Issue #413, opened 2016-11-01
pvs studio static analysis amd64


igouss:
FYI

[out1.tasklist.txt.zip](https://github.com/eclipse/omr/files/565262/out1.tasklist.txt.zip)

mgaudet:
Thanks! There's certainly some things in there that ought to be looked at. 

I'm going to try to triage this slowly into some new issues, but much appreciated :D

amankumar1:
I have made a summary of all the errors and warnings from the file, sorted by the total count of each error and warning encountered along with a brief description. See `summary.txt`.

There is a separate text file for each error code in the `warn` and `err` folders. Each text file contains the complete PVS Studio output for all occurrences of that error or warning.

[pvs.zip](https://github.com/eclipse/omr/files/579385/pvs.zip)

mgaudet:
Just so people don't have to download the zip file, I have put up [`summary.txt` in a Gist](https://gist.github.com/mgaudet/91cff155b5f370b6c915bcf4025aefac) Thanks for formatting the list @amankumar1 

IMO, there's some clear choices for things we ought to look at early: 

- [ ] `V595 : 48  : Pointer was utilized before it was verified against nullptr` (modern
- [ ] `V614 : 13  : Potentially uninitialized variable used`
- [ ] `V561 : 5   : It's probably better to assign value to a pertivular variable than to declare it anew` (potentially dangerous variable shadowing) 
- [ ] `V512 : 2   : Call to function leads to buffer overflow`
- [ ] `V598 : 9   : 'memcpy' is used to copy fields of class. Virtual table pointer damaged`

There's also some silly low hanging fruit: 

- [ ] [This silly ternary!](https://github.com/eclipse/omr/blob/830ebbb/compiler/optimizer/OMRSimplifierHandlers.cpp#L11045)

mgaudet:
@0xdaryl Can we get this labelled with 'help wanted'?

PR #532, opened 2016-11-24
Fix ternary statement in optimizer

amankumar1:
The ternary statement of the form `a?b:b` was changed to `b`.

Issue: #413

Signed-off-by: Aman Kumar <amank@ca.ibm.com>


vijaysun-omr:
Simple cleanup. Merging.

PR #535, opened 2016-11-25
Replace CS2/HashTab with std/map in debug counters

doubletea:
Currently, `CS2::HashTable` is the data structure for keeping track unique debug counters in DebugCounterGroup. Since CS2 elements are moving away, migrate the table to `std::map`, which is also better documented.

All `CS2::HashTable` APIs that are used for debug counters should functionally match the `std::map` equivalent.

Signed-off-by: John Xu <xujohn@ca.ibm.com>

### DIFF
--- a/compiler/ras/DebugCounter.cpp
+++ b/compiler/ras/DebugCounter.cpp
@@ -365,14 +365,12 @@ TR::DebugCounter *TR::DebugCounterGroup::findCounter(const char *nameChars, int3
    strncpy(name, nameChars, nameLength);
    name[nameLength] = 0;
 
-   // There's a race here if we do parallel compilation
-
    OMR::CriticalSection findCounterLock(_countersMutex);
 
-   CS2::HashIndex hi;
-   if (!_countersHashTable.Locate(name, hi))
+   auto search = _countersHashTable.find(name);
+   if(search == _countersHashTable.end())
       return NULL;
-   return _countersHashTable[hi];
+   return search->second;
    }
 
 TR::DebugCounterAggregation *TR::DebugCounterGroup::createAggregation(TR::Compilation *comp)
@@ -430,12 +428,10 @@ TR::DebugCounter *TR::DebugCounterGroup::createCounter(const char *name, int8_t 
    TR::DebugCounter *result = new (persistentMemory) TR::DebugCounter(name, fidelity, denominator, flags);
    TR_ASSERT(result, "DebugCounter *result must not be null. Ensure availability of persistent memory");
    _counters.add(result);
-   
-   // There's a race here if we do parallel compilation
 
    OMR::CriticalSection createCounterLock(_countersMutex);
 
-   _countersHashTable.Add(result->getName(), result);
+   _countersHashTable.insert(std::make_pair(result->getName(), result));
 
    return result;
    }
--- a/compiler/ras/DebugCounter.hpp
+++ b/compiler/ras/DebugCounter.hpp
@@ -21,16 +21,17 @@
 
 #include "ras/Debug.hpp"
 
-#include <stdarg.h>            // for va_list
-#include <stddef.h>            // for NULL
-#include <stdint.h>            // for int32_t, int8_t, uint32_t, int64_t, etc
-#include "cs2/hashtab.h"       // for HashTable
-#include "env/CompilerEnv.hpp" // for target->is64Bit()
-#include "env/TRMemory.hpp"    // for TR_MemoryBase::ObjectType::DebugCounter, etc
-#include "env/jittypes.h"      // for intptrj_t
-#include "infra/Flags.hpp"     // for flags8_t
-#include "infra/List.hpp"      // for TR_PersistentList
-#include "infra/Monitor.hpp"   // for createCounter race conditions
+#include <stdarg.h>                // for va_list
+#include <stddef.h>                // for NULL
+#include <stdint.h>                // for int32_t, int8_t, uint32_t, int64_t, etc
+#include <map>                     // for _countersHashTable
+#include "env/TypedAllocator.hpp"  // for typed_allocator
+#include "env/CompilerEnv.hpp"     // for target->is64Bit()
+#include "env/TRMemory.hpp"        // for TR_MemoryBase::ObjectType::DebugCounter, etc
+#include "env/jittypes.h"          // for intptrj_t
+#include "infra/Flags.hpp"         // for flags8_t
+#include "infra/List.hpp"          // for TR_PersistentList
+#include "infra/Monitor.hpp"       // for createCounter race conditions
 
 namespace TR { class Compilation; }
 namespace TR { class Node; }
@@ -229,7 +230,20 @@ public:
 
 class DebugCounterGroup
    {
-   CS2::HashTable<const char*, DebugCounter*, TRPersistentMemoryAllocator> _countersHashTable;
+   struct debugCounterNameComp
+      {
+      bool operator()(const char* lhs, const char* rhs) const
+         {
+         while (*lhs == *rhs)
+            {
+            if (*lhs==0) return false;
+            lhs+=1; rhs+=1;
+            }
+         return *lhs > *rhs ? true : false;
+         }
+      };
+
+   std::map<const char*, DebugCounter*, debugCounterNameComp, TR::typed_allocator<std::pair<const char*, DebugCounter*>, TRPersistentMemoryAllocator>> _countersHashTable;
    TR_PersistentList<DebugCounter> _counters;
    TR_PersistentList<DebugCounterAggregation> _aggregations;
    DebugCounter *createCounter (const char *name, int8_t fidelity, TR_PersistentMemory *mem);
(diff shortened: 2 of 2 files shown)


compiler/ras/DebugCounter.hpp
```diff
@@ -229,7 +230,20 @@ class DebugCounterAggregation : public DebugCounterBase
...
 class DebugCounterGroup
    {
-   CS2::HashTable<const char*, DebugCounter*, TRPersistentMemoryAllocator> _countersHashTable;
+   struct debugCounterNameComp
+      {
+      bool operator()(const char* lhs, const char* rhs) const
+         {
+         while (*lhs == *rhs)
+            {
+            if (*lhs==0) return false;
```
andrewcraik: Don't you need something similar for rhs?

doubletea: I think *rhs==0 is implied by the while condition, so in this case the two char arrays are the same. We want to return true only if the lhs char array is greater than the rhs char array.

andrewcraik: ok yes - I buy that

doubletea:
To elaborate, the `CS2::HashTable` in debug counters is used to ensure we have unique names; when we increment a counter that already exists, it uses the corresponding debug counter object instead of making a new one with the same name. This uniqueness is ensure by the compare struct `debugCounterNameComp` (see the [cppreference on std::map](http://en.cppreference.com/w/cpp/container/map))

doubletea:
In testing std::map with all debug counters turned on, it passes all the same tests on all platforms that the current debug counter does with CS2::HashTable. The debug counter numbers match exactly with CS2::HashTable for test cases I performed on z/OS and z/Linux. Quick tests shows the numbers also agree on X and P.

charliegracie:
I was recently doing some work using STL map and I noticed it performed very poorly on OSX compared to Linux x86.  Have you performed any performance verification on the speed at which you can add / remove / find elements in the STL map?  If so which platforms have you compared?

andrewcraik:
A significant performance difference would be worrying since it risks perturbing the state of the system by increasing compile times when trying to instrument the system to find out what is going wrong so performance is important for these tables from the point of view of not being significantly worse.

doubletea:
I performed some quick tests on z/Linux64, x86-64/Linux, and ppc64le-64/Linux by running java -version with all debug counters enabled and count=0. Even this simple test case creates 10,000+ debug counters and many times that many increments across all platforms. For ten iterations of java -version, when I compare CS2::HashTable to STL Map in terms of compile time (`TR_PrintCompTime=1`), on average:

- Map is about 5% faster on z/Linux64
- Map is about 10% faster on z/Linux64
- Map is about the same as CS2 on ppc64le-64/Linux

Of course, this test case is not the most thorough, so I am currently setting up ilog performance runs as well. 

I am pretty confident that STL Map does not perform significantly worse than the current CS2 Hashtable, at least in the context of debug counters. Since a goal is to remove the CS2 libraries in the future, how important is a slight performance difference towards this?

andrewcraik:
Slight performance degradation is not a major issue to me, but degradation that affects order, time, or footprint of compilations to the point where the compilation order changes, the footprint changes to cause compilations to fail, or the time for compilation increases significantly are concerns since we do not want turning the counters on to perturb the system in a noticeable way (thereby altering the system we are trying to observe). I think for evaluating STL changes we need to make sure we evaluate the platforms where there are different STL implementations (ie Linux will be fairly common and standardized, but Windows etc will be different and we need to ensure things are consistently ok).

mstoodle:
Any update on this, @doubletea ? Could we at least do the same tests you did for Linux on Z and POWER on Windows to see if there's any sign of different performance behaviour?

doubletea:
Yes, sorry I had difficulties running the ilog benchmarks on all the platforms, and I wanted to wait on that before I gave some more performance numbers. For the same quick tests I did in my last comment, I did on x86-64 Windows and STL Map turned out to have nearly the same compile time on average, within about 5%.

I think for most use cases for debug counters, there are not going to be thousands of different ones. And if a developer chooses to have so many, they should be aware of the performance impacts regardless of how much the underlying data structure may be slightly better or worse depending on implementation.

andrewcraik:
I would point out that because debug counter names often contain things like compile levels, method names, block numbers and bytecode indices it is not unusual to have a few thousand enabled at any one time. A good example would be the coldBlock* counters from CodeGenPrep which basically puts a unique counter in each cold block.

The main platform of concern is Linux on the major codegen platforms since those are the ones where the majority of debug counter use occurs at present. Each compiler will have a different STL implementation and so the effects of the change may well be different.

@doubletea could you clarify the measurement variance vs the 5% delta - is the 5% significant or in the noise?

@mpirvu any comments on this one? I know it isn't on the critical path for startup/rampup but you have been doing a lot of digging on STL and I know that we have used debug counters in some cases to help investigate those problems.

doubletea:
@andrewcraik It was in the noise, +/- 5%

mpirvu:
I am ok with the change given that debug counters are not used in production and that performance delta between this STL based implementation and the CS2 based implementation is small.
I was never a fan of CS2::HashTable. A while ago when we started to migrate to CS2 we had a big footprint regression due to replacing some data structure (I don't remember which one) with CS2::HashTable.
Speaking of which, have we looked at impact on JIT scratch memory requirements?

andrewcraik:
@mpirvu do you think we should merge the change, but use it as a test bed for footprint growth ahead of other such refactorings? Or should we do the assessment first?

compiler/ras/DebugCounter.cpp
```diff
@@ -361,18 +361,19 @@ TR::DebugCounter *TR::DebugCounterGroup::findCounter(const char *nameChars, int3
    {
    if (nameChars == NULL)
       return NULL;
+
+   // Null terminate the name
    char *name = (char*)alloca(nameLength+1);
    strncpy(name, nameChars, nameLength);
    name[nameLength] = 0;
 
    // There's a race here if we do parallel compilation
```
mgaudet: Is this comment still relevant given the locking below?

mpirvu:
We can do the assessment first because it's not complicated: same tests, but adding verbose={compilePerformance},vlog=vlog.txt  to the jit command line option. The verbose log will have lines like:

`+ (warm) java/lang/Integer.intValue()I @ 0x000007FEBD4BDE60-0x000007FEBD4BDE80 OrdinaryMethod - Q_SZ=0 Q_SZI=0 QW=2 j9m=000000000F0665B0 bcsz=5 sync time=542us mem=[region=1984 system=16384]KB compThread=0 CpuLoad=106%(13%avg) JvmCpu=71%`

**mem=[region=1984 system=16384]KB**  is what we need to look at. This has changed from what I was used to. I think that "system" represents the size of all segments allocated from VM and "region" is how much we actually use from those segments. Luc would know for sure @lmaisons

compiler/ras/DebugCounter.cpp
mgaudet: (this is about the lock, not the name null termination)

compiler/ras/DebugCounter.cpp
```diff
@@ -361,18 +361,19 @@ TR::DebugCounter *TR::DebugCounterGroup::findCounter(const char *nameChars, int3
    {
    if (nameChars == NULL)
       return NULL;
+
+   // Null terminate the name
```
0xdaryl: I don't think the "Null terminate" comment is adding much value.  This is a common pattern and its fairly obvious from the code what it's up to.

This did, however, draw my eye to this code and the `alloca` call seems unsafe from a security (stack overflow) perspective if the debug counter name and length passed in can be arbitrary and specified by a user.  I'll create an issue to investigate this.  This appears to be the only `alloca` usage in the OMR compiler code.  I think there are safer alternatives to use.

lmaisons:
> I think that "system" represents the size of all segments allocated from VM and "region" is how much we actually use from those segments.

'system' is quantity of memory we have requested from the operating system.  This is done (for Java) in 16 MiB chunks.  From those chunks we carve up TR::MemorySegments for use by Regions.  Unless a TR::MemorySegment is a non-standard size, once created, it will stick around until the end of a compilation.  'region' is the amount of memory we have made available as TR::MemorySegments.

compiler/ras/DebugCounter.cpp
doubletea: Good point, that comment is not needed anymore.

compiler/ras/DebugCounter.cpp
doubletea: That's fair, the comment is not necessary.

doubletea:
Something went wrong with the branch, fixing it now.

lmaisons:
Is this still wanted?  What's needed to get this restarted?

andrewcraik:
probably needs a rebase and retest

mpirvu:
Looking through the comments I see that we decided we are going to measure the impact of this change on JIT scratch memory. Unless Testarossa code already computes some averages that are printed at shutdown time, we need a verbose log and compute stats on the region sizes:
mem=[region=1984 system=16384]KB

mstoodle:
Ok, let's try to figure out a solution for this PR or else it seems like it will just sit here forever.

@mpirvu what is your preference at this point? I'm not sure @doubletea will be able to provide the measurements you asked for (please correct me if I'm wrong). Do you want to do the measurements? @fjeremic did the motivation for this change originally come from you, or do you know what the motivation was?

The PR will need to be rebased before we accept it, as @andrewcraik indicates, but it doesn't seem like anyone has any hard blockers (just potential concerns) to merge.

mstoodle:
Ok, sounds good @fjeremic. Thanks for the quick response!

PR #536, opened 2016-11-25
 Add late block ordering optimization to JitBuilder warm strategy 

ShuyuLi-ca:
Add an optimization strategy called `basicBlockOrdering` in order to
clean up block orders. Set the methodSymbol() in `IlBuilder` in order to
use `basicBlockOrdering`.

Signed-off-by: Shuyu Li <shuyuli@ca.ibm.com>

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1799,7 +1799,8 @@ IlBuilder::ForLoop(bool countsUp,
                    TR::IlValue *increment)
    {
    //ILB_REPLAY_BEGIN();
-
+   
+   methodSymbol()->setMayHaveLoops(true);
    TR_ASSERT(loopCode != NULL, "ForLoop needs to have loopCode builder");
    *loopCode = createBuilderIfNeeded(*loopCode);
 
@@ -1868,6 +1869,7 @@ IlBuilder::DoWhileLoop(const char *whileCondition, TR::IlBuilder **body, TR::IlB
    {
    //ILB_REPLAY_BEGIN();
 
+   methodSymbol()->setMayHaveLoops(true);
    TR_ASSERT(body != NULL, "doWhileLoop needs to have a body");
 
    if (!_methodBuilder->symbolDefined(whileCondition))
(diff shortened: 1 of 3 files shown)


mstoodle:
I think I would prefer the pass to be run "ifEnabled" and to have generalStoreSinking enable it. Refinement would be if store sinking inserts stores along edge placements since that is what introduces new basic blocks. With this approach, other opts can also enable block ordering if they introduce new blocks.

Arbitrarily setting "mayHaveLoops" so that block ordering will run IfLoops is not the right way to make it happen.

mstoodle:
Tchniceally, you're not adding an optimization strategy, you're adding an optimization to the JitBuilder warm optimization strategy. Perhaps "Add late block ordering optimization to JitBuilder warm strategy" would be a more appropriate title, @ShuyuLi-ca ?

ymanton:
> Arbitrarily setting "mayHaveLoops" so that block ordering will run IfLoops is not the right way to make it happen.

These are probably two separate (but related) issues. Many opts check whether `mayHaveLoops` is set or not, but it never seems to be set by anyone other than the bytecode iterator, so @ShuyuLi-ca is setting it in the JitBuilder loop services as a starting point.

I don't think that's enough though; one can construct a loop without using those services, so this flag needs to get set somehow since it is checked by many opts.

ShuyuLi-ca:
@mstoodle @ymanton , I just changed a little bit, thanks a lot.

mstoodle:
Thanks @ymanton .

@ShuyuLi-ca I think these two changes should be split into two separate commits (one for the "may have loops" and one to add the late block ordering pass, since they do different things and afaik do not depend on one another.

ymanton:
Can you change the subject from "Add a service to check whether have loops" to something like "Set mayHaveLoops flag in JitBuilder loop services" instead? A service is one of the high-level APIs that JitBuilder provides to clients; this change is simply setting a flag in some of the existing services.

mstoodle:
It looks good at this point.

I know the motivation for this work was to improve the ordering for the blocks introduced by store sinking when it had to split a control flow edge to insert the store only onto the path where it's live.

@ShuyuLi-ca @ymanton Did this change result in improving the position of those blocks?

ymanton:
@mstoodle Yes, it does the trick, the new blocks are put in the fall-through position rather than having to be jumped to.

mstoodle:
Ok great! Thanks @ymamton !

PR #538, opened 2016-11-27
fix build on osx 10.12

gurinderu:
Replace deprecated syscall() interface for macOS

syscall() has been deprecated and is not available on macOS 10.12. 

This fixes the build on macOS 10.12, by replacing the syscall for threadID with 
pthread_threadid_np on macOS

Signed-off-by: Nick Pavlov <gurinderu@gmail.com>

### DIFF
--- a/thread/unix/rasthrsup.c
+++ b/thread/unix/rasthrsup.c
@@ -27,6 +27,7 @@
 #include <sys/syscall.h>
 #endif /* __GLIBC_PREREQ(2,4) */
 #elif defined(OSX)
+#include <pthread.h>
 #include <sys/syscall.h>
 #endif /* defined(LINUX) */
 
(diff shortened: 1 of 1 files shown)


mgaudet:
Thanks for splitting this apart: I do think we need a better commit message though. I would suggest something along the lines of 

    Replace deprecated syscall() interface for macOS

    syscall() has been deprecated and is not available on macOS 10.12. 
    
    This fixes the build on macOS 10.12, by replacing the syscall for threadID with 
    pthread_threadid_np on macOS

See CONTRIBUTING.md for our commit message guidelines.

gurinderu:
@mgaudet done

charliegracie:
After looking around it appears as though pthread_threadid_np was added in OSX 10.6 so we would no longer support releases older than that once we apply that change.  I think that it is ok but @mstoodle what is your opinion as an OSX user?

mstoodle:
@charliegracie 10.6 is already pretty ancient (released in 2009, I believe) .

I am ok to "limit" us to 10.6 and later :) . I don't have a good feeling for how much the earlier versions of OSX are used, though, so not sure how far forward we can take that statement at the moment. But that shouldn't hold up this commit and doesn't seem like urgent question to answer at the moment.

Thanks for the contribution, @gurinderu !

charliegracie:
Thanks!

PR #543, opened 2016-11-29
Add linkage information for each runtime helper

0dvictor:
As there are many different ABI, each runtime helper potentially
requires a specific ABI. Existing code does not take this important
information, and hence this PR proposes adding linkage convention
to each runtime helper.

### DIFF
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -25,7 +25,7 @@
 #include <string.h>                            // for strlen, strncmp, etc
 #include "codegen/CodeGenerator.hpp"           // for CodeGenerator
 #include "codegen/FrontEnd.hpp"                // for TR_FrontEnd, feGetEnv
-#include "env/KnownObjectTable.hpp"        // for KnownObjectTable, etc
+#include "env/KnownObjectTable.hpp"            // for KnownObjectTable, etc
 #include "codegen/LinkageConventionsEnum.hpp"
 #include "codegen/Machine.hpp"                 // for Machine
 #include "codegen/RealRegister.hpp"            // for RealRegister
@@ -677,11 +677,12 @@ OMR::SymbolReferenceTable::createIsOverriddenSymbolRef(TR::ResolvedMethodSymbol 
 
 
 TR::SymbolReference *
-OMR::SymbolReferenceTable::createRuntimeHelper(
-   TR_RuntimeHelper index, bool canGCandReturn, bool canGCandExcept, bool preservesAllRegisters)
+OMR::SymbolReferenceTable::createRuntimeHelper(TR_RuntimeHelper index,
+                                               bool             canGCandReturn,
+                                               bool             canGCandExcept,
+                                               bool             preservesAllRegisters)
    {
-
-   TR::MethodSymbol * methodSymbol = TR::MethodSymbol::create(trHeapMemory(),TR_Helper);
+   TR::MethodSymbol * methodSymbol = TR::MethodSymbol::create(trHeapMemory(),runtimeHelperLinkage(index));
    methodSymbol->setHelper();
    methodSymbol->setMethodAddress(runtimeHelperValue(index));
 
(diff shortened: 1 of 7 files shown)


0dvictor:
As the linkage table may be zero initialized, the linkage may be incorrectly reported as TR_Private if unset; hence, this PR also introduce a TR_None linkage to avoid this situation.

0dvictor:
Thank you for the feedback, I've updated code with comments.

vijaysun-omr:
I would like @andrewcraik to review this please.

compiler/compile/OMRSymbolReferenceTable.hpp
```diff
@@ -240,8 +240,8 @@ class SymbolReferenceTable
    TR::SymbolReference * getSymRef(CommonNonhelperSymbol i);
    TR::SymbolReference * getSymRef(int32_t i) { return baseArray.element(i); }
 
-   TR::SymbolReference * createRuntimeHelper(TR_RuntimeHelper, bool = false, bool = false, bool preservesAllRegisters = false);
```
andrewcraik: Can you explain the rational in removing the default parameter values? It would also be more ideal to have names for the bool parameters in the header so it is easy to find out what the different params are and what they mean without having to drill into the cpp.

compiler/compile/OMRSymbolReferenceTable.cpp
```diff
@@ -713,7 +717,7 @@ OMR::SymbolReferenceTable::findOrCreateCodeGenInlinedHelper(CommonNonhelperSymbo
    {
    if (!element(index))
       {
-      TR::MethodSymbol * sym = TR::MethodSymbol::create(trHeapMemory(),TR_Helper);
+      TR::MethodSymbol * sym = TR::MethodSymbol::create(trHeapMemory(),TR_None);
```
0dvictor: This change should not cause behavior change since CodeGenInlinedHelper are not, and should not be, handled by any linkage; instead, CodeGen recognizes them and generates inlined assembly. Therefore, they do not use any of existing linkage convention, and hence, I believe, it is a perfect example that one helper with no linkage, noted as TR_None.

compiler/compile/OMRSymbolReferenceTable.hpp
0dvictor: The only caller to createRuntimeHelper(), findOrCreateRuntimeHelper() always gives all parameters. In addition, the function looks a dedicated helper of findOrCreateRuntimeHelper() and should not be called by anyone else; and hence allowing createRuntimeHelper() have default parameters looks confusing to me.
For the second suggestion, I agree it makes the code easier to understand by having names with those bool parameters, I'll update them.

0dvictor: Updated the function declarations with names of each parameter.

compiler/compile/OMRSymbolReferenceTable.cpp
vijaysun-omr: Thanks for your answer. So anything in the category of CodeGenInlinedHelper is appropriate for the TR_None linkage convention; there are other examples of helpers that require actual "helper linkage" I assume and that is why TR_Helper is still a category in the new enum correct ?

0dvictor: Correct. For helpers that require actual linkage, it may need TR_Helper, TR_System, TR_Private, or even other enum, depending on the implementation of the implementation of the helper. For example, TR_newObject needs TR_Helper,  TR_AMD64getTimeOfDay needs TR_System, TR_volatileReadLong needs TR_Private, etc.

vijaysun-omr:
I will merge this once the CI build passes and @andrewcraik is happy with the responses to his questions.

compiler/runtime/Runtime.hpp
```diff
@@ -163,21 +165,46 @@ enum TR_RuntimeHelper
...
+   TR_LinkageConventions getLinkage(TR_RuntimeHelper h)
+      {
+      return h < TR_numRuntimeHelpers ? _linkage[h] : TR_None;
+      }
+   // Linkage convention is essential when calling a helper
+   // For example, on X86, there are private linkage, System V ABI, fastcall, cdecl, etc.
+   void setAddress(TR_RuntimeHelper h, void * a, TR_LinkageConventions lc = TR_Private)
+      {
+      _helpers[h] = translateAddress(a);
+#ifdef TR_TARGET_X86
```
andrewcraik: ok we can live with this for now I guess.

compiler/compile/OMRSymbolReferenceTable.hpp
andrewcraik: ok that makes more sense now - thanks for the change

andrewcraik:
@vijaysun-omr I think this is now good to merge

vijaysun-omr:
I will wait for Matthew's questions to be addressed before merging.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -490,6 +490,8 @@ OMR::X86::CodeGenerator::createLinkage(TR_LinkageConventions lc)
 
    switch (lc)
       {
+      case TR_None:
```
0dvictor: Good suggestion, updated.

andrewcraik:
@vijaysun-omr I think Matthew's question has been answered so this should be good to merge.

PR #544, opened 2016-11-30
Code Cleanup - Remove unused function declaration

0dvictor:
The function declaration is never used, and hence should be removed.


vijaysun-omr:
Simple cleanup. Merging.

PR #545, opened 2016-11-30
Upgrade concurrentScavengerInProgress flag to a toggle counter

amicic:
There is a global and a thread local counter. Whenever a local counter
is behind the global one, a concurrent switch  should occur, and update
local counter. Global counter is incremented on every start and end STW
phase. 
The counter is needed to help with scenarios when a thread goes asleep
during one GC cycle and gets awaken during another GC cycle. We still
need to trigger the switch, since relevant parameters (like Evacuate
boundaries) change from a cycle to a cycle. 


charliegracie:
Just for documentation, this could in fact still be an issue if the thread was asleep for a full 64bit's worth of GC cycles exactly but we can revisit this in the future if we think it is going to cause any issues.

PR #547, opened 2016-11-30
Add support for CUDA 8.0

keithc-ca:
This change adds support for discovery of the CUDA version 8.0 runtime shared library and ties the minimum supported runtime to the version of the toolkit used at compile time (instead of hard-coding a lower bound of version 5.5).

Issue: 546


charliegracie:
@keithc-ca could you update the commit message to be more similar to the Pull Request initial comment.  Having those details in the commit message make it much easier to follow what changes were made by looking at the Git log.  Thanks

keithc-ca:
Rebased and reworded the commit message.

Issue #380, opened 2016-10-25
Decouple the OMR Compiler's Error Codes and Return Codes


dsouzai:
The current error handling story in the compiler involves (ideally) setting an error code using the `setErrorCode` method, and then calling the `outOfMemory` which throws a `TR::CompilationException`. This happens when an error occurs deep in a compile such that, currently, there **isn't** a way to consistently propagate the error up without throwing an exception. Unfortunately, the error codes that are set using `setErrorCode` are sometimes also used as return codes that the control logic will check and then throw an exception. This happens when there we are not deep in a compile and it **is** possible to consistently propagate the error up without throwing an exception. As such, there are inconsistencies with the concepts of return and error codes. 

The plan is:
1. Replace all uses of `outOfMemory` with the throwing of appropriate C++ exceptions. 
2. Remove all uses of `setErrorCode`
3. Evaluate the #defines and remove all except those that are used as return codes

By getting rid of `outOfMemory` and `setErrorCode` calls, we remove the need to determine the type of error by checking the error code since we can simply catch the appropriate exceptions. Once this is done, step 3 is simply a cleanup step.

dsouzai:
From what I can tell, the only valid return codes are:
COMPILATION_SUCCEEDED
COMPILATION_IL_GEN_FAILURE
COMPILATION_UNIMPL_OPCODE
and the one Xiaili is adding in #375 (COMPILATION_REQUESTED)

perhaps we also need one along the line of COMPILATION_FAILED for the cases when an exception is thrown.

dsouzai:
These are the list of error codes that, once #412 gets merged, are replaced by TR::CompilationException or it's children:

Excessive Complexity Exception
    COMPILATION_LOOPS_OR_BASICBLOCKS_EXCEEDED
    COMPILATION_REMOVE_EDGE_NESTING_DEPTH_EXCEEDED
    COMPILATION_MAX_DEPTH_EXCEEDED
    COMPILATION_MAX_NODE_COUNT_EXCEEDED
    COMPILATION_INTERNAL_POINTER_EXCEED_LIMIT

Max caller index exceeded Exception
    COMPILATION_MAX_CALLER_INDEX_EXCEEDED

Insufficiently Aggressive Compilation Exception
    COMPILATION_NEEDED_AT_HIGHER_LEVEL
    
Recoverable Code Cache Error Exception
    COMPILATION_ALL_CODE_CACHES_RESERVED
[...]

PR #548, opened 2016-11-30
Remove unused error codes from the compiler

dsouzai:
This PR completes step 3 in #380. The only thing a compiler returns is COMPILATION_SUCCEEDED. All other error conditions are described via C++ exceptions. The compileMethodFromDetails method sets the few remaining return codes depending on the exception thrown. The reason these few remaining return codes continue to exist is because it is not trivial to remove their dependency from the various FVTests. However, as a step toward cleaner code, this PR also includes a change that replaces the remaining `#defined`'d return codes with an enum.


dsouzai:
@mpirvu  @0xdaryl Could you guys please review.

fyi @xliang6 (since it is related to the COMPILATION_REQUESTED change you added a little while ago).

0xdaryl:
The changes look good to me.  I'll wait for @mpirvu 's opinion before merging.

mpirvu:
Changes are good and ca be delivered

PR #549, opened 2016-11-30
Changes to improve the split and splitEdge APIs for OMR::Block

cathyzhyi:


charliegracie:
It appears there is an issue when running the linter with this Pull Request.  You can view this issue by looking into the continuous integration failure.

mgaudet:
```
../../compiler/il/OMRBlock.cpp:797:11: error: Implicit this receivers are prohibited in extensible classes
   return split(startOfNewBlock, cfg, fixupCommoning, copyExceptionSuccessors, methodSymbol);
          ^
          self()->
1 error generated.
make[1]: *** [../../compiler/il/OMRBlock.cpp.linted] Error 1
```

(just to avoid having to dig around a bit :smile: )

cathyzhyi:
@vijaysun-omr The fix was because of a crash I found when forcing fsd and disaliasing fsd and mimic interpreter shape. When loop strider convert IV back to 32 bits it splits all the exit edges and there happens to be an OSR catch block in the failing case. @jdmpapin

vijaysun-omr:
Is there an exception edge from the new OSR catch block created by splitEdge to the original OSR catch block ? How does flow of control actually get transferred from the new OSR catch block to the original OSR catch block ?

cathyzhyi:
@vijaysun-omr, the control flow is like: fromBlock->**exception** edge->conversion block(the new block)-> **normal** edge-> toBlock(OSRCatch block).

vijaysun-omr:
Okay that makes sense. I will merge this now.

PR #554, opened 2016-12-01
Define TR_HOST_64BIT correctly for Power

### LINKED ISSUE #553: TR_HOST_64BIT is defined wrongly
TR_HOST_64BIT was intended to mean the build rather than the underlying ISA architecture.  Currently OMR defines.h defines it according to the underlying ISA architecture.  As a result, 32bit build defines TR_HOST_64BIT too. All files can be compiled successfully but most wrongly. 

Currently:
#  if defined(__PPC64__) || defined(_ARCH_PPC64)
#    define TR_HOST_64BIT 1

Fix:
#  if defined(__PPC64__) || defined(__64BIT__)
#    define TR_HOST_64BIT 1

zl-wang:
TR_HOST_64BIT was not intended for the underlying ISA architecture but the build.
Fixed with compiler pre-defined macros instead:

Fixes issue:  #553

### DIFF
--- a/compiler/env/defines.h
+++ b/compiler/env/defines.h
@@ -111,7 +111,7 @@
 #  define TR_HOST_32BIT 1
 #elif defined(__PPC__) || defined(_ARCH_PPC)
 #  define HOST_ARCH ARCH_POWER
-#  if defined(__PPC64__) || defined(_ARCH_PPC64)
+#  if defined(__PPC64__) || defined(__64BIT__)
 #    define TR_HOST_64BIT 1
 #  else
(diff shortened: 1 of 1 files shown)


zl-wang:
Done.

Yes, the fix appeared to be good for AIX64/32,  LinuxBE 64/32, and LinuxLE (64), and at the same time for at least two types of compilers I tested.

For details, refer to the issue #553 itself.   That was the guideline given the other day.

"Fixes #number" format conforms to the guideline as well though:  https://help.github.com/articles/closing-issues-via-commit-messages/

Issue #465, opened 2016-11-10
Add Clang Linux64 platform


noryb009:
The Test compiler and JitBuilder do not currently have `amd64-linux64-clang` as a supported platform. This shouldn't be too difficult, since it did work at some point.

A good starting place is copying [this file](https://github.com/rubyomr-preview/rbjitglue/blob/master/ruby/build/platform/host/amd64-linux64-clang.mk) to either the Test compiler or JitBuilder, then running `PLATFORM=amd64-linux64-clang make`.

PR #556, opened 2016-12-01
Support clang in the Test compiler and JitBuilder

noryb009:
This adds support for building the Test compiler and JitBuilder using clang/LLVM on Linux x86-64.

Issue: #465


fvtest/compilertest/build/toolcfg/gnu/common.mk
```diff
@@ -26,18 +26,27 @@ SED_PATH?=sed
 AR_PATH?=ar
 PERL_PATH?=perl
 
+ifeq ($(OS),osx)
```
mstoodle: This seems unintuitive enough that it deserves at least a comment to explain. "as" works for me on OSX. Is there some version of OSX/clang where "clang" is the right way to access the assembler (or have I misunderstood) ?

jitbuilder/build/toolcfg/gnu/common.mk
```diff
@@ -21,18 +21,27 @@ SED_PATH?=sed
 AR_PATH?=ar
 PERL_PATH?=perl
 
+ifeq ($(OS),osx)
```
mstoodle: same comment here

noryb009:
Despite `__ --version` being the same for `as` and `clang`, `as` doesn't appear to support the `-D` flag (at least, on the machine I have access to). I did add a comment to explain the code.

mstoodle:
Thanks @noryb009 !

PR #557, opened 2016-12-01
Avoid duplicated reporting fragmentation info in gc-start tag of verbosegclog

LinHu2016:
reset fragmentation flag after end of gc reporting to avoid duplicated
fragmentation information inside tag gc-start in some cases.


charliegracie:
@LinHu2016 The commit message does not match the actual change.  Can you please update it?

PR #558, opened 2016-12-02
Fix bugs in getting CPU number on IBM i

bjxyji:
CPU type OMRPORT_CPU_TARGET is not supported on IBM i, so we get CPU
number directly through _system_configuration.ncpus.

Signed-off-by: xiang yang ji <bjxyji@cn.ibm.com>


charliegracie:
Thanks!  I will merge as soon as the travis build passes

PR #560, opened 2016-12-02
Optimize Object.clone only for cloneable objects

liqunl:
In java, only instances of classes that implement Cloneable can be
cloned, otherwise a CloneNotSupportedException will be thrown. This fix
checks whether the object is cloneable before adding the object to the
candidate list and re-enable the optimization.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -5704,6 +5704,16 @@ TR::Node *constrainAcall(TR::ValuePropagation *vp, TR::Node *node)
             TR::VPResolvedClass *newTypeConstraint = NULL;
             if (constraint)
                {
+               // Do nothing if the class of the object doesn't implement Cloneable
+               if (constraint->getClass() && !vp->comp()->fej9()->isCloneable(constraint->getClass()))
+                  {
+                  if (vp->trace())
+                     traceMsg(vp->comp(), "Object Clone: Class of node %p is not cloneable, quit\n", node);
+
+                  TR::DebugCounter::incStaticDebugCounter(vp->comp(), TR::DebugCounter::debugCounterName(vp->comp(), "inlineClone/unsuitable/(%s)/%s/block_%d", vp->comp()->signature(), vp->comp()->getHotnessName(vp->comp()->getMethodHotness()), vp->_curTree->getEnclosingBlock()->getNumber()));
+
+                  return node;
+                  }
                if ( constraint->isFixedClass() )
                   {
                   newTypeConstraint = TR::VPFixedClass::create(vp, constraint->getClass());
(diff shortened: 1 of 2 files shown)


liqunl:
@andrewcraik Andrew, can you review my commit. It has been tested with acceptance build and should be merged after the gitlab MR is closed.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -5704,6 +5704,13 @@ TR::Node *constrainAcall(TR::ValuePropagation *vp, TR::Node *node)
             TR::VPResolvedClass *newTypeConstraint = NULL;
             if (constraint)
                {
+               // Do nothing if the class of the object doesn't implement Cloneable
+               if (constraint->getClass() && !vp->comp()->fej9()->isCloneable(constraint->getClass()))
+                  {
+                  if (vp->trace())
+                     traceMsg(vp->comp(), "Object Clone: Class constraint on node %p doesn't exist or the class is not cloneable, quit\n", node);
+                  return node;
```
andrewcraik: I don't think we want to return early here since it will mess up the hit/miss debug counters below. This check probably needs to go down around the code where we add the clone to the transformation list and we should probably move the debug counter logic down there so that the counters count whether we hit or missed on adding it to the transformation list.

liqunl: Do we really want to include Object.clone on a uncloneable object into the hit/miss debug counters? A miss on such object is expected thus should not be considered as a miss.

andrewcraik: That is a really good point - maybe we should add an unsuitable counter as well just so we know when we hit this case then?

liqunl: Yes, we can do that. Will do the change and in a moment.

liqunl: New change is recompiled and retested on X86, and the debug counter is successfully incremented on Object.clone test.

andrewcraik:
The change looks good to merge once the CI passes.

liqunl:
@andrewcraik What is CI?

andrewcraik:
CI = continuous integration - eg the travis-ci checks which are currently outstanding for the item :)

andrewcraik:
@vijaysun-omr could you review and merge if this looks good - it seems fine to me and will allow an optimization to be re-enabled as well as solving an OMR issue.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -5704,6 +5704,16 @@ TR::Node *constrainAcall(TR::ValuePropagation *vp, TR::Node *node)
             TR::VPResolvedClass *newTypeConstraint = NULL;
             if (constraint)
                {
+               // Do nothing if the class of the object doesn't implement Cloneable
+               if (constraint->getClass() && !vp->comp()->fej9()->isCloneable(constraint->getClass()))
```
andrewcraik: I don't think the site would have been queued for transformation, but I agree this could be made extra safe here with a ! and an ||.

liqunl: The transformation is only done when the constraint is a FixedClass constraint or a ResolvedClass constraint, which ensures constraint->getClass() returns non-NULL.

vijaysun-omr: Okay thanks for the explanation, I will merge now.

Issue #563, opened 2016-12-02
internalPointer flag on a node is incorrectly set


liqunl:
Some places set internalPointer flag on a node base on the result of isArrayRef on its opcode. The query isArrayRef returns true when the opcode is aiadd or aladd and is not sufficient to decide whether the node points to the middle of an object, causing gc failures when the gc tries to scan the object the node points to.

liqunl:
This has been fixed.

PR #564, opened 2016-12-02
WIP: Add canBeInternalPtrOfObject query

liqunl:
The query isArrayRef by its name is supposed to return true for nodes
who are addresses pointing to the middle of objects. However, it is a
query on an opcode and returns true for aiadd and aladd, which is
insufficient to serve its original purpose. In some places in the code,
the isInternalPtr flag on a node will be incorrectly set if isArrayRef
returns true, resulting crashes in the gc when it tries to scan the
object the node points to. Unfortunately, isArrayRef is used everywhere
to check if the opcode is aiadd or aladd or to serve other purposes,
thus we can not simply change the implementation of this query or
replace it everywhere with a different query. This commit adds a new
query canBeInternalPtrOfObject on the node and ask the new query
whereever isInternalPtr flag might be set when isArrayRef returns true.

issue: #563

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -2382,7 +2382,9 @@ OMR::Node::canBeInternalPtrOfObject()
       // Come across a node that has already been checked
       TR::ILOpCode op = curNode->getOpCode();
       if (printOp)
+         {
          fprintf(stderr,"->%s", op.getName());
+         }
 
       if (visited.contains(curNode))
          break;
@@ -2434,12 +2436,12 @@ OMR::Node::canBeInternalPtrOfObject()
                    ops == TR::variableNewArray ||
                    ops == TR::multianewarray ||
                    ops == TR::aconst ||
-                   ops == TR::i2a ||
-                   ops == TR::l2a ||
+                   //ops == TR::i2a ||
+                   //ops == TR::l2a ||
                    ops == TR::aiadd ||
                    ops == TR::aladd))
             {
-            fprintf(stderr, "Found unsupported opcode: %s",childOp.getName());
+            fprintf(stderr, "Found unsupported opcode: ");
             printOp = true;
             }
          }
@@ -2449,8 +2451,13 @@ OMR::Node::canBeInternalPtrOfObject()
 
    newResult = newResult && !printOp;
    if (printOp)
+      {
       fprintf(stderr, "Node %p %s internal pointer and we return true\n", self(), newResult ? "is" : "is not");
-   TR_ASSERT(newResult, "Return true when we should really return false, node is "POINTER_PRINTF_FORMAT, self());
+      fflush(stderr);
+      //comp->setOutFile(TR::FilePointer::Stdout());
+      //comp->dumpMethodTrees("About to assert, tree is");
+      }
+   TR_ASSERT(!printOp, "Return true when we should really return false, node is %p", self());
    return true;
    }
 
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -2336,6 +2336,7 @@ OMR::Node::dontEliminateStores(bool isForLocalDeadStore)
 
 
 
+// false is not equivalent to isCollected
 bool
 OMR::Node::isNotCollected()
    {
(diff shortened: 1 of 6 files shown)


liqunl:
@andrewcraik Andrew, this is the fix for incorrect internal pointer, can you review it when you have time? The change has been tested with acceptance build and toronto personal build.

compiler/il/OMRNode.cpp
```diff
@@ -2352,6 +2353,47 @@ OMR::Node::isNotCollected()
 
 
 bool
+OMR::Node::canBeInternalPtrAtObject()
```
andrewcraik: can this please be canBeInternalPtrOfObject for grammatical reasons. Can you please add a document comment describing what the function does as well?

compiler/il/OMRNode.cpp
```diff
@@ -2352,6 +2353,47 @@ OMR::Node::isNotCollected()
...
+         TR::Symbol *symbol = curNode->getSymbolReference()->getSymbol();
+         TR::ILOpCode op = curNode->getOpCode();
+         return op.isNew() ||
+                (op.getDataType() == TR::Address && op.isCall()) ||
+                symbol->isInternalPointer() ||
+                symbol->isCollectedReference();
+         }
+
+      visited.add(curNode);
+      if (curNode->getNumChildren() >= 1)
```
andrewcraik: I think this assumption is meant to handle something like a*sub and a*add. There are things like aternary that I am not sure will be handled correctly so we should put an assert to make sure only the cases we expect to handle will be handled.

liqunl: Maybe we should support aternary?

liqunl: Sorry, I just realized that we can't support aternary since the result is not known statically. Will add the assert.

liqunl: @andrewcraik Changes have been made according to your comment.

compiler/il/OMRNode.cpp
```diff
@@ -2351,6 +2352,56 @@ OMR::Node::isNotCollected()
...
+         {
+         curNode = curNode->getFirstChild();
+         TR_ASSERT(curNode->getOpCode() != TR::aternary, "canBeInternalPtrOfObject does not work for nodes who originate from a node with data type TR::aternary, node is " POINTER_PRINTF_FORMAT, self());
```
andrewcraik: I'd prefer this to be a whitelist rather than a black list - if someone adds a new opcode and it isn't accounted for here we should assert I believe. Thoughts?

liqunl: Yeah, we can have a whitelist, I'll get you the list in a second.

liqunl: @andrewcraik Whitelist has been added, I did not create a new function because the whitelist is not general for other purposes.

compiler/il/OMRNode.cpp
```diff
@@ -2351,6 +2352,82 @@ OMR::Node::isNotCollected()
...
+                   op == TR::iu2a ||
+                   op == TR::l2a ||
+                   op == TR::lu2a ||
+                   op == TR::b2a ||
+                   op == TR::bu2a ||
+                   op == TR::s2a ||
+                   op == TR::su2a ||
+                   op == TR::loadaddr ||
+                   op == TR::aRegLoad ||
+                   op == TR::compressedRefs ||
```
andrewcraik: Can this really appear as a first child - I was sure it was a treetop?

compiler/il/OMRNode.cpp
```diff
@@ -2351,6 +2352,82 @@ OMR::Node::isNotCollected()
...
+                   op == TR::New ||
+                   op == TR::newarray ||
+                   op == TR::anewarray ||
+                   op == TR::variableNew ||
+                   op == TR::variableNewArray ||
+                   op == TR::multianewarray ||
+                   op == TR::acalli ||
+                   op == TR::aiadd ||
+                   op == TR::aiuadd ||
```
andrewcraik: This opcode is deprecated please do not support it.

compiler/il/OMRNode.cpp
```diff
@@ -2351,6 +2352,82 @@ OMR::Node::isNotCollected()
...
+                   op == TR::aladd ||
+                   op == TR::aluadd ||,
```
andrewcraik: This opcode is deprecated please do not support it.

compiler/il/OMRNode.cpp
```diff
@@ -2351,6 +2352,82 @@ OMR::Node::isNotCollected()
...
+
+      visited.add(curNode);
+      if (curNode->getNumChildren() >= 1)
+         {
+         curNode = curNode->getFirstChild();
+         TR_ASSERT(op == TR::aload ||
+                   op == TR::aloadi ||
+                   op == TR::areturn ||
+                   op == TR::acall ||
+                   op == TR::i2a ||
```
andrewcraik: Can't this list of conversions be simplified by asking if the opcode is a conversion since we know conversions to address type will follow your rule?

compiler/il/OMRNode.cpp
liqunl: When the environment variable TR_UseTranslateInTrees is set, it'll appear as a first child. Not sure if we still use that code, but it's there.

compiler/il/OMRNode.cpp
liqunl: I took a deeper thought into it and  I think we can't support conversions, because if we have something like the following, we can't check which lload is the address.
```
l2a
   ladd
     lload
     lload
```

andrewcraik: materializing an arbitrary pointer will always be tricky - we need to consider what the current code would do and will we do anything different using your fix in this case? I would imagine we would assume the pointer was not collected and carry on, but I am not sure.

andrewcraik: A default assumption that anything coming from a conversion to address type is not a collected reference is probably a good starting point and if there are issues with that we can go from there.

andrewcraik:
@liqunl please fix the lint failure in the CI build so this can be re-reviewed for a merge.

liqunl:
The assertion has triggered in a bunch of jobs, looks like we have quite a lot of places where aiadd/aladd's children using incorrect opcode. I'm trying to fix them and it might take some time.

mstoodle:
May I ask after current status here, please? Are you finding value in having a WIP merge request open here, or could we close it until such time as you're ready to complete it?

Not trying to push you away, just not sure a lingering inactive PR is providing much value.

liqunl:
The fix has been paused due to other work. Closing this PR and will open a new one once the fix is ready.

PR #565, opened 2016-12-02
Load Power Table Of Constants in method prologue

R-Santhir:
The Power Table Of Contents(TOC) was not being loaded into r2 as specified by the Power System
Linkage. When the caller does not do this, the callee will try and use
the TOC base register, r2, resulting in an error. To fix this the TOC
will be loaded when the prologue is created.

### DIFF
--- a/compiler/p/codegen/PPCSystemLinkage.cpp
+++ b/compiler/p/codegen/PPCSystemLinkage.cpp
@@ -641,6 +641,7 @@ TR_PPCSystemLinkage::createPrologue(
    cg()->setStackPointerRegister(machine->getPPCRealRegister(properties.getNormalStackPointerRegister()));
    TR::RealRegister        *sp = cg()->getStackPointerRegister();
    TR::RealRegister        *metaBase = cg()->getMethodMetaDataRegister();
+   TR::RealRegister        *gr2 = machine->getPPCRealRegister(TR::RealRegister::gr2);
    TR::RealRegister        *gr0 = machine->getPPCRealRegister(TR::RealRegister::gr0);
    TR::RealRegister        *gr11 = machine->getPPCRealRegister(TR::RealRegister::gr11);
    TR::RealRegister        *cr0 = machine->getPPCRealRegister(TR::RealRegister::cr0);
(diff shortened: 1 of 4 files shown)


0xdaryl:
I'll review this thoroughly when the PR is no longer a WIP.  In the meantime, however, please change your commit message to refer to `Power` rather than `PPC`.

It may be worth expanding "TOC" in its first use for clarity.  i.e., TOC (table of contents)

R-Santhir:
@ymanton

R-Santhir:
@ymanton, let me know if I have missed anything

R-Santhir:
@0xdaryl, I've update the commit with your suggestions

compiler/p/codegen/PPCSystemLinkage.cpp
```diff
@@ -641,13 +641,17 @@ TR_PPCSystemLinkage::createPrologue(
...
    TR::RealRegister        *gr0 = machine->getPPCRealRegister(TR::RealRegister::gr0);
    TR::RealRegister        *gr11 = machine->getPPCRealRegister(TR::RealRegister::gr11);
    TR::RealRegister        *cr0 = machine->getPPCRealRegister(TR::RealRegister::cr0);
    TR::Node                   *firstNode = comp()->getStartTree()->getNode();
    TR::RealRegister::RegNum regIndex;
    int32_t                    size = bodySymbol->getLocalMappingCursor();
    int32_t                    argSize;
+    
+   //Move TOCBase into r2
+   cursor = loadConstant(cg(), firstNode, (int64_t)(cg()->getTOCBase()), gr2, cursor, false, false);
```
ymanton: You will need to guard this with `#ifdef __LITTLE_ENDIAN__` since this only applies to `ppc64le`.

R-Santhir: I've made the appropriate change

R-Santhir:
Personal Build:
http://zaxxon.torolab.ibm.com/build_info.php?build_id=128785

0xdaryl:
Please change your commit title to something like "Load Power Table Of Constants in method prologue" as per the project Contribution Guidelines.

Looks OK otherwise.

PR #567, opened 2016-12-05
Simplify packed decimal restore snippet generation

NigelYiboYu:
Simplify generateS390 instruction functions for packed decimal operations
so that they use the same function to generate restore snippets.

### DIFF
--- a/compiler/z/codegen/S390GenerateInstructions.cpp
+++ b/compiler/z/codegen/S390GenerateInstructions.cpp
@@ -780,26 +780,7 @@ generateRXInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op, TR::N
 #ifdef J9_PROJECT_SPECIFIC
    if (op == TR::InstOpCode::CVB || op == TR::InstOpCode::EX)
       {
-      TR::Node* BCDCHKNode = cg->getCurrentCheckNodeBeingEvaluated();
-
-      if (BCDCHKNode && BCDCHKNode->getOpCodeValue() == TR::BCDCHK)
-         {
-         // Explicitly mark as a DAA intrinsic instruction
-         instr->setThrowsImplicitException();
-
-         //there might be cases that two or more insts generated will throw exception, and we only need one of the label.
-         //so we will assume that at this moment the second child of BCDCHKNode is the label.
-         TR::LabelSymbol * label = (TR::LabelSymbol *)BCDCHKNode->getChild(1);
-
-         TR::S390RestoreGPR7Snippet * restoreSnippet =
-                        new (INSN_HEAP) TR::S390RestoreGPR7Snippet(cg, n,  TR::LabelSymbol::create(INSN_HEAP,cg), label);
-         cg->addSnippet(restoreSnippet);
-         TR::Instruction * nop = new (INSN_HEAP) TR::S390NOPInstruction(TR::InstOpCode::NOP, 2, n, cg);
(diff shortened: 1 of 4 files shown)


compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -3380,3 +3240,67 @@ TR::Instruction *generateZeroVector(TR::Node *node, TR::CodeGenerator *cg, TR::R
    {
    return generateVRIaInstruction(cg, TR::InstOpCode::VREPI, node, vecZeroReg, 0, 3);
    }
+
+#ifdef J9_PROJECT_SPECIFIC
+
+/**
+ * \brief This function returns true if the spcified instruction is designed to throw decimal overflow
```
fjeremic: These types of statement should go under `\return` clause. Perhaps something like this:

```cpp
/**
 * \brief 
 *    Determines if an instruction can throw decimal overflow.
 * 
 * \param op 
 *    The instruction to check.
 * 
 * \returns
 *    true if \p op is designed to throw decimal overflow exception according to the PoP; false otherwise.
 *
 * \details 
 *    Linux on z runs a process with this exception disabled. As a result of this, the JIT signal handler
 *    does not receive a signal in case decimal overflow exception happens. This helper function decides 
 *    whether to rely on condition code in such cases.
*/
```

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -3380,3 +3240,67 @@ TR::Instruction *generateZeroVector(TR::Node *node, TR::CodeGenerator *cg, TR::R
...
+       return false;
+      }
+   }
+
+void
+generateS390DAAExceptionRestoreSnippet(TR::CodeGenerator* cg,
+                                       TR::Node* n,
+                                       TR::Instruction* instr,
+                                       TR::InstOpCode::Mnemonic op,
+                                       bool isHasNOP)
```
fjeremic: Similarly here, `is` and `Has` is confusing. Perhaps just `hasNOP`, or even better `generateNOP`. In either case this function needs to be documented as it is not immediately clear what it does.

NigelYiboYu: Changed to `hasNOP`

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -3380,3 +3240,74 @@ TR::Instruction *generateZeroVector(TR::Node *node, TR::CodeGenerator *cg, TR::R
...
+/**
+ * \brief
+ *    Determines if an instruction can throw a decimal overflow exceptions.
+ *
+ * \param op
+ *    The instruction to check.
+ *
+ * \return
+ *    true if \p op is designed to throw decimal overflow exception according to the PoP; false otherwise.
+ *
```
fjeremic: The name is slightly confusing with both `is` and `Can` in the name. Would `canThrowDecimalOverflowException` be better?

NigelYiboYu: OK, fixed. 
I realized that it's a bit confusing to name it like such. Given that the TR and OMR naming convention doesn't specify method naming, I'm therefore following google's naming convention, which is something that I'm familiar with: method names should start with verbs; and for methods that return booleans, use 'is' in front of them.

vijaysun-omr:
Looks like good cleanup with a lot of code mapped to the same routine. Nice job Nigel. Merging now.

PR #568, opened 2016-12-05
Jitbuilder vm

mstoodle:
Add new support for modelling virtual machine state variables (including an operand stack) to the JitBuilder API. Provides the basis for a poor-person's On Stack "Replacement".

Add tests for the operand stack modelling.

Some minor cleanup also included.

Issue: #570

### DIFF
--- a/compiler/ilgen/BytecodeBuilder.cpp
+++ b/compiler/ilgen/BytecodeBuilder.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2016
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -215,40 +215,112 @@ OMR::BytecodeBuilder::transferVMState(TR::BytecodeBuilder **b)
    }
 
 void
+OMR::BytecodeBuilder::Goto(TR::BytecodeBuilder **dest)
+   {
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   Goto(*dest);
+   }
+
+void
 OMR::BytecodeBuilder::Goto(TR::BytecodeBuilder *dest)
    {
    AddSuccessorBuilder(&dest);
-   OMR::IlBuilder::Goto((TR::IlBuilder **)&dest);
+   OMR::IlBuilder::Goto(dest);
    }
 
 void
 OMR::BytecodeBuilder::IfCmpEqual(TR::BytecodeBuilder **dest, TR::IlValue *v1, TR::IlValue *v2)
    {
-   bool added = false;
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpEqual(*dest, v1, v2);
+   }
 
-   if (*dest != NULL)
-      {
-      AddSuccessorBuilder(dest);
-      added = true;
-      }
-   OMR::IlBuilder::IfCmpEqual((TR::IlBuilder **)dest, v1, v2);
+void
+OMR::BytecodeBuilder::IfCmpEqual(TR::BytecodeBuilder *dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpEqual(dest, v1, v2);
+   }
 
-   if (!added)
-      AddSuccessorBuilder(dest); // if dest didn't exist previously, it can't be a merge
+void
+OMR::BytecodeBuilder::IfCmpEqualZero(TR::BytecodeBuilder **dest, TR::IlValue *c)
+   {
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpEqualZero(*dest, c);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpEqualZero(TR::BytecodeBuilder *dest, TR::IlValue *c)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpEqualZero(dest, c);
    }
 
 void
 OMR::BytecodeBuilder::IfCmpNotEqual(TR::BytecodeBuilder **dest, TR::IlValue *v1, TR::IlValue *v2)
    {
-   bool added = false;
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpNotEqual(*dest, v1, v2);
+   }
 
-   if (*dest != NULL)
-      {
-      AddSuccessorBuilder(dest);
-      added = true;
-      }
-   OMR::IlBuilder::IfCmpNotEqual((TR::IlBuilder **)dest, v1, v2);
+void
+OMR::BytecodeBuilder::IfCmpNotEqual(TR::BytecodeBuilder *dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpNotEqual(dest, v1, v2);
+   }
 
-   if (!added)
-      AddSuccessorBuilder(dest); // if dest didn't exist previously, it can't be a merge
+void
+OMR::BytecodeBuilder::IfCmpNotEqualZero(TR::BytecodeBuilder **dest, TR::IlValue *c)
+   {
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpNotEqualZero(*dest, c);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpNotEqualZero(TR::BytecodeBuilder *dest, TR::IlValue *c)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpNotEqualZero(dest, c);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpLessThan(TR::BytecodeBuilder **dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpLessThan(*dest, v1, v2);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpLessThan(TR::BytecodeBuilder *dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpLessThan(dest, v1, v2);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpGreaterThan(TR::BytecodeBuilder **dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   if (*dest == NULL)
+      *dest = _methodBuilder->OrphanBytecodeBuilder(_bcIndex, _name);
+   IfCmpGreaterThan(*dest, v1, v2);
+   }
+
+void
+OMR::BytecodeBuilder::IfCmpGreaterThan(TR::BytecodeBuilder *dest, TR::IlValue *v1, TR::IlValue *v2)
+   {
+   TR_ASSERT(dest != NULL, "service cannot be called with NULL destination builder");
+   AddSuccessorBuilder(&dest);
+   OMR::IlBuilder::IfCmpGreaterThan(dest, v1, v2);
    }
--- a/compiler/ilgen/BytecodeBuilder.cpp
+++ b/compiler/ilgen/BytecodeBuilder.cpp
@@ -21,6 +21,7 @@
 #include "env/FrontEnd.hpp"
 #include "infra/List.hpp"
 #include "il/Block.hpp"
+#include "ilgen/VirtualMachineState.hpp"
 #include "ilgen/IlBuilder.hpp"
 #include "ilgen/MethodBuilder.hpp"
 #include "ilgen/BytecodeBuilder.hpp"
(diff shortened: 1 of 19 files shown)


mstoodle:
botched some stuff. working on it...

jitbuilder/release/src/OperandStackTests.hpp
```diff
@@ -0,0 +1,62 @@
...
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+
+#ifndef OPERANDSTACKTESTS_INCL
+#define OPERANDSTACKTESTS_IMPL
```
Leonardo2718: Should be

```c++
#define OPERANDSTACKTESTS_INCL
```

Leonardo2718:
Awesome work on the documentation btw! :+1:

compiler/ilgen/BytecodeBuilder.cpp
```diff
@@ -37,22 +38,29 @@ OMR::BytecodeBuilder::BytecodeBuilder(TR::MethodBuilder *methodBuilder,
...
-    _methodSymbol = methodSymbol;
-    _fe = fe;
-    _symRefTab = symRefTab;
-    _comp = TR::comp();
+    this->OMR::IlInjector::initialize(details, methodSymbol, fe, symRefTab);
+    //_details = details;
+    //_methodSymbol = methodSymbol;
+    //_fe = fe;
+    //_symRefTab = symRefTab;
+    //_comp = TR::comp();
```
mstoodle: yes, thanks

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -0,0 +1,197 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2000, 2016
```
mstoodle: nope. fixed.

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -0,0 +1,197 @@
...
+   _pushAmount(other->_pushAmount)
+   {
+   int32_t numBytes = _stackMax * sizeof(TR::IlValue *);
+   _stack = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
+   memcpy(_stack, other->_stack, numBytes);
+   }
+
+
+// commits the simulated operand stack of values to the virtual machine state
+// the given builder object is the path to the interpreter
```
mstoodle: i'll remove that as it's been described in other places as the "typical" use of the Commit() function, and I agree this version sounds a bit confusing.

compiler/ilgen/VirtualMachineOperandStack.hpp
```diff
@@ -0,0 +1,190 @@
...
+ *   value onto the virtual machine's operand stack (more details at definition).
+ * Reload() is left empty; assumption is that each BytecodeBuilder handler will
+ *   update the state of the operand stack appropriately on return from the
+ *   interpreter.
+ * MakeCopy() copies the state of the operand stack
+ * MergeInto() is slightly subtle. Operations may have been already created
+ *   below the merge point, and those operations will have assumed the
+ *   expressions are stored in the TR::IlValue's for the state being Merged
+ *   *to*. So the purpose of Merge() is to store the values of the current
+ *   state into the same variables as in the "other" state.
```
mstoodle: damn, I thought I'd caught all of those. thanks!

compiler/ilgen/VirtualMachineOperandStack.hpp
```diff
@@ -0,0 +1,190 @@
...
+ *   Pop() pops and returns a TR::IlValue from the stack
+ *   Top() returns the TR::IlValue at the top of the stack
+ *   Pick() returns the TR::IlValue "depth" elements from the top
+ *   Drop() discards "depth" elements from the stack
+ *   Dup() is a convenience function for Push(Top())
+ *
+ * To facilitate propagating an operand stack as control flows from one bytecode
+ * to another, VirtualMachineOperandStack provides a function MakeCopy() which
+ * allocates a new VirtualMachineOperandStack and makes it an identical
+ * copy of the current operand stack. It is typically used to
```
mstoodle: actually, that whole paragraph is an earlier (obviously incomplete) version of the comment. i've removed it.

compiler/ilgen/VirtualMachineState.hpp
```diff
@@ -0,0 +1,123 @@
...
+ * ##Usage
+ *
+ * Commit() is typically called when transitioning from compiled code to the interpreter.
+ *
+ * Reload() is typically called on the transition back from the interpreter to compiled
+ * code.
+ *
+ * MakeCopy() is typically called when control flow edges are created: the current vm state
+ * must also become the initial vm state at the target of the control flow edge and be
+ * able to evolve independently from the current vm state.
```
mstoodle: Yes. Otherwise, any change to the state in the "source" builder object would also change the state in the "target" builder object. On top of that, every builder object needs to maintain the "initial" vm state (i.e. the state when control first gets there) as well as the "current" vm state (i.e. the state as it evolves due to the operations performed by that builder). You need the initial state so that any MergeInto() operations to that builder object synchronize to the proper vm state...

mstoodle:
I believe I have addressed your comments @Leonardo2718 . Thanks for the review !

compiler/ilgen/VirtualMachineState.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+   /**
+    * @brief create an identical copy of the current object.
+    * @returns the copy of the current object
+    *
+    * Typically used when propagating the current state along a flow edge to another builder to
+    * capture the input state for that other builder.
+    * Default implementation simply returns the current object.
+    */
+   virtual VirtualMachineState *MakeCopy() { return this; }
```
mstoodle: Maybe, but since VirtualMachineState is just an interface, it's unlikely this would be the right answer for any subclass and would just have to be reimplemented anyway. At least "this" is guaranteed to be the expected kind of object for any subclass that inherits this implementation, and VirtualMachineRegister and VirtualMachineRegisterInStruct (currently) can reuse it.

Leonardo2718: I see. Thanks for clarifying!

compiler/ilgen/VirtualMachineState.hpp
```diff
@@ -0,0 +1,123 @@
...
+   public:
+
+   /**
+    * @brief Cause all simulated aspects of the virtual machine state to become real.
+    * @param b builder object where the operations will be added to change the virtual machine state.
+    *
+    * The builder object b is assumed to be along a control flow path transitioning
+    * from compiled code to the interpreter. Base implementation does nothing.
+    */
+   virtual void Commit(TR::IlBuilder *b) { }
```
Leonardo2718: Hm... I see your point about `MergeInto`. However, I can't think of a case where `Commit` and `Reload` don't need to be re-implemented. If there are such cases, I would think they are rather rare. So, to help enforce the idea that this is really an interface, I would prefer if these were pure virtual.

That said, since this is more of a conceptual issue than a functional issue, I'm okay with leaving it as is.

Leonardo2718:
LGTM :+1:

charliegracie:
I am really looking forward to being able to consume this in my SOM++ project :)

compiler/ilgen/VirtualMachineRegister.hpp
```diff
@@ -0,0 +1,179 @@
...
+      }
+
+   /**
+    * @brief used in the compiled method to add to the (simulated) register's value
+    * @param b the builder where the operation will be placed to add to the simulated value
+    * @param amount the constant value that should be added to the simulated value, after multiplication by _adjustByStep
+    * Adjust() is really a convenience function for the common operation of adding a value to the register. More
+    * complicated operations (e.g. multiplying the value) can be built using Load() and Store() if needed.
+    */
+   virtual void Adjust(TR::IlBuilder *b, int64_t amount)
```
vijaysun-omr: Maybe this can call the Adjust routine taking an arg of type TR::IlValue * to keep the (small) logic in one place. 

Wondered about the name "Adjust" which is'nt the most descriptive, but it is an aesthetic point.

compiler/ilgen/VirtualMachineRegisterInStruct.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * The simulated register value is simply stored in a single local variable, which
+ * gives the compiler visibility to all changes to the register (and enables
+ * optimization / simplification). Because there is just a single local variable,
+ * the Merge() function does not need to do anything (the value is accessible from
+ * the same location at all locations). The Commit() and Reload() functions simply
+ * move the value back and forth between the local variable and the structure that
+ * holds the actual virtual machine state.
+ */
+
+class VirtualMachineRegisterInStruct : public ::OMR::VirtualMachineRegister
```
vijaysun-omr: My understanding is that VirtualMachineRegisterInStruct represents a "field" inside some struct whereas VirtualMachineRegister just represents a field loaded from some "static" address. 

If so, is there any key reason why the former extends the latter ? I see that some helper routines like Load and Store are inherited as a convenience because of that hierarchy but could that effect not have been achieved if those routines were in another common superclass that was directly extended by both VirtualMachineRegister and VirtualMachineRegisterInStruct ?

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+void
+VirtualMachineOperandStack::checkSize()
+   {
+   if (_stackTop == _stackMax)
+      grow();
+   }
+
+void
+VirtualMachineOperandStack::grow()
```
vijaysun-omr: Did you consider a grow method that takes an "increment amount" as a parm as opposed to having a "standard" increment amount applied inside of grow() ?

compiler/ilgen/VirtualMachineOperandStack.hpp
```diff
@@ -0,0 +1,186 @@
...
+   protected:
+   void copyTo(OMR::VirtualMachineOperandStack *copy);
+   void checkSize();
+   void grow();
+
+   private:
+   TR::MethodBuilder *_mb;
+   OMR::VirtualMachineRegister *_stackTopRegister;
+   int32_t _stackMax;
+   int32_t _stackTop;
```
vijaysun-omr: A highly pedantic question (and please be gentle/genteel in your response to this one) 

Should you be accounting for the possibility that an operand stack for some esoteric language might be allowed to have more than 32 bits worth of elements ?

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -0,0 +1,197 @@
...
+   for (int32_t i=_stackTop;i >= 0;i--)
+      {
+      // only need to do something if the two entries aren't already the same
+      if (other->_stack[i]->getCPIndex() != _stack[i]->getCPIndex())
+         {
+         // what if types don't match? could use ConvertTo, but seems...arbitrary
+         // nobody *should* design bytecode set where corresponding elements of stacks from
+         // two incoming control flow edges can have different primitive types. objects, sure
+         // but not primitive types! Expecting to be disappointed here some day...
+         TR_ASSERT(_stack[i]->getSymbol()->getDataType() == other->_stack[i]->getSymbol()->getDataType(), "disappointing stack merge: primitive type mismatch at same depth stack elements");
```
vijaysun-omr: How does this code know it's dealing with primitive types as opposed to objects ? Maybe I am missing something subtle...

mstoodle:
Yes, after helping @vijaysun-omr through it, I could have done a much better job ordering and structuring the commits to help people understand the content. Sorry about that!

compiler/ilgen/VirtualMachineOperandStack.cpp
mstoodle: Maybe I should fix up that comment. The argument was that two different objects (i.e. two different pointers to objects) could come from different predecessor edges, but that the underlying type of the value in the operand stack should be the same. i.e. one edge shouldn't have pushed an Int32 while the other side pushed a Float.

The "disappointed" comment is just my funny way of commiserating with whoever tracks down this particular assertion firing in future.

compiler/ilgen/VirtualMachineOperandStack.hpp
mstoodle: 32 is just a hint. There are two functions checkSize() and grow() that will grow the stack if needed. Come to think of it, the current OperandStackTests.cpp test does not ensure that these two functions operate correctly <blush>. I'll open an issue for that once this is merged.

compiler/ilgen/VirtualMachineOperandStack.cpp
mstoodle: it's a good idea as an extension.

compiler/ilgen/VirtualMachineRegisterInStruct.hpp
mstoodle: 2 "reasons": 1) I estimated that the amount of code different between such a base class and the current VirtualMachineRegister class would be very small, and 2) I originally had visions of implementing VMRegInStruct using VMReg (i.e. just loading the address of the field and passing it to VirtualMachineRegister). I ultimately did not go the 2nd way, but arranging them this way it still could (and therebyshare even more implementation).

Do you think it would be importantly better that way?

compiler/ilgen/VirtualMachineOperandStack.cpp
vijaysun-omr: Sorry I meant the TR_ASSERT message can say "invalid" instead of "disappointing".

Regarding the other issue, I understand reading the code again that we won't have any mismatch for objects anyway since both stack entries would have the same type; the only time the assert would fire would be for 2 primitive types that are not the same or a primitive and an object type, both of which I agree should be invalid.

compiler/ilgen/VirtualMachineRegisterInStruct.hpp
vijaysun-omr: Not "importantly" :) I was just curious because there did not seem to be an obvious inheritance relationship to me.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1561,54 +1568,95 @@ void
...
 IlBuilder::IfCmpGreaterThan(TR::IlBuilder **target, TR::IlValue *left, TR::IlValue *right)
    {
    *target = createBuilderIfNeeded(*target);
+   IfCmpGreaterThan(*target, left, right);
+   }
+
+void
+IlBuilder::IfCmpGreaterThan(TR::IlBuilder *target, TR::IlValue *left, TR::IlValue *right)
+   {
    ILB_REPLAY("%s->IfCmpGreaterThan(%s, %s, %s);", REPLAY_BUILDER(this), REPLAY_BUILDER(target), REPLAY_VALUE(left), REPLAY_VALUE(right));
```
vijaysun-omr: REPLAY_BUILDER used to take target as an argument when it was a TR::IlBuilder ** and it seems to still take target even though the type is now TR::IlBuilder * (?) i.e. we now have one extra level of indirection and I am not sure this is intentional.

vijaysun-omr:
I have now reviewed all of the commits. I can merge in once my last couple of review comments are addressed.

compiler/ilgen/VirtualMachineOperandStack.cpp
mstoodle: ah! I missed the one in the assert message. I can fix that, but perhaps I could do it as part of another PR since it's not a functional issue, just to avoid another testing go-around on this one?

vijaysun-omr:
I am merging in this PR now; the amendments are all minor ones that are not functional issues per se and I would like them to be delivered via a separate PR sometime soon. Thanks for the responses to the comments.

PR #572, opened 2016-12-06
Reorganize OMR simplifier code

xliang6:
Move J9 project specific code out of OMR simplifier code. Create
virtual methods in OMR simplifier class to allow different actions
to be taken for OMR and J9.

Remove OMR simplifier helper and handler functions and other code
snippets that effectively didn't do anything.

### DIFF
--- a/compiler/optimizer/OMRSimplifier.hpp
+++ b/compiler/optimizer/OMRSimplifier.hpp
@@ -30,6 +30,7 @@
 #include "il/Node.hpp"                        // for Node, etc
 #include "il/Node_inlines.hpp"                // for Node::getFirstChild, etc
 #include "infra/Assert.hpp"                   // for TR_ASSERT
+#include "infra/Checklist.hpp"                // for {Node,Block}Checklist
 #include "infra/HashTab.hpp"                  // for TR_HashTabInt
 #include "optimizer/Optimization.hpp"         // for Optimization
 #include "optimizer/OptimizationManager.hpp"  // for OptimizationManager
@@ -56,7 +57,7 @@ typedef enum
    ConditionCodeLast = ConditionCodeInvalid
    } TR_ConditionCodeNumber;
 
-/*
+/**
  * Class Simplifier
  * ================
  *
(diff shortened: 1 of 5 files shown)


xliang6:
JIT libraries for x86-64 with and without my changes were given to Andrew's team. It has been verified that there is no noticeable performance difference between the two libraries.

xliang6:
Marius also verified that the two builds look the same from startup/footprint/CPU point of view.

compiler/optimizer/OMRSimplifier.hpp
```diff
@@ -121,7 +122,127 @@ class Simplifier : public TR::Optimization
...
+
+   /**
+    * Checks if the bound value is great than or equal to the length value.
+    *
+    * @param[in] boundChild The tree node corresponding to the bound value
+    * @param[in] lengthChild The tree node corresponding to the length value
+    * @return Returns true if bound is great than or equal to length; 
+    * false otherwise
+    */
+   virtual bool isLengthGEBound(TR::Node *boundChild, TR::Node *lengthChild);
```
mstoodle: This function name and description is a bit misleading, to me. It's actually asking if the length is known to be >= the bound, isn't it? Even if it returns false, the length may be >= bound. But the way it's named, one could be easily conclude that false means length <  bound. Perhaps name it isLengthDefinitelyGEBound() or something like that?

The actual implementation has nothing to do with lengths and bounds, either. It's just testing whether its first operand is definitely less than the second operand, right? Does it needs to be "tied" to the concept of array bounds checking?

compiler/optimizer/OMRSimplifier.hpp
```diff
@@ -121,7 +122,34 @@ class Simplifier : public TR::Optimization
    void prepareToReplaceNode(TR::Node * node);
 
    void anchorOrderDependentNodesInSubtree(TR::Node *node, TR::Node *replacement, TR::TreeTop* anchorTree);
-   TR::Node * unaryCancelOutWithChild(TR::Node * node, TR::Node * firstChild, TR::TreeTop* anchorTree, TR::ILOpCodes opcode, bool anchorChildren=true);
+
```
xliang6: Comments for all new methods have been added.

compiler/optimizer/OMRSimplifier.hpp
xliang6: Good point, the name has been changed to isBoundDefinitelyGELength. I agree with you that the comparison currently has nothing to do with lengths and bounds, but a method that compares its two arguments might add other IL tree patterns in the future that might not apply here so I would like to leave the name with its "tie" to array bound checking.

xliang6:
@fjeremic There is some packed code being refactored in the simplifier class. Can you review it and make sure it is okay? The changes have something to do with checking whether a node is of packed type and then calling setPDMulPrecision on the transformed node. Thanks.

fjeremic:
@xliang6 Just confirming that these PD related simplifications are not being removed, but being pushed up into J9?

xliang6:
@fjeremic Yes, they are pushed into J9.

fjeremic:
LGTM.

xliang6:
Thanks, @fjeremic 

@andrewcraik @vijaysun-omr Can you review my changes? Thanks.

compiler/optimizer/OMRSimplifierHelpers.cpp
```diff
@@ -602,60 +602,47 @@ TR::Node *reduceExpTwoAndGreaterToMultiplication(int32_t exponentValue, TR::Node
...
-      TR::Node * node = NULL;
-      TR::Node ** subTrees = (TR::Node**)s->comp()->trMemory()->allocateStackMemory((maxCeilingExp+1)*sizeof(TR::Node*));
+      TR::Node **subTrees = (TR::Node**)s->comp()->trMemory()->allocateStackMemory((maxCeilingExp+1)*sizeof(TR::Node*));
       subTrees[0] = baseNode;
       int32_t i = 0;
       for (i = 1; exponentValue >= (CONSTANT64(1) << i); ++i) // i can reach maxCeilingExp+1
          {
          int32_t j = i-1;
-         node = subTrees[i] = TR::Node::create(multOp, 2, subTrees[j], subTrees[j]);
-#ifdef J9_PROJECT_SPECIFIC
```
vijaysun-omr: How do you expect this kind of precision setting to be done if the virtual routine you created for this purpose (I believe setNodePrecisionIfNeeded ?) is not called from this spot in the simplifier ?

xliang6: Yes, the recursive method setNodePrecisionIfNeeded for the node is called at the end of the method before returning the node.

vijaysun-omr:
Looks okay to me. I can merge once @andrewcraik also gives his approval.

andrewcraik:
LGTM

Issue #573, opened 2016-12-06
Ability to call arbitrary C functions with JitBuilder


mstoodle:
While you can use DefineFunction() along with Call or ComputedCall to call an arbitrary function from any IlBuilder object, sometimes all you want to do is to call a function using a set of arguments (think libffi). In this scenario, what you really want to do is to create a thunk that's specific to the signature of parameter types and return type and be able to use that thunk to call functions that have that signature by passing an array of arguments.

The model I suggest is to create a thing called a ThunkBuilder, which will be a subclass of MethodBuilder, to create these thunks. Different ThunkBuilder instances will correspond to different native method signatures. Compiling a ThunkBuilder instance will generate a thunk for calling any function that has that signature. Since DefineFunction and ComputedCall already exist, ThunkBuilder should require only a very small amount of new code.

mstoodle:
ThunkBuilder support was merged via #577 .

PR #577, opened 2016-12-06
Add ThunkBuilder for libffi-like call building

mstoodle:
Provide a new kind of builder object that can call C functions from
an array of argument values. A ThunkBuilder object takes a name, a
return type, a number of parameters, and an array of parameter types.
The compiled code generated by a ThunkBuilder takes two arguments and
returns whatever kind of value was specified by the constructor call.
The first argument is the target call address, and the second argument
is an array of Word sized arguments, which will be passed to the
target method as the types indicated by the ThunkBuilder constructor
call.

Also created Thunk.cpp test that, initially, has only single character
and single short parameters with no return type. To be extended later.

Issue: #573 

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -930,11 +930,20 @@ TR::IlValue *
 IlBuilder::ConvertTo(TR::IlType *t, TR::IlValue *v)
    {
    appendBlock();
-   TR::IlValue *convertedValue = NewValue(t);
    TR::DataType t1 = v->getSymbol()->getDataType();
    TR::DataType t2 = t->getPrimitiveType();
+   if (t1 == t2)
+      {
+      TraceIL("IlBuilder[ %p ]::%d is ConvertTo (already has type %s) %d\n", this, v->getCPIndex(), t->getName(), v->getCPIndex());
+      return v;
+      }
+
    TR::ILOpCodes convertOp = TR::DataType::getDataTypeConversion(v->getSymbol()->getDataType(), t->getPrimitiveType());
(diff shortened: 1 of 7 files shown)


mstoodle:
Just getting some early testing; will rebase and update after #568 is merged.

mstoodle:
@DanHeidinga you probably haven't had much chance to try ThunkBuilder from my fork, but any feedback would be welcome

compiler/ilgen/IlBuilder.cpp
```diff
@@ -919,12 +919,22 @@ TR::IlValue *
...
    TR::DataType t1 = v->getSymbol()->getDataType();
    TR::DataType t2 = t->getPrimitiveType();
+   if (t1 == t2)
+      {
+      TraceIL("IlBuilder[ %p ]::%d is ConvertTo %d (already has type %s)\n", this, v->getCPIndex(), v->getCPIndex(), t->getName());
+      return v;
+      }
+
    TR::ILOpCodes convertOp = TR::DataType::getDataTypeConversion(v->getSymbol()->getDataType(), t->getPrimitiveType());
+   TR_ASSERT(convertOp != TR::BadILOp, "unknown type conversion required");
```
Leonardo2718: In the assert message, would it make more sense to say "unknown type conversion __requested__" ? Also, it would be helpful to have some debug information right in the message, eg: source type, target type, `IlBuilder` object.

mstoodle: agreed, i'll update it as part of rebasing

compiler/ilgen/ThunkBuilder.hpp
```diff
@@ -0,0 +1,99 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2016
```
0xdaryl: Fix copyright:  2016, 2016

PR #578, opened 2016-12-06
Set the 'swappedChildren' node flag in swapChildren()

mpirvu:
swapChildren() is an utility method used for swapping the children of a node.
When that happens for an 'if' node we also need to set the 'swappedChildren' flag.


mpirvu:
@vijaysun-omr could you please review this change set? Thanks

PR #581, opened 2016-12-06
Replace binary search in OMROptions with STL

amankumar1:
Replace the binary search in `processOption` that finds options
in a table with the STL implementation.

Signed-off-by: Aman Kumar <amank@ca.ibm.com>

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -3403,6 +3403,29 @@ OMR::Options::processOptionSet(
    return options;
    }
 
+/**
+ * Custom comparison function to compare two options.
+ * Used for STL binary search.
+ *
+ * @param a The first option.
+ * @param b The second option.
+ */
+
+bool
+OMR::Options::compareOptionsForBinarySearch(const TR::OptionTable &a, const TR::OptionTable &b)
+   {
+   int32_t lengthOfTableEntry;
+
+   // Check which option is the one passed in (optionToFind). We want to match
+   // only up to the number of characters of the option in the table, not the one
+   // passed in.
+   if (a.msg == "optionToFind")
+      lengthOfTableEntry = b.length;
+   else
+      lengthOfTableEntry = a.length;
+
+   return (strnicmp_ignore_locale(a.name, b.name, lengthOfTableEntry) < 0);
+   }
 
 char *
 OMR::Options::processOption(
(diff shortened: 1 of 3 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -3403,6 +3403,47 @@ OMR::Options::processOptionSet(
...
+      optionInTable = &a;
+      }
+
+   lengthOfTableEntry = optionInTable->length;
+
+   // Ensure 'log@lf' is the same as 'log=lf' (Z language environment transform)
+   if (optionInTable->name[lengthOfTableEntry - 1] == '=' &&
+       optionToFind->length >= optionInTable->length &&
+       optionToFind->name[lengthOfTableEntry - 1] == '@' &&
+       strnicmp_ignore_locale(a.name, b.name, lengthOfTableEntry - 1) == 0)
```
mgaudet: @fjeremic Is this transform still required? (IIRC, it was only needed for JCL, and I don't know if we still support JCL)

fjeremic: Not quite sure. First time I'm seeing this. @joransiu do you might know if this code is still relevant?

Leonardo2718: I recall investigating this a while ago and finding that it is no longer needed.

joransiu: I'm not familiar with this piece of code.   If the JCL that @mguadet refers to is z/OS JCL, it is common for folks to launch the Java applications via JCL (either through BPXBatch or JZOS launcher).

amankumar1:
FYI: Travis CI build passed but it didn't update on github for some reason.

compiler/control/OMROptions.cpp
```diff
@@ -3420,95 +3443,42 @@ OMR::Options::processOption(
...
+   // Create an object for the option to find in the table and set attributes
+   TR::OptionTable optionToFind = TR::OptionTable();
+   optionToFind.name = startOption;
+   optionToFind.length = strlen(optionToFind.name);
+
+   // Since STL binary search requires total ordering, there need to be a way to differentiate
+   // between the option to find and an option in the table. The msg field is used here to
+   // specify which option is the option to find. Note that this is not normal usage of the
+   // msg field.
+   optionToFind.msg = "optionToFind";
```
0xdaryl: While a clever way to leverage what's already in the OptionTable struct without adding anything new, I would prefer adding a `bool` to indicate this special property to make it crystal clear what the intent is.  It also places an undocumented dependency on an actual option message.  It is highly unlikely that this would ever cause a problem but we're leaving the possibility open.

amankumar1:
@0xdaryl Made the requested change. Waiting for the tests to pass.

0xdaryl:
I think the changes look fine and are ready to be merged.  Thanks.

Before I do that, can you briefly outline how you tested this?  Options testing is arguably weak in OMR (TBD) yet Options are used quite extensively.

amankumar1:
@0xdaryl Sorry for the late response. I used a number of input strings (manually generated and from other projects) to make sure the STL binary search was doing the right thing. A close inspection of the new code shows that it is functionally same as the old one.

PR #582, opened 2016-12-06
Fix linting errors on P and Z

noryb009:
Travis CI doesn't run the linter on Power and Z specific code. This resolves the linting errors that slipped through.

These changes allow the compiler to correctly apply static polymorphism to the calls. Two of these changes do so by adding `self()->` to calls to member functions, while the rest add `TR::TreeEvaluator::` to calls to TreeEvaluator static functions.


mstoodle:
Can you make the commit message a bit more descriptive of the fixes that needed to be applied?

noryb009:
@mstoodle Updated.

mstoodle:
Thanks @noryb009

Issue #574, opened 2016-12-06
JitBuilder should have a work list for BytecodeBuilders


mstoodle:
JitBuilder clients should probably use a worklist to process BytecodeBuilders in control flow order rather than any arbitrary order. That's because the propagation of virtual machine states added in PR #568 relies on visiting byte code builders in order (i.e. in order to have a vm state for a byte code builder B you must have already processed another bytecodebuilder that had a vm state and that created a control flow edge to B).

It makes sense for MethodBuilder to provide a work list for BytecodeBuilders; adding to the worklist could/should even be hidden behind the services that create control edges between BytecodeBuilders. A fringe benefit would be that bytecode builders for bytecodes that aren't reachable would never be added to the worklist and would never be processed / translated.

mstoodle:
Worklist support was merged via #583 .

PR #583, opened 2016-12-07
JitBuilder worklist

mstoodle:
Add bytecode index worklist service to MethodBuilder
    
The automatic propagation of virtual machine state variables performed by BytecodeBuilders requires that the language compiler visit bytecodes in a particular order. In particular, no bytecode should be visited before at least one of its predecessors has been visited. This requirement stems from the fact that it is the creation of that flow edge from the predecessor to the bytecode that causes the vm state to propagate. If you visit a bytecode before any of its predecessors, it will almost certainly have a NULL vmState.
    
To facilitate the correct traversal of the bytecodes, I'm going to add a worklist facility to MethodBuilder, which will keep track of which bytecode indices have already been visited and which bytecode indices are ready to be visited. A pair of bit vectors serves the purpose nicely. Then, by augmenting the services in MethodBuilder and BytecodeBuilder that create control flow edges between BytecodeBuilder objects, the insertion of bytecode indices to the worklist can be done automatically.
    
The language compiler must create an initial VM state and explicitly store it to the MethodBuilder object. The first control flow edge from the MethodBuilder to the first bytecode (typically at bytecode index 0) must be created by calling AppendBuilder() on the MethodBuilder. From there, the language compiler need only iterate calling GetNextBytecodeFromWorklist() to get the next bytecode index to be processed until it returns -1. When -1 is returned, there are no remaining bytecode builders to visit. If there are any other special entry points, then the bytecode builders for those special entry points will need to be explicitly added to the worklist by passing them to addBytecodeBuilderToWorklist().
    
This commit also adds a test of the worklist, which verifies that the reachable bytecode builders and only those builders are traversed by the worklist. The method will fail to compile if the worklist does not visit exactly the right builders in exactly the expected (from lowest to highest bc index ) order.

Issue: #574

### DIFF
--- a/jitbuilder/release/src/Worklist.cpp
+++ b/jitbuilder/release/src/Worklist.cpp
@@ -1,0 +1,347 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2016, 2016
+ *
+ *  This program and the accompanying materials are made available
+ *  under the terms of the Eclipse Public License v1.0 and
+ *  Apache License v2.0 which accompanies this distribution.
+ *
+ *      The Eclipse Public License is available at
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+
+#include <iostream>
+#include <stdlib.h>
+#include <stdint.h>
+#include <dlfcn.h>
+#include <errno.h>
+
+#include "Jit.hpp"
+#include "ilgen/TypeDictionary.hpp"
+#include "ilgen/MethodBuilder.hpp"
+#include "ilgen/BytecodeBuilder.hpp"
+#include "ilgen/VirtualMachineState.hpp"
+#include "Worklist.hpp"
+
+using std::cout;
+using std::cerr;
+
+#define REPORT_BCI(bci,e_bci)   { if ((bci) != (e_bci)) { cerr << "Error: bytecode " << (bci) << " not found in worklist ( expected" << (e_bci) << ")\n"; return false; } }
+
+#define REPORT_VMSTATE(b,s)
+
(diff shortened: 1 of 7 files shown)


jitbuilder/release/src/Worklist.cpp
```diff
@@ -0,0 +1,343 @@
...
+static int32_t success = 0;
+
+// BC0: r = p;
+// BC1: r++;
+//      if (r > 10) goto BC3;
+// BC2: if (r > 5)  goto BC5;
+// BC4: if (r < 3)  goto BC1; // loop back
+// BC9: return 9;
+//
+// BC5: if (r < 8) goto BC11  (return 11)
```
charliegracie: There is no BC11 listed which is likely fine since I do not see any other way to get there and it is just supposed to return 11.

jitbuilder/release/src/Worklist.cpp
```diff
@@ -0,0 +1,343 @@
...
+// BC17: return 17;
+//
+// BC7: if (r < 17) goto BC1: // loop back
+// BC8: return 8;
+//
+//
+// BC3: if (r > 15) goto BC7  (return 7)
```
charliegracie: I do not believe this line would return the value 7.  Or does the (return 7) just signify the goto BC7?

jitbuilder/release/src/Worklist.cpp
mstoodle: yeah, I was trying to simplify; you gathered correctly that BC11 just returns 11.

I'll add 3 more lines (sounds pretty simple, doesn't it?) that make that a bit more obvious.

jitbuilder/release/src/Worklist.hpp
```diff
@@ -0,0 +1,32 @@
...
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+
+#ifndef WORKLIST_INCL
+#define WORKLIST_IMPL
```
mstoodle: Bah, will that IMPL never die!! Thanks for catching it

Issue #539, opened 2016-11-28
Delete empty enum in OMRBlock.hpp


mgaudet:
In OMRBlock.hpp it appears that [there's an empty enum](https://github.com/eclipse/omr/blob/830ebbb/compiler/il/OMRBlock.hpp#L512), an artifact of old refactoring I suspect.

```cpp
   enum //  flag bits for _moreflags
      {
      // lowest 8 bits are used for partial flags. Public declaration above.

      };

````
This is doubly confusing, since[ what _actually_ appears](https://github.com/eclipse/omr/blob/830ebbb/compiler/il/OMRBlock.hpp#L434) to be stored in `_moreflags` are 

```

   enum partialFlags // Stored in lowest 8 bits of _moreflags
      {
      _unsanitizeable         = 0x00000001,
      _containsCall           = 0x00000002,
      _restartBlock           = 0x00000004,
      _partialInlineBlock     = 0x00000008,
      _endBlock               = 0x00000010,
      _difficultBlock         = 0x00000020,
      // Available            = 0x00000040,
      // Available            = 0x00000080
      };
```

Thanks to @gurinderu for pointing this out.

mstoodle:
believe is fixed with #584, please reopen if I'm mistaken

PR #584, opened 2016-12-07
Cleanup uses of option flags

noryb009:
The first commit removes an empty enum that was supposed to be used for options, but has not been needed.

The second commit cleans up processing the `verbose` compiler option flag, specifically in how the default options are applied.

Issue: #539

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -4780,14 +4780,11 @@ char *
 OMR::Options::setVerboseBits(char *option, void *base, TR::OptionTable *entry)
    {
    VerboseOptionFlagArray *verboseOptionFlags = (VerboseOptionFlagArray*)((char*)base+entry->parm1);
-   if (entry->parm2 != 0) // This is used for -Xjit:verbose  without any options and will set first and third bit
+   if (entry->parm2 != 0) // This is used for -Xjit:verbose without any options
       {
-      intptrj_t parm2 = entry->parm2;
-      for (int32_t i = 0; parm2; i++, parm2 >>= 1)
-         {
-         if (parm2 & 1)
-            verboseOptionFlags->set((TR_VerboseFlags)i);
-         }
+      // Since no verbose options are specified, add the default options,
+      // specified in parm2 of the options table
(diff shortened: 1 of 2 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -4780,14 +4780,11 @@ char *
...
       {
-      intptrj_t parm2 = entry->parm2;
-      for (int32_t i = 0; parm2; i++, parm2 >>= 1)
-         {
-         if (parm2 & 1)
-            verboseOptionFlags->set((TR_VerboseFlags)i);
-         }
+      // Since no verbose options are specified, use the default options,
+      // specified in parm2 of the options table
+      verboseOptionFlags->replaceWord(0, entry->parm2);
```
mstoodle: I agree this fix is more straight-forward than the original implementation.

If I'm reading this change properly, though, behaviour would be different if anyone specified both the default verbose option as well as other specific options (because replaceWord would overwrite those other specific options).

@mpirvu @vijaysun-omr is that something you think people do and count on it to work additively?

mpirvu: Personally I use either -Xjit:verbose  or  -Xjit:verbose={...}, but not both at the same time.
I need some time to read and understand the verbose options code because it's convoluted. At the very least we need more documentation on how things work.

mpirvu: With the proposed code the order of options matters, that is the "verbose={...}" options that come before "verbose" will be silently eliminated, while the ones that come after will take effect.
Actually, because verboseOptionFlags->replaceWord(0, entry->parm2) nukes only the first word of verboseOptionFlags, then only the first 64 verbose options are going to be affected. Someday this is going to create a lot of confusion to someone.

noryb009: I see your point. I'll likely change it to logically ORing the first word and parm2 together. Note that parm2 is an uintptr_t, so if someone wants parm2 to hold a value > 32 bits, there will be issues setting parm2 before there are problems with the OR.

noryb009:
I've switched to using an OR to set the default flags.

PR #585, opened 2016-12-07
Replace getHighGlobalRegisterNumber call with the correct test for long type

yanluo7:
Replace getHighGlobalRegisterNumber call with the correct test for 
long type

On 64 bit, getHighGlobalRegisterNumber returns value is undefined and 
unreliable. So we need to replace it with the correct test for long type
test. If the test succeeds, we can then safely call 
getHighGlobalRegisterNumber 

### DIFF
--- a/compiler/p/codegen/OMRRegisterDependency.cpp
+++ b/compiler/p/codegen/OMRRegisterDependency.cpp
@@ -233,7 +233,7 @@ OMR::Power::RegisterDependencyConditions::RegisterDependencyConditions(
 
       if (reg!=NULL /* && reg->getKind()==TR_GPR */)
 	 {
-	 if (child->getHighGlobalRegisterNumber() > -1)
+	 if (TR::Compiler->target.is32Bit() && child->getType().isInt64())
 	    numLongs++;
 	 }
       }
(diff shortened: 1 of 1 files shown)


0xdaryl:
* Please sign your commit as per the project Contribution Guidelines

* Please change your PR and Commit titles to be imperative (as per the CG) and for the commit message to be more descriptive of the change that you are proposing

yanluo7:
@0xdaryl Daryl, I've changed the commit title and message to be more descriptive, and signed the CLA agreement. Is there anything else I need to change?

mgaudet:
I suspect your push was rejected when you `--ammend --signoff`ed it: This is the kind of circumstance where you'll want to use `git push --force` to ensure the re-written branch gets pushed. As it is, we can't see your edit, hence the ip-validation check still showing as failed.

compiler/p/codegen/OMRRegisterDependency.cpp
```diff
@@ -233,7 +233,7 @@ OMR::Power::RegisterDependencyConditions::RegisterDependencyConditions(
 
       if (reg!=NULL /* && reg->getKind()==TR_GPR */)
 	 {
-	 if (child->getHighGlobalRegisterNumber() > -1)
+	 if (TR::Compiler->target.is32Bit() && child->getType().isInt64())
```
yanluo7: Yes, this would probably be equivalent. I made these change to supplement what Julian fixed for the same issue in other P codegen parts, where he used the same isInt64 check. For consistency reason, I think we can use the same check as in other places.

vijaysun-omr:
Thanks for clarifying and fixing the commit message. Merging.

PR #587, opened 2016-12-07
Refactor OverflowCHK service in JIT builder

ShuyuLi-ca:
First, add UnsignedOverflow CHK Opcode, then refactory OverflowCHK service, such as `AddWithOverflowCHK`, `AddWithUnsignedOverflow`, etc.


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1154,16 +1154,17 @@ IlBuilder::setHandlerInfo(uint32_t catchType)
...
+   TR::ILOpCodes overflow = isSignedOverflow? TR::OverflowCHK: TR::UnsignedOverflowCHK;
+   TR::Node *overflowChkNode = TR::Node::createWithRoomForOneMore(overflow, 3, symRefTab()->findOrCreateOverflowCheckSymbolRef(_methodSymbol), operationNode, operationNode->getFirstChild(), operationNode->getSecondChild());
    overflowChkNode->setOverflowCHKOperation(operationNode->getOpCodeValue());
    genTreeTop(overflowChkNode);
    return overflowChkNode;
    }
 
 TR::IlValue *
-IlBuilder::operationWithOverflow(TR::ILOpCodes op, TR::Node *leftNode, TR::Node *rightNode, TR::IlBuilder **handler)
+IlBuilder::operationWithOverflow(TR::ILOpCodes op, TR::Node *leftNode, TR::Node *rightNode, TR::IlBuilder **handler, bool isSignedOverflow)
```
mstoodle: maybe this should be called genOverflowOperation rather than operationWithOverflow?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1193,26 +1194,36 @@ IlBuilder::operationWithOverflow(TR::ILOpCodes op, TR::Node *leftNode, TR::Node
    return resultValue;
    }
 
-TR::IlValue *
-IlBuilder::AddWithOverflow(TR::IlBuilder **handler, TR::IlValue *left, TR::IlValue *right)
+// This function takes 4 arguments and generate the addValue.
+// This function is called by AddWithOverflow and UnsignedAddWithOverflow.
+TR::ILOpCodes 
+IlBuilder::getOpCodeAndOperandNodes(TR::IlValue *leftValue, TR::IlValue *rightValue, TR::Node **leftNode, TR::Node **rightNode)
```
mstoodle: This function, once it is streamlined to only compute the opcode, should also be shared with the Add() service, I think.

compiler/il/OMRILOpCodesEnum.hpp
```diff
@@ -659,6 +659,7 @@
    ResolveAndNULLCHK, // Resolve check a static, field or method and Null check the underlying pointer.  child 1 is reference to be resolved. Symbolref indicates failure action/destination
    DIVCHK,   // Divide by zero check. child 1 is the divide. Symbolref indicates failure action/destination
    OverflowCHK, // Overflow check. child 1 is the operation node(add, mul, sub). Child 2 and child 3 are the operands. Symbolref indicates failure action/destination
+   UnsignedOverflowCHK, // UnsignedOverflow check. child 1 is the operation node(add, mul, sub). Child 2 and child 3 are the operands. Symbolref indicates failure action/destination
```
mstoodle: "Child 2 and Child 3 are the operands of the operation"

please fix for OverflowCHK too.

compiler/x/amd64/codegen/TreeEvaluatorTable.hpp
```diff
@@ -613,7 +613,8 @@
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::ResolveCHK (J9)
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::ResolveAndNULLCHK (J9)
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::DIVCHK (J9)
-   TR::TreeEvaluator::OverflowCHKEvaluator,                // TR::OverflowCHK 
+   TR::TreeEvaluator::OverflowCHKEvaluator,                // TR::OverflowCHK
+   TR::TreeEvaluator::badILOpEvaluator,                    // TR::UnsignedOverflowCHK
```
mstoodle: is the idea that a subsequent commit will implement an unsigned overflow evaluator?

If that's the case, please open an issue now to implement the unsigned overflow evaluator. I'm also slightly concerned that we're building technical debt on other platforms by only implementing these things on x86 (only amd64 even). Perhaps we need an issue to implement the overflowchk evaluators on other platforms.

mstoodle:
I'm afraid "refectory" isn't a word: please change all these to "refactor"

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -8864,6 +8864,21 @@
    },
 
    {
+   /* .opcode               = */ TR::UnsignedOverflowCHK,
+   /* .name                 = */ "UnsignedOverflowCHK",
+   /* .properties1          = */ ILProp1::TreeTop | ILProp1::HasSymbolRef,
+   /* .properties2          = */ ILProp2::Check | ILProp2::MayUseVMThread | ILProp2::MayUseSystemStack,
```
mstoodle: if we're using an exception edge and a handler block, shouldn't the OverflowCHK opcodes have CanRaiseException as well?

compiler/il/OMRILOps.hpp
```diff
@@ -261,7 +261,9 @@ class ILOpCode
    bool isTwoChildrenAddress()  const { return isArrayRef(); }
    bool isCase()                const { return getOpCodeValue() == TR::Case; }
    bool isAnchor()              const { return getOpCodeValue() == TR::compressedRefs; }
-   bool isOverflowCHK()         const { return getOpCodeValue() == TR::OverflowCHK; }
+   bool isOverflowCHK()         const { return (getOpCodeValue() == TR::OverflowCHK)         ||
```
mstoodle: I am not a huge fan of identity based queries like these. Are we using these two queries anywhere? If not, I would prefer we not create them until they are needed (at which point we can evaluate if they're really the thing that's needed).

compiler/ilgen/IlBuilder.cpp
andrewcraik: if the genOverflowCHKTreeTop is going to take an overflow opcode shouldn't this take the opcode as well since the op here doesn't tell you if you want signed or unsinged overflow checking? If this is the case perhaps a better name still would be genOperationWithOverflowCHK or genArithmeticWithOverflowCHK

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1262,7 +1260,7 @@ IlBuilder::UnsignedSubWithOverflow(TR::IlBuilder **handler, TR::IlValue *left, T
 
    TR::Node *leftNode = loadValue(left);
    TR::Node *rightNode = loadValue(right);
-   TR::IlValue *unsignedSubValue = operationWithOverflow(TR::ILOpCode::unsignedSubtractOpCode(leftNode->getDataType()), leftNode, rightNode, handler);
+   TR::IlValue *unsignedSubValue = operationWithOverflow(TR::ILOpCode::unsignedSubtractOpCode(leftNode->getDataType()), leftNode, rightNode, handler, false /*isSignedOverflow*/);
```
andrewcraik: This function sounds like it is generating some kind of usub like iusub or csub or lusub - these are deprecated opcodes - we should be generating the isub, ssub, lsub and the signedness only comes in to the overflow check so I think this needs to be fixed.

mgaudet: Should we have an issue open to delete the unsigned opcodes?

andrewcraik: Soon - I need to consider how mul and div fit with add/sub/etc since there is a bit of cognitive dissonance there atm. Once that is sorted I'll open an issue for the work that needs to be done. For now I'm just trying to reduce the spread of the opcodes pending a final design. Note we will need to keep all unsigned conversions.

compiler/il/OMRILOps.hpp
cathyzhyi: it's currently used in an assert in OMR::Node::setNodeRequiresConditionCodes(bool v) to check if the node needs conditional check is reasonable. The assert checks "op.isArithmetic() || op.isLoadConst() || op.isOverflowCHK();" [https://github.com/eclipse/omr/blob/master/compiler/il/OMRNode.cpp#L8482](url)  The || here serves for this purpose.

compiler/x/codegen/OMRTreeEvaluator.hpp
```diff
@@ -346,6 +347,10 @@ class OMR_EXTENSIBLE TreeEvaluator: public OMR::TreeEvaluator
 
    static intptrj_t integerConstNodeValue(TR::Node *node, TR::CodeGenerator *cg);
 
+   static TR::Block *GetOverflowCatchBlock(TR::Node *node, TR::CodeGenerator *cg);
+
+   static void GenArithmeticInstructions(TR::Node *node, TR::CodeGenerator *cg);
```
andrewcraik: same thing here - lowercase first letter

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2431,7 +2431,31 @@ TR::Register *OMR::X86::TreeEvaluator::andORStringEvaluator(TR::Node *node, TR::
...
+   TR_ASSERT((bbstartNode->getOpCodeValue() == TR::BBStart), "catch block entry %p must be TR::BBStart\n", bbstartNode);
+   if (!bbstartNode->getLabel())
+      {
+      TR::LabelSymbol *label = generateLabelSymbol(cg);
+      bbstartNode->setLabel(label);
+      }
+   return overflowCatchBlock;
+   }
+
+void OMR::X86::TreeEvaluator::GenArithmeticInstructions(TR::Node *node, TR::CodeGenerator *cg)
```
andrewcraik: The fact that this takes an OverflowCHK / UnsignedOverflowCHK node is surprising given the node - I think this should have a better name and assertion to check the opcode of node is what is expected.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1230,14 +1230,14 @@ IlBuilder::AddWithOverflow(TR::IlBuilder **handler, TR::IlValue *left, TR::IlVal
...
+IlBuilder::AddWithUnsignedOverflow(TR::IlBuilder **handler, TR::IlValue *left, TR::IlValue *right)
    {
    TR::Node *leftNode = loadValue(left);
    TR::Node *rightNode = loadValue(right);
    TR::ILOpCodes opcode = getOpCode(left, right);
    TR::IlValue *addValue = genOperationWithOverflowCHK(opcode, leftNode, rightNode, handler, TR::UnsignedOverflowCHK);
-   TraceIL("IlBuilder[ %p ]::%d is UnsignedAddWithOverflow %d + %d\n", this, addValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
-   //ILB_REPLAY("%s = %s->UnsignedAddWithOverflow(%s, %s, %s);", REPLAY_VALUE(addValue), REPLAY_BUILDER(this), REPLAY_PTRTOBUILDER(handler), REPLAY_VALUE(left), REPLAY_VALUE(right));
+   TraceIL("IlBuilder[ %p ]::%d is AddWithUnsignedOverflow %d + %d\n", this, addValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
+   //ILB_REPLAY("%s = %s->AddWithUnsignedOverflow(%s, %s, %s);", REPLAY_VALUE(addValue), REPLAY_BUILDER(this), REPLAY_PTRTOBUILDER(handler), REPLAY_VALUE(left), REPLAY_VALUE(right));
```
andrewcraik: Why is this code commented out?

ShuyuLi-ca: It seems that `REPLAY_PTRTOBUILDER` haven't been implemented yet.

mstoodle: If you mean the replay macro: that's because the replay support is currently stuck in time and can't handle builder pointers properly. I've been asking people to create the line in new services, so it's there but, since it can't compile, to comment it out for now.

compiler/il/OMRNode.cpp
```diff
@@ -4675,14 +4675,18 @@ OMR::Node::setArrayComponentClassInNode(TR_OpaqueClassBlock *c)
 TR::ILOpCodes
 OMR::Node::setOverflowCHKOperation(TR::ILOpCodes op)
```
mstoodle: similarly, i think this should be setOverflowCheckOperation

compiler/il/OMRNode.cpp
```diff
@@ -4675,14 +4675,18 @@ OMR::Node::setArrayComponentClassInNode(TR_OpaqueClassBlock *c)
...
+             (self()->getOpCodeValue() == TR::UnsignedOverflowCHK), 
+             "set OverflowCHK operation info for no OverflowCHK node");
    return _unionBase._extension.getExtensionPtr()->setElem<TR::ILOpCodes>(3, op);
    }
 
 TR::ILOpCodes
 OMR::Node::getOverflowCHKOperation()
    {
-   TR_ASSERT(self()->getOpCodeValue() == TR::OverflowCHK, "get OverflowCHK operation info for no OverflowCHK node ");
+   TR_ASSERT((self()->getOpCodeValue() == TR::OverflowCHK) ||
```
mstoodle: ...and getOverflowCheckOperation...

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2560,37 +2566,28 @@ TR::Register *OMR::X86::TreeEvaluator::OverflowCHKEvaluator(TR::Node *node, TR::
...
+   TR::TreeEvaluator::genArithmeticInstructionsForOverflowCHK(node, cg);
+   generateLabelInstruction(JO4, node, overflowCatchBlock->getEntry()->getNode()->getLabel(), cg);
+   cg->setVMThreadRequired(false);
+   cg->decReferenceCount(node->getFirstChild());
+   return NULL;
    }
 
-   TR_X86OpCodes opcode = isUnsigned? JB4: JO4; 
-   generateLabelInstruction(opcode, node, overflowCatchBlock->getEntry()->getNode()->getLabel(), cg);
+TR::Register *OMR::X86::TreeEvaluator::unsignedOverflowCHKEvaluator(TR::Node *node, TR::CodeGenerator *cg)
```
mstoodle: This code looks virtually identical to overflowCHKEvaluator. Could you use the same function and just pick JO4 versus JB4 based on the node's opcode?

mstoodle:
Making good progress, @ShuyuLi-ca . It's almost ready for merging, but still a couple of issues to address, please.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2560,34 +2566,16 @@ TR::Register *OMR::X86::TreeEvaluator::OverflowCHKEvaluator(TR::Node *node, TR::
...
-           isUnsigned = true;
-           break;
    }
 
-   TR_X86OpCodes opcode = isUnsigned? JB4: JO4; 
+TR::Register *OMR::X86::TreeEvaluator::overflowCHKEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR::Block *overflowCatchBlock = TR::TreeEvaluator::getOverflowCatchBlock(node, cg);
+   TR::TreeEvaluator::genArithmeticInstructionsForOverflowCHK(node, cg);
+   TR_X86OpCodes opcode = node->getOpCodeValue() == TR::OverflowCHK? JO4: JB4;
```
mstoodle: I would prefer this line to be expanded for the two OverflowCHK cases we have now along with a TR_ASSERT if an opcode other than those two go through here, as some small amount of future proofing. i.e. something like:
```
if (node->getOpCodeValue() == TR::OverflowCHK)
   opcode = JO4;
else if (node->getOpCodeValue() == TR::UnsignedOverflowCHK)
   opcode = JB4;
else
   TR_ASSERT(0, "unrecognized overflow operation in overflowCHKEvaluator");
```

ShuyuLi-ca:
@mstoodle Thanks a lot. So I deleted the previous TR_Assert that I added in Line 2468 as well. :)

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -8852,7 +8852,22 @@
    /* .opcode               = */ TR::OverflowCHK,
    /* .name                 = */ "OverflowCHK",
    /* .properties1          = */ ILProp1::TreeTop | ILProp1::HasSymbolRef,
-   /* .properties2          = */ ILProp2::Check | ILProp2::MayUseVMThread | ILProp2::MayUseSystemStack,
+   /* .properties2          = */ ILProp2::Check | ILProp2::MayUseVMThread | ILProp2::MayUseSystemStack | ILProp2::CanRaiseException | ILProp2::CanRaiseException,
```
mstoodle: Why is ILProp2::CanRaiseException here twice?

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -8852,7 +8852,22 @@
...
+   /* .reverseBranchOpCode  = */ TR::BadILOp,
+   /* .booleanCompareOpCode = */ TR::BadILOp,
+   /* .ifCompareOpCode      = */ TR::BadILOp,
+   },
+
+   {
+   /* .opcode               = */ TR::UnsignedOverflowCHK,
+   /* .name                 = */ "UnsignedOverflowCHK",
+   /* .properties1          = */ ILProp1::TreeTop | ILProp1::HasSymbolRef,
+   /* .properties2          = */ ILProp2::Check | ILProp2::MayUseVMThread | ILProp2::MayUseSystemStack | ILProp2::CanRaiseException | ILProp2::CanRaiseException,
```
mstoodle: ILProp2::CanRaiseException is also here twice

compiler/il/OMRNode.cpp
```diff
@@ -4673,16 +4673,20 @@ OMR::Node::setArrayComponentClassInNode(TR_OpaqueClassBlock *c)
    }
 
 TR::ILOpCodes
-OMR::Node::setOverflowCHKOperation(TR::ILOpCodes op)
+OMR::Node::setOverflowCheckOperation(TR::ILOpCodes op)
    {
-   TR_ASSERT(self()->getOpCodeValue() == TR::OverflowCHK, "set OverflowCHK operation info for no OverflowCHK node");
+   TR_ASSERT((self()->getOpCodeValue() == TR::OverflowCHK) ||
```
mstoodle: you should leverage isOverflowCheck() here :)

compiler/il/OMRNode.cpp
```diff
@@ -4673,16 +4673,20 @@ OMR::Node::setArrayComponentClassInNode(TR_OpaqueClassBlock *c)
...
+             "set OverflowCHK operation info for no OverflowCHK node");
    return _unionBase._extension.getExtensionPtr()->setElem<TR::ILOpCodes>(3, op);
    }
 
 TR::ILOpCodes
-OMR::Node::getOverflowCHKOperation()
+OMR::Node::getOverflowCheckOperation()
    {
-   TR_ASSERT(self()->getOpCodeValue() == TR::OverflowCHK, "get OverflowCHK operation info for no OverflowCHK node ");
+   TR_ASSERT((self()->getOpCodeValue() == TR::OverflowCHK) ||
```
mstoodle: You should leverage isOverflowCheck() here

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1222,7 +1222,7 @@ IlBuilder::getOpCode(TR::IlValue *leftValue, TR::IlValue *rightValue)
    TR::ILOpCodes op;
    if (leftValue->getSymbol()->getDataType() == TR::Address)
       {
-      if (leftValue->getSymbol()->getDataType() == TR::Int32)
+      if (rightValue->getSymbol()->getDataType() == TR::Int32)
```
mstoodle: i would prefer you fix this problem in the commit where leftValue was originally introduced. Otherwise you introduced a commit with broken code in it that doesn't need to be broken.

mstoodle:
LGTM, just waiting for the CI tests to complete now.

PR #592, opened 2016-12-09
Complete CS cycle in progress before starting STW global GC

amicic:
More generic solution of https://github.com/eclipse/omr/pull/579. This
one additionally covers the case of plain Parallel Global GC, rather
than more typical ConcurrentGC (STW phase of it).

An example is running Java's Gencon GC policy with Concurrent Marking
explicitly disabled.

### DIFF
--- a/gc/base/Collector.cpp
+++ b/gc/base/Collector.cpp
@@ -16,14 +16,18 @@
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
  ******************************************************************************/
 
+#include "AllocateDescription.hpp"
 #include "Collector.hpp"
 #include "CollectorLanguageInterface.hpp"
 #include "GCExtensionsBase.hpp"
-#include "ModronAssertions.h"
 #include "Heap.hpp"
-#include "AllocateDescription.hpp"
-#include "OMRVMThreadListIterator.hpp"
 #include "MemorySubSpace.hpp"
+#include "ModronAssertions.h"
+#include "OMRVMThreadListIterator.hpp"
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+#include "Scavenger.hpp"
+#endif /* OMR_GC_CONCURRENT_SCAVENGER */
+
 
 class MM_MemorySubSpace;
 class MM_MemorySpace;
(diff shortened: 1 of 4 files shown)


gc/base/Collector.cpp
```diff
@@ -192,6 +196,10 @@ MM_Collector::preCollect(MM_EnvironmentBase* env, MM_MemorySubSpace* subSpace, M
 {
 	MM_GCExtensionsBase* extensions = env->getExtensions();
 
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+	completeConcurrentScavenge(env);
```
charliegracie: @amicic should this be on GlobalCollector instead?  Isn't the Scavenger also a sub class of collector?

amicic: Valid question. To avoid this being called for Scavenger, I check cycleType within completeConcurrentScavenge to not be Scavenger.

Some background facts before deciding if we want move it somewhere else:
This should be called before internalPreCollect (or early in the method), because it is the one that will issue cycle start (for global GC). The goal of this change is to enforce completion of Scavenger before any real action is done with global cycle.
Collector::internalPreCollect is virtual and has its implementation for ParallelGlobalGC, ConcurrentGC (and Scavenger). It does not have its implementation in GlobalCollector.
Collector::preCollect is not virtual, hence does not have implementation in GlobalCollector.

Currently, there is no place in GlobalCollector to call it from while perserving the requirement to be called before internalPreCollect.

An alternative approach is to just call it at two spots, early in internalPreCollect for both ConcurrentGC and ParallelGlobalGC (and than remove that cycleType check in completeConcurrentScavenge).

Another alternative is to make preCollect virtual and implement for GlobalCollector, which would call completeConcurrentScavenge and than parent preCollect, which would than call appropriate internalPreCollect. Since there is no implementation of preCollect for Scavenger, the parent one would be called right away, thus no call to completeConcurrentScavenge  would be done.

Other suggestions or preferences on the proposed alternatives?

charliegracie: Thanks for the explanation.  I think this is likely the best location for now.

PR #593, opened 2016-12-09
Remove dead libraryHandle from compiler TestDriver

mgaudet:
Not clear what the intention of this was, but is no longer used anywhere
and can be safely removed.


mstoodle:
At one point very early in the development of the test compiler, I was trying to put the test compiler into a shared object that could be linked against a (much smaller) test executable. I abandoned this approach at the time, but I think it's something we should revisit.

Nonetheless, there's no point keeping this vestigial stuff around. It can be readily recreated later.

PR #595, opened 2016-12-12
Add more documentation to extensible classes

noryb009:
This rephrases and extends the current documentation. New documentation is also added, including extensible enums, comparisons to other approaches, and connector guards.

### DIFF
--- a/doc/compiler/extensible_classes/Extensible_Classes.md
+++ b/doc/compiler/extensible_classes/Extensible_Classes.md
@@ -4,70 +4,94 @@
 
 Extensible classes are a way of implementing static polymorphism in C++. They allow various projects to override and extend the
 functionality of an existing class, through inheritance, in order to suit their specific needs. This is done by carefully organizing a
-project in a way that lets the compiler find the most extended implementation of a function.
+project in such a way that the compiler can find the most extended implementation of a member function, no matter where it is called from.
 
-This document describes the general characteristics of extensible classes as well as their implementation. Although there are alternative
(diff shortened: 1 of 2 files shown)


doc/compiler/extensible_classes/Extensible_Classes.md
```diff
@@ -161,11 +184,45 @@ Pros:
...
+### Namespace Aliases
+
+Namespace aliases are a proposed alternative to connectors.
+This method makes namespaces accumulative, where if a class is not found in a namespace at compile time, the compiler looks in its parent namespace.
+
+This greatly reduces the boilerplate code required to implement connectors, including the macros defined above a file's includes.
+Additionally, this simplifies the question of what to inherit from.
+
+This method however has a few cons.
+If the wrong class is included when inherited from, it would not be used in the class hierarchy.
```
mgaudet: This sentence is a little strange...

mgaudet:
You appear to be working w/ the 1-line per sentence model of prose. While that makes sense for rendered content (preserves diffability), IMO, for our markdown we ought to wrap at 80.

noryb009:
@mgaudet Thanks for reviewing.

There's currently a mix of documentation styles (eg. each paragraph on one line, to 80 characters per line), so we might need to have a wider discussion about what style we should use moving forward. For now, I've changed this patch to the 80 characters per line style.

0xdaryl:
$0.02.  From a reviewer's perspective, the other advantage of wrapping at 80 chars is that its easier to add per-line comments closer to where the comment actually applies (as opposed to the end of a paragraph).

PR #596, opened 2016-12-13
Move and beef up dependency helper class

zl-wang:
TR_PPCRegisterDependencyMap which can be used in other places.


zl-wang:
Also, edited the auto TABs there historically.

zl-wang:
Helper class TR_PPCRegisterDependencyMap was in OMRRegisterDependency.cpp.  That means it cannot be used in other places, such that I moved it to OMRRegisterDependency.hpp.  Also, adding a couple utility routines for convenience:

void addDependency(TR::RegisterDependency *dep, uint32_t index)

and

uint8_t getTargetIndex(TR::RealRegister::RegNum rr)

Also, cleaning up the TABs in original file.

zl-wang:
Done.

Issue #569, opened 2016-12-05
Optimization Test Framework


noryb009:
Currently, there are 2 ways to test optimizations:

* Optimize a program, and ensure it gives the same output as without the optimization.
* Run with `traceFull,log=logfile` and verify the IL was correctly modified.

Nether of these are great tests. While the former usually ensures the output is valid, it doesn't check if the optimization was applied correctly, or even at all. The latter addresses this shortcoming, but the process is manual, time consuming, and one needs to understand the optimization to properly verify the optimization was correctly applied.

This is a proposed method to allow the latter method to be run as an automated test. It is essentially a framework for running a given sequence of optimizations on an ILInjector, and exposes a hook to allow the test to verify IL. Doing codegen is optional.

An example test can be seen [here](https://github.com/noryb009/omr/commit/3a34b8343e391896e92aecf1920cf9daffbd365e#diff-9ede7d3de302142136adcf5f1353741dR32). Essentially, the test sets a MethodInfo (which itself takes an IlInjector), a verifier, and a list of optimizations (in this case, the list is in the test fixture constructor).

Verifiers have very few restrictions - they are only expected to return a non-zero status code on an error. In test cases, they are expected to use gtest's `EXPECT` and `ASSERT` macros to fail the test, if required.

Verifiers can also be re-purposed for general IL verification - `verify{Trees,Blocks}` could be rewritten as verifiers, and specified when required by a downstream project. This also allows downstream projects to apply their own constraints to IL, if required.

The full change set can be seen in the last few commits on my [opt-test](https://github.com/noryb009/omr/commits/opt-test) branch.

I'm interested to what the community thinks about this. Thoughts and constructive criticism is welcomed.

I would like to thank Filip and Mark for their ideas in #382, as well as Matthew and Luc for their suggestions made offline.

Leonardo2718:
Awesome work! :+1:

I really think this is a great start.

As an item for later work, I'd like to see if it's possible to further decouple the verification from the compilation process (i.e. be able to "hide" the test compiler a little more). I think this would help clean up and reduce the size of test cases, hopefully making it easier to create better tests.

noryb009:
@mgaudet 
1. That is definitely possible. I'll switch it soon.
[...]

PR #599, opened 2016-12-14
Initial framework for testing optimizations

noryb009:
This adds a framework for testing individual optimizations. It allows the test to perform tasks required for optimization testing without a lot of boilerplate code. Notably, it provides a hook that allows tests to examine IL after an optimization has been applied in order to confirm it has been applied correctly.

These commits should likely be reviewed independently, since each commit stands on its own (except for some parts not being used until the end). Motivation can be see in the associated issue, while implementation and design details are addressed in the commit messages and comments.

Issue: #569


fvtest/compilertest/ilgen/TestIlGeneratorMethodDetails.hpp
```diff
@@ -71,9 +78,14 @@ class OMR_EXTENSIBLE IlGeneratorMethodDetails : public OMR::IlGeneratorMethodDet
                                           bool forceClassLookahead,
                                           TR_InlineBlocks *blocksToInline);
 
+#define IL_VERIFY
```
mgaudet: Not a huge fan of the define in here: IMO, it should be added to the fvtest makefiles

fvtest/compilertest/ras/IlVerifierHelpers.hpp
```diff
@@ -0,0 +1,132 @@
...
+      {
+      _hasRun = true;
+
+      if(_ilVerifier)
+         {
+         // If verify() does not return, have a non-zero return code.
+         _rc = 1;
+         _rc = _ilVerifier->verify(methodSymbol);
+         }
+      else
```
mgaudet: Style wise, I think we should have braces around this (to match the `if `above)

fvtest/compilertest/ilgen/MethodInfo.hpp
```diff
@@ -0,0 +1,94 @@
...
+   void DefineILInjector(TR::IlInjector *ilInjector)
+      {
+      _ilInjector = ilInjector;
+      }
+
+   /**
+    * Return a TR::ResolvedMethod based on the defined function and
+    * IlInjector.
+    *
+    * Before calling this, the function signature and IlInjector must
```
mgaudet: Prime candidate for an assertion.

fvtest/compilertest/tests/SimplifierFoldAndTest.cpp
```diff
@@ -19,54 +19,181 @@
...
+ * Errors are added by `ASSERT_*` and `EXPECT_*`.
+ * `ASSERT_*` immediately end compilation, so other
+ * verifiers and functional tests do not run.
+ * `EXPECT_*` only adds an error, which allows the user to
+ * see if there are other failures later.
+ *
+ * While anything can be done in `verify`, this test
+ * is implmented by traversing the IL, and checking if
+ * there are AND operations.
+ */
```
mgaudet: Is it just me or is this comment not actually describing the body below (`ASSERT/EXPECT` vs `ADD_FAILURE`)

compiler/control/CompileMethod.cpp
```diff
@@ -312,6 +312,10 @@ compileMethodFromDetails(
             compiler.setDebug(createDebugObject(&compiler));
             }
 
+#ifdef VERIFY_IL
```
0xdaryl: Is there a reason why this is guarded with an `#ifdef` rather than a compiler option?  I'd rather have it live and enable-able via an option.

mgaudet:
Once this gets merged, we should open followup issues based on @andrewcraik's excellent summary of other-things-to-keep-in-mind [here](https://github.com/eclipse/omr/issues/569#issuecomment-268811761) . In particular, providing ways to inject frequency seems a top need.

0xdaryl:
Thanks for the great feedback @andrewcraik .  We will certainly incorporate that input into the evolution of this framework.

Just to be clear, in your mind does any of your feedback preclude merging this PR in its current form?

andrewcraik:
I don't think we should hold this out. I don't think it is ready to use in production, but that is no reason to not contribute the code as a starting point and then work on rounding it in subsequent pull requests out before putting it to use.

compiler/control/CompileMethod.cpp
noryb009: 1. There is no "good" default for the option, unless you make it depend on the debug option (and even then, it could just use upon debug or prod with assumes instead of a new option).
2. Currently, this is only implemented in the Test compiler, making it not make sense to either enable in a different project, or disable it in the Test compiler (since it would just stop some tests from working).

If verifiers are used in other projects in the future, this will need to be revisited. In particular, I feel verifiers would be implemented project-wide, not method-wide, meaning this code would need to be removed anyway.

0xdaryl: Having an option depend on a build flag or configuration is not unprecedented (its not necessarily ideal, but our options framework needs a lot of re-work).

I think a runtime option is ideal, but I'm not going to press the issue for this PR.  However, if this is only dependent on the Test compiler at the moment and will be for some time, then I think using the `#ifdef` `TEST_PROJECT_SPECIFIC` is a better choice than `VERIFY_IL`.  Is `VERIFY_IL` used anywhere else?

noryb009: No, it is not. I will switch to `TEST_PROJECT_SPECIFIC`.

I hope it won't be this way for very long - in January, I plan on building out this framework further and will try to refactor current verifiers into using the new system, however I am unsure of how much free time I will have to do this.

ymanton:
I happened to be doing something that would benefit from this sort of testing so I gave it a try and it was pretty simple to get it working.

One comment, have you given any thought on how a test can generate multiple compiled methods to test? I can sort of see how this might work -- add some sort of interface to the `Info` class that influences the function it defines and what the IL injector emits and make multiple calls to that together with `Verify*()` in your `TEST_F()` -- but I haven't tried it.

Having a test be able to generate multiple compiled methods is pretty useful, you don't want to force someone to generate one giant function if they have a lot of variations to test, you would prefer that each variation be tested in it's own compiled.

0xdaryl:
@younes : thanks for your comment.  Unless its specific to this PR, would you mind moving it to the Issue (#569) where we can continue the discussion?

noryb009:
The build has failed due to JitBuilder thinking it is the Test compiler. This is addressed in #635.

Issue #223, opened 2016-09-21
OS/X JIT support 


mgaudet:
Today we do not build on OS/X, however, it is a popular platform that we should be targeting, especially since the rest of OMR already supports OS/X.

charliegracie:
@mgaudet do you know how different the ABI is when comparing Linux x86-64 to OSX?

Do you have any gut feeling on the type of work required to bring up OSX?  If there are any easy steps it may be a good opportunity to have someone external to the project work on it.

lmaisons:
Not Matthew, but...

For reference: https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/LowLevelABI/000-Introduction/introduction.html

x86-64 is stated as using the AMD64 System V ABI and Apple provides no information beyond a link to the standard document (version 0.90 from December 2003).  This should be directly compatible with Linux unless something significant happened across different versions (latest obvious version is 0.99.6 from December 2013) *and* Linux / GCC decided to re-target the newer version.  The main problem I would expect to encounter is assumptions in our code that Linux is the only POSIX-like platform of relevance for x86-64, and thus some of our uses of various system services will likely require re-evaluation / refactoring / re-implementation for OS X / FreeBSD.

lmaisons:
At one point FreeBSD user space and OS X were *very* similar.  If that continues to be the case, it may be worth using a FreeBSD build environment as a proxy if Macs are scarce.

charliegracie:
Cool thanks.  We have access to OS X machines on travis so we can always enable it here and get basic testing.  Also I have been thinking ahead and we have some OS X machines in Ottawa :)

0xdaryl:
@noryb009 will investigate and implement what needs to be done for OMR compiler support on OS X.

noryb009:
#418 provides support for OS X in the JIT.

Be aware that there are likely still bugs in OS X that aren't encountered by the Test compiler and JitBuilder.

mgaudet:
... this has worked for a while now :)

PR #600, opened 2016-12-15
Label TRlist workaround as libc++ only

noryb009:
A workaround, previously thought to be caused by OS X, was actually caused by libc++. The issue has been fixed upstream in [LLVM bug 31378](https://llvm.org/bugs/show_bug.cgi?id=31378), and will be in libc++ 4.0. This corrects the include guard and adds a comment describing the reason for the workaround.


noryb009:
@mstoodle I don't believe for this problem specifically. This originally came from #223, but most of this change is unrelated.

mstoodle:
Ok, thanks, @noryb009 . I'll merge it as the fix is very self-contained and clear as-is.

PR #605, opened 2016-12-15
Fix exception egdes for OSRCatch block in Block::splitEdge

cathyzhyi:
Fix the control flow when spliting an exception edge to OSRCatch block to be **from -> exceptionEdge -> newBlock -> normalEdge -> to**

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1268,6 +1268,14 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
          }
       }
 
+   //For OSRCatchBlock, the newBlock to be created below needs to match the bcInfo of the induction point.
+   //Since the original OSRCatchBlock has the same bcInfo as the induction point, we just use entry node from the
+   //original OSRCatchBlock.
+   if (to->isOSRCatchBlock())
+      {
+      exitNode = to->getEntry()->getNode();
+      }
+
    TR::CFG * cfg = c->getFlowGraph();
    TR::Compilation *comp = cfg->comp();
    TR::TreeTop *entryTree = to->getEntry();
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
I would like @andrewcraik to also review this.

compiler/il/OMRBlock.cpp
```diff
@@ -1334,23 +1334,27 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
...
-                  -1, 
-                  to->getOwningMethod(), 
-                  c);
-         else
-            newBlock->setHandlerInfo(
-               to->getCatchType(),
-               to->getInlineDepth(),
-               to->getHandlerIndex(),
-               to->getOwningMethod(),
+         newBlock->setHandlerInfoWithOutBCInfo(
```
andrewcraik: I am worried about the BCInfo of the catch since the BC info is going to match the induction point and probably not the BCinfo of the block. It, in general, should not make a difference, but the potential difference makes me worry about a hard to find bug possibility resulting from this code.

andrewcraik:
The control flow in splitEdge seems very complicated since there is an
if (isExceptionEdge && to->isOSRCatchBlock())
then
if (isExceptionEdge && !to->isOSRCatchBlock())
then
if (isExceptionEdge) with an isOSRCatchBlock test inside.

Shouldn't this be tidied yup as part of this change since it seem a bit of a mess and could be hiding bugs?

compiler/il/OMRBlock.cpp
cathyzhyi: Thanks a lot for pointing this out. Modified the code to make the catch block match BCInfo of induction point.

cathyzhyi:
I am closing this pull request after discussion with Andrew. The fix is too risky because it will change the assumption that each caller index only have one osrCatchBlock. We will fix the problem in a more conservative way.

PR #610, opened 2016-12-16
Add transactional memory services in JitBuilder

fredqiu:
1. Add two new services (transaction, transactionAbort) in JitBuilder and a new test case.
2. In transaction, 
    a) if platform supports xbegin, service will branch into three directions: persistentFailure (permanent failure, go back to use atomic way), transientFailure (temporary failure, will retry), transaction(if success, use non-atomic way to handle).
    b) if platform doesn't support xbegin, we will walk around transaction and transientFailure paths and go to persistentFailure directly.
3.  Test case TransactionOperations.cpp/hpp: tested on both platform supports xbegin and the one doesn't.  
4. Since now we only have tstart evaluator in X, I use an assertion in the beginning of transaction() service to rule out other platforms.
5. There are many versions of Haswell and a small number of Broadwell having defects with TM. (Now we assume only the version of Broadwell supports TM, ignoring that small number of problematic ones for now. Also, I leave a "TODO" in the code.) Firstly, I added Broadwell in ProcessorInfo; Secondly, now I added an "if" in x/OMRCodeGenerator.cpp. If processor is haswell, even though CPUID shows it supports TM, don't do setSupportsTM().


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1564,6 +1564,114 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    TraceIL("IlBuilder[ %p ]::transactionBegin %p, %p, %p, %p)\n", this, *persistentFailureBuilder, *transientFailureBuilder, *fallThroughBuilder);
+   
+    appendBlock();
+
+    TR::Block *mergeBlock = emptyBlock();
+    *persistentFailureBuilder = createBuilderIfNeeded(*persistentFailureBuilder);
+    *transientFailureBuilder = createBuilderIfNeeded(*transientFailureBuilder);
+    *fallThroughBuilder = createBuilderIfNeeded(*fallThroughBuilder);
+
+    if (!TR::Compiler->target.cpu.getX86SupportsTM(comp()))
```
mgaudet: Is there no query that will also support P/Z TM?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1564,6 +1564,114 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+ *    transactionalRegion
+ *    ...
+ * unlock (TrasanctionEnd) 
+ *
+ * \structure & \param value
+ *     tstart
+ *       |
+ *       ----persistent// This is a permanent failure, try atomic way to do it instead
+ *       |
+ *       ----transient//Temperal failure, user can retry
```
mgaudet: Spelling error: 'Temporary'

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1564,6 +1564,114 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
 
+/**
+ * \brief
+ *  the service is for generating a treetop for transaction begin when the user needs to use transactional memory.
+ *
+ * lock (TransactionBegin)
+ *    ...
+ *    transactionalRegion
+ *    ...
+ * unlock (TrasanctionEnd)
```
mgaudet: While it is possible to implement transactional lock elision using this service, it doesn't directly implement it; IMO, it's misleading to talk about a lock in this context here. I'd drop this section, and replace with a pseudocode implementation or basic-block diagram of control flow.

jitbuilder/release/src/TransactionalOperations.cpp
```diff
@@ -0,0 +1,131 @@
...
+TransactionalMethod::buildIL()
+   {
+   cout << "TransactionalMethod::buildIL() running!\n";
+
+   TR::IlBuilder *persistent=NULL;
+   TR::IlBuilder *transient=NULL;
+   TR::IlBuilder *fallthrough=NULL; 
+
+   //Service will return true if the platform supports TM
+   TransactionBegin(&persistent, &transient, &fallthrough);
```
mgaudet: Comment doesn't appear to match code.

compiler/il/OMRNode.hpp
```diff
@@ -333,7 +333,7 @@ class OMR_EXTENSIBLE Node
       TR::ILOpCode opcode; opcode.setOpCodeValue(opvalue);
       TR_ASSERT(!opcode.isIf(), "use createif or createbranch on this node\n");
       TR_ASSERT(opvalue != TR::arraycopy, "use createArraycopy to create this node");
-      TR_ASSERT(opvalue != TR::tstart, "use createWithRoomForFive to create node: %s", opcode.getName());
+      //TR_ASSERT(opvalue != TR::tstart, "use createWithRoomForFive to create node: %s", opcode.getName());//since using create(3,a,b,c) instead, need to comment this for now
```
andrewcraik: Yeah that doesn't seem like it should be right with a superficial look at the code...

compiler/x/env/OMRCPU.cpp
```diff
@@ -144,5 +150,8 @@ OMR::X86::CPU::getX86OSSupportsSSE2(TR::Compilation *comp)
 bool
 OMR::X86::CPU::getX86SupportsTM(TR::Compilation *comp)
    {
-   return false;
+   uint32_t flags8 = self()->getX86ProcessorFeatureFlags8(comp);
```
andrewcraik: I'm not sure if this is the right query due to the possibility that some Haswell chips could be running with out the microcode update that disabled this feature due the hardware having implementation bugs. For true safety this check probably needs to ensure the processor is of a sufficiently new model as well as having the feature flag set to avoid weird problems from bugs in earlier hardware which may not have been patched.

andrewcraik: Also does this Flags8 need to be guarded for older hardware that may not have this many flags? ie when was flags8 introduced and will we hit any problems in old hardware if we attempt to read it?

compiler/x/i386/codegen/TreeEvaluatorTable.hpp
```diff
@@ -564,9 +564,9 @@
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::monent (J9)
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::monexit (J9)
    TR::TreeEvaluator::badILOpEvaluator,                    // TR::monexitfence (J9)
-   TR::TreeEvaluator::badILOpEvaluator,                  // TR::tstart (J9, WCode)
-   TR::TreeEvaluator::badILOpEvaluator,                 // TR::tfinish (J9, WCode)
-   TR::TreeEvaluator::badILOpEvaluator,                 // TR::tabort(J9, WCode)
+   TR::TreeEvaluator::tstartEvaluator,                  // TR::tstart (J9, WCode)
```
andrewcraik: Why the mentions of WCode here? Why the mentions of J9? This is generic OMR code right?

jitbuilder/release/Makefile
```diff
@@ -44,6 +45,7 @@ ALL_TESTS = \
             toiltype
 
 goal: $(ALL_TESTS)
+	./transactionaloperations
```
andrewcraik: What happens if the test is run on hardware without support?

jitbuilder/release/src/TransactionalOperations.cpp
```diff
@@ -0,0 +1,131 @@
+/*******************************************************************************
```
andrewcraik: The test will hit the success pathways ie the TSTART and TCOMMIT - I am not sure it tests anything related to TABORT. Arranging a test of TABORT will be much harder, but perhaps that should be considered to ensure full functional validation?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4779,6 +4779,148 @@ TR::Register *OMR::X86::TreeEvaluator::atomicorEvaluator(TR::Node *node, TR::Cod
...
+   {
+   generateInstruction(XEND, node, cg);
+   return NULL;
+   }
+
+TR::Register *
+OMR::X86::TreeEvaluator::tabortEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   //we hardcode an abort reason as 0x04 just for simplicity.
+   generateImmInstruction(XABORT, node, 0x04, cg);
```
andrewcraik: This seems dangerous and ill advised - if it has to survive there had better at least be a big TODO here to fix this.

compiler/x/i386/codegen/TreeEvaluatorTable.hpp
fredqiu: Yes, this is for generic OMR code. I should remove the WCode thing.

jitbuilder/release/Makefile
fredqiu: In my service, I handled the situation where the platform doesn't support xbegin by directly going to persistentFailure path. I tested my test case both on platform does support and the one doesn't.

andrewcraik: Make sure that the behaviour you describe is documented somewhere on the service to make it clear as it could be surprising to some.

jitbuilder/release/src/TransactionalOperations.cpp
fredqiu: yes, TABORT is related to nothing here. I added it only considering the integrity of TM services (since tstart/tend/tabort always appears together) and just in case others might use it in the future. Should I remove it?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+ *     tstart
+ *       |
+ *       ----persistentFailure // This is a permanent failure, try atomic way to do it instead
+ *       |
+ *       ----transientFailre // Temporary failure, user can retry 
+ *       |
+ *       ----fallthrough // Success, use general(non-atomic) way to do it
+ */
+void
+IlBuilder::TransactionBegin(TR::IlBuilder **persistentFailureBuilder, TR::IlBuilder **transientFailureBuilder, TR::IlBuilder **fallThroughBuilder)
```
mstoodle: So fallThroughBuilder is the transactional code? I think we should use a better name for it, like transactionBuilder or something like that.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+ *              +----------------+   
+ *
+ * \endverbatim 
+ *
+ * \structure & \param value
+ *     tstart
+ *       |
+ *       ----persistentFailure // This is a permanent failure, try atomic way to do it instead
+ *       |
+ *       ----transientFailre // Temporary failure, user can retry
```
mstoodle: typo: Failre - Failure

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+ *       ----persistentFailure // This is a permanent failure, try atomic way to do it instead
+ *       |
+ *       ----transientFailre // Temporary failure, user can retry 
+ *       |
+ *       ----fallthrough // Success, use general(non-atomic) way to do it
+ */
+void
+IlBuilder::TransactionBegin(TR::IlBuilder **persistentFailureBuilder, TR::IlBuilder **transientFailureBuilder, TR::IlBuilder **fallThroughBuilder)
+    {   
+    //ILB_REPLAY("%s->TransactionBegin(%s, %s, %s);", REPLAY_BUILDER(this), REPLAY_BUILDER(persistentFailureBuilder), REPLAY_BUILDER(transientFailureBuilder), REPLAY_BUILDER(fallThroughBuilder));
```
mstoodle: you should do the createBuilderIfNeeded calls before you do the ILB_REPLAY line.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+ *       |
+ *       ----transientFailre // Temporary failure, user can retry 
+ *       |
+ *       ----fallthrough // Success, use general(non-atomic) way to do it
+ */
+void
+IlBuilder::TransactionBegin(TR::IlBuilder **persistentFailureBuilder, TR::IlBuilder **transientFailureBuilder, TR::IlBuilder **fallThroughBuilder)
+    {   
+    //ILB_REPLAY("%s->TransactionBegin(%s, %s, %s);", REPLAY_BUILDER(this), REPLAY_BUILDER(persistentFailureBuilder), REPLAY_BUILDER(transientFailureBuilder), REPLAY_BUILDER(fallThroughBuilder)); 
+    TraceIL("IlBuilder[ %p ]::transactionBegin %p, %p, %p, %p)\n", this, *persistentFailureBuilder, *transientFailureBuilder, *fallThroughBuilder);
```
mstoodle: You should also do this line after the createBuilderIfNeeded calls.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    cfg()->addEdge(_currentBlock, (*persistentFailureBuilder)->getEntry());
+    cfg()->addEdge(_currentBlock, (*transientFailureBuilder)->getEntry());
+    
+    AppendBuilder(*fallThroughBuilder);
+    gotoBlock(mergeBlock);
+
+    AppendBuilder(*persistentFailureBuilder);
+    gotoBlock(mergeBlock);
+    AppendBuilder(*transientFailureBuilder);
+
```
mstoodle: I think it would be better to lay out your "fallThroughBuilder" last so it can fall through to the mergeBlock. Following the tstart node, you should Goto the fallThroughBuilder (around the persistent and transient failure builders).

Why isn't the tend node automatically put at the end of the "fallThroughBuilder" ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,119 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    AppendBuilder(*transientFailureBuilder);
+    
+    appendBlock(mergeBlock);
+    }  
+
+/** 
+ * Generate XEND instruction at then end of the region
+ */
+void
+IlBuilder::TransactionEnd()
```
mstoodle: separating this from the tstart seems dangerous to me: what if the client forgets? you end up with weirdness.

compiler/ilgen/IlBuilder.hpp
```diff
@@ -198,6 +198,9 @@ class IlBuilder : public TR::IlInjector
    TR::IlValue *IndexAt(TR::IlType *dt, TR::IlValue *base, TR::IlValue *index);
    TR::IlValue *AtomicAddWithOffset(TR::IlValue *baseAddress, TR::IlValue *offset, TR::IlValue *value);
    TR::IlValue *AtomicAdd(TR::IlValue *baseAddress, TR::IlValue * value);
+   void TransactionBegin(TR::IlBuilder **persistentFailureBuilder, TR::IlBuilder **transientFailureBuilder, TR::IlBuilder **fallThroughBuilder);
```
mstoodle: I would prefer to have just Transaction() and TransactionAbort().

compiler/ilgen/IlBuilder.cpp
fredqiu: Agreed. I've re-arranged the sequence. Now I layout fallThrough as the last one and put a "tEnd" at the end of fallThrough automatically.

compiler/ilgen/IlBuilder.hpp
fredqiu: Yes Mark, now I moved what was in TransactionEnd() to the end of fallThrough IlBuilder and change the service name to Transaction().

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -371,6 +371,7 @@ class OMR_EXTENSIBLE CodeGenerator
    TR_HasRandomGenerator randomizer;
 
    virtual bool supportsAtomicAdd() {return false;}
+   virtual bool supportsTM()    {return false;}
```
0xdaryl: In general, because CodeGenerator is an extensible class, marking a member function as virtual isn't what you should use unless you really need to achieve dynamic polymorphism.  In your case, you shouldn't use it.

compiler/x/env/OMRCPU.cpp
fredqiu: Hi Andrew, I added another commit in my PR. I added another processor info for Broadwell and we assume only the version after broadwell supporting TM. (Even though a small number of Broadwell also have such problem, ignore it for now and I leave a TODO there. )

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -240,9 +240,21 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
...
+	  * Due to many verions of Haswell and a small number of Broadwell have defects for TM and then disabled by Intel, 
+	  * we will return false for any versions before Broadwell.
+	  *
+	  * TODO: Need to figure out from which mode of Broadwell start supporting TM
+	  */
+      if (!_targetProcessorInfo.isIntelHaswell())
+         {
+         if (TR::Compiler->target.is64Bit())
+            {
+            self()->setSupportsTM(); // disable tm on 32bits for now
```
mstoodle: why is it disabled on 32-bit "for now"? I can tell it's disabled from the code; the comment should explain why it's disabled for now.

compiler/codegen/OMRCodeGenerator.hpp
fredqiu: Hi Daryl, I make it virtual because I hope to disable TM for all platforms except X. (because only X has TM evaluator now.) I think I should change a name for supportsTM(), perhaps change to hasTMEvaluator(). Sorry it's very confusing similar names used for different checks. a) supportsTM() is used for checking if this platform has TM evaluator in code. Early this week we found some jitbuilder testcases failed on P and Z, so Mark asked me to add an assertion in the beginning of the jitbulder service. That's why now I have virtual supportsTM() returning false by default and supportsTM() in X/codgen is true b) getSupportsTM() is used for detecting CPUID to tell if the processor supports TM.

compiler/x/codegen/OMRCodeGenerator.cpp
fredqiu: For this, I simply copied from previous version. (github shows I added this line but I didn't) I didn't dare to remove them in case previous contributor has some concerns with 32-bits. Should I remove them since they don't have further explanation?

andrewcraik: Removing them due to a lack of an explanation is not a justification that the code will work in 32bit mode. I'd rather we keep the change to the control infrastructure here as minimal as possible since it impacts projects other than JITBuilder I believe.

fredqiu: Agreed. I will keep it this way.

jitbuilder/release/src/TransactionalOperations.cpp
```diff
@@ -0,0 +1,126 @@
...
+      cerr << "FAIL: compilation error " << rc << "\n";
+      exit(-2);
+      }
+
+   cout << "Step 4: invoke compiled code and print results\n";
+   typedef int32_t (TransactionalMethodFunction)(int32_t*);
+   TransactionalMethodFunction *transaction_increment = (TransactionalMethodFunction *) entry;
+
+   int32_t v;
+   v=0;   cout << "transaction_increment(" << v; cout << ") == " << transaction_increment(&v) << "\n";
```
mstoodle: strictly speaking, this test would be better if it were multi-threaded and if there were an example that would test the abort path. Perhaps you could open an issue to extend this test case to make it more realistic, so that we at least capture that this should be done?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,122 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    *persistentFailureBuilder = createBuilderIfNeeded(*persistentFailureBuilder);
+    *transientFailureBuilder = createBuilderIfNeeded(*transientFailureBuilder);
+    *transactionBuilder = createBuilderIfNeeded(*transactionBuilder);
+
+    TR::Node *persistentFailureNode = TR::Node::create(TR::branch, 0, (*persistentFailureBuilder)->getEntry()->getEntry());
+    TR::Node *transientFailureNode = TR::Node::create(TR::branch, 0, (*transientFailureBuilder)->getEntry()->getEntry());
+    TR::Node *transactionNode = TR::Node::create(TR::branch, 0, (*transactionBuilder)->getEntry()->getEntry());
+
+    TR::Node *tStartNode = TR::Node::create(TR::tstart, 3, persistentFailureNode, transientFailureNode, transactionNode);   
+    tStartNode->setSymbolReference(comp()->getSymRefTab()->findOrCreateTransactionEntrySymbolRef(comp()->getMethodSymbol()));
```
mstoodle: Forgive my not knowing the detailed semantics of the start node. Will the tstart node redirect control flow the transaction builder automatically if the builders are laid out this way? I'm just worried that the persistent failure builder appears as the fall through path from the tstart node.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1565,6 +1565,122 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    
+    genTreeTop(tStartNode);
+
+    //connecting the block having tstart with persistentFailure's and transaction's blocks 
+    cfg()->addEdge(_currentBlock, (*persistentFailureBuilder)->getEntry());
+    cfg()->addEdge(_currentBlock, (*transactionBuilder)->getEntry());
```
mstoodle: why isn't an edge needed to the transientFailureBuilder's entry?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4779,6 +4779,152 @@ TR::Register *OMR::X86::TreeEvaluator::atomicorEvaluator(TR::Node *node, TR::Cod
    return NULL;
    }
 
+TR::Register *                                                                                                                                                             
+OMR::X86::TreeEvaluator::tstartEvaluator(TR::Node *node, TR::CodeGenerator *cg)                                                                                             
+   {                                                                                                                                                                       
+   /*         
+   .Lstart:
+      xbegin       .Lfallback
+      jmp          .LfallThrough
```
mstoodle: not sure i like the layout of these instructions...seems like Lfallback should be placed somewhere "else" and it would be better to have the fallthrough label code directly below the xbegin (which would mean rearranging the builders laid down by the JitBuilder Transaction() service. Is this how these opcodes work on other platforms as well?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4779,6 +4779,152 @@ TR::Register *OMR::X86::TreeEvaluator::atomicorEvaluator(TR::Node *node, TR::Cod
...
+      persistentFailureLabel = generateLabelSymbol(cg);  
+      persistentFailureNode->getBranchDestination()->getNode()->setLabel(persistentFailureLabel);
+   }     
+       
+   TR::Register *accReg = cg->allocateRegister();                                                                                                                          
+   TR::RegisterDependencyConditions *endLabelConditions;
+   TR::RegisterDependencyConditions *fallThroughConditions = NULL;
+   TR::RegisterDependencyConditions *persistentConditions = NULL;
+   TR::RegisterDependencyConditions *transientConditions = NULL;
+
```
mstoodle: This area of the code is not my specialty. @andrewcraik have you or a delegate reviewed this code already?

compiler/x/codegen/OMRCodeGenerator.cpp
mstoodle: my remark was only about the fact that the comment itself is not informative. The code itself tells me that it is disabled for 32-bit; i don't need a comment to tell me that. What is warranted, IMO, is an explanation for why it is disabled on 32-bits. I wasn't suggesting we change the code at all.

If we have no additional information on why it's disabled for 32-bits, then I guess we can just leave it as-is.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -371,6 +371,7 @@ class OMR_EXTENSIBLE CodeGenerator
    TR_HasRandomGenerator randomizer;
 
    virtual bool supportsAtomicAdd() {return false;}
+   virtual bool hasTMEvaluator()    {return false;}
```
mstoodle: Daryl's earlier comment that these don't need to be virtual is right because CodeGenerator is an extensible class. For extensible classes, these queries will always be made against the most derived CodeGenerator class (that's what TR::CodeGenerator is). Thus, making them virtual provides no real value.

jitbuilder/release/src/TransactionalOperations.cpp
mstoodle: Don't remove it, but we should devise a test for it. See my other comment about opening an issue to extend the test case.

jitbuilder/release/src/TransactionalOperations.cpp
fredqiu: Agreed. I was wondering the same thing. Will open it now.

jitbuilder/release/Makefile
```diff
@@ -237,6 +237,11 @@ toiltype : libjitbuilder.a ToIlType.o
 ToIlType.o: src/ToIlType.cpp
 	g++ -o $@ $(CXXFLAGS) $<
 
+transactionaloperations : libjitbuilder.a TransactionalOperations.o
```
mstoodle: Please also add transactionaloperations to jitbuilder/release/.gitignore

mstoodle: And also, please add transactionaloperations to the "testall" goal.

compiler/ilgen/IlBuilder.cpp
fredqiu: oh, it's because I see there is cfg()-addEdge() in AppendBuilder() in next line. I want to avoid adding repetitive code.

compiler/ilgen/IlBuilder.cpp
mstoodle: ok, @fredqiu explained to me that tstart has no implicit fall-through control flow edge: it will always branch to one of its children, and there is no requirement (though it's customary) for any of those destinations to immediately follow the block that ends in tstart.

compiler/ilgen/IlBuilder.cpp
mstoodle: Since tstart does not actually fall-through, I would prefer you to add explicit control flow edges to all three blocks and then call appendNoFallThroughBlock(). It may feel strange, and it may end up doing the same thing, but I feel it more correctly models what's happening for the block containing that tstart node.

compiler/x/codegen/OMRTreeEvaluator.cpp
mstoodle: changing it would presumably affect things other than JitBuilder though, so let's leave it for now.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1619,6 +1619,124 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+    //ending the transaction at the end of transactionBuilder
+    appendBlock();
+    TR::Node *tEndNode=TR::Node::create(TR::tfinish,0);
+    tEndNode->setSymbolReference(comp()->getSymRefTab()->findOrCreateTransactionExitSymbolRef(comp()->getMethodSymbol()));
+    genTreeTop(tEndNode);
+
+	//Three IlBuilders above merged here
+    appendBlock(mergeBlock);
+
+	}
```
mstoodle: minor formatting error; can be fixed independently

mstoodle:
lgtm, other than a small formatting error. since I know you're off for the holidays I'll fix it myself once I merge this change (once the CI testing completes :) ).

Thanks, @fredqiu !

compiler/x/codegen/OMRTreeEvaluator.cpp
ymanton: The ideal way to do this is to have the fallback path be an out-of-line code section, where in you figure out why the transaction was aborted and branch to either the persistent or transient failure path. The `tstart` evaluator can then check whether the block representing the body of the transaction follows the current block or lives somewhere else and needs to be branched to.

PR #613, opened 2016-12-19
Resolve issue where the operand stack commits to the wrong address

charliegracie:
When the VirtualMachineOperandStack *copy* constructor is invoked it
does not copy the _stackOffset variable.  This causes commits to the
actual stack to happen at the incorrect location. The solution is to
copy the value of _stackOffset to the new VirtualMachineOperandStack


mstoodle:
Good catch @charliegracie ! I'll merge once the ci tests complete.

PR #614, opened 2016-12-19
Support for aborted Concurrent Scavenger during root processing

amicic:
If aborted Scavenge occurs during root processing (rare, but still
possible if a very large object is directly referenced from the roots)
proceed with the rest of the steps in the cycle while still in STW. Currently,
we would resume the app and concurrent scavenge operation, which is
dangerous since not all roots might not properly updated.
The code  changes are contained with CS specific state machine. 

We will now be able to report all the necessary events, and it will be
as clear in verbose gc that aborted cycle occurred, is if it were the
more common case (during main scan phase).

Also, piggybacking some general (not only CS specific) cleanup in
Scavenger - SemiSpace interaction relevant to the cycle
setup/teardown/abort.

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -383,6 +383,7 @@ MM_Scavenger::masterSetupForGC(MM_EnvironmentStandard *env)
 
 	/* Find tenure memory sub spaces for collection ( allocate and survivor are context specific) */
 	/* Find the allocate, survivor and tenure memory sub spaces for collection */
+	/* Evacuate range for GC is what allocate space is for allocation */
 	_evacuateMemorySubSpace = _activeSubSpace->getMemorySubSpaceAllocate();
 	_survivorMemorySubSpace = _activeSubSpace->getMemorySubSpaceSurvivor();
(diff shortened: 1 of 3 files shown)


gc/base/standard/Scavenger.cpp
```diff
@@ -3343,25 +3331,10 @@ MM_Scavenger::masterThreadGarbageCollect(MM_EnvironmentBase *envBase, MM_Allocat
...
-		}
-		scavengeIncremental(env, 3);
-		if (concurrent_state_scan == _concurrentState) {
-			_activeSubSpace->flip(env, MM_MemorySubSpaceSemiSpace::set_evacuate);
-			_activeSubSpace->flip(env, MM_MemorySubSpaceSemiSpace::set_allocate);
-
-		}
-	}
-	else
+	scavengeIncremental(env, 3);
```
charliegracie: What does 3 represent?

amicic: 3ms default increment duration, if/when incremental mode is implemented. It's just a remnant of other Garbage Collectos that rely on MasterGCThread infrastructures, and it's ignored for now. That and some other code within scavengeIncremental method is subject to removal or will evolve in something more meaningful, time permitting.

PR #616, opened 2016-12-19
Remove atomicOperations test case and add assertion in service

fredqiu:
1) Remove atomicOperations.cpp/hpp from main goal of JitBuilder Makefile.
2) Add Assertion in AtomicAddWithOffset() to ensure that the platform supports atomic add. Now we only support it in X.

Signed-off-by: fredqiu <fredqiu@ca.ibm.com>


mstoodle:
We need to figure out a story for supporting platform specific features in JitBuilder. I'll open an issue for the longer term solution to this problem.

PR #617, opened 2016-12-19
WIP: A temporary fix for AOT compile VarHandle methods

liqunl:
VarHandle optimization creates calls to out-of-thin-air, leaving some
symbols with cp index -1, which is not relocatable in AOT. This change
will catch the AOT exception thrown from VarHandle optimization and
abort the compile if the method is the outer most method, or set an
ilgen flag to tell inliner the method is not inlinable so that the
outer most method can still be AOT compiled.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -32,6 +32,9 @@
 #include "compile/ResolvedMethod.hpp"
 #include "compile/SymbolReferenceTable.hpp"
 #include "control/Options.hpp"
+#ifdef J9_PROJECT_SPECIFIC
+#include "exceptions/AOTFailure.hpp"
+#endif
 #include "control/Options_inlines.hpp"
 #include "env/TRMemory.hpp"
 #include "env/jittypes.h"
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik Andrew, can you review my code?

liqunl:
@andrewcraik Should we wrap the exception handling in frontend? Or it's ok to do it this way as a temporary fix?

liqunl:
I have tried catching the exception to a TR::CompilationException variable , and then casting it down to AOTHasInvokeVarHandle with dynamic_cast in J9 frontend, but dynamic_cast is not permitted with -fno-rtti. We need to throw subclass exceptions so that the control logic gets the right type and sets the right flag. Maybe we can compare the description message of the exception.

andrewcraik:
I'm not a fan of the Java specific nature of this code given this is OMR. I think the immediate issue has been worked around that this was addressing. The fundamental issue here is that there are two classes of compilation exceptions: class one are exceptions which when they occur all compilation must abort (out of memory etc) and class two means that the current ILGen/ilopts should be aborted but the outer compilation can continue.

Rather than putting in this temporary fix, I would rather get the classification above codified and then make sure the catch uses whatever the infrastructure we build is to test if the exception is a class 2 exception. If it is then we fail ilgen, if it is a class one exception we just rethrow it bring down the compilation.

@dsouzai @lmaisons I think this classification is a design you need to be involved with so let's iterate on what Liqun has crafted here and generalize things so we can differentiate class one and class two exceptions and do the right thing. This is an important and pressing item so we should aim to get this done in the next few days. We may wish to consider at what point to merge it given the changes in behaviour it will bring, a small delay between ready to merge and merge may be needed so the project is at a good point for the change to land.

0xdaryl:
I agree with @andrewcraik 's comments.  In general, we're trying to avoid introducing "temporary fixes" or "todos" to the code without clear justification and an issue created to track making them "permanent".

liqunl:
I played with the c++ rethrow feature and looks like we don't need to dynamic cast the exception. Will update the PR with the new fix.

liqunl:
A new pull request has been created to better describe the new fix. https://github.com/eclipse/omr/pull/631
Close this PR.

PR #619, opened 2016-12-20
Miscellaneous JitBuilder fixes

mstoodle:
Collection of minor improvements and fixes for JitBuilder services

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -240,6 +240,7 @@ IlBuilder::defineValue(const char *name, TR::IlType *type)
    {
    TR::DataType dt = type->getPrimitiveType();
    TR::SymbolReference *newSymRef = symRefTab()->createTemporary(methodSymbol(), dt);
+   newSymRef->getSymbol()->setNotCollected();
    defineSymbol(name, newSymRef);
    }
 
@@ -247,6 +248,7 @@ TR::IlValue *
 IlBuilder::newValue(TR::DataType dt)
    {
    TR::SymbolReference *newSymRef = symRefTab()->createTemporary(methodSymbol(), dt);
+   newSymRef->getSymbol()->setNotCollected();
    char *name = (char *) _comp->trMemory()->allocateHeapMemory(5 * sizeof(char));
    sprintf(name, "_T%-3d", newSymRef->getCPIndex());
    newSymRef->getSymbol()->getAutoSymbol()->setName(name);
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1379,9 +1379,23 @@ IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
    return returnValue;
    }
 
+void
+IlBuilder::integerizeAddresses(TR::IlValue **leftPtr, TR::IlValue **rightPtr)
+   {
+   TR::IlValue *left = *leftPtr;
+   TR::IlValue *right = *rightPtr;
+   if (left->getSymbol()->getDataType() == TR::Address)
+      {
+      TR_ASSERT(right->getSymbol()->getDataType() == TR::Address, "expecting both operands to be addresses");
+      *leftPtr = ConvertTo(Word, left);
+      *rightPtr = ConvertTo(Word, right);
+      }
+   }
+
 TR::IlValue *
 IlBuilder::LessThan(TR::IlValue *left, TR::IlValue *right)
    {
+   integerizeAddresses(&left, &right);
    TR::IlValue *returnValue=compareOp(TR_cmpLT, false, left, right);
    TraceIL("IlBuilder[ %p ]::%d is LessThan %d < %d?\n", this, returnValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
    ILB_REPLAY("%s = %s->LessThan(%s, %s);", REPLAY_VALUE(returnValue), REPLAY_BUILDER(this), REPLAY_VALUE(left), REPLAY_VALUE(right));
@@ -1391,6 +1405,7 @@ IlBuilder::LessThan(TR::IlValue *left, TR::IlValue *right)
 TR::IlValue *
 IlBuilder::LessOrEqualTo(TR::IlValue *left, TR::IlValue *right)
    {
+   integerizeAddresses(&left, &right);
    TR::IlValue *returnValue=compareOp(TR_cmpLE, false, left, right);
    TraceIL("IlBuilder[ %p ]::%d is LessOrEqualTo %d <= %d?\n", this, returnValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
    ILB_REPLAY("%s = %s->LessOrEqualTo(%s, %s);", REPLAY_VALUE(returnValue), REPLAY_BUILDER(this), REPLAY_VALUE(left), REPLAY_VALUE(right));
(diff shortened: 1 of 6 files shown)


mstoodle:
Two things are missing from the current set of commits: documentation needs to be added for the various conditional services, and tests should also be added. Marking this pull request as WIP while I finish those pieces off.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1366,7 +1378,7 @@ TR::IlValue *
 IlBuilder::UnsignedShiftR(TR::IlValue *v, TR::IlValue *amount)
    {
    TR::IlValue *returnValue=binaryOpFromOpMap(TR::ILOpCode::unsignedShiftRightOpCode, v, amount);
-   TraceIL("IlBuilder[ %p ]::%d is shr %d << %d\n", this, returnValue->getCPIndex(), v->getCPIndex(), amount->getCPIndex());
+   TraceIL("IlBuilder[ %p ]::%d is unsigned shr %d >>> %d\n", this, returnValue->getCPIndex(), v->getCPIndex(), amount->getCPIndex());
```
Leonardo2718: Is `>>>` supposed to be `>>` or do you use this to signify that it's unsigned?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1380,24 +1392,59 @@ IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
...
    }
 
+void
+IlBuilder::integerizeAddresses(TR::IlValue **leftPtr, TR::IlValue **rightPtr)
+   {
+   TR::IlValue *left = *leftPtr;
+   TR::IlValue *right = *rightPtr;
+   if (left->getSymbol()->getDataType() == TR::Address)
+      {
+      TR_ASSERT(right->getSymbol()->getDataType() == TR::Address, "mismatched types in address compare");
```
Leonardo2718: I see how this function would mostly be useful when comparing address, but assuming that this function will only ever be called for comparisons doesn't seem quite right. Maybe the assert message should say something like "both types are not addresses", just to make it clear that it failed in this function as opposed to the actual comparison function.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1380,24 +1392,59 @@ IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
...
    ILB_REPLAY("%s = %s->LessThan(%s, %s);", REPLAY_VALUE(returnValue), REPLAY_BUILDER(this), REPLAY_VALUE(left), REPLAY_VALUE(right));
    return returnValue;
    }
 
 TR::IlValue *
+IlBuilder::LessOrEqualTo(TR::IlValue *left, TR::IlValue *right)
+   {
+   integerizeAddresses(&left, &right);
+   TR::IlValue *returnValue=compareOp(TR_cmpLE, false, left, right);
+   TraceIL("IlBuilder[ %p ]::%d is LessOrEqualTo %d < %d?\n", this, returnValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
```
Leonardo2718: Should the message use `<=` instead of just `<` or is this symbol reserved for something else?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1380,24 +1392,59 @@ IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
...
+   ILB_REPLAY("%s = %s->LessOrEqualTo(%s, %s);", REPLAY_VALUE(returnValue), REPLAY_BUILDER(this), REPLAY_VALUE(left), REPLAY_VALUE(right));
+   return returnValue;
+   }
+
+TR::IlValue *
 IlBuilder::GreaterThan(TR::IlValue *left, TR::IlValue *right)
    {
+   integerizeAddresses(&left, &right);
    TR::IlValue *returnValue=compareOp(TR_cmpGT, false, left, right);
    TraceIL("IlBuilder[ %p ]::%d is GreaterThan %d < %d?\n", this, returnValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
```
Leonardo2718: Should the `<` be a `>`?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1380,24 +1392,59 @@ IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
...
    ILB_REPLAY("%s = %s->GreaterThan(%s, %s);", REPLAY_VALUE(returnValue), REPLAY_BUILDER(this), REPLAY_VALUE(left), REPLAY_VALUE(right));
    return returnValue;
    }
 
+TR::IlValue *
+IlBuilder::GreaterOrEqualTo(TR::IlValue *left, TR::IlValue *right)
+   {
+   integerizeAddresses(&left, &right);
+   TR::IlValue *returnValue=compareOp(TR_cmpGE, false, left, right);
+   TraceIL("IlBuilder[ %p ]::%d is GreaterOrEqualTo %d < %d?\n", this, returnValue->getCPIndex(), left->getCPIndex(), right->getCPIndex());
```
Leonardo2718: Should the `<` be a `>=`?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1645,13 +1699,29 @@ IlBuilder::IfCmpLessThan(TR::IlBuilder *target, TR::IlValue *left, TR::IlValue *
    }
 
 void
+IlBuilder::IfCmpLessOrEqual(TR::IlBuilder *target, TR::IlValue *left, TR::IlValue *right)
+   {
+   TR_ASSERT(target != NULL, "This IfCmpLessOrEqual requires a non-NULL builder object");
+   ILB_REPLAY("%s->IfCmpLessOrEqual(%s, %s, %s);", REPLAY_BUILDER(this), REPLAY_BUILDER(target), REPLAY_VALUE(left), REPLAY_VALUE(right));
+   TraceIL("IlBuilder[ %p ]::IfCmpLessOrEqual %d < %d? -> [ %p ] B%d\n", this, left->getCPIndex(), right->getCPIndex(), target, target->getEntry()->getNumber());
```
Leonardo2718: Should the `<` be a `<=`?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1660,6 +1730,14 @@ IlBuilder::IfCmpGreaterThan(TR::IlBuilder *target, TR::IlValue *left, TR::IlValu
    }
 
 void
+IlBuilder::IfCmpGreaterOrEqual(TR::IlBuilder *target, TR::IlValue *left, TR::IlValue *right)
+   {
+   ILB_REPLAY("%s->IfCmpGreaterOrEqual(%s, %s, %s);", REPLAY_BUILDER(this), REPLAY_BUILDER(target), REPLAY_VALUE(left), REPLAY_VALUE(right));
+   TraceIL("IlBuilder[ %p ]::IfCmpGreatOrEqual %d < %d? -> [ %p ] B%d\n", this, left->getCPIndex(), right->getCPIndex(), target, target->getEntry()->getNumber());
```
Leonardo2718: Should the `<` be a `>=`?

mstoodle:
Thanks for the review, @Leonardo2718 . Clearly I missed a lot of copy&paste bugs, which I will clean up.

mstoodle:
I changed my mind about documenting the various conditional services: I feel they are so obvious that introducing 5+ lines of comments per service adds too much clutter to the file. I'm willing to be convinced otherwise on this point, but for this PR, I think I'll leave them as-is (which means they fit in very well with the 50 or so other services that also aren't documented around them).

I also want to form a more considered opinion about how and where we should be testing such basic services, so I'm also going to defer writing tests for the conditional services in this particular PR. I have opened #627 for the community to discuss these issues before moving forward on any approach.

PR #621, opened 2016-12-20
Shutdown issues

charliegracie:
When using the OMR_Initialize_VM and OMR_Shutdown_VM APIs you can get a crash in the flushing TLHs and there was memory being leaked.

The 2 commits in this PR resolve both of those issues.


charliegracie:
@jason-hall can you verify that this change resolves your issues?  Thanks

jason-hall:
Yes. I can confirm the issue I observed does not occur with these changes.

charliegracie:
@jason-hall thank you.

PR #622, opened 2016-12-20
Adding new option to change number of profiled classes

r30shah:
In all cases we used to set number of profiled classes for profiledClassTest
using preprocessor directive #define. This commit will allow us to change the
number of inlined profiled class using option.


andrewcraik:
I'm not sure the name of inlined profiled classes makes sense since it sounds like it belongs with the inliner while I think this is probably related to some kind of codegen cache. Perhaps a name without inline in it would make more sense.

ymanton:
This is the number of profiled class tests to emit for `instanceof` and `checkcast` so the name should probably reflect that, e.g. `getInstanceOfOrCheckCastMaxProfiledClassTests()` (or shorter if you can think of something). `getMaxProfiledClasses()` is way too generic for anyone to know what it refers to without digging around to see how/where it's used.

ymanton:
This also raises the question of why this is here in OMR rather than whichever downstream project actually uses it. Did you look into that already?

r30shah:
@ymanton  Yes you are right. My mistake in skipping that. Thanks for pointing that out. Closing this Pull Request as I can add this to the appropriate downstream project.

PR #626, opened 2016-12-21
Special handling for OSRCatchBlock for IndVarConversion

cathyzhyi:
Bail out IndVarConversion if there is any exit edge
from inside the loop to an OSRCatchBlock. This is because
edge splitting is needed on every exit to shrink the induction
variable back but it is not safe to split edge to an OSRCatchBlock.

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1248,6 +1248,7 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
    {
 
    //traceMsg("Splitting edge (%d,%d)\n", from->getNumber(), to->getNumber());
+   TR_ASSERT(!to->isOSRCatchBlock(), "Splitting edge to OSRCatchBlock (block_%d -> block_%d) is not supported\n", from->getNumber(), to->getNumber());
    TR::Node *exitNode = from->getExit()->getNode();
    TR_Structure *fromContainingLoop = NULL;
    if (from->getStructureOf())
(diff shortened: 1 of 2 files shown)


compiler/optimizer/InductionVariable.cpp
```diff
@@ -4223,9 +4223,24 @@ void TR_LoopStrider::detectLoopsForIndVarConversion(
...
+      //in fsd, bail out if there is any exception edges to OSRCatchBlock because there is no
+      //safe way to split those edges
+      if (TR::comp()->getOption(TR_FullSpeedDebug) && nextBlock->hasExceptionSuccessors())
+         {
+         TR_SuccessorIterator sit(nextBlock);
+         for (TR::CFGEdge *e = sit.getFirst(); e != NULL; e = sit.getNext())
+            {
+            TR::Block *dest = toBlock(e->getTo());
+            if (dest->isOSRCatchBlock())
+               return;
```
andrewcraik: This needs trace output for sure - silent aborts are never good!

compiler/il/OMRBlock.cpp
```diff
@@ -1336,21 +1336,13 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
...
-               to->getOwningMethod(),
-               c);
-            TR::ResolvedMethodSymbol *method = c->getMethodSymbol();
+         newBlock->setHandlerInfo(
+            to->getCatchType(),
+            to->getInlineDepth(),
+            to->getHandlerIndex(),
+            to->getOwningMethod(),
+            c);
+         TR::ResolvedMethodSymbol *method = c->getMethodSymbol();
```
andrewcraik: This code should be guarded to make sure we only do this for exception edges going to non-OSR catch blocks. We should also do something nasty like return NULL if we fail to split so that it causes the code to fall over aggressively in the unsupported case.

andrewcraik:
I have asked for a couple of changes before this gets merged.

compiler/il/OMRBlock.cpp
cathyzhyi: Thanks, Andrew! I've added an assert to guard the OSRCatch block case. Other than that, the edge splitting should be able to split and insert a valid newBlock.

compiler/optimizer/InductionVariable.cpp
cathyzhyi: Agreed. Added new trace for debugging purpose.

andrewcraik:
@jdmpapin can you do a quick review since the induction variable conversion code is something you are very familiar with. Change looks good to me and should sort the issue which should only show under involuntary OSR modes (ie FSD).

jdmpapin:
Just as a note, the change looks at first glance slightly more conservative than it really is. It will skip any loop containing a block with an outgoing OSR exception edge, even when that edge is not a loop exit. But strider already skips any loop that contains a catch block, so there's no difference when the OSR catch block is within the loop.

LGTM

andrewcraik:
LGTM

compiler/optimizer/InductionVariable.cpp
```diff
@@ -4223,9 +4223,30 @@ void TR_LoopStrider::detectLoopsForIndVarConversion(
...
    for (nextBlock = bi.getCurrent(); nextBlock; nextBlock = bi.getNext())
+      {
       if (nextBlock->hasExceptionPredecessors())
          return;
 
+      //in fsd, bail out if there is any exception edges to OSRCatchBlock because there is no
+      //safe way to split those edges. Also, there is no need to check if the OSRCatchBlock is 
+      //within the loop because loop strider already skips any loop that contains a 
+      //catch block
+      if (TR::comp()->getOption(TR_FullSpeedDebug) && nextBlock->hasExceptionSuccessors())
```
vijaysun-omr: Is there any reason why FSD must be checked here ? 

It might be more general to skip this if OSR catch blocks and exception edges end up used in other cases.

andrewcraik: agreed - though at the point where strider runs the OSR transitions should be explicit and so removing the check shouldn't have a functional impact. Good catch Vijay - thanks!

cathyzhyi: agreed, it should be more general. Modified as suggested.

vijaysun-omr:
Thanks for making the change. I will merge once the checks have passed.

PR #631, opened 2016-12-22
Handle exceptions thrown in ilgen opts

liqunl:
optimizations can throw exceptions, including ilgen opts. Exceptions
thrown from non-ilgen opts will be catched by compilation object and
in most cases will cause a compilation to abort which is a desired
behavior. However, aborting the compilation during ilgen opts triggered
by inlining will stop jit from compiling a compilable method, leading
to potential performance issues. Thus, we should only abort the
compilation when not in inliner.

This commit introduces a recoverable ilgen exception class, who can be
inherited by project specific exceptions with potential to be recovered
in ilgen. genIL will catch the base class of the recoverable ilgen
exceptions and choose the right action.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1282,7 +1282,17 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
 
          if (optimizer)
             {
-            optimizer->optimize();
+            try
+               {
+               optimizer->optimize();
+               }
+            catch (const TR::RecoverableILGenException &e)
+               {
+               if (self()->comp()->isOutermostMethod())
+                  throw;
+               _methodFlags.set(IlGenSuccess, false);
+               }
+
             comp->setOptimizer(previousOptimizer);
             }
          else
--- a/compiler/compile/CompilationException.hpp
+++ b/compiler/compile/CompilationException.hpp
@@ -35,6 +35,11 @@ struct ILGenFailure : public virtual CompilationException
    virtual const char* what() const throw() { return "IL Gen Failure"; }
    };
 
+struct RecoverableILGenException : public virtual CompilationException
+   {
+   virtual const char* what() const throw() { return "Recoverable IL Gen Exception"; }
+   };
+
 struct ExcessiveComplexity : public virtual CompilationException
    {
    virtual const char* what() const throw() { return "Excessive Complexity"; }


liqunl:
@andrewcraik This is the new fix for the exception handling. The j9 front end will override handleExceptionThrownFromILGen and decide whether to rethrow the exception.

andrewcraik:
To me this seems like a reasonable change and provides the flexibility we need to handle non-fatal ilgen failures.

liqunl:
Will test the commit together with the j9 change, then ask for a merge.

lmaisons:
I think we need to introduce the concept of transient / local / potentially recoverable IL generation failures into our understanding of exceptions, and classify specific exceptions with this concept as appropriate.

I'm also not entirely convinced that genIL is the right place for the try / catch block.  Did we consider putting exception handling in genMethodILForPeeking / tryToGenerateILForMethod (or their underscore prefixed equivalents)?  With some (but not exhaustive) consideration those feel like they would have better context to discern the right action should IL gen encounter an exception.

andrewcraik:
I think we do need to introduce the concept and then leverage it properly in OMR. This infrastructure would be the first part of that since the classification of recoverable vs non-recoverable could be project specific.

My feeling on where to do the catch is at this base level of the ilgen. The simple reason being that IL gen can already fail for other reasons and the service wrappers like genMethodILForPeeking and tryToGenerateILForMethod can handle a failure coming from the core service that actually builds the trees which is the level we are at. This would seem to centralize the fact that we are going to say yes we can recover from this kind of error and then carry on. We would need to consider moving up if the error classification were different for peeking and inlining etc, but I am hoping that is a degree of freedom that we don't actually need or that we should delay introducing until such times as we have a compelling use case for it.

dsouzai:
I think this approach is fine. However, handleExceptionThrownFromILGen should be carefully designed since the aim is to continue compiling. It should not do anything that could result in a different exception being thrown; for example, scratch memory (heap or stack) should **not** be allocated because an std::bad_alloc could be thrown resulting in std::terminate. In addition, care needs to be taken to make sure that the state of the compilation is valid after handleExceptionThrownFromILGen has returned success (eg no dangling references, no locks acquired, etc). 

Ideally, it would be nice to have a special class of exceptions along the lines of "Recoverable IL Gen Exception". The benefit of this would be that handleExceptionThrownFromILGen would be simplified since the catch block wouldn't need to catch the high level TR::CompilationException; any non recoverable IL gen exception would escape all the way to the top level catch block in the control code, and any recoverable IL gen exception that somehow managed to escape to the top level catch block could have a TR_ASSERT in its catch block.

liqunl:
@andrewcraik Andrew, can you review the commit again so that we can merge it?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1260,7 +1260,15 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
...
             {
-            optimizer->optimize();
+            try
+               {
+               optimizer->optimize();
+               }
+            catch (const TR::CompilationException &e)
+               {
+               bool success = comp->fe()->handleExceptionThrownFromILGen();
+               _methodFlags.set(IlGenSuccess, success);
```
andrewcraik: Since the _methodFlags are really for cross-compilation communication does this do the right thing for peeking? I would assume the flag would be reset by the call to do a peek, but can you confirm this is the case?

compiler/codegen/FrontEnd.hpp
```diff
@@ -125,6 +125,7 @@ class TR_FrontEnd : public TR_Uncopyable
    // Method
    // --------------------------------------------------------------------------
 
+   virtual bool handleExceptionThrownFromILGen();   // This API is to be used in a catch block
```
andrewcraik: Should we do the rethrow in the catch block and return an enum saying yes, no, or rethrow? Thoughts?

andrewcraik:
So the comment for @dsouzai is valid, but for the time being we are not adding exception throws that will result in compilation success, what we want to do is fail the ilgen for the compilation so as long as you are not throwing with locks held or other craziness then we should be fine. A comment warning of the dangers and that the usual practice should be to return false whenever possible should be added since this mechanism is for truly exceptional cases

I remain firmly of the opinion that the try catch should be done in genIL since genIL is allowed to fail and callers already need to handle this fact. We need to make sure that the failure is properly handled by callers, but I don't think that someone calling a service wrapper cares why the failure happened. I think keeping the handling in genIL simplifies things and maintains the strong encapsulation genIL currently achieves. I don't want exception handling logic spreading into the optimizer so genIL is a nice logical spot for now and we can reconsider if there is a case that it doesn't handle well.

How are we going to handle the classification of exceptions into those where we can fail ILgen and those where we need to rethrow? That is the real benefit we are trying to introduce here - is that follow-on work?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
liqunl: Peeking does not check this flag, nor does inlining. Confirmed with Marius, no one cares about this flag. Maybe we should clean up this flag.

andrewcraik: ok well leave that as it is for now - the clean-up can wait.

compiler/codegen/FrontEnd.hpp
liqunl: If we plan to rethrow in the catch block instead of in the frontend, we can not try catch the exception in the frontend, thus we will not be able get the exception details. If we want to treat all the exceptions the same, then yes, we can make the frontend return an enum. @andrewcraik What do you think?

liqunl:
Re-append WIP to this PR because we want to add the capability to tell the type of the exception in runtime with RTTI off. And we want to avoid doing exception catch or rethrow in the handler, just to keep the implementation clean.

liqunl:
i just recalled that callers of peeking does not check the return value nor the method flag, instead, they check if there are trees generated for the method, if so, they consider the peeking is successful, which is wrong when ilgen can fail during ilgen opts. So this PR should also add a fix to callers of peeking.

liqunl:
@andrewcraik @dsouzai I have added an exception kind enum to each exceptions, can you take a look? We might need better naming for TR_ILGenState and its enums, suggestions are welcomed

dsouzai:
Why do we need the `getKind` method? The exception type already gives information on what the exception is. If you look at `compileMethodFromDetails` in CompileMethod.cpp, it does something similar to this:

```
catch (const TR::CompilationException &exception)
   {
   try
      {
      throw;
      }
   catch (const SpecificExceptionType1 &e)
      {
      // do something
      }
   }
```

You can do the same in the `handleExceptionThrownFromILGen` method. This would remove the need for adding a redundant getKind method in each of the exception objects. In addition, you'll notice in `compileMethodFromDetails` that there's this line:

```
#if defined(J9ZOS390)
         // Compiling with -Wc,lp64 results in a crash on z/OS when trying
         // to call the what() virtual method of the exception.
         printCompFailureInfo(jitConfig, &compiler, "");
#else
         printCompFailureInfo(jitConfig, &compiler, exception.what());
#endif
```

There's a bug in the V2R2 XLC for z/OS that prevents us from actually reading the exception object; it seems to be ok to throw and catch it but if you want to access anything in it, it causes a crash. This is another reason (albeit not a very good reason) to not use getKind.

Also worth getting @lmaisons to give this a once-over

andrewcraik:
The suggestion for the getKind came from me - the thing I was trying to avoid was having the mechanism for error signalling (currently exceptions, previously longjump) being in the policy processing logic about whether we need to abort the compile or just fail ILgen to make future changes easier. If the policy logic does the rethrow to a specific catch then the fact we use exceptions will end up baked into every policy implementation and I was trying to avoid that to make it easier to change in future if required.

I was not aware of the issue on zOS which complicates things, but rethrowing to recatch doesn't seem like a clean or nice design to me.

andrewcraik:
If reading the exception object is not possible on all platforms this has to support would it be better to have a static getKind which does the rethrow to catch to return a kind enum so that you can get the 'kind' and we can centralize that idiom to one place for removal once the exception object can be read?

dsouzai:
I guess it comes down to what easier means. In this PR, genIL already has a try/catch so some exception-ism will be baked in there. As long as we use exceptions, it will only make sense to use getKind inside a catch block. As such, as long as we use high level functions like `handleExceptionThrownFromILGen` which could be renamed to `handleILGenError` then there's still really only one place that we would need to change if we moved from exceptions to the next error handling scheme. However, if in the future we have various handle<misc>Error functions, then having each of them do a rethrow and recatch is unnecessary. In that case, having a static getKind would make sense. 

I do have a couple of thoughts with using a getKind approach:

1. It should not prevent an exception from being rethrown. Namely, we'd need to make sure that in this scenario:
```
bool checkException()
   {
   bool shouldThrow = true;
   try
      {
      throw;
      }
   catch (mySpecificException &e)
      {
      shouldThrow = false;
      }
   }`
...
catch (myException &exception)
   {
   if (checkException())
      throw;
   }
...
```
if `checkException()` returns true, it should throw the same exception (and not cause std::terminate instead).

2. We would need to figure out how to define getKind for project specific code since there will be (and are) project specific exceptions defined.

liqunl:
If the xlc bug is going to be fixed, we can use this implementation for non-zos platforms, and for zos, we can use the static getKind, once the bug is fixed, we can remove the static getKind.

As for the static getKind, we need to have in omr, and one in each project, below can be an example for j9 getKind
TR::ExceptionKind getKind()
   {
   if (TR::TR_UnknownException != getOMRKind())
      return getOMRKind();
   return getJ9Kind();
   }

TR::ExceptionKind getOMRKind()
   {
   try
      {
      throw;
      }
   catch (const TR::ILGenFailure &e)
      {
      return TR::TR_ILGenFailure;
      }
   /*
    * catch other omr exceptions here
   */
   catch (...)
      {
      return TR::TR_UnknownException;
      }
    }

similar to getOMRKind(), getJ9Kind will only try to catch J9 specific exceptions.

dsouzai:
I'd rather we didn't have two different implementations. I prefer the static getKind method over the virtual one because 1) it does not require two implementations, and 2) because in the control code logic, we can remove the rethrow/recatch and instead call getKind.

liqunl:
With the discussion with Andrew, Irwin and Luc, we have reached an agreement on the following design:
We will have one new exception called RecoverableILGenException inherited from CompilationException in OMR, and AOTHasInvokeVarHandle would inherit from RecoverableILGenException. We will catch it in ilgen and check if we are in the outer most method, if so, return false, otherwise rethrow to abort the compile. The code would look like

try
   {
   optimizer->optimize();
   }
catch (const TR::RecoverableILGenException &e)
   {
   if (self()->comp()->isOutermostMethod())
       throw;
    _methodFlags.set(IlGenSuccess, false);
    }

liqunl:
Before the fix in ilgen is delivered, changes are needed for callers of genMethodILForPeeking. They tell whether ilgen is successful by looking at the method symbol's trees, if trees exist for the method symbol, they think ilgen has succeeded. This is not the case with this fix since ilgen can fail in ilgen opts now, when the trees have been generated for the method in walker.

liqunl:
I will not be able to work on this PR within 2 weeks, if someone else is to take over the fix, please read the 2 comments above.

liqunl:
@andrewcraik can you review the change? It has passed the personal build.

andrewcraik:
I think this looks good to merge.

vijaysun-omr:
I can merge this once checks pass.

liqunl:
There seems to be a network issue, I've pasted the link to the failed job in trtalk, hope the infra team can get it sort out soon.

liqunl:
Rebased the commit and hopefully the ci build will pass this time.

liqunl:
@0xdaryl is the TravisCI network issue fixed? My ci builds seem to be waiting since 10am this morning.

mgaudet:
@liqunl https://www.traviscistatus.com/ :smile:

liqunl:
@mgaudet Thanks Matthew, so to confirm, is it caused by Heroku outage? If so, do you know when it will be back?

mgaudet:
Not the outage; it's the backlogs that are the issue.

andrewcraik:
@vijaysun-omr can we get this merged now?

vijaysun-omr:
Merging since checks have passed now.

PR #633, opened 2016-12-22
Fix disableLongDispStackSlot option on Z

doubletea:
`GPR7` and `HPR7` are locked when the long displacement stack slot is disabled, however they are still added to the Global Register table for GRA. This change implements the missing conditions in  `initializeGlobalRegisterTable()` to fix this problem.

The binary encoding for `LG, LGF, and LLGF` assumes the source and target registers are the same `scratchReg` taken from `getRegisterOperand(1)`. However, this is not true when the `scratchReg` is taken from `ExtCodeBaseRealRegister`, where the source and target are different. This is fixed by correctly differentiating the `scratchReg` and `sourceReg` locals.

Signed-off-by: John Xu <xujohn@ca.ibm.com>


fjeremic:
LGTM.

Issue #632, opened 2016-12-22
ILGen opts do not run in noOpt


liqunl:
ILGen opts are fundamental optimizations and should run at all optLevels. Currently, it's disabled for noOpt which is wrong and can cause functional problems.

mgaudet:
So this can be closed now right?

PR #634, opened 2016-12-22
Run ilgen opts at all optLevels

liqunl:
ilgen opts are fundamental opts that modify the tree to a good shape
and if are disabled, funational problems will happen. Currently, ilgen
opt strategy is disabled in noopt. With this fix, it will run at all
optLevels.

issue: #632


liqunl:
Personal build have passed. @andrewcraik Can you review this commit?

andrewcraik:
Looks good to me and I think it makes sense to run ilgen opts at all times. This is one for @vijaysun-omr to review when he gets time.

liqunl:
@vijaysun-omr Vijay, I have rebased the branch and retested it with a personal build, can you merge it if you're ok with the change?

vijaysun-omr:
No issues with this; we were anyway running IL gen opts at >= cold method hotness and so it should be impacting only no-opt, and I think affecting that hotness in this way is fine.

PR #444, opened 2016-11-07
Change configure flags for JIT components

noryb009:
Add separate configure flags for the Test compiler and JitBuilder; `--enable-TEST_COMPILER` and `--enable-JITBUILDER` respectively. These were picked to be consistent with other configure flags.

Issue: #285

PR #635, opened 2016-12-22
Fix two build issues

noryb009:
First, JitBuilder used to specify `TEST_COMPILER_SPECIFIC`, but not `JITBUILDER_SPECIFIC`. Since neither is used in the code, we can safely change this define.

Second, the Test compiler and JitBuilder builds on OS X were accidentally disabled in #444. This re-enables them by default. No other instances of this problem exist in other configurations.


mstoodle:
lgtm, thanks @noryb009 !

0xdaryl:
I would prefer the macro to be `JITBUILDER_PROJECT_SPECIFIC` to be consistent with macros used in other projects.  Since this is already merged I will create a separate PR for that.

Issue #500, opened 2016-11-16
Impove error checking and handling in JitBuilder


Leonardo2718:
Currently, error handling in JitBuilder is rather light. Specifically, there are several instances where there are no checks being performed for things that really aught to be checked. For example, the definition of `TypeDictionary::CloseStruct` does not check whether the struct being closed has been created:

```c++
void
TypeDictionary::CloseStruct(const char *structName, unsigned int finalSize)
   {
   TR_HashId structID = 0;
   _structsByName->locate(structName, structID);
      // what if there is not `structName` in `_structByName` ?
   StructType *theStruct = (StructType *) _structsByName->getData(structID);
   theStruct->Close(finalSize);
   }
```

Leonardo2718:
When creating a new struct or union, no check is done to see if a type with the same name already exists. Create two types with the same name should be disallowed (likely using an assert) as it causes problems when looking up a type in `TypeDictionary`.

A related issue is that when create a struct or union field, no check is done to see if a field with the same name already exists in the type. This needs to be disallowed for the same reason as above.

PR #644, opened 2016-12-28
Add JitBuilder support for unions

### LINKED ISSUE #643: JitBuilder should support unions
JitBuilder does not currently have an interface for dealing with unions. This can make interfacing with a VM that makes heavy use of unions unnecessarily difficult or cumbersome. For this reason, JitBuilder should have an interface for defining and using unions. For consistency, this interface should be similar to the one provided for structs.

As a note, it is *very* important that JitBuilder be able to alias all union members together to prevent incorrect optimizations.

Leonardo2718:
This implements JitBuilder support for union types. Some minor code clean ups are also done.

The implementation is split into two commits. One commit implements the underlying structures used to represent unions. The other adds the interface methods to `TypeDictionary` used to create and interact with union types. The remaining commits do some light code clean up and add testing for JitBuilder unions.

Fixes #643

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1232,7 +1232,9 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
 
       TR_IlGenerator *ilGen = customRequest.getIlGenerator(self(), fe, comp, symRefTab);
 
-      _methodFlags.set(IlGenSuccess, ilGen->genIL());
+      auto genIL_rc = ilGen->genIL();
+      _methodFlags.set(IlGenSuccess, genIL_rc);
+      traceMsg(self()->comp(), "genIL() returned %d\n", genIL_rc);
 
       if (_methodFlags.testAny(IlGenSuccess))
          {
--- a/compiler/ilgen/TypeDictionary.cpp
+++ b/compiler/ilgen/TypeDictionary.cpp
@@ -371,8 +371,7 @@ TR::IlReference *
 UnionType::getFieldSymRef(const char *fieldName)
    {
    FieldInfo *info = findField(fieldName);
-   if (NULL == info)
-      return NULL;
+   TR_ASSERT(info, "Struct %s has no field with name %s\n", getName(), fieldName);
 
    TR::SymbolReference *symRef = info->getSymRef();
    if (NULL == symRef)
--- a/compiler/ilgen/TypeDictionary.cpp
+++ b/compiler/ilgen/TypeDictionary.cpp
@@ -28,6 +28,7 @@
 #include "env/Region.hpp"
 #include "env/SystemSegmentProvider.hpp"
 #include "env/TRMemory.hpp"
+#include "infra/BitVector.hpp"
 
 
 namespace OMR
@@ -285,6 +286,121 @@ StructType::getFieldSymRef(const char *fieldName)
    return (TR::IlReference *)symRef;
    }
 
+class UnionType : public TR::IlType
+   {
+public:
+   TR_ALLOC(TR_Memory::IlGenerator)
+
+   UnionType(const char *name, TR_Memory* trMemory) :
+      TR::IlType(name),
+      _firstField(0),
+      _lastField(0),
+      _size(0),
+      _closed(false),
+      _symRefBV(4, trMemory),
+      _trMemory(trMemory)
+      { }
+
+   TR::DataType getPrimitiveType()                 { return TR::Address; }
+   void Close();
+
+   void AddField(const char *name, TR::IlType *fieldType);
+   TR::IlType * getFieldType(const char *fieldName);
+
+   TR::SymbolReference *getFieldSymRef(const char *name);
+   virtual bool isUnion() { return true; }
+   virtual size_t getSize() { return _size; }
+
+protected:
+   FieldInfo * findField(const char *fieldName);
+
+   FieldInfo * _firstField;
+   FieldInfo * _lastField;
+   size_t      _size;
+   bool        _closed;
+   TR_BitVector _symRefBV;
+   TR_Memory* _trMemory;
+   };
+
+void
+UnionType::AddField(const char *name, TR::IlType *typeInfo)
+   {
+   if (_closed)
+      return;
+
+   auto fieldSize = typeInfo->getSize();
+   if (fieldSize > _size) _size = fieldSize;
+
+   FieldInfo *fieldInfo = new (PERSISTENT_NEW) FieldInfo(name, 0 /* no offset */, typeInfo);
+   if (0 != _lastField)
+      _lastField->setNext(fieldInfo);
+   else
+      _firstField = fieldInfo;
+   _lastField = fieldInfo;
+   }
+
+void
+UnionType::Close()
+   {
+   _closed = true;
+   }
+
+FieldInfo *
+UnionType::findField(const char *fieldName)
+   {
+   FieldInfo *info = _firstField;
+   while (NULL != info)
+      {
+      if (strncmp(info->_name, fieldName, strlen(fieldName)) == 0)
+         return info;
+      info = info->_next;
+      }
+   return NULL;
+   }
+
+TR::IlType *
+UnionType::getFieldType(const char *fieldName)
+   {
+   FieldInfo *info = findField(fieldName);
+   if (NULL == info)
+      return NULL;
+   return info->_type;
+   }
+
+TR::IlReference *
+UnionType::getFieldSymRef(const char *fieldName)
+   {
+   FieldInfo *info = findField(fieldName);
+   if (NULL == info)
+      return NULL;
+
+   TR::SymbolReference *symRef = info->getSymRef();
+   if (NULL == symRef)
+      {
+      // create a symref for the new field and set its bitvector
+      TR::Compilation *comp = TR::comp();
+      auto symRefTab = comp->getSymRefTab();
+      TR::DataType type = info->getPrimitiveType();
+
+      TR::Symbol *symbol = TR::Symbol::createShadow(comp->trHeapMemory(), type);
+      symRef = new (comp->trHeapMemory()) TR::SymbolReference(symRefTab, symbol, comp->getMethodSymbol()->getResolvedMethodIndex(), -1);
+      symRef->setOffset(0);
+      symRef->setReallySharesSymbol();
+
+      TR_SymRefIterator sit(_symRefBV, symRefTab);
+      for (TR::SymbolReference *sr = sit.getNext(); sr; sr = sit.getNext())
+          {
+          symRefTab->makeSharedAliases(symRef, sr);
+          }
+
+      _symRefBV.set(symRef->getReferenceNumber());
+
+      info->cacheSymRef(symRef);
+      }
+
+   return static_cast<TR::IlReference *>(symRef);
+   }
+
 class PointerType : public TR::IlType
    {
 public:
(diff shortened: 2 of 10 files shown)


compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+
+void
+UnionType::AddField(const char *name, TR::IlType *typeInfo)
+   {
+   if (_closed)
+      return;
+
+   TR::DataType primitiveType = typeInfo->getPrimitiveType();
+   auto fieldSize = typeInfo->getSize();
+   _size = fieldSize > _size ? fieldSize : _size;
```
mgaudet: What you're doing here is `_size = max(_size, fieldSize)` -- I'd just say that.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+   FieldInfo * _lastField;
+   size_t      _size;
+   bool        _closed;
+   TR_BitVector _symRefBV;
+   };
+
+void
+UnionType::AddField(const char *name, TR::IlType *typeInfo)
+   {
+   if (_closed)
```
mgaudet: shouldn't adding to a closed union assert or throw or something?

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+      return;
+
+   TR::DataType primitiveType = typeInfo->getPrimitiveType();
+   auto fieldSize = typeInfo->getSize();
+   _size = fieldSize > _size ? fieldSize : _size;
+
+
+   FieldInfo *fieldInfo = new (PERSISTENT_NEW) FieldInfo(name, 0 /* no offset */, typeInfo);
```
mgaudet: What is the expected lifetime of a FieldInfo... a little surprised to see it `PERSISTENT_NEW`'d

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+   if (0 != _lastField)
```
mgaudet: Why a hand rolled linked list of fieldinfo, and not just a `std::vector`?

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+   _lastField = fieldInfo;
+
+   // create a symref for the new field and set its bitvector
+   TR::Compilation *comp = TR::comp();
+
+   TR::Symbol *symbol = NULL;
+   if (TR::Int32 == primitiveType)
+      symbol = comp->getSymRefTab()->findOrCreateGenericIntShadowSymbol();
+   else
+      symbol = TR::Symbol::createShadow(comp->trHeapMemory(), primitiveType);
```
mgaudet: I _think_ I know what you're doing here, but I the special handling of `TR::Int32` deserves some explanation.

compiler/ilgen/TypeDictionary.cpp
mgaudet: Oh

NVM. `FieldInfo` is not your creation.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+      }
+   }
+
+FieldInfo *
+UnionType::findField(const char *fieldName)
+   {
+   FieldInfo *info = _firstField;
+   while (NULL != info)
+      {
+      if (strncmp(info->_name, fieldName, strlen(fieldName)) == 0)
```
mgaudet: Maybe add `operator==` to `FieldInfo`?

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,117 @@ StructType::getFieldSymRef(const char *fieldName)
...
+   fieldInfo->_symRef = new (comp->trHeapMemory()) TR::SymbolReference(comp->getSymRefTab(), symbol, comp->getMethodSymbol()->getResolvedMethodIndex(), -1);
+   _symRefBV.set(fieldInfo->_symRef->getReferenceNumber());
+   }
+
+void
+UnionType::Close()
+   {
+   _closed = true;
+
+   // alias all the fields together
```
mgaudet: :+1:

compiler/ilgen/TypeDictionary.cpp
mstoodle: agreed, it probably should. In his defence, I'm sure he's mimicking what I did for adding a field to a struct. They should both assert.

compiler/ilgen/TypeDictionary.cpp
mstoodle: TypeDictionary was meant to have a lifetime longer than a compilation, so persistent allocation is the right approach here.

compiler/ilgen/TypeDictionary.cpp
mstoodle: or even just "if (fieldSize > _size) _size = fieldSize;"

compiler/ilgen/TypeDictionary.cpp
mstoodle: for background: JitBuilder currently avoids STL because JitBuilder builds into a static library and it's a lot harder to link against that static library on various platforms if it depends on STL.

Leonardo2718:
Now that #711 has been merged, I've been able to clean up my code.

@0xdaryl or @vijaysun-omr: this PR is now ready for review.

charliegracie:
This change would be very useful and allow a lot of clean up in the Lua-Vermelha implementation.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,121 @@ StructType::getFieldSymRef(const char *fieldName)
...
+      { }
+
+   TR::DataType getPrimitiveType()                 { return TR::Address; }
+   void Close();
+
+   void AddField(const char *name, TR::IlType *fieldType);
+   TR::IlType * getFieldType(const char *fieldName);
+
+   TR::SymbolReference *getFieldSymRef(const char *name);
+   bool isUnion() { return true; }
```
0xdaryl: Should this be `virtual` ?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1232,7 +1232,9 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
```
0xdaryl: Please make the commit title for 30418641 more descriptive.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -577,6 +586,47 @@ TypeDictionary::CloseStruct(const char *structName)
...
+   }
+
+void
+TypeDictionary::UnionField(const char *unionName, const char *fieldName, TR::IlType *type)
+   {
+   TR_HashId unionID=0;
+   if (_unionsByName->locate(unionName, unionID))
+      {
+      UnionType *unionType = (UnionType *) _unionsByName->getData(unionID);
+      unionType->AddField(fieldName, type);
```
vijaysun-omr: Should we check if a field by the same name was added to this union before ? There seems to be no protection either here or in AddField in case some code ended up calling these apis twice for the same field in a union. I know this is not a good practice but it is cleaner to think duplication won't happen.

vijaysun-omr:
I reviewed these changes and approved all but one of the commits. I'll approve the final commit here once my question is addressed.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -475,11 +592,25 @@ TypeDictionary::PointerTo(TR::IlType *baseType)
...
 TR::IlReference *
-TypeDictionary::FieldReference(const char *structName, const char *fieldName)
+TypeDictionary::FieldReference(const char *typeName, const char *fieldName)
    {
    TR_HashId structID = 0;
-   _structsByName->locate(structName, structID);
-   StructType *theStruct = (StructType *) _structsByName->getData(structID);
-   return theStruct->getFieldSymRef(fieldName);
+
+   auto found = _structsByName->locate(typeName, structID);
```
xliang6: Rename structID to fieldID?

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -285,6 +286,121 @@ StructType::getFieldSymRef(const char *fieldName)
...
+
+      TR::Symbol *symbol = TR::Symbol::createShadow(comp->trHeapMemory(), type);
+      symRef = new (comp->trHeapMemory()) TR::SymbolReference(symRefTab, symbol, comp->getMethodSymbol()->getResolvedMethodIndex(), -1);
+      symRef->setOffset(0);
+      symRef->setReallySharesSymbol();
+
+      TR_SymRefIterator sit(_symRefBV, symRefTab);
+      for (TR::SymbolReference *sr = sit.getNext(); sr; sr = sit.getNext())
+          {
+          comp->getSymRefTab()->makeSharedAliases(symRef, sr);
```
xliang6: minor change:
symRefTab->makeSharedAliases(symRef, sr);

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -475,11 +592,25 @@ TypeDictionary::PointerTo(TR::IlType *baseType)
...
+      return theStruct->getFieldSymRef(fieldName);
+      }
+
+   found = _unionsByName->locate(typeName, structID);
+   if (found)
+      {
+      UnionType *theUnion = (UnionType *) _unionsByName->getData(structID);
+      return theUnion->getFieldSymRef(fieldName);
+      }
+
```
xliang6: What happens if you have a struct field and a union field with the same name, and you actually want the one from the union?

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: Technically, it doesn't need to be marked as virtual because `OMR::IlType::isUnion()` already is virtual :stuck_out_tongue_closed_eyes: 

For clarity though (and to keep things consistent), I'll mark it as virtual explicitly.

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: Looking at it again, I agree that there should be a check for this.

In my implementation, I just copied what `StructType::AddField` does. It doesn't check for an existing field with the same name so I didn't check for it either.

I can add a check to both.

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: Good catch!

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: :stuck_out_tongue: fixed.

Leonardo2718: Actually should be `typeID`.

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: Hm...

I think the problem here is not what happens when you have the same field name in a struct and a union, but rather what happens when a union and a struct have the same name.

This should of course never happen in a well-formed program, but JitBuilder does not currently check for this. It looks like it doesn't even check for two structs with the same name.

This is a large enough issue that I think it should be solved in a separate pull request.

compiler/ilgen/TypeDictionary.cpp
Leonardo2718: Come to think of it, this is really more related to issue #500. Would it be okay if I delayed this fix to a different PR? I can add a comment to #500 that explains this issue specifically so it's not forgotten.

vijaysun-omr: Okay yes, it does fall in the general category of making JitBuilder more resilient and documenting under #500 is fine.

compiler/ilgen/TypeDictionary.cpp
xliang6: Yes, if a struct and union have the same name.

Leonardo2718:
Thanks everyone for contributing!

0xdaryl:
Admiral, there be unions here!

PR #649, opened 2017-01-03
Fix handling of induction offsets in OSR analyses

andrewcraik:
Recent changes added an API to allow an OSR point to have an induction offset and when that offset was positive to perform OSR def and liveness analyses differently. It turns out that you not only need the state at the transition after a potential OSR point, you need the state before. This is specifically for calls where inlining the call my require that you are able to obtain the state ahead of the call. This commit removes the induction offset API in favour of adding additional OSR points to track this necessary information.


andrewcraik:
@vijaysun-omr ping?

Issue #650, opened 2017-01-03
Unused variable in CollectorLanguageInterface without scavenger or concurrent mark enabled


ktbriggs:
Found by @youngar while compiling OMR for Ruby. If OMR_GC_MODRON_CONCURRENT_MARK and OMR_GC_MODRON_SCAVENGER are undefined for build gcc complains about unused env variable in MM_CollectorLanguageInterface::writeBarrierStore/Update() methods.

```
CollectorLanguageInterface.cpp: In member function 'void MM_CollectorLanguageInterface::writeBarrierStore(OMR_VMThread*, void*, fomrobject_t*, void*)':
CollectorLanguageInterface.cpp:37: error: unused variable 'env'
CollectorLanguageInterface.cpp: In member function 'void MM_CollectorLanguageInterface::writeBarrierUpdate(OMR_VMThread*, void*, void*)':
CollectorLanguageInterface.cpp:49: error: unused variable 'env'
```

charliegracie:
Closing as this was resolved via #651

PR #651, opened 2017-01-03
Fix unused variable errors in CollectorLanguageInterface.cpp

ktbriggs:
Must declare unused offending variable (env) only when 
OMR_GC_MODRON_CONCURRENT_MARK or OMR_GC_MODRON_SCAVENGER are defined.

Also did some light refactoring to simplify the methods involved.

Peer reviewer: @youngar 
Senior reviewer: @charliegracie 
Issue: #650 

### DIFF
--- a/gc/base/CollectorLanguageInterface.cpp
+++ b/gc/base/CollectorLanguageInterface.cpp
@@ -29,37 +29,12 @@
 struct OMR_VMThread;
 
 void
-MM_CollectorLanguageInterface::writeBarrierStore(OMR_VMThread *omrThread, omrobjectptr_t parentObject, fomrobject_t *parentSlot, omrobjectptr_t childObject)
+MM_CollectorLanguageInterface::writeBarrier(OMR_VMThread *omrThread, omrobjectptr_t parentObject, omrobjectptr_t childObject)
 {
-	GC_SlotObject slotObject(omrThread->_vm, parentSlot);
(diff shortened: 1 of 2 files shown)


gc/base/CollectorLanguageInterface.cpp
```diff
@@ -29,37 +29,12 @@
 struct OMR_VMThread;
 
 void
-MM_CollectorLanguageInterface::writeBarrierStore(OMR_VMThread *omrThread, omrobjectptr_t parentObject, fomrobject_t *parentSlot, omrobjectptr_t childObject)
+MM_CollectorLanguageInterface::writeBarrier(OMR_VMThread *omrThread, omrobjectptr_t parentObject, omrobjectptr_t childObject)
```
charliegracie: Could this function be marked as MMINLINE (and leave it in the cpp) so that it gets inlined into the 2 callers.?

ktbriggs: Yes. That was an oversight.

ktbriggs: It is declared as private MMINLINE in CollectorLanguageInterface.hpp:

```
private:
	MMINLINE void writeBarrier(OMR_VMThread *omrThread, omrobjectptr_t parentObject, omrobjectptr_t childObject);
```

PR #653, opened 2017-01-09
Updates to support Inlining for Ruby 2.4

mgaudet:
A couple of small tweaks to support inlining for Ruby 2.4 branch

### DIFF
--- a/compiler/compile/VirtualGuard.cpp
+++ b/compiler/compile/VirtualGuard.cpp
@@ -134,22 +134,37 @@ TR_VirtualGuard::createRubyInlineGuard
 (TR_VirtualGuardKind kind, TR::Compilation * comp, int16_t calleeIndex,
  TR::Node* callNode, TR::TreeTop * destination, TR_OpaqueClassBlock *thisClass)
 {
+#ifdef RUBY_PROJECT_SPECIFIC
   //Ensure that ILGen hasn't changed the children's layout we expect.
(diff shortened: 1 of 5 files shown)


compiler/compile/VirtualGuard.cpp
```diff
@@ -134,22 +134,35 @@ TR_VirtualGuard::createRubyInlineGuard
...
-  TR::Node* isInlineable = TR::Node::create(TR::lcall, 2);
+  TR::Node* isInlineable = TR::Node::create(TR::lcall, 3);
+
+  // Nail down what the serial numbers are at time of inlining -- these will 
+  // match the inlined body serial numbers. 
   isInlineable->setSymbolReference(inline_guard_helperSymRef);
-  isInlineable->setAndIncChild(0, TR::Node::aconst(callInfo, (uintptrj_t)callInfo->getAddress()));
-  isInlineable->setAndIncChild(1, receiver);
+
+  //TODO: These trees ought to be using i|l consts... depending on the platform size,
```
0xdaryl: The general policy with OMR has been to not introduce TODO's into the code.  If you want to track outstanding work, please create an issue so it doesn't get forgotten.

mgaudet: Fixed by static asserting -- turns out, `rb_serial_t` is always 64 bits right now in the interpreter.

Issue #655, opened 2017-01-10
Virtual guard head merger should not move monitor stores up across guards


liqunl:
A monitor store stores the resource that's going to be synchronized to a temp at the beginning of a inlined body for a synchronized method, and it indicates an entry to the lock which will be cleared once an exitfence is encountered. Moving such a store up across the guard that's protecting it would cause a wrong monitor map to be generated,  potentially leading to a gc crash.

PR #656, opened 2017-01-10
Virtual guard head merger fixes

liqunl:
Virtual guard head merger can move trees ahead of a guard up, including
trees ahead of an HCR guard. Those trees can have side effect which
should stay in place. This commit fixes the problem and allows the
merging of a runtime guard when there are only priv arg stores and live
monitor stores ahead of it.

The commit also disaliases the liveMonitorInitStore flag and the
privatizedInlinerArg flag. And code is added to sinkstore to prevent it
from sinking live monitor stores which was protected by
privatizedInlinerArg flag.

issue: #655

### DIFF
--- a/compiler/optimizer/VirtualGuardHeadMerger.cpp
+++ b/compiler/optimizer/VirtualGuardHeadMerger.cpp
@@ -46,6 +46,35 @@ static bool isMergeableGuard(TR::Node *node)
    return mergeOnlyHCRGuards ? node->isHCRGuard() : node->isNopableInlineGuard();
    }
 
+// Splits monitor store from priv args
+// Return the block containing only the priv args if exist
+static TR::Block* splitMonitorStoreFromPrivArg(TR::Block* block, TR::CFG *cfg)
+   {
+   TR::TreeTop *start = block->getFirstRealTreeTop();
+   TR::TreeTop *end = block->getLastRealTreeTop();
+   bool hasMonitorStore = false;
+   TR::TreeTop *firstPrivArg = NULL;
+   for(TR::TreeTop *tt = start; tt && tt != end; tt = tt->getPrevTreeTop())
+      {
+      TR::Node * node = tt->getNode();
+      if (node->getOpCode().hasSymbolReference() && node->getSymbol()->holdsMonitoredObject())
+         hasMonitorStore = true;
+      else if(node->chkIsPrivatizedInlinerArg())
+         {
+         firstPrivArg = tt;
+         // Monitor stores always appear before priv args, so we can safely stop here
+         break;
+         }
+      }
+   if (hasMonitorStore && firstPrivArg)
+      return block->split(firstPrivArg, cfg, true /* fixupCommoning */, false /* copyExceptionSuccessors */);
+   else if (firstPrivArg)
+      return block;
+   else
+      // This block does not contain a priv arg
+      return NULL;
+   }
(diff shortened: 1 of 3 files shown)


liqunl:
@andrewcraik @cathyzhyi Can you review my vghm fix?

andrewcraik:
I think the change looks good and fits in well with the code.

cathyzhyi:
LGTM.

liqunl:
@vijaysun-omr Hi Vijay, can you merge my PR when you have time? It has passed personal build and acceptance build.

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -260,28 +292,41 @@ int32_t TR_VirtualGuardHeadMerger::perform() {
...
-				  runtimeIns = privargBlock;
-			   privargIns = privargBlock;
-
-			   // refresh the insertPoint since it could be stale after the above updates
-			   insertPoint = runtimeIns;
+               // We should leave code ahead of an HCR guard in place because:
+               // 1, it might have side effect to runtime guards after it, moving it up might cause us to falsely merge
+               //    the subsequent runtime guards
+               // 2, it might contain live monitor, moving it up above a guard can affect the monitor's live range
+               if (!guard2->isHCRGuard())
```
vijaysun-omr: Could you please elaborate on why this is legal for non HCR guards but illegal for HCR guards ?

andrewcraik: So the logic here is as follows - HCR guards are allowed to move across any arbitrary code as long as the treetops crossed cannot gc and return - ie we cannot stop and so the state at the destination of the move must match the start. Because the HCR guard does not rely on the receiver or any of the arguments to the method being protected we don't need to move the privargs around with the guard itself. For a runtime guard we have to make sure to move the privargs up when the guard is moved up to ensure the guard operates correctly. This code handles moving the privargs up and so it is legal and required for a runtime guard, but it is not legal or correct for an HCR guard due to the arbitrary nature of the code being crossed.

liqunl: I was working on 3 paragraphs to explain, but found that we should move the priv arg ahead of an HCR guard up as well, because the priv arg  of next guard might move across it and there might be data dependency between them. I'm working on a new fix to cover that case.

liqunl: BTW, a priv arg store is movable only when the store does not depend on code ahead of it.

liqunl: As per discuss with Andrew, we need a extra check to decide whether it's safe to move the runtime guard, i.e. if there are priv arg stores for other calls ahead of the runtime guard, it's not safe to move it. Builds are running and will remove WIP once they have passed.

liqunl:
An extra check is added in isSafeToMoveGuard to prevent merging runtime guards when there are priv arg stores for other guards ahead of it. @andrewcraik @vijaysun-omr please review and merge when you have time.

andrewcraik:
LGTM

liqunl:
@vijaysun-omr I have rebased and retested the commit, can you review and merge it if you think the change is ok?

vijaysun-omr:
Looks okay to me. Merging.

PR #660, opened 2017-01-11
Fix to return correct currentMethod/Symbol for comp

cathyzhyi:
Before this fix, *getCurrentMethod* and *getMethodSymbol* API on comp
return compiling(caller) method/symbol for ilgen but callee method/symbol for
optimization. The reason behind this is that ilgen needs the 'caller' in
some cases and *_methodSymbol* on *IlGenerator* would keep callee
method/symbol if 'callee' is needed. However, this can cause
inconsistency between the ilgen and optimizer when a temp is created
in ilgen but analyzed in optimizer.

This fix changes the 2 APIs to make ilgen and optimizer get the same
current method or symbol. The compiling method is indeed needed in
several places and there is another API getJittedMethodSymbol on comp
for this.

The changes are in 3 kinds:
1. add methodSymbol() to TR_IlGenerator to facilitate 'kind 2'
2. change getCurrentMethod and getMethodSymbol on comp to return callee
symbol
3. fix several places should have used getJittedMethodSymbol rather than
getMethodSymbol


vijaysun-omr:
I would also like @andrewcraik to review this.

cathyzhyi:
mark as WIP to investigate an issue found in personal builds.

vijaysun-omr:
Actually add a WIP tag as was apparently Yi's intent based on her last comment.

cathyzhyi:
@vijaysun-omr @andrewcraik could you guys take a look at this changeset. I've verified it passes all the personal builds.

andrewcraik:
The change to make the current method sane in ilgen is a good one. The main risk is with people who expected the compilation method now getting the ilgen method and having to vet the code for all of those occurrences and fix them as part of the change. I see a couple were caught here and I haven't found more myself yet. @vijaysun-omr not sure if you see any I missed?

vijaysun-omr:
I do not see any other calls in the OMR layer that could change behavior in ways that could lead to problems. The owning method changes for some symbols being created but I don't see what could go wrong in any obvious way.

PR #661, opened 2017-01-12
Remove duplicate decrement in simplifyChildren()

jdmpapin:
The `decFutureUseCount()` call appears to have been accidentally
duplicated. Incorrect future use count values prevent at least the
simplification of cascading `aladd` operations with constants, i.e.

    aladd
      aladd              aladd
        ...        ==>     ...
        lconst a           lconst (a + b)
      lconst b


compiler/optimizer/OMRSimplifierHelpers.cpp
```diff
@@ -113,7 +113,6 @@ void simplifyChildren(TR::Node * node, TR::Block * block, TR::Simplifier * s)
       {
       TR::Node * child = node->getChild(i);
       child->decFutureUseCount();
-      child->decFutureUseCount();
```
vijaysun-omr: Okay, this fixes a clear bug and so I am merging in right away.

PR #662, opened 2017-01-12
Extend signed comparisons in loop strider widening

jdmpapin:
As part of its sign-extension pass, loop strider now identifies 32-bit
signed comparisons made directly against any replaced IV (induction
variable), and changes them into 64-bit signed comparisons by wrapping
the children in `i2l`. At least the `i2l` around the replaced IV will
be eliminated, so that afterward it will still be a direct child of the
comparison.

### DIFF
--- a/compiler/optimizer/InductionVariable.cpp
+++ b/compiler/optimizer/InductionVariable.cpp
@@ -4658,6 +4658,7 @@ void TR_LoopStrider::walkTreesAndFixUseDefs(
    replaceLoadsInStructure(
       loopStructure,
       _loopDrivingInductionVar,
+      newIntValue,
       newSymbolReference,
       exLoads,
       visited);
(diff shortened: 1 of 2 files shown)


compiler/optimizer/InductionVariable.cpp
```diff
@@ -4724,6 +4732,110 @@ void TR_LoopStrider::replaceLoadsInSubtree(
...
+   \verbatim
+      iflcmpge
+        i2l
+          l2i
+            lload k
+        i2l
+          iload n
+   \endverbatim
+ *
+ * Because the \c l2i node here is an ex-load, it will cancel the \c i2l later:
```
vijaysun-omr: An l2i operation followed by an i2l operation is not necessarily a cancelling pair, is it ? 
The highest bit in the lower 32 bits of the lload will be used to fill in the upper 32 bits in the result of the i2l and this need not be the same value as the llload. Am I missing something ?

jdmpapin: You're right: `i2l` of `l2i` doesn't cancel in general. This one will cancel only because the `l2i` is an "ex-load," i.e. it was originally `iload i`, but was transmuted into `l2i (lload k)` in the process of replacing the 32-bit IV `i` with a 64-bit IV `k`. The cancellation happens later in `signExtend()`, and is [justified in the documentation](https://github.com/eclipse/omr/blob/a5d1fcbddc9a5d064fe28e93c090942db69a06ef/compiler/optimizer/InductionVariable.cpp#L4961-L4984).

vijaysun-omr: Okay thanks for the explanation; it makes sense since it was not a random lload but one created by the strider for an ind var that was an int originally.

vijaysun-omr:
Looks good to me. I can merge once @andrewcraik reviews and gives his approval too.

andrewcraik:
LGTM

PR #663, opened 2017-01-12
Move ineliminable sign-extensions in loop strider

jdmpapin:
When loop strider replaces a 32-bit induction variable `i` with a new
64-bit `k` within a loop, the original loads of `i` in that loop
become `l2i` of loads of `k`. Such an `l2i` node may occur under
arithmetic in a context such as the following, where it is not
possible to eliminate the outer `i2l` node.

    i2l
      iadd (cannotOverflow)
        iload foo
        l2i
          lload k

In this case, changing the `i2l` into `ladd` would require creating a
new `i2l` wrapping `iload foo`, which so far strider elects never to
do. But with this change, the above tree will now transform into the
following tree that does fewer conversions:

    ladd (cannotOverflow)
      i2l
        iload foo
      lload k

### DIFF
--- a/compiler/optimizer/InductionVariable.cpp
+++ b/compiler/optimizer/InductionVariable.cpp
@@ -4745,7 +4745,7 @@ void TR_LoopStrider::replaceLoadsInSubtree(
  *
  * \param exLoads the set of \c l2i nodes that used to be \c iload
  * \see detectLoopsForIndVarConversion(TR_Structure*, TR::NodeChecklist&)
- * \see signExtend(TR::Node*, TR::NodeChecklist&, NodeMap&)
+ * \see signExtend(TR::Node*, TR::NodeChecklist&, SignExtMemo&)
  * \see transmuteDescendantsIntoTruncations(TR::Node*, TR::Node*)
  */
 void TR_LoopStrider::eliminateSignExtensions(TR::NodeChecklist &exLoads)
@@ -4754,11 +4754,18 @@ void TR_LoopStrider::eliminateSignExtensions(TR::NodeChecklist &exLoads)
 
    auto alloc = comp()->allocator();
    auto mapAlloc = getTypedAllocator<std::pair<ncount_t, TR::Node*> >(alloc);
-   NodeMap signExtMemo(std::less<ncount_t>(), mapAlloc);
+   SignExtMemo signExtMemo(std::less<ncount_t>(), mapAlloc);
 
    TR::TreeTop *start = comp()->getStartTree();
    for (TR::TreeTop *tt = start; tt != NULL; tt = tt->getNextTreeTop())
       eliminateSignExtensionsInSubtree(tt->getNode(), exLoads, visited, signExtMemo);
+
+   // References from signExtMemo were included in reference counts; so now
+   // release all of them. This avoids incorrectly inflating the reference
+   // count of any pre-existing node that was wrapped in an i2l but for which
+   // the new i2l node was not used.
+   for (auto it = signExtMemo.begin(); it != signExtMemo.end(); ++it)
+      it->second.extended->recursivelyDecReferenceCount();
 
    if (trace())
       comp()->dumpMethodTrees("trees after eliminating sign extensions");
@@ -4769,7 +4776,7 @@ void TR_LoopStrider::eliminateSignExtensionsInSubtree(
    TR::Node *parent,
    TR::NodeChecklist &exLoads,
    TR::NodeChecklist &visited,
-   NodeMap &signExtMemo)
+   SignExtMemo &signExtMemo)
    {
    if (visited.contains(parent))
       return;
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
I would like @andrewcraik to review this as well.

compiler/optimizer/InductionVariable.cpp
```diff
@@ -4946,70 +4978,123 @@ TR::Node *TR_LoopStrider::signExtend(
...
    if (!node->cannotOverflow())
-      return NULL;
-
-   TR::Node *extLeft = signExtend(node->getChild(0), exLoads, memo);
+      return rejected;
+
+   static const char * const moveDisableEnv =
+      feGetEnv("TR_disableLoopStriderMoveSignExtIntoChild");
+   static const bool moveDisable =
+      moveDisableEnv != NULL && moveDisableEnv[0] != '\0';
```
andrewcraik: This is a strange convention - normally a disable envVar will disable if the env var is set regardless of value. Any particular reason for this divergence from convention here?

andrewcraik:
I believe the logic is sound and addresses the issue outlined in the pull request. I made one comment on how and env var was being used. Once that is addressed I'll read the PR again.

compiler/optimizer/InductionVariable.cpp
jdmpapin: I was under the impression that it's rare (in general) to treat unset environment variables differently from those set to the empty string. For example, shell scripts often test for environment variables with something like `test -n "$VAR"`. Maybe this is a peculiarity of my own, but if the code tested only `moveDisableEnv != NULL`, I would be surprised at the behaviour of a script like this:

    for disable in '' 1; do
        TR_disableLoopStriderMoveSignExtIntoChild="$disable" vm-with-jit [...]
    done

It also isn't entirely without precedent. I've introduced multiple other environment variable tests following the same convention, even just in this file. Most recently, we have `TR_disableLoopStriderWidenComparison` from #662.

Of course, it's not as though this convention helps all that much to prevent error. Just about anybody might still be surprised if the above loop said, for example, `for disable in 0 1`, or `for disable in no yes`. And nothing prevents a typo in the variable name.

For that reason, this should ideally be a JIT option instead, if we want to be particular about it. The name would be checked, and it could be restricted to an option set, and so on. I used an environment variable here for expediency, because I often get merge conflicts when I'm adding a JIT option. (I think this is the story of most of our environment variables.)

So with that said, at your preference I can change this environment variable to use the prevailing convention of `moveDisable = moveDisableEnv != NULL` (full-stop), or make it into a JIT option, or leave it as is. You should just be aware that there are others like this one. I hope nobody is out there with the opposite of my expectation, trying to disable things with `TR_disableSomething=''`.

andrewcraik: Ok - we need to get this fix merged, but the usage is non-standard compared to the rest of the JIT where exporting a variable (regardless of value or non-value) is enough and you have to unset to clear it. We should probably try to standardize the usage across opts, but since there are other instances already we can merge this and look to clean-up the usage as another commit after this is sorted.

mgaudet: Did you guys want to open an Issue to track standardizing environment variable usage?

andrewcraik:
@jdmpapin @vijaysun-omr the env var is consistent with the file so can we get this merged and we can then fix up the env vars if we need to as a separate item.

vijaysun-omr:
@jdmpapin could you please give a brief summary of your reasoning for why the 2 IL trees you show in your first comment would result in the same 64-bit (long) value ? In particular, I am wondering about the value of the bit (MSB of lower 32-bits) used to fill the upper 32-bits of the i2l nodes in the two cases being different (from iadd in one case and from iload in the other). While I do not have example values to produce a different answer at the top level of the tree you showed, I also don't have a concise rationale for why the 2 trees would produce the same answer (and that's what I am asking you for). Maybe the transformation is valid precisely because the cannotOverflow flag is set and not otherwise ?

jdmpapin:
Right, the transformation requires that `cannotOverflow` is set on the arithmetic node. Otherwise, `signExtend` won't touch it. The test is in the first few lines of `signExtendBinOp`.

jdmpapin:
In the documentation of `signExtend`, there is [an explanation of the original transformation](https://github.com/eclipse/omr/blob/b1ce2a99a27453711434ce1003c3c00e88f15e74/compiler/optimizer/InductionVariable.cpp#L4825-L4857), and also [of this broader transformation](https://github.com/eclipse/omr/blob/b1ce2a99a27453711434ce1003c3c00e88f15e74/compiler/optimizer/InductionVariable.cpp#L4884-L4900). In the latter `cannotOverflow` is implicit because it's irrelevant to the point made there, and repeating it is tiresome.

vijaysun-omr:
Okay thanks a lot. That helps clarify the picture in a scenario without any overflow possible.

vijaysun-omr:
Code looks okay to me. Merging.

Issue #665, opened 2017-01-13
Ruby JIT multiplexing pseudo-TOC on sys TOC


zl-wang:
After pull request #612 was merged, Ruby tests started to fail. 
Finding out unfortunately Ruby JIT is multiplexing JIT pseudo-TOC on sys TOC register.

mgaudet:
Is there a more permanent fix we should be making about the TOC? _ie_ Mandating a certain usage pattern for it?

zl-wang:
TOC usage pattern is fixed and specified by the ABI:  caller needs to set up for the callee across a module. And, caller needs to restore its own upon return.

JIT pseudo-TOC, on the other hand, is another single table used by all JIT-ed code.  From original design, it was based off another (preserved) register.

Discovering whether it is multiplexed as in Ruby JIT took the roundabout.  Not sure about 32bit Ruby (is there?) status, but keeping the original code is a safe bet.

mgaudet:
No such thing as 32 bit Ruby. :) So that's one good thing. 

When you say 'Multiplexed' here, and in the commit message, I am not quite sure what you mean; Is it just that the preserved register (`r2`) is being shared between JIT and non-JIT code for TOC usage -- [Hence the need to generate a TOC restore](https://github.com/zl-wang/omr/commit/dd1b0efc9428279c42acce8cc3e99b61f0242204#diff-8667a28b4e5624311cf3283984358c37R1472)?

zl-wang:
There is system TOC and there is JIT pseudo TOC.  In Ruby, it appeared that r2 is used to hold the base address of JIT pseudo-TOC when the control flow is in JIT-ed code.  r2 needs to hold the base address of system TOC when the control flow is in ABI-conforming code (by definition).  This deviates from Java JIT where r2 is not multiplexed as such (r16 is dedicated for JIT pseudo-TOC).

ymanton:
I think the system linkage code in OMR still needs some work (probably on all platforms). This code cannot stand on it's own as-is; without knowing about the ABI conventions of the **caller** it can't really do it's job, it knows about the ABI conventions of the **callee** but you really need some coordination between the two.

In the past this worked fine because the system linkage code was written to assume that the caller was following the Java JIT private linkage conventions, and the code would then generate code that was appropriate for transitioning from Java JIT to system and back. This implicit assumption cannot be made in OMR; in order to generate calls to a callee that follows a particular ABI you need to know some things about the caller's ABI as well.

[...]

PR #666, opened 2017-01-13
Issue with pseudo-TOC mulitplexing on sys-TOC reg

zl-wang:
Unforfunately, Ruby JIT multiplexed gr2 for JIT pseudo-TOC.
I figured it is true as well for Python and jittest.

Since gr2 doesn't participate in sysLinkage reg assigning and
there is only one pseudo-TOC in existence, I don't restore
that not-changed pseudo-TOC on the way out for the multiplexing
case for performance reason of course.

Tested against both Ruby and Java.

Fixed issue: #665

### DIFF
--- a/compiler/p/codegen/PPCSystemLinkage.cpp
+++ b/compiler/p/codegen/PPCSystemLinkage.cpp
@@ -1441,6 +1441,8 @@ void TR::PPCSystemLinkage::buildDirectCall(TR::Node *callNode,
    int32_t                        refNum = callSymRef->getReferenceNumber();
 
    //This is not JIT pseudo TOC, but jit-module system TOC.
+   //Unfortunately, Ruby JIT is multiplexing gr2(as system module TOC) for JIT-pseudo TOC
+
 
    if(aix_style_linkage)
          {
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
Perhaps one of the folks more familiar with code generation such as @ymanton or @fjeremic can review this one ?

ymanton:
Looks good to me. Unfortunately I think this code needs to be refactored to allow for what was originally intended, but that is a bigger undertaking. This fix is sufficient for now.

compiler/p/codegen/PPCSystemLinkage.cpp
```diff
@@ -1441,14 +1441,19 @@ void TR::PPCSystemLinkage::buildDirectCall(TR::Node *callNode,
...
          {
          //Implies TOC needs to be restored (ie not 32bit ppc-linux-be)
          //This is wrong with regard to where system TOC is restored from, but happens to be right for JIT
          //for the time being.
          if(TR::Compiler->target.cpu.isBigEndian())
-            TR::TreeEvaluator::restoreTOCRegister(callNode, cg(), dependencies);
+#if !(defined(PYTHON) || defined(OMR_RUBY) || defined(JITTEST))
+            TR::TreeEvaluator::restoreTOCRegister(callNode, cg(), dependencies)
+#endif
+            ;
```
fjeremic: Interesting use of a semicolon statement. You may want to consider introducing brackets for the if clause (create a scope for the ifdef) and moving the semicolon to line 1454. This should improve readability. It also guards against someone in the future accidentally adding another statement in the ifdef not realizing what is happening here.

zl-wang:
made the changes suggested by @fjeremic

vijaysun-omr:
Merging this now that everyone is fine with the change. Hope the devil's number (666) for this PR does not cause any havoc :)

Issue #667, opened 2017-01-13
Reinstate z9 / z10 ALS on z Systems


fjeremic:
Reinstate z9 / z10 ALS

Reinstate z9 / z10 architecture level set (ALS) in the OMR z Systems codegen. This is the minimal architecture set which is to be supported by compiler built on top of OMR moving forward. Hence we need to modify all locations that generate instructions introduced in subsequent iterations of  z/Architecture.

0xdaryl:
@fjeremic : are there any remaining tasks with this issue?  If not, please close it.

fjeremic:
@0xdaryl yes there seems to be one outstanding issue I need to fix. I'm waiting on infrastructure issues to be resolved on the z9 machine so I can hop on and reproduce the problem.

PR #668, opened 2017-01-13
Reinstate z9 / z10 ALS

fjeremic:
Reinstate z9 / z10 architecture level set (ALS) in the OMR z Systems
codegen. This is the minimal architecture set which is to be supported
by compiler built on top of OMR moving forward. Hence we need to modify
all locations that generate instructions introduced in subsequent
iterations of  z/Architecture.

This change effectively guards many sequences that we generate in the
compiler with architecture checks, and provides alternative
implementations that are supported on z9 / z10 if needed.

Issue: #667


fjeremic:
Would be good to have someone familiar with z also take a look in addition to the committer. Any takers?

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -948,6 +948,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
 
    bool opCodeIsNoOpOnThisPlatform(TR::ILOpCode &opCode);
 
+   bool supportsPrefetch() { return getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10); }
```
0xdaryl: I don't see this function used anywhere in this commit or the existing OMR codebase.  Was it added as an anticipated convenience for future commits?  If so, I think the name could use some refinement as "prefetch" is too generic a term (unless on Z it truly isn't, so correct me if I'm wrong).

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -3666,17 +3712,50 @@ OMR::Z::TreeEvaluator::ternaryEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+         {
+         TR::LabelSymbol *branchDestination = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
+         branchDestination->setEndInternalControlFlow();
+         TR::Instruction *brcInst = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNERC,
+               node, branchDestination); //, cursor);
+         brcInst->setStartInternalControlFlow();
+         TR::Instruction *conditionalLoad = generateRRInstruction(cg, trueVal->getOpCode().is8Byte() ?
+               TR::InstOpCode::LGR : TR::InstOpCode::LR, node, trueReg, falseReg);
+         TR::RegisterDependencyConditions * conditions =
+               new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 4, cg);
```
joransiu: Not necessarily a bug, but I counted only max of 3 post conditions being added here.  Do we need "4"?

joransiu: The z196 path has code to handle if trueReg/falseReg is a register pair.  However, the non-z196 path doesn't seem to handle this case.  Is that okay?

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -3637,22 +3640,65 @@ OMR::Z::TreeEvaluator::ternaryEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+            {
+            conditions->addPostCondition(falseReg, TR::RealRegister::AssignAny);
+            }
+
+         if (trueReg != falseReg && trueReg != firstReg && trueReg != secondReg)
+            {
+            conditions->addPostCondition(trueReg, TR::RealRegister::AssignAny);
+            }
+
+         TR::Instruction *conditionalLoad = generateRRInstruction(cg, trueVal->getOpCode().is8Byte() ? TR::InstOpCode::LGR : TR::InstOpCode::LR, node, trueReg, falseReg);
```
joransiu: The z196 path has code to handle if trueReg/falseReg is a register pair.  However, the non-z196 path doesn't seem to handle this case.  Is that okay?

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -63,63 +63,84 @@
 #include "z/codegen/S390GenerateInstructions.hpp"
 #include "z/codegen/S390Instruction.hpp"
 #include "z/codegen/S390OutOfLineCodeSection.hpp"
+
 static TR::InstOpCode::Mnemonic getIntToFloatLogicalConversion(TR::CodeGenerator *cg, TR::InstOpCode::Mnemonic convertOpCode)
    {
-   switch(convertOpCode)
+   if(cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z196))
```
joransiu: Nit.  Space missing between if and '('

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -63,63 +63,84 @@
...
+   return convertOpCode;
    }
 
 /**
  * "Convert to Logical" opcodes. Requires Arch(9) or above and IEEE FP
  */
 static TR::InstOpCode::Mnemonic getFloatToIntLogicalConversion(TR::CodeGenerator *cg, TR::InstOpCode::Mnemonic convertOpCode)
    {
-   switch(convertOpCode)
+   if(cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z196))
```
joransiu: Nit.  Space missing between if and '('

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2375,20 +2375,23 @@ generateRegLitRefInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op
...
-      targetSnippet->setReloType(reloType);
-      AOTcgDiag4(comp, "generateRegLitRefInstruction constantDataSnippet=%x symbolReference=%x symbol=%x reloType=%x\n",
-         targetSnippet, targetSnippet->getSymbolReference(), targetSnippet->getSymbolReference()->getSymbol(), reloType);
-
-      cursor = (TR::S390RILInstruction *) generateRILInstruction(cg, (op == TR::InstOpCode::LG)?TR::InstOpCode::LGRL:TR::InstOpCode::LRL, node, treg, targetSnippet, preced);
-      return cursor;
+      if (op == TR::InstOpCode::LG || op == TR::InstOpCode::L)
+         {
+         TR::S390ConstantDataSnippet *targetSnippet;
+         if(op == TR::InstOpCode::LG)
```
joransiu: Nit. missing space.

compiler/z/codegen/OMRCodeGenerator.hpp
joransiu: I think the z/Architecture instructions this query is guarding are the two PREFETCH DATA (PFD and PFDRL).   The other forms of prefetch are named quite differently (branch preload for i-cache prefetching, NIAI for software hints on loading data).  If we want to improve it, I guess supportsPrefetchData() would be a more suitable name...

fjeremic: Fixed in 79e4476. I removed the query altogether as there are currently no consumers of it. The `TR::Prefetch` IL evaluators on all platforms have architecture level set checks before emitting instructions for a prefetch which renders this query obsolete.

compiler/z/codegen/S390GenerateInstructions.cpp
fjeremic: Fixed in 21c9db0.

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2484,7 +2487,7 @@ generateRegLitRefInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op
       TR::S390ConstantDataSnippet * constDataSnip = cg->create64BitLiteralPoolSnippet(TR::Int64, imm);
 
       // HCR in generateRegLitRefInstruction 64-bit: register const data snippet used by Z6
```
fjeremic: Fixed in 21c9db0.

compiler/z/codegen/ControlFlowEvaluator.cpp
fjeremic: Fixed in 21c9db0.

compiler/z/codegen/ControlFlowEvaluator.cpp
fjeremic: Fixed (both) in 21c9db0.

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -3666,17 +3712,50 @@ OMR::Z::TreeEvaluator::ternaryEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+         branchDestination->setEndInternalControlFlow();
+         TR::Instruction *brcInst = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNERC,
+               node, branchDestination); //, cursor);
+         brcInst->setStartInternalControlFlow();
+         TR::Instruction *conditionalLoad = generateRRInstruction(cg, trueVal->getOpCode().is8Byte() ?
+               TR::InstOpCode::LGR : TR::InstOpCode::LR, node, trueReg, falseReg);
+         TR::RegisterDependencyConditions * conditions =
+               new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 4, cg);
+         conditions->addPostCondition(trueReg, TR::RealRegister::AssignAny);
+         if(falseReg != trueReg)
```
fjeremic: Fixed in 21c9db0.

compiler/z/codegen/FPTreeEvaluator.cpp
fjeremic: Fixed in 21c9db0.

compiler/z/codegen/FPTreeEvaluator.cpp
fjeremic: Fixed in 21c9db0.

fjeremic:
All issues identified fixed. Re-running some tests now. While we wait I'd like to understand something better. @joransiu while looking at the ternary evaluator I noted that we evaluate both conditions before any branching occurs. More specifically:

https://github.com/eclipse/omr/blob/master/compiler/z/codegen/ControlFlowEvaluator.cpp#L3601-L3614

```c++
  TR::Register *trueReg = 0;
  if(TR::TreeEvaluator::treeContainsAllOtherUsesForNode(condition,trueVal) &&
     (!trueVal->isUnneededConversion() || TR::TreeEvaluator::treeContainsAllOtherUsesForNode(condition, trueVal->getFirstChild())))
     {
     if (comp->getOption(TR_TraceCG))
        traceMsg(comp, "Calling evaluate (instead of clobber evaluate for node %p because all other uses are in the compare tree)\n",trueVal);
     trueReg = cg->evaluate(trueVal);
     }
  else
     {
     trueReg =  cg->gprClobberEvaluate(trueVal); //cg->evaluate(trueVal);
     }

   TR::Register *falseReg = cg->evaluate(falseVal); //cg->gprClobberEvaluate(falseVal);
```

How does this work exactly with children that have side-effects? For example, what if both children were method / function calls? Is there some implicit assumption that the children of a ternary IL must be side-effect free? Perhaps @andrewcraik or @vijaysun-omr can shed some light on this.

andrewcraik:
The key point about the ternary opcode is that it is NOT the same as the ?: operator. The OMR ternary opcodes will evaluate their children completely and the value of the node is determined based on the condition _after_ all children are evaluated.

If the node had the same semantics as ?: you would have control flow in the trees not represented in the CFG and that would cause a lot of issues for the optimizer and its dataflow analysis, hence the definition of all children are evaluated (like any other node) before the value selection occurs. An overload of the meaning of ternary, but the OMR definition is all children and then select the value.

fjeremic:
Thanks for the explanation @andrewcraik. I'm going to open an issue linking to your comment so we can add some documentation here as a naive developer such as myself could assume TR::ternary IL maps directly to ternary operator, which is clearly not the case.

mstoodle:
Just to complete the point: a tree should only ever have one major side effect (call, store, etc.), so just like every other kind of node, a ternary operation shouldn't have side effects under its children. Calls should be anchored directly under a treetop or store tree.

fjeremic:
> Just to complete the point: a tree should only ever have one major side effect (call, store, etc.), so just like every other kind of node, a ternary operation shouldn't have side effects under its children. Calls should be anchored directly under a treetop or store tree.

@mstoodle Thanks. I suppose call is the extreme case. Perhaps a load would do as well:

```c++
int foo(bool a)
{
    return a ? 1 : *static_cast<int*>(0);
}
```

This is where my confusion arises and why I think we should better document the ternary IL opcode. If one assumes that :? maps to TR::ternary they are in for a world of trouble as described by @andrewcraik.

andrewcraik:
Thanks @mstoodle for making the side-effect anchorning point and documenting the behaviour of ternary is a good idea - it has tripped up a number of people of the course of its existence.

mstoodle:
Yeah, basic blocks should stay basic blocks (ok, super blocks if you want to get technical because of exception checks) at the IL level: no internal incoming flow edge.

mstoodle:
@fjeremic Note that a load is not considered a side effect (the nullcheck that precedes it for Java is though :) ). You can put multiple loads into a single tree with impunity.

fjeremic:
Once reviewers give final approval I will squash the changes.

0xdaryl:
LGTM.  @joransiu, any final concerns before Filip squashes commits?

fjeremic:
@joransiu any other concerns?

joransiu:
LGTM.

fjeremic:
Thanks for the reviews. I have squashed my changes and the TravisCI build is kicked off. When it completes this change is ready to be merged. @0xdaryl FYI.

PR #673, opened 2017-01-16
Flag for OSR infrastructure removal

ncough:
OSR infrastructure must be present to support induce OSR, however,
components of this infrastructure, particularly exception edges, may
have been removed. This adds a flag to determine if the exception
edges have been removed and extends existing functions to consider it.

Also modifies conditions for induce OSR, such that it is supported
along side MimicInterpreterFrameShape in the event of full speed
debugging.


andrewcraik:
@ncough can you remove the signed-off line from your commit message please? Otherwise I think this should achieve the desired affect.

andrewcraik:
@vijaysun-omr I think this is ready for review - LGTM

vijaysun-omr:
Looks reasonable to me to disallow induction once OSR exception edges have been removed. Merging.

PR #608, opened 2016-12-15
DDR DwarfParser for AIX

VermaSh:
These files contain the code to parse the symbol table of a c or c++ compiled file for AIX. What's needed now to resolve the issue is to read the unorderd map and create dies. There is useful information regarding reading the unorderd map in notes.txt attached to https://github.com/eclipse/omr/issues/606

PR #674, opened 2017-01-16
Populate Dwarf Compilation Unit Contexts for AIX

twlogger:
Much of the code in AixDwarfParser.cpp and AixDwarfParser.hpp is the same as in DwarfParser.cpp and DwarfParser.hpp.


mgaudet:
This branch should be rebased to clean up the history (and at least one commit is failing IP validation).  

From the looks of it, this may fit under a single commit.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+	"DW_DLE_FNO (4) file not open",
+	"DW_DLE_IA (5) invalid argument",
+	"DW_DLE_IOF (6) I/O failure",
+	"DW_DLE_MAF (7) memory allocation failure",
+	"DW_DLE_NOB (8) not an object file or dSYM bundle not found",
+	"DW_DLE_VMM (9) dwarf format/library version mismatch",
+};
+
+/* temporary debugging methods */
+void printDie (Dwarf_Die* die, int spaces) {
```
jason-hall: Functions main(), printCU(), printDie(), etc should not be in here. If you want *minimal* debug output, use the "DEBUGPRINTF" macro (as used in DwarfScanner.cpp) and not cout. I suggest using it during normal operation, such as printing out a CU or Die name as you create it, and not in a separate printSomething() function.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ 	 * 2: inside of a file, parse to get information to populate dies, check if end of file reached
+ 	 */
+	if (0 == _startNewFile && regex_search(data, m_s, startOfFileSection)) {
+		_startNewFile = 1;
+		
+		/* Create one compilation unit context per file */
+		Dwarf_CU_Context *newCU = new Dwarf_CU_Context();
+		
+		/* Create a DIE for the CU */
+		Dwarf_Die newDie = new Dwarf_Die_s ();
```
jason-hall: Excess space after "Dwarf_Die_s" and before " ()".

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+		}
+	} else if (1 == _startNewFile) {
+		if (regex_search(data ,m_s, fileInfoRow)) {
+			/* Update the _fileID */
+			_fileID = extractFileID (m_s.prefix().str());
+			
+			/* Add the file to the list of files */
+			_fileList.push_back(strip(m_s.suffix().str(), '\n'));
+			
+			Dwarf_Attribute name = new Dwarf_Attribute_s ();
```
jason-hall: Excess space after "Dwarf_Attribute_s" and before " ()".

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ * @return bound type based on the data string
+ */
+string
+parseBound(const string data)
+{
+	string ret = "INVALID";
+
+	if ("J" == data) {
+		ret = "indeterminable";
+	} else if ('0' <= data[0] && '9' >= data[0]) {
```
jason-hall: I know you probably did not write this, but it should be: "} else if (('0' <= data[0]) && ('9' >= data[0])) {"

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+		regex typeDefExpression ("(^):t"+INTEGER+"="+"(c|d|e|g|D|k|m|w|M|\\*|\\&|V|Z|f|a)?(.*)");
+		regex constantExpression ("(^)"+NAME+":c="+"(b|c|i|r|s|C|S)?;(.*)");
+		regex structureExpression ("=s"+NUM_BYTES);
+		regex unionExpression ("=u"+NUM_BYTES);
+		regex classExpression ("=Y"+NUM_BYTES+"(?:"+CLASS_KEY+")+(.*)(\\()");
+		regex classStructure ("Y"+NUM_BYTES+"s(.*)");
+		regex classUnion ("Y"+NUM_BYTES+"u(.*)");
+		regex enumExpression ("=e"+INTEGER);
+		
+		//regex labelExpression ("(^)"+NAME+":t(-)?[0-9]+="+INTEGER+"($)");
```
jason-hall: This line should probably be either deleted or preserved in a proper comment.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+			
+		}
+	}
+	return ret;
+}
+
+static int
+parseDwarfDie(const string line)
+{
+
```
jason-hall: Excess blank line

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ *
+ * @return argument value as string
+ */
+string
+toString(const double value)
+{
+	stringstream ss;
+	ss << value;
+	return ss.str();
+}
```
jason-hall: You are missing blank line at end of file.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.hpp
```diff
@@ -0,0 +1,284 @@
...
+
+double shiftDecimal(const double value);
+
+double toDouble(const string value);
+
+int toInt(const string value);
+
+size_t toSize (const string size);
+
+string toString (const double value);
```
jason-hall: You are missing blank line at end of file.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ * Returns type of bound
+ *
+ * @param[in] data: bound declaration string
+ *
+ * @return bound type based on the data string
+ */
+string
+parseBound(const string data)
+{
+	string ret = "INVALID";
```
jason-hall: Just a generate note on error handling: 
- If a function can fail, it should have a return type of DDR_RC and return either DDR_RC_OK or DDR_RC_ERROR. Other returned data can then be by parameters.
- Function return types for the functions you create should always be checked. If DDR_RC_ERROR was returned, execution should halt and not continue to make additional function calls. Note that our coding standard is still to have only one "return" per function.
- If a function calls any functions which returns DDR_RC, then it itself can fail and it should also return a DDR_RC.

I would suggest checking some of the DwarfScanner.cpp functions for error handling style if you are unsure.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+			parsedData->members = m.suffix();
+			parsedData->format = NEW_FORMAT;
+
+			parsedData->type = stripLeading(parsedData->type, '=');
+			parsedData->type = striptrailing(parsedData->type, '(');
+			parsedData->members = striptrailing(parsedData->members, ';');
+			
+			/* Check if the class declaration is for a structure, union or class*/
+			unsigned int index = 1;
+			while ((parsedData->type).length() != index && !isalpha((parsedData->type)[index])) {
```
jason-hall: Fully parameterize 'if's:
while ((parsedData->type.length() != index) && (!isalpha(parsedData->type[index]))) {

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+	/* get the type of the data of the line being parsed */
+	Parsed_Declaration_Data *parsedDeclarationData = new Parsed_Declaration_Data();
+	
+	Dwarf_Half tag = parseDeclarationLine(line, parsedDeclarationData);
+	
+	/* Parse the line further based on the type of declaration */
+	if (DW_TAG_unknown != tag) {
+		Dwarf_Die newDie;
+		double typeID = extractTypeID(parsedDeclarationData->typeID);
+
```
jason-hall: Excess blank line

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ * @param[in]  data	  : string that needs to be split
+ * @param[in]  delimeters: what the string needs to be seperated on
+ *
+ * @return string vector containing the strings after resulting from the split operation
+ */
+str_vect
+split(const string data, char delimeters)
+{
+	str_vect elements;
+
```
jason-hall: Excess blank lines

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ * @param[in] chr: the character to be removed from str
+ *
+ * @return string after trimming char from the end of the string
+ */
+string
+striptrailing(const string str, const char chr)
+{
+	string ret = str;
+	if (!ret.empty()) {
+
```
jason-hall: Excess blank line

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+ * @param[in] value: double value that you want to move the decimal point for
+ *
+ * @return double with decimal point shifted left by length of the argument plus 1
+ */
+double
+shiftDecimal(const double value)
+{
+	double ret = value;
+	if (ret >= 1) {
+		ret = shiftDecimal (value/100);
```
jason-hall: Excess space after "shiftDecimal"

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.hpp
```diff
@@ -0,0 +1,284 @@
...
+
+string striptrailing(const string str, const char chr);
+
+double shiftDecimal(const double value);
+
+double toDouble(const string value);
+
+int toInt(const string value);
+
+size_t toSize (const string size);
```
jason-hall: Space after "toSize"

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+		
+		double ID = _fileID + typeID;
+		
+		/* Search for the die in the createdDies before creating a new one */
+		unordered_map<double, Dwarf_Die>::iterator existingEntry = createdDies.find(ID);
+		
+		if (createdDies.end() != existingEntry) {
+			newDie = existingEntry->second;
+		} else {
+			newDie = new Dwarf_Die_s();
```
jason-hall: Initialize all fields of Dwarf_Die_s either by setting them individually or using memset() or however you prefer

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,1034 @@
...
+	/* state machine processes each stage of the input
+	 * 0: has not reached the start of a new file yet
+	 * 1: reached the start of a new file, parse to get info about the file, populate _currentCU
+ 	 * 2: inside of a file, parse to get information to populate dies, check if end of file reached
+ 	 */
+	if (0 == _startNewFile && regex_search(data, m_s, startOfFileSection)) {
+		_startNewFile = 1;
+		
+		/* Create one compilation unit context per file */
+		Dwarf_CU_Context *newCU = new Dwarf_CU_Context();
```
jason-hall: Initialize all fields of Dwarf_CU_Context either by setting them individually or using memset() or however you prefer

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
jason-hall: Also initialize all fields of Dwarf_Die_s

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
jason-hall: Also initialize all fields of Dwarf_Attribute_s

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
jason-hall: It also looks like function parseBound() is unused and can be deleted?

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -705,18 +283,19 @@ parseDeclarationLine (const string data, Parsed_Declaration_Data *parsedData)
 static int
 parseDwarfDie(const string line)
 {
-	
 	int ret = DW_DLV_OK;
 	/* get the type of the data of the line being parsed */
 	Parsed_Declaration_Data *parsedDeclarationData = new Parsed_Declaration_Data();
```
jason-hall: This should probably not be a pointer. As it is, you have used "new" to allocate this, but never freed it, so it leaks memory. I suggest using this instead:
Parsed_Declaration_Data parsedDeclarationData = {0};
Dwarf_Half tag = parseDeclarationLine(line, &parsedDeclarationData);

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -555,27 +156,6 @@ isCompilerGenerated(const string data)
...
 
- /**
- * Returns type of bound
- *
- * @param[in] data: bound declaration string
- *
- * @return bound type based on the data string
- */
-string
-parseBound(const string data)
```
jason-hall: You should also remove this unused function from the header.

jason-hall:
I'd also suggest cleaning up the commit description. It doesn't need to be signed off 5 times.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -705,18 +279,15 @@ parseDeclarationLine (const string data, Parsed_Declaration_Data *parsedData)
 static int
 parseDwarfDie(const string line)
 {
-	
 	int ret = DW_DLV_OK;
 	/* get the type of the data of the line being parsed */
 	Parsed_Declaration_Data *parsedDeclarationData = new Parsed_Declaration_Data();
```
jason-hall: It seems you updated it without following my suggestion. The variable parsedDeclarationData should not be a pointer. Please see the code example in my previous comment.

twlogger: The Parsed_Declaration_Data structure has strings in it, so I can't initialize it in that way without getting `std::logic_error: basic_string::_s_construct null not valid` since strings can't be initialized with the value 0.

jason-hall: 1. I was not trying to make a point about how to initialize it. It was an example. Temporary local variables like this should not use new to allocate memory. Its unnecessary, adds excess error checking, and adds complication with freeing the memory. As is, in the case of running out of memory, your new call will return null and the program will crash when it attempts to use the variable. No null check exists in the code. Don't add null checks, just get rid of this use of new and of Parsed_Declaration_Data.

2. The Parsed_Declaration_Data structure should not even exist in this commit anyway, since the point was to populate the Dwarf structures instead of it.

3. When you do need to use new, it should be paired with delete() and not free(). free() is for malloc()'ed memory only. It also requires error checking: you should always check if the result of new/malloc is null before using it and be sure to only free variables that are not null.

tools/ddrgen/src/scanners/dwarf/DwarfFunctions.hpp
```diff
@@ -153,6 +149,9 @@ struct Dwarf_Die_s
 	Dwarf_Half _tag;
 	Dwarf_Die_s *_parent;
 	Dwarf_Die_s *_sibling;
+	#if defined(AIXPPC)
```
jason-hall: The "#if" and "#endif" lines should have no indentation

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -153,125 +153,93 @@ isCompilerGenerated(const string data)
...
-			/* Check if the class declaration is for a structure, union or class*/
-			unsigned int index = 1;
-			while ((parsedData->type.length() != index) && (!isalpha(parsedData->type[index]))) {
-				++index;
-			}
-			if (0 == isCompilerGenerated(parsedData->members)) {
-				if (regex_search(parsedData->type, m, classStructure)) {
+			/* Check if the declaration is for a class, strucutre or union */
+			if (0 == isCompilerGenerated(members)) {
+				switch(sizeAndType[indexOfFirstNonInteger]){
```
jason-hall: Missing space: "){"

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -153,125 +153,93 @@ isCompilerGenerated(const string data)
...
-			parsedData->typeID = tmp[1];
-			parsedData->type = m[0];
-			parsedData->members = m.suffix();
-
-			parsedData->type = stripLeading(parsedData->type, '=');
-			parsedData->members = stripLeading(parsedData->members, ':');
-			parsedData->members = striptrailing(parsedData->members, ';');
-			
 		}
+	} else if ("t" == declarationData.substr(0, 1)){
```
jason-hall: Perhaps consider this: 't' == declarationData.front()
Missing space: "){"

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.hpp
```diff
@@ -19,41 +19,16 @@
...
-const string BIT_OFFSET (INTEGER);
-const string CLASS_TYPEID (NON_NEG_INTEGER);
-const string ORD_VALUE (INTEGER);
-const string CLASS_KEY ("(s|u|c)");
-const string ACCESS_SPEC ("(i|o|u)"); /* MUST ALWAYS BE THERE */
-const string ANON_SPEC ("a");
-const string BASE_CLASS_OFFSET (NON_NEG_INTEGER);
-const string VIRTUAL_ACCESS_SPEC ("(v?"+ACCESS_SPEC+"?)");
-const string OPT_BASE_SPEC (VIRTUAL_ACCESS_SPEC+BASE_CLASS_OFFSET+"\\:"+CLASS_TYPEID+",?"); /* comma if we have multiple of this */
+const string START_OF_FILE = "debug     3    FILE";
```
jason-hall: As discussed: consider a more robust approach that does not require an exact amount of whitespace and column width

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -153,125 +153,93 @@ isCompilerGenerated(const string data)
...
+			ret = DW_TAG_volatile_type;
+			break;
+		case 'a': /* Array */
+			ret = DW_TAG_array_type;
+			break;
+		default:
+			ret = DW_TAG_unknown;
+			break;
+		}
+	} else if ("T" == declarationData.substr(0, 1)) {
```
jason-hall: declarationData.first()?

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -153,125 +153,93 @@ isCompilerGenerated(const string data)
...
  *
  * @param[in]  data: the string ot be parsed.
  * @param[out] tag                   : the type of the declaration of the data string
- * @param[out] parsedDeclarationData : the declaration data split into name, type and members (if applicable)
+ * @param[out] typeID                : the ID of the entity being declared
  */
 static Dwarf_Half
-parseDeclarationLine(const string data, Parsed_Declaration_Data *parsedData)
-{	
+parseDeclarationLine(const string data, double &typeID)
```
jason-hall: Our coding standard does not allow reference types. I suggest pointers instead: double *typeID

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -280,14 +248,13 @@ static int
 parseDwarfDie(const string line)
 {
 	int ret = DW_DLV_OK;
+	double typeID;
 	/* get the type of the data of the line being parsed */
-	Parsed_Declaration_Data *parsedDeclarationData = new Parsed_Declaration_Data();
-	Dwarf_Half tag = parseDeclarationLine(line, parsedDeclarationData);
+	Dwarf_Half tag = parseDeclarationLine(line, typeID);
```
jason-hall: Since our coding standard does not allow reference types, consider the equivalent without reference type arguments: Dwarf_Half tag = parseDeclarationLine(line, &typeID);

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -153,125 +153,93 @@ isCompilerGenerated(const string data)
...
-			
-			/* Split the half containing the name into another 2 halves, to get the name and typeID */
-			str_vect tmp2 = split(tmp[0], ':');
-			parsedData->name = tmp2[0];
-			parsedData->typeID = tmp2[1];
-			parsedData->type = tmp[1];
-		} else if (regex_search(data, m, constantExpression)) {
+	
+	/* split the declaration into the name and the information about the declaration */
+	size_t indexOfFirstColon = data.find_first_of(':');
```
jason-hall: Is it possible for a line to not contain ":"? If so, will this code behave if a line does not contain ":"?

twlogger: I'll add code to check that `data` contains a colon, and add documentation for the parameters being passed in.

twlogger:
@charliegracie could you please open an Eclipse CQ for this PR please? Also, I think that the previous PR https://github.com/eclipse/omr/pull/608 is no longer needed. Thanks.

twlogger:
Since all the code in this PR is in #721 anyways, I don't think we require this one (since we'd need two Eclipse CQs, rather than just one).

PR #675, opened 2017-01-16
Add OSR points after monents

ncough:
This extends OSR to add points after a monent, similar to what is done
with calls and asyncchecks. These points are used for analysis and
may result in the insertion of OSR guards. This breaks the assumption
that OSR points are always before and optionally after a node. The
code has been altered to make them both optional, but one mandatory.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -718,10 +718,10 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    }
 
 bool
-OMR::Compilation::requiresPreOSRPoint(TR::Node *node)
+OMR::Compilation::requiresLeadingOSRPoint(TR::Node *node)
    {
-   // When there is no OSR induction offset, a pre OSR point is required
-   // This currently only covers async checks as they save the state when the check is performed
+   // Without an induction offset, a leading OSR point is required
+   // This point results in analysis of liveness before the side effect has occured
    if (getOSRInductionOffset(node) == 0)
       {
       return true;
@@ -729,11 +729,10 @@ OMR::Compilation::requiresPreOSRPoint(TR::Node *node)
 
    switch (node->getOpCodeValue())
       {
-      // Monents only require an offset OSR point as they will perform OSR when executing the
-      // monitor and there is no side effect due to the monent
+      // Monents only require a trailing OSR point as they will perform OSR when executing the
+      // monitor and there is no change in liveness due to the monent
       case TR::monent: return false;
-      // Calls require pre and offset OSR points as they may have to store the original and the
-      // modified state, for example when inlining
+      // Calls require leading and trailing OSR points as liveness may change across them
       default: return true;
       }
    }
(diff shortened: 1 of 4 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -669,6 +670,8 @@ bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node)
              self()->getSymRefTab()->getNonhelperIndex(self()->getSymRefTab()->getLastCommonNonhelperSymbol()))
             potentialOSRPoint = disableGuardedCallOSR == NULL;
          }
+      else if (node->getOpCodeValue() == TR::monent)
+         potentialOSRPoint = disableMonentOSR == NULL;
```
vijaysun-omr: potentialOSRPoint = (disableMonentOSR == NULL); 

would be slightly easier to read. Could you please make this change ?

compiler/compile/OMRCompilation.cpp
```diff
@@ -702,12 +705,25 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
+         default: return 3;
+         }
+      }
    return 0;
    }
 
 bool
+OMR::Compilation::requiresPreOSRPoint(TR::Node *node)
+   {
+   return node->getOpCodeValue() != TR::monent;
```
vijaysun-omr: Could you please add a comment here explaining why monent is treated differently to async check and call in terms of a PreOSRPoint ?

compiler/compile/OMRCompilation.cpp
```diff
@@ -720,7 +720,22 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
 bool
 OMR::Compilation::requiresPreOSRPoint(TR::Node *node)
    {
-   return node->getOpCodeValue() != TR::monent;
+   // When there is no OSR induction offset, a pre OSR point is required
+   // This currently only covers async checks as they save the state when the check is performed
+   if (getOSRInductionOffset(node) == 0)
```
vijaysun-omr: Just curious why asyncchecks and monents are not in the same category as in neither requiring a pre OSR point ? Is'nt OSR really only possible once an asynccheck returns back to compiled code (same as a monent) ?

andrewcraik: The issue here is subtle - asyncchecks are always generated _before_ the side effect of the operation whose bytecode they carry so they have an induction offset of 0. When you have an induction offset of 0 you will always have a prepoint because that is the transition point. For the monent we will transition after the monent, but unlike a call we will not blow the monent into pieces where we could see the pre-osr point on the stack (calls can be inlined). Essentially, this code is saying that if the offset is 0 the prepoint is the point and if the offset is non-zero we need the prepoint if we may need the pre-bytecode execution state as part of setting up a transition which is true for a call, but not for a monent.

All very subtle but I think this is probably the simplest way to encode this notion of do we need two osr points or one.

vijaysun-omr: I see, so the asyncheck is somewhat special because it's not actually in the bytecodes of the method. 

monent and calls are both in the method bytecodes and behave in the ways we would expect : we have an OSR point after we return from the monent and for the call is both before and after to account for the inlining possibility and the place of the actual OSR respectively.

PR #677, opened 2017-01-17
Split Runtime.hpp into omr and J9 specific parts

mpirvu:
The parts from omr/compiler/runtime/Runtime.hpp that are J9 specific will
be moved into a new file trj9/runtime/Runtime.hpp. This new file will include
compiler/runtime/Runtime.hpp header


mpirvu:
Changes are ready to go. The counterpart GitLab changes are here: https://gitlab-polyglot.hursley.ibm.com/jit/tr.open/merge_requests/953

PR #678, opened 2017-01-17
Fix Virtual Guard Head Merger bug

cathyzhyi:
Fix Virtual Guard Head Merger bug

The intention of the original code is to iterate *forward* from the
start of a block. Fix the code to go forward instead of backward.


mgaudet:
This will need a larger commit message (see [CONTRIBUTING.md](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md))

andrewcraik:
@cathyzhyi LGTM once the commit message is fixed

cathyzhyi:
@andrewcraik @mgaudet thanks! Fixed already.

vijaysun-omr:
The change is clearly better than what's there already since we were not iterating correctly over trees. Merging.

PR #680, opened 2017-01-18
addition of new constructor to enable growsUp and startinStackOffset

jduimovich:
The virtual functions growsUp() and stackPtrStartingOffset() were not being called in subclasses because they are being called in the constructor. 

The fix is to add two new parameters to the constructor to set the values for stack direction and offset.  A constructor matching the old constructor sets the values to the old defaults so old code continues to work as before. 

make operandstacktests still passed all tests.

### DIFF
--- a/compiler/ilgen/VirtualMachineOperandStack.cpp
+++ b/compiler/ilgen/VirtualMachineOperandStack.cpp
@@ -28,14 +28,16 @@
 namespace OMR
 {
 
-VirtualMachineOperandStack::VirtualMachineOperandStack(TR::MethodBuilder *mb, int32_t sizeHint,
-   TR::IlType *elementType, OMR::VirtualMachineRegister *stackTopRegister)
+   VirtualMachineOperandStack::VirtualMachineOperandStack(TR::MethodBuilder *mb, int32_t sizeHint,
+   TR::IlType *elementType, OMR::VirtualMachineRegister *stackTopRegister, bool growsUp, int32_t stackInitialOffset)
    : VirtualMachineState(),
    _mb(mb),
    _stackTopRegister(stackTopRegister),
    _stackMax(sizeHint),
    _stackTop(-1),
-   _elementType(elementType)
+   _elementType(elementType),
+   _pushAmount(growsUp ? +1 : -1),
+   _stackOffset(stackInitialOffset)
    {
    int32_t numBytes = _stackMax * sizeof(TR::IlValue *);
    _stack = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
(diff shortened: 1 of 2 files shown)


compiler/ilgen/VirtualMachineOperandStack.hpp
```diff
@@ -80,8 +80,31 @@ class VirtualMachineOperandStack : public VirtualMachineState
...
+    * @brief public constructor, must be instantiated inside a compilation because uses heap memory
+    * @param mb TR::MethodBuilder object of the method currently being compiled
+    * @param sizeHint initial size used to allocate the stack; will grow larger if needed
+    * @param elementType TR::IlType representing the underlying type of the virtual machine's operand stack entries
+    * @param stackTop previously allocated and initialized VirtualMachineRegister representing the top of stack
+    * @param growsUp to configure virtual machine stack growth direction, set to true if virtual machine stack grows towards larger addresses, false otherwise 
+    * @param stackInitialOffset to configure virtual machine stack stack offset 
+    * set to the difference in elements between initial stack pointer and actual bottom of stack
+    * Some stacks Push by incrementing stack pointer then storing, some by storing and then
+    * incrementing stack pointer. In the first case, stackPtrStartingOffset() should return -1
```
mstoodle: please change the comments to correctly refer to the stackInitialOffset parameter's value rather than the old stackPtrStartingOffset() method's return value. Otherwise, looks good to me.

mgaudet:
Probably can also squash this branch down to one commit. :)

compiler/ilgen/VirtualMachineOperandStack.hpp
jduimovich: Ok, after a couple commits, proof-reading, fixed the comments to the correct name and clarified the default values for the old constructor.

jduimovich:
Squashed to a single commit.

mstoodle:
Thanks for getting to this one before I did, @jduimovich ! I'll merge once the CI build passes.

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -44,9 +46,13 @@ VirtualMachineOperandStack::VirtualMachineOperandStack(TR::MethodBuilder *mb, in
...
    // to recreate the stack as the interpreter would have
    mb->Store("OperandStack_base", stackTopRegister->Load(mb));
+   }
 
-   _pushAmount = (growsUp()) ? +1 : -1;
-   _stackOffset = stackPtrStartingOffset();
+   // compatibility with prior constructor. pass same defaults as before for growup and stackOffset
+   VirtualMachineOperandStack::VirtualMachineOperandStack(TR::MethodBuilder *mb, int32_t sizeHint,
+   TR::IlType *elementType, OMR::VirtualMachineRegister *stackTopRegister)
+   : VirtualMachineOperandStack(mb, sizeHint,elementType,stackTopRegister, true, -1)
```
mstoodle: Compiler level on linux x86 won't accept constructor delegation like you've done here :( .

You could use default parameters in the header prototype as an alternative.

jduimovich: Yeah, my Mac has an excessively fresh C++ I guess but when I tested, my linux machine seems to be similarly afflicted with too new a gcc.

Fixed to used defaults for parameters.

PR #682, opened 2017-01-18
Cleanup INC/DEC r16/r32 instruction

0dvictor:
[X86-64 Only] Existing heuristic to handle unencodable INC/DEC r16/r32
is confusing and inefficient. Cleanup the code by supplying the correct
opcode on X86-64 so that no heuristics are needed anymore.


andrewcraik:
@0dvictor Can you please removed the signed off by line from the commit message?

Other than the commit message I think this is a good piece of clean-up.

mgaudet:
We can't merge commits that aren't signed off. (See [CONTRIBUTING.md](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#legal-considerations)) I think this is actually mergable right now, given the above LGTM :)

andrewcraik:
correction - I mean the pull request message not the commit message.

andrewcraik:
@mgaudet pointed out that the text of the PR doesn't merge into the history so ignore my request.

LGTM

PR #685, opened 2017-01-19
Fix hold time and clock skew comparison

babsingh:
Presently, we evaluate hold time average when clock skew is greater than
the hold time sample. If clock skew is greater than the hold time
sample, then the hold time sample is indeterminate. Evaluating hold time
average using an indeterminate hold time sample will yield an inaccurate
hold time average. Inaccurate hold time average will impact the decision
making process of the adaptive spin heuristic: spin or do not spin. When
clock skew is less than the hold time sample, then the hold time sample
can be considered valid. We must evaluate hold time average using a
valid hold time sample. In order to fix the implementation, the
comparison between clock skew and hold time has been updated.

### DIFF
--- a/thread/common/threaddef.h
+++ b/thread/common/threaddef.h
@@ -542,7 +542,7 @@ enum {
 				if ((self)->tracing->pause_count == (monitor)->tracing->enter_pause_count) { \
 					omrtime_delta_t holdTime = GET_HIRES_CLOCK() - (monitor)->tracing->enter_time; \
 					if (holdTime > 0) { \
-						if ((0 == (self)->library->clock_skew) || ((omrtime_t)holdTime < (self)->library->clock_skew)) { \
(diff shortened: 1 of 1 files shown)


mstoodle:
@babsingh the motivation sounds reasonable. I'm just curious if you can describe what kind of testing you did to measure the impact of this change, or if some particular scenario (that you can talk about in the open :) ) led you to identify this particular problem and that this change fixed?

babsingh:
@mstoodle I found the bug while reading the code. During initialization, clock skew is set to 0. So, we never utilize the incorrect implementation. Only a user can set clock_skew to a non-zero value by setting omrthread_global("clockSkewHi") via a prospective command line option. The impact will vary depending upon how accurately a user can measure and set the clock skew value. I did not get to measure the impact of this change.

charliegracie:
Good catch

PR #686, opened 2017-01-19
Add Runtime-Helper enum entities for ECC methods on Power

hjooybar:
Add entities in Runtime-Helper enum to handle Elliptic curve cryptography (ECC) acceleration on Power processors.
ECC algorithm contains heavy amount of arithmetic operations on big integers. To accelerate ECC, jit recognizes basic methods of big integer arithmetic operations and replaces them with optimized assembly.
Four helper functions are added to the Runtime-Helper enum. These helpers are called by jit to accelerate 256-bit multiplication, addition and subtraction (modular or simple).


0xdaryl:
Please shorten your commit title as per the CG.

PR #687, opened 2017-01-19
Fix bug in rematerialization safety logic

andrewcraik:
The RematTools server to allow you to try and recompute a store on a different control flow path. This is used in the inliner to do privatized inliner argument rematerialization in the blocks on the taken side of virtual guards. The RematTools include a bunch of safety checking logic to ensure it is safe to remat a store. There is a bug in this remat logic that could cause some nodes that should be checked to be skipped. This patch fixes this problem and adds additional helpful trace to the RematTools.

### DIFF
--- a/compiler/optimizer/RematTools.cpp
+++ b/compiler/optimizer/RematTools.cpp
@@ -190,13 +190,23 @@ void RematTools::walkNodesCalculatingRematSafety(TR::Compilation *comp,
    {
    for (uint16_t i = 0; i < currentNode->getNumChildren(); ++i)
       {
-      if (visitedNodes.ValueAt(currentNode->getGlobalIndex()))
-         return;
+      if (visitedNodes.ValueAt(currentNode->getChild(i)->getGlobalIndex()))
+         continue;
 
-      visitedNodes[currentNode->getGlobalIndex()] = true;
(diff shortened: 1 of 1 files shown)


compiler/optimizer/RematTools.cpp
```diff
@@ -190,21 +190,35 @@ void RematTools::walkNodesCalculatingRematSafety(TR::Compilation *comp,
...
 
       walkNodesCalculatingRematSafety(comp, currentNode->getChild(i),
             scanTargets, enabledSymRefs, unsafeSymRefs, trace, visitedNodes);
       }
 
+   if (trace)
+      {
+      traceMsg(comp, "Remat Safety Walk visiting n%dn\n", currentNode->getGlobalIndex());
+      traceMsg(comp, "Enabled symrefs: ");
+      (*comp) << enabledSymRefs;
```
fjeremic: Cool! Did not know compilation had overloaded the stream operator.

mgaudet:
I think you have a typo in your commit message ("The RematTools serve**r**") :smile:

andrewcraik:
Fixed

vijaysun-omr:
Looks like a clear bug fix; we were iterating over the children of the node but checking if the node was visited every time rather than the children.

PR #688, opened 2017-01-20
Validated calls to getHighGlobalRegisterNumber

R-Santhir:
On 64 bit machines, the getHighGlobalRegisterNumber call is not valid nor
needed. On 32 bit machines, a register dependency node's children must
be checked to see if they are of the long type, before performing the
getHighGlobalRegisterNumber call. If the register dependency node has
a child without a type, the node's child's child will be checked.

### DIFF
--- a/compiler/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/codegen/OMRTreeEvaluator.cpp
@@ -147,6 +147,26 @@ bool OMR::TreeEvaluator::instanceOfOrCheckCastNeedSuperTest(TR::Node * node, TR:
    return false;
    }
 
+TR_GlobalRegisterNumber 
+OMR::TreeEvaluator::getHighGlobalRegisterNumberIfAny(TR::Node *node, TR::CodeGenerator *cg)
+   {
+    //No need for register pairs in 64-bit mode
+    if (TR::Compiler->target.is64Bit())
+        return -1;
+
+    //if the node itself doesn't have a type (e.g passthrough) we assume it has a child with a type
+    //However we keep track of the initial node as it will contain the register information.
+    TR::Node *rootNode = node;
+
+    while (node->getType() == TR::NoType) {
+        node = node->getFirstChild();
+        TR_ASSERT(node, "The node should always be valid while looking for a Child with a type");
+    }
+    TR_ASSERT(node->getType() != TR::NoType, "Expecting node %p, to have a specific type here", node);
+
+    //Only need a register pair if the node is a 64bit Int
+    return node->getType().isInt64() ? rootNode->getHighGlobalRegisterNumber() : -1;
+   }
 
 // Returns n where value = 2^n.
 //   If value != 2^n, returns -1
--- a/compiler/p/codegen/OMRRegisterDependency.cpp
+++ b/compiler/p/codegen/OMRRegisterDependency.cpp
@@ -34,6 +34,7 @@
 #include "codegen/RegisterDependency.hpp"
 #include "codegen/RegisterDependencyStruct.hpp"  // for RegisterDependency
 #include "codegen/RegisterPair.hpp"              // for RegisterPair
+#include "codegen/TreeEvaluator.hpp"             // for getHighGlobalRegisterNumberIfAny()
 #include "compile/Compilation.hpp"               // for Compilation
 #include "control/Options.hpp"
 #include "control/Options_inlines.hpp"
@@ -156,7 +157,7 @@ OMR::Power::RegisterDependencyConditions::RegisterDependencyConditions(
 
       if (reg!=NULL /* && reg->getKind()==TR_GPR */)
 	 {
-	 if (child->getHighGlobalRegisterNumber() > -1)
+	 if (TR::TreeEvaluator::getHighGlobalRegisterNumberIfAny(child, cg) != -1)
 	    numLongs++;
 	 }
       }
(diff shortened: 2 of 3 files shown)


R-Santhir:
@ymanton, @zl-wang can you guys take a look?

zl-wang:
Looks good to me.

mgaudet:
Is `getHighGlobalRegisterNumber` invalid in all contexts, or only `RegisterDependency` calculation? -- _ie_ should the function be asserting in 64 bit contexts?

R-Santhir:
I'm fairly certain that it has no use in 64bit contexts, @ymanton / @zl-wang could confirm that. 
 Given this, it would make sense to have it assert.

zl-wang:
Looking at how it is used in all codegen(s) and what I remembered,  it was guaranteed to be valid in GRA context (valid in the sense:  it must be -1 if the node doesn't need regPair or fails to be assigned with a highReg).  But this guarantee silently changed I don't know when.  It is now valid if and only if we are dealing with 32bit codegen (potentially needing highReg) and the type is Int64.

ymanton:
The more I look at this patch the more I dislike the fact that it repeats the same check (3 times here, another time downstream). It feels like the sort of functionality that should be wrapped in a helper function of some sort.

In addition, it looks like `x`, `z`, and `arm` all suffer from the same original bug in that they call `getHighGlobalRegisterNumber()` without checking the target arch or child type, so if we can no longer assume that `getHighGlobalRegisterNumber()` will return a valid global register number or `-1` then we should probably have a helper function in a common place and use it everywhere it's needed.

It would also be nice to figure out how this broke.

zl-wang:
Notified the relevant parties of this bug ...

R-Santhir:
I've added a helper function, ```getHighglobalRegisterNumberIfAny()``` inside ```OMRTreeEvaluator``` that performs the checks I was doing before, and if they are successful calls ```getHighGlobalRegisterNumber()```. @ymanton mentioned this would be useful, as other platforms can use the helper as well if needed.

compiler/codegen/OMRTreeEvaluator.cpp
```diff
@@ -147,6 +147,24 @@ bool OMR::TreeEvaluator::instanceOfOrCheckCastNeedSuperTest(TR::Node * node, TR:
...
+TR_GlobalRegisterNumber 
+OMR::TreeEvaluator::getHighGlobalRegisterNumberIfAny(TR::Node *node, TR::CodeGenerator *cg)
+   {
+    //No need for register pairs in 64-bit mode
+    if (TR::Compiler->target.is64Bit())
+        return -1;
+
+    //if the node itself doesn't have a type (e.g passthrough) we assume it has a child with a type
+    //However we keep track of the initial node as it will contain the register information.
+    TR::Node *topmost_node = node;
```
ymanton: Try calling it `root` or something perhaps. At least stick to camelCase like the rest of the code.

R-Santhir: I changed it to `rootNode`

ymanton:
Looks good. Thanks.

R-Santhir:
@0xdaryl - I've run pBuilds in both Toronto and Ottawa, and they have had no failures, can you look at this when you have a chance?

compiler/codegen/OMRTreeEvaluator.cpp
```diff
@@ -147,6 +147,24 @@ bool OMR::TreeEvaluator::instanceOfOrCheckCastNeedSuperTest(TR::Node * node, TR:
...
+    //No need for register pairs in 64-bit mode
+    if (TR::Compiler->target.is64Bit())
+        return -1;
+
+    //if the node itself doesn't have a type (e.g passthrough) we assume it has a child with a type
+    //However we keep track of the initial node as it will contain the register information.
+    TR::Node *rootNode = node;
+
+    while (node->getType() == TR::NoType)
+        node = node->getFirstChild();
```
0xdaryl: Are you certain this will always find a node with a type by traversing the first child?  I mean, it seems reasonable to assume and I can't think of a counter example, but the loop proceeds always assuming its going to find a valid first child (i.e., no check for `node` null-ness).

compiler/codegen/OMRTreeEvaluator.cpp
```diff
@@ -147,6 +147,24 @@ bool OMR::TreeEvaluator::instanceOfOrCheckCastNeedSuperTest(TR::Node * node, TR:
...
+    TR_ASSERT(node && node->getType() != TR::NoType, "Expecting node %p, to have a specific type here", node);
```
0xdaryl: This check for `node` in this assert should never happen because the loop above would have crashed trying to dereference it.

compiler/p/codegen/OMRRegisterDependency.cpp
```diff
@@ -179,15 +180,16 @@ OMR::Power::RegisterDependencyConditions::RegisterDependencyConditions(
       TR::RealRegister::RegNum regNum = (TR::RealRegister::RegNum)cg->getGlobalRegister(child->getGlobalRegisterNumber());
 
       TR::RealRegister::RegNum highRegNum;
+      TR_GlobalRegisterNumber validHighReg = TR::TreeEvaluator::getHighGlobalRegisterNumberIfAny(child, cg);
```
0xdaryl: I find the use of the suffix `Reg` on the variable name confusing in this context.  In other contexts in this function and file it is used to hold `TR::Register`s.  Perhaps `validHighRegNum` would be a better name.

compiler/p/codegen/OMRRegisterDependency.cpp
```diff
@@ -179,15 +180,16 @@ OMR::Power::RegisterDependencyConditions::RegisterDependencyConditions(
...
          {
-         highRegNum = (TR::RealRegister::RegNum)cg->getGlobalRegister(child->getHighGlobalRegisterNumber());
 
+         highRegNum = (TR::RealRegister::RegNum)cg->getGlobalRegister(validHighReg);
          TR::RegisterPair *regPair = reg->getRegisterPair();
          TR_ASSERT(regPair, "assertion failure");
-	 highReg = regPair->getHighOrder();
-	 reg = regPair->getLowOrder();
+         highReg = regPair->getHighOrder();
+	     reg = regPair->getLowOrder();
```
0xdaryl: Fix indenting.

compiler/p/codegen/OMRRegisterDependency.cpp
R-Santhir: That makes sense, I've changed it to what you've suggested

compiler/codegen/OMRTreeEvaluator.cpp
R-Santhir: You're right, I've moved a separate assert into the while loop, this address the other concern as well, there should not be a NULL that is encountered when looking for a child with a Type

0xdaryl:
Will merge after CI passes.

gita-omr:

PR #690, opened 2017-01-20
Only perform bound checks on deque access in DEBUG or ASSUME

ncough:
To improve performance in production, the accesses for deque
now map to std::deque accesses directly, rather than performing at()
as this incurs a performance penalty due to bound checks.


andrewcraik:
LGTM

vijaysun-omr:
Thanks Nick. I am going to merge this now since the checks had passed originally and Nick only added comments since then.

PR #692, opened 2017-01-20
Fix for incorrect MoveDataType

0dvictor:
TR::Address and TR_GPR should correspond to TR_MovDataTypes::Int4 on X86-32,
while previous code always returns TR_MovDataTypes::Int8


andrewcraik:
These look like legitimate oopses - LGTM

vijaysun-omr:
Fix datatypes based on bit-ness. Merging.

PR #693, opened 2017-01-22
Enablement for more Ruby optimization

mgaudet:
* Add support for a 'hot' optLevel in Ruby.
* Add two new instance variable access helpers. 
* Stop asserting 'notImplemented' in getClassFromNewArray Type. Default NULL return is acceptable!


mgaudet:
Fixed

PR #694, opened 2017-01-23
Stop inlining MethodHandles in fsd mode

cathyzhyi:
We need to stop inlining MethodHandle in fsd because those trees are
generated by JIT out of thin air and dont have corresponding
bytecode for interpreter to transit from. Note that this changeset
only solves part of the problem and there are still trees generated
from ilgen without inlining that will be fixed later.

### DIFF
--- a/compiler/optimizer/OMROptimizer.hpp
+++ b/compiler/optimizer/OMROptimizer.hpp
@@ -114,7 +114,8 @@ enum
    IfEnabledAndMoreThanOneBlock,
    IfEnabledAndMoreThanOneBlockMarkLastRun,
    IfAOTAndEnabled,
-   IfMethodHandleInvokes, // JSR292: Extra analysis required to optimize MethodHandle.invoke
+   IfMethodHandleInvokes, //JSR292: Extra analysis required to optimize MethodHandle.invoke
+   IfMethodHandleInvokesNoDebug, // JSR292 optimizations that should only apply in non-debug mode
    IfNotQuickStart,
    IfFullInliningUnderOSRDebug,
    IfNotFullInliningUnderOSRDebug,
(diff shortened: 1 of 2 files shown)


andrewcraik:
@cathyzhyi I'm not sure I follow this - this code does not seem to be specific to the inliner - I don't see where the inliner specific part of the check is located. Could you point it out?

cathyzhyi:
@andrewcraik Hi Andrew, I am disabling the entire methodHandleInvokeInliningGroup. The flag IfMethodHandleInvokes is used for this group and an globalCopyPropagation that should be performed after methodHandleInvokeInliningGroup.  By adding the fsd checking condition here will disable both.

andrewcraik:
For cleanliness I think we need another flag IfMethodHandleInvokesAndNoDebug to capture this case otherwise the behaviour is surprising relative to the other flags - this is subtle and non-obvious as currently implemented.

cathyzhyi:
@andrewcraik Added a new flag as suggested. Thanks!

compiler/optimizer/OMROptimizer.hpp
```diff
@@ -114,7 +114,8 @@ enum
    IfEnabledAndMoreThanOneBlock,
    IfEnabledAndMoreThanOneBlockMarkLastRun,
    IfAOTAndEnabled,
-   IfMethodHandleInvokes, // JSR292: Extra analysis required to optimize MethodHandle.invoke
+   IfMethodHandleInvokes, //JSR292: Extra analysis required to optimize MethodHandle.invoke
+   IfMethodHandleInvokesNoDebug, // JSR292 optimizations that should only apply in non-debug mode
```
andrewcraik: There is a convention of having an And in the name when there is a union of two conditions - eg AOTAndEnabled. Could you please add the And in?

cathyzhyi:
closing this as we might not need it anymore with the new design.

PR #697, opened 2017-01-24
Revert "Fix for incorrect MoveDataType"

cathyzhyi:
This reverts commit e28a9b987bfcf9a1d339f4d8fb3341520fc65469. The original change exposes several failures in tests and it is reverted for now until we have a better understanding of the problem.


0xdaryl:
Could you provide more details on this revert please?  Why is it necessary, what problem did it uncover, and why did it slip past our review and testing?  Thanks.

cathyzhyi:
@0xdaryl Thanks for the comment. I added more detail about the reason we need to revert it.

Issue #699, opened 2017-01-24
Correctly set `PLATFORM` for JIT tests


mgaudet:
In `.travis.yml` we have the following stanza

```
env:
  - SPEC=osx_x86-64
  - SPEC=linux_x86-64
  - SPEC=linux_x86-64_cmprssptrs
  - SPEC=linux_x86
```

However, in the case of `SPEC=linux_x86` when you look at the actual build log, you'll see 

```
../../fvtest/compilertest/build/platform/common.mk:12: PLATFORM not set. Guessing 'amd64-linux64-gcc'
```

This is because the JIT makefiles haven't been hooked up to use SPEC, and use their own indicator of build configuration. When `PLATFORM` isn't set, a script `guessPlatform` is invoked, which as you might suspect, guesses what the host machine is, and sets `PLATFORM` appropriately. 

In this case, despite `SPEC` being set for 32-bit linux, we end up testing the 64 Bit test compiler. Which means the 32 Bit test compiler hasn't been exercised by Travis yet. 

This lead to an issue where https://github.com/eclipse/omr/pull/692 looked clean, but actually was `ASSERT`ing in a 32 bit build; however, this was merged because its Travis build was clean, leading to the revert in https://github.com/eclipse/omr/pull/697

mgaudet:
Fixed by #702

PR #702, opened 2017-01-24
Explicitly set PLATFORM for Travis builds

mgaudet:
The JIT makefiles haven't been hooked up to use SPEC, they use their own
indicator of build configuration PLATFORM. When PLATFORM isn't set, a
script guessPlatform is invoked, which as you might suspect, guesses
what the host machine is, and sets PLATFORM appropriately.

In this case, despite SPEC being set for 32-bit linux, we end up testing
the 64 Bit test compiler in Travis builds.

This hardcodes the PLATFORM for our build matrix.

Fixes Issue #699

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@
 language: c++
 sudo: false
 
-os:
-  - linux
-  - osx
-
 # https://docs.travis-ci.com/user/caching/
 cache: ccache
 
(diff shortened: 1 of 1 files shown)


mgaudet:
This actually shouldn't pass the Travis builds until it has been rebased on top of #697 -- but I wanted this here to a) verify we actually build the platforms we think we do b) make sure no one else is trying to fix #699 :D

mgaudet:
Somehow this spawned 6 builds instead of 4....

mgaudet:
Rebased on top of #697 -- should pass. 

This changes the `.travis.yml` to explicitly list the matrix contents, as opposed to dancing through hoops to get the excludes right 😄

mgaudet:
Delivering this is blocked by #703

mgaudet:
Whoops. Last rebase was done on a stale clone. 

Will fix when I get to work, which has my non-stale clone.

Issue #703, opened 2017-01-25
32 Bit JIT Builder Travis Builds fail


mgaudet:
In #702 We are attempting to explicitly set PLATFORM to ensure we produce and test a 32 bit JIT when SPEC is set to a 32 bit linux configuration. 

However, [the Travis builds continues to fail in this scenario](https://travis-ci.org/eclipse/omr/builds/195009170) because of a mismatch between the architecture of the JITBuilder archive. 

```
-Wl,-z,origin,-rpath,\$ORIGIN,--disable-new-dtags,-rpath-link,../..
make[2]: Leaving directory `/home/travis/build/eclipse/omr/fvtest/vmtest'
make -C jitbuilder/release all
make[2]: Entering directory `/home/travis/build/eclipse/omr/jitbuilder/release'
g++ -o Call.o -g -std=c++0x -O2 -c -fno-rtti -fPIC -I./include/compiler -I./include src/Call.cpp
g++ -g -fno-rtti -o call Call.o -L. -ljitbuilder -ldl
/usr/bin/ld: skipping incompatible ./libjitbuilder.a when searching for -ljitbuilder
/usr/bin/ld: cannot find -ljitbuilder
collect2: ld returned 1 exit status
make[2]: *** [call] Error 1
make[2]: Leaving directory `/home/travis/build/eclipse/omr/jitbuilder/release'
make[1]: *** [jitbuilder/release] Error 2
make[1]: Leaving directory `/home/travis/build/eclipse/omr'
make: *** [tests] Error 2
```

mgaudet:
Also, it turns out in internal testing, 32 Bit JitBuilder tests were also never hooked up.

mgaudet:
Closing this -- Fixed by #707

Issue #705, opened 2017-01-25
32 Bit Jit Builder Support: Makefile Changes


mgaudet:
`jitbuilder/release/Makefile` will need to have edits made to support 32 bit builds. 

- [ ] `LDFLAGS`
- [ ] `CXXFLAGS`

Issue #704, opened 2017-01-25
32 Bit JIT Builder Support: TR::IA32SystemLinkage::buildIndirectDispatch


mgaudet:
Solving #703 found that 32 bit dispatch on x86 isn't done yet

```
 Step 5: compile ComputedCall method builder
Assertion failed at ../compiler/x/i386/codegen/IA32SystemLinkage.cpp:227: false
	TR::IA32SystemLinkage::buildIndirectDispatch is not supported.
```

PR #707, opened 2017-01-25
Only enable JITBuilder on 64-bit SPECs.

mgaudet:
There are a number of outstanding issues before 32-Bit JITBuilder can be
enabled:

- Issue #705
- Issue #704

However, this should resolve Issue #703, and unblock Pull #702, which will allow us to do 32 Bit opcode testing in Travis .


mgaudet:
I'm going to need to verify by hand that JITBuilder tests are disabled for the `SPEC=linux_86` build.

fjeremic:
> The first line should be less than 50 characters, sentence case, and does not end in a period.

Nit: https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines

mgaudet:
That period. Trips me up every time.

Fixed. :smile: 

I've also verified that the 32 Bit JitBuilder tests aren't running.

PR #709, opened 2017-01-25
Clean up aliasing interface

Leonardo2718:
This commit does some minor clean ups to the aliasing interface by:

- combining separate if statements into one when there is no functional need for the separation
- removing functions for which there exists another function that does the same thing

This clean up should simplify later work to be done on the aliasing interface.


vijaysun-omr:
The code changes look safe to me. Please fix the commit message typo and I can merge this.

Leonardo2718:
@vijaysun-omr I've fixed the type.

vijaysun-omr:
Only change was to the commit message and the checks had passed earlier. I am merging even though checks are in progress now.

PR #710, opened 2017-01-25
Add option to make inliner only inline natives

cathyzhyi:
Add this option to help performance investigation when we only care about natives that got inlined.


vijaysun-omr:
I'll wait for the checks to pass before merging this in.

vijaysun-omr:
Useful option to have to control the kind of inlining we do. No change in default behavior. Merging.

PR #711, opened 2017-01-26
Enable aliasing support for unions

xliang6:
Add a hash table of bitvectors in symbol reference table class. The hash
table is keyed by a symbol reference's reference number. Each bitvector
keeps track of what other symbol references it can be aliased to. Add
code to include the bitvector in the shared shadow alias calculation.

### DIFF
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -117,7 +117,8 @@ OMR::SymbolReferenceTable::SymbolReferenceTable(size_t sizeHint, TR::Compilation
      _hasUserField(false),
      _aggregateShadowSymbolMap(8, comp->allocator("SymRefTab")),
      _aggregateShadowSymbolReferenceMap(8, comp->allocator("SymRefTab")),
-     _registerSymbolRefs(NULL)
+     _registerSymbolRefs(NULL),
+     _sharedAliasMap(NULL)
    {
    _numHelperSymbols = TR_numRuntimeHelpers + 1;;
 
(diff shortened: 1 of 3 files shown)


compiler/compile/OMRSymbolReferenceTable.cpp
```diff
@@ -1899,3 +1900,57 @@ TR::SymbolReference * & OMR::SymbolReferenceTable::element(CommonNonhelperSymbol
...
+void OMR::SymbolReferenceTable::makeSharedAliases(TR::SymbolReference *sr1, TR::SymbolReference *sr2)
+   {
+   int32_t symRefNum1 = sr1->getReferenceNumber();
+   int32_t symRefNum2 = sr2->getReferenceNumber();
+   TR_BitVector *aliases1 = NULL;
+   TR_BitVector *aliases2 = NULL;
+
+   if (_sharedAliasMap == NULL)
+      {
+      _sharedAliasMap = new (comp()->trHeapMemory()) AliasMap(std::less<int32_t>(), getTypedAllocator<std::pair<int32_t, TR_BitVector *> >(comp()->allocator()));
```
lmaisons: If I'm reading this correctly and this is per symbol reference table, why not make this a real member instead of a pointer allocated on demand?

Also, you shouldn't need "getTypedAllocator" anymore.  If you pass comp()->allocator() directly, it should get converted into the right kind of typed_allocator automatically, thanks to a conversion operator I added to CS2::shared_allocator.

compiler/compile/OMRSymbolReferenceTable.cpp
```diff
@@ -1899,3 +1900,57 @@ TR::SymbolReference * & OMR::SymbolReferenceTable::element(CommonNonhelperSymbol
...
+       AliasMap::iterator iter2 = _sharedAliasMap->find(symRefNum2);
+       if (iter2 != _sharedAliasMap->end())
+          aliases2 = iter2->second;
+       }
+
+    if (aliases1 == NULL)
+       {
+       aliases1 = new (comp()->trHeapMemory()) TR_BitVector(self()->getNumSymRefs(), comp()->trMemory(), heapAlloc);
+       aliases1->empty();
+       _sharedAliasMap->insert(std::make_pair(symRefNum1, aliases1));
```
lmaisons: I'm not sure how well TR_BitVector copy constructs, but if these fields are canonically part of the map, would it make sense for them physically reside there? In other words, have the map contain actual TR_BitVector objects instead of pointers?

compiler/compile/OMRSymbolReferenceTable.cpp
Leonardo2718: I believe `_sharedAliasMap` is allocated on demand to minimize the footprint overhead when it's not needed.

compiler/compile/OMRSymbolReferenceTable.cpp
Leonardo2718: In the code that I've looked at, `TR_BitVector`s are almost always passed around as pointers or as references. This may only be for performance reasons though.

Although I would generally agree that it makes sense for them to physically be in the container, in this case, I think using pointers may be "safer" in the sense that it is more consistent with other uses of `TR_BitVector`.

compiler/compile/OMRSymbolReferenceTable.cpp
xliang6: Yes, `_sharedAliasMap` is a pointer to minimize its memory footprint to current consumers of symbol reference table, i.e. Java.  It is created on demand by JitBuilder when union types are created.

I will remove "getTypedAllocator", thanks.

compiler/compile/OMRSymbolReferenceTable.cpp
xliang6: `_sharedAliasMap` can potentially hold as many bit vectors as there are symbol references in the symbol reference table. But we only need to create a bit vector when a symbol reference has shared aliases, i.e. JitBuilder creating union types.

xliang6:
@vijaysun-omr Please review. Thanks.

compiler/il/Aliases.cpp
```diff
@@ -797,6 +797,10 @@ OMR::SymbolReference::setSharedShadowAliases(TR_BitVector * aliases, TR::SymbolR
       while (symRef = i.getNext())
          if (symRef->getSymbol() == self()->getSymbol())
             aliases->set(symRef->getReferenceNumber());
+
+      // include symbol reference's own shared alias bitvector
```
vijaysun-omr: This code is under a condition

void
OMR::SymbolReference::setSharedShadowAliases(TR_BitVector * aliases, TR::SymbolReferenceTable * symRefTab)
   {
   if (self()->reallySharesSymbol() && !_symbol->isUnsafeShadowSymbol())

Can you please confirm that these conditions will be satisfied in the cases you are targetting with this PR, i.e. unions ? Otherwise your code won't be used.

xliang6: Yes, the conditions will be met. JitBuilder union symbol references will be created as regular shadow symbol references (i.e. not unsafe) and setReallySharesSymbol() will be called on them.

Leonardo2718: Just for the record, I have tested this patch with some un-pushed changes and can confirm that it works as expected for JitBuilder unions.

vijaysun-omr: Okay that's good in the sense that aliasing between JitBuilder generated sym refs for unions will work (meaning that languages whose IL gen uses JitBuilder will work fine). 

One last question : Is it worth considering the setting of really shares symbol flag inside of makeSharedAliases so that other languages that call the general makeSharedAliases API on two sym refs actually get the answer they expect when they get the aliases ?

xliang6: The current convention appears to be calling setReallySharesSymbol() to set the flag right after a symbol reference is created, so setting the flag in makeSharedAliases might not make the most sense if a symbol reference wasn't created initially to share the underlying symbol. However, I can add an assert in makeSharedAliases when a symbol reference is passed in without its flag being set to address the case you brought up?

vijaysun-omr: Okay, I see your point about setting the flag being something expected at sym ref creation time currently (though it seems like a convention that could be relaxed if we wanted since it's not very pervasive in the codebase). My point was : why not do everything needed to make a sym ref an alias of the other if the makeSharedAliases api was called ?  Having said that, at least for union sym refs (the use case you are going after here) one should know that sharing is possible at sym ref creation time. So, maybe an assert is good enough for now to raise the red flag if the api was used without "really shares symbol" being set in some future scenario.

vijaysun-omr:
The asserts seem to have been added. Merging.

PR #714, opened 2017-01-27
Add vector related macro instructions for Power 

hjooybar:
A new set of macros and instruction encoding has been added for Power architecture.
Here is a list of newly added instructions:
```
vspltisb 	Vector Splat Immediate Signed Byte
vshasigmad 	Vector SHA-512 Sigma Doubleword
vshasigmaw 	Vector SHA-256 Sigma Word
vaddudm	        Vector Add Unsigned Doubleword Modulo
vmrgew 		Vector Merge Even Word
lxvd2x		Load VSX Vector Doubleword*2 Indexed
stxvd2x 	Store VSX Vector Doubleword*2 Indexed
```

### DIFF
--- a/compiler/p/runtime/ppcasmdefines.inc
+++ b/compiler/p/runtime/ppcasmdefines.inc
@@ -416,6 +416,37 @@
         .int 0x100005c8 | \vrt << 21 | \vra << 16
         .endm
 
+        .macro lxvd2x   xt, ra, rb
+        .int 0x7C000698 | (\xt & 31) << 21 | \ra << 16 | \rb << 11 | (\xt & 32) >> 5
+        .endm
+        .macro stxvd2x   xt, ra, rb
+        .int 0x7C000798 | (\xt & 31) << 21 | \ra << 16 | \rb << 11 | (\xt & 32) >> 5
+        .endm
+        .macro vspltw   vrt, vra, uim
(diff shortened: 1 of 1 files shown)


fjeremic:
Please amend the commit message so as to conform to the commit guidelines:

> The first line describes the change made. It is written in the imperative mood, say what happens when the patch is applied. Keep it short and simple. **The first line should be less than 50 characters**, sentence case, and does not end in a period. **Leave a blank line between the first line and the message body.**

May I also suggest a more descriptive header message, perhaps:

_Add vector related macro instructions for Power_

vijaysun-omr:
@zl-wang and @ymanton Please review this.

0xdaryl:
The changes themselves look fine to me, but I'll defer to @zl-wang or @ymanton for the final say.

Could you clean up the commit message text please?  It isn't formatted very cleanly in terms of spaces, and the case of the instruction mnemonics is wildly inconsistent (choose either upper or lower case).

hjooybar:
@zl-wang 
There is no missing define. Assembler on AIX has support for more instruction comparing the Linux assembler. There is no need to define some of these instructions on AIX because the assembler already know them.
I applied all other review comments and updated the merge request.

vijaysun-omr:
Checks have passed and all the review comments addressed. Merging.

PR #716, opened 2017-01-30
Consolidate logics related to opcode generation

0dvictor:
De-duplicate logics related to opcode length calculation and binary generation.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -2985,12 +2985,10 @@ void OMR::X86::CodeGenerator::apply32BitLoadLabelRelativeRelocation(TR::Instruct
    {
    TR::Instruction *movInstructionX86   = movInstruction;
 
-
    TR_ASSERT(movInstructionX86->getOpCodeValue() == MOV4RegImm4,"wrong load instruction used for apply32BitLoadLabelRelativeRelocation\n");
 
    uint8_t *cursor = movInstructionX86->getBinaryEncoding();
-   cursor += movInstructionX86->rexPrefixLength();
-   cursor += movInstructionX86->getOpCode().getOpCodeLength();
+   cursor += movInstructionX86->getOpCode().length(movInstructionX86->rexBits()) + movInstructionX86->rexRepeatCount();
 
    int32_t distance = int32_t(endLabel->getCodeLocation() - startLabel->getCodeLocation());
    TR_ASSERT(IS_32BIT_SIGNED(distance),
@@ -3429,14 +3427,14 @@ uint8_t *OMR::X86::CodeGenerator::generatePadding(uint8_t              *cursor,
          if (length >= 5)
             {
             length -= 5;
-            cursor = TR_X86OpCode::copyBinaryToBuffer(JMP4, cursor);
+            cursor = TR_X86OpCode(JMP4).binary(cursor);
             *(int32_t*)cursor = length;
             cursor += 4;
             }
          else
             {
             length -= 2;
-            cursor = TR_X86OpCode::copyBinaryToBuffer(JMP1, cursor);
+            cursor = TR_X86OpCode(JMP1).binary(cursor);
             *(int8_t*)cursor = length;
             cursor += 1;
             }
--- a/compiler/x/codegen/X86Ops.hpp
+++ b/compiler/x/codegen/X86Ops.hpp
@@ -1388,6 +1388,67 @@ class TR_X86OpCode
 
    uint8_t getOpCodeLength() {return _binaryEncodings[_opCode]._length;}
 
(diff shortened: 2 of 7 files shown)


0xdaryl:
This isn't an official review, but just a heads up while this is a WIP that you'll need to shorten the commit title.

0dvictor:
Thank you for pointing it out. The final commit title is shortened.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -2985,12 +2985,10 @@ void OMR::X86::CodeGenerator::apply32BitLoadLabelRelativeRelocation(TR::Instruct
    {
    TR::Instruction *movInstructionX86   = movInstruction;
 
-
    TR_ASSERT(movInstructionX86->getOpCodeValue() == MOV4RegImm4,"wrong load instruction used for apply32BitLoadLabelRelativeRelocation\n");
 
    uint8_t *cursor = movInstructionX86->getBinaryEncoding();
-   cursor += movInstructionX86->rexPrefixLength();
-   cursor += movInstructionX86->getOpCode().getOpCodeLength();
+   cursor += movInstructionX86->getOpCode().length(movInstructionX86->rexBits()) + movInstructionX86->rexRepeatCount();
```
andrewcraik: I'm not quite sure how this is right - if the meaning of length(movInstructionX86->rexBits()) is the length of one rex prefix then shouldn't we be multiplying the count?

andrewcraik: or are we relying on this being a single byte to make this math right? If so how does the addition of a unnecessary 0x40 work in this scheme with the changes?

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1388,6 +1388,64 @@ class TR_X86OpCode
...
+         }
+      if (rex)
+         {
+         *cur++ = rex;
+         }
+      if (needsSSE42OpcodePrefix())
+         {
+         *cur++ = 0x0f;
+         }
+      return copyBinaryToBuffer(cur);
```
andrewcraik: prior to this it would make sense that, when asserts are enabled, we verify the number of bytes written (ie number of cur bumps) matches the length returned from the length function to help prevent accidental partial updates.

compiler/x/codegen/OMRCodeGenerator.cpp
0dvictor: rexBits() is always a single byte which contains only a necessary REX.
The addition of unnecessary 0x40 is counted as part of rexRepeatCount(), which returns the number of unnecessary REX bytes.

compiler/x/codegen/X86Ops.hpp
0dvictor: Good suggestion, it has been updated.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1388,6 +1388,67 @@ class TR_X86OpCode
...
+      if (rex)
+         {
+         *cur++ = rex;
+         }
+      if (needsSSE42OpcodePrefix())
+         {
+         *cur++ = 0x0f;
+         }
+      cur = copyBinaryToBuffer(cur);
+      TR_ASSERT(cur-start == length(rex), "Actual binary length should be the same as calculated one");
```
andrewcraik: hmm - do we have to worry about getOpCodeLength in the length function?

compiler/x/codegen/OMRCodeGenerator.cpp
andrewcraik: Ok can you please comment rexRepeatCount to describe when it is non-zero and how the number returned from there relates to required and non-required REX prefixes because it doesn't seem to be completely obvious to me so could confuse others.

0dvictor: Good suggestion, updated.

compiler/x/codegen/X86Ops.hpp
0dvictor: It makes more sense to include getOpCodeLength in the length function; length() returns the length of the whole opcode, which includes both prefixes and the actual opcode. In addition, since current binary encoding scheme arbitrarily divide the instruction into prefixes (in op properties) and opcode, only the total length has a consistent and valid meaning.

andrewcraik: Ok - as lng as copyBinaryToBuffer copies all of the opcode and we are comparing total lengths then this should be fine.

andrewcraik:
I think this is good to merge since my comments have been addressed. Would appreciate a second pair of eyes over the code to help make sure I haven't missed anything.

cathyzhyi:
LGTM

andrewcraik:
@0xdaryl any additional concerns?

Issue #717, opened 2017-01-30
LoopVersioner creates certain guard tests unsafely


ymanton:
When creating `asynccheck` guard tests for loops `LoopVersioner` duplicates trees from inside the loop to create these tests without doing anything to clear any flags that may be set in the duplicated nodes. These flags need to be cleared because they may no longer be valid at the location where the duplicated trees are inserted.

mstoodle:
PR looks merged so I'll aggressively close this issue. If there's more to do here @ymanton please say so and I'll reopen it.

PR #718, opened 2017-01-30
Make sure LoopVersioner creates async check guard tests safely

ymanton:
This patch set fixes LoopVersioner so that it will create guard tests for `asynccheck` versioning safely by making sure that any duplicated trees have their flags reset. The first patch introduces a new `OMR::Node` method that duplicates trees for the purposes of inserting them arbitrarily; the second patch uses it in `LoopVersioner`; the third is cosmetic and cleans up the surrounding code in `LoopVersioner` to improve readability.

Issue: #717

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -1733,6 +1733,14 @@ OMR::Node::duplicateTree(bool duplicateChildren)
    return unsafeNode;
    }
 
+TR::Node *
+OMR::Node::duplicateTreeForCodeMotion()
+   {
+   TR::Node *result = self()->duplicateTree(true);
+   result->resetFlagsForCodeMotion();
+   return result;
+   }
+
 /**
  * Method to duplicate an entire subtree. Tries to do so 'safely' by maintaining a mapping of visited (and duplicated) nodes
  */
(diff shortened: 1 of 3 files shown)


ymanton:
> There seem to be several calls to duplicateTree from other parts of LoopVersioner.cpp unrelated to async check versioning. In many of those cases, the same sort of code motion is being done after duplicating a node.
Is there a reason for limiting your change to the async check versioning case ?

Considered that, but I wasn't sure that it would be justified. The case addressed by this PR is a verified problem that fixes a failing downstream Java test. There are many uses of `duplicateTree()` in this part of the code, changing so many of them could expose other functional and performance bugs for downstream.

Never the less, I can make the change if we think it's worthwhile.

vijaysun-omr:
Since the implications relate to functional correctness I would like to see us try and fix those other places as well. However, I do not want to hold up this PR since what you have here is obviously better than what we had before and will fix the test case failure you mentioned. 

My suggestion would be to open a new PR and organize the changes logically (i.e. different kinds of versioning you are changing the code for) in terms of different commits in that PR so that it is easier for someone doing a binary search in case there is a problem. Please create that new PR and paste a link in here (even if it's a WIP) and I will merge this in.

ymanton:
I've opened #723 to fix the other cases.

vijaysun-omr:
Thanks I have merged this PR now.

PR #719, opened 2017-01-30
Re-architect TR_AliasSetInterface to use the CRTP

### LINKED ISSUE #715: Convert AliastSetInterface to use the CRTP
The current architecture of AliasSetInterface is rather confusing and difficult to reason about. Re-architecting to something that is simpler and easier to understand will greatly simplify later work. The CRTP seems like a good candidate as it should preserve the statically polymorphic behavior of the current architecture. Being a relatively small architectural change, it should not change anything functionally. However, it will require an inversion of class hierarchies, so some care must be taken when making the change.

Leonardo2718:
The previous implementation of `TR_AliasSetInterface` was rather
confusing due to the nature of its class hierarchy. This commit changes
the code to use the CRTP, making the class hierarchy easier to read and
understand, since the CRTP is a well known C++ pattern.

Fixes #715 

### DIFF
--- a/compiler/il/AliasSetInterface.hpp
+++ b/compiler/il/AliasSetInterface.hpp
@@ -36,24 +36,16 @@
 
 
 template <class AliasSetInterface>
-class TR_AliasSetInterface : public AliasSetInterface {
+class TR_AliasSetInterface {
 public:
 
-  TR_AliasSetInterface(TR::SymbolReference *symRef, bool isDirectCall = false,
-                       bool includeGCSafePoint = false)
-    : AliasSetInterface(symRef),
-      _isDirectCall(isDirectCall),
-      _includeGCSafePoint(includeGCSafePoint) {
-  }
-
-  TR_AliasSetInterface(TR::Node *node, bool isDirectCall = false, bool includeGCSafePoint = false) :
-    AliasSetInterface(node),
+  TR_AliasSetInterface(bool isDirectCall = false, bool includeGCSafePoint = false) :
     _isDirectCall(isDirectCall),
     _includeGCSafePoint(includeGCSafePoint) {
   }
 
   TR_BitContainer getTRAliases() {
-    return AliasSetInterface::getTRAliases(_isDirectCall, _includeGCSafePoint);
+    return static_cast<AliasSetInterface*>(this)->getTRAliases_impl(_isDirectCall, _includeGCSafePoint);
   }
 
    template <class BitVector>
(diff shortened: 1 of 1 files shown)


xliang6:
LGTM.

0xdaryl:
Looks fine.  Please document here the fact that you ran a higher-than-usual amount of testing on this change on J9 Java builds and it passed cleanly.

@vijaysun-omr , any concerns with this one?

Leonardo2718:
I confirm that I have tested this change by running extensive J9 Java tests and that they passed cleanly.

andrewcraik:
Why do we have a function naming convention introduced in these files that contains an _? That seems unlike the rest of the code - if there is a good reason then I'm willing to go along, but it just seems out-of-place.

compiler/il/AliasSetInterface.hpp
```diff
@@ -36,24 +36,16 @@
...
-  TR_AliasSetInterface(TR::SymbolReference *symRef, bool isDirectCall = false,
-                       bool includeGCSafePoint = false)
-    : AliasSetInterface(symRef),
-      _isDirectCall(isDirectCall),
-      _includeGCSafePoint(includeGCSafePoint) {
-  }
-
-  TR_AliasSetInterface(TR::Node *node, bool isDirectCall = false, bool includeGCSafePoint = false) :
-    AliasSetInterface(node),
+  TR_AliasSetInterface(bool isDirectCall = false, bool includeGCSafePoint = false) :
```
andrewcraik: Is the dropping of the node here due to the removal of node aliasing?

Leonardo2718: Not quite. It's because of the inversion of the class hierarchy. 

`TR_AliasSetInterface` used to be in the middle of the hierarchy and needed to pass the node to it's parent. Now that it's at the top, it doesn't need to know about the node (or symref for that matter). It is the derived class `TR_NodeAliasSetInterface` that now keeps track of the node.

Leonardo2718:
I used `_` because that seems to be the convention when using the CRTP (at least from what I've seen). It supposed to give a visual queue that a function is used as a delegate and shouldn't be called from outside of the class hierarchy.

vijaysun-omr:
LGTM

0xdaryl:
@andrewcraik , have your questions been addressed?

andrewcraik:
LGTM

Leonardo2718:
Thanks @andrewcraik, @xliang6, @0xdaryl, and @vijaysun-omr for reviewing.

PR #721, opened 2017-01-30
DDR Dwarf Parser for AIX - Add parsing for dwarf dies

twlogger:
This should be merged in after https://github.com/eclipse/omr/pull/674


tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -441,6 +720,618 @@ parseDwarfInfo(const char *line)
...
+					_builtInTypeDie = newDie;
+				}
+				if (NULL !=  previousDie) {
+					previousDie->_sibling = newDie;
+					newDie->_previous = previousDie;
+				}
+				previousDie = newDie;
+
+				/* Insert the newly created DIE into the unordered map for use later on */
+				pair<double, Dwarf_Die> result;
```
jason-hall: This unitialized variable can be directly set to the make_pair on the next line

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -114,7 +146,7 @@ extractFileID(const string data)
 {
 	stringstream ss;
 	double ret = 0;
-	ss << data.substr(1, data.length() - 1);
+	ss << data.substr(1, data.find_first_of(']')-1);
```
jason-hall: ) - 1);

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -151,31 +183,200 @@ isCompilerGenerated(const string data)
...
- * @param[in]  data: the string to be parsed. data should be the portion of the declaration line after "debug     0    decl" and whitespace. 
+ * Parses the data passed in and populates a DIE as a base type
+ * param[in] data: the data to be parsed, it should consist of the following characters only: "1234567890-"
+ * param[out] currentDie: the DIE to populate with the parsed data.
+ */
+static int
+parseBaseType(const string data, Dwarf_Die currentDie, Dwarf_Error *error)
+{
+	int ret = DW_DLV_OK;
+	int index = toInt(data)*-1;
```
jason-hall: ) * -1;

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -441,6 +720,618 @@ parseDwarfInfo(const char *line)
...
+				}
+				if (NULL !=  previousDie) {
+					previousDie->_sibling = newDie;
+					newDie->_previous = previousDie;
+				}
+				previousDie = newDie;
+
+				/* Insert the newly created DIE into the unordered map for use later on */
+				pair<double, Dwarf_Die> result;
+				result = make_pair<double, Dwarf_Die>((double)(i*(double)-1), (Dwarf_Die)newDie);
```
jason-hall: (double)(i * -1)

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -151,31 +183,200 @@ isCompilerGenerated(const string data)
...
+			Dwarf_Attribute subrangeType = new Dwarf_Attribute_s();
+			Dwarf_Attribute subrangeUpperBound = new Dwarf_Attribute_s();
+			if ((NULL == type) || (NULL == newDie) || (NULL == subrangeType) || (NULL == subrangeUpperBound)) {
+				ret = DW_DLV_ERROR;
+				setError(error, DW_DLE_MAF);
+			} else {
+				type->_type = DW_AT_type;
+				type->_form = DW_FORM_ref1;
+				type->_nextAttr = NULL;
+				/* If the type's typeID is built-in, leave it as is, otherwise, add the current file's ID to make it unique */
```
jason-hall: This is an incorrect use of an attribute. When _form is DW_FORM_ref, only _ref can be set. _sdata should not be used unless the _form is DW_FORM_SDATA.

An attribute represents only one piece of information, such as a reference or file id, but never both.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -441,6 +720,618 @@ parseDwarfInfo(const char *line)
...
+}
+
+/**
+ * Looks up and populates the reference attributes which were added while populating Dwarf DIEs
+ */
+static void
+populateAttributeReferences()
+{
+	/* Iterate through the vector containing all attributes to be populated*/
+	for (unsigned int i = 0; i < refsToPopulate.size(); i++) {
```
jason-hall: refsToPopulate would make more sense as vector of pairs or some such collection. That would allow you to store a pair of values: an attribute which needs to be set, and an ID for what it needs to be set to. The _sdata of the attribute should not be used for this purpose. From the perspective of the DwarfScanner or someone consuming the Dwarf API, its not obvious why an attribute has 2 values or what the garbage value could mean.

As is, the code assumes that the attribute structure is defined such that both the _sdata and _ref are available, separate locations in memory which can be used simultaneously. While this is true for the current implementation, th Dwarf spec for attributes is such that only one should ever be set.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -441,6 +720,618 @@ parseDwarfInfo(const char *line)
...
+			size->_nextAttr = declFile;
+			size->_form = DW_FORM_udata;
+			size->_sdata = 0;
+			size->_udata = dieSize;
+			size->_refdata = 0;
+			size->_ref = NULL;
+
+			declFile->_type = DW_AT_decl_file;
+			declFile->_nextAttr = NULL;
+			declFile->_form = DW_FORM_string;
```
jason-hall: DW_AT_decl_file is actually to be of form DW_FORM_sdata, not string. The number is to be an index into the list of file names returned by dwarf_srcfiles() rather than the file name itself.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.hpp
```diff
@@ -19,6 +19,19 @@
...
 #include "DwarfFunctions.hpp"
 
+/* PTR_SIZE varies for 32bit, 64bit */
+const int PTR_SIZE = 8;
+
+const string BUILT_IN_TYPES [36] = {"", "int", "char", "short", "long", "unsigned char",
+"signed char", "unsigned short", "unsigned int", "unsigned", "unsigned long", "void", 
+"float", "double", "long double", "integer", "boolean", "short real", "real", "stringptr",
+"character", "logical*1", "logical*2", "logical*4", "logical", "complex", "complex", "integer*1",
+"integer*2", "integer*4", "wchar", "long long", "unsigned long long", "logical*8", "integer*8",
```
jason-hall: What is an integer*1, integer*2, logical*1, complex, etc? Are the previous base types insufficient to cover all cases? ie char for size 1, short for size 2, int for size 4, etc.

If the DwarfScanner and superset generator produce base types of "logical" and "integer", how would one translate these when generating the java code from the superset? The set of lowest base types should probably only have signed/unsigned versions for each size. For other types, I'd suggest creating a Die for a typedef with the name and setting its type to the corresponding base type of the appropriate size.

charliegracie:
I have not opened up the Eclipse CQ yet.  If this is just an addition/fix to #674 can you just add it to that so that I do not need another CQ because this PR is over 1000 lines. Hopefully we hear from @VermaSh in #608 soon

jason-hall:
@charliegracie Yes, this is a continuation of #674.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.hpp
twlogger: These are just the types I found [here](https://sourceware.org/gdb/current/onlinedocs/stabs/Negative-Type-Numbers.html#Negative-Type-Numbers), I'm not sure which we need and which we do not. I will look into it.

jason-hall: Well, I wouldn't deem some as not needed and delete them. I suggest creating one signed/unsigned for each possible size, then creating typedefs that reference them for the types of redundant sizes. That link suggests some are Fortran-specific.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -218,14 +217,16 @@ parseArray(const string data, Dwarf_Die currentDie, Dwarf_Error *error)
...
 				int ID = toInt(arrayInfo[3]);
 				if (ID > 0) {
-					ID = ID + _fileID;
+					ID = ID + _fileID + NUM_BUILT_IN_TYPES;
+				} else {
+					ID = ID * -1;
 				}
 				type->_udata = 0;
-				type->_refdata = 0;
+				type->_refdata = ID;
```
jason-hall: As we discussed in person: I strongly do not recommend making the refdata numbers dependent on the IDs extracted from the AIX stabstrings. You should define your own numbering which is guaranteed to be unique. We have no guarantees for the ranges of the numbers extracted from stabstrings.

My suggestion was to create a static number. Every time you create a new Die, associate that Die with the static number and increment it. As such, the first Die would be associated with X, the next with X + 1, etc. Please refer to the OSX version of this code.

As you said to me already, these IDs based on the AIX stabstring info are not necessarily guaranteed to be unique. For example, ID 101 in file 1000 would be the same as ID 100 in file 1001. This is a poor design unless you can guarantee that (1) the ID numbers are documented to always be in a specific range for projects of all sizes, (2) the file ids are documented to always be in a specific range for projects of all sizes, (3) the file and type ids are always unique when added together for all projects, and (4) the ranges of these values will never change in future versions of AIX/stabstring info.

The pair of file id and type id (not added together) is the best we've got for determining the type of a variable in the stabstring info. However, since we cannot guarantee the range of these values, adding them together is not a good idea. The best way to futureproof it and guarantee unique numbering for refdata is to come up with **your own** numbering and **not use refdata numbering that relies on IDs from AIX**.

charliegracie:
@twlogger As per the Contributing.md the first line of each commit message has to be 50 characters or less.  Can you update your commit messages that do not conform to this.  Thanks

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -0,0 +1,2012 @@
...
+static int parseNewFormat(const string data, const string dieName, Dwarf_Die currentDie, Dwarf_Error *error);
+static int parseOldFormat(const string data, const string dieName, Dwarf_Die currentDie, Dwarf_Error *error);
+static int parseTypeDef(const string data, const string dieName, Dwarf_Die currentDie, Dwarf_Error *error);
+static void populateAttributeReferences();
+static int populateBuiltInTypeDies(Dwarf_Error *error);
+static void populateNestedClasses();
+static int setDieAttributes(const string dieName, const unsigned int dieSize, Dwarf_Die currentDie, Dwarf_Error *error);
+/* temporary debugging methods */
+using namespace std;
+void printType (Dwarf_Die * t) {
```
jason-hall: You seem to have re-added these debug functions, possibly unintentionally? They use "cout" and do not follow the coding standards...

I suggest adding more DEBUGPRINTF's instead and print out the Die's as they are created. These prints can be turned on/off easily and as such can be left in instead of added and removed as you debug your code.

jason-hall:
A more robust solution would probably be to move the PdbScanner implementation of "blacklisting" to superset/blob generation and modify the tool to accept a "blacklist" file as a command line option. That should maximize transparency of the blacklist to the user.

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -105,10 +105,16 @@ dwarf_init(const char *filepath,
 	/* Populate the built-in types */
 	ret = populateBuiltInTypeDies(error);
 
+	/* Get the path of the file descriptor */
+	buffer = strdup("/proc/");
```
jason-hall: This code actually can segfault as you have written in.

1. strdup("/proc/") will malloc a char array that can hold precisely 6 characters.
2. Then toString(getpid()) creates a string object, allocating more memory.
3. Now, you have a char [7] == {"/proc/"} and you are concatenating the string object converted to a char array to the end of it, overrunning the end of the buffer.

Instead of malloc'ing a tiny array on the heap with strdup() and creating 2 std::string objects then freeing it, why not just create a local array of an appropriate size and put it all together with one sprintf()? General coding practice is to not allocate heap memory for something (buffer) you are only using temporarily/locally.

This will use a local variable instead of the heap, not overrun your buffer, and not wastefully create and convert std::strings:
char buffer[100] = {'/0'};
sprintf(buffer, "/proc/%d/fd/%d", getpid(), fd);

tools/ddrgen/src/scanners/dwarf/AixDwarfParser.cpp
```diff
@@ -73,16 +73,25 @@ dwarf_finish(Dwarf_Debug dbg, Dwarf_Error *error)
...
 
 	/* Delete the undefined type reference DIE */
 	if (NULL != _undefinedRef) {
 		deleteDie(_undefinedRef);
+		_undefinedRef = NULL;
 	}
 
+	_lastDie = NULL;
 	Dwarf_CU_Context::_fileList.clear();
+	createdDies.clear();
```
jason-hall: All of these collections should not be needed after dwarf_init() completes. At the end of dwarf_init, the main context's pointers should lead to all dwarf_die's. If the stabstring parsing is complete in dwarf_init, there is no reason to keep any of these collections until dwarf_finish().

charliegracie:
The Eclipse CQ can be found here:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13118

charliegracie:
The CQ was approved for contributing.

PR #722, opened 2017-01-31
Fix potential bugs and refactor spinlock code

babsingh:
Potential bugs:

1) omrthread_spinlock_acquire gets called from two locations:
monitor_enter_three_tier and omrthread_monitor_try_enter_using_threadId.
In monitor_enter_three_tier, there is an attempt to acquire the monitor
after a call to omrthread_spinlock_acquire. In
omrthread_monitor_try_enter_using_threadId, there is no attempt to
acquire the monitor after a call to omrthread_spinlock_acquire. Thus,
omrthread_spinlock_acquire should attempt to acquire the lock atleast
once when spinning is disabled. In eclipse#457, changes were pushed that
violated this requirement. In eclipse#457 changes,
omrthread_spinlock_acquire would make no attempts to acquire the monitor
if it was decided not to spin. In the new changes, this bug has been
fixed.

2) Adaptive spin feature can disable spinning when other threads are
already in the process of spinning. Threads that are spinning do not use
any synchronization primitive to read spin counts from
omrthread_monitor_t. In omrthread_spinlock_acquire, we currently read
spin counts multiple times from omrthread_monitor_t in the nested for
loops. If the adaptive spin feature suddenly disables spinning while a
thread is already in the process of spinning, then the spin counts will
get corrupted, and incorrect JLM information will be recorded. Also, we
currently do not stop spinning if the adaptive spin feature disables
spinning while a thread is already in the process of spinning. The new
changes fix these bugs. Spin counts are read once before spinning
instead of reading spin counts multiple times while spinning. In the
nested for loops, a check has been added to recognize if spinning has
been disabled by the adaptive spin feature.

Refactor code: 

1) J9_ARE_ALL_BITS_SET is more reliable to make sure if a flag is set.
In the new changes, J9_ARE_ALL_BITS_SET is used for flag checks.

2) Updating JLM counts at two places makes it difficult to read code.
In the new changes, JLM counts are updated in a more efficient manner at
one location. 

3) A new build flag is added: OMR_THR_SPIN_CODE_REFACTOR. The new
changes are wrapped around this build flag. This build flag helps
preserve the old implementation. This build flag is currently disabled
so the old implementation is active.

4) Enabling build flags: OMR_THR_SPIN_CODE_REFACTOR and
OMR_THR_SPIN_WAKE_CONTROL.

5) We have done thorough testing on x86. Build flag 
OMR_THR_SPIN_CODE_REFACTOR is added to allow thorough testing on
other platforms. This build flag and old changes will be removed once 
thorough testing is complete on other platforms. Issue to remove build
flag and old changes - https://github.com/eclipse/omr/issues/773.


thread/common/threadhelpers.cpp
```diff
@@ -50,52 +50,57 @@ intptr_t
...
+	BOOLEAN spinning = TRUE;
+ 	if ((monitor->spinThreads < lib->maxSpinThreads) && J9_ARE_NO_BITS_SET(monitor->flags, J9THREAD_MONITOR_DISABLE_SPINNING)) {
+ 		VM_AtomicSupport::add(&monitor->spinThreads, 1);
+ 	} else {
+ 		spinning = FALSE;
+ 		spinCount3Init = 1;
+ 		spinCount2Init = 1;
+ 		spinCount1Init = 1;
+ 		spinCount3 = spinCount3Init;
+ 		spinCount2 = spinCount2Init;
```
charliegracie: Are these 2 lines correct?  The are going to set spinCount3 and spinCount2 to 1 because you just modified the values of the Init variables.  If they are correct it may be more clear to just set them to 1 explicitly.

babsingh: Yes. spinCount* are used by the nested for loops, and these variables will change in the for loops. spinCount*Init stores the initial snap shot of spin counts. We want to initialize spinCount* using spinCount*Init values. Code is written in this manner in order to reflect dependence between spinCount* and spinCount*Init.

babsingh:
We have done a lot of testing before where we stop spinning when the number of spin threads exceeds the maximum number of allowed spin threads. The new changes follow the same concept; we do not spin but we execute compare and swap once on the lock. The variable initialization code at the beginning should have no impact on performance since the most critical portion is the nested for loops. The check added in the nested for loops to identify if spinning is disabled is necessary. Currently, we will keep spinning if spinning is disabled since the changes to monitor->spinCount* will not reflect immediately. This will cause a regression. I can benchmark to verify the impact of changes. Please agree upon the final changes.

charliegracie:
Do you have the details on the performance?

babsingh:
Benchmark results [local environment]:
i) High contention micro-benchmark: 2 core machine (4% regression), 4 core machine (1.5% improvement), 8 core machine (10% improvement) and 16 core machine (13.5% improvement). 
ii) VolanoMark: 1%-4% improvement and no regression

charliegracie:
@babsingh as per the Contributing.md can you limit the first line of the commit message to 50 characters.  Thanks

babsingh:
Can you limit the first line of the commit message to 50 characters?
done.

charliegracie:
Why are we adding this under a build flag that is not enabled? Is this a work in progress change and you are expecting follow on changes?

babsingh:
Build flag has been enabled.

charliegracie:
@babsingh if it is enabled then why are we keeping the old support? If we believe this is a valid change then I would prefer to only have one copy of the code.  One copy of the code is easier to keep track of and work with.

babsingh:
Update -- Enabled two flags: OMR_THR_SPIN_CODE_REFACTOR and OMR_THR_SPIN_WAKE_CONTROL.

It is easier to go to the old version in case of emergency. We would like to monitor the new changes for a month or two. I have created a task to clean up the old code.

PR #723, opened 2017-01-31
WIP: Make sure LoopVersioner creates various guard tests safely

ymanton:
`LoopVersioner` creates various guards unsafely. See #718 for one example. This PR will be used to fix other cases that appear problematic but that haven't caused any known issues yet.

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -1733,6 +1733,14 @@ OMR::Node::duplicateTree(bool duplicateChildren)
    return unsafeNode;
    }
 
+TR::Node *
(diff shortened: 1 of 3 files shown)


mstoodle:
What's the current status here, please?

ymanton:
Currently on the back burner for a while. This is simply a clean up task to rewrite existing code that is not known to have any issue to use a new API. The one case that did have a bug in it has already been fixed via https://github.com/eclipse/omr/issues/718.

mstoodle:
@ymanton so maybe we can close this PR until such time as it comes off the backburner?

ymanton:
Sure, I'll keep track of this task elsewhere.

PR #724, opened 2017-01-31
Add exception asynccheck insertion opt

ncough:
Adds the exception handler asynccheck insertion optimization to the
ilgen opt strategy and a tracing option.

Corrects the case where an asynccheck has been inserted into an
exception handler, which is being split to use a goto. In this
situation, the original asynccheck is retained in the exception
handler and a duplicate is added to the new block. This prevents
the goto edge from creating a loop without an asynccheck and
ensures the exception handler still has its asynccheck.

### DIFF
--- a/compiler/optimizer/ValuePropagation.cpp
+++ b/compiler/optimizer/ValuePropagation.cpp
@@ -7739,11 +7739,21 @@ void TR::ValuePropagation::doDelayedTransformations()
       TR::TreeTop * firstTT = predictedCatchBlock->getFirstRealTreeTop();
 
       // Find the first real tree in catch block; ignoring the
-      // profiling trees possibly inserted by catch block profiler
-      //
-      while (firstTT &&
-             firstTT->getNode()->isProfilingCode())
-        firstTT = firstTT->getNextRealTreeTop();
+      // profiling trees possibly inserted by catch block profiler or
+      // asyncchecks for OSR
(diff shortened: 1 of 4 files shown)


0xdaryl:
Please trim the length of your commit title <50 chars per the CG.

compiler/optimizer/ValuePropagation.cpp
```diff
@@ -7739,11 +7739,21 @@ void TR::ValuePropagation::doDelayedTransformations()
...
+      // asyncchecks for OSR
+      TR::TreeTop * asyncCheck = NULL;
+      while (firstTT)
+         {
+         if (firstTT->getNode()->isProfilingCode())
+            firstTT = firstTT->getNextRealTreeTop();
+         else if (firstTT->getNode()->getOpCodeValue() == TR::asynccheck)
+            {
+            asyncCheck = firstTT;
+            firstTT = firstTT->getNextRealTreeTop();
```
vijaysun-omr: This tracks an async check tree for duplication later on, but it also skips ahead to the next tree top for an async check tree whereas this was not being done in the original loop. Is this skipping for an async check tree intentional ?

ncough: Yes, it is necessary to leave an async check tree in the original exception handler, so it is skipped as well.

vijaysun-omr:
Looks pretty safe to me. Merging.

Issue #706, opened 2017-01-25
Log File tests aren't cleaning up after themselves


mgaudet:
Running make test leaves behind 

- [ ] `traceCG.log`
- [ ] `traceFull.log`

mgaudet:
I think this was fixed with #728

PR #728, opened 2017-02-02
Fix Log File tests to clean up its files

ldferrar:
Fixed lifetime of string constants holding file name by changing const char * to std::string.

Resolves issue #706. 
Signed-off-by: Lauren Ferraro <lferraro@ca.ibm.com>


mgaudet:
Looks good to me!

0xdaryl:
The changes themselves look reasonable to me.

Perhaps I'm being obtuse, but I'm missing the connection as to why this is needed to cause the test artifacts to be removed.   Can you elaborate why this wasn't happening when they were simple const chars ?

ldferrar:
In the previous implementation, a shallow copy was made of a stack allocated c-style string. This caused the stored file paths to point to invalid memory. Changing the c-style strings to std::string ensures a deep-copy is done.

PR #732, opened 2017-02-03
Add opts for arrayset on short const length array

jingge-ibm:
Currently in arrayset evaluator on X86, we do not take advantage of
the fact that we can know the length of some arrays at compile time.
Some optimizaitons can be enabled if we have the size of array at
compile time. In particular, two optimizations are implemented.
First, for short constant length array, a better sequence of
instructions will be generated. Second, in additional to the short
constant length array, if the value is also zero, we are able to
generate a even better sequence.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -3403,6 +3403,339 @@ TR::Register *OMR::X86::TreeEvaluator::encodeUTF16Evaluator(TR::Node *node, TR::
    }
 #endif
 
+static void packUsingShift(TR::Node* node, TR::Register* tempReg, TR::Register* sourceReg, int32_t size, TR::CodeGenerator* cg)
+   {
+   generateRegRegInstruction(MOV8RegReg, node, tempReg, sourceReg, cg);
+   generateRegImmInstruction(SHL8RegImm1, node, sourceReg, size*8, cg);
+   generateRegRegInstruction(OR8RegReg, node, sourceReg, tempReg, cg);
+   }
+
+static void generateMovToMemInstructionsForArrayset(TR::Node* node, TR::Register* valueReg, uint8_t size, TR::Register* addressReg, int32_t index, TR::CodeGenerator* cg)
+   {
+   TR_X86OpCodes opcode;
+   switch(size)
+      {
+      case 1:
+         opcode = S1MemReg;
+         break;
+      case 2:
+         opcode = S2MemReg;
+         break;
+      case 4:
+         opcode = S4MemReg;
+         break;
+      case 8:
+         opcode = S8MemReg;
+         break;
+      case 16:
+         opcode = MOVDQUMemReg;
+         break;
+      }
+   generateMemRegInstruction(opcode, node, generateX86MemoryReference(addressReg, index, cg), valueReg, cg);
+   }
+
+static void packXMMWithMultipleValues(TR::Node* node, TR::Register* XMMReg, TR::Register* sourceReg, int8_t size, TR::CodeGenerator* cg)
+   {
+   switch(size)
+      {
+      case 1:
+      case 2:
+         {
+         static uint8_t MASKOFSIZEONE[] =
+            {
+            0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00,
+            };
+
+         static uint8_t MASKOFSIZETWO[] =
+            {
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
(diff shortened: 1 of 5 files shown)


jingge-ibm:
Personal build: http://zaxxon.torolab.ibm.com/build_info.php?build_id=131926

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3403,146 +3403,441 @@ TR::Register *OMR::X86::TreeEvaluator::encodeUTF16Evaluator(TR::Node *node, TR::
...
+         else if (reminder != 0)
+            {
+            GenerateMoveInstruction(node, XMM, 16, addressReg, totalSize-16, cg);
+            }
+         cg->stopUsingRegister(XMM);
+         }
+      }
+   }
+
+inline void ArraySet64BitPrimitiveOn32BitISA(TR::Node* node, TR::Register* addressReg, TR::Register* valueReg, TR::Register* sizeReg, TR::CodeGenerator* cg)
```
0xdaryl: * Please elaborate on why this function is marked `inline` rather than `static`?

* I'd drop the `ISA` on the function name.  If you really want to say `32BitISA` then say `IA32` instead for consistency with the rest of the codegen.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3403,146 +3403,441 @@ TR::Register *OMR::X86::TreeEvaluator::encodeUTF16Evaluator(TR::Node *node, TR::
...
          break;
       case 4:
          generateRegImmInstruction(SHRRegImm1(cg), node, sizeReg, 2, cg);
+         generateInstruction(REPSTOSD, node, stosDependencies, cg);
          break;
       case 8:
          generateRegImmInstruction(SHRRegImm1(cg), node, sizeReg, 3, cg);
+         generateInstruction(REPSTOSQ, node, stosDependencies, cg);
          break;
       default:
```
0xdaryl: Three of the case statements are identical except for the opcode value and the shift amount.  Consider refactoring to save on code bloat.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3403,146 +3403,441 @@ TR::Register *OMR::X86::TreeEvaluator::encodeUTF16Evaluator(TR::Node *node, TR::
...
+   generateLabelInstruction(JRCXZ1, node, endLabel, cg);
+   generateLabelInstruction(LABEL, node, loopLabel, cg);
+   generateMemRegInstruction(MOVQMemReg, node, generateX86MemoryReference(addressReg, sizeReg, 3, -8, cg), XMM, cg);
+   generateLabelInstruction(LOOP1, node, loopLabel, cg);
+
+   generateLabelInstruction(LABEL, node, endLabel, deps, cg);
+   cg->stopUsingRegister(XMM);
+   }
+
+inline void ArraySetDefault(TR::Node* node, uint8_t elementSize, TR::Register* addressReg, TR::Register* valueReg, TR::Register* sizeReg, TR::CodeGenerator* cg)
```
0xdaryl: Please elaborate on why this function is `inline` rather than `static`.

jingge-ibm:
I am going to change my commit message when I do the squash.

jingge-ibm:
I would be more than willing to test my code before merge. Where can I find documentations on how to write a compiler test?

andrewcraik:
Any update on this - it is a good change that provides better performance and we need to get it integrated to ensure we find any bugs sooner rather than later. @0xdaryl @jingge-ibm ?

jingge-ibm:
I have changed the commit message.

0xdaryl:
@jingge-ibm : Sadly, there isn't a great documentation story (yet) for writing unit tests.  The typical development pattern is to find an existing test to get the basic idea and then innovate as necessary.  I suggest you start with `fvtest/compilertest/tests/X86OpCodesTest.cpp`.

I will review your new commit today.

vijaysun-omr:
@0xdaryl anything more you would like done as part of this PR before it gets merged ? Thanks.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3403,146 +3403,434 @@ TR::Register *OMR::X86::TreeEvaluator::encodeUTF16Evaluator(TR::Node *node, TR::
...
+            0x00, 0x00, 0x00, 0x00,
+            };
+
+         uint8_t MASKOFSIZETWO[] =
+            {
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
+            0x00, 0x01, 0x00, 0x01,
+            };
```
0xdaryl: Minor performance point:  I think it would be better to mark these `static` to avoid the auto array initialization each time through here.

jingge-ibm: I made the change. Can you check it? @0xdaryl

vijaysun-omr:
Merging this since all checks have passed and comments from reviewers have been addressed.

PR #733, opened 2017-02-06
Value Propagation performance bug fix

mpirvu:
A feature that avoids doing VP on low frequency blocks (in order
to save compilation time) has a bug such that the JIT will not
do VP at hot/scorching opt levels for blocks that are NOT cold.
The condition should be reversed so that the JIT avoids VP for
basic blocks that ARE cold.


mpirvu:
@andrewcraik     FYI. Optimizer change.

andrewcraik:
LGTM @vijaysun-omr this one looks like one for you to own.

PR #734, opened 2017-02-07
WIP: Add option to disable exception asynccheck insertion

ncough:
This adds an option to disable exception asynccheck insertion,
replacing the prior used environment variable. Also shifts its
optimization enum out of the project.


andrewcraik:
This is a transient state while the new stuff is worked on. LGTM

PR #736, opened 2017-02-07
Stop adding exception edges for OSRBock of callee

cathyzhyi:
When merging callee cfg to caller cfg, inliner iterates through all
exception successor blocks of the original block containing that call
and add exception edges between these blocks and all the callee blocks.
However, the OSR blocks (both code block and catch block) of callee
should be excluded because the OSR blocks will transfer to VM and never
go back to any jitted block.


andrewcraik:
The key point here is that while the edges are not wrong the issue is the confusion they cause in analysis of locks. The unlock catch-all blocks generated for synchronized methods are the key issue here. All normal code needs to ensure they have edges to this block to ensure appropriate unlocking happens, but edges from the OSR infrastructure blocks confuse the lock analysis and cause the live ranges to be incorrectly calculated - the OSR infrastructure will not throw exceptions during the transition to the VM (once started) and so we should not pollute them with these extra edges.

LGTM.

vijaysun-omr:
The exception edges being skipped are valid to do so because callee blocks cannot take such edges to reach the OSR catch block in the caller directly; they will have to go through the OSR catch block in the callee and it's that callee OSR catch block that will go to the caller OSR code/catch block.

PR #737, opened 2017-02-07
Handle exceptions from genIL()

cathyzhyi:
Make the original try/catch block more general to include exceptions from both genIL() and ilgen opts

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1212,6 +1212,10 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
    // inliner modifies the generated tree so we need to regenerate it
    // each time...
    // replaced  "if (!_firstTreeTop || 1)" with "if (!_firstTreeTop || !comp->isPeekingMethod())"
+
+   TR::Optimizer *optimizer = NULL;
+   TR::Optimizer *previousOptimizer = NULL;
+   try {
    if (!_firstTreeTop || !comp->isPeekingMethod())
       {
       _firstTreeTop = 0;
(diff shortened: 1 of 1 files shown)


andrewcraik:
Despite the complexity of the diff I think this is a pretty straight forward change to increase the scope of the try-catch added to genIL so exceptions form IL generation as well as from ilgenopts will both be caught and if an exception is caught that only requires us to fail IL generation then we do so otherwise we let the exception escape as it would have previously.

LGTM

cathyzhyi:
For simplicity of diff, I split the pull request into 2 commits. All the essential changes are in the first commit and the second is just for adjusting the indents.

vijaysun-omr:
Thanks for splitting up the change to make it easier to review. 

Just for my information, under what circumstances can a TR::RecoverableILGenException be thrown ? Can this exception ever be thrown outside of the IL gen, i.e. can it ever be thrown by the optimizer per se ?

andrewcraik:
The TR::RecoverableILGenException itself is never thrown - there are specific subclasses which are thrown and those are the ones being caught. The current ones are Java specific and reflect cases where we cannot handle a given method because of the compilation environment, but if all we are doing is peeking/inlining we can keep going with the compile.

The intention is for derivatives of RecoverableILGenException to only be thrown in ilgen or ilgenopts (ie they can be thrown from optimizations that are in the ilgen opts strategy). They could be thrown outside this context (nothing explicitly stops that), but then they would be treated like any other exception and they would be caught at the top level leading to a compilation abort/JIT crash like all other exceptions not derrived from RecoverableILGenException.

The purpose of this is to let us fail an ilgen using indirect flow of control using throw without bringing down the entire compile as happens currently.

@vijaysun-omr if we can get this merged in the next day or so that would be great since we have a few things depending on it.

vijaysun-omr:
Thanks I will merge it in now.

PR #738, opened 2017-02-08
JitBuilder struct/union field address loading

Leonardo2718:
This change implements helper functions to IlGen code for loading the address of struct/union fields, as well as some testing for them.

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -817,6 +817,33 @@ IlBuilder::IndexAt(TR::IlType *dt, TR::IlValue *base, TR::IlValue *index)
    return address;
    }
 
+/**
+ * @brief Generate IL to load the address of a struct field.
+ *
+ * The address of the field is calculated by adding the field's offset to the
+ * base address of the struct. The address is also converted (type casted) to
+ * a pointer to the type of the field.
+ */
+TR::IlValue *
+IlBuilder::StructFieldInstanceAddress(const char* structName, const char* fieldName, TR::IlValue* obj) {
+   auto offset = typeDictionary()->OffsetOf(structName, fieldName);
+   auto ptype = typeDictionary()->PointerTo(typeDictionary()->GetFieldType(structName, fieldName));
+   auto addr = Add(obj, ConstInt64(offset));
+   return ConvertTo(ptype, addr);
+}
+
+/**
+ * @brief Generate IL to load the address of a union field.
+ *
+ * The address of the field is simply the base address of the union, since the
+ * offset of all union fields is zero.
+ */
+TR::IlValue *
+IlBuilder::UnionFieldInstanceAddress(const char* unionName, const char* fieldName, TR::IlValue* obj) {
+   auto ptype = typeDictionary()->PointerTo(typeDictionary()->UnionFieldType(unionName, fieldName));
+   return ConvertTo(ptype, obj);
+}
+
 TR::IlValue *
 IlBuilder::NullAddress()
    {
--- a/compiler/ilgen/IlBuilder.hpp
+++ b/compiler/ilgen/IlBuilder.hpp
@@ -204,7 +204,18 @@ public:
    TR::IlValue *AtomicAdd(TR::IlValue *baseAddress, TR::IlValue * value);
    void Transaction(TR::IlBuilder **persistentFailureBuilder, TR::IlBuilder **transientFailureBuilder, TR::IlBuilder **fallThroughBuilder);
    void TransactionAbort();
-   
+
+   /**
+    * `StructFieldAddress` and `UnionFieldAddress` are two functions that
+    * provide a workaround for a limitation in JitBuilder. Becuase `TR::IlValue`
+    * cannot represent an object type (only pointers to objects), there is no
+    * way to generate a load for a field that is itself an object. The workaround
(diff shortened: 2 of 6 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -770,6 +770,34 @@ IlBuilder::VectorLoadAt(TR::IlType *dt, TR::IlValue *address)
...
 
+/**
+ * @brief IlBuilder::StructFieldAddress IlGen to get the address of a struct field
+ * @param structName is the name of the struct type the field belongs to
+ * @param fieldName is the name of the field itself
+ * @param obj is an IlValue representing the instance (address) of the struct
+ * @return IlValue representing the address of the field in the struct
+ */
+TR::IlValue *
+IlBuilder::StructFieldAddress(const char* structName, const char* fieldName, TR::IlValue* obj) {
```
0xdaryl: I would prefer a more descriptive name, such as `StructFieldAddressWithinInstance` to make it clear what the address is relative to.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -770,6 +770,34 @@ IlBuilder::VectorLoadAt(TR::IlType *dt, TR::IlValue *address)
...
+/**
+ * @brief IlBuilder::StructFieldAddress IlGen to get the address of a union field
+ * @param unionName is the name of the union type the field belongs to
+ * @param fieldName is the name of the field itself
+ * @param obj is an IlValue representing the instance (address) of the union
+ * @return IlValue representing the address of the field in the union
+ */
+TR::IlValue *
+IlBuilder::UnionFieldAddress(const char* unionName, const char* fieldName, TR::IlValue* obj) {
+   auto ptype = typeDictionary()->PointerTo(typeDictionary()->UnionFieldType(unionName, fieldName));
```
vijaysun-omr: Am I reading this correctly ? 

Was the intent to get the offset of the field inside the union and calculate the address by adding to the obj address (as in the struct case) ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -770,6 +770,34 @@ IlBuilder::VectorLoadAt(TR::IlType *dt, TR::IlValue *address)
...
+ * @param structName is the name of the struct type the field belongs to
+ * @param fieldName is the name of the field itself
+ * @param obj is an IlValue representing the instance (address) of the struct
+ * @return IlValue representing the address of the field in the struct
+ */
+TR::IlValue *
+IlBuilder::StructFieldAddress(const char* structName, const char* fieldName, TR::IlValue* obj) {
+   auto offset = typeDictionary()->OffsetOf(structName, fieldName);
+   auto ptype = typeDictionary()->PointerTo(typeDictionary()->GetFieldType(structName, fieldName));
+   auto addr = Add(obj, ConstInt64(offset));
```
vijaysun-omr: While this is "a" correct way to represent the address of a field, I do want to ask about how pervasively you intend to use this api. 

One way to access the field in question is to simply generate an indirect load or store node with the child node being the address node returned by this routine you added using a symbol reference that itself has an offset of zero. 

Another equivalent way to access a struct field is to use a symbol reference having an (potentially non zero) offset corresponding to the struct field on the indirect load or store node and use the "obj" TR::IlValue as the child node. 

The latter is a less verbose representation since we have fewer nodes: no constant offset node and no add node either.

Leonardo2718: I see your point.

I think I understand how your second approach works in Trees, but I'm not sure how it maps to existing JitBuilder abstractions.

Currently, TR::IlValue cannot represent a struct "obj", only a pointer to an obj. This makes it impossible to load a field that is itself a struct. The purpose of this API is to load the address of the field so that an indirect load/store can then be done.

Admittedly, this is a dirty hack and I should document it is as such. I'm only contributing this change as I've found some use for it in my own project and it's general enough to be useful in other projects as well. Perhaps @mstoodle should have a say in this. @charliegracie might also have some thoughts.

compiler/ilgen/IlBuilder.cpp
Leonardo2718: In the case of unions, all fields have an offset of 0. So, it's enough to simply cast the pointer to the union to a pointer to the field type; no calculation is require.

compiler/ilgen/IlBuilder.cpp
0xdaryl: @Leonardo2718 , please document the fact that this change is less than ideal, but it is a reasonable step forward for JitBuilder.  It may be revisited in the future as part of many other JitBuilder enhancements.  When @mstoodle returns we can review the change with him to get his opinion.  Vijay will merge after the comment is added.

0xdaryl: Also, please address my function naming comment above (or convince me otherwise).

compiler/ilgen/IlBuilder.cpp
vijaysun-omr: That makes sense. Thanks for clarifying.

vijaysun-omr:
I am merging this. @Leonardo2718 can add a link to the issue that Daryl suggested in one of his review comments but it should not hold up the delivery of these changes.

Leonardo2718:
I have opened issue #752 to revisit this change and try to develop a better solution to the problem.

PR #740, opened 2017-02-08
Exception asynccheck insertion in ilgen

ncough:
This removes options and enums added for the
ExceptionAsyncCheckInsertion optimization pass.
It also deals with situations where exception handlers
are cloned or split, to ensure the asynccheck remains
in the right place or is duplicated as needed.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1998,8 +1998,31 @@ OMR::ResolvedMethodSymbol::detectInternalCycles(TR::CFG *cfg, TR::Compilation *c
                            }
                         }
 
-                     //rip out trees
-                     clonedCatch->getEntry()->join(clonedCatch->getExit());
+                     TR::TreeTop *retain = clonedCatch->getEntry();
+                     // As this method is performed soon after ilgen, the exception handler
+                     // may be prepended with an asynccheck and pending pushes
+                     // These should be retained in the copy, so skip them when ripping out trees
+                     if (comp->getHCRMode() == TR::osr)
+                        {
+                        TR::TreeTop *next = retain->getNextTreeTop();
+                        if (next && next->getNode()->getOpCodeValue() == TR::asynccheck)
+                           {
+                           TR_ByteCodeInfo osrPointBCI = next->getNode()->getByteCodeInfo();
+                           retain = next;
+                           next = next->getNextTreeTop();
+                           // Pending pushes should have the same BCI as the OSR point
+                           while (next
+                                 && next->getNode()->getOpCode().isStoreDirect()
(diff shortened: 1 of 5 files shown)


andrewcraik:
I think this looks good to merge - @ncough can you please make sure this is rebased since it was created a little while ago. @vijaysun-omr a set of small changes to keep asynchecks added at the start of catch blocks with the catch block when splitting/doing other operations. This is needed for OSR correctness and should only really affect OSR modes where asynchecks in catch blocks are needed.

ncough:
This branch has been rebased.

vijaysun-omr:
Could you please explain why the ExceptionAsyncCheckInsertion optimization pass needed to be deleted and this code added as part of IL gen ?

andrewcraik:
The main issue here is that we only need the async checks in real catch handlers that exist in the user's program. Ilgen also manufactures catch blocks, in Java this could include catch handlers to unlock method scoped locks etc, and we do not want the overhead of asynchecks in those methods. Further the implementation in ilgen was much more clean than as a separate pass - it is only a couple of lines of code in the two places that handle generation of catch blocks.

ncough:
A further problem existed with adding OSR points for the asyncchecks, as the optimization pass lacked all the necessary state to add a transition for these points. This state was available in the IL gen walk, and so adding the asyncchecks earlier allowed for the existing infrastructure to correctly manage them.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1998,8 +1998,24 @@ OMR::ResolvedMethodSymbol::detectInternalCycles(TR::CFG *cfg, TR::Compilation *c
...
+                     TR::TreeTop *retain = clonedCatch->getEntry();
+                     // As this method is performed soon after ilgen, the exception handler
+                     // may be prepended with an asynccheck and pending pushes
+                     // These should be retained in the copy, so skip them when ripping out trees
+                     if (comp->getHCRMode() == TR::osr)
+                        {
+                        TR::TreeTop *next = retain->getNextTreeTop();
+                        while (next &&
+                               (next->getNode()->getOpCodeValue() == TR::asynccheck ||
+                               (next->getNode()->getSymbolReference() &&
```
vijaysun-omr: We should probably call hasSymbolReference() on the opcode first before getting the symbol reference (to be safe). 
Might also be safer to more explicitly check for an auto store rather than the somewhat more generic CP index < 0 test here.

ncough: Pending pushes are now detected by matching their bytecode index against the OSR point's.

ncough:
Marked as WIP whilst OSR transitions are tested.

vijaysun-omr:
Thanks for making the changes to tighten up the condition for detecting the stores based on the prior review/discussion. Merging.

PR #741, opened 2017-02-08
Use proper macro to detect Windows specific code

0dvictor:
According to the build files (<PATH>\build\toolcfg\host\win.mk),
"WINDOWS" is the correct macro to detect Windows OS.


0dvictor:
Some notes about the change:
1. Macro "WIN32" was expected to be defined on both 32-bit and 64-bit Windows, same as macro "WINDOWS".
2. Macro "WIN64" was expected to be defined only on 64-bit Windows; however, in the scenario of compiler/x/amd64/runtime/AMD64cpuid.inc, checking "WINDOWS" is sufficient since the file is only used by 64-bit code.

andrewcraik:
LGTM

PR #742, opened 2017-02-08
Add synchronizedFieldLoad codegen symbol

fjeremic:
Add a new codegen inlined method call symbol which will model a codegen inlined sequence for reducing a synchronized field load using hardware instructions which support loading memory by means of block-concurrent interlocked fetch. In addition to this we add a new option to support such a feature and deprecate a similar but less documented and intrusive option.

### DIFF
--- a/compiler/codegen/OMRInstruction.cpp
+++ b/compiler/codegen/OMRInstruction.cpp
@@ -191,20 +191,10 @@ Instruction::move(TR::Instruction *newLocation)
    {
    TR_ASSERT(self() != newLocation, "An instruction cannot be its own predecessor");
 
-   TR::Instruction *prev = self()->getPrev();
-   TR::Instruction *next = self()->getNext();
-
-   if (prev)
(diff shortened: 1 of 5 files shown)


fjeremic:
@0dvictor you have recently added some codegen inlined symbols. It would be great if you could review this PR.

0xdaryl:
Is d1565a5 needed to fix a problem relating to the other commits in this PR?  I'm just trying to understand the relationship to adding the new codegen symbol.

fjeremic:
> Is d1565a5 needed to fix a problem relating to the other commits in this PR? I'm just trying to understand the relationship to adding the new codegen symbol.

@0xdaryl yes this commit addresses an issue in what I can see is the only consumer of this API within an internal consumer of OMR.

0xdaryl:
@0dvictor , any concerns?

0dvictor:
LGTM

PR #743, opened 2017-02-09
Add support for new core specifiers on Linux

dnakamura:
- Add support for the %E,%i,%I,%P core specifier in Linux (see core(5)).
%I and %P use wildcard matching since it is impossible to determine
PID/TID value in the root namespace.

 - Existing wild card matching only supports 1 wildcard field per core
file. Added check to ensure that only 1 is specified in core_pattern.

Task: 121414

### DIFF
--- a/port/linux/omrosdump_helpers.c
+++ b/port/linux/omrosdump_helpers.c
@@ -30,6 +30,7 @@
 #if defined(LINUX)
 #include <sys/prctl.h>
 #include <linux/prctl.h>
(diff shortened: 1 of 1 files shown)


charliegracie:
@dnakamura Can you update your commit message to remove the task you are referencing as it is too large a number to be referencing an OMR GitHub Issue.  Thanks

dnakamura:
@charliegracie should be good now

PR #744, opened 2017-02-09
Fix inconsistent switching for regular / nonZero TLHs

dmitripivkine:
Fixing the problem in allocation path when refreshed TLH is requested
for nonZero TLH but actually taken for regular TLH. As a result an
object would not be allocated in nonZero TLH but out-of-TLH. The heap
memory taken for refreshed TLH would be lost and also might create a
problem with heap walk ability. This problem exists on non-Batch Clear
TLH platforms


charliegracie:
Hi @dmitripivkine  could you limit the first line in the commit message to 50 characters or less.  Also the commit appears to be authored by @vijaysun-omr and committed by you.  Is that correct?

dmitripivkine:
Yes, thank you. Postpone it for now. Currently have problem with tools. Will fix Monday.

PR #746, opened 2017-02-09
Require OSRGuard NOPing in profiling compilations

ncough:
Adds OSRGuard to the kinds of virtual guards that will be
NOPed when conducting a profiling compilation.


ncough:
@andrewcraik

vijaysun-omr:
Safe change since it's not valid to leave an explicit OSR guard without NOPing under any circumstances (including in a profiling compilation).

PR #749, opened 2017-02-10
WIP: Improve VP Known Object Exploitation - Part I

liqunl:
Fold the following calls into constants when all of their arguments are
known objects.

Class.getModifiersImpl
Class.isAssignableFrom
Class.isInterface
J9VMInternals.identityHashCode

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -653,7 +653,6 @@ static bool containsUnsafeSymbolReference(TR::ValuePropagation *vp, TR::Node *no
          return true;
          }
 
-
       if (node->getSymbolReference() == vp->getSymRefTab()->findInstanceShapeSymbolRef() ||
           node->getSymbolReference() == vp->getSymRefTab()->findInstanceDescriptionSymbolRef() ||
           node->getSymbolReference() == vp->getSymRefTab()->findDescriptionWordFromPtrSymbolRef() ||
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik This change depends on the j9 change.

andrewcraik:
This code really belongs somewhere in trj9 - we probably just need to add an extension point and the bulk of this can move down to j9.

liqunl:
Talked to Xiaoli, she said I can take simplifier as an example and add a j9 vp myself. So eventually, this PR will only contain a declaration of the extension point and a call to it in constrainCall.

liqunl:
After looking into creating J9 VP, I found that it requires quite a bit of work and I'm really not the right person to do it. So giving the fact that there is not a place in j9 to put the code in, and we're not adding ifdefs of J9_PROJECT_SPECIFIC, we'll leave the code in OMR.

liqunl:
This PR will be ready once the J9 MR is merged.

liqunl:
@andrewcraik Since the J9 part is merged, can you review this change (if you haven't done so) and put a comment if you think it's good?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -5239,6 +5238,98 @@ TR::Node *constrainCall(TR::ValuePropagation *vp, TR::Node *node)
...
+             && classChildConstraint->getClassType()->asFixedClass())
+            {
+            int32_t isInterface = TR::Compiler->cls.isInterfaceClass(vp->comp(), classChildConstraint->getClass());
+            TR::VPConstraint *isInterfaceConstraint = TR::VPIntConst::create(vp, isInterface);
+            vp->replaceByConstant(node, isInterfaceConstraint, classChildGlobal);
+            TR::DebugCounter::incStaticDebugCounter(vp->comp(), TR::DebugCounter::debugCounterName(vp->comp(), "constrainCall/%s", "isInterface"));
+            return node;
+            }
+         }
+      else if (isIsAssignableFrom)
```
andrewcraik: You can probably get more milage here - we don't need both classes to be fixed classes - it may be sufficient to have a fixed class and a class constraint holding a top type...

compiler/optimizer/VPHandlers.cpp
```diff
@@ -5239,6 +5238,98 @@ TR::Node *constrainCall(TR::ValuePropagation *vp, TR::Node *node)
       bool isFloatAbs =   (symbol->getRecognizedMethod() == TR::java_lang_Math_abs_F);
       bool isDoubleAbs =  (symbol->getRecognizedMethod() == TR::java_lang_Math_abs_D);
 
+      bool isGetModifiersImpl = (symbol->getRecognizedMethod() == TR::java_lang_Class_getModifiersImpl);
```
andrewcraik: This would probably look nicer as a switch case and would run a little bit faster too.

andrewcraik:
I think this looks good to merge.

liqunl:
@vijaysun-omr This PR has passed all the checks and has been reviewed, can you merge it when you're free?

liqunl:
This PR no long needed.

PR #750, opened 2017-02-10
Fix  inliner suitableForRemat

cathyzhyi:
The second part of logic to decide suitableForRemat should
only apply when the first part decide suitableForRemat is true


vijaysun-omr:
Fixes a clear flaw in the logic that was making us do guarded arg remat in a run with the option to disable it set. I will merge as soon as the checks pass.

vijaysun-omr:
The checks had in fact passed for this before Yi cleaned up something in a comment. So I am merging now.

PR #753, opened 2017-02-13
 Update alloc pointer only after winning fwd-ing op

amicic:
Concurrent Scavenger only. If a thread loses the race during atomic
forwarding operation, it can still re-use the pre-reserved space for
subsequent object copies. This will partially reduce the problem of duplicate
copies (in Concurrent Scavenger, we still copy the object before the
forwarding operation).


charliegracie:
@amicic Can you update the first line of the commit message to be 50 characters or less as per the Contributing.md.  Thanks

fjeremic:
> @amicic Can you update the first line of the commit message to be 50 characters or less as per the Contributing.md. Thanks

While amending the commit might as well fix the grammar in the first sentence. Perhaps:

> If a thread loses the race during a forwarding operation, ...

PR #758, opened 2017-02-14
Provide managed destruction of Region allocations

lmaisons:
Under some circumstances, we will want to Region-allocate objects that
require having their destructors invoked when they are destroyed.
Currently there is no mechanism for doing this.  This change set
provides such a mechanism for types that would allow themselves to be
copy-constructed into the Region.

In order to provide this new facility, the allocate function is extended
via a new function template such that invocations passing a parameter
that is not coercable to size_t are treated as instead providing an
object prototype with which to construct a new instance via copying.
When the Region is itself destroyed, it will iterate through these
managed objects, destroying them in LIFO ordering.

### DIFF
--- a/compiler/env/Region.cpp
+++ b/compiler/env/Region.cpp
@@ -28,7 +28,8 @@ Region::Region(TR::SegmentProvider &segmentProvider, TR::RawAllocator rawAllocat
    _segmentProvider(segmentProvider),
    _rawAllocator(rawAllocator),
    _initialSegment(_initialSegmentArea.data, INITIAL_SEGMENT_SIZE),
-   _currentSegment(TR::ref(_initialSegment))
+   _currentSegment(TR::ref(_initialSegment)),
+   _lastDestructable(NULL)
    {
    }
 
@@ -36,12 +37,25 @@ Region::Region(const Region &prototype) :
    _segmentProvider(prototype._segmentProvider),
    _rawAllocator(prototype._rawAllocator),
    _initialSegment(_initialSegmentArea.data, INITIAL_SEGMENT_SIZE),
-   _currentSegment(TR::ref(_initialSegment))
+   _currentSegment(TR::ref(_initialSegment)),
+   _lastDestructable(NULL)
    {
    }
 
 Region::~Region() throw()
    {
+   /*
+    * Destroy all object instances that depend on the region
+    * to manage their lifetimes.
+    */
+   Destructable *lastDestructable = _lastDestructable;
+   while (lastDestructable)
+      {
+      Destructable * const currentDestructable = lastDestructable;
+      lastDestructable = currentDestructable->prev();
+      currentDestructable->~Destructable();
+      }
+
    for (
       TR::reference_wrapper<TR::MemorySegment> latestSegment(_currentSegment);
       latestSegment.get() != _initialSegment;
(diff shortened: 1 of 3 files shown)


lmaisons:
Updated PR with commit that doesn't include unrelated fix

0xdaryl:
I'd appreciate @mpirvu 's review of this as well.

mgaudet:
Would it be possible to get some documentation that describes how one opts into destruction semantics? 

Just reading through the code I get to this, 

```c++
TR::compiler->regionAllocator.allocate<Typename>(instanceOfType);
```

but that doesn't seem right to me.

lmaisons:
It's more or less that, generally without having to explicitly identify the template specialization:
```cpp
cfg->trMemory()->currentStackRegion().allocate(TR_StackForRevPost(n, ++edge));
```
(In case it's not obvious TR_StackForRevPost is a type being used to create a temporary)

mpirvu:
I am ok with the code, but in my opinion it needs comments and examples of intended usage.
Do you think that most of the allocations will be done using the pattern above where a temporary is created for the sole purpose of having an object to copy from? I am concerned about overhead obviously. Move semantics in C++11 is supposed to deal nicely with overhead from temporaries, but I don't think it will work here.

lmaisons:
@mpirvu I agree about move constructors; I tried various forms of passing by value / passing by const reference[1] unfortunately none of the options elided all the secondary copies.

I added comments to the header.  Are they sufficient or is more information required?

[1]http://ideone.com/eeehqI

mpirvu:
Maybe the comment could be more clear by adding "Object instances allocated in this form will be destroyed in LIFO order **when the owning region gets destroyed**".
I didn't completely understand the comment regarding the lifespan of objects allocated in R0 vs R1. In my mind the objects go away when the region goes away, therefore what matters is which region goes destroyed first.

Regarding the move semantics I remember reading that the object passed in must not be const because it is going to be modified during the move process. I have no practical experience with move operations in C++11 though. I was thinking that the move may not work because the end result must reside inside the region, so the temporary must also be created in the region to avoid copying.

lmaisons:
I don't believe we can rely on moves at all while our various compilers are at various stages of incompatibility, which is why I haven't actively sought out to explicitly make them available.

What I thought would be an option was using pass-by-value + copy elision, but that seems to be fickle / generally unreliable (the ideone links were my attempts to re-create and better understand how they would operate under this context).

The comment about lifetimes is there because, when I was considering the implications resulting from a second Region being allocated 'into' a first, I found them surprising.  My naive understanding was that objects in the top region would universally outlive the objects in the sub-Region (like would be the case with stack memory) but that's not actually the case.  This ordering matters if there are interdependencies between Region allocated objects that are relevant during destruction.

lmaisons:
Latest push should include your requested clarification.

0xdaryl:
@mpirvu , please post any concerns with @lmaisons latest commit.

mpirvu:
Another thing caught my attention
```static const size_t INITIAL_SEGMENT_SIZE = 8;```
This used to be 4096. Any reason for the change from 4096 to 8?

lmaisons:
>  This used to be 4096. Any reason for the change from 4096 to 8?
Yikes, thanks for catching that!  I changed that to try and help catch the cause of the debug STL failures.  I thought I had changed it back, but I clearly had not.

mpirvu:
I have no more comments and the change can be delivered.

lmaisons:
@mpirvu Thank you for the detailed and thorough review.

PR #761, opened 2017-02-16
Improve IVA's usesUnchangedValueInLoopTest analysis

jdmpapin:
In induction variable analysis (IVA), when `analyzeExitEdges` sees that `valueNode` is a load, it will now determine the point of evaluation of that load relative to the pertinent stores by iterating through the extended block's nodes in postorder. For this purpose, and for symmetry, the set of node iterators is filled out with a new `PostorderNodeIterator`.

### DIFF
--- a/compiler/infra/ILWalk.cpp
+++ b/compiler/infra/ILWalk.cpp
@@ -138,11 +138,6 @@ TR::PreorderNodeIterator::PreorderNodeIterator(TR::TreeTop *start, TR::Compilati
    push(start->getNode());
    }
 
-TR::Node *TR::PreorderNodeIterator::currentNode()
-   {
-   return _stack.top()._node;
-   }
-
 bool TR::PreorderNodeIterator::alreadyBeenPushed(TR::Node *node)
    {
    return _checklist.contains(node);
@@ -197,7 +192,7 @@ void TR::PreorderNodeIterator::stepForward()
       }
    }
 
-void TR::PreorderNodeIterator::logCurrentLocation()
+void TR::NodeIterator::logCurrentLocation()
    {
    // TODO: Log even without an _opt.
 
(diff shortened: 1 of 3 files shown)


Leonardo2718:
Could you specify what IVA stands for in your description, for those of us less familiar with this part of the code :)

vijaysun-omr:
I would like @andrewcraik to review this as well.

andrewcraik:
LGTM - this is an improvement to make sure that we only transform when safe to do so.

andrewcraik:
@jdmpapin can you please resolve the conflicts on this branch and then let @vijaysun-omr know so we can get this merged assuming he is ok with the change?

jdmpapin:
@vijaysun-omr This is updated now to resolve the conflict, which was just whitespace.

vijaysun-omr:
Thanks, I started reviewing this but need more time before I finish it.

vijaysun-omr:
Looks good. The usage of the post order node iterator certainly makes the code much cleaner for this kind of "happened in which order" sort of a problem.

vijaysun-omr:
I suppose the other alternative to solve the problem you showed in the 3rd commit is to force local CSE to run and ensure that we never have the situation where we have 2 different loads of the same variable before the store. Did you consider that ?

jdmpapin:
Relying on local CSE for this purpose would complicate the use of IVA by imposing an unnecessary trade-off of the compile-time overhead for an additional opt pass, levied quite broadly, against the benefit of more precise analysis in IVA, gained much more narrowly. There would be the potential for such a trade-off before each occurrence of IVA in an opt strategy, if there might be loads like these that aren't commoned.

I call such trade-offs "unnecessary" because, by using the improvement put forth here, we can get the benefit without adding any opt passes, and therefore without having to determine whether it's worth it to add them. This isn't to say that there's absolutely no increase in compile-time; the updated analysis will visit more nodes than previously. However, this code only runs in limited circumstances.

By using local CSE, we would get to avoid changing IVA, but I'd like to think that this change is an improvement even just in clarity. Please feel free to contradict me though :smile:

vijaysun-omr:
Point taken. 

Induction variable analysis is used by enough different passes nowadays (not always the case in the past) that we would have had to add a pass of local CSE at several points in the opt strategy. This is probably true even at warm (the hotness level where compile time matters most in the real world). The "unnecessary coupling" point is also a fair one from a conceptual viewpoint.

My suggestion was based on looking at the non trivial changes you had to do here and thinking that adding an extra pass of local CSE would have been (much) less risky. But given that you have already done the work to implement the enhancement in IVA and having reviewed the (as clean as possible) way in which you did so, we should just go ahead and accept your changes. 

So I am merging this in.

PR #763, opened 2017-02-16
Inhibit array splitting in Scavenger backout

ktbriggs:
This was previously done for the RS overflow case but also needs to
be done in non-RS overflow backout. This fix covers both cases. The
boolean backOutFlag member has been extended to 3 states to distinguish
backOutFlagsRaised and backOutStarted conditions. Array splitting is 
inhibited only in backOutStarted state.

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -345,7 +345,7 @@ void
 MM_Scavenger::masterSetupForGC(MM_EnvironmentStandard *env)
 {
 	/* Make sure the backout state is cleared */
-	setBackOutFlag(env, false);
+	setBackOutFlag(env, backOutFlagCleared);
 
 #if !defined(OMR_GC_CONCURRENT_SCAVENGER)
 	_rescanThreadsForRememberedObjects = false;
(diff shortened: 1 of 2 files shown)


charliegracie:
@ktbriggs Can you limit the first line of the commit message to 50 characters please?  Also why did so many lines in Scavenger.hpp change? Is it the line endings?

ktbriggs:
@charliegracie I added another commit pending Dmitri's review. Will squash and shorten commit message when change is approved and tested. As for Scavenger.hpp diff, yes it appears to be line endings, append "?w=1" to githib diff url to suppress lineend differences. Not sure how lineends got whacked -- usually egit/eclipse takes care of that.

ktbriggs:
@charliegracie this has been reviewed by Aleks & Dmitri and is ready (squashed, reworded) to go.

charliegracie:
@ktbriggs Scavenger.hpp still has every line changed.  Can you resolve this and push again?  I know I can ignore the whitespace changes but it is better to not have them.

ktbriggs:
@charliegracie Line endings in Scavenger.hpp have been fixed and squashed into last commit.

PR #765, opened 2017-02-18
Teach arrayset evaluators the size of compressed refs

jdmpapin:
With compressed references, the size of a reference in the heap does not
agree with `Node::getSize` for reference-typed nodes.


vijaysun-omr:
This looks good to me as a straightforward bug fix. I would like @andrewcraik to review as well.

andrewcraik:
Looks good to me.

vijaysun-omr:
Merging since it's a clear bug fix for compressed references.

Issue #767, opened 2017-02-21
Vector Evaluator fixes and improvements on p and z


IBMJimmyk:
Vector Evaluator fixes and improvements on p and z

Some vector evaluators on p currently transfer data between vector registers and GPRs by going through memory. The proposed changes will change them to use instructions that directly move data between the two register types.

This change will also fix some bugs such as:
-vneg Evaluator on Power is incorrect when an input value is the minimum signed 32 bit int (-2147483648). vsubuwm should be used instead of vsumsws.
-vand, vor and vxor on Power may attempt to use a VMX instruction on a VSX register. VSX instruction should be used instead.
-getvelem and vsetelem returned incorrect values on z when working with 64 bit data on a 31 bit system. The cause was incorrect handling of pair registers.
-generateMvFprGprInstructions on Power does not work under Little Endian. Support needs to be added.

Other various improvements:
-extra entries in convertScalarToVector to handle cases like frem and lxor.
-updated getSupportsOpCodeForAutoSIMD on p and z to reflect the current state of vector evaluator support.

mstoodle:
In future, I would prefer more focused issues rather than this kind of "grab bag" approach. Its harder to see what's actually being addressed by an issue with multiple seemingly unrelated "fixes" inside. It's not that hard to open another issue :) .

IBMJimmyk:
Okay. Even when I was writing this I could tell it was a bit all over the place. I'll definitely split something like this up into multiple Issues in the future.

PR #768, opened 2017-02-21
Vector Eval bug fixes and perf improvements

IBMJimmyk:
Added more entries to convertScalarToVector such as frem, drem and lxor which are converted to vrem, vrem and vxor.

Changed vsplats evaluator on Power to use direct move instructions to transfer data from GPRs to vector registers instead of transferring the data through memory. This applies to splatting Int32 and Int64 data types since Double and Float did not transfer data through memory.

Fixed a bug in the vsplats evaluator on Power for Float data where a TR_VSX_VECTOR register was used for the result register. It was changed to a TR_VRF register because float data may use VMX instructions in other places.

generateMvFprGprInstructions normally generates direct move instructions to move data between GPRs and FPRs on Power. It has a fallback path which transfers the data through memory instead when direct move isn't supported. The fallback path was modified to support little endian power.

getSupportsOpCodeForAutoSIMD on Power and z were modified to more accurately represent which vector evaluators are currently supported and for which data types.

vand, vor and vxor evaluators on Power were expanded to support Int64 data type. Int64 vectors uses the VSX registers so the previously used VMX instruction was invalid for Int64 data.

Changed getvelem evaluator on Power to use direct move instructions to transfer data from vector registers to GPRs instead of transferring the data through memory.

Fixed bug with vneg evaluator on Power. Using the vsubsws would produce the wrong result when the input was -2147483648 (minimum signed 32 bit int) due to being a saturate instruction. It was changed to vsubuwm which is a modulo instruction.

Fixed a bug with getvelem evaluator on z. When operating on 64 bit data on a 31 bit system, the result is supposed to be returned in a pair register. This was not happening and has now been corrected.

Fixed a bug with vsetelem evaluator on z. When operating on 64 bit data on a 31 bit system, the wrong parts of the vector register were being read out into the pair register due to the index calculation being wrong. This was fixed.

Issue: #767

### DIFF
--- a/compiler/p/codegen/FPTreeEvaluator.cpp
+++ b/compiler/p/codegen/FPTreeEvaluator.cpp
@@ -562,11 +562,10 @@ TR::Register *OMR::Power::TreeEvaluator::dloadEvaluator(TR::Node *node, TR::Code
    return TR::TreeEvaluator::dloadHelper(node, cg, tempReg, TR::InstOpCode::lfd);
    }
 
-
-
 TR::Register *OMR::Power::TreeEvaluator::vsplatsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    TR::Node *child = node->getFirstChild();
+   static bool disableDirectMove = feGetEnv("TR_disableDirectMove") ? true : false;
 
    if (node->getDataType() == TR::VectorInt8)
       {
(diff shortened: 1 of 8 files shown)


compiler/p/codegen/FPTreeEvaluator.cpp
```diff
@@ -562,11 +562,10 @@ TR::Register *OMR::Power::TreeEvaluator::dloadEvaluator(TR::Node *node, TR::Code
    return TR::TreeEvaluator::dloadHelper(node, cg, tempReg, TR::InstOpCode::lfd);
    }
 
-
-
 TR::Register *OMR::Power::TreeEvaluator::vsplatsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    TR::Node *child = node->getFirstChild();
+   static bool disableDirectMove = feGetEnv("TR_disableDirectMove") ? true : false;
```
0xdaryl: Is there a reason why this is an environment variable rather than an Option?

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1676,17 +1676,47 @@ TR::Register *OMR::Power::TreeEvaluator::vimaxEvaluator(TR::Node *node, TR::Code
...
+   switch (node->getDataType())
+      {
+      case TR::VectorInt8:
+      case TR::VectorInt16:
+      case TR::VectorInt32:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::vxor);
+      case TR::VectorInt64:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxlxor);
+      default:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxlxor);
```
0xdaryl: These cases only differ by the opcode.  Can you rework to only have one call to `inlineVectorBinaryOp` which will reduce code footprint?

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1803,6 +1833,245 @@ TR::Register *OMR::Power::TreeEvaluator::vigetelemEvaluator(TR::Node *node, TR::
 
 TR::Register *OMR::Power::TreeEvaluator::getvelemEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
+   static bool disableDirectMove = feGetEnv("TR_disableDirectMove") ? true : false;
```
0xdaryl: Same comment as above for environment variable.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1676,17 +1676,47 @@ TR::Register *OMR::Power::TreeEvaluator::vimaxEvaluator(TR::Node *node, TR::Code
...
+   switch (node->getDataType())
+      {
+      case TR::VectorInt8:
+      case TR::VectorInt16:
+      case TR::VectorInt32:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::vor);
+      case TR::VectorInt64:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxlor);
+      default:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxlor);
```
0xdaryl: These cases only differ by the opcode. Can you rework to only have one call to inlineVectorBinaryOp which will reduce code footprint?

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1676,17 +1676,47 @@ TR::Register *OMR::Power::TreeEvaluator::vimaxEvaluator(TR::Node *node, TR::Code
...
+   switch (node->getDataType())
+      {
+      case TR::VectorInt8:
+      case TR::VectorInt16:
+      case TR::VectorInt32:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::vand);
+      case TR::VectorInt64:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxland);
+      default:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxland);
```
0xdaryl: These cases only differ by the opcode. Can you rework to only have one call to inlineVectorBinaryOp which will reduce code footprint?

compiler/p/codegen/FPTreeEvaluator.cpp
IBMJimmyk: A environment variable was used there to match the one used inside generateMvFprGprInstructions() in tr.open/omr/compiler/p/codegen/GenerateInstructions.cpp. If TR_disableDirectMove is set, my code changes do not want to call generateMvFprGprInstructions anymore. The reason for this is if TR_disableDirectMove is set, a code path that's meant for float registers rather than vector registers in general will be used. Maybe a better solution would be to change generateMvFprGprInstructions to properly support all vector registers?

compiler/p/codegen/OMRTreeEvaluator.cpp
IBMJimmyk: I can do that. I will drop the Int64 case and just have the default case cover Int64 as well. I will do this for all three locations (vand, vor, vxor).

IBMJimmyk:
For testing I created small unit tests and logged the method with the test in it. I confirmed that both the correct result was outputted by the program and that the expected instructions were generated in the method log file. I tested multiple data types each of the modified evaluators.

On Power, I tested:
-Int64 - vor, vand, vxor
-Int32 - vsplats - direct move and memory move
-Int64 - vsplats - direct move and memory move
-Int32 - getvelem - direct move and memory move, constant and non-constant index
-Int64 - getvelem - direct move and memory move, constant and non-constant index
-Float - getvelem - direct move and memory move, constant and non-constant index
-Double - getvelem - direct move and memory move, constant and non-constant index
-Int32 - vneg - specifically tested the previously bad input of -214748648

On z, I tested:
-Int32 - vrem
-Int64 - vrem
-Int64 - vor, vand, vxor
-Int64 - getvelem, 64 bit and 31 bit system

While I fixed a bug in vsetelem for handling 64bit data on 31 bit z, I don't think we ever actually use that code path so I wasn't able to test it. Because of that, I made sure to review that section more throughly.

In the future, I'll try to break up changes like this into more organized pieces.

compiler/p/codegen/FPTreeEvaluator.cpp
0xdaryl: Possibly.  Or, both functions could upgrade to a true Option rather than an environment variable, which is my preference.  However, given the overwhelming precedent to use both in this codebase already, I'm fine with it.

compiler/p/codegen/OMRTreeEvaluator.cpp
0xdaryl: FYI.. I believe this comment is still outstanding in your latest commit.

IBMJimmyk: Unless I misunderstood what you meant, I think the commit I added 2 days ago addressed this: https://github.com/eclipse/omr/pull/768/commits/ee80fe0e4c8668cf0b95d8a0c811a6007a949f30

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1676,17 +1676,41 @@ TR::Register *OMR::Power::TreeEvaluator::vimaxEvaluator(TR::Node *node, TR::Code
...
    {
-   return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::vand);
+   switch (node->getDataType())
+      {
+      case TR::VectorInt8:
+      case TR::VectorInt16:
+      case TR::VectorInt32:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::vand);
+      default:
+         return TR::TreeEvaluator::inlineVectorBinaryOp(node, cg, TR::InstOpCode::xxland);
```
0xdaryl: I think my comments around these functions were all lost, likely when a new commit was pushed.

My comment was to only have a single function call since the cases only differ by the opcode.  The switch should be to choose a different opcode.  You will generate far less code with only a single call.  I suggest that these case statements (and the ones below) all be restructured this way.

IBMJimmyk: Okay, that makes sense. I'll try reducing the number of different functions.

IBMJimmyk: My latest commit should address this issue. There will only be one call to inlineVectorBinaryOp per evaluator now.

PR #769, opened 2017-02-22
Use proper CMP opcode for divide check

0dvictor:
The old code generated a long compare by modifying
REX prefix from an int compare instruction, which is
quite confusing and hacky. This changeset correctly
use int or long compare opcode.


andrewcraik:
LGTM

andrewcraik:
@0xdaryl a fairly simple x86 codegen change for you to consider.

PR #771, opened 2017-02-22
Canonicalize while loops with post-inc/decrement

jdmpapin:
Skipping past `treetop` allows while loops with tests of the following
form to be canonicalized:

        treetop
    n0n   iload i
        istore i
          iadd
            iload i
            iconst -1
        ificmple --> exit
    n0n   ==>iload i
          bound


andrewcraik:
LGTM

andrewcraik:
@vijaysun-omr a fairly simple canonicalizer improvement for your consideration.

vijaysun-omr:
The proposed change seems quite safe to me since it only increases the cases that get considered a while loop while doing canonicalization. The test as it stands seems like it would skip over not just side effect free treetops like the one shown in the first comment but also ones with side effects. However, this kind of skipping over treetops with potential side effects underneath would just make this code consistent with what seems to be done for checks anyway. So I am fine with this change.

PR #772, opened 2017-02-22
Consolidate logics related to ModRM/SIB byte

0dvictor:
1. Provide unified APIs to construct ModRM and SIB bytes.
2. Properly set the bits without assuming the memory being set to zero.

### DIFF
--- a/compiler/x/amd64/codegen/OMRMemoryReference.cpp
+++ b/compiler/x/amd64/codegen/OMRMemoryReference.cpp
@@ -622,8 +622,8 @@ OMR::X86::AMD64::MemoryReference::generateBinaryEncoding(
          {
          // Addresses in the low 2GB (or high 2GB) can be accessed using a SIB byte
          //
-         self()->setModField(modRM, IA32SIBPresent);
-         *cursor++ = 0x25; // [disp32] encoding
+         self()->ModRM(modRM)->setBase()->setHasSIB();
+         self()->SIB(cursor++)->setScale()->setNoIndex()->setIndexDisp32();
          *(uint32_t*)cursor = (uint32_t)displacement;
          }
       else
@@ -631,7 +631,7 @@ OMR::X86::AMD64::MemoryReference::generateBinaryEncoding(
          TR_ASSERT(IS_32BIT_RIP(displacement, rip), "assertion failure");
          TR_ASSERT(!(comp->getOption(TR_EnableHCR) && sr.getSymbol() && sr.getSymbol()->isClassObject()),
             "HCR runtime assumptions currently can't patch RIP-relative offsets");
-         self()->setModField(modRM, IA32Offset32);
+         self()->ModRM(modRM)->setIndexOnlyDisp32();
          *(uint32_t*)cursor = (uint32_t)(displacement - (intptrj_t)rip);
          }
 
--- a/compiler/x/codegen/OMRInstruction.hpp
+++ b/compiler/x/codegen/OMRInstruction.hpp
@@ -69,6 +69,95 @@ class EnlargementResult
 
 class OMR_EXTENSIBLE Instruction : public OMR::Instruction
    {
+   public:
+
+   struct REX
+      {
+      uint8_t B : 1;
+      uint8_t X : 1;
+      uint8_t R : 1;
+      uint8_t W : 1;
+      uint8_t _padding : 4;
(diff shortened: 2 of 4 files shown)


andrewcraik:
@cathyzhyi another one for an opinion from you please. I will review as well.

cathyzhyi:
LGTM

andrewcraik:
Agreed - another nice piece of clean up to make the encoding more robust. LGTM. @0xdaryl over to you since I think this is in your domain of expertise.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1217,6 +1217,102 @@ typedef enum
    IA32EFlags_CF = 0x10
    } TR_EFlags;
 
+namespace TR
+   {
+   namespace X86
+      {
+      namespace Instruction
```
mgaudet: This would be the first use of namespaces under `TR::` right? 

(Note: Just because it's different doesn't mean we shouldn't merge it, but it's certainly a bit curious compared to other uses of namespaces we have. )

0xdaryl:
The change itself is a nice cleanup.

Given the pervasiveness of this kind of change to binary encoding I'd like to understand how you verified that each of the encoding scenarios that were changed in MemoryReference have been encoded correctly for both 32-bit and 64-bit.

0dvictor:
For all places that changed, I had assertion code to verify the new way gave the same result as the old way did.
For example,
old code:
   self()->setModField(modRM, IA32SIBPresent);
new code:
   self()->ModRM(modRM)->setBase()->setHasSIB();
verification code:
   {
   self()->setModField(modRM, IA32SIBPresent);
   auto old = *modRM;
   *modRM = 0xff;
   self()->ModRM(modRM)->setBase()->setHasSIB();
   if (*modRM != old)
      {
      TR_ASSERT(false, "incorrect coding");
      *(char*)0 = 0xdd;
      }
   }
Running through all required continuous tests, no assertions or segfault was triggered; hence I believe the new code generates same code as the old one.

0xdaryl:
Thanks for explaining that.  But the assert approach is only effective IF a particular encoding combination is attempted from the set of tests you choose to run.  My question is more about how you verified that each of the encodings you changed were exercised (i.e., did you get 100% coverage?)  Using your assert approach you'd have to add counters or printfs to verify that.

I'm not trying to be overly difficult, but subtle binary encoding errors can lead to problems that are very difficult to diagnose.  Which is why I'd like to have some assurance that each of these paths that follow your new scheme has been exercised before merging.

0dvictor:
Agreed. During development, I had printfs for each place I touched and all routes were used from my local tests; although I did not run continuous builds with printfs.

compiler/x/codegen/X86Ops.hpp
0xdaryl: In my mind `X86::`  is an acceptable subclassification under `TR::`, but I feel chopping it up even finer under `Instruction::` is providing little additional isolation for this particular feature.  I prefer the `Instruction::` namespace removed.

0dvictor: Sure, I have removed Instruction:: namespace.

0dvictor:
Instruction:: namespace is removed. Do you mind take a look at the change once more?

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1217,6 +1217,99 @@ typedef enum
    IA32EFlags_CF = 0x10
    } TR_EFlags;
 
+namespace TR
```
0xdaryl: My apologies for the delay in responding, but I've been giving this some deeper thought on how to proceed and I have a different suggestion.  There's been some discussion around the use of the `TR` namespace (issue #788), and I have to admit I'm more convinced of its use as a pure namespace without further refinement.

Looking at what you've implemented, why don't you implement these as public static members of the `OMR::X86::Instruction` class?  Then, you would simply reference them as `TR::Instruction::ModRM` (for example).  The way you reference things in the `TR` namespace now makes the developer think they are static members, but they're not.  Making them static members is cleaner and conceptually pleasing.

0dvictor: Good suggestion, they are members of OMR::X86::Instruction now.

0xdaryl:
Looks good now.  I'll merge when CI completes.

PR #774, opened 2017-02-22
Prevent inlining for body of Modron Assertions

dmitripivkine:
Modron Assertions with message in OMR_GC_DEBUG_ASSERTS mode allocate
print buffer in the stack. It my be a reason for significant size of
stack frame in case of aggressive inlining. An execution of the body of
assertion however is a part of slow path so inlining here is not
necessary. This implementation allows to reduce size of stack frames.


gc/base/ModronAssertions.cpp
```diff
@@ -0,0 +1,58 @@
...
+#if defined(OMR_GC_DEBUG_ASSERTS)
+
+#define ASSERTION_MESSAGE_BUFFER_SIZE	((uintptr_t)256)
+
+/*
+ * The purpose of moving of the body of macros GC Assertions with message to separate function is to prevent inlining
+ * on fast path to keep stack frames small
+ */
+
+intptr_t
```
charliegracie: Does this need to have a return value as it is never used?

charliegracie:
@dmitripivkine The change looks good.  Can you squash the 2 commits together and push the resulting new commit to the same branch.  You will have to force push the new commit.

dmitripivkine:
Running personal builds on Linux and Windows to be sure there is no compilation problems. If succeed, will squash commits to one.

PR #775, opened 2017-02-23
Change longSubtractAnalyserImpl arguments type

cathyzhyi:
The argument's value is changed in the function so change to
reference type to make sure the change is visible to the caller.


andrewcraik:
LGTM

PR #777, opened 2017-02-23
Add architecture guard for RXSBG

fjeremic:
RXSBG is not supported on z9. The generation of this instruction must be
guarded with an architecture check.

Issue: #667


vijaysun-omr:
Looks like a straightforward change to guard against generating code that is'nt supported on older hardware. I will merge once checks pass.

PR #778, opened 2017-02-23
Use 32-bit shr in x86 arrayset evaluator on 32-bit

jdmpapin:
Passing `cg` to `SHRRegImm1` inadvertently selects the overload
`SHRRegImm1(bool is64Bit)`, and `cg` implicitly converts to `true`
because it is non-null. As a result, when `arraySetDefault` generates a
`shr`, it unconditionally generates the 64-bit variant. The 64-bit `shr`
is encoded using a `REX.W` prefix, which in 32-bit code is interpreted
as an additional instruction `dec eax` preceding the desired 32-bit
`shr` instruction. Since `rep stos` gets the value to store from (the
low part of) `eax`, the stored values are off by 1 in this case.

Calling the nullary overload of `SHRRegImm1` determines bitness instead
using `TR::Compiler->target.is64Bit()`.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -3731,7 +3731,7 @@ static void arraySetDefault(TR::Node* node, uint8_t elementSize, TR::Register* a
          break;
       }
 
-   if (shiftAmount) generateRegImmInstruction(SHRRegImm1(cg), node, sizeReg, shiftAmount, cg);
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
I would like @andrewcraik to review this.

andrewcraik:
LGTM

vijaysun-omr:
That bug sounds nasty; good work @jdmpapin finding and fixing. I see how the fix would use the right sized shr for 32-bit. Merging.

PR #779, opened 2017-02-23
Substitute usage of stck with stckf

dnakamura:
stckf is significantly faster than stck (1 cycle vs hundreds), but looses
the guarantee of uniqueness between calls (which is not needed here).

### DIFF
--- a/port/linuxs390/omrrttime.s
+++ b/port/linuxs390/omrrttime.s
@@ -26,7 +26,7 @@ maxprec:
     basr %r1,0
 .LT0:
     
-    stck 48(%r15)
+    stckf 48(%r15)
     lm  %r2,%r3,48(%r15)
     sl  %r3,.LC0-.LT0+4(%r1)
     brc 3,.LCarry
--- a/port/linuxs39064/omrrttime.s
+++ b/port/linuxs39064/omrrttime.s
@@ -25,7 +25,7 @@ maxprec:
 .text
     larl    %r3,.LT0
     
-    stck 64(%r15)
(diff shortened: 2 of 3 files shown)


dnakamura:
Updated as stckf is not available on 31bit

fjeremic:
> Updated as stckf is not available on 31bit

Why is that? Both STCK and STCKF store 8-bytes worth of data to a memory location. I do not understand why addressability introduced a restriction here. Could you elaborate?

charliegracie:
@dnakamura Any update on the question by @fjeremic?

dnakamura:
@fjeremic Thinking about it you could still do stckf on 31 bit. I was thinking 31bit == ESA/390.  It boils down to how backward comparable you want to be.

fjeremic:
@dnakamura thanks for the explanation. IMO the benefit of consistency far outweighs the probability of someone needing to run this code on ESA/390 which is decades old now. Given the performance implications of using STCK on modern z/Architecture I am in favor of using STCKF on 31-bit as well.

If a future requirement arises for a consumer of OMR to run on ESA/390 we can loop back and add this support. Until then lets take the benefit of the additional performance STCKF gives us.

charliegracie:
@dnakamura and @fjeremic I think we should go with the STCKF all of the time.

charliegracie:
@dnakamura I will merge the pull request once you update the commit to use stckf for both 32 and 64 bit

dnakamura:
@charliegracie  should be good to go

PR #780, opened 2017-02-24
Enable ArraySet for X86

0dvictor:
ArraySet evaluator has been optimized and tested; turing it on.
In addition, SupportsArraySetToZero flag is removed since all CodeGen now either supports arrayset to an arbitrary value or not support arrayset at all. SupportsArraySetToZero flag served only as an intermediate workaround when X86 CodeGen did not properly support arrayset.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -365,12 +365,12 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    self()->setSupportsArrayCmp();
    self()->setSupportsArrayCopy();
 
-   if (comp->getOption(TR_EnableX86AdvancedMemorySet))
-      self()->setSupportsArraySet();
-
    static char *disableArraySet = feGetEnv("TR_disableArraySet");
    if (!disableArraySet)
+      {
       self()->setSupportsArraySetToZero();
+      self()->setSupportsArraySet();
+      }
 
    self()->setSupportsScaledIndexAddressing();
    self()->setSupportsConstantOffsetInAddressing();
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -733,7 +733,6 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableVirtualPersistentMemory",      "M\tenable persistent memory to be allocated using virtual memory allocators",
                                           SET_OPTION_BIT(TR_EnableVirtualPersistentMemory), "F", NOT_IN_SUBSET},
    {"enableVpicForResolvedVirtualCalls",  "O\tenable PIC for resolved virtual calls",         SET_OPTION_BIT(TR_EnableVPICForResolvedVirtualCalls), "F"},
-   {"enableX86AdvancedMemorySet",         "C\tEnable advanced memory support on x86", SET_OPTION_BIT(TR_EnableX86AdvancedMemorySet), "F", NOT_IN_SUBSET },
    {"enableYieldVMAccess",                "O\tenable yielding of VM access when GC is waiting", SET_OPTION_BIT(TR_EnableYieldVMAccess), "F"},
    {"enableZAccessRegs",                "O\tenable use of access regs as spill area on 390.", SET_OPTION_BIT(TR_Enable390AccessRegs), "F"},
    {"enableZEpilogue",                  "O\tenable 64-bit 390 load-multiple breakdown.", SET_OPTION_BIT(TR_Enable39064Epilogue), "F"},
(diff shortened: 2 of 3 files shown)


compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -365,12 +365,12 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    self()->setSupportsArrayCmp();
    self()->setSupportsArrayCopy();
 
-   if (comp->getOption(TR_EnableX86AdvancedMemorySet))
-      self()->setSupportsArraySet();
-
    static char *disableArraySet = feGetEnv("TR_disableArraySet");
```
vijaysun-omr: Any particular reason this is an env var and not an option check ? Options can be applied per compiled method whereas env vars cannot.

0dvictor: It is historical code. I'm not sure why it was designed that way but I guess the env var was purely for debugging when it was first developed and meant to be deleted later. Z has neither env var or option; P has SupportsArraySet off by default. Some optimizer has its own way to turn arrayset related optimization on/off, for example TR_DisableSeparateInitFromAlloc.

vijaysun-omr: Okay, using an option can be a valuable debugging aid where you can narrow down to a particular method without needing to re-build. @andrewcraik what is your opinion ?

andrewcraik: This matches the p codegen so I think it is ok for now. It will at least allow us to determine if the evaluator is involved and then more debugging will be needed anyway and Victor would be the one doing that so since he made the PR and seemed happy I think we go with it.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -1475,9 +1475,6 @@ class OMR_EXTENSIBLE CodeGenerator
    bool getSupportsArraySet() {return _flags1.testAny(SupportsArraySet);}
    void setSupportsArraySet() {_flags1.set(SupportsArraySet);}
 
-   bool getSupportsArraySetToZero() {return _flags3.testAny(SupportsArraySetToZero);}
```
andrewcraik: I'm not sure we really want to get rid of this - it is a useful feature being able to allow only arrayset to zero or general arrayset. @vijaysun-omr thoughts?

vijaysun-omr: I agree that it's useful to distinguish; some JIT optimizations only care about "zero" arrayset support and it would be easier to get those optimizations working on a new platform if we did not have to have to implement the "full" arrayset support to begin with. I assume this was your reasoning as well @andrewcraik ?
 
@0dvictor Do you know what the status is on the ARM code generator re: arrayset support ?

0dvictor: ARM does not support arrayset; neither set to zero or to arbitrary value.

compiler/x/codegen/OMRCodeGenerator.cpp
0dvictor: I personally prefer keep the code as it using an env var for being consistent with other platforms as well as historical code.
I agree that an option is more flexible and useful; but I think it might be better to introduce the option in a separate changeset and update all platforms to check the flag.

compiler/codegen/OMRCodeGenerator.hpp
andrewcraik: My reasoning was exactly that - a codegen may only want to support setting to zero rather than arbitrary setting and having this degree of freedom seems useful in the abstract and since it is already implemented I don't think we need to throw it away.

0dvictor: Sure, we can keep it.

0dvictor: Since supporting arrayset to zero is technically a 'sub-feature' of arrayset, we need to ensure that if a CG declares supporting for arbitrary arrayset, it must also declare supporting for arrayset to zero; otherwise, it may lead to confusion.

vijaysun-omr: Makes sense. Were you planning to make that change as part of this PR ?

0dvictor: I prefer not to; as this PR is only to enable arrayset on X86.
With this PR, X declares supports for both arrayset to zero and arbitrary arrayset. Z already declares supports for both. ARM supports neither. The only platform violates this proposed rule is P, which, under TR_AggressiveOpts, only declares support for arbitrary arrayset. I'll contact folks in charge of P and see if they would add declaration for supports arrayset to zero.

andrewcraik:
LGTM with those changes done - can we get a committer to recheck the change now?

andrewcraik:
I think the change for the option will be handled in a separate PR since it will be a change to P codegen to set the supportsArraySetToZero flag. @0dvictor was just telling me that he had discussed it with @ymanton and the pull request will be made shortly.

vijaysun-omr:
Thanks for the changes and the clarifications wrt the two queries on the different platforms. Merging this one now.

PR #784, opened 2017-02-26
Make options initialization functions protected

0xdaryl:
The following functions in OMROptions should be `protected`
rather than `private` so they can be extended and specialized
by other projects:

* jitPreProcess()
* fePreProcess(void *base)
* jitPostProcess()
* fePostProcessAOT(void *base)
* fePostProcessJIT(void *base)
* jitLatePostProcess(TR::OptionSet *optionSet, void *jitConfig)
* feLatePostProcess(void *base, TR::OptionSet *optionSet)

### DIFF
--- a/compiler/control/OMROptions.hpp
+++ b/compiler/control/OMROptions.hpp
@@ -1981,6 +1981,15 @@ public:
    static bool isQuickstartDetected() { return _quickstartDetected; }
    void disableCHOpts(); // disable CHOpts, but also IPA and prex which depend on the chtable
 
+protected:
+   void  jitPreProcess();
+   bool  fePreProcess(void *base);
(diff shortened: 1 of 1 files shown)


fjeremic:
Should they be named with a _compiler_ prefix? These are these really _JIT_ compiler specific?

0xdaryl:
@fjeremic , you're right.  But this PR is really about simply changing the access specifier for those existing functions.  This is one tiny step toward getting things in a more understandable state before tackling a rework of the Options processing (an issue is forthcoming).  That rework will ensure function names are aligned with the more generic naming of compiler technology in OMR.

vijaysun-omr:
Change allows extension of the mentioned APIs in the future; no change in current functionality. Merging.

Issue #508, opened 2016-11-17
Replace MM_ConfigurationLanguageInterface* with MM_ConfigurationDelegate


ktbriggs:
This is a first step toward eliminating the MM_*LanguageInterface classes that were used to catch language-dependent factors while extracting core OMR out of J9 Java. These factors represent necessary integration points but the resulting MM_*LanguageInterface classes are fragmented and difficult to maintain. Moreover they force virtual implementations of member functions which can be a performance issue for functions that are frequently called in the runtime.

Each MM_*LanguageInterface class will be replaced with a corresponding MM_*Delegate class that will implement equivalent factors as the language interface, without forcing virtual methods. In the process the delegate methods will be rationalized and simplified wherever possible so that the delegate class interface will be easier to implement, use, and document.

For this issue, MM_ConfigurationLanguageInterface will be eliminated. A MM_ConfigurationDelegate class will replace derived MM_ConfigurationLanguageInterface* subclasses in the example/glue and in glue for other existing language implementations.

ktbriggs:
Closed via https://github.com/eclipse/omr/pull/785

ktbriggs:
Reopening this -- some build flag issues need to be cleaned up.

charliegracie:
Closing as #884 has been merged

PR #785, opened 2017-02-27
Deprecate MM_ConfigurationLanguageInterface etc

ktbriggs:
Refactored MM_Configuration and subclasses to clean up convoluted
inheritance factors and eliminate all MM_ConfigurationLangageInterface
usage, replacing with equivalent ConfigurationDelegate method calls.

Client language will be required to implement all methods defined
for MM_ConfigurationDelegate in glue/ConfigurationDelegate. This 
is a more focused articulation of the integration points required for
client languages to configure GC.

Deleted obsolete ConfigurationLanguageInterface*.*.

Updated copyright on all files touched in this commit.

Fixed newlines in ConfigurationGenerational.[ch]pp.

Issue: #508


charliegracie:
@ktbriggs quick note before I have reviewed.  Can you limit the first line of the commit message to 50 characters?  Thanks

ktbriggs:
Shortened first line of commit message. Also added comments for methods in example/glue/ConfigurationDelegate.hpp and added MM_EnvironmentBase* parameter to two MM_ConfigurationDelegate methods that may require environment to access extensions, vm, port, etc.

ktbriggs:
Not sure where error in Travis build is coming from. Travis reports build failure for 32-bit linux only, all other platforms pass. Travis log output for 32-bit linux build is empty. I am going to touch commit message only and push again to force new Travis build.

gc/base/Configuration.cpp
```diff
@@ -162,41 +143,51 @@ MM_EnvironmentBase*
...
 	MM_EnvironmentBase* env = allocateNewEnvironment(extensions, omrVMThread);
-	if (NULL == env) {
-		return NULL;
-	}
-	if (!initializeEnvironment(env)) {
-		env->kill();
-		return NULL;
+	if (NULL != env) {
+		if (initializeEnvironment(env)) {
+			if (extensions->fvtest_disableInlineAllocation) {
```
charliegracie: It is not clear where this change came from as it did not exist in this function before. It feels like it belongs in the initializeEnvironment method.

example/glue/ConfigurationDelegate.hpp
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * This method is called just after the heap has been initialized. Delegate may perform
+	 * any additional heap-related initialization at this point.
+	 * @param env The calling thread environment
+	 * @return false if heap initialization fails
+	 */
+	bool
+	initializeHeap(MM_EnvironmentBase* env)
```
charliegracie: The name of this function sounds to me like it is initializing the heap.  Maybe we can change it to heapInitialized or something to make it clear that it is called after the heap is initialized to allow the language to do its own initialization of structures, etc. for the heap.

example/glue/ConfigurationDelegate.hpp
```diff
@@ -0,0 +1,146 @@
...
+		return 0;
+	}
+
+	/**
+	 * The OMR segregated GC requires GC clients to specify a distribution of segment sizes with which
+	 * it will partition the heap. This method is called to retrieve the size distribution metadata.
+	 * @param env The calling thread environment
+	 * @return A pointer to segregated heap size distribution metadata, or NULL if not using segregated heap
+	 */
+	OMR_SizeClasses *getSegregatedSizeClasses(MM_EnvironmentBase *env)
```
charliegracie: Should this function be wrapped in an #if defined since it only needs to be there if you are using a segregated heap?

example/glue/ConfigurationDelegate.hpp
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * Perform GC client-specific initialization for a new MM_EnvironmentBase subclass instance.
+	 * @param env The thread environment to initialize
+	 * @return false if environment initialization fails
+	 */
+	bool
+	initializeEnvironment(MM_EnvironmentBase* env)
+	{
```
charliegracie: Similar to the initializeHeap function this is not actually initializing an environment but rather allowing the language to do its own initialization when an environment is initialized.  Also this code seems like it would be common to all languages and does not belong in the delegate.

example/glue/ConfigurationDelegate.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+	/**
+	 * The OMR GC preallocates a pool of collector threads. This method is called to allow GC clients
+	 * to determine the maximum number of concurrent collector threads to run during GC cycles.
+	 * @param env The calling thread environment
+	 * @return The maximum number of GC threads to allocate for the runtime GC
+	 */
+	uintptr_t getMaxGCThreadCount(MM_EnvironmentBase* env)
+	{
+		return 1;
```
charliegracie: Is this for concurrent threads or for slave threads?

gc/base/standard/ConfigurationGenerational.hpp
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 1991, 2016
+ * (c) Copyright IBM Corp. 1991, 2016, 2017
```
charliegracie: This is not the correct way to update the copyright.  It is initial year, current year.  Can you fix this in all of the files please?

ktbriggs: Ok.

example/glue/ConfigurationDelegate.hpp
ktbriggs: Concurrent threads.
```
void
MM_Configuration::initializeGCThreadCount(MM_EnvironmentBase* env)
{
	...
	extensions->gcThreadCount = omrsysinfo_get_number_CPUs_by_type(OMRPORT_CPU_TARGET);
	if (_delegate.getMaxGCThreadCount(env) < extensions->gcThreadCount) {
		extensions->gcThreadCount = _delegate.getMaxGCThreadCount(env);
	}
	...
```
MM_GCExtensionsBase::gcThreadCount is used to limit threads in MM_ParallelDispatcher:
```
MM_ParallelDispatcher::initialize(MM_EnvironmentBase *env)
{
	...
	_threadCountMaximum = env->getExtensions()->gcThreadCount;
```

example/glue/ConfigurationDelegate.hpp
ktbriggs: Renaming this function to environmentInitialized(), consistent with heapInitialized(). This (example glue) code is necessary for now but will obsolete when MM_EnvironmentLanguageInterface is refactored into MM_EnvironmentDelegate (WIP). Example code has nothing to add to environment initialization but other languages should have this hook.

example/glue/ConfigurationDelegate.hpp
ktbriggs: It is unconditionally called from ConfigurationRealtime. All of the realtime code in Java is compiled unconditionally.

example/glue/ConfigurationDelegate.hpp
ktbriggs: I agree. This and initializeEnvironment() are both informational (hooks) called from OMR to permit language to respond to GC-related events. Using your suggested name pattern for both of these and will carry this convention forward into ELI and GCLI component delgates.

ktbriggs:
@charliegracie requested changes are in most recent commit (103396e). All 3 commits have been rebased on eclipse/master@5939598.

example/glue/ConfigurationDelegate.hpp
charliegracie: That is actually the Slave threads and not the concurrent threads.  Can you update the documentation to reflect this?  Instead of using slave threads you can talk about this number being used to limit the number of threads created by the paralleldispatcher or something like that.

example/glue/ConfigurationDelegate.hpp
charliegracie: Since I have requested another change below can you wrap this code in #if defined()?  As we add new APIs I would like them to all be properly #if defined so it is obvious when you need to worry about them.

ktbriggs: Done. I removed MM_Configuration::initializeSizeClasses(), which previously called this function, and let MM_ConfigurationSegregated and MM_ConfigurationRealtime (Java) call it directly as they are the only consumers of hte segregated heap. The delegate method getSegregatedSizeClasses() is conditionally compiled only when OMR_GC_SEGREGATED_HEAP is enabled.

example/glue/ConfigurationDelegate.hpp
ktbriggs: Done.

ktbriggs:
@charliegracie changes reflecting updated from your most recent comments are in commit 4c25835. Will squash these commits when review is complete.

PR #786, opened 2017-02-27
Use resolved method for skipping HCR decisions

fjeremic:
Callee method symbols may be null depending on whether we are peeking
inside of a top-level callee method to be inlined or not. For a method
two call stacks below the current compiled body the callee method symbol
will be null as it has not been constructed yet. This means we are
unable to properly make a decision on whether or not to generate HCR
guards for such methods.

Alternatively a resolved method object should always be available for an
inline callee candidate. Thus we can always use the resolved method
object for determining whether or not to skip generating HCR guards.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1782,8 +1782,7 @@ TR_InlinerBase::addGuardForVirtual(
 
    static const char *disableHCRGuards = feGetEnv("TR_DisableHCRGuards");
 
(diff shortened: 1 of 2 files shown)


fjeremic:
@cathyzhyi could you please review this change? Thank you!

andrewcraik:
LGTM

vijaysun-omr:
The changes look better to me than what was there before in the sense that the resolved method will be available/non-null in more cases than the resolved method symbol. Merging.

PR #790, opened 2017-02-28
Use a typedef for the standard CFGEdge* container

lmaisons:
Use of the container type for holding CFGEdge pointers has been
splattered throughout the codebase.  This impedes the ability to modify
the container type used, which is required to resolve performance
issues.

Rather than identifying the container type explicitly, replace the
declarations with a shared typedef.

Instances of this container were found using the regex:
```TR::list<\s*(TR::)?CFGEdge\s*[*]\s*>```

### DIFF
--- a/compiler/env/ExceptionTable.cpp
+++ b/compiler/env/ExceptionTable.cpp
@@ -65,7 +65,7 @@ TR_ExceptionTableEntryIterator::TR_ExceptionTableEntryIterator(TR::Compilation *
 
          // create exception ranges from the list of exception predecessors
          //
-         TR::list<TR::CFGEdge*> & exceptionPredecessors = catchBlock->getExceptionPredecessors();
+         TR::CFGEdgeList & exceptionPredecessors = catchBlock->getExceptionPredecessors();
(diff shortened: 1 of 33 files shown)


lmaisons:
@0xdaryl @mstoodle This one is pretty boring but will help immensely when swapping out std::list with something else.  Should I be soliciting wider input for this one or is it OK for merging as-is?

0xdaryl:
Changes look reasonable to me.  @vijaysun-omr @andrewcraik : any objections?

andrewcraik:
I think this is a reasonable change since it provides more freedom of action to update list data types in API compatible ways going forward.

FYI @jdmpapin @cathyzhyi @liqunl @ncough @0dvictor - probably doesn't affect you but to warn you of the new typedef that is proposed and that new code should use this typedef and you may need to update inflight changes based to be compatible with this change if you are relying on CFGEdge lists directly.

andrewcraik:
FYI @fjeremic @NigelYiboYu you may also want to know about this since I know you have also been looking at the optimizer a bit of late which is the primary area of impact of this change.

PR #791, opened 2017-02-28
Add efficient single bit dataflow analysis

andrewcraik:
The current dataflow engine is instantiated using BitVectors as backing storage to support analysis requiring an arbitrary number of bits. While flexible, the use of BitVectors is especially inefficient for single-bit data flow analyses. This change introduces a new TR_SingleBitContainer which implements the BitVector API used by the dataflow engine, but which stores only a single boolean value. The dataflow framework is
instantiated for this new container to allow for much cheaper single-bit dataflow analyses to be written.

### DIFF
--- a/compiler/infra/BitVector.cpp
+++ b/compiler/infra/BitVector.cpp
@@ -211,6 +211,15 @@ void TR_BitVector::print(TR::Compilation *comp, TR::FILE *file)
       }
    }
 
+void TR_SingleBitContainer::print(TR::Compilation *comp, TR::FILE *file)
+   {
+   if (comp->getDebug())
+      {
+      if (file == NULL)
+         file = comp->getOutFile();
+      comp->getDebug()->print(file, this);
+      }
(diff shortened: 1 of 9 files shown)


andrewcraik:
@ncough can you review this please and then let Vijay know it is ready for his review as well?

ncough:
@vijaysun-omr LGTM

vijaysun-omr:
Looks good and should be a fine addition to the dataflow framework.

0xdaryl:
The `TR_SingleBitContainer` class is a new addition to OMR infrastructure and appears to be self-contained.  Prior to merging, I think this PR should have provided a number of unit test cases in fvtest to exercise each of its fundamental operations.  Even post-merge I'd still like to see them.

Also, the comment preceding the implementation in BitVector.hpp (lines 94-98) contains several spelling mistakes.  Please correct them.

Code looks good otherwise.

PR #792, opened 2017-02-28
Cleanup deadcode in OSRExceptionEdgeRemoval

andrewcraik:
As part of optimization prototyping an alternative implementation of this optimization was added in a method called newperfom. The newperform method has been dormant for some time and is no longer needed. This commit removes this dead code.


vijaysun-omr:
Simple deletion of unused code. I will merge once checks pass.

PR #795, opened 2017-02-28
add some IlBuilder documentation and a new IlBuilder service StoreOver

mstoodle:


compiler/ilgen/IlBuilder.cpp
```diff
@@ -592,6 +592,12 @@ IlBuilder::indirectLoadNode(TR::IlType *dt, TR::Node *addr, bool isVectorLoad)
    return loadValue;
    }
 
+/**
+ * @brief Store an IlValue into a named local variable
+ * @param varName the name of the local variable to be stored into. if the name has not been used before, this local variable will
+ *                take the same data type as the value being written to it.
+ * @param value IlValue that should be written to the local variable, which should the same data type
```
Leonardo2718: "should the same data type" -> "should the __be__ same data type"

compiler/ilgen/IlBuilder.cpp
```diff
@@ -592,6 +592,12 @@ IlBuilder::indirectLoadNode(TR::IlType *dt, TR::Node *addr, bool isVectorLoad)
    return loadValue;
    }
 
+/**
+ * @brief Store an IlValue into a named local variable
+ * @param varName the name of the local variable to be stored into. if the name has not been used before, this local variable will
```
Leonardo2718: "if the name has" -> "If the name has"

compiler/ilgen/IlBuilder.cpp
```diff
@@ -605,10 +611,17 @@ IlBuilder::Store(const char *varName, TR::IlValue *value)
    storeNode(sym, loadValue(value));
    }
 
+/**
+ * @brief Store a vector IlValue into a named local variable
+ * @param varName the name of the local variable to be vector stored into. if the name has not been used before, this local variable will
```
Leonardo2718: "if the name has" -> "If the name has"

fjeremic:
It may be a good idea to write the Doxygen documentation in the header files where the interface for the function is defined so that consumers of the library who are including the header files have the documentation that comes with them. The details of the implementation should go into the object files.

This may not make as much sense here since OMR / IlBuilder is not a standalone library but for binary / statically linked distributions of a library all you would have are header files and the documentation should go along with them.

This may be up to debate though and we should collectively as a group decide this for consistency reasons.

mstoodle:
So for this PR, I just put them in the same place everything else is (the C++ file). I understand the concern, though.

I considered putting the comments in the header, but (my opinion was that) those comments make the header a lot harder to read as code. That is very possibly just my own bias, but I prefer being able to quickly scan the list of services in the header, and if I need help understanding what a service needs, I can go look in the C++ file for the details (including the implementation :) ).

For me, I found a header file with 6-ish lines of comment for every line of actual code annoying to read.

compiler/ilgen/IlBuilder.cpp
mstoodle: is that a typo in your fix for my typo? :)

mstoodle: should fixed, thanks

compiler/ilgen/IlBuilder.cpp
mstoodle: should be fixed, thanks

fjeremic:
Thank for opening the discussion over at #800. I have nothing to add to the review comments, however could you please amend the commit headers so as to conform to the [commit guidelines](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines)?

mstoodle:
Sorry, brain not working well @fjeremic . Can you be more specific on what in the commit headers needs conforming?

mstoodle:
No worries, @fjeremic :) . I just couldn't see the lower case 'a's for some reason.

vijaysun-omr:
LGTM. I will merge once the checks pass.

PR #796, opened 2017-03-01
Introduce JIT option to disable arrayset optimizations

0dvictor:
Added new JIT option, disableArraySetOpts By applying this options,
CodeGen would not declare support for arrayset.


vijaysun-omr:
Thanks for taking the suggestion to convert the env var to an option Victor. I have a feeling it will come in handy when you are debugging a failure and need to restrict the transformation to a subset of methods to narrow it down.

PR #797, opened 2017-03-01
Restore id field on CFGEdge

andrewcraik:
It can be very handy, during optimization, to be able to identify CFG
edges by an ID number. The OMR code has all the required APIs, but the
backing field is not present. This change restores the backing field so
that an optimization can store an ID number of a CFGEdge to simplify
booking keeping and analysis data structures. There is no requirement
for IDs to be unique or to have any meaning outside of a single
optimization - it is simply a convenience field.

### DIFF
--- a/compiler/infra/OMRCfg.cpp
+++ b/compiler/infra/OMRCfg.cpp
@@ -426,7 +426,7 @@ TR_OrderedExceptionHandlerIterator::getCurrent()
 
 
 TR::CFGEdge::CFGEdge(TR::CFGNode *pF, TR::CFGNode *pT, TR_AllocationKind allocKind)
-   : _pFrom(pF), _pTo(pT), _visitCount(0), _frequency(0)
+   : _pFrom(pF), _pTo(pT), _visitCount(0), _frequency(0), _id(-1)
    {}
 
 TR::CFGEdge * TR::CFGEdge::createEdge (TR::CFGNode *pF, TR::CFGNode *pT, TR_Memory* trMemory, TR_AllocationKind allocKind)
--- a/compiler/infra/TRCfgEdge.hpp
+++ b/compiler/infra/TRCfgEdge.hpp
@@ -38,7 +38,7 @@ class CFGEdge : public TR_Link<CFGEdge>
    TR_ALLOC(TR_Memory::CFGEdge)
 
 
-   CFGEdge() : _pFrom(NULL), _pTo(NULL), _visitCount(0), _frequency(0)
+   CFGEdge() : _pFrom(NULL), _pTo(NULL), _visitCount(0), _frequency(0), _id(-1)
    {}
 
    // Construct a normal edge between two nodes
@@ -71,8 +71,8 @@ class CFGEdge : public TR_Link<CFGEdge>
       {
       return ++_visitCount;
       }
-   int32_t    getId()            {return -1;}
-   int32_t    setId(int32_t id)  {return -1;}
+   int32_t    getId()            {return _id;}
+   int32_t    setId(int32_t id)  {return (_id = id);}
 
    int16_t    getFrequency()          { return _frequency; }
 
@@ -106,6 +106,7 @@ class CFGEdge : public TR_Link<CFGEdge>
 
    int16_t _frequency;
 
+   int32_t _id;
    };
 
 }


vijaysun-omr:
Just curious, if considered making the id a unique value that is assigned when the CFG edge is created and persists across optimizations ? Kind of like a CFG node number behaves. I am not sure if that would have worked just as well in your specific use case.

mstoodle:
Just curious: has there been any evaluation of the dynamic footprint impact of this change, since CFG edges can be pretty plentiful.

andrewcraik:
I have not done a dynamic footprint measurement. This implementation matches an older implementation that existed in older versions of this code prior to contribution. A globally assigned unique ID would work for what I have in mind, but with a great dynamic footprint when processing the edges if they are not compacted since there would be gaps in my otherwise packed data structures. I don't really want to pay the cost of always numbering the edges since no one else needs the functionality right now. We could look to make it like a node number later if necessary.

If there is a real concern on footprint I can create my own mapping data structure during optimization/analysis, but that will incur an extra compile-time cost since the data structure would need to be based on hashing of edge addresses or similar.

@mpirvu how worried are you by the addition of this field?

mpirvu:
Since we started allocating JIT scratch memory from mmaped segments the JIT is less sensitive to footprint increases. However, there is the hidden danger that some expensive compilations (hot/scorching) may start failing due to exceeding the 256MB limit. This is going to be a silent performance issue.
I see that the new field follows a 16-bit field (_frequency). If a 16-bit ID is sufficient, then the new field can be had for free.

andrewcraik:
For my purpsoes yes - if we had 2^15-1 edges then I think we would have blown past other limits a while ago - I can shink the id to int16_t - I will just need to touch more code. @mstoodle would shrinking the field size and getting it for free satisfy your concern?

mpirvu:
Now I remember that I had plans to extend that frequency field from 16-bit to 32-bit to avoid the truncation we do when we set frequencies
```
   void       setFrequency(int32_t f)
      {
      if (f >= SHRT_MAX)
         f = SHRT_MAX-1;

      _frequency = f;
      }
```
This code distorts the real frequency. If we couple this with the truncation done by the IProfiler (which also keeps frequency on 16-bit fields) then we can artifically improve throughput just by running the interpreter for longer time: most blocks will appear to be at frequency 10000 and the JIT loves to inline in such blocks.

mstoodle:
If you can hide it in otherwise unused 16 bits then there isn't any growth so nothing to worry about. But with @mpirvu 's last point, it now it sounds like there needs to be a decision at some point on diagnostic value (which I understood to be the value of this new id field) versus dynamic footprint.

In any case, I was just trying to raise the performance question at the point of change rather than later as something is finally found to be worse off. Do we have any notion of how many control flow edges are created during one of those massive Java compiles @mpirvu mentioned. That would help to understand the impact for a real use scenario for the OMR project.

Independently of this PR, maybe it would be useful for the J9 project to help characterize which data structures are the most important ones to pay attention to (because they have large multipliers) when new fields show up in pull requests. Just a thought, I'm open to other ideas too. We can always just continue to react to issues as they are raised by J9 when problems are discovered.

andrewcraik:
The field is not a diagnostic field, it is a field used by the optimization to save having to use a hash data structure based on the CFGEdge address/pointer which is what my initial implementation used and which proved to be very expensive in terms of compile time.

I'm not quite sure what the conclusion of the discussion is...

mstoodle:
Ah, ok. I read the initial comment too quickly the first time through.

I'm not really worried about footprint, and if no one else is highly concerned about the dynamic footprint increase, then we should move forward, though I would like to request that @vijaysun-omr do the merge as the strongest overall voice on performance topics.

I just wanted to make sure we considered (and documented that it was done) performance effects before we blindly accepted a new field in such a common data structure. We haven't always been explicit about documenting that discussion in the OMR project and it can be hard to recollect when problems are found months later so I think documenting here is a generally good idea.

vijaysun-omr:
I think the overall impact on runtime footprint here is unlikely to be significant enough to prevent this change from going in. Even in a compilation with 30k CFG edges (quite a large compilation) we are looking at a 120k increase in footprint which is'nt going to regress anything that we measure on the Java side too much. Obviously, we need to avoid making a habit of adding fields in these core classes without careful consideration but I do not see a trend in that direction (this is one of the rare PRs that has added a field to one of the core classes and we having this discussion before we are accepting of even this change). 

The other concern is if the extra footprint used by this change pushes us over some threshold that makes some compilation(s) fail. This could happen with the change, but if we were within 100k of failing a key compilation in some benchmark we care about, we would have encountered that issue one of these days anyway (e.g. with some different inlining decisions or different sampling/profiling information). 

Hopefully the addition of a generic id field makes it easier to handle some things that might been candidates for being placed as fields inside a CFG edge class via more transient side tables; so there is reason to allow this change in particular in my opinion. 

I do want to encourage bringing up these kinds of performance questions though and documenting them for general awareness. 

I am merging this now.

PR #799, opened 2017-03-01
Add missing z Systems instructions

fjeremic:
Adds missing z Systems instruction definitions which are to
be used for miscellaneous hardware exploitation.

Add missing processor detection skeleton and corresponding option which
is to be used to query support for miscellaneous hardware support.

Add common codegen support for generating and querying missing z Systems
instructions.


0xdaryl:
LGTM.  Will merge when CI passes.

PR #801, opened 2017-03-01
Make DDRgen run without crashing on Linux x86-64

twlogger:
Made minor changes to genSuperset.cpp and DwarfScanner.cpp. 

Signed-off-by: Tony Wang <twlogger@gmail.com>

### DIFF
--- a/tools/ddrgen/src/scanners/dwarf/DwarfScanner.cpp
+++ b/tools/ddrgen/src/scanners/dwarf/DwarfScanner.cpp
@@ -60,6 +60,7 @@ DwarfScanner::getBlacklist(Dwarf_Die die)
 
 	char **fileNamesTableConcat = NULL;
 	char *compDir = NULL;
+	Dwarf_Bool hasAttr = false;
 	Dwarf_Error error = NULL;
 	Dwarf_Attribute attr = NULL;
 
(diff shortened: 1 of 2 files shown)


charliegracie:
@jason-hall Can you review this please?

tools/ddrgen/src/scanners/dwarf/DwarfScanner.cpp
```diff
@@ -77,9 +77,11 @@ DwarfScanner::getBlacklist(Dwarf_Die die)
 		goto Failed;
 	}
 
-	if (DW_DLV_ERROR == dwarf_formstring(attr, &compDir, &error)) {
-		ERRMSG("Getting compilation directory string: %s\n", dwarf_errmsg(error));
-		goto Failed;
+	if (NULL != attr) {
```
jason-hall: If the attribute DW_AT_comp_dir does not exist, the Dwarf API will not necessarily just return null. I've had it often crash when attempting to use dwarf_attr() on an attribute that a Die does not have. Thus, your change can still segfault. The dwarf spec states one should use dwarf_hasattr().

The correct way to do this with the dwarf_hasattr() function is like this:
Dwarf_Bool hasAttr = false;
if (DW_DLV_ERROR == dwarf_hasattr(die, DW_AT_comp_dir, &hasAttr, &err)) {
   ERRMSG(...
   rc = DDR_RC_ERROR;
   goto Failed;
}
if (hasAttr) {
    if (DW_DLV_ERROR == dwarf_attr(die, DW_AT_comp_dir, &attr, &error)) {
        ERRMSG("Getting compilation directory attribute: %s\n", dwarf_errmsg(error));
        goto Failed;
    }
    ....

tools/ddrgen/src/scanners/dwarf/DwarfScanner.cpp
```diff
@@ -191,7 +193,10 @@ DwarfScanner::blackListedDie(Dwarf_Die die, bool *dieBlackListed)
 		/* If the decl_file matches an entry in the file table not beginning with "/usr/",
 		 * then we are interested in it. Also filter out decl file "<built-in>".
 		 */
-		if ((NULL == strstr(_fileNamesTable[declFile - 1], "<built-in>"))
+		if (0 == declFile) {
+			/* What happens when there is no file? */
```
jason-hall: Re "What happens when there is no file?"
If a Die has no decl_file attribute, then it will typically not have the decl_file attribute and the above dwarf_hasattr() check will return false. Its strange that it would crash here now but did not before. I guess there must be some odd case where it uses 0 to indicate no file.

twlogger: I took a look at the DWARF documentation, and it says "The value 0 indicates that no source file has been specified."

So the decl_file attribute being 0 is supported, so I think we need this check for that case.

jason-hall:
You should probably squash it to a single commit and use a name more descriptive than "Initial commit". The second commit's name is better.

Issue #781, opened 2017-02-24
Dis-aliasing unsafe array shadows from field/static shadows


gita-omr:
We currently dis-alias unsafe array shadows from all symbols in cpSymRefs. The latter is populated by findOrCreateCPSymbol() and contains references to constants in CP. 

We also would like to keep track of other symbols with cpIndex >= 0 (e.g. fields and statics) and dis-alias them from unsafe array shadows as well.

PR #802, opened 2017-03-01
Track all CP symbols using cpSymRefs

gita-omr:
See Issue #781 for high level rational.

In this PR:
-  rename cpSymRefs into cpConstantSymRefs since this bitvector currently tracks only constants in CP
-  add new cpSymRefs bitvector for tracking *all* symbol references in CP
-  set cpSymRefs whenever cpConstantSymRefs is set 

cpSymRefs can now be used for tracking any symbol in CP. These symbols will be automatically dis-aliased from unsafe array shadows since there is already code for that in Aliases.cpp. The only difference is that this used to be done only for constants and now all CP symbols will be dis-aliased from unsafe array shadows, as long as they are included into cpSymRefs by a particular language implementation.

Issue: #781

### DIFF
--- a/compiler/compile/OMRAliasBuilder.cpp
+++ b/compiler/compile/OMRAliasBuilder.cpp
@@ -46,6 +46,7 @@ OMR::AliasBuilder::AliasBuilder(TR::SymbolReferenceTable *symRefTab, size_t size
      _unsafeSymRefNumbers(sizeHint, c->trMemory(), heapAlloc, growable),
      _unsafeArrayElementSymRefs(sizeHint, c->trMemory(), heapAlloc, growable),
      _gcSafePointSymRefNumbers(sizeHint, c->trMemory(), heapAlloc, growable),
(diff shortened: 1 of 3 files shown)


fjeremic:
@gita-omr could you please update the commit header to conform with our commit guidelines and could you please add the message from this PR into the commit body?

vijaysun-omr:
To be certain on one point : the dis-aliasing you mentioned would specifically be for unsafe "array" shadows right ? I ask, because it is incorrect to dis-alias against "non array" unsafe shadows and I did not remember seeing separate tracking for unsafe array shadows specifically (though maybe you or someone else added that tracking).

gita-omr:
We specifically track unsafe array shadows vs. all unsafe shadows. This code was added into VP last year and was motivated by a Spark test case in which unsafe shadows where used exclusively to access array elements but where aliased with field shadows.

vijaysun-omr:
Thanks, I will merge once checks pass.

PR #807, opened 2017-03-01
Run treeSimplification before fieldPrivatization

gita-omr:
Run treeSimplification before FieldPrivatization to remove redundant code left by LoopVersioner that can prevent field privatization (due to nullchecks, calls, etc.)


vijaysun-omr:
I assume the newly added pass only runs at hotness levels >= hot, correct ?

gita-omr:
Correct, this opt group runs at optlevel >= hot.

vijaysun-omr:
Okay I am not concerned with the very small compilation time increase that would happen due to this extra pass; we run many more expensive optimizations at >= hot such that the extra pass of tree simplification should not result in noticeable compilation time increase. Merging.

vijaysun-omr:
Checks had passed before; Gita updated the commit message now and that's why the CI build is in progress.

PR #813, opened 2017-03-02
Implement Reload

jduimovich:
Implemented a default Reload. It will simply reload all the same values as Commit had saved.  

This version conservatively reloads everything which is probably ok for garbage collected systems where objects may move. 

We should consider some enhancements at some point.
1) option to reload some or all of the elements via selective reload to ensure you don't refetch from memory if not needed. 
2) option to take into account stack size changes before reload. This is be useful in VMs where the virtual machine changes the stack with pops of arguments , and pushed results. In that case, you would adjust the stack for pops, and the reload the number of new elements pushed by the VM.

### DIFF
--- a/compiler/ilgen/VirtualMachineOperandStack.cpp
+++ b/compiler/ilgen/VirtualMachineOperandStack.cpp
@@ -87,8 +87,25 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
       }
    }
 
-void
-VirtualMachineOperandStack::MergeInto(OMR::VirtualMachineOperandStack *other, TR::IlBuilder *b)
+   void
+   VirtualMachineOperandStack::Reload(TR::IlBuilder* b)
+   {
+       TR::IlType* Element = _elementType;
+       TR::IlType* pElement = _mb->typeDictionary()->PointerTo(Element);
+       // reload the elements back into the simulated operand stack
+       // If the # of stack element has changed, the user should adjust the # of elements 
+       // using Drop beforehand to add/delete stack elements.
+       TR::IlValue* stack = b->Load("OperandStack_base");
+       for (int32_t i = _stackTop; i >= 0; i--) {
+           _stack[i] = b->LoadAt(pElement,
+               b->IndexAt(pElement,
+                   stack,
+                   b->ConstInt32(i - _stackOffset)));
+       }
+   }
+
+   void
+   VirtualMachineOperandStack::MergeInto(OMR::VirtualMachineOperandStack* other, TR::IlBuilder* b)
    {
    TR_ASSERT(_stackTop == other->_stackTop, "stacks are not same size");
    for (int32_t i=_stackTop;i >= 0;i--)
--- a/compiler/ilgen/VirtualMachineOperandStack.hpp
+++ b/compiler/ilgen/VirtualMachineOperandStack.hpp
@@ -106,6 +106,14 @@ class VirtualMachineOperandStack : public VirtualMachineState
     * @param b the builder where the operations will be placed to recreate the virtual machine operand stack
     */
    virtual void Commit(TR::IlBuilder *b);
+   
+   /**
+    * @brief read the virtual machine stack back into the simulated operand stack
+    * @param b the builder where the operations will be placed to recreate the simulated operand stack 
+    * Users can call Drop beforehand with the appropriate positive or negative count to ensure the simulated
+    * stack accounts for new or dropped virtual machine stack elements. 
+    */
+   virtual void Reload(TR::IlBuilder *b);
 
    /**
     * @brief create an identical copy of the current object.
--- a/jitbuilder/release/src/OperandStackTests.cpp
+++ b/jitbuilder/release/src/OperandStackTests.cpp
@@ -311,6 +311,30 @@ verifyResult12(STACKVALUETYPE top)
    OperandStackTestMethod::verifyStack("11", 3, 3, 5, 4, expectedResult12Top);
    }
 
+// used to compare expected values and report fail it not equal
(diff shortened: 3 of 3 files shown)


fjeremic:
@jduimovich could you please amend the commit to include your PR comment into the commit body so as to conform with the [commit guidelines](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines)? You will also need to sign off on your commit.

Interestingly enough the ip-validation succeeded for b2ed23a even though the commit wasn't signed which is contrary to what the contribution guidelines suggest:

> Remember, if a blank line is found anywhere after the Signed-off-by line, the Signed-off-by: will be considered outside of the footer, and will fail the automated Signed-off-by validation.

@charliegracie is the validation perhaps broken?

charliegracie:
There does not appear to be a commit message at all.  I am not sure if the GitHub UI is having issues right now or not.  @jduimovich the text from the Pull Request would make a great Commit message.  You do have to sign the commit.

@fjeremic 
Pull Requests do NOT require and Issue. Also a commit message does NOT have to include the Issue: tag.  It is good practice to include it if you do have an Issue open though.  I do not believe there is an Issue for this so it would not be required.  However if you do have an Issue and the commit will resolve the Issue using one of the GitHub keywords (Fixes, Fixed, Resolves, etc.) instead of Issue will ensure that the Issue is closed when the PR is merged.

compiler/ilgen/VirtualMachineOperandStack.hpp
```diff
@@ -106,6 +106,14 @@ class VirtualMachineOperandStack : public VirtualMachineState
     * @param b the builder where the operations will be placed to recreate the virtual machine operand stack
     */
    virtual void Commit(TR::IlBuilder *b);
+   
+   /**
+    * @brief read the virtual machine stack back into the simulated operand stack
+    * @param b the builder where the operations will be placed to recreate the simulated operand stack 
+    * Users can call Drop with the appropriate positive or negative count to ensure the simulated
```
mstoodle: maybe add a "beforehand" just to be super clear? Users can call Drop *beforehand* with the appropriate...

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -87,8 +87,25 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
...
+   {
+       TR::IlType* Element = _elementType;
+       TR::IlType* pElement = _mb->typeDictionary()->PointerTo(Element);
+       // reload the elements back into the simulated operand stack
+       // If the # of stack element has changed, the user should adjust the # of elements 
+       // using Drop to add/delete stack elements.
+       TR::IlValue* stack = b->Load("OperandStack_base");
+       for (int32_t i = _stackTop; i >= 0; i--) {
+           _stack[i] = b->LoadAt(Element,
+               b->IndexAt(pElement,
```
mstoodle: not the indentation I normally apply here, though I suppose it's not mandatory. I like to line up the "b"s and insert indentation between the -> and the first operation to reflect the structure of the expression tree.

jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -452,6 +482,25 @@ OperandStackTestMethod::testStack(TR::BytecodeBuilder *b, bool useEqual)
 
    COMMIT(mergeBB);
    mergeBB->Call("verifyResult12", 1, TOP(mergeBB));
+ 
+   int amountToAdd = 10;
+  // Reload test. Call a routine that modifies stack elements passed to it. 
+   // Test by reloading an test the popped values
```
mstoodle: typo: an -> and

jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -311,6 +311,30 @@ verifyResult12(STACKVALUETYPE top)
    OperandStackTestMethod::verifyStack("11", 3, 3, 5, 4, expectedResult12Top);
    }
 
+// used to compare expected values and report fail it not equal
+void
+verifyValuesEqual(STACKVALUETYPE v1, STACKVALUETYPE v2)
+   { 
+   REPORT1(v1 == v2, "verifyValuesEqual", v1);
```
mstoodle: If the values aren't equal, maybe use REPORT2 and pass "v2", v2 as well?

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -97,7 +97,7 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
        // using Drop to add/delete stack elements.
        TR::IlValue* stack = b->Load("OperandStack_base");
        for (int32_t i = _stackTop; i >= 0; i--) {
-           _stack[i] = b->LoadAt(Element,
+           _stack[i] = b->LoadAt(pElement,
```
mstoodle: I would prefer you squash this commit with the previous one, so there isn't a "broken" point in the history.

jitbuilder/release/src/OperandStackTests.cpp
jduimovich: Changed to REPORT2 in refreshed commit.

jduimovich:
Well, all the rebasing and squashing has punted on some typoes/comments needed fixedin the review  but they have been changed in the current branch.

mstoodle:
Weird. Your signed-off-by header still doesn't show in the commit, but the ip validation check details claim it's there. I assume you did sign the (now single) commit, @jduimovich ?

Thanks for taking the time to fix according to the comments..I know git wasn't being your friend for this one.

PR #814, opened 2017-03-03
Refactor ValuePropagation into OMR layer and TR layer

liqunl:
VP refactor

This commit contains the following change:
1) Move ValuePropagation, GlobalValuePropagation and
   LocalValuePropagation to OMR and map them to TR
2) Move ArraycopyTransformation to its own hpp and cpp
3) Move ValuePropagation::getBCDSignConstraints out of OMR


liqunl:
This is the first version of vp refactor change and more change is to be added to this PR. Here is a quick summary of what has been changed:
The 5 commits move most of the vp code into OMR::ValuePropagation. GlobalValuePropagation and LocalValuePropagation keep staying in TR but are moved to their own header file since the use of TR::ValuePropagation is not allowed in OMRValuePropagation.cpp without seeing its implementation, and same to functions containing use of TR::ValuePropagation. The implementations of OMR::VP, GlobalVP and LocalVP are in OMRValuePropagation.cpp except some of VP's and LocalVP's member functions are implemented in ValuePropagationCommon.cpp. 

ValuePropagation::getBCDSignConstraints is moved from trj9 to omr and is enclosed in J9_PROJECT_SPECIFIC macro for now, but it will be moved to J9 VP once we have one, and we will have one in this pull request later.

liqunl:
Can't find Xiaoli in github but have poked her on slack. FYI @andrewcraik @vijaysun-omr

compiler/optimizer/ValuePropagation.hpp
```diff
@@ -16,1099 +16,23 @@
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
  *******************************************************************************/
 
-#ifndef VALUEPROP_INCL
-#define VALUEPROP_INCL
+#ifndef TR_VALUEPROP_INCL
```
xliang6: Please spell out TR_VALUEPROPAGATION_INCL.

jitbuilder/build/files/common.mk
```diff
@@ -174,7 +174,7 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPConstraint.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlers.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlersCommon.cpp \
-    $(JIT_OMR_DIRTY_DIR)/optimizer/ValuePropagation.cpp \
```
xliang6: This appears to be a duplicated set of the same lines above. Can you delete this set since we only need one?

liqunl: Yes, this need to be deleted. Will carefully review the change before delivery.

compiler/optimizer/GlobalValuePropagation.hpp
```diff
@@ -0,0 +1,75 @@
...
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *******************************************************************************/
+
+#ifndef GLOBALVALUEPROP_INCL
+#define GLOBALVALUEPROP_INCL
```
xliang6: Please spell out: GLOBALVALUEPROPAGATION_INCL

compiler/optimizer/GlobalValuePropagation.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "env/jittypes.h"                     // for uintptrj_t
+#include "il/DataTypes.hpp"                   // for DataTypes, TR_YesNoMaybe
+#include "il/ILOpCodes.hpp"                   // for ILOpCodes::DIVCHK
+#include "il/Node.hpp"                        // for Node, vcount_t
+#include "infra/Link.hpp"                     // for TR_LinkHead, TR_Link, etc
+#include "infra/List.hpp"                     // for List, TR_ScratchList
+#include "optimizer/Optimization.hpp"         // for Optimization
+#include "optimizer/OptimizationManager.hpp"  // for OptimizationManager
+#include "optimizer/ValuePropagation.hpp"
+
```
xliang6: The set of includes can be trimmed if ValuePropagation.hpp is included already?

compiler/optimizer/LocalValuePropagation.hpp
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *******************************************************************************/
+
+#ifndef LOCALVALUEPROP_INCL
+#define LOCALVALUEPROP_INCL
```
xliang6: Please spell out: LOCALVALUEPROPAGATION_INCL

compiler/optimizer/LocalValuePropagation.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include "env/jittypes.h"                     // for uintptrj_t
+#include "il/DataTypes.hpp"                   // for DataTypes, TR_YesNoMaybe
+#include "il/ILOpCodes.hpp"                   // for ILOpCodes::DIVCHK
+#include "il/Node.hpp"                        // for Node, vcount_t
+#include "infra/Link.hpp"                     // for TR_LinkHead, TR_Link, etc
+#include "infra/List.hpp"                     // for List, TR_ScratchList
+#include "optimizer/Optimization.hpp"         // for Optimization
+#include "optimizer/OptimizationManager.hpp"  // for OptimizationManager
+#include "optimizer/ValuePropagation.hpp"
+
```
xliang6: The set of includes can be trimmed if ValuePropagation.hpp is included already?

liqunl:
@xliang6 Can you review the last 4 commits I just pushed?

liqunl:
@andrewcraik Can you also take a look at my change? The commits are a little be messy but you can read them as a whole if you like. There is J9 change that i'll create a MR for tmr. The changes have been tested with personal build and all jobs passed.

jitbuilder/build/files/common.mk
```diff
@@ -174,8 +176,6 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPConstraint.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlers.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlersCommon.cpp \
```
xliang6: Please delete the following as they are duplicated as well:

```
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPConstraint.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlers.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VPHandlersCommon.cpp \

File: compiler/optimizer/OMRValuePropagation.hpp [hunk 8]
```diff
@@ -445,6 +444,7 @@ class ValuePropagation : public TR::Optimization
    void removeRestOfBlock();
    void mustTakeException();
    void processTrees(TR::TreeTop *startTree, TR::TreeTop *endTree);
+   virtual void transformRecognizedMethod(TR::Node *node) {}
```
xliang6: I can't find where this new method is called? OMR VP needs to invoke this method to get the different behaviors we want for different projects.

liqunl:
@xliang6 It looks like ArraycopyTransformation is a j9 specific opt, maybe we should leave it in TR and refactor it later.

compiler/optimizer/OMRValuePropagation.hpp
liqunl: This will be called near the beginning of constrainCall in VPHandlers.cpp, together with a few other lines of change, as part of my vp opt. I guess I should leave them to a different PR.

liqunl:
@xliang6 Can you take a quick look as the final review so that I can ask @vijaysun-omr to merge it?

0xdaryl:
Please create a more descriptive title for this PR.

andrewcraik:
@liqunl I am wanting to confirm there is no functional change in this beyond breaking VP apart into two components correct? What is the fate of ArraycopyTransformation is it going to stay put or move? Is any other code moving? I think this needs to be spelled out explicitly given the enormity of this change.

I would prefer having one commit to rename the files etc and a separate one to move that transformation out of OMR if that is what we are doing - it will make testing and fault isolation later easier.

Have we verified compile-time using a benchmark of some kind to make sure that the refactoring is not going to cause a regression in that area?

liqunl:
@andrewcraik There is no function change. The plan in this commit is to move ArraycopyTransformation to its own hpp and cpp, nothing has changed for it in terms of namespace. The entire class can be moved to J9 later. No code has been moved other than the 3 items I listed in the first comment.

> I would prefer having one commit to rename the files etc and a separate one to move that transformation out of OMR if that is what we are 

You mean for all the classes I've touched in this commit?

I have not run any benchmark to measure the compile time, I can get a few runs tonight or this afternoon if the booking allows.

andrewcraik:
My comment about the commits is to say in future when changing functionality and renaming files there should be two commits - one to rename and one to change so that the history doesn't get lost. @0xdaryl not sure your thoughts in this case.

I don't think I like ArraycopyTransformation moving into its own file - there are many different transformations and so grouping them by project would seem to make sense - ie there should be more than one transformation in a file otherwise we will end up with dozens of files so I think we need a more generic name like OMRVPTransformers or something like that.

compiler/optimizer/OMRValuePropagation.hpp
xliang6: I see. Please create this method as a part of that set of changes to make it logically complete.

liqunl: This method has been deleted in the latest commit and will be added in other PR.

liqunl:
@andrewcraik The commit is split to 2 parts, one is renaming the file, the other is the real change.

liqunl:
Change tested on dt3, no perf regression nor compile time regression is seen.

liqunl:
@andrewcraik @xliang6 The change has been modified so that only VP is refactored to OMR, everything else stays the same (except getBCDSignConstraints is moved to J9 because that's the right place for it). And personal build passed, could you review it again?

liqunl:
I want to add that the implementation of some inner classes of VP have been moved from OMRValuePropagation.hpp to OMRValuePropagation.cpp due to the fact that their member functions use TR::VP's member function and TR::VP is just a forward declaration in OMRValuePropagation.hpp.

jitbuilder/build/files/common.mk
```diff
@@ -171,11 +171,6 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     $(JIT_OMR_DIRTY_DIR)/optimizer/ValueNumberInfo.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VirtualGuardCoalescer.cpp \
     $(JIT_OMR_DIRTY_DIR)/optimizer/VirtualGuardHeadMerger.cpp \
-    $(JIT_OMR_DIRTY_DIR)/optimizer/VPConstraint.cpp \
```
andrewcraik: Why are these removed?

liqunl: This is a duplicate, if you look for it in the entire file, you'll find two of them.

liqunl: @andrewcraik Has my answer resolved your concern? If so, can you review the rest of the code?

compiler/optimizer/GlobalValuePropagation.hpp
andrewcraik: This question is still valid - I don't know that all these includes need to be here.

jitbuilder/build/files/common.mk
andrewcraik: Ok - good catch; the deletion is fine.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -136,18 +136,17 @@ TR::ValuePropagation::ValuePropagation(TR::OptimizationManager *manager)
      _arrayCloneCalls(trMemory()),
      _objectCloneTypes(trMemory()),
      _arrayCloneTypes(trMemory()),
-     _bcdSignConstraints(NULL),
      _parmInfo(NULL),
      _parmTypeValid(NULL),
      _constNodeInfo(comp()->allocator())
    {
-   _vcHandler.setVP(this);
+   _vcHandler.setVP((TR::ValuePropagation *)this);
```
andrewcraik: Shouldn't this be a static_cast?

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -456,7 +454,7 @@ TR::VPConstraint *TR::ValuePropagation::getConstraint(TR::Node *node, bool &isGl
       if (trace())
          {
          traceMsg(comp(), "   %s [%p] has existing constraint:", node->getOpCode().getName(), node);
-         rel->print(this, valueNumber, 1);
+         rel->print((TR::ValuePropagation *)this, valueNumber, 1);
```
andrewcraik: static_cast me thinks here too

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -489,7 +487,7 @@ TR::VPConstraint *TR::ValuePropagation::getConstraint(TR::Node *node, bool &isGl
       if (trace())
          {
          traceMsg(comp(), "   %s [%p] has existing global constraint:", node->getOpCode().getName(), node);
-         rel->print(this, valueNumber, 1);
+         rel->print((TR::ValuePropagation *)this, valueNumber, 1);
```
andrewcraik: Another static_cast I think here.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1937,13 +1935,13 @@ void TR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
 
    // We can now make assertions about the arguments to the arraycopy call
    //
-   addBlockConstraint(srcObjNode, TR::VPNonNullObject::create(this));
-   addBlockConstraint(dstObjNode, TR::VPNonNullObject::create(this));
+   addBlockConstraint(srcObjNode, TR::VPNonNullObject::create((TR::ValuePropagation *)this));
```
andrewcraik: I'm not a fan of this casting - isn't there something we can do to avoid this? At the very least these should be static_casts, but they look like self() calls which I am not a fan of. I would have hoped we could find a way to work around this...

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -3861,7 +3859,7 @@ void TR::ValuePropagation::transformConverterCall(TR::TreeTop *callTree)
          TR::TreeTop* thresholdTestTree = TR::TreeTop::create(comp(), TR::Node::createif(TR::ificmplt, TR::Node::createLoad(callNode, lenRef), TR::Node::create(TR::iconst, 0, threshold)));
          createAndInsertTestBlock(comp(), thresholdTestTree, callTree, origCallBlock, slowArraytranslateBlock);
 
-         TR_InlineCall inlineFallback(optimizer(), this);
+         TR_InlineCall inlineFallback(optimizer(), (TR::ValuePropagation *)this);
```
andrewcraik: Another static cast that I am not a fan of...

andrewcraik:
Overall, the change is looking pretty good. The main issue is the use of C-style casts and I want to understand why we need to cast to TR::ValuePropagation. These casts are messy and I suspect we could end up with some very subtle bugs if they were missed when needed - is there a way we could jig the API to avoid this problem because it makes for ugly code...

compiler/optimizer/ValuePropagationCommon.cpp
liqunl: I haven't thought of any solution better than using self().

liqunl:
@andrewcraik The reason why we need to cast to TR::ValuePropagation is that the APIs taking VP as an argument are all expecting TR::ValuePropagation. I can change the code to call self() and use static_cast there, but I cannot think of any better solution.

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -281,118 +284,32 @@ class ValuePropagation : public TR::Optimization
...
+      virtual ValueConstraint * copy (ValueConstraint * vc);
+      virtual TR::Compilation * comp ();
       private:
       TR::ValuePropagation *_vp;
       };
 #else
    class ValueConstraintHandler
       {
       public:
-      void setVP(TR::ValuePropagation *vp) {_vp = vp;}
```
xliang6: This can be changed to void setVP(OMR::ValuePropagation *vp) {_vp = vp;}. As long as we pass in the right ValuePropagation instance to this method, calling its virtual functions should dispatch to the right methods for a given project. Am I missing something?

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -281,118 +284,32 @@ class ValuePropagation : public TR::Optimization
    ValueConstraint *copyValueConstraints(ValueConstraints &valueConstraints);
 
 #if USE_TREES
-   class ValueConstraintHandler : public TREE_HANDLER<ValueConstraint>
+   class ValueConstraintHandler : public TREE_HANDLER <ValueConstraint>
       {
       public:
-      void setVP(TR::ValuePropagation *vp) {_vp = vp;}
```
xliang6: This can be changed to void setVP(OMR::ValuePropagation *vp) {_vp = vp;}. As long as we pass in the right ValuePropagation instance to this method, calling its virtual functions should dispatch to the right methods for a given project. Am I missing something?

compiler/optimizer/ValuePropagationCommon.cpp
xliang6: We need to pass in the right instance of ValuePropagation to setVP to ensure correct behavior for its virtual methods for a given project. We can either add the following to the constructor of the derived class of ValuePropagation or its initialize() method:

_vcHandler.setVP(this);

compiler/optimizer/OMRValuePropagation.hpp
liqunl: No, so we can use OMR::ValuePropagation in OMR layer, and J9::ValuePropagation in J9 layer, which means we do not need the namespace mapping, each layer will be talking about their version of VP since it's virtual inheritance. The only thing is that GVP and LVP need to subclass from the right VP, so we still need something similar to mapping, we're thinking about using a define, for example in ValuePropagation.hpp in trj9
#define TR::ValuePropagation J9::ValuePropagation

What do you think? Is there a known pattern to resolve this issue?

andrewcraik: We can't use a macro #define we would need to use a typedef or something like that...

xliang6: I am not sure I understand your question completely. So I am proposing to keep setVP as the following:

void setVP(OMR::ValuePropagation *vp) {_vp = vp;}

And when we invoke setVP, we need to make sure the right VP object (OMR or VP for another project) is passed as the argument to setVP. After that, invoking virtual methods on the right VP object will dispatch to the desired methods automatically.

liqunl: We agree that we need to take OMR::VP in setVP and other similar places in OMR, in addition, we're thinking of replacing the mapping from creating a VP in TR namespace to a typedef or something similar, for example, in trj9/optimizer/ValuePropagation.hpp
namespace TR { 
typedef J9::ValuePropagation ValuePropagation;
}

So the question is, is there better way or known pattern to do such typedef?

vijaysun-omr:
This is a work in progress and so I am changing the title to reflect that.

compiler/optimizer/OMRValuePropagation.hpp
xliang6: Currently we are creating a new TR::MyClass which inherits from OMR::MyClass or MyClass from a project of your choice for the mapping. One thing that jumps to mind is that MyClass could have platform-specific code, in which case we need to create OMR::X86::MyClass and making sure TR::MyClass is mapped to the right MyClass (for the correct platform) requires extra care. In either case, we will need to putting the correct mapping in ValuePropagation.hpp so consumers of the VP class only needs to include this one file.

@0xdaryl Do you have any thoughts on this?

andrewcraik: I like this solution myself - I think it fixes all the problems: no need to cast, each layer is self contained (you want to add a constraint, a handler, and transformer for something in your language it stays in your language layer and the virtual inheritance makes it work), and best of all everyone else just include ValuePropagation.hpp and uses the TR::VP typedef.

0xdaryl: I'm not proposing a solution yet, just trying to understand.  Do you only need the mapping to the `TR::` namespace in order to construct the proper VP object?  In all other cases, is passing around the base of the hierarchy (i.e., a `OMR::ValuePropagation` class) sufficient?

A bit of history...

At one point for extensible classes the "concrete class" in the TR namespace didn't exist and we simply used a typedef of the most derived class into the TR namespace.  However, this wasn't feasible for a few reasons, some of which may be relevant here, and especially if this is a general pattern that we will use as a precedent:

* typedefs cannot be forward declared,
* typedefs cannot "override" each other, so you have to manage it so that only one is visible to the compiler,
* typedefs of classes cannot be friended (actually they can in C++11, but not all build compilers support this feature yet)

The point about forward declaration may be the most relevant in this discussion and was the biggest deal-breaker for OMR.

liqunl: @andrewcraik Given the concerns from Daryl, plus all other classes are using inheritance to do the mapping, I personally think we should follow the convention we already have, instead of having two ways of doing mapping. What do you think?

andrewcraik: The critical issue here is that we are doing virtual inheritance (true virtual inheritance not extensible inheritance because VP is too complex to be complicated by the extensible class pattern) and we need to make sure that that the LocalValuePropagation and GlobalValuePropagation instances constructed in the OMROptimizationManager for the various optimization groups appropriately specialize when a language specific ValuePropagation has been declared. Daryl is not proposing a solution and we shouldn't pick a solution just because he asked a question.

The key thing we need to figure out is how the OMROptimizationManager can end up talking about the right optimizations specializations - having the OMR ValuePropagation classes talk about OMR::ValuePropagation and, for example, the J9 ValuePropagation classes talk about J9::ValuePropagation is fine. The external interface (where external interface is the set of methods called from outside these files) should be able to talk about TR::ValuePropagation and get the correct behavior. If we can figure that issue out then that is the design to go for.

liqunl: The IPATH will guarantee that the external interface talks about the right TR::ValuePropagation. The problem here is what should we put in ValuePropagation.hpp, the typedef or a new ValuePropagation class in TR namespace. Both will serve our purpose but we need to choose one.

andrewcraik: @mgaudet @0xdaryl @xliang6 can we have your input on the two choices available here and what should be done because we really want to get this finished so we can continue development work to further improve VP.

mgaudet: Sorry, I'm going to bow out of this one, mostly because I don't think I have enough background on what's being asked here. If there's still confusion Monday, maybe Andrew and Liqun and I can talk about this in person with a whiteboard, and get the background filled in. 

Hopefully it doesn't come to that tho, and @xliang6 and @0xdaryl have answers :)

xliang6: @andrewcraik @0xdaryl Here are my two cents...

Thanks @0xdaryl. I looked around the source code and the limitations of typedefs don't really apply in this case. However, given that we are trying to set a precedent on how one will map a class to the TR namespace and I can't predict how future classes will be used (i.e. one of the typedef limitations might become relevant), I vote for creating a new VP class in TR namespace in ValuePropagation.hpp. FWIW, I also don't see a clear downside to this option.

I had a quick chat with Vijay about this as well. He votes for the second option, i.e creating a new TR::VP class. His reasoning is that he would like us to keep the number of concepts used to construct an optimization class simple and since we are already using (virtual) inheritance for these opts, we should stick to it for name mapping as well.

liqunl:
@andrewcraik @xliang6 Could you review the new commit? The new commit will be merged with its parent commit once the review is done. The change includes replacing all TR::VP used in OMRValuePropagation, VPConstraint classes, and VPHandlers to OMR::VP to avoid casting.

liqunl:
A quick test on x86-32 has passed, more testing will be done.

compiler/optimizer/GlobalValuePropagation.hpp
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2000, 2016
```
andrewcraik: Please fix copyright dates so they are correct.

andrewcraik:
One high-level request - the LocalValuePropagation class has a descriptive comment - please doxygen format that comment and make sure it is up to date. GlobalValuePropagation should have a similar comment. Both should have a statement that the classes are not intended to be extended and that the core VP engine is in ValuePropagation which is home of extension points for language specific things.

andrewcraik:
This design is a good first step on making VP extensible per language and will allow the extension point currently needed. There are a couple of issues with the current design that are worth documenting:
- The VPConstraint class is codified to take an OMR constraint and there are a number of places in VP where constraints can be created. We may have issues if one wants to make a language specific constraint which accepts a language specific VP to access language specific VP functions. There are a few possible fixes, but because the constraints lie outside of the VP hierarchy it will involve some kind of casting, templates, or similar to coerce the VP to the correct type so designing something nicer will be necessary when the need for that extension point comes up
- The handler table is not currently extensible - again we don't need this now, but we will eventually need to provide that extension point (both to handle language specific opcodes and, more commonly, wrap or replace OMR constraint logic for OMR nodes)
- There is no nice extension pattern/point for language specific delayed transformations (worklists and transformation processing in doDelayedTransformations) these will eventually be needed as well.

I think this is a good first step and avoids cluttering the code too much. I think we do need to be aware of the extension limitations and some things may need some modifications going forward as more extension points are added. A key design point here is to avoid complicating this already very complicated code with lots of casts and other C++ goop which this implementation avoids nicely.

I'll wait for @xliang6 to comment before making a final judgement, but I think this change has finally arrived at a good spot in terms of not casting and providing the first cut of extension in this opt.

compiler/optimizer/GlobalValuePropagation.hpp
liqunl: What should I change it to? These lines are copied from an existing file.

andrewcraik: Year should be updated to 2017

xliang6: Looking at the changes now... But as for the year, it should be changed to 2017 for all the files changed, not just the new files.

liqunl: Shouldn't we leave it to a different PR?

andrewcraik: No copyright headers should be updated to have the second year match the current year when the file is touched in general. We haven't always been as vigilant as we should, but for the VP files touched they should all have 2017 as the second year as part of this commit.

andrewcraik:
I believe that the answer to @xliang6 question is that we are using virtual inheritance here so no special casting is required - the this in OMR will be a J9 if we have instantiated a J9 so there is nothing special required as long as the extension points are marked virtual.

liqunl:
Comments are formated and a few include files are changed to include the most accurate ones.

liqunl:
@xliang6 As to your question, The VP instance will be either GlobalValuePropagation or LocalValuePropagation, who derive from TR::ValuePropagation, which is declared in ValuePropagation.hpp and thanks to the IPATH, only the ValuePropagation.hpp in the project will be included.

@andrewcraik The change has been tested with my personal build and perf runs yesterday shows no throughput regression nor compile time regression. Do you think we can merge it now?

andrewcraik:
I think we have arrived at a good first cut of the refactoring this optimization requires. There is more to be done in the future but this creates the basic structure and LGTM from me. We just need @xliang6 to confirm her concerns have been addressed and this can go in.

xliang6:
LGTM. As we discussed earlier, please add a reminder not to call any virtual methods in VP constructor for expected correct behavior.

liqunl:
Warning is added to the constructor of OMR::ValuePropagation. @mstoodle  Xiaoli, Andrew and I have come to an agreement that this commit is good to merge, since neither Daryl nor Vijay is around, can you merge this PR?

PR #819, opened 2017-03-03
Relax CFGEdgeList implementation dependencies

lmaisons:
Currently there are many locations where we make the assumption that
TR::CFGEdgeList is implemented using std::list. The most prolific
assumption is assuming the availability of a remove() function.

This change set eliminates those dependancies enabling us to use any of
the standard sequence containers.

**NOTE: This change does not attempt to address or in any way rectify
iterator and reference invalidation safety.  Attempting to use a
container with stricter requirements would require a thorough review and
updating of our container usage.**

### DIFF
--- a/compiler/optimizer/Structure.cpp
+++ b/compiler/optimizer/Structure.cpp
@@ -440,22 +440,30 @@ void TR_RegionStructure::removeEdge(TR::CFGEdge *edge, bool isExitEdge)
    //
    TR::CFGNode *fromNode = edge->getFrom();
    TR::CFGNode *toNode   = edge->getTo();
-   if(std::find(fromNode->getSuccessors().begin(), fromNode->getSuccessors().end(), edge) != fromNode->getSuccessors().end())
-      {
-      fromNode->getSuccessors().remove(edge);
-      TR_ASSERT((std::find(toNode->getPredecessors().begin(), toNode->getPredecessors().end(), edge) != toNode->getPredecessors().end())
-                ,"bad structure found trying to remove an out edge");
-      toNode->getPredecessors().remove(edge);
-      }
-   else
-      {
-      TR_ASSERT((std::find(fromNode->getExceptionSuccessors().begin(), fromNode->getExceptionSuccessors().end(), edge) != fromNode->getExceptionSuccessors().end())
-                ,"bad structure found trying to remove an out edge");
-      fromNode->getExceptionSuccessors().remove(edge);
-      TR_ASSERT((std::find(toNode->getExceptionPredecessors().begin(), toNode->getExceptionPredecessors().end(), edge) != toNode->getExceptionPredecessors().end())
-                ,"bad structure found trying to remove an out edge");
-      toNode->getExceptionPredecessors().remove(edge);
-      }
+   auto successorIterator = std::find(fromNode->getSuccessors().begin(), fromNode->getSuccessors().end(), edge);
+   bool isRegularEdge = successorIterator != fromNode->getSuccessors().end();
+
+   TR::CFGEdgeList &successorList = isRegularEdge ?
+      fromNode->getSuccessors() :
+      fromNode->getExceptionSuccessors();
+
+   TR::CFGEdgeList &predecessorList = isRegularEdge ?
+      toNode->getPredecessors() :
+      toNode->getExceptionPredecessors();
+
+   auto predecessorIterator = std::find(predecessorList.begin(), predecessorList.end(), edge);
+
+   successorIterator = isRegularEdge ?
+      successorIterator :
+      std::find(successorList.begin(), successorList.end(), edge);
+
+   TR_ASSERT(
+      (predecessorIterator != predecessorList.end()) && (successorIterator != successorList.end()),
+      "bad structure found trying to remove an out edge"
+      );
+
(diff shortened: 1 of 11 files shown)


lmaisons:
The wording at http://en.cppreference.com/w/cpp/container/list/remove would seem to suggest that it's perfectly valid to call ```std::list::remove``` with a value that does not appear in the list.

However:

1. The result of calling ```std::find(begin_iterator, end_iterator, value)``` to search for a value which is not actually in a container yields ```end_iterator``` which, when searching a whole container is the past-the-end iterator
2. According to http://en.cppreference.com/w/cpp/container/list/erase it is not permissible to call ```generic_sequence_container_type::erase``` on the past-the-end iterator, as the iterator must be dereferenceable.

We thus need to include a guard whenever converting ```std::list::remove``` to ```std::find``` and ```generic_sequence_container_type::erase```

lmaisons:
Rebased on upstream

0xdaryl:
This seems OK with me.  @vijaysun-omr @ajcraik : any concerns?

compiler/optimizer/Structure.cpp
```diff
@@ -1223,8 +1231,11 @@ void TR_RegionStructure::collapseIntoParent()
                      if ((*pred)->getFrom()->getNumber() == getNumber() &&
                          toStructureSubGraphNode((*pred)->getFrom())->getStructure() == this)
                         {
-                        nodeInParent->getSuccessors().remove(*pred);
-                        target->getPredecessors().remove(*pred);
+                        TR::CFGEdgeList &successorList = nodeInParent->getSuccessors();
+                        auto successorIterator = std::find(successorList.begin(), successorList.end(), *pred);
+                        if (successorIterator != successorList.end()) successorList.erase(successorIterator);
```
mstoodle: please don't put the "then" code on the same line as the condition

compiler/optimizer/Structure.cpp
```diff
@@ -1241,7 +1252,9 @@ void TR_RegionStructure::collapseIntoParent()
             // Disconnect this edge from the original target node in the child
             // region. This target node is not used in the parent region.
             //
-            toNode->getPredecessors().remove(*edge);
+            TR::CFGEdgeList &predecessorList = toNode->getPredecessors();
+            auto predecessorIterator = std::find(predecessorList.begin(), predecessorList.end(), *edge);
+            if (predecessorIterator != predecessorList.end()) predecessorList.erase(predecessorIterator);
```
mstoodle: same here, please put the "then" code on a separate line

compiler/optimizer/Structure.cpp
```diff
@@ -1276,8 +1289,11 @@ void TR_RegionStructure::collapseIntoParent()
                      if (((*pred)->getFrom()->getNumber() == getNumber()) &&
                          (toStructureSubGraphNode((*pred)->getFrom())->getStructure() == this))
                         {
-                        nodeInParent->getExceptionSuccessors().remove(*pred);
-                        target->getExceptionPredecessors().remove(*pred);
+                        TR::CFGEdgeList &successorList = nodeInParent->getExceptionSuccessors();
+                        auto successorIterator = std::find(successorList.begin(), successorList.end(), *pred);
+                        if (successorIterator != successorList.end()) successorList.erase(successorIterator);
```
mstoodle: and here :)

compiler/optimizer/Structure.cpp
```diff
@@ -1287,7 +1303,9 @@ void TR_RegionStructure::collapseIntoParent()
                }
             TR_ASSERT(target, "Structure error collapsing regions");
 
-            toNode->getExceptionPredecessors().remove(*edge);
+            TR::CFGEdgeList &predecessorList = toNode->getExceptionPredecessors();
+            auto predecessorIterator = std::find(predecessorList.begin(), predecessorList.end(), *edge);
+            if (predecessorIterator != predecessorList.end()) predecessorList.erase(predecessorIterator);
```
mstoodle: and here :)

compiler/optimizer/Structure.cpp
```diff
@@ -1223,8 +1231,11 @@ void TR_RegionStructure::collapseIntoParent()
                      if ((*pred)->getFrom()->getNumber() == getNumber() &&
                          toStructureSubGraphNode((*pred)->getFrom())->getStructure() == this)
                         {
-                        nodeInParent->getSuccessors().remove(*pred);
-                        target->getPredecessors().remove(*pred);
+                        TR::CFGEdgeList &successorList = nodeInParent->getSuccessors();
+                        auto successorIterator = std::find(successorList.begin(), successorList.end(), *pred);
+                        if (successorIterator != successorList.end()) successorList.erase(successorIterator);
+
+                        target->getPredecessors().erase(pred);
```
mstoodle: Is the find->erase just not required on the getPredecessors case because you know it was found in the successor's list? if so, that might be worthy of a comment in these places where you skip doing the find in the predecessor lists. Or (wait for it) maybe one generic service on cfgnode could take care of it without mostly repeating the code on the client side :) ? Maybe I'm missing something...

lmaisons:
I'm going to close this as pull request #1135 represents what I believe to be the best way forward.

PR #820, opened 2017-03-03
Consolidate logics related to binary generation

0dvictor:
De-duplicate logics related to instruction binary generation.

### DIFF
--- a/compiler/x/codegen/OMRInstruction.hpp
+++ b/compiler/x/codegen/OMRInstruction.hpp
@@ -198,7 +198,10 @@ class OMR_EXTENSIBLE Instruction : public OMR::Instruction
    virtual bool dependencyRefsRegister(TR::Register *reg);
 
    virtual uint8_t *generateBinaryEncoding();
-   virtual int32_t estimateBinaryLength(int32_t currentEstimate);
+   virtual int32_t  estimateBinaryLength(int32_t currentEstimate);
+   virtual uint8_t* generateOperand(uint8_t* cursor) { return cursor; }
+   virtual bool     needsRepPrefix();
+   virtual bool     needsLockPrefix();
    virtual EnlargementResult enlarge(int32_t requestedEnlargementSize, int32_t maxEnlargementSize, bool allowPartialEnlargement) { return EnlargementResult(0, 0); }
 
    enum TR_UpperHalfRefConditions
--- a/compiler/x/codegen/X86BinaryEncoding.cpp
+++ b/compiler/x/codegen/X86BinaryEncoding.cpp
@@ -176,25 +176,50 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
 
 // -----------------------------------------------------------------------------
 // OMR::X86::Instruction:: member functions
-
-uint8_t *OMR::X86::Instruction::generateBinaryEncoding()
+bool OMR::X86::Instruction::needsRepPrefix()
    {
-   uint8_t *instructionStart = self()->cg()->getBinaryBufferCursor();
-   uint8_t *cursor           = instructionStart;
-   if (self()->getOpCode().needsRepPrefix())
+   return self()->getOpCode().needsRepPrefix();
+   }
+
+bool OMR::X86::Instruction::needsLockPrefix()
+   {
+   return self()->getOpCode().needsLockPrefix();
+   }
+
+uint8_t* OMR::X86::Instruction::generateBinaryEncoding()
+   {
+   // Normally the loop only executes once,
+   // it is needed to avoid recursive calls when generateOperand() requests to regenerate the binary code by returning NULL;
+   for(;;)
       {
-      *cursor++ = 0xf3;
+      uint8_t *instructionStart = self()->cg()->getBinaryBufferCursor();
+      uint8_t *cursor           = instructionStart;
+      self()->setBinaryEncoding(instructionStart);
+      if (self()->needsRepPrefix())
+         {
+         *cursor++ = 0xf3;
+         }
+      if (self()->needsLockPrefix())
+         {
+         *cursor++ = 0xf0;
+         }
+      cursor = self()->generateRepeatedRexPrefix(cursor);
+      cursor = self()->getOpCode().binary(cursor, self()->rexBits());
+      cursor = self()->generateOperand(cursor);
+      // cursor is normal not NULL, and hence we can finish binary encoding and exit the loop
+      // cursor is NULL when generateOperand() requests to regenerate the binary code, which may happen during encoding of memref with unresolved symbols on 64-bit
+      if (cursor)
+         {
+         self()->setBinaryLength(cursor - instructionStart);
+         self()->cg()->addAccumulatedInstructionLengthError(self()->getEstimatedBinaryLength() - self()->getBinaryLength());
+         return cursor;
+         }
       }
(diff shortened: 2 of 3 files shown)


compiler/x/codegen/OMRInstruction.hpp
```diff
@@ -198,7 +198,9 @@ class OMR_EXTENSIBLE Instruction : public OMR::Instruction
    virtual bool dependencyRefsRegister(TR::Register *reg);
 
    virtual uint8_t *generateBinaryEncoding();
-   virtual int32_t estimateBinaryLength(int32_t currentEstimate);
+   virtual int32_t  estimateBinaryLength(int32_t currentEstimate);
+   virtual uint8_t* operand(uint8_t* cursor) { return cursor; }
```
fjeremic: What are the semantics of this function? May I suggest to include some Doxygen documentation as it is not intuitive what purpose this function serves.

0dvictor: Thank-you for point it out your concerns. I have renamed the function to generateOperand which explains itself.

0xdaryl:
In general, the change looks good and is a nice conceptual cleanup.

As this change is fairly pervasive and fundamental to X86 binary encoding, please describe your testing methodology to give assurance that the affected code paths were exercised correctly.

0dvictor:
I've done traceCG on testjit for both 32-bit and 64-bit; the trace logs confirmed that the generated binaries before and after this PR were same.

andrewcraik:
@0dvictor this needs rebasing to fix the conflicts

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -176,21 +176,36 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
...
+         *cursor++ = 0xf3;
+         }
+      if (self()->needsLockPrefix())
+         {
+         *cursor++ = 0xf0;
+         }
+      cursor = self()->generateRepeatedRexPrefix(cursor);
+      cursor = self()->getOpCode().binary(cursor, self()->rexBits());
+      cursor = self()->generateOperand(cursor);
+      if (cursor)
```
andrewcraik: When can cursor be false - it would seem like you would have generated instructions into the stream but we would fail to set th einstruction length? What do we return if the cursor does become NULL?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -176,21 +176,36 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
 
 // -----------------------------------------------------------------------------
 // OMR::X86::Instruction:: member functions
+bool OMR::X86::Instruction::needsLockPrefix()
```
andrewcraik: Why this helper for lockPrefix and not one for needsRepPrefix? Shouldn't we be consistent?
What is the value of this one-liner?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -176,21 +176,36 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
...
+   }
 
-uint8_t *OMR::X86::Instruction::generateBinaryEncoding()
+uint8_t* OMR::X86::Instruction::generateBinaryEncoding()
    {
-   // *this    swipeable for debugging purposes
-   uint8_t *instructionStart = self()->cg()->getBinaryBufferCursor();
-   uint8_t *cursor           = instructionStart;
-   if (self()->getOpCode().needsRepPrefix())
+   for(;;)
```
andrewcraik: The need for this infinite loop is not immediately clear to me and I think dserves a comment since I would have thought an instruction would encode once and be done?

compiler/x/codegen/X86BinaryEncoding.cpp
0dvictor: cursor is NULL when operand() requests to regenerate the binary code, which may happen during encoding of memref with unresolved symbols on 64-bit.
Comments have been updated.

compiler/x/codegen/X86BinaryEncoding.cpp
0dvictor: Good suggestion. Helper for needsRepPrefix() is added.

compiler/x/codegen/X86BinaryEncoding.cpp
0dvictor: We normally exit the loop within one iteration. The loop is to handle the case operand() requests regenerating current instruction by returning NULL. The previous approach was to recursive call itself, this proposed approach uses a loop to be safer and more efficient.

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -176,25 +176,48 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
...
    {
-   uint8_t *instructionStart = self()->cg()->getBinaryBufferCursor();
-   uint8_t *cursor           = instructionStart;
-   if (self()->getOpCode().needsRepPrefix())
+   return self()->getOpCode().needsLockPrefix();
+   }
+
+uint8_t* OMR::X86::Instruction::generateBinaryEncoding()
+   {
+   for(;;)
```
andrewcraik: Can you comment here on the need for the loop as well please?

0dvictor: Updated.

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -176,25 +176,50 @@ uint8_t getMemoryBarrierBinaryLengthLowerBound(int32_t barrier, TR::CodeGenerato
...
+         }
+      if (self()->needsLockPrefix())
+         {
+         *cursor++ = 0xf0;
+         }
+      cursor = self()->generateRepeatedRexPrefix(cursor);
+      cursor = self()->getOpCode().binary(cursor, self()->rexBits());
+      cursor = self()->generateOperand(cursor);
+      // cursor is normal not NULL, and hence we can finish binary encoding and exit the loop
+      // cursor is NULL when operand() requests to regenerate the binary code, which may happen during encoding of memref with unresolved symbols on 64-bit
```
mgaudet: typo here; comment should be `... when generateOperand()... ` right?

mgaudet: Oh -- ditto above (I think it was just missed when `operand` became `generateOperand`)

0dvictor: Thank you for point it out. It has been fixed.

andrewcraik: should the operand failing to be renamed to generateOperand generate a compile-time error? If not what have you done to check you haven't missed any more?

0dvictor: This is an outdated diff. The comment containing operand() has been updated. Failing to renaming will trigger a compile-time error. This was the only function that refers operand() in comments.

andrewcraik:
Ok, probably needs a rebase but LGTM. @0xdaryl over to you.

0dvictor:
Rebased.

0dvictor:
@0xdaryl Do you mind also review this PR?

PR #823, opened 2017-03-06
Correct off by 1 mistake in test for real method call

ncough:
Whilst calls are potential OSR points, helper calls and
common non-helper symbols are not. This distinction is tested by
comparing the call's symbol reference number with the index of the
last common non-helper symbol. This index marks the end of the
predefined symbols, but can be used as the reference number for a
real method call.


vijaysun-omr:
I can picture the tough bug that led to this one being discovered :)

PR #832, opened 2017-03-06
Start testing MetaDataManager

mgaudet:
With the aim of starting to support tracking metadata, this pull request starts testing the MetaData Manager.


mgaudet:
This build fails in Travis; It turns out that the build dependencies are wrong; `libj9avl` isn't built by the time `testjit` is being linked. 

I'll dig into that tomorrow (though, we can still do some review).

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+   data.startPC = 10;
+   data.endPC   = 100;
+   data.name    = "test";
+
+   ASSERT_TRUE(manager.insertMetaData(&data));
+   ASSERT_TRUE(manager.containsMetaData(&data));
+
+   auto* metadata = manager.findMetaDataForPC(10);
+   ASSERT_TRUE(NULL != metadata);
+   ASSERT_TRUE(NULL !=  metadata->name);
```
Leonardo2718: I don't think this assert is required (https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#string-comparison). Is there a reason for it?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+   TR::MethodMetaDataPOD data;
+   data.startPC = 10;
+   data.endPC   = 100;
+   data.name    = "test";
+
+   ASSERT_TRUE(manager.insertMetaData(&data));
+   ASSERT_TRUE(manager.containsMetaData(&data));
+
+   auto* metadata = manager.findMetaDataForPC(10);
```
Leonardo2718: I think
```c++
auto* metadata = manager.findMetaDataForPC(data.startPC);
```
is clearer.

Also is there a reason for using `auto*` instead of just `auto`?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+   ASSERT_TRUE(NULL != metadata);
+   ASSERT_TRUE(NULL !=  metadata->name);
+   ASSERT_STREQ("test", metadata->name);
```
Leonardo2718: Could this just be

```c++
ASSERT_STREQ(data.name, metadata->name);
```
?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+
+TEST(MetaDataTest, ExactInsertion) {
+   TestEnhancedCodeMetaDataManager manager;
+   ASSERT_TRUE(manager.initializeCodeMetaDataManager());
+
+   // Set up code cache hash table.
+   ASSERT_TRUE(NULL != manager.addFalseCacheInfo(0,1000));
+
+   TR::MethodMetaDataPOD data;
```
Leonardo2718: Would it be useful to have two copies of this object, one for input and one for expected output?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+   ASSERT_TRUE(manager.insertMetaData(&data));
+   ASSERT_TRUE(manager.containsMetaData(&data));
+
+   auto* metadata = manager.findMetaDataForPC(10);
+   ASSERT_TRUE(NULL != metadata);
+   ASSERT_TRUE(NULL !=  metadata->name);
+   ASSERT_STREQ("test", metadata->name);
+
+   ASSERT_TRUE(manager.removeMetaData(&data));
```
Leonardo2718: Could some of these asserts be moved to a separate test case:

```c++
ASSERT_TRUE(manager.insertMetaData(&data));
ASSERT_TRUE(manager.containsMetaData(&data));
ASSERT_TRUE(manager.removeMetaData(&data));
```

?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+   // Set up code cache hash table.
+   ASSERT_TRUE(NULL != manager.addFalseCacheInfo(0,1000));
+
+   TR::MethodMetaDataPOD data;
+   data.startPC = 10;
+   data.endPC   = 100;
+   data.name    = "test";
+
+   ASSERT_TRUE(manager.insertMetaData(&data));
+   ASSERT_TRUE(manager.containsMetaData(&data));
```
Leonardo2718: This entire snippet looks the same as the one for `TEST(MetaDataTest, ExactInsertion)`. Could it be moved into a test fixture (https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests)?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+   std::vector<int> test_points = {11, 25, 99};
+   for (auto i : test_points) { 
+      auto* metadata = manager.findMetaDataForPC(i);
+      ASSERT_TRUE(NULL != metadata) << "Didn't find pc" << i;
+      ASSERT_TRUE(NULL !=  metadata->name);
+      ASSERT_STREQ("test", metadata->name);
+   }
```
Leonardo2718: It's not really clear to me what this test cases is actually testing. Could you add a description? Also, have you considered breaking these up into different test cases.

One last more advanced question (mostly just curios), is there a reason for picking this weak combination of equivalence partitioning and boundary-value analysis?

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+   int iter = 0; 
+   for (auto i : test_points) { 
+      auto* metadata = manager.findMetaDataForPC(i);
+      ASSERT_TRUE(NULL != metadata) << "Didn't find pc" << i;
+      ASSERT_TRUE(NULL !=  metadata->name);
+      ASSERT_STREQ(names[iter++], metadata->name);
+      ASSERT_TRUE(manager.removeMetaData(metadata)); 
+      ASSERT_TRUE(NULL == manager.findMetaDataForPC(i));
+   }
+
```
Leonardo2718: Lots of "magic" numbers here (where did they come from?) so it's not really clear what the test case is trying to detect. Maybe used named values instead and add some more messages to each assert?

fvtest/compilertest/tests/MetaDataTest.cpp
mgaudet: Agree on the first point. To the second, IMO, `auto*` provides a readability hint.

fvtest/compilertest/tests/MetaDataTest.cpp
mgaudet: Honestly, didn't notice that STREQ does the right thing with `NULL`. Will remove.

fvtest/compilertest/tests/MetaDataTest.cpp
mgaudet: Agreed.

fvtest/compilertest/tests/MetaDataTest.cpp
Leonardo2718: Fair enough.

fvtest/compilertest/tests/MetaDataTest.cpp
mgaudet: I can break them into their own test case; your thinking is if I assert on those other test cases, I can then do `insert`, `contains`, etc,  and not assert in a test case not explicitly trying to test `insert` or `contains`?

mgaudet:
Alright -- based on Leo's feedback I'm going to mark this as WIP, and take another stab at this :)

Leonardo2718:
So I had a few general notes here but GitHub decided to get rid of them for some reason.

First, typo in (4712563cb655f32974069c2fd4bb6bfd4ebfc269)[https://github.com/eclipse/omr/pull/832/commits/4712563cb655f32974069c2fd4bb6bfd4ebfc269]: "assering" -> "asserting"

Second, I think that exposing protected members for the sake of testability is completely OK. In Java, this problem is usually solved using reflection but I think this is (much) more elegant.

Lastly, I think it would be good to document test cases a little more with comments and assert messages.

fvtest/compilertest/tests/MetaDataTest.cpp
Leonardo2718: Correct. Having the asserts there is not incorrect (sorry for the double negative), strictly speaking, but it does make it a bit more difficult to understand what's being tested.

fvtest/compilertest/tests/MetaDataTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * The design of the CodeMetaData manager is such that it expects each code
+ * cache to be added to the CodeMetaDataManager. In order to avoid that
+ * coupling in this test, the below class allows allocating 'fake' homes for
+ * tree elements
+ *
+ * As well, this resets the singleton on destruction.
+ */
+class TestEnhancedCodeMetaDataManager : public TR::CodeMetaDataManager {
```
0xdaryl: `TestCompiler` namespace

compiler/runtime/OMRCodeMetaDataManager.cpp
```diff
@@ -294,6 +295,12 @@ CodeMetaDataManager::insertMetaDataRangeInHash(
       return 1;
       }
 
+   // Don't insert zero sized ranges.
+   if (startPC == endPC)
+      {
+      return 1;
+      }
+
```
0xdaryl: What's wrong with inserting zero-sized ranges?  Will this code not function?  Suppose someone wants to insert metadata for a particular code address in their runtime (not sure why, but they might have a need for it).  If we require a range of at least one to cover a code range then it should be documented somewhere in the API.

fvtest/compilertest/runtime/CodeMetaDataPOD.hpp
```diff
@@ -0,0 +1,34 @@
...
+#ifndef TR_METHOD_METADATAPOD_INCL
+#define TR_METHOD_METADATAPOD_INCL
+
+#include "runtime/TestCodeMetaDataPOD.hpp"
+#include "infra/Annotations.hpp"           // for OMR_EXTENSIBLE
+
+namespace TR
+{
+
+struct OMR_EXTENSIBLE MethodMetaDataPOD : public Test::MethodMetaDataPODConnector
```
0xdaryl: namespace Test -> TestCompiler

fvtest/compilertest/runtime/TestCodeMetaDataPOD.hpp
```diff
@@ -0,0 +1,52 @@
...
+#ifndef TEST_METHOD_METADATAPOD_INCL
+#define TEST_METHOD_METADATAPOD_INCL
+
+/*
+ * The following #define(s) and typedef(s) must appear before any #includes in this file
+ */
+#ifndef TEST_METHOD_METADATAPOD_CONNECTOR
+#define TEST_METHOD_METADATAPOD_CONNECTOR
+namespace Test { struct MethodMetaDataPOD; }
+namespace Test { typedef Test::MethodMetaDataPOD MethodMetaDataPODConnector; }
```
0xdaryl: Namespace Test -> TestCompiler for both

fvtest/compilertest/runtime/TestCodeMetaDataPOD.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "infra/Annotations.hpp"  // for OMR_EXTENSIBLE
+#include "runtime/OMRCodeMetaDataPOD.hpp" 
+
+/*
+ * This structure describes the shape of the method meta data information.
+ * It must be a C++ POD object and follow all the rules behind POD formation.
+ * Because its fields may be extracted by a runtime system, its exact layout
+ * shape MUST be preserved.
+ */
+namespace Test
```
0xdaryl: Namespace Test -> TestCompiler

compiler/runtime/OMRCodeMetaDataManager.cpp
mgaudet: Currently, you can insert metadata with a zero sized range, but the query logic will never find it. So you'll insert things into the table that can never be removed. I'll add a comment to this.

mgaudet:
Addressed comments.

PR #835, opened 2017-03-07
Begin to remove warm/cold snippet distinction

0xdaryl:
* Begin to remove deprecated warm snippet support
* Remove unused isWarm parm from snippet functions, and constant fold

This isn't the complete set of work to remove the distinction, but its a cleanup task on the way there.


andrewcraik:
@0dvictor heads up and you may want to verify there isn't anything concerning in the x codegen - I don't think so, but you are working in there more atm.

0xdaryl:
@fjeremic , @0dvictor : I'd appreciate a code review.

compiler/arm/codegen/OMRCodeGenerator.hpp
```diff
@@ -126,7 +126,7 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    void doRegisterAssignment(TR_RegisterKinds kindsToAssign);
    void doBinaryEncoding();
 
-   void emitDataSnippets(bool isWarm = 0);
+   void emitDataSnippets();
    bool hasDataSnippets();
    int32_t setEstimatedLocationsForDataSnippetLabels(int32_t estimatedSnippetStart, bool isWarm = 0);
```
fjeremic: Do we also need to remove the `bool isWarm = 0` here?

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -2931,14 +2931,14 @@ OMR::CodeGenerator::setEstimatedLocationsForSnippetLabels(int32_t estimatedSnipp
...
-OMR::CodeGenerator::emitSnippets(bool isWarm)
+OMR::CodeGenerator::emitSnippets()
    {
    uint8_t *codeOffset;
    uint8_t *retVal;
 
    for (auto iterator = _snippetList.begin(); iterator != _snippetList.end(); ++iterator)
       {
-      if ((*iterator)->isWarmSnippet() == isWarm)
+      if ((*iterator)->isWarmSnippet() == 0)
```
fjeremic: Should this be `if (!(*iterator)->isWarmSnippet())`? Similarly elsewhere.

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -7953,13 +7953,13 @@ TR_S390OutOfLineCodeSection * OMR::Z::CodeGenerator::findS390OutOfLineCodeSectio
...
 TR::S390ConstantDataSnippet *
-OMR::Z::CodeGenerator::findOrCreateConstant(TR::Node * node, void * c, uint16_t size, bool isWarm)
+OMR::Z::CodeGenerator::findOrCreateConstant(TR::Node * node, void * c, uint16_t size)
    {
    CS2::HashIndex hi;
    TR_S390ConstantDataSnippetKey key;
    key.c      = c;
    key.size   = size;
-   key.isWarm = isWarm;
+   key.isWarm = false;
```
fjeremic: Is the `isWarm` field still relevant following this change?

0xdaryl: No it's not.  I have that cleaned up in a subsequent PR.

compiler/codegen/OMRCodeGenerator.cpp
0xdaryl: You're right.  But my next pass through here will remove `isWarmSnippet()` entirely and fold entirely (will always return false).  That's coming in the next PR.

compiler/arm/codegen/OMRCodeGenerator.hpp
0xdaryl: Yes.  There are another batch of functions that need such cleanup (and I have in a subsequent commit).  I was staging these commits as there is surrounding logic that needs to be reasoned about and cleaned up.

fjeremic:
My concerns have been addressed in the above comments. LGTM assuming there is followup work to clean up the rest. @0dvictor CC for review.

0dvictor:
LGTM. The proposed changes should be functional the same as existing X code.

compiler/codegen/OMRCodeGenerator.cpp
vijaysun-omr: Appreciate the staged approach since it makes each commit logically easier for the the reviewer to reason about.

Issue #833, opened 2017-03-07
Fill out JitBuilder conditional support


mstoodle:
JitBuilder (specifically, IlBuilder and BytecodeBuilder) currently do not offer a full complement of conditional services. That includes missing lower level services like unsigned comparisons on IlBuilder, the OrEqual services are also not provided on BytecodeBuilder which makes JIT construction more awkward, and there isn't any support for building short-circuited conditional structures.

All of these things are supported by the underlying OMR compiler, so JitBuilder should be able to offer them.

PR #836, opened 2017-03-07
Enhance JitBuilder conditional services

mstoodle:
Add all IlBuilder IfCmp* conditionals to BytecodeBuilder
Add Unsigned varieties for all conditional services
Create IfAnd and IfOr services to help to create short-circuit conjunction/disjunctions
Create jitbuilder/release/src/Conditionals code sample

Issue: #833

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1480,6 +1480,110 @@ IlBuilder::UnsignedShiftR(TR::IlValue *v, TR::IlValue *amount)
    return returnValue;
    }
 
+/*
+ * @brief IfAnd service for constructing short circuit AND conditional nests (like the && operator)
+ * @param allTrueBuilder builder containing operations to execute if all conditional tests evaluate to true
+ * @param anyFalseBuilder builder containing operations to execute if any conditional test is false
+ * @param numTerms the number of conditional terms
+ * @param ... for each term, provide a TR::IlBuilder object and a TR::IlValue object that evaluates a condition (builder is where all the operations to evaluate the condition go, the value is the final result of the condition)
+ *
+ * Example:
+ * TR::IlBuilder *cond1Builder = OrphanBuilder();
+ * TR::IlValue *cond1 = cond1Builder->GreaterOrEqual(
+ *                      cond1Builder->   Load("x"),
+ *                      cond1Builder->   Load("lower"));
+ * TR::IlBuilder *cond2Builder = OrphanBuilder();
+ * TR::IlValue *cond2 = cond2Builder->LessThan(
+ *                      cond2Builder->   Load("x"),
+ *                      cond2Builder->   Load("upper"));
+ * TR::IlBuilder *inRange = NULL, *outOfRange = NULL;
+ * IfAnd(&inRange, &outOfRange, 2, cond1Builder, cond1, cond2Builder, cond2);
+ */
+void
+IlBuilder::IfAnd(TR::IlBuilder **allTrueBuilder, TR::IlBuilder **anyFalseBuilder, int32_t numTerms, ...)
+   {
+   TR::IlBuilder *mergePoint = OrphanBuilder();
+   *allTrueBuilder = createBuilderIfNeeded(*allTrueBuilder);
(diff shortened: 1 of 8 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1494,9 +1494,8 @@ IlBuilder::integerizeAddresses(TR::IlValue **leftPtr, TR::IlValue **rightPtr)
    {
    TR::IlValue *left = *leftPtr;
    TR::IlValue *right = *rightPtr;
-   if (left->getSymbol()->getDataType() == TR::Address)
+   if (left->getSymbol()->getDataType() == TR::Address && right->getSymbol()->getDataType() == TR::Address)
```
Leonardo2718: From a usability point of view, I'm not sure I like the idea of this silently failing if the two IlValues are not addresses. I'm thinking of a scenario where I accidentally pass something that's an address and another thing that isn't. The call will silently fail to convert the address, causing some assert to fail later on because an operation is not compatible with addresses. Then, I get really confused because I've clearly called `integerizeAddresses` but I'm still getting an address.

jitbuilder/release/src/Conditionals.cpp
```diff
@@ -0,0 +1,234 @@
...
+            Load("L")),
+         Sub(
+            Load("U"),
+            Load("L"))));
+   rc3True->Store("rc3",
+   rc3True->   ConstInt32(1));
+   rc3False->Store("rc3",
+   rc3False->   ConstInt32(0));
+
+   // if (x - L >= (U-L)
```
Leonardo2718: Missing `)` at the end of the comment.

jitbuilder/release/src/Conditionals.cpp
```diff
@@ -0,0 +1,234 @@
...
+                         x_ge_U_builder->   Load("U"));
+
+   TR::IlBuilder *rc2True = NULL, *rc2False = NULL;
+   IfOr(&rc2False, &rc2True, 2, x_lt_L_builder, x_lt_L, x_ge_U_builder, x_ge_U);
+   rc2True->Store("rc2",
+   rc2True->   ConstInt32(1));
+   rc2False->Store("rc2",
+   rc2False->   ConstInt32(0));
+
+   // if (x - L < (U-L)
```
Leonardo2718: Missing `)` at the end of the comment.

charliegracie:
Great change @mstoodle.  This will help a lot. 

It is nice to have a test for all of these ILBuilder API's.  Is there a plan to test all of the BytecodeBuilder APIs as well?  I think there is an Issue for this already but I did not see it quickly in my scan.

mstoodle:
I'm getting there :) . As things get changed, I'm trying to get tests written against them, though technically the code in the release/src directory should be called "samples" rather than "tests" :) . Worklist.cpp does test some (but not all) of the BytecodeBuilder API.

compiler/ilgen/IlBuilder.cpp
mstoodle: you're right about that though the code was in a weirder state to start with. Technically, this integerizeAddresses joker was created for the GT,GE,LT,LE comparisons against addresses, because the OMR compiler doesn't actually define all the opcodes needed to compare against opcodes. I'll take another look at this aspect of the change to see if I can improve it.

Thanks for keeping me honest, @Leonardo2718 !

jitbuilder/release/src/Conditionals.cpp
mstoodle: fixed

jitbuilder/release/src/Conditionals.cpp
mstoodle: fixed

compiler/ilgen/IlBuilder.cpp
mstoodle: ok, i separated the handling of each operand, so that if either operand is an Address it will convert it to Word. I think that's a better solution. I don't want to encourage anyone to use this function, however, without good reason. We aren't missing all that many evaluators, I don't think, and the long term prospect for integerizeAddresses should be gleeful deletion.

PR #837, opened 2017-03-07
Rematerialize call constant nodes

NigelYiboYu:
Constant integers in call parameters can be uncommoned and rematerialized
if it is cheap to load them as immediate values.

1. Add an extra pass before the instruction selection phase to uncommon
constant call parameters

2. Add Z specific rematerialization range, which is the range of signed
32-bit integers.

### DIFF
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -505,6 +505,30 @@ OMR::CodeGenPhase::performLowerTreesPhase(TR::CodeGenerator * cg, TR::CodeGenPha
    }
 
 
+void
+OMR::CodeGenPhase::performUncommonCallConstNodesPhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
+   {
+   TR::Compilation* comp = cg->comp();
+
+   if(comp->getOption(TR_DisableCallConstUncommoning))
+      {
+      traceMsg(comp, "Skipping Uncommon Call Constant Node phase\n");
+      return;
+      }
+
+   phase->reportPhase(UncommonCallConstNodesPhase);
+
+   if (comp->getOption(TR_TraceCG) || comp->getOption(TR_TraceTrees))
+      comp->dumpMethodTrees("Pre Uncommon Call Constant Node Trees");
+
+   TR::LexicalMemProfiler mp(phase->getName(), comp->phaseMemProfiler());
+   LexicalTimer pt(phase->getName(), comp->phaseTimer());
+
+   cg->uncommonCallConstNodes();
+
+   if (comp->getOption(TR_TraceCG) || comp->getOption(TR_TraceTrees))
+      comp->dumpMethodTrees("Post Uncommon Call Constant Node Trees");
+  }
 
 void
 OMR::CodeGenPhase::performReserveCodeCachePhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
@@ -576,6 +600,8 @@ OMR::CodeGenPhase::getName(PhaseValue phase)
    {
    switch (phase)
       {
+      case UncommonCallConstNodesPhase:
+         return "UncommonCallConstNodesPhase";
       case ReserveCodeCachePhase:
          return "ReserveCodeCache";
       case LowerTreesPhase:
--- a/compiler/codegen/OMRCodeGenPhaseFunctionTable.hpp
+++ b/compiler/codegen/OMRCodeGenPhaseFunctionTable.hpp
@@ -25,6 +25,7 @@
  // The entries in this file must be kept in sync with compiler/codegen/OMRCodeGenPhaseEnum.hpp
    TR::CodeGenPhase::performReserveCodeCachePhase,                                           //ReserveCodeCachePhase
    TR::CodeGenPhase::performLowerTreesPhase,                                                 //LowerTreesPhase
+   TR::CodeGenPhase::performUncommonCallConstNodesPhase,                                     //UncommonCallConstNodesPhase
    TR::CodeGenPhase::performSetupForInstructionSelectionPhase,                               //SetupForInstructionSelectionPhase
    TR::CodeGenPhase::performInstructionSelectionPhase,                                       //InstructionSelectionPhase
    TR::CodeGenPhase::performCreateStackAtlasPhase,                                           //CreateStackAtlasPhase
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -94,6 +94,7 @@
 #include "infra/Link.hpp"                           // for TR_LinkHead, etc
 #include "infra/List.hpp"                           // for ListIterator, etc
 #include "infra/Stack.hpp"                          // for TR_Stack
+#include "infra/Checklist.hpp"                      // for TR::NodeCheckList
 #include "infra/TRCfgEdge.hpp"                      // for CFGEdge
 #include "infra/TRCfgNode.hpp"                      // for CFGNode
 #include "optimizer/Optimizations.hpp"
@@ -563,6 +564,91 @@ OMR::CodeGenerator::setUpForInstructionSelection()
       }
 
   }
+
+/**
+ * \brief The uncommon-call-constant-node pass is an extra pre-codegen pass
+ * before instruction selection. It aims to avoid the situation where
+ * a constant integer is made alive across calls and, thereby, lower the preserved register
+ * pressure caused due to this.
+ *
+ * \details
+ *
+ * \verbatim
+ * Take z/Architecture as an example, we could run into the following situation where constant
+ * integer 16 is loaded to a preserved register GPR6, which is then loaded to GPR2 as call
+ * parameter. The GPR6 will be preserved by the callee and can incur extra register shuffling.
+ *
+ * LHI     GPR6,0x10           # immediate value in preserved register
+ * LGR     GPR2,GPR6           # load parameter from preserved register again
+ * LG      GPR14,#609 -216(GPR7)
+ * LGHI    GPR0,0xff28
+ * BASR    GPR14,GPR14         # call virtual function, which will unnecessarily preserve GPR6.
+ * LGR     GPR2,GPR6           # load parameter from preserved register again for the next use
+ * ...
+ *
+ * \endverbatim
+ *
+ * Loading integer constants as immediate values [i.e. LHI GPR2, 0x10] is preferred.
+ * The ranges for immediate value loading on different platforms are governed by the following
+ * platform-specific query:
+ *
+ * \verbatim
+ *   bool materializesLargeConstants()
+ *   bool shouldValueBeInACommonedNode(int64_t value)
+ * \endverbatim
+ *
+*/
+void
+OMR::CodeGenerator::uncommonCallConstNodes()
+   {
+   TR::Compilation* comp = self()->comp();
+   if(comp->getOption(TR_TraceCG))
+      {
+      traceMsg(comp, "Performing uncommon call constant nodes\n");
+      }
+
+   TR::NodeChecklist checklist(comp);
+
+   for (TR::TreeTop* tt = self()->comp()->getStartTree(); tt; tt = tt->getNextTreeTop())
+      {
+       TR::Node* node = tt->getNode();
+       if(node->getNumChildren() >= 1 &&
+               node->getFirstChild()->getOpCode().isFunctionCall())
+          {
+          TR::Node* callNode = node->getFirstChild();
+          if(checklist.contains(callNode))
+             {
+             if(comp->getOption(TR_TraceCG))
+                {
+                traceMsg(comp, "Skipping previously visited call node %d\n", callNode->getGlobalIndex());
+                }
+             continue;
+             }
+
+          checklist.add(callNode);
+
+          for(uint32_t i = 0; i < callNode->getNumChildren(); ++i)
+             {
+             TR::Node* paramNode = callNode->getChild(i);
+             if(paramNode->getReferenceCount() > 1 &&
+                     paramNode->getOpCode().isLoadConst() &&
+                     !self()->isMaterialized(paramNode))
+                {
+                if(self()->comp()->getOption(TR_TraceCG))
+                   {
+                   traceMsg(comp, "Uncommon const node %X [n%dn]\n",
+                            paramNode, paramNode->getGlobalIndex());
+                   }
+
+                TR::Node* newConstNode = TR::Node::create(paramNode->getOpCodeValue(), 0);
+                newConstNode->setConstValue(paramNode->getConstValue());
+                callNode->setAndIncChild(i, newConstNode);
+                paramNode->decReferenceCount();
+                }
+             }
+          }
+      }
+   }
 
 void
 OMR::CodeGenerator::doInstructionSelection()
@@ -3499,27 +3585,8 @@ OMR::CodeGenerator::isMaterialized(TR::Node * node)
    else
       return false;
 
-   return self()->isMaterialized(value);
+   return self()->shouldValueBeInACommonedNode(value);
    }
-
-
-// determine if value fits in the immediate field (if any) of instructions that machine provides
-bool
-OMR::CodeGenerator::isMaterialized(int64_t value)
-   {
-   if(self()->materializesLargeConstants()) //check the int case
-      {
-      int64_t smallestPos = self()->getSmallestPosConstThatMustBeMaterialized();
-      int64_t largestNeg = self()->getLargestNegConstThatMustBeMaterialized();
-
-      if ((value >= smallestPos) ||
-          (value <= largestNeg))
-         return true;
-      }
-   return false;
-   }
-
-
 
 bool
 OMR::CodeGenerator::canNullChkBeImplicit(TR::Node *node)
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -309,6 +309,8 @@ class OMR_EXTENSIBLE CodeGenerator
    TR_FrontEnd *fe();
    TR_Debug *getDebug();
 
+   void uncommonCallConstNodes();
+
    void preLowerTrees();
    void postLowerTrees() {}
 
@@ -614,10 +616,8 @@ class OMR_EXTENSIBLE CodeGenerator
    bool canBeAffectedByStoreTagStalls() { return false; } // no virt, default
 
    bool isMaterialized(TR::Node *); // no virt, cast
-   bool isMaterialized(int64_t); // no virt, cast
-   bool materializesLargeConstants() { return false;}
-   int32_t getLargestNegConstThatMustBeMaterialized() {return -32769;} // no virt, cast
-   int32_t getSmallestPosConstThatMustBeMaterialized() {return 32768;} // no virt, cast
+   bool shouldValueBeInACommonedNode(int64_t) { return false; } // no virt, cast
+   bool materializesLargeConstants() { return false; }
 
    bool canUseImmedInstruction(int64_t v) {return false;} // no virt
    bool needsNormalizationBeforeShifts() { return false; } // no virt, cast
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -255,6 +255,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableBlockVersioner",              "O\tdisable block versioner",                        SET_OPTION_BIT(TR_DisableBlockVersioner), "P"},
    {"disableBranchOnCount",               "O\tdisable branch on count instructions for s390",  SET_OPTION_BIT(TR_DisableBranchOnCount), "F"},
    {"disableBranchPreload",               "O\tdisable return branch preload",                  SET_OPTION_BIT(TR_DisableBranchPreload), "F"},
+   {"disableCallConstUncommoning",        "O\tdisable uncommon call constant node phase",  SET_OPTION_BIT(TR_DisableCallConstUncommoning), "F"},
    {"disableCallGraphInlining",           "O\tdisable Interpreter Profiling based inlining and code size estimation",  SET_OPTION_BIT(TR_DisableCallGraphInlining), "P"},
    {"disableCatchBlockRemoval",           "O\tdisable catch block removal",                    TR::Options::disableOptimization, catchBlockRemoval, 0, "P"},
    {"disableCFGSimplification",           "O\tdisable Control Flow Graph simplification",      TR::Options::disableOptimization, CFGSimplification, 0, "P"},
--- a/compiler/z/codegen/OMRCodeGenerator.hpp
+++ b/compiler/z/codegen/OMRCodeGenerator.hpp
@@ -343,6 +343,20 @@ public:
    void insertInstructionPrefetchesForCalls(TR_BranchPreloadCallData * data);
    bool hasWarmCallsBeforeReturn();
 
+   /**
+    * \brief Tells the optimzers and codegen whether a load constant node should be rematerialized.
+    *
+    * \details Large constants should be materialized (constant node should be commoned up)
+    * because loading them as immediate values is expensive.
+    *
+    * A constant is large if it's outside of the range specified by the largest negative
+    * const and smallest positive const. And the ranges are hardware platform dependent.
+    */
+   bool materializesLargeConstants() { return true; }
+   bool shouldValueBeInACommonedNode(int64_t value);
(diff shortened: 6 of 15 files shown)


compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -343,6 +343,13 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    void insertInstructionPrefetchesForCalls(TR_BranchPreloadCallData * data);
    bool hasWarmCallsBeforeReturn();
 
+   // Large constants should be materialized because loading them are considered expensive.
+   // A constant is large if it's outside of the range specified by the largetst negative
+   // const and smallest positive const.
+   bool materializesLargeConstants() { return true;}
```
fjeremic: Could we Doxygen these comments?

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3525,7 +3611,11 @@ OMR::CodeGenerator::isMaterialized(TR::Node * node)
    }
 
 
-// determine if value fits in the immediate field (if any) of instructions that machine provides
+// Determine if value fits in the immediate field (if any) of instructions that machine provides.
+// A node with large constant can be materialized and left as commoned nodes.
+// Smaller constants can be uncommoned so that they re-materialize everytime when needed as a call
+// parameter. This query is platform specific as constant loading can be expensive on some platforms
+// but cheap on others, depending on the their magnitude.
```
fjeremic: Could we Doxygen these comments?

compiler/z/codegen/OMRCodeGenerator.hpp
NigelYiboYu: Done

compiler/codegen/OMRCodeGenerator.cpp
NigelYiboYu: Done

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,92 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+      {
+      traceMsg(self()->comp(), "Performing uncommon call constant nodes\n");
+      }
+
+   for (TR::TreeTop* tt = self()->comp()->getStartTree(); tt; tt = tt->getNextTreeTop())
+      {
+       TR::Node* node = tt->getNode();
+       if(node &&
+               node->getNumChildren() >= 1 &&
+               node->getFirstChild() &&
```
fjeremic: If a node has a child the child will always be non-null so this check is redundant. Similarly below with `paramNode`.

fjeremic:
@andrewcraik @ymanton would be good to get a review here as this is relevant to x86 and Power.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,92 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+                TR::Node* newConstNode = TR::Node::create(paramNode->getOpCodeValue(), 0);
+                newConstNode->setConstValue(paramNode->getConstValue());
+                callNode->setAndIncChild(i, newConstNode);
+                paramNode->decReferenceCount();
+
+                if(self()->comp()->getOption(TR_TraceCG))
+                   {
+                   traceMsg(self()->comp(),
+                            "Uncommon const node %X [n%dn]\n",
+                            paramNode, paramNode->getNodePoolIndex());
```
andrewcraik: by convention you want getNodeGlobalIndex not the NodePoolIndex which is internal and not generally logged/exposed.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,92 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+   {
+   if(self()->comp()->getOption(TR_TraceCG))
+      {
+      traceMsg(self()->comp(), "Performing uncommon call constant nodes\n");
+      }
+
+   for (TR::TreeTop* tt = self()->comp()->getStartTree(); tt; tt = tt->getNextTreeTop())
+      {
+       TR::Node* node = tt->getNode();
+       if(node &&
```
andrewcraik: I don't think a tt can ever have a null node can it? I don't think that means anything...

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,92 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+               node->getNumChildren() >= 1 &&
+               node->getFirstChild() &&
+               node->getFirstChild()->getOpCode().isCall())
```
andrewcraik: And here you really do mean isCall and not just opcodes that are some kind of call? They are not the same in TR since isCall has been used to mean has side-effects and there are some surprising things that say yes to isCall

andrewcraik:
I think a control option for this would not hurt. Eventually we would want something akin to lastOptSubIndex as well, but that is more a nice to have.

The first use of a call should be under a treetop or check - you can probably reduce compile-time by limiting to only inspecting those trees. You probably also want to have a checklist of nodes you have modified to save repeatedly processing the same call for some strange reason.

compiler/codegen/OMRCodeGenPhaseFunctionTable.hpp
```diff
@@ -26,6 +26,7 @@
    TR::CodeGenPhase::performReserveCodeCachePhase,                                           //ReserveCodeCachePhase
    TR::CodeGenPhase::performLowerTreesPhase,                                                 //LowerTreesPhase
    TR::CodeGenPhase::performSetupForInstructionSelectionPhase,                               //SetupForInstructionSelectionPhase
+   TR::CodeGenPhase::performUncommonCallConstNodesPhase,                                     //UncommonCallConstNodesPhase
```
ymanton: I think you want to do this before `SetupForInstructonSelection` has been done, since you're creating new nodes and you want them to go through the setup process.

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -505,6 +505,29 @@ OMR::CodeGenPhase::performLowerTreesPhase(TR::CodeGenerator * cg, TR::CodeGenPha
...
+   TR::LexicalMemProfiler mp(phase->getName(), comp->phaseMemProfiler());
+   LexicalTimer pt(phase->getName(), comp->phaseTimer());
+
+   cg->uncommonCallConstNodes();
+
+   if (comp->getOption(TR_TraceCG) || comp->getOption(TR_TraceTrees) || comp->getOptions()->getTraceCGOption(TR_TraceCGPreInstructionSelection))
+      comp->dumpMethodTrees("Post Uncommon Call Constant Node Trees");
+
+   // check interrupt
+   if (comp->compilationShouldBeInterrupted(AFTER_INSTRUCTION_SELECTION_CONTEXT))
```
ymanton: I don't know if you really need to check for interruption after what should be a fairly quick pass, but you'll be reporting the wrong yield point if you're interrupted. Personally I'd get rid of this, but if you want to keep it you may as well report the right thing.

NigelYiboYu: check interrupt is removed.

compiler/codegen/OMRCodeGenPhaseFunctionTable.hpp
NigelYiboYu: I just moved this phase before the `SetupForInstructonSelection` phase. 

As an aside, I do notice that there's a `RemoveUnusedLocalsPhase` stuck between the setup instruction selection and instruction selection phase ( in `trj9\codegen\CodeGenPhaseToPerform.hpp`). Is this intended?

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -505,6 +505,29 @@ OMR::CodeGenPhase::performLowerTreesPhase(TR::CodeGenerator * cg, TR::CodeGenPha
    }
 
 
+void
+OMR::CodeGenPhase::performUncommonCallConstNodesPhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
+   {
+   TR::Compilation* comp = cg->comp();
+   phase->reportPhase(InstructionSelectionPhase);
```
NigelYiboYu: Fixed

compiler/codegen/OMRCodeGenerator.cpp
NigelYiboYu: Done. I'll re-test after all these changes.

compiler/codegen/OMRCodeGenerator.cpp
NigelYiboYu: Changed to use `getGlobalIndex()`

compiler/codegen/OMRCodeGenPhaseFunctionTable.hpp
ymanton: `RemoveUnusedLocals` doesn't do anything to the trees or nodes, it just scans the list of autos and removes unreferenced symbols, so it's OK.

compiler/codegen/OMRCodeGenerator.cpp
NigelYiboYu: The first `node` in the if statement is removed.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,89 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+   if(self()->comp()->getOption(TR_TraceCG))
+      {
+      traceMsg(self()->comp(), "Performing uncommon call constant nodes\n");
+      }
+
+   for (TR::TreeTop* tt = self()->comp()->getStartTree(); tt; tt = tt->getNextTreeTop())
+      {
+       TR::Node* node = tt->getNode();
+       if(node->getNumChildren() >= 1 &&
+               node->getFirstChild()->getOpCode().isFunctionCall())
```
andrewcraik: This API does not yet exist as far as I can tell so this change is not ready for merging - please mark as WIP or revise the code prevent accidental merges

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +569,89 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+             if(paramNode->getReferenceCount() > 1 &&
+                     paramNode->getOpCode().isLoadConst() &&
+                     !self()->isMaterialized(paramNode))
+                {
+                TR::Node* newConstNode = TR::Node::create(paramNode->getOpCodeValue(), 0);
+                newConstNode->setConstValue(paramNode->getConstValue());
+                callNode->setAndIncChild(i, newConstNode);
+                paramNode->decReferenceCount();
+
+                if(self()->comp()->getOption(TR_TraceCG))
```
andrewcraik: Convention says the trace should be before the transformation so that the log contains the line of the last transformation attempted in the event of a crash rather than leaving no trace of diagnostic output of what was about to happen.

compiler/codegen/OMRCodeGenerator.cpp
andrewcraik: You probably want to have a list of processed nodes to save compile-time - a NodeChecklist to record nodes you have already transformed would be good so you don't keep re-examining the same node under multiple different treetops (not uncommon when OSR is enabled for example).

NigelYiboYu: 1. isFunctionCall() exists in OMR. I'm making a new one in trj9 to override the OMR one. This is a temporary solution until OMR has a new IL property and query.

2. I'm using NodeChecklist in my latest commit. Will push it up for you to review.

compiler/codegen/OMRCodeGenerator.cpp
NigelYiboYu: Fixed.

compiler/codegen/OMRCodeGenerator.cpp
andrewcraik: So the only concern here is that we don't uncommon things we shouldn't uncommon. isFunctionCall doesn't quite catch only true calls so we just need to make sure this subset is right.

0xdaryl:
What is the reason this rematerialization is being done as a codegen phase rather than as part of the rematerialization (uncommoning) local opt?  I'm not saying it shouldn't be, but just trying to understand why.

andrewcraik:
The logic here is that it is always legal to common nodes and that the common of constants is especially pervasive. Uncommoning nodes can lead to some optimizations not working as well since they rely on identity for a cheap determination of equivalence. The idea here is that we don't want the optimizer polluted with platform specific commonings which may change the idioms seen by many optimization and that it is cleaner conceptually to uncommon the nodes once all optimization has finished when it is appropriate and beneficial to do so. Basically, we want a cleaner separation between optimization and platform-specific deoptimization.

andrewcraik:
LGTM from me

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -616,8 +619,8 @@ class OMR_EXTENSIBLE CodeGenerator
    bool isMaterialized(TR::Node *); // no virt, cast
    bool isMaterialized(int64_t); // no virt, cast
    bool materializesLargeConstants() { return false;}
-   int32_t getLargestNegConstThatMustBeMaterialized() {return -32769;} // no virt, cast
-   int32_t getSmallestPosConstThatMustBeMaterialized() {return 32768;} // no virt, cast
+   int64_t getLargestNegConstThatMustBeMaterialized() {return -32769;}
+   int64_t getSmallestPosConstThatMustBeMaterialized() {return 32768;}
```
0xdaryl: To my point above, these default constants are completely arbitrary in the common code generator.

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -343,6 +343,19 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    void insertInstructionPrefetchesForCalls(TR_BranchPreloadCallData * data);
    bool hasWarmCallsBeforeReturn();
 
+   /**
+    * \brief Tells the optimzers and codegen whether a load constant node should be rematerialized.
+    *
+    * \details Large constants should be materialized (constant node should be commoned up)
+    * because loading them as immediate values is expensive.
+    *
+    * A constant is large if it's outside of the range specified by the largetst negative
```
0xdaryl: Typo: largest

compiler/control/OMROptions.cpp
```diff
@@ -254,6 +254,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableBlockVersioner",              "O\tdisable block versioner",                        SET_OPTION_BIT(TR_DisableBlockVersioner), "P"},
    {"disableBranchOnCount",               "O\tdisable branch on count instructions for s390",  SET_OPTION_BIT(TR_DisableBranchOnCount), "F"},
    {"disableBranchPreload",               "O\tdisable return branch preload",                  SET_OPTION_BIT(TR_DisableBranchPreload), "F"},
+   {"disableCallConstUncommon",           "O\tdisable uncommon call constant node phase",  SET_OPTION_BIT(TR_DisableCallConstUncommon), "F"},
```
0xdaryl: Would calling this `disableCallConstUncommoning` read better?

compiler/codegen/OMRCodeGenerator.hpp
NigelYiboYu: These are not arbitrary. The two default values here are max/min 16-bit signed numbers used by PPC. On Z, I have override functions that return max/min 32-bit signed numbers.

I'll leave a comment on these two.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+          node->getOpCodeValue() == TR::lcalli      ||
+          node->getOpCodeValue() == TR::lucalli     ||
+          node->getOpCodeValue() == TR::fcalli      ||
+          node->getOpCodeValue() == TR::dcalli      ||
+          node->getOpCodeValue() == TR::acalli      ||
+          node->getOpCodeValue() == TR::calli;
+   }
+
+/**
+ * \brief The uncommon-call-constant-node pass is an extra pre-codegen pass right
```
NigelYiboYu: I will delete the word `right`. This was true in my first implementation and is no longer so.

compiler/codegen/OMRCodeGenerator.hpp
andrewcraik: Wouldn't it make more sense to do the extension point as Darryl suggested above and basically say no by default and anyone who wants it opts in. That way PPC specifically specifies what range they want and won't be broken by changes at the common codegen layer? It would seem to be more robust.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
 
+/**
+ * TODO: Remove this function once OMR provides a proper API query function.
+ * OMR issue https://github.com/eclipse/omr/issues/842
+ *
+ *
+ * \brief The ILOpCode property `ILProp1::Call` is a property used in a broader sense.
+ * An opCode that bears this property has call-like side effects; and they
+ * may not be real function calls. This query is used to check if an OpCode is a
+ * functional call by excluding non-funciton-call opCodes.
```
NigelYiboYu: Fixed.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+/**
+ * \brief The uncommon-call-constant-node pass is an extra pre-codegen pass right
+ * before instruction selection. It aims to avoid the situation where
+ * a constant integer is made alive across calls and, thereby, lower the preserved register
+ * pressure caused due to this.
+ *
+ * \details
+ *
+ * \verbatim
+ * Take z/Architecture as an exmaple, we could run into the following situation where constant
```
NigelYiboYu: fixed

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+ *   bool materializesLargeConstants() { return false;}
+ *   int64_t getLargestNegConstThatMustBeMaterialized() {return -32769;}
+ *   int64_t getSmallestPosConstThatMustBeMaterialized() {return 32768;}
+ * \endverbatim
+ *
+*/
+void
+OMR::CodeGenerator::uncommonCallConstNodes()
+   {
+    TR::Compilation* comp = self()->comp();
```
NigelYiboYu: Fixed in the next commit.

compiler/z/codegen/OMRCodeGenerator.hpp
NigelYiboYu: Fixed in the next commit.

compiler/control/OMROptions.cpp
NigelYiboYu: Fixed in the next commit.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+ * ASSOCREGS
+ * Label L0041:
+ * LG      GPR14,#610 -216(GPR6)
+ * LGHI    GPR0,0xff28
+ * BASR    GPR14,GPR14         # Call "java/io/PrintStream.println(I)V"
+ *
+ * \endverbatim
+ *
+ * Loading integer constants as immediate values [i.e. LHI GPR2, 0x10] is preferred.
+ * The ranges for immedate value loading on different platforms are governed by the following queries:
```
NigelYiboYu: Will fix it in the next commit

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3525,7 +3659,17 @@ OMR::CodeGenerator::isMaterialized(TR::Node * node)
...
+ *
+ * \details
+ *
+ * A node with large constant can be materialized and left as commoned nodes.
+ * Smaller constants can be uncommoned so that they re-materialize everytime when needed as a call
+ * parameter. This query is platform specific as constant loading can be expensive on some platforms
+ * but cheap on others, depending on the their magnitude.
+ */
 bool
 OMR::CodeGenerator::isMaterialized(int64_t value)
```
NigelYiboYu: How about sinking the following three functions to the `omr/compiler/<platform>/codegen` folders so that they are not exposed in the common code generator layer?

`isMaterialized(int64_t value)`
`SmallestPosConstThatMustBeMaterialized()`
`LargestNegConstThatMustBeMaterialized()`

NigelYiboYu: How about `shouldValueBeInCommonedNode()` or `shouldValueBeMaterialized()` ?

fjeremic: Perhaps `shouldRematerializeConstant(int64_t value)`?

NigelYiboYu:
@0xdaryl 
Updated with new changes. Please review

NigelYiboYu:
@0xdaryl 

My apologies. I didn't see your test methodology request.

1. The functional verification is covered by JIT and VM personal builds. See the dependent J9 merge request for build details: https://gitlab-polyglot.hursley.ibm.com/jit/tr.open/merge_requests/1037

2. I've also sampled a few compiled Java method traces and verified that the constant nodes under function calls are indeed uncommoned; as a result, the J9 codegen is also emitting the intended load-immediate machine instructions.

3. Performance tests were done manually on the affected platforms: PowerPC and Z/Architecture using DayTrader3 benchmark. I collected 10-15 samples on each of the platforms and gained ~0.7% throughput on PowerPC and ~0% on Z.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
 
   }
 
+/**
+ * TODO: Remove this function once OMR provides a proper API query function.
+ * OMR issue https://github.com/eclipse/omr/issues/842
```
NigelYiboYu: This query is no longer used. After reviewing all ILOpCodes that has the `Call` property, the existing `isFunctionCall` query is the proper query to use.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -569,6 +570,139 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+ * before instruction selection. It aims to avoid the situation where
+ * a constant integer is made alive across calls and, thereby, lower the preserved register
+ * pressure caused due to this.
+ *
+ * \details
+ *
+ * \verbatim
+ * Take z/Architecture as an exmaple, we could run into the following situation where constant
+ * integer 16 is loaded to a preserved register GPR6, which is then loaded to GPR2 as call
+ * parameter. The GPR6 will be preserved by the callee and can incur extra register shuffling.
```
NigelYiboYu: The comment is shortened.

compiler/codegen/OMRCodeGenerator.hpp
NigelYiboYu: My latest commit basically does this: answer NO by default and answer yes in the P and Z layer. And for the platforms that answer yes to `isMaterialized`, they'll provide platform specific ranges using the `SmallestPosConstThatMustBeMaterialized` and `SmallestPosConstThatMustBeMaterialized` queries, which are not provided in the common codegen layer.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -564,6 +565,91 @@ OMR::CodeGenerator::setUpForInstructionSelection()
...
+      traceMsg(comp, "Performing uncommon call constant nodes\n");
+      }
+
+   TR::NodeChecklist checklist(comp);
+
+   for (TR::TreeTop* tt = self()->comp()->getStartTree(); tt; tt = tt->getNextTreeTop())
+      {
+       TR::Node* node = tt->getNode();
+       if(node->getNumChildren() >= 1 &&
+               node->getOpCode().isFunctionCall())
```
vijaysun-omr: Should'nt this be testing node->getFirstChild()->getOpCode().isFunctionCall() ?

NigelYiboYu: fixed and re-testing

vijaysun-omr:
I see that the typo in the previous commit is now fixed. I will merge this now since the checks have passed.

PR #839, opened 2017-03-07
Improve JitBuilder opt strategy

mstoodle:
Revamping especially the initial optimizations in the JitBuilder optimization
strategy can more quickly clean up the initial IL generated. I also added some
more global optimizations into the jitBuilder strategy (global copy propagation,
global value propagation) to improve the final code quality.

Need to iterate on this more, but for now most consumers I tested this with
got a performance improvement from this new strategy, some just because the IL
didn't blow the thresholds to start disabling things (like use-def analysis).

### DIFF
--- a/jitbuilder/optimizer/JBOptimizer.cpp
+++ b/jitbuilder/optimizer/JBOptimizer.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
Don't see anything risky wrt the ordering of optimizations that might have cause functional issues. 
Performance tradeoff in terms of compile time vs throughput here is something I am curious about but from your commit message, it seemed to help most consumers. I am merging so that others waiting to consume this can proceed.

PR #847, opened 2017-03-10
Incremental cleanup toward removing warm snippets

0xdaryl:
* Remove references to last warm block/instruction
* Remove optional `isWarm` parm from snippet generation
* Remove `isWarm` parm from snippet printing

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -650,17 +650,6 @@ OMR::CodeGenerator::doInstructionSelection()
             }
 #endif
          }
-      else if (opCode == TR::BBEnd)
-         {
-         TR::Block *b = self()->getCurrentEvaluationBlock();
-         if (b->isLastWarmBlock())
-            {
-            // Mark the split point between warm and cold instructions, so they
-            // can be allocated in different code sections.
-            //
-            prevInstr->setLastWarmInstruction(true);
-            }
-         }
 
       self()->setLiveLocals(liveLocals);
 
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -2891,7 +2891,7 @@ OMR::CodeGenerator::alignBinaryBufferCursor()
 
 
 int32_t
-OMR::CodeGenerator::setEstimatedLocationsForSnippetLabels(int32_t estimatedSnippetStart, bool isWarm)
+OMR::CodeGenerator::setEstimatedLocationsForSnippetLabels(int32_t estimatedSnippetStart)
    {
    TR::Snippet *cursor;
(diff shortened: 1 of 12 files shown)


0xdaryl:
@fjeremic , @ymanton : I'd appreciate a review please.

0xdaryl:
**Note To Reviewers**

Feel free to poke holes in my logic.

Removing the notion of "last warm block" and "last warm instruction" is predicated on eliminating the following behaviour:

* The notion of "last warm instruction" is only set by code guarded by `isLastWarmBlock()`

* `isLastWarmBlock()` never returns true in OMR because nothing sets the last warm block.
Looking into all known downstream projects, only the J9 Z code generator has the possibility 
of setting it.  Therefore, neither of these settings have any use for ARM, Power, and X86.

* For Z, even though J9 might set the last warm block and therefore the last warm
instruction, there is no code in either OMR or any known downstream project that consumes
`isLastWarmInstruction` on Z.

Therefore, both will be eliminated.  Once disconnected in OMR, downstream projects
will be modified to not use any of the functions, and then they will be eliminated
from OMR.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -2902,20 +2891,20 @@ OMR::CodeGenerator::alignBinaryBufferCursor()
...
    if (self()->hasTargetAddressSnippets())
       {
-      estimatedSnippetStart = self()->setEstimatedLocationsForTargetAddressSnippetLabels(estimatedSnippetStart, isWarm);
+      estimatedSnippetStart = self()->setEstimatedLocationsForTargetAddressSnippetLabels(estimatedSnippetStart);
       }
 
    for (auto iterator = _snippetList.begin(); iterator != _snippetList.end(); ++iterator)
       {
-      if ((*iterator)->isWarmSnippet() == isWarm)
+      if ((*iterator)->isWarmSnippet() == 0)
```
fjeremic: Assuming `isWarmSnippet` function returns a `bool` can this be changed to:

```
if (!(*iterator)->isWarmSnippet())
```

to avoid comparison between `bool` and `int`. Similarly elsewhere in this change.

0xdaryl: Yes, but the very next PR to be submitted after this one is merged will remove all `isWarmSnippet()` calls assume they are always false (which they are).  See https://github.com/0xdaryl/omr/tree/nowarmsnippets for a sneak peak.  It has passed internal testing; I'm just waiting for this PR to be merged.

fjeremic:
LGTM. My concerns have been addressed. Very good explanation of the background for this change @0xdaryl! Readers should take note :)

vijaysun-omr:
@ymanton have you reviewed this change ?

ymanton:
LGTM.

PR #849, opened 2017-03-10
Fix Simplifier::create method

xliang6:
Change Simplifier::create method to instantiate TR::Simplifier so the
right simplifier will be created for a given project.


xliang6:
@0xdaryl Done.

PR #850, opened 2017-03-10
Cleanup logics related to CPU feature detection

0dvictor:
1. Eliminated the two almost identical unnecessary assembly helpers for CPUID
2. Use supportsTM() in a way which is consistent with other checks.


vijaysun-omr:
@andrewcraik and @0xdaryl should review this.

andrewcraik:
@cathyzhyi a review from you due to the change to the TM handling

andrewcraik:
This looks like a generally good change which cleans up a lot of handcoded assembly and seems conceptually sound. I defer to @0xdaryl for a more detailed review since he may be more familiar with subtleties in the old implementation that we might care about as well as @cathyzhyi since she knows more about TM than I do.

cathyzhyi:
The TM part looks good to me. There were 2 ways to check supporting TM and was very confusing. Now it's much clearer.

andrewcraik:
@0xdaryl over to you LGTM from me

0xdaryl:
Thanks for this PR.  This is an improvement over what's already been implemented in the compiler
using its custom assembly routines.  

However, ultimately, the compiler technology should not be responsible for determining the target
processor to which its generating code for.  That is really what the 'target' of the `Compiler` `Env` pointer is for--its a step in that direction.  The code generator shouldn't care where that 'target' came from, only that the 'target' is what it should be asking for processor capabilities.  For example, the 'target' could come from the runtime which is filled in by the port library, or it could be a completely different target processor than the host in the world of cross compilation (we're not there yet with this technology, but we're getting there).

So, for this PR, I think the code to eliminate the assembly code can stay.  But I would like to see the environment queried for whether the processor supports TM or not (i.e., leave it the way it was).  I know its different than the other queries, but arguably those other queries should actually be converted to using the 'target' rather than the other way around.

More of the code generator processor queries should be based on the Env pointer, but that's probably for another PR. All of the CPU detection code should live in the port library.  There was/is some work being done to achieve that, but its progress is slow.

andrewcraik:
@0xdaryl until the port library layer changes are done I think we do still need a PR like this to clean-up the handling within the x code generator. As I see it there are 3 steps 1) clean-up what is there, 2) move queries over to Env and 3) move implementation to the port library. The goal of this PR is 1 and as long as queries are not moved out of env, I don't see why it should be blocked since it is cleaning up some very ugly code and making future refactorings easier. We need access to the entire set of CPU bits which is what this is trying to achieve.

0xdaryl:
My long-winded review agrees with you.  The only part that I object to is removing the query from the the CPUEnv class which is ultimately the question that should be asked in the future.  I'd hate for this code to be removed and then added back later.

andrewcraik:
Ok awesome - think I must have misunderstood something. I'm sure @0dvictor will be happy to make that change to the PR and re-request a review. Thanks!

0dvictor:
Sure, we can keep checks for TM in CPUEnv class. It has been updated.

andrewcraik:
LGTM from me - over to @0xdaryl for reconsideration.

PR #851, opened 2017-03-11
Move forwarded object bit from 0x2 to 0x4

amicic:
This will allow to set forwarded bit and hole bit (0x1) at the same
(which will be needed for a special variant of forwarded object in
future) and be able to distinguish it from a single slot hole (0x3 bits
set).

### DIFF
--- a/gc/structs/ForwardedHeader.hpp
+++ b/gc/structs/ForwardedHeader.hpp
@@ -68,7 +68,7 @@ private:
 	omrobjectptr_t _objectPtr;					/**< the object on which to act */
 	MutableHeaderFields _preserved; 			/**< a backup copy of the header fields which may be modified by this class */
 	const uintptr_t _forwardingSlotOffset;		/**< fomrobject_t offset from _objectPtr to fomrobject_t slot that will hold the forwarding pointer */
-	static const uintptr_t _forwardedTag = 2;	/**< bit mask used to mark forwarding slot value as forwarding pointer */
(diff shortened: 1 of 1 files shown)


fjeremic:
I suspect this change will require updates to KCA as it analyzes forwarded objects as potential GC map bugs. Do we need updates to DDR?

amicic:
This change alone might not be an issue, but accompanied change specific to J9 will almost sure break it.

klangman:
Currently KCA does look for GC dark matter, moved objects and GC free list pointers using the following logic:

Dark matter:  ptr sized 0x1 followed by a ptr sized length with a value of 512 (or less) followed by valid object or another ptr sized 0x1 followed by a ptr sized length of 512 or less.

Moved Object:  a low bit tagged (0x2) pointer to a valid object

GC Free list: At least 5 nodes in a linked list where the 0th element of each node is a low tagged (0x1) next ptr, or if we hit a free list terminator (ptr sized value of 0x1).

charliegracie:
KCA will need to be update to handle this.  I am not sure if it has support for different values for different VM versions.  Eventually KCA should be updated to use the DDR generated types / structs and fetch them from the core file so it will continue to work with these types of changes without any updates.

I do not believe DDR needs to be changed for this.

klangman:
KCA already can read all the DDR struct data.. What struct/field/const should I be looking at to determine the bits that are used?

PR #853, opened 2017-03-13
Make omrmmap_map_file honor offset on z/OS

dnakamura:
Currently offset parameter is ignored on z/OS, leading to unexepected
behaviour when offset != file position


charliegracie:
Looks good to me.  Thanks for catching this.

mgaudet:
Cursory inspection of `fvtest/porttest/omrmmapTest.cpp` suggests that there is no testing of the offset parameter of `omrmmap_map_file`. Perhaps a follow up item for tests ought to be opened?

charliegracie:
I agree there should be testing added for this in a follow up item.

PR #854, opened 2017-03-13
Prefer fully qualified namespace for definitions

mgaudet:
Makes it much easier to search for function names. This is already the
dominant style in the code base.

[ci skip] as documentation only change.

### DIFF
--- a/doc/CodingStandard.md
+++ b/doc/CodingStandard.md
@@ -1441,3 +1441,21 @@ uint32_t size = getSize();
 uint32_t averageSize = size / count;
 ```
 
+## Specify Namespace in Function Definition
+
+When writing the definition of a C++ function separate from the declaration,
+use the fully qualified name to allow easier search.
+
+Bad
+```
+namespace OMR { 
+
+Foo::Func(int*) ... 
+
+} 
+```
+
+Good 
+```
+OMR::Foo::Func(int*) 
+```


andrewcraik:
An excellent addition to the standard - definitely a best practice for everyone's sanity in my opinion.

charliegracie:
@rwy0717 and @youngar is this a good first step in the direction you were thinking for namespace?

rwy0717:
Unambiguous code is important. My advice is to eschew any but the most basic/predictable use of namespaces and classes, and prefer unique base names instead. You can’t use the fully qualified name everywhere, after all.

As an example:

Good:
```C++
namespace OMR {
  namespace GC {
    class SegregatedMarkingScheme;
    class MarkingScheme;
  } // GC
} // OMR
```

Bad:
```C++
namespace OMR {
  namespace GC {
    namespace Segregated {
      class MarkingScheme;
    } // Segregated
    class MarkingScheme;
  } // GC
} // OMR
```

When we follow these rules, is the fully-qualified-name still helpful, or just unnecessarily verbose? But hey, I don’t know, clarity is a big deal.

mgaudet:
*cough* 

That last example is case in point why we need / want fully qualified names. In the compiler component we've structured core classes into layers, where the distinction between layers is _only_ the namespace.

rwy0717:
Oh in that case, I completely agree 😸

mgaudet:
@0xdaryl  Corrected.

PR #855, opened 2017-03-13
Split block containing the call under FSD mode

cathyzhyi:
In FSD, we need to split block even for cases without virtual
guard. This is because under FSD a block with OSR point must have an
exception edge to the osrCatchBlock of correct callerIndex. Split
the block here so that the OSR points from callee and from caller
are separated.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -5011,6 +5011,18 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
             _disableTailRecursion = true;
          }
       }
+   else if (comp()->getOption(TR_FullSpeedDebug) && tif->crossedBasicBlock())
+      {
+      /**
+       * In FSD mode, we need to split block even for cases without virtual guard. This is 
+       * because in FSD a block with OSR point must have an exception edge to the osrCatchBlock
+       * of correct callerIndex. Split the block here so that the OSR points from callee
+       * and from caller are separated.
+       */
+      TR::Block * blockOfCaller = previousBBStartInCaller->getNode()->getBlock();
+      TR::Block * blockOfCallerInCalleeCFG = nextBBEndInCaller->getNode()->getBlock()->split(callNodeTreeTop, callerCFG);
+      callerCFG->copyExceptionSuccessors(blockOfCaller, blockOfCallerInCalleeCFG);
+      }
 
    // move the NULLCHK to before the inlined code
    //


vijaysun-omr:
Do you need any special handling for callees that are so simple that they just get inlined into the block containing the call originally without any splitting of blocks at all ? 

Perhaps a simple diagram or elaboration of what the picture was before this change compared with what the picture is with this change would help make it completely clear.

andrewcraik:
So the issue here is what happens to the OSR exception edges. If the caller's block containing the call to be inlined has an OSRExceptionEdge to OSRCatchBlock A and the callee block has OSRExceptionEdge to OSRCatchBlock B, merging the two blocks together causes the block to have two OSRExceptionEdges which is a confusing state - especially if the callee can be simplified to a non-treetop. The thrust of this change is to make sure that we do not merge the block being inlined so that the OSR exception edges remain correct and each block has one OSR catch.

In the case of a multi-block method which does not need a guard, the caller block will be fused with the callee entry block and the callee exit block will have appended to it the treetops from the caller following the call. This leads to some very confusing looking bytecode info on the blocks and the fused blocks needed two different OSR exception edges to two different OSR catch blocks.

Handling these fused blocks with multiple exception edges is not done by the current OSR infrastructure and rather than complicating the already complicated OSR book keeping even more trying to keep straight information about fused blocks, keeping the blocks separate achieves the same result with less complexity.

vijaysun-omr:
Okay that clarifies things. Thanks. 

It's those blocks at the transitions from caller->callee bytecodes (and vice versa) that are affected by the change and I agree that we should probably avoid having multiple OSR catch blocks in those cases since the rest of the OSR infrastructure might not work correctly.

cathyzhyi:
Add special handling for cases where the callee is a single block method. In such case, there is no need to split the block because a single block callee in FSD mode mustn't have any OSR point nor OSR block.

andrewcraik:
New version looks clean - I think this should be good to go - LGTM.

compiler/optimizer/Inliner.cpp
```diff
@@ -5011,6 +5011,18 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
             _disableTailRecursion = true;
          }
       }
+   else if (comp()->getOption(TR_FullSpeedDebug) && tif->crossedBasicBlock())
```
vijaysun-omr: Does'nt this need to be done whenever "involuntary OSR" is needed ? 
I agree that FSD is the only case in which we do involuntary OSR today but wanted to ask if a more general check here would be better so that this code "just works" in some future (not FSD) scenario where involuntary OSR would be used.

andrewcraik: @ncough is working on just such a query now. He should be making his pull request soon so maybe he would be kind enough to sweep this test for FSD up into his new API when he asks to put it in? would that work for you @vijaysun-omr?

vijaysun-omr: Yes I am fine with @ncough doing it as part of his PR. Otherwise this change is good and I will merge it now.

PR #856, opened 2017-03-13
Rewrite post- to pre-increment in loop tests

jdmpapin:
Canonicalizer now looks within loops for inequality tests against the
old value of a variable that is incremented or decremented just prior to
the test, and attempts to rewrite those tests in terms of the new value.
As an example in pseudocode, it attempts to change `if (i++ < n)` into
`if (++i <= n)`.

Particularly for loop tests against induction variables, the optimizer
is better attuned to tests against the new value, because those are
commonly generated for typical for-loops (after canonicalization).

### DIFF
--- a/compiler/il/OMRILOps.hpp
+++ b/compiler/il/OMRILOps.hpp
@@ -83,7 +83,7 @@ public:
 
    /// Get the opcode to be used if the children of this opcode are swapped.
    /// e.g. ificmplt --> ificmpgt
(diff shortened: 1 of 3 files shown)


vijaysun-omr:
Looks good to me. I would appreciate a look by @andrewcraik as well; there is a fair amount of intricate logic for the several different cases being handled by the same piece of code and I reviewed it myself but would appreciate another set of eyes over the code before merging.

andrewcraik:
LGTM

PR #859, opened 2017-03-14
WIP: Simplify integer add/sub under comparison nodes

NigelYiboYu:
Simplify integer add and subtract operations under integer comparison
nodes to avoid emitting add/sub instructions in the codegen phase.

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
    return node;
    }
 
+/**
+ * \brief
+ * Simplify add/subtract operations under integer comparison nodes
+ * to save an add/sub instruction during codegen.
+ *
+ * \details
+ *
+ * e.g. This following is a before-after comparison.
+ *
+ * Before:
+ * ificmple --> block_17 BBStart at n13n ()
+ *   iadd (X>=0 cannotOverflow )
+ *     ==>iRegLoad (in GPR_0112) (cannotOverflow canBeAssignedToHPR SeenRealReference )
+ *     iconst -1 (X!=0 X<=0 )
+ *   ==>iconst 0 (X==0 X>=0 X<=0 )
+ *
+ *
+ * After:
+ * ificmple --> block_17 BBStart at n13n ()
+ *   ==>iRegLoad (in GPR_0112) (cannotOverflow canBeAssignedToHPR SeenRealReference )
+ *   iconst 1 (X==0 X>=0 X<=0 )
+ *
+ *
+ * The iadd node needs to be single reference and marked as 'cannotOverflow' to
+ * ensure functional correctness.
+*/
+TR::Node* reduceIntegerCompareArithmetics(TR::Node* node, TR::Simplifier * s)
+   {
+   TR::Node* firstChild = node->getFirstChild();
+   TR::Node* secondChild = node->getSecondChild();
+   TR::ILOpCodes op = firstChild->getOpCodeValue();
+
+   if((op == TR::iadd || op == TR::isub) &&
+           firstChild->chkCannotOverflow()  &&
+           firstChild->isSingleRef() &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
+           firstChild->getSecondChild()->getConstValue() != 0 &&
+           secondChild->getOpCode().isLoadConst())
+      {
+      int64_t const1 = firstChild->getSecondChild()->getConstValue();
+      int64_t const2 = secondChild->getConstValue();
+      int64_t const3 = 0;
+
+      if(op == TR::iadd)
+         {
+         const3 = const2 - const1;
+         }
+      else if(op == TR::isub)
+         {
+         const3 = const2 + const1;
+         }
+
+      if(const3 > ((int64_t)0x000000007FFFFFFF) ||
+              const3 < ((-1ll) << 31))
+         {
+         // Can't set iconst value if const3 is outside of int32_t range.
+         return node;
+         }
+
+      dumpOptDetails(s->comp(), "%sEliminating add/sub operation under node n%dn\n",
+                     s->optDetailString(),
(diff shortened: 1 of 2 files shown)


compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
+ * The iadd node needs to be single reference and marked as 'cannotOverflow' to
+ * ensure functional correctness.
+*/
+TR::Node* reduceIntegerCompareArithmetics(TR::Node* node, TR::Simplifier * s)
+   {
+   TR::Node* firstChild = node->getFirstChild();
+   TR::Node* secondChild = node->getSecondChild();
+   TR::ILOpCodes op = firstChild->getOpCodeValue();
+
+   if((op == TR::iadd || op == TR::isub) &&
```
fjeremic: Nitpick: Space between `if` and parenthesis. Similarly in other palaces in this commit.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
+           firstChild->chkCannotOverflow()  &&
+           firstChild->isSingleRef() &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
+           firstChild->getSecondChild()->getConstValue() != 0 &&
```
fjeremic: Why is this check necessary?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
+         {
+         const3 = const2 - const1;
+         }
+      else if(op == TR::isub)
+         {
+         const3 = const2 + const1;
+         }
+
+      if(const3 > ((int64_t)0x000000007FFFFFFF) ||
+              const3 < ((-1ll) << 31))
```
fjeremic: This seems cryptic to read. How about:

```
if ((static_cast<uint64_t>(const3) >> 32) > 0)
```

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
+ *
+ *
+ * After:
+ * ificmple --> block_17 BBStart at n13n ()
+ *   ==>iRegLoad (in GPR_0112) (cannotOverflow canBeAssignedToHPR SeenRealReference )
+ *   iconst 1 (X==0 X>=0 X<=0 )
+ *
+ *
+ * The iadd node needs to be single reference and marked as 'cannotOverflow' to
+ * ensure functional correctness.
```
ymanton: Is the single reference limitation really necessary if the non-constant child of the `iadd` is still being evaluated in the original location? I don't see what can go wrong if the `iadd` is evaluated later in this case.

Also, why is this limited to integer types? Is there a problem with doing this for longs?

compiler/optimizer/OMRSimplifierHandlers.cpp
NigelYiboYu: Will remove this `const!=0` check later.

compiler/optimizer/OMRSimplifierHandlers.cpp
NigelYiboYu: One of my old habits. I will add spaces in my next commit.

compiler/optimizer/OMRSimplifierHandlers.cpp
andrewcraik: Or shorts or bytes for that matter?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12882,6 +12882,88 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
+ * \brief
+ * Simplify add/subtract operations under integer comparison nodes
+ * to save an add/sub instruction during codegen.
+ *
+ * \details
+ *
+ * e.g. This following is a before-after comparison.
+ *
+ * Before:
+ * ificmple --> block_17 BBStart at n13n ()
```
andrewcraik: This example looks to have too much detail like the node numbers and block numbers - I thik abstracting those details out to make it more clear what the reader should be focusing on makes more sense.

compiler/optimizer/OMRSimplifierHandlers.hpp
```diff
@@ -267,5 +267,5 @@ TR::Node * NewSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
 TR::Node * lowerTreeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s);
 TR::Node * arrayLengthSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s);
 
-
+TR::Node * reduceIntegerCompareArithmetics(TR::Node* node,TR::Simplifier * s);
```
andrewcraik: Not a big fan of the name - this almost seems like a form of comparison strength reduction (see https://en.wikipedia.org/wiki/Strength_reduction), but I'm not sure it is truly strength reduction. A better, more descriptive name is certainly warranted.

compiler/optimizer/OMRSimplifierHandlers.cpp
NigelYiboYu: I'm currently working on signed/unsigned long, int, byte, short, and etc. I'll test and update this MR.

compiler/optimizer/OMRSimplifierHandlers.hpp
NigelYiboYu: Noted. I'll give it a better name.

PR #860, opened 2017-03-14
Propagate through relative global constraints in VP

jdmpapin:
When adding a block constraint, VP (value propagation) may attempt to
propagate it through any relative constraints on the same value number.
For this purpose VP now looks at relative global constraints in addition
to block constraints, which helps it to constrain the value of an
induction variable based on a comparison against its previous value.


vijaysun-omr:
Looks like a safe change to make us more consistent with respect to relative constraints. Merging.

PR #861, opened 2017-03-14
OMR codegen full support for certain evaluators

0dvictor:
integerPairMulEvaluator
integerPairShlEvaluator
integerPairRolEvaluator
integerPairShrEvaluator
integerPairUshrEvaluator

For integerPairMulEvaluator, it does not handle the case
with non-constant multiplicand when TR_NoInlineLongMultiply
is set. The env var is purely for backing off to deprecated
code. Hence the path is removed.

For the shift evaluators, the whole code paths handling
non-constant shift amount are missing. Therefore, this commit
implements them.


0xdaryl:
I haven't reviewed this in detail (yet), but in your commit message can you elaborate on what you changed to give these evaluators "full support" ?  Presumably, code on some paths were missing.  Can you try and generalize a description of those changes?

0dvictor:
For integerPairMulEvaluator, it does not handle the case with non-constant multiplicand when TR_NoInlineLongMultiply is set. The env var is purely for backing off to long deprecated code. Hence the path is removed.

mgaudet:
(Pretty good summaries: Can you put all that into the commit message?)

andrewcraik:
Discussion of the testing performed is needed for these changes.

0dvictor:
Both OMR testjit and J9 sanity tests passed and showed the correct instruction sequence. In addition, it passed all internal tests.

andrewcraik:
Looks like another good change - @0xdaryl over to you - I'd appreciate a second opinion on the code as relates to 32bit since I'm less familiar with the 32bit platform and I think you know it a bit better.

0dvictor:
I have also enabled ShiftOrRolTest for long type as this changeset introduces support for these evaluators.

PR #862, opened 2017-03-14
Introduce SIMD evaluators on X86

0dvictor:
Initial supported evaluators:
    vadd
    vmul
    vsplats
    vload
    vloadi
    vstore
    vstorei
    vbRegLoad
    vsRegLoad
    viRegLoad
    vlRegLoad
    vfRegLoad
    vdRegLoad
    vbRegStore
    vsRegStore
    viRegStore
    vlRegStore
    vfRegStore
    vdRegStore

Issue #997 

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -5262,3 +5262,64 @@ OMR::X86::TreeEvaluator::ibyteswapEvaluator(TR::Node *node, TR::CodeGenerator *c
    cg->decReferenceCount(child);
    return target;
    }
+
+enum BinaryArithmeticOps : uint32_t
+   {
+   BinaryArithmeticInvalid,
+   BinaryArithmeticAdd,
+   BinaryArithmeticSub,
+   BinaryArithmeticMul,
+   BinaryArithmeticDiv,
+   NumBinaryArithmeticOps
+   };
+static const TR_X86OpCodes BinaryArithmeticOpCodes[TR::NumOMRTypes][NumBinaryArithmeticOps] =
+   {
+   //  Invalid,       Add,         Sub,       Mul,         Div
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // NoType
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Int8
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Int16
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Int32
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Int64
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Float
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Double
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Address
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt8
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt16
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt32
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt64
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorFloat
+   { BADIA32Op, ADDPDRegReg, BADIA32Op, MULPDRegReg, BADIA32Op }, // VectorDouble
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Aggregate
+   };
+// For ILOpCode that can be translated to single SSE/AVX instructions
+TR::Register* OMR::X86::TreeEvaluator::FloatingPointAndVectorBinaryArithmeticEvaluator(TR::Node* node, TR::CodeGenerator* cg)
+   {
+   auto arithmetic = BinaryArithmeticInvalid;
+   switch (node->getOpCodeValue())
+      {
+      case TR::vadd:
+         arithmetic = BinaryArithmeticAdd;
+         break;
+      case TR::vmul:
+         arithmetic = BinaryArithmeticMul;
+         break;
+      default:
+         TR_ASSERT(false, "Unsupported OpCode");
+      }
+   TR::Node* operandNode0 = node->getChild(0);
+   TR::Node* operandNode1 = node->getChild(1);
+   TR::Register* operandReg0 = cg->evaluate(operandNode0);
+   TR::Register* operandReg1 = cg->evaluate(operandNode1);
+
+   TR::Register* resultReg = cg->allocateRegister(operandReg0->getKind());
+   generateRegRegInstruction(MOVDQURegReg, node, resultReg, operandReg0, cg);
+
+   TR_X86OpCodes opCode = BinaryArithmeticOpCodes[node->getDataType()][arithmetic];
+   TR_ASSERT(opCode != BADIA32Op, "FloatingPointAndVectorBinaryArithmeticEvaluator: unsupported data type or arithmetic.");
+   generateRegRegInstruction(opCode, node, resultReg, operandReg1, cg);
+
+   node->setRegister(resultReg);
+   cg->decReferenceCount(operandNode0);
+   cg->decReferenceCount(operandNode1);
+   return resultReg;
+   }
--- a/compiler/x/codegen/OMRTreeEvaluator.hpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.hpp
@@ -287,6 +287,16 @@ class OMR_EXTENSIBLE TreeEvaluator: public OMR::TreeEvaluator
    // routines for floating point values that can fit in one GPR
    static TR::Register *floatingPointStoreEvaluator(TR::Node *node, TR::CodeGenerator *cg);
 
+   // For ILOpCode that can be translated to single SSE/AVX instructions
+   static TR::Register *FloatingPointAndVectorBinaryArithmeticEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+
+   // SIMD evaluators
+   static TR::Register *SIMDRegLoadEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+   static TR::Register *SIMDRegStoreEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+   static TR::Register *SIMDloadEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+   static TR::Register *SIMDstoreEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+   static TR::Register *SIMDsplatsEvaluator(TR::Node *node, TR::CodeGenerator *cg);
+
    static TR::Register *icmpsetEvaluator(TR::Node *node, TR::CodeGenerator *cg);
    static TR::Register *bztestnsetEvaluator(TR::Node *node, TR::CodeGenerator *cg);
 
--- a/compiler/x/codegen/SIMDTreeEvaluator.cpp
+++ b/compiler/x/codegen/SIMDTreeEvaluator.cpp
@@ -1,0 +1,153 @@
+/*******************************************************************************
+*
+* (c) Copyright IBM Corp. 2017, 2017
+*
+*  This program and the accompanying materials are made available
+*  under the terms of the Eclipse Public License v1.0 and
+*  Apache License v2.0 which accompanies this distribution.
+*
+*      The Eclipse Public License is available at
+*      http://www.eclipse.org/legal/epl-v10.html
+*
+*      The Apache License v2.0 is available at
+*      http://www.opensource.org/licenses/apache2.0.php
+*
+* Contributors:
+*    Multiple authors (IBM Corp.) - initial implementation and documentation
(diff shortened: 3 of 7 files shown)


vijaysun-omr:
This should be reviewed by @0xdaryl and @andrewcraik

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+enum BinaryArithmetic_t : uint32_t
+   {
+   BinaryArithmeticInvalid,
+   BinaryArithmeticAdd,
+   BinaryArithmeticSub,
+   BinaryArithmeticMul,
+   BinaryArithmeticDiv,
+   BinaryArithmeticNum
+   };
+static const TR_X86OpCodes BinaryArithmeticOpCodes[TR::NumOMRTypes][BinaryArithmeticNum] =
```
andrewcraik: This seems very generic for a SIMD pull request and deserves some discussion/comment in the PR. What is the design here and why are we fusing the code paths for SIMD and (seemingly eventually) non-SIMD types?

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt64
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorFloat
+   { BADIA32Op, ADDPDRegReg, BADIA32Op, MULPDRegReg, BADIA32Op }, // VectorDouble
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Aggregate
+   };
+
+TR::Register* OMR::X86::TreeEvaluator::BinaryArithmeticEvaluator(TR::Node* node, TR::CodeGenerator* cg)
+   {
+   BinaryArithmetic_t arithmetic = BinaryArithmeticInvalid;
+   switch (node->getOpCodeValue())
```
andrewcraik: The above table seems very generic and yet this seems very specific despite the name of the function - this seems confusing. At the very least a comment on this function is warranted.

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+      cg->setImplicitExceptionPoint(instr);
+   return NULL;
+   }
+
+TR::Register* OMR::X86::TreeEvaluator::SIMDsplatsEvaluator(TR::Node* node, TR::CodeGenerator* cg)
+   {
+   TR::Node* childNode = node->getChild(0);
+   TR::Register* childReg = cg->evaluate(childNode);
+
+   uint8_t shufconst = 0;
```
andrewcraik: This deserves a comment and what is the hardcoded 0x44 - it seems a bit of a magic number - should it be in a constant or something with a name to make it more sensible.

andrewcraik:
A discussion of the testing done would be warranted for these changes - how do we know the new evaluators work?

compiler/x/codegen/OMRTreeEvaluator.hpp
```diff
@@ -286,6 +286,15 @@ class OMR_EXTENSIBLE TreeEvaluator: public OMR::TreeEvaluator
    // routines for floating point values that can fit in one GPR
    static TR::Register *floatingPointStoreEvaluator(TR::Node *node, TR::CodeGenerator *cg);
 
+   // SIMD evaluators
+   static TR::Register *BinaryArithmeticEvaluator(TR::Node *node, TR::CodeGenerator *cg);
```
andrewcraik: The code around this makes it look more generic than just SIMD as does the name - is it really intended for SIMD? If so it should have SIMD in its name no?

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+ * US Government Users Restricted Rights - Use, duplication or disclosure
+ * restricted by GSA ADP Schedule Contract with IBM Corp.
+ *
+ * This file is part of product 5655-W32.
+ *
+ */
+
+#include <stdint.h>                                   // for uint8_t, etc
+#include <stdio.h>                                    // for NULL, printf, etc
+#include "codegen/CodeGenerator.hpp"                  // for CodeGenerator, etc
```
andrewcraik: I'm not sure where this list of includes was copied from but it probably deserves evaluation for trimming since I don't think all of these entries are actually used in this file and so the compiler is working harder than needed and the code dependencies are less clear than they could be with at trimmed include list.

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+   TR_X86OpCodes opCode = BADIA32Op;
+   switch (node->getSize())
+      {
+      case 16:
+         opCode = MOVDQURegMem;
+         break;
+      default:
+         break;
+      }
+   TR_ASSERT(opCode != BADIA32Op, "SIMD Load Evaluator: unsupported fill size");
```
andrewcraik: A CG trace statement is warranted in these cases too incase someone see a problem but does not have asserts enabled...

compiler/x/codegen/SIMDTreeEvaluator.cpp
0dvictor: Yes, it is designed to be generic although now is only used by vadd and vmul. BinaryArithmeticEvaluator is for ILOpCode which can be translated to single SSE/AVX instructions. It has the ability to support not only SIMD but also non-SIMD types; although whether non-SIMD types should migrate to use this evaluator is likely to be a separate issue.

0dvictor: Fusing the code path also provides future benefits if we want to utilize three-operand AVX instructions. Instead duplicating almost identical code, this evaluator can handle almost, if not all, logic relating to three-operand AVX instructions.

compiler/x/codegen/SIMDTreeEvaluator.cpp
0dvictor: Yes, as its name indicates, it is a generic evaluator for binary arithmetic. Comment has been added.

compiler/x/codegen/SIMDTreeEvaluator.cpp
0dvictor: Updated.

compiler/x/codegen/OMRTreeEvaluator.hpp
0dvictor: No, it is not purely for SIMD. Although it is only used by vadd and vmul, the evaluator is designed to handle generic cases where ILOpCode can be translate to a single SSE/AVX instruction.

compiler/x/codegen/SIMDTreeEvaluator.cpp
0dvictor: It has been trimmed.

compiler/x/codegen/SIMDTreeEvaluator.cpp
0dvictor: Updated.

0dvictor:
Testing was done utilizing matmult test from jitbuilder - the only current use case for SIMD evaluators on X86. Trace logs have been examined to ensure the evaluators work as they intend to.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5240,3 +5240,64 @@ OMR::X86::TreeEvaluator::VMifArrayCmpEvaluator(TR::Node*, TR::CodeGenerator*)
...
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Address
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt8
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt16
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt32
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorInt64
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // VectorFloat
+   { BADIA32Op, ADDPDRegReg, BADIA32Op, MULPDRegReg, BADIA32Op }, // VectorDouble
+   { BADIA32Op, BADIA32Op,   BADIA32Op, BADIA32Op,   BADIA32Op }, // Aggregate
+   };
+// For ILOpCode that can be translated to single SSE/AVX instructions
```
andrewcraik: I'm still not seeing it - could you show an example of a translation you are envisioning here to justify the generic name? Please turn the comment into doxygen.

0dvictor: For example, TR::vadd can be translated to a single instruction ADDPD and hence can be handled by this evaluator. Other examples of potential uses of this evaluator are TR::dadd (ADDSD), TR::vdiv (DIVPD), etc.
Examples of ILOpCode that cannot be handled by this evaluator include TR::drem, there is no corresponding SSE/AVX instruction.
BTW, any other suggestions about the name?

andrewcraik: I think it will be cleaner if this code just handles the v* arithmetic opcodes that it can and that the normal arithmetic cases like dadd are handled in the normal arithmetic paths elsewhere in the JIT. This would mean this could be called VectorArithemticEvaluator and the table simplified in some way as a result I think. Unless there is a compelling reason to fuse the handling vector and non-vector I think it will be cleaner from a design and understanding point of view to keep them quite separate.

0dvictor: I don't think the table can be simplified unless doing some math on the enum; however this leads to more confusion.

andrewcraik: The x86 floating point operations operate on FPRs in the sense of virtual registers which happen to map down to the same physical registers as the VPRs. I see your point, I will ponder this a bit more and post again soon. I want to absorb what you said and consider it in the context of the existing floating point logic.

andrewcraik: I see your point about merging the two areas of code. I think this makes sense once x87 support is removed. Until x87 support is removed keeping the floating point code together will make more sense, but this is a good home for it in the future.

I think the name BinaryArithmeticEvaluator should be changed. I don't have a great name, but FPVBinaryArithmeticEvaluator would be an improvement. We could also have a FloatingPointBinaryArithmeticEvaluator and a VectorBinaryArithmeticEvaluator which call into this method as the implementation.

Thoughts?

0dvictor: I personally do not think FPV is an intuitive acronym; I'd prefer to use the full term if floating point and vector operations really need be emphasized in its name. i.e. FloatingPointAndVectorBinaryArithmeticEvaluator. The function is only called from the evaluator table and hence the length of name should not bring any inconvenience to developers.

andrewcraik: I'm good with the full title.

0dvictor: Updated.

andrewcraik:
I think this is looking good.

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,153 @@
+/*******************************************************************************
+*
+* (c) Copyright IBM Corp. 2000, 2017
```
0dvictor: Updated.

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,153 @@
...
+   TR_X86OpCodes opCode = BADIA32Op;
+   switch (node->getSize())
+      {
+      case 16:
+         opCode = MOVDQURegMem;
+         break;
+      default:
+         TR_ASSERT(false, "Unsupported fill size");
+         if (cg->comp()->getOption(TR_TraceCG))
+            traceMsg(cg->comp(), "Unsupported fill size\n");
```
0dvictor: Updated.

compiler/x/codegen/SIMDTreeEvaluator.cpp
```diff
@@ -0,0 +1,153 @@
...
+   TR_X86OpCodes opCode = BADIA32Op;
+   switch (node->getSize())
+      {
+      case 16:
+         opCode = MOVDQUMemReg;
+         break;
+      default:
+         TR_ASSERT(false, "Unsupported fill size");
+         if (cg->comp()->getOption(TR_TraceCG))
+            traceMsg(cg->comp(), "Unsupported fill size\n");
```
0dvictor: Updated.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5240,3 +5240,64 @@ OMR::X86::TreeEvaluator::VMifArrayCmpEvaluator(TR::Node*, TR::CodeGenerator*)
    {
    return 0;
    }
+
+enum BinaryArithmetic_t : uint32_t
```
0dvictor: Updated.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5240,3 +5240,64 @@ OMR::X86::TreeEvaluator::VMifArrayCmpEvaluator(TR::Node*, TR::CodeGenerator*)
...
+   {
+   BinaryArithmeticInvalid,
+   BinaryArithmeticAdd,
+   BinaryArithmeticSub,
+   BinaryArithmeticMul,
+   BinaryArithmeticDiv,
+   BinaryArithmeticNum
```
0dvictor: Updated.

0dvictor:
- Thank you for the valuable comments.
- This PR is the most basic SIMD support for packed double with simplicity and functional correctness in-mind. Further optimizations are being planned and will be implemented. I'll create an issue.
- Conflicts are resolved.

0xdaryl:
Thanks.  I'll approve/merge once an issue has been created.

0dvictor:
Issue #997 Vector Evaluator Support on X86 has been created.

andrewcraik:
@0xdaryl issue is created - should be good to approve/merge.

PR #864, opened 2017-03-15
Fix mtvsrwz macro to use intended shift operators

### LINKED ISSUE #863: GNU assembler macro for mtvsrwz instruction generates an illegal instruction
The following macro is intended to generate a `mtvsrwz` instruction when building with older versions of GNU `as` that predate its existence.
```
.macro mtvsrwz  vrt, ra
.int 0x7c0001e6 | \ra < 16 | (\vrt & 31) < 21 | (\vrt & 32) > 5
.endm
```
However where `>>` and `<<` are intended, `>` and `<` are used which results in a valid macro that always emits 0.

ymanton:
Fix an obvious typo to use shift operators instead of relational operators.

Fixes #863.


vijaysun-omr:
Simple fix for a typo. Merging.

PR #866, opened 2017-03-15
WIP: Allow for remat of stores dependent on indirect loads

ncough:
This extends RematTools to check if indirect loads are safe in a
region, based on whether there are indirect stores present.
It can then remat stores dependent on indirect loads, if possible.

### DIFF
--- a/compiler/optimizer/RematTools.cpp
+++ b/compiler/optimizer/RematTools.cpp
@@ -71,7 +71,8 @@ void RematSafetyInformation::dumpInfo(TR::Compilation *comp)
  */
 TR_YesNoMaybe RematTools::gatherNodesToCheck(TR::Compilation *comp,
       TR::Node *privArg, TR::Node *currentNode, TR::SparseBitVector &scanTargets,
-      TR::SparseBitVector &symRefsToCheck, bool trace, TR::SparseBitVector &visitedNodes)
+      TR::SparseBitVector &symRefsToCheck, bool trace, TR::SparseBitVector &visitedNodes,
+      bool safeIndirectLoads)
    {
    visitedNodes[currentNode->getGlobalIndex()] = true;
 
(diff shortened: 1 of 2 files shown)


compiler/optimizer/RematTools.hpp
```diff
@@ -53,20 +53,23 @@ class RematTools
    {
    static TR_YesNoMaybe gatherNodesToCheck(TR::Compilation *comp,
       TR::Node *privArg, TR::Node *currentNode, TR::SparseBitVector &scanTargets,
-      TR::SparseBitVector &symRefsToCheck, bool trace, TR::SparseBitVector &visitedNodes);
+      TR::SparseBitVector &symRefsToCheck, bool trace, TR::SparseBitVector &visitedNodes,
+      bool safeIndirectLoads);
```
andrewcraik: This should probably go before visitedNodes since it is meant to mirror the public API.

compiler/optimizer/RematTools.cpp
```diff
@@ -316,3 +323,25 @@ void RematTools::walkTreeTopsCalculatingRematFailureAlternatives(TR::Compilation
...
       }
    }
+
+/*
+ * This method is used to determine if indirect loads are safe for the nodes between
+ * the start inclusive and the end exclusive. It scans for indirect stores and
+ * determines the loads to be safe if none are found. The result can be passed to
+ * gatherNodesToCheck to allow remat of stores dependent on indirect loads.
+ */
+bool RematTools::checkForSafeIndirectLoads(TR::Compilation *comp, TR::TreeTop *start,
```
andrewcraik: Don't we also need to check for volatile reads/writes which create happens before relationships in the memory model which could make the remat invalid as well.

vijaysun-omr: Why are indirect stores sufficient to look for ? e.g. calls can also modify something that could be loaded indirectly right ? Perhaps one way would be to consult aliasing in this routine to assess which calls (and which stores, come to think of it) are worth panicking about.

andrewcraik: agreed - a check using the aliasing sets like done in the safety walk is probably the best idea.

PR #867, opened 2017-03-15
Configure options for NextGenHCR

ncough:
This ensures that the options required by NextGenHCR are
correctly configured when it is enabled.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1988,6 +1988,15 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
       self()->setOption(TR_DisableMethodHandleThunks); // Can't yet transition a MH thunk frame into equivalent interpreter frames
       }
 
+   if (self()->getOption(TR_EnableNextGenHCR) && !self()->getOption(TR_DisableOSR))
+      {
+      self()->setOption(TR_EnableOSR);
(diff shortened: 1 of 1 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -1988,6 +1988,15 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
       self()->setOption(TR_DisableMethodHandleThunks); // Can't yet transition a MH thunk frame into equivalent interpreter frames
       }
 
+   if (self()->getOption(TR_EnableNextGenHCR) && !self()->getOption(TR_DisableOSR))
+      {
+      self()->setOption(TR_EnableOSR);
+      self()->setOption(TR_EnableHCR);
+      self()->setOption(TR_DisableTM);
+      self()->setDisabled(redundantGotoElimination, true);
+      self()->setDisabled(redundantAsyncCheckRemoval, true);
```
andrewcraik: the disables need all need comments on why they are required. redundantAsyncCheckRemoval is fundamentally not compatible so has to be off. The need to turn off TM and redundantGotoElimination are more dubious and we should verify what the problem is and document that in the options since turning off TM and goto elimination are undesirable if there isn't a good moral reason for having to do so.

vijaysun-omr: I agree. Please update with comments and I can merge afterwards.

PR #868, opened 2017-03-15
Check if call is anchored to prior treetop in OSR

ncough:
OSR points are intended to occur at the first evaluation of a call.
However, the current check allows references to the call after it's
first evaluation to appear as OSR points. This change ensures the
call is anchored under the treetop being checked.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -706,12 +706,14 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
          // call node to ensure we see the correct state of the doNotProfile flag
          node = node->getFirstChild();
 
-         // In OSR HCR mode, the OSR point for a call occurs on its first evaluation.
-         // Therefore, it is necessary to check if the call is anchored under a prior treetop.
+         // The OSR point applies where the node is anchored, rather than where it may
+         // be commoned. Therefore, it is necessary to check if the node is anchored under
+         // a prior treetop.
          if (node->getReferenceCount() > 1)
             {
             TR::TreeTop *cursor = tt->getPrevTreeTop();
-            while (cursor && cursor->getNode()->getOpCodeValue() != TR::BBStart)
+            TR::TreeTop *extendedBlockStart = tt->getNode()->getBlock()->startOfExtendedBlock()->getEntry();
+            while (cursor && cursor != extendedBlockStart)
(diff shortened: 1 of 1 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -698,16 +698,39 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
    if (potentialOSRPoint && !self()->getOption(TR_FullSpeedDebug))
       {
-      // When in OSR HCR mode we need to make sure we check the BCI of the original
-      // call node to ensure we see the correct state of the doNotProfile flag
+
       if (self()->getHCRMode() == TR::osr &&
           (node->getOpCode().isCheck() || node->getOpCodeValue() == TR::treetop))
+         {
+         // When in OSR HCR mode we need to make sure we check the BCI of the original
+         // call node to ensure we see the correct state of the doNotProfile flag
```
andrewcraik: There is nothing explicit about calls here - I think this applies to anything with a ref count greater than 1, just at present only calls and asyncchecks are transition points and asyncchecks are not commoned - I don't know if the call needs to be handled differently so this comment may be spurious.

compiler/compile/OMRCompilation.cpp
```diff
@@ -698,16 +698,39 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
-      TR::ResolvedMethodSymbol *method = bci.getCallerIndex() == -1 ?
-         self()->getMethodSymbol() : self()->getInlinedResolvedMethodSymbol(bci.getCallerIndex());
-      potentialOSRPoint = method->supportsInduceOSR(bci, tt->getEnclosingBlock(), NULL, self(), false);
+
+         // In OSR HCR mode, the OSR point for a call occurs on its first evaluation.
+         // Therefore, it is necessary to check if the call is anchored under a prior treetop.
+         if (node->getReferenceCount() > 1)
+            {
+            TR::TreeTop *cursor = tt->getPrevTreeTop();
+            while (cursor && cursor->getNode()->getOpCodeValue() != TR::BBStart)
```
vijaysun-omr: Nodes can be commoned across an extended basic block, rather than just a single basic block. While this pass may not be running after block extension today it's probably better to end the search at the extended block start treetop (which other local optimizations do use via an existing API).

PR #869, opened 2017-03-15
Add flag for pending push symbols

ncough:
The current method to determine if a symbol is a pending push is to
check if it is stored in the method's array of pending pushes or by
checking its CP index. This adds a flag to determine if a symbol is
a pending push, applied to it when it is created.


andrewcraik:
Looks like a good change to me - makes the handling explicit which can only help I think.

vijaysun-omr:
Just so I am clear, what was the issue with the original way of checking for pending pushes that needed the addition of the flag ? Was it a performance reason or a functional one ?

ncough:
It was not desirable to check against the CPIndex as this was too general. The other options was to check against the ResolvedMethodSymbol’s collection of pending pushes, however, as it was necessary to frequently check for pending pushes in methods related to OSR, the use of a flag instead improved performance.

Issue #425, opened 2016-11-03
X86 Binary Encoding Scheme Redesign


0dvictor:
There are three major issues with current binary encoding scheme:
1. Binary encoding table only holds at most 3 bytes as instruction op code; however certain instructions requires more. Existing code works around this issue by arbitrarily dividing an instruction into prefix and opcode and then holds prefix information as op properties. The criteria differentiating prefix and opcode seems unclear and the actual usage is inconsistent.
2. Human readability is not optimal as entries in the encoding table appear quite different from Intel® 64 and IA-32 Architectures Software Developer’s Manual.
3. It does not provide a way to switch between SSE and AVX instructions. As the support for AVX is in-demand and mixing SSE/AVX instructions is a bad practice, codegen ideally should generate either instructions without minimal or even zero modifications on evaluators.

The design attached tackles all above three issues and aims to simplify the codegen.

[X86OpCode.pdf](https://github.com/eclipse/omr/files/570149/X86OpCode.pdf)

0dvictor:
Closing the Issue as all its work has been merged.

PR #875, opened 2017-03-15
X86 Binary Encoding Redesign

0dvictor:
Initial implementation of the redesigned X86 binary encoding scheme.
Refer to Issue #425 for design details.

### DIFF
--- a/compiler/x/codegen/OpBinary.cpp
+++ b/compiler/x/codegen/OpBinary.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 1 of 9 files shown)


compiler/x/codegen/OpBinary.cpp
```diff
@@ -911,3 +911,904 @@ const TR_X86OpCode::TR_OpCodeBinaryEntry TR_X86OpCode::_binaryEncodings[IA32NumO
...
+// X87 is only available on X86-32
+#ifdef TR_TARGET_64BIT
+#define X87_________________(x) 0,0,0
+#else
+#define X87_________________(x) (uint8_t)((x & 0xE0) >> 5), (uint8_t)((x & 0x18) >> 3), (uint8_t)(x & 0x07)
+#endif
+
+const TR_X86OpCode::OpCode_t TR_X86OpCode::_binaries[] =
+{
+    { VEX_L___, VEX_vNONE, PREFIX___, REX__, ESCAPE_____, 0xcc, 0, ModRM_NONE, Immediate_0 },    // BADIA32Op/int 3
```
andrewcraik: There should be a comment above this table listing the fields and a reference to where these contstants are defined so that they can be easily referenced - full understanding of the entries should be possible without hunting for the prefixes.

andrewcraik: There is a very important piece of implicit data for these instructions - namely the name of the instruction that each corresponds to... I'm wondering if we should try to make that more explicit some how - having information carried in comments is not a good way to make sure that information stays accurate and relevant.

One thought would be that since _binaries is an array it could be initialized as _binaries[numOpCodes] or something and each line would be something like:
_binaries[BADIAS32Op] = { .... } which would make the list not strictly dependent on order and it would enforce that the opcode was associated with the instance. I'm not sure if that incurs a penalty in initializing the array? I'm thinking not because the offsets are constant so it can all still be baked into the DLL.

Another option would be to add the op code to the encoding information, but that would increase footprint so I'm less of a fan of that.

@0dvictor would be interested to hear your thoughts. One additional benefit of the _binaries[<opcode>] format is that the instruction is on the left so you see it first before reading the encoding details which seems a more logical approach to me.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,6 +1219,115 @@ typedef enum
...
+      uint8_t vex_l : 2;
+      uint8_t vex_v : 1;
+      uint8_t prefixes : 2;
+      uint8_t rex_w : 1;
+      uint8_t escape : 2;
+      uint8_t opcode;
+      uint8_t modrm_opcode : 3;
+      uint8_t modrm_form : 2;
+      uint8_t immediate_size : 3;
+#ifdef TR_TARGET_32BIT
```
andrewcraik: Wouldn't it be better to have isX87 implemented but hardcoded to false for 64bit to just avoid a bit width specific API?

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,6 +1219,115 @@ typedef enum
...
+      inline bool isX87() const
+         {
+         return (prefixes == PREFIX___) && (opcode >= 0xd8) && (opcode <= 0xdf);
+         }
+#endif
+      inline bool isGroup07() const
```
andrewcraik: This needs a comment about what it means - the name is not obvious.

compiler/x/codegen/X86Ops_inlines.hpp
```diff
@@ -0,0 +1,104 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2000, 2016
```
andrewcraik: Please make sure all copyright dates are correct.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1418,7 +1528,7 @@ class TR_X86OpCode
          }
       return len + getOpCodeLength();
       }
-   uint8_t* binary(uint8_t* cur, uint8_t rex = 0)
+   uint8_t* deprecated_binary(uint8_t* cur, uint8_t rex)
```
andrewcraik: deprecated functions need a comment about why they were deprecated and what their replacements is/are.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,6 +1219,115 @@ typedef enum
...
+   enum TR_OpCodeImmediate : uint8_t
+      {
+      Immediate_0 = 0x0,
+      Immediate_1 = 0x1,
+      Immediate_2 = 0x2,
+      Immediate_4 = 0x3,
+      Immediate_8 = 0x4,
+      Immediate_S = 0x7,
+      };
+   template <typename TCurcor>
```
andrewcraik: what is the purpose of this template? I'm sure it is logical, but templates are used sparingly in TR due to the binary growth they can cause so a comment about their use in the PR would seem justified to me.

andrewcraik:
This is a good clean-up of the encoding in general and will make it a lot easier to support new encodings going forward. A few high-level comments beyond the specific comments above which are what stood out on initial review:
1) This code as nearly 0 comments - New classes and methods should have some kind of doxygen associated with them to make it easier to follow the code.
2) A reference to the GitHub issue in the code of where to find the design would seem sensible to me, but I'm unsure on project policy. @0xdaryl thoughts on this point?
3) We should be avoiding use of #ifdef macros where possible because they make the code harder to reason about. The main offender in this case is stuff related to x87 - my preference would be to hardcode the isX87 or whatever to false using a macro, but avoid having #ifdefs about 32bit spread around the code.
4) A change on this scale (like with your previous pull requests related to the codegen) needs some discussion of the testing that was done and how it was done. If you have any unit tests you wrote to exercise this thing then contribution of those where possible would be great.

Once you have addressed the above I'll continue my review.

mgaudet:
Regarding designs, maybe one option would be to turn that PDF presentation into some prose that could be checked in under `doc/compiler/x/OpCodeEncoding.md`?

compiler/x/codegen/X86Ops.hpp
0dvictor: The logic to check isX87() is valid on both 32-bit and 64-bit. The only reason behind the ifdef is that checking isX87() is unnecessary on 64-bit since X87 instruction should never appear under 64-bit. The ifdef can be simply removed.

compiler/x/codegen/OpBinary.cpp
0dvictor: If you meant to add a line of comment listing the meaning of each column, the name of the macros are designed to explain itself and I believe it is intuitive enough. In addition, it is such a long table, adding a line on the top does not help much as the line would be out of readable panel while looking majority of instructions.

If you meant to comments list values of enums containing VEX_L, PREFIX__, etc.; it is in fact to duplicate the code. It would be a hassle for maintaining. It is expected that anyone who tries to understand, modifies and/or debug this array to first understand the structure TR_X86OpCode::OpCode_t, a simple structure; and once understood, this array, TR_X86OpCode::_binaries[] is very intuitive. It is dangerous trying to interpret content of any array without understanding it element type. In addition, most, if not all, modern IDE can locate TR_X86OpCode::OpCode_t from any entry of this array. Therefore, I personally dislike the idea to comment on any structure details or enum values here, as it would likely induce the reader not to read TR_X86OpCode::OpCode_t before reading this array.

andrewcraik: I was meaning more comments on the column headings. If no one else thinks this is a problem I'm happy to leave it. If we aren't going to comment on the structure perhaps a pointer to the header in a comment would be appropriate since jumping to def doesn't work in GitHub source browsing for example:
// see compiler/x/codegen/OMRInstruction.hpp for structural information.

The point I am more keen on is trying to document the opcode in the source next to the binary information rather than in a comment. What are your thoughts on this?

0dvictor: Good suggestion, I'll add the comment as it warns readers to under the structure first.

0dvictor: I completely agree that relying on the order to correlate binaries to its meaning is extremely dangerous. We have this problem not only in the binary table but also in op properties, op names, etc. I have been thinking about this for awhile and am eager to get this restructured so that we do not rely on the order.
However, I believe this should be handled in a separate issue or pull request. I'm concerned that it may be more confusing and problematic if we resolve the binary table but not other places. Another reason why the initial version of new binary table is still rely on order is to be consist with the old design. In this way, we can keep the old approach so that we can verify the new implementation with the old one.
The plan to fully move away  from depending on orders is on the to-do list but I have got a chance fully think about it yet. One idea is to use macro expansion tricks like what we use for Optimizations.enum.

andrewcraik: Ok yes, that is a very valid point. Let's keep this to the current change to the encoding method and we can do the reformat to a generated table as a follow-on item. Thanks!

compiler/x/codegen/X86Ops.hpp
0dvictor: Updated.

compiler/x/codegen/X86Ops_inlines.hpp
0dvictor: Updated.

compiler/x/codegen/X86Ops.hpp
0dvictor: These are old implementations. Comments added.

compiler/x/codegen/X86Ops.hpp
0dvictor: These are to unify length calculation and binary generation as they are the same except one increases counter and the other one writes bytes. Comments added.

0dvictor:
@andrewcraik I have updated code according your feedback, please take a look again.
1. I've added more comments in places that I believed necessary, let me know if any other places are worth commenting.
2. Sure, once we are cleared about the project policy, I can update to include the link to the Issue.
3. All unnecessary ifdef/ifndef are removed.
4. This PR includes two commits, the first commit contains logic to generate binaries twice with both the old and new scheme; then verify the two byte per byte. A build with this logic has passed all required functional test and hence it is believed that the new scheme generates identical code as the old one. Since these logic would have performance penalty, the second commit removes them.

0dvictor:
@mgaudet I like this idea to check-in the documentation. I'll study the .md format and see if I can rewrite the PDF to .md.

andrewcraik:
Could the checking code be put under an environment variable or option temporarily (eg we don't usually generate the old encoding or do anything with the old encoding, but if we set an envvar like paranoidX86Encoding. This would provide an easy verification to do on x86 failures while the change settles in - it can be cleaned up after the code has soaked for a while. If there is a performance reason to not have this option please say, but the verification seems like useful logic to keep around for a short time.

andrewcraik:
The implementation looks good to me and the code does look much more clean and self-contained now. I'm interested to see if others feel the level of commenting now is good - it seems sparse but sufficient to me, but am happy to hear other points of view.

0dvictor:
Good suggestion, now the verification logic is guarded by an environment variable.

doc/compiler/x/OpCodeEncoding.md
```diff
@@ -0,0 +1,99 @@
...
+    Immediate_1 = 0x1,
+    Immediate_2 = 0x2,
+    Immediate_4 = 0x3,
+    Immediate_8 = 0x4,
+    Immediate_S = 0x7,
+};
+```
+
+# Generate Non-AVX Instruction
+1. Generate lagacy prefixes according to OpProperties and OpProperties2.
```
mgaudet: spelling error ('lagacy' - > 'legacy')

doc/compiler/x/OpCodeEncoding.md
```diff
@@ -0,0 +1,99 @@
...
+2. Generate prefixes according to prefixes field.
+3. Obtain REX prefix from operand and set REX.W according to rex_w field.
+3.1 Generate REX prefix
+4. Generate opcode escape according to escape field.
+5. Write opcode
+6. Set and write ModR/M field if necessary
+
+# CONTENT BELOW ARE NOT IMPLEMENTED YET, TO BE UPDATED.
```
mgaudet: Maybe just call this 'future work?'

doc/compiler/x/OpCodeEncoding.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Generate AVX Instruction
```
mgaudet: Subheads should use `##` to reduce size :)

doc/compiler/x/OpCodeEncoding.md
0dvictor: updated.

doc/compiler/x/OpCodeEncoding.md
0dvictor: Good suggestion.

doc/compiler/x/OpCodeEncoding.md
0dvictor: Updated.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,7 +1220,121 @@ typedef enum
...
+         }
+      // check if the instuction is part of Group 7 OpCode Extension
+      inline bool isGroup07() const
+         {
+         return (escape == ESCAPE_0F__) && (opcode == 0x01);
+         }
+      // TBuffer should only be one of the two: Estimator when calculating length, and Writer when generating binaries.
+      template <class TBuffer> typename TBuffer::cursor_t encode(typename TBuffer::cursor_t cursor, uint8_t rexbits) const;
+      };
+   template <typename TCurcor>
```
0dvictor: Updated.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,7 +1220,121 @@ typedef enum
...
+   class BufferBase
+      {
+      public:
+      typedef TCurcor cursor_t;
```
0dvictor: Updated.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -1220,7 +1220,121 @@ typedef enum
...
+         {
+         *((T*)cursor) = binaries;
+         cursor+= sizeof(T);
+         }
+      };
+   // TBuffer should only be one of the two: Estimator when calculating length, and Writer when generating binaries.
+   template <class TBuffer> inline static typename TBuffer::cursor_t encode(TR_X86OpCodes op, typename TBuffer::cursor_t cursor, uint8_t rex)
+      {
+      // OpCodes beyond FENCE are not X86 instructions and hence they do not have binary encodings.
+      return op < FENCE ? _binaries[op].encode<TBuffer>(cursor, rex) : cursor;
```
0dvictor: Updated.

doc/compiler/x/OpCodeEncoding.md
```diff
@@ -0,0 +1,99 @@
+# X86 Binary Encoding Scheme - Issue [#425](https://github.com/eclipse/omr/issues/425)
+
+OMR's X86 Binary Encoding Scheme is inspired by Intel's VEX and EVEX prefixes, which compact mandaroty prefixes and multi-byte opcode escape codes into 2 bits each.
```
0dvictor: Updated.

doc/compiler/x/OpCodeEncoding.md
```diff
@@ -0,0 +1,99 @@
...
+
+Following structure is used to hold an instruction:
```
0dvictor: Updated.

0dvictor:
Thank you for the review and valuable comments.
1. By "all required functional tests", I meant both TravisCI tests and IBM's internal much deeper test suit.
2. The changeset to delete deprecated code is ready and I will deliver it in a subsequent PR 3-4 days after this PR gets merged. I completely agree to and am eager to delete deprecated code; however, I would like to have this version on record so that other developers on OMR could have a chance to see the changes on X86 CG.
3. Yes, compile-time performance tests are cleared. No observable performance difference was observed.
4. AVX instructions will be supported by a subsequent PR, including VEX256/512. It will be delivered as soon as the deprecated code is deleted.

0xdaryl:
Thanks for addressing my comments.

PR #877, opened 2017-03-15
Defer _stackTopRegister update until commit

jduimovich:
This change defers _stackTopRegister updates until the Commit operation. This ensures that once the stack elements are visible in the vm stack, the vm stack pointer (_stackTopRegister) matches that stack state.
Prior code maintained the _stackTopRegister at each push/pop/drop which meant the stack top visible by  the vm code, may not match vm stack as they are not committed at that point.

Signed-off-by: John Duimovich <jduimovich@gmail.com>


mstoodle:
LGTM and has the additional benefit that it reduces the number of nodes and locals generated and reduces the load on the optimizer to see through them all. It's cleaner than the original approach. Thanks @jduimovich !

Just have to wait for testing to complete.

PR #878, opened 2017-03-16
Remove warm snippets and code cache estimation

0xdaryl:
* Remove deprecated warm code cache estimation logic
* Remove `isWarm` field from Z constant snippet hash
* Remove warm snippet references

### DIFF
--- a/compiler/arm/codegen/OMRCodeGenerator.cpp
+++ b/compiler/arm/codegen/OMRCodeGenerator.cpp
@@ -47,10 +47,6 @@
 #include "il/TreeTop_inlines.hpp"
 #include "il/symbol/ParameterSymbol.hpp"
 
-// Amount to be added to the estimated code size to ensure that there are long
-// branches between warm and cold code sections (must be multiple of 8 bytes).
-//
-#define MIN_DISTANCE_BETWEEN_WARM_AND_COLD_CODE (32768+8)
 TR_Processor OMR::ARM::CodeGenerator::_processor=TR_NullProcessor;
 
 static int32_t identifyFarConditionalBranches(int32_t estimate, TR::CodeGenerator *cg);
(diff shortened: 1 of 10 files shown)


0xdaryl:
**Note to Reviewers on Power**

The logic to eliminate the warm estimation code on Power is more complicated than
other architectures, hence it deserves additional explanation of my reasoning:

* In `OMR::Power::CodeGenerator::doBinaryEncoding()`, `data.warmEstimate` is initialized to 0

* The first opportunity for `data.warmEstimate` to be anything other than 0 is
after a call to `identifyFarConditionalBranches()`.  That call takes two parameters of
interest to us: 1) a pointer to an `incWarmSize` that will be added to `data.warmEstimate`
upon completion of that call, and 2) the current `data.warmEstimate` (which is 0).
However, `incWarmSize` will never be anything other than zero after that function
because the increment logic within is gated on the warm estimate being greater than the
offset of some branch instruction in question.  This will never occur since it is always 0.
Hence, after the call `data.warmEstimate` remains 0.

* The next opportunity for modifying `data.warmEstimate` is gated on
whether `getGlobalLabelsEncountered().empty()` returns false (i.e., there's something
in the list).  Nothing is ever added to this list in OMR or any downstream project.
Hence, the logic to update `data.warmEstimate` will never execute.  The "global
labels encountered" logic has been removed as unused.

* Following this, `data.warmEstimate` should only be 0 and hence subsquent paths
can be folded.

0xdaryl:
@fjeremic @ymanton : I'd appreciate a review.

ymanton:
Power changes look good to me.

(Most of this functionality was intended for static compilation and was never actually used in any meaningful way.)

fjeremic:
LGTM, a much welcomed change.

PR #879, opened 2017-03-16
Allow changing MSVC runtime dll via preprocessor define

dnakamura:
omrsig has a hard coded dependency on the MSVC runtime dll name. This
patch substitutes the name with a preprocessor macro to allow changing the
name via compile time flags


charliegracie:
@dnakamura Can you limit the first line of the commit message to 50 characters?  Thanks!

PR #880, opened 2017-03-16
Add support for processing individual files in tracegen

dnakamura:
Allow tracegen to process induividual files, rather than searching a root
directory. Default behaviour remains the same.

### DIFF
--- a/tools/tracegen/TraceGen.cpp
+++ b/tools/tracegen/TraceGen.cpp
@@ -58,12 +58,13 @@ help(int argc, char *argv[])
 			|| 0 == strcmp(argv[i], "/h")
 			|| 0 == strcmp(argv[i], "/help")
 		) {
-			printf("%s [-threshold num] [-w2cd] [-generateCfiles] [-treatWarningAsError] [-root rootDir] [-force]\n", argv[0]);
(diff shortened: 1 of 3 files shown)


tools/tracegen/TraceGen.cpp
```diff
@@ -58,12 +58,13 @@ help(int argc, char *argv[])
 			|| 0 == strcmp(argv[i], "/h")
 			|| 0 == strcmp(argv[i], "/help")
 		) {
-			printf("%s [-threshold num] [-w2cd] [-generateCfiles] [-treatWarningAsError] [-root rootDir] [-force]\n", argv[0]);
+			printf("%s [-threshold num] [-w2cd] [-generateCfiles] [-treatWarningAsError] [-root rootDir] [-file file.tdf] [-force]\n", argv[0]);
```
charliegracie: Is passing both -root and -file ok?

dnakamura: Depends on your definition of ok.  You probably wouldn't want to, but it will still work correctly

charliegracie: Ok thanks.  Was just wondering if we should make it fail if you specify both but if it works then I am ok with the current change.

PR #882, opened 2017-03-16
Disable propagation via global relative constraints

jdmpapin:
Runaway recursion has been observed in VP (value propagation) when
propagating through global relative constraints using adde6ad58284.
Disabling until more is known.


jdmpapin:
[Whitespace-insensitive diff](https://github.com/eclipse/omr/pull/882/files?w=1)

PR #883, opened 2017-03-16
Remove code related to -Xnoquickstart

mpirvu:
-Xnoquickstart is an option introduced as a workaround for micro-benchmarks
that had most/all of the code in a method invoked only once. Thus, even if
the JIT compiled that important method, the transition to the new body
never happened.
This restriction was eliminated with the advent of DLT (dynamic loop transfer) 
which allows the transition from interpreted code to jitted code during the same
invocation. As a consequence -Xnoquickstart is more or less useless now.
This change set deprecates the -Xnoquickstart option (the JIT still recognizes
the option, but does nothing).

For documentation purposes: -Xnoquickstart compiles loopy methods from their
first invocation (count=0) and uses hot opt level.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -3978,9 +3978,6 @@ OMR::Options::setCounts()
       }
    else // no counts string specified
       {
-      if (self()->getOption(TR_noquickstart) && !self()->getOption(TR_MimicInterpreterFrameShape))
-         _optLevel = hot; // Fixed opt level
-
       // No need for sampling thread if only one level of compilation and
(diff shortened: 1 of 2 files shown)


mpirvu:
I removed the reference to the RTC work item.

vijaysun-omr:
"For documentation purposes: Xquickstart compiles loopy methods from their
first invocation (count=0) and use hot opt level."....you meant -Xnoquickstart, right ? 
If so, please amend the message.

vijaysun-omr:
Code changes look okay to me. Please amend the message and I shall merge once checks pass.

mpirvu:
I have changed the comment. Thanks for spotting the mistake.

PR #884, opened 2017-03-16
Deprecate MM_ConfigurationLanguageInterface etc

ktbriggs:
Additional changes required to ensure compilation without
OMR_GC_SCAVENGER and OMR_GC_SEGREGATED_HEAP flags.

Issue: #508


ktbriggs:
@jduimovich these are the changes required to build with your stripped-down osx configuration.

ktbriggs:
Rebased and pushed again to force new Travis build. No code changes, Last Travis build for linux_x86-64 timed out near end of tests.

charliegracie:
@ktbriggs you do not need to push to get the job restarted. Committers looking at the change can just restart the job.

gc/base/GCExtensionsBase.hpp
```diff
@@ -1084,6 +1084,7 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
 		, rootScannerStatsEnabled(false)
 		, softMx(0) /* softMx only set if specified */
 		, gcThreadCountForced(false)
+		, scavengerEnabled(false)
```
charliegracie: scavengerEnabled needs to stay under the OMR_GC_MODRON_SCAVENGER compile flag.  Everywhere that looks at it needs to be properly #if defined.

ktbriggs:
@charliegracie this is ready to go but overlaps @rwy0717's changes from https://github.com/eclipse/omr/pull/890. I prefer my solution for scavengerEnabled flag and suggest that it be extended to deal with concurrentMark if latter is to be conditionally included in MM_GCExtensionsBase. 

So either @rwy0717 can update his branch to include unconditional inline tests for scavenger/concurrent capability or I can extend mine to deal with concurrentMark. @charliegracie, @rwy0717 it's your call.

ktbriggs:
@charliegracie actually I (as non-committer) do need to push in order to restart a build, and sometimes committers will wait until all checks have passed before looking at changes. This is unfortunate and costly (only 1/4 jobs timed out in this case, pushing idempotent commit forced 4/4 rebuilds). Would be convenient if non-committers could restart individual jobs when they fail owing to machine/travis issues.

ktbriggs:
Commit f738c73 extends previous change included in 702ddb9 (conditionally compile MM_GCExtensionsBase::scavengerEnabled, define and use MM_GCExtensionsBase::isScavengerEnabled() unconditionally) to treat MM_GCExtensionsBase fields concurrentmark, concurrentSweep similarly. This obviates @rwy0717's changes in #890, which can be discarded.

gc/base/GCExtensionsBase.hpp
```diff
@@ -763,6 +763,36 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
...
+	{
+#if defined(OMR_GC_MODRON_SCAVENGER)
+		return scavengerEnabled;
+#else
+		return false;
+#endif /* defined(OMR_GC_MODRON_SCAVENGER) */
+	}
+
+	MMINLINE bool
+	isConcurrentMark()
```
charliegracie: Can we change this function to isConcurrentMarkEnabled()?  This way the functions all have similar names.

gc/base/GCExtensionsBase.hpp
```diff
@@ -763,6 +763,36 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
...
+	{
+#if defined(OMR_GC_CONCURRENT_MARK)
+		return concurrentMark;
+#else
+		return false;
+#endif /* defined(OMR_GC_CONCURRENT_MARK) */
+	}
+
+	MMINLINE bool
+	isConcurrentSweep()
```
charliegracie: Can we change this function to isConcurrentSweepEnabled()?  This way the functions all have similar names.

ktbriggs:
@charliegracie 

1. yes scavengerEnabled, concurrentMark, concurrentSweep fields (if compiled) are still public in MM_GCExtensionsBase and referenced directly in several dozen places. I was conservative in applying use of new isX() methods, making code these changes only where they would make a safe and significant difference. 
2. should we also change above variable names accordingly if renaming isConcurrent{Mark,Sweep}Enabled()? or leave all as is, so that familiar names are retained and corresponding functions match variable names (following common pattern for these sorts of methods)?

charliegracie:
I was just asking questions to fully understand what was being changed and what was left to do in future changes.  I would like to see the new functions named as I suggested so they are consistent.  Under this change you do not have to change the variable names or replace all of the uses as I agree that those are follow on issues.

ktbriggs:
Ok. Method names changed (isConcurrent{Mark,Sweep}()->isConcurrent{Mark,Sweep}Enabled()). The same files were touched as in most recent commit so squashed. Only method names were updated in this commit.

Pull request is on hold though pending resolution of a persistent failure of JIT sanity mode 119 tests. They have been failing only since previous commit (squashed into this one) and they only fail on Axxon runs -- same commands run fine when run from terminal and same machine and SDK as with Axxon build. It looks like the job is relying on PATH to run java, but SDK under test is not on PATH.

ktbriggs:
This commit squashes two previous commits, without code changes. All tests have passed.

@charliegracie this is ready to go (pending completion of Travis build).

charliegracie:
Can you remove the WIP from the title if it is completed?  Thanks.

ktbriggs:
WIP removed.

PR #886, opened 2017-03-17
Improve reg pressure sim for small 64-bit consts

mstoodle:
I came across an example block with a sequence of code using
lconsts to address an array of structs. Unhappy with the
register allocation in the method, I tracked the problem
down to the handling of register pressure for these lconsts:
each one increased register pressure by 1 until the sequence
blew the register budget. The immediate reason for the
increased register pressure is that these lconsts are assumed
to require a register for the duration of the block, even
though the constants are small and generally end up folded
into memory references.

There is an escape hatch for 32-bit integers to say they do
not require a register. This commit adds 32-bit integer
constants masquerading as 64-bit integer constants due to
IL type consistency requirements on 64-bit platforms. It
allows 64-bit integer constants where node->isHighWordZero().

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -3103,9 +3103,10 @@ uint8_t OMR::X86::CodeGenerator::nodeResultGPRCount(TR::Node *node, TR_RegisterP
    TR_ASSERT(!self()->comp()->getOption(TR_DisableRegisterPressureSimulation), "assertion failure");
 
    // 32-bit integer constants practically never need a register on x86
+   //  (includes 64-bit integer constants with high word zero needed to maintain IL correctness)
    //
    if (  node->getOpCode().isLoadConst()
-      && node->getSize() <= 4
+      && (node->getSize() <= 4 || (node->getType().isInt64() && node->isHighWordZero()))
       && (node->getType().isAddress() || node->getType().isIntegral())
       && !(  self()->simulatedNodeState(node, state)._keepLiveUntil != NULL // Check if parent will become a RegStore that keeps this node live
          && state->_currentTreeTop->getNode()->getOpCode().isStoreDirect()


compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -3117,9 +3117,10 @@ uint8_t OMR::X86::CodeGenerator::nodeResultGPRCount(TR::Node *node, TR_RegisterP
    TR_ASSERT(!self()->comp()->getOption(TR_DisableRegisterPressureSimulation), "assertion failure");
 
    // 32-bit integer constants practically never need a register on x86
+   //  (includes 64-bit integer constants with high word zero needed to maintain IL correctness)
    //
    if (  node->getOpCode().isLoadConst()
-      && node->getSize() <= 4
+      && (node->getSize() <= 4 || node->getType().isInt64() && node->isHighWordZero())
```
fjeremic: May I suggest parenthesis around the `node->getType().isInt64() && node->isHighWordZero()`? A keen reader will correctly parse this however mere mortals may not remember that `&&` has higher precedence than `||`  in C++ and may interpret this line incorrectly.

mstoodle: yeah, i'll do that, it's a good suggestion

mstoodle:
I'm afraid the "test case" where I encountered this problem on was a 370MB log file from a Rosie compilation using JitBuilder, which may not run on Z. I don't recommend that particular example.

You need to get a small lconst being used in the addressing arithmetic for a memory access. Then from the register simulation output (from traceGRA) you can see the pressure go up by one and never drop for the rest of the block.

I discovered the problem while tracking how register pressure was calculated through a sequence of IL that basically stored constants into a the fields for a set of structures in an array (so A[0].f1=C01;A[0].f2=C02;...;A[1].f1=C11;A[2].f2=C12; ...


It might be possible to concoct a JitBuilder test that would run on Z and exhibit the IL pattern I was seeing ....Maybe a modified version of StructArray in jitbuilder/release/src ?

mstoodle:
IBM has kindly decided to do some performance measurements with this change in/out, just to measure the impact (since GRA can be a touchy thing). Hopefully we'll hear back on the results in the next few days.

andrewcraik:
Performance measurements are done and the results say that there is no throughput performance regression from this change - we didn't see a measurable benefit in the cases we tried, but it seems like a fine change to make and could improve some specific situations quite noticeably as described in the PR. Compile time appears to have head near constant as well. LGTM.

vijaysun-omr:
Merging since performance result is in.

ymanton:
I think the proper way to do this would be to use the `isMaterialized()` codegen query, that way you don't need the original escape hatch or the node flag check.

mstoodle:
Hm, those isMaterialized() queries are a bit inconsistent in their definition, though I suppose you're right that isMaterialized(int64_t) could be used to replace the line of code I changed in this PR and would make it more precise.

ymanton:
You should be aware of #837 as @NigelYiboYu is making some changes to `isMaterialized()` and its friends. X would still need an appropriate implementation.

andrewcraik:
@0dvictor FYI since you are doing work on the x codegen changes in this area might be something you want to look at.

PR #888, opened 2017-03-17
Add UpdateStack API to handle stack moves

charliegracie:
In a lot of Virtual Machines the Stack is allocated on the heap.  This
means that the Stack may move.  If the Virtual Machine stack moves we
need an API to update the simulated Operand Stack.

The API needs to update OperandStack_base and _stackTopRegister.  I
updated the operandstacktest to move the stack a few times throughout
the execution and cleaned it up so the stack is properly freed.

### DIFF
--- a/compiler/ilgen/VirtualMachineOperandStack.cpp
+++ b/compiler/ilgen/VirtualMachineOperandStack.cpp
@@ -93,25 +93,26 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
       }
    }
 
-   void
-   VirtualMachineOperandStack::Reload(TR::IlBuilder* b)
+void
+VirtualMachineOperandStack::Reload(TR::IlBuilder* b)
    {
-       TR::IlType* Element = _elementType;
-       TR::IlType* pElement = _mb->typeDictionary()->PointerTo(Element);
-       // reload the elements back into the simulated operand stack
-       // If the # of stack element has changed, the user should adjust the # of elements 
-       // using Drop beforehand to add/delete stack elements.
-       TR::IlValue* stack = b->Load("OperandStack_base");
-       for (int32_t i = _stackTop; i >= 0; i--) {
-           _stack[i] = b->LoadAt(pElement,
-               b->IndexAt(pElement,
-                   stack,
-                   b->ConstInt32(i - _stackOffset)));
+   TR::IlType* Element = _elementType;
+   TR::IlType* pElement = _mb->typeDictionary()->PointerTo(Element);
+   // reload the elements back into the simulated operand stack
+   // If the # of stack element has changed, the user should adjust the # of elements
+   // using Drop beforehand to add/delete stack elements.
+   TR::IlValue* stack = b->Load("OperandStack_base");
+   for (int32_t i = _stackTop; i >= 0; i--)
+      {
+      _stack[i] = b->LoadAt(pElement,
+                  b->   IndexAt(pElement,
+                           stack,
+                  b->      ConstInt32(i - _stackOffset)));
        }
    }
 
-   void
-   VirtualMachineOperandStack::MergeInto(OMR::VirtualMachineOperandStack* other, TR::IlBuilder* b)
+void
+VirtualMachineOperandStack::MergeInto(OMR::VirtualMachineOperandStack* other, TR::IlBuilder* b)
    {
    TR_ASSERT(_stackTop == other->_stackTop, "stacks are not same size");
    for (int32_t i=_stackTop;i >= 0;i--)
@@ -128,6 +129,14 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
          b->StoreOver(other->_stack[i], _stack[i]);
          }
       }
+   }
+
+// Update the OperandStack_base and _stackTopRegister after the Virtual Machine moves the stack.
+// This call will normally be followed by a call to Reload if any of the stack values changed in the move
+void
+VirtualMachineOperandStack::UpdateStack(TR::IlBuilder *b, TR::IlValue *stack)
+   {
+   b->Store("OperandStack_base", stack);
    }
 
 // Allocate a new operand stack and copy everything in this state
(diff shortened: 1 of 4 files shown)


jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -180,6 +181,42 @@ STACKVALUETYPE *OperandStackTestMethod::_realStack = NULL;
...
+OperandStackTestMethod::moveStack()
+   {
+   STACKVALUETYPE *newStack = (STACKVALUETYPE *) malloc (_realStackSize * sizeof(STACKVALUETYPE));
+   int delta = 0;
+   if (useThreadSP)
+      delta = thread.sp - _realStack;
+   else
+      delta = _realStackTop - _realStack;
+   memcpy(newStack, _realStack, _realStackSize * sizeof(STACKVALUETYPE));
+   free(_realStack);
```
mstoodle: maybe paint the old stack area before freeing it?

jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -180,6 +181,42 @@ STACKVALUETYPE *OperandStackTestMethod::_realStack = NULL;
...
+   _realStackTop = _realStack + delta;
+   thread.sp = _realStackTop;
+
+   return _realStack - 1;
+   }
+
+void
+OperandStackTestMethod::freeStack()
+   {
+   free(_realStack);
```
mstoodle: same as the other free, maybe paint it first

jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -401,16 +438,17 @@ OperandStackTestMethod::OperandStackTestMethod(TR::TypeDictionary *d)
...
-#define COMMIT(b)       ((b)->vmState()->Commit(b))
-#define RELOAD(b)       ((b)->vmState()->Reload(b))
-#define PUSH(b,v)	(STACK(b)->Push(b,v))
-#define POP(b)          (STACK(b)->Pop(b))
-#define TOP(b)          (STACK(b)->Top())
-#define DUP(b)          (STACK(b)->Dup(b))
-#define DROP(b,d)       (STACK(b)->Drop(b,d))
-#define PICK(b,d)       (STACK(b)->Pick(d))
+#define STACK(b)	       (((TestState *)(b)->vmState())->_stack)
+#define STACKTOP(b)	       (((TestState *)(b)->vmState())->_stackTop)
```
mstoodle: could you convert the tabs in these two lines to spaces? :)

jitbuilder/release/src/OperandStackTests.cpp
```diff
@@ -401,16 +438,17 @@ OperandStackTestMethod::OperandStackTestMethod(TR::TypeDictionary *d)
...
+#define COMMIT(b)          ((b)->vmState()->Commit(b))
+#define RELOAD(b)          ((b)->vmState()->Reload(b))
+#define UPDATESTACK(b, v)  (STACK(b)->UpdateStack(b, v))
+#define PUSH(b,v)	       (STACK(b)->Push(b,v))
```
mstoodle: another tab here :)

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -93,25 +93,26 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
...
+   // reload the elements back into the simulated operand stack
+   // If the # of stack element has changed, the user should adjust the # of elements
+   // using Drop beforehand to add/delete stack elements.
+   TR::IlValue* stack = b->Load("OperandStack_base");
+   for (int32_t i = _stackTop; i >= 0; i--)
+      {
+      _stack[i] = b->LoadAt(pElement,
+                  b->   IndexAt(pElement,
+                           stack,
+                  b->      ConstInt32(i - _stackOffset)));
```
mstoodle: thanks for cleaning up here

compiler/ilgen/VirtualMachineOperandStack.cpp
```diff
@@ -130,6 +131,18 @@ VirtualMachineOperandStack::Commit(TR::IlBuilder *b)
...
+// Update the OperandStack_base and _stackTopRegister after the Virtual Machine moves the stack.
+// This call will normally be followed by a call to Reload if any of the stack values changed in the move
+void
+VirtualMachineOperandStack::UpdateStack(TR::IlBuilder *b, TR::IlValue *stack)
+   {
+   b->Store("OperandStack_base", stack);
+   // Adjust the vm _stackTopRegister by number of elements that have been pushed onto the stack.
+   // _stackTop is -1 at 0 pushes, 0 for 1 push, so # of elements to adjust by is _stackTop+1
+   _stackTopRegister->Store(b, stack);
+   _stackTopRegister->Adjust(b, (_stackTop+1)*_pushAmount);
```
mstoodle: I don't think we need to adjust _stackTopRegister here, after John's change to remove all interim adjustments to the stack top register. _stackTop doesn't change with UpdateStack(), and Commit() will update _stackTopRegister in exactly the same way you have here (you clearly copied the lines from Commit :) ) when the next Commit() call is made.

jitbuilder/release/src/OperandStackTests.cpp
charliegracie: done.

jitbuilder/release/src/OperandStackTests.cpp
charliegracie: done

jitbuilder/release/src/OperandStackTests.cpp
charliegracie: yep thought I had got them all but I guess I missed a few

jitbuilder/release/src/OperandStackTests.cpp
charliegracie: done

compiler/ilgen/VirtualMachineOperandStack.cpp
charliegracie: You are correct.  I thought that the _stackTopRegister would need to be updated to Reload to work but it does not.  Removed those two lines.

mstoodle:
looks good, thanks @charliegracie !

PR #889, opened 2017-03-17
JitBuilder service so nodes get bytecode indices

mstoodle:
Currently, all IL nodes generated by JitBuilder services
get the practically useless bytecode index of -1. This commit
provides a single service that enables JitBuilder clients
to get correct bytecode indices, but it requires an additional
call to be inserted into their bytecode translation loop.

On a BytecodeBuilder *b, call b->SetCurrentIlGenerator() and
nodes generated from that point forward will be annotated
with b->_bcIndex value, which is stored from the construction
of the object.

Strategically, I think we should make the worklist in
MethodBuilder actually know about the BytecodeBuilder
objects rather than just their bytecode indices, and
that way the worklist could call this new function
transparently before it hands the object back to the
client compiler to translate its operations.

### DIFF
--- a/compiler/ilgen/BytecodeBuilder.cpp
+++ b/compiler/ilgen/BytecodeBuilder.cpp
@@ -48,6 +48,19 @@ OMR::BytecodeBuilder::BytecodeBuilder(TR::MethodBuilder *methodBuilder,
    initSequence();
    }
 
+/**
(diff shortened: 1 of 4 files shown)


vijaysun-omr:
Adds a new service that makes the IL trees "more complete" in that they have a way to specify the bytecode index on the node.

PR #890, opened 2017-03-18
Fix missing GC feature ifdefs

rwy0717:
Recent versions of OMR were using some concurrent mark and scavenger
features without the needed `#if defined()` guards. This patch allows
the GC to compile without concurrent or scavenger enabled.

### DIFF
--- a/gc/base/Configuration.cpp
+++ b/gc/base/Configuration.cpp
@@ -169,9 +169,11 @@ MM_Configuration::initializeEnvironment(MM_EnvironmentBase* env)
 	case gc_modron_allocation_type_tlh:
 		env->_objectAllocationInterface = MM_TLHAllocationInterface::newInstance(env);
 		break;
+#if defined(OMR_GC_SEGREGATED_HEAP)
 	case gc_modron_allocation_type_segregated:
 		env->_objectAllocationInterface = MM_SegregatedAllocationInterface::newInstance(env);
 		break;
+#endif /* defined(OMR_GC_SEGREGATED_HEAP) */
 	default:
 		Assert_MM_unreachable();
 		break;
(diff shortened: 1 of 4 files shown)


rwy0717:
@ktbriggs This is my version of the feature flags fix you're working on in #884.

gc/base/Configuration.cpp
```diff
@@ -169,9 +169,11 @@ MM_Configuration::initializeEnvironment(MM_EnvironmentBase* env)
 	case gc_modron_allocation_type_tlh:
```
charliegracie: There is a TLH build flag as well.  Since you are cleaning up this function can you also wrap this under the appropriate #if defined as well?  Thanks

rwy0717: Sure, but then we have the awkward problem that there are no valid allocation types. I'm not really familiar with this code, but maybe we should introduce a generic allocation type, and initialize with an `MM_AllocationInterfaceGeneric` here.

As well, It looks like the entirety of `omrgcconsts.h` is in need of ifdefs.

ktbriggs:
@rwy0717 and @charliegracie I see Robert's changes re scavengerEnabled and concurrentMark flags and they conflict/extend my fix for scavengerEnabled. I would really like to minimize requirement for #ifdef'ing
checks for these capabilities as this makes a mess of the source code and should be minimized. 

For scavengerEnabled I defined in MM_GCExtensionsBase:
```
	MMINLINE bool
	isScavengerEnabled()
	{
#if defined(OMR_GC_MODRON_SCAVENGER)
		return scavengerEnabled;
#else
		return false;
#endif /* defined(OMR_GC_MODRON_SCAVENGER) */
	}
```
which can be called without performance penalty and without requiring #ifdef around the call. 

Can we do this for concurrentMark as well? And I suggest that only one of use (Robert or myself) complete this change to avoid messy conflicts etc.

charliegracie:
I am closing this item in favor of #884 as it improves more locations.  Thanks @rwy0717

PR #891, opened 2017-03-18
Improve struct and union field names in logs

mstoodle:
JitBuilder struct and union field names are not well presented in JIT logs because of the way their symbols are created. Using createNamedSymbol instead so names appear as "struct.field" or "union.field" in logs.

### DIFF
--- a/compiler/ilgen/TypeDictionary.cpp
+++ b/compiler/ilgen/TypeDictionary.cpp
@@ -379,7 +379,9 @@ UnionType::getFieldSymRef(const char *fieldName)
       auto symRefTab = comp->getSymRefTab();
       TR::DataType type = info->getPrimitiveType();
 
-      TR::Symbol *symbol = TR::Symbol::createShadow(comp->trHeapMemory(), type);
+      char *fullName = (char *) comp->trMemory()->allocateHeapMemory((strlen(info->_name) + 1 + strlen(_name) + 1) * sizeof(char));
+      sprintf(fullName, "%s.%s", _name, info->_name);
+      TR::Symbol *symbol = TR::Symbol::createNamedShadow(comp->trHeapMemory(), type, info->_type->getSize(), fullName);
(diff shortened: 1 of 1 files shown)


andrewcraik:
looks good to me - always better to have better RAS

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -261,11 +261,9 @@ StructType::getFieldSymRef(const char *fieldName)
 
       TR::DataType type = info->getPrimitiveType();
 
-      TR::Symbol *symbol = NULL;
-      if (TR::Int32 == type)
-         symbol = comp->getSymRefTab()->findOrCreateGenericIntShadowSymbol();
-      else
-         symbol = TR::Symbol::createShadow(comp->trHeapMemory(), type);
+      char *fullName = (char *) TR::comp()->trMemory()->allocateHeapMemory((strlen(info->_name) + 1 + strlen(_name) + 1) * sizeof(char));
```
0xdaryl: Use the cached `comp` instead of the more expensive `TR::comp()`

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -381,7 +379,9 @@ UnionType::getFieldSymRef(const char *fieldName)
       auto symRefTab = comp->getSymRefTab();
       TR::DataType type = info->getPrimitiveType();
 
-      TR::Symbol *symbol = TR::Symbol::createShadow(comp->trHeapMemory(), type);
+      char *fullName = (char *) TR::comp()->trMemory()->allocateHeapMemory((strlen(info->_name) + 1 + strlen(_name) + 1) * sizeof(char));
```
0xdaryl: Use the cached `comp` instead of the more expensive `TR::comp()`

mstoodle:
Good catch on both problems, @0xdaryl . Thanks, and should be fixed now!

PR #895, opened 2017-03-20
Deprecate EnvironmentLanguageInterface etc

ktbriggs:
@charliegracie this is open for comments pending successful build/test.

Add MM_EnvironmentDelegate to MM_EnvironmentBase. Include GC_Environment
class in MM_EnvironmentDelegate to contain language-specific thread
local caches and stats, accessible to language support classes using 
MM_EnvironmentBase::getGCEnvironment().

Replace MM_EnvironmentLanguageInterface artifacts with 
MM_EnvironmentDelegate.

Delete EnvironmentLanguageInterface*.[ch]pp

Issue: #870


gc/base/EnvironmentBase.hpp
```diff
@@ -562,6 +589,49 @@ class MM_EnvironmentBase : public MM_BaseVirtual
...
+	 * points into global structures.
+	 */
+	MMINLINE GC_Environment *getGCEnvironment() { return _delegate.getGCEnvironment(); }
+
+	/**
+	 * Called on each participating GC thread before starting mark phase. This can be used
+	 * to reset local thread stats and reference buffers prior to commencing marking.
+	 */
+	MMINLINE void
+	markingPhaseStart()
```
charliegracie: Similar to the last set of delegate changes can you change the tense of the verb in the function name?.  These functions make it sound like they are performing this action when in fact they are getting called when the action is beginning or completed.  Can you also change the APIs in the delegate itself?  Thanks

example/glue/ConfigurationDelegate.hpp
```diff
@@ -117,8 +116,7 @@ class MM_ConfigurationDelegate
 	bool
 	environmentInitialized(MM_EnvironmentBase* env)
 	{
-		env->_envLanguageInterface = MM_EnvironmentLanguageInterfaceImpl::newInstance(env);
-		return (NULL != env->_envLanguageInterface);
```
charliegracie: Do you believe any delegates will still require this function?

gc/base/AllocateInitialization.hpp
```diff
@@ -1,277 +1,276 @@
-/*******************************************************************************
```
charliegracie: Is this change going to conflict with the change in #884?

ktbriggs: Yes. Rebasing now.

example/glue/ConfigurationDelegate.hpp
ktbriggs: Yes. Java delegate will require it.

gc/base/EnvironmentBase.hpp
ktbriggs: Methods renamed.

MM_EnvironmentBase:
  markingPhaseStart() -> markingPhaseStarted()
  markingPhaseFinish() -> markingPhaseFinished()
  
MM_EnvironmentDelegate: (omr/glue)
  markStart() -> markingStarted()
  markFinish() -> markingFinished

Similarly for other language environment delegates.

ktbriggs:
Will squash when requested changes in 264d446 have been reviewed.

example/glue/EnvironmentDelegate.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * (this file).
+ *
+ * OMR threads must acquire shared VM access and hold it as long as they are
+ * accessing VM internal structures. In the event that another thread requests
+ * exclusive VM access, threads holding shared VM access must release shared
+ * VM access and suspend activities involving direct access to VM structures.
+ *
+ * This (example) environment delegate implements a simplistic VM access framework
+ * that is not pre-emptive; that is, it relies on threads that are holding non-
+ * exclusive VM access to check frequently (sub-millisecond) whether any other
```
ktbriggs: Yup. What is really needed here is a preemptive framework that blocks mutator threads at safe points when an exclusive access request (for GC) is pending. Since we do not yet have a generic framework for preemption in omr, mutator threads must poll to check for this condition and, if the polling frequency is lower than a few ms, GC cycles will be held up longer than they should be.

So sub-millisecond is just a guideline and suggests some urgency around this, as is appropriate.  A lot can happen in a millisecond, and polling once or twice in that interval would introduce less slowdown than would a thread that did not poll and yield for >1 ms while a request for exclusive VM access was pending.

ktbriggs:
@charliegracie commit 16fbfb8 contains the changes requested in your comment. Will squash into previous commit when you have reviewed and approved these changes.

PR #896, opened 2017-03-20
Avoid printing guard's callee symbol if not specified

ncough:
Virtual guards may contain the symbol reference for the method they
guard, however, it is not guaranteed. This can be seen when creating
SideEffect guards in VP, where the guard is constructed using the
current block's start node, rather than a call node. However, when
dumping the CHTable, it was assumed that a guard would always
have this symbol reference specified.

### DIFF
--- a/compiler/ras/Debug.cpp
+++ b/compiler/ras/Debug.cpp
@@ -3615,8 +3615,9 @@ TR_Debug::dump(TR::FILE *pOutFile, TR_CHTable * chTable)
          char guardKindName[38];
          sprintf(guardKindName, "%s %s", getVirtualGuardKindName((*info)->getKind()), (*info)->mergedWithHCRGuard()?"+ HCRGuard ":"");
 
-         if ((*info)->getKind() == TR_OSRGuard)
-            trfprintf(pOutFile, "[%4d] %-38s OSRGuard\n", i, guardKindName);
+         if (!(*info)->getSymbolReference())
+            trfprintf(pOutFile, "[%4d] %-38s %s%s\n",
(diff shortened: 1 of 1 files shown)


compiler/ras/Debug.cpp
```diff
@@ -3615,8 +3615,9 @@ TR_Debug::dump(TR::FILE *pOutFile, TR_CHTable * chTable)
          char guardKindName[38];
          sprintf(guardKindName, "%s %s", getVirtualGuardKindName((*info)->getKind()), (*info)->mergedWithHCRGuard()?"+ HCRGuard ":"");
 
-         if ((*info)->getKind() == TR_OSRGuard)
-            trfprintf(pOutFile, "[%4d] %-38s OSRGuard\n", i, guardKindName);
+         if (!(*info)->getSymbolReference())
+            trfprintf(pOutFile, "[%4d] %-38s %s%s\n",
+                  i, guardKindName ,(*info)->isInlineGuard()?"inlined ":"", guardKindName);
```
mstoodle: why print out guardKindName twice on this line?

ncough: The prior behavior, which applied to only OSR Guards, was to repeat the guardKindName in the place of the callee symbol. This function is printing out a table, so I believe it is desirable to not leave a column empty.

andrewcraik: I'd agree - leaving the field blank would make it look like the print failed since side there is already code doing something similar I think keep it the same for readability of the logs is a good idea.

mstoodle: ok, fair enough.

PR #900, opened 2017-03-21
Add missing break statements

fjeremic:
Add two obviously missing break statements for CGFR and CGFRJ in
splitIntoCompareAndBranch and splitIntoCompareAndLongBranch,
respectively.


vijaysun-omr:
Clearly a case where having the breaks in question makes more sense. Merging.

PR #904, opened 2017-03-22
Crash when -Xjit:timingCumulative option is used

mpirvu:
The option -Xjit:timingCumulative is used to print ilGen/optimizer/codegen
times across all compilations. This is performed at shutdown time in a
static method OMR::Compilation::shutdown(TR_FrontEnd * fe).
However, this method internally makes use of the compilation object
retried from local thread storage: TR::comp(). This is invalid because
the shutdown() routine is not executed on a compilation thread.
There is another bug related to  TR_SingleTimer code: the function used
to print the time taken uses a static char buffer which makes this routine
not multithreaded safe. I have replaced this timeTakenString() method
with a new one, secondsTaken(), that can be used in a multithreaded
environment.

### DIFF
--- a/compiler/env/OMRVMEnv.cpp
+++ b/compiler/env/OMRVMEnv.cpp
@@ -107,13 +107,7 @@ static uint64_t highResClockResolution()
    }
 
 uint64_t
-OMR::VMEnv::getHighResClockResolution(TR::Compilation *comp)
-   {
-   return ::highResClockResolution();
-   }
-
-uint64_t
-OMR::VMEnv::getHighResClockResolution(OMR_VMThread *omrVMThread)
+OMR::VMEnv::getHighResClockResolution()
    {
    return ::highResClockResolution();
    }
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -1956,10 +1956,10 @@ OMR::Compilation::shutdown(TR_FrontEnd * fe)
    bool printCummStats = ((fe!=0) && TR::Options::getCmdLineOptions() && TR::Options::getCmdLineOptions()->getOption(TR_CummTiming));
    if (printCummStats)
       {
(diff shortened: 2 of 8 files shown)


compiler/env/OMRVMEnv.cpp
```diff
@@ -118,6 +118,12 @@ OMR::VMEnv::getHighResClockResolution(OMR_VMThread *omrVMThread)
    return ::highResClockResolution();
    }
 
+uint64_t 
+OMR::VMEnv::getHighResClockResolution(TR_FrontEnd *fe)
```
mstoodle: Do we need to add a front-end parameter here at all? Shouldn't there just be a getHighResClockResolution() call (no parameter) and everything should use that?

Would you entertain the notion to add this call without a parameter and modify your calls to use that, then do another quick PR to change the remaining calls to getHighResClockResolution() that pass alternate parameters? (I only see 6 other calls to the "comp()" one and none to the OMRVMThread one).

Or you could just add a second commit to clean those up in this PR...

mstoodle: I think that would actually simplify this commit too, since you wouldn't need two of everything or a forward declaration of TR_FrontEnd...

mpirvu: getHighResClockResolution() with no parameter does not exist. The function needs access to the portlib where the real implementation resides.

mpirvu:
As discussed, I changed the implementation so that getHighResClockResolution() and TR_SingleTimer::secondsTaken() do not need a parameter. In trj9 getHighResClockResolution() will retrieve the portLib from a global variable and from there it will find the clock resolution. In OMR getHighResClockResolution() will use a hardcoded value.

mstoodle:
Much better, but any chance we could also fix the name of the option to spell "cumulative" properly? :)

mpirvu:
Fixed the name of the option: timingCummulative --> timingCumulative

mstoodle:
Thanks @mpirvu . I'll merge once testing passes.

PR #905, opened 2017-03-22
Add read barrier flags

tajila:
Add read barrier flags

TASK 125560

reviewed by @charliegracie 

Signed-off-by: tobi <atobia@ca.ibm.com>


mgaudet:
So, just a heads up for future reference. Once you've opened a pull request, modifications can be made by simply force pushing to the same branch. 

This way you needn't close and re-open PR's for updates, and it will keep review comments when they arrive associated with the code, even as it's modified.

PR #906, opened 2017-03-22
Recovery on aborted Concurrent Scavenger

amicic:
If unable to copy object, we will return the original location of the
object in Evacuate. But to ensure that no other threads raced and
successfully copied the object  (and thus having two 'valid' version of
the object) the object will be self-forwarded. 

Self-forwarding is an atomic operation that sets forwarding bit same as
regular forwarding operation does. But self-forwarding will not change
the class information. It will set another bit (hole bit) to distinguish
from strictly forwarded objects.

Recovery from aborted Concurrent Scavenge is completely different (that it
is for regular Scavenger). There is no backout procedure per say.
Instead, we leave objects where they are in Nursery, and update
all reference slots that still point to strictly forwarded objects.
Then, percolate global GC will execute sliding compact 
that will move all objects in Nursery into low address range.

There are various other supporting changes, like:
- making heap object iterator aware of forwarding objects and 
- making Nursery unified space so that all objects in Survivor and
Evacuate are grouped together during the sliding compaction


gc/base/HeapMap.cpp
```diff
@@ -214,7 +214,7 @@ MM_HeapMap::numberBitsInRange(MM_EnvironmentBase *env, void *lowAddress, void *h
 	uintptr_t baseIndex, topIndex;
 	
 	/* Validate passed heap references */
-	Assert_MM_true(lowAddress < highAddress);
+	Assert_MM_true(lowAddress <= highAddress);
```
charliegracie: is lowAddress allowed to be equal to highAddress?

gc/base/MemoryPoolAddressOrderedListBase.hpp
```diff
@@ -48,6 +48,7 @@ class MM_MemoryPoolAddressOrderedListBase : public MM_MemoryPool
 	MM_HeapLinkedFreeHeader ** _referenceHeapFreeList;
 	MM_SweepPoolState* _sweepPoolState;	/**< GC Sweep Pool State */
 	MM_SweepPoolManagerAddressOrderedListBase* _sweepPoolManager;		/**< pointer to SweepPoolManager class */
+	MM_HeapLinkedFreeHeader *_lastFreeEntry;							/**< address of the last free entry in the pool; valid after compact; NOT maintained during allocation */
```
charliegracie: Does this make sense to move up to the super class instead of having the default case of NULL?

amicic: I can move up. I do not see big difference. It's probably here because I put it first here (since last entry makes sense only for ordered pools), but then I realized that CompactScheme deals with top level MemoryPool (although it is very order aware).

gc/base/HeapMap.cpp
amicic: With this change a part (Subspace) of the heap can be. When Nursery is unified (100% tilted) prior to compact, the other part of Nursery is literately 0%. If we happen to clear mark map during percolate global on the empty part we hit this assertion. 

An alternative is to keep this assertion on strict '<' and prevent even an attempt of clearing the mark map on empty subspace. I did try that but did not find an elegant way to do it. I can try again...

amicic: I restored the assert to the original, and added a check in the caller (MM_ConcurrentGC::determineInitWork) to skip empty region.

gc/base/standard/Scavenger.cpp
```diff
@@ -4311,6 +4537,9 @@ MM_Scavenger::mutatorFinalReleaseCopyCaches(MM_EnvironmentBase *envBase, MM_Envi
 {
 	MM_EnvironmentStandard *env = MM_EnvironmentStandard::getEnvironment(envBase);
 	MM_EnvironmentStandard *threadEnvironment = MM_EnvironmentStandard::getEnvironment(threadEnvironmentBase);
+//	OMRPORT_ACCESS_FROM_OMRPORT(env->getPortLibrary());
```
charliegracie: Did you mean to leave this commented code in here?  If so could it be wrapped under a DEBUG if defined or something.

gc/base/standard/Scavenger.cpp
```diff
@@ -4403,10 +4632,14 @@ MM_Scavenger::scavengeComplete(MM_EnvironmentBase *envBase)
 bool
 MM_Scavenger::scavengeIncremental(MM_EnvironmentBase *env, int64_t scavengeIncrementEndTime)
 {
+	Assert_MM_mustHaveExclusiveVMAccess(env->getOmrVMThread());
 	bool result = false;
 	bool timeout = false;
 
 	while (!timeout) {
+//		OMRPORT_ACCESS_FROM_OMRPORT(env->getPortLibrary());
```
charliegracie: Same as commented out code above

gc/base/MemoryPoolAddressOrderedListBase.hpp
charliegracie: No that is ok.  I agree that _lastFreeEntry really only makes sense on the MPAOL hierarchy.  In the future maybe the compact scheme could be switched so that it knows about the MPAOL hierarchy instead of the top level MP since it only works with MPAOL.  At that point the two APIs on MP could be removed.

amicic: agreed, this is a compromise between the correct solution and not too much changes within this PR

gc/base/standard/Scavenger.cpp
amicic: I intend to remove them (already did in my work space). If I have to modify this PR again for some other reasons, I'll do it as part of it, if not, I'll defer it for next PR (to avoid nowdays rather slow Travis compile)

PR #908, opened 2017-03-22
Unlock free spill list during out of line RA on P

ymanton:
Keeping the free list locked during RA on out of line code sections prevents any spills freed in from ever being re-used. For large methods with many out of line sections this causes us to allocate many spill slots that are ever only used once, leading to excessively bloated stack frames and metadata, and compilation failure if we exceed our compilation memory budget.

Locking the spill list is required for RA between the branch to and merge back from an out of line section, in order to insure that any unspilled registers in that path don't return their spill slots to the free list so we can re-unspill the register to the same location once we move to RA in the out of line path, but keeping the free list locked during RA in the out of line section appears to be completely unnecessary and detrimental. Registers can and should return their spill slots to the free list once unspilled out of line because they may never be used again, let alone spilled again, and will have no further opportunity of relinquishing their spill slots.

Regression tested against Java and by inspecting a few select compilation logs for the desired behaviour.

### DIFF
--- a/compiler/p/codegen/PPCInstruction.cpp
+++ b/compiler/p/codegen/PPCInstruction.cpp
@@ -102,18 +102,12 @@ void TR::PPCLabelInstruction::assignRegisters(TR_RegisterKinds kindToBeAssigned)
       TR_PPCOutOfLineCodeSection *oi = cg()->findOutLinedInstructionsFromLabel(getLabelSymbol());
       TR_ASSERT(oi, "Could not find PPCOutOfLineCodeSection stream from label.  instr=%p, label=%p\n", this, getLabelSymbol());
 
+      cg()->unlockFreeSpillList();
       if (!oi->hasBeenRegisterAssigned())
          oi->assignRegisters(kindToBeAssigned);
 
       if (cg()->getDebug())
          cg()->traceRegisterAssignment("OOL: Finished register assignment in OOL section\n");
-
-      // Unlock the free spill list.
-      //
-      // TODO: live registers that are not spilled at this point should have their backing
-      // storage returned to the free spill list.
-      //
-      cg()->unlockFreeSpillList();
       }
    }
 
@@ -220,17 +214,13 @@ void TR::PPCConditionalBranchInstruction::assignRegisters(TR_RegisterKinds kindT
       //
       TR_PPCOutOfLineCodeSection *oi = cg()->findOutLinedInstructionsFromLabel(getLabelSymbol());
       TR_ASSERT(oi, "Could not find PPCOutOfLineCodeSection stream from label.  instr=%p, label=%p\n", this, getLabelSymbol());
+
+      cg()->unlockFreeSpillList();
       if (!oi->hasBeenRegisterAssigned())
          oi->assignRegisters(kindToBeAssigned);
+
       if (cg()->getDebug())
             cg()->traceRegisterAssignment("OOL: Finished register assignment in OOL section\n");
-
-      // Unlock the free spill list.
-      //
-      // TODO: live registers that are not spilled at this point should have their backing
-      // storage returned to the free spill list.
-      //
-      cg()->unlockFreeSpillList();
       }
    }
 


mstoodle:
@zl-wang would you please review?

zl-wang:
when you unlock it at the start of OOL RA, it is unclear to me why you bother with to lock it to begin with. we should discuss.

ymanton:
Locking the free list and preventing "freed" spill slots from being returned to the list is a common design for all our codegens that implement OOL code sections.

My understanding of why is that we need to unspill registers on both sides of the diamond that's formed by OOL code, so any register that's unspilled on one side needs to be unspilled from the same location on the other. By locking the free list and preventing unspilled registers from returning their spill slots we prevent future spills from getting their hands on the spill slots before we've had a chance to unspill registers on the OOL side.

```
+           +
|           |
|           |
| branch +--------------------->+-----------+
|           |                   |           |
|           |                   | unspill X |
|           |                   |           |
| unspill X |                   | unspill Y |
|           |                   |           |
|           |                   |   spill Y |
|           |                   |           |
| label  <----------------------+-----------+
|           |
|   spill X |
|           |
|           |
+           +
```

Here's one example, where we are doing bottom-up RA and have to `spill X` at some point. Later during RA we `unspill X` between the branch to and merge back from the out of line code section. At this point we cannot allow X's spill slot to be made available to any later spills, like `spill Y`, because we have not yet done RA on the OOL side. The approach used by all of our codegens is to lock the free list to prevent `unspill X` from being able to return its spill slot to the free list.

The problem with this approach is that if we keep the list locked during OOL RA the second `unspill X` will also not return its slot to the free list and future spills will never be able to re-use it.

zl-wang:
the branched-to is OutOfLine section where spill/unspill happens for its being cold.  The main line (inlined hot path) is not expected to spill/unspill for performance reason.  That is my mental model.  Plus, it is not going to work if there are multiple branches to the same OOL section and if you allow spill/unspill in the between (i.e. different conditions going to the OOL):  e.g.  WriteBarrier has this pattern.

lock/unlockSpillList is for there being branches inside OOL section.  We can discuss further.

ymanton:
> the branched-to is OutOfLine section where spill/unspill happens for its being cold. The main line (inlined hot path) is not expected to spill/unspill for performance reason.

We have to unspill in the main line code path if there is a use of the virtual register. We don't want or expect to **spill** in the main code path, but we can't prevent **unspilling**, that is simply a consequence of a spill that already happened. The problem is that an unspill in the main line code path obligates us to also unspill in the out of line code path which is what forces us to do this locking/unlocking business.

> That is my mental model. Plus, it is not going to work if there are multiple branches to the same OOL section and if you allow spill/unspill in the between (i.e. different conditions going to the OOL): e.g. WriteBarrier has this pattern.

This is an existing limitation in all of our codegens, OOL sections only cleanly support diamond control flow, otherwise lots of assumptions will be broken. If you have multiple out of line branches you have to make sure nothing goes wrong on a case by case basis.

I think these issues are all criticisms of the existing OOL implementation, which I agree with, but they exist independently of this PR.

ymanton:
Spoke to @zl-wang offline, and he suggested that we think about requiring that no RA happen in the main line path between the branch to and merge back (similar to how we deal with ICF today). This is contrary to the original OOL design goal of having no constraints on RA in the main line path and doing the right thing in the out of line path, but that has proven to be a difficult thing to get right.

Eliminating the possibility of unspills on the main line path would reduce the complexity of this whole scheme by allowing us to get rid of the free spill list lock, various checks in our spilling/unspilling paths, and so on. The cost is that we extend the live range of some virtuals and probably introduce more RA action in the out of line path, but that might be an acceptable trade-off. Anyone who has an interest in codegen and OOL implementations feel free to comment on this idea.

Anyhow, none of that prevents this change from being merged to allow the implementation we currently have to deal with excessive spilling without blowing up, so I believe @zl-wang is agreeable to this being merged and can give his OK.

zl-wang:
Ready to be merged as temporary solution.

PR #909, opened 2017-03-22
Fix CCR register assignment bug on P

ymanton:
This PR contains two patches, the first of which fixes a rare bug in the P register assigner, the other which cleans up formatting and variable names in surrounding code to better follow conventions.

In order to fix the register assigner we make sure to assign CCRs before GPRs if and when required when assigning registers for dependencies. This is to avoid the case where the dependency assigns and therefore blocks every single available GPR prior to assigning any CCR. If any CCRs are in the spilled state and must be reverse spilled we will require an unblocked GPR to use as a scratch register, since on Power we can't load/store a physical CCR directly.

Regression tested on Java and by inspection of a few logs to insure the expected behaviour.

### DIFF
--- a/compiler/p/codegen/OMRRegisterDependency.cpp
+++ b/compiler/p/codegen/OMRRegisterDependency.cpp
@@ -633,8 +633,8 @@ findDependencyChainHead(TR::RegisterDependency *dep,
    }
 
 static void assignFreeRegisters(TR::Instruction              *currentInstruction,
-                                TR::RegisterDependency    *dep,
-                                TR_PPCRegisterDependencyMap& map,
+                                TR::RegisterDependency       *dep,
+                                TR_PPCRegisterDependencyMap&  map,
                                 TR::CodeGenerator            *cg)
    {
    TR::Machine *machine = cg->machine();
@@ -653,22 +653,20 @@ static void assignFreeRegisters(TR::Instruction              *currentInstruction
    }
 
 static void assignContendedRegisters(TR::Instruction              *currentInstruction,
-                                     TR::RegisterDependency    *dep,
-                                     TR_PPCRegisterDependencyMap& map,
-                                     bool                         depsBlocked,
+                                     TR::RegisterDependency       *dep,
+                                     TR_PPCRegisterDependencyMap&  map,
+                                     bool                          depsBlocked,
                                      TR::CodeGenerator            *cg)
    {
    TR::Machine *machine = cg->machine();
 
    dep = findDependencyChainHead(dep, map);
 
-
    TR::Register *virtReg = dep->getRegister();
    TR::RealRegister::RegNum targetRegNum = dep->getRealRegister();
    TR::RealRegister *targetReg = machine->getPPCRealRegister(targetRegNum);
    TR::RealRegister *assignedReg = virtReg->getAssignedRealRegister() ?
       toRealRegister(virtReg->getAssignedRealRegister()) :  NULL;
-
 
    // Chain of length 1
    if (!assignedReg || !map.getDependencyWithTarget(assignedReg->getRegisterNumber()))
@@ -677,6 +675,7 @@ static void assignContendedRegisters(TR::Instruction              *currentInstru
       virtReg->block();
       return;
       }
+
    // Chain of length 2, handled here instead of below to get 3*xor exchange on GPRs
    if (map.getDependencyWithTarget(assignedReg->getRegisterNumber()) == map.getDependencyWithAssigned(targetRegNum))
       {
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
I would like @zl-wang and @fjeremic to review this.

zl-wang:
Is there any downside to assign CCRs (or FPRs/VRs/VSRs) first, no matter blockGPRs is true or not?

blockYYY appeared to mean requiring-not-full-set of YYY.  Maybe naming them accordingly to reduce confusion with register blocked state.

we should not have spilled CCRs at all, since CCR is not a result visible to optimizer (no reuse).  Having spilled CCRs is purely a side-effect of artificially-made-longer live-range of CCRs. Loop-ending condition testing and AsyncCheckMsg testing ... I suspected it is the only special case where two live CCRs exist at the same time. If we throw away that special case, brave enough we may be able to skip spilling/unspilling CCRs completely.

ymanton:
> Is there any downside to assign CCRs (or FPRs/VRs/VSRs) first, no matter blockGPRs is true or not?

No functional downside, I just wanted to save some compilation time since the condition is so rare. We are iterating through the dependencies to count the number of GPRs/FPRs/etc there is an opportunity to detect this situation and avoid an extra trip through the dependencies if it is not needed.

> blockYYY appeared to mean requiring-not-full-set of YYY. Maybe naming them accordingly to reduce confusion with register blocked state.

I've changed the names of those variables to `haveSpareYYY` to make it clearer as to what is going on.

> we should not have spilled CCRs at all, since CCR is not a result visible to optimizer (no reuse). Having spilled CCRs is purely a side-effect of artificially-made-longer live-range of CCRs. Loop-ending condition testing and AsyncCheckMsg testing ... I suspected it is the only special case where two live CCRs exist at the same time.

That's right, most of the cases I saw were due to our evaluators re-using virtual CCRs across helper calls. It is hard to avoid this if you also have internal control flow and are forced to extend the live range of your working registers.

> If we throw away that special case, brave enough we may be able to skip spilling/unspilling CCRs completely.

Yes I tried this for the stuff I've been working on and it has the desired effect. It can be done in private linkage since OMR is not affected.

compiler/p/codegen/OMRRegisterDependency.cpp
```diff
@@ -844,74 +844,86 @@ void TR_PPCRegisterDependencyGroup::assignRegisters(TR::Instruction   *currentIn
...
+               ++numGPRs;
+               break;
+            case TR_FPR:
+               ++numFPRs;
+               break;
+            case TR_VRF:
+               ++numVRFs;
+               break;
+            case TR_CCR:
+               if (!virtReg->getAssignedRegister())
```
fjeremic: Is this condition enough? In an OOL section if a virtual register is not assigned to a real register it may not always mean the register is spilled. What is this is the first time we are encountering this virtual register in an OOL path but the future use count != total use count. That is, what if the register is used in the mainline hot path but is dead at the merge label of the OOL path? In that case the `virtReg->getAssignedRegister()` will return `NULL` however the virtual register is not spilled. It has just never been assigned.

Small nitpick as well: `getAssignedRegister` does not return a boolean, so shouldn't the if be checking for `virtReg->getAssignedRegister() == NULL` (I really wish we supported `nullptr`).

ymanton: > Is this condition enough? In an OOL section if a virtual register is not assigned to a real register it may not always mean the register is spilled.

That's true, it's not accurate, and `haveSpilledCCRs` should really be `mayHaveSpilledCCRs` if the check stays as-is, but a false positive is not harmful here because we can always assign CCRs before GPRs without penalty (see my response to @zl-wang).

The most accurate/least costly check I can think of given the current RA implementation would be `getAssignedRegister() != NULL && getBackingStorage() != NULL`.

> Small nitpick as well: `getAssignedRegister` does not return a boolean, so shouldn't the if be checking for `virtReg->getAssignedRegister() == NULL` (I really wish we supported `nullptr`).

Personal preference but yes you are right.

ymanton: Fixed both issues.

fjeremic: `getAssignedRegister() != NULL && getBackingStorage() != NULL` is effectively what we use on z Systems. This should work fine. Thanks for the followup.

fjeremic:
LGTM (as far as my Power expertise goes).

vijaysun-omr:
@zl-wang are you okay with these changes ?

zl-wang:
yes, i am ok with it.

PR #911, opened 2017-03-22
Add runtime flag for concurrent scavenge support

fjeremic:
Concurrent scavanger related code enabling the code generator to emit
read barriers for object reference loads is currently guarded by a
compile time OMR_GC_CONCURRENT_SCAVENGER define.

This change effectively deprecates the OMR_GC_CONCURRENT_SCAVENGER flag
in the JIT in favor of a runtime check. This is a pre-emptive change in
preparation of full deprecation of the OMR_GC_CONCURRENT_SCAVENGER flag.
However this removal has to come from the OMR GC front and a
correpsonding runtime flag for concurrent scavanger needs to be
introduced in the GC. Only at this point may we remove the final
reference to OMR_GC_CONCURRENT_SCAVENGER in the JIT.

### DIFF
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -420,11 +420,11 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
    {
    TR::Compilation *comp = cg->comp();
 
-   if (cg->supportsConcurrentScavange())
+   if (cg->isConcurrentScavengeEnabled())
       {
       // TODO (GuardedStorage): We need to come up with a better solution than anchoring tree tops of compressedrefs
       // sequences to enforce certain evaluation order. Perhaps not lowering compressedrefs in the first place when
-      // concurrent scavange is supported is the correct solution here.
+      // concurrent scavenge is supported is the correct solution here.
       traceMsg(comp, "GuardedStorage: in performSetupForInstructionSelectionPhase\n");
 
       auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -419,39 +419,43 @@ void
 OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
    {
    TR::Compilation *comp = cg->comp();
-   // TODO (GuardedStorage)
-#if defined(OMR_GC_CONCURRENT_SCAVENGER)
-   traceMsg(comp, "GuardedStorage: in performSetupForInstructionSelectionPhase\n");
 
-   auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
-
-   std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop*, TR::TreeTop*>, TR::Allocator> >
-      currentTreeTopToappendTreeTop(std::less<TR::TreeTop*> (), mapAllocator);
-
-   for (TR::PreorderNodeIterator iter(comp->getStartTree(), comp); iter != NULL; ++iter)
+   if (cg->supportsConcurrentScavange())
       {
-      TR::Node *node = iter.currentNode();
+      // TODO (GuardedStorage): We need to come up with a better solution than anchoring tree tops of compressedrefs
+      // sequences to enforce certain evaluation order. Perhaps not lowering compressedrefs in the first place when
+      // concurrent scavange is supported is the correct solution here.
+      traceMsg(comp, "GuardedStorage: in performSetupForInstructionSelectionPhase\n");
 
-      traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
+      auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
 
-      if ((comp->useCompressedPointers() && node->getOpCodeValue() == TR::l2a) || (!comp->useCompressedPointers() && node->getOpCodeValue() == TR::aloadi))
+      std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop*, TR::TreeTop*>, TR::Allocator> >
+         currentTreeTopToappendTreeTop(std::less<TR::TreeTop*> (), mapAllocator);
+
+      for (TR::PreorderNodeIterator iter(comp->getStartTree(), comp); iter != NULL; ++iter)
          {
-         TR::TreeTop* anchorTreeTop = TR::TreeTop::create(comp, TR::Node::create(TR::treetop, 1, node));
-         TR::TreeTop* appendTreeTop = iter.currentTree();
+         TR::Node *node = iter.currentNode();
 
-         if (currentTreeTopToappendTreeTop.count(appendTreeTop) > 0)
+         traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
+
+         if ((comp->useCompressedPointers() && node->getOpCodeValue() == TR::l2a) || (!comp->useCompressedPointers() && node->getOpCodeValue() == TR::aloadi))
             {
-            appendTreeTop = currentTreeTopToappendTreeTop[appendTreeTop];
+            TR::TreeTop* anchorTreeTop = TR::TreeTop::create(comp, TR::Node::create(TR::treetop, 1, node));
+            TR::TreeTop* appendTreeTop = iter.currentTree();
+
+            if (currentTreeTopToappendTreeTop.count(appendTreeTop) > 0)
+               {
+               appendTreeTop = currentTreeTopToappendTreeTop[appendTreeTop];
+               }
+
+            // Anchor the l2a before the current treetop
+            appendTreeTop->insertBefore(anchorTreeTop);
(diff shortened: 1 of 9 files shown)


compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -494,6 +494,18 @@ class OMR_EXTENSIBLE CodeGenerator
...
    bool supportsMergingOfHCRGuards() {return false;} // no virt, default
 
+   /** \brief
+    *     Determines whether this code generator supports concurrent scavange of objects during garbage collection.
+    *
+    *  \return
+    *     true if the code generator will emit read barriers for loads of object references from the heap in support
+    *     of concurrent scavange; false otherwise.
+    */
+   bool supportsConcurrentScavange()
```
mstoodle: Typo, I think: isn't it scavenge not scavange?

Also, I wonder about the choice of "supports" as the verb here. This name suggests a question like "can this code generator operate with concurrent scavenge?" rather than "should I generate code for concurrent scavenge?". Maybe concurrentScavengeEnabled() or something like that?

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -485,19 +485,21 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
    */
 
-   // TODO (GuardedStorage)
-   TR::Register *source;
-   if (comp->useCompressedPointers() && (TR::Compiler->om.compressedReferenceShift() == 0 || firstChild->containsCompressionSequence()) && !node->isl2aForCompressedArrayletLeafLoad())
-      cg->setEvalCompressionSequence(true);
+   TR::Register* source = NULL;
 
-   if ((TR::Compiler->target.cpu.getS390SupportsGuardedStorageFacility()) && comp->useCompressedPointers() && (firstChild->containsCompressionSequence()) && !node->isl2aForCompressedArrayletLeafLoad())
+   // TODO (GuardedStorage): Why is isl2aForCompressedArrayletLeafLoad check necessary?
```
mstoodle: This doesn't look like a straight-forward refactoring to me. It appears that the condition for compressedReferenceShift() == 0 has been eliminated, and consequently you've sunk the calls to setEvalCompressionSequence() so they no longer wrap around the code to evaluate the first child. Is there a reason the shift==0 condition is no longer required?

Side and probably irrelevant question because I have no idea what effect setEvalCompressionSequence() actually has. Is it impossible for there to be another l2a nested underneath the grandchild? On its face, it looks dangerous to me to simply setEvalCompressionSequence(false) on the way out if a nested l2a is possible, but there are lots of reasons why that could be perfectly correct.

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Fixed in c5ec84e and 18d6ac5.

compiler/z/codegen/UnaryEvaluator.cpp
fjeremic: This is a bug. I was under the impression that `containsCompressionSequence` would return true for `compressedReferenceShift() == 0` and hence I thought the check was redundant. To my surprise this is not true and the check is needed. Fixed in a4bdb5b.

As an aside I'm curious if anyone can answer why a `compressedReferenceShift() == 0` tree is not defined as a compression sequence? It technically is a compression sequence with a shift of zero.

fjeremic: As for the other part of the question I do not think this can currently happen. All compression sequences are anchored under `TR::compressedRefs` tree anchors, and following the compressedrefs rematerialization phase they are expanded into the compression / decompression sequences. 

Though I do share your concern in that something like lower trees could decide to sink a reference down to a grandchild in which case we would encounter the scenario you speak of. This problem is however avoided with a pre-instruction selection pass which anchors all `TR::l2a` trees to their own treetops. See:

https://github.com/eclipse/omr/blob/master/compiler/codegen/OMRCodeGenPhase.cpp#L423-L454

This whole idea needs to be reworked. I do not think we should be rematerializing `TR::compressedRefs` trees to begin with and avoiding to generate the compression sequence. The guarded load instruction models the load and compression together, and as such we should not explicitly represent the compression in the trees. This is work that I will be doing next.

fjeremic: I'm a rather cautious person though when it comes to assumptions so just in case during this transition period as part of a4bdb5b I have made the change to use a counter based mechanism instead of setting a flag to avoid any potential nested scenarios.

mstoodle: +1 for caution :) . But maybe easier to store the current state before you set it to true, and then restore it to the previous state rather than just set to false? fewer moving parts...

fjeremic: > +1 for caution :) . But maybe easier to store the current state before you set it to true, and then restore it to the previous state rather than just set to false? fewer moving parts...

I think I want to keep the depth counter intact to reuse for debugging. I want to place some extra debugging checks to see if the nested scenario happens and if it does in what scenario and how often. We are currently having a lot of trouble debugging issues where we miss to generate a guarded load somewhere in a JIT body and we end up loading a stale pointer during concurrent scavange. These are not easy problems to debug unfortunately as the failure can happen much further down than where the missed guarded load occurred.

fjeremic:
@ebadyano given your expertise in this area could I please get a review from you?

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -419,39 +419,43 @@ void
...
-#if defined(OMR_GC_CONCURRENT_SCAVENGER)
-   traceMsg(comp, "GuardedStorage: in performSetupForInstructionSelectionPhase\n");
 
-   auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
-
-   std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop*, TR::TreeTop*>, TR::Allocator> >
-      currentTreeTopToappendTreeTop(std::less<TR::TreeTop*> (), mapAllocator);
-
-   for (TR::PreorderNodeIterator iter(comp->getStartTree(), comp); iter != NULL; ++iter)
+   if (cg->isConcurrentScavengeEnabled())
```
0xdaryl: I think this code needs to make a distinction between whether concurrent scavenge
is SUPPORTED by a particular code generator versus whether it is ENABLED at all.

The question that you want to be asking at this point in the code is whether
concurrent scavenge is both ENABLED and SUPPORTED, unless I am totally misreading
what you're trying to do here.  Looking into how this was actually implemented in
`OMR::Z::CodeGenerator::isConcurrentScavengeEnabled`, this query only returns
true if a particular Z processor feature is available, as opposed to whether
concurrent scavenge is actually enabled by the GC in this VM.

Also, as written, this code path will always be taken on Z regardless of the GC
policy in place.  Was that your intent?

0xdaryl: I am also wondering about the value of this code in this (very) common phase in
the code generator.  As written, it would be very easy to sink all this logic into
its own codegen phase that would only be enabled on Z (should that be the right
thing to do).  Can you look into that?

fjeremic: @0xdaryl as stated in the commit message this is a pre-emptive change. The GC currently does not have a runtime flag which determines whether concurrent scavenge is enabled or not, i.e. the query which determines the concurrent scavenge GC policy.

As such the `isConcurrentScavengeEnabled` is currently guarded by a compile time flag. When GC implements a runtime query for the concurrent scavenge GC policy `isConcurrentScavengeEnabled` will change accordingly. The "ENABLED" part you mention above will be implemented as such.

The "SUPPEROTED" part is implmented already by `getS390SupportsGuardedStorageFacility` query which checks for processor support. The value returned by this call is a cached value obtained from the port library by extracting STFLE bits from the processor. Once the GC implements their runtime check they too will query the port library in a similar fashion so the GC and JIT are both synchronized with the value each respective library returns for a "isConcurrentScavengeEnabled" query.

Hope this clears things up a bit. Did I answer your question?

fjeremic: I can definitely sink this down into a z Systems specific codegen phase. Though given that Power also supports concurrent scavenge I suspect we want something like this for Power as well. I am questioning the existence of this code to begin with however. I think a better solution which I want to explore is to remove this code altogether (as noted in the comment near this snippet) and avoid rematerializing `TR::compressedRefs` trees into decompression sequences as this more closely models what our hardware instruction does (it loads and shifts all in one go so we shouldn't really model the load before the shift in the IL).

Given my explanation please recommend whether you would like this pushed into a z Systems specific codegen phase.

fjeremic:
@mstoodle @0xdaryl @ebadyano any further comments / actions you can suggest?

mstoodle:
Sorry for the delay @fjeremic . I would have preferred 18d6ac5 be folded into the original commit, but that's a relatively minor point I'm willing to accept. I have no further concerns.

fjeremic:
@mstoodle I wanted to keep the commits intact during review process so once can gauge the progression of the PR when addressing reviews. Once all reviewers are happy with the state of the changes I will squash all my commits so as to not pollute the repo. Thanks for your review!

mstoodle:
@fjeremic very considerate, thank you !

ebadyano:
@fjeremic lgtm

fjeremic:
@0xdaryl any other concerns?

compiler/codegen/OMRCodeGenPhase.cpp
0xdaryl: The haze I was in was due to the fact that I didn't realize that the downstream implementer of this technology only provided *one* GC policy (i.e., concurrent scavenge) in effect at a time.  If this work is on the path to not make that restriction (and ultimately contribute to Eclipse OMR) then I'm fine with it.

0xdaryl: Agreed re/ your better solution.  I look forward to seeing that in a future PR.  :-)

PR #913, opened 2017-03-23
Compilation Time Reduction

dsouzai:
This PR contains various changes to improve Compilation CPU Time.

### DIFF
--- a/compiler/optimizer/LocalDeadStoreElimination.cpp
+++ b/compiler/optimizer/LocalDeadStoreElimination.cpp
@@ -1139,9 +1139,6 @@ bool TR::LocalDeadStoreElimination::examineNewUsesForKill(TR::Node *node, TR::No
 
 void TR::LocalDeadStoreElimination::killStoreNodes(TR::Node *node)
    {
-   TR::SparseBitVector usedef_vector(comp()->allocator());
-   node->getSymbolReference()->getUseDefAliases().getAliasesWithClear(usedef_vector);
-
    for (auto it = _storeNodes->begin(); it != _storeNodes->end(); ++it)
       {
       TR::Node *storeNode = *it;
@@ -1151,8 +1148,7 @@ void TR::LocalDeadStoreElimination::killStoreNodes(TR::Node *node)
          TR::SymbolReference *storeSymRef=storeNode->getSymbolReference();
 
          // TODO: improve by not killing stores that are definitely disjoint
-         // if (node->getSymbolReference()->getUseDefAliases().contains(storeSymRef, comp()))
-         if (usedef_vector[storeSymRef->getReferenceNumber()])
+         if (node->getSymbolReference()->getUseDefAliases().contains(storeSymRef, comp()))
         *it = NULL;
          }
       }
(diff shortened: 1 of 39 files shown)


dsouzai:
@mpirvu  Could you please review :)

dsouzai:
Travis CI failed because the of the following linter complaint:

```
../../compiler/x/codegen/Instruction.hpp:40:7: error: Implicit this receivers are prohibited in extensible classes
      setOpCodeValue(op);
      ^
      self()->
```

However, I don't believe this is an error. I explicitly removed the self() calls from Instruction.hpp specifically because at the TR level, the `this` pointer is already the most derived type (unless I'm misunderstood, @0xdaryl could you verify that?). The reason I removed it because in the process of inlining OMR::Instruction::self(), I came across this use of self() in a hpp file. I'm not sure if it is ok to use self() in Instruction.hpp unless by that point TR::Instruction is a complete type.

mgaudet:
This is a weakness in the linter. I'm going to open an item to address it.

dsouzai:
After talking to Matthew, I will update the PR with changes that will make the linter happy; while the current changes are correct, merging them would result in all future PRs also failing the linter until the issue is fixed.

mpirvu:
I am ok with the code.
Do you know why the new implementation for TR::LocalDeadStoreElimination::killStoreNodes(TR::Node *node) is faster/better?

dsouzai:
The "new" implementation is just reverting to what the code did before the `TR::SparseBitVector usedef_vector` container code was added. Creating and using the container first involves going through the process of allocating memory and copying over the bits from the  `node->getSymbolReference()->getUseDefAliases()` bit vector so there is a local copy of the bit vector. Given that killStoreNodes iterated through the list of store nodes and only used the bit vector to check whether it should kill a node or not, the cost of manually checking (via `node->getSymbolReference()->getUseDefAliases().contains(storeSymRef, comp())`) was still cheaper than having to create the container and copy over the bits.

compiler/optimizer/LocalDeadStoreElimination.cpp
```diff
@@ -1151,8 +1148,7 @@ void TR::LocalDeadStoreElimination::killStoreNodes(TR::Node *node)
          TR::SymbolReference *storeSymRef=storeNode->getSymbolReference();
 
          // TODO: improve by not killing stores that are definitely disjoint
-         // if (node->getSymbolReference()->getUseDefAliases().contains(storeSymRef, comp()))
-         if (usedef_vector[storeSymRef->getReferenceNumber()])
+         if (node->getSymbolReference()->getUseDefAliases().contains(storeSymRef, comp()))
```
vijaysun-omr: Since you essentially reversed the usage of sparse bit vector here and went back to the original way of doing the check, I thought I would ask if you know why a sparse bit vector usage was added here, i.e. it would be good to know what was the case where it was thought to be beneficial.

dsouzai: From the looks of it, it was added before we split and released OMR; it was done by another project where the change actually improved CPU time for them considerably.

dsouzai:
@vijaysun-omr I think it should be ok to accept the merge request :)

vijaysun-omr:
Okay thanks for the explanation. Merging.

Issue #902, opened 2017-03-21
Mac OS X uses MAP_ANON instead of MAP_ANONYMOUS


rednaxelafx:
I was trying to play with the JitBuilder examples on Mac OS X 10.9.
Building JitBuilder following instructions from Page 30 of Mark's January OMR JIT talk:
```
$ git clone https://github.com/eclipse/omr
$ cd omr && ./configure [ --disable-warnings-as-errors ]
$ cd jitbuilder && make # builds libjitbuilder.a
$ cd release && make test # builds and runs examples
```
... leads me to this error: 
```
In file included from ../jitbuilder/env/FrontEnd.cpp:30:
../compiler/env/FEBase_t.hpp:94:42: error: use of undeclared identifier 'MAP_ANONYMOUS'
                           MAP_PRIVATE | MAP_ANONYMOUS,
                                         ^
```
I guess we'd have to do conditional compilation here for OS X.

Side question: what degree of support should I expect from OMR/TR on Mac OS X as of now?

mstoodle:
Hey @rednaxelafx ,

Sorry, those directions aren't perfect, I'm afraid. I've got a blog article in the works with the real directions (which include working on OS X, note the SPEC setting):

    $ git clone https://github.com/eclipse/omr
    $ cd omr && make -f run_configure.mk SPEC=osx OMRGLUE=./example/glue
    $ cd jitbuilder && make
    $ cd release && make
    $ make test | testall

I work on OS X myself for all my JitBuilder and OMR development (unless something is Linux-specific), so everything should work just fine on OS X. And if you find something that doesn't work, please open an issue!

mstoodle:
Mind you, I'm running 10.12 ... Please post back here if it still doesn't work with the updated instructions.

rednaxelafx:
Hi Mark @mstoodle ,

Cool! Thanks for your reply.

I did a fresh clone of the `omr` repo and followed your updated instructions. Unfortunately it still fails the same way on my machine.
I wonder if it has anything to do with OS X version that I'm running on. As I said I'm running on a fairly old version, 10.9 (x86_64-apple-darwin13.4.0), and compiling with an old Clang (3.5.x) as well.

Looking forward to your next installment of the JitBuilder blog series!

Best regards,
Kris Mok

mstoodle:
[...]

PR #916, opened 2017-03-23
Use MAP_ANON when MAP_ANONYMOUS not available

mstoodle:
This commit is a temporary requirement for JitBuilder and other
OMR compiler based projects to be able to build on older versions
of OS X where MAP_ANONYMOUS has not been defined. This problem
came up in Issue #902 while building JitBuilder and the OMR compiler
component on Mac OS X 10.9 . In these kinds of environments, MAP_ANON
is usually defined and so we can simply define MAP_ANONYMOUS to
be MAP_ANON where we need it. The port library already does this
swizzling but we have not yet hooked up a depenedency for the
compiler or JitBuilder on the port library to benefit from that
code.

### DIFF
--- a/compiler/env/FEBase_t.hpp
+++ b/compiler/env/FEBase_t.hpp
@@ -80,6 +80,19 @@ FEBase<Derived>::allocateCodeMemory(TR::Compilation *comp, uint32_t warmCodeSize
    return warmCode;
    }
 
+// This code does not really belong here (along with allocateRelocationData, really)
+// We should be relying on the port library to allocate memory, but this connection
+// has not yet been made, so as a quick workaround for platforms like OS X <= 10.9,
+// where MAP_ANONYMOUS is not defined, is to map MAP_ANON to MAP_ANONYMOUS ourselves
+#if !defined(MAP_ANONYMOUS)
+  #define NO_MAP_ANONYMOUS
+  #if defined(MAP_ANON)
+    #define MAP_ANONYMOUS MAP_ANON
+  #else
+    #error unexpectedly, no MAP_ANONYMOUS or MAP_ANON definition
+  #endif
+#endif
+
 template <class Derived>
 uint8_t *
 FEBase<Derived>::allocateRelocationData(TR::Compilation* comp, uint32_t size)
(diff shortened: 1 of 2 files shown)


compiler/env/FEBase_t.hpp
```diff
@@ -80,6 +80,19 @@ FEBase<Derived>::allocateCodeMemory(TR::Compilation *comp, uint32_t warmCodeSize
...
+// This code does not really belong here (along with allocateRelocationData, really)
+// We should be relying on the port library to allocate memory, but this connection
+// has not yet been made, so as a quick workaround for platforms like OS X <= 10.9,
+// where MAP_ANONYMOUS is not defined, is to map MAP_ANON to MAP_ANONYMOUS ourselves
+#if !defined(MAP_ANONYMOUS)
+  #define NO_MAP_ANONYMOUS
+  #if defined(MAP_ANON)
+    #define MAP_ANONYMOUS MAP_ANON
+  #else
+    #define MAP_ANONYMOUS 0
```
charliegracie: Does it make sense to have a #error here instead since we are are not sure what to do if both MAP_ANONYMOUS and MAP_ANON are not defined.

jitbuilder/runtime/JBCodeCacheManager.cpp
```diff
@@ -72,6 +72,18 @@ JitBuilder::CodeCacheManager::allocateCodeCacheSegment(size_t segmentSize,
...
    {
+   // We should really rely on the port library to allocate memory, but this connection
+   // has not yet been made, so as a quick workaround for platforms like OS X <= 10.9
+   // where MAP_ANONYMOUS is not defined, is to map MAP_ANON to MAP_ANONYMOUS ourselves
+   #if !defined(MAP_ANONYMOUS)
+      #define NO_MAP_ANONYMOUS
+      #if defined(MAP_ANON)
+         #define MAP_ANONYMOUS MAP_ANON
+      #else
+         #define MAP_ANONYMOUS 0
```
charliegracie: Same as my first comment

compiler/env/FEBase_t.hpp
mstoodle: Yeah, you're probably right; that's probably a more appropriate default than to silently do nothing. I'll update with your suggestion.

Thanks @charliegracie !

mstoodle: @charliegracie fix (and rebase) has been pushed.

jitbuilder/runtime/JBCodeCacheManager.cpp
mstoodle: Should be fixed now. Thanks.

PR #918, opened 2017-03-23
Get extended block whilst walking in OSR check

ncough:
Rather than performing the walk back to the start of the current block
to determine whether it is currently in an extended block and then
repeating the walk to perform its check, both can be done in a single
pass.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -712,8 +712,7 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
          if (node->getReferenceCount() > 1)
             {
             TR::TreeTop *cursor = tt->getPrevTreeTop();
-            TR::TreeTop *extendedBlockStart = tt->getNode()->getBlock()->startOfExtendedBlock()->getEntry();
-            while (cursor && cursor != extendedBlockStart)
+            while (cursor)
                {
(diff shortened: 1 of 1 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -712,15 +712,17 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
+            while (cursor)
                {
                if ((cursor->getNode()->getOpCode().isCheck() || cursor->getNode()->getOpCodeValue() == TR::treetop)
                    && cursor->getNode()->getFirstChild() == node)
                   {
                   potentialOSRPoint = false;
                   break;
                   }
+               if (cursor->getNode()->getOpCodeValue() == TR::BBStart &&
+                   cursor == cursor->getNode()->getBlock()->startOfExtendedBlock()->getEntry())
```
andrewcraik: You want to test isExtensionOfPreviousBlock here - startOfExtendedBlock will walk back across blocks to return to you the treetop. You stop searching once BBStart says no to isExtensionOfPreviousBlock

andrewcraik:
LGTM

mstoodle:
Your commit title is longer than 50 characters as requested in our contribution guidelines. Can you please shorten it?

andrewcraik:
New title looks good - back to @vijaysun-omr following the requested update being made.

vijaysun-omr:
Looks like a safe change to me to avoid the extra walk over trees to get to extended block start. Merging.

PR #919, opened 2017-03-23
Detect more CPU features on X86

0dvictor:
Allowing CG to query supports for:
SSE3
SSSE3
AVX
AVX2
FMA
CLMUL

This commit also include some cleanup work:
1. Removed duplicated TR_X86ProcessorFlags2, TR_X86ProcessorFeatures8, TR_ProcessorDescription;
2. Added more feature flags to their corresponding enums for future use.

### DIFF
--- a/compiler/env/ProcessorInfo.hpp
+++ b/compiler/env/ProcessorInfo.hpp
@@ -94,6 +94,7 @@ enum TR_X86ProcessorFeatures
    TR_MachineCheckException         = 0x00000080,
    TR_CMPXCHG8BInstruction          = 0x00000100,
    TR_APICHardware                  = 0x00000200,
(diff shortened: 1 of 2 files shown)


mstoodle:
Your commit message should also mention you're removing the duplicate TR_X86ProcessorFlags enum and friends from x/codegen/OMRCodeGenerator.hpp .

I would also say that the title of the commit message is misleading (the commit message more accurately describes the change).

andrewcraik:
Commit message and description aside the move of the flags out of the OMRCodeGenerator.hpp seems to be in keeping with the intended design and adding flags for features people will want to check for seems reasonable.

0dvictor:
Thank you both for the review. The commit message has been updated.

PR #920, opened 2017-03-24
Fix makefile dependency for JitBuilder tests

Leonardo2718:
This commit updates the makefile to ensure that GTest is built before
the JitBuilder tests that depend on it.


mstoodle:
About as straight-forward as it gets. Thanks @Leonardo2718 !

PR #921, opened 2017-03-24
Extend remat API to analyse beyond one block

ncough:
The remat API has several functions that require a start and end
node to be specified. In this form, they are limited to a block
or a continuous set of blocks, as it is not possible to specify
which edges should be taken. This change extends the API to include
a BitVector of allowable blocks. When these functions reach the
end of their current block, the first allowable block in the
set of successors will be taken, allowing for more aggressive
remat.

### DIFF
--- a/compiler/optimizer/RematTools.cpp
+++ b/compiler/optimizer/RematTools.cpp
@@ -20,6 +20,7 @@
 #include "il/ILOps.hpp"                                   // for ILOpCode, etc
 #include "il/Node_inlines.hpp"
 #include "il/TreeTop_inlines.hpp"
+#include "il/Block.hpp"
 
 void RematSafetyInformation::add(TR::TreeTop *argStore, TR::SparseBitVector &symRefDependencies)
    {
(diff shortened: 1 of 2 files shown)


andrewcraik:
The copying of the blocks to visit bitvector is not ideal. It seems to be trying to handle cycles, but cycles in the middle of an extended block sequence are invalid - the only possible safe merge point is the start of the sequence so this code could probably get cleaned up and made more efficient to exploit that fact and add additional checking under debug and prod with assumes to validate the block sequence. Even under prod a cheap check that each block you are processing has a single predecessor is probably warranted (if violated we should return that no remat is possible and just bail).

compiler/optimizer/RematTools.cpp
```diff
@@ -290,7 +389,7 @@ void RematTools::walkTreeTopsCalculatingRematFailureAlternatives(TR::Compilation
          traceMsg(comp, "  priv arg remat: visiting [%p]: isStore %d privArg %d failedTargetMatch %d", start->getNode(),
             start->getNode()->getOpCode().isStoreDirect(),
             start->getNode()->getSymbol() && start->getNode()->getSymbol()->isAuto(),
-            failedTargets.ValueAt(start->getNode()->getFirstChild()->getGlobalIndex()));
+            start->getNode()->getNumChildren() ? failedTargets.ValueAt(start->getNode()->getFirstChild()->getGlobalIndex()) : -1);
```
andrewcraik: can we please make the > 0 explicit here since that is the intention - it isn't truly doing a truth test.

compiler/optimizer/RematTools.cpp
```diff
@@ -242,42 +303,80 @@ void RematTools::walkNodesCalculatingRematSafety(TR::Compilation *comp,
...
    }
-   
+
 /*
  * This method walks the nodes between the start treetop inclusive and
  * the end treetop exclusive. When it encounters a node in the scanTargets
  * it adds the symref of the scan target to its set of interest and continues
  * scanning. Any symrefs killed after they have been added to the set of interest
  * are added to the unsafeSymRefs vector for later use
+ * Returns true to indicate it successfully completed the analysis.
```
andrewcraik: Please add a comment describing what should be in blocksToVisit for a single block walk and a multi-block walk.

andrewcraik:
LGTM - over to @vijaysun-omr

vijaysun-omr:
I don't have all the context for all the routines that were touched here but @andrewcraik does. From my one review of the logic, it seems reasonable what is being done here. Merging.

PR #923, opened 2017-03-24
minor fixes/RAS improvements to fsd

cathyzhyi:
3 changes are in this PR
1. Add more info to assert msg in OSRDefAnalysis 
2. Fix TR_OSRCompilationData::addInstruction arg type 
    The intention of the loop in this API is just walking up the call stack but not change the bcInfo. Change reference type to normal type to prevent accidentally change the argument value.
3. Delete redundant exception edge copying
     The code's intention was to avoid exception edge copying for OSR blocks. This line was intended to be deleted.

### DIFF
--- a/compiler/compile/OSRData.hpp
+++ b/compiler/compile/OSRData.hpp
@@ -119,7 +119,7 @@ class TR_OSRCompilationData
                            int32_t symRefOrder, int32_t symSize, bool takesTwoSlots);
    void ensureSlotSharingInfoAt(const TR_ByteCodeInfo& T);
    void addInstruction(TR::Instruction* instr);
-   void addInstruction(int32_t instructionPC, TR_ByteCodeInfo &bcInfo);
+   void addInstruction(int32_t instructionPC, TR_ByteCodeInfo bcInfo);
 
    void checkOSRLimits();
 
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -91,7 +91,7 @@ TR_OSRCompilationData::addInstruction(TR::Instruction* instr)
    }
 
 void
-TR_OSRCompilationData::addInstruction(int32_t instructionPC, TR_ByteCodeInfo &bcInfo)
+TR_OSRCompilationData::addInstruction(int32_t instructionPC, TR_ByteCodeInfo bcInfo)
    {
    bool trace = comp->getOption(TR_TraceOSR);
    if (trace)
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -4966,7 +4966,6 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
             callerCFG->copyExceptionSuccessors(blockContainingTheCall, n, succAndPredAreNotOSRBlocks);
          else
             debugTrace(tracer(),"\ndon't add exception edges from callee OSR block(block_%d) to caller catch blocks\n",n->getNumber());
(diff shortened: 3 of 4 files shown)


andrewcraik:
can you coordinate with @ncough about getting the query into the compiler to say whether OSR is voluntary or involuntary as well as whether it is before or after? I don't want tests for FSD serving as a proxy for the OSR mode. We merged one yesterday because I believed @ncough was about to pull request his change to add the API. If he is about to do that then can we get that one merged and then rebase this on top of that? If not can we please add that API as he intends as part of this and also fix the FSD test added yesterday?

compiler/compile/OSRData.hpp
```diff
@@ -119,7 +119,7 @@ class TR_OSRCompilationData
                            int32_t symRefOrder, int32_t symSize, bool takesTwoSlots);
    void ensureSlotSharingInfoAt(const TR_ByteCodeInfo& T);
    void addInstruction(TR::Instruction* instr);
-   void addInstruction(int32_t instructionPC, TR_ByteCodeInfo &bcInfo);
+   void addInstruction(int32_t instructionPC, TR_ByteCodeInfo bcInfo);
```
andrewcraik: Please explain why a reference is wrong - your justification here is the type is wrong - it isn't wrong you are choosing to change from a ref to an actual bcInfo. Is it that we can possibly modify it?

andrewcraik:
Can you explain why the FSD check being where it was was not a 'proper' disable? It seems like we should not have been able to remove edges with where it was placed.

andrewcraik:
I think the title of the PR could be improved based on the answers to the questions posed. One of these is not an issue - it is a RAS improvement...

compiler/compile/OSRData.hpp
cathyzhyi: Hey Andrew, sry the explanation was in the commit msg and I copied them to the comments of this PR. The intention of the loop in this API is just walking up the call stack but not change the bcInfo. Change reference type to normal type to prevent accidentally changing the argument value.

cathyzhyi:
@andrewcraik I just talked to @ncough and the "Properly disable TR_OSRExceptionEdgeRemoval in fsd" could be removed after his refactoring. It was needed because of the setCanAffordOSRControlFlow at the start of the method.

compiler/compile/OSRData.hpp
andrewcraik: Ah ok - that makes sense

andrewcraik:
ok - so if you are going to remove that commit from the PR then the rest can go in and he will clean up the rest when he pull requests shortly (all being well).

cathyzhyi:
removed the commit that's not needed.

compiler/compile/OSRData.hpp
vijaysun-omr: While I agree with the risk of modification point, would we end up copying the TR_ByteCodeInfo structure more now (at parameter passing) that the reference type has been removed ?

andrewcraik: Yes we would. The external API to the rest of the compiler should not be a reference because bcInfo is modified in the iterative loop inside the implementation on line 139 of OSRData.cpp. The implementation internally is iterative so I don't think it makes sense to create an impl method that takes a reference which is internal. I believe the copy is required and is a necessary safety mechanism - otherwise the bcInfo in the caller context will be modified by the search being performed which is not the desired behavior and would be entirely surprising.

vijaysun-omr: Okay I did not get that the value was actually being changed (at line 139) until your response to my question. I agree with the change in that case since the caller side value should not be changed just because of the iteration over the call stack in this routine.

PR #924, opened 2017-03-24
Add extension point constrainRecognizedMethod in VP

liqunl:
This commit contains the following changes:

1) Add extension point as home to language specific optimizations on
recognized methods.

2) Return if the node is not a call after the transformation.

3) Make doDelayedTransformations virtual so that it can be overriden in
   order to host language specific optimizations that have to be done
   near the end of VP.

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.hpp
+++ b/compiler/optimizer/OMRValuePropagation.hpp
@@ -457,7 +457,7 @@ class ValuePropagation : public TR::Optimization
    void launchNode(TR::Node *node, TR::Node *parent, int32_t whichChild);
 
    bool checkAllUnsafeReferences(TR::Node *node, vcount_t visitCount);
-   void doDelayedTransformations();
+   virtual void doDelayedTransformations();
 
    void replacePackedArrayLoad(TR::Node *loadNode, TR::Node *packedNode, TR::Node *curNode, vcount_t visitCount);
 
@@ -606,6 +606,7 @@ class ValuePropagation : public TR::Optimization
    void transformUnknownTypeArrayCopy(TR_TreeTopWrtBarFlag *);
    void transformReferenceArrayCopy(TR_TreeTopWrtBarFlag *);
    void transformReferenceArrayCopyWithoutCreatingStoreTrees(TR_TreeTopWrtBarFlag *arrayTree, TR::SymbolReference *srcObjRef, TR::SymbolReference *dstObjRef, TR::SymbolReference *srcRef, TR::SymbolReference *dstRef, TR::SymbolReference *lenRef);
+   virtual void constrainRecognizedMethod(TR::Node *node);
 
    struct ObjCloneInfo {
       TR_ALLOC(TR_Memory::ValuePropagation)
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -5237,6 +5237,11 @@ TR::Node *constrainCall(OMR::ValuePropagation *vp, TR::Node *node)
          }
       }
 
+   vp->constrainRecognizedMethod(node);
+
+   // Return if the node is not a regular call (xcall/xcalli) anymore
+   if (!node->getOpCode().isCall() || node->getSymbol()->castToMethodSymbol()->isHelper())
+      return node;
 
    if ( symbol )
       {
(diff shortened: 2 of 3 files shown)


liqunl:
@andrewcraik Could you review my change? There are two places checking whether the node is still a call, I wonder if we should move them into the same commit. Personal build is running, will put the result once it's finished.

andrewcraik:
isCall answers yes to many nodes that are not actually of the form xcall or xcalli (for example arraycopy, dcos etc). What is the limitation we are trying to impose on VP at this point because your comments seem to say on xcall or xcalli, but the query is much more general.

liqunl:
Yes, I'm trying to avoid non xcall/xcalli entering the code after constrainCall/transformRecognizedMethod, basically only allow whatever opcodes that are constrained by constrainCall/constrainACall to be able to continue. I guess I need another check to say isNonHelperCall.

liqunl:
Or we should just return if any transformation has done to the node.

liqunl:
@andrewcraik Chang is updated.

liqunl:
@andrewcraik What do you think of the new change?

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -606,6 +606,7 @@ class ValuePropagation : public TR::Optimization
    void transformUnknownTypeArrayCopy(TR_TreeTopWrtBarFlag *);
    void transformReferenceArrayCopy(TR_TreeTopWrtBarFlag *);
    void transformReferenceArrayCopyWithoutCreatingStoreTrees(TR_TreeTopWrtBarFlag *arrayTree, TR::SymbolReference *srcObjRef, TR::SymbolReference *dstObjRef, TR::SymbolReference *srcRef, TR::SymbolReference *dstRef, TR::SymbolReference *lenRef);
+   virtual void transformRecognizedMethod(TR::Node *node) { }
```
andrewcraik: This deserves some kind of doxygen comment about what it is, what it is used for, where it is called from when etc. It should probably be implemented in a cpp file rather than the header.

liqunl: Documentation is added in the new commit which will be merged with other commits before deliver.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2161,8 +2161,7 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
           && base->isNonNullObject()
           && base->getClassType()
           && base->getClassType()->asFixedClass()
-          && base->getClass()
-          && TR::Compiler->cls.isClassInitialized(vp->comp(), base->getClass()))
```
andrewcraik: My question remains as to why it is safe to remove this check - can you please comment with your justification?

liqunl: It's explained in my first commit message. And I'll forward you an email from Devin with more details.

liqunl: This change is removed from this PR.

liqunl:
@andrewcraik Changes have been made to the commit, could you please review again?

andrewcraik:
LGTM

liqunl:
Personal build is passed.

liqunl:
@vijaysun-omr Could you review and merge when you have time?

liqunl:
WIP this item because new change might need to handle HCR.

liqunl:
@andrewcraik Changes have been made according to our discussion yesterday, could you review again? Hopefully this is the last review for this PR.

andrewcraik:
LGTM from me

liqunl:
@vijaysun-omr Can you review and merge when you have time?

liqunl:
@andrewcraik constrainRecognizedMethod() will be called before devirtualizeCall, I wonder if we should move it to after devirtualizeCall. Haven't looked deep into the effect the position could have, putting the comment here so that we don't merge the PR before we get clear on that.

andrewcraik:
If we are going to do special processing on a method by recognizing it I believe that means that we either know a) that it won't be a virtual call or b) there will be only 1 implementation. I guess moving it later would ensure that. @liqunl will your changes still work if the code is moved to after devirtualize call? It may provide more opportunities for it to kick in...

liqunl:
Moving it later will benefit constrainRecognizedMethod. Tested with one of my VP opts and it still works.

liqunl:
@andrewcraik Any more concerns from you? If not we can get this merged.

andrewcraik:
LGTM

liqunl:
@vijaysun-omr Andrew and I have arrived an conclusion that this change is in a good shape, could you review and merge it when you get the time?

vijaysun-omr:
Looks okay to me. Merging.

PR #925, opened 2017-03-24
Update z Systems processor IDs

fjeremic:
Fix a TODO to update he z Systems processor ID for zNext and introduce
the zNexts processor and its corresponding ID.


compiler/env/ProcessorInfo.hpp
```diff
@@ -46,9 +46,9 @@ enum TR_S390MachineType
    TR_ZEC12_RESERVE           =  2829,  // reserve for zHelix
    TR_Z13                     =  2964,  // z13
    TR_Z13s                    =  2965,  // z13s
+   TR_ZNEXT                   =  3906,  // zNext
+   TR_ZNEXTs                  =  3907,  // zNexts
```
vijaysun-omr: Just for my information, what is the difference between TR_ZNEXT and TR_ZNEXTs in this context ?

fjeremic: I'm not sure what the "s" is shorthand for, perhaps _slim_? For z13 for example, z13s is defined as a "smaller footprint" z13. More information here:

http://www-03.ibm.com/systems/z/hardware/z13s.html

PR #926, opened 2017-03-24
Extend OSR API to represent its different modes

ncough:
This change adds checks that abstract over the different OSR
modes, specifically whether it is operating in voluntary
or involuntary OSR mode and whether it will insert the OSR
transition before or after the OSR point.

Moreover, it adds the concept of OSR analysis points, which
exist only to hold OSR information at certain bytecode indices and will
not become a transition point.

A further change is partial de-aliasing of the osrInfrastructureRemoved
and the canAffordOSRControlFlow flags. canAffordOSRControlFlow is used
to indicate whether the additional compile time of OSR is afforadable,
whilst osrInfrastructureRemoved indicates that OSR exception edges
have been removed. However, canAffordOSRControlFlow was being used
to achieve the latter in some cases.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -586,6 +586,15 @@ bool OMR::Compilation::canAffordOSRControlFlow()
       return false;
       }
 
+   if (self()->osrInfrastructureRemoved())
+      {
+      if (self()->getOption(TR_TraceOSR))
+         {
+         traceMsg(self(), "canAffordOSRControlFlow returning false due to removal of OSR infrastructure\n");
+         }
+      return false;
+      }
+
    return _canAffordOSRControlFlow;
    }
 
@@ -597,6 +606,13 @@ void OMR::Compilation::setSeenClassPreventingInducedOSR()
 
 bool OMR::Compilation::supportsInduceOSR()
    {
+   if (_osrInfrastructureRemoved)
+      {
+      if (self()->getOption(TR_TraceOSR))
+         traceMsg(self(), "OSR induction cannot be performed after OSR infrastructure has been removed\n");
+      return false;
+      }
+
    if (!self()->canAffordOSRControlFlow())
       {
       if (self()->getOption(TR_TraceOSR))
@@ -622,13 +638,6 @@ bool OMR::Compilation::supportsInduceOSR()
       {
       if (self()->getOption(TR_TraceOSR))
          traceMsg(self(), "Cannot guarantee OSR transfer of control to the interpreter will work for calls preventing induced OSR (e.g. Quad) because of differences in JIT vs interpreter representations\n");
-      return false;
-      }
-
-   if (_osrInfrastructureRemoved)
-      {
-      if (self()->getOption(TR_TraceOSR))
-         traceMsg(self(), "OSR induction cannot be performed after OSR infrastructure has been removed\n");
       return false;
       }
 
@@ -659,7 +668,7 @@ bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node)
    static char *disableMonentOSR = feGetEnv("TR_disableMonentOSR");
 
    bool potentialOSRPoint = false;
-   if (self()->getHCRMode() == TR::osr)
+   if (self()->getOSRTransitionTarget() == TR::postExecutionOSR)
       {
       if (_osrInfrastructureRemoved)
          potentialOSRPoint = false;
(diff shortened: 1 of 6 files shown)


compiler/compile/OSRData.hpp
```diff
@@ -44,6 +44,21 @@ namespace TR
       osr,
       traditional
       };
+  enum OSRTransitionTarget
```
andrewcraik: Can you please add some descriptive comments on these enums describing what they are and what the different values means since these are now very important in understanding the OSR infrastructure

compiler/compile/OMRCompilation.cpp
```diff
@@ -740,38 +749,78 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
    return potentialOSRPoint;
    }
 
+/*
+ * There are two OSR modes: voluntary and involuntary OSR.
+ * In voluntary OSR mode, the JIT is aware of where OSR transitions will occur
+ * and can manage them. In involuntary OSR mode, the OSR transition point is not known
+ * and can occur at any point.
```
andrewcraik: This should read something more like: In involuntary OSR mode, the JITed code does not control when an OSR transition occurs - a transfer can be initiated externally at any potential OSR point.

compiler/compile/OSRData.hpp
```diff
@@ -44,6 +44,55 @@ namespace TR
...
+
+/*
+ * OSR can operate in two modes, voluntary and involuntary.
+ *
+ * In voluntary OSR, the JIT is aware of the OSR points and controls
+ * the placements of their transitions, limiting the necessary book keeping
+ * to them.
+ *
+ * In involuntary OSR, the transition can occur at any bytecode index, 
+ * requiring the book keeping for a transition at any time.
```
andrewcraik: Again we need to be clear that involuntary is just that a transition can occur at any potentialOSR point - the generated code does not determine if the transition will take place.

compiler/compile/OMRCompilation.cpp
```diff
@@ -740,38 +749,81 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
+      return false;
 
    switch (node->getOpCodeValue())
       {
-      // Monents only require a trailing OSR point as they will perform OSR when executing the
-      // monitor and there is no change in liveness due to the monent
-      case TR::monent: return false;
-      // Calls require leading and trailing OSR points as liveness may change across them
-      default: return true;
+      case TR::monent:
```
andrewcraik: shouldn't the comment be preserved?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1268,19 +1261,16 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
             if (doOSR)
                {
                TR_ASSERT(comp->getOSRCompilationData(), "OSR compilation data is NULL\n");
-               if (comp->canAffordOSRControlFlow())
+               self()->genAndAttachOSRCodeBlocks(comp->getCurrentInlinedSiteIndex());
```
andrewcraik: Could you provide the logic behind removing the canAffordOSRControlFlow query here - we won't end up generating more code after this change than we did before right?

ncough: The addition of [osrInfrastructureRemoved](https://github.com/eclipse/omr/commit/955f4104b8f026fdb21c8047c746586e22399061) added a call to supportsInduceOSR() to doOSR, which will check both canAffordOSRControlFlow and osrInfrastructureRemoved. The calls between the evaluation of doOSR and the check create an ILGen optimizer and detecting cycles in the CFG. Neither of these will modify the flags as canAffordOSRControlFlow is never modified and osrInfrastructureRemoved is only changed if OSRExceptionEdgeRemoval is performed. Therefore, the check was redundant.

andrewcraik: Ok makes sense - thanks!

andrewcraik:
LGTM now, but needs a squash before merging since one commit is only comments that clarify the first commit. @vijaysun-omr this one is a pre-req on several other things @ncough wants to merge so if you could review when you get a moment that would be appreciated.

ncough:
Squashed the commits and rebased.

compiler/compile/OMRCompilation.cpp
```diff
@@ -735,38 +744,85 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
...
       {
-      switch (node->getOpCodeValue())
-         {
-         case TR::monent: return 1;
-         case TR::asynccheck: return 0;
-         default: return 3;
-         }
+      case TR::monent: return 1;
+      case TR::asynccheck: return 0;
+      default: return 3;
```
vijaysun-omr: What does the "default" mean here wrt returning 3 ?

vijaysun-omr: I mean, why is that a valid value for every other opcode ?

vijaysun-omr:
My question as part of the code review should not hold up the delivery of this PR since the logic was the same before this change and my question was related more to why the logic is what it is.

compiler/compile/OMRCompilation.cpp
andrewcraik: That is a good point - this should be case case case and default should assert that you have an unexpected opcode. @ncough FYI since one of us should fix this.

PR #927, opened 2017-03-27
Small JitBuilder fixes

mstoodle:
A few small fixes for JitBuilder:
1. put some exposed ilgen tracing code under option traceIlGen
2. eliminate a bunch of compiler warnings due to const return values in OMRDataTypes.hpp.cpp
3. added virtual destructors to TR_IlGenerator, IlInjector, IlBuilder, MethodBuilder, ThunkBuilder, IlType, PointerType, StructType, UnionType

### DIFF
--- a/compiler/il/OMRDataTypes.cpp
+++ b/compiler/il/OMRDataTypes.cpp
@@ -307,7 +307,7 @@ static int32_t OMRDataTypeSizes[] =
 
 static_assert(TR::NumOMRTypes == (sizeof(OMRDataTypeSizes) / sizeof(OMRDataTypeSizes[0])), "OMRDataTypeSizes is not the correct size");
 
-const int32_t
+int32_t
 OMR::DataType::getSize(TR::DataType dt)
    {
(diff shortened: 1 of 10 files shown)


mstoodle:
Note that the IlBuilder hierarchy does use virtual functions, and MethodBuilder can be allocated outside of the JIT framework, so it really should have a virtual destructor. Possibly we could get away with only a virtual descructor in that class, but I prefer the whole hierarchy to have it. TR::IlType can also be used outside of a compilation, even though its allocations are via persistent new.

mstoodle:
Argh. I managed to fold part of another change into one of these commits. I will update and push again (and will comment here when it's *really* ready).

mstoodle:
Ok, extraneous partial commit removed, so this one should be good to go again.

mstoodle:
i'm not sure that constexpr on top of inline buys us all that much for these particular queries, it won't help in one case (getSize whose implementation isn't in the header file), and since we haven't used them anywhere else I'm suspicious of their support level although I haven't verified.

mstoodle:
commit comment reworded to preserve decorum :) .

PR #928, opened 2017-03-27
Supporting options for heuristics related to LPQ

mpirvu:
LPQ is a secondary queue used to store compilation requests of lower
priority. This change set introduces 3 new options:
-Xjit:concurrentLPQ  ==>  Compilation requests from LPQ are allowed to
proceed in parallel with compilations from main queue
-Xjit:earlyLPQ ==> Compilation requests from LPQ are allowed to happen
early, during startup
-Xjit:suspendCompThreadsEarly ==> Suspend compilation threads when
compilation queue weight drops under a threshold. The alternative is
to suspend compilation threads when there is nothing else in the queue.


mstoodle:
@mpirvu Can you please shorten the commit title to less than 50 characters as described in our contributing guidelines, and also add a bit of explanation for what these options mean in the commit message? I know the commit itself provides a description, but it would be nice to have that text in the commit message as well.

mpirvu:
Shorter title and more verbose commit message. Note that the options are useful only in tr.open.

vijaysun-omr:
Just adding new options for controlling LPQ. No change in behaviour yet.

PR #929, opened 2017-03-27
Allow TypeDictionary reuse across compilations

mstoodle:
TypeDictionary cannot currently be reused across multiple
compilations because it caches TR::SymbolReference pointers
for field references of struct and union types. These pointers
only have meaning in a single compilation.

This commit adds NotifyCompilationDone() to TypeDictionary
which iterates over all defined StructTypes and UnionTypes
and clears their _symRef field back to 0. It also
reinitializes the symrefBV bit vector in each UnionType.
Finally, compileMethodBuilder will call NotifyCompilationDone
on the supplied MethodBuilder's TypeDictionary just before
returning.

To test, I modified the StructArray test, which used to
allocate two identical TypeDictionaries for use to compile
two different methods, so that it reuses a single
TypeDictionary. I then ran structarray 10,000 times without
seeing any crash.

### DIFF
--- a/compiler/ilgen/TypeDictionary.cpp
+++ b/compiler/ilgen/TypeDictionary.cpp
@@ -126,10 +126,11 @@ public:
 
    void cacheSymRef(TR::SymbolReference *symRef) { _symRef = symRef; }
    TR::SymbolReference *getSymRef()              { return _symRef; }
+   void clearSymRef()                            { _symRef = NULL; }
 
    TR::IlType *getType()                         { return _type; }
 
-   TR::DataType getPrimitiveType()             { return _type->getPrimitiveType(); }
+   TR::DataType getPrimitiveType()               { return _type->getPrimitiveType(); }
 
    size_t getOffset()                            { return _offset; }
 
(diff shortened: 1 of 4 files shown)


mgaudet:
The issue here is that now you can recycle a type dictionary, but it's also now thread-unsafe....

mgaudet:
(to avoid criticizing without suggesting a way forward; Maybe what needs to happen is Type Dictionaries need to be made compilation-state-less, with a separate state object being created by the compilation instead, which gets boiled by processing a TypeDictionary)

mstoodle:
TypeDictionary has always been thread unsafe :) . I agree that longer term it would be better if SymbolReferenceTable was the thing caching the symbol references, but that probably requires sinking the functionality of the guts of TypeDictionary into OMR::SymbolReferenceTable, leaving TypeDictionary to be just an API.

In the meantime, I suppose clients who want to have multiple compilation threads should create a TypeDictionary for each thread (which is at least better than having to create a TypeDictionary for every compilation, as they do now).

mstoodle:
Also note that it's actually StructType and UnionType that cache symbol references, not TypeDictionary itself.

mstoodle:
However, making TypeDictionary thread-safe is more than just not caching symbol references. TypeDictionary manages all kinds of data structures that are not thread-safe.

mgaudet:
Fair enough: I also don't know whether or not a compilation will mutate a type dictionary.

PR #930, opened 2017-03-27
Enable on-demand method lookup for JitBuilder

mstoodle:
Currently if JitBuilder clients want to Call a function using either
Call or ComputedCall, that function must have been described to
JitBuilder via the DefineFunction call on the MethodBuilder object.
All Calls lookup the function name provided and will fail to call it
if DefineFunction has not yet been called. The only way to make sure
the function has been defined is to make sure you call it yourself
before that function name is called.

This commit adds one check inside the Call services so that, if the
first lookup fails, it will call a virtual instance on MethodBuilder
called RequestFunction. Clients can implement this instance function
in their subclass of MethodBuilder to locate and call DefineFunction
for the requested function name. When RequestFunction() comes back,
the Call service will try to look the function up again and will
proceed if it has now been defined. This approach frees JitBuilder
clients from having to define all possible functions up front, which
may not be practical. Instead, required functions can be defined
on demand.

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1720,6 +1720,10 @@ IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, ...)
    TR::IlValue **argValues = processCallArgs(_comp, numArgs, args);
    va_end(args);
    TR::ResolvedMethod *resolvedMethod = _methodBuilder->lookupFunction(functionName);
+   if (resolvedMethod == NULL && _methodBuilder->RequestFunction(functionName))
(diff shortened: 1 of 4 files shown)


mstoodle:
there's not much to it to test, but I know what you mean @Leonardo2718 .

I'll come up with a sample for it.

mstoodle:
Apologies for the delay in responding to the request for a sample here; I have been swamped by non-OMR related work. I'll try to get this one wrapped up by the end of the week.

mstoodle:
I modified the Switch code sample to show how to use RequestFunction. There wasn't any really good code sample to choose, so I picked a very simple one. Not ideal, but writing a whole new code sample just for this service felt a bit silly.

There are still lots of examples for the "other" way to do it.

PR #931, opened 2017-03-27
Remove unnecessary condition from ificmpeqSimplifier

JamesKingdon:
ificmpeqSimplifier contains an additional condition in one of its
if statements compared to the otherwise similar ificmpneSimplifier
function, and the extra test has a negative impact on the performance
of the generated code. This change deletes the extra condition to
make the two functions behave the same way.

Additionally, uses of secondChild->getInt in the affected conditional
block were changed to secondChild->getLongInt.

Signed-off-by: James Kingdon <jkingdon@ca.ibm.com>


andrewcraik:
This change looks good to me - the key point to note is that storing a node's const value writes the value to a 64bit field in the back data store. For narrower types the narrowed value is widened to 64bits. Since both these values are positive the widening used does not matter (signed vs unsigned) and so equality comparison to 0 or 1 will work - it will just correctly handle cases where a long constant has only 0 or 1 in its low half and something other than 0 in its high half.

andrewcraik:
@JamesKingdon please shorten your title - this one is 52 characters which exceeds the 50 characters required in the contribution guidelines.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12721,17 +12721,17 @@ TR::Node *ificmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
 
    bitwiseToLogical(node, block, s);
 
-   if (firstChild->getOpCode().isBooleanCompare() && !firstChild->getOpCode().isLongCompare() &&
+   if (firstChild->getOpCode().isBooleanCompare() &&
        (secondChild->getOpCode().isLoadConst()) &&
-       ((secondChild->getInt() == 0) || (secondChild->getInt() == 1)) &&
+       ((secondChild->getLongInt() == 0) || (secondChild->getLongInt() == 1)) &&
```
vijaysun-omr: secondChild is the second child of ificmpeq and so must be an iconst to have a type correct tree; given this, why is getInt not the right api to call ?

andrewcraik: Vijay is right here - this might have been brain fog on my part - the compare does return int so getInt is right.

JamesKingdon: Ok, I'll revert those calls to getInt.

vijaysun-omr:
Thanks for making the change suggested in the previous review. Merging.

PR #934, opened 2017-03-27
Fix JitBuilder broken recursive method name test

mstoodle:
Hugely silly test to check for a recursive call to the function
being compiled only used strncmp (with the length of the current
function) without also checking that the name to lookup has the same
length as the current method's name (oops). So looking up "mandelhelp"
would actually short circuit the lookup if you were compiling "mandel"
because, well, the substring matched :( . Unanticipated recursions ensue...

This commit adds a first check that the names have the same
length before using strncmp with one of the string's lengths.


vijaysun-omr:
Simple change. Merging now that checks have passed.

PR #936, opened 2017-03-28
Add opcode ibyteswap and X86 evaluator

liqunl:
Add ibyteswap opcode to reverse the byte order of a 32-bit integer:
bits 0 to 7 are swapped with bits 24 to 31, and bit 8 to 15 are swapped
with bits 16 to 23. An X86 evaluator is also added in this change.

### DIFF
--- a/compiler/arm/codegen/TreeEvaluatorTable.hpp
+++ b/compiler/arm/codegen/TreeEvaluatorTable.hpp
@@ -917,4 +917,6 @@
 	TR::TreeEvaluator::longNumberOfTrailingZeros,
 	TR::TreeEvaluator::longBitCount,
 
+   TR::TreeEvaluator::unImpOpEvaluator,                 // TR::ibyteswap
+
    TR::TreeEvaluator::NOPEvaluator,         // Temporarily using for TR::Prefetch
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -1038,6 +1038,21 @@ OMR::CodeGenerator::getSupportsTLE()
    return self()->getSupportsTM();
    }
 
+/** \brief
+ *     A query about whether ibyteswap is supported
+ *
+ *  \return
+ *     True if ibyteswap is supported in codegen
+ *
+ *  \note
+ *     Override the query on the platform where ibyteswap is implemented
+ */
+bool
+OMR::CodeGenerator::getSupportsIbyteswap()
+   {
+   return false;
+   }
+
 bool
 OMR::CodeGenerator::getSupportsConstantOffsetInAddressing(int64_t value)
    {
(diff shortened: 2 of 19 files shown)


liqunl:
@vijaysun-omr This is the change for ibyteswap, tested with personal build, can you review and merge it?

andrewcraik:
I'm not sure this is safe without the opcode being added in general is it? what happens for languages without ibyteswap?

0xdaryl:
I'd also like to see a more informative commit message.  Please indicate that this is a J9-specific change.

liqunl:
@andrewcraik The opcode and x86 evaluator have been moved to OMR.

liqunl:
@0xdaryl I moved the opcode to OMR, it is not j9 specific now. I also modified the commit message to contain more details.

liqunl:
@ymanton @fjeremic Could you or someone on your team work on the ibyteswap evaluator on your platform? This is needed for an optimization that's to be delivered after this change. The x86 evaluator wil probably go in first so that we can start the review process for the optimization.

andrewcraik:
Until all codegens support it the optimization will need to check for the correct platform in some way, but I'm sure you are already looking into sorting that.

liqunl:
Yes, already told James that his opt has to be limited to x86 until all codegens support it.

andrewcraik:
Will the limit be done with a codegen support query? If so should that support query be added to this PR since without it the opcode is not much use...

liqunl:
Yes, a query will be added to this PR.

liqunl:
@andrewcraik I added a codegen support query to return false in omr codegen and overrided it in x86 codegen to return true.

ymanton:
@R-Santhir, this happens to be similar to something you're doing.

vijaysun-omr:
This looks okay to me and the checks have passed. Anything more you would like done @andrewcraik ?

andrewcraik:
LGTM from me - it will be dormant until someone starts generating it which I assume will come in another PR. We should also add some tests for the opcode, but we can get this in for now since there will be other changes depending on it that need to go up for review.

0xdaryl:
1)  I'd like to see a 64-bit version as well for completeness: `lbyteswap`.  Your x86 evaluator will already do the right thing for both 32-bit and 64-bit.

2) In terms of testing, since this is the PR that introduces the new opcode I think it should be unit tested as part of its introduction.  Testing byteswap correctness should be relatively straightforward.

liqunl:
Re 1:
I can create a git issue for this item but the code will be in next PR.
Re 2:
I've done a unit test locally and the result is good.

liqunl:
For 2, I'll try to add a unit test in fvtest and put it in a different commit in this PR.

sophia-guo:
"Add ibyteswap unit test commit" lgtm.

liqunl:
@sophia-guo super thanks for your review and help on adding the unit test. @0xdaryl Could you review and add your comments? If it looks good to you, can you merge it?

And I have to mention that personal build passed with this commit, and I checked the detailed result of the test to make sure it is a valid test.

liqunl:
A git issue is created for lbyteswap https://github.com/eclipse/omr/issues/958

vijaysun-omr:
Thanks for addressing the suggestions Liqun. I am merging.

PR #937, opened 2017-03-28
Remove the initialized check on a class

liqunl:
Loosen the condition on constraining indirect load from a J9Class by
removing the initialized check. This is because java/lang/Class
reference and the component type if the class is an array, are
properties of the class file, and will be initialized before the class
is sufficiently initialized, so it's safe to dereference them at
compile time without the initialized check.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -2072,6 +2072,7 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
             break; // OK to attempt compile-time load
          }
 
+      const static char *needInitializedCheck = feGetEnv("TR_needInitializedCheck");
       TR::VPConstraint *base = vp->getConstraint(node->getFirstChild(), isGlobal);
       if (base && node->getOpCode().hasSymbolReference() &&
           (node->getSymbolReference() == vp->comp()->getSymRefTab()->findVftSymbolRef()))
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik This is the change for removing class initialization check, could you review when you have time? Personal build has passed.

andrewcraik:
LGTM

liqunl:
This is a j9 specific change.

mgaudet:
There's a grammar error with 'loose' here -- not sure if supposed to be 'Loosen' or 'Lose'; Both scan OK to me in a way that 'loose' doesn't.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2162,7 +2163,7 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
           && base->getClassType()
           && base->getClassType()->asFixedClass()
           && base->getClass()
-          && TR::Compiler->cls.isClassInitialized(vp->comp(), base->getClass()))
+          && (!needInitializedCheck || TR::Compiler->cls.isClassInitialized(vp->comp(), base->getClass())))
          {
          if (symRef == vp->comp()->getSymRefTab()->findJavaLangClassFromClassSymbolRef())
```
vijaysun-omr: This is'nt directly related to your commit but while reviewing this change I saw another snippet of code guarded by if (symRef == vp->comp()->getSymRefTab()->findJavaLangClassFromClassSymbolRef()) further down around line 2214 as part of the same if-then-else if construct that starts at line 2168. If you agree the code around line 2214 is unnecessary, please delete it to clean up the code.

liqunl: That is not in the if-then-else if construct starting at line 2168. It belongs to the if-then-else starting at line 2077

liqunl:
Grammar error with 'loose' fixed.

PR #940, opened 2017-03-28
Add OSR call site remat table

ncough:
When OSR is enabled, it is necessary to store the contents of the
stack before a call, as the call may be inlined and could contain
an OSR transition. If this transition were to occur, it would
have to reconstruct the stack for all of the callers.

Often, the stack contains values stored elsewhere which will
not be modified within the inlined call. Therefore, it is safe to
remat the pending push, representing the stack contents, within
OSR code blocks. This change adds the infrastructure for such 
an optimization by adding rematable pending pushes to the call site info.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -2229,6 +2229,87 @@ OMR::Compilation::isInlinedDirectCall(uint32_t index)
    return _inlinedCallSites[index].directCall();
    }
 
+/*
+ * Get the number of pending push slots for a caller, as this will be the size of
+ * the OSRCallSiteRemat table. In the event that the table has not been initialized,
+ * it will return 0.
+ */
+uint32_t
+OMR::Compilation::getOSRCallSiteRematSize(uint32_t callSiteIndex)
+   {
+   if (!_inlinedCallSites[callSiteIndex].osrCallSiteRematTable())
+      return 0;
+
+   int32_t callerIndex = self()->getInlinedCallSite(callSiteIndex)._byteCodeInfo.getCallerIndex();
+   return callerIndex < 0 ? self()->getMethodSymbol()->getNumPPSlots() :
+      self()->getInlinedResolvedMethodSymbol(callerIndex)->getNumPPSlots();
+   }
+
+/*
+ * Get the pending push symbol reference and the corresponding load, to later remat the pending push 
+ * within OSR code blocks inside the callee. To get a mapping, the call site index for the callee and
+ * the caller's pending push slot should be provided.
+ */
+void
+OMR::Compilation::getOSRCallSiteRemat(uint32_t callSiteIndex, uint32_t slot, TR::SymbolReference *&ppSymRef, TR::SymbolReference *&loadSymRef)
+   {
+   int32_t *table = _inlinedCallSites[callSiteIndex].osrCallSiteRematTable();
+   if (!table)
+      {
+      ppSymRef = NULL;
+      loadSymRef = NULL;
+      return;
+      }
+
+#if defined(DEBUG) || defined(PROD_WITH_ASSUMES)
+   // Ensure the requested slot is valid, based on the total number of PP slots
+   uint32_t callerNumPPSlots = self()->getOSRCallSiteRematSize(callSiteIndex);
+   TR_ASSERT(slot < callerNumPPSlots, "can only perform call site remat for the caller's pending pushes");
+#endif
+
+   TR::SymbolReferenceTable *symRefTab = self()->getSymRefTab();
+   ppSymRef = table[slot * 2] == 0 ? NULL : symRefTab->getSymRef(table[slot * 2]);
+   loadSymRef = table[slot * 2 + 1] == 0 ? NULL : symRefTab->getSymRef(table[slot * 2 + 1]);
+   }
+
+/*
+ * Set the pending push symbol reference and the corresponding load, to later remat the
+ * pending push within OSR code blocks inside the callee. To correctly add the entry, the callee's
+ * call site index and the caller's pending push with its matching load should be provided.
+ *
+ * It is necessary to provide the pending push symbol reference, rather than just its slot, as the slots
+ * may be shared. The information necessary to find the symref from the slot is stored, as the call site
+ * should be an OSR point, however the pending push symref is expected to be already known, so it is
+ * cheaper to use directly.
+ */
+void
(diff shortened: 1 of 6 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -2173,6 +2173,81 @@ OMR::Compilation::isInlinedDirectCall(uint32_t index)
...
+      }
+
+#if defined(DEBUG) || defined(PROD_WITH_ASSUMES)
+   // Ensure the requested slot is valid, based on the total number of PP slots
+   uint32_t callerNumPPSlots = self()->getOSRCallSiteRematSize(callSiteIndex);
+   TR_ASSERT(slot < callerNumPPSlots, "can only perform call site remat for the caller's pending pushes");
+#endif
+
+   TR::SymbolReferenceTable *symRefTab = self()->getSymRefTab();
+   ppSymRef = table[slot * 2] == 0 ? NULL : symRefTab->getSymRef(table[slot * 2]);
```
andrewcraik: Remind me again why the slot and the caller index aren't enough to find the ppSymRef?

ncough: Pending push slots are shared, so it would be necessary to work out which symbol reference is within the slot at the call site. It would be possible to do with the OSR data, as it stores this information.

andrewcraik: Ok but this saves compile-time which makes sense. Could you add a comment about that somewhere helpful in this code?

compiler/compile/OMRCompilation.cpp
```diff
@@ -2173,6 +2173,86 @@ OMR::Compilation::isInlinedDirectCall(uint32_t index)
    return _inlinedCallSites[index].directCall();
    }
 
+/*
+ * Get the size of the OSRCallSiteRemat table, based on the number of pending push
```
andrewcraik: This sounds a bit confusing. Perhaps it should read: Get the number of pending push slots in the given caller. This is the number of entries in the CallSiteRemat entry for the caller - or something like that.

compiler/optimizer/Inliner.cpp
```diff
@@ -2382,6 +2382,63 @@ TR_ParameterToArgumentMapper::lookForModifiedParameters(TR::Node * node)
       }
    }
 
+void
+TR_ParameterToArgumentMapper::mapOSRCallSiteRematTable(uint32_t siteIndex)
```
andrewcraik: Can we please have a descriptive comment on what this does and emphasize that it only does something if OSR is enabled? Should there be a check for OSR enablement in here explicitly in addition to testing the remat size?

andrewcraik: This is not useful under involuntary OSR so will we not do this in that mode?

andrewcraik: Does this need to check for OSR infrastructure being removed as part of late inlining?

andrewcraik:
In general a fine addition to the infrastructure - just a few small comments to address.

ncough:
@vijaysun-omr As this is a pre-req for other changes and I have addressed current concerns, I would appreciate your thoughts on this PR.

vijaysun-omr:
Looking at the commit code changes, am I right in thinking the rematerialization is only considered for values that are arguments to a call that is being inlined ? In other words, other pending push temp stores will likely still be left on the main line path and not be considered eligible for rematerialization but this is maybe considered acceptable since a large majority of pending push stores before calls are in fact for the arguments ?

ncough:
For postExecution OSR (and preExecution?), the pending pushes before a call do not include the arguments, as they would already have been popped off the stack for the call. The remaining items on the stack are typically for calls after the inlined call, which take these stack items and the inlined call's result as arguments. Moreover, these items are typically autos or parms which will not be modified by the inlined call, making them safe to use for remat within OSR code blocks. Other pending push temp stores will be left on the main line path.

compiler/optimizer/Inliner.cpp
```diff
@@ -2382,6 +2382,71 @@ TR_ParameterToArgumentMapper::lookForModifiedParameters(TR::Node * node)
...
+            }
+
+      if (!parmMap)
+         TR::DebugCounter::incStaticDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "osrCallSiteRemat/mapParm/missing/(%s)",  comp()->signature()));
+      }
+
+   // Update the remat tables for calls within the current
+   for (int32_t childIndex = 0; childIndex < comp()->getNumInlinedCallSites(); ++childIndex)
+      {
+      TR_InlinedCallSite &ics = comp()->getInlinedCallSite(childIndex);
```
vijaysun-omr: Why does this routine need to create the table for the children recursively ? Won't the parameter to argument mapper code get called for each inlined method (i.e. siteIndex) anyway ?

compiler/optimizer/Inliner.cpp
```diff
@@ -2382,6 +2382,71 @@ TR_ParameterToArgumentMapper::lookForModifiedParameters(TR::Node * node)
...
+      for (; parmMap; parmMap = parmMap->getNext())
+         if (symbol == parmMap->_parmSymbol)
+            {
+            if (parmMap->_isConst)
+               {
+               // Should be able to do const, current side table does not allow it
+               comp()->setOSRCallSiteRemat(siteIndex, ppSymRef, NULL);
+               TR::DebugCounter::incStaticDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "osrCallSiteRemat/mapParm/const/(%s)",  comp()->signature()));
+               }
+            else if (loadSymRef->getOffset() > 0)
```
vijaysun-omr: What exactly does this case denote ? i.e. what is special about the loadSymRef->getOffset() > 0 case ?

ncough: This case was only added to replicate behavior in the existing mapper. It is there to handle the case where a loadaddr is replaced, however, due to the constraints on what can be used for remated, this will never occur.

compiler/optimizer/Inliner.cpp
ncough: The purpose of the recursive call is to map the tables for any methods inlined within the current, as their tables would have already been generated. This tables may have used a parm from the current method to remat a pending push, therefore it is necessary to map this symref to a new one, as parms for inlined methods will all be mapped.

vijaysun-omr: Ah ok, so whether the callee's table already been generated before or not, we need it as part of this context since it's part of the parent (caller) call graph being inlined here.

vijaysun-omr:
Thanks for answering my questions. Merging now.

PR #941, opened 2017-03-28
Handle unequal number of children in identicalSubTrees

r30shah:
This function takes two nodes and traverse the subtrees recursively
to check if we have identical sub trees. When we pass two nodes with
unequal number of children there is a potential of segmentation fault.
Adding another check which will return false if nodes have unequal number
of children.


andrewcraik:
LGTM

PR #942, opened 2017-03-29
Consolidate X86 OpCode Information

0dvictor:
All information related to an OpCode is now in the same file.
1. Merge instruction name, mnemonic, binary information, properties
into one unified table.
2. De-duplicate logics generating size-parameterized opcodes.


andrewcraik:
Couple of formatting comments for the new enum file:
- the indent for the lines under each instruction seem quite high; they should probably be reduced to match the current code style to reduce the width of the line and the amount of right scrolling one needs to do on narrower screens
- the property0 and property1 lines are far too long - I think you need to come up with some nice convention how to wrap those lines so that they are easier to read and to see the different properties

I am a big fan of this change - having all the tables merged together greatly reduces the chances of mistakes when adding opcodes and will make sure everything matches up.

Given the scope of the change I think some comment about how the testing has been done and how the conversion to the new format was done is warranted. Throughput performance and compile-time for the OMR compiler should not be affected, but build time might be (though I wouldn't expect it to amount to much).

The Pull Request description is too terse and vague - it should describe the consolidation and how things are changed etc.

0dvictor:
I believe the intent for the lines under each instruction is needed. It guarantees entries under INSTRUCTION macro are aligned; and hence make it easier to read. I don't think reducing 5-6 spaces in the beginning of the line would increase readability; but the misalignment, on the other hand, would make it harder to read, especially most IDE does not provide good syntax highlight for this file.
In addition, the intent matches the current code style: when break arguments into multiple lines, each argument is aligned. For example, in compiler/il/OMRBlock.cpp:
```
void
OMR::Block::uncommonNodesBetweenBlocks(TR::Compilation *comp, TR::Block *newBlock, TR::ResolvedMethodSymbol *methodSymbol)
   {
   TR_ScratchList<TR::SymbolReference> symbolReferenceTempsA(self()->trMemory());
   TR_ScratchList<TR::SymbolReference> injectedBasicBlockTemps(self()->trMemory());
   TR_ScratchList<TR::SymbolReference> symbolReferenceTempsC(self()->trMemory());

   TR_HandleInjectedBasicBlock ibb(comp, NULL, methodSymbol? methodSymbol: comp->getMethodSymbol(),
                                   symbolReferenceTempsA,
                                   injectedBasicBlockTemps,
                                   symbolReferenceTempsC);
   ibb.findAndReplaceReferences(self()->getEntry(), newBlock, NULL);
```

0dvictor:
I agree that property0 and property1 lines are quite long now; however, they still contains quite a few deprecated/unused entries caused by previous restructuring. Subsequent changesets will cleanup these entries and hence the property line would be much shorter.

0dvictor:
The conversion was done via a custom made executable that parses original code and generates the new table. It guarantees a line-to-line match of all four entries: name, mnemonic, property0, and propery1.
All required tests from TravisCI and IBM's much deeper test suits have cleared.
Throughput and compile-time performance are not expected to change as the code would be identical to the previous version after the C Preprocessor.

0dvictor:
Commit message has been updated.

andrewcraik:
Ok if you really prefer the larger indent I can probably live with it unless any of the committers has an issue with it. If the property lines will shrink that is good, but won't wrapping them make it easier to make those changes or are you wanting this format temporarily to make automated clean-up easier and we can maybe reformat the properties later to make them much easier to read?

0dvictor:
Yes, it is a good idea to format property lines in a better way; and I will definitely do in subsequent changesets as follow-ups.
For this PR, I prefer to keep this unwrapped format. Besides its benefits for automated clean-up, it would be much easier to re-format the properties once all unused/deprecated entries are deleted.

andrewcraik:
Ok I buy the need to keep it easy for automation over the short term with a view that it will be reformatted in a small integer number of weeks.

andrewcraik:
@0xdaryl I think this is good to go - it probably needs a rebase from @0dvictor, but could you see if you have any objections in its current form so we can keep this moving forward. The concerns about formatting were addressed in my discussion with @0dvictor where he said that he needs the current format to help facilitate automated clean-up, but the number of properties etc should drop shortly.

compiler/x/codegen/OpBinary.cpp
```diff
@@ -16,901 +16,33 @@
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
  *******************************************************************************/
 
-#include "x/codegen/X86Ops.hpp"  // for TR_X86OpCode, ::IA32NumOpCodes, etc
+#include "X86Ops.hpp"
```
0dvictor: Good point. I have reverted this line.

compiler/x/codegen/X86Ops.enum
```diff
@@ -0,0 +1,3560 @@
+/*******************************************************************************
```
0dvictor: Updated.

compiler/x/codegen/X86Ops.enum
```diff
@@ -0,0 +1,3560 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2000, 2017
```
0dvictor: Updated.

compiler/x/codegen/X86Ops.enum
```diff
@@ -0,0 +1,3560 @@
...
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *******************************************************************************/
+
+/*
+ * This file will be included within an enum.  Only comments, preprocessor macros,
```
0dvictor: Updated.

compiler/x/codegen/OpBinary.cpp
```diff
@@ -16,901 +16,33 @@
...
-    { VEX_L___, VEX_vNONE, PREFIX_66, REX__, ESCAPE_____, 0x83, 1, ModRM_EXT_, Immediate_S },    // OR2RegImms
-    { VEX_L___, VEX_vNONE, PREFIX___, REX__, ESCAPE_____, 0x81, 1, ModRM_EXT_, Immediate_4 },    // OR4RegImm4
-    { VEX_L___, VEX_vNONE, PREFIX___, REX_W, ESCAPE_____, 0x81, 1, ModRM_EXT_, Immediate_4 },    // OR8RegImm4 (AMD64)
-    { VEX_L___, VEX_vNONE, PREFIX___, REX__, ESCAPE_____, 0x83, 1, ModRM_EXT_, Immediate_S },    // OR4RegImms
-    { VEX_L___, VEX_vNONE, PREFIX___, REX_W, ESCAPE_____, 0x83, 1, ModRM_EXT_, Immediate_S },    // OR8RegImms (AMD64)
-    { VEX_L___, VEX_vNONE, PREFIX___, REX__, ESCAPE_____, 0x80, 1, ModRM_EXT_, Immediate_1 },    // OR1MemImm1
-    { VEX_L___, VEX_vNONE, PREFIX_66, REX__, ESCAPE_____, 0x81, 1, ModRM_EXT_, Immediate_2 },    // OR2MemImm2
-    { VEX_L___, VEX_vNONE, PREFIX_66, REX__, ESCAPE_____, 0x83, 1, ModRM_EXT_, Immediate_S },    // OR2MemImms
-    { VEX_L___, VEX_vNONE, PREFIX___, REX__, ESCAPE_____, 0x81, 1, ModRM_EXT_, Immediate_4 },    // OR4MemImm4
-    { VEX_L___, VEX_vNONE, PREFIX___, REX_W, ESCAPE_____, 0x81, 1, ModRM_EXT_, Immediate_4
```
0xdaryl: This should be a relative path as well `codegen/X86Ops.ins'

0xdaryl: This should be a relative path as well `codegen/X86Ops.ins'

0xdaryl: This should be a relative path as well `codegen/X86Ops.ins'

compiler/x/codegen/X86Debug.cpp
```diff
@@ -2694,886 +2694,18 @@ TR_Debug::print(TR::FILE *pOutFile, TR::AMD64WriteBarrierSnippet * snippet)
...
-   "VirtualGuardNOP",
-   "LABEL",
-   "FCMPEVAL",
-   "RestoreVMThread",
-   "PPS_OPCount",
-   "PPS_OPField",
-   "AdjustFramePtr",
-   "ReturnMarker"
+#define INSTRUCTION(name, mnemonic, binary, property0, property1) #name
+#include "X86Ops.ins"
```
0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

compiler/x/codegen/X86Debug.cpp
```diff
@@ -2694,886 +2694,18 @@ TR_Debug::print(TR::FILE *pOutFile, TR::AMD64WriteBarrierSnippet * snippet)
...
+#undef INSTRUCTION
+   };
+
+static const char * opCodeToMnemonicMap[] =
+   {
+#define INSTRUCTION(name, mnemonic, binary, property0, property1) #mnemonic
+#include "X86Ops.ins"
```
0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

compiler/x/codegen/X86Debug.cpp
```diff
@@ -2694,886 +2694,18 @@ TR_Debug::print(TR::FILE *pOutFile, TR::AMD64WriteBarrierSnippet * snippet)
 
 #endif
 
-static const char * opCodeToNameMap[IA32NumOpCodes] =
+static const char * opCodeToNameMap[] =
    {
```
0xdaryl: For a future follow-on item, it should be possible to roll these names into your unified table as well.

compiler/x/codegen/OpBinary.cpp
0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

compiler/x/codegen/X86Ops.hpp
```diff
@@ -25,892 +25,12 @@
...
-   PPS_OPFIELD,     // Phase profiling site (object field counting)
-   AdjustFramePtr,  // Adjust frame pointer
-   ReturnMarker,    // Return Marker for linkages that do not end with a RET instruction
-   IA32LastOp = ReturnMarker,
-   IA32NumOpCodes = IA32LastOp + 1
-   } TR_X86OpCodes;
+enum TR_X86OpCodes : uint32_t
+   {
+#define INSTRUCTION(name, mnemonic, binary, property0, property1) name
+#include "X86Ops.ins"
```
0xdaryl: This should be a relative path as well `codegen/X86Ops.ins`

compiler/x/codegen/OpBinary.cpp
0dvictor: Updated.

0dvictor: Updated.

0dvictor: Updated.

0dvictor: Updated.

0dvictor: Updated.

0dvictor: Updated

compiler/x/codegen/X86Debug.cpp
0dvictor: Updated.

compiler/x/codegen/X86Debug.cpp
0dvictor: Updated.

compiler/x/codegen/X86Ops.hpp
0dvictor: Updated.

compiler/x/codegen/X86Debug.cpp
0dvictor: Good suggestion, I'll definitely keep it in mind.

PR #944, opened 2017-03-29
Fix caller index for osr meta data related logic

cathyzhyi:
For late inlining, the immediate caller symbol might not be on the
callStack but the callerSymbol in addGuardForVirtual
is read from the top of the callStack and therefore is incorrect for
purpose of finding OSRMethodMetaData. Change the code to use
callNode instead to more reliably find the caller symbol.

Also, removed the computeNumLivePendingSlotsAndNestingDepth which not used anywhere.


andrewcraik:
The actual fix looks good. I'm not sure what the history of that unused function is - @vijaysun-omr what are your thoughts on the code clean-up to remove that method?

vijaysun-omr:
We were using the "wrong" caller symbol and this PR uses the one on the call node (that should always be the right one, even during late inlining).

PR #945, opened 2017-03-29
Support MethodBuilders without line or file info

mgaudet:
Also adds a test for the same, and, restructures the naming in JitBuilder's method file to make methods easier to find with `grep`-like tools.

### DIFF
--- a/fvtest/jitbuildertest/AnonymousTest.cpp
+++ b/fvtest/jitbuildertest/AnonymousTest.cpp
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2017, 2017
+ *
+ *  This program and the accompanying materials are made available
+ *  under the terms of the Eclipse Public License v1.0 and
+ *  Apache License v2.0 which accompanies this distribution.
+ *
+ *      The Eclipse Public License is available at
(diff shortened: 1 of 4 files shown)


fvtest/jitbuildertest/AnonymousTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+   }
+
+typedef uint32_t (*AnonFunc)();
+
+class AnonymousTest : public JitBuilderTest {};
+
+TEST_F(AnonymousTest, AnonymousTest) 
+   {
+   AnonFunc func; 
+   ASSERT_COMPILE(AnonTypes, Anonymous, func);
```
Leonardo2718: I'll just make a note here that since AnnonTypes doesn't define any types, you could also just do `ASSERT_COMPILE(TR::TypeDictionary, Anonymous, func);`.

mgaudet: That's actually a great point, and I should do precisely that.

mstoodle:
LGTM @mgaudet . Welcome to the JitBuilder development club :) .

I'll wait for you to update according to @Leonardo2718 's suggestion, which I agree is a good one, and then merge.

mgaudet:
Just did that :)

PR #947, opened 2017-03-29
Rename setSupportsArrayCopy

fjeremic:
Rename setSupportsArrayCopy to setSupportsReferenceArrayCopy. The 
setSupportsArrayCopy API is currently overloaded for both primitive
and reference array copies. There already exists an API
setSupportsPrimitiveArrayCopy which only seems to be active on z
Systems.

This commit decouples z Systems from setSupportsPrimitiveArrayCopy and
in addition decouples setSupportsArrayCopy into
setSupportsReferenceArrayCopy and setSupportsPrimitiveArrayCopy. All
codegens currently supporting J9 are marked to reduce array copies, both
primititive and reference.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -1435,8 +1435,8 @@ class OMR_EXTENSIBLE CodeGenerator
    bool getSupportsPrimitiveArrayCopy() {return _flags2.testAny(SupportsPrimitiveArrayCopy);}
    void setSupportsPrimitiveArrayCopy() {_flags2.set(SupportsPrimitiveArrayCopy);}
 
-   bool getSupportsReferenceArrayCopy() {return _flags2.testAny(SupportsReferenceArrayCopy);}
-   void setSupportsReferenceArrayCopy() {_flags2.set(SupportsReferenceArrayCopy);}
+   bool getSupportsReferenceArrayCopy() {return _flags1.testAny(SupportsReferenceArrayCopy);}
+   void setSupportsReferenceArrayCopy() {_flags1.set(SupportsReferenceArrayCopy);}
 
    bool getSupportsEfficientNarrowIntComputation() {return _flags2.testAny(SupportsEfficientNarrowIntComputation);}
    void setSupportsEfficientNarrowIntComputation() {_flags2.set(SupportsEfficientNarrowIntComputation);}
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -1432,11 +1432,11 @@ class OMR_EXTENSIBLE CodeGenerator
    bool getSupportsDivCheck() {return _flags1.testAny(SupportsDivCheck);}
    void setSupportsDivCheck() {_flags1.set(SupportsDivCheck);}
 
-   bool getSupportsArrayCopy() {return _flags1.testAny(SupportsArrayCopy);}
-   void setSupportsArrayCopy() {_flags1.set(SupportsArrayCopy);}
-
    bool getSupportsPrimitiveArrayCopy() {return _flags2.testAny(SupportsPrimitiveArrayCopy);}
    void setSupportsPrimitiveArrayCopy() {_flags2.set(SupportsPrimitiveArrayCopy);}
+
+   bool getSupportsReferenceArrayCopy() {return _flags2.testAny(SupportsReferenceArrayCopy);}
+   void setSupportsReferenceArrayCopy() {_flags2.set(SupportsReferenceArrayCopy);}
 
    bool getSupportsEfficientNarrowIntComputation() {return _flags2.testAny(SupportsEfficientNarrowIntComputation);}
    void setSupportsEfficientNarrowIntComputation() {_flags2.set(SupportsEfficientNarrowIntComputation);}
@@ -1706,7 +1706,7 @@ class OMR_EXTENSIBLE CodeGenerator
       HasResumableTrapHandler                            = 0x00000040,
       //                                                 = 0x00000080,   // Available
       SupportsPartialInlineOfMethodHooks                 = 0x00000100,
-      SupportsArrayCopy                                  = 0x00000200,
+      SupportsReferenceArrayCopy                         = 0x00000200,
       SupportsJavaFloatSemantics                         = 0x00000400,
       SupportsInliningOfTypeCoersionMethods              = 0x00000800,
       VMThreadRequired                                   = 0x00001000,
--- a/compiler/optimizer/LoopReducer.cpp
+++ b/compiler/optimizer/LoopReducer.cpp
@@ -928,7 +928,7 @@ bool
 TR_LoopReducer::generateArraycopy(TR_InductionVariable * indVar, TR::Block * loopHeader)
    {
    // for aladds
-   if (!comp()->cg()->getSupportsArrayCopy() && !comp()->cg()->getSupportsPrimitiveArrayCopy())
+   if (!comp()->cg()->getSupportsReferenceArrayCopy() && !comp()->cg()->getSupportsPrimitiveArrayCopy())
       {
       dumpOptDetails(comp(), "arraycopy not enabled for this platform\n");
       return false;
@@ -967,7 +967,7 @@ TR_LoopReducer::generateArraycopy(TR_InductionVariable * indVar, TR::Block * loo
 
    bool needWriteBarrier = comp()->getOptions()->needWriteBarriers();;
    //FUTURE: can eliminate wrtbar when src and dest are equal. Currently we don't reduce arraycopy like this.
-   if (arraycopyLoop.hasWriteBarrier() && needWriteBarrier && !comp()->cg()->getSupportsArrayCopy())
+   if (arraycopyLoop.hasWriteBarrier() && needWriteBarrier && !comp()->cg()->getSupportsReferenceArrayCopy())
       {
       dumpOptDetails(comp(), "arraycopy arraystore tree has write barrier as root and write barriers are enabled but no support for this platform- no arraycopy reduction\n");
       return false;
@@ -1088,23 +1088,24 @@ TR_LoopReducer::generateArraycopy(TR_InductionVariable * indVar, TR::Block * loo
    if (arraycopyLoop.getStoreAddress()->getIncrement() < 0)
       //&& (storeAddr->getFirstChild()->getSymbol()->getRegisterMappedSymbol() == loadAddr->getFirstChild()->getSymbol()->getRegisterMappedSymbol()))
       {
-      if (comp()->cg()->getSupportsPrimitiveArrayCopy())
-         arraycopy->setBackwardArrayCopy(true); /* bit available only to primitive arraycopy */
+      arraycopy->setBackwardArrayCopy(true);
       }
    else
       {
       arraycopy->setForwardArrayCopy(true);
       }
 
-   if (!comp()->cg()->getSupportsPrimitiveArrayCopy()) /* bits only available for non-primitive arraycopy */
+   switch (arraycopyLoop.getCopySize())
       {
-      switch(arraycopyLoop.getCopySize())
-         {
-         case 2: arraycopy->setHalfWordElementArrayCopy(true); break;
-         case 4: case 8: arraycopy->setWordElementArrayCopy(true); break;
-         }
-      }
+      case 2:
+         arraycopy->setHalfWordElementArrayCopy(true);
+         break;
 
+      case 4:
+      case 8:
+         arraycopy->setWordElementArrayCopy(true);
+         break;
+      }
 
    TR::Node * top = TR::Node::create(TR::treetop, 1, arraycopy);
    arrayStoreTree->setNode(top);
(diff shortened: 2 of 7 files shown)


fjeremic:
@andrewcraik @jdmpapin could you please review? This code touches areas you both have expertise in. After we are satisfied with the changes I will do a performance evaluation of the final change as this touches a very performance critical piece of code and make sure we are OK on this front.

Just to give you some context this effectively decouples reference and primitive array copies. z Systems needs such decoupling for concurrent scavenger as we currently cannot handle reference array copies and we need a clean way of disabling this while enabling primitive array copies.

@dsouzai @ebadyano FYI.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -1432,12 +1432,12 @@ class OMR_EXTENSIBLE CodeGenerator
    bool getSupportsDivCheck() {return _flags1.testAny(SupportsDivCheck);}
    void setSupportsDivCheck() {_flags1.set(SupportsDivCheck);}
 
-   bool getSupportsArrayCopy() {return _flags1.testAny(SupportsArrayCopy);}
-   void setSupportsArrayCopy() {_flags1.set(SupportsArrayCopy);}
-
    bool getSupportsPrimitiveArrayCopy() {return _flags2.testAny(SupportsPrimitiveArrayCopy);}
    void setSupportsPrimitiveArrayCopy() {_flags2.set(SupportsPrimitiveArrayCopy);}
 
+   bool getSupportsReferenceArrayCopy() {return _flags2.testAny(SupportsReferenceArrayCopy);}
```
andrewcraik: why is this flags2? The enum the flag is allocated in is commented as flags1 - isn't this going to test the wrong bit?

compiler/optimizer/LoopReducer.cpp
```diff
@@ -1088,23 +1088,24 @@ TR_LoopReducer::generateArraycopy(TR_InductionVariable * indVar, TR::Block * loo
    if (arraycopyLoop.getStoreAddress()->getIncrement() < 0)
       //&& (storeAddr->getFirstChild()->getSymbol()->getRegisterMappedSymbol() == loadAddr->getFirstChild()->getSymbol()->getRegisterMappedSymbol()))
       {
-      if (comp()->cg()->getSupportsPrimitiveArrayCopy())
-         arraycopy->setBackwardArrayCopy(true); /* bit available only to primitive arraycopy */
```
andrewcraik: This comment should be preserved somewhere because it is a warning not to copy the flag to reference copying code.

andrewcraik: I'm trying to prove to myself that only primitives get here, but I'm having trouble doing that... your intention is that the flag is still only set for primitives correct? If so maybe you could help me with a line number for the guard you see still protecting this. If it should apply to reference arrays have you checked that the bit does not have an overloaded meaning like the comment would suggest?

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Argh, this is such a silly but very costly mistake. I'm very glad we do thorough code reviews and I appreciate your level of detail. This is a bug. Fixed in 13f34f045b33633a1a598a18000da9a21cf328dc.

compiler/optimizer/LoopReducer.cpp
fjeremic: I don't think this flag should apply only to primitive array copies. The code just above can definitely create a 5-child arraycopy which is a reference arraycopy. And the original program was copying elements backwards. I don't see what the distinction here between primitive and reference arrays is.

In fact I've checked all the codegens we currently support. No codegen ever checks for the `isBackwardArrayCopy` flag. They all check for `isForwardArrayCopy` and if this returns false we assume its a backwards array copy:

Power:
https://github.com/eclipse/omr/blob/master/compiler/p/codegen/OMRTreeEvaluator.cpp#L3223

x86:
https://github.com/eclipse/omr/blob/master/compiler/x/codegen/OMRTreeEvaluator.cpp#L2894

z/Architecture:
https://github.com/eclipse/omr/blob/master/compiler/z/codegen/OMRTreeEvaluator.cpp#L13481

In addition a few lines below this change we set the `setHalfWordElementArrayCopy` and `setWordElementArrayCopy` flags. Before this change z Systems was the only codegen that answers `true` to `getSupportsPrimitiveArrayCopy` and not surprisingly the above two flags I just mentioned are not at all used withing the z Systems codegen. So there is no reason not to set them for primitive arrays as well.

andrewcraik: isBackwardArrayCopy is used in ValuePropagation.cpp:5518 ... still trying to work out if this code is limited in the cases it handles - it seems tied to the arraycopy implementations in the codegen...

fjeremic: Yes but the only call to this method is guarded by `getSupportsPostProcessArrayCopy` which only z Systems implements returning `true`. It is no surprise that z Systems is (currently) the only platform that answers `true` to `getSupportsPrimitiveArrayCopy` so the backwards array copy check there does not matter.

jdmpapin: > In fact I've checked all the codegens we currently support. No codegen ever checks for the isBackwardArrayCopy flag. They all check for isForwardArrayCopy and if this returns false we assume its a backwards array copy

I'm still trying to understand the surface area of all of these flags, but one thing to note is that `isBackwardArrayCopy` and `isForwardArrayCopy` are not independent. Setting one unsets the other. Furthermore, it's not obvious that the interpretation of `isForwardArrayCopy` is the same on all platforms. A comment at the start of the z `arraycopyEvaluator` describes its interpretation in terms that make claims about "390 codegen" rather than about the meaning of the flag more broadly.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1170,6 +1170,16 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
...
 
+      if (primitiveArray1 || primitiveArray2)
+         {
+         transformTheCall = primitiveTransform;
+         }
+
+      if (primitiveArray1 || primitiveArray2)
+         {
+         transformTheCall = referenceTransform;
+         }
```
jdmpapin: Did you mean `if (referenceArray1 || referenceArray2)`?

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1159,6 +1158,7 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
         srcOffHigh >= 0 && dstOffHigh >= 0 && copyLenHigh >= 0)
       {
       transformTheCall = true;
```
jdmpapin: We might not have any type information about either array. Do you still want this default in that case? Maybe it should default to something like `primitiveTransform && referenceTransform`.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1352,26 +1349,9 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
...
-      {
-      // If only primitive arraycopy is supported for this target, and the call is known to
-      // go to a reference arraycopy (both src and dst are known to be ref arrays)
-      // then transform the arraycopy into a call to
-      // java/lang/System.arraycopy(Object[], Object[], int, Object[], int, int)
-      // which is a private method in the java.lang.System class we ship
-      //
-      TR_ResolvedMethod * newMethod =
-         findResolvedClassMethod(comp(), "Ljava/lang/System;", "arraycopy", "([Ljava/lang/Object;I[Ljava/lang/Object;II)V");
-      }
```
jdmpapin: Thanks for cleaning this up

compiler/optimizer/ValuePropagationCommon.cpp
fjeremic: Yes I did. Fixed in 9289c54b0b26838c5753fc386b9f2dd7ce8ba98b. Thanks for catching this!

compiler/optimizer/LoopReducer.cpp
jdmpapin: It looks safe to call `setBackwardArrayCopy(true)` here regardless of `getSupportsPrimitiveAraryCopy`. It does unset `is[RarePath]ForwardArrayCopy`, but that's already not set on this new `arraycopy` node. So the only difference should be the answer to `isBackwardArrayCopy`, which AFAICT is only *really* queried by `ArraycopyTransformation::createMultipleArrayNodes`, which is only called on z where we were already explicitly marking it as backward.

The other uses I found of `isBackwardArrayCopy` are just in assertions in related `arraycopy` flag setter methods, and code that copies properties from one `arraycopy` node to another.

compiler/optimizer/ValuePropagationCommon.cpp
fjeremic: Did you mean to use OR instead of AND there? I thought of that as well. This change keeps the current behavior in that if primitive **or** reference array are supported and we have no type information then we carry out the transformation anyway.

I could not figure out why this was even correct to begin with. In my opinion setting this value to true by default and figuring out based on information we know whether to set it to false is too aggressive. We should be setting it to false by default and figuring out based on the information we know whether it should be true.

This is a much less aggressive approach but also more risky and probably outside the scope of this change. Do any of the reviewers have other thoughts?

compiler/optimizer/LoopReducer.cpp
```diff
@@ -1088,23 +1088,24 @@ TR_LoopReducer::generateArraycopy(TR_InductionVariable * indVar, TR::Block * loo
...
-      }
+      case 2:
+         arraycopy->setHalfWordElementArrayCopy(true);
+         break;
 
+      case 4:
+      case 8:
+         arraycopy->setWordElementArrayCopy(true);
+         break;
+      }
```
jdmpapin: These also seem to be safe to set regardless of the value of `getSupportsPrimitiveArrayCopy`. They will still be set on the platforms that query them, and they'll be ignored on platforms where they weren't previously set.

The availability of the flag bits cannot depend on whether the particular `arraycopy` node is for copying primitives, since nothing here queries the node about that.

The flag bits are also available on all platforms. One of the bits collides with `dontTransformArrayCopyCall`, but we already better not transform a call node into `arraycopy` in-place when that flag is set.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1149,7 +1148,7 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
    // If it is OK to convert this call to a possible call to the (fast)
    // arraycopy helper, do it.
    //
-   if ((primitiveTransform || fullTransform) &&
+   if ((primitiveTransform || referenceTransform) &&
        !comp()->getOption(TR_DisableArrayCopyOpts) &&
        !node->isDontTransformArrayCopyCall() &&
```
jdmpapin: Note that we respect `dontTransformArrayCopyCall` here

compiler/optimizer/ValuePropagationCommon.cpp
jdmpapin: I think the original `fullTransform` means that the transformation can be done even without any type information here. Previously it just so happened that `primitiveTransform` implied `fullTransform`, so the `primitiveTransform || fullTransform` condition actually guaranteed `fullTransform`. In that case, `primitiveTransform && referenceTransform` seems to be the equivalent of the old `fullTransform`.

I agree that larger changes to the logic here are outside the scope of this change. However, changing the default away from `true` is safe because it can only make the transformation more conservative. The default I suggested above wouldn't affect performance either, because it's still the case (for now) that all platforms have both `primitiveTransform` and `referenceTransform`.

By the way, I've verified that your changes to the logic below do not affect the resulting values of `transformTheCall` and `needArrayCheck`, given the current and previous values on each platform of `primitiveTransform` and `fullTransform`/`referenceTransform`. I checked each (possible) combination of values of the following:
* `primitiveArray1`,
* `primitiveArray2`,
* `referenceArray1`,
* `referenceArray2`,
* `identicalTypes`, standing in for `[...] && srcObject...->getClass() == dstObject...->getClass()`, and
* `alwaysCallWriteBarrier`, standing in for `comp()->getOptions()->alwaysCallWriteBarrier()`.

fjeremic: @jdmpapin I agree. We should move towards a more conservative approach. I've made the change in 5273acf.

andrewcraik:
Given the nature of the changes in this item a full perf validation will ben needed before we consider merging - failure to transform could cause serious performance issues.

andrewcraik:
Ok I think this is ready for some performance testing - we can collaborate offline about running tests on the version of the code built by @fjeremic.

fjeremic:
@andrewcraik @jdmpapin performance testing is complete on a downstream project. Across multiple benchmarks on multiple platforms we support this change was compared and the performance results indicate we are on par across throughput, compile time, and startup. If we are all satisfied with the results please comment and let the commiter know. Thank you very much!

andrewcraik:
I think this is now god to go from a perf perspective based on the testing done downstram of OMR. @jdmpapin if you can do your last checks I think this should be good to merge once you green light it.

jdmpapin:
Looks good with the small adjustments made above

On to you, @vijaysun-omr

compiler/optimizer/LoopReducer.cpp
vijaysun-omr: We should probably also look at the ARM code generator to be complete on discussions relating to "all the codegens".

fjeremic: @vijaysun-omr ARM code generator currently does not support array copy reduction so we don't have to worry about preexisting code in the case of ARM.

PR #948, opened 2017-03-30
Avoid unnecessary async checks at scorching

JamesKingdon:
Added a test to avoid unnecessary async checks being added to code
that isn't going to be recompiled because it's already at scorching.


andrewcraik:
LGTM

andrewcraik:
IBM did some performance evaluation on this change it and there were no regressions on the test cases we tried.

vijaysun-omr:
Pretty straightforward change to avoid adding sampling points when we are already at the highest possible optimization level. Merging.

Issue #950, opened 2017-03-30
Expanding AutoSIMD support to Float datatype on z


IBMJimmyk:
Modified the evaluators for the following opcodes to handle float data type on z:
vadd, vsub, vmul, vdiv, vneg
vsplats, getvelem, vsetelem

The following opcode evaluators already supported float data type (since they don't care about data type) but are now marked as supported in getSupportsOpCodeForAutoSIMD.
vload, vloadi, vstore, vstorei

inlineVectorUnaryOp and inlineVectorBinaryOp are used inside many of the previously mentioned evaluators. They were changed to set the right bits in an instruction to indicate that a floating point version should be used.

As an example, prior to any changes, the mask4 bits in a VFA instruction were always set to 0x3 to indicate double type data since it was only used for doubles at the time. This was changed so that for float type data, the mask4 bit are now set to 0x2 to indicate the vector contains floats that need to be added together.

generateFusedMultiplyAddIfPossible was modified to try and generate VFMA or VFMS for float data as well as the previously supported double data.

In most of the modified functions, a check for zNext is performed since it is necessary for the floating point version of the vector instructions.

After these changes, AutoSIMD support for floats on z now matches the support for doubles.

0xdaryl:
@IBMJimmyk : can this issue be closed now, or is there more work that needs to be done?

IBMJimmyk:
This can be closed. All outstanding work on this item has been completed. Any future work would go under a new Issue.

PR #951, opened 2017-03-30
Expand AutoSIMD support on z to floats

IBMJimmyk:
Updates the implementation of the following evaluators on z to support the
float data type:
vadd, vsub, vmul, vdiv, vneg
vsplats, getvelem, vsetelem

Many of those evaluators made calls to inlineVectorUnaryOp and
inlineVectorBinaryOp. Those functions were also updated to handle the
float data case. They now set the proper bits in a vector instruction when
a float version is used.

In most of the modified functions, a check for zNext was added since it is
necessary for the float versions of the vector instructions to be used.

getSupportsOpCodeForAutoSIMD was updated to reflect the new support for
floats on zNext.

Issue: #950

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -16506,10 +16506,6 @@ OMR::Z::TreeEvaluator::inlineVectorUnaryOp(TR::Node * node,
          generateVRRaInstruction(cg, op, node, returnReg, sourceReg1, 0, 0, getVectorElementSizeMask(node));
          break;
       case TR::InstOpCode::VFPSO:
-         if (node->getDataType() == TR::VectorFloat && !cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zNext))
-            {
-            TR_ASSERT(false, "Vector Float operation only supported on zNext onward for node : %s", cg->getDebug()->getName(node));
-            }
          breakInst = generateVRRaInstruction(cg, op, node, returnReg, sourceReg1, 0 /* invert sign */, 0, getVectorElementSizeMask(node));
          break;
       default:
@@ -16562,10 +16558,6 @@ OMR::Z::TreeEvaluator::inlineVectorBinaryOp(TR::Node * node, TR::CodeGenerator *
       case TR::InstOpCode::VFS:
       case TR::InstOpCode::VFM:
       case TR::InstOpCode::VFD:
-         if (node->getDataType() == TR::VectorFloat && !cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zNext))
-            {
-            TR_ASSERT(false, "Vector Float operation only supported on zNext onward for node : %s", cg->getDebug()->getName(node));
-            }
          mask4 = getVectorElementSizeMask(node);
          breakInst = generateVRRcInstruction(cg, op, node, targetReg, sourceReg1, sourceReg2, 0, 0, mask4);
          break;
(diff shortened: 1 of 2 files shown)


compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -16549,12 +16553,22 @@ OMR::Z::TreeEvaluator::inlineVectorBinaryOp(TR::Node * node, TR::CodeGenerator *
          mask4 = getVectorElementSizeMask(node);
          breakInst = generateVRRcInstruction(cg, op, node, targetReg, sourceReg1, sourceReg2, 0, 0, mask4);
          break;
-      // These require mask4 = 0x3, other values illegal
-      // These only operate on long BFP elements,
+      /*
+       * Before zNext, these required mask4 = 0x3 and other values were illegal
+       * This was because they only operated on long BFP elements.
+       * From zNext onward, mask4 is variable and sets the element size to operate on
+       */
```
fjeremic: Perhaps this comment is better suited in `getSupportsOpCodeForAutoSIMD` where we query for support?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -19318,8 +19407,16 @@ OMR::Z::TreeEvaluator::vsetelemEvaluator(TR::Node *node, TR::CodeGenerator *cg)
                }
             else if (size == 4)
                {
-               uint32_t value = valueNode->getIntegerNodeValue<uint32_t>();
-               offset = cg->fe()->findOrCreateLiteral(cg->comp(), &value, size);
+               if (valueNode->getOpCode().isFloat())
+                  {
+                  float value = valueNode->getFloat();
```
fjeremic: Minor syntactical nitpick: we call it `value` here but `v` above. Its good to stay consistent for easier parsing / readability.

compiler/z/codegen/OMRTreeEvaluator.cpp
IBMJimmyk: This comment was to update a preexisting comment. I can also add a similar comment to "getSupportsOpCodeForAutoSIMD". Should I also remove the comment from this location?

compiler/z/codegen/OMRTreeEvaluator.cpp
IBMJimmyk: I used value in vsetelem because the other cases in vsetelem used value. Similarly the existing cases in getvelem used 'v' so I used 'v' in getvelem. I will change the 'v' in getvelem to "value" to match vsetelem and because it's more descriptive than just 'v'.

fjeremic:
LGTM.

IBMJimmyk:
Okay. I squashed the commits together.

0xdaryl:
This sequence of code (or a slight variant thereof) is repeated often in the evaluators:

```
if (node->getDataType() == TR::VectorFloat && !cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zNext))
   {
   TR_ASSERT(false, "Vector Float operation only supported on zNext onward for node : %s", cg->getDebug()->getName(node));
   return NULL;
   }
```

Its arguably there to catch the unlikely case where someone has generated a float type vector opcode despite the underlying processor not supporting it (this should have been prohibited at opcode generating time).

Rather than pollute the code (and increase .so footprint) with this, I think it would be better to live in its own function that's simply called when needed.  It will also make it easier to add conditions to this in the future or drop the processor check, etc. (i.e., its all in one place).  Arguably, you ought to abort the compile rather than returning your way out of here because something has gone wrong and you likely can't recover from it.

IBMJimmyk:
Do you feel like it would be better to remove the checks for zNext inside the evaluators and assume the proper checks were done at opcode generation time? I do feel like I'm adding in too many extra checks.

The only check that was absolutely necessary was the one inside getSupportsOpCodeForAutoSIMD. Getting rid of the others would remove nine of the checks for zNext and only leave one left. At the same time this will address your second concern so that future changes to the check only need to be done in one place.

IBMJimmyk:
I decided to go with removing all the extra checks and rely on OpCode generation making sure the processor supports vector float operations. These changes have been added to the PR.

0xdaryl:
Will merge once CI completes.

PR #953, opened 2017-03-30
Experimental AVX Support on X86

0dvictor:
Initial support for AVX/AVX2/FMA instructions that uses VEX prefixes.

### DIFF
--- a/compiler/x/codegen/X86Ops.hpp
+++ b/compiler/x/codegen/X86Ops.hpp
@@ -359,15 +359,20 @@ class TR_X86OpCode
                break;
             }
          }
-      // check if the instruction has mandatory prefix(es)
-      inline bool hasMandatoryPrefix() const
+      // check if the instruction can be encoded as AVX
+      inline bool supportsAVX() const
          {
-         return prefixes == PREFIX___;
+         return vex_l != VEX_L___;
          }
       // check if the instruction is X87
       inline bool isX87() const
          {
          return (prefixes == PREFIX___) && (opcode >= 0xd8) && (opcode <= 0xdf);
+         }
+      // check if the instruction has mandatory prefix(es)
+      inline bool hasMandatoryPrefix() const
+         {
+         return prefixes == PREFIX___;
          }
       // check if the instruction is part of Group 7 OpCode Extension
       inline bool isGroup07() const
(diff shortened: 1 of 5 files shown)


andrewcraik:
LGTM - not enabled by default at present, but is the initial contribution ready for enabling soon.

andrewcraik:
@0xdaryl ping?

andrewcraik:
@0xdaryl any update on this one?

compiler/x/codegen/X86Ops.hpp
```diff
@@ -414,6 +423,11 @@ class TR_X86OpCode
       // OpCodes beyond FENCE are not X86 instructions and hence they do not have binary encodings.
       return op < FENCE ? _binaries[op].encode<TBuffer>(cursor, rex) : cursor;
       }
+   static inline void finalize(TR_X86OpCodes op, uint8_t* cursor)
+      {
+      if (op < FENCE)
```
0xdaryl: This test for finalization seems too fragile, even guarded with a comment.  I wonder if you could add a "finalization" property to an opcode and peek at the appropriate `_properties` table for the given `op` to decide whether to do this?

0dvictor: _properties would not work since the op can encoded as either SSE or AVX instructions; only finalization only works on AVX instructions. Adding such property does not help as we still have to detect if the instruction is SSE or AVX. An instruction is AVX if and only if its first byte is 0xC4 or 0xC5.
The root cause of this issue is that AVX instructions have three operands but SSE instructions only have two. Our CG defines instructions with two-operand as well. We need somehow convert the instruction to three-operand mode; but required information is only available after operands are encoded. I am working on a better design for such conversion but have not finalized yet.
This approach is one required least code refactoring and hence I would like to deliver as a intermediate path to unblock queued features that require AVX. As mentioned in the X86 Binary Encoding and SIMD changesets, we are working on supporting three-operand instructions. Once three-operand instructions are properly supported, this "finalization" approach will be replaced.

0xdaryl: My apologies for letting this languish.  My comment wasn't about the need to finalize an instruction.  Rather, its about how you decide on whether to finalize based on its relative position to another opcode in the table.  Its about coming up with a property to answer the real question you're asking: is this an instruction that has a real binary encoding?   The `_properties` tables are visible to this function, and a property like `PseudoOp` is close to what you want on the current opcodes that currently live after `FENCE` in that enum.  Maybe you need a `NotEncodable` property on the 18 special instructions in that enum, or repurpose the little-used `PseudoOp` property.

0dvictor: Thanks for clarification. Good point, let me look further into it and make updated.

andrewcraik:
@0dvictor any update on getting this ready to merge?

0dvictor:
I updated the code as @0xdaryl suggested; however, the new code requires PR #1064. I'll continue this PR once the prerequisite is delivered.

0dvictor:
As #1064 is merged, this PR can now continue.

compiler/x/codegen/X86Ops.hpp
0dvictor: Updates have been made. Now `PseudoOp` is checked when determine whether finalization should be called.

andrewcraik:
@0xdaryl over to you on this one.

0xdaryl:
Looks great Victor!

The Appveyor problems are still being worked out and are not related to this PR. Since the rest of CI passed cleanly I will merge.

PR #954, opened 2017-03-30
Use correct instruction generation function

dsouzai:
In the aloadHelper in the z tree evaluator, the LGG instruction needs to
be generated using the generateRXYInstruction but currently it is being
generated using the generateRXInstruction which is incorrect.


vijaysun-omr:
Looks right to me. @fjeremic please review as well.

fjeremic:
LGTM this is an obvious mistake.

PR #957, opened 2017-03-31
Remove stale HWProfiler options

mpirvu:
This changeset removes some old HWProfiler options that are not
referenced by any method.


vijaysun-omr:
Simple cleanup of some options code. Merging.

PR #959, opened 2017-03-31
Add common functions to OSR API concerning PPS

ncough:
OSR points are surrounded by pending push stores, to ensure
the stack is recreated after the transition, and treetops of
pending push loads, to avoid the side effects of the prior mentioned
stores. This adds functionality to determine if a node is one of
these and to extract the transition point for an OSR point based
on the current mode.

Moreover, it corrects an issue with OSR analysis as liveness and
use/def information should be calculated at the transition or
analysis point, not the OSR point.

### DIFF
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -686,18 +686,49 @@ TR_OSRMethodData::inlinesAnyMethod() const
    }
 
 void
-TR_OSRMethodData::addLiveRangeInfo(int32_t byteCodeIndex, TR_BitVector *liveRangeInfo)
+TR_OSRMethodData::addLiveRangeInfo(int32_t byteCodeIndex, TR::OSRPointType osrPoint, TR_BitVector *liveRangeInfo)
    {
-   bcLiveRangeInfoHashTab.Add(byteCodeIndex, liveRangeInfo);
+   TR_BCLiveRangeInfoHashKey key(byteCodeIndex, osrPoint);
+   bcLiveRangeInfoHashTab.Add(key, liveRangeInfo);
    }
 
-
+/*
+ * Get the live range info for a bytecode index
+ * In postExecution OSR, it is possible for the same bytecode index to have different
+ * liveness information, based on whether it was generated for a induction or
+ * an analysis point.
+ *
+ * For example, consider two calls, where the result of one feeds into the other.
+ * For a transition point after the first call, its result is on the stack, whilst,
+ * for an analysis point before the second call, the result has been taken as an
+ * argument and is no longer on the stack.
+ *
+ * This method will default to using the induction OSR point value, as it will
+ * always contain the live values at the analysis point. However, a more exact
+ * result can be achieve by specifing the point type.
+ */
 TR_BitVector *
 TR_OSRMethodData::getLiveRangeInfo(int32_t byteCodeIndex)
    {
+   TR_BitVector* liveRangeInfo = NULL;
+   if (getMethodSymbol()->comp()->getOSRTransitionTarget() == TR::postExecutionOSR)
+      {
+      liveRangeInfo = getLiveRangeInfo(byteCodeIndex, TR::inductionOSR);
+      if (!liveRangeInfo)
+         liveRangeInfo = getLiveRangeInfo(byteCodeIndex, TR::analysisOSR);
+      }
+   else
+      liveRangeInfo = getLiveRangeInfo(byteCodeIndex, TR::inductionOSR);
+   return liveRangeInfo;
+   }
+
+TR_BitVector *
+TR_OSRMethodData::getLiveRangeInfo(int32_t byteCodeIndex, TR::OSRPointType osrPoint)
+   {
+   TR_BCLiveRangeInfoHashKey key(byteCodeIndex, osrPoint);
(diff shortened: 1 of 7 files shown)


compiler/compile/OSRData.cpp
```diff
@@ -686,18 +686,34 @@ TR_OSRMethodData::inlinesAnyMethod() const
...
+   TR_BCLiveRangeInfoHashKey key(byteCodeIndex, osrPoint);
+   bcLiveRangeInfoHashTab.Add(key, liveRangeInfo);
    }
 
-
 TR_BitVector *
 TR_OSRMethodData::getLiveRangeInfo(int32_t byteCodeIndex)
    {
+   TR_BitVector* liveRangeInfo = NULL;
+   if (getMethodSymbol()->comp()->getOSRTransitionTarget() == TR::postExecutionOSR)
```
andrewcraik: This deserves a comment to explain the methodology here.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1419,7 +1419,69 @@ OMR::ResolvedMethodSymbol::getNumberOfBackEdges()
...
+ * Checks the provided node is pending push store or load expected to be
+ * around an OSR point.
+ */
+bool
+OMR::ResolvedMethodSymbol::isOSRRelatedNode(TR::Node *node)
+   {
+   return (node->getOpCode().isStoreDirect()
+         && node->getOpCode().hasSymbolReference()
+         && node->getSymbolReference()->getSymbol()->isPendingPush())
+      || (node->getOpCodeValue() == TR::treetop
```
andrewcraik: What is the logic behind anchored loads - do we get this structure as a result of the way trees are generated? There is normally no guarantee a load is 'anchored' so a comment on the logic surrounding this check would be good.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1419,7 +1419,73 @@ OMR::ResolvedMethodSymbol::getNumberOfBackEdges()
...
+   }
+
+/*
+ * Checks that the provided node is related to the OSR point's BCI.
+ * This applies to pending push stores and loads sharing the BCI.
+ */
+bool
+OMR::ResolvedMethodSymbol::isOSRRelatedNode(TR::Node *node, TR_ByteCodeInfo osrBCI)
+   {
+   TR_ByteCodeInfo bci = node->getByteCodeInfo();
```
andrewcraik: Should this be a reference type to save a copy?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1419,7 +1419,73 @@ OMR::ResolvedMethodSymbol::getNumberOfBackEdges()
...
+         && node->getFirstChild()->getOpCode().hasSymbolReference()
+         && node->getFirstChild()->getSymbolReference()->getSymbol()->isPendingPush());
+   }
+
+/*
+ * Checks that the provided node is related to the OSR point's BCI.
+ * This applies to pending push stores and loads sharing the BCI.
+ */
+bool
+OMR::ResolvedMethodSymbol::isOSRRelatedNode(TR::Node *node, TR_ByteCodeInfo osrBCI)
```
andrewcraik: since the osrBCI is not modified should this be a reference type as well to save a copy?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1419,7 +1419,73 @@ OMR::ResolvedMethodSymbol::getNumberOfBackEdges()
    return numBackEdges;
    }
 
+/*
+ * Returns the bytecode info for an OSR point, taking into account the
+ * possibility that the treetop for a call has different BCI to the OSR
+ * point.
+ */
+TR_ByteCodeInfo
+OMR::ResolvedMethodSymbol::getOSRByteCodeInfo(TR::Node *node)
```
andrewcraik: Should this return a reference since the nodes return a reference to save compile-time and memory?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
ncough: During ILGen, anchored loads are generated to avoid the side effects of storing the stack into pending push temps. These anchored loads can exist between the OSR point and the induction point, so it is necessary to skip them.

ncough:
The API has been updated to use reference types for ByteCodeIndex functions.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
andrewcraik: Ok this is surprising since most loads are not anchored usually - could you please put a comment to the effect of what you stated into the code?

andrewcraik:
@ncough request for you to document the anchored load convention in the code with a comment since it is surprising. Otherwise I think this is ready for the next reviewer.

andrewcraik:
My comment request was actually handled in the function doxygen and I missed it - thanks @ncough. Over to @vijaysun-omr now.

andrewcraik:
LGTM @vijaysun-omr

PR #960, opened 2017-03-31
WIP: Add assert for unknown OSR points

ncough:
This change explicitly adds checks for OSR points that are calls,
as in the past it would default to a call if no other type of
point was found. It will now default to preExecutionOSR behaviour
and assert.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -787,11 +787,16 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    if (self()->getOSRTransitionTarget() != TR::postExecutionOSR)
       return 0;
 
+   if ((node->getOpCodeValue() == TR::treetop || node->getOpCode().isCheck()) && node->getFirstChild()->getOpCode().isCall())
+      return 3;
+
    switch (node->getOpCodeValue())
       {
       case TR::monent: return 1;
       case TR::asynccheck: return 0;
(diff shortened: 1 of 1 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -787,11 +787,16 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    if (self()->getOSRTransitionTarget() != TR::postExecutionOSR)
       return 0;
 
+   if (node->getOpCodeValue() == TR::treetop || node->getOpCode().isCheck())
```
vijaysun-omr: "This change explicitly adds checks for OSR points that are calls"....um, where is the explicit check for a call ? I only see us checking the treetop node for opcodes that would be above calls.

ncough: Throughout OSR functions an assumption is made that a treetop or check will have a call as a child, as this is the only treetop/check that can pass isPotentialOSRPoint. As this was the same assumption that resulted in the defaults this commit is attempting to address, I will add the check for calls as well.

andrewcraik:
I'm not sure if you really want isCall here - there are a number of nodes that respond yes to isCall that are not yield points... At the very least some helpers don't need to be included nor do arraycopys for example.

mstoodle:
@ncough Is this change still progressing? If not, could we close this PR and get you to submit a new one once the change is ready?

PR #961, opened 2017-03-31
Trigger a tracepoint for VM hooks that exceed a threshold

LinHu2016:
- "time threshold" = 100 milliseconds
 - new function J9HookRegisterWithCallSite(), pass CALLSITE to
J9HookRecord via J9HookRegisterWithCallSite()
 - use portLib to calculate elapsed time for callback function
 - add tracepoint in J9Hookable
 - collect hook information for hook dump section

### DIFF
--- a/gc/base/OMRVMInterface.cpp
+++ b/gc/base/OMRVMInterface.cpp
@@ -67,8 +67,8 @@ GC_OMRVMInterface::initializeExtensions(MM_GCExtensionsBase *extensions)
 {
 	J9HookInterface** mmPrivateHooks = J9_HOOK_INTERFACE(extensions->privateHookInterface);
 
-	(*mmPrivateHooks)->J9HookRegister(mmPrivateHooks, J9HOOK_MM_PRIVATE_WALK_HEAP_START, hookWalkHeapStart, NULL);
-	(*mmPrivateHooks)->J9HookRegister(mmPrivateHooks, J9HOOK_MM_PRIVATE_WALK_HEAP_END, hookWalkHeapEnd, NULL);
+	(*mmPrivateHooks)->J9HookRegisterWithCallSite(mmPrivateHooks, J9HOOK_MM_PRIVATE_WALK_HEAP_START, hookWalkHeapStart, OMR_GET_CALLSITE(), NULL);
+	(*mmPrivateHooks)->J9HookRegisterWithCallSite(mmPrivateHooks, J9HOOK_MM_PRIVATE_WALK_HEAP_END, hookWalkHeapEnd, OMR_GET_CALLSITE(), NULL);
 }
 
 /**
--- a/include_core/omrhookable.h
+++ b/include_core/omrhookable.h
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 1991, 2016
+ * (c) Copyright IBM Corp. 1991, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -37,7 +37,11 @@ extern "C" {
 /* derives the common interface pointer from a module-specific hook interface */
 #define J9_HOOK_INTERFACE(interface) (&(interface).common.hookInterface)
 
+#define J9HOOK_LIB_CONTROL_TRACE_START "trace_start"
+#define J9HOOK_LIB_CONTROL_TRACE_STOP "trace_stop"
 
+intptr_t
+omrhook_lib_control(const char *key, uintptr_t value);
 
 struct J9HookInterface; /* Forward struct declaration */
 typedef void (*J9HookFunction)(struct J9HookInterface **hookInterface, uintptr_t eventNum, void *eventData, void *userData); /* Forward struct declaration */
@@ -46,6 +50,7 @@ typedef struct J9HookInterface {
 	intptr_t (*J9HookDisable)(struct J9HookInterface **hookInterface, uintptr_t eventNum);
 	intptr_t (*J9HookReserve)(struct J9HookInterface **hookInterface, uintptr_t eventNum);
 	intptr_t (*J9HookRegister)(struct J9HookInterface **hookInterface, uintptr_t eventNum, J9HookFunction function, void *userData, ...);
+	intptr_t (*J9HookRegisterWithCallSite)(struct J9HookInterface **hookInterface, uintptr_t eventNum, J9HookFunction function, const char *callsite, void *userData, ...);
 	void (*J9HookUnregister)(struct J9HookInterface **hookInterface, uintptr_t eventNum, J9HookFunction function, void *userData);
 	void (*J9HookShutdownInterface)(struct J9HookInterface **hookInterface);
 	intptr_t (*J9HookIsEnabled)(struct J9HookInterface **hookInterface, uintptr_t eventNum);
@@ -65,12 +70,34 @@ typedef struct J9HookInterface {
 #define J9HOOK_AGENTID_DEFAULT  ((uintptr_t)1)
 #define J9HOOK_AGENTID_LAST  ((uintptr_t)-1)
 
+/* time threshold (=100 milliseconds) for triggering the tracepoint  */
+#define OMRHOOK_DEFAULT_THRESHOLD_IN_MILLISECONDS_WARNING_CALLBACK_ELAPSED_TIME	100
+
+/* array of OMREventInfo4Dump is added in individual hookInterface by Hook generation tool to avoid
+   rumtime native memory allocation(malloc), use this macro to access &infos4Dump[event] */
+#define J9HOOK_DUMPINFO(interface, event) (&((OMREventInfo4Dump *)&((uint8_t*)((interface) + 1))[(interface)->eventSize])[event])
+
+typedef struct OMRHookInfo4Dump {
+	const char *callsite;
+	void * func_ptr;		/* use function pointer instead, if callsite == NULL */
+	uint64_t startTime;
+	uint64_t duration;
+}OMRHookInfo4Dump;
+
+typedef struct OMREventInfo4Dump {
+	struct OMRHookInfo4Dump longestHook;
+	struct OMRHookInfo4Dump lastHook;
+}OMREventInfo4Dump;
+
 typedef struct J9CommonHookInterface {
 	struct J9HookInterface *hookInterface;
 	uintptr_t size;
 	omrthread_monitor_t lock;
 	struct J9Pool *pool;
 	uintptr_t nextAgentID;
+	struct OMRPortLibrary *portLib;		/* for accessing PortLibrary  */
+	uint64_t threshold4Trace;			/* the threshold for triggering tracepoint */
+	uintptr_t eventSize;				/* how many events supported by this hook interface */
 } J9CommonHookInterface;
 
 
@@ -82,6 +109,7 @@ typedef struct J9CommonHookInterface {
 typedef struct J9HookRecord {
 	struct J9HookRecord *next;
 	J9HookFunction function;
+	const char *callsite;
 	void *userData;
 	uintptr_t count;
 	uintptr_t id;
--- a/omr/omrtrcinit.c
+++ b/omr/omrtrcinit.c
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2014, 2016
+ * (c) Copyright IBM Corp. 2014, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -22,6 +22,7 @@
 #include "omrport.h"
 #include "omrrasinit.h"
 #include "omrtrace.h"
+#include "omrhookable.h"
 #include "ut_omrvm.h"
 
 #undef UT_MODULE_LOADED
@@ -98,6 +99,9 @@ omr_ras_initTraceEngine(OMR_VM *omrVM, const char *traceOptString, const char *d
 	/* Load module for thread library tracepoints */
 	omrthread_lib_control(J9THREAD_LIB_CONTROL_TRACE_START, (uintptr_t)omrVM->_trcEngine->utIntf);
 
+	/* Load module for hookable library tracepoints */
+	omrhook_lib_control(J9HOOK_LIB_CONTROL_TRACE_START, (uintptr_t)omrVM->_trcEngine->utIntf);
+
 	UT_OMRVM_MODULE_LOADED(omrVM->_trcEngine->utIntf);
 	UT_OMRTI_MODULE_LOADED(omrVM->_trcEngine->utIntf);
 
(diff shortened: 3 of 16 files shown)


include_core/omrhookable.h
```diff
@@ -65,12 +70,35 @@ typedef struct J9HookInterface {
...
 #define J9HOOK_AGENTID_LAST  ((uintptr_t)-1)
 
+/* time threshold (=100 milliseconds) for triggering the tracepoint  */
+#define DEFAULT_THRESHOLD_IN_MILLISECONDS_WARNNING_CALLBACK_ELAPSED_TIME	100
+
+/* for generating Hook dump section */
+#define INITIALIZE_HOOKDUMP(interface) {interface.common.eventSize = sizeof(interface.flags) / sizeof(interface.flags[0]); interface.common.infos4Dump = &interface.infos4Dump[0];}
+
+
+typedef struct HookInfo4Dump {
```
DanHeidinga: Struct should be prefixed with 'OMR' to prevent potential name conflicts.

include_core/omrhookable.h
```diff
@@ -65,12 +70,35 @@ typedef struct J9HookInterface {
...
+	const char *callsite;
+	void * func_ptr;		/* use function pointer instead, if callsite == NULL */
+	uint64_t startTime;
+	uint64_t duration;
+}HookInfo4Dump;
+
+typedef struct EventInfo4Dump {
```
DanHeidinga: Struct should be prefixed with 'OMR' to prevent potential name conflicts.

util/hookable/hookable.cpp
```diff
@@ -78,6 +82,28 @@ static J9CONST_TABLE J9HookInterface hookFunctionTable = {
...
 #define HOOK_VALID_ID(id) ( (((id) | 1) + 1) )
 
+
+intptr_t
+omrhook_lib_control(const char *key, uintptr_t value)
+{
+	intptr_t rc = -1;
+
+	if (0 != value) {
+#if defined(OMR_RAS_TDF_TRACE)
```
DanHeidinga: With the way this is written, the function will fail 100% of the time if `OMR_RAS_TDF_TRACE` is not defined.

gc/base/OMRVMInterface.cpp
```diff
@@ -81,6 +81,15 @@ GC_OMRVMInterface::getOmrHookInterface(MM_GCExtensionsBase *extensions)
 }
 
 /**
+ * Return the OMR private hook interface for the Memory Manager
+ */
+J9HookInterface**
+GC_OMRVMInterface::getOmrPrivateHookInterface(MM_GCExtensionsBase *extensions)
+{
```
DanHeidinga: Is this really what you want to do?  It exposes the private hook interface to everyone.... which means it isn't private anymore.

include_core/omrhookable.h
```diff
@@ -65,12 +70,35 @@ typedef struct J9HookInterface {
 #define J9HOOK_AGENTID_DEFAULT  ((uintptr_t)1)
 #define J9HOOK_AGENTID_LAST  ((uintptr_t)-1)
 
+/* time threshold (=100 milliseconds) for triggering the tracepoint  */
+#define DEFAULT_THRESHOLD_IN_MILLISECONDS_WARNNING_CALLBACK_ELAPSED_TIME	100
```
DanHeidinga: This should be namespaced as well - either 'OMR' or 'J9' (to match existing constants in the file)

ie: `OMRHOOK_DEFAULT_THRESHOLD_IN_MILLISECONDS_WARNNING_CALLBACK_ELAPSED_TIME`

include_core/omrhookable.h
```diff
@@ -65,12 +70,35 @@ typedef struct J9HookInterface {
...
+
+/* for generating Hook dump section */
+#define INITIALIZE_HOOKDUMP(interface) {interface.common.eventSize = sizeof(interface.flags) / sizeof(interface.flags[0]); interface.common.infos4Dump = &interface.infos4Dump[0];}
```
DanHeidinga: This needs to be namespaced as well.

Multi-line macros are usually embedded in `do{ }while(0)` loops.

util/hookable/hookable.cpp
```diff
@@ -182,7 +209,42 @@ J9HookDispatch(struct J9HookInterface **hookInterface, uintptr_t taggedEventNum,
...
+				}
+
+				if (commonInterface->threshold4Trace <= timeDelta) {
+					const char *callsite = "UNKNOWN";
+					char buffer[32];
+					if (NULL != record->callsite) {
+						callsite = record->callsite;
+					} else {
+						/* if the callsite info can not be retrieved, use callback function pointer instead  */
+						omrstr_printf(buffer, sizeof(buffer), "0x%p", record->function);
```
DanHeidinga: Is this supposed to do `callsite = buffer` here?

util/hookable/hookable.cpp
```diff
@@ -342,6 +404,115 @@ J9HookRegister(struct J9HookInterface **hookInterface, uintptr_t taggedEventNum,
 					(emptyRecord->next->agentID >= agentID)))
 		) {
 			emptyRecord->function = function;
+			emptyRecord->callsite = NULL;
```
DanHeidinga: Rather than modifying both `J9HookRegister` and the new `J9HookRegisterWithCallSite`, would it be better to have the old function call the new one passing 'NULL' as the callsite?  

This would mean there aren't two implementations that need to be kept in sync.

include_core/omrhookable.h
0xdaryl: Also, WARNNING is spelled wrong.

DanHeidinga:
Apologies if I've missed it, but it took a bit of work to track down how data structures all fit together.  Can you add a description of the layout of these structures to the pull request (or better even, in the code)?

ie: How 'EventInfo4Dump' gets connected into the hook interface, etc.

gc/base/GCExtensionsBase.cpp
```diff
@@ -182,10 +182,12 @@ MM_GCExtensionsBase::initialize(MM_EnvironmentBase* env)
 	if (J9HookInitializeInterface(getPrivateHookInterface(), OMRPORTLIB, sizeof(privateHookInterface))) {
 		goto failed;
 	}
+	INITIALIZE_HOOKDUMP(privateHookInterface);
```
DanHeidinga: Why is the `INITIALIZE_HOOKDUMP` macro needed?  Could you roll it into the `J9HookInitializeInterface()` call?  

It's easier for consumers of this code if they don't have to do anything special after calling `J9HookInitializeInterface` as they may forget / not know about the extra macro step.

util/hookable/hookable.cpp
LinHu2016: J9HookRegister were used in omr, j9vm and jit, in order to smooth migration from J9HookRegister() to J9HookRegisterWithCallSite(), we temporarily keep old interface J9HookRegister to avoid dependence issue between omr and other components.

DanHeidinga: Right  - so why not have J9HookRegister forward to J9HookRegisterWithCallSite?

```
J9HookRegister(...) {
   return J9HookRegisterWithCallSite(......, NULL);
}
```
So you only have 1 implementation?

gc/base/GCExtensionsBase.cpp
LinHu2016: array of OMREventInfo4Dump is added in individual hookInterface by Hook generation code to avoid rumtime native memory allocation(malloc), but hookable code is not aware of specific Hook interface except J9CommonHookInterface, the macro OMRHOOK_INITIALIZE_HOOKDUMP try to pass array pointer and size from the  individual hookInterface to J9CommonHookInterface, then hookable code and javadump code can easily access OMREventInfo4Dump.

util/hookable/hookable.cpp
LinHu2016: you are right, I will update it to one implementation. Thanks

gc/base/GCExtensionsBase.cpp
DanHeidinga:
```
 	if (J9HookInitializeInterface(getPrivateHookInterface(), OMRPORTLIB, sizeof(privateHookInterface))) {
 		goto failed;
 	}
+	INITIALIZE_HOOKDUMP(privateHookInterface);
```
Isn't `getPrivateHookInterface()` the same as `privateHookInterface`?  If they are, then `J9HookInitializeInterface` should be able to do the "right" thing.

gc/base/OMRVMInterface.cpp
LinHu2016: I need a simple way to access all of hook interfaces from javadump code like below, 
any suggestion for access omr private hook interface? Thanks,

j9JavaVM->memoryManagerFunctions->j9gc_get_omr_hook_interface(j9JavaVM->omrVM)
j9JavaVM->memoryManagerFunctions->j9gc_get_omr_private_hook_interface(j9JavaVM->omrVM)
j9JavaVM->memoryManagerFunctions->j9gc_get_hook_interface(j9JavaVM)

j9JavaVM->internalVMFunctions->getVMHookInterface(j9JavaVM)
zip_getVMZipCachePoolHookInterface((J9ZipCachePool *) j9JavaVM->zipCachePool)
j9JavaVM->internalVMFunctions->getJITHookInterface(j9JavaVM)

DanHeidinga: Can you expose the `getOmrPrivateHookInterface` on the J9 side without making access to it part of the api?

gc/base/GCExtensionsBase.cpp
LinHu2016: yes, getPrivateHookInterface() == privateHookInterface, but J9HookInitializeInterface() cast it to common pointer ( struct J9HookInterface ** ), I could not figure out which HookInteface inside method J9HookInitializeInterface()

charliegracie:
@LinHu2016 I do not see any users of J9HookRegister anymore.  All callers seem to have been updated to J9HookRegisterWithCallSite.  If I am not missing anything I would like to suggest that instead of adding a new function you just modify the original one.

LinHu2016:
@charliegracie , Hi Charlie, JIT still use J9HookRegister, also there is dependence between omr and j9vm/jit, j9vm/jit changes need wait omr's change to promote, keeping old interface is just for change  transition, finally we will remove old interface, might be mark it as deprecated for now

gc/base/OMRVMInterface.cpp
LinHu2016: Thanks Dan, move the function to modronapi(j9gc_)

gc/base/GCExtensionsBase.cpp
LinHu2016: typical HookInterface like below
typedef struct xxxxHookInterface {
	struct J9CommonHookInterface common;
	U_8 flags[xx];
	struct EventInfo4Dump infos4Dump[xx];
	J9HookRecord* hooks[xx];
} xxxxHookInterface;

to access flags and hooks, we use the below macros, accessing flags from the end of "common" interface, and accessing hooks from end of the interface in descending order, without knowing array size, we can not locate start address of EventInfo4Dump array.  create new macro J9HOOK_INITIALIZE_INTERFACE_AND_DUMP instead of J9HookInitializeInterface() to avoid missing the extra macro step. 

/* flags are stored at the beginning of the interface just after the common interface fields in ascending order */
#define HOOK_FLAGS(interface, event) (((uint8_t*)((interface) + 1))[event])
/* records are stored at the END of the interface in descending order */
#define HOOK_RECORD(interface, event) (((J9HookRecord**)( (uint8_t*)(interface) + (interface)->size ))[ -1 - (event)])

include_core/omrhookable.h
```diff
@@ -65,12 +70,56 @@ typedef struct J9HookInterface {
...
+#define OMRHOOK_DEFAULT_THRESHOLD_IN_MILLISECONDS_WARNING_CALLBACK_ELAPSED_TIME	100
+
+/*
+ * array of OMREventInfo4Dump is added in individual hookInterface by Hook generation tool to avoid
+ * rumtime native memory allocation(malloc), this macro would pass the array pointer and size from
+ * the individual hookInterface to J9CommonHookInterface, then hookable code and javadump code can
+ * easily access OMREventInfo4Dump[].
+ */
+#define OMRHOOK_INITIALIZE_HOOKDUMP(interface) \
+	do { \
```
charliegracie: Is this still called or can it be removed?

include_core/omrhookable.h
```diff
@@ -65,12 +70,56 @@ typedef struct J9HookInterface {
...
+		interface.common.eventSize = sizeof(interface.flags) / sizeof(interface.flags[0]); \
+		interface.common.infos4Dump = &interface.infos4Dump[0]; \
+	} while (0)
+
+/* this macro would initialize hookInterface and dump struct for hook */
+#define J9HOOK_INITIALIZE_INTERFACE_AND_DUMP(hookInterface, portLib) ({ \
+	intptr_t result = 0; \
+	do { \
+		J9HookInitializeInterface(J9_HOOK_INTERFACE(hookInterface), portLib, sizeof(hookInterface)); \
+		if (0 == result) { \
```
charliegracie: If you look at the macros for HOOK_FLAGS and HOOK_RECORD could you do something similar so we do not have to have the extra infos4Dump pointer and so we do not need this code.

omr/omrtrcinit.c
```diff
@@ -22,6 +22,7 @@
 #include "omrport.h"
 #include "omrrasinit.h"
 #include "omrtrace.h"
+#include "omrhookable.h"
```
charliegracie: Can you update the copyright header for this file?

util/hookable/hookable.cpp
charliegracie: @LinHu2016 does your thumbs up mean you are going to change this or that the failure is expected?

include_core/omrhookable.h
```diff
@@ -65,12 +70,56 @@ typedef struct J9HookInterface {
...
+
+/*
+ * array of OMREventInfo4Dump is added in individual hookInterface by Hook generation tool to avoid
+ * rumtime native memory allocation(malloc), this macro would pass the array pointer and size from
+ * the individual hookInterface to J9CommonHookInterface, then hookable code and javadump code can
+ * easily access OMREventInfo4Dump[].
+ */
+#define OMRHOOK_INITIALIZE_HOOKDUMP(interface) \
+	do { \
+		interface.common.eventSize = sizeof(interface.flags) / sizeof(interface.flags[0]); \
```
charliegracie: eventSize could be set during the generation of the hook structure.

util/hookable/hookable.cpp
LinHu2016: I am going to change the code.

include_core/omrhookable.h
LinHu2016: new macro J9HOOK_DUMPINFO similar with HOOK_FLAGS, so canceled OMRHOOK_INITIALIZE_HOOKDUMP

omr/omrtrcinit.c
LinHu2016: updated copyright date in header

include_core/omrhookable.h
LinHu2016: done

include_core/omrhookable.h
LinHu2016: removed

LinHu2016:
@charliegracie , Hi Charlie, all of changes are done, Could you please review it again, Thanks

util/hookable/hookable.cpp
charliegracie: @DanHeidinga are you ok with this code now?

DanHeidinga: Yes, lgtm

mgaudet:
It's interesting to note that in GitHub it's hard to see, but this contained a couple of files where the line-endings were totally rewritten. ie `util/hookable/hookable.exportlist` became CLRF, and `util/hookable/j9hook.tdf` was added with CLRF line endings. 

I don't know if we've nailed down rules about line endings, but unix style easily dominates in the project.

PR #962, opened 2017-03-31
Add isDebuggingMode query to compilation

cathyzhyi:
There are 2 debugging modes, mimic interpreter frame shape
and full speed debug based on OSR. Those 2 are aliased
until recently. Some code checking for mimic interpreter frame
shape is actually asking whether we are in debugging mode and
should be using the new query after the 2 debugging modes are
dis-aliased.

### DIFF
--- a/compiler/arm/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/arm/codegen/OMRTreeEvaluator.cpp
@@ -1346,8 +1346,8 @@ TR::Register *OMR::ARM::TreeEvaluator::BBStartEvaluator(TR::Node *node, TR::Code
    TR::Node * fenceNode = TR::Node::createRelative32BitFenceNode(node, &block->getInstructionBoundaries()._startPC);
    TR::Instruction * fence = generateAdminInstruction(cg, ARMOp_fence, node, deps, fenceNode);
 
-   if (block->isCatchBlock() && comp->getOption(TR_FullSpeedDebug))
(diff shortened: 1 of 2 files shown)


andrewcraik:
looks like a good change to decouple specific debugging implementations from queries where we just want to know if we are in a debug mode of some kind.

compiler/arm/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1346,8 +1346,8 @@ TR::Register *OMR::ARM::TreeEvaluator::BBStartEvaluator(TR::Node *node, TR::Code
    TR::Node * fenceNode = TR::Node::createRelative32BitFenceNode(node, &block->getInstructionBoundaries()._startPC);
    TR::Instruction * fence = generateAdminInstruction(cg, ARMOp_fence, node, deps, fenceNode);
 
-   if (block->isCatchBlock() && comp->getOption(TR_FullSpeedDebug))
-      fence->setNeedsGCMap(); // a catch entry is a gc point in FSD mode
+   if (block->isCatchBlock() && comp->isDebuggingMode())
```
vijaysun-omr: So the thinking here is that this action would be needed even if we were not in "full speed debug" mode but were in "mimic interpreter shape" mode ? If so, could you elaborate on what it is about this spot in the code that would make it applicable to when we are mimicking interpreter frame shape (regardless of whether we were debugging or not). While I understand that mimic interpreter frame shape is only used under full speed debug currently, is it valid to consider the former as implying that we are in some sort of "debug mode" ?

PR #963, opened 2017-03-31
Add missing ARM constant re-materialization queries

NigelYiboYu:
Add missing constant re-materialization threshold queries for ARM
to avoid build failures.

### DIFF
--- a/compiler/arm/codegen/OMRCodeGenerator.hpp
+++ b/compiler/arm/codegen/OMRCodeGenerator.hpp
@@ -162,6 +162,11 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    void apply24BitLabelRelativeRelocation(int32_t * cursor, TR::LabelSymbol *);
    void apply8BitLabelRelativeRelocation(int32_t * cursor, TR::LabelSymbol *);
 
+   // ARM specific thresholds for constant re-materialization
+   int64_t getLargestNegConstThatMustBeMaterialized() {return -32769;}  // minimum 16-bit signed int minus 1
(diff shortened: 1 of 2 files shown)


compiler/arm/codegen/OMRCodeGenerator.hpp
```diff
@@ -162,6 +162,10 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    void apply24BitLabelRelativeRelocation(int32_t * cursor, TR::LabelSymbol *);
    void apply8BitLabelRelativeRelocation(int32_t * cursor, TR::LabelSymbol *);
 
+   // ARM specific thresholds for constant re-materialization
+   int64_t getLargestNegConstThatMustBeMaterialized() { TR_ASSERT(0, "Not Implemented on ARM"); return 0; }
```
andrewcraik: If you are implementing these methods on arm why are these asserting - they should return a sensible value and not assert otherwise why are we creating an implementation?

NigelYiboYu: Will change this so that ARM and PPC have the same returns.

vijaysun-omr:
Safe/simple change that adds missing query for ARM. Merging.

vijaysun-omr:
Actually, before I merge this, one question : what is the rationale for the particular values you have chosen for ARM wrt smallest positive value and largest negative value that can be materialized ?

NigelYiboYu:
Similar to PPC, loading 16bit number on ARM is considered cheap. And loading anything outside of 16bit signed range is considered expensive. That's why the ARM implementation is the same as PPC.

PR #964, opened 2017-03-31
Enhance the linter to better understand calls

mgaudet:
The linter needs to catch calls that are missing downcasts. What happens
however when the 'downcast' would really resolve to the current type?
This is a case we have encountered in the codebase where a
most-derived concrete TR:: type makes a member function call (see PR#913).

This PR resolves this by teaching the linter to check whether or not a
call resolves to the most derived type, and the current type is the most
derived type. 

Some refactoring is also provided.


mgaudet:
Going to address @Leonardo2718's concerns.

mgaudet:
I've addressed Leo's concerns (which were valid! Dropping that test caused false-negatives to escape!)

I'm going to leave this in WIP until I can do a little more robust testing however, which will probably be Monday.

mgaudet:
Unmarked as WIP after testing.

PR #965, opened 2017-03-31
WIP: Add profileInfoIsUntrustworthy to OMR::Node

JamesKingdon:
The profileInfoIsUntrustworthy flag can be used to transfer
information between optimisations to avoid generating code that
depends on profiling data that is unlikely to be good.

The following functions have been added relating to the new flag:

   bool profileInfoCannotBeTrusted();
   void setProfileInfoCannotBeTrusted();
   const char * printIsProfileInfoCannotBeTrusted();
   void setProfileInfoCannotBeTrusted(bool v);
   bool canCheckIfProfileInfoCannotBeTrusted();

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -5173,6 +5173,27 @@ OMR::Node::printIsNonZero()
    return self()->isNonZero() ? "X!=0 " : "";
    }
 
+bool                   OMR::Node::profileInfoCannotBeTrusted()    { return _flags.testAny(profileInfoIsUntrustworthy); }
+void                   OMR::Node::setProfileInfoCannotBeTrusted() {  _flags.set(profileInfoIsUntrustworthy); }
+const char *           OMR::Node::printIsProfileInfoCannotBeTrusted() { return self()->profileInfoCannotBeTrusted() ? "profileInfoIsUntrustworthy " : ""; }
+ 
+void OMR::Node::setProfileInfoCannotBeTrusted(bool v)
+       {
+       TR::Compilation * c = TR::comp();
+       if (performNodeTransformation2(c,"O^O NODE FLAGS: Setting profileInfoIsUntrustworthy flag on node %p to %d\n", this, v))
+          _flags.set(profileInfoIsUntrustworthy, v);
+       }
+ 
+bool OMR::Node::canCheckIfProfileInfoCannotBeTrusted()
+       {
+       switch (self()->getOpCodeValue())
+          {
+          case TR::instanceof:
+             return true;
+          default:
+             return false;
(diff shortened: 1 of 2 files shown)


compiler/il/OMRNode.cpp
```diff
@@ -5173,6 +5173,27 @@ OMR::Node::printIsNonZero()
...
+const char *           OMR::Node::printIsProfileInfoCannotBeTrusted() { return self()->profileInfoCannotBeTrusted() ? "profileInfoIsUntrustworthy " : ""; }
+ 
+void OMR::Node::setProfileInfoCannotBeTrusted(bool v)
+       {
+       TR::Compilation * c = TR::comp();
+       if (performNodeTransformation2(c,"O^O NODE FLAGS: Setting profileInfoIsUntrustworthy flag on node %p to %d\n", this, v))
+          _flags.set(profileInfoIsUntrustworthy, v);
+       }
+ 
+bool OMR::Node::canCheckIfProfileInfoCannotBeTrusted()
```
fjeremic: This seems like something that should return false by default and an extensible class should implement the below logic. This seems like "J9-ish" type logic in the base `OMR::Node` class. Do others agree?

fjeremic:
@JamesKingdon can you provide an explicit use case in which using such a node flag is better than just not providing the said untrusty profiling info in the first place?

ymanton:
I had the same thought as @fjeremic, this feels like a problem that can be solved better locally.

Can you simply throw away or otherwise invalidate the compilation's copy of the profiling info once you've made a transformation that makes it invalid and relieve the rest of the code base from having to know anything about this?

andrewcraik:
There could be a valid use for this - it is possible to have two or more copies of a node and you may want to indicate that a given copy does not reflect the profiling info gathered, but you don't want to throw it away for the whole compilation. The optimizer can often arrive in these situations - the isCold block flag can highlight blocks that suffer from this problem. @JamesKingdon should justify the change, but having it in OMR and having at the top level is not necessarily wrong if it is addressing a problem created by an optimization cloning nodes or transforming the CFG.

ymanton:
I can see that happening, so throwing away the data is a bad idea. But I think in this case the J9 JIT profiler has enough info to deal with its problem on its own; it knows the node opcode, it knows the meaning of the source bytecode, and it knows its own limitations and that it shouldn't be giving back profiling info for an `instanceof` node when the original bytecode was something else.

In general I don't think exposing the possibility of incorrect profiling information and shifting the responsibility of sanitizing it to potentially every consumer of IL is a good idea unless our hands are tied. Ideally even the original opt doing this transformation shouldn't have to know anything about its effects on profiling data.

PR #968, opened 2017-04-03
Update register dependencies of arraytranslate

jingge-ibm:
Add the allocation of XMM3, which was used and but not allocated before.
Remove the allocation of termChar in OT case, as it is not used in the
new helper

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -3242,10 +3242,10 @@ TR::Register *OMR::X86::TreeEvaluator::arraytranslateEvaluator(TR::Node *node, T
    // arraytranslate
    //    input ptr
    //    output ptr
-   //    translation table (dummy)
-   //    terminal character (dummy when src is byte and dest is word, otherwise, it's a mask)
+   //    translation table (not used)
+   //    terminal character (not used when src is byte and dest is word, otherwise, it's a mask)
    //    input length (in elements)
-   //    stopping char (dummy for X)
+   //    stopping char (not used for X)
    // Number of elements translated is returned
    //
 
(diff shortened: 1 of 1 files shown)


compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3255,19 +3255,24 @@ TR::Register *OMR::X86::TreeEvaluator::arraytranslateEvaluator(TR::Node *node, T
    TR::Register *srcPtrReg, *dstPtrReg, *transTableReg, *termCharReg, *lengthReg;
    bool stopUsingCopyReg1 = TR::TreeEvaluator::stopUsingCopyRegAddr(node->getChild(0), srcPtrReg, cg);
    bool stopUsingCopyReg2 = TR::TreeEvaluator::stopUsingCopyRegAddr(node->getChild(1), dstPtrReg, cg);
-   bool stopUsingCopyReg4 = TR::TreeEvaluator::stopUsingCopyRegInteger(node->getChild(3), termCharReg, cg);
+   bool stopUsingCopyReg4 = false;
+   if (!sourceByte)
+      {
+      stopUsingCopyReg4 = TR::TreeEvaluator::stopUsingCopyRegInteger(node->getChild(3), termCharReg, cg);
+      }
```
0xdaryl: Before your change child(3) was always evaluated.  Now its only evaluated conditionally if `sourceByte` is false.  If `sourceByte` is true its reference count is simply decremented later.  Is that what you intended?  If so, please explain.  There is a comment nearby that says if the source is a byte then child(3) is a "dummy".  I'm not sure what that means, but I'll assume that's why you skip the evaluation.  But please do clarify.

jingge-ibm: "dummy" means that the value is not used in the helper (It is a free register). In the old helper, when sourceBye is false,  we overwrite child(3) register in the very beginning  and we just use it as a free register. But in the new helper, I realized that we have other free registers to use so we don't actually need to evaluate child(3).

0xdaryl: Normally, an opcode evaluator needs to evaluate and decrement the reference count on all its children.  I've seen many problems over the years where this wasn't followed and one problem or another was introduced.  Exceptions to this rule can (and have) been made in the past in controlled circumstances, and this opcode may very well be one of those.

But I'm seeing the code for this evaluator for the first time and I don't have deep familiarity with the opcode.  If you're confirming that skipping the evaluation of the 3rd child is correct in this case then I'm fine with your assessment.

I'm troubled by the implementation of this opcode, but that's not your problem... :-)

Issue #939, opened 2017-03-28
Deprecate isLiteralPoolOnDemandOn in the z Systems codegen


fjeremic:
The `isLiteralPoolOnDemandOn` is a z Systems particular codegen query. Historically it was used to lock a particular register and designate it to hold the literal pool address throughout the lifetime of the compilation.

On modern z/Architecture, more specifically with the introduction of load relative instructions in z10, we have the ability to load values from the literal pool with a single relative load. The relative loads, LRL for example, are RIL type instructions meaning we can address a literal pool that it at most 4GB away. With the deprecation of tiered code cache it is inconceivable that the literal pool cannot be addressed with relative load instructions. As such, we always benefit from freeing up the literal pool register for assignment as the register pressure decreases at no cost to our ability to load values from the literal pool.

In fact if the `DynamicLiteralPool` optimization runs, which is the case in pretty much most optimization strategies we will always free the literal pool register. As such I propose we deprecate this the ability to lock the literal pool register and prevent it from being assigned as it frankly does not provide any benefit currently.

I would like to hear some opinions or alternative viewpoints on this issue from subject matter experts. @joransiu @ebadyano @NigelYiboYu any concerns regarding this?

lmaisons:
Pardon my ignorance, but how did tiered code cache affect this?  As far as I know, code caches were never anywhere near 4G large, and thus anywhere in both hot and cold areas should have always been reachable within 4G.

I guess I'm worried your thinking instead of trampolines, which could potentially still be needed.

fjeremic:
@lmaisons if I'm understanding this correctly tiered code caches were code caches that were allocated separately. We placed "cold" code in the cold cache which is where the snippets would end up. As such we could run into a scenario where the code caches themselves got allocated with a gap of 4GB. They don't necessarily have to be 4GB large. Do I understand this correctly?

joransiu:
I suspect the two aspects of Tiered Code Cache (historically) is breeding some confusion here.

[...]

PR #969, opened 2017-04-03
Enable dynamic literal pool by default

fjeremic:
On modern z/Architecture, more specifically with the introduction of
load relative instructions in z10, we have the ability to load values
from the literal pool with a single relative load. The relative loads,
LRL for example, are RIL type instructions meaning we can address a
literal pool that it at most 4GB away. With the deprecation of tiered
code cache it is inconceivable that the literal pool cannot be addressed
with relative load instructions. As such, we always benefit from freeing
up the literal pool register for assignment as the register pressure
decreases at no cost to our ability to load values from the literal
pool.

Issue: #939

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -647,7 +647,7 @@ OMR::Z::CodeGenerator::CodeGenerator()
          self()->setSupportsHighWordFacility(true);
       }
 
-   self()->setOnDemandLiteralPoolRun(false);
+   self()->setOnDemandLiteralPoolRun(true);
    self()->setGlobalStaticBaseRegisterOn(false);
 
    self()->setGlobalPrivateStaticBaseRegisterOn(false);


fjeremic:
Note to reviewers:

As discussed in #939 we will eventually deprecate the dynamic literal pool. The first step in this direction is to enable the dynamic literal pool by default to get additional testing coverage on downstream projects consuming OMR. Once we have adequate testing complete I will be forging ahead and deprecating the notion of dynamic literal pool all together.

vijaysun-omr:
So. the intent would be deprecate the ability to lock a register for literal pool after you have turned on the optimization that allows you to do so unconditionally. i.e. after this change is merged, we would no longer be locking a register for the literal pool in practice although the ability to do so would still exist in the compiler codebase (to be deprecated soon afterwards based on testing results) ?

fjeremic:
@vijaysun-omr not quite. Currently the literal pool register is locked by default unless some optimization (in a downstream project consuming OMR) chooses to unlock it. With this change we no longer rely on an optimization to run in order for the literal pool register to be unlocked. As discussed in #939 on modern z/Architecture and due to the OMR ALS locking the literal pool register is undesirable in all scenarios.

vijaysun-omr:
Okay so this setting itself unlocks the register that used to be reserved for the lit pool. Thanks for the clarification. I'll merge this now.

PR #970, opened 2017-04-03
Free out of line spills on Power

ymanton:
Prior attempt at improving the way we handle register spills during out of line register assignment (#908) was not sufficient to deal with the case where a register is spilled between the branch and merge of an out of line section, is also spilled out of line, is subsequently unspilled out of line, and therefore must be re-spilled at the end of out of line RA in order to restore the register file to its expected state. Under these conditions we have to make sure that the unspill of the register does not free the spill slot for re-use, since we will need to forcefully re-spill the register.

This PR reverts #908 and attempts to address this problem by only unlocking the free spill list in out of line code sections in specific, safe circumstances and by freeing spill slots after out of line RA where possible.

First, during out of line RA, we free the spill slot of any register who's future use count hits 0, as dead registers should definitely have no futher need for their spill slots. In order to do this we momentarily unlock the free spill list.

Second, after out of line RA we iterate through all of the currently assigned registers and free any spill slots still held by a register, since at this point out of line RA is complete and the free spill list has been unlocked.

### DIFF
--- a/compiler/p/codegen/OMRMachine.cpp
+++ b/compiler/p/codegen/OMRMachine.cpp
@@ -55,6 +55,7 @@
 #include "p/codegen/PPCInstruction.hpp"
 #include "p/codegen/PPCOpsDefines.hpp"         // for Op_load, Op_st
 #include "ras/Debug.hpp"                       // for TR_DebugBase
+#include "env/IO.hpp"                          // for PRINTF_POINTER_FORMAT
 
 namespace TR { class AutomaticSymbol; }
 
@@ -70,6 +71,25 @@ static void registerCopy(TR::Instruction *precedingI,
                                         TR::RealRegister *sReg,
                                         TR::CodeGenerator   *cg);
 
+static int32_t spillSizeForRegister(TR::Register *virtReg)
+   {
+   switch (virtReg->getKind())
+      {
+      case TR_GPR:
+         return TR::Compiler->om.sizeofReferenceAddress();
+      case TR_FPR:
+         return virtReg->isSinglePrecision() ? 4 : 8;
+      case TR_VSX_SCALAR:
+         return 8;
+      case TR_CCR:
+         return 4;
+      case TR_VSX_VECTOR:
+      case TR_VRF:
+         return 16;
+      }
+   TR_ASSERT(false, "Unexpected register kind");
+   return 0;
+   }
 
 static const char *
 getRegisterName(TR::Register * reg, TR::CodeGenerator * cg)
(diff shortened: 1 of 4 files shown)


vijaysun-omr:
@zl-wang and @fjeremic please review.

zl-wang:
That was my suspicion previously.  Lock/unlockFreeSpillList were actually meant to be a state how backingStorage should be handled.  Free it on a per assignment basis or end-of-life basis.  This looked better.

0xdaryl:
Please say "Power" rather than "P" in your commit titles/messages.

compiler/p/codegen/OMRMachine.cpp
```diff
@@ -1994,14 +1999,60 @@ OMR::Power::Machine::decFutureUseCountAndUnlatch(TR::Register *virtualRegister)
...
+         if (trace)
             {
             traceMsg(comp,"\nOOL: %s's remaining uses are out-of-line, unlatching\n", self()->cg()->getDebug()->getName(virtualRegister));
             }
          }
       virtualRegister->getAssignedRealRegister()->setState(TR::RealRegister::Unlatched);
       virtualRegister->setAssignedRegister(NULL);
       }
+
+   // If this register is dead and we're assigning on on an out of line path
```
fjeremic: Typo: "on on"

compiler/p/codegen/OMRMachine.cpp
```diff
@@ -1994,14 +1999,60 @@ OMR::Power::Machine::decFutureUseCountAndUnlatch(TR::Register *virtualRegister)
...
+         traceMsg(comp, "\nFreeing backing storage " POINTER_PRINTF_FORMAT " of size %u from dead virtual %s\n", location, size, cg->getDebug()->getName(virtualRegister));
+      cg->unlockFreeSpillList();
+      cg->freeSpill(location, size, 0);
+      virtualRegister->setBackingStorage(NULL);
+      location->setMaxSpillDepth(0);
+      cg->lockFreeSpillList();
+      }
+   }
+
+void OMR::Power::Machine::disassociateUnspilledBackingStorage()
```
fjeremic: Could you please provide Doxygen documentation for this newly added function?

compiler/p/codegen/OMRMachine.cpp
```diff
@@ -1994,14 +1999,60 @@ OMR::Power::Machine::decFutureUseCountAndUnlatch(TR::Register *virtualRegister)
...
+         TR_BackingStore *location = virtReg->getBackingStorage();
+
+         if (location != NULL)
+            {
+            int32_t size = spillSizeForRegister(virtReg);
+            if (trace)
+               traceMsg(comp, "\nDisassociating backing storage " POINTER_PRINTF_FORMAT " of size %u from assigned virtual %s\n", location, size, cg->getDebug()->getName(virtReg));
+            cg->freeSpill(location, size, 0);
+            virtReg->setBackingStorage(NULL);
+            location->setMaxSpillDepth(0);
```
fjeremic: Could we use the max spill depth as an alternative to `lockFreeSpillList`? They seem to serve an almost identical purpose.

fjeremic:
I really wish we had a common code generation. We shouldn't have to replicate this logic across platforms and waste time fixing these super corner case bugs. Here is the z Systems implementation of protecting the spill slot in an OOL section if we have spills on either side of the ICF:

https://github.com/eclipse/omr/blob/master/compiler/z/codegen/OMRMachine.cpp#L4924

Note the use of `getMaxSpillDepth` to address the issue. As an aside this whole infrastructure is in need of an overhaul.

ymanton:
Power has the same code actually, but it's not sufficient to allow the free spill list to be unlocked during out of line RA after all. As far as I can recall that code was added to both Power and Z in multiple commits to fix multiple bugs that locking/unlocking the free spill list did not address, but the details are buried somewhere.

Rather than re-discovering how and why `maxSpillDepth` and free spill list locking interact this PR takes a much simpler approach that is easier to reason about:

1. dead registers have no need for spill slots
2. registers in the assigned state **after** out of line RA has been completed also have no need for spill slots.

The only reason either of these two have their spill slots is because the free spill list was locked at the time they should have given them up. The x86 code generator [already does `#2`](https://github.com/eclipse/omr/blob/6b2d054765c0d75ed24c1cdbf4beb919dde3c325/compiler/x/codegen/OMRMachine.cpp#L2296) and this PR deletes some code in `compiler/p/codegen/OMRRegisterDependency.cpp` that apparently was trying to do something similar but is doomed to failure because the free spill list is still locked at the point where it executes. `#1` I believe is the only change that doesn't have an equivalent in at least one of the other code generators, but it's also hard for me to see where it could possibly go wrong.

vijaysun-omr:
@fjeremic any other comments ?

fjeremic:
@vijaysun-omr no other comments.

PR #971, opened 2017-04-03
Evaluate relocatable constant child of compare equality node

r30shah:
Relocation was not provided for contant child of compare equality node
if that node is not profile guarded. This commit will fix
this issue by calling evaluate() on constant child for
it is a method pointer or a class pointer constant and
relocation is required.

### DIFF
--- a/compiler/p/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/p/codegen/ControlFlowEvaluator.cpp
@@ -1242,8 +1242,10 @@ TR::Register *OMR::Power::TreeEvaluator::compareIntsForEquality(TR::InstOpCode::
    TR::Node     *secondChild = node->getSecondChild();
 
 #ifdef J9_PROJECT_SPECIFIC
-   if (cg->profiledPointersRequireRelocation() && node->isProfiledGuard() && secondChild->getOpCodeValue() == TR::aconst &&
-       (secondChild->isClassPointerConstant() || secondChild->isMethodPointerConstant()))
+if (cg->profiledPointersRequireRelocation() && secondChild->getOpCodeValue() == TR::aconst &&
+   (secondChild->isClassPointerConstant() || secondChild->isMethodPointerConstant()))
+   {
+   if (node->isProfiledGuard())
       {
       TR_VirtualGuard *virtualGuard = comp->findVirtualGuardInfo(node);
       TR_AOTGuardSite *site = comp->addAOTNOPSite();
@@ -1252,6 +1254,12 @@ TR::Register *OMR::Power::TreeEvaluator::compareIntsForEquality(TR::InstOpCode::
       site->setNode(node);
       site->setAconstNode(secondChild);
       }
+   else
+      {
+      TR_ASSERT(!(node->isNopableInlineGuard()),"Should not evaluate class or method pointer constants underneath NOPable guards as they are runtime assumptions handled by virtualGuardHelper");
+      cg->evaluate(secondChild);
+      }
+   }
 #endif
 
    if (firstChild->getOpCodeValue() == TR::b2i &&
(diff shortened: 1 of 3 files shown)


compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1242,15 +1242,22 @@ TR::Register *OMR::Power::TreeEvaluator::compareIntsForEquality(TR::InstOpCode::
...
        (secondChild->isClassPointerConstant() || secondChild->isMethodPointerConstant()))
       {
-      TR_VirtualGuard *virtualGuard = comp->findVirtualGuardInfo(node);
-      TR_AOTGuardSite *site = comp->addAOTNOPSite();
-      site->setType(TR_ProfiledGuard);
-      site->setGuard(virtualGuard);
-      site->setNode(node);
-      site->setAconstNode(secondChild);
+		if (node->isProfiledGuard()
+			{
```
andrewcraik: The indenting here is messed up - please fix.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1625,15 +1632,22 @@ TR::Register *OMR::Power::TreeEvaluator::compareLongsForEquality(TR::InstOpCode:
...
          {
-         TR_VirtualGuard *virtualGuard = comp->findVirtualGuardInfo(node);
-         TR_AOTGuardSite *site = comp->addAOTNOPSite();
-         site->setType(TR_ProfiledGuard);
-         site->setGuard(virtualGuard);
-         site->setNode(node);
-         site->setAconstNode(secondChild);
+			if (node->isProfiledGuard())
+				{
+         	TR_VirtualGuard *virtualGuard = comp->findVirtualGuardInfo(node);
```
andrewcraik: indenting again

andrewcraik:
x change LGTM

fjeremic:
LGTM. Address constants related to methods or classes need to be AOT relocated irregardless of whether the node has a profiling guard or not. There are some tab vs. spaces issues in your change. Also please reword the commit header to be at most 50 characters in length so as to conform to the commit contributing guidelines:

0dvictor:
LGTM - changes on X86.

ymanton:
Your commit message should probably refer to relocatable code rather than AOT, and there's a spelling mistake of 'contant' in there. Aside from the formatting, the change itself looks OK to me.

vijaysun-omr:
If we have a class pointer constant (or a method pointer constant) under a NOPable guard, why does it need to be relocated ?

r30shah:
Please correct me if I am wrong, I think it is checking that the comparison we are generating here is for to guard some code like in-lining of method or so and as it needs to be relocatable in AOT as constant against which are comparing can change in different JVM invocations. 
My change is to create a AOT relocations when we generate sequence to compare address and address constant child needs to be AOT relocatable.

andrewcraik:
@r30shah the point is that you do not need to evaluate the aconst of a nop'd guard - the constant will be part of the runtime assumption and handled by the runtime assumption relocation logic. Basically, rather than testing for a profiled guard you want to test that the if is not a nopable guard to ensure you don't consume a register evaluating the constant needlessly. A good catch by Vijay!

r30shah:
The query sent to check if needs relocation is this one  "cg->profiledPointersRequireRelocation()" which at the end sends isAOT query so will be evaluating a child only if it is aconst and it is Class or Method Pointer Constant in AOT compilations. 

Also to handle the VGNOP case, All the platform calls virtualGuardHelper which generates VGNOP instruction for eligible cases and skips rest of the evaluation.
```
TR::Register *
OMR::Z::TreeEvaluator::ificmpeqEvaluator(TR::Node * node, TR::CodeGenerator * cg)
   {
   PRINT_ME("ificmpeq", node, cg);

   if (virtualGuardHelper(node, cg))
      {
      return NULL;
      }
```

andrewcraik:
Ok so if there is an assumption that a nop'able guard won't have its aconst evaluated there should be an assert to that affect in the code just to be sure we aren't causing a problem. If the change still works with that assert we should be good to merge.

vijaysun-omr:
Please add the assert suggested by Andrew and I can review afterwards.

r30shah:
I have added assertions so now if node is NOPABLE and in AOT compilation we try to evaluate, it is going to generate asserts.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1252,6 +1254,12 @@ TR::Register *OMR::Power::TreeEvaluator::compareIntsForEquality(TR::InstOpCode::
       site->setNode(node);
       site->setAconstNode(secondChild);
       }
+   else
+      {
+      TR_ASSERT(!(node->isNopableInlineGuard()),"We should not reach this case with nopable guard node");
```
fjeremic: I suggest we use a more explicit assert. The message should describe _why_ we shouldn't reach this case with a nopable guard. At the moment it provides no feedback to the user should the assertion trigger. Perhaps saying:

```
TR_ASSERT(!(node->isNopableInlineGuard()), "NOPable guards register runtime assumptions and do not need relocation records, hence class or method pointer constants underneath NOPable guards should not be evaluated");
```

Similarly in other places in this commit.

r30shah: Makes sense. Made the chages. Thanks for the suggestions.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1252,6 +1254,12 @@ TR::Register *OMR::Power::TreeEvaluator::compareIntsForEquality(TR::InstOpCode::
       site->setNode(node);
       site->setAconstNode(secondChild);
       }
+   else
+      {
+      TR_ASSERT(!(node->isNopableInlineGuard()),"Should not evaluate class or method pointer constants underneath NOPable guards as NOPable guard registers are runtime assumptions and do not need relocation records");
```
andrewcraik: NOPable guards do not have registers. Perhaps "Should not evaluate class or method pointer constants under NOPable guards - relocations are handled by runtime assumption relocation logic" or similar.

r30shah: As we handle the NOPable node using virtual guard helper, how about,
"Should not evaluate class or method pointer constants underneath NOPable guards as they are runtime assumptions handled by virtualGuardHelper");

?

andrewcraik: I am fine we your proposed wording.

r30shah: Made the changes. Thanks.

andrewcraik:
LGTM

vijaysun-omr:
Thanks for the changes. I will merge in once checks pass.

PR #973, opened 2017-04-04
Rename JProfiling options to SamplingJProfiling

andrewcraik:
JProfiling is a next generation technology for generating execution
profiling information from compiled methods. The infrastructure
currently in OMR reflects a hardware sampling based approach. We have
had greater success with a more deterministic apporach. This change
renames the current JProfiling infrastructure to reflect its sampling
basis so we can add a deterministic version along side it. At present we
think the sampling approach is likely to be removed, but aren't ready to
that yet so we just want to avoid name collisions and confusion.


vijaysun-omr:
Simple renaming of the older implementation of JProfiling related options. Merging.

PR #975, opened 2017-04-04
Fix block split assert

andrewcraik:
There is an assert in the edge splitting code which asserts a block
connected to the exit must end in a throw or a return. It checks this
by looking at the treetop. These nodes are allowed to appear under check
treetops and this change fixes the assert to correctly handle these
cases.

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1383,8 +1383,8 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
    else
       {
       TR::TreeTop *lastTree = from->getLastRealTreeTop();
-      TR::ILOpCodes opCode = lastTree->getNode()->getOpCodeValue();
-      TR_ASSERT(opCode == TR::athrow || lastTree->getNode()->getOpCode().isReturn(), "edge to EXIT should end in throw or return");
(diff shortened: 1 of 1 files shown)


0xdaryl:
Minor nit: please fix spelling of 'exits' in the first sentence of your commit message.

andrewcraik:
Fixed

compiler/il/OMRBlock.cpp
```diff
@@ -1383,8 +1383,8 @@ OMR::Block::splitEdge(TR::Block *from, TR::Block *to, TR::Compilation *c, TR::Tr
    else
       {
       TR::TreeTop *lastTree = from->getLastRealTreeTop();
-      TR::ILOpCodes opCode = lastTree->getNode()->getOpCodeValue();
-      TR_ASSERT(opCode == TR::athrow || lastTree->getNode()->getOpCode().isReturn(), "edge to EXIT should end in throw or return");
+      TR::ILOpCode &opCode = lastTree->getNode()->getOpCode().isCheck() ? lastTree->getNode()->getFirstChild()->getOpCode() : lastTree->getNode()->getOpCode();
```
vijaysun-omr: Should we also check for TR_treetop opcode here ? I know that throw/return can appear at the top of the tree but some optimization could in theory change a null check to a treetop for example and wondering if it might be better to future proof this check.

andrewcraik:
I have update the assert to handle treetops as well as checks. @vijaysun-omr back to you.

PR #976, opened 2017-04-04
Remove deprecated packed object support

0xdaryl:
Remove the remnants of a prototype feature implemented in J9 Java to support
a packed object representation.  The feature is no longer supported and should
not be in the Eclipse OMR code base.


vijaysun-omr:
Removing only code that was used under the deprecated PackedObject feature. Merging.

PR #977, opened 2017-04-04
Add infra to save basic block offset and frequency

andrewcraik:
Reassociating profiling information can be very tricky. In general the
OMR compiler will use bytecode metadata stored alongside profiling data
to faciliate re-associating the profiling data with basic blocks. It is
general practice in the compiler to reset all block frequencies post
inlining to regenerate a re-nomalized set of frequency data for the
composed method as a whole. This means we can see blocks which have been
split by inlining and we need to find the correct bytecode index at
which to lookup their frequency data. It is also possible that we are
not able to recompute a frequency for a block even though a frequency
was computed for it when it was IL generated in isolation. For this
reason it would be desirable to save frequency information from il
generation as a fallback. This change adds the infrastructure to save
bytecode ranges for basic blocks as seen by IL generation and their
associated frequencies for later use during profile re-association.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -131,7 +131,8 @@ OMR::ResolvedMethodSymbol::ResolvedMethodSymbol(TR_ResolvedMethod * method, TR::
      _pythonConstsSymRef(NULL),
      _pythonNumLocalVars(0),
      _pythonLocalVarSymRefs(NULL),
-     _properties(0)
+     _properties(0),
+     _bytecodeProfilingOffsets(comp->allocator())
    {
    _flags.setValue(KindMask, IsResolvedMethod);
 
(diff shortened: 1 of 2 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -2273,6 +2274,107 @@ OMR::ResolvedMethodSymbol::setNoTemps(bool b)
...
+   {
+   for (auto itr = _bytecodeProfilingOffsets.begin(), end = _bytecodeProfilingOffsets.end(); itr != end; ++itr)
+      {
+      if (itr->first >= bytecodeIndex)
+         itr->second.second = frequency;
+      }
+   }
+
+/**
+ * \brief Return the saved profilign frequency for the given bytecode index
```
vijaysun-omr: minor type "profilign"

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -2273,6 +2274,107 @@ OMR::ResolvedMethodSymbol::setNoTemps(bool b)
...
+ * @param startBCI The bytecode index of the start of the basic block
+ * @param endBCI The bytecode index of the end of the basic block
+ */
+void OMR::ResolvedMethodSymbol::addProfilingOffsetInfo(int32_t startBCI, int32_t endBCI)
+   {
+   for (auto itr = _bytecodeProfilingOffsets.begin(), end = _bytecodeProfilingOffsets.end(); itr != end; ++itr)
+       {
+       if (itr->first >= endBCI)
+          {
+          _bytecodeProfilingOffsets.insert(itr, std::make_pair(endBCI, std::make_pair(startBCI, -1)));
```
vijaysun-omr: Could you please clarify what the -1 value for second.second represents here ? It seems to be set in setProfilerFrequency but because the usage of the api is not included in this commit (perfectly fine that it is'nt) I cannot see who sets that field value under what circumstance.

Possible to state up front what first, second.first and second.second denote at a high level in the summary for the PR ?

andrewcraik: so the -1 value is the stashed profiling frequency associated with the bytecode block that begin at startBCI and ends at endBCI. This is the default 'safe' value and the profiling frequency is stashed at a point much later in the compile, after the blocks have been created and saved.

andrewcraik: The first and second uses map directly to the std::pairs in the list. so first on its own is the endBCI, second is the second pair which consists of startBCI and profiling frequency (second.first and second.second). The naming isn't great due to the use of std::pair, but it save creating yet another data type and gets the benefits of all things std::pair. Once you are used to std::pair I feel the code reads pretty well - it great, but it is understandable IMO.

andrewcraik:
Comment updated - there were a couple of pushes because of some confusion in my local repo about the correct base, but the PR delta is once again correct.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
vijaysun-omr: Thanks. So there is a single entry for every block that existed in the original program and as new blocks are created as a result of splitting those original blocks, new entries would get added to the list for the startBCI/endBCI corresponding to the split point ? I guess I am wondering if this addition of new entries is typically skipped in some case(s) because the original block's frequency can still be found since it is a sorted list and so you will find "some" entry based on the endPC test you have implemented.

andrewcraik: New blocks are not added to the list. Once we finish ILGen the set of blocks if fixed. We map new blocks to old blocks using the bytecode index. Given the lack of updates to the list I don't think there is an issue. This is to preserve the ILGen state of blocks so we can map bytecodes back to it at a later point in a compile.

vijaysun-omr: Thanks. That clarifies the picture.

vijaysun-omr:
Checks had passed earlier and the only change since then is to a comment. So I am merging this now.

Issue #897, opened 2017-03-21
Migrate JitBuilder examples to use the new testing framework


Leonardo2718:
Now that #881 and #892 have been merged, JitBuilder examples that are better suited as test cases should be migrated to use the new testing framework.

mstoodle:
note that those examples also serve as "samples" for how to use JitBuilder. Can you provide a list of those ones you're thinking to migrate (and do you really mean "migrate" or do you mean also implement as a test) ?

PR #979, opened 2017-04-05
Add some JitBuilder tests based on existing examples

Leonardo2718:
This is some initial work to address issue #897.

In this change, the `ControlFlowTest` is removed from the examples and replaced by an equivalent test case. A test equivalent to the `WorklistTest` example is also added but the example is not removed as there is no other example that demonstrates the use of JitBuilder worklists.


fvtest/jitbuildertest/WorklistTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+#include "JBTestUtil.hpp"
+
+#include "ilgen/BytecodeBuilder.hpp"
+#include "ilgen/VirtualMachineState.hpp"
+
+#define EXPECT_BICEQ(expected_index, actual_index) do {\
```
mstoodle: why BIC ? I was expecting BCI, but maybe I'm misreading your intent?

fvtest/jitbuildertest/WorklistTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+   EXPECT_EQ((expected_index), (actual_index)) << "Unexpected bytecode index (likely implies bytecode was not found in worklist)."; \
```
mstoodle: I think the indices not being equal implies the worklist either missed or added a bytecode it should not have. "not found in worklist" implies some kind of search which I think is misleading in this use case.

fvtest/jitbuildertest/WorklistTest.cpp
Leonardo2718: :stuck_out_tongue: Definitely should be BCI. Fixing.

fvtest/jitbuildertest/WorklistTest.cpp
Leonardo2718: Good point. Fixed.

Leonardo2718:
@mstoodle Have you had a chance to look over my changes?

fvtest/jitbuildertest/WorklistTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+#include "JBTestUtil.hpp"
+
+#include "ilgen/BytecodeBuilder.hpp"
+#include "ilgen/VirtualMachineState.hpp"
+
+#define EXPECT_BCIEQ(expected_index, actual_index) do {\
+   EXPECT_EQ((expected_index), (actual_index)) << "Unexpected bytecode index (implies bytecode was either missed or erroniously added in worklist)."; \
```
mstoodle: typo: should be erroneous

mstoodle:
sorry for the late review, @Leonardo2718 . One typo and I'll merge. Please feel free to poke me faster if I don't respond, but I'll try to pay more attention this time :) .

fvtest/jitbuildertest/WorklistTest.cpp
Leonardo2718: Fixed.

PR #980, opened 2017-04-05
DirectJNI is not supported in fsd mode

cathyzhyi:
Only enable direct JNI for ARM for non fsd mode

### DIFF
--- a/compiler/arm/codegen/OMRCodeGenerator.cpp
+++ b/compiler/arm/codegen/OMRCodeGenerator.cpp
@@ -90,8 +90,8 @@ OMR::ARM::CodeGenerator::CodeGenerator()
    _linkageProperties = &self()->getLinkage()->getProperties();
    _linkageProperties->setEndianness(TR::Compiler->target.cpu.isBigEndian());
 
-
-   self()->setSupportsDirectJNICalls();
+   if (!self()->comp()->getOption(TR_FullSpeedDebug))
+      self()->setSupportsDirectJNICalls();
    self()->setSupportsVirtualGuardNOPing();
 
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
To be clear, is this bringing ARM in line with what is already the case on our other platforms ?

cathyzhyi:
that's correct.

andrewcraik:
Yes that is correct - the other platforms are handled in a downstream project, but the ARM codegen is not yet factored that way so we are making the equivalent change here.

vijaysun-omr:
FYI @JBKingdon @0xdaryl and @andrewcraik

andrewcraik:
LGTM from me.

Leonardo2718:
The change LGTM.

One question for completeness, how was this tested (since the Travis builds do not currently test ARM)?

JamesKingdon:
Interesting, directToJNI is one of two features I had to disable last night to get further in my testing.

andrewcraik:
@Leonardo2718 The ARM change was not directly run. We do know from talking to the VM guys that the sequence used for direct JNI is not supported under FSD and testing on the other code generators showed that to be the case. directJNI is an optimization to increase performance so turning it off can leave perf on the table, but will not be functionally incorrect. The change is to keep inline with the platforms directly tested - we are making the structurally equivalent change.

@JamesKingdon it is possible that recent changes in the code to handle this in the downstream project consuming this codegen could have introduced bugs leading to directJNI failing to operate correctly in all cases.

Leonardo2718:
@andrewcraik thank you for the explanation.

PR #983, opened 2017-04-06
Ensure virtual calls for inlined calls have PP

ncough:
Virtual calls for inlined calls are moved to the taken side of
their guards. In OSR, however, it is necessary to also ensure
the necessary pending push temps are present. This change
ensures the call result is stored into the necessary pending
push temp slot, as this is the only change to stack that should
occur. Other pending push temps should be located before the
guard.


andrewcraik:
Only request is could we please have some inliner trace in the treetop skipping to document when we create the pending push store in block 4 and when we skip treetops.

charliegracie:
@ncough The failures in the travis build appear to be real compile failures and not any of the known intermittent failures.

andrewcraik:
@charliegracie I think this is because this change depends on another PR that is waiting to go in - should not be merged until that is sorted and a new build triggered. @ncough force push once the dependent change is merged.

vijaysun-omr:
Looks okay to me. Merging.

PR #984, opened 2017-04-06
Prevent gotoElim from skipping asynccheck in OSR

ncough:
Goto elimination will attempt to remove blocks that contain
a series of asyncchecks followed by a goto. To ensure the asynccheck
is still present, it is currently moved to all of its predecessors.
However, these may have other successors, some of which may not
correctly manage the addition of a yield point. To correct this,
the asynccheck is shifted to the start of the goto's destination block
which should contain anything needed to account for the yield.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1992,9 +1992,6 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
       // TM results in difficult situations when determining where the transition should occur,
       // so it is currently disabled. The implications of disabling TM should be investigated further.
       self()->setOption(TR_DisableTM);
-      // RedundantGotoElimination creates complications for NextGenHCR, potentially due to
-      // the manipulation of asyncchecks. The implications of disabling it should be investigated further.
-      self()->setDisabled(redundantGotoElimination, true);
       // Asynchecks cannot be removed in NextGenHCR as they may be used for transitions
       self()->setDisabled(redundantAsyncCheckRemoval, true);
       }
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -3276,9 +3276,8 @@ void TR_EliminateRedundantGotos::renumberExitEdges(TR_RegionStructure *region, i
       }
    }
 
-void TR_EliminateRedundantGotos::placeAsyncCheck(TR::Block *block)
+void TR_EliminateRedundantGotos::placeAsyncCheckBefore(TR::TreeTop *tt)
    {
-   TR::TreeTop *tt   = block->getLastRealTreeTop();
    TR::TreeTop *actt =
       TR::TreeTop::create(comp(), TR::Node::createWithSymRef(tt->getNode(), TR::asynccheck, 0,
                                      getSymRefTab()->findOrCreateAsyncCheckSymbolRef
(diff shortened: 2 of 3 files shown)


andrewcraik:
The bigger issue here is that the shift to the predecessor could place the asynccheck above a branch allowing a yield to occur where we may not have accounted for it. Moving to the successor prevents this problem... If you agree with that perhaps you can clarify the PR comment to reflect this.

ncough:
PR comment has been updated to correctly reflect the issue being fixed.

andrewcraik:
LGTM

vijaysun-omr:
Could moving the async check to the start of the successor also cause "a yield to occur where we may not have accounted for it" in some case ? e.g. if the successor was the header of a loop that the blocks are part of.

ncough:
In the event that the asynccheck is a yield point that must be accounted for, the additional required trees would have already been generated for it and may be located in the block with it. In this case, the optimization will not be applied as there will be other treetops in the block. They will otherwise be in the destination block, where it is correct to place the asynccheck before them.

Otherwise, if there is no invalidation point to account for at this yield point, there must be either no invalidation points remaining or they must be managed by another yield. If there are no invalidation points remaining, it is safe to shift the yield point forward as there is nothing to account for. If the invalidation point is managed by another yield, this other yield and its required trees must be located between the start of the destination block and the invalidation point. Therefore, it is safe to shift the yield to the start of the destination, as the invalidation will be accounted for by the later yield.

vijaysun-omr:
Thanks. So the reason why downward motion of the async check is valid is because there is only a goto at the end of the block (and not some other additional required tree pattern that could be an OSR guard). Moving into the predecessor on the other hand could move the async check past an OSR guard which is what makes that operation incorrect.

vijaysun-omr:
I thought about this some more and I still have a question around the simple case when we have a loop consisting of only 2 blocks 

1) a loop entry that ends in the tree pattern that is an OSR guard
2) successor of block 1 that has only an async check and a goto back to block 1

In this case moving the async check back past the predecessor edge should result in the same code as moving the async check forward past the successor (goto) edge. i.e. it seems we have done the same transformation (that we were trying to avoid) just by different means. Could you please clarify if this is a potential problem ?

ncough:
Moving the asynccheck back past an OSR guard is not an incorrect operation, as there will still be an OSR guard between the yield point and any invalidation points. The original issue this patch attempts to address is when the asynccheck is moved back past any other branch, as the predecessor's other successor may not have an OSR guard before an invalidation point.

vijaysun-omr:
Okay, then let's modify the example to end block 1 in some other branch (i.e. not an OSR guard). Is that still a potential problem ?

ncough:
This should not be a potential problem. If there is an invalidation point inside of block 1, there should already be an OSR guard between it and the asynccheck, so it is safe to move the asynccheck to the top of block 1. This is also true if the invalidation point is instead in the branch's other successor, as an OSR guard must have existed in the successor or, similar to the prior situation, in block 1. This last situation is where moving the asynccheck back would be invalid, as it would sit between the guard and the invalidation point.

vijaysun-omr:
Okay so the key difference comes from the fact that we move to the "top" of the successor. This avoids the problem with moving to the "bottom" of the predecessor, i.e. the async check could be executed before an OSR guard but after an invalidation.

ncough:
Yes, that is the issue addressed here.

vijaysun-omr:
Thanks I will merge now.

PR #985, opened 2017-04-06
Add check for OSR induce block

ncough:
When OSR transitions are implemented using OSR guards, the taken
side of the guard manages several book keeping aspects of the
transition. For example, it will contain stores to dead symbol
references and will remat symbol references required for the
transition. This block can be ignored by several optimizations.
To allow for this, this change adds a check for such a block.

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1708,6 +1708,46 @@ OMR::Block::getGlobalNormalizedFrequency(TR::CFG * cfg)
    return frequency;
    }
 
+/*
+ * An OSR induce block is used in voluntary OSR to contain the OSR point's dead stores
+ * and to induce the transition. It will always have the exit and at least one exception
+ * successor to an OSR catch block.
+ */
+bool
+OMR::Block::isOSRInduceBlock(TR::Compilation *comp)
+   {
+   // An OSR induce block should have one successor, the end, and will not exist in involuntary OSR
+   if (comp->getOSRMode() != TR::voluntaryOSR || self()->getSuccessors().size() != 1 ||
+       self()->getSuccessors().front()->getTo() != comp->getFlowGraph()->getEnd())
+      return false;
+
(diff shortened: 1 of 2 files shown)


andrewcraik:
This is only valid in voluntary OSR mode right - in involuntary normal blocks have the exception edges and the test will be misleading...

andrewcraik:
Maybe under debug and prod with assumes we should check the block to see if the call is actually there?

compiler/il/OMRBlock.cpp
```diff
@@ -1708,6 +1708,19 @@ OMR::Block::getGlobalNormalizedFrequency(TR::CFG * cfg)
...
+/*
+ * An OSR induce block is used in postExecutionOSR to contain the OSR point's dead stores
+ * and to induce the transition. It will always have no successors and one exception
+ * successor to an OSR catch block.
+ */
+bool
+OMR::Block::isOSRInduceBlock(TR::Compilation *comp)
+   {
+   return comp->getOSRMode() == TR::voluntaryOSR
+      && self()->getExceptionSuccessors().size() == 1
```
vijaysun-omr: Do we really need the test that there must only be a single exception successor ? 
Is it possible that the induce block has a non OSR catch block as an exception successor in addition to an OSR catch block ?

ncough: As this function could be called on any arbitrary block, it is necessary to check the number of exception successors to safely call front().

andrewcraik: right, I think Vijay's point is that the block could potentially have other exception successors (ie real ones) and we should account for that. Perhaps this could read more like:
`if (comp->getOSRMode() != TR::voluntaryOSR)
   return false;
if (self()->getExcetionSuccessors().size() == 1
   && ,,,(
   return true;
// loop to check for an OSRCatch exception successor`

vijaysun-omr:
Thanks Nick and Andrew for updating the commit. Looks much more correct now. I will merge once checks pass.

PR #986, opened 2017-04-06
Ignore OSR induce blocks in GRA

ncough:
OSR induce blocks will contain loads and stores of symbol references
to remat those that are live at the transition. As these blocks will
typically never be visited and, in the event that they are, performance
is expected to be bad, they should be ignored when determining the
weight for register candidates. This allows other symbol references,
used in the main line execution, to have registers allocated to them
and reduces spilling.

### DIFF
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -1320,12 +1320,11 @@ TR_RegisterCandidate::processLiveOnEntryBlocks(TR::Block * * blocks, int32_t *bl
          blockWeight = blockStructureWeight[bnum];
          }
 
-      if ((!dontAssignInColdBlocks(comp) || !block->isCold()) &&
-          (blockWeight >= maxFrequency*freqRatio))
+      bool ignoreBlock = (dontAssignInColdBlocks(comp) && block->isCold()) || block->isOSRInduceBlock(comp);
+      if (!ignoreBlock && (blockWeight >= maxFrequency*freqRatio))
          ++origNumberOfBlocks;
 
-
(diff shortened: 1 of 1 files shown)


andrewcraik:
I'm not an expert on GRA, but this looks fairly safe and simple - OSR transition blocks should not factor into GRA's consideration since they are so rare. LGTM @vijaysun-omr

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -955,6 +955,14 @@ TR_RegisterCandidate::processLiveOnEntryBlocks(TR::Block * * blocks, int32_t *bl
          if (trace)
             traceMsg(comp, "\tExamining live-on-entry block_%d with loads and stores = %d\n", blockNumber, (*_loadsAndStores)[blockNumber]);
 
+         // Ignore uses in OSR induce blocks
+         if (blocks[blockNumber]->isOSRInduceBlock(comp))
+            {
+            _liveOnEntry.reset(blockNumber);
```
vijaysun-omr: Do we need to reset the live on entry for the induce OSR block ? 

I understand the reasoning behind the "continue" statement skipping the weighing process in this change, but resetting live on entry for the induce OSR block could lead to stores being emitted on the main line path before the OSR guard which is undesirable.

vijaysun-omr:
Thanks for making the requested change. Did you feel this required a performance evaluation ?

ncough:
Yes, performance was evaluated with the requested change and the original PR. Both versions demonstrated the throughput improvements expected when ignoring OSR induce blocks. Reduced spilling was also confirmed in common culprits of bad register allocation.

vijaysun-omr:
Thanks, I will merge this.

PR #989, opened 2017-04-07
Add initial infrastructure for new JProfiling

andrewcraik:
Now that the old JProfiling implementation based on hardware sampling
has been renamed to SamplingJProfiling, I would like to add initial
placeholder infrastructure to support the new deterministic JProfiling
that is in development in a downstream project. The plan is to
contribute this new implementation, but I need to do additional
performance verification and tuning first as well as some refactoring.
This change adds minimal placeholders to make it easier to contribute
the new implementation in the (hopefully near) future.


charliegracie:
What is JProfiling?

andrewcraik:
JProfiling is runtime profiling of OMR compiler generated code. There is currently a JProfiling implementation in OMR called SamplingJProfiling and it uses hardware sampling to try and build an execution profile. The implementation in OMR of SamplingJProfiling is a prototype and we have not had great success when evaluating it in our downstream project. The new JProfiling I am working to contribute is a very low overhead method of placing software counters to derive execution profiles from compiled code. The main use is to allow methods to be recompiled post-profiling to boost performance since the compiler optimizations are highly profile dependent.

compiler/optimizer/Inliner.cpp
```diff
@@ -2617,7 +2617,8 @@ TR_TransformInlinedFunction::transform()
    // If the first block has exception predecessors or multiply predecessors then we can't merge
    // the first block with the caller's block
    //
-   if ((firstBlock->getPredecessors().size() > 1) ||
+   if (comp()->isJProfilingCompilation() ||
+       (firstBlock->getPredecessors().size() > 1) ||
```
vijaysun-omr: Could you please add a brief comment here on the reason why this logic needs to be guarded by the extra test you added ?

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -1943,7 +1943,8 @@ void OMR::X86::CodeGenerator::doBinaryEncoding()
 
 #ifdef J9_PROJECT_SPECIFIC
    if ((!self()->comp()->getOptions()->getOption(TR_DisableGuardedCountingRecompilations) && TR::Options::getCmdLineOptions()->allowRecompilation()) ||
-       (self()->comp()->getOptions()->getEnableGPU(TR_EnableGPU) && self()->comp()->hasIntStreamForEach()))
+       (self()->comp()->getOptions()->getEnableGPU(TR_EnableGPU) && self()->comp()->hasIntStreamForEach()) ||
+       self()->comp()->isJProfilingCompilation())
```
vijaysun-omr: Same here. A brief comment on why this logic is guarded.

compiler/optimizer/Inliner.cpp
andrewcraik: This code controls whether we will merge the method entry block of an inlined method with the caller's block. When running profiling, this causes confusion when profiling owing to the fact we cannot see the entry block of the inlined method. This change forces us to split the block which fixes this confusion quite neatly.

compiler/x/codegen/OMRCodeGenerator.cpp
andrewcraik: This change is to record the startPC of the compiled method so that when JProfiling triggers recompilation we can locate the method metadata to perform the recompilation. With out this the recompilation logic crashes unable to find the method that is trying to be recompiled.

vijaysun-omr:
Thanks for the explanations. They help clarify the reasoning. Merging.

Issue #469, opened 2016-11-11
numa.h presence isn't detected and warned during configure


mgaudet:
One thing a number of people have encountered is `./configure`ing OMR on a new machine, then hitting make, only to have it not succeed when `numa.h` isn't found (latest example [here](https://github.com/rubyomr-preview/rubyomr-preview/issues/19)). We should be detecting it in configure (and if possible, disabling).

youngar:
Resolved by #991

PR #991, opened 2017-04-07
Ensure numa.h is available if required

charliegracie:
Modify OMRCFG_DEFINE_FLAG to execute the <default action> for both
default yes and command line set options.

Modify the OMR_PORT_NUMA_SUPPORT configure define to check for numa.h if
it is --enabled and print an error message if it is not available.

Only set --enable-OMR_PORT_NUMA_SUPPORT for one of the Linux x86
platforms. We do not need to test this in all specs.

Update INSTALL to remove the requirement for numa-dev


mgaudet:
So this will resolve https://github.com/eclipse/omr/issues/469 correct?

charliegracie:
Yes it will resolve #469.

PR #993, opened 2017-04-10
Code cleanup related to X86 Binary Encoding

0dvictor:
Remove code deprecated by new X86 Binary Encoding Scheme.


andrewcraik:
looks like another good piece of clean-up following on from previous work. Transformations look fairly mechanical and seem correct.

PR #994, opened 2017-04-10
WIP: Fail ilgen if involuntary OSR cannot do OSR

cathyzhyi:
A few non-optional transformations in ilgen can result in OSR failures.
In such cases we need to fail the ilgen by throwing
TransformationUnsupportedUnderInvolentaryOSR excpetion to either abort
the compilation or stop inlininig the callee depending on whether it's
top most method or inlined callee.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1254,6 +1254,14 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
                comp->getOption(TR_EnableOSR) && comp->supportsInduceOSR() &&
                !comp->isPeekingMethod()  && (comp->getOption(TR_EnableNextGenHCR) || !comp->getOption(TR_EnableHCR)) ;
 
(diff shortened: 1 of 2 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1254,6 +1254,14 @@ OMR::ResolvedMethodSymbol::genIL(TR_FrontEnd * fe, TR::Compilation * comp, TR::S
                comp->getOption(TR_EnableOSR) && comp->supportsInduceOSR() &&
                !comp->isPeekingMethod()  && (comp->getOption(TR_EnableNextGenHCR) || !comp->getOption(TR_EnableHCR)) ;
 
+            if (!comp->isPeekingMethod())
+               {
+               if (comp->getOSRMode() == TR::involuntaryOSR && !doOSR)
```
andrewcraik: This catches a general case, but isn't there a more specific check - namely we must support OSR for every node which returns true to isPotentialOSRPoint and shouldn't that check be enforced during ilgen? I suspect genTreeTop or similar may be the place we want to do something. This is only going to pick-up on global incompatibility not individual sites...

PR #998, opened 2017-04-11
WIP: Enable folding for test of array class in VP

JamesKingdon:
Detect Iand operating on a fixed array class so that optimizations
can fold away runtime tests for array types. This extends the
existing functionality in VP for folding an isArray test based on
an object to the case where we don't have an object, but we do have
a Class in hand.


andrewcraik:
@JamesKingdon please talk to @liqunl - this belongs in a downstream project and I believe she is already working to deliver equivalent code.

liqunl:
Sorry I didn't check my inbox frequently. I am not working on something equivalent. All of my vp opts relate to folding a call, not a test.

PR #999, opened 2017-04-11
Add property to merge guards with OSR guards

ncough:
Existing guards may be reused to implement OSR transitions. These
guards should be marked appropriately so that they can be
patched when necessary, similar to OSR guards. This change adds
a property to signify this, similar to the existing mergedWithHCR flag.


andrewcraik:
LGTM

vijaysun-omr:
Looks good. I will merge once checks pass.

PR #1001, opened 2017-04-11
Add required files for Eclipse

charliegracie:

### DIFF
--- a/Notice.html
+++ b/Notice.html
@@ -1,0 +1,108 @@
+
+<?xml version="1.0" encoding="ISO-8859-1" ?>
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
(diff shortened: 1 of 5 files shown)


mstoodle:
Thanks @charliegracie for putting this together. I asked one very minor question from the Eclipse lawyers on one part of the wording. Once we hear back, either I'll merge or request a very small change.

mstoodle:
Ok, my question to Eclipse legal was around the use of the word "plug-in" which seemed an inappropriate description of our particular project. But I hadn't noticed all the places where the term was used, so changing it to something like "component" is a more complicated change than I had envisioned.

The Eclipse legal team is now aware of my concern, and will hopefully address it as part of a more general review of the Eclipse legal documentation guide. They're ok with this PR, and I don't see a strong reason to figure out the "plug-in" use, so I'm just going to merge it. We'll use follow-on commits to improve the state of these documents.

Thanks again, @charliegracie !

mstoodle:
Removing the WIP marker as Eclipse legal has communicated their approval with the current commit.

PR #1003, opened 2017-04-11
Add DAA sub-option and disable/enable trap option

NigelYiboYu:
This commit adds two sub-options for the packed decimal optimization
group: the `disableMarshallingIntrinsics` option and the
`disablePackedDecimalIntrinsics` option, the latter of which disables the
optimizer and codegen from emitting exception-triggering packed decimal
hardware instructions. With these new options, one can separately disable
the exception-triggering and non-exception-triggering packed decimal
instructions.

Also, add `disableTraps` and `enableTraps` options to control the generation of 
trap options such as compare-and-trap. Traps are now disabled
on zOS by default.


NigelYiboYu:
The purpose of splitting the old `disableIntrinsics` into two sub-options: `disableMarshallingInstrinsics` and `disablePackedDecimalIntrinsics`, the latter of which disables the optimizer and codegen from emitting exception-triggering hardware instructions. With these two sub-options, one can separately disable the exception-triggering and non-exception-triggering packed decimal instructions. 

This commit also fixes a DataAccessAccelerator options bug where neither `-Xrs` nor  `-Xjit:noResumableTrapHandler` disables the generation of packed decimal instructions. Now, `diablePackedDecimalIntrinsics` is set if one of the `-Xrs` and `noResumableTrapHandler` is set.

Also, add the `enableTraps` and `disableTraps` options for all platforms so that they can be used to enable/disable the generation of exception-triggering instructions.

mgaudet:
[That comment](https://github.com/eclipse/omr/pull/1003#issuecomment-293399932) would make a great addition to your commit message.

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -715,14 +715,24 @@ OMR::Z::CodeGenerator::CodeGenerator()
    self()->setSupportsBCDToDFPReduction();
    self()->setSupportsIntDFPConversions();
 
+   // traps are causing some problems on z/OS and not providing much if any perf benefit
```
fjeremic: Could you please provide a better description of the problems you encountered which lead to the decision to disable traps on z/OS? It will help a future reader understand _why_ traps are disabled.

NigelYiboYu: Updated with a more detailed description.

compiler/compile/Method.cpp
```diff
@@ -254,6 +284,14 @@ bool TR_ResolvedMethod::isDAAIntrinsicMethod()
    return false;
    }
 
+bool TR_ResolvedMethod::isDAAIntrinsicMethod()
+   {
+#ifdef J9_PROJECT_SPECIFIC
+   return isDAAMarshallingIntrinsicMethod() || isDAAPackedDecimalIntrinsicMethod();
+#endif
```
fjeremic: There should be an `#elif` here for the false path.

compiler/compile/ResolvedMethod.hpp
```diff
@@ -150,7 +150,12 @@ class TR_ResolvedMethod
 
    bool isDAAMethod();
```
fjeremic: Are these used anywhere anymore following this change?

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -715,14 +715,42 @@ OMR::Z::CodeGenerator::CodeGenerator()
...
+    * 2. on z/OS Management Facility, disabling traps gave performance boost because of fewer exceptions from the JCL.
+    * The path length on z/OS for a trap is longer than on other operating systems so the trade-off and
+    * penalty for using traps on z/OS is significantly worse than other platforms. This gives the potential
+    * for significant performance loss caused by traps.
+    *
+    * 3. Depending on sysadmin settings, every time a trap fires, the system may get a "system dump"
+    * message which shows a 0C7 abend and with lots of exceptions being thrown this clutters up the syslog.
+    *
+    * 4. Certain products can get in the way of the JIT signal handler so the JIT doesn't
+    * receive the 0C7 signal causing the JVM process to get killed
```
fjeremic: > signal causing the ~JVM~ **product** to get killed

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -715,14 +715,42 @@ OMR::Z::CodeGenerator::CodeGenerator()
...
    if (_processorInfo.supportsArch(TR_S390ProcessorInfo::TR_z10))
       {
       self()->setSupportsTranslateAndTestCharString();
 
       self()->setSupportsBCDToDFPReduction();
       self()->setSupportsIntDFPConversions();
 
-      if (!comp->getOptions()->getOption(TR_NoResumableTrapHandler) && TR::Compiler->vm.hasResumableTrapHandler(comp))
+      // change from checking the noResumableTrapHandler to the new disableTrap option to decouple disabling
+      // the handler which is also used by DAA with disabling trap instructions. However, keep the
```
fjeremic: This comment needs an update. There is no concept of "DAA" or "service" in OMR. I would in fact just remove the comment all together as the above explanation is more than enough.

compiler/z/codegen/OMRCodeGenerator.cpp
NigelYiboYu: Fixed.

compiler/z/codegen/OMRCodeGenerator.cpp
NigelYiboYu: This comment has been removed.

compiler/compile/ResolvedMethod.hpp
NigelYiboYu: This is no longer used. Deleting it.

compiler/compile/Method.cpp
NigelYiboYu: Added `#elif return false;` to all similar places.

fjeremic:
LGTM.

compiler/arm/codegen/OMRCodeGenerator.cpp
```diff
@@ -160,7 +160,7 @@ OMR::ARM::CodeGenerator::CodeGenerator()
       // wrapping-around.
```
0xdaryl: |

0xdaryl:
I'm worried that this one commit seems to be conflating changes that deserve commits in their own right (but could be merged under the same PR of course).  In particular, I see four commits:

1)  Add `TR_EnableTraps` and `TR_DisableTraps` options.  For this one, however, because there are two seemingly opposite options I think you owe more of an explanation as to what each option's semantics are and how they are expected to be used and co-exist.  Are they polar opposites?  If so, why have two options?  If not, how do you manage the settings of both these options so they don't conflict?

As an example, some of your code now looks like this:
```
   if (!self()->comp()->getOptions()->getOption(TR_DisableTraps))
```
Its not clear whether this is equivalent to:
```
   if (self()->comp()->getOptions()->getOption(TR_EnableTraps))
 ```
and why.

Furthermore, I don't see either `EnableTraps` or `DisableTraps` set on any platform other than zOS and the code in this PR.  Please explain what the semantics of this is.

2)  Replacing the use of the `TR_NoResumableTrapHandler` option with one of your new trap options on other platforms.  Please explain the equivalence between `!TR_DisableTraps`
and `!TR_NoResumableTrapHandler` as in my mind they mean different things.  Furthermore, what is to become of the `TR_NoResumableTrapHandler` option as you did not remove it in this PR?  Is it now obsolete in your mind?  Please explain.

3)  Add two sub-options for the packed decimal opt group

4)  Fix the DAA option bug

NigelYiboYu:
@0xdaryl 

This commit is indeed a z-only commit. No other platform is currently using these new options.

I've told Simon and he can better answer your questions about the use of `enableTraps`, `disableTraps` and what is to become of the existing `NoResumableTrapHandler` options.

I will split this single commit into multiple ones and merge them together under the same request.

sehirst:
- This disables traps on z/OS by default so we need a way to turn them back on in case people want them, which explains the enableTraps. It's a no-op on all other platforms.
- Xjit:noResumableTrapHandler was suppose to disable both DAA and traps but it wasn't disabling DAA. This MR fixes that. disableTraps was added so that traps could be turned off separately from DAA. disableTraps is obviously a no-op on z/OS. noResumableTrapHandler should set disableTraps as the way for it to set disable traps. It used to be checked directly and so that's why it has been replaced with disableTraps in various places.
- If both disableTraps and EnableTraps are used, I think in current releases I made it so that the disable overrides the enable.

NigelYiboYu:
Correction to Simon's comment:
`enableTraps` overrides `disableTraps`. This piece of code is not in the OMR project because it's not common code.

Addition to what Simon said.

`enableTraps` and `disableTraps` are opposite of each other; but they are set and used differently. Currently, `enableTraps` is only used once in the constructor of CodeGenerator class (not defined in the OMR project).

`disableTraps` is a sub-option of `noResumableTrapHandler` and it's true on zOS by default and false on all other platforms.  `disableTraps` is true if `noResumableTrapHandler` is true. Due to the options hierarchy and the fact that `disableTraps` option has different default values on different platforms, we need a way to turn on trap option just for zOS. This is why a separate `enableTraps` option is needed.

0xdaryl:
Thank you for partitioning the work into multiple commits.

Perhaps I'm being obtuse, but the `enableTraps`/`disableTraps` option duality does not make sense to me.  Especially given @nigelyiboyu's explanation that "`enableTraps` and `disableTraps` are opposite of each other; but they are set and used differently".  This statement worries me and I'm sure will cause confusion.

It sounds as if these are not actually polar opposites, but worthy of two separate options with clearly defined purposes (perhaps one specific to zOS).  I suggest you try and restructure this work around one option that enables/disables trap support (true polar opposites) and a secondary option (perhaps zOS specific) that influences the default setting of that option.  I think it will be far easier to explain and document that way.

Adding to the confusion, neither of these proposed options are set on some platforms, so what is the semantics of that?  Traps are neither disabled nor enabled, so what's the expected behaviour?

Also, @sehirst suggests that the `enableTraps` option is there to enable it "in case people want it".  Is this intended to be an externally supported option on your downstream project that your customers could enable?  If so, that needs a very clear description of what the option is and how to use it.

sehirst:
noResumableTrapHandler disables DAA and traps. We need a way to disable each individually, hence disableTraps.

On platforms where traps are disabled by default, we need a way to re-enable them for potential perf investigations by us or customers, hence enableTraps.

Neither option needs to be externalized outside of an APAR. They exist for diagnostic and RAS purposes.

They are indeed polar opposites. Using them both in the same command line would be no different than doing something like -Xjit:disableInlining,tryInline={...}.

I wouldn't characterize them as being "set and used differently" like Nigel wrote.

0xdaryl:
In the interests of moving this forward, at the very least for this PR to be acceptable to merge you need to find a way for your 'disableTraps' and 'enableTraps' options to share a single Option bit.  Otherwise, they represent four possible states rather than just two.  The 'enable' option should have a direct effect on the 'disable' option.

If this is not possible, to me it suggests that these options are not doing what you think they're doing and the design needs to be reconsidered.

In any case, all options should have a meaningful default on all platforms that are affected.

NigelYiboYu:
Update

-- The `enableTraps` and `disableTraps` are now sharing the same option bit.

PR #1004, opened 2017-04-11
Runtime concurrentScavenger flag

amicic:
Use the flag to decide in runtime whether some part of Scavenger code
runs or not, depending if CS is requested or not. 

This will allow running both plain Scavenger and Concurrent Scavenger
under same build (as long as compile time CONCURRENT_SCAVENGER flag is
enabled in the build).

### DIFF
--- a/example/glue/ScavengerRootScanner.hpp
+++ b/example/glue/ScavengerRootScanner.hpp
@@ -98,9 +98,7 @@ public:
 		}
 	}
 	
-#if !defined(OMR_GC_CONCURRENT_SCAVENGER)
 	void rescanThreadSlots(MM_EnvironmentStandard *env) { }
-#endif
 
 	void scanClearable(MM_EnvironmentBase *env)
 	{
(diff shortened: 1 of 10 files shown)


fjeremic:
Do we have plans to deprecate `OMR_GC_CONCURRENT_SCAVENGER` eventually?

amicic:
No, the compile time flag will not be removed. It will be selectively enabled on specific platforms. It also gives somewhat cleaner separation of most of CS code (useful for code understanding and debugging). The platforms that do have compile time flag on will be able to query if CS is run-time enabled or not via a J9 specific API (there is a whole other J9 specific change that comes along this one).

charliegracie:
@amicic would it be possible to create a helper on extensions similarly to isScavengerEnabled(), etc. to check for this instead of a macro? Also you will need to rebase the change and force push it again because it can not longer be automatically merged.  Thanks

fjeremic:
> @amicic would it be possible to create a helper on extensions similarly to isScavengerEnabled(), etc. to check for this instead of a macro?

To add to this it would be great if this could somehow also be exposed to the OMR compiler.

amicic:
@fjeremic While you are asking for a reasonable thing, all external GC helpers are language specific. I cannot do it until that whole section of code (well, language independent part) migrates to OMR.

amicic:
@charliegracie Will do in the next increment of changes, when I end up enabling the compile time flag for select platforms.

PR #1005, opened 2017-04-11
Add async check for exception handlers for fsd

cathyzhyi:
Catch event could make JIT yield to vm in involuntary OSR mode and
the inserted async check is used as an indication of an OSR point
for the start of the catch block. This change is to prevent the
syncchk from being removed.


andrewcraik:
LGTM

andrewcraik:
@vijaysun-omr given the OSR related changes you have been merging could you take a look at this one please?

vijaysun-omr:
Simple change to capture the fact that a catch block is an OSR point in involuntary OSR mode. Merging.

PR #1006, opened 2017-04-12
Remove the prohibition on operator new and delete

mgaudet:
This prohibition will be sunk into downstream projects where it makes
sense. While OMR code still needs to use the correct allocators for
memory management, this particular method of prohibition is too leaky
and broad to stand in the OMR project as a whole.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -213,11 +213,6 @@ OMR::Compilation::getNOPTranslateTable()
    }
 
 
-
-
-/* See comment below about operator new */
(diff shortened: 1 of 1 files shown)


0xdaryl:
I'd like @lmaisons to review please.

lmaisons:
Looks OK to me.

andrewcraik:
Anyway the linter could be enhanced to help detect what this was doing, but in a more robust way only in the OMR code?

mgaudet:
That's a great alternative we should puruse in the future. I just opened https://github.com/eclipse/omr/issues/1008

PR #1007, opened 2017-04-12
Specialized 'fe' option processing routines

mpirvu:
Provide specialized functions for option processing functions that need
to operate on JitConfig. The JitConfig is actually cached in a static
field in OMR::Options::_feBase, so it is readily available.
This change paves the way to splitting the option flags into OMR generic
ones and option flags that are only used by subprojects.

Issue: #1055

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1359,6 +1359,13 @@ OMR::Options::set32BitNumeric(char *option, void *base, TR::OptionTable *entry)
 
 
 char *
+OMR::Options::set32BitNumericInJitConfig(char *option, void *base, TR::OptionTable *entry)
+   {
+   return TR::Options::set32BitNumeric(option, _feBase, entry);
+   }
+
+
+char *
 OMR::Options::set64BitSignedNumeric(char *option, void *base, TR::OptionTable *entry)
    {
    int64_t sign = 1;
@@ -1458,12 +1465,17 @@ OMR::Options::setString(char *option, void *base, TR::OptionTable *entry)
    return dummy_string;
    }
 
+char *
+OMR::Options::setStringInJitConfig(char *option, void *base, TR::OptionTable *entry)
+   {
+   return TR::Options::setString(option, _feBase, entry);
+   }
 
 char *
 OMR::Options::setStringForPrivateBase(char *option, void *base, TR::OptionTable *entry)
    {
 #ifdef J9_PROJECT_SPECIFIC
-   base = TR_J9VMBase::getPrivateConfig(base);
+   base = TR_J9VMBase::getPrivateConfig(_feBase);
    return TR::Options::setString(option, base, entry);
 #else
    return 0;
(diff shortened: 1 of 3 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -1359,6 +1359,13 @@ OMR::Options::set32BitNumeric(char *option, void *base, TR::OptionTable *entry)
 
 
 char *
+OMR::Options::set32BitNumericInJitConfig(char *option, void *base, TR::OptionTable *entry)
```
0xdaryl: This is a new function.  What is the purpose of the `base` parameter because its unused here and in all implementations in this PR?  Please add a Doxygen-style comment (perhaps in the header file where the other comment is already).

compiler/control/OMROptions.cpp
```diff
@@ -1458,12 +1465,17 @@ OMR::Options::setString(char *option, void *base, TR::OptionTable *entry)
    return dummy_string;
    }
 
+char *
+OMR::Options::setStringInJitConfig(char *option, void *base, TR::OptionTable *entry)
```
0xdaryl: This is a new function.  What is the purpose of the `base` parameter because its unused here and in all implementations in this PR?  Please add a Doxygen-style comment (perhaps in the header file where the other comment is already).

compiler/control/OMROptions.hpp
```diff
@@ -2064,6 +2064,11 @@ class OMR_EXTENSIBLE Options
    //
    static char *set32BitNumeric(char *option, void *base, TR::OptionTable *entry);
 
+   // Scan the option for a numeric value and set 32bit word at offset "offset"
```
0xdaryl: Not clear about this comment.  `offset` isn't a parameter to this function.

compiler/control/OMROptions.hpp
```diff
@@ -2115,6 +2120,12 @@ class OMR_EXTENSIBLE Options
    //
    static char *setString(char *option, void *base, TR::OptionTable *entry);
 
+   // Scan the option for a string value and set pointer at offset "offset"
```
0xdaryl: Not clear about this comment.  `offset` isn't a parameter to this function.

mpirvu:
I have created an issue #1055 to group all related pull requests

compiler/control/OMROptions.cpp
mpirvu: I have added Doxygen-style comments. 'base' is an unused parameter. It needs to be present because set32BitNumericInJitConfig() is an option processing routine that has a fixed format.

compiler/control/OMROptions.cpp
mpirvu: I have added Doxygen-style comments. Same as above, 'base' is an-used parameter that needs to be present due to the fixed format of the option processing routines.

mpirvu:
@0xdaryl I have addressed the suggestions from the review. Please review this commit again.

PR #1009, opened 2017-04-12
Disable ref. array copy for concurrent scavenge

fjeremic:
With the distinction between reference arraycopy and primitive arraycopy
available we can now enable primitive arraycopy under concurrent
scavenge and disable reference arraycopy under concurrent scavenge.

A TODO is left to explain why this restriction is imposed. We plan on
addressing this restriction in a future PR.


fjeremic:

dsouzai:
LGTM

vijaysun-omr:
Good change to allow primitive array copies under concurrent scavenge. Merging.

PR #1010, opened 2017-04-12
Move remove duplicates to gen superset

twlogger:
The old removeDuplicates() algorithm has been replaced entirely with a new one which simply checks that only one structure with a given full name is printed in the superset and blob, significantly improving performance.

As a side-effect of this change, several missing structures in the generated superset have been fixed. In addition, minor tweaks to the DwarfScanner have fixed additional missing structures.

Some missing JVM structures (around ~200) caused by bugs in the AIX parser are also fixed.

Signed-off-by: Tony Wang <twlogger@gmail.com>


charliegracie:
@twlogger can you provide a bit more details in each of the commit messages?  Also make sure to limit the first line to 50 characters please.

Thanks

twlogger:
I'll squash the latest commit.

PR #1011, opened 2017-04-12
Fix accidental modification of BCI

ncough:
When requesting the ByteCodeInfo for an OSR point that is the
child of a treetop, it is necessary to get the child's ByteCodeInfo.
This is required as the treetop may have a different value, confusing
the OSR infrastructure which relies on this value to determine the
transition point.

The API's function to fetch this value would accidentally modify
the treetop's BCI to match that of its child in the described case.
This would limit profiling later on as it relies on these values
to remain consistent.


andrewcraik:
LGTM obvious bug fix

vijaysun-omr:
Must not have been an easy bug to track down. Thanks, and I will merge now.

PR #1012, opened 2017-04-12
Treat OSR Guards appropriately in codegen

ncough:
OSR guards are NOP patch sites, which are patched at a GC point
into an unconditional branch, much like HCR guards. This change
updates Z and Power codegen to treat them appropriately.


andrewcraik:
Looks good from a quick glance @ymanton @fjeremic you would be the experts on the code where these changes are being made - could you please review?

fjeremic:
LGTM.

ymanton:
LGTM.

PR #1014, opened 2017-04-13
WIP: Fixes for DDR

youngar:
I was trying to run ddrgen on my computer, and ran into a couple problems.  I think it's working on my computer now. I'll test it more tomorrow.
/cc @jason-hall @twlogger

### DIFF
--- a/example/glue/MarkingDelegate.cpp
+++ b/example/glue/MarkingDelegate.cpp
@@ -51,12 +51,15 @@ MM_MarkingDelegate::scanRoots(MM_EnvironmentBase *env)
 void
 MM_MarkingDelegate::masterCleanupAfterGC(MM_EnvironmentBase *env)
 {
+	OMRPORT_ACCESS_FROM_OMRVM(env->getOmrVM());
 	J9HashTableState state;
 	ObjectEntry *rEntry = NULL;
 	OMR_VM_Example *omrVM = (OMR_VM_Example *)env->getOmrVM()->_language_vm;
 	rEntry = (ObjectEntry *)hashTableStartDo(omrVM->objectTable, &state);
 	while (rEntry != NULL) {
 		if (!_markingScheme->isMarked(rEntry->objPtr)) {
+			omrmem_free_memory((void *)rEntry->name);
+			rEntry->name = NULL;
 			hashTableDoRemove(&state);
 		}
 		rEntry = (ObjectEntry *)hashTableNextDo(&state);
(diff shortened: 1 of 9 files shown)


youngar:
I just saw I'm fixing one of the same bugs as in #1010.

tools/ddrgen/src/scanners/dwarf/DwarfScanner.cpp
```diff
@@ -74,20 +74,25 @@ DwarfScanner::getBlacklist(Dwarf_Die die)
 
 	if (DW_DLV_ERROR == dwarf_hasattr(die, DW_AT_comp_dir, &hasAttr, &error)) {
 		ERRMSG("Checking for compilation directory attribute: %s\n", dwarf_errmsg(error));
-		goto Failed;
+		/* The DIE didn't have a compilationDirectory attribute so we can skip
```
jason-hall: dwarf_hasattr() only returns DW_DLV_ERROR in the case of an internal dwarf library error, not in the case of no attribute present. If the Die has no compilation directory, hasAttr will be false. The correct thing to do here is to check the value of hasAttr, as was in the code somehow removed by the mentioned 055c3bf.

youngar: I thought that might be the case. Thanks for the clarification Jason :smile: .

youngar:
@jason-hall I'm using Arch Linux, and libdwarf is not linked to libelf, so I had to add it to the link line. Is this normal?
```
Linux aryoung-tp 4.10.8-1-ARCH #1 SMP PREEMPT Fri Mar 31 16:50:19 CEST 2017 x86_64 GNU/Linux

ldd /usr/lib/libdwarf.so.1
ldd: warning: you do not have execution permission for `/usr/lib/libdwarf.so.1'
	linux-vdso.so.1 (0x00007ffec0b83000)
	libz.so.1 => /usr/lib/libz.so.1 (0x00007f92c6c0c000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f92c6868000)
	/usr/lib64/ld-linux-x86-64.so.2 (0x0000558c47381000)
```

jason-hall:
Typically, in my experience, one must compile libdwarf themselves. I haven't seen a machine that already has it. I guess its different for you.

youngar:
Ubuntu has a package `libdwarf-dev` which provides `/usr/lib/libdwarf.a`.  I'll see if I can get the ddrgen test running in the travis build.

Also, thank you for sending me those instructions for DDR today. I'm going to add them to this commit.

doc/ddr.md
```diff
@@ -0,0 +1,68 @@
+(c) Copyright IBM Corp. 2016
```
fjeremic: Probably needs a copyright update.

PR #1015, opened 2017-04-13
Disalias mimic interpreter frame shape and FSD

cathyzhyi:
FSD has been using mimic interpreter frame shape to facilitate
the transition from JITted code to interpreter which largely
limit the optimizations can be applied. By disaliasing FSD and
mimic interpreter frame shape, fsd uses OSR instead
and therefore allows more optimizations to happen.

### DIFF
--- a/compiler/control/OMROptions.hpp
+++ b/compiler/control/OMROptions.hpp
@@ -108,8 +108,7 @@ enum TR_CompilationOptions
    TR_RegisterMaps               = 0x00001000,
    TR_CreatePCMaps               = 0x00002000,
    TR_OptimizeForSpace           = 0x00004000,
-   TR_FullSpeedDebug             = 0x00008000,
-   TR_MimicInterpreterFrameShape = 0x00008000, // ** NOTE ** Alias for TR_FullSpeedDebug until we're prepared to separate these two concepts
+   TR_MimicInterpreterFrameShape = 0x00008000,
 
    TR_TraceBC                    = 0x00010000,
    TR_TraceInfo                  = 0x00020000,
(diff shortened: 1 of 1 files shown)


compiler/control/OMROptions.hpp
```diff
@@ -108,7 +108,6 @@ enum TR_CompilationOptions
    TR_RegisterMaps               = 0x00001000,
    TR_CreatePCMaps               = 0x00002000,
    TR_OptimizeForSpace           = 0x00004000,
-   TR_FullSpeedDebug             = 0x00008000,
    TR_MimicInterpreterFrameShape = 0x00008000, // ** NOTE ** Alias for TR_FullSpeedDebug until we're prepared to separate these two concepts
```
vijaysun-omr: Can we delete the comment to reflect the fact that there is no more aliasing with your change ?

compiler/control/OMROptions.hpp
```diff
@@ -267,7 +266,8 @@ enum TR_CompilationOptions
    TR_EnableNewAllocationProfiling        = 0x00010000 + 5, // enable tracing of fields load and store
    TR_IgnoreIEEERestrictions              = 0x00020000 + 5, // enable more aggressive, nonIEEE compliant xforms
    TR_ProcessHugeMethods                  = 0x00040000 + 5, // allow processing of huge methods
-   // Available                           = 0x00080000 + 5,
+   // Available                           =
```
vijaysun-omr: Again, the "Available" comment is no longer right.

vijaysun-omr:
If you can please update the comments I mentioned in my review, I will merge the change in.

compiler/control/OMROptions.hpp
cathyzhyi: yea, modified as suggested.

compiler/control/OMROptions.hpp
cathyzhyi: Deleted the "Available" comments.

vijaysun-omr:
Thanks for making the changes requested. Since only comments changed and the checks had passed earlier, I am merging this now.

PR #1017, opened 2017-04-18
Recognize AES-CBC cipher on x86

hjooybar:
AES-CBC cryptography can be accelerated using 64bit x86 instructions.
This changeset proposes two new recognized methods for accelerating
AEC-CBC-Encrypt and AES-CBC-Decrypt.
Jit is going to recognize these methods and replace them with handcrafted assembly routines


compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4184,6 +4186,18 @@ TR::Register *OMR::X86::TreeEvaluator::directCallEvaluator(TR::Node *node, TR::C
       {
       return TR::TreeEvaluator::andORStringEvaluator(node, cg);
       }
+   else if ((symbol->getRecognizedMethod() == TR::com_ibm_crypto_provider_AEScryptInHardware_cbcEncrypt)
+         && cg->enableAESInHardwareTransformations() && !disableECCP256AESCBC
+         && !(gcMode == TR_WrtbarCardMarkIncremental || gcMode == TR_WrtbarRealTime) && TR::Compiler->target.is64Bit())
```
charliegracie: Why can this optimization not be used for those GC barriers?

hjooybar: As far as I know, this code is there to ensure we do not apply this optimization on arraylets.
The code is added from VM28. It might be a better idea to use a function to check whether it is an arraylet mode.

fjeremic:
Please try to use x86 rather than "X" or "X86" in your commit messages to promote consistency and avoid any ambiguity.

compiler/x/codegen/OMRTreeEvaluator.cpp
hjooybar: Fixed

vijaysun-omr:
Thanks for updating the commit message. Since checks had passed earlier, I will merge this.

PR #1018, opened 2017-04-18
Avoid calling hardcoded helpers for FP conversion

0dvictor:
On X86-64, none of FP to integer conversion evaluators call hardcoded helpers;
On X86-32, only f2l and d2l still need hardcoded helpers.


vijaysun-omr:
I need a review from @0xdaryl on this since Andrew is away.

0dvictor:
Thank-you for the review, it has been rebased.

PR #1020, opened 2017-04-18
Provide an untyped pointer type in TypeDictionary

### LINKED ISSUE #511: Add pNoType to TR::TypeDictionary
Add the `pNoType` type to `TR::TypeDictionary`. This JitBuilder type maps to `void*` in C/C++.

lmaisons:
Add a pNoType definition to TypeDictionary.  This turns various indexing
and derefering oprations into potentially dangerous operations.  To help
prevent forest fires, add asserts to those operations to catch
inadvertent [and nonsensical] dereferences of untyped pointers.

Resolves #511.

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -574,6 +574,7 @@ IlBuilder::indirectLoadNode(TR::IlType *dt, TR::Node *addr, bool isVectorLoad)
    TR_ASSERT(dt->isPointer(), "indirectLoadNode must apply to pointer type");
    TR::IlType * baseType = dt->baseType();
    TR::DataType primType = baseType->getPrimitiveType();
+   TR_ASSERT(primType != TR::NoType, "Dereferencing an untyped pointer.");
    TR::DataType symRefType = primType;
    if (isVectorLoad)
       symRefType = symRefType.scalarToVector();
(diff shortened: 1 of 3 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -813,7 +814,7 @@ IlBuilder::IndexAt(TR::IlType *dt, TR::IlValue *base, TR::IlValue *index)
    {
    TR::IlType *elemType = dt->baseType();
    TR_ASSERT(elemType != NULL, "IndexAt should be called with pointer type");
-
+   TR_ASSERT(elemType->getPrimitiveType() != TR::NoType, "IndexAt of untyped pointer is nonsensical.");
```
mstoodle: Maybe "Cannot use IndexAt with pointer to NoType" ?

lmaisons:
I tried looking at jitbuilder/release/src/ToIlType.cpp and found that it already had coverage of pVoid_type.  Am I missing the intent?

Leonardo2718:
`toIlType<>()` doesn't depend on "cached" `TR::IlType` instances for pointer types. Instead, it unconditionally creates a new instance using `PointerTo()` (not ideal but it's simple and it works). So, `toIlType<void *>()` already works because it just returns whatever `PointerTo(NoType)` returns.

Just thought I'd clear that up :slightly_smiling_face:

mstoodle:
@lmaisons @Leonardo2718 you're right, it was an ill-considered request in the first place. Thanks for pursuing it to its inevitable conclusion.

PR #1021, opened 2017-04-18
Fix live registers issue for concurrent scavenge

ebadyano:
When running with compressed refs shift > 0, we are not evaluating shift
child since guarded load does both load and shift. We need to decrement
ref count on the right node. This code will get removed once we will
deliver proper fix for not anchoring l2a nodes for concurrent scavenge.

Signed-off-by: evgeniab@ca.ibm.com <evgeniab@ca.ibm.com>

### DIFF
--- a/compiler/z/codegen/UnaryEvaluator.cpp
+++ b/compiler/z/codegen/UnaryEvaluator.cpp
@@ -520,8 +520,10 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    node->setRegister(source);
 
 
-   //printf("Node %p is first child\n", firstChild); fflush(stdout);
-   //traceMsg(comp, "Node %p is first child\n", firstChild);
+   if (shouldGenerateGuardedLoadsForCompressedLoad && firstChild->containsCompressionSequence())
+      {
+      cg->decReferenceCount(firstChild->getFirstChild());
+      }
    cg->decReferenceCount(firstChild);
 
    return source;


fjeremic:
LGTM. This fixes an subtle but important issue. Without this change all reference registers end up staying live until the end of the compilation thus causing many unnecessary spills.

vijaysun-omr:
"We need to decrement ref count on the right node."...could you please elaborate on this ? It seems obvious that we should use the "right" node but how were we getting it wrong ? 

Also, what does getting the node right have to do with anchoring l2a nodes for concurrent scavenge ?

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -520,9 +520,14 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    node->setRegister(source);
 
 
-   //printf("Node %p is first child\n", firstChild); fflush(stdout);
-   //traceMsg(comp, "Node %p is first child\n", firstChild);
-   cg->decReferenceCount(firstChild);
+   if (shouldGenerateGuardedLoadsForCompressedLoad)
```
0xdaryl: Is this test sufficient?  Looking at the code above that reaches past the `firstChild` to evaluate it's firstChild, its gated with a `firstChild->containsCompressionSequence()` check.  Why is that not needed here?  There seems to be a case when `TR::Compiler->om.compressedReferenceShift() == 0` that will take you through this path as well.  Perhaps I don't understand all the subtle interactions behind these trees for concurrent scavenge, but on the surface there appears to be a small hole in the logic.

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -520,9 +520,14 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
+      {
+      cg->decReferenceCount(firstChild->getFirstChild());
```
0xdaryl: In light of your proposed fix, who now decrements the ref count on `firstChild` in this case?  For that matter, who evaluates the first child when `shouldGenerateGuardedLoadsForCompressedLoad` is true?

compiler/z/codegen/UnaryEvaluator.cpp
ebadyano: Good catch. I will add TR::Compiler->om.compressedReferenceShift() == 0 to the check

compiler/z/codegen/UnaryEvaluator.cpp
ebadyano: When we generate guarded load we don't need to do a shift, the result from load is shifted. This is a way to skip it. Is there a better way to skip evaluating the shift, do I need to decrement fistChild (shift node) if we never evaluate it?

ebadyano: is there a reason we don't use recursive decReferenceCount here?

0xdaryl: I don't think in general it's safe to skip evaluating a node.  It might be possible in some cases, but the circumstances have to be well controlled (and documented that this is what you're doing and why).  I'll defer to an optimizer expert to make the determination as to whether its safe in this case.

I don't think you need a recursive decReferenceCount here because the firstChild was evaluated, which should have done the node decrementing as part of that process.

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -520,9 +520,14 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    node->setRegister(source);
 
 
-   //printf("Node %p is first child\n", firstChild); fflush(stdout);
-   //traceMsg(comp, "Node %p is first child\n", firstChild);
-   cg->decReferenceCount(firstChild);
+   if (shouldGenerateGuardedLoadsForCompressedLoad || firstChild->containsCompressionSequence())
```
fjeremic: I don't think this is correct. If concurrent scavenge is not enabled then we wouldn't be anchoring the `l2a` nodes to treetops, hence `shouldGenerateGuardedLoadsForCompressedLoad` will be `false`. This means you will skip evaluation of `l2a` nodes when concurrent scavenge is not enabled which is incorrect. Please run a personal build with this change on upstream projects to see what I mean.

As a general rule every node you evaluate within an evaluator you should decrement its reference count. Any register you allocate that is not attached to any node you need to call `stopUsingRegister` on it. The fix should look something like this:

```
if (shouldGenerateGuardedLoadsForCompressedLoad && firstChild->containsCompressionSequence())
      {
      cg->decReferenceCount(firstChild->getFirstChild());
      }
   else
      {
      cg->decReferenceCount(firstChild);
      }
```

However I don't think this is the final solution. As @0xdaryl who decrements the reference count of `firstChild`? I don't think this is properly handled at the moment. Please run your use case with `-Xjit:paranoidOptCheck,enableParanoidRefCountChecks` and I suspect you will see asserts regarding reference counts of nodes not being zero at the end of the compilation.

ebadyano: right, && is the correct one, I agree. Checked in the fix. 

I will run with -Xjit:paranoidOptCheck,enableParanoidRefCountChecks on p47. Do I need to rebuild with prod assume or will these options work on a regular build?

ebadyano:
As per @fjeremic comment I ran with prod assumes build and I didn't see any asserts related to not dereferencing firstChild (shift node) as well as firstChild->getFirstChild() for cases comressrefs shift > 0 and shouldGenerateGuardedLoadsForCompressedLoad . But it might make sense to dereference firstChild even when we don't evaluate it. I ran with that fix on p47 and it completed okay.

@vijaysun-omr @0xdaryl  Just to clarify,  Irwin is currently working on a better solution to avoid rematerializing TR::compressedRefs trees into decompression sequences as this more closely models what our hardware instruction does (it loads and shifts all in one go so we shouldn't really model the load before the shift in the IL). Once that change is delivered we can remove this code and also code referenced by Filip in comment under https://github.com/eclipse/omr/pull/911

 Given the explanation please recommend whether we want this pushed into a z Systems specific codegen phase.

ebadyano:
@0xdaryl any other concerns?

0xdaryl:
I think the changes are ok given the context.

What do you mean by "please recommend whether we want this pushed into a z Systems specific codegen phase"?  It already is.  Or do you mean remove from OMR?  Either way its probably fine where it is.

ebadyano:
@vijaysun-omr @0xdaryl Can this get merged?

PR #1022, opened 2017-04-18
Add jitThrowIncompatibleReceiver helper

jdmpapin:
This is a J9-specific helper, but unfortunately the helpers are listed
in OMR at the moment.


compiler/il/Aliases.cpp
```diff
@@ -136,6 +136,7 @@ OMR::SymbolReference::getUseonlyAliasesBV(TR::SymbolReferenceTable * symRefTab)
             case TR_overflowCheck:
             case TR_nullCheck:
             case TR_methodTypeCheck:
+            case TR_incompatibleReceiver:
```
fjeremic: Perhaps we can guard them with `J9_PROJECT_SPECIFIC` to emphasize the fact that this is J9 specific and to make an easy eye catcher for future refactoring?

jdmpapin: I agree that would be good in general, but I'm not sure it should be part of this change. It would be strange to make `TR_incompatibleReceiver` conditional on `J9_PROJECT_SPECIFIC`, but not, say, `TR_methodTypeCheck` or `TR_multiANewArray`. And for some reason the declarations give an explicit value for each enumerator, which means that guarding just the declaration of `TR_incompatibleReceiver` would be awkward.

PR #1026, opened 2017-04-20
Heap Allocation Support for Concurrent Scavenger

dmitripivkine:
Concurrent Scavenger required for entire Nursery to be covered by
Concurrent Scavenger Page which should have size power of 2 and it's
start address should be aligned to page size.
The Concurrent Scavenger Page is split to 64 equal size sections.
Set region size equal section size to get proper heap alignment.
Change memory allocation for object heap to guarantee Concurrent
Scavenger Page alignment.  


gc/base/standard/ConfigurationGenerational.cpp
```diff
@@ -219,4 +219,20 @@ MM_ConfigurationGenerational::defaultMemorySpaceAllocated(MM_GCExtensionsBase *e
 	extensions->setGuaranteedNurseryRange(nurseryStart, (void*)UDATA_MAX);
 }
 
+uintptr_t
+MM_ConfigurationGenerational::calculateDefaultRegionSize(MM_EnvironmentBase *env)
+{
+	uintptr_t regionSize = STANDARD_REGION_SIZE_BYTES;
+
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
```
charliegracie: There was a new API added by @amicic.  Can you switch to using that API for code like this?  Thanks

gc/base/MemoryManager.cpp
```diff
@@ -67,8 +67,24 @@ MM_MemoryManager::createVirtualMemoryForHeap(MM_EnvironmentBase* env, MM_MemoryH
 	Assert_MM_true(0 != pageSize);
 
 	uintptr_t allocateSize = size;
-	if (heapAlignment > pageSize) {
-		allocateSize += (heapAlignment - pageSize);
+
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
```
charliegracie: Can you use the new API I mentioned below here as well?  Thanks

gc/base/MemoryManager.cpp
```diff
@@ -252,6 +291,74 @@ MM_MemoryManager::createVirtualMemoryForHeap(MM_EnvironmentBase* env, MM_MemoryH
 		instance->incrementConsumerCount();
 		handle->setMemoryBase(instance->getHeapBase());
 		handle->setMemoryTop(instance->getHeapTop());
+
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+		/*
```
charliegracie: Same thing here

charliegracie:
@dmitripivkine you will also need to rebase on the latest changes as the PR can not be automatically merged anymore.

gc/base/standard/ConfigurationGenerational.cpp
amicic: Well, in some places that API (ext->isConcurrentScavengerEnabled()) cannot be used... if symbols are defined for CS only, like here concurrentScavengerPageSectionSize.

dmitripivkine: Well, What I should do? Use compile flag OMR_GC_CONCURRENT_SCAVENGER + new Aleks API? Otherwise to have run-time check only I should move CS specific variables out from OMR_GC_CONCURRENT_SCAVENGER flag to make them common. Is it what we need?

dmitripivkine: It is fine for me however we need to be consistent with this

dmitripivkine: I am talking about fields in GCExtensionsBase.hpp
#if defined(OMR_GC_CONCURRENT_SCAVENGER)
	bool concurrentScavenger;	/* flag for check is Concurrent Scavenger active */
	bool debugConcurrentScavengerPageAlignment;
	uintptr_t concurrentScavengerPageSectionSize; /* selected section size for Concurrent Scavenger Page */
	void *concurrentScavengerPageStartAddress; /* start address for Concurrent Scavenger Page, UDATA_MAX if it is not initialized */
#endif /* defined(OMR_GC_CONCURRENT_SCAVENGER) */

PR #1027, opened 2017-04-20
Add option to disable OSR call site remat

ncough:
This adds an option to disable OSR call site remat, as there is a
significant risk it could present problems and will assist in
debugging.


vijaysun-omr:
Adding a new option and uses of it that look safe. Merging.

PR #1028, opened 2017-04-21
Add MVC support for reference array copy

fjeremic:
Currently the z Systems codegen is generating a load-store loop for
forward reference arraycopy because of a historic bug that is now moot.
This commit enables generation of the same MVC sequence for reference
arraycopy that we use for primitive arraycopy.

We take this time to reorganize, simplify, and document all arraycopy
evaluators on z Systems. We now have a consolidated path for primitive
and reference arraycopy with maximal code reusage.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -13461,454 +13461,133 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
    return NULL;
    }
 
-
-/**
- * arraycopyEvaluator - call to System.arraycopy that may be partially inlined
- */
-TR::Register *
-OMR::Z::TreeEvaluator::arraycopyEvaluator(TR::Node * node, TR::CodeGenerator * cg)
+TR::Register*
+OMR::Z::TreeEvaluator::arraycopyEvaluator(TR::Node* node, TR::CodeGenerator* cg)
    {
-   PRINT_ME("arraycopy", node, cg);
-
-   TR::Instruction * instr;
-   TR::LabelSymbol * snippetLabel;
-   TR::Snippet * snippet;
-
-   // Only a simple memmove or memcpy type of arraycopy is done as a node operation
-   // for 390 (this is different than other platforms which support other forms of
-   // arraycopy).
-   // For this case there are 3 children:
-   //    1) The byte source pointer
-   //    2) The byte destination pointer
-   //    3) The byte count
-   //
-   // Also, note that 390 codegen -requires- that isForwardArrayCopy() really means
-   // that a forward array copy is safe and that if it is -not- set it really means
-   // that a backward array copy is safe - the common code should have generated the
-   // correct tests around the arraycopy.
-   //
-   // Also, note that 390 codegen does not need to check for a particular frequency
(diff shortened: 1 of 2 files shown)


fjeremic:

mgaudet:
sp. not moot -> now moot. :D

charliegracie:
Are the reference array copies still happening a `word` at a time?

fjeremic:
> sp. not moot -> now moot. :D

Fixed.

> Are the reference array copies still happening a word at a time?

Quoting _Principles of Operation_ Section 5-125:

> When destructive overlap does not exist, the operands of MOVE (**MVC**), MOVE WITH KEY, MOVE TO
PRIMARY, and MOVE TO SECONDARY are accessed as follows:
>1. The first operand is accessed in a left-to-right direction, and all bytes accessed within a double 
word appear to be accessed concurrently as observed by other CPUs.
>2. The second operand is accessed in a left-to-right direction, and all bytes within a doubleword in the second operand that are moved into a single doubleword in the first operand appear to be fetched concurrently as observed by other CPUs. Thus, if the first and second operands begin on the same byte offset within a doubleword, the fetch of the second operand appears to be doubleword concurrent as observed by other CPUs. If the offsets within a doubleword differ by 4, the fetch of the second operand appears to be word concurrent as observed by other CPUs.

So yes MVC is block concurrent on a **double-word** boundary.

charliegracie:
Thanks for the info! Just making sure concurrent GCs do not see incorrect values during concurrent marking.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13461,454 +13461,133 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
-   if (copyByteDstReg)
-      {
-      cg->stopUsingRegister(byteDstReg);
-      }
-   if (copyByteLenReg)
+   else
       {
-      cg->stopUsingRegister(byteLenReg);
+      // TODO: There are cases under which the following assert will trigger. These need to be fixed and this assert needs to be re-enabled.
+      // TR_ASSERT(node->getArrayCopyElementType() == TR::Address, "Reference arraycopy element type should be a TR::Address but was '%s'", node->getArrayCopyElementType().toString());
```
dsouzai: It seems like this assert could be re-enabled now that we can only get here if it's a ref array copy but I'm not 100% sure about that. Worth making a note of this anyway.

dsouzai:
Looks fine to me. Probably worth having Joran review this for a better functional correctness review (I don't know what his GitHub handle is...).

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: This is not correct. Currently for example in VP we generate reference arraycopy with `TR::Int32` data type in some cases. This change is outside the scope of this PR hence why I left it as a TODO.

fjeremic:
> Looks fine to me. Probably worth having Joran review this for a better functional correctness review (I don't know what his GitHub handle is...).

@joransiu

fjeremic:
@joransiu any concerns or suggestions regarding this commit?

joransiu:
LGTM.  looks a lot cleaner now. :+1:

fjeremic:
@0xdaryl any suggestions from you?

PR #1031, opened 2017-04-21
WIP: Add new makefile variable OMR_ZOS_TARGET

babsingh:
-Wc,TARGET and -Wl,compat are zOS compiler options. Both must agree with
each other. OMR_ZOS_TARGET will allow values for -Wc,TARGET and
-Wl,compat to be specified from outside.

### DIFF
--- a/configure
+++ b/configure
@@ -741,6 +741,7 @@ LDFLAGS
 CFLAGS
 CC
 OMR_CROSS_CONFIGURE
+OMR_ZOS_TARGET
 OMR_TOOLCHAIN
 OMR_TARGET_DATASIZE
 OMR_HOST_ARCH
@@ -896,6 +897,7 @@ OMR_HOST_OS
 OMR_HOST_ARCH
 OMR_TARGET_DATASIZE
 OMR_TOOLCHAIN
+OMR_ZOS_TARGET
 OMR_CROSS_CONFIGURE
 CC
 CFLAGS
@@ -1690,6 +1692,8 @@ Some influential environment variables:
               One of: 31,32,64
   OMR_TOOLCHAIN
               The toolchain used to build the package. One of: gcc,xlc,msvc
+  OMR_ZOS_TARGET
+              The zOS target used during compilation. One of: zOSV1R13,zOSV2R1
   OMR_CROSS_CONFIGURE
               Indicates that configure is not being run on a system that is
               capable of building the package. Compiler and feature detection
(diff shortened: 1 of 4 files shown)


charliegracie:
@babsingh  is this more of a generic feature that other platforms may also want to set?

babsingh:
it is zOS compiler specific.

charliegracie:
@babsingh I think that may make sense.  @dnakamura and @rwy0717 does that seem like it would also be something that could be worked into the CMake work you have been prototyping? Is there already a generic way for doing it in CMake?

dnakamura:
No generic way that I am aware of. Would be fairly trivial to port to the cmake builds

babsingh:
@charliegracie making it a more generic feature will allow multiple compiler options in a single variable. also, it will be defined differently for different compilers. the various permutations of the generic variable will be large. this will make it impossible to set default values and error check. i think we should proceed with a variable only for -Wc,TARGET and -Wl,compat.

rwy0717:
looks good to me. In CMake I'd provide a [toolchain file](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) for setting z/os platform flags (Or just let the user set CXXFLAGS and call it a day). I get anxious about setting non-optional build flags.

dumb observation, but with the current solution you can't cross compile to a newer z/os target, because the build tools wouldn't run.

babsingh:
@rwy0717 is there an override option which turns AC_MSG_ERROR (errors) to warnings? this will let one compile with an older or newer z/os target that is unsupported.

rwy0717:
You can use `AC_MSG_WARN([message])`instead. I think that is an improvement to erroring out, but it's not solving my original issue.

My point was that, a packager should be able to set the build tool's z/os target independently of the OMR libs. However, this is not something anyone would likely ever need, so don't worry about it.

More on printing Messages [here](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Printing-Messages.html).

mstoodle:
@babsingh @charliegracie @rwy0717 what's the current status on this older PR? it's got a conflict now, but are there any outstanding problems to be resolved before it can be merged?

babsingh:
@mstoodle Due to build dependencies, flags don't properly propagate to OMR. This task has been paused until those dependencies are fixed. I will have this task merge ready once those dependencies have been fixed.

0xdaryl:
@babsingh, thanks for the explanation.  However, is there any value in keeping this PR open until your blocking prereqs are ready?  Unless this will happen within days I suggest closing this PR.

babsingh:
@0xdaryl Agreed. Closing this PR.

PR #1032, opened 2017-04-21
Correct CUDA runtime forward compatibility

### LINKED ISSUE #1030: CUDA version selection is inverted
As the comment in omrcuda.cpp says:
```
 * Include forward-compatible support for runtime libraries.
```
the intent was to support versions at runtime that are newer than the version used at compile time. However, the conditions are inverted, so, for example, if you compile with CUDA 5.5, only version 5.5 will be considered at runtime.

keithc-ca:
Corrects the preprocessor conditions controlling runtime support for CUDA runtime versions based on the CUDA version used at compile. time.

fixes #1030

### DIFF
--- a/port/common/omrcuda.cpp
+++ b/port/common/omrcuda.cpp
@@ -995,29 +995,29 @@ const J9CudaLibraryDescriptor runtimeLibraries[] = {
 /*
  * Include forward-compatible support for runtime libraries.
  */
-#if 8000 <= CUDA_VERSION
+#if CUDA_VERSION <= 8000
 	OMRCUDA_LIBRARY_ENTRY(8, 0),
-#endif /* 8000 <= CUDA_VERSION */
+#endif /* CUDA_VERSION <= 8000 */
 
-#if 7050 <= CUDA_VERSION
+#if CUDA_VERSION <= 7050
 	OMRCUDA_LIBRARY_ENTRY(7, 5),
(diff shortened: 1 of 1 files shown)


port/common/omrcuda.cpp
```diff
@@ -995,29 +995,29 @@ const J9CudaLibraryDescriptor runtimeLibraries[] = {
 /*
  * Include forward-compatible support for runtime libraries.
  */
-#if 8000 <= CUDA_VERSION
+#if CUDA_VERSION <= 8000
```
charliegracie: Where is the CUDA_VERSION flag set?

keithc-ca:
CUDA_VERSION is defined by cuda.h.

port/common/omrcuda.cpp
charliegracie: My understanding is that this code is building a list of all possible libraries that could be loaded based on the CUDA_VERSION macro from cuda.h.  If the version is 8000 create an entry for all of the known libraries <= to 8000.  Then code later on will try to open the libraries in descending order so that it opens the most `recent` library?

keithc-ca: No, that's what is used to do, but it's not what we wanted. If you build with CUDA 7.0, the 7.x and 8.x runtime libraries are compatible with the API in the 7.0 header files. The reverse is not true: you can't compile with 8.0 header files and expect to succeed using a 7.x runtime.

PR #1033, opened 2017-04-21
Adjust big callee threshold on z Systems

fjeremic:
It has been empirically tested that because z/Architecture lacks a call
return stack we tend to incur a higher penalty for calling methods. As
such inlining slightly more helps overall performance.

Increasing the big callee threshold has been found to causea marginal
increase in compile time at the benefit of very observable throughput on
the platform. As such making this increase the default.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -2478,6 +2478,8 @@ OMR::Options::jitPreProcess()
 
 #if defined(TR_HOST_POWER)
    _bigCalleeThreshold = 300;
+#elif defined(TR_HOST_S390)
+   _bigCalleeThreshold = 600;
 #else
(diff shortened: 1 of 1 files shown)


fjeremic:
@vijaysun-omr @mpirvu could you please review?

mpirvu:
Looks fine to me. Is it ok to provide the actual values of the improvement/regression here?

fjeremic:
> Looks fine to me. Is it ok to provide the actual values of the improvement/regression here?

On an upstream OMR project on a large server type application at the cost of ~13.5% in compile time we get back almost ~1% in peak throughput. Given the default settings throttle for throughput this is a worth while trade.

vijaysun-omr:
I agree with the tradeoff mentioned in this PR. Merging.

PR #1034, opened 2017-04-21
Trampoline Support for zOS

pli-:
Includes options to enable Trampolines on zOS, enable Trampolines
all the time for stress testing, enable code cache allocation
above the 2gb bar, use IIHF and IILF rather than BASR to jump,
and options to enable one big code cache.

Signed-off-by: Pingda Li <pingda@ca.ibm.com>


0xdaryl:
@fjeremic , could you review as well please?

mgaudet:
This should get rebased to remove the merge commit.

compiler/control/OMROptions.cpp
```diff
@@ -2690,7 +2692,19 @@ OMR::Options::jitPreProcess()
...
+#elif defined(TR_HOST_64BIT) && defined(J9ZOS390)
+      
+      // If we enabled the 2GB bar then addresses will now be from 0x80000000 to 0x7fffffffffffffff
+      // The way zOS allocates memory is by looking for 2MB free chunks and uses that for its 
+      // code caches. When we enable the 2GB bar then the addresses will be from 0x80000000
+      // to 0x7fffffffffffffff. This means two code caches could be very far apart from one another
+      // which is something we want to avoid as it may cause performance issues. The solution will 
+      // be to enable the code cache consolidation which will allocate 256MB at startup just for 
+      // code caches. This way the code caches will be together and improve performance. 
+      // Note: If all 256MB are used up then the program will crash.
```
mgaudet: Can we not just start aborting compiles when we hit 256mb of code cache usage?

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2319,6 +2325,71 @@ generateDirectCall(TR::CodeGenerator * cg, TR::Node * callNode, bool myself, TR:
 
       AOTcgDiag1(comp, "\ntempInst=%p\n", tempInst);
       return tempInst;
+#elif defined(TR_TARGET_64BIT) && defined(J9ZOS390) 
+      // If trampolines are enabled we will see if we need to use one
+      if (comp->getOption(TR_EnableTrampolines))
+         {
+         // With MultiCodeCache, we ALWAYS can jump to at least a trampoline.
+         #if !defined(PUBLIC_BUILD)
```
mgaudet: What purpose does this macro serve?

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2319,6 +2325,71 @@ generateDirectCall(TR::CodeGenerator * cg, TR::Node * callNode, bool myself, TR:
...
+            (new (INSN_HEAP) TR::S390RILInstruction(TR::InstOpCode::BRASL, callNode, RegRA, imm, callSymRef, cg));
+
+         AOTcgDiag1(comp, "\ntempInst=%p\n", tempInst);
+         return tempInst;
+         }
+      else
+         {
+         if (cg->canUseRelativeLongInstructions(imm) || isHelper)
+            {
+            #if !defined(PUBLIC_BUILD)
```
mgaudet: See Q above.

compiler/z/codegen/S390GenerateInstructions.cpp
pli-: Currently trampolines are always enabled on zLinux64, trampolines are not needed in zLinux32 and zOS32. So for the case of zOS64 we want it to go through this path.
I will delete the "#if !definied(PUBLIC_BUILD)", it used to be there from previous code and has now been removed.

compiler/control/OMROptions.cpp
pli-: My mistake, we will just abort compiles after we hit 256mb of code cache. I will fix the comment to say the correct message

charliegracie:
@mstoodle will this allow more of the JitBuilder tests/examples to be run on zOS?

mstoodle:
@charliegracie I don't *think* so, but it looks like it might do the hook up to the needed facilities when the trampoline infrastructure is available for OMR "methods". @fjeremic or @0xdaryl (who was looking into the method trampolines for POWER and Z a while back) can probably give a more definitive answer.

fjeremic:
> @mstoodle will this allow more of the JitBuilder tests/examples to be run on zOS?

In theory yes once enabled by default. Though the current code cache is 128mb which can fit (hundreds of?) thousands of compiled bodies so I'm not sure how this was a concern prior to this change.

mstoodle:
@fjeremic JitBuilder code cache repository is 16MB with 2MB code caches. But main problem is, I think, that the OMR code cache implementation does not have the method trampoline hash table(s) used by at least one downstream consumer to share method trampolines and I am not sure that the dispatch opcodes are taking the right path to generate trampolines all on their own.

charliegracie:
Ok thanks for the details everyone

compiler/control/OMROptions.cpp
```diff
@@ -607,6 +607,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"dumpIprofilerMethodNamesAndCounts",  "O\tDebug Printing of Method Names and Persisted Counts.", SET_OPTION_BIT(TR_DumpPersistedIProfilerMethodNamesAndCounts), "F"},
    {"dynamicThreadPriority",              "M\tenable dynamic changing of compilation thread priority", SET_OPTION_BIT(TR_DynamicThreadPriority), "F", NOT_IN_SUBSET},
    {"earlyLPQ",                           "M\tAllow compilations from low priority queue to happen early, during startup", SET_OPTION_BIT(TR_EarlyLPQ), "F", NOT_IN_SUBSET },
+   {"enableAboveTwoGBBar",                "O\tenable allocation of executable memory above the 2gb bar (zOS64)", SET_OPTION_BIT(TR_EnableAboveTwoGBBar), "F"},
```
fjeremic: We should change this option to `enableRMODE64`. Please take a look at [1] for reference. Perhaps the description could be "enable resident mode of compiled bodies on z/OS to reside above the the 2-gigabyte bar".

[1] https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.ieab100/resmode.htm

compiler/control/OMROptions.cpp
```diff
@@ -732,6 +733,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableSpecializedEpilogues",         "O\tapply restore non-volatiles selectively",         SET_OPTION_BIT(TR_EnableSpecializedEpilogues), "F"},
    {"enableTailCallOpt",                  "R\tenable tall call optimization in peephole", SET_OPTION_BIT(TR_EnableTailCallOpt), "F"},
    {"enableThisLiveRangeExtension",       "R\tenable this live range extesion to the end of the method", SET_OPTION_BIT(TR_EnableThisLiveRangeExtension), "F"},
+   {"enableTrampolines",                  "O\tenable trampolines to always be used (zLinux64 and zOS64)", SET_OPTION_BIT(TR_EnableTrampolines), "F"},
```
fjeremic: The comment implies this _forces_ trampolines to be used. Is this the case? Reading the code below it seems to imply that trampolines _can_ be used. Perhaps a better comment may be

```
enable generation of trampolines for method dispatch
```

Also while this option is currently used on Linux on z and z/OS there should be nothing preventing us from using the same option on other platforms. Hence this option should be platform agnostic.

compiler/control/OMROptions.cpp
```diff
@@ -2690,7 +2692,19 @@ OMR::Options::jitPreProcess()
...
+#elif defined(TR_HOST_64BIT) && defined(J9ZOS390)
+      
+      // If we enabled the 2GB bar then addresses will now be from 0x80000000 to 0x7fffffffffffffff
+      // The way zOS allocates memory is by looking for 2MB free chunks and uses that for its 
+      // code caches. When we enable the 2GB bar then the addresses will be from 0x80000000
+      // to 0x7fffffffffffffff. This means two code caches could be very far apart from one another
+      // which is something we want to avoid as it may cause performance issues. The solution will 
+      // be to enable the code cache consolidation which will allocate 256MB at startup just for 
+      // code caches. This way the code caches will be together and improve performance. 
+      // Note: If all 256MB are used up then the program will abort compiles.
```
fjeremic: In this comment we should replace "2GB bar" with "RMODE(64)" because we are talking about resident mode virtual addresses.

compiler/runtime/OMRCodeCacheManager.cpp
```diff
@@ -125,13 +125,13 @@ OMR::CodeCacheManager::initialize(
    config._needsMethodTrampolines =
       !(config.trampolineCodeSize() == 0
         || config.maxNumberOfCodeCaches() == 1
-#if !defined(STRESS_TRAMPOLINES) && !defined(TR_HOST_POWER)
+#if !defined(STRESS_TRAMPOLINES) && !defined(TR_HOST_POWER) && !defined(TR_TARGET_64BIT)
```
fjeremic: Could you please elaborate why the extra 32-bit check is needed here?

compiler/z/codegen/S390HelperCallSnippet.cpp
```diff
@@ -116,6 +116,20 @@ TR::S390HelperCallSnippet::emitSnippetBody()
...
       }
+#elif defined(TR_TARGET_64BIT) && defined(J9ZOS390)
+   if (cg()->comp()->getOption(TR_EnableTrampolines) && NEEDS_TRAMPOLINE(destAddr, cursor, cg()))
+      {
+      // Destination is beyond our reachable jump distance, we'll find the
+      // trampoline.
+      destAddr = cg()->fe()->indexedTrampolineLookup(helperSymRef->getReferenceNumber(), (void *)cursor);
+      this->setUsedTrampoline(true);
+      
+      // We clobber rEP if we take a trampoline.  Update our register map if necessary.
```
fjeremic: Extra space after the period.

compiler/z/codegen/S390Instruction.cpp
```diff
@@ -2010,7 +2010,12 @@ TR::S390RILInstruction::adjustCallOffsetWithTrampoline(int32_t offset, uint8_t *
 
    // Check to make sure that we can reach our target!  Otherwise, we need to look up appropriate
    // trampoline and branch through the trampoline.
+   
+#if defined(TR_TARGET_64BIT) && defined(STRESS_TRAMPOLINES)
+   if (1)
```
fjeremic: Can we change this to `if (true)`?

fjeremic: Also is it possible to create a compiler option `stressTrampolines` rather than this being a compile time flag? Is there any downside to this?

compiler/z/codegen/S390Instruction.cpp
```diff
@@ -2457,6 +2468,25 @@ TR::S390RILInstruction::generateBinaryEncoding()
...
+         if (calleeSymRef != NULL)
+            {
+            if (calleeSymRef->getReferenceNumber() >= TR_S390numRuntimeHelpers)
+               cg()->fe()->reserveTrampolineIfNecessary(comp, calleeSymRef, true);
+            }
+         else
+            {
+               #ifdef DEBUG
+               printf("Missing possible re-reservation for trampolines.\n");
+               #endif
```
fjeremic: Extra indentation here on `printf` and the preprocessor macros should not be indented at all.

compiler/runtime/OMRCodeCacheManager.cpp
pli-: The extra 32 bit check is added because when I was doing personal build tests the 32 bit tests would fail when stress trampoline is defined as they do not have trampoline code. I added the check so that running personal builds would be easier in the future. Should I remove it?

compiler/z/codegen/S390Instruction.cpp
pli-: I like the idea of stressTrampolines compiler option, should be a simple change from if(X) to if(getOption(stressTrampoline) OR X). I will talk with Joran to confirm.

lmaisons: From what I remember, STRESS_TRAMPOLINES is broken beyond recognition due to total incompatibility with code cache switching. It would probably be to remove it then to staple on another ifdef as though it mattered.

compiler/runtime/OMRCodeCacheManager.cpp
fjeremic: You can leave it if enabling stress trampolines is causing issues.

compiler/z/codegen/S390Instruction.cpp
fjeremic: @lmaisons it seemed to work fine when @pli- was stressing this change.

fjeremic:
LGTM.

pli-:
Can one of the committers restart the test, there was a time out, thanks.

compiler/control/OMROptions.cpp
```diff
@@ -732,6 +733,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableSpecializedEpilogues",         "O\tapply restore non-volatiles selectively",         SET_OPTION_BIT(TR_EnableSpecializedEpilogues), "F"},
    {"enableTailCallOpt",                  "R\tenable tall call optimization in peephole", SET_OPTION_BIT(TR_EnableTailCallOpt), "F"},
    {"enableThisLiveRangeExtension",       "R\tenable this live range extesion to the end of the method", SET_OPTION_BIT(TR_EnableThisLiveRangeExtension), "F"},
+   {"enableTrampolines",                  "O\tenable generation of trampolines for method dispatch", SET_OPTION_BIT(TR_EnableTrampolines), "F"},
```
0xdaryl: I find the presence of this option in its currently named form confusing.  Every other platform X86, Power, ARM, Z (to some extent) already require trampolines as a matter of correctness.  Enablement isn't really an option, its a requirement.  I don't see the existing platforms starting to use this option because I'm not sure how it would be applied and what the semantics would be, which is why I think it needs to be clarified and made specific to the context in which you need these trampolines generated (zOS).  If I'm not thinking about this the right way can you please clarify?

I do think a context where trampolines are always generated at a call site would be useful, but that's more along the lines of your other `stressTrampoline` option.

pli-: This option only enables zOS64 trampolines as I believe the feature will not be enabled by default yet. Perhaps renaming it to enablezOSTrampolines would make it clearer.

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2296,57 +2301,40 @@ generateDirectCall(TR::CodeGenerator * cg, TR::Node * callNode, bool myself, TR:
...
-         }
-
-      TR::S390RILInstruction *tempInst =
-            (new (INSN_HEAP) TR::S390RILInstruction(TR::InstOpCode::BRASL, callNode, RegRA, imm, callSymRef, cg));
-
-      AOTcgDiag1(comp, "\ntempInst=%p\n", tempInst);
-      return tempInst;
-#else
-      if (cg->canUseRelativeLongInstructions(imm) || isHelper)
+#if !defined(TR_TARGET_64BIT) || (defined(TR_TARGET_64BIT) && J9ZOS390)
```
0xdaryl: Shouldn't this be `... && defined(J9ZOS390))` ?

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2296,57 +2301,40 @@ generateDirectCall(TR::CodeGenerator * cg, TR::Node * callNode, bool myself, TR:
...
-
             }
 
          TR::S390RILInstruction *tempInst =
             new (INSN_HEAP) TR::S390RILInstruction(TR::InstOpCode::BRASL, callNode, RegRA, imm, cg);
 
-         if (isHelper)
-            tempInst->setSymbolReference(callSymRef);
+#if !defined(TR_TARGET_64BIT) || (defined(TR_TARGET_64BIT) && J9ZOS390)
+#if (defined(TR_TARGET_64BIT) && J9ZOS390)
```
0xdaryl: Shouldn't both of these lines be `... && defined(J9ZOS390))` ?

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2296,57 +2301,40 @@ generateDirectCall(TR::CodeGenerator * cg, TR::Node * callNode, bool myself, TR:
...
+#endif
+            {
+            if (isHelper)
+               tempInst->setSymbolReference(callSymRef);
+            }
+#endif
          AOTcgDiag1(comp, "\ntempInst=%x\n", tempInst);
          return tempInst;
          }
+#if !defined(TR_TARGET_64BIT) || (defined(TR_TARGET_64BIT) && J9ZOS390)
```
0xdaryl: Shouldn't this be `... && defined(J9ZOS390))` ?

pli-: Oops, fixed

0xdaryl:
As this project does not test Z as part of its build CI, please confirm that you tested this change on BOTH Linux and zOS and then I'll merge once TravisCI completes.

0xdaryl:
@pli- , can you confirm this passed Z testing on Linux and zOS?

pli-:
@0xdaryl I ran zOS and zLinux tests and there was a bug. I spent the day finding and fixing it and kicked off more tests. Will get back to you with results tomorrow morning. Thanks.

pli-:
http://zaxxon.torolab.ibm.com/build_info.php?build_id=136908
Toronto Personal build passing

0xdaryl:
@fjeremic , a few changes were made since your last review.  Please let me know if there are any concerns with merging.  Otherwise, I'll merge once Travis CI passes.

fjeremic:
@0xdaryl LGTM. Just went through the change again. It looks very clean now.

0xdaryl:
The Appveyor problems are still being worked out and are not related to this PR. Since the rest of CI passed cleanly I will merge.

PR #1036, opened 2017-04-23
Fix bcInfo of asyncCHK node

cathyzhyi:
AsyncCHKs are OSR points and VM relies on correct bytecode info
to transfer to the next bytecode execution. When inserting an
asyncCHK node, we need to make sure it has the same bytecode info
as the node right before or after. The original code makes the
inserted AsynChk have bytecode info of the next **treetop** which is
incorrect beause it might skip children of the node hanging under
that treetop and make VM believe the bytecode represented by the
children have already been executed and resume execution from the
wrong place.


andrewcraik:
LGTM

vijaysun-omr:
Simple change to make the bytecode info on an inserted async check more correct in a fairly straightforward way. Merging.

Issue #981, opened 2017-04-05
Extract MM_CollectorLanguageInterface::markingScheme_*(..) methods into MM_MarkingDelegate


ktbriggs:
This is a next step toward eliminating the MM_*LanguageInterface classes that were used to catch language-dependent factors while extracting core OMR out of J9 Java. These factors represent necessary integration points but the resulting MM_*LanguageInterface classes are fragmented and difficult to maintain. Moreover they force virtual implementations of member functions which can be a performance issue for functions that are frequently called in the runtime.

MM_CollectorLanguageInterface is a catch-all container for what are essentially delegate methods for a number of distinct GC components. Each of these methods is syntactically bound to a specific GC component by a prepending the GC component method name name onto the method name, viz:

parallelGlobalGC_\*(..)
concurrentGC_\*(..)
markingScheme_\*(..)
compactScheme_\*(..)
scavenger_\*(..)

Each of these groups of methods will be refactored into a corresponding delegate class that will implement equivalent factors, without forcing virtual methods. In the process the delegate methods will be rationalized and simplified wherever possible so that the delegate class interface will be easier to implement, use, and document. The essential delegate interface (delegate method signatures) will be common for all languages, but each implementation will be language-specific. The delegate classes (headers and implementation files) will be located in the language glue directory.

For this issue, markingScheme_*(..) methods will be extracted into a MM_MarkingDelegate class, and a persistent instance of this delegate class will be bound to the runtime MM_MarkingScheme component instance at run time. 

Also will extend GC_ObjectScanner framework as required to allow scanners to be used in marking contexts (currently, scanners are deployed only in scavenger). Marking scanner support will require scanners to indicate which reference slots point to leaf objects (if leaf optimization enabled).

ktbriggs:
Closed via https://github.com/eclipse/omr/pull/1037 (main commit) plus follow-up commits in https://github.com/eclipse/omr/pull/1063 and https://github.com/eclipse/omr/pull/1071.

PR #1037, opened 2017-04-24
Extract MM_MarkingDelegate from CLI

ktbriggs:
Refactored markingScheme_* methods and supporting artifacts from
MM_CollecotrLanguageInterface implementation in example/glue into a new
MM_MarkingDelegate class. This is continuation of effort to eliminate
*LanguageInterface classes in favor of a consistent component/delegate
model.

Use GC_ObjectScanner subclasses to scan objects for marking.

Issue: #981


gc/base/segregated/OverflowSegregated.cpp
```diff
@@ -154,7 +153,7 @@ MM_OverflowSegregated::handleOverflow(MM_EnvironmentBase *env)
 					/* scan overflowed object */
 					/* TODO Fix this as it is wrong for metronome GC policy */
 					/* The subclassing of markingscheme by segregatedmarking scheme needs a lot of work and means metronome is pretty broken */
-					_extensions->collectorLanguageInterface->markingScheme_scanObject(env, object, MM_CollectorLanguageInterface::SCAN_REASON_OVERFLOWED_OBJECT);
+					markingScheme->scanObject(env, object, MM_MarkingDelegate::SCAN_REASON_OVERFLOWED_OBJECT);
```
charliegracie: I think the scanReasons should have moved to MarkingScheme and not the delegate.

gc/base/MarkMap.hpp
```diff
@@ -33,6 +33,8 @@
 #define J9MODRON_HEAP_SLOTS_PER_MARK_SLOT J9MODRON_HEAP_SLOTS_PER_HEAPMAP_SLOT
 #define J9MODRON_HEAP_BYTES_PER_MARK_BYTE J9MODRON_HEAP_BYTES_PER_HEAPMAP_BYTE
 
+#define BITS_PER_BYTE 8
```
charliegracie: Is this true for all build types? 32 bit, 64 but, compressed references, etc?

gc/base/MarkingScheme.cpp
```diff
@@ -272,14 +294,15 @@ MM_MarkingScheme::setMarkBitsInRange(MM_EnvironmentBase *env, void *heapBase, vo
 void
 MM_MarkingScheme::completeScan(MM_EnvironmentBase *env)
 {
-	omrobjectptr_t objectPtr = NULL;
-	MM_WorkPackets *packets = getWorkPackets();
-
+	uintptr_t *bytesScanned = &(env->_markStats._bytesScanned);
```
charliegracie: Why are the stats just not being updated in the scanObject call?  Also why are they being cached locally? I feel like that confuses the code.  The compiler should be generating the same code even if you find them via the env everytime.

gc/base/IndexableObjectScanner.hpp
```diff
@@ -19,7 +19,7 @@
 #if !defined(INDEXABLEOBJECTSCANNER_HPP_)
 #define INDEXABLEOBJECTSCANNER_HPP_
 
-#include "ObjectScanner.hpp"
+#include "../ObjectScanner.hpp"
```
charliegracie: Can you remove the relative path please?

gc/base/MarkingScheme.cpp
```diff
@@ -263,7 +262,30 @@ MM_MarkingScheme::setMarkBitsInRange(MM_EnvironmentBase *env, void *heapBase, vo
...
+ * Private internal. Called exclusively from completeScan();
+ */
+uintptr_t
+MM_MarkingScheme::scanObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr)
+{
+	uintptr_t sizeToDo = UDATA_MAX;
+	GC_ObjectScannerState objectScannerState;
+	GC_ObjectScanner *objectScanner = _delegate.getObjectScanner(env, objectPtr, &objectScannerState, MM_MarkingDelegate::SCAN_REASON_PACKET, &sizeToDo);
+	if (NULL != objectScanner) {
+		bool isLeafSlot = false;
```
charliegracie: Instead of using the local does it make not more sense to have a isCurrentSlotLeaf() on the scanner?

gc/base/MarkingScheme.cpp
```diff
@@ -336,19 +359,16 @@ MM_MarkingScheme::completeMarking(MM_EnvironmentBase *env)
...
 }
 
 MM_WorkPackets *
 MM_MarkingScheme::createWorkPackets(MM_EnvironmentBase *env) {
 	MM_WorkPackets *workPackets = NULL;
 
-#if defined(OMR_GC_MODRON_CONCURRENT_MARK)
-	if(_extensions->concurrentMark) {
+	if (_extensions->isConcurrentMarkEnabled()) {
 		workPackets = MM_WorkPacketsConcurrent::newInstance(env);
```
charliegracie: This code should still be wrapped in the #if defined OMR_GC_MODRON_CONCURRENT_MARK since the class itself is probably wrapped in it.  If you compiled without it, it would likely fail to compile.

gc/base/MarkingScheme.hpp
```diff
@@ -139,13 +141,7 @@ class MM_MarkingScheme : public MM_BaseVirtual
 	MMINLINE bool
 	inlineMarkObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, bool leafType=false)
 	{
-		bool didMark = false;
```
charliegracie: I am not sure about these changes.  They make the code harder to read. Why did you want to change this code?

gc/base/MarkingScheme.hpp
```diff
@@ -154,13 +150,8 @@ class MM_MarkingScheme : public MM_BaseVirtual
 	MMINLINE bool
 	isMarked(omrobjectptr_t objectPtr)
 	{
-		bool shouldCheck = isHeapObject(objectPtr);
```
charliegracie: Same as previous comment

gc/base/MarkingScheme.hpp
```diff
@@ -169,43 +160,49 @@ class MM_MarkingScheme : public MM_BaseVirtual
 
 	void completeScan(MM_EnvironmentBase *env);
 	
-	MMINLINE void
-	scanObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, MM_CollectorLanguageInterface::MarkingSchemeScanReason reason)
+	/**
+	 * Public. Called from external context, eg concurrent GC.
+	 */
+	MMINLINE uintptr_t
+	scanObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, MM_MarkingDelegate::MarkingSchemeScanReason reason, uintptr_t sizeToDo = UDATA_MAX)
```
charliegracie: Couldn't this method be merged with the other scanObject and just take a default parameter of UDAT_MAX for the size to do?

gc/base/MarkingScheme.hpp
```diff
@@ -169,43 +160,49 @@ class MM_MarkingScheme : public MM_BaseVirtual
...
 		 */
-		env->_markStats._bytesScanned += sizeScanned;
-	}
-
-#if defined(OMR_GC_MODRON_CONCURRENT_MARK)
-	MMINLINE uintptr_t
-	scanObjectWithSize(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, MM_CollectorLanguageInterface::MarkingSchemeScanReason reason, UDATA sizeToDo)
-	{
-		if(MM_CollectorLanguageInterface::SCAN_REASON_PACKET == reason) {
+		env->_markStats._bytesScanned += sizeToDo;
```
charliegracie: Where is this function called from? What is the purpose?  It just seems to be updating some stats.

gc/base/ObjectScanner.hpp
```diff
@@ -19,14 +19,20 @@
...
 #include "ModronAssertions.h"
 #include "objectdescription.h"
 
 #include "BaseVirtual.hpp"
-#include "EnvironmentStandard.hpp"
+#include "EnvironmentBase.hpp"
 #include "GCExtensionsBase.hpp"
 #include "SlotObject.hpp"
 
+/* TODO: define & configure OMR_GC_LEAF_BITS for omrcfg.h */
```
charliegracie: Would it be possible to include this change in the PR?

gc/base/standard/ParallelGlobalGC.cpp
```diff
@@ -1035,11 +1035,11 @@ MM_ParallelGlobalGC::collectorStartup(MM_GCExtensionsBase* extensions)
 void
 MM_ParallelGlobalGC::collectorShutdown(MM_GCExtensionsBase *extensions)
 {
-#if defined(J9VM_GC_MODRON_SCAVENGER)
+#if defined(OMR_GC_MODRON_SCAVENGER)
```
charliegracie: good catch.  Thanks you for fixing this!

gc/base/segregated/OverflowSegregated.cpp
ktbriggs: I agree. But MarkingDelegate.hpp can't include MarkingScheme.hpp. I will move them to omrgcconsts.h.

gc/base/MarkMap.hpp
ktbriggs: Yup. I just moved it here from MarkMap.hpp. Definition is unchanged.

gc/base/MarkingScheme.cpp
ktbriggs: It would be messy. Scanners advance position before returning from getNextSlot(). So isCurrentSlotLeaf() would have to backtrack.

gc/base/MarkingScheme.cpp
ktbriggs: This (completeScan()) is the innermost scanning loop. I also factored out of scanObject() a more efficient private version of scanObject() that avoids duplicate NULL check on objects returned by scanners. When I did that it just made sense to move stats update into completeScan() call and cache them. I agree that, with this refactoring of stats update out of reduced scanObject(), smart compiler should cache stats addresses anyway, but not so sure that it would do so if stats were always updated in scanObject().

The caching is done only the private scanObject(), and takes away any question of whether the compiler is smart enough to do this. I don't agree that this caching makes the code more complex. And I think that the private scanObject() has an overall positive impact on marking scan times.

gc/base/MarkingScheme.cpp
ktbriggs: I will fix this.

gc/base/IndexableObjectScanner.hpp
ktbriggs: These are being inserted by configure. I did not add relative path (avoid them altogether if I can).

Will remove and look for others. Should probably open an issue relating to relative paths and configure. This was Robert's ides if I recall. If he backed it out he may have missed something.

gc/base/MarkingScheme.hpp
ktbriggs: Because the code was hard to read. Unnecessarily verbose, spurious use of local variables obfuscates how result is obtained. Current form is much more clear and explicit IMO.

gc/base/MarkingScheme.hpp
ktbriggs: Same reply.

gc/base/MarkingScheme.hpp
ktbriggs: See previous comment. At least, private scanObject() avoids unnecessary NULL check, and allows for more efficient updating of mark stats.

gc/base/MarkingScheme.hpp
ktbriggs: It was called from concurrent GC, overflow handling, other contexts external to marking scheme. It has been replaced with the public scanObject(.., <optional size parameter>) method presented in MM_MarkingDelegate.

gc/base/ObjectScanner.hpp
ktbriggs: Yes. My intention. Would also like to discuss renaming MM_MarkingDelegate functions (current names are kind of disjointed IMO).

ktbriggs:
Latest commit addresses comments from review of OMR pull request (https://github.com/eclipse/omr/pull/1037). Most of the changes relate to handling of MarkingSchemeScanReason definition (moved to omrgcconsts.h, renamed to provide MM_ prefix).

Some of the comments above haven't been addressed in this commit, for reasons given in replies above. I am open to further discussion on these.

gc/base/MarkMap.hpp
charliegracie: Thanks

gc/base/MarkingScheme.hpp
charliegracie: So it is no longer called? Or it is just called to update the stats?

ktbriggs: It no longer exists. All sites that previously called this method now call the public version of scanObject(...), which receives size as a parameter. This version updates stats inside, as did previous scanObjectWithSize(). The private version of scanObject(), which is called exclusively from the main completeScan() loop, leaves stats update to caller, avoids NULL object check, and avoids  test of `SCAN_REASON_PACKET == reason` which is necessary in the public version.

ktbriggs:
Latest commit addresses comments from review of OMR pull request (https://github.com/eclipse/omr/pull/1037). Most of the changes relate to handling of MarkingSchemeScanReason definition (moved to omrgcconsts.h, renamed to provide MM_ prefix).

Some of the comments above haven't been addressed in this commit, for reasons given in replies above. I am open to further discussion on these.

gc/base/MarkingScheme.hpp
charliegracie: Sorry I mis-read the diff.

gc/base/MarkingScheme.hpp
charliegracie: Verbosity never hurts as source file size is never an issue.  I am ok with these changes as there are only 2 logical operations happening on the line.  In the future for changes like these can they be pulled into a separate PR / commit as they are just refactoring to functions that otherwise would not have been changed.  Thanks

gc/base/ObjectScanner.hpp
charliegracie: So you are going to add this to this PR?

gc/base/MarkingScheme.hpp
```diff
@@ -169,43 +161,49 @@ class MM_MarkingScheme : public MM_BaseVirtual
...
+	/**
+	 * Public. Called from external context, eg concurrent GC.
+	 */
+	MMINLINE uintptr_t
+	scanObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, MM_MarkingSchemeScanReason reason, uintptr_t sizeToDo = UDATA_MAX)
 	{
-		if(MM_CollectorLanguageInterface::SCAN_REASON_PACKET == reason) {
-			env->_markStats._objectsScanned += 1;
+		GC_ObjectScannerState objectScannerState;
+		GC_ObjectScanner *objectScanner = _delegate.getObjectScanner(env, objectPtr, &objectScannerState, reason, &sizeToDo);
```
charliegracie: I understand why the sizeToDo is passed so that it can be limited.  I feel like that complicates the scanner code and it will make it harder for implementers to implement their own scanner.

gc/base/MarkingScheme.hpp
```diff
@@ -169,43 +161,49 @@ class MM_MarkingScheme : public MM_BaseVirtual
...
+		if (NULL != objectScanner) {
+			bool isLeafSlot = false;
+			GC_SlotObject *slotObject;
+#if defined(J9VM_GC_LEAF_BITS)
```
charliegracie: Can you switch this to the OMR version?

gc/base/MarkingScheme.cpp
ktbriggs: Caching removed.

ktbriggs:
All comments have been addressed. I'm going to defer renaming methods in the marking delegate until I have completed the other GCs in CLI. I think there will be a pattern across GCs and if so will rename consistently for all GC component delegates.

gc/base/MarkingScheme.hpp
charliegracie: Actually after discussing this with a few people I believe I would like these changes reverted.  You lose debugability by having all of these on the same line.  The code is way easier to add break points and debug when things are on separate lines.  Once these two functions are revered and the changesets squashed I will accept the PR.  Thanks

ktbriggs:
Ok. Squashed all commits for pull. Only changes made since last commit (reviewed) are:

```
MMINLINE bool
inlineMarkObject(MM_EnvironmentBase *env, omrobjectptr_t objectPtr, bool leafType=false)
{
	bool objectMarked = false;

	if (NULL != objectPtr) {
		objectMarked = inlineMarkObjectNoCheck(env, objectPtr, leafType);
	}

	return objectMarked;
}

MMINLINE bool
isMarked(omrobjectptr_t objectPtr)
{
	bool marked = true;
		
	/* Everything off-heap is considered marked, everything on-heap must be checked */
	if (isHeapObject(objectPtr)) {
		marked = _markMap->isBitSet(objectPtr);
	}

	return marked;
}
```

Mildly protesting these changes. To me, they are more difficult to parse. Inline check in return statement is much easier to parse and understand at a glance. As for debugging, just set breakpoint on return statement and debug away.

PR #1038, opened 2017-04-24
JProfiling bug fix

mpirvu:
comp->isJProfilingCompilation() was recently changed to always return false
on the account that there is a similar query in CompilationInfo.
However, comp->isJProfilingCompilation() is used in the inliner.
This query will be replaced by  comp->getOption(TR_EnableJProfiling) which is
correct as long as the JIT doesn't change its mind while optimizing the
method in hand. Currently the JIT never changes its mind once it decided it
is going to generate a JProfiling body. Even if the JIT will turn off
JProfiling midway while optimizing a method, the resulting code is still
correct with no runtime performance penalty. The only downside is a small
compilation time increase.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -2639,7 +2639,7 @@ TR_TransformInlinedFunction::transform()
    // If the first block has exception predecessors or multiply predecessors then we can't merge
    // the first block with the caller's block
    //
-   if (comp()->isJProfilingCompilation() ||
+   if (comp()->getOption(TR_EnableJProfiling) ||
        (firstBlock->getPredecessors().size() > 1) ||
        firstBlock->hasExceptionSuccessors() ||
        comp()->fe()->isMethodEnterTracingEnabled(calleeResolvedMethod->getPersistentIdentifier()) ||


mpirvu:
@andrewcraik Andrew, please review this PR.

andrewcraik:
LGTM

andrewcraik:
@mpirvu seems to be an issue with getJittedBodyInfo not existing in some OMR contexts? That will need to be fixed, but the spirit of the change looks fine.

mpirvu:
I checked how we get to getJittedBodyInfo from other parts of OMR code. It turns out all instances are protected by #ifdef J9_PROJECT_SPECIFIC. Is that an acceptable solution?

andrewcraik:
I think it would be better to provide an implementation here that returns false and then we can have a J9 implementation in the J9 Compilation for the J9 project which has the correct logic I think. No need for a macro when the extension point already exists IMO.

mpirvu:
ok. In this commit I will leave only the Inliner change.
In gitlab I will change J9Compilation.hpp/cpp with a new definition for isJProfilingCompilation()

mpirvu:
@andrewcraik  It appears that OMR::jProfiling is only defined in trj9 despite being part of the OMR namespace.
I may have to use the option TR_EnableJProfiling which is defined in OMROptions.hpp.
On the whole it's not ideal to talk in the OMR layer (inliner in this case) about something only defined/used in trj9. Is there a preferred alternative?

andrewcraik:
The OMR::jProfiling is a J9 only optimization and the way extensible enums work the J9 specific opts are added to the enum in the OMR namespace - this is by design. Ultimately the whole thing should eventually be moved up to OMR, but that is a bigger job since the whole profiling infrastructure needs to be refactored and contributed. If it makes it easier we could declare the opt enum in OMR as a placeholder for the later contribution.

mpirvu:
This PR is ready to be merged. It's J9 specific sibling has already been merged.

andrewcraik:
LGTM

PR #1040, opened 2017-04-24
Allow late decision whether to obey CS request

amicic:
Concurrent Scavenger: Introduce an additional 'requested' flag that will be set during parsing
of the CS command line option. There might be platform/language specific
requirements after we parsing is done that will dictate whether the
request is obeyed or not.

Add a line in 'initialized' verbose GC stanza to report if we actually
run CS or not.


charliegracie:
@amicic can you modify the first line of the commit to 50 characters

PR #1043, opened 2017-04-25
Add flag to skip setting frequency

ncough:
Currently, frequencies will be calculated and set before an
optimization when all necessary information is present. It is then
maintained across modifications to the CFG. This necessary information
includes structure analysis.

This change allows an optimization to explicitly prevent frequencies
from being set before itself. This is useful for early opts that require
structure, but not frequency, and may significantly alter the structure
to the point where the frequency data would not have been adequately
maintained.


andrewcraik:
LGTM - a helpful change to prevent premature frequency setting IMO.

vijaysun-omr:
Okay as an infrastructural change to allow an optimization to avoid setting frequencies. I'm a bit curious which optimization we intend to set the flag for, though they are all in downstream projects.

PR #1044, opened 2017-04-25
Find stores/returns after inlined calls in OSR

ncough:
In the event that the result of an inlined call is referenced once
after it, in a store or a return, the inliner is able to use that
treetop directly rather than requiring a temp to return the result.

This behaviour is broken by postExecutionOSR as all calls that are
OSR points will be followed by a pending push that stores its
result. The correction here skips this pending push to attempt
the existing analysis. This is valid under OSR as the inlined call
is no longer an OSR point and its result should not be on the stack
if it is stored or returned immediately.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -3625,16 +3625,40 @@ bool TR_TransformInlinedFunction::onlyMultiRefNodeIsCallNode(TR::Node * callNode
 
 TR::TreeTop * TR_TransformInlinedFunction::findSimpleCallReference(TR::TreeTop * callNodeTreeTop, TR::Node * callNode)
    {
+   // If we are in postExecution OSR and the call is an OSR point, it will always be followed
+   // by a series of PP treetops, one of which will reference the call result.
+   //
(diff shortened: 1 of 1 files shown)


compiler/optimizer/Inliner.cpp
```diff
@@ -3625,24 +3625,52 @@ bool TR_TransformInlinedFunction::onlyMultiRefNodeIsCallNode(TR::Node * callNode
...
+
+      // Search the corresponding OSR nodes for a reference to the call
+      //
+      TR::TreeTop * pps = NULL;
+      if (osrPP)
+         {
+         TR_ByteCodeInfo &bci = callNode->getByteCodeInfo();
+         while (comp()->getMethodSymbol()->isOSRRelatedNode(nextTreeTop->getNode(), bci))
+            {
+            if (nextTreeTop->getNode()->getOpCode().isStore() && nextTreeTop->getNode()->getFirstChild() == callNode)
```
andrewcraik: Should there be an assertion that we found only one rather than assuming the last one that matches is the "right" one to look at?

PR #1045, opened 2017-04-25
Use consistent exclusive access backoff

gacholio:
The tryAcquireExclusiveVMAccess path in the GC will wait for another
thread if the other thread is already acquiring exclusive VM access.  It backs off
using releaseCriticalHeapAccess and reacquireCriticalHeapAccess.  The
acquireExclusiveVMAccess path uses releaseVMAccess(bool) and acquireVMAccess.
Change acquireExclusiveVMAccess to be consistent with
tryAcquireExclusiveVMAccess so the delegate may more easily detect the
exclusive backoff case.


charliegracie:
This change makes sense to me.

PR #1047, opened 2017-04-25
Default isConcurrentScavengeEnabled to false

fjeremic:
Concurrent scavenge requires GC support, hardware checks, OS level
checks, and guarded storage block initialization to be properly done.
The infrastructure for such plumbing work should be done at a downstream
OMR project level. As such we are defaulting concurrent scavenge to be
disabled by default unless a downstream project extends the
CodeGenerator class and overrides the value.

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -6095,16 +6095,6 @@ OMR::Z::CodeGenerator::supportsMergingOfHCRGuards()
           !self()->comp()->compileRelocatableCode();
    }
 
-bool
-OMR::Z::CodeGenerator::isConcurrentScavengeEnabled()
-   {
-#if defined(OMR_GC_CONCURRENT_SCAVENGER)
-   return TR::Compiler->target.cpu.getS390SupportsGuardedStorageFacility();
-#else
-   return false;
-#endif
-   }
-
 // Helpers for profiled interface slots
 void
 OMR::Z::CodeGenerator::addPICsListForInterfaceSnippet(TR::S390ConstantDataSnippet * ifcSnippet, TR::list<TR_OpaqueClassBlock*> * PICSlist)
--- a/compiler/z/codegen/OMRCodeGenerator.hpp
+++ b/compiler/z/codegen/OMRCodeGenerator.hpp
@@ -433,15 +433,6 @@ public:
 
    bool supportsMergingOfHCRGuards();
 
-   /** \brief
-    *     Determines whether concurrent scavenge of objects during garbage collection is enabled.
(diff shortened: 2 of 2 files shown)


fjeremic:

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -6095,16 +6095,6 @@ OMR::Z::CodeGenerator::supportsMergingOfHCRGuards()
...
 
-bool
-OMR::Z::CodeGenerator::isConcurrentScavengeEnabled()
-   {
-#if defined(OMR_GC_CONCURRENT_SCAVENGER)
-   return TR::Compiler->target.cpu.getS390SupportsGuardedStorageFacility();
-#else
-   return false;
-#endif
-   }
```
dsouzai: I'm not sure you can delete this, unless every single project that builds on top of OMR implements this function - for example there's code in `omr/compiler/z/codegen/ControlFlowEvaulator.cpp` that does a `if (cg->isConcurrentScavengeEnabled())` query which would fail to resolve at runtime. I think you would need to keep it in and have it return false (probably just in the .hpp file)

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -433,15 +433,6 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
...
 
-   /** \brief
-    *     Determines whether concurrent scavenge of objects during garbage collection is enabled.
-    *
-    *  \return
-    *     true if the code generator will emit read barriers for loads of object references from the heap in support
-    *     of concurrent scavenge; false otherwise.
-    */
-   bool isConcurrentScavengeEnabled();
-
```
dsouzai: Same as comment above

compiler/z/codegen/OMRCodeGenerator.cpp
fjeremic: @dsouzai the function is already defined in the base code generator. We were only overloading it here via extensible class. See [1].

[1] https://github.com/eclipse/omr/blob/master/compiler/codegen/OMRCodeGenerator.hpp#L499-L509

compiler/z/codegen/OMRCodeGenerator.hpp
fjeremic: See above for explanation.

compiler/z/codegen/OMRCodeGenerator.cpp
dsouzai: Ah ok, then we're all good.

dsouzai:
LGTM

PR #1051, opened 2017-04-25
Add const length arraycopy opt on X86

jingge-ibm:
Generate more efficient instruction sequences for arraycopy when the
length is constant and less than 256 (for forward and backward copy)
or less than 64 (for no direction copy)


compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,244 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+         case 4:
+            storeOpcode = MOVDMemReg;
+            break;
+         case 8:
+            storeOpcode = MOVQMemReg;
+            break;
+         case 16:
+            storeOpcode = MOVDQUMemReg;
+            break;
+         }
```
mgaudet: IMO `default:` should exist (and probably be a `TR_ASSERT` (`_FATAL`?)).

Same comment for new switches below.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,244 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+ * Generate a store instruction
+ * Parameters:
+ * node: the tree node
+ * addressReg: the base register for the destination mem address
+ * index: the index for the destination mem address
+ * valueReg: the value that needs be stored
+ * size: the size of valueReg
+ * cg: the code generator
+ */
+static void generateStoreInstructions(TR::Node* node, TR::Register* addressReg, int32_t index, TR::Register* valueReg, uint8_t size,  TR::CodeGenerator* cg)
```
fjeremic: Awesome that you wrote documentation for newly added functions! It would be great if you could Doxygen this function comment so as to stay consistent with other parts of our codebase. There are numerous examples around if you grep. Similarly elsewhere in this commit.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3423,9 +3661,31 @@ TR::Register *OMR::X86::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::Co
...
+   if (sizeNode->getOpCode().isLoadConst() && TR::Compiler->target.is64Bit())
+      {
+      size = TR::TreeEvaluator::integerConstNodeValue(sizeNode, cg);
+      if (node->isForwardArrayCopy() || node->isBackwardArrayCopy())
+         {
+         if (size <= 256) isShortConstArrayWithDirection = true;
+         }
+      else
+         {
+         if (size <= 16*4) isShortConstArrayWithoutDirection = true;
```
fjeremic: Would it be better to name these magic constants even as static private const variables or `#defines`? This way we could also document them and have an easy way to change them if needed in the future.

fjeremic:
Please try to use "x86" rather than "X" in your commit comments or code to be more explicit and prevent any ambiguity.

compiler/x/codegen/OMRTreeEvaluator.cpp
andrewcraik: I would much prefer consts/static consts to #defines - makes it way easier for the compiler and other tools to deal with the constants as what they are rather than magic token pasting from the preprocessor which is evil.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,244 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+            break;
+         case 8:
+            storeOpcode = MOVQMemReg;
+            break;
+         case 16:
+            storeOpcode = MOVDQUMemReg;
+            break;
+         }
+      }
+   else
```
andrewcraik: There are vector registers as well as GPRs - for safety make the GPR case explict and add an else and assert in case you don't get either of the two handled kinds of registers. Makes future extension easier to handle too.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,244 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+ * Generate a load instruction
+ * Parameters:
+ * node: the tree node
+ * valueReg: the destination register
+ * size: the size of the loaded value
+ * addressReg: the base register for the source mem address
+ * index: the index for the source mem address
+ * cg: the code generator
+ */
+static void generateLoadInstructions(TR::Node* node, TR::Register* valueReg, uint8_t size, TR::Register* addressReg, int32_t index, TR::CodeGenerator* cg)
```
andrewcraik: There seems to be a very similar structure to the store case - any particular reason you chose not to combine them in some way? Is this more clear? Just seems like it could be a bit of duplication...

jingge-ibm: Hi Andrew. It did this mostly because it is more clear. The cost of duplication shouldn't be very big.

andrewcraik: Ok that makes sense.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+
+static void arraycopyForShortConstArrayWithDirection(TR::Node* node, TR::Register* dstReg, TR::Register* srcReg, uint32_t size, TR::CodeGenerator *cg)
+   {
+   uint32_t totalSize = size;
+   static uint32_t regSize[] = {16, 8, 4, 2, 1};
+
+   TR::Register* xmmReg = NULL;
+   TR::Register* gprReg = NULL;
+
+   int i = 0;
```
0xdaryl: int32_t for code consistency

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+
+         }
+      }
+   else
+      {
+      TR_ASSERT(0, "Unsupported register type in generateLoadInstructions");
+      }
+   generateRegMemInstruction(loadOpCode,  node, valueReg, generateX86MemoryReference(addressReg, index, cg), cg);
+   }
+
```
0xdaryl: Doxygen comment for new function with multiple parms?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+      else
+         {
+         i++;
+         }
+      }
+   if (xmmReg) cg->stopUsingRegister(xmmReg);
+   if (gprReg) cg->stopUsingRegister(gprReg);
+   }
+
+static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Register* dstReg, TR::Register* srcReg, uint32_t size, TR::CodeGenerator *cg)
```
0xdaryl: Doxygen comment for new function with multiple parms?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+      generateLoadInstructions(node, xmmReg, 16, srcReg, i*16, cg);
+      xmmUsed[i] = xmmReg;
+      }
+
+   // Note: The worst case, moves = [X, 1, 1, 1, 1]
+   //
+   // If we can do it just using one gpr, do it
+   // If we can do it with one gpr and one xmm, do it
+   // If total size is >= 16, use shift window
+   // Use two gpr
```
0xdaryl: Should this be "Otherwise, use two gprs"  ?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+ *     The value that needs be stored
+ *
+ *  \param size
+ *     The size of valueReg
+ *
+ *  \param cg
+ *     The code generator
+ *
+ */
+static void generateStoreInstructions(TR::Node* node, TR::Register* addressReg, int32_t index, TR::Register* valueReg, uint8_t size,  TR::CodeGenerator* cg)
```
jingge-ibm: This function is used by arraycopy and arrayset.

andrewcraik: The issue isn't the callers the issue is that you said generateStoreInstructions when the implementation will generate a single store instruction is what @0xdaryl was pointing out.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3371,6 +3371,287 @@ static void arrayCopyDefault(TR::Node* node, uint8_t elementSize, TR::Register*
...
+ *  \param addressReg
+ *     The base register for the source mem address
+ *
+ *  \param index
+ *     The index for the destination mem address
+ *
+ *  \param cg
+ *     The code generator
+ */
+static void generateLoadInstructions(TR::Node* node, TR::Register* valueReg, uint8_t size, TR::Register* addressReg, int32_t index, TR::CodeGenerator* cg)
```
jingge-ibm: This function is also used by arraycopy and arrayset.

compiler/x/codegen/OMRTreeEvaluator.cpp
jingge-ibm: Sorry for the confusing, by my previous comment, I meant that probably we don't want to rename it to be something specific to arraycopy because this function is used in arraycopy and arrayset; this is for the second point in @0xdaryl 's comment.
As for the first point (plural problem), I will change it to singular.

compiler/x/codegen/OMRTreeEvaluator.cpp
jingge-ibm: Sorry I was wrong in the previous comment. This function is not used in arrayset, only in arraycopy. I can rename it to something specific to arraycopy

compiler/x/codegen/OMRTreeEvaluator.cpp
0xdaryl: Plurality is only a minor issue.  I think the name of the function is still far too generic and the addressing form too specialized in and amongst a file of many other evaluators.  Even `generateArrayElementStore` is better in my opinion.

jingge-ibm:
@0dvictor helped me verify the benchmark result, it does not bring any regression. 
For functional testing, the personal tests have passed

0dvictor:
Yes, both OMR's test and IBM's internal testsuites passed. In terms of performance, no measurable regression was observed.

Issue #1050, opened 2017-04-25
ThreadAbortTest.sleeping timeouts


mgaudet:
Maybe this is selection bias, but I feel like I'm seeing more [ThreadAbortTest failures than usual](https://travis-ci.org/eclipse/omr/jobs/225693207). Is this a test we can tweak to make more reliable? 

```
[----------] 1 test from ThreadAbortDeathTest (207 ms total)
[----------] 3 tests from ThreadAbortTest
[ RUN      ] ThreadAbortTest.Running
[       OK ] ThreadAbortTest.Running (1 ms)
[ RUN      ] ThreadAbortTest.Sleeping
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

We just restart the builds to fix this typically.

rwy0717:
I believe this is resolved with PR #1052.

PR #1052, opened 2017-04-25
Fix a race in ThreadAbortTest.Sleeping

rwy0717:
In the original test, the sleeper could enter and notify on enterSync
before main hits wait.  This could result in the main thread waiting
for a notify that never comes.

The new test also replaces a hardcoded sleep in main with a notify. In
an ideal world, it would be safe to call abort as soon as
omrthread_create returns.

This probably fixes issue #1050, where this test was hanging. It now
takes ~1second on my laptop.

### DIFF
--- a/fvtest/threadtest/abortTest.cpp
+++ b/fvtest/threadtest/abortTest.cpp
@@ -129,63 +129,46 @@ runningMain(void *arg)
 }
 
 typedef struct sleep_testdata_t {
-	omrthread_monitor_t startSync;
-	omrthread_monitor_t exitSync;
-	volatile intptr_t rc;
-	volatile bool started;
+	omrthread_monitor_t sync;
 } sleep_testdata_t;
 
 TEST(ThreadAbortTest, Sleeping)
 {
 	omrthread_t t;
-	sleep_testdata_t mythread;
+	sleep_testdata_t d;
+	sleep_testdata_t *data = &d;
 
-	omrthread_monitor_init(&mythread.startSync, 0);
-	omrthread_monitor_init(&mythread.exitSync, 0);
+	omrthread_monitor_init(&data->sync, 0);
 
-	createDefaultThread(&t, sleepingMain, &mythread);
+	omrthread_monitor_enter(data->sync);
+	createDefaultThread(&t, sleepingMain, data);
+	omrthread_monitor_wait(data->sync); // startup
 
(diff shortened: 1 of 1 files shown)


fvtest/threadtest/abortTest.cpp
```diff
@@ -129,64 +129,47 @@ runningMain(void *arg)
...
-        mythread.started = false;
-	omrthread_monitor_enter(mythread.startSync);
-	while (!mythread.started) {
-		omrthread_monitor_wait(mythread.startSync);
-	}
-	omrthread_monitor_exit(mythread.startSync);
-
-	omrthread_monitor_enter(mythread.exitSync);
-	omrthread_sleep(START_DELAY);
 	omrthread_abort(t);
```
charliegracie: When we call `omrthread_abort(t)` from the main thread if `t` has not called `omrthread_sleep_interruptable()` will it see the interrupt when the call happens or will it actually get aborted?  I feel like there is still a timing hole here

fvtest/threadtest/abortTest.cpp
```diff
@@ -129,64 +129,47 @@ runningMain(void *arg)
...
-	sleep_testdata_t *testdata = (sleep_testdata_t *)arg;
-	intptr_t rc;
+	sleep_testdata_t *data = (sleep_testdata_t *)arg;
 
-	omrthread_monitor_enter(testdata->startSync);
-	testdata->started = true;
-	omrthread_monitor_notify(testdata->startSync);
-	omrthread_monitor_exit(testdata->startSync);
+	omrthread_monitor_enter(data->sync);
+	omrthread_monitor_notify(data->sync);
```
charliegracie: The enter,notify,exit pattern where no state is changed before the notify is always risky and provides the possibility for timing holes

fvtest/threadtest/abortTest.cpp
rwy0717: `omrthread_abort(t)` will set `J9THREAD_FLAG_ABORTED` on `t` and interrupt it. Abort will not actually cancel `t` (I.E. stop execution of the physical thread).

`omrthread_sleep_interruptable(timeout, 0)` will return `J9THREAD_PRIORITY_INTERRUPTED` if the thread
has been aborted, either at entry or while sleeping.

PR #1053, opened 2017-04-26
Make the travis builds faster

rwy0717:
- Shallow git clone
- Make --job parallelism
- CMake: Use the Ninja generator

It's difficult to time, when caching behaviour seems to be a little
unpredictable, but these changes seems to help.

A one-off measurement (with ccache populated):
- Make:  17m -> 15m
- CMake: 17m -> 09m
- Lint:  06m -> 02m

I've also started timing major build steps, to help evaluate build
system changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ addons:
       - llvm-3.4
       - gcc-multilib
       - g++-multilib
+      - ninja-build
 env:
   matrix:
     - BUILD_WITH_CMAKE=yes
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -21,7 +21,10 @@ os:
 language: cpp
 sudo: false
 cache: ccache # https://docs.travis-ci.com/user/caching/
+compiler: clang
```
charliegracie: This change appears to have no current impact and I worry that it would lead to hidden changes in the future.  Could you remove this?

rwy0717: Yep, it's gone.

PR #1054, opened 2017-04-26
Support pending push loads in OSR liveness

ncough:
Anchored pending push loads may be placed within an OSR point's
collection of related treetops to mitigate the impact of using
pending push loads in place of the original stack contents.

This change corrects behaviour in OSR liveness analysis to identify
these anchored loads and update the liveness accordingly. It also
extends behaviour here to remove these loads once they have served
their purpose and are not commoned elsewhere.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -55,6 +55,7 @@
 #include "optimizer/Optimizer.hpp"                    // for Optimizer
 #include "optimizer/DataFlowAnalysis.hpp"             // for TR_Liveness, etc
 #include "optimizer/StructuralAnalysis.hpp"
+#include "optimizer/TransformUtil.hpp"
 #include "optimizer/UseDefInfo.hpp"
 #include "runtime/Runtime.hpp"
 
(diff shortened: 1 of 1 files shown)


compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -921,9 +927,23 @@ TR::TreeTop *TR_OSRLiveRangeAnalysis::collectPendingPush(TR_ByteCodeInfo bci, TR
...
+         int32_t localIndex = local->getLiveLocalIndex();
+         _liveVars->set(localIndex);
+         if (comp()->getOption(TR_TraceOSR))
+            traceMsg(comp(), "+++ local index %d OSR PENDING PUSH LOAD LIVE\n", localIndex);
+
+         if (node->getFirstChild()->getReferenceCount() == 1)
+            {
+            if (comp()->getOption(TR_TraceOSR))
+               traceMsg(comp(), "----> removing node %p\n", node);
+            tt->getPrevTreeTop()->join(tt->getNextTreeTop());
```
andrewcraik: I know node pools etc are not used really at the moment, but is letting the nodes leak like this correct or is there something else that should be done - I'm not sure. Thoughts @vijaysun-omr ?

andrewcraik:
LGTM

andrewcraik:
@vijaysun-omr I think this is ready for you to look at...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -910,6 +912,11 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
    return 0;
    }
 
+// 
+// To reduce the impact of pending push temps in postExecutionOSR, the live pending push symrefs can be stored or
+// referenced in anchored loads, between the OSR point and the transition. In the event that the anchored loads
+// are not commoned, they will be removed.
+//
```
vijaysun-omr: What is the significance of the loads "not being commoned" ? I mean, when can they be commoned and when would they be left with reference count == 1 ?

ncough: If the load was anchored purely to inform this stage of liveness, it would exist with a reference count of 1. If it was instead anchored to avoid side effects of a pending push store, a reference count greater than 1 would be expected, and therefore can't be removed.

vijaysun-omr: Okay so these loads with ref count == 1 have served their purpose when you are removing them. Thanks.

PR #1058, opened 2017-04-26
Enforce reference arraycopy has type TR::Address

fjeremic:
Reference arraycopies are important when concurrent scavenge is enabled.
We need to determine at code generation time that a particular arraycopy
is indeed a reference arraycopy such that we can generate read barriers
for every array element access. As such it is important that all
arraycopy nodes that are in fact reference array copies specify the data
type of the array elements is a TR::Address.

We will use this fact in the code generator to determine whether we need
to generate read barriers for array reference loads.

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -4597,39 +4597,35 @@ OMR::Node::freeExtensionIfExists()
       }
    }
 
-
-
 TR::DataType
 OMR::Node::getArrayCopyElementType()
    {
-   TR::Compilation * comp = TR::comp();
-   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy  || self()->getOpCodeValue() == TR::arrayset, "assertion failure");
+   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy || self()->getOpCodeValue() == TR::arrayset, "getArrayCopyElementType called on an invalid node");
 
-   if (_numChildren==3 || _numChildren==4 || _numChildren==6)
+   if (_numChildren == 3)
       {
-       return (TR::DataTypes)_unionBase._extension.getExtensionPtr()->getElem<uint32_t>(_numChildren);
+      return static_cast<TR::DataTypes>(_unionBase._extension.getExtensionPtr()->getElem<uint32_t>(_numChildren));
       }
+   else
+      {
+      TR_ASSERT(_numChildren == 5, "getArrayCopyElementType called on an arraycopy node with an invalid number of children");
 
-   TR_ASSERT(comp, "Query only valid during compilation\n");
-   if (TR::Compiler->target.is64Bit() && !comp->useCompressedPointers())
-      return TR::Int64;
-   return TR::Int32;
+      return TR::Address;
+      }
    }
 
 void
 OMR::Node::setArrayCopyElementType(TR::DataType type)
    {
-   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy || self()->getOpCodeValue() == TR::arrayset, "assertion failure");
+   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy || self()->getOpCodeValue() == TR::arrayset, "setArrayCopyElementType called on an invalid node");
 
-   if (_numChildren==3 || _numChildren==4 || _numChildren==6)
+   if (_numChildren == 3)
       {
-      uint16_t numExtensionElems = _unionBase._extension.getNumElems();
-      TR_ASSERT(numExtensionElems > _numChildren, "need an extra slot to set arrayCopyElementType\n");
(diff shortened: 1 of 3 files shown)


fjeremic:
@dsouzai @vijaysun-omr could you please review?

compiler/il/OMRNode.cpp
```diff
@@ -4597,40 +4597,36 @@ OMR::Node::freeExtensionIfExists()
...
-
 TR::DataType
 OMR::Node::getArrayCopyElementType()
    {
-   TR::Compilation * comp = TR::comp();
-   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy  || self()->getOpCodeValue() == TR::arrayset, "assertion failure");
+   TR_ASSERT(self()->getOpCodeValue() == TR::arraycopy || self()->getOpCodeValue() == TR::arrayset, "getArrayCopyElementType called on an invalid node");
 
-   if (_numChildren==3 || _numChildren==4 || _numChildren==6)
+   if (_numChildren == 3)
```
vijaysun-omr: The cases for number of children being 4 or 6 are no longer in use ?

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1528,56 +1541,30 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
             }
          }
 
-      // -------------------------------------------------------------------
-      // Determine element size and array stride
-      // -------------------------------------------------------------------
-      if (isStringCompressedArrayCopy)
+      if (srcArrayLength)
```
dsouzai: Previously this was guarded with a `if(elementSize)`. Is that check not needed anymore or was it never a necessary check?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13576,8 +13576,7 @@ OMR::Z::TreeEvaluator::referenceArraycopyEvaluator(TR::Node* node, TR::CodeGener
       }
    else
       {
-      // TODO: There are cases under which the following assert will trigger. These need to be fixed and this assert needs to be re-enabled.
-      // TR_ASSERT(node->getArrayCopyElementType() == TR::Address, "Reference arraycopy element type should be a TR::Address but was '%s'", node->getArrayCopyElementType().toString());
+      TR_ASSERT_FATAL(node->getArrayCopyElementType() == TR::Address, "Reference arraycopy element type should be TR::Address but was '%s'", node->getArrayCopyElementType().toString());
```
vijaysun-omr: The prior comment suggests some cases that could trigger the assert (on downstream projects I assume). Has there been some testing done already on this front ?

compiler/il/OMRNode.cpp
fjeremic: No, we should have exactly two types of arraycopies. One with 3 children and one with 5 children.

compiler/optimizer/ValuePropagationCommon.cpp
fjeremic: To reduce reference arraycopy one of two things has to be true, either `primitiveArray1 || primitiveArray2` or `referenceArray1 || referenceArray2`. The above logic I added covers both cases and properly sets `type` and `elementSize`. Thus `elementSize` will always be non-zero at this point which is why I removed the check.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: Yes I have done proper testing on downstream projects with all assertions enabled to make sure we never hit this assert. I have also changed from `TR_ASSERT` to `TR_ASSERT_FATAL` to force a fatal crash if this assert triggers even if we compiled with assertions disabled. The reason for this choice is that missing read barriers under concurrent scavenege is a nightmare to debug as the missed read barrier could result in a stale reference to be propagated through the heap and we may not access this stale reference until much later in the application execution path.

As such debugging these types of issues is very difficult and I'd rather fatally assert if we are ever about to miss a read barrier than spend days debugging such problems in core dumps.

fjeremic:
@0dvictor could you please review as well? There are two occurrences of `getArrayCopyElementType` in the x86 codegen. I have verified that for 5 child arraycopies in both cases it is correct to return TR::Address as we properly check for this and use the size of reference field from the Object Model helper as the stride. Please independently verify this is the case so I'm not regressing anything here.

compiler/il/OMRNode.cpp
0dvictor: It's nice to see that this part is cleaned.

0dvictor:
The changes look good to me. For `getArrayCopyElementType`, it is expected to perform the same on X86.

fjeremic:
@vijaysun-omr any other concerns? I have rebased and fixed up minor aesthetic issues (indentation, extra space).

vijaysun-omr:
Thanks for the explanation. Merging.

PR #1059, opened 2017-04-26
Addition to the include path to support future work.

jdekonin:
Signed-off-by: Joe Dekoning <joe_dekoning@ca.ibm.com>


charliegracie:
@jdekonin can you update the first line of your commit message to limit it to 50 characters or less and also remove the `.`.  Thanks

PR #1060, opened 2017-04-26
Disabling Inlined Array Copy on Power

R-Santhir:
When the array length is not known at compile time, a sequence of close
to 300 instructions is needed to cover all cases of the array copy.
Inlining this is very costly in terms of performance for small arrays.
Given that most of the arrays seen are a few dozen bytes, it is not
worth inlining. When the length of the array is known at compile time,
inlining can still occur.

### DIFF
--- a/compiler/p/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/p/codegen/OMRTreeEvaluator.cpp
@@ -4576,7 +4576,7 @@ TR::Register *OMR::Power::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::
    bool  supportsVSX = (processor >= TR_PPCp8) && !disableVSXArrayCopy && TR::Compiler->target.cpu.getPPCSupportsVSX();
 
    static bool disableLEArrayCopyHelper  = (feGetEnv("TR_disableLEArrayCopyHelper") != NULL);
-   static bool disableVSXArrayCopyInlining = (feGetEnv("TR_disableVSXArrayCopyInlining") != NULL);
+   static bool disableVSXArrayCopyInlining = (feGetEnv("TR_enableVSXArrayCopyInlining") == NULL); // Disabling due to a performance regression
 
    bool  supportsLEArrayCopy = !disableLEArrayCopyHelper && TR::Compiler->target.cpu.isLittleEndian() && TR::Compiler->target.cpu.hasFPU();
 


compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4577,6 +4577,7 @@ TR::Register *OMR::Power::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::
 
    static bool disableLEArrayCopyHelper  = (feGetEnv("TR_disableLEArrayCopyHelper") != NULL);
    static bool disableVSXArrayCopyInlining = (feGetEnv("TR_disableVSXArrayCopyInlining") != NULL);
+   disableVSXArrayCopyInlining = true; // Disabling due to a performance regression
```
ymanton: If you're going to leave the code in place and just disable it perhaps

```
static bool disableVSXArrayCopyInlining = (feGetEnv("TR_enableVSXArrayCopyInlining") == NULL);
```

would be better.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4576,7 +4576,7 @@ TR::Register *OMR::Power::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::
    bool  supportsVSX = (processor >= TR_PPCp8) && !disableVSXArrayCopy && TR::Compiler->target.cpu.getPPCSupportsVSX();
 
    static bool disableLEArrayCopyHelper  = (feGetEnv("TR_disableLEArrayCopyHelper") != NULL);
-   static bool disableVSXArrayCopyInlining = (feGetEnv("TR_disableVSXArrayCopyInlining") != NULL);
+   static bool disableVSXArrayCopyInlining = (feGetEnv("TR_disableVSXArrayCopyInlining") == NULL); // Disabling due to a performance regression
```
vijaysun-omr: Is it better to rename this env var ? As it stands, the env var is named TR_disableVSXArrayCopyInlining but if someone sets it, then it enables the arraycopy inlining (which is a bit odd).

vijaysun-omr:
I also agree with Filip's earlier comment about the header. 

Is it possible to explain at a high level why this regressed performance ? Just curious since that is'nt really mentioned anywhere.

zl-wang:
Yes, the commit message and header need to change.  Compile-time known length of copy is still inlined.

Blind inlining of arrayCopy is a bad idea on POWER, since it needs ~300 instructions to cover all cases of copy.  Just doing this inlining will involve shuffling around 10s kilobytes of memory.  Most likely you cannot recover this overhead, given java arraycopy is usually of a (few) dozen bytes without hitting the code 1000s times.  The code bloat (and effect on iCache) is not even considered here.

The straight trade-off is the arrayCopy helper trampoline overhead vs. code-bloat.   If the copy-length is small, you unlikely can recover the compilation overhead alone.  If the copy-length is big, the trampoline overhead doesn't matter.

R-Santhir:
I have changed the env var name as per @vijaysun-omr 's suggestion, The commit header and message have been changed as well to address @fjeremic 's concern with @zl-wang 's description.

gita-omr:

PR #1019, opened 2017-04-18
correct concurrent sweepStates after heap resize

LinHu2016:
- update concurrent sweepStates after heap resize at end of global gc
to avoid losting free memory and wrong memory usgage reporting
 - update verbosegc schema to allow warning in <mem-info>

Signed-off-by: Lin Hu <linhu@ca.ibm.com>

PR #1062, opened 2017-04-27
Revert "correct concurrent sweepStates after heap resize"

charliegracie:
Reverts eclipse/omr#1019

This change is causing issues when concurrent sweep is happening.  I am reverting the PR until the concurrent sweep issues can be resolved


charliegracie:
Merging as this is just reverting code.

Issue #1066, opened 2017-04-27
Enable AutoSIMD support on x86


IBMJimmyk:
This is related to Issue #525 and Issue #862 

Issue #862 implements several SIMD evaluators on x86. getSupportsOpCodeForAutoSIMD in compiler/x/codegen/OMRCodeGenerator.cpp needs to be updated to reflect the newly implemented evaluators.

When implementing Issue #525, the call to setSupportsAutoSIMD was intentionally left out of initialize in OMRCodeGenerator.cpp. This was because AutoSIMD was not possible at the time due to a lack of vector evaluators. Since x86 will now have some support for AutoSIMD, the call needs to be added in.

Everything to do with isFPRUsedAsVRF added in Issue #525 needs to be removed. I made a mistake when looking at how prepareRegistersForAssignment worked in compiler/codegen/CodeGenRA.cpp. The code related to isFPRUsedAsVRF does not help in any way and in some cases causes registers to not be assigned when they should be. isFPRUsedAsVRF was introduced recently and no other code that I didn't write uses it so I think it's best to just get rid of it.

PR #1067, opened 2017-04-27
Enable AutoSIMD support on x86

IBMJimmyk:
Updated getSupportsOpCodeForAutoSIMD in x/codegen/OMRCodeGenerator.cpp to
reflect the recently implemented vector evaluators. Instead of returning
false in all cases, getSupportsOpCodeForAutoSIMD will now return true for
loads, stores and vector splats for int32, int64, float and double data
and also double type add and multiply. All of these operations are
supported in the new vector evaluators.

A call to setSupportsAutoSIMD is now made inside initialize in
OMRCodeGenerator.cpp to indicate that autoSIMD is now supported. This call
is behind a check for SSE2 support.

The function isFPRUsedAsVRF and any references to it were removed. It was
added to prepareRegistersForAssignment in error. It is not needed for
AutoSIMD and sometimes caused registers to not be assigned properly. It
was only used in one place, so the entire fuction is just being removed.

Issue: #1066

### DIFF
--- a/compiler/codegen/CodeGenRA.cpp
+++ b/compiler/codegen/CodeGenRA.cpp
@@ -409,15 +409,6 @@ OMR::CodeGenerator::prepareRegistersForAssignment()
          registerCursor->setFutureUseCount(registerCursor->getTotalUseCount());
       km = registerCursor->getKindAsMask();
 
-      /*
-       * Sometimes TR_VRF virtual registers need to be assigned even when there are no TR_VRF real registers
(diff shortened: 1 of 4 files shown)


IBMJimmyk:
The changes in this PR have already been tested for performance impact. These changes relied on the changes in Issue #862 being checked in first so I couldn't check them in until now.

I think @andrewcraik and @gita-omr wanted to look at these changes as well.

andrewcraik:
@gita-omr can you please review - once you are happy I can review and we can try to get this in.

gita-omr:
LGTM

andrewcraik:
LGTM

PR #1069, opened 2017-04-28
Releases managed heap free pages without resizing

Param-S:
### Overview
Systems where the runtime wants to relinquish the allocated free resources like
memory, cpu & etc to make efficient/optimal use of those resources. For example,
On cloud, the processes are charged based on GB/hours for memory usage. If more
than one runtime runs on the same account and if runtimes can explicitly share
the memory based on the usage, it will be beneficial. To support such scenarios,
this PR adds a method onto the existing interface "systemGarbageCollect" to
release the managed heap's free memory pages without resizing the heap.

### Design
Updates the existing interface "systemGarbageCollect" to recognise the new GC
code J9MMCONSTANT_EXPLICIT_GC_IDLE_GC and perform Global GC and release managed
heap free memory pages when it is called with new GC code.

Invocation of Global GC is determined by the allocation made to the memory space
since previous global GC cycle. If the allocation is lesser than the threshold
value, then Global GC is not called and directly continues with the process of
managing heap free memory pages. By default, value/percentage for the allocation
threshold is set to 20. The default value can be customised through the new
attribute "gcOnIdleRatio"(which is added into MM_GCExtensionsBase).

The managed heap free memory pages are released by walking through the entries
of the memory pool and releases the pages associated with the memory using
system interface madvice(on Linux platform). By default, the complete managed
heap free memory pages get released. The default value can be customised using
the new attribute "idleMinimumFree"(which is added into MM_GCExtensionsBase).
 
### PSECUDO CODE
```
ENTER systemGarbageCollect
IF GCCode equals J9MMCONSTANT_EXPLICIT_GC_IDLE_GC
THEN
    IF allocation since last global GC cycle is greater than 20 %
    THEN
        CALL Global GC 
    ENDIF
    WALK through free memory pool and decommits pages of free memory
ENDIF
EXIT systemGarbageCollect
```

#### Updates to existing structures/class:
MM_GCExtensionsBase {
....
idleMinimumFree = 0 default, release all free memory pages
lastGCFreeBytes = to keep record of free memory after last global GC
gcOnIdleRatio = 20 default, < allocation since last global gc, perform GC 
compactOnIdle = enforce compaction as part the global GC
}

New Virtual function is added to perform releasing memory pages
MM_MemorySpace, MM_MemorySubSpace, MM_MemoryPool{
    releaseFreeMemoryPages();
}

New GC code is added & classified accordingly.
J9MMCONSTANT_EXPLICIT_GC_IDLE_GC - explicit GC, aggressive GC.

#### Updates to verbose GC stanza/tag <heap-resize:
Added a new type "releases free pages" into existing types "expand, contract, &
etc) of <heap-resize tag. 

### Example
If runtime invokes systemGarbageCollect with GC code = J9MMCONSTANT_EXPLICIT_GC_IDLE_GC, we see a similar output in verbose GC. 
```
<heap-resize id="3279" type="release free pages" space="tenure" amount="222228480" count="1" timems="3.202" reason="idle" timestamp="2017-05-04T16:21:06.393" />
```
### Scope
This PR is currently scoped to only Generational Concurrent GC policy and Linux
x86(32bit + 64bit) platform.

### Future
This PR will be implemented for other GC policies and supported platform


charliegracie:
Hi @Param-S,

Can you provide more details in this PR about what it is trying to accomplish?  Why is this code all under a build flag?  It looks like it is fairly generic and could be determined by a runtime flag without any extra cost.

You will need to rebase your changes on the latest version of the OMR code so that the merge conflict is resolved.

You will also need to follow the instructions in [CONTRIBUTING.md](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md).  Please update your commit message to follow all of the guidelines and you will need to sign an Eclipse ECA before the change can be merged.

gc/base/MemoryPoolAddressOrderedListBase.cpp
```diff
@@ -168,3 +168,34 @@ MM_MemoryPoolAddressOrderedListBase::abandonMemoryInPool(MM_EnvironmentBase* env
...
+uintptr_t
+MM_MemoryPoolAddressOrderedListBase::releaseMemoryPoolFreeEntryMemoryPages(MM_EnvironmentBase* env, MM_HeapLinkedFreeHeader* freeEntry)
+{
+	uintptr_t freedMemory = 0;
+	MM_HeapLinkedFreeHeader* currentFreeEntry = freeEntry;
+	uintptr_t pageSize = env->getExtensions()->heap->getPageSize();
+	while (NULL != currentFreeEntry) {
+		if (currentFreeEntry->getSize() >= ((2 * pageSize) + sizeof(MM_HeapLinkedFreeHeader))) {
+			/* release/decommit memory after Header */
+			uint8_t* addressBase = (uint8_t*)MM_Math::roundToCeiling(pageSize, (uintptr_t)((uint8_t*)currentFreeEntry + sizeof(MM_HeapLinkedFreeHeader)));
```
dmitripivkine: Would it be cleaner to have "uintptr_t addressBase" and cast it as a (void *) in decommitMemory() call?

gc/base/MemoryPoolAddressOrderedListBase.cpp
```diff
@@ -168,3 +168,34 @@ MM_MemoryPoolAddressOrderedListBase::abandonMemoryInPool(MM_EnvironmentBase* env
...
+			uintptr_t totalFreePagesCount = (currentFreeEntry->getSize() - (uintptr_t)(addressBase - (uint8_t*)currentFreeEntry)) / pageSize;
```
dmitripivkine: again, there is no need to cast to uint8_t pointer, cast straight to uintptr_t

gc/base/MemorySubSpace.cpp
```diff
@@ -919,16 +919,23 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
-
 		env->acquireExclusiveVMAccessForGC(_collector);
-		reportSystemGCStart(env, gcCode);
-
-		/* system GCs are accounted into "user" time in GC/total time ratio calculation */
-		_collector->garbageCollect(env, this, NULL, gcCode, NULL, NULL, NULL);
-
-		reportSystemGCEnd(env);
+		#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+		if (isSystemGCAllowed(env, gcCode)) {
```
dmitripivkine: Bad function name. I would not even create this function, just put plain code there

gc/base/MemorySubSpace.cpp
```diff
@@ -919,16 +919,23 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
+		#endif
+			/* TODO: This is bogus for multiple memory spaces - should ask the space, not the heap */
+			_extensions->heap->getResizeStats()->setFreeBytesAtSystemGCStart(getApproximateActiveFreeMemorySize());
```
dmitripivkine: This is change of behavior. This call use to be issued before taking Exclusive. Aleks, please comment

gc/base/MemorySubSpace.cpp
```diff
@@ -919,16 +919,23 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
+
+			reportSystemGCStart(env, gcCode);
+			/* system GCs are accounted into "user" time in GC/total time ratio calculation */
+			_collector->garbageCollect(env, this, NULL, gcCode, NULL, NULL, NULL);
+			reportSystemGCEnd(env);
+		#if defined(OMR_GC_IDLE_HEAP_MANAGER)
```
dmitripivkine: # operators should start in column 1 ... here and below

gc/base/MemorySubSpace.cpp
```diff
@@ -1929,3 +1936,28 @@ MM_MemorySubSpace::wasContractedThisGC(uintptr_t gcCount)
...
+MM_MemorySubSpace::isSystemGCAllowed(MM_EnvironmentBase* env, uint32_t gcCode)
+{
+	bool invokeGC = true;
+	if (J9MMCONSTANT_IMPLICIT_GC_IDLE_GC == gcCode) {
+		MM_MemorySpace* defaultMemorySpace = _extensions->heap->getDefaultMemorySpace();
+		uintptr_t freeMemorySize = defaultMemorySpace->getApproximateActiveFreeMemorySize(MEMORY_TYPE_OLD);
+		uintptr_t actvMemorySize = defaultMemorySpace->getActiveMemorySize(MEMORY_TYPE_OLD);
+		uintptr_t previousMemorySize = 0 < _extensions->lastGCFreeBytes ? _extensions->lastGCFreeBytes : actvMemorySize;
+
+		if ((0 < freeMemorySize) && (20 > (((previousMemorySize - freeMemorySize) * 100) / actvMemorySize))) {
```
dmitripivkine: use \# define for "20" with appropriate name

gc/base/MemorySubSpace.cpp
```diff
@@ -919,16 +919,23 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
-		_extensions->heap->getResizeStats()->setFreeBytesAtSystemGCStart(getApproximateActiveFreeMemorySize());
-
 		env->acquireExclusiveVMAccessForGC(_collector);
-		reportSystemGCStart(env, gcCode);
-
-		/* system GCs are accounted into "user" time in GC/total time ratio calculation */
-		_collector->garbageCollect(env, this, NULL, gcCode, NULL, NULL, NULL);
-
-		reportSystemGCEnd(env);
+		#if defined(OMR_GC_IDLE_HEAP_MANAGER)
```
dmitripivkine: \# operators should start in column 1

gc/base/MemorySubSpace.cpp
```diff
@@ -919,16 +919,23 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
+		#endif
+			/* TODO: This is bogus for multiple memory spaces - should ask the space, not the heap */
+			_extensions->heap->getResizeStats()->setFreeBytesAtSystemGCStart(getApproximateActiveFreeMemorySize());
+
+			reportSystemGCStart(env, gcCode);
+			/* system GCs are accounted into "user" time in GC/total time ratio calculation */
+			_collector->garbageCollect(env, this, NULL, gcCode, NULL, NULL, NULL);
+			reportSystemGCEnd(env);
+		#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+		}
```
dmitripivkine: if bool would be generally defined there is no need to have this bracket under \#if defined(OMR_GC_IDLE_HEAP_MANAGER)

gc/base/MemorySubSpace.cpp
amicic: I understand that is moved down so that we do not take this snapshot if we do not end up doing the system GC.  bigger concern is that 'is allowed' logic is AFTER acquiring exclusive access - seems like a wasted exercise if sysGC is not performed. Could we invoke 'is allowed' logice prior to invoking top level sysGC? 

I did not look into the rest of the code yet, but what happens if we sysGC is not 'allowed' - do we still take exclusive access to walk&decommit Tenure free pool?

amicic: Ah I see the answer to my question... releaseFreeMemoryPages is invoked just after. 

I'll think a bit what's the best way... But for now, I'll say this: walking&decommit a pool should be safe by just holding tenure allocate lock - exclusive access is not really needed. So, one way to do it is 1) to pull out the 'is allowed' logic out of sysGC call. 2) if sysGC call is done, invoke releaseFreeMemoryPages somwhere in Global GC code, toward the end. 3) invoke releaseFreeMemoryPages  in else clause of 'is allowed', outside of sysGC, but add pool locking the the release logic. Dmitiri would you find this resonable?

dmitripivkine: Yes, looks like a reasonable proposal

gc/base/MemoryPoolAddressOrderedListBase.cpp
```diff
@@ -168,3 +168,34 @@ MM_MemoryPoolAddressOrderedListBase::abandonMemoryInPool(MM_EnvironmentBase* env
...
+			if (1 < totalFreePagesCount) {
+				uintptr_t commitPagesCount = (0 < _extensions->idleMinimumFree) ? (totalFreePagesCount * _extensions->idleMinimumFree / 100) : 0;
```
dmitripivkine: Suggestion: I would try to avoid complicated "arithmetic if" operators. It makes code hard to read. 
Example:
uintptr_t commitPagesCount = 0;
if (0 <_extensions->idleMinimumFree) {
   commitPagesCount = totalFreePagesCount * _extensions->idleMinimumFree / 100;
}
is much easier to read. Compiler would generate the same code any way.

gc/base/MemorySubSpace.cpp
amicic: Since we are supposedly idling, exclusive access acquire should be farily cheap to obtain, so we not saving that much by not doing that 'exercise' after all. 

If if end up just acquiring pool lock, i'll add that 
1) the thread must have vm access as a min (to prevent any other party acquiring exclusive and messing up with heap (like very GC just from another thread).
2) i'd like to see verbose GC stanza reporting this decommit if running standalone (probalby even if it's run within sysGC, although that would be implied from sysGC reason'). But I want to see durationms of the operation, in any case. This would not be really a gc-op, but more like info (see for example classunload-info stanza).

gc/base/gcutils.cpp
```diff
@@ -239,6 +239,10 @@ getSystemGCReasonAsString(uint32_t gcCode)
 		return "native out of memory";
 	case J9MMCONSTANT_EXPLICIT_GC_RASDUMP_COMPACT:
 		return "rasdump";
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+	case J9MMCONSTANT_IMPLICIT_GC_IDLE_GC:
+		return "vm idle";
+#endif
```
amicic: Rename it to EXPLICIT type. Implicit are ones caused by AF.

gc/base/MemoryPoolAddressOrderedListBase.cpp
amicic: go with a shorted name: releaseFreeEntryMemoryPages

amicic: Seems to me that this line is redunant:
if (currentFreeEntry->getSize() >= ((2 * pageSize) + sizeof(MM_HeapLinkedFreeHeader))) {

it will be later verified with totalFreePagesCount.

gc/base/MemoryPoolAddressOrderedListBase.cpp
amicic: Shouldn't be 0 <  totalFreePagesCount? You could decommit even that one single page.

gc/base/MemoryPoolAddressOrderedListBase.cpp
```diff
@@ -168,3 +168,34 @@ MM_MemoryPoolAddressOrderedListBase::abandonMemoryInPool(MM_EnvironmentBase* env
...
+						freedMemory += decommitPagesCount * pageSize;
+					}
+				}
+			}
+		}
+		currentFreeEntry = currentFreeEntry->getNext();
+	}
+	return freedMemory;
+}
+#endif
```
amicic: To be consistent with the method name, the return variable should be releasedMemory.

gc/base/MemorySubSpace.cpp
amicic: The threshould should be a configurable parameter, via command line option. 

The logic is not right to me - I believe we discussed this. It should be relative to historic free memory, not relative to active. The logic is: if we only have <60 or so % of free memory (currently) available relative to what we historically have had, just after a global GC, than we should do an explicit sysGC to regain even those extra 40%.

The current criteria does not guarantee that any new free memory will be made available. For example, if we historically free up 30 and currently we are also 30 you'll get (30 - 30) < 20, and sysGC will be triggered, but actually 0 additional memory will be freed.

The current criteria has its own benefits, though. In an opposite example, if we already have very high hisstorical free memory, like 80&, and we are currently at 59, it will prevent a sysGC. 

A compromise is perhaps, to keep both...

I also suggest that you maintain historical weighted average of free memory. For example, avg = (50% of current + 50% of prev average) 
.

gc/base/MemorySubSpace.cpp
```diff
@@ -1929,3 +1936,28 @@ MM_MemorySubSpace::wasContractedThisGC(uintptr_t gcCount)
...
+	bool invokeGC = true;
+	if (J9MMCONSTANT_IMPLICIT_GC_IDLE_GC == gcCode) {
+		MM_MemorySpace* defaultMemorySpace = _extensions->heap->getDefaultMemorySpace();
+		uintptr_t freeMemorySize = defaultMemorySpace->getApproximateActiveFreeMemorySize(MEMORY_TYPE_OLD);
+		uintptr_t actvMemorySize = defaultMemorySpace->getActiveMemorySize(MEMORY_TYPE_OLD);
+		uintptr_t previousMemorySize = 0 < _extensions->lastGCFreeBytes ? _extensions->lastGCFreeBytes : actvMemorySize;
+
+		if ((0 < freeMemorySize) && (20 > (((previousMemorySize - freeMemorySize) * 100) / actvMemorySize))) {
+			invokeGC = false;
+		}
```
amicic: I a previous commented on the criteria, without realizing that that invokeGC is being set to false (for some reason I assumed true). Now the logic makes more sense, so  you can disregard my previous comment in the original form (which have been updated meanwhile)

gc/base/GCExtensionsBase.hpp
```diff
@@ -706,6 +706,13 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
 
 	uintptr_t darkMatterSampleRate;/**< the weight of darkMatterSample for standard gc, default:32, if the weight = 0, disable darkMatterSampling */
 
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+	uintptr_t idleMinimumFree;   /** < percentage of free heap to be retained as committed, default=0 for gencon, complete tenture free memory will be decommitted */
+	uintptr_t lastGCFreeBytes;  /** < records the free memory size from last Global GC cycle */
+	uint32_t gcOnIdleRatio; /**< the percentage of allocation since the last GC allocation determines the invocation of global GC, default global GC is invoked if allocation is > 20% */
```
dmitripivkine: Is it a reason to have gcOnIdleRatio explicitly 32-bit?

gc/base/MemorySubSpace.cpp
```diff
@@ -919,17 +919,51 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
 	if (_collector && _usesGlobalCollector) {
-		/* TODO: This is bogus for multiple memory spaces - should ask the space, not the heap */
-		_extensions->heap->getResizeStats()->setFreeBytesAtSystemGCStart(getApproximateActiveFreeMemorySize());
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+		bool invokeGC = true;
+		if (J9MMCONSTANT_EXPLICIT_GC_IDLE_GC == gcCode ) {
+			MM_MemorySpace* defaultMemorySpace = _extensions->heap->getDefaultMemorySpace();
+			uintptr_t freeMemorySize = defaultMemorySpace->getApproximateActiveFreeMemorySize(MEMORY_TYPE_OLD);
+			uintptr_t actvMemorySize = defaultMemorySpace->getActiveMemorySize(MEMORY_TYPE_OLD);
+			uintptr_t previousMemorySize;
```
dmitripivkine: local variable should be initialized... Suggestion: There is might be a case when using of arithmetic if is reasonable

gc/base/MemorySubSpace.cpp
```diff
@@ -919,17 +919,51 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
 	}
 
 	if (_collector && _usesGlobalCollector) {
-		/* TODO: This is bogus for multiple memory spaces - should ask the space, not the heap */
-		_extensions->heap->getResizeStats()->setFreeBytesAtSystemGCStart(getApproximateActiveFreeMemorySize());
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+		bool invokeGC = true;
```
dmitripivkine: I would move invokeGC outside of defined(OMR_GC_IDLE_HEAP_MANAGER). In this case there is no need to have if(invokeGC) and correspondent brackets under ifdef

gc/base/MemorySubSpace.cpp
```diff
@@ -919,17 +919,51 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
+				previousMemorySize = actvMemorySize;
+			}
 
-		env->acquireExclusiveVMAccessForGC(_collector);
-		reportSystemGCStart(env, gcCode);
+			if ((0 < freeMemorySize) && (_extensions->gcOnIdleRatio > (((previousMemorySize - freeMemorySize) * 100) / actvMemorySize))) {
+				invokeGC = false;
+			}
+		}
+		if (invokeGC) {
```
dmitripivkine: if you move invokeGC out of ifdef you will be able to move out this line too

gc/base/MemorySubSpace.cpp
```diff
@@ -919,17 +919,51 @@ MM_MemorySubSpace::systemGarbageCollect(MM_EnvironmentBase* env, uint32_t gcCode
...
 
-		reportSystemGCEnd(env);
-		env->releaseExclusiveVMAccessForGC();
+			/* system GCs are accounted into "user" time in GC/total time ratio calculation */
+			_collector->garbageCollect(env, this, NULL, gcCode, NULL, NULL, NULL);
+
+			reportSystemGCEnd(env);
+			env->releaseExclusiveVMAccessForGC();
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+		}
```
dmitripivkine: ...and this bracket too

gc/verbose/VerboseHandlerOutput.cpp
```diff
@@ -950,3 +950,23 @@ verboseHandlerHeapResize(J9HookInterface** hook, uintptr_t eventNum, void* event
...
+{
+	MM_ReleaseFreeMemPagesEvent* event = (MM_ReleaseFreeMemPagesEvent*)eventData;
+	MM_VerboseManager* manager = getManager();
+	MM_VerboseWriterChain* writer = manager->getWriterChain();
+	MM_EnvironmentBase* env = MM_EnvironmentBase::getEnvironment(event->currentThread);
+	OMRPORT_ACCESS_FROM_ENVIRONMENT(env);
+	char tagTemplate[200];
+	getTagTemplate(tagTemplate, sizeof(tagTemplate), manager->getIdAndIncrement(), omrtime_current_time_millis());
+	enterAtomicReportingBlock();
+	writer->formatAndOutput(env, 0, "<heap-decommit %s>", tagTemplate);
```
dmitripivkine: I would avoid word decommit here and below

gc/base/GCExtensionsBase.hpp
Param-S: value range of the variable "gcOnIdleRatio" is 0 - 100. It doesn't need 8 bytes (uintptr_t).  
The next variable "compactOnIdle) is of type bool. 
So, we define the variable gcOnIdleRatio as 4 bytes, both the variables gcOnIdleRation + compactOnIdle can be accomdated in same 64 bit word.

gc/base/MemorySubSpace.cpp
Param-S: I consciously left it uninitialised as I previously received a comment related to initialisation.

dmitripivkine 8 days ago
again, no need to set variable to 0 if next line set it's real value. For the following scenario.

+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
 +uintptr_t
 +MM_MemoryPoolLargeObjects::releaseFreeMemoryPages(MM_EnvironmentBase* env)
 +{
 +	uintptr_t memReleased = 0;

did i understand it correctly? If not, apologise, please correct me.

dmitripivkine: Why not to do:
uintptr_t previousMemorySize  = actvMemorySize;
if (0 < _extensions->lastGCFreeBytes) {
  previousMemorySize = _extensions->lastGCFreeBytes;
}

gc/base/GCExtensionsBase.hpp
dmitripivkine: It is ok for me... however keep in mind that if this variable will be involved in operations together with uintptr_t variables compiler will need to do cast 32-bit to 64-bit and back. Sure, most of the modern processors have 32-bit instructions so probably cost nothing. Also 4 bytes you are trying to save will be lost any way most likely. Also in some cases (probably not in this) it might be a reason for warning in compilation time if explicit cast is missed.

gc/verbose/VerboseHandlerOutput.cpp
Param-S: **Keyword for new verbose tag:**
<heap-decommit id="4283" timestamp="2017-05-03T04:19:52.563">
  <decommit decommitBytes="411484160" durationms="6.539" />
</heap-decommit>
**>>>to<<<**
<heap-pages id="4283" timestamp="2017-05-03T04:19:52.563">
  <pages releasedBytes="411484160" durationms="6.539" />
</heap-pages>
is it fine?

gc/base/MemorySubSpace.cpp
Param-S: agree. corrected it

gc/base/MemorySubSpace.cpp
Param-S: agree. rearranged compiler flag.

Param-S:
Hi @charliegracie,
I have updated PR with more details.  I have also signed the ECA.  
Thanks
Param

gc/base/MemoryPoolAddressOrderedList.cpp
```diff
@@ -1648,3 +1648,14 @@ MM_MemoryPoolAddressOrderedList::appendCollectorLargeAllocateStats()
 	_largeObjectCollectorAllocateStats = _largeObjectAllocateStats;
 }
 
+#if defined(OMR_GC_IDLE_HEAP_MANAGER)
+uintptr_t
+MM_MemoryPoolAddressOrderedList::releaseFreeMemoryPages(MM_EnvironmentBase* env)
+{
+	uintptr_t releasedBytes;
```
charliegracie: Please initialize this local variable.

gc/base/GCExtensionsBase.hpp
charliegracie: I would prefer to have this variable become a `uintptr_t`. This will just future proof the code and remove the requirement that everyone has to continue to verify that the value has not become too large for a 32 bit variable.

gc/base/GCExtensionsBase.hpp
```diff
@@ -706,6 +706,13 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
...
+	bool compactOnIdle; /**< Forces compaction if global GC executed while VM Runtime State set to IDLE, default is false */
```
dmitripivkine: I don't see any functional use for this variable

gc/base/MemoryPoolAddressOrderedList.cpp
Param-S: Updated code.

Param-S:
@charliegracie updated the code as per your last comments.

PR #1072, opened 2017-05-01
Rename guard merging checks to not specify HCR

ncough:
OSR guards now support guard merging, under the same conditions
as HCR guards. This renames checks required to merge these
guards to not specify HCR.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -494,7 +494,7 @@ class OMR_EXTENSIBLE CodeGenerator
    //
    bool supports32bitAiadd() {return true;}  // no virt, default
    bool supportsMarshallingUnmarshallingIntrinsics() {return true;} // no virt, default
-   bool supportsMergingOfHCRGuards() {return false;} // no virt, default
+   bool supportsMergingGuards() {return false;} // no virt, default
 
    /** \brief
     *     Determines whether concurrent scavenge of objects during garbage collection is enabled.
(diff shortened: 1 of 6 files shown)


compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -494,7 +494,7 @@ class OMR_EXTENSIBLE CodeGenerator
    //
    bool supports32bitAiadd() {return true;}  // no virt, default
    bool supportsMarshallingUnmarshallingIntrinsics() {return true;} // no virt, default
-   bool supportsMergingOfHCRGuards() {return false;} // no virt, default
+   bool supportsMergingGuards() {return false;} // no virt, default
```
fjeremic: @mgaudet are these "no virt, default" comments still useful? This should be obvious given we are in the base OMRCodeGenerator.

andrewcraik:
The comment is not new and matches its friends. The merging query is just being given a more applicable name so I think this should be fine to merge. We can handle the comment later, downstream projects want to switch the API name sooner rather than later to avoid confusion. LGTM.

fjeremic:
> The comment is not new and matches its friends. The merging query is just being given a more applicable name so I think this should be fine to merge. We can handle the comment later, downstream projects want to switch the API name sooner rather than later to avoid confusion.

I completely agree. My intention was not to block the merging of this PR, hence why I left a comment review as I'm curious to whether the "no virt, default" comments are still relevant and whether the should be removed (outside the scope of this PR of course).

mgaudet:
Those comments can be stripped. They were inserted as notes during refactoring.

PR #1074, opened 2017-05-01
Enable Array Copy under Concurrent Scavenge

dsouzai:
This commit enables the previously disabled (under concurrent scavenge)
transformation of the array copy call into the arraycopy node.

In order to handle a current scavenge cycle, the compiler generates code
to check a runtime flag. Execution conditionally branches to an Out Of Line
sequence which contains the necessary guarded load, shift, and store
instructions. This code is only active under `#ifdef J9_PROJECT_SPECIFIC`.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -13510,6 +13510,56 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
    byteSrcReg = cg->gprClobberEvaluate(byteSrcNode);
    byteDstReg = cg->gprClobberEvaluate(byteDstNode);
 
+   bool alreadyEvaluatedByteLenNode = false;
+
+#ifdef J9_PROJECT_SPECIFIC
+   TR::LabelSymbol* mergeLabel;
+   bool mustGenerateOOLGuardedLoadPath = cg->isConcurrentScavengeEnabled() &&
+                                         node->getArrayCopyElementType() == TR::Address;
+
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      // The OOL path needs the byte length for the guarded load, compress, and store loop.
+      // However, the main line fast path also needs the byte length. As such, the byteLenNode
+      // must be evaluated outside of the OOL ICF region.
+      byteLenReg = cg->gprClobberEvaluate(byteLenNode);
+      alreadyEvaluatedByteLenNode = true;
+
+      mergeLabel = generateLabelSymbol(cg);
+      TR::LabelSymbol* slowPathLabel = generateLabelSymbol(cg);
+
+      TR_ASSERT_FATAL(J9_PRIVATE_FLAGS_CONCURRENT_SCAVENGER_ACTIVE == 0x20000,
+                 "GSCS: The OOL sequence branch is dependant on the flag being 0x20000");
+      TR_ASSERT_FATAL(TR::Compiler->target.is64Bit(),
+                 "GSCS: Guarded Load OOL Path only defined in 64 bit mode");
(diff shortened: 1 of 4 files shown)


dsouzai:
@fjeremic Could you please review.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13522,7 +13570,11 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
          {
          bool evaluatedLengthMinusOne = false;
 
-         byteLenReg = cg->evaluateLengthMinusOneForMemoryOps(byteLenNode, true, evaluatedLengthMinusOne);
+         // Don't evaluate length minus one if byteLenNode was already evaluated
+         if (!alreadyEvaluatedByteLenNode)
```
fjeremic: Lets stay consistent with `evaluatedLengthMinusOne` and name this `evaluatedByteLenNode`.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13532,18 +13584,30 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
       vtOp.generate(byteDstReg, byteSrcReg, strideRegister);
       vtOp.cleanUpReg();
 
       cg->stopUsingRegister(strideRegister);
       }
 
+#ifdef J9_PROJECT_SPECIFIC
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      // merge point
```
fjeremic: Unnecessary comment. `mergeLabel` already implies this.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13510,6 +13510,54 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
+   bool alreadyEvaluatedByteLenNode = false;
+
+#ifdef J9_PROJECT_SPECIFIC
+   TR::LabelSymbol* mergeLabel;
+   bool mustGenerateOOLGuardedLoadPath = cg->isConcurrentScavengeEnabled() &&
+                                         node->getArrayCopyElementType() == TR::Address;
+
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      // We must evaluate this before switching to the OOL path
```
fjeremic: Perhaps explain why we need to evaluate this. You should mention that the OOL path needs the byte length for the guarded load-compress-store loop and as such it must be evaluate outside of the OOL ICF region as this register is also used in the fast path.

fjeremic:
A note to commiters: While we are introducing another use of `J9_PROJECT_SPECIFIC` here please don't be alarmed. I have in my plan to push this code up into the J9 project and decouple the arraycopy implementation. This work will happen in the next weeks.

fjeremic:
!1076 is a prerequisite to this PR. This PR relies on the fixes from !1076. Please do not merge until the prerequisite PR has been merged.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: Done.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: Done.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: Done.

fjeremic:
LGTM.

dsouzai:
@vijaysun-omr #1076 is merged so I think it should be ok to merge this PR if everything looks good to you as well.

PR #1076, opened 2017-05-01
Correctly block/unblock even/odd register pairs

fjeremic:
When assigning register pairs we need to consider two cases of either
the even or odd register in the register pair being spilled. In such a
case we must reverse the spill state of the respective register, however
we must be careful here not to cause changes to the register state of
the other register in the register pair.

For example if we are attempting to assign a pair of registers (A,B)
where A has been spilled onto the stack and B is currently in a register
we must be careful to not evict B from the register state when reversing
the spill state of A.

To avoid such issues we block register B until the spill state of
register A has been reversed. We must also unblock register B following
this action.

This commit addresses a problem in which we do not correctly unblock
register B, and thus the register stays blocked for the remainder of the
register assignment.

### DIFF
--- a/compiler/z/codegen/OMRMachine.cpp
+++ b/compiler/z/codegen/OMRMachine.cpp
@@ -1822,8 +1822,13 @@ OMR::Z::Machine::assignBestRegisterPair(TR::Register    *regPair,
          {
          freeRegisterHigh->block();
          }
+
       freeRegisterLow = self()->reverseSpillState(currInst, lastReg);
-      freeRegisterLow->block();
+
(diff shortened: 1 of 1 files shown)


mgaudet:
(Commit message typo: 'teh')

fjeremic:
> (Commit message typo: 'teh')

Thanks! Fixed.

fjeremic:
@joransiu @hjooybar could you please review?

joransiu:
LGTM.

hjooybar:
LGTM.

PR #1078, opened 2017-05-01
Set arraycopy type in correct order

fjeremic:
Compressed and decompressed array copies are also primitive array copies
and as such we must set the type and element size of these special array
copies after setting the type and element size as derived from the fact
that they are primitive array copies as well.

This is done to enforce the fact that for the specially recognized
decompressed array copies we have type set to TR::Int16 and element size
derived from the respective type.


fjeremic:
@ncough could you please review?

ncough:
LGTM and tested on a known error case.

vijaysun-omr:
This fixed order is better than the original where some overwriting of the special decompressed arraycopy case could have happened.

PR #1080, opened 2017-05-01
Extend isPotentialOSRPoint to find OSR node

ncough:
isPotentialOSRPoint will now account for monents anchored under
treetops. It will also find the OSR node within the tree. This
is used throughout the rest of the API to determine what type
of OSR point is being handled. Failure cases have been
added to the other API calls if the OSR point is not known.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -659,7 +659,7 @@ bool OMR::Compilation::isShortRunningMethod(int32_t callerIndex)
    return false;
    }
 
-bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node)
+bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node, TR::Node **osrPointNode)
    {
    static char *disableAsyncCheckOSR = feGetEnv("TR_disableAsyncCheckOSR");
    static char *disableGuardedCallOSR = feGetEnv("TR_disableGuardedCallOSR");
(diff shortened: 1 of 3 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -784,12 +782,23 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    // If no induction after the OSR point, offset must be 0
    if (self()->getOSRTransitionTarget() != TR::postExecutionOSR)
       return 0;
+   
+   TR::Node *osrNode;
+   if (!self()->isPotentialOSRPoint(node, &osrNode))
```
andrewcraik: This should be in the assert so it gets compiled out

andrewcraik:
LGTM - prevents logic for finding OSR nodes from spreading in the compiler - we get it as a side-effect of checking the tree which is ideal.

andrewcraik:
@vijaysun-omr can you please review?

compiler/compile/OMRCompilation.cpp
```diff
@@ -784,12 +782,23 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
...
+      {
+      TR_ASSERT(0, "getOSRInductionOffset should only be called on OSR points");
```
vijaysun-omr: What is the relationship between this point in the code and getOSRInductionOffset, i.e. I don't see a call immediately to that routine.

PR #1081, opened 2017-05-01
Add exception points as potential OSR point in involuntaryOSR

cathyzhyi:
In involuntaryOSR, exception points could yield to VM and should
be treated as potential OSR points.


vijaysun-omr:
Should the header and commit message say "involuntary OSR" instead of "fsd" (since that is what the code is checking for) ?

cathyzhyi:
@vijaysun-omr , yea, thought about that too but "involuntary OSR" would make the header longer than the 50 characters limit.

PR #1082, opened 2017-05-01
Detect new Intel X86 microarchitecture

0dvictor:
1. Detect Skylake based Intel X86 CPU;
2. Inline logic querying old Intel CPUs.

### DIFF
--- a/compiler/env/ProcessorInfo.hpp
+++ b/compiler/env/ProcessorInfo.hpp
@@ -213,6 +213,7 @@ enum TR_ProcessorDescription
    TR_ProcessorIntelIvyBridge   = 0x0000000e,
    TR_ProcessorIntelHaswell     = 0x0000000f,
    TR_ProcessorIntelBroadwell   = 0x00000010,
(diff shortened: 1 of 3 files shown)


vijaysun-omr:
I assume "inlining logic querying old Intel CPUs" is a compile time savings attempt. How much did the change add to the JIT module size ? Just so that we keep an eye on that metric to ensure the tradeoff was a good one.

0xdaryl:
I agree with getting rid of the `isIntelOldMachine` query from OMR.  The naming isn't great because "Old" is very much a point-in-time statement, but that's to be dealt with in any code you're sinking this into.

0dvictor:
`isIntelOldMachine` is only used in one place and hence there should not be any footprint issue by inlining it.

0dvictor:
`isIntelOldMachine` is on the list of code cleanups. The code querying this function is actually being deprecated and will be removed soon. After that, `isIntelOldMachine` will be deleted as well.

PR #1084, opened 2017-05-02
Use 64bit std/ld instructions on a rare spill path

klangman:
When executing on 64bit platforms we must use ld/std rather then lwz/stw to spill registers.

Kevin Langman <langman@ca.ibm.com>


vijaysun-omr:
@zl-wang and @ymanton please review.

zl-wang:
I discussed this issue with Kevin before.  It is the right fix.

Issue #1085, opened 2017-05-03
Verbose bits not set in subproject


ncough:
Subprojects that rely on `setVerboseBitsInJitPrivateConfig()` will no longer correctly set verbose bits after #1007.

This is due to changes to `setVerboseBits()` as it no longer consider its base argument, rather it uses _feBase directly. However, `setVerboseBitsInJitPrivateConfig()` uses this argument to point at the `TR_JitPrivateConfig`, which contains the verbose bits. The immediate fix seems to be to revert to using the base argument, rather than _feBase, but I am not certain if this clashes with the greater intentions of the PR.

@mpirvu Will other changes related to #1055 account for this, or should this be fixed?

mpirvu:
I understand the issue. The code needs some refactoring and I'll work on it.

mpirvu:
Nick, could you please review my fix for this issue? Thanks

PR #1086, opened 2017-05-03
Fix processing of verbose options

mpirvu:
This commit fixes a bug introduced by PR #1007.
The new implementation of setVerboseBitsInJitPrivateConfig() is able to find
the JitConfig without looking at the value of parameter 'base'. However,
when trying to set the verbose flags in PrivateJitConfig for the trj9
subproject it uses setVerboseBits() which, similarly, was changed to
ignore the parameter 'base' and retrieve the OMR JitConfig from _feBase.
This is wrong.
The solution is to factor out the commong code that sets verbose bits
and call it from two different places:
1) setVerboseBitsInJitPrivateConfig() used by trj9 subproject
2) setVerboseBits() used by OMR generic code

Issue: #1085


ncough:
This change is failing in my local build due to `TR_Debug::dumpOptions`. Its unhappy when performing a comparison with the overloaded `setVerboseBits()`:

`../omr/compiler/ras/OptionsDebug.cpp:382: error: assuming cast to type ‘char* (*)(char*, void*, TR::OptionTable*)’ from overloaded function`

mpirvu:
The code compiled on my laptop, but apparently Linux complains about the overload. I have changed the name of the helper from the overloaded setVerboseBits() to setVerboseBitsHelper()

ncough:
LGTM, thanks for fixing the issue.

mpirvu:
Everything is good now. PR is ready to be merged.

PR #1089, opened 2017-05-04
Split block containing the call under OSR

ncough:
In any OSR mode, blocks containing OSR points should have exception
edges to the method's OSRCatchBlock. However, when inlining and
splitting a block in the caller, the callee's final block
and the second half of the split block may be merged, preventing
the addition of an edge to the OSRCatchBlock. This was fixed
under FSD by always splitting the block, however, it should also
be applied to other OSR modes as long as OSR exception edges
are expected to be found.


ncough:
@cathyzhyi Could you review this?

cathyzhyi:
LGTM.

andrewcraik:
LGTM - over to @vijaysun-omr

PR #1090, opened 2017-05-04
Do not generate decompression sequence under CS

dsouzai:
These commits remove the need for the compiler to generate the decompression sequence trees.

### DIFF
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -423,9 +423,8 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
 
    if (cg->isConcurrentScavengeEnabled())
       {
-      // TODO (GuardedStorage): We need to come up with a better solution than anchoring tree tops of compressedrefs
-      // sequences to enforce certain evaluation order. Perhaps not lowering compressedrefs in the first place when
-      // concurrent scavenge is supported is the correct solution here.
+      // TODO (GuardedStorage): We need to come up with a better solution than anchoring aloadi's
+      // to enforce certain evaluation order
       traceMsg(comp, "GuardedStorage: in performSetupForInstructionSelectionPhase\n");
 
       auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
@@ -439,7 +438,7 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
 
          traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
 
-         if ((comp->useCompressedPointers() && node->getOpCodeValue() == TR::l2a) || (!comp->useCompressedPointers() && node->getOpCodeValue() == TR::aloadi))
+         if (node->getOpCodeValue() == TR::aloadi)
             {
             TR::TreeTop* anchorTreeTop = TR::TreeTop::create(comp, TR::Node::create(TR::treetop, 1, node));
             TR::TreeTop* appendTreeTop = iter.currentTree();
@@ -449,7 +448,7 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
                appendTreeTop = currentTreeTopToappendTreeTop[appendTreeTop];
                }
 
-            // Anchor the l2a before the current treetop
+            // Anchor the aloadi before the current treetop
             appendTreeTop->insertBefore(anchorTreeTop);
             currentTreeTopToappendTreeTop[iter.currentTree()] = anchorTreeTop;
 
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -432,13 +432,66 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
       std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop*, TR::TreeTop*>, TR::Allocator> >
          currentTreeTopToappendTreeTop(std::less<TR::TreeTop*> (), mapAllocator);
 
+      TR_BitVector *unAnchorableAloadiNodes = comp->getBitVectorPool().get();
+
       for (TR::PreorderNodeIterator iter(comp->getStartTree(), comp); iter != NULL; ++iter)
          {
          TR::Node *node = iter.currentNode();
 
          traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
 
-         if (node->getOpCodeValue() == TR::aloadi)
+         // isNullCheck handles both TR::NULLCHK and TR::ResolveAndNULLCHK
+         // both of which do not operate on their child but their
+         // grandchild (or greatgrandchild).
+         if (node->getOpCode().isNullCheck())
+            {
+            // An aloadi cannot be anchored if there is a Null Check on
+            // its child. There are two situations where this occurs.
+            // The first is when doing an aloadi off some node that is
+            // being NULLCHK'd (see Ex1). The second is when doing an
+            // icalli in which case the aloadi loads the VFT of an
+            // object that must be NULLCHK'd (see Ex2).
+            //
+            // Ex1:
+            //    n1n NULLCHK on n3n
+            //    n2n    aloadi f    <-- First Child And Parent of Null Chk'd Node
+            //    n3n       aload O
+            //
+            // Ex2:
+            //    n1n NULLCHK on n4n
+            //    n2n    icall foo        <-- First Child
+            //    n3n       aloadi <vft>  <-- Parent of Null Chk'd Node
+            //    n4n          aload O
+            //    n4n       ==> aload O
+
+            TR::Node *nodeBeingNullChkd = node->getNullCheckReference();
+            if (nodeBeingNullChkd)
+               {
+               TR::Node *firstChild = node->getFirstChild();
+               TR::Node *parentOfNullChkdNode = NULL;
+
+               if (firstChild->getOpCode().isCall() &&
+                   firstChild->getOpCode().isIndirect())
+                  {
+                  parentOfNullChkdNode = firstChild->getFirstChild();
+                  }
+               else
+                  {
+                  parentOfNullChkdNode = firstChild;
+                  }
+
+               if (parentOfNullChkdNode &&
+                   parentOfNullChkdNode->getOpCodeValue() == TR::aloadi &&
+                   parentOfNullChkdNode->getNumChildren() > 0 &&
+                   parentOfNullChkdNode->getFirstChild() == nodeBeingNullChkd)
+                  {
+                  unAnchorableAloadiNodes->set(parentOfNullChkdNode->getGlobalIndex());
+                  traceMsg(comp, "GuardedStorage: Cannot anchor  %p\n", firstChild);
+                  }
+               }
+            }
+         else if (node->getOpCodeValue() == TR::aloadi &&
+                  !unAnchorableAloadiNodes->isSet(node->getGlobalIndex()))
             {
             TR::TreeTop* anchorTreeTop = TR::TreeTop::create(comp, TR::Node::create(TR::treetop, 1, node));
             TR::TreeTop* appendTreeTop = iter.currentTree();
@@ -455,6 +508,8 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
             traceMsg(comp, "GuardedStorage: Anchored  %p to treetop = %p\n", node, anchorTreeTop);
             }
          }
+
+      comp->getBitVectorPool().release(unAnchorableAloadiNodes);
       }
 
    if (cg->shouldBuildStructure() &&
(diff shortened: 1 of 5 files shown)


dsouzai:
@fjeremic could you review.

dsouzai:
@andrewcraik could you check if commit ce08df9 makes sense.

fjeremic:
@0xdaryl as promised this PR will address the concerns we raised when merging #911.

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -423,23 +423,63 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
...
+            // Ex2:
+            //    n1n NULLCHK on n4n
+            //    n2n    icall foo
+            //    n3n       aloadi <vft>
+            //    n4n          aload O
+
+            TR::Node *nodeBeingNullChkd = node->getNullCheckReference();
+            TR::Node *firstChild = node->getFirstChild();
+            TR::Node *grandChild = firstChild->getFirstChild();
+            TR::Node *greatGrandChild = grandChild->getFirstChild();
```
fjeremic: Are we sure the `grandChild` will always exist, i.e. is it always true that `grandChild != NULL`?

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -449,13 +489,15 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
...
-            // Anchor the l2a before the current treetop
+            // Anchor the aloadi before the current treetop
             appendTreeTop->insertBefore(anchorTreeTop);
             currentTreeTopToappendTreeTop[iter.currentTree()] = anchorTreeTop;
 
             traceMsg(comp, "GuardedStorage: Anchored  %p to treetop = %p\n", node, anchorTreeTop);
             }
          }
+
+      comp->getBitVectorPool().release(unAnchorableAloadiNodes);
```
fjeremic: Not relevant to this PR but this seems like a great candidate for RAII. @mgaudet any thoughts?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5065,17 +5066,6 @@ genericLoadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference *
...
-            // TODO (GuardedStorage): If we are in the evaluation of a compressedrefs sequence and are about to generate
-            // compressed load we override the instruction opcode to generate a guarded load and shift instead. We should
-            // figure out a better way to handle this.
-            if (cg->isEvaluatingCompressionSequence() && load == TR::InstOpCode::LLGF)
-               {
-               load = TR::InstOpCode::LLGFSG;
-               }
-            }
-
          generateRXInstruction(cg, load, node, targetRegister, tempMR);
```
fjeremic: Actually now looking at it this code was incorrect in its previous state. LLGFSG is an RXY instruction, not an RX instruction. Though I'm almost certain for all intents and purposes we would have generated the correct thing anyway.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6121,17 +6111,37 @@ aloadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference * tempM
...
+
                   if (cg->isConcurrentScavengeEnabled() &&
-                      !comp->useCompressedPointers() &&
                       node->getOpCodeValue() == TR::aloadi &&
                       tempReg->containsCollectedReference())
                      {
-                     generateRXYInstruction(cg, TR::InstOpCode::LGG, node, tempReg, tempMR);
+                     if (comp->useCompressedPointers())
+                        {
+                        if (TR::TransformUtil::fieldShouldBeCompressed(node, comp))
```
fjeremic: @vijaysun-omr FYI this is why its impotent that this API never returns false negatives.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6121,17 +6111,37 @@ aloadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference * tempM
...
+                           loadMnemonic = TR::InstOpCode::LLGFSG;
+                        else
+                           loadMnemonic = TR::InstOpCode::getLoadOpCode();
```
fjeremic: I'm not quite convinced this is correct. Why isn't this an LGG?

compiler/z/codegen/OMRTreeEvaluator.cpp
vijaysun-omr: Okay, I assume you meant "important" :)

compiler/codegen/OMRCodeGenPhase.cpp
mgaudet: You're right it's a good candidate, but I don't know if the machinery exists.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: > Okay, I assume you meant "important" :)

Thats a pretty bad typo haha!

andrewcraik:
commit ce08df9 makes sense, but the get of the grandchild needs to be guarded to make sure it is valid to get it.

compiler/codegen/OMRCodeGenPhase.cpp
dsouzai: My understanding is that the grandchild of a NULLCHK should always be defined (at least by the time we get to codegen - @andrewcraik can correct me if I'm wrong). The greatGrandChild might be NULL though. However I will add checks for both the grandChild and the greatGrandChild.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: You're right this should be a LGG (even if this evaluates to loading from the stack where a reference isn't compressed, it doesn't hurt).

dsouzai:
@fjeremic updated commits with code review changes.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: The guarded loads guard against the value _being loaded_ not the location it is being loaded from. Consider a case where we store a decompressed object from the nursery on the stack. This object still needs a read barrier when being loaded.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6121,17 +6111,37 @@ aloadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference * tempM
...
+                     if (comp->useCompressedPointers())
+                        {
+                        if (TR::TransformUtil::fieldShouldBeCompressed(node, comp))
+                           loadMnemonic = TR::InstOpCode::LLGFSG;
+                        else
+                           loadMnemonic = TR::InstOpCode::LGG;
+                        }
+                     else
+                        {
+                        loadMnemonic = TR::InstOpCode::LGG;
```
fjeremic: We can now merge the two else paths since they are equivalent.

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -487,28 +487,7 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
-         source = cg->evaluate(firstChild);
-         }
-         
-      cg->decEvaluatingCompressionSequence();
-      }
-   else
-      {
-      source = cg->evaluate(firstChild);
-      }
+   source = cg->evaluate(firstChild);
```
fjeremic: We can merge this line up to the declaration on line 488.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: Done.

compiler/z/codegen/UnaryEvaluator.cpp
dsouzai: Done.

compiler/z/codegen/OMRTreeEvaluator.cpp
dsouzai: Wouldn't any value on the reference that is stored on the stack be part of the root set?

compiler/codegen/OMRCodeGenPhase.cpp
andrewcraik: So the NULLCHK always operates on the grandchild so that is safe, the greatGrandChild is not guaranteed to exist and needs an appropriate check.

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -432,13 +432,64 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
...
+      TR_BitVector *unAnchorableAloadiNodes = comp->getBitVectorPool().get();
+
       for (TR::PreorderNodeIterator iter(comp->getStartTree(), comp); iter != NULL; ++iter)
          {
          TR::Node *node = iter.currentNode();
 
          traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
 
-         if (node->getOpCodeValue() == TR::aloadi)
+         if (node->getOpCodeValue() == TR::NULLCHK)
```
vijaysun-omr: Is there a similar need to consider TR::ResolveAndNULLCHK ? If yes, then perhaps the test above should read node->isNullCheck()

dsouzai: I made the change in 8f58ce352176a6e0aa1ce6a1db876373e320191b to check for the other *AndNULLCHK scenarios.

dsouzai:
travis-ci failed on osx because of this (as far as i can tell):

```
ld: unexpected bindingNone in '__ZN12TestCompiler15S390OpCodesTest18invokeAddressTestsEv' from ../../objs/compilertest_prod/fvtest/compilertest/tests/S390OpCodesTest.o for architecture x86_64

clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Why is the linker linking S390OpCodesTest.o on x86?

charliegracie:
@dsouzai I am not sure why it is being compiled but in your previous push it compiled [successfully](https://travis-ci.org/eclipse/omr/jobs/228977396)

dsouzai:
Not sure what the problem is given that the only change between the two pushes is this:

```
git diff c0a461c b065364
diff --git a/compiler/codegen/OMRCodeGenPhase.cpp b/compiler/codegen/OMRCodeGenPhase.cpp
index d4a8636..790d377 100644
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -440,7 +440,9 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
 
          traceMsg(comp, "GuardedStorage: Examining node = %p\n", node);
 
-         if (node->getOpCodeValue() == TR::NULLCHK)
+         // TR::checkcastAndNULLCHK does not set the ILProp2::NullCheck flag so we need to
+         // explicity check for it
+         if (node->getOpCode().isNullCheck() || node->getOpCodeValue() == TR::checkcastAndNULLCHK)
             {
             // An aloadi cannot be anchored if there is a Null Check on
             // its child. There are two situations where this occurs.
```

@mgaudet do you have any idea?

compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -432,13 +432,66 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
...
+            //
+            // Ex2:
+            //    n1n NULLCHK on n4n
+            //    n2n    icall foo
+            //    n3n       aloadi <vft>
+            //    n4n          aload O
+
+            TR::Node *nodeBeingNullChkd = node->getNullCheckReference();
+            TR::Node *firstChild = node->getFirstChild();
+            TR::Node *grandChild = firstChild->getFirstChild();
```
vijaysun-omr: I'm not sure this is a valid access for a checkcastAndNULLCHK since the node being null checked is just the first child of the checkcastAndNULLCHK node (look in the getNullCheckReference implementation to see what I mean) and it may not have a first child.

While I was looking at getNullCheckReference() I also noticed that it returns an answer for TR:ZEROCHK; do you therefore need to handle this case too ?

mgaudet:
Not sure! Internet suggests it's a compiler issue... maybe?

compiler/codegen/OMRCodeGenPhase.cpp
dsouzai: In conversation with Andrew he mentioned that TR::ZEROCHK was meant for divide by zero checks, so I don't think that's an issue. However I will have to fix up the checkcastAndNULLCHK scenario as well as the indirect call scenario after looking through the getNullCheckReference implementation. Thanks for pointing it out to me.

vijaysun-omr: I thought TR::DIVCHK was meant for divide by zero checks.

dsouzai: Yes you are right. I must have misunderstood the conversation. After more discussion I believe I do not need to worry about TR::ZEROCHK or TR::checkCastAndNULLCHK because I only need to not anchor aloadi nodes who's child is being NULLCHK'd. TR::ZEROCHK and TR::checkCastAndNULLCHK perform the null check on their child which is always safe to evaluate.

dsouzai:
@vijaysun-omr could you take a look at 6423e7eafc37c01bd9006ebdf74aeb078e980e6a with my review changes?

dsouzai:
The changes that this PR depended on are merged. It should be ok to merge this request if everything looks good.

vijaysun-omr:
Thanks for making the requested changes. Merging now.

PR #1091, opened 2017-05-04
Check runtime flag if CS code should be run

amicic:
In configuration, where Concurrent Scavenger is compile time enabled,
but run time disabled, we have to check run time flag before accessing
any CS functionality.

Fixing a specific spot in Global GC, that calls Scavenger to complete
any active concurrent cycle.


charliegracie:
@amicic Can you edit your commit message to limit the first line to 50 characters?  Thanks

amicic:
@charliegracie ok, will do. I had thought the rule was imposed on pull request title, only.

PR #1092, opened 2017-05-04
WIP: Separate cannot OSR at and during BCI

ncough:
This change modifies the implementation and use of the resolved
method symbol's API to determine if a BCI cannot OSR. It will now
consider the transition BCI, rather than the OSR node's BCI and
can distinguish between cases where it is not possible to
transition at a bytecode index and transition during an index,
eg. an inlined call.

This change includes:
 - Separate API into cannotAttemptOSRAt and cannotAttemptOSRDuring
 - Store cannotAttemptOSRDuring result against call site info
 - Fix check to verify OSR blocks for all callers exist and
   perform unreachable block cleanup
 - Don't generate OSR infrastructure if this call site cannot
   transition
 - Remove unused shouldNotOSR bitvector


andrewcraik:
I think this looks good, but I am a bit close to the change. @vijaysun-omr would you mind gives us your opinion on this?

PR #1093, opened 2017-05-05
Set array copy flags in VP for recognized methods

fjeremic:
VP may not have the type information of the source or destination arrays
we are copying. However for the specialized String compression array
copies we can be certain that both the source and destination arrays are
primitive arrays. As such it is always safe to set the primitive array
copy flags in VP for such recognized methods.

Failure to set these flags can get us in an invalid state in which we
fail to properly generate array store checks or misinterpret the special
String compression array copies as reference array copies, both of which
are incorrect.

### DIFF
--- a/compiler/optimizer/ValuePropagationCommon.cpp
+++ b/compiler/optimizer/ValuePropagationCommon.cpp
@@ -1295,6 +1295,13 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
          type = TR::Int8;
 
          elementSize = TR::Symbol::convertTypeToSize(type);
+
+         // VP may not know anything about the types of the objects we are copying, hence primitiveArray1 and
+         // and primitiveArray2 would both return false. However because we are dealing with a recognized method we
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM

mstoodle:
Slightly confusing because only the comment (and the motivating scenario) seem to have anything to do with AOT. Does AOT even need to be mentioned here? Presumably the parameters are arrays of primitives if the recognized methods are being used.

I would prefer the comment not make AOT the motivation, but rather an explanation for why this "override" behaviour is warranted for primitiveArray1 and primitiveArray2. Have I misread the motivation @fjeremic ?

fjeremic:
@mstoodle I agree. AOT is not worth mentioning here even though it is the motivator. Its a general problem of VP not having type information but we know the signature of the recognized method. Fixed in c522887.

mstoodle:
Looks good @fjeremic ! For pedantry, I'll wait until CI completes to merge, but please feel free to poke me if I don't notice right away!

PR #1094, opened 2017-05-05
Cleanup and fix GC ExclusiveVMAccess APIs

charliegracie:
There is a lot of duplicated code and makes modifications very
difficult. I am removing all of the APIs except 1 to keep things as
simple as possible.

The GC ExclusiveVMAccess APIs were not properly handling the case where
a system collect was called by a thread that already has
ExclusiveVMAccess outside of the GC. This could easily lead to a dead
lock because of the way the GC handles getting ExclusiveVMAccess. The
change also fixes the potential hang.

### DIFF
--- a/example/glue/EnvironmentDelegate.cpp
+++ b/example/glue/EnvironmentDelegate.cpp
@@ -26,10 +26,20 @@
 #include "GCExtensionsBase.hpp"
 
 void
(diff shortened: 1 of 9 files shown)


charliegracie:
@dmitripivkine Can you take a look and provide your thoughts on this change please?

dmitripivkine:
Looks good for me. This fix resolves problem observed.

charliegracie:
@youngar can you review this please?

youngar:
Looks good to me

PR #1095, opened 2017-05-05
Enable NextGenHCR by default

ncough:
Enable NextGenHCR, HCR implemented with on-stack replacement, by
default. This will result in OSR and HCR being enabled by default,
as this feature requires them.

These defaults are performed after OSR may be disabled in
`OMR::Compilation`, as defaulting before this point could result
in situations that do not require HCR having the older implementation
enabled instead.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -433,22 +433,25 @@ OMR::Compilation::Compilation(
 
    static const char *enableOSRAtAllOptLevels = feGetEnv("TR_EnableOSRAtAllOptLevels");
 
+   // If the outermost method is native, OSR is not supported
+   if (_methodSymbol->isNative())
+      self()->setOption(TR_DisableOSR);
+
+   // Do not default OSR and HCR on if:
+   // NextGenHCR is disabled, as they are enabled for it
+   // OSR is explicitly disabled
+   // FSD is enabled, as HCR cannot be enabled with it
+   if (!self()->getOption(TR_DisableNextGenHCR) && !self()->getOption(TR_DisableOSR) && !self()->getOption(TR_FullSpeedDebug))
+      {
+      self()->setOption(TR_EnableOSR); // OSR and HCR must be enabled for NextGenHCR
+      self()->setOption(TR_EnableHCR);
+      }
+
    if (self()->isDLT() || (((self()->getMethodHotness() < warm) || self()->compileRelocatableCode() || self()->isProfilingCompilation()) && !enableOSRAtAllOptLevels && !_options->getOption(TR_FullSpeedDebug)))
       {
       self()->setOption(TR_DisableOSR);
       _options->setOption(TR_EnableOSR, false);
       _options->setOption(TR_EnableOSROnGuardFailure, false);
-      }
-
-   if (_options->getOption(TR_EnableHCR))
-      {
-      if (!self()->getOption(TR_DisableOSR))
-         self()->setOption(TR_EnableOSR); // Make OSR the default for HCR
-      if (_options->getOption(TR_EnableNextGenHCR))
-         {
-         self()->setOption(TR_EnableOSR);
-         _options->setOption(TR_DisableOSR, false);
-         }
       }
 
    if (_options->getOption(TR_EnableOSR))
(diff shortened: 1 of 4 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -440,21 +440,29 @@ OMR::Compilation::Compilation(
       _options->setOption(TR_EnableOSROnGuardFailure, false);
       }
 
-   if (_options->getOption(TR_EnableHCR))
+   // Do not default OSR and HCR on if:
+   //  NextGenHCR is disabled, as they are enabled for it
+   //  OSR is explicitly disabled
+   //  FSD is enabled, as HCR cannot be enabled with it
```
fjeremic: Some spacing inconsistencies here.

compiler/compile/OMRCompilation.cpp
```diff
@@ -440,21 +440,29 @@ OMR::Compilation::Compilation(
...
-         {
-         self()->setOption(TR_EnableOSR);
-         _options->setOption(TR_DisableOSR, false);
-         }
+      self()->setOption(TR_EnableOSR); // OSR and HCR must be enabled for NextGenHCR
+      self()->setOption(TR_EnableHCR);
+
+      // TM results in difficult situations when determining where the transition should occur,
+      // so it is currently disabled. The implications of disabling TM should be investigated further.
+      self()->setOption(TR_DisableTM);
```
fjeremic: Have we done performance investigations to confirm this is safe on all platforms? Do we plan on addressing this problem in the future?

compiler/compile/OMRCompilation.cpp
ncough: These have been amended.

compiler/compile/OMRCompilation.cpp
ncough: This has been investigated and its no longer necessary to disable. Merge requests have been opened against the respective projects to address this.

compiler/control/OMROptions.cpp
```diff
@@ -1982,6 +1982,11 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
    // OSR is not available for ARM yet
    self()->setOption(TR_EnableOSR, false);
    self()->setOption(TR_EnableOSROnGuardFailure, false);
+   self()->setOption(TR_DisableNextGenHCR);
+#endif
+
+#ifndef J9_PROJECT_SPECIFIC
+   self()->setOption(TR_DisableNextGenHCR);
```
fjeremic: Is it certain that `OMR::Options::jitLatePostProcess` will run before `OMR::Compilation::Compilation` constructor which queries this option and make some decisions based on it?

compiler/compile/OMRCompilation.cpp
```diff
@@ -440,21 +440,25 @@ OMR::Compilation::Compilation(
...
+      self()->setOption(TR_EnableOSR); // OSR and HCR must be enabled for NextGenHCR
+      self()->setOption(TR_EnableHCR);
       }
 
    if (_options->getOption(TR_EnableOSR))
       {
       // Current implementation of partial inlining will break OSR
       self()->setOption(TR_DisablePartialInlining);
+      // Can't yet transition a MH thunk frame into equivalent interpreter frames
+      self()->setOption(TR_DisableMethodHandleThunks);
```
andrewcraik: This isn't the right answer - when compiling a MH thunk we want to disable OSR - we cannot OSR out of the top most method frame (ie method -1) since it is a thunk with no interpreter equivalent. That doesn't mean we can't support HCR since thunks themselves are immune from redefinition. We use traditional HCR guards for all the 'normal' bodies inlined into the thunk. For thunks inlined into thunks and similar the redefition of the thunk archetype only changes the code template and so future compiles will pickup the new bytecodes, but we do not compensate and cause existing thunks to pickup the new code - per discussion with Dan.

This needs to be revised since it will negatively affect performance downstream from here.

ncough: This has been addressed by disabling OSR when compiling MH thunks, rather than disabling MH thunks in OSR.

ncough:
@mpirvu As this is control related, could you review this?

ncough:
I agree with the point concerning the non-Java disable. A concern was raised as to whether `OMR::Options::jitLatePostProcess` will always run before `OMR::Compilation::Compilation`. Seeing as you are recommending this, I assume it will. The commit has been updated.

Currently, a great deal of OSR and HCR specific code is guarded by these options. As there is an intention to clarify the Disable/Enable OSR situation, I would prefer to delay a shift to persistentInfo until then, if thats alright with you.

mpirvu:
OMR::Options::jitLatePostProcess always runs before any OMR::Compilation::Compilation.

mpirvu:
I am not requesting changes to use the persistentInfo; just saying that the logic could become clearer that way: take input from user in terms of options, combine with input from the runtime and set runtime flags according to whatever you want to do.

vijaysun-omr:
Reviewed the options related changes here; they look okay to me. Merging.

PR #1096, opened 2017-05-05
Code cleanup - reduce compilation warnings

0dvictor:
Eliminated compilation warnings related to X86 binary encoding.

### DIFF
--- a/compiler/x/codegen/X86Ops.hpp
+++ b/compiler/x/codegen/X86Ops.hpp
@@ -421,99 +421,99 @@ class TR_X86OpCode
          }
       };
    // TBuffer should only be one of the two: Estimator when calculating length, and Writer when generating binaries.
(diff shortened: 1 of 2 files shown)


andrewcraik:
why so many whitespace changes mixed with the functional changes? it is a bit hard to review...

0dvictor:
They are actually not whitespace changes. The property querying functions are changed from `TYPE FUNC()` to `inline TYPE FUNC() const`.

andrewcraik:
I see that there are real changes, they are just intermixed with whitespace changes which just makes reviewing harder.

0dvictor:
Good point, I've separate them into two commits.

andrewcraik:
The refactored commits diff very nicely - I think this is LGTM. @0xdaryl can you please review?

andrewcraik:
The warnings in the x86 builds a quite annoying due to their verbosity could we please get this reviewed for merging to get the builds cleaned up please @0xdaryl ?

PR #1097, opened 2017-05-08
Avoid adding tabort to OSR guards

ncough:
Currently, a tabort to end a transaction will be placed before
any prepareForOSR. This is desirable as it prevents repeated
failures when performing an OSR transition. However, when using
OSR guards, it is more suitable to place the abort on the
taken side of the guard, as this will abort before any remat or
an OSR induce occurs.


vijaysun-omr:
I see that the change will prevent inserting a tabort before prepareForOSR. This is one of the two changes talked about in the description of the change, the other being "it is more suitable to place the abort on the
 taken side of the guard". Will that be delivered in another PR ?

ncough:
Cancelling this PR as the additional tabort will not result in incorrect behaviour.

PR #1100, opened 2017-05-08
Fix calculation of unneeded memory

### LINKED ISSUE #1098: Fix calculation of unneeded memory
omrmmap_dont_need() calculates the amount of unneeded memory using math on
"uintptr_t *" which gives a result in pointers, not bytes.

Cast the start pointer to "uintptr_t" and do the math.

pdbain-ibm:
The code was calculating the unneeded memory in units of pointers, not bytes.

Fixes #1098

Signed-off-by: Peter Bain <pdbain@ca.ibm.com>


pdbain-ibm:
Oops.  Forgot a platform...

PR #1102, opened 2017-05-08
Allow for pre and post OSR in one comp

ncough:
This change allows for pre and post OSR to both be enabled
in one compliation. This is necessary for NextGenHCR and
OSRG. This change also cleans up the resolved method symbol's API
to determine if a BCI cannot perform OSR.

This change includes:
 - Allow for preExecutionOSR and postExecutionOSR to both be set
 - Separate API into cannotAttemptOSRAt and cannotAttemptOSRDuring
 - Fix check to verify OSR blocks for all callers exist and
   perform unreachable block cleanup
 - Don't generate OSR infrastructure if this call site cannot
   transition
 - Remove unused shouldNotOSR bitvector

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -666,7 +666,7 @@ bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node, TR::Node **osrPointNo
    static char *disableMonentOSR = feGetEnv("TR_disableMonentOSR");
 
    bool potentialOSRPoint = false;
-   if (self()->getOSRTransitionTarget() == TR::postExecutionOSR)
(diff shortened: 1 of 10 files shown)


andrewcraik:
Not a fan of the & in conditions and Nic is working on addressing that. Overall the rest seems ok on an initial read - need to give it a bit more thought tomorrow since this is a very subtle area.

andrewcraik:
A separate future change would be to get rid of tests of EnableOSR to test for the OSR transition mode at least in the optimizer core to help clean up the story around the processing of EnableOSR and DisableOSR

andrewcraik:
LGTM - further clean up and refinement - seems sensible to me.

vijaysun-omr:
Looks valid to me. Merging.

PR #1103, opened 2017-05-08
Insert more asyncchecks for sampling before return

jdmpapin:
The compiler makes an effort to insert asyncchecks before returns in
methods that might otherwise fail to collect samples. However, some
methods could sometimes still get stuck at a low optimization level.

This change brings asynccheck insertion's heuristic more in line with
the one used in redundant asynccheck removal, by having it detect when
asyncchecks have been versioned out.

Furthermore, both asynccheck insertion and redundant asynccheck removal
are now allowed to run under voluntary OSR, since it's still safe to
insert asyncchecks immediately before returns.

### DIFF
--- a/compiler/optimizer/AsyncCheckInsertion.cpp
+++ b/compiler/optimizer/AsyncCheckInsertion.cpp
@@ -25,6 +25,7 @@
 #include "compile/Compilation.hpp"             // for Compilation
 #include "compile/ResolvedMethod.hpp"          // for TR_ResolvedMethod
 #include "compile/SymbolReferenceTable.hpp"    // for SymbolReferenceTable
+#include "control/Recompilation.hpp"           // for TR_Recompilation
 #include "env/StackMemoryRegion.hpp"
 #include "env/jittypes.h"                      // for TR_ByteCodeInfo, etc
 #include "il/Block.hpp"                        // for Block, toBlock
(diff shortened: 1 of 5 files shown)


vijaysun-omr:
@andrewcraik please review when you get a chance please.

andrewcraik:
LGTM - the heuristics in async check insertion mirror those in asynccheckremoval and asynccheckremoval is appropriately shackled to avoid generating invalid transition points when OSR is enabled.

compiler/optimizer/AsyncCheckInsertion.cpp
```diff
@@ -102,7 +119,18 @@ bool TR_AsyncCheckInsertion::shouldPerform()
...
+   // required infrastructure may not exist
+   //
+   if (comp()->getOption(TR_EnableOSR) && comp()->getOSRMode() == TR::involuntaryOSR)
+      return false;
+
+   // Should not be performed at very hot or scorching. Sampling does
+   // not matter at these points, as recompilation will not occur based
+   // on it.
+   //
+   if (comp()->getMethodHotness() >= veryHot)
```
jdmpapin: Probably not - I wasn't aware of non-profiling very hot.

This test isn't really necessary anyway because higher opt levels use `redundantAsyncCheckRemoval` instead. If you like I can delete it to let the strategy decide, or change it to be more precise, [perhaps something like this](https://github.com/eclipse/omr/blob/3a5b9f9b193b68cad98a07421aa34f6df6ae1614/compiler/optimizer/RedundantAsyncCheckRemoval.cpp#L157-L164).

vijaysun-omr: Thanks. Yes I would support using the more precise condition as in the commit that you linked to.

andrewcraik:
@jdmpapin could you please have a look at Vijay's last comment and amend appropriately so we can get this merged?

compiler/optimizer/AsyncCheckInsertion.cpp
jdmpapin: Updated

vijaysun-omr:
Thanks for the change. I'll merge once checks pass.

PR #1104, opened 2017-05-09
Remove commit message markdown suggestion

mgaudet:
Remove the suggestion that commit messages be written in markdown. This 
suggestion can confuse, as GitHub will not render messages as markdown, 
and will produce worse commit messages overall.                         
                                                                        
For example, we don't want people writing `[inline link](foo.example) `  
style.


charliegracie:
I think this change makes a lot of sense.  Thanks @mgaudet

PR #1106, opened 2017-05-09
Avoid heap resize in middle of concurrent sweep

LinHu2016:
- Do not resize LOA if concurrent sweep is active(require complete stw
sweep for LOA resize to avoid potential concurrent sweep issue)
  - Move resizeLOA from GlobalGCComlete hook to postCollect
  - Remove GlobalGCComlete hook

### DIFF
--- a/gc/base/MemoryPoolLargeObjects.cpp
+++ b/gc/base/MemoryPoolLargeObjects.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 1991, 2016
+ * (c) Copyright IBM Corp. 1991, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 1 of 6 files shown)


LinHu2016:
1, currently sweep code does not handle LOA resize in middle of concurrent sweep case.
2, updating concurrent sweep to handle LOA resize is complicated and risk.
3, currently we do have existing logic, which tries to complete concurrent sweep before LOA resize.
4, the condition check for triggering complete concurrent sweep is the same as the condition check for LOA resize.(ApproximateSOAFree < minimumSOAFreeRequired)
5, the condition check for triggering complete concurrent sweep is in sweep(in Collect), and LOA resize is in postCollect.
6, between two condition checks, ApproximateSOAFree could "shrink" due to resume an allocation or replenishment attempt at the end of collect.
7, there would be a case, which isn't protected (ApproximateSOAFree > minimumSOAFreeRequired during condition check for completing concurrent sweep and ApproximateSOAFree < minimumSOAFreeRequired when condition check during LOA resize)
8, adding condition check for completed concurrent sweep for invocation of LOA resize to prevent LOA resize in middle of concurrent sweep case.

charliegracie:
@LinHu2016 there is a compile failure in the PR.  Also can you capitalize the first letter in the first line of your commit message.  Thanks

charliegracie:
@LinHu2016 If a GC is using a concurrent sweep scheme it will most likely never do a contract of LOA with this change.  Is that correct? If so I am not sure that is the appropriate action for this.

LinHu2016:
@charliegracie , Hi Charlie, the main logic is unchanged (if we need to contract LOA, we would stop concurrent sweep mode and complete the rest sweep first, then try to contract LOA), in current code, 99% cases we handle well, but there is a hole we did not protected, we run condition check for contract LOA twice, one for determining if need complete the rest concurrent sweep, and another for determing if contract LOA, between two condition checks there could be one allocation, which causes result of first check is false and result of second check is true. this change just try to prevent hole case.

PR #1107, opened 2017-05-10
WIP: Handle method entry for involuntaryOSR

cathyzhyi:
In involuntaryOSR, method entries, although not explicitly represented
in the trees,  are yield points. Therefore OSRCatchBlock is not optional 
for this mode. This change adds  asynccheck at method entry to represent 
the OSR point if there is no other explicit OSR points in this method. 
Otherwise, the rest of the optimizer would try to remove the OSR infrastructure.


vijaysun-omr:
Is it possible that there was some potential OSR point(s) to begin with but was removed by some optimization later (based on analysis that it did) ? i.e. the async check was not added because there was some potential OSR point but a later analysis removes it and the OSR infrastructure ?

cathyzhyi:
after a better understanding between JIT and VM, this is no longer needed.

Issue #834, opened 2017-03-07
JitBuilder IL generation could be a lot less wasteful


mstoodle:
When I created JitBuilder, I saved time by using the simplest possible IL generation approach to avoid dealing with some of the subtleties of the OMR compiler's IL (primarily around node commoning and expression anchoring). Unfortunately, that means the initial version of the IL (before optimizations run) uses tons of expression nodes, tons of local variables, and tons of basic blocks that aren't, strictly speaking, necessary. Fortunately, the initial optimizer passes are pretty good at cleaning up the IL and getting it to a place you'd more or less expect from how the JitBuilder code is written. Then subsequent passes optimize it even further.

If the methods being compiled aren't very large (like the code samples in jitbuilder/release/src) then the inefficiency of the initial IL isn't a big issue because the optimizer cleans it up. Unfortunately, for JIT compilers or for more complex code generation tasks, this wasteful approach can explode resource usage and cross thresholds that can turn off certain kinds of analyses leaving the code in a less than hygienic state, and can even cause compilations to fail by running out of memory.

The basic idea I have is to generate every node underneath a TreeTop so it's anchored by design in the current block. If an IlValue is used within the same block, then IlBuilder can just use the node's address. I think a large number of uses will just be able to use the node address, which will save a lot on blocks, symbols, symbol references, and nodes. But if the IlValue comes from a different block, then we'll need to promote the TreeTop to a Store to  temp and then load the temp at the use point (much like the way it works now). 

I think the way forward here is to finally take advantage of the IlValue type name and store enough information to be able to tell if the value was created inside the current block or not. We'll lose some efficiency to store that info, but it's nowhere near the benefit we'll get from generating less IL in most cases. There will be some fallout to this approach as IlBuilder and other code in omr/compiler/ilgen knows that IlValue is currently just a different name for a SymbolReference and reaches into it. We'll need to fix that.

mstoodle:
Closing as this problem as been mostly addressed for the moment, at least.

PR #1110, opened 2017-05-11
Improve JitBuilder's initial IL quality

mstoodle:
enable Java "float" semantics for X86 code generator if using SSE for floats
Fix for #834 : first stage more efficient IL generation for JitBuilder
Added descriptions for fields of IlBuilder class

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -226,7 +226,7 @@ IlBuilder::printBlock(TR::Block *block)
    comp()->getDebug()->print(comp()->getOutFile(), tt);
    }
 
-TR::IlValue *
+TR::SymbolReference *
 IlBuilder::lookupSymbol(const char *name)
    {
    TR_ASSERT(_methodBuilder, "cannot look up symbols in an IlBuilder that has no MethodBuilder");
@@ -234,10 +234,10 @@ IlBuilder::lookupSymbol(const char *name)
    }
 
 void
-IlBuilder::defineSymbol(const char *name, TR::IlValue *v)
+IlBuilder::defineSymbol(const char *name, TR::SymbolReference *symRef)
    {
    TR_ASSERT(_methodBuilder, "cannot define symbols in an IlBuilder that has no MethodBuilder");
-   _methodBuilder->defineSymbol(name, v);
+   _methodBuilder->defineSymbol(name, symRef);
    }
 
 void
@@ -250,21 +250,29 @@ IlBuilder::defineValue(const char *name, TR::IlType *type)
    }
 
 TR::IlValue *
-IlBuilder::newValue(TR::DataType dt)
+IlBuilder::newValue(TR::DataType dt, TR::Node *n)
    {
-   TR::SymbolReference *newSymRef = symRefTab()->createTemporary(methodSymbol(), dt);
-   newSymRef->getSymbol()->setNotCollected();
-   char *name = (char *) _comp->trMemory()->allocateHeapMemory(5 * sizeof(char));
-   sprintf(name, "_T%-3d", newSymRef->getCPIndex());
-   newSymRef->getSymbol()->getAutoSymbol()->setName(name);
-   defineSymbol(name, newSymRef);
-   return newSymRef;
+   // make sure TreeTop is well formed
+   TR::Node *ttNode = n;
+   if (!ttNode->getOpCode().isTreeTop())
+      ttNode = TR::Node::create(TR::treetop, 1, n);
+
+   TR::TreeTop *tt = TR::TreeTop::create(_comp, ttNode);
+   _currentBlock->append(tt);
+   TR::IlValue *value = new (_comp->trHeapMemory()) TR::IlValue(n, tt, _currentBlock, _methodBuilder);
+   return value;
+   }
+
+TR::IlValue *
+IlBuilder::newValue(TR::IlType *dt, TR::Node *n)
+   {
+   return newValue(dt->getPrimitiveType(), n);
    }
 
 TR::IlValue *
 IlBuilder::NewValue(TR::IlType *dt)
    {
-   return newValue(dt->getPrimitiveType());
+   TR_ASSERT_FATAL(0, "should not create a value without a TR::Node");
    }
 
 TR::TreeTop *
@@ -440,8 +448,11 @@ IlBuilder::zero(TR::DataType dt)
       {
       case TR::Int8 :  return TR::Node::bconst(0);
       case TR::Int16 : return TR::Node::sconst(0);
+      case TR::Int32 : return TR::Node::iconst(0);
+      case TR::Int64 : return TR::Node::lconst(0);
       default :        return TR::Node::create(TR::ILOpCode::constOpCode(dt), 0, 0);
       }
+   TR_ASSERT(0, "should not reach here");
    }
 
 TR::Node *
@@ -453,14 +464,13 @@ IlBuilder::zero(TR::IlType *dt)
 TR::Node *
 IlBuilder::zeroNodeForValue(TR::IlValue *v)
    {
-   return zero(v->getSymbol()->getDataType());
+   return zero(v->getDataType());
    }
 
 TR::IlValue *
 IlBuilder::zeroForValue(TR::IlValue *v)
    {
-   TR::IlValue *returnValue = newValue(v->getSymbol()->getDataType());
-   storeNode(returnValue, zeroNodeForValue(v));
+   TR::IlValue *returnValue = newValue(v->getDataType(), zeroNodeForValue(v));
    return returnValue;
    }
 
@@ -549,14 +559,13 @@ IlBuilder::AppendBuilder(TR::IlBuilder *builder)
 TR::Node *
 IlBuilder::loadValue(TR::IlValue *v)
    {
-   return TR::Node::createLoad(v);
+   return v->load(_currentBlock);
    }
 
 void
-IlBuilder::storeNode(TR::IlValue *dest, TR::Node *v)
+IlBuilder::storeNode(TR::SymbolReference *symRef, TR::Node *v)
    {
-   appendBlock();
-   genTreeTop(TR::Node::createStore(dest, v));
+   genTreeTop(TR::Node::createStore(symRef, v));
    }
 
 void
@@ -590,8 +599,7 @@ IlBuilder::indirectLoadNode(TR::IlType *dt, TR::Node *addr, bool isVectorLoad)
 
    TR::Node *loadNode = TR::Node::createWithSymRef(loadOp, 1, 1, addr, storeSymRef);
 
-   TR::IlValue *loadValue = NewValue(baseType);
-   storeNode(loadValue, loadNode);
+   TR::IlValue *loadValue = newValue(baseType, loadNode);
    return loadValue;
    }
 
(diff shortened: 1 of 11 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1794,6 +1754,8 @@ IlBuilder::genCall(TR::SymbolReference *methodSymRef, int32_t numArgs, TR::IlVal
    for (int32_t a=0;a < numArgs;a++)
       {
       TR::IlValue *arg = argValues[a];
+      if (arg->getDataType() == TR::Int8 || arg->getDataType() == TR::Int16 || Word == Int64 && arg->getDataType() == TR::Int32)
+         arg = ConvertTo(Word, arg);
```
Leonardo2718: I'm not sure I follow this. Why is it needed now and not before?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1802,8 +1764,8 @@ IlBuilder::genCall(TR::SymbolReference *methodSymRef, int32_t numArgs, TR::IlVal
 
    if (returnType != TR::NoType)
       {
-      TR::IlValue *returnValue = newValue(callNode->getDataType());
-      genTreeTop(TR::Node::createStore(returnValue, callNode));
+      TR::IlValue *returnValue = newValue(callNode->getDataType(), callNode);
+      //genTreeTop(TR::Node::createStore(returnValue->getSymbolReference(), callNode));
```
Leonardo2718: I think this line should be removed.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1853,8 +1813,8 @@ IlBuilder::AtomicAddWithOffset(TR::IlValue * baseAddress, TR::IlValue * offset,
       }
 
    genTreeTop(callNode); 
-   TR::IlValue *returnValue = newValue(callNode->getDataType());
-   genTreeTop(TR::Node::createStore(returnValue, callNode)); 
+   TR::IlValue *returnValue = newValue(callNode->getDataType(), callNode);
+   //genTreeTop(TR::Node::createStore(returnValue->getSymbolReference(), callNode));
```
Leonardo2718: I think this line should be removed.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1950,68 +1910,67 @@ IlBuilder::AtomicAdd(TR::IlValue * baseAddress, TR::IlValue * value)
...
+   //ending the transaction at the end of transactionBuilder
+   appendBlock();
+   TR::Node *tEndNode=TR::Node::create(TR::tfinish,0);
+   tEndNode->setSymbolReference(comp()->getSymRefTab()->findOrCreateTransactionExitSymbolRef(comp()->getMethodSymbol()));
+   genTreeTop(tEndNode);
 
-	}  
+   //Three IlBuilders above merged here
+   appendBlock(mergeBlock);
+   }
```
Leonardo2718: Separating the white space changes in a separate commit would make this kind of thing a lot easier to review in future (but I don't need to remind _you_ of this right :wink:).

compiler/ilgen/IlValue.cpp
```diff
@@ -0,0 +1,109 @@
...
+      {
+      // probably not very common scenario, but if both values are in the same block as current
+      // then storeOver just needs to update node pointer
+      _nodeThatComputesValue = value->_nodeThatComputesValue;
+      }
+   else
+      {
+      // more commonly, if any value is in another block or this use will be in a different block, first ensure this value is
+      // stored to an auto symbol
+      // NOTE this may mean nodes may be stored to more than one auto, but that's ok
```
Leonardo2718: Out of curiosity, why exactly is this considered okay?

mstoodle:
Thanks for the quick review @Leonardo2718 : some valuable feedback and I'll address the points you raised.

I'm surprised at the build results, since I've been working with this code for quite a while on my laptop, but clearly need to resolve some stuff.

compiler/ilgen/IlValue.cpp
mstoodle: it will only store to at least one additional auto (i.e. the original creation of the node may be fed into a store to an auto), and it's ok because a specific auto is needed to make sure the correct value will be communicated to the use. We cannot reuse any other auto because that auto's value may change between the store and the use point.

mstoodle: global value numbering or global register allocator should also be able to remove any unnecessary duplicate auto's too.

compiler/ilgen/IlBuilder.cpp
mstoodle: thanks for pointing that out i actually forgot I had done this one....i'll separate it out, and maybe separate out some of the API changes that tended to cast a wide net.

compiler/ilgen/IlValue.cpp
```diff
@@ -0,0 +1,109 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2017, 2017
```
charliegracie: Is this code new or copied from other files? Just curious if the initial copyright is correct

mstoodle: Hm, it's virtually new (when I first created the file it was 100% new :) ) but some code from another file crept in. I'll change the initial date to 2016. Good catch, Charlie!

compiler/ilgen/IlValue.cpp
```diff
@@ -0,0 +1,109 @@
...
+   {
+   if (_symRefThatCanBeUsedInOtherBlocks == NULL)
+      {
+      TR::Compilation *comp = TR::comp();
+
+      // first use from another block, need to create symref and insert store tree where node  was computed
+      TR::SymbolReference *symRef = comp->getSymRefTab()->createTemporary(_methodBuilder->methodSymbol(), _nodeThatComputesValue->getDataType());
+      symRef->getSymbol()->setNotCollected();
+      char *name = (char *) comp->trMemory()->allocateHeapMemory(8 * sizeof(char));
+      sprintf(name, "_T%-5d", symRef->getCPIndex());
```
fjeremic: `getCPIndex` returns an `int32_t`. Won't this overrun the buffer `name` if the CP index is 6 characters or more?

mstoodle: You're correct. If we managed to create a million symbol references in a single compilation, it would overflow. I suppose a bug that permitted a garbage value in there could also overflow.

This problem pre-existed my change, but I'll add a commit here to address it. Thanks for  pointing it out @fjeremic .

mstoodle:
This PR turned into something of a grab bag of commits.

I'll separate out some of the "hygiene" commits so this merge is more about the actual subject of this PR, and I'll create a separate hygiene PR.

mstoodle:
I trimmed this PR down to 3 commits that relate to improving the IL created by JitBuilder.

compiler/ilgen/IlValue.cpp
```diff
@@ -0,0 +1,109 @@
...
+      symRef->getSymbol()->getAutoSymbol()->setName(name);
+
+      // create store and its treetop
+      TR::Node *storeNode = TR::Node::createStore(symRef, _nodeThatComputesValue);
+      TR::TreeTop *prevTreeTop = _treeTopThatAnchorsValue->getPrevTreeTop();
+      TR::TreeTop *newTree = TR::TreeTop::create(comp, storeNode);
+      newTree->insertNewTreeTop(prevTreeTop, _treeTopThatAnchorsValue);
```
vijaysun-omr: Is it a given that the _nodeThatComputesValue can be evaluated before the _treeTopThatAnchorsValue ? I am thinking of an example of an idiv node that was evaluated for the first time under a DIVCHK parent originally but now gets evaluated under the store to the temp inserted as the newTree. Maybe such cases cannot be encountered at this spot in the code for some other reason (?)

mstoodle: That's a subtle point, it's true. There are currently only the *WithOverflow style of tests that do this kind of thing right now (DIVCHK isn't currently used in JitBuilder <gasp>) and they do not expose the node underneath as an IlValue so it doesn't matter there.

If we start inserting the check trees for something like idiv, then my model would be: JitBuilder should insert the DIVCHK tree with the node underneath, then NewValue should create a second TreeTop where this storing-to-temp business occurs (if needed). That way, the check tree will always anchor the first evaluation so that it happens before the store, which is the proper order to be able to reuse the value anyway. Dead trees will clean up the extra treetop if it's not needed just like it does for all the other extra treetops being left around.

Does that address the general concern (even if I made idiv seem even scarier to you now)?

vijaysun-omr: So, if a check opcode like DIVCHK were ever added to OMR, the solution for this case (above code) would be one of 
1) Explicitly look for the check opcode and if we find one, add a store after the check tree to preserve the ordering constraint in that case, but continue to add a store before the tree in every other case.
2) Do not look for check opcode in the above logic but instead change the transformation to always create a treetop with the value we need anchored under it and then a store with the anchored value as the child (rhs) after that treetop ? I only ask about this because there is talk of dead trees elimination in the above description and I associated that somehow  with a treetop that must be created somewhere (though I am clearly unsure why the treetop is needed).

Is it safer to add a store after the _treeTopThatAnchorsValue in every case except when it is a control flow opcode (i.e. if or goto or switch or return etc.) ?

mstoodle: Actually I was thinking that any future service that wants to use some kind of CHK opcode would explicitly generate the check as a treetop (there are some examples in IlBuilder where genTreeTop() is called directly, without using newValue() ).

The value to be checked should be created using newValue(), which would append *another* treetop with the same node as its child (even though this treetop is redundant in this case because the check treetop will have already anchored the node). If that value were ever used outside the block containing the node, then the another treetop (wow, sounds like a patent application now) would be replaced with a store. 

Imagine a DIVCHK is created for a new value that is the result of an idiv.

```
DIVCHK
    idiv
       left
       right
treetop
    ==>idiv
```

If the IlValue associated with the idiv node is used outside the block, the existing code in IlValue will transform that second treetop so that it looks like:
```
DIVCHK
    idiv
       left
       right
istore <temp slot X>
    ==>idiv
```

So the DIVCHK, because it precedes the treetop that newValue adds, is the *real* anchor point for the idiv. When there is no CHK opcode, the treetop inserted by newValue guarantees the right anchor point.

Dead trees is going to be (PR coming :) ) the first optimization to run, and it will end up removing lots of these treetops if they aren't really needed as anchor points.

vijaysun-omr: I am still unclear on how the store would end up after the DIVCHK in your picture given that I though _treeTopThatAnchorsValue is the DIVCHK tree and if I look at the code above the store goes before that tree. Maybe _treeTopThatAnchorsValue is'nt the DIVCHK tree...

mstoodle: Ah yes, I can see how the name of the field introduces some confusion since, in the example I was describing, _treeTopThatAnchorsValue is actually the "treetop" not the DIVCHK. The DIVCHK needs to be created independently. When IlBuilder::newValue(TR::Int32, idivNode) is called, it will generate "treetop" and pass that as _treeTopThatAnchorsValue to the IlValue constructor.

So the IlValue will be completely unaware that there is a DIVCHK tree there. It will stay there, and IlValue::storeToAuto() will only work with the "treetop".

Does that help?

vijaysun-omr: Yes that makes things clearer. Thanks for the explanation.

PR #1111, opened 2017-05-11
Check the number of pending pushes on a method

ncough:
Under OSR, it will not be possible to transition correctly
if we allocate more pending push slots than the VM expects. To ensure
this never occurs, the JIT will perform a fatal assertion if it
detects a requested slot beyond the limit.

### DIFF
--- a/compiler/ilgen/ByteCodeIteratorWithState.hpp
+++ b/compiler/ilgen/ByteCodeIteratorWithState.hpp
@@ -54,6 +54,7 @@ template<typename ByteCode,
         _tryCatchInfo(comp->allocator("IlGen"))
       {
       _printByteCodes = (comp->getOutFile() != NULL && comp->getOption(TR_TraceBC) && (comp->isOutermostMethod() || comp->getOption(TR_DebugInliner) || comp->trace(OMR::inlining)));
+      _cannotAttemptOSR = comp->getOption(TR_EnableOSR) && !comp->isPeekingMethod() && methodSym->cannotAttemptOSRDuring(comp->getCurrentInlinedSiteIndex(), comp);
       }
 
    void printByteCodes()
@@ -421,6 +422,7 @@ protected:
    TR_LinkHead<IndexPair>                         _backwardBranches;
    bool                                           _inExceptionHandler;
    bool                                           _printByteCodes;
+   bool                                           _cannotAttemptOSR;
 
    flags8_t *                                     _flags;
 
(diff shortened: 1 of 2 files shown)


mstoodle:
could you expand "PP" in the commit message and title please; should be "pending pushes" from the context, but that may not be clear to others.

compiler/ilgen/ByteCodeIteratorWithState.hpp
```diff
@@ -54,6 +54,7 @@ template<typename ByteCode,
         _tryCatchInfo(comp->allocator("IlGen"))
       {
       _printByteCodes = (comp->getOutFile() != NULL && comp->getOption(TR_TraceBC) && (comp->isOutermostMethod() || comp->getOption(TR_DebugInliner) || comp->trace(OMR::inlining)));
+      _cannotAttemptOSR = comp->getOption(TR_EnableOSR) && !comp->isPeekingMethod() && methodSym->cannotAttemptOSRDuring(comp->getCurrentInlinedSiteIndex(), comp);
```
andrewcraik: I assume this is ok even if getCurrentInlinedSiteIndex returns -1?

andrewcraik:
Change looks good - just one question to resolve and then it should be good to merge.

compiler/ilgen/ByteCodeIteratorWithState.hpp
ncough: Yes, `cannotAttemptOSRDuring()` will always safely return false for a site index of -1, as it is only concerned with methods above it on the stack.

andrewcraik: ok sounds good and the EnableOSR check handles -1 at the top level which makes sense I guess.

vijaysun-omr:
The commit (and PR description) message still has "PP" in it.

andrewcraik:
@vijaysun-omr I believe nic has now corrected the PP issue

vijaysun-omr:
Could you please elaborate on

"To assist in preventing this from occuring, ByteCodeIteratorWithState
will determine if OSR is possible within a method. Pending push
stores can then be avoided if the stack were to not match the VM's
expectation."

How does determining if OSR is possible in the method related to preventing the "wrong" number of pending pushes ?

ncough:
If the method is an internal implementation, different to what the VM expects, they will disagree on the number of pending push slots that are required. Therefore, ilgen will have to avoid requesting the slots. It also provides some compile time savings as the OSR overhead can be skipped in ilgen if its already known to not be possible.

vijaysun-omr:
Okay I understand. The usages of the boolean you added are probably expected in upstream projects.

PR #1112, opened 2017-05-11
Lint error fixes and travis integration

sajidahmed21:
Contains the following:

- Fixes for lint errors in p and z specific code
- Build jobs for running linter for p and z targets on Travis CI

### DIFF
--- a/scripts/build-on-travis.sh
+++ b/scripts/build-on-travis.sh
@@ -43,6 +43,17 @@ else
   if test "x$RUN_LINT" = "xyes"; then
     llvm-config --version
     clang++ --version
+
+    # Run linter for x86 target
+    time make --jobs $JOBS lint
+
+    # Run linter for p and z targets
+    export TARGET_ARCH=p
+    export TARGET_BITS=64
+    time make --jobs $JOBS lint
+
+    export TARGET_ARCH=z
+    export TARGET_BITS=64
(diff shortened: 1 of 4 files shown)


mgaudet:
Your commit messages need to be word-wrapped at 72chars. 

It might be worth trying to combine the three platform lints into a single Travis job. We're limited on the number of concurrent jobs we're able to run.

mgaudet:
Everything else LGTM.

sajidahmed21:
I've made the suggested changes. Please check.

scripts/build-on-travis.sh
```diff
@@ -43,6 +43,16 @@ else
...
+
+    # Run linter for x86 target
+    time make --jobs $JOBS lint
+
+    # Run linter for p and z targets
+    export TARGET_ARCH=p
+    export TARGET_BITS=64
+    time make --jobs $JOBS lint
+
+    export TARGET_ARCH=z
```
mstoodle: I guess you're "inheriting" the earlier setting of TARGET_BITS here, but I would prefer you set it explicitly so that it's clear what it is being tested  for the z platform and also will be less brittle to subsequent changes.

sajidahmed21: I was probably trying to adhere to DRY, but I agree, it makes sense not to, in this case.

Issue #1113, opened 2017-05-12
heap-resize reports incorrect amount value for type "release free pages"


Param-S:
Random cases where the value printed for heap-resize is not valid wrt "release free pages" type

<heap-resize id="6109" type="release free pages" space="tenure" **amount="18446744073416224768"** count="1" timems="4.125" reason="idle" timestamp="2017-05-08T03:15:29.279" />

from debugging, it seems to me that the condition where the FreeEntry size falls below the page size, is not taken into consideration by the following statement.

```
uintptr_t
MM_MemoryPoolAddressOrderedListBase::releaseFreeEntryMemoryPages(...)
{
...
uintptr_t totalFreePagesCount = (currentFreeEntry->getSize() - (addressBase - (uintptr_t)currentFreeEntry)) / pageSize;
...
}
```

ashu-mehra:
I think this issue is fixed by https://github.com/eclipse/omr/pull/1114 and can be closed.

charliegracie:
Closing as this was resolved in #1114

PR #1114, opened 2017-05-12
Corrects heap-resize tag conflict in idle case

Param-S:
While calculating the no. of pages to be released from a heap free
entry, it failed to recognize the entries which are lesser than page
size.  Variables used in the calculation is of type uintptr_t, while
subtraction the negative value treated as positive value because of
unsigned.

```
MM_MemoryPoolAddressOrderedListBase::releaseFreeEntryMemoryPages(...)
{
...
uintptr_t totalFreePagesCount = (currentFreeEntry->getSize() - (addressBase - (uintptr_t)currentFreeEntry)) / pageSize;
...
}
```

Issue: #1113

Signed-off-by: Parameswaran Selvam <parselva@in.ibm.com>


charliegracie:
@Param-S could you limit the first line of the commit message to 50 characters please?

Param-S:
@charliegracie corrected commit message.

amicic:
Wouldn't changing totalFreePagesCount to a signed int fix the problem? You have effectively same checks in lines 180 and 184.

Param-S:
True, It did not strike me.  I will correct it my next PR.

PR #1115, opened 2017-05-12
Add OSR infra for NextGenHCR in outermost method

ncough:
Under NextGenHCR, it may be necessary to transition from
the start of the method, in the event that an assumption
was invalidated during the method prologue. Therefore,
it is necessary to always add OSR infrastructure for
the outermost method with an edge to the catch block
from the start.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -856,6 +856,22 @@ OMR::ResolvedMethodSymbol::genAndAttachOSRCodeBlocks(int32_t currentInlinedSiteI
 
    TR::TreeTop *lastTreeTop = NULL;
    TR_OSRMethodData *osrMethodData = self()->comp()->getOSRCompilationData()->findOrCreateOSRMethodData(currentInlinedSiteIndex, self());
+
+   // Under NextGenHCR, the method entry for the outermost method is an implicit OSR point
+   // To ensure a transition from this point is possible, it is necessary to generate and attach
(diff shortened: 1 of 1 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -856,6 +856,20 @@ OMR::ResolvedMethodSymbol::genAndAttachOSRCodeBlocks(int32_t currentInlinedSiteI
 
    TR::TreeTop *lastTreeTop = NULL;
    TR_OSRMethodData *osrMethodData = self()->comp()->getOSRCompilationData()->findOrCreateOSRMethodData(currentInlinedSiteIndex, self());
+
+   // Under OSR, the method entry for the outermost method is an implicit OSR point
+   // To ensure a transition from this point is possible, it is necessary to generate and attach
+   // a catch and code block now, with an exception edge to the catch block from the first block
+   //
+   bool genForOuterEntry = self()->comp()->isOutermostMethod() && self()->comp()->getHCRMode() == TR::osr;
```
andrewcraik: What about FSD and other OSR modes? The comment makes it sound like this should be a test for any OSR at all not just nextGenHCR

ncough: The comment has been updated to reflect NextGenHCR's requirements.

andrewcraik:
LGTM - need an exception edge from the first block so we can protect fear reaching the entry from potential yields in the method prologue

vijaysun-omr:
Seems like a self contained safe change. Merging.

PR #1125, opened 2017-05-15
Add OSR point for NextGenHCR in outermost method

ncough:
Under NextGenHCR, it may be necessary to transition from
the start of the method, in the event that an assumption
was invalidated during the method prologue. A prior change
added exceptions edges to handle this, however, parts
of the OSR API require an OSR point to also be present. This
change adds the OSR point, with the correct bytecode index.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -863,13 +863,19 @@ OMR::ResolvedMethodSymbol::genAndAttachOSRCodeBlocks(int32_t currentInlinedSiteI
    // This exception edge should remain until OSR guards have been inserted, at which point it is no
    // longer needed
    //
-   bool genForOuterEntry = self()->comp()->isOutermostMethod() && self()->comp()->getHCRMode() == TR::osr;
-   if (genForOuterEntry)
+   if (self()->comp()->isOutermostMethod() && self()->comp()->getHCRMode() == TR::osr)
       {
+      // Add an exception edge to the OSR catch block from the first block
       TR::Block *OSRCatchBlock = osrMethodData->findOrCreateOSRCatchBlock(self()->getFirstTreeTop()->getNode());
       TR::Block *firstBlock = self()->getFirstTreeTop()->getEnclosingBlock();
       if (!firstBlock->hasExceptionSuccessor(OSRCatchBlock))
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -863,13 +863,19 @@ OMR::ResolvedMethodSymbol::genAndAttachOSRCodeBlocks(int32_t currentInlinedSiteI
...
       {
+      // Add an exception edge to the OSR catch block from the first block
       TR::Block *OSRCatchBlock = osrMethodData->findOrCreateOSRCatchBlock(self()->getFirstTreeTop()->getNode());
       TR::Block *firstBlock = self()->getFirstTreeTop()->getEnclosingBlock();
       if (!firstBlock->hasExceptionSuccessor(OSRCatchBlock))
          self()->comp()->getFlowGraph()->addEdge(TR::CFGEdge::createExceptionEdge(firstBlock, OSRCatchBlock, self()->comp()->trMemory()));
+
+      // Create an OSR point with bytecode index 0
+      TR_ByteCodeInfo firstBCI = self()->getFirstTreeTop()->getNode()->getByteCodeInfo();
+      firstBCI.setByteCodeIndex(0);
```
vijaysun-omr: Why does the bytecode index have to be set to 0 ? Why not use what's there ?

andrewcraik: The reason for the transition being to 0 is that the guard is being inserted at the top of the method with no trees having been run. We have to set this transition up in the il generator for the language concerned and so the state we record is for bytecode index 0. If any trees have moved or changed we don't want to use their bytecode index, we want to go from the entry of the method to the entry of the method in the interpreter. These points are equivalent by construction since nothing will have been run. Using what is there is dangerous because it may not have been considered an OSR point and we can't allow anything to run - we literally want to go from having executed none of the method semantics to the point of having run none of the method semantics. This hard coding achieves this.

vijaysun-omr: Okay so we have prepared for a transition at bytecode index 0 specifically in the IL gen which is what makes it the value you want to use here when creating an OSR point.

ncough: As assumptions may be invalidated during the method prologue and these may affect the JITed code representing bytecode index 0, it is necessary to allow for a transition before this is executed. There is no guarantee that the first treetop has this bytecode index correctly set, so this is done to be certain.

Issue #1126, opened 2017-05-15
GRA support for Vector registers on x86


IBMJimmyk:
To further expand on autoSIMD support on x, GRA support will be added for vector registers on x. This change will indicate to GRA that TR_VRF registers exist and that GRA can be performed on them. This is done by indicating which range of registers can hold TR_VRF registers. Also various functions that deal with registers will need to be updated to recognize TR_VRF registers. In most cases, these functions will perform a similar action to how TR_FPR registers are handled but slightly modified for the TR_VRF register case.

0xdaryl:
@IBMJimmyk , can this issue be closed now or is there outstanding work to be done?

PR #1127, opened 2017-05-15
Support for GRA for vector registers on x

IBMJimmyk:
Modified initialize inside OMRCodeGenerator.cpp to indicate the first and
last real registers that can hold vector values. This information is used
by GRA to perform GRA.

GRA can lead to the creation of register dependencies so the register
dependency code inside OMRRegisterDependency.cpp was updated to handle
TR_VRF registers. This change also updates several other register related
functions to handle TR_VRF registers. In most cases, this results in
perform actions similar to the TR_FPR case but slightly modified for the
TR_VRF case.

Issue: #1126


vijaysun-omr:
I would like a review from @0xdaryl and @andrewcraik

IBMJimmyk:
@gita-omr also wanted to review the code.

gita-omr:
Sorry, this is something we should have put in earlier. This is necessary for auto-SIMD support in GRA on Intel. Since current auto-SIMD support on Intel is very basic there is very low chance that GRA will even kick in. But in case it does, these changes are necessary.

0xdaryl:
Looks generally OK as an extension to what's already there.

PR #1128, opened 2017-05-15
Fix support query of vector multiply sum instruction on Z

hjooybar:
VMSL instruction is used to improve ECC performance.
Like other vector instructions, this instruction can only be used on machines with proper hardware and OS support for vector instructions.
This PR prevents using VMSL instruction when hardware or OS do not have support for vector instructions.


vijaysun-omr:
Change looks safe me to. Check for vector support before using such instructions. Merging.

PR #1129, opened 2017-05-15
Improve some compiler messages, logs, warnings

mstoodle:
Sundry small improvements to messages, log files, and compiler warnings. Only 2 could be considered functional issues: 1) an assertion testing the size of something assuming it was a 32-bit integer when it was, in fact, a 16-bit integer, and 2) a "unrecognized" method test using -1 when TR::unknownMethod exists and is zero, not -1.

### DIFF
--- a/compiler/il/NodeExtension.hpp
+++ b/compiler/il/NodeExtension.hpp
@@ -58,8 +58,7 @@ public:
     template <class T>
     T  setElem(uint16_t index, T elem)
        {
-       (T)(_data[index] = (uintptr_t)elem);
-       return (T)_data[index];
+       return (T)(_data[index] = (uintptr_t)elem);
        }
 
     template <class T>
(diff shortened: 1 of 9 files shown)


compiler/il/NodeExtension.hpp
```diff
@@ -58,8 +58,7 @@ class NodeExtension
     template <class T>
     T  setElem(uint16_t index, T elem)
        {
-       (T)(_data[index] = (uintptr_t)elem);
-       return (T)_data[index];
+       return (T)(_data[index] = (uintptr_t)elem);
```
charliegracie: Is this change fixing a compiler warning?

Leonardo2718: I think the compiler warns that `(T)(_data[index] = (uintptr_t)elem);` is an "unused expression" (or something along those lines). Basically, since the resulting value of the expression `_data[index] = (uintptr_t)elem` isn't used, there's no need to cast it to type `T`.

I *think* an equivalent fix would be:

```
_data[index] = (uintptr_t)elem;
return (T)(_data[index]);
```
or even

```
_data[index] = (uintptr_t)elem;
return elem;
```

I have to admit I generally really don't like type casts involving template arguments.

mstoodle: The first statement was causing a compiler warning because the cast on the first line was not being consumed by anything, as @Leonardo2718 stated.

PR #1131, opened 2017-05-16
Do not decrease the ref count of compress sequence of l2aEvaluator

r30shah:
Previously we used to evaluate the shift left child when
l2a node contains compression sequence under Concurrent
Scavenge mode. As this part of code was refactored, we
do not need to decrement reference count of the shift left
child of the tree as we don't evaluate it in this function.

### DIFF
--- a/compiler/z/codegen/UnaryEvaluator.cpp
+++ b/compiler/z/codegen/UnaryEvaluator.cpp
@@ -496,11 +496,6 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
    node->setRegister(source);
 
-
-   if (firstChild->containsCompressionSequence())
-      {
-      cg->decReferenceCount(firstChild->getFirstChild());
-      }
    cg->decReferenceCount(firstChild);
 
    return source;


r30shah:
FYI: @dsouzai , @fjeremic

dsouzai:
This is an oversight on my part in #1090:

```
 -   if (shouldGenerateGuardedLoadsForCompressedLoad && firstChild->containsCompressionSequence())
 +   if (firstChild->containsCompressionSequence())
        {
        cg->decReferenceCount(firstChild->getFirstChild());
        }
```

since `shouldGenerateGuardedLoadsForCompressedLoad` was removed, the entire if statement should have been a NOP.

compiler/z/codegen/UnaryEvaluator.cpp
```diff
@@ -496,11 +496,6 @@ OMR::Z::TreeEvaluator::l2aEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
    node->setRegister(source);
 
-
-   if (firstChild->containsCompressionSequence())
```
vijaysun-omr: Could you please clarify if this is a functional fix ? i.e. what can go wrong if we decrement the reference count of the shift left tree that is not evaluated ?

r30shah: This is a functional fix as I hit the assertions when we attempted to reduce the reference count of child.  I believe we also evaluate the children of firstChild when we call evaluate on it.

vijaysun-omr: I see, thanks.

Issue #793, opened 2017-02-28
Generalize ARM support in the OMR compiler


Leonardo2718:
Currently, ARM support in the OMR compiler is very specific to J9.

This code needs to be generalized so that other projects, like JitBuilder, are able to run on ARM.

Leonardo2718:
The existing arm code generator, evaluators for the following opcodes are *declared but not implemented*:

- `badd`
- `bsub`
- `bmul`
- `bdiv`
- `brem`
- `bneg`
- `bshl`
- `bshr`
- `bushr`
- `band`
- `bor`
- `bxor`
- `bcmpeq`
- `bcmplt`
- `bcmpge`
- `bcmpgt`
- `bcmple`

Leonardo2718:
Tree Simplifier needs to be taught how to handle/fold 8 and 16 bit integers.

knn-k:
Hi, I enabled 4 evaluators (acmp*) for ARM in Issue #2133.

Leonardo2718:
Thanks @knn-k .

When you have PR ready and it gets merged, I will check off the corresponding items in the list.

knn-k:
Hi @Leonardo2718 ,
My fix was merged, and #2133 was closed.

knn-k:
I think I can implement those evaluators for byte and short types.   How can I test them?

Leonardo2718:
Greate @knn-k!

The current way of testing evaluators is to write Tril tests in `fvtest/compilertriltest`. You should be able to use [fvtest/compilertriltest/ArithmeticTest.cpp](https://github.com/eclipse/omr/blob/master/fvtest/compilertriltest/ArithmeticTest.cpp) as an example.

I haven't had the chance to build Tril on ARM yet, so let me know if you run into problems.

PR #1132, opened 2017-05-16
ARM system linkage improvements

Leonardo2718:
This series of commits improve/fix the ARM code generator in the OMR compiler. The main goal of this change is to complete the system linkage implementation. System linkage is the part of the code generator responsible for generating the instruction sequences at function call boundaries to satisfy the platform ABI.

These commits can be roughly grouped into 3 sections:

1. The first 6 commits do some necessary clean up to enable the other commits.
2. The next 4 commits constitute the main changes. These generalize parts of the ARM binary encoding and add missing functionality to the system linkage implementation.
3. The last 2 commits add a test case for the ARM system linkage and some trace in the code generator that was useful when debugging the previous changes. The test case is intended to exercises some edge cases in the system linkage implementation.

I am marking as WIP so that the changes can be reviewed while I run some more tests.

Issue #793

### DIFF
--- a/compiler/arm/codegen/ARMSystemLinkage.cpp
+++ b/compiler/arm/codegen/ARMSystemLinkage.cpp
@@ -36,6 +36,11 @@
 #include "il/Node.hpp"
 #include "il/Node_inlines.hpp"
 
+struct UnsupportedParameterType : public virtual TR::CompilationException
+   {
+   virtual const char* what() const throw() { return "Unsupported Parameter Type"; }
+   };
+
 TR::ARMLinkageProperties TR::ARMSystemLinkage::properties =
     {                           // TR_System
     IntegersInRegisters|        // linkage properties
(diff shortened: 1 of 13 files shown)


compiler/arm/codegen/ARMSystemLinkage.cpp
```diff
@@ -167,156 +172,270 @@ TR::ARMLinkageProperties TR::ARMSystemLinkage::properties =
...
+   reg = machine->getARMRealRegister(TR::RealRegister::gr13);
+   reg->setState(TR::RealRegister::Locked);
+   reg->setAssignedRegister(reg);
+
+   // make r12 (IP) unavailable for RA
+   reg = machine->getARMRealRegister(TR::RealRegister::gr12);
+   reg->setState(TR::RealRegister::Locked);
+   reg->setAssignedRegister(reg);
+
+   // make r9 unavailable for RA (just in case, because it's meaning is platform defined)
```
0xdaryl: In the prologue/epilogue code below, r9 is treated as a preserved register.  If that's the case, I'm wondering why it isn't eligible for register assignment?

compiler/arm/codegen/ARMSystemLinkage.cpp
```diff
@@ -167,156 +172,270 @@ TR::ARMLinkageProperties TR::ARMSystemLinkage::properties =
...
+               nextFltArgReg += 1;
+               mapSingleParameter(parameter, stackIndex);
+               }
+            else
+               {
+               nextFltArgReg = getProperties().getNumFloatArgRegs() + 1;
+               }
+            break;
+         case TR::Aggregate:
+            TR_ASSERT(false, "Function parameters of aggretate types are not currently supported on ARM.");
```
0xdaryl: aggretate -> aggregate

compiler/arm/codegen/ARMSystemLinkage.cpp
Leonardo2718: According to the [AAPCS](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf) (section 5.1.1), the meaning of r9 is platform dependent. Since it is unclear how the register is used on platforms targeted by downstream projects, I chose to do what seemed safest and made it both preserved and unavailable for register assignment.

compiler/arm/codegen/ARMSystemLinkage.cpp
```diff
@@ -326,12 +445,122 @@ TR::ARMLinkageProperties& TR::ARMSystemLinkage::getProperties()
 
 void TR::ARMSystemLinkage::createPrologue(TR::Instruction *cursor)
    {
-   TR_ASSERT(0, "unimplemented");
+   TR::CodeGenerator   *codeGen    = cg();
+   const TR::ARMLinkageProperties& properties = getProperties();
+   TR::Machine         *machine    = codeGen->machine();
+   TR::ResolvedMethodSymbol* bodySymbol = comp()->getJittedMethodSymbol();
+   TR::Node                 *firstNode  = comp()->getStartTree()->getNode();
+   TR::RealRegister    *stackPtr   = machine->getARMRealRegister(properties.getStackPointerRegister());
```
Leonardo2718: This code actually came from a downstream project. I will fix.

Leonardo2718:
@0xdaryl have I addressed all your concerns?

compiler/arm/codegen/ARMSystemLinkage.cpp
lmaisons: >the meaning of r9 is platform dependent

Does this imply that there are asynchronous contexts where, despite using r9, we could be violating the expectation of the platform?

I'm having a hard time picturing the scenario that intersects both "requires preservation" and "don't touch this"

Leonardo2718: Not sure.

I'm open to changing this if someone can tell me which is the preferred alternative (between preserved vs unavailable for register assignment).

0xdaryl: Not being an ARM environment expert, is there any way to determine how r9 is intended to be used (e.g., build environment variable, runtime query, etc.) based on the "platform" (in ARM parlance) we're running in?  It would be a shame to default to the most conservative position (preserved and not available for RA) all the time when I suspect it actually is available in most cases.

What does the C runtime expect?

Leonardo2718: After researching the supported platform, there does not appear to be any straightforward documentation of how r9 is used. I don't really see any solution being viable without more detailed information about the target platforms, so I still think the existing implementation is safest for the moment.

PR #1133, opened 2017-05-16
Move logic querying getOption from jitPreProcess

fjeremic:
The option flags are not initialized in the jitPreProcess function. As
such any logic querying an option flag will return its default value set
at initialization. This is an undesired effect. This logic is now moved
to jitPostProcess where we are guaranteed to have initialized option
bits.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -2314,7 +2314,6 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
    if (self()->requiresLogFile() || self()->getBreakOnCreate() || self()->getDebugOnCreate())
       _addressToEnumerate |= TR_EnumerateRegister;
 
-
    if (self()->getOption(TR_ImmediateCountingRecompilation))
       self()->setOption(TR_EnableGCRPatching, false);
 
@@ -2422,10 +2421,6 @@ OMR::Options::jitPreProcess()
    // --------------------------------------------------------------------------
    // All projects
    //
-   if (self()->getOption(TR_EnableLargePages))
-      {
-      self()->setOption(TR_EnableLargeCodePages);
-      }
 
    //Disabling Shrink Wrapping on all platforms (functional issues)
    self()->setOption(TR_DisableShrinkWrapping);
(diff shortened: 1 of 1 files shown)


fjeremic:
@mpirvu could you please review?

compiler/control/OMROptions.cpp
```diff
@@ -2328,6 +2332,23 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
...
+   // 2GB bar. When RMODE(64) is enabled however we are able to allocate executable memory above the 2GB bar. This
+   // means two code caches could be very far apart from one another which is something we want to avoid as it may
+   // cause performance issues. The solution is to enable the code cache consolidation which will allocate larger
+   // (~256MB) code caches at startup. The code caches will then have locality which improves performance, but
+   // moreover the size of the allocated larger code cache will ensure no trampolines are needed in JIT private
+   // linkage.
+   if (self()->getOption(TR_EnableRMODE64))
+#endif
+      {
+      self()->setOption(TR_EnableCodeCacheConsolidation);
```
mpirvu: This will not work. TR_EnableCodeCacheConsolidation is used in onLoadInternal() when the code caches are created. At that point the command line options have been processed. However, jitLatePostProcess() is done at a later stage in aboutToBootstrap(). I suggest to move this code one phase earlier in jitPostProcess

fjeremic: Fixed in 3272dad.

mpirvu: code looks good

mstoodle:
The commit message should be updated to reflect the change @mpirvu requested.

compiler/control/OMROptions.cpp
```diff
@@ -3681,6 +3659,28 @@ OMR::Options::jitPostProcess()
    if (_hotMaxStaticPICSlots < 0)
       _hotMaxStaticPICSlots = -_hotMaxStaticPICSlots * _maxStaticPICSlots;
 
+   if (self()->getOption(TR_EnableLargePages))
+      {
+      self()->setOption(TR_EnableLargeCodePages);
+      }
+
+#if defined(TR_HOST_64BIT)
```
mstoodle: I realize this was in the original code as well, but shouldn't this be TR_TARGET_64BIT ?

fjeremic: Yes, you are correct. We are talking about the target here. Fixed in 7a7678e. I have also updated the commit message and the PR description. Thanks for the review!

mstoodle:
Thanks @fjeremic looks good now. Once CI tests finish I'll merge (feel free to poke me if you notice it finishing before I do :) ).

PR #1135, opened 2017-05-16
Fix compiler CPU usage regression

lmaisons:
Restore compiler CPU usage that was regressed when changing the implementation of the container used by CFG nodes to keep track of incoming and outgoing edges.

### DIFF
--- a/compiler/env/ExceptionTable.cpp
+++ b/compiler/env/ExceptionTable.cpp
@@ -83,9 +83,17 @@ TR_ExceptionTableEntryIterator::TR_ExceptionTableEntryIterator(TR::Compilation *
                bool found = false;
                TR_ASSERT(tt->getNode(), "a treetop doesn't have a node ????");
                TR::Block * prevBlock = tt->getNode()->getBlock();
-               for (auto e = exceptionPredecessors.begin(); e != exceptionPredecessors.end(); ++e)
-                  if (toBlock((*e)->getFrom()) == prevBlock)
-                     { exceptionPredecessors.erase(e); found = true; break; }
+               for (auto previousEdge = exceptionPredecessors.before_begin(), currentEdge = exceptionPredecessors.begin();
+                    currentEdge != exceptionPredecessors.end();
+                    ++previousEdge, ++currentEdge)
+                  {
+                  if (toBlock((*currentEdge)->getFrom()) == prevBlock)
+                     {
+                     currentEdge = exceptionPredecessors.erase_after(previousEdge);
+                     found = true;
+                     break;
+                     }
+                  }
                if (!found) break;
                first = prevBlock;
                addSnippetRanges(tableEntries, prevBlock, catchBlock, catchType, method, comp);
@@ -99,9 +107,17 @@ TR_ExceptionTableEntryIterator::TR_ExceptionTableEntryIterator(TR::Compilation *
                if (!tt) break;
                bool found = false;
                TR::Block * nextBlock = tt->getNode()->getBlock();
-               for (auto e = exceptionPredecessors.begin(); e != exceptionPredecessors.end(); ++e)
-                  if (toBlock((*e)->getFrom()) == nextBlock)
-                     { exceptionPredecessors.erase(e); found = true; break; }
+               for (auto previousEdge = exceptionPredecessors.before_begin(), currentEdge = exceptionPredecessors.begin();
+                    currentEdge != exceptionPredecessors.end();
+                    ++previousEdge, ++currentEdge)
+                  {
+                  if (toBlock((*currentEdge)->getFrom()) == nextBlock)
+                     {
+                     exceptionPredecessors.erase_after(previousEdge);
+                     found = true;
+                     break;
+                     }
+                  }
                if (!found) break;
                last = nextBlock;
                addSnippetRanges(tableEntries, nextBlock, catchBlock, catchType, method, comp);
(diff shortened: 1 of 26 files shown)


lmaisons:
@mstoodle @0xdaryl @vijaysun-omr Here is [what is hopefully] the final version.

vijaysun-omr:
Thank you very much for all your work on this @lmaisons

andrewcraik:
Looks like a good set of changes - I know there has been a lot of testing. I think it would be helpful to have a bit of documentation about the new API and how it should be used, but it shouldn't hold up this code being accepted since it addresses a number of issues with the current list implementation.

compiler/infra/ForwardList.hpp
```diff
@@ -0,0 +1,684 @@
+#ifndef FORWARD_LIST_HPP
```
0xdaryl: This file requires a copyright header.

vijaysun-omr:
@mpirvu could you please review ?

compiler/infra/ForwardList.hpp
```diff
@@ -0,0 +1,684 @@
...
+#define FORWARD_LIST_HPP
+
+#pragma once
+
+#if 0
```
0xdaryl: General comment throughout this new file:  I am loathe to introduce large amounts of code that are disabled or commented out without justification.  Are these just debugging artifacts that haven't been cleaned up, future extensions, early experimental implementations, etc.?  I'd rather they not be introduced as I'm doubtful they will ever be cleaned up.

0xdaryl: At the very least there should be some Doxygen-style header on this file that describes what this data structure is and what its intended use is over and above the C++ std forward_list.

compiler/infra/ForwardList.hpp
0xdaryl: Previous extensions of std structures in OMR used a lower-case name matching the std equivalent (e.g., infra/deque.hpp).  Please rename this file to `forward_list.hpp` for consistency with that precedent.

`TRlist.hpp` is yet another way of naming an extending class that I'm not too comfortable with, and I don't think that naming is the precedent that should be used here.

lmaisons: Oops.

lmaisons: I didn't realize deque.hpp had been lower-cased.  The previous example that I followed was ReferenceWrapper.hpp do we want to open an issue to rename it to reference_wrapper.hpp?

compiler/infra/ForwardList.hpp
lmaisons: Contained within the `#if 0` section is what is required to implement forward_list as an extension to std::forward_list, and is, ideally, what we would switch to once all our target build environments support using std::forward_list.

compiler/infra/ForwardList.hpp
0xdaryl: Is ReferenceWrapper an extension of std, or simply a wrapper around it?

I don't really have a preference for naming, I just want it consistent for these kinds of extensions.  If I had to pick, I'd go with lowercase like deque.hpp introduced.

mpirvu:
There is a lot of new code and thus easy to miss something. As far as I can tell the code looks good.
I second Daryl's comment that we need to document how the new forward_list is different that the one in STL.

compiler/infra/ForwardList.hpp
lmaisons: TR::reference_wrapper is a re-implementation of std::reference_wrapper [I.E. it should be drop-in replaceable with std::reference_wrapper] that is sufficient for our current requirements.

compiler/infra/ForwardList.hpp
mpirvu: Can we actually extend STL containers? I remember reading somewhere that STL containers do not have virtual destructors and therefore should not be extended.

lmaisons: I believe it's safe to extend STL containers as long as they destroyed using the most derived type.

lmaisons: Is this close to what you had in mind:
```
/**
 * @class TR::forward_list
 *
 * @brief A partial re-implementation of std::forward_list for C++03 supplemented by a linear-time size method.
 *
 * std::forward_list is a container introduced in the 2011 revision of the C++ standard for which
 * the objective is to provide the performance of a singly-linked list. Because the continer is singly-linked,
 * the interfaces common to other containers, where manipulations are performed on the half-closed interval
 * [begin, end), are problematic. To provide a work-around, std::forward_list provides similar operations that
 * operate on the open interval (before_begin, end), and provide [operation]_after variants that perform the
 * requested operation on the position immediately following the provided iterator.
 *
 * A further result of the strict requirement of being equally performant to a manually maintained singly-linked
 * list, is that std::forward_list does not provide a size method, as it cannot provide one that operates in O(1)
 * time. However, there are a number of places in our code base that depend on being able to request the size of
 * a container, where the container is one targetted for re-implementation as a forward_list. To reconcile these
 * requirements, we provide an extension to std::forward_list that provides a size() operation that runs in time
 * linear to the number of objects stored in the container.
 *
 */
```

mpirvu: Right. This means to never hold a pointer to the base class. I don't know how to enforce that.
I am also thinking that if STL containers were never meant to be extended, future implementations may add the keyword 'final' to the class making the extension impossible.

lmaisons: While my first comment addresses the portion of code that is disabled via `#if 0`, there is also the remaining C++03 compatible method definitions from std::forward_list which I skipped implementing, instead commenting out their declarations entirely. Is the preference to remove these?

lmaisons:
Updated commits with changes that [I think] address all issues modulo the commented out methods.

vijaysun-omr:
Comments @0xdaryl and @mpirvu regarding the updates from Luc ?

mpirvu:
I am ok with the changes.

vijaysun-omr:
I asked Daryl for his approval verbally and so I am merging.

PR #1138, opened 2017-05-17
Disable markAncestors

zl-wang:
in RedundantAsyncCheckRemoval.cpp, that is.

It is found out that 100s of asyncCheck are inserted hurting the performance.

### DIFF
--- a/compiler/optimizer/RedundantAsyncCheckRemoval.cpp
+++ b/compiler/optimizer/RedundantAsyncCheckRemoval.cpp
@@ -1310,6 +1310,8 @@ void TR_RedundantAsyncCheckRemoval::getNearestAncestors(TR_StructureSubGraphNode
 
 void TR_RedundantAsyncCheckRemoval::markAncestors(TR_StructureSubGraphNode *node, TR_StructureSubGraphNode *entry)
    {
+   return;  // Disable it for performance. For more details, refer to https://github.com/eclipse/omr/pull/1138
+
    if (node == entry)
       return;
 
@@ -1318,6 +1320,8 @@ void TR_RedundantAsyncCheckRemoval::markAncestors(TR_StructureSubGraphNode *node
 
    node->setVisitCount(comp()->getVisitCount());
 
+   if (trace())
+      traceMsg(comp(),"<===markAncestors start=== ssg node: %d, ssg entry: %d\n", node->getNumber(), entry->getNumber());
 
    for (auto edge = node->getPredecessors().begin(); edge != node->getPredecessors().end(); ++edge)
       {
@@ -1355,6 +1359,9 @@ void TR_RedundantAsyncCheckRemoval::markAncestors(TR_StructureSubGraphNode *node
             _ancestors.add(pred);
 	    }
          }
+
+      if (trace())
+         traceMsg(comp(),"<===markAncestors recursion=== ssg pred: %d, ssg entry: %d\n", pred->getNumber(), entry->getNumber());
 
       markAncestors(pred, entry);
       }


compiler/optimizer/RedundantAsyncCheckRemoval.cpp
```diff
@@ -1310,6 +1310,8 @@ void TR_RedundantAsyncCheckRemoval::getNearestAncestors(TR_StructureSubGraphNode
 
 void TR_RedundantAsyncCheckRemoval::markAncestors(TR_StructureSubGraphNode *node, TR_StructureSubGraphNode *entry)
    {
+   return;  // Disable it for Spark performance
```
fjeremic: This doesn't seem like a good way of disabling something. I also think a bigger comment here is warranted as to why this is disabled and there should be a way to re-enable the feature in case this has undesirable side-effects elsewhere. Perhaps an envvar could be better here?

zl-wang:
It appeared rough, but this markAncestors didn't help performance if any.  As Vijay explained to me, the 100s asyncChecks were inserted because of the code structure:  one if-then-else is inside another, and the embedding depth is 100s. After markAncestors marked these ancestors, asyncCheck was eventually inserted without removing them.  In one previous log, two dozens were removed and more than 300 were inserted.  In a recent log, 10 were removed and 189 inserted.

In any case, removal appeared not dependent on markAncestors.  So, this simple fix, although it may appear rough.

vijaysun-omr:
I think we should do some performance evaluation at the very least before we make this change.

zl-wang:
I am collecting performance data for various workloads.  TPC-DS is only helped so far.  The relevant queries see ~10% range.  Others will come later.

For the specific method (10 removed, 189 inserted), I also logged to see with this fix:  10 removed, 6 inserted.  At least, it appeared relatively "sane".

liqunl:
There is no measurable regression on daytrader.

zl-wang:
No regression neither benefit on POWER with SPECjbb2015 and DayTrader

zl-wang:
No regression on Spark HiBench on X86
looks like 1% benefit for Terasort on X86

vijaysun-omr:
To answer @0xdaryl and @fjeremic I believe markAncestors is misbehaving in the case Julian mentioned. It was very apparent in that case was because there was a very deeply nested if-then-else structure and there were 100s of async checks being added. 

Redundant async check removal works by removing all async checks from loops that contained them and then doing an analysis to place async checks optimally (taking into account calls and inner loops) in that loop such that we have at least one yield point (async check, call or inner loop with an async check) on every path through the loop. As such, it should never be adding multiple async checks into the same loop where more than one of those async checks could be executed on a given control flow path through the loop. I cannot see how markAncestors as implemented avoids this; it seems to be capable of visiting a predecessor P of a block B that is only partially covered (i.e. does not have an async check on every path down stream of the async check before the loop back edge) and adding in an async check in that predecessor P without considering the fact that the reason it was partially covered was because block B was partially covered and if we place an async check in block B then that would cover the control flow paths down stream of block B and therefore not need to an async check in predecessor P. 

Given the time constraints around implementing the proper fix in time for a down stream project deadline, this is a somewhat temporary conservative measure (not running the markAncestors logic will mean we simply have an async check placement similar to before the opt ran, namely that there is an async check before the backwards branch(es) in the loop). A proper fix for the problem will need more elaborate functional testing (because a missing async check can cause weird problems like a thread getting stuck in an infinite loop without a yield point)  and performance testing and will be delivered once that has occurred.

0xdaryl:
Thanks for the detailed explanation @vijaysun-omr .  At the very least I'd say the comment where the code is short-circuited should reference this PR.

PR #1141, opened 2017-05-17
Fix incorrect CPU detection for Linux 32-bit

amitin:
Fix incorrect host machine CPU detection for Linux 32-bit.
Fix typo for cmake variable.


andrewcraik:
You need to sign off the commit for IP validation to pass.

PR #1146, opened 2017-05-18
JitBuilder small performance improvements

mstoodle:
A sequence of small performance improvements for JitBuilder, mostly focused on prolog/epilog corner cases for small methods, but there is also a change to the JitBuilder optimization strategy now that the initial IL quality has been improved via #1110.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -972,6 +972,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
         TR::Options::set32BitNumeric, offsetof(OMR::Options,_numRestrictedGPRs), 0, "F%d"},
    {"oldDataCacheImplementation",         "I\trevert to old data cache implementation.", SET_OPTION_BIT(TR_OldDataCacheImplementation),"F", NOT_IN_SUBSET},
    {"oldJVMPI",           "D\told way of determining which jit options to use with JVMPI",    SET_OPTION_BIT(TR_OldJVMPI), "P" },
+   {"omitFramePointer",         "I\tdo not dedicate a frame pointer register for X86 system linkage.", SET_OPTION_BIT(TR_OmitFramePointer),"F", NOT_IN_SUBSET},
    {"onlyInline=",                        "O{regex}\tlist of methods that can be inlined",
                                           TR::Options::setRegex, offsetof(OMR::Options, _onlyInline), 0, "P"},
    {"optDetails",         "L\tlog all optimizer transformations",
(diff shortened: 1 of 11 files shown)


mstoodle:
Assigning @0xdaryl since bulk of the changes are in x86 codegen.

charliegracie:
Are there any issues with these changes from a compiler point of view? They look ok to me from a JitBuilder point of view.

charliegracie:
@mstoodle is there a way to request a method be compiled using a particular strategy (ie cold vs warm)?

mstoodle:
@charliegracie not yet, but it's not terribly hard to do. I created #1187 to cover it.

compiler/control/OMROptions.cpp
```diff
@@ -972,6 +972,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
         TR::Options::set32BitNumeric, offsetof(OMR::Options,_numRestrictedGPRs), 0, "F%d"},
    {"oldDataCacheImplementation",         "I\trevert to old data cache implementation.", SET_OPTION_BIT(TR_OldDataCacheImplementation),"F", NOT_IN_SUBSET},
    {"oldJVMPI",           "D\told way of determining which jit options to use with JVMPI",    SET_OPTION_BIT(TR_OldJVMPI), "P" },
+   {"omitFramePointer",         "I\tdo not dedicate a frame pointer register for X86 system linkage.", SET_OPTION_BIT(TR_OmitFramePointer),"F", NOT_IN_SUBSET},
```
0xdaryl: I don't think this option needs to be specific to "x86" (or "system linkage" for that matter).

mstoodle: That's true, but the implementation is currently x86 system linkage specific and will have no effect on any other platform, so I felt it was appropriate to make that clear in the option description. I refrained from putting it in the actual name of the option because I agree it doesn't have to be so specific.

compiler/x/codegen/X86SystemLinkage.cpp
```diff
@@ -429,7 +429,7 @@ TR::X86SystemLinkage::createPrologue(TR::Instruction *cursor)
 
    const TR::X86LinkageProperties &properties = getProperties();
 
-   const uint32_t outgoingArgSize = cg()->getLargestOutgoingArgSize();
+   const uint32_t outgoingArgSize = (cg()->isLeafMethod()) ? 0 : cg()->getLargestOutgoingArgSize();
```
mstoodle: heh, good point. I'll review that code to see if I can reconstruct why I decided to solve it here.

mstoodle: Reviewing the code that sets `_largestOutgoingArgSize`, I think it's just pedantic caution. Do you want me to remove it?

0xdaryl: My reading of the code wherever `_largestOutgoingArgSize` is set leads me to believe it should only be non-zero when a call with arguments is present (and hence this shouldn't be a leaf method).  I would prefer it being removed if not necessary as it might add confusion in the future.

mstoodle: fair enough, i'll update it.

mstoodle: @0xdaryl update pushed.

PR #1147, opened 2017-05-18
First zTPF Pull to OMR

jjohnst-us:
Description:
This pull represents a learning effort of committing omr changes for the z/TPF platform.   These four files are new copies (based on the other platforms) and reside in the new port/ztpf folder and will not affect any other platforms.  A successful personal build was done in the IBM environment (build_id=348248), as would be expected because only the z/TPF platform has the capability of building its unique environment.

The files were randomly chosen and deal with dump processing and virtual memory.  A larger pull request is in the works for a later date.

### DIFF
--- a/port/ztpf/omrosdump.c
+++ b/port/ztpf/omrosdump.c
@@ -1,0 +1,362 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 1991, 2017
+ *
+ *  This program and the accompanying materials are made available
+ *  under the terms of the Eclipse Public License v1.0 and
+ *  Apache License v2.0 which accompanies this distribution.
+ *
+ *      The Eclipse Public License is available at
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial API and implementation and/or initial documentation
+ *    Multiple authors (IBM Corp.) - refactoring and modifications for z/TPF platform
+ *******************************************************************************/
+
+/**
+ * \file
+ * \ingroup Port
+ * \brief Dump formatting externals referenced from portLibrary
+ */
+
+#define ZTPF_NEEDS_GETCONTEXT
+
+#include <signal.h>
+#include <tpf/c_proc.h>
+#include <tpf/c_cinfc.h>
+#include <tpf/tpfapi.h>
+#include <tpf/sysapi.h>
+#include <tpf/c_idsprg.h>
+#include <tpf/c_stck.h>
+#include <sys/ucontext.h>
+#include <pthread.h>
+#define OSDUMP_SYMBOLS_ONLY
+
+#include <sys/types.h>
+#include <stdint.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include <errno.h>
+#include <tpf/c_eb0eb.h>
+#include <tpf/c_proc.h>
+#include "omrport.h"
+#include "portpriv.h"
+#include "omrosdump_helpers.h"
+#include "omrsignal.h"
+#include <signal.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include <libgen.h>
+
+typedef struct iproc IPROC;
+#include "safe_storage.h"
+
+/**
+ * \fn Callback from CCCPSE: preemptible error.
+ *
+ * The z/TPF kernel will call us back at this function when it (the z/TPF
+ * program interrupt handler) determines that the error is 'preemptible'
+ * (i.e. is a 'soft' interrupt which is caused by division by zero or a
+ * null-pointer reference).
+ *
+ */
+static void __attribute__((used))
+ztpf_preemptible_err_ep(void)
+{
+	// let the port library do the work.
+	ztpf_preemptible_helper();
+}
+
+/**
+ * \fn Callback from CCCPSE: non-preemptible error.
+ *
+ * The z/TPF kernel will call us back at this function when it (the z/TPF
+ * program interrupt handler) determines that the error is 'non-preemptible'
+ * (i.e. is a 'hard' interrupt) caused by any number of programming errors
+ * considered serious enough to question the state of the JVM.
+ */
+static void
+__attribute__((used))
+ztpf_nonpreemptible_err_ep(void)
+{
+	// let the port library do the work.
+	ztpf_nonpreemptible_helper();
+}
+
+/**
+ * \fn  Create an ELF core dump file of the user space
+ *
+ *
+ *        This function determines if there is a Java Dump Buffer available pertinent to this
+ *        thread/process. If one is not, we know we were called from the system dump agent's
+ *        executing a hook (such as -Xdump:system:events=vmstop) and arrange for one to be
+ *        created right now. If a dump pertinent to us is available, we use it. We then call
+ *        ztpfBuildCoreDump() in order to build an ELF core dump from its contents, then
+ *        write it to file.
+ *
+ *        After we're done writing the file, zeroize the Java Dump Buffer pointer in the
+ *        DIB block.
+ *
+ * \param[in] portLibrary       The port library.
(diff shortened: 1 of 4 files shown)


port/ztpf/omrosdump.c
```diff
@@ -0,0 +1,343 @@
...
+ * \return 0 on success, non-zero otherwise.
+ *
+ * \note <TT>filename</TT> buffer can not be NULL.
+ * \note User must allocate and free <TT>filename</TT> buffer.
+ * \note <TT>filename</TT> buffer length is platform dependent, assumed to be <TT>EsMaxPath</TT> or <TT>MAX_PATH</TT>
+ * \note If <TT>filename</TT> buffer is empty, an error will be returned. Should it
+ *        contain only a filename (no path), one will be chosen for it.
+ */
+UDATA
+j9dump_create(struct J9PortLibrary *portLibrary, char *filename, char *dumpType, void *userData)
```
fjeremic: We seem to be introducing quite a lot of J9 and Java related code in this commit. I'm not sure this belongs at the OMR layer. Can this be abstracted to be used with OMR? And if not shouldn't this code belong in an upstream project consuming OMR?

port/ztpf/omrosdump.c
```diff
@@ -0,0 +1,343 @@
...
+{
+        U_8     workspace[PATH_MAX];
+        U_8     *errMsg = "";
+        U_8 *JVMfn = NULL;
+    U_8 holdkey =0;
```
fjeremic: Please fix spacing inconsistencies here and elsewhere in the commit. See other obvious spots in comments and code below.

jjohnst-us:
Waiting on some legal discussions before signing ECA.  I made copyright header updates, but haven't committed them yet.  Still need to fix spacing issues.

Not sure why you mentioned these might not be OMR layer updates... These segments were copied from other platforms from the OMR repo.   Either from port/common or port/linux.  Could you elaborate?  Thanks.

port/ztpf/omrosdump.c
jjohnst-us: @fjeremic
Could you elaborate and provide an example of functions which should be abstracted out?  The original location of this file in the original IBM repo was under omr.  Should I be renaming these functions to omrdump_create instead of j9dump_create?
Could you explain then what OMR should contain?  For example the z/OS platform has a omrosdump.c file with an omrdump_create?

Much appreciated.

port/ztpf/omrosdump.c
jjohnst-us: I used the source formatter from eclipse to fix spacing you are referring too.  I need to configure the source formatter to match conventions for omr.

jjohnst-us:
@Leonardo2718 
@fjeremic 

Hi, I cleaned up these segments to use the omr labeling convention based on other segments I saw from other platforms.  I was able to get the IP stuff cleared and submitted.

Once you accept this I can do the rest of them in similar fashion.  Remember, our platform is small and currently we are the only ones able to build this environment.  Thanks for taking a look.

port/ztpf/omrosdump.c
```diff
@@ -0,0 +1,357 @@
...
+omrdump_create(struct OMRPortLibrary *portLibrary, char *filename, char *dumpType, void *userData)
+{
+	uint8_t workspace[PATH_MAX];
+	uint8_t *errMsg = "";
+	uint8_t *JVMfn = NULL;
+	uint8_t holdkey = 0;
+
+	DIB *pDIB = NULL;
+	DBFHDR *pDBF = NULL;
+	safeStorage *s = allocateDurableStorage();
```
charliegracie: Can this call fail? Does this require a NULL check?

port/ztpf/omrosdump_helpers.h
```diff
@@ -0,0 +1,292 @@
...
+
+/* This data structure is the argument block whose address we need to pass to
+ * ztpfBuildCoreFile() as its one and only parameter. Function
+ * translateInterruptContexts() also uses it since it requires so much of the
+ * same data.
+ *
+ * There are ten (10) data members in this block, it is 0x50 bytes long.
+ */
+typedef struct {
+   siginfo_t				  *sii;			  /* argv[0] = ptr to siginfo_t				*/
```
charliegracie: This code is ok for this PR but in the future I would prefer if the field names were not aligned as small changes force them to become unaligned like below.  It is better to just have 1 space between the type and the name.

port/ztpf/omrosdump_helpers.h
```diff
@@ -0,0 +1,292 @@
...
+   int32_t					  rc;			  /* argv[3] = return code area				*/
+   uint32_t					  flags;		  /* argv[5] = 32 bits of output flags		*/
+   uint32_t					  wkspcSize;	  /* argv[6] = How big is argv[7]?			*/
+   uint8_t					  *wkSpace;		  /* argv[7] = ptr to workspace				*/
+   struct J9PortLibrary		  *portLibrary;   /* argv[8] = ptr to J9PortLibrary			*/
+   DIB						  *dibPtr;		  /* argv[9] = ptr to Dump Interchange Blk.	*/
+} args;
+
+/* Argument pointer Accessors */
+#define dumpSiginfo(p)	   (p->sii)
```
charliegracie: Similar comment to my previous one

port/ztpf/omrosdump_helpers.h
```diff
@@ -0,0 +1,292 @@
...
+typedef struct fake_auxv {
+   uint8_t		fake_data[304];		/* Gotta go with 304 zeros here.			*/
+} AUXV;
+
+typedef struct { 
+   uint64_t		hi_word;
+   uint64_t		lo_word;
+} psw_t;
+
+//
```
charliegracie: Is there any reason to keep the commented out code?

port/ztpf/omrvmem.c
```diff
@@ -0,0 +1,1994 @@
...
+	/* clear page info data, this has the effect of starting off in a standard state */
+	memset(&vmem_page_info, 0x00, sizeof(vmem_hugepage_info_t));
+	get_hugepages_info(portLibrary, &vmem_page_info);
+
+	/* 0 terminate the table */
+	memset(PPG_vmem_pageSize, 0, OMRPORT_VMEM_PAGESIZE_COUNT * sizeof(uintptr_t));
+	memset(PPG_vmem_pageFlags, 0, OMRPORT_VMEM_PAGESIZE_COUNT * sizeof(uintptr_t));
+
+	/* First the default page size */
+#if defined(J9ZTPF)
```
charliegracie: What is the build flag used for? Isn't all of the code in this directory ztpf code?

Also would it be possible to rename this flag OMRZPTF?

port/ztpf/omrosdump.c
jjohnst-us: I checked for NULL just in case.

port/ztpf/omrosdump_helpers.h
jjohnst-us: Sure thing, I cleaned some of it up.  Going forward will just leave one space.

port/ztpf/omrosdump_helpers.h
jjohnst-us: Removed spaces, going forward will just leave one space.

port/ztpf/omrosdump_helpers.h
jjohnst-us: Removed extra code.

jjohnst-us: Removed J9ZTPF guard macros and code that should never get built.  We will use the OMRZTPF convention per your suggestion for other segments.

port/ztpf/omrvmem.c
jjohnst-us: Removed J9ZTPF guard macros and code that should never get built. We will use the OMRZTPF convention per your suggestion for other segments.

port/ztpf/omrosdump_helpers.h
jjohnst-us: On a side note, I used the eclipse code formatter, I forget if it added that alignment between type and variable name.

jjohnst-us:
@charliegracie Let me know if there are any other updates you would like to see, otherwise this should be a low risk merge.  Thanks for the review.

lmaisons:
Just a heads-up that the files introduced by this PR seem to be in CR+LF  format rather than the LF format we [seem] to use everywhere else.

PR #1149, opened 2017-05-19
Removed undocumented branch conditions in omr

arianneb:
Replaced branch conditions not documented in the Z Architecture Reference Summary / Principles of Operation with their corresponding mask values. Removed same branch conditions from enumerations and branch to name map. 


mstoodle:
@arianneb your ci build seems to have been "cancelled". Was that on purpose?

@fjeremic would you please review here?

PR #1150, opened 2017-05-19
Add option disableProfilingDataReclamation

ojsheikh:
Add an option that should to disable the cleanup of memory allocated for profile information, in case an issue arises with profile data cleanup

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -471,6 +471,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableProfileGenerator",            "O\tdisable profile generator",                      TR::Options::disableOptimization, profileGenerator, 0, "P"},
 #endif
    {"disableProfiling",                   "O\tdisable profiling",                              SET_OPTION_BIT(TR_DisableProfiling), "P"},
+   {"disableProfilingDataReclamation",    "O\tdisable reclamation for profiling data",         SET_OPTION_BIT(TR_DisableProfilingDataReclamation), "F", NOT_IN_SUBSET},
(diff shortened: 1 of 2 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -538,6 +538,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableTrivialStoreSinking",         "O\tdisable trivial store sinking", RESET_OPTION_BIT(TR_EnableTrivialStoreSinking), "F"},
    {"disableTrueRegisterModel",           "C\tdisable use of true liveness model in local RA instead of Future Use Count (Dangerous use unless no Global Virtual Regs used)", RESET_OPTION_BIT(TR_EnableTrueRegisterModel), "F"},
    {"disableUncountedUnrolls",            "O\tdisable GLU from unrolling uncoutned loops ",SET_OPTION_BIT(TR_DisableUncountedUnrolls), "F"},
+   {"disableProfilingDataReclamation",    "O\tdisable reclamation for profiling data",         SET_OPTION_BIT(TR_DisableProfilingDataReclamation), "F", NOT_IN_SUBSET},
```
mpirvu: Options need to be added in lexicographic order. In this case, 'disableProfilingDataReclamation' needs to stay after 
   {"disableProfiling",                   "O\tdisable profiling",                              SET_OPTION_BIT(TR_DisableProfiling), "P"},

ojsheikh: Forgot to move the option when I renamed the option, moved.

vijaysun-omr:
Change looks safe to me too. Merging.

PR #1152, opened 2017-05-23
Enforce singleton property in JitBuilder::CodeCacheManager

sajidahmed21:
- Fatal assertion to enforce singleton property in JitBuilder::CodeCacheManager
- Fatal assertion to ensure that JitBuilder::CodeCacheManager::instance() never returns a null pointer

### DIFF
--- a/jitbuilder/runtime/JBCodeCacheManager.hpp
+++ b/jitbuilder/runtime/JBCodeCacheManager.hpp
@@ -58,7 +58,12 @@ public:
 
    void *operator new(size_t s, TR::CodeCacheManager *m) { return m; }
 
-   static TR::CodeCacheManager *instance()  { return _codeCacheManager; }
+   static TR::CodeCacheManager *instance()
+      {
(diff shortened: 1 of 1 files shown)


mstoodle:
unfortunately, the commit message titles both exceed the 50 character limit documented in our contribution guidelines. Would you please shorten them?

sajidahmed21:
@mstoodle is the code cache manager currently being used in multi-threaded environments? If yes, isn't it absolutely necessary to place a lock inside the constructor right away?

mstoodle:
Yes, there are. It just comes down to what is the contract. I would prefer not to penalize single threaded scenarios which just means that, in the multi-threaded case, it's up to the client to do the synchronization (for example, if using JitBuilder, acquire a monitor or mutex around the initializeJit() call).

We can revisit it in future if it's too much of a hassle for consumers.

lmaisons:
>is the code cache manager currently being used in multi-threaded environments

Sort-of.  We are frequently constructing it under contexts that are essentially single-threaded, and subsequently accessing it in contexts that are multi-threaded.

PR #1154, opened 2017-05-23
WIP: Add OSR data for method entry

ncough:
The method entry may become an OSR point, however,
it will not be treated as an OSR point by OSR
liveness and usedef analysis. This change ensures
that it is correctly identified.


vijaysun-omr:
@andrewcraik please review as well.

andrewcraik:
LGTM

vijaysun-omr:
Looks okay to me. The special casing of the implicit OSR point on method entry is needed to make sure we do not incur extra overheads. Merging.

PR #1155, opened 2017-05-23
Allow building with recent XLC versions

AgostonSzepessy:
Changes to enable building with recent versions of XLC on PPCLE Linux.

### DIFF
--- a/compiler/env/jittypes.h
+++ b/compiler/env/jittypes.h
@@ -149,13 +149,13 @@ typedef struct TR_InlinedCallSite
       #define FLUSH_MEMORY(smp)
    #endif
 #else
-   #if (__IBMC__ || __IBMCPP__)
-      void do_lwsync (void);
-      #pragma mc_func do_lwsync {"7c2004ac"}
-      #pragma reg_killed_by do_lwsync
-      #define FLUSH_MEMORY(smp) if( smp ) do_lwsync();
+   #if defined(__IBMC__) || defined(__IBMCPP__)
+      #if defined(__cplusplus)
+      #include <builtins.h>
+      #endif /* __cplusplus */
+      #define FLUSH_MEMORY(smp) if( smp ) __lwsync();
    #elif defined(LINUX)
-      #define FLUSH_MEMORY(smp) if( smp ) __asm__("sync");
+      #define FLUSH_MEMORY(smp) if( smp ) __asm__("lwsync");
    #endif
 #endif
 
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -2249,7 +2249,7 @@ static void assign_candidate_loop_trace_increment(TR::Compilation *comp, TR_Regi
 bool
 TR_RegisterCandidates::assign(TR::Block ** cfgBlocks, int32_t numberOfBlocks, int32_t & lowestNumber, int32_t & highestNumber)
    {
-#if (defined(__IBMCPP__) || defined(__IBMC__))
+#if (defined(__IBMCPP__) || defined(__IBMC__)) && !defined(__ibmxl__)
    // __func__ is not defined for this function on XLC compilers (Notably XLC on Linux PPC and ZOS)
    static const char __func__[] = "TR_RegisterCandidates::assign";
 #endif
--- a/compiler/p/runtime/PPCCodeSync.inc
+++ b/compiler/p/runtime/PPCCodeSync.inc
@@ -22,25 +22,46 @@
 extern "C" void *getCodeSync();
 
 #if defined(TR_HOST_POWER)
-#if defined(__ibmxl__)
-#define dcbf(x)  __dcbf(x)
-#define sync()  __sync()
+#if defined(__IBMC__) || defined(__IBMCPP__)
+
+#if defined(__cplusplus)
+#include <builtins.h>
+#endif /* __cplusplus */
+
+#define dcbf(x) __dcbf(x)
+#define sync() __sync()
 #define isync() __isync()
-#elif (__IBMC__ || __IBMCPP__)
-void dcbf(unsigned char *);
-void icbi(unsigned char *);
-void sync();
-void isync();
-#pragma mc_func dcbf  {"7c0018ac"}
-#pragma mc_func icbi  {"7c001fac"}
-#pragma mc_func sync  {"7c0004ac"}
-#pragma mc_func isync {"4c00012c"}
-#pragma reg_killed_by dcbf
-#pragma reg_killed_by icbi
-#pragma reg_killed_by sync
-#pragma reg_killed_by isync
-#endif
-#endif
+
+#else
+static inline void dcbf(unsigned char *addr)
+   {
+   __asm__(
+      "dcbf 0,%0"
+      : /* no outputs */
+      : "r" (addr) );
+   }
+
+static inline void sync()
+   {
+   __asm__("sync");
+   }
+
+static inline void isync()
+   {
+   __asm__("isync");
+   }
+
+#endif /* __IBMC__ || __IBMCPP__ */
+
+static inline void icbi(unsigned char *addr)
+   {
+   __asm__(
+      "icbi 0,%0"
+      : /* no outputs */
+      : "r" (addr) );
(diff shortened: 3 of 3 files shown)


compiler/env/jittypes.h
```diff
@@ -149,7 +149,9 @@ typedef struct TR_InlinedCallSite
       #define FLUSH_MEMORY(smp)
    #endif
 #else
-   #if (__IBMC__ || __IBMCPP__)
+   #ifdef __ibmxl__
+      #define FLUSH_MEMORY(smp) if ( smp ) __lwsync();
+   #elif (__IBMC__ || __IBMCPP__)
       void do_lwsync (void);
       #pragma mc_func do_lwsync {"7c2004ac"}
       #pragma reg_killed_by do_lwsync
```
ymanton: I think both old and new XL support the `__lwsync()` builtin, so can you try getting rid of the `mc_func` part and using the same define for both?

compiler/p/runtime/PPCCodeSync.inc
```diff
@@ -112,7 +112,7 @@ void ppcCodeSync(unsigned char *codeStart, unsigned int codeSize)
 	for(addr = codeStart ; addr < limit; addr += cacheLineSize)
 	   {
 
-		#if defined(__IBMC__) || defined(__IBMCPP__)
+		#if defined(__IBMC__) || defined(__IBMCPP__) && !defined(LINUX)
 		icbi(addr);
 		#elif defined(LINUX)
 		__asm__(
```
ymanton: I think we can simplify this by defining a suitable `icbi` at the top of this file where similar macros are defined. Can you take a crack at defining everything at the top and using it in the rest of the code and cleaning up this mix and match we have going on here? In general prefer to use builtins when available, then inline assembly for GCC and new XLC, and either inline assembly or `mc_funcs` for old XLC, though I would lean towards inline assembly.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -2249,7 +2249,7 @@ static void assign_candidate_loop_trace_increment(TR::Compilation *comp, TR_Regi
 bool
 TR_RegisterCandidates::assign(TR::Block ** cfgBlocks, int32_t numberOfBlocks, int32_t & lowestNumber, int32_t & highestNumber)
    {
-#if (defined(__IBMCPP__) || defined(__IBMC__))
+#if (defined(__IBMCPP__) || defined(__IBMC__)) && (!defined(LINUXPPC))
```
ymanton: I think you want to check for `!defined(__ibmxl__)` here? (Since whoever put this in was suggesting that it was a workaround for old XLC on Linux and it to continue to work in that case.)

compiler/env/jittypes.h
```diff
@@ -149,11 +149,8 @@ typedef struct TR_InlinedCallSite
...
 #else
-   #if (__IBMC__ || __IBMCPP__)
-      void do_lwsync (void);
-      #pragma mc_func do_lwsync {"7c2004ac"}
-      #pragma reg_killed_by do_lwsync
-      #define FLUSH_MEMORY(smp) if( smp ) do_lwsync();
+   #if defined(__IBMC__) || defined(__IBMCPP__)
+      #define FLUSH_MEMORY(smp) if( smp ) __lwsync();
    #elif defined(LINUX)
       #define FLUSH_MEMORY(smp) if( smp ) __asm__("sync");
```
ymanton: This has nothing to do with XLC, but as long as you're making changes in this code I think we can change this "sync" to "lwsync" to keep the behaviour consistent across all compilers.

compiler/p/runtime/PPCCodeSync.inc
```diff
@@ -24,23 +24,38 @@ extern "C" void *getCodeSync();
...
+	{
+	__asm__("isync");
+	}
+
+#endif /* __ibmxl__ */
+
+void icbi(unsigned char *addr)
+	{
+	__asm__(
+	   "dcbf 0,%0"
```
ymanton: This should be `icbi` not `dcbf`.

AgostonSzepessy:
@ymanton Could you please look over the changes I made?

compiler/p/runtime/PPCCodeSync.inc
```diff
@@ -24,23 +24,38 @@ extern "C" void *getCodeSync();
 #if defined(TR_HOST_POWER)
 #if defined(__ibmxl__)
 #define dcbf(x)  __dcbf(x)
-#define sync()  __sync()
+#define sync() __sync()
 #define isync() __isync()
-#elif (__IBMC__ || __IBMCPP__)
```
ymanton: Both old and new XLC support these 3 as builtins, so we can check for `__IBMC__ || __IBMCPP__` instead of `__ibmxl__` and leave the inline assembly for GCC.

AgostonSzepessy:
@ymanton Could you please review my changes?

compiler/env/jittypes.h
```diff
@@ -149,13 +149,10 @@ typedef struct TR_InlinedCallSite
       #define FLUSH_MEMORY(smp)
    #endif
 #else
-   #if (__IBMC__ || __IBMCPP__)
-      void do_lwsync (void);
-      #pragma mc_func do_lwsync {"7c2004ac"}
-      #pragma reg_killed_by do_lwsync
-      #define FLUSH_MEMORY(smp) if( smp ) do_lwsync();
+   #if defined(__IBMC__) || defined(__IBMCPP__)
```
ymanton: Let's add
```
#if !defined(__ibmxl__)
#include <builtins.h>
#endif /* __ibmxl__ */
```
here as well. It must be getting included implicitly somehow for `__lwsync()` to work on old XLC, but it's better to be explicit.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -2249,7 +2249,7 @@ static void assign_candidate_loop_trace_increment(TR::Compilation *comp, TR_Regi
 bool
 TR_RegisterCandidates::assign(TR::Block ** cfgBlocks, int32_t numberOfBlocks, int32_t & lowestNumber, int32_t & highestNumber)
    {
-#if (defined(__IBMCPP__) || defined(__IBMC__))
+#if (defined(__IBMCPP__) || defined(__IBMC__)) && (!defined(__ibmxl__))
```
ymanton: Get rid of the extra parenthesis around `!defined` to keep it consistent with the rest of the code.

compiler/p/runtime/PPCCodeSync.inc
```diff
@@ -22,25 +22,46 @@
...
+
+void icbi(unsigned char *addr)
+	{
+	__asm__(
+	   "icbi 0,%0"
+	   : /* no outputs */
+	   : "r" (addr) );
+	}
+
+#endif /* TR_HOST_POWER */
```
ymanton: Can you fix the indenting for these functions? Should be soft tabs of 3 spaces. Also, and more importantly, can you declare these functions as `static`?

AgostonSzepessy:
@ymanton Could you please review my changes?

AgostonSzepessy:
@ymanton Could you please review my changes?

ymanton:
I don't see any other issues. Unless @zl-wang has any comments I'm OK with this going in.

zl-wang:
Hmm ... we should take this opportunity to revamp PPCCodeSync.  Since POWER5, iCache has been coherent. We don't need dcbf and icbi line by line.  Instead, we only need one icbi to scoreboard for isync. Furthermore, the second sync is only needed for the reason of G4 processor I believed.  i.e.  For POWER5 and later,  no matter how big the "new" code region is, the CodeSync sequence always looks like:  (only one icbi is enough)

       sync
       icbi
       isync

Since java9 targeted supporting only recent processors (later than POWER5?), we can do the above.

Dragging-behind is the reality that this code base is commonly used in older releases (which might need to support older processors).  Sigh ... we need to revisit this issue when it is divested of older releases.

I hoped you already make sure that those "static" routines (e.g.  dcbf(x)) are inlined. 

Otherwise, this looks fine.

ymanton:
Looks like this is ready, right? I don't see any more issues.

PR #1156, opened 2017-05-23
Turn-off default optimization for constant length array copy

0dvictor:
Disable optimization for constant length array copy by default,
but leave two environment variables to turn it on either fully or
partially.


andrewcraik:
LGTM - @vijaysun-omr can we merge this urgently since there is a bug in the code that is being disabled.

vijaysun-omr:
Just curious, how "temporary" is this disabling ? i.e. do you expect to debug the problem that led to this urgent PR disabling the code and then re-enable ? Or it requires significantly more work (akin to a re-design) ?

vijaysun-omr:
I will wait for checks to  pass before merging in any case.

andrewcraik:
I don't think we know the extent of the problem in functional terms - the analysis is still being done. We have also had a few performance problems so we need to understand what is going on and we may need to revert to the previous implementation if it can't be changed. Regardless, we will either re-enable a fixed version of this or revert the changes in the not too distant future.

vijaysun-omr:
Checks have passed, the change seems to be a conservative one in that it disables some (buggy) code that is "optional" (to get extra performance) and earlier comments suggest this is required with some urgency. Merging.

PR #1157, opened 2017-05-23
Disable OSR opts when OSR is not supported

ncough:
OSR liveness and reaching definitions analysis 
should not be performed if OSR is not possible 
during this compilation. They will waste compile
time and the expected infrastructure will not
be present.


andrewcraik:
Seemingly sensible change to stop running opts that shouldn't be running... LGTM

vijaysun-omr:
Seems like a valid compile time saving change in some scenarios. I will wait for checks to pass.

PR #1159, opened 2017-05-23
Update ftops.pl and associated files

ymanton:
The `ftops.pl` tool used by the Power code generator is out of date and needs some attention. This patch cleans up the code to make it easier for developers to run and maintain, re-generates the artifacts, and removes dead artifacts and code.


ymanton:
`TRSRC` doesn't need to be set anymore, you can run the script from any directory and it will not depend on anything defined in your environment. Instead the script assumes the top level OMR directory can be reached via `../../..` relative to its own directory, so it will need to be adjusted whenever it moves and breaks that relation.

vijaysun-omr:
@zl-wang could you please review ?

zl-wang:
It looked to me that both PPCOps.hpp and PPCOps2.hpp files should  be deleted, instead of checking-in an empty file.

ymanton:
The GitHub UI makes it appear as if only the contents were deleted, but the file `PPCOps.hpp` was actually deleted entirely. I've deleted `PPCOps2.hpp` as well.

Issue #1160, opened 2017-05-24
Fix for unsafe generation of VFMA and VFMS on z


IBMJimmyk:
generateFusedMultiplyAddIfPossible tries to check if it is working on floating point data before generating fused instructions. The problem is caused by the fact that a call to isFloatingPoint on a type returns false
for VectorDouble and VectorFloat types. To address this issue, it is proposed that a specific check for VectorDouble and VectorFloat is added to the function generateFusedMultiplyAddIfPossible.

A more complete solution would be to modify isFloatingPoint to return true for VectorDouble and VectorFloat types. However, isFloatingPoint is used in many locations so a lengthy investigation into potential side effects would be required which is not something I can do at this time.

PR #1161, opened 2017-05-24
Fix for unsafe generation of VFMA and VFMS on z

IBMJimmyk:
generateFusedMultiplyAddIfPossible tries to check if it is working on
floating point data before generating fused instructions. The problem is
caused by the fact that a call to isFloatingPoint on a type returns false
for VectorDouble and VectorFloat types. To address this issue, a specific
check for VectorDouble and VectorFloat was added to the function
generateFusedMultiplyAddIfPossible.

Issue: #1160

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -18264,7 +18264,8 @@ generateFusedMultiplyAddIfPossible(TR::CodeGenerator *cg, TR::Node *addNode, TR:
       }
 
    // if the FloatMAF option is not set and the node is not FP strict, we can't generate FMA operations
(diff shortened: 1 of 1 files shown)


mgaudet:
Is it just me, or does this reduce restriction on generating FMA, not tighten? (It's not clear what the 'problem' is here. To me this reads as an enhancement)

IBMJimmyk:
Sorry, I didn't notice the new comment until now. That section of the code is supposed to return false when FMA is not supposed to be used. My change increases the number of scenarios where false is returned to include cases where the node type is VectorFloat and VectorDouble.

The problem was that using a FMA instead of a multiply then an add instruction won't always meet the Java spec because it skips the rounding between the multiply and the add which is important when handling floating point data.

mgaudet:
Ok. That makes some more sense.

PR #1163, opened 2017-05-24
Small CMake Improvements

mgaudet:
A couple small improvements to the CMakeLists.


youngar:
LGTM
/cc @dnakamura

rwy0717:
I suspect `OMR_GLUE` won't be cached if it is set by the user, which is not ideal.

PR #1164, opened 2017-05-24
 Disable/Sampling Hook trace for frequent events

LinHu2016:
Disable/Sampling Hook trace for frequent events

 - disable/sampling hook tracepoint and hook dump via new tag <trace-sampling intervals="N" /> in hdf for high-frequency events to reduce performance impacts of tracking hooks
 - new invocation count in hook dump structure to trace event frequency
 - atomic operation for updating hook dump information to keep data
consistent in race condition

### DIFF
--- a/util/hookable/hookable.cpp
+++ b/util/hookable/hookable.cpp
@@ -181,6 +181,9 @@ J9HookDispatch(struct J9HookInterface **hookInterface, uintptr_t taggedEventNum,
 	uintptr_t eventNum = taggedEventNum & J9HOOK_EVENT_NUM_MASK;
 	J9CommonHookInterface *commonInterface = (J9CommonHookInterface *)hookInterface;
 	J9HookRecord *record = HOOK_RECORD(commonInterface, eventNum);
+	OMREventInfo4Dump *eventDump = J9HOOK_DUMPINFO(commonInterface, eventNum);
+	uintptr_t samplingInterval = (taggedEventNum & J9HOOK_TAG_SAMPLING_MASK) >> 16;
+	bool sampling = false;
 
 	if (taggedEventNum & J9HOOK_TAG_ONCE) {
 		uint8_t oldFlags;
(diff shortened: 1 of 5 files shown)


charliegracie:
@LinHu2016 can you limit the first line of your commit message to 50 characters please?

LinHu2016:
update first line of commit message(under 50 characters)

0xdaryl:
I'm not sure what to make of your commit message and how to parse it.  Is that a point-form design for how it works?  Could you please re-write it in a more descriptive form that describes what this commit provides, perhaps separating it from how to tune it (which is what I think the latter part of your commit message is getting at).

LinHu2016:
updated commit message and comment for pull request

util/hookable/hookable.cpp
```diff
@@ -213,39 +216,56 @@ J9HookDispatch(struct J9HookInterface **hookInterface, uintptr_t taggedEventNum,
...
-						eventDump->longestHook.startTime = eventDump->lastHook.startTime;
-						eventDump->longestHook.func_ptr = eventDump->lastHook.func_ptr;
-						eventDump->longestHook.duration = eventDump->lastHook.duration;
+
+				if (sampling) {
+					uint64_t timeDelta = omrtime_current_time_millis() - startTime;
+					uint64_t lastStart = 0;
+
+					/* here is not thread safe, keep eventDump data consistency from race condition */
+					do {
```
charliegracie: Can you provide some details as to why this has to be an atomic operation?

LinHu2016: the hook dispatch is not thread safe, the same event could be triggered by multiple threads at the same time(I have observed events, which is registered by jit, were invoked by multithreads), so there might be race condition to update the last hook or longest hook information(there are three hook infors - callsite, starttime and duration, we need to keep three of them consistent).

util/hookable/hookable.cpp
```diff
@@ -213,39 +216,56 @@ J9HookDispatch(struct J9HookInterface **hookInterface, uintptr_t taggedEventNum,
 			/* now read the id again to make sure that nothing has changed */
 			VM_AtomicSupport::readBarrier();
 			if (record->id == id) {
-				uint64_t timeDelta = 0;
+				uint64_t startTime = 0;
+				uint64_t count = 0;
+				if (NULL != eventDump) {
+					count = VM_AtomicSupport::addU64(&eventDump->count,1);
+					sampling = (1 >= samplingInterval) || ((100 >= samplingInterval) && (0 == (count ^ samplingInterval)));
```
amicic: I beileve it should be: count % samplingInterval

util/hookable/hookable.cpp
amicic: It's still strange. Appears that event dump fields are overwritten before winning the atomic op.

I personally would not care for atomically updating 'last' as much as for 'longest'.

util/hookable/hookable.cpp
LinHu2016: sampling = (1 >= samplingInterval) || ((100 >= samplingInterval) && (0 == (count % samplingInterval)));

util/hookable/hookable.cpp
charliegracie: I do not feel like this is the right way to update this structure.  Some fields are updated outside of the atomic and there are two atomic operations. Let me think a bit about this to see a bit more.

LinHu2016:
remove atomic access

PR #1166, opened 2017-05-25
Disable new arraycmp and arraycopy on X86

ncough:
Due to issues discovered in testing, the new implementations
of arraycmp and arraycopy on X86 will be disabled by default.
The prior implementations will be used by default, with options
to enable the new behaviour.


andrewcraik:
LGTM - this is a code restoration and new functionality disable - our testing in a downstream project says that this works well and fixes the issues we were seeing when we used the new evaluators.

andrewcraik:
@0xdaryl can you please review for merging?

PR #1167, opened 2017-05-25
Fold enableZOSTrampolines into enableRMODE64

fjeremic:
Because of the nature of executable memory allocation on z/OS, unless
RMODE64 is enabled we can only allocate executable memory below the 2GB
bar. The relative immediate instructions used for linkage dispatch can
address 32-bits worth of half-words in a signed format, meaning we can
branch -4GB to +4GB from the branch instruction. This region exceeds the
2GB limit and hence when RMODE64 is disabled we never need to generate
trampolines on z/OS.

When RMODE64 is enabled however, the executable memory may be allocated
anywhere within the 64-bit virtual address space, and hence we may
exceed the reachable addressability of the linkage branch instructions.
In such cases we always need to generate trampolines which makes the
enableZOSTrampolines synonymous with enableRMODE64.


fjeremic:
@NigelYiboYu could you please review?

NigelYiboYu:
Look good to me.

PR #1175, opened 2017-05-26
Fix compile err in MSSS::getActiveMemorySize

amicic:
Some compilers enforce return even in the 'unreachable' path.


mgaudet:
If that `return 0` is actually unreachable, shouldn't it be the `Assert_MM__unreachable()` that rolls down to become the default exit?

amicic:
Things are more clear if you look at the whole method which is not very large. return 0 is reachable in what used to be top level else branch. Only one specific branch is not reachable. But that branch did not have return. Now it will have. Since, it's irrelevant what value is returned on unreachable branch, I just shared the existing return 0 from the top level else branch.

mgaudet:
I mismatched my braces when reading :D I see it now.

PR #1176, opened 2017-05-26
Fold memory reference to SS format for MVC instr

r30shah:
Force memory reference to use SS format limits right after
they are generated when we get success to reduce store node
to direct memory-memory copy and use MVC instruction to perform
operation. This avoids scenario when index child of source and
target memory reference shares common nodes and we clobber evaluate
that node before it is used.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -6412,8 +6412,15 @@ bool directMemoryStoreHelper(TR::CodeGenerator* cg, TR::Node* storeNode)
                TR::DebugCounter::incStaticDebugCounter(cg->comp(), "z/optimization/directMemoryStore/indirect");
 
                // Force the memory references to not use an index register because MVC is an SS instruction
+               // After generating a memory reference, Enforce it to generate LA instruction.
+               // This will avoid scenarios when we have common base/index between destination and source
+               // And when generating the source memory reference, it clobber evaluates one of the node shared between 
+               // target memory reference as well.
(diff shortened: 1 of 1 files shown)


r30shah:
FYI @joransiu

joransiu:
LGTM

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6412,8 +6412,16 @@ bool directMemoryStoreHelper(TR::CodeGenerator* cg, TR::Node* storeNode)
                TR::DebugCounter::incStaticDebugCounter(cg->comp(), "z/optimization/directMemoryStore/indirect");
 
                // Force the memory references to not use an index register because MVC is an SS instruction
+               // After generating a memory reference, Enforce it to generate LA instruction.
+               // This will avoid scenarios when we have common base/index between destination and source
+               // And when generating the source memory reference, it clobber evaluates one of the node shared between 
+               // target memory reference as well.
+               // TODO : Fix this. It should not clobber evaluate the node in first place. Workaround just suppresses the bug.
```
0xdaryl: In general we shouldn't be adding TODO's to the code because they will simply be forgotten.  Is there an issue open to track the problem this fix is working around?  If not, please create one and reference this PR.

Does this PR have any value once that overarching issue is fixed?

r30shah: I have opened up issue against this and linked this PR with that with rmoving TODO from the comment.

r30shah:
@0xdaryl I have opened up issue in OMR and also removed TODO from code along with shortening the title. Let me know if you have any other suggestion.

PR #1178, opened 2017-05-26
Add queries to answer a node's collectedness

liqunl:
We used to mark aladd/aiadd nodes as internal pointers, which is wrong
when the address child of the aladd/aiadd node is not collected. This
pull request contains the following change to fix this problem:

1. Add two queries to answer node collectedness
2. Stop converting references to integers

issue: #563
Signed-off-by: liqunl liqunl@ca.ibm.com


liqunl:
Tested with daytrader, spark and specjvm2008, no measurable regression.
Tested with personal builds, one continuous level, one acceptance level. No failure is caused by the change set.

@andrewcraik @jdmpapin Can you review my change set?

Noticed that I didn't remove the switch cases in OMR::ILOpCode::compareOpCode for TR::Address when UnsignedCompare is false, because jitbuilder assumes address comparisons are signed, removing them will break jitbuilder. compareOpCode is only used by IVA and jitbuilder, so it should be safe to leave it like this and we can fix jitbuilder after this PR.

andrewcraik:
I have not reviewed the change yet, but the commit message needs to be shortened to be inline with the contribution guidelines.

liqunl:
@andrewcraik Which part do you mean? The commit messages' headlines are no more than 50 letters and the commit messages are no more than 72 letters a line.

mstoodle:
@liqunl can you provide a bit more detail for your statement "jitbuilder assumes address comparisons are signed" ?

liqunl:
@mstoodle It looks like OMR::ILOpCode::compareOpCode is created for jitbuilder because only jitbuilder uses this function before I use it. Before my change, this function returns valid opcode for address iff UnsignedCompare=false, and it returns BadILOp if the caller tries to ask for an unsigned compare opcode for address type.

andrewcraik:
@liqunl the first line of the commit and the title of the pull request is "Fix address to integer conversions and add canBeInternalPointer query" which is 72 characters long - please shorten.

compiler/il/OMRNode.cpp
```diff
@@ -2357,6 +2357,83 @@ OMR::Node::isNotCollected()
...
+         TR::Symbol *symbol = curNode->getSymbolReference()->getSymbol();
+         if (op.isNew() ||
+             (op.getDataType() == TR::Address && op.isCall()) ||
+             symbol->isInternalPointer() ||
+             symbol->isCollectedReference())
+             return true;
+         }
+
+      visited.add(curNode);
+      if (curNode->getNumChildren() >= 1)
```
andrewcraik: There are plenty more opcodes with more than 1 child - this seems quite fragile... aren't we really talking about nodes with symrefs for the most part or something else? I am worried valid opcodes are going to needlessly trip your assert.

compiler/optimizer/GeneralLoopUnroller.cpp
```diff
@@ -537,6 +533,8 @@ void TR_LoopUnroller::modifyBranchTree(TR_RegionStructure *loop,
 
    if (originalLimitNode->getOpCode().isLoadConst() && !const_overflow)
       {
+      // Address compare generated by loop strider does not have constant node,
+      // thus does not fall into this category
```
andrewcraik: I'm not sure this comment and the one above add much value. If there is an assumption about the nodes an assert is better or at the very least what case doesn't fall in here should be spelled out and not defined by the implementation of another opt which could change independently.

compiler/optimizer/GeneralLoopUnroller.cpp
```diff
@@ -937,6 +924,8 @@ void TR_LoopUnroller::modifyOriginalLoop(TR_RegionStructure *loop, TR_StructureS
          modifyBranchTree(loop, loopNode, branchNode);
          }
 
+      // IVA does not create PIV for ifacmpne and ifacmpeq, thus loops with ifacmpne are not counted loop
+      // and they won't enter this if so there is no need to check for ifacmpne
```
andrewcraik: This needs an assert.

compiler/optimizer/Inliner.cpp
```diff
@@ -3265,13 +3265,13 @@ TR::TreeTop * OMR_InlinerUtil::storeValueInATemp(
...
-      if (
-         (value->getOpCode().isArrayRef() ||
+      if (value->canBeInternalPtrOfObject() ||
             (value->getOpCode().isLoadVarDirect() &&
              value->getSymbolReference()->getSymbol()->isAuto() &&
-             value->getSymbolReference()->getSymbol()->castToAutoSymbol()->isInternalPointer())))
+             value->getSymbolReference()->getSymbol()->castToAutoSymbol()->isInternalPointer()))
          isInternalPointer = true;
 
+      // value->isNotCollected() == false does not mean value is collected
```
andrewcraik: Not sure what you are trying to tell us here... perhaps elaborate about what we are guarding and why the check is the way it is?

compiler/optimizer/ValuePropagationTable.hpp
```diff
@@ -403,10 +403,10 @@ const ValuePropagationPtr constraintHandlers[] =
...
    constrainCmpeq,           // TR::acmpeq
    constrainCmpne,           // TR::acmpne
-   constrainCmplt,           // TR::acmplt
-   constrainCmpge,           // TR::acmpge
-   constrainCmpgt,           // TR::acmpgt
-   constrainCmple,           // TR::acmple
+   constrainCmp,             // TR::acmplt
+   constrainCmp,             // TR::acmpge
+   constrainCmp,             // TR::acmpgt
+   constrainCmp,             // TR::acmple
```
andrewcraik: Could you elaborate on why this seems to have changed from something more specific to something more general?

liqunl: Previous handlers call to constrainCmplessthan which doesn't handle unsigned compares.

compiler/il/OMRNode.cpp
liqunl: According to previous discussion, there are opcodes this query cannot handle properly, such as aternary. The assertion is to make sure only the cases we expect to handle will be handled. The first if in this function ensures we only deal with aiadd/aladd. This if contains a whitelist of opcodes this query can handle properly.

compiler/il/OMRNode.cpp
```diff
@@ -2357,6 +2357,83 @@ OMR::Node::isNotCollected()
...
+
+      if (curNode->getOpCode().hasSymbolReference())
+         {
+         TR::ILOpCode op = curNode->getOpCode();
+         TR::Symbol *symbol = curNode->getSymbolReference()->getSymbol();
+         if (op.isNew() ||
+             (op.getDataType() == TR::Address && op.isCall()) ||
+             symbol->isInternalPointer() ||
+             symbol->isCollectedReference())
+             return true;
```
liqunl: @andrewcraik Maybe we can return false when above if is false, but I'm not sure if we can make such decision based on the symRef.

compiler/optimizer/Inliner.cpp
liqunl: Sorry, this is a note I forgot to remove.

compiler/optimizer/GeneralLoopUnroller.cpp
liqunl: Added an assert here and after above comment.

liqunl:
@andrewcraik I modified the commits, can you review it again?

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+               bool isFirstChildCollected = curNode->getFirstChild()->computeIsCollectedReference();
+               bool isSecondChildCollected = curNode->getSecondChild()->computeIsCollectedReference();
+               TR_ASSERT(isFirstChildCollected != isSecondChildCollected, "Computation does not support address node from aternary with one child collectable the other uncollectable, node is " POINTER_PRINTF_FORMAT, self());
+               return isFirstChildCollected && isSecondChildCollected;
+               }
+            default:
+               return false;
+            }
+         }
+      else
```
andrewcraik: Probably don't need this else - you can let it fall through to the return at the bottom of the method I think.

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+               curNode = curNode->getFirstChild();
+               break;
+            case TR::aternary:
+               {
+               bool isFirstChildCollected = curNode->getFirstChild()->computeIsCollectedReference();
+               bool isSecondChildCollected = curNode->getSecondChild()->computeIsCollectedReference();
+               TR_ASSERT(isFirstChildCollected != isSecondChildCollected, "Computation does not support address node from aternary with one child collectable the other uncollectable, node is " POINTER_PRINTF_FORMAT, self());
+               return isFirstChildCollected && isSecondChildCollected;
+               }
+            default:
```
andrewcraik: So I think the only opcodes not explicitly handled in this switch are i2a and l2a since they are type 'address' with 1 one or more children. Since we explicitly know that we treat them as false for now I think we should have a pair of cases with a return false and a comment saying if a language can handle a collected reference going via a long/int then the logic would need to be augmented to handle that. The default case can then assert for an unhandled opcode so if someone adds a new opcode of type 'address' with 1 or more children they are prompted to add it to this method - makes sure someone doesn't get a false by accident.

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+            case TR::multianewarray:
+            case TR::aiadd:
+            case TR::aladd:
+               curNode = curNode->getFirstChild();
+               break;
+            case TR::aternary:
+               {
+               bool isFirstChildCollected = curNode->getFirstChild()->computeIsCollectedReference();
+               bool isSecondChildCollected = curNode->getSecondChild()->computeIsCollectedReference();
+               TR_ASSERT(isFirstChildCollected != isSecondChildCollected, "Computation does not support address node from aternary with one child collectable the other uncollectable, node is " POINTER_PRINTF_FORMAT, self());
```
andrewcraik: I think this assert should say that 'aternary' does not support having one child collected and the other not collected. It won't work in the codegen and is just not supported so I would state that rather than the computation doesn't support it - the compiler in general doesn't support it.

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+ *    from a collectable reference, i.e. anything that is an address to a collected object.
+ */
+bool
+OMR::Node::computeIsCollectedReference()
+   {
+   TR::Node *curNode = self();
+   if (!curNode->getType().isAddress())
+      return false;
+
+   while (curNode)
```
andrewcraik: I worry about an exponential traversal here because of the aternary... Should we be keeping a checklist of visited nodes?

compiler/optimizer/GeneralLoopUnroller.cpp
```diff
@@ -657,9 +654,7 @@ void TR_LoopUnroller::modifyBranchTree(TR_RegionStructure *loop,
       TR::ILOpCodes opCode;
       if (getIndexType().isAddress())
          {
-         opCode = getTestChildType().isInt32() ?
-            (isIncreasingLoop() ? TR::ifiucmplt : TR::ifiucmpgt) :
-            (isIncreasingLoop() ? TR::iflucmplt : TR::iflucmpgt) ;
+         opCode =isIncreasingLoop() ? TR::ifacmplt : TR::ifacmpgt;
```
andrewcraik: whitespace around =

compiler/il/OMRNode.cpp
andrewcraik: Oh probably also an aconst as well right?

andrewcraik:
Please create an issue for the JITBuilder use of signed addresses - that isn't correct and should be fixed.

compiler/il/OMRNode.cpp
liqunl: Yes, spelling the false cases out sounds like the right thing to do.

compiler/il/OMRNode.cpp
liqunl: The reason why I removed the checklist is that I think we probably don't allow cycles in the tree, because if we do allow, there would be problems evaluating the nodes.

andrewcraik: Ah, but with an aternary you could still get an exponential walk because of the two children - I agree with the single child walk it wouldn't be a problem

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+ *    Return true if the node is an internal pointer, false otherwise.
+ *
+ * \note
+ *    This query currently only works on opcodes that can have a pinning array pointer,
+ *    i.e. aiadd, aiuadd, aladd, aluadd. Suppport for other opcodes can be added if necessary.
+ */
+bool
+OMR::Node::computeIsInternalPointer()
+   {
+   if (self()->getOpCode().hasPinningArrayPointer())
```
jdmpapin: I think the intended condition is `if (!hasPinningArrayPointer())`

compiler/il/OMRNode.cpp
jdmpapin: We need to know not only that a node has been visited, but also the correct result for that node. So to avoid an exponential walk, we would need to memoize with maybe a map or a pair of checklists

compiler/il/OMRNode.cpp
jdmpapin: Looks like this should be `TR_ASSERT(isFirstChildCollected == isSecondChildCollected, ...)`, rather than `!=`. And there's no need to `&&` the results afterward. Either one can be returned because they're required to be interchangeable

compiler/il/OMRNode.cpp
jdmpapin: Without this `else` the loop would restart with the same value of `curNode`. In this case there is no child that `curNode` could have been set to. Maybe we should invert the condition to make the control flow clearer?

    if (curNode->getNumChildren() == 0)
        return false;

And then go on assuming there are children without all of the rightward drift

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,101 @@ OMR::Node::isNotCollected()
...
+            case TR::New:
+            case TR::newarray:
+            case TR::anewarray:
+            case TR::variableNew:
+            case TR::variableNewArray:
+            case TR::multianewarray:
+            case TR::aiadd:
+            case TR::aladd:
+               curNode = curNode->getFirstChild();
+               break;
```
jdmpapin: A lot of these opcodes seem like they don't really belong here:
* `aload` and `aRegLoad` don't have children
* `aloadi` can be collected or not independent of the child, with e.g. VFT pointer loads
* `areturn` and `compressedRefs` don't produce a value
* Calls and allocations would have been handled above, again independent of their children

compiler/il/OMRNode.cpp
jdmpapin: I think the answer should be in terms of the symref alone, when there is one. We can't assume in general that a node is collected iff its first child is collected

compiler/il/OMRNode.cpp
jdmpapin: Constant null is allowed both as collected and as non-collected. If we tell ourselves that we know one way or the other, there will be reasonable cases that trip the ternary assertion

compiler/il/OMRNode.cpp
jdmpapin: On second thought, given the restriction in the doc comment, maybe this should be an assertion forcing callers to check. Currently callers assume that `computeIsInternalPointer()` implies `hasPinningArrayPointer()`, which gets in the way of later returning true for other opcodes, e.g. for `ternary` of two `aladd`s, or for `aload` of a pinning array temp

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -3474,11 +3474,11 @@
    /* .opcode               = */ TR::acmpeq,
    /* .name                 = */ "acmpeq",
    /* .properties1          = */ ILProp1::BooleanCompare,
-   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::SupportedForPRE,
+   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::UnsignedCompare | ILProp2::SupportedForPRE,
    /* .properties3          = */ ILProp3::CompareTrueIfEqual,
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::Int32,
-   /* .typeProperties       = */ ILTypeProp::Size_4 | ILTypeProp::Integer,
+   /* .typeProperties       = */ ILTypeProp::Size_4 | ILTypeProp::Integer | ILTypeProp::Unsigned,
```
jdmpapin: I think `typeProperties` here describes the type of the result rather than the type of the operands. The way you have it is consistent with the `[il]ucmpxx` opcodes, but I don't know why those include `Unsigned`

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -3474,11 +3474,11 @@
    /* .opcode               = */ TR::acmpeq,
    /* .name                 = */ "acmpeq",
    /* .properties1          = */ ILProp1::BooleanCompare,
-   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::SupportedForPRE,
+   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::UnsignedCompare | ILProp2::SupportedForPRE,
```
jdmpapin: I'm not sure whether we want to say that this is an unsigned comparison, when (in)equality is insensitive to signedness

mstoodle:
@liqunl sorry I didn't get back to you on the JitBuilder connection. So, what that means is that anyone who uses something like LessThan on values that have Address type will get a comparison, but anyone using LessThanUnsigned on such values would assert in the compiler.

The right answer here is the reverse: anyone comparing things that are addresses in JitBuilder should be using Unsigned comparisons, so I think you should just remove the address compare opcodes from compareOpCode() on the "!needsUnsigned" branch. That's a change in behaviour, but I would also call it fixing a bug. If many people trip on it, I'll add an assert to JitBuilder to provide a more helpful message (actually, maybe I'll just add another "beginner" issue to add that assert once your PR is merged).

compiler/il/OMRNode.cpp
liqunl: Code has been updated to account those cases.

compiler/il/OMRILOps.cpp
```diff
@@ -291,3 +297,30 @@ OMR::ILOpCode::compareOpCode(TR::DataType dt,
...
+      return TR_cmpGT;
+   else if (isLessCmp(op))
+      return TR_cmpLE;
+   else if (isGreaterCmp(op))
+      return TR_cmpGE;
+   else if (isEqualCmp(op))
+      return TR_cmpEQ;
+   else if (isNotEqualCmp(op))
+      return TR_cmpNE;
+   TR_ASSERT(true, "Expect comparison opcode");
```
jdmpapin: You probably wanted `TR_ASSERT(false)`

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,115 @@ OMR::Node::isNotCollected()
...
+      // The following are all opcodes that are address type, non-TreeTop and non-conversion
+      switch (opValue)
+         {
+         case TR::aloadi:
+         case TR::aiadd:
+         case TR::aiuadd:
+         case TR::aladd:
+         case TR::aluadd:
+            curNode = curNode->getFirstChild();
+            break;
```
jdmpapin: `aloadi` still doesn't fit here. The VFT load was just one example to illustrate that garbage collected memory and native memory may have pointers into each other both ways. Another example would be `<classFromJavaLangClass>` and `<javaLangClassFromClass>`. Rather than trying to special-case these things, I think the collectedness of the result of a load should be determined by its symref

jdmpapin: Hopefully it's not actually valid for an address-type load to have a symbol with `!isInternalPointer() && !isCollectedReference() && !isNotCollected()`. If so, then `aloadi` (and `aload` and `aRegLoad`) nodes won't even reach here

compiler/il/ILOpCodeProperties.hpp
liqunl: Right, I did not think about that. Removing that should not affect the functionality, I'll do a build and remove the typeProperty if it passes.

compiler/il/OMRNode.cpp
liqunl: > Hopefully it's not actually valid for an address-type load to have a symbol with !isInternalPointer() && !isCollectedReference() && !isNotCollected(). If so, then aloadi (and aload and aRegLoad) nodes won't even reach here

That is what I'm worrying about. I can do a personal build with asserts on those cases. If we do not see such cases now, we should not expect them in the future.

compiler/il/OMRILOps.cpp
liqunl: My bad, fixed.

compiler/il/OMRNode.cpp
liqunl: @andrewcraik @jdmpapin Sorry I didn't notice your comments until just now. I still couldn't come up with a case where we can have cycles in the tree, which is the reason causing an exponential walk. And it makes no sense for a tree to have cycles whichever path is taken. The only problem I see here is that the two children of aternary can be the same node. I can check that to avoid examining the same node twice.

andrewcraik: The checklist isn't for cycles it is to avoid rewalking nodes in non-cyclic trees in such a way that causes exponential time walks.

mstoodle: Remember the "A" in "DAG" is Acyclic :) .

But @andrewcraik and @jdmpapin are right: you can still get an exponential visit if you're descending past commoned nodes (imagine how you would traverse an aternary where both its operands are themselves aternary nodes, etc.). This particular scenario is *probably* pretty rare, but I've hit examples of exponential walks in the past (just not with aternaries). Maybe you only have to memoize the results for aternary nodes (really, any node with more than one address child) to prevent the exponential walk?

liqunl: Thanks, now I understand. Will add a checklist to the code.

liqunl:
issue #1225 has been created for the signedness problem in jitBuilder.

liqunl:
@andrewcraik @jdmpapin I removed aternary from the support list, will create a git issue for removing aternary opcode. Could you review my change again?

compiler/il/OMRILOps.cpp
```diff
@@ -291,3 +297,30 @@ OMR::ILOpCode::compareOpCode(TR::DataType dt,
...
+      return TR_cmpGT;
+   else if (isLessCmp(op))
+      return TR_cmpLE;
+   else if (isGreaterCmp(op))
+      return TR_cmpGE;
+   else if (isEqualCmp(op))
+      return TR_cmpEQ;
+   else if (isNotEqualCmp(op))
+      return TR_cmpNE;
+   TR_ASSERT(false, "Expect comparison opcode");
```
jdmpapin: This fallthrough path should either return a value, or fail to return entirely. Otherwise, when assertions are disabled, it will return garbage that's unlikely to be a valid `TR_ComparisonTypes`.

One way to avoid returning would be to use `TR_ASSERT_FATAL` instead of `TR_ASSERT`. Although I'm not sure whether that would be clear to compilers. It would be good to avoid compiler warnings of "not all paths return a value."

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,103 @@ OMR::Node::isNotCollected()
...
+               return symbol->isCollectedReference();
+            }
+         case TR::aconst:
+         case TR::getstack:
+            return false;
+         default:
+            TR_ASSERT(false, "Unsupported opcode %s on node " POINTER_PRINTF_FORMAT, op.getName(), curNode);
+            return false;
+         }
+      }
```
jdmpapin: As with `getCompareType`, the lack of return here after falling off of the `while` loop would produce garbage if it were ever to occur

compiler/il/OMRNode.cpp
```diff
@@ -2357,7 +2357,103 @@ OMR::Node::isNotCollected()
...
+         case TR::aload:
+         case TR::loadaddr:
+            {
+            TR::Symbol *symbol = curNode->getSymbolReference()->getSymbol();
+            // isCollectedReference() responds false to generic int shadows because their type
+            // is int. However, address type generic int shadows refer to collected slots.
+            if (opValue == TR::aloadi && symbol == TR::comp()->getSymRefTab()->findGenericIntShadowSymbol())
+               return true;
+            else
+               return symbol->isCollectedReference();
```
jdmpapin: What's the result of this when `symbol` is an internal pointer temp?

compiler/optimizer/ValuePropagationTable.hpp
```diff
@@ -488,10 +488,10 @@ const ValuePropagationPtr constraintHandlers[] =
...
    constrainIfcmpeq,         // TR::ifacmpeq
    constrainIfcmpne,         // TR::ifacmpne
-   constrainCondBranch,      // TR::ifacmplt
-   constrainCondBranch,      // TR::ifacmpge
-   constrainCondBranch,      // TR::ifacmpgt
-   constrainCondBranch,      // TR::ifacmple
+   constrainIfcmplt,         // TR::ifacmplt
+   constrainIfcmpge,         // TR::ifacmpge
+   constrainIfcmpgt,         // TR::ifacmpgt
+   constrainIfcmple,         // TR::ifacmple
```
jdmpapin: These handlers look safe for address inequalities, but only because they don't do much with unsigned comparisons

compiler/il/OMRNode.cpp
liqunl: It should be ture. Internal pointer temps are address type, and that query is equivalent to `return (self()->getDataType()==TR::Address || self()->isLocalObject()) && !self()->isNotCollected();`

compiler/il/OMRILOps.cpp
liqunl: I think I can make the last `else if` an `else`. There is compare type other than `TR_cmpNE` that would reach there.

compiler/il/OMRNode.cpp
liqunl: Added a return at the end of the function.

andrewcraik:
@jdmpapin have your review comments been addressed?

jdmpapin:
Yes, LGTM

liqunl:
I'm investigating a failure on z in my personal build where we seem to create a store from aladd of aconst null and a long to a temp without setting the temp's uncollected flag. This results in the temp slot being marking as collected and then a crash in gc.

mstoodle:
Hi @liqunl any updates here, please?

I hope you're intending to squash the commits (updates, Update II, Update III) once the final review is complete :) .

liqunl:
Yes. I'm waiting for a memory change to merge this one. Will resume this work when the memory change is merged.

0xdaryl:
@liqunl, is there a PR already created for the feature you need before merging this?  Which of the two original commits is this fix needed for?  I ask because it seems like this PR has two independent features its trying to contribute (am I wrong?).  Is it possible to create a separate PR for just adding the internal pointer query and getting that committed first?  Because this current PR needs some work (rebasing and squashing and potentially other fixes) prior to merging I suspect you might have an easier path forward with that approach.

If there are still issues to resolve and you're not close to remedying them then I suggest closing this PR and working on them in your own personal fork.

andrewcraik:
The PR for the required feature is marked as Mem1 and is ready for merging. Unfortunately, separating the two issues in this PR is non-trivial due to the complex interplay between the different queries and getting the notion of internal pointer correct. Once the Mem1 change is merged a rebase of this should resolve the issue and we can get it merged. I am against this being broken apart of closed at this time - it is a vital bit of work that has exposed nasty bugs which is why it is waiting on the other change.

liqunl:
@andrewcraik @jdmpapin This is the up-to-date change. The Toronto personal build has passed, I'm struggling with some infra issues in personal builds in Ottawa, will update once have a build up running and finished.

liqunl:
The Ottawa pbuild has also passed.

liqunl:
@andrewcraik @jdmpapin All the tests have passed, could you review my change so we can get this merged? Will rebase the commits once review is done.

andrewcraik:
The only controversial part of this commit is the walk computing if an address is derived from a collected reference. The original review was pointing out that you could get an exponential walk with an aternary opcode. Discussions have said that we don't think aternary is a good idea due to the possibility of its children being of different 'collectedness'. The opcode is not currently used and we should be able to get rid of it. As a result I'm happy to not support it in that walk, but there needs to be an assert so that we don't do the 'wrong' thing with it.

Otherwise I think I am happy with the change. @jdmpapin any further comments? Otherwise I think this should be rebased and un-WIPed so it can be merged.

jdmpapin:
It's hard to see what the differences are from last time. Primarily I see that while `aconst NULL` is considered collected, the following is not, regardless of `C`, and in particular even when `C` is `NULL`.

    a[il]add
      aconst C
      ...

I guess this is to address the issue in your comment above,

> where we seem to create a store from aladd of aconst null and a long to a temp without setting the temp's uncollected flag. This results in the temp slot being marking as collected and then a crash in gc.

The solution looks good. I'm perfectly happy with forbidding fabrication of pointers into GCed memory via offsets from a constant address. That would just be `l2a` in disguise.

@ajcraik An `aternary` node would already flow into the default case and hit the assertion there on unsupported opcodes

liqunl:
The changes in this commit compared to last reviewed one are:
1, Treat aconst null as collected
2, Change isNotCollected query on node to utilize the more accurate computeIsCollectedReference query

0xdaryl:
My only (non-technical) comment at this point would be to drastically simplify your PR message.  You don't need to cut-and-paste the commit messages of each of its constituent commits because that's just overly redundant.  A PR message with just the title of each constituent commit should be sufficient, perhaps with a bit of surrounding context.

jdmpapin:
`isNotCollected` looks good as well :+1:

liqunl:
Commits have been rebased and retested with Toronto personal build. It is safe to merge once all the checks have completed.

andrewcraik:
@vijaysun-omr over to you - finally ready to get this one merged

liqunl:
@vijaysun-omr Can you merge this MR when you have time? It has been reviewed by Andrew and Devin and has passed all the test. I just redid a personal build today and every job has passed.

compiler/il/OMRNode.cpp
```diff
@@ -2316,17 +2316,115 @@ OMR::Node::dontEliminateStores(bool isForLocalDeadStore)
...
+/**
+ * \brief
+ *    Check if the node is an internal pointer.
+ *
+ * \return
+ *    Return true if the node is an internal pointer, false otherwise.
+ *
+ * \note
+ *    This query currently only works on opcodes that can have a pinning array pointer,
+ *    i.e. aiadd, aiuadd, aladd, aluadd. Suppport for other opcodes can be added if necessary.
```
vijaysun-omr: I think aload and aRegload opcodes can also have a pinning array pointer. It seems to me the comment (and possibly the code) needs to be changed to ensure that those opcodes are included.

jdmpapin: I believe that `aiadd`, `aiuadd`, `aladd`, and `aluadd` are the only opcodes this function is currently meant to handle. The comment likely means to say that it only works for nodes whose opcodes satisfy `hasPinningArrayPointer()`, which is perhaps poorly named. Callers check `hasPinningArrayPointer()` before calling this function.

Originally the result for any other opcode would have been false. Precisely because we might later want to return true for loads of internal pointer temps, or possibly in other reasonable cases, Liqun documented this opcode restriction and added the assertion below to check that it holds. This way, callers can't assume that a true result informs them about the opcode of the node, and we're free to relax the opcode restriction in the future.

Since then, the implementing `computeIsCollectedReference()` has changed so that it actually will return true for loads of internal pointer temps ([see here](https://github.com/eclipse/omr/pull/1178/files/ade5a1ccac7694304d9d7ab63e2a6b3dacb03cb8#r120992574)). However, I think the same logic will also produce true for typical loads producing (non-internal) references into the GC heap. But because of the restriction on the opcodes of the input, `computeIsInternalPointer()` makes no claim about the result in such cases.

compiler/il/OMRNode.cpp
```diff
@@ -2316,17 +2316,115 @@ OMR::Node::dontEliminateStores(bool isForLocalDeadStore)
...
+         case TR::New:
+         case TR::newarray:
+         case TR::anewarray:
+         case TR::variableNew:
+         case TR::variableNewArray:
+         case TR::multianewarray:
+              return true;
+         // aload, aloadi, aRegLoad and loadaddr are associated with a symref, we should
+         // be able to tell its collectedness from its symref.
+         case TR::aRegLoad:
```
vijaysun-omr: Calling getSymbolReference on a register load (e.g. aregload) is a bit risky if the opcode does not return true for hasSymbolReference. That being said, I think there is at least one other spot in the compiler that accesses the symbol reference on a register load and I don't know of a specific reason why this would cause a functional problem as things stand.

compiler/optimizer/OMRLocalCSE.cpp
```diff
@@ -1355,7 +1355,8 @@ TR::Node* OMR::LocalCSE::getAvailableExpression(TR::Node *parent, TR::Node *node
          }
       }
 
-   if ((node->getOpCode().isArrayRef()) &&
+   if (node->hasPinningArrayPointer() &&
```
vijaysun-omr: Maybe you considered wrapping the hasPinningArrayPointer && computeIsInternalPointer tests inside a query routine so that you do not have the 2 tests in several places. It's not a big deal though.

compiler/il/OMRNode.cpp
liqunl: The original query only supported those opcodes and it assumed whoever uses it knows that. To make the new query less risky, this query currently keeps that assumption.  

We have two `hasPinningArrayPointer`, one is on `OMR::Node`, the other is on `OMR::ILOpCode`. It should be better to use the `OMR::Node` here since it checks the property of a node. It currently uses the `OMR::ILOpCode::hasPinningArrayPointer` internally. We can change it to have its own implementation if the two queries were to diverge.

PR #1179, opened 2017-05-26
Fix properties of VLGV instruction on Z architecture

hjooybar:
“VECTOR LOAD GR FROM VR ELEMENT” (VLGV instruction) loads an element of a vector register in a scalar register.
However, regardless of the element size, the remaining of 64-bit output GR fills with zeros.
This indicates that the instruction clobbers all 64 bits of its output register and has to be marked as a 64bit instruction.


vijaysun-omr:
Looks safe to me. I would like an okay from @joransiu too.

joransiu:
LGTM

PR #1180, opened 2017-05-26
Recognize and accelerate ECC methods on x86

hjooybar:
256-bit multiply and mod operations are vastly used in Elliptic curve cryptography.
These operations can be accelerated using handcrafted assembly instruction sequence.
This changeset recognizes method signature of these operations and replaces them with
an optimized instruction sequence.


hjooybar:
@andrewcraik
 This is a dependent changeset and has to be committed after its Java side counterpart changeset

mstoodle:
@hjooybar @andrewcraik Comments like the above don't really belong in this project, please. Change coordination requirements for consuming projects don't make sense to be documented in this project.

Imagine if we had 20 projects doing the same thing in some fraction of our PRs...

andrewcraik:
@mstoodle agreed.

andrewcraik:
If the evaluator is not in OMR the hookup doesn't belong in OMR. If the hookup is generally useful and contribution of the evaluator is possible then that is an alternative option.

vijaysun-omr:
@0xdaryl please review as well.

hjooybar:
As requested by @andrewcraik  the hook will not be added to OMR project

PR #1182, opened 2017-05-29
Try removing if node in all compare simplifiers

liqunl:
Having an if node with its branch target the same as its immediate fall
through block results in an illy formed cfg that will confuse other
optimizations. Removing such node is supported in most comparison
simplifiers but is missing in ifacmpne and ifacmpeq, and in generic
simplifiers that handles if nodes such as dftSimplifier. This commit
adds code to try to remove such if node in the simplifier.

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -1902,22 +1902,6 @@ static TR::Node *intDemoteSimplifier(TR::Node * node, TR::Block * block, TR::Sim
    return node;
    }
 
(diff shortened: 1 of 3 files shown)


liqunl:
@andrewcraik This is the fix. Will update the result of personal build once it finishes.

liqunl:
Move removeIfToFollowingBlock to OMRSimplifierHelpers.cpp so that J9 can also use this function. Personal build with both J9 and OMR change is running, will update the result once it finishes.

andrewcraik:
LGTM - @vijaysun-omr could you have a look please?

liqunl:
Personal build has passed.

vijaysun-omr:
Looks okay to me. Merging.

Issue #1183, opened 2017-05-29
No template instantiation for CodeGenerator::loadAndStoreMayOverlap that leads to "undefined symbol" at runtime


last5bits:
The following template method violates the C++ standard ([StackOverflow](https://stackoverflow.com/questions/26745973/potential-evaluation-of-inline-function-bodies-and-instatiation-of-template-memb)):

`omr/compiler/z/codegen/OMRCodeGenerator.hpp`

`template <class TR_AliasSetInterface> bool loadAndStoreMayOverlap(TR::Node *store, size_t storeSize, TR::Node *load, size_t loadSize, TR_AliasSetInterface &storeAliases);`

since the template method is implemented in the .cpp file rather than the header file.

This leads to an `undefined symbol` error during runtime. The code works on GCC 4.4.7; however, GCC 4.8.5 seems to inline the template straight into the other overload candidate without emitting the template instantiation (hence, `undefined symbol`).

last5bits:
There are two potential solutions:
1) Move the template definition into the header
2) Explicitly instantiate the template

Solution 1) is preferable since it won't require manually instantiating the template function after adding new calls (with different template arguments).

I will push the corresponding fix shortly.

mstoodle:
Hi @last5bits , would you please change the text in this issue to refer to a problem in OMR without directly referencing the project you're working on that consumes OMR?

PR #1184, opened 2017-05-29
 Fix an "undefined symbol" runtime error on zLinux

last5bits:
Fix an "undefined symbol" runtime error on zLinux

The template method `loadAndStoreMayOverlap` in `OMR::Z::CodeGenerator`
was implemented in the implementation file rather than the header file,
and, hence, was violating the C++ standard. GCC 4.8.5 is inlining the
template straight into the overload candidate and not emitting the
template instantiation. The result was an "undefined symbol" error
during runtime on zLinux.
Issue: #1183

Signed-off-by: Alexey Zhikhartsev <alexey.zhikhartsev@ibm.com>

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator_inlines.hpp
+++ b/compiler/z/codegen/OMRCodeGenerator_inlines.hpp
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2000, 2017
+ *
+ *  This program and the accompanying materials are made available
+ *  under the terms of the Eclipse Public License v1.0 and
+ *  Apache License v2.0 which accompanies this distribution.
+ *
+ *      The Eclipse Public License is available at
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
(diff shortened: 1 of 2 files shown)


compiler/z/codegen/OMRCodeGenerator_inlines.hpp
```diff
@@ -0,0 +1,31 @@
...
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
+
+#ifndef OMRCODEGENERATOR_INLINES_INCL
```
0xdaryl: This #ifdef is too similar to it's parent in OMR, and offers no indication that its for the Z codegen.  Please use `OMR_Z_CODEGENERATOR_INLINE_INCL` instead.

last5bits: Fixed.

Issue #1185, opened 2017-05-29
Deprecate isConcurrentScavengeEnabled codegen query


last5bits:
Deprecate `isConcurrentScavengeEnabled` codegen query and instead expose a query on the Object Model class querying whether we need read barriers for field loads. The compiler shouldn't technically be querying the codegen to ask if concurrent scavenge is enabled, that should be something that the Object Model should answer.

PR #1186, opened 2017-05-29
Deprecate isConcurrentScavengeEnabled codegen query

last5bits:
Instead, expose a query on the Object Model class querying whether we
need read barriers for field loads. The compiler shouldn't technically
be querying the codegen to ask if concurrent scavenge is enabled, that
should be something that the Object Model should answer.
Issue: #1185

Signed-off-by: Alexey Zhikhartsev <alexey.zhikhartsev@ibm.com>


joransiu:
LGTM

PR #1197, opened 2017-05-31
Fix x86 encoding of mov[dq] between xmm and memory

jdmpapin:
The encoding table erroneously specified `REX.W` for `movd` instead of `movq`, so the JIT would encode `movd` as `movq` and vice versa. This change removes the `REX.W` prefix from `movd`, and changes `movq` to an alternate encoding that allows it to be used in 32-bit code.


andrewcraik:
LGTM @vijaysun-omr or @0xdaryl can we get this merged - the previous encodings are wrong and can cause some nasty, subtle issues for people running on x86.

PR #1198, opened 2017-06-01
Zero offset OSR induction for sync method monent

ncough:
Under NextGenHCR, the OSR transition occurs after
the yield point. In sync methods, transitions
can occur at the monent in the method entry. However,
as this monent does not exist, transitioning with
an offset will result in the first bytecode being
skipped.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -815,14 +815,22 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    if (osrNode->getOpCode().isCall())
       return 3;
 
-   switch (osrNode->getOpCodeValue())
+   // If the monent has a bytecode index of 0, it must be the
+   // monent for a synchronized method. A transition here
+   // must target the method entry.
+   if (osrNode->getOpCodeValue() == TR::monent)
       {
-      case TR::monent: return 1;
-      case TR::asynccheck: return 0;
-      default:
-         TR_ASSERT(0, "OSR points should only be calls, monents or asyncchecks");
+      if (osrNode->getByteCodeIndex() == 0)
          return 0;
+      else
+         return 1;
       }
+
+   if (osrNode->getOpCodeValue() == TR::asynccheck)
+      return 0;
+
+   TR_ASSERT(0, "OSR points should only be calls, monents or asyncchecks");
+   return 0;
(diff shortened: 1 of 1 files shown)


andrewcraik:
Change looks quite reasonable - please comment on testing once complete and then I think we can look to merge this.

ncough:
This has passed testing in a down stream project and successfully transitions when forced at the method monitor enter.

andrewcraik:
LGTM @vijaysun-omr I think you are the best committer to look at this one.

compiler/compile/OMRCompilation.cpp
```diff
@@ -815,14 +815,22 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
...
+   // If the monent has a bytecode index of 0, it must be the
+   // monent for a synchronized method. A transition here
+   // must target the method entry.
+   if (osrNode->getOpCodeValue() == TR::monent)
       {
-      case TR::monent: return 1;
-      case TR::asynccheck: return 0;
-      default:
-         TR_ASSERT(0, "OSR points should only be calls, monents or asyncchecks");
+      if (osrNode->getByteCodeIndex() == 0)
```
vijaysun-omr: Is this test sufficient to distinguish between 

the "implicit" monent on method entry for synchronized methods 
vs 
an "explicit" monent that may happen to be the first bytecode (and therefore have bytecode index 0) for a method ?

I believe you were thinking of the former when you implemented this change but I am just wondering if anything could go wrong in the latter case.

ncough: From my understanding of how monitor enters are currently implemented, there is no clear way to distinguish between the implicit and the explicit entries. A node flag exists, however, it is not reliable. Bytecode index is checked because an explicit monitor entry requires an argument and therefore couldn't be the first bytecode, only the implicit could be.

andrewcraik:
Looks good to me - the bytecode index 0 is a bit fragile, but I don't think there is a better alternative at present. Open to suggestions, however.

compiler/compile/OMRCompilation.cpp
vijaysun-omr: ..."an explicit monitor entry requires an argument and therefore couldn't be the first bytecode, only the implicit could be"....good point!

PR #1199, opened 2017-06-01
Stash call arguments against BCI for OSR induce

ncough:
Under NextGenHCR, the OSR transition occurs after
the yield point. In sync methods, transitions
can occur at the monent in the method entry. However,
as this monent does not exist, transitioning with
an offset will result in the first bytecode being
skipped.


andrewcraik:
LGTM - we treat the arguments to the transition target differently than other stack slots and this accounts for that in the post transition case so makes sure the stack is setup correctly.

PR #1200, opened 2017-06-01
WIP: Insert a GC Map for Unsafe Shadow Symbol [X86]

0dvictor:
An exception may occur on a node with Unsafe Shadow Symbol,
and hence a GC Map is needed.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -1075,6 +1075,12 @@ TR::Register *OMR::X86::TreeEvaluator::integerStoreEvaluator(TR::Node *node, TR:
    if (instr && node->getOpCode().isIndirect())
       cg->setImplicitExceptionPoint(instr);
 
+   if (node->getSymbol()->isUnsafeShadowSymbol())
+      {
+      instr->setNeedsGCMap(0xFF00FFFF);
+      instr->setNode(node);
+      }
+
    if (usingCompressedPointers)
       cg->decReferenceCount(node->getSecondChild());
 


compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1075,6 +1075,12 @@ TR::Register *OMR::X86::TreeEvaluator::integerStoreEvaluator(TR::Node *node, TR:
    if (instr && node->getOpCode().isIndirect())
       cg->setImplicitExceptionPoint(instr);
 
+   if (node->getSymbol()->isUnsafeShadowSymbol())
```
vijaysun-omr: Do you need to handle loads as well (you seem to have changed a store evaluator) ?

vijaysun-omr: Is there some reason why only "integer" store evaluator was changed ? What about the other datatypes that are possible using unsafe ?

vijaysun-omr: What about the other plattforms (apart from x86) ?

0dvictor: This version only handles store, I'll update the changeset to add support for load.

0dvictor: integerStoreEvaluator handles all kinds of integer data types, including Int8, Int16, Int32, Int64, and Address.
For float and double, call stack shows it eventually calls putInt/putLong, and the exception comes from putInt/putLong.
Exception in thread "main" java.lang.InternalError
        at sun.misc.Unsafe.putLong(Native Method)
        at java.nio.DirectByteBuffer.putDouble(DirectByteBuffer.java:1030)
        at java.nio.DirectByteBuffer.putDouble(DirectByteBuffer.java:1053)
        at TTest.put(TTest.java:36)
        at TTest.main(TTest.java:23)

0dvictor: I guess they may use similar, if not the same, strategy by setting GC or exception information at store/load instructions.  However, this evaluator is platform specific.

vijaysun-omr: Okay FYI to @ymanton @zl-wang @fjeremic and @joransiu and @JBKingdon so that the other platforms can take note of the need to do similar work.

ymanton:
Not sure I follow exactly what's going on here, but I don't think Power needs this.

Hard to discuss this in general terms without Java-isms so I won't bother, but IIRC the x86 code generator foregoes explicitly generating checks for null pointers in favour of piggybacking Java exceptions onto the resulting OS exceptions (e.g. `NullPointerException` onto `SIGSEGV` from a load from 0), which I guess explains why you would want a GC map on loads/stores to/from unsafe symbols, since the base register can be null. Is that what is going on here @0dvictor?

On Power we generate trap instructions to test for null and those should have GC maps already. Probably similar for the other code generators as well.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1075,6 +1075,12 @@ TR::Register *OMR::X86::TreeEvaluator::integerStoreEvaluator(TR::Node *node, TR:
...
+      {
+      instr->setNeedsGCMap(0xFF00FFFF);
```
ymanton: Do you actually need registers in your GC map? Wherever this exception ends up being caught, whether in the same method or elsewhere, I don't think that the currently live registers will be live at the destination so there's no point in having them in the map, unless this is an x codegen specific quirk that I'm not aware of.

0dvictor: On X, GC map is required to find Byte Code Information (BCI) for the instruction. When exception occurs, BCI is required to generate correct back trace.

ymanton: I understand that part, but the `0xFF00FFFF` mask results in the GC map containing registers, which unless I am missing something I suspect you don't really want in this case.

0dvictor:
No, this is to avoid crash when SIG_BUS occurs. It is a X86 Linux signal that may occur when accessing memory mapped file. It may or may not apply to other platforms.

zl-wang:
it is not wrong to put register map in GC map though.  it didn't waste space either, except that line of code, although i agree with you to get rid of it.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Sure, I'll change it to not preserving live registers.

0dvictor:
Sure, I'll change it to not preserving live registers.

ymanton:
OK, I was pointed to the downstream bug that prompted this fix, which explains that load and store nodes are being created as part of an "unsafe" API, and that if they cause OS exceptions the downstream project wants to map those to language exceptions, so it needs bytecode information at the `pc` that caused the exception. It's not even that downstream wants to do this for all loads/stores to unsafe symbols, it just wants to do this for a particular few, and only on particular platforms...

It seems a shame that we have to fix this in the OMR compiler, for all unsafe symbols, and in each code generator no less.

ymanton:
Perhaps what you really want here is the `canGCandExcept` node query, which just looks at the corresponding `canGCandExcept` symref flag. Maybe downstream should set this flag for the particular symrefs that require this behaviour and the evaluator can simply check `canGCandExcept` before deciding whether or not to generate a GC map. That would free us from hardcoding this behaviour for every unsafe load/store for every OMR consumer just to fix a very limited problem for one downstream project.

zl-wang:
Could you (Victor) explain the entitlement BCI should be asked for in this particular crash?  It appeared the exact BCI doesn't help in fixing the crash, since the memory-mapped region went haywire and it was not caused by the Unsafe access itself.  If the crash (SIGBUS) happened in native code (C code for example), you probably only got that crash too.

Issue #1203, opened 2017-06-01
Fix for incorrect register range when creating register dependency on Power


IBMJimmyk:
The register enums in p/codegen/RealRegisterEnum.hpp are listed in this order: GPR, FPR, VRF, CCR

createCondForLiveAndSpilledGPRs in p/codegen/OMRMachine.cpp tries to iterate over those four types of registers by setting a loop range to go from FirstGPR to LastCCR. It does this to generate a register dependency. This logic is correct so there are no problems so far.

createCondForLiveAndSpilledGPRs then checks to see if there are any live VRF registers. If so, it changes the end of the loop range to LastVRF. This is a problem. The range from FirstGPR to LastVRF only covers GPR, FPR and VRF registers.  The existence of a live VRF register causes CCR registers to be excluded from consideration to be added to the register dependency.

The proposed fix is to simply remove the check for live VRF registers. Setting the loop range to go from FirstGPR to LastCCR will cover the four register types already.

PR #1204, opened 2017-06-01
Fix for bad register dependency creation on Power

IBMJimmyk:
The register enums in p/codegen/RealRegisterEnum.hpp are listed in this
order: GPR, FPR, VRF, CCR. createCondForLiveAndSpilledGPRs in
p/codegen/OMRMachine.cpp tries to iterate over those four types of
registers by setting a loop range to go from FirstGPR to LastCCR. It does
this to generate a register dependency. This logic is correct so there are
no problems so far.

createCondForLiveAndSpilledGPRs then checks to see if there are any live
VRF registers. If so, it changes the end of the loop range to LastVRF.
This is a problem. The range from FirstGPR to LastVRF only covers GPR, FPR
and VRF registers. The existence of a live VRF register causes CCR
registers to be excluded from consideration to be added to the register
dependency.

This fix removes the check for live VRF registers. It is unnecessary since
the default range of FirstGPR to LastCCR already includes vector
registers.

Issue: #1203

### DIFF
--- a/compiler/p/codegen/OMRMachine.cpp
+++ b/compiler/p/codegen/OMRMachine.cpp
@@ -1864,12 +1864,7 @@ OMR::Power::Machine::createCondForLiveAndSpilledGPRs(bool cleanRegState, TR::lis
    // it is space conscious.
    //
    int32_t c = 0;
-   int32_t endReg = TR::RealRegister::LastCCR;
-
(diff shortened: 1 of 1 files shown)


ymanton:
Straight forward fix; LGTM, thanks.

IBMJimmyk:
I changed the commit title to say Power instead of just "p".

PR #1205, opened 2017-06-01
Fix iu2l not being evaluated on z systems

NigelYiboYu:
iu2l was not evaluated and resulted in incorrect array indexing in
BigInteger Multiplication operations.

This is fixed by calling generic load helper if the input is unsigned; and
we skip the node evaluations only if the nodes are signed and already
sign exnended.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -5331,8 +5331,23 @@ OMR::Z::TreeEvaluator::extendCastEvaluator(TR::Node * node, TR::CodeGenerator * 
    // reusing the source 64bit GPR, as this would trigger an assume in RA.
    if (cg->supportsHighWordFacility() && !comp->getOption(TR_DisableHighWordRA) && (numberOfExtendBits==64))
       canClobberSrc = false;
-
-   if (!node->isUnneededConversion() && !childRegister->alreadySignExtended())
+    
+   /**
+   * A load (for proper type conversion) is required if the node is unsigned, not sign extended and
+   * not marked as unneeded conversion.
+   *
+   * For example, codegen can encounter the following tree:
+   *
+   * n4188n   (  0)  iRegStore GPR6  (Unsigned NeedsSignExt privatizedInlinerArg )
+   * n4109n   (  3)    l2i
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
Could you please elaborate on "why" the iu2l was not being evaluated ? This would allow me to review the fix with proper context.

vijaysun-omr:
I would like @zlwang and @joransiu to review.

joransiu:
The proposed change LGTM.   The register flag "childRegister->alreadySignExtended()" is set in iRegStoreEvaluator to signify that the value was **sign** extended.   The existing code uses that flag to determine whether to skip the genericLoadHelper which would generate the sign extension instruction.  In this case, we incorrectly skipped generation, as iu2l would require a **zero** extension instead.

NigelYiboYu:
Added the following comment to this commit:
```
 A load (for proper type conversion) is required if the node is unsigned, not sign extended and
 not marked as unneeded conversion.

 For example, codegen can encounter the following tree:

 n4188n   (  0)  iRegStore GPR6  (Unsigned NeedsSignExt privatizedInlinerArg )
 n4109n   (  3)    l2i
 n19339n  (  1)      ==>lRegLoad (in GPR_3616) (highWordZero X>=0 cannotOverflow SeenRealReference )
  .....
 n4182n   (  0)      iu2l (in GPR_3616) (highWordZero X>=0 )
 n4109n   (  0)        ==>l2i (in GPR_3616)

 In this case, a load is needed to zero the high half of the long to corretly convert iu to l.
```

vijaysun-omr:
Thanks for the explanation @joransiu and also for your "negatives packed" explanation @NigelYiboYu. I have never seen so many "un" and "not" used in such a small amount of text (just kidding) :-) I will merge now.

zl-wang:
isSourceTypeSigned  is really confusing.  It is more appropriate to call it isTargetTypeSigned (i.e.  the result type is signed or unsigned).   It is a correct fix though, when you are looking for a signed result and the child register is sign-extended already.

A potential more generic fix:  TargetTypeSigned-ness == childTypeSigned-ness, and child register is extended already (be it zero or sign extended).  Leave it for the future (e.g.  something like ui2ul).

zl-wang:
This bug happened in the situation:  you are looking for an unsigned result but the childRegister is sign-extended, and skipping the necessary action.

PR #1206, opened 2017-06-01
Ensure JitBuilder creates calls under treetops

mstoodle:
IlBuilder::genCall() was only generating call nodes under their
own treetop if the call node did not return a value. Unfortunately,
the way IlValue works, it could cause the call node to swing down
to places it should never go. Resolve that by anchoring the call
node under its own treetop explicitly in genCall().

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1758,14 +1758,14 @@ IlBuilder::genCall(TR::SymbolReference *methodSymRef, int32_t numArgs, TR::IlVal
       callNode->setAndIncChild(childIndex++, loadValue(arg));
       }
 
+   // callNode must be anchored by itself
+   genTreeTop(callNode);
+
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
This change looks right to me since needing to anchor call nodes under their own treetop is necessary to enforce the "only one side effect per treetop" rule in the compiler. If the call node could swing down to some other treetop (as would have been possible without the fix) we could have more than side effect in that other treetop.

My only question is if there are any other IL opcodes for which a similar anchoring solution is needed in JitBuilder ?

mstoodle:
@vijaysun-omr it would be a node that's already a treetop that also has a value. I believe calls were the only ones in that category that weren't handled properly. Can you think of any others?

vijaysun-omr:
Thanks, one of the main points of asking my question was for that quick review to check if there were other opcodes to be handled the same way. I will merge this now.

PR #1210, opened 2017-06-02
Skip OSR related trees when splitting block

ncough:
An OSR point may have related trees, such as
pending push stores, immediately after it.
It is not acceptable to insert other trees in
this region. As block splitter may try to
uncommon, it must skip the entire OSR region
when creating the store.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -3087,6 +3087,11 @@ TR_HandleInjectedBasicBlock::createTemps(bool replaceAllReferences)
             if (tt->getNode()->getOpCode().isBranch() || tt->getNode()->getOpCode().isSwitch())
                tt = tt->getPrevTreeTop();
 
+            // If this treetop is an OSR point, a store cannot be placed between it and the 
+            // transition treetop in postExecutionOSR
+            if (comp()->isPotentialOSRPoint(tt->getNode()))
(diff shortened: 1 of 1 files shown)


andrewcraik:
A sensible fix to keep a transition point and the associated pending pushes together in the same block to save a lot of confusion later on.

compiler/optimizer/Inliner.cpp
```diff
@@ -3087,6 +3087,11 @@ TR_HandleInjectedBasicBlock::createTemps(bool replaceAllReferences)
             if (tt->getNode()->getOpCode().isBranch() || tt->getNode()->getOpCode().isSwitch())
                tt = tt->getPrevTreeTop();
 
+            // If this treetop is an OSR point, a store cannot be placed between it and the 
+            // transition treetop in postExecutionOSR
```
vijaysun-omr: Do you need to check for postExecutionOSR here or is that done in some other way before this code gets called or when getOSRTransitionTreeTop is called ?

ncough: getOSRTransitionTreeTop will consider the mode and act correspondingly, returning the provided treetop in the preExecutionOSR case.

vijaysun-omr: Okay thanks for the explanation.

PR #1211, opened 2017-06-04
Improve alias info in FieldPrivatizer

gita-omr:
FieldPrivatizer does not privatize fields that have any aliases.
Improve this by checking if any of the field's aliases are actually used inside the loop:

  - collect all symbol references used in the loop in _allSymRefs
  - AND getUseDefAliases with _allSymRefs when checking if field can be privatized

### DIFF
--- a/compiler/optimizer/FieldPrivatizer.cpp
+++ b/compiler/optimizer/FieldPrivatizer.cpp
@@ -113,8 +113,18 @@ bool TR_FieldPrivatizer::isFieldAliasAccessed(TR::SymbolReference * symRef)
 
    if (symRef->getUseDefAliases().hasAliases())
       {
-      aliasAccessed = true;
+
+      TR_ASSERT(_allSymRefs[symRef->getReferenceNumber()] == true, "symRef should have been seen in the loop");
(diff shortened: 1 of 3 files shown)


0xdaryl:
Commit title must be < 50 chars per the CG and without a trailing period.

gita-omr:
@vijaysun-omr, could you take a look at this one? Thanks.

vijaysun-omr:
Is _allSymRefs cleared after one loop has been analyzed and before the next loop is analyzed ? Just making sure the results from one loop won't affect the analysis of the next.

gita-omr:
This is done per loopStructure, similar to _neverWritten, by these calls in FieldPrivatizer.cpp:
       if (isPredictableLoop <= 0)
          {
          vcount_t visitCount = comp()->incVisitCount();
          initializeSymbolsWrittenAndReadExactlyOnce(symRefCount, notGrowable);
          collectSymbolsWrittenAndReadExactlyOnce(loopStructure, visitCount);
          }

vijaysun-omr:
Thanks for the clarification. Looks okay to me.

PR #1212, opened 2017-06-05
Support long/double auto-SIMD operations on Power

gita-omr:
for spark-sql-perf.

DatasetPerfromance benchmark inside spark-sql-perf performs avg() operation on
a Dataset of long numbers. Since avg() returns double there is a conversion
from long to double. In order to support this we need to:

  - add vector long 'add' and 'sub' instructions to Power codegen
    (they are supported starting with Power8)
  - add new vector IL opcode for l2d


vijaysun-omr:
@zl-wang and @ymanton please review.

compiler/p/codegen/FPTreeEvaluator.cpp
```diff
@@ -652,7 +652,7 @@ TR::Register *OMR::Power::TreeEvaluator::vsplatsEvaluator(TR::Node *node, TR::Co
    else if (node->getDataType() == TR::VectorInt64)
       {
       TR::Register *srcReg = cg->evaluate(child);
-      TR::Register *trgReg = cg->allocateRegister(TR_VSX_VECTOR);
+      TR::Register *trgReg = cg->allocateRegister(TR_VRF);
 
       if (!disableDirectMove && TR::Compiler->target.cpu.id() >= TR_PPCp8 && TR::Compiler->target.cpu.getPPCSupportsVSX())
```
zl-wang: This is relatively limiting.  This operation is VSX operation in itself, but I presumed you changed VSX_VECTOR to VRF because you expected VMX operation will follow.  Otherwise, this change is not a good move.

gita-omr: Yeah, I actually meant to ask about this. If I don't make this change I get an ASSERT in vadd/vsubEvaluator. Since we allocate VRF for Int32 I thought it's ok to do that for Int64.

compiler/il/OMRILOpCodesEnum.hpp
```diff
@@ -534,6 +534,7 @@
    vcalli,     // indirect call returning a vector
    vternary,   // vector ternary operator
    v2v,        // vector to vector conversion. preserves bit pattern (noop), only changes datatype
+   vl2vd,      // vector to vector conversion. converts each long element to double
    vconst,     // vector constant
```
zl-wang: Do we expect vi2vf (vector integer to vector single precision floating point) or opCodes for other combinations of data types in the future?  I am wondering if we can treat conversions similarly to vadd/vsub etc, element data type being kept somewhere else.  I know there are two types involved in conversions, different from vadd/vsub etc.

gita-omr: Another good find. I made the opcode typed for two reasons:

1) there have been a lot of discussions about pros and cons of typeless opcodes. There is a possibility that we make them typed again
2) currently we need a very safe change and making vl2vd typed is much safer solution for now

compiler/p/codegen/FPTreeEvaluator.cpp
zl-wang: it looked like evaluators emitting VMX operations didn't try to/check  (but ASSERT/assuming) move the source vectors (if VSX_VECTOR not in range of VRF) to VRF. It is functionally correct to use the VRF subset of VSX_VECTOR, but you give up on half of the vector registers.

compiler/il/OMRILOpCodesEnum.hpp
zl-wang: Acceptable to me at least for now.

vijaysun-omr:
"Enhance" is not spelled correctly in the PR/commit titles.

0xdaryl:
Further to @vijaysun-omr 's comment, the PR title needs to be < 50 characters per the CG.

compiler/p/codegen/FPTreeEvaluator.cpp
gita-omr: I suspect long vectors have not been used on Power. Since we use VRF's for integers, perhaps the safest thing would be to do the same for longs. @zl-wang, @ymanton : what do you think? Let's discuss tomorrow if necessary. Thanks.

zl-wang: yes, this is the safest approach for the time being.  the reason is clear (VSX sources feeding into VMX operations) though.  to work it out completely, need VECTOR register type feedback in a certain way, or type checking and move.

vijaysun-omr:
Could you please also fixing the typo around "Enhance" in the commit message ? I will merge after you do so. Thanks.

gita-omr:
@vijaysun-omr, just changed the commit message. Thanks.

PR #1216, opened 2017-06-05
CMake and Appveyor improvements

dnakamura:
A number of small improvements  for appveyor and CMake builds.
See the individual commits for details


charliegracie:
@dnakamura why is the PortSysinfoTest.sysinfo_test0 disabled?

dnakamura:
@charliegracie  That test confirms that the executable name returned by the port lib matches what is passed via argv[0]. However for whatever reason, on windows, CTest does not do path conversion, so argv[0] has '/' directory separators instead of the expected '\'

cmake/platform.cmake
```diff
@@ -141,23 +151,34 @@ if(OMR_HOST_OS STREQUAL "win")
...
 	else()
 		set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -entry:_DllMainCRTStartup@12")
 	endif()
 
 	#strip out exception handling flags (added by default by cmake)
-	string(REPLACE "/EHsc" "" filtered_cxx_flags ${CMAKE_CXX_FLAGS})
-	string(REPLACE "/GR" "" filtered_cxx_flags ${filtered_cxx_flags})
-	set(CMAKE_CXX_FLAGS "${filtered_cxx_flags} ${common_flags}")
+	omr_remove_option(CMAKE_CXX_FLAGS "/EHsc")
+	omr_remove_option(CMAKE_CXX_FLAGS "/GR")
```
rwy0717: I dislike mandating language feature flags, some uses might need these things. I'd rather these flags be configured externally.  If it's needed for CI,  can we (un)set these flags through a cmake cache or the environment?

cmake/caches/Appveyor.cmake
```diff
@@ -0,0 +1,15 @@
...
+set(OMR_OMRSIG ON CACHE BOOL "")
+set(OMR_FVTEST ON CACHE BOOL "")
+set(OMR_GLUE ${CMAKE_SOURCE_DIR}/example/glue)
+set(OMR_GC_SEGREGATED_HEAP ON CACHE BOOL "")
+set(OMR_GC_MODRON_SCAVENGER ON CACHE BOOL "")
+set(OMR_GC_MODRON_CONCURRENT_MARK ON CACHE BOOL "")
+set(OMR_THR_CUSTOM_SPIN_OPTIONS ON CACHE BOOL "")
+set(OMR_NOTIFY_POLICY_CONTROL ON CACHE BOOL "")
+set(OMR_THR_SPIN_WAKE_CONTROL ON CACHE BOOL "")
+set(OMR_THR_SPIN_CODE_REFACTOR ON CACHE BOOL "")
```
dnakamura: I was thinking about it, but I'm thinking in the not too distant future the two caches will likely diverge

cmake/platform.cmake
dnakamura: I agree in general.  However I would consider it out of scope for this PR. Will likely take that into account in my upcoming platform config refactor.  Also as it stands this does not force flags for the user. This is all variable scoped, however AFAIK,  user set cache variables will take precedence

rwy0717:
looks good to me.

PR #1217, opened 2017-06-05
Fix properties of VLVG instruction (z Systems)

hjooybar:
“VECTOR LOAD VR ELEMENT FROM GR” (VLVG instruction) stores a general register in a vector register.
For double-word element size (mask = 3) this instruction uses all 64 bits of its source register. Therefore, it has to be marked as a 64bit instruction.
The same logic applies to VLVGP instruction as well.


vijaysun-omr:
Code looks okay to me. I will wait for checks to pass.

vijaysun-omr:
I am going to merge this even though the AppVeyor (windows) build has not passed; it is a z codegen change, so that should be fine.

PR #1218, opened 2017-06-05
Add public API to get OS kernel info

babsingh:
omrsysinfo_os_kernel_info has been added to the port library in
order to retrieve kernel information. Only Linux kernel is supported at
this point; Linux kernel version is retrieved.

### DIFF
--- a/fvtest/porttest/si.cpp
+++ b/fvtest/porttest/si.cpp
@@ -2048,7 +2048,7 @@ TEST(PortSysinfoTest, sysinfo_test_get_os_description)
 	intptr_t rc = 0;
 	reportTestEntry(OMRPORTLIB, testName);
 
-	OMROSDesc desc;
+	struct OMROSDesc desc;
 	rc =  omrsysinfo_get_os_description(&desc);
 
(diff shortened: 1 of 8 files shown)


babsingh:
@charliegracie @rwy0717 in `omr/port`, there are folders for `linuxppc64` and `linuxs390`. these folders don't have a definition of `omrsysinfo.c`. will `linuxppc64` and `linuxs390` end up using `common/omrsysinfo.c` or `unix/omrsysinfo.c`?

rwy0717:
Linux will fall back to `unix/*` before `common/*`.

charliegracie:
@babsingh Is this an API that should be more generic and not Linux specific?

babsingh:
@charliegracie yes, it can be an API to get kernel version for any OS kernel. At this point, we don't have need to support other kernels. 

I will create a data struct to store kernel information and rename API to sound generic. is this sufficient?

charliegracie:
Yes that sounds perfect.

babsingh:
@charliegracie commit has been updated to reflect a generic API.

babsingh:
@charliegracie would you like me to make changes to this pull request?

PR #1219, opened 2017-06-06
Allow RMODE64 to be disabled via an option

fjeremic:
Introduce an option disableRMODE64 to allow the user to override the
RMODE64 option and disable it. In addition enable code cache
consolidation on z/OS 64-bit by default and disable it on z/OS 64-bit if
RMODE64 is disabled.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -488,6 +488,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableRegisterPressureSimulation",  "O\tdon't walk the trees to estimate register pressure during global register allocation", SET_OPTION_BIT(TR_DisableRegisterPressureSimulation), "F"},
    {"disableRematerialization",           "O\tdisable rematerialization",                      TR::Options::disableOptimization, rematerialization, 0, "P"},
    {"disableReorderArrayIndexExpr",       "O\tdisable reordering of index expressions",        TR::Options::disableOptimization, reorderArrayExprGroup, 0, "P"},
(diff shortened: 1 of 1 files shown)


fjeremic:
@mpirvu could you please review?

compiler/control/OMROptions.cpp
```diff
@@ -3667,10 +3672,13 @@ OMR::Options::jitPostProcess()
    // (~256MB) code caches at startup. The code caches will then have locality which improves performance, but
    // moreover the size of the allocated larger code cache will ensure no trampolines are needed in JIT private
    // linkage.
-   if (self()->getOption(TR_EnableRMODE64))
+   //
+   // Code cache consolidation is enabled on all 64-bit platforms by default. However if RMODE64 is not available
+   // on the OS or if the user specified to disable it we must also disable code cache consolidation.
+   if (!self()->getOption(TR_EnableRMODE64))
 #endif
```
mpirvu: This #endif is misplaced. You need to push it down near the next #endif, otherwise codecacheconsolidation is disabled on all platforms except zOS.

fjeremic: Wow pretty big mistake on my part. Thanks for catching this. Fixed in da3e152.

0xdaryl:
Your commit message says "In addition enable code cache consolidation on all 64-bit platforms by default...".  Wasn't code cache consolidation for 64-bit already enabled by default?  I just want to make sure this PR isn't overstating what its doing.

fjeremic:
> Your commit message says "In addition enable code cache consolidation on all 64-bit platforms by default...". Wasn't code cache consolidation for 64-bit already enabled by default? I just want to make sure this PR isn't overstating what its doing.

No, it was enabled for all (64-bit) platforms **except** z/OS.

mpirvu:
@0xdaryl Please let me know when you are merging this PR, so that I can merge its counterpart in tr.open.

PR #1221, opened 2017-06-06
Misc memory leaks and initialization fixes

youngar:


youngar:
Changed to WIP as there are errors in consumer builds with this change.  I'll update when I determine or fix the cause.

youngar:
@charliegracie this is good to go

PR #1222, opened 2017-06-06
Fixes for DDR

youngar:
Continuing from #1014.

This PR introduces the following changes:
- Code reorganization
- Cmake compilation support
- Two phase superset generation:
  - generate IL for all compilation units
  - combine generated IL into one cohesive IL
- Many small improvements to better support IL generation for different platforms and compilers 
- Building DDR in our CI


mstoodle:
Uh, any update on what's going on in this PR? :) @youngar ?

ddr/include/ddr/ir/Modifiers.hpp
```diff
@@ -19,7 +19,7 @@
 #ifndef MODIFIERS_HPP
 #define MODIFIERS_HPP
 
-#include <string>
+#include "ddr/std/string.hpp"
 #include <vector>
```
rwy0717: Does vector also need a compatibility header?

ddr/include/ddr/ir/Type.hpp
```diff
@@ -22,7 +22,7 @@
 #include "ddr/config.hpp"
 
 #include <set>
```
rwy0717: And set too?

ddr/include/ddr/ir/Modifiers.hpp
jason-hall: I can't think of a reason it would

youngar:
I have created a CQ and uploaded a patch for this change.  For those who are members of the project, the CQ is here: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13962.  I will try to keep this PR updated with the CQ status.

youngar:
This CQ has been approved!

PR #1226, opened 2017-06-07
Safely dup trees for loop versioning tests

ymanton:
Insure that all loop versioning tests that use duplicated nodes from inside the loop do so safely by making sure that the duplicated nodes' flags are cleared, since the flags may no longer be accurate outside the loop.

This is a follow up fix to 72f1d48904936cc352e52516e81ea7bb5353a8e5, which addressed one guard test that was known to be the cause of a downstream bug. This patch addresses all the remaining guard tests, including another test (that checks the induction variable against the loop lower bound) that is causing another bug for downstream.

The tests modified by this patch were identified by looking for

* nodes being duplicated via `duplicateTree()` and
* being consumed directly or indirectly by `if*cmp*` nodes that are
* inserted prior to the loop, and guarded with a `performTransformation()`

and replacing `duplicateTree()` calls with `duplicateTreeForCodeMotion()`. In some cases the duplicated nodes had their flags manually cleared because the author was aware of the need to do so; such code was removed because it is now redundant. The many other calls to `duplicateTree()` were left intact because it was not clear what the intention of the auther was or whether or not node flags should be cleared.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -4817,13 +4817,13 @@ void TR_LoopVersioner::buildNullCheckComparisonsTree(List<TR::Node> *nullChecked
                TR::Node *dupInvariantNullCheckReference;
                if(nullCheckReferenceisField)
                   {
-                  dupInvariantNullCheckReference=nextNode->getData()->duplicateTree();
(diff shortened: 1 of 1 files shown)


ymanton:
FYI @vijaysun-omr, @andrewcraik, this fixes a high priority bug for downstream.

andrewcraik:
Looks good to me.

vijaysun-omr:
Looks okay to me. Merging.

PR #1228, opened 2017-06-08
Fix optimization manager enabled field

xliang6:
Initialize optimization manager enabled field based on its related
options flag and start using the manager field to query from then on.

### DIFF
--- a/compiler/optimizer/OMROptimizationManager.cpp
+++ b/compiler/optimizer/OMROptimizationManager.cpp
@@ -49,7 +49,7 @@ OMR::OptimizationManager::OptimizationManager(TR::Optimizer *o, OptimizationFact
         _numPassesCompleted(0),
         _optData(NULL),
         _optPolicy(NULL),
-        _enabled(false),
+        _enabled(!o->comp()->isDisabled(optNum)),
         _requested(false),
         _requestedBlocks(o->trMemory()),
         _trace(o->comp()->getOptions()->trace(optNum))
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -198,7 +198,7 @@ int32_t TR_TrivialInliner::perform()
    comp()->generateAccurateNodeCount();
 
    TR::ResolvedMethodSymbol * sym = comp()->getMethodSymbol();
-   if (sym->mayHaveInlineableCall() && !comp()->isDisabled(OMR::inlining))
+   if (sym->mayHaveInlineableCall() && optimizer()->isEnabled(OMR::inlining))
       {
       uint32_t initialSize = comp()->getOptions()->getTrivialInlinerMaxSize();;
       if (comp()->getOption(TR_Randomize) || comp()->getOption(TR_VerbosePseudoRandom))
@@ -245,7 +245,7 @@ TR_InlinerBase::TR_InlinerBase(TR::Optimizer * optimizer, TR::Optimization *opti
    _util = optimizer->getInlinerUtil();
    _policy->setInliner(this);
    _util->setInliner(this);
-   if (comp()->getOptions()->isDisabled(OMR::innerPreexistence))
+   if (!optimizer->isEnabled(OMR::innerPreexistence))
       _disableInnerPrex = true;
 
    setInlineVirtuals(true);
@@ -1058,7 +1058,7 @@ bool
 TR_InlineCall::inlineCall(TR::TreeTop * callNodeTreeTop, TR_OpaqueClassBlock * thisClass, bool recursiveInlining, TR_PrexArgInfo *argInfo, int32_t initialMaxSize)
    {
    TR_InlinerDelimiter delimiter(tracer(),"TR_InlineCall::inlineCall");
-   if (comp()->isDisabled(OMR::inlining))
+   if (!getOptimizer()->isEnabled(OMR::inlining))
       return false;
 
    TR::Node * parent = callNodeTreeTop->getNode();
--- a/compiler/optimizer/LoopReducer.cpp
+++ b/compiler/optimizer/LoopReducer.cpp
@@ -4438,7 +4438,7 @@ TR_LoopReducer::perform()
    // enable only if the new loop reduction framework is
    // disabled
    //
-   if (!comp()->getOptions()->isDisabled(OMR::idiomRecognition))
+   if (optimizer()->isEnabled(OMR::idiomRecognition))
       {
       dumpOptDetails(comp(), "idiom recognition is enabled, skipping loopReducer\n");
       return 0;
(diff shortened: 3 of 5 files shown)


compiler/optimizer/OMROptimizationManager.cpp
```diff
@@ -49,7 +49,7 @@ OMR::OptimizationManager::OptimizationManager(TR::Optimizer *o, OptimizationFact
         _numPassesCompleted(0),
         _optData(NULL),
         _optPolicy(NULL),
-        _enabled(false),
+        _enabled(!o->comp()->isDisabled(optNum)),
```
andrewcraik: I am not sure that is enough... the optimizer dispatch loop checks a regex as well as the isDisabled call (see line 1624 onward in OMROptimizer.cpp).

If we are going to standardize this logic then optimizer invocation sequence should be using this same query for its early exit at the point indicated.

Further rather than always writing optimizer()->isEnabled.... wouldn't it make sense for the root Optimization to gain an enabled() or disabled() function that wraps up this logic so it is easy to redirect the handling again if desired in the future?

A compile-time verification is needed to make sure that the refactoring isn't going to increase compile time.

mstoodle: @andrewcraik agree with most of that, but compile time verification? you're really worried about the compile time impact of the queries to check if an optimization is enabled/disabled?

If you're worried about compile time regressions for this change, then I think you should be concerned about compile time regressions for any change to any optimization, which seems a trifle overboard, wouldn't you agree?

andrewcraik: My main concern in regards to compile-time was just recent experience with self() calls in extensible classes where the self() wasn't inlined leading to a hard-to-find compile-time increase. You may be right in this case since in many cases it is going to replace existing calls and could memoize some of the costs of the existing checking logic. To be honest most of the optimizer changes do a have a compile-time verification these days to avoid creep over time, but I agree it probably isn't strictly necessary as long as the number of indirects per query etc looks to be consistent for the most part.

xliang6: Thanks for the feedback, @andrewcraik @mstoodle 

> I am not sure that is enough... the optimizer dispatch loop checks a regex as well as the isDisabled call (see line 1624 onward in OMROptimizer.cpp).

> If we are going to standardize this logic then optimizer invocation sequence should be using this same query for its early exit at the point indicated.

The regex logic from line 1624 onward in OMROptimizer.cpp is currently **only** used in the optimizer dispatch loop and all the other consumers today (mostly other optimizations) are using comp()->isDisabled(optNum) directly. So adding this logic to the isDisabled/isEnabled method will potentially change the current behavior in a number of places.

> Further rather than always writing optimizer()->isEnabled.... wouldn't it make sense for the root Optimization to gain an enabled() or disabled() function that wraps up this logic so it is easy to redirect the handling again if desired in the future?

We do have an enabled() query on the root OptimizationManager class today. The optimizer()->isEnabled method was added as a wrapper for the underlying enabled() method as a convenience.

So another option that just occurred to me is that the isEnabled query on optimizer can contain the regex logic and the enabled query on optimization doesn't. This allows consumers to pick and choose what kind of checking they want to perform. However, this might generate more confusion as two seemingly similar queries actually do different checking under the hood?

andrewcraik: What is the use case for ignoring the regex in some places? When can we run an optimization if the regex says no and we don't want to know that in the optimization?

xliang6: Here is one use case I can think of that might generate different outcomes if we change how isDisabled/isEnabled works:

Opt A runs and checks to see whether opt B is disabled or not and performs differently based on the query result. Today, it only uses comp()->isDisabled(optNum) query without checking against the regex. If we add the regex check in the new isDisabled/isEnabled query, this can change what opt A does.

mstoodle: do you always have an opt index in hand to be able to do the regex test? .... I don't think Opt A can ask "is Opt B disabled" with the regex unless A knows what B's opt index *will* be.

This suggests to me that asking the isDisabled() question via the regex is a more "timely" kind of query (you have to ask it "in the moment", as it were) rather than a timeless notion implied by the other query. Maybe (and I'll be impressed if someone actually put this much thought into it) it's called "isDisabled()" because returning true means the opt won't run, but returning false doesn't necessarily mean it will run (because it might fail the regex check). Normally I don't like "negative" queries because I hate to see code like !isDisable() but, in this case, it might be the most correct question to ask...

xliang6: Agreed, opt A doesn't know (at least not easily) what opt B's opt index is when the optimizer calls B's perform so the regex opt index check has to happen in the optimizer dispatcher loop but the regex opt name check can still be added to isDisabled() if desired. IMO, regex opt index check can be a part of query asking whether an opt will run or not; whereas regex opt name check is a part of query asking whether an opt has been disabled or not (though that is not how the code has been working so far).

Please note that the query isDisabled() currently only returns true if a command-line option like disableFoo has been detected under the original assumption that all opts are enabled unless  instructed otherwise. So asking isDisabled() doesn't necessarily mean we are asking "the most correct question". It could actually be misleading today since a project using OMR may not even support opt Foo for instance. This was addressed with the new query implementation in this PR.

andrewcraik: Wow, how awkward. In light of the above, I agree with the approach of just trying to make it mirror the current enable/disable state in light of the above discussion (the ramifications of the regex did not dawn on me until the above discussion).

mstoodle: I'm not sure I agree that _enabled(!o->comp()->isDisabled()) is necessarily more correct than _disabled(o->comp()->isDisabled()), especially given the common case is that every optimization in the strategy is enabled unless disabled.

Q: If an opt has been disabled, do we still need to allocate an entire OptimizationManager for it so that it can tell anyone who asks (optimizer?) that it's not enabled?

xliang6: It can be better explained with the following use case: Opt A will do X if opt B is enabled. Currently if opt B is not even supported by a project, !comp()->isDisabled() will still return true, which is wrong. With this PR, if an opt is not supported by a project, we will not create an optimization manager for that opt so its _enabled field is nonexistent. Instead we will call optimizer->isEnabled() which in turn checks to see if an opt is supported first and if it's not, it will return false; otherwise it will return the value of the opt's _enabled field.

A: Yes, an opt manager is created even if the opt has been disabled (but is supported), however the opt class it manages will not be created. The opt manager class is a very lightweight class meant to hold "static" information about an opt such as its name, opts in its group if it's a group of opts, flags it needs to set, project-specific heuristics for the opt (which was added later), etc. The actual opt class is the one holding any data structures it needs during a particular pass of an opt. This can take a lot of memory depending on the opt and it used to live for the duration of the optimizer whether it was disabled or not.

PR #1229, opened 2017-06-08
Add implementation of Tril

Leonardo2718:
Tril is a small language/tool for generating Testarossa IL (Intermediate Language). This change adds a prototype for the trilc compiler that reads, compiles, and executes Tril code.

Eventually, Tril should evolve into a tool for testing Testarossa. For now, only a very simple implementation of the compiler as well as some documentation and examples are contributed.

### DIFF
--- a/fvtest/tril/.gitignore
+++ b/fvtest/tril/.gitignore
@@ -1,0 +1,2 @@
+build*/
+*.user
--- a/fvtest/tril/CMakeLists.txt
+++ b/fvtest/tril/CMakeLists.txt
@@ -41,6 +41,7 @@ add_executable(trilc
     ${FLEX_tril_scanner_OUTPUTS}
     ast.c
     ilgen.cpp
+    main.cpp
 )
 
 target_include_directories(trilc PUBLIC
--- a/fvtest/tril/CMakeLists.txt
+++ b/fvtest/tril/CMakeLists.txt
@@ -1,0 +1,62 @@
+###############################################################################
+#
+# (c) Copyright IBM Corp. 2017
+#
+#  This program and the accompanying materials are made available
+#  under the terms of the Eclipse Public License v1.0 and
+#  Apache License v2.0 which accompanies this distribution.
+#
+#      The Eclipse Public License is available at
+#      http://www.eclipse.org/legal/epl-v10.html
+#
+#      The Apache License v2.0 is available at
+#      http://www.opensource.org/licenses/apache2.0.php
+#
+# Contributors:
+#    Multiple authors (IBM Corp.) - initial implementation and documentation
+###############################################################################
+
+cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+
+project(tril LANGUAGES C CXX)
+
+find_package(BISON)
+find_package(FLEX)
+
+set(OMR_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../../omr)
+set(JITBUILDER_PATH ${OMR_PATH}/jitbuilder/release)
+
+set(CMAKE_POSITION_INDEPENDENT_CODE True)
+
+add_definitions(-g -fdiagnostics-color=always -std=c++0x -fno-rtti -fno-threadsafe-statics -Wno-deprecated -Wno-enum-compare -Wno-invalid-offsetof -Wno-write-strings -O3 -pthread -fomit-frame-pointer -fasynchronous-unwind-tables -Wreturn-type -fno-dollars-in-identifiers -m64 -fPIC -fno-strict-aliasing   -DBITVECTOR_BIT_NUMBERING_MSB -DUT_DIRECT_TRACE_REGISTRATION -DJITTEST -DJITBUILDER_SPECIFIC -DPROD_WITH_ASSUMES -DTR_HOST_X86 -DTR_HOST_64BIT -DBITVECTOR_64BIT -DLINUX -DTR_TARGET_X86 -DTR_TARGET_64BIT -DSUPPORTS_THREAD_LOCAL -D_LONG_LONG -DJ9HAMMER)
+
+BISON_TARGET(tril_parser tril.y ${CMAKE_CURRENT_BINARY_DIR}/tril.parser.c)
+FLEX_TARGET(tril_scanner tril.l ${CMAKE_CURRENT_BINARY_DIR}/tril.scanner.c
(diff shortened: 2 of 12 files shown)


ymanton:
Nice job. Just a few comments from my quick look.

Is there any value in having to explicitly specify a block's fallthrough? From what I can tell it looks like you require it if the block ends in a conditional jump or a non-jump but not a goto, so why not just infer it in the first two cases as well? It also seems like it leaves the door open for someone to create the wrong CFG if they accidentally specify a fallthrough other than the following block.

Does every block have to be named, even if nothing branches to it? I guess because `fallthrough=` exists it forces you to then name everything. Unless it adds to the expressiveness of the language and/or is required due to other limitations I think people will appreciate this more if it can reduce their mental burden (without complicating its own implementation too much).

Having to know the offset of indirect loads and stores ahead of time is problematic -- probably OK for test programs and if you're dealing strictly with blobs of memory under your control, but I'm struggling to think of how to cope with layouts that can't be known ahead of time.

Speaking of testing, since you're using JitBuilder this is not really aimed at testing is it? It seems like it's aimed at writing standalone programs. How do you see this being used in the test compiler context?

Leonardo2718:
@ymanton 

> Is there any value in having to explicitly specify a block's fallthrough? From what I can tell it looks like you require it if the block ends in a conditional jump or a non-jump but not a goto, so why not just infer it in the first two cases as well? It also seems like it leaves the door open for someone to create the wrong CFG if they accidentally specify a fallthrough other than the following block.

I am definitely considering this as a point for future improvement. My original thinking was that the CFG should be explicit built so that users can generate wrong CFGs for testing purposes. I'm not convinced though that this provides much value, so I'm open to changing it.

> Does every block have to be named, even if nothing branches to it? I guess because fallthrough= exists it forces you to then name everything. Unless it adds to the expressiveness of the language and/or is required due to other limitations I think people will appreciate this more if it can reduce their mental burden (without complicating its own implementation too much).

Naming a block is optional. The name is only used to specify branch and fallthrough targets. So, if a block isn't used as a target, it doesn't need a name. So far, naming every block has helped _me_ reduce the mental burden of writing the examples. I won't know whether this applies for others as well until I get their feedback though :stuck_out_tongue: .

> Having to know the offset of indirect loads and stores ahead of time is problematic -- probably OK for test programs and if you're dealing strictly with blobs of memory under your control, but I'm struggling to think of how to cope with layouts that can't be known ahead of time.

You're absolutely right. While developing Tril, I did try to focus on testing rather than standalone programs (see next point). However, being able to handle layouts that are unknown ahead of time could be useful. I'll have to do a bit more thinking on how to handle this.

> Speaking of testing, since you're using JitBuilder this is not really aimed at testing is it? It seems like it's aimed at writing standalone programs. How do you see this being used in the test compiler context?

Excellent question! The main goal of Tril _is_ indeed to write tests. The fact that you _can_ write a standalone program is mostly a side effect of trying to be keep as much generality as I can.

I used JitBuilder as a way to jump-start the development work. Eventually, I would like to drop the dependency on JitBuilder in favor of something that provides more control / less overhead.

My vision for Tril goes something like this: when writing tests for a fancy new compiler feature, input Trees are written in the Tril language. A test can then be written that takes these Trees as input and feeds them to the component(s) of the compiler being tested. The test compiler should provided the necessary machinery for setting up the test and compilation environments.

This is only the beginning and there's still a lot of work to be done and questions to be answered. Currently, the trilc compilers serve as a convenient way of demoing what can be done using Tril. The Mandelbrot example is a bit of an extreme case. I don't see many use cases for programs that complex.

Thanks for the feedback! :slightly_smiling_face:

mgaudet:
I think before we merge this, we need two more things, which really, could be the same thing -- as much as I know @Leonardo2718 hates this :P 

1. A set of _small_ examples; ie, it'd be good to see how you'd do `int foo(int a) { return a + 2; }` 
2. A small amount of validation testing -- just to verify it's 'working' as expected, should someone want to play around with it.

ymanton:
Further to @mgaudet's comment, a description of the syntax and a definition of the grammar are probably the most important pieces of documentation besides how to build and run.

> Naming a block is optional. The name is only used to specify branch and fallthrough targets. So, if a block isn't used as a target, it doesn't need a name. So far, naming every block has helped me reduce the mental burden of writing the examples. I won't know whether this applies for others as well until I get their feedback though :stuck_out_tongue: .

Since `fallthrough` is not optional that makes the names defacto required though, right? Anyhow, I don't feel strongly about it either way.

Leonardo2718:
@mgaudet 

More examples are definitely going to be required, and I have plans for them. Currently, the file `test.tril` contains an implementation for `int foo(int x) { return x >= 0 ? x + 1 : x - 1; }`, which is a start.

However, the `*.tril` file is only half of what's *currently* needed to have a working example. The actual compilation and method dispatch also have to be implemented. In addition, the current implementation of the MethodBuilder instance limits the type of functions that can be compiled. So, creating new examples actually requires changing (very few) parts of the `trilc` implementation. This is definitely *not* ideal and I have plans for resolving these issues very soon. Once the tool starts to become usable, I will add examples accordingly.

With regards to testing, the Mandelbrot example is actually the best I have right now in terms of system verification. I have more unit / automated tests coming though (it's all part of the plan).

@ymanton 

> Since `fallthrough` is not optional that makes the names defacto required though, right?

Actually, `fallthrough` is also optional. At least according to the rules of the language (yeah, I really need to document it). Tril is supposed to provide very fine control over the IL that gets generated and even allow invalid IL to be generated. So, I tried to make the default behavior "do nothing", even if the output is invalid IL or a wrong CFG. The goal is to reduce the amount of "magic" involved as much as possible; nothing is generated unless explicitly asked for. I know the implementation is not quite there yet but that's what I'm aiming for. This may prove to be the wrong approach, but it seems reasonable for now.

To all:
As I've already mentioned, there's still a lot of work to be done and many questions to be answered. I'm hoping that merging this (very incomplete) change is going to allow others to help and contribute. Once merged, I will start opening issues for the various items I have planned.

Leonardo2718:
Based on feedback so far, I've been re-thinking Tril's inner workings and architecture quit a bit. I have also significantly improved the code, building on the work in this PR.

As a follow up, I will be opening another PR shortly with architectural improvements, testing, and documentation.

@0xdaryl if you can merge this PR, I will open the new PR when ready.

PR #1232, opened 2017-06-08
Revert x86 arraytranslate register dependencies

jdmpapin:
Recent changes to the register dependencies for `arraytranslate` on x86 rely on the new helpers. These changes are reverted in order to allow the previous helpers to run again. However, the addition of `xmm3` is retained because it is a necessary fix.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -4275,12 +4275,13 @@ TR::Register *OMR::X86::TreeEvaluator::arraytranslateEvaluator(TR::Node *node, T
    TR::Register *dummy1 = cg->allocateRegister();
    TR::Register *dummy2 = cg->allocateRegister(TR_FPR);
    TR::Register *dummy3 = cg->allocateRegister(TR_FPR);
+   TR::Register *dummy4 = cg->allocateRegister(TR_FPR);
 
    bool arraytranslateOT = false;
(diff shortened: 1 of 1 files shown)


jdmpapin:
I don't know why the AppVeyor claims that this PR is "non-mergeable." Currently it should even be fast-forwardable to 23d7c31 from 59c3544

0xdaryl:
I should have pulled on this thread earlier, but I didn't so I'll do it now.  

I'm a bit uncomfortable that an evaluator in OMR is susceptible to the linkage requirements of a helper in an unseen downstream project.  This is compounded by the fact that the purpose of this evaluator is to simply call such a helper, and there is no implementation of that helper available in OMR.

Arguably, this is an artifact of the initial compiler technology contribution, but that doesn't make it right.

I see a couple of courses of action:

1)  If the code is truly project-neutral and a benefit to OMR, I suggest surfacing the runtime code from your project into OMR (provided you are allowed to do so) so this evaluator has a purpose and the linkage dependencies are obvious.

2)  If it's not project neutral, I suggest that the contents of this evaluator be folded away back into your downstream project where there is a tighter coupling between your helper and this evaluator.

andrewcraik:
I agree with @0xdaryl that the code as currently factored doesn't really make sense for OMR and this probably needs to move downstream. Since the issue here is a correctness issue affecting our use of OMR I'd really appreciate it if we can merge this fix and track the larger item of moving this code in an issue here on GitHub. I realize this is not an ideal solution and an exception to usual rules, but it would be greatly appreciated to help fix a problem we are experiencing while clean-up is done.

@jdmpapin can you please create an issue to track cleaning this up?

PR #1238, opened 2017-06-09
Adding Ibyteswap evaluator on Power

R-Santhir:
This is change adds the evaluator and codegen query on Power. The ibyteswap unit test is also enabled on Power as well.

This is a follow-up to #936

### DIFF
--- a/compiler/p/codegen/OMRCodeGenerator.hpp
+++ b/compiler/p/codegen/OMRCodeGenerator.hpp
@@ -172,6 +172,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
 
    bool is64BitProcessor();
 
+   bool getSupportsIbyteswap();
+
    void generateBinaryEncodingPrologue(TR_PPCBinaryEncodingData *data);
 
    void beginInstructionSelection();
--- a/compiler/p/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/p/codegen/OMRTreeEvaluator.cpp
@@ -6220,3 +6220,40 @@ TR::Register *OMR::Power::TreeEvaluator::retrieveTOCRegister(TR::Node *node, TR:
 {
    return cg->machine()->getPPCRealRegister(TR::RealRegister::gr2);
 }
+
+TR::Register * OMR::Power::TreeEvaluator::ibyteswapEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+{
+   TR_ASSERT(node->getNumChildren() == 1, "Wrong number of children in ibyteswapEvaluator");
+
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *tgtRegister = cg->allocateRegister();
+
+   if (!firstChild->getRegister() &&
+       firstChild->getOpCode().isMemoryReference() &&
+       firstChild->getReferenceCount() == 1)
+      {
+      TR::MemoryReference *tempMR = new (cg->trHeapMemory()) TR::MemoryReference(firstChild, 4, cg);
+      tempMR->forceIndexedForm(firstChild, cg);
+      generateTrg1MemInstruction(cg, TR::InstOpCode::lwbrx, node, tgtRegister, tempMR);
+      tempMR->decNodeReferenceCounts(cg);
+      }
+   else
+      {
+      TR::Register *srcRegister = cg->evaluate(firstChild);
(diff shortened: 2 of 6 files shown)


R-Santhir:
@ymanton can you take a look when you get a chance? It's a pretty straight forward change.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6220,3 +6220,41 @@ TR::Register *OMR::Power::TreeEvaluator::retrieveTOCRegister(TR::Node *node, TR:
 {
    return cg->machine()->getPPCRealRegister(TR::RealRegister::gr2);
 }
+
+TR::Register * OMR::Power::TreeEvaluator::ibyteswapEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+{
+   TR_ASSERT(node->getNumChildren() == 1, "Wrong number of children in inlineIntegerReverseBytes");
```
ymanton: The message string is referring to the wrong method.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6220,3 +6220,41 @@ TR::Register *OMR::Power::TreeEvaluator::retrieveTOCRegister(TR::Node *node, TR:
...
+
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *tgtRegister = cg->allocateRegister();
+
+   if (TR::Compiler->target.cpu.id() >= TR_PPCp7 &&
```
ymanton: I know the code you're refactoring had this constraint, but I don't think we need to limit this to >=P7, support for this instruction goes back a long way. @zl-wang, did we limit this to P7 and later for any particular reason?

compiler/p/codegen/OMRCodeGenerator.hpp
```diff
@@ -172,6 +172,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
 
    bool is64BitProcessor();
 
+   virtual bool getSupportsIbyteswap();
```
ymanton: This isn't your doing, but I don't think this needs to be virtual here or in the base class. Since this class is part of an extensible hierarchy you can override `getSupportsIbyteswap()` in the sub class without it being a virtual method (unless I am mistaken?).

compiler/p/codegen/TreeEvaluatorTable.hpp
```diff
@@ -748,5 +748,5 @@
    TR::TreeEvaluator::longNumberOfLeadingZeros,         // TR::lnolz (J9)
    TR::TreeEvaluator::longNumberOfTrailingZeros,        // TR::lnotz (J9)
    TR::TreeEvaluator::longBitCount,                     // TR::lpopcnt (J9)
-   TR::TreeEvaluator::unImpOpEvaluator,                 // TR::ibyteswap
+   TR::TreeEvaluator::ibyteswapEvaluator,                 // TR::ibyteswap
```
ymanton: Can you line the `// TR::ibyteswap` with the other comments above/below?

compiler/p/codegen/OMRTreeEvaluator.cpp
zl-wang: the checking for POWER7 and later should only be applied to using the double-word form of instructions (i.e.  ldbrx/stdbrx), since the word form (lwbrx/stwbrx) and half-word form (lhbrx/sthbrx) existed day one of PPC ISA.

R-Santhir: Okay I'm removing that part of the check. So there's no reason for it to have been there in the first place?

zl-wang: There is no reason for (>=p7) to be there for lwbrx/stwbrx/lhbrx/sthbrx instructions.  But it should be there for ldbrx/stdbrx instructions.  Please fix it consistently everywhere ...

vijaysun-omr:
@ymanton is this ready to be merged in your opinion ?

ymanton:
Yes, LGTM.

vijaysun-omr:
@R-Santhir could you please re-base (since this spent a few days on the review queue) ?

R-Santhir:
@vijaysun-omr rebased and all checks have passed

vijaysun-omr:
Thanks for adding a test case Ryan and re-basing. Merging.

gita-omr:

PR #1239, opened 2017-06-11
Fix incorrect target platform detection

amitin:


jitbuilder/build/platform/host/ia32-linux-gcc.mk
```diff
@@ -0,0 +1,24 @@
+################################################################################
+##
+## (c) Copyright IBM Corp. 2016, 2016
```
0xdaryl: Should be 2016, 2017

rwy0717:
The cmake part looks good to me. /cc @dnakamura

mstoodle:
Hi @amitin, this PR looks soooooo close to being merged: thanks for your efforts so far!

If you could update the commit message titles to meet our contribution guidelines as @0xdaryl requested :) then I think we can merge it.

PR #1241, opened 2017-06-12
Aborted CS: Handle indirect refs

amicic:
When fixing up references in root/RS scanning in Aborted Concurrent
Scavenger, also handle indirect references (references to Nursery from
off heap structures associated with the object). Language specific code
will actually do the fixup (for example, in Java it will fixup class
structure)


0xdaryl:
Minor nit:  please remove the '.' from your commit title per the CG.

PR #1242, opened 2017-06-12
Begin to remove FrontEnd dependence on CCM

0xdaryl:
Begin to eliminate `OMR::CodeCacheManager` dependence on deprecated FrontEnd functionality by either eliminating it outright or refactoring the FrontEnd functionality into it's own class.

### DIFF
--- a/compiler/env/OMRCompilerEnv.hpp
+++ b/compiler/env/OMRCompilerEnv.hpp
@@ -38,6 +38,7 @@ namespace OMR { typedef OMR::CompilerEnv CompilerEnvConnector; }
 #include "env/ObjectModel.hpp"
 #include "env/ArithEnv.hpp"
 #include "env/VMEnv.hpp"
+#include "env/VMMethodEnv.hpp"
 
 namespace TR { class CompilerEnv; }
 
@@ -73,6 +74,10 @@ public:
    // Information about the VM environment.
    //
    TR::VMEnv vm;
+
+   // Information about methods in this compilation environment
+   //
+   TR::VMMethodEnv mth;
 
    // Object model in this compilation environment
    //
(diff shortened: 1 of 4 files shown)


0xdaryl:
I found a potential problem.  This isn't quite ready to merge yet.

vijaysun-omr:
The 3 routines in VMMethodEnv currently were chosen because of their relationship to code cache management ? Would you be proactively moving more of the front end functionality to the class as well ?

0xdaryl:
Yes.  These were just the first few chosen based on their simplicity.  I have other PRs that will move more of the FrontEnd TR_OpaqueMethodBlock functions to this class.

vijaysun-omr:
Thanks. Was there something about this class that made you consider making it extensible (as opposed to a class inheritance/virtual method approach ?) Just wanted to get that reasoning made explicit so that we can make a choice based on the right criteria in other cases in the future.

0xdaryl:
The decision to make this class extensible follows the same reasoning as many of the other compiler classes within the base OMR project:  for encapsulation and performance.

This technique uses static polymorphism and allows each compiler project derived from OMR to encapsulate its own custom functionality within its own extension class without introducing those functions in the topmost OMR layer.  By making it extensible it also follows the precedent of each of the other classes that compose a CompilerEnv class.

Making the functions in VMMethodEnv virtual would have also added an extra level of indirection on each function call (the first being to find the VMMethodEnv instance on the CompilerEnv structure, and the second being the dispatch through vtable for the function).  Using extensible classes makes each of these function calls direct and determinable at compile-time.

vijaysun-omr:
Thanks for the summary. Merging now.

PR #1244, opened 2017-06-13
Fix usage of rbx to allocate a stack frame

### LINKED ISSUE #1243: JitBuilder uses incorrect x86-64 register to allocate small stack frames in function prologues
When allocating a stack frame that is 8-bytes or less on x86-64, the function prologue generated by JitBuilder uses a `push` of a dummy register. In the epilogue, it pops the same register to de-allocate the frame.

However, the dummy register currently used (on x86-64) is `rbx`, which is a preserved register. This is a problem because when the register is popped in the epilogue, it overwrites the value of the register with whatever was left on the stack.

Leonardo2718:
This change introduces the helper function `getSingleWordFrameAllocationRegister()` that returns an appropriate register for allocating/deallocating the stack on a given platform. Now, instead of pushing/popping rbx on x86, r11 is used on x86-64 and ecx is used on x86-32.

Fixes #1243


Leonardo2718:
@0xdaryl I have updated the change.

PR #1248, opened 2017-06-13
Replace redundant lwz with rlwinm

AgostonSzepessy:
Power processors aren't optimized for storing values and then fetching
them back, so the compiler optimizes away redundant loads after
stores. In some cases, lwz was used to clear the high 32 bits of a
variable, but the compiler optimized it away regardless. Replacing it
with rlwinm, which will clear the high 32 bits, solves this.


0xdaryl:
I'd appreciate a review from @ymanton .

Code changes look fine to me.  Could you change the capitalization of POWER in your commit message to Power?

AgostonSzepessy:
@0xdaryl I've changed the commit message.

ymanton:
LGTM.

PR #1249, opened 2017-06-13
Ignore generated file omr.rc

lmaisons:
omr.rc is a generated file introduced by commit 8ce8861 / a40dfc7.  Add
it to .gitignore to stop it from showing up in untracked changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ config.cache
 config.log
 config.status
 SPEC
+omr.rc
 
 OMR_VERSION_STRING
 OMR_JIT_VERSION_STRING


rwy0717:
LGTM, with nit: should probably be `/omr.rc`.

lmaisons:
I copied the other configure artifacts.  What's the difference?

jdmpapin:
`/omr.rc` will only ignore `omr.rc` in the repository root, and not in subdirectories.

rwy0717:
Yeah, since it is consistent, this looks good as-is.

PR #1251, opened 2017-06-14
Fix redundantGotoElim asynccheck insert for OSR

ncough:
Changes were made to redundantGotoElim to allow it
to perform under NextGenHCR without placing an
asynccheck where it may invalidate an assumption
without being protected by an OSR guard.

Incorrect changes were made when adding the possibility
of both pre and post execution in one compile.
Moreover, the modification to the optimization is
specific to how assumptions are made, not the OSR
mode.

### DIFF
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -3569,7 +3569,7 @@ int32_t TR_EliminateRedundantGotos::process(TR::TreeTop *startTree, TR::TreeTop 
             TR::Block *predBlock = current->getFrom()->asBlock();
             requestOpt(OMR::treeSimplification, true, predBlock);
 
-            if (asyncMessagesFlag && comp()->isOSRTransitionTarget(TR::postExecutionOSR))
+            if (asyncMessagesFlag && comp()->getHCRMode() != TR::osr)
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM

compiler/optimizer/LocalOpts.cpp
```diff
@@ -3661,7 +3661,7 @@ int32_t TR_EliminateRedundantGotos::process(TR::TreeTop *startTree, TR::TreeTop
       // Place an asynccheck as the first treetop of the successor, if there was one in the removed block
       // This is necessary as placing it in the predecessor may result in a seperation from its OSR guard
       //
-      if (asyncMessagesFlag && comp()->isOSRTransitionTarget(TR::postExecutionOSR))
+      if (asyncMessagesFlag && comp()->getHCRMode() == TR::osr)
```
vijaysun-omr: Should this be != TR::osr instead of == ?

ncough: This is the case where the asynccheck should be inserted under NextGenHCR, as it won't invalidate an assumption without an OSR guard. All cases being treated the same was a bug.

vijaysun-omr: I see. Thanks for clarifying this particular spot was different to the others in terms of correctness.

PR #1254, opened 2017-06-14
Remove liveness diff between post and pre OSR

ncough:
With the addition of argument stashing for OSR
infrastructure, pre and post OSR points no longer
require a distinction between their dead stores.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -528,8 +528,15 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
    if (shouldSplitBlock)
        firstHalfBlock->split(insertionPoint, self()->comp()->getFlowGraph(), true);
 
+   // The arguments for a transition to this bytecode index may have already been stashed.
+   // If they have, use these directly rather than copying the reference node.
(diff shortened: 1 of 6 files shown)


andrewcraik:
LGTM @vijaysun-omr

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1821,14 +1821,23 @@ OMR::ResolvedMethodSymbol::insertStoresForDeadStackSlotsBeforeInducingOSR(TR::Co
    TR::TreeTop *prev = induceOSRTree->getPrevTreeTop();
    TR::TreeTop *next = induceOSRTree;
 
-   TR::OSRTransitionTarget pointType = TR::preAndPostExecutionOSR;
+   /*
+    * Any arguments used by this bytecode index should be
+    * kept live
+    */
+   TR_BitVector *deadSymRefs = osrMethodData->getLiveRangeInfo(byteCodeIndex);
```
vijaysun-omr: Is this change also related to the pre/post OSR distinction that ceased to exist because of argument stashing ? I mean if we need to do this unconditionally now, why was it not being done under pre OSR before ?

ncough: This MR was missing a change so that pre OSR would use these stashed arguments when they were available, rather than copying a reference node. With this added, the copied symrefs must be kept live for the transition, so this is done unconditionally.

PR #1255, opened 2017-06-14
Add OSRInduceBlock flag and skip opts

ncough:
OSRInduceBlocks will be filled with dead stores
and remat trees, resulting in a significant
compile time overhead if not ignored.

This change:
 - Adds a flag to identify OSRInduceBlocks
 - Check the flag in several opts
 - Remove the prior expensive version of the check

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1710,43 +1710,51 @@ OMR::Block::getGlobalNormalizedFrequency(TR::CFG * cfg)
 
 /*
  * An OSR induce block is used in voluntary OSR to contain the OSR point's dead stores
- * and to induce the transition. It will always have the exit and at least one exception
- * successor to an OSR catch block.
+ * and to induce the transition. It will always have a single edge to the exit and at least one
+ * exception successor to an OSR catch block.
+ *
+ * This function verifies that any blocks marked as an OSRInduceBlock hold these properties.
  */
 bool
-OMR::Block::isOSRInduceBlock(TR::Compilation *comp)
+OMR::Block::verifyOSRInduceBlock(TR::Compilation *comp)
    {
-   // An OSR induce block should have one successor, the end, and will not exist in involuntary OSR
-   if (comp->getOSRMode() != TR::voluntaryOSR || self()->getSuccessors().size() != 1 ||
-       self()->getSuccessors().front()->getTo() != comp->getFlowGraph()->getEnd())
-      return false;
+   // This check only applies when in voluntary OSR
+   if (comp->getOSRMode() != TR::voluntaryOSR)
+      return true;
 
-   // At least one of the exception edges must be to an OSR catch block
-   auto e = self()->getExceptionSuccessors().begin();
-   for (; e != self()->getExceptionSuccessors().end(); ++e)
-      if ((*e)->getTo()->asBlock()->isOSRCatchBlock())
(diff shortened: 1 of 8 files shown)


compiler/il/OMRBlock.cpp
```diff
@@ -1708,46 +1708,6 @@ OMR::Block::getGlobalNormalizedFrequency(TR::CFG * cfg)
    return frequency;
    }
 
-/*
- * An OSR induce block is used in voluntary OSR to contain the OSR point's dead stores
- * and to induce the transition. It will always have the exit and at least one exception
- * successor to an OSR catch block.
- */
-bool
-OMR::Block::isOSRInduceBlock(TR::Compilation *comp)
```
andrewcraik: Don't we want to keep this to be able to validate and/or update the flag?

compiler/env/ExceptionTable.cpp
```diff
@@ -50,7 +50,7 @@ TR_ExceptionTableEntryIterator::TR_ExceptionTableEntryIterator(TR::Compilation *
       catchBlock = tt->getNode()->getBlock();
 
       // Skip OSR blocks
-      if (catchBlock->isOSRCodeBlock() || catchBlock->isOSRCatchBlock())
+      if (catchBlock->isOSRCodeBlock() || catchBlock->isOSRCatchBlock() || catchBlock->isOSRInduceBlock())
```
vijaysun-omr: I guess this is okay for now since we may never be inducing OSR in a catch block (?) but maybe this could be a place for an assertion to avoid having to debug a nasty exception table related problem in case someone did add an OSR induction in a catch block in the future (?)

compiler/optimizer/LiveVariableInformation.cpp
```diff
@@ -229,7 +229,7 @@ void TR_LiveVariableInformation::initializeGenAndKillSetInfo(TR_BitVector **regu
          block = node->getBlock();
          blockNum = block->getNumber();
          seenException  = false;
-         if (_ignoreOSRUses && (block->isOSRCodeBlock() || block->isOSRCatchBlock()))
+         if (_ignoreOSRUses && (block->isOSRCodeBlock() || block->isOSRInduceBlock() || block->isOSRCatchBlock()))
```
vijaysun-omr: I don't completely understand this change. The original code was checking if OSR transition related uses could be ignored (based on the _ignoreOSRUses boolean) and I think this change seems to ignore blocks containing OSR induction call too. I don't see how this is legal because there could other uses that have nothing to do with OSR in a block ending in an OSR induction call.

compiler/optimizer/OMRSimplifier.cpp
```diff
@@ -337,6 +337,13 @@ OMR::Simplifier::simplifyExtendedBlock(TR::TreeTop * treeTop)
       if (block && !b->isExtensionOfPreviousBlock())
          break;
 
+      if (b->isOSRCodeBlock() || b->isOSRCatchBlock() || b->isOSRInduceBlock())
```
vijaysun-omr: Again, can we afford to omit tree simplification (one of the few opts we require for functional correctness) in a block that could have other trees before the OSR induction call ? I can understand this in the limited context of an OSR catch/code block but is it legal for a block ending in an OSR inductio call ?

vijaysun-omr:
Thanks for the change. Merging.

PR #1256, opened 2017-06-14
Add helper method to check all points can OSR

ncough:
To simplify analysis, optimizations may ask if
there are any OSR points that cannot transition.
This adds a helper to determine if OSR is always
safe, excluding checks to the OSR infrastructure.


andrewcraik:
LGTM @vijaysun-omr please review

PR #1257, opened 2017-06-14
Use dense BitVectors in region analysis

ncough:
Use dense BitVectors, rather than sparse, to
implement region analysis. This reduces the
compile overhead, as allocations for the sparse
BitVectors are costly.

### DIFF
--- a/compiler/optimizer/StructuralAnalysis.cpp
+++ b/compiler/optimizer/StructuralAnalysis.cpp
@@ -43,15 +43,15 @@
 #include "ras/Debug.hpp"                              // for TR_DebugBase
 
 void TR_RegionAnalysis::simpleIterator (TR_Stack<int32_t>& workStack,
-                                        SparseBitVector& vector,
+                                        StructureBitVector& vector,
                                         WorkBitVector &regionNodes,
                                         WorkBitVector &nodesInPath,
                                         bool &cyclesFound,
                                         TR::Block *hdrBlock,
                                         bool doThisCheck)
    {
-   SparseBitVector::Cursor cursor(vector);
-   for (cursor.SetToLastOne(); cursor.Valid(); cursor.SetToPreviousOne())
+   StructureBitVector::Cursor cursor(vector);
+   for (cursor.SetToFirstOne(); cursor.Valid(); cursor.SetToNextOne())
       {
       StructInfo &next = getInfo(cursor);
 
@@ -345,7 +345,7 @@ TR_RegionStructure *TR_RegionAnalysis::findNaturalLoop(StructInfo &node,
 
    int32_t numBackEdges = 0;
 
-   SparseBitVector::Cursor cursor(node._pred);
+   StructureBitVector::Cursor cursor(node._pred);
    for (cursor.SetToFirstOne(); cursor.Valid(); cursor.SetToNextOne())
       {
       StructInfo &backEdgeNode = getInfo(cursor);
@@ -472,14 +472,14 @@ void TR_RegionAnalysis::addNaturalLoopNodes(StructInfo &node, WorkBitVector &reg
    regionNodes[index] = true;
    nodesInPath[index] = true;
 
-   SparseBitVector::Cursor cursor(node._pred);
+   StructureBitVector::Cursor cursor(node._pred);
    for (cursor.SetToFirstOne(); cursor.Valid(); cursor.SetToNextOne())
       {
       StructInfo &next = getInfo(cursor);
       if (_dominators.dominates(hdrBlock, next._originalBlock))
          addNaturalLoopNodes(next, regionNodes, nodesInPath, cyclesFound, hdrBlock);
       }
-   SparseBitVector::Cursor eCursor(node._exceptionPred);
+   StructureBitVector::Cursor eCursor(node._exceptionPred);
    for (eCursor.SetToFirstOne(); eCursor.Valid(); eCursor.SetToNextOne())
       {
       StructInfo &next = getInfo(eCursor);
@@ -604,7 +604,7 @@ void TR_RegionAnalysis::addRegionNodes(StructInfo &node, WorkBitVector &regionNo
    regionNodes[index] = true;
    nodesInPath[index] = true;
 
-   SparseBitVector::Cursor cursor(node._succ);
+   StructureBitVector::Cursor cursor(node._succ);
    for (cursor.SetToFirstOne(); cursor.Valid(); cursor.SetToNextOne())
       {
       StructInfo &next = getInfo(cursor);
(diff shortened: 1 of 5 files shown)


andrewcraik:
LGTM @vijaysun-omr

compiler/optimizer/StructuralAnalysis.hpp
```diff
@@ -58,7 +58,7 @@ class TR_RegionAnalysis
 
    class StructInfo;
    typedef CS2::TableOf<StructInfo, TR::Allocator> InfoTable;
-   typedef TR::SparseBitVector SparseBitVector;
+   typedef TR::BitVector SparseBitVector;
```
0xdaryl: This is quite subtle.  By the name alone, someone not deeply familiar with this code would likely be expecting a sparse data structure.  Perhaps a different name should be used.

andrewcraik:
So the logic behind the change is as follows: following a conversation with @lmaisons I was given the understanding that a SparseBitVector can become more efficient than a BitVector when the number of set bits is less than 1/16. In a SparseBitVector nearly every element being set causes an allocation. The cost of the allocation is readily seen in execution profiles - especially since when running with OSR on you need to run more passes of structural analysis and one of the most expensive parts of that analysis when profiled are the allocations for the SparseBitVector. Based on the average size of methods the savings of the SparseBitVector seem small compared to the cost of running the region memory allocators and the overhead associated with them.

I am open to hearing other points of view on this one, but the significant allocation overhead for SparseBitVector made it feel like this wasn't the right tradeoff any more. @mpirvu interested to hear your thoughts as well.

compiler/optimizer/StructuralAnalysis.hpp
andrewcraik: I agree - a rename here would be appropriate.

compiler/optimizer/Structure.cpp
```diff
@@ -363,8 +363,7 @@ void TR_RegionStructure::addGlobalRegisterCandidateToExits(TR_RegisterCandidate
    }
 
 
-
-static bool findCycle(TR_StructureSubGraphNode *node, TR::SparseBitVector &regionNodes, TR_BitVector &nodesSeenOnPath, TR_BitVector &nodesCleared, int32_t entryNode)
+static bool findCycle(TR_StructureSubGraphNode *node, TR::BitVector &regionNodes, TR_BitVector &nodesSeenOnPath, TR_BitVector &nodesCleared, int32_t entryNode)
```
mgaudet: Maybe what should be done is moving the `typedef` such that we can use the `SparseBitVector` typedef in this signature -- (after it has been renamed of course. Perhaps `StructureBitVector`?

compiler/optimizer/Structure.hpp
```diff
@@ -626,8 +626,8 @@ class TR_RegionStructure : public TR_Structure
         void reset() {_iter.SetToFirstOne();}
 
         private:
-          TR::SparseBitVector _nodes;
-          TR::SparseBitVector::Cursor _iter;
+          TR::BitVector _nodes;
+          TR::BitVector::Cursor _iter;
```
mgaudet: Ditto to my previous comment; perhaps the `BitVector` `typedef` ought to be in Structure.hpp?

compiler/optimizer/StructuralAnalysis.cpp
```diff
@@ -51,7 +51,7 @@ void TR_RegionAnalysis::simpleIterator (TR_Stack<int32_t>& workStack,
                                         bool doThisCheck)
    {
    SparseBitVector::Cursor cursor(vector);
-   for (cursor.SetToLastOne(); cursor.Valid(); cursor.SetToPreviousOne())
+   for (cursor.SetToFirstOne(); cursor.Valid(); cursor.SetToNextOne())
```
mgaudet: Do sparse and dense bitvectors order their bits differently? Or is this a separate optimization?

andrewcraik: Looking at the code there may be a small benefit to backwards iteration, but the main reason for the change is that the reverse traversal is only currently supported by the SparseBitVector::Cursor and not the BitVector::Cursor - it seems to be the only asymmetry in the API...

mpirvu:
I vaguely recollect that SparseBitVectors were mostly useful to the static compiler.
However, to cover all bases I think we should measure the impact of this change alone on startup-time/CPU/footprint.

lmaisons:
If the reverse iteration is beneficial, it might be worth revisiting an old patch I created to allow swapping out CS2::SparseBitVector with CS2::ABitVector using a pre-processor macro.

vijaysun-omr:
@ncough I am assuming this needs the evaluation requested by Marius to be done before it can be merged. I am waiting for an update on this before I consider it.

andrewcraik:
@ncough can we please work on getting the evaluation done for this one?

mpirvu:
I performed some startup experiments using the new Liberty 17.0.0.2. 

With SCC/AOT but no -Xquickstart (the way WAS teams runs it) dense bit vectors manage to:
1) Reduce footprint by 0.26% (negligible considering CI=0.22%)
2) Increase CompCPU by 1.3% (CI=0.6%)
3) Startup time is flat.

With SCC/AOT and -Xquickstart dense bit vectors manage to:
1) Reduce footprint by 0.35% (CI=0.22%, so noise)
2) Increase CompCPU by 0.44% (CI=0.4%, so noise)
3) Improve startup time by 0.48% (CI=0.43%, so it could be noise)

I believe the expectation was that dense b it vectors would increase footprint, but improve CompCPU. If I ignore fluctuations for a moment, the results above show the exact opposite behavior. Taking fluctuations into consideration I would say the behavior is more or less the same.

andrewcraik:
Marius evaluation shows noise - we would expect the dense BitVector to reduce the highwater mark in memory usage and, potentially, have some small improvement in compile time due to less indirection. These changes have a big impact on structure analysis which does not run much at cold. The impact will grow more for warm, hot and scorching compiles as structure is rebuilt more and the size of the structure increases.

compiler/optimizer/Structure.cpp
andrewcraik: I think that is probably a fine change, but it seems like a much bigger item than correcting the bad memory behaviour we are seeing with the current BitVector - I would be more comfortable decoupling these two issues.

compiler/optimizer/Structure.hpp
andrewcraik: Ditto to my reaction :)

ncough:
This change has been updated to address the naming concerns. 

Testing in a benchmark that consists mostly of warm compilations shows no significant difference to the memory usage high water mark and a small reduction in compile time. This reduction is also seen in perf profiles, where the SparseBitVector allocations are immediately apparent, further amplified by OSR due to its the reliance on structure during IlGen opts. With this change applied, structural analysis and the optimizations that require it show a reduction in compile time.

vijaysun-omr:
Looks ready to go in. Merging.

PR #1258, opened 2017-06-14
Fix pending push liveness shortcut

ncough:
Accidental processing of removed pending push trees
resulted in pending push temps remaining live to method
entry and other OSR points when they should not have.
Moreover, any loads with a reference count greater than 1 
are not loads specific to this OSR point's shortcut and should 
be ignored.


vijaysun-omr:
Seems like a valid change to me. Merging.

PR #1259, opened 2017-06-14
Calculate OSRDef during OSRLiveRangeAnalysis

ncough:
To reduce the compile time overhead of OSR, it is possible
to calculate which slots are being used at an OSR
point based on the set of live symrefs during
OSRLiveRangeAnalysis, rather than having to perform both
it and OSRDefAnalysis.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -41,6 +41,7 @@
 #include "il/TreeTop.hpp"                             // for TreeTop
 #include "il/TreeTop_inlines.hpp"
 #include "il/symbol/AutomaticSymbol.hpp"
+#include "il/symbol/ParameterSymbol.hpp"
 #include "il/symbol/ResolvedMethodSymbol.hpp"
 #include "infra/Array.hpp"                            // for TR_Array
 #include "infra/Assert.hpp"                           // for TR_ASSERT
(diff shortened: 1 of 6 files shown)


compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -596,6 +596,16 @@ int32_t TR_OSRDefAnalysis::perform()
          }
       return 0;
       }
+   else if (comp()->getOSRMode() == TR::voluntaryOSR)
```
andrewcraik: This doesn't account for the liveness analysis being disabled or anything like that...

vijaysun-omr:
I have a high level question about the optimization being done to save compile time here. I wonder if using the OSR liveness is the right answer instead of OSR def analysis. 

OSR def analysis is meant to specify which one of various local variables that share a particular slot is the one that is "in use" at a given program point. "In use" is not the same thing "is live" though. The example I am thinking of is

Let us assume t1 is a local variable that share the same auto slot as some other local variable.

t1 = RHS1

...use of t1;

OSR point

...no use of t1

t1 = RHS2

At this OSR point, t1 is not live since there are no downstream uses of t1. However, if an OSR does take place at that point we need the value in local variable t1 to be written to the shared auto slot and I don't see how this will happen based on liveness. In other words I do not see what value will be written to the shared slot in this case ?

ncough:
In the example above, the shared auto slot does does not need t1 written to it because it is not used by the interpreter. Under the current implementation, OSR def analysis will ensure t1 gets written to the shared slot, but OSR liveness will result in a dead store for t1 under NextGenHCR and OSRG, so there would be no guarantee that the value written to the slot at transition would be valid.

ncough:
It is assumed that the interpreter and the jit initially sharing the same liveness definition and the interpreter only requires slots to be filled in that would have been live under this definition. This allows for dead stores to be generated in OSR induce blocks and OSR def analysis to be based on the results of OSR live range analysis, reducing the overhead of voluntary OSR.

vijaysun-omr:
Thanks, the argument of other main line stores being eliminated based on the "dead "stores we generate on OSR induction path is a good one since it could leave us without a main line store and coupled with the fact that the "dead" stores on the OSR induction path are deleted anyway the interpreter has to be capable of dealing with a local variable that is not stored.

vijaysun-omr:
Merging this now.

PR #1260, opened 2017-06-14
Version OSR guards out of loops

ncough:
If all yield points have been removed from a loop,
for example, the removal of asyncchecks, it is possible
to version out any OSR guards that have been left behind.
A single OSR guard will be added, which will patch to the
slow version of the loop.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -4314,6 +4314,54 @@ void TR_LoopVersioner::versionNaturalLoop(TR_RegionStructure *whileLoop, List<TR
       dumpOptDetails(comp(), "The node %p has been created for testing if virtual guard is required\n", nextComparisonNode);
       }
 
+   // If all yield points have been removed from the loop but OSR guards remain, 
+   // they can be versioned out. Currently, this code assumes all OSR guards
+   // will be patched by the same runtime assumptions, so only one OSR guard is added
+   // to branch to the slow loop.
+   //
+   bool safeToRemoveOSRGuards = false;
+   bool seenOSRGuards = false;
+   static char *disableLoopOSR = feGetEnv("TR_DisableOSRGuardLoopVersioner");
+   if (comp()->getHCRMode() == TR::osr && disableLoopOSR == NULL)
+      {
+      safeToRemoveOSRGuards = true;
+      ListIterator<TR::Block> blocksIt(&blocksInWhileLoop);
+      TR::Node *osrGuard = NULL;
+      for (TR::Block *block = blocksIt.getCurrent(); safeToRemoveOSRGuards && block; block = blocksIt.getNext())
+         {
+         for (TR::TreeTop *tt = block->getEntry(); tt != block->getExit(); tt = tt->getNextTreeTop())
+            {
+            if (comp()->isPotentialOSRPoint(tt->getNode()))
+               {
+               safeToRemoveOSRGuards = false;
+               break;
+               }
+            else if (tt->getNode()->isOSRGuard())
+               {
+               osrGuard = tt->getNode();
+               seenOSRGuards = true;
+               }
+            }
+         }
+      if (seenOSRGuards && safeToRemoveOSRGuards)
+         {
+         if (performTransformation(comp(), "%sCreate versioned OSRGuard\n", OPT_DETAILS_LOOP_VERSIONER))
+            {
+            TR_ASSERT(osrGuard, "should have found an OSR guard to version");
+
+            // Duplicate the OSR guard tree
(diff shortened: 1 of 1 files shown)


compiler/optimizer/LoopVersioner.cpp
```diff
@@ -4775,6 +4815,22 @@ void TR_LoopVersioner::versionNaturalLoop(TR_RegionStructure *whileLoop, List<TR
...
+      {
+      ListIterator<TR::Block> blocksIt(&blocksInWhileLoop);
+       for (TR::Block *block = blocksIt.getCurrent(); block; block = blocksIt.getNext())
+          {
+          TR::TreeTop *lastRealTT = block->getLastRealTreeTop();
+          if (lastRealTT
+              && lastRealTT->getNode()->isOSRGuard()
+              && performTransformation(comp(), "%sRemove OSR guard n%dn from hot loop\n", OPT_DETAILS_LOOP_VERSIONER, lastRealTT->getNode()->getGlobalIndex()))
+             {
+             _cfg->removeEdge(block->getEdge(lastRealTT->getNode()->getBranchDestination()->getEnclosingBlock()));
```
andrewcraik: We should probably remove the the virtual guard as well using removeVirtualGuard since that is what we are doing here. We might also want to use the Block removeBranch since that would be the more official way to sort this out I think.

ncough: The removal stage has been updated to use removeVirtualGuard and removeBranch.

andrewcraik:
There is a stray file in this commit OMRResolvedMethodSymbol.cpp.orig which should be removed.

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -4314,6 +4314,49 @@ void TR_LoopVersioner::versionNaturalLoop(TR_RegionStructure *whileLoop, List<TR
       dumpOptDetails(comp(), "The node %p has been created for testing if virtual guard is required\n", nextComparisonNode);
       }
 
+   bool safeToRemoveOSRGuards = false;
```
andrewcraik: This could use a comment to describe the logic of when it is safe to remove an OSRGuard and there is an assumption in this code that all OSRGuards are the same which is currently true, but which could potentially have to change in the future if OSRGuards patch more specifically in the future.

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -4775,6 +4815,24 @@ void TR_LoopVersioner::versionNaturalLoop(TR_RegionStructure *whileLoop, List<TR
       properRegion->removeExternalEdgeTo(predBlock->getStructureOf(), succBlock->getStructureOf()->getNumber());
       }
 
+   if (seenOSRGuards && safeToRemoveOSRGuards)
```
andrewcraik: Can probably use a comment as well to describe what is going to happen and why like most of the other big blocks in this area.

ncough:
Comments have been added and the stray file removed.

andrewcraik:
LGTM @vijaysun-omr please review.

vijaysun-omr:
This will work but I think if you did not have anything else to version with respect to, you won't even call versionNaturalLoop. 

The way the versioner works is by collecting up everything it might consider for versioning in an early "analysis phase" and if none of those checks etc. were versionable in the end (not invariant nor based on induction variable) then it won't even call versionNaturalLoop. The alternative to implement this could be to include the test we want in the analysis phase, e.g. you could do the initial part where you check the legality of the transformation by looking for OSR points in detectInvariantTrees. The latter part where you eliminate the guard from the loop could be done in buildConditionalTree.

andrewcraik:
I understand the way the current implementation works by gathering up and then transforming. There are two issues with implementing the OSR versioning this way:
1) the ability to version OSR guards depends on all OSR points being versioned - determining this in the analysis phase is very hard because the unreachable blocks from other versionings are still present in the loop etc
2) an OSR guard will only be in the loop if there was an OSR point originally in the loop - that is we must have to version an asynccheck, guard, or other structure to remove an OSR point to allow us to remove the OSR guard
As a result allowing versioner to version only an OSR guard does not make sense and we chose to implement it as a post pass to clean up guards that are no longer needed due to other versionings.

vijaysun-omr:
Thanks for the clarification. CI build was not successful it seems (though it's probably some issue that warrants a simple retry) ?

andrewcraik:
@vijaysun-omr The CI build failed because Travis CI timed out on one platform - all other platforms passed clean. Is there any way for you to restart the one failed platform to save time?

vijaysun-omr:
Since all checks have passed now and my question from before is answered, I will merge.

PR #1262, opened 2017-06-15
Find the actual virtual guard for OSR guards

ncough:
When searching for virtual guards, the actual guard
should be used, rather than one with a matching
BCI.


andrewcraik:
LGTM - OSR guards are like HCR guards so we need the same special case for this logic to work properly.

vijaysun-omr:
Looks safe to me to mimic what happens for HCR guards in this logic. Merging.

PR #1263, opened 2017-06-15
Remove unnecessary uses of TR::comp()

0xdaryl:
`TR::comp()` is primarily useful in contexts where the Compilation
object isn't already cached somewhere.  While retrieving the
current compilation's Compilation object is always legal using
`TR::comp()` it does incur a performance penalty because the data
is retrieved from thread local storage (TLS).

In places where the Compilation object is already cached, retrieve
it from there rather than TLS.  This change should only be a
compile-time performance win.

This PR is split into commits from each major subdirectory under
`compile`.


vijaysun-omr:
Thanks for this change saving compile time. Merging.

PR #1266, opened 2017-06-16
More Tril Improvements

Leonardo2718:
These commits add various improvements to Tril, including:

* code clean up
* added documentation
* added tests

[ci skip] since Tril has not been added to continuous integration


mstoodle:
Is there a reason you chose not to give parameters or methods names ? Or did I just miss them?

Leonardo2718:
@mstoodle 
Method and argument info is currently only used to generate the method signature needed for compilation. Names aren't currently needed anywhere else in Tril, so I opted to only implement the pieces I needed first.

I will likely add the capability for naming both methods and arguments in a later pull request though.

PR #1268, opened 2017-06-16
CMake: Add vmtest and threadextendedtest

dnakamura:
This adds the mentioned tests and has some minor supporting changes.


charliegracie:
I do not want this to be stalled so I am going to merge it now as I believe it looks ok. If @rwy0717 has any suggestions we can make them in a new PR.

PR #1272, opened 2017-06-19
Do not enable HCR by default

ncough:
NextGenHCR currently forces HCR by default. However,
support for HCR should be requested, defaulting to the
NextGenHCR implementation instead.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -437,14 +437,14 @@ OMR::Compilation::Compilation(
    if (_methodSymbol->isNative())
       self()->setOption(TR_DisableOSR);
 
-   // Do not default OSR and HCR on if:
-   // NextGenHCR is disabled, as they are enabled for it
+   // Do not default OSR on if:
+   // NextGenHCR is disabled, as it is enabled for it
    // OSR is explicitly disabled
    // FSD is enabled, as HCR cannot be enabled with it
-   if (!self()->getOption(TR_DisableNextGenHCR) && !self()->getOption(TR_DisableOSR) && !self()->getOption(TR_FullSpeedDebug))
+   // HCR has not been enabled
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM @vijaysun-omr

compiler/compile/OMRCompilation.cpp
```diff
@@ -441,10 +441,9 @@ OMR::Compilation::Compilation(
    // NextGenHCR is disabled, as they are enabled for it
```
0xdaryl: Does the comment one line above this (L440) need to change to:   "Do not default OSR on if:"  ?

compiler/compile/OMRCompilation.cpp
```diff
@@ -441,10 +441,9 @@ OMR::Compilation::Compilation(
...
    // OSR is explicitly disabled
    // FSD is enabled, as HCR cannot be enabled with it
-   if (!self()->getOption(TR_DisableNextGenHCR) && !self()->getOption(TR_DisableOSR) && !self()->getOption(TR_FullSpeedDebug))
+   if (!self()->getOption(TR_DisableNextGenHCR) && !self()->getOption(TR_DisableOSR) && !self()->getOption(TR_FullSpeedDebug) && self()->getOption(TR_EnableHCR))
       {
       self()->setOption(TR_EnableOSR); // OSR and HCR must be enabled for NextGenHCR
```
0xdaryl: Does this comment need to change as well?

compiler/compile/OMRCompilation.cpp
ncough: Comments have been updated to reflect the change in behaviour.

vijaysun-omr:
I will merge this once checks pass.

andrewcraik:
@vijaysun-omr This has now passed everything except osx - osx is still waiting to start its build. Unlikely that this has any platform specific component.

PR #1275, opened 2017-06-19
WIP: Use std::string in JitBuilder interfaces

sajidahmed21:
Replace C strings with `std::string` in Jit Builder interface.


sajidahmed21:
This is WIP. But would appreciate reviews on the current changeset.

compiler/ilgen/MethodBuilder.hpp
```diff
@@ -73,7 +74,7 @@ class MethodBuilder : public TR::IlBuilder
 
    TR::TypeDictionary *typeDictionary()                      { return _types; }
 
-   const char *getDefiningFile()                             { return _definingFile; }
+   std::string getDefiningFile()                       { return _definingFile; }
```
Leonardo2718: Can you fix the alignment here so that the `{` is aligned with the other ones (as it was before)?

compiler/ilgen/MethodBuilder.hpp
```diff
@@ -102,7 +103,7 @@ class MethodBuilder : public TR::IlBuilder
    void AppendBuilder(TR::BytecodeBuilder *bb);
    void AppendBuilder(TR::IlBuilder *b)    { this->OMR::IlBuilder::AppendBuilder(b); }
 
-   void DefineFile(const char *file)                         { _definingFile = file; }
+   void DefineFile(std::string file)                   { _definingFile = file; }
```
Leonardo2718: Same comment about '{' alignment.

fvtest/compilertest/compile/Method.hpp
```diff
@@ -103,7 +104,7 @@ class ResolvedMethod : public ResolvedMethodBase, public Method
    public:
    ResolvedMethod(TR_OpaqueMethodBlock *method);
    ResolvedMethod(TR::MethodBuilder *methodBuilder);
-   ResolvedMethod(const char      * fileName,
+   ResolvedMethod(std::string       fileName,
```
Leonardo2718: `const std::string&` ?

fvtest/compilertest/compile/Method.hpp
```diff
@@ -194,14 +195,14 @@ namespace TR
...
 
-         ResolvedMethod(char            * fileName,
-                        char            * lineNumber,
-                        char            * name,
-                        int32_t           numArgs,
-                        TR::IlType     ** parmTypes,
-                        TR::IlType      * returnType,
-                        void            * entryPoint,
-                        TR::IlInjector  * ilInjector)
+         ResolvedMethod(std::string      fileName,
```
Leonardo2718: `const std::string&` ?

jitbuilder/compile/Method.hpp
```diff
@@ -194,7 +195,7 @@ namespace TR
             : JitBuilder::ResolvedMethod(method)
             { }
 
-         ResolvedMethod(char            * fileName,
+         ResolvedMethod(std::string       fileName,
```
Leonardo2718: `const std::string&` ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1682,18 +1682,18 @@ IlBuilder::processCallArgs(TR::Compilation *comp, int numArgs, va_list args)
  *    The list is a computed address followed by the actual arguments
  */
 TR::IlValue *
-IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, ...)
+IlBuilder::ComputedCall(std::string functionName, int32_t numArgs, ...)
```
Leonardo2718: `const std::string&` ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1706,46 +1706,46 @@ IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, ...)
  *    the computed address followed by the actual arguments
  */
 TR::IlValue *
-IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, TR::IlValue **argValues)
+IlBuilder::ComputedCall(std::string functionName, int32_t numArgs, TR::IlValue **argValues)
```
Leonardo2718: `const std::string&` ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1706,46 +1706,46 @@ IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, ...)
...
-   TR_ASSERT(resolvedMethod, "Could not identify function %s\n", functionName);
+   TR_ASSERT(resolvedMethod, "Could not identify function %s\n", functionName.c_str());
 
    TR::SymbolReference *methodSymRef = symRefTab()->findOrCreateComputedStaticMethodSymbol(JITTED_METHOD_INDEX, -1, resolvedMethod);
    return genCall(methodSymRef, numArgs, argValues, false /*isDirectCall*/);
    }
 
 TR::IlValue *
-IlBuilder::Call(const char *functionName, int32_t numArgs, ...)
+IlBuilder::Call(std::string functionName, int32_t numArgs, ...)
```
Leonardo2718: `const std::string&` ?

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1706,46 +1706,46 @@ IlBuilder::ComputedCall(const char *functionName, int32_t numArgs, ...)
...
-   TR_ASSERT(resolvedMethod, "Could not identify function %s\n", functionName);
+   TR_ASSERT(resolvedMethod, "Could not identify function %s\n", functionName.c_str());
 
    TR::SymbolReference *methodSymRef = symRefTab()->findOrCreateStaticMethodSymbol(JITTED_METHOD_INDEX, -1, resolvedMethod);
    return genCall(methodSymRef, numArgs, argValues);
    }
 
 TR::IlValue *
-IlBuilder::Call(const char *functionName, int32_t numArgs, TR::IlValue ** argValues)
+IlBuilder::Call(std::string functionName, int32_t numArgs, TR::IlValue ** argValues)
```
Leonardo2718: `const std::string&` ?

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -380,13 +380,12 @@ MethodBuilder::lookupSymbol(const char *name)
    }
 
 TR::ResolvedMethod *
-MethodBuilder::lookupFunction(const char *name)
+MethodBuilder::lookupFunction(std::string name)
```
Leonardo2718: `const std::string&` ?

fvtest/compilertest/compile/Method.hpp
```diff
@@ -106,7 +106,7 @@ class ResolvedMethod : public ResolvedMethodBase, public Method
    ResolvedMethod(TR::MethodBuilder *methodBuilder);
    ResolvedMethod(std::string       fileName,
                   const char      * lineNumber,
-                  char            * name,
+                  std::string       name,
```
Leonardo2718: `const std::string&` ?

fvtest/compilertest/compile/Method.hpp
```diff
@@ -197,7 +197,7 @@ namespace TR
 
          ResolvedMethod(std::string      fileName,
                         char           * lineNumber,
-                        char           * name,
+                        std::string      name,
```
Leonardo2718: `const std::string&` ?

jitbuilder/compile/Method.hpp
```diff
@@ -106,7 +106,7 @@ class ResolvedMethod : public ResolvedMethodBase, public Method
    ResolvedMethod(TR::MethodBuilder *methodBuilder);
    ResolvedMethod(std::string       fileName,
                   const char      * lineNumber,
-                  char            * name,
+                  std::string       name,
```
Leonardo2718: `const std::string&` ?

jitbuilder/compile/Method.hpp
```diff
@@ -197,7 +197,7 @@ namespace TR
 
          ResolvedMethod(std::string       fileName,
                         char            * lineNumber,
-                        char            * name,
+                        std::string       name,
```
Leonardo2718: `const std::string&` ?

jitbuilder/release/src/Switch.cpp
```diff
@@ -58,13 +58,13 @@ SwitchMethod::SwitchMethod(TR::TypeDictionary *d)
  * "printString" is always going to be called anyway.
  */
 bool
-SwitchMethod::RequestFunction(const char *name)
+SwitchMethod::RequestFunction(std::string name)
```
Leonardo2718: `const std::string&` ?

sajidahmed21:
Agree. Using string references for function params is certainly more efficient.
Have made all suggested changes.

charliegracie:
I am not that familiar with std::string.  Can you easily create one from a char *?

sajidahmed21:
Yes, you can use the `std::string(const char* s)` constructor to create one as follows:

`char *s = "Example String";`
`std::string std_str(s);`

However, if you just want to directly create a `std::string`, you could do:

`std::string s = "Example String";`

compiler/ilgen/IlBuilder.hpp
```diff
@@ -27,7 +27,9 @@
 
 #include <fstream>
 #include <stdarg.h>
-#include <string.h>
+#include <string.h>   // TODO: Remove
```
0xdaryl: I don't like TODOs.  What's preventing this from being removed now?

compiler/ilgen/MethodBuilder.hpp
```diff
@@ -119,28 +121,28 @@ class MethodBuilder : public TR::IlBuilder
...
+                       TR::IlType        ** parmTypes);
 
    /**
     * @brief will be called if a Call is issued to a function that has not yet been defined, provides a
     *        mechanism for MethodBuilder subclasses to provide method lookup on demand rather than all up
     *        front via the constructor.
     * @returns true if the function was found and DefineFunction has been called for it, otherwise false
     */
-   virtual bool RequestFunction(const char *name) { return false; }
+   virtual bool RequestFunction(std::string name) { return false; }
```
0xdaryl: Pass string by reference.

fvtest/compilertest/compile/Method.cpp
```diff
@@ -67,8 +67,10 @@ ResolvedMethod::signature(TR_Memory * trMemory, TR_AllocationKind allocKind)
    {
    if( !_signature )
       {
-      char * s = (char *)trMemory->allocateMemory(strlen(_fileName) + 1 + strlen(_lineNumber) + 1 + strlen(_name) + 1, allocKind);
-      sprintf(s, "%s:%s:%s", _fileName, _lineNumber, _name);
+      char * s = (char *)trMemory->allocateMemory(_fileName.length() + 1 + strlen(_lineNumber) + 1 + _name.length() + 1, allocKind);
+
+      // TODO: Change this to use C++ string concatenation
+      sprintf(s, "%s:%s:%s", _fileName.c_str(), _lineNumber, _name.c_str());
```
0xdaryl: I don't like TODOs.  Presumably, doing this in std:strings and then converting the result to a c_str would eliminate the explicit allocation above?  It would look cleaner I suppose, but let's reason about its efficiency.

jitbuilder/compile/Method.cpp
```diff
@@ -63,8 +63,10 @@ JitBuilder::ResolvedMethod::signature(TR_Memory * trMemory, TR_AllocationKind al
    {
    if( !_signature )
       {
-      char * s = (char *)trMemory->allocateMemory(strlen(_fileName) + 1 + strlen(_lineNumber) + 1 + strlen(_name) + 1, allocKind);
-      sprintf(s, "%s:%s:%s", _fileName, _lineNumber, _name);
+      // TODO: Create using C++ string concatenation
```
0xdaryl: Same comment as above for this TODO.

compiler/ilgen/ThunkBuilder.cpp
```diff
@@ -47,8 +47,8 @@ ThunkBuilder::ThunkBuilder(TR::TypeDictionary *types, const char *name, TR::IlTy
    _numCalleeParams(numCalleeParams),
    _calleeParamTypes(calleeParamTypes)
    {
-   DefineLine(__FILE__);
```
mgaudet: Good catch :D

compiler/ilgen/IlBuilder.hpp
sajidahmed21: I was planning to remove it when I got rid of all `char *` in this file. However, I don't see any `string.h` functions being used anyway - so I'll remove it as part of this commit.

fvtest/compilertest/compile/Method.cpp
sajidahmed21: I am unsure if we can get rid of the explicit allocation, given that the lifetime of the (signature) string has to be longer than the lifetime of this function call (and therefore, the caller of this function is providing the alloc info). However, if it is known that the string will not be used / accessed beyond the lifetime of this `ResolvedMethod` instance, then we could just store it as a member variable, and return the corresponding c_str.

As far as the TODO is concerned, I can't think of a reason as to why we should use C++ string concat. Doing it using `sprintf` will likely be more efficient. I will remove it (the TODO).

jitbuilder/compile/Method.cpp
sajidahmed21: As mentioned above, need to discuss this.

mstoodle:
@sajidahmed21 Does this pull request include commits from #1288 ?

compiler/ilgen/MethodBuilder.hpp
```diff
@@ -73,7 +74,7 @@ class MethodBuilder : public TR::IlBuilder
 
    TR::TypeDictionary *typeDictionary()                      { return _types; }
 
-   const char *getDefiningFile()                             { return _definingFile; }
+   std::string getDefiningFile()                             { return _definingFile; }
```
mstoodle: should this return a reference or did you intend to make a copy? If you intended it, where will the copy be allocated?

mstoodle: Maybe I missed it, but I don't see _definingFile being changed to a std::string & in this commit?

jitbuilder/compile/Method.hpp
```diff
@@ -166,7 +167,7 @@ class ResolvedMethod : public ResolvedMethodBase, public Method
                                TR::SymbolReferenceTable *symRefTab);
 
    protected:
-   const char *_fileName;
+   std::string _fileName;
```
mstoodle: where does the backing character array get allocated for this string?

compiler/ilgen/MethodBuilder.hpp
sajidahmed21: Making it return by reference would certainly be more efficient. I intended to make it return by value to avoid the possibilities of accidental modification and dangling reference. However, if you prefer, I could make it return a const reference, which would help us _get around_ the first problem.

As far as allocation is concerned, the copy would get allocated by the caller.

sajidahmed21: I had (mistakenly) included the type change for `_definingFile` along with the next commit. I have rebased my commits to include it as a part of this now.

sajidahmed21:
@mstoodle Yes, I have rebased my commits to include the ones in #1288 as some of the std::string changes are dependent on the hash table ones.

jitbuilder/compile/Method.hpp
sajidahmed21: I am pretty confident, the underlying character array is on the heap (or global data segment?). It's the responsibility of the `std::string` class to manage that, and therefore, it will get deallocated automatically once the corresponding `std::string` is no longer being used.

Just curious, any particular reason why you are interested about this?

mstoodle: Because the JIT is not supposed to allocate memory on the malloc heap. Compiler classes do not have destructors to manage memory the way you're imagining. You need to allocate these objects using an Allocator so that any supplementary memory allocated comes from the JIT allocator rather than, say, ::new() . 

Take a look at the definition and the constructor's initialization of, say, the field `_dependentSymRefs` in `compiler/optimizer/RematTools.hpp`

compiler/ilgen/MethodBuilder.hpp
mstoodle: Allocation of memory is a subject requiring care in the compiler component, please see answer below.

jitbuilder/compile/Method.hpp
mgaudet: So, in the OMR core, I agree that we need to be careful with not using the system allocator. However, in JitBuilder, we have to be careful about where we draw the line. 

I have experience the [ `PersistentMemory` destructor assertion](https://github.com/eclipse/omr/blob/7b53b24271a7e10407d7bd4ccedf69fe1964dcc3/compiler/env/TRMemory.hpp#L684) prevent me from wrapping JitBuilder with a library, because in that wrapper I ceded memory management of the JitBuilder objects to the library -- so when it went to destroy the objects, i would fail.

mstoodle: This change I commented on isn't a JitBuilder "object"; it is not visible to clients of the library. 

Method is a core OMR class that has been extended by the JitBuilder layer. As such, all the memory allocated by its OMR base class(es) will be properly tracked, but if JitBuilder layer is "free" to neglect to use the same Allocator infrastructure, then anything done at the JitBuilder layer will not be similarly tracked.

This path leads to JitBuilder providing misleading answer to the memory tracking features of the OMR compiler / OMR project, because it will only provide part of the answer (with some other part invisibly sitting on the global heap).

Unless I've misunderstood something here, I don't agree with this particular direction.

mstoodle: Also note that IlBuilder objects other than MethodBuilders and ThunkBuilders should really only be created during compilation (i.e. in the buildIL call of some MethodBuilder). As such, they are actually running as part of the compiler, so I think we should really be trying hard to allocate that memory according to the compiler and not ceding control of those objects to the client.

Granted, MethodBuilder should really be cleaned up on the way out of compileMethodBuilder to eliminate compile-time artifacts like IlBuilder objects (just like I did for SymbolReferences in the TypeDictionary).

mstoodle: We could change some of that story by creating a JitBuilder bytecode set or persistent structure (which you've proposed before for other reasons) and then having the JitBuilder il generator become a visitor of that structure that generates OMR compiler IL. That's more analogous to how LLVM IR builders work.

I think we should enable this path (by implementing a particular kind of "standard" MethodBuilder that parses said structure to generate IL), but I don't (yet anyway) think we should change JitBuilder to operate in only that way.

mgaudet: I missed that this was on Method.hpp -- so I agree, it either needs to not use `std::string` or it needs to use an allocator to do its allocation. 

@sajidahmed21 asked if a MethodBuilder can be expected to outlive `ResolvedMethod` ... I think the answer is yes here.

mstoodle:
All right, that means #1288 must be merged before this pull request.

mstoodle:
Hi @sajidahmed21 if you rebase this one now that #1288 has been merged, I'll go over it. Please ping me when it's been rebased.

sajidahmed21:
Sure, will do. It will likely take a few days since some of these changes need to be redone to incorporate the use of custom regions. I'll ping you once its done.

mstoodle:
Any update here, @sajidahmed21 ?

mstoodle:
@sajidahmed21 will you be updating this PR to address conflicts and take your other changes into effect? If not, I'll probably close this PR .

mstoodle:
Ok, closing for now.

This change represents a fundamental improvement to our API and a solid step towards a v1 for the JitBuilder release, so I still keenly support this change happening.

PR #1280, opened 2017-06-20
Tell VP that tstart has no fallthrough

jdmpapin:
Without this, global VP (value propagation) would look for a fallthrough edge from a block ending with `tstart` to the textually following block, which may not be a branch target of the `tstart` node. Failure to find the edge results in dereferencing a past-the-end iterator.


andrewcraik:
LGTM

vijaysun-omr:
I am going to merge this now since we need it urgently for a downstream project.

andrewcraik:
Travis CI has passed everything except OSX at this point.

PR #1283, opened 2017-06-21
WIP: Insert a GC map for nodes that can cause GC [X86]

0dvictor:
Certain load/store nodes can cause GC; hence a GC map is
required.


mgaudet:
Surely that pattern can be encapsulated into a function, rather than copy-pasted 8 times?

compiler/x/codegen/FPTreeEvaluator.cpp
```diff
@@ -314,6 +314,12 @@ TR::Register *OMR::X86::TreeEvaluator::performFload(TR::Node *node, TR::MemoryRe
       targetRegister = cg->allocateSinglePrecisionRegister(TR_X87);
       instr = generateFPRegMemInstruction(FLDRegMem, node, targetRegister, sourceMR, cg);
       }
+
+   if (node && node->canCauseGC())
```
0xdaryl: General comment thoughout your PR:  while I understand (and appreciate!) your thoroughness, I don't think `node` non-NULLness needs to be checked at each stage because in many cases it is already assumed to be non-NULL and used as if it were (e.g., see L323 just after this change).

PR #1284, opened 2017-06-21
Allow preemption of optimization initialization

lmaisons:
Allow language runtimes consuming the OMR compiler to preempt the
default initialization of OptimizationManager objects, providing instead
a complete alternative that initializes OptimizationManagers only for
the optimizations and optimization groups needed by the consuming
project.

### DIFF
--- a/compiler/optimizer/OMROptimizer.cpp
+++ b/compiler/optimizer/OMROptimizer.cpp
@@ -732,6 +732,13 @@ OMR::Optimizer::Optimizer(TR::Compilation *comp, TR::ResolvedMethodSymbol *metho
    // zero opts table
    memset(_opts, 0, sizeof(_opts));
 
+/*
+ * Allow downstream projects to disable the default initializaiton of optimizations
+ * and allow them to take full control over this process.  This can be an advantage
+ * if they don't use all of the optimizations initialized here as they can avoid
+ * getting linked in to the binary in their entirety.
+ */
+#if !defined(TR_OVERRIDE_OPTIMIZATION_INITIALIZATION)
    // initialize OMR optimizations
 
    _opts[OMR::andSimplification] =
@@ -915,6 +922,7 @@ OMR::Optimizer::Optimizer(TR::Compilation *comp, TR::ResolvedMethodSymbol *metho
       new (comp->allocator()) TR::OptimizationManager(self(), NULL, OMR::finalGlobalGroup, finalGlobalOpts);
 
    // NOTE: Please add new OMR optimization groups here!
+#endif
 
 }
 
@@ -2815,7 +2823,8 @@ void OMR::Optimizer::doStructureChecks()
 
 bool OMR::Optimizer::getLastRun(OMR::Optimizations opt)
    {
-   TR_ASSERT(_opts[opt], "Optimization manager for %d should be initialized first", opt);
+   if (!_opts[opt])
+      return false;
    return _opts[opt]->getLastRun();
    }
 
--- a/jitbuilder/optimizer/JBOptimizer.cpp
+++ b/jitbuilder/optimizer/JBOptimizer.cpp
@@ -64,6 +64,9 @@
 #include "optimizer/SinkStores.hpp"
 #include "optimizer/ShrinkWrapping.hpp"
 #include "optimizer/TrivialDeadBlockRemover.hpp"
+#include "optimizer/GlobalValuePropagation.hpp"
+#include "optimizer/LocalValuePropagation.hpp"
+#include "optimizer/Inliner.hpp"
 
 
 static const OptimizationStrategy tacticalGlobalRegisterAllocatorOpts[] =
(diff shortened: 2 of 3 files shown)


lmaisons:
This change enables us to save ~1MB of binary size for JitBuilder projects:

-rwxr-xr-x 1 lmaisons users 8881088 Jun 21 19:53 luav.baseline
-rwxr-xr-x 1 lmaisons users 7966592 Jun 21 19:39 luav.trim

compiler/optimizer/OMROptimizer.cpp
```diff
@@ -723,6 +723,7 @@ OMR::Optimizer::Optimizer(TR::Compilation *comp, TR::ResolvedMethodSymbol *metho
    // zero opts table
    memset(_opts, 0, sizeof(_opts));
 
+#if !defined(TR_OVERRIDE_OPTIMIZATION_INITIALIZATION)
```
mgaudet: Probably should document the expected usage of this somewhere?

vijaysun-omr:
@andrewcraik please review

andrewcraik:
Since our opt strategies are all in static arrays wouldn't it make more sense to collect these arrays together and use the opt strategies themselves to do the init so that adding an opt to a strategy would cause it to be initialized appropriately? It seems strange to provide an init and then ifdef it out when you could init by walking the strategy and/or group arrays? What would be the advantage of this proposal vs the init based on usage approach?

lmaisons:
The goal here is to avoid a link-time dependency on native symbols for functionality that will go unused.  For opt implementations, it turns out that *the* thing that causes them to show up in the link dependency graph is their initialization.  In order to achieve this with opt strategies, the opt strategies would have to stop being enum values and instead refer directly to the opts themselves.

While I think it's likely possible, I expect it would take quite a bit of heavy lifting to accomplish, and would still require some form of compile-time machinery to emit or exclude code on a per-project basis.

vijaysun-omr:
@andrewcraik please respond to Luc's clarification when you get a chance.

compiler/optimizer/OMROptimizer.cpp
```diff
@@ -2787,7 +2789,8 @@ void OMR::Optimizer::doStructureChecks()
 
 bool OMR::Optimizer::getLastRun(OMR::Optimizations opt)
    {
-   TR_ASSERT(_opts[opt], "Optimization manager for %d should be initialized first", opt);
+   if (!_opts[opt])
```
mstoodle: I'm not sure why this change was needed? Does the optimizer ever ask about last run for an opt that isn't in a strategy list (and that presumably would be initialized if it were in a strategy list) ?

mstoodle:
Another ping to review, please, @andrewcraik . I just merged another PR that will force Luc to add another optimization into this PR before it can be merged...

compiler/optimizer/OMROptimizer.cpp
lmaisons: I was getting bit by this conditional in VP:
```
if (!comp()->getFlowGraph()->getStructure() &&
          (((comp()->mayHaveLoops() && optimizer()->getLastRun(OMR::loopVersioner)) || !comp()->mayHaveLoops()) &&
           optimizer()->getLastRun(OMR::basicBlockExtension)))
```

mstoodle: ah, ok, that makes sense. I foolishly didn't think about opts asking about other opts :( .

compiler/optimizer/OMROptimizer.cpp
andrewcraik: This needs to be addressed IMO - make it clear what this is and why one would use it.

jitbuilder/optimizer/JBOptimizer.cpp
```diff
@@ -166,7 +168,36 @@ Optimizer::Optimizer(TR::Compilation *comp, TR::ResolvedMethodSymbol *methodSymb
       new (comp->allocator()) TR::OptimizationManager(self(), TR_TrivialDeadBlockRemover::create, OMR::trivialDeadBlockRemover, "O^O TRIVIAL DEAD BLOCK REMOVAL ");
    _opts[OMR::regDepCopyRemoval] =
       new (comp->allocator()) TR::OptimizationManager(self(), TR::RegDepCopyRemoval::create, OMR::regDepCopyRemoval, "O^O REGISTER DEPENDENCY COPY REMOVAL: ");
+
    // NOTE: Please add new IBM optimizations here!
```
andrewcraik: Looks like the is comment needs freshening up.

andrewcraik:
The one thing I think that needs to happen before we merge this is to move the optimization eye catcher strings (eg O^O OTPNAME HERE) into the optimizations as a field of some kind and to pass that field rather than a hardcoded string. Since initialization can now, potentially, occur in OMR or in downstream projects duplicating those strings seems wasteful and could cause un-necessary divergence in logs and debug output over time.

lmaisons:
Latest push just adds OMR::loopCanonicalization initialization.  Still hasn't addressed other comments.

vijaysun-omr:
Any update on this ? Last comment suggests there is more work to do to address some of the earlier comments.

vijaysun-omr:
Another ping about this PR (since it's been open for a while) to make sure it is planned to be progressed based on the feedback so far (if not, we should probably close it).

lmaisons:
Updated to latest master

mstoodle:
@andrewcraik in the interest of merging this PR, would you be willing to accept a follow-on item to change the name initialization so that the optimization defines the name rather than the project that initializes it?

Or are you adamant that change should be included as part of this PR?

lmaisons:
I thought I fixed that concern by moving the opt description strings into virtual functions :shrug:

mstoodle:
Indeed, I don't see the strings anymore, though it isn't clear from the commentary in this PR when that happened.

@andrewcraik are you ok to merge this one now?

andrewcraik:
I think I am good with this now - sorry missed the @ in my inbox.

mstoodle:
genie-omr build

mstoodle:
Ok, i'm not sure why only ppc discovered the problem, but it looks like a recent PR that added trivial inliner to the JitBuilder opt strategy warrants a small update in this PR.

@lmaisons any chance you could rebase (again) and make that (hopefully quick) update? Sorry for the extra work we caused you with our slow merging "process" here :( .

0xdaryl:
The PR in question is #1780.  @xliang6 FYI only.

lmaisons:
I'm happy to do this, but I think it probably needs to be handled slightly strangely to keep the lawyers happy.  I'm fairly sure that if I simply amend the existing commit it's going to trip up the CLA check.

From a purely copyright assertion point of view, the thing that strikes me the most likely as being 'correct' is to add another change set signed off by my Rubikloud email.

Alternatively, it might be easier if someone else with less CLA entanglement push this across the line.

FWIW, https://github.com/lmaisons/omr/commit/fd3265a8f2f69ad5ff41643c9c50761d886ed67c contains a merged and rebased commit with the additional initialization.

mstoodle:
@lmaisons If it's ok with you, I'd like to try having you do it (just to see what will happen; I would like to believe the IP check is smart enough to handle multiple ECA contributors in the same pull request). Please check if you should be adding (not replacing :) ) a Rubikloud copyright statement to the file(s) you change with your current account.

We can always retreat to your suggested approach later if things don't go "swimmingly".

mstoodle:
@lmaisons just a ping to see if this PR is still something you're willing to run to completion?

lmaisons:
Yep, in the process of getting dispensation to do this.

mstoodle:
@lmaisons what's the outlook here, do you think?

lmaisons:
Sorry for the delay, but the good news is that I got permission from Rubikloud to submit this as my own copyright.  I will be rebasing and updating the branch shortly.

mstoodle:
Hey @lmaisons, just wondering if you've got any notable updates here?

mstoodle:
Hi @lmaisons. So I verified with #2305 that your pull request works after rebasing (that zOS failure shown here isn't really a problem with your change, as I'm sure you already realized).

All that's needed here is a rebase and then filling out the ECA with your updated infomation so that the IP validation can pass.

I still think you should submit this yourself, but since you worked for IBM when the original submission was made, I will re-submit the change myself (since I also work for IBM) if you haven't done so by the end of this week just so we can move this work forward (and reduce JitBuilder based JITs by about a megabyte, which means by about 15% for the code samples -- pretty impressive improvement we've been missing out on).

mstoodle:
Ok, I'll try to get this re-submitted this week under my own commit.

Thanks for the original attempt, @lmaisons.

PR #1285, opened 2017-06-21
Check return temp exists splitting in inlining

ncough:
Under OSR, its necessary to ensure the last block
of the callee and the caller's block immediately
after must be separated for distinct OSR exception
edges. However, this splitting was not taking into
account whether the return value was commoned across.
This was working in the past because the callee
always had an OSR catch block, resulting in
the inliner generating a temporary for the result.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -5062,13 +5062,18 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
             _disableTailRecursion = true;
          }
       }
-   else if (comp()->getOption(TR_EnableOSR) && tif->crossedBasicBlock() && !comp()->osrInfrastructureRemoved())
+   else if (comp()->getOption(TR_EnableOSR) && tif->crossedBasicBlock() && !comp()->osrInfrastructureRemoved()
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
There was a rationale for splitting the block that you give both in the description of this PR and in the comment in the code next to where the change was made. If splitting the block is required for preserving correct OSR exception edge structure, it is unclear to me why we can legally choose to not split the block in cases based on something quite unrelated (the return value being commoned) ?

ncough:
This only excludes situations where the inliner believes the callee is simple enough to be directly inlined, allowing it to common the return value across the callee's final block and the next block in the caller. In such a situation, the callee must have OSR disabled, as the existence of any catch block (such as the OSR catch block) prevents it from being inlined directly and would result in a temporary being generated for the return value. With OSR disabled, a distinction between exception edges to catch blocks is not needed.

The check is performed against the return node's reference count, as it will be zero in the event that a temporary was generated or null if there is no return. This check, and the check for `tif->crossedBasicBlock()`, are short cuts to determine if the callee has OSR infrastructure.

vijaysun-omr:
Okay that's quite the non obvious explanation but it's what I was missing. Thanks for clarifying that this essentially means that OSR is disabled on the callee.

PR #1287, opened 2017-06-22
typed_allocator should take references to Regions

lmaisons:
C++ standard library containers will frequently create ephemeral
instances of type coercions of the the containers allocator type,
expecting any allocator instantiation within the same container instance
to be equivalent for allocations and deallocations, with one instance
being able to deallocate memory provided by another instance. Implied by
this requirement is that memory allocations must be able to outlive the
allocator instances from which they were allocated.

Given the above, when using TR::Regions and TR::typed_allocator as the
allocator for C++ containers, one needs to ensure that the instances of
typed_allocator use references to an externally defined Region. A Region
whose lifetime is under explicit control of its user.

This change corrects region use in LocalCSE's uses of std::multimap to
enforce these requirements.


lmaisons:
@andrewcraik FYI

andrewcraik:
LGTM

PR #1288, opened 2017-06-22
WIP: Replace custom hash tables in JitBuilder with std::map

sajidahmed21:
Replace custom hash tables in JitBuilder with `std::map`. This
will ensure better type safety, and later, help in simplifying the
JitBuilder interfaces to use std::string.


mgaudet:
Looks like a good start! I would finish the rest of the hash tables before we merge this I think, but keep going one table per commit.  🎉

sajidahmed21:
As per the uses in the following methods, `_symbolTypes` table in MethodBuilder seems to be mapping symbol names to their corresponding `TR::IlType`.

[MethodBuilder::DefineLocal](https://github.com/eclipse/omr/blob/9cde53884e2701e54b02e7981590d7a06d7a823f/compiler/ilgen/MethodBuilder.cpp#L436)
[MethodBuilder::DefineMemory](https://github.com/eclipse/omr/blob/9cde53884e2701e54b02e7981590d7a06d7a823f/compiler/ilgen/MethodBuilder.cpp#L444)
[MethodBuilder::lookupSymbol](https://github.com/eclipse/omr/blob/9cde53884e2701e54b02e7981590d7a06d7a823f/compiler/ilgen/MethodBuilder.cpp#L358)

However, in [MethodBuilder::defineSymbol](https://github.com/eclipse/omr/blob/9cde53884e2701e54b02e7981590d7a06d7a823f/compiler/ilgen/MethodBuilder.cpp#L341), it seems to be (incorrectly) inserting the corresponding `TR::DataType` by casting it to a `void *`. 

As a fix, I would think that a pointer to the corresponding `TR::IlType` should be passed in as a parameter of `MethodBuilder::defineSymbol` (so that we could store it in the `_symbolTypes` table). However, it would be worthwhile to discuss if there are any ways to derive `TR::IlType` from `TR::DataType` or `TR::SymbolReference`, so that we could get around this without having to change the function signature.

Leonardo2718:
Wow that's nasty! This is exactly the kind of bug that moving to `std::map` is going to catch.

For your fix, I believe `_symbolTypes->add(name, id3, (void*)typeDictionary()->PrimitiveType(symRef->getSymbol()->getDataType()->getDataType()))` should do the trick.

sajidahmed21:
Yes, this is what I was considering to do at first. But couldn't this symbol actually be of a non-primitive type (i.e. `StructType`, `UnionType`, etc)? In which case, wouldn't it be incorrect to represent it using a `PrimitiveType`?

Leonardo2718:
> But couldn't this symbol actually be of a non-primitive type (i.e. StructType, UnionType, etc)?

The type of the symbol comes from `symref` which, being a `TR::SymbolReference`, can only a primitive type.

sajidahmed21:
@mgaudet Please review

sajidahmed21:
@mgaudet I've moved the map initializations to the constructor initializer list, and have updated the description above to reason about STLUtils.

Please review.

mgaudet:
I don't quite understand why 7038d093521f82280af1b067009f1c0289a0597b switches to a `FATAL_ASSERT`? @Leonardo2718 ?

mgaudet:
Regarding 93c123eea7b6ce06f270d306d03e1c45c9ef3264: Does no one ever read that map?

mgaudet:
@Leonardo2718 Is there a way to tweak the API of `LookupUnion` to avoid the wierd casts in https://github.com/eclipse/omr/pull/1288/commits/2a871d6da3f0a53687930f0d7592f4f7c1424792?

mgaudet:
I am pretty comfortable with this PR.

sajidahmed21:
Regarding 7038d093521f82280af1b067009f1c0289a0597b: As per my understanding, a user is not expected to try and lookup undefined functions. So, if `LookupFunction` happens to be called with a non-existing function name, it is almost certainly an error, which should be brought to attention right away. Returning NULL is also an option, but a fatal assert will make it easier for the user to catch such bugs.

Leonardo2718:
Regarding https://github.com/eclipse/omr/commit/7038d093521f82280af1b067009f1c0289a0597b, I think returning `NULL` is the correct thing to do because it is required for the `RequestFunction()` mechanism to work. `lookupFunction()` is part of the internal IlBuilder API and should never be called directly by client code.

Leonardo2718:
Regarding https://github.com/eclipse/omr/commit/2a871d6da3f0a53687930f0d7592f4f7c1424792 : `LookupUnion()` and `LookupStruct()` are part of the client facing API, so I think their interface should be unchanged. To avoid the casts, I would create internal (protected?) versions of these functions that return `UnionType*` and `StructType*`, respectively. `LookupUnion()` and `LookupStruct()` could then call the internal functions and the returned value will be implicitly cast to `TR::IlType*`.

sajidahmed21:
Alright, I'll drop that commit (7038d093521f82280af1b067009f1c0289a0597b) then.

Also, as Matthew mentioned, we are indeed only ever inserting into the memory locations table, and not reading from it. Would like to know if we need it for the future, or is it safe to get rid of it, along with the `DefineMemory` API?

mgaudet:
I think another PR to remove the `DefineMemory` API is warranted. It appears to be a path no longer pursued... though @mstoodle would know its history better.

sajidahmed21:
I've dropped the fatal assert commit, and have created helpers for `LookupStruct` and `LookupUnion` in order to avoid the explicit casts.

@mgaudet Please review.

mstoodle:
In case anyone is interested, `_symbolTypes` is not used after a `TR::SymbolReference` is created (except in one case where the symbols are known to be parameters and so were created before you could create `TR::SymbolReferences`). `MethodBuilder` needs to remember the types of parameters before you have a compilation and so before you can create `TR::SymbolReference` objects. Any time you have a `TR::SymbolReference` (i.e. from `defineSymbol()` ) you add that to `_symbols` and its presence in _symbols means you won't ever access `_symbolTypes` and refer to the returned data as a `TR::DataType`. Really, `defineSymbol()` shouldn't really be writing to `_symbolTypes` at all because it's never used.

I agree it's kludgy, inconsistent, and was simply a hack to get going so it can be fixed.

@magaudet `DefineMemory()` was an early (incomplete) attempt to provide access to global variables. It will still be needed at some point, but the interface will need to be updated (plus we'll want to give things names so we can create appropriate relocations for them). Some of the "machinery" that's there will need to be updated, but it could also be easily recreated.

sajidahmed21:
Thanks for that explanation @mstoodle 

@0xdaryl This change is ready for final review and merging.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -95,7 +96,14 @@ MethodBuilder::MethodBuilder(TR::TypeDictionary *types, OMR::VirtualMachineState
    _allBytecodeBuilders(0),
    _vmState(vmState),
    _bytecodeWorklist(NULL),
-   _bytecodeHasBeenInWorklist(NULL)
+   _bytecodeHasBeenInWorklist(NULL),
+   _parameterSlot(str_comparator),
```
mstoodle: This list looks out of order to me; these fields should be listed in order of their definition in the class. Putting them here in a different order is misleading since the fields won't be initialized in this order (not that there are dependencies, but it's cleaner if the order matches the order the fields will be initialized).

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -139,7 +139,7 @@ MethodBuilder::asMethodBuilder()
 void
 MethodBuilder::initMaps()
    {
-   _parameterSlot = new (PERSISTENT_NEW) TR_HashTabString(typeDictionary()->trMemory());
+   _parameterSlot = std::map<const char *, int32_t>();
```
mstoodle: I notice you changed these to all be inline objects in your last commit, so I won't complain about where these all got allocated, other than to (continue to?) hammer on the point that we have to be super careful where these objects get allocated and how their memory is being managed.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -358,11 +358,11 @@ MethodBuilder::lookupSymbol(const char *name)
...
    TR::DataType type = ((TR::IlType *)(_symbolTypes->getData(typesID)))->getPrimitiveType();
 
-   TR_HashId slotID;
-   if (_parameterSlot->locate(name, slotID))
+   std::map<const char *, int32_t>::iterator it = _parameterSlot.find(name);
+   if (it != _parameterSlot.end())
       {
       symRef = symRefTab()->findOrCreateAutoSymbol(_methodSymbol,
-                                                   (int32_t)(uintptr_t)_parameterSlot->getData(slotID),
+                                                   it->second,
```
mstoodle: Although I set a bad example here by directly using "getData()"... I confess, I am not a fan of using `it->second` directly. I would prefer you put it in a meaningful variable name to make it more clear (to me, I suppose) and rely on the compiler to optimize the renaming away.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -328,7 +328,7 @@ MethodBuilder::symbolDefined(const char *name)
    // be called in a MethodBuilder contructor. In contrast, ::DefineSymbol
    // which inserts into _symbols, can only be called from within a MethodBuilder's
    // ::buildIL() method ).
-   return (_symbolTypes.find(name) != _symbolTypes.end());
+   return _symbolTypes.find(name) != _symbolTypes.end();
```
mstoodle: This change belongs in a different commit.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -677,10 +649,11 @@ TypeDictionary::FieldReference(const char *typeName, const char *fieldName)
    {
    TR_HashId typeID = 0;
 
-   auto found = _structsByName->locate(typeName, typeID);
+   StructIterator it = _structsByName.find(typeName);
+   bool found = it != _structsByName.end();
    if (found)
```
mstoodle: You should just simplify here and eliminate the use of the `found` bool since comparing the iterator to the end "value" is idiomatic.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -699,10 +672,9 @@ void
 TypeDictionary::NotifyCompilationDone()
    {
    // clear all symbol references for fields
-   TR_HashTabIterator structIterator(_structsByName);
-   for (StructType *aStruct = (StructType *)structIterator.getFirst();aStruct;aStruct = (StructType *)structIterator.getNext())
+   for (StructIterator it = _structsByName.begin(); it != _structsByName.end(); it++)
       {
-      aStruct->clearSymRefs();
+      it->second->clearSymRefs();
```
mstoodle: same comment about using "it->second" directly.

compiler/ilgen/TypeDictionary.hpp
```diff
@@ -36,9 +36,9 @@
 #include "ilgen/IlBuilder.hpp"
 
 class TR_Memory;
-class TR_HashTabString;
 
 namespace OMR { class StructType; }
+namespace OMR { class UnionType; }
```
mstoodle: I'm not super happy about these two type names showing up in the public header (I was trying to keep these types completely private), but I'll trade my happiness for the simplifying value of getStruct() and getUnion().

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -137,18 +142,6 @@ MethodBuilder::asMethodBuilder()
    }
 
 void
-MethodBuilder::initMaps()
```
mstoodle: very happy to see this function go away; i always hated it!

mstoodle:
Just a note for awareness: #1275 is dependent on this PR being merged first.

compiler/ilgen/TypeDictionary.hpp
sajidahmed21: Yeah, I wasn't too happy either when I added them. Technically, we _could_ still avoid this declaration by using Pimpl pattern. But I doubt if it would be worthwhile doing it considering the tradeoffs.

sajidahmed21:
Thanks @mstoodle !
I've made the suggested changes. Please review.

compiler/ilgen/MethodBuilder.cpp
sajidahmed21: I only refactored these initalizations to the constructor initializer list. Does this affect the memory allocation or lifetime of these objects in any way? Also, can you please explain what you mean by "inline" objects here?

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -675,42 +632,59 @@ TypeDictionary::PointerTo(TR::IlType *baseType)
...
+   UnionIterator unionIterator = _unionsByName.find(typeName);
+   if (unionIterator != _unionsByName.end())
       {
-      UnionType *theUnion = (UnionType *) _unionsByName->getData(typeID);
+      UnionType *theUnion = unionIterator->second;
       return theUnion->getFieldSymRef(fieldName);
       }
 
-   TR_ASSERT(false, "No type with name `%s`", typeName);
+   TR_ASSERT_FATAL(false, "No type with name `%s`", typeName);
```
0xdaryl: Change quotes to ' ' for consistency with other assert messages.

compiler/ilgen/TypeDictionary.hpp
```diff
@@ -415,46 +416,55 @@ class TypeDictionary
...
+   TR::IlType                                             * pWord;
+   TR::IlType                                             * pFloat;
+   TR::IlType                                             * pDouble;
+   TR::IlType                                             * pAddress;
+   TR::IlType                                             * pVectorInt8;
+   TR::IlType                                             * pVectorInt16;
+   TR::IlType                                             * pVectorInt32;
+   TR::IlType                                             * pVectorInt64;
+   TR::IlType                                             * pVectorFloat;
+   TR::IlType                                             * pVectorDouble;
```
0xdaryl: The abundance of formatting spaces introduced to line all these fields up with `_structsByName` and `_unionsByName` is needlessly excessive to satisfy those two fields.  I say go back to the tighter formatting that was there originally.

compiler/infra/STLUtils.cpp
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2017
```
0xdaryl: Please use the 2017, 2017 format.

compiler/infra/STLUtils.hpp
```diff
@@ -0,0 +1,34 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2017
```
0xdaryl: Please use the 2017, 2017 format.

compiler/infra/STLUtils.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+/** 
+ * Declares utility functions to be used with STL containers.
+ */
+
+#ifndef TR_STL_HPP
+#define TR_STL_HPP
+
+/**
+ * Returns true if the string pointed by s1 is 'less than' the one pointed
```
0xdaryl: Both "pointed"s should be "pointed to"

mstoodle:
@sajidahmed21 Can I ask how far away you think this pull request is from being complete?

In future, I think I would prefer to see a logical set of smaller pull requests, if possible, so we can get changes into the code base, because this set of changes, as it stands, is going to present a huge merge conflict for anyone with an active JitBuilder pull request (he said, selfishly, but also thinking of others :) ).

sajidahmed21:
@mstoodle As you may already know, currently I am working on integrating our custom allocator in these hash tables. We were having some difficulty earlier in doing that due to implicit generation of constructors (thanks to C++). Thankfully, I consider this issue to be solved for now, and I am currently planning to get these changes done soon (hopefully, by today), unless something new pops up during the review(s).

I agree, going forward it'll be better to break down these changes into smaller chunks.

sajidahmed21:
@mstoodle @0xdaryl As discussed, I've modified all the maps to use the corresponding TypeDictionary memory region for allocating their contents.
In order to do this, I had to delete the TypeDictionary copy constructor to avoid ambiguous compiler errors in external compilation units.

Some of the naming of the maps (and their `typedef`d types) hasn't been great, I've to say.
Also, note that Travis builds are expected to fail as of now, since I haven't updated the CMake builds to include the new `compiler/infra/STLUtils.*` files that I have created.

Please review the current changes. The CMake change will likely be a very small one and should not affect the existing changes.

sajidahmed21:
Update: CMake list has been updated to include the new file. Travis builds should be passing now.

sajidahmed21:
@mstoodle @0xdaryl Can you please review this PR? The earlier I get feedback, the easier it will be for me to propagate the changes to the dependent `std::string` changes in #1275

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -32,8 +32,10 @@
 #include "compile/Compilation.hpp"
 #include "compile/SymbolReferenceTable.hpp"
 #include "control/Recompilation.hpp"
+#include "infra/Assert.hpp"
 #include "infra/Cfg.hpp"
 #include "infra/HashTab.hpp"
```
0xdaryl: I don't think there's a need for this #include any longer.  Please remove.

compiler/ilgen/MethodBuilder.hpp
```diff
@@ -27,8 +27,10 @@
 #endif
 
 
+#include <map>
 #include <fstream>
 #include "ilgen/IlBuilder.hpp"
+#include "env/TypedAllocator.hpp"
```
mstoodle: You should really add env/TypedAllocator.hpp and its dependent files to jitbuilder/build/rules/common.mk (the list of copied header files) in this commit. Otherwise, isn't jitbuilder broken with this commit because TypedAllocator.hpp won't be on the usual include path?

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -349,15 +352,16 @@ MethodBuilder::lookupSymbol(const char *name)
...
 
    TR::SymbolReference *symRef;
-   TR_HashId typesID;
-   _symbolTypes->locate(name, typesID);
+   SymbolTypeMap::iterator symTypesIterator =  _symbolTypes.find(name);
+
+   TR_ASSERT_FATAL(symTypesIterator != _symbolTypes.end(), "Symbol '%s' doesn't exist", name);
 
-   TR::DataType type = ((TR::IlType *)(_symbolTypes->getData(typesID)))->getPrimitiveType();
+   TR::DataType type = symTypesIterator->second->getPrimitiveType();
```
mstoodle: I confess I'm not a big fan of these iterator->second references. Would it be too painful to put them into a variable of the same type, even if you just use auto? I just loathe templates when one tries to debug this code, so if I can avoid grubbing around in a std::pair, that feels worth it to me.

compiler/ilgen/TypeDictionary.cpp
```diff
@@ -676,9 +660,9 @@ TypeDictionary::NotifyCompilationDone()
       }
 
    // clear all symbol references for union fields
-   TR_HashTabIterator unionIterator(_unionsByName);
-   for (UnionType *aUnion = (UnionType *)unionIterator.getFirst();aUnion;aUnion = (UnionType *)unionIterator.getNext())
+   for (UnionMap::iterator it = _unionsByName.begin(); it != _unionsByName.end(); it++)
       {
+      UnionType *aUnion = it->second;
```
mstoodle: yay!

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -331,6 +331,8 @@ MethodBuilder::symbolDefined(const char *name)
 void
 MethodBuilder::defineSymbol(const char *name, TR::SymbolReference *symRef)
    {
+   TR_ASSERT_FATAL(_symbols.find(name) == _symbols.end(), "Symbol '%s' already defined", name);
```
mstoodle: in future, commits like this one could be made in its own pull request. As it stands, this commit doesn't *really* fit in with the title of the MR.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -522,14 +522,16 @@ MethodBuilder::DefineFunction(const char* const name,
...
-   TR::ResolvedMethod *method = new (PERSISTENT_NEW) TR::ResolvedMethod((char*)fileName,
-                                                                        (char*)lineNumber,
-                                                                        (char*)name,
-                                                                        numParms,
-                                                                        parmTypes,
-                                                                        returnType,
-                                                                        entryPoint,
-                                                                        0);
+   // Store ResolvedMethod in the heap region of the TypeDictionary memory segment
+   TR_Memory *memory = typeDictionary()->trMemory();
```
mstoodle: Maybe it would be better for the MethodBuilder object to have its own cached _trMemory field, initialized from the TypeDictionary? that way all the allocations could be against the MethodBuilder's memory region and we would then be free to have the MethodBuilder allocate its own memory region in future without having to update all the allocations?

I thought we had discussed this point before, so maybe I'm just not remembering the reason why that's impractical...

sajidahmed21: I don't see any issue with doing that. @lmaisons can you think of any reason why this might not work out?

compiler/ilgen/MethodBuilder.hpp
sajidahmed21: I had already included them with an earlier commit (e27da1b63c836c655e42fb3c17e227f74f8dc78e) :)

sajidahmed21:
@0xdaryl @mstoodle I've made the changes you suggested, which includes MethodBuilder having its own memory region.
I've updated the description at the top to summarize everything that I've included so far in this PR.

Please review.

jitbuilder/build/rules/common.mk
```diff
@@ -120,6 +123,7 @@ $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/IlGen.hpp: $(FIXED_SRCBASE)/$(JIT_
 JITBUILDER_FILES=$(RELEASE_DIR)/Makefile \
              $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/env/defines.h \
              $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/env/jittypes.h \
+             $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/env/TypedAllocator.hpp \
```
mstoodle: didn't you also add #include "infra/Assert.hpp" which itself requires Annotations.hpp and CompilationException.hpp ?

sajidahmed21: @mstoodle I don't see infra/Assert.hpp being required anywhere in the JitBuilder headers. Not sure why you would want to add that.

mstoodle: Technically, I (mistakenly) asked a question whose answer is "no". So it's ok.

sajidahmed21: Sorry, I mistakenly read your question as "why didn't you add infra/Assert.hpp to the JitBuilder includes?"

0xdaryl:
@sajidahmed21 , please look into the OSX CI failure to verify its not an infra problem.

mstoodle:
@0xdaryl @sajidahmed21 I actually restarted the OSX job once last night, so it's unlikely to be an infrastructure problem (i.e. the current failure is the second one).

0xdaryl:
Its actually the third--I restarted it too.  :-)

sajidahmed21:
Yes, I'm currently investigating it, and have been able to reproduce locally - pretty certain it's not an infra problem.

Will update as I figure out the cause.

sajidahmed21:
@0xdaryl @mstoodle As it turns out, C++ destructors cleanup member variables **after** executing whatever is in the body of the user defined destructor. Since we were cleaning up the MethodBuilder memory region in the destructor body, the std::maps in MethodBuilder ended up with dangling references to their contents, which produced seg faults when the std::map destructors were later called.

I've fixed this by explicitly clearing up the map contents (in both MethodBuilder and TypeDictionary) **before** destroying the respective memory regions.

(Hopefully) this PR should now be ready for merging.

mgaudet:
Does the destruction of the region have to happen as part of the body? If so, it rather seems this is a trap others will fall into, and we ought to open an issue about seeing if we can't give regions normal destruction semantics. Otherwise, is it possible to reorder the members so that the region is destroyed last?

mstoodle:
@mgaudet just curious if I've missed some subtlety: why do say regions don't have normal destruction semantics? Isn't this just an ordering issue since the region and the objects stored in that region need to be constructed/destructed in the proper order?

mgaudet:
So, I've not looked deeply into this. What @sajidahmed21 seems to be saying above is that for some reason, [the MethodBuilder region destruction](https://github.com/eclipse/omr/pull/1288/commits/7bf9e3c4a30b423f7fa90d238417595374e54a6a#diff-fd4d54cfe9f8e717db659292cb14074cR147) must happen as part of the destructor body. This decision however appears to be viral: Because it's destroyed in the body, so too must everything that's dependent on that region.

My question here is mostly: Do regions have to be destroyed as part of the destructor body, if so why, and maybe we ought to have an issue to see if that requirement can be removed, as this appears to be a fairly substantial footgun when it comes to regions.

lmaisons:
Regions have normal destructor semantics.  The problem here is that the member of the class isn't a `Region` but a `Region *`.  The only place one can *explicitly* destruct an indirectly-allocated is in the body of the destructor.  If one wants a destruction order that is the same as what would occur had the Region been a direct member, one needs some form of RAII wrapper.

sajidahmed21:
Currently, we are storing `TR::Region` and the associated `TR::SegmentProvider` as pointers within MethodBuilder and TypeDictionary. This demands that we clean up these objects explicitly in the destructor body.

An alternative that would not require us to destroy these objects as part of the destructor body, would be to store them by value (instead of pointers). This would ensure that they are destroyed by normal destruction semantics, just like the other objects. However, this would mean that we need to add a whole bunch of header files in the JitBuilder release includes:

env/Region.hpp
env/SystemSegmentProvider.hpp

# Needed by Region.hpp
infra/ReferenceWrapper.hpp
env/TypedAllocator.hpp (already included)
env/MemorySegment.hpp
env/RawAllocator.hpp

# Needed by MemorySegment.hpp
env/RawAllocator.hpp (already included)
infra/Assert.hpp

# Needed by SystemSegmentProvider.hpp
env/SegmentAllocator.hpp

# Needed by SegmentAllocator.hpp
env/SegmentProvider.hpp

# Needed by Assert.hpp
infra/Annotations.hpp
infra/CompilationException.hpp

sajidahmed21:
@mstoodle As discussed, I've added comments warning about the use of regions. Please check if its good for merging now.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -80,16 +86,22 @@ namespace OMR
 
 MethodBuilder::MethodBuilder(TR::TypeDictionary *types, OMR::VirtualMachineState *vmState)
    : TR::IlBuilder(asMethodBuilder(), types),
+   // Note: Any allocations in the following memory region that will get deallocated by C++ destruction semantics along with this MethodBuilder class, should instead
+   // be explicitly cleared / deallocated in the MethodBuilder destructor *before* the region itself is destroyed in order to avoid undefined behaviour from invalid
+   // memory references.
```
mstoodle: Sorry, I don't understand the comment. It says "Any allocations in the following memory region" which sounds a lot like the std::map's that are allocated as part of this MethodBuilder object. But you must not mean those allocations because the destructor below doesn't explicitly clear / deallocate those std::maps.

After the conversation earlier today, I was expecting a description along the lines of: because the memory region is stored here as a pointer, we can't rely on the usual C++ destruction semantic to destruct and deallocate the memory region  only after the objects allocated from it have been destructed. Also, because the impact is really in the destructor, I expected the comment to be in the destructor rather than in the constructor.

Special points if you explain that the memory region is stored as a pointer to avoid increasing the number of header files needed to compile against the JitBuilder library.

compiler/ilgen/MethodBuilder.cpp
```diff
@@ -80,16 +86,22 @@ namespace OMR
...
+   _segmentProvider(static_cast<TR::SegmentProvider *>(new(TR::Compiler->persistentAllocator()) TR::SystemSegmentProvider(MEM_SEGMENT_SIZE, TR::Compiler->rawAllocator))),
```
mstoodle: While I'm reviewing...why is this static_cast needed?

compiler/ilgen/MethodBuilder.cpp
sajidahmed21: The comment doesn't recommend deallocating the maps, since as you said, they are allocated within the MethodBuilder object. It recommends to deallocate "any allocations *in the region*", which in our case are the actual `std::pairs` stored by the map.

To avoid any ambiguity, I'm replacing the words "following memory region" with `_memoryRegion` and rephrasing the sentence overall to make it more explicit.

Also, my intention for putting the comment in the constructor was to make it more noticeable, since its a more likely place for a new contributor to allocate objects in the region. (I did add a note on the destructor though).

compiler/ilgen/MethodBuilder.cpp
sajidahmed21: I don't see it being required, and I think I may have asked @lmaisons about the [existing one](https://github.com/eclipse/omr/blob/master/compiler/ilgen/TypeDictionary.cpp#L469) in TypeDictionary.cpp, not sure what was the answer.

I'm getting rid of both since you don't seem to see it being required either.

compiler/ilgen/MethodBuilder.cpp
mstoodle: Sorry. My comment wasn't well constructed, mostly because I think I reviewed the file before all the changes had been made to it. I'll accept the comment being at the constructor.

mstoodle:
Thanks for sticking it out, @sajidahmed21 ! LGTM :)

PR #1291, opened 2017-06-23
Remove project code from x86 directCallEvaluator

0xdaryl:
Remove the J9-specific parts of OMR::X86::TreeEvaluator::directCallEvaluator
as they will be handled downstream.


vijaysun-omr:
Straightforward cleanup. Merging.

PR #1292, opened 2017-06-23
Add two new generateInstruction APIs

fjeremic:
Add two new APIs:

- generateS390BranchInstruction overload targeting RR instructions
- generateDataConstantInstruction

The former is an overload to generate RR instruction formats such as
BCR, and the latter is a new API to generate data constant
pseudo-instructions in the instruction stream.

### DIFF
--- a/compiler/z/codegen/S390GenerateInstructions.hpp
+++ b/compiler/z/codegen/S390GenerateInstructions.hpp
@@ -88,6 +88,32 @@ TR::Instruction *generateS390BranchInstruction(
                    TR::LabelSymbol *sym,
                    TR::Instruction *preced = 0);
 
+/** \brief
+ *     Generates a branch instruction accepting a branch condition and a target register.
+ *
+ *  \param cg
+ *     The code generator used to generate the instructions.
+ *
+ *  \param op
+ *     The mnemonic of the branch instruction.
+ *
+ *  \param node
+ *     The node to which the generated instruction will be associated with.
+ *
+ *  \param compareOpCode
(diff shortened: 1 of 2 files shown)


compiler/z/codegen/S390GenerateInstructions.hpp
```diff
@@ -1300,6 +1326,29 @@ TR::Instruction *generateDirectCall(
...
+ *     Generates a data constant in the instruction stream.
+ *
+ *  \param cg
+ *     The code generator used to generate the instructions.
+ *
+ *  \param op
+ *     The mnemonic of the data constant pseudo instruction.
+ *
+ *  \param node
+ *     The node to which the generated instruction will be associated with.
```
joransiu: Maybe "The node with which the generated instruction will be associated."?

compiler/z/codegen/S390GenerateInstructions.hpp
```diff
@@ -1300,6 +1326,29 @@ TR::Instruction *generateDirectCall(
...
+ *
+ *  \param data
+ *     The data to inline in the instruction stream.
```
joransiu: data -> data constant

fjeremic: Fixed in 9e5eb16.

compiler/z/codegen/S390GenerateInstructions.hpp
fjeremic: Fixed in 9e5eb16.

PR #1297, opened 2017-06-25
Add some compiler tests using Tril

Leonardo2718:
These changes add some compiler tests that use Tril to generate input IL. These include Tree Verifier tests and some initial opcode tests. The opcode tests are intended to show how tests that could eventually replace the existing opcode tests (in the test compiler) can be written.

It should be noted that some of the Tree Verifier tests currently fail because the Tree verifier has historically been poorly maintained and is not behaving as expected.

[ci skip] because Tril is not part of continuous integration

### DIFF
--- a/fvtest/compilertriltest/.gitignore
+++ b/fvtest/compilertriltest/.gitignore
@@ -1,0 +1,2 @@
+build*/
+CMakeLists.txt.user
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -33,6 +33,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fdiagnostics-color=always -fno-rtti 
 add_executable(comptest
    main.cpp
    TreeVerifierTest.cpp
+   OpCodeTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -1,0 +1,46 @@
+###############################################################################
+#
+# (c) Copyright IBM Corp. 2017, 2017
+#
+#  This program and the accompanying materials are made available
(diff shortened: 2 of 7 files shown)


0xdaryl:
I think this is a great start on some of these kinds of tests.

I must admit I'm not too keen on the directory name `compilertriltest` but I don't have a great alternative suggestion.

I'm also musing about namespaces and whether it makes sense to create the test functions within a namespace like the rest of the projects that extend the compiler technology.  Nothing is likely to extend these tests though, and they're more of a consumer of JitBuilder and the compiler technology rather than an extender.  Do you (or anyone) have a strong opinion either way?

Leonardo2718:
@0xdaryl 

I agree that `compilertriltest` is not a good name but I couldn't think of anything else.

The question of namespaces and extensions is rather complex. This is partially because it depends heavily on how the compiler technology is expected to be consumed and some limitations imposed by the gtest framework. I will try summarizing my personal opinion without going into the details of my reasoning.

Simplifying things tremendously, I think downstream projects should create their own tests and not rely on tests provided in OMR to test _their_ code. If a downstream project makes extensions to the compiler technology, the OMR tests wont cover the new functionality. If they override some behavior in the compiler technology, the OMR tests will break as they are designed to test the behavior implemented in OMR. Furthermore, combining the gtest framework with the extension mechanism used in the compiler technology is likely to be very tricky. Downstream projects are therefore likelier to benefit more from creating their own tests than from reusing the tests in OMR.

That said, I could see a downstream project wanting to reuse some of the testing _infrastructure_. For example, some of the convenience functions for the opcode tests could be useful to a project that adds some custom opcodes to the compiler technology. In this scenario, extensions could be made either using the compiler technology's extension mechanism or using regular (dynamic) polymorphism.

All this considered, I don't really see a need for putting the tests themselves in a namespace. In fact, this may complicate things a little for gtest, as it's rather sensitive to scope resolution.

However, it makes sense for the testing infrastructure to be in a namespace to ease consumability in downstream projects. But, as with the directory name, I don't know what a good namespace would be.

PR #1304, opened 2017-06-26
Compilation object cleanup

lmaisons:
Change the Compilation object's region member type to a reference,
allowing comp->region() to return the same region used by heap memory.

Clean up the initializer list of OMR::Compilation so that members
initialization are specified in the same order as their declaration
inside the class definition. (Which is the order they get initialized
anyway, regardless of their order in the initializer list.)

Change permissions of _signature member and replace its external uses
with a call to the signature() method. _signature should still be the
first member, as this can significantly ease defect investigations.

Remove _aotClassInfo from OMR::Compilation as it never gets used.


lmaisons:
This change will require downstream changes from projects that extend the definition of OMR::Compilation

lmaisons:
Going to close / re-open to see if can pick up the copyright changes

lmaisons:
Reopening as per last...

lmaisons:
>.< wrong PR

compiler/compile/OMRCompilation.hpp
```diff
@@ -969,15 +965,15 @@ class OMR_EXTENSIBLE Compilation
    void reportFailure(const char *reason);
 
 protected:
+   // Signature of the method we are compiling -- keep this at offset 0
```
fjeremic: Perhaps incorporating a piece of the PR comment would make sense as to _why_ one would want to keep this field at offset 0, i.e.

> _signature should still be the first member, as this can significantly ease defect investigations.

0xdaryl: It's not clear to me whether this is a hard requirement driven by some downstream project built on OMR or merely a suggestion.  There are certainly examples of other places in OMR where a downstream project has built tooling (and hence an undocumented dependence) on the location of a particular field in a data structure.  This is wrong in my opinion.

I am more in favour of removing the comment entirely, especially if its only a "soft" requirement whose presence will likely discourage future refactoring.  If it truly is a "hard" requirement then it should be documented as such and properly justified in comments in OMR.

lmaisons: @mpirvu @klangman Would either of you happen to know why we kept signatures at offset 0?

0xdaryl: In the interests of moving this forward and merging, please remove the comment unless there is a justifiable, hard requirement for this field to be the first one in this class in OMR.  I don't believe there is.

lmaisons:
Removed the comment about the `_signature` field

PR #1306, opened 2017-06-27
Cleanup undocumented branch conditions

arianneb:
Removed all undocumented branch conditions and replaced
them with their corresponding documented condition.
Ex. COND_BERC changed to either COND_BE or COND_BER.
Those conditions which did not have an obvious corresponding
documented branch condition have been replaced with their
corresponding mask value.

Changed function useBranchOnZeroConditionMaskIfPossible(...)
in omr/compiler/z/codegen/OMRTreeEvaluator.cpp to accomodate
changes.

Added documentation in OMRInstOpCode.hpp for each branch
condition. Added documentation to modified function and changed
function name from useBranchOnZeroConditionMaskIfPossible(...) to
convertComparisonBranchConditionToTestUnderMaskBranchCondition(...)

Removed unused jump conditions in OMRInstOpCode.hpp, InstOpCode.hpp,
and S390BranchCondNames.cpp.


mgaudet:
Just for clarification: When you say "undocumented", do you mean these aren't documented conditions in TR, or aren't documented in the z/Architecture principles of operation?

arianneb:
Those that were removed were not documented in the z/Architecture Reference Summary or in the POPS. Most of them were also not documented in the code base.

0xdaryl:
Looks good on my pass through it.  Are all these branch substitutions performance neutral?

arianneb:
As far as I know they are performance neutral. The substitutions all translate to the same mask value that they had before, so nothing should have changed under the hood.

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1403,11 +1403,13 @@ OMR::Z::TreeEvaluator::ificmpeqEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
       {
-      reg = generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BERC, TR::InstOpCode::COND_BERC);
+      reg = generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BE, TR::InstOpCode::COND_BE);
       }
    else
       {
-      reg = generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNERC, TR::InstOpCode::COND_BNERC);
+      reg = generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNE, TR::InstOpCode::COND_BNE);
+
+      traceMsg(TR::comp(), "Generated op code for %p\n",  node);
```
fjeremic: This looks like a debugging line. Is it necessary?

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1536,12 +1538,12 @@ OMR::Z::TreeEvaluator::iflcmpeqEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
       {
-      generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BERC, TR::InstOpCode::COND_BERC);
+      generateS390CompareBranch(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BE, TR::InstOpCode::COND_BE);
       }
    else
       {
-      generateS390lcmpEvaluator(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNERC, TR::InstOpCode::COND_NOP, TR::InstOpCode::COND_BERC, CMP4CONTROLFLOW);
-      }
+      generateS390lcmpEvaluator(node, cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BNE, TR::InstOpCode::COND_NOP, TR::InstOpCode::COND_BE, CMP4CONTROLFLOW);
+}
```
fjeremic: Indentation needed here.

compiler/z/codegen/OMRInstOpCode.hpp
```diff
@@ -390,94 +390,60 @@ class InstOpCode: public OMR::InstOpCode
...
-      COND_BPRC,
-      COND_BRC,
-      COND_BZ,
-      COND_BZR,
-      COND_BZRC,
-      COND_NOP,
-      COND_VGNOP,
+                        /* Instruction Type    |  Meaning (Operands)
+                           ----------------------------------------- */
+      COND_NOPR,        // Conditional Branch     No operation (RR)
```
fjeremic: Perhaps this branch condition can be moved to be beside `COND_NOP`?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3952,31 +3952,39 @@ genCompareAndBranchInstructionIfPossible(TR::CodeGenerator * cg, TR::Node * node
    return cursor;
    }
 
-TR::InstOpCode::S390BranchCondition useBranchOnZeroConditionMaskIfPossible(TR::Node * node, TR::InstOpCode::S390BranchCondition brCond)
+/**
+ * \fn convertComparisonBranchConditionToTestUnderMaskBranchCondition(TR::Node * node, TR::InstOpCode::S390BranchCondition brCond)
```
fjeremic: I don't think this is necessary. Doxygen will automatically group this function comment with the subsequent definition.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3952,31 +3952,39 @@ genCompareAndBranchInstructionIfPossible(TR::CodeGenerator * cg, TR::Node * node
...
+ * \brief Converts a branch condition mask to the correct mask for the test under mask instruction being generated 
+ * 
+ * \param node Constant, 1st or 2nd child of IL tree root
+ * \param brCond Branch condition mask 
+ * \return newCond New mask to be used in generated test under mask instruction 
+ *
+ */
+TR::InstOpCode::S390BranchCondition convertComparisonBranchConditionToTestUnderMaskBranchCondition(TR::Node * node, TR::InstOpCode::S390BranchCondition brCond)
```
fjeremic: This function returns a branch condition, not a mask. Would renaming `node` to `constNode` be more intuitive? The comment also seems to imply restrictions on where `node` comes from, i.e. first or second child. This restriction does not make sense. How about:

```
/** \brief
 *     Converts a comparison branch condition to the correct branch condition for the test under mask instruction 
 *     being generated.
 *
 *  \param constNode
 *     The constant the comparison branch condition is comparing against.
 *
 *  \param brCond
 *     The comparison branch condition to be converted.
 *
 *  \return
 *     The branch condition to be used in generated test under mask instruction.
 */
```

Regardless the comment seems to imply the consumer of this function must pass in a constant. Lets enforce this invariant with an assert at the start of the function.

compiler/z/codegen/S390BranchCondNames.cpp
```diff
@@ -21,91 +21,54 @@
...
-   "BHRC",
    "BL",
    "BLR",
-   "BLRC",
    "BM",
-   "BMR",
-   "BMRC",
-   "BNANRC",
-   "BNCRC",
+   "BMR ",
```
fjeremic: There is an extra space before the end of quote here.

Issue #1305, opened 2017-06-27
IlValue::storeOver() creates illegal IL trees


mstoodle:
This problem was discovered as part of the Lua Vermelha port that @charliegracie has been working on. He has a "register file" version of omr/compiler/ilgen/VirtualMachineOperandStack.cpp that he has been using to better optimize Lua code. It turns out that the MergeInto facility, which uses IlValue::storeOver(), can create a block along an edge into a merge that can look like:

````
n351n     BBStart <block_78>                                                                  [0x0000dc8268] bci=[-1,11,-] rc=0 vc=0 vn=- li=- udi=- nc=0
n355n     lstore  <temp slot 9>[#624  _T9] [flags 0x10004 0x0 ]                               [0x0000dc8348] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=1
n189n       ==>ladd
n358n     lstore  <temp slot 19>[#634  _T19] [flags 0x10004 0x0 ]                             [0x0000dc83f0] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=1
n237n       ==>lload
n361n     lstore  <temp slot 20>[#635  _T20] [flags 0x10004 0x0 ]                             [0x0000dc8498] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=1
n264n       ==>lsub
n364n     lstore  <temp slot 21>[#636  _T21] [flags 0x10004 0x0 ]                             [0x0000dc8540] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=1
n250n       ==>lconst 2
n367n     lstore  <temp slot 22>[#637  _T22] [flags 0x10004 0x0 ]                             [0x0000dc85e8] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=1
n255n       ==>lconst 1
n369n     goto --> block_25 BBStart at n47n                                                   [0x0000dc8658] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=0
n352n     BBEnd </block_78>                                                                   [0x0000dc82a0] bci=[-1,0,-] rc=0 vc=0 vn=- li=- udi=- nc=0
````

Note the commoned references being used even though the node is not computed in this basic block (and it does not extend another basic block). That's illegal IL.

[...]

Issue #1298, opened 2017-06-25
Fix the compiler's Tree Verifier


Leonardo2718:
As reported by #1297, the Tree Verifier in the compiler is not currently behaving as expected.

The Tree Verifier is supposed to check that some given Testarossa Intermediate Language (IL)  is well-formed and conforms to the rules of the language. However, as is shown by the tests in #1297, it is both accepting invalid IL and rejecting valid IL.

Running the test in #1297 using `TR_Options=paranoidoptcheck,tracefull,log=trace.log TR_DEBUG="paranoidoptcheck checkTypes" ./comptest --gtest_filter="TreeVerifierTest/*"` outputs:

```Note: Google Test filter = TreeVerifierTest/*
[==========] Running 15 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 11 tests from TreeVerifierTest/IllformedTrees
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/0
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/0 (78 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/1
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/1 (26 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/2
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/2 (49 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/3
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/3 (497 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/4
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/4 (494 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/5
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/5 (38 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/6
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/6 (43 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/7
[       OK ] TreeVerifierTest/IllformedTrees.FailCompilation/7 (38 ms)
[ RUN      ] TreeVerifierTest/IllformedTrees.FailCompilation/8
[...]

PR #1307, opened 2017-06-27
Store value to auto in IlValue::storeOver

mstoodle:
IlValue::storeOver does not handle the incoming parameter "value"
properly, in that it does not ensure that it has been stored to
an auto symbol if it's computed in a different block. This mistake
can result in illegal IL in the form of node references that cross
basic block boundaries.

The fix ensures that value has been written to an auto if necessary,
and also leaves the IlValue in a more consistent state (earlier
version would update _nodeThatComputesValue without also updating
_blockThatComputesValue; really neither of these should be updated).

Fixes: #1305

### DIFF
--- a/compiler/ilgen/IlValue.cpp
+++ b/compiler/ilgen/IlValue.cpp
@@ -95,15 +95,24 @@ OMR::IlValue::storeOver(TR::IlValue *value, TR::Block *block)
       {
       // more commonly, if any value is in another block or this use will be in a different block, first ensure this value is
       // stored to an auto symbol
-      // NOTE this may mean nodes may be stored to more than one auto, but that's ok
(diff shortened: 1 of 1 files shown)


mstoodle:
I verified that all of the JitBuilder code samples pass even with "export TR_Options=paranoidOptCheck" set, which verifies the trees between each optimization pass (at least, as well as we verify trees given observations in #1298)

mstoodle:
I verified that, with this change in place, operandstacktests does work when logging is enabled (and failed with logging enabled before this change).

charliegracie:
I have also verified that the proposed change fixes my Lua Vermelha issue as well.  Thanks Mark!

Issue #826, opened 2017-03-06
Valgrind support for the Garbage Collector and JIT memory


mgaudet:
_This issue is derived from [our Google Summer of Code 2017 Ideas List](https://wiki.eclipse.org/Google_Summer_of_Code_2017_Ideas#Eclipse_OMR)_ 
> Valgrind support for the Garbage Collector and JIT memory http://valgrind.org/docs/manual/mc-manual.html#mc-manual.mempools

Varun-garg:
I am sharing my results for Valgrind on sample gctest to help get started with this issue. (ubuntu, gcc 5.4.0)

```
$ valgrind ./omrgctest -configListFile=fvtest/gctest/configuration/fvConfigListFile.txt
==38115== Memcheck, a memory error detector
==38115== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==38115== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==38115== Command: ./omrgctest -configListFile=fvtest/gctest/configuration/fvConfigListFile.txt
==38115== 
[==========] Running 8 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 8 tests from configFile/GCConfigTest
[ RUN      ] configFile/GCConfigTest.test/0
==38115== Warning: set address range perms: large range [0x64cd000, 0x264dc000) (noaccess)
==38115== Warning: set address range perms: large range [0x64cd000, 0x264dc000) (noaccess)
[       OK ] configFile/GCConfigTest.test/0 (3403 ms)
[ RUN      ] configFile/GCConfigTest.test/1
==38115== Warning: set address range perms: large range [0x64cd000, 0x264dc000) (noaccess)
==38115== Warning: set address range perms: large range [0x64cd000, 0x264dc000) (noaccess)
[       OK ] configFile/GCConfigTest.test/1 (3264 ms)
[ RUN      ] configFile/GCConfigTest.test/2
[       OK ] configFile/GCConfigTest.test/2 (1879 ms)
[ RUN      ] configFile/GCConfigTest.test/3
[       OK ] configFile/GCConfigTest.test/3 (2113 ms)
[ RUN      ] configFile/GCConfigTest.test/4
[       OK ] configFile/GCConfigTest.test/4 (1507 ms)
[ RUN      ] configFile/GCConfigTest.test/5
[       OK ] configFile/GCConfigTest.test/5 (2020 ms)
[...]

PR #1309, opened 2017-06-27
WIP: Valgrind memcheck api integration in GC

Varun-garg:
#I would like to share the progress that I and @youngar's have made to add Valgrind support in garbage collector. This pull request does the following

1. Allocation of valgrind pool, addition new memory blocks in the pool and new objects in these blocks.
2. target `omrgcexample` works correctly, running `valgrind ./omrgcexample` throws no errors. It did not show errors before valgrind api was integrated, but after marking heap memory as no access we came across many errors which were fixed by allocating objects via valgrind's api and marking places in omr runtime as accessable when needed.
3. Added valgrind package to linux and mac in travis configuration to allow automatic builds with memcheck in future.
4. Added `--enable-valgrind` argument as configure argument. It is now disabled by default to avoid unnecessary requirement of valgrind package and also fixes appveyor builds.

Currently I am working on `omrgctest` as tons of valgrind errors are still there in it. Most errors are overlapping errors which may be due to not using api at places when we are releasing objects or moving them. 

References: #826


Varun-garg:
This pull request has some issues. I will open another one shortly.

Varun-garg:
All of the commits are amended and new pull request is #1311

Issue #1281, opened 2017-06-21
Expose limits imposed by cgroups via port library APIs


ashu-mehra:
Runtimes like JVM rely on resource limits like memory and cpus to adjust their internal parameters which includes heap size, GC helper threads, JIT helper threads etc.
Cgroups in Linux provide a mechanism to impose limits on such resources. Therefore, when running as part of a cgroup, the runtime should adjust its internal parameters based on the resource constraints imposed by the cgroup.
Currently the information about resource constraints imposed by cgroup is not provided by the port library in any way. If a runtime wants to take into consideration the limits imposed by the cgroup, it would have to figure this out on its own.
It would be useful for the runtimes if port library can provide APIs that would enable them to query the resource constraints imposed by the cgroup.
The proposal is to add mechanism for following tasks:
1. Add capability to read `/proc/<pid>/cgroup` file.
Using this file, the process can figure out the cgroup it belongs to for a given subsystem.
2. Expose APIs for providing memory limits and CPU limits imposed by the cgroup. For start only these two limits would be exposed. More can be added later as and when required.
3. Now, the above mechanism would work when running on bare metal or a VM, but would break when running in a docker container. This is because in a docker container the `/proc/<pid>/cgroup` gives the cgroup of the process on the host, and not the one in the container. AFAIK when running in container, the cgroup would always be the root cgroup. Therefore, we also need to detect whether the process is running in a docker container or not, and use correct cgroup to get the resource limits.

As an example, if I log in to a docker container, and cat `/proc/1/cgroup`, I get following contents:
```
$ docker exec -it dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612 /bin/bash
root@dfee854963d4:/# cat /proc/1/cgroup
11:pids:/docker/dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612
10:perf_event:/docker/dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612
9:freezer:/docker/dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612
8:cpuset:/docker/dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612
7:blkio:/docker/dfee854963d4575d263785ddee0f3c398355b354e6608c5777a3a2e40fef3612
[...]

PR #1310, opened 2017-06-27
Add support for cgroup v1 in Linux

ashu-mehra:
Add support for cgroup v1 in Linux
    
Added code for reading `/proc/<pid>/cgroup` file which provides
information about the cgroups that the process belongs to.
This information can be used to get resource constraints imposed
by the cgroup by reading appropriate file(s) under
`/sys/fs/cgroup/<subsystem>/<cgroup name>`.
    
Added following APIs to the port library to interact with cgroups:
1. omrsysinfo_is_cgroup_limits_supported()
2. omrsysinfo_is_cgroup_limits_enabled()
3. omrsysinfo_enable_cgroup_limits()
4. omrsysinfo_cgroup_get_memlimit()
    
omrsysinfo_cgroup_get_memlimit() returns the memory limit imposed
by the cgroup by reading the contents of
`/sys/fs/cgroup/memory/cgroup name>/memory.limit_in_bytes` file.
    
There is more work required to fully support cgroups, which includes
determine if we are in container or not,
adding APIs to expose other resource limits,
adding support for cgroup v2 etc.
These will be done in subsequent commits.

Issue: #1281


ashu-mehra:
@charliegracie can you please review.

ashu-mehra:
@DanHeidinga this is the initial PR for adding support for cgroup.

ashu-mehra:
I am updating the changes based on this [comment](https://github.com/eclipse/omr/issues/1281#issuecomment-312176400)

charliegracie:
I have been reading through the changeset.  I am not that familiar with cgroups so it is taking a bit longer as I investigate them and the appropriate way to get information about them.

ashu-mehra:
@charliegracie can this be merged now?

port/unix_include/omrcgroup.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+#ifndef omrcgroup_h
+#define omrcgroup_h
+
+#if defined(LINUX)
+
+typedef struct CgroupEntry {
```
DanHeidinga: Should this be prefixed with OMR to avoid namespace issues?  ie `omrCgroupEntry`

port/unix_include/omrcgroup.h
```diff
@@ -0,0 +1,32 @@
...
+        int32_t hierarchyId; /**< cgroup hierarch ID*/
+	char *subsystem; /**< name of the subsystem*/
```
DanHeidinga: Nitpick: some of the fields have an extra leading space.

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+	cgroup = getCgroupNameForSubsystem(portLibrary, PPG_cgroupEntryList, subsystemNames[subsystem]);
+	if (NULL == cgroup) {
+		char errBuf[512];
+
+		portLibrary->str_printf(portLibrary, errBuf, sizeof(errBuf), "cgroup name for subsystem %s is not available", subsystemNames[MEMORY]);
+		rc = portLibrary->error_set_last_error_with_message(portLibrary, OMRPORT_ERROR_SYSINFO_CGROUP_NAME_NOT_AVAILABLE, errBuf);
+		goto _end;
+	}
+
+	portLibrary->str_printf(portLibrary, fileToRead, sizeof(fileToRead), "/sys/fs/cgroup/%s/%s/%s", subsystemNames[subsystem], cgroup, fileName);
```
DanHeidinga: Do we need to check for truncation here?  If the string being written is larger than the allowed buffer (very unlikely), we might fopen the wrong thing.

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+{
+	int32_t rc = OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;
+#if defined(LINUX)
+	char fileToTest[PATH_MAX];
+	struct statfs buf = {0};
+
+	if (NULL == PPG_cgroupEntryList) {
+		/* Check if tmpfs is mounted on /sys/fs/cgroup */
+		portLibrary->str_printf(portLibrary, fileToTest, sizeof(fileToTest), "/sys/fs/cgroup/");
+		rc = statfs(fileToTest, &buf);
```
DanHeidinga: Why do the str_printf() here?  Couldn't the `"/sys/fs/cgroup/"` string be passed directly to `statsfs` saving the printf call and allowing the `fileToTest` buffer to be removed?

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+	}
+
+	va_start(args, format);	
+	rc = vfscanf(file, format, args);
+	va_end(args);
+
+	if (numItemsToRead != rc) {
+		char errBuf[512];
+
+		portLibrary->str_printf(portLibrary, errBuf, sizeof(errBuf), "unexpcted format of file %s", fileName);
```
DanHeidinga: nitpick: typo `unexpcted` --> `unexpected`

DanHeidinga: Is there also a possible truncation issue here?  More minor as it only affects the error message output...

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+	if (!portLibrary->sysinfo_is_cgroup_limits_enabled(portLibrary)) {
+		rc = portLibrary->error_set_last_error_with_message(portLibrary, OMRPORT_ERROR_SYSINFO_CGROUP_LIMITS_DISABLED, "cgroup limits is disabled");
+		goto _end;
+	}
+	
+	cgroup = getCgroupNameForSubsystem(portLibrary, PPG_cgroupEntryList, subsystemNames[subsystem]);
+	if (NULL == cgroup) {
+		char errBuf[512];
+
+		portLibrary->str_printf(portLibrary, errBuf, sizeof(errBuf), "cgroup name for subsystem %s is not available", subsystemNames[MEMORY]);
```
DanHeidinga: I'll whine about checking for truncation here as well but its less of a concern as the subsystem names are more constrained.

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+
+			if (NULL == cgName) {
+				rc = addCgroupEntry(portLibrary, &cgEntryList, hierId, cursor, cgroup);
+				if (0 != rc) {
+					goto _end;
+				}
+			} else {
+				char errBuf[512];
+
+				portLibrary->str_printf(portLibrary, errBuf, sizeof(errBuf), "multiple entries for subsystem %s found in process cgroup file", cursor);
```
DanHeidinga: Truncation concern?  See review comments below for more details. (I started reviewing at the bottom)

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3158,356 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+
+		/* Following is the description of /proc/<pid>/cgroup copied from 'man' page for proc:
+ 		 *
+ 		 * Each entry in /proc/<pid>/cgroup is of type:
+		 * 	5:cpuacct,cpu,cpuset:/daemons
+		 * The colon-separated fields are, from left to right:
+		 * 	1. hierarchy ID number
+		 * 	2. set of subsystems bound to the hierarchy
+		 * 	3. control group in the hierarchy to which the process belongs
+		 */
```
DanHeidinga: Love this comment!  It makes the code below very clear.

port/common/omrport.c
```diff
@@ -250,6 +250,10 @@ static OMRPortLibrary MasterPortLibraryTable = {
 	omrsysinfo_get_os_description, /* sysinfo_get_os_description */
 	omrsysinfo_os_has_feature, /* sysinfo_os_has_feature */
 	omrsysinfo_os_kernel_info, /* sysinfo_os_kernel_info */
+	omrsysinfo_is_cgroup_limits_supported, /* sysinfo_is_cgroup_limits_supported */
+	omrsysinfo_is_cgroup_limits_enabled, /* sysinfo_is_cgroup_limits_enabled */
+	omrsysinfo_enable_cgroup_limits, /* sysinfo_enable_cgroup_limits */
+	omrsysinfo_cgroup_get_memlimit, /* sysinfo_cgroup_get_memlimit */
```
DanHeidinga: To namespace this more consistently, and to make it easier to search the code base for cgroup related functions, can the names be changed to:
```
omrsysinfo_cgroup_is_limits_supported, /* sysinfo_is_cgroup_limits_supported */
omrsysinfo_cgroup_is_limits_enabled, /* sysinfo_is_cgroup_limits_enabled */
omrsysinfo_cgroup_enable_limits, /* sysinfo_enable_cgroup_limits */
omrsysinfo_cgroup_get_memlimit, /* sysinfo_cgroup_get_memlimit */
```

port/unix_include/omrcgroup.h
ashu-mehra: Shouldn't it be OmrCgroupEntry, i.e. start with uppercase letter?

DanHeidinga: Good point.  Grepping through the code for 'typedef struct' I see we haven't been terribly consistent but newer apis tend to be in the style of `OMRCgroupEntry` where the OMR is entirely uppercase.

ashu-mehra:
@DanHeidinga there were multiple places where truncation was a possibility.
I have added a new commit to address that.
Please review.

port/unix/omrsysinfo.c
```diff
@@ -236,6 +241,44 @@ struct  {
 #endif
 };
 
+#if defined(LINUX)
+
+#define OMR_CGROUP_V1_MOUNT_POINT "/sys/fs/cgroup"
+
+/* Currently 12 subsystems/resource controllers are defined.
+ */
+typedef enum CgroupSubsystems {
```
DanHeidinga: I missed this one in the first review. Sorry.  If this enum ever "escapes" this file, it should probably be prefixed with "OMR".

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3161,360 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+	
+	cgroup = getCgroupNameForSubsystem(portLibrary, PPG_cgroupEntryList, subsystemNames[subsystem]);
+	if (NULL == cgroup) {
+		rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_CGROUP_NAME_NOT_AVAILABLE, "cgroup name for subsystem %s is not available", subsystemNames[MEMORY]);
+		goto _end;
+	}
+
+	/* absolute path of the file to be read is: /sys/fs/cgroup/subsystemNames[subsystem]/cgroup/filenName */
+	fileNameLen = portLibrary->str_printf(portLibrary, NULL, (uint32_t)-1, "%s/%s/%s/%s", OMR_CGROUP_V1_MOUNT_POINT, subsystemNames[subsystem], cgroup, fileName);
+	if (fileNameLen > PATH_MAX) {
```
DanHeidinga: I think this needs to be `>=` as fileNameLen won't include the NUL as per the omrstr_printf doc:
```
* @return The number of characters printed not including the NUL terminator.
```

port/unix/omrsysinfo.c
```diff
@@ -3071,3 +3161,360 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+	char cgroup[PATH_MAX];
+	/* This array should be large enough to read names of all subsystems. 1024 should be enough. */
+	char subsystems[1024];
+	FILE *cgroupFile = NULL;
+	OMRCgroupEntry *cgEntryList = NULL;
+	int32_t rc = 0;
+
+	Assert_PRT_true(NULL != cgroupEntryList);
+	
+	portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
```
DanHeidinga: This is another possible truncation but I'd be tempted to address this with an assert as it's never going to happen.

port/common/omrerror.c
```diff
@@ -293,3 +293,70 @@ omrerror_set_last_error_with_message(struct OMRPortLibrary *portLibrary, int32_t
...
+	ptBuffers->portableErrorCode = portableCode;
+
+	va_start(args, format);
+	requiredSize = portLibrary->str_vprintf(portLibrary, NULL, (uint32_t)-1, format, args);
+	va_end(args);
+
+	/* Store the message, allocate a bigger buffer if required.  Keep the old buffer around
+	 * just in case memory can not be allocated
+	 */
+	requiredSize = requiredSize < J9ERROR_DEFAULT_BUFFER_SIZE ? J9ERROR_DEFAULT_BUFFER_SIZE : requiredSize;
```
DanHeidinga: Use parenthesis here: `requiredSize = (requiredSize < J9ERROR_DEFAULT_BUFFER_SIZE) ? J9ERROR_DEFAULT_BUFFER_SIZE : requiredSize;`

port/common/omrerror.c
```diff
@@ -293,3 +293,70 @@ omrerror_set_last_error_with_message(struct OMRPortLibrary *portLibrary, int32_t
...
+ *
+ * @note If per thread buffers @see omrportptb.h are not available the error code is not stored.
+ * This event would only occur if the per thread buffers could not be allocated, which is highly unlikely. In this
+ * case an application will receive a generic message/errorCode when querying for the last stored values.
+ */
+int32_t
+omrerror_set_last_error_with_message_format(struct OMRPortLibrary *portLibrary, int32_t portableCode, const char *format, ...)
+{
+	PortlibPTBuffers_t ptBuffers;
+	uint32_t requiredSize;
```
DanHeidinga: This needs a default value.

port/unix/omrsysinfo.c
ashu-mehra: Thats the case when `buf` is not NULL. When its NULL, that is the user wants to estimate the buffer size required, then it includes the NUL terminator.
```
 * @note When buf is NULL, the size of the buffer required to print to the string, including
 * the NUL terminator is returned.
```

port/unix/omrsysinfo.c
ashu-mehra: I noticed that but didn't bother to change because, as you said, "/proc/%d/cgroup" is unlikely to be more than PATH_MAX.
I am wondering what kind of assert are you suggesting. We don't know the actual length of the string - "/proc/%d/cgroup".

port/unix/omrsysinfo.c
DanHeidinga: You're right.  Good catch.

port/unix/omrsysinfo.c
DanHeidinga:
```
uintptr_t used = portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
Assert_PRT_true(used < PATH_MAX);
```

ashu-mehra: But that assert would always hold true as per the contract of `omrstr_printf()` API, even when there is truncation, isn't it?

DanHeidinga: A somewhat hacky solution would be:
```
uintptr_t used = portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
Assert_PRT_true(......("u" == cgroup[used-2]) && ("p" == cgroup[used - 1])); /* Ensure truncation didn't occur. */
```
but that's probably trying to be too clever.

Better solution is to just write the real code:
```
uintptr_t used = portLibrary->str_printf(portLibrary, NULL, 0, "/proc/%d/cgroup", pid);
Assert_PRT_true(used < PATH_MAX);
portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
```

ashu-mehra: yes, that makes sense. Thanks for the code changes.

ashu-mehra: I also added couple of more asserts on `subsystem` and `cgroup` after the call to fscanf to check for overflow.
```
                rc = fscanf(cgroupFile, "%d:%[^:]:%s", &hierId, subsystems, cgroup);
                /* Ensure we didn't overflow */
                Assert_PRT_true(strlen(subsystems) < 1024);
                Assert_PRT_true(strlen(cgroup) < PATH_MAX);
```

PR #1312, opened 2017-06-28
Add support for VirtualMachineOperandArray

charliegracie:
Not all Virtual Machines have an Operand Stack. Another common construct
is to have a fixed sized array of "registers". The new
VirtualMachineOperandArray is very similar to the VMOS but it does not
allow for growth and it has to deal with NULL entries as not all entries
may have a value.

The VirtualMachineOperandArray::Move API requires that a copy of the
IlValue stored at srcIndex be stored at dstIndex. To accomplish this I
added a new API to IlBuilder call CopyValue.

I have added tests to verify the all of the functionality.

### DIFF
--- a/compiler/ilgen/IlBuilder.hpp
+++ b/compiler/ilgen/IlBuilder.hpp
@@ -142,6 +142,8 @@ public:
    // create a new local value (temporary variable)
    TR::IlValue *NewValue(TR::IlType *dt);
 
+   TR::IlValue *Copy(TR::IlValue *value);
+
    // constants
    TR::IlValue *NullAddress();
    TR::IlValue *ConstInt8(int8_t value);
--- a/compiler/ilgen/VirtualMachineOperandArray.cpp
+++ b/compiler/ilgen/VirtualMachineOperandArray.cpp
@@ -1,0 +1,181 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2017,2017
+ *
+ *  This program and the accompanying materials are made available
+ *  under the terms of the Eclipse Public License v1.0 and
+ *  Apache License v2.0 which accompanies this distribution.
+ *
+ *      The Eclipse Public License is available at
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *******************************************************************************/
+
+#include "ilgen/VirtualMachineOperandArray.hpp"
+#include "compile/Compilation.hpp"
+#include "il/SymbolReference.hpp"
+#include "il/symbol/AutomaticSymbol.hpp"
+#include "ilgen/BytecodeBuilder.hpp"
+#include "ilgen/MethodBuilder.hpp"
+#include "ilgen/TypeDictionary.hpp"
+
+#define TraceEnabled    (TR::comp()->getOption(TR_TraceILGen))
+#define TraceIL(m, ...) {if (TraceEnabled) {traceMsg(TR::comp(), m, ##__VA_ARGS__);}}
+
+namespace OMR
+{
+VirtualMachineOperandArray::VirtualMachineOperandArray(TR::MethodBuilder *mb, int32_t numOfElements, TR::IlType *elementType, VirtualMachineRegister *arrayBaseRegister)
+   : VirtualMachineState(),
+   _mb(mb),
+   _numberOfElements(numOfElements),
+   _elementType(elementType),
+   _arrayBaseRegister(arrayBaseRegister)
+   {
+   int32_t numBytes = _numberOfElements * sizeof(TR::IlValue *);
+   _values = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
+   memset(_values, 0, numBytes);
+
+   // store current operand stack pointer base address so we can use it whenever we need
+   // to recreate the stack as the interpreter would have
+   mb->Store("OperandArray_base", arrayBaseRegister->Load(mb));
+   }
+
+VirtualMachineOperandArray::VirtualMachineOperandArray(OMR::VirtualMachineOperandArray *other)
+   : VirtualMachineState(),
+   _mb(other->_mb),
+   _numberOfElements(other->_numberOfElements),
+   _elementType(other->_elementType),
+   _arrayBaseRegister(other->_arrayBaseRegister)
+   {
+   int32_t numBytes = _numberOfElements * sizeof(TR::IlValue *);
+   _values = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
+   memcpy(_values, other->_values, numBytes);
+   }
+
+
+// commits the simulated operand array of values to the virtual machine state
+// the given builder object is where the operations to commit the state will be inserted
+// into the array which is assumed to be managed independently, most likely
+void
+VirtualMachineOperandArray::Commit(TR::IlBuilder *b)
+   {
+   TR::IlType *Element = _elementType;
+   TR::IlType *pElement = _mb->typeDictionary()->PointerTo(Element);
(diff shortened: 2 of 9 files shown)


compiler/ilgen/VirtualMachineOperandArray.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+   TR::IlValue *arrayBase = b->Load("OperandArray_base");
+
+   for (int32_t i = 0;i < _numberOfElements;i++)
+      {
+      TR::IlValue *element = _values[i];
+      if (element != NULL)
+         {
+         b->StoreAt(
+         b->       IndexAt(pElement,
```
mstoodle: not standard indentation for compiler component files. I know one day we'll change it, but for now please keep it consistent even if more spaces makes the tree structure a tad more readable.

compiler/ilgen/VirtualMachineOperandArray.cpp
```diff
@@ -0,0 +1,183 @@
...
+                   b->              ConstInt32(i)));
+       }
+   }
+
+void
+VirtualMachineOperandArray::MergeInto(OMR::VirtualMachineState *o, TR::IlBuilder *b)
+   {
+   VirtualMachineOperandArray *other = (VirtualMachineOperandArray *)o;
+   TR_ASSERT(_numberOfElements == other->_numberOfElements, "array are not same size");
+   TR::BytecodeBuilder *builder = (TR::BytecodeBuilder *)b;
```
mstoodle: i don't see `builder` being used anywhere, so please remove it.

compiler/ilgen/VirtualMachineOperandArray.hpp
```diff
@@ -0,0 +1,136 @@
...
+ * Commit() simply iterates over the simulated operand array and stores each
+ *   value onto the virtual machine's operand array (more details at definition).
+ * Reload() read the virtual machine array back into the simulated operand array
+ * MakeCopy() copies the state of the operand array
+ * MergeInto() is slightly subtle. Operations may have been already created
+ *   below the merge point, and those operations will have assumed the
+ *   expressions are stored in the TR::IlValue's for the state being merged
+ *   *to*. So the purpose of MergeInto() is to store the values of the current
+ *   state into the same variables as in the "other" state.
+ *
```
mstoodle: should add UpdateArray() into this list

jitbuilder/release/src/OperandArrayTests.hpp
```diff
@@ -0,0 +1,61 @@
...
+ ******************************************************************************/
+
+
+#ifndef OPERANDARRAYTESTS_INCL
+#define OPERANDARRAYTESTS_INCL
+
+#include "ilgen/MethodBuilder.hpp"
+
+#define ARRAYVALUEILTYPE	Int32
+#define	ARRAYVALUETYPE		int32_t
```
mstoodle: stray tabs here :)

compiler/ilgen/IlBuilder.hpp
```diff
@@ -142,6 +142,8 @@ class IlBuilder : public TR::IlInjector
    // create a new local value (temporary variable)
    TR::IlValue *NewValue(TR::IlType *dt);
 
+   TR::IlValue *CopyValue(TR::IlValue *value);
```
mstoodle: given the API, I would be tempted to call it "Copy" rather than "CopyValue" .... I know it's kinda like NewValue, but I think of it more like Load, Store, Add, Sub which also could (but by choice did not) get a Value suffix.

I could be argued out of it, but just think of all the characters you will no longer have to type :) .

jitbuilder/release/src/OperandArrayTests.cpp
```diff
@@ -0,0 +1,551 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2016, 2016
```
mstoodle: should be modified in 2017

jitbuilder/release/src/OperandArrayTests.cpp
```diff
@@ -0,0 +1,551 @@
...
+#define COMMIT(b)        (STACK(b)->Commit(b))
+#define RELOAD(b)        (STACK(b)->Reload(b))
+#define SET(b,i,v)       (STACK(b)->Set((i),(v)))
+#define GET(b,i)         (STACK(b)->Get((i)))
+#define MOVE(b, d, s)    (STACK(b)->Move(b, d, s))
+
+bool
+OperandArrayTestMethod::testArray(TR::BytecodeBuilder *builder, bool useEqual)
+   {
+   //VMOA      [null,null,null,...,null]
```
mstoodle: I like this style, makes it very clear!

jitbuilder/release/src/OperandArrayTests.hpp
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
+ *
+ * (c) Copyright IBM Corp. 2016, 2016
```
mstoodle: should be modified in 2017

jitbuilder/build/rules/common.mk
```diff
@@ -108,6 +108,9 @@ $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineRegister.hpp: $(FIXE
 $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineRegisterInStruct.hpp: $(FIXED_SRCBASE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineRegisterInStruct.hpp $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen
 	cp $< $@ || cp $< $@
 
+$(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineOperandArray.hpp: $(FIXED_SRCBASE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineOperandArray.hpp $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen
```
mstoodle: please update the copyright header in this file

charliegracie:
I believe I addressed all of the issues. 

I made a change to `VirtualMachineOperandArray::MergeInto`.  It now asserts that if the destination array has a NULL value at a given index that the source array is also NULL.

jitbuilder/release/src/Simple.cpp
```diff
@@ -89,7 +89,8 @@ SimpleMethod::buildIL()
    cout << "SimpleMethod::buildIL() running!\n";
    Return(
       Add(
-         Load("value"),
+         Copy(
```
mstoodle: Sorry, Charlie, I just realized that we probably shouldn't be mucking with Simple.cpp because the jitbuilder blog articles (and several of the presentations) reference it and describes it in detail. So I'd prefer not to just add Copy here (though it didn't register with me in my first review).

Given you have OperandArrayTests which presumably exercise Copy, I'm not sure we *really* need it in the code samples.

charliegracie:
I updated it accordingly

compiler/ilgen/VirtualMachineOperandArray.cpp
```diff
@@ -0,0 +1,181 @@
...
+void
+VirtualMachineOperandArray::MergeInto(OMR::VirtualMachineState *o, TR::IlBuilder *b)
+   {
+   VirtualMachineOperandArray *other = (VirtualMachineOperandArray *)o;
+   TR_ASSERT(_numberOfElements == other->_numberOfElements, "array are not same size");
+   for (int32_t i=0;i < _numberOfElements;i++)
+      {
+      if (NULL == other->_values[i])
+         {
+         TR_ASSERT(_values[i] == NULL, "if an element in the dest array at index is NULL it has to be NULL in the src array");
```
mstoodle: This assert establishes a "static" rule that all operands uses must be reached by an earlier definition on every incoming path. I think this is the only reasonable choice, but I wanted to call out that this assertion codifies it (but projects that leverage VirtualMachineOperandArray, anyway :) ).

PR #1323, opened 2017-07-05
Compress OSR infra after all OSR points are added

ncough:
OSR slot sharing information is intended to be
compressed after the map from instruction to slot
information has been generated. However, it is
being compressed before the slot sharing information
for snippets has been added. This can result in
a transition failing if a snippet's instruction PC
is within a range of compressed PCs and has different
slot sharing information.


andrewcraik:
I think this makes sense, but a second opinion is warranted.

ncough:
@vijaysun-omr Could you please review this?

vijaysun-omr:
Looks like a safe change to me to delay the compression operation. Merging.

PR #1324, opened 2017-07-05
Fix live monitor meta data to detect invalid cases

cathyzhyi:
In original lmmd code, when propagating monitor state from one block
to its sucessors the blocks that are already visited are skipped
and it's assumed that the monitor state for one block should be the
same no matter which predecessor it's propagated from.
This code detect the cases where the monitor state for one block can vary
when propagating from different predecessor blocks. In those cases,
lmmd should resort to the most conservative solution.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -1563,7 +1563,7 @@ int32_t TR_SetMonitorStateOnBlockEntry::addSuccessors(TR::CFGNode * cfgNode,
       if (walkOnlyExceptionSuccs && !succBlock->isCatchBlock())
          continue;
 
-      if (succBlock->getEntry() && succBlock->getVisitCount() != _visitCount)
+      if (succBlock->getEntry())
          {
          bool addInfo = true;
          if (monitorType == MonitorEnter)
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
Need a review from @andrewcraik for this.

andrewcraik:
I think this change makes sense and is good to go, but an opinion from @vijaysun-omr would also be appreciated since I think you have more experience in here.

vijaysun-omr:
The changes look okay to me.

One question I have is around testing this change. If we incorrectly determined that the live monitor stacks were inconsistent on two control flow paths leading into a block, then we would silently initialize autos in the method without any other overt sign that something went wrong. Is it possible to test this in a way that prints a trace message when inconsistent state is detected so that you can see how often inconsistent state is detected and assess if it's valid ?

Another question is around the scenario when you expect to encounter inconsistent state, e.g. what was the situation that made you implement this fix ?

cathyzhyi:
the static debug counter is added and running it with a fairly large application shows such unbalanced cases are very rare. 
The bytecodes lead to this fix are manually manipulated, resulting in a jump from inside the synchronized region directly to the outside without executing the monitor exit.

vijaysun-omr:
Thanks. I will merge this now.

PR #1325, opened 2017-07-05
WIP: Replace SparseBitVector with dense in dominators

ncough:
The dominators analysis used a SparseBitVector
to represent semidominant nodes. Due to the
heap allocation overhead of the CS2 SparseBitVector
and the density of CFG node numbers, this has
been replaced with a dense bit vector.

### DIFF
--- a/compiler/optimizer/Dominators.hpp
+++ b/compiler/optimizer/Dominators.hpp
@@ -63,7 +63,7 @@ class TR_Dominators
 
    protected:
 
-   typedef TR::SparseBitVector SparseBitVector;
+   typedef TR::BitVector SparseBitVector;
 
    private:
 


vijaysun-omr:
Is the "overhead" that you are concerned about a footprint OR compile time ?

ncough:
This change is concerned with the impact on compile time and heap usage.

Due to the use of DominatorAnalysis in OSRLiveRangeAnalysis and SparseBitVector's reliance on the heap allocations when setting bits, this data structure was accounting for 10% of the total heap usage in testing. As this usage cannot be recovered, compilations would later reach their maximum memory limit and fail.

BitVectors instead have cheaper allocations, in terms of compile time, and they can placed in a stack region, allowing the memory to be recovered after OSRLiveRangeAnalysis.

PR #1326, opened 2017-07-05
Add dead stores to OSR code block

ncough:
To reduce OSR overhead, the dead stores for an entire call
site can be placed in its OSR code block, rather than
every OSR induce block. This is done during IlGen for the
callee, as OSR live range analysis would have already been
performed for the caller.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1174,6 +1174,12 @@ OMR::ResolvedMethodSymbol::genOSRHelperCall(int32_t currentInlinedSiteIndex, TR:
       }
    else
       {
+      // Add the dead stores for this call site
+      TR_OSRMethodData* callerOSRData = self()->comp()->getOSRCompilationData()->findCallerOSRMethodData(osrMethodData);
+      TR_InlinedCallSite &callSiteInfo = self()->comp()->getInlinedCallSite(currentInlinedSiteIndex);
+      callerOSRData->getMethodSymbol()->insertStoresForDeadStackSlots(self()->comp(), callSiteInfo._byteCodeInfo,
+         OSRCodeBlock->getExit(), false);
+
       //create an appropritely-typed return node and append it to the OSR code block and add
       //the corresponding cfg edge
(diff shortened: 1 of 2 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1799,80 +1805,66 @@ OMR::ResolvedMethodSymbol::getOSRTransitionTreeTop(TR::TreeTop *tt)
    }
 
 void
-OMR::ResolvedMethodSymbol::insertStoresForDeadStackSlotsBeforeInducingOSR(TR::Compilation *comp, int32_t inlinedSiteIndex, TR_ByteCodeInfo &byteCodeInfo, TR::TreeTop *induceOSRTree, TR::ResolvedMethodSymbol *callSymbolForDeadSlots)
+OMR::ResolvedMethodSymbol::insertStoresForDeadStackSlots(TR::Compilation *comp, TR_ByteCodeInfo &byteCodeInfo, TR::TreeTop *insertTree, bool ignoreStashedArgs)
```
andrewcraik: Can you discuss the need for this new bool parameters and what is signifies? A the very least this needs some doxygen to describe how it is used.

vijaysun-omr:
I assume this only adds the caller's dead slots zeroing to a callee's OSR code block. I feel a more visual explanation of the "before" and "after" state would be beneficial in making it clear exactly what the transformation is. Perhaps sketch out using an example call chain A->B->C ?

ncough:
An example call chain has been added, with a simplified sketch of the OSR blocks.

vijaysun-omr:
Thanks for the diagram as a comment in the code. It helps clarify things a lot. I assume the purpose of the change is to reduce the code bloat from having the stores for the callers in each and every OSR induce block. This way the caller stores are more concentrated and can be shared for each OSR induce point in the callee.

ncough:
Yes, you are correct. This can significantly reduce the node count if there are several of OSR transitions in a callee and the caller has a large number of dead symrefs. These can be killed once, rather than being repeated at each OSR induce block.

vijaysun-omr:
Merging.

PR #1328, opened 2017-07-05
Add breakpoint guard to support inlining in FSD

cathyzhyi:
This changeset is part of the code to add breakpoint guards to support
inlining in FSD mode.The breakpoint guard reads a bit from
a field of the callee method which is set when any breakpoints
are set for that particular method.

This changeset create the actual breakpoint guard and add it to each
callee under FSD mode.

### DIFF
--- a/compiler/compile/VirtualGuard.cpp
+++ b/compiler/compile/VirtualGuard.cpp
@@ -213,6 +213,71 @@ TR_VirtualGuard::createMethodGuard
    return createMethodGuardWithReceiver (kind, comp, calleeIndex, callNode, destination, calleeSymbol, thisClass, callNode->getSecondChild());
    }
 
+/*
+ * generating the actual test node for breakpoint guard
+ */
+TR::Node*
+TR_VirtualGuard::createBreakpointGuardNode
+(TR::Compilation * comp, int16_t calleeIndex,
+ TR::Node* callNode, TR::TreeTop * destination, TR::ResolvedMethodSymbol * calleeSymbol)
+   {
+#ifdef J9_PROJECT_SPECIFIC
+/*
+ * Shape of a breakpoint guard:
+ * iflcmpeq/ificmpeq
+ *    land/iand
+ *       lloadi/iloadi <ConstantPool shadow>
+ *          aconst J9Method
+ *       isBreakpointedBit 
+ *    isBreakpointedBit 
+ */
+   bool is64Bit = TR::Compiler->target.is64Bit();
+   TR::SymbolReferenceTable * symRefTab = comp->getSymRefTab();
+   TR::SymbolReference * fieldSymRef = symRefTab->findOrCreateJ9MethodConstantPoolFieldSymbolRef(offsetof(struct J9Method, constantPool));
+   TR::Node * aconstNode = TR::Node::aconst(callNode, (uintptrj_t)calleeSymbol->getResolvedMethod()->getPersistentIdentifier());
+   TR::Node * constantPool = TR::Node::createWithSymRef(is64Bit? TR::lloadi: TR::iloadi, 1, 1, aconstNode, fieldSymRef);
+   aconstNode->setIsMethodPointerConstant(true);
+   aconstNode->setInlinedSiteIndex(calleeIndex);
+   aconstNode->setByteCodeIndex(0);
+   TR::Node * flagBit = NULL;
+   TR::Node *guard = NULL;
+   if (TR::Compiler->target.is64Bit())
+      {
+      flagBit = TR::Node::create(callNode, TR::lconst, 0, 0);
+      flagBit->setLongInt(comp->fej9()->offsetOfMethodIsBreakpointedBit());
+
+      }
+   else
+      {
+      flagBit = TR::Node::create(callNode, TR::iconst, 0, comp->fej9()->offsetOfMethodIsBreakpointedBit());
+      }
+
+   guard =  TR::Node::createif(is64Bit? TR::iflcmpeq: TR::ificmpeq,
+            TR::Node::create(is64Bit? TR::land: TR::iand, 2, constantPool, flagBit),
+            flagBit,
+            destination);
+   return guard;
+#else
+   TR_ASSERT(false, "need project specific implementation to generate the breakpoint guard node");
+#endif
+   }
+ 
+/*
+ * a breakpoint guard jumps to the slow path if a breakpoint is set for the inlined callee
+ */
+TR::Node*
+TR_VirtualGuard::createBreakpointGuard
+(TR::Compilation * comp, int16_t calleeIndex,
+ TR::Node* callNode, TR::TreeTop * destination, TR::ResolvedMethodSymbol * calleeSymbol)
+   {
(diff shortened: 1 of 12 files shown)


compiler/codegen/FrontEnd.hpp
```diff
@@ -135,11 +135,14 @@ class TR_FrontEnd : public TR_Uncopyable
 
    virtual TR_OpaqueMethodBlock * getMethodFromName(char * className, char *methodName, char *signature, TR_OpaqueMethodBlock *callingMethod=0);
    virtual uint32_t offsetOfIsOverriddenBit();
+   virtual uint32_t offsetOfIsBreakpointedBit();
 
    // Needs VMThread
    virtual bool isMethodEnterTracingEnabled(TR_OpaqueMethodBlock *method);
    virtual bool isMethodExitTracingEnabled(TR_OpaqueMethodBlock *method);
 
+   virtual bool isSelectiveMethodEnterExitEnabled();
```
fjeremic: I would say any new APIs added to the common frontend class ought to be Doxygen documented for precise specification so that any downstream project wishing to implement their own frontend knows exactly how to implement these APIs.

vijaysun-omr:
@andrewcraik should review this as well.

compiler/codegen/FrontEnd.hpp
andrewcraik: Not only that, this needs a better name more applicable to the OMR project and languages in general.

cathyzhyi: Thanks for the suggestions. Added comments and moved the API to project specific place.

compiler/compile/VirtualGuard.cpp
```diff
@@ -213,6 +213,60 @@ TR_VirtualGuard::createMethodGuard
...
+      guard =  TR::Node::createif(TR::iflcmpeq,
+                        TR::Node::create(TR::land, 2, constantPool, flagBit),
+               flagBit,
+               destination);
+      }
+   else
+      {
+      TR::Node * flagBit = TR::Node::create(callNode, TR::iconst, 0, comp->fej9()->offsetOfMethodIsBreakpointedBit());
+
+      guard =  TR::Node::createif(TR::ificmpne,
```
vijaysun-omr: How come the branch opcode is ificmpne in the case of 32-bit whereas it is iflcmpeq in the 64-bit case ? Should'nt the "sense" of the branch opcode be the same given all the other children are the same ?

compiler/il/OMRNode.cpp
```diff
@@ -7529,14 +7529,33 @@ OMR::Node::setIsOSRGuard()
       _flags.set(osrGuard);
    }
 
+void
+OMR::Node::setIsBreakpointGuard()
+   {
+   TR::Compilation *c = TR::comp();
+   TR_ASSERT(self()->getOpCode().isIf(), "assertion failure");
+   if (performNodeTransformation1(c, "O^O NODE FLAGS: Setting fsdGuard flag on node %p\n", self()))
+      _flags.set(fsdGuard);
```
vijaysun-omr: Minor question, is there a reason why the flag bit is called "fsdGuard" whereas the routine calls it a "breakpoint guard" ? i.e. why not make it consistent ?

compiler/compile/VirtualGuard.cpp
```diff
@@ -213,6 +213,60 @@ TR_VirtualGuard::createMethodGuard
...
+   aconstNode->setInlinedSiteIndex(calleeIndex);
+   aconstNode->setByteCodeIndex(0);
+
+   TR::Node *guard = NULL;
+   if (TR::Compiler->target.is64Bit())
+      {
+      TR::Node * flagBit = TR::Node::create(callNode, TR::lconst, 0, 0);
+      flagBit->setLongInt(comp->fej9()->offsetOfMethodIsBreakpointedBit());
+
+      guard =  TR::Node::createif(TR::iflcmpeq,
```
andrewcraik: Can you make the formatting in this if block match the else so that it is more obvious they are equivalent? An alternative to save code duplication might be just to create two opcode variables to hold the if and and and and then generate the trees which could prevent the 32 and 64 cases from diverging over time...

compiler/compile/VirtualGuard.cpp
```diff
@@ -213,6 +213,60 @@ TR_VirtualGuard::createMethodGuard
...
+ *          aconst J9Method
+ *       isBreakpointedBit 
+ *    isBreakpointedBit 
+ */
+TR::Node*
+TR_VirtualGuard::createBreakpointGuard
+(TR::Compilation * comp, int16_t calleeIndex,
+ TR::Node* callNode, TR::TreeTop * destination, TR::ResolvedMethodSymbol * calleeSymbol)
+   {
+#ifdef J9_PROJECT_SPECIFIC
```
andrewcraik: The only part of this that is J9 Project specific is the actual test. If we have a convention about the test and have a method that generates the test trees the rest of the code is OMR common and doesn't need to be hidden - the Java specific part could then move out of OMR to leave a common language framework that just requires implementing languages to generate the trees to do the test for the breakpoint. If that is a huge amount of work then could we at least isolate the J9 Project specific part in a function that describes what it is generating and for non Java languages asserts or something similar?

compiler/compile/VirtualGuard.hpp
```diff
@@ -64,7 +65,8 @@ enum TR_VirtualGuardTestType
    TR_VftTest,
    TR_MethodTest,
    TR_NonoverriddenTest,
-   TR_RubyInlineTest
+   TR_RubyInlineTest,
+   TR_FSDTest
```
andrewcraik: Would be good to document what this is somewhere (either above the enum or at the enum some how.

compiler/compile/VirtualGuard.cpp
cathyzhyi: the field loaded from the method data structure is 64bits on 64-bits platform but 32bits on 32-bits and therefore the compare opcode is different.

compiler/compile/VirtualGuard.cpp
cathyzhyi: Thanks for the suggestion. Modified accordingly.

compiler/compile/VirtualGuard.cpp
andrewcraik: we agree the l needs to change to an i but one case is iflcmpeq and in the other it is ificmpne - why do you flip the sense of the test - that is not what you want  don't think they should both be eq or ne.

compiler/compile/VirtualGuard.hpp
cathyzhyi: thanks for the suggestion. added documentation accordingly.

compiler/compile/VirtualGuard.cpp
cathyzhyi: moved the code to generate the actual test node to a separate function.

vijaysun-omr:
Both the issues I raised seem to have been addressed by @cathyzhyi; she fixed the if compare opcode to be the same "sense" on 64 and 32 bit and renamed the flag bit. If @andrewcraik is okay with this I can consider for merging.

andrewcraik:
LGTM

vijaysun-omr:
Merging.

PR #1330, opened 2017-07-06
Remove allocator pointer in CfgNode

ncough:
To simplify memory management, CfgNodes should
not have a pointer to the allocator. Instead,
the allocator can be accessed from the CFG
or compilation object. Furthermore, a region
can be specified when creating a CfgNode.

### DIFF
--- a/compiler/env/TypedAllocator.hpp
+++ b/compiler/env/TypedAllocator.hpp
@@ -63,6 +63,7 @@ public:
       return !(left == right);
       }
 
+   untyped_allocator backingAllocator() { return _backingAllocator; }
 protected:
    explicit typed_allocator( untyped_allocator backingAllocator ) throw() : _backingAllocator(backingAllocator) {}
 
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -125,7 +125,7 @@ OMR::Block::Block(TR::TreeTop *entry, TR::TreeTop *exit, TR_Memory * m) :
 
 /// Copy constructor
 OMR::Block::Block(TR::Block &other, TR::TreeTop *entry, TR::TreeTop *exit) :
-   TR::CFGNode(other.trMemory()),
+   TR::CFGNode(other.getSuccessors().get_allocator().backingAllocator()),
    _pEntry(entry),
    _pExit(exit),
    _pStructureOf(NULL),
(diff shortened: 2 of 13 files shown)


vijaysun-omr:
I would appreciate a review from @lmaisons on this.

vijaysun-omr:
Thanks for the reviews you did today @lmaisons Wanted to ask if you could also look at this PR.

compiler/il/OMRBlock.cpp
```diff
@@ -427,8 +428,8 @@ OMR::Block::breakFallThrough(TR::Compilation *comp, TR::Block *faller, TR::Block
          comp->getFlowGraph()->addNode(gotoBlock, faller->getCommonParentStructureIfExists(fallee, comp->getFlowGraph()));
       else
          comp->getFlowGraph()->addNode(gotoBlock);
-      comp->getFlowGraph()->addEdge(TR::CFGEdge::createEdge(faller,  gotoBlock, self()->trMemory()));
-      comp->getFlowGraph()->addEdge(TR::CFGEdge::createEdge(gotoBlock,  fallee, self()->trMemory()));
+      comp->getFlowGraph()->addEdge(TR::CFGEdge::createEdge(faller,  gotoBlock, comp->getFlowGraph()->trMemory()));
+      comp->getFlowGraph()->addEdge(TR::CFGEdge::createEdge(gotoBlock,  fallee, comp->getFlowGraph()->trMemory()));
```
lmaisons: Given that the compilation object has direct access to trMemory, are we using the flow graph's trMemory for reasons of conceptual clarity?  Or have we started walking down a path where there is more than a single trMemory object per compilation?

compiler/env/TypedAllocator.hpp
```diff
@@ -63,6 +63,7 @@ class typed_allocator<void, Alloc>
       return !(left == right);
       }
 
+   untyped_allocator backingAllocator() { return _backingAllocator; }
```
lmaisons: This sits strangely to me.  I wonder if this shouldn't instead be implemented as a conversion operator.

compiler/il/OMRBlock.cpp
```diff
@@ -1545,9 +1546,9 @@ OMR::Block::setInstructionBoundaries(uint32_t startPC, uint32_t endPC)
    }
 
 void
-OMR::Block::addExceptionRangeForSnippet(uint32_t startPC, uint32_t endPC)
+OMR::Block::addExceptionRangeForSnippet(OMR::Compilation *comp, uint32_t startPC, uint32_t endPC)
```
lmaisons: If this is only using the compilation for access to trMemory, it should probably just take one of those directly instead.

ncough:
Similar changes to those included here will instead be implemented in a future pull request.

PR #1331, opened 2017-07-06
Version HCR guards out of loops

ncough:
If there are no canGCandReturn or canGCandExcept
nodes in a loop, any HCR guards within it
can be versioned out. This is similar to the OSR
guard versioning, however, an individual HCR
guard must be created for each that is removed
due to the runtime assumptions varying between
them.


vijaysun-omr:
Looks valid to me along similar lines to the OSR guard versioning merged a few weeks back.

PR #1332, opened 2017-07-07
z/TPF Port updates for OMR Part III

jjohnst-us:
This commit contains further updates for z/TPF OMR support.
The commit contains both build related updates and single 
source related updates.

### DIFF
--- a/include_core/AtomicSupport.hpp
+++ b/include_core/AtomicSupport.hpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 1991, 2016
+ * (c) Copyright IBM Corp. 1991, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -14,6 +14,7 @@
  *
  * Contributors:
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *    James Johnston (IBM Corp.)   - initial z/TPF Port Updates
  *******************************************************************************/
 
 #if !defined(ATOMIC_SUPPORT_HPP_)
@@ -24,6 +25,10 @@
 #if defined(TYPESTUBS_H)
 #define ATOMIC_SUPPORT_STUB
 #endif /* defined(TYPESTUBS_H) */
+
+#if defined(OMRZTPF)
+#include <tpf/cmpswp.h>
+#endif
 
 #include <stdlib.h>
 #if !defined(ATOMIC_SUPPORT_STUB)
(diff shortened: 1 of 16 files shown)


jjohnst-us:
@charliegracie 
Internal IBM build_id=355437 ran clean accept for one unrelated failure.

include_core/AtomicSupport.hpp
```diff
@@ -14,6 +14,7 @@
  *
  * Contributors:
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *    James Johnston (IBM Corp.)   - initial z/TPF Port Updates
```
charliegracie: @mstoodle I think adding the contributor is ok but just checking with you as well.

port/unix/protect_helpers.c
```diff
@@ -45,6 +46,12 @@ struct {
 intptr_t
 protect_memory(struct OMRPortLibrary *portLibrary, void *address, uintptr_t length, uintptr_t flags)
 {
+#if defined(OMRZTPF)
+	/* Return success, zTPF does not support mprotect. */
+	IDATA rc = 0;
+
+	return rc;
```
charliegracie: @pdbain-ibm Is the right answer to return success or failure when the feature is not supported?

tools/tracegen/StringUtils.cpp
```diff
@@ -44,8 +45,13 @@ StringUtils::startsWithUpperLower(const char *text, const char *prefix)
 const char *
 StringUtils::containsUpperLower(const char *text, const char *toFind)
 {
+#if !defined(OMRZTPF)
 	while ('\0' != *text) {
 		if (tolower(*text) == tolower(*toFind)) {
+#else /* !defined(OMRZTPF) */
+	while ('\0' != (unsigned char)*text) {
```
charliegracie: zPTF compilers require a cast to compare a char to the dereference of a char *?

jjohnst-us: I recompiled the code undoing the z/TPF mod and compiled it cleanly.  Not sure why we originally had that mod (maybe at the time a different compiler?).  Anways, because it was the only mod we had in StringUtils.cpp  I removed it from the commit.

port/unix/protect_helpers.c
jjohnst-us: Our intention was to no-op the operation, we didn't want to fail JVM operations as I believe a while ago architects deemed it was okay for now if we didn't support it.

jjohnst-us:
@charliegracie let us know if you need anything else.

port/unix/omrsyslog.c
```diff
@@ -153,6 +154,11 @@ get_conv_txt_num_bytes(struct OMRPortLibrary *portLibrary, const char *message)
 static void
 syslog_write(struct OMRPortLibrary *portLibrary, const char *buf, int priority)
 {
+#if defined(OMRZTPF)
+	sprintf(buf, "%s", buf);
```
keithc-ca: This looks suspicious: `buf` should be considered read-only and shouldn't be the target of `sprintf` (even if it only writes what's already there).

jjohnst-us: @keithc-ca 
Thanks, I will put it on my radar.  
I just undid the change in our vm28 build environment and it compiles cleanly.  
Not sure why we had that update to begin with.  I need discuss it with Pete L. to see if we can remember why we made this update.   If we can't remember, I'll undo it in our next pull request coming shortly with updates we need for omrsysinfo.  This update only affects the z/TPF platform and shouldn't have affected anyone else.

PR #1334, opened 2017-07-10
WIP: Modify BitVector to use regions for allocations

ncough:
This change modifies BitVector to use regions
for allocations, rather than being restricted
to heap memory or the current stack. The original
API must be used for persistent memory, as it
cannot be expressed as a region.

### DIFF
--- a/compiler/infra/BitVector.hpp
+++ b/compiler/infra/BitVector.hpp
@@ -106,6 +106,7 @@ class TR_SingleBitContainer
 
    TR_SingleBitContainer() : _value(false) {}
    TR_SingleBitContainer(int64_t initBits, TR_Memory * m, TR_AllocationKind allocKind = heapAlloc) : _value(false) { }
+   TR_SingleBitContainer(int64_t initBits, TR::Region &region) : _value(false) { }
    int32_t get(int32_t n) { TR_ASSERT(n == 0, "SingleBitContainers only contain one bit\n"); return _value; }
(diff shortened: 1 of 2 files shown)


compiler/infra/BitVector.hpp
```diff
@@ -156,25 +156,71 @@ class TR_BitVector
...
          {
-         _chunks = (chunk_t*)m->allocateMemory(_numChunks*sizeof(chunk_t), _allocationKind, ot);
+         if (_region)
+            {
+            _chunks = (chunk_t*)_region->allocate(_numChunks*sizeof(chunk_t));
+            }
+         else
+            {
+            TR_ASSERT(allocKind == persistentAlloc, "Should not allocate through trMemory except for persistent memory");
+            _chunks = (chunk_t*)m->allocateMemory(_numChunks*sizeof(chunk_t), persistentAlloc, ot);
```
lmaisons: My preference would be for this to use `jitPersistentAlloc` for consistency with the rest of the choices

lmaisons:
Other that the nit, This looks OK to me.

lmaisons:
I hate to ask this but... Using TR_BitVector with stack memory today can result in the underlying allocations coming from a different stack area than the one active when the bit vector is constructed.  While that path is rife with hazards, it would not surprise me to learn that some aspect of the compiler depends on this to reduce memory costs.  Have we done any performance analysis to quantify this risk?

PR #1335, opened 2017-07-10
Tracegen tool fixes

LinaSerry:
Tracepoints/TraceAssert normally have a default parameter which is
the current thread.  There is an option NoEnv to remove the default
current thread parameter.  The template for generating the TraceAssert
was not taking the envParam into consideration.  This means that
TraceAsserts which did not use NoEnv were generating incorrect data.
The fix is to take it into consideration and generate the correct code
for these asserts.

Header file creation now throws an error if the Trace point is not
properly formatted rather than being generated and then later causing
a compilation error.


tools/tracegen/TDFParser.cpp
```diff
@@ -229,6 +229,9 @@ TDFParser::processTracePointDetail(const char *line, J9TDFTracepoint *tp, const
 			tok = strtok(NULL, " \t");
 		} else if (StringUtils::startsWithUpperLower(tok, (UT_ASSERT_TYPE == tp->type ? "Assert=" : "Template="))) {
 			tp->format = getTemplate(line, (UT_ASSERT_TYPE == tp->type ? "Assert=" : "Template="));
+			if (NULL == tp->format){
```
charliegracie: Are there any malloc'd structures that we need to free before goto failed?

LinaSerry: no, this should be okay.

PR #1336, opened 2017-07-10
Remove special OSR Induce API for guards

ncough:
OSR is used for certain guarded callees. In the past
a specialized function was used to generate and insert
the trees for these. With simplification of OSR
infrastructure, this difference is no longer necessary
and the common function can be used.


andrewcraik:
LGTM - a logical clean up IMO.

vijaysun-omr:
Looks like a safe cleanup. Merging.

PR #1337, opened 2017-07-10
RAS capabilities for compiler scratch memory

lmaisons:
This commit provides two new capabilities for interacting with scratch
memory: 1) A RegionProfiler class, and 2) A DebugSegmentProvider class.

1) The RegionProfiler class is intended for tracking changes in Region
memory consumption between the construction and destruction of the
RegionProfiler.

2) The DebugSegmentProvider class provides an aggressive failure mode
used to detect invalid reads and writes to expired memory regions. This
is achieved by changing the memory mapping of a memory segment when it
is relinquished by a Region back to the SegmentProvider, freeing the
physical pages and marking the mapping as having an empty permission
set.


lmaisons:
@andrewcraik FYI

lmaisons:
Rebased to latest master

compiler/env/RegionProfiler.hpp
```diff
@@ -0,0 +1,79 @@
...
+#define OMR_REGION_PROFILER_HPP
+
+#pragma once
+
+#include "env/Region.hpp"
+#include "env/SegmentProvider.hpp"
+#include "compile/Compilation.hpp"
+
+namespace TR {
+
```
0xdaryl: I think some level of Doxygen commentary is appropriate here.  Unstated, but implied by this feature as well is that its only appropriate if you have a `TR::Compilation` in hand.

compiler/env/DebugSegmentProvider.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <stddef.h>
+#include <set>
+#include "env/TypedAllocator.hpp"
+#include "infra/ReferenceWrapper.hpp"
+#include "env/SegmentAllocator.hpp"
+#include "env/RawAllocator.hpp"
+
+namespace TR {
+
```
0xdaryl: Please provide a Doxygen comment for this new class.

lmaisons:
Updated based on review comments & rebased to master

PR #1338, opened 2017-07-10
CMake: Change glue to be interface library

dnakamura:
Instead of taking cache variable OMR_GLUE which points to directory
with glue, now take OMR_GLUE_TARGET which contains name of glue target

Old system was limited to having glue reside in a subdirectory of
omr, which is not particularily useful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,13 +32,13 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules" ${CMAKE_MODULE_PATH})
 
 
 ###
-### Getting the glue directory
+### Getting the glue target
 ###
-
-if (NOT DEFINED OMR_GLUE)
-    set(OMR_GLUE "./example/glue" CACHE PATH "The glue directory")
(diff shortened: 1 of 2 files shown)


charliegracie:
I am not sure this change is "correct" the glue currently is just not for Java.  There is thread and RAS glue as well.  In the future the JIT and other components may also have glue so it should not be included by getting the GC.  I may be missing something as CMake is still new to me

dnakamura:
@charliegracie  Have update and the glue is no longer built as part of the gc

dnakamura:
@charliegracie  Should be updated now. glue is no longer built into the gc

CMakeLists.txt
```diff
@@ -84,6 +84,17 @@ endif()
 configure_file(./omrcfg.CMakeTemplate.h omrcfg.h)
 configure_file(./omrversionstrings.CMakeTemplate.h omrversionstrings.h)
 
+add_library(omrglue STATIC)
+# lInk as private since we dont want the interface sources to propogate
```
charliegracie: Can you correct the capitalization?

PR #1340, opened 2017-07-11
Remove enableAESInHardwareTransformations()

0xdaryl:
This codegen query is not used in OMR.  It will be moved to
its downstream project.


vijaysun-omr:
Simple cleanup. Merging.

PR #1341, opened 2017-07-11
Add a git commit message hook script

xliang6:
1. Add a commit message script that can be installed in a local repository
    under .git/hooks to validate commit messages against OMR contributing
    guidelines.
2. Add installation instructions for the script in the guidelines.

[ci skip] git hook script only

### DIFF
--- a/scripts/commit-msg
+++ b/scripts/commit-msg
@@ -1,0 +1,108 @@
+#!/usr/bin/env ruby
+###############################################################################
+##
+## (c) Copyright IBM Corp. 2017
+##
+##  This program and the accompanying materials are made available
+##  under the terms of the Eclipse Public License v1.0 and
+##  Apache License v2.0 which accompanies this distribution.
+##
+##      The Eclipse Public License is available at
+##      http://www.eclipse.org/legal/epl-v10.html
+##
+##      The Apache License v2.0 is available at
+##      http://www.opensource.org/licenses/apache2.0.php
+##
+## Contributors:
+##    Multiple authors (IBM Corp.) - initial implementation and documentation
+################################################################################
+
+# This script is intended to be used under omr_repo_dir/.git/hooks to check
+# each git commit message against OMR contributing guidelines. The script
+# needs to be executable and it will be invoked when a 'git commit' command 
+# is issued. OMR contributing guidelines can be found at:
+# https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines
+
+message_file = ARGV[0]
+message = File.read(message_file)
+
+# remove git generated messages
+commit = ""
+message.each_line do |line|
+  if (line.match(/^(?!#).*$/))
+    commit.concat(line)
+  end
+end
+
+title, body = commit.split("\n", 2)
+
+# check commit title length
+title_len = title.length
+#puts "Title is #{title}, length is #{title_len}"
+
+if (title_len > 0 && title_len < 50)
+  # shortcut to skip commit message checking
+  if (title.match(/^skip.*/) || title.match(/^fixup.*/))
+    puts "PASSED: Skipping commit checks!\n"
+    exit 0
+  end
+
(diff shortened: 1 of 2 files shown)


fjeremic:
Does this automatically integrate with GitHub?

xliang6:
@fjeremic This is a client-side hook script, i.e. it needs to be installed in a local repository and then it is executed when a `git commit` command is issued. Installation is very easy: Copy the script into .git/hooks as commit-msg and make it executable.

fjeremic:
> Installation is very easy: Copy the script into .git/hooks as commit-msg and make it executable.

@xliang6 could you as part of this PR include a commit in the contribution guidelines that this exists. Presumably when someone wishes to contribute to OMR they will read through Contributing.md and unless there is some documentation there that this hook exists it will be lost in the sands of time.

xliang6:
@fjeremic +1 on adding instructions to CONTRIBUTING.md.

scripts/commit-msg
```diff
@@ -0,0 +1,87 @@
...
+  state = "failure"
+  description = "Commit title should be less than 50 characters: actual #{title_len}"
+end
+
+# check commit message body length
+if (state == "success")
+  body.each_line.with_index() do |line, j|
+    line.chomp!
+#    puts "Line length is #{line.length}"
+    if (line.length >= 0 && line.length <= 72)
```
0xdaryl: The 72 character limit is really only a recommendation in the CG.  There are circumstances where it may be difficult or inconvenient to ensure a 72 char limit (e.g., pasting source code, or an entry from a compiler log).  Hence, exceeding 72 chars should illicit an informational warning that some lines have exceeded 72 chars and the message body should be checked manually.

scripts/commit-msg
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env ruby
+
+# This script is intended to be used under repo_dir/.git/hooks to check
+# each git commit message against OMR contributing guidelines. The script
```
0xdaryl: Perhaps put the path to the CG file (CONTRIBUTING.md) in the comment.

scripts/commit-msg
```diff
@@ -0,0 +1,87 @@
...
+    state = "failure"
+    description = "Please sign off your commit in the footer."
+  elsif (lines.count < 3)
+    state = "failure"
+    description = "Please provide a commit message body explaining your work."
+  else
+    state = "success"
+    description = "This commit satisfies all commit guidelines."
+  end
+end
```
0xdaryl: My preference would be for this script to output as little as possible if the commit actuall conforms to the guidelines (i.e., only output an "All OK!" on overall compliance rather than at each stage).

scripts/commit-msg
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+end
+
+# check signed-off line
+if (state == "success")
+  lines = body.split(/\n+/)
+  if (lines.empty?)
+    state = "failure"
+    description = "Please provide a commit message body and sign off your commit."
```
0xdaryl: Lame comment...  The messages in this part of the script have become more polite.  Can you make the messages in this script consistent:  either all "terse" or all "polite", eh?

scripts/commit-msg
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env ruby
```
0xdaryl: Please provide a copyright header.

scripts/commit-msg
xliang6: Okay, I will print the information and still return success for such commits.

xliang6:
@0xdaryl Thanks for the comments. I have made changes as you requested.

FYI, I also added a shortcut to bypass any checking if user puts "skip" or "fixup" at the beginning of a commit message title. I personally found it useful as I sometimes intend to discard a commit message when I do `git rebase -i` with the fixup option.

PR #1342, opened 2017-07-11
Fix spin counts

babsingh:
If spinning is disabled, we should try to acquire lock once and
immediately exit. When spinning is disabled, operations such as nops,
yieldCPU and omrthread_yield should be avoided since they utilize CPU.

JLM counts should not be updated if spinning is disabled. For disabled
spinning, spin counts are 0 since we try acquiring the lock once and
then exit.


babsingh:
No significant improvement spotted on x86. Also, regression seen on zOS. Thus, closing this pull request.

PR #1343, opened 2017-07-11
Fix casting error on windows

LinaSerry:
casting strlen in forloop to be an int to fix
possible loss of data error on windows plateform

### DIFF
--- a/tools/tracegen/TDFParser.cpp
+++ b/tools/tracegen/TDFParser.cpp
@@ -340,7 +340,7 @@ TDFParser::getTemplate(const char *line, const char *key)
 	char *format = NULL;
 	size_t len = 0;
 	/*
-	 * Number of quotes in the string. Must be 2 when the string is fully parsed to ensure there is an opening and closing quote.
+	 * Number of quotes in the string. Must be even when the string is fully parsed to ensure there is an opening and closing quote.
 	 */
 	int count = 0;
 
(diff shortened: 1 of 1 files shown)


tools/tracegen/TDFParser.cpp
```diff
@@ -365,7 +365,7 @@ TDFParser::getTemplate(const char *line, const char *key)
 	}
 
 	len = (pos - vpos);
-	for (int i = strlen(line); i >= 0; i--){
+	for (size_t i = strlen(line); i >= 0; i--){
```
charliegracie: Can you put a space after the `)` before the `{`.  Thanks

tools/tracegen/TDFParser.cpp
```diff
@@ -365,7 +365,7 @@ TDFParser::getTemplate(const char *line, const char *key)
...
 		if (line[i] == ' '){
```
charliegracie: Since this is just doing a continue can we not remove it and change the next `else if` to an `if`?

tools/tracegen/TDFParser.cpp
```diff
@@ -365,7 +365,7 @@ TDFParser::getTemplate(const char *line, const char *key)
...
 			continue;
 		}
```
charliegracie: Can you put a space after the `)` before the `{`. Thanks

0xdaryl:
Also, at the very least please report what casting error you are fixing with this PR.  Your commit message is light on details.

tools/tracegen/TDFParser.cpp
```diff
@@ -365,15 +365,15 @@ TDFParser::getTemplate(const char *line, const char *key)
 	}
 
 	len = (pos - vpos);
-	for (int i = strlen(line); i >= 0; i--){
-		if (line[i] == ' '){
-			continue;
-		}
-		else if (line[i] == '\"'){
+	for (int i = (int)(strlen(line)) - 1; i >= 0; i--) {
```
charliegracie: Sorry instead of casting the result of strlen to an int can we just give i a type of size_t? Did I miss a previous comment where someone suggested this instead?

PR #1345, opened 2017-07-12
[Mem1] Allocate using regions in TR_List and TR_BitVector

andrewcraik:
The current implementations of TR_BitVector and TR_List allocate memory
using trMemory(). This is problematic for growable data structures using
stack memory since new stack marks may cause the lifetime of internal
data structures to be inconsistent. In addition, the lack of a
TR::Region based API prohibits modelling memory lifetimes which do not
follow the existing stack mark or heap.

This change modifies the data structures to store a region to do their
memory management and to always use the same region for their
allocation. The trMemory based APIs derrive the appropriate default
region to minimize impact on the code.

To facilitate allocation from region overrides of the pacement new and
delete operators taking TR:Region are supplied to mirror those for
trMemory and friends.

### DIFF
--- a/compiler/infra/BitVector.hpp
+++ b/compiler/infra/BitVector.hpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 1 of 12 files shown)


lmaisons:
Does this preempt pull #1334 ?

compiler/infra/BitVector.hpp
```diff
@@ -156,25 +156,68 @@ class TR_BitVector
...
+      if (_numChunks)
+         {
+         if (_region)
+            {
+            _chunks = (chunk_t*)_region->allocate(_numChunks*sizeof(chunk_t));
+            }
+         else
+            {
+            TR_ASSERT(allocKind == persistentAlloc, "Should not allocate through trMemory except for persistent memory");
+            _chunks = (chunk_t*)m->allocateMemory(_numChunks*sizeof(chunk_t), persistentAlloc, ot);
```
lmaisons: Nit: Could this please use jitPersistentAlloc for consistency? This appears to be the only place left in the using this form.

lmaisons:
Beyond the nit, this looks OK to me.

PR #1346, opened 2017-07-12
[Mem2] UseDef memory allocation cleanup

andrewcraik:
Currently the UseDef analysis has an auxiliary structure used to hold
data only live during usedef analysis. This structure, while allocated o
the stack, is making heavy use of the CS2 allocator. Much of the heap
memory used is not reclaimed and is leading to large footprint
overheads.

These changes refactor the AuxiliaryData structure and UseDefInfo
itself to allocate from their own regions using non-CS2 data strucutres
to facilitate freeing memory efficiently.

### DIFF
--- a/compiler/il/OMRNode.hpp
+++ b/compiler/il/OMRNode.hpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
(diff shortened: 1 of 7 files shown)


andrewcraik:
Note the first commit 0a23551 is pull requested separately as Mem1 - this change consists of the other 3 changes and would need to be merged after Mem1

andrewcraik:
I have revised vector.hpp and deque.hpp per request.

lmaisons:
Updated changes LGTM

andrewcraik:
commit messages amended per request and rebased.

PR #1347, opened 2017-07-12
[Mem3] Remove CS2 BitVectors from DataFlow

andrewcraik:
The DataFlow engine was designed to operate with stack memory and the
associated stack marks and releases. The BitVectors were previously
changed to use CS2 dense BitVectors allocated out of a CS2::TableOf
pool. The pool is leaking large amounts of memory and the CS2 dense
BitVectors are not compatible with region allocation. This change swaps
the BitVectors back to region allocated TR_BitVectors and removes the
CS2::TableOf pool for significant memory savings.

### DIFF
--- a/compiler/optimizer/BackwardBitVectorAnalysis.cpp
+++ b/compiler/optimizer/BackwardBitVectorAnalysis.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -19,7 +19,6 @@
 #include <stddef.h>                                 // for NULL
 #include <stdint.h>                                 // for int32_t
 #include "compile/Compilation.hpp"
-#include "cs2/bitvectr.h"
 #include "env/TRMemory.hpp"                         // for BitVector, etc
 #include "il/Block.hpp"                             // for Block
 #include "il/Node.hpp"                              // for Node, vcount_t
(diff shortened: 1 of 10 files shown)


andrewcraik:
The change consists only of commit a7e2a19 the other commits are from Mem1 and Mem2 - this should be merged after them.

lmaisons:
The changes for `Remove CS2 BitVectors from DataFlow` [delta between this and mem2] look OK to me.

andrewcraik:
rebased on revised Mem2

compiler/optimizer/BackwardBitVectorAnalysis.cpp
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
```
0xdaryl: Is there any need for the  `#include "cs2/bitvectr.h"` anymore in this file?  We should be breaking as many CS2 dependencies as possible.  If not needed, please remove.

compiler/optimizer/BackwardBitVectorAnalysis.cpp
```diff
@@ -825,13 +825,13 @@ template<class Container>void TR_BackwardDFSetAnalysis<Container *>::initializeG
       }
 
 
-   TR::BitVector pendingList(this->comp()->allocator());
+   TR_BitVector pendingList(this->comp()->trMemory()->currentStackRegion());
    //TR_BitVector *originalPendingList = new (this->trStackMemory()) TR_BitVector(this->_numberOfNodes, this->trMemory(), stackAlloc);
```
0xdaryl: Does this commented line have significance anymore?

compiler/optimizer/BackwardBitVectorAnalysis.cpp
andrewcraik: Removed.

compiler/optimizer/BackwardBitVectorAnalysis.cpp
andrewcraik: No significance - removed.

PR #1349, opened 2017-07-12
[Mem4] Clean RegisterCandidates memory management

andrewcraik:
The RegisterCandidates data structure has several CS2::TableOf data structures which are leaking a  lot of memory. This change removes the TableOf structures, introduces a memory allocation region, allocates the candidates from the region, and efficiently cleans up candidates when freed. The top two commits are in this pull request - Mem1 - Mem3 are the base.

### DIFF
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -82,35 +82,28 @@
 #define keepAllStores false
 
 TR::GlobalSet::GlobalSet(TR::Compilation * comp, const TR::Allocator &alloc)
-   :_refAutosPerBlock(alloc),_comp(comp),_maxEntries(0)
+   :_refAutosPerBlock((RefMapComparator()),(RefMapAllocator(comp->trMemory()->heapMemoryRegion()))),
+    _comp(comp),_maxEntries(0)
    {
    }
 
 
 void TR::GlobalSet::DenseSet::print(TR::Compilation * comp)
    {
-   TR::BitVector::Cursor bit(_refs);
-   for (bit.SetToFirstOne(); bit.Valid(); bit.SetToNextOne())
+   TR_BitVectorIterator bits(_refs);
+   for (int32_t bit = bits.getFirstElement(); bits.hasMoreElements(); bit = bits.getNextElement())
      {
-       CS2::BitIndex bitNum = bit;
-       if (_refs[bitNum])
-         {
-          traceMsg(comp,"%d ",bitNum);
-         }
+     traceMsg(comp,"%d ",bit);
      }
    traceMsg(comp,"\n");
    }
 
 void TR::GlobalSet::SparseSet::print(TR::Compilation * comp)
    {
-   TR::SparseBitVector::Cursor bit(_refs);
-   for (bit.SetToFirstOne(); bit.Valid(); bit.SetToNextOne())
+   TR_BitVectorIterator bits(_refs);
+   for (int32_t bit = bits.getFirstElement(); bits.hasMoreElements(); bit = bits.getNextElement())
      {
-       CS2::SparseBitIndex bitNum = bit;
-       if (_refs[bitNum])
-         {
-          traceMsg(comp,"%d ",bitNum);
-         }
+     traceMsg(comp,"%d ",bit);
      }
    traceMsg(comp,"\n");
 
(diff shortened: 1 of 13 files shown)


vijaysun-omr:
I would like @lmaisons to review this too.

compiler/optimizer/UseDefInfo.cpp
```diff
@@ -2215,14 +2218,14 @@ void TR_UseDefInfo::dereferenceDef(TR_UseDefInfo::BitVector &useDefInfo, int32_t
    {
    TR_ASSERT(defIndex < getTotalNodes(), "TR_UseDefInfo::getNode index(%d) is bigger than total(%d)\n", defIndex, getTotalNodes());
 
-   CS2::LinkedListOf<CS2::Pair<TR::Node *,TR::TreeTop *>, TR::Allocator> defIndices(comp()->allocator());
-   defIndices.Push(_atoms[defIndex]);
+   TR::list<std::pair<TR::Node *,TR::TreeTop *>, TR::Region&> defIndices(_region);
```
lmaisons: This likely would perform better as a deque, or even possibly a vector if we swapped to using push_back / back.  Though it may not enough use to matter.

compiler/optimizer/UseDefInfo.hpp
```diff
@@ -373,10 +374,10 @@ class TR_UseDefInfo : public TR::Allocatable<TR_UseDefInfo, TR::Allocator>
 
       friend class TR_UseDefInfo;
       };
-   typedef CS2::LinkedListOf<MemorySymbol,TR::Allocator> MemorySymbolList;
+   typedef TR::list<MemorySymbol, TR::Region&> MemorySymbolList;
 
    int32_t                 _numMemorySymbols;
-   CS2::ArrayOf<MemorySymbolList, TR::Allocator> _valueNumbersToMemorySymbolsMap;
+   TR::vector<MemorySymbolList *, TR::Region&> _valueNumbersToMemorySymbolsMap;
```
lmaisons: What was the reason for making this a container of pointers rather than storing the sub-containers directly?  Also, the only use I could see of the sub-lists was them being used as a stack.  This appears to be another candidate for replacement with deque / vector.

andrewcraik: This change was not in this commit - it was in Mem2. This was converted to a pointer because of issues compiling with GCC 4.8 where value typed members were causing the allocator to be copied when the vector was resized.

compiler/optimizer/RegisterCandidate.hpp
```diff
@@ -50,19 +47,20 @@ namespace TR
...
    }
 
    void collectReferencedAutoSymRefs(TR::Block * BB);
-   bool isEmpty() { return _refAutosPerBlock.IsEmpty(); }
-   void makeEmpty() { _refAutosPerBlock.MakeEmpty(); }
+   bool isEmpty() { return _refAutosPerBlock.empty(); }
+   void makeEmpty() { _refAutosPerBlock.clear(); }
    void initialize(uint32_t numSymRefs, int32_t numBlocks) {_maxEntries = numSymRefs * numBlocks;}
 
    //Set, SparseSet and DenseSet
```
andrewcraik: I'm open to seeing if there is a better solution. For now I think this will do, but we can consider it. I believe these are all using bitvectors so that might be an advantage in this particular case.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -3847,24 +3828,28 @@ TR_RegisterCandidates::computeAvailableRegisters(TR_RegisterCandidate *rc, int32
...
 
-         if (!overlapTable->Exists(blockNumber))
+         if (overlapTable->find(blockNumber) == overlapTable->end())
             {
-            Coordinates c(comp()->allocator());
-            overlapTable->AddEntryAtPosition(blockNumber, c);
-            Coordinates &overlaps = (*overlapTable)[blockNumber];
-            ComputeOverlaps(b, comp(), overlaps);
+            Coordinates *c = new (comp()->trMemory()->currentStackRegion()) Coordinates((CoordinatesComparator()), (CoordinatesAllocator(comp()->trMemory()->currentStackRegion())));
+            //Coordinates *c = new (comp()->trMemory()->currentStackRegion()) Coordinates(comp()->allocator());
```
andrewcraik: Nope, cleaning it up now.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -3847,24 +3828,28 @@ TR_RegisterCandidates::computeAvailableRegisters(TR_RegisterCandidate *rc, int32
...
          TR::Block * b = blocks[blockNumber];
 
-         if (!overlapTable->Exists(blockNumber))
+         if (overlapTable->find(blockNumber) == overlapTable->end())
             {
-            Coordinates c(comp()->allocator());
-            overlapTable->AddEntryAtPosition(blockNumber, c);
-            Coordinates &overlaps = (*overlapTable)[blockNumber];
-            ComputeOverlaps(b, comp(), overlaps);
+            Coordinates *c = new (comp()->trMemory()->currentStackRegion()) Coordinates((CoordinatesComparator()), (CoordinatesAllocator(comp()->trMemory()->currentStackRegion())));
```
andrewcraik: So there was a dance because of the copy construction - conversion to a pointer makes this particular bit easier to handle. Further, the Coordinate won't escape - the table is private to the opt, there is no getter, and the table is lexically allocated in the assign function and so the table dies once that scope ends. Stack allocation is , therefore, reasonable for this I believe.

andrewcraik:
Updated with the removal of the commented line.

compiler/optimizer/RegisterCandidate.cpp
lmaisons: Makes sense; happy to leave as-is.

compiler/optimizer/UseDefInfo.hpp
lmaisons: Ah this was the cause of that.  I opened #1366 for this as this should get fixed.  Meanwhile, no need to block this.

PR #1350, opened 2017-07-12
[Mem5] Clean Dominators memory management

andrewcraik:
Currently a Dominators analysis has a number of data structures scoped
to the lifetime of the analysis. These structures are currently CS2
containers allocated using the CS2 allocator. This change refactors
Dominators to have an embedded region and switches the data structures
to all allocate from that region. The TableOf in the current
implementation was especially bad at leaking memory.

### DIFF
--- a/compiler/optimizer/Dominators.cpp
+++ b/compiler/optimizer/Dominators.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -24,9 +24,6 @@
 #include "compile/Compilation.hpp"             // for Compilation, etc
 #include "control/Options.hpp"
 #include "control/Options_inlines.hpp"
-#include "cs2/bitvectr.h"                      // for ABitVector<>::BitRef
-#include "cs2/sparsrbit.h"
-#include "cs2/tableof.h"                       // for TableOf
 #include "env/IO.hpp"
 #include "env/TRMemory.hpp"                    // for Allocator
 #include "il/Block.hpp"                        // for Block, toBlock
@@ -39,10 +36,11 @@
 #include "ras/Debug.hpp"                       // for TR_DebugBase
 
 TR_Dominators::TR_Dominators(TR::Compilation *c, bool post) :
+   _region(c->trMemory()->heapMemoryRegion()),
    _compilation(c),
-   _info(c->getFlowGraph()->getNextNodeNumber()+1, c->allocator(), c->allocator()),
-   _dfNumbers(c->getFlowGraph()->getNextNodeNumber()+1, 0, c->allocator()),
-   _dominators(c->getFlowGraph()->getNextNodeNumber()+1, static_cast<TR::Block *>(NULL), c->allocator())
+   _info(c->getFlowGraph()->getNextNodeNumber()+1, BBInfo(_region), _region),
+   _dfNumbers(c->getFlowGraph()->getNextNodeNumber()+1, 0, _region),
+   _dominators(c->getFlowGraph()->getNextNodeNumber()+1, static_cast<TR::Block *>(NULL), _region)
    {
    LexicalTimer tlex("TR_Dominators::TR_Dominators", _compilation->phaseTimer());
 
@@ -124,7 +122,7 @@ TR_Dominators::TR_Dominators(TR::Compilation *c, bool post) :
    if (trace())
       traceMsg(comp(), "End of %sdominator calculation\n", _postDominators ? "post-" : "");
 
-   // Release no-longer-used CS2 data
+   // Release no-longer-used data
    _info.clear();
    }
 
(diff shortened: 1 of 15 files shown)


vijaysun-omr:
I would like a review from @lmaisons on this.

compiler/optimizer/Dominators.hpp
```diff
@@ -55,30 +54,26 @@ class TR_Dominators
...
-   typedef TR::SparseBitVector SparseBitVector;
+   typedef TR_BitVector BitVector;
 
    private:
 
    struct BBInfo
       {
-      BBInfo(TR::Allocator allocator) :
-         _bucket(allocator), _block(NULL), _parent(-1), _idom(-1), _ancestor(-1), _label(-1), _child(-1), _sdno(-1), _size(0)
+      BBInfo(TR::Region &region) :
```
lmaisons: I clearly missed it myself when going through this, but it might make sense to make this constructor explicit and explicitly construct the BBInfo from from a region in locations where this currently happens automagically.

compiler/optimizer/Dominators.cpp
```diff
@@ -39,10 +36,11 @@
 #include "ras/Debug.hpp"                       // for TR_DebugBase
 
 TR_Dominators::TR_Dominators(TR::Compilation *c, bool post) :
+   _region(c->trMemory()->heapMemoryRegion()),
    _compilation(c),
-   _info(c->getFlowGraph()->getNextNodeNumber()+1, c->allocator(), c->allocator()),
-   _dfNumbers(c->getFlowGraph()->getNextNodeNumber()+1, 0, c->allocator()),
-   _dominators(c->getFlowGraph()->getNextNodeNumber()+1, static_cast<TR::Block *>(NULL), c->allocator())
+   _info(c->getFlowGraph()->getNextNodeNumber()+1, _region, _region),
```
lmaisons: I think one of these `_region` parameters ends up being an implicit cast to a BBInfo.  It would be nice if this were made more obvious via an explicit construction-style cast.

compiler/optimizer/Dominators.cpp
```diff
@@ -459,14 +457,15 @@ void TR_PostDominators::findControlDependents()
    int32_t nextNodeNumber = _cfg->getNextNodeNumber();
    int32_t i;
 
+   _directControlDependents = (TR_BitVector**)_region.allocate(nextNodeNumber*sizeof(TR_BitVector *));
```
lmaisons: Was double indirection necessary or cleaner here?  At first glance this looks like it could 'just' be a dynamically allocated array of values, but that it might also be more trouble than it's worth.

compiler/optimizer/Dominators.cpp
andrewcraik: Yes the first should be explicit.

compiler/optimizer/Dominators.hpp
andrewcraik: Good point

andrewcraik:
revision to make the BBInfo constructor explicit

lmaisons:
BBInfo changes look OK to me.

compiler/optimizer/Dominators.cpp
andrewcraik: I'm not sure whether the in place value TR_BitVector will work nicely with all the allocator copying? This mathces how a number of data structures are declared in the optimizer so I was following that pattern. We could consider reworking it as you suggest in the future.

compiler/optimizer/Dominators.hpp
```diff
@@ -55,30 +54,26 @@ class TR_Dominators
...
+         _bucket(region), _block(NULL), _parent(-1), _idom(-1), _ancestor(-1), _label(-1), _child(-1), _sdno(-1), _size(0)
          {}
       TR::Block     *_block;     // The block whose info this is
       int32_t       _parent;    // The parent in the depth-first spanning tree
       int32_t       _idom;      // The immediate dominator for this block
       int32_t       _ancestor;  // The ancestor in the forest
       int32_t       _label;     // The node in the ancestor chain with minimal
                                 // semidominator number
-      SparseBitVector _bucket;    // The blocks whose semidominator is this node
+      BitVector _bucket;    // The blocks whose semidominator is this node
```
vijaysun-omr: This looks like a place to expect a sparse bit vector to be an appropriate choice based on the thought that the set of "blocks whose semidominator is this node" is likely not a set that approaches the full set of blocks. However it may be that that is okay as a high level thought, but when one looks at the details of the sparse bit vector implementation it is possible that it takes more memory in some cases like reasonable sized (though not small) control flow graphs. So my question is : is there something inherent here that this set is'nt as sparse as I think it is OR is it that our sparse bit vector implementation does'nt quite work as well ?

andrewcraik: So the issue here is that the BitVector that was being used was a CS2 sparse bit vector. The sparseness is good, but it allocates its memory from the heap and it is not recycled effectively due to heap fragmentation. The TR_BitVector is semi-sparse (compared to the CS2 dense BitVector). but the major advantage here is that the TR_BitVector is compatible with region allocators and so the memory can be fully cleaned up. I changed the bit vector for that reason - all the data structures from the analysis now are allocated in a region and that region will be fully cleaned up leading to lower heap memory usage over the course of a compile rather than the slow leak seen with the CS2 allocator for the SparseBitVector.

vijaysun-omr:
Usage of the TR_BitVector in bit vector analysis makes that code more readable (at least for me). Thanks for doing that.

vijaysun-omr:
Looks ok to me. Merging.

PR #1351, opened 2017-07-12
[Mem6] Clean memory management in LocalDSE

andrewcraik:
Currently LocalDSE allocates a CS2::TableOf that is never used. Thiscommit removes that table entirely and switches other data structures to be allocated in the stack scope of the optimization so that they are properly cleaned up.


lmaisons:
`Clean memory management in LocalDSE` looks OK to me.

andrewcraik:
rebased

vijaysun-omr:
Looks ok to me. Merging.

PR #1352, opened 2017-07-14
Link and compile with -pthread

rwy0717:
We were previously consuming pthread in an ad-hoc way.
This now propagates -pthread through anything linking against
j9thrstatic (that's nearly everything).

`-pthread` has the same effect as `-lpthread` and `-D_REENTRANT`. Now,
we don't need to define `_REENTRANT` ourselves.

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -77,8 +77,8 @@ endif()
 if(OMR_HOST_OS STREQUAL "linux")
 	add_definitions(
 		-DLINUX
-		-D_REENTRANT
 		-D_FILE_OFFSET_BITS=64
+		-pthread
 	)
 endif()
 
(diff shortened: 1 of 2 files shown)


dnakamura:
The only thing I would be concerned about is -D_REENTRANT gets applied to all the targets, but not all the targets link against j9thrstatic

rwy0717:
I used `j9thrstatic.a` as an indicator of threads. But I see your point--libraries not using threads will need `_REENTRANT`. Should we compile/link **all** targets with `-pthread`? I am starting to think yes.

rwy0717:
The reason why we don't want to simply define `_REENTRANT`:
1. `_REENTRANT` is a compiler detail
2. `-pthread` could in theory affect compilation in subtle ways (E.G. memory model)
3. `-pthread` is the documented Right Way™

rwy0717:
I've fixed this up to use `-pthread` everywhere, not just as part of the `j9thrstatic.a` interface.

dnakamura:
lgtm

PR #1353, opened 2017-07-14
Disable spin/wake feature on zOS

babsingh:
The spin/wake feature is disabled on zOS due to poor performance.

The spin/wake feature requires the OMR_THR_THREE_TIER_LOCKING flag to be
enabled. On AIX, the OMR_THR_THREE_TIER_LOCKING flag is disabled. So,
the spin/wake feature is disabled on AIX.

As of now, the OMR_THR_SPIN_WAKE_CONTROL is excluded from the following
configure_*.mk files:

1. configure_common.mk
2. configure_aix_ppc.mk
3. configure_zos_390.mk


fjeremic:
This PR fixes an important scalability issue on z/OS. Higher n-way workloads are unable to saturate CPU resources when the spin/wake feature is enabled.

PR #1355, opened 2017-07-14
Fix a crash in analyzeExitEdges()

klangman:
In analyzeExitEdges() we need to ensure we are only considering blocks
for the target loop and not any inner loops

Signed-off-by: Kevin Langman <langman@ca.ibm.com>


vijaysun-omr:
Looks like a safe change to me since an inner loop having an update to a variable should basically mean we do not consider it well behaved enough to be an induction variable in an outer loop.

PR #1360, opened 2017-07-14
Replace TreeVerifier with the ILValidator

### LINKED ISSUE #1298: Fix the compiler's Tree Verifier

Leonardo2718:
This change takes the previously implemented ILValidator, brings it up to date, and uses it to replace the old TreeVerifier in the compiler component.

This change takes the implementation of the (previously unused) ILValidator and brings it up to date with the rest of the code base. It also implements the necessary mechanisms to dispatch the ILValidator from the optimizer. The `useIlValidator` option is added which, if specified, will cause the optimizer to use the new ILValidator instead of the old TreeVerifier. This will allow downstream projects to continue using the TreeVerifier until they are ready to switch to the new ILValidator. Since the ILValidator is more robust/aggressive than the TreeVerifier, some projects may find that they have to fix problems that were previously unreported. JitBuilder and the test compiler already use the ILValidator by default by hard coding the `useIlValidator` option in their JIT initialization routines.

This change also adds some Tril-based tests for the ILValidator.

Lastly, this change also fixes some of the issues that were reported by the ILValidator during testing. These include both legitimate problems in existing projects as well as problems in the validator itself.

Fixes #1298


jitbuilder/control/Jit.cpp
```diff
@@ -179,7 +179,7 @@ extern "C"
 bool
 initializeJit()
    {
-   return initializeJitBuilder(0, 0, 0, (char *)"-Xjit:acceptHugeMethods,enableBasicBlockHoisting,omitFramePointer");
+   return initializeJitBuilder(0, 0, 0, (char *)"-Xjit:acceptHugeMethods,enableBasicBlockHoisting,omitFramePointer,useIlValidator");
```
mgaudet: Do we want the validator turned on by default for JitBuilder? That's a bit surprising to me.

compiler/ras/ILValidator.cpp
```diff
@@ -313,6 +315,36 @@ bool TR::ILValidator::isLoggingEnabled()
    return (comp()->getOption(TR_TraceILWalks));
    }
 
+void TR::ILValidator::printDiagnositc(const char *formatStr, ...)
```
mgaudet: Typo in name: `Diagnostic` (not `Diagnositc`)

compiler/infra/Assert.cpp
```diff
@@ -38,6 +38,38 @@
 #include "env/VMJ9.h"
 #endif
 
+void OMR_NORETURN TR::trap()
+   {
+   static char * noDebug = feGetEnv("TR_NoDebuggerBreakPoint");
+   if (noDebug)
```
Leonardo2718: Fixed.

compiler/ras/ILValidator.cpp
Leonardo2718: Fixed.

jitbuilder/control/Jit.cpp
Leonardo2718: The `useIlValidator` option only makes it so that *if* validation is enabled (e.g. using `paranoidOptCheck`), the ILValidator will be used instead of the TreeVerifier. It does not enable validation by default.

mgaudet: Ooooooo. 

I see now. 🆒

Thanks!

compiler/optimizer/LocalOpts.cpp
```diff
@@ -8131,7 +8132,7 @@ TR_ColdBlockMarker::perform()
    static char *validate = feGetEnv("TR_validateBeforeColdBlockMarker");
    if (validate)
       {
-      TR::ILValidator validator(this);
+      TR::ILValidator validator(comp());
       validator.checkSoundness(comp()->getStartTree());
```
0xdaryl: The first thing that `treesAreValid` does is `checkSoundness`.  Is this additional call really needed here?

compiler/ras/ILValidator.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+   TR::Compilation *comp();
+
+   void updateNodeState(Location &newLocation);
+
+   void printDiagnostic(const char *formatStr, ...);
+   void vprintDiagnostic(const char *formatStr, va_list ap);
+   };
+
+class JavaILValidator: public ILValidator
```
Leonardo2718: Originally, `JavaILValidator` was in `omr/compiler/infra/ILWalk.{hpp,cpp}`. It just got moved around as part of my refactoring. I've removed it from the OMR compiler though and it will now live as part of the Java project.

compiler/optimizer/LocalOpts.cpp
Leonardo2718: This is how it was previously implemented. I'll remove though because it really is redundant.

compiler/ras/ILValidator.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+static bool hasGlRegDeps(TR::Node* node)
+   {
+   for (auto i = 0; i < node->getNumChildren(); ++i)
+      {
+      if (node->getChild(i)->getOpCodeValue() == TR::GlRegDeps)
+         {
+         return true;
+         }
+      }
```
vijaysun-omr: Yes it should be the last child in every case I know of.

compiler/infra/Assert.hpp
```diff
@@ -21,6 +21,9 @@
  *
  * Defines a number of assert macros.
```
Leonardo2718: Very true. Fixed!

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -1627,7 +1627,7 @@
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::Int32,
    /* .typeProperties       = */ ILTypeProp::Size_4 | ILTypeProp::Integer,
-   /* .childProperties      = */ TWO_SAME_CHILD(TR::Int32),
+   /* .childProperties      = */ TWO_CHILD(TR::Int32, TR::Int32),
```
Leonardo2718: I'm afraid this nothing more than an automated refactoring artifact :sweat_smile:. Fixed!

compiler/infra/ILWalk.hpp
```diff
@@ -60,15 +59,17 @@ class TreeTopIteratorImpl
...
+
+   TR::Compilation * comp() { return _comp; }
 
    void logCurrentLocation();
 
    void stepForward();
    void stepBackward();
 
-   TreeTopIteratorImpl(TR::TreeTop *start, TR::Optimization *opt=NULL, const char *name=NULL);
+   TreeTopIteratorImpl(TR::TreeTop *start, TR::Compilation *comp=NULL, const char *name=NULL);
```
Leonardo2718: In the original implementation, it was the `TR::Optimization` that was optional. I just changed the type of the parameter.

I don't think it makes sense for these APIs to be used outside of a compilation, since they operate exclusive on the compiler's IL, which is only (currently) available during compilation.

Personally, I would prefer if the parameter was mandatory, but I think making this change is outside the scope of this already large pull request.

compiler/ras/ILValidator.cpp
```diff
@@ -0,0 +1,356 @@
...
+         }
+      else
+         {
+         soundnessRule(currentTree, stop == NULL, "Reached the end of the trees after n%dn without encountering the stop tree n%dn", currentTree->getNode()->getGlobalIndex(), stop? stop->getNode()->getGlobalIndex() : 0);
+         checkNodeSoundness(currentTree, currentTree->getNode(), ancestorNodes, visitedNodes);
+         }
+      }
+   }
+
+void TR::ILValidator::checkNodeSoundness(TR::TreeTop *location, TR::Node *node, NodeChecklist &ancestorNodes, NodeChecklist &visitedNodes)
```
Leonardo2718: I didn't originally write this code, I only moved it from `compiler/ras/ILWalk.cpp`, so I can only make educated guesses for why it was written this way.

However, I believe the reason for calling the first parameter `location` is that it is only used to print the "location" of a failure in diagnostic messages; it is not actually used to check anything. In this sense, it keeps the interface consistent with the rest of the API.

I actually prefer the name `location` to `treetop` because it makes the purpose of the parameter clear: to specify where the node being checked is "located". However, I'm open to re-naming it to something like `treetopLocation` or `parentTreetop`. Both of these names make the relationship with the `node` parameter clear (at least in my head).

compiler/ras/ILValidator.cpp
```diff
@@ -0,0 +1,356 @@
...
+      TR::Node *child = node->getChild(i);
+      soundnessRule(location, child != NULL, "n%dn child %d must not be NULL", node->getGlobalIndex(), i);
+
+      checkNodeSoundness(location, child, ancestorNodes, visitedNodes);
+      }
+
+   ancestorNodes.remove(node);
+   }
+
+void TR::ILValidator::soundnessRule(TR::TreeTop *location, bool condition, const char *formatStr, ...)
```
Leonardo2718: Same comment as above.

compiler/ras/ILValidator.cpp
Leonardo2718: If this is indeed the case, I'm open to changing this.

andrewcraik: If it isn't last I think it will surprise a lot of other code.

compiler/ras/ILValidator.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+      //
+      // Additional specific kinds of validation
+      //
+
+      TR::Node *node = iter.currentNode();
+      if (node->getOpCodeValue() == TR::BBEnd)
+         {
+         // Determine whether this is the end of an extended block
+         //
```
andrewcraik: There would be an opportunity here to catch another kind of oops - where you have broken a block and failed to extend when it makes sense to do so. This would be more of a warning, but it could be a nice extra in another iteration to check if there is a single successor, single predecessor relationship and block extension has been run then we should expect the next block to be marked an extension.

Leonardo2718: This does sound like something that could be worth checking. Is there a reason this couldn't be done in the CFGVerifier instead?

andrewcraik: Probably not - it just occurred to me while I was reading all the code about checking extension here.

compiler/ras/ILValidator.cpp
Leonardo2718: Excellent! Fixed!

compiler/ras/ILValidator.cpp
```diff
@@ -0,0 +1,339 @@
...
+               const auto actChildType = childOpcode.getDataType().getDataType();
+               const auto expChildTypeName = expChildType == ILChildProp::UnspecifiedChildType ? "UnspecifiedChildType" : TR::DataType::getName(expChildType);
+               const auto actChildTypeName = TR::DataType::getName(actChildType);
+               validityRule(iter, expChildType == ILChildProp::UnspecifiedChildType || actChildType == expChildType,
+                            "Child %d has unexpected type %s (expected %s)" , i, actChildTypeName, expChildTypeName);
+               }
+            else
+               {
+               // make sure the node is allowed to have a GlRegDeps child
+               // and make sure that it is the last child
```
mgaudet: 👍 Nice. I was happy to see that got converted to a validity rule.

compiler/infra/ILWalk.hpp
0xdaryl: I think the parameter should be mandatory too.  I don't quite understand your "scope" comment though because to fix this is simply a matter of removing the default value.  If there are places in the code that rely on the default value and don't pass in a valid compilation object I would argue that they are now potentially incorrect anyway because some of the underlying validator code relies on having a compilation object (whose access is not always guarded).

Leonardo2718: Fair enough.

Leonardo2718:
Thanks @0xdaryl!

> Given recent experiences with memory leaks in other data structures I think the data structures supporting IL validation need a comprehensive review as well to be sure their lifetimes are understood and memory is reclaimed accordingly. I don't think this should hold up merging this initial PR, but I do think its something that needs to get done before this technology become ubiquitous.

Completely agree. I'll open an issue to track this once the PR is merged.

> It is unfortunate that the shiftOrRol name from the Test compiler got morphed into "shift and roll" which I don't think makes sense (unless I'm way off the mark on the context). It's likely named as such because OMR only supports a rotate left opcode, which I'm a bit surprised to discover only now! The Test compiler tests should have been more aptly named shiftAndRotate, and hence your descendants written in Tril should be named accordingly and avoid the mis-named "roll". I'd like this to be fixed in your Tril implementation and commit messages before this gets merged to avoid perpetuating confusion.

Should be fixed now.

> I appreciate the effort you put into re-writing the invalid shiftOrRol tests in Tril. As we are actively working to upgrade the Google Test framework in order to enable Tril as part of TravisCI I am fine with deprecating the Test compiler tests as you have done so in this PR. I doubt this is leaving us exposed to any significant extent on the quality front for a short period of time.

Agreed. Hopefully we can get Tril enable as soon :slightly_smiling_face:.

compiler/ras/ILValidator.cpp
0xdaryl: Alright, it can stay the way it is.

compiler/ras/ILValidator.cpp
0xdaryl: Same reply as above.

0xdaryl:
@mgaudet , please approve Leo's changes if you're OK with them.  I'll merge once you've signed off.

PR #1363, opened 2017-07-17
Extend OSR Induce API to accept optional CFG

ncough:
OSR induce calls are added during inlining for
OSROnGuardFailure. The OSR Induce API currently
assumes that the outermost CFG should be used
when manipulating blocks to add an induce call.
This is not valid when inlining a call whose
caller is not the outermost method. To correct
this, the API has been extended with an optional
CFG argument that defaults to the outermost.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.hpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.hpp
@@ -152,16 +152,16 @@ public:
 
    bool genPartialIL(TR_FrontEnd *, TR::Compilation *, TR::SymbolReferenceTable *, bool forceClassLookahead = false, TR_InlineBlocks *blocksToInline = 0);
 
-   void genOSRHelperCall(int32_t inlinedSiteIndex, TR::SymbolReferenceTable* symRefTab);
+   void genOSRHelperCall(int32_t inlinedSiteIndex, TR::SymbolReferenceTable* symRefTab, TR::CFG *cfg = NULL);
    void genAndAttachOSRCodeBlocks(int32_t inlinedSiteIndex) ;
    int  matchInduceOSRCall(TR::TreeTop* insertionPoint,
                                int16_t callerIndex,
                                int16_t byteCodeIndex,
                                const char* childPath);
 
-   TR::TreeTop *genInduceOSRCallNode(TR::TreeTop* insertionPoint, int32_t numChildren, bool copyChildren, bool shouldSplitBlock = true);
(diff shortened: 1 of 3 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.hpp
```diff
@@ -152,16 +152,16 @@ class OMR_EXTENSIBLE ResolvedMethodSymbol : public TR::MethodSymbol
...
    void genAndAttachOSRCodeBlocks(int32_t inlinedSiteIndex) ;
    int  matchInduceOSRCall(TR::TreeTop* insertionPoint,
                                int16_t callerIndex,
                                int16_t byteCodeIndex,
                                const char* childPath);
 
-   TR::TreeTop *genInduceOSRCallNode(TR::TreeTop* insertionPoint, int32_t numChildren, bool copyChildren, bool shouldSplitBlock = true);
-   TR::TreeTop *genInduceOSRCall(TR::TreeTop* insertionPoint, int32_t inlinedSiteIndex, int32_t numChildren, bool copyChildren, bool shouldSplitBlock);
-   TR::TreeTop *genInduceOSRCall(TR::TreeTop* insertionPoint, int32_t inlinedSiteIndex, TR_OSRMethodData *osrMethodData, int32_t numChildren, bool copyChildren, bool shouldSplitBlock);
+   TR::TreeTop *genInduceOSRCallNode(TR::TreeTop* insertionPoint, int32_t numChildren, bool copyChildren, bool shouldSplitBlock = true, TR::CFG *cfg=NULL);
```
andrewcraik: for style add spaces around your NULLs in these three lines.

andrewcraik:
I would say for clarity the default argument should be comp()->getFlowGraph() if that is possible in the default argument. If it isn't and it has to be NULL, each function accepting null should test for the NULL parameter as the first if and set the parameter to comp()->getFlowGraph() so that it is obvious in the rest of the function that cfg is non-NULL and is always ok to use - I'd rather the logic wasn't sprayed deeper down in the function for maintainability and readability.

ncough:
Updated to address the style and maintainability issues raised by @andrewcraik.

andrewcraik:
LGTM - @vijaysun-omr this is an important issue to address since the OSR infrastructure can currently graph the wrong CFG leading to compilation issues.

vijaysun-omr:
I will wait for checks to pass before merging. Looks safe to me since it uses the "right" CFG more often.

PR #1364, opened 2017-07-17
Add missing scavenger ifdefs

rwy0717:
/cc: @ktbriggs @charliegracie 


gc/base/standard/EnvironmentStandard.hpp
```diff
@@ -93,7 +95,9 @@ class MM_EnvironmentStandard : public MM_EnvironmentBase
 	 */
 	void flushRememberedSet()
```
charliegracie: Should we wrap the function in a #if defined or just the code inside?  I am not sure about this one.  Rememberedset in this context is very scavenger specific so should anyone even call this unless they are using the scavenger?

rwy0717: I think you're right, I'll update to wrap the entire function.

charliegracie: Also I believe the flag is OMR_GC_MODRON_SCAVENGER and not OMR_GC_SCAVENGER

PR #1367, opened 2017-07-17
Increase the limit of constant data snippet

r30shah:
On z Systems, constant data snippet size was limited to
16 bytes. This commit increases that to 64 bytes.

### DIFF
--- a/compiler/z/codegen/ConstantDataSnippet.hpp
+++ b/compiler/z/codegen/ConstantDataSnippet.hpp
@@ -26,6 +26,7 @@
 #include "compile/Compilation.hpp"  // for comp, Compilation (ptr only)
 #include "env/jittypes.h"           // for uintptrj_t
 #include "infra/Assert.hpp"         // for TR_ASSERT
+#include "OMRCodeGenerator.hpp"
 
 class TR_Debug;
 namespace TR { class CodeGenerator; }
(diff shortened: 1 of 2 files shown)


r30shah:
fyi @fjeremic

compiler/z/codegen/ConstantDataSnippet.hpp
```diff
@@ -47,7 +47,7 @@ class S390ConstantDataSnippet : public TR::Snippet
    protected:
    union
       {
-      uint8_t                       _value[16];
+      uint8_t                       _value[64];
```
fjeremic: Can this be `1 << TR_DEFAULT_DATA_SNIPPET_EXPONENT` so as to link the two entities together. I also question the value of `TR_DEFAULT_DATA_SNIPPET_EXPONENT` to begin with. Is there any possibility to simplify this code? In what context is `constantDataSnippetExponent` used?

r30shah: @fjeremic  I have changed the size of _value array with `1 << TR_DEFAULT_DATA_SNIPPET_EXPONENT` to link them. With regard to your question, We can only allocate Data snippet in power of 2 and in method, writable snippets are emitted in order of decreasing size. So `constantDataSnippetExponent`to get the maximum size of snippet and we start from that point.

PR #1369, opened 2017-07-18
Add new Xjit option enableAggressiveLoopVersioning

JamesKingdon:
enableAggressiveLoopVersioning is a new jit option that will be used to
guard a number of changes to the behaviour of LoopVersioner.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -615,6 +615,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"dynamicThreadPriority",              "M\tenable dynamic changing of compilation thread priority", SET_OPTION_BIT(TR_DynamicThreadPriority), "F", NOT_IN_SUBSET},
    {"earlyLPQ",                           "M\tAllow compilations from low priority queue to happen early, during startup", SET_OPTION_BIT(TR_EarlyLPQ), "F", NOT_IN_SUBSET },
    {"enableAggressiveLiveness",           "I\tenable globalLiveVariablesForGC below warm", SET_OPTION_BIT(TR_EnableAggressiveLiveness), "F"},
+   {"enableAggressiveLoopVersioning", "O\tOptions and thresholds that result in loop versioning occurring in more cases", SET_OPTION_BIT(TR_EnableAggressiveLoopVersioning), "F" },
    {"enableAllocationOfScratchBTL",       "M\tAllow the allocation scratch memory below the line (zOS 31-bit)", RESET_OPTION_BIT(TR_DontAllocateScratchBTL), "F", NOT_IN_SUBSET },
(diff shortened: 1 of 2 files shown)


JamesKingdon:
This will be a pre-req for four upcoming changes.

compiler/control/OMROptions.cpp
```diff
@@ -100,6 +100,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
         TR::Options::setStaticNumeric, (intptrj_t)&OMR::Options::_abstractTimeToReduceInliningAggressiveness, 0, "F%d", NOT_IN_SUBSET },
    {"acceptHugeMethods",     "O\tallow processing of really large methods", SET_OPTION_BIT(TR_ProcessHugeMethods), "F" },
    {"activateCompThreadWhenHighPriReqIsBlocked", "M\tactivate another compilation thread when high priority request is blocked",  SET_OPTION_BIT(TR_ActivateCompThreadWhenHighPriReqIsBlocked), "F", NOT_IN_SUBSET},
+   {"aggressiveLoopVersionerOpts", "O\tUse more loop versioning", SET_OPTION_BIT(TR_aggressiveLoopVersionerOpts), "F" },
```
fjeremic: This option sounds weird because its not in an imperative mood mood. IMO options should be commands. Would `aggressivelyVersonLoops` not sounds more imperative?

JamesKingdon: We have several precedents for non-imperative options, and I wasn't aware of an intent to make Xjit options follow an imperative naming scheme. Isn't it reasonable to consider any boolean option foo to be implicitly equivalent to enableFoo, with the 'enable' omitted for brevity?

I have no particular attachment to the name though, so happy to change it if you wish.

fjeremic: @JamesKingdon this was just _my_ personal preference and as such I did not request changes in my review, but rather a 'review comment' in GitHub terms. To explain my rationale to me `-Xjit:aggressiveLoopVersionerOpts` doesn't really say much. Are we enabling aggressive loop versioner opts? Are we tracing them? Are we disabling them?

fjeremic: I've created https://github.com/eclipse/omr/issues/1370 to kick things off.

JamesKingdon: Excellent, that looks like a useful discussion to have. I'll push forward on the dependent changes and circle round to the name for the option at the end, by which time I can take into account any feedback from #1370

JamesKingdon:
I changed the name of the option to take into account the comments from Filip and #1370, and rebased to resolve a conflict on the choice of bit to represent the option.

vijaysun-omr:
"aggressiveLoopVersionerOpts" is not the name of the option but is mentioned in the PR description ?

PR #1371, opened 2017-07-18
Revert "Link and compile with -pthread"

youngar:
Reverts eclipse/omr#1352

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -77,8 +77,8 @@ endif()
 if(OMR_HOST_OS STREQUAL "linux")
 	add_definitions(
 		-DLINUX
+		-D_REENTRANT
 		-D_FILE_OFFSET_BITS=64
-		-pthread
 	)
 endif()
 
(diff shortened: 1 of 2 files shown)


charliegracie:
This is just reverting the previous change using the "revert PR" button so I am going to accept it without waiting for the build to complete.

charliegracie:
The IP Test will always fail on the revert PRs so it is not expected to pass

charliegracie:
eclipse/omr#1352 was breaking CMake builds.  Cmake builds are currently not make as required to pass so it was not noticed that they were broken.  This PR reverts the PR which was causing CMake failures and the original PR is being investigated by the committer to resolve the issue as it was CMake specific.

charliegracie:
Here is a direct link to the failing travis build:
https://travis-ci.org/eclipse/omr/jobs/254967590

rwy0717:
The issue was that `-pthread` wasn’t in the link flags, causing missing symbols.`add_definitions()` only affects the compile flags.

Issue #1278, opened 2017-06-20
Segfault when passing `--gtest-filter="*"` to omrgctest


youngar:
When running omrgctest with a test filter, the test segfaults while parsing command line arguments.  I would have expected it to run only the tests which match the filter.
```
./omrgctest --gtest_filter="*" -configListFile=fvtest/gctest/configuration/fvConfigListFile.txt

[==========] Running 8 tests from 1 test case.
[----------] Global test environment set-up.
fish: “./omrgctest --gtest_filter="*"…” terminated by signal SIGSEGV (Address boundary error)
```

```
(gdb) bt
#0  0x00007ffff6c95597 in __strncmp_sse42 () from /usr/lib/libc.so.6
#1  0x000000000041689c in BaseEnvironment::SetUp (this=0x860000) at ../util/testEnvironment.hpp:78
#2  0x0000000000426473 in testing::internal::SetUpEnvironment (env=0x860000) at ../../third_party/gtest-1.7.0/src/gtest.cc:4212
#3  0x000000000043da98 in std::for_each<__gnu_cxx::__normal_iterator<testing::Environment* const*, std::vector<testing::Environment*, std::allocator<testing::Environment*> > >, void (*)(testing::Environment*)> (__first=0x860000, __last=0x545345545f485441, __f=0x426450 <testing::internal::SetUpEnvironment(testing::Environment*)>) at /usr/include/c++/7.1.1/bits/stl_algo.h:3884
#4  0x0000000000438e6a in testing::internal::ForEach<std::vector<testing::Environment*, std::allocator<testing::Environment*> >, void (*)(testing::Environment*)> (c=std::vector of length 1, capacity 1 = {...}, functor=0x426450 <testing::internal::SetUpEnvironment(testing::Environment*)>) at ../../third_party/gtest-1.7.0/src/gtest-internal-inl.h:296
#5  0x00000000004266fb in testing::internal::UnitTestImpl::RunAllTests (this=0x85fc20) at ../../third_party/gtest-1.7.0/src/gtest.cc:4309
[...]

PR #1375, opened 2017-07-18
Fix segfault when running omrgctest with options

LinaSerry:
When initGoogle is called it will remove the arguments it recognizes from
argv and decrements argc accordingly.

The seg fault was occurring because argv was being updated while argc was not. This is because argc is used to initialize _argc before initGoogle is called. Generally initGoogle should be the first thing that is called.

Instead of fetching the testcases from the config file passed into the command line, the test now already knows which testcases it will be running. To choose which set of sets to run the --gtest_filter="" option can be used.

Issue:#1278


fvtest/gctest/main.cpp
```diff
@@ -30,17 +30,14 @@ testMain(int argc, char **argv, char **envp)
...
-	gcTestEnv = (GCTestEnvironment *)testing::AddGlobalTestEnvironment(new GCTestEnvironment(argc, argv));
-	/* This must be called before InitGoogleTest. Because we need to initialize the parameter vector before InitGoogleTest
-	 * calls INSTANTIATE_TEST_CASE_P and takes the vector. */
-	gcTestEnv->GCTestSetUp();
-
 	/*
 	 * ASSERT_NO_FATAL_FAILURE() cannot be used in non-void function. Therefore,
 	 * we need to use HasFatalFailure() to check for fatal failures in sub-routine.
 	 */
 	if (!testing::Test::HasFatalFailure()) {
```
charliegracie: Do we still need this check?  I am not sure there is any chance for failure before this point anymore

perftest/omrperftest.mk
```diff
@@ -39,7 +39,7 @@ endif
 all: test
 	
 omr_perfgctest:
-	./omrgctest -configListFile=perftest/gctest/configuration/perfConfigListFile.txt -keepVerboseLog
+	./omrgctest --gtest_filter="perfTest*" -keepVerboseLog
 	./omrperfgctest
```
charliegracie: Should the perf test be filtering the functional tests?

charliegracie:
@renfeiw Are you ok with these changes?

perftest/omrperftest.mk
LinaSerry: the perf test was only running the tests in perfConfigListFile, that's why it filters the functional tests. @renfeiw what do you think?

renfeiw:
I am ok with the change. We can use --gtest_filter to select what test we want to run. The only different is that the list is not provided during runtime now. During debugging, we now need to recompile gctest after any change is made to the list, which I think should be fine.

fvtest/gctest/GCConfigTest.cpp
```diff
@@ -37,6 +37,17 @@
 #define STRINGFY(str) DO_STRINGFY(str)
 #define DO_STRINGFY(str) #str
 
+const char *gcTests[] = {"fvtest/gctest/configuration/sample_GC_config.xml",
```
ktbriggs: I don't see why this is necessary -- the config files allow tests to be added/excluded using a text editor and without recompiling, which is useful in debugging scenarios. we can live with this but why not just initGoogle() appropriately as per commit comment and google test prescribed usage?

LinaSerry: because if we pass in the config file in the command line, we will have to parse the config file before calling initGoogle() so we could pass it into  INSTANTIATE_TEST_CASE_P and we don't want to be doing anything before initGoogle() . Or is there another way @ktbriggs ?

charliegracie: Are you worried about adding new tests / removing tests during debugging?  If it is removing tests you can just change the filter so it is not included.  If it is adding tests then yes we have to recompile but it is better than the alternative which seems buggy and error prone.

Leonardo2718: A quick addition to the discussion, if I may: it's generally considered bad practice to do anything complicated before `initGoogleTest()` because it can "obscure" the test environment. This makes it much more difficult for someone who is *not already familiar* with the tests  (e.g. me :slightly_smiling_face:) to figure out what's going on.

As @charliegracie pointed out, removing tests when debugging can be accomplished easily enough using a filter. Although adding new tests does require re-compilation, I would argue that the tests should be added once and then stay there (maybe marked as `DISABLED`, if it doesn't make sense to run them unless the user explicitly asks for them). Some accompanying documentation of what's special about these tests and how they're useful when debugging would also be tremendously helpful (especially for us newbies :sweat_smile:).

PR #1378, opened 2017-07-18
Remove the use of colons in feGetEnv

dsouzai:
The colon character cannot be used as part of a name in an export due to
its special treatment to separate exported directories. We must revert
back to using underscores.


dsouzai:
@fjeremic

PR #1379, opened 2017-07-18
Use `-pthreads` on linux, osx

rwy0717:


charliegracie:
With this change will we still need to specify the re-entrant compile flag?

rwy0717:
We won't, as long as every target links against `Threads::Threads` (via `j9thrstatic` or otherwise).

rwy0717:
I've updated this pull request to use the `-pthread` flag. We've decided not to use the built-in Threads package for now, in the interest of keeping things simple and predictable. In the future, Threads may be the best cross-platform option.

Here, we use `-pthread` to define the `_REENTRANT` flag when compiling. `-pthread` is applied globally.

We also use the `-pthread` flag to link `libpthread`. Only libraries that link against `j9thrstatic` (directly or indirectly) will have the `-pthread` flag on the link command. `j9thrstatic` seems to be a sufficient indicator of threading in the GC/VM code. This was the missing step in #1352.

charliegracie:
@dnakamura and @mgaudet can either of you give this a thumbs up?  thanks

dnakamura:
lgtm

PR #1380, opened 2017-07-19
Fix integer cast in omrerror.c

ktbriggs:
Port library method str_vprintf() requires a uintptr_t value for buffer
length and returns uintptr_t, return value was being assigned to
uint32_t value that is later assigned to a uint32_t-valued
field. 

Buffer length arg for str_vprintf() is not used when buffer pointer is
NULL. Replaced value for buffer length arg with 0 (was (unit32_t)-1).

Method will now reuse existing buffer if reallocation fails or
allocation size is >= 2^32. Error message will be truncated and null
terminated if too long to fit in current buffer.

Additional related fixes to use uintptr_t consistently for buffer sizes
and to fix inconsistent use of str_vprintf() args and return value.

Unexpected problem found in Port.cpp. OSX needs sys/statfs.h. I don't
know how it ever got into master without including it. 

### DIFF
--- a/port/common/j9nls.c
+++ b/port/common/j9nls.c
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 1991, 2015
+ * (c) Copyright IBM Corp. 1991, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -189,10 +189,10 @@ j9nls_get_variant(struct OMRPortLibrary *portLibrary)
 static void
 writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *format, va_list args)
 {
-	char outputBuffer[512];
-	char *allocatedBuffer;
-	uintptr_t numberWritten, length;
-	va_list copyOfArgs;
+	va_list argsCopy;
+	char localBuffer[256];
+	char *writeBuffer = NULL;
+	uintptr_t bufferSize = 0;
 
 	BOOLEAN options_info = ((PPG_syslog_flags & J9NLS_INFO) == J9NLS_INFO);
 	BOOLEAN options_warn = ((PPG_syslog_flags & J9NLS_WARNING) == J9NLS_WARNING);
@@ -234,34 +234,32 @@ writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *for
 		return;
 	}
 
-	/* Attempt to write string to on-stack buffer */
-	COPY_VA_LIST(copyOfArgs, args);
-	numberWritten = portLibrary->str_vprintf(portLibrary, outputBuffer, sizeof(outputBuffer), format, copyOfArgs);
+	/* What is size of buffer required ? str_vprintf(..,NULL,..) result includes the null terminator */
+	COPY_VA_LIST(argsCopy, args);
+	bufferSize = portLibrary->str_vprintf(portLibrary, NULL, 0, format, argsCopy);
 
-	/* str_vprintf returns number characters written excluding the null terminator */
-	if (sizeof(outputBuffer) > (numberWritten + 1)) {
-		portLibrary->syslog_write(portLibrary, effectiveSyslogFlag, outputBuffer);
-		return;
+	/* use local buffer if possible, allocate a buffer from system memory if local buffer not large enough */
+	if (sizeof(localBuffer) >= bufferSize) {
+		writeBuffer = localBuffer;
+	} else {
+		writeBuffer = portLibrary->mem_allocate_memory(portLibrary, bufferSize, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
 	}
 
-	/* Didn't fit into outputBuffer - find out how long the buffer needs to be */
-	COPY_VA_LIST(copyOfArgs, args);
-	length = portLibrary->str_vprintf(portLibrary, NULL, (uint32_t)(-1), format, copyOfArgs);
-	length += 1; /* For the \0 */
-
-	/* Allocate buffer from the heap */
-	allocatedBuffer = portLibrary->mem_allocate_memory(portLibrary, length, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
-	if (NULL == allocatedBuffer) {
-		/* Might as well write the truncated string out */
-		portLibrary->syslog_write(portLibrary, effectiveSyslogFlag, outputBuffer);
+	/* format and write out the buffer (truncate into local buffer as last resort) */
+	COPY_VA_LIST(argsCopy, args);
+	if (NULL != writeBuffer) {
+		portLibrary->str_vprintf(portLibrary, writeBuffer, bufferSize, format, argsCopy);
+		portLibrary->syslog_write(portLibrary, effectiveSyslogFlag, writeBuffer);
+		/* dispose of buffer if not on local */
+		if (writeBuffer != localBuffer) {
+			portLibrary->mem_free_memory(portLibrary, writeBuffer);
+		}
+	} else {
+		portLibrary->str_vprintf(portLibrary, localBuffer, sizeof(localBuffer), format, argsCopy);
+		localBuffer[sizeof(localBuffer) - 1] = '\0';
+		portLibrary->syslog_write(portLibrary, effectiveSyslogFlag, localBuffer);
 		return;
 	}
-
-	portLibrary->str_vprintf(portLibrary, allocatedBuffer, length, format, args);
-
-	portLibrary->syslog_write(portLibrary, effectiveSyslogFlag, allocatedBuffer);
-
-	portLibrary->mem_free_memory(portLibrary, allocatedBuffer);
 }
 
 /**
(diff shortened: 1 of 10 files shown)


port/common/omrfile.c
```diff
@@ -630,7 +630,7 @@ omrfile_vprintf(struct OMRPortLibrary *portLibrary, intptr_t fd, const char *for
 	COPY_VA_LIST(copyOfArgs, args);
 
 	/* What is size of buffer required ? Does not include the \0 */
```
charliegracie: I thought the API did return the size including the `\0`?  Can you verify whether or not it does?  Thanks

port/common/j9nls.c
```diff
@@ -234,34 +229,33 @@ writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *for
 		return;
 	}
 
-	/* Attempt to write string to on-stack buffer */
-	COPY_VA_LIST(copyOfArgs, args);
-	numberWritten = portLibrary->str_vprintf(portLibrary, outputBuffer, sizeof(outputBuffer), format, copyOfArgs);
+	char localBuffer[256];
+	char *writeBuffer = NULL;
```
charliegracie: This is not valid C code.  "For C89, you must declare all of your variables at the beginning of a scope block"

tools/tracegen/Port.cpp
```diff
@@ -30,8 +30,10 @@
 #define J9FILE_UNC_EXTENDED_LENGTH_PREFIX (L"\\\\?\\")
 #else /* defined(WIN32) */
 #include <sys/types.h>
-#if !defined(OMRZTPF)
+#if defined(AIXPPC)
 #include <sys/statvfs.h>
+#elif (defined(LINUX) && !defined(J9ZTPF)) || defined(OSX)
```
charliegracie: Why are you making this change?  All of the builds were passing.  Did the compiler recently change on the travis builds? Other builds than ran today passed on OSX.

port/common/omrfile.c
```diff
@@ -607,41 +607,33 @@ EsTranslateOpenFlags(int32_t flags)
 void
 omrfile_vprintf(struct OMRPortLibrary *portLibrary, intptr_t fd, const char *format, va_list args)
 {
-	char outputBuffer[256];
-	char *allocatedBuffer;
-	uint32_t numberWritten;
-	va_list copyOfArgs;
-
-	/* Attempt to write output to stack buffer */
-	COPY_VA_LIST(copyOfArgs, args);
```
charliegracie: Why did you remove the copy of the args?

tools/tracegen/Port.cpp
ktbriggs: OSX needs sys/statfs.h. Compile failed for Travis build on my fork, but passed on eclipse/omr validation build for same push. Scratching head, but code inspection clearly shows problems with statvfs and statfs usage vis-a-vis includes and build flags. I believe my changes are correct.

Only explanation I can think of is that OSX builds for eclipse/omr validation include sys/statfs.h indirectly, but for briggrs/omr they do not.

port/common/omrfile.c
ktbriggs: I don't believe it is required. I can't find implementation for str_vprintf(), but I think it handles va_list.

tools/tracegen/Port.cpp
charliegracie: That code has not really changed since the initial commit.  The zptf changes recently would not have changed anything for OSX.  I am wary of changes we do not fully understand

port/common/j9nls.c
ktbriggs: Fixed in most recent commit.

tools/tracegen/Port.cpp
ktbriggs: Port.cpp changes backed out in most recent commit.

port/win32_include/omrportptb.h
```diff
@@ -39,6 +39,7 @@
 #include "omrport.h"
 
 #define J9ERROR_DEFAULT_BUFFER_SIZE 256 /**< default customized error message size if we need to create one */
+#define J9ERROR_MAXIMUM_BUFFER_SIZE 0xFFFFFFFF /**< maximum customized error message size if we need to create one */
```
charliegracie: Is this maximum buffer size correct for both 32 and 64 bit platforms? Doesn't the stdint library provide a max for unintptr_t?

port/unix_include/omrportptb.h
```diff
@@ -41,6 +41,7 @@
 #include "omriconvhelpers.h"
 
 #define J9ERROR_DEFAULT_BUFFER_SIZE 256 /**< default customized error message size if we need to create one */
+#define J9ERROR_MAXIMUM_BUFFER_SIZE 0xFFFFFFFF /**< maximum customized error message size if we need to create one */
```
charliegracie: Is this maximum buffer size correct for both 32 and 64 bit platforms? Doesn't the stdint library provide a max for unintptr_t (UINTPTR_MAX)?

port/include/omrportptb.h
```diff
@@ -39,6 +39,7 @@
 #include "omrport.h"
 
 #define J9ERROR_DEFAULT_BUFFER_SIZE 256 /**< default customized error message size if we need to create one */
+#define J9ERROR_MAXIMUM_BUFFER_SIZE 0xFFFFFFFF /**< maximum customized error message size if we need to create one */
```
charliegracie: Is this maximum buffer size correct for both 32 and 64 bit platforms? Doesn't the stdint library provide a max for unintptr_t (UINTPTR_MAX)?

port/common/omrfile.c
charliegracie: You are correct it takes a va_list but as you consume args from a va_list it moves the list along and you can not get back to the beginning again.  If there are multiple calls using the same va_list I believe you will not see any values in the second, third, etc usage of the same va_list.  I believe that is why the tests were failing originally with your changes.

charliegracie: You are correct it takes a va_list but as you consume args from a va_list it moves the list along and you can not get back to the beginning again. If there are multiple calls using the same va_list I believe you will not see any values in the second, third, etc usage of the same va_list. I believe that is why the tests were failing originally with your changes.

port/common/j9nls.c
```diff
@@ -234,34 +232,30 @@ writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *for
 		return;
 	}
 
-	/* Attempt to write string to on-stack buffer */
-	COPY_VA_LIST(copyOfArgs, args);
-	numberWritten = portLibrary->str_vprintf(portLibrary, outputBuffer, sizeof(outputBuffer), format, copyOfArgs);
+	/* What is size of buffer required ? str_vprintf(..,NULL,..) result includes the null terminator */
+	uintptr_t bufferSize = portLibrary->str_vprintf(portLibrary, NULL, 0, format, args);
```
charliegracie: You can not declare the variable here in c89 which is used on some of our platforms

port/common/j9nls.c
```diff
@@ -234,34 +232,30 @@ writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *for
 		return;
 	}
 
-	/* Attempt to write string to on-stack buffer */
-	COPY_VA_LIST(copyOfArgs, args);
```
charliegracie: You will need to bring back all of the COPY_VA_LIST code.  See other comments about this.

port/common/j9nls.c
```diff
@@ -234,34 +232,30 @@ writeSyslog(struct OMRPortLibrary *portLibrary, uintptr_t flags, const char *for
...
-		return;
+	/* use local buffer if possible, allocate a buffer from system memory if local buffer not large enough */
+	if (sizeof(localBuffer) >= bufferSize) {
+		writeBuffer = localBuffer;
+	} else {
+		writeBuffer = portLibrary->mem_allocate_memory(portLibrary, bufferSize, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
 	}
 
-	/* Didn't fit into outputBuffer - find out how long the buffer needs to be */
-	COPY_VA_LIST(copyOfArgs, args);
```
charliegracie: You will need to bring back all of the COPY_VA_LIST code.  See other comments about this.

port/common/omrfile.c
```diff
@@ -607,41 +607,33 @@ EsTranslateOpenFlags(int32_t flags)
...
-	/* str_vprintf always null terminates, returns number characters written excluding the null terminator */
-	if (sizeof(outputBuffer) > (numberWritten + 1)) {
-		/* write out the buffer */
-		portLibrary->file_write_text(portLibrary, fd, outputBuffer, numberWritten);
-		return;
-	}
+	char localBuffer[256];
+	char *writeBuffer = NULL;
 
-	/* Either the buffer was too small, or it was the exact size.  Unfortunately can't tell the difference,
```
charliegracie: You will need to bring back all of the COPY_VA_LIST code.  See other comments about this.

port/common/omrfile.c
```diff
@@ -607,41 +607,33 @@ EsTranslateOpenFlags(int32_t flags)
...
-	 * need to determine the size of the buffer (another call to str_vprintf) then print to the buffer,
-	 * a third call to str_vprintf
-	 */
-	COPY_VA_LIST(copyOfArgs, args);
+	/* What is size of buffer required ? str_vprintf(..,NULL,..) result includes the null terminator */
+	uintptr_t bufferSize = portLibrary->str_vprintf(portLibrary, NULL, 0, format, args);
```
charliegracie: When the COPY_VA_LIST is brought back you will need to move the declaration of this variable up.

port/win32_include/omrportptb.h
ktbriggs: This guard is meant to catch wildly invalid results from str_vprintf(.., NULL, ...) calls on 64-bit platforms. It limits range to 32-bits (unsigned), so works for both32- and 64-bit platforms. Even 2^32 is too large to used as a reasonable limit but this will catch negative values etc. It is harmless to leave in although it would never trap an error unless str_vprintf() was badly broken.

port/unix_include/omrportptb.h
ktbriggs: As above.

port/include/omrportptb.h
ktbriggs: As above.

port/common/j9nls.c
ktbriggs: All changes requested pushed in latest commit (also, copyright updates to touched files).

PR #1384, opened 2017-07-19
z/TPF omr dump updates

pelemie:
Cleaned up our opensource version of sys/ucontext.h to use unmodified
definitions. The omr/port/ztpf file updates support this change.

Signed-off-by: pelemie <lemie@us.ibm.com>

### DIFF
--- a/compiler/env/defines.h
+++ b/compiler/env/defines.h
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -14,6 +14,7 @@
  *
  * Contributors:
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *    Multiple authors (IBM Corp.) - z/TPF platform initial port to OMR environment
  ******************************************************************************/
 
 
@@ -82,11 +83,13 @@
 #endif
 
 /*
-  Standarize the OS macros Use HOST_OS instead of various different
+  Standardize the OS macros Use HOST_OS instead of various different
   ways of checking for a particular host os.  As a reference, see
   http://sourceforge.net/p/predef/wiki/OperatingSystems/
 */
 #if defined(__linux__)
+#  define HOST_OS OMR_LINUX
+#elif defined(_TPF_SOURCE)
 #  define HOST_OS OMR_LINUX
 #elif defined(_AIX)
 #  define HOST_OS OMR_AIX
--- a/compiler/z/codegen/SystemLinkage.cpp
+++ b/compiler/z/codegen/SystemLinkage.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -14,6 +14,7 @@
  *
  * Contributors:
  *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *    Multiple authors (IBM Corp.) - z/TPF platform initial port to OMR environment
  *******************************************************************************/
 
 // See also S390Linkage.cpp which contains more S390 Linkage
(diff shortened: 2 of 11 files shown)


port/ztpf/safe_storage.h
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+ * Licensed Materials - Property of IBM
+ * "Restricted Materials of IBM"
+ *
+ * (c) Copyright IBM Corp. 2003, 2017 All Rights Reserved
+ *
+ * US Government Users Restricted Rights - Use, duplication or disclosure
+ * restricted by GSA ADP Schedule Contract with IBM Corp.
+ *******************************************************************************/
```
fjeremic: I think the Eclipse licensing is missing here.

charliegracie: You will need to update the license and amend your commit and force push it to the repo.

pelemie: Thanks guys, sorry for the n00b err0r.

pelemie: @charliegracie I added a change to omrthread.c to this pull request (avoid setjmp call in thread_wrapper() if platform is linux and not z/tpf).

charliegracie:
Hi,  Can you squash the copyright commit into the original commit so that when this PR is accepted there is no point in history where the file has the wrong copyright.  Thanks

pelemie:
completed (my first) squash.

charliegracie:
Sorry on the squash you lost all of the commit message details about what the commit is doing.  Could you update it one more time to include the details of the change.  Similar to the text in the PR description.  Thanks

pelemie:
@charliegracie  I'm not sure if I did what you asked for; and, it seems, I did it it twice.

charliegracie:
Almost.  If you do an interactive rebase you can modify the commit message.  You need to make sure the first line is 50 characters or less and then you only need one signed-off tag on the last line.

pelemie:
@charliegracie  Thank you for the advice! It looks better now, even to me.

fjeremic:
> @charliegracie Thank you for the advice! It looks better now, even to me.

Thanks @pelemie. One last thing. Sorry for the nitpicks. Please fix the commit header for consistency sake:

> ... The first line should be less than 50 characters, sentence case, and **does not end in a period**.

pelemie:
Let's try this again. Included two additional files and made sure the copyright dates were correct (not all of them were?)

pelemie:
@charliegracie  @fjeremic  Comments should be up to date

charliegracie:
@vijaysun-omr or @0xdaryl  or @mstoodle  Can you take a quick look at the compile changes?  They look ok to me but I want a compiler expert to review them as well.

vijaysun-omr:
I would defer to @fjeremic or @joransiu for z codegen changes.

compiler/env/defines.h
```diff
@@ -82,12 +83,14 @@
...
 
 /*
-  Standarize the OS macros Use HOST_OS instead of various different
+  Standardize the OS macros Use HOST_OS instead of various different
   ways of checking for a particular host os.  As a reference, see
   http://sourceforge.net/p/predef/wiki/OperatingSystems/
 */
 #if defined(__linux__)
 #  define HOST_OS OMR_LINUX
+#elif defined(_TPF_SOURCE)
```
fjeremic: Can this be folded as an `|| defined(_TPF_SOURCE)` with `#if defined(__linux__)`?

compiler/z/codegen/SystemLinkage.cpp
```diff
@@ -931,10 +932,21 @@ TR::S390zLinuxSystemLinkage::S390zLinuxSystemLinkage(TR::CodeGenerator * codeGen
...
+
+#if defined(OMRZTPF)
+/*
+ * z/TPF os is a 64 bit target
+ */
+   setOffsetToRegSaveArea     (16);    // z/TPF is a 64 bit target only
+   setGPRSaveAreaBeginOffset  (16);    //x'10'  see ICST_R2 in tpf/icstk.h
+   setGPRSaveAreaEndOffset    (160);   //x'a0'  see ICST_PAT in tpf/icstk.h
+   setOffsetToFirstParm       (448);   //x'1c0' see ICST_PAR in tpf/icstk.h
+#else
```
fjeremic: Why can't z/TPF define `TR_TARGET_64BIT`? This code does not need to exist. See uses of `OMR::Environment::setBitness` and how it is set.

port/ztpf/omrintrospect.c
```diff
@@ -753,122 +749,11 @@ count_threads( struct PlatformWalkData *data ) {
...
-			break;
-		}
-	} while (!timedOut(data->state->deadline1));
-
-	if (timedOut(data->state->deadline1)) {
-		data->threadsOutstanding = data->threadCount - 1;
-		return -5;
-	}
-
-	return data->threadCount -1;
```
fjeremic: This seems to be deleting a whole whack of code. This does not look correct. @charliegracie PTAL.

compiler/z/codegen/SystemLinkage.cpp
jjohnst-us: @fjeremic just in case you weren't aware... z/TPF needs to push at least 448 byte stack frames.  Anything smaller then that will corrupt our linkage code.

compiler/env/defines.h
pelemie: I will do that, I just wanted to maintain the style that used a stream of #elif's.

compiler/z/codegen/SystemLinkage.cpp
pelemie: Thanks @jjohnst-us . I will update the comment to note the difference in stack sizes between z/tpf and z/linux and remove the first line of code; z/tpf can and does use the 64bit setting so no need to have that. The other three lines use values that directly reference our stack and need to be there or z/tpf can't start a jvm properly.

port/ztpf/omrintrospect.c
pelemie: our 28 version of j9introspect has the following code:
#if defined(J9ZTPF)
    errno = EBADF;
    return -1;
#endif
In other words, we aren't implementing this function.  We could leave that code there even though we aren't executing it.  I suppose, too, we can change the callers to avoid this function; it seems better to update the function itself.

compiler/z/codegen/SystemLinkage.cpp
fjeremic: Apologies, you're right! I did not notice the changed offsets at first. Those are good updates @pelemie.

port/ztpf/omrintrospect.c
fjeremic: I see. The function is implemented in /port/ztpf. Is it safe to say no other callers will access this? I incorrectly assumed `suspend_all_preemptive` is a common function to other backends and removing this code would be incorrect, but it does not appear so if z/TPF is the only caller of this function.

pelemie: suspend_all_preemptive is only called in port/ztpf/omrintrospect.c

charliegracie:
Thanks for the reviews everyone.  Merging now :)

PR #1385, opened 2017-07-20
Fix test for `str_ftime()` in Singapore time zone

pparkkin:
Test for `str_ftime()` fails when run in Singapore time zone.

The test expects all time zone offsets to UTC at Unix timestamp `0` to
be full hours. In Singapore at Unix timestamp `0` the offset was 7
hours and 30 minutes.

The change simply makes it so the test ignores differences in minutes
between UTC and local time in addition to ignoring differences in
hours.

Signed-off-by: Paavo Parkkinen <pparkkin@gmail.com>


charliegracie:
Thanks :)

PR #1388, opened 2017-07-20
[Mem8] Use TR_BitVector in LocalDSE

andrewcraik:
LocalDSE is currently using a shared bit vector for internal liveness
tracking. These bit vectors leak into the heap and consume large amounts
of memory. This change converts the optimization to use a stack
allocated TR_BitVector to save space.


andrewcraik:
@vijaysun-omr ping?

vijaysun-omr:
Looks straightforward. Merging.

PR #1389, opened 2017-07-20
Fix str_vprintf() usage in port library

ktbriggs:
Usage of this method is incorrect in several port methods. This commit
fixes usage and simplifies affected methods somewhat.

### DIFF
--- a/fvtest/porttest/omrstrTest.cpp
+++ b/fvtest/porttest/omrstrTest.cpp
@@ -179,46 +179,33 @@ test_omrstr_vprintf(struct OMRPortLibrary *portLibrary, const char *testName, co
 	/* Buffer larger than required */
 	va_start(args, expectedResult);
 	validate_omrstr_vprintf(OMRPORTLIB, testName, actualResult, sizeof(actualResult), expectedResult, format, args);
-	va_end(args);
 
 	/* Exact size buffer (+1 for NUL)*/
-	va_start(args, expectedResult);
 	validate_omrstr_vprintf(OMRPORTLIB, testName, actualResult, strlen(expectedResult) + 1, expectedResult, format, args);
-	va_end(args);
 
 	/* Truncated buffer - use strlen(expectedResult) as length of buffer thus reducing the size by 1*/
 	if (strlen(expectedResult) > 1) {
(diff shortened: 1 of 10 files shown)


youngar:
I'm not sure why the appveyer build failed, it's possible github was slow to create the merge commit. Reading online it looks like the build can be restarted by opening and closing this PR.

fvtest/porttest/omrstrTest.cpp
```diff
@@ -179,46 +179,33 @@ test_omrstr_vprintf(struct OMRPortLibrary *portLibrary, const char *testName, co
 	/* Buffer larger than required */
 	va_start(args, expectedResult);
```
charliegracie: Should you have removed this va_start?

fvtest/porttest/testHelpers.cpp
```diff
@@ -240,18 +240,16 @@ outputErrorMessage(struct OMRPortLibrary *portLibrary, const char *fileName, int
 	va_start(args, format);
```
charliegracie: If the va_start/end moved into the callee then shouldn't we remove all usage inside the caller?

fvtest/porttest/omrstrTest.cpp
ktbriggs: va_list variables are effectively const in methods that instantiate them -- these guys just need to call va_start() once and balance it with va_end() after use. So it can be used multiple times in declaring method and in called methods that do not call va_arg() to iterate the contained vars. 

This is so as long as called methods use va_arg() to obtain vars call va_copy() and iterate the copy, leaving the caller's original instance untouched.

fvtest/porttest/testHelpers.cpp
ktbriggs: Not sure I understand your question. The va_start() and va_end() calls are made only by variadic methods, which encapsulate variable arg list in va_list that is then passed down the call chain until some receiving method needs to access the contained vars. The receiving method must call va_copy() and iterate over the copy to obtain vars, leaving the original alone.

charliegracie: Ok makes sense

Issue #1279, opened 2017-06-20
The maxprec() call to acquire the current time on Z lacks precision


last5bits:
Starting from the Z10 generation,  Z machines have sub-nanosecond precision. However, the `maxprec()` call on z/OS and zLinux does not use this capability.

The `maxprec()` call is implemented with `STCKF`: the instruction that stores the Time-of-Day (TOD) data in the supplied register. The data is a 64-bit unix-time number, in which the bit at the position 51 changes every microsecond.

The current implementation for `maxprec()` adheres to the following:

1. Get the current time by using the STCKF instruction
1. Shift this number by 9 bit to the right, effectively using only bits 0--54 of the initial number; thus, decreasing the precision to 1/8 of a microsecond.

When e.g. the port library call to `omr_time_nano_time()` calculates the time in nanosecods, it merely multiplies the result of `maxprec()` by 125. This precision can be improved since, starting from Z10, the bit at the position 62 is incremented, which gives the precision of approximately 0.5 of a nanosecond.

This problem is to be fixed in both the port library and in the jit-compiler.

last5bits:
Relevant issue: in port library, `omrtime_hires_delta()` employs casts to `double` that also can decrease precision of the result.

PR #1390, opened 2017-07-20
Z nanotime precision

last5bits:
Previously, the high-resolution timer was truncated to 1/8 of a microsec and then multiplied by a constant to e.g. obtain the current time in nanoseconds.

Now, we obtain high-resolution timer that has the precision of approximately 0.5 of a nanosecond and adjusts it to produce e.g. the current time in nanoseconds without losing precision.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -646,12 +646,6 @@ class OMR_EXTENSIBLE CodeGenerator
 
    uint32_t getMaxObjectSizeGuaranteedNotToOverflow() { return _maxObjectSizeGuaranteedNotToOverflow; }
 
-   // high precision timer can be converted into millisecond timer or nanosecond timer through appropriate
-   // division or multiplication (respectively). For example, 390 high precision timer provides accuracy to
-   // 1/8 of a microsecond so converters are 8000 and 125 respectively (multiplying the two should give you 1000000)
-   int64_t getCurrentTimeMillisDivisor() { return 0; } // no virt
-   int64_t getNanoTimeMultiplier() { return 0; } // no virt
-
    // --------------------------------------------------------------------------
    // OSR, not code generator
    //
--- a/port/linuxs390/omrtime.c
+++ b/port/linuxs390/omrtime.c
@@ -30,7 +30,6 @@
 #include "omrportpg.h"
 #include "omrport.h"
 
-#define OMRTIME_HIRES_CLOCK_FREQUENCY J9CONST_U64(8000000)
 #define OMRTIME_CLOCK_DELTA_ADJUSTMENT_INTERVAL_USEC J9CONST_I64(60 * 1000 * 1000)
 
 #define OMRTIME_NANOSECONDS_PER_SECOND J9CONST_I64(1000000000)
@@ -40,7 +39,23 @@ extern int64_t maxprec();
 
 static void omrtime_calculate_hw_time_delta(struct OMRPortLibrary *portLibrary);
 
-
+/**
+ * Multiply a number by a fraction
+ *
+ * @param[in] num The number to be multiplied by the fraction
+ * @param[in] numerator The numerator of the fraction
+ * @param[in] denominator The denominator of the fraction
+ *
+ * @return The result of the multiplication
+ */
+static int64_t
+muldiv64(const int64_t num, const int64_t numerator, const int64_t denominator)
+{
+	const int64_t quotient = num / denominator;
+	const int64_t remainder = num - quotient * denominator;
+	const int64_t res = quotient * numerator + ((remainder * numerator) / denominator);
+	return res;
+}
 
 /**
  * Query OS for timestamp.
@@ -55,7 +70,7 @@ uintptr_t
 omrtime_msec_clock(struct OMRPortLibrary *portLibrary)
 {
 	int64_t msec;
-	int64_t usec = maxprec() / 8;
+	int64_t usec = maxprec() / OMRPORT_TIME_HIRES_MICROTIME_DIVISOR;
 
 	msec = usec / 1000;
 
@@ -82,7 +97,7 @@ omrtime_msec_clock(struct OMRPortLibrary *portLibrary)
 uintptr_t
 omrtime_usec_clock(struct OMRPortLibrary *portLibrary)
 {
-	int64_t usec = maxprec() / 8;
+	int64_t usec = maxprec() / OMRPORT_TIME_HIRES_MICROTIME_DIVISOR;
 
 	if ((usec - PPG_last_clock_delta_update > OMRTIME_CLOCK_DELTA_ADJUSTMENT_INTERVAL_USEC)
 	 || (usec < PPG_last_clock_delta_update)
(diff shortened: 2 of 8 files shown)


last5bits:
Issue #1279

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -648,9 +648,11 @@ class OMR_EXTENSIBLE CodeGenerator
 
    // high precision timer can be converted into millisecond timer or nanosecond timer through appropriate
    // division or multiplication (respectively). For example, 390 high precision timer provides accuracy to
-   // 1/8 of a microsecond so converters are 8000 and 125 respectively (multiplying the two should give you 1000000)
+   // 1/2048 of a microsecond so, to convert maximum-precision timer to
+   // - milliseconds, divide maximum-precision timer by 2,048,000
+   // - nanoseconds, multiply maximum-precision timer by 125/256
    int64_t getCurrentTimeMillisDivisor() { return 0; } // no virt
-   int64_t getNanoTimeMultiplier() { return 0; } // no virt
```
fjeremic: I don't see any uses of these two APIs within OMR. Would these be good candidates for removal or upstreaming to the project making use of these APIs?

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -648,9 +648,11 @@ class OMR_EXTENSIBLE CodeGenerator
...
+   std::pair<int64_t, int64_t> getNanoTimeMultiplierFraction() { return std::make_pair(0, 0); } // no virt
```
fjeremic: This new function should be documented. However see previous comment. It sounds like this function belongs in an upstream project.

compiler/codegen/OMRCodeGenerator.hpp
last5bits: There are no uses *now* but I feel like this is platform-specific information and it belongs to OMR. In addition, OMR has other mentions of these constants in port library; I would prefer not to smear this over two projects.

fjeremic: I'm suggesting we _move_ these (and document them) to the upstream project which consumes the APIs, not duplicate them. The names ("currentTimeMillis" and "nanoTime") seem to imply these are Java specific queries and as such IMO they belong in the upstream J9 code generator.

last5bits: Sure, but keep in mind that after moving these constants into the upstream projects, OMR will still have mentions of these constants in port library. We already don't have one single method for getting e.g. `CurrentTimeMillisDivisor`, now we are gonna put this constant into two separate projects.

I'm worried about a scenario in which some updates the constant in OMR without updating it in the upstream project. This can lead to nasty bugs.

charliegracie: They do belong in OMR but they belong in the port library and the compiler should call the port library to make use of them

fjeremic: > They do belong in OMR but they belong in the port library and the compiler should call the port library to make use of them

I agree. @last5bits any thoughts?

last5bits: Sounds good! I'll update the code.

last5bits:
> Some administrative comments only from me for now
@0xdaryl Fixed.

compiler/codegen/OMRCodeGenerator.hpp
last5bits: Fixed.

compiler/codegen/OMRCodeGenerator.hpp
last5bits: Fixed.

port/zos390/omrtime.c
```diff
@@ -63,16 +78,22 @@ omrtime_msec_clock(struct OMRPortLibrary *portLibrary)
...
 omrtime_current_time_nanos(struct OMRPortLibrary *portLibrary, uintptr_t *success)
 {
 	*success = 1;
-	uint64_t nanos = MAXPREC() * 125;
-	return nanos;
+
+	const int64_t subnanosec = MAXPREC();
+
+	// calculate nanosec = subnanosec * NUMERATOR / DENOMINATOR via integer arithmetics
+	const int64_t nanosec = muldiv64(subnanosec,
```
charliegracie: This calculation is much more expensive now than it was I believe.  There are now 7 math operations instead of 1.  If you made the code `uint64_t nanos = MAXPREC() * OMRPORT_TIME_HIRES_NANOTIME_NUMERATOR / OMRPORT_TIME_HIRES_NANOTIME_DENOMINATOR;' some of the math would become a compile time constant and return back to a single math operation

port/zos390/omrtime.c
```diff
@@ -87,15 +108,20 @@ omrtime_current_time_nanos(struct OMRPortLibrary *portLibrary, uintptr_t *succes
...
 }
 
 int64_t
 omrtime_nano_time(struct OMRPortLibrary *portLibrary)
 {
-	int64_t nanos = MAXPREC() * 125;
-	return nanos;
+	const int64_t subnanosec = MAXPREC();
+
+	// calculate nanosec = subnanosec * NUMERATOR / DENOMINATOR via integer arithmetics
```
charliegracie: Same as my previous comment

port/zos390/omrtime.c
```diff
@@ -158,9 +184,10 @@ omrtime_hires_delta(struct OMRPortLibrary *portLibrary, uint64_t startTime, uint
 	if (OMRTIME_HIRES_CLOCK_FREQUENCY == requiredResolution) {
 		/* no conversion necessary */
 	} else if (OMRTIME_HIRES_CLOCK_FREQUENCY < requiredResolution) {
-		ticks = (uint64_t)((double)ticks * ((double)requiredResolution / (double)OMRTIME_HIRES_CLOCK_FREQUENCY));
+		ticks = muldiv64(ticks, requiredResolution, OMRTIME_HIRES_CLOCK_FREQUENCY);
```
charliegracie: Again same as above

port/linuxs390/omrtime.c
```diff
@@ -230,9 +245,10 @@ omrtime_hires_delta(struct OMRPortLibrary *portLibrary, uint64_t startTime, uint
 	if (OMRTIME_HIRES_CLOCK_FREQUENCY == requiredResolution) {
 		/* no conversion necessary */
 	} else if (OMRTIME_HIRES_CLOCK_FREQUENCY < requiredResolution) {
-		ticks = (uint64_t)((double)ticks * ((double)requiredResolution / (double)OMRTIME_HIRES_CLOCK_FREQUENCY));
+		ticks = muldiv64(ticks, requiredResolution, OMRTIME_HIRES_CLOCK_FREQUENCY);
```
charliegracie: Since required resolution has to be one of a few options this code would be modified so that we significantly reduce the cost of calculating the results.  See my comments below about muldiv64 being a much more expensive operation than just using the constants everywhere

port/zos390/omrtime.c
last5bits: Charlie, it *is* more expensive but it is also functionally correct (unlike the previous version that didn't have enough precision).

First, `OMRPORT_TIME_HIRES_NANOTIME_NUMERATOR / OMRPORT_TIME_HIRES_NANOTIME_DENOMINATOR` does not become a compile-time constant (at least, it didn't in my experiments). Even if it was, it would evaluate to `125 / 256 = 0`.

So, the order of the operations is 1) `MAXPREC() * NANOTIME_NUMERATOR` and then 2) `res / NANOTIME_DENUMERATOR`; there is a possibility of overflow in 1).

PR #1391, opened 2017-07-21
Remove optDetailString from OptimizationManager

lmaisons:
Currently the optimization details string used during tracing is
initialized as part of the optimization manager construction. Because
OMR compiler consumers may initialize some optimizations themselves,
this opens us up to a scenario where these strings start to diverge
between the various consuming projects.

As developers rely on the consistency of those strings in order to aid
in cross-project debugging, this change modifies these optimization
string definitions such that they are an inherent part of the class
definition, and are accessed via a virtual function, for which each
implementation returns a statically defined string.

### DIFF
--- a/compiler/optimizer/AsyncCheckInsertion.cpp
+++ b/compiler/optimizer/AsyncCheckInsertion.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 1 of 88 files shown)


0xdaryl:
@xliang6, I'd appreciate a review on this one please.

xliang6:
LGTM. Do you need to change the copyright year to 2017 for these files too?

lmaisons:
Probably :S

lmaisons:
Copyright headers should have been fixed by latest push.

vijaysun-omr:
There was an intent to do coalescing of virtual guards in more cases than a tail split would have been done originally, e.g. nested guards could be "coalesced" was the thinking but that does'nt really legitimize the inconsistency between the naming string and the class name.

andrewcraik:
Fixing the inconsistency doesn't help too much since the opt is in a file called VirtualGuardCoalescer... Perhaps something to tackle in a separate item?

PR #1392, opened 2017-07-21
Improve VirtualGuard canBeRemoved and merged API

ncough:
This change extends VirtualGuard so that guards
can be labelled as unmerged and canBeRemoved can
now be asked to ignore whether guards have been
merged.


andrewcraik:
A sensible improvement - I think the motivating case here is that you may be able to replace a runtime guard with a nopable guard. To do so you would have to see that the guard itself is removable its self, but then also note that it is merged with a non-removable guard and transform the guard appropriately. A useful addition. LGTM.

PR #1393, opened 2017-07-21
Merge OptAttributes into Node

ncough:
OptAttributes hold information about a Node, such
as its reference count and local index. They are held separate
so that the OptAttributes can be freed when no longer needed.
Currently this is limited to before codegen on certain platforms.

However, they are already allocated into the heap region
using a CS2 TableOf and only have a limited chance of their memory
being reused by another CS2 data structure.

Moving these fields into the Node itself with proper ordering
results in only an 8 byte increase to its size and improves
locality, as they are frequently accessed.

### DIFF
--- a/compiler/codegen/CodeGenRA.cpp
+++ b/compiler/codegen/CodeGenRA.cpp
@@ -104,12 +104,9 @@ OMR::CodeGenerator::checkForLiveRegisters(TR_LiveRegisters *liveRegisters)
       traceMsg(self()->comp(), "\n\n");
       for (TR_LiveRegisterInfo *p = liveRegisters->getFirstLiveRegister(); p; p = p->getNext())
          {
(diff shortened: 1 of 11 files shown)


andrewcraik:
I think this is a good change overall - the side data structure doesn't buy much if anything and simplifying the whole node pool construct would seem to be a good thing. @mpirvu you may want to assess this one.

mgaudet:
🎉🎉🎉

0xdaryl:
Any comments @mpirvu?

mpirvu:
If I understand correctly this commit may increase the peak JIT scratch space used by a compilation, but has the potential of speeding up compilations. Both effects are small if anything.
My opinion is that compilation speed trumps compilation memory and if this change improves code robustness and maintainability we should go for it.

PR #1394, opened 2017-07-21
Increase inlining node count for OSR

ncough:
OSR book keeping is represented in the IL. Therefore,
it can significantly increase the number of nodes needed.
To retain the prior inlining performance, the limit based
on node count must be increased under OSR.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -346,6 +346,22 @@ TR_InlinerBase::setInlineThresholds(TR::ResolvedMethodSymbol *callerSymbol)
    if (e)
       _nodeCountThreshold = atoi(e);
 
+   // Under voluntary OSR, OSR blocks and additional stores can
+   // significantly increase the node count, particularly dead stores
+   // in OSRCodeBlocks. To maintain the same inlining behaviour, the
+   // threshold must be increased.
+   //
+   // The compile time overhead of this threshold increase has been tested
+   // and was not significant, as this threshold is typically only reached
+   // in hot or above compiles.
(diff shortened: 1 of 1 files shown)


compiler/optimizer/Inliner.cpp
```diff
@@ -346,6 +346,14 @@ TR_InlinerBase::setInlineThresholds(TR::ResolvedMethodSymbol *callerSymbol)
    if (e)
       _nodeCountThreshold = atoi(e);
 
+   if (comp()->getOption(TR_EnableOSR))
+      {
+      static const char *f = feGetEnv("TR_OSRNodeCountThreshold");
+      if (f)
+         _nodeCountThreshold = atoi(f);
+      else
+         _nodeCountThreshold *= 2;
```
fjeremic: How accurate is this node count estimate? Does OSR introduce twice the number of trees on average?

andrewcraik: Note this is not trees - this is nodes. And empirical evidence so far says this number is correct - we have to generate the OSR catch block, code block, and dead stores so the number can be quite large - especially for small methods.

ncough:
The change has been updated to limit when the threshold is increased, matching conditions in OSRLiveRange analysis. Particularly, the threshold will not be increased for late inlining or under FSD.

vijaysun-omr:
Has this change been tested to see the impact on compile time ?

ncough:
The compile time overhead of this threshold increase has been tested
and was found to not be significant, as this threshold is typically only reached
in hot or above compiles.

andrewcraik:
LGTM from me.

vijaysun-omr:
Thanks for the answer. Merging.

PR #1395, opened 2017-07-21
Set OSR block freq during ColdBlockMarker

ncough:
Setting the frequencies on blocks during IlGen
can prevent the JitProfiler from properly applying
its raw frequencies. OSR blocks, which can be created
during IlGen, have always received a frequency of 0.
This modifies them to receive a frequency of -1, 
which is later set to 0 during ColdBlockMarker.

### DIFF
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -8166,8 +8166,15 @@ TR_ColdBlockMarker::identifyColdBlocks()
    for (TR::AllBlockIterator iter(optimizer()->getMethodSymbol()->getFlowGraph(), comp()); iter.currentBlock(); ++iter)
       {
       TR::Block *block = iter.currentBlock();
+
       if (block->isCold())
+         {
+         // OSR blocks may not have had a chance to set their frequencies yet
+         if (block->isOSRCodeBlock() || block->isOSRCatchBlock())
+            block->setFrequency(UNKNOWN_COLD_BLOCK_COUNT);
+
          foundColdBlocks = true;
+         }
       else
          {
          int32_t coldness = isBlockCold(block);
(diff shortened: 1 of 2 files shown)


compiler/optimizer/LocalOpts.cpp
```diff
@@ -8166,8 +8166,15 @@ TR_ColdBlockMarker::identifyColdBlocks()
    for (TR::AllBlockIterator iter(optimizer()->getMethodSymbol()->getFlowGraph(), comp()); iter.currentBlock(); ++iter)
       {
       TR::Block *block = iter.currentBlock();
+
       if (block->isCold())
+         {
+         // OSR blocks may not have had a chance to set their frequencies yet
+         if (block->isOSRCodeBlock() || block->isOSRCatchBlock())
+            block->setFrequency(0);
```
andrewcraik: Since the cold block frequencies mean something special (ie the source) I think there is a number reserved for ilgen or similar - perhaps that number should be used here.

andrewcraik:
Other than the specific frequency of the cold block, the spirit of this change looks good and it something we should merge sooner rather than later due to the issues bad frequencies can cause.

compiler/optimizer/LocalOpts.cpp
ncough: The only cold block frequency I see being set during IlGen is `UNKNOWN_COLD_BLOCK_COUNT`. `CATCH_COLD_BLOCK_COUNT` may be another choice, but does not seem suitable for the OSRCodeBlock.

andrewcraik: Ok I can live with that I think.

andrewcraik:
@vijaysun-omr over to you.

vijaysun-omr:
Seems to be a safe change in that it allows OSR blocks to participate in JIT frequency setting. Merging.

PR #1410, opened 2017-07-25
Run the unimportantLoopCount at all optlevels

JamesKingdon:
In TR_LoopVersioner::loopIsWorthVersioning we currently only
run the test for unimportantLoopCount for optlevels <= warm. It is
likely to be beneficial in many circumstances to run this test at
all optlevels.

This change is currently guarded by the Xjit option added in
PR #1369 until such time as the behaviour is tested accross a broad
set of applications.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -185,9 +185,14 @@ bool TR_LoopVersioner::loopIsWorthVersioning(TR_RegionStructure *naturalLoop)
    TR::Block *entryBlock = naturalLoop->getEntryBlock();
 
    if (entryBlock->isCold())
+      {
+      if (trace()) traceMsg(comp(), "loopIsWorthVersioning returning false for cold block\n");
       return false;
+      }
 
-   if (comp()->getMethodHotness() <= warm)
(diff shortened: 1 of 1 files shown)


compiler/optimizer/LoopVersioner.cpp
```diff
@@ -187,7 +187,9 @@ bool TR_LoopVersioner::loopIsWorthVersioning(TR_RegionStructure *naturalLoop)
    if (entryBlock->isCold())
       return false;
 
-   if (comp()->getMethodHotness() <= warm)
+   // if aggressive loop versioner flag is set then run at any optlevel, otherwise only at warm or less
+   bool aggressive = TR::Options::getCmdLineOptions()->getOption(TR_EnableAggressiveLoopVersioning);
+   if (aggressive || comp()->getMethodHotness() <= warm)
```
andrewcraik: This probably deserves a trace message to show the decision made - it should be guarded for tracing of the opt in question.

JamesKingdon:
@andrewcraik Additional trace added.

vijaysun-omr:
Change looks okay to me. Could you please re-base and ensure checks pass so that I can consider merging ?

vijaysun-omr:
Checks have passed. Merging.

PR #1411, opened 2017-07-25
Reduce the default lvBlockFreqCutoff value

JamesKingdon:
The current default threshold of 5000 for lvBlockFreqCutoff was
found to be too high for some use-cases. A lower threshold of
500 was introduced, currently subject to
-Xjit:enableAggressiveLoopVersioning
being set until the change is tested across more
applications. The env var TR_LoopVersionerFreqCutoff remains
available to over-ride the default setting.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -208,8 +208,26 @@ bool TR_LoopVersioner::loopIsWorthVersioning(TR_RegionStructure *naturalLoop)
             }
          }
 
+      bool aggressive = TR::Options::getCmdLineOptions()->getOption(TR_EnableAggressiveLoopVersioning);
+
+      int32_t lvBlockFreqCutoff;
       static const char * b = feGetEnv("TR_LoopVersionerFreqCutoff");
-      int32_t lvBlockFreqCutoff = b ? atoi(b) : 5000;
+      if (b) 
+         {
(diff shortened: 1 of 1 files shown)


compiler/optimizer/LoopVersioner.cpp
```diff
@@ -208,8 +208,21 @@ bool TR_LoopVersioner::loopIsWorthVersioning(TR_RegionStructure *naturalLoop)
...
+      if (b) 
+         {
+         lvBlockFreqCutoff = atoi(b);
+         }
+      else if (aggressive)
+         {
+         lvBlockFreqCutoff = 500;
+         }
+      else
+         lvBlockFreqCutoff = 5000;
```
andrewcraik: braces around this one please

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -208,8 +208,21 @@ bool TR_LoopVersioner::loopIsWorthVersioning(TR_RegionStructure *naturalLoop)
...
+
```
andrewcraik: Can we add a trace statement to log the frequency guarded by tracing so that we know what was going on?

JamesKingdon:
@andrewcraik Braces and trace added, PB run

andrewcraik:
LGTM

vijaysun-omr:
Looks okay to me. Merging.

PR #1413, opened 2017-07-25
Only run the unimportant block test at <=warm

JamesKingdon:
It was found that the test for unimportant blocks based on block
frequency in TR_LoopVersioner::detectChecksToBeEliminated was
detrimental to performance at higher optimisation levels. An
additional test was added to restrict the code path to optlevels
warm and lower. The new behaviour is gated by the Xjit option
enableAggressiveLoopVersioning added by PR #1369


vijaysun-omr:
Looks okay to me. Merging.

PR #1416, opened 2017-07-26
[CMAKE] Use -fno-strict-aliasing for GC

mgaudet:
This matches the behaviour in the makefiles, and allows more modern
versions of GCC and clang to pass the GC tests.

### DIFF
--- a/gc/CMakeLists.txt
+++ b/gc/CMakeLists.txt
@@ -266,6 +266,12 @@ add_library(omrgc
 )
 add_dependencies(omrgc omrgc_hookgen)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
+    target_compile_options(omrgc PRIVATE -fno-strict-aliasing)
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "XL|VisualAge")
+    target_compile_options(omrgc PRIVATE -qalias=noansi)
+endif()
+
 target_include_directories(omrgc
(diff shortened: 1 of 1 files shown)


0xdaryl:
Looks fine to me.  But would appreciate a thumbs up from at least one of @rwy0717 or @dnakamura before merging.

0xdaryl:
Actually, I'll defer merging to @charliegracie (GC subject matter expert).

rwy0717:
Unfortunately, this is something that we need.
I think it should be `Clang`, not `CLANG`. Also, missing `AppleClang`.
Since you're using `MATCHES`, IMO it would be better as:
```
if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
```

I think this style is more readable than `STREQUAL`/`OR`.

We should get an issue up about our aliasing.

mgaudet:
Switched to `MATCHES` syntax, and added `AppleClang`

gc/CMakeLists.txt
```diff
@@ -266,6 +266,10 @@ add_library(omrgc
 )
 add_dependencies(omrgc omrgc_hookgen)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
```
charliegracie: Do we not require a similar option for XLC as well?

mgaudet: Good point: Just pushed an update for that.

PR #1418, opened 2017-07-26
[Mem9] Create a memory region to hold structure

andrewcraik:
Currently structure nodes are allocated on the heap. The structure information is regularly invalidated during compilation and rebuilt. Each invalidates structure graph currently leaks until the end of the compilation and can amount to a non-trivial amount of memory. This change creates a TR::Region in the CFG to hold structure so that we can invalidate it.

This change is based on the [Mem8] change so the 'Use TR_BitVector in LocalDSE' change is from that PR.

### DIFF
--- a/compiler/optimizer/StructuralAnalysis.cpp
+++ b/compiler/optimizer/StructuralAnalysis.cpp
@@ -25,9 +25,6 @@
 #include "compile/Compilation.hpp"                    // for Compilation
 #include "control/Options.hpp"
 #include "control/Options_inlines.hpp"
-#include "cs2/arrayof.h"                              // for StaticArrayOf
-#include "cs2/bitvectr.h"
-#include "cs2/sparsrbit.h"
 #include "env/TRMemory.hpp"
 #include "il/Block.hpp"                               // for Block, toBlock
 #include "il/symbol/ResolvedMethodSymbol.hpp"
@@ -58,7 +55,7 @@ void TR_RegionAnalysis::simpleIterator (TR_Stack<int32_t>& workStack,
       // The exit block must only be made part of the region headed by the entry
       // block. It provides an exit for all other regions.
       //
-      if (doThisCheck && next._succ.IsZero() &&
+      if (doThisCheck && next._succ.isEmpty() &&
           next._originalBlock == _cfg->getEnd())
          {
          if (hdrBlock->getNumber() != 0)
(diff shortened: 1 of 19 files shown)


andrewcraik:
There may be a case in the ras code that was missed in this commit where we could read a blcok structureOf pointer when the structure has been invalidated. Will advise once the issue has been assessed. I'd like the review of the changes to continue because it will be a simple case of a missed check to see if the flow graph structure is valid.

andrewcraik:
I think I found fixed the bad structure read in ras/Debug.cpp

andrewcraik:
This change is fine, but I am looking an issue in a downstream project where the project can try to read an invalid pointer from invalidated structure.

0xdaryl:
Do you have any reason to believe the Travis CI failure is related to this PR?

andrewcraik:
No I do not.

compiler/optimizer/StructuralAnalysis.hpp
```diff
@@ -57,10 +57,10 @@ class TR_RegionAnalysis
...
 
    class StructInfo;
-   typedef CS2::TableOf<StructInfo, TR::Allocator> InfoTable;
-   typedef TR::BitVector StructureBitVector;
-   typedef TR::BitVector WorkBitVector;
-   typedef TR::deque<TR_StructureSubGraphNode*> SubGraphNodes;
+   typedef StructInfo** InfoTable;
+   typedef TR_BitVector StructureBitVector;
+   typedef TR_BitVector WorkBitVector;
+   typedef TR::deque<TR_StructureSubGraphNode*, TR::Region&> SubGraphNodes;
```
0xdaryl: Are the `#include "cs2/arrayof.h"` and `#include "cs2/tableof.h"` still required in this file?  Please remove if not.

compiler/optimizer/StructuralAnalysis.cpp
```diff
@@ -58,7 +58,7 @@ void TR_RegionAnalysis::simpleIterator (TR_Stack<int32_t>& workStack,
       // The exit block must only be made part of the region headed by the entry
       // block. It provides an exit for all other regions.
       //
-      if (doThisCheck && next._succ.IsZero() &&
+      if (doThisCheck && next._succ.isEmpty() &&
```
0xdaryl: This comment isn't applicable to this line, but I need a place to put it.  :-)

Are these #includes still relevant in this file?  Please remove otherwise.
```
#include "cs2/arrayof.h"                              // for StaticArrayOf
#include "cs2/bitvectr.h"
#include "cs2/sparsrbit.h"
```

andrewcraik:
@0xdaryl I have clean up the includes and corrected the typos you found in the commit messages.

PR #1419, opened 2017-07-26
[Mem10] Reduce memory leakage from LocalCSE

andrewcraik:
This commit changes the data storage of LocalCSE from using CS heap allocated structures to using region allocated structures so that we can release the memory consumed during the execution of LocalCSE. Without this improved memory management the optimization can leak hundereds of MB of memory during some large compilations.

This change is based on [Mem 9] and so commits from 'Convert LocalCSE CS2 TableOf into std::map' onward should be considered. This should be merged after Mem8 and Mem9

### DIFF
--- a/compiler/optimizer/OMRLocalCSE.cpp
+++ b/compiler/optimizer/OMRLocalCSE.cpp
@@ -75,7 +75,6 @@ OMR::LocalCSE::LocalCSE(TR::OptimizationManager *manager)
      _seenCallSymbolReferences(comp()->allocator()),
      _seenSymRefs(comp()->allocator()),
      _parentAddedToHT(comp()->allocator()),
-     _killedNodes(comp()->allocator()),
      _availableLoadExprs(comp()->allocator()),
      _availablePinningArrayExprs(comp()->allocator()),
      _killedPinningArrayExprs(comp()->allocator()),
@@ -232,8 +231,10 @@ void OMR::LocalCSE::prePerformOnBlocks()
    _storeMap = new (stackRegion) StoreMap((StoreMapComparator()), StoreMapAllocator(stackRegion));
 
    int32_t symRefCount = comp()->getSymRefCount();
+   int32_t nodeCount = comp()->getNodeCount();
    _possiblyRelevantNodes.init(symRefCount, stackRegion, growable);
    _relevantNodes.init(symRefCount, stackRegion, growable);
+   _killedNodes.init(nodeCount, stackRegion, growable);
 
    comp()->incVisitCount();
    _mayHaveRemovedChecks = false;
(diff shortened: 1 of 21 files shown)


compiler/optimizer/OMRLocalCSE.cpp
```diff
@@ -280,7 +280,7 @@ void OMR::LocalCSE::transformBlock(TR::TreeTop * entryTree, TR::TreeTop * exitTr
    _numNodes = 0;
    _numNullCheckNodes = 0;
    _numCopyPropagations = 0;
-   _arrayRefNodes.deleteAll();
+   _arrayRefNodes = new (stackMemoryRegion) TR_ScratchList<TR::Node>(trMemory());
```
vijaysun-omr: How does this lead to a reduction in memory usage ? Would'nt the deleteAll at least keep the list object around ?

compiler/optimizer/OMRLocalCSE.cpp
```diff
@@ -585,29 +595,28 @@ void OMR::LocalCSE::examineNode(TR::Node *node, SharedSparseBitVector &seenAvail
...
                  }
               }
            else
               {
-              CS2::TableOf<TR::Node *, TR::Allocator>::Cursor nc(_storeNodes);
-              for (nc.SetToFirst(); nc.Valid(); nc.SetToNext())
+              TR_BitVector storeMapSymRefs(comp()->trMemory()->currentStackRegion());
+              for (auto itr = _storeMap->begin(), end = _storeMap->end(); itr != end; ++itr)
+                 storeMapSymRefs.set(itr->first);
+
```
vijaysun-omr: This pattern of code is seen multiple times. Is populating the bit vector here by walking the _storeMap and iterating over it potentially something more expensive in terms of compile time vs iterating over the already populated _storeNodes ?

compiler/optimizer/OMRLocalCSE.cpp
andrewcraik: So the arrayRefNodes was originally allocating out of the stack region for the optimization (rather than the per block stack region wrapped around transformBlock. As a result, the deleteAll memory was leaked until the end of the optimization. This change allocates a new list, but it is allocated out of a fresh scratch region all the time meaning that the memory will not live beyond the transform block leading to a lower memory footprint overall.

compiler/optimizer/OMRLocalCSE.cpp
andrewcraik: Walking the entries of the _storeMap should not be significantly more expensive than walking the list. I was avoiding any semantic changes beyond memory lifetime so we could investigate further in a subsequent revision.

PR #1422, opened 2017-07-26
[Mem12] Clean up TableOf from code generator

andrewcraik:
The TableOf data structure is used to store monitor information during codegen which results in increased memory usage. This commit replaces the TableOf with an std::map for efficiency.

This change is based on top of Mem9, 10, and 11. It consists of one commit.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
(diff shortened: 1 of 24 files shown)


compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
```
0xdaryl: Remove:

```
OMRCodeGenerator.cpp:64:#include "cs2/tableof.h"                            // for TableOf
```

andrewcraik: fixed in latest revision of the commit.

PR #1423, opened 2017-07-26
[Mem13] Remove node pool TableOf

andrewcraik:
The NodePool currently uses a CS2 TableOf as an allocator of sorts. This inefficient for space and memory compared to directly allocating memory. This change converts the NodePool to use a region, removes the pool index since it is no longer useful, and generally cleans up node allocation to save memory and reduce compile time.

This is based on Mem 9, 10, 11 and 12. It consists of 3 commits.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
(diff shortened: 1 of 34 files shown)


mgaudet:
Travis failure looks like a missed removed member -- I think it shows up only there because I think the linter builds are the only `DEBUG` builds we have hooked up to Travis.

andrewcraik:
Fixed the member just before you commented.

mgaudet:
Heh :) I was trying to make sure it wasn't a CMake issue.

andrewcraik:
Yeah I was doing builds with out debug enabled for my testing hence the missed remove :S Thanks for checking!

PR #1424, opened 2017-07-26
Provide APIs for enabling cgroup subsystems

ashu-mehra:
Instead of providing APIs that enable cgroup limits for all
subsystems, a new set of APIs is added which allows to query
which subsystems are available and enable subsystems individually.
This would allow runtimes to enable cgroup subsystems based on their
requirements.
Note that there is another subtle change in the design of the
port library APIs related to cgroup system.
Enabling a cgroup subsystem in portlibrary (by calling
omrsysinfo_cgroup_enable_subsystems) would make the port library APIs
to use cgroup limits internally.
Any port library API that directly returns cgroup limits,
eg omrsysinfo_cgroup_get_memlimit(), does not check whether the
corresponding subsystem (eg memory) is enabled or not.
Therefore, when using omrsysinfo_cgroup_* APIs,
the caller doesn't need to explicitly enable the corresponding
subsystem in the port library.

Consider the case of omrsysinfo_get_physical_memory() which returns
physical memory of the system.
If the system is running in a cgroup, and the cgroup's "memory"
subsystem is enabled, then omrsysinfo_get_physical_memory()
would internally call omrsysinfo_cgroup_get_memlimit(), and
return memory limit imposed by the cgroup.
However, if the user wants to just get the memory limit imposed by the
cgroup, it can directly call omrsysinfo_cgroup_get_memlimit(),
without enabling the "memory" subsystem.

Issue: #1281


ashu-mehra:
@charliegracie @DanHeidinga can you please review.

ashu-mehra:
@charliegracie @DanHeidinga any estimate when this could be reviewed?

charliegracie:
I am not sure why the mutex is requried? Should the mutex be a pthread mutex or a omrthread mutex?  Should the mutex be a global variable or a field in the port lirbary struct?  

Waiting for a thumbs up from @DanHeidinga as well as the answers to my questions above.  Thanks

ashu-mehra:
> I am not sure why the mutex is requried?

`cgroupEntryListMutex` is protecting `PPG_cgroupEntryList` which is initialized in `omrsysinfo_cgroup_system_is_available()`. That API can be called by multiple threads, so the need to protect `PPG_cgroupEntryList`.

> Should the mutex be a pthread mutex or a omrthread mutex? Should the mutex be a global variable or a field in the port lirbary struct?

Since the mutex is only required for protecting `PPG_cgroupEntryList` which is internal to the port library and the mutex is required only in omrsysinfo.c, I used pthread mutex and didn't put it in a port library struct.

charliegracie:
Re: the mutex.  Is that what other places in the port library do?

ashu-mehra:
I see omrsignal.c using a `static omrthread_monitor_t`. I think I should also change `cgroupEntryListMutex` to be of type `omrthread_monitor_t`.

ashu-mehra:
I have updated the change set to use omrthread_monitor_t instead of pthread_mutex_t.
@charliegracie @DanHeidinga Please review.

include_core/omrport.h
```diff
@@ -1438,12 +1443,16 @@ typedef struct OMRPortLibrary {
...
 	/** see @ref omrsysinfo.c::omrsysinfo_os_kernel_info "omrsysinfo_os_kernel_info"*/
 	BOOLEAN  ( *sysinfo_os_kernel_info)(struct OMRPortLibrary *portLibrary, struct OMROSKernelInfo *kernelInfo) ;
-	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_is_limits_supported "omrsysinfo_cgroup_is_limits_supported"*/
-	int32_t ( *sysinfo_cgroup_is_limits_supported)(struct OMRPortLibrary *portLibrary);
-	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_is_limits_enabled "omrsysinfo_cgroup_is_limits_enabled"*/
-	BOOLEAN ( *sysinfo_cgroup_is_limits_enabled)(struct OMRPortLibrary *portLibrary);
-	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_enable_limits "omrsysinfo_cgroup_enable_limits"*/
-	int32_t ( *sysinfo_cgroup_enable_limits)(struct OMRPortLibrary *portLibrary);
+	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_system_is_available "omrsysinfo_cgroup_system_is_available"*/
+	int32_t ( *sysinfo_cgroup_system_is_available)(struct OMRPortLibrary *portLibrary) ;
```
DanHeidinga: Should this be `sysinfo_cgroup_is_system_available()`?

port/common/omrsysinfo.c
```diff
@@ -808,42 +808,73 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
- * @return 0 if the port library can use cgroup limits, otherwise negative error code
+ * @return 0 if the the cgroup system is available, otherwise negative error code
  */
-int32_t 
-omrsysinfo_cgroup_is_limits_supported(struct OMRPortLibrary *portLibrary)
+int32_t
+omrsysinfo_cgroup_system_is_available(struct OMRPortLibrary *portLibrary)
 {
-	return OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;
+	return OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;;
```
DanHeidinga: nitpick: double `;` at end of line

DanHeidinga: Looking at the name and the comment, I'm not confident I know what this method would do.  Is it testing if the runtime is running in a cgroup?

port/common/omrsysinfo.c
```diff
@@ -808,42 +808,73 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[in] subsystems bitwise-OR of flags of type OMR_CGROUP_SUBSYSTEMS_* 
+ * indicating the subsystems to be enabled 
+ *                           
+ * @return bitwise-OR of flags of type OMR_CGROUP_SUBSYSTEMS_* indicating the subsystems that are enabled
+ */
+uint64_t
+omrsysinfo_cgroup_enable_subsystems(struct OMRPortLibrary *portLibrary, uint64_t requestedSubsystems)
+{
```
DanHeidinga: How are errors handled / communicated to the caller by this function?

What if the user tries to enable a subsystem that isn't enabled in the system?

port/unix/omrsysinfo.c
```diff
@@ -1585,8 +1594,15 @@ omrsysinfo_shutdown(struct OMRPortLibrary *portLibrary)
 			PPG_si_executableName = NULL;
 		}
 #if defined(LINUX)
+		omrthread_monitor_enter(cgroupEntryListMonitor);
 		freeCgroupEntries(portLibrary, PPG_cgroupEntryList);
+		omrthread_monitor_exit(cgroupEntryListMonitor);
```
DanHeidinga: The monitor needs to protect the nulling of the ` 		PPG_cgroupEntryList = NULL;` as well as the call to freeCgroupEntries()

port/unix/omrsysinfo.c
```diff
@@ -3369,6 +3425,29 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
...
+	switch(subsystemFlag) {
+	case OMR_CGROUP_SUBSYSTEM_CPU:
+		return CPU;
+	case OMR_CGROUP_SUBSYSTEM_MEMORY:
+		return MEMORY;
+	default:
+		Trc_PRT_Assert_ShouldNeverHappen();
+	}
+	/* unreachable code */
+	Trc_PRT_Assert_ShouldNeverHappen();
```
DanHeidinga: This should issue a compiler warning as technically trace asserts can be disabled, resulting in falling off the end of a non-void function.

DanHeidinga: I think we need a return here and then just the assert in the default case should be sufficient,

port/unix/omrsysinfo.c
```diff
@@ -3191,6 +3213,40 @@ freeCgroupEntries(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cgEntryLis
...
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[in] cgEntryList pointer to OMRCgroupEntry representing a circular list
+ * containing entry for each subsystem
+ *
+ * @return bitwise-OR of flags of type OMR_CGROUP_SUBSYSTEM_* indicating the
+ * subsystems that are available
+ */
+uint64_t
+getAvailableSubsystems(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cgEntryList)
```
DanHeidinga: Shouldn't this be static so it's not called elsewhere?

port/unix_include/omrportpg.h
```diff
@@ -120,7 +121,8 @@ typedef struct OMRPortPlatformGlobals {
 #endif
 
 #if defined(LINUX)
-#define PPG_cgroupLimitsEnabled (portLibrary->portGlobals->platformGlobals.cgroupLimitsEnabled)
+#define PPG_cgroupSubsystemsAvailable (portLibrary->portGlobals->platformGlobals.cgroupSubsystemsAvailable)
```
DanHeidinga: There are three port library globals in play here and it looks like the two new ones are only valid when `PPG_cgroupEntryList ` is non-null?

Can you add a comment here (and in the commit) that describes the relationship between the three port globals?

DanHeidinga:
I haven't reviewed the test changes yet.  I'll look at those once we're in agreement on the rest of the code changes.

port/common/omrsysinfo.c
ashu-mehra: Error can occur during `omrsysinfo_cgroup_is_system_available` in which case no subsystem is enabled, but the error is not communicated back to the user.
I am expecting the APIs would be used in following manner:
```
if (0 == omrsysinfo_cgroup_is_system_available()) {
    availableSubsystems = omrsysinfo_cgroup_get_available_subsystems();
    if (requiredSubsystem & availableSubsystems) {
        omrsysinfo_cgroup_enable_subsystems(requiredSubsystems);
    }
}
```

> What if the user tries to enable a subsystem that isn't enabled in the system?
No error is returned. If the APIs are used as above, this situation shouldn't happen.
And the API actually returns the subsystems that are enabled. So the user can check which subsystems got enabled on this call.

port/unix_include/omrportpg.h
ashu-mehra: The condition `PPG_cgroupEntryList` being non-null is used only to check if `PPG_cgroupSubsystemsAvailable` and `PPG_cgroupSubsystemsAvailable` have already been set. As such there is no relation between the three.

include_core/omrport.h
ashu-mehra: Right, this sounds better.

port/common/omrsysinfo.c
ashu-mehra: Done

ashu-mehra:
Updated the change set as per the comments.

ashu-mehra:
@DanHeidinga any more review comments?

port/unix/omrsysinfo.c
```diff
@@ -3459,38 +3542,68 @@ omrsysinfo_cgroup_is_limits_supported(struct OMRPortLibrary *portLibrary)
 			goto _end;
 		}
 
+		omrthread_monitor_enter(cgroupEntryListMonitor);
 		rc = readCgroupFile(portLibrary, getpid(), &PPG_cgroupEntryList);
+		omrthread_monitor_exit(cgroupEntryListMonitor);
+		if (0 != rc) {
+			goto _end;
+		}
+		PPG_cgroupSubsystemsAvailable = getAvailableSubsystems(portLibrary, PPG_cgroupEntryList);
```
DanHeidinga: Once readCgroupFile() has set PPG_cgroupEntryList to non-null, the PPG_cgroupSubsystemsAvailable global isn't read and this function will always return 0.

It seems strange that this function calls `getAvailableSubsystems`.  I would have (maybe naively?) expected that to be called by `omrsysinfo_cgroup_get_available_subsystems` if the `PPG_cgroupSubsystemsAvailable` global was null...

port/unix/omrsysinfo.c
```diff
@@ -3459,38 +3542,68 @@ omrsysinfo_cgroup_is_limits_supported(struct OMRPortLibrary *portLibrary)
...
-BOOLEAN 
-omrsysinfo_cgroup_is_limits_enabled(struct OMRPortLibrary *portLibrary)
+uint64_t
+omrsysinfo_cgroup_get_available_subsystems(struct OMRPortLibrary *portLibrary)
+{
+#if defined(LINUX)
+	if (NULL == PPG_cgroupEntryList) {
+		int32_t rc = portLibrary->sysinfo_cgroup_is_system_available(portLibrary);
+		if (0 != rc) {
+			PPG_cgroupSubsystemsAvailable = 0;
```
DanHeidinga: I don't like the way this global is sometimes set here, sometimes set in sysinfo_cgroup_is_system_available.  It needs to be set in a single place (single point of initialization)

port/unix_include/omrportpg.h
DanHeidinga: Doesn't that state a relationship?  Neither PPG_cgroupSubsystemsAvailable or PPG_cgroupSubsystemsAvailable are valid if PPG_cgroupEntryList is null.  

The code should state that invariant.

port/win32/omrsysinfo.c
```diff
@@ -1613,22 +1613,34 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
 	return FALSE;
 }
 
-int32_t 
-omrsysinfo_cgroup_is_limits_supported(struct OMRPortLibrary *portLibrary)
+int32_t
+omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
 {
-	return OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;
+	return OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;;
```
DanHeidinga: nitpick: double `;` on this line

ashu-mehra:
@DanHeidinga, there are few things pointed out by @dinogun on need for this PR and design of cgroup APIs.

Main point is -if we update existing portlib APIs to use cgroup limits transparently, then it would change the meaning of the API.
For example, omrsysinfo_get_physical_memory() is expected to return total physical memory in the system. So if we update it to return cgroup memory limit then it would break the API. Users of this API may really want the 'total physical memory in the system' and not the 'total physical memory available to the process'. Moreover, there wouldn't be any way left for the user to get 'total physical memory in the system' if the API is changed to return cgroup memory transparently.

@dinogun, can you please add your thoughts as well.

DanHeidinga:
The rationale behind having existing apis return the cgroup limits is that it makes it way way easier for users of the port library to be docker aware.  If the existing apis aren't updated to use the cgroup limits, then every caller will need to be updated to check if cgroups are enabled, get the cgroup limit otherwise get the original limit.

That's a lot of code for every user to write at every callsite... and its easy to get wrong.  

The design here is to have the runtime **opt-into** the cgroup limits deliberately which mitigates the concerns about "change the meaning of the API".

Can you point to runtimes using this api that would be broken by this change?

ashu-mehra:
@DanHeidinga I understand your point.

One example where modifying existing API may have undesired effect is J9's OperatingSystemMXBean.getTotalPhysicalMemory() which uses `sysinfo_get_physical_memory()`. Javadoc for this API states:
`Returns the total available physical memory on the system in bytes.`
Since this is externally exposed API, we cannot be sure for what purpose this API would be used by the users. If the user is actually relying on this API to get physical memory in the system (eg some load balancer may be using it to get total physical memory in the system based on which it may balance the load), then updating sysinfo_get_physical_memory() to return container limit would cause the user to malfunction.

DanHeidinga:
The "saving grace" for APIs like the OSMXBean is to have the user opt into enabling cgroup limits system wide through some mechanism - in code, command line options, etc.

ashu-mehra:
> have the user opt into enabling cgroup limits system wide through some mechanism - in code, command line options, etc.

@DanHeidinga I don't think I got this, can you please explain.

DanHeidinga:
I don't have a problem with `sysinfo_get_physical_memory` returning the cgroup limit if:

- The user has specified a commandline option (like -XX:+EnableCGroupLimits), or
- The runtime has asked the port library to enable cgroup limits through `sysinfo_cgroup_enable_subsystems`, or
- some other explicit opt-in has occurred.

DanHeidinga:
Just to be clear - if something is OSMXBean and wants to return the physical memory even though the runtime has enabled cgroup limits, then I'm ok with addressing that particular use case through a new bug / issue when it occurs.

port/common/omrsysinfo.c
ashu-mehra: Actually there is no need to call omrsysinfo_cgroup_get_available_subsystems() here:

```
if (0 == omrsysinfo_cgroup_is_system_available()) {
    uint64_t enabledSubsystems = omrsysinfo_cgroup_enable_subsystems(requiredSubsystems);
    if ((enabledSubsystems & requiredSybsystems) != enabledSubsystems) {
            /* do error handling */
    }
}
```

include_core/omrport.h
```diff
@@ -1438,12 +1443,16 @@ typedef struct OMRPortLibrary {
...
-	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_is_limits_enabled "omrsysinfo_cgroup_is_limits_enabled"*/
-	BOOLEAN ( *sysinfo_cgroup_is_limits_enabled)(struct OMRPortLibrary *portLibrary);
-	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_enable_limits "omrsysinfo_cgroup_enable_limits"*/
-	int32_t ( *sysinfo_cgroup_enable_limits)(struct OMRPortLibrary *portLibrary);
+	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_is_system_available "omrsysinfo_cgroup_is_system_available"*/
+	int32_t ( *sysinfo_cgroup_is_system_available)(struct OMRPortLibrary *portLibrary) ;
+	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_get_available_subsystems "omrsysinfo_cgroup_get_available_subsystems"*/
+	uint64_t ( *sysinfo_cgroup_get_available_subsystems)(struct OMRPortLibrary *portLibrary);
+	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_is_subsystem_available "omrsysinfo_cgroup_is_subsystem_available"*/
+	BOOLEAN ( *sysinfo_cgroup_is_subsystem_available)(struct OMRPortLibrary *portLibrary, uint64_t subsystem);
```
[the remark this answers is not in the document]
ashu-mehra: `sysinfo_cgroup_is_system_available` can return negative error code to indicate the reason for unavailability of cgroup system. `sysinfo_cgroup_is_subsystem_available` just checks the subsystem flags which are set.

port/common/omrsysinfo.c
```diff
@@ -808,42 +808,74 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
 
 /**
- * Checks if the port library can use cgroup limits
+ * Returns TRUE if the platform supports cgroup system and
+ * the port library can use it
  *
  * @param[in] portLibrary pointer to OMRPortLibrary
  *
- * @return 0 if the port library can use cgroup limits, otherwise negative error code
+ * @return 0 if the the cgroup system is available, otherwise negative error code
```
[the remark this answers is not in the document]
ashu-mehra: It returns 0 if available, else negative error code. The negative error code indicates its not available, and it includes unsupported platform as well.

port/common/omrsysinfo.c
```diff
@@ -808,42 +808,74 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+/**
+ * Returns TRUE if specified cgroup subsystem is available for port library to use, FALSE otherwise
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[in] subsystem one of the flags of type OMR_CGROUP_SUBSYSTEMS_*
+ *
+ * @return TRUE if specified cgroup subsystem is available for port library to use, FALSE otherwise
+ */
+BOOLEAN
+omrsysinfo_cgroup_is_subsystem_available(struct OMRPortLibrary *portLibrary, uint64_t subsystem)
```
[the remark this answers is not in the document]
ashu-mehra: I updated the API to allow for checking multiple subsystems.

port/unix/omrsysinfo.c
```diff
@@ -1509,14 +1516,18 @@ omrsysinfo_get_physical_memory(struct OMRPortLibrary *portLibrary)
 {
 	uint64_t result = 0;
 
-	if (portLibrary->sysinfo_cgroup_is_limits_enabled(portLibrary)) {
+#if defined(LINUX)
+	uint64_t enabledSubsystems = portLibrary->sysinfo_cgroup_get_enabled_subsystems(portLibrary);
+
+	if (J9_ARE_ALL_BITS_SET(enabledSubsystems, OMR_CGROUP_SUBSYSTEM_MEMORY)) {
```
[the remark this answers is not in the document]
ashu-mehra: Added a new API `sysinfo_cgroup_are_subsystems_enabled()` on same lines as `sysinfo_cgroup_are_subsystems_available()`.

port/unix/omrsysinfo.c
```diff
@@ -1509,14 +1516,18 @@ omrsysinfo_get_physical_memory(struct OMRPortLibrary *portLibrary)
...
 		int rc = 0;
```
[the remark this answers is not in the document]
ashu-mehra: fixed.

port/unix/omrsysinfo.c
```diff
@@ -1585,8 +1596,15 @@ omrsysinfo_shutdown(struct OMRPortLibrary *portLibrary)
 			PPG_si_executableName = NULL;
 		}
 #if defined(LINUX)
+		omrthread_monitor_enter(cgroupEntryListMonitor);
 		freeCgroupEntries(portLibrary, PPG_cgroupEntryList);
 		PPG_cgroupEntryList = NULL;
+		omrthread_monitor_exit(cgroupEntryListMonitor);
+		attachedPortLibraries -= 1;
```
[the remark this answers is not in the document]
ashu-mehra: yes, some fvtests create fake portlibrary to check its management functions. This causes problems with the globals.
Similar pattern to use `attachedPortLibraries` is there in `omrsignal.c`.

port/unix/omrsysinfo.c
```diff
@@ -3459,38 +3542,68 @@ omrsysinfo_cgroup_is_limits_supported(struct OMRPortLibrary *portLibrary)
 			goto _end;
 		}
 
+		omrthread_monitor_enter(cgroupEntryListMonitor);
```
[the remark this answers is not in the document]
ashu-mehra: They would end up creating two lists of OMRCgroupEntry nodes and only one would be assigned to PPG_cgroupEntryList. Other one would be leaked.

port/unix/omrsysinfo.c
ashu-mehra: `PPG_cgroupSubsystemsAvailable` cannot be checked for null; its not a pointer.
`omrsysinfo_cgroup_is_system_available` calls `getAvailableSubsystems` to ensure `PPG_cgroupSubsystemsAvailable` is set properly at the time when `PPG_cgroupEntryList` is created.
If `getAvailableSubsystems` is to be called by `omrsysinfo_cgroup_get_available_subsystems`, then it cannot rely on `PPG_cgroupEntryList` being null to call it, as `PPG_cgroupEntryList` may have been set previously by a call to `omrsysinfo_cgroup_is_system_available`.

port/unix/omrsysinfo.c
ashu-mehra: I moved it to `omrsysinfo_cgroup_is_system_available`.

ashu-mehra:
@DanHeidinga I have updated the change set to address the review comments.

ashu-mehra:
@DanHeidinga can you please review updated change set.

DanHeidinga:
I'm hoping to get through this review this week.  Its been somewhat crazy due to internal deadlines recently and I haven't had a lot of cycles for this.  Sorry.

port/common/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: Isn't this really a boolean then?  Indicating if the system is available or not?

I'm struggling to see what the user code would do with the failure case that that a boolean wouldn't give us.

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: This is great info to put in the code - a comment that describes the problem due to multiple port libraries and indicating the tests that run into this problem.  

That way no one has to rediscover the rationale for this code when they investigating it in the future.

port/unix/omrsysinfo.c
```diff
@@ -3191,6 +3212,40 @@ freeCgroupEntries(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cgEntryLis
...
+	}
+
+	do {
+		if (!strcmp(cgEntry->subsystem, subsystemNames[CPU])) {
+			availableSubsystems |= OMR_CGROUP_SUBSYSTEM_CPU;
+		} else if (!strcmp(cgEntry->subsystem, subsystemNames[MEMORY])) {
+			availableSubsystems |= OMR_CGROUP_SUBSYSTEM_MEMORY;
+		}
+		cgEntry = cgEntry->next;
+	} while (cgEntry!= cgEntryList);
```
DanHeidinga: How long is this list on a typical system?  Vanilla hw vs inside a container?  If the list is typically capped at 2-3 entries, this code is fine.  If we expect longer lists, we'll want to write this differently, especially if it gets called at startup

port/common/omrsysinfo.c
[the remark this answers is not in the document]
ashu-mehra: Right, this too can be a boolean. Error code would be useful from debugging/service perspective, but then I think it should be displayed in a trace point, and need not be a return value.

fvtest/porttest/si.cpp
```diff
@@ -2109,51 +2109,35 @@ TEST(PortSysinfoTest, sysinfo_cgroup_get_memlimit)
...
-
-	rc = omrsysinfo_cgroup_get_memlimit(&cgroupMemLimit);
-
-#if defined(LINUX)
-	if (0 == cgroupLimitEnabled) {
-		if (OMRPORT_ERROR_SYSINFO_CGROUP_LIMITS_DISABLED == rc) {
-			outputErrorMessage(PORTTEST_ERROR_ARGS, "omrsysinfo_cgroup_get_memlimit returned %d when cgroup limits was enabled successfully\n", rc);
+	/* Compare sysinfo_get_physical_memory and sysinfo_cgroup_get_memlimit after enabling memory subsystem */
+	enabledSubsystems = omrsysinfo_cgroup_enable_subsystems(OMR_CGROUP_SUBSYSTEM_MEMORY);
+	if (OMR_CGROUP_SUBSYSTEM_MEMORY == enabledSubsystems) {
```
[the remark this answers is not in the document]
ashu-mehra: Thats correct.

port/unix/omrsysinfo.c
ashu-mehra: Currently `cgEntryList` has all cgroup subsystems configured in the system. In cgroup v1, that comes to 12 subsystems.
We can probably restrict it to only the ones we need - memory and cpu for now.
Let me add another change set on top of existing one which would highlight these changes.

ashu-mehra: @DanHeidinga please take a look at commit 175daad that avoids adding every subsystem to `PPG_cgroupEntryList` and also sets `PPG_cgroupSubsystemsAvailable` more efficiently than before.

ashu-mehra:
@DanHeidinga can you please review commit 175daad.

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: @ashu-mehra Can you update this code to include the details about the problem with multiple port libraries and that it typically is only seen in test situations?

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: This should check `PPG_cgroupEntryList` under the mutex to determine whether it needs to call `readCgroupFile()` so that it doesn't leak.

port/unix/omrsysinfo.c
```diff
@@ -3317,31 +3345,41 @@ getCgroupNameForSubsystem(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cg
...
-		rc = portLibrary->error_set_last_error_with_message(portLibrary, OMRPORT_ERROR_SYSINFO_MEMORY_ALLOC_FAILED, "memory allocation for cgroup entry failed");
-		goto _end;
-	}
-	cgEntry->hierarchyId = hierId;
-	cgEntry->subsystem = (char *)(cgEntry + 1);
-	strcpy(cgEntry->subsystem, subsystem);
-	cgEntry->cgroup = cgEntry->subsystem + strlen(subsystem) + 1;
-	strcpy(cgEntry->cgroup, cgroupName);
+	for (i = 0; i < sizeof(supportedSubsystems) / sizeof(supportedSubsystems[0]); i++) {
+		if (J9_ARE_NO_BITS_SET(PPG_cgroupSubsystemsAvailable, supportedSubsystems[i].flag)
```
DanHeidinga: Is it possible for two threads to be executing this method at once?  If it's not, then the function should document what lock needs to be held when calling it.  Otherwise, it shouldn't be reading / modifying global state (`PPG_cgroupSubsystemsAvailable`) without a lock.

port/unix_include/omrportpg.h
DanHeidinga: Can you add the above invariant as a comment in the code?

port/unix/omrsysinfo.c
ashu-mehra: I have updated `addCgroupEntry()` to avoid using PPG_cgroupSubsystemsAvailable.
Available subsystems are now determined by the `readCgroupFile()` and returned back via one of its parameters.

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
ashu-mehra: Done

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
ashu-mehra: Done

port/unix_include/omrportpg.h
ashu-mehra: Done

ashu-mehra:
@DanHeidinga added new commit https://github.com/eclipse/omr/pull/1424/commits/5a9e49fb3a0b3eb9658c92d33ce13b76661bcd12 to address the comments.
Please review.

port/unix/omrsysinfo.c
```diff
@@ -3408,24 +3442,23 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
...
 			}
 
-			cgName = getCgroupNameForSubsystem(portLibrary, cgEntryList, cursor);
-
-			if (NULL == cgName) {
-				rc = addCgroupEntry(portLibrary, &cgEntryList, hierId, cursor, cgroup);
-				if (0 != rc) {
-					goto _end;
+			for (i = 0; i < sizeof(supportedSubsystems) / sizeof(supportedSubsystems[0]); i++) {
+				if (J9_ARE_NO_BITS_SET(available, supportedSubsystems[i].flag)
```
DanHeidinga: This should use `OMR_ARE_NO_BITS_SET` instead of the `J9` version.  This will solve a compile break in this change.

DanHeidinga:
@ashu-mehra Thanks for addressing the locks, globals and comments questions.

The Travis and AppVeyor builds are failing now.  Can you fix up the issues and then I'll give this another look over?

ashu-mehra:
Checking travis-ci build failure...

ashu-mehra:
Pushed a commit to fix compile failure reported by travis-ci build.

DanHeidinga:
@charliegracie Can you or one of the other committers give this a second look and merge?  The commits should probably be squashed prior to the merge.

charliegracie:
Hi @ashu-mehra would it be possible to refactor this commits to resolve the compilation issues into the appropriate original commits.  It would clean up the commit history and make it easier to test the codebase at different points in history.  If this was merged there would be at least 3 known points in history that the codebase would not compile.  Thanks

ashu-mehra:
@charliegracie I squashed the commits.

charliegracie:
genie-omr build all

ashu-mehra:
It looks like I inserted a bug in `omrsysinfo_cgroup_is_system_available` while addressing review comments.
I will be updating the last commit with following change to fix it. Since its trivial, I don't think we need to run tests again.

```
@@ -3603,7 +3603,7 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
                }
 
                omrthread_monitor_enter(cgroupEntryListMonitor);
-               if (NULL != PPG_cgroupEntryList) {
+               if (NULL == PPG_cgroupEntryList) {
                        rc = readCgroupFile(portLibrary, getpid(), &PPG_cgroupEntryList, &PPG_cgroupSubsystemsAvailable);
                }
                omrthread_monitor_exit(cgroupEntryListMonitor);
```

charliegracie:
@genie-omr build all

port/unix/omrsysinfo.c
```diff
@@ -3408,24 +3442,23 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
...
 			}
 
-			cgName = getCgroupNameForSubsystem(portLibrary, cgEntryList, cursor);
-
-			if (NULL == cgName) {
-				rc = addCgroupEntry(portLibrary, &cgEntryList, hierId, cursor, cgroup);
-				if (0 != rc) {
-					goto _end;
+			for (i = 0; i < sizeof(supportedSubsystems) / sizeof(supportedSubsystems[0]); i++) {
+				if (OMR_ARE_NO_BITS_SET(available, supportedSubsystems[i].flag)
```
charliegracie: I would have preferred that this if statement be broken into 2 so it is easier to debug but that is a stylistic issue that is not worth changing the code for.  I will accept this change as is.

charliegracie:
This is a great start to the Cgroup support.  Thank you

DanHeidinga:
Echoing Charlie's comments - thanks for working through the reviews on this with us @ashu-mehra!  Great to see it merged!

ashu-mehra:
:relieved:

PR #1425, opened 2017-07-26
Split platform.cmake in two

mgaudet:
Split the platform detection code into a routine that exports variables, and a file that sets definitions. 

~(Will be rebased after #1416 is merged, so WIP for now)~ Not sure what I was thinking


0xdaryl:
@rwy0717 , @dnakamura : please review.

rwy0717:
I like the style of defining then calling a macro. I've only just glanced at it, but this looks like a good idea to me. Are we planning on updating our other cmake files to follow this pattern?

Re naming conventions:
Should the macro start with `omr_`?
Should the file be called `DetectSystemInformation.cmake`?

Also, it looks like linux is in there twice.

cmake/system_detection.cmake
```diff
@@ -0,0 +1,93 @@
...
+#
+# Contributors:
+#    Multiple authors (IBM Corp.) - initial implementation and documentation
+###############################################################################
+
+# Translate from CMake's view of the system to the OMR view of the system. 
+# Exports a number of variables indicicating platform, os, endianness, etc. 
+# - OMR_ARCH_{X86,ARM,S390} # TODO: Add POWER
+# - OMR_ENV_DATA{32,64}  
+# - OMR_ENV_TARGET_DATASZIZE (either 32 or 64)
```
rwy0717: Minor typo

cmake/system_detection.cmake
```diff
@@ -0,0 +1,93 @@
...
+   if(OMR_HOST_OS STREQUAL "zos")
+      # ZOS specifics
+   endif()
+
+
+   if(OMR_HOST_OS STREQUAL "osx")
+      # OSX specifics
+   endif()
+
+endmacro(detect_system_information)
```
rwy0717: Do we want to print a summary here, instead of in `platform.cmake`?
or define a `print_system_summary` macro?

cmake/system_detection.cmake
```diff
@@ -0,0 +1,93 @@
...
+      set(OMR_TARGET_DATASIZE 32)
+      set(OMR_ENV_LITTLE_ENDIAN ON)
+   elseif(CMAKE_SYSTEM_NAME MATCHES "OS390")
+      set(OMR_ARCH_S390 ON)
+      set(OMR_ENV_DATA32 ON)
+      set(OMR_TARGET_DATASIZE 32)
+      set(OMR_HOST_OS "zos")
+   else()
+      message(FATAL_ERROR "Unknown processor: ${CMAKE_SYSTEM_PROCESSOR}")
+   endif()
```
rwy0717: Are you OK to change to `MATCHES` instead of `STREQUAL`/`OR`?

cmake/system_detection.cmake
mgaudet: I think that's a good idea, but I'll do it in another PR.

mgaudet:
Updated: 

* Cleaned up the platform translation code 
* renamed file and macro
* Fixed typo. 
* Removed duplicated linux stanza

rwy0717:
LGTM

dnakamura:
For things like `OMR_HOST_OS`. Do we want those as variables, or cache variables? Having them as variables means that uses can override them by setting cache variables, which may or may not be desired behavior.  On the one hand if everything is working as it should you will probably break the build. On the other, if platform detection is screwy it lets the user overrule it

rwy0717:
I think @mgaudet  is right, cached values always override set values. The user can do this with `-Dvar=val`.

In the future, it may make sense to cache the result of this macro.

However, I think users should be setting `CMAKE_SYSTEM_PROCESSOR`/etc, not the OMR flags. Caching IMO signals that these are user flags. What about a combination of caching and [`mark_as_advanced`](https://cmake.org/cmake/help/v3.9/command/mark_as_advanced.html)?

dnakamura:
@rwy0717  you can also set as internal which means cmake wont show them even if you tell it to show advanced.
@mgaudet  Since they are regular variables now the user can effectively override them by specifying `-DOMR_HOST_OS=xyz` on the command line.  This would override any setting we put since variable resolution always prefers cache variables to local ones

mgaudet:
Ok. I think I'm good with leaving them as normal variables for now -- while a user could certainly make this go hairy by caching incorrect values, I'm not sure how changing these to cache variables improves the situation.

rwy0717:
If we ever needed to, this might be how i'd rewrite the OMR_HOST_OS detection:
```cmake
set(OMR_HOST_OS "OMR_HOST_OS-NOTFOUND" CACHE STRING "The normalized host OS. Derived from CMAKE_SYSTEM_NAME")
mark_as_advanced(OMR_HOST_OS)

if(NOT OMR_HOST_OS)
   #TODO: Detect AIX 
   if(CMAKE_SYSTEM_NAME MATCHES "Linux")
      set(OMR_HOST_OS "linux")
   elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
      set(OMR_HOST_OS "osx")
   elseif(CMAKE_SYSTEM_NAME MATCHES "Windows")
      set(OMR_HOST_OS "win")
   elseif(CMAKE_SYSTEM_NAME MATCHES "OS390")
      set(OMR_HOST_OS "zos")
   else()
      message(FATAL_ERROR "Unsupported OS: ${CMAKE_SYSTEM_NAME}")
   endif()
endif()
```

It's not as obvious with the CPU.

For now, I +1 leaving it as is.

charliegracie:
@youngar once @dnakamura gives the last changes a thumbs up can you review this and merge it if you are happy with it?  Thanks

dnakamura:
@mgaudet 
>Ok. I think I'm good with leaving them as normal variables for now -- while a user could certainly make this go hairy by caching incorrect values, I'm not sure how changing these to cache variables improves the situation.
In the cmake file we could be doing `set(OMR_HOST_OS "win" CACHE STRING FORCE)` which would override any setting provided by the user. 
That being said, I'm fine with merging as is

PR #1427, opened 2017-07-26
Disable traps when z10 architecture is disabled

NigelYiboYu:
If one chooses to disable z10 architecture level, traps should be disabled
so that the optimizations and codegen can avoid generating trap-related IL
nodes and instructions.

Move project specific random z architecture level disabling code out of
OMR and into the relevant project.

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -553,60 +553,6 @@ OMR::Z::CodeGenerator::CodeGenerator()
    // Initialize Linkage for Code Generator
    self()->initializeLinkage();
 
-   // Do random Disable<Platform> when -Xjit randomGen
-   if (comp->getOption(TR_Randomize))
-      {
-      switch(randomizer.randomInt(TR::Compiler->target.cpu.id() - TR_s370gp7))
-         {
-         case 1:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z9);
-            traceMsg(comp,"disablez9");
-            break;
-            }
-
-         case 2:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z10);
-            traceMsg(comp, "disablez10");
-            break;
-            }
-
-         case 3:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z196);
-            traceMsg(comp, "disablez196");
-            break;
-            }
-
-         case 4:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_zEC12);
-            traceMsg(comp, "disablezEC12");
-            break;
-            }
-
-         case 5:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z13);
-            traceMsg(comp, "disablez13");
-            break;
-            }
-         case 6:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z14);
-            traceMsg(comp, "disablez14");
-            break;
-            }
-         case 7:
-            {
-            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_zNext);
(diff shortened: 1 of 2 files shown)


compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -552,65 +552,16 @@ OMR::Z::CodeGenerator::CodeGenerator()
...
+   
    // Check for -Xjit disable options and target this specific compilation for the proper target
    if (comp->getOption(TR_DisableZ10))
       {
       _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z10);
+
+      // If z10 is disabled, there will no longer be trap instructions.
+      // Set disable traps so that the optimizations and codegen can avoid generating
+      // trap-specific nodes or instructions.
+      comp->setOption(TR_DisableTraps);
```
fjeremic: This code doesn't belong here. What if the actual hardware we're running on is z9 and the user didn't specify any compiler options? You would never enter this path and `comp->getOption(TR_DisableTraps)` would return `true`. 

Please move this code to other z10 related queries below:
https://github.com/eclipse/omr/blob/d183d21dc7304ed1f39039bed26c81be9cc5ff2e/compiler/z/codegen/OMRCodeGenerator.cpp#L753-L756

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -1178,6 +1173,12 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    TR::list<TR::S390ConstantDataSnippet*>  _constantList;
    TR::list<TR::S390ConstantDataSnippet*>  _snippetDataList;
 
+   /**
+    * _processorInfo contains the targetted hardware level for the compilation.
+    * This may be different than the real hardware the JVM is currently running
+    * on, due to user specified options.
+    */
+   TR_S390ProcessorInfo            _processorInfo;
```
fjeremic: Awesome to see people documenting related code as part of their changes. Keep it up!

compiler/z/codegen/OMRCodeGenerator.cpp
NigelYiboYu: OK. Didn't realize that we didn't have existing code to disable traps on z9.

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -1178,6 +1173,12 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    TR::list<TR::S390ConstantDataSnippet*>  _constantList;
    TR::list<TR::S390ConstantDataSnippet*>  _snippetDataList;
 
+   /**
+    * _processorInfo contains the targetted hardware level for the compilation.
+    * This may be different than the real hardware the JVM is currently running
```
0xdaryl: Is this Java specific?  Or should "JVM" be rephrased to something runtime-neutral?

compiler/z/codegen/OMRCodeGenerator.hpp
```diff
@@ -1178,6 +1173,12 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    TR::list<TR::S390ConstantDataSnippet*>  _constantList;
    TR::list<TR::S390ConstantDataSnippet*>  _snippetDataList;
 
+   /**
+    * _processorInfo contains the targetted hardware level for the compilation.
```
0xdaryl: Spelling: targeted

NigelYiboYu:
1. Fixed typo in comments and also replaced the word `JVM` with something project neutral.
2. Split into two commits for logical separation

0xdaryl:
Thanks @NigelYiboYu .  Please remove the "Task:" from your commit messages as it is meaningless to Eclipse OMR.  After that, this should be good to merge.

NigelYiboYu:
@0xdaryl `Task` numbers removed.

PR #1429, opened 2017-07-27
Upgrade Google Test from 1.7 to 1.8

### LINKED ISSUE #1381: Upgrade Googletest from 1.7 to 1.8
As mentioned in #1267, we would like to use the newer version of the Googletest framework.  The upgrade requires a few additional changes to Googletest so that tests will run on all supported platforms, including z/OS.  If there are not significant changes to gtest between 1.7 and 1.8, then we will still encounter an issue (related to ascii to ebcdic conversion, port library atoe layer, and resulting in garbled test output on z/OS) that needs to be ported forward as part of this upgrade.

For this upgrade, we will add a new directory https://github.com/eclipse/omr/tree/master/third_party/gtest-1.8.0 populated from a release of the opensource tool based off this tag: https://github.com/google/googletest/releases/tag/release-1.8.0
and make the few necessary changes to continue to be able to run on all platforms.

While we transition, I suggest keeping the 1.7.0 version until we are confident that the newer version runs well (and we see no concerning test infrastructure issues).

renfeiw:
Fixes: #1381
Signed-off-by: renfeiw@ca.ibm.com <renfeiw@ca.ibm.com>


0xdaryl:
Thanks @renfeiw.  I will create the Eclipse CQ in order to get this merged.

charliegracie:
@0xdaryl can you paste the link to the CQ in here?  Thanks

0xdaryl:
Eclipse CQ:   https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13953

charliegracie:
You do not need to use `Issue:  #1381` as `Fixes: #1381` is an appropriate GitHub tag which will auto close the Issue if that is what you want.  If you believe this is the final fix `Fixes:` is right but if you believe there are other PRs required just use `Issue:`

third_party/gtest-1.8.0/include/gtest/internal/gtest-param-util.h
```diff
@@ -57,9 +57,10 @@ namespace testing {
 // Consists of the parameter value and the integer parameter index.
 template <class ParamType>
 struct TestParamInfo {
-  TestParamInfo(const ParamType& a_param, size_t an_index) :
-    param(a_param),
-    index(an_index) {}
+  TestParamInfo(const ParamType& a_param, size_t an_index) {
```
charliegracie: which platform / compiler did this not work on?

third_party/gtest-1.8.0/src/gtest.cc
```diff
@@ -3212,6 +3212,8 @@ void PrettyUnitTestResultPrinter::OnTestIterationEnd(const UnitTest& unit_test,
     PrintFailedTests(unit_test);
     printf("\n%2d FAILED %s\n", num_failures,
                         num_failures == 1 ? "TEST" : "TESTS");
+  } else {
+	  ColoredPrintf(COLOR_GREEN,  "[  ALL TESTS PASSED  ] \n");
```
charliegracie: Have we considered pushing this change back to GoogleTest?  To me having an extra print saying that ALL TESTS PASSED makes sense.  Or should we be checking the return value of RUN_ALL_TESTS in our code to print this so they we have one less modification?

third_party/gtest-1.8.0/include/gtest/internal/gtest-param-util.h
renfeiw: It doesn't work on zos with the following error:
"../../third_party/gtest-1.8.0/include/gtest/internal/gtest-param-util.h", line 62.19: CCN5861 (S) Too few arguments are specified for macro "index". Empty arguments are used.
"/usr/include/strings.h", line 206.16: CCN5425 (I) "index" is defined on line 206 of "/usr/include/strings.h".

third_party/gtest-1.8.0/src/gtest.cc
renfeiw: This is only required by Axxon, I don' think Axxon can check the return value of the test. This line can be removed when we get rid of Axxon in the future.

charliegracie: Axxon is not used by Eclipse OMR.  I do like the idea of the test cases printing this sort of thing as it makes it easier to see errors or the lack of errors.  What I am suggesting is that our testMain functions could check the return of RUN_ALL_TESTS and print the `ALL TESTS PASSED` if the return value is 0.  This change does not have to be made under this PR as this is exactly what we were doing in the previous version but I am just curious if you think my idea would work.

renfeiw: I think it should work. I will create another issue for this.

0xdaryl:
The CQ has been approved by the PMC.

0xdaryl:
@charliegracie , are your concerns with this addressed?

PR #1430, opened 2017-07-27
Add more aot relocation types to X86ImmInstruction

cathyzhyi:
There are cases where a X86ImmInstruction contains method/class
address and relocation records with correct type are needed.

### DIFF
--- a/compiler/x/codegen/X86BinaryEncoding.cpp
+++ b/compiler/x/codegen/X86BinaryEncoding.cpp
@@ -948,6 +948,22 @@ TR::X86ImmInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
                   (uint8_t *)(intptr_t)getNode()->getInlinedSiteIndex() , (TR_ExternalRelocationTargetKind) getReloKind(), cg()),
                   __FILE__, __LINE__, getNode());
                break;
+            case TR_MethodPointer:
+               if (getNode() && getNode()->getInlinedSiteIndex() == -1 && 
+                  (void *) getSourceImmediate() == cg()->comp()->getCurrentMethod()->resolvedMethodAddress())
+                  setReloKind(TR_RamMethod);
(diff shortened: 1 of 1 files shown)


cathyzhyi:
@mstoodle Hey Mark, could you help me review this AOT change?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -948,6 +948,21 @@ TR::X86ImmInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
                   (uint8_t *)(intptr_t)getNode()->getInlinedSiteIndex() , (TR_ExternalRelocationTargetKind) getReloKind(), cg()),
                   __FILE__, __LINE__, getNode());
                break;
+            case TR_MethodPointer:
+               if (getNode() && getNode()->getInlinedSiteIndex() == -1)
```
mstoodle: I think you should include in this condition a test that the immediate value matches the J9Method of the compiled method. Otherwise, any instruction generated from a node in the method being compiled that has a MethodPointer will be replaced with the compiled method's J9Method (because that's what RamMethod does). Although that *probably* works in all existing cases, I don't think it's necessarily the right thing to do. Only if the method pointer is the current method's J9Method is it definitely the right thing to do.

cathyzhyi: Thanks for the suggestion. Modified accordingly and fixed other similar conditions in the code.

andrewcraik:
@mstoodle ping? I believe @cathyzhyi has made the requested changes.

mstoodle:
Wow, bang-on 50 character commit message :) . LGTM, but @0xdaryl self assigned so I won't merge without his ok.

andrewcraik:
@0xdaryl can you please review and merge? we need this change to allow other changes in an OMR consuming project.

Issue #1397, opened 2017-07-22
Remove unused `canOptimizeEdit` from Z code generator


0xdaryl:
This function was left over from early compiler technology.  It is not used and should be removed.

`OMR::Z::CodeGenerator::canOptimizeEdit(uint8_t *parm, int32_t length)` in `compiler/z/codegen/OMRCodeGenerator.cpp`

PR #1432, opened 2017-07-28
Remove `canOptimizeEdit` from Z code generator

pparkkin:
This function was left over from early compiler technology. It is not
used and should be removed.

Issue https://github.com/eclipse/omr/issues/1397

Signed-off-by: Paavo Parkkinen <pparkkin@gmail.com>


pparkkin:
Should this be tested in System z?

0xdaryl:
IBM will take care of System Z testing because at the moment such systems are not part of CI.  Thanks for your contribution!

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -11416,14 +11416,6 @@ bool TR_S390Peephole::forwardBranchTarget()
       }
    return false;
    }
-bool
-OMR::Z::CodeGenerator::canOptimizeEdit(uint8_t *parm, int32_t length)
-   {
-   bool leadingSigs = false, trailingStr = false;
-   uint8_t fill = 0;
-   int32_t numDigits = 0, numSigs = 0, sigOff = 0, lastDigit = 0, digitsAfterSig = 0;
-   return self()->parseEditParm(parm, length, leadingSigs, trailingStr, fill, numDigits, numSigs, sigOff, lastDigit, digitsAfterSig);
```
0xdaryl: `parseEditParm` is actually dead too.  I should have mentioned that in the issue.  Would you mind removing that as part of this PR as well?

pparkkin: Removed.

0xdaryl: Excellent!  Thanks.

PR #1434, opened 2017-07-28
[Mem15] Reduce LoopStrider memory usage

andrewcraik:
Currently the loop strider allocates large amounts memory during it recursive analysis routines and this memory leaks during the analysis. This memory leaks until the optimization has finished running which has caused some massive growth in memory consumption. Further, some data structures are pre-sized to the number of symbol references when the actual number of live entries is very small. To reduce this wasted space, several key data structures are transformed from arrays to maps keyed by symref index.

This commit is based on top of Mem 9, 10, 11, 12 and 13 and so should be merged after these other changes. The delta in this commit is 2 commits beyond those in Mem 13.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -94,6 +94,10 @@
 #include "control/Recompilation.hpp"           // for TR_Recompilation, etc
 #include "runtime/CodeCacheExceptions.hpp"
 #include "ilgen/IlGen.hpp"                     // for TR_IlGenerator
+
+// this ratio defines how full the alias memory region is allowed to become before
+// it is recreated after an optimization finishes
+#define ALIAS_REGION_LOAD_FACTOR 0.75
 
 #ifdef J9_PROJECT_SPECIFIC
 #include "control/RecompilationInfo.hpp"           // for TR_Recompilation, etc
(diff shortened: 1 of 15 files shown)


andrewcraik:
Rebased onto the merge of Mem 9 - 13 inclusive.

compiler/compile/OMRCompilation.cpp
```diff
@@ -2713,3 +2714,17 @@ OMR::Compilation::CompilationPhaseScope::~CompilationPhaseScope()
    {
    _comp->restoreCompilationPhase(_savedPhase);
    }
+
+TR::Region &OMR::Compilation::aliasRegion()
+   {
+   return self()->_aliasRegion;
+   }
+
```
vijaysun-omr: I would like @lmaisons to also review this commit

compiler/optimizer/OMROptimizer.cpp
```diff
@@ -2050,6 +2050,10 @@ int32_t OMR::Optimizer::performOptimization(const OptimizationStrategy *optimiza
          }
 
       delete opt;
+      // we cannot easily invalidate during IL gen since we could be peeking and we cannot destory our
+      // caller's alias sets
+      if (!isIlGenOpt())
```
vijaysun-omr: How does the case when we are doing IL gen for inlining get handled ? i.e. could we be destroying our caller's alias sets (as the comment suggests) in that case ?

compiler/compile/OMRCompilation.cpp
andrewcraik: This file is changed in a commit under Mem14 already reviewed by @0xdaryl, but @lmaisons is welcome to comment in Mem14

compiler/optimizer/OMROptimizer.cpp
andrewcraik: isIlGenOpt will return true in both cases - we found it in the context of peeking but isIlGenOpt will be true for peeking or during inlining. Basically, only after 'true' optimization steps that are not nested can we invalidate the alias set which is the intent behind this check.

vijaysun-omr: Okay I see it defers the invalidation until the next "top level compile" opt is finished.

vijaysun-omr:
I'm ready to merge this PR once @lmaisons has looked at the commit I requested he review.

andrewcraik:
Mem14 is WIP so this should not merge until Mem14 has merged.

andrewcraik:
The issue we were tracking has been resolved and this change should be good to merge @vijaysun-omr

vijaysun-omr:
Merging this now that Mem14 has been merged.

PR #1436, opened 2017-07-31
Ignore OSR exit edges for induction variables

ncough:
Exit edges to OSR blocks can be ignored under
induction variable analysis as they do not
depend on the induction variable and will
not modify it. They are mostly ignored already
unless the exit edge is from an inner region.

### DIFF
--- a/compiler/optimizer/InductionVariable.cpp
+++ b/compiler/optimizer/InductionVariable.cpp
@@ -5459,7 +5459,12 @@ static inline int64_t ceiling(int64_t numer, int64_t denom)
    }
 
 TR_InductionVariableAnalysis::TR_InductionVariableAnalysis(TR::OptimizationManager *manager)
-   : TR::Optimization(manager), _ivs(0), _blockInfo(0), _dominators(NULL)
+   : TR::Optimization(manager),
+     _ivs(0),
+     _blockInfo(0),
+     _dominators(NULL),
+     _seenInnerRegionExit(0, trMemory(), stackAlloc, growable),
+     _isOSRInduceBlock(0, trMemory(), stackAlloc, growable)
    {}
 
 int32_t TR_InductionVariableAnalysis::perform()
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
@andrewcraik please review as well.

andrewcraik:
Looks good to me - the OSR edges are not something IV needs to worry about - they will transition to equivalent state in the VM and execution will continue. The danger case of improperly nested loops is not going to occur with these edges since they will never merge back to the method being compiled.

ncough:
The old implementation of this ignore would always fail due to structure not being set on the destination of an exit edge. To find the destination block of an exit edge, it is necessary to search for a block with a matching block number. As OSR induce blocks have edges to the exit, search the exit's predecessors.

compiler/optimizer/InductionVariable.cpp
```diff
@@ -6382,6 +6382,27 @@ TR_InductionVariableAnalysis::analyzeExitEdges(TR_RegionStructure *loop,
 
       if (subNode->getStructure()->asRegion())
          {
+         // OSR blocks can be ignored here as they will not be impacted by a loop controlling
+         // condition. Exit edges do not have structure for their destination. Instead, find
+         // the corresponding block as a predecessor of the CFG exit based on the number.
+         TR::Block *dest = NULL;
+         int32_t toNum = edge->getTo()->getNumber();
+         TR::CFGNode *end = comp()->getFlowGraph()->getEnd();
+         for (auto predEdge = end->getPredecessors().begin(); predEdge != end->getPredecessors().end(); ++predEdge)
```
andrewcraik: This seems an expensive way of checking and are we sure it will always find the right exit block? You are using the exit of the CFG to find single level predecessors? I don't follow how that will always work...

ncough: OSR induce blocks will always have an edge to the exit, so they should be found by this check. To reduce its cost, the results of the check have been memoized.

andrewcraik:
LGTM from me @vijaysun-omr over to you.

vijaysun-omr:
Looks okay to me. Merging.

PR #1439, opened 2017-07-31
Partial GC initialization for Concurrent Scavenger

dmitripivkine:
Concurrent Scavenger might be called earlier then GC Master Thread
initialization is complete. So pre-initialization required.
Add preMasterGCThreadInitialize() implementation to the Scavenger.


dmitripivkine:
I believe this change is good to go. Personal build passes and original problem is fixed (use to be ~100% on some machines)

PR #1443, opened 2017-08-01
New option disableLoopTransfer

JamesKingdon:
The new option disableLoopTransfer replaces the
use of environment variable TR_disableLoopTransfer. The option
is applied at both the original location and two further points
to more completely suppress the artifacts related to loop
transfer.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -412,6 +412,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableLoopReplicator",              "O\tdisable loop replicator",                        TR::Options::disableOptimization, loopReplicator, 0, "P"},
    {"disableLoopReplicatorColdSideEntryCheck","I\tdisable cold side-entry check for replicating loops containing hot inner loops", SET_OPTION_BIT(TR_DisableLoopReplicatorColdSideEntryCheck), "P"},
    {"disableLoopStrider",                 "O\tdisable loop strider",                           TR::Options::disableOptimization, loopStrider, 0, "P"},
+   {"disableLoopTransfer",                "O\tdisable the loop transfer part of loop versioner", SET_OPTION_BIT(TR_DisableLoopTransfer), "F"},
    {"disableLoopVersioner",               "O\tdisable loop versioner",                         TR::Options::disableOptimization, loopVersioner, 0, "P"},
    {"disableMarkingOfHotFields",          "O\tdisable marking of Hot Fields",                  SET_OPTION_BIT(TR_DisableMarkingOfHotFields), "F"},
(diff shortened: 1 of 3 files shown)


0xdaryl:
Why not simply beef up what TR_disableLoopTransfer does?  I'm not clear why you need a new option.  Its name is explicit too:  I'm not sure what "LVloopTransfer" means...

JamesKingdon:
The env var is all or nothing, i.e. it impacts every method whereas an option can be applied with a method filter. Env vars are also notoriously difficult to apply reliably in some cases, so an option is much more consumable. I added the 'LV' to the name because of the possible confusion with DLT. This new option is specifically disabling the loop transfer aspect of loop versioning. 'disableLoopVersioningLoopTransfer' would have been more explicit but seems very unwieldy, so I tried for a compromise. Names are hard.

JamesKingdon:
@andrewcraik for thoughts on the option name?

JamesKingdon:
Reverting to WIP to respond to Andrew's comment:

> I'm not a fan of that new name - it is quite cryptic - loop transfer is a known thing for people working on the optimizer. If it has to change then something more like disableVersionedLoopTransfer or similar

Taking both Andrew and Daryl's reactions together, it's clear that I'm in the minority as far as being concerned about any confusion with DLT, so the best bet seems to stay closer to the original naming. I'll change the option name to disableLoopTransfer and the corresponding enum identifier TR_DisableLoopTransfer.

andrewcraik:
LGTM

compiler/control/OMROptions.cpp
```diff
@@ -412,6 +412,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableLoopReplicator",              "O\tdisable loop replicator",                        TR::Options::disableOptimization, loopReplicator, 0, "P"},
    {"disableLoopReplicatorColdSideEntryCheck","I\tdisable cold side-entry check for replicating loops containing hot inner loops", SET_OPTION_BIT(TR_DisableLoopReplicatorColdSideEntryCheck), "P"},
    {"disableLoopStrider",                 "O\tdisable loop strider",                           TR::Options::disableOptimization, loopStrider, 0, "P"},
+   {"disableloopTransfer",                "O\tdisable the loop transfer part of loop versioner", SET_OPTION_BIT(TR_DisableLoopTransfer), "F"},
```
0xdaryl: Option should be "disableLoopTransfer" for case consistency.

JamesKingdon: It certainly should. Thanks for catching.

0xdaryl:
Approved, but CI needs to succeed before merging.

vijaysun-omr:
Merging this since Travis CI build has passed.

PR #1444, opened 2017-08-01
Simplify byte, short, int, and long comparisons

NigelYiboYu:
Remove integer arithmetics in integer comparisons to reduce the number of add/sub
instructions emitted.

e.g. an addition can be avoided if x-1>0 is transformed into x>1.

### DIFF
--- a/compiler/control/OMROptions.hpp
+++ b/compiler/control/OMROptions.hpp
@@ -386,7 +386,7 @@ enum TR_CompilationOptions
    TR_VerboseInlineProfiling              = 0x00004000 + 9,
    // Available                           = 0x00008000 + 9,
    // Available                           = 0x00010000 + 9,
-   // Available                           = 0x00020000 + 9,
+   TR_DisableIntegerCompareSimplification = 0x00020000 + 9,
    TR_DisableAutoSIMD                      = 0x00040000 + 9,
    TR_DisableOOL                          = 0x00080000 + 9,
    TR_DisableWriteBarriersRangeCheck      = 0x00100000 + 9,
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -12801,6 +12801,196 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
    return node;
    }
 
+
+/**
+ * \brief
+ *
+ * Remove add/subtract operations under comparison nodes to save an
+ * addition or subtraction instruction during codegen. This simplification
+ * applies to:
+ *
+ * 1. Signed comparisons of byte, short, int, and long values.
+ * 2. Unsigned (in)equality of byte, short, int, and long values.
+ *
+ * \details
+ *
+ * \verbatim
+ * As an example, assume the following is the original expression
+ *        x + oldConst1 > oldConst2
+ *
+ * This expression can be simplified to
+ *        x > newConst
+ *        where newConst = oldConst2 - oldConst1
+ * The newConst calculation should not produce overflow or underflow.
+ * \endverbatim
+ *
+ * Also note that for unsigned comparison types other than equality and inequality,
+ * add/sub simplifications are not possible due to how values are interpreted
+ * in the current IL nodes. Essentially, the value of an iconst node is neither
+ * signed nor unsigned; and its parent is free to interpret its value.
+ *
+ * For example, the following transformation is incorrect:
+ *
+ * \varbatim
+ *
+ * Before:
+ *
+ * ifiucmplt --> goto ...
+ *     isub (cannotOverflow)
+ *         iload x
+ *         iconst -1
+ *     iconst 100
+ *
+ * After:
+ *
+ * ifiucmplt --> goto ...
+ *     iload x
+ *     iconst 99
+ *
+ * \endverbatim
+ *
+ * If x were -1, the first tree would evaluate to true since it's comparing if zero is less than 100.
+ * But in the second tree, ifiucmplt interprets the value of x as UINT_MAX. Thus,
+ * the second tree is comparing UINT_MAX with 99; and the whole comparison node evalutes
+ * to false.
+ *
+ * Unsigned equality and inequality comparisons are not sensitive to sign interpretation problems.
+ *
+ * \param node          the comparison node being simpilfied
+ * \param s             the simplifier object
+*/
+TR::Node* removeArithmeticsUnderIntegralCompare(TR::Node* node,
+                                                TR::Simplifier * s)
+   {
+   if (s->comp()->getOption(TR_DisableIntegerCompareSimplification) ||
+         !(node->getOpCode().isBooleanCompare()   // The node must be an integer comparison node.
+            && node->getNumChildren() > 0
+            && node->getFirstChild()->getOpCode().isInteger()))
+      return node;
+
+   TR::Node* opNode      = node->getFirstChild();
+   TR::Node* secondChild = node->getSecondChild();
+   TR::ILOpCodes op      = opNode->getOpCodeValue();
+
+   // Get recognized operation types. Unsigned IL OpCodes are deprecated and not supported here.
+   bool isAddOp = (op == TR::iadd || op == TR::ladd || op == TR::sadd || op == TR::badd);
+   bool isSubOp = (op == TR::isub || op == TR::lsub || op == TR::ssub || op == TR::bsub);
+
+   bool isUnsignedCompare = node->getOpCode().isUnsignedCompare();
+   bool isEqOrNotEqCompare = node->getOpCode().isCompareForEquality();
+
+   bool isSignednessTypeSupported = isEqOrNotEqCompare || (!isUnsignedCompare && opNode->cannotOverflow());
+
+   if ((isAddOp || isSubOp)
+         && isSignednessTypeSupported
+         && opNode->getSecondChild()->getOpCode().isLoadConst()
+         && secondChild->getOpCode().isLoadConst()
+         // Only perform the simplification on the first occurance of the add/sub node
+         && (opNode->getFutureUseCount() == opNode->getReferenceCount() - 1))
+      {
+      int64_t signedMax = 0, signedMin = 0;
+      uint64_t oldUConst1 = 0, oldUConst2 = 0, newUConst = 0;
+
+      if (opNode->getOpCode().is1Byte())
+         {
+         oldUConst1 = opNode->getSecondChild()->getUnsignedByte();
+         oldUConst2 = secondChild->getUnsignedByte();
+
+         signedMax = TR::getMaxSigned<TR::Int8>();
+         signedMin = TR::getMinSigned<TR::Int8>();
+         }
+      else if (opNode->getOpCode().is2Byte())
+         {
+         oldUConst1 = opNode->getSecondChild()->getUnsignedShortInt();
+         oldUConst2 = secondChild->getUnsignedShortInt();
+
+         signedMax = TR::getMaxSigned<TR::Int16>();
+         signedMin = TR::getMinSigned<TR::Int16>();
+         }
+      else if (opNode->getOpCode().is4Byte())
+         {
+         oldUConst1 = opNode->getSecondChild()->getUnsignedInt();
+         oldUConst2 = secondChild->getUnsignedInt();
+
+         signedMax = TR::getMaxSigned<TR::Int32>();
+         signedMin = TR::getMinSigned<TR::Int32>();
+         }
+      else if (opNode->getOpCode().is8Byte())
+         {
+         oldUConst1 = opNode->getSecondChild()->getUnsignedLongInt();
+         oldUConst2 = secondChild->getUnsignedLongInt();
+
+         signedMax = TR::getMaxSigned<TR::Int64>();
+         signedMin = TR::getMinSigned<TR::Int64>();
+         }
+      else
+         {
+         if (s->trace())
+            traceMsg(s->comp(),
+                     "\nEliminating add/sub under compare node n%dn failed due to opcode data type\n",
+                     node->getGlobalIndex());
+
+         return node;
+         }
+
+      if (!isEqOrNotEqCompare)
+         {
+         // check for signed overflow
+         // Signed integer overflow is undefined behavior in C++11; and unsigned
+         // arithmnetic operations are defined. Need to make sure that signed integer results don't overflow.
+         int64_t oldConst1 = static_cast<int64_t>(oldUConst1);
+         int64_t oldConst2 = static_cast<int64_t>(oldUConst2);
+
+         bool canTransformAdd = isAddOp
+                 && !(oldConst1 > 0 && (oldConst2 < signedMin + oldConst1))
+                 && !(oldConst1 < 0 && (oldConst2 > signedMax + oldConst1));
+
+         bool canTransformSub = isSubOp
+                     && !(oldConst1 > 0 && (oldConst2 > signedMax - oldConst1))
+                     && !(oldConst1 < 0 && (oldConst2 < signedMin - oldConst1));
+
+         if (!(canTransformAdd || canTransformSub))
+            {
+            if (s->trace())
+               traceMsg(s->comp(),
+                        "\nEliminating add/sub under order compare node n%dn failed due to overflow\n",
+                        node->getGlobalIndex());
+
+            return node;
+            }
+         }
+
+      // Calculate new constant value.
+      // For both signed and unsigned comparisons, the new values are calculated using
+      // the old unsigned values. The reason being that signed integer overflow is undefined behavior in
+      // C++, and that unsigned integer overflow is defined behavior: they shall obey the laws of
+      // arithmetic modulo 2^n and produce wrapped values
+      // The IL expects integer wrapping in case of overflow; and this matches unsigned integer behaviors in C++.
+      newUConst = isAddOp ? (oldUConst2 - oldUConst1) : (oldUConst2 + oldUConst1);
+
+      // Done checking constant values. Transform the tree.
+      if (performTransformation(s->comp(),
+                                "%sEliminating add/sub operation under integer comparison node n%dn %s\n",
+                                s->optDetailString(),
+                                node->getGlobalIndex(),
+                                node->getOpCode().getName()))
+         {
+         // It's not recommended to mutate the value of an existing constant node by
+         // calling constNode->setConstValue()
+         // Hence, creating a new node here and copy the old node's internal info.
+         TR::Node* newConstNode = TR::Node::create(secondChild, secondChild->getOpCodeValue(), 0);
+         newConstNode->setUnsignedLongInt(newUConst);
+         node->setAndIncChild(0, opNode->getFirstChild());
+         node->setAndIncChild(1, newConstNode);
+
+         secondChild->decReferenceCount();
+         opNode->recursivelyDecReferenceCount();
+         }
+      }
+
+   return node;
+   }
+
 //---------------------------------------------------------------------
 // Integer if compare equal (signed and unsigned)
 //
@@ -12913,6 +13103,7 @@ TR::Node *ificmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
 
 
    addressCompareConversion(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    if (s->getLastRun())
       convertToTestUnderMask(node, block, s);
@@ -13017,6 +13208,7 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
 
 
    addressCompareConversion(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
 
    return node;
@@ -13061,6 +13253,7 @@ TR::Node *ificmpltSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       }
 
 
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13103,6 +13296,7 @@ TR::Node *ificmpleSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       unsignedIntCompareNarrower(node, s, TR::ifsucmple, TR::ifbucmple);
       }
 
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    partialRedundantCompareElimination(node, block, s);
    return node;
@@ -13146,6 +13340,7 @@ TR::Node *ificmpgtSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       unsignedIntCompareNarrower(node, s, TR::ifsucmpgt, TR::ifbucmpgt);
       }
 
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    partialRedundantCompareElimination(node, block, s);
    return node;
@@ -13193,6 +13388,7 @@ TR::Node *ificmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
    if (ifxcmpReducer.isReducible())
       node = ifxcmpReducer.reduce();
 
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13231,6 +13427,7 @@ TR::Node *iflcmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       longCompareNarrower(node, s, TR::ificmpeq, TR::ifsucmpeq, TR::ifscmpeq, TR::ifbcmpeq);
       }
 
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13268,6 +13465,7 @@ TR::Node *iflcmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       longCompareNarrower(node, s, TR::ificmpne, TR::ifsucmpne, TR::ifscmpne, TR::ifbcmpne);
       }
    addressCompareConversion(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    partialRedundantCompareElimination(node, block, s);
    return node;
@@ -13305,6 +13503,7 @@ TR::Node *iflcmpltSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       if (firstChild->getOpCode().isLoadConst() && conditionalBranchFold((((uint64_t)originalFirst->getLongInt())<((uint64_t)originalSecond->getLongInt())), node, firstChild, secondChild, block, s))
          return node;
       }
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13341,6 +13540,7 @@ TR::Node *iflcmpleSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       if (firstChild->getOpCode().isLoadConst() && conditionalBranchFold((((uint64_t)originalFirst->getLongInt())<=((uint64_t)originalSecond->getLongInt())), node, firstChild, secondChild, block, s))
          return node;
       }
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13377,6 +13577,7 @@ TR::Node *iflcmpgtSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
       if (firstChild->getOpCode().isLoadConst() && conditionalBranchFold((((uint64_t)originalFirst->getLongInt())>((uint64_t)originalSecond->getLongInt())), node, firstChild, secondChild, block, s))
          return node;
       }
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13419,6 +13620,7 @@ TR::Node *iflcmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier 
    if (ifxcmpReducer.isReducible())
       node = ifxcmpReducer.reduce();
 
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
    return node;
    }
@@ -13812,6 +14014,7 @@ TR::Node *ifCmpWithEqualitySimplifier(TR::Node * node, TR::Block * block, TR::Si
          }
       }
    bitTestingOp(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
 
    return node;
@@ -13912,6 +14115,7 @@ TR::Node *ifCmpWithoutEqualitySimplifier(TR::Node * node, TR::Block * block, TR:
          }
       }
    bitTestingOp(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
    partialRedundantCompareElimination(node, block, s);
 
    return node;
@@ -14067,6 +14271,8 @@ TR::Node *icmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
          }
       }
 
+   removeArithmeticsUnderIntegralCompare(node, s);
+
    return node;
    }
 
@@ -14088,7 +14294,7 @@ TR::Node *icmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    orderChildren(node, firstChild, secondChild, s);
-
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
@@ -14111,7 +14317,7 @@ TR::Node *icmpltSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
-
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
@@ -14134,7 +14340,7 @@ TR::Node *icmpleSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
-
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
@@ -14157,7 +14363,7 @@ TR::Node *icmpgtSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
-
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
@@ -14180,7 +14386,7 @@ TR::Node *icmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
-
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
@@ -14256,6 +14462,7 @@ TR::Node *lcmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
       }
 
    orderChildren(node, firstChild, secondChild, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
 
    return node;
    }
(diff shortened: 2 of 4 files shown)


compiler/control/OMROptions.hpp
```diff
@@ -307,7 +307,7 @@ enum TR_CompilationOptions
    TR_DisableNewLoopTransfer              = 0x10000000 + 6, // loop versioning for virtual guards
    TR_UseSamplingJProfilingForDLT                 = 0x20000000 + 6,
    TR_UseSamplingJProfilingForInterpSampledMethods= 0x40000000 + 6,
-   // Available                           = 0x80000000 + 6,
+   TR_disableIntegerCompareSimplification = 0x80000000 + 6,
```
fjeremic: Lets be consistent here and capitalize `Disable`.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -73,7 +73,19 @@
...
 
+// Returns true if signed 64-bit a - b has overflow or underflow
+#define IS_INT64_SUB_OVERFLOW(a, b)  (b < 0 && a > TR::getMaxSigned<TR::Int64>() + b)
+#define IS_INT64_SUB_UNDERFLOW(a, b) (b > 0 && a < TR::getMinSigned<TR::Int64>() + b)
+
+// Returns true if unsigned 64-bit a - b has underflow
+#define IS_UINT64_SUB_UNDERFLOW(a, b) (a < b)
+
+// Returns true if unsigned 64-bit a + b has overflow
+#define IS_UINT64_ADD_OVERFLOW(a, b) (a > TR::getMaxUnsigned<TR::Int64>() - b)
```
fjeremic: These seem like great utility functions. Perhaps another reviewer may be able to suggest a more visible place to put these?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+   bool isSub = (op == TR::isub || op == TR::lsub || op == TR::iusub || op == TR::lusub ||
+                 op == TR::ssub || op == TR::bsub || op == TR::busub);
+
+   if ((isAdd || isSub) &&
+           firstChild->chkCannotOverflow()  &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
+           secondChild->getOpCode().isLoadConst() &&
+           firstChild->isSingleRef() &&
+           !(!firstChild->isSingleRef() &&
+             firstChild->getFutureUseCount() != firstChild->getReferenceCount() - 1))
```
fjeremic:
```
!(!firstChild->isSingleRef() && firstChild->getFutureUseCount() != firstChild->getReferenceCount() - 1)
```

can be simplified to

```
(firstChild->isSingleRef() || firstChild->getFutureUseCount() == firstChild->getReferenceCount() - 1)
```

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+            unsignedMin = isUnsigned  ? TR::getMinUnsigned<TR::Int8>() : 0;
+            }
+         else if (firstChild->getOpCode().is2Byte())
+            {
+            signedMax   = !isUnsigned ? TR::getMaxSigned<TR::Int16>()   : 0;
+            signedMin   = !isUnsigned ? TR::getMinSigned<TR::Int16>()   : 0;
+            unsignedMax = isUnsigned  ? TR::getMaxUnsigned<TR::Int16>() : 0;
+            unsignedMin = isUnsigned  ? TR::getMinUnsigned<TR::Int16>() : 0;
+            }
+         else if (firstChild->getOpCode().is4Byte())
```
fjeremic: Why is the 64-bit int any different than the above cases?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+*/
+TR::Node* removeArithmeticsUnderIntegralCompare(TR::Node* node,
+                                                TR::Simplifier * s,
+                                                bool isInt)
+   {
+   TR::Node* firstChild     = node->getFirstChild();
+   TR::Node* secondChild    = node->getSecondChild();
+   TR::ILOpCodes op         = firstChild->getOpCodeValue();
+
+   bool isAdd = (op == TR::iadd || op == TR::ladd || op == TR::iuadd || op == TR::luadd ||
```
andrewcraik: the unsigned add and sub operations are deprecated - adding is done the same regardless of signed-ness. Only mod/div/conversion worry about signedness.

Why are these not using opcode property tests?

andrewcraik:
@jdmpapin a review please

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+           firstChild->chkCannotOverflow()  &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
+           secondChild->getOpCode().isLoadConst() &&
+           firstChild->isSingleRef() &&
+           !(!firstChild->isSingleRef() &&
+             firstChild->getFutureUseCount() != firstChild->getReferenceCount() - 1))
+      {
+      bool isUnsigned = firstChild->getOpCode().isUnsigned();
+
+      int64_t const1 = 0, const2 = 0, const3 = 0;
```
andrewcraik: These really need better names

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+ * Remove add/subtract operations under comparison nodes to save an
+ * add/sub instruction during codegen.
+ *
+ * \details
+ *
+ * e.g. This following is a before-after comparison.
+ *
+ * \verbatim
+ * Before:
+ * ificmple --> block_17 BBStart at n13n ()
```
andrewcraik: remove the node reference (eg n13n) as it is not relevant to these examples.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12759,6 +12960,12 @@ TR::Node *ificmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
 
 
    addressCompareConversion(node, s);
+
+   if (s->getLastRun() && !s->comp()->getOption(TR_disableIntegerCompareSimplification))
```
andrewcraik: I am not a fan of this control logic being sprayed everywhere - there should be an early return in removeArithmeticsUnderIntegralCompare so that changes to this logic are easily made.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13209,6 +13457,7 @@ TR::Node *iflcmpgtSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
    TR::Node *originalSecond = secondChild;
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
+
```
andrewcraik: please clean up this whitespace change.

compiler/optimizer/OMRSimplifierHandlers.cpp
andrewcraik: I am really not a fan of macros - I don't see why we need them in this case.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+   bool isAdd = (op == TR::iadd || op == TR::ladd || op == TR::iuadd || op == TR::luadd ||
+                 op == TR::sadd || op == TR::badd || op == TR::buadd);
+
+   bool isSub = (op == TR::isub || op == TR::lsub || op == TR::iusub || op == TR::lusub ||
+                 op == TR::ssub || op == TR::bsub || op == TR::busub);
+
+   if ((isAdd || isSub) &&
+           firstChild->chkCannotOverflow()  &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
```
andrewcraik: how do we know getSecondChild is valid to call?

andrewcraik:
Overall, a reasonable idea. I think the current implementation is quite verbose and could do with some simplification to make it easier to follow. The overflow checks for the int64/uint64 case worry me, but I haven't figure out if that is warranted or not yet. @jdmpapin may have some thoughts on this.

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: > I am really not a fan of macros

I would tend to agree. They offer no benefit over inline functions which are properly typed.

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: `isAdd` and `isSub` are about `firstChild`, not `node`

andrewcraik: ah missed that - all good.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+           secondChild->getOpCode().isLoadConst() &&
+           firstChild->isSingleRef() &&
```
jdmpapin: This `isSingleRef()` seems unintentional. It makes the following check against future use count irrelevant

NigelYiboYu: Yes. Thanks for pointing this out.

jdmpapin: It might be possible to remove both calls to `isSingleRef()`, since the only occurrence of a singly-reference node is its first occurrence anyway

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+ * The iadd node should be marked as 'cannotOverflow' to ensure functional correctness.
+ *
+ * \param node          the comparison node being simpilfied
+ * \param s             the simplifier object
+ * \param isInt         true if the integral types is less than or euqal to 32-bit wide;
+ *                      false if it's 64-bit.
+*/
+TR::Node* removeArithmeticsUnderIntegralCompare(TR::Node* node,
+                                                TR::Simplifier * s,
+                                                bool isInt)
```
jdmpapin: Can't `isInt` be determined from `node`? It has to be well-typed

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+
+   if ((isAdd || isSub) &&
+           firstChild->chkCannotOverflow()  &&
+           firstChild->getSecondChild()->getOpCode().isLoadConst() &&
+           secondChild->getOpCode().isLoadConst() &&
+           firstChild->isSingleRef() &&
+           !(!firstChild->isSingleRef() &&
+             firstChild->getFutureUseCount() != firstChild->getReferenceCount() - 1))
+      {
+      bool isUnsigned = firstChild->getOpCode().isUnsigned();
```
jdmpapin: As Andrew mentioned above, addition and subtraction are independent of signedness, and the unsigned opcodes are deprecated. The relevant thing here would be whether `node->isUnsignedCompare()`.

Unfortunately, `cannotOverflow` tells us that the operation doesn't signed-overflow, and I don't think it has an unsigned version, so we may have to give up on unsigned inequalities.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+
+         int64_t signedMax = 0, signedMin = 0;
+         uint64_t unsignedMax = 0, unsignedMin = 0;
+
+         if (firstChild->getOpCode().is1Byte())
+            {
+            signedMax   = !isUnsigned ? TR::getMaxSigned<TR::Int8>()   : 0;
+            signedMin   = !isUnsigned ? TR::getMinSigned<TR::Int8>()   : 0;
+            unsignedMax = isUnsigned  ? TR::getMaxUnsigned<TR::Int8>() : 0;
+            unsignedMin = isUnsigned  ? TR::getMinUnsigned<TR::Int8>() : 0;
```
jdmpapin: Is it important to set these to zero when `isUnsigned` disagrees? it looks like we just won't consult `signed{Min,Max}` when `isUnsigned`, and vice versa

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: Smaller integers are extended to 64 bits for extra range so that we can check the true result. For integers that are already 64 bits wide, it would take an even wider integer type to do that

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+            return node;
+            }
+         }
+      else  // Long
+         {
+         if (isUnsigned)
+            {
+            uint64_t uconst1 = firstChild->getSecondChild()->getUnsignedLongInt();
+            uint64_t uconst2 = secondChild->getUnsignedLongInt();
+            uint64_t uconst3 = 0;
```
jdmpapin: This `uconst3` is unused after assigned, and the value of `const3` is left unchanged in this branch

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+               }
+            else
+               {
+               return node;
+               }
+            }
+         }
+
+      // Done checking constant values. Change the nodes.
+      dumpOptDetails(s->comp(), "%sEliminating add/sub operation under comparison node n%dn\n",
```
jdmpapin: Can we use `performTransformation` here?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+         {
+         if (isUnsignedLong)
+            {
+            secondChild->setUnsignedLongInt(const3);
+            }
+         else
+            {
+            secondChild->setConstValue(const3);
+            }
+         }
```
jdmpapin: A node shouldn't be transmuted in a way that changes its value, so this should always create a new constant

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: Also, does this transformation have to be restricted to the last run? Seems like it could happen earlier

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: Although `eq` and `ne` comparisons can be transformed regardless of signedness and whether or not there is overflow.

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: > A node shouldn't be transmuted in a way that changes its value, so this should always create a new constant

Curious why this API exist then for `isLoadConst` IL opcodes?

andrewcraik: the setConstValue exists for cases where you are build trees in ilgen and don't know the constant and will fill it in later. Not an ideal usage scenario, but there you go. The use outside of ilgen is not permitted for the reasons Devin stipulated.

jdmpapin: Historically there has been code that changes constant values in-place, but it causes issues. For example, from time to time I've seen constants with `(X>=0 X<=0 X!=0)`.

The method is still useful for creating a new constant node in this case, since there is currently no `Node::create` for integer constants of arbitrary size. There is a variant that takes `int32_t intValue`, but it claims to be invalid for smaller integer types, and it's unsuitable for long constants.

We also need to set the value when transmuting a previously non-constant node into a constant. For example, with

    iadd
      iconst 3
      iconst 4

We do something like the following, which does not change the node's value:

    iaddNode->removeAllChildren();
    TR::Node::recreate(iaddNode, TR::iconst);
    iaddNode->setInt(7); // computed from the children, of course

It could have been set by a variant of `recreate` if we had one, and that might be an improvement, but having access to this method is no worse in this sense than having access to `setAndIncChild()`.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12647,6 +12659,195 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+      else
+         {
+         TR::Node* newConstNode = TR::Node::create(secondChild->getOpCodeValue(), 0);
```
jdmpapin: This should have bci specified

vijaysun-omr:
What is the latest on this PR ? It had WIP in the title which usually means I don't start reviewing. Just making sure it really is a work in progress in which case I will just wait for the work to progress.

andrewcraik:
@NigelYiboYu ping? is this stale or being worked on?

NigelYiboYu:
This has been the lowest priority on my list. I'll work on it now and get this pull request refreshed.

NigelYiboYu:
I've fixed some errors and addressed some concerns from previous reviews. Please review @jdmpapin

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+   TR::Node* opNode      = node->getFirstChild();
+   TR::Node* secondChild = node->getSecondChild();
+   TR::ILOpCodes op      = opNode->getOpCodeValue();
+   bool isInt            = !opNode->getOpCode().isLong();
+
+   // Get recognized operation types. Unsigned IL OpCodes are deprecated and not supported here.
+   bool isAddOp = (op == TR::iadd || op == TR::ladd || op == TR::sadd || op == TR::badd);
+   bool isSubOp = (op == TR::isub || op == TR::lsub || op == TR::ssub || op == TR::bsub);
+
+   if ((isAddOp || isSubOp) &&
```
andrewcraik: this is easier to read if you put the && at the start of each line rather than the end of the line before. The indent seems very large here in the browser - can you line up on the ( of (isAddOp || .. or something else reasonable in that area - indent is 3 levels at a time in this code :)

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+                     !(oldConst1 > 0 && (oldConst2 > TR::getMaxSigned<TR::Int64>() - oldConst1)) &&
+                     !(oldConst1 < 0 && (oldConst2 < TR::getMinSigned<TR::Int64>() - oldConst1)))
+               newConst = oldConst2 + oldConst1;
+            else
+               return node;
+            }
+         }
+
+      // Done checking constant values. Transform the tree.
+      performTransformation(s->comp(), "%sEliminating add/sub operation under integer comparison node n%dn\n",
```
andrewcraik: this needs to be in an if so it can be diabled

andrewcraik:
I'll let @jdmpapin make a pass before reading more closely. The code could use more tracing when simplifier tracing is enabled - it should log the case detected, the transformation being done etc to make debugging from the log easy.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+            signedMax   = TR::getMaxSigned<TR::Int32>();
+            signedMin   = TR::getMinSigned<TR::Int32>();
+            unsignedMax = TR::getMaxUnsigned<TR::Int32>();
+            unsignedMin = TR::getMinUnsigned<TR::Int32>();
+            }
+         else
+            return node;
+
+         if ((!isUnsignedCompare && (newConst > signedMax || newConst < signedMin)) ||
+                 (isUnsignedCompare && (newConst > unsignedMax || newConst < unsignedMin)))
```
andrewcraik: indent here seems large

compiler/optimizer/OMRSimplifierHandlers.cpp
NigelYiboYu: The function `removeArithmeticsUnderIntegralCompare` will return right away if this simplification is disabled. I don't think we need another `if` around this line.

jdmpapin: The purpose of `performTransformation` is to allow disabling a particular occurrence of a transformation without disabling the class of transformation, e.g. so that it's possible to debug with `lastOptSubIndex`

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: This overflow test is sensitive to whether the old constants were set using `setInt()` or `setUnsignedInt()` (or similar for narrower types). For example, suppose we have a signed 32-bit comparison `x + 1 <= INT_MIN`. If this `INT_MIN` comes from `setInt()`, then overflow would be correctly detected. OTOH, if it comes from `setUnsignedInt()`, then I believe we'd get `newConst == (int64_t)INT_MAX`, which would incorrectly appear not to have overflowed

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+            if (isAddOp &&
+                    !(oldConst1 > 0 && (oldConst2 < TR::getMinSigned<TR::Int64>() + oldConst1)) &&
+                    !(oldConst1 < 0 && (oldConst2 > TR::getMaxSigned<TR::Int64>() + oldConst1)))
+               newConst = oldConst2 - oldConst1;
+            else if (isSubOp &&
+                     !(oldConst1 > 0 && (oldConst2 > TR::getMaxSigned<TR::Int64>() - oldConst1)) &&
+                     !(oldConst1 < 0 && (oldConst2 < TR::getMinSigned<TR::Int64>() - oldConst1)))
+               newConst = oldConst2 + oldConst1;
+            else
+               return node;
```
jdmpapin: The overflow checking logic here looks good! 👍 

Is it possible to abstract this a bit so that it can be used for the narrower cases as well? Widening to prevent overflow is good, but since this code is still needed anyway, maybe it can be used to simplify the function

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+         }
+
+      // Done checking constant values. Transform the tree.
+      performTransformation(s->comp(), "%sEliminating add/sub operation under integer comparison node n%dn\n",
+                            s->optDetailString(),
+                            node->getGlobalIndex());
+
+      // Handle the iconst node in the compare
+      TR::Node* newConstNode = TR::Node::create(secondChild->getOpCodeValue(), 0);
+      newConstNode->setByteCodeIndex(secondChild->getByteCodeIndex());
```
jdmpapin: `newConstNode` needs to have its entire bytecode info set, not just the index. Since it's created right here, probably the most convenient way would be to pass an "originating bytecode node" to `TR::Node::create()`, e.g. with this overload:

    TR::Node *
    OMR::Node::create(TR::Node * originatingByteCodeNode, TR::ILOpCodes op, uint16_t numChildren)

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12922,6 +13065,7 @@ TR::Node *ificmpeqSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
 
 
    addressCompareConversion(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
```
jdmpapin: It looks like `addressCompareConversion()` can transmute `node` into an address comparison. At the moment `removeArithmeticsUnderIntegralCompare()` should ignore address comparisons due to its explicit list of accepted left child opcodes. But if you want to use `isAdd()`, then just be aware that `aladd` and `aiadd` respond with true

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+ * \endverbatim
+ *
+ *  And the newConst calculation should not produce overflow or underflow.
+ *
+ * \param node          the comparison node being simpilfied
+ * \param s             the simplifier object
+*/
+TR::Node* removeArithmeticsUnderIntegralCompare(TR::Node* node,
+                                                TR::Simplifier * s)
+   {
```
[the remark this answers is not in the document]
NigelYiboYu: `removeArithmeticsUnderIntegralCompare()` is only invoked in byte, short, integer or long comparison simplifer handlers. Besides, it tests the first child of the `node` to see if it's an `iadd`, `isub`, and etc.  So I don't think we need to test the comparison node.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,149 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+      return node;
+
+   TR::Node* opNode      = node->getFirstChild();
+   TR::Node* secondChild = node->getSecondChild();
+   TR::ILOpCodes op      = opNode->getOpCodeValue();
+   bool isInt            = !opNode->getOpCode().isLong();
+
+   // Get recognized operation types. Unsigned IL OpCodes are deprecated and not supported here.
+   bool isAddOp = (op == TR::iadd || op == TR::ladd || op == TR::sadd || op == TR::badd);
+   bool isSubOp = (op == TR::isub || op == TR::lsub || op == TR::ssub || op == TR::bsub);
```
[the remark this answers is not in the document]
NigelYiboYu: `TR::ILOpCode::isAdd()` returns true for int, long, double,  and all other additions.. The spectrum is broader than what `removeArithmeticsUnderIntegralCompare()` can handle. It's only removing add/sub from Integer types.

compiler/optimizer/OMRSimplifierHandlers.cpp
NigelYiboYu: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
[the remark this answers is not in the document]
andrewcraik: I think an assert is probably warranted as an additional safety measure just in case the function use is changed or abused later. Use an assert that is not enabled in prod builds would be my vote.

compiler/optimizer/OMRSimplifierHandlers.cpp
andrewcraik: +1 the perform transformation needs to be the condition of an if statement guarding the rest of the code.

compiler/optimizer/OMRSimplifierHandlers.cpp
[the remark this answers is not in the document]
fjeremic: I agree. If you make an implicit assumption, enforce it with a `TR_ASSERT`.

NigelYiboYu:
Update:

1. Do no handle unsigned comparison nodes except for equality and inequality tests
2. Add `TR_ASSERT` to test opCode and prevent misuse of this tree simplification function.
3. Generalized constant calculation overflow detection logic so that it can apply to 1,2,4,8 byte types.

NigelYiboYu:
Added the add/sub removal for all these combinations of IL op codes:   `[l, i, lu, iu] cmp [gt, lt, ge, le, eq, ne]`

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,235 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+      else
+         {
+         if (isAddOp
+               && !(oldConst1 > 0 && (oldConst2 < signedMin + oldConst1))
+               && !(oldConst1 < 0 && (oldConst2 > signedMax + oldConst1)))
+            newConst = oldConst2 - oldConst1;
+         else if (isSubOp
+                     && !(oldConst1 > 0 && (oldConst2 > signedMax - oldConst1))
+                     && !(oldConst1 < 0 && (oldConst2 < signedMin - oldConst1)))
+            newConst = oldConst2 + oldConst1;
```
jdmpapin: Thanks for combining these paths :)

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: With the new assertion at the beginning of `removeArithmeticsUnderIntegralCompare()`, this should guard against passing in `node` when it has been transmuted into an address comparison

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13026,6 +13256,7 @@ TR::Node *ificmpneSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
 
 
    addressCompareConversion(node, s);
+   removeArithmeticsUnderIntegralCompare(node, s);
```
jdmpapin: There's another call to `addressCompareConversion()` here

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,235 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+         {
+         oldConst1 = opNode->getSecondChild()->getLongInt();
+         oldConst2 = secondChild->getLongInt();
+
+         oldUConst1 = opNode->getSecondChild()->getUnsignedLongInt();
+         oldUConst2 = secondChild->getUnsignedLongInt();
+
+         signedMax   = TR::getMaxSigned<TR::Int64>();
+         signedMin   = TR::getMinSigned<TR::Int64>();
+         unsignedMax = TR::getMaxUnsigned<TR::Int64>();
```
jdmpapin: It might be good to use a template and `getConst<T>()` to remove this duplication

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,235 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+                          || node->getOpCodeValue() == TR::lcmpne
+                          || node->getOpCodeValue() == TR::icmpeq
+                          || node->getOpCodeValue() == TR::icmpne
+                          || node->getOpCodeValue() == TR::lucmpeq
+                          || node->getOpCodeValue() == TR::lucmpne
+                          || node->getOpCodeValue() == TR::iucmpeq
+                          || node->getOpCodeValue() == TR::iucmpne;
+
+   bool isSignednessTypeSupported = (!isUnsignedCompare && opNode->chkCannotOverflow())
+           || (isUnsignedCompare && isEqOrNotEqCompare);
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12810,6 +12810,235 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+   bool isEqOrNotEqCompare = TR::ILOpCode::isEqualCmp(node->getOpCodeValue())
+                          || TR::ILOpCode::isNotEqualCmp(node->getOpCodeValue())
+                          || node->getOpCodeValue() == TR::lcmpeq
+                          || node->getOpCodeValue() == TR::lcmpne
+                          || node->getOpCodeValue() == TR::icmpeq
+                          || node->getOpCodeValue() == TR::icmpne
+                          || node->getOpCodeValue() == TR::lucmpeq
+                          || node->getOpCodeValue() == TR::lucmpne
+                          || node->getOpCodeValue() == TR::iucmpeq
+                          || node->getOpCodeValue() == TR::iucmpne;
```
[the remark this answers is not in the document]
NigelYiboYu: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12804,6 +12804,220 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+      // Thus, equality comparisons' new values can be obtained using unsigned old values.
+      if (isEqOrNotEqCompare)
+         {
+         newUConst = isAddOp ? (oldUConst2 - oldUConst1) : (oldUConst2 + oldUConst1);
+         newConst = static_cast<int64_t>(newUConst);
+         }
+      else
+         {
+         newConst = isAddOp ? (oldConst2 - oldConst1) : (oldConst2 + oldConst1);
+         }
```
jdmpapin: The logic for the `isEqOrNotEqCompare` case should work in all cases

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12804,6 +12804,220 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+         && isSignednessTypeSupported
+         && opNode->getSecondChild()->getOpCode().isLoadConst()
+         && secondChild->getOpCode().isLoadConst()
+         // Only perform the simplification on the first occurance of the add/sub node
+         && (opNode->getFutureUseCount() == opNode->getReferenceCount() - 1))
+      {
+      int64_t oldConst1 = 0, oldConst2 = 0, newConst = 0;     // for signed compare types
+      int64_t signedMax = 0, signedMin = 0;
+
+      uint64_t oldUConst1 = 0, oldUConst2 = 0, newUConst = 0;
```
jdmpapin: The two sets of constants here are redundant, and they cause a lot of repeated code just below. Could we use `uint64_t` only?

vijaysun-omr:
genie-omr build all

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -12801,6 +12801,196 @@ TR::Node *su2dSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
...
+   TR::ILOpCodes op      = opNode->getOpCodeValue();
+
+   // Get recognized operation types. Unsigned IL OpCodes are deprecated and not supported here.
+   bool isAddOp = (op == TR::iadd || op == TR::ladd || op == TR::sadd || op == TR::badd);
+   bool isSubOp = (op == TR::isub || op == TR::lsub || op == TR::ssub || op == TR::bsub);
+
+   bool isUnsignedCompare = node->getOpCode().isUnsignedCompare();
+   bool isEqOrNotEqCompare = node->getOpCode().isCompareForEquality();
+
+   bool isSignednessTypeSupported = isEqOrNotEqCompare || (!isUnsignedCompare && opNode->chkCannotOverflow());
```
vijaysun-omr: Should this be using the "cannotOverflow" query ? I believe that is what the rest of the optimizer uses.

fjeremic: > Should this be using the "cannotOverflow" query ? I believe that is what the rest of the optimizer uses.

Should `chkCannotOverflow` exist then? What is it's purpose? Is this duplication we can remove?

vijaysun-omr: I believe the chk* variants of each query are called from the print* routines that exist for each Node query (to print Node flags in a log file). The chk* variants of the queries check the opcode and the flags value (both have to be checked before a particular flags value string can be printed in the log file) whereas the queries themselves check the flags value only (on the assumption that the caller is already aware that the query being done is relevant for the particular opcode).

NigelYiboYu: Changed the query to use `cannotOverflow()`.

vijaysun-omr:
Thanks for making all the requested changes. Merging.

PR #1453, opened 2017-08-01
Added Makefile rule for building JitBuilder library

sajidahmed21:
- <s>Added `operandarraytests` (which is the binary of a JitBuilder example) to the corresponding gitignore list.</s>

- Added a rule in the JitBuilder release Makefile for building the JitBuilder library, if its not present. This usability enhancement will eliminate an extra step (and any ambiguity) when trying to build the JitBuilder examples in the absence of a JitBuilder binary.

### DIFF
--- a/jitbuilder/release/Makefile
+++ b/jitbuilder/release/Makefile
@@ -21,6 +21,9 @@
 CXX?=g++
 CXXFLAGS=-g -std=c++0x -O2 -c -fno-rtti -fPIC -I./include/compiler -I./include
 
+OMR_ROOT_DIR=../..
+JIT_BUILDER_DIR=$(OMR_ROOT_DIR)/jitbuilder
+
 .SUFFIXES: .cpp .o
 
 # NOTE: replay purposefully left out of these lists
@@ -303,6 +306,8 @@ thunks : libjitbuilder.a Thunk.o
 Thunk.o: src/Thunk.cpp
 	$(CXX) -o $@ $(CXXFLAGS) $<
 
+libjitbuilder.a :
+	make -C $(JIT_BUILDER_DIR)
 
 clean:
 	@rm -f $(ALL_TESTS) replay *.o


mstoodle:
I think your first commit is also introduced in #1448

sajidahmed21:
You are right. I've removed it from this PR.

mstoodle:
The original point of the "release" directory is that it could stand by itself and was an "output" of the omr/jitbuilder project. The makefile was meant to be self-contained, which you're now breaking with this change.

Technically, there is more than one dependence here, since the .a is only part of what jitbuilder generates when it builds (it also copies header files).

Do you strongly feel this is an important addition to this Makefile?

sajidahmed21:
I see what you are saying. But in the current state, I am not sure if this Makefile is fully self-contained, since it does not build the examples if jitbuilder has not already been built.
It relies on the existence of the `libjitbuilder` binary without actually telling the user how to build one.

Although, technically there are other dependencies here (i.e. the header files), as you said they are handled as part of building jitbuilder.

I considered this to be useful, particularly for those who are new to the codebase and are trying to start off by building and running the examples.

However, if we are not comfortable with this change, I would think that we should update the README to reflect how to generate `libjitbuilder.a` (and / or output some helpful error message specifying the process).

mstoodle:
Sorry, this one dropped off my radar for some reason.

When I described it as "self contained" I meant: if you zip up everything in the "release" directory (after building libjitbuilder.a) then you can take that zip file to another machine, unzip it, and write code against the JitBuilder API.

The Makefile in the release directory is for building the code samples against a library.

In that scenario, there's nothing you can do if there is no libjitbuilder.a in the directory (just like there's nothing you can do if the appropriate headers haven't been copied into the include directory).

Even the README is meant to be describing the contents of the release directory.

How about a Makefile rule that issues a suggestive error message if libjitbuilder.a is not present in the top level directory?

mstoodle:
@sajidahmed21 if you're ok with the suggestion, maybe best for us to open an issue/PR recording the new choice and then close this PR. What do you think?

mstoodle:
Closing since the action has been captured in issue #1655 .

PR #1454, opened 2017-08-01
Use OSR method info to check for OSR blocks

jdmpapin:
IL (intermediate language) generation sometimes assumed that the IL for a method has OSR (on-stack replacement) code and catch blocks whenever `getOSRPoints()` is nonempty for that method. However, it's possible that all OSR points for the method belong to other inlined call sites, since the set of OSR points can differ when IL is generated for a single method repeatedly. For instance, symbol references that are unresolved when first inlining a method may be resolved when later inlining the same method again.

This commit changes IL generation to determine the presence or absence of the OSR code and catch block based on the OSR method info for the current inlined site index, without consulting `getOSRPoints()`.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -958,15 +958,17 @@ OMR::ResolvedMethodSymbol::genAndAttachOSRCodeBlocks(int32_t currentInlinedSiteI
          lastTreeTop = tt;
       }
 
-   //if an OSR code block has been created, attach its treetops and osr catch block's
-   //treetops to the method's treetops
-   if (!self()->getOSRPoints().isEmpty())
-      {
-      TR_OSRMethodData *osrMethodData = self()->comp()->getOSRCompilationData()->findOrCreateOSRMethodData(currentInlinedSiteIndex, self());
-      TR::Block *OSRCodeBlock = osrMethodData->getOSRCodeBlock();
-      TR::Block *OSRCatchBlock = osrMethodData->getOSRCatchBlock();
-      lastTreeTop->insertTreeTopsAfterMe(OSRCatchBlock->getEntry(), OSRCodeBlock->getExit());
-      }
+   // If OSR code and catch blocks have been created, stitch them into the
+   // method's trees. Check for the OSR blocks directly, since they're specific
+   // to the current inlined call site, unlike getOSRPoints().
+   TR::Block * const osrCodeBlock = osrMethodData->getOSRCodeBlock();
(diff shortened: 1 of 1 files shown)


andrewcraik:
@ncough your two cents - looks good to me.

ncough:
We base another decision on the size of `getOSRPoints()` in `genInduceOSRCall()` for the late addition of OSR infrastructure, however, this should never happen as indicated by the comment near that code. LGTM

jdmpapin:
Was AppVeyor going to run anything using OSR? And if so, can we restart it? It just failed to check out the commit

vijaysun-omr:
Just so I am completely clear, the case where the check that was being done originally would have been incorrect is if we had 2 different calls to the same method and these calls use the same symbol. We want to inline both calls and when we generate IL for the first call, we do not find any OSR points but when we generate IL for the second call, something (e.g. resolution) has changed leading to an OSR point being encountered. Since the symbol is the same, the second IL gen has now overwritten what the symbol had in the OSR points list from the first IL gen which leads to the wrong answer for the first inlined site. This commit gets the info from the OSR data instead and since that is specific to each inlined site, this fixes the problem.

vijaysun-omr:
@jdmpapin is my understanding correct ?

jdmpapin:
That's close. We don't clear out `getOSRPoints()`. Instead it holds the OSR points for *every* call site where the method was inlined. Each OSR point has bci that identifies the site it belongs to.

The problem scenario I saw does involve inlining the same method twice, sharing the same `ResolvedMethodSymbol`, but where the *first* site has at least one OSR point, and the *second* has none. When this happens, no OSR blocks are created for the second site, because they're unnecessary. Without this change, we would nonetheless attempt to attach the second site's (nonexistent) OSR blocks to the trees, simply because there are OSR points on the method symbol (belonging to the first site).

An OSR point can disappear when we have an unresolved symref resulting in a `ResolveCHK` where there would otherwise not have been an OSR point. When generating IL for the same method again later, we may no longer have an unresolved symref, because (say) a Java constant pool entry has been resolved in the meantime.

There could be other sources of variability in the set of OSR points generated by Java's IL generator, although I'm not aware of any. IL generators for other projects should probably be free to generate different sets of OSR points based on run-time state as well.

vijaysun-omr:
Thanks, makes sense to me.

jdmpapin:
Is there anything further I need to do before this can be merged? I [asked about AppVeyor](https://github.com/eclipse/omr/pull/1454#issuecomment-319750337) above, since I suspect that it might not test OSR anyway. Please let me know how to proceed

vijaysun-omr:
I am merging because the appveyor "failure" is not a real issue.

PR #1456, opened 2017-08-01
Cmake: Link the system thread library portably

rwy0717:
j9thrstatic is our portable library for threading. If gtest needs
threading, this is how it should be getting it.

Also link omr_core for basic OMR flags.

### DIFF
--- a/fvtest/omrGtestGlue/CMakeLists.txt
+++ b/fvtest/omrGtestGlue/CMakeLists.txt
@@ -16,9 +16,12 @@
 #    Multiple authors (IBM Corp.) - initial implementation and documentation
 ###############################################################################
 
-
 add_library(omrGtest STATIC
-	omrGtest.cpp)
+	omrGtest.cpp
+)
+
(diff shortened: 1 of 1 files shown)


charliegracie:
waiting for a thumbs up from @dnakamura or @mgaudet

dnakamura:
My concern would be on platforms where pthreads is available, but not used by `j9thrstatic` (potentially aix and z/os). Seems like this patch would break under said conditions.

mgaudet:
Curious: In the appveyor build we're seeing ref to `/pthreads`: https://ci.appveyor.com/project/eclipsewebmaster/omr/build/1.0.1316/messages

charliegracie:
@rwy0717 any updates or comments based on the the comments by @dnakamura and @mgaudet?

rwy0717:
Devin is right that this will break when j9thrstatic doesn't link pthreads.
I'm going to close this PR, we'll need to solve this another way.

It's troubling that /pthreads is showing up on the windows builds. I'll look into that as well.

PR #1457, opened 2017-08-02
[Mem14] Create a region to hold alias set BitVectors

andrewcraik:
The current aliasing infrastructure allocates bit vectors to return to
callers of the aliasing interface. These bit vectors are heap allocated
and leak to the end of the compile. This change introduces a region to
hold these bit vectors and an API on the compilation to signal when it
is safe to invalidate the region and release the memory. The change adds
a call in LocalCSE to this API as well as calling it between
optimizations to reduce memory usage.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * (c) Copyright IBM Corp. 2000, 2016
+ * (c) Copyright IBM Corp. 2000, 2017
  *
  *  This program and the accompanying materials are made available
  *  under the terms of the Eclipse Public License v1.0 and
@@ -94,6 +94,10 @@
 #include "control/Recompilation.hpp"           // for TR_Recompilation, etc
 #include "runtime/CodeCacheExceptions.hpp"
 #include "ilgen/IlGen.hpp"                     // for TR_IlGenerator
+
+// this ratio defines how full the alias memory region is allowed to become before
+// it is recreated after an optimization finishes
+#define ALIAS_REGION_LOAD_FACTOR 0.75
 
 #ifdef J9_PROJECT_SPECIFIC
 #include "control/RecompilationInfo.hpp"           // for TR_Recompilation, etc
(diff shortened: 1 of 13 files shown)


andrewcraik:
Rebased onto the merge of Mem 9 - 13 inclusive.

compiler/optimizer/OMROptimizer.cpp
```diff
@@ -2050,6 +2050,10 @@ int32_t OMR::Optimizer::performOptimization(const OptimizationStrategy *optimiza
          }
 
       delete opt;
+      // we cannot easily invalidate during IL gen since we could be peeking and we cannot destory our
```
0xdaryl: typo:  destroy

compiler/compile/OMRCompilation.cpp
```diff
@@ -2713,3 +2714,17 @@ OMR::Compilation::CompilationPhaseScope::~CompilationPhaseScope()
...
    }
+
+TR::Region &OMR::Compilation::aliasRegion()
+   {
+   return self()->_aliasRegion;
+   }
+
+void OMR::Compilation::invalidateAliasRegion()
+   {
+   if (self()->_aliasRegion.bytesAllocated() > (0.75) * TR::Region::initialSize())
```
0xdaryl: How was the `0.75` threshold chosen?  Can this constant be pulled out into at least a `#define` so its not so hidden?

andrewcraik: The 0.75 was not scientifically tuned - it was chosen to avoid recreating the region after every opt, but to try and make sure that we had sufficient space to allocate at least a few BitVectors before the region would need to allocate from the allocator. I will pull the 0.75 out into a `#define`

andrewcraik:
@0xdaryl I have updated the commit with the requested changes

andrewcraik:
Looks like LocalTransparency may have an issue with the region invalidation will update once we know more.

andrewcraik:
The issue we had with this change has now been fixed and should now be ready to merge - @0xdaryl FYI

0xdaryl:
Was the fix in this PR and requires a review?

andrewcraik:
The fix was in here - the issue was that copying the bit vector on return from the aliasing API into the heap was not being done correctly - = assignment on a TR_BitVector copies the region to allocate form as well as the data. The fix is to create the new bit vector and or in the values. Very minor change.

PR #1458, opened 2017-08-02
CMake: Add omr_assert utility

rwy0717:
Add a barebones assert function for checking assumptions in cmake.

/cc @dnakamura @mgaudet 


rwy0717:
Just fixed a small formatting issue. This should be good to go.

PR #1460, opened 2017-08-02
JitBuilder cleanup

mstoodle:
Three small corrections:

1. Reverse order of induction variable analysis and loop canonicalization in JitBuilder opt strategy, plus add a block ordering pass beforehand to clean up the blocks.
2. Workaround fix for an assertion in the OperandArrayTests code sample. The correct fix will be a broader change so I want to get this test working and into our pull request testing first.
3. For JitBuilder testing, run "make testall" on linux_x86 and osx platforms, but "test" elsewhere.

### DIFF
--- a/jitbuilder/optimizer/JBOptimizer.cpp
+++ b/jitbuilder/optimizer/JBOptimizer.cpp
@@ -117,11 +117,12 @@ static const OptimizationStrategy JBwarmStrategyOpts[] =
    { OMR::treeSimplification                                                       },
    { OMR::trivialDeadTreeRemoval,                    OMR::IfEnabled                },
 
-   { OMR::inductionVariableAnalysis,                 OMR::IfLoops                  },
-   { OMR::loopCanonicalization,                      OMR::IfLoops                  },
+   { OMR::basicBlockOrdering,                        OMR::IfLoops                  }, // clean up block order for loop canonicalization, if it will run
+   { OMR::loopCanonicalization,                      OMR::IfLoops                  }, // canonicalization must run before inductionVariableAnalysis else indvar data gets messed up
+   { OMR::inductionVariableAnalysis,                 OMR::IfLoops                  }, // needed for loop unroller
    { OMR::generalLoopUnroller,                       OMR::IfLoops                  },
-   { OMR::basicBlockExtension,                       OMR::MarkLastRun              }, // extend blocks; move trees around if reqd
-   { OMR::treeSimplification                                                       }, // revisit; not really required ?
+   { OMR::basicBlockExtension,                       OMR::MarkLastRun              }, // clean up order and extend blocks now
+   { OMR::treeSimplification                                                       },
    { OMR::localCSE                                                                 },
    { OMR::treeSimplification,                        OMR::IfEnabled                },
    { OMR::trivialDeadTreeRemoval,                    OMR::IfEnabled                },
--- a/fvtest/omrtest.mk
+++ b/fvtest/omrtest.mk
@@ -44,8 +44,18 @@ omr_algotest:
 omr_gctest:
 	./omrgctest --gtest_filter="gcFunctionalTest*"
 
+# jitbuilder can run different sets of tests on linux_x86 and osx than on other platforms
+# until we common this up, run "testall" on linux_x86 and osx but run "test" everywhere else
 omr_jitbuilderexamples:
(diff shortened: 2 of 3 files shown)


mstoodle:
Verified that the linux_x86 and osx testing for this PR ran the extended JitBuilder testing, so that's nice.

jitbuilder/optimizer/JBOptimizer.cpp
```diff
@@ -117,11 +117,12 @@ static const OptimizationStrategy JBwarmStrategyOpts[] =
    { OMR::treeSimplification                                                       },
    { OMR::trivialDeadTreeRemoval,                    OMR::IfEnabled                },
 
-   { OMR::inductionVariableAnalysis,                 OMR::IfLoops                  },
-   { OMR::loopCanonicalization,                      OMR::IfLoops                  },
+   { OMR::basicBlockOrdering                                                       }, // clean up block order going into canonicalizer
+   { OMR::loopCanonicalization,                      OMR::IfLoops                  }, // canonicalizer first, otherwise indvar data gets messed up
```
0xdaryl: Sorry for the pedantic request.  You say "canonicalizer" twice in a row.

jitbuilder/optimizer/JBOptimizer.cpp
```diff
@@ -117,11 +117,12 @@ static const OptimizationStrategy JBwarmStrategyOpts[] =
    { OMR::treeSimplification                                                       },
    { OMR::trivialDeadTreeRemoval,                    OMR::IfEnabled                },
 
-   { OMR::inductionVariableAnalysis,                 OMR::IfLoops                  },
-   { OMR::loopCanonicalization,                      OMR::IfLoops                  },
+   { OMR::basicBlockOrdering                                                       }, // clean up block order going into canonicalizer
```
0xdaryl: If BasicBlockOrdering is only needed to clean things up for the canonicalizer, should it be conditional on OMR::IfLoops too?

0xdaryl: Upon further reflection, I suppose there are other benefits to cleaning up block order outside of loop canonicalizer, though perhaps not as noticable...

jitbuilder/optimizer/JBOptimizer.cpp
mstoodle: Do you mean I should refer to it as "canonicalization" ? or should I remove the first one (since it's not specific to having loops ?)

0xdaryl: I read that as one continuous sentence with two "canonicalizer"s in a row.  If that's what you intended then please remove the first one (or second one :-)  ).  If that's not what you intended then please add a period at the end of the first comment.

jitbuilder/optimizer/JBOptimizer.cpp
mstoodle: @0xdaryl for methods without loops, not much will happen between this run of block ordering and the subsequent block extension pass (which also does block ordering). So your first suggestion was right, I think. I updated it to only run block ordering OMR::IfLoops.

jitbuilder/optimizer/JBOptimizer.cpp
mstoodle: @0xdaryl I modified the text to hopefully make it more clear. I think each line's comment pertains to that optimization and shouldn't run on multiple lines, but I'm sure we're far from consistent on this point :( .

jitbuilder/optimizer/JBOptimizer.cpp
```diff
@@ -117,11 +117,12 @@ static const OptimizationStrategy JBwarmStrategyOpts[] =
    { OMR::treeSimplification                                                       },
    { OMR::trivialDeadTreeRemoval,                    OMR::IfEnabled                },
 
-   { OMR::inductionVariableAnalysis,                 OMR::IfLoops                  },
-   { OMR::loopCanonicalization,                      OMR::IfLoops                  },
+   { OMR::basicBlockOrdering                         OMR::IfLoops                  }, // clean up block order for loop canonicalization, if it will run
```
0xdaryl: Thanks.  This needs a ',' after `OMR::basicBlockOrdering` to avoid the TravisCI break.

mstoodle: *sigh* missed the notification email on this one. Silly mistake; I'll fix it.

PR #1455, opened 2017-08-01
CMake: use OMR_JITBUILDER and OMR_TEST_COMPILER feature flags

rwy0717:
/cc @dnakamura @mgaudet

PR #1461, opened 2017-08-02
Add JitBuilder tests in CMake builds

sajidahmed21:
This change enables building and running the JitBuilder tests in
`<omr_root>/fvtest/jitbuildertest` using CMake.

### DIFF
--- a/fvtest/jitbuildertest/CMakeLists.txt
+++ b/fvtest/jitbuildertest/CMakeLists.txt
@@ -1,6 +1,6 @@
 ###############################################################################
 #
-# (c) Copyright IBM Corp. 2017
+# (c) Copyright IBM Corp. 2017, 2017
 #
 #  This program and the accompanying materials are made available
 #  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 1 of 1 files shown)


mgaudet:
Initial scan LGTM, however there is a cross dep here: https://github.com/eclipse/omr/pull/1455 

Depends on what gets merged first.

rwy0717:
Your patch needs to check for `OMR_JITBUILDER` before adding the jitbuildertest subdir. `OMR_JITBUILDER` is disabled on appveyor.

mgaudet:
(Now that #1455 is in )

sajidahmed21:
Alright, I'm adding that check. But just curious:

- How do AppVeyor and Travis builds differ (in terms of environment and the functionality that they test)?
- Why do we not want JitBuilder builds on AppVeyor?

mgaudet:
Appveyor tests windows builds (Travis can only test Linux and OS/X) 

JitBuilder has never been built on Windows. While there's not technical reason it shouldn't be doable, it's never been done thus far. As a result, the CMake work needs to exclude JitBuilder and JitBuilder's tests from Windows CI, in this case, AppVeyor. 

The features for each CI system are set inside of [`cmake/caches/`](https://github.com/eclipse/omr/tree/master/cmake/caches).

rwy0717:
This looks good, thanks!

Also I'll add that with cmake, we're (temporarily) hardcoding some linux-x86-64 constants, so the jit won't build anywhere else.

sajidahmed21:
@0xdaryl Done

PR #1464, opened 2017-08-03
[Mem16] Reduce GRA memory usage

andrewcraik:
We are seeing problems where GRA is running away with its memory usage. This commit compacts a number of data structures and reduces duplication to significantly reduce GRA's peak memory usage.

This is built on top of Mem14 and Mem15.

### DIFF
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -83,12 +83,12 @@
 
 TR::GlobalSet::GlobalSet(TR::Compilation * comp, TR::Region &region)
    :_refAutosPerBlock((RefMapComparator()),(RefMapAllocator(region))),
-    _comp(comp),_maxEntries(0)
+    _comp(comp)
    {
    }
 
 
-void TR::GlobalSet::DenseSet::print(TR::Compilation * comp)
+void TR::GlobalSet::Set::print(TR::Compilation * comp)
    {
    TR_BitVectorIterator bits(_refs);
    for (int32_t bit = bits.getFirstElement(); bits.hasMoreElements(); bit = bits.getNextElement())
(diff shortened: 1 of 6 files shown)


andrewcraik:
marking WIP due to a potential stability issue identified when running with the latest versions of OMR - will update once investigated.

andrewcraik:
@vijaysun-omr would still appreciate your thoughts on the change since I am hoping whatever issue there is should only be a small change to fix.

andrewcraik:
The suspicions around these changes causing stability issues seem to be resolved, but I'm leaving this WIP for now because Mem14 it depends on is WIP'd due to an issue with LocalTransparency.

compiler/optimizer/RegisterCandidate.hpp
```diff
@@ -63,25 +62,13 @@ class GlobalSet
...
-      public:
-      TR_ALLOC(TR_Memory::RegisterCandidates);
-      Set(TR::Region &region):_refs(region){}
-      bool get(uint32_t refId) { return _refs.get(refId); }
-      void set(uint32_t refId) { _refs.set(refId); }
-      void print(TR::Compilation * comp);
-      private:
-      TR_BitVector _refs;
-      };
-
```
vijaysun-omr: "very thin wrapper" for bit vector sounds right :) Not clear if there was some cleaner abstraction kind of reason why this was done but it is just a memory consumer the way it is at least.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -571,89 +571,52 @@ TR_RegisterCandidate::setWeight(TR::Block * * blocks, int32_t *blockStructureWei
    _loadsAndStores = new (comp->trStackMemory()) TR_Array<uint32_t>(comp->trMemory(), numberOfBlocks, true, stackAlloc);
```
vijaysun-omr: The commit message needs to be fixed to say "contains old code" instead of "contains old cold".

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -608,11 +608,12 @@ TR_RegisterCandidate::setWeight(TR::Block * * blocks, int32_t *blockStructureWei
...
-         if ((uint32_t)blockWeight > (*_loadsAndStores)[extendedBlock->getNumber()])
-            (*_loadsAndStores)[extendedBlock->getNumber()] = blockWeight;
+         if ((uint32_t)blockWeight > _blocks.getNumberOfLoadsAndStores(extendedBlock->getNumber()))//(*_loadsAndStores)[extendedBlock->getNumber()])
+            //(*_loadsAndStores)[extendedBlock->getNumber()] = blockWeight;
+            _blocks.setNumberOfLoadsAndStores(extendedBlock->getNumber(), blockWeight);
          }
-      if ((uint32_t)blockWeight > (*_loadsAndStores)[blockNumber])
-         (*_loadsAndStores)[blockNumber] = blockWeight;
+      //if ((uint32_t)blockWeight > (*_loadsAndStores)[blockNumber])
+      //   (*_loadsAndStores)[blockNumber] = blockWeight;
```
vijaysun-omr: Is this change in logic intentional ? The thrust of most of the other changes in this commit are to keep existing behaviour.

compiler/optimizer/RegisterCandidate.hpp
andrewcraik: Originally this abstraction served to wrap two different BitVector implementations using the CS2 sparse and dense bit vectors. These bit vectors were change to TR_BitVectors for memory savings and compatibility with region memory allocation. That rendered the hierarchy meaningless and a memory consumer hence why it is removed in my commits.

compiler/optimizer/RegisterCandidate.cpp
andrewcraik: The change is is more subtle that this if disappearing. the _loadAndStores structure was, to a large extent, duplicating the information in the _blocks data structure. This if was copying the data from _blocks to _loadAndStores. The difference here is that there is an accumulation of frequencies over part of the extended basic block for _loadAndStores that does not happen for _blocks before my change. The change to accumulate the info in _blocks seemed to be logical and unlikely to affect current usage while saving the duplicate data structure. Performance testing so far say there is no impact of throughput performance. The semantic change is due to the unification of _blocks and _loadAndStores.

PR #1469, opened 2017-08-03
Add copyright to README.md

rwy0717:
I was looking at what files were missing copyrights, and noticed that most markdown files are guilty. As a demo, this commit adds a copyright header as a comment. It won't show in the GitHub rendering. I've also added a shorter copyright footer, which is displayed.  To be honest, I'm not sure this is wanted, but here it is.

You can check out the render here: https://github.com/rwy0717/omr/tree/copyright

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,20 @@
+<!--
+(c) Copyright IBM Corp. 2016, 2017
+
+This program and the accompanying materials are made available
+under the terms of the Eclipse Public License v1.0 and
+Apache License v2.0 which accompanies this distribution.
+
+    The Eclipse Public License is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    The Apache License v2.0 is available at
+    http://www.opensource.org/licenses/apache2.0.php
+
+Contributors:
+  Multiple authors (IBM Corp.) - initial implementation and documentation
+-->
+
 [![Travis Status](https://api.travis-ci.org/eclipse/omr.svg?branch=master)](https://travis-ci.org/eclipse/omr)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,20 @@
+<!--
+(c) Copyright IBM Corp. 2017
```
mstoodle: should be 2016, 2017 since this file was created in 2016.

rwy0717: Fixed :)

rwy0717:
As this patch is, there are two places to maintain the "last updated year". I think that's too much effort. Are we OK to remove the copyright footer? Or at least the year?

It's worth figuring out here, because we're establishing a convention for all our markdown files, including documentation.

mstoodle:
If the copyright message (footer) is included, it must include the years. It does feel a bit repetitive, but each one also serves a purpose (the header one so there's a consistent header in all our files with all the relevant info, and the footer as part of the actual displayed text). I'm not sure I have a strong opinion.

Thoughts, @charliegracie ? I guess it would also make sense to see if Eclipse IP has an opinion.

charliegracie:
I think we should keep it and sadly keep it up to date

charliegracie:
I looked around and it seems like we should just keep it.  If it becomes too much of an issue in the future we can ask the Eclipse IP team about removing it.  @rwy0717 can you rebase (I merged your other PR first) and I will get this merged.

rwy0717:
I've rebased, and addressed @mstoodle's issue. This is good to go.

mstoodle:
Thanks, @rwy0717 and @charliegracie for the research!

PR #1472, opened 2017-08-05
Add Tril tests to Travis builds

Leonardo2718:
This change set does some clean up to Tril and its cmake files to allow better integrate with the other OMR cmake files. It also adds the Tril tests and Tril-based compiler tests to the Travis builds.

### DIFF
--- a/cmake/compiler_support.cmake
+++ b/cmake/compiler_support.cmake
@@ -95,6 +95,7 @@ function(make_compiler_target TARGET_NAME)
       TR_HOST_X86
       TR_TARGET_X86
       TR_TARGET_64BIT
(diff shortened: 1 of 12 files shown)


charliegracie:
@Leonardo2718 This is a great change and I really like the Tril test suite.  I am approving these changes but for a future change can you fix the compiler warnings in fvtest/tril/test/ASTTest.cpp around `const char *` and `char *`.  They were not introduced in this changeset but they are now visible because we are compiling Tril in the travis builds :)

Leonardo2718:
@charliegracie Will do :+1: :slightly_smiling_face:

Issue #1426, opened 2017-07-26
Segfault due to incorrect privatization of wrtbar


IBMJimmyk:
Field Privatizer recognizes it can privatize a wrtbar node that stores to a static. The wrtbar is changed so it now stores to an auto. However, a wrtbar is unnecessary when storing to an auto so this is an unexpected construct. Optimizations farther down the line assume it can't exist which can eventually leads to a segfault.

To address the problem, Field Privatizer should also convert the wrtbar into an aStore node during privatization.

PR #1473, opened 2017-08-08
Change privatized wrtbar into an aStore

IBMJimmyk:
Field privatizer can turn wrtbars that store to a static into a wrtbar
that stores to an auto. However, stores to autos do not need a wrtbar.
This change causes field privatizer to instead convert the wrtbar to an
aStore.

The second child of the original wrtbar is not needed by the new aStore so
it get anchored. This change set also fixes the issue with privatizing
indirect loads and stores where the children that were not being used were
not being anchored. Those children will now be anchored as well.
Particular care was taken to ensure that the nodes are anchored in the
same order they would have normally have been evaluatated in.

Issue: #1426

### DIFF
--- a/compiler/optimizer/FieldPrivatizer.cpp
+++ b/compiler/optimizer/FieldPrivatizer.cpp
@@ -709,7 +709,7 @@ void TR_FieldPrivatizer::privatizeFields(TR::Node *node, bool postDominatesEntry
             node->setSymbolReference(autoForField);
             TR::Node *newFirstChild = 0;
             int32_t newFirstChildNum = -1;
-            if ( opCode.isIndirect() )
+            if (opCode.isIndirect())
                {
                if (opCode.isStore())
                   {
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
Anchoring is a good practice in general but I wonder if is is needed here specifically since there are fairly stringent requirements on being able to do field privatization, one of them being that the subtree under the load/store in question should be invariant in the loop. Given this, none of those values would change inside a block in the loop either meaning we don't have to anchor for correctness in this optimization.

IBMJimmyk:
To confirm, that applies to both children of the wrtbar? If so, it does sound like anchoring might not be needed here.

I was also talking with @andrewcraik about anchoring the nodes so he might have more to add.

vijaysun-omr:
The second child of the wrtbar is the RHS value (being stored) and that will be a child of the store node anyway after privatization. The first child should have already been checked to be invariant I think.

IBMJimmyk:
I talked to @gita-omr . We took a look inside detectFieldsThatCannotBePrivatized in FieldPrivatizer.cpp and it looks like the check to see if a child is loop invariant only occurs if the parent node is indirect. Also the code only checks if the first child is loop invariant. Since a wrtbar is direct, the check to see if its children are loop invariant does not occur. Also an iwrtbar has three children and currently the third child is not checked to see if it is loop invariant even though it can get decremented (only the second child is kept after privitization).

From this it looks like the anchoring code is actually needed. @vijaysun-omr does this sound right to you?

vijaysun-omr:
I think for a wrtbar the second child is assumed to be loop invariant since it represents the runtime's internal class representation of the class of the static variable being written by the wrtbar. The assumption is that a language runtime only has a single internal class representation for a given class in a given run (this would make that second child loop invariant).

IBMJimmyk:
Since it looks like the child nodes that aren't kept by Privatizer must be loop invariant, I'm going to remove the node anchoring code.

As mentioned earlier, the second child of a wrtbar must be loop invariant. The third child of a wrtbari is either the same node as the first child or in one of the first child's subtrees. This third child should also always be loop invariant if the first child is loop invariant (which we check for).

vijaysun-omr:
Looks okay to me. Merging.

PR #1474, opened 2017-08-08
CMake: Set the tool configuration for osx, z/os.

rwy0717:
The z/os configuration is wrong, but at least we'll try something.

/cc @dnakamura @mgaudet @charliegracie 

### DIFF
--- a/cmake/modules/OmrDetectSystemInformation.cmake
+++ b/cmake/modules/OmrDetectSystemInformation.cmake
@@ -67,7 +67,7 @@ macro(omr_detect_system_information)
 
    if(OMR_HOST_OS STREQUAL "linux")
       # Linux specifics
-      set(OMR_TOOLCONFIG "gnu") 
+      set(OMR_TOOLCONFIG "gnu")
    endif()
 
    if(OMR_HOST_OS STREQUAL "win")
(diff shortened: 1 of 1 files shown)


rwy0717:
Rebased and fixed conflict.

cmake/modules/OmrDetectSystemInformation.cmake
```diff
@@ -83,11 +83,13 @@ macro(omr_detect_system_information)
 
    if(OMR_HOST_OS STREQUAL "zos")
       # ZOS specifics
+      set(OMR_TOOLCONFIG "gnu")
    endif()
 
 
    if(OMR_HOST_OS STREQUAL "osx")
       # OSX specifics
+      set(OMR_TOOLCONFIG "gnu")
```
charliegracie: Do we want to create a clang tool config?

rwy0717: I believe the long-run intention is that gnu will cover all gnu-ish toolchains, including clang. This is just the smallest change I could make to get osx/cmake building.

rwy0717:
This is needed to enable travis builds in #1475. Check out the failed build [here](https://travis-ci.org/rwy0717/omr/jobs/262567501). We fail to set the toolchain and `exit 1`.

cmake/modules/OmrDetectSystemInformation.cmake
```diff
@@ -83,11 +83,13 @@ macro(omr_detect_system_information)
 
    if(OMR_HOST_OS STREQUAL "zos")
       # ZOS specifics
+      set(OMR_TOOLCONFIG "gnu")
```
rwy0717: yeah I wrote in the commit, this setting is wrong, but at least there's something here while we work on proper xlc support.

charliegracie: I am ok with this for now so that we can move forward

PR #1475, opened 2017-08-08
CMake: Build with CMake on Travis OSX

rwy0717:
Delete the old osx build, because they are slow and keeping both will
bog down our CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,6 @@ env:
 matrix:
   exclude:
     - os: osx
-      env: BUILD_WITH_CMAKE=yes CMAKE_GENERATOR=Ninja
-    - os: osx
       env: BUILD_WITH_CMAKE=yes CMAKE_GENERATOR="Unix Makefiles"
   include:
     - os: linux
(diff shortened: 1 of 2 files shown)


mgaudet:
Probably can fix the OS/X issue with a `brew install ninja`

rwy0717:
I'm marking this as WIP until the osx build actually passes :)

dnakamura:
May also be useful to use `make` on the osx builds. That way we can catch possible issues with the makefile generator

mgaudet:
That's not a bad idea.

rwy0717:
In the interest of keeping osx builds fast, maybe we should add makefile builds for linux instead?

rwy0717:
cmake/osx is passing, this patch is no longer WIP.

dnakamura:
Wont this conflict with #1513 ?  Also I would suggest installing ninja if `CMAKE_GENERATOR==Ninja`

rwy0717:
Yes it will. TBH I didn't think this would be good to go so soon, I thought I'd be babysitting this PR for a while. Once #1513  goes in, I'll update to conditionally install ninja if enabled, and fix the conflict.

Then we can enable OSX builds at last 👍

rwy0717:
Fixed @dnakamura

dnakamura:
LGTM

PR #1477, opened 2017-08-08
CMake: Update OMR Platform code

rwy0717:


mgaudet:
Mostly LGTM; the only 'concern' I guess I have is the `-D` prefixing.

I guess this just means if you're a downstream project with a script that needs to know about the defines, but doesn't use -D, you'll do something like 

```
set(defs "${OMR_ARCH_DEFINITIONS} ...") 
string(REPLACE -D --defsym defs ${defs})
```

rwy0717:
I'm OK with undoing this change if there a specific use case. Was this for our masm2gas compiler or something?

My view is that a downstream project would need definitions from a variety of sources. It's cleaner to process flags in the "standard" format, than "some prefixed with -D, some not".

Plus this is just plain idiomatic cmake.

mgaudet:
I quoted 'concern' for a reason :) I actually agree, but just was thinking aloud, and wanted to raise the concern should it have impact on anyone else.

dnakamura:
Look like the windows build is complaining about unknown flags `/GM` and `/GD` Not sure where thats coming in from

mgaudet:
I think this is a good step in the right direction myself.

rwy0717:
The flag error @dnakamura mentioned was caused by a bug in `omr_remove_flags`. This function will delete the whitespace around any target flags, accidentally concatenating surrounding flags. This bug is hidden as long as the target flag is either first or last in the string.

This PR moves some flag additions earlier, pushing unwanted flags into the middle of the strings, which exposes the bug.

I've added a commit to fix `omr_remove_option`.

rwy0717:
OK. This should be good to go.

PR #1481, opened 2017-08-09
Move platform code to platform files

mgaudet:
Extract Windows and z/OS platform code to their respective platform files.


charliegracie:
@rwy0717 and @dnakamura any comments?

rwy0717:
I think the windows configuration code needs to be better split between os/win and toolcfg/msvc. A lot of the flag configuration is msvc only.

I like the introduction of the macro for importing os config. It's another step forward.

mgaudet:
Moved the windows setup code to the MSVC file. Altered the macro system a touch.

rwy0717:
LGTM

PR #1483, opened 2017-08-09
CMake: Build object libraries

dnakamura:
Build omrutil, j9thrstatic, and omrport as object libraries so that consumers can extend them if they wish.  The normal static libraries are still built from the object libs for internal use


rwy0717:
LGTM

mgaudet:
LGTM

PR #1485, opened 2017-08-10
Improve JitBuilder handling of shift operands

mstoodle:
After the new IL validator showed up, it exposed some weaknesses in
how operands of shift operations were being handled by JitBuilder.
In particular, shift operands are validated using the same mechanism
as for other binary operations, and that mechanism is entirely
inappropriate for shifts. In fact, it only really works for 32-bit
shift operands

This commit separates out the handling of shift operations from the
generic binaryOp* handling and makes the operand testing a bit more
relevant: the left operand must be integer typed, if the left operand
is narrower than 32-bits then it will be automatically widened to
32-bits, the right operand will be automatically converted to a
32-bit integer. If no conversion exists for either operand,
ConvertTo should assert.

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1076,8 +1076,8 @@ IlBuilder::doVectorConversions(TR::Node **leftPtr, TR::Node **rightPtr)
 
 TR::Node*
 IlBuilder::binaryOpNodeFromNodes(TR::ILOpCodes op,
-                             TR::Node *leftNode,
-                             TR::Node *rightNode) 
+                                 TR::Node *leftNode,
+                                 TR::Node *rightNode) 
    {
    TR::DataType leftType = leftNode->getDataType();
    TR::DataType rightType = rightNode->getDataType();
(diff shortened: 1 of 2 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1443,10 +1443,53 @@ IlBuilder::Xor(TR::IlValue *left, TR::IlValue *right)
...
+   } 
+
+TR::IlValue *
+IlBuilder::shiftOpFromOpMap(OpCodeMapper mapOp,
+                            TR::IlValue *left,
+                            TR::IlValue *right)
+   {
+   if (left->getDataType() == TR::Int8 || left->getDataType() == TR::Int16)
+      left = ConvertTo(Int32, left);
+   TR_ASSERT(left->getDataType() == TR::Int32 || left->getDataType() == TR::Int64, "left operand of shift must be Int32 or Int64");
```
Leonardo2718: What's the reason for this restriction? Unless I'm missing something, I would have thought that using one of the [`bshl`, `sshl`, `bshr`, `sshr`] opcodes would have worked here.

mstoodle: I had a weak impression (which I should have verified) that not all platforms supported the narrower types. But after verifying and discussing with @Leonardo2718, it looks like my impression was incorrect. I'll update the validation here. Thanks for keeping me honest, @Leonardo2718

mstoodle: @Leonardo2718 please take a look at the commit below to see if you agree with this version. Once we're both happy, I'll squash into a single commit.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1450,7 +1450,7 @@ IlBuilder::shiftOpNodeFromNodes(TR::ILOpCodes op,
    {
    TR::DataType leftType = leftNode->getDataType();
    TR::DataType rightType = rightNode->getDataType();
-   TR_ASSERT((leftType == TR::Int32 || leftType == TR::Int64) && rightType == TR::Int32, "shift operation first operand must be 32-bit or 64-bit integer, and shift amount must be 32-bit integer");
+   TR_ASSERT((leftType == TR::Int8 || leftType == TR::Int16 || leftType == TR::Int32 || leftType == TR::Int64) && rightType == TR::Int32, "shift operation first operand must an integer, and shift amount must be 32-bit integer");
```
Leonardo2718: Maybe use `leftType.isIntegral()`?

Also "must **be** an integer" in the assert message.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1470,19 +1470,16 @@ IlBuilder::shiftOpFromOpMap(OpCodeMapper mapOp,
...
-   if (left->getDataType() == TR::Int8 || left->getDataType() == TR::Int16)
-      left = ConvertTo(Int32, left);
-   TR_ASSERT(left->getDataType() == TR::Int32 || left->getDataType() == TR::Int64, "left operand of shift must be Int32 or Int64");
-
-   if (right->getDataType() != TR::Int32)
-      right = ConvertTo(Int32, right);
-
    TR::Node *leftNode = loadValue(left);
+   TR::DataType leftType = leftNode->getDataType();
+   TR_ASSERT(leftType == TR::Int8 || leftType == TR::Int16 || leftType == TR::Int32 || leftType == TR::Int64, "left operand of shift must be integer type");
```
Leonardo2718: Same comment as above.

compiler/ilgen/IlBuilder.cpp
```diff
@@ -1443,10 +1443,50 @@ IlBuilder::Xor(TR::IlValue *left, TR::IlValue *right)
...
+IlBuilder::shiftOpFromOpMap(OpCodeMapper mapOp,
+                            TR::IlValue *left,
+                            TR::IlValue *right)
+   {
+   TR::Node *leftNode = loadValue(left);
+   TR::DataType leftType = leftNode->getDataType();
+   TR_ASSERT(leftType.isIntegral(), "left operand of shift must be integer type");
+
+   TR::Node *rightNode = loadValue(right);
+   if (!rightNode->getDataType().isInt32())
```
charliegracie: You can only shift a 64bit value for a 32bit shift?  I would assume you could shift a 64 bit value by a 64 bit value.  I may be missing something though so just asking.

Leonardo2718: As it turns out, the shift value should always be 32-bits according the rules of Testarossa IL. This is what the ILValidator checks and what is generally expected by the optimizer.

mstoodle: @charliegracie where do you want to shift that value, Charlie? :)  That would be 9,223,372,036,854,775,807 bit positions!

Note that it will convert to 32-bits for you, so will 2B shifts be enough :) ?

mstoodle:
ah, nuts, I should fix the commit message :(

mstoodle:
I squashed down to a single commit that I believe addresses all the review comments so far, unless I've misunderstood Charlie's concern.

compiler/ilgen/IlBuilder.cpp
charliegracie: Thanks for correcting me.  I guess I was tired and convinced myself shifting by a 32 bit number limited you to shifting by 32 bits :).  You definitely do not need to shift by a 64 bit value.

mstoodle: No worries!

PR #1486, opened 2017-08-10
CMake: Add OMR_FVTEST

rwy0717:
This flag is already enabled in CI, but was never used properly.

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -22,13 +22,14 @@ set(OMR_WARNINGS_AS_ERRORS ON CACHE BOOL "Treat compile warnings as errors")
 ### Major Feature Flags
 ###
 
-set(OMR_GC ON CACHE BOOL "Enable the GC")
 set(OMR_COMPILER OFF CACHE BOOL "Enable the compiler")
+set(OMR_FVTEST OFF CACHE BOOL "Enable the FV Testing.")
+set(OMR_GC ON CACHE BOOL "Enable the GC")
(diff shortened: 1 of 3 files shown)


cmake/config.cmake
```diff
@@ -22,13 +22,14 @@ set(OMR_WARNINGS_AS_ERRORS ON CACHE BOOL "Treat compile warnings as errors")
...
+set(OMR_GC ON CACHE BOOL "Enable the GC")
 set(OMR_JITBUILDER OFF CACHE BOOL "Enable building JitBuilder")
-set(OMR_TEST_COMPILER OFF CACHE BOOL "Enable building the test compiler")
+set(OMR_OMRSIG ON CACHE BOOL "Enable the OMR signal compatibility library")
 set(OMR_PORT ON CACHE BOOL "Enable portability library")
+set(OMR_TEST_COMPILER OFF CACHE BOOL "Enable building the test compiler")
 set(OMR_THREAD ON CACHE BOOL "Enable thread library")
-set(OMR_OMRSIG ON CACHE BOOL "Enable the OMR signal compatibility library")
 # set(OMR_TOOLS On CACHE BOOL "Enable the build tools")
 # TODO: Support building only tools for cross-compilation build
```
rwy0717: changes: Added flag OMR_FVTEST (default OFF). Alphabetized major features.

Leonardo2718:
This is likely outside the scope of this PR, but in future, would it be possible to also add flags for enabling/disabling tests for specific components (like the test compiler already has)?

PR #1487, opened 2017-08-10
CMake: Do not use OMR_GLUE as a subdirectory

rwy0717:
We can't realistically use the OMR_GLUE as a subdirectory. This will
break when the glue is outside OMR. Instead, we have to rely on a parent
project defining all the right glue targets and variables, before
building omr.

Instead of testing for the existence of the OMR_GLUE variable,
The OMR_EXAMPLE flag can be used to conditionally add the example
directory, which will define the glue library.

OMR_EXAMPLE is already enabled in the CI.

There is an issue of interdependence here that we have yet to sort out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,15 +33,6 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules" ${CMAKE_MODULE_PATH})
 include(OmrAssert)
 include(OmrFindFiles)
 
-###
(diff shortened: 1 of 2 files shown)


rwy0717:
Building the example had to be moved lower in the cmakelists, after we set up our global configuration.

dnakamura:
fixed in #1338

rwy0717:
Closing, as devin pointed out #1338 solves this issue.

PR #1488, opened 2017-08-10
[CMake] Refactor JIT platform detection

mgaudet:
This patch turns the JIT platform detection into a function, and starts
to separate some pieces into variables. A side effect of this patch is
that it adds support for building compiler components with OS/X.

### DIFF
--- a/cmake/compiler_support.cmake
+++ b/cmake/compiler_support.cmake
@@ -31,22 +31,65 @@ if (NOT PERL_FOUND )
    message(FATAL_ERROR "Perl not found")
 endif()
 
+
 set(MASM2GAS_PATH ${OMR_ROOT}/tools/compiler/scripts/masm2gas.pl CACHE INTERNAL "MASM2GAS PATH")
 
 include(${OMR_ROOT}/cmake/AddPrefix.cmake) 
 
-# Platform setup code!
-# TODO: THis needs to be abstracted for cross platform builds
+# Fetch the OMR view of the system.
+include(OmrDetectSystemInformation)
 
-set(TR_TARGET_ARCH    x     CACHE INTERNAL "The architecture directory used for the compiler code (x, p, arm, or z)")
-set(TR_TARGET_SUBARCH amd64 CACHE INTERNAL "The subarchitecture directory used for the compiler code. May be empty (i386 or amd64)")
-set(TR_TARGET_BITS    64    CACHE INTERNAL  "Bitness of the target architecture")
+macro(tr_detect_system_information)
 
-set(TR_HOST_ARCH    x     CACHE INTERNAL "The architecture directory used for the compiler code (x, p, or z)")
-set(TR_HOST_SUBARCH amd64 CACHE INTERNAL "The subarchitecture directory used for the compiler code. May be empty (i386 or amd64)")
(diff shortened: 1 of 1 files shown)


cmake/compiler_support.cmake
```diff
@@ -31,22 +31,65 @@ if (NOT PERL_FOUND )
...
+	#  - Support more platforms, and, separate host and target arch. 
+	#  - Once we support all the platforms OMR supports with CMake, this can be
+	#    largely integrated into OmrDetectSystemInformation.cmake. 
+	if(OMR_ARCH_X86)
+		set(TR_HOST_ARCH    x     CACHE INTERNAL "The architecture directory used for the compiler code (x, p, or z)")
+		list(APPEND TR_COMPILE_DEFINITIONS TR_HOST_X86 TR_TARGET_X86)
+		if(OMR_ENV_DATA64)
+			set(TR_HOST_SUBARCH amd64 CACHE INTERNAL "The subarchitecture directory used for the compiler code. May be empty (i386 or amd64)")
+			set(TR_HOST_BITS    64    CACHE INTERNAL  "Bitness of the target architecture")
+			set(CMAKE_ASM-ATT_FLAGS "--64 --defsym TR_HOST_X86=1 --defsym TR_HOST_64BIT=1 --defsym BITVECTOR_64BIT=1 --defsym LINUX=1 --defsym TR_TARGET_X86=1 --defsym TR_TARGET_64BIT=1" CACHE INTERNAL "ASM FLags")
```
charliegracie: This is something we need to look into further as we decide how we are going to handle ASM files.  These options will break other components using the assembler.  I do not see the assembler being enabled in CMake as a language yet for the compiler so this set line is not being used as far as I can tell. This is not new code but I am just pointing out as a future consideration before I forget :)

cmake/compiler_support.cmake
```diff
@@ -31,22 +31,65 @@ if (NOT PERL_FOUND )
...
+	set(TR_COMPILE_DEFINITIONS "")
+
+
+	# Platform setup code! 
+	# TODOs:
+	#  - Support more platforms, and, separate host and target arch. 
+	#  - Once we support all the platforms OMR supports with CMake, this can be
+	#    largely integrated into OmrDetectSystemInformation.cmake. 
+	if(OMR_ARCH_X86)
+		set(TR_HOST_ARCH    x     CACHE INTERNAL "The architecture directory used for the compiler code (x, p, or z)")
```
charliegracie: I am not sure about all of the extra spaces to attempt to line things up.  Is this something the other CMake files are doing?  This type alignment usually ends up being broken and also causes a 1 line change to turn into much larger changes to fix alignment issues.

mgaudet: So, I would personally defend alignment: You read the code more than you read diffs, and if the alignment make it easier to read the code (as I think it does here), I'd put that higher in importance than minimizing diffs.

Having said that, this patch is known to not be final form. I'm trying to get something in here that will allow #1475 to go in, but the CMake platform story us under active evolution.

cmake/compiler_support.cmake
mgaudet: Yep. It's definitely something that needs changing, but the whole CMake platform support code is a WIP

mgaudet:
@charliegracie Any objections to this going in?

PR #1490, opened 2017-08-10
[Comp17] Reduce runtime complexity of copy propagation

andrewcraik:
Performance analysis of the compile time of the OMR compiler has shown that copy propagation is consuming a very large amount of time when compiling methods with 1000s of symbol references and 1000s of blocks. The inner loop of the optimization was found to be iterating over all def indices and all use indicies for each block. This set of commits improves the performance of this inner loop to provide significant compile time savings.

This commit depends on Mem14-Mem16.

### DIFF
--- a/compiler/optimizer/CopyPropagation.cpp
+++ b/compiler/optimizer/CopyPropagation.cpp
@@ -457,8 +457,6 @@ int32_t TR_CopyPropagation::perform()
       currentTree = currentTree->getNextTreeTop();
       }
 
-   int32_t i = 0, j;
-
    int32_t numDefsOnEntry    = useDefInfo->getNumDefsOnEntry();
    int32_t firstRealDefIndex = useDefInfo->getFirstDefIndex() + numDefsOnEntry;
    int32_t lastDefIndex      = useDefInfo->getLastDefIndex();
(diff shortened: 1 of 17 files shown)


compiler/optimizer/CopyPropagation.cpp
```diff
@@ -540,18 +544,32 @@ int32_t TR_CopyPropagation::perform()
...
                            }
-                        else if (equivalentDefs[j-firstRealDefIndex] == j)
+                        else if (equivalentDefs.find(j) == equivalentDefs.end())
                            {
                            if (trace())
                                traceMsg(comp(), "111setting j %d to i %d\n", j, i);
-                           equivalentDefs[j-firstRealDefIndex] = i;
+                           if (i != j)
+                              equivalentDefs[j] = i;
                            }
```
vijaysun-omr: Is there a reason there is'nt logic similar to what I saw in an earlier instance of this kind of condition in this commit : 

else if (i == equivalentDefLookup->second)
                              {
                              equivalentDefs.erase(i);
                              }

that I assume is essentially saying that if we are just marking a def as it's own equivalent def, then we do not need an entry for it in the map data structure that we have now.

vijaysun-omr:
The PR title needs to say "of" instead of "if"

compiler/optimizer/CopyPropagation.cpp
```diff
@@ -745,16 +745,18 @@ int32_t TR_CopyPropagation::perform()
    for (auto itr = usesToBeFixed.begin(), end = usesToBeFixed.end(); itr != end; )
       {
       int32_t fixUseIndex = itr->first;
+      TR::Node *fixUseNode = useDefInfo->getNode(fixUseIndex);
       invalidateDefUseInfo = true;
+      if (useDefInfo->hasLoadsAsDefs())
+         {
+         printf("here\n");
```
vijaysun-omr: This printf needs to be removed.

compiler/optimizer/CopyPropagation.cpp
```diff
@@ -745,16 +745,18 @@ int32_t TR_CopyPropagation::perform()
    for (auto itr = usesToBeFixed.begin(), end = usesToBeFixed.end(); itr != end; )
       {
       int32_t fixUseIndex = itr->first;
+      TR::Node *fixUseNode = useDefInfo->getNode(fixUseIndex);
       invalidateDefUseInfo = true;
+      if (useDefInfo->hasLoadsAsDefs())
```
vijaysun-omr: Is this condition really needed ? I ask, because it is not used before the later loop in this same commit. I'd have expected us to be consistent. Or is there some subtle reason why the first spot is different ?

ncough: This condition short cuts the potentially expensive loop. The loop is only necessary if loads are treated as defs and can therefore be avoided if not operating in that mode.

compiler/optimizer/CopyPropagation.cpp
ncough: This check should be redundant in both cases, as `i` should not appear in any of the mappings so far. The first section does not make this assumption, adding this condition to manage it, whilst the second does.

ncough:
These changes have been moved to a new PR: https://github.com/eclipse/omr/pull/1510
Please close this one.

vijaysun-omr:
Nick and I have both reviewed this. Merging.

PR #1491, opened 2017-08-10
[Comp18] Improve performance of global dead store elimination

andrewcraik:
Global dead store elimination has an inner loop which scales based on the highest symbol reference number in a method. This is very wasteful since only a small fraction of symrefs participate in global dead store elimination. This commit rewrites this inner loop to scale better as the number of symbol references grows.

This commit depend on Mem14 - Comp17

### DIFF
--- a/compiler/optimizer/IsolatedStoreElimination.cpp
+++ b/compiler/optimizer/IsolatedStoreElimination.cpp
@@ -1010,6 +1010,7 @@ TR_IsolatedStoreElimination::findStructuresAndNodesUsedIn(TR_UseDefInfo *info, T
       TR_RegionStructure *regionStructure = structure->asRegion();
 
       if (regionStructure->isNaturalLoop() &&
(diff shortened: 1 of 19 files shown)


compiler/optimizer/IsolatedStoreElimination.cpp
```diff
@@ -1181,6 +1183,10 @@ TR_IsolatedStoreElimination::findStructuresAndNodesUsedIn(TR_UseDefInfo *info, T
                }
             }
          }
+      // if we can remove the strucutre then these defs are ok here and in all parent structures
```
ncough: Structure is spelt incorrectly here.

ncough:
LGTM - Besides the spelling mistake.

vijaysun-omr:
Merging this since the prereq stacked changes are in now and both Nick and I have reviewed.

PR #1493, opened 2017-08-10
Fix _patchVirtualGuard to not assert on false positives

fjeremic:
When patching happens synchronously we are effectively patching in a
jump with a guarantee no processor will be executing the patch location.
Thus there is no race condition. This means we will be overwriting
existing instructions. If it so happens that the instruction at the
patch location has opcode 0xA7 or 0xC0 as the first byte we may
encounter a false positive and incorrectly assert. Such an example can
be seen if an LHI R8,0 (0xA7880000) instruction is after the VGNOP.

This fix strengthens the condition of when one byte patching happens,
i.e. if and only if we are certain a BRC or BRCL is present.

### DIFF
--- a/compiler/z/runtime/VirtualGuardRuntime.cpp
+++ b/compiler/z/runtime/VirtualGuardRuntime.cpp
@@ -56,10 +56,8 @@ extern "C" void _patchVirtualGuard(uint8_t* locationAddr, uint8_t* destinationAd
 
    int64_t displacement = static_cast<int64_t>(destinationAddr - locationAddr) / 2;
 
-   if (locationAddr[0] == 0xA7 || locationAddr[0] == 0xC0)
+   if ((locationAddr[0] == 0xA7 || locationAddr[0] == 0xC0) && locationAddr[1] == 0x04)
       {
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
You mentioned the false positive condition when we may assert. 
Could you please elaborate on when we actually want to assert ?

fjeremic:
> Could you please elaborate on when we actually want to assert ?

We want to assert if a NOP branch has been placed. A NOP branch on IBM Z is defined as a branch which has the correct target but does not branch on any condition code, i.e. the mask value of the branch is zero.

The code generator needs to ensure that all branches which can be patched asynchronously (with the application potentially executing the same branch we are patching) are NOP branches with the correct target location. Thus `_patchVirtualGuard` should assert if there is a NOP branch placed (indicating patching is being done while application code is executing) and the destination of the branch is the same in the binary encoding as it is in the argument that is passed to `_patchVirtualGuard`.

If the branch destination is not correct we cannot safely patch the branch to a different destination because that involves storing either 2 or 4 bytes to memory which may not be block concurrent on IBM Z processors if the memory location happens to cross a 4 byte boundary. IBM Z processors only ensure block concurrent stores of 4 bytes on a 4-byte alignment.

vijaysun-omr:
Thanks for the clear explanation. I will merge now.

PR #1500, opened 2017-08-11
Tweak phrasing of CONTRIBUTING.md

mgaudet:
Fixes some grammatical errors, and make an example commit a little
clearer (so that it actually complies with the commit guidelines)

[ci skip] documentation change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 Thanks for your interest in this project.
 
 We welcome and enourage all kinds of contribution to the project, not only code. 
-This can incuding bug reports, user experience feedback, assistance in reproducing
+This includes bug reports, user experience feedback, assistance in reproducing
 issues and more.
 
 This project uses GitHub Issues to track ongoing development, discuss project
@@ -46,9 +46,10 @@ Following these guidelines will help us to merge your pull requests smoothly:
 ## Commit Guidelines
 
 The first line describes the change made. It is written in the imperative mood,
-say what happens when the patch is applied. Keep it short and simple. The first
-line should be less than 50 characters, sentence case, and does not end in a
-period. Leave a blank line between the first line and the message body.
+and should say what happens when the patch is applied. Keep it short and
+simple. The first line should be less than 50 characters, sentence case, and
+does not end in a period. Leave a blank line between the first line and the
+message body.
 
 The body should be wrapped at 72 characters, where reasonable.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 Thanks for your interest in this project.
 
 We welcome and enourage all kinds of contribution to the project, not only code. 
-This can incuding bug reports, user experience feedback, assistance in reproducing
+This incudes bug reports, user experience feedback, assistance in reproducing
```
rwy0717: `s/incudes/includes/` :)

CONTRIBUTING.md
```diff
@@ -46,9 +46,10 @@ Following these guidelines will help us to merge your pull requests smoothly:
 ## Commit Guidelines
 
 The first line describes the change made. It is written in the imperative mood,
-say what happens when the patch is applied. Keep it short and simple. The first
-line should be less than 50 characters, sentence case, and does not end in a
-period. Leave a blank line between the first line and the message body.
+and should say what happens when the patch is applied. Keep it short and
+simple. The first line should be less than 50 characters, sentence case, and
```
charliegracie: I feel like we should probably change this to say like this:
```The first line should be less than 50 characters, whenever possible and written in sentence case.` ``
I do not like enforcing the 50 character limit as a hard limit as then we end with descriptions that do not quite make sense or have people remove vowels from words to make it fit. I think the 50 character limit is a good guideline but I do not feel like we should be so pedantic about it.  Also if the first line ends in a period what harm does that cause?  Enforcing some of these rules just puts extra pressure on the CI environments and may scare people away from the project as we have sooo many rules without adding any real value.

@mstoodle what are your thoughts?

File: CONTRIBUTING.md (hunk 1 above)
Comment (mgaudet) [inline CONTRIBUTING.md]: 👍 Thanks, fixed in 235e7289

File: CONTRIBUTING.md (hunk 2 above)
Comment (mgaudet) [inline CONTRIBUTING.md]: I'd love to have this discussion, but I think that's an issue, not something for this PR.

Comment (mstoodle) [inline CONTRIBUTING.md]: I created #1501 for the ongoing discussion.

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -46,9 +46,10 @@ Following these guidelines will help us to merge your pull requests smoothly:
 ## Commit Guidelines
 
 The first line describes the change made. It is written in the imperative mood,
```
alin00: Missed opportunity to fix "imperative mood" -> "imperative mode" here.

PR #1502, opened 2017-08-11
Add portlib api to return max heap size

ashu-mehra:
A new port library API omrsysinfo_get_physical_memory_for_heap()
is added that returns the default max heap size taking into account
following factors:
1. physical memory in the system
2. cgroup limit on memory (conditional)
3. address space limit on memory

Another API omrsysinfo_cgroup_use_memlimit_for_heap() is added
that sets a flag which will cause omrsysinfo_get_physical_memory_for_heap()
to use cgroup limit on memory for determining the max heap size.

### DIFF
--- a/gc/base/GCExtensionsBase.cpp
+++ b/gc/base/GCExtensionsBase.cpp
@@ -58,9 +58,6 @@ bool
 MM_GCExtensionsBase::initialize(MM_EnvironmentBase* env)
 {
 	OMRPORT_ACCESS_FROM_OMRPORT(env->getPortLibrary());
-	uint64_t physicalMemory = 0;
-	uint64_t memoryLimit = 0;
-	uint64_t usableMemory = 0;
 	uint64_t memoryToRequest = 0;
 	uintptr_t *pageSizes = NULL;
 	uintptr_t *pageFlags = NULL;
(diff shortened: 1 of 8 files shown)


ashu-mehra:
@charliegracie @dmitripivkine fyi

DanHeidinga:
@ashu-mehra The "heap" in this context is the managed runtimes garbage collected heap, correct?  Assuming that's correct then I don't think this is really a Port Library api.  Something akin to this should be implemented in the garbage collector.

The port library should be the thin interface between the rest of the runtime and the operating system.  This feels (to me at least) to be on the wrong side of that design point.

charliegracie:
I think there are a few different ideas being covered under the one idea here.  Some of these decisions are the GC's responsibility and some are the port libraries responsibility in my opinion.  The GC deciding to use 1/2 or 1/3 of the physical memory available is a GC decision.  The GC should ask the port library 1 question.  How much memory is available on the system.  If there are cgroups or other OS / virtualization concerns those should be covered in the one call to the port library.  This port library API may require a parameter so that the callers can specify which of the OS/virtualization possibilities that it wants to respect. At that point the GC gets back information from the port library and then makes it modifications to the value of physical available memory that are GC specific.

DanHeidinga:
> This port library API may require a parameter so that the callers can specify which of the OS/virtualization possibilities that it wants to respect

That seems like a reasonable approach.

Given the design of the port library which makes having these kinds of common / platform independent helpers inconvenient from a code maintenance point of view (ie: need to duplicate across all platforms), I'd still suggest putting this logic into the gc.

ashu-mehra:
> The GC should ask the port library 1 question. How much memory is available on the system. If there are cgroups or other OS / virtualization concerns those should be covered in the one call to the port library.

We already have an API omrsysinfo_get_physical_memory(). To convey this API to use cgroup memory limit, there is another API omrsysinfo_cgroup_enable_subsystem() which would cause omrsysinfo_get_physical_memory() to respect cgroup memory limit.

> This port library API may require a parameter so that the callers can specify which of the OS/virtualization possibilities that it wants to respect

I can add a new API, say omrsysinfo_get_memory_limit(LIMITTYPE type) where LIMITTYPE could be PHYSICAL, CGROUP etc. Question is when would GC use omrsysinfo_get_memory_limit(CGROUP)?
Requirement is for GC to call omrsysinfo_get_memory_limit(CGROUP) if a specific option/flag is set by the runtime, else use omrsysinfo_get_memory_limit(PHYSICAL).
If the GC code is actually checking for this option/flag, then it can as well call omrsysinfo_cgroup_get_memlimit() instead of calling omrsysinfo_get_memory_limit(CGROUP).

ashu-mehra:
@charliegracie @DanHeidinga Can you please respond to my previous comment.

DanHeidinga:
@ashu-mehra I see your point and made similar suggestions as well due to a desire to keep the port library api small / easy to port.  

I can also see @charliegracie's point though - he doesn't want the GC littered with multiple if this special case, call that api, else some other special case, call api 2, etc.  

Do you have a proposal for an API that can balance between the two concerns?  Are there are other cases in the GC that could be rolled into such a "single query" api?  It might be worthwhile to write the code for that api in the GC and let it do the separate portlib calls and see how the code looks.  Looking at a prototype of it may make it obvious if it should live in the port library or the gc.

DanHeidinga:
First, my apologies on the delayed responses here - my time has been taken by internal deliveries.

The API you proposed localizes the 3 calls into a single method which is a good start.  Is there a way to reduce those 3 portlib calls to a single call?

Can the GC hold onto the flags to be passed to the 'get_limit' api?  i.e. have the GC store the flag in one of it's data structures?  When parsing command line options, it can update the flag so the "respect cgroups for heap" option can set the flag to `CGROUP`, etc.

```
uint64_t 
MM_GCExtensionsBase::getMaxMemoryForHeap()
{
      availableMemory = omrsysinfo_get_memory_limit(_gcMemoryLimitType);
      memoryLimit = omrsysinfo_get_limit(OMRPORT_RESOURCE_ADDRESS_SPACE);
      return OMR_MIN(availableMemory, memoryLimit);
}
```

Based on @charliegracie's comments, it sounds like he would prefer to have `OMRPORT_RESOURCE_ADDRESS_SPACE` rolled into the flags passed to `omrsysinfo_get_memory_limit`.

ashu-mehra:
> Can the GC hold onto the flags to be passed to the 'get_limit' api? i.e. have the GC store the flag in one of it's data structures

Makes sense. I can replace `bool _useCgroupMemLimit` with `int _heapMemoryLimitType`.

> it sounds like he would prefer to have OMRPORT_RESOURCE_ADDRESS_SPACE rolled into the flags passed to omrsysinfo_get_memory_limit

Did you mean the port library API `omrsysinfo_get_memory_limit` should apply the address space limit internally before returning the value? I am wondering if the name `omrsysinfo_get_usable_memory_with_limit(int limitType)` would be more appropriate in that case?

ashu-mehra:
> Did you mean the port library API omrsysinfo_get_memory_limit should apply the address space limit internally before returning the value? I am wondering if the name omrsysinfo_get_usable_memory_with_limit(int limitType) would be more appropriate in that case?

@DanHeidinga your thoughts on this?

DanHeidinga:
> Did you mean the port library API omrsysinfo_get_memory_limit should apply the address space limit internally before returning the value?

That sounds like @charliegracie's preferred approach for this.  

Ideally, I'd like to see what other similar APIs for making it easier to deal with system limits we could create.  This would make it easier to end up with a good, consistent set of APIs for dealing with system limits.  I know its more work to do that investigation so I'm somewhat hesitant to ask you to do it, though the API at the end would be stronger for the work.

mstoodle:
Hi @ashu-mehra , this PR has been open and not very active for quite a while now. Would you please outline the remaining plan here, or should this PR be closed and re-submitted when it's ready?

ashu-mehra:
Better to close it. If required I will open a new one.

PR #1507, opened 2017-08-14
CMake: Add some new utilities

rwy0717:
This adds a new cmake module, OmrUtility, with some convenience functions for manipulating strings as flags.

The additions are:
`omr_remove_flags`: Renamed version of `omr_remove_option`. Accepts multiple flags.
`omr_stringify`: Stringify a list to a space separated string of items.
`ome_append_flags`: Append a list of flags to a string, intended to be used with `CMAKE_<LANG>_FLAGS` variable.


charliegracie:
Has @dnakamura or @mgaudet reviewed this yet?  Looking for a thumbs up

rwy0717:
Whoops, should have CC'd. That explains why nothing's going on.

dnakamura:
I feel like there was a reason I made `omr_remove_option` a macro, but im not sure why. But I guess it wasn't really a valid reason if it still works.

dnakamura:
LGTM

rwy0717:
I'm guessing it was done as a macro instead of using `PARENT_SCOPE`.

dnakamura:
actually it was probably So I could avoid the `set()` all together and just rely on `string(REGEX REPLACE` to do all the work

PR #1509, opened 2017-08-15
Correct guidance regarding issue references

keithc-ca:
According to https://help.github.com/articles/closing-issues-using-keywords/, references to issues should _not_ include a colon. That is,
```
Issue: #1234
```
doesn't work; it needs to be
```
Issue #1234
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,7 @@ Correct race in frobnicator
 
 This patch eliminates the race condition in issue #1234.
 
-Issue: #1234
+Issue #1234
 ```
 
 Sign off on your commit in the footer. By doing this, you assert original
@@ -113,7 +113,7 @@ The guidelines are changed to:
 - Encourage contributors to put more information into the commit
   message.
 
-Issue: #124
+Issue #124
 Signed-off-by: Robert Young <rwy0717@gmail.com>
 ```
 


mstoodle:
Ok, I guess you're right that none of the examples has a colon in it.

Of course, none of the examples has "issue" either. Perhaps we should also update the recommendation to use "issue" to "fixes" ?

@charliegracie any opinion here?

charliegracie:
I am ok with removing the `:`.  In the future I would also like to suggest that people use one of the key words that actually closes the Issue if they believe this is the final fix for the issue.

mstoodle:
Sounds good to me. @keithc-ca would you like to take a crack at adding some words to describe the use of, say, "fixes" for the final commit? Or would you like to leave that to someone else's PR (in which case I'll just merge this one) ?

mstoodle:
(incidentally, you can add [ci-skip] to your commit message for updates to the doc files to avoid running all the tests on travis/appveyor)

rwy0717:
The `Issue: #1234` syntax is not meant to close issues. It's just a conventional/grepable place to establish references. @mstoodle is right that it's the `#1234` that generates the link, and this can technically appear anywhere in the message.

Our guidelines say you should explain the relationship between a commit and any relevant issues in the body of the commit. This is where you would use the `Closes #1234` syntax. Since commit footers are conventionally organized as `Key: value`, I am against removing the colon.

Example:
```
Fix nullptr dereference in frobnicator

Closes #1234, where the frobnicator was segfaulting.

Issue: #1234
Signed-off-By: Robert Young <rwy0717@gmail.com>
```

mgaudet:
If you say `Closes #1234`, are we really going to expect people to also use the footer key pair? 

Not a big fan of this redundancy.

rwy0717:
That *is* what's in our guidelines, the issue numbers should each appear twice.

How about:
For each relevant issue, we require only a footer-tag, and we start to accept `Closes: ` and `Fixes: ` as valid footer-tags? We can make the in-body explanation optional.

Also check this out: github works with the colon.
https://github.com/rwy0717/github-sandbox/pull/2

keithc-ca:
It appears the presence of a colon doesn't interfere with automatically closing an issue referenced via 'closes', etc. I retract this pull request.

PR #1512, opened 2017-08-15
Disable PartialRedundancyElimination when profiling

ncough:
When profiling with HCR enabled, the extra blocks
and temps required in this configuration can result
in PRE taking up a significant amount of compile
time and memory. This can often result in the
compilation running out of scratch space during
or soon after PRE, degrading performance.

Disabling this opt in these conditions has been
verified on several benchmarks, with no evident
regression, as the following recompile will still
include the opt.

### DIFF
--- a/compiler/optimizer/PartialRedundancy.cpp
+++ b/compiler/optimizer/PartialRedundancy.cpp
@@ -213,6 +213,20 @@ bool TR_PartialRedundancy::ignoreNode (TR::Node *node)
 
 int32_t TR_PartialRedundancy::perform()
    {
+   // Calculate the number of expressions commoned by PRE that are allowed
+   // to be profiled. This is to keep the value profiling related code expansion
+   // within some limits
+   //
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
Loop specialization in the scorching compilation depends on loop invariant values being profiled in the profiling compilation. Profiling of loop invariant values is done in PRE (partial redundancy elimination) currently and this commit will effectively disable that optimization. So I have some doubts about whether this sweeping change (disabling PRE without looking at the size of the method for example) is a good one in general. At the very least I would prefer to have some kind of heuristic based on the size of the method to decide whether to disable it or not in a profiling compilation (assuming we cannot "fix" the problem of large memory consumption in some other way given your project deadline).

ncough:
The condition to disable has been updated to consider an existing heuristic which controls the number of values PRE can profile. It is based on the number of nodes, limiting this disable so that it only applies for larger bodies.

vijaysun-omr:
Thanks for changing the heuristic to be more selective. This looks better now. 
I will merge once checks pass.

PR #1515, opened 2017-08-16
Explicitly list include directories

### LINKED ISSUE #1514: Disable CMAKE_INCLUDE_CURRENT_DIR
Our `CMakeLists.txt` currently [sets `CMAKE_INCLUDE_CURRENT_DIR` to on.](https://github.com/eclipse/omr/blob/master/CMakeLists.txt#L61)

This will have to be disabled. The compiler technology requires control over the ordering of include paths. When CMAKE_INCLUDE_CURRENT_DIR is enabled, should one of the include paths automagically added by CMake collide with one of the include paths provided by the JIT, it will end up floating out of order, breaking the ability to build the JIT in some scenarios.

mgaudet:
This is required by the compiler technology because of the include path magic used to implement static polymorphism

Fixes #1514

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,9 +58,6 @@ include(cmake/config.cmake)
 
 enable_testing()
 
-#include current source dir and current bin dir automatically
-set(CMAKE_INCLUDE_CURRENT_DIR ON)
-
 include_directories(
   ${PROJECT_BINARY_DIR}
   ./include/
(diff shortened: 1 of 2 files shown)


mgaudet:
Closing this one for now. 

My local testing bit me with a dirty source dir.

mgaudet:
Whoops. Just ran into [this! ](https://github.com/isaacs/github/issues/361)

PR #1517, opened 2017-08-16
Reduce OSR shared slot map memory usage

ncough:
The OSR shared slot map can be compressed by representing
a continuous set of entries that share the same mapping
with the first. This was implemented by walking the
map and removing any duplicates that were identified,
however, TR_Array implements remove by copying all remaining
elements forward by one index.

The new implementation builds up a compressed list as it goes
along, rather than removing in place. This will not always be
cheaper, but will reduce overall usage due to the high frequency
of entries that can be removed.

### DIFF
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -221,25 +221,36 @@ TR_OSRCompilationData::addInstruction2SharedSlotMapEntry(
 void
 TR_OSRCompilationData::compressInstruction2SharedSlotMap()
    {
-   for (int i = 0; i < instruction2SharedSlotMap.size()-1;)
+   // This approach isn't ideal, as it will result in copies when moving to the new list
+   // and back, however, it is cheaper than the prior implementation.
+   //
+   // TODO: Replace TR_Array with a TR::deque in TR_Instruction2SharedSlotMap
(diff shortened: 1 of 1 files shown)


compiler/compile/OSRData.cpp
```diff
@@ -221,25 +221,36 @@ TR_OSRCompilationData::addInstruction2SharedSlotMapEntry(
 void
 TR_OSRCompilationData::compressInstruction2SharedSlotMap()
    {
-   for (int i = 0; i < instruction2SharedSlotMap.size()-1;)
+   // This approach isn't ideal, as it will result in copies when moving to the new list
+   // and back, however, it is cheaper than the prior implementation.
+   //
+   // TODO: Replace TR_Array with a TR::deque in TR_Instruction2SharedSlotMap
```
fjeremic: Any reason this can't be done right now?

ncough: I don't want to make significant changes this close to a deadline. The change proposed here is sufficient to cut this method's memory usage to the point where it is no longer evident on profiles.

fjeremic: Fair enough. Thanks for the explanation.

vijaysun-omr:
Looks okay to me. Merging.

PR #1521, opened 2017-08-16
Reduce OSR logging and disable by default

ncough:
Ensure OSR related tracing is properly guarded
by the TraceOSR option and disable it by default.


vijaysun-omr:
Changes only to tracing of OSR and they look fine to me. I will merge once checks pass.

PR #1526, opened 2017-08-17
Fix lowering methodExitHook simplifier

cathyzhyi:
Fix lowering methodExitHook simplifier

With inlining in debugging mode, there might be multiple method
enter/exit hooks in one extended basic block and the old way of
lowering them is wrong. This PR collect enter/exit treetops and nodes
into a list and process them after finishing all other nodes in
the same extended basic blocks to avoid spliting a block while in
the process of analyzing the block.


vijaysun-omr:
I have some concerns about splitting the block for method exit lowering while still in the process of analyzing the block; the original code would have only done method exit lowering after the block was analyzed completely (the bug in the original code being it would only lower the "last" method exit).

vijaysun-omr:
Thanks for changing the fix to address my concern. I see that both the method enter and method exit hooks are now lowered after the block is analyzed, like was happening before this commit. The only difference is that this lowering is done for multiple hooks in the same block since this was what was causing the bug originally.

cathyzhyi:
The change has been tested with forcing method enter/exit hook in local tests.

PR #1527, opened 2017-08-17
Refactor and clean up CodeCacheManager

### LINKED ISSUE #1191: JBCodeCacheManager should not define "pyfe()"
The original JitBuilder compiler project grew from an older CPython compiler project that has not (yet?) been open sourced. One holdover from that evolution is the presence of the "pyfe()" call in jitbuilder/runtime/JBCodeCacheManager.hpp . This function should be removed.

There is only one use of pyfe() which is in JBCodeCacheManager.cpp, and it should be easily adjusted to instead rely on fe().

lmaisons:
- Modifies the CodeCacheManager constructor, adding a TR::RawAllocator
parameter and removing the TR_FrontEnd parameter.

- Uses the newly provided TR::RawAllocator when allocating and freeing
memory, moving the implementation to the OMR Layer.

- Removes the _fe field and the fe and pyfe methods

- Removes the static JitConfig field, as well as its getter method in
the test compiler and JitBuilder environments.

Resolves #1191 (JBCodeCacheManager should not define "pyfe()")

### DIFF
--- a/fvtest/compilertest/runtime/TestCodeCacheManager.cpp
+++ b/fvtest/compilertest/runtime/TestCodeCacheManager.cpp
@@ -34,45 +34,16 @@
 
 
 TR::CodeCacheManager *TestCompiler::CodeCacheManager::_codeCacheManager = NULL;
-TestCompiler::JitConfig *TestCompiler::CodeCacheManager::_jitConfig = NULL;
-
+TestCompiler::CodeCacheManager::CodeCacheManager(TR::RawAllocator rawAllocator)
+   : OMR::CodeCacheManagerConnector(rawAllocator)
+   {
+   _codeCacheManager = self();
+   }
 
 TR::CodeCacheManager *
 TestCompiler::CodeCacheManager::self()
    {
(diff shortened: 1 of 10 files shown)


fvtest/compilertest/runtime/TestCodeCacheManager.cpp
```diff
@@ -31,47 +31,18 @@
...
+   }
 
 TR::CodeCacheManager *
 TestCompiler::CodeCacheManager::self()
    {
    return static_cast<TR::CodeCacheManager *>(this);
    }
 
-TestCompiler::FrontEnd *
-TestCompiler::CodeCacheManager::pyfe()
```
mstoodle: yay!

mstoodle:
You should add to the commit message that it contains a fix for #1191 :) .

mstoodle:
It's really looks a lot cleaner without the frontend and jitconfig (never felt right with those being in there). You might consider renaming the commit to "Refactor and clean up CodeCacheManager" since there's a fair amount of just nice clean-up that falls out of the refactoring here.

Looks good, @lmaisons!

PR #1528, opened 2017-08-17
Ignore irrelevant stores in UseDef control

ncough:
UseDef will give up if there are too many
uses and defs. The number of defs can increase
dramatically under voluntary OSR, due to the
use of irrelevant stores at each transition
point. This change counts these stores and
prevents them from being considered in the
UseDef control logic.


vijaysun-omr:
Looks good to me. Merging.

PR #1529, opened 2017-08-17
WIP: CMake: Add toolchain targeting x86

rwy0717:
This toolchain can be used to generate binaries for x86, with any gnuish
platform.


rwy0717:
Closing this PR. We've decided that rolling toolchains with OMR is not how we're going to address the issue of 32-bit builds on 64-bit platforms. Per offline discussion, we'll be introducing a user-facing 32-bit option. This shouldn't stop consumers from providing or using their own toolchains to build 32-bit omr.

PR #1533, opened 2017-08-18
Add exception throwing case for ZEROCHK

cathyzhyi:
Modify OMR::Node::exceptionsRaised to indicate the fact that Node
with zerochk opcode can throw exception


vijaysun-omr:
Correct change since ZEROCHK is obviously capable of raising an exception. Merging.

PR #1535, opened 2017-08-18
Compiler CMake support for POWER platforms

mgaudet:
As well as miscellaneous refactoring derived from said work.

### DIFF
--- a/cmake/compiler_support.cmake
+++ b/cmake/compiler_support.cmake
@@ -32,7 +32,6 @@ if (NOT PERL_FOUND )
    message(FATAL_ERROR "Perl not found")
 endif()
 
-set(MASM2GAS_PATH ${OMR_ROOT}/tools/compiler/scripts/masm2gas.pl CACHE INTERNAL "MASM2GAS PATH")
 
 macro(tr_detect_system_information)
 	macro(jit_not_ready)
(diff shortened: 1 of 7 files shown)


charliegracie:
@dnakamura and @rwy0717 ?

cmake/compiler_support.cmake
```diff
@@ -72,19 +71,65 @@ macro(tr_detect_system_information)
...
+		
+	elseif(OMR_HOST_OS MATCHES "aix")
+		list(APPEND TR_COMPILE_DEFINITIONS
+			${upcase_os}
+			SUPPORTS_THREAD_LOCAL
+			_XOPEN_SOURCE_EXTENDED=1 
+    			_ALL_SOURCE
+		)
 	else()
 		jit_not_ready()
```
dnakamura: maybe an `omr_assert(FALSE, .... )` would be more clear here?

dnakamura:
LGTM. Had small note re: using `omr_assert` but I'm also ok merging in as is

mgaudet:
`OMR_COMPILER_COMPILE_OPTIONS`? Screaming awkward. Historically the compiler technology has been commonly referred to by the name Testarossa, and I don't feel a particular need to wander from that particular heritage.  

The isolation here shouldn't be permanent though: Longer term, it ought to be better integrated into the rest of the compiler technology.

mgaudet:
A lot of the indenting issues pre-date this change. I will  apply a whitespace correction change separately.

mgaudet:
Addressed Robert's review comments in the above trio of commits.

rwy0717:
LGTM

PR #1538, opened 2017-08-19
Correct LoopVersioners use of isPotentialOSRPoint

ncough:
isPotentialOSRPoint will check if OSR infrastructure
has been removed, preventing yield points from appearing
as though they require special consideration after OSR
transitions can no longer be added. However, loop versioner
was using the call to identify any yield point, irregardless
of the infrastructure state.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -632,7 +632,7 @@ bool OMR::Compilation::isShortRunningMethod(int32_t callerIndex)
    return false;
    }
 
-bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node, TR::Node **osrPointNode)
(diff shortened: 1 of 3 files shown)


vijaysun-omr:
What makes this use of isPotentialOSRPoint special ? i.e. should other callers be re-examined to see if they should set the new argument you added to true ?

ncough:
Other callers use `isPotentialOSRPoint` to identify special cases where optimizations are limited by OSR. However, loop versioner is asking whether the node could be considered a yield, at which point assumptions could be invalidated during execution. This is unique to LoopVersioner, as it is the only optimization that attempts to move OSR guards.

vijaysun-omr:
Thanks for checking the other callers and clarifying. Merging.

PR #1543, opened 2017-08-21
Fix long displacement for RSL instructions

fjeremic:
RSL instructions (such as TP) do not have an index register. As such we
must treat them specially when generating long displacement logic.
Similar logic already exists for various other kinds of instructions
with memory references that do not have an index register.

To prevent future problems in this area we also take the opportunity to
add a fatal assert if this issue arises in the future.


vijaysun-omr:
Looks okay to me. I will merge once checks pass.

ncough:
LGTM

PR #1544, opened 2017-08-21
CMake: Refactor linking

dnakamura:
Refactor linking in order to better represent dependencies.
In addition provide new cache variables to allow consumer to
override the use of trace, hook, port, thread, and gc libs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,15 +85,22 @@ configure_file(./omrcfg.CMakeTemplate.h omrcfg.h)
 configure_file(./omrversionstrings.CMakeTemplate.h omrversionstrings.h)
 
 add_library(omrglue STATIC)
-# Link as private since we dont want the interface sources to propagate
 target_link_libraries(omrglue
-	PUBLIC
-	omrgc
 	PRIVATE
+	# Link as private since we dont want the interface sources to propagate
 	${OMR_GLUE_TARGET}
 )
 #but we need to propagate the include paths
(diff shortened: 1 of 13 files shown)


mgaudet:
I notice some lib duplication for link reasons -- I am guessing this works, but is that documented as working for dependency resolution? 

Other than that, I think it LGTM.

dnakamura:
@mgaudet are you referring to things with omrglue? If so, yeah the situation is a bit weird, because there are a bunch of circular dependencies between omrglue and various components.

mgaudet:
Yeah, I was talking circular include resolution. It wasn't clear to me if that was an acceptable way to solve that issue in CMake, or if some generators were free to de-duplicate the link line unless the debs were correctly indicated to CMake.

dnakamura:
Its not just the include resolution there are actual circular symbol resolution issues. From what I've looked at the link dependency issue is resolved before it hits the generator. The problem with specifying the dependencies further up is that it wont work when you are building shared libraries as cmake only allows cycles in the dependency graph when all the nodes in the cycle are static libraries  (or maybe object/interface libraries as well)

dnakamura:
Ah. just realized this will break things for z/OS since i removed the linkage between omrsig and omr_shared

charliegracie:
So I am waiting for a change to re-add the link between omrsig and omr_shared?

dnakamura:
@charliegracie  now has the `omr_shared` fix

charliegracie:
The testing says it is not complete but when I click the link the build has finished successfully.  I know travis was having issues so it is likely that it missed an update or something.  I am going to merge.

PR #1549, opened 2017-08-21
Add two prototype Tril utilities

mgaudet:
- tril_dumper takes a tril program on stdin or as a file argument,
  parses it and then dumps the AST to stdout.
- tril_compiler does the same, then compiles the code. It does nothing
  with the compiled code, but this can be useful for checking validation
  errors quickly.


vijaysun-omr:
@mstoodle a review from you would be good here too.

PR #1550, opened 2017-08-22
Remove assertions in _patchVirtualGuard

fjeremic:
It does not appear for there to be a safe way to tell whether we are
patching synchronously or asynchronously without false positives. The
asserts placed in this function are firing off on upstream projects
because runtime virtual guards that are placed as NOP guards could have
arbitrary code after them, even other NOP branches. 

As such for a synchronous virtual guard we will be clobbering the block
it is guarding and we cannot be certain what is at the location we will
be patching. It could in fact be another NOP branch which may confuse us
when trying to distinguish whether we are patching a NOP BRC / BRCL.

All assertions have been removed from this function. Furthermore the
patching order has been re-ordered to be as safe as possible with the
information we have at hand. It will be up to the caller / codegen to
ensure the non-atomic patching is safe.

### DIFF
--- a/compiler/z/runtime/VirtualGuardRuntime.cpp
+++ b/compiler/z/runtime/VirtualGuardRuntime.cpp
@@ -26,8 +26,7 @@
 
 
 /** \brief
-  *    Patch an existing BRC / BRCL virtual guard at \p locationAddr to branch to \p destinationAddr or encode a BRC /
-  *    BRCL at \p locationAddr to branch to \p destinationAddr if a NOP BRC / BRCL does not exist.
+  *    Encode a BRC / BRCL at \p locationAddr to branch to \p destinationAddr.
   *
   * \param locationAddr
   *    The location to patch.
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
I would like a review from @joransiu on this one.

joransiu:
LGTM.

The only question would be if it'd be even safer to set the 0 condition mask value first, before updating the op-code in the first byte.  I can't think of a scenario where that would be an issue (unless we mixed BRCL/BRC -- which shouldn't really happen).

fjeremic:
> The only question would be if it'd be even safer to set the 0 condition mask value first, before updating the op-code in the first byte. I can't think of a scenario where that would be an issue (unless we mixed BRCL/BRC -- which shouldn't really happen).

Fixed in 8a8f804

mgaudet:
So, random question (please don't tweak based on this): To ensure the compiler doesn't re-order your writes... shouldn't `locationAddr` be `volatile`? This obviously hasn't been an issue yet, but peeking at this it was my first thought as soon as I saw the comment  'order matters'.

fjeremic:
> So, random question (please don't tweak based on this): To ensure the compiler doesn't re-order your writes... shouldn't locationAddr be volatile? This obviously hasn't been an issue yet, but peeking at this it was my first thought as soon as I saw the comment 'order matters'.

Though about that as well, but didn't want to change it because I wanted the API to stay consistent between all platforms. You are likely right in that `locationAddr` should be made `volatile` and on all platforms consistently. Perhaps lets do this under a new PR so we can have a cross platform discussion there. I'll open one up shortly after this is merged to avoid dependent PRs.

vijaysun-omr:
Will merge once checks pass.

PR #1552, opened 2017-08-22
WIP: Prototype for representing the symbol reference table in Tril

Leonardo2718:
This prototype allows the OMR compiler's Symbol Reference Table (symreftab) to be represented in Tril. The prototype is not complete and should not be merged in its current state, so I'm marking as WIP. I am looking for feedback on the prototype (particularly from @0xdaryl, @mstoodle, and @mgaudet) regarding what (additional) information should be serialized and the technique used to serialize flags.

## Abstract

This PR does five important things:

1. Patches `OMR::Compilation::compile()` to dump the generated IL as Tril to stdout immediately after IL is generated (this needs to be undone before the PR can be merged)
2. Applies some fixes to the `TrilDumper` and `ILTraverser` classes
3. Adds the "Enum Table Factory" mechanism for mapping stringified enum names to the corresponding enum values
4. Adds a mechanism for dumping the symreftab as Tril
5. Adds a mechanism for (re-)generating the symreftab from Tril code

The last two points are the focus of this PR. The following sections describe some details of how the symreftab is represented in Tril.

## Symreftab Serialization

Currently, this prototype only supports a basic representation of symbol references (symrefs), symbols, and flags associated with each of them. (If you have suggestions for other specific things that should also be represented, please comment!)

The symreftab is a child-node of a Tril method (sibling of the method blocks):

```
(method ...
  (symreftab ...)
  (block ...)
   ...
)
```

Symrefs in the table are specified as children of `symreftab`:

```
(symreftab
  (symref ...)
  (symref ...)
   ...
)
```

Each `symref` expects one child representing the symbol it contains:

```
(symref ... (symbol ... ) )
```

The symbol kind is specified as the node-name:

```
(parm ...)   ; a parameter
(auto ...)   ; an automatic
(static ...) ; a static
(symbol ...) ; a generic symbol (e.g. a shadow)
```

A complete `symreftab` would look something like:

```
(method name="foo" returns=Int32
  (symreftab
    (symref ... (auto ...) )
    (symref ... (static ...) )
    (symref ... (symbol ...) )
     ...
  )
  (block ...)
  (block ...)
   ...
)
```

## Serializing Flags

Both symbols and symrefs can have flags associated with them. In the code, these are stored as bit-fields with enum values used as masks for the different flags. Because the enum names do not have a one-to-one mapping with values (i.e. multiple names map to the same value), serializing them directly is rather challenging. The mechanism proposed in this prototype tries to make manually specifying flags in Tril code easy and dumping flags from IL efficient.

To achieve these goal, flag states can be represented in three parts:

1. `rawflags`: the "initial" raw value of the bit-field used to store the flags (an integer value)
2. `setflags`: a list of additional flags to be set (modifier for `rawflags`)
3. `clearflags`: a list of additional flags to be cleared (modifier for `rawflags`)

If multiple bit-fields are used to store flags, the `*flags` argument names have an additional number at the end that corresponds to the bit-field number:

- `rawflags`, `setflags`, and `clearflags` apply to the first bit-field
- `rawflags2`, `setflags2`, and `clearflags2` apply to the second bit-field
- `rawflags3`, `setflags3`, and `clearflags3` apply to the third bit-field
- etc.

When the symreftab is dumped, flags are dumped as raw bit-fields. An example symreftab dump:

```
(symreftab
  (symref offset=0 rawflags=0x20000000 (parm name="(null)" type=Address size=8 rawflags=0x40000107 rawflags2=0))
  (symref offset=0 rawflags=0x20000000 (symbol name="foo" type=Int32 size=4 rawflags=0x603 rawflags2=0x200))
  (symref offset=0 rawflags=0x20000000 (symbol name="bar" type=Int32 size=4 rawflags=0x603 rawflags2=0x200))
  (symref offset=0 rawflags=0x20000000 (auto name="(null)" type=Int32 size=4 rawflags=0x3 rawflags2=0))
  (symref offset=0 rawflags=0x20000000 (symbol name="quux" type=Int32 size=4 rawflags=0x603 rawflags2=0x200))
)
```

When the symreftab is (re-)generated from Tril code, the raw bit-field (defaults to 0 if not specified) is first copied as-is into memory; the integer value is simply written to the bit-field used to store the flags. Then, if specified, additional flags that must be set/cleared are iterated over and the bit-field is modified accordingly.

This mechanism makes it easy to dump flags as no processing is done to identify the meaning of the raw bit-field. It is also easy for a user to manually set/clear flags by name if so desired.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -878,6 +878,9 @@ static int32_t strHash(const char *str)
    return result;
    }
 
+#include "infra/TrilDumper.hpp"
+#include <stdio.h>
+
 int32_t OMR::Compilation::compile()
    {
 
@@ -977,6 +980,17 @@ int32_t OMR::Compilation::compile()
      _ilGenSuccess = _methodSymbol->genIL(self()->fe(), self(), self()->getSymRefTab(), _ilGenRequest);
      if (printCodegenTime) genILTime.stopTiming(self());
    }
+
+   if (_ilGenSuccess)
+      {
+      auto comp = self();
+      auto sym = comp->getJittedMethodSymbol();
+      TR::TrilDumper dumper{stderr};
+      TR::ILTraverser traverser(comp);
+      traverser.registerObserver(&dumper);
+      traverser.traverse(sym);
+      fprintf(stderr, "\n");
+      }
 
    // Force a crash during compilation if the crashDuringCompile option is set
    TR_ASSERT_FATAL(!self()->getOption(TR_CrashDuringCompilation), "crashDuringCompile option is set");
--- a/compiler/il/OMRSymbolReference.hpp
+++ b/compiler/il/OMRSymbolReference.hpp
@@ -304,26 +304,12 @@ protected:
    TR_BitVector *getUseonlyAliasesBV(TR::SymbolReferenceTable *symRefTab);
    TR_BitVector *getUseDefAliasesBV( bool isDirectCall = false, bool gcSafe = false);
 
-   enum // flags
-      {
-      Unresolved                        = 0x00000001,
-      CanGCandReturn                    = 0x00000002,
-      CanGCandExcept                    = 0x00000004,
-      ReallySharesSymbol                = 0x00000008,
-      StackAllocatedArrayAccess         = 0x00000010,
-      SideEffectInfo                    = 0x00000020,
-      LiteralPoolAddress                = 0x00000040,
-      FromLiteralPool                   = 0x00000080,
-      OverriddenBitAddress              = 0x00000100,
-      InitMethod                        = 0x00000200, ///< J9
-      TempVariableSizeSymRef            = 0x00000400,
-      Adjunct                           = 0x00000800, ///< auto symbol represents the adjunct part of the dual symbol
-      Dual                              = 0x00001000, ///< auto symbol represents a dual symbol consisting of two parts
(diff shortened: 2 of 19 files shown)


mgaudet:
My feedback is split into two pieces: 

### Tril

I don't mind the symref format, though, there's a couple things that _might_ need to be addressed

1. I'd prefer to optimize Tril for writability where possible, which here I think means trying to the extent possible to keep as many of these annotations on nodes optional. _i.e_ `offset` elided can be assumed to mean zero, etc. 
2. I have no intuition for how often this happens, but in theory, SymRefs can share symbols -- it's not clear how you'd represent that above (may I suggest `@id` style commoning. 

### Changes

This makes it so clear the benefits and clarity we would get from adopting a descriptor-generator pattern for structured data like enums and opcode tables. I don't think we need to go full [TableGen](http://llvm.org/docs/TableGen/), but I would rather invest in that kind of a story than another one-of-a-kind magical machine (clever though it is!) for dealing with that.

mstoodle:
Wondering what we should do with this PR at this point, @Leonardo2718 . You are understandably busy at the moment with school, but I'm wondering if this PR sitting here is really providing much value... Maybe close and re-submit when you have time to get back to it?

Leonardo2718:
The original purpose of this PR was to collect feedback on the approach used to represent the symbol reference table in Tril. It has indeed been a while since I opened this PR, so it's probably about time I close it.

I will open a new PR once I have time to work on this some more.

PR #1554, opened 2017-08-22
Add missing JitBuilder examples to CMake 

mgaudet:
Also add the missed virtual machine modelling files to the CMake source lists.


vijaysun-omr:
Looks safe to me. Merging.

Issue #1556, opened 2017-08-22
Reduce ifcmpge to ifcmpeq


last5bits:
The following is a canonical case for the test under mask (TM) optimization. The issue here is that we only carry out that optimization on if*eq type trees. However, the above tree given the iand can be reduced from ifiucmpge to ifiucmpeq and we should be able to perform the optimization. The reduction is possible since the constants are equal and the case of greater-than will never happen. In other words, the result of `iand` cannot be greater than the constant.

```
 n347n    (  0)  ifiucmpge --> block_12 BBStart at n211n ()                                           [0x000003ff7e314998] bci=[0,106,352] rc=0 vc=551 vn=- li=- udi=- nc=3 flg=0x20
 n342n    (  0)    iand (in GPR_0131) (X>=0 cannotOverflow )                                          [0x000003ff7e314830] bci=[0,103,352] rc=0 vc=551 vn=- li=- udi=- nc=2 flg=0x1100
 n340n    (  0)      iloadi  <isArray>[#532  Shadow +16] [flags 0x603 0x0 ] (in GPR_0131) (cannotOverflow )  [0x000003ff7e3147a0] bci=[0,103,352] rc=0 vc=551 vn=- li=- udi=- nc=1 flg=0x1000
 n338n    (  0)        aloadi  <classRomPtr>[#525  Shadow +8] [flags 0x10607 0x0 ] (in GPR_0133) (X>=0 isUnsignedLoad sharedMemory )  [0x000003ff7e314710] bci=[0,103,352] rc=0 vc=565 vn=- li=- udi=- nc=1 flg=0x100
 n337n    (  0)          aloadi  <classFromJavaLangClass>[#527  Shadow +8] [flags 0x10607 0x0 ] (in GPR_0135) (X!=0 X>=0 isUnsignedLoad sharedMemory )  [0x000003ff7e3146c8] bci=[0,103,352] rc=0 vc=564 vn=- li=- udi=- nc=1 flg=0x104
 n528n    (  1)            ==>aRegLoad (in &GPR_0128) (X!=0 X>=0 SeenRealReference sharedMemory )
 n341n    (  0)      iconst 0x10000 (in GPR_0130) (X!=0 X>=0 )                                        [0x000003ff7e3147e8] bci=[0,103,352] rc=0 vc=551 vn=- li=- udi=- nc=0 flg=0x104
 n341n    (  0)    ==>iconst 0x10000 (in GPR_0130) (X!=0 X>=0 )
[...]

PR #1557, opened 2017-08-22
Reduce ifcmpge to ifcmpeq

last5bits:
Please see #1556


last5bits:
@fjeremic Please take a look whenever you have time.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
    }
 
 
+/* Reduce Ifcmp Greater-Than-Or-Equal to Equal in the following case. If
+* - the first child is an "and" node
+* - the second child of the "and" node is a constant N
+* - the second child of the ifcmple node is a constant N,
```
fjeremic: Did you mean `ifcmpge` rather than `ifcmple`?

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+   auto param = GetParam();
+
+   const char *type = std::get<0>(param);
+   const char *sign = std::get<1>(param);
+   const char *lege = std::get<2>(param);
+   const char *val  = std::get<3>(param);
+
+   char inputTrees[600] = {0};
+   std::snprintf(inputTrees, 600,
+         R"((method return="NoType"
```
Leonardo2718: Have you verified that raw string literals are available on all supported compilers? (I didn't think they were but I could be wrong).

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+* ifiucmpeq --> block_5
+*  iand
+*    iload  <...>
+*    iconst 16
+*  iconst 16
+*
+* Since the constants are equal, the case of greater-than/less-than is impossible,
+* and we can reduce a GE/LE operation to an EQ operation. After the reduction is performed,
+* the test-under-mask optimization can be applied to the "ifcmpeq" node.
+*/
```
fjeremic: Test-under-mask is an IBM Z specific optimization we do at code generation time. Although we take advantage of it this simplifier case simply strengthens the condition from `>=` to `==` the latter of which conveys more information about the operands for other optimizations.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+class IfcmpReducer
+   {
+   public:
+   IfcmpReducer(TR::Node *node, TR::Simplifier *s) : _node(node), _s(s) {}
+
+   static TR::ILOpCodes getNewOpCode(TR::Node *node)
+      {
+      auto opCodeValue = node->getOpCodeValue();
+      if (opCodeValue == TR::ifbcmple || opCodeValue == TR::ifbcmpge)
```
fjeremic: The `if` `else` seems overkill for this many cases. Is it possible to use a `switch` here?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+* Since the constants are equal, the case of greater-than/less-than is impossible,
+* and we can reduce a GE/LE operation to an EQ operation. After the reduction is performed,
+* the test-under-mask optimization can be applied to the "ifcmpeq" node.
+*/
+class IfcmpReducer
+   {
+   public:
+   IfcmpReducer(TR::Node *node, TR::Simplifier *s) : _node(node), _s(s) {}
+
+   static TR::ILOpCodes getNewOpCode(TR::Node *node)
```
fjeremic: Can we Doxygen document this function?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+         return TR::iflucmpeq;
+         }
+      else
+         {
+         TR_ASSERT(false, "Ifcmp reduction is not implemented for node [%p]\n", node);
+         }
+      }
+
+   template<typename T>
+   bool reducible() const
```
fjeremic: Documentation needed here.

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+ * Then, open the trace log and witness the transformation.
+ */
+TEST_P(IfcmpReductionTest, Reduction)
+   {
+   auto param = GetParam();
+
+   const char *type = std::get<0>(param);
+   const char *sign = std::get<1>(param);
+   const char *lege = std::get<2>(param);
+   const char *val  = std::get<3>(param);
```
Leonardo2718: Since you access the tuple both here and in `reductionTestNameGenerator()`, have you considered defining a struct that gives names to the tuple elements and defining a `to_struct()` function that converts an `IfcmpReductionParamType` to a struct instance (similar to `BinaryOpParamType` and `BinaryOpParamStruct` in `OpCodeTest.hpp`)?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+
+      if (firstChild->getOpCode().isAnd() &&
+         firstChild->getSecondChild()->getOpCode().isIntegralConst() &&
+         secondChild->getOpCode().isIntegralConst())
+         {
+         T constAnd = firstChild->getSecondChild()->getConst<T>();
+         T constIfcmp = secondChild->getConst<T>();
+
+         if (constAnd == constIfcmp)
+            return true;
```
fjeremic: This may not be correct. For example `if ((x & 3) <= 3)` is not equivalent to `if ((x & 3) == 3)` (let `x == 1` and this becomes apparent). In fact for all `<=` cases you capture `if ((x & 3) <= 3)` can be reduced to `true`. Its the `>=` that are special.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,111 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+         }
+      return false;
+      }
+
+   TR::Node *reduce()
```
fjeremic: Documentation needed here.

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+ * 2) Signed / unsigned (e.g., ificmp / ifiUcmp)
+ * 3) Less than or equal / greater than or equal
+ * 4) Reduction can be performed (the const value in "and" is equal to the value in "ifcmp") / reduction cannot be performed
+ */
+INSTANTIATE_TEST_CASE_P(IfcmpReduction, IfcmpReductionTest,
+        ::testing::Combine(
+            ::testing::Values("i", "l", "b", "s"),
+            ::testing::Values("u", ""),
+            ::testing::Values("ge", "le"),
+            ::testing::Values("16", "17")),
```
fjeremic: Can you add some more test cases covering what was observed above with `if ((x & 3) <= 3)` for example. We should also cover a variety of corner cases here.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: I agree that your description would be more clear to the reader. However, just to clarify, I am fairly sure that we have the same optimization for x86 too (at least).

fjeremic:
> I agree that your description would be more clear to the reader. However, just to clarify, I am fairly sure that we have the same optimization for x86 too (at least).

Not sure. @0dvictor can we do this on x86? Do we have an in-memory test instruction so we can avoid loading a value into a register?

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: > I agree that your description would be more clear to the reader. However, just to clarify, I am fairly sure that we have the same optimization for x86 too (at least).

Not sure. @0dvictor can we do this on x86? Do we have an in-memory test instruction so we can avoid loading a value into a register?

0dvictor: Yes, X86 support memory vs constant comparison. By reducing  `ifcmpge/le` to `ifcmpeq`, X86 could potentially generate one instruction less. It can never be worse (in theory).

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+ ******************************************************************************/
+
+#include <gtest/gtest.h>
+#include "Jit.hpp"
+#include "jitbuilder_compiler.hpp"
+
+#define ASSERT_NULL(pointer) ASSERT_EQ(nullptr, (pointer))
+#define ASSERT_NOTNULL(pointer) ASSERT_TRUE(nullptr != (pointer))
+
+extern "C" bool initializeJitWithOptions(char *options);
```
last5bits: Fixed.

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+   public:
+
+   static void SetUpTestCase()
+      {
+      std::string defaultOptions = "-Xjit:acceptHugeMethods,enableBasicBlockHoisting,omitFramePointer,useIlValidator,paranoidoptcheck";
+      // Disable the following optimizations; otherwise, the trees are optimized into nothing.
+      std::string userOptions = "disableGlobalVP,disableLocalCSE";
+      std::string options = defaultOptions + "," + userOptions;
+
+      auto initSuccess = initializeJitWithOptions((char*)options.c_str());
```
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: The comment is fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Thanks, Filip, I didn't think the less-than-or-equals case through. Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

last5bits:
@Leonardo2718 Leonardo, thank you for your valuable comments. Could you please give me some advice about the following problem with GTest. As per Filip's suggestion, I want to instantiate tests that would verify corner cases: small/large integers, positives/negatives, etc. I tried to use `TRTest::const_value_pairs` that you provided buy I ended up duplicating my test cases (you can see that in the newer version of the code I've just pushed).

Do you have some template magic in mind that could help me with the duplication? Thank you in advance.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: The issue for the LE case: #1564.

Leonardo2718:
@last5bits The duplication is annoying, I agree. It's a problem that will have to be dealt with in all Tril-based tests eventually.

In your specific case, instead of passing the constant values themselves, you can pass the stringified values (like you had before). So you have two options (that I can think of off the top of my head):

1. Specify the input values as strings the same way you had before, but covering more edge-cases (maybe taking inspiration from `TRTest::const_values()`)
2. Use `TRTest::const_value_pairs()` to generate input values and then stringify them before feeding them to the test

Option 1 is the easiest, though perhaps less elegant, of the two.

Options 2 has much more potential for generalization but likely requires a bit of template programming. My approach would be to use [std::transform](http://en.cppreference.com/w/cpp/algorithm/transform) (or a custom wrapper for it) to stringify the list of values returned by `TRTest::const_value_pairs()`.

For now, I'm okay with either solution so I will let you choose whichever one you prefer (or come up with a different solution altogether). If you choose option 1 BTW, someone later on can spend the time to properly develop option 2.

Hopefully that helps :wink:.

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+class IfcmpReductionTest : public ::testing::TestWithParam<IfcmpReductionParamType>
+   {
+   public:
+
+   static void SetUpTestCase()
+      {
+      std::string defaultOptions = "-Xjit:acceptHugeMethods,enableBasicBlockHoisting,omitFramePointer,useIlValidator,paranoidoptcheck";
+      // Disable the following optimizations; otherwise, the trees are optimized into nothing.
+      std::string userOptions = "disableGlobalVP,disableLocalCSE";
+      std::string options = defaultOptions + "," + userOptions;
```
last5bits: Fixed. Not sure that the why is necessary; I assume that a curious reader would eyeball the tree and decide why for themselves.

fvtest/compilertriltest/IficmpTest.cpp
last5bits: Fixed.

fvtest/compilertriltest/IficmpTest.cpp
last5bits: They are not, good catch. Fixed.

fvtest/compilertriltest/IficmpTest.cpp
last5bits: Fixed. However, bear in mind that currently tests are not automatic. To verify that the transformation is happening correctly, the developer should run the test manually and examine the trace log.

There is an issue opened in regard to that: #1558.

last5bits:
@Leonardo2718 Thanks a lot, your advice definitely helped. I went with the option two. The code is ready for a second review, feel free to take another look.

fvtest/compilertriltest/IficmpTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+ * Then, open the trace log and witness the transformation.
+ */
+TEST_P(IfcmpReductionTest, Reduction)
+   {
+   auto param = to_struct(GetParam());
+
+   std::string type = param.type;
+   std::string sign = param.sign;
+   std::string val1 = param.val1;
+   std::string val2 = param.val2;
```
Leonardo2718: I suppose this works, though I would have probably just used `param.*` everywhere :smile: .

last5bits: That's way too much to type :P

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13570,10 +13695,37 @@ TR::Node *ifCmpWithEqualitySimplifier(TR::Node * node, TR::Block * block, TR::Si
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
 
+   TR::ILOpCodes opCode = node->getOpCodeValue();
+
+   // Reduce ifcmpge/le to ifcmpeq in specific cases
```
fjeremic: Comment needs updating. `ifcmple` is not reduced.

last5bits: Fixed.

andrewcraik:
@jdmpapin review of the simplifier change commit please.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,118 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+class IfcmpgeReducer
+   {
+   public:
+   IfcmpgeReducer(TR::Node *node, TR::Simplifier *s) : _node(node), _s(s) {}
+
+   /**
+    * @brief Get a specific ifcmpeq opcode based on a ifcmpge opcode.
+    *
+    * Each ifcmpge opcode is mapped to its corresponding ifcmpeq opcode.
+    * E.g., ifiucmpge should be substituted with ifiucmpeq.
```
jdmpapin: Equality is insensitive to signedness, so there's no issue changing `ifiucmpge` into `ificmpeq` (without the `u`).  Then this function would just be `TR::ILOpCodes::ifcmpeqOpCode(node->getDataType())`.  The `ucmpeq` and `ucmpne` opcodes should probably be considered deprecated, as with `uadd` and `usub`

last5bits: Thanks for your suggestion, Davin. I changed the function but now I'm getting this assertion when I run my tests:
```
Assertion failed at /sandbox/azhikhar/omr/compiler/il/OMRILOps.hpp:1053: 0

        no ifcmpeq opcode for this datatype NoType
```

I might be missing something. @Leonardo2718 is it possible that Tril does not set the data type for the node?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13019,11 +13131,15 @@ TR::Node *ificmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
...
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
 
+   IfcmpgeReducer ifcmpReducer(node, s);
+
    if (node->getOpCodeValue() == TR::ificmpge)
       {
       if (firstChild->getOpCode().isLoadConst() && conditionalBranchFold((originalFirst->getInt() >= originalSecond->getInt()), node, firstChild, secondChild, block, s))
          return node;
       intCompareNarrower(node, s, TR::ifsucmpge, TR::ifscmpge, TR::ifbcmpge);
+      if (ifcmpReducer.reducible<int32_t>())
```
jdmpapin: The preceding call to `intCompareNarrower` can change the comparison opcode from `ificmpge` to one of `ifsucmpge`, `ifscmpge`, or `ifbcmpge`, in which case we'd want `reducible<uint16_t>`, `reducible<int16_t>`, or `reducible<int8_t>`, respectively.

It's called again in the `ifiucmpge` branch below, and `iflcmpgeSimplifier` has a similar call to `longCompareNarrower`.

I'd recommend making the existing `reducible` private and renaming it, and wrapping it with something like the switch you have below in `ifCmpWithEqualitySimplifier`, so that callers can simply ask `ifcmpReducer.reducible()`

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13570,10 +13695,37 @@ TR::Node *ifCmpWithEqualitySimplifier(TR::Node * node, TR::Block * block, TR::Si
...
+         case TR::ifscmpge:
+            if (ifcmpReducer.reducible<int16_t>())
+               node = ifcmpReducer.reduce();
+            break;
+
+         case TR::ifsucmpge:
+            if (ifcmpReducer.reducible<uint16_t>())
+               node = ifcmpReducer.reduce();
+            break;
+      }
```
jdmpapin: If you wrap `reducible<T>` as I suggested above, this switch would simplify into something like this:

    TR::ILOpCode op = node->getOpCode();
    if (op.isCompareForOrder() && op.isCompareTrueIfGreater() && ifcmpReducer.reducible())
       node = ifcmpReducer.reduce();

Note that `op.isCompareTrueIfEqual()` is guaranteed by the choice of opcodes that use `ifCmpWithEqualitySimplifier`, and it should be preserved by `makeConstantTheRightChildAndSetOpcode` (through `getOpCodeForSwapChildren`).

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,118 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+
+      if (firstChild->getOpCode().isAnd() &&
+         firstChild->getSecondChild()->getOpCode().isIntegralConst() &&
+         secondChild->getOpCode().isIntegralConst())
+         {
+         T constAnd = firstChild->getSecondChild()->getConst<T>();
+         T constIfcmp = secondChild->getConst<T>();
+
+         if (constAnd == constIfcmp)
+            return true;
```
jdmpapin: This needs to also check that either the comparison is unsigned, or the mask has the high/sign bit clear.  Otherwise it's possible for the result of `iand` to compare greater.  For example, consider the following tree:

    ificmpge
      iand
        foo
        iconst -2
      iconst -2

If `foo` is non-negative, then the result of `iand` will also be non-negative, and therefore > -2.  Slightly more aggressively, it should be possible to transform this tree in cases where `foo` is known to be negative.

compiler/optimizer/OMRSimplifierHandlers.cpp
jdmpapin: Oh sorry, I should have said to take the data type of a child

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: Nice catch @jdmpapin. @Leonardo2718 surely this should have been caught by the unit tests as `std::vector<T> TR::Test::const_values()` which is used by @last5bits returns all combinations including `negative_value<T>` which was -42 in this case. Any idea why we don't have a failing unit test here?

fjeremic: On second thought presumably the following test would have been generated:

```
(method return="NoType"
  (block name="b" fallthrough="u"
    (istore temp="val"
       (iconst 1) )
      (ificmpge target="t
          (iand"
              (iload temp="val")
              (iconst -42))
          (iconst -42) ) )
  (block name="u" (return))
  (block name="t" (return))
```

However I don't see us comparing the result of the test against a valid output. So there is a failing test case here but we are just not verifying it.

Leonardo2718: I think this may be a good enough reason to justify calling the compiled body.

fjeremic: What we really need to do for these tests is to modify the TRIL generated code to be something like this (using above as example):

```
(method return="Int32"
  (block name="b" fallthrough="u"
    (istore temp="val"
       (iconst 1) )
      (ificmpge target="t
          (iand"
              (iload temp="val")
              (iconst -42))
          (iconst -42) ) )
  (block name="u" (ireturn (iconst 1)))
  (block name="t" (ireturn (iconst 0)))
```

And test this method with:

```
ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;

auto entry_point = compiler.getEntryPoint<int32_t (*)(void)>();
volatile auto exp = param.oracle(param.lhs, param.rhs);
volatile auto act = entry_point();
ASSERT_EQ(exp, act);
```

Where `param.orcale` is constructed using the lambda:

```
[]() { return ((1 & -42) >= 1); }
```

Leonardo2718: Seems reasonable to me @fjeremic.

There is some room for simplifying things here (compared to what's in `ArithmeticTest.cpp`) because we're only dealing with one opcode.

> We should also add several more constants in `TRTest::const_values()`.

Definitely. This will likely help other tests as well.

jdmpapin: I don't know what optimizations are applied when compiling from tril, but the test above would just constant fold the entire branch if local CSE were allowed to run before simplifier.  It may be a good idea to change `val` into a parameter, to ensure that it's completely opaque

Leonardo2718: @jdmpapin 

Tril uses JitBuilder as a backend so it will run whatever optimization strategy is used there. However, this specific test is explicitly run with `disableGlobalVP` and `disableLocalCSE`. So, that deals with part of the problem.

> It may be a good idea to change val into a parameter, to ensure that it's completely opaque.

Definitely a worthwhile option.

Ideally, I would prefer to test both cases (when val is a parameter and when val is constant), but that's an improvement for later.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,151 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+*    iload  <...>
+*    iconst 16
+*  iconst 16
+*
+* Since the constants are equal, the case of greater-than is impossible,
+* and we can reduce ifcmpge to ifcmpeq. In other words, the result of iand
+* cannot be greater than the constant, and we can simply perform an equality check,
+* which provides further optimization opportunities.
+*/
+class IfcmpgeReducer
```
mstoodle: I'm concerned about the relatively generic verb "reduce" here. I think either this class should be named more specifically something like `IfcmpgeToIfcmpeqReducer` or the `reducible()`, `reduce()` and `reduceSpecific()` API should indicate "reduction" towards an equality comparison opcode.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,151 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+    * @brief Get a specific ifcmpeq opcode based on a ifcmpge opcode.
+    *
+    * E.g., ificmpge should be substituted with ificmpeq.
+    * Note, that equality is insensitive to signedness, so there's no
+    * issue changing ifiucmpge into ificmpeq (without the u).
+    *
+    * @param node An ifcmpge node
+    * @return An ifcmpeq node
+    */
+   static TR::ILOpCodes getNewOpCode(TR::Node *node)
```
mstoodle: Somewhat weaker complaint about generic "new" opcode request, though I think the same future proofing should be applied here.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,151 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+               return reducibleSpecific<int32_t>();
+            case TR::ifiucmpge:
+               return reducibleSpecific<uint32_t>();
+
+            case TR::iflcmpge:
+               return reducibleSpecific<int64_t>();
+            case TR::iflucmpge:
+               return reducibleSpecific<uint64_t>();
+         }
+      }
```
jdmpapin: This should return a value in all cases.  Maybe default to false?

compiler/optimizer/OMRSimplifierHandlers.cpp
mstoodle: it would still be correct for LE if the constant is a power of 2, though seems like a weird scenario. I'm ok to not catch this case in this PR :) .

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Ah, gotcha. Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,151 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+    *
+    * @param node An ifcmpge node
+    * @return An ifcmpeq node
+    */
+   static TR::ILOpCodes getNewOpCode(TR::Node *node)
+      {
+      // TODO: uncomment once the assertion is fixed (the assertion is exposed in the tril tests):
+      // Assertion failed at omr/compiler/il/OMRILOps.hpp:1053: 0
+      //         no ifcmpeq opcode for this datatype NoType
+      //return TR::ILOpCode::ifcmpeqOpCode(node->getDataType());
```
jdmpapin: In case my other comment was buried, you should be able to use the type of either child here, e.g. `ifcmpeqOpCode(node->getChild(0)->getDataType())`.  Both children have to have the same type for the comparison to be well-typed.  My bad for suggesting `node->getDataType()` 😅 

If that works for you, then this function is probably unnecessary

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13570,10 +13725,18 @@ TR::Node *ifCmpWithEqualitySimplifier(TR::Node * node, TR::Block * block, TR::Si
 
    makeConstantTheRightChildAndSetOpcode(node, firstChild, secondChild, s);
 
+   TR::ILOpCodes opCode = node->getOpCodeValue();
```
jdmpapin: Looks like this is now unused except in its original scope

compiler/optimizer/OMRSimplifierHandlers.cpp
mstoodle: I find this comment confusing. How does a tril test manage to as for the ifcmpeqOpCode for a node that has NoType? Especially given your implementation throws an assertion for anything other than the 8 well defined opcodes in your switch?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13035,6 +13180,10 @@ TR::Node *ificmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
       unsignedIntCompareNarrower(node, s, TR::ifsucmpge, TR::ifbucmpge);
       }
 
+   IfcmpgeReducer ifcmpReducer(node, s);
+   if (ifcmpReducer.reducible())
+      return ifcmpReducer.reduce();
+
```
mstoodle: is there no possible reason to call partialRedundantCompareElimination on the node even if we reduce it? it seems like there are other places where we just reassign node and continue along ...

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -13256,6 +13406,11 @@ TR::Node *iflcmpgeSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier
       if (firstChild->getOpCode().isLoadConst() && conditionalBranchFold((((uint64_t)originalFirst->getLongInt())>=((uint64_t)originalSecond->getLongInt())), node, firstChild, secondChild, block, s))
          return node;
       }
+
+   IfcmpgeReducer ifcmpReducer(node, s);
+   if (ifcmpReducer.reducible())
+      return ifcmpReducer.reduce();
```
mstoodle: same comment: do we have to return here?

compiler/optimizer/OMRSimplifierHandlers.cpp
mstoodle: not a big deal, but why did this definition move here?

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: `node->getDataType()` returns NoType when the node is an `ifcmpge` node. Or, at least, that's what I've witnessed. Regardless, as per Davin's suggestion, I fixed it.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Historical reasons :sweat_smile: Fixed.

last5bits:
> Also, please clean up the commit titles (you're not targeting ifcmple anymore) and put a bit more detail into the commit messages.

Fixed. Not sure if the test-adding commit needs more elaboration, please let me know if so.

compiler/optimizer/OMRSimplifierHandlers.cpp
last5bits: Fixed the code but I'd rather keep the function; I feel like it is more readable this way.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -5260,6 +5260,144 @@ static bool skipRemLowering(int64_t divisor, TR::Simplifier *s)
...
+      TR::Node * firstChild = _node->getFirstChild(), * secondChild = _node->getSecondChild();
+
+      if (firstChild->getOpCode().isAnd() &&
+         firstChild->getSecondChild()->getOpCode().isIntegralConst() &&
+         secondChild->getOpCode().isIntegralConst())
+         {
+         T constAnd = firstChild->getSecondChild()->getConst<T>();
+         T constIfcmp = secondChild->getConst<T>();
+
+         bool isCmpUnsigned = _node->getOpCode().isUnsigned();
```
jdmpapin: I think the query you want here is `isUnsignedCompare`

last5bits:
Folks, thank you so much for all your valuable suggestions.

@fjeremic Filip, I think I fixed most of them; one important thing left is to improve the tests. I leave it to you, trust me, Tril is fun :) And very convenient to work with.

Filip, I also added you to my fork's collaborators. If you wish to build tril under Linux on Z, rebase off the `reduce-ificmp-new` branch, it has all the needed cmake fixes. However, it is not necessary since x86 has the same test-under-mask optimization, so you'll see a similar instruction after the reduction (`test` instead of `TM`).

To build the tests, go to `omr/build` and run
```
cmake -DOMR_JITBUILDER=1 -DOMR_FVTEST=1 .. && make
```
After that, you'll find a `comptest` binary in `omr/build/fvtest/compilertriltest`. The command line for running comptest is described in `IficmpTest.cpp`.

Well, it was a great ending for an awesome internship :) Thank you all and good luck!

mgaudet:
ILVerifier support has been added to Tril in https://github.com/eclipse/omr/pull/1664 

This will allow automated verification to replace the hand verification currently in the tests for this.

andrewcraik:
@last5bits this item looks like it was getting close to being done - are you planning to finish the last few things up so it can be re-reviewed and merged?

fjeremic:
> @last5bits this item looks like it was getting close to being done - are you planning to finish the last few things up so it can be re-reviewed and merged?

I'm working on finishing this up. I was blocked by https://github.com/eclipse/omr/issues/1697 for a bit but I've got it working now on Linux on z. Just a matter of fixing up the Tril tests. I'll allocate some time to work on this today.

andrewcraik:
Ok super it isn't a huge rush - it was more this seemed like some good stuff to get in and I'd like to be supportive if there are questions/concerns etc around next steps and what else is needed etc etc

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: Fixed in 8f65726

compiler/optimizer/OMRSimplifierHandlers.cpp
fjeremic: Fixed all of this in bb54ec4

fjeremic:
> Also, please clean up the commit titles (you're not targeting ifcmple anymore) and put a bit more detail into the commit messages.

@mstoodle fixed. As the committer assigned to this PR could you modify the PR title to "Reduce ifxcmpge to ifxcmpeq" as @last5bits may no longer be active on this PR?

fjeremic:
@Leonardo2718 @mgaudet could I please get a review of the Tril tests? @andrewcraik @jdmpapin could I please get a final review of the simplifier changes? Thanks everyone!

mgaudet:
Just looking at the test commit, looks pretty good to me. 

Given #1664, would it be possible to write a verifier for this transform? 

( #1661 has stagnated a bit, but I have a local branch that has a more stable optimization specification system that I would like to build on this test once I have it delivered too)

fjeremic:
> Given #1664, would it be possible to write a verifier for this transform?

I think so. Though I'm tempted to leave this as follow-on work. The way `IlVerifier` seems to work I would have to manually partition the 648 tests into two distinct sets of tests on which the reduction is expected to perform (i.e. look for the `ifxcmpeq` tree) and tests on which it is not possible to perform the reduction. This seems like a non-trivial amount of work. Would it be ok to open an issue for this as an extension?

mgaudet:
Absolutely! I'd encourage it in fact.

fvtest/compilertriltest/IfxcmpgeReductionTest.cpp
```diff
@@ -0,0 +1,440 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2017 IBM Corp. and others
```
Leonardo2718: Was (the contents of) this file really created in 2000?

fjeremic: Good ol' copy/paste bugs. This is indeed incorrect. Fixed in eeff48c

jdmpapin:
Simplifier changes lgtm

fjeremic:
@mstoodle any additional concerns regarding this change? It seems ip-validation also failed for some reason. It seems it doesn't like that there are multiple commit authors. Not sure how to get around this though.

charliegracie:
@fjeremic If you look at the author and committer tags are they both the same and have your proper email address?

mgaudet:
(Might be worth trying `git commit --amend --reset-author` )

fjeremic:
> @fjeremic If you look at the author and committer tagsfor commit eeff48c are they both the same and have your proper email address?

Yes it's the same one I use for all my commits. The tooling might be confused because @last5bits added me as a contributor to his fork so I can complete this PR and I've added new commits to his branch.


> (Might be worth trying git commit --amend --reset-author )

Seems this did the trick, thanks! Though @last5bits does not get credit for being the author now 😢. It seems the tooling only likes it if the committer and author match.

last5bits:
I can live with that :)

andrewcraik:
@fjeremic was this ready to go? looks like there is a merge conflict now so some kind of rebase is probably needed before it can be reviewed...

fjeremic:
> @fjeremic was this ready to go? looks like there is a merge conflict now so some kind of rebase is probably needed before it can be reviewed...

Yes it was ready to go. Waiting on @mstoodle to review as he requested changes which have been addressed. I'll resolve the merge conflict now.

fjeremic:
@mstoodle any other concerns?

mstoodle:
genie-omr build all

mstoodle:
@fjeremic could you look into the comptest failure please?

fjeremic:
> @fjeremic could you look into the comptest failure please?

So I've just confirmed these Power failures are a duplicate of https://github.com/eclipse/omr/issues/1804. The tests fail on the master branch even without my changes here. This is a pre-existing issue and is a byproduct of switching to CMake builds on Power which now transparently starts to invoke `ctest` which runs the Tril tests which fail on Power.

If you ran the genie-omr build on any other rebased PR on top of current master you will find the build will also fail. I've left a note on the #1804. We should address this soon.

charliegracie:
I am surprised about that.  Other PR and branch builds successfully passed today when using CMake.  You can see this one as an example:
https://ci.eclipse.org/omr/job/Build-linux_ppc-64_le_gcc/77.

Issue #1804 discussing one known set of comptest failures but there was a PR to exclude the known failures.

fjeremic:
I'm getting the following failures:

```
[==========] 3271 tests from 27 test cases ran. (6169 ms total)
[  PASSED  ] 3240 tests.
[  FAILED  ] 31 tests, listed below:
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/24, where GetParam() = (('\x1' (1), 7), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/28, where GetParam() = (('\xFF' (-1), 0), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/30, where GetParam() = (('\xFF' (-1), 1), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/32, where GetParam() = (('\xFF' (-1), 5), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/38, where GetParam() = (('\xFF' (-1), 7), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/56, where GetParam() = (('\xD6' (-42), 0), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/58, where GetParam() = (('\xD6' (-42), 1), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/60, where GetParam() = (('\xD6' (-42), 5), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/70, where GetParam() = (('\x80' (-128), 0), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/86, where GetParam() = (('\x7F' (127), 1), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/88, where GetParam() = (('\x7F' (127), 5), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/94, where GetParam() = (('\x7F' (127), 7), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/98, where GetParam() = (('\x81' (-127), 0), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/108, where GetParam() = (('\x81' (-127), 7), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/114, where GetParam() = (('~' (126, 0x7E), 1), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/Int8ShiftAndRotate.UsingLoadParam/116, where GetParam() = (('~' (126, 0x7E), 5), ("bshl", 0x103a5d80))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/14, where GetParam() = (('\xFF' (255), 0), ("bushr", 0x103a5e40))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/28, where GetParam() = (('\xD6' (214), 0), ("bushr", 0x103a5e40))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/42, where GetParam() = (('\xFF' (255), 0), ("bushr", 0x103a5e40))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/56, where GetParam() = (('\xFE' (254), 0), ("bushr", 0x103a5e40))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/14, where GetParam() = ((65535, 0), ("sushr", 0x103a5e60))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/28, where GetParam() = ((65494, 0), ("sushr", 0x103a5e60))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/42, where GetParam() = ((65535, 0), ("sushr", 0x103a5e60))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/56, where GetParam() = ((65534, 0), ("sushr", 0x103a5e60))
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/22, where GetParam() = ('\xFF' (255), '\xD6' (214), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/26, where GetParam() = ('\xFF' (255), '\xFE' (254), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/40, where GetParam() = ('\xD6' (214), '\xD6' (214), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/58, where GetParam() = ('\xFF' (255), '\xD6' (214), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/62, where GetParam() = ('\xFF' (255), '\xFE' (254), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/76, where GetParam() = ('\xFE' (254), '\xD6' (214), 0x103dac40)
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/80, where GetParam() = ('\xFE' (254), '\xFE' (254), 0x103dac40)
```

The first failures do not seem to occur on the automated build. Likely due to different gcc versions that we use. It is quite odd. Either way the other 7 failures are from test cases in this PR. Going to investigate now.

fjeremic:
@mstoodle @charliegracie could we please get a re-launch of the genie-omr build?

I hopped on a Power Ubuntu box and reproduced the issues here:

```
/tmp/filip/omr/fvtest/compilertriltest/IfxcmpgeReductionTest.cpp:173: Failure
     Expected: param.oracle(param.val1, param.val2)
     Which is: 1
To be equal to: entry_point(param.val1)
     Which is: 0
[  FAILED  ] IfxcmpgeReductionTest/UInt8ReductionTest.Reduction/22, where GetParam() = ('\xFF' (255), '\xD6' (214), 0x103dac40) (8 ms)
```

The JIT was generated the following code:

```
------------------------------
 n8n      (  0)  ifbcmpeq --> block_4 BBStart at n5n ()                                               [     0x1003acdab60] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=- nc=2 flg=0x20
 n9n      (  0)    band (in GPR_0016)                                                                 [     0x1003acdaba0] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=58000 nc=2
 n10n     (  0)      i2b (in GPR_0017)                                                                [     0x1003acdabe0] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=58192 nc=1
 n11n     (  0)        iload  Parm  0<parm 0 I>[#592  Parm] [flags 0xc0000103 0x0 ] (in GPR_0017)     [     0x1003acdac20] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=58192 nc=0
 n12n     (  0)      bconst -42 (in GPR_0018)                                                         [     0x1003acdac60] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=58624 nc=0
 n13n     (  0)    bconst -42                                                                         [     0x1003acdaca0] bci=[-1,0,-] rc=0 vc=61 vn=- li=2 udi=- nc=0
------------------------------

 [     0x1003ad8e470]   0       lwz     GPR_0017, [gr1, 0]              # SymRef  Parm  0<parm 0 I>[#592  Parm] [flags 0xc0000103 0x0 ]
 [     0x1003ad8e5c0]   0       li      GPR_0018, 0xffffffffffffffd6
 [     0x1003ad8e650]   0       and     GPR_0016, GPR_0017, GPR_0018
 [     0x1003ad8e7a0]   0       cmpwi   CCR_0019, GPR_0016, -42
 [     0x1003ad8e830]   0       beq     CCR_0019, Label L0016
```

Hopping in a debugger I checked the value of `GPR_0017` is actually `0xFF`. This makes sense since my function prototype is:

```
auto entry_point = compiler.getEntryPoint<int32_t (*)(uint8_t)>();
```

Meaning the C++ compiler will do a logical sign extension to promote from `uint8_t` to it's equivalent 32-bit type which is `uint32_t` since implicit promotions of a signed type always result in a signed type. The solution here is to modify the test case to accept a signed type, so `int8_t` in this case since the JIT function definition is defined as accepting a 32-bit signed integer parameter.

The C++ compiler will then sign extend `int8_t` to it's equivalent 32-bit type which is `int32_t` and send that to the JIT function which at runtime receives a value of `0xFFFFFFFF`.

charliegracie:
genie-omr build all

jdmpapin:
Looks like the generated code is sensitive to the high bits of the argument when the conditional tree isn't supposed to be. From the perspective of the caller, the argument has type `uint8_t`, which only has 8 bits, and I don't think C++ makes any claim about the high bits of a register used to pass the value. If that's specified anywhere, it would be in the ABI

fjeremic:
> If that's specified anywhere, it would be in the ABI

Right we are talking about the ABI here. On Linux at least the arguments must come through with at least a width of 32. Here is the documentation for IBM Z:

http://refspecs.linuxbase.org/ELF/zSeries/lzsabi0_zSeries.html#PARAMETERPASSING

jdmpapin:
Sorry, my main point was that even if the ABI affects the high bits, the conditional tree above explicitly ignores those bits, so their contents should be irrelevant

fjeremic:
> the conditional tree above explicitly ignores those bits, so their contents should be irrelevant

@ymanton and I spoke about this as well. It seems we also have a discrepancy on what exactly the truncation trees such as `i2b` are semantically supposed to do. Should they clear the high order 48 bits? Should it be a no-op? On Power we seem to explicitly truncate. On IBM Z this tree is a no-op. What does x86 do?

Back to your question though - most architectures do not support comaprison operations on sub-integer types (< 32-bit that is). Both Power and IBM Z do not support them. x86 seems to. How do you evaluate a `ifbcmpeq` tree on an architecture where the only comparison instructions you can generate operate on 32-bit values in registers?

The assumption the Power and IBM Z codegens make because of these architectural restrictions is that all sub-integer types have valid 32-bit representations when loaded in a register.

mstoodle:
> The assumption the Power and IBM Z codegens make because of these architectural restrictions is that all sub-integer types have valid 32-bit representations when loaded in a register.

Conveniently inherited from the Java world :) .

Doesn't this assumption imply there is a difference between i2b and iu2bu (an opcode which doesn't currently exist and is also conveniently not needed in the Java world) ?

jdmpapin:
A byte isn't inherently signed or unsigned. Rather, the signedness is a matter of interpretation by the operations that consume it, which can disagree. So I don't think it's possible to ensure that the value of a byte-typed node is always represented as the appropriate 32-bit extension. For example, if such registers were expected to always hold the sign-extension of the byte, then they wouldn't be directly usable in unsigned comparisons. If they were expected to hold the zero-extension, they wouldn't be directly usable in signed comparisons.

x86 has narrow forms of `cmp`, e.g. `cmp al, bl` to compare the low byte of `rax` with that of `rbx`. I'm not sure that we allow the high bits to be arbitrary on x86, but for the purpose of comparison I think we should be able to.

A code generator might want to impose a convention, rather than allow the high bits to be arbitrary, but if so, then `i2b` has to establish that convention, since its child's result may have arbitrary values in the high bits. In that case `i2b` could only be treated as a no-op under special circumstances, i.e. when we happen to know that the value of the child is already a sign-extension of a byte

andrewcraik:
I believe on x86 the i2b will just cause us to use the narrower forms and there are no guarantees about the upper part of the register. A subsequent b2i or bu2i would require the appropriate sign or zero extension. On an architecture where the comparison does not go to a width less than the value and i2b is treated as a nop the 'overreading' of the register would need to be compensated for by a mask (eg an and) to clear the high bits so you only compare the byte portion of the register. The contents of the upper part just have to be guaranteed equal since the trees say you are only comparing the low byte IMO.

fjeremic:
> So I don't think it's possible to ensure that the value of a byte-typed node is always represented as the appropriate 32-bit extension.

I think this ties back into @mstoodle's remark above that things are not very clear in OMR at the moment for sub 32-bit types since the code generators are based off Java semantics where unsigned or sub 32-bit operations do not really exist.

fjeremic:
> Doesn't this assumption imply there is a difference between i2b and iu2bu (an opcode which doesn't currently exist and is also conveniently not needed in the Java world) ?

Yes I think it does. I think our evaluator for this node is incorrect. I am under the impression that we should always be clearing the unneeded bits when truncating and always sign or logical extending when promoting. Thoughts?

andrewcraik:
sub 32bit operations do not exist in the semantics of the bytecodes, but they can be generated through optimization.

Having a convention that requires maintenance of bits outside of the width specified for the type of the node is going to be very costly. In cases where operations could permute bits outside of the range (for example a badd using two registers - on a platform there may only be a two register add) you would say the badd needs to add and then clear the high bits of the register. I would argue that one only needs to care when one is using the register in a way where you will consider the high-bits in an invalid way (eg the compare). This greatly increases the scope for optimization and saves having to correct the intermediate values at every step - you just make sure that the value is limited to the correct range when being observed.

fjeremic:
> I would argue that one only needs to care when one is using the register in a way where you will consider the high-bits in an invalid way (eg the compare).

So what do you generate for the compare in such a case where your architecture only supports 32-bit comparisons for example? Would you have to sign extend both the operands before performing a `bcmpeq` comparison for example?

fjeremic:
@mstoodle all builds passed here. Can you modify the WIP: from the title? I do not own the PR so I cannot do so.

andrewcraik:
> So what do you generate for the compare in such a case where your architecture only supports 32-bit comparisons for example? Would you have to sign extend both the operands before performing a bcmpeq comparison for example?

The `bcmpeq` is not signed or unsigned since it is an equality so on a platform without a byte width compare equal we would have to ensure the upper 3 bytes (or 7 bytes if the register is 64bits) are set equal on both operands - this could be done with zero or sign extension. If it were an ordered comparison such as lt or gt then there are signed and unsigned versions of that operation since it has to interpret the bit pattern in the byte and so the appropriate extension would be required.

The key point here that has been pointed out by @jdmpapin is that signed-ness only exists for select operations (conversion, div, mulh, ordered comparison being a few - the remainder are sign agnostic eg add) and it is not possible to maintain the "correct" representation in the register because a single byte value could be treated as both signed and unsigned at different points. The code gen could try to track which representation it has generated in a register to avoid sign or zero extension later, but at the IL level that is irrelevant - the value in the tree is a single byte and there are no other bytes outside of the single byte that represents the value. Everything else is a detail for the codegens and their component evaluators.

jdmpapin:
> Doesn't this assumption imply there is a difference between i2b and iu2bu (an opcode which doesn't currently exist and is also conveniently not needed in the Java world) ?

Without reintroducing a distinction between signed and unsigned types, a hypothetical `iu2bu` would behave identically to `i2b`, with interpretation still left to consumers. A single value can be consumed by multiple parents, treated sometimes as signed and sometimes as unsigned.

> I am under the impression that we should always be clearing the unneeded bits when truncating

I'll second Andrew that it wouldn't help to maintain the high bits through intermediate operations such as `badd`.

> So what do you generate for the compare in such a case where your architecture only supports 32-bit comparisons for example? Would you have to sign extend both the operands before performing a `bcmpeq` comparison for example?

Since equality is insensitive to signedness, `bcmpeq` wouldn't require sign-extensions as such, but it would require the high bits to be determined by the low bits, in order to avoid spurious differences like the one observed above. The best way to do that might depend on the circumstance. For example, masking the operands would do, but there's no opportunity to mask negative immediates like -42 above, which appear to be sign-extended . With an immediate like -42, it would be fine to sign-extend, or simply to set all of the high bits (`GPR_0016 | 0xffffff00`).

A signed inequality would require its inputs to be sign-extended, even if the high bits of these registers were cleared. In order to avoid sign-extending for signed comparisons, byte values would have to be sign-extended on truncation and also after any intermediate calculations

mstoodle:
> Without reintroducing a distinction between signed and unsigned types, a hypothetical iu2bu would behave identically to i2b, with interpretation still left to consumers.

"Doesn't this assumption imply" is how I led :) . The code generators assume that upper bits of short types are affected by the conversion operation, which is what make the hypothesized iu2bu opcode different than i2b. I don't think we should have such an opcode, for the same reasons you and @andrewcraik raised.

If a platform wants to store bytes or shorts or ints in a container larger than a byte/short/int, that's up to it to do properly. I don't think it makes sense to expect the optimizer to imagine and somehow manage bits that aren't actually specified in the IL.

jdmpapin:
By the way, was this tree originally `ifbcmpge`? i.e.

    ifbcmpge --> block_4
      band
        i2b
          iload  <parm 0 I>
        bconst -42
      bconst -42

If so, then I think this is one of the cases that was supposed to have been prevented by checking that either the comparison is unsigned or the constant is non-negative. Under signed 8-bit comparison we have 

    (0x02 & 0xd6) == 0x02 > 0xd6

mstoodle:
@jdmpapin are you suggesting there might still be something wrong here?

fjeremic:
> By the way, was this tree originally `ifbcmpge`? i.e.

No the original tree was an unsigned compare, which is why we were able to carry out the transformation:

```
n8n       ifbucmpeq --> block_4 BBStart at n5n ()                                              [     0x1002e2fab40] bci=[-1,0,-] rc=0 vc=8 vn=- li=- udi=- nc=2 flg=0x20
n9n         band                                                                              [     0x1002e2fab80] bci=[-1,0,-] rc=1 vc=8 vn=- li=- udi=- nc=2
n10n          i2b                                                                             [     0x1002e2fabc0] bci=[-1,0,-] rc=1 vc=8 vn=- li=- udi=- nc=1
n11n            iload  Parm  0<parm 0 I>[#592  Parm] [flags 0x40000103 0x0 ]                  [     0x1002e2fac00] bci=[-1,0,-] rc=1 vc=8 vn=- li=- udi=- nc=0
n12n          bconst -42                                                                      [     0x1002e2fac40] bci=[-1,0,-] rc=1 vc=8 vn=- li=- udi=- nc=0
n13n        bconst -42                                                                        [     0x1002e2fac80] bci=[-1,0,-] rc=1 vc=8 vn=- li=- udi=- nc=0
```

jdmpapin:
Thanks, I have no issues then 😄

mstoodle:
Thanks for all the discussion and hard work to get this one done @last5bits @fjeremic @jdmpapin @Leonardo2718 !

mgaudet:
I was just catching up on this discussion. Am I correct in summarizing, that as far as concrete issues to be opened as derived from this (interesting!) discussion, we have 

-  [ ] Create Tree (Tril?) level tests for `i2s` and `i2b`, in particular with a focus on ensuring high bits don't affect computation.  

Is that the only one, or would people in this thread call out others as important?

Leonardo2718:
@mgaudet I think that's a good summary.

As an add-on, I'd like to have a discussion to refine and get a consensus on what code generators are expected to do/handle for these opcodes (which we can then be verify/enforce using tests).

PR #1559, opened 2017-08-23
Mark locationAddr volatile in _patchVirtualGuard

fjeremic:
Stores to locationAddr in _patchVirtualGuard should not be reordered by
the compiler as they could be potentially patching asynchronously with
the application executing the patch location. Marking this parameter as
volatile ensures this does not happen.

### DIFF
--- a/compiler/z/runtime/VirtualGuardRuntime.cpp
+++ b/compiler/z/runtime/VirtualGuardRuntime.cpp
@@ -44,14 +44,16 @@
   */
 extern "C" void _patchVirtualGuard(uint8_t* locationAddr, uint8_t* destinationAddr, int32_t smpFlag)
    {
+   volatile uint8_t* volatileLocationAddr = locationAddr;
+
    static bool debugTrace = debug("traceVGNOP") != NULL;
 
    if (debugTrace)
       {
-      printf("####> Patching VGNOP at locationAddr %p (%x), destinationAddr %p (%x), smpFlag: %d\n", locationAddr, *reinterpret_cast<intptrj_t*>(locationAddr), destinationAddr, *reinterpret_cast<intptrj_t*>(destinationAddr), smpFlag);
+      printf("####> Patching VGNOP at locationAddr %p (%x), destinationAddr %p (%x), smpFlag: %d\n", volatileLocationAddr, *reinterpret_cast<volatile intptrj_t*>(volatileLocationAddr), destinationAddr, *reinterpret_cast<intptrj_t*>(destinationAddr), smpFlag);
       }
 
-   int64_t displacement = static_cast<int64_t>(destinationAddr - locationAddr) / 2;
+   int64_t displacement = static_cast<int64_t>(destinationAddr - volatileLocationAddr) / 2;
 
    // If there is a NOP BRC / BRCL at locationAddr the application could be asynchronously executing this branch as
    // we are patching it. Because the caller / codegen ensured this branch will never execute if it is to be patched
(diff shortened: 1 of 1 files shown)


fjeremic:
This is inspired by the suggestion by @mgaudet in https://github.com/eclipse/omr/pull/1550#issuecomment-324128810.

fjeremic:
@0dvictor @ymanton could I please get a review?

0dvictor:
The changeset is simple and looks correct. LGTM.

ymanton:
It's news to me that virtual guards can be patched while the application is running. Is this function being called for something other than patching a (traditional) virtual guard?

The general idea looks fine to me, but maybe you're better off hiding this in the implementation(s) that need it, rather than in the signature for everyone -- e.g. it doesn't look ARM needs it since it only writes a single (indivisible) value at the location, Power wouldn't need it if it shared the signature because it too would be limited to writing a single value.

fjeremic:
> It's news to me that virtual guards can be patched while the application is running. Is this function being called for something other than patching a (traditional) virtual guard?

As far as I can see this is the only patching function we have. Some virtual guards are for stop the world events, others are not:
https://github.com/eclipse/omr/blob/master/compiler/z/codegen/S390Instruction.cpp#L5792-L5804

 This function prototype is shared across all platforms and hence I assume the patching can be done asynchronously. 

This is why x86 for example patches in an infinite loop to avoid the race condition:
https://github.com/eclipse/omr/blob/master/compiler/x/runtime/VirtualGuardRuntime.cpp#L44-L45

ymanton:
Actually you're right, I forgot all of the contexts in which downstream called this function.

Anyhow, I would still suggest not putting this in the signature unless it has a purpose for everyone, since it's an implementation detail only relevant to some platforms.

fjeremic:
> Anyhow, I would still suggest not putting this in the signature unless it has a purpose for everyone, since it's an implementation detail only relevant to some platforms.

Can you suggest an alternative implementation? Will something like:

```
+extern "C" void _patchVirtualGuard(uint8_t *locationAddr, uint8_t *destinationAddr, int32_t smpFlag)
    {
        volatile uint8_t* volatileLocationAddr = locationAddr;
        // Use volatileLocationAddr instead of locationAddr in this function
    }
```

work as intended? This would need to be replicated in all platforms, so not sure what the benefit of this approach is. `locationAddr` *is* a volatile memory location. I don't quite understand why making this explicit in the signature is an implementation detail. To me it is a specification detail informing developers implementing the API that other threads can observe stores to the said memory location and such stores should be ordered.

ymanton:
> Can you suggest an alternative implementation? Will something like: ... This would need to be replicated in all platforms, so not sure what the benefit of this approach is.

Yeah that looks fine to me, although like I said it's not needed at all on ARM the way it's currently implemented, and on x86 it's not needed on the path that does the patch with a single write.

> locationAddr is a volatile memory location. I don't quite understand why making this explicit in the signature is an implementation detail. To me it is a specification detail informing developers implementing the API that other threads can observe stores to the said memory location (and vice-versa) and any stores to locationAddr should be ordered, irrelevant of whether you only need one store or not.

That's not an appropriate use of `volatile` though, a function with a better name or a comment will accomplish that more effectively; `volatile` should only be used by the implementer if and when they know they need it.

What volatile does is compiler dependent, which is why the implementer should decide to use it or not. It has nothing to do with threads or making things visible to other threads, it's not like Java's `volatile` which is better defined and provides both compiler and hardware ordering guarantees for multi-threaded code. C/C++ `volatile` tells the compiler that the memory holding the volatile variable can be written to without it knowing and/or has access side effects, which usually **prevents** it from doing things like optimizing away multiple redundant reads, re-ordering accesses to it (though it doesn't prevent the hardware from re-ordering), combining multiple smaller reads/writes into one bigger read/write, and so on. On MSVC it actually does more by default unless you disable this MS-specific behaviour, but that goes to my point that the implementer needs to know what it's doing on their platform and if it's doing what they want.

If you perform your patching by writing to parts of an instruction in a specific order and in such a way that after each write the instruction is still valid then you want the compiler to generate stores for your writes in exactly the order you wrote them and you also don't want the compiler to combine a sequence of single byte writes into a larger write, so it makes sense to use `volatile` in that context to prevent the compiler from being too smart. The more correct way to make a compiler respect the order and composition of your writes to memory is to use a compiler barrier between them, but not every compiler gives you those, so people use `volatile` to get the same effect. If you perform your patching via a single write `volatile` is completely useless, if not harmful.

> However do we not have to issue memory fences after the writes on some systems? How would you know you have to issue a fence if locationAddr was not marked as volatile?

On most architectures you have to do something to flush caches to make sure processors don't keep executing the old unpatched instruction from their icache (which is what `armCodeSync`, `patchingFence16`, etc are mostly concerned with), but that's totally separate from `volatile`. `volatile` is not documenting what's going on or what's needed to the implementer though, it actually just makes the code more mysterious because now the implementer has to wonder why the memory pointed to is `volatile` since, like I said, the meaning in C/C++ is not defined with threading and visibility in mind.

fjeremic:
@ymanton thank you for investing the time to write up your explanation. It is very much appreciated. I fully understand your perspective now and appreciate your feedback. I have learned some facts about `volatile` in C which I had incorrectly assumed translated over from Java. Given that on IBM Z we store a byte at a time I have taken the suggestion to locally mark `locationAddr` as volatile on the IBM Z implementation only in dfe2c12

ymanton:
That looks good to me, it should prevent the compiler from doing anything clever to your sequence of writes.

PR #1560, opened 2017-08-23
Fix compiler warning in sigtest on Linux 390

charliegracie:
When compiling Linux 390 the size of sa_flags changes based which version of the C library you are using.  This causes some ugly code so instead I just always store the value into an int64_t and print that. This way there is not weird checks to decide the printf specifier.


youngar:
The current plan is to cast the integer to the right size, and then print it.  This should let us get rid of the `#if defined(...)` code.

PR #1562, opened 2017-08-23
Stash use node to tree mapping in CopyProp

ncough:
The treetop containing a use node is often required
in CopyProp. The prior approach was to repeatedly
search the CFG for the corresponding treetop, however,
this was costly. Instead, a mapping is generated
early in copy prop and used throughout.

### DIFF
--- a/compiler/optimizer/CopyPropagation.cpp
+++ b/compiler/optimizer/CopyPropagation.cpp
@@ -64,7 +64,8 @@ void collectNodesForIsCorrectChecks(TR::Node * n, TR::list< TR::Node *> & checkN
 
 TR_CopyPropagation::TR_CopyPropagation(TR::OptimizationManager *manager)
    : TR::Optimization(manager),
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
Need a review from @andrewcraik

andrewcraik:
LGTM

vijaysun-omr:
Is the reason this solution works is because the key used in the new map you added is the node itself (value being the treetop) and transformations in copy propagation modify the node to have a different sym ref but never in fact change the tree that a particular node belongs to ?

ncough:
Correct, the only significant optimizations that occur before the final use of the new map are `rematerializeIndirectLoadsFromAutos` and `replaceCopySymbolReferenceByOriginalIn`. Neither of these will move nodes to new trees. They may create or remove nodes, however, use nodes are fetched from UseDefInfo, which is not updated. As node pointer equality is used in the original implementation and the new, only those initial nodes that existing during UseDef analysis require mappings.

Issue #1530, opened 2017-08-18
OS/X CMake builds are producing lots of pthread warnings


mgaudet:
```
[725/1106] Linking CXX shared library omrsigcompat/libomrsig.dylib
clang: warning: argument unused during compilation: '-pthread'
clang: warning: argument unused during compilation: '-pthread'
[734/1106] Linking CXX executable fvtest/algotest/omralgotest
clang: warning: argument unused during compilation: '-pthread'
clang: warning: argument unused during compilation: '-pthread'
[742/1106] Linking CXX executable fvtest/gctest/omrgctest
clang: warning: argument unused during compilation: '-pthread'
clang: warning: argument unused during compilation: '-pthread'
[772/1106] Linking CXX executable fvtest/porttest/omrporttest
```

(On [master](https://travis-ci.org/eclipse/omr/jobs/265789735#L3324))

PR #1563, opened 2017-08-23
Fix up compile options

rwy0717:
## Clang and `-pthread`

With clang, `-pthread` is a compile-only option, It can't be used when linking. We've been using -pthread to link on our osx builds, and generating a lot of warning (#1530).

We were adding compile options to `CMAKE_CXX_FLAGS`, but it's used when linking CXX objects. Instead, this PR uses `add_compile_options`. Now, these options will only appear on the compile command.

This PR also adds `-pthread` to the `OMR_PLATFORM_LINKER_OPTIONS` when using GCC. For Clang, add `pthread` to `OMR_PLATFORM_LIBRARIES`.

## The MSVC Runtime and OMRSig

(copied from below)

In MSVC, the `/Md[d]` option sets what version of the MSVC RT DLL is used, (the little d means "with debug symbols"). Since I've updated how platform compile options are added, `/MD` (added by `platform/toolcfg/msvc.cmake`) has propagated to new places.

However, OMR Sig only supports finding the `signal` symbol in the *debug* MSVC RT DLL, which is linked with `/MDd`. The new `/MD` flag is overriding the `/MDd` flag, and breaking us.

I've updated OMR Sig to support multiple versions of the MSVC RT DLL. Now, OMR Sig will check for a number of different MSVC RT DLL versions (with/without debug info), at run time. This means you can link OMR sig against a variety of MSVC RT DLLs, and it should Just Work™ in more situations.

There is another detail: cmake already adds the `/MD[d]` automatically, and picks the version using CMAKE_BUILD_TYPE. In the interest of simplicity, I've removed `/MD` from our msvc toolcfg.


rwy0717:
In MSVC, the `/Md[d]` option sets what version of the MSVC RT DLL is used, (the little d means "with debug symbols"). Since I've updated how platform compile options are added, `/MD` (added by `platform/toolcfg/msvc.cmake`) has propagated to new places.

However, OMR Sig only supports finding the `signal` symbol in the *debug* MSVC RT DLL, which is linked with `/MDd`. The new `/MD` flag is overriding the `/MDd` flag, and breaking us.

I've updated OMR Sig to support multiple versions of the MSVC RT DLL. Now, OMR Sig will check for a number of different MSVC RT DLL versions (with/without debug info), at run time. This means you can link OMR sig against a variety of MSVC RT DLLs, and it should Just Work™ in more situations.

There is another detail: cmake already adds the `/MD[d]` automatically, and picks the version using CMAKE_BUILD_TYPE. In the interest of simplicity, I've removed `/MD` from our msvc toolcfg.

rwy0717:
This is good to go.

mgaudet:
Looks good to me. `tril` will have to have its pthread issues fixed separately.

omrsigcompat/omrsig.cpp
```diff
@@ -276,6 +277,34 @@ sigaction(int signum, const struct sigaction *act, struct sigaction *oldact) __T
 }
 #endif /* defined(POSIX_SIGNAL) */
 
+#if defined(WIN32)
+HMODULE
```
charliegracie: Can we not figure out at runtime what the current MSVC runtime module is?

rwy0717: When omrsig uses a different version of the MSVCR than the application, the whole thing is borked anyways. If we want, I could determine the version of MSVCR in cmake, and we can load that specific version here. That would essentially move this check to compile-time. Sound okay?

rwy0717:
I've updated this PR and reverted the runtime search for an MSVC C Runtime (MSVC CRT). omrsig is back to normal, and will look for a specific version of the MSVC CRT determined by cmake.

I've expanded the cmake MSVC CRT guessing code and moved it to `OmrMsvcRuntime`. The `OMR_MSVC_CRT` is set using the `MSVC_VERSION` and the `CMAKE_BUILD_TYPE`. If someone is using custom flags to set the runtime, this could guess the wrong runtime. `OmrMsvcRuntime` respects the `MSVC_CRT` environment variable, and is meant to be user configurable.

cmake/modules/OmrPlatform.cmake
```diff
@@ -56,13 +56,13 @@ macro(omr_platform_global_setup)
 
 	set(OMR_PLATFORM_GLOBALLY_INITIALIZED 1)
 
+	add_compile_options(${OMR_PLATFORM_COMPILE_OPTIONS})
```
charliegracie: This will also add these flags to the assembler line.  On some platforms that will not work.

rwy0717:
The VS generator does not use CMAKE_BUILD_TYPE, the solution it generates provides multiple configurations simultaneously instead. To properly support multiple build configurations, `OMR_MSVC_RT` has to use generator expressions to detect the build type.

This new version of the patch only supports Debug or Release builds.

cmake/modules/OmrPlatform.cmake
charliegracie: @rwy0717 I believe this will break PPC LE.

charliegracie:
This needs to be rebased before it can be merged

rwy0717:
I've moved the msvc support in omrsig to a new PR: #1616 .

PR #1565, opened 2017-08-24
Make ILvalidator handle tstart correctly

### LINKED ISSUE #1553: ILValidator doesn't understand `tstart` branches
If you run the JitBuilder `transactionaloperations` example, you'll get an ILValidation error. 

It appears the validation error is itself an error however: 

```
14:56 $ TR_Options=traceFull,log=0log,traceIlGen ./jitbuilder/release/transactionaloperations 
Step 1: initialize JIT
Step 2: define type dictionary
Step 3: compile method builder
TransactionalMethod::buildIL() running!
*** VALIDATION ERROR ***
Node: branch n19n
Method: ../jitbuilder/release/src/TransactionalOperations.cpp:77:transaction_increment
Child count 0 matches neither expected values of 1 (without GlRegDeps) nor 2 (with GlRegDeps)
Assertion failed: (0), function trap, file ../compiler/infra/Assert.cpp, line 69.
Abort trap: 6
```

Looking at the generated log, you'll see 

```
n3n       BBStart <block_3>                                                     
n22n      tstart  unknown helper[#180  helper Method] [flags 0x400 0x0 ] ()     
n19n        branch --> block_5 BBStart at n7n                                   
n20n        branch --> block_7 BBStart at n11n                                  
n21n        branch --> block_9 BBStart at n15n                                  
n4n       BBEnd </block_3> =====                                                
```

Which I believe is the valid shape of a `tstart` node.

Leonardo2718:
These commits allow the ILValidator to correctly check `branch` opcodes.

Fixes #1553

### DIFF
--- a/compiler/il/ILOpCodeProperties.hpp
+++ b/compiler/il/ILOpCodeProperties.hpp
@@ -8658,7 +8658,7 @@
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::NoType,
    /* .typeProperties       = */ 0,
-   /* .childProperties      = */ ILChildProp::NoChildren,
+   /* .childProperties      = */ THREE_SAME_CHILD(TR::NoType),
    /* .swapChildrenOpCode   = */ TR::BadILOp,
    /* .reverseBranchOpCode  = */ TR::BadILOp,
    /* .booleanCompareOpCode = */ TR::BadILOp,
--- a/compiler/il/ILOpCodeProperties.hpp
+++ b/compiler/il/ILOpCodeProperties.hpp
@@ -10416,7 +10416,7 @@
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::NoType,
    /* .typeProperties       = */ 0,
-   /* .childProperties      = */ ONE_CHILD(ILChildProp::UnspecifiedChildType),
+   /* .childProperties      = */ ILChildProp::NoChildren,
    /* .swapChildrenOpCode   = */ TR::BadILOp,
    /* .reverseBranchOpCode  = */ TR::BadILOp,
    /* .booleanCompareOpCode = */ TR::BadILOp,
--- a/compiler/il/OMRILOpCodesEnum.hpp
+++ b/compiler/il/OMRILOpCodesEnum.hpp
@@ -725,7 +725,7 @@
    dexp,     // double exponent
 
 
-   branch,     // generic branch --> first child is a constant indicating the mask
+   branch,     // generic branch --> DEPRECATED use TR::case instead
    igoto,      // indirect goto, branches to the address specified by a child
 
    bexp,     // signed byte exponent  (raise signed byte to power)


compiler/ras/ILValidator.cpp
```diff
@@ -219,6 +220,14 @@ bool TR::ILValidator::treesAreValid(TR::TreeTop *start, TR::TreeTop *stop)
          if (isEndOfExtendedBlock)
             validateEndOfExtendedBlock(iter);
          }
+      else if (opcodeValue == TR::branch &&
+               iter.currentParent() != NULL &&
+               iter.currentParent()->getOpCodeValue() == tstart)
```
mstoodle: I know the opcode properties table says that TR::branch has a single child, but I don't see any code anywhere that creates or treats a TR::branch node that has anything other than zero children.

So that makes me wonder if the TR::branch child property is wrong, as well as if this correctness condition is therefore more complicated than it needs to be. @magaudet @cathyzhyi can you shed light on TR::branch?

Leonardo2718: The child property in the table is derived from a comment in `omr/compiler/il/OMRILOpCodesEnum.hpp:726`:

```c++
   branch,     // generic branch --> first child is a constant indicating the mask
```

If this is indeed incorrect, I'd be more than happy to simplify the validator and fix the child property (and the comment). Also, a follow up question from me (perhaps better palced in a separate issue), if TR::branch doesn't expect a child, what's the difference between TR::branch and TR::Goto?

mstoodle: Botched Matthew's @ reference...@mgaudet do you happen to know the history of the TR::branch opcode ?

mstoodle:
discussed with @mgaudet a bit. The TR::branch opcode dates back quite a way and my guess is that the original use has changed and the first child requirement has been removed. Further, @mgaudet believes we should just be using TR::Case rather than TR::branch and I agree that seems cleaner.  I've asked him to open an issue describing that change, which can go onto our backlog.

To address this problem, I think we should just update the description of TR::branch to say it has zero children, which should simplify the condition being checked.

compiler/ras/ILValidator.cpp
mstoodle: @Leonardo2718 TR::Goto is a treetop, for one. TR::Case is a much more appropriate relationship.

Leonardo2718:
@mstoodle @mgaudet 

Now that I've had a few minutes to spare, I've updated this PR to correct the expected children of both `TR::branch` and `TR::tstart`.

I've also update the `TR::branch` documentation to say the opcode is deprecated (in light of #1847).

mstoodle:
LGTM. Once CI tests pass and we get a thumbs up from @mgaudet, this one can be merged.

mgaudet:
LGTM as well.

mgaudet:
@mstoodle Ping :) I think this is G2G

mstoodle:
for thoroughness: genie-omr build all

PR #1569, opened 2017-08-24
Fix VirtualGuardHeadMerger to uncommon correctly

ncough:
VirtualGuardHeadMerger will rearrange privarg stores
and monitor stores between virtual guards. However,
it did not correctly uncommon nodes between these
two types of stores, resulting in loads of
uninitialized values.

### DIFF
--- a/compiler/optimizer/VirtualGuardHeadMerger.cpp
+++ b/compiler/optimizer/VirtualGuardHeadMerger.cpp
@@ -37,6 +37,7 @@
 #include "optimizer/Optimizer.hpp"             // for Optimizer
 #include "optimizer/TransformUtil.hpp"         // for TransformUtil
 #include "ras/DebugCounter.hpp"
+#include "infra/Checklist.hpp"
 
 #define OPT_DETAILS "O^O VG HEAD MERGE: "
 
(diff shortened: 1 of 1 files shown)


compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -46,33 +47,96 @@ static bool isMergeableGuard(TR::Node *node)
...
       }
-   if (hasMonitorStore && firstPrivArg)
-      return block->split(firstPrivArg, cfg, true /* fixupCommoning */, false /* copyExceptionSuccessors */);
-   else if (firstPrivArg)
-      return block;
-   else
-      // This block does not contain a priv arg
-      return NULL;
+
+   // If there are privargs and monitors, they must be swapped around, such that all privargs are
```
andrewcraik: Is there a risk of a privarg load under a monitor store or is that not the case by construction? I suspect that because the code is generated in the caller's context the privarg won't be used, but wanted to ask the question to confirm the tree reordering isn't likely to introduce the same problem.

ncough: As the monitor stores are before the privarg stores, this will not be the case. It is also not possible to have the inverse, as monitor meta data should not be loaded as arguments, as indicated by a comment in the prior implementation.

It should also be noted that volatile stores are not considered here, as this optimization will see their loads under separate treetops and give in.

vijaysun-omr:
Is this ready ? It has the WIP prefix..so I thought I would check.

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -46,33 +47,96 @@ static bool isMergeableGuard(TR::Node *node)
...
+   TR_ASSERT(isMergeableGuard(guard->getNode()), "last node must be guard %p", guard->getNode());
+
+   // Search for privarg and monitor stores, anchoring any commoned loads at the start of the block
+   bool anchoredTemps = false;
+   for (TR::TreeTop *tt = start; tt && tt->getNode()->getOpCodeValue() != TR::BBEnd; tt = tt->getNextTreeTop())
       {
       TR::Node * node = tt->getNode();
+
+      anchoredTemps |= anchorCommonNodes(comp, node, start, checklist);
+
```
vijaysun-omr: Is it safe to anchor the nodes at the start of the block ? Can we be sure that none of those nodes would evaluate to a different value if moved to the start of the block ? i.e. a def to the node in between.

andrewcraik: The only treetops allowed by VGHM between guards are are privarg stores and monitored object stores. Unless a privarg store was loaded as part of calculating a subsequent privarg or the monitored object we loaded to be used as part of one of the nodes then it shouldn't be a problem. To be fully safe one could have a BitVector of alias indices we have seen defined and abort if we see an intersection with a node we are anchoring. That being said I would be surprised if that happened given the highly constrained case that is permitted. Thoughts?

ncough: Anchoring all commoned nodes was unnecessary here. Only nodes commoned between the monitor stores and guard must be specially managed, as the guard will be shifted before the stores. Therefore, anchoring any commoned nodes within the guard's treetop into the privarg block is sufficient. Also, it is already assumed that the guard does not load any of the monitor stores, due to the prior mentioned shift of the guard, so anchoring nodes before the defs must be safe.

PR #1570, opened 2017-08-24
PPC fixes

mstoodle:
First, resolve a POWER System Linkage "accounting" bug in the parameter slot assignments when double parameters are used.

Second, enable repeated `initializeJit()`, `shutdownJit()` cycles on POWER, which are used in the JitBuilderTests.

### DIFF
--- a/compiler/p/codegen/PPCSystemLinkage.cpp
+++ b/compiler/p/codegen/PPCSystemLinkage.cpp
@@ -1717,7 +1717,6 @@ void TR::PPCSystemLinkage::setParameterLinkageRegisterIndex(TR::ResolvedMethodSy
                }
             numFloatArgs++;
             numIntArgs++;
-            if (paramCursor->getDataType() == TR::Double) numIntArgs++;
             break;
          case TR::Aggregate:
             {
--- a/compiler/control/CompileMethod.cpp
+++ b/compiler/control/CompileMethod.cpp
@@ -136,7 +136,25 @@ int32_t commonJitInit(OMR::FrontEnd &fe, char *cmdLineOptions)
 
    void *pseudoTOC = NULL;
 #if defined(TR_TARGET_POWER)
-   pseudoTOC = (void *) TR_PPCTableOfConstants::initTOC(&fe, fe.getPersistentInfo(), jitConfig->getInterpreterTOC());
+
+   static bool initializedTOC = false;
+   // some silliness  here needed to support projects like JitBuilder where the JIT
+   // can be initialized, shutdown, then initialized again. If we're initializing the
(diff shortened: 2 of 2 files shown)


mstoodle:
Would you please review at least the system linkage change, @ymanton ?

mstoodle:
For the record, these two changes enable our check-in tests to run and pass on PPCLE.

compiler/p/codegen/PPCSystemLinkage.cpp
```diff
@@ -1717,7 +1717,8 @@ void TR::PPCSystemLinkage::setParameterLinkageRegisterIndex(TR::ResolvedMethodSy
                }
             numFloatArgs++;
             numIntArgs++;
-            if (paramCursor->getDataType() == TR::Double) numIntArgs++;
+            if (!TR::Compiler->target.is64Bit() && paramCursor->getDataType() == TR::Double)
+               numIntArgs++;
```
vijaysun-omr: Naive question. Why are numIntArgs being incremented when this is a double (and numFloatArgs has already been incremented) ?

mstoodle: The ABI wants what it wants. I found it odd, too.

But incrementing it twice for a double on a 64-bit platform was even more odd ;) .

ymanton: This is correct for ppc64le, floating point args passed in registers cause the corresponding integer registers to be 'burned'.

Having said that you don't actually have to do anything for 32-bit targets because LE is 64-bit only (I realize there's another `is32Bit()` check in this code, it too is unnecessary and was probably copied from some other Power linkage that has both 32- and 64-bit targets).

mstoodle: Interesting; I hadn't realized this function was PPCLE specific (there are lots of references to AIX and 32-bit in other spots in this class). I suppose you're recommending that I clear out the 32-bit code from this path then, @ymanton :) ?

ymanton: Yeah, the

```
if (paramCursor->getDataType() == TR::Double) numIntArgs++;
```

line definitely needs to go to make it correct, but the two added lines aren't necessary. Otherwise, the changes in this PR look fine to me. This class can be refactored properly later if/when it needs to support other platforms.

mstoodle: Ok, I updated the commit to simply remove the second numIntArgs bump.

Thanks @ymanton !

vijaysun-omr:
This looks good to me. @ymanton please let me know if you are okay with it and if so, I will merge it.

ymanton:
LGTM.

PR #1572, opened 2017-08-24
Prevent LoopVersioner from checking guards

ncough:
LoopVersioner will search for profiled nodes,
to build versions of loops where the profiling
is disabled. If a node beneath a nopable guard
shares BCI with a profiled value, LoopVersioner may
mistake the guard's child as being profiled. This
change prevents the method from analysing the children
for a nopable guard.


andrewcraik:
the basic principle seems sound - the risk is of perf loss, but the contents of the trees being skipped are not worthy of versioning as described.

vijaysun-omr:
Is this ready ? It is marked WIP...

ncough:
Corrected mistakes in the commit message.

vijaysun-omr:
There are some typos in the commit message. Please use the word "guard" instead of "VG".

PR #1575, opened 2017-08-25
Correct the jit builder test platform coverage

charliegracie:
OSX and Linux x86 are the only platforms which can run testall. The
conditional for the test makefile was checking the wrong values. This
change fixes the conditionals so the correct tests run on each platform.


mstoodle:
thanks, charlie, i found that one just recently but hadn't pushed a fix yet :(

mstoodle:
will merge once the PR testing inevitably proves that testall still runs on linux_x86 and osx :)

mstoodle:
Ok, so this PR fixes the fvtest/omrtest.mk file, but I don't think the cmake based builds (OSX is now only tested via cmake) include the 'testall' set of samples in the jitbuilder testing. Nonetheless, I believe this PR is valid, so I'm going to merge.

Issue #1574, opened 2017-08-25
LibDwarf path isn't persisting properly in CMake


mgaudet:
If you configure OMR using CMake, it will work, and successfully build DDR. 

However, if you then touch an arbitrary CMake file, such that CMake self-reconfigures, it then will fail to build omr, as it no longer is able to find libdwarf.

```
Processing tdf file /home/magaudet/open/omr/util/hashtable/hashtable.tdf
Calculating groups for hashtable
Creating header file: /home/magaudet/open/omr/build/util/hashtable/ut_hashtable.h
Creating pdat file: /home/magaudet/open/omr/build/util/hashtable/ut_hashtable.pdat
Creating c file: /home/magaudet/open/omr/build/util/hashtable/ut_hashtable.c
[166/166] Linking CXX executable ddr/tools/ddrgen/omr_ddrgen
magaudet@ubuntu1604VM:~/open/omr/build (detached*)$ touch ../cmake/compiler_support.cmake 
magaudet@ubuntu1604VM:~/open/omr/build (detached*)$ ninja omr_ddrgen
[1/1] Re-running CMake...
-- Starting with CMake version 3.5.1
CMake Warning at CMakeLists.txt:24 (message):
  CMake support is currently experimental.
<snip> 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/magaudet/open/omr/build
[3/10] Building CXX object ddr/lib/ddr-scanner/CMakeFiles/omr_ddr_scanner.dir/dwarf/DwarfScanner.cpp.o
[...]

PR #1585, opened 2017-08-29
cmake: Fix up our FindLibDwarf module

rwy0717:
This PR beefs up the FindLibDwarf package and starts requiring it in linux builds.

Now, FindLibDwarf should work correctly in subsequent reruns.


rwy0717:
This fixes #1574

PR #1586, opened 2017-08-30
cmake: Support tools cross compilation

rwy0717:
This patch lets us compile and use the tools when cross compiling for
x86 on an x86-64 machine.

I've simplified the tools directory, moving the add_tracegen and
add_hookgen macros to seperate modules. This lets them be used outside
of OMR more easily. Now, we can completely skip the tools directory when
importing.

Do not skip the tools directory when cross compiling. Instead, emit a
message that the tools will be cross compiled.

The OMR_TOOLS flag is now respected. If you have the hookgen or tracegen
tools installed on your system, you should be able to disable building
the bundled tools completely. Imported tools will always be used,
regardless of the OMR_TOOLS state.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,9 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules" ${CMAKE_MODULE
 
 include(OmrAssert)
 include(OmrFindFiles)
+include(OmrHookgen)
 include(OmrPlatform)
+include(OmrTracegen)
 
 ###
 ### Set up the global platform configuration
@@ -95,10 +97,12 @@ endif()
 configure_file(./omrcfg.CMakeTemplate.h omrcfg.h)
 configure_file(./omrversionstrings.CMakeTemplate.h omrversionstrings.h)
 
-add_subdirectory(tools)
-# Yeah, its dumb doing this here. Read note in tools/CMakeLists.txt
-if(CMAKE_CROSSCOMPILING)
-	include(${OMR_TOOLS_IMPORTFILE})
+if(OMR_TOOLS_IMPORTFILE)
+	include("${OMR_TOOLS_IMPORTFILE}")
+elseif(OMR_TOOLS)
+	add_subdirectory(tools)
+else()
+	message(FATAL_ERROR "OMR: Build tools are required. See OMR_TOOLS and OMR_TOOLS_IMPORTFILE")
 endif()
 
 add_subdirectory(thread)
--- a/cmake/modules/OmrHookgen.cmake
+++ b/cmake/modules/OmrHookgen.cmake
@@ -1,6 +1,6 @@
 ###############################################################################
 #
-# (c) Copyright IBM Corp. 2017, 2017
+# (c) Copyright IBM Corp. 2017
 #
 #  This program and the accompanying materials are made available
 #  under the terms of the Eclipse Public License v1.0 and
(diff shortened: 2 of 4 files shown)


charliegracie:
@dnakamura can you review this please?

@rwy0717 you will need to rebase this change before it can be merged.  Thanks

cmake/modules/OmrHookgen.cmake
```diff
@@ -0,0 +1,40 @@
...
+	return()
+endif()
+set(OMR_HOOKGEN_ 1)
+
+#TODO: currently output in source tree, should be in build tree
+#TODO: Dependecy checking is broken, since it checks for output in build tree rather than src
+function(omr_add_hookgen input)
+	get_filename_component(input_dir "${input}" DIRECTORY)
+	#message("Hookgen, Output = ${ARGN}")
+	#message("Hookgen, depends = ${CMAKE_CURRENT_SOURCE_DIR}/${input}")
```
dnakamura: While we are working on this may as well strip out these debug  messages

cmake/modules/OmrTracegen.cmake
```diff
@@ -0,0 +1,47 @@
...
+		set(base_name "${ARGV1}")
+	else()
+		get_filename_component(base_name "${input}" NAME_WE)
+	endif()
+	#file(TO_CMAKE_PATH "${CMAKE_CURRENT_BINARY_DIR}/${input_dir}/ut_${base_name}" generated_filename)
+	file(TO_CMAKE_PATH "${CMAKE_CURRENT_BINARY_DIR}/ut_${base_name}" generated_filename)
+
+	add_custom_command(
+		OUTPUT "${generated_filename}.c" "${generated_filename}.h" "${generated_filename}.pdat"
+		COMMAND $<TARGET_FILE:tracegen> -w2cd -treatWarningAsError -generatecfiles -threshold 1 -file ${CMAKE_CURRENT_SOURCE_DIR}/${input}
```
dnakamura: '$<TARGET_FILE:...>` wont work if we are expecting to be pulling in tracegen from the path. Substituting with just `tracegen`  should do the trick

tools/CMakeLists.txt
```diff
@@ -16,61 +16,17 @@
...
 
-	add_subdirectory(hookgen)
-	add_subdirectory(tracemerge)
-	add_subdirectory(tracegen)
+omr_assert(TEST OMR_TOOLS)
+omr_assert(TEST NOT OMR_TOOLS_IMPORTFILE)
 
-	export(TARGETS hookgen tracemerge tracegen FILE "ImportTools.cmake")
+if(CMAKE_CROSSCOMPILING)
+	message(Status "OMR: Using cross compiled tools")
```
dnakamura: Maybe a bit better message is in order. Had to do a double take on this one.  Not entirely sure how to make it better.  Maybe I just need more coffee

CMakeLists.txt
```diff
@@ -110,10 +112,10 @@ add_dependencies(omrglue
...
 )
 
-add_subdirectory(tools)
-# Yeah, its dumb doing this here. Read note in tools/CMakeLists.txt
-if(CMAKE_CROSSCOMPILING)
-	include(${OMR_TOOLS_IMPORTFILE})
+if(OMR_TOOLS_IMPORTFILE)
+	include("${OMR_TOOLS_IMPORTFILE}")
+elseif(OMR_TOOLS)
+	add_subdirectory(tools)
```
dnakamura: We should probably put a final else case here to complain if it cant find hookgen & tracegen on your path

charliegracie: @rwy0717 and @dnakamura is this something we are going to do in this PR?

rwy0717: I'm not sure. I've thought about it, and supporting installed tools properly might be outside the scope of this PR. I might just abort instead, and leave "support installed build tools" for a later PR. @dnakamura does that sound good?

dnakamura: fair enough

rwy0717: I've updated the code here to error out if the tools are disabled and not imported. I'll see about using installed hookgen/toolgen binaries in a later patch.

PR #1592, opened 2017-08-30
Test improvements

charliegracie:
A lot of the OMR tests take a very long time to complete. This change
reduces a number of these tests to a more managable time.  The
reductions were the result of fewer loops and smaller sleep/timout
times.


rwy0717:
LGTM
/cc @youngar

PR #1594, opened 2017-08-31
Merge stats across CS phases

amicic:
In Concurrent Scavenger gather per phase/increment Scavenger stats.
These are typically used for reported purposes. Than at the end of a
cycle merge those stats. Those are typically used for various heuristics
(like tilting) that are performed at the end of a cycle. Currently,
since we clear per increment stats and do not merge them, the heuristics
rely only on last increment, which is vary small data subset.

Standard Scavenger is not meant to be affected, although a lot of code
is shared with CS. There is still a notion of an increment, but there is
only one. We still merge that one with the cycle stats, so there are
identical.

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -363,8 +363,8 @@ MM_Scavenger::masterSetupForGC(MM_EnvironmentStandard *env)
 	uintptr_t regionSize = _extensions->heap->getHeapRegionManager()->getRegionSize();
 	Assert_MM_true((0 != regionSize) && (0 == ((uintptr_t)_heapBase % regionSize)));
 
-	/* Clear the gc statistics */
-	clearGCStats(env);
+	/* Clear the cycle gc statistics. Increment level stats will be cleared just prior to increment start. */
+	clearCycleGCStats(env);
 
 	/* invoke collector language interface callback */
 	_cli->scavenger_masterSetupForGC(env);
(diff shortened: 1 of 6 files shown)


gc/base/standard/Scavenger.cpp
```diff
@@ -784,15 +779,52 @@ MM_Scavenger::mergeGCStats(MM_EnvironmentBase *env)
...
+	Assert_MM_true(env->isMasterThread());
+	MM_ScavengerStats *finalGCStats = &_extensions->scavengerStats;
+	mergeGCStatsBase(env, finalGCStats, &_extensions->incrementScavengerStats);
+
+	/* Language specific stats were supposed to be merged directly from thread local to cycle global. No need to merge them here. */
+
+	if (lastIncrement) {
+		finalGCStats->getFlipHistory(0)->_tenureMask = _tenureMask;
+		uintptr_t tenureAge = 0;
+		for (tenureAge = 0; tenureAge <= OBJECT_HEADER_AGE_MAX; ++tenureAge) {
```
charliegracie: Can you put a brief explanation for this code in a comment?

amicic: I don't know exactly what it does. It was just moved out from common merge path, into the marge path that is done only once by master thread. Back there it was executed by every GC thread and protected with is-master-thread check.

amicic: There seem to be two parts, one is "Recalculate tenure age" and the other "Update Flip/tenure stats for age 0". Makes sense to you ?

charliegracie: Sorry I missed that it was removed from the other location.  I thought it was new code.  Since it is not new code it is ok as is.

amicic: I added those two comments anyway

PR #1595, opened 2017-08-31
Fixes to virtual guard head merger

ncough:
This change includes two fixes to virtual guard head merger. The first adjusts its
behaviour to detect cases where it is moving a privarg, whose temp has
been recycled and is live in the outer call site's cold path.

The second adjusts the optimization's behaviour to treat OSR guards as stop the
world and prevents it from moving them if the destination is not a valid OSR
point.

### DIFF
--- a/compiler/optimizer/VirtualGuardHeadMerger.cpp
+++ b/compiler/optimizer/VirtualGuardHeadMerger.cpp
@@ -150,8 +150,58 @@ static TR::Block* splitRuntimeGuardBlock(TR::Compilation *comp, TR::Block* block
    return block;
    }
 
+/**
+ * Collect direct loads in a node and its children, adding them to the provided BitVector.
+ *
+ * @param node The node to consider.
+ * @param loadSymRefs BitVector of symbol reference numbers seen.
+ * @param checklist Checklist of visited nodes.
+ */
+static void collectDirectLoads(TR::Node *node, TR_BitVector &loadSymRefs, TR::NodeChecklist &checklist)
+   {
+   if (checklist.contains(node))
+      return;
+   checklist.add(node);
+
+   if (node->getOpCode().isLoadVarDirect())
+      loadSymRefs.set(node->getSymbolReference()->getReferenceNumber());
+
+   for (int i = 0; i < node->getNumChildren(); i++)
+      collectDirectLoads(node->getChild(i), loadSymRefs, checklist);
+   }
+
+/**
+ * Search for direct loads in the taken side of a guard
+ *
+ * @param firstBlock The guard's branch destination
+ * @param coldPathLoads BitVector of symbol reference numbers for any direct loads seen until the merge back to mainline
+ */
+static void collectColdPathLoads(TR::Block* firstBlock, TR_BitVector &coldPathLoads)
+   {
+   TR_Stack<TR::Block*> blocksToCheck(TR::comp()->trMemory(), 8, false, stackAlloc);
+   blocksToCheck.push(firstBlock);
+   TR::NodeChecklist checklist(TR::comp());
+
+   coldPathLoads.empty();
+   while (!blocksToCheck.isEmpty())
+      {
+      TR::Block *block = blocksToCheck.pop();
+
+      for (TR::TreeTop *tt = block->getFirstRealTreeTop(); tt->getNode()->getOpCodeValue() != TR::BBEnd; tt = tt->getNextTreeTop())
+         collectDirectLoads(tt->getNode(), coldPathLoads, checklist);
+
+      // Search for any successors that have not merged with the mainline
+      for (auto itr = block->getSuccessors().begin(), end = block->getSuccessors().end(); itr != end; ++itr)
+         {
+         TR::Block *dest = (*itr)->getTo()->asBlock();
+         if (dest != TR::comp()->getFlowGraph()->getEnd() && dest->getPredecessors().size() == 1)
+            blocksToCheck.push(dest);
+         }
+      }
+   }
+
 static bool safeToMoveGuard(TR::Block *destination, TR::TreeTop *guardCandidate,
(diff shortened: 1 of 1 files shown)


compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -185,10 +222,40 @@ static bool safeToMoveGuard(TR::Block *destination, TR::TreeTop *guardCandidate,
...
+         {
+         TR::Block *block = blocksToCheck.pop();
+
+         // Search this block for any uses of the privargs we plan to move
+         for (TR::TreeTop *tt = block->getFirstRealTreeTop(); tt->getNode()->getOpCodeValue() != TR::BBEnd; tt = tt->getNextTreeTop())
+            if (searchNodeForUses(tt->getNode(), privArgSymRefs, checklist))
+               return false;
+
+         // Search for any successors that have not merged with the mainline
+         for (auto e = block->getSuccessors().begin(); e != block->getSuccessors().end(); ++e)
```
andrewcraik: I would prefer to have an itr and an end to match other code and to stash the end to save compile time. Something like:
for (auto itr = block->getSuccessors().begin(), end = block->getSuccessors().end(); itr != end; ++itr)

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -150,27 +155,59 @@ static TR::Block* splitRuntimeGuardBlock(TR::Compilation *comp, TR::Block* block
...
+static bool searchNodeForUses(TR::Node *node, TR_BitVector &symRefsToCheck, TR::NodeChecklist &checklist)
+   {
+   if (checklist.contains(node))
+      return false;
+   checklist.add(node);
+
+   if (node->getOpCode().isLoadVarDirect() && symRefsToCheck.get(node->getSymbolReference()->getReferenceNumber()))
+      return true;
+
+   for (int i = 0; i < node->getNumChildren(); i++)
```
andrewcraik: braces here for clarity would be good

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -185,10 +222,40 @@ static bool safeToMoveGuard(TR::Block *destination, TR::TreeTop *guardCandidate,
...
+      // taken. Give up if this situation is detected.
+      TR_Stack<TR::Block*> blocksToCheck(TR::comp()->trMemory(), 8, false, stackAlloc);
+      blocksToCheck.push(branchDest->getNode()->getBlock());
+      TR::NodeChecklist checklist(TR::comp());
+      while (!blocksToCheck.isEmpty())
+         {
+         TR::Block *block = blocksToCheck.pop();
+
+         // Search this block for any uses of the privargs we plan to move
+         for (TR::TreeTop *tt = block->getFirstRealTreeTop(); tt->getNode()->getOpCodeValue() != TR::BBEnd; tt = tt->getNextTreeTop())
```
andrewcraik: braces on this loop would help readability I think

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -185,10 +222,40 @@ static bool safeToMoveGuard(TR::Block *destination, TR::TreeTop *guardCandidate,
...
              && tt->getNode()->getBranchDestination() != branchDest)
             return false;
          }
+
+      // Search for loads of the collected privargs in the taken side of guard1
+      // This is possible if a privarg temp was recycled for the inner call site, with a prior use as an
+      // argument for the outer call site. As the privargs for the inner call site must be evaluated before
+      // both guards, this would result in the recycled temp holding the incorrect value if the guard is ever
+      // taken. Give up if this situation is detected.
+      TR_Stack<TR::Block*> blocksToCheck(TR::comp()->trMemory(), 8, false, stackAlloc);
```
andrewcraik: would it be cheaper to flip this the other way - gather those used on the taken side of the guard when we pick a new first guard and have a candidate to merge and keep reusing the set for the subsequent priv args? Might be lower in compile time overall, but possibly more complex. Thoughts?

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -47,6 +47,11 @@ static bool isMergeableGuard(TR::Node *node)
    return mergeOnlyHCRGuards ? node->isHCRGuard() : node->isNopableInlineGuard();
    }
 
+static bool isStopTheWorldGuard(TR::Node *node)
+   {
+   return node->isHCRGuard() || node->isOSRGuard();
+   }
+
```
fjeremic: While true, this query is incomplete. Profiled guards are also "stop the world" guards as seen in the following comment:
https://github.com/eclipse/omr/blob/28a04889bbb3906fa8dfc9635cfcf4798f95c827/compiler/z/codegen/S390Instruction.cpp#L5792-L5804

Is it possible to add profiled guards to this list? If so that would be great, because we could abstract this API into a common place and reuse it both here and in the codegen location linked above.

ncough: Profiled guards aren't included in this query as this optimization will ignore them in `isMergeableGuard `. As all guards must first be mergeable to be considered, they weren't a concern. Therefore, using a common API shouldn't modify the behaviour.

ncough: Profiled guards shouldn't be stop the world, so Z codegen might be mistaken here. A common API to check for stop the world guards should be added though, in a later change.

ncough:
The readability concerns have been corrected and the recycled temp check has been rearranged to use the analysis result if other guards are being merged.

andrewcraik:
LGTM

andrewcraik:
@vijaysun-omr could you have a look?

PR #1597, opened 2017-08-31
Remove ZOS Linkage support

dmitripivkine:
Remove ZOS Linkage support required for correct linkage of classes with
long names for old compilers. This code is not necessary any more


dmitripivkine:
This change set should be applied after promotion of correspondent part in J9

dmitripivkine:
This item is ready to be merged

PR #1609, opened 2017-09-05
CMake: Overhaul gc options

dnakamura:
Modify the gc config options to include documentation and dependency
info


0xdaryl:
What is the current disposition of this PR?  Are the CI failures real?  Is it ready for a review?

dnakamura:
yeah ci failures are real. Its been on the back burner. There is some option that is now silently falling back because it doesn't meet its pre-requisites

dnakamura:
Actually. I'm not sure I really like this approach because things silently fail. Will open new PR when I have better solution

PR #1610, opened 2017-09-06
Use CCACHE on OSX

charliegracie:
- Install CCACHE on OSX via homebrew
- Use clang as the compiler for OSX. It was getting used since gcc/g++
are symlinks to clang but we should be explicit 


youngar:
All of the travis test passed for this, even though GitHub currently says it is still in progress.

PR #1612, opened 2017-09-06
ELF relocatable object file support

lmaisons:


charliegracie:
@lmaisons it looks like you did not update the CMakeLists.txt for your changes.

compiler/codegen/ELFObjectFileGenerator.cpp
```diff
@@ -0,0 +1,307 @@
...
+   elfSymbols[0].st_other = 0;
+   elfSymbols[0].st_shndx = 0;
+   elfSymbols[0].st_value = 0;
+   elfSymbols[0].st_size = 0;
+
+   ELFSymbol *elfSym = elfSymbols + 1;
+   char *names = elfSymbolNames + 1;
+   for (auto it = _symbols.begin(); it != _symbols.end(); ++it)
+      {
+      fprintf(stderr, "Writing elf symbol %d, name(%d) = %s\n", (elfSym - elfSymbols), it->_nameLength, it->_name);
```
0xdaryl: This comment applies to all the `fprintf(stderr, ...)`'s used in this PR.  While useful, I think they should be guarded with some kind of trace option to control the amount of output.

compiler/codegen/ELFObjectFileGenerator.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include "runtime/CodeCacheManager.hpp"
+#include "codegen/ELFRelocationResolver.hpp"
+
+class TR_Memory;
+
+namespace TR {
+
+class ELFObjectFileGenerator
+   {
+public:
```
0xdaryl: The API of this new class should have Doxygen comments.

compiler/codegen/StaticRelocation.hpp
```diff
@@ -0,0 +1,74 @@
...
+#define STATICRELOCATION_HPP
+
+#pragma once
+
+#include <inttypes.h>
+
+class TR_ResolvedMethod;
+
+namespace TR {
+
```
0xdaryl: The API of this new class should have Doxygen comments.

compiler/codegen/ELFObjectFileGenerator.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ *******************************************************************************/
+
+#ifndef ELFFILEGENERATOR_HPP
+#define ELFFILEGENERATOR_HPP
+
+#if defined(LINUX)
+
+#include <elf.h>
```
0xdaryl: Will `elf.h` always be present on Linux systems Eclipse OMR will be built on or is there a separate package to install?

compiler/codegen/ELFObjectFileGenerator.cpp
```diff
@@ -0,0 +1,307 @@
...
+   _elfHeader.e_phentsize = 0;
+   _elfHeader.e_phnum = 0;
+   _elfHeader.e_shentsize = sizeof(ELFSectionHeader);
+   _elfHeader.e_shnum = 6; // number of sections in trailer
+   _elfHeader.e_shstrndx = 4; // index to shared string table section in trailer
+   }
+
+void
+TR::ELFObjectFileGenerator::initializeELFHeaderForPlatform()
+   {
```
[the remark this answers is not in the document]
lmaisons: Largely because at this point it no longer requires its own tangle of `#if defined` (having already gone through that pain in the header)

lmaisons:
The duplication between `ELFObjectFileGenerator` and `CodeCacheManager` is real, and yes, would ideally be de-duplicated in some base class.  Because of time constraints, and because the goal was a prototype, I prioritized merging what existed and worked over being as aggressive as possible in reducing duplication.

0xdaryl:
I will accept these changes as-is.  However, please open an issue to track the ELF object refactoring and consolidation.

Issue #1606, opened 2017-09-05
Request contributor approval to relicense all contributions


mstoodle:
With the committer vote in #1596 having passed (well, technically, it hasn't closed yet but enough committers have +1'd that the result cannot change), we need to collect approvals from all contributors to relicense their contributions under the new license.  For reference, here is the new license information:

To be more specific, the Eclipse OMR project has decided to change its license to the following:
```
This program and the accompanying materials are made available under the terms of the Eclipse Public License 2.0 which accompanies this distribution and is available at http://eclipse.org/legal/epl-2.0 or the Apache License, Version 2.0 which accompanies this distribution and is available at https://www.apache.org/licenses/LICENSE-2.0.

This Source Code is also Distributed under one or more Secondary Licenses, as those terms are defined by the Eclipse Public License, v. 2.0: GNU General Public License, version 2 with the GNU Classpath Exception [1] and GNU General Public License, version 2 with the OpenJDK Assembly Exception [2].

[1] https://www.gnu.org/software/classpath/license.html 
[2] http://openjdk.java.net/legal/assembly-exception.html
```

This license updates our existing Eclipse Public License from v1 to the latest v2, and adds a secondary license compatibility statement so that our project code can be combined with the GPLv2 license described above. This change only expands the scenarios where the Eclipse OMR code can be used, which consequently requires contributors to permit broader licensing terms on their contributions. This broader licensing requirement is why we need this approval request from all existing contributors to the project.

To simplify the approval process, any contributor who was an IBM employee at the time of their contribution can be covered by a single executive-level IBM approver. After discussion with the Eclipse Foundation, @jduimovich can provide this approval for all IBM contributions (including his own :) ).

Not all contributions have come from IBM employees, however, so there are still a few people who we must explicitly ask to give their approval to re-license all their contributions, specifically @amitin @erik-brangs @gurinderu @Krovatkin .
 
Would everyone @ mentioned directly in this issue please respond with a +1 to confirm that you agree to relicense your contributions (all IBM contributions in @jduimovich's case) according to the license outlined above (I suppose -1 is also a possible response)?

If there are any non-IBM contributors that are not identified in this list, please mention it here (and give your +1/-1).

amitin:
+1

jduimovich:
+1

Krovatkin:
+1

erik-brangs:
+1

gurinderu:
+1

mstoodle:
[...]

Issue #1596, opened 2017-08-31
Updating our license to EPLv2


mstoodle:
With this project's largest stakeholder becoming an open source project (Eclipse OpenJ9), a licensing compatibility issue has arisen that affects Eclipse OMR. I would like to resolve this via a committer vote, but we also need to collect some data from our contributors.

What's the issue?

The Eclipse OpenJ9 Java Virtual Machine project both consumes Eclipse OMR project and must have a compatible license story to work within the context of OpenJDK. The current state is:

- Eclipse OMR has dual Apache License 2.0 and Eclipse Public License 1.0

- OpenJDK has GNU General Public License, version 2 with the GNU Classpath Exception [1] and GNU General Public License, version 2 with the OpenJDK Assembly Exception [2].

OpenJ9 is stuck in the middle :) . What they have opted to do is a more complicated license to support two primary activities: 1) moving code back and forth between OMR and OpenJ9, and 2) being used alongside the precise flavour of GPLv2 license (with the noted exceptions) used OpenJDK.

To be more specific, Eclipse OpenJ9 is going with this license:
```
This program and the accompanying materials are made available under the terms of the Eclipse Public License 2.0 which accompanies this distribution and is available at http://eclipse.org/legal/epl-2.0 or the Apache License, Version 2.0 which accompanies this distribution and is available at https://www.apache.org/licenses/LICENSE-2.0.

This Source Code is also Distributed under one or more Secondary Licenses, as those terms are defined by the Eclipse Public License, v. 2.0: GNU General Public License, version 2 with the GNU Classpath Exception [1] and GNU General Public License, version 2 with the OpenJDK Assembly Exception [2].
```

For that license to work for the two activities mentioned, Eclipse OMR would have to have this same license. I propose that Eclipse OMR should adopt the license above. We lose nothing that I can see, and will continue to work with our largest stakeholder, the Eclipse OpenJ9 project.

I would like the committers to vote on this proposal with Agree (+1), Disagree (-1), or Abstain (0). I will hold the voting open for one week, but if the vote has been decided (i.e. enough +1's or -1's have been submitted that the remaining committer votes cannot change the result), then I will call the vote complete. There are additional tasks to perform if the vote succeeds (to ensure all contributors will provide their contributions under the new license).

[...]

PR #1613, opened 2017-09-07
WIP: Relicense the port library to EPLv2 & Apache v2

DanHeidinga:
Update the license as per the outcome of the vote in issue #1596.

This can't be accepted until all contributors in  #1606 have signed off on the relicensing their contributions.

### DIFF
--- a/port/zos390/omrvmem_support_below_bar_64.s
+++ b/port/zos390/omrvmem_support_below_bar_64.s
@@ -1,20 +1,26 @@
 ***********************************************************************
+* Copyright (c) 1991, 2016 IBM Corp. and others
 *
-* (c) Copyright IBM Corp. 1991, 2016
+* This program and the accompanying materials are made available under
+* the terms of the Eclipse Public License 2.0 which accompanies this
+* distribution and is available at
+* https://www.eclipse.org/legal/epl-1.0/ or the Apache License,
+* Version 2.0 which accompanies this distribution and is available at
+* https://www.apache.org/licenses/LICENSE-2.0.
 *
-*  This program and the accompanying materials are made available
-*  under the terms of the Eclipse Public License v1.0 and
-*  Apache License v2.0 which accompanies this distribution.
+* This Source Code is also Distributed under one or more Secondary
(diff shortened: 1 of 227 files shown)


DanHeidinga:
@mstoodle Can you confirm this is the correct license text?  In particular, not sure about the gpl / class path part

charliegracie:
I looked at a few files and the text appears to be correct.  Thanks for jumping on this work but note we have not given the ok to make the switch yet so it may be a while before we can merge these changes.

port/zos390/omrvmem_support_below_bar_64.s
```diff
@@ -204,30 +204,6 @@ OMRSTORAGE RMODE ANY                                                     000000
 @@CCN@19 ALIAS C'omrfree_memory_below_bar'                               000000
 @@CCN@11 ALIAS C'omrallocate_4K_pages_below_bar'                         000000
 @@CCN@1  ALIAS C'omrallocate_1M_pageable_pages_below_bar'                000000
-************************************************************************
```
charliegracie: I believe the duplicate copy was there so that the generated file also contained the copyright

DanHeidinga: That's annoying.  I'll check on that and update here once I've confirmed

DanHeidinga: Compiled on zOS and only see these files:
```
$ cd ~/omr/port
$ find . -name "omrvmem_support_*"                                    
./omrvmem_support_above_bar.o
./omrvmem_support_below_bar_64.o
./zos390/omrvmem_support_below_bar_64.s
./zos390/omrvmem_support_below_bar_31.s
./zos390/omrvmem_support_above_bar.s
```

Is there a specific generated file that needs to be looked at?  I'm not able to find it...

charliegracie:
This will need to be rebased.

There were last minute changes to the copyright header which will need to be incorporated into these changes.  You can see this change in #1606

DanHeidinga:
Closing this in favour of https://github.com/eclipse/omr/pull/1622 which has the correct copyrights.

PR #1614, opened 2017-09-07
WIP: update copyrights on util/ & tools/ to EPLv2 & Apache v2

DanHeidinga:
Update the license as per the outcome of the vote in issue #1596.

This can't be accepted until all contributors in #1606 have signed off on the relicensing their contributions.

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

PR #1617, opened 2017-09-07
Relicense under EPLv2 & Apache v2

keithc-ca:
Update the license as per the outcome of the vote in issue #1596.

This can't be accepted until all contributors in #1606 have signed off on the relicensing their contributions.

This addresses files in the cmake and compiler folders.


0xdaryl:
What's the intersection between this PR and https://github.com/eclipse/omr/pull/1613 ?

keithc-ca:
There should be no overlap among #1613, #1614, #1617 and #1618 (and we're not done yet).

charliegracie:
There were last minute changes to the copyright header which will need to be incorporated into these changes.  You can see this change in #1606

keithc-ca:
Rebased with updated license text.

compiler/arm/codegen/ARMDebug.cpp
```diff
@@ -1,19 +1,22 @@
 /*******************************************************************************
+ * Copyright (c) 2000, 2016 IBM Corp. and others
```
mstoodle: Hm, technically this file is being changed in 2017 to update its copyright header....shouldn't we just unilaterally update the second date to 2017 in all cases?

mstoodle:
The requirement in #1606 has been met: when it's ready, this PR can be merged.

compiler/arm/codegen/ARMDebug.cpp
keithc-ca: It was my understanding that the date should change only if something other than the copyright itself is different.

charliegracie: I tend to agree with @keithc-ca that we should not update the dates because we are changing the copyright header. @mstoodle ?

DanHeidinga: @mstoodle If you're ok with the dates, can you approve this so its obviously ready to merge once the PR builds pass?

mstoodle: technically, the copyright statement applies to the copyright statement too :) but I think it's defensible not to change the date.

keithc-ca:
Closing: changes are included in https://github.com/eclipse/omr/pull/1629.

PR #1624, opened 2017-09-08
Remove omr_shared

rwy0717:
OmrPlatform is meant to be re-includable. However, when we define
omr_shared in this module, this is no longer the case.

omr_shared was used to pass compile options needed to create shared
libraries. Any shared library in OMR would link privately against this
lib.

Instead, we introduce a new platform variable
OMR_PLATFORM_SHARED_COMPILE_OPTIONS

This variable contains the same set of compile options normally
inherited through omr_shared. All usages of omr_shared have been replace
with target_add_compile_options and this new variable.

Besides making OmrPlatform reincludable, this also makes shared compile
options more conventional.

/cc @dnakamura @mgaudet @charliegracie

### DIFF
--- a/cmake/modules/OmrPlatform.cmake
+++ b/cmake/modules/OmrPlatform.cmake
@@ -47,11 +47,6 @@ include(platform/toolcfg/${OMR_TOOLCONFIG})
 # Verify toolconfig!
 include(platform/toolcfg/verify)
 
-
-# interface library for exporting symbols from dynamic library
-# currently does nothing except on zos
-add_library(omr_shared INTERFACE)
-
 macro(omr_platform_global_setup)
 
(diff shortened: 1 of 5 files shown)


dnakamura:
I was looking at this and we could be making cmake do this automatically for us. CMake does have a seem to have variables to handle this kind of thing, but It not documented so I'm not sure if we want to rely on it. (Although i dont think they will break it, since you need to rely on the interface if you are writing your own compiler detection)

charliegracie:
Do we want to try that @dnakamura as another PR?  I am ok with accepting this as is and then trying the other way afterwards.  Up to you and @rwy0717 to decide.

dnakamura:
nah. merge as is. Im not even 100% sure that such magic actually exists.

PR #1633, opened 2017-09-12
Test output cleanup

charliegracie:
Cleanup hashtable, avl and pool testing to remove extra output

Add a new listener which prints less data
- Create a new TestEventListener which does not output as much data by
default as the default TestEventListener.  Make this new
TestEventListener the listener used by OMR testing.

### DIFF
--- a/fvtest/algotest/main.cpp
+++ b/fvtest/algotest/main.cpp
@@ -34,6 +34,8 @@ testMain(int argc, char **argv, char **envp)
 {
 	::testing::InitGoogleTest(&argc, argv);
 
+	OMREventListener::setDefaultTestListener();
+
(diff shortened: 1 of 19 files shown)


fvtest/algotest/main.cpp
```diff
@@ -34,6 +34,8 @@ testMain(int argc, char **argv, char **envp)
 {
 	::testing::InitGoogleTest(&argc, argv);
 
+	OMREventListener::setDefaultTestListner();
```
rwy0717: Typo: `s/listner/listener/`

charliegracie: That would be a problem everywhere :)  I will update it and push again

PR #1635, opened 2017-09-12
Fix omr/compiler warnings

charliegracie:
The compiler component has a lot of compiler warnings.  This PR will resolve some of the simple ones.  There will be a follow on Issue/PR for some of the more complicated ones.

### DIFF
--- a/compiler/env/OMRIO.hpp
+++ b/compiler/env/OMRIO.hpp
@@ -55,12 +55,12 @@ namespace OMR { typedef OMR::IO IOConnector; }
    #if defined(LINUX) || defined(OSX)
       #ifdef TR_HOST_64BIT
          #ifdef TR_TARGET_X86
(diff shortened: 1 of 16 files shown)


vijaysun-omr:
Looks straightforward to me. Merging.

compiler/env/OMRIO.hpp
```diff
@@ -55,12 +55,12 @@ namespace OMR { typedef OMR::IO IOConnector; }
    #if defined(LINUX) || defined(OSX)
       #ifdef TR_HOST_64BIT
          #ifdef TR_TARGET_X86
-            #define POINTER_PRINTF_FORMAT "%012p"
+            #define POINTER_PRINTF_FORMAT "%12p"
```
keithc61: This has the effect of eliminating the leading zeros: was that intentional?

PR #1639, opened 2017-09-13
Fix compilation issues on zOS

charliegracie:
m4 on zOS can not handle C/C++ style comments. This file is included in
a C++ header and a few m4 files so the copyright header needs to work
for both. Remove the copyright header while a solution is being found.

zOS assembler can not handle lines with more than 71 characters even in
comments. When a solution is found ensure that the length of each line
in the copyright header is a maximum of 71 characters.

[ci skip] since this is just removing a copyright header


mstoodle:
LGTM, thanks @charliegracie !

PR #1654, opened 2017-09-14
Remove .swp file & improve .gitignore

keithc-ca:
The file /.swp appears to have been (unintentionally) added during the mass license change.
Updated `.gitignore` to reduce the likelihood it happens again.


youngar:
LGTM.  What program makes the .swp file?

rservant:
Vim is one application that creates .swp files.

https://unix.stackexchange.com/questions/27923/what-causes-swap-files-to-be-created

PR #1656, opened 2017-09-14
Remove cold code paths

arianneb:
Modified function allowSplitWarmAndColdBlocks() to return true which sets the 
cold code length to 0. This ensures that the cold code paths are never taken. 
After testing with this change, it has been verified that everything works without 
the cold code paths. The cold code paths were then removed and  
allowSplitWarmAndColdBlocks() was reverted back to return false (such that any 
dependencies in openj9 may continue to use it normally, for now). 

[ci skip]

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -6322,16 +6322,8 @@ OMR::Z::CodeGenerator::doBinaryEncoding()
       _extentOfLitPool = self()->setEstimatedOffsetForConstantDataSnippets(_extentOfLitPool);
       }
 
-   if (self()->allowSplitWarmAndColdBlocks())
-      {
-      self()->setEstimatedWarmLength(data.estimate);
-      self()->setEstimatedColdLength(0);
-      }
-   else
-      {
-      self()->setEstimatedWarmLength(0);
-      self()->setEstimatedColdLength(data.estimate);
-      }
+   self()->setEstimatedWarmLength(data.estimate);
+   self()->setEstimatedColdLength(0);
 
    data.cursorInstruction = self()->comp()->getFirstInstruction();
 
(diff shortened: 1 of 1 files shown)


compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -376,7 +376,7 @@ class OMR_EXTENSIBLE CodeGenerator
    bool buildInterpreterEntryPoint() { return false; }
    void generateCatchBlockBBStartPrologue(TR::Node *node, TR::Instruction *fenceInstruction) { return; }
    bool supportsUnneededLabelRemoval() { return true; }
-   bool allowSplitWarmAndColdBlocks() { return false; }
```
mgaudet: Is the goal to remove this query entirely? 

I'm a little surprised to see it flip from disabled to enabled, given most places in this it's been deleted. If this is to influence a downstream project like OpenJ9, shouldn't the code be removed from OpenJ9, then subsequently this PR removes the OMR impl?

arianneb: I believe the goal is to remove it, yes. There are still j9 dependencies on it. We are completing this task in 3 phases. That piece will come later once we can verify that what we've already done is kosher. @mgaudet

mgaudet: I guess the thing I don't understand is why this flips sense from false to true though. Your PR message seems to suggest that you had tried to revert it, but it didn't take.

arianneb: So when allowSplitWarmAndColdBlocks returns true, it sets the cold code length to 0. This ensures that the cold code paths will never be taken. I made this simple modification and ran a personal build to ensure that things will pass without using any of the cold code paths. Once that was verified, I went on to remove the cold code paths altogether and reverted allowSplitWarmAndColdBlocks to return false, such that any j9 dependencies can continue to use it for now. @mgaudet

xliang6:
@arianneb I would ask a review from @fjeremic and @0xdaryl once you address my concerns. Thanks.

xliang6:
LGTM.

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -6496,33 +6467,8 @@ OMR::Z::CodeGenerator::doBinaryEncoding()
...
-
-   if ( !self()->allowSplitWarmAndColdBlocks() )
-      {
-      // We need this, otherwise getWarmCodeLength() would compute a -ive value,
-      // which in-turn would make the later assert check fail. The negative
-      // value comes about because getCodeStart() is larger _binaryBufferStart,
-      // which is what warmCodeEnd is initlaized to.
-      self()->setWarmCodeEnd(self()->getCodeStart());
-      }
-
```
xliang6: On second thought, can we get rid off code setting ```doBinaryEncoding``` to true and the ```if``` statement checking if ```doBinaryEncoding``` is true (if that is the only live code path)?

arianneb:
I tested locally with `TR_ASSERT` enabled and by adding a `TR_ASSERT(0, ...)` in `omr/compiler/compile/OMRCompilation.cpp` in function `OMR::Compilation::Compilation(…)`. Then I ran a java -version with `-Xjit:count=0` and the program crashed as expected. I will test on the farm with `TR_ASSERT` enabled once everything is back up and running. @fjeremic @0xdaryl

arianneb:
Commits are now split. Please append the URL of the second commit (Further cleanup of doBinaryEncoding()) with `?w=1` to ignore the white space. The un-indenting that I did makes the change difficult to read.

0xdaryl:
omr-genie build all

0xdaryl:
genie-omr build all

arianneb:
I fixed the spelling. Testing is complete. This should be good to merge now. @0xdaryl

PR #1657, opened 2017-09-14
Fix Windows compile issue

charliegracie:
Older versions of msvc run with c89 support by default. The AppVeyor
builds run with c99 by default so this was not caught in the CI testing


charliegracie:
A simple fix for a compile issue so I added multiple Assignees so it can get merged once the CI passes

PR #1660, opened 2017-09-15
Change windows compiler to catch c89 issues

charliegracie:
Strictly speaking visual studio does not support c89 or c99.  It supports which ever features it needs to be a compliant c++ compiler for the c++ spec you are using.  I am downgrading the visual studio version to a lower level that catches this issue since we want to 2012 or maybe even 2010 still.

This also fixes compile failures related to the downgrade.

~This depends on #1657.  I will rebase once I see the AppVeyor build fail so we know that it fails and then I can show that it is resolved when I rebase.~ Rebased and ready to go :)


charliegracie:
@mstoodle and @youngar  what are your thoughts on this?  I would like to use the older Visual Studio so that we catch these types of errors and continue to be backwards compatible with older visual studio versions.

charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
genir-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

PR #1661, opened 2017-09-15
Support Tests using a "Mock" strategy 

mgaudet:
This functionality allows code external to the compiler (such as a test
case) to specify a mockOptimization strategy that overrides the
optimization strategy a particular method might be expected to use.

This is currently not a thread-safe implementation beause doing a
thread-safe implementation would require a more invasive compilation
control modification (since the strategy would need to be injected into
the Compilation object, but due to where it's created, there's little
opportunity for this.)

The good news is that Google Test doesn't test in threads, so for now
we are spared the necessity for dealing with this.


compiler/optimizer/OMROptimizer.hpp
```diff
@@ -169,6 +169,19 @@ class Optimizer
    static const char * getOptimizationName(OMR::Optimizations opt);
 
    static const OptimizationStrategy *optimizationStrategy( TR::Compilation *c);
+
+   /**
+    * Override what #OptimizationStrategy to use. While this has the same
+    * functionality as the `optTest=` parameter, it does not require
+    * reinitializing the JIT, nor manually changing the optFile.
+    *
+    * Unless _mockStrategy is non-null, has no effect on the optimizer.
```
0xdaryl: Typo: "this has"

compiler/optimizer/OMROptimizer.hpp
```diff
@@ -317,6 +331,13 @@ class Optimizer
    bool                          _isIlGen;
 
    const OptimizationStrategy *          _strategy;
+
+   /* 
+    * Since mock strategies are only used in testing right now, we make this
+    * static to ease implementation. 
+    */
```
0xdaryl: I think a summary of this blurp from the commit as a comment here will help future archaeologists understand why you're making it static:
```
This is currently not a thread-safe implementation beause doing a
thread-safe implementation would require a more invasive compilation
control modification (since the strategy would need to be injected into
the Compilation object, but due to where it's created, there's little
opportunity for this.)
```

0xdaryl:
Also, 'because' is spelled incorrectly in your commit message.

compiler/optimizer/OMROptimizer.hpp
mgaudet: Sorry -- I don't see the issue here 😞

mgaudet:
Renamed and updated this PR, as I intentionally re-scoped it to include a demo.

0xdaryl:
genie-omr build all

vijaysun-omr:
Tests pass and @0xdaryl has approved the changes. Merging.

PR #1663, opened 2017-09-15
Correct type of sprintf for a tril test

mgaudet:
Tiny type change.


mgaudet:
So it turns out the correct version of this is still blocked on [another parser issue](https://github.com/eclipse/omr/pull/1665#issuecomment-330262008)

mgaudet:
#1665 has gone in, unblocking this.

PR #1665, opened 2017-09-15
Fix incorrect parsing of Int values

rbanerjee:
There was an error in tril when trying to parse a large hex value (such as  0xFFFFFFFF00000000). This corrects that, and makes the parser test more robust.

### DIFF
--- a/fvtest/tril/test/ParserTest.cpp
+++ b/fvtest/tril/test/ParserTest.cpp
@@ -26,25 +26,28 @@
 #define ASSERT_NULL(pointer) ASSERT_EQ(nullptr, (pointer))
 #define ASSERT_NOTNULL(pointer) ASSERT_TRUE(nullptr != (pointer))
 
-TEST(ParserTest, SingleNodeWithJustName) {
-    auto trees = parseString("(nodeName)");
+
+typedef std::pair<const char *, const char *> TestParamWithExpectedResultString;
+
+class SingleNodeWithName : public testing::TestWithParam<TestParamWithExpectedResultString> {};
+
+TEST_P(SingleNodeWithName, VerifyNode) {
+    auto trees = parseString(GetParam().first);
 
     ASSERT_NOTNULL(trees);
-    ASSERT_STREQ("nodeName", trees->getName());
+    ASSERT_STREQ(GetParam().second, trees->getName());
     ASSERT_NULL(trees->getArgs());
     ASSERT_NULL(trees->getChildren());
     ASSERT_NULL(trees->next);
 }
 
-TEST(ParserTest, SingleCommandNodeWithJustName) {
-    auto trees = parseString("(@commandName)");
-
-    ASSERT_NOTNULL(trees);
-    ASSERT_STREQ("@commandName", trees->getName());
-    ASSERT_NULL(trees->getArgs());
-    ASSERT_NULL(trees->getChildren());
-    ASSERT_NULL(trees->next);
-}
+INSTANTIATE_TEST_CASE_P(
+    ParserTest,
+    SingleNodeWithName,
+    testing::Values(
+        TestParamWithExpectedResultString("(nodeName)", "nodeName"),
+        TestParamWithExpectedResultString("(@commandName)", "@commandName")
+));
 
 TEST(ParserTest, TwoNodesWithJustName) {
     auto trees = parseString("(nodeName)(otherNode)");
@@ -100,8 +103,12 @@ TEST(ParserTest, NodeWithChildAndSibling) {
     ASSERT_NULL(sibling->next);
 }
 
-TEST(ParserTest, SingleNodeWithIntArg) {
-    auto trees = parseString("(nodeName 3)");
+typedef std::pair<const char *, ASTValue::Integer_t> TestParamWithExpectedResultInt;
+
+class SingleNodeWithIntArg : public testing::TestWithParam<TestParamWithExpectedResultInt> {};
+
+TEST_P(SingleNodeWithIntArg, VerifyNode) {
+    auto trees = parseString(GetParam().first);
 
     ASSERT_NOTNULL(trees);
     ASSERT_STREQ("nodeName", trees->getName());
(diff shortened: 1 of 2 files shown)


fvtest/tril/test/ParserTest.cpp
```diff
@@ -26,26 +26,33 @@
...
+    ASSERT_STREQ(GetParam().second, trees->getName());
     ASSERT_NULL(trees->getArgs());
     ASSERT_NULL(trees->getChildren());
     ASSERT_NULL(trees->next);
 }
 
+// The generated tests from this instantiation will have the following names:
+// ParserTest/SingleNodeWithName.VerifyNode/0 for TestParamWithExpectedResultString("(nodeName)", "nodename") ..
+// With the convention being:
+// InstantiationName/FooTest.DoesBlah/n for val_n
```
Leonardo2718: Not sure this comment is needed.

Leonardo2718: Not sure this comment is needed, but I guess it doesn't hurt to have around.

fvtest/tril/test/ParserTest.cpp
```diff
@@ -100,24 +107,16 @@ TEST(ParserTest, NodeWithChildAndSibling) {
     ASSERT_NULL(sibling->next);
 }
 
-TEST(ParserTest, SingleNodeWithIntArg) {
-    auto trees = parseString("(nodeName 3)");
+// aliases for the underlying C++ types used to store the different AST types.
+using Integer_t = uint64_t;
+using FloatingPoint_t = double;
```
Leonardo2718: Could you use `ASTValue::Integer_t` and `ASTValue::FloatingPoint_t` instead? This will guarantee that the types are the same as the ones used by Tril.

fvtest/tril/test/ParserTest.cpp
```diff
@@ -128,29 +127,35 @@ TEST(ParserTest, SingleNodeWithHexArg) {
...
     ASSERT_EQ(ASTValue::Integer, arg->getValue()->getType());
-    ASSERT_EQ(0xabc123, arg->getValue()->getInteger());
+    ASSERT_EQ(GetParam().second, arg->getValue()->getInteger());
     ASSERT_NULL(arg->next);
 }
 
-TEST(ParserTest, SingleNodeWithNegativeHexArg) {
-    auto trees = parseString("(nodeName -0x249DEF)");
+INSTANTIATE_TEST_CASE_P(
+    ParserTest,
```
Leonardo2718: For consistency with the other test, I would rather this be `ParserTestSingleNodeWithIntArgAsDecValue`.

fvtest/tril/test/ParserTest.cpp
```diff
@@ -161,10 +166,18 @@ TEST(ParserTest, SingleNodeWithFloatArg) {
...
+    ASSERT_EQ(GetParam().second, arg->getValue()->getFloatingPoint());
     ASSERT_NULL(arg->next);
 }
 
+INSTANTIATE_TEST_CASE_P(
+    ParserTest,
+    SingleNodeWithFloatArg,
+    testing::Values(
+        TestParamWithExpectedResultFloat("(nodeName 3.00)", 3.00),
+        TestParamWithExpectedResultFloat("(nodeName -3.00)", -3.00)
```
Leonardo2718: Might be worth having a few (at least one) edge cases here.

fvtest/tril/tril/tril.l
```diff
@@ -43,7 +43,7 @@
             };
 
 [-]?0x[0-9a-fA-F]+      {
-                yylval.integer = strtol(yytext, NULL, 16);
+                yylval.integer = strtoul(yytext, NULL, 16);
```
Leonardo2718: Good catch!

fvtest/tril/test/ParserTest.cpp
rbanerjee: I'm okay either way, don't think it matters that much. @mgaudet, thoughts?

mgaudet: I'd say pull the comment -- just because it's documenting the expected behaviour of GoogleTest, and so isn't adding a huge amount value for experienced readers (though, I understand how it would help people new to the project; we just can't optimize the code for them, or else the code would be so thick with explanatory comment that it would be difficult to see what's going on)

mgaudet:
Speaking of more edge cases: 

```
diff --git a/fvtest/tril/test/ParserTest.cpp b/fvtest/tril/test/ParserTest.cpp
index 7075d0e2..21113b17 100644
--- a/fvtest/tril/test/ParserTest.cpp
+++ b/fvtest/tril/test/ParserTest.cpp
@@ -138,7 +138,8 @@ INSTANTIATE_TEST_CASE_P(
         TestParamWithExpectedResultInt("(nodeName 0)", 0),
         TestParamWithExpectedResultInt("(nodeName 3)", 3),
         TestParamWithExpectedResultInt("(nodeName -3)", -3),
-        TestParamWithExpectedResultInt("(nodeName 9223372036854775807)", 9223372036854775807U)
+        TestParamWithExpectedResultInt("(nodeName 9223372036854775807)", 9223372036854775807U),
+        TestParamWithExpectedResultInt("(nodeName 18446744073709551615)", 18446744073709551615ULL)
 ));
 
```

produces

```

[----------] 5 tests from ParserTest/SingleNodeWithIntArg
../fvtest/tril/test/ParserTest.cpp:130: Failure
      Expected: GetParam().second
      Which is: 18446744073709551615
To be equal to: arg->getValue()->getInteger()
      Which is: 9223372036854775807
[  FAILED  ] ParserTest/SingleNodeWithIntArg.VerifyNode/4, where GetParam() = (0x107d491ae pointing to "(nodeName 18446744073709551615)", 18446744073709551615) (0 ms)
```

rbanerjee:
Addressed the comments except for the one asking for more tests that test the parsing for floating point numbers. I will that as a separate pull request  

Also modified the previous commit messages to reflect that this now fixes parsing issues with Integers(represented as *both hex and decimal) that are longer than 'long'.

PR #1668, opened 2017-09-17
Add support for nopable breakpoint guard

cathyzhyi:
To lower the overhead for the breakpoint guard instruction sequence,
the guard is nopped when compiling a method and patched
only when a breakpoint is set for that method. This changeset contains
the code for adding a new runtime assumption for breakpoint guard

### DIFF
--- a/compiler/compile/VirtualGuard.cpp
+++ b/compiler/compile/VirtualGuard.cpp
@@ -227,12 +227,12 @@ TR_VirtualGuard::createBreakpointGuardNode
 #ifdef J9_PROJECT_SPECIFIC
 /*
  * Shape of a breakpoint guard:
- * iflcmpeq/ificmpeq
+ * iflcmpne/ificmpne
  *    land/iand
  *       lloadi/iloadi <ConstantPool shadow>
  *          aconst J9Method
  *       isBreakpointedBit 
- *    isBreakpointedBit 
+ *    iconst 0 
  */
    bool is64Bit = TR::Compiler->target.is64Bit();
    TR::SymbolReferenceTable * symRefTab = comp->getSymRefTab();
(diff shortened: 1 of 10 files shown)


compiler/compile/VirtualGuard.cpp
```diff
@@ -227,38 +227,31 @@ TR_VirtualGuard::createBreakpointGuardNode
...
 /*
  * Shape of a breakpoint guard:
- * iflcmpeq/ificmpeq
- *    land/iand
- *       lloadi/iloadi <ConstantPool shadow>
- *          aconst J9Method
+ * ificmpne
+ *    iand
+ *       bu2i
+ *          buloadi <ConstantPool shadow>
```
andrewcraik: buloadi - ie byte unsigned load indirect - is deprecated; we should be using bloadi since loading a byte width has no sign and then the bu2i conversion will handle the sign extension when widening to integer. In theory this could be a band and ifbcmpne and bconst but I'm not sure how well they would be handled by the optimizer so having an i and living with the zero extension for symmetry with other tests could make sense.

charliegracie:
genie-omr build

charliegracie:
genie-omr build

compiler/compile/VirtualGuard.cpp
```diff
@@ -227,38 +227,31 @@ TR_VirtualGuard::createBreakpointGuardNode
...
-   TR::Node * constantPool = TR::Node::createWithSymRef(is64Bit? TR::lloadi: TR::iloadi, 1, 1, aconstNode, fieldSymRef);
    aconstNode->setIsMethodPointerConstant(true);
    aconstNode->setInlinedSiteIndex(calleeIndex);
    aconstNode->setByteCodeIndex(0);
    TR::Node * flagBit = NULL;
    TR::Node *guard = NULL;
-   if (TR::Compiler->target.is64Bit())
-      {
-      flagBit = TR::Node::create(callNode, TR::lconst, 0, 0);
-      flagBit->setLongInt(comp->fej9()->offsetOfMethodIsBreakpointedBit());
```
vijaysun-omr: So this is a 32 bit field I assume regardless of whether one is on 64-bit or 32-bit platform ?

cathyzhyi: Thanks for the comments! The intention was to only load the last byte of that field to make the instruction sequence faster but on a second thought, this won't work for big endian platforms. I changed it back to the old implementation. The efficiency of the generated sequence shouldn't matter anyway because we would nop them.

vijaysun-omr:
Thanks for changing the commit in response to the question. I can merge this if @andrewcraik is okay with it.

andrewcraik:
LGTM

vijaysun-omr:
Checks passed. Merging.

0xdaryl:
I believe the changes made to `compiler/runtime/OMRRuntimeAssumptions.[ch]pp` should really live in the Eclipse OpenJ9 project in `trj9/runtime/J9RuntimeAssumptions.[ch]pp` as they are not consumed by OMR and introduce the use of an enum value (`RuntimeAssumptionOnMethodBreakPoint`) only defined by the OpenJ9 project.

PR #1676, opened 2017-09-18
Remove unused TR_HOST_Power macro

0xdaryl:
`TR_HOST_Power` is arguably a typo for the correct host macro `TR_HOST_POWER`.
As nothing in Eclipse OMR nor any known downstream projects sets this incorrect
macro, replace it with the correct macro.

However, because the "wrong" path has been taken for such a long time I question
the value of providing a separate value just for Power and folded the setting of
this option into a single store for all architectures.


0xdaryl:
@gita-omr @ymanton : I would appreciate a review from at least one of you please.

ymanton:
No objection from me.

gita-omr:
Looks right to me as well.

charliegracie:
genie-omr build

PR #1677, opened 2017-09-18
Add optimization options for OSR modes

ncough:
This adds optimization options for two OSR modes,
voluntary or involuntary OSR. It also updates
OSRDefAnalysis and OSRLiveRangeAnalysis to
rely on these options, rather than checking
for slot sharing information.


0xdaryl:
genie-omr build

andrewcraik:
LGTM

andrewcraik:
@ncough Looks like everyone is ok with this so please remove WIP once we can get this in - seems like a nice change for clarity and understanding of what is going on.

PR #1678, opened 2017-09-18
Cache CFG end for repeated OSR induce calls

ncough:
When adding an OSR guard, it is necessary to append
OSR blocks to the end of the CFG. The final treetop
is necessary for this, found by walking the entire
IL. Therefore, it is cheaper to maintain a pointer
to the last treetop through repeated OSR guard
insertions.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -594,7 +594,8 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
 
 
 bool
-OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR_ByteCodeInfo induceBCI, TR::TreeTop* branch, bool extendRemainder, int32_t offset)
+OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR_ByteCodeInfo induceBCI, TR::TreeTop* branch,
+    bool extendRemainder, int32_t offset, TR::TreeTop ** lastTreeTop)
    {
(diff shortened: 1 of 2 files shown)


charliegracie:
genie-omr build

andrewcraik:
LGTM

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -628,7 +629,12 @@ OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR_ByteCo
       osrBlock->getEntry()->getNode()->setByteCodeInfo(induceBCI);
       osrBlock->getExit()->getNode()->setByteCodeInfo(induceBCI);
 
-      cfg->findLastTreeTop()->join(osrBlock->getEntry());
+      // Load the cached CFG end if available, and store it when finished
```
vijaysun-omr: How do we ensure that no other logic (unrelated to adding OSR induction logic) appends code at the end of the list of treetops making the cached value "stale" ? 

I also find the reference to the "CFG" in this change a bit clunky in that what is being talked about is the end of the treetops in list order, which would seem to be unrelated to the CFG directly as such. The "CFG end" is already a phrase that has a meaning in the compiler, namely the dummy block without any treetops associated with it that all blocks ending in a return have CFG edge(s) to.

ncough: Other than OSR induction logic, no other blocks should be added by `induceOSRAfter `. If the caller could potentially add blocks between calls to this method, they must update the cached treetop accordingly. An assert has been added to reflect the final treetop must have no trees following it.

I have corrected the names to indicate the last treetop rather than the CFG end.

vijaysun-omr:
Looks much better now and thanks for adding the assert. Merging.

PR #1679, opened 2017-09-18
Store feGetEnv results in statics

ncough:
feGetEnv is a common debug tool to disable or enable
behaviour based on environment variables. To avoid
significant compile time overhead, the result of the
call should be stored in a static.

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -4608,7 +4608,8 @@ static bool isBitwiseLongComplement(TR::Node * n)
 template <typename T>
 static bool checkAndReplaceRotation(TR::Node *node,TR::Block *block, TR::Simplifier *s)
    {
-   if (feGetEnv("TR_DisableROLSimplification"))
+   static char* disableROLSimplification = feGetEnv("TR_DisableROLSimplification");
+   if (disableROLSimplification)
       return false;
 
    TR::Node *firstChild = node->getFirstChild();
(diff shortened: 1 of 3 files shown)


compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -4608,7 +4608,8 @@ static bool isBitwiseLongComplement(TR::Node * n)
 template <typename T>
 static bool checkAndReplaceRotation(TR::Node *node,TR::Block *block, TR::Simplifier *s)
    {
-   if (feGetEnv("TR_DisableROLSimplification"))
+   static char* env = (feGetEnv("TR_DisableROLSimplification"));
```
0xdaryl: Rather than `env`, why not call the auto `disableROLSimplification` to make it explicit?

compiler/x/codegen/OMRX86Instruction.cpp
```diff
@@ -991,7 +991,8 @@ void TR::X86RegRegInstruction::assignRegisters(TR_RegisterKinds kindsToBeAssigne
 
       bool regRegCopy = isRegRegMove();
       TR_X86OpCodes opCode = getOpCodeValue();
-      if (feGetEnv("TR_UseOutOfDatedRegRevMoveList"))
+      static char* env = feGetEnv("TR_UseOutOfDatedRegRevMoveList");
```
0xdaryl: I do take exception to a couple of spelling mistakes in the environment variable itself.  Shouldn't the spelling be `TR_UseOutOfDateRegRegMoveList` or am I completely missing the point of this env var?

Rather than `env`, why not call the auto `useOutOfDateRegRegMoveList` to make it explicit?

ncough: I am not certain about the original point of the env var, but based on the context the spelling you propose seems valid.

charliegracie:
genie-omr build

charliegracie:
genie-omr build

PR #1680, opened 2017-09-18
Ignore irrelevant stores in local DSE

ncough:
Local DSE will build up a list of stores to check
as it progresses through the block. Due to the
large number of dead stores in OSR blocks, this
can result in a significant compile time increase.

### DIFF
--- a/compiler/optimizer/LocalDeadStoreElimination.cpp
+++ b/compiler/optimizer/LocalDeadStoreElimination.cpp
@@ -754,7 +754,7 @@ void TR::LocalDeadStoreElimination::adjustStoresInfo(TR::Node *node, TR_BitVecto
    // being stored into as unused so that any dead store encountered above
    // can be removed if it does not feed a use.
    //
-   if (node->getOpCode().isStore())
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM

compiler/optimizer/LocalDeadStoreElimination.cpp
```diff
@@ -754,7 +754,7 @@ void TR::LocalDeadStoreElimination::adjustStoresInfo(TR::Node *node, TR_BitVecto
    // being stored into as unused so that any dead store encountered above
    // can be removed if it does not feed a use.
    //
-   if (node->getOpCode().isStore())
+   if (node->getOpCode().isStore() && !node->chkStoredValueIsIrrelevant())
```
vijaysun-omr: Should this be checking !node->storedValueIsIrrelevant() instead ?

vijaysun-omr: To make it consistent with what every other place that checks for irrelevant stores is using...

ncough: This has been updated to use `storedValueIsIrrelevant`, requiring the explicit check of `isAutoOrParm` as well.

vijaysun-omr:
Thanks for the change. Merging since checks have passed.

PR #1681, opened 2017-09-18
Use vector rather than hash table in CompactLocals

ncough:
CompactLocals frequently has to find the entry
in its interference graph for a live local index.
This was done with the interference graph's built
in hash table, however, this is expensive.
CompactLocals also had an array of symbols, indexed
by live local indices.

This change replaces the array with a vector of
interference graph nodes, allowing CompactLocals
to cheaply look up the interference graph node
for a live local index.

### DIFF
--- a/compiler/optimizer/CompactLocals.cpp
+++ b/compiler/optimizer/CompactLocals.cpp
@@ -77,7 +77,7 @@ TR_CompactLocals::TR_CompactLocals(TR::OptimizationManager *manager)
    : _liveVars(NULL),
      _prevLiveVars(NULL),
      _temp(NULL),
-     _localIndexToSymbolMap(NULL),
+     _localIndexToIGNode(NULL),
      _localsIG(NULL),
      TR::Optimization(manager)
    {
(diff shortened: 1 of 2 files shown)


charliegracie:
genie-omr build

compiler/optimizer/CompactLocals.cpp
```diff
@@ -458,32 +460,29 @@ TR_CompactLocals::createInterferenceBetween(TR_BitVector *bv)
...
 
-         sym1 = (*_localIndexToSymbolMap)[i1];
-         sym2 = (*_localIndexToSymbolMap)[i2];
+         ig1 = (*_localIndexToIGNode)[i1];
+         ig2 = (*_localIndexToIGNode)[i2];
 
-         if (sym1 && sym2 &&
-             eligibleLocal(sym1) && eligibleLocal(sym2) &&
-             !_localsIG->hasInterference(sym1, sym2))
+         if (ig1 && ig2)
```
vijaysun-omr: The notion of an eligibleLocal is no longer required ? Just checking because the code used to check for that but does not do so anymore with the change having been done.

ncough: This is safe as `_localIndexToIGNode` will only receive an entry for a symbol if it is an `eligibleLocal`. Moreover, whether it is an `eligibleLocal` will not change throughout this analysis.

vijaysun-omr:
Thanks for the response. I see how the logic is equivalent now. Merging.

PR #1682, opened 2017-09-18
Improvements to GRA compile time

ncough:
These changes reduce the compile time overhead of GlobalRegisterAllocator.
It includes the removal of a check that was no longer being used and a fix to
GRAs behaviour when identifying overlapping allocations.

The most significant change converts loops in GRA's setup into BitVector
operations. This was enabled by recent memory improvements, which allow
a RegisterCandidate's BlockInfo to specify interested blocks with a weight
of 0 using a BitVector, rather than adding new map entries. With this,
all concerns could be represented as BitVectors of block numbers, except
for GlobalSet, which had to be pivoted from BitVectors of symbol reference
numbers to match the others.

### DIFF
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -574,7 +574,6 @@ TR_RegisterCandidate::setWeight(TR::Block * * blocks, int32_t *blockStructureWei
       if (!b) continue;
 
       TR_ASSERT(blockNumber == b->getNumber(),"blocks[x]->getNumber() != x");
-      bool hasLoadNearStart = !isExtensionOfPreviousBlock.isSet(blockNumber) && findLoadNearStartOfBlock(b, getSymbolReference());
       TR_ASSERT((blockNumber < cfg->getNextNodeNumber()) && (blocks[blockNumber] == b),"blockNumber is wrong");
       
       int32_t blockWeight = _blocks.getNumberOfLoadsAndStores(blockNumber);
--- a/compiler/optimizer/RegisterCandidate.cpp
+++ b/compiler/optimizer/RegisterCandidate.cpp
@@ -3758,7 +3758,7 @@ TR_RegisterCandidates::computeAvailableRegisters(TR_RegisterCandidate *rc, int32
             overlaps = overlapLookup->second;
             }
 
-         TR_RegisterCandidate *rc2 = b->getGlobalRegisters(comp())[i].getRegisterCandidateOnEntry();
+         TR_RegisterCandidate *rc2 = b->getGlobalRegisters(comp())[i].getRegisterCandidateOnExit();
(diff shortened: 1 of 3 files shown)


compiler/optimizer/RegisterCandidate.hpp
```diff
@@ -54,32 +54,32 @@ class GlobalSet
...
+   void collectReferencedAutoSymRefs(TR::Node *node, TR_BitVector &referencedAutoSymRefs, TR::NodeChecklist &visited);
 
    TR::Region &_region;
-   TR_BitVector _empty;
    typedef TR::typed_allocator<std::pair<uint32_t, TR_BitVector*>, TR::Region&> RefMapAllocator;
    typedef std::less<uint32_t> RefMapComparator;
    typedef std::map<uint32_t, TR_BitVector*, RefMapComparator, RefMapAllocator> RefMap;
-   RefMap _refAutosPerBlock;
+   RefMap _blocksPerAuto;
+   TR_BitVector _empty;
```
andrewcraik: if this a singleton for representing empty then I think the name should be _EMPTY which is a hint at the singleton-ness

andrewcraik:
I think the pull request message could be expanded to include a few more details to save having to read the changesets to see the overall effect of the change.

andrewcraik:
The overall diff is a bit daunting but the individual commits are LGTM modulo a small nit I made on the naming of a field.

andrewcraik:
@vijaysun-omr I think this is one in your field of expertise I think this needs cross platform testing so once updated a genie build is probably something to fire off.

ncough:
I have corrected the field naming and expanded the pull request message.

compiler/optimizer/RegisterCandidate.cpp
```diff
@@ -3758,7 +3758,7 @@ TR_RegisterCandidates::computeAvailableRegisters(TR_RegisterCandidate *rc, int32
             overlaps = overlapLookup->second;
             }
 
-         TR_RegisterCandidate *rc2 = b->getGlobalRegisters(comp())[i].getRegisterCandidateOnEntry();
+         TR_RegisterCandidate *rc2 = b->getGlobalRegisters(comp())[i].getRegisterCandidateOnExit();
```
vijaysun-omr: Good catch. Is this a recent bug related to the compile time improvements OR was it just a pre-existing functional bug you are fixing now ?

ncough: This was a pre-existing functional bug I encountered when trying to reduce the compile time in `computeAvailableRegisters`, using the `GlobalSet` to shortcut some of the checks. The reduction was not significant enough to warrant the additional complexity.

ncough:
genie-omr build all

vijaysun-omr:
Since my comments have been addressed I am merging this now.

PR #1683, opened 2017-09-19
Fixing heap top address calculation

dmitripivkine:
- The top heap address calculation should use aligned start address.
Otherwise a region aligned heap size might be shorter by region.

- Concurrent Scavenger code in Memory Manager should not step back by
tail padding as far as top address does not include it. tailpadding info
has been removed from Modron Assertion message

- Modron Assertions has been improved to work in startup time in case if
OMR VM Thread is not initialized. Tracepoint output should be skip in
this case.  


youngar:
genie-omr build

PR #1691, opened 2017-09-19
GC: Update the GC Forge Allocator API

rwy0717:
* Do not take an environment in initialize/shutdown. The forge is
  typically initialized before any environments are attached.
* Wrap the forge in the OMR namespace. Provide a backwards compatible
  typedef. Do not put MemoryStatistic or AllocationCategory inside the
  OMR namespace just yet, they are used in C code.
* Provide a set of new/delete operators for wrapping forge allocations.
  Now you can pass the forge as a parameter of the new operator. This
  should simplify the newInstance idiom used throughout the GC.

/cc @youngar @charliegracie @ktbriggs

### DIFF
--- a/gc/base/Forge.cpp
+++ b/gc/base/Forge.cpp
@@ -25,20 +25,23 @@
 #include "omrcomp.h"
 #include "EnvironmentBase.hpp"
 
-typedef struct MM_MemoryHeader {
+namespace OMR {
+namespace GC {
+
+struct MemoryHeader {
 	uintptr_t allocatedBytes;
 	MM_AllocationCategory::Enum category;
-} MM_MemoryHeader;
+};
 
-typedef union MM_AlignedMemoryHeader {
+union AlignedMemoryHeader {
 	double forAlignment;
-	MM_MemoryHeader header;
-} MM_AlignedMemoryHeader;
+	MemoryHeader header;
+};
 
 bool
-MM_Forge::initialize(MM_EnvironmentBase* env)
+Forge::initialize(OMRPortLibrary* port)
 {
-	_portLibrary = env->getPortLibrary();
+	_portLibrary = port;
 
 	if (0 != omrthread_monitor_init_with_name(&_mutex, 0, "MM_Forge")) {
 		return false;
@@ -54,7 +57,7 @@ MM_Forge::initialize(MM_EnvironmentBase* env)
 }
 
 void 
-MM_Forge::tearDown(MM_EnvironmentBase* env)
+Forge::tearDown()
 {
 	_portLibrary = NULL;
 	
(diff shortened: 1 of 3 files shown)


genie-omr:
Can one of the admins verify this patch?

genie-omr:
Can one of the admins verify this patch?

charliegracie:
genie-omr build

charliegracie:
@rwy0717 I think this is a great change.  Historically we supported platforms whose C++ compilers did not work properly with this style of code but I believe those compilers have all been fixed or we no longer support the platform :).

The current change should not cause any downstream consumers compile issues as far as I can tell.  Do you believe this to be correct as well?

lmaisons:
Two things worth considering:
1) Specifically for the operators new: If I remember correctly, the GC compiles with '-fno-exceptions'.  To get `operator new` to return `nullptr` without invoking the supposedly allocated object's constructor with `this == nullptr`, you need to mark them as `throw()`

2) If OMR is going to be used as namespace project wide (and there's no great reason it shouldn't be), we should probably be having a discussion about how we want to avoid name collisions.

lmaisons:
Created an issue for the shared namespace in #1706

charliegracie:
I am going to spend some more time dijesting these changes and do some reading to make sure I understand them fully.

rwy0717:
OK, awesome writeup :D  I may end up moving the operator new/delete changes to a separate PR where we can figure out all the details, but for now:

Re 1:

If we want, I can move the forge to `OMR::GC::*`, since this is a GC-specific class. We can talk in the new issue about where GC code should live, but I think either GC or MM is a natural fit.

Re 2:

Yep, the GC builds with `-fnoexceptions`. My thinking is that Its safe to leave out `throw()`, since that option will cause the new operation to check the pointer before constructing. I can mark the new operators as `throw()`  to be safe.

I also think that the Forge has the potential to be a general purpose allocator. But if we start to use the forge outside of the GC, exceptions *may* be enabled. Would there be any interest in providing both a throwing and nothrow set of new operators? The nothrow variant would look like this:

```c++
/// Takes a std::nothrow_t tag
/// an unused argument "tagging" the operator new as a throw() variant.

void*
operator new(std::size_t size, Forge& forge, AllocationCategory::Enum category, const char* callsite, const std::nothrow_t&) throw();
```

I'm not sure how to gracefully degrade the throwing operator new into a nothrow implementation when exceptions are disabled. Possibly we could use `#ifdef __cpp_exceptions` to completely disable the throwing API. The GC would only use the non-throwing version.

While we're at it, I'd also like to settle on either providing a `Forge*` or `Forge&` API, and not implement both.

Re 3:

I left out the scalar APIs for the time being, and will add them as a later patch.

Re 4/5:

Good points, I'll add throw() to the delete operators.

rwy0717:
I pushed a new version with the following changes:
1. Move `Forge` to the `OMR::GC` namespace
2. Remove the operator delete code -- they'll never be called.
3. Update the operator new calls to be `throw()` and take a `std::nothrow` argument.
4. Add an `operator new[]`

charliegracie:
genie-omr build

rwy0717:
Rebased PR.

charliegracie:
genie-omr build all

PR #1695, opened 2017-09-19
Add missed _inlines header to PartialRedundancy.hpp

mgaudet:
Fixes a warning from GCC

    In file included from ../compiler/optimizer/Optimization.hpp:25:
    ../compiler/optimizer/OMROptimization.hpp:123:31: warning: inline function 'OMR::Optimization::trace' is not defined [-Wundefined-inline]
       inline bool                trace();
                                  ^
    ../compiler/optimizer/PartialRedundancy.hpp:207:46: note: used here
       bool trace() { return _partialRedundancy->trace(); }
                                                 ^


0xdaryl:
genie-omr build

PR #1700, opened 2017-09-20
GC: Don't use the GC in omrcore when OMR_GC is OFF

rwy0717:
This fixes some issues in base9, where we want to use only jitbuilder, but OMR can't compile with the GC disabled.

/cc @dnakamura @youngar


charliegracie:
genir-omr build

charliegracie:
genie-omr build

charliegracie:
genie-omr build

rwy0717:
I had forgotten to `#include "omrcfg.h"`

PR #1701, opened 2017-09-20
Refactor double/float remainder helper

r30shah:
Function floatRemHelper is used to generate code for node drem/frem.
Generated assembly uses divide-to-integer instructions DIxBR to get
remainder of two floating point numbers. In rare case there is a
call-out to helper function to get the result. This commit uses
system linkage to call the fast-path helper function in OOL section.

### DIFF
--- a/compiler/z/codegen/FPTreeEvaluator.cpp
+++ b/compiler/z/codegen/FPTreeEvaluator.cpp
@@ -66,6 +66,7 @@
 #include "z/codegen/S390GenerateInstructions.hpp"
 #include "z/codegen/S390Instruction.hpp"
 #include "z/codegen/S390OutOfLineCodeSection.hpp"
+#include "z/codegen/OMRLinkage.hpp"
 
 static TR::InstOpCode::Mnemonic getIntToFloatLogicalConversion(TR::CodeGenerator *cg, TR::InstOpCode::Mnemonic convertOpCode)
    {
@@ -1059,14 +1060,11 @@ OMR::Z::TreeEvaluator::ddivEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    return node->getRegister();
    }
 
-/**
- * Generate code for a % b for both float and double.
- * The code calls out to an assembler version of fmod that uses standard
- * JIT linkage, except that it is known that the function does not kill
- * any registers other than FPR0/FPR1/FPR2/GPR-ra
- * The call-out does not always happen - in fact, it is hopefully rare.
- * The DIxBR instruction is used to do a divide-to-integer. The call-out is
- * only done if the result is not exact.
+/** \brief Generates code for a % b for both float and double values
+ *  \details
+ *  The code uses divide-to-integer instructions DIxBR to get remainder of two double/float values. 
+ *  In rare cases when result of DIxBR is not exact it calls out to double/float remainder helper which
+ *  calls fmod to get remainder and uses System linkage.
  */
 TR::Register *
 OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
@@ -1076,15 +1074,9 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
    TR::Register * secondRegister = cg->fprClobberEvaluate(node->getSecondChild());
    TR::Register * tempRegister = cg->allocateRegister(TR_FPR);
    TR::Register * targetRegister = cg->allocateRegister(TR_FPR);
-   TR::Register * dummyRegister3 = cg->allocateRegister(TR_FPR);
-   TR::Register * dummyRegister4 = cg->allocateRegister(TR_FPR);
-   TR::Register * dummyRegister5 = cg->allocateRegister(TR_FPR);
-   TR::Register * dummyRegister6 = cg->allocateRegister(TR_FPR);
-   TR::Register * dummyRegister7 = cg->allocateRegister(TR_FPR);
    TR::LabelSymbol * labelNotExact = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
    TR::LabelSymbol * labelOK = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
-   TR::Register * raReg = cg->allocateRegister();
-
+   TR::ILOpCodes opCode = node->getOpCodeValue();
    //
    // Algorithm: c = a rem b;
    //              target = a;
@@ -1093,6 +1085,8 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
    //              if (target is +0 AND a is < +0) then target is -0;
    // The check for the special case of 0's should be on rare path code, not in the mainline
 
+   // callNode is used as dangling node used to pass to buildSystemLInkageDispatch to generate a call to fmod helper
+   TR::Node *callNode = NULL;
 
    if (node->getDataType() == TR::Float)
       {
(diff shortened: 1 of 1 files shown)


r30shah:
@fjeremic Can you please review this PR. I made a change in FPTreeEvaluator to use System linkage from OMR for generating a call-out dispatch sequence.

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -1061,9 +1062,8 @@ OMR::Z::TreeEvaluator::ddivEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
 /**
  * Generate code for a % b for both float and double.
- * The code calls out to an assembler version of fmod that uses standard
- * JIT linkage, except that it is known that the function does not kill
- * any registers other than FPR0/FPR1/FPR2/GPR-ra
+ * The code calls out to double/float remainder helper which
+ * calls fmod to get remainder and uses System linkage.
```
fjeremic: Can we properly Doxygen this function while we are in this area of code?

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -1107,6 +1101,7 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
       generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BE, node, labelOK);        // it is not < +0
       generateRRInstruction(cg, TR::InstOpCode::LCEBR, node, targetRegister, targetRegister); // negate answer to be -0
       generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, labelOK);        // it is not < +0
+      TR::Node::recreate(node, TR::fcall);
```
fjeremic: I do not believe this is the best way to achieve what you want. When you recreate a node the underlying infrastructure will remove all non-applicable facts you know about this IL. For example if you knew this particular node was always non-negative and the node was marked as such the `recreate` API may wipe out this property since it is not applicable to calls (this is hypothetical to illustrate the point; not sure if it actually happens for the non-negative flag but surely it does for others). Then when you `recreate` again later you've lost a piece of information.

To achieve what you want here it's best to create a new node if the infrastructure for `buildSystemLinkageDispatch` does not support a variadic call where you can pass arguments directly. This way you also avoid the hairiness of the `unsetRegister` hack below.

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -1121,45 +1116,36 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+   TR::Register *helperReturnRegister = linkage->buildSystemLinkageDispatch(node);
+   // In addition to returning the result register, above dispatch call also set node with result register.
+   // Main line sets result in targetRegister so in OOL we copy helperReturnRegister to targetRegistert and unset node register.
+   generateRRInstruction(cg, node->getDataType() == TR::Float ? TR::InstOpCode::LER : TR::InstOpCode::LDR, node,  targetRegister, helperReturnRegister);
+   node->unsetRegister();
+   cg->stopUsingRegister(helperReturnRegister);
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, labelOK);
+   outlinedSlowPath->swapInstructionListsWithCompilation();
+
+   TR::RegisterDependencyConditions * postDeps = generateRegisterDependencyConditions(0, 4, cg);
```
fjeremic: Can we mark the ICF in which this post dependency is being used with `setStartInternalControlFlow` / `setEndInternalControlFlow`?

fjeremic:
@r30shah do the helper functions exist in OMR? For example if one were to create a Tril test in which we take the OOL path will it even be functionally correct?

It seems we are calling the native C implementation directly now. Is there a need for the helper glue in Math.s to exist?

r30shah:
@fjeremic I does not exist in the OMR layer. So if someone uses tril to generate drem/frem with a test case that forces us to take OOL path, it won't work. It is a bug which I believe still exists with this change. 
We do not need floating/double remainder helper glue in Math.s . We can get rid of those.

r30shah:
@fjeremic Following your suggestions on instead of recreating a node that might cause a loss of information on original node, I am calling `buildSystemLinkageDispatch` using a new dangling node to generate call out sequence.

compiler/z/codegen/FPTreeEvaluator.cpp
r30shah: We are using OOL path to generate a call out sequences so there is no ICF in OOL and in mainline we postDeps are attached to `labelOK` which is done Label for this node. This seemed to me like a simple case of ICF and I thought attaching postDeps to `labelOK` is sufficient.  Do we still need to mark instructions as setStart/EndInternalControlFlow ?

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -1125,14 +1127,13 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
    outlinedSlowPath->swapInstructionListsWithCompilation();
    
    generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, labelNotExact);
+   callNode->setChild(0, node->getFirstChild());
+   callNode->setChild(1, node->getSecondChild());
```
fjeremic: You can fold these two lines into `createWithSymRef` above via the following API:
https://github.com/eclipse/omr/blob/3271be89f2df16683616fe0e38c85693cea982bb/compiler/il/OMRNode.hpp#L187

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -1125,14 +1127,13 @@ OMR::Z::TreeEvaluator::floatRemHelper(TR::Node * node, TR::CodeGenerator * cg)
...
-   node->setSymbolReference(cg->symRefTab()->findOrCreateRuntimeHelper(node->getOpCodeValue() == TR::fcall ? TR_S390jitMathHelperFREM : TR_S390jitMathHelperDREM, false, false, false));
-   TR::Register *helperReturnRegister = linkage->buildSystemLinkageDispatch(node);
-   // In addition to returning the result register, above dispatch call also set node with result register.
-   // Main line sets result in targetRegister so in OOL we copy helperReturnRegister to targetRegistert and unset node register.
+   TR::Register *helperReturnRegister = linkage->buildSystemLinkageDispatch(callNode);
    generateRRInstruction(cg, node->getDataType() == TR::Float ? TR::InstOpCode::LER : TR::InstOpCode::LDR, node,  targetRegister, helperReturnRegister);
-   node->unsetRegister();
-   cg->stopUsingRegister(helperReturnRegister);
+   callNode->setReferenceCount(1);
+   cg->decReferenceCount(callNode);
```
fjeremic: Why do we need this? Is the intent to decrement the reference count of the call's children? If so we should do this explicitly so it is obvious to the reader.

r30shah: No. We do not recursively decrease reference count of the node here. Reason behind the setting ref count to one and decrease it is because, nativeDispatch sets the node with the return register, by artificially setting ref count to 1 and decreasing it I assume will make sure that helperReturnRegister is no longer live.

compiler/z/codegen/FPTreeEvaluator.cpp
r30shah: This API increases the ref count of the child as well. The use of older API makes sure that we do not mess up with the nodes from original trees as newly created node is used to just generate a call using nativeDispatch.

compiler/z/codegen/FPTreeEvaluator.cpp
fjeremic: Ok, I understand. @r30shah this is not so obvious though just looking at the code. Would you agree it may be a good idea to leave a short one sentence comment as to why we are doing this?

fjeremic:
You may need to squash the changes before a commiter can merge.

r30shah:
@fjeremic  Thanks for reviewing this. I have squashed all commits.

Issue #1604, opened 2017-09-05
Add CMake option for tril


mgaudet:
This should allow us to build just JitBuilder and not Tril.

PR #1704, opened 2017-09-21
CMake compiler fixes

charliegracie:
- Make tril find_package REQUIRED
- Fix -pthread warning while linking compiletest


mgaudet:
LGTM, though makes #1604 a higher priority IMO

mgaudet:
genie-omr build

Issue #1707, opened 2017-09-21
OMR no longer compiles with Xcode 9


mgaudet:
Building with Xcode 9 fails in a couple of different places. 

Here's my configuration: 

```
13:20 $ cmake -C ../cmake/caches/Travis.cmake -GNinja  .. 
loading initial cache file ../cmake/caches/Travis.cmake
-- Starting with CMake version 3.9.0
CMake Warning at CMakeLists.txt:28 (message):
  CMake support is currently experimental.

  Various components may fail to compile with default configuration options,
  and some are not even attempted.


-- The CXX compiler identification is AppleClang 9.0.0.9000037
-- The C compiler identification is AppleClang 9.0.0.9000037
-- Check for working CXX compiler: /usr/local/opt/ccache/libexec/c++
-- Check for working CXX compiler: /usr/local/opt/ccache/libexec/c++ -- works
<snip>
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/magaudet/open/xcode_fail/build
✔ ~/open/xcode_fail/build [xcode_fail L|✔] 
```

DDR Compilation failure
------------------------

```
FAILED: ddr/lib/ddr-blobgen/CMakeFiles/omr_ddr_blobgen.dir/java/genBinaryBlob.cpp.o 
/usr/local/opt/ccache/libexec/c++  -DJ9HAMMER -DOSX -DUT_DIRECT_TRACE_REGISTRATION -D_FILE_OFFSET_BITS=64 -I. -I.././include -I.././include_core -I.././third_party -I../ddr/include -fPIC   -pthread -m64 -std=gnu++11 -MD -MT ddr/lib/ddr-blobgen/CMakeFiles/omr_ddr_blobgen.dir/java/genBinaryBlob.cpp.o -MF ddr/lib/ddr-blobgen/CMakeFiles/omr_ddr_blobgen.dir/java/genBinaryBlob.cpp.o.d -o ddr/lib/ddr-blobgen/CMakeFiles/omr_ddr_blobgen.dir/java/genBinaryBlob.cpp.o -c ../ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
In file included from ../ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp:26:
In file included from ../ddr/include/ddr/std/sstream.hpp:39:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/sstream:174:
[...]

PR #1708, opened 2017-09-21
Correct type error in GC

mgaudet:
Corrects a type error flagged by AppleClang 9

Related issue is #1707.


charliegracie:
Looks good to me. I am surprised that was not a warning on every compiler.

charliegracie:
genie-omr build

PR #1709, opened 2017-09-21
Convert CompareAndSwap and PatchFence to Intrinsics

0dvictor:
Migrated assembly utilities CompareAndSwap and PatchFence16 to intrinsics.

### DIFF
--- a/compiler/x/runtime/VirtualGuardRuntime.cpp
+++ b/compiler/x/runtime/VirtualGuardRuntime.cpp
@@ -21,6 +21,7 @@
 
 #include <stdint.h>
 #include "infra/Assert.hpp"
+#include "x/runtime/X86Runtime.hpp"
 
 #define IS_32BIT_SIGNED(x)   ((x) == ( int32_t)(x))
 
@@ -47,7 +48,7 @@ extern "C" void _patchVirtualGuard(uint8_t *locationAddr, uint8_t *destinationAd
       // Self-loop
       *(uint16_t*)locationAddr = 0xfeeb;
 
-      _patchingFence16(locationAddr);
+      patchingFence16(locationAddr);
 
       // Bytes 2-4
       locationAddr[2] = (displacement >> 8);
@@ -55,7 +56,7 @@ extern "C" void _patchVirtualGuard(uint8_t *locationAddr, uint8_t *destinationAd
       locationAddr[4] = (displacement >> 24);
 
       // Bytes 0-1; unlock the self-loop/JMP+3
-      _patchingFence16(locationAddr);
+      patchingFence16(locationAddr);
       *(uint16_t*)locationAddr = 0xe9 + ((displacement & 0xff) << 8);
       }
    }
--- a/compiler/x/runtime/X86Runtime.hpp
+++ b/compiler/x/runtime/X86Runtime.hpp
@@ -25,11 +25,14 @@
 #include "env/ProcessorInfo.hpp"
 
 #ifdef WINDOWS
-#include <intrin.h> // for __cpuid, __cpuidex
+#include <intrin.h>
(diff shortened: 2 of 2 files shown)


charliegracie:
x86 only change so no reason to run the Linux PPCLE and 390 testing via the genie

andrewcraik:
looks reasonable to me

compiler/x/runtime/VirtualGuardRuntime.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include <stdint.h>
 #include "infra/Assert.hpp"
+#include "X86Runtime.hpp"
```
0xdaryl: Please precede this include filename with a path stem:  e.g., `runtime/X86Runtime.hpp`

compiler/x/runtime/X86Runtime.hpp
```diff
@@ -71,4 +74,30 @@ inline bool jitGetCPUID(TR_X86CPUIDBuffer* pBuffer)
...
+inline bool InterlockedCompareAndSwap(volatile uint16_t* ptr, uint16_t old_val, uint16_t new_val)
+   {
+#ifdef WINDOWS
+   return old_val == (uint16_t)_InterlockedCompareExchange16((volatile short*)ptr, (short)new_val, (short)old_val);
+#else
+   return __sync_bool_compare_and_swap(ptr, old_val, new_val);
+#endif
+   }
+
+inline void X86PatchingFence16(void* addr)
```
0xdaryl: Why did you call this `X86PatchingFence16` as opposed to simply `patchingFence16`?  It just feels a little awkward.

compiler/x/runtime/VirtualGuardRuntime.cpp
```diff
@@ -47,15 +48,15 @@ extern "C" void _patchVirtualGuard(uint8_t *locationAddr, uint8_t *destinationAd
       // Self-loop
       *(uint16_t*)locationAddr = 0xfeeb;
 
-      _patchingFence16(locationAddr);
```
0xdaryl: Can you now delete the assembly proc `_patchingFence16` ?

compiler/x/runtime/X86Runtime.hpp
```diff
@@ -71,4 +74,30 @@ inline bool jitGetCPUID(TR_X86CPUIDBuffer* pBuffer)
       }
    }
 
+inline bool InterlockedCompareAndSwap(volatile uint32_t* ptr, uint32_t old_val, uint32_t new_val)
```
0xdaryl: I'm not overly keen on the `Interlocked` prefix to this function despite the fact that the Microsoft compiler uses it.  My main objection is that it's a term that is foreign to the Testarossa compiler technology and hasn't been used before in this context.  `Atomic` is probably more appropriate (e.g., `AtomicCompareAndSwap`).

0dvictor: Good suggestion, I've updated the name.

compiler/x/runtime/X86Runtime.hpp
0dvictor: I actually like the name `patchingFence16` more. It has been updated.

compiler/x/runtime/VirtualGuardRuntime.cpp
0dvictor: Updated.

compiler/x/runtime/VirtualGuardRuntime.cpp
0dvictor: Probably not right now, as downstream projects may still use it. However, I'll clean it up as soon as downstream projects adopt the new intrinsic-based version.

PR #1710, opened 2017-09-21
Revert back heap top calculation

dmitripivkine:
Use unaligned address for heap start and tolerate the fact that heap
might be shorter by region in case of unfortunate alignment. Shorter by
region heap is allowed by logic.


youngar:
genie-omr build

Issue #1688, opened 2017-09-19
Make Tril and Tril's tests 'backend' agnostic


mgaudet:
The OMR compiler technology by itself is not a complete artifact. As a result of this, a testing framework like Tril needs to be able to hook itself into a completed artifact. In a very real sense, this is is Tril acting as a compiler front-end, plugging into one a completed artifact as a backend.

Currently, the backend is JitBuilder. However, Tril was designed to support multiple backends (and conceptually, could host them simultaneously, though building would be a challenge). Unfortunately, because that support was more theoretical than practical, what has happened is that the concrete backend selection has leaked, and so you'll find tests riddled with references to the `Tril::JitBuilderCompiler`. 

The question is ultimately how to pursue this goal: 

* Should the selection of backend be compile time only? 
* Should we support multiple backends in one build (and would that work with the compiler technology, or would colliding symbols cause ruin). 

My initial thoughts are that we need to replace `#include "jitbuilder_compiler.hpp"` with  `#include "default_compiler.hpp"`, and `Tril::DefaultCompiler`, then setup DefaultCompiler to alias with `Tril::JitBuilderCompiler`

Leonardo2718:
Being able to support different compiler backends is indeed an important requirement for Tril.

For simplicity, I would start by only supporting using a single backend at a time and selecting it at compile time.

As a first step to support running Tril-based tests using different backends, we can add a `virtual Tril::MethodCompiler& GetTrilCompiler()` method (maybe `*` instead of `&`) to `TRTest::JitTest`; the base test fixture for all Tril-based compiler tests. This method can return a concrete instance of the `Tril::MethodCompiler` interface (e.g. `Tril::JitBuilderCompiler`). Then, instead of explicitly creating an instance of `Tril::JitBuilderCompiler`, tests can simply call `TRTest::JitTest::getTrilCompiler()` and use whatever instance is returned. Different implementations of the method can be guarded with `#if defined()` or `#elif defined()`. Command line arguments to the compiler can then be adjusted to select different backends.

A more sophisticated approach could use the extensible class mechanism used by the compiler technology (or something similar).

Actually doing this work will likely require some redesign however. So, as part of this issue, I think it is important to consider what the *expectations* of the `Tril::MethodCompiler` interface are. These need to be the common properties of all backends that Tril will support. A non-exhaustive list:

- What is a backend required to do when `Tril::MethodCompiler::compile()` is called?
[...]

PR #1711, opened 2017-09-21
Use Test Compiler as a Tril Backend

mgaudet:
An implementation for #1688 

Notes I'd like to call out: 

* ~This adds a header to the root of the compiler directory, called `Jit.hpp`, that mirrors the one in `jitbuilder/release/include`. This is a **controversial decision** as we've no had headers at that level before. The header exports a simple interface that I think most compilers might be able to meet, _but_ OpenJ9 may be a challenge.~
* Changes from `JitBuilderCompiler`, to `DefaultCompiler`, but interestingly, since both the test compiler and JitBuilder can meet the `Jit.hpp` interface, this allows us to create a concrete backend that's shared called `SimpleCompiler.` To use Tril with another backend one could connect up that backend by modifying Default.hpp.

### DIFF
--- a/fvtest/tril/tril/simple_compiler.hpp
+++ b/fvtest/tril/tril/simple_compiler.hpp
@@ -29,15 +29,15 @@ namespace TR { class IlVerifier; }
 namespace Tril {
 
 /**
- * @brief Concrete realization of MethodCompiler that uses JitBuilder for compilation
+ * @brief Concrete realization of MethodCompiler 
  */
-class JitBuilderCompiler : public Tril::MethodCompiler {
+class SimpleCompiler : public Tril::MethodCompiler {
     public:
-        explicit JitBuilderCompiler(const ASTNode* methodNode)
+        explicit SimpleCompiler(const ASTNode* methodNode)
             : MethodCompiler{methodNode} {}
 
         /**
-         * @brief Compiles the Tril method using JitBuilder
+         * @brief Compiles the Tril method
          * @return 0 on compilation success, an error code otherwise
          */
         int32_t compile() /* override */ ;
--- a/fvtest/compilertest/CMakeLists.txt
+++ b/fvtest/compilertest/CMakeLists.txt
@@ -70,6 +70,10 @@ create_omr_compiler_library(
 		PROD_WITH_ASSUMES
 		INCLUDES ${OMR_ROOT}/fvtest
 )
+
+# Export paths for dependent objects
+make_compiler_target(testcompiler INTERFACE COMPILER testcompiler)
+
 add_executable(compilertest
 	tests/main.cpp
 	tests/BarIlInjector.cpp
--- a/fvtest/compilertest/Jit.hpp
+++ b/fvtest/compilertest/Jit.hpp
@@ -1,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2017 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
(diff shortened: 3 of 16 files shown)


Leonardo2718:
So, this approach is a bit different than what I had in mind. It uses more of a statically-polymorphic approach than what I described in the [issue](https://github.com/eclipse/omr/issues/1688#issuecomment-331250892), which is more dynamically polymorphic.

The original intention of the `Tril::MethodCompiler` interface was precisely to avoid running into issues with `Jit.hpp` (as well as others). It is up to users to define how the backend's interface maps to the Tril compilation interface. So whether or not a backend provides a `Jit.hpp` file becomes less relevant.

This may complicate building a little, but it should still work as long as different backends can be linked separately. Technically, Tril doesn't need to be linked against JitBuilder as long as `Tril::JitBuilderCompiler` is not used.

Need to do more thinking on this.

mgaudet:
So, this is WIP at least partially because it needs feedback. 

Further thinking over the weekend has suggested to me that I'd rather actually provide a `Jit.hpp` inside the `fvtest/compilertest/` directory than `compiler/` 

Jit.hpp is pretty fundamental to the rest of the test-case rigging in Tril -- while Tril itself may be separated from it, it's hard to pull it out of all the tests.

Leonardo2718:
> Jit.hpp is pretty fundamental to the rest of the test-case rigging in Tril -- while Tril itself may be separated from it, it's hard to pull it out of all the tests.

I'm not sure I see the dependence. It should be possible to move the implementation of `TRTest::JitTest::SetUpTestCase()` and `TRTest::JitTest::TearDownTestCase()` into a separate cpp file. Then, `JitTest.hpp` shouldn't need to include `Jit.hpp`. Am I missing something?

mgaudet:
Rebased this, and removing WIP tag. 

The bit that I think is missed here is that even if the `Jit.hpp` was in a different CPP file, it'd still be necessary in `fvtest/compilertriltest`, and thus you end up with some level of 'backend awareness' in the tests. 

I think I'd prefer to open a follow up issue about further making Trip backend agnostic; IMO, this is good to go as a preliminary implementation. 

Follow up work I think would be to provide some sort of Tril initialization code that moves the backed knowledge into `fvtest/tril`, which is where it ought to be IMO.

Leonardo2718:
So, I think that with currently available information, both solutions discussed are equally valid. So I'm okay with this change.

> Follow up work I think would be to provide some sort of Tril initialization code that moves the backend knowledge into `fvtest/tril`, which is where it ought to be IMO.

I'm not sure I completely agree with this. Originally, backend knowledge was intentionally not put in Tril. The theory here is that Tril should only worry about IL generation and not care what happens to IL afterwards. Instead, an interface is defined for handing over generated IL to a backend. This leaves it up to the "backend implementer" to define how to it's going to work with Tril using the provided interface. 

Conceptually, this does not prevent Tril-based tests from also being backend agnostic, since a similar approach can be used to further deffer backend definition.

Admittedly, this is more of an aspiration as Tril is not currently so decoupled from the backend. It is nevertheless, IMO, a valid goal worth aiming for.

All this said, I can see there are big technical challenges to be overcome. The major problem that comes to mind is the fact that a user of the OMR compiler (e.g. OpenJ9) is able to not only extend but also override the IL implementation, which complicates generic serialization.

In any case, this is likely better left for a later discussion.

mgaudet:
@Leonardo2718 I guess the thing I find mildly distasteful is that Tril _tests_ are currently 'informed' of the backend via the initialization and teardown requirements. 

Having said that, I don't think it's a critical thing to fix now, and in this PR. 

Rebased to address merge conflicts. @0xdaryl I think this is ready to go, assuming a find 👍 from Leo.

Leonardo2718:
> Having said that, I don't think it's a critical thing to fix now, and in this PR.

Agreed.

👍 for the change.

mgaudet:
@0xdaryl ping?

mgaudet:
This needs to be rebased to adjust new tests. After that, this would be good to get in. Too hard to keep up to date.

mgaudet:
Marking as WIP, as I'll rebase Monday.

mgaudet:
Rebased: 

@0xdaryl assuming CI passes, this is ready to go.

mgaudet:
@0xdaryl Any notes on this one?

fvtest/tril/tril/simple_compiler.hpp
```diff
@@ -29,15 +29,15 @@ namespace TR { class IlVerifier; }
 namespace Tril {
 
 /**
- * @brief Concrete realization of MethodCompiler that uses JitBuilder for compilation
+ * @brief Concrete realization of MethodCompiler 
  */
-class JitBuilderCompiler : public Tril::MethodCompiler {
+class SimpleCompiler : public Tril::MethodCompiler {
```
0xdaryl: Perhaps I'm being obtuse, but can you explain what's wrong with calling this a JitBuilderCompiler?  Is it because a "SimpleCompiler" is just one kind of an instance of a JitBuilder-based compiler?  I'm just trying to connect the dots...

mgaudet: The distinction is that this is a concrete compiler implementation that can be satisfied by both the test compiler and JitBuilder backends -- it's not directly related to JitBuilder anymore.

mgaudet:
Rebased to address conflicts, and addressed review comments.

0xdaryl:
genie-omr build all

charliegracie:
genie-omr build zlinux

PR #1712, opened 2017-09-21
Correct allocator types

mgaudet:
The allocator types were missing consts, however, only recently did
compilers start checking this (see #1707 for details).

### DIFF
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -438,7 +438,7 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
 
       auto mapAllocator = getTypedAllocator<std::pair<TR::TreeTop*, TR::TreeTop*> >(comp->allocator());
 
-      std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop*, TR::TreeTop*>, TR::Allocator> >
+      std::map<TR::TreeTop*, TR::TreeTop*, std::less<TR::TreeTop*>, TR::typed_allocator<std::pair<TR::TreeTop* const, TR::TreeTop*>, TR::Allocator> >
          currentTreeTopToappendTreeTop(std::less<TR::TreeTop*> (), mapAllocator);
 
       TR_BitVector *unAnchorableAloadiNodes = comp->getBitVectorPool().get();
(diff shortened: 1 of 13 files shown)


mgaudet:
WIP because this needs to be tested 1) for compatibility with OpenJ9 2) for compatibility with _older_ compilers.

rwy0717:
Built and tested locally (xcode 9, macOS Sierra) and works, thank you!!

The only nitpick I have, is that you've added the `const` to the right of the integer types:
eg: ` int32_t const` vs `const int32_t`, which I think is stylistically backwards.

mgaudet:
No longer WIP: As far as the style nitpick, I can change it, but I don't have a strong opinion on that one: @lmaisons, any strong pref (given I was working off your template in #1707)?

lmaisons:
My weak preference is to have const after the type its modifying [as it is now] for consistency, but I recognize it's generally idiomatic to have const first for non-pointers...

mgaudet:
Ok; Given I've tested in this configuration, I'd rather this go in as is, and we can discuss const location preferences elsewhere.

rwy0717:
It really is no big deal (and personally I like trailing modifiers), but I don't think it is consistent. Even in this diff, it's written `const char* const`, not `char const* const` (Sorry for the pedantry).

mgaudet:
Well, I won't be changing this till Friday at the earliest. So, I leave the merge question to a committer.

andrewcraik:
given the number of platforms where the build compilers don't currently check the const-ness and the ease with which someone may add another one of these without the const is there going to be a change to the linter to pick-up on the non-const cases so that once the clean-up is done we can be more sure that future contributions will not re-add the problem?

In general I am not a fan of const-ness due to its viral nature and the limited protection it provides. That being said if the compilers require it then we have to do it at least on these and impact on the rest of the code is limited because of where it needs to be put.

lmaisons:
I don't know about the linter, but in the meantime, perhaps we could request using the latest xcode (9) on our OS X builds as that is the platform that exposed the requirement for these changes.

charliegracie:
OS X resources are so low on travis we can only afford to have 1 build on that platform.  Since we support older versions of XCode I would prefer not to move up to a newer version of XCode there on the travis build.  If we do not test the older versions we will likely break them very quickly.

0xdaryl:
I would rather the OMR linter focus on checking for things that are unique to the design patterns and code structures within the Eclipse OMR compiler (like the rules around extensible classes), as opposed to things that could be caught by running an actual compiler or general static analyzer like Clang.  I understand the concern, but I'm not convinced (yet) the impact is great enough to warrant changing the original motivation for the OMR linter.

0xdaryl:
genie-omr build all

mgaudet:
This needn't necessarily be checked only via OS/X; It's possible this will manifest in newer Clang releases, which we could test on a linux. 

I'll open an issue to track that work

PR #1714, opened 2017-09-22
Cleanup logics related to SSE availability check 

0dvictor:
1. X86 CodeGen now properly checks testOSForSSESupport() from target.cpu.
2. Unused redundant helpers related to SSE check are deleted.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -101,8 +101,6 @@
 namespace OMR { class RegisterUsage; }
 namespace TR { class RegisterDependencyConditions; }
 
-extern "C" bool jitTestOSForSSESupport(void);
-
 // Hack markers
 #define CANT_REMATERIALIZE_ADDRESSES (TR::Compiler->target.is64Bit()) // AMD64 produces a memref with an unassigned addressRegister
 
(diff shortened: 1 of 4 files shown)


compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -281,7 +279,7 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    // 32-bit platforms must check the processor and OS.
    // 64-bit platforms unconditionally support prefetching.
    //
-   if (_targetProcessorInfo.supportsSSE() && jitTestOSForSSESupport())
+   if (_targetProcessorInfo.supportsSSE() && TR::Compiler->target.cpu.testOSForSSESupport(comp))
```
charliegracie: The only implementation of testOSForSSESupport I see in OMR unconditionally returns false.  Is that what it is supposed to do?

0dvictor: Yes, now `testOSForSSESupport()` in OMR always returns false as it does not support SSE on X86-32 yet. Each downstream project can optionally override it to return true if the project properly support SSE. However, I believe there is an on-going effort to bring full SSE support to OMR on X86-32 and `testOSForSSESupport()` should give the correct answer after that.

charliegracie: cool thanks for the info.

andrewcraik:
LGTM

compiler/x/env/OMRCPU.cpp
```diff
@@ -116,39 +116,3 @@ OMR::X86::CPU::testOSForSSESupport(TR::Compilation *comp)
...
-   //
-   uint32_t flags = self()->getX86ProcessorFeatureFlags(comp);
-   if ((flags & 0x05000000) != 0x05000000)
-      return false;
-
-   return self()->testOSForSSESupport(comp);
-   }
-
-bool
-OMR::X86::CPU::getX86SupportsTM(TR::Compilation *comp)
```
0xdaryl: Just curious why this is being deleted because its orthogonal to the purpose of this PR.

I realize it's unused and likely could be done better so I don't have any objections.  However,  if you want to roll this deletion into this PR then please amend your commit message to broaden the scope of this PR.

0dvictor: Thank-you for pointing it out.
Yes, they are being deleted because they are redundant and unused. I've updated the commit message.

0xdaryl:
genie build is not necessary as this is an x86-specific change.

PR #1715, opened 2017-09-22
Avoid inflating frequencies in block splitter

jdmpapin:
When block splitter copies a merge block, it distributes some of the frequency of the original block to the copy, based on the proportion of the predecessors' total frequency belonging to the hot predecessor. A subsequent check raises the resulting frequency if necessary to ensure that it's no less than that of the predecessor, with a static debug counter suggesting that otherwise the frequencies would be inconsistent. However, when the predecessor branches, it's expected for the new block to be less frequent.

In particular, if the predecessor is in a loop not containing the merge block, the duplicate's frequency would be raised to match frequencies within the loop, interfering with proper layout of the loop.

This commit restricts the frequency adjustment to cases in which the predecessor does not branch, i.e. cases in which it really would restore consistency.

### DIFF
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -6390,8 +6390,12 @@ int32_t TR_BlockSplitter::perform()
          int16_t coldFreq = mergeNode->getFrequency() - hotFreq;
          if (hotFreq < splitPred->getFrequency())
             {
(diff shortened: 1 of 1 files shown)


jdmpapin:
@vijaysun-omr From what I could see this was the cause of the problem running without `quickProfile` in eclipse/openj9#39

charliegracie:
genie-omr build

andrewcraik:
LGTM

vijaysun-omr:
Looks like a safe change to me in that it make the frequency setting more correct. Checks have also passed. So I am merging this.

PR #1716, opened 2017-09-22
Fix compile failure related to PRId64

charliegracie:
This resolves issues with compiling on RHEL 6 and ensures that the format macros are available.

The single format macro used in this file is known not to exist on older version of visual studio so it is defined in this file.  I updated the check to only define the macro for Windows and zOS and cause an error on other platforms so that we can find platforms that do not provide these specifiers.


charliegracie:
genie-omr build

charliegracie:
genie-omr build

charliegracie:
@youngar are you ok with this change?

Issue #1697, opened 2017-09-19
Investigate the challenges in downgrading Tril's minimum compiler version


mgaudet:
There is some desire to use Tril in contexts where the compiler may not have full C++11 support; most of the compiler technology outside Tril is capable of building in this situation. 

It would be good to know exactly what kind of changes would be required to lower the minimum compiler version, and what features we'd lose or be forced to make conditional on language support version.

fjeremic:
In particular attempting to build Tril on Linux on z with g++ 4.4.6 will not work due to missing C++11 support.

mgaudet:
When thinking about this, it might also be worthwhile to split the sizing effort in two: 1) The effort to get Tril, the parser and Ilgenerator up and running with older compilers and 2) the effort to get the test cases supported. 

It may still be supremely useful to use Tril on older build compilers, even if the cost of converting the extent test cases would be prohibitive.

Leonardo2718:
For Tril, I believe the only C++11 feature that is heavily relied upon is the `<type_traits>` standard library. Theoretically, the library could be implemented in pure C++03 (at least the parts that are used in Tril). However, doing so would require writing a fair bit of boilerplate template code.

Another C++11 feature that is somewhat used in Tril are template type aliases. Not having these would only be an inconvenience, leading to some duplicate code.

Everything else can either be worked around easily enough, or should already be supported by older compilers (e.g. `auto`).

For the Tril-based tests, in addition to the features I've mentioned above, some also use C++11 variadic templates. This will be quite tricky to replace, though it should be possible to create something that's close enough using `std::pair`.

fjeremic:
As a temporary workaround I've compiled gcc 4.8.5 on my RHEL 6 and I am able to build Tril correctly now. Interestingly enough there are some test failures for which I'll open up issues for so it was a worth while exercise.

mgaudet:
@rbanerjee is investigating downgrading the minimum supported version. We're not yet sure how complicated this will be, but are hopeful.

fjeremic:
> @rbanerjee is investigating downgrading the minimum supported version. We're not yet sure how complicated this will be, but are hopeful.

In my experience only `ast.hpp` caused problem because of the use of `using = <some template>` which we can work around with `typedef`. Best to hop onto a machine with gcc 4.4.6 and attempt to compile Tril then just fix problems as they come.

rbanerjee:
[...]

PR #1718, opened 2017-09-22
Downgrade Tril's minimum compiler version

rbanerjee:
There was some desire to use Tril in contexts where the compiler may not have full C++11 support.
This adds that support to tril and it's accompanying examples, while still retaining some of the C++11 only code (by guarding them with appropriate macros).

### DIFF
--- a/fvtest/tril/examples/mandelbrot/main.cpp
+++ b/fvtest/tril/examples/mandelbrot/main.cpp
@@ -25,7 +25,7 @@
 #include <assert.h>
 #include <stdio.h>
 
-using MandelbrotFunction = void(int32_t, int32_t, int32_t*);
+typedef void (MandelbrotFunction) (int32_t, int32_t, int32_t*);
 
 int main(int argc, char const * const * const argv) {
     assert(argc == 2);
@@ -33,7 +33,7 @@ int main(int argc, char const * const * const argv) {
 
     // parse the input Tril file
     FILE* inputFile = fopen(argv[1], "r");
-    assert(inputFile != nullptr);
+    assert(inputFile != NULL);
     ASTNode* trees = parseFile(inputFile);
     fclose(inputFile);
 
@@ -45,8 +45,8 @@ int main(int argc, char const * const * const argv) {
     assert(mandelbrotCompiler.compile() == 0);
     auto mandelbrot = mandelbrotCompiler.getEntryPoint<MandelbrotFunction*>();
 
-    constexpr auto size = 80;                   // number of rows/columns in the output table
-    constexpr auto iterations = 1000;           // number of iterations to be performed
+    const auto size = 80;                   // number of rows/columns in the output table
+    const auto iterations = 1000;           // number of iterations to be performed
     int32_t table[size][size] = {{0}};          // the output table
 
     mandelbrot(iterations, size, &table[0][0]);
--- a/fvtest/tril/tril/ast.hpp
+++ b/fvtest/tril/tril/ast.hpp
@@ -23,6 +23,7 @@
 #define AST_HPP
 
 #include "ast.h"
+#include "enable_if.hpp"
 
 #include <type_traits>
 #include <assert.h>
@@ -59,9 +60,9 @@ struct ASTValue {
     enum ASTType {Integer, FloatingPoint, String} ;
 
     // aliases for the underlying C++ types used to store the different AST types
-    using Integer_t = uint64_t;
-    using FloatingPoint_t = double;
-    using String_t = const char *;
+    typedef uint64_t Integer_t;
+    typedef double FloatingPoint_t;
+    typedef const char* String_t;
 
     /**
      * STL compatible type traits for working with AST data types
@@ -76,17 +77,11 @@ struct ASTValue {
      * is_FloatintPoint_Compatible<double>::value; // true
      * is_String_Compatible<const char*>::value;   // true
      */
-    template <typename T>
-    using is_Integer_Compatible = std::is_integral<T>;
-    template <typename T>
-    using is_FloatingPoint_Compatible = std::is_floating_point<T>;
-    template <typename T>
-    using is_String_Compatible = std::is_same<String_t, T>;
 
     // constructors
-    explicit ASTValue(Integer_t v) : _type{Integer}, next{nullptr} { _value.integer = v; }
-    explicit ASTValue(FloatingPoint_t v) : _type{FloatingPoint}, next{nullptr} { _value.floatingPoint = v; }
-    explicit ASTValue(String_t v) : _type{String}, next{nullptr} { _value.str = v; }
+    explicit ASTValue(Integer_t v) : _type{Integer}, next{NULL} { _value.integer = v; }
+    explicit ASTValue(FloatingPoint_t v) : _type{FloatingPoint}, next{NULL} { _value.floatingPoint = v; }
+    explicit ASTValue(String_t v) : _type{String}, next{NULL} { _value.str = v; }
 
     /**
      * @brief Return the contained value as the specified type
@@ -108,18 +103,19 @@ struct ASTValue {
      * b.get<double>(); // returns 4.5 as a double
      * b.get<int>();    // causes an assertion failure
      */
+
     template <typename T>
-    typename std::enable_if<is_Integer_Compatible<T>::value, T>::type get() const {
+    typename Tril::enable_if<std::is_integral<T>::value , T>::type get() const {
         assert(Integer == _type);
         return static_cast<T>(_value.integer);
     }
     template <typename T>
-    typename std::enable_if<is_FloatingPoint_Compatible<T>::value, T>::type get() const {
+    typename Tril::enable_if<std::is_floating_point<T>::value, T>::type get() const {
         assert(FloatingPoint == _type);
         return static_cast<T>(_value.floatingPoint);
     }
     template <typename T>
-    typename std::enable_if<is_String_Compatible<T>::value, T>::type get() const {
+    typename Tril::enable_if<std::is_same<String_t, T>::value, T>::type get() const {
         assert(String == _type);
         return static_cast<T>(_value.str);
     }
@@ -155,16 +151,17 @@ struct ASTValue {
      * b.isCompatibleWith<double>(); // returns true
      * b.isCompatibleWith<int>();    // returns false
      */
+
     template <typename T>
-    typename std::enable_if<is_Integer_Compatible<T>::value, bool>::type isCompatibleWith() const {
+    typename Tril::enable_if<std::is_integral<T>::value , bool>::type isCompatibleWith() const {
         return Integer == _type;
     }
     template <typename T>
-    typename std::enable_if<is_FloatingPoint_Compatible<T>::value, bool>::type isCompatibleWith() const {
+    typename Tril::enable_if<std::is_floating_point<T>::value, bool>::type isCompatibleWith() const {
         return FloatingPoint == _type;
     }
     template <typename T>
-    typename std::enable_if<is_String_Compatible<T>::value, bool>::type isCompatibleWith() const {
+    typename Tril::enable_if<std::is_same<String_t, T>::value, bool>::type isCompatibleWith() const {
         return String == _type;
     }
 
@@ -220,7 +217,7 @@ struct ASTNodeArg {
     public:
     ASTNodeArg* next;
 
-    ASTNodeArg(const char* name, ASTValue* value, ASTNodeArg* next = nullptr)
+    ASTNodeArg(const char* name, ASTValue* value, ASTNodeArg* next = NULL)
         : _name{name}, _value{value}, next{next} {}
 
     const char* getName() const { return _name; }
@@ -265,7 +262,7 @@ struct ASTNode {
     const ASTNodeArg* getArgument(int index) const {
         auto arg = _args;
 
-        while (arg != nullptr && index > 0) {
+        while (arg != NULL && index > 0) {
             arg = arg->next;
             --index;
         }
@@ -297,9 +294,9 @@ struct ASTNode {
     const ASTNodeArg* getPositionalArg(int index) const {
         auto arg = _args;
 
-        while (arg != nullptr) {
+        while (arg != NULL) {
             const auto name = arg->getName();
-            if (name == nullptr || name[0] == '\0') {
+            if (name == NULL || name[0] == '\0') {
                 if (index > 0) { --index; }
                 else { break; }
             }
@@ -315,7 +312,7 @@ struct ASTNode {
         auto arg = _args;
         auto i = 0;
 
-        while (arg != nullptr) {
+        while (arg != NULL) {
             ++i;
             arg = arg->next;
         }
--- a/fvtest/tril/tril/enable_if.hpp
+++ b/fvtest/tril/tril/enable_if.hpp
@@ -19,37 +19,20 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
  *******************************************************************************/
 
-#ifndef JITBUILDER_COMPILER_HPP
-#define JITBUILDER_COMPILER_HPP
-
-#include "method_compiler.hpp"
-
-namespace TR { class IlVerifier; } 
+#ifndef ENABLE_IF_HPP
+#define ENABLE_IF_HPP
 
 namespace Tril {
+    // Create an implementation of std::enable_if which was only
+    //  added since C++11.
+    template <bool, typename T = void>
 
-/**
- * @brief Concrete realization of MethodCompiler that uses JitBuilder for compilation
- */
-class JitBuilderCompiler : public Tril::MethodCompiler {
-    public:
-        explicit JitBuilderCompiler(const ASTNode* methodNode)
-            : MethodCompiler{methodNode} {}
+    struct enable_if {};
 
-        /**
-         * @brief Compiles the Tril method using JitBuilder
-         * @return 0 on compilation success, an error code otherwise
-         */
-        int32_t compile() /* override */ ;
-
-        /**
-         * @brief Start compilation with a verifier. 
-         * @param verifier The verifier to run. 
-         * @return 0 on complilation success, an error code or exception otherwise. 
-         */
-        int32_t compileWithVerifier(TR::IlVerifier* verifier);
-};
-
+    template <typename T>
+    struct enable_if<true, T> {
+        typedef T type;
+    };
 } // namespace Tril
 
-#endif // JITBUILDER_COMPILER_HPP
+#endif // ENABLE_IF_HPP
--- a/fvtest/tril/tril/ilgen.cpp
+++ b/fvtest/tril/tril/ilgen.cpp
@@ -45,7 +45,9 @@ class OpCodeTable : public TR::ILOpCode {
       static TR::ILOpCodes getOpCodeFromName(const std::string& name) {
          auto opcode = _opcodeNameMap.find(name);
          if (opcode == _opcodeNameMap.cend()) {
-            for (const auto& p : _opCodeProperties) {
+            for (int i = TR::FirstOMROp; i< TR::NumIlOps; i++) {
+               const auto p_opCode = static_cast<TR::ILOpCodes>(i);
+               const auto& p = TR::ILOpCode::_opCodeProperties[p_opCode];
                if (name == p.name) {
                   _opcodeNameMap[name] = p.opcode;
                   return p.opcode;
@@ -79,7 +81,7 @@ std::unordered_map<std::string, TR::ILOpCodes> OpCodeTable::_opcodeNameMap;
  * reason, special opcodes are detected using opcode properties.
  */
 TR::Node* Tril::TRLangBuilder::toTRNode(const ASTNode* const tree) {
-     TR::Node* node = nullptr;
+     TR::Node* node = NULL;
 
      auto childCount = tree->getChildCount();
 
@@ -93,7 +95,7 @@ TR::Node* Tril::TRLangBuilder::toTRNode(const ASTNode* const tree) {
          }
          else {
              TraceIL("Failed to find node for commoning (@id \"%s\")\n", id)
-             return nullptr;
+             return NULL;
          }
      }
      else if (strcmp("@common", tree->getName()) == 0) {
@@ -108,7 +110,7 @@ TR::Node* Tril::TRLangBuilder::toTRNode(const ASTNode* const tree) {
          }
          else {
              TraceIL("Failed to find node for commoning (@id \"%s\")\n", id)
-             return nullptr;
+             return NULL;
          }
      }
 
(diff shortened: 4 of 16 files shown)


fvtest/tril/tril/ast.hpp
```diff
@@ -28,6 +28,25 @@
 #include <assert.h>
 #include <cstring>
 
+#if __cplusplus <= 199711L
+const class nullptr_t {
```
rbanerjee: This is kind of silly but nullptr was only added as recent as C++11. And just using NULL seems like a very crude solution (maybe I'm missing something here).
Open to suggestions as to whether there's a more elegant workaround.

fvtest/tril/examples/mandelbrot/main.cpp
```diff
@@ -45,8 +49,13 @@ int main(int argc, char const * const * const argv) {
     assert(mandelbrotCompiler.compile() == 0);
     auto mandelbrot = mandelbrotCompiler.getEntryPoint<MandelbrotFunction*>();
 
+#if __cplusplus <= 199711L
+    const auto size = 80;                   // number of rows/columns in the output table
+    const auto iterations = 1000;           // number of iterations to be performed
+#else
     constexpr auto size = 80;                   // number of rows/columns in the output table
```
rbanerjee: Not sure why we are using constexpr here (instead of just const).

fvtest/tril/tril/ast.hpp
```diff
@@ -76,13 +101,17 @@ struct ASTValue {
      * is_FloatintPoint_Compatible<double>::value; // true
      * is_String_Compatible<const char*>::value;   // true
      */
+#if __cplusplus <= 199711L
+     // Never used for the version without C++11 support.
```
mgaudet: I'd rather flip the sense of this check, and avoid the 

```
#if FOO 
//nothing
#else
stuff
#endif
```
pattern. It can be replaced with 

```
#if !FOO
stuff
#endif
```

fvtest/tril/tril/ast.hpp
```diff
@@ -155,6 +213,21 @@ struct ASTValue {
      * b.isCompatibleWith<double>(); // returns true
      * b.isCompatibleWith<int>();    // returns false
      */
+#if __cplusplus <= 199711L
```
mgaudet: So, there's two distinctions here: 

1. The use of `enable_if` rather than `std::enable_if` 
2. The use of `std::is_integral` directly, rather than via the alias `is_Integer_Compatible` 

@Leonardo2718 Why the indirection of is_Integer_Compatible? 

@rbanerjee If we were to to standardize on `std::is_integral`, this could collapse nicely with a `using std::enable_if<T>` (syntax may be wrong).

fvtest/tril/tril/ilgen.hpp
```diff
@@ -50,7 +50,11 @@ class TRLangBuilder : public TR::IlInjector {
         TRLangBuilder(const ASTNode* trees, TR::TypeDictionary* d)
               : TR::IlInjector(d), _trees(trees) {}
 
+#if __cplusplus <= 199711L // Check compiler support for C++11.
```
mgaudet: For changes like this, since override is not mandatory, I think I'd prefer to skip the `#ifdef` and instead simply annotate the function with a comment;

```
bool injectIL() /* override */ ;
```

mgaudet: Same applies elsewhere

fvtest/tril/tril/ast.hpp
mgaudet: I'll admit to my gut not having a good feeling about what the 'right' answer here is, but, I would suggest this might be best put in another header; I don't think it makes sense here.

fvtest/tril/examples/mandelbrot/main.cpp
mgaudet: I think I'd just keep the const auto versions.

fvtest/tril/examples/mandelbrot/main.cpp
```diff
@@ -25,7 +25,11 @@
 #include <assert.h>
 #include <stdio.h>
 
+#if __cplusplus <= 199711L
```
mgaudet: I think I'd just keep the typedef, rather than #ifdef.

fvtest/tril/tril/ast.hpp
```diff
@@ -59,9 +78,15 @@ struct ASTValue {
     enum ASTType {Integer, FloatingPoint, String} ;
 
     // aliases for the underlying C++ types used to store the different AST types
+#if __cplusplus <= 199711L // Check for C++11 compiler support.
+    typedef uint64_t Integer_t;
```
mgaudet: I think I'd just keep the typedefs, rather than ifdefs. It doesn't bring extra value preserving the the `using`

fvtest/tril/tril/ast.hpp
Leonardo2718: The indirection is there mostly for two reasons:

1. In future, I was planning on making some changes to the type system in Tril, so having the indirection meant only having to change the code in one place.
2. Conceptually, I wanted there to be a distinction between the Tril type system and the C++ type system. The indirection was part of the interface for "bridging" the two, the goal being to hide as much of the details of how types are handled in the background behind a clear and precise interface.

Although I would prefer to keep the extra indirection for the sake of mental clarity, I'm okay with removing it as part of this change. In this case, I see no value in keeping both implementations, since it would defeat the goal of reducing code duplication.

fvtest/tril/tril/ilgen.cpp
```diff
@@ -45,7 +45,13 @@ class OpCodeTable : public TR::ILOpCode {
       static TR::ILOpCodes getOpCodeFromName(const std::string& name) {
          auto opcode = _opcodeNameMap.find(name);
          if (opcode == _opcodeNameMap.cend()) {
+#if __cplusplus <= 199711L
+            for (int i = TR::FirstOMROp; i< TR::NumIlOps; i++) {
+               TR::ILOpCodes p_opCode = (TR::ILOpCodes)i;
```
Leonardo2718: I would prefer `const auto p_opCode = static_cast<TR::ILOpCodes>(i);`

fvtest/tril/tril/ilgen.cpp
```diff
@@ -45,7 +45,13 @@ class OpCodeTable : public TR::ILOpCode {
...
+               OMR::OpCodeProperties &p = TR::ILOpCode::_opCodeProperties[p_opCode];
+#else
```
Leonardo2718: I would prefer `const auto& p = TR::ILOpCode::_opCodeProperties[p_opCode];`

fvtest/tril/tril/ilgen.cpp
```diff
@@ -45,7 +45,13 @@ class OpCodeTable : public TR::ILOpCode {
...
             for (const auto& p : _opCodeProperties) {
```
Leonardo2718: Not sure there's any value in having both implementations.

fvtest/tril/tril/ast.hpp
charliegracie: I would really prefer if `NULL` was used.  Whenever possible we should try to be consistent throughout the entire codebase.

rbanerjee:
@Leonardo2718 , @mgaudet 

> Although a theoretically sound approach to checking for C++11 support, in practice, using #if __cplusplus <= 199711L is not usually considered safe/portable because some compilers have __cplusplus set to 199711L even if C++11 support is incomplete."

About using the __cplusplus macro. There's actually no "correct" way of doing this it seems.

"Implementations rarely leap from one formal revision of the standard directly to the next; the implementation process generally proceeds by smaller steps. As a result, testing for a specific revision of the standard (e.g. by examining the value of the __cplusplus macro) often gives the wrong answer. Implementers generally don’t want to appear to be claiming full conformance to a standard revision until all of its features are implemented. That leaves programmers with no portable way to determine which features are actually available to them."

^ this is from https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#recs.hasinc

Which is partly why I left most of Leo's old code behind (the ones which strictly require C++11 support) while guarding them with a macro. Because at IMO,  in most cases they are doing the right thing (given the C++11 tools are actually available). 

My idea was to remove the "work arounds" once we actually have a C++11 supported build compiler. So I was kind of surprised as to why we would want to get rid of all the C++11 code altogether (cause in theory at least, they are "better"). This also kind of answers the questions about "Not sure if there's any value in having both implementations of X?". Unfortunately though, what I was unaware of was how often we updated our build compilers to begin with. 

So given it might take a while for was to switch to a build compiler that actually has C++11 support, I think it does make sense as to just get rid of things like std::enable_if, lambdas and all the usings.

 At least that's my understanding of the situation. Please do correct me if I'm wrong.

fvtest/tril/tril/ast.hpp
```diff
@@ -108,6 +137,34 @@ struct ASTValue {
      * b.get<double>(); // returns 4.5 as a double
      * b.get<int>();    // causes an assertion failure
      */
+
+#if __cplusplus <= 199711L
+    // Create create an equivalent implementation of std::enable_if.
+    template <bool, typename T = void>
+    struct enable_if {};
```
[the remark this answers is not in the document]
rbanerjee: It doesn't (at least I couldn't get it to work on GCC 4.4.6 [also, notice it says *at least as far back as GCC 4.4.6] by default).

But more importantly one of the original reasons why are doing this to begin with, to quote @fjeremic was 

>In particular attempting to build Tril on Linux on z with g++ 4.4.6 will not work due to missing C++11 support.

So instead of getting it to compile with some particular version of GCC, this patch is supposed to remove/work around the C++11 only features. And as far as I know, std::enable_if was "officially" only added since C++11.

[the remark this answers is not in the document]
mgaudet: Did you end up compiling Tril (You should check the command lines) w/o any C++11 support? Or just whatever comes out of GCC 4.4.6 (i.e., `-std=c++0x`)?

mgaudet:
Personally, I'd like to keep the supported subset that the rest of the compiler is already using -- It'd be silly to get rid of `auto` for example, given that the rest of the compiler already uses `auto`.

charliegracie:
It may be a really long time before we have a C++11 compiler available on all of our platforms. To provide a consistent environment for all platforms weave to use the LCD. 

I agree with @mgaudet on keeping the features which the rest of the compiler already uses. I wish his example was not 'auto' though as it I should one of the worst features c++ added in my opinion. It makes debugging and code inspection a nightmare but that is a bridge that was already crossed so we are stuck with it.

mgaudet:
Heh. It's easily my favourite!

fvtest/tril/tril/ast.hpp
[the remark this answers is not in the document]
rbanerjee: Can confirm, doesn't work.

rbanerjee:
ptal

fvtest/tril/tril/jitbuilder_compiler.cpp
```diff
@@ -40,6 +40,12 @@ int32_t Tril::JitBuilderCompiler::compile() {
    return compileWithVerifier(NULL);
 }
 
+#if __cplusplus <= 199711L // Check for C++11 compiler support.
```
mgaudet: Given the way the rest of this change looks, I'd just pull this, and use the lambda version universally.

rbanerjee: Sorry, but did you mean the 'version *without the lambda universally'? (Lambdas were again only added since C++11).

fvtest/tril/tril/enable_if.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef ENABLE_IF_HPP
+#define ENABLE_IF_HPP
+
+
+#if __cplusplus <= 199711L
+    // Create an equivalent implementation of std::enable_if.
+    template <bool, typename T = void>
+
+    struct enable_if {};
```
Leonardo2718: Can you put this in the `Tril` namespace?

fvtest/tril/tril/jitbuilder_compiler.cpp
mgaudet: I'm cool with this.

fvtest/tril/tril/enable_if.hpp
rbanerjee: Done.

rbanerjee:
ptal

fvtest/tril/tril/enable_if.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
+#ifndef ENABLE_IF_HPP
+#define ENABLE_IF_HPP
+
+#if __cplusplus > 199711L
```
lmaisons: This version check seems somewhat aggressive to me, I would have thought we would want something more along the lines of
```
#if (__cplusplus < 201103L) || !defined(_MSVC_LANG)
```
with `_MSVC_LANG` assuming that MSVC++2015+ is acceptable...

Do we know for sure that all partial updates to the C++ language standard on the compilers we care about either
1. Don't update the _cplusplus version or,
2. Are compatible with our requirements?

rbanerjee: @lmaisons 

This was brought up in an earlier comment, but I don't think there's any way of coming up with a combination of macros that is guaranteed to be infallible when trying to detect whether the compiler does or doesn't have C++11 support [checking `__cplusplus` by itself *theoretically should be enough] . 

>"Implementations rarely leap from one formal revision of the standard directly to the next; the implementation process generally proceeds by smaller steps. As a result, testing for a specific revision of the standard (e.g. by examining the value of the __cplusplus macro) often gives the wrong answer. Implementers generally don’t want to appear to be claiming full conformance to a standard revision until all of its features are implemented. That leaves programmers with no portable way to determine which features are actually available to them."

 [ ref- https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#recs.hasinc . I guess C++17 has `__has_cpp_attribute` which kind of does the trick, but of course that's no use to us.  ]

Right now how it's set up, it at least works for our build compiler and the newest GCC one. Not sure how much marginal benefit we might potentially be able to extract from complicating the `#if` statements and whether it's even "worth" doing that. I might be missing something here.

rbanerjee:
@mgaudet  @0xdaryl  

Think this is good to merge.

fjeremic:
Any news on this item? We have some pending work that is blocked by this.

dchopra001:
I see the following on gcc 4.4.6 on zLinux:

```[100%] Building CXX object fvtest/compilertriltest/CMakeFiles/comptest.dir/__/omrGtestGlue/argmain.cpp.o
In file included from /gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/OpCodeTest.hpp:25,
                 from /gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/ArithmeticTest.cpp:22:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:221: error: expected constructor, destructor, or type conversion before ‘T’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:222: error: expected constructor, destructor, or type conversion before ‘T’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:223: error: expected constructor, destructor, or type conversion before ‘T’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:224: error: expected constructor, destructor, or type conversion before ‘T’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:225: error: expected constructor, destructor, or type conversion before ‘T’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp: In function ‘std::vector<T, std::allocator<_CharT> > TRTest::const_values()’:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:233: error: ‘zero_value’ was not declared in this scope
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:234: error: ‘one_value’ was not declared in this scope
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:235: error: ‘negative_one_value’ was not declared in this scope
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:236: error: ‘positive_value’ was not declared in this scope
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:237: error: ‘negative_value’ was not declared in this scope
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:233: error: expected primary-expression before ‘{’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:233: error: expected ‘;’ before ‘{’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:233: error: expected primary-expression before ‘>’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:233: error: expected primary-expression before ‘)’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:234: error: expected primary-expression before ‘>’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:234: error: expected primary-expression before ‘)’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:235: error: expected primary-expression before ‘>’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:235: error: expected primary-expression before ‘)’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:236: error: expected primary-expression before ‘>’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:236: error: expected primary-expression before ‘)’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:237: error: expected primary-expression before ‘>’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:237: error: expected primary-expression before ‘)’ token
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/JitTest.hpp:242: error: expected ‘;’ before ‘}’ token
...
```

rbanerjee:
Looking into it now.

rbanerjee:
@fjeremic  So with this PR, it seems like `Tril` and the examples work fine on `zLinux` but some of the Tril Compiler Tests still need to be ported. As the original Issue #1697 suggested, should we do this in two parts?

As for this one, it takes care of getting Tril, the parser and the Ilgenerator up and running with older compilers. I think it's best we merge this and I can open up a separate PR for porting the tests, as there is a certain amount of discussion to be had there. Does that sound good?

rbanerjee:
Just to confirm, the blocked PR is #1557, right?

fjeremic:
> @fjeremic Just to confirm, the blocked PR is #1557, right?

Correct.

rbanerjee:
> So with this PR, it seems like Tril and the examples work fine on zLinux but some of the Tril Compiler Tests still need to be ported. As the original Issue #1697 suggested, should we do this in two parts?

>As for this one, it takes care of getting Tril, the parser and the Ilgenerator up and running with older compilers. I think it's best we merge this and I can open up a separate PR for porting the tests, as there is a certain amount of discussion to be had there. Does that sound good?

@fjeremic Do you have any thoughts on this part?

@0xdaryl  ptal. If there aren't any objections then I would l like to have this merged please.

fjeremic:
> @fjeremic Do you have any thoughts on this part?

Sounds reasonable to me.

0xdaryl:
omr-genie build all

0xdaryl:
genie-omr build all

charliegracie:
I would prefer to see all C++11 code completely removed as I see no win in maintaining both copies. I think this is agreeing with @0xdaryl.  Other components have sadly let C++0x extensions creep in so that is ok but until C++11 is fully supported on all platforms and is a specification level we can target I think removing it is the best course of action.

rbanerjee:
@0xdaryl  Addressed the comments. ptal.

>I would prefer to see all C++11 code completely removed as I see no win in maintaining both copies. I think this is agreeing with @0xdaryl. Other components have sadly let C++0x extensions creep in so that is ok but until C++11 is fully supported on all platforms and is a specification level we can target I think removing it is the best course of action.

@charliegracie The checking of `__cplusplus <= 199711L ` macros have now completely been removed in this patchset. Actually there were only two of them left form the last iteration. 

One for using `std::enable_if` instead of our own implementation. IIUC  using the one  in `std` doesn't really provide any added benefits. So I don't have any issues removing that one.

The other one had to do with using `std:: bind1st` and `std::ptr_fun` to "emulate" lambdas. The issue with that one being, from my understanding `std::bind1st` has been deprecated since C++11, and C++17 completely removes it. So once we upgrade our compilers, having only the implementation with `std::bind1st` should break our builds. However, don't think we are going to be moving to C++17 complaint compilers anytime in the near future. So either I can open up an issue to document this behaviour, or we can just use the macro for that one specific case?

fvtest/tril/tril/jitbuilder_compiler.cpp
```diff
@@ -51,7 +55,7 @@ int32_t Tril::JitBuilderCompiler::compileWithVerifier(TR::IlVerifier* verifier)
     auto argTypes = methodInfo.getArgTypes();
     auto argIlTypes = std::vector<TR::IlType*>{argTypes.size()};
     std::transform(argTypes.cbegin(), argTypes.cend(), argIlTypes.begin(),
-                   [&types](TR::DataTypes d){ return types.PrimitiveType(d); } );
+                   std::bind1st(std::ptr_fun(&getPrimitiveType),types));
```
mgaudet: (We discussed this in person) 

Given the simplicity of `std::transform`, IMO, rather than chasing the 'functional purity', it'd be better to just do the loop, which removes the need to have a function defined, and eliminates both `std::bind1st` and `std::ptr_fun`.

rbanerjee:
@0xdaryl  ptal. Addressed the comments and testing is done. (Removing the WIP tag.)

0xdaryl:
omr-genie build all

0xdaryl:
genie-omr build all

0xdaryl:
I believe all concerns have been addressed.  There are no compiler level macros remaining in this PR.

PR #1719, opened 2017-09-22
Cmake fixes

dnakamura:
Small misc cmake fixes. See individual commit messages for details

### DIFF
--- a/omr/CMakeLists.txt
+++ b/omr/CMakeLists.txt
@@ -59,8 +59,6 @@ if(OMR_GC)
 	target_link_libraries(omrcore
 		PRIVATE
 			${OMR_CORE_GLUE_TARGET}
-			omrtrace
-			omrgc
 	)
 
 	target_include_directories(omrcore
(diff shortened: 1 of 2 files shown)


mgaudet:
I've no objections :) LGTM 👍

rwy0717:
Lgtm

charliegracie:
No need for a CI build on pLinux and zLinux since they do not use CMake yet.

omr/CMakeLists.txt
```diff
@@ -59,8 +59,6 @@ if(OMR_GC)
 	target_link_libraries(omrcore
 		PRIVATE
 			${OMR_CORE_GLUE_TARGET}
-			omrtrace
```
charliegracie: I thought the OMR core code was responsible for calling the GC startup APIs and also used tracepoints. Doesn't that mean that it needs to link the gc and trace libraries?

dnakamura: It means the end consumer needs to link against those (if they actually end up calling said functions). However the issue is that almost everyone links against omrcore, thus making omrcore depend on anything quickly introduces cycles

PR #1721, opened 2017-09-24
Do not define _XOPEN_SOURCE globally

rwy0717:
Important bits of OSX's pthread library are disabled when this symbol is
defined. This seems to be new behaviour as of xcode 9. This fixes many
compilation errors in the C++ standard library.

This fixes one of the compilation issues in #1707.

### DIFF
--- a/include_core/omrport.h
+++ b/include_core/omrport.h
@@ -921,6 +921,9 @@ typedef struct J9MmapHandle {
 #define _XOPEN_SOURCE
 #endif /* defined(OSX) */
 #include <ucontext.h>
+#if defined(OSX)
+#undef _XOPEN_SOURCE
+#endif /* OSX */
 #endif /* !WIN32 */
 
 #if defined(J9ZOS390)


charliegracie:
@rwy0717 do we need to specify _XOPEN_SOURCE on OS X at all?  With the global value removed with your PR I only see a few instances in c/cpp files for OS X.  I did a quick google search and it looks like most of them could be removed.  Can you take a quick look?  Thanks

rwy0717:
On macOS, `_XOPEN_SOURCE` is needed to include `<ucontext.h>`. The ucontext APIs were deprecated in POSIX 2004, and removed in POSIX 2008. We are using `ucontext_t,{get,set}context` for thread introspection and signal handling, and I don't think it's going to be realistic to get rid of it.

What we can do, is try to limit how the ucontext header (and _XOPEN_SOURCE) pollutes our port library public APIs. Right now, the port library uses the raw `ucontext_t` type, which must be defined in `port.h`, so even this kind of cleanup will be non-trivial.

If it's OK with you, I'd like this to go in as-is to solve the immediate compilation failure, and I'll open an issue about limiting / hiding our usage of ucontext in the port library?

charliegracie:
Sounds ok to me. Thanks for the details

PR #1722, opened 2017-09-25
Fix maxprec() call to avoid fixed-point overflows on Z

deesebas:
The maxprec() routine that obtains current time on Z-plaforms
triggers fixed-point overflows if the PSW Fixed Point Oveflow mask
happens to be turned on. The reason for the overflow is that a subtract
operation used for calculating the current time mistakenly uses signed
rather than unsigned subtraction. The time value of the TOD-clock
(returned via STCK/STCKF instruction) will always be unsigned, so a
subtract logical instruction should be used instead of subtract - this 
would avoid Fixed-point Overflows.

The fix applies to z/OS, z/Linux and z/TPF platforms.

Signed-off-by: Deepthi Sebastian <deesebas@in.ibm.com>

### DIFF
--- a/port/linuxs390/omrrttime.s
+++ b/port/linuxs390/omrrttime.s
@@ -36,7 +36,7 @@ maxprec:
     brc 3,.LCarry
     ahi %r2,-1
 .LCarry:
-    s   %r2,.LC0-.LT0(%r1)
+    sl   %r2,.LC0-.LT0(%r1)
     srdl %r2,1
     br  %r14
 
--- a/port/linuxs39064/omrrttime.s
+++ b/port/linuxs39064/omrrttime.s
@@ -31,7 +31,7 @@ maxprec:
     
     stckf 64(%r15)
     lg  %r1,64(%r15)
-    sg %r1,.LC0-.LT0(%r3)
+    slg %r1,.LC0-.LT0(%r3)
     srlg %r1,%r1,1
     lgr %r2,%r1
     br  %r14
--- a/port/zos390/omrrttime.s
+++ b/port/zos390/omrrttime.s
@@ -49,7 +49,7 @@ LCarry0  sl   r2,LC0
          sl  r3,CVTLSO+4
          brc NOBORROW,LCarry1
          ahi r2,-1
-LCarry1  s   r2,CVTLSO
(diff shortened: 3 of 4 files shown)


charliegracie:
genie-omr build zlinux

deesebas:
@joransiu , @DanHeidinga, please review. Thanks.

DanHeidinga:
I'll defer to @joransiu or someone with more z-assembly knowledge (Fillip?) to review this

DanHeidinga:
@fjeremic Is there a different instruction for 64bit?  i.e.:  `SLGB` of `SLBG`?

deesebas:
Thanks for the suggestion @fjeremic

@DanHeidinga, there is an SLBG instruction, but in the 64-bit case, SLG will be sufficient.

In the 32-bit case, we need to split the obtained 64-bit STCK value into two halves to perform the subtraction. Currently for example, for 31-bit  z/os:

maxprec  EDCXPRLG BASEREG=8
         stckf CLOCK31(r4)
         lm  r2,r3,CLOCK31(r4)
         sl  r3,LC0+4                       Lower half of subtraction
         brc NOBORROW,LCarry0    Check for borrow 
         ahi r2,-1                            If borrow=yes, subtract 1 from higher half of the STCK value
LCarry0  sl   r2,LC0                      Higher half of subtraction

The improvement Fillip suggests (and this is not related to the current issue) is that the two instructions (brc, ahi) can be avoided by using a SLB instruction for the higher half of the subtraction. First perform SL on the lower half, which will set the 'Borrow' bit in the PSW to 0, if the SL resulted in a borrow. Then perform SLB on the higher half, which will automatically take care of the borrow from the previous SL instruction.

deesebas:
There are a few different places where the improvement suggested by @fjeremic needs to be done. For now, I'd like to go ahead with the current changes. Created #1725 for the improvements.

deesebas:
@fjeremic in my second commit I've made an additional change in the z/os codepath.

DanHeidinga:
@deesebas If this is ready to commit, please remove the 'WIP' from the title

DanHeidinga:
@0xdaryl @charliegracie Can one of you merge this?

charliegracie:
genie-omr build zlinux

PR #1723, opened 2017-09-25
Add support for nopable breakpoint guard

cathyzhyi:
To lower the overhead for the breakpoint guard instruction sequence,
the guard is nopped when compiling a method and patched
only when a breakpoint is set for that method. This changeset cleans
up related queries and adds the option to disable nopping breakpoint
guard

### DIFF
--- a/compiler/compile/VirtualGuard.cpp
+++ b/compiler/compile/VirtualGuard.cpp
@@ -227,38 +227,46 @@ TR_VirtualGuard::createBreakpointGuardNode
 #ifdef J9_PROJECT_SPECIFIC
 /*
  * Shape of a breakpoint guard:
- * iflcmpeq/ificmpeq
+ * iflcmpne/ificmpne
  *    land/iand
  *       lloadi/iloadi <ConstantPool shadow>
  *          aconst J9Method
  *       isBreakpointedBit 
- *    isBreakpointedBit 
+ *    iconst 0 
  */
-   bool is64Bit = TR::Compiler->target.is64Bit();
    TR::SymbolReferenceTable * symRefTab = comp->getSymRefTab();
    TR::SymbolReference * fieldSymRef = symRefTab->findOrCreateJ9MethodConstantPoolFieldSymbolRef(offsetof(struct J9Method, constantPool));
    TR::Node * aconstNode = TR::Node::aconst(callNode, (uintptrj_t)calleeSymbol->getResolvedMethod()->getPersistentIdentifier());
-   TR::Node * constantPool = TR::Node::createWithSymRef(is64Bit? TR::lloadi: TR::iloadi, 1, 1, aconstNode, fieldSymRef);
+   TR::Node * constantPool = NULL;
    aconstNode->setIsMethodPointerConstant(true);
    aconstNode->setInlinedSiteIndex(calleeIndex);
    aconstNode->setByteCodeIndex(0);
    TR::Node * flagBit = NULL;
    TR::Node *guard = NULL;
+   TR::Node * zero = NULL;
    if (TR::Compiler->target.is64Bit())
       {
       flagBit = TR::Node::create(callNode, TR::lconst, 0, 0);
       flagBit->setLongInt(comp->fej9()->offsetOfMethodIsBreakpointedBit());
-
+      zero = TR::Node::create(callNode, TR::lconst, 0, 0);
+      constantPool = TR::Node::createWithSymRef(TR::lloadi, 1, 1, aconstNode, fieldSymRef);
+      guard =  TR::Node::createif(TR::iflcmpne,
+                  TR::Node::create(TR::land, 2, constantPool, flagBit),
+                  zero,
+                  destination);
       }
    else
       {
       flagBit = TR::Node::create(callNode, TR::iconst, 0, comp->fej9()->offsetOfMethodIsBreakpointedBit());
+      zero = TR::Node::create(callNode, TR::iconst, 0, 0);
+      constantPool = TR::Node::createWithSymRef(TR::iloadi, 1, 1, aconstNode, fieldSymRef);
+      guard =  TR::Node::createif(TR::ificmpne,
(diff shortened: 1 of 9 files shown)


vijaysun-omr:
@andrewcraik please review this as well.

andrewcraik:
@cathyzhyi can you please explain why the 3 static helper functions have now been exposed as public APIs? Are they exposed because they need to be extended/replaced or because a new part of the system needs to call them? I had a look at this change and I don't quite see why they are being exposed.

compiler/compile/VirtualGuard.cpp
```diff
@@ -244,20 +244,22 @@ TR_VirtualGuard::createBreakpointGuardNode
...
+      zero = TR::Node::create(callNode, TR::lconst, 0, 0);
       }
    else
       {
       flagBit = TR::Node::create(callNode, TR::iconst, 0, comp->fej9()->offsetOfMethodIsBreakpointedBit());
+      zero = TR::Node::create(callNode, TR::iconst, 0, 0);
       }
 
-   guard =  TR::Node::createif(is64Bit? TR::iflcmpeq: TR::ificmpeq,
+   guard =  TR::Node::createif(is64Bit? TR::iflcmpne: TR::ificmpne,
```
andrewcraik: Can't both of these opcodes be assigned in the if else above using a local to remove the ternary operator?

compiler/compile/VirtualGuard.cpp
```diff
@@ -227,12 +227,12 @@ TR_VirtualGuard::createBreakpointGuardNode
...
- * iflcmpeq/ificmpeq
+ * iflcmpne/ificmpne
  *    land/iand
  *       lloadi/iloadi <ConstantPool shadow>
  *          aconst J9Method
  *       isBreakpointedBit 
- *    isBreakpointedBit 
+ *    iconst 0 
  */
    bool is64Bit = TR::Compiler->target.is64Bit();
```
andrewcraik: Not a fan of having both this and the if..else.. now that the logic is getting blended - would be nice to simplify this down a bit.

cathyzhyi: Thanks a lot for the suggestion. Modified the code to get rid of the ternary sign.

cathyzhyi:
The isStopTheWorldGuard is called from  `static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)` so that the condition check a meaningful query instead of checking HCRGuard, OSRGuard and BreakpointGuard individually.

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -2531,12 +2531,13 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
...
+
    TR::Instruction *vgnopInstr = generateVirtualGuardNOPInstruction(node, site, deps, label, cg);
    TR::Instruction *patchPoint = cg->getVirtualGuardForPatching(vgnopInstr);
 
-   // HCR guards are patched when the threads are stopped.
-   // No issues with multithreaded patching, therefore alignment is not required
-   if (TR::Compiler->target.isSMP() && !node->isHCRGuard() && !node->isOSRGuard())
+   // Guards patched when the threads are stopped have no issues with multithreaded patching. 
+   // therefore alignment is not required
+   if (TR::Compiler->target.isSMP() && !cg->isStopTheWorldGuard(node))
```
andrewcraik: Is a similar change needed in any other codegens?

andrewcraik: Thanks @fjeremic can you review the z change?

cathyzhyi: The other codegen doesn't need patching alignment but after discussion with @fjeremic there is a binary encoding opt that could be applied to breakpoint guard as it is a stop the world guard. I am changing that code to use this query.

compiler/z/codegen/S390Instruction.cpp
```diff
@@ -5804,17 +5804,17 @@ TR::S390VirtualGuardNOPInstruction::generateBinaryEncoding()
    // in) if the patching occurs during GC pause times.  The patching of up to 6-bytes is potentially
    // not atomic.
 
-   bool performEmptyPatch = getNode()->isHCRGuard() || getNode()->isProfiledGuard() || getNode()->isOSRGuard();
+   bool performEmptyPatch = cg()->isStopTheWorldGuard(getNode()) || getNode()->isProfiledGuard();
```
fjeremic: Outside the scope of this item but just confirming with the experts here that profiled guards are *not* stop the world guards? In which case I'm not sure this code is safe because we would not emit any instructions for profiled guards.

cathyzhyi: The code is safe but needs cleanup. The concept of stop the world guard is only relevant for nopable guards which are nops got patched at runtime. Profiled guards are not nopable guards and they are explicit compare and jump instructions. Therefore, the getNode()->isProfiledGuard() should never return true when calling from S390VirtualGuardNOPInstruction.

cathyzhyi: Sry, what I said about isProfiledGuard() can never be true is wrong. When a profiled guard is merged with HCRGuard both explicit compare instructions and the S390VirtualGuardNOPInstruction are generated for that profiled guard. The gist for this code is that when this happens the explicit compare instruction can be a patched point and nothing needs to be generated for S390VirtualGuardNOPInstruction. See the comments here: https://github.com/eclipse/omr/blob/94eaf838db94badd7a7c57bebc621994a471cf11/compiler/z/codegen/S390Instruction.cpp#L5809-L5811

vijaysun-omr:
@andrewcraik and @fjeremic are you fine with these changes ?

andrewcraik:
@vijaysun-omr I'm fine with these

PR #1727, opened 2017-09-26
Enable AVX support on X86

0dvictor:
X86 CodeGen now generates AVX instructions unless the machine does
not support AVX or environment variable TR_DisableAVX is set.

### DIFF
--- a/compiler/x/codegen/X86BinaryEncoding.cpp
+++ b/compiler/x/codegen/X86BinaryEncoding.cpp
@@ -1432,6 +1432,9 @@ TR::X86RegInstruction::enlarge(int32_t requestedEnlargementSize, int32_t maxEnla
    if (disableRexExpansion || TR::comp()->getOption(TR_DisableZealousCodegenOpts))
       return OMR::X86::EnlargementResult(0, 0);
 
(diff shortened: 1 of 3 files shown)


andrewcraik:
@0dvictor could you clarify if by enable AVX support you mean we will generate the AVX encodings by default or if you are just allowing AVX encodings to be added to the code generator by subsequent changes.

0dvictor:
Good question. This PR will make X86 CG always use AVX encodings unless it is not supported or explicitly disabled.

lmaisons:
Is it worth updating the commit message to explicitly indicate that this is enabled by default?

andrewcraik:
This sounds good to me though there is a caveat emptor to consumers of OMR that they can see the new encodings now if they run on new enough hardware. That isn't a bad thing, it is something to be aware of.

0dvictor:
True, and I believe significant portion of consumers will see the new encodings, as AVX was introduced with Sandy Bridge around 2011.

0dvictor:
@lmaisons Thank you for pointing it out. The commit message has actually already been updated.

PR #1729, opened 2017-09-26
Change TOC size from 256KB to 64KB in -Xquickstart mode

harryyu1994:
We are changing the TOC size from 256KB back to the original 64KB in -Xquickstart mode in an effort to decrease footprint.

The default TOC size was changed from 64KB to 256KB in commit bdc1efa to help with performance in Liberty/Hadoop/Spark. However, since throughput is not very important in -Xquickstart mode, we can afford to revert back to 64KB in this mode. 

This change will decrease footprint.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -5025,6 +5025,7 @@ void OMR::Options::setQuickStart()
    {
    _quickstartDetected = true;
    _initialOptLevel = cold;
+   _tocSizeInKB = 64;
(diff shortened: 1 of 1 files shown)


harryyu1994:
@mpirvu Could you please review this change, thanks.

mpirvu:
Code looks good to me.

mpirvu:
@harryyu1994  You can remove the WIP header now.

vijaysun-omr:
Simple change under -Xquickstart. Merging.

PR #1734, opened 2017-09-27
isArray test on a class flag from RAM class

liqunl:
isArray test should be testing a flag from RAM class instead of ROM
class to save us one indirect load. The folding optimization needs
modification as well to detect the right tree. This change also remove
current limitation where folding only happens to isArray test generated
by LoopVersioner, it will work for isArray test generated by ilgen and
any optimization.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -7575,10 +7575,18 @@ void TR_LoopVersioner::collectAllExpressionsToBeChecked(List<TR::TreeTop> *nullC
                TR::Node *duplicateCheckedValue = node->getFirstChild()->duplicateTreeForCodeMotion();
                TR::Node *vftLoad = TR::Node::createWithSymRef(TR::aloadi, 1, 1, duplicateCheckedValue, comp()->getSymRefTab()->findOrCreateVftSymbolRef());
                //TR::Node *componentTypeLoad = TR::Node::create(TR::aloadi, 1, vftLoad, comp()->getSymRefTab()->findOrCreateArrayComponentTypeSymbolRef());
-               TR::Node *romClassLoad = TR::Node::createWithSymRef(TR::aloadi, 1, 1, vftLoad, comp()->getSymRefTab()->findOrCreateClassRomPtrSymbolRef());
-               TR::Node *isArrayField = TR::Node::createWithSymRef(TR::iloadi, 1, 1, romClassLoad, comp()->getSymRefTab()->findOrCreateClassIsArraySymbolRef());
-               TR::Node *andConstNode = TR::Node::create(isArrayField, TR::iconst, 0, TR::Compiler->cls.flagValueForArrayCheck(comp()));
-               TR::Node * andNode   = TR::Node::create(TR::iand, 2, isArrayField, andConstNode);
+               TR::Node *classFlag = NULL;
+               if (TR::Compiler->target.is32Bit())
+                  {
+                  classFlag = TR::Node::createWithSymRef(TR::iloadi, 1, 1, vftLoad, comp()->getSymRefTab()->findOrCreateClassAndDepthFlagsSymbolRef());
+                  }
+               else
+                  {
+                  classFlag = TR::Node::createWithSymRef(TR::lloadi, 1, 1, vftLoad, comp()->getSymRefTab()->findOrCreateClassAndDepthFlagsSymbolRef());
+                  classFlag = TR::Node::create(TR::l2i, 1, classFlag);
+                  }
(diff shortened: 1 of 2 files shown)


liqunl:
The change has been tested at acceptance level,. @andrewcraik Could you review my change?

andrewcraik:
Can the old set of trees that VP used to match still be generated anywhere in the compiler? Before we remove the support we need to be sure of this fact. The new implementation makes sense given the test of the flag can be done that way.

liqunl:
Places where the old set of trees used to be generated are modified to generate the new set of trees.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -7782,42 +7782,54 @@ TR::Node *constrainIand(OMR::ValuePropagation *vp, TR::Node *node)
...
-                        if (objConstraint->getClassType()->isArray() == TR_yes)
-                           constraint = TR::VPIntConst::create(vp, rhs->asIntConst()->getLowInt()/*, isUnsigned*/);
-                        else
-                           constraint = TR::VPIntConst::create(vp, 0/*, isUnsigned*/);
-                        }
+                     if (vp->trace())
+                        traceMsg(vp->comp(), "Fold isArray test in %p\n", baseExpression);
+
+                     if (baseExpressionConstraint->getClassType()->isArray() == TR_yes)
+                        constraint = TR::VPIntConst::create(vp, rhs->asIntConst()->getLowInt()/*, isUnsigned*/);
```
andrewcraik: why is isUnsigned commented out?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -7782,42 +7782,54 @@ TR::Node *constrainIand(OMR::ValuePropagation *vp, TR::Node *node)
...
                      else
-                        {
-                        //if (vp->_isGlobalPropagation)
-                        //   {
-                        //   dumpOptDetails(vp->comp(), "NOT Folding isArray test in %p\n", objectLoad);
-                        //   printf("NOT Folding isArray test in %s\n", vp->comp()->signature()); fflush(stdout);
-                        //   }
-                        }
+                        constraint = TR::VPIntConst::create(vp, 0/*, isUnsigned*/);
```
andrewcraik: why is isUnsigned commented out?

compiler/optimizer/VPHandlers.cpp
liqunl: That is legacy code from the original implementation. TR::VPIntConst::create takes only 2 arguments now. Commit is modified to remote isUnsigned.

compiler/optimizer/VPHandlers.cpp
liqunl: Same above.

andrewcraik:
super thank LGTM @vijaysun-omr over to you for a second review to make sure I didn't miss anything

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -7575,10 +7575,18 @@ void TR_LoopVersioner::collectAllExpressionsToBeChecked(List<TR::TreeTop> *nullC
                TR::Node *duplicateCheckedValue = node->getFirstChild()->duplicateTreeForCodeMotion();
                TR::Node *vftLoad = TR::Node::createWithSymRef(TR::aloadi, 1, 1, duplicateCheckedValue, comp()->getSymRefTab()->findOrCreateVftSymbolRef());
                //TR::Node *componentTypeLoad = TR::Node::create(TR::aloadi, 1, vftLoad, comp()->getSymRefTab()->findOrCreateArrayComponentTypeSymbolRef());
-               TR::Node *romClassLoad = TR::Node::createWithSymRef(TR::aloadi, 1, 1, vftLoad, comp()->getSymRefTab()->findOrCreateClassRomPtrSymbolRef());
-               TR::Node *isArrayField = TR::Node::createWithSymRef(TR::iloadi, 1, 1, romClassLoad, comp()->getSymRefTab()->findOrCreateClassIsArraySymbolRef());
-               TR::Node *andConstNode = TR::Node::create(isArrayField, TR::iconst, 0, TR::Compiler->cls.flagValueForArrayCheck(comp()));
-               TR::Node * andNode   = TR::Node::create(TR::iand, 2, isArrayField, andConstNode);
+               TR::Node *classFlag = NULL;
+               if (TR::Compiler->target.is32Bit())
```
vijaysun-omr: Should this take into account the fact that class pointers are compressed to be 32-bits wide in compressed references mode ?

vijaysun-omr: Never mind, it's not the vft load that is being loaded via an int...it's the class depth and flags and that should be okay to load using the below trees as shown on all platforms.

vijaysun-omr:
Looks okay to me. Merging.

PR #1735, opened 2017-09-27
Fix BNDCHK versioning tests under substitution

jdmpapin:
In order to identify appropriate expressions of induction variables (IVs) under `BNDCHK` nodes, loop versioner looks through loads of autos to their definitions (when there are unique definitions).  In this way a `BNDCHK` may be versioned even if no load of its IV feeds directly into the index.  For instance, the following `BNDCHK` can be versioned as long as `n` and `array` are invariant.

    BBStart <loop0>
    istore derived
      isub
        iload n
        iload i
    if... --> exit
    BBEnd </loop0>

    BBStart <loop1>
    BNDCHK
      arraylength
        aload array
      iload derived
    ...
    istore i
      isub
        iload i
        iconst -1
    ificmple --> loop0
      ==>isub
      iload n
    BBEnd </loop1>

This commit fixes some problems with the versioning tests generated in cases like this one:

1. Substitution into the third test sometimes used the IV instead of the substituted variable's definition.  In the above example, derived would be replaced by `i`, not `n - i`.  This only happened in cases where the entire index expression is a load.  If the index were instead `derived + 1`, then it would be correctly rewritten to `n - i + 1`.

2. All of the `BNDCHK` versioning tests are sensitive to whether the IV occurs in a negative position, but versioner did not account for subtractions within the definition of a substituted variable.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -2102,7 +2102,13 @@ TR::Node *TR_LoopVersioner::isDependentOnInvariant(TR::Node *useNode)
    }
 
 
-TR::Node *TR_LoopVersioner::isDependentOnInductionVariable(TR::Node *useNode, bool noArithmeticAllowed, bool &isIndexChildMultiplied, TR::Node * &mulNode, TR::Node * &strideNode)
+TR::Node *TR_LoopVersioner::isDependentOnInductionVariable(
+   TR::Node *useNode,
+   bool noArithmeticAllowed,
+   bool &isIndexChildMultiplied,
+   TR::Node * &mulNode,
+   TR::Node * &strideNode,
+   bool &indVarOccursAsSecondChildOfSub)
    {
    TR_UseDefInfo *useDefInfo = optimizer()->getUseDefInfo();
    if (!useDefInfo)
@@ -2155,6 +2161,9 @@ TR::Node *TR_LoopVersioner::isDependentOnInductionVariable(TR::Node *useNode, bo
                   bool isFirstChildInvariant = isExprInvariant(child->getFirstChild());
                   if (isFirstChildInvariant)
                      {
+                     if (child->getOpCode().isSub())
+                        indVarOccursAsSecondChildOfSub = !indVarOccursAsSecondChildOfSub;
+
                      if(isIndexChildMultiplied)
                          strideNode=child->getFirstChild();
                      child = child->getSecondChild();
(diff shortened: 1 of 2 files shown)


jdmpapin:
This has been tested using J9 based on the idea that, when we version a `BNDCHK` that uses a temporary in this way instead of directly using the IV, the generated versioning tests should be the same as those we would get if the definition of the temporary were pre-substituted into the index. I generated 128 pairs of loops that should correspond in this way, e.g.

    public static void go00111110(int[] a, int n, int m) {
        // Params(substitute=False, count_backward=True, offset_count=True, inclusive_bound=True, flip_tmp=True, offset_tmp=True, flip_index=False, bump_early=False)

        int bound = 1;
        int i = n;
        while (i >= bound) {
            int tmp = m + 1 - i;    // simplified from: m - 1 - (i - 1) + 1
            if (bail) throw new Use(tmp);

            a[tmp - 1] = 42;
            i--;
        }
    }

    public static void go00111111(int[] a, int n, int m) {
        // Params(substitute=True, count_backward=True, offset_count=True, inclusive_bound=True, flip_tmp=True, offset_tmp=True, flip_index=False, bump_early=False)

        int bound = 1;
        int i = n;
        while (i >= bound) {
            int tmp = m + 1 - i;    // simplified from: m - 1 - (i - 1) + 1
            if (bail) throw new Use(tmp);

            a[m - i] = 42;  // simplified from: a[m - 1 - (i - 1) + 1 - 1]
            i--;
        }
    }

Passing the same value for `n` and `m`, these loops will all set `a[k] = 42` for each `k` in `[0, n)`, but in various different ways described by the `Params`:

- `count_backward`: loop over decreasing `i`, rather than increasing;
- `offset_count`: include an extra `+1` in `i`, compensating in `tmp`;
- `inclusive_bound`: use an inclusive rather than exclusive loop bound;
- `flip_tmp`: reverse the order of `tmp`, putting `i` in negative position;
- `offset_tmp`: include an extra `+1` in `tmp`, compensating in the index;
- `flip_index`: reverse the order of indexing, putting `tmp` in negative position;
- `bump_early`: adjust `i` before indexing rather than afterward.

I compiled each method with tracing, and parsed the versioning tests from the log. While the tests generated for each pair can't be expected to be identical, a few simplifications should reveal that they are equivalent. I applied simplifications similar to those mentioned in the above generated Java source.

With this fix all pairs of loops generate equivalent tests, where previously 72 of the 128 pairs had differences.

vijaysun-omr:
I have to commend the dedication to testing this as thoroughly as you have done with the variations of the loops. I looked at the versioner code now and it looks correct to me. So I am merging.

PR #1736, opened 2017-09-27
CMake: Add basic support for installing targets

dnakamura:
At the moment only installs tooling and omrsig (only the minimum required
for building openJ9).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,23 @@ endif()
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 ###
+### Set up install path stuff
+###
+# Note the paths are a bit odd compared to standard gnu install paths
+# This is because we arent expecting to actually be installing globally
+# since we don't have a good story for installing OMR system-wide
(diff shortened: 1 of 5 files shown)


charliegracie:
@rwy0717 and @mgaudet any comments?

charliegracie:
I am accepting this as-is so further progress can be made.  If @mgaudet or @rwy0717 have any comments we can address them in a future commit.

rwy0717:
I don't see any problems, but could you just describe how and why this is being used?

I guess J9 will build and install OMR, as part of it's build process? In J9, Is OMR built using `add_subdirectory`, or are we doing something different?

dnakamura:
basically the vm is just using it so it puts the important outputs of the build in one place (ie compose a folder with all the shared libs it needs). Also sets up the RPATH.
yes J9 is just  calling `add_subdirectory` on omr, which means that the global install target just inherits all the install stuff from omr.

rwy0717:
OK cool, thanks for explaining @dnakamura :)

PR #1737, opened 2017-09-27
Remove unused CS2 files

lmaisons:
Given that we want to actively avoid introducing new usages of CS2 into
the compiler component, remove the files that are not / no longer in
use.

Furthermore, remove the CS2 test directory as it has never been used and
does not fit with the rest of our test framework.


andrewcraik:
Sounds good to LGTM - I will add that usage of the CS2 bitvectors is in decline so these data structures are slowly leaving the codebase. Nice to see the dead ones being cleaned up.

0xdaryl:
genie-omr build

0xdaryl:
genie-omr build all

Issue #1726, opened 2017-09-26
Document compilation exception semantics


0xdaryl:
The compiler technology uses C++ exceptions to indicate abnormal circumstances while a method is being compiled.  Many of them are fatal for the method being compiled, but some are not.

While many appear to be self-explanatory by their name alone, each exception should be documented and describe to runtime implementors the circumstances in which they should be thrown and the implications of handling them.

Documentation on how a runtime implementor can add their own compilation exception and how a downstream project can add one is helpful too.

From `runtime/CodeCacheExceptions.hpp`:
* CodeCacheError
* RecoverableCodeCacheError

From `compile/CompilationException.hpp`:
* CompilationException
* ILGenFailure
* RecoverableILGenException
* ExcessiveComplexity
* MaxCallerIndexExceeded
* CompilationInterrupted
* UnimplementedOpCode
* InsufficientlyAggressiveCompilation
* GCRPatchFailure

dsouzai:
Since I added these exception types, I will document them. Should it be a comment in the *.hpp files or should it be in a .md file in the doc directory?

0xdaryl:
Thanks @dsouzai !  I would prefer comments in the code to avoid staleness.

dsouzai:
I think this issue can be closed.

PR #1741, opened 2017-09-28
Document Exception Semantics

dsouzai:
Addresses https://github.com/eclipse/omr/issues/1726. This PR also includes some minor cleanup, namely using the `TR::Compilation::failCompilation` method instead of explicitly throwing an exception.


andrewcraik:
Documentation itself looks good to me - covers the exceptions and their use cases nicely.

0xdaryl:
genie-omr build all

PR #1742, opened 2017-09-28
z/TPF Fixes during initial sniff testing.

jjohnst-us:
z/TPF fixes during initial sniff testing.  Fixes unresolved functions.
Fixes High Resolution Timer defines.  Modified setstacksize to match 
z/TPF architecture.

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -1780,12 +1780,18 @@ omrsysinfo_get_limit(struct OMRPortLibrary *portLibrary, uint32_t resourceID, ui
 #else /* !defined(OMRZTPF) */
 		/* z/TPF does not have virtual address support */
 		*limit = OMRPORT_LIMIT_UNKNOWN_VALUE;
-		rc = OMRPORT_LIMIT_UNKNOWN;
 		Trc_PRT_sysinfo_get_limit_Exit(rc);
 		return rc;
 #endif /* !defined(OMRZTPF) */
 	} else if (OMRPORT_RESOURCE_CORE_FILE == resourceRequested) {
+#if !defined(OMRZTPF)
 		resource = RLIMIT_CORE;
+#else /* !defined(OMRZTPF) */
+		/* z/TPF does not have virtual address support */
+		*limit = OMRPORT_LIMIT_UNKNOWN_VALUE;
+		Trc_PRT_sysinfo_get_limit_Exit(rc);
+		return rc;
+#endif /* !defined(OMRZTPF) */
 	}
 
 	switch (resourceRequested) {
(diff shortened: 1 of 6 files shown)


charliegracie:
genie-omr build all

port/unix/omrsysinfo.c
```diff
@@ -2201,6 +2220,11 @@ omrsysinfo_limit_iterator_next(struct OMRPortLibrary *portLibrary, J9SysinfoLimi
 
 		rc = 0;
 	}
+#else /* !defined(OMRZTPF) */
+		rc = OMRPORT_ERROR_SYSINFO_OPFAILED;
```
charliegracie: It looks like these 3 lines are tabbed in 1 too many times

port/ztpf/omrsignal_context.c
```diff
@@ -216,6 +216,7 @@ allocateDurableStorage( void )
 	rtnv->argv.sii = &(rtnv->siginfo);
 	rtnv->argv.uct = &(rtnv->ucontext);
 	rtnv->argv.sct = &(rtnv->sigcontext);
+	rtnv->sigcontext.sregs = &(rtnv->_sregs);   //point to actual _sigregs struct
```
charliegracie: You should not be using `//` comments in C code

port/ztpf/safe_storage.h
```diff
@@ -46,6 +46,7 @@ typedef struct _durable_storage {
 	siginfo_t	siginfo;
 	ucontext_t ucontext;
 	struct sigcontext sigcontext;
+	_sigregs _sregs; //sigcontext struct contains pointer to _sigregs
```
charliegracie: You should not be using `//` comments in C code

vijaysun-omr:
@dsouzai Can I please get a review done on the first file changed in this commit ?

dsouzai:
The changes to DebugSegmentProvider.cpp look fine to me. That said, the changes make it so that the DebugSegmentProvider uses the `TR::RawAllocator` to allocate/free memory. Because `TR::RawAllocator` uses `malloc` to allocate memory, it means that the DebugSegmentProvider loses some of its ability to provide debugging for memory corruption related bugs under OMRZTPF - with `mmap`, the memory pages are protected which will cause a crash immediately when that memory is accessed after it was freed, whereas with `TR::RawAllocator` the memory is painted with `0xEF` which may cause a crash, but not necessarily when the memory was accessed. As such, at some point in the future, there should be something similar to `mmap` on OMRZTPF for RAS purposes.

jjohnst-us:
@charliegracie 
All your requested comments were applied.  Let me know if you need anything else.

charliegracie:
genie-omr build zlinux

jjohnst-us:
genie-omr build zlinux

PR #1744, opened 2017-09-28
Fix compilation issue with older MSVC compilers

andrewcraik:
The use of inttypes.h in StaticRelocation.hpp is not compatible with
many older versions of the Microsoft Visual C/C++ compiler. This commit
changes this to stdint.h which should be more portable without affecting
the implementation.


0xdaryl:
genie-omr build all

PR #1749, opened 2017-09-29
Move ELF test make dependencies

### LINKED ISSUE #1748: JitBuilder PPC LE CI Failure: multiple definition of `unknown helper (trampoline)'
Introduced by #1612 

Building static ELF tests requires running dynamic call test.  Call doesn't work on platforms without trampoline support.

lmaisons:
Move the make dependencies for the ELF tests from ALL_TESTS to explicit
dependencies of experimental_goal.

This is required because entries in ALL_TESTS get built on all
platforms, even if they can't run on all platforms. However, the
distinction between 'building' and 'running' the ELF tests if fuzzy
enough that if a platform doesn't support running the test, it doesn't
support building it either.

Fixes #1748

### DIFF
--- a/jitbuilder/release/Makefile
+++ b/jitbuilder/release/Makefile
@@ -56,8 +56,6 @@ ALL_TESTS = \
             transactionaloperations \
             union \
             worklist \
-            useIncrement \
-            useCall \
 
 
 # Compile all the tests by default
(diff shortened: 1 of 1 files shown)


charliegracie:
genie-omr build all

PR #1751, opened 2017-09-29
Flush GC caches in Concurrent Scavenger

amicic:
Push any open GC copy caches to scan queue, and abandon unused parts:

1) for mutator threads (that created copy caches while copying in read
barrier), when requested heap walk in a middle of a CS cycle. 

2) at the end of first STW phase of CS
This will help with scan work created by GC threads that were involved
in this phase, but not involved in following concurrent phase (which
generally uses less GC threads). That work will be then completed during
concurrent phase, instead of waiting for those GC threads to be
re-activated at the final STW phase. 
This will also help with heap walkability (in a middle of a GC cycle).


charliegracie:
genie-omr build all

PR #1753, opened 2017-09-29
Ignore OSR shared slot info under voluntary OSR

ncough:
In voluntary OSR, OSR shared slot information will
be generated for all OSR points initially, but only
a limited set of these eventually have to be
commited to the OSR instruction table. This change
adds a condition to eliminate a majority of these,
cutting footprint and compile time overhead.


vijaysun-omr:
Looks like a safe change to me since we can only do OSR at the calls that induce OSR in voluntary mode. Merging.

PR #1754, opened 2017-09-29
Implement OSR slot bookkeeping as linked list

ncough:
It is cheaper to implement the OSR slot sharing
infrastructure as a linked list, due to the
removals during its compression.


vijaysun-omr:
Looks okay to me. I would like a review from @andrewcraik

andrewcraik:
LGTM

PR #1755, opened 2017-09-29
Merge HCR guards in all HCR modes

ncough:
HCR guards were not merged with other guards
in the OSR implementation of HCR to simplify the
analysis. This change enables the merging, to reduce
compile time overhead, as the analysis should now
support it.


vijaysun-omr:
Just so it is documented @ncough could you please elaborate on what was the extra support needed to handle the merging of HCR guards under OSR ? i.e. why it could not be done before and what was changed to allow it now ?

ncough:
In the past, it was not possible to remove merged HCR guards, however, this functionality was recently added. An analysis in a downstream project, OpenJ9, which removed HCR guards has been extended to make use of this functionality and treat merged guards appropriately. Therefore, the inliner can keep the merged guards, reducing overhead.

PR #1756, opened 2017-09-29
Use direct TR_BitVector iterator

ncough:
Structure analysis was recently changed to use
TR_BitVectors rather than their CS2 equivalent.
However, the iterators were not updated. Instead,
TR_BitVector relied on a wrapper to match the
CS2 BitVector implementation. This changes
structural analysis to use the iterator directly.


andrewcraik:
LGTM

vijaysun-omr:
Looks like a straightforward change to the iterator used. Merging.

PR #1761, opened 2017-10-02
Concurrent Scavenger Nursery resizing

amicic:
Increase Nursery alignment in Subspace logic to 2 regions (from 1
region). Later in PhysicalSubArena code, the 2 region alignment is enforced
anyway. Hence attempts to resize with less than two regions would be
ignored. Since region is rather large in CS, most of attempts,
especially early ones, would be ignored.

Also, adding very rudimental adjustment to sizing heuristics, to
accommodate for the low number of background GC threads used.


amicic:
@dmitripivkine Please, have a look.

dmitripivkine:
Looks good for me

charliegracie:
genie-omr build all

PR #1762, opened 2017-10-02
Zero OSR loads that are dead on all paths

ncough:
Once OSR infrastructure has been removed, OSR
transitions can no longer be added. Therefore, it
is simple to identify stack slots that are dead
at all transition points, reducing the number of
loads and dead stores, cutting compile time.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -1300,6 +1300,62 @@ TR_OSRLiveRangeAnalysis::optDetailString() const throw()
    return "O^O OSR LIVE RANGE ANALYSIS: ";
    }
 
+/**
+ * Collect the dead stores in the provided block. Either initialize the provided bitvector with
+ * the dead stores symrefs that were found or take the intersection of the two, filtering down to
+ * symrefs that are always dead.
+ */
+bool TR_OSRExceptionEdgeRemoval::addDeadStores(TR::Block* osrBlock, TR_BitVector& dead, bool inited)
+   {
+   TR_ASSERT(osrBlock->isOSRCodeBlock() || osrBlock->isOSRInduceBlock(), "It is only safe to identify dead stores in OSRCodeBlocks and OSRInduceBlocks");
+
+   _seenDeadStores->empty();
+   for (TR::TreeTop *tt = osrBlock->getFirstRealTreeTop(); tt != osrBlock->getExit(); tt = tt->getNextTreeTop())
+      {
+      TR::Node *node = tt->getNode();
+      if (node->getOpCode().isStoreDirect() && node->getSymbol()->isAutoOrParm() && node->storedValueIsIrrelevant())
+         _seenDeadStores->set(node->getSymbolReference()->getReferenceNumber());
+      }
+
+   if (inited)
+      dead &= *_seenDeadStores;
+   else
+      dead |= *_seenDeadStores;
+
+   if (comp()->getOption(TR_TraceOSR))
+      {
+      traceMsg(comp(), "Identified dead stores for block_%d:\n", osrBlock->getNumber());
+      _seenDeadStores->print(comp());
+      traceMsg(comp(), "\nRemaining dead stores:\n");
+      dead.print(comp());
+      traceMsg(comp(), "\n");
+      }
+
+   return !_seenDeadStores->isEmpty();
+   }
+
+/**
+ * If loads of a symbol along a OSR paths are zeroed, the dead stores can be removed
+ * as well. Given an OSRCodeBlock or an OSRInduceBlock and a bitvector of dead symbol
+ * references, this function will remove them.
+ */
+void TR_OSRExceptionEdgeRemoval::removeDeadStores(TR::Block* osrBlock, TR_BitVector& dead)
+   {
+   TR_ASSERT(osrBlock->isOSRCodeBlock() || osrBlock->isOSRInduceBlock(), "It is only safe to remove dead stores from OSRCodeBlocks and OSRInduceBlocks");
+
+   for (TR::TreeTop *tt = osrBlock->getFirstRealTreeTop(); tt != osrBlock->getExit(); tt = tt->getNextTreeTop())
+      {
+      TR::Node *node = tt->getNode();
+      if (node->getOpCode().isStoreDirect() && node->getSymbol()->isAutoOrParm() && node->storedValueIsIrrelevant() &&
+          dead.get(node->getSymbolReference()->getReferenceNumber()))
+         {
+         if (comp()->getOption(TR_TraceOSR))
+            traceMsg(comp(), "Removing dead store n%dn of symref #%d\n", node->getGlobalIndex(), node->getSymbolReference()->getReferenceNumber());
+         TR::TransformUtil::removeTree(comp(), tt);
+         } 
+      }
+   }
+
 int32_t TR_OSRExceptionEdgeRemoval::perform()
    {
    if (comp()->getOption(TR_EnableOSR))
@@ -1347,6 +1403,7 @@ int32_t TR_OSRExceptionEdgeRemoval::perform()
 
    TR::CFG *cfg = comp()->getFlowGraph();
 
+   // Remove edges to OSR catch blocks that are no longer necessary
    TR::CFGNode *cfgNode;
    for (cfgNode = cfg->getFirstNode(); cfgNode; cfgNode = cfgNode->getNext())
       {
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
I would like a review from @andrewcraik

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1314,6 +1314,35 @@ TR_OSRLiveRangeAnalysis::optDetailString() const throw()
...
+   if (inited)
+      dead &= *_seenDeadStores;
+   else
+      dead |= *_seenDeadStores;
+
+   return !_seenDeadStores->isEmpty();
+   }
+
+void TR_OSRExceptionEdgeRemoval::removeDeadStores(TR::Block* osrBlock, TR_BitVector& dead)
+   {
```
andrewcraik: An assert to enforce the correct kind of bloc is given would be a good idea here...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1378,6 +1408,82 @@ int32_t TR_OSRExceptionEdgeRemoval::perform()
...
+   //
+   // As this is cheap to do and will reduce the overhead of analysis like UseDef and
+   // CompactLocals, identify the loads that have dead stores on all paths
+   // and zero them.
+   //
+   static const char *disableDeadStoreCleanup = feGetEnv("TR_DisableDeadStoreCleanup");
+   if (!disableDeadStoreCleanup)
+      {
+      TR_BitVector alwaysDead(comp()->trMemory()->currentStackRegion());
+      TR_Stack<TR::Block *> osrBlocks(trMemory(), 8, false, stackAlloc);
```
andrewcraik: A comment on why we can't process when we find them with a bitvector of already processed blocks would be handy here if this is the case...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1378,6 +1408,82 @@ int32_t TR_OSRExceptionEdgeRemoval::perform()
...
+      // frame.
+      for (cfgNode = cfg->getFirstNode(); cfgNode; cfgNode = cfgNode->getNext())
+         {
+         TR::Block *block = toBlock(cfgNode);
+         if (!block->isOSRCodeBlock())
+            continue;
+
+         // Find the prepareForOSR call
+         TR::TreeTop *tt = block->getFirstRealTreeTop();
+         while (tt && !(tt->getNode()->getNumChildren() > 0 && tt->getNode()->getFirstChild()->getOpCode().isCall()))
```
andrewcraik: Shouldn't we check that the call is the call we expect by checking the symref for safety?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1378,6 +1408,82 @@ int32_t TR_OSRExceptionEdgeRemoval::perform()
...
+
+         // Remove loads that are always dead
+         for (int32_t i = 0 ; i < prepareForOSR->getNumChildren(); ++i)
+            {
+            TR::Node *child = prepareForOSR->getChild(i);
+            if (child->getOpCode().isLoadVar() && alwaysDead.get(child->getSymbolReference()->getReferenceNumber()))
+               {
+               if (performTransformation(comp(), "%s: Remove dead OSR load [%p] from OSR code block_%d\n", OPT_DETAILS, child, block->getNumber()))
+                  {
+                  child->decReferenceCount();
```
andrewcraik: Shouldn't we just recreate the node in place because the node has an identical value by the nature of the transform and so any use will benefit from knowing it is zero...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1378,6 +1408,82 @@ int32_t TR_OSRExceptionEdgeRemoval::perform()
...
+                  prepareForOSR->setAndIncChild(i,
+                     TR::Node::createConstZeroValue(child, child->getSymbol()->getDataType()));
+                  }
+               else
+                  alwaysDead.reset(child->getSymbolReference()->getReferenceNumber());
```
andrewcraik: matching {}s please for symmetry

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1314,6 +1314,35 @@ TR_OSRLiveRangeAnalysis::optDetailString() const throw()
...
+   }
+
+void TR_OSRExceptionEdgeRemoval::removeDeadStores(TR::Block* osrBlock, TR_BitVector& dead)
+   {
+   for (TR::TreeTop *tt = osrBlock->getFirstRealTreeTop(); tt != osrBlock->getExit(); tt = tt->getNextTreeTop())
+      {
+      TR::Node *node = tt->getNode();
+      if (node->getOpCode().isStoreDirect() && node->storedValueIsIrrelevant() &&
+          dead.get(node->getSymbolReference()->getReferenceNumber()))
+         TR::TransformUtil::removeTree(comp(), tt);
```
andrewcraik: We should trace at this point...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1314,6 +1314,35 @@ TR_OSRLiveRangeAnalysis::optDetailString() const throw()
...
+      if (node->getOpCode().isStoreDirect() && node->storedValueIsIrrelevant())
+         _seenDeadStores->set(node->getSymbolReference()->getReferenceNumber());
+      }
+
+   if (inited)
+      dead &= *_seenDeadStores;
+   else
+      dead |= *_seenDeadStores;
+
+   return !_seenDeadStores->isEmpty();
```
andrewcraik: Trace would seem to be a good idea here...

andrewcraik:
Basic idea looks fine - a liberal sprinkling of appropriately guarded trace is warranted to make this code easier to debug

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Tracing has been added for the store removal and for finding the dead stores.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Added these in.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: A recreate to the corresponding type has been added. I could not find a similar example of this, but I believe this is correct.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: A check has been added for the symref, with a more specific check for the treetop node.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Added the comment. The blocks are revisited to clean up dead stores. A stack isn't required, but it seems like a cheap solution.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Added an assert for both helper functions.

andrewcraik:
LGTM with the updates - @vijaysun-omr over to you

vijaysun-omr:
Are the stores added at the start of the OSR code block ? Is that the "common" point where the stores can be added without duplicating for each OSR transition point ?

ncough:
Yes, dead stores will be placed after the `prepareForOSR` call in an OSR code block. Thanks to this, the analysis here only has to check the immediate predecessors to an OSR code block to find all of the dead stores that apply to its frame.

vijaysun-omr:
Why "after" the prepareForOSR call ? I would have guessed we place dead stores before the OSR call, no ?

ncough:
The dead stores in an OSRCodeBlock apply to its caller's frame, a feature added in https://github.com/eclipse/omr/pull/1326 and explained in the comment here: https://github.com/eclipse/omr/blob/2d351f73b65c7ebefa18203e0fb145b37f0140de/compiler/il/symbol/OMRResolvedMethodSymbol.cpp#L1908-L1945

The dead stores are placed after the `prepareForOSR` in case a symbol has been reused from the caller to translate the callee's frame. If the symbol appeared to be dead in the caller, placing the dead store before the callee's `prepareForOSR` could result in the slot that reused the symbol being invalid.

vijaysun-omr:
Makes sense for the caller's stores to be after the prepareForOSR call of the callee. I was missing that we were talking about the caller in one case and the callee in another. Thanks for clarifying.

PR #1763, opened 2017-10-02
Reduce OSRLiveRangeAnalysis overhead

ncough:
This is a set of changes to reduce OSRLiveRangeAnalysis overhead, mostly focused on skipping the analysis where possible. It consists of:

- When calculating liveness at a treetop level, skip blocks that contain no OSR points
- Remove allocations, left over from disabled sections of the analysis
- Add a missing option to disable the analysis
- Skip the analysis for bodies that don't contain OSR points at all
- Skip the analysis for bodies that don't have symbols that require it
- Add the option to solve pending push liveness in IlGen

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -847,6 +847,25 @@ OMR::Compilation::requiresAnalysisOSRPoint(TR::Node *node)
       }
    }
 
+/**
+ * To reduce OSR overhead, OSRLiveRangeAnalysis supports solving liveness analysis
+ * during IlGen, as an approximation of this information is commonly known at this
+ * stage.
+ *
+ * If this function returns true, OSRLiveRangeAnalysis expects pending push liveness
+ * information to be stashed in OSRData using addPendingPushLivenessInfo prior to
+ * its execution in IlGenOpts. Returning true without the correct information stashed
+ * will result in reduced performance and failed OSR transitions.
+ *
+ * If this function returns false, OSRLiveRangeAnalysis will perform liveness analysis
+ * on pending push symbols.
+ */
+bool
+OMR::Compilation::pendingPushLivenessDuringIlgen()
+   {
+   return false;
+   }
+
 bool
 OMR::Compilation::isProfilingCompilation()
    {
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -636,7 +636,11 @@ TR_OSRMethodData::TR_OSRMethodData(int32_t _inlinedSiteIndex, TR::ResolvedMethod
         _linkedToCaller(false),
         slot2ScratchBufferOffset(comp()->allocator()),
         _numSymRefs(0),
-        bcInfoHashTab(comp()->allocator()),  bcLiveRangeInfoHashTab(comp()->allocator()), argInfoHashTab(comp()->allocator()) {};
+        bcInfoHashTab(comp()->allocator()),
+        bcLiveRangeInfoHashTab(comp()->allocator()),
+        bcPendingPushLivenessInfoHashTab(comp()->allocator()),
+        argInfoHashTab(comp()->allocator())
+   {}
 
 TR::Block *
 TR_OSRMethodData::findOrCreateOSRCodeBlock(TR::Node* n)
@@ -790,6 +794,37 @@ TR_OSRMethodData::getArgInfo(int32_t byteCodeIndex)
       args = argInfoHashTab.DataAt(hashIndex);
       }
    return args;
+   }
+
+/*
+ * Add pending push live range info for a BCI.
+ *
+ * During IlGen, the live pending push symbol references may be available.
+ * They can be stashed against the BCI, so that OSRLiveRangeAnalysis can
+ * skip them, avoiding costly region and liveness analysis.
+ */
+void
+TR_OSRMethodData::addPendingPushLivenessInfo(int32_t byteCodeIndex, TR_BitVector *livenessInfo)
+   {
+   bcPendingPushLivenessInfoHashTab.Add(byteCodeIndex, livenessInfo);
+   }
+
+/*
+ * Get pending push live range info for a BCI. This represents all pending pushes
+ * that are live at a BCI. It will only be valid if they are stash here using
+ * addPendingPushLivenessInfo.
+ */
+TR_BitVector *
+TR_OSRMethodData::getPendingPushLivenessInfo(int32_t byteCodeIndex)
+   {
+   CS2::HashIndex hashIndex;
+   TR_BitVector* livenessInfo = NULL;
+   if (bcPendingPushLivenessInfoHashTab.Locate(byteCodeIndex, hashIndex))
+      {
+      livenessInfo = bcPendingPushLivenessInfoHashTab.DataAt(hashIndex);
+      }
+
+   return livenessInfo;
    }
 
 void
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -461,6 +461,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
 #ifdef J9_PROJECT_SPECIFIC
    {"disableOSRGuardRemoval",             "O\tdisable OSR guard removal",                      TR::Options::disableOptimization, osrGuardRemoval, 0, "P"},
 #endif
+   {"disableOSRLiveRangeAnalysis",        "O\tdisable live range analysis for on-stack replacement", SET_OPTION_BIT(TR_DisableOSRLiveRangeAnalysis), "F"},
    {"disableOSRLocalRemat",               "O\tdisable use of remat when inserting guards for on-stack replacement", SET_OPTION_BIT(TR_DisableOSRLocalRemat), "F"},
    {"disableOSRSharedSlots",              "O\tdisable support for shared slots in on-stack replacement", SET_OPTION_BIT(TR_DisableOSRSharedSlots), "F"},
    {"disableOutlinedNew",                 "O\tdo object allocation logic inline instead of using a fast jit helper",  SET_OPTION_BIT(TR_DisableOutlinedNew), "F"},
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -885,10 +885,25 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
             //
             _liveVars->empty();
             TR::Block *succ;
-            for (auto edge = block->getSuccessors().begin(); edge != block->getSuccessors().end(); ++edge)
+            auto edge = block->getSuccessors().begin();
+            for (; edge != block->getSuccessors().end(); ++edge)
                {
                succ = toBlock((*edge)->getTo());
                *_liveVars |= *liveLocals._blockAnalysisInfo[succ->getNumber()];
+               }
+
+            // If the block contains an OSR point, it should have an exception edge
+            // to an OSR catch block
+            for (edge = block->getExceptionSuccessors().begin(); edge != block->getExceptionSuccessors().end(); ++edge)
+               {
+               if (toBlock((*edge)->getTo())->isOSRCatchBlock())
+                  break;
+               }
+            if (edge == block->getExceptionSuccessors().end())
+               {
+               if (comp()->getOption(TR_TraceOSR))
+                  traceMsg(comp(), "skipping block_%d without OSR points", block->getNumber());
+               break;
                }
 
             if (comp()->getOption(TR_TraceOSR))
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -681,77 +681,324 @@ TR::TreeTop* findOSRHelperCall(TR::Compilation* comp, TR::Block * osrCodeBlock)
 // class TR_OSRLiveRangeAnalysis
 TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manager)
    : TR::Optimization(manager),
-     _deadVars(NULL),
-     _liveVars(NULL)
+     _liveVars(NULL),
+     _pendingPushSymRefs(NULL),
+     _sharedSymRefs(NULL),
+     _workBitVector(NULL)
    {}
 
-int32_t TR_OSRLiveRangeAnalysis::perform()
+bool TR_OSRLiveRangeAnalysis::shouldPerformAnalysis()
    {
-   if (comp()->getOption(TR_EnableOSR))
-      {
-      if (comp()->getOption(TR_DisableOSRLiveRangeAnalysis))
-         {
-         if (comp()->getOption(TR_TraceOSR))
-            traceMsg(comp(), "OSR is enabled but OSR live range analysis is not.\n");
-         return 0;
-         }
-      }
-   else
+   if (!comp()->getOption(TR_EnableOSR))
       {
       if (comp()->getOption(TR_TraceOSR))
-         traceMsg(comp(), "OSR Option not enabled -- returning from OSR live range analysis.\n");
-      return 0;
-      }
-
-   // Determine if the analysis is necessary
-   if (!canAffordAnalysis())
-      {
-      if (comp()->getOption(TR_TraceOSR))
-         {
-         traceMsg(comp(), "%s OSR live range analysis is not affordable for this method because of the expected cost\n",
-            optimizer()->getMethodSymbol()->signature(comp()->trMemory()));
-         traceMsg(comp(), "Returning...\n");
-         }
-
-      return 0;
+         traceMsg(comp(), "Should not perform OSRLiveRangeAnalysis -- OSR Option not enabled\n");
+      return false;
       }
    else if (comp()->isPeekingMethod())
       {
       if (comp()->getOption(TR_TraceOSR))
-         {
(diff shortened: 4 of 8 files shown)


compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -886,7 +1081,7 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
    block = comp()->getStartBlock();
    while (block)
       {
-      memset(_pendingSlotValueParents, 0, numBits*sizeof(NodeParentInfo *));
+      //memset(_pendingSlotValueParents, 0, numBits*sizeof(NodeParentInfo *));
```
andrewcraik: please remove deadcode, don't just comment it

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -842,38 +1043,32 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
...
+               {
+               liveLocalIndexToSymRefNumberMap[nextSymRef->getSymbol()->castToAutoSymbol()->getLiveLocalIndex()] = nextSymRef->getReferenceNumber();
+               nextSymRef->getSymbol()->setLocalIndex(0);
+               }
             }
          }
       }
 
-   _pendingSlotValueParents = (NodeParentInfo **)trMemory()->allocateStackMemory(numBits*sizeof(NodeParentInfo *));
+   //_pendingSlotValueParents = (NodeParentInfo **)trMemory()->allocateStackMemory(numBits*sizeof(NodeParentInfo *));
```
andrewcraik: please remove dead code don't just comment it

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1075,9 +1288,9 @@ void TR_OSRLiveRangeAnalysis::maintainLiveness(TR::Node *node,
          //
          if (local->getLocalIndex() == 0)
             {
-            if (_pendingPushVars->get(localIndex))
+            if (_pendingPushSymRefs.get(node->getSymbolReference()->getReferenceNumber()))
                {
-               if (_pendingSlotValueParents[localIndex])
+               if (0 && _pendingSlotValueParents[localIndex])
```
andrewcraik: This is bad form - if this is dead remove it.

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1140,7 +1353,7 @@ void TR_OSRLiveRangeAnalysis::maintainLiveness(TR::Node *node,
             local->setLocalIndex(local->getLocalIndex() + node->getReferenceCount());
             }
 
-         if (0 && _pendingPushVars->get(localIndex))
+         if (0 && _pendingPushSymRefs.get(node->getSymbolReference()->getReferenceNumber()))
```
andrewcraik: if this is dead remove it as well.

compiler/optimizer/OSRDefAnalysis.hpp
```diff
@@ -99,19 +99,30 @@ class TR_OSRLiveRangeAnalysis : public TR::Optimization
...
        TR_Liveness *liveLocals, TR_BitVector *liveVars, TR::Block *block);
    TR::TreeTop *collectPendingPush(TR_ByteCodeInfo bci, TR::TreeTop *pps, TR_BitVector *liveVars);
 
    TR_BitVector *_deadVars;
    TR_BitVector *_liveVars;
-   TR_BitVector *_pendingPushVars;
+
+   TR_BitVector _sharedSymRefs;
+   TR_BitVector _pendingPushSymRefs;
+   TR_BitVector _workBitVector;
```
andrewcraik: Can you provide a justification for why these need to be heap allocated and not stack allocated like _deadVars and _liveVars? This would seem to be almost an effective 'leak' of memory since these are only live in the optimization...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1254,34 +1467,42 @@ void TR_OSRLiveRangeAnalysis::buildOSRSlotSharingInfo(TR::Node *node, TR_BitVect
...
+   TR_BitVector *deadSymRefs = NULL;
+   if (!_deadVars->isEmpty())
       {
-      deadSymRefs = new (trHeapMemory()) TR_BitVector(maxSymRefNumber+1, trMemory(), heapAlloc);
+      // Load the existing BitVector
+      deadSymRefs = osrMethodData->getLiveRangeInfo(byteCodeIndex);
+      if (!deadSymRefs)
+         {
+         deadSymRefs = new (trHeapMemory()) TR_BitVector(0, trMemory(), heapAlloc);
+         allocated = true;
```
andrewcraik: why do we need allocated rather than just testing the NULLness of the field?

andrewcraik: Or something similar

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
       if (comp()->getOption(TR_TraceOSR))
-         traceMsg(comp(), "OSR Option not enabled -- returning from OSR live range analysis.\n");
-      return 0;
+         traceMsg(comp(), "Should not perform OSRLiveRangeAnalysis -- Not required because we are peeking\n");
+      return false;
+      }
+   else if (!comp()->supportsInduceOSR())
+      {
+      if (comp()->getOption(TR_TraceOSR))
+         traceMsg(comp(), "Should not perform OSRLiveRangeAnlysis -- OSR is not supported\n");
```
andrewcraik: OSR is not supported? don't you mean that we are not in a voluntary OSR mode or something like that?

compiler/control/OMROptions.cpp
```diff
@@ -460,6 +460,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
 #ifdef J9_PROJECT_SPECIFIC
    {"disableOSRGuardRemoval",             "O\tdisable OSR guard removal",                      TR::Options::disableOptimization, osrGuardRemoval, 0, "P"},
 #endif
+   {"disableOSRLiveRangeAnalysis",        "O\tdisable live range analysis for on-stack replacement", SET_OPTION_BIT(TR_DisableOSRLiveRangeAnalysis), "F"},
```
andrewcraik: I take it that a more normal disable of an opt pass doesn't work because of ilgen or something?

compiler/compile/OSRData.cpp
```diff
@@ -624,7 +624,10 @@ TR_OSRMethodData::TR_OSRMethodData(int32_t _inlinedSiteIndex, TR::ResolvedMethod
         _linkedToCaller(false),
         slot2ScratchBufferOffset(comp()->allocator()),
         _numSymRefs(0),
-        bcInfoHashTab(comp()->allocator()),  bcLiveRangeInfoHashTab(comp()->allocator()), argInfoHashTab(comp()->allocator()) {};
+        bcInfoHashTab(comp()->allocator()),
+        bcLiveRangeInfoHashTab(comp()->allocator()),
+        bcPendingPushLivenessInfoHashTab(comp()->allocator()),
+        argInfoHashTab(comp()->allocator()) {};
```
andrewcraik: if we are breaking this into two lines the indenting isn't right. I do see some strange HashIndex stuff because of the CS2 HashTable usage - would these be better as an std::map to simplify that code?

compiler/compile/OMRCompilation.cpp
```diff
@@ -848,6 +848,12 @@ OMR::Compilation::requiresAnalysisOSRPoint(TR::Node *node)
    }
 
 bool
+OMR::Compilation::pendingPushLivenessDuringIlgen()
```
andrewcraik: This deserves some doxygen to explain when it should be true and what implementing languages need to support before they can set it to true etc

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
+            {
+            containsAuto = true;
+            if (methodSymbol->sharesStackSlot(symRef))
+               _sharedSymRefs.set(symRef->getReferenceNumber());
+            }
          }
+      }
 
-      return 0;
+   TR_Array<List<TR::SymbolReference>> *pendingPushSymRefs = comp()->getMethodSymbol()->getPendingPushSymRefs();
```
andrewcraik: if we don't specifically need a TR_Array for perf reasons could we use an std::deque, std::list or similar with stack scoped memory to the same effect?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
+      return fullAnalysis(sharesParm, containsPendingPush);
+
+   return 0;
+   }
+
+/*
+ * The partial analysis is only concerned with solving which pending push temps
+ * are live at certain OSR points.
+ */
+int32_t TR_OSRLiveRangeAnalysis::partialAnalysis()
```
andrewcraik: This feels like it needs more tracing... it logs the start, but not much else...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -842,38 +1043,32 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
...
          }
       }
 
-   for (i=0;i<numPendingSlots;i++)
+   if (containsPendingPushes)
       {
-      List<TR::SymbolReference> symRefsAtThisSlot = (*pendingPushSymRefs)[i];
+      for (i=0;i<numPendingSlots;i++)
+         {
+         List<TR::SymbolReference> symRefsAtThisSlot = (*pendingPushSymRefs)[i];
```
andrewcraik: Should this copy the list or be a reference?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -899,12 +1094,27 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
             //
             _liveVars->empty();
             TR::Block *succ;
-            for (auto edge = block->getSuccessors().begin(); edge != block->getSuccessors().end(); ++edge)
+            auto edge = block->getSuccessors().begin();
+            for (; edge != block->getSuccessors().end(); ++edge)
```
andrewcraik: I'd consider a while loop here...

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
+      traceMsg(comp(), "OSR reaching live range analysis can be done\n");
+
+   bool containsAuto = false, sharesParm = false, containsPendingPush = false;
+
+   // Detect autos, pending push temps and whether there is a shared parm slot
+   TR::ResolvedMethodSymbol *methodSymbol = optimizer()->getMethodSymbol();
+   TR_Array<List<TR::SymbolReference>> *autosListArray = methodSymbol->getAutoSymRefs();
+   for (uint32_t i = 0; autosListArray && i < autosListArray->size(); ++i)
+      {
+      List<TR::SymbolReference> autosList = (*autosListArray)[i];
```
andrewcraik: Should this copy or be a reference?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
+         TR_BitVector *ppInfo = osrMethodData->getPendingPushLivenessInfo(bci.getByteCodeIndex());
+         pendingPushLiveRangeInfo(node, ppInfo, _pendingPushSymRefs, osrPoint, osrMethodData);
+         pendingPushSlotSharingInfo(node, ppInfo, _sharedSymRefs, osrPoint);
          }
-      return 0;
       }
-   else
+
+   // Certain modes require the method entry to be an OSR point
+   if (comp()->isOutermostMethod() && comp()->getHCRMode() == TR::osr)
```
andrewcraik: Feels like this should have a trace statement

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -699,75 +699,280 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
...
+   {
+   if (liveSymRefs && !liveSymRefs->isEmpty())
+      {
+      TR::SymbolReferenceTable *symRefTab = comp()->getSymRefTab();
+      TR_ByteCodeInfo& bcInfo = osrPoint->getByteCodeInfo();
+
+      if (trace())
+         traceMsg(comp(), "Shared PP slots at OSR point [%p] at %d:%d\n", node, bcInfo.getCallerIndex(), bcInfo.getByteCodeIndex());
+
+      _workBitVector = *liveSymRefs;
```
andrewcraik: Not sure about the allocator used with this copy - I feel something else would make the lifetime of _workBitVector more explicit...

andrewcraik:
The spirit of the changes looks reasonable, but there are a number of implementation details that we need to look at before this is ready IMO.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: `supportsInduceOSR` will return false for a variety of modes that don't support OSR, such as DLT, with tracing for the reasoning behind it. I'll extend the message to make this clearer.

compiler/control/OMROptions.cpp
ncough: OSRLiveRangeAnalysis is required under certain modes for OSR transitions to be correct. It is left as an option to allow for these cases to be detected.

andrewcraik: Ok that makes sense - thanks for the clarification.

compiler/optimizer/OSRDefAnalysis.cpp
andrewcraik: Ok that would be great - thanks!

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Rather than disabling the copy propagation allocations, it has been removed entirely, addressing the four comments concerning it.

compiler/optimizer/OSRDefAnalysis.hpp
ncough: There was no suitable justification. They have been moved to the stack. Moreover, `workBitVector` has been used to replace `deadVars`.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Based on the prior implementation of this method, it was believed if there was at least one dead var, an allocation should be done. It has now been delayed until it is also confirmed that the var is one of interest.

compiler/compile/OMRCompilation.cpp
ncough: A comment has been added.

compiler/compile/OSRData.cpp
ncough: It would likely be simpler and easy to replace. It should be noted that the existing CS2 hash table will not overwrite entries, keeping the original value instead. This behaviour should be matched, or OSR logic will have to be updated. I believe this should be done in a separate PR.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Tracing has been added to the treetop walk and the method entry. The liveness and shared slot functions already had tracing.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: All of these should have been references, an error made prior to this change. They have all been updated to use references.

compiler/optimizer/OSRDefAnalysis.cpp
ncough: Both loops have been change to while loops.

andrewcraik:
Thanks @ncough for all the clean-up - the new version is looking good. I think this should work as intended - @vijaysun-omr over to you.

andrewcraik:
@vijaysun-omr ping?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -885,12 +885,27 @@ int32_t TR_OSRLiveRangeAnalysis::perform()
...
+                  break;
+               }
+            if (edge == block->getExceptionSuccessors().end())
+               {
+               if (comp()->getOption(TR_TraceOSR))
+                  traceMsg(comp(), "skipping block_%d without OSR points", block->getNumber());
+               break;
+               }
+
             if (comp()->getOption(TR_TraceOSR))
```
vijaysun-omr: Is'nt there some maintainLiveness logic later on that is'nt predicated on an OSR point being present in the block ? If so, why is it okay to skip that code ?

ncough: `maintainLiveness` is only concerned with retaining liveness between treetops in a block. When a new block is seen, `_liveVars` is reset according to the liveness results for the block's successors. Therefore, blocks can be processed out of order and skipped.

PR #1764, opened 2017-10-03
Change J9 prefix to OMR

charliegracie:
This PR cleans up a lot of code. These changes are a first step to removing the J9 `namespace` from OMR.  These references to J9 are historical as that is where the code was originally donated from but to ease consumption of the project and clean up the code it makes sense to try and remove all of them.

This PR does not fix all occurrences but is the first step in the cleanup.


charliegracie:
genie-omr build all

dnakamura:
@charliegracie  As part of this effort do we want to start renaming the libraries? Ie `j9thrstatic`, `j9pool`, `j9hashtable`, etc? It's probably not worth effort to rename them on the makefile side, but I see no reason we cant rename them in cmake

charliegracie:
Yes that is a great idea @dnakamura.  I agree that there is no reason to update the makefile versions.

Issue #1760, opened 2017-09-30
New datatypes and operations for AutoSIMD on x86


IBMJimmyk:
Currently AutoSIMD support on x86 is fairly limited. It only supports Double type add and mul operations.

The proposed changes will significantly expand AutoSIMD support on x86 to include the following data type and operation combinations:

Add - Int, Long, Float, Double
Sub - Int, Long, Float, Double
Mul - Int, Float, Double
Div - Float, Double
And - Int, Long
Or - Int, Long
Xor - Int, Long

Furthermore, support for reduction operations will also be added.

The increase in AutoSIMD support will come from updating the x86 evaluators to handle the new math and logical operations. Also, a new evaluator will be created for getvelem to implement reduction support.

These evaluator changes require previously unused vector instructions so changes to X86Ops.ins are also necessary.

PR #1766, opened 2017-10-03
New datatypes and operations for AutoSIMD on x86

IBMJimmyk:
Currently AutoSIMD support on x86 is fairly limited. It only supports
Double type add and mul operations.

This change sets significantly expands AutoSIMD support on x86 to include
the following data type and operation combinations:

Add - Int, Long, Float, Double
Sub - Int, Long, Float, Double
Mul - Int, Float, Double
Div - Float, Double
And - Int, Long
Or - Int, Long
Xor - Int, Long

Furthermore, support for reduction operations was also added.

getSupportsOpCodeForAutoSIMD() was updated to reflect these changes.

The increase in AutoSIMD support came from updating the x86 evaluators to
handle the new math and logical operations. Also, a new evaluator was
created for getvelem to implement reduction support.

These evaluator changes required previously unused vector instructions so
changes to X86Ops.ins was also made to support new vector instructions.

Issue: #1760


andrewcraik:
given the changes to the x86 codegen I think getting the thoughts of @0dvictor on this would be a good idea since he works in that area a lot.

IBMJimmyk:
I think @gita-omr @0dvictor and @andrewcraik wanted to review this before it got committed.

IBMJimmyk:
There's one thing about about my changes that I wanted to highlight because it also came up in a previous change I submitted.

In OMRTreeEvaluator.cpp I use the instructions PADDD and PMULLD instead of "PADDDRegReg" or "PMULLDRegReg" which I use for everything else. The reason for this is PADDD and PMULLD already existed prior to my changes and were in use by code outside of the AutoSIMD sections of the code. I felt that the changes in the PR were already large enough and updating the areas to use "PADDDRegReg" or "PMULLDRegReg" instead would be better left to a future PR.

0dvictor:
Maybe we should combine `VectorLogicalEvaluator` into `FloatingPointAndVectorBinaryArithmeticEvaluator`. They are similar enough that, by consolidating them, we can apply same optimizations on. For example, we can potentially use three-operand forms of the instructions in the same place.

andrewcraik:
The spirit of the changes looks fine - @0dvictor are you happy with the level of trace? Otherwise my main question would be what kind of perf validation this has been through since a big change like this could impact performance for the better or the worse.

0dvictor:
The trace information looks fine to me. The evaluators are straightforward enough to find out where an instruction is generated.
Performance measurements definitely need be measured further. Maybe guard these new logics with an environment variable and delete it later?

IBMJimmyk:
After talking with @0dvictor I merged VectorLogicalEvaluator into FloatingPointAndVectorBinaryArithmeticEvaluator to address a previous comment.

0dvictor:
Thank you. LGTM.

0xdaryl:
Please fix the linter error:
```
./../compiler/x/codegen/OMRCodeGenerator.cpp:984:73: error: Implicit this
      receivers are prohibited in extensible classes
  ...TR::Float || dt == TR::Double || (dt == TR::Int32 && comp()->cg()->getX8...
                                                          ^
                                                          self()->
```

You are already in the `CodeGenerator` class so a `self()->getX86ProcessorInfo()` call should be sufficient to get at what you need.

IBMJimmyk:
I fixed the error. I changed the line to use self()->getX86ProcessorInfo().

gita-omr:
Daryl, these are all valid questions. I will let @IBMJimmyk answer.

IBMJimmyk:
For testing, I created unit tests that attempted to trigger AutoSIMD for each of the data type and operation combinations. The test itself would confirm a correct calculation. On top of that, I would create a log file for the compiled method and manually verify that the correct vector instructions were generated. A similar process was perform to ensure that the getvelem evaluator was working as expected. I ran a small test to perform a reduction operation that would trigger AutoSIMD and manually confirmed that the correct code was being generated.

For performance, we tried out a version of OpenJ9 with this AutoSIMD feature added in and saw no regressions in HiBench. Furthermore, IBM helped out by running some of their own Java tests and they did not see any regressions either.

0xdaryl:
Additional Z and P testing is not required as all changes are within files only built on x86.

PR #1772, opened 2017-10-04
Fix comptest failures on Linux 390

charliegracie:
Shifting a 32 bit integer by more than 31 bits is undefined behaviour in
the C++ spec. Currently the OMR compiler  masks the shift value by 0x1F.
This means a shift of 32 or larger on a 32 bit integer effectively
returns the same value. To ensure the C++ version matches the compiled
version of the test the C++ version needs to also mask the shift value.

### DIFF
--- a/fvtest/compilertriltest/ShiftAndRotateTest.cpp
+++ b/fvtest/compilertriltest/ShiftAndRotateTest.cpp
@@ -56,6 +56,34 @@ template <typename T> static T rotate(T a, int32_t b)
    return a;
    }
 
+template <typename Left> static Left shift_left(Left left, int32_t right)
+   {
+   int32_t r = right;
+   if (sizeof(Left) <= sizeof(int32_t))
+      {
+      r = (r & (8 * sizeof(int32_t) - 1));
+      }
+   else
+      {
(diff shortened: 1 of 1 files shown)


charliegracie:
Currently the tril tests are only built via CMake and require C++11 so running the genie build should not be required.

fvtest/compilertriltest/ShiftAndRotateTest.cpp
```diff
@@ -56,6 +56,34 @@ template <typename T> static T rotate(T a, int32_t b)
    return a;
    }
 
+template <typename Left, typename Right> static Left shift_left(Left left, Right right)
```
Leonardo2718: Since, the "right" argument to a shift opcode must always be `Int32`, can you get rid of the `Right` template parameter and hard code `int32_t ` instead? This will make it consistent with the implementation of `rotate`.

charliegracie: Changed

vijaysun-omr:
I was looking for the same issue that Leo already brought up that the second child of a shift is an int no matter what the type of the shift. Looks okay otherwise. Merging.

PR #1774, opened 2017-10-04
Support for building Test Compiler on s390x using CMake

rbanerjee:
This commit makes it possible to have a cmake based build
for the Test Compiler on zLinux_s390x.

### DIFF
--- a/fvtest/compilertest/CMakeLists.txt
+++ b/fvtest/compilertest/CMakeLists.txt
@@ -53,6 +53,12 @@ elseif(OMR_ARCH_POWER)
 		p/codegen/Evaluator.cpp
 		${OMR_ROOT}/compiler/p/env/OMRDebugEnv.cpp
 	)
+elseif(OMR_ARCH_S390)
+	list(APPEND COMPILERTEST_FILES 
+		z/codegen/Evaluator.cpp
+		z/codegen/TestCodeGenerator.cpp
+		${OMR_ROOT}/compiler/z/env/OMRDebugEnv.cpp
+	)
 endif()
 
 create_omr_compiler_library(
@@ -62,6 +68,7 @@ create_omr_compiler_library(
 		JITTEST
 		TEST_PROJECT_SPECIFIC
 		PROD_WITH_ASSUMES
+		INCLUDES ${OMR_ROOT}/fvtest
 )
 add_executable(compilertest
 	tests/main.cpp


fvtest/compilertest/CMakeLists.txt
```diff
@@ -54,11 +54,18 @@ elseif(OMR_ARCH_POWER)
...
+		z/codegen/Evaluator.cpp
+		${OMR_ROOT}/compiler/z/env/OMRDebugEnv.cpp
+	)
 endif()
 
 create_omr_compiler_library(NAME       testcompiler 
                             OBJECTS    ${COMPILERTEST_FILES}
-							DEFINES JITTEST TEST_PROJECT_SPECIFIC PROD_WITH_ASSUMES)
+							DEFINES JITTEST TEST_PROJECT_SPECIFIC PROD_WITH_ASSUMES
+			                                INCLUDES ${OMR_ROOT}/fvtest)
```
charliegracie: Is this the right change or should the `#include` should be changed similar to how it is in jitbuilder?

charliegracie:
@mgaudet can you review please?

fvtest/compilertest/CMakeLists.txt
mgaudet: Which `#include` in JitBuilder? IMO, includes in the compiler technology shouldn't include the project name. Now, that's not universally held, but my pref would be to alter the include path rather than alter the `#include` -- @0xdaryl may have a differing opinion here. 

In general, I'd like to get rid of the `#include "project/..."` includes.

mgaudet:
LGTM, though I do think there's a stylistic question to be determined around includes here.

fvtest/compilertest/CMakeLists.txt
charliegracie: In `omr/fvtest/compilertest/z/codegen/TestCodeGenerator.hpp" there is the following include `#include "compilertest/codegen/TestCodeGenerator.hpp"`.  This include in my opinion should either be `#include "codegen/TestCodeGenerator.hpp"` if the include path is modified to have `omr/fvtest/compilertest' near the beginning or `#include "fvtest/compilertest/codegen/TestCodeGenerator.hpp"` since that is effectively how the include works in the similar file in jitbuilder.

mgaudet: So the last one, I think is wrong. The first one though, I think is how it ought to be. 

```
#include "codegen/TestCodeGenerator.hpp" // < -- Right
#include "compilertest/codegen/TestCodeGenerator.hpp" // < -- wrong
#include "fvtest/compilertest/codegen/TestCodeGenerator.hpp" // < -- more wrong
```

@rbanerjee In light of the above, do you want to just change this PR to instead correct the include to the first line above?

charliegracie:
@rbanerjee I believe the codegenerator part of this commit is breaking x86 compiles for some reason.  Maybe to get the zLinux builds working you can remove that part again and open a new PR to change the codegenerator so we can get CMake based zLinux builds working.

rbanerjee:
@charliegracie ptal

The Codegen code inside Compiler Test still needs some cleaning up. In particular, the `*CodeGenerator` files inside `fvtest/compilertest` don't extend the functionality
    provided by the CodeGenerator extensible class in any meaningful way. I will do that in a separate PR.

rbanerjee:
@0xdaryl  sorry about that. Should be good now.

rbanerjee:
@charliegracie  Not sure why the OSX build timed out last time around. Seems to be fine now.

@0xdaryl  ptal

PR #1778, opened 2017-10-05
WIP: Convert OMR_PLATFORM_COMPILE_OPTIONS to flag append

mgaudet:
Currently in our AIX toolconfig we have `OMR_PLATFORM_COMPILE_OPTIONS` set
to (among other flags) `-qlanglvl=extended`.

In the compiler technology, the `TR_CXX_COMPILE_OPTIONS` flag is used to
upgrade C++ compilations to the `-qlanglvl=extended0x`, as the compiler
technology takes advantage of some of the C++11 features provided by
that compiler option.

The problem is that using `add_compile_options`, the arguments are
appended to the compilation string very late. This produces a
compilation string like

    ... -qlanglvl=extended0x ... -qlanglvl=extended

In most traditional command line processors, XLC included, the last
option superceeds earlier options, so this breaks AIX compilation of 
the compiler component. 

This commmit stops using `add_compile_options`, replacing it with
appending to both the `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`. This is at
least mildly distasteful, however, I'm not sure how else to be able to
override definitions from the 'global' platform options in a
subcomponent, if the global options are set with `add_compile_options`.

(A side note: One initial option that you might want to consider is
having the compiler technology add its own options with
add_compile_options. The issue here is that there is no way to add a
compile option using that command for only one language, which is
a requirement here, without using generator expressions. Now, it may be
acceptable to do so for XLC, however, language specific generator
expressions are documented to not work for MSVC, and so we have
historically avoided them)

### DIFF
--- a/cmake/modules/OmrPlatform.cmake
+++ b/cmake/modules/OmrPlatform.cmake
@@ -58,7 +58,13 @@ macro(omr_platform_global_setup)
 
 	set(OMR_PLATFORM_GLOBALLY_INITIALIZED 1)
 
-	add_compile_options(${OMR_PLATFORM_COMPILE_OPTIONS})
+	omr_append_flags(CMAKE_C_FLAGS
+		${OMR_PLATFORM_COMPILE_OPTIONS}
+	)
+
+	omr_append_flags(CMAKE_CXX_FLAGS
+		${OMR_PLATFORM_COMPILE_OPTIONS}
+	)
 
 	omr_append_flags(CMAKE_C_FLAGS
 		${OMR_PLATFORM_C_COMPILE_OPTIONS}


mgaudet:
WIP because I'd like some feedback from @rwy0717 and @dnakamura at least.

rwy0717:
I think this will cause `-pthread` to be passed to clang++ during linking, which is going to give us that unused argument warning. How about we remove the language level from the global platform flags, and set it on a target-by-target basis.

Also, have we looked into using the CXX_STANDARD property? Were we sure that it doesn't work on XLC?

mgaudet:
That seems like a sensible middle ground approach :) I'll work on that

Regarding CXX_STANDARD, it has been my experience it's not successful nearly often enough to rely on.

mgaudet:
I took a meander down the road of seeing if we could get using CXX_STANDARD, as I'd love to use it. The issue is that in most version of CMake, the XLC support is rudimentary at best, and missing entirely at worst. Much of the support for CXX_STANDARD in [XLC was merged only in April of 2017](https://gitlab.kitware.com/cmake/cmake/commit/6555286c5551e3e051ba9814e9f02f62b2963446). 

I was also incapable of building a newer version of CMake that included those changes, so I was unable to test them out. 

```
[ 94%] Linking CXX executable ../bin/ctest
ld: 0711-781 ERROR: TOC overflow. TOC size: 68728       Maximum size: 65536
collect2: ld returned 12 exit status
make: 1254-004 The error code from the last command is 1.
```

rwy0717:
Sadness that it won't build. You can probably compile and link with `-qpic=large` to fix the error. It's the option we use in openj9. via: https://www.ibm.com/developerworks/rational/library/overview-toc-aix/

If `CXX_STANDARD` support is patchy, maybe we could just fill in the missing parts with OmrPlatform. I'm thinking literally a copy-paste of that patch, inside `toolconfig/xlc.cmake`, could suffice.

mgaudet:
So... maybe? The issue is that as near as I can tell, what has to happen there is essentially the same as doing a CMake toolchain for XLC -- (which might be totally sane). 

-sigh- 

I have a simple-ish patch using the above proposal I need to test tomorrow. We will see what comes of that.

mgaudet:
#1790 is the simple approach I described above.

devin122:
@mgaudet  re: `ld: 0711-781 ERROR: TOC overflow` on gcc it would be caused by using `-fpic` instead of `-fPIC` there is some similar flag for xlc. I think you need to use `-qpic=large` instead of just `-qpic`

PR #1779, opened 2017-10-05
OMR updates to address z/TPF build errors

pelemie:
VPHandlers.cpp - rename swap() to VPHswap() as z/TPF already has an
opensource swap() function in our libstdc++ library.
TypedAllocator.hpp - compiler couldn't find a matching matching function
for call to "construct()".
Signed-off-by: pete lemieszewski <lemie@us.ibm.com>

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -4360,14 +4360,6 @@ static int32_t longNumberOfLeadingZeros (int64_t n) {return leadingZeroes (n);}
 static int32_t longNumberOfTrailingZeros (int64_t n) { return trailingZeroes(n); }
 static int32_t longBitCount (int64_t n) {return populationCount(n);}
 
-template <typename T>
-void swap (T& a, T& b)
-   {
-   T tmp = b;
-   b = a;
-   a = tmp;
-   }
-
 
 template <typename FUNC, typename FUNC2, typename T>
 void addValidRangeBlockOrGlobalConstraint (OMR::ValuePropagation *vp,
@@ -4382,7 +4374,7 @@ void addValidRangeBlockOrGlobalConstraint (OMR::ValuePropagation *vp,
 
    if (pLow > pHigh)
       {
-      swap (pLow, pHigh);
+      std::swap (pLow, pHigh);
       }
 
    if (vp->trace())


compiler/env/TypedAllocator.hpp
```diff
@@ -125,6 +125,14 @@ class typed_allocator : public typed_allocator<void, Alloc>
       }
 
    static void construct(pointer p, const_reference prototype) { new(p) value_type(prototype); }
+#if defined (OMRZTPF)
+/*
+ * z/TPF update to avoid compiler error when calling construct:
+ * "...from opensource/include46/g++/bits/stl_uninitialized.h:547:6: error:"
+ * "no matching function for call to �TR::typed_allocator<unsigned int, TR::Region&>::construct(unsigned int*)�..."
+ */
+   static void construct(pointer p) { new(p) value_type();  }
```
dsouzai: What compiler version are you guys using? From this link https://software.intel.com/en-us/forums/intel-many-integrated-core/topic/475660:

>gcc changed the call in __uninitialized_default_n_a (and __uninitialized_default_a) from:
__alloc.construct(std::__addressof(*__cur));
to:
__traits::construct(__alloc, std::__addressof(*__cur));
between 4.7.0 and 4.7.3.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -4382,7 +4386,11 @@ void addValidRangeBlockOrGlobalConstraint (OMR::ValuePropagation *vp,
 
    if (pLow > pHigh)
       {
+#if defined (OMRZTPF)
+           VPHswap (pLow, pHigh);
```
dsouzai: There is too much indentation in this line.

compiler/env/TypedAllocator.hpp
pelemie: we are using gcc 4.6.3 (built as a cross compiler for our platform)

pelemie: I hadn't noticed this writeup (or the one in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51626).  the suggestions in the bugzilla item suggests
"The fix for libstdc++ would be to avoid calling construct with a single argument, or to backport a minimal version of allocator_traits from 4.7. " And, I can do that + withdraw this change.
Did this/similar error occur with a gcc.4.4 build?

dsouzai: The concern I have that because `static void construct(pointer p, const_reference prototype) { new(p) value_type(prototype); }` copy constructs into `p` from `prototype`, there could be issues when dealing with non primitive types because some code in OMR may depend on the copy constructor being invoked for functional correctness. It's unlikely since for the most part TR::typed_allocator is supposed to be used when dealing with STL containers, but there's no guarantee it won't cause issues.

dsouzai: >Did this/similar error occur with a gcc.4.4 build?
I'm not sure, for the most part I've only dealt with OMR via OpenJ9 so I believe the compiler being used there is gcc 4.8.

dsouzai: @charliegracie @mstoodle Do you guys know if we've ever built OMR with gcc 4.4?

mgaudet: I believe mostly OMR has been compiled with GCC 4.4.7. So yes, but not the earliest of the 4 series.

dsouzai: Odd that this issue has never shown up before then, though I suppose this is also dependent on the version of libstdc++.

mgaudet: Ehhh -- I mean, it's not unheard of for compiler bugs to have quite a dependence on context.

pelemie: Thanks Everyone! ok, I found the patch to libstdc++ and it is in... gcc 4.6.3! 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51626
Not sure why we didn't get this patch in our version of gcc 4.6.3. So the basic problem statement is [C++0x] can't use C++98 allocators with -std=c++0x. Without checking, I'm guessing that gcc 4.4.7 didn't cause the problem because there wasn't enough integrated c++11 development. Certainly, it is cleared up in gcc 4.8. I've verified that applying the patch fixes the problem so I'll withdraw this patch once I update the indentation in VPHandlers.cpp (comment below.).

compiler/optimizer/VPHandlers.cpp
pelemie: Thanks... I hope I updated/squashed the changes correctly!

mgaudet:
maybe just universally renaming `swap` to `VPSwap` makes sense.

pelemie:
I always guess wrong :) I originally had that and thought it wouldn't pass muster.

dsouzai:
The changes LGTM. You'll probably have to change the commit message though, specifically the `Squashed updates` line since I doubt the committers will accept that hehe.

pelemie:
The template function that was swap has the same signature in my opensource copy of move.h

0xdaryl:
genie-omr build all

0xdaryl:
Looks fine to me.  @andrewcraik , any objections?

PR #1781, opened 2017-10-05
Add and use spec specific CMake caches

charliegracie:
Add a cache per spec so that each spec can selectively turn OMR features
on / off. The CMake builds running on Travis  and Appveyor had to be changed to pass
the SPEC as part of the env.


charliegracie:
@dnakamura, @rwy0717 and @mgaudet what are your thoughts on this?

mgaudet:
Personally, I dislike the proliferation of `SPEC`, and `SPEC`-like-concepts, but, I and hard pressed to answer what a much better answer is.

charliegracie:
I agree with you but I also do not know how to do it any better.  We need the ability to compile 32, 64 bit and 64 bit compressed versions of each platform as a minimum.  I can change the wording SPEC to cache if that makes it any better.  We can either have a cache for each concrete build we would like to create or we can specify 1-10 command line options for each build we would like to do. Having the cache seems more convenient to me.

rwy0717:
I like to think of the caches as curated sets of user-facing features. Features are platform-agnostic, even if they aren't always available. We avoid exposing platform configuration to the user. I think that in general we've accomplished that. The big exception to this is the `OMR_ENV_DATA32` toggle, which we discovered was not easy to handle in any other way. I would argue that the bit width is not a feature, but compressed references is.

It makes sense to me to start with two caches: `most_features.cmake` (in essence the travis or  appveyor cache) and `most_features+crefs.cmake`. If you try to enable compressed references on an unsupported (ie 32bit) platform, cmake should just error out, like any other feature.

Since the bit width doesn't belong in the cache, and it's just one option on some builds, it makes sense to just set it on the command line. After all, It's just a convenience for hammering the target CPU architecture, when you don't want to use toolchain file.

Organizing this way would also help stop the platforms*features explosion of caches.

charliegracie:
Ok that sort of makes sense to me.  I guess downstream projects that have very fixed sets of requirements can create their own caches if it makes it easier for them.  I will close this for now.

devin122:
@rwy0717  Does it make sense to have `most_features.cmake` and most_features+crefs.cmake`? I would argue that it would make more sense to toggle compressed refs from the command line, since the two cmake files would be identical except for 1 line

PR #1782, opened 2017-10-05
Improve CMake JitBuilder support for Linux P

charliegracie:
JitBuilder requires 2 defines which were not being set and the
compilation was missing a file for Linux P.

### DIFF
--- a/jitbuilder/CMakeLists.txt
+++ b/jitbuilder/CMakeLists.txt
@@ -53,6 +53,7 @@ elseif(OMR_ARCH_POWER)
 		p/runtime/CodeSync.cpp
 		p/runtime/AsmUtil.spp
 		p/runtime/CodeDispatch.spp
+		${OMR_ROOT}/compiler/p/env/OMRDebugEnv.cpp
 	)
 endif()
 
@@ -61,7 +62,7 @@ endif()
 create_omr_compiler_library(
 	NAME    jitbuilder
 	OBJECTS ${JITBUILDER_OBJECTS}
-	DEFINES PROD_WITH_ASSUMES
+	DEFINES PROD_WITH_ASSUMES JITTEST
 )
 
 # Add interface path so that include paths propagate.


jitbuilder/CMakeLists.txt
```diff
@@ -61,7 +62,7 @@ endif()
 create_omr_compiler_library(
 	NAME    jitbuilder
 	OBJECTS ${JITBUILDER_OBJECTS}
-	DEFINES PROD_WITH_ASSUMES
+	DEFINES PROD_WITH_ASSUMES JITTEST JITBUILDER_SPECIFIC
```
mgaudet: `JITBUILDER_SPECIFIC` isn't used anywhere; It is referenced in the Makefiles, but that's incorrect, and should be removed. 

Hypothetically `JITBUILDER_PROJECT_SPECIFC`would be the correct macro, but I don't think it's used anywhere at the moment.

mgaudet: Definitely we should open an issue about getting rid of JITTEST, but I think this is the right choice right now.

jitbuilder/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ elseif(OMR_ARCH_POWER)
 		p/runtime/CodeSync.cpp
 		p/runtime/AsmUtil.spp
 		p/runtime/CodeDispatch.spp
+		${OMR_ROOT}/compiler/p/env/OMRDebugEnv.cpp
```
mgaudet: This should be added to `compiler/p/CMakeLists.txt`, as it's shared across projects.

charliegracie: This is how compilertest does it. I would prefer to keep both products working the same way.  If that is the correct way then compiletest should also be changed.  I would prefer to make a follow on PR to change both if that is the decision.

jitbuilder/CMakeLists.txt
charliegracie: I removed JITBUILDER_SPECIFIC.

jitbuilder/CMakeLists.txt
mgaudet: KK.

mgaudet:
You may want to change the commit message now that you've only added 1 define.

mgaudet:
So, I just took a look in more detail, and it seems it's really the test compiler that's in error here; so it may actually be better to just move the TestCompiler's include into the appropriate compiler CMakeLists.txt 

    10:54 $ ag OMRDebugEnv.cpp 
    compiler/env/CMakeLists.txt
    28:	${CMAKE_CURRENT_SOURCE_DIR}/OMRDebugEnv.cpp
    
    compiler/x/CMakeLists.txt
    62:	${CMAKE_CURRENT_SOURCE_DIR}/env/OMRDebugEnv.cpp
    
    compiler/z/CMakeLists.txt
    57:	${CMAKE_CURRENT_SOURCE_DIR}/env/OMRDebugEnv.cpp
    
    fvtest/compilertest/CMakeLists.txt
    54:		${OMR_ROOT}/compiler/p/env/OMRDebugEnv.cpp

Every other `OMRDebugEnv` is being handled by the compiler support module.

charliegracie:
I already have that change in another set of commits to further improve the CMake compiler support

0xdaryl:
@mstoodle , do you want to review this PR or can I merge?

PR #1788, opened 2017-10-05
Quiet output from LinkedList test

charliegracie:
Currently this test outputs a lot of information to the console when it
works correctly. Only print information when there is a failure. Also a
DEBUG_OUTPUT compile time flag as added to re-enable some of the more
verbose output to aid in debugging. 


charliegracie:
genie-omr build all

mstoodle:
LGTM

PR #1790, opened 2017-10-06
WIP: Split Language standard out in CMake

mgaudet:
Currently in our AIX toolconfig we have OMR_PLATFORM_COMPILE_OPTIONS set
to (among other flags) -qlanglvl=extended.

In the compiler technology, the TR_CXX_COMPILE_OPTIONS flag is used to
upgrade C++ compilations to the -qlanglvl=extended0x, as the compiler
technology takes advantage of some of the C++11 features provided by
that compiler option.

The problem is that using add_compile_options, the arguments are
appended to the compilation string very late. This produces a
compilation string like

    ... -qlanglvl=extended0x ... -qlanglvl=extended

In most traditional command line processors, XLC included, the last
option superceeds earlier options, so this breaks AIX compilation of
the compiler component.

This commit hoists the language standard out to two new platform
controllable variables: OMR_C_LANGUAGE_STANDARD and
OMR_CXX_LANGUAGE_STANDARD. These flags are put in the command line
via modification of the CMAKE_C_FLAGS and CMAKE_CXX_FLAGS respectively.

As a result, the compiler technology remains free to append a higher
standard later (or, if desired, we could later us omr_remove_flag,
to replace it entirely).

### DIFF
--- a/cmake/modules/OmrPlatform.cmake
+++ b/cmake/modules/OmrPlatform.cmake
@@ -62,10 +62,12 @@ macro(omr_platform_global_setup)
 
 	omr_append_flags(CMAKE_C_FLAGS
 		${OMR_PLATFORM_C_COMPILE_OPTIONS}
+		${OMR_C_LANGUAGE_STANDARD}
 	)
 
 	omr_append_flags(CMAKE_CXX_FLAGS
 		${OMR_PLATFORM_CXX_COMPILE_OPTIONS}
+		${OMR_CXX_LANUAGE_STANDARD} 
 	)
 
 	omr_append_flags(CMAKE_EXE_LINKER_FLAGS
--- a/cmake/modules/platform/toolcfg/gnu.cmake
+++ b/cmake/modules/platform/toolcfg/gnu.cmake
@@ -34,6 +34,9 @@ else()
 	)
 endif()
 
+set(OMR_C_LANGUAGE_STANDARD "") 
+set(OMR_CXX_LANGUAGE_STANDARD "") 
+
 # Testarossa build variables. Longer term the distinction between TR and the rest
 # of the OMR code should be heavily reduced. In the mean time, we keep
 # the distinction
--- a/cmake/modules/platform/toolcfg/msvc.cmake
+++ b/cmake/modules/platform/toolcfg/msvc.cmake
@@ -36,6 +36,9 @@ else()
 	set(TARGET_MACHINE "i386")
 endif()
 
+set(OMR_C_LANGUAGE_STANDARD "") 
+set(OMR_CXX_LANGUAGE_STANDARD "") 
+
 list(APPEND OMR_PLATFORM_LINKER_OPTIONS
 	-subsystem:console
 	-machine:${TARGET_MACHINE}
--- a/cmake/modules/platform/toolcfg/xlc.cmake
+++ b/cmake/modules/platform/toolcfg/xlc.cmake
@@ -38,12 +38,16 @@ else()
 	)
 endif()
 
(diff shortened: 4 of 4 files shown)


mgaudet:
This is another take on the same problem addressed by #1778

cmake/modules/OmrPlatform.cmake
```diff
@@ -62,10 +62,12 @@ macro(omr_platform_global_setup)
 
 	omr_append_flags(CMAKE_C_FLAGS
 		${OMR_PLATFORM_C_COMPILE_OPTIONS}
+		${OMR_C_LANGUAGE_STANDARD}
 	)
 
 	omr_append_flags(CMAKE_CXX_FLAGS
 		${OMR_PLATFORM_CXX_COMPILE_OPTIONS}
+		${OMR_CXX_LANUAGE_STANDARD}
```
rwy0717: There's a typo here.

dnakamura:
I would much rather see this done relying more heavily on CMakes infrastructure (ie its notion of language level). It doesn't have the lang level info in the current releases (although it looks like it was added in 3.9), I think we may be able to just set some variables and have it work. Will spend some time looking into it

rwy0717:
I agree, I'd rather use cmake's built in support than roll our own. We [kind-of talked](https://github.com/eclipse/omr/pull/1778#issuecomment-334597491)  about "filling in" the missing variables from inside our toolcfg files. I actually think that's a good idea.

I also don't really understand where this patch is going. Why not just append the standard to the `OMR_PLATFORM_{C,CXX}_OPTIONS` variables directly?

mgaudet:
> I also don't really understand where this patch is going. Why not just append the standard to the `OMR_PLATFORM_{C,CXX}_OPTIONS` variables directly? 

If I do that, then the language level changes for all projects, not just the compiler, which I suspect is even less desirable. 

I'm OK with marking this as WIP pending @dnakamura's investigation of the CMake infrastructure

dnakamura:
Looking at things it gets a little weird. On the newer cmakes it will only use `-qlanglvl=extended` for c++, however it looks like we are using it for c and c++. It also seems that for makefile builds we are using the c++ compiler to build c files so that could be part of the reason too.  I dont really like the idea of pushing `-qlanglvl=extended` to the c compiler , and conversely I'm not really sure how to replicate the current behavior where we compile c files as c++

dnakamura:
That being said, the following snippet could be added to `cmake/modules/platform/toolcfg/xlc.cmake` to get the basis of language standards level support (at least for c++)
```
if(NOT CMAKE_CXXX_STANDARD_DEFAULT)
	set(CMAKE_CXX_STANDARD_DEFAULT 98)
	set(CMAKE_CXX98_EXTENSION_COMPILE_OPTION "-qlanglvl=extended")
endif()
```

mgaudet:
Maybe I screwed something up, but I did briefly flirt with that sort of thing, but it seemed like `CMAKE_CXX98_EXTENSION_COMPILE_OPTION` wasn't being paid attention to.

dnakamura:
@mgaudet  yes. that was driving me nuts. Thats what I have spent my day so far figuring out.  Turns out that language standards
 are completely ignored unless `CMAKE_<LANG>_STANDARD_DEFAULT` is defined
```
  const char* defaultStd =
    this->Makefile->GetDefinition("CMAKE_" + lang + "_STANDARD_DEFAULT");
  if (!defaultStd || !*defaultStd) {
    // This compiler has no notion of language standard levels.
    return;
  }```

mgaudet:
hahahahah. Wow. Ok. 

If that works, I'm totally fine with closing this PR in lieu of your solution, which seems just flat out better.

PR #1791, opened 2017-10-06
Move initialization of JitOptTest _strategy

mgaudet:
This avoids a mis-sized buffer. Without this change, the buffer is
allocated before the correct size has been determined, and is not
resized before usage.

Issue found with -fsanitize=address, running the Tril tests.


charliegracie:
genie-omr build all

PR #1796, opened 2017-10-06
Support folding immutable array shadows

liqunl:
If we know an array is not going to be changed nor is its content, we
can fold the load of its element. This is done by adding a bit in
KnownObjectTable to indicate whether a known object is an array with
immutable content. When VP sees an array shadow load from such array,
it will improve the array shadow to an ImmutableArrayShadow, a new
symbol added for this optimzation, and ask transformIndirectLoadChain
to fold it. ImmutableArrayShadow is also served as an indicator to
transformIndirectLoadChain that it is safe to fold an array shadow with
such symbol.

### DIFF
--- a/compiler/il/Aliases.cpp
+++ b/compiler/il/Aliases.cpp
@@ -647,6 +647,22 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
             aliases->set(symRefTab->getArrayShadowIndex(_symbol->getDataType().scalarToVector()));
             }
 
+         if (_symbol->isArrayShadowSymbol() &&
+             !symRefTab->aliasBuilder.immutableArrayElementSymRefs().isEmpty())
+            {
+            if (!aliases)
+               aliases = new (aliasRegion) TR_BitVector(bvInitialSize, aliasRegion, growability);
+
+            TR::DataType type = _symbol->getDataType();
+            TR_BitVectorIterator bvi(symRefTab->aliasBuilder.arrayElementSymRefs());
+            int32_t symRefNum;
+            while (bvi.hasMoreElements())
+               {
+               symRefNum = bvi.getNextElement();
+               if (symRefTab->getSymRef(symRefNum)->getSymbol()->getDataType() == type)
+                  aliases->set(symRefNum);
+               }
+            }
 
          if (_symbol->isArrayShadowSymbol() &&
              supportArrayRefinement &&
--- a/compiler/optimizer/VPConstraint.cpp
+++ b/compiler/optimizer/VPConstraint.cpp
@@ -616,6 +616,13 @@ TR_YesNoMaybe TR::VPKnownObject::isJavaLangClassObject()
    return _isJavaLangClass ? TR_yes : TR_no;
    }
 
+bool TR::VPKnownObject::isArrayWithConstantElements(TR::Compilation * comp)
+   {
+   TR::KnownObjectTable *knot = comp->getKnownObjectTable();
+   TR_ASSERT(knot, "TR::KnownObjectTable should not be null");
+   return knot->isArrayWithConstantElements(_index);
+   }
+
 TR_YesNoMaybe TR::VPClassType::isArray()
    {
    if (_sig[0] == '[')
--- a/compiler/optimizer/VPConstraint.hpp
+++ b/compiler/optimizer/VPConstraint.hpp
@@ -958,6 +958,7 @@ class VPKnownObject : public TR::VPFixedClass
    virtual TR::VPConstraint *intersect1(TR::VPConstraint *other, OMR::ValuePropagation *vp);
 
    virtual TR_YesNoMaybe isJavaLangClassObject();
+   virtual bool isArrayWithConstantElements(TR::Compilation * comp);
 
    virtual bool mustBeEqual(TR::VPConstraint *other, OMR::ValuePropagation *vp);
    virtual bool mustBeNotEqual(TR::VPConstraint *other, OMR::ValuePropagation *vp);
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -355,8 +355,22 @@ static bool tryFoldCompileTimeLoad(
       TR::KnownObjectTable *knot = vp->comp()->getKnownObjectTable();
       TR::Node *baseExpression = NULL;
       TR::KnownObjectTable::Index baseKnownObject = TR::KnownObjectTable::UNKNOWN;
-      for (TR::Node *curNode = node->getFirstChild(); !baseExpression; curNode = curNode->getFirstChild())
+      TR::Node *curNode = node;
+
+      for (; !baseExpression; )
          {
+         // Track last curNode because its symref might be improved to ImmutableArrayShadow
+         TR::Node* prevNode = curNode;
+
+         // Look pass aladd/aiadd to get the array object for array shadows
+         if (curNode->getSymbol()->isArrayShadowSymbol()
+             && curNode->getFirstChild()->getOpCode().isArrayRef()
+             && curNode->getFirstChild()->getSecondChild()->getOpCode().isLoadConst())
+            {
+            curNode = curNode->getFirstChild()->getFirstChild();
+            }
+         else curNode = curNode->getFirstChild();
+
          bool curIsGlobal;
          TR::VPConstraint *constraint = vp->getConstraint(curNode, curIsGlobal);
          if (!curIsGlobal)
@@ -375,6 +389,31 @@ static bool tryFoldCompileTimeLoad(
             }
          else if (constraint->getKnownObject())
             {
+            TR::Symbol* prevNodeSym = prevNode->getSymbol();
+            // Improve regular array shadows from arrays with immutable content
+            if (prevNodeSym->isArrayShadowSymbol() &&
+                // UnsafeShadow responds true to above question, but the type of the array element might not match the type of the symbol,
+                // it needs to be derived from the array
+                !prevNodeSym->isUnsafeShadowSymbol() &&
+                !vp->comp()->getSymRefTab()->isImmutableArrayShadow(prevNode->getSymbolReference()))
+               {
+               if (constraint->getKnownObject()->isArrayWithConstantElements(vp->comp()))
+                  {
+                  // Use the DataType of the symbol because it represents the element type of the array
+                  TR::SymbolReference* improvedSymRef = vp->comp()->getSymRefTab()->findOrCreateImmutableArrayShadowSymbolRef(prevNodeSym->getDataType());
+                  if (vp->trace())
+                     traceMsg(vp->comp(), "Found arrayShadow load %p from array with constant elements %p\n", prevNode, curNode);
+                  if (performTransformation(vp->comp(), "%sUsing ImmutableArrayShadow symref #%d for node %p\n", OPT_DETAILS, improvedSymRef->getReferenceNumber(), prevNode))
+                     prevNode->setSymbolReference(improvedSymRef);
+
+                  // Alias info of calls are computed in createAliasInfo, which is only called if alias info is not available or is invalid.
+                  // Invalidate alias info so that it can be recomputed in the next optimization that needs it
+                  vp->optimizer()->setAliasSetsAreValid(false);
+                  }
+               else
+                  break;
+               }
+
             baseExpression  = curNode;
             baseKnownObject = constraint->getKnownObject()->getIndex();
             if (vp->trace())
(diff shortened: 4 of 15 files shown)


jdmpapin:
How is the folding restricted to cases in which the array's contents are immutable?

andrewcraik:
@liqunl can you address the question from @jdmpapin please?

liqunl:
@jdmpapin The change supports the folding of array shadows, however, whether to fold array shadows and under what circumstances an array shadow is foldable is a decision that each language has to make.
I updated my commit message to avoid confusion.

andrewcraik:
I don't see the bit that allows a language to opt into the folding or not - this looks like it will kick in for everything - could you explain the language extension point please?

liqunl:
@andrewcraik The language extension point is `TR::TransformUtil::transformIndirectLoadChain()`. If the base expression, in this case the array, has known object constraint, we'll call `TR::TransformUtil::transformIndirectLoadChain()` to transform it and the most specific implementation will be called.

andrewcraik:
Right but I can have a known object below an arrayshadow and that isn't enough to transform it - I need to know the array is ok as well which I don't see being checked.

liqunl:
As per discussion with @andrewcraik and @jdmpapin, we're going to add a bit in the known object constraint to indicate whether the content of the array is also immutable and we'll check that bit before proceeding to the actual transformation.

liqunl:
@jdmpapin @andrewcraik Any more comments? I'll launch do a personal build and some tests before the merge.
@AdamBrousseau Do you have any idea why the CI build failed with the following error? It looks like an infra issue.
`/usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! (RuntimeError)`

compiler/optimizer/VPConstraint.hpp
```diff
@@ -934,7 +934,7 @@ class VPKnownObject : public TR::VPFixedClass
 
    protected:
    VPKnownObject(TR_OpaqueClassBlock *klass, TR::Compilation * comp, TR::KnownObjectTable::Index index, bool isJavaLangClass)
-      : TR::VPFixedClass(klass, comp, KnownObjectPriority), _index(index), _isJavaLangClass(isJavaLangClass) {}
+      : TR::VPFixedClass(klass, comp, KnownObjectPriority), _index(index), _isJavaLangClass(isJavaLangClass), _isArrayWithConstantElements(false) {}
```
andrewcraik: I am not sure I'm a fan of hard coding this to false - I could see a default parameter to say false unless you say otherwise? @jdmpapin thoughts since you are familiar with this constraint.

compiler/optimizer/VPConstraint.cpp
```diff
@@ -616,6 +616,16 @@ TR_YesNoMaybe TR::VPKnownObject::isJavaLangClassObject()
    return _isJavaLangClass ? TR_yes : TR_no;
    }
 
+void TR::VPKnownObject::setIsArrayWithConstantElements()
+   {
+   _isArrayWithConstantElements = true;
+   }
+
+TR_YesNoMaybe TR::VPKnownObject::isArrayWithConstantElements()
```
andrewcraik: why does this return yes/no/maybe when it can only be yes/no?

compiler/optimizer/VPConstraint.hpp
```diff
@@ -958,6 +958,8 @@ class VPKnownObject : public TR::VPFixedClass
    virtual TR::VPConstraint *intersect1(TR::VPConstraint *other, OMR::ValuePropagation *vp);
 
    virtual TR_YesNoMaybe isJavaLangClassObject();
+   virtual TR_YesNoMaybe isArrayWithConstantElements();
+   virtual void setIsArrayWithConstantElements();
```
andrewcraik: Setters on constraints are generally a bad idea - we need to create a new constraint to mutate it - I'd prefer to handle the extra parameter in the constructor somehow (either explicitly or a default param)

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2236,6 +2252,13 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
       TR::Symbol *sym = node->getSymbol();
       switch (sym->getRecognizedField())
          {
+         case TR::Symbol::Java_lang_invoke_VarHandle_handleTable:
+            {
+            TR::VPConstraint *vhConstraint = vp->getConstraint(node, isGlobal);
```
andrewcraik: This warrants some kind of trace

compiler/optimizer/VPConstraint.hpp
jdmpapin: Since constraints are supposed to be immutable (like you mentioned elsewhere), it will need to be provided when the constraint is constructed

liqunl:
Update:
After another round of discussion with Devin and Andrew, we'll try another approach, which is:
1. We'll add a bit in the known object able for each entry to indicate whether the object is a constant array with constant content
2. VP will create the known object constraint with _isArrayWithConstantElements bit set for this known object
3. We'll add a new symbol called something like immutableArrayShadow and it is only to be created by VP for the purpose of folding array shadow loads
4. VP will improve a regular array shadow load from an array with constant elements to an immutableArrayShadow load, and ask transformIndirectLoadChain to fold it
5. transformIndirectLoadChain will look for immutableArrayShadow on the load node as a safety check and fold the shadow if the node has that symbol
6. The trigger point of this optimization is the bit in the known object table. Each language is responsible for setting this bit if they want to fold array shadow loads from specific objects

liqunl:
@andrewcraik @jdmpapin The change is ready for review, a personal build is still running and will update the result once it finishes.

liqunl:
Sorry, I need to add aliasing, so putting this PR back to WIP.

liqunl:
@jdmpapin @andrewcraik Can you review my change when you have time? This change supports aliasing between immutable array shadow and regular array shadow, but we probably don't need that since immutable array shadow indicates the array has been created and there are no stores to it. Moving the loads around has no side effect since it's loading a constant anyway.

liqunl:
The CI build failure is due to a job running overtime, it is not a hang nor a functional failure, will run the test locally.

liqunl:
With the discussion with Devin and Andrew, we want to make the concept of `ImmutableArrayShadow` general in the way that `ImmutableArrayShadow` implies when the program executes to a load of `ImmutableArrayShadow`, it is guaranteed that the same value will be returned if one reads that memory again, and more specifically, no store will be issued to that memory.  
`ImmutableArrayShadow` does not imply that the array has been allocated at compile time.
Thus, the aliasing is needed. I'll add the aliasing between the `ImmutabelArrayShadow` and the `KnownObjectSymRef` that's created for it.

liqunl:
@jdmpapin @andrewcraik The changes are ready for review, can you take a look?

compiler/il/Aliases.cpp
```diff
@@ -647,6 +647,22 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
             aliases->set(symRefTab->getArrayShadowIndex(_symbol->getDataType().scalarToVector()));
             }
 
+         if (_symbol->isArrayShadowSymbol() &&
+             !symRefTab->aliasBuilder.immutableArrayElementSymRefs().isEmpty())
```
jdmpapin: Because a usual array shadow shares its `Symbol` with the corresponding immutable array shadow, and both have `reallySharesSymbol` set, I think these aliases should already be included via `setSharedShadowAliases()`

compiler/optimizer/VPHandlers.cpp
```diff
@@ -375,6 +386,22 @@ static bool tryFoldCompileTimeLoad(
             }
          else if (constraint->getKnownObject())
             {
+            // Improve regular array shadows from arrays with immutable content
+            if (node->getSymbol()->isArrayShadowSymbol() &&
+                !vp->comp()->getSymRefTab()->isImmutableArrayShadow(node->getSymbolReference()))
```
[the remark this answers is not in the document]
liqunl: `if (constraint->getKnownObject()->isArrayWithConstantElements(vp->comp()))` will check the array child. 
`aloadi <immutable-array-shadow>` will be improved to a symref with a known object index but not necessarily an immutable array. So `node`'s symref will remain the same

compiler/il/Aliases.cpp
liqunl: Putting our discussion here:
`setSharedShadowAliases` only search for symrefs generated for non-array shadows with the same symbol, and `reallySharesSymbol` is needed for other types of aliasing such as `canKill`, so we have to keep `reallySharesSymbol` flag set as well as the loop iterating through `aliasBuilder.arrayElementSymRefs` to look for the alias set.

andrewcraik:
@liqunl and @jdmpapin has your discussion concluded or are further updates to this being worked on?

liqunl:
@andrewcraik Just made some changes that need to be reviewed.

andrewcraik:
@jdmpapin are you good with this change now?

liqunl:
@jdmpapin My personal build has passed, do you still have concerns with this change?

andrewcraik:
LGTM from me - @vijaysun-omr I think this is now ready for your consideration - some fairly big changes to add the new concept of immutable array shadows and infrastructure for compile-time folding.

vijaysun-omr:
I'm unclear about

"With the discussion with Devin and Andrew, we want to make the concept of ImmutableArrayShadow general in the way that ImmutableArrayShadow implies when the program executes to a load of ImmutableArrayShadow, it is guaranteed that the same value will be returned if one reads that memory again, and more specifically, no store will be issued to that memory.
ImmutableArrayShadow does not imply that the array has been allocated at compile time.
Thus, the aliasing is needed."

If it is guaranteed that there cannot be any stores issued to that memory, then I am unsure why we need aliasing. Could you please elaborate ?

I'm also a bit intrigued by the whole notion of an immutable array shadow to begin with and had the same question as Devin early on in this PR, namely how is this limited to cases when the array's contents are immutable. I know you said that the situation in which we can know that an array's contents are immutable is dependent on the language, but am I right in thinking it's based on some specific semantic knowledge about the program (and array) in question that the compiler developer has based on their familiarity with the code in every case (at least currently) ? i.e. is the compiler ever able to "prove" that an array's contents are immutable ?

andrewcraik:
The intended use of this shadow, for the time being, is cases where we are dealing with special implementation internal arrays. The motivational case here would for VarHandles in Java - the handle has an array of MethodHandles that implement the various VarHandle operations. The contents of this array will not be changed (by design) and we need to be able to dereference the elements of the array at compile time to get hold of the handle required at a particular VarHandle callsite to inline through the handle and get good performance. A second place we could use them would be for the array of characters inside of a String - strings are immutable in that they will not mutable to represent a different string at compile-time and so if we see an operation offsetting into the character array of the String we could dereference into it at compile-time.

It is non-trivial to prove an array's contents are never modified - the current approach is for the compiler to substitute an immutable array shadow when we know for sure, based on language implementation details/semantics that the particular array can be safely folded. As such there is no proof mechanism at the present time.

I believe the aliasing is necessary should we ever be able to see the point where an immutable array is being initialized or potentially initialized. In the initialization code the array would appear as a mutable array so to ensure correct behavior the aliasing is required in general would be my line on this. @jdmpapin may have additional insights to share since he has studied this in more detail than I have recently.

liqunl:
The concept of immutable array shadow should imply that in the program execution order, there is no store to that memory after a load off an immutable array shadow, i.e. storing to an immutable array shadow is not allowed (we can make TreeVerification to abort the compilation if it sees such store). However, we cannot prove that an array's contents are never modified. There can be a store to a regular array shadow which points to the same memory. The onus is whoever creates `ImmutableArrayShadow` or  `ArrayWithConstantConstant` knows for sure that the array content will remain unchanged once initialized based on the language and specific semantic knowledge about the program.

vijaysun-omr:
Okay, I can see why some aliasing might be needed. I assume we do not alias immutable array shadows to each other but do alias all of them to the original (mutable) array shadow. Do we invalidate aliasing information at the end of an optimization that creates a new immutable array shadow so that aliasing info can be rebuilt or is this not required for some reason ?

liqunl:
Immutable array shadows used to access array element with the same type share the same symref, they're self aliased. Aliasing between immutable array shadows does not affect optimizations because those are loads. They are also aliased to regular/mutable array shadow with the same type.

For your second question: we don't invalidate aliasing information at the end of an optimization that creates a new immutable array shadow. I'm not aware of any API in `SymbolReferenceTable` that does that. AFAIK, we don't keep alias set around or stash it for next optimization; we request one when we need it. Is there anything I'm missing? If so, can you explain?

vijaysun-omr:
I was thinking of

optimizer()->setAliasSetsAreValid(false);

that gets called by optimizations to signal to the optimizer infrastructure that one or more of the fundamental bit vectors (e.g. symbol references that belong to a certain category) that are used to construct alias sets on demand may need to be recomputed. If alias sets are invalidated as shown above, the createAliasInfo in symreftab should get called. I am not sure if any of those fundamental bit vectors need to be updated or not in your case.

liqunl:
Creating an `ImmutableArrayShadow` symref will affect bit vectors related to array shadows, methods and calls. Since the bit vectors for methods and calls are kept between optimizations and will only be updated if they're invalid, so I added a line in VP to invalidate the alias info after a new symref is created.

We do not keep the alias info for other symrefs, which means we'll recompute it everytime it is requested. The aliasing for the `ImmutableArrayShadow` is handled by `getUseDefAliasesBV`

vijaysun-omr:
Thanks for making the change to invalidate aliasing in VP that is the only opt that creates these immutable array shadow sym refs at the moment.

vijaysun-omr:
This looks okay to me. Merging.

Issue #1801, opened 2017-10-11
Some files in omr/util/hookable use "dos" line feeds which cause issue on AIX


groeges:
Some of the files omr/util/hookable use "dos" line feeds which cause issue on AIX.
This was identified when trying to build OpenJ9 which uses omr where an error: 
```
ld -o ../omr/../libj9hookable29.so hookable.o ut_j9hook.o copyright.o   -lj9utilcore -lj9pool -lj9thr29  -L. -L../omr/.. -L../omr/../lib  -lc_r -lC_r -lm -lpthread -b64 -G -bnoentry -bernotok -bmap:j9hookable29.map -bE:j9hookable29.exp
ld: 0711-319 WARNING: Exported symbol not defined: J9HookInitializeInterface
ld: 0711-319 WARNING: Exported symbol not defined: omrhook_lib_control
ld: 0711-244 ERROR: No csects or exported symbols have been saved.
../omr/omrmakefiles/rules.mk:284: recipe for target '../omr/../libj9hookable29.so' failed
```
This was caused by the j9hookable29.exp which was created by
```
sh ../omr/omrmakefiles/generate-exports.sh xlc j9hookable29 ../omr/util/hookable/hookable.exportlist j9hookable29.exp
```
having the "dos" line feeds (ie ^M) in the file - which are in the ../omr/util/hookable/hookable.exportlist, which caused the `ld` command to not find the symbols in the hookable.o object file. 

Converting the ../omr/util/hookable/hookable.exportlist to be unix compatible (using dos2unix) resolves the issue. This also sees to work on a Windows system so having the file in this format seems to be OK.

groeges:
Fixed by PR: https://github.com/eclipse/omr/pull/1802

PR #1802, opened 2017-10-11
Change to not use "dos" line breaks

groeges:
Some files in util/hookable were modified to not use "dos" line breaks,
ie ^M which caused issues when building omr for OpenJ9 on an AIX
platform

Signed-off-by: Steven Groeger <groeges@uk.ibm.com>


groeges:
This is a PR for issue: https://github.com/eclipse/omr/issues/1801

DanHeidinga:
@charliegracie @0xdaryl @rwy0717 Can on the of the OMR committers restart the travis build?  The clang job fails for what looks like a machine config issue:
```
The command "if [ $TRAVIS_OS_NAME = osx ]; then brew install ccache ; fi" failed and exited with 1 during .
```

charliegracie:
Travis OS X jobs that use homebrew are all broken with the latest updates to the build infrastructure by Travis.  @rwy0717 is working on a solution.  ALL PRs that currently have the homebrew failures will need to be rebased on the changes in his PR once it gets accepted

charliegracie:
genie-omr build all

charliegracie:
accepting without rebasing for the OS X fix since all other platforms worked successfully and this PR contains no OS X specific changes.

PR #1806, opened 2017-10-11
Aborted CS and RS overflow

amicic:
Handle a special case of aborted Concurrent Scavenger when Remembered
Set is in overflow. This is not much different from handling it when RS
is not overflowed - we just clear RS. All heap fixup (restoring
references to forwarded objects and self-forwarded objects) occurs in
the following global GC and fixup hook.

Also, fixign a bug in Parallel Heap Walker that was ignoring walk flags.


charliegracie:
genie-omr build all

PR #1808, opened 2017-10-12
GC: Reorganize forge structures

rwy0717:
Move AllocationCategory to the OMR::GC namespace. Leave
MM_MemoryStatistics as is, it needs to be forward-declarable in the
hookgen headers. However, add an OMR::GC::MemoryStatistics typedef.

While I'm here, add a GC_HEAP category that aliases the old JAVA_HEAP category.

/cc @charliegracie @ktbriggs 


charliegracie:
genie-omr build all

charliegracie:
@rwy0717 does this change cause any downstreem projects to have to change their code to consume this change?

charliegracie:
genie-omr build all

rwy0717:
This is compatible openj9 (thanks to the typedefs), but once this goes through, I will be updating openj9 to use the new identifiers. We'll eventually be removing the backwards-compatible typedefs, and at that point, any other downstream project will require a code change.

PR #1810, opened 2017-10-12
Add static debug counters for loop replicator

yanluo7:
Add static debug counters for different type of loop replication failures.
Fixed minor logic bug to return proper values in heuristics() functions.

Signed-off-by: Yan Luo <yan_luo@ca.ibm.com>

### DIFF
--- a/compiler/optimizer/LoopReplicator.cpp
+++ b/compiler/optimizer/LoopReplicator.cpp
@@ -60,6 +60,17 @@ TR_LoopReplicator::TR_LoopReplicator(TR::OptimizationManager *manager)
    _maxNestingDepth = 0;
    }
 
+
+//Add static debug counter for a given replication failure
+static void countReplicationFailure(char *failureReason, int32_t regionNum)
+   {
+   //Assemble format string: "LoopReplicator/<failureReason>/%s/(%s)/region_%d"
+   TR::DebugCounter::incStaticDebugCounter(TR::comp(), TR::DebugCounter::debugCounterName(TR::comp(),
+      "LoopReplicator/%s/%s/(%s)/region_%d", failureReason,
+      TR::comp()->getHotnessName(TR::comp()->getMethodHotness()),
+      TR::comp()->signature(), regionNum));
+   }
+
 int32_t TR_LoopReplicator::perform()
    {
    // mainline entry
@@ -155,6 +166,7 @@ int32_t TR_LoopReplicator::perform(TR_Structure *str)
    TR::Block *entryBlock = region->getEntryBlock();
    if (entryBlock->isCold())
       {
+      countReplicationFailure("ColdLoop", region->getNumber());
       dumpOptDetails(comp(), "region (%d) is a cold loop\n", region->getNumber());
       return 0;
       }
@@ -229,6 +241,8 @@ int32_t TR_LoopReplicator::perform(TR_Structure *str)
       _loopType = doWhile;
       return replicateLoop(region, branchNode);
       }
+
+   countReplicationFailure("UnsupportedLoopStructure", region->getNumber());
 
    dumpOptDetails(comp(), "loop (%d) does not conform to required form & will not be replicated\n",
                     region->getNumber());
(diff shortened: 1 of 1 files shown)


charliegracie:
@yanluo7 please remove the Task line from the PR description and the Commit message.  Thanks

compiler/optimizer/LoopReplicator.cpp
```diff
@@ -147,6 +147,8 @@ int32_t TR_LoopReplicator::perform(TR_Structure *str)
 
    if (!region->isNaturalLoop())
       {
+      TR::DebugCounter::incStaticDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "LoopReplicator/NotNaturalLoop/%s/(%s)/region_%d",
+         comp()->getHotnessName(comp()->getMethodHotness()), comp()->signature(), region->getNumber()));
```
jdmpapin: This case is unlikely ever to be interesting, unless perhaps `region` is an improper region. For example, I think it would always count at least `LoopReplicator/NotNaturalLoop/*/(*)/region_0`

jdmpapin: Is it possible to use shorter lines? It's hard to read the diffs because they scroll horizontally

compiler/optimizer/LoopReplicator.cpp
```diff
@@ -331,6 +338,8 @@ int32_t TR_LoopReplicator::replicateLoop(TR_RegionStructure *region,
    TR::TreeTop *lastTT = cBlock->getLastRealTreeTop();
    if (!lastTT->getNode()->getOpCode().isBranch())
       {
+      TR::DebugCounter::incStaticDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "LoopReplicator/NoBranchFoundInLoop/%s/(%s)/region_%d",
+         comp()->getHotnessName(comp()->getMethodHotness()), comp()->signature(), region->getNumber()));
```
jdmpapin: These are getting pretty repetitive. Could they be moved into a function to clean up these sites? e.g.

    countUnreplicated("NoBranchFoundInLoop", region);

compiler/optimizer/LoopReplicator.cpp
```diff
@@ -60,6 +60,23 @@ TR_LoopReplicator::TR_LoopReplicator(TR::OptimizationManager *manager)
...
+   {
+   //Assemble format string: "LoopReplicator/<failureReason>/%s/(%s)/region_%d"
+   char *formatPre = "LoopReplicator/";
+   char *formatPost = "/%s/(%s)/region_%d";
+   //Make the buffer sufficiently large.
+   int32_t formatStringLength = strlen(formatPre) + strlen(failureReason) + strlen(formatPost) + 20;
+   char formatString[formatStringLength];
+   strcpy(formatString, formatPre);
+   strcat(formatString, failureReason);
+   strcat(formatString, formatPost);
```
jdmpapin: Rather than put `failureReason` into the format string, it would be better to fill it in `printf`-style into `LoopReplicator/%s/%s/(%s)/region_%d`

compiler/optimizer/LoopReplicator.cpp
```diff
@@ -729,7 +748,7 @@ bool TR_LoopReplicator::heuristics(LoopInfo *lInfo)
    if (!lInfo->_replicated)
       dumpOptDetails(comp(), "no side entrance found into trace; no replication will be performed\n");
 
-   return true;
+   return lInfo->_replicated;
```
andrewcraik: These are functional changes (this and the other return further down - could you just document here if you observed a problem form this and what this change did since it could have an impact on perf etc and I'd like to understand the problem case.

jdmpapin: The return value of `heuristics` only affects trace messages and the result of `TR_LoopReplicator::perform`. The result of `perform` seems to be unused. (It feeds into the result of `Optimizer::performOptimization`, which is itself unused except to feed into itself recursively.)

andrewcraik: Ok sounds good.

yanluo7: Echoing Devin, this does not affect performance, only enhancing correctness by making the caller of heuristics methods generating the correct traceMsg. I can add some comment in the code.

andrewcraik:
@vijaysun-omr over to you

compiler/optimizer/LoopReplicator.cpp
andrewcraik: code comment not required - just wanted to document the justification of the change in the PR - I'm fine with it especially in light of the comments from @jdmpapin

vijaysun-omr:
Looks safe to me; adding in debug counters for a particular condition. Merging.

PR #1813, opened 2017-10-17
Increase the limit of constant data snippet

r30shah:
Increase the size of maximum allowable data snippet from
64 bytes to 128 bytes.


r30shah:
@fjeremic  This commit increases the limit of constant data snippet to 128 bytes. This is needed if we want to use more than 3 onsite cache slots for this https://github.com/eclipse/openj9/pull/310 feature. Can you please review?

fjeremic:
I'm fine with merging this but we should really think about simplifying this code. We could just dynamically allocate the right amount.

0xdaryl:
genie-omr build zlinux

PR #1814, opened 2017-10-17
Fix slot assertions logging JitBuilder compiles

mstoodle:
We added more strenuous assertions when looking up symbol names
by slot number and that exposed a number of problems logging
JitBuilder compiles because there are a few places where we
did not register the symbol name after creating a new local slot.
Apparently, we never tested the new assertions while logging.

Spots fixed by this commit:
   1) CreateLocalArray
   2) CreateLocalStruct
   3) Copy
   4) IlValue when it stores a value needed in another block


charliegracie:
genie-omr build plinux

mstoodle:
Any reason you specifically decided to test on plinux, @charliegracie ? :)

charliegracie:
@mstoodle no sorry copy and paste from another comment. 
genie-omr build zlinux

PR #1816, opened 2017-10-17
Fix abort during JitBuilder compile logging

mstoodle:
There are cases when the code generate manufactures a symbol reference
and cannot assign an appropriate cpIndex (e.g. converting an indirect
access to a stack allocated array or object into a "more" direct access
via the stack pointer). In these cases, if the JIT is logging its
output, the JIT will ask MethodBuilder for the name of a symbol at
slot -1 which naturally causes an assertion. This commit returns
the name "Unknown" for slot -1 because this assertion only occurs
during JIT logging and it's usually better to allow the logging
to continue than to abort prematurely just because such a symbol
was created and logged.


mstoodle:
genie-omr build all

PR #1817, opened 2017-10-17
Fix for missing memory fence

0dvictor:
A memory fence need be applied when applying patching fence.


0xdaryl:
x86 only.  No genie build needed.

0xdaryl:
Its not clear to me why this fence (and similar assembly fences in the AsmUtil.inc files) doesn't apply to 32-bit.  I realize this wasn't the problem being addressed by this fix, but now that my eyes are on it I can't recall if this was by design or an oversight.  In my mind the same rules apply.  Nevertheless, I think the code patching fence sequences on x86 in general need a review in light of modernization of processors since this code was written, and the 32-bit question can be addressed then.  I'll open an issue to track that investigation.

Issue #1811, opened 2017-10-12
comptest takes 5mins to run on OS x


charliegracie:
I have been trying to improve the Travis build/test times and I noticed that OS X is still a lot longer than Linux x86.  One of the biggest differences is the time spent in comptest.  On Linux x86 it takes ~4s but on OS X it takes ~350s.  It is actually 2 of the subtests which take most of the time.

Linux x86:
1 test from CommoningDeathTest (105 ms total)
13 tests from ILValidatorDeathTest/IllformedTrees (1305 ms total)

OS X:
1 test from CommoningDeathTest (18580 ms total)
13 tests from ILValidatorDeathTest/IllformedTrees (324226 ms total)

charliegracie:
@Leonardo2718 this seems really strange to me.  When I was working on SOM++ though I noticed that the std::map data structure was orders of magnitudes of slower on OS X than it was on Linux.  The trivial use case in SOM++ was adding a 2X throughput loss for the entire benchmark run so I could believe something like that could be causing this issue. FYI @mstoodle and @0xdaryl

Leonardo2718:
Without profiling information, it will be difficult to pin-point what exactly is consuming so much time.

However, my initial thought is that it probably has more to do with the Death Test implementation in gtest rather than the tests themselves. Those death tests always run slower than any other test.

If this is correct, I don't think there's a lot that can be done, though it might be worth researching if someone else has also run into perf problems with gtest on OSX and found a way around it.

charliegracie:
I finally got my OS X machine up and running.  If I run comptest by itself it only takes a few seconds so the issue was not quickly reproducible on my machine.  I will run it under ctest and see if that makes any difference.  If it does not then I will check my system configuration to see what if any differences there are.

charliegracie:
I am using AppleClang 9.0.0.9000038 and the Travis builds are using AppleClang 7.3.0.7030031.

charliegracie:
I tried a build that used [XCode 9](https://travis-ci.org/charliegracie/omr/jobs/288795551).  The time to run the comptest was not improved.

This was the compiler version in that test: AppleClang 9.0.0.9000037

charliegracie:
I got a co-worker who has OS X 10.11 to try and the time was also fast on his machine.

My next step will be to see if I can use "Instruments" on the command line to gather performance data.

charliegracie:
[...]

PR #1820, opened 2017-10-18
Switch OS X to use Unix Makefiles

charliegracie:
Simplify the yml file and speed up the OS X builds slightly but not
using the Ninja generator.


charliegracie:
@youngar it took longer than it did before because there was no ccache for clang and unix makefiles.  After it gets merged OS X should be about 1 minute faster.  Still digging into why OS X is so slow.  Some investigation is in Issue #1811

Issue #1818, opened 2017-10-18
PROPOSAL: Provide record and replay capability for JitBuilder API


mstoodle:
The JitBuilder API is currently leveraged primarily during compilation (specifically during IL generation) to create the input to the JIT compiler. Right now, different language compilers (for Smalltalk, Lua, Rosie, Swift, etc.) exist and make calls to the JitBuilder API to direct compilation. But the only way to test those compilers is in their own specific environment. That makes it relatively difficult to reflect those compiler needs back into the Eclipse OMR project.

If we could record the JitBuilder API calls and replay them to an independent JitBuilder library, we could at least recreate the compilation effort from other language compilers directly in OMR and  bring tests into OMR that protect those other compilers as the JitBuilder API and implementation continue to evolve. In principle, it should be possible to execute that code as well, but without mock interfaces to the environments in which that code is expected to run, I do not expect it to be easy to validate the generated code using this technique.

Nonetheless, I think having the ability to store JitBuilder IL (JBIL) to a file would be very useful for reflecting our consumers use of JitBuilder at OMR. There are some other interesting facilities we could build on top of record-and-replay, but I think the ability to test should be valuable enough.

As I open other issues to move this work forward, I will add comments here to treat this as an umbrella item. I confess, I have already been moving down this path with an implementation, just to see how it might pan out, but I am still interested in any feedback including support or negative responses.

fyi @charliegracie @jduimovich @Leonardo2718

mstoodle:
This issue was mistakenly closed when the PR to remove the pre-existing replay support was merged.

Leonardo2718:
Pull requests #2915 and #2968 added a generator for the C++ JitBuilder client API with the intention of allowing language bindings to be generated from a high-level description of the JitBuilder API.

Pull requests #3019 and #3056 (unmerged at time of writing) add Record and Replay functionality to JitBuilder based on what is discussed at a high level in this issue.

In an attempt to minimize the amount of work required to maintain the Recorder mechanism when the JitBuilder API changes, @mstoodle and I have been discussing the possibility of generating parts of the Recorder mechanism in a similar way as the client API.

[...]

PR #1821, opened 2017-10-18
Eliminate early jitbuilder replay support

mstoodle:
The file ReplayMethod.cpp is a remnant of an earlier prototype for replay
compilation for JitBuilder, along with a few targets in release/Makefile.
There are also macros defined and called by IlBuilder and MethodBuilder
that are about to be replaced by a better implementation of JBIL recording/
replaying.

These can all be eliminated.

Issue: #1818


mstoodle:
genie-omr build all

compiler/ilgen/IlBuilder.cpp
```diff
@@ -165,6 +155,7 @@ IlBuilder::setupForBuildIL()
    }
 
 
+#if 0
```
0xdaryl: I'll let this go cuz I know this is a WIP.  :-)

PR #1822, opened 2017-10-19
Prevent suspending native stack walk threads for z/TPF.

jjohnst-us:
Prevent suspending native stack walk threads for z/TPF.
In addition, call special z/TPF thread initialization routines.

### DIFF
--- a/port/ztpf/omrintrospect.c
+++ b/port/ztpf/omrintrospect.c
@@ -1079,90 +1079,10 @@ omrintrospect_threads_startDo_with_signal(struct OMRPortLibrary *portLibrary, J9
  *	work on z/TPF. We'll leave this gate open for now, with a pending decision to close
  *	it or not.
  */
-#ifdef ZOS64
-	RECORD_ERROR(state, UNSUPPORT_PLATFORM, 0);
+
+	RECORD_ERROR(state, UNSUPPORTED_PLATFORM, 0);
 	return NULL;
-#endif
 
-	/* construct the walk state and thread structures */
-	state->heap = heap;
-	state->portLibrary = portLibrary;
-	data = (struct PlatformWalkData*)portLibrary->heap_allocate(portLibrary, heap, sizeof(struct PlatformWalkData));
-	state->platform_data = data;
-	state->current_thread = NULL;
-
-	if (!data) {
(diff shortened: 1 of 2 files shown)


jjohnst-us:
Low risk mode if it can make 4Q 2017 it would be nice.

jjohnst-us:
Had to re-push an update, discovered a compilation issue on z/TPF build only.

port/ztpf/omrintrospect.c
```diff
@@ -1079,12 +1079,12 @@ omrintrospect_threads_startDo_with_signal(struct OMRPortLibrary *portLibrary, J9
...
  */
-#ifdef ZOS64
-	RECORD_ERROR(state, UNSUPPORT_PLATFORM, 0);
+
+	RECORD_ERROR(state, UNSUPPORTED_PLATFORM, 0);
 	return NULL;
-#endif
 
-	/* construct the walk state and thread structures */
+	/* Maintaining original code here for potential future use.
```
charliegracie: I think I would prefer to not have a large commented out piece of code.  If we decided to use this in the future we could always go back to previous revisions in the repo to get the code.  @mstoodle what are your thoughts?

port/ztpf/omrintrospect.c
```diff
@@ -1163,6 +1163,7 @@ omrintrospect_threads_startDo_with_signal(struct OMRPortLibrary *portLibrary, J9
 cleanup:
 	resume_all_preempted(data);
 	return NULL;
+	*/
```
charliegracie: Is this function now missing a return?

port/ztpf/omrintrospect.c
jjohnst-us: Thanks for the feedback.  I removed the commented code.

port/ztpf/omrintrospect.c
jjohnst-us: The return was located at the top of the function.

jjohnst-us:
@charliegracie 
Let me know if you need anything else thanks!

jjohnst-us:
@charliegracie 
Once merged I assume I would need to apply this to 4Q 17 Branch as well if we wanted it to make cut-off?

Issue #1482, opened 2017-08-09
MIN_INT / MIN_INT evaluates to -1 when TreeSimplifier is disabled


mgaudet:
Reproduction Steps
---------------------

```
cd omr
mkdir build
cd build
cmake -GNinja -C../cmake/caches/Travis.cmake ..
ninja comptest
TR_Options=lastOptIndex=-1  ./fvtest/compilertriltest/comptest --gtest_filter=DivArithmeticTest/Int32Arithmetic.UsingConst/28
```

Error Output
-------------

```
Note: Google Test filter = DivArithmeticTest/Int32Arithmetic.UsingConst/28
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from DivArithmeticTest/Int32Arithmetic
[ RUN      ] DivArithmeticTest/Int32Arithmetic.UsingConst/28
../fvtest/compilertriltest/ArithmeticTest.cpp:38: Failure
      Expected: param.oracle(param.lhs, param.rhs)
      Which is: 1
To be equal to: entry_point()
      Which is: -1
[  FAILED  ] DivArithmeticTest/Int32Arithmetic.UsingConst/28, where GetParam() = ((-2147483648, -2147483648), ("idiv", 0x750a40)) (0 ms)
[----------] 1 test from DivArithmeticTest/Int32Arithmetic (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DivArithmeticTest/Int32Arithmetic.UsingConst/28, where GetParam() = ((-2147483648, -2147483648), ("idiv", 0x750a40))

 1 FAILED TEST

```

Preliminary Investigation: 
-------------------------

As expected, the tree 

```
n1n       BBStart <block_2>       
n3n       ireturn                 
n4n         idiv                  
n5n           iconst 0x80000000   
n6n           iconst 0x80000000   
n2n       BBEnd </block_2>        
``` 

Is returned from IlGen. However, this passes without `lastOptIndex` because the treeSimplifier sees this tree and does the correct substitution with `iconst 1`. 

Without tree simplification however, it appears there's a sign-extending widening happening at some point that interacts poorly with another simplification.

mgaudet:
[lastOptIndex1.txt](https://github.com/eclipse/omr/files/1211850/lastOptIndex1.txt)

[...]

Issue #566, opened 2016-12-05
Prima facie valid IL opcodes are invalid depending on platform


lmaisons:
Some opcodes in the compiler IL that appear to be universally valid for use may or may not be valid depending on the platform we're compiling on.

An example that came up recently is 'aiadd':
* Purports to add an 'integer' sized value to an address
* We expect binary operations to operate on fields of the same length
* In OMR compiler IL, 'integer' sized opcodes are always 32 bits

Thus, on 64 bit platforms, aiadd is invalid.

mstoodle:
in this particular case, I suppose we could introduce an aiuadd opcode to fill the gap in reasoning about whether to sign extend the integer operand. But that would be a bit of a weird scenario and I'm not sure I like it.

Better would be for the position you've advocated, which is that aiadd is an invalid opcode on 64-bit platforms (the opcode that should be used is aladd and there should be either an iu2l or i2l on the integer operand). That means we should have an assertion somewhere if anyone tries to (re)create such a node. Sadly, it also passes on some pain to people working with the IL :( .

Still, many things really need to know what that type is, anyway, so maybe the pain forces us to think about it properly (he suggested, squinting dubiously).

mgaudet:
Another option would be to replace `aiadd` and `aladd` with a 'typeless' `aadd`, that takes either a `Int32` or `Int64` argument depending on address width.  

All of this is evidence we also need a much better IL verification story.

Leonardo2718:
@mstoodle Actually, I believe there already exists a `aiuadd`.

I tend to prefer @mgaudet option as it's one less thing to think about when generating IL. Otherwise, one has to think about which opcode to generate *and* whether a `iu2l` or `i2l` is needed.

I also agree that IL verification definitely needs to be able to catch these kinds of errors.

mstoodle:
The problem with the "generic" aadd opcode is that you're basically punting the complexity into the optimizer that's reasoning about what that thing actually means based on the platform you're running on.

I'm not sure that's the right trade-off compared to: figure out what you really mean and provide the proper input to the compiler (once).

We have an IL verification story and it has worked for a bunch of common problems. I agree that it's not perfect and could be made better. IMO improvements are very welcome so long as the verification doesn't impact production compile time (unless you can make an argument that the cost is worth it to find mistakes that are made *often*).

andrewcraik:
[...]

PR #1826, opened 2017-10-23
Improve the IL Validation process

rbanerjee:
The primary aim of this PR is to Improve the IL Validation process and lay the necessary groundwork for future improvements.

In light of the discussions here (https://github.com/eclipse/omr/issues/1482), we clearly need a better way to reason about our IL. This involves enforcing existing IL Validation Rules and formalizing some of the more implicit expectations.

The Goals are listed below (loosely arranged in order of priority) :

1. Give more control over "when" and "what" we are validating against.
   - Connect the ILValidator to the existing compiler technology. The ILValidator is initialized right after IL generation. And can then be used to Validate the IL at any point after that.
    - This allows a more "state aware" form of IL Validation. For example, we can choose to `validate` the IL using a certain set of Rules right after IL Generation which in turn can then be different from the set of Rules we choose to use immediately before going into Code Generation.
The state is passed to the `validateIL` method (as an enum value) and depending on that `state`, we decide which subset of all the available Rules we should check against (the set of `ALL` available rules is specified during the creation of the ILValidator object).
Some of these Rules could also be architecture aware in which case it gives individual Code Generators finer control over the IL that it’s expected to handle (#1901). Meaning potential “discrepancies” (`ABI` specific IL quirks) can be caught potentially before Codegen. 

2. Compartmentalize Rules.
There are a bunch of dimensions along which we could have categorized the IL Validation Rules. The one chosen here involves classifying rules based on the “context” required to check whether they are valid or not.

   The types flow somewhat naturally form there:
   - Rules that require one to check the entire method (TR::ResolvedMethodSymbol) for a certain property (i.e. Soundness checks)
   - Rules that only confine themselves to looking at a particular Block(TR::Block).
   - Rules that only check whether a particular node (TR::Node) conforms to a certain expectation. For example checking whether `ireturn` has child types that are either int | short | byte.

3. Decoupling existing Rules.
    A large part of which involves decoupling the `Soundness Checks` from the rest of the rules. Also split up complicated rules into more manageable (somewhat atomic) chunks.
    Making it easier to add new rules. Also provide a set of utility functions in order to write said Rules.
    
Related Issue: #1482, #1901, #566, #1971

Signed-off-by: Rwitaban (Ray) Banerjee <rayb@ca.ibm.com>


compiler/optimizer/OMROptimizationManager.cpp
```diff
@@ -286,9 +286,9 @@ void OMR::OptimizationManager::performChecks()
          }
       else
          {
-         TR::ILValidator validator(self()->comp());
-         auto methodSymbol = self()->comp()->getMethodSymbol();
-         validator.treesAreValid(methodSymbol->getFirstTreeTop(), methodSymbol->getLastTreeTop());
+           TR::ILValidatorDefaultDriver validator(self()->comp());
```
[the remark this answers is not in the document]
rbanerjee: Must have been a typo. But more importantly the indentation needs to be consistent throughout the code in general. I'll move everything to `Whitesmiths style` in the next iteration.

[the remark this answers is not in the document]
rbanerjee: Actually this wasn't up for review yet :)
Agree with a lot of the suggestions though. Will ping you with a `ptal` when the next iteration is ready. Appreciate the help.

rbanerjee:
Complete revision of the previous approach. Updated the Title and Description accordingly.

rbanerjee:
@mgaudet ptal. @Leonardo2718  I would appreciate a review from you as well, specially regarding 
> This is part of a long term goal to merge the IlVerifier and
the ILValidator into one entity.

Will add more reviewers as the PR progresses.

compiler/ras/BlockValidationRules.hpp
```diff
@@ -0,0 +1,79 @@
...
+   };
+
+
+// TODO: Even though the Rule(class) names are somewhat "self explanatory",
+//       we might still want to formally define each of these rules.
+//       And if we want these definitions to live with the code, then
+//       this seems like a good place to do that.
+//       (Another option would be to create a doc on github for these, but I can
+//        justify providing a brief description here as well.)
+class ValidateNodeRefCountWithinBlock : public TR::BlockValidationRule
```
mgaudet: I think this comment should try to reflect the property that's being enforced.

compiler/ras/BlockValidationRules.cpp
```diff
@@ -0,0 +1,97 @@
...
+   {
+   _nodeChecklist.empty();
+   for (TR::TreeTop *tt = firstTreeTop; tt != exitTreeTop->getNextTreeTop(); tt = tt->getNextTreeTop())
+      {
+      TR::Node *node = tt->getNode();
+      node->setLocalIndex(node->getReferenceCount());
+      validateRefCountPass1(node);
+      }
+
+   _nodeChecklist.empty();
```
rbanerjee: Oooops! I don't need to empty the nodeChecklist twice. Will remove in the next iteration.

compiler/ras/ILValidator.hpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2016, 2017 IBM Corp. and others
```
rbanerjee: Typo (a weird one at that) will fix in the next patch set.

compiler/ras/ILValidator.cpp
```diff
@@ -21,333 +21,132 @@
...
-            }
-         traceMsg(comp(), "}\n");
-         }
-      }
-
+   return _comp;
    }
 
-bool TR::ILValidator::treesAreValid(TR::TreeTop *start, TR::TreeTop *stop)
+bool TR::ILValidator::validate()
```
mgaudet: Given that the return value here is never checked, and the complexity that is injected below trying to figure out what to return, perhaps the return type should just be removed, the returns removed.

compiler/ras/ILValidator.hpp
```diff
@@ -19,187 +19,51 @@
...
+   // The Validation rules are categorized into three types based on the
+   // "scope" required to validate them.
+   // For example, a MethodValidationRule would require information about
+   // the entire Method(encapsulated as a TR::ResolvedMethodSymbol) in order
+   // to find out if the IL satisifies the given condition.
+   // Whereas a NodeValidationRule checks whether a particular TR::Node has
+   // some property. Meaning a NodeValidationRule doesn't need to keep track
+   // of already encountered nodes or peek into other blocks to see whether
+   // a particular Node is valid or not.
+   std::vector<TR::MethodValidationRule *> _methodValidationRules;
```
mgaudet: I'd change these comments to split them to a per-member basis, and you should use Doxygen style throughout.

compiler/ras/MethodValidationRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "infra/Checklist.hpp"                       // for NodeChecklist
+#include "infra/ILWalk.hpp"                          // for PostorderNodeOccurrenceIterator
+
+// Exit protocol upon encountering a failure.
+#if defined(DEBUG) || defined(PROD_WITH_ASSUMES)
+#define ABORT() TR::trap()
+#else
+#define ABORT() comp->failCompilation<TR::CompilationException>("Validation error: IL is unsound")
+#endif
+#define FAIL() if (!feGetEnv("TR_continueAfterValidationError")) ABORT()
```
mgaudet: The (almost exact) duplication of `ABORT` / `FAIL` is a bit curious;

compiler/ras/MethodValidationRules.hpp
```diff
@@ -0,0 +1,107 @@
...
+   //           Note that TR::checkCondition is guranteed to FAIL() under a certain protocol
+   //           (which is defined in ILValidationUtils.cpp) if the `condition` is not fulfilled.
+   //           So as long as this function returns, the node upholds the given rule.
+   //           Therefore we could very well make it a boolean or even a void function.
+   //           BUT there is a case to be made for future MethodValidationRules that don't use
+   //           the utilities defined in ILValidationUtils. In which case they might
+   //           choose to use error codes for specific scenerios (personally I'm not a big
+   //           fan of doing things that way, specially since there are much better alternatives in
+   //           this case) [same applies for Node and Block Validation Rules].
+   virtual int32_t validate(TR::ResolvedMethodSymbol *methodSymbol) = 0;
```
mgaudet: IMO, the above comment makes the argument in favour of ditching void -- I don't think it's worth the awkwardness for functionality not used.

compiler/ras/NodeValidationRules.cpp
```diff
@@ -0,0 +1,137 @@
...
+   auto opcode = node->getOpCode();
+   if (opcode.getOpCodeValue() == TR::ireturn)
+      {
+      const auto childCount = node->getNumChildren();
+      for (auto i = 0; i < childCount; ++i)
+         {
+         auto childOpcode = node->getChild(i)->getOpCode();
+         const auto actChildType = childOpcode.getDataType().getDataType();
+         const auto childTypeName = TR::DataType::getName(actChildType);
+         TR::checkCondition(node, (actChildType == TR::Int32 ||
```
mgaudet: Maybe this rule should have a link to #1901?

compiler/ras/NodeValidationRules.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+   public:
+
+   Validate_ireturnReturnType(TR::Compilation *comp);
+
+   int32_t validate(TR::Node *node);
+   };
+
+// TODO: Add NodeValidationRules regarding platform specific opcodes.
+//       (See: Issue #566 about one on `aiadd`)
```
mgaudet: Todo is done .

compiler/ras/ILValidator.cpp
```diff
@@ -21,333 +21,132 @@
...
+     // of the available ones we should call.
+     // This also lets one specify the ordering in which the rules are validated.
+     TR::MethodValidationRule* temp_method_rules[] = { new  TR::SoundnessRule(_comp),
+                                                       new  TR::ValidateLivenessBoundaries(_comp) };
+
+     TR::BlockValidationRule* temp_block_rules[] = { new  TR::ValidateNodeRefCountWithinBlock(_comp) };
+
+     TR::NodeValidationRule* temp_node_rules[] = { new  TR::ValidateChildCount(_comp),
+                                                   new  TR::ValidateChildTypes(_comp),
+                                                   new  TR::Validate_ireturnReturnType(_comp) };
```
mgaudet: I don't think we can have use of `new` without making it placement-`new` on to a TR memory type: See #2003.

compiler/compile/OMRCompilation.cpp
```diff
@@ -1987,6 +2003,17 @@ void OMR::Compilation::switchCodeCache(TR::CodeCache *newCodeCache)
       }
    }
 
+// TODO: Add Enum overrides for this.
+void OMR::Compilation::validateIL()
+   {
+   if (_ilValidator == NULL)
+      self()->failCompilation<TR::CompilationException>("Attempting to validate the IL without the ILValidator being initialized");
+   else
```
mgaudet: If this is an unexpected state, prefer `TR_ASSERT_FATAL` over failing a compilation, which could be a silent performance issue.

fvtest/compilertriltest/ILValidatorTest.cpp
```diff
@@ -41,10 +43,24 @@ TEST_P(IllformedTrees, FailCompilation) {
...
+     //       So the Trees below will always fail compilation on non-32 bit platforms
+     //       and it only makes sense to test with them when running on a 32 bit platform.
+     //       Overall we need a better strategy for writing platform specific Tril Tests.
+     //       See: Issue #1843
+     // aiadd | auiadd childProperties = TWO_CHILD(TR::Address, TR::Int32)
+#if defined(TR_TARGET_32BIT)
+    "(method return=Address (block (areturn (aiadd (aconst 4) (lconst 1)))))",
+    "(method return=Address (block (areturn (aiadd (iconst 1) (aconst 4)))))",
+#endif
+    // Illformed because aiadd is invalid on non-32 bit platforms.
```
mgaudet: Could just use `#else`

mgaudet:
From a high level, one other comment I'd have: I am not clear that all the new files are necessary right now -- overall we don't have that many validators or validation categories. IMO, collapsing most of this into a single header might be an acceptable choice. 

The addition of new files may be unavoidable, but is going to produce a downstream dependency in OpenJ9 at the very least.

rbanerjee:
@mgaudet  ptal :)

rbanerjee:
>I think I'd like to see a lot of the specified cleanup in TODOs happen.

@mgaudet  most of the `TODO`s have now been addressed. For the remaining ones (this applies for the newly introduced ones as well), I would like to address them as separate PRs. Some of these even require a bit more discussion on the topic and I would be happy to open up Issues for those once this gets merged.

>From a high level, one other comment I'd have: I am not clear that all the new files are necessary right now -- overall we don't have that many validators or validation categories. IMO, collapsing most of this into a single header might be an acceptable choice.

The `{Method, Block, Node}ValidationRule.cpp/hpp` files have now been collapsed under one set of files titled `ILValidationRules.cpp/hpp`(I have still tried partitioning these files in way such that it's easy to find and add rules).

As for the remaining newly introduced files, I think they are somewhat necessary as they serve very distinct purposes. Maybe their existence is a little more justifiable now than before (since this iteration is quite different from the last patch-set). Also, I would be happy to open up a paired dependency change for (at least) OpenJ9 once this gets approved.

rbanerjee:
I would like a review from @0xdaryl as well.

rbanerjee:
Also, the last commit (8628c89) seems to be causing one of the JitTests to fail on certain platforms. https://github.com/eclipse/omr/blob/8b089c3af5fe41dc700aacf5ade826761687c2cf/jitbuilder/release/src/OperandStackTests.cpp#L416-L447  Note that ILValidation isn't turned on by default so my guess (just by skimming over the code) would be that it is actually trying to test whether the generated IL is Valid, and it turns out that it's not under the new definition introduced here.

The particular rule that it fails to satisfy is `aiadd is only valid on 32 bit platforms`.
(Another possible scenario could be that my interpretation of the requirement is wrong hence the Rule itself doesn't make sense.)

Now as pointed out in one of the `TODO`s (https://github.com/eclipse/omr/pull/1826/files#diff-87c67c5a8fbce035687150196c38c015R59) , currently there is only one single compiler option that controls whether we treat ILValidation errors as Fatal or not (it was added in this PR). This is currently the default functionality; as in if  ILValidation is turned on, then all `Rules` are by implicitly "Strict" (The Tril ILValidation Death Tests work in the same manner).

As argued in the aforementioned `TODO`, this decision (on "Strictness) should be made at the `ILValidation Strategy` Level and the Compiler Option (`continueAfterILValidationError`) should only act as a way of overriding the said decision.

This way we could make the `Rule` about `aiadd only being valid on 32 bit platforms` a non-strict one. (Until it is resolved. Or maybe we always want it to be a "Lenient" one in which case it just acts a "Warning" instead of a "Rule". But I think that's  a separate discussion.)

Now as for resolving this issue, I could add the functionality mentioned in the `TODO` as part of this PR. The only thing stopping me is the fact that, I fear this PR has gotten complicated enough already and I don't want to it to reach a point where it tries to "boil the ocean".

So essentially, is this seen as a *required* functionality that the ILValidator must have? If so then I can add it to this PR. Otherwise I would prefer not to.

In which case the alternative would be to take out that specific Rule (which would be the last commit) and make it part of a different PR (potentially one that adds the functionality of being able to make a Rule "Strict" or "Lenient" on an `ILValidationStrategy` level).

mgaudet:
All LGTM, though, I'm a little curious about why 33d5c7a isn't squashed... somewhere. 

IMO, this is ready to go if you strip the TODOs and replace them with issues, but I will defer to a committer on the actual reqs for merge. I like this conceptually, and definitely support the strict, lenient pairing.

rbanerjee:
>  I'm a little curious about why 33d5c7a isn't squashed... somewhere.

@mgaudet I assumed it would make it easier to fix downstream dependencies (the ones caused by the addition of new files) :) 
I will squash it in the next one.

rbanerjee:
In relation to my comment above regarding the `jitbuilder` test failure, I think it's best I pull out the associated commit and address this on a separate PR. Having the ability to specify Rules to be either `Strict` or `Lenient` is very useful but IMO not *necessary* enough that it needs to be addressed in this PR.

Will address this is in the next iteration.

compiler/ras/ILValidationUtils.cpp
```diff
@@ -0,0 +1,150 @@
...
+      /**
+       *Safely bring down the VM if the continueAfterILValidationError Compiler Option
+       *is not set.
+       */
+      if (!comp->getOption(TR_continueAfterILValidationError))
+         TR::trap();
+      }
+   }
+
+void TR::updateNodeState(TR::Node *node,
```
Leonardo2718: I don't think it makes sense to use this function outside of `ValidateLivenessBoundaries`. Could you make it a helper function of that class instead?

rbanerjee:
> I have to admit I'm not a fan of ILValidationStrategy, but I can live with it for now.

Can I ask you to elaborate a little there? :)

Leonardo2718:
> Can I ask you to elaborate a little there?

Writing up a complete explanation is going to take me a bit of time. As I said, I'm okay with it for the time being. I don't want to hold back this PR. I'll open a follow up issue to document my concern later on.

compiler/ras/ILValidationUtils.cpp
rbanerjee: Good Call! Done.

rbanerjee:
@mgaudet, @Leonardo2718 ptal, this is ready for review again.

mgaudet:
Looks good to me. 

One small comment being: I am not sure describing `TR::trap` as safely taking down the VM is accurate: more like aggressively taking it down.

rbanerjee:
@mgaudet  ptal

> One small comment being: I am not sure describing TR::trap as safely taking down the VM is accurate: more like aggressively taking it down.

@mgaudet Agreed. Upon closer inspection that comment seems somewhat superfluous. Removed in the latest patch-set.

rbanerjee:
@mgaudet @Leonardo2718 I think I have addressed all of the requests. Please do let me know if I missed anything.

@0xdaryl over to you. (Testing is done, removing the`WIP` tag.)

compiler/ras/ILValidator.cpp
```diff
@@ -19,335 +19,243 @@
...
+      */
+
+     _methodValidationRules.assign(begin(temp_method_rules), end(temp_method_rules));
+     _blockValidationRules.assign(begin(temp_block_rules), end(temp_block_rules));
+     _nodeValidationRules.assign(begin(temp_node_rules), end(temp_node_rules));
    }
 
-void TR::ILValidator::validityRule(Location &location, bool condition, const char *formatStr, ...)
+
+TR::Compilation *TR::ILValidator::comp()
```
0xdaryl: Shouldn't this be inline in the header file?

compiler/ras/ILValidator.cpp
```diff
@@ -19,335 +19,243 @@
...
+   std::vector<TR::MethodValidationRule *> reqMethodValidationRules;
+   while (strategy->id != OMR::endRules)
       {
-      validityRule(newLocation, state._futureReferenceCount >= 1, "Node already has reference count 0");
-      if (--state._futureReferenceCount == 0)
+      for (auto it = _methodValidationRules.begin(); it != _methodValidationRules.end(); ++it)
          {
-         _liveNodes.remove(node);
+         /**
+          *Each *ValidationRule has a unique `id`. These ids are defined in
```
0xdaryl: Spaces after * please.

compiler/control/OMROptions.cpp
```diff
@@ -183,6 +183,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
                                TR::Options::setStaticNumeric, (intptrj_t)&OMR::Options::_compThreadCPUEntitlement, 0, "F%d", NOT_IN_SUBSET },
    {"concurrentLPQ", "M\tCompilations from low priority queue can go in parallel with compilations from main queue", SET_OPTION_BIT(TR_ConcurrentLPQ), "F", NOT_IN_SUBSET },
    {"conservativeCompilation","O\tmore conservative decisions regarding compilations", SET_OPTION_BIT(TR_ConservativeCompilation), "F"},
+   {"continueAfterILValidationError", "O\tAll other things held constant, do not abort upon encountering an ILValidation failure.", SET_OPTION_BIT(TR_ContinueAfterILValidationError), "F"},
```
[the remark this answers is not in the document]
rbanerjee: Done.

> Is it needed? 

No, not yet at least. So good catch!

But I do believe once we introduce the notion of `Strict` vs `Lenient` Rules, we will need to have something like this. Since if a `Rule` is set to `Lenient`, we do indeed _continue after an IL Validation Error_ even if the "continueAfterILValidationError" option is not set. I will have to give it some more thought, once that functionality is added. 

The original intent was for this option to be able to able to override this behaviour if required. As in, even if a Rule was set as a `Strict` one (`Rules` for which we abort compilation rather than just emit an diagnostic message), this option would allow us to override that behaviour and continue compilation. However I have decided to add the `Strict` | `Lenient` split as a followup `PR`.
So although compilation is expected to abort under conditions, can fail or the relation can be abolished by intervening factors.

The "All other things held constant" part was essentially alluding to that. Thanks for catching that. I will find a better way to phrase it next time.

0xdaryl:
genie-omr build all

0xdaryl:
@mgaudet , please approve if you're OK with the current changes.

compiler/ras/ILValidator.cpp
rbanerjee: Done.

compiler/ras/ILValidator.cpp
rbanerjee: Done.

compiler/ras/ILValidationUtils.cpp
```diff
@@ -0,0 +1,92 @@
...
+      va_start(args, formatStr);
+      vprintDiagnostic(comp, formatStr, args);
+      va_end(args);
+      printDiagnostic(comp, "\n");
+      if (!comp->getOption(TR_ContinueAfterILValidationError))
+         TR::trap();
+      }
+   }
+
+void TR::printDiagnostic(TR::Compilation *comp, const char *formatStr, ...)
```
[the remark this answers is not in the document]
rbanerjee: See my comment above.

compiler/ras/ILValidationUtils.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TR::LiveNodeWindow::LiveNodeWindow(NodeSideTable<NodeState> &sideTable,
+                                   TR_Memory *memory)
+   :_sideTable(sideTable)
+   ,_basis(0)
+   ,_liveOffsets(10, memory, stackAlloc, growable)
+   {
+   }
+
+bool TR::isLoggingEnabled(TR::Compilation *comp)
```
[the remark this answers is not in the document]
rbanerjee: `isLoggingEnabled` has been changed to `isILValidationLoggingEnabled` (a little long but I guess it works ...)

For similar reasons, the following have now been renamed in this change list. 

`checkCondition` -> `checkILCondition`
`printDiagnostic` -> `printILDiagnostic`
`vprintDiagnostic` -> `vprintILDiagnostic`

Thanks for pointing that out!

compiler/ras/ILValidationStrategies.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+   /* Used to mark the end of a ILValidationStrategy */
+   endRules
+   };
+
+struct ILValidationStrategy
+   {
+   OMR::ILValidationRule        id;
+   /**
+    * TODO: Eventually ILValidationStrategy should be implemented
```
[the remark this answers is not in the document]
rbanerjee: Removed in the latest change list.

Since the said `TODO` does not pre-exist this PR, I will wait for this to go through before opening up an Issue. (If you're alright with that).

compiler/ras/ILValidationRules.hpp
```diff
@@ -0,0 +1,229 @@
...
+    */
+   virtual void validate(TR::ResolvedMethodSymbol *methodSymbol) = 0;
+
+   OMR::ILValidationRule id()   { return _id; }
+   };
+
+
+class SoundnessRule : public MethodValidationRule
+   {
+   TR::Compilation  *_comp;
```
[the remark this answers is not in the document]
rbanerjee: Done.

compiler/ras/ILValidationRules.hpp
```diff
@@ -0,0 +1,229 @@
...
+ *          Validation Strategy.
+ *     
+ *       3. Finally, the ILValidator is responsible for validating
+ *          the IL based on a certain ILValidationStrategy.
+ *          So please instantiate the said `*ValidationRule` object
+ *          during the creation of TR::ILValidator.
+ *
+ */
+
+#ifndef ILVALIDATIONRULES_HPP
```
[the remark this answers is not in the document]
rbanerjee: Done.

compiler/ras/ILValidationRules.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+   if (opcode.expectedChildCount() != ILChildProp::UnspecifiedChildCount)
+      {
+      const auto expChildCount = opcode.expectedChildCount();
+      const auto actChildCount = node->getNumChildren();
+
+      /* Validate child count. */
+      if (!opcode.canHaveGlRegDeps())
+         {
+         // In the common case, no GlRegDeps child is expect nor present.
```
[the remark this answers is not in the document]
rbanerjee: Done.

compiler/ras/ILValidationRules.cpp
```diff
@@ -0,0 +1,499 @@
...
+      }
+   else
+      {
+      TR::checkCondition(node, node->getOpCode().isTreeTop(), comp,
+                   "Node has already gone dead");
+      }
+
+   if (TR::isLoggingEnabled(comp))
+      {
+      // TODO: This should be made into a compiler option.
```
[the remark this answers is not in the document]
rbanerjee: The said `TODO` is no longer needed in the context of this updated change list.

compiler/ras/ILValidationRules.cpp
```diff
@@ -0,0 +1,499 @@
...
+    */
+   if (liveNodes.isEmpty())
+      {
+      liveNodes.startNewWindow();
+      }
+   }
+
+void TR::ValidateLivenessBoundaries::updateNodeState(TR::Node *node,
+                                                     TR::NodeSideTable<TR::NodeState>  &nodeStates,
+                                                     TR::LiveNodeWindow &liveNodes, TR::Compilation *comp)
```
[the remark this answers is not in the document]
rbanerjee: Good catch! Done now.

compiler/ras/ILValidationRules.cpp
```diff
@@ -0,0 +1,499 @@
...
+   {
+   for (LiveNodeWindow::Iterator lnwi(liveNodes); lnwi.currentNode(); ++lnwi)
+      {
+      TR::checkCondition(node, false, _comp,
+		     "Node cannot live across block boundary at n%dn",
+		     lnwi.currentNode()->getGlobalIndex());
+      }
+
+   /**
+    *At the end of an extended block, no node we've already seen could ever be seen again.
```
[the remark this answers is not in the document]
rbanerjee: Done.

compiler/ras/ILValidationRules.cpp
```diff
@@ -0,0 +1,499 @@
...
+                               _comp, "Expected BBStart after BBEnd");
+            isEndOfExtendedBlock = ! nextTree->getNode()->getBlock()->isExtensionOfPreviousBlock();
+            }
+         else
+            {
+            isEndOfExtendedBlock = true;
+            }
+
+         if (isEndOfExtendedBlock)
+            /* Ensure there are no nodes live across the end of a block */
```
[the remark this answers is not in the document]
rbanerjee: Done.

compiler/control/OMROptions.cpp
```diff
@@ -1163,6 +1164,8 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"traceILDeadCode=",                 "L{regex}\tlist of additional traces to enable: basic, listing, details, live, progress",
         TR::Options::setBitsFromStringSet, offsetof(OMR::Options, _traceILDeadCode), 0, "P"},
    {"traceILGen",                       "L\ttrace IL generator",                           SET_OPTION_BIT(TR_TraceILGen), "F"},
+   {"traceILValidation",                "L\tsynonym for traceILValidator",                          SET_OPTION_BIT(TR_TraceILValidation), "F" },
```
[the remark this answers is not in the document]
rbanerjee: Removed the synonym in this change list.

compiler/ilgen/IlInjector.cpp
```diff
@@ -36,7 +36,8 @@
 #include "ilgen/IlInjector.hpp"
 #include "ilgen/TypeDictionary.hpp"
 #include "infra/Cfg.hpp"
-#include "ras/ILValidator.hpp"
+#include "ras/ILValidationStrategies.hpp"      // for TR::postILgenValidation
```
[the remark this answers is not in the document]
rbanerjee: Not necessary, but a lot of the preexisting files follow this convention. This particular file being an exception I suppose.

I agree on the fact that they are not that useful. Removed in the updated change list.

compiler/control/OMROptions.cpp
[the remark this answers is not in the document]
rbanerjee: Done.

0xdaryl:
Should this still be a WIP?

PR #1833, opened 2017-10-23
arraytranslate StringCoding.implEncodeISOArray

andrewcraik:
Handling of arraytranslate for ISO-8859-1 encoding from UTF-16 is currently in the
ValuePropagationCommon file. This change adds another recognized method
to the list of methods supported by this code.


DanHeidinga:
This looks like code that belongs in OpenJ9 as it appears to handle special cases for Java methods.  Can this be migrated to the OpenJ9 project?  At a minimum, I'd like to see an issue created to cover moving the code.

andrewcraik:
The J9_PROJECT_SPECIFIC macros are marking the code that needs to move in here. Unentangling all this stuff is a major piece of work taking a lot longer than the change I'm trying to do. We could open an item for this one, but it is a case of J9_PROJECT_SPECIFIC code needs to move down which is being done as people get time. Moving this code down properly is going to need more than a few hours of thought and testing. Given that if you want an issue to move J9_PROJECT_SPECIFIC code down I'll open it - just the places where it is used are non-trivial and so need time and care to change.

DanHeidinga:
I'd vote for opening issues for areas that have been recently touched as it makes the required refactoring explicitly visible in the issues list rather than only implicitly visible in the code.

andrewcraik:
Created issue #1852 for the guarded code.

charliegracie:
no need for genie testing as this code appears to all be under the J9_PROJECT_SPECIFIC define

vijaysun-omr:
Looks okay to me. I would like a review from @fjeremic too since there is a string compression related change in the commit.

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -3730,6 +3734,10 @@ void OMR::ValuePropagation::transformConverterCall(TR::TreeTop *callTree)
 
    TR::Node *child1 = TR::Node::create(TR::iadd, 2, srcOff, len);
    TR::Node *child2 = TR::Node::create(TR::arraylength, 1, srcObjNode);
+   // the source for the decompressed string version of this in StirngCoding takes a byte array
+   // used as a char array so convert the size appropriately for the check >> 1
+   if (rm == TR::java_lang_StringCoding_implEncodeISOArray)
+      child2 = TR::Node::create(TR::ishr, 2, child2, TR::Node::iconst(child2, 0));
```
fjeremic: Is this supposed to be a shift right by 1? Is the `TR::Node::iconst(child2, 0)` here correct? Also StringCoding is spelled incorrectly.

andrewcraik: fixed

vijaysun-omr:
Thanks for the buddy check @fjeremic and catching that typo. Merging this now.

PR #1835, opened 2017-10-23
Manage additional blocks in OSR infrastructure

ncough:
If OSR exception edge removal is run sufficiently
late, blocks may have been added between OSR
catch and code blocks. In this situation, the
assumptions made by DeadStoreCleanup may be
invalid, so it should stop.

Moreover, OSR exception edge removal should not
run if infrastructure has already been removed.


andrewcraik:
LGTM

vijaysun-omr:
Looks pretty safe to me. Merging.

PR #1836, opened 2017-10-23
Introducing Recognized Call Transformer

0dvictor:
Recognized Call Transformer is a new optimization pass which
provides a unified place for downstream projects to convert
certain calls into ILOps and/or intrinsics.


Leonardo2718:
Any chance we could get some documentation for how a downstream project might make use of this optimization?

0dvictor:
Good suggestion, I'll make an example using Eclipse OpenJ9.

0dvictor:
https://github.com/eclipse/openj9/pull/427 is an example to adopt this proposed Recognized Call Transformer.

andrewcraik:
I think OMR is going to need a stub do nothing pass which can be extened using normal C++ inheritance by languages to add their own custom support. At the moment this is just forward declarations of an opt that doesn't exist in OMR.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,61 @@
...
+   {
+   return new (manager->allocator()) TR::RecognizedCallTransformer(manager);
+   }
+
+int32_t OMR::RecognizedCallTransformer::perform()
+   {
+   auto methodSymbol = comp()->getMethodSymbol();
+   for (_treetop = methodSymbol->getFirstTreeTop(); _treetop != NULL; _treetop = _treetop->getNextTreeTop())
+      {
+      _node = treetop()->getNode()->getChild(0); // Get the first child of the tree
```
andrewcraik: Should use getFirstChild here

compiler/optimizer/OMRRecognizedCallTransformer.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+   protected:
+   inline TR::TreeTop* treetop() const { return _treetop; }
+   inline TR::Node*    node()    const { return _node   ; }
+
+   virtual bool isInlineable() { return false; }
+   virtual void transform() {}
+
+   private:
+   TR::TreeTop* _treetop;
```
andrewcraik: I'm not a fan of these fields - this is shared mutable state that is important for the optimization - if transform needs the treetop pass it the treetop and let it extract the node. If it needs a specific child node as well as the treetop pass both. This is more in keeping with the style of the other optimizations and results in a more functional style to the helper functions making them easier to refactor and move later.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "optimizer/RecognizedCallTransformer.hpp"
+
+#include "il/Node.hpp"
+#include "il/Node_inlines.hpp"
+#include "il/TreeTop.hpp"
+#include "il/TreeTop_inlines.hpp"
+#include "codegen/CodeGenerator.hpp"                      // for CodeGenerator
+#include "il/ILOpCodes.hpp"
+#include "il/ILOps.hpp"                                   // for ILOpCode, etc
+#include "ilgen/IlGenRequest.hpp"
```
andrewcraik: Why are these needed - it is unusual for an optimization to have dependencies on the ilgenerator...

compiler/optimizer/OMRRecognizedCallTransformer.cpp
0dvictor: Updated.

compiler/optimizer/OMRRecognizedCallTransformer.hpp
0dvictor: Sure, I removed these fields from the class and now they are parameters of the helper functions.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
0dvictor: Updated.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,55 @@
...
+TR::Optimization* OMR::RecognizedCallTransformer::create(TR::OptimizationManager *manager)
+   {
+   return new (manager->allocator()) TR::RecognizedCallTransformer(manager);
+   }
+
+int32_t OMR::RecognizedCallTransformer::perform()
+   {
+   for (auto treetop = comp()->getMethodSymbol()->getFirstTreeTop(); treetop != NULL; treetop = treetop->getNextTreeTop())
+      {
+      auto node = treetop->getNode()->getFirstChild();
```
andrewcraik: You should check that the result of getNode() has a first child before calling this.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,58 @@
...
+TR::Optimization* OMR::RecognizedCallTransformer::create(TR::OptimizationManager *manager)
+   {
+   return new (manager->allocator()) TR::RecognizedCallTransformer(manager);
+   }
+
+int32_t OMR::RecognizedCallTransformer::perform()
+   {
+   for (auto treetop = comp()->getMethodSymbol()->getFirstTreeTop(); treetop != NULL; treetop = treetop->getNextTreeTop())
+      {
+      if (treetop->getNode()->getNumChildren() > 0)
```
vijaysun-omr: Should this somehow be checking for and only doing the transformation at the first occurrence of a commoned call node ? I am thinking of a case when more than one tree (regardless of the opcode at the top node, e.g. treetop or store) can have a particular call node as the first child and the first reference to it is the one where the call node will actually get evaluated. While it's possible that the project specific "transform" handles this somehow, should this be considered in the "tree walking" pass that is being considered at this (OMR) layer ?

0dvictor: My apology for the late reply. This OMR layer code will call downstream project's "transform" for each occurrence of `call` node it sees; however, the project's "transform" may choose to transform or not. For example, should OpenJ9 uses it to transform `Unsafe`, OpenJ9's "transform" would probably only transform the `call` nodes marked in its fast-path. OMR does not have to and should not differentiate different occurrences of `call` node.

vijaysun-omr: Okay that sounds fine in theory but in practice I feel most downstream projects will end up duplicating the logic to avoid visiting the same call node multiple times. The call really "happens" at the first occurrence of the call node and so I'd actually argue the latter occurrences are not really calling anything and so no transformation should be trying to transform the later occurrences. Nor should it be necessary if the transformation of the call node happened correctly at the first occurrence.

0dvictor: I see your point; however, by not limiting the transformation to the first occurrence, downstream projects can be more flexible. They can choose to perform different transformations in different occurrences. I feel it is improper to limit this flexibility. In addition, downstream projects can implement their own logic to avoid duplicating logic, such as checking visit count before transformation.

andrewcraik: I have to say that on principle I would agree with @vijaysun-omr in that uncommoning a node (ie transforming a non-first occurrence of a node) is a non-trivial operation and doesn't seem like it belongs with a utility pass like the RecognizedCallTransformer. If it is not ok to transform the first occurrence of a call, then some additional complex analysis will be needed to do the necessary uncommoning and then transform a subset. Unless there is a compelling use case in a downstream project for the additional functionality I would be more comfortable if the nodechecklist existed in OMR and the option only existed to transform the first instance of a call in place which would change the call evaluation which is the true intent of this utility.

0dvictor: Thank you @vijaysun-omr and @andrewcraik for pointing out your concerns and ideas. I've updated the code so that now it only visits each call node once.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+TR::Optimization* OMR::RecognizedCallTransformer::create(TR::OptimizationManager *manager)
+   {
+   return new (manager->allocator()) TR::RecognizedCallTransformer(manager);
+   }
+
+int32_t OMR::RecognizedCallTransformer::perform()
+   {
+   TR_BitVector visited;
+   visited.init(0, comp()->trMemory(), heapAlloc, growable);
```
andrewcraik: This BitVector should not use heapAlloc - it is sufficient to use stack alloc so the memory is freed at the end of the optimization instead of at the end of the compilation. The correct declaration would be something like:
`TR_BitVector visited(comp()->trMemory()->currentStackRegion());`
There is a separate facility that already exists in the compilation to give you a special kind of BitVector that is recycled by the compilation for the purposes of doing this kind of tracking. The advantage of this special type is that the compilation manages a pool of them which saves constantly having to allocate them. To use these you want to look in the OMR project at the compiler/infra/Checklist.hpp header. In there you will find `NodeChecklist` and `BlockChecklist`. You would just want to do:
`TR::NodeChecklist visited(comp());`
That checklist has appropriate `add` `contains` and `remove` methods which will make things easy.

If you could have a look and rework to use the `TR::NodeChecklist` that would be great since that is the direction we want to move to for checking for repeated visits.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -0,0 +1,61 @@
...
+   for (auto treetop = comp()->getMethodSymbol()->getFirstTreeTop(); treetop != NULL; treetop = treetop->getNextTreeTop())
+      {
+      if (treetop->getNode()->getNumChildren() > 0)
+         {
+         auto node = treetop->getNode()->getFirstChild();
+         if (node && node->getOpCode().isCall())
```
andrewcraik: moving the visited check up to this level will save compile-time without changing behavior. It will save repeated calls to isInlineable.

andrewcraik:
The new version is looking good - I just asked for a couple of changes around the infrastructure for checking if nodes exist. Thank you for continuing to work on this and for your contribution to OMR - this will be a useful utility for languages going forward so we appreciate your help in refining the infrastructure.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
0dvictor: Good point, updated.

compiler/optimizer/OMRRecognizedCallTransformer.cpp
0dvictor: Wonderful, the `TR::NodeChecklist` approach is much cleaner than the `TR_BitVector` one.  Updated.

0dvictor:
Thank you, I've updated the code.

andrewcraik:
Looks good - @vijaysun-omr any further concerns from you?

vijaysun-omr:
Thanks for the change @0dvictor and your patience through the review process. Merging now.

DanHeidinga:
@vijaysun-omr I don't see any indication that the OMR genie builds were triggered for this change?  Were they handled in some way that doesn't show up in this PR?

We're investigating a build break in openj9-omr on ppcle that looks like it may be related to this PR and I'd like to confirm that its passed a ppcle build with just OMR.

DanHeidinga:
Looks like the OMR nightly (?) build on ppcle containing this change passed - https://ci.eclipse.org/omr/job/Build-linux_ppc-64_le_gcc/239/ - which means the issue should be on the openj9 side of the world

andrewcraik:
Yes I believe the issue is that OpenJ9 is not compiling the new files that were added because they are not in the OpenJ9 makefiles which are separate from those here in OMR.

PR #1842, opened 2017-10-24
Validate ireturn child types

mgaudet:
~This has a knock on effect, requiring the addition of extra conversions
in some test cases, as well as some new tests for the validator.~

Validate the child types of ireturn

### DIFF
--- a/compiler/il/ILOpCodeProperties.hpp
+++ b/compiler/il/ILOpCodeProperties.hpp
@@ -657,7 +657,7 @@
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::Int32,
    /* .typeProperties       = */ ILTypeProp::Size_4 | ILTypeProp::Integer,
-   /* .childProperties      = */ ONE_CHILD(ILChildProp::UnspecifiedChildType),
+   /* .childProperties      = */ ONE_CHILD(ILChildProp::UnspecifiedChildType), // ireturn is used to return all types smaller than Int32
    /* .swapChildrenOpCode   = */ TR::BadILOp,
    /* .reverseBranchOpCode  = */ TR::BadILOp,
    /* .booleanCompareOpCode = */ TR::BadILOp,
--- a/compiler/ras/ILValidator.cpp
+++ b/compiler/ras/ILValidator.cpp
@@ -260,12 +260,12 @@ bool TR::ILValidator::treesAreValid(TR::TreeTop *start, TR::TreeTop *stop)
          for (auto i = 0; i < actChildCount; ++i)
             {
             auto childOpcode = node->getChild(i)->getOpCode();
+            const auto expChildType = opcode.expectedChildType(i);
+            const auto actChildType = childOpcode.getDataType().getDataType();
+            const auto expChildTypeName = (expChildType >= TR::NumTypes) ? "UnspecifiedChildType" : TR::DataType::getName(expChildType);
+            const auto actChildTypeName = TR::DataType::getName(actChildType);
             if (childOpcode.getOpCodeValue() != TR::GlRegDeps)
                {
-               const auto expChildType = opcode.expectedChildType(i);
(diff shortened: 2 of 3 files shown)


Leonardo2718:
Although I completely agree with the intention of this change, there is a slight subtlety here that must be considered. In `/home/leonardo/Projects/omr/compiler/il/OMRILOps.hpp`, `OMR::ILOpCode::returnOpCode()` is implemented as follows:

```c++
static TR::ILOpCodes returnOpCode(TR::DataType type)
   {
   switch(type)
      {
      case TR::Int8:     return TR::ireturn;
      case TR::Int16:    return TR::ireturn;
      case TR::Int32:    return TR::ireturn;
      case TR::Int64:    return TR::lreturn;
      case TR::Address:  return TR::areturn;
      case TR::Float:    return TR::freturn;
      case TR::Double:   return TR::dreturn;
      case TR::NoType:   return TR::Return;   // void return
      default: TR_ASSERT(0, "no return opcode for this datatype");
      }
   return TR::BadILOp;
   }
```

This has the consequence of effectively allowing an IL generator in a downstream project to accidentally miss the required type conversion. This is also the source of the validation failures in the Travis builds for this PR.

mgaudet:
Interesting. From an ABI perspective I can sort of see why this is.  The question is how to resolve this. 

Is it mandated that the `ireturn` evaluators handle smaller types then? @fjeremic / @ymanton ? ( @Leonardo2718 tells me offline that the evaluator on x86 already 'just works') 

 Or should the correct answer here be that all attempts to return a smaller type explicitly widen via `b2i` or `s2i` before hitting `ireturn`? 

Or does this get resolved by implementing `breturn` and `sreturn`, and having them act like `ireturn [b|s]2i`

mgaudet:
One option here is I remove the child type annotation from `ireturn` for now,  and add an explicit check to the validator, ensuring the type is one of `i`, `s`, or `b`.  

Then there will need to be more testing of the `ireturn` evaluator for making sure it does the correct widening.

mgaudet:
Ok, I've gone ahead with the second option.

ymanton:
> Is it mandated that the ireturn evaluators handle smaller types then? @fjeremic / @ymanton ? ( @Leonardo2718 tells me offline that the evaluator on x86 already 'just works')

I think this is a Java-ism, not an ABI thing (although it happens to look ABI-ish). `ireturn` is only expected return `Int32` because in the world of Java bytecode that's the smallest integer you can return. In our IL if you have something smaller you'll have to widen it with the appropriate conversion because the code generator (at least the Power one that I'm looking at) will not do any sort of widening for you. The OpenJ9 IL generator does this explicitly when loading integers smaller than 32 bits and it does this everywhere, not just as a special case for `ireturn`, so small integers have a very limited existence in general.

> Or should the correct answer here be that all attempts to return a smaller type explicitly widen via b2i or s2i before hitting ireturn?
> 
> Or does this get resolved by implementing breturn and sreturn, and having them act like ireturn [b|s]2i

I think this is a mismatch between how OMR wants/expects our IL to behave and how it and our various code generators actually behave due to their Java heritage. I think what we want (or maybe it's just me) is for the OMR IL to be free of these sorts of Java-isms and treat small integers (and unsigned integers for that matter) as first class citizens that can be operated on in the same way as their signed 32- and 64-bit counterparts.

mgaudet:
Not quite sure what's up with the travis build failure: On my machine, the log just looks blank: https://travis-ci.org/eclipse/omr/jobs/292280898

0xdaryl:
Restarted...

0xdaryl:
genie-omr build all

mgaudet:
Given the comment above, I'm going to open issues for implementing `breturn` and `sreturn`. Once those are done, we will be able to edit this validation rule to require using those returns, or explicit widening conversions.

PR #224, opened 2016-09-21
Add more documentation for optimizations

xliang6:
Add documentation for dead trees elimination and
local value propagation.

Signed-off-by: Xiaoli Liang <xsliang@ca.ibm.com>

PR #1849, opened 2017-10-25
Allow for non-JNI Unsafe methods

JamesKingdon:
In Java9 sun.misc.Unsafe JNI methods are moved to jdk.internal and
replaced with simple wrappers. This leads to a few places in the
code where we do the wrong thing by assuming that the recognized
Unsafe method is a JNI method.
This commit adds explicit tests to differentiate between the JNI
methods and the non-JNI wrappers.

### DIFF
--- a/compiler/optimizer/RedundantAsyncCheckRemoval.cpp
+++ b/compiler/optimizer/RedundantAsyncCheckRemoval.cpp
@@ -420,12 +420,21 @@ bool TR_RedundantAsyncCheckRemoval::callDoesAnImplicitAsyncCheck(TR::Node *callN
        (symbol->getRecognizedMethod()==TR::java_lang_Integer_rotateLeft) ||
        (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateLeft) ||
        (symbol->getRecognizedMethod()==TR::java_lang_Integer_rotateRight) ||
-       (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateRight) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapInt_jlObjectJII_Z) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapLong_jlObjectJJJ_Z) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapObject_jlObjectJjlObjectjlObject_Z)
+       (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateRight)
        )
        return false;
+
+   // Beginning in Java9 we use the same enum values for both the
+   // sun.misc.Unsafe wrappers and the jdk.internal.misc.Unsafe JNI methods, so
+   // for these we need to do an additional test to check if they are the native
+   // versions or not.
+   if (symbol->isNative() && 
+       ((symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapInt_jlObjectJII_Z) ||
+       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapLong_jlObjectJJJ_Z) ||
+       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapObject_jlObjectJjlObjectjlObject_Z)) 
+      )
+      return false;
+
 
    if ((symbol->getRecognizedMethod()==TR::java_math_BigDecimal_DFPPerformHysteresis)  ||
        (symbol->getRecognizedMethod()==TR::java_math_BigDecimal_DFPUseDFP) ||
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -625,21 +625,33 @@ int32_t OMR::X86::CodeGenerator::getMaximumNumbersOfAssignableGPRs()
    }
 
 /*
- * this method returns TRUE for all the cases we decide NOT to inline the CAS native
- * it checks the same condition as inlineCompareAndSwapNative in tr.source/trj9/x/codegen/J9TreeEvaluator.cpp
+ * This method returns TRUE for all the cases we decide NOT to replace the call to CAS
+ * with inline assembly.
+ * It checks the same condition as inlineCompareAndSwapNative in tr.source/trj9/x/codegen/J9TreeEvaluator.cpp
  * so that GRA and Evaluator should be consistent about whether to inline CAS natives
  */
 static bool willNotInlineCompareAndSwapNative(TR::Node *node,
       int8_t size,
-      bool isObject,
       TR::Compilation *comp)
    {
 #ifdef J9_PROJECT_SPECIFIC
+   TR::SymbolReference *callSymRef = node->getSymbolReference();
+   TR::MethodSymbol *methodSymbol = callSymRef->getSymbol()->castToMethodSymbol();
+
    if (TR::Compiler->om.canGenerateArraylets() && !node->isUnsafeGetPutCASCallOnNonArray())
       return true;
    static char *disableCASInlining = feGetEnv("TR_DisableCASInlining");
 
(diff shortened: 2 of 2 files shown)


JamesKingdon:
This PR addresses https://github.com/eclipse/openj9/issues/321 for the OMR code.
A further PR will be required for changes in OpenJ9 to fully address the problem. I think the two PRs will be independent, although both will be required to resolve the issue.

JamesKingdon:
Personal build #224 was clean with these changes.

vijaysun-omr:
I would like a review from @andrewcraik

compiler/optimizer/RedundantAsyncCheckRemoval.cpp
```diff
@@ -420,12 +420,21 @@ bool TR_RedundantAsyncCheckRemoval::callDoesAnImplicitAsyncCheck(TR::Node *callN
...
+   // In Java9 we use the same enum values for both the sun.misc.Unsafe wrappers
+   // and the jdk.internal.misc.Unsafe JNI methods, so for these we need to do an
+   // additional test to check if they are the native versions or not.
+   if ((
        (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapInt_jlObjectJII_Z) ||
        (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapLong_jlObjectJJJ_Z) ||
        (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapObject_jlObjectJjlObjectjlObject_Z)
-       )
-       return false;
+       ) && symbol->isNative()
```
andrewcraik: This test (isNative) is probably cheaper so we should put it first to short circuit the check...

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -662,7 +674,8 @@ static bool willNotInlineCompareAndSwapNative(TR::Node *node,
 #endif
    }
 
-//some recognized methods might be transformed into very simple hardcoded assembly sequence which doesn't need register spills as a real call does
+// Some recognized methods are transformed into very simple hardcoded
+// assembly sequences which don't need register spills as real calls do
```
andrewcraik: Not sure how this comment relates to the method?

JamesKingdon: As far as I can see the comment is accurate. What sort of changes did you have in mind?

andrewcraik: I think a proper doxygen stating this method return false if the code generator has special handling for a given method that will result in a simple inline assembly sequence or similar. The current comment just seems a bit out of context - albeit related.

JamesKingdon: Ah, ok. I'll do that.

andrewcraik:
Thanks @JamesKingdon new code looks good to me - over to @vijaysun-omr. The commits should be squashed before merging.

JamesKingdon:
Rebased and squashed.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -625,23 +625,35 @@ int32_t OMR::X86::CodeGenerator::getMaximumNumbersOfAssignableGPRs()
    }
 
 /*
- * this method returns TRUE for all the cases we decide NOT to inline the CAS native
- * it checks the same condition as inlineCompareAndSwapNative in tr.source/trj9/x/codegen/J9TreeEvaluator.cpp
+ * This method returns TRUE for all the cases we decide NOT to replace the call to CAS
```
0xdaryl: Instead of tinkering with this function in OMR and adding even more Java-specific code, I would rather see it moved directly into the OpenJ9 project.  It is a static helper function that is ONLY called from code guarded with `#ifdef J9_PROJECT_SPECIFIC`.  You should be able to make it extern, hoist it out of this file with relatively little effort, and then do whatever is necessary for Java.

The comment below also suggests that there's similar logic in J9TreeEvaluator that it must be kept in sync with.  It seems to me that having both of those checks in the same project would be a step toward commoning them into one and would be easier to maintain.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -662,7 +674,17 @@ static bool willNotInlineCompareAndSwapNative(TR::Node *node,
 #endif
    }
 
-//some recognized methods might be transformed into very simple hardcoded assembly sequence which doesn't need register spills as a real call does
+   
+/** Identify methods which are not transformed into inline assembly.
```
0xdaryl: Thanks for adding the Doxygen comment.  Could you use `@brief` in front of the first couple sentences?

compiler/optimizer/RedundantAsyncCheckRemoval.cpp
```diff
@@ -420,13 +420,21 @@ bool TR_RedundantAsyncCheckRemoval::callDoesAnImplicitAsyncCheck(TR::Node *callN
...
        (symbol->getRecognizedMethod()==TR::java_lang_Integer_rotateRight) ||
-       (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateRight) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapInt_jlObjectJII_Z) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapLong_jlObjectJJJ_Z) ||
-       (symbol->getRecognizedMethod()==TR::sun_misc_Unsafe_compareAndSwapObject_jlObjectJjlObjectjlObject_Z)
+       (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateRight)
        )
        return false;
 
+   // In Java9 we use the same enum values for both the sun.misc.Unsafe wrappers
```
0xdaryl: The "Java9" comments seem very point-in-time to me.  Will someone reading these comments 5 years from now think this only applies to Java9?   Perhaps saying "Beginning in Java9..." might be better.

compiler/x/codegen/OMRCodeGenerator.cpp
JamesKingdon: I'm very much in favour of this, but would it be ok if I opened a new issue for the work? I'd like to get the Unsafe fix into the code promptly, and this would be my first attempt to move code between projects, so I don't want to try and rush it.

andrewcraik: I think that moving it down is a good idea, but since it is broken at present in at least some cases fixing it then moving it seems reasonable to me - that way we only move working code which is probably a good principle to have in general when there is a choice. Your thoughts @0xdaryl ?

0xdaryl: OK, I'm fine with an issue.

JamesKingdon: Opened #1942

compiler/optimizer/RedundantAsyncCheckRemoval.cpp
```diff
@@ -420,13 +420,22 @@ bool TR_RedundantAsyncCheckRemoval::callDoesAnImplicitAsyncCheck(TR::Node *callN
        (symbol->getRecognizedMethod()==TR::java_lang_Integer_rotateLeft) ||
        (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateLeft) ||
        (symbol->getRecognizedMethod()==TR::java_lang_Integer_rotateRight) ||
-       (symbol->getRecognizedMethod()==TR::java_lang_Long_rotateRight) ||
```
vijaysun-omr: If this code is shared between Java8 (where the methods are native) and Java9 (where they are not native as you mentioned in the abstract), how does this change work in both cases ?

vijaysun-omr:
It looks like the requested changes have either been made or issues opened. I reviewed the change myself now and it looks okay to me. @0xdaryl are you fine with this getting merged (just thought I would check before doing so) ?

0xdaryl:
@vijaysun-omr : yes, please merge.

PR #1853, opened 2017-10-26
Reverse logic in getDataType

mgaudet:
This avoids the peculiar double-negative construction of
if (not has no data type)


charliegracie:
genie-omr build all

PR #1861, opened 2017-10-26
Enhance ArrayCopy for 16-bit primitives on X86

0dvictor:
Use REP MOVSD to copy 16-bit primitives instead of REP MOVSW.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -3496,6 +3496,24 @@ static void generateRepMovsInstruction(TR_X86OpCodes repmovs, TR::Node *node, TR
    generateInstruction(repmovs, node, dependencies, cg);
    }
 
+/** \brief
+*    Generate instructions to do array copy for 64-bit primitives on IA-32.
+*
+*  \param node
+*     The tree node
+*
+*  \param dstReg
+*     The destination array address register
+*
+*  \param srcReg
+*     The source array address register
+*
+*  \param sizeReg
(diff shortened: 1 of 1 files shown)


compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3992,6 +4049,7 @@ TR::Register *OMR::X86::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::Co
 
    TR::DataType dt = node->getArrayCopyElementType();
    uint32_t elementSize = 1;
+   static bool avoidREPMOVSW = feGetEnv("TR_AvoidREPMOVSW");
```
0xdaryl: Please provide a slightly more descriptive env var to indicate context.  e.g., `TR_AvoidREPMOVSWForArrayCopy`

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3568,6 +3568,63 @@ static void arrayCopy64BitPrimitiveOnIA32(TR::Node* node, TR::Register* dstReg,
    cg->stopUsingRegister(scratch);
    }
```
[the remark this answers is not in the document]
0dvictor: Good suggestion, updated.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Sure, it has been updated.

0xdaryl:
Genie build is not required as x86-only files were modified.

PR #1899, opened 2017-10-30
X86 fastcall convention support

0dvictor:
Introducing a new linkage TR_FastCall. It is mapped to the following calling convention in each platform:
64-bit Linux: System V AMD64 ABI (same as TR_System)
64-bit Windows: Microsoft x64 calling convention (same as TR_System)
32-bit Linux: fastcall with sseregparm
32-bit Windows: fastcall

### DIFF
--- a/compiler/x/amd64/codegen/AMD64FastCallConfig.hpp
+++ b/compiler/x/amd64/codegen/AMD64FastCallConfig.hpp
@@ -1,0 +1,110 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "codegen/X86FastCallLinkage.hpp"
+
+// Windows: Microsoft x64 calling convention
+// Linux:   System V AMD64 ABI.
+
+const TR::RealRegister::RegNum TR::X86FastCallCallSite::IntParamRegisters[] =
+   {
+#ifdef WINDOWS
+   TR::RealRegister::ecx,
+   TR::RealRegister::edx,
+#else
+   TR::RealRegister::edi,
+   TR::RealRegister::esi,
+   TR::RealRegister::edx,
+   TR::RealRegister::ecx,
+#endif
+   TR::RealRegister::r8,
+   TR::RealRegister::r9,
+   };
+
+const TR::RealRegister::RegNum TR::X86FastCallCallSite::FloatParamRegisters[] =
+   {
+   TR::RealRegister::xmm0,
+   TR::RealRegister::xmm1,
+   TR::RealRegister::xmm2,
+   TR::RealRegister::xmm3,
+#ifndef WINDOWS
+   TR::RealRegister::xmm4,
+   TR::RealRegister::xmm5,
+   TR::RealRegister::xmm6,
+   TR::RealRegister::xmm7,
+#endif
+   };
+
+const TR::RealRegister::RegNum TR::X86FastCallCallSite::CallerSavedRegisters[] =
+   {
+   TR::RealRegister::eax,
+   TR::RealRegister::ecx,
+   TR::RealRegister::edx,
+#ifndef WINDOWS
+   TR::RealRegister::edi,
+   TR::RealRegister::esi,
+#endif
+   TR::RealRegister::r8,
+   TR::RealRegister::r9,
+   TR::RealRegister::r10,
+   TR::RealRegister::r11,
+   TR::RealRegister::xmm0,
+   TR::RealRegister::xmm1,
+   TR::RealRegister::xmm2,
+   TR::RealRegister::xmm3,
+   TR::RealRegister::xmm4,
+   TR::RealRegister::xmm5,
+   TR::RealRegister::xmm6,
+   TR::RealRegister::xmm7,
+   TR::RealRegister::xmm8,
+   TR::RealRegister::xmm9,
+   TR::RealRegister::xmm10,
+   TR::RealRegister::xmm11,
+   TR::RealRegister::xmm12,
+   TR::RealRegister::xmm13,
+   TR::RealRegister::xmm14,
+   TR::RealRegister::xmm15,
+   };
+
+const TR::RealRegister::RegNum TR::X86FastCallCallSite::CalleeSavedRegisters[] =
+   {
+   TR::RealRegister::ebx,
+#ifdef WINDOWS
+   TR::RealRegister::edi,
+   TR::RealRegister::esi,
+#endif
+   TR::RealRegister::ebp,
+   TR::RealRegister::esp,
+   TR::RealRegister::r12,
+   TR::RealRegister::r13,
+   TR::RealRegister::r14,
+   TR::RealRegister::r15,
+   };
+
+const bool   TR::X86FastCallCallSite::CalleeCleanup = false;
+const size_t TR::X86FastCallCallSite::StackSlotSize = 8;
+
+// Windows X86-64 requires caller reserves shadow space for parameters passed via registers
+#ifdef WINDOWS
+const bool TR::X86FastCallCallSite::RegisterParameterShadowOnStack = true;
+#else
+const bool TR::X86FastCallCallSite::RegisterParameterShadowOnStack = false;
+#endif
--- a/compiler/x/codegen/RealRegisterManager.hpp
+++ b/compiler/x/codegen/RealRegisterManager.hpp
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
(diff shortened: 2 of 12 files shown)


0xdaryl:
FYI...  @mgaudet is working on extern call support with a user-specified linkage in Tril (#1910) that should make creating unit test cases to exercise this kind of code relatively straightforward.  I would like to see some unit test cases written with it once its ready.

mgaudet:
So, given the linkage convention enum is shared across architectures, is there a particular expectation for `TR_Fastcall` linkage? Or is just a linkage expected to be 'faster' than the default linkage? Would aliasing this to `TR_SystemLinkage` break expectations? 

A naming aside: Not being a linkage expert, squinting, it seems like this code is more trying to implement something similar to the [Microsoft vectorcall linkage](https://en.wikipedia.org/wiki/X86_calling_conventions#Microsoft_vectorcall), rather than fast call.

compiler/x/codegen/RealRegisterManager.hpp
```diff
@@ -0,0 +1,91 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2017 IBM Corp. and others
```
fjeremic: This looks like a new file. The initial copyright date seems incorrect. Similarly elsewhere in this PR.

compiler/x/codegen/RealRegisterManager.hpp
```diff
@@ -0,0 +1,91 @@
...
+      for (uint8_t i = (uint8_t)TR::RealRegister::NoReg; i < (uint8_t)TR::RealRegister::NumRegisters; i++)
+         {
+         if (_Registers[i] != NULL)
+            {
+            _cg->stopUsingRegister(_Registers[i]);
+            }
+         }
+      }
+   // Find or allocate a virtual register which is bind to the given real register
+   TR::Register* Use(TR::RealRegister::RegNum RealRegister)
```
fjeremic: `Use` seems to be orthogonal to the camelCase naming convention used throughout the entire compiler. Similarly for `BuildRegisterDependencyConditions`, `NumberOfRegistersInUse`, etc. We should try to stay consistent if possible.

0dvictor:
The implementation is actually gcc's `fastcall` with `sseregparm`, which is slightly different from Microsoft's `vectorcall`. `vectorcall` is only supported by MSVS 2013 or newer; however OMR is required to support MSVS 2010.
Linkages on X86-32 is such a complicated topic. There are too many choices and none of them is the standard. For OMR, `TR_System` is mapped to `cdecl`, which is used by Linux APIs and Microsoft's MSVCRT. `cdecl` requires floating point return value being passed by X87 register ST(0) even if both caller and callee use only SSE/AVX. Therefore I'm proposing OMR to adopt `fastcall` with `sseregparam`, to avoid the cost of unnecessary value movements between SSE/AVX and X87, for OMR's internal calls and helper calls.
On X86-64, `TR_FastCall` is identical to `TR_System`, and I'm expecting they are identical on ARM, Power, and Z.

compiler/x/codegen/RealRegisterManager.hpp
0dvictor: Thanks for pointing it out, it has been updated.

compiler/x/codegen/RealRegisterManager.hpp
0dvictor: Sure, I've updated them.

compiler/x/amd64/codegen/AMD64FastCallConfig.hpp
```diff
@@ -0,0 +1,111 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2017 IBM Corp. and others
```
mstoodle: Are these files marked as copyright 2000 because they contain excerpts from other files that have copyright 2000?

mstoodle:
@0xdaryl @0dvictor #1910 was merged a while back. Is the expectation now that tril unit tests should be written to test this fastcall linkage before this PR is merged?

compiler/x/amd64/codegen/AMD64FastCallConfig.hpp
0dvictor: Thank you for point it out. In this case, it was a mistake; X86FastCallLinkage.hpp and X86FastCallLinkage.cpp contain modified code from OpenJ9, which have copyright 2000.

0dvictor: Updated.

mstoodle:
This PR blocks another PR at OpenJ9 and they've both been open for several months without much recent documented progress. Would you please outline the plan here @0dvictor @andrewcraik ?

0dvictor:
The functional change is ready; however, its related tests are not there yet. I don't know whether we should wait for the tests or get this in first.

mstoodle:
So, I guess one on-point question: can you explain what you've done to verify that the functional implementation is correct?

0dvictor:
I have test the functionalities with https://github.com/eclipse/openj9/pull/477 . The combination passed all IBM's internal Java tests.

0dvictor:
I've created some sanity tests using ones for system linkage as template. Note that tril based OMR tests do not cover all code paths of fastcall linkage. It may or may not be a limitation that tril based OMR tests always generate a caller-callee pair such that their addresses are more than 2G apart. Therefore, these tests only exercise the path that a trampoline is used. Fortunately, OpenJ9 always exercises the other path that caller and callee are less than 2G apart. I hope by tests on both OMR and OpenJ9 side, it can be confidence that this PR is correct and good to go.

compiler/x/codegen/X86FastCallLinkage.cpp
```diff
@@ -0,0 +1,240 @@
...
+   // Build parameters, parameters in _Params are Right-to-Left (RTL)
+   TR::Register* IntParams[NumberOfIntParamRegisters] = {};
+   TR::Register* FloatParams[NumberOfFloatParamRegisters] = {};
+   TR_Stack<TR::Register*> StackParams(cg()->trMemory());
+
+   size_t IntParamIndex = 0;
+#ifndef WINDOWS
+   size_t FloatParamIndex = 0;
+#else
+   auto& FloatParamIndex = IntParamIndex;
```
mstoodle: I'm not a big fan of this construct. While it moves the #ifdef out of the loop, it makes that loop code somewhat harder to understand (the parameter indices could be the same variable or not, depending on the platform). This is an example where I think I'd actually prefer to see the #ifdef code inside the loop (gasp! I know...)

compiler/x/codegen/X86FastCallLinkage.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
```
mstoodle: another out of date SPDX identifier.

compiler/x/codegen/X86FastCallLinkage.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include "codegen/Linkage.hpp"
+#include "env/jittypes.h"
+
+namespace TR { class CodeGenerator; }
+namespace TR { class Instruction; }
+namespace TR { class LabelSymbol; }
+namespace TR { class Node; }
+namespace TR { class RegisterDependencyConditions; }
+
+#define IMCOMPLETELINKAGE  "This class is only used to generate call-out sequence but no call-in sequence, so it is not used as a complete linkage."
```
mstoodle: typo: INCOMPLETELINKAGE (note second letter)

compiler/x/i386/codegen/IA32FastCallConfig.hpp
```diff
@@ -0,0 +1,69 @@
...
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
```
mstoodle: yup, another SPDX id needs to be updated :)

mstoodle:
@0dvictor what do you think about changing that remaining loop structure?

mstoodle:
I can see what you're saying. I'm just not sure (as a "reader" of that code) that I agree. 
But I'm not sure I feel strongly enough about it to require you to change it. At the very least, an informative comment is warranted.

I'll defer to a second opinion from @0xdaryl on whether he thinks this variable aliasing for the Windows platform is better or worse than putting the #ifdef inside the loop (or maybe he'll suggest a third option for even more fun).

mstoodle:
Ping to @0xdaryl for his opinion :)

andrewcraik:
@0dvictor, @0xdaryl and @mstoodle this looks to have gone stale - do we know what the next steps are?

mstoodle:
Still marked as WIP @0dvictor what do you think of closing this PR and then we can look at it with fresh eyes once ready for merging (though I'd still like to see a review from @0xdaryl who's currently on vacation).

mstoodle:
Thanks @0dvictor please @ ping me when the PR is ready.

0dvictor:
@mstoodle FYI, this changeset has been rebased and re-verified.

compiler/x/codegen/X86FastCallLinkage.cpp
mstoodle: I am still not a huge fan, but I can't argue that it doesn't match how the Windows ABI maps registers. I'll withdraw the opinion as not important enough to block the change.

mstoodle:
genie-omr build all

0dvictor:
For sure, I'll update the commit message to the following once tests pass:

Introducing a new linkage TR_FastCall. It is mapped to the following calling convention in each platform:
64-bit Linux: System V AMD64 ABI (same as TR_System)
64-bit Windows: Microsoft x64 calling convention (same as TR_System)
32-bit Linux: fastcall with sseregparm
32-bit Windows: fastcall

mstoodle:
@0dvictor looks like a possibly valid failure ...

0dvictor:
The failure occurred at the trampoline to call an address that not fit into a 32-bit RIP relative. It is never used in downstream projects that I run all tests, and did not occur in my local OMR sanity test. It has been fixed.

0dvictor:
@mstoodle do you mind take a look at the updated change?

mstoodle:
genie-omr build all

mstoodle:
I'm ok with the commits as they are, but I'm not the x86 expert. @0xdaryl would you please review the x86 details of this change?

0xdaryl:
I am aiming to give this a comprehensive review by EOW.

mstoodle:
Still pondering @0xdaryl :) ?

Leonardo2718:
ping @0xdaryl

Leonardo2718:
@0dvictor there seem to be a few merge conflicts that need to be resolved.

0dvictor:
The conflicts have been resolved.

Leonardo2718:
Seems like `getX86RealRegister` is being used instead of `getRealRegister`. An artefact of merge conflict resolution?

0dvictor:
Good catch, fixed.

Leonardo2718:
@genie-omr build all

0xdaryl:
My principal concerns with this PR are that it's being done in a different way than how the other linkages are implemented and introduces a subtly different coding style than found in other parts of the x86 code generator (or the compiler for that matter).  For instance, this implementation does not use the LinkageProperties structure to describe the registers.  While I do think the LinkageProperties needs an overhaul (it's about to celebrate its 20th birthday!), I'm not convinced that the solution proposed in this PR is entirely the way to go (I do like some of the ideas though), and I am hesitant to introduce another design variant into the codebase without a plan to bring all linkages to use a common design (not just x86, but across all platforms).  I will ultimately be creating an issue to revisit the LinkageProperties design as part of #2082.

While my concerns could be addressed in this PR, given that the author is no longer actively involved in the project and is unlikely to make progress with moving this PR along, I am closing it.

PR #1904, opened 2017-10-30
Fix AIX CMake compiles

charliegracie:
- remove omrcore when it is not actually required
- DDR blob reader requires omrutil
- sigtest requires fvtest/util on the include path
- thread needs to handle adding pthread for AIX


charliegracie:
This PR allows you to compile AIX 64 bit OMR using CMake.  I did not test with the COMPILER flag on since I did not have access to a machine with flex installed or the ability to install it.

charliegracie:
This is the command line I used to run CMAKE:
`cmake -Wdev -DCMAKE_XL_CreateExportList="/usr/vac/bin/CreateExportList -X64" -C../cmake/caches/Travis.cmake ..`

charliegracie:
FYI @rwy0717 and @dnakamura

dnakamura:
LGTM

PR #1905, opened 2017-10-31
Fix z/TPF unique build error for ELFObjectFileGenerator.cpp

jjohnst-us:
Minor update to assist z/TPF platorm in compiling 
ELFObjectFileGenerator.cpp 

### DIFF
--- a/compiler/codegen/ELFObjectFileGenerator.cpp
+++ b/compiler/codegen/ELFObjectFileGenerator.cpp
@@ -21,6 +21,7 @@
 
 #if defined(LINUX)
 
+#include <stdio.h>
 #include <string.h>
 #include <unistd.h>                     // for getpid, pid_t
 #include "codegen/ELFObjectFileGenerator.hpp"


0xdaryl:
Could you elaborate on what the "unique build error" is?  Perhaps the problem can be fixed by reworking the code rather than introducing a zTPF-specific macro.

jjohnst-us:
z/TPF is an ebcdic platform.

So we make use of an a2e environment similar to z/OS where segments are compiled with intermediary standard headers which redefine printf (eg) to atoe_printf and links to a library which has logic to translate function parameters and function returns making it transparent to the compiled unit.  (no updates required in theory).

However, C++ standard library include file also undefine symbols as part of namespace definitions for standard functions, and it doesn't play well with this library.

So we get something like:
In file included from /ztpf/commit/opensource/include46/g++/ext/string_conversions.h:39:0,
                 from /ztpf/commit/opensource/include46/g++/bits/basic_string.h:2756,
                 from /ztpf/commit/opensource/include46/g++/string:54,
                 from /tpfjava/cleanbuild/jvmupgrade/build/tmp/bld_367600/bld_linux_ztpf_390-64/tr.source/../omr/compiler/codegen/ELFObjectFileGenerator.hpp:31,
                 from /tpfjava/cleanbuild/jvmupgrade/build/tmp/bld_367600/bld_linux_ztpf_390-64/tr.source/../omr/compiler/codegen/ELFObjectFileGenerator.cpp:28:
/projects/jvmport/userfiles/opensource/include46/g++/cstdio:110:11: error: '::fputs' has not been declared
/projects/jvmport/userfiles/opensource/include46/g++/cstdio:123:11: error: '::putc' has not been declared
/projects/jvmport/userfiles/opensource/include46/g++/cstdio:125:11: error: '::puts' has not been declared

It turns out to be a circular header include nightmare.

I was able to pull out the stdarg.h relatively trivially and include it in our a2e stdio.h header.  However the stdio.h circilry dependency looks like it would require an opensource header update to <string>.

It makes it a whole lot easier if we can add the udpate into the omr side, eventhough its a bit odd, but clean.  The only reason we added the OMRZTPF guardmacro is so that in the future it would make a contributor think twice before removing the header.

So if I remove the stadarg.h wold that be a compromise?  It makes our z/TPF maintenance easier.

Thoughts?

jjohnst-us:
Is this okay?

compiler/codegen/ELFObjectFileGenerator.cpp
```diff
@@ -21,6 +21,10 @@
 
 #if defined(LINUX)
 
+#if defined(OMRZTPF)
+#include <stdio.h>
+#endif /* defined(OMRZTPF) */
```
0xdaryl: Can you just include `stdio.h` unconditionally?  This source file does use features found in stdio.h, so it's not wrong to include it regardless of the build target.  Or am I missing something?

jjohnst-us: @0xdaryl So the only reason I added our platform guard macro was so that someone in the future would think twice about removing it but I think that is highly unlikely someone would remove stdio.h.

I removed the guard macros per your request.  Thanks.!

0xdaryl:
Thanks.  Please remember to also fix the spelling of `ELFObjectFileGenerator.cpp` in your commit titles.

0xdaryl:
genie-omr build all

PR #1906, opened 2017-10-31
Remove z/TPF version of omrrttime.s

jjohnst-us:
z/TPF can simply use the s39064 version of omrtime
support.


charliegracie:
genie-omr build zlinux

PR #1908, opened 2017-10-31
Stop versioning unimportant checks in outer loops

jdmpapin:
Upon entry into a loop, a very low frequency check may not run at all before control exits the loop. If a check isn't going to run, the loop can complete normally even if that check would fail. As a result, versioning the check is likely to spuriously send execution to the cold loop. Furthermore, there is little benefit to be had from removing a check that seldom runs. For these reasons, versioner tries to avoid versioning "unimportant" checks.

Unimportant checks have so far been identified based on the frequency ratio between the check and the header of the loop under consideration. But this can cause an inconsistency when an outer loop contains an inner loop that contains a check, and the check is invariant w.r.t. to the outer loop.

If the inner loop iterates more frequently than the outer loop, as is often the case, then it's possible that an unimportant check within the inner loop is too frequent to look unimportant at the level of the outer loop. In this case, the entire outer loop, including the inner loop! can be versioned based on the check, even though it would not be considered important enough to version the inner loop only.

This commit avoids the inconsistency by considering the headers of any inner loops within which a check is nested. To determine whether the check is unimportant, its frequency is compared to the frequency of the hottest header. The original importance test is available under the environment variable `TR_loopVersionerUseOuterHeaderForImportance`.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -3173,6 +3173,14 @@ bool TR_LoopVersioner::detectChecksToBeEliminated(TR_RegionStructure *whileLoop,
    whileLoop->getBlocks(&blocksInWhileLoop);
    int32_t loop_size = blocksInWhileLoop.getSize();
 
+   // Blocks are listed in order of a traversal over the tree of regions, i.e.
+   // they are grouped by region. As we iterate, maintain the current innermost
+   // region, and current innermost containing loop. The hottest intervening
+   // loop header only has to be recomputed when the innermost loop changes.
+   TR_RegionStructure *rememberedInnermostRegion = NULL;
+   TR_RegionStructure *rememberedInnermostLoop = NULL;
+   TR::Block *rememberedHotLoopHeader = NULL;
+
    ListIterator<TR::Block> blocksIt(&blocksInWhileLoop);
    TR::Block *nextBlock;
    for (nextBlock = blocksIt.getCurrent(); nextBlock; nextBlock=blocksIt.getNext())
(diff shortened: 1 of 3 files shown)


jdmpapin:
Commit 9a2a33d is shared with #1909.

vijaysun-omr:
Interesting case you point out. 

I thought of the case when the "infrequently executed block" containing the check in the inner loop runs once every time the outer loop runs and at first, was wondering why such a check ought not to be versioned when considering the outer loop (after all it runs in every iteration of the outer loop). I think the logic is that it's not worth taking the risk to version the check for the outer loop 
since it might result in the slow version of the outer loop running and this would mean the slow version of the inner loop also runs. Instead we are okay leaving such an opportunity on the table when versioning the outer loop since the hotter inner loop would be the one that mattered from a performance standpoint anyway and we are choosing to ensure that the fast version of the hotter inner loop (that is expected to be more performance critical) does execute. 

Is that correct ?

jdmpapin:
Right, that's pretty much it. A check that's too risky to version in the inner loop should also be too risky to version in the outer loop, because the versioning tests for the outer loop have even higher stakes (if only by a little).

I considered comparing against the hottest block in the entire loop, to deal with situations where a check that doesn't run is in the outer loop, next to the hot inner loop. In that case, versioning the check can still cause the cold version of the inner loop to run. However, that test would prevent versioning of checks that aren't inside nested loops, and which run in every iteration of the outer loop. Furthermore, the risk that versioning a check will cause cold code to run for other parts of the loop is a general issue in versioner that doesn't require nested loops, e.g.

    boolean useArray = ...;
    int[] array = useArray ? new int[n] : null;
    for (int i = 0; i < n; i++) {
        // [important code here]
        if (useArray) {
            array[i] = ...; // NULLCHK, BNDCHK
        }
    }

I do think that in some cases we might want to exempt certain blocks from the importance test if we can tell that they'll run, especially if the importance test is made even stricter

vijaysun-omr:
I understand the more general issue you point out and agree with being more selective in the case of infrequently executed check(s) only inside an inner loop.

PR #1909, opened 2017-10-31
Add structure exit extraction

jdmpapin:
After deleting edges from the control flow graph, there may be blocks, or structure subgraph nodes more generally, from which execution is guaranteed to exit a containing loop. A fresh structural analysis would consider such nodes to be outside the loop, which is preferable for the sake of later loop optimizations.

Setting aside the general benefit of smaller loops, leaving loop regions to hold onto nodes that don't belong can cause significant problems, such as versioning of checks that are guaranteed to fail.

This series of commits adds `TR_RegionStructure::extractUnconditionalExits`, which searches for nodes that no longer belong within loops, and pulls them out of those loops. This is not done as part of automatic structure repair, because it can be surprising and unwanted when the control flow graph is in an intermediate state. Currently this is done only at the end of value propagation.


jdmpapin:
Commit 9a2a33d is shared with #1908.

compiler/optimizer/Structure.cpp
```diff
@@ -184,7 +186,788 @@ void TR_RegionStructure::cleanupAfterEdgeRemoval(TR::CFGNode *node)
...
+      removeContentsFromRegion(initialStructure, region);
+
+      if (_trace && _comp->getOutFile() != NULL)
+         {
+         traceMsg(_comp, "Structure after moving node into parent:\n<structure>\n");
+         _comp->getDebug()->print(_comp->getOutFile(), _cfg->getStructure(), 0);
+         traceMsg(_comp, "</structure>\n");
+         }
+
+#ifdef DEBUG
```
andrewcraik: This looks loop invariant unless I am reading the nesting wrong - perhaps it should move out?

compiler/optimizer/Structure.cpp
```diff
@@ -184,7 +186,788 @@ void TR_RegionStructure::cleanupAfterEdgeRemoval(TR::CFGNode *node)
...
+
+   private:
+   typedef TR::typed_allocator<TR_Structure*, TR::Region&> StructureAlloc;
+   typedef std::vector<TR_Structure*, StructureAlloc> StructureVec;
+   typedef std::set<TR_Structure*, std::less<TR_Structure*>, StructureAlloc> StructureSet;
+
+   typedef TR::typed_allocator<TR::CFGEdge*, TR::Region&> EdgeVecAlloc;
+   typedef std::vector<TR::CFGEdge*, EdgeVecAlloc> EdgeVec;
+
+   typedef std::pair<TR_RegionStructure*, TR_BitVector> RcEntry;
```
andrewcraik: Any particular reason the TR_BitVectors are by value rather than just being allocated from the region and only storing the pointer to reduce copying?

jdmpapin: It just saves an indirection. [No bit vector contents are copied](https://github.com/jdmpapin/omr/blob/80480340ac05b8cd8da535e4843f452613e4aace/compiler/optimizer/Structure.cpp#L641-L644)

compiler/optimizer/Structure.cpp
jdmpapin: The value of `checkStructure` could be determined outside the loop, but I think it's clearer next to its use here, and the overhead of reading an option should be negligible in this context

vijaysun-omr:
@andrewcraik are you okay with this commit ?

compiler/optimizer/Structure.cpp
andrewcraik: Ok sounds good to me - thanks!

compiler/optimizer/Structure.cpp
andrewcraik: Given the distance from the loop header I can buy this - in general we do want to avoid having too many little things like this creep in since they do eventually add up. I do agree, however, in this situation the clarity probably warrants the current placement.

andrewcraik:
LGTM

andrewcraik:
@vijaysun-omr back to you

vijaysun-omr:
Looks okay to me.

PR #1911, opened 2017-10-31
Clean up genericLongAndAsRotateHelper

dchopra001:
-Rename the function to tryToReplaceLongAndWithRotateInstruction
as this is more descriptive

-Clean up the function to avoid skipping the evaluation of i2l
and iu2l nodes. This should be handled by the simplifier (and
it currently is). Also, insert a static debug counter to catch
potential opportunities in the future.

-Add asserts during binary encoding and instruction construction
to make sure the reserved bits in the instruction's operands are
not being set. This is a source of printing errors in trace logs.

-Test the function using tril + google tests

### DIFF
--- a/compiler/z/codegen/BinaryEvaluator.cpp
+++ b/compiler/z/codegen/BinaryEvaluator.cpp
@@ -2467,35 +2467,38 @@ genericRotateAndInsertHelper(TR::Node * node, TR::CodeGenerator * cg)
    }
 
 TR::Register *
-genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
+OMR::Z::TreeEvaluator::tryToReplaceLongAndWithRotateInstruction(TR::Node * node, TR::CodeGenerator * cg)
    {
    if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10))
       {
       TR::Node * firstChild = node->getFirstChild();
       TR::Node * secondChild = node->getSecondChild();
-      TR::Node * skipedConversion = NULL;
       TR::Compilation *comp = cg->comp();
 
-      // Transform land into RISBG as it may be better than discrete pair of ands
+      // Given the right case, transform land into RISBG as it may be better than discrete pair of ands
       if (node->getOpCodeValue() == TR::land && secondChild->getOpCode().isLoadConst())
          {
-         uint64_t value = secondChild->getLongInt();
+         uint64_t longConstValue = secondChild->getLongInt();
 
-         if ((firstChild->getOpCodeValue() == TR::i2l || firstChild->getOpCodeValue() == TR::iu2l)
-               && firstChild->getReferenceCount()==1 && firstChild->getRegister() == NULL && firstChild->isHighWordZero() )
-            {
-            skipedConversion = firstChild;
-            firstChild = firstChild->getFirstChild();
-            value &= 0x00000000FFFFFFFFL;
-            }
+         int32_t tZeros = trailingZeroes(longConstValue);
+         int32_t lZeros = leadingZeroes(longConstValue);
+         int32_t tOnes  = trailingZeroes(~longConstValue);
+         int32_t lOnes  = leadingZeroes(~longConstValue);
+         int32_t popCnt = populationCount(longConstValue);
+ 
+         // if the population count is 0 then the result of the AND will also be 0,
+         // because a popCnt=0 means there are no overlapping bits between the two AND
+         // operands. We cannot generate a RISBG for this case as RISBG cannot be
+         // used to zero out the contents of the entire register. The starting and ending
+         // positions are inclusive, so even if start=end, we will preserve at least one bit.
+         if (popCnt == 0)
+            return NULL;
 
-         int32_t tZeros = trailingZeroes(value);
-         int32_t lZeros = leadingZeroes(value);
-         int32_t tOnes  = trailingZeroes(~value);
-         int32_t lOnes  = leadingZeroes(~value);
-         int32_t popCnt = populationCount(value);
-         int32_t msBit = -1;
-         int32_t lsBit = -1;
+         // the minimum value for the bit positions is 0, and lsBit can equal msBit iff popCnt = 1
+         // if popCnt == 0 then we cannot generate a RISBG instruction because RISBG preserves at
+         // least one bit position
+         int32_t msBit = 0;
+         int32_t lsBit = 0;
 
          if (comp->getOption(TR_TraceCG))
             {
@@ -2503,29 +2506,73 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
             traceMsg(comp,"          => rotated-and-insert: lOnes %d, tOnes %d\n", lOnes, tOnes);
             }
 
-         bool doIt = false;
+         bool doTransformation = false;
+
+         // We now try to detect if the lconst's bit pattern is such that we can use a single RISBG
+         // instead of multiple AND instructions
+
+         // The first block will catch all cases where the lConstValue is a series of contiguous ones
+         // surrounded by zeros
+         // ex: 0001111000, 011110, 111111, 001111, 11100 etc...
+         //
+         // The 'else if' block will consider all cases where we have a series of contiguous zeros
+         // surrounded by ones
+         // ex: 1110000111, 1000001
          if (popCnt == (64 - lZeros - tZeros))
             {
             msBit = lZeros;
             lsBit = 63 - tZeros;
-            if (firstChild->getReferenceCount()>1 || skipedConversion != NULL   // Better of with non-destructive RISBG as AND?
-                  || lZeros > 31 || tZeros > 31 || (lZeros > 0 && tZeros > 0))  // Better off with a single AND?
+            // The below checks make sure we are not better off using one of Z architecture's
+            // 16-bit AND immediate instructions (NILF,NILH,NILL, NIHF,NIHH, NIHL).
+            // Details for each check below:
+            //
+            // "firstChild->getReferenceCount() > 1"
+            //    If the firstChild's refCount>1, we will need to clobber evaluate this case
+            //    because we cannot destroy the value. In such a case we're better off with 
+            //    RISBG because RISBG won't clobber anything and hence can do the operation
+            //    faster.
+            //
+            // "|| lZeros > 31 || tZeros > 31"
+            //    If there are more than 31 contiguous zeros, then we must use RISBG because
+            //    no NI** instruction do zero out that many bits in a single instruction.
+            //
+            // "|| (lZeros > 0 && tZeros > 0"
+            //    NI** instructions need to explicitly specify which bit positions to zero out.
+            //    If our constant at minimum looks like 0111...110 (there are zeros in the top
+            //    half and bottom half of the register), then it's better to use RISBG.
+            //    This is because there are no NI** instructions allowing us to specify bits in
+            //    the top half and bottom half of the register to zero out.
+
+            if (firstChild->getReferenceCount() > 1 || lZeros > 31 || tZeros > 31
+                || (lZeros > 0 && tZeros > 0))
                {
-               doIt = true;
+               doTransformation = true;
                }
             }
          else if (popCnt == (lOnes + tOnes))
             {
             msBit = 64 - tOnes;
             lsBit = lOnes - 1;
-            if (firstChild->getReferenceCount()>1 || skipedConversion != NULL // Better of with non-destructive RISBG as AND?
-                  || (lOnes < 32 && tOnes < 32) )                             // Better off with a single AND?
+            // The below checks make sure we are not better off using one of Z architecture's
+            // 16-bit AND immediate instructions (NILF,NILH,NILL, NIHF,NIHH, NIHL).
+            // Details for each check below:
+            //
+            // "firstChild->getReferenceCount() > 1"
+            //    same as above
+            //
+            // "(lOnes < 32 && tOnes < 32)"
+            //    This case implies there are zeros at the high/low boundary of the register.
+            //    Example value: 11111000011111. The NI** instructions can only operate on
+            //    the top half or bottom half of a register at a time. So such a case
+            //    will require two NI** instructions. Hence we are better off using a single RISBG
+
+            if (firstChild->getReferenceCount()> 1 || (lOnes < 32 && tOnes < 32))
                {
-               doIt = true;
+               doTransformation = true;
                }
             }
 
-         if (doIt && performTransformation(comp, "O^O Use RISBG instead of 2 ANDs for %p.\n", node) )
+         if (doTransformation && performTransformation(comp, "O^O Use RISBG instead of 2 ANDs for %p.\n", node))
             {
             TR::Register * targetReg = NULL;
             TR::Register * sourceReg = cg->evaluate(firstChild);
@@ -2539,19 +2586,18 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
                targetReg = sourceReg;
                }
 
-            TR::InstOpCode::Mnemonic opCode = cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zEC12) ? TR::InstOpCode::RISBGN : TR::InstOpCode::RISBG;
+               // if possible then use the instruction that doesn't set the CC as it's faster
+               TR::InstOpCode::Mnemonic opCode = cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zEC12) ? TR::InstOpCode::RISBGN : TR::InstOpCode::RISBG;
+ 
+               // this instruction sets the rotation factor to 0 and sets the zero bit(0x80).
+               // So it's effectively zeroing out every bit except the inclusive range of lsBit to msBit
+               // The bits in that range are preserved as is
                generateRIEInstruction(cg, opCode, node, targetReg, sourceReg, msBit, 0x80 + lsBit, 0);
 
-            if (skipedConversion)
-               {
-               skipedConversion->setRegister(targetReg);
-               cg->decReferenceCount(firstChild);
-               }
             return targetReg;
             }
          }
       }
-
    return NULL;
    }
 
@@ -4725,7 +4771,7 @@ OMR::Z::TreeEvaluator::landEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
    if ((TR::Compiler->target.is64Bit() || cg->use64BitRegsOn32Bit()) &&
        ((targetRegister = genericRotateAndInsertHelper(node, cg))
-             || (targetRegister = genericLongAndAsRotateHelper(node, cg)))     )
+             || (targetRegister = TR::TreeEvaluator::tryToReplaceLongAndWithRotateInstruction(node, cg))))
       {
       node->setRegister(targetRegister);
 
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -41,6 +41,7 @@ add_executable(comptest
 	IfxcmpgeReductionTest.cpp
 	VectorTest.cpp
 	CallTest.cpp
+	LongAndAsRotateTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/LongAndAsRotateTest.cpp
+++ b/fvtest/compilertriltest/LongAndAsRotateTest.cpp
@@ -1,0 +1,514 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2017 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
+#include "JitTest.hpp"
+#include "default_compiler.hpp"
+
+// these two arrays hold the value loaded by the iload node.
+int32_t i2lInput []  = {0x0, 0x1, 0x0000FFFF, 0x2, 0x3E, (int32_t)0x80000000, (int32_t)0xFFFF0000,
+                          0x40000000, 0x7C000000, (int32_t)0xFFFFFFFF, 0x0003C000, 0x40000001};
+
+uint32_t iu2lInput [] = {0x0,0x1,0x0000FFFF,0x2,0x3E,0x80000000,0xFFFF0000,0x40000000,0x7C000000,
+                           0xFFFFFFFF,0x0003C000,0x40000001};
+
+// This array consists of all cases where contiguous 1's appear at most once in the binary value
+// ex: 001111000, 1111111, 011110, 1110000, 00001111
+uint64_t lconstInput_zerosAroundOnes [] =               {   0x7000000000000000,
+                                                            0x0,
+                                                            0x1,                //00000..00001
+                                                            0x00000000FFFFFFFF, //00000..11111
+                                                            0x2,                //00000..00010
+                                                            0x3E,               //00...0111110    
+                                                            0x8000000000000000, //10000000..00
+                                                            0xFFFFFFFF00000000, //111111..0000
+                                                            0x4000000000000000, //010000..0000
+                                                            0x7C00000000000000, //0111100..000
+                                                            0xFFFFFFFFFFFFFFFF, //1111111..111
+                                                            0x00000003C0000000, //0..011110..0
+                                                            0x00007FFFFFFFFFFF,
+                                                            0x0000FFFFFFFFFFFF,
+                                                            0x000000000000FFFF,
+                                                            0xFFFFFFFFFFFE0000,
+                                                            0xFFFFFFFFFFFF0000,
+                                                            0xFFFF000000000000
+                                                        };
+
+// this array is for tests where contiguous zeros only appear once in the binary value.
+// ex: 11100001111, 1000001, 0000000
+uint64_t lconstInput_onesAroundZeros [] = {0x0, 0x8000000000000001, 0xE000000000000003, 0xFFFFFFF00FFFFFFF,
+                                             0xFFFFFFFEFFFFFFFF, 0xFFFFFFFE7FFFFFFF, 0xF0FFFFFFFFFFFFFF,
+                                             0xFFFFFFFFFFFFFF0F, 0XFFFF000000000001, 0x8000FFFFFFFFFFFF};
+
+// These input values cannot be handled by a RISBG instruction
+uint64_t lconstInput_invalidPatterns [] = {0x4000000000000001, 0xF0F0F0F0F0F0F0F0, 0x8000000000000002};
+
+template <typename VarType, typename ConstType>
+class LongAndAsRotateTest : public ::testing::TestWithParam<std::tuple<VarType, ConstType, ConstType (*) (VarType, ConstType)>>
+   {
+   public:
+
+   static void SetUpTestCase()
+      {
+      const char *options = "-Xjit:disableTreeSimplification,disableAsyncCompilation,acceptHugeMethods,enableBasicBlockHoisting,"
+                            "omitFramePointer,useIlValidator,paranoidoptcheck";
+      auto initSuccess = initializeJitWithOptions(const_cast<char*>(options));
+
+      ASSERT_TRUE(initSuccess) << "Failed to initialize JIT.";
+      }
+
+   static void TearDownTestCase()
+      {
+      shutdownJit();
+      }
+   };
+
+template <typename VarType, typename ConstType>
+struct LongAndAsRotateParamStruct
+   {
+   VarType variableValue;  //not known at compile time
+   ConstType constantValue;  //known at compile time
+   ConstType (*oracle)(VarType, ConstType);
+   };
+
+template <typename VarType, typename ConstType>
+LongAndAsRotateParamStruct<VarType, ConstType> to_struct(const std::tuple<VarType, ConstType, ConstType (*) (VarType, ConstType)> &param)
+   {
+   LongAndAsRotateParamStruct<VarType, ConstType> s;
+
+   s.variableValue = std::get<0>(param);
+   s.constantValue = std::get<1>(param);
+   s.oracle = std::get<2>(param);
+
+   return s;
+   }
+
+template <typename VarType, typename ConstType>
+static ConstType longAndAsRotateOracle(VarType a, ConstType b)
+   {
+   return ((ConstType)a & b);
+   }
+
+class i2lLongAndAsRotateTest : public LongAndAsRotateTest<int32_t, uint64_t> {};
+
+TEST_P(i2lLongAndAsRotateTest, SimpleTest)
+   {
+   auto param = to_struct(GetParam());
+
+   char inputTrees[512] = {0};
+   std::snprintf(inputTrees, sizeof(inputTrees),
+         "(method return=Int64 args=[Int32]  "
+         " (block                            "
+         "  (lreturn                         "
+         "  (land                            "
+         "      (i2l (iload parm=0))         "
+         "      (lconst %llu)))))            ",
+         param.constantValue
+         );
+
+   auto trees = parseString(inputTrees);
+
+   ASSERT_NOTNULL(trees);
+
+   Tril::DefaultCompiler compiler {trees};
+
+   ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\nInput trees: " << inputTrees;
+
+   auto entry_point = compiler.getEntryPoint<uint64_t (*)(int32_t)>();
+
+   ASSERT_EQ(param.oracle(param.variableValue, param.constantValue), entry_point(param.variableValue));
+   }
+
+TEST_P(i2lLongAndAsRotateTest, iConstTest)
+   {
+   auto param = to_struct(GetParam());
+
+   // this is an arbitrary value that will be used to create
+   // a tree where the optimizer will set the highWordZero
+   // flag on an i2l node. Since treeSimplification is
+   // disabled for these google tests, such a tree is required
+   // for basic block extension to set the highWordZero flag.
+   const int32_t constValue = 0x4321;
+
+   char inputTrees[512] = {0};
+   std::snprintf(inputTrees, sizeof(inputTrees),
+         "(method return=Int64 args=[Int32]  "
+         " (block                            "
+         "  (lreturn                         "
+         "  (land                            " 
+         "      (i2l                         "
+         "         (iand                     "
+         "            (iload parm=0)         "
+         "            (iconst %d)            "
+         "         )                         "
+         "      )                            " 
+         "      (lconst %llu)))))            ",
+         constValue,
+         param.constantValue
+         );
+
+   auto trees = parseString(inputTrees);
+
+   ASSERT_NOTNULL(trees);
+
+   Tril::DefaultCompiler compiler {trees};
+
+   ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\nInput trees: " << inputTrees;
+
+   auto entry_point = compiler.getEntryPoint<uint64_t (*)(int32_t)>();
+
+   ASSERT_EQ(param.oracle((param.variableValue & constValue), param.constantValue), entry_point(param.variableValue)); 
+
+   }
+
+TEST_P(i2lLongAndAsRotateTest, GreaterThanOneRefCount1)
+   {
+   auto param = to_struct(GetParam());
+
+   // this is an arbitrary value that will be used to create
+   // a tree where an i2l node has reference count greater than 1.
+   // The actual value doesn't really matter for constLongValue (as long
+   // as it's not simple enough to get optimized away).
+   const uint64_t constLongValue = 0xFFFFFFFFFFFF0000;
+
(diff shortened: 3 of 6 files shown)


fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+   ASSERT_NOTNULL(trees);
+
+   Tril::JitBuilderCompiler compiler {trees};
+
+   ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+   auto entry_point = compiler.getEntryPoint<uint64_t (*)(int32_t)>();
+
+   ASSERT_EQ(param.oracle((param.variableValue & 0x4321), param.constantValue), entry_point(param.variableValue));
```
mgaudet: Where's the 0x4321 coming from?

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+using LongAndAsRotateParamType = std::tuple<VarType, ConstType, ConstType (*) (VarType, ConstType)>;
+
+template <typename VarType, typename ConstType>
+class LongAndAsRotateTest : public ::testing::TestWithParam<LongAndAsRotateParamType<VarType, ConstType>>
+   {
+   public:
+
+   static void SetUpTestCase()
+      {
+      const char *options = "-Xjit:disableTreeSimplification,disableAsyncCompilation,acceptHugeMethods,enableBasicBlockHoisting,omitFramePointer,useIlValidator,paranoidoptcheck";
```
mgaudet: I wonder if this test case wouldn't be a good candidate to test out the MockOptimization strategies, just to give you some more direct control over the strategy. (See #1661 and [this example](https://github.com/mgaudet/omr/blob/3049c28624e9b989d289687e28a615cba201c502/fvtest/compilertriltest/SimplifierFoldAndTest.cpp#L69-L84))

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+      auto initSuccess = initializeJitWithOptions(const_cast<char*>(options));
```
mgaudet: Can you open an issue to change the signature of `initializeJitWithOptions` so that this `const_cast` becomes un-necessary?

mgaudet:
(As an aside, though I wouldn't have you change on this basis alone: It seems like your commit message is wrapped at ~40 chars... which is weirdly narrow. CONTRIBUTING.md says 72 is a best effort IIRC)

0xdaryl:
I'd appreciate a review from @fjeremic please.

0xdaryl:
I haven't reviewed yet, but @dchopra001 could you fix the formatting in the commit and PR message as @mgaudet suggests and rebase to fix the merge conflict.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "      (lreturn                                       "
+         "         (land                                       "
+         "            (@id \"myNode\")                         "
+         "            (lconst %s)                              "
+         "         )                                           "
+         "      )                                              "
+         "   )                                                 "
+         ")                                                    ",  
+         std::to_string(param.constantValue).c_str()
+ */       //above tree is compacted below:
```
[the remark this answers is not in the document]
dchopra001: I thought the extra white spaces was wasteful :). Let me know if it's better to stick to the original form though.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: I should have defined this more clearly. The 0x4321 is just a positive value that I use for the `iconst` node to get the optimizer to set the `i2l` node's highWordZero flag. 
The comment `//basic block extension will realize that this i2l will have highWordZero` was entered in relation to this. I'll make this more clear.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: I'm guessing you already got to it here?
https://github.com/eclipse/omr/issues/1916

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2466,8 +2466,36 @@ genericRotateAndInsertHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+ * ex. If in the above tree the value of our iload parameter is 
+ * 0x80000000. This will be converted to a long and become
+ * 0xFFFFFFFF80000000. Traditionally we would need to use two
+ * AND instructions to capture the correct result (we need to 
+ * AND the top half and bottom half of the word separately using 
+ * NIHF and NILF instructions). However, with RISBG, we can just choose
+ * to preserve the first 36 bits (lconst's first 36 bits are 1's) and 
+ * accomplish this in one instruction.  
+ *
+ */
```
fjeremic: Can we properly Doxygen this function an its parameters in the header and leave the details Doxygen'd in the object file implementation?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2466,8 +2466,36 @@ genericRotateAndInsertHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+ * This is a helper function that will try to use a rotate instruction
+ * to perform a 'land' operation. For example, if we have the following tree:
+ *
+ * land
+ *    i2l
+ *       iload parm=0
+ *    lconst 0xFFFFFFFFF0000000
+ *
+ * We can avoid the 'land' operation altogether in such a case. We know that
+ * the result of this operation will preserve the first 16 bits of the iload
```
fjeremic: > will preserve the first 16 bits of the iload

This does not sound correct. This example is kind of an edge case. I would focus more on the intention of the use of `RISBG` to avoid two AND instructions, i.e. the case where you are ANDing with something like `0xFF00000000000003` where there is a contiguous sequence of one bits somewhere in the constant and where the alternative to evaluating such a sequence is to either materialize the constant in a register (expensive) or to use two ANDs such as `NIHF` + `NILF` to and the most-significant and least-significant 32-bit portions of the first operand.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
       TR::Node * skipedConversion = NULL;
```
fjeremic: There doesn't seem to be any uses of this variable. Are we able to fold it away?

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: Interesting. I remember wishing there was something like this when I was testing edge cases. Some test cases were particularly designed at the time to test under different optimization settings. Not sure if it's needed now though, as we made some changes such that the testing was no longer needed. But I'll look into this.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
 
-      // Transform land into RISBG as it may be better than discrete pair of ands
+      // Given the right case, transform land into RISBG as it may be better than discrete pair of ands
       if (node->getOpCodeValue() == TR::land && secondChild->getOpCode().isLoadConst())
          {
-         uint64_t value = secondChild->getLongInt();
+         uint64_t longConstValue = secondChild->getLongInt();
 
          if ((firstChild->getOpCodeValue() == TR::i2l || firstChild->getOpCodeValue() == TR::iu2l)
                && firstChild->getReferenceCount()==1 && firstChild->getRegister() == NULL && firstChild->isHighWordZero() )
```
fjeremic: `firstChild->getReferenceCount()==1 && firstChild->getRegister() == NULL` can be replaced by `firstChild->isSingleRefUnevaluated()`. Why does the reference count matter here?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
             {
-            skipedConversion = firstChild;
-            firstChild = firstChild->getFirstChild();
-            value &= 0x00000000FFFFFFFFL;
-            }
+            TR::DebugCounter::incStaticDebugCounter(cg->comp(), TR::DebugCounter::debugCounterName(cg->comp(), "codegen/z/BinaryEvaluator/simplifierShouldHaveFoldedThisTree/%s", cg->comp()->signature()));
```
fjeremic: This is not true. The tree should have been folded by the simplifier if and only if `value & 0x00000000FFFFFFFFL == 0`. We should not blindly increment the counter here without such a check.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+         // If such a case arises where we can figure out at compile time that the 
+         // constant's value is 0, then we must handle that case in the simplifier.
          if (popCnt == (64 - lZeros - tZeros))
             {
-            msBit = lZeros;
+            msBit = lZeros; 
             lsBit = 63 - tZeros;
-            if (firstChild->getReferenceCount()>1 || skipedConversion != NULL   // Better of with non-destructive RISBG as AND?
-                  || lZeros > 31 || tZeros > 31 || (lZeros > 0 && tZeros > 0))  // Better off with a single AND?
+            if (firstChild->getReferenceCount() > 1 // if we use an AND and refCount>1, then we will have to backup the value in another register beforehand, as AND is destructive
```
fjeremic: Can't we clobber evaluate `firstChild` and perform the transformation anyway? Can you further explain why reference count matters here?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+                || lZeros > 31 || tZeros > 31 // this makes sure we are not better off using one of the following AND instructions: NILF,NILH,NILL, NIHF,NIHH, NIHL
+                || (lZeros > 0 && tZeros > 0)) // if our constant at minimum looks like 0111...110, then it's better to use RISBG as this case will require multiple ANDs
```
fjeremic: This does not look correct. What if `lZeros == 1` and `tZeros == 60`. In this case you would carry out the transformation even though you could have generated `NILL` which is a 4-byte instruction as opposed to the 6-byte `RISBG`. This needs more careful consideration.

z/Architecture has all variants of AND immediate instructions for the 16-bit non-overlapping parts of a register. We should only be using `RISBG` in the cases where we cannot handle the AND with a single such 16-bit AND immediate operation. We need to have checks for this.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
                {
-               doIt = true;
+               doTransformation = true;
                }
-            }
-         else if (popCnt == (lOnes + tOnes))
+            }
```
fjeremic: Unnecessary space here.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+                || lZeros > 31 || tZeros > 31 // this makes sure we are not better off using one of the following AND instructions: NILF,NILH,NILL, NIHF,NIHH, NIHL
+                || (lZeros > 0 && tZeros > 0)) // if our constant at minimum looks like 0111...110, then it's better to use RISBG as this case will require multiple ANDs
                {
-               doIt = true;
+               doTransformation = true;
                }
-            }
-         else if (popCnt == (lOnes + tOnes))
+            }                         
+         else if (popCnt == (lOnes + tOnes) && popCnt > 0)
```
fjeremic: In terms of compile time would it be better to first compute `popCnt` and have an early exit if `popCnt == 0`?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2476,56 +2504,75 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
-            }
-         else if (popCnt == (lOnes + tOnes))
+            }                         
+         else if (popCnt == (lOnes + tOnes) && popCnt > 0) 
             {
             msBit = 64 - tOnes;
             lsBit = lOnes - 1;
-            if (firstChild->getReferenceCount()>1 || skipedConversion != NULL // Better of with non-destructive RISBG as AND?
-                  || (lOnes < 32 && tOnes < 32) )                             // Better off with a single AND?
+            if (firstChild->getReferenceCount()>1 // if we use an AND and refCount>1, then we will have to backup the value in another register beforehand, as AND is destructive
```
fjeremic: Same as above. Can we not use clobber evaluate if it will save generating two ANDs?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -2539,7 +2586,12 @@ genericLongAndAsRotateHelper(TR::Node * node, TR::CodeGenerator * cg)
...
-            TR::InstOpCode::Mnemonic opCode = cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zEC12) ? TR::InstOpCode::RISBGN : TR::InstOpCode::RISBG;
+               // if possible then use the instruction that doesn't set the CC as it's faster
+               TR::InstOpCode::Mnemonic opCode = cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_zEC12) ? TR::InstOpCode::RISBGN : TR::InstOpCode::RISBG;
+               
+               // this instruction sets the rotation factor to 0 and sets the zero bit(0x80).
+               // So it's effectively zeroing out every bit except the inclusive range of lsBit to msBit
+               // The bits in that range are preserved as is
                generateRIEInstruction(cg, opCode, node, targetReg, sourceReg, msBit, 0x80 + lsBit, 0);
 
             if (skipedConversion)
```
fjeremic: This looks like dead code and should be removed.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -4725,7 +4776,7 @@ OMR::Z::TreeEvaluator::landEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
    if ((TR::Compiler->target.is64Bit() || cg->use64BitRegsOn32Bit()) &&
        ((targetRegister = genericRotateAndInsertHelper(node, cg))
-             || (targetRegister = genericLongAndAsRotateHelper(node, cg)))     )
+             || (targetRegister = tryToReplaceLongAndWithRotateInstruction(node, cg)))     )
```
fjeremic: Can we fix the spacing here on the last parenthesis?

fvtest/compilertriltest/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ add_executable(comptest
 	ShiftAndRotateTest.cpp
 	SimplifierFoldAndTest.cpp
 	IfxcmpgeReductionTest.cpp
+        LongAndAsRotateTest.cpp
```
fjeremic: I think the CMakeLists.txt files use tabs, hence the spacing inconsistency here. Can we modify the spaces to tabs?

fvtest/compilertriltest/LongAndAsRotateTest.cpp
mgaudet: Would it be useful do you think to be able to directly set the 'highWordZero' flag on a node? If so, maybe open an enhancement issue. It's something milling about in my mental backlog, but I don't have an explicit issue for it yet

fvtest/compilertriltest/LongAndAsRotateTest.cpp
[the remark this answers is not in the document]
mgaudet: But now you have both the compacted and original forms 😄 

I heavily prefer the whitespace laden form as it's more readable. I'd only keep it around unless there's a compelling reason the compacted form is better.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
mgaudet: Actually, that's a separate issue entirely

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: So I spoke to Filip about this and realized I need to rethink the optimization strategy for these tests. In most cases we should not be having any optimizations in the strategy because these trees will generally be folded by the optimizer. I will enable the trees in certain cases when I want the optimizer to detect certain patterns and set appropriate flags. Perhaps we can also revisit this when we have a feature in tril that allows us to set flags (as you suggested in another comment here). I'll open an issue relating to this today.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: Done: https://github.com/eclipse/omr/issues/1946

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: I opened an issue: https://github.com/eclipse/omr/issues/1947

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: So the way I understand is that we would originally skip the evaluation of the `i2l` or `iu2l` node if this entire condition holds true. `firstChild` would be set to firstChild->getFirstChild() and if a RISBG instruction can be generated, then we wouldn't evaluate the `i2l` or `iu2l` node anymore. If the reference count is greater than 1 then we cannot skip the evaluation because we might need the result later on. 

I left this check here because my initial thought was that a case like this should never reach the code generator because the optimizer should have further simplified a tree like this. Is this not true? (I can see why the name of the counter can be misleading. I'll try coming up with a better one).

mgaudet: (_Outsider Opinion_ If it's expected that the optimizer **must** have done a simplification, we need to documented in some fashion to avoid tripping people up when doing investigations, or when disabling some or all of the optimizer. Right now, there's not a great home for this, but we're working on developing a 'pre-codegen validation' which is where you could sock these kinds of facts away. See #1482 )

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: So based on our discussion earlier my understanding was that this type of code should be in the optimizer. The optimizer would/should be able to simplify a tree that looks like this, and not just for the specific case where `value & 0x00000000FFFFFFFFL == 0`. So we would leave a check like this here because it could help identify future opportunities. Is this not true?

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: So the RISBG instruction is non-destructive and we can keep the value in a register as is. So if the reference count > 1 we can generate something like this: `RISBGN R1, R2, 0,129,0` in order to preserve the data in R2. I don't understand clobberEvaluate completely, but from what I understand so far it would keep a copy of the data in R2 in a separate register. Isn't this unnecessary since we can use `RISBGN` to not clobber the value in the first place?

This is what my comment on the line kind of refers to as well. If I'm missing something here then let me know. Otherwise I think I'll add a better comment in the code to document this.

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: So I don't thing that example would work. A value such as that looks like this in binary:
011100000...0. We would need to explicitly zero out everything except those three bits. As per the specification for the 6 AND immediate instructions in Z, 

>"the second operand is ANDed with bits of the first operand, and the result replaces those bits of the first operand. The remainder of the operand remains unchanged."

The 6 AND immediate operands are below:

andimm

The problem is in the fact that the remainder of the operand remains unchanged. So we can't use NILL because that would mean bits 0-47 would remain unchanged when they should be explicitly zeroed out. In fact we would probably need to generate a NIHF and NILF instruction because we need to make changes to the entire top half and bottom half of the register. 

I think a good way to think about this is that we are better off generating a RISBG if there's a 0 in the top half AND the bottom half of the constant (given that we have the leadingOnes or leadingZeros pattern). A 0 means that we need to explicitly zero out a particular bit in the register value if we use these AND Immediate instructions. If the zero exists in both halves then we will need two AND Immediate instructions. 

What you're suggesting does work for the converse case though. If we had something like 1000111111...111 instead, then we could use NILL to zero out those three bits and keep the rest of the bits in the register unchanged. This is exactly why the `if` clause on line 2569 seems a lot less restrictive than the one here because it can handle that case.

Perhaps this also deserves a better comment?

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: Yeah it's better to pull this check up and bail as soon as we calculate that popCnt is 0. It will make more sense as well when someone is reading this code in the future. I'll change this in the next commit.

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: As mentioned in another comment, I don't think we need clobber evaluate. We don't need to store the original value in another register because the RISBG instruction is not destructive. So we can just store the result of the RISBG instruction in another register. I would think a regular evaluate is probably faster in terms of compile time performance as well?

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: > So the way I understand is that we would originally skip the evaluation of the i2l or iu2l node if this entire condition holds true.

The function `tryToReplaceLongAndWithRotateInstruction` as it stands does not skip the evaluation of any children. If a truncation operation should not happen a simplifier case should be added to eliminate it. The function implemented here is responsible for determining if an evaluation of an AND node can benefit from being evaluated with an `RISBG` instruction. The reference count of the children of the AND operation should not matter here. Is the logic here correct?

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: That is only partially correct. Consider the following tree:

```
land
  i2l (highWordZero)
    ...
  lconst 1
```

This tree would enter the above `if` statement and your debug counter would be incremented. However this tree cannot be folded by the simplifier because `value & 0x00000000FFFFFFFFL != 0`, i.e. you have to carry out the AND operation. So the debug counter spots a false positive.

I'm more inclined to say we just remove this entire `if` all together. I don't particularly see much value in it.

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: > So the RISBG instruction is non-destructive and we can keep the value in a register as is. So if the reference count > 1 we can generate something like this: RISBGN R1, R2, 0,129,0 in order to preserve the data in R2. I don't understand clobberEvaluate completely, but from what I understand so far it would keep a copy of the data in R2 in a separate register. Isn't this unnecessary since we can use RISBGN to not clobber the value in the first place?

That's correct. Then there is no need to clobber evaluate in this case at all. I don't see why the reference count would then be a pre-requisite for this optimization?

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: > I would think a regular evaluate is probably faster in terms of compile time performance as well?

Yes we just need a regular evaluation of the first child. I don't think we need to reference count check here as a guard.

compiler/z/codegen/BinaryEvaluator.cpp
dchopra001: If we were to remove the `firstChild->getReferenceCount()>1` condition, the inner if clause will become: `if (lZeros > 31 || tZeros > 31 || (lZeros > 0 && tZeros > 0)`. 

We can still have a case where the outer if clause (  `if (popCnt == (64 - lZeros - tZeros))`) is true, but the new inner one is not (ex. if `lConstValue` is originally `0x0000FFFFFFFFFFFF`). In this scenario we would bail from the helper and end up generating an `NIHH` instruction. If the firstChild's refCount>1, we will need to clobber evaluate this case because we cannot destroy the value. Wouldn't this result in an extra instruction since we need to backup the original value somewhere? 
If we went with a `RISBG` instruction instead, we can accomplish this in one instruction since we don't need to clobber evaluate. The number of registers used would be the same in both cases. This is why I believe if the refCount>1 for the firstChild we should always use RISBG since we save an instruction.

I'm going to test this case out once I get past the compile failures I'm currently seeing. Let me know if I'm missing something here.

fjeremic: > We can still have a case where the outer if clause ( if (popCnt == (64 - lZeros - tZeros))) is true, but the new inner one is not (ex. if lConstValue is originally 0x0000FFFFFFFFFFFF). In this scenario we would bail from the helper and end up generating an NIHH instruction. If the firstChild's refCount>1, we will need to clobber evaluate this case because we cannot destroy the value. Wouldn't this result in an extra instruction since we need to backup the original value somewhere?

Right, but we miss all the opportunities where `lConstValue` is not one where we can `NIHH` for example. The point I'm trying to drive is that `firstChild->getReferenceCount() == 1` does not always imply that using `RISBG` is a bad idea. 

What if `firstChild->getReferenceCount() == 1` and `lConstValue == 0x0000FFFFFFFFFFF0` for example? This function would choose not to generate an `RISBG` and the AND evaluator will have to generate at least 2 ANDs.

fjeremic: > We can still have a case where the outer if clause ( if (popCnt == (64 - lZeros - tZeros))) is true, but the new inner one is not (ex. if lConstValue is originally 0x0000FFFFFFFFFFFF). In this scenario we would bail from the helper and end up generating an NIHH instruction.

The difference between generating `NIHH` and `RISBG` in this case is that `RISBG` is 2 bytes longer. The alternative to not generating `RISBG` in case of `firstChild->getReferenceCount() == 1` and `lConstValue == 0x0000FFFFFFFFFFF0` is that multiple instructions will get generated, which is always worse than the extra 2 bytes of icache we pay.

The best of both worlds is to determine in which case can we *NOT* use one of the `NI**` variants to evaluate the sequence when `firstChild->getReferenceCount() == 1` and generate an `RISBG` in such cases (if possible of course).

dchopra001: Okay so I think we're both trying to say the same thing here.   

>The point I'm trying to drive is that firstChild->getReferenceCount() == 1 does not always imply that using RISBG is a bad idea.

 The `firstChild->getReferenceCount() > 1` check is not a prerequisite. The entire if clause looks like the following in a single line: 
`if (firstChild->getReferenceCount() > 1 || lZeros > 31 || tZeros > 31 || (lZeros > 0 && tZeros > 0))`

I separated these into multiple lines to add comments explaining what each one is doing. I'll clean this up and put a comment above the `if` to make this more clear.

>What if firstChild->getReferenceCount() == 1 and lConstValue == 0x0000FFFFFFFFFFF0

We do handle this case. We would catch it with this `OR` : `(lZeros > 0 && tZeros > 0)`. The reference count is irrelevant here. We will generate RISBG for this case no matter what. 

So if our lConstValue satisfies the pattern that the outer if-clause is trying to detect AND if the refCount>1 then it's ALWAYS better to generate a RISBG instruction. The alternative to a RISBG is multiple instructions (back up the value, then use destructive AND instruction).

>The best of both worlds is to determine in which case can we NOT use one of the NI** variants to evaluate the sequence when firstChild->getReferenceCount() == 1 and generate an RISBG in such cases (if possible of course).

We do essentially do that. The inner-if will only allow a RISBG to be generated if the alternative is to generate multiple `ANDs`. The refCount>1 check is done for the reasons explained above. The remaining checks in this clause can be explained by the following explanation:
>I think a good way to think about this is that we are better off generating a RISBG if there's a 0 in the top half AND the bottom half of the constant (given that we have the leadingOnes or leadingZeros pattern). A 0 means that we need to explicitly zero out a particular bit in the register value if we use these AND Immediate instructions. If the zero exists in both halves then we will need two AND Immediate instructions.

fjeremic: Thanks for the explanation. My mistake was that I was reading the code as:

```
if (firstChild->getReferenceCount())
   {
   ...
   }
```

fvtest/compilertriltest/LongAndAsRotateTest.cpp
dchopra001: So I played around with MockOptimization strategies, and I can make something work using that as well. However I still don't think this is the best long term solution. (Also, the segment of code I was trying to test will be removed in the next commit, but we'll ignore that for now as it can still be important to test this corner case).

The JIT code modified here is in the code generator and ideally I would like to run with noOpt to test the code. However I can't do that right now because some of the nodes require certain flags to be set. So my work around was to disable just enough optimizations so that the optimizer will set the flag I want. The way to do this with MockOptimization would be to run with noOpts except for a couple of test cases (where we would have to add then remove an optimization, or run these tests under a different TestFixture.

The best long term solution IMO here is to not use any optimization and set the flags myself. However the feature to set flags is not ready yet (https://github.com/eclipse/omr/issues/1947). I suggest we leave this as is and revisit these tests once we have the ability to set flags. Neither of the current solutions are particularly "cleaner" and are not the long term solution. So it doesn't really matter which one we go with right now. We will have to revisit this regardless. 

Let me know what you think.

compiler/z/codegen/BinaryEvaluator.cpp
rbanerjee: Kind of a small nit, but do you mind wrapping the lines for the newly added code. I believe we don't have a limit on line length, but some of the line lengths here do seem inconsistent with the convention followed around them.

compiler/z/codegen/BinaryEvaluator.cpp
rbanerjee: Please see the comment above about wrapping really long lines.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+   char inputTrees[512] = {0};
+   std::snprintf(inputTrees, sizeof(inputTrees),
+         "(method return=Int64 args=[Int32]  "
+         " (block                            "
+         "  (lreturn                         "
+         "  (land                            " 
+         "      (i2l (iload parm=0))         "
+         "      (lconst %s)))))              ",
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Sorry but this isn't going to work on our build compiler. 

I believe `std::to_string` only got a constructor overload for `int` starting `C++11`. 

A possible workaround would be to do the following :

```
...

(lconst %lld) ))))) 

...
param.constantValue

```

Please do the same for the rest of the cases where you are using ` std::to_string(param.constantValue).c_str()` 
is used.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "      )                                              "
+         "   )                                                 "
+         ")                                                    ",  
+         std::to_string(param.constantValue).c_str()
+ */       //above tree is compacted below:
+         "(method name=\"AddAndLand\" return=Int64 args=[Int32]"
+         "(block(lstore temp=\"x\"(land(i2l id=\"myNode\"      "
+         "(iload parm=0))(lconst 0xFFFFFFFFFFFF0000)))         "
+         "(lreturn(land(@id \"myNode\")(lconst %s)))))         ",  
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: See comment about the `std::to_string` constructor.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "   )                                                 "
+         ")                                                    ",  
+  */       //above tree is compacted below:
+         "(method name=\"AddAndLand\" return=Int64 args=[Int32]  "
+         "(block(lstore temp=\"x\"(land(i2l                      "
+         "(iand(iload parm=0 id=\"myNode\")(iconst 0x4321)))     "
+         "(lconst 0xFFFFFFFFFFFF0000)))                          "
+         "(lreturn(land(i2l(iand(@id \"myNode\")(iconst 0x1234)))"
+         "(lconst %s)))))                                        ",  
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "  (lreturn                         "
+         "  (land                            " 
+         "      (i2l                         " //basic block extension will realize that this i2l will have highWordZero
+         "         (iand                     "
+         "            (iload parm=0)         "
+         "            (iconst 0x4321)        "
+         "         )                         "
+         "      )                            " 
+         "      (lconst %s)))))              ",
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+   char inputTrees[256] = {0};
+   std::snprintf(inputTrees, sizeof(inputTrees),
+         "(method return=Int64 args=[Int32]  "
+         " (block                            "
+         "  (lreturn                         "
+         "  (land                            " 
+         "      (iu2l (iload parm=0))         "
+         "      (lconst %s)))))              ",
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "      )                                              "
+         "      (lreturn                                       "
+         "         (land                                       "
+         "            (@id \"myNode\")                         "
+         "            (lconst %s)                              "
+         "         )                                           "
+         "      )                                              "
+         "   )                                                 "
+         ")                                                    ",  
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+ */       //above tree is compacted below:
+         "(method name=\"AddAndLand\" return=Int64 args=[Int32]"
+         "(block(lstore temp=\"x\"(land(iu2l id=\"myNode\"      "
+         "(iload parm=0))(lconst 0xFFFFFFFFFFFF0000)))         "
+         "(lreturn(land(@id \"myNode\")(lconst %s)))))         ",  
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "   )                                                 "
+         ")                                                    ",  
+  */       //above tree is compacted below:
+         "(method name=\"AddAndLand\" return=Int64 args=[Int32]  "
+         "(block(lstore temp=\"x\"(land(iu2l                      "
+         "(iand(iload parm=0 id=\"myNode\")(iconst 0x4321)))     "
+         "(lconst 0xFFFFFFFFFFFF0000)))                          "
+         "(lreturn(land(iu2l(iand(@id \"myNode\")(iconst 0x1234)))"
+         "(lconst %s)))))                                        ",  
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+         "  (lreturn                         "
+         "  (land                            " 
+         "      (iu2l                        " //basic block extension will realize that this i2l will have highWordZero
+         "         (iand                     "
+         "            (iload parm=0)         "
+         "            (iconst 0x4321)        "
+         "         )                         "
+         "      )                            " 
+         "      (lconst %s)))))              ",
+         std::to_string(param.constantValue).c_str()
```
rbanerjee: Same as above.

fvtest/compilertriltest/LongAndAsRotateTest.cpp
```diff
@@ -0,0 +1,494 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
+#include "JitTest.hpp"
+#include "jitbuilder_compiler.hpp"
+#include <gtest/gtest.h>
+#include <Jit.hpp>
+
+// these two vectors hold the value loaded by the iload node.
+std::vector<int32_t> i2lInput = {0x0,0x1,0x0000FFFF,0x2,0x3E,(int32_t)0x80000000,(int32_t)0xFFFF0000,0x40000000,0x7C000000,(int32_t)0xFFFFFFFF,0x0003C000,0x40000001};
```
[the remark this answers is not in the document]
dchopra001: How about I just use a c style array then to avoid all confusion?

```int32_t i2lInput [] = {0x0,0x1,0x0000FFFF,0x2,0x3E,(int32_t)0x80000000,
                         (int32_t)0xFFFF0000,0x40000000,0x7C000000,
                         (int32_t)0xFFFFFFFF,0x0003C000,0x40000001};```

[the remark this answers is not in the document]
Leonardo2718: Yep, that might work.

dchopra001:
@fjeremic @mgaudet I've addressed all the comments/changes you requested. PTAL.

@rbanerjee @Leonardo2718 Thank you for the review. I'd missed the function pointer casting and vector initializations when downgrading the tests to be compatible with our compilers. I believe all the other changes you guys requested should be in this commit. PTAL and let me know if I missed something.

Thanks!

mgaudet:
High level inspection of the tests: LGTM.

0xdaryl:
genie-omr build all

0xdaryl:
@rbanerjee : are you OK to merge?  Please approve.

rbanerjee:
Looks Great! (Thanks for addressing the formatting issues as well) 😁

Issue #1903, opened 2017-10-30
Correct ireturn child types in test cases


mgaudet:
Regardless of the outcome of the discussion in #1901 , we should correct the child return types in our test cases to make sure that everything works _the way it is now_. 

As well, we should undo the neutering of the ILValidator I delivered in #1842 .

mgaudet:
(@rbannerjee is currently tackling this)

Issue #1901, opened 2017-10-30
Returning Types smaller than Int32


mgaudet:
### Background 

In #1842 I first implemented a check that ensured all `ireturn` nodes had an `Int32` child type. However, this caused some failures in the compiler tests.

@Leonardo2718 pointed out that `returnOpCode` would return `ireturn` for all types smaller than or equal to Int32. 

After some discussion, and [an insightful comment from](https://github.com/eclipse/omr/pull/1842#issuecomment-339125394) @ymanton I opened #1888 and #1887, to track the creation of `sreturn` and `breturn` as opcodes.

#1842 was modified to allow Int16 and Int8 children to `ireturn`. 

I think I'd like to start some more discussion of this 

# The new thinking: 

On Friday, in pursuit of #1845 @rbanerjee and I were working on a test case for the conversions to Int8 and Int16 types. It was then that we realized that widening is an operation that needs to be sign bit aware: hence the existence of the `b2i` and `bu2i` / `s2i` and `su2i` pairs. The former of the pairs will sign-extend to Int32 and the latter will zero-extend. 

It's likely that `breturn` and `sreturn` would be widening operations, which means the possibility of having to create a `bureturn`and `sureturn`. We already have `iureturn` and `lureturn`, but I don't think that they are considered to be widenings (to be honest, I'm a little surprised to see them, unless `lureturn` is expected to widen an `Int32` child). 

In parallel, @rbanerjee has been investigating failures on s390x. It turns out that these failures are precisely missing widening operations and the s390x ireturn evaluator not correctly handling undersized arguments. The validator not raising the flag on these nodes caused more investigation was necessary. 

# The issues, summarized: 

The most major question is precisely: **What should the canonical way be to return integers smaller than Int32?**. 

I see three answers: 

* Mandate widening operation, and continue to use Int32 as the return type. 
* Mandate the CodeGenerators support widening non-Int32 children on `ireturn` 
* Implement `breturn`,`bureturn`,`sreturn`,`sureturn` 

There are attached subquestions: 

-  If we mandate a conversion to a wider type to be explicit in trees, what would the right return answer be for `OMR::ILOpCode::returnOpCode()`
-  How awkward is the multi-typed children of `ireturn` if we allow it do the widening in codegen -- is this a wart on the IL that isn't worth allowing?
- Can anyone fill in some of the history behind the unsigned returns? @vijaysun-omr / @mstoodle ? 
[...]

PR #1900, opened 2017-10-30
Downgrade minimum compiler version for Compiler Tril Tests

rbanerjee:
This commit lowers the minimum compiler version required to
run the compiler tests built using Tril, to at least as far
as GCC 4.4.6. Essentially not requiring full C++11 support.

Related to PR #1718 

Fixes #1697

PR #1914, opened 2017-11-01
WIP: Fix ShiftAndRotate test failures and remove ireturn Il validation

rbanerjee:
This fixes `ShiftAndRotate` `comptest` failures on s390x. The tests fail due to illformed trees being passed to the Codegenrator where the `ireturn` opcode has an incorrect child type. Also removes existing IL Validation rules that check whether `ireturn` has children with any of the said incorrect types.

Note that as outlined in PR #1900, this is one of the two `comptest` Test Suites where it passes on the `x86` architecture but fails on `s390x`. Though the problem itself is architecture agnostic.
See https://github.com/eclipse/omr/issues/1901 for more on this.

Fixes Issue https://github.com/eclipse/omr/issues/1903

Related to: Issue https://github.com/eclipse/omr/issues/1901, PR #1842 

Signed-off-by: Rwitaban (Ray) Banerjee <rayb@ca.ibm.com>

### DIFF
--- a/compiler/il/ILOpCodeProperties.hpp
+++ b/compiler/il/ILOpCodeProperties.hpp
@@ -657,7 +657,7 @@
    /* .properties4          = */ 0,
    /* .dataType             = */ TR::Int32,
    /* .typeProperties       = */ ILTypeProp::Size_4 | ILTypeProp::Integer,
-   /* .childProperties      = */ ONE_CHILD(ILChildProp::UnspecifiedChildType), // ireturn is used to return all types smaller than Int32
+   /* .childProperties      = */ ONE_CHILD(ILChildProp::UnspecifiedChildType),
    /* .swapChildrenOpCode   = */ TR::BadILOp,
    /* .reverseBranchOpCode  = */ TR::BadILOp,
    /* .booleanCompareOpCode = */ TR::BadILOp,
--- a/compiler/ras/ILValidator.cpp
+++ b/compiler/ras/ILValidator.cpp
@@ -260,12 +260,12 @@ bool TR::ILValidator::treesAreValid(TR::TreeTop *start, TR::TreeTop *stop)
          for (auto i = 0; i < actChildCount; ++i)
             {
             auto childOpcode = node->getChild(i)->getOpCode();
-            const auto expChildType = opcode.expectedChildType(i);
-            const auto actChildType = childOpcode.getDataType().getDataType();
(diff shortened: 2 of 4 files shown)


rbanerjee:
@mgaudet ptal.

mgaudet:
So, b628c69 looks fine. However, I think we had a communication failure on 3838da8. That completely removes all child type validation for `ireturn`. 

What I had been thinking (and failed to communicate, so apologies from me regarding that) is to remove the special case validation for `ireturn`, but in addition _to set the child property of `ireturn` to `TR::Int32`_. 

Now this almost certainly will require changes to the non-tril compiler tests, so it may be necessary to do that in another PR, as I'd really liked to get b628c69 in to correct the functional issues.

Leonardo2718:
> What I had been thinking [...] is to remove the special case validation for ireturn, but in addition to set the child property of ireturn to TR::Int32.

One thing to lookout for is that this may break JitBuilder based JITs (including JitBuilder tests/examples) because JitBuilder uses `TR::ILOpCode::returnOpCode()` internally.

mgaudet:
So, that would be yet-another-required fix -- special casing code in JitBuilder that calls `TR::ILOpCode::returnOpCode()`, though, I suspect this wouldn't be terribly intrusive...

rbanerjee:
@mgaudet Hmm. I assumed we would wait for https://github.com/eclipse/omr/issues/1901 to reach a conclusion before changing the `expected child type` for `ireturn`.

More importantly, Issue https://github.com/eclipse/omr/issues/1903 only goes as far as to say "undo the neutering of the ILValidator I delivered in #1842". The changing of `ireturn`'s child type is a more significant one and I would surely like to have it be part of a separate PR altogether. 

So in light of that, I still think having a standalone commit that just reverts back the checks added in #1842 is worth having. I will shortly open up another PR that changes the `expected child type` for `ireturn` and modifies the surrounding technology accordingly. Does that sound reasonable? (I might be missing something here.)

mgaudet:
Alright -- I'm OK with proceeding down this road.

rbanerjee:
@mgaudet  and I had a chat about this offline and in light of that discussion, I think it's best I abandon this PR (I'm not too fond of the branch name staying the same if this PR becomes just about `comptest` failures : ] ).

The work will be split into multiple PRs the first of which is https://github.com/eclipse/omr/pull/1928 (that just takes care if the `ShiftAndRotate` failures).

PR #1915, opened 2017-11-01
WIP: Introduce OMR Runtime Helper for CRC32

0dvictor:
CRC32 is widely used for many applications; OMR should
provide such an utility so that each downstream project
does not have to implement its own.


0dvictor:
@andrewcraik this PR is about to introduce a helper for CRC32, it requires https://github.com/eclipse/omr/pull/1899. Feel free to use and/or rework it as I'll be away for awhile.

mstoodle:
Hi @0dvictor, what's the plan here, please?

mstoodle:
@0dvictor maybe we should close this one until you're ready to finish it off?

0dvictor:
This PR needs features introduces by https://github.com/eclipse/omr/pull/1899, but https://github.com/eclipse/omr/pull/1899 has not been merged yet although this PR is feature complete and tested. Hence it is hard to give a plan until https://github.com/eclipse/omr/pull/1899 is accepted.

ymanton:
I'm dubious of the value of this, is this really something that belongs here?

There are multiple variants of CRC32 and this one looks like it uses a table pre-calculated for one of them, but it's not specified which. Is there reason to believe someone using OMR would want this variant and in particular this implementation v.s. something from their language runtime if they have one or from zlib or elsewhere? Conversely, if someone doesn't need it do they really want a 1KiB table + associated code in their compiler's runtime lib?

mstoodle:
@0dvictor would you please address @ymanton's concerns? Obviously we need to get #1899 cleaned up first, but in the meantime can we please resolve that this PR is the right direction?

0dvictor:
Having been thinking about this for a while, I think it is a good point that not all project wants this specific implementation of CRC32 even though they may need CRC32. Different implementations take different trade-offs, each downstream project may choose different trade-offs. Should we close this PR and let the downstream projects implement their own?

0xdaryl:
I'm fine with closing this.  The use case I had in mind for this is no longer of interest to me.

0xdaryl:
@0dvictor, if you no longer have a reason to keep this open, please close it.  Thanks.

0dvictor:
Thank you for reminding. I'm closing this pull request.

Issue #1917, opened 2017-11-01
Enable SIMDRegLoad/Store Evaluators on 32bit x86


IBMJimmyk:
The SIMDRegLoadEvaluator and SIMDRegStoreEvaluator were already written and enabled for 64bit x86. They are used for handling the following opcodes:

vbRegLoad, vbRegLoad, viRegLoad, vlRegLoad, vfRegLoad, vdRegLoad
vbRegStore, vsRegStore, viRegStore, vlRegStore, vfRegStore, vdRegStore

Those evaluators need to be added to the 32bit side of TreeEvaluatorTable.hpp as well. Right now it is possible to generate opcodes like vlRegStore under 32 bit x86 which causes problems if the corresponding evaluators are not enabled.

PR #1918, opened 2017-11-01
Added SIMD Evaluators to i386 TreeEvaluatorTable

IBMJimmyk:
TreeEvaluatorTable.hpp was updated to include SIMDRegLoad and SIMDRegStore
evaluators. The evaluators already existed for use with amd64. The
evaluators were already compatible with i386 so they only needed to be
added to TreeEvaluatorTable so they would be used.

Issue: #1917


IBMJimmyk:
@andrewcraik Could you take a look at this? This change is essentially just enabling other code so I think SIMDRegLoadEvaluator and SIMDRegStoreEvaluator inside SIMDTreeEvaluator.cpp should also be looked at to confirm that they really will work with 32bit as well as their current use under 64bit.

0xdaryl:
Looks OK to me.  I'll await @andrewcraik 's opinion though.

andrewcraik:
LGTM from me

0xdaryl:
A genie build is not required as the files affected are only built on x86.

PR #1919, opened 2017-11-01
Enable warnings as errors by default

charliegracie:
Enable the warnings as errors flag for targets that are warning
free.  Other targets should be working on compiling clean for
warnings as errors and as they complete the work they should also
enable this compile flag.

Also enable -Wall for "gnu" compilers and /W3 for the msvc
compilers.  Similarly to warnings as errors add this new option
to targets that compile clean with it enabled.


charliegracie:
FYI @dnakamura and @rwy0717

charliegracie:
genie-omr build plinux

charliegracie:
genie-omr build plinux

dnakamura:
LGTM. I'm not crazy about the verbosity required. ie 
```
if(OMR_WARNINGS_AS_ERRORS)
 	target_compile_options(omrgc PRIVATE ${OMR_WARNING_AS_ERROR_FLAG})
 endif()
 
 if(OMR_ENHANCED_WARNINGS)
 	target_compile_options(omrgc PRIVATE ${OMR_ENHANCED_WARNING_FLAG})
 endif()
```
But I cant see a better solution

PR #1920, opened 2017-11-01
Fix exponential walk in dead trees elimination

jdmpapin:
An earlier attempt to prevent `fixUpTree()` taking exponential time was previously reverted (via commenting) due to a functional issue.

The problem with the original fix is the early return when `(node->getVisitCount() <= oldCompVisitCount)`. This condition will tend to hold not only for nodes evaluated in earlier treetops, but also for nodes that haven't been encountered yet by the current pass of dead trees elimination, resulting in a failure to anchor.

Anchoring should often be unnecessary because of advance analysis by `isSafeToReplaceNode()`, but in some cases the child of treetop is exempted from that analysis, most notably when it has reference count 1. In these cases it's necessary to anchor any previously unevaluated children that might occur below. For example, consider the following trees:

    treetop
      isub with rc=1
        iload x
        iconst -1
    istore x
      ...independent of previous value of x...
    ...
    ireturn
      ==>iload x

In order to eliminate the subtraction, the load must be anchored. With the early return mentioned above, if the visit count of the load is <= `oldCompVisitCount` (which is likely), then `fixUpTree()` will allow the load to swing down past the store.

Dead trees elimination has a particular visit count value reserved to indicate that a node is already evaluated: `visitCount` in `process()`. This can be seen by the way `visitCount` is used in `process()` itself (after setting future use counts), and in `isSafeToReplaceNode()`, `collectSymbolReferencesInNode()`, and `containsNode()`. This commit uses that value to identify nodes that have already been evaluated, which never need to be anchored.

Furthermore, this commit addresses the strange usage pattern of `seenNodes`, which allowed `fixUpTree()` to descend into the children of nodes that it had already anchored.

### DIFF
--- a/compiler/optimizer/DeadTreesElimination.cpp
+++ b/compiler/optimizer/DeadTreesElimination.cpp
@@ -77,23 +77,19 @@ static OMR::TreeInfo *findOrCreateTreeInfo(TR::TreeTop *treeTop, List<OMR::TreeI
    return t;
    }
 
-// temporarily revert this fix
-//static bool fixUpTree(TR::Node *node, TR::TreeTop *treeTop, TR::SparseBitVector &seenNodes, bool &highGlobalIndex, TR::Compilation *comp, vcount_t oldCompVisitCount)
-static bool fixUpTree(TR::Node *node, TR::TreeTop *treeTop, TR::SparseBitVector &seenNodes, bool &highGlobalIndex, TR::Optimization *opt)
+static bool fixUpTree(TR::Node *node, TR::TreeTop *treeTop, TR::NodeChecklist &visited, bool &highGlobalIndex, TR::Optimization *opt, vcount_t evaluatedVisitCount)
    {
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM from me - @vijaysun-omr I think this is in your area of expertise.

vijaysun-omr:
@jdmpapin maybe you can add a brief description of both the original issue (which was just an exponential walk) as well as the prior attempt at a fix that was disabled temporarily. It looks like we needed to check for two kinds of "visited" status here, 1) visited in a prior tree and 2) visited previously when walking the current tree for "fixup". Was the problem with the prior attempt at a fix that it was not visiting some node that it should have ?

jdmpapin:
I've edited the commit message and the PR comment

vijaysun-omr:
Thanks, I understand the case now where the original fix was leading to wrong results. I will merge this now since checks had already passed before and the only reason checks are being run again is that the commit message was changed.

PR #1922, opened 2017-11-01
Improve synchronizedFieldLoadSymbol documentation

fjeremic:
lloadi opcodes are also supported now so this is reflected in the
documentation.

[ci-skip] Documentation only.


0xdaryl:
The CI failure is another instance of the thread death "hang".  This PR is a documentation change so it should have no effect on it at all.  That failure should not prevent this PR from being merged.

Issue #1913, opened 2017-10-31
Remove unsigned loads from IL Opcodes


rbanerjee:
As per the discussion in Issue #1902, the unsigned loads should be removed. 

These are all now deprecated and also there is no functional difference between a signed load and it's unsigned counterpart.

To be more specific this is the list of `unsigned load` Opcodes that are currently part  of  the IL and need to be removed

- [ ] buload 
- [ ] buloadi
- [ ] cload  (the unsigned version of `sload`) 
- [ ] cloadi  (the unsigned version of `sloadi`) 
- [ ] iuload
- [ ] iuloadi 
- [ ] luload
- [ ] luloadi

For each of these one would have to follow the steps outlined below -

- Remove the opcode from the OpCode enum,
- Remove it from the Opcode Properties table.
- Remove from the various RAS data structures
- Remove in the Optimizer
- Remove associated entries from OMRSimplifierTableEnum and the ValuePropagationTable
- Remove the evaluator for X | P | Z | ARM (given that they exist)

0xdaryl:
To make this easier to digest for someone considering taking on this issue in the future, could you articulate precisely which opcodes should be removed in this issue?

rbanerjee:
Updated the description to provide a brief overview for what needs to be done.

andrewcraik:
I will also need to be removed from the various RAS data structures as well as some careful removals from in the optimizer - there are tables in the Simplifier and Value Propagation that are enum index based as well so they will need to be updated too.

rbanerjee:
@andrewcraik  thanks. Updated the description accordingly.

0xdaryl:
I just discovered this issue.  Issue #2657 is a superset.  Closing this one.

PR #1928, opened 2017-11-01
Correct ireturn child types in ShiftAndRotate tests

rbanerjee:
These tests fail on s390x due to the missing conversions before `ireturn`.

The choice of doing a `signed` vs `unsigned` widening is based off of how the oracle is set up right now. From where our IL stands today, `ireturn` has no way of deciphering whether the left child of the `*shr` OpCode (which conventionally would dictate the signedness of the resulting shift) was signed or unsigned to begin with. So the “right choice” here, is to mimic how the oracle interprets the resulting bits while producing its result.

Something to keep mind, in this instance (and for future tests similar to this), the goal is to validate whether the associated Opcode (`*shr` in this case) behaves as expected. `ireturn` and the said conversion before it, are just part of "extracting the computed result”. So the “extraction process” must follow the same convention as the associated Test Oracle for the Test Case to be deemed valid.

(see the discussion below for details)

Fixes Issue: #1903

Related Issues: #1901

### DIFF
--- a/fvtest/compilertriltest/ShiftAndRotateTest.cpp
+++ b/fvtest/compilertriltest/ShiftAndRotateTest.cpp
@@ -180,7 +180,7 @@ TEST_P(Int8ShiftAndRotate, UsingConst) {
     auto param = TRTest::to_struct(GetParam());
 
     char inputTrees[120] = {0};
-    std::snprintf(inputTrees, 120, "(method return=Int8 (block (ireturn (%s (bconst %d) (iconst %d)) )))", param.opcode.c_str(), param.lhs, param.rhs);
+    std::snprintf(inputTrees, 120, "(method return=Int8 (block (ireturn (b2i (%s (bconst %d) (iconst %d)) ))))", param.opcode.c_str(), param.lhs, param.rhs);
     auto trees = parseString(inputTrees);
 
     ASSERT_NOTNULL(trees);
@@ -197,7 +197,7 @@ TEST_P(Int8ShiftAndRotate, UsingLoadParam) {
     auto param = TRTest::to_struct(GetParam());
 
     char inputTrees[120] = {0};
-    std::snprintf(inputTrees, 120, "(method return=Int8 args=[Int8, Int32] (block (ireturn (%s (bload parm=0) (iload parm=1)) )))", param.opcode.c_str());
+    std::snprintf(inputTrees, 120, "(method return=Int8 args=[Int8, Int32] (block (ireturn (b2i (%s (bload parm=0) (iload parm=1)) ))))", param.opcode.c_str());
     auto trees = parseString(inputTrees);
 
     ASSERT_NOTNULL(trees);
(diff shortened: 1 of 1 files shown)


rbanerjee:
@mgaudet  ptal.

mgaudet:
Would you be able to describe your rationale for choosing to widen aligned vs. unsigned, and potentially add that to the commit message?

rbanerjee:
@mgaudet What a great question (more on why this is such a great question later) 😁

Initially my idea of choosing whether to use `unsigned` vs `signed` widening was based off of what I had imagined to be the appropriate choice from a semantic perspective.
Something that quickly fell apart when I tried to justify it, even to myself.

So my original reasoning went something like this:
In the case of `bshr` (`shift byte right arithmetically (child1 >> child2)`) the signed bit gets copied over when shifting, thus preserving the sign of the byte. So it makes sense to let that information be carried over while widening and therefore `b2i` (i.e `signed` widening) made the most sense.
On the other hand, `bushr` (`shift byte right logically (child1 >> child2)`) by definition doesn't preserve the sign bit when shifting `child1`. Resulting in it's return type being always agnostic of the sign of `child1`. So before passing off the result to `ireturn`, `bu2i` (an unsigned widening) seemed the most appropriate choice.

However, there's the case where you shift by `zero` bits in both cases. Now the result of such a shift is trivial as in `child1` stays the same in the case of both `bshr` and `bushr`. So then the question becomes, "How does `ireturn` know whether to `sign` extend or not?".  Well the answer that is - it just doesn't.
(The same applies for `sshr` and `sushr`.)

(Now in the case of c++, the `type_of` `l << r` is dictated by the `type_of` `l`. But in our case, since we are doing the operation on bytes and there's no notion of `breturn`,  we must do an explicit cast going from `bshr / bushr` to `ireturn`. But `ireturn` has no idea of whether `child1` in `bshr` was `signed` or `unsigned` to begin with. In fact from my understanding, where our IL stands today, it doesn't even want to know that. One of the reasons why things like `buload` and other unsigned loads are considered redundant ( See https://github.com/eclipse/omr/issues/1913 ). At the end of the day, the evaluation rule in C++ is still just one placed by the language.) 

So essentially there is no correct answer to whether I should be doing a `b2i` or `bu2i` before invoking `ireturn` in any of these cases.  As anticlimactic as it sounds, the "right thing" to do is just mimic however the `oracle` is set up right now. Because at the end of the day, the goal in this case is to ensure `bshr/bushr` opcodes work as expected. It just so happens to be the case that there is no "one right way" of `extracting that result`. (somewhat relevant to the conversation we are having in https://github.com/eclipse/omr/issues/1901 ).
II'll summarize some of this in the commit description.

But yeah, thanks for asking that. It made me deeply revaluate the expectations that I have of our IL :)  . Also there is a point to be made whether these tests could be made better by abstracting away the aforementioned `extracting the result` part by doing the comparison between the expected and the computed result in the IL itself using something like `ifbucmpge `. I will try to not make this answer any longer and open up a separate issue for that  : ]

rbanerjee:
@mgaudet ptal. Updated the PR and commit descriptions accordingly.

@0xdaryl  Removing the `WIP` tag. Should be good merge once testing finishes.

0xdaryl:
genie-omr build all

PR #1910, opened 2017-10-31
Preliminary Call Support in Tril

mgaudet:
This adds support for creating calls from Tril code to externally defined functions. 

~Currently WIP because the patch as presented is dependent on #1858 being merged first, as it's the PR that contains 0ae512e~

PR #1930, opened 2017-11-02
Add a preliminary language reference for Tril

mgaudet:
[ci skip] documentation only change.

### DIFF
--- a/doc/compiler/tril/TrilLanguageReference.md
+++ b/doc/compiler/tril/TrilLanguageReference.md
@@ -1,0 +1,254 @@
+<!--
+Copyright (c) 2017, 2017 IBM Corp. and others
+
+This program and the accompanying materials are made available under
+the terms of the Eclipse Public License 2.0 which accompanies this
+distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+or the Apache License, Version 2.0 which accompanies this distribution and
+is available at https://www.apache.org/licenses/LICENSE-2.0.
+
+This Source Code may also be made available under the following
+Secondary Licenses when the conditions for such availability set
+forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+General Public License, version 2 with the GNU Classpath 
+Exception [1] and GNU General Public License, version 2 with the
+OpenJDK Assembly Exception [2].
+
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+-->
+
+# Tril Language Reference
+
+In many ways, the Tril language mirrors the Testarossa Intermediate Language,
+_Trees_. 
+
+This language reference intends to document the places where special behaviour
+has been created in Tril to make writing test cases easier. 
+
+# Basics
+
+Tril is built out of two main components: 
+
+1. A very permissive S-expression parser.
+2. An IlGenerator that walks the AST produced by said parser, and generates
+   Trees. 
+
+# Syntax
+
+Tril represents Trees using S-expressions, with key-value properties to augment
+a particular node with more information. 
+
+* `(x (y) (z))` represents some `x` with `y` and `z` as children.
+* `(x foo=bar baz=braz (y) (z)` is the same tree as above, but where the node
+  for `x` is augmented with two properties with values. 
+
+  Anything may be provided as a property for a node, however, unknown
+  properties will be silently ignored.
+
+## Literals
+
+* Signed integer literals can be input directly, and will always be treated internally
+  as a 64-bit integer.  
+* Hexadecimal integer literals are supported as well, and are treated as
+  64-bit integers. Use a leading `0x`. 
+* Floating point numbers can be input by including a decimal point.  
+
+## Comments 
+
+`;` starts a comment in Tril, and the comment proceeds to the end of a line. 
+
+## Special Forms: 
+
+The key to Tril is the unlocking of extra behaviour using 'special forms', or
+atoms that represent certain behaviours in Testarossa. 
+
+### Types
+
+The Testarossa IL, and therefore Tril, deals with the following data types. 
+
+* `Int8`
+* `Int16`
+* `Int32`
+* `Int64`
+* `Address`
+* `Float`
+* `Double`
+* `VectorInt8`
+* `VectorInt16`
+* `VectorInt32`
+* `VectorInt64`
+* `VectorFloat`
+* `VectorDouble`
+* `NoType`
+
+Signedness is a property of operations, not the data itself.
+
+### Methods
+
+A whole method in Tril is wrapped in a `method` expression. 
+
+#### Properties 
+
+* `return` _Mandatory_ is the return type of the method. 
+* `args` _Optional_ is a square-bracketed list of arguments representing the
+  types of the parameters, and the order in which they appear. 
+
+  eg. `args=[Int32,Double,Float]`
+    
+### Block
+
+In the Testarossa IL, all trees exist within the context of a basic block. In 
+Tril, blocks must be children of a `method`, or they will not be recognized.
+
+#### Properties
+
+* `name` _Optional_ Blocks can be named in order to target them with branches. 
+
+### Stores and loads
+
+Currently the creation of new symbol references in Tril is not supported.
+However, store and load opcodes provide two options for where the store / load
+can come from, which can be specified with one of the following two properties: 
+
+* `parm` can specify the parameter from which a load can come. Parameters are 
+  counted from zero.  eg. `(iload parm=2)` loads the third parameter from the
+  signature. 
+* `temp` Stores to temps create new temporary variables that can later be
+  referenced. eg. `(istore temp="x" ...)`
+
+### Branches 
+
+Branches evaluate a condition then either jump to a target, or falls through. 
+
+#### Properties
+
+* `target` _Mandatory_ The `name` of the block a successful computation of the
+  condition will continue executing in. 
+* `fallthrough` _Optional_ In some circumstances it is necessary to explicitly
+  list the `name` of the fallthrough block of a branch to ensure the desired CFG is
+  constructed. In addition to a block `name`, there are two special values that
+  may be used here: 
+   * `@none` indicates no fallthrough edge need be generated: this can be the
+     case for certain opcodes. 
+   * `@exit` falls through to the exit block of the CFG. 
+
+#### Example
+
+```
+(method ...
+   (block fallthrough="sub" name="start"
+      (ificmpge target="add"                 
+        ...) )
+   (block name="sub"
+      (ireturn
+         ...) )
+   (block name="add"
+      (ireturn                               
+        ... ) )
+```
+   
+### Commoning 
+
+Nodes can be given identifiers to allow creating commoned subtrees. To common a
+reference, on first computation annotate the node with the `id="<some name>"`
+property. Then, where commoning is desired, replace the node with `(@id "<some
+name>")` 
+
+```
+(imul id="multiplyResult"
+    ...) 
+(iadd
+  (@id "multiplyResult")
+  (@id "multiplyResult")
+```
+
+### Vectors 
+
+Vector support in Tril is very preliminary, but present. One of the most
+important pieces is the ability to annotate indirect vector loads and stores
+with the data type of the load. 
+
+This is required because the vector support in Testarossa infers the vector
+types from certain roots, as opposed to expressing the data types in the
+opcodes directly like other opcodes in Testarossa. 
+
+    (vloadi type=VectorDouble (aload parm=0))
+    
+####Properties
+
+* `type` _Mandatory_ The data type being loaded or stored. 
+
+
+
+### Calls
+
+Call opcodes are supported with in Tril by providing extra annotation of the
+argument types and the function pointer to be invoked.
+
+    (icall address=0x1234 args=[Int32] ...) 
+
+####Properties
+
+* `address` _Mandatory_ The address of the function to be invoked by the call opcode
+* `args` _Optional_ The list of argument types passed to the call. 
+
+
+# Examples 
+
+## Incordec 
+
+Checked in as `fvtest/tril/examples/incordec/incordec.tril`
+
+```
+; This simple method takes as argument a pointer to a 32-bit integer. If the
+; integer pointed to by the argument has a negative value, then the value
+; decremented by 1 is returned. Otherwise (value is positive), the value
+; incremented by 1 is returned.
+;
+; An equivalent C implementation:
+;
+; int incordec(int* parm0) {
+;    if (*parm0 < 0) {
+;     return *parm0 - (int)1.0;
+;    } else { 
+;      int x = 1;
+;      return *parm0 + x;
+;    }
+; }
+
+(method name="incordec" return="Int32" args=["Address"]
+   (block fallthrough="sub" name="start"     ; start:
+      (ificmpge target="add"                 ; if (*parm0 >= 0) goto add;
+         (iloadi offset=0 (aload parm=0) )
+         (iconst 0) ) )
+   (block name="sub"                         ; sub:
+      (ireturn                               ; return *parm0 - 1;
+         (isub
+            (iloadi offset=0 (aload parm=0) )
+            (d2i (dconst 1.0) ) ) ) )
+   (block name="add"                         ; add:
+      (istore temp="x" (iconst 1))           ; int x = 1;
+      (ireturn                               ; return *parm0 + x;
+         (iadd
+            (iloadi offset=0 (aload parm=0) )
+            (iload temp="x") ) ) ) )
+```
+
+## Vector Support: 
+
+Vector addition of two input double vectors into a third output double vector.
+Hooked up as a test case in `fvtest/compilertriltest/VectorTest.cpp`.
+
+```
+(method return= NoType args=[Address,Address,Address]          
+  (block                                                       
+     (vstorei type=VectorDouble offset=0                       
+         (aload parm=0)                                        
+            (vadd                                              
+                 (vloadi type=VectorDouble (aload parm=1))     
+                 (vloadi type=VectorDouble (aload parm=2))))   
+     (return)))                                                
+```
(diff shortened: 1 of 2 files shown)


doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
+# Tril Language Reference
+
+In may ways, the Tril language mirrors the Testarossa Intermediate Language,
+_Trees_. 
+
+This language reference intends to document mostly the places where special
```
0xdaryl: Do you need "mostly" ?

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+# Syntax
+
+Tril represents Trees using S-expressions, with key-value properties to augment
+a particular node with more information. 
+
+* `(x (y) (z))` represents some `x` with `y` and `z` as children.
+* `(x foo=bar baz=braz (y) (z)` is the same tree as above, but where the node
+  for `x` is augmented with two properties with values. 
+
+  Anything may be provided as a property for a node, however, unknown properties will be silently ignored.
```
0xdaryl: I think it would be useful to be more specific on the syntax requirements for literals and constants.  For example, can strings be quoted?  How do you specify floating point constants (10.0f)?  Can you use hex constants "0x1234" ?  etc.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+    (vloadi type=VectorDouble (aload parm=0))
+    
+####Properties
+
+* `type` _Mandatory_ The data type being loaded or stored. 
+
+
+
+### Calls
+_(Pending delivery as part of [#1910](https://github.com/eclipse/omr/pull/1910))_
```
0xdaryl: Either make sure the PR for #1910 removes this, or remove all this text and make sure PR #1910 adds it.

mgaudet:
A note to myself when I circle back to addressing feedback on this PR: Should also add a copyright header to the file.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+This language reference intends to document mostly the places where special
+behaviour has been created in Tril to make writing test cases easier. 
+
+# Basics
+
+Tril is built out of two main components: 
+
+1. A very permissive S-expression parser.
+2. An IlGenerator that walks the AST produced by said parser, and generates
+   trees.
```
Leonardo2718: Capitalize "trees" for consistency?

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+* `args` _Optional_ is a square-bracketed list of arguments representing the
+  types of the parameters, and the order in which they appear. 
+
+  eg. `args=[Int32,Double,Float]`
+    
+
+
+### Block
+
+In the Testarossa IL, all trees exist within the context of a basic block.
```
Leonardo2718: Maybe mention that blocks _must_ children of a method (otherwise, they will not be recognized as blocks).

Leonardo2718: Also maybe mention that all child expressions of a block are interpreted as opcodes.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+* `name` _Optional_ Blocks can be named in order to target them with branches. 
+
+### Stores and loads
+
+Currently the creation of new symbol references in Tril is not supported.
+However, store and load opcodes provide two options for where the store / load
+can come from, which can be specified with one of the following two proprties: 
+
+* `parm` can specify the parameter from which a load can come. eg. `(iload
+  parm=2)` loads the third parameter from the signature.
```
Leonardo2718: Maybe mention explicitly that parameters are counted from 0. Otherwise, this looks like it might be a typo (at least to someone who is not already familiar with Trees).

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+can come from, which can be specified with one of the following two proprties: 
+
+* `parm` can specify the parameter from which a load can come. eg. `(iload
+  parm=2)` loads the third parameter from the signature. 
+* `temp` Stores to temps create new temporary variables that can later be
+  referenced. eg. `(istore temp="x" ...)`
+
+### Branches 
+
+Branches evaluate a condition then either jump to a target, or falls through.
```
Leonardo2718: Not sure if documenting the behavior of opcodes is appropriate in this document. It really should be part of a document on Trees.

Although, I can see how something short like this can be helpful for readability and clarity.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+  referenced. eg. `(istore temp="x" ...)`
+
+### Branches 
+
+Branches evaluate a condition then either jump to a target, or falls through. 
+
+#### Properties
+
+* `target` _Mandatory_ This is the block a successful computation of the
+  condition will continue executing in.
```
Leonardo2718: I think it would be helpful to explicitly mention that the target is specified by the `name` of the target block.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+   (block name="add"
+      (ireturn                               
+        ... ) )
+```
+   
+### Commoning 
+
+Nodes can be given identifiers to allow creating commoned subtrees. To common a
+reference, on first compuation annotate the node with the `id` property. Then,
+where commoning is desired, replace the node with `(@id "...")`
```
Leonardo2718: Maybe use something like `id="<some name>"` and then `(@id "<some name>")`?

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,194 @@
...
+* `address` _Mandatory_ The address of the function to be invoked by the call opcode
+* `args` _Optional_ The list of argument types passed to the call. 
+
+
+# Examples 
+
+## Incordec 
+
+Checked in as `fvtest/tril/examples/incordec/incordec.tril`
+
```
Leonardo2718: Might be worth mentioning somewhere that line comments start with a `;`.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: #1910 can't add the text, because the file doesn't exist yet, nor can it remove it. I don't think there's a sensible way to keep these two independent other than seeing what gets merged first, and compensating from that.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: I think I'd like to keep this just because it increases the readability of the document.

mgaudet:
All review comments addressed, except for those I've explicitly called above.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: done.

mgaudet:
(Curious -- AppVeyor appears not to have respected `[ci skip]` on this force push...)

doc/compiler/tril/TrilLanguageReference.md
mgaudet: 1910 was merged, so this has now been removed.

mgaudet:
Rebased and corrected. I think this is ready for merge.

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+Branches evaluate a condition then either jump to a target, or falls through. 
+
+#### Properties
+
+* `target` _Mandatory_ The `name` of the block a successful computation of the
+  condition will continue executing in. 
+* `fallthrough` _Optional_ In some circumstances it is necessary to explicitly
+  list the `name` of the fallthrough block of a branch to ensure the desired CFG is
+  constructed. In addition to a block `name`, there are two special values that
```
0xdaryl: "...two optional special values..."   (add optional)

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+         ...) )
+   (block name="add"
+      (ireturn                               
+        ... ) )
+```
+   
+### Commoning 
+
+Nodes can be given identifiers to allow creating commoned subtrees. To common a
+reference, on first compuation annotate the node with the `id="<some name>"` property. Then,
```
0xdaryl: sp: computation

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+```
+; This simple method takes as argument a pointer to a 32-bit integer. If the
+; integer pointed to by the argument has a negative value, then the value
+; decremented by 1 is returned. Otherwise (value is positive), the value
+; incremented by 1 is returned.
+;
+; An equivalent C implementation:
+;
+; int incordec(int* parm0) {
+;    if (*value < 0) return *parm - 1;
```
0xdaryl: Should this be `*parm0 < 0` instead?  (there is no `value` defined)

0xdaryl: parm -> parm0

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+;    else return *parm + 1;
```
0xdaryl: parm -> parm0

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+         (isub
+            (iloadi offset=0 (aload parm=0) )
+            (d2i (dconst 1.0) ) ) ) )
+   (block name="add"                         ; add:
+      (istore temp="x" (iconst 1))           ; int x = 1;
+      (ireturn                               ; return *parm0 + x;
+         (iadd
+            (iloadi offset=0 (aload parm=0) )
+            (iload temp="x") ) ) ) )
+```
```
0xdaryl: While functionally equivalent to the C snippet above, I think its worth pointing out that you exercised your creative license in order to demonstrate some additional features of Tril in order to achieve that equivalent functionality (like the conversion of the dconst and the store/load from a temporary).

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+[1] https://www.gnu.org/software/classpath/license.html
+[2] http://openjdk.java.net/legal/assembly-exception.html
+
+SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+-->
+
+# Tril Language Reference
+
+In may ways, the Tril language mirrors the Testarossa Intermediate Language,
+_Trees_.
```
0xdaryl: Is _Trees_ the official name of the Testarossa IL?  I didn't think it had one other than "Testarossa IL".  People refer to "the trees" when talking about the IL, but I don't think that made its way into an official name.  Also, you don't refer to Testarossa IL as "Trees" subsequently in this document.

I'd say purge the ", _Trees_." from this sentence.

cc'ing Testarossa greybeards for opinion:   @mstoodle @vijaysun-omr @andrewcraik

doc/compiler/tril/TrilLanguageReference.md
andrewcraik: Does this need to consider table lookups and switches? what about goto - ie unconditional branch?

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+  for `x` is augmented with two properties with values. 
+
+  Anything may be provided as a property for a node, however, unknown
+  properties will be silently ignored.
+
+## Literals
+
+* Integer literals can be input directly, and will always be treated internally
+  as a 64-bit integer.  
+* Hexadecimal floating point literals are supported as well, and are treated as
```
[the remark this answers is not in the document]
Leonardo2718: That should be "integers".

doc/compiler/tril/TrilLanguageReference.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Special Forms: 
+
+The key to Tril is the unlocking of extra behaviour using 'special forms', or
+atoms that represent certain behaviours in Testarossa. 
+
+### Methods
+
+A whole method in Tril is wrapped in a `method` expression. 
+
```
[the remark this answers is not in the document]
Leonardo2718: Maybe types should be specified under their own header, since they can be used for more than just method definitions.

doc/compiler/tril/TrilLanguageReference.md
Leonardo2718: Unconditional branches work the same way, so that phrase should be generalized to include these as well

Based on my weak understanding of table lookups and switches, I think they should also work, but I've never tested them myself.

doc/compiler/tril/TrilLanguageReference.md
Leonardo2718: Not sure if the "optional" makes sense. In my mind, these are values that carry a special meaning; they assign names to targets that are otherwise unnamed. They are not elements that can be "optionally" added. Maybe an example using these would help clarify this: `fallthrough="foo"` vs `fallthrough=@exit`.

doc/compiler/tril/TrilLanguageReference.md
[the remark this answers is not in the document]
0xdaryl: Sorry for not being clear.  That's what I meant.

doc/compiler/tril/TrilLanguageReference.md
0xdaryl: Yeah, on second thought "optional" is not the right word here and makes more sense without it.  +1 to an example though.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: I think I'd keep switches as undocumented until we have a test case that's verifying it all works the way it ought to :)

doc/compiler/tril/TrilLanguageReference.md
andrewcraik: I have heard Trees as a proper name used for the IL with IBM in the past and I have seen documents that have used that name. That being said it isn't particularly distinctive. I'm on the fence about using here, but having a proper name for the IL would seem to be a good idea since what was Testarossa is now split over the OMR and OpenJ9 projects and having a name in the context of OMR that is inherited by others building on this technology would seem like a good idea to make documentation and discussion more clear going forward.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: I actually don't yet have enough experience to write out an example for `fallthrough=@exit`; 

    (goto target="unconditionalTarget" fallthrough=@none) 

makes sense to me -- not sure though, is that a mandatory fallthrough on an unconditional branch or not?

doc/compiler/tril/TrilLanguageReference.md
mgaudet: https://github.com/eclipse/omr/issues/2000

doc/compiler/tril/TrilLanguageReference.md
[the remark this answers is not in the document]
mgaudet: Done.

doc/compiler/tril/TrilLanguageReference.md
mgaudet: Done.

mgaudet:
So everything except clarifying branch fallthrough semantics is done. I think I am biased towards leaving some things un-documented until we have test cases or examples prepared that show off particular elements.

doc/compiler/tril/TrilLanguageReference.md
andrewcraik: A throw with a fallthrough of @exit would make sense with a target being some catch block, but not sure if that is implemented yet or if the @exit in that case would be implicit?

mgaudet: I don't think we've got any support for `throw` -- I mean, if it worked due to the cleverness of the way that @Leonardo2718 cooked up Tril, I wouldn't be surprised, but we definitely haven't tried it 🤔

doc/compiler/tril/TrilLanguageReference.md
mgaudet: There's a compromise point here: Could rephrase as 

> In may ways, the Tril language mirrors the Testarossa Intermediate Language,
colloquially known as _Trees_.

doc/compiler/tril/TrilLanguageReference.md
Leonardo2718: So, it used to be that specifying `fallthrough` was mandatory in Tril. This created the need to have a way of expressing "no fall-through" (`@none`) and "fall-through to the exit block" (`@exit`).

However, it was later decided that in most cases, it is easy enough to figure out what the fall-through path should be (the next block in TreeTop order). So, the `fallthrough` property was made optional. Now, Tril will implicitly generate fall-through edges as needed by walking the AST, *unless an explicit `fallthrough` is specified*. Explicitly specifying the expected (correct) fall-through target (as is done in @mgaudet's example above) is therefore redundant.

Today, `fallthrough` is used to *override* the default fall-through. So far, this has been most useful for explicitly generating incorrect CFGs for testing purposes.

re: throw, I don't know enough about how exceptions work in Testarossa to say whether or not they will work. Quickly scanning the code base, it doesn't look like Tril will correctly handle throws. At least, I don't think it will generate the correct CFG (though it might generate the correct Trees; cleverness can only go so far 😛 ).

In any case, the documentation and examples should make the context under which `@none` and `@exit` are/should be used clear.

mgaudet:
I've updated the branch descriptor. Hopefully this scans better to @Leonardo2718

mgaudet:
@0xdaryl I think this has addressed all feedback.

PR #1940, opened 2017-11-02
CMake: Allow out-of-tree hookgen

dnakamura:
Modify the hookgen handling to allow generating files outside of the
source tree. This is not trivial since the output files are specified
as a relative path in the .hdf file (which are relative to the .hdf).

To work arround this omr_add_hookgen, reads in the .hdf file, identifies
the output header names, rewrites them as relative path (relative to
the current binary dir), and then outputs the modified hdf into
CMAKE_CURRENT_BINARY_DIR, and runs hookgen on that. This allows us to
generate the hookgen files in the build tree, without making any changes
which would likely break the existing autconf/makefile builds.

Signed-off-by: Devin Nakamura <dnakamura@ca.ibm.com>

### DIFF
--- a/cmake/modules/OmrHookgen.cmake
+++ b/cmake/modules/OmrHookgen.cmake
@@ -23,21 +23,61 @@ if(OMR_HOOKGEN_)
 	return()
 endif()
 set(OMR_HOOKGEN_ 1)
+include(CMakeParseArguments)
 
 # Process an input hookgen file to generate output source files.
-# Usage: omr_add_hookgen(<input> <output>...)
-# TODO: currently output in source tree, should be in build tree
-# TODO: Dependecy checking is broken, since it checks for output in build tree rather than src
-function(omr_add_hookgen input)
-	get_filename_component(input_dir "${input}" DIRECTORY)
-	add_custom_command(
-		OUTPUT ${ARGN}
-		COMMAND hookgen ${CMAKE_CURRENT_SOURCE_DIR}/${input}
-		DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${input}"
-		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${input_dir}
-	)
-endfunction(omr_add_hookgen)
+# Usage: omr_add_hookgen(INPUT <input> [PRIVATE_DIR <private_header_dir>] [PUBLIC_DIR <public_header_dir>]
+# public headers are written to public_header_dir (defaults to current binary dir)
+# private headers are written to private_heade_dir (defautls to current binary dir)
+function(omr_add_hookgen)
+	cmake_parse_arguments(OPT "" "INPUT;PRIVATE_DIR;PUBLIC_DIR" "TARGETS" ${ARGN})
 
-macro(add_hookgen)
-	omr_add_hookgen(${ARGN})
-endmacro(add_hookgen)
+	if(NOT OPT_INPUT)
+		message(FATAL_ERROR "value for INPUT must be specified")
+	endif()
+
+	#if private_dir or public_dir not specified, default to CMAKE_CURRENT_BINARY_DIR
+	if(NOT OPT_PRIVATE_DIR)
+		set(OPT_PRIVATE_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+	endif()
+	if(NOT OPT_PUBLIC_DIR)
+		set(OPT_PUBLIC_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+	endif()
+
+	get_filename_component(input_filename "${OPT_INPUT}" NAME)
+
+	file(READ "${OPT_INPUT}" hook_def)
+	string(REGEX MATCH "<privateHeader>[^<]*" private_header "${hook_def}")
+	string(REGEX MATCH "<publicHeader>[^<]*" public_header "${hook_def}")
+	if((NOT public_header) OR (NOT private_header))
+		messeage(FATAL_ERROR "Header extraction failed for ${input}")
+	endif()
+
+	string(REPLACE "<privateHeader>" "" private_header "${private_header}")
+	string(STRIP "${private_header}" private_header)
+	get_filename_component(private_header "${private_header}" NAME)
+
+	string(REPLACE "<publicHeader>" "" public_header "${public_header}")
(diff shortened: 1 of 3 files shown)


dnakamura:
I believe with this change we no longer output  anything into the source tree, which means you *should* be able to run multiple builds off of the same source tree
/cc @rwy0717

charliegracie:
genie-omr build plinux

charliegracie:
@rwy0717 any opinions?

cmake/modules/OmrHookgen.cmake
```diff
@@ -23,21 +23,61 @@ if(OMR_HOOKGEN_)
...
 
 # Process an input hookgen file to generate output source files.
-# Usage: omr_add_hookgen(<input> <output>...)
-# TODO: currently output in source tree, should be in build tree
-# TODO: Dependecy checking is broken, since it checks for output in build tree rather than src
-function(omr_add_hookgen input)
-	get_filename_component(input_dir "${input}" DIRECTORY)
+# Usage: omr_add_hookgen(INPUT <input> [PRIVATE_DIR <private_header_dir>] [PUBLIC_DIR <public_header_dir>]
+# public headers are written to public_header_dir (defaults to current binary dir)
+# private headers are written to private_heade_dir (defautls to current binary dir)
```
rwy0717: typo: private_heade_dir and defuatls

cmake/modules/OmrHookgen.cmake
```diff
@@ -23,21 +23,61 @@ if(OMR_HOOKGEN_)
...
+
+	string(REPLACE "<privateHeader>" "" private_header "${private_header}")
+	string(STRIP "${private_header}" private_header)
+	get_filename_component(private_header "${private_header}" NAME)
+
+	string(REPLACE "<publicHeader>" "" public_header "${public_header}")
+	string(STRIP "${public_header}" public_header)
+	get_filename_component(public_header "${public_header}" NAME)
+
+	# Hookgen does not accept absoulte paths, they must be relative to the input file
```
rwy0717: typo: absoulte

cmake/modules/OmrHookgen.cmake
```diff
@@ -23,21 +23,61 @@ if(OMR_HOOKGEN_)
...
+		set(OPT_PRIVATE_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+	endif()
+	if(NOT OPT_PUBLIC_DIR)
+		set(OPT_PUBLIC_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+	endif()
+
+	get_filename_component(input_filename "${OPT_INPUT}" NAME)
+
+	file(READ "${OPT_INPUT}" hook_def)
+	string(REGEX MATCH "<privateHeader>[^<]*" private_header "${hook_def}")
```
rwy0717: I think this would be clearer (and shorter) as: 
```cmake
string(REGEX REPLACE "<privateHeader>(.*?)</privateHeader>" "\\1" private_header "${hook_def}")
```

cmake/modules/OmrHookgen.cmake
```diff
@@ -23,21 +23,61 @@ if(OMR_HOOKGEN_)
...
+	file(RELATIVE_PATH private_path "${CMAKE_CURRENT_BINARY_DIR}" "${OPT_PRIVATE_DIR}/${private_header}")
+	file(RELATIVE_PATH public_path "${CMAKE_CURRENT_BINARY_DIR}" "${OPT_PUBLIC_DIR}/${public_header}")
+
+	string(REGEX REPLACE "<privateHeader>[^<]*" "<privateHeader>${private_path}" hook_def "${hook_def}")
+	string(REGEX REPLACE "<publicHeader>[^<]*" "<publicHeader>${public_path}" hook_def "${hook_def}")
+
+	file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${input_filename}" "${hook_def}")
+
+	# Make cmake step depend on .hdf file, since we need to regen the modified .hdf
+	set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${OPT_INPUT}")
```
rwy0717: nice, I would have forgotten to do this 👍

dnakamura:
@rwy0717  I'm not crazy about it either. However its probably the best way without breaking the autoconf build. I agree that once we have killed it off we should come back and revisit this

charliegracie:
@dnakamura once you fix up the typos, etc. this PR looks ready to go

dnakamura:
done

cmake/modules/OmrHookgen.cmake
```diff
@@ -26,9 +26,9 @@ set(OMR_HOOKGEN_ 1)
 include(CMakeParseArguments)
 
 # Process an input hookgen file to generate output source files.
-# Usage: omr_add_hookgen(INPUT <input> [PRIVATE_DIR <private_header_dir>] [PUBLIC_DIR <public_header_dir>]
+# Usage: (INPUT <input> [PRIVATE_DIR <private_header_dir>] [PUBLIC_DIR <public_header_dir>]
 # public headers are written to public_header_dir (defaults to current binary dir)
-# private headers are written to private_heade_dir (defautls to current binary dir)
+# private headers are written to private_heade_dir (defaults to current binary dir)
```
charliegracie: Sorry one more type private_heade_dir.

rwy0717:
You're not doing anything with the `TARGETS ...` arguments, is that intentional?

dnakamura:
@rwy0717  was trying to figure out where you were seeing that. Finally found it. Was a remnant from an old prototype

gc/CMakeLists.txt
```diff
@@ -20,25 +20,26 @@
...
+	PUBLIC_DIR "${omr_BINARY_DIR}"
+	TARGETS omrgc_hookgen
 )
-add_hookgen(base/omrmmprivate.hdf
-	${CMAKE_CURRENT_SOURCE_DIR}/base/mmprivatehook.h
-	${CMAKE_CURRENT_SOURCE_DIR}/base/mmprivatehook_internal.h
+
+omr_add_hookgen(
+	INPUT base/omrmmprivate.hdf
+	TARGETS omrgc_hookgen
```
rwy0717: But wait, you're still using `TARGETS` in the different call sites 😜

PR #1944, opened 2017-11-03
Replace missed nullptr with NULL

mgaudet:


charliegracie:
genie-omr build plinux

PR #1959, opened 2017-11-06
Fixes to ValuePropagation

liqunl:
This commit contains the following changes
1) Early return in VPHandlers if addKnownObjectConstraints has created
   a constraint. This change will save compilation time in each pass
   of VP.
2) Create KnownObject constraints for newly folded references so that
   the folding can benefit current pass of VP.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -1419,15 +1419,22 @@ static const char *getFieldSignature(OMR::ValuePropagation *vp, TR::Node *node, 
    return NULL;
    }
 
-static void addKnownObjectConstraints(OMR::ValuePropagation *vp, TR::Node *node)
+/**
+ * Try to add constraint to known objects or constant strings.
+ * @parm vp The VP object.
+ * @parm node The node whose value might be a known object or a constant string.
+ *
+ * @return True if the constraint is created, otherwise false.
+ */
+static bool addKnownObjectConstraints(OMR::ValuePropagation *vp, TR::Node *node)
    {
    TR::KnownObjectTable *knot = vp->comp()->getKnownObjectTable();
    if (!knot)
-      return;
+      return false;
 
    TR::SymbolReference *symRef = node->getSymbolReference();
    if (symRef->isUnresolved())
-      return;
+      return false;
 
    uintptrj_t *objectReferenceLocation = NULL;
    if (symRef->hasKnownObjectIndex())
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik Can you review my change?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1419,15 +1419,15 @@ static const char *getFieldSignature(OMR::ValuePropagation *vp, TR::Node *node,
    return NULL;
    }
 
-static void addKnownObjectConstraints(OMR::ValuePropagation *vp, TR::Node *node)
```
mgaudet: Maybe a comment here to indicate what true and false mean to consumers would be good?

andrewcraik:
LGTM - over @vijaysun-omr since this is in his area of expertise. @jdmpapin FYI - review not required I don't think, but feel free to comment since you know this code well.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1516,9 +1525,7 @@ TR::Node *constrainAload(OMR::ValuePropagation *vp, TR::Node *node)
       return node;
 
    // before undertaking aload specific handling see if the default load transformation helps
-   // But even if it does help, node may still be an aload that we can constrain.
-   if (TR::TransformUtil::transformDirectLoad(vp->comp(), node)
-       && node->getOpCodeValue() != TR::aload)
+   if (TR::TransformUtil::transformDirectLoad(vp->comp(), node))
```
vijaysun-omr: Why did we drop the != TR::aload test ?

liqunl: If `transformDirectLoad` has done something, it must have changed the node to a constant. For non-collected reference, it'll be an `aconst` node; for collected reference, it'll be a known object index on the symref of the `TR::aload` node. Both cases can be handled by `constrainNewlyFoldedConst` now. Notice this commit contains a change to `constrainNewlyFoldedConst` to create known object constraint.

vijaysun-omr: Okay so it is specifically to allow the known object constraint to be created in the "current" pass then.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1540,7 +1547,8 @@ TR::Node *constrainAload(OMR::ValuePropagation *vp, TR::Node *node)
          vp->addGlobalConstraint(node, TR::VPObjectLocation::create(vp, TR::VPObjectLocation::J9ClassObject));
          }
 
-      addKnownObjectConstraints(vp, node);
+      if (addKnownObjectConstraint(vp, node))
```
keithc-ca: Should this be a call to `addKnownObjectConstraints()` (note the name is plural)?

liqunl: My bad, I forgot that I have made this little change after tested my branch. I've opened https://github.com/eclipse/omr/pull/1978 for it. @vijaysun-omr Can you review and merge it? I have tested it this time, and so sorry for the trouble.

PR #1960, opened 2017-11-06
Small fixes and additions to Tril printing routines

mgaudet:
- Correct defaulting to stdout in some places.
- Add printTreesToStdErr to support calling in a
  debugger that may not have a good definition for
  stderr or stdout


charliegracie:
genie-omr build plinux

PR #1961, opened 2017-11-06
Support doing linkage definition on call sites in Tril

mgaudet:
~_Note: Only 8e87394 is new, as this PR builds on top of #1910_~

This PR is a feedback request for specifying linkage on a call. The hope is that this kind of functionality can be used to test a PR similar to #1899. 

This PR only covers the setting the linkage on calls. However, a very similar mechanism should be usable to set the linkage on the Tril described method itself, so that it receives parameters based on the specified linkage.

I also explored the use of Type Parameterized tests in google test, coming to the conclusion I didn't really like them for this, as it makes the test cases awkward.

### DIFF
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -41,6 +41,7 @@ add_executable(comptest
 	LongAndAsRotateTest.cpp
 	MockStrategyTest.cpp
 	LogicalTest.cpp
+	LinkageTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/LinkageTest.cpp
+++ b/fvtest/compilertriltest/LinkageTest.cpp
@@ -1,0 +1,140 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2017 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
+#include "JitTest.hpp"
+#include "default_compiler.hpp"
+
+/* Template for mapping a C/C++ type to
+ * a TR type, and type prefix.
+ */
+template<typename T>
+struct TypeToString {
+   static const char * type;
+   static const char * prefix;
+};
+
+template<> const char * TypeToString<int32_t>::type = "Int32";
+template<> const char * TypeToString<int64_t>::type = "Int64";
+template<> const char * TypeToString<float>  ::type = "Float";
+template<> const char * TypeToString<double> ::type = "Double";
+
+template<> const char * TypeToString<int32_t>::prefix = "i";
+template<> const char * TypeToString<int64_t>::prefix = "l";
+template<> const char * TypeToString<float>  ::prefix = "f";
+template<> const char * TypeToString<double> ::prefix = "d";
+
+TEST(TypeToString,TestTemplate) {
+   EXPECT_STREQ("Int32", TypeToString<int32_t>::type);
+   EXPECT_STREQ("Double", TypeToString<double>::type);
+}
+
+
+template <typename T>
+T passThrough(T x) { return x; }
+
+
+template <typename T>
+class LinkageTest : public TRTest::JitTest {};
+
+typedef ::testing::Types<int32_t, int64_t, float, double> InputTypes;
+TYPED_TEST_CASE(LinkageTest, InputTypes);
+
+TYPED_TEST(LinkageTest, SystemLinkageParameterPassingSingleArg) {
+    char inputTrees[200] = {0};
+    const auto format_string = "(method return=%s args=[%s] (block (%sreturn (%scall address=%p args=[%s] linkage=system (%sload parm=0)) )  ))";
+    std::snprintf(inputTrees, 200, format_string, TypeToString<TypeParam>::type, // Return
+                                                  TypeToString<TypeParam>::type, //Args
+                                                  TypeToString<TypeParam>::prefix, //return
+                                                  TypeToString<TypeParam>::prefix, //call
+                                                  &passThrough<TypeParam>, //address
+                                                  TypeToString<TypeParam>::type, //args
+                                                  TypeToString<TypeParam>::prefix //load
+                                                  );
+    auto trees = parseString(inputTrees);
+
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+
+    auto entry_point = compiler.getEntryPoint<TypeParam (*)(TypeParam)>();
+
+    EXPECT_EQ(static_cast<TypeParam>(0), entry_point( static_cast<TypeParam>(0)))  << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(1), entry_point( static_cast<TypeParam>(1)))  << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(-1), entry_point(static_cast<TypeParam>(-1))) << "Input Trees: " << inputTrees;
+#endif
+}
+
+template <typename T>
+T fourthArg(T a, T b, T c, T d) { return d; }
+
+
+TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
+    char inputTrees[400] = {0};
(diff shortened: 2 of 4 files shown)


fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) { 
+    char inputTrees[400] = {0};
+    const auto format_string = "(method return=%s args=[%s,%s,%s,%s] (block (%sreturn (%scall address=%p args=[%s,%s,%s,%s] linkage=system"
+                                 " (%sload parm=0)"
+                                 " (%sload parm=1)"
+                                 " (%sload parm=2)"
+                                 " (%sload parm=3)"
+                                 ") )  ))";
+
+    std::snprintf(inputTrees, 400, format_string, TypeToString<TypeParam>::type,   // Return
```
mgaudet: This might be the biggest argument in favour of exposing a type-genericized builder pattern for the Tril AST. 

My thought being something like 

```
return<T>().
  call<T>().
    load<T>(0);
```

mgaudet:
I started sketching together adding linkage selection to `(method` blocks, so you could compile a method expecting to _receive_ parameters under another linkage. 

It doesn't seem terribly complicated, though, there's a design question here. In the Non-J9 compilers, the linkage is set inside `compileMethodFromDetails`

https://github.com/eclipse/omr/blob/c788ba35b810073df3e8e56a154ec21011cec3d1/compiler/control/CompileMethod.cpp#L357

The question is how would you communicate a linkage requirement to that method. I don't particularly want to add another parameter to that function, but the `JittedMethodSymbol` doesn't exist until after the `TR::Compilation` exists, which is created inside `compileMethodFromDetails`. 

My first thought on this is to augment `TR::IlGenMethodDetails` with a linkage parameter. @mstoodle  How does that sound to you from a conceptual integrity standpoint?

mstoodle:
I don't think that's the right spot. I would have thought linkage would be an aspect of TR_ResolvedMethod. i.e. at the point you know precisely which concrete method implementation you're talking about, you should be able to know what its linkage should be.

@vijjaysun-omr @0xdaryl agree/disagree?

mgaudet:
Yes, and that's how it works today -- the question for me is how to smuggle linkage requirement into and through `compileMethodFromDetails`; the ResolvedMethodSymbol isn't created until TR::Compilation exists.

vijaysun-omr:
I agree with the resolved method being a good place for the linkage info.

mstoodle:
Um, but a `TR::IlGeneratorMethodDetails` has a `TR::ResolvedMethod`, does it not?

mgaudet:
So, right now the linkage information is on the <tt>ResolvedMethod<b>Symbol</b></tt>; is this suggesting that we should be moving it to the `ResolvedMethod` itself? Of should it be transmitted from `ResolvedMethod` to `ResolvedMethodSymbol` on the symbol's creation?

mstoodle:
Linkage doesn't seem like something that's only relevant for the Symbol. I would call it an attribute of the ResolvedMethod.

I'm afraid I don't have a good answer for why linkage convention currently hangs off MethodSymbol and not ResolvedMethod . @vijaysun-omr any memories of this choice?

mgaudet:
Can you generate a direct call with only a `ResolvedMethodSymbol`, or do you need to have an associated `ResolvedMethod`?

If the former, I could see why linkage is on the symbol not the method.

ymanton:
I'm thinking it's probably necessary to know the linkage of a method even if you don't know exactly which method you're calling, since you have to generate code to perform the call following some convention or another.

vijaysun-omr:
I don't think there was a philosophical reason for putting it on the resolved method symbol and not on the resolved method (at least far as I know). 

In response to

Can you generate a direct call with only a ResolvedMethodSymbol, or do you need to have an associated ResolvedMethod?

from @mgaudet, whether a direct call is generated or not depends on the call node's opcode I would have thought.

vijaysun-omr:
Maybe I mis-understood @mgaudet 's question and he was actually asking if we had enough info on the resolved method symbol to generate the direct call (assuming it was using a direct call opcode in the trees) if it was not allowed to look at the underlying resolved method. If so, the answer is "no" in that would probably know what linkage to use today but not know which specific method to call.

mgaudet:
So, totally my own fault, but this PR diverged pretty heavily from its original intent. I've just rebased and pushed the original intent of this as a single commit: This allows specifying the linage of an outgoing generated call. 

The problem of specifying a linkage for a compiled method can deferred to another issue, which I will open separately. 

Now, in the mean time, I did open #1974 which has a system that breaks apart `compileMethodFromDetails` into a stateful object that can be manipulated in such a way to make accessing the `ResolvedMethodSymbol` for a compilation possible. This is one potential alternative solution, instead of trying to relocate the linkage info from the `ResolvedMethodSymbol` to the `ResolvedMethod`. 


(_Note to self: There's a follow up item here to coordinate this change with #1930, to provide documentation of linkage specification_)

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+
+    auto entry_point = compiler.getEntryPoint<TypeParam (*)(TypeParam,TypeParam,TypeParam,TypeParam)>();
+
+    EXPECT_EQ(static_cast<TypeParam>(1024), entry_point(0,0,0,static_cast<TypeParam>(1024)))  << "Input Trees: " << inputTrees;
```
Leonardo2718: Maybe use more randomish numbers here, rather than 0 and 1024.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "JitTest.hpp"
+#include "default_compiler.hpp"
+
+
+/* 
+int32_t int32PassThrough(int32_t x) { return x; } 
+int64_t int64PassThrough(int64_t x) { return x; } 
+float   floatPassThrough(float x)   { return x; } 
+double  doublePassThrough(double x) { return x; } 
+*/
```
Leonardo2718: Are these still needed?

mgaudet:
Addressed Leo's comments. This is good to go I think

fvtest/compilertriltest/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 add_compile_options(-pthread)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fno-rtti -fno-threadsafe-statics -Wno-deprecated -Wno-enum-compare -Wno-invalid-offsetof -Wno-write-strings -O3 -fomit-frame-pointer -fasynchronous-unwind-tables -Wreturn-type -fno-dollars-in-identifiers -m64 -fno-strict-aliasing")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fno-threadsafe-statics -Wno-deprecated -Wno-enum-compare -Wno-invalid-offsetof -Wno-write-strings -O3 -fomit-frame-pointer -fasynchronous-unwind-tables -Wreturn-type -fno-dollars-in-identifiers -m64 -fno-strict-aliasing")
```
mstoodle: This change seems like an important one. Can you explain why you need to remove the -fno-rtti flag to "support doing linkage definition on call sites in tril" ?

fvtest/tril/tril/compiler_util.hpp
```diff
@@ -73,6 +74,26 @@ static std::vector<TR::DataTypes> parseArgTypes(const ASTNode* node) {
...
+ *         recognized. 
+ */
+static TR_LinkageConventions convertStringToLinkage(const char * linkageName) { 
+   std::string ln = linkageName;  
+   if (ln == "private") 
+      return TR_Private; 
+   if (ln == "system")
+      return TR_System;
+   if (ln == "allregister") 
+      return TR_AllRegister;
```
mstoodle: Given we only have a test for "system" linkage, maybe it would be better to just define "system" linkage here and then add other linkages as we write tests that use and validate them?

fvtest/tril/tril/ilgen.cpp
```diff
@@ -287,8 +290,22 @@ TR::Node* Tril::TRLangBuilder::toTRNode(const ASTNode* const tree) {
...
+           TR_LinkageConventions lc = convertStringToLinkage(linkageString); 
+           if (lc != TR_None) {
+              methodSymRef->getSymbol()->castToMethodSymbol()->setLinkage(lc);     
+              TraceIL("  customizing linakge of call to %s (lc=%d)\n", linkageString, lc);
+           } else { 
+              TraceIL("  failed to find customized linkage %s, aborting parsing\n", linkageString);
+              return NULL; 
+           }
+        }
+        node  = TR::Node::createWithSymRef(opcode.getOpCodeValue(), childCount, methodSymRef);
```
mstoodle: If you're going to optionally override the linkage here I think I'd prefer the default linkage setting to be made explicitly here (even if it's redundant because it's already set somewhere else).

fvtest/compilertriltest/CMakeLists.txt
mgaudet: Ah, whoops. I should remove that from this PR. 

That change makes Google Test's type-parameterized tests actually report the type they are testing, rather than just saying `<type>` all the time. 

But this work should be done as a different PR which I will start right away

fvtest/tril/tril/compiler_util.hpp
mgaudet: Done

fvtest/tril/tril/ilgen.cpp
mgaudet: Done.

mgaudet:
Fixed and rebased.

mgaudet:
rebased to address conflict again.

mgaudet:
rebased to address conflict again.

mgaudet:
@0xdaryl Any other feedback on this?

mstoodle:
Not a blocking concern, per se, but a test designed to fail with a linkage other than the default (and only supported) linkage might be appropriate just to make sure everything is *really* operational. i.e. a test that tries to do a call with, say, linkage=private or some string other than system.

mgaudet:
Thanks Mark. That's a good point. I have now added that test (which actually caught a problem in IlGenFailure propagation)

0xdaryl:
genie-omr build all

fvtest/compilertriltest/LinkageTest.cpp
dchopra001: FYI I think this is also causing compilation failures on older compiler versions. More here: https://github.com/eclipse/omr/issues/2152

Issue #1955, opened 2017-11-04
Compilation failures in omr/tools when cross compiling


JamesKingdon:
In the context of attempting an OpenJDK/OpenJ9 cross compile for arm 32 bit, the compilation fails in the tools directory. It appears the makefiles are picking up the 32 bit size of the target platform and applying it to the tools which are being compiled for the host. Since the host is 64 bit and not configured for multi-arch this fails with missing include files (note the -m32 in the options)
```
g++  -I.  -DUT_DIRECT_TRACE_REGISTRATION  -DLINUX -D_REENTRANT -D_FILE_OFFSET_BITS=64 -DJ9X86 -c  -MMD -MP -fno-exceptions -fno-rtti -fno-threadsafe-statics -fpic -m32 -msse2 -I/usr/include/nptl -Wall -Wno-non-virtual-dtor -O3 -fno-strict-aliasing -march=pentium4 -mtune=prescott   -o FileReader.o FileReader.cpp
In file included from /usr/include/stdlib.h:24:0,
                 from FileReader.cpp:23:
/usr/include/features.h:367:25: fatal error: sys/cdefs.h: No such file or directory
```
The makefiles need to differentiate between the target and host architectures.

charliegracie:
The tools can/should be configured to be compiled for the host and not the target.  @youngar can you provide some guidance here?

JamesKingdon:
The work around I used was to add OMR_ENV_DATA64=1 to toolconfigure.mk.in in the conditional section for cross compile. I have no idea if this is the 'correct' solution or not.

youngar:
Hi, i'm seeing what you describe; `OMR_ENV_DATA64` is picked up from the target platform.  We do have separate variables `OMR_TARGET_DATASIZE... 32` and `OMR_BUILD_DATASIZE... 64` to detect the host and target platform.  We need to change how the tools are configured to set `OMR_ENV_DATA64` in toolconfigure.mk.in based on the host. Our build farm doesn't have a problem building the tools with `-m32`, so we never noticed.

JamesKingdon:
Many thanks @youngar, I believe this problem is solved now.

PR #1963, opened 2017-11-07
Set OMR_ENV_DATA64 when configuring build tools

youngar:
Build tools require to be configured according to the current host
platform. This variable currently defaults to the target platform, which
causes cross compilation issues.

issue: #1955


rwy0717:
LGTM, but you might as well also `AC_SUBST` the 32bit flag as well, right?

youngar:
`OMR_ENV_DATA32` is not a flag that we use.  It looks like it was added in the cmake build system though:
```
cmake/modules/OmrDetectSystemInformation.cmake:69:      if(NOT (OMR_ENV_DATA64 OR OMR_ENV_DATA32))
cmake/modules/OmrDetectSystemInformation.cmake:75:                      set(OMR_ENV_DATA32 ON)
cmake/modules/OmrDetectSystemInformation.cmake:83:                      #TODO assert OMR_ENV_DATA32 is not also set
cmake/modules/OmrDetectSystemInformation.cmake:85:              elseif(OMR_ENV_DATA32)
cmake/modules/platform/toolcfg/msvc.cmake:63:elseif(OMR_ENV_DATA32)
```

rwy0717:
Oh OK, don't worry about it then :)

charliegracie:
Is this PR ready?

youngar:
I haven't tested it properly with a Java build. My builds have been failing every time for various unrelated issues (e.g. machine issues). I don't see why it shouldn't be okay, but I want to test it first.

youngar:
@charliegracie good to go

charliegracie:
genie-omr build zlinux

PR #1964, opened 2017-11-07
Disable recomp in MimicInterpreterFrameShape

ncough:
Full speed debug compilations, implemented
using MimicInterpreterFrameShape, must use
the no-opt optimization level. Therefore,
they do not support recompilations.

This change disables recompilations for
any methods compiled in such a configuration.


vijaysun-omr:
Simple change in a mode where recompilation should not be needed since we are not after that level of performance when we are running noOpt and mimicing interpreter frame shape anyway.

PR #1965, opened 2017-11-07
Small corrections to Tril

mgaudet:
* Correct parsing function so that it will work on a 32 bit platform that has `unsigned long` be 32 bits.
* Correct typo in test case name.


charliegracie:
genie-omr build plinux

PR #1966, opened 2017-11-07
Restore CMAKE_ prefix on CXX_EXTENSIONS support var

mgaudet:
Somehow the CMAKE_ prefix was lost on the CMAKE_CXX_EXTENSIONS variable,
and subsequently CMake defaulted to extensions on, rather than off.

Thanks to @dnakamura for spotting this.


charliegracie:
genie-omr build plinux

rwy0717:
LGTM !

Issue #1957, opened 2017-11-06
Unable to include `omrgc.h` in C


pparkkin:
Trying to include `omrgc.h` in C code results in the following error.

```
.../omr/gc/include/omrgc.h:31:1: error: unknown type name ‘class’
 class MM_AllocateInitialization;
```

Compiling with `-Wall -g --std=gnu99`.

```
$ cc --version
cc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
```

charliegracie:
@ktbriggs Can you take a look at this please since the change causing this issue was part of a PR you created for ObjectModelDelegate?  Thanks

pparkkin:
Would it be possible to get something like the example (https://github.com/eclipse/omr/blob/master/example/main.cpp) but using C?

PR #1967, opened 2017-11-07
Fix C interface issues in omrgc.h

ktbriggs:
C interface as presented included a C++ artifact in a function parameter
list. The C++ function is now wrapped in a method that has an acceptable
C parameter list.

omrgc.h is a C header file and should not have included the forward
class declaration or included it in a function parameter list (my bad). 
However, I have left it in the omrgc.h header file, enclosed with `#ifdef 
__cplusplus` for backwards compatibility. AFAIK only GCConfigTest.cpp
uses the C++ interface for this method, within OMR.

Issue: #1957


charliegracie:
genie-omr build all

PR #1969, opened 2017-11-08
Allow restoring default Google Test output

mgaudet:
Sometimes when debugging a test case it's extremely valuable to see
the complete Google Test output. This change allows restoring that
behaviour under an environment variable OMR_VERBOSE_TEST.

### DIFF
--- a/fvtest/omrGtestGlue/omrTest.h
+++ b/fvtest/omrGtestGlue/omrTest.h
@@ -122,6 +122,7 @@ public:
 	}
 
 	static void setDefaultTestListener(bool showTestCases = true, bool showTests = false, bool showSuccess = false, bool showFailures = true) {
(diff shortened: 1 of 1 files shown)


charliegracie:
I like the idea of being able to control how much output is being written out but I am not sure this is the best way.  I think it would be much better if we could control each of the output types individually or have an option to enable / disable them all.

mgaudet:
More control on the output of the OMR listener would be good, but IMO, it's outside the scope of this PR, which is really intended to be a debugging back door for someone facing a challenge. 

If you're still opposed to the all-or-none approach, I don't mind, but I'll close this PR rather than try to build out a listener control system.

charliegracie:
No this is ok as it helps with debugging.

PR #1972, opened 2017-11-08
Improve 32 bit compiler support

charliegracie:
This PR will allow 32 bit Linux x86 to compile the compiler technology.  The JitBuilder, compilertest and compilertriltest now compile successfully on 32 bit.  There are still test failures to be addressed.


mgaudet:
LGTM

One question: Does c3dc671 impact downstream consumers, or does the isolation via the `*Connectors` suffice? It's likely a similar change should also be made in downstream projects regardless, I'm just not sure if this ends up being a breaking change or not.

charliegracie:
FYI @mstoodle, @0xdaryl, @fjeremic, @vijaysun-omr, @0dvictor I am very curious about your thoughts (and others) on the namespace change.  Sadly GCC defines an i386 macro which gets substituted by cpp.  At least on one version of GCC I tried just by enabling -m32 on the command line caused this macro to be defined which means the compiler technology can not be compiled for 32 bit Linux. I noticed that for amd64 we used a capitalized AMD64 for the namespace.  I would prefer to continue using i386 but since we know about this issue now I think it would be best to future proof the code.

charliegracie:
@mgaudet that is a good question.  I will check it out later today or tomorrow.  My current belief is that the connectors isolate the change so it does not break things but I will verify.  I think if we go forward with this change any downstream consumers should also make similar changes even if this is not breaking.

mgaudet:
(worth noting as well, in terms of evidence in favour of the change: `i386` **is** listed as a predefined compiler macro at the [Predefined Compiler Macros wiki](https://sourceforge.net/p/predef/wiki/Architectures/))

Leonardo2718:
From what I recall, I am fairly certain that the connectors mechanism effectively isolates this change from downstream projects so as to not cause any build breakage.

However, some downstream projects like OpenJ9 also use the `i386` namesapce. A similar fix should therefore be applied there also, even if this hasn't cause problems thus far.

0xdaryl:
It's unfortunate to change the name, but I support this proposal.  `i386` is not used that pervasively in the codebase, and tends to hide in some of the darker corners.  I don't believe the extensible class connector interface should be affected.

0xdaryl:
A genie build is not required because these changes are x86-specific.

PR #1979, opened 2017-11-09
Improve locking

babsingh:
Currently, there is a limit (spin_thread_limit) on how many threads can
spin at a time on a monitor/lock. Over this spin_thread_limit, threads
don't spin but try to acquire the lock once before waiting. This can
cause high CPU utilization if there are large number of threads that try
to acquire a monitor/lock.

Now onwards, threads over the spin_thread_limit will immediately wait
instead of trying to acquire the monitor/lock once. This will help avoid
high CPU utilization.

### DIFF
--- a/thread/common/threadhelpers.cpp
+++ b/thread/common/threadhelpers.cpp
@@ -70,20 +70,16 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
 	uintptr_t spinCount2Init = monitor->spinCount2;
 	uintptr_t spinCount1Init = monitor->spinCount1;
 
+	uintptr_t spinCount3 = spinCount3Init;
+	uintptr_t spinCount2 = spinCount2Init;
+
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
-	BOOLEAN spinning = TRUE;
  	if (monitor->spinThreads < lib->maxSpinThreads) {
  		VM_AtomicSupport::add(&monitor->spinThreads, 1);
  	} else {
- 		spinCount1Init = 1;
- 		spinCount2Init = 1;
- 		spinCount3Init = 1;
- 		spinning = FALSE;
+ 		goto exit;
  	}
 #endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
-
-	uintptr_t spinCount3 = spinCount3Init;
-	uintptr_t spinCount2 = spinCount2Init;
 
 	for (; spinCount3 > 0; spinCount3--) {
 		for (spinCount2 = spinCount2Init; spinCount2 > 0; spinCount2--) {
@@ -134,11 +130,10 @@ update_jlm:
 #endif /* OMR_THR_JLM */
 
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
-	if (spinning) {
-		VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
-	}
-#endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
+	VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
 
+exit:
+#endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
 	return result;
 }
 


babsingh:
fyi - @charliegracie; these changes can be merged.

thread/common/threadhelpers.cpp
```diff
@@ -134,11 +130,10 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
 #endif /* OMR_THR_JLM */
 
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
-	if (spinning) {
-		VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
-	}
+	VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
 #endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
 
+exit:
```
DanHeidinga: The `exit` label is unconditionally compiled in but only used under `OMR_THR_SPIN_WAKE_CONTROL` - does this compile successfully if the #define isn't set?

babsingh: no, it will not compile properly if the label is unused. moved `exit:` label within `OMR_THR_SPIN_WAKE_CONTROL` ifdefs.

DanHeidinga:
This skips all JLM updates for this lock.  Have you looked at the JLM data to determine the effects of this?  Are the highly contended locks as visible as before?

babsingh:
The skip is only performed when the number of spin threads exceed `maxSpinThreads`. Before, we incremented JLM counters by `+1` if the number of spin threads exceeded `maxSpinThreads`. 
Also, it is very rare for number of spin threads to be `>=` to `maxSpinThreads`. The skip should have minimal impact on JLM data. We should still be able to identify contended locks using JLM data.

charliegracie:
Can you provide a more detailed description of the reason for this change?  What is the reason for this change?  IF there is CPU available you may get better performance by spinning hence the reason the code currently spins.  What is the value of lib->maxSpinThreads and how is it calculated?  Why is the calculation only based on the current monitor and not on the thread library?  If there are multiple monitors causing heavy spinning there can still be more threads spinning than there are CPUs.

babsingh:
**Can you provide a more detailed description of the reason for this change?**
Performance regression is observed due to micro-contention. Micro-contention - lot of threads trying to acquire a lock with a small critical section. To acquire a lock, a lot of threads can spin simultaneously. Threads may also try to repetitively acquire the lock. This can bottleneck CPU utilization. Majority of CPUs will only be used for spinning. This will prevent other threads to do useful work. Hence, an application's performance (throughput, response time ...) will be adversely impacted.

**What is the reason for this change?**
The aim of this change is to prevent CPU hogging caused due to lock spinning. If an application is negatively impacted by micro-contention, this change will improve the application's performance.

**If there is CPU available you may get better performance by spinning hence the reason the code currently spins.**
These changes don't completely stop spinning. It just adds a control limiter on how many threads should spin simultaneously. 

**What is the value of lib->maxSpinThreads and how is it calculated?**
Within OMR, lib->maxSpinThreads is initialized to 1. In the JVM, lib->maxSpinThreads is currently initialized to number of available CPUs on a machine. New changes pending; lib->maxSpinThreads will be initialized differently in the JVM (Reference - https://github.com/eclipse/openj9/pull/553). New initialization protocol for lib->maxSpinThreads in the JVM:
```
For 1-4 CPU machines, maxSpinThreads is set to (available_CPUs).
For 5-16 CPU machines, maxSpinThreads is set to (available_CPUs/2).
For 17-64 CPU machines, maxSpinThreads is set to (available_CPUs/3).
For greater than 64 CPU machines, maxSpinThreads is set to
(available_CPUs/4).
```

**Why is the calculation only based on the current monitor and not on the thread library?**
We tried global heuristics (in the thread library) to control spinning but it had a negative impact on performance. We found that local/per-monitor heuristics are better to control spinning; this approach didn't have a negative impact on performance.

**If there are multiple monitors causing heavy spinning there can still be more threads spinning than there are CPUs.**
True. Working on a design to vary spinning per monitor (feedback/control/self-learn). This will help minimize lib->maxSpinThreads to 1 and avoid access spinning (efficient CPU utilization). Also, working on a design to wake threads early and spin before a lock is released. These improvements will be an addition to our adaptive spin algorithm.

Currently, adaptive spin only stops spinning if upper bounds are violated for `holdtime` and `slow_percent`. Upper bounds are constants. During an investigation, I found that adaptive spin doesn't trigger or stops spinning. I think the upper bounds shouldn't be constants. Upper bounds should be set dynamically depending upon the environment. Also, working on a design to improve existing adaptive spin approach.
```
holdtime_threshold = 1 million
slow_percent_threshold = 10

if ((holdtime_count > 0) && (avg_holdtime > holdtime_threshold))
     || ((enter_count > 0) && (slow_percent >= slow_percent_threshold)) {
         STOP_SPINNING;
}
```

babsingh:
@charliegracie Ran benchmarks (Volanomark and Apache JMeter/WAS) to verify JLM data. Heavily contended locks are as visible as before. Noticed an improvement. For heavily contended locks, JLM counters used to overflow before; this led to inaccurate JLM results. With the new changes, the probability of overflow has reduced; more accurate JLM results are seen for heavily contended locks since we don't increment JLM counters for `spinThreads > maxSpinThreads`.

charliegracie:
genie-omr build all

PR #1981, opened 2017-11-09
Disable *Death*Tests for OS X CI testing

charliegracie:
These tests take up to 10s of minutes when running on the OS X CI
machines which causes the tests to fail. When running locally
these tests on take seconds. This disables the tests from the
OS X CI until we can resolve the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,6 @@ before_install:
 env:
   global:
     - GTEST_COLOR=1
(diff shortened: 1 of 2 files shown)


charliegracie:
I do not think there is any need for the genie tests since this is just an update for the Travis testing.

mgaudet:
LGTM :shipit:

mstoodle:
Looks like #1811 captures the root problem, so we have a reminder to fix it. Merging.

PR #1982, opened 2017-11-09
Correct JitBuilder's optimizer strategy selection

mgaudet:
Previously by providing it's own optimizer strategy selection, JitBuilder ended up bypassing the mock optimizer support. 

This PR changes the way the JitBuilder chooses it's strategies, and adds a test for Mock optimization strategy support.

### DIFF
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -42,6 +42,7 @@ add_executable(comptest
 	VectorTest.cpp
 	CallTest.cpp
 	LongAndAsRotateTest.cpp
+	MockStrategyTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/MockStrategyTest.cpp
+++ b/fvtest/compilertriltest/MockStrategyTest.cpp
@@ -29,35 +29,34 @@
 
 
 /**
- * Test Fixture for SimplifierFoldAndTest that 
- * selects only the relevant opts for the test case
+ * Uses the mockStrategy support to select only an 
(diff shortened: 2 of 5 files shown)


fvtest/compilertriltest/SimplifierFoldAndTest.cpp
```diff
@@ -117,3 +118,46 @@ TEST_F(SimplifierFoldAndTest, FoldHappens) {
...
+/*
+ * Given the above, if we disable the simplifier
+ * the same trees should _not_ fold. This test case
+ * ensures that the mock optimizer support is working. 
+ *
+ * Coupling note: This is assuming the default optimization
+ * strategy contains the simplifier. If it did not, this test
+ * case would be invalid
+ */
+TEST_F(MockOptimizationTest, FoldDoesntHappen) {
```
Leonardo2718: If I understand correctly, this test is verifying that the mock optimizer is being used and is working. Would it not be better than to put this test in it's own file? This would allow more extensive testing of the mock optimizer later on.

mgaudet: I realized after I went home last night that calling it "mock optimizer" is the wrong phrase -- it's a custom strategy; But I will split this test out. The downside is the duplication of the test case string.

mgaudet:
OS/X timeout failure in TravisCI (Will go away once #1981 is merged)

mgaudet:
Rebased atop #1981, so the OS/X CI failure _should_ go away.

mgaudet:
Rebased (and corrected a whitespace issue) 

Should be good to go.

0xdaryl:
genie-omr build all

PR #1983, opened 2017-11-09
Memoize getLongestPathOfDAG() to run in O(nlogk)

jdmpapin:
where n is the number of nodes in the subtree, and k is the number of nodes found by getLongestPathOfDAG() in the same extended block.

The previous implementation was O(n^2 logn) because it allowed each node to be queued multiple times, up to the max path length.

### DIFF
--- a/compiler/optimizer/DeadTreesElimination.cpp
+++ b/compiler/optimizer/DeadTreesElimination.cpp
@@ -199,43 +199,27 @@ bool collectSymbolReferencesInNode(TR::Node *node,
    return true;
    }
 
-// TODO: Path retrieval is not implemented, but is trivial, making this O(N)
-static int32_t getLongestPathOfDAG(TR::Node *entry, TR::Compilation *cm)
+typedef std::pair<TR::Node*, int32_t> LPEntry;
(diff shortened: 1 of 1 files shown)


jdmpapin:
This survived tests in J9 with a `TR_ASSERT_FATAL` checking that the results are always the same as the previous implementation

vijaysun-omr:
@andrewcraik should also review this

andrewcraik:
The implementation itself looks good. I do see that there can be a number of calls to this since this helper is called from isSafeToReplaceNode and was wondering if there would be an opportunity to cache results between calls to save re-computation, but I don't think that is likely given the current usage of isSafeToRepalceNode. @jdmpapin do you agree with this assessment?

Otherwise I think the code looks fine and is a good bit of clean-up.

jdmpapin:
Now that you mention it, I think the results for each node can be memoized if they're computed bottom-up instead of top-down. Marking this WIP for the moment

jdmpapin:
Updated to remember the value for each node within the extended block

andrewcraik:
LGTM from me over to @vijaysun-omr since he is familiar with this part of the compiler.

vijaysun-omr:
New implementation looks equivalent. Merging.

Issue #1854, opened 2017-10-26
Compiler Memory Management Hierarchy Discussion


dsouzai:
This issue includes structure from https://github.com/eclipse/openj9.

### Background
**Providers**: These classes have APIs to provide a segment (indentation is inheritance; hierarchy/naming could be subject to change)
```
TR::SegmentProvider
'-> J9::DebugSegmentProvider
'-> TR::SegmentPool
'-> TR::SegmentAllocator
    '-> J9::SystemSegmentProvider
    '-> OMR::SystemSegmentProvider
    '-> OMR::DebugSegmentProvider
```
**Allocators**: These classes are used by the above classes to do the actual allocation of memory (indentation is inheritance; hierarchy/naming could be subject to change)
```
TR::RawAllocator

J9::J9SegmentProvider
'-> J9::J9SegmentCache
'-> J9::SegmentAllocator
```
See https://github.com/eclipse/omr/issues/788 for what it means for a class to be in the `TR` namespace as opposed to `J9` or `OMR`.

### Discussion
The point of this discussion is to help create a firm understanding of the memory management hierarchy in TR, in order to help me with the documentation I will be writing in the near future, but also to separate distinct concepts. 

This discussion requires a distinction between what it means to be a Provider and an Allocator. Given the way the code exists today, I’ve come up with these definitions:

* An Allocator is the “allocator” part of a Memory Manager. It is used by a Provider to acquire and free memory. Its job is simply to get memory and free memory. When an Allocator is asked to free memory, it must free that memory; it does not manage any memory by holding it in a free list.
* A Provider is the “manager” part of a Memory Manager. It manages memory it requests from an Allocator. It is the external interface by which higher level constructs request and release memory. When a Provider is asked to free memory, it does not have to release it back to the OS; it can keep it in free lists, or protect that memory to help debug memory corruption issues. A Provider is also not obligated to use an Allocator (as defined above) but instead can use `mmap` or `malloc` or `sbrk` or whatever - as the “manager” part of a Memory Manager, it is free to use whatever it wants to actually acquire the memory it will manage.  

[...]

PR #2003, opened 2017-11-13
Add Compiler Memory Manager Documentation

dsouzai:


dsouzai:
Thanks to @lmaisons for his help with this via https://github.com/eclipse/omr/issues/1854.

doc/compiler/memory/MemoryManager.md
```diff
@@ -0,0 +1,214 @@
+# OMR Compiler Memory Manager
+
+The OMR Compiler doesn't necessarily operate in a vacuum, but rather 
+as one component of a Managed runtime. Additionally,  communications 
+between components could be extern "C". Therefore, one can't simply 
+use something like `::operator new` since it can throw. Additionally, 
+mutli-threaded performance of malloc historically has been hit or miss, 
+and any generic common allocator would have poor interaction between 
+components, with little to no insight into usage.
```
mgaudet: sp. multi

I don't think throwing shade at `malloc` here is a good call - If `malloc`'s threaded performance were really at issue, there would be other allocators we could be deploying instead. The fact we don't is because we believe that we can derive substantial gains in both performance and functionality by having a compiler-domain-specific allocator. 

(I object mildly to the description of the port allocator as 'domain-specialized', because in the context of the compiler, it is precisely _not_ domain specific)

doc/compiler/memory/MemoryManager.md
```diff
@@ -0,0 +1,214 @@
...
+in the Port Library, is expensive (both in terms of memory footprint 
+and execution time) for many of the types of allocations performed 
+by the Compiler; there are too many participants for informal trust, 
+book-keeping is expensive, and optimizing for divergent component 
+use-cases is difficult to impossible.
+
+Therefore, the compiler manages memory its own memory.
+
+
+## Compiler Memory Manager Hierarchy
```
mgaudet: IMO, leading this description with an [ASCII-art diagram](http://asciiflow.com) of the hierarchies would be a great choice.

doc/compiler/memory/MemoryManager.md
```diff
@@ -0,0 +1,214 @@
...
+## Miscellaneous Information
+
+`TR::PersistentAllocator` is not a `TR::Region`. That said, if one wanted
+to create a `TR::Region` that used `TR::PersistentMemory` as its backing
+provider, they would simply need to extend `TR::SegmentProvider`, perhaps
+calling it `TR::PersistentSegmentProvider`, and have it use
+`TR::PersistentAllocator`. The `TR::Region` would then be provided this 
+`TR::PersistentSegmentProvider` object.
+
+`TR::Region` allocations are untyped raw memory. In order to have a region
```
mgaudet: I would actually hoist some of the information about destructor semantics to near the top of the document: It's a peculiar facet of our use of memory and so should be highlighted earlier.

doc/compiler/memory/MemoryManager.md
dsouzai: I reworded the line regarding the port allocator to
>Using an allocator specific to a  Managed Runtime, such as `omrmem_allocate_memory` in the Port Library...

Regarding the comment about `malloc`, I think the point there is that from (our) historical empirical observation, `malloc` has been wasteful for memory when there are multiple simultaneous compilations occurring (since it uses per thread arenas + a global arena); it also results in more fragmentation since memory allocated in a compilation gets freed at the end of the compile but `malloc` doesn't release that memory back to the OS. In terms of other allocators we could be deploying, that's addressed by the line
>any generic common allocator would have poor interaction between components, with little to no insight into usage.

So it's not so much that we're putting down malloc so much as we're stating that in certain compiler configurations, `malloc` ends up hurting more than helping.

I can update the doc to be more explicit about what "hit or miss" means, or I can still take it out if you think it doesn't belong there.

doc/compiler/memory/MemoryManager.md
dsouzai: I can bring that information up, but the concern I have is that a reader is not going to know what it all means until the read the rest of the document. It currently also flows a bit better, since the doc starts with the the motivation, the hierarchy, how the compiler uses memory, and then subtleties and information. Personally I would find it confusing if I saw information at the top that only made sense after reading the whole document, since I would probably not read in order but skip over to the section where a particular term is first described. However, if you think the average reader would be fine with it on top, then I can move it to the top.

doc/compiler/memory/MemoryManager.md
mgaudet: :D I think putting in the above is a pretty good call -- but at the end of the day, we have a story where we believe we can do pretty well for our allocation patterns by taking matters into our own hands.

doc/compiler/memory/MemoryManager.md
mgaudet: I guess I'm torn a bit. 

Maybe surround it with 🚨 emojis to call attention to the fact constructors do not fire by default. That's the thing I think should be made clear as possible.

doc/compiler/memory/MemoryManager.md
lmaisons: Re: shade on malloc. I thing the odds are high that a large part of why we went down this road is because malloc's _historical_ multi-threaded performance was awful.  A lot of this work started when dlmalloc was still the leading implementation.  I have huge respect for dlmalloc, but it is _not_ thread friendly.  Whether or not the historical context belongs in document describing current practice is :shrug: from me.

malloc's multi-threaded performance notwithstanding.  The reason this is all justified today, is that the compiler produces a prolific amount of garbage during a compilation, and if it were all allocated individually from a generic allocator, we'd be at serious risk of long-term memory fragmentation.  That's on top of any heavy-weight RAS requirements for a common runtime allocator.

doc/compiler/memory/MemoryManager.md
lmaisons: +1 for highlighting the nasty gotyas inherent in naive slab allocators.

dsouzai:
@mgaudet updated the document with your requested changes :)

mgaudet:
Looks good to me! 🚨

dsouzai:
@0xdaryl This should be good to merge if everything looks good to you.

PR #2005, opened 2017-11-14
GC: Don't get the extensions if we don't use them

rwy0717:
/cc @charliegracie 


charliegracie:
genie-omr build all

rwy0717:
This fixes a warning about unused variables

PR #2006, opened 2017-11-14
Zero out dead slots at OSR transition for FSD

cathyzhyi:
In debugging mode, the debugger may dereference an object slot even
after it's gone dead and we must make sure such dead slots are zeroed
out at OSR transition to avoid program crashing.

VoluntaryOSR could zero out the dead slots at induceOSR in trees
because there is one induceOSR for each osrPoint. However, under
involuntarOSR, which debbuging mode runs with, multiple osrPoints
would transit to the same code block and therefore the zero stores
can not be expressed in trees shared between all the different
osrPoints.

This changeset adds the infrastructure to bookkeep deadslots while
compiling the Java code. At OSR transition at runtime, the prepareForOSR
helper reads the data and decide which slots to zero out.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -603,6 +603,20 @@ int32_t TR_OSRDefAnalysis::perform()
          }
       return 0;
       }
+   else 
+      {
+      TR_OSRMethodData *osrMethodData = comp()->getOSRCompilationData()->findOrCreateOSRMethodData(comp()->getCurrentInlinedSiteIndex(), comp()->getMethodSymbol());
+      if (osrMethodData->hasSlotSharingOrDeadSlotsInfo())
+         {
+         if (trace())
+            {
+            traceMsg(comp(), "%s OSR reaching definition analysis is not required as it has already been calculated\n",
+               optimizer()->getMethodSymbol()->signature(comp()->trMemory()));
+            traceMsg(comp(), "Returning...\n");
+            }
+         return 0;
+         }
+      }
 
    if (trace())
       {
@@ -687,7 +701,9 @@ TR_OSRLiveRangeAnalysis::TR_OSRLiveRangeAnalysis(TR::OptimizationManager *manage
      _liveVars(NULL),
      _pendingPushSymRefs(NULL),
      _sharedSymRefs(NULL),
-     _workBitVector(NULL)
+     _workBitVector(NULL),
+     _workDeadSymRefs(NULL),
+     _visitedBCI(NULL)
    {}
 
 bool TR_OSRLiveRangeAnalysis::shouldPerformAnalysis()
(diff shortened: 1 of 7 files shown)


andrewcraik:
@cathyzhyi this change now has merge conflicts could you please rebase and resolve so the review can continue? Is this now ready for review?

cathyzhyi:
@andrewcraik just resolved the conflict. waiting for the tests pass.

andrewcraik:
@ncough could you provide a review please?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1474,26 +1754,37 @@ void TR_OSRLiveRangeAnalysis::buildOSRLiveRangeInfo(TR::Node *node, TR_BitVector
...
-            deadSymRefs->set(liveLocalIndexToSymRefNumberMap[nextDeadVar]);
-            }
+            deadSymRefsTemp.set(liveLocalIndexToSymRefNumberMap[nextDeadVar]);
+         }
+
+      // Load the existing BitVector
+      deadSymRefs = osrMethodData->getLiveRangeInfo(byteCodeIndex);
+      if (!deadSymRefs)
+         {
+         deadSymRefs = new (trHeapMemory()) TR_BitVector(0, trMemory(), heapAlloc);
```
ncough: The conditions for an allocation here have changed, which could result in more heap usage. Would it be possible to prevent an allocation if its know that the result will be an empty bit vector, by checking `deadSymRefsTemp `?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1474,26 +1754,37 @@ void TR_OSRLiveRangeAnalysis::buildOSRLiveRangeInfo(TR::Node *node, TR_BitVector
    *_workBitVector -= *liveVars;
 
    TR_BitVector *deadSymRefs = NULL;
+   TR_BitVector deadSymRefsTemp(0, comp()->trMemory()->currentStackRegion());
```
ncough: Other temporary bitvectors in this optimization are allocated once and cleared, such as `_workBitVector`. I'm not sure if this is actually worthwhile for compile time, just an observation.

ncough:
LGTM

compiler/optimizer/OSRDefAnalysis.cpp
cathyzhyi: Thanks for the suggestion. Changed the code to allocate deadSymRefs bit vector only when necessary.

compiler/optimizer/OSRDefAnalysis.cpp
cathyzhyi: Thanks for pointing that out! Changed the deadSymRefsTemp to work the same way as _workBitVector to save compilation time.

cathyzhyi:
Testing updated changes.

cathyzhyi:
Finished with testing and is ready to be merged.

vijaysun-omr:
genie-omr build all

vijaysun-omr:
Looks okay to me. Merging.

PR #2007, opened 2017-11-14
Encourage the use of issue closing tags

0xdaryl:
Modify the Contribution Guidelines to encourage the use of a Closing tag
when merging the final commit that completely addresses an issue.  This
will cause the issue to automatically close.

[ci skip]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,14 @@ Key: Value
 ```
 
 When a commit has related issues or commits, explain the relation in the message
-body. You should also leave an `Issue` tag in the footer. For example:
+body. You should also leave an `Issue` tag in the footer. However, if this is the
+final commit that fixes an issue you should leave a `Closes` tag (or one of the
+tags described [here](https://help.github.com/articles/closing-issues-using-keywords/)
+in the footer instead which will automatically close the referenced issue when
+this pull request is merged.
+
+For example, if this is just one of the commits necessary to address Issue 1234
+then the following is a valid commit message:
 
 ```
 Correct race in frobnicator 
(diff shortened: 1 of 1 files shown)


0xdaryl:
@mstoodle @charliegracie : please review

CONTRIBUTING.md
```diff
@@ -95,7 +95,13 @@ Key: Value
 ```
 
 When a commit has related issues or commits, explain the relation in the message
-body. You should also leave an `Issue` tag in the footer. For example:
+body. You should also leave an `Issue` tag in the footer. However, if this is the
+final commit that fixes an issue you should leave a `Closes` tag in the footer
```
charliegracie: There is more than one tag which will automatically close the PR.  Maybe change the text to `Closes` or one of the tags found [here](https://help.github.com/articles/closing-issues-using-keywords/).

CONTRIBUTING.md
```diff
@@ -105,6 +111,17 @@ This patch eliminates the race condition in issue #1234.
 Issue: #1234
 ```
 
+However, if this is the final commit that addresses the issue then the following
```
charliegracie: I would prefer less strongly worded text, similar to `However, if this is the final commit that addresses the issue the following is a valid commit message`.

charliegracie: If the Author of a commit put the `Issue:` tag in the commit message the Committer who is merging the commit can always modify the merge commit message to include the `closes` or `fixes`, etc. tag.

CONTRIBUTING.md
0xdaryl: Fixed.

CONTRIBUTING.md
0xdaryl: Added your less strongly worded text.  Your second comment is a true statement, but is it something you want in the CG?

charliegracie: No second comment was just a comment :).  I agree it would not make sense in the CG.

PR #2008, opened 2017-11-15
Support building OMR with Runtime Sanitizers

mgaudet:
Some compiler systems (gcc and clang) support the notion of Runtime
Sanitizers. These are builds that are instrumented to flag bugs at
runtime. Sanitizers are nice because they typically have a very low
false positive rate, and as such, when an issue is flagged, it typically
can be relied on.

### DIFF
--- a/cmake/modules/OmrSanitizerSupport.cmake
+++ b/cmake/modules/OmrSanitizerSupport.cmake
@@ -1,0 +1,40 @@
+###############################################################################
+# Copyright (c) 2017, 2017 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+# or the Apache License, Version 2.0 which accompanies this distribution and
+# is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# This Source Code may also be made available under the following
+# Secondary Licenses when the conditions for such availability set
+# forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+# General Public License, version 2 with the GNU Classpath
+# Exception [1] and GNU General Public License, version 2 with the
+# OpenJDK Assembly Exception [2].
+#
+# [1] https://www.gnu.org/software/classpath/license.html
+# [2] http://openjdk.java.net/legal/assembly-exception.html
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+###############################################################################
+
+include(OmrUtility)
+
+set(OMR_SANITIZER OFF CACHE STRING "-fsanitize parameter")
+set_property(
+	CACHE OMR_SANITIZER
+	PROPERTY STRINGS OFF undefined address thread
+	)
+
+# GNU style sanitizers 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU") 
+	if (OMR_SANITIZER) 
+		message(STATUS "${OMR_SANITIZER} Sanitizer is enabled")
+		# Extra options for the sanitizers, often documented to improve output
+		set(ADDITIONAL_SANITIZER_FLAGS "-fno-omit-frame-pointer")
+		omr_append_flags(CMAKE_CXX_FLAGS ${ADDITIONAL_SANITIZER_FLAGS} "-fsanitize=${OMR_SANITIZER}")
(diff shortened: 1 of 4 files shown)


cmake/modules/OmrSanitizerSupport.cmake
```diff
@@ -0,0 +1,37 @@
...
+	# Extra options for the sanitizers, often documented to improve output
+	set(ADDITIONAL_SANITIZER_FLAGS "-fno-omit-frame-pointer") 
+	if(OMR_SANITIZE_UNDEFINED)
+		message(STATUS "Undefined Behaviour Sanitization Enabled")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ADDITIONAL_SANITIZER_FLAGS} -fsanitize=undefined")
+		set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}   ${ADDITIONAL_SANITIZER_FLAGS} -fsanitize=undefined")
+	elseif(OMR_SANITIZE_ADDRESS) 
+		message(STATUS "Address Sanitization Enabled")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ADDITIONAL_SANITIZER_FLAGS} -fsanitize=address")
+		set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}   ${ADDITIONAL_SANITIZER_FLAGS} -fsanitize=address")
```
rwy0717: I would write this as:
```cmake
omr_append_flags(CMAKE_C_FLAGS ${ADDITIONAL_SANTIZER_FLAGS} "-fsanitize=address")
```

mgaudet: 👍 Done

rwy0717:
Can sanitizers be combined? Right now the cmake UI allows the user to turn both on, but one overrides the other. Maybe we want one string variable instead? IE:

```cmake
set(OMR_SANITIZER OFF CACHE STRING "The sanitizer")
set_property(
  CACHE OMR_SANITIZER
  PROPERTY STRINGS OFF Address UndefinedBehaviour Thread
)
```

And don't forget to `include(OmrUtility)` add the cached defaults to config.cmake.

rwy0717:
One reason I can think to explicitly support these sanitizers in the build system is, they may one day require in-source support. We may want to introduce [conditional compilation](https://clang.llvm.org/docs/AddressSanitizer.html#conditional-compilation-with-has-feature-address-sanitizer) (`#ifdef OMR_ASAN`), or [custom attributes](https://clang.llvm.org/docs/AddressSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-address) (`OMR_NO_UBSAN`). These would require cmake integration to keep portable.

But if this isn't a direction we're going to take, couldn't we have just had a toolchain file under scripts for special travis sanitizer runs? Are the sanitization flags something we expect developers to be using in a general way, or is this only useful for a small set of specialized test runs on specific platforms?

dnakamura:
@rwy0717 there is some in source support stuff now. Ie you can mark functions with attributes to skip sensitization (in the case of false positives). Although i suppose we could just leave them there, and they should be a `nop` when we arent built with sensitization

rwy0717:
My comment^ is mostly just FUD about accumulating cruft in the build system 😅. I actually am okay with implementing sanitizer support in cmake. If things ever start to get "too big", moving specialized build flags (like these) to toolchain files is one way we could pare down the bloat. That's not a problem we're facing today.

> there is some in source support stuff now. Ie you can mark functions with attributes to skip sensitization (in the case of false positives). Although i suppose we could just leave them there, and they should be a nop when we arent built with sensitization

Yeah yeah that's what I'm saying, the sanitizers have in-source attributes, but we're not using them yet, and if we wanted to, we'd need portable wrapper macros. Though, now that I'm thinking about it, it would be easy to implement the wrapper macros in C without any cmake support (beyond the platform definitions already in place).

mgaudet:
Chose to follow @rwy0717 's lead, and use the cached string to enable or disable.

dnakamura:
@rwy0717 After watching some cppcon talks about undefined behaviour etc, I'm thinking it may actually be productive to turn it on address and UB sanitization on by default in the debug builds

mgaudet:
Different PR 😛

rwy0717:
LGTM

0xdaryl:
@charliegracie , do you have any concerns with this?

PR #2011, opened 2017-11-15
Reset lastFreeEntry

amicic:
On Global GCs reset _lastFreeEntry (to point to NULL) when AOL pool is
reset. The only user of this info is aborted GC handling in Concurrent
Scavenger, which will always be set this through compaction step that is
done as a part of Percolate GGC (in CS world). Still, for cleanness and
for future changes (if/when compaction is made an optional part of
Percolate GGC), this is correct to reset.


charliegracie:
genie-omr build all

PR #1752, opened 2017-09-29
Fix casting issues in generateRILInstruction() API

sehirst:
Splitting the generateRILInstruction() API to handle the different
types of immediates that RIL instructions can take separately.
One kind is a pure immediate which is used for say an ADD
instruction. The other is when the immediate is used as a relative
offset for address calculation like LARL or BRASL instructions.
This allows calls to the API to have more sensible casting.
They are modified for this change and unsafe casting is switched
to static casting and some code has its style improved.

Signed-off-by: Simon Hirst <shirst@ca.ibm.com>

PR #2012, opened 2017-11-15
Make correct call to getOpCodeName in z codegen

sehirst:
A call to getOpCode().getMnemonicName() was recently added but
it is not implemented in z codegen. This is causing problems
building omr on z. This commit fixes this by calling the
correct API.

Signed-off-by: Simon Hirst <shirst@ca.ibm.com>

### DIFF
--- a/compiler/z/codegen/S390Instruction.hpp
+++ b/compiler/z/codegen/S390Instruction.hpp
@@ -2154,7 +2154,7 @@ class S390RILInstruction : public TR::Instruction
       : TR::Instruction(op, n, cg), _targetPtr(NULL), _targetSnippet(NULL), _targetSymbol(NULL), _flagsRIL(0), _mask(0xffffffff),  _targetLabel(NULL), _symbolReference(NULL), _sourceImmediate(imm)
       {
       TR_ASSERT_FATAL(getOpCode().isExtendedImmediate(), "Incorrect S390RILInstruction constructor used.\n"
-                                                       "%s expects an address as the immediate but this constructor is for integer immediates.", getOpCode().getMnemonicName());
+                                                       "%s expects an address as the immediate but this constructor is for integer immediates.", cg->getDebug()->getOpCodeName(&this->getOpCode()));
       checkRegForGPR0Disable(op,treg);
       if (!getOpCode().setsOperand1())
          useSourceRegister(treg);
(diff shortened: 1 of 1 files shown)


sehirst:
@fjeremic Here is the Pull Request which fixes the problem in the `getOpCodeName` call I introduced in my recent Pull Request https://github.com/eclipse/omr/pull/1752

fjeremic:
@0xdaryl could we get an OMR genie build for this hotfix?

charliegracie:
genie-omr build all

charliegracie:
I am curious as to why the previous builds did not pick up this issue.  Is this something that only causes issues in downstream projects?

fjeremic:
> Is this something that only causes issues in downstream projects?

I don't think so. This issue was caught when building comptest in OMR, so no downstream project was involved. As to why this was not caught in one of the builds launched I'm not sure. The latest commit in https://github.com/eclipse/omr/pull/1752 does have a green check mark beside it meaning the tests did run. I'm not sure how it would have not caught a linker error such as this one. Unless compiler tests never ran?

fjeremic:
@charliegracie / @0xdaryl checks passed. Could we get a merge for this?

charliegracie:
I figured out why this was not found in testing.  Linux 390 is not using CMake yet for the CI testing and comptest is only build / executed with the CMake builds

fjeremic:
> I figured out why this was not found in testing. Linux 390 is not using CMake yet for the CI testing and comptest is only build / executed with the CMake builds

Do we have an issue tracking this? Is there a functional issue blocking us from using CMake on Linux on z?

charliegracie:
Merging.... we will need to get the CI testing on Linux 390 using CMake soon so these types of things get caught during testing :)

charliegracie:
@fjeremic last time I tried to use it comptest and the 390 opcode tests were failing a lot of tests.  I was hoping for them to get resolved before I added the CMake testing.  Since they were failing I would have needed to disable them which means we would not have found this failure either most likely.

fjeremic:
> Actually @fjeremic you reviewed and are happy with the change correct?

Yes should be good to merge.

> @fjeremic last time I tried to use it comptest and the 390 opcode tests were failing a lot of tests. I was hoping for them to get resolved before I added the CMake testing. Since they were failing I would have needed to disable them which means we would not have found this failure either most likely.

We've been making some progress here. @dchopra001 has a fix for quite a lot of the failures. We'll see what else is left once that is merged.

PR #2013, opened 2017-11-16
NOP-ing of check involving the static StringCompression field

yanluo7:
Signed-off-by: Yan Luo <yan_luo@ca.ibm.com>

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.cpp
+++ b/compiler/optimizer/OMRValuePropagation.cpp
@@ -8031,9 +8031,11 @@ void OMR::ValuePropagation::doDelayedTransformations()
 
          if ((staticName && (staticNameLen > 0) &&
              (!strncmp(staticName, "com/ibm/websphere/ras/TraceComponent.fineTracingEnabled", 55) ||
-              !strncmp(staticName, "com/ibm/ejs/ras/TraceComponent.anyTracingEnabled", 48))) ||
+              !strncmp(staticName, "com/ibm/ejs/ras/TraceComponent.anyTracingEnabled", 48) ||
+              !strncmp(staticName, "java/lang/String.compressionFlag", 32))) ||
              ((cii->_len == 41) && !strncmp(cii->_sig, "Lcom/ibm/websphere/ras/TraceEnabledToken;", cii->_len)) ||
-             ((cii->_len == 35) && !strncmp(cii->_sig, "Lcom/ibm/ejs/ras/TraceEnabledToken;", cii->_len)))
+             ((cii->_len == 35) && !strncmp(cii->_sig, "Lcom/ibm/ejs/ras/TraceEnabledToken;", cii->_len)) ||
+             ((cii->_len == 40) && !strncmp(cii->_sig, "Ljava/lang/String$StringCompressionFlag;", cii->_len)))
             {
             recognizedStatic = true;
             ifNode->setAndIncChild(0, origFirst->duplicateTree());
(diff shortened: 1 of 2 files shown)


yanluo7:
Currently, we check if a String is compressed by checking its count field is non-negative. We can optimize by checking if the static compressionFlag field is null instead. Because StringCompressionFlag class is only initialized if an uncompressed String is constructed, we can tell there's no uncompressed String objects in the JVM as long as the field is null.
On the JIT side, we can tell the optimizer to generate guarded NOP for these null checks, effectively skipping (compressionFlag == null || count >= 0). When an uncompressed String is constructed, 'StringCompressionFlag' class is initialized and this triggers the JIT to patch all the NOP sites, restoring the (compressionFlag == null || count >= 0) check.

yanluo7:
Currently, we check if a String is compressed by checking its `count` field is non-negative. We can optimize by checking if the static `compressionFlag` field is null instead. Because `StringCompressionFlag` class is only initialized if an uncompressed String is constructed, we can tell there's no uncompressed String objects in the JVM as long as the field is null.
On the JIT side, we can tell the optimizer to generate guarded NOP for these null checks, effectively skipping `(compressionFlag == null || count >= 0)`. When an uncompressed String is constructed, `StringCompressionFlag` class is initialized and this triggers the JIT to patch all the NOP sites, restoring the `(compressionFlag == null || count >= 0)` check.

DanHeidinga:
This looks pretty OpenJ9-specific to me (as does the surrounding code).  Since its being changed now, would it be a good time to refactor this Java-centric code out of OMR and into the OpenJ9 project?

yanluo7:
@andrewcraik Re. Dan's comment about refactoring this into OpenJ9, is it feasible to refactor all the VP code or is it something too big and outside the scope of the string compression work?

andrewcraik:
The refactor is a much bigger undertaking that needs separate design work - I would vote for the status quo - there is already an issue for refactoring VP I believe to move some of this stuff down. It is just a large task with many complications and subtleties that will have to be dealt with.

0xdaryl:
Please remove the trailing `.` from your commit titles per the Contribution Guidelines.

0xdaryl:
genie-omr build all

DanHeidinga:
Re @andrewcraik's comment on maintaining the status quo.  Keeping this code in OMR increases the maintenance costs and makes it harder to evolve the code as OpenJ9 changes need to be merged (and synced!) across two projects.  The OpenJ9-specific code is a barrier to entry for other runtimes interested in OMR.

I don't object to this being merged as is though I worry about the ongoing costs incurred by not refactoring this code.

andrewcraik:
LGTM form me - I agree there is a danger with on going cost - the refactoring is a larger piece of work so needs its own dedicated time and thought. It has not escaped my attention.

vijaysun-omr:
I'd like to ask about your thoughts around the testing of this kind of a change. 
I would suggest testing that the code patching etc. work as expected when the class in question is initialized, i.e. let's test that rare path that is needed for functional correctness via a unit test at least.

yanluo7:
Bumberbench which I have been running can generate uncompressed String for certain special character(s) in the middle of a run, which triggers the patching (after checking generated code/verbose/instrumentation). For a more formal test, I can think of a multithreaded microbenchmark that we used for testing Transactional Memory, where the worker thread(s) can be modified to pause and another thread can be used to trigger the patching.

vijaysun-omr:
Okay I would strongly suggest the more formal multithreaded test to be created as well since there is a slow path that is going to execute quite rarely and it won't be easy to diagnose a problem in this area. Given that your other benchmark is already testing the patching (slow) code I will merge this now.

PR #2019, opened 2017-11-17
Build definingMap for OSRUses symbols

cathyzhyi:
OSRUses under prepareForOSR prolong the live range of autos and
pending push temps. Especially, those fake uses can cause many
autos and temps appear to be alive at the method entry and requires
zero initialization in the method prologue which is big overhead.

A way to solve this problem is to find out the `real` uses
of those autos and pending push temps at each OSRPoint and use them
during liveness analysis. The `real` uses can be found by figuring
out which symbols are needed to calculate the auto or temps at each
prepareForOSRCall and combining this with the liveness information
for the autos and temps from original bytecode.

This changeset builds the map, named definingMap, to map each auto
and temp to the symbols needed to calculate them at any OSRPoint.

### DIFF
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -38,6 +38,7 @@
 #include "env/CompilerEnv.hpp"
 #include "il/Block.hpp"                        // for Block
 #include "il/Node.hpp"                         // for Node
+#include "il/Node_inlines.hpp"                 // for getInt
 #include "il/Symbol.hpp"                       // for Symbol
 #include "il/SymbolReference.hpp"              // for SymbolReference
 #include "il/TreeTop.hpp"                      // for TreeTop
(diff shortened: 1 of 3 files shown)


vijaysun-omr:
I would like a review from @andrewcraik and @ncough please.

andrewcraik:
LGTM from me

compiler/compile/OSRData.cpp
```diff
@@ -513,6 +514,388 @@ uint32_t TR_OSRCompilationData::getOSRStackFrameSize(uint32_t methodIndex)
...
+      if (node->getOpCode().isStoreDirect())
+         symRef = node->getSymbolReference();
+      else if (node->getOpCode().isStoreReg())
+         symRef = node->getRegLoadStoreSymbolReference();
+
+      if (symRef)
+         {
+         if (symRef->getSymbol()->isAutoOrParm())
+            {
+            TR_BitVector subTreeSymRefs;
```
ncough: A `TR_BitVector` without a specified region will perform allocation in persistent memory. Should this be in the current stack region?

andrewcraik: Good catch - yes I think @ncough is right

compiler/compile/OSRData.cpp
```diff
@@ -513,6 +514,388 @@ uint32_t TR_OSRCompilationData::getOSRStackFrameSize(uint32_t methodIndex)
...
+               {
+               symRefs = (*blockMap)[symRef->getReferenceNumber()];
+               symRefs->empty(); // the current store kills previous defs
+               }
+
+            if (!subTreeSymRefs.isEmpty())
+               {
+               if (!symRefs)
+                  symRefs = new (comp()->trStackMemory()) TR_BitVector(0, comp()->trMemory(), stackAlloc);
+               updateDefiningSymRefs(&subTreeSymRefs, blockMap, symRefs);
```
ncough: `symRefs` is extracted from the current map if there is already a mapping for the `symRef`, otherwise its allocated here. In the event its allocated, its never added back to the map. Is this intentional? If so, why allocate and run `updateDefiningSymRefs` at all?

cathyzhyi: You are right. The new allocated symRefs should be added back to the map. Nice Catch!

compiler/compile/OSRData.cpp
cathyzhyi: modified to use current stack region. thanks! @ncough

ncough:
LGTM

andrewcraik:
@vijaysun-omr over to you on this one - could you take a look?

vijaysun-omr:
Looks okay to me. Merging.

PR #2021, opened 2017-11-17
Reorganize part of test compiler

mgaudet:
This commit moves the IlInjectors used in the Test Compiler into the
tests subdirectory. This is the right place for these becase they are
involved mostly in the construction of test cases, not directly related
to the adaptation of the compiler technology to the test compiler.

In essence, this tries to match the physical layout to the pre-existing
logical layout.


0xdaryl:
genie-omr build all

PR #2022, opened 2017-11-17
lastFreeEntry only at the end

amicic:
After Percolate Global GC due to aborted Concurrent Scavenger, rely on
the last free entry (out of which a new Survivor space will be created)
only if it is at the very end of SemiSpace (no objects after it). 


Leonardo2718:
genie-omr build all

charliegracie:
genie-omr add to whitelist

Leonardo2718:
genie-omr build all

PR #2027, opened 2017-11-20
Add runtime assumption with multiple patch sites

ncough:
Runtime assumptions incur a persistent memory
and runtime overhead. It is possible for assumptions
to require several patch sites in a method for the same
event. It is also possible for several events to require
the same set of patch sites.

This adds a runtime assumption capable of representing
such a situation with reduced overhead, by creating
a shared collection of patch sites.

### DIFF
--- a/compiler/runtime/OMRRuntimeAssumptions.cpp
+++ b/compiler/runtime/OMRRuntimeAssumptions.cpp
@@ -40,3 +40,107 @@ void TR::PatchNOPedGuardSite::compensate(bool isSMP, uint8_t *location, uint8_t 
    {
    _patchVirtualGuard(location, destination, isSMP);
    }
+
+TR::PatchSites::PatchSites(TR_PersistentMemory *pm, size_t maxSize) :
+    _maxSize(maxSize),
+    _size(0),
+    _refCount(0),
+    _patchPoints(NULL),
+    _firstLocation(NULL),
+    _lastLocation(NULL)
+   {
+   size_t bytes = sizeof(uint8_t*) * 2 * maxSize;
+   TR_ASSERT(maxSize == bytes/(sizeof(uint8_t*) * 2),  "Requested number of patch sites %d exceeds supported size\n", maxSize);
+   _patchPoints = (uint8_t**) pm->jitPersistentAlloc(bytes);
+   }
+
+/**
+ * Add a patch site to the collection.
+ */
+void TR::PatchSites::add(uint8_t *location, uint8_t *destination)
+   {
+   TR_ASSERT_FATAL(_size < _maxSize, "Cannot add more patch sites, max size is %d", _maxSize);
+   _patchPoints[_size * 2] = location;
+   _patchPoints[_size * 2 + 1] = destination;
+   _size++;
+
+   if (_firstLocation == NULL || location < _firstLocation)
+      _firstLocation = location;
+   if (_lastLocation == NULL || location > _lastLocation)
+      _lastLocation = location;
+   }
+
+/**
+ * Compare this collection with another.
+ * Will return true if both will patch the same locations, in the same order.
+ * Duplicate locations are also respected.
+ */
+bool TR::PatchSites::equals(PatchSites *other)
+   {
+   if (other->getSize() != _size)
+      return false;
+
+   if (_firstLocation != other->getFirstLocation() || _lastLocation != other->getLastLocation())
+      return false;
+
+   for (size_t i = 0; i < _size; ++i)
+      {
+      if (getLocation(i) != other->getLocation(i))
+         return false;
+      }
+   return true;
+   }
+
+/**
+ * Check if a location is contained within this collection.
+ */
+bool TR::PatchSites::containsLocation(uint8_t *location)
+   {
+   if (_firstLocation > location || _lastLocation < location)
+      return false;
+   return internalContainsLocation(location);
+   }
+
+bool TR::PatchSites::internalContainsLocation(uint8_t *location)
+   {
+   for (size_t i = 0; i < _size; ++i) 
+      {
+      if (getLocation(i) == location)
+         return true;
+      }
+   return false;
+   } 
+
+uint8_t *TR::PatchSites::getLocation(size_t index)
+   {
+   TR_ASSERT(index < _size, "Invalid index for patch site %d but size %d", index, _size);
+   return _patchPoints[index * 2];
+   }
+uint8_t *TR::PatchSites::getDestination(size_t index)
+   {
+   TR_ASSERT(index < _size, "Invalid index for patch site %d but size %d", index, _size);
+   return _patchPoints[index * 2 + 1];
+   }
+
+/**
+ * Functions to maintain reference counts from runtime assumptions.
+ * Each new runtime assumption using the list should call addReferences
+ * and should then call reclaim when the runtime assumption is being removed.
+ */
+void TR::PatchSites::addReference()
+   {
+   _refCount++;
+   TR_ASSERT(_refCount > 0, "Reference count overflow");
+   }
+
+void TR::PatchSites::reclaim(PatchSites *sites)
+   {
+   TR_ASSERT(sites->_refCount > 0, "Attempt to reclaim patch sites without a reference");
+   sites->_refCount--;
+
+   if (sites->_refCount == 0)
+      {
+      TR_PersistentMemory::jitPersistentFree(sites->_patchPoints);
+      TR_PersistentMemory::jitPersistentFree(sites);
+      }
+   }
--- a/compiler/runtime/OMRRuntimeAssumptions.hpp
+++ b/compiler/runtime/OMRRuntimeAssumptions.hpp
@@ -33,6 +33,7 @@
 class TR_FrontEnd;
 class TR_PatchJNICallSite;
 namespace TR { class PatchNOPedGuardSite; }
+namespace TR { class PatchMultipleNOPedGuardSites; }
 class TR_PreXRecompile;
 class TR_RedefinedClassPicSite;
 class TR_UnloadedClassPicSite;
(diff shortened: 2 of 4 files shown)


compiler/runtime/OMRRuntimeAssumptions.cpp
```diff
@@ -40,3 +40,36 @@ void TR::PatchNOPedGuardSite::compensate(bool isSMP, uint8_t *location, uint8_t
...
+
+void TR::PatchSites::set(size_t index, uint8_t *location, uint8_t *destination)
+   {
+   TR_ASSERT(index < _size, "Index %d beyond patch sites size %d\n", index, _size);
+   index *= 2;
+   _patchPoints[index] = location;
+   _patchPoints[index + 1] = destination;
+   }
+
+bool TR::PatchSites::equals(PatchSites *other)
```
andrewcraik: Given that other patch sites compare based on equality of the location should this be a comparison of locations? Is there an ordering constraint on the locations? It would make sense to order them in some cases, but I am unsure of the overheads so would appreciate your thoughts.

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -185,6 +187,59 @@ class PatchNOPedGuardSite : public OMR::LocationRedirectRuntimeAssumption
    uint8_t *_destination;
    }; // TR::PatchNOPedGuardSite
 
+class PatchSites
+   {
+   private:
+   int       _refCount;
+   int       _size; 
+   uint8_t **_patchPoints;
```
andrewcraik: The choice of the location and destination being in the same array is interesting - what is the reasoning behind this as opposed to having two arrays of the same length - one for location and one for destination. Is it for locality of data at runtime? If so a comment would be appropriate.

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -185,6 +187,59 @@ class PatchNOPedGuardSite : public OMR::LocationRedirectRuntimeAssumption
...
+
+   public:
+   PatchSites(TR_PersistentMemory *pm, int size);
+
+   int getSize() { return _size; }
+   uint8_t *getLocation(size_t index) { return _patchPoints[index * 2]; }
+   uint8_t *getDestination(size_t index) { return _patchPoints[index * 2 + 2]; }
```
andrewcraik: +2 ? why isn't this +1 given the pairs pattern elsewhere?

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -185,6 +187,59 @@ class PatchNOPedGuardSite : public OMR::LocationRedirectRuntimeAssumption
...
+      return site != 0 && _patchSites->equals(site->getPatchSites());
+      }
+
+   virtual void reclaim()
+      {
+      PatchSites::reclaim(_patchSites);
+      }
+
+   virtual PatchMultipleNOPedGuardSites *asPMNGSite() { return this; }
+   virtual uint8_t *getAssumingPC() { return _patchSites->getLocation(0); }
```
andrewcraik: A comment on why index 0 is the right thing to return here would be helpful.

compiler/runtime/OMRRuntimeAssumptions.cpp
ncough: Ordering has not been enforced in this implementation, as its current uses (in a downstream project) don't provide it cheaply or require it. Considering how large these assumptions can become and how infrequently `equals` is used on `RuntimeAssumptions`, I don't believe its a suitable restriction to make. I have added a `O(n^2)` comparison to `equals` as a tradeoff.

compiler/runtime/OMRRuntimeAssumptions.hpp
ncough: This was a mistake and has been corrected.

compiler/runtime/OMRRuntimeAssumptions.hpp
ncough: I have updated the function with a comment. Existing uses of `getAssumingPC` were used to test whether the assumption was in an address range, such as a method. As all assumptions where assumed to be in the same method, retuning the first location appeared to be safe. As this restriction isn't explicitly enforced anywhere and comparisons may desire smaller ranges, `OMR::RuntimeAssumption` has been modified to consider a range of assuming PCs.

compiler/runtime/OMRRuntimeAssumptions.cpp
```diff
@@ -40,3 +40,88 @@ void TR::PatchNOPedGuardSite::compensate(bool isSMP, uint8_t *location, uint8_t
...
+
+/**
+ * Compare this collection with another.
+ * Will return true if all locations are the same.
+ */
+bool TR::PatchSites::equals(PatchSites *other)
+   {
+   if (other->getSize() != _size)
+      return false;
+
```
andrewcraik: Would it make sense to add a fast path exit based on comparing the ranges of the two sets of assumptions - if they don't match exactly then the sets won't be equal right? That should cause complete equality checks in a minority of cases would be my thoughts...

compiler/runtime/OMRRuntimeAssumptions.hpp
ncough: It improves locality and reduces the number of allocations that are necessary. A single allocation could achieve the same effect, with separate points for the first and second halves of the list, but this would waste an additional field on the class. I have added a comment that expresses this.

compiler/runtime/OMRRuntimeAssumptions.cpp
ncough: Thanks, I have added the fast path.

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -185,6 +197,77 @@ class PatchNOPedGuardSite : public OMR::LocationRedirectRuntimeAssumption
...
+   size_t   getSize() { return _size; }
+   uint8_t *getLocation(size_t index);
+   uint8_t *getDestination(size_t index);
+   uint8_t *getFirstLocation() { return _firstLocation; }
+   uint8_t *getLastLocation() { return _lastLocation; }
+
+   void add(uint8_t *location, uint8_t *destination);
+
+   bool equals(PatchSites *other);
+   bool containsLocation(uint8_t *location);
```
andrewcraik: Is this called publically? If so do we need to add fast paths to this like equals and have an internal implementation that is used by both equals and this one to save scanning? Otherwise this can be the private/protected helper...

ncough: At the moment, its not called publically, but it could provide useful functionality. An internal helper has been added.

andrewcraik:
Looks good to me now - @vijaysun-omr your thoughts?

vijaysun-omr:
Is it possible to push past the size_t limits for reference counts etc. in some case with a pathologically large method ? If so, does this need some kind of conservative handling at least ?

compiler/runtime/OMRRuntimeAssumptions.cpp
```diff
@@ -40,3 +40,102 @@ void TR::PatchNOPedGuardSite::compensate(bool isSMP, uint8_t *location, uint8_t
...
+   if (_firstLocation != other->getFirstLocation() || _lastLocation != other->getLastLocation())
+      return false;
+
+   for (size_t i = 0; i < _size; ++i)
+      {
+      if (!other->internalContainsLocation(getLocation(i)))
+         return false;
+      }
```
jdmpapin: This loop only checks to see that the locations in `this->_patchPoints` are a subset of the ones in `other->_patchPoints`. Since the sizes are equal, that would imply equality as long as locations are not allowed to be repeated, which would be reasonable, but maybe it's worth mentioning

compiler/runtime/OMRRuntimeAssumptions.cpp
```diff
@@ -40,3 +40,102 @@ void TR::PatchNOPedGuardSite::compensate(bool isSMP, uint8_t *location, uint8_t
...
+      _firstLocation = location;
+   if (_lastLocation == NULL || location > _lastLocation)
+      _lastLocation = location;
+   }
+
+/**
+ * Compare this collection with another.
+ * Will return true if all locations are the same.
+ */
+bool TR::PatchSites::equals(PatchSites *other)
```
jdmpapin: `RuntimeAssumption::equals()` looks like it's unused. Just an observation - not to say that it should be removed in this commit, but maybe the behaviour of this method isn't all that relevant

compiler/runtime/OMRRuntimeAssumptions.cpp
ncough: The original implementation assumed locations would be unique, but the use of this feature in OpenJ9 could not make that guarantee. Moreover, enforcing it could incur significant compile time overhead. Instead, the definition of `equals` has been modified to be true when two lists have the same locations, including duplicates, in the same order.

ncough:
More conservative handling has been added to manage the possibility of a reference count overflow, which will avoid reclaiming the memory if unexpected reference count behaviour is seen.

vijaysun-omr:
Could'nt there be some issue with the add routine as well ? It might overwrite an earlier entry ?

ncough:
Better bound management has been added to the `add` and the `init` that will avoid invalid memory accesses in the event of overflows.

vijaysun-omr:
Do you need a boolean value returned by add ? How would a caller realize that the addition was not done (I think this is needed to add the assumption separately if the add did not succeed) ?

ncough:
Whilst `size_t` is only specified to be at least 16 bit, it will be at least 32 bit on all platforms currently supported by OMR. At this size, the possibility of overflow is unlikely without hitting another form of constraint. Therefore, the overflow checks on reference count and list size have been removed.

andrewcraik:
maybe the checks should be in for debug?

ncough:
A subset of the checks remain under debug, which will assert if an overflow has been detected.

Issue #2015, opened 2017-11-16
Build Failure with Xcode 9


mgaudet:
This is new instance of a problem we'd addressed before in #1707, and puts more pressure on #1758 

```
14:07 $ ninja -j1 comptest
[1/66] Building CXX object jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/optimizer/Structure.cpp.o
FAILED: jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/optimizer/Structure.cpp.o 
/usr/local/opt/ccache/libexec/c++  -DBITVECTOR_BIT_NUMBERING_MSB -DJ9HAMMER -DJITTEST -DOSX -DPROD_WITH_ASSUMES -DSUPPORTS_THREAD_LOCAL -DTR_HOST_64BIT -DTR_HOST_X86 -DTR_TARGET_64BIT -DTR_TARGET_X86 -DUT_DIRECT_TRACE_REGISTRATION -D_FILE_OFFSET_BITS=64 -I../jitbuilder/x/amd64 -I../jitbuilder/x -I../jitbuilder -I../compiler/x/amd64 -I../compiler/x -I../compiler -I../ -I. -I.././include -I.././include_core -I.././third_party -Wno-write-strings -std=c++0x -g -fPIC   -pthread -m64 -MD -MT jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/optimizer/Structure.cpp.o -MF jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/optimizer/Structure.cpp.o.d -o jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/optimizer/Structure.cpp.o -c ../compiler/optimizer/Structure.cpp
In file included from ../compiler/optimizer/Structure.cpp:33:
In file included from ../compiler/il/AliasSetInterface.hpp:28:
In file included from ../compiler/compile/SymbolReferenceTable.hpp:25:
In file included from ../compiler/compile/OMRSymbolReferenceTable.hpp:37:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/map:820:5: error: static_assert failed "Allocator::value_type must be same type as value_type"
    static_assert((is_same<typename allocator_type::value_type, value_type>::value),
[...]

PR #2031, opened 2017-11-20
The std::map value pairs have const keys

rwy0717:
Update two places where we're creating an allocator that allocates the wrong type. On recent mac OS versions, this fixes a compilation failure.


0xdaryl:
Is this a fix for #2015?

0xdaryl:
genie-omr build all

rwy0717:
Yes, it looks like it is. I came across this while moving base9 to the newest OMR.

Issue #2035, opened 2017-11-21
Fix GRA for Vector registers on 32bit x86


IBMJimmyk:
GRA for vector registers on 32bit x86 is broken. It attempts to use an ST real register to hold a value from a 128 bit TR_VRF virtual register. The generated instructions actually expect xmm real registers to be assigned to TR_VRF virtual registers. Assigning incompatible registers to an instruction ends up confusing binary encoding leading to incorrect code generation.

This problem is specific to 32bit x86. On 64bit, GRA assigns xmm registers as expected.

IBMJimmyk:
#2036 Disables vector register GRA on 32bit x86. It also disables autosimd reduction operations on 32bit x86. These features will need to be reenabled after the GRA problem in this issue has been resolved.

PR #2036, opened 2017-11-21
Disable vector register GRA and reduction on 32bit x86

IBMJimmyk:
GRA does not properly assign xmm real registers to TR_VRF virtual
registers on 32bit x86. This can lead to code that produces an incorrect
result. GRA for vector registers is temporarily being disabled on 32bit
x86 until this issue is resolved.

The removal of GRA may cause a performance problem in autosimd reductions
so reductions have also been disabled. This was done by temporarily
disabling the use of getvelem in autosimd on 32bit x86 which is needed
for reduction.

Issue: #2035

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -299,8 +299,16 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    self()->setLastGlobalGPR(self()->machine()->getLastGlobalGPRRegisterNumber());
    self()->setLast8BitGlobalGPR(self()->machine()->getLast8BitGlobalGPRRegisterNumber());
    self()->setLastGlobalFPR(self()->machine()->getLastGlobalFPRRegisterNumber());
-   self()->setFirstGlobalVRF(self()->getFirstGlobalFPR());
-   self()->setLastGlobalVRF(self()->getLastGlobalFPR());
+   /*
+    * GRA does not work with vector registers on 32 bit due to a bug where xmm registers are not being assigned.
+    * This disables GRA for vector registers on 32 bit.
+    * This code will be reenabled as part of Issue 2035 which tracks the progress of fixing the GRA bug.
+    */
+   if (TR::Compiler->target.is64Bit())
+      {
+      self()->setFirstGlobalVRF(self()->getFirstGlobalFPR());
+      self()->setLastGlobalVRF(self()->getLastGlobalFPR());
+      }
 
    // Initialize Linkage for Code Generator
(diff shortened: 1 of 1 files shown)


compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -299,8 +299,15 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    self()->setLastGlobalGPR(self()->machine()->getLastGlobalGPRRegisterNumber());
    self()->setLast8BitGlobalGPR(self()->machine()->getLast8BitGlobalGPRRegisterNumber());
    self()->setLastGlobalFPR(self()->machine()->getLastGlobalFPRRegisterNumber());
-   self()->setFirstGlobalVRF(self()->getFirstGlobalFPR());
-   self()->setLastGlobalVRF(self()->getLastGlobalFPR());
+   /*
+    * GRA does not work with vector registers on 32 bit due to a bug where xmm registers are not being assigned
+    * This will temporary disable GRA for vector registers on 32 bit until the problem has been resolved
```
0xdaryl: * Please use periods at the end of your sentences rather than just line breaks.
* Rather than say this fix is temporary, please reference the issue you created to track the resolution of the problem in the comment instead.  Part of the work in that issue should be to re-enable this code and remove the comment.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -1027,11 +1034,21 @@ OMR::X86::CodeGenerator::getSupportsOpCodeForAutoSIMD(TR::ILOpCode opcode, TR::D
...
-      case TR::getvelem:
          if (dt == TR::Int32 || dt == TR::Int64 || dt == TR::Float || dt == TR::Double)
             return true;
          else
             return false;
+      /*
+       * GRA does not work with vector registers on 32 bit due to a bug where xmm registers are not being assigned
+       * This can potentially cause a performance problem in autosimd reductions
+       * This function is where AutoSIMD checks to see if getvelem is suppored for use in reductions
+       * getvelem is temporarily marked as unsupported until the GRA problem has been fixed.
```
0xdaryl: * Please use periods at the end of your sentences.

* Rather than say this fix is temporary, please reference the issue you created to track the resolution of the problem in the comment instead.  Part of the work in that issue should be to re-enable this code and remove the comment.

0xdaryl:
Genie build is not required because an x86-only file was modified.

IBMJimmyk:
I fixed the concerns mentioned in the previous comments. I fixed the punctuation in the code comments and added references to Issue #2035 and the conditions for reenabling the code.

PR #2047, opened 2017-11-23
Add support for detecting docker container

ashu-mehra:
Add support in port library for detecting if we are running in
docker container or not.
This is required for cgroup related APIs in port library,
such as `omrsysinfo_cgroup_get_memlimit`, to work properly.
This is because when running in a docker container,
the cgroups listed in `/proc/self/cgroup` do not match the cgroups
of the process in the container. `/proc/self/cgroup` actually
shows the cgroups as seen from the host.

Eg if a process is running in a docker container,
its `/proc/PID/cgroup` on docker will show something like:
```
$ cat /proc/8364/cgroup
11:cpuset:/docker/<container id>
10:perf_event:/docker/<container id>
9:memory:/dockera/<container id>
...
```

but the actual cgroup of the process would be "/".

Once we figure out that we are running in docker container,
then cgroup APIs in port library should use "/" as cgroup name
for reading resource limits.

Note that as and when docker containers start supporting
cgroup namespace, then above mentioned mismatch would also be fixed.
Following issues and PRs are related to adding support for
cgroup namespace in docker container:
https://github.com/opencontainers/runc/pull/1184
https://github.com/opencontainers/runtime-spec/issues/62
https://github.com/opencontainers/runtime-spec/issues/66

There is no direct mechanism to detect if we are running in
docker container.
Most reliable mechanism that I have come across is
by checking the cgroup for PID 1 (by reading `/proc/1/cgroup` file)
for all subsystems. If all cgroups are "/", then it indicates the
process is running on host system.
Else we conclude it to be running in a docker container.
Note that this mechanism will break when docker containers start
supporting cgroup namespace.

### DIFF
--- a/port/common/omrsysinfo.c
+++ b/port/common/omrsysinfo.c
@@ -902,6 +902,7 @@ omrsysinfo_cgroup_are_subsystems_enabled(struct OMRPortLibrary *portLibrary, uin
  * omrsysinfo_cgroup_enable_limits() before calling this function.
  * When the fuction returns OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM,
  * value of *limits is unspecified.
+ * Note that 'limit' parameter must not be NULL.
  *
  * @param[in] portLibrary pointer to OMRPortLibrary
  * @param[out] limit pointer to uint64_t which successful return contains memory limit imposed by cgroup
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -256,8 +256,17 @@ struct  {
 #if defined(LINUX)
 
 #define OMR_CGROUP_V1_MOUNT_POINT "/sys/fs/cgroup"
+#define ROOT_CGROUP "/"
 
-/* Currently 12 subsystems/resource controllers are defined.
+/* An entry in /proc/<pid>/cgroup is of following form:
+ * 	<hierarchy ID>:<subsystem>[,<subsystem>]*:<cgroup name>
+ *
+ * An example:
+ * 	7:cpuacct,cpu:/mycgroup
+ */
+#define PROC_PID_CGROUP_ENTRY_FORMAT "%d:%[^:]:%s"
+
+/* Currently 12 subsystems or resource controllers are defined.
  */
 typedef enum OMRCgroupSubsystem {
 		INVALID_SUBSYSTEM = -1,
@@ -331,12 +340,14 @@ static uint16_t getPhysicalMemory();
 #endif /* defined(OMRZTPF) */
 
 #if defined(LINUX) && !defined(OMRZTPF)
+static BOOLEAN isCgroupV1Available(struct OMRPortLibrary *portLibrary);
 static void freeCgroupEntries(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cgEntryList);
 static char * getCgroupNameForSubsystem(struct OMRPortLibrary *portLibrary, OMRCgroupEntry *cgEntryList, const char *subsystem);
 static int32_t addCgroupEntry(struct OMRPortLibrary *portLibrary, OMRCgroupEntry **cgEntryList, int32_t hierId, const char *subsystem, const char *cgroupName);
-static int32_t readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgroupEntryList, uint64_t *availableSubsystems);
+static int32_t readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, BOOLEAN inContainer, OMRCgroupEntry **cgroupEntryList, uint64_t *availableSubsystems);
 static OMRCgroupSubsystem getCgroupSubsystemFromFlag(uint64_t subsystemFlag);
 static int32_t readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFlag, const char *fileName, int32_t numItemsToRead, const char *format, ...);
+static int32_t isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer);
 #endif /* defined(LINUX) */
 
 
@@ -1701,7 +1712,7 @@ omrsysinfo_startup(struct OMRPortLibrary *portLibrary)
 #if defined(LINUX) && !defined(OMRZTPF)
 	PPG_cgroupEntryList = NULL;
 	/* To handle the case where multiple port libraries are started and shutdown,
-	 * as done by some fvtests (eg fvtest/porttest/j9portTest.cpp) that create fake portlibrary 
+	 * as done by some fvtests (eg fvtest/porttest/j9portTest.cpp) that create fake portlibrary
 	 * to test its management and lifecycle,
 	 * we need to ensure globals like cgroupEntryListMonitor are initialized and destroyed only once.
 	 */
@@ -3316,6 +3327,34 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
 
 /** 
  * @internal
+ * Checks if cgroup v1 system is available
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ *
+ * @return TRUE if cgroup v1 system is available, FALSE otherwise 
+ */
+static BOOLEAN 
+isCgroupV1Available(struct OMRPortLibrary *portLibrary)
+{
+	struct statfs buf = {0};
+	int32_t rc = 0;
+	BOOLEAN result = TRUE;
+
+	/* If tmpfs is mounted on /sys/fs/cgroup, then it indicates cgroup v1 system is available */
+	rc = statfs(OMR_CGROUP_V1_MOUNT_POINT, &buf);
+	if (0 != rc) {
+		portLibrary->error_set_last_error(portLibrary, errno, OMRPORT_ERROR_SYSINFO_SYS_FS_CGROUP_STATFS_FAILED);
+		result = FALSE;
+	} else if (TMPFS_MAGIC != buf.f_type) {
+		portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_SYS_FS_CGROUP_TMPFS_NOT_MOUNTED, "tmpfs is not mounted on " OMR_CGROUP_V1_MOUNT_POINT);
+		result = FALSE;
+	}
+
+	return result;
+}
+
+/** 
+ * @internal
  * Free resources allocated for OMRCgroupEntry
  *
  * @param[in] portLibrary pointer to OMRPortLibrary
@@ -3420,21 +3459,20 @@ _end:
  *
  * @param[in] portLibrary pointer to OMRPortLibrary
  * @param[in] pid process id
+ * @param[in] inContainer if set to TRUE then ignore cgroup in /proc/<pid>/cgroup and use ROOT_CGROUP instead  
  * @param[out] cgroupEntryList pointer to OMRCgroupEntry *. On successful return, *cgroupEntry
  * points to a circular linked list. Each element of the list is populated based on the contents 
  * of /proc/<pid>/cgroup file.
- * @param[out] availableSubsystems on successful return, contains bitwise-OR of flags of type OMR_CGROUP_SUBSYSTEMS_* 
+ * @param[out] availableSubsystems on successful return, contains bitwise-OR of flags of type OMR_CGROUP_SUBSYSTEMS_*
  * indicating the subsystems available for use
  *
  * returns 0 on success, negative code on error
  */
 static int32_t
-readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgroupEntryList, uint64_t *availableSubsystems)
+readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, BOOLEAN inContainer, OMRCgroupEntry **cgroupEntryList, uint64_t *availableSubsystems)
 {
-	char cgroup[PATH_MAX];
+	char cgroupFilePath[PATH_MAX];
 	uintptr_t requiredSize = 0; 
-	/* This array should be large enough to read names of all subsystems. 1024 should be enough. */
-	char subsystems[1024];
 	FILE *cgroupFile = NULL;
 	OMRCgroupEntry *cgEntryList = NULL;
 	uint64_t available = 0;
@@ -3444,34 +3482,29 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
 	
 	requiredSize = portLibrary->str_printf(portLibrary, NULL, (uint32_t)-1, "/proc/%d/cgroup", pid);
 	Assert_PRT_true(requiredSize <= PATH_MAX);
-	portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
-	cgroupFile = fopen(cgroup, "r");
+	portLibrary->str_printf(portLibrary, cgroupFilePath, sizeof(cgroupFilePath), "/proc/%d/cgroup", pid);
+
+	/* Even if 'inContainer' is TRUE, we need to parse the cgroup file to get the list of subsystems */
+	cgroupFile = fopen(cgroupFilePath, "r");
 	if (NULL == cgroupFile) {
 		rc = portLibrary->error_set_last_error(portLibrary, errno, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_FOPEN_FAILED);
 		goto _end;
 	}
 
 	while (0 == feof(cgroupFile)) {
+		char cgroup[PATH_MAX];
+		/* This array should be large enough to read names of all subsystems. 1024 should be enough based on current supported subsystems. */
+		char subsystems[1024];
 		char *cursor = NULL;
 		char *separator = NULL;
 		int32_t hierId = -1;
 
-		/* Following is the description of /proc/<pid>/cgroup copied from 'man' page for proc:
- 		 *
- 		 * Each entry in /proc/<pid>/cgroup is of type:
-		 * 	5:cpuacct,cpu,cpuset:/daemons
-		 * The colon-separated fields are, from left to right:
-		 * 	1. hierarchy ID number
-		 * 	2. set of subsystems bound to the hierarchy
-		 * 	3. control group in the hierarchy to which the process belongs
-		 */
-		rc = fscanf(cgroupFile, "%d:%[^:]:%s", &hierId, subsystems, cgroup);
+		rc = fscanf(cgroupFile, PROC_PID_CGROUP_ENTRY_FORMAT, &hierId, subsystems, cgroup);
 		/* Ensure we didn't overflow */
 		Assert_PRT_true(strlen(subsystems) < 1024);
 		Assert_PRT_true(strlen(cgroup) < PATH_MAX);
 
 		if (EOF == rc) {
-			rc = 0;
 			break;
 		} else if (3 != rc) {
 			rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_READ_FAILED, "unexpcted format of /proc/%d/cgroup file", pid);
@@ -3490,7 +3523,12 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
 				if (OMR_ARE_NO_BITS_SET(available, supportedSubsystems[i].flag)
 					&& !strcmp(cursor, supportedSubsystems[i].name)
 				) {
-					rc = addCgroupEntry(portLibrary, &cgEntryList, hierId, cursor, cgroup);
+					const char *cgroupToUse = cgroup;
+
+					if (TRUE == inContainer) {
+						cgroupToUse = ROOT_CGROUP;
+					}
+					rc = addCgroupEntry(portLibrary, &cgEntryList, hierId, cursor, cgroupToUse);
 					if (0 != rc) {
 						goto _end;
 					}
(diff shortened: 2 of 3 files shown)


ashu-mehra:
@DanHeidinga can you please review this.

ashu-mehra:
Sorry, last minute changes resulted in those failures. I have fixed them. It should pass now.

charliegracie:
genie-omr build all

ashu-mehra:
@DanHeidinga can you please review this.

port/unix/omrsysinfo.c
```diff
@@ -3603,6 +3650,56 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+	*containerType = OMR_CONTAINER_NONE;
+
+	if (isCgroupV1Available(portLibrary)) {
+		/* Read PID 1's cgroup file /proc/1/cgroup and check cgroup name for each subsystem.
+		 * If cgroup name for each subsystem points to the root cgroup "/",
+		 * then the process is not running in a container.
+		 * For any other cgroup name, assume we are in a 'docker' container.
+		 */
+		rc = readCgroupFile(portLibrary, 1, FALSE, &cgEntryList, NULL);
+		if (0 != rc) {
```
DanHeidinga: The `goto` is unnecessary if the `!=` is changed to `==` and the code below is moved into the if block.

port/unix/omrsysinfo.c
```diff
@@ -3613,25 +3710,23 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
-			goto _end;
+		if (isCgroupV1Available(portLibrary)) {
+			rc = portLibrary->sysinfo_get_container_type(portLibrary, &PPG_containerType);
+			if (0 != rc) {
+				goto _end;
+			}
+			omrthread_monitor_enter(cgroupEntryListMonitor);
+			if (NULL == PPG_cgroupEntryList) {
+				if (OMR_CONTAINER_DOCKER == PPG_containerType) {
+					rc = readCgroupFile(portLibrary, getpid(), TRUE, &PPG_cgroupEntryList, &PPG_cgroupSubsystemsAvailable);
```
DanHeidinga: A local variable for the `TRUE` vs `FALSE` would be clearer rather than duplicating the `readCgroupFile` call

port/unix/omrsysinfo.c
```diff
@@ -3613,25 +3710,23 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
-		}
-
-		omrthread_monitor_enter(cgroupEntryListMonitor);
-		if (NULL == PPG_cgroupEntryList) {
-			rc = readCgroupFile(portLibrary, getpid(), &PPG_cgroupEntryList, &PPG_cgroupSubsystemsAvailable);
-		}
-		omrthread_monitor_exit(cgroupEntryListMonitor);
-		if (0 != rc) {
-			goto _end;
+		if (isCgroupV1Available(portLibrary)) {
```
DanHeidinga: Do I read this code correctly that the cgroup detection is only possible on cgroup v1 systems?

port/unix/omrsysinfo.c
```diff
@@ -3603,6 +3650,56 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+			if (strcmp(ROOT_CGROUP, cgroupName)) {
+				*containerType = OMR_CONTAINER_DOCKER;
+				break;
+			}
+			temp = temp->next;
+		} while (temp != cgEntryList);
+	}
+_end:
+	if (NULL != cgEntryList) {
+		freeCgroupEntries(portLibrary, cgEntryList);
```
DanHeidinga: This seems allocation heavy for the information being gathered.  Can this be done without allocating and freeing the cgroup list?

include_core/omrport.h
```diff
@@ -1012,6 +1012,13 @@ typedef struct OMROSKernelInfo {
 #define OMR_CGROUP_SUBSYSTEM_MEMORY ((uint64_t)0x2)
 #define OMR_CGROUP_SUBSYSTEM_ALL (OMR_CGROUP_SUBSYSTEM_CPU | OMR_CGROUP_SUBSYSTEM_MEMORY)
 
+typedef enum OMRContainerType {
+	OMR_CONTAINER_NONE,
+	OMR_CONTAINER_DOCKER,
```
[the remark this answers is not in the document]
ashu-mehra: I checked `runc` container and it displays same behavior with respect to `/proc/pid/cgroup` file. So I believe the detection mechanism should work for any `runc`-style container, not just Docker.
Since Docker is most popular system at the moment, I named it so.
Do you want to rename it as something like `OMR_CONTAINER_RUNC`?

port/unix/omrsysinfo.c
```diff
@@ -3431,6 +3472,7 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
 	requiredSize = portLibrary->str_printf(portLibrary, NULL, (uint32_t)-1, "/proc/%d/cgroup", pid);
 	Assert_PRT_true(requiredSize <= PATH_MAX);
 	portLibrary->str_printf(portLibrary, cgroup, sizeof(cgroup), "/proc/%d/cgroup", pid);
+
```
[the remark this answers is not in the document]
ashu-mehra: `cgroup` at this point is storing path `/proc/pid/cgroup`. Same var is also used for storing cgroup name by parsing `/proc/pid/cgroup` file. This may be the cause of confusion. I will update code to rename the variable.
Use of `ROOT_CGROUP` is intentional as we don't want to use the cgroup name we get from `/proc/pid/cgroup` file.

port/unix/omrsysinfo.c
ashu-mehra: For cgroup v2, the file system type mounted on `/sys/fs/cgroup` is different. The function `isCgroupV1Available()` is currently only handling cgroup v1, therefore I named it so.
As and when we add support for cgroup v2, I believe we can rename the function then.

port/unix/omrsysinfo.c
```diff
@@ -3735,6 +3831,48 @@ omrsysinfo_cgroup_get_memlimit(struct OMRPortLibrary *portLibrary, uint64_t *lim
...
+	int32_t rc = 0;
+
+	Assert_PRT_true(NULL != containerType);
+
+	/* Set default value */
+	*containerType = OMR_CONTAINER_NONE;
+
+#if defined(LINUX) && !defined(OMRZTPF)
+	{
+		const char *envValue = getenv("CONTAINER_TYPE");
```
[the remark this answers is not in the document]
ashu-mehra: Its a fallback mechanism for current detection logic. As I mentioned in the commit message, container detection logic is flaky and can break in future. If that happens, then we can use this env variable to indicate to runtime that its running in a container.
For example, we can set this env variable in Dockerfile to enable container awareness in the runtime.

ashu-mehra:
> I'm not sure I follow the need for (2). Can you clarify why these apis are useful and what port library consumers would do with this info?

I thought it may come handy if runtimes want to display information about the container they are running in for diagnostic purpose. For example, OpenJ9 can display the container type in javacore?

port/unix/omrsysinfo.c
ashu-mehra: It looks possible. Let me see if I can refactor it to make it light weight.

ashu-mehra:
@DanHeidinga I have updated the code as per the comments. Can you please take a look.

ashu-mehra:
> This is a little hard to review as it mixes together two things:
> 1. fixing a bug(?) around detecting and handling non-v1 cgroup support

I should point out that the change is not to handle non-v1 cgroup support. Changes are mainly to  make cgroup support work properly in (docker) container by detecting that we are in the container environment.

ashu-mehra:
@DanHeidinga can you please review the updated change set.

mstoodle:
another ping for @DanHeidinga who maybe missed the first one due to conference circuit prep distractions :) .

DanHeidinga:
Thanks @mstoodle and @ashu-mehra - My goal is to have reviewed this by end of week (hold me to it!)

include_core/omrport.h
```diff
@@ -1463,6 +1471,10 @@ typedef struct OMRPortLibrary {
 	uint64_t ( *sysinfo_cgroup_are_subsystems_enabled)(struct OMRPortLibrary *portLibrary, uint64_t subsystemFlags);
 	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_get_memlimit "omrsysinfo_cgroup_get_memlimit"*/
 	int32_t (*sysinfo_cgroup_get_memlimit)(struct OMRPortLibrary *portLibrary, uint64_t *limit);
+	/** see @ref omrsysinfo_get_container_type "omrsysinfo_get_container_type"*/
+	int32_t ( *sysinfo_get_container_type)(struct OMRPortLibrary *portLibrary, OMRContainerType *containerType);
+	/** see @ref omrsysinfo_get_container_name "omrsysinfo_get_container_name"*/
+	const char * ( *sysinfo_get_container_name)(struct OMRPortLibrary *portLibrary, OMRContainerType containerType);
```
DanHeidinga: These two APIs don't belong in this changeset.  Given the VM only supports detecting linux cgroups or nothing, we don't have enough examples to generalize to a good portable api.  

I would rather forgo these APIs now and add them when we have second (or third) container system to work with.  That will result in a better designed api that is easier to maintain longterm.

port/unix/omrsysinfo.c
```diff
@@ -256,6 +256,18 @@ struct  {
 #if defined(LINUX)
 
 #define OMR_CGROUP_V1_MOUNT_POINT "/sys/fs/cgroup"
+#define ROOT_CGROUP "/"
+
+/* Following is the description of /proc/<pid>/cgroup copied from 'man' page for proc:
```
DanHeidinga: We can't copy from the `man` page for proc as the documentation is under GPLv2: http://man7.org/linux/man-pages/man5/proc.5.license.html

port/unix/omrsysinfo.c
```diff
@@ -299,6 +311,12 @@ struct {
 	{ "memory", OMR_CGROUP_SUBSYSTEM_MEMORY }
 };
 
+const char *OMRContainerNames[OMR_CONTAINER_NUM_CONTAINERS] = {
```
DanHeidinga: As per comment above, this belongs in a separate changeset or should wait until we have more examples of container systems to build the api on.

port/unix/omrsysinfo.c
```diff
@@ -3300,6 +3320,37 @@ omrsysinfo_os_kernel_info(struct OMRPortLibrary *portLibrary, struct OMROSKernel
...
+		portLibrary->error_set_last_error(portLibrary, errno, OMRPORT_ERROR_SYSINFO_SYS_FS_CGROUP_STATFS_FAILED);
+		result = FALSE;
+		goto _end;
+	} else if (TMPFS_MAGIC != buf.f_type) {
+		portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_SYS_FS_CGROUP_TMPFS_NOT_MOUNTED, "tmpfs is not mounted on " OMR_CGROUP_V1_MOUNT_POINT);
+		result = FALSE;
+		goto _end;
+	}
+
+_end:
```
DanHeidinga: label can be removed.

port/unix/omrsysinfo.c
```diff
@@ -3603,6 +3652,69 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+ * Return container type
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[out] containerType pointer to OMRContainerType which on successful return contains the
+ *                           container type in which the process is running.
+ *
+ * @return 0 on success, otherwise negative error code
+ */
+static int32_t
+getContainerType(struct OMRPortLibrary *portLibrary, OMRContainerType *containerType)
```
DanHeidinga: Please remove this method from this PR.

port/unix/omrsysinfo.c
```diff
@@ -3613,25 +3725,24 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
 
-		omrthread_monitor_enter(cgroupEntryListMonitor);
-		if (NULL == PPG_cgroupEntryList) {
-			rc = readCgroupFile(portLibrary, getpid(), &PPG_cgroupEntryList, &PPG_cgroupSubsystemsAvailable);
-		}
-		omrthread_monitor_exit(cgroupEntryListMonitor);
-		if (0 != rc) {
-			goto _end;
+				if (OMR_CONTAINER_DOCKER == PPG_containerType) {
+					useRootCgroup = TRUE;
```
DanHeidinga: This will need to be reworked when the container type api is moved out.

port/unix/omrsysinfo.c
```diff
@@ -3706,13 +3817,14 @@ int32_t
 omrsysinfo_cgroup_get_memlimit(struct OMRPortLibrary *portLibrary, uint64_t *limit)
 {
 	int32_t rc = OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;
+
+	Assert_PRT_true(NULL != limit);
```
DanHeidinga: Please update the function prototype to indicate `limit` cannot be null.

port/unix/omrsysinfo.c
ashu-mehra: I updated the comment to put it in my own words instead of copying from `man` page.

port/unix/omrsysinfo.c
ashu-mehra: Done

ashu-mehra:
@DanHeidinga I have updated the change set to remove `sysinfo_get_container_type` and `sysinfo_get_container_name` APIs.
Other comments are also addressed. Please review the updated change set.

port/common/omrsysinfo.c
```diff
@@ -902,6 +902,7 @@ omrsysinfo_cgroup_are_subsystems_enabled(struct OMRPortLibrary *portLibrary, uin
  * omrsysinfo_cgroup_enable_limits() before calling this function.
  * When the fuction returns OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM,
  * value of *limits is unspecified.
+ * Note that 'limit' parameter should not be NULL.
```
DanHeidinga: Nitpick: `should not` -> `must not`  

This is a bit of a nitpick but `must` is a stronger requirement than `should`.  Since the code will assert it is not null, `must` is a better choice.

port/win32/omrsysinfo.c
```diff
@@ -1653,3 +1653,4 @@ omrsysinfo_cgroup_get_memlimit(struct OMRPortLibrary *portLibrary, uint64_t *lim
 {
 	return OMRPORT_ERROR_SYSINFO_CGROUP_UNSUPPORTED_PLATFORM;
 }
+
```
DanHeidinga: Can you remove this file from the PR?  It doesn't have any real changes.

port/unix/omrsysinfo.c
```diff
@@ -3444,34 +3483,28 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, OMRCgroupEntry **cgr
...
+	/* Even if 'useRootCgroup' is TRUE, we need to parse the cgroup file to get the list of subsystems */
+	cgroupFile = fopen(cgroupFilePath, "r");
 	if (NULL == cgroupFile) {
 		rc = portLibrary->error_set_last_error(portLibrary, errno, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_FOPEN_FAILED);
 		goto _end;
 	}
 
 	while (0 == feof(cgroupFile)) {
+		char cgroup[PATH_MAX];
+		char subsystems[1024];
```
DanHeidinga: Can you add a comment that points to the `subsystemNames` array to indicate why `1024` is likely sufficient?

port/unix/omrsysinfo.c
```diff
@@ -3617,6 +3655,63 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+			rc = fscanf(cgroupFile, PROC_PID_CGROUP_ENTRY_FORMAT, &hierId, subsystems, cgroup);
+			/* Ensure we didn't overflow */
+			Assert_PRT_true(strlen(subsystems) < 1024);
+			Assert_PRT_true(strlen(cgroup) < PATH_MAX);
+
+			if (EOF == rc) {
+				break;
+			} else if (3 != rc) {
+				rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_READ_FAILED, "unexpected format of /proc/1/cgroup file");
+				goto _end;
```
DanHeidinga: `@return 0 on success, otherwise negative error code`

This may exit the method with a positive, non-zero return value contrary to the function contract.

port/unix/omrsysinfo.c
```diff
@@ -3617,6 +3655,63 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+			}
+
+			if (strcmp(ROOT_CGROUP, cgroup)) {
```
DanHeidinga: Can you make this a boolean comparison and explicitly add the `!= 0` check to the if statement?  The port library (or more specifically, new code) tries to ensure that all comparisons result in boolean values.

port/unix/omrsysinfo.c
```diff
@@ -3617,6 +3655,63 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+/** 
+ * Checks if the process is running inside docker container 
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[out] inDocker pointer to BOOLEAN which on successful return indicates if the process is running in docker container or not 
+ *
+ * @return 0 on success, otherwise negative error code
+ */
+static int32_t
+isRunningInDocker(struct OMRPortLibrary *portLibrary, BOOLEAN *inDocker)
```
DanHeidinga: Should this be `isRunningInContainer`?  It appears to generically check for a container runtime, not docker specifically.

port/unix/omrsysinfo.c
```diff
@@ -3617,6 +3655,63 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
 	return rc;
 }
 
+/** 
+ * Checks if the process is running inside docker container 
+ *
+ * @param[in] portLibrary pointer to OMRPortLibrary
+ * @param[out] inDocker pointer to BOOLEAN which on successful return indicates if the process is running in docker container or not
```
DanHeidinga: Why not return `TRUE` if running in a container and `FALSE` for not / errors?  

The current approach means user code needs to check two things:
```
  BOOLEAN inDocker = false;
   if (0 == isRunningInDocker(portLib, &inDocker)) {
      if (inDocker == TRUE) {
        /* take action */
      }
    }
```
Is there a meaningful distinction between error and not in docker?

port/unix/omrsysinfo.c
```diff
@@ -3627,25 +3722,26 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
-			rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_SYS_FS_CGROUP_TMPFS_NOT_MOUNTED, "tmpfs is not mounted on " OMR_CGROUP_V1_MOUNT_POINT);
-			goto _end;
-		}
+			rc = isRunningInDocker(portLibrary, &inDocker);
+			if (0 != rc) {
+				goto _end;
+			}
+			omrthread_monitor_enter(cgroupEntryListMonitor);
+			if (NULL == PPG_cgroupEntryList) {
+				BOOLEAN useRootCgroup = FALSE;
```
DanHeidinga: Is `useRootCgroup` required or can the code pass `inDocker` instead?

port/unix/omrsysinfo.c
ashu-mehra: `rc` is reassigned return value of `error_set_last_error_with_message_format` which is negative error code.

port/unix/omrsysinfo.c
ashu-mehra: I am inclined to make this distinction between error and not in docker.
Its caller `omrsysinfo_cgroup_is_system_available` should not attempt to do any processing if any error is encountered.
If we treat error case same as 'not in docker' case, then we may end up in trouble if the error happened when its actually running in docker. In such case it would attempt to read non-existing cgroup control files for determining resource limits.
To be on safer side, if any error occurs, better to not use cgroup limits at all and fallback to regular behavior.

port/unix/omrsysinfo.c
ashu-mehra: `isRunningInContainer` is also fine. The mechanism works for runc based containers, not specific to docker.

port/unix/omrsysinfo.c
ashu-mehra: We can use `inDocker` instead.

ashu-mehra:
@DanHeidinga change set is updated. Please review.

port/unix/omrsysinfo.c
DanHeidinga: I missed that - sorry.

DanHeidinga:
@ashu-mehra I've applied these changes to an OpenJ9 build and added some debug prints but I'm not seeing any output.  Is there a commandline flag / env var I need to set to enable cgroup detection?

port/unix/omrsysinfo.c
```diff
@@ -3627,25 +3722,21 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
-		}
-
-		omrthread_monitor_enter(cgroupEntryListMonitor);
-		if (NULL == PPG_cgroupEntryList) {
-			rc = readCgroupFile(portLibrary, getpid(), &PPG_cgroupEntryList, &PPG_cgroupSubsystemsAvailable);
-		}
-		omrthread_monitor_exit(cgroupEntryListMonitor);
-		if (0 != rc) {
-			goto _end;
+			rc = isRunningInContainer(portLibrary, &inContainer);
```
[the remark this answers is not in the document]
ashu-mehra: I noticed that but didn't bother as `isCgroupV1Available` just adds an overhead of one syscall.
But if we want to common it out, how about calling the new function `checkCgroupPrereqs(&cgroupVersion, &inContainer)` where `cgroupVersion` would be an `enum 
 { V1, V2 }` (currently the code only supports V1, but sooner than later we would have to consider V2 as well).

port/unix/omrsysinfo.c
```diff
@@ -3627,25 +3722,21 @@ omrsysinfo_cgroup_is_system_available(struct OMRPortLibrary *portLibrary)
...
+			if (0 != rc) {
+				goto _end;
```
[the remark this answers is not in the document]
ashu-mehra: >  we should capture some diagnostics to know why we failed to determine if we were in a container or not.

I agree. I actually want to add tracepoints in the code at various error conditions and thought of taking that up it in a separate PR. Is that ok?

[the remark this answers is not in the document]
DanHeidinga: Works for me.

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: Since we'll have to touch this code again to add V2 support, I'm OK with leaving it as is now.

charliegracie:
@genie-omr build all zos

PR #2051, opened 2017-11-23
Add a query for targeted inlining

liqunl:
Add an extension point in InlinerUtil to answer whether targeted
inlining is needed for a method after it is inlined.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -5422,17 +5422,11 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
    updateCallersFlags(callerSymbol, calleeSymbol, _optimizer);
 
 
-   if (calleeSymbol->getMethod()->isArchetypeSpecimen() || calleeSymbol->hasMethodHandleInvokes())
+   if (getUtil()->needTargetedInlining(calleeSymbol))
       {
-      // Trees from archetype specimens may not match the archetype method's bytecodes,
-      // so there may be some calls things that inliner missed.
-      //
-      // Tactically, we also inline again based on hasMethodHandleInvokes because EstimateCodeSize
-      // doesn't yet cope with invokeHandle, invokeHandleGeneric, and invokeDynamic (but it should).
-      //
       _optimizer->setRequestOptimization(OMR::methodHandleInvokeInliningGroup);
       if (comp()->trace(OMR::inlining))
-         heuristicTrace(tracer(),"Requesting another pass of MethodHandle invoke inlining due to %s\n", tracer()->traceSignature(calleeSymbol));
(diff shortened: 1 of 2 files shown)


liqunl:
@andrewcraik Can you review this change? I'll unWIP the PR once my pbuild passes.

andrewcraik:
Looks good to me

compiler/optimizer/Inliner.cpp
```diff
@@ -6376,6 +6376,15 @@ OMR_InlinerUtil::computePrexInfo(TR_CallTarget *target)
    return NULL;
    }
 
+/**
+ * Check if targeted inlining is needed if callee is inlined
```
0xdaryl: What is "targeted inlining" ?  A description of that is needed in this comment if this is going to be a useful API.

liqunl: I've made the change, could you continue the review?

compiler/optimizer/Inliner.cpp
```diff
@@ -5422,17 +5422,26 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
...
+
+    Each method except the last one in the chain usually contains simple code that manipulates the arguments for the next
+    call. The last method in the chain is usually the most important one, it does the real job and can contain arbitrary
+    code.
+
+    The goal of targeted inlining is to inline along the call chain to get the most important method inlined, which is the
+    last one in the chain. It does not care about other callsites.
+
+    Currently, inliner does not have heuristics to examine into each callsite to find the ones with those properties, the
+    JIT developers have to tell it what and when to inline based on the language.
```
0xdaryl: I don't think this sentence is necessary.

compiler/optimizer/Inliner.cpp
```diff
@@ -5422,17 +5422,26 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
...
+    Targeted inlining is meant to deal with method call chains where the receiver of current call stores the information
+    of the next call, which is usually the receiver or the method of the next call. Knowing the receiver of the first call
+    means we can devirtualize all the calls along the chain.
+
+    Each method except the last one in the chain usually contains simple code that manipulates the arguments for the next
+    call. The last method in the chain is usually the most important one, it does the real job and can contain arbitrary
+    code.
+
+    The goal of targeted inlining is to inline along the call chain to get the most important method inlined, which is the
+    last one in the chain. It does not care about other callsites.
```
0xdaryl: Thanks for the detailed description.  Could you move it to `Inliner.hpp` as part of a proper Doxygen comment (which also describes the input arg and the return value) before `needTargetedInlining` ?  I think its better to describe there what the API does so that a potential implementor knows whether it applies to them or not.

liqunl: The comment's been moved to `Inliner.hpp` and is formatted in Doxygen style.

vijaysun-omr:
Looks safe. Merging.

Issue #1830, opened 2017-10-23
Problems building on z/OS


daveyc:
I'm getting the following error building on z/OS. I have set `_CC_CCMODE=1   `

```
FSUM3010 Specify a file with the correct suffix (.C, .hh, .i, .c, .i, .s, .o, .x, .p, .I, or .a), or a corresponding data set name, instead of -L..
gmake[2]: *** [../../hookgen] Error 1
```

mgaudet:
Can you show more of you configuration output.

daveyc:
I didn't supply any options to configure. If you can give me the heads up what I need for a z/OS build it will be greatly received. I've ran ./configure --help and see lots of options but don't have time right now to work it all out.

daveyc:
If I can get the build working I will document it in a readme for z/OS build.

mgaudet:
Did you provide a value for `SPEC`?

daveyc:
No I didn't supply a value for SPEC. I was assuming the configure script would be able to deduce it's a z/OS system by running `uname` and I wouldn't have to specify any arguments or set any environment variables.

daveyc:
Here's the output from configure. OMR_HOST_OS is correctly set to zos which the makefiles use to control the linker.

```
AR='ar'
ARFLAGS=''
AS='as'
CC='cc'
CCLINK='$(CC)'
CCLINKEXE='$(CCLINK)'
CCLINKSHARED='$(CCLINK)'
CFLAGS=''
CPP='cc -E'
CPPFLAGS=''
CXX='xlC_x'
CXXFLAGS=''
CXXLINK='$(CXX)'
CXXLINKEXE='$(CXXLINK)'
CXXLINKSHARED='$(CXXLINK)'
DEFS='-DHAVE_CONFIG_H'
ECHO_C='\c'
ECHO_N=''
ECHO_T=''
EGREP='/u/doc/bin/grep -E'
EXEEXT=''
GLOBAL_ARFLAGS=''
GLOBAL_CFLAGS=''
GLOBAL_CPPFLAGS=''
GLOBAL_CXXFLAGS=''
GLOBAL_INCLUDES=''
GLOBAL_LDFLAGS=''
GLOBAL_LIBPATH=''
GLOBAL_SHARED_LIBS=''
GLOBAL_STATIC_LIBS=''
GREP='/u/doc/bin/grep'
LDFLAGS=''
LIBOBJS=''
LIBS=''
LTLIBOBJS=''
[...]

PR #2053, opened 2017-11-23
Modify makefile rules for zOS

charliegracie:
Currently the OMR zOS makefiles require that a user sets a few 
evnironment variables that control the placement of compile
options. This change uses the default placement of compile options
so that the environment variables are not required.

Remove the LIST option since we do not need to generate the 
assembly listing to stdout.

This change also removes the -g9 option that was added for
generating debug information since it is not valid.

[ci skip] - on zOS changes

### DIFF
--- a/omrmakefiles/rules.zos.mk
+++ b/omrmakefiles/rules.zos.mk
@@ -36,7 +36,6 @@ endif
 
 # Enable Debugging Symbols
 ifeq ($(OMR_DEBUG),1)
-  GLOBAL_FLAGS+=-Wc,g9
 endif
 
 # Enable Optimizations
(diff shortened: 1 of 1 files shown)


charliegracie:
This PR should remove the requirement found in #1830 to have to set the _C*_MODE environment variables.  Next up is to get a .gitattributes file checked in.

omrmakefiles/rules.zos.mk
```diff
@@ -36,7 +36,6 @@ endif
 
 # Enable Debugging Symbols
 ifeq ($(OMR_DEBUG),1)
-  GLOBAL_FLAGS+=-Wc,g9
 endif
```
fjeremic: This `ifeq` is now empty. Can it be folded away?

charliegracie: I was leaving it the same as the 2 above so if we figure out the right options we can add them in.  I can remove it / them if you prefer

charliegracie:
genie-omr buil zos

omrmakefiles/rules.zos.mk
fjeremic: Oh I see. I didn't notice the ones above. I was just curious. I vote to stay consistent then and keep the change as is since `OMR_WARNINGS_AS_ERRORS` and  `OMR_ENHANCED_WARNINGS` `ifeq` checks are empty as well. Thanks for the clarification!

charliegracie:
genie-omr build zos

charliegracie:
we can ignore the CI failure as there are still machine configuration issues and a successful build will require `.gitattributes` to be added.

PR #2054, opened 2017-11-24
Introduce Trampoline exceptions

0xdaryl:
Relocate the Trampoline exceptions from the OpenJ9 project to OMR.
Move them to the `TR` namespace and add some basic documentation.

### DIFF
--- a/compiler/runtime/CodeCacheExceptions.hpp
+++ b/compiler/runtime/CodeCacheExceptions.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2016 IBM Corp. and others
+ * Copyright (c) 2016, 2017 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


0xdaryl:
@dsouzai : would you mind reviewing please?

This is a WIP only while detailed testing is running.

0xdaryl:
I didn't discover eclipse/openj9#669 until now (which has not been merged yet).  My documentation is slightly different than @dsouzai's.  Please let me know if you'd prefer me to use yours instead.

dsouzai:
I think your documentation is fine :).

0xdaryl:
Testing is complete.  Removing WIP.

@mstoodle @vijaysun-omr : please merge.

0xdaryl:
I think the way the RecoverableTrampolineError is currently used in OpenJ9 it is only thrown during the binary encoding phase.  That's probably why it was documented that way in the assert.  I'd prefer to just remove the mentions to binary encoding to make this as general as possible recoverable trampoline exception.  Unless @dsouzai feels otherwise??

dsouzai:
In https://github.com/eclipse/openj9/pull/669 I documented the `J9::RecoverableTrampolineError` exception as:
>Thrown on an Trampoline Failure condition which the compiler can recover from by allowing a recompilation to occur.

even though it is only thrown in Binary Encoding because I figured that that could change in the future. So I agree that it should be made as general as possible.

On that note, you may want to also change https://github.com/eclipse/omr/blob/c624dd395a4baa454c3db16ecff3112a858bc546/compiler/runtime/CodeCacheExceptions.hpp#L45-L46
to be more consistent with your documentation for `TR::RecoverableTrampolineError`, specifically the bit about "recompilation" since your documentation doesn't assume recompilation.

0xdaryl:
Addressed comments.  Please review again.

dsouzai:
LGTM.

mstoodle:
Random question: the two ones you added are "class" whereas the previous one is "struct". There's not much difference, really, but is there a reason to prefer one over the other (or to use both) ?

0xdaryl:
I just cut-pasted from OpenJ9 and in fact didn't notice that distinction at all.  OpenJ9 uses both seemingly indiscriminately.  Given that C++ would make the `what()` function default private, is that useful at all?  Perhaps these should all be `structs`?

I'll change it for this PR, but there's likely a deeper discussion to have.

dsouzai:
>Given that C++ would make the what() function default private, is that useful at all? 
Because the `what()` method is virtual and is in the public scope in `std::exception`, doesn't that mean that the `what()` method of all derived types is also public?

Also, amusingly enough, `std::exception` is defined as a class and not a struct. I think you're right, there should be an issue opened in all relevant projects to determine a consistent story.

0xdaryl:
Pushed again...  Please review.

dsouzai:
:+1:

PR #2057, opened 2017-11-25
Updates for docker based ARM build

JamesKingdon:
configure_linux_arm.mk and rules.linux.mk are updated to make
use of the environment variables provided by the docker file
(https://github.com/eclipse/openj9/pull/622).

This allows the names of the toolchain executables to be passed
from the docker file and also avoids the need to hard-code the
location of the global include files.

Signed-off-by: James Kingdon <jkingdon@ca.ibm.com>

### DIFF
--- a/omrmakefiles/rules.linux.mk
+++ b/omrmakefiles/rules.linux.mk
@@ -104,6 +104,8 @@ endif
 # option, it is silently ignored.
 GLOBAL_ASFLAGS+=-noexecstack
 
+ARM_ARCH_FLAGS=-march=armv6 -marm -mfpu=vfp -mfloat-abi=hard
+
 ifeq (ppc,$(OMR_HOST_ARCH))
     ifeq (gcc,$(OMR_TOOLCHAIN))
         ifeq (1,$(OMR_ENV_DATA64))
@@ -139,7 +141,7 @@ else
             endif
         else
             ifeq (arm,$(OMR_HOST_ARCH))
-                GLOBAL_ASFLAGS+=-march=armv6 -mfpu=vfp -mfloat-abi=hard
+                GLOBAL_ASFLAGS+=$(ARM_ARCH_FLAGS)
             else
                 # Nothing
             endif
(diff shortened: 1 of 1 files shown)


glue/configure_includes/configure_linux_arm.mk
```diff
@@ -27,36 +27,54 @@ else
   TEMP_TARGET_DATASIZE:=32
 endif
 
+ifeq ($(OPENJ9_DOCKER_CC_ARM),1)
+  TMP_HOST = $$(OPENJ9_CC_PREFIX)
```
charliegracie: Can you change the name of the environment variable to remove the OpenJ9 from it?  The OMR project is / can be used for many downstream projects so we should not be adding any new OpenJ9 specific options.  I am assuming this would be useful for other projects as well if it had a different name.  Thanks.

JamesKingdon: Hi @charliegracie ,
In this case OMR is the consumer of the variable which is defined in the Docker file added to openJ9 in https://github.com/eclipse/openj9/pull/622. The variable is named for consistency with other variables defined in that file. OPENJ9_DOCKER_CC_ARM specifically is destined to be removed as soon as resources permit, as it is something of a quick fix to allow rapid prototyping of builds for armhf.

jdekonin: If I understand the interaction between OpenJ9 and OMR correctly, openj9/runtime/gc_glue_java/configure_includes/configure_linux_arm.mk is the actualmk used by openj9 and the omr/glue is a set of examples to work from.  If this is the case, then these changes need to happen in the openj9 repository.  Can you confirm or deny that @charliegracie?

JamesKingdon: Ah, that's interesting. Changes to openj9/runtime/gc_glue_java/configure_includes/configure_linux_arm.mk were contributed in https://github.com/eclipse/openj9/pull/686.
If the files changed in this PR are not actually needed for building OpenJDK then we can scrap them and avoid the whole issue of what to call the env vars.

JamesKingdon: Something in this PR is needed as the build fails if I revert to the original branch.

glue/configure_includes/configure_linux_arm.mk
```diff
@@ -27,36 +27,54 @@ else
...
+else
+  TMP_HOST = arm-bcm2708hardfp-linux-gnueabi
+endif
+
 ifeq (linux_arm, $(SPEC))
```
JamesKingdon: This looks like the wrong test, it should be
ifneq (,$(findstring linux_arm,$(SPEC)))
to also accept linux_arm_docker.

glue/configure_includes/configure_linux_arm.mk
JamesKingdon: It's the change to omrmakefiles/rules.linux.mk that is needed in order to get the -marm option to the new compiler, so I can reduce this PR to just that change.

JamesKingdon:
@charliegracie I removed the unnecessary changes containing the OPENJ9 variables.

JamesKingdon:
Squashed the commits.

JamesKingdon:
@charliegracie any more changes needed?

Issue #1838, opened 2017-10-23
Remove LOCK and UNLOCK pseudo instructions from Z compiler


0xdaryl:
Remove deprecated `LOCK` and `UNLOCK` pseudo instructions from the Z code generator.  Both were part of the initial OMR compiler contribution but apply to a historical project and are no longer relevant.

A case-sensitive grep of the code for `::LOCK` or `::UNLOCK` is the starting point.

PR #2058, opened 2017-11-27
Remove LOCK and UNLOCK pseudo inst from Z compiler

shivammittal99:
Closes: #1838


0xdaryl:
genie-omr build zlinux

0xdaryl:
genie-omr build zlinux

PR #2064, opened 2017-11-27
Fail to initialize on bad limit or inline file

sehirst:
Currently if the JIT is unable to read the inline or limit file
supplied, it simply prints a warning and carries on. This is very
easy to miss with automated testing, and the whole purpose of
these files is for debugging and narrowing down a problem, so
there is no point to continue if the file isn't being picked up,
and worse it can lead one to false conclusions. Additionally,
the inlinefileOption accepts a range but does nothing with it.
This is obviously bad, but it makes no sense to have a range
for an inlinefile because there can easily be dependencies i.e.
one method needs to be inlined for another one to be. Therefore
the range code was removed. Doxygen documentation was also added.

Signed-off-by: Simon Hirst <shirst@ca.ibm.com>

### DIFF
--- a/compiler/ras/LimitFile.cpp
+++ b/compiler/ras/LimitFile.cpp
@@ -380,20 +380,39 @@ TR_Debug::scanInlineFilters(FILE * inlineFile, int32_t & lineNumber, TR::Compila
    return !inlineFileError;
    }
 
+/** \brief
+ *     This function opens the inlinefile and adds its entries to a TR::CompilationFilters.
+ *
+ *     An inlinefile is a file containing a list of methods, one per line, which the inliner will limit itself to
+ *     when trying to perform inlining. In other words, only methods from the file can be inlined, but there is no
+ *     guarantee that any of them will be inlined. The format for entry is:
+ *
+ *     + signature
+ *
+ *  \param option
+ *     Points to the first char after inlinefile=
+ *
+ *  \param base
+ *     Unused variable needed for downstream projects.
+ *
+ *  \param entry
+ *     The option table entry for this option.
+ *
+ *  \param cmdLineOptions
+ *     Unused variable needed for downstream projects.
+ *
+ *  \return
+ *     The unmodified parameter option if there is a problem with the file, aborting JIT initialization.
+ *     Otherwise a pointer to the next comma or NULL.
+ */
 char *
 TR_Debug::inlinefileOption(char *option, void *base, TR::OptionTable *entry, TR::Options * cmdLineOptions)
    {
    char *endOpt = option;
    char *name = option;
+   char *fail = option;
 
-   bool range = false;
-   if (*endOpt == '(')
-      {
-      ++endOpt;
-      ++name;
-      range = true;
-      }
-
+   // move to the end of this option
    for (; *endOpt && *endOpt != ','; endOpt++)
       {}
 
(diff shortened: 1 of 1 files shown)


sehirst:
@fjeremic Here is the change to fail on a bad limit file. @dsouzai Assisted by showing me how to make the JIT fail to initialize here.

compiler/ras/LimitFile.cpp
```diff
@@ -441,7 +441,8 @@ TR_Debug::inlinefileOption(char *option, void *base, TR::OptionTable *entry, TR:
       }
    else
       {
-      TR_VerboseLog::write("<JIT: warning: unable to read inline file --> '%s' will be ignored>\n", inlineFileName);
+      TR_VerboseLog::write("<JIT: fatal: unable to read inline file --> '%s'>\n", inlineFileName);
+      return option; // This is how to make the JIT fail to initialize. We want to fail if we can't read the file because it is too easy to miss that the file wasn't picked up
```
fjeremic: This comment seems like something that belongs in a Doxygen comment as it is non-obvious and as there are multiple exit points of this function. Would it be possible to Doxygen the these two functions and move this comment up to the documentation section?

sehirst: Absolutely. I'll get on that today.

0xdaryl:
Well here's a rock I haven't turned over before.  I had no idea you could do this.  Would you mind pointing me to the code that actually checks this string for "fatal" just so I understand how this mechanism works?  For some reason it's escaping me...

sehirst:
@0xdaryl This is pretty confusing, and as per Filip's request for more documentation I'm making it clearer. It's not the "fatal" which makes the JIT abort. It's returning the unmodified parameter `option`. That is a bit bananas but I'm not going to try and fix that in this PR.

sehirst:
Adding a WIP as this task has grown a bit in size and it will take me a bit more time to complete.

sehirst:
@0xdaryl @fjeremic Take a look at my latest commit. If you're happy with it I'll do the same type of documentation for `limitfileOption` and any potential code clean-up.

cathyzhyi:
Changes made to inlinefileOption removed the part processing the range which makes sense. LGTM once the build error is fixed.

compiler/ras/LimitFile.cpp
```diff
@@ -380,78 +380,108 @@ TR_Debug::scanInlineFilters(FILE * inlineFile, int32_t & lineNumber, TR::Compila
...
    }
 
+/** \brief
+ *     This function opens the inlinefile and adds its entries to a TR::CompilationFilters.
+ *
+ *     An inlinefile is a file containing a list of methods, one per line, which the inliner will limit itself to
+ *     when trying to perform inlining. In other words, only methods from the file can be inlined, but there is no
+ *     guarantee that any of them will be inlined. The format for entry is:
+ *
+ *     + package/Class.method(signature)
```
fjeremic: The `+ package/Class.method(signature)` seems very Java-esque and may not relate to the rest of the projects consuming OMR. Perhaps just leaving it as `+ signature` might be more clear? Especially since we have the following API which will return the exact string that should go after the plus sign:

https://github.com/eclipse/omr/blob/8b089c3af5fe41dc700aacf5ade826761687c2cf/compiler/compile/OMRCompilation.hpp#L421

fjeremic:
> @0xdaryl @fjeremic Take a look at my latest commit. If you're happy with it I'll do the same type of documentation for limitfileOption and any potential code clean-up.

Looks good other than the single comment I've noted above. Please amend a29d476 with a capitalized sentence case title as per the contribution guidelines:

compiler/ras/LimitFile.cpp
sehirst: Good point.

sehirst:
Not quite done yet. Had to revert some changes because of extensible class restrictions. Can't remove unused variables because of downstream dependencies.

sehirst:
Successfully built locally. Removed WIP. Ready to merge.

sehirst:
Whoops, forgot to change a request Filip had about the signature, and I'm doing a tiny bit more fine tuning on error handling. Put back the WIP but will remove it shortly. One more push and I'm done.

sehirst:
Ok, done now. Ready to merge.

0xdaryl:
genie-omr build all

charliegracie:
The zOS build failed because this PR does not contain the required fixes to compile successfully.  If it was rebased it should work.  If this does not contain zOS specific changes I would not bother rebasing.

Issue #1738, opened 2017-09-27
Remove specialized epilogues from compiler


0xdaryl:
Specialized epilogues are an experimental OMR compiler feature currently disabled in the Z and P code generators.  As it has been disabled for many years and never shown a strong performance benefit, the logic, option, and supporting code should all be removed.

A starting point for this cleanup is a case-insensitive search for `specializedEpilogue`.

PR #2066, opened 2017-11-28
Remove deprecated specialized epilogues

0xdaryl:
* Remove specialized epilogue code and fold surrounding code as necessary

Closes: #1738

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -1234,8 +1234,7 @@ void OMR::Compilation::performOptimizations()
    if (_optimizer)
       _optimizer->optimize();
 
-   if (!self()->getOption(TR_EnableSpecializedEpilogues) &&
-       self()->getOption(TR_DisableShrinkWrapping) &&
+   if (self()->getOption(TR_DisableShrinkWrapping) &&
        !TR::Compiler->target.cpu.isZ() && // 390 now uses UseDefs in CodeGenPrep
        !self()->getOptions()->getVerboseOption(TR_VerboseCompYieldStats))
       _optimizer = NULL;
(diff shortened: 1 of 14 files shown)


0xdaryl:
I'd appreciate a review from @fjeremic and @ymanton please.  This primarily impacts P and Z.

fjeremic:
Curious since it seems related, are we keeping the shrink wrapping code around as well or is that slated for future deprecation? IIRC there were still functional issues with it and it never showed any real performance benefit.

ymanton:
P changes LGTM.

0xdaryl:
@fjeremic , the shrink wrapping code was contributed primarily to demonstrate how one could do dataflow over `TR::Instruction`s as we didn't want to lose that code.  As an actual optimization, its worth was never really proven and it was never 100% functionally correct on all platforms.

Now that its part of the permanent git history of Eclipse OMR I'm fine with removing it.  It will simplify some aspects of the CodeGenerator as well.  The removal should be done as a single PR so that future historians can see all the changes that make up shrink wrapping conveniently should they need to.

vijaysun-omr:
I can only think of one (somewhat speculative) reason the shrink wrapping optimization in the OMR codebase. The main reason why the optimization did not have the impact as originally envisaged was that the inliner became quite aggressive in Java and (generally speaking) has a high success rate for small/medium methods on hot paths. One question I cannot answer with certainty is : will the inliner be as successful in other languages that OMR is being considered for ? If it won't be as successful in some other language due to constraints unique to that scenario, then it might be more important to optimize prologue/epilogue costs (that shrink wrapping type optimization would be a part of).

0xdaryl:
I don't have any evidence to say whether it would be more beneficial in an inlining-constrained environment or not.  The inliner technology in OMR is somewhat immature (the bulk of the smarts live in OpenJ9 and are highly tuned for Java) which means it may not be as effective or aggressive as Java.  This may expose more opportunities @vijaysun-omr postulates.  I think the only way we'll know for sure is to dust the code off and make it available for non-Java runtimes.  I think the code as-written is reasonably general, but as we've seen with most pieces of code derived with a Java origin there will be some Java-specific semantics that will need to be generalized.  Also, I do recall there were functional issues on at least Power that need investigation.

I think the shrink wrapping discussion should move to a separate issue (I'll create one).  The fate of shrink wrapping in OMR is independent of this PR.

vijaysun-omr:
Fair enough, I will merge this PR independently of the broader shrink wrapping discussion.

PR #2085, opened 2017-11-30
Use populationCount instead of bitCount32

### LINKED ISSUE #2081: Use populationCount rather than bitCount32
The compiler `OMR::CodeGenerator` class defines its own extern function `bitCount32` to do a population count.  This is historical and code should be using `populationCount` from the `infra/Bit.hpp` include instead.

Remove the `bitCount32` function and replace references with `populationCount`.

shivammittal99:
Closes #2081

Signed-off-by: Shivam Mittal <shivammittal99@gmail.com>


0xdaryl:
genie-omr build all

PR #2087, opened 2017-11-30
Call omrsig_signalOS_internal() on SIG_IGN/SIG_DFL

### LINKED ISSUE #2076: Spark daemon stops running with Java V8 SR5
When Java is launched with nohup in ZOS, nohup will register a SIGHUP signal handler to ignore the SIGHUP signal.

In Java 8 SR 5, the above SIGHUP signal handler is overwritten and therefore the Java process will be terminated with a SIGHUP signal.

hangshao0:
Recover our previous behaviour of omrsig_primary_sigaction():
When SIG_IGN/SIG_DFL has been registered for a signal, user should be able
to query such actions via omrsig_primary_sigaction().
omrsig_primary_sigaction() always calls omrsig_sigaction_internal() with
primary=true, in the case of SIG_IGN/SIG_DFL, the cached
sigData[signum].primaryAction might not be the same as sigaction registered
with the OS. 

Fixes #2076

Signed-off-by: Hang Shao <hangshao@ca.ibm.com>

### DIFF
--- a/omrsigcompat/omrsig.cpp
+++ b/omrsigcompat/omrsig.cpp
@@ -333,11 +333,28 @@ omrsig_signalOS_internal(int signum, const struct sigaction *act, struct sigacti
 	if (NULL == signalOS) {
 		rc = -1;
 	} else {
-		sighandler_t old = signalOS(signum, act->sa_handler);
+		sighandler_t handler = SIG_DFL;
+		/* signal() on WIN only takes SIG_IGN/SIG_DFL/function pointer as the second parameter (NULL is mapped to SIG_DFL).
+		 * If only querying the existing sighandler_t, use SIG_DFL to get the existing sighandler_t and then set it back.
+		 */
+		bool setAction = false;
+		if (NULL != act) {
+			handler = act->sa_handler;
+			setAction = true;
+		}
+		sighandler_t old = signalOS(signum, handler);
 		if (SIG_ERR == old) {
 			rc = -1;
-		} else if (NULL != oldact) {
-			oldact->sa_handler = old;
+		} else {
+			if (NULL != oldact) {
+				oldact->sa_handler = old;
+			}
+			if (!setAction) {
+				/* signal action was set to SIG_DFL, set old action back */
+				if (SIG_ERR == signalOS(signum, old)) {
+					rc = -1;
+				}
+			}
 		}
 	}
 #endif /* defined(POSIX_SIGNAL) */
@@ -362,9 +379,29 @@ omrsig_sigaction_internal(int signum, const struct sigaction *act, struct sigact
 		}
 
 		SIGLOCK(sigMutex);
-		/* Get previous handler from slot. */
 		if (NULL != oldact) {
-			*oldact = *savedAction;
+			struct sigaction oact = {{0}};
+			bool returnOldAction = false;
+			rc = omrsig_signalOS_internal(signum, NULL, &oact);
+			if (-1 != rc) {
+#if defined(POSIX_SIGNAL)
+				if (OMR_ARE_NO_BITS_SET(oact.sa_flags, SA_SIGINFO)) {
+#endif /* defined(POSIX_SIGNAL) */
+					if (((sighandler_t)SIG_DFL == oact.sa_handler)
+						|| ((sighandler_t)SIG_IGN == oact.sa_handler)
+					) {
+						returnOldAction = true;
+					}
+#if defined(POSIX_SIGNAL)
+				}
+#endif /* defined(POSIX_SIGNAL) */
+			}
+			if (returnOldAction) {
+				*oldact = oact;
+			} else {
+				/* Get previous handler from slot. */
+				*oldact = *savedAction;
+			}
 		}
 
 		if (NULL != act) {


omrsigcompat/omrsig.cpp
```diff
@@ -362,9 +364,28 @@ omrsig_sigaction_internal(int signum, const struct sigaction *act, struct sigact
...
 		SIGLOCK(sigMutex);
-		/* Get previous handler from slot. */
 		if (NULL != oldact) {
-			*oldact = *savedAction;
+			struct sigaction oact = {{0}};
+			bool retOact = false;
+			rc = omrsig_signalOS_internal(signum, NULL, &oact);
+
+#if defined(POSIX_SIGNAL)
+			if (!(oact.sa_flags & SA_SIGINFO)) {
```
DanHeidinga: Can you use the OMR_ALL_BITS_SET macros or explicitly check the `SA_SIGINFO == ....`?

hangshao0: I searched the whole OMR repo, there is no OMR_ALL_BITS_SET.

omrsigcompat/omrsig.cpp
```diff
@@ -333,11 +333,13 @@ omrsig_signalOS_internal(int signum, const struct sigaction *act, struct sigacti
 	if (NULL == signalOS) {
 		rc = -1;
 	} else {
-		sighandler_t old = signalOS(signum, act->sa_handler);
-		if (SIG_ERR == old) {
-			rc = -1;
-		} else if (NULL != oldact) {
-			oldact->sa_handler = old;
+		if (NULL != act) {
```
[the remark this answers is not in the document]
hangshao0: This piece of code is only active on Windows. 
We won't return old action on window if act is NULL, as the OS call at line 377 is signalOS(signum, act->sa_handler), act cannot be NULL.

omrsigcompat/omrsig.cpp
```diff
@@ -362,9 +364,28 @@ omrsig_sigaction_internal(int signum, const struct sigaction *act, struct sigact
 		}
 
 		SIGLOCK(sigMutex);
-		/* Get previous handler from slot. */
 		if (NULL != oldact) {
-			*oldact = *savedAction;
+			struct sigaction oact = {{0}};
+			bool retOact = false;
```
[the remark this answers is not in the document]
hangshao0: Renamed.

omrsigcompat/omrsig.cpp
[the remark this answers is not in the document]
DanHeidinga: Doesn't that leave the existing problem of missing the ignored signal on Windows?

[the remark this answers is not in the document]
hangshao0: Fixed.

omrsigcompat/omrsig.cpp
```diff
@@ -333,7 +333,11 @@ omrsig_signalOS_internal(int signum, const struct sigaction *act, struct sigacti
 	if (NULL == signalOS) {
 		rc = -1;
 	} else {
-		sighandler_t old = signalOS(signum, act->sa_handler);
+		sighandler_t handler = NULL;
+		if (NULL != act) {
+			handler = act->sa_handler;
+		}
+		sighandler_t old = signalOS(signum, handler);
```
DanHeidinga: Have you tested this on a Windows machine?  What was the result?

My reading of the [msdn doc](https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx) for `signal` indicates it can only be passed `SIG_IGN`, `SIG_DFL`, or a user function.

Unless `SIG_IGN` or `SIG_DFL` map to NULL, based on the doc I don't think this is valid and expect it to return `SIG_ERR`.

One ugly workaround would be to set it to SIG_IGN, get the old handler, and then set the old handler back.  I don't like that kind of hack though it opens a window where the "wrong" signal handler is installed.

hangshao0: Tested on Windows:

Passing NULL to handler, signal() does not return SIG_ERR, it returns a valid value. 
SIG_DFL is 0, SIG_IGN is 1.

hangshao0: Probably we shouldn't let user query the handler by passing NULL act. As SIG_DFL is mapped to NULL. Query using NULL act would accidentally set signal action to SIG_DFL

omrsigcompat/omrsig.cpp
```diff
@@ -333,11 +333,25 @@ omrsig_signalOS_internal(int signum, const struct sigaction *act, struct sigacti
 	if (NULL == signalOS) {
 		rc = -1;
 	} else {
-		sighandler_t old = signalOS(signum, act->sa_handler);
+		sighandler_t handler = SIG_DFL;
```
DanHeidinga: Can you add a comment to this section to describe why we're defaulting to the SIG_DFL so that the next person looking at this understands?

hangshao0: Comment added

charliegracie: @DanHeidinga are you ok with the comment that was added?

DanHeidinga: Yes - comment explains the "dance" going on here.

PR #2088, opened 2017-11-30
Add hasSlotSharingOrDeadSlotsInfo to OSRMethodData

cathyzhyi:
Add this API to query whether there is information about slots sharing or dead slots stored in OSRMethodData

### DIFF
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -779,6 +779,13 @@ TR_OSRMethodData::getArgInfo(int32_t byteCodeIndex)
    return args;
    }
 
+
+bool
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
@cathyzhyi could you please add in a message to the commit as well ?

andrewcraik:
The commit message and pull request should describe the intent of the change on its own without reference to other projects or changes. I think another iteration would make the message much more clear what this is doing.

andrewcraik:
The actual API makes sense to me.

cathyzhyi:
@andrewcraik Thanks for pointing this out. Changed the commit msg.

PR #2096, opened 2017-12-01
Deprecate get(.*)OpCodeFromNode in IBM Z codegen

fjeremic:
These functions are obsolete and not needed. Simplify the code by
deprecating the class of functions.


fjeremic:
This PR has a hard dependency on https://github.com/eclipse/openj9/pull/742. Commiters please note not to merge this PR until https://github.com/eclipse/openj9/pull/742 has been merged. Marking this item as WIP until that is done.

fjeremic:
https://github.com/eclipse/openj9/pull/742 has been merged. Unmarking WIP. This PR should be ready to be reviewed.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3369,7 +3369,7 @@ generateS390CompareAndBranchOpsHelper(TR::Node * node, TR::CodeGenerator * cg, T
                if(testRegister->getKind() == TR_GPR64 || cg->use64BitRegsOn32Bit())
                   compareOpCode = isUnsignedCmp ? TR::InstOpCode::CLG : TR::InstOpCode::CG;
                else
-                  compareOpCode = isUnsignedCmp ? TR::InstOpCode::getCmpLogicalOpCodeFromNode(nonConstNode) : TR::InstOpCode::getCmpOpCodeFromNode(nonConstNode);
+                  compareOpCode = isUnsignedCmp ? TR::InstOpCode::getCmpLogicalOpCode() : TR::InstOpCode::getCmpOpCode();
```
fjeremic: The reason why this change is safe is because we know at this point `constType` is a `TR::Address` or a `TR::Int64`. We also know `node` is an `TR::if*cmp*` tree which must have type safe children. This implies `nonConstType` must also be equivalent to the type of `constNode`.

As such on 64-bit we must always use a 64-bit compare (since the `constType` is a 64-bit value) and on 32-bit we must always use a 32-bit compare.

compiler/z/codegen/OpMemToMem.cpp
```diff
@@ -1809,7 +1809,7 @@ MemToMemTypedVarLenMacroOp::generateLoop()
 
    // Skip the loop if length is zero.
    TR::LabelSymbol * doneLoop = TR::LabelSymbol::create(_cg->trHeapMemory(),_cg);
-   startControlFlow = generateS390CompareAndBranchInstruction(_cg, TR::InstOpCode::getCmpOpCodeFromNode(_lenNode), _dstNode, _lenReg, (int32_t)0, TR::InstOpCode::COND_BNH, doneLoop, false, false);
+   startControlFlow = generateS390CompareAndBranchInstruction(_cg, TR::InstOpCode::getCmpOpCode(), _dstNode, _lenReg, (int32_t)0, TR::InstOpCode::COND_BNH, doneLoop, false, false);
```
fjeremic: The reason this and the subsequent change is correct is because the type of the `TR::arraycopy` node will be the same width as the system pointer type. This can clearly be seen here:

https://github.com/eclipse/omr/blob/e0b8377e50cbc004895cd3aa36e0c3f05dba5322/compiler/optimizer/ValuePropagationCommon.cpp#L871-L932

These `MemToMemTypedVarLenMacroOp` are only used in the context of evaluating `TR::arraycopy`, and hence `getCmpOpCodeFromNode` is equivalent to `getCmpOpCode`. Furthermore a few lines below the context in which `_lenReg` is used also reinforces the fact that on 64-bit target we need to use 64-bit compares and on 32-bit target we need to use 32-bit compares.

fjeremic:
@0xdaryl to ease the review process I've noted down some subtleties on why these changes are semantically correct via a self-review. Hope this helps.

0xdaryl:
Thanks.  I'm OK with all of this.

Why did you call it "IBM Z codegen" in your commit/PR titles?  I suggest dropping the "IBM" unless there's a reason for it.

fjeremic:
> Why did you call it "IBM Z codegen" in your commit/PR titles? I suggest dropping the "IBM" unless there's a reason for it.

They rebranded in August to IBM Z:
https://www.ibm.com/it-infrastructure/z/hardware

But you're right, the "IBM" doesn't serve any purpose.

0xdaryl:
genie-omr build zlinux zos

PR #2097, opened 2017-12-02
Fix for X86-32 minmaxEvaluator Issue

### LINKED ISSUE #2099: Incorrect Result from minmaxEvaluator on X86-32
https://github.com/eclipse/omr/pull/1839 introduced a bug into minmaxEvaluator on X86-32. The issue was because signed comparison was used between two lower words, where unsigned comparison should be used.
One example that the failure occurs is: max(0xF234567872345678L, 0xF2345678f2345678L) incorrectly returning 0xF234567872345678L.

0dvictor:
minmaxEvaluator on X86-32 should use unsigned
comparison for low word instead of signed one.
Fixing it.

Closes #2099 

### DIFF
--- a/compiler/x/i386/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/i386/codegen/OMRTreeEvaluator.cpp
@@ -3736,14 +3736,17 @@ OMR::X86::I386::TreeEvaluator::integerPairByteswapEvaluator(TR::Node *node, TR::
 TR::Register*
 OMR::X86::I386::TreeEvaluator::integerPairMinMaxEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   TR_X86OpCodes SETcc = BADIA32Op;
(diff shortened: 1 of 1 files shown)


andrewcraik:
LGTM - @0xdaryl or @vijaysun-omr would one of you mind taking a look at this fix?

0xdaryl:
A genie build is not required because the commit only affects x86 files.

0dvictor:
Sorry, I meant there was a "problem/bug" in minmaxEvaluator on X86-32. As far as I know, there isn't any OMR or OpenJ9 Issue.

andrewcraik:
@0dvictor I think the request was to create an issue documenting the problem and the failure mode/symptoms and to reference it with the PR so that when the PR is merged the issue is resolved.

0dvictor:
Sure, I've created https://github.com/eclipse/omr/issues/2099

PR #2098, opened 2017-12-04
Fix OSR on guard failure during late inlining

ncough:
OSR on guard failure will use an OSR transition
in place of a virtual call. This can occur
during late inlining. For bookkeeping to be
correctly updated, the OSR blocks for the
caller must be linked with the callee's.
However, the incorrect resolved method
symbol was being used during late inlining,
resulting in a crash.


andrewcraik:
LGTM - @vijaysun-omr over to you, but looks like a fairly straight forward fix.

vijaysun-omr:
Simple OSR related bug fix. Merging.

PR #2109, opened 2017-12-05
CMake: Organize targets into folders

dnakamura:
Organize the various targets into folders when using a generator with
folder support.

See
https://cmake.org/cmake/help/v3.2/prop_gbl/USE_FOLDERS.html
https://cmake.org/cmake/help/v3.2/prop_tgt/FOLDER.html


charliegracie:
genie-omr build plinux

PR #2114, opened 2017-12-06
Add minimum ALS for each instruction in Z codegen

fjeremic:
Often times when implementing a new feature we make use of instructions
which have been introduced in a specific processor. To make use of such
instructions we have to guard the block of code which will emit the
respective instruction with an architecture guard. If such a guard is
omitted we end may end up executing at runtime an unsupported
instruction, depending on the environment in which the compiler is
executing.

To prevent this from happening in the future we add a minimum
architecture level set (ALS) for each instruction and add a soft assert
in the constructor of each instruction to validate that we will never
crash with a SIGILL due to a missed architecture guard.

### DIFF
--- a/compiler/z/codegen/ArchInstOpCodeEncoding.hpp
+++ b/compiler/z/codegen/ArchInstOpCodeEncoding.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2017 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
--- a/compiler/z/codegen/ArchInstOpCodeEncoding.hpp
+++ b/compiler/z/codegen/ArchInstOpCodeEncoding.hpp
@@ -25,1144 +25,1143 @@
  */
 
 
-   {0x00, 0x00, PSEUDO},   // BAD
-   {0x5A, 0x00, RX_FORMAT},   // A
-   {0xED, 0x1A, RXE_FORMAT},  // ADB
-   {0xB3, 0x1A, RRE_FORMAT},  // ADBR
-   {0xB3, 0xD2, RRF_FORMAT},  // ADTR
-   {0xED, 0x0A, RXE_FORMAT},  // AEB
-   {0xB3, 0x0A, RRE_FORMAT},  // AEBR
-   {0xC2, 0x09, RIL_FORMAT},  // AFI
-   {0xE3, 0x08, RXY_FORMAT},  // AG
-   {0xE3, 0x18, RXY_FORMAT},  // AGF
-   {0xC2, 0x08, RIL_FORMAT},  // AGFI
-   {0xB9, 0x18, RRE_FORMAT},  // AGFR
-   {0xA7, 0x0B, RI_FORMAT},   // AGHI
-   {0xEC, 0xD9, RIE_FORMAT},  // AGHIK
-   {0xB9, 0x08, RRE_FORMAT},  // AGR
-   {0xB9, 0xE8, RRR_FORMAT},  // AGRK
-   {0x4A, 0x00, RX_FORMAT},   // AH
-   {0xB9, 0xC8, RRR_FORMAT},  // AHHHR
-   {0xB9, 0xD8, RRR_FORMAT},  // AHHLR
-   {0xA7, 0x0A, RI_FORMAT},   // AHI
-   {0xEC, 0xD8, RIE_FORMAT},  // AHIK
-   {0xE3, 0x7A, RXY_FORMAT},  // AHY
-   {0xCC, 0x08, RIL_FORMAT},  // AIH
-   {0x5E, 0x00, RX_FORMAT},   // AL
-   {0xE3, 0x98, RXY_FORMAT},  // ALC
-   {0xE3, 0x88, RXY_FORMAT},  // ALCG
-   {0xB9, 0x88, RRE_FORMAT},  // ALCGR
-   {0xB9, 0x98, RRE_FORMAT},  // ALCR
-   {0xC2, 0x0B, RIL_FORMAT},  // ALFI
-   {0xE3, 0x0A, RXY_FORMAT},  // ALG
-   {0xE3, 0x1A, RXY_FORMAT},  // ALGF
-   {0xC2, 0x0A, RIL_FORMAT},  // ALGFI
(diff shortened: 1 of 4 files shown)


fjeremic:
Note to reviewers that `TR_S390ProcessorInfo` was simply moved so we have exposure to the right type. Also note that this is the first PR of several PRs which will clean up the instruction classes on the Z codegen. The entire contents of ArchInstOpCodeEncoding.hpp will be reformatted and very possibly folded into other places. This is just a temporary state to make this particular PR consumable for review.

0xdaryl:
I'm wondering if modifying the OMR base constructor to explicitly use `setOpCodeValue` to set the opcode (rather than in the member initialization list) and then having Z provide its own specialization of `setOpCodeValue` to do the architecture check as you have in this PR might be a better way to go?  It would keep the architecture check in one place (I think) and you should have access to the CodeGenerator object.

In any case, you do need to add some smarts into `setOpCodeValue` so that your well-intentioned checks aren't subverted.

0xdaryl:
genie-omr build zlinux zos

fjeremic:
> In any case, you do need to add some smarts into setOpCodeValue so that your well-intentioned checks aren't subverted.

@0xdaryl I'm actually just seeing this API now. To me this API does not make sense as I believe `InstOpCode` should be immutable. I cannot think of any reason why it shouldn't be as such. This is outside the scope of this PR but I do want to hear your thoughts on this. Please let me know if we should open up an issue to discuss.

compiler/z/codegen/OMRInstruction.hpp
```diff
@@ -88,6 +88,13 @@ class OMR_EXTENSIBLE Instruction : public OMR::Instruction
    int32_t getBlockIndex() { return _blockIndex; }
    void setBlockIndex(int32_t i) { _blockIndex = i; }
 
+   void setOpCodeValue(TR::InstOpCode::Mnemonic op)
+      {
+      OMR::Instruction::setOpCodeValue(op);
+
+      TR_ASSERT(_cg->getS390ProcessorInfo()->supportsArch(_opcode.getMinimumALS()), "Processor detected does not support instruction %s\n", TR::InstOpCode::getMnemonicName(op));
+      }
```
fjeremic: @0xdaryl just making sure this will work the way I think it will. This function is not virtual but it should still "override" the base OMR implementation when being used in the Z codegen, correct?

fjeremic:
So there seem to be a ton of compilation failures on x86 and I suspect ARM will fail as well. The reason is because x86 does not build on top of the `Mnemonic` enum for their opcodes:

https://github.com/eclipse/omr/blob/45adab45525a3cf3f397434050ed6c441dc77929/compiler/codegen/OMRInstOpCode.hpp#L46-L50

Instead x86 seems to use their own enum:

https://github.com/eclipse/omr/blob/45adab45525a3cf3f397434050ed6c441dc77929/compiler/x/codegen/X86Ops.hpp#L31-L36

With their own setter:

https://github.com/eclipse/omr/blob/45adab45525a3cf3f397434050ed6c441dc77929/compiler/x/codegen/OMRInstruction.hpp#L98

And as such we get ambiguity when we call `self()->setOpCodeValue(op)` in the `OMR::Instruction` base constructor. The solution here is for x86 to use the proper `Mnemonic` enum to define the opcodes, otherwise `setOpCodeValue` cannot be properly extended via the extension mechanism without breaking x86 and ARM. This work is outside the scope of this item though. I'll be opening up an issue for this.

@0xdaryl given the above observations I propose we do not extend the `setOpCodeValue` on Z until the appropriate changes are made. We will only lose out on the assertion check in the currnetly 5 places which use `setOpCodeValue` on Z. Do you agree with this approach?

0xdaryl:
> To me this API does not make sense as I believe InstOpCode should be immutable. I cannot think of any reason why it shouldn't be as such. This is outside the scope of this PR but I do want to hear your thoughts on this. Please let me know if we should open up an issue to discuss.

In general, I think your philosophy is correct about the immutability of an instruction.  However, I believe the current uses of `setOpCodeValue` in the code base are for "slight" refinements to the instruction being used (all other properties being the same).

One of the reasons this is done was because there isn't a mechanism in the Code Generators for recycling instruction memory within a compilation.  A real example that comes to mind where this was particularly important was when we were implementing x87 floating point support.  Because the x87 stack (i.e., the floating point registers) is "managed" by the register assigner which runs after instruction selection, it was not always possible to know at instruction selection time whether a pop form instruction was needed or not to clear the stack (that is, to free a register).  The register assigner became responsible for inserting the "operation-with-pop instruction" in place of the "operation-without-pop instruction", and the cheapest way to do that was to simply replace the opcode on the instruction (all other properties being equal).  This happens a lot in the course of a method (usually whenever the live range of a register ends, though I recall it wasn't as simple as that as we can detect that during instruction selection).  In the early days of Testarossa, we were very miserly on the memory used due to the embedded environment constraints, and we still hold that philosophy today.  Hence, you can make an argument for re-using instruction memory in this case.

There are other examples in the X and P code generators where substituting one opcode for another in controlled circumstances is necessary.  But it comes down to the convenience of reusing a `TR::Instruction`'s memory.  If there are places in the Code Generators where changing the opcode changes either the form or the properties on an instruction, that's wrong to me and a new instruction is warranted.

Note that the `TR::Node` class went through a similar existential crisis a few years back.  At one point, it was very common to simply change the opcode on a Node in a transformation via its own `setOpCodeValue()` function.  However, in many cases the properties of the Node weren't changed (or weren't changed in their entirety).  This led to subtle bugs and hence the `recreate()` function was introduced to re-use Node memory properly when its identity changes, and all `setOpCodeValue()` calls were replaced with `recreate()`'s.  The other thing about the Node class is that there are mechanisms for recycling Node memory because some compilation units have the compile memory dominated by Node instances.

I don't think we need to go so far on `TR::Instruction` yet.  However, after more across-the-board refactoring of the Instruction hierarchy occurs (and what you're doing starting with this PR is certainly a big part of it), then I'm willing to take another look at it.

fjeremic:
In light of the above response I'm removing the `setOpCodeValue` commit until #2126 is addressed. The value proposition of this change is to avoid generating unsupported instructions. There are currently only 5 locations in the Z codegen where `setOpCodeValue` is used. These locations will not benefit from the architecture check introduced in this PR (to be fixed later once #2126 is addressed), however we are no worse off than we are today (we have no checks today at all). As such having asserts for the additional 5 locations should IMO not block the merging of this PR. @0xdaryl do you agree with my proposition?

fjeremic:
> My only minor comment with this PR is to update the copyright date on compiler/z/codegen/ArchInstOpCodeEncoding.hpp and it should be good to merge.

Fixed in c0506f5

PR #2121, opened 2017-12-07
Value profiling API changes

ncough:
A new value profiling implementation, JProfiling, is being
added to OpenJ9, along with some changes to the value
profiling API. These change updates these calls as necessary
and adds configuration options for the profiling modes.


charliegracie:
genie-omr build all

andrewcraik:
This looks like general clean-up and adding a few options and hook points. LGTM from me pending any further changes.

compiler/optimizer/Inliner.hpp
```diff
@@ -72,7 +72,6 @@ const float SECOND_BEST_MIN_CALL_FREQUENCY = .2275f; //65% of 35%
 
 class TR_BitVector;
 class TR_CallStack;
-class TR_ExtraAddressInfo;
```
vijaysun-omr: We seem to be removing TR_ExtraAddressInfo and using TR_ExtraValueInfo instead of TR_ExtraAbstractInfo. I assume these two changes are related in that TR_ExtraAddressInfo is the only other subclass of TR_ExtraAbstractInfo. Could you please elaborate on why this change is being made ?

ncough: `TR_ExtraAddressInfo` has been removed here as it wasn't being used and its a small pain to forward declare, as its now a template. The other changes from `TR_ExtraAbstractInfo` to `TR_ExtraValueInfo` are due to the removal of `TR_ExtraAbstractInfo`. Rather than having to cast back and forth between the two classes, `TR_ValueInfo`'s methods now use `TR_ExtraValueInfo` directly.

compiler/control/OMROptions.cpp
```diff
@@ -4797,6 +4798,7 @@ char *OMR::Options::_verboseOptionNames[TR_NumVerboseOptions] =
    "hookDetailsClassLoading",
    "hookDetailsClassUnloading",
    "sampleDensity",
+   "profiling"
```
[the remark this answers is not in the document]
ncough: For verbose options, the option name has to match its position in `TR_VerboseFlags`, rather than being alphabetically sorted like the main option list. Here is matches the value for `TR_VerboseProfiling`.

PR #2122, opened 2017-12-07
Enable optimized ArrayCopy as default on X86

0dvictor:
Although X86 has a new optimized ArrayCopy implementation,
the deprecated ones are still default. Enabling the new ones.


0dvictor:
For the purpose of debugging and record keeping, the deprecated code is not deleted with this pull request. The deprecated code will be removed via a separate pull request a few days after.

charliegracie:
genie-omr build zlinux

PR #2128, opened 2017-12-09
Remove obsolete RegisterSymbols

0xdaryl:
`TR::RegisterSymbol`s are a remnant from an old register allocator and
do not have relevance in the OMR project.  These symbols are unused by
any known downstream project.

Remove references to `TR::RegisterSymbol`s and fold code around symbol
identity checks assuming register symbols never exist.

### DIFF
--- a/compiler/codegen/Analyser.cpp
+++ b/compiler/codegen/Analyser.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2017 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -60,7 +60,6 @@ void TR_Analyser::setInputs(TR::Node     *firstChild,
    if (  firstChild->getOpCode().isMemoryReference()
       && firstChild->getSymbolReference() != vftPointerSymRef
       && firstChild->getReferenceCount() == 1
-      && !firstChild->getSymbol()->isRegisterSymbol()
       && !lockedIntoRegister1)
       {
       setMem1();
@@ -69,7 +68,6 @@ void TR_Analyser::setInputs(TR::Node     *firstChild,
    if (  secondChild->getOpCode().isMemoryReference()
       && secondChild->getSymbolReference() != vftPointerSymRef
       && secondChild->getReferenceCount() == 1
-      && !secondChild->getSymbol()->isRegisterSymbol()
       && !lockedIntoRegister2)
       {
       setMem2();
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -120,7 +120,6 @@ OMR::SymbolReferenceTable::SymbolReferenceTable(size_t sizeHint, TR::Compilation
      _hasUserField(false),
      _aggregateShadowSymbolMap(8, comp->allocator("SymRefTab")),
      _aggregateShadowSymbolReferenceMap(8, comp->allocator("SymRefTab")),
-     _registerSymbolRefs(NULL),
      _sharedAliasMap(NULL)
    {
    _numHelperSymbols = TR_numRuntimeHelpers + 1;;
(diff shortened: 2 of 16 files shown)


mstoodle:
I would like to hear @vijaysun-omr and @andrewcraik 's opinion on this one. I know this support was only every used on Z, which makes it less useful than it could be, but I'm wondering if any of the common machinery to, say, express registers in IL (that's being removed here) could be usefully repurposed in OMR.

For example, it might be a nice feature for JitBuilder to be able to express register hints all the way from ilgen.

Thoughts?

0xdaryl:
There's the question of whether they're needed and the question of whether they should be implemented _this way_.  This implementation appears to change the symbols on loads and stores, which potentially requires special casing throughout (you see some of that in this PR).  The question needs to be asked why not extend the RegLoad/RegStore infrastructure to better support the notion of register operations in IL, as it seems more "natural".

vijaysun-omr:
Yes in my opinion, having the clear notion of a load/store being "from memory" is worthwhile to keep the specification of those opcodes simple (one of the IL design principles we should keep in mind). Otherwise it leads to the complications that every place that deals with loads/stores needs to be special cased for these cases. If we need to represent source code locals that "must" be mapped to registers (as was the original requirement in this case), we should consider a separate store/load IL opcode family for that if regLoad/regStore are considered too heavy (lots of GlRegDeps everywhere and not all optimizations can cope with regLoads/regStores/GlRegDeps currently).

andrewcraik:
I am not a fan of a load/store not being "from memory" either. I think my opinion on how to guide the register assignment may be a bit different. At present the optimizer doesn't really consider registers until after Global Register Assigner (GRA) and so not all opts expect things like regLoad and regStore as stated by @vijaysun-omr . I wonder if a path of less resistance to get this support would be to add a notion of mapping a temp/auto to a register and requiring GRA to globally allocate that temp/auto to a specific register and for the various local register assigners to also obey this mapping in their assignment choice. That means that even without GRA the temp will be correctly assigned and we would minimize the impact on the rest of the compiler. The main downside is that we would need to add some optimizer support to avoid eliminating temps that are bound to registers. I'm not quite clear how with a new opcode we would capture the target register unless we encode it in the load/store opcode's node somehow...

mstoodle:
Given the entire optimizer can deal with loads/stores, and not all of the optimizer can deal with regLoads/regStores, I guess I was thinking the notion of a register symbol tied to an automatic symbol was a reasonable way to express a hint, and that GRA would either take the hint or change it to a real load/store (and codegen would just treat it like a load/store to an auto).

After all, a load/store that gets changed by GRA to a regLoad/regStore isn't very different in concept and the optimizer seems to deal with that just fine. That register symbol simply provides a mechanism to capture the preferred register in which to store that temp (and also expresses the importance of that allocation to GRA).

@vijaysun-omr am I making any sense? is there a better way to attach that hint? I am not sure I like the "side table" notion mapping auto's to preferred registers. Given what I wrote above, wouldn't the symbol be a more natural place to express it? But maybe it should have its own subclass of AutomaticSymbol rather than being a specially configured version of an AutomaticSymbol...

vijaysun-omr:
I think we should be clear on a couple of points before going too far in this discussion 

1. Is the register symbol representing something that is just a hint or a mandatory requirement o register assign ? Use cases have come up in both categories in the past, but the real question is : what do we think we want to support in the future at the OMR layer (and whether this is how we would choose to implement that) ? Using GRA to implement a hint sounds more acceptable than for implementing a mandatory requirement (what about noOpt ?).

2. I think there is also a variant wherein the real register is chosen ahead of time for a given local vs the real register is not chosen but whatever the real register is chosen for this local by the compiler, it remains consistent for that local.  

Clarifying these points might help focussing the discussion so that we are not talking about different things.

mstoodle:
Yes, ok. I've pulled this pull request too far afield. No matter where the rest of the discussion goes, it will never circle back to needing the exact code that's being removed in this pull request (which is decidedly not the proper way to do any of it).

I think we're all in agreement that this code can be removed. We should move the rest of it to an issue.

Sorry for the diversion, Daryl. Please rebase your pull request and I'll merge it.

0xdaryl:
Rebased.

genie-omr build all

mstoodle:
Thanks @0xdaryl

PR #2130, opened 2017-12-11
Prepare to relocate ValuePropagation functions to OpenJ9

0xdaryl:
Functions that are wholly guarded with J9_PROJECT_SPECIFIC macros will
be relocated to the OpenJ9 project.  The code that calls them, however,
is guarded with J9_PROJECT_SPECIFIC macros but is nested within functions
that are not.  To enable relocation, these calls need to have their
receiver object (i.e., `this`) downcast to `TR::ValuePropagation`.

Functions affected:

```
transformRTMultiLeafArrayCopy
transformObjectCloneCall
transformArrayCloneCall
transformConverterCall
```

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.cpp
+++ b/compiler/optimizer/OMRValuePropagation.cpp
@@ -7260,8 +7260,7 @@ void OMR::ValuePropagation::doDelayedTransformations()
    for (converterCallTree = convIt.getFirst();
 		   converterCallTree; converterCallTree = convIt.getNext())
       {
-
-      transformConverterCall(converterCallTree);
+      static_cast<TR::ValuePropagation *>(this)->transformConverterCall(converterCallTree);
       }
    _converterCalls.deleteAll();
 
@@ -7272,7 +7271,7 @@ void OMR::ValuePropagation::doDelayedTransformations()
       ObjCloneInfo *cloneInfo = objCloneTypeIt.getFirst();
       while (callTree && cloneInfo)
          {
-         transformObjectCloneCall(callTree, cloneInfo);
+         static_cast<TR::ValuePropagation *>(this)->transformObjectCloneCall(callTree, cloneInfo);
          callTree = objCloneIt.getNext();
          cloneInfo = objCloneTypeIt.getNext();
          }
@@ -7287,7 +7286,7 @@ void OMR::ValuePropagation::doDelayedTransformations()
       TR_OpaqueClassBlock *clazz = arrayCloneTypeIt.getFirst();
       while (callTree && clazz)
          {
-         transformArrayCloneCall(callTree, clazz);
+         static_cast<TR::ValuePropagation *>(this)->transformArrayCloneCall(callTree, clazz);
          callTree = arrayCloneIt.getNext();
          clazz = arrayCloneTypeIt.getNext();
          }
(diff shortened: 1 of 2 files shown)


andrewcraik:
I thought that VP was using virtual inheritance so the need for the casting seems a bit mysterious to me - what am I missing in relation to the casts? I'd prefer to minimize casting in this complicated code if possible (and I thought that the virtual approach would allow this).

0xdaryl:
The code that calls these functions is guarded by `J9_PROJECT_SPECIFIC` macros and lives in the `OMR::ValuePropagation` class.  Once the functions themselves that are wholly guarded with the macros are moved to the OpenJ9 project, they will live in the `J9::ValuePropagation` class which is actually a subclass of `OMR::ValuePropagation`.

In order for a function in `OMR::ValuePropagation` to call a function that only exists in `J9::ValuePropagation` you either have to do the downcast or declare those functions virtual in `OMR::ValuePropagation`.

The first approach is somewhat ugly in that you're calling a function in a subclass, which normally wouldn't be generally safe to do but in this case you're guaranteed that it exists because of the project-specific ifdefs.

The second approach introduces project-specific virtual functions in OMR, functions that will only ever have an implementation in OpenJ9.  This could be mitigated somewhat by guarding those function definitions with project-specific macros, but I'm not keen on adding new project-specific macros to OMR which is why it wasn't my first choice.

We could add a `self()` method to `OMR::ValuePropagation` that hides some of the verbosity of the static casts.

andrewcraik:
hmm, I suspect the right answer for these methods in the long term is to move the recognition logic itself down to OpenJ9 so we don't need the casts. These are all specific call method matches and so giving projects a virtual method hook where a project can put code which recognizes and transforms/queues for transformation specific calls would make conceptual sense. That way these methods would move down along with their recognition logic and the hook would be only visible entry point to this in OMR. If you want to move these down temporarily as part of that larger clean up I can go along with that, I just don't think the casting is a safe state to leave things in over the long term and doesn't solve the problem of providing a nice hook point for other framework adopters.

mstoodle:
@0xdaryl I notice you're also removing findResolvedMethodClass in this PR? Is that a freebie you're throwing in, or something that shouldn't be in this PR? Maybe it should have its own commit with meaningful message if it is meant to be included, since it's not related to the other changes you're describing in the PR and making in the existing commit.

mstoodle:
Just a ping @0xdaryl to see where we're at here :) .

0xdaryl:
@andrewcraik : I agree with your concerns and I opened #2317 to track the refactoring effort.

@mstoodle : PR now split into two commits.

vijaysun-omr:
genie-omr build all

vijaysun-omr:
@andrewcraik just checking that you have reviewed @0xdaryl update and are okay with this PR as it stands ?

andrewcraik:
I am still not a fan of the downcasting in the change - the movement to OpenJ9 is one portion of each feature - object clone transformation, for example, has all the infrastructure for finding and tracking the calls of interest in OMR and the only part that is being moved down is the transformation. It would make more sense to put the hook points in and then move one entire transformation at a time down.

My main worry is the possibility of bugs creeping in over time because of the casting and because features will be split over two projects (leading to more coordinated changes between the projects being needed) since the full refactoring is going to be some time coming. If this is holding other things up then I suppose we can live with it.

Removal of the dead method is fine with me.

vijaysun-omr:
@0xdaryl what is the next step on this ?

0xdaryl:
Closing this for now.  I'll re-open when I have time to advance it in the future.

PR #1851, opened 2017-10-26
z/TPF Update for I-Stream Determination

jjohnst-us:
Change structure member name by which z/TPF picks-up
I-Stream count for System because of new features
added to z/TPF Operating System. 

Signed-off-by: jjohnst-us <jjohnst@us.ibm.com>

PR #2131, opened 2017-12-11
z/TPF minor updates for omrosdump_helpers.c

jjohnst-us:
Update to change valid programs to write out into dumps.  Removed
explicitly translating pathconf function call parameter as that function
was added into z/TPF a2e library support.

### DIFF
--- a/port/ztpf/omrosdump_helpers.c
+++ b/port/ztpf/omrosdump_helpers.c
@@ -55,10 +55,10 @@ extern void masterSynchSignalHandler(int nbr, siginfo_t *siginfo,
 #define    KEY0(a)    do { cinfc(CINFC_WRITE, CINFC_CMMJDB); } while(0)
 #define    UNKEY(a) do { keyrc(); } while(0)
 
-char javaPgmsToDump[][5] = { "DJBV", "DJHK", "DJNC", "DJVB", "DJCK", "DJHS",
+char javaPgmsToDump[][5] = { "DJHK", "DJNC", "DJVB", "DJCK", "DJHS",
 		"DJPR", "DJVM", "DJDD", "DJHT", "DJRD", "DJZL", "DJDP", "DJHV", "DJSG",
 		"D9VM", "DJDY", "DJHY", "DJSH", "DJ7B", "DJGC", "DJIT", "DJTH", "DJAR",
-		"DJGK", "DJMT", "DJTR" };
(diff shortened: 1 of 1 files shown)


port/ztpf/omrosdump_helpers.c
```diff
@@ -851,7 +848,8 @@ buildELFNoteArea(Elf64_Nhdr *buffer, DIB *dibPtr)
 	 * fractional. Does it really matter? Nah. But if you're gonna do it, you might
 	 * as well do it right ....
 	 */
-	prstat->stime.tv_usec = ecbptr()->ce1istim / 4098; /* Used CPU time    */
+	/* prstat->stime.tv_usec = ecbptr()->ce1istim / 4098; Used CPU time    */
```
charliegracie: Is the commented out line valuable?  If so can you document why we are keeping it around?  Thanks

port/ztpf/omrosdump_helpers.c
```diff
@@ -424,10 +424,7 @@ ztpfBuildCoreFile(void *argv_block)
 	sprintf((char *)workSpace(arg), "core.%lX.%d", dibPtr->dstckf,
 			dumpSiginfo(arg)->si_pid);
 	{
-		int pathLen = strlen(pathName);
-		char ebcdicPath[pathLen + 1];
-		a2e_len(pathName, ebcdicPath, strlen(pathName));
-		int fsystype = pathconf(ebcdicPath, _TPF_PC_FS_TYPE);
+		int fsystype = pathconf(pathName, _TPF_PC_FS_TYPE);
```
charliegracie: Should all a2e conversions be removed from ztpf? Just asking purely out of curiosity.

jjohnst-us: z/TPF has the same issue translation issue as z/OS (both being ebcdic platforms).  To assist with porting middleware I think we do the same as z/OS where we redefine standard functions to an atoe_* version (eg. atoe_printf) and we link in a library which implements those functions which simply translates string/char parameters or returns as required.  Our a2e environment is still progressing, there still are a handful of functions we need to implement.

In the future we are looking to do this in a better way that works better with C++, but for now we maintaining the same strategy.

jjohnst-us:
@charliegracie , updates made per you request.  Also could you take a look at pull request
https://github.com/eclipse/omr/pull/1851

Much appreciated.

charliegracie:
genie-omr build all

PR #2132, opened 2017-12-11
Add perf record utility for compilation thread

ncough:
This adds a utility to generate a performance profile
for a specific compilation. This is useful when attempting
to cut compile time overhead, particularly when there are
a handful of expensive compiles (>10s in duration) but
they are drowned out others.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -49,6 +49,7 @@
 #include "cs2/allocator.h"                     // for heap_allocator
 #include "cs2/sparsrbit.h"
 #include "env/CompilerEnv.hpp"
+#include "env/CompileTimeProfiler.hpp"         // for TR::CompileTimeProfiler
 #include "env/IO.hpp"                  // for IO (trfflush)
 #include "env/ObjectModel.hpp"                 // for ObjectModel
 #include "env/KnownObjectTable.hpp"            // for KnownObjectTable
@@ -991,6 +992,9 @@ int32_t OMR::Compilation::compile()
    if (printCodegenTime) compTime.startTiming(self());
    if (_recompilationInfo)
       _recompilationInfo->startOfCompilation();
+
+   // Create the compile time profiler
+   TR::CompileTimeProfiler perf(self(), "compileTimePerf");
 
    {
      if (printCodegenTime) genILTime.startTiming(self());
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1026,6 +1026,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"printErrorInfoOnCompFailure",        "O\tPrint compilation error info to stderr", SET_OPTION_BIT(TR_PrintErrorInfoOnCompFailure), "F", NOT_IN_SUBSET},
    {"privatizeOverlaps",  "O\tif BCD storageRefs are going to overlap then do the move through a temp", SET_OPTION_BIT(TR_PrivatizeOverlaps), "F"},
    {"profile",            "O\tcompile a profiling method body", SET_OPTION_BIT(TR_Profile), "F"},
+   {"profileCompileTime",   "I\tgenerate a perf report for a specific compilation", SET_OPTION_BIT(TR_CompileTimeProfiler), "F" },
    {"profileMemoryRegions", "I\tenable the collection of scratch memory profiling data", SET_OPTION_BIT(TR_ProfileMemoryRegions), "F" },
    {"profilingCompNodecountThreshold=", "M<nnn>\tthreshold for doubling the method to do a profiling compile is considered expensive",
         TR::Options::setStaticNumeric, (intptrj_t)&OMR::Options::_profilingCompNodecountThreshold, 0, "F%d", NOT_IN_SUBSET},
--- a/compiler/env/CompileTimeProfiler.hpp
+++ b/compiler/env/CompileTimeProfiler.hpp
@@ -1,0 +1,216 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2017 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *******************************************************************************/
+
+#ifndef OMR_COMPILE_TIME_PROFILER_HPP
+#define OMR_COMPILE_TIME_PROFILER_HPP
+
+#include "compile/Compilation.hpp"
+
+#include <stdlib.h>
+#include <stdio.h>
+
+#if defined(LINUX)
+
+#include <sys/wait.h>
+#include <sys/types.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+
+namespace TR {
+
+/**
+ * Utility to generate a profile for the current compilation,
+ * to assist in determining the source of compile time overhead. This
+ * is useful in situations where there is a particularly long running
+ * compile, such as a profiled very-hot or scorching compile, but profiling
+ * all compilations results in too much noise.
+ *
+ * This implementation is Linux specific, due to its reliance on perf to
+ * generate the profile.
+ *
+ * It will fork and execute perf, configured such that it will
+ * profile the current compilation thread until the current compilation ends.
+ *
+ * Perf, by default, is configured to collect call graph information at a
+ * high frequency, with the following options:
+ *   --call-graph -F9999
+ * These defaults can be overridden using the TR_CompileTimeProfiler env var.
+ *
+ * The current implementation only supports perf, but other profiling
+ * utilities could be introduced. 
+ *
+ * In testing, the initial portion of the compile was lost, due to perf's
+ * start up. A delay has been added to manage this.
+ */
+class CompileTimeProfiler
+   {
+public:
+   CompileTimeProfiler(TR::Compilation *comp, const char *identifier)
+      {
+      _pid = 0;
+      if (comp->getOption(TR_CompileTimeProfiler))
+         {
+         static char *cacheOptions[_optionsLength];
+         static size_t filenamePos;
+         static size_t threadIDPos;
+
+         char timestr[_timeLength];
+         time_t timer = time(NULL);
+         snprintf(timestr, sizeof(timestr), "%i", (int32_t)timer % 100000);
+         
+         char tidstr[_threadIDLength];
+         snprintf(tidstr, sizeof(tidstr), "%d", syscall(SYS_gettid));
+
+         // Construct the output filename, with the thread id and time in seconds
+         char filename[_fileLength];
+         snprintf(filename, sizeof(filename), "%s.%s.%s.data", identifier ? identifier : "perf", tidstr, timestr);
+
+         // Build up the constant options, with env var override
+         if (filenamePos == 0) 
+            parseOptions(cacheOptions, filenamePos, threadIDPos); 
+
+         // Copy and specialize options
+         char *options[_optionsLength];
+         memcpy(options, cacheOptions, sizeof(options));
+         options[filenamePos] = filename;
+         options[threadIDPos] = tidstr;
+
+         // Dump the options
+         if (TR::Options::getVerboseOption(TR_VerbosePerformance))
+            {
+            TR_VerboseLog::vlogAcquire();
+            TR_VerboseLog::write("\nProfiling %s compile time with:", comp->signature());
+            char **iter = options;
+            while (*iter)
+               {
+               TR_VerboseLog::write(" %s", *iter);
+               iter++;
+               }
+            TR_VerboseLog::write("\n");
+            TR_VerboseLog::vlogRelease();
+            }
+
+         _pid = fork();
+         if (_pid == 0)
+            {
+            execv("/usr/bin/perf", options);
+            exit(0);
+            }
+
+         // Give perf some time to start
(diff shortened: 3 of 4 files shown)


charliegracie:
genie-omr build all

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * compile, such as a profiled very-hot or scorching compile, but profiling
+ * all compilations results in too much noise.
+ *
+ * This will fork and execute perf, configured such that it will
+ * profile the current compilation thread until a time limit
+ * is met or the current compilation ends.
+ *
+ * Perf is configured to collect call graph information using frame
+ * pointers at a high frequency, using the options:
+ *   --call-graph fp -m 4096 -F9999 -r 1
```
ymanton: Perhaps not a great idea to hard code some of these options, it will force people to edit the code and rebuild if they ever try this in an environment where the defaults don't work.

Regarding `--call-graph fp`, did you test this on something other than AMD64 or did you build with frame pointers enabled? This method of call stack sampling basically doesn't work on AMD64 unless you build your code with FP support, you'll just get bogus results; `--call-graph dwarf` is the more reliable option, but may not work on anything other than AMD64 unfortunately.

`-r` needs root or the user needs to have the right caps IIRC, and the specific `-m` and `-F` values you're using can fail you on some systems depending on how they're set up.

compiler/compile/OMRCompilation.cpp
```diff
@@ -97,6 +97,7 @@
 #include "control/Recompilation.hpp"           // for TR_Recompilation, etc
 #include "runtime/CodeCacheExceptions.hpp"
 #include "ilgen/IlGen.hpp"                     // for TR_IlGenerator
+#include "env/CompileTimeProfiler.hpp"         // for TR::CompileTimeProfiler
```
dsouzai: Would this build on non Linux platforms? I'm not sure `#include <sys/syscall.h>` exists on Windows, for example. You may want to guard this include with an `#if defined(LINUX)`.

compiler/compile/OMRCompilation.cpp
```diff
@@ -978,6 +979,9 @@ int32_t OMR::Compilation::compile()
       methodInfo->setProfileInfo(NULL);
 #endif
 
+   // Create the compile time profiler
+   TR::CompileTimeProfiler perf("compileTimePerf");
```
dsouzai: This would have to be guarded with an `#if defined(LINUX)` as well; either that or the implementation of `TR::CompileTimeProfiler` would need to moved to `compiler/x/env/CompileTimeProfiler.hpp` and an empty implementation would have to be added to `compiler/env/CompileTimeProfiler.hpp`.

compiler/env/CompileTimeProfiler.hpp
dsouzai: >Perhaps not a great idea to hard code some of these options, it will force people to edit the code and rebuild if they ever try this in an environment where the defaults don't work.

You could probably accept a string from the user, similar to  how it's done with the counts string option:

https://github.com/eclipse/omr/blob/28966b753f1c58bb25d8a56a311dc1cb065138e3/compiler/control/OMROptions.cpp#L192-L197

andrewcraik: I know from having been working with @ncough that we did do runs where we build the JIT with the frame pointers enabled so we got accurate profiling - hence the option. I agree either an option or an env var to set the string is worth while. These defaults have been very helpful so maybe they can be the default string if none is supplied.

compiler/compile/OMRCompilation.cpp
andrewcraik: Should this be guarded in some way to save creating it if it won't be used?

andrewcraik:
I think this is a great addition to our tools to find compile-time problems since it allows high fidelity profiling of individual compilations. I think this is Linux specific for now, but that I think it is still a useful tool regardless.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ *      The Eclipse Public License is available at
+ *      http://www.eclipse.org/legal/epl-v10.html
+ *
+ *      The Apache License v2.0 is available at
+ *      http://www.opensource.org/licenses/apache2.0.php
+ *
+ * Contributors:
+ *    Multiple authors (IBM Corp.) - initial implementation and documentation
+ ******************************************************************************/
```
0xdaryl: Please use the current license header.  This is the old one referencing EPL v1.0.

ncough: Thanks for pointing that out. This has been corrected.

compiler/env/CompileTimeProfiler.hpp
ncough: The implementation has been extended to accept arguments through an environment variable, with a set of sane defaults. 

@ymanton You are right, it was necessary to build with frame pointers enabled, as the particular machines tested at the time did not support dwarf. However, as this should be the exception, the default has been changed to dwarf.

compiler/compile/OMRCompilation.cpp
ncough: The implementation followed that of `RegionProfiler`, making use of RAII to control the duration of the profile and perform a runtime check internally. This has been modified to avoid the overhead in the disabled case.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,169 @@
...
+                  options[pos++] = iter;
+                  newArg = false;
+                  }
+               iter++;
+               }
+            }
+         else
+            {
+            options[pos++] = "--call-graph";
+            options[pos++] = "dwarf";
```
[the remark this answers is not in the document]
ncough: I was under the impression dwarf was always preferable, but deferring the choice to `perf` with management for AMD64 makes sense. The commit has been updated to reflect your suggestion.

compiler/compile/OMRCompilation.cpp
```diff
@@ -978,6 +982,13 @@ int32_t OMR::Compilation::compile()
       methodInfo->setProfileInfo(NULL);
 #endif
 
+#if defined(LINUX)
+   // Create the compile time profiler
+   TR::CompileTimeProfiler *perf;
```
andrewcraik: This needs to be set NULL for the if clean up to work properly in call cases since perf is not guaranteed to be 0.

ncough:
Rebased to resolve a conflict with a recent merge.

compiler/compile/OMRCompilation.cpp
```diff
@@ -992,6 +996,13 @@ int32_t OMR::Compilation::compile()
    if (_recompilationInfo)
       _recompilationInfo->startOfCompilation();
 
+#if defined(LINUX)
+   // Create the compile time profiler
+   TR::CompileTimeProfiler *perf = NULL;
+   if (self()->getOption(TR_CompileTimeProfiler))
+      perf = new (self()->trHeapMemory()) TR::CompileTimeProfiler(self(), "compileTimePerf");
```
dsouzai: I think in order for this to work, you need to add `TR_ALLOC` inside the class definition. For example, with `TR_Dominators`:

https://github.com/eclipse/omr/blob/34b0b3e14425eeadde434dd2496098d83cb2530d/compiler/optimizer/InductionVariable.cpp#L5521

https://github.com/eclipse/omr/blob/34b0b3e14425eeadde434dd2496098d83cb2530d/compiler/optimizer/Dominators.hpp#L50-L54

dsouzai: Nvm, turns out there are global definitions...

https://github.com/eclipse/omr/blob/34b0b3e14425eeadde434dd2496098d83cb2530d/compiler/env/TRMemory.hpp#L607-L617

andrewcraik:
@vijaysun-omr over to you

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+            {
+            printf(" %s", *iter);
+            iter++;
+            }
+         printf("\n");
+
+         _pid = fork();
+         if (_pid == 0)
+            {
+            execv("/usr/bin/perf", options);
```
0xdaryl: I haven't thought this through but I'll throw it out there while noodling it over and to invite other comments.  I'm concerned we're creating a security hole here, though how to exploit it I haven't decided.  If someone were to replace /usr/bin/perf with their own malicious application (not easy, but theoretically possible) what kind of mayhem could ensue with the privileges it is granted from the JVM process?  The JIT option that controls this is enabled in production builds (i.e., its not a debug option), hence you can force this code to run from any Linux build.  I'd like some reassurance that we're not creating an opportunity for exploitation here, and if we are we need to protect ourselves.  Thinking...

0xdaryl: Having said that though, in order to add this option to the command-line or JIT environment you'll have to be able to launch the runtime process anyway.  So maybe you're no worse off.  This code might be ok, I just have an uneasy feeling about it.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+
+#include <sys/wait.h>
+#include <sys/types.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+
```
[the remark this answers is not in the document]
ncough: I have gone for a LINUX ifdef wrapping the implementation, with a stub class for the rest. Thanks for the suggestion.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * utilities could be introduced. 
+ *
+ * In testing, the initial portion of the compile was lost, due to perf's
+ * start up. A delay has been added to manage this.
+ */
+class CompileTimeProfiler
+   {
+public:
+   CompileTimeProfiler(TR::Compilation *comp, const char *identifier)
+      {
```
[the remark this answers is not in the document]
ncough: I have shifted these into `static const` fields and made the relationships clear.

compiler/control/OMROptions.cpp
```diff
@@ -1026,6 +1026,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"privatizeOverlaps",  "O\tif BCD storageRefs are going to overlap then do the move through a temp", SET_OPTION_BIT(TR_PrivatizeOverlaps), "F"},
    {"profile",            "O\tcompile a profiling method body", SET_OPTION_BIT(TR_Profile), "F"},
    {"profileMemoryRegions", "I\tenable the collection of scratch memory profiling data", SET_OPTION_BIT(TR_ProfileMemoryRegions), "F" },
+   {"profileCompileTime",   "I\tgenerate a perf report for a specific compilation", SET_OPTION_BIT(TR_CompileTimeProfiler), "F" },
```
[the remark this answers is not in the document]
ncough: Thanks, I have fixed this.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+      _pid = 0;
+      if (comp->getOption(TR_CompileTimeProfiler))
+         {
+         char timestr[7];
+         time_t timer = time(NULL);
+         sprintf(timestr, "%i", (int32_t)timer % 100000);
```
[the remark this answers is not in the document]
ncough: Originally, it was because the string representation of the integer couldn't exceed the bounds. I have made it `snprintf` anyway.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+#if defined(TR_TARGET_X86) && defined(TR_TARGET_64BIT)
+            options[pos++] = "dwarf";
+#endif
+            options[pos++] = "-F9999";
+            }
+
+         options[pos++] = NULL;
+
+         // Dump the arguments
+         printf("\nProfiling %s compile time with:", comp->signature());
```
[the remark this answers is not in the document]
ncough: These have been shifted to the verbose logs, behind an option.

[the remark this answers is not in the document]
ncough: Concerning both this suggestion and stashing the options, this tool is most useful when targeting a single or small number of compiles, which is why this wasn't really considered. Moreover, this output (now in the verbose log) is the only way to link a compilation to the results. If these reasons aren't sufficient, a more significant filename could be generated and this output could be dumped only the first time.

compiler/env/CompileTimeProfiler.hpp
ncough: I am not aware of an approach to easily make this more secure. I am happy to hide it behind a build option, if that would help.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+      if (_pid != 0)
+         {
+         // Kill perf at the end of the compile
+         kill(_pid, SIGINT);
+         waitpid(_pid, NULL, 0);
+         _pid = 0;
+         }
+      }
+
+   ~CompileTimeProfiler()
```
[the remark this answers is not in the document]
ncough: The class has been reverted to its original design. Its used as a stack allocated object, as it is sufficiently light even in its disabled state. This should ensure the destructor is called.

compiler/env/CompileTimeProfiler.hpp
```diff
@@ -0,0 +1,171 @@
...
+         size_t pos = 0;
+         options[pos++] = "perf";
+         options[pos++] = "record";
+         options[pos++] = "-o";
+         options[pos++] = filename;
+         options[pos++] = "-t";
+         options[pos++] = tidstr;
+
+         static char *envOptions = feGetEnv("TR_CompileTimeProfiler");
+         if (envOptions)
```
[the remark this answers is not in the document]
ncough: Options are now cached, as suggested.

compiler/env/CompileTimeProfiler.hpp
ymanton: In order to replace `/usr/bin/perf` you'd have to acquire root or equivalent privileges, at which point you have the keys to the kingdom, so this doesn't seem like a hole in that context. Programs should be able to execute other programs, particularly those in secure locations like `/usr`, and OpenJ9 already allows you to invoke arbitrary commands via `-Xdump`.

ncough:
@0xdaryl Thank you for the suggestions and review. I believe I have addressed the issues. If you get the opportunity, could you review the corrected version?

compiler/env/CompileTimeProfiler.hpp
ncough: Considering @ymanton's contribution, this seems to be safe due to the fully specified path and the use of a command line option to enable the feature.

andrewcraik:
@0xdaryl over to you on this one I think.

andrewcraik:
@vijaysun-omr and @0xdaryl can we get the review of this one looked at again now that the review requests have been addressed by @ncough?

0xdaryl:
genie-omr build all

vijaysun-omr:
Looks okay to me. Merging since tests have passed now.

PR #2140, opened 2017-12-12
Extend memory region profiler

ncough:
This change simplifies use of the memory region
profiler and extends its use to codegen and the
general compilation strategy.

### DIFF
--- a/compiler/env/RegionProfiler.hpp
+++ b/compiler/env/RegionProfiler.hpp
@@ -48,13 +48,21 @@ namespace TR {
 class RegionProfiler
    {
 public:
-   RegionProfiler(TR::Region &region, TR::Compilation &compilation, const char *identifier) :
+   RegionProfiler(TR::Region &region, TR::Compilation &compilation, const char *format, ...) :
       _region(region),
       _initialRegionSize(_region.bytesAllocated()),
       _initialSegmentProviderSize(_region._segmentProvider.bytesAllocated()),
-      _compilation(compilation),
-      _identifier(identifier ? identifier : "[ANONYMOUS]")
+      _compilation(compilation)
       {
+      if (_compilation.getOption(TR_ProfileMemoryRegions))
+         {
+         va_list args;
+         va_start(args, format);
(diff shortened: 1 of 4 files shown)


compiler/env/RegionProfiler.hpp
```diff
@@ -89,7 +94,7 @@ class RegionProfiler
    size_t const _initialRegionSize;
    size_t const _initialSegmentProviderSize;
    TR::Compilation &_compilation;
-   const char * const _identifier;
+   char _identifier[512];
```
fjeremic: Is this the size per region profilier? This is just to hold the description string right? If so 512 seems like a very large number given most of the current strings are "comp/opt", "comp/codegen", etc.

charliegracie:
genie-omr build all

compiler/env/RegionProfiler.hpp
ncough: Good point. I have cut it back to the original 256.

compiler/env/RegionProfiler.hpp
```diff
@@ -48,13 +48,20 @@ namespace TR {
...
-      _compilation(compilation),
-      _identifier(identifier ? identifier : "[ANONYMOUS]")
+      _compilation(compilation)
       {
+      if (_compilation.getOption(TR_ProfileMemoryRegions))
+         {
+         va_list args;
+         va_start(args, format);
+         vsnprintf(_identifier, sizeof(_identifier), format, args);
+         _identifier[sizeof(_identifier) - 1] = '\0';
```
andrewcraik: Should there be an ASSERT or similar here if the supplied result was truncated? Not sure if we can tell, but asserting when there is a truncation rather than silently truncating would seem desirable if it isn't a huge amount of work.

ncough: An assert has been added based on the return value of `vsnprintf`.

andrewcraik:
@vijaysun-omr over to you

vijaysun-omr:
Looks okay to me. Merging.

PR #2142, opened 2017-12-12
Fix compiler warning on older compilers

charliegracie:


charliegracie:
genie-omr build all

fvtest/threadtest/abortTest.cpp
```diff
@@ -215,7 +215,7 @@ TEST(ThreadAbortTest, Waiting)
 		J9ThreadAbstractMonitor *mon = (J9ThreadAbstractMonitor *)testdata.waitSync;
 		EXPECT_TRUE(NULL == mon->waiting);
 		EXPECT_TRUE(NULL == mon->blocking);
-		EXPECT_EQ(0, mon->count);
+		EXPECT_TRUE(0 == mon->count);
```
Leonardo2718: Out of curiosity, what warning does this fix?

charliegracie: It is a signed vs unsigned comparison.  It is due to the older compiler version running on RHEL 6 by default and the macros used to generate the EXPECT_EQ.

Issue #2144, opened 2017-12-13
`comptest` fails to build on `Z`


rbanerjee:
Trying to build `comptest` on `zlinux_s390x`generates the following error -

```
[ 92%] Built target jitbuilder
[ 96%] Built target tril
[ 96%] Built target omrGtest
Linking CXX executable comptest
../../jitbuilder/libjitbuilder.a(OMRInstruction.cpp.o): In function `OMR::Z::Instruction::Instruction(TR::CodeGenerator*, OMR::InstOpCode::Mnemonic, TR::Node*)':
OMRInstruction.cpp:(.text+0x316): undefined reference to `OMR::InstOpCode::getMnemonicName(OMR::InstOpCode::Mnemonic)'
../../jitbuilder/libjitbuilder.a(OMRInstruction.cpp.o): In function `OMR::Z::Instruction::Instruction(TR::CodeGenerator*, TR::Instruction*, OMR::InstOpCode::Mnemonic, TR::Node*)':
OMRInstruction.cpp:(.text+0x51a): undefined reference to `OMR::InstOpCode::getMnemonicName(OMR::InstOpCode::Mnemonic)'
collect2: ld returned 1 exit status
make[3]: *** [fvtest/compilertriltest/comptest] Error 1
make[2]: *** [fvtest/compilertriltest/CMakeFiles/comptest.dir/all] Error 2
make[1]: *** [fvtest/compilertriltest/CMakeFiles/comptest.dir/rule] Error 2
make: *** [comptest] Error 2
```

Machine specifications:
```
Architecture:          s390x
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Big Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s) per book:    1
Book(s):               8
Vendor ID:             IBM/S390
BogoMIPS:              21881.00
Hypervisor:            z/VM 6.4.0
Hypervisor vendor:     IBM
Virtualization type:   full
Dispatching mode:      horizontal
```

Building on `gcc 4.4.6`.

Potentially related to this: https://github.com/eclipse/omr/pull/2012

Tagging the participants from https://github.com/eclipse/omr/pull/2012:
@sehirst  @fjeremic @charliegracie

dchopra001:
@fjeremic 
Looks like it's due to this minumum ALS commit: https://github.com/eclipse/omr/pull/2114

[...]

PR #2145, opened 2017-12-13
Remove calls to `getMnemonicName` in `z` codegen

rbanerjee:
There are a couple of Assertions in OMRInstruction.cpp
which make calls to `OMR::InstOpCode::getMnemonicName`.
However, `getMnemonicName` is yet to be implemented on `Z`
and therefore these calls generate compile time failures.

This commit replaces the said calls by using an alternate API.

Closes: #2144

Signed-off-by: Rwitaban (Ray) Banerjee rayb@ca.ibm.com

### DIFF
--- a/compiler/z/codegen/OMRInstruction.cpp
+++ b/compiler/z/codegen/OMRInstruction.cpp
@@ -85,7 +85,7 @@ OMR::Z::Instruction::Instruction(TR::CodeGenerator* cg, TR::InstOpCode::Mnemonic
    OMR::Instruction(cg, op, node),
    CTOR_INITIALIZER_LIST
    {
-   TR_ASSERT(cg->getS390ProcessorInfo()->supportsArch(_opcode.getMinimumALS()), "Processor detected does not support instruction %s\n", TR::InstOpCode::getMnemonicName(op));
+   TR_ASSERT(cg->getS390ProcessorInfo()->supportsArch(_opcode.getMinimumALS()), "Processor detected does not support instruction %s\n", cg->getDebug()? cg->getDebug()->getOpCodeName(&_opcode) : "(unknown)");
 
    self()->initialize();
    }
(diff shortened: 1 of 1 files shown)


rbanerjee:
I'd appreciate a review from @fjeremic please. This primarily impacts Z.

compiler/z/codegen/OMRInstruction.cpp
```diff
@@ -85,7 +85,7 @@ OMR::Z::Instruction::Instruction(TR::CodeGenerator* cg, TR::InstOpCode::Mnemonic
    OMR::Instruction(cg, op, node),
    CTOR_INITIALIZER_LIST
    {
-   TR_ASSERT(cg->getS390ProcessorInfo()->supportsArch(_opcode.getMinimumALS()), "Processor detected does not support instruction %s\n", TR::InstOpCode::getMnemonicName(op));
+   TR_ASSERT(cg->getS390ProcessorInfo()->supportsArch(_opcode.getMinimumALS()), "Processor detected does not support instruction %s\n", cg->getDebug()->getOpCodeName(&this->getOpCode()));
```
rbanerjee: Note to reviewers: The Travis Lint failure seems to be related  to `(&this->getOpCode..`. 

However the usage of `this` here seems to be consistent with the code around it. So my guess would be that it's not relevant.

fjeremic: Will using `self()` work here?

0xdaryl: Just use `_opcode` to be consistent with the `getMinimalALS` query earlier in the line.

rbanerjee:
@0xdaryl @fjeremic  ptal

0xdaryl:
This is a Z-specific change.  Genie builds not required on non-Z platforms.

genie-omr build zlinux zos

Issue #1887, opened 2017-10-27
Create the opcodes `breturn`, `bureturn` and evaluators


mgaudet:
Based on discussion in #1842, it seems that to help raise `Int8` to a first class citizen, the right answer is to implement `breturn and `bureturn` to be used in functions expected to return an `Int8`

This task should involve: 

- [ ] Add the opcodes to the OpCode enum, 
- [ ] Fill in the opcode properties table for them.
- [ ] Change the implementation of `OMR::ILOpCode::returnOpCode()` to use `breturn`. 
- [ ] Add an evaluator for X
- [ ] Add an evaluator for P
- [ ] Add an evaluator for Z
- [ ] Add an evaluator for ARM
- [ ] Add test cases to exercise the new opcode and evaluator.

mgaudet:
Closing this for now, pending a discussion in #1901

mgaudet:
Reopening based on discussion in #1901

PR #2147, opened 2017-12-14
WIP: Create the opcode breturn

rbanerjee:
Add `breturn` to the list of available  `ILOpCode`s and implement the required
codegen support on `X`, `P`, `Z`  and `arm`. 

This is part of an effort to raise `TR::Int8` to a first class citizen.

Please see https://github.com/eclipse/omr/issues/1901 for the relevant discussion. 

Fixes Issue: #1887 (NOTE: Does not address the `bureturn` aspect of it)

Signed-off-by: Rwitaban (Ray) Banerjee rayb@ca.ibm.com

### DIFF
--- a/compiler/arm/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/arm/codegen/ControlFlowEvaluator.cpp
@@ -72,7 +72,7 @@ TR::Register *OMR::ARM::TreeEvaluator::gotoEvaluator(TR::Node *node, TR::CodeGen
    return NULL;
    }
 
-// also handles areturn
+/* Also handles areturn and breturn. */
 TR::Register *OMR::ARM::TreeEvaluator::ireturnEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    TR::Register *returnRegister = cg->evaluate(node->getFirstChild());
@@ -81,7 +81,17 @@ TR::Register *OMR::ARM::TreeEvaluator::ireturnEvaluator(TR::Node *node, TR::Code
    addDependency(deps, returnRegister, cg->getProperties().getIntegerReturnRegister(), TR_GPR, cg);
 
    generateAdminInstruction(cg, ARMOp_ret, node, deps);
-   cg->comp()->setReturnInfo(TR_IntReturn);
+   switch (node->getDataType())
+     {
+     case TR::Int8:
+       cg->comp()->setReturnInfo(TR_ByteReturn);
+       break;
+     case TR::Int64:
+       cg->comp()->setReturnInfo(TR_IntReturn);
+       break;
+     default:
+       TR_ASSERT(false, "The ireturnEvaluator in arm should only be used when the data being returned is of type TR::Int8 || TR::Int32");
+     }
    return NULL;
    }
 
@@ -100,7 +110,7 @@ TR::Register *OMR::ARM::TreeEvaluator::lreturnEvaluator(TR::Node *node, TR::Code
    return NULL;
    }
 
-// areturn handled by ireturnEvaluator
+// breturn and areturn handled by ireturnEvaluator
 
 TR::Register *OMR::ARM::TreeEvaluator::returnEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
--- a/compiler/compile/CompilationTypes.hpp
+++ b/compiler/compile/CompilationTypes.hpp
@@ -81,21 +81,22 @@ enum TR_CallingContext {
 
 
 
-// function return type flags for word preceding the entry point in any
+// function return types for word preceding the entry point in any
 // linkage that needs to be callable from interpreted.  Also used in pic
 // for use when dispatching methods that are interpreted in megamorphic send
 // case
 enum TR_ReturnInfo
    {
-   TR_VoidReturn        = 0,
-   TR_IntReturn         = 1,
-   TR_LongReturn        = 2,
-   TR_FloatReturn       = 3,
-   TR_DoubleReturn      = 4,
-   TR_ObjectReturn      = 5, // Currently 64-bit platforms only
-   TR_FloatXMMReturn    = 6, // IA32 and AMD64 only
-   TR_DoubleXMMReturn   = 7, // IA32 and AMD64 only
-   TR_ConstructorReturn = 8  // Return from constructor
+   TR_VoidReturn,
+   TR_ByteReturn,
+   TR_IntReturn,
+   TR_LongReturn,
+   TR_FloatReturn,
+   TR_DoubleReturn,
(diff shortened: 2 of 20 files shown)


compiler/arm/codegen/ControlFlowEvaluator.cpp
```diff
@@ -81,7 +81,14 @@ TR::Register *OMR::ARM::TreeEvaluator::ireturnEvaluator(TR::Node *node, TR::Code
    addDependency(deps, returnRegister, cg->getProperties().getIntegerReturnRegister(), TR_GPR, cg);
 
    generateAdminInstruction(cg, ARMOp_ret, node, deps);
-   cg->comp()->setReturnInfo(TR_IntReturn);
+   if (node->getDataType() == TR::Int8)
```
andrewcraik: This would be much more elegant as a switch on node->getDataType() with unsupported types hitting a fatal assertion with an informative message.

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1265,7 +1268,14 @@ TR::Register *OMR::X86::TreeEvaluator::integerReturnEvaluator(TR::Node *node, TR
          }
       else
          {
-         comp->setReturnInfo(TR_IntReturn);
+         if (node->getDataType() == TR::Int8)
```
andrewcraik: Another good place for a switch to be dropped in rather than an if ladder

compiler/arm/codegen/ControlFlowEvaluator.cpp
mstoodle: ...particularly since there's probably a follow-on PR to add support for sreturn coming, I would guess :)

mstoodle:
@rbanerjee just a ping to see if you'll be able to make the two small updates requested by @andrewcraik ?

rbanerjee:
@mstoodle Thanks for the ping!

@andrewcraik  @mstoodle  I have updated the PR to address your comments. Please do let me know if I missed anything.
(There was also a bug in the `p` codegen evaluator for `breturn`, which is fixed now.)

rbanerjee:
Testing is done. Removing the `WIP` tag.

@0xdaryl  ptal

0xdaryl:
@genie-omr build all

rbanerjee:
@0xdaryl this is ready for review again. Please take a look.

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1149,6 +1149,24 @@ OMR::Z::TreeEvaluator::returnEvaluator(TR::Node * node, TR::CodeGenerator * cg)
          comp->setReturnInfo(TR_ObjectReturn);
          dependencies->addPostCondition(returnValRegister, linkage->getIntegerReturnRegister());
          break;
+      case TR::breturn:
+         dependencies->addPostCondition(returnValRegister, linkage->getIntegerReturnRegister());
+         comp->setReturnInfo(TR_ByteReturn);
+         if (linkage->isNeedsWidening())
+            {
+            /* LGBR equiv to {Operands R_1,R_2 : Load Byte (64<-8)} */
+            new (cg->trHeapMemory()) TR::S390RRInstruction(TR::InstOpCode::LGBR,
```
fjeremic: Please avoid using placement new. We have helpers which may perform extra checks via `generateRRInstruction`. There are many examples of uses of this API throughout the Z codegen.

compiler/compile/CompilationTypes.hpp
```diff
@@ -81,21 +81,22 @@ enum TR_CallingContext {
...
+   TR_VoidReturn,
+   TR_ByteReturn,
+   TR_IntReturn,
+   TR_LongReturn,
+   TR_FloatReturn,
+   TR_DoubleReturn,
+   TR_ObjectReturn,     // Currently 64-bit platforms only
+   TR_FloatXMMReturn,   // IA32 and AMD64 only
+   TR_DoubleXMMReturn,  // IA32 and AMD64 only
+   TR_ConstructorReturn // Return from constructor
```
0xdaryl: I believe OpenJ9 encodes this value as part of the linkage info word in the code cache before the JITed entry point.  I don't think you can arbitrarily insert a new value in this enum without synchronizing the change with the VM in OpenJ9.  Please check this--maybe I'm wrong.

I realize this is an OpenJ9 concern but since the code for this lives in OMR (for now) it should be addressed here.

0xdaryl: Also, I like the explicit enumeration of these values.  It makes it much easier to decode.  Please keep the explicit values.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -800,6 +800,29 @@ TR::Register *OMR::Power::TreeEvaluator::gotoEvaluator(TR::Node *node, TR::CodeG
...
+/**
+ * (breturn TR::Int8) ≡ (ireturn (b2i (TR::Int8)))
+ */
+TR::Register *OMR::Power::TreeEvaluator::breturnEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR::Compilation *comp = cg->comp();
+   TR::Node *child  = node->getFirstChild();
+   TR::Register *returnRegister = cg->evaluate(child);
+   /**
+    * Extend the signed eight-bit value to form a 32 bit word, copying the data in place.
```
0xdaryl: eight -> 8 for consistency

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1228,7 +1231,18 @@ TR::Register *OMR::X86::TreeEvaluator::integerReturnEvaluator(TR::Node *node, TR
          }
       else
          {
-         comp->setReturnInfo(TR_IntReturn);
+         switch (node->getDataType())
+            {
+            default:
+              TR_ASSERT(0, "On non-64bit target platforms, the only supported return types are TR::Int8 and TR::Int32");
```
0xdaryl: What about TR::Address ?

compiler/compile/CompilationTypes.hpp
0xdaryl: I just realized it isn't a fair request on my part to ask you to delve into OpenJ9 to figure out whether or not this will break anything there since this PR is not being made in support of OpenJ9 work.  Sorry about that.  Given that, I'll ask that you move `TR_ByteReturn` to the end of this enum to minimize potential downstream churn and replace the explicit enumerator values that you deleted.

0xdaryl:
Hi @rbanerjee.  What is the current status of this PR?  Would you consider closing it while you work on addressing the review comments and then re-open when ready?

PR #2148, opened 2017-12-14
Fix unresolved case in duplicate check

ncough:
The method `isUnsafeToDuplicateAndExecuteAgain` on
nodes will determine whether its safe to duplicate
a node and its children. However, unresolved
symbol references were treated as safe for duplication,
even though they should remain under their resolve
checks.


charliegracie:
genie-omr build all

PR #2149, opened 2017-12-14
Enable optimizations with JProfiling

ncough:
This change enables various optimizations
that were disabled under JitProfiling due
to issues with the duplication it performed,
such as the increased node count or its
limitations on asyncchecks.

### DIFF
--- a/compiler/optimizer/AsyncCheckInsertion.cpp
+++ b/compiler/optimizer/AsyncCheckInsertion.cpp
@@ -120,7 +120,7 @@ bool TR_AsyncCheckInsertion::shouldPerform()
    {
    // Don't run when profiling
    //
-   if (comp()->isProfilingCompilation() || comp()->generateArraylets())
+   if (comp()->getProfilingMode() == JitProfiling || comp()->generateArraylets())
       return false;
 
    // It is not safe to add an asynccheck under involuntary OSR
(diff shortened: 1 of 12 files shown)


andrewcraik:
LGTM from me - over to @vijaysun-omr for a review since this is probably in his area of expertise

compiler/optimizer/CompactLocals.cpp
```diff
@@ -90,6 +90,9 @@ int32_t TR_CompactLocals::perform()
       return 0;
       }
 
+   if (comp()->getProfilingMode() == JitProfiling)
```
vijaysun-omr: Unlike the other changes in this commit this seems to be a new optimization that has been disabled under JitProfiling (?)

vijaysun-omr:
genie-omr build all

compiler/optimizer/CompactLocals.cpp
ncough: This optimization was disabled using strategy flags in the past. The initial version of this commit shifted these checks to the optimizations themselves, as they weren't part of the strategy, rather they were a restriction due to the profiling implementation.

The change is being reverted to use strategy flags, as this makes the disable clearer.

ncough:
This change now enables optimizations under JProfiling through the use of optimization flags, in particular `IfNotJitProfiling` and `IfEnabledAndNotJitProfiling`.

andrewcraik:
The travis failure seems odd - 10 mins inactivity?

I prefer this approach and think this looks good.

andrewcraik:
@vijaysun-omr over to you

vijaysun-omr:
This commit does look much more straightforward. However since there are significant changes, I will do another genie build.

vijaysun-omr:
genie-omr build all

vijaysun-omr:
Merging since the CI issue not a real one.

PR #2150, opened 2017-12-15
Add options to control JProfiler thread

ncough:
JProfiling includes an optional thread to
manage its value profiling during execution.
This change adds options to disable it.


andrewcraik:
LGTM from me @0xdaryl or @vijaysun-omr could one of you review & merge please? This is a simple option addition.

PR #2151, opened 2017-12-15
Deprecate unused instruction properties in Z codegen

fjeremic:
Some instruction properties in the Z codegen are not used and are
slated for deprecation. We currently do not envision any of the removed
properties to be used by any downstream projects now or in the future
thus eliminating stale properties allows us to minimize the size of the
instruction opcode tables and consolidate properties and properties2
static fields.

### DIFF
--- a/compiler/z/codegen/OMRInstOpCode.hpp
+++ b/compiler/z/codegen/OMRInstOpCode.hpp
@@ -400,19 +400,20 @@ namespace Z
 #define   VSI_FORMAT    59
 
 /* Instruction Properties (One hot encoding) */
+#define S390OpProp_None                   static_cast<uint64_t>(0x0000000000000000ull)
 #define S390OpProp_UsesTarget             static_cast<uint64_t>(0x0000000000000001ull)
 #define S390OpProp_SingleFP               static_cast<uint64_t>(0x0000000000000002ull)
 #define S390OpProp_DoubleFP               static_cast<uint64_t>(0x0000000000000004ull)
 #define S390OpProp_SetsZeroFlag           static_cast<uint64_t>(0x0000000000000008ull)
 #define S390OpProp_SetsSignFlag           static_cast<uint64_t>(0x0000000000000010ull)
-#define S390OpProp_SetsCarryFlag          static_cast<uint64_t>(0x0000000000000020ull)
+// Available                              static_cast<uint64_t>(0x0000000000000020ull)
 #define S390OpProp_SetsOverflowFlag       static_cast<uint64_t>(0x0000000000000040ull)
 #define S390OpProp_BranchOp               static_cast<uint64_t>(0x0000000000000080ull)
 #define S390OpProp_IsLoad                 static_cast<uint64_t>(0x0000000000000100ull)
 #define S390OpProp_IsStore                static_cast<uint64_t>(0x0000000000000200ull)
-#define S390OpProp_IsSync                 static_cast<uint64_t>(0x0000000000000400ull)
-#define S390OpProp_ReadsMultipleRegs      static_cast<uint64_t>(0x0000000000000800ull)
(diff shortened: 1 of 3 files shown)


fjeremic:
Note to reviewers that as part of this PR the intention is only to remove unused properties. In a subsequent PR we will clean up the `properties` and `properties2` fields and further consolidate the lists. I wanted to do this in steps so as to make sure no bugs are introduced and to make the commits easier for peer review.

compiler/z/codegen/OMRInstOpCode.hpp
```diff
@@ -631,13 +631,11 @@ class InstOpCode: public OMR::InstOpCode
    uint64_t is32bit() {return properties[_mnemonic] & S390OpProp_Is32Bit;}
    uint64_t is32to64bit() {return properties[_mnemonic] & S390OpProp_Is32To64Bit;}
 
-   uint64_t usesImplicitRegisters() { return properties[_mnemonic] & S390OpProp_UsesImplicitRegisters;}
    uint64_t hasLongDispSupport() {return properties[_mnemonic] & S390OpProp_LongDispSupported;}
    uint64_t usesRegPairForTarget() {return properties[_mnemonic] & S390OpProp_UsesRegPairForTarget; }
-   uint64_t mayUseRegPairForTarget() { return properties[_mnemonic] & S390OpProp_MayUseRegPairForTarget; }
    uint64_t usesRegPairForSource() {return properties[_mnemonic] & S390OpProp_UsesRegPairForSource; }
    uint64_t usesRegRangeForTarget(){return properties[_mnemonic] & S390OpProp_UsesRegRangeForTarget; }
-   uint64_t canUseRegPairForTarget() {return usesRegPairForTarget() || usesRegRangeForTarget() || mayUseRegPairForTarget(); }
```
fjeremic: `mayUseRegPairForTarget` has been deprecated as it was only set on a single instruction. This instruction was `SIGP` which _may_ set either the first or the second register in the first operand, whichever is odd. To make things simple we enforce a restriction that the first operand must be an even-odd register pair, such that we always know `SIGP` will use the odd register of the respective pair.

compiler/z/codegen/OMRInstOpCode.hpp
```diff
@@ -659,33 +657,18 @@ class InstOpCode: public OMR::InstOpCode
...
    uint64_t readsCC() {return properties[_mnemonic] & S390OpProp_ReadsCC;}
 
-   uint64_t readsMultipleRegs() {return properties[_mnemonic] & S390OpProp_ReadsMultipleRegs;}
-   uint64_t setsMultipleRegs() {return properties[_mnemonic] & S390OpProp_SetsMultipleRegs;}
    uint64_t setsZeroFlag() {return properties[_mnemonic] & S390OpProp_SetsZeroFlag;}
    uint64_t setsSignFlag() {return properties[_mnemonic] & S390OpProp_SetsSignFlag;}
-   uint64_t setsCarryFlag() {return properties[_mnemonic] & S390OpProp_SetsCarryFlag;}
    uint64_t setsOverflowFlag() {return properties[_mnemonic] & S390OpProp_SetsOverflowFlag;}
    uint64_t setsCompareFlag() {return properties[_mnemonic] & S390OpProp_SetsCompareFlag;}
+   uint64_t setsCarryFlag() { return setsZeroFlag(); }
```
fjeremic: Carry and zero flags are synonymous. All instructions that have either have both, so the flags can be merged. I've left the query here because depending on context we may be talking about either carry or zero flag. Even though they map to the same condition code from a code reading perspective it helps to talk about the correct flag depending on context.

joransiu:
LGTM

0xdaryl:
Files modified are Z codegen specific.  No need for Power tests.

genie-omr zlinux zos

0xdaryl:
Files modified are Z codegen specific. No need for Power tests.

genie-omr build zlinux zos

PR #2153, opened 2017-12-15
Fix option enum collision

ncough:
JProfiling added a new option in a prior commit,
replacing a value marked as available in the
options enum. However, this value was already
taken. This change removes the collision.


0xdaryl:
genie-omr build all

PR #2154, opened 2017-12-18
Disable redefinition of 'remove' for z/TPF Platform

jjohnst-us:
z/TPF builds runtimes with an intermediary character
translation library which uses intercepting headers
to redefine system functions which require translation
from ascii to ebcdic. However, for some scenarios
translations is not desired. The 'remove' member function
is one of those instances where we don't want to
redefine to do translations.

### DIFF
--- a/compiler/infra/Checklist.hpp
+++ b/compiler/infra/Checklist.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2017 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -21,6 +21,10 @@
 
 #ifndef OMR_CHECKLIST_INCL
 #define OMR_CHECKLIST_INCL
+
+#ifdef OMRZTPF
+#define __TPF_DO_NOT_MAP_ATOE_REMOVE
+#endif
 
 class TR_BitVector;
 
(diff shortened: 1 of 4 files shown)


0xdaryl:
I'm not questioning the need for this change, but I am trying to understand the placement of this macro in this eclectic group of files and why other files weren't chosen.  Is the problem with *any* member function named `remove` regardless of class?  Does this macro need to accompany the definition, declaration, or use of such `remove` functions?

0xdaryl:
genie-omr build all

jjohnst-us:
@0xdaryl good question.  So I took the pragmatic approach of adding the macros wherever a compilation failure occurred.  The complicating factor is that the wrapping "ascii to ebcdic" header could have been processed before the class with the defining problematic 'remove' is processed such that it wouldn't see the 'swtich' macro to preven the a2e header from redefining remove.

jjohnst-us:
Keep in mind because its a z/TPF platform only update it won't affect other platforms.

jjohnst-us:
@fjeremic @0xdaryl 
I believe this is a JIT related update as well, would you be the reviewer for this?

Although there is one GC segment update in this commit.

jjohnst-us:
@fjeremic Hi felipe, I assume you would be doing the merge on this one as well?  tia..

fjeremic:
> @fjeremic Hi felipe, I assume you would be doing the merge on this one as well? tia..

Unfortunately no, I'm not a committer at the OMR project. We'll have to wait for one of the committers to pick up this PR.

jjohnst-us:
@charliegracie when you get a chance, thanks charlie.

jjohnst-us:
@charliegracie Hey Charlie, was there anyone else you wanted to review the Pull Request?  tia..

charliegracie:
Sorry this was tagged as a compiler PR so it slipped off of my radar.

PR #2155, opened 2017-12-18
Enable project root and system root variables for z/TPF Platform

jjohnst-us:
Introduce makefile variables to hold project and system root
variables which specify include base directories for OMR
compilations.

### DIFF
--- a/omrmakefiles/rules.ztpf.mk
+++ b/omrmakefiles/rules.ztpf.mk
@@ -77,6 +77,9 @@ ifneq (,$(findstring executable,$(ARTIFACT_TYPE)))
 	GLOBAL_LDFLAGS+=$(DEFAULT_LIBS)
 endif
 
+ZTPF_ROOT?=/ztpf/commit
+PROJECT_ROOT?=/projects/jvmport/userfiles
+
 ###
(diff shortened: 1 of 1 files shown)


jjohnst-us:
@charliegracie 
Here is another PR where it really only affects z/TPF and should be considered low risk from another platforms perspective.

omrmakefiles/rules.ztpf.mk
```diff
@@ -77,6 +77,9 @@ ifneq (,$(findstring executable,$(ARTIFACT_TYPE)))
 	GLOBAL_LDFLAGS+=$(DEFAULT_LIBS)
 endif
 
+ZTPF_ROOT+=/ztpf/commit
```
charliegracie: This will add `/ztpf/commit` to the end of `ZTPF_ROOT`.  If someone overrides this option on the command line or in the environment you will end up concatenating this value to what the user set.  I think you wanted this `ZTPF_ROOT?=/ztpf/commit` and `PROJECT_ROOT?=/projects/jvmport/userfiles`

jjohnst-us:
@charliegracie Thanks Charlie.  Update made and rebased.

PR #2157, opened 2017-12-18
GC: Don't use scavenger cache size to pick TLH size when scavenger is off

rwy0717:
Originally, we always used the tlhMaximumSize from the extensions.
This was recently changed to use the scavenger cache size, but this only
compiles if the scavenger is enabled.

I've added an ifdef that resorts to the original behaviour if the
scavenger is off, fixing a compilation error.

### DIFF
--- a/gc/base/MemoryPoolAddressOrderedList.cpp
+++ b/gc/base/MemoryPoolAddressOrderedList.cpp
@@ -91,7 +91,11 @@ MM_MemoryPoolAddressOrderedList::initialize(MM_EnvironmentBase *env)
 
 #if defined(OMR_GC_THREAD_LOCAL_HEAP)
 	/* this memoryPool can be used by scavenger, maximum tlh size should be max(_extensions->tlhMaximumSize, _extensions->scavengerScanCacheMaximumSize) */
+#if defined(OMR_GC_MODRON_SCAVENGER)
 	uintptr_t tlhMaximumSize = OMR_MAX(_extensions->tlhMaximumSize, _extensions->scavengerScanCacheMaximumSize);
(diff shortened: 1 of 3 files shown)


rwy0717:
@nbhuiyan: This PR should fix your problem building the GC. Thanks for pointing it out.

charliegracie:
@amicic what are your thoughts?

charliegracie:
genie-omr build all

amicic:
yeah, looks right. @LinHu2016, we missed this detail in the original change

youngar:
@rwy0717 one spelling mistake in a comment: `OMR_GC_MODERON_SCAVENGER `

gc/base/MemoryPoolAddressOrderedList.cpp
```diff
@@ -91,7 +91,11 @@ MM_MemoryPoolAddressOrderedList::initialize(MM_EnvironmentBase *env)
 
 #if defined(OMR_GC_THREAD_LOCAL_HEAP)
 	/* this memoryPool can be used by scavenger, maximum tlh size should be max(_extensions->tlhMaximumSize, _extensions->scavengerScanCacheMaximumSize) */
+#if defined(OMR_GC_MODRON_SCAVENGER)
 	uintptr_t tlhMaximumSize = OMR_MAX(_extensions->tlhMaximumSize, _extensions->scavengerScanCacheMaximumSize);
+#else /* OMR_GC_MODERON_SCAVENGER */
```
charliegracie: Spelling OMR_GC_MODERON_SCAVENGER to OMR_GC_MODRON_SCAVENGER

PR #2163, opened 2017-12-20
Add JCL programs to static data dump list for z/TPF.

jjohnst-us:
z/TPF dumps out program static data areas for OMR and
OPENJ9 programs.  This PR extends the list to include
JCL programs as well.  This update enhances RAS for
z/TPF system cores.


jjohnst-us:
@charliegracie 
This should be considered low risk as this segment is only compiled for z/TPF platform.  We are simply adding to a list of program names which can only be recognized by z/TPF.  Let me know if you need any other information.

tia

PR #2164, opened 2017-12-20
Build liveness with real uses on OSR points

cathyzhyi:
OSRUses under prepareForOSR prolong the live range of autos and
pending push temps. Especially, those fake uses can cause many
autos and temps appear to be alive at the method entry and requires
zero initialization in the method prologue which is big overhead.

A way to solve this problem is to find out the real uses
of those autos and pending push temps at each OSRPoint and use them
during liveness analysis. The real uses can be found by figuring
out which symbols are needed to calculate the auto or temps at each
prepareForOSRCall and combining this with the liveness information
for the autos and temps from original bytecode.

This changeset changes liveness analysis to consider real uses at
each OSRPoint.

### DIFF
--- a/compiler/optimizer/DataFlowAnalysis.hpp
+++ b/compiler/optimizer/DataFlowAnalysis.hpp
@@ -775,12 +775,12 @@ class TR_LiveVariableInformation
    void findLocalUsesInBackwardsTreeWalk(TR::Node *node, int32_t blockNum,
                                          TR_BitVector *genSetInfo, TR_BitVector *killSetInfo,
                                          TR_BitVector *commonedLoads, vcount_t visitCount);
-   void findUseOfLocal(TR::Node *node, int32_t blockNum,
+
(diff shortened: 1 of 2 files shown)


cathyzhyi:
This PR is dependent on https://github.com/eclipse/omr/pull/2006

andrewcraik:
I'll take a look, but an opinion from @ncough would also be helpful.

ncough:
LGTM

vijaysun-omr:
genie-omr build all

cathyzhyi:
This change has passed tests and ready to be merged.

vijaysun-omr:
Could you please re-base this change ? I can merge once you do so.

cathyzhyi:
@vijaysun-omr  Rebased to the newest version. Thanks!

vijaysun-omr:
Tests have passed. Merging.

Issue #1941, opened 2017-11-02
comptest (VectorTest.cpp) generates an illegal instruction on zlinux_s390x


rbanerjee:
While downgrading the required compiler version for Tril Based compiler tests (`comptest`)
(https://github.com/eclipse/omr/pull/1900), I encountered the following when trying to run comptest on `zlinux_s390x` `using GCC 4.4.6` : 

```
[==========] Running 3400 tests from 28 test cases.
[----------] 1 test from CommoningDeathTest
[----------] 1 test from CommoningDeathTest (89 ms total)

[----------] 14 tests from ILValidatorDeathTest/IllformedTrees
[----------] 14 tests from ILValidatorDeathTest/IllformedTrees (160 ms total)

[----------] 1 test from VectorTest
Illegal instruction (core dumped)
```

`comptest` seems to at least run when excluding `VectorTest`:

```
fvtest/compilertriltest/comptest --gtest_filter=-VectorTest*

...

[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/14, where GetParam() = (('\xFF' (255), 0), ("bushr", 0x803c043c))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/28, where GetParam() = (('\xD6' (214), 0), ("bushr", 0x803c043c))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/42, where GetParam() = (('\xFF' (255), 0), ("bushr", 0x803c043c))
[  FAILED  ] ShiftAndRotateTest/UInt8ShiftAndRotate.UsingConst/56, where GetParam() = (('\xFE' (254), 0), ("bushr", 0x803c043c))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/14, where GetParam() = ((65535, 0), ("sushr", 0x803c044c))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/28, where GetParam() = ((65494, 0), ("sushr", 0x803c044c))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/42, where GetParam() = ((65535, 0), ("sushr", 0x803c044c))
[  FAILED  ] ShiftAndRotateTest/UInt16ShiftAndRotate.UsingConst/56, where GetParam() = ((65534, 0), ("sushr", 0x803c044c))

[...]

Issue #1843, opened 2017-10-24
Figure out how to do platform specific testing and test-disablement in Tril


mgaudet:
Some test cases will be written that cannot be run on all platforms, intentionally, leading to a need to disable/enable test cases on a per platform basis. 

[One option is to guard them with `#ifdef TR_TARGET` macros](https://github.com/eclipse/omr/pull/1805/files), however, this is coarse choice, and doesn't help distinguish between

* Tests temporarily disabled due to a known defect
* Tests disabled temporarily on a platform while waiting for implementation
* Tests disabled permanently on a platform because some feature doesn't exist, or never will.

mgaudet:
IMO, a lot of this needs to be done with a custom 'skip' patch to GoogleTest that aborts execution of a test case based on a predicate. Skips should then be reported based on categories: 

```
TEST(Foo,Bar) { 
   SKIP(platform != x86-64, PLATFORM)
   ... 
}

TEST(Foo,Bar) { 
   SKIP(bits == 32, KNOWN_BUG)
   ... 
}

[==========] 6490 tests from 43 test cases ran. (6507 ms total)
[==========] 12 tests skipped: 8 PLATFORM 4 KNOWN_BUG
[  PASSED  ] 6490 tests.
```

Leonardo2718:
I have opened #3539 to introduce a basic `SKIP_IF()` macro. It is not nearly as sophisticated as what's discussed in this issue but I'm hoping it will enable further work by introducing some basic machinery for skipping a test.

ehsankianifar:
I believe [this PR](https://github.com/eclipse-omr/omr/pull/7912) will fix this issue.

PR #2165, opened 2017-12-20
Disable VectorTest on Z

dchopra001:
Z12 and lower architectures don't have vector support. This causes
illegal instruction failures on Z. Since we cannot currently
determine at runtime which Z architecture the test is running on, the
tril test for VectorTest is disabled on Z altogether. We can re-enable
once we are able to check at runtime.

### DIFF
--- a/fvtest/compilertriltest/VectorTest.cpp
+++ b/fvtest/compilertriltest/VectorTest.cpp
@@ -39,7 +39,13 @@ TEST_F(VectorTest, VDoubleAdd) {
     auto trees = parseString(inputTrees);
 
     ASSERT_NOTNULL(trees);
-
+    //TODO: Re-enable this test on S390 after issue #1843 is resolved. 
+    //
+    // This test is currently disabled on Z platforms because not all Z platforms
+    // have vector support. Determining whether a specific platform has the support
+    // at runtime is currently not possible in tril. So the test is being disabled altogether
+    // on Z for now.
+#ifndef TR_TARGET_S390
     Tril::DefaultCompiler compiler{trees};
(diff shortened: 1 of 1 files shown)


dchopra001:
FYI This is a short term fix for the problem mention in https://github.com/eclipse/omr/issues/1941. The long term fix would be for tril to skip the test after figuring out that the machine it's running on doesn't have Vector support.

dchopra001:
Tagging @fjeremic for review.

fvtest/compilertriltest/VectorTest.cpp
```diff
@@ -39,7 +39,11 @@ TEST_F(VectorTest, VDoubleAdd) {
     auto trees = parseString(inputTrees);
 
     ASSERT_NOTNULL(trees);
-
+    // This test is currently disabled on Z platforms because not all Z platforms
+    // have vector support. Determining whether a specific platform has the support
+    // at runtime is currently not possible in tril. So the test is being disabled altogether
+    // on Z for now.
+#ifndef TR_TARGET_S390
```
fjeremic: Could we please add a TODO and in it reference the Issue which will eventually fix this problem? Otherwise this comment will get lost in the sands of time (and code).

dchopra001: Okay, so looks like #1843 is the actual issue that should resolve this problem once and for all. So I'll add the "Closes: #1941" tag for this problem as well since that problem should technically be solved by this PR.

dchopra001:
Tagging @rbanerjee and @mgaudet as well due to tie in with #1843 and #1941.

mgaudet:
LGTM

rbanerjee:
LGTM

0xdaryl:
genie-omr build all

Issue #2152, opened 2017-12-15
LinkageTest.cpp fails to compile on gcc 4.4.6 and 4.4.7


dchopra001:
I've tried compiling on x86 and z, and am getting the following compilation failures :

```
/gsa/tlbgsa/home/d/c/dchopra/development/omr/compiler/ras/DebugCounter.hpp: In constructor ‘TR::DebugCounterGroup::DebugCounterGroup(TR_PersistentMemory*)’:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/compiler/ras/DebugCounter.hpp:250: warning: deprecated conversion from string constant to ‘char*’
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp: In member function ‘void LinkageTest_SystemLinkageParameterPassingFourArg_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = double]’:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp:154:   instantiated from here
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp:118: error: address of overloaded function with no contextual type information
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp: In member function ‘void LinkageTest_SystemLinkageParameterPassingSingleArg_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = double]’:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp:154:   instantiated from here
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp:76: error: address of overloaded function with no contextual type information
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp: In member function ‘void LinkageTest_InvalidLinkageTest_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = double]’:
/gsa/tlbgsa/home/d/c/dchopra/development/omr/fvtest/compilertriltest/LinkageTest.cpp:154:   instantiated from here
[...]

PR #2166, opened 2017-12-20
Remove incompatible compiler features in LinkageTest

dchopra001:
Tril tests inside LinkageTest.cpp use function pointers to template functions. This is incompatible with gcc 4.4.7. This commit implements a workaroud so that we can successfully compile and run LinkageTest using gcc 4.4.7.

Closes: #2152

### DIFF
--- a/fvtest/compilertriltest/LinkageTest.cpp
+++ b/fvtest/compilertriltest/LinkageTest.cpp
@@ -58,8 +58,8 @@ TYPED_TEST_CASE(LinkageTest, InputTypes);
 
 TYPED_TEST(LinkageTest, InvalidLinkageTest) { 
    char inputTrees[200] = {0};
-   const auto format_string = "(method return=Int32  args=[Int32] (block (ireturn (icall address=%p args=[Int32] linkage=noexist  (iload parm=0)) )  ))";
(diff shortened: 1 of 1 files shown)


dchopra001:
Not sure if this is the best approach to solving the issue. 

@rbanerjee @Leonardo2718 @mgaudet thoughts? 

Let me know if you have a better approach. Otherwise I'll clean the current solution up and we can go from there.

Leonardo2718:
I believe the currently accepted "workaround" is to explicitly cast the function pointers, for example: `static_cast<TypeParam (*)(TypeParam)>(passThrough<TypeParam>)`.

(just `static_cast<TypeParam (*)(TypeParam)>(passThrough)` might also work).

dchopra001:
Ahh yes. Good point. I've made the changes. 

Going to remove the WIP tag as well.

0xdaryl:
genie-omr build all

0xdaryl:
This PR only affects the LinkageTest, so the zOS CI failure is not related to this PR.

dchopra001:
> Please change your commit message wording to indicate that you're just trying to get the LinkageTest to build and run with gcc 4.4.7.

Sorry about the delay on this. I've changed it now.

0xdaryl:
genie-omr build all

PR #2171, opened 2017-12-21
Remove references to libaux library

mstoodle:
The initial OMR contribution included a library called "auxv" as part of
the port library (port/unix/auxv.[ch]). This library was written by IBM
and incorporated into the OMR code base (then the J9 JVM) long before
this code was separately released as an open source library called
libauxv. The code in OMR is, in fact, unrelated to that open source
library.

A mistake was made referencing libauxv in our license materials, however:
our LICENSE and longabout.html files reference libauxv as if it has a
different license than the rest of OMR and even mention the open source
project. This commit removes these references because they are wrong.

[ci skip] since this update does not affect the code

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 Product: OMR
 
-Copyright (c) 1991, 2016 IBM Corp. and others
+Copyright (c) 1991, 2017 IBM Corp. and others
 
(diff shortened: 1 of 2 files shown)


0xdaryl:
I think a project lead (@charliegracie) should be the assignee on this one.

PR #2178, opened 2017-12-22
Delete unused code in virtual guard coalescer

vijaysun-omr:
This commit deletes some routines that were
prototyped a few years ago but are not used
currently. Newer features in the compiler such
as OSR and argument rematerialization in inliner
make it very unlikely this code will ever be used.


vijaysun-omr:
FYI @0xdaryl

0xdaryl:
genie-omr build all

Issue #2175, opened 2017-12-21
Do not double compute usedef use result


fjeremic:
There is a lot of code that looks like this:

```
_useDefInfo->getUseDef(info, load->getUseDefIndex());
if (!info.IsZero())
   {
   ...
   }
```

The `getUseDef` function is defined as follows:
https://github.com/eclipse/omr/blob/78177d1357a00cc175191775635f147d5b76c22b/compiler/optimizer/UseDefInfo.cpp#L2637-L2641

Note the return value of `getUseDef` is identical to what we check in the subsequent `if`. This is wasteful because the `isZero` function has a loop inside which means we are double computing:
https://github.com/eclipse/omr/blob/78177d1357a00cc175191775635f147d5b76c22b/compiler/cs2/bitvectr.h#L1062-L1072

We need to simplify all such code. There are dozens of such references.

0xdaryl:
I believe this is fixed now.

PR #2183, opened 2017-12-22
Avoid IsZero overhead after calling getUseDef

vijaysun-omr:
As mentioned in issue #2175 the result of getUseDef
has already been checked for zero (the return value
of getUseDef is false if zero). So there is no need
to check IsZero again and this is the change in this commit.

Issue: #2175


vijaysun-omr:
@0xdaryl I won't be looking to merge this without some more testing (probably only in the new year) but I would appreciate a review when you get a chance (no rush).

vijaysun-omr:
genie-omr build all

fjeremic:
@vijaysun-omr anything left to do in this PR before unmarking the WIP?

PR #2185, opened 2017-12-23
Add missing directory stem prefix to some #include directives

0xdaryl:

### DIFF
--- a/compiler/runtime/CodeCacheTypes.cpp
+++ b/compiler/runtime/CodeCacheTypes.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2017 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


0xdaryl:
genie-omr build all

0xdaryl:
The zOS failure is not related to these compiler changes.

0xdaryl:
genie-omr build all

compiler/runtime/CodeCacheTypes.cpp
```diff
@@ -19,7 +19,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
  *******************************************************************************/
 
-#include "CodeCacheTypes.hpp"
+#include "runtime/CodeCacheTypes.hpp"
```
fjeremic: We don't have to do this here as we are inconsistent elsewhere but what do you think about alphabetizing the include directives in the compiler subdirectories?

fjeremic:
The tags seem to imply this is a bug. Does it close a particular issue? And if not a minor comment in the commit on why this is a bug per-say should be warranted.

compiler/runtime/CodeCacheTypes.cpp
mstoodle: not an unreasonable goal to move towards, I think

Issue #2187, opened 2017-12-28
Load Extensions optimization should respect global sign extension decisions


fjeremic:
## Background

The Load Extensions optimization is a local optimization (operating on extended basic block boundaries) which attempts to identify in which context sub-64-bit loads are being used, signed or unsigned, based on the context of the load reference. While counting such loads the optimization also looks for conversion nodes which the respective loads feed into.

The opportunity here for sub-64-bit loads is that based on the preference of the uses of the respective load we can choose to sign or zero extend the loads at the source to a 64-bit width and in doing so we can skip some (or all) of the subsequent conversions because we can assume the load register has already been zero or sign extended. A canonical example could look like this:

```
n10n     NULLCHK on n12n
n11n       iloadi <someInt>
n12n         aload <this>

...

n13n     lstore <auto slot 1>
n14n       iu2l
n11n         ==>iloadi
```

A normal evaluation sequence would force us to generate a logical zero extension for node `n14n`, however this is unnecessary. If we were to instead load and zero extend when evaluating `n11n` we could avoid a future conversion at `n14n` thus saving an instruction from being generated.

The Load Extensions optimization would in this case mark the load at `n11n` with a `force64Bit` and `isUnsignedLoad` flags which the load evaluator will check for an emit a load (32-bit) and zero extend (to 64-bit) instruction. The optimization would also set an `unneededConv` flag on the conversion at `n14n` signaling the evaluators that this conversion can be skipped and no instructions need be emitted.

## Problem

The problem with this local optimization is that it is unaware of and does not respect global sign extension decisions. GRA in particular can set the `NeedsSignExt` flag on `iRegStores` signifying that a sign extension should be emitted during the evaluation of the global register store. This directly conflicts with the Load Extensions optimization. Consider the following trees:

```
n10n     NULLCHK on n12n
n11n       iloadi <someInt>
n12n         aload <this>

...
n15n     istore <auto slot 2>
n11n       ==>iloadi
...

n13n     lstore <auto slot 1>
n14n       iu2l
n11n         ==>iloadi
```
     
[...]

PR #2188, opened 2017-12-28
Exclude nodes marked with NeedsSignExt flag in Load Extensions

fjeremic:
Load Extensions is a local optimization is that it is unaware of and
does not respect global sign extension decisions. GRA in particular can
set the NeedsSignExt flag on iRegStores signifying that a sign
extension should be emitted during the evaluation of the global
register store. This directly conflicts with the Load Extensions
optimization.

We introduce an excluded nodes list and prevent the local optimization
on any load that feeds into a store marked with the NeedsSignExt flag.

Closes: #2187

### DIFF
--- a/compiler/optimizer/LoadExtensions.cpp
+++ b/compiler/optimizer/LoadExtensions.cpp
@@ -54,7 +54,7 @@
 
 TR_LoadExtensions::TR_LoadExtensions(TR::OptimizationManager *manager)
    : TR::Optimization(manager),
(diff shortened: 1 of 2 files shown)


fjeremic:
I kindly request a review from either @vijaysun-omr or @andrewcraik.

vijaysun-omr:
The change looks good to me.

vijaysun-omr:
genie-omr build all

vijaysun-omr:
Checks have passed. Merging this now.

PR #2190, opened 2017-12-30

amitin:
Changes I made:
Modified cmake scripts 
Removed unused variables
Fixed some type casts


babsingh:
@amitin what warning did Clang issue for using {{0}} initializers?

amitin:
@babsingh, if I restore for example this: 
`struct sigaction act = {{0}};`
I can see the following as an error message issued by clang: 
C:\workspace\omr\omrsigcompat\omrsig.cpp(257,26): error : braces around scalar initializer [-Werror,-Wbraced-scalar-init]

PR #2191, opened 2017-12-30

amitin:
Changes I made:
Modified cmake scripts
Removed unused variables
Fixed some type casts


omrsigcompat/omrsig.cpp
```diff
@@ -197,8 +197,12 @@ omrsig_handler(int sig, void *siginfo, void *uc)
 sighandler_t
 omrsig_primary_signal(int signum, sighandler_t handler)
 {
-	struct sigaction act = {{0}};
-	struct sigaction oldact = {{0}};
+	struct sigaction act;
+	struct sigaction oldact;
+
+	memset(&act, 0, sizeof(struct sigaction));
```
charliegracie: The zero initializer gives a warning on clang?

port/win32/omrfile_blockingasync.c
```diff
@@ -139,7 +139,7 @@ omrfile_blockingasync_read(struct OMRPortLibrary *portLibrary, intptr_t fd, void
 
 
 intptr_t
-omrfile_blockingasync_write(struct OMRPortLibrary *portLibrary, intptr_t fd, void *buf, intptr_t nbytes)
+omrfile_blockingasync_write(struct OMRPortLibrary *portLibrary, intptr_t fd, const void *buf, intptr_t nbytes)
```
charliegracie: Can you update all instances of `omrfile_blockingasync_write` to have the proper signature?  I see a few other hits in the codebase missing the `const`.  Thanks

port/win32/omrintrospect.c
```diff
@@ -135,8 +136,7 @@ suspend_all_preemptive(struct PlatformWalkData *data)
 		}
 
 		if (!Thread32First(data->snapshot, &data->thread32)) {
-			result = GetLastError();
-			goto error_return;
```
charliegracie: Can you provide a bit more context why you are changing the behaviour here?

port/win32/omrintrospect.c
```diff
@@ -176,13 +175,8 @@ suspend_all_preemptive(struct PlatformWalkData *data)
 		passA = !passA;
 	} while (a_count != b_count);
 
-	if (a_count != b_count) {
-		goto error_return;
```
charliegracie: Why are you removing this?

omrsigcompat/omrsig.cpp
amitin: Yes, it does.

port/win32/omrfile_blockingasync.c
amitin: Sure.

port/win32/omrintrospect.c
amitin: The code at 'error_return' label does SetLastError(result) and returns -1. So doing GetLastError() and then immediately invoking SetLastError() seems to be a noop. Why don't just return -1 having last error set by Thread32First() function?
See also the code above: CreateToolhelp32Snapshot() also sets last error if it fails, and the check against invalid handle just returns -1 - the last error is set.

port/win32/omrintrospect.c
amitin: This condition is always false. The execution flow only gets there if the loop above exits normally.
Additionally, if the condition somehow wasn't false, then there is a chance to get to 'error_return' label having 'result' variable uninitialised.

amitin:
Closing this. See #2198.

Issue #2192, opened 2018-01-02
DefiningMapAllocator in OSRData.hpp needs const key


mstoodle:
...or else recent XCode versions complain *bitterly* compiling on a Mac...

mstoodle:
First fix was not correct and broke builds on AIX and z/OS (thanks xlC). I think I just added an extra "const" that wasn't needed. Will test those platforms before contributing a fix.

PR #2193, opened 2018-01-02
Give DefiningMap a const key

mstoodle:
std::map uses require a const key or else recent version of XCode (more
recent than used on our build machines apparently) will not match the
template classes properly. This commit simply makes the key const.

Fixes: #2192

### DIFF
--- a/compiler/compile/OSRData.hpp
+++ b/compiler/compile/OSRData.hpp
@@ -91,9 +91,9 @@ namespace TR
       };
    }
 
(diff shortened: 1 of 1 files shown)


mstoodle:
genie-omr build all

vijaysun-omr:
Simple fix and checks have passed. Merging.

mstoodle:
Reverting because it breaks on AIX and zOS (using xlC). I will need to figure out a way to accommodate OSX as well as those platforms.

mstoodle:
ok, i suspect I just added one too many "const" to the fix. will test downstream before contributing an updated fix.

Issue #2049, opened 2017-11-23
C-API for JitBuilder


dinfuehr:
It would be helpful to have an additional C-API for JitBuilder. This would make it much easier to use OMR in languages other than C++. Many languages like Rust or Swift provide good interoperability with C, but can't consume C++-APIs. LLVM for example also provides a C-bindings for this reason.

mstoodle:
Thanks for making the suggestion, @dinfuehr. There has been some discussion in the past to do something with either clang plugins or something like Swig to create alternative language bindings for JitBuilder. I've also done a manual prototype of Java bindings for JitBuilder last year.

There's no technical reason it cannot be done, of course, since the C++ used in JitBuilder could largely be viewed as syntactic sugar on C. I think the more important challenge is how to do it in a way that can be fairly easily carried along as the JitBuilder API evolves forward (and that projects that extend the JitBuilder API can also leverage). Thinking required, suggestions welcome!

mstoodle:
fwiw @dinfuehr , I'll be covering this topic a bit in my JitBuilder status and directions update webcast tomorrow at 3:30pm EST (see https://github.com/eclipse/omr/issues/2339).

dinfuehr:
Thanks, might not be able to join but definitely gonna watch the recording..

mstoodle:
There's a question related to how the C API might look pasted into #2397 but I'll reproduce here as it's the better place to discuss it.

The basic question is around name JitBuilder services, since C doesn't have a namespace to hide them behind. 

I was imagining that every JitBuilder function will be prefixed by a "short name" for the JitBuilder class.

So `IlBuilder` functions are called `IB_Add`, `IB_Sub`, etc. and take an explicit `IlBuilder *` "receiver" parameter.

`BytecodeBuilder` functions are called `BB_Add`, `BB_Sub`, etc. and take an explicit `BytecodeBuilder *` "receiver" parameter.

`TypeDictionary` functions are called `TD_DefineStruct`, etc. and take an explicit `TypeDictionary *` "receiver" parameter.

Thoughts?

mstoodle:
BTW, the early stages (i.e. there is some very wrong code in here) of a C API generator implemented in a branch of my OMR fork here: https://github.com/mstoodle/omr/tree/jitbuilder_api). Look in, e.g., `jitbuilder/client/C_Binding.[ch]pp` .)

[...]

Issue #1819, opened 2017-10-18
PROPOSAL: Add semantic versioning to JitBuilder API


mstoodle:
In preparation to record JitBuilder IL (JBIL) under Issue #1818, I would like to start versioning the JitBuilder API so that the version can be written out as part of a JBIL record. With an embedded semantic version (see [http://semver.org/](http://semver.org/)) , any particular JitBuilder library should be able to self certify whether it is qualified to replay the JBIL created from any other (valid :) ) source. Why now, given the JitBuilder API is still missing major features? Because I believe we should be building the versioning notion into our infrastructure now to avoid having to retrofit later. I also think that our existing JitBuilder consumers would benefit from a better notion of how refreshing their JitBuilder library is expected to affect their implementation.

Note that this versioning specifically concerns the API, not the implementation (though, of course, some implementation changes will also implact the API). I have purposefully excluded the implementation because clients should program against the API not the implementation.

Semantic versioning will require a higher standard of code review associated with changes to the JitBuilder API, in that it will only work if committers consistently respect that changes in the API must be properly reflected in updates to the semantic versioning. _**Pull requests should not be merged if they include commits that impact the JitBuilder API without including the appropriate update to the semantic version.**_ Commits will need to be updated to include the version update.

Specifically, the proposal is to:
   1) create a base class for all the *Builder classes and for TypeDictionary called something like *BuilderAPI (name subject to feedback)
   2) in one particular class, a set of const integers will be introduced that describe the MAJOR, MINOR, and PATCH levels for the JitBuilder API. For hopefully adequate future proofing, I propose MAJOR as a 32-bit unsigned integer value, and MINOR/PATCH as 16-bit unsigned integer values.
   3) in general, as the JitBuilder API evolves:
         PATCH is increased when bug fixes change the API but in backwards compatible ways
         MINOR is increased when new function changes the API but in backwards compatible ways
         MAJOR is increased when the API changes but not in a backwards compatible way
    4) any JitBuilder API change that affects the MAJOR number may need to be deferred depending on project plans and release schedule (you know, once we have those)
  
I propose to start the versioning at 1.0.0 since the JitBuilder API is actually quite functional at this point. I expect MAJOR version bumps won't be all that unusual for the next little while since the JitBuilder API is still "coming together".

[...]

PR #2194, opened 2018-01-02
WIP Define JitBuilder client API

mstoodle:
Separate out the client API into its own set of classes distinct from implementation, with the goals to semantically version the client API (#1819), to eventually be able to cleanly separate different possible implementations of the client API (#1818), and to make it somewhat easier to create different language bindings into the JitBuilder library (e.g. #2049, https://github.com/eclipse/openj9/issues/806).

This is work in progress because all the changes are all scrunched together in mostly one massive commit and all the commit messages are minimal, but I now have everything rebased to master and working on my laptop so I wanted to check whether there are any platform issues remaining. Before removing the WIP, I'll tease the one massive commit into several more meaningful commits, though I am unlikely to make the interim commits fully build-able due to the pervasiveness of this change.

I will be changing history on this branch, so I don't recommend anyone to build upon it in its current form.

I'll also make this PR message more meaningful :) .

### DIFF
--- a/compiler/compile/OSRData.hpp
+++ b/compiler/compile/OSRData.hpp
@@ -91,9 +91,9 @@ namespace TR
       };
    }
 
-typedef TR::typed_allocator<std::pair<int32_t , TR_BitVector*>, TR::Region&> DefiningMapAllocator;
+typedef TR::typed_allocator<std::pair<const int32_t, TR_BitVector*>, TR::Region&> DefiningMapAllocator;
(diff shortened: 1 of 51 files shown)


mstoodle:
This PR requires more work; I realized I need to do more work to cleanly support fvtest compilers that are based on the IlInjector / JitBuilder classes.

Please ignore for the time being.

mstoodle:
In fact, I am rethinking some of the fundamental approach here now. I'll close this PR and open a new one once the world order is re-established.

PR #2196, opened 2018-01-03
Eliminate last bits of early jitbuilder replay support

mstoodle:
There is one last piece of code in IlBuilder.cpp and in
the JitBuilder makefile to finish removing the early replay
support.

Fixes: #1818


vijaysun-omr:
Trivial cleanup that is'nt significant enough for more cross platform testing in my opinion. Merging.

PR #2197, opened 2018-01-03
Revert "Give DefiningMap a const key"

mstoodle:
Reverts eclipse/omr#2193


mstoodle:
Revert, so testing is technically not useful. Merging to address downstream build breaks.

PR #2198, opened 2018-01-04

amitin:
Closing #2191 as it has several unnecessary merge commits.
Changes I made:
Modified cmake scripts
Removed unused variables
Fixed some type casts

### DIFF
--- a/port/win32/omrintrospect.c
+++ b/port/win32/omrintrospect.c
@@ -111,6 +111,7 @@ resume_all_preempted(struct PlatformWalkData *data)
  * of our threads barring the one executing the function. It then takes a new snapshot to determine
  * if new threads were created during the suspension. This is repeated until the thread count matches
  * between the two snapshots.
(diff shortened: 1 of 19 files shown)


port/win32/omrintrospect.c
```diff
@@ -176,13 +175,8 @@ suspend_all_preemptive(struct PlatformWalkData *data)
 		passA = !passA;
 	} while (a_count != b_count);
 
-	if (a_count != b_count) {
-		goto error_return;
```
charliegracie: I agree I do not see how the code would exist the loops above without the 2 counts being equal to one another without just returning.   I am going to open an issue for this function though.  If the thread count did change I do not see how the loop would ever exit since the counts are never reset.

charliegracie:
@dnakamura can you look at the CMake changes.  They look good to me.

charliegracie:
genie-omr build all

dnakamura:
lgtm

PR #2200, opened 2018-01-04
Enable compilation of the compiler technology on zOS

charliegracie:


charliegracie:
genie-omr build all

charliegracie:
The compiler technology makefiles in OMR currently do not support compiling on zOS.  I will try again when we have CMake support on zOS.

PR #2201, opened 2018-01-05
Give DefiningMapAllocator a const key type

mstoodle:
TR::typed_allocator requires a const key type or else recent version of XCode
(more recent than used on our build machines apparently) will not match the
template classes properly. This commit simply makes the key type const.

I tested on all platforms to verify builds, since an earlier version of
this fix broke builds on some platforms.

Fixes: #2192

### DIFF
--- a/compiler/compile/OSRData.hpp
+++ b/compiler/compile/OSRData.hpp
@@ -91,7 +91,7 @@ namespace TR
       };
    }
 
-typedef TR::typed_allocator<std::pair<int32_t , TR_BitVector*>, TR::Region&> DefiningMapAllocator;
+typedef TR::typed_allocator<std::pair<const int32_t, TR_BitVector*>, TR::Region&> DefiningMapAllocator;
 typedef std::less<int32_t> DefiningMapComparator;
 typedef std::map<int32_t, TR_BitVector*, DefiningMapComparator, DefiningMapAllocator> DefiningMap;
 


mstoodle:
genie-omr build all

compiler/compile/OSRData.hpp
```diff
@@ -91,7 +91,7 @@ namespace TR
       };
    }
 
-typedef TR::typed_allocator<std::pair<int32_t , TR_BitVector*>, TR::Region&> DefiningMapAllocator;
+typedef TR::typed_allocator<std::pair<const int32_t, TR_BitVector*>, TR::Region&> DefiningMapAllocator;
```
fjeremic: Should all `std::map` keys be defined as `const` as a good rule of thumb? I have a PR in flight with a similar change.

mstoodle: @fjeremic yes, otherwise I think you'll generate another example of the same problem this PR fixes :(

mstoodle: (thanks for noticing :) )

mstoodle: note that the std::map key type is still not const, though.... it's only on the allocator that you need the 'const'. If you do the key type on DefiningMap, for example, that will break on AIX and zOS (as I found with my earlier version of this fix). Unfortunately, we do not currently have compiler component testing on those platforms at Eclipse OMR (we're working on it!)

fjeremic: Makes sense given the definition of the allocator from the documentation (http://en.cppreference.com/w/cpp/container/map). Thanks! I'll make the same change accordingly.

mstoodle:
@vijaysun-omr I think this one is ready for your consideration, at your convenience.

Issue #2172, opened 2017-12-21
SPDX-License-Identifier text in file header needs to be refined


mstoodle:
In #1596 we updated our license according to the directions we were given from the Eclipse Foundation. There has bee, however, and update to the SPDX identifier we should be using to make our license more precise.

We've been asked to update this line in every copyright header to the following:

```
SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
```

Hopefully, this will be the last time we'll need to do this, as it will require a commit that touches basically every file in the repository :( .

fjeremic:
On a related note do we have a script that does this copyright change? If so we should check this into the repository so it is easy to do this next time.

pshipton:
It should be a simple search and replace, but due to the 72 column line limit restriction on z assembler files, these files need to be handled specially. Or the new identifier needs to be broken into multiple lines for all files. For OpenJ9 I fixed the z assembly files by hand since there aren't that many.

PR #2202, opened 2018-01-05
Correct SPDX license identifier

mstoodle:
The original SPDX license identifier was missing the refined details
about the GPL-2.0 special clauses. The new identifier includes these
details explicitly and more precisely depicts the Eclipse OMR license.

For every file in the repository, this commit changes this line in
the file header :
SPDX-License-Identifier: EPL-2.0 OR Apache-2.0

to :
SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception

Fixes: #2172


mstoodle:
this change *should* not affect the code, but given its breadth, let's test it anyway for the relatively cheap cost of a couple of builds :)

genie-omr build all

fjeremic:
Only think to be careful about is if we modified an Z assembler files the copyright change may push us over the 71 character limit which will cause z/OS failures. See https://github.com/eclipse/openj9/pull/873 for details.

util/omrutil/unix/zos/64/j9memclrz10_64.s
```diff
@@ -18,7 +18,7 @@
 * [1] https://www.gnu.org/software/classpath/license.html
 * [2] http://openjdk.java.net/legal/assembly-exception.html
 *
-* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
```
fjeremic: This will need to be broken up into two lines as on z/OS we hit the 71 character limit for the assembler. Similarly in all other z/OS assembler files and any .m4 files which will generate assembler files.

fjeremic: Also do we need copyright date updates for license changes?

mstoodle: blah, good points. Thanks @fjeremic !

mstoodle:
genie-omr build zos

mstoodle:
yay, I think I got them all! Thanks @fjeremic for pointing out the problem.

@charliegracie I think this one is ready for your consideration, at your earliest convenience :) .

charliegracie:
Does the SPDX line have to be on a single line?  All of the other lines in the copyright header wrap at a fairly consistent length.  Would it be possible to wrap this line as well?

mstoodle:
@charliegracie I actually wasn't sure wrapping it for zOS was even the right thing to do....now you're suggesting we wrap all of them? :) I'll look into whether there are guidelines for such things; stay tuned!

mstoodle:
@charliegracie 

From [SPDX 2.1, Appendix V](http://bit.ly/2aWiuRA)
"The tag should appear on its own line in the source file, generally as part of a comment."

We don't have any other restrictions on line length in source code files other than z/OS assembler files and that's for functional reasons. Although I understand the reason for the request and agree that it looks better wrapped, I think we should try to keep it on a single line where we can.

PR #2206, opened 2018-01-09
Check OSXSAVE when detecting AVX support

0dvictor:
OSXSAVE flag tells if OS enables XSAVE instruction,
which is a strong indication that AVX is enabled by
the OS.

### DIFF
--- a/compiler/x/runtime/X86Runtime.hpp
+++ b/compiler/x/runtime/X86Runtime.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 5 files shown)


0xdaryl:
These changes are only built on x86, so a genie build is not required.

compiler/x/runtime/X86Runtime.hpp
```diff
@@ -66,6 +73,15 @@ inline bool jitGetCPUID(TR_X86CPUIDBuffer* pBuffer)
       // EAX = 7, ECX = 0
       cpuidex(CPUInfo, 7, 0);
       pBuffer->_featureFlags8 = CPUInfo[EBX];
+
+      // Check for XSAVE
+      if(pBuffer->_featureFlags2 & 0x08000000) // OSXSAVE
+         {
+         if((6 & _xgetbv(0) != 6) || feGetEnv("TR_DisableAVX")) // Unset OSXSAVE if not enabled via CR0
```
0xdaryl: Can you annotate the naked constant `6` with a comment describing what XCR bits it represents?  e.g.,
```
// '6' = mask for XCR0[2:1]='11b' (XMM state and YMM state are enabled)
```

0dvictor: Good suggestion, updated.

0xdaryl:
Thanks for the change.  Not sure if you saw my copyright comment above?

0dvictor:
Sorry I missed your comments about copyright headers. Now they are all updated.

PR #2214, opened 2018-01-12
Preparations for use of ddrgen

keithc-ca:
Preparations for use of ddrgen in support of openj9#378

* correct references to getmacros script
* add missing ddr_namespace annotations
* use J9HASH_TABLE_AVL_TREE_TAG_BIT instead of hard-coded constant
* ignore objs directory

### DIFF
--- a/ddr_artifacts.mk
+++ b/ddr_artifacts.mk
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2016, 2017 IBM Corp. and others
+# Copyright (c) 2016, 2018 IBM Corp. and others
 # 
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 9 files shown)


fjeremic:
Cross linking the OpenJ9 issue https://github.com/eclipse/openj9/issues/378.

ddr_artifacts.mk
```diff
@@ -30,10 +30,10 @@ top_srcdir = .
 include $(top_srcdir)/omrmakefiles/configure.mk
 
 # Ensure the source tree path is canonically-formed and ends with /
-abs_srcroot = $(abspath $(TOP_SRCDIR))/
+abs_srcroot := $(subst //,/,$(abspath $(TOP_SRCDIR))/)
 
 all: check-vars
-	./tools/ddrgen/src/macros/getMacros.sh $(abs_srcroot)
+	bash $(top_srcdir)/ddr/tools/getmacros $(abs_srcroot)
```
charliegracie: should we be relying on `bash`?  Can we not just use `sh`?  I know there are sometimes issues with `bash` on AIX and zOS.

keithc-ca: getmacros begins with
```sh
#!/bin/bash
```
and I think it does depend on features in `bash` that are not in `sh`.

charliegracie: makes sense then to use `bash`

PR #2215, opened 2018-01-12
Enable OMRChecker testing script to read CLANG environment variable

nbhuiyan:
Currently, the test script always uses "clang++" to call Clang for
testing OMRChecker. This causes problems, when "clang++" calls an
unsupported clang version by default. There are a number of problems
that can arise from this, such as when using different Linux distros.
The increased flexibility was achieved by 1 change:
  - The constructor of class OMRChecker has been modified to pull and use
    Clang calling info from the environment, by looking for the env
    variable 'CLANG'. If no such env variable exists, then the default
    'clang++' is used.

This change is a requirement for addressing [issue 729](https://github.com/eclipse/openj9/issues/729) in eclipse/openj9.

### DIFF
--- a/tools/compiler/OMRChecker/test.py
+++ b/tools/compiler/OMRChecker/test.py
@@ -1,7 +1,7 @@
 #! /usr/bin/python
 
 ###############################################################################
-# Copyright (c) 2016, 2016 IBM Corp. and others
+# Copyright (c) 2016, 2018 IBM Corp. and others
 #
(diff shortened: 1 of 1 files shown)


nbhuiyan:
I have removed the WIP label after test running it locally and on Travis CI. Please merge this PR so that I can get started with eclipse/openj9#729

0xdaryl:
genie-omr build all

PR #2216, opened 2018-01-12
Add Rem service to IlBuilder

aviansie-ben:
Currently, there is no way to emit IL for getting the remainder of an
integer division using IlBuilder's services. This commit adds a new Rem
service on IlBuilder to allow this using existing support in
TR::ILOpCode for generating remainder opcodes.

Signed-off-by: Benjamin Thomas <ben@benthomas.ca>

### DIFF
--- a/compiler/ilgen/IlBuilder.cpp
+++ b/compiler/ilgen/IlBuilder.cpp
@@ -1033,6 +1033,15 @@ IlBuilder::doVectorConversions(TR::Node **leftPtr, TR::Node **rightPtr)
       *leftPtr = TR::Node::create(TR::vsplats, 1, left);
    }
 
+TR::IlValue *
+IlBuilder::widenIntegerTo32Bits(TR::IlValue *v)
+   {
+   if (v->getDataType() == TR::Int8 || v->getDataType() == TR::Int16)
+      return ConvertTo(Int32, v);
+
+   return v;
+   }
+
 TR::Node*
 IlBuilder::binaryOpNodeFromNodes(TR::ILOpCodes op,
                                  TR::Node *leftNode,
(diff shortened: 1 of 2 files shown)


compiler/ilgen/IlBuilder.cpp
```diff
@@ -1363,6 +1363,14 @@ IlBuilder::Div(TR::IlValue *left, TR::IlValue *right)
    return returnValue;
    }
 
+TR::IlValue *
+IlBuilder::Rem(TR::IlValue *left, TR::IlValue *right)
+   {
+   TR::IlValue *returnValue=binaryOpFromOpMap(TR::ILOpCode::remainderOpCode, left, right);
```
mstoodle: This looks reasonable (it's consistent with implementations of the other arithmetic services).

But none of our code generators currently support brem and srem, so anyone who uses those types will hit an assert deeper into a compilation than probably desirable. So, maybe we could do an automatic widening here if either of the parameters are Int8 or Int16 (similar but not quite the same as the widening that's done as part of Call for parameters, for example)? 

```
if (left->getDataType() == TR::Int8 || left->getDataType() == TR::Int16)
   left = ConvertTo(Int32, left);
```

...and same for right. Or, even better, you could create a private service `TR::IlValue *widenIntegerTo32Bits(TR::IlValue *value)` to convert both left and right.

For consistency (and in case we revert this widening operation later), I suggest converting the return value to the type of the (arbitrary, but consistent with how other operations choose their type) left operand.

Please also put a comment explaining why the widening is done here, so there's a reminder to people to undo this stuff if/when someone gets around to implementing the opcodes natively.

aviansie-ben: Should be fixed now

mstoodle:
Changes look good to me. We should really add a test for these, but I would not be surprised if we're missing examples of some of the other arithmetic operators as well.

I have created #2223 to capture the requirement to create tests/samples for all the jitbuilder services and specifically mentioned Rem() there.

mstoodle:
Since Rem doesn't have a test, it's unlikely additional platform testing will provide any benefit (and it's written in the same style as the rest of the services so unlikely to show build problems either <ahem> ).

PR #2217, opened 2018-01-13
Re-introduce the lurem opcode

### LINKED ISSUE #2212: Revive lurem opcode
#269 removed the `lurem` opcode. However, this opcode represents a distinct and general operation that may be useful in some language runtimes (in fact, its 32-bit counterpart, `iurem`, is still present). For generality, this opcode should be re-introduced, either by reverting #269 or by re-introducing a subset of the functionality removed in that PR. In addition, Tril tests should be written for this opcode to ensure it functions as expected, with a reasonable level of certainty.

Followup tasks would be to implement the corresponding 8-bit and 16-bit variants of this opcode, `burem` and `surem`.

Leonardo2718:
This changeset re-introduces the `lurem` opcode by reverting #269 and applying necessary patches to make the code up-to-date with more recent changes. Some tests for this and related opcodes are also added.

Fixes #2212

### DIFF
--- a/compiler/arm/codegen/TreeEvaluatorTable.hpp
+++ b/compiler/arm/codegen/TreeEvaluatorTable.hpp
@@ -119,6 +119,7 @@
    TR::TreeEvaluator::unImpOpEvaluator,     // TR::brem
    TR::TreeEvaluator::unImpOpEvaluator,     // TR::srem
    TR::TreeEvaluator::iremEvaluator,        // TR::iurem
+   TR::TreeEvaluator::lremEvaluator,        // TR::lurem
    TR::TreeEvaluator::inegEvaluator,        // TR::ineg
    TR::TreeEvaluator::lnegEvaluator,        // TR::lneg
 #if (defined(__VFP_FP__) && !defined(__SOFTFP__))
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -6866,7 +6866,7 @@ TR::Node *removeRedundantREM(OMR::ValuePropagation *vp, TR::Node *node, TR::VPCo
          //    child prec=3 (i.e. max value is 999)
          //    iconst 1000
          //
-         // this doesn't work for unsigned (iurem) as these ops treat the operands as unsigned so, for example, a prec=1 dividend that is negative
+         // this doesn't work for unsigned (iurem/lurem) as these ops treat the operands as unsigned so, for example, a prec=1 dividend that is negative
          // when interpreted as unsigned will actually have 10 digits of precision (-1 would be 0xFFffFFff = 4,294,967,295)
          //
          return vp->replaceNode(node, node->getFirstChild(), vp->_curTree);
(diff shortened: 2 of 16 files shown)


Leonardo2718:
The build failure is due to a time-out error in `ThreadAbortTest` and is unrelated to this particular PR.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -7004,6 +7004,62 @@ TR::Node *constrainLrem(OMR::ValuePropagation *vp, TR::Node *node)
...
+   TR::VPConstraint *rhs = vp->getConstraint(node->getSecondChild(), rhsGlobal);
+   TR::VPConstraint *constraint = NULL;
+   lhsGlobal &= rhsGlobal;
+   if (lhs && lhs->asLongConst() &&
+       rhs && rhs->asLongConst())
+      {
+      //
+      // REMOVED AS IT WOULD LEAD TO AN ASSERT POST FORK
+      //
+      TR_ASSERT(0, "constrainLurem not supported");
```
0xdaryl: This comment and assert needs to be made more relevant in the open-source world.  It's really hiding a limitation of the implementation, namely that we can't fold constant unsigned dividend and divisors for long remainder at compile-time (I think that just needs to be implemented).  I would like to either see the complete implementation here, or have this code fall-back on the most general constraint with an issue created to add unsigned long remainder support to the ArithEnv class.

compiler/optimizer/ValuePropagationTable.hpp
```diff
@@ -250,6 +251,7 @@ const ValuePropagationPtr constraintHandlers[] =
    constrainChildren,        // TR::brem
    constrainChildren,        // TR::srem
    constrainIrem,            // TR::iurem
+   constrainLurem,            // TR::lurem
```
0xdaryl: Please remove leading space for consistency.

compiler/p/codegen/BinaryEvaluator.cpp
```diff
@@ -2372,6 +2372,8 @@ TR::Register *lrem64Evaluator(TR::Node *node, TR::CodeGenerator *cg)
    int64_t divisor = 0;
    TR::Compilation * comp = cg->comp();
 
+   TR_ASSERT(node->getOpCodeValue() != TR::lurem, "TR::ludiv is not impelemented yet for 64-bit target\n");
```
0xdaryl: spelling:   implemented

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -116,3 +118,170 @@ INSTANTIATE_TEST_CASE_P(DivArithmeticTest, Int32Arithmetic, ::testing::Combine(
...
+
+    auto entry_point = compiler.getEntryPoint<uint64_t (*)(uint64_t, uint64_t)>();
+    ASSERT_EQ(param.oracle(param.lhs, param.rhs), entry_point(param.lhs, param.rhs));
+}
+
+INSTANTIATE_TEST_CASE_P(DivArithmeticTest, UInt64Arithmetic, ::testing::Combine(
+    ::testing::ValuesIn(
+        TRTest::filter(TRTest::const_value_pairs<uint64_t, uint64_t>(), div_filter<uint64_t> )),
+    ::testing::Values(
+        std::make_tuple("lurem", lurem) )));
```
0xdaryl: Won't the creation of these constant parameter tests for lurem cause the assert in VP that I commented on above?  Presumably the tests need to run at a high enough opt level, but theoretically possible nonetheless.

compiler/arm/codegen/TreeEvaluatorTable.hpp
```diff
@@ -119,6 +119,7 @@
    TR::TreeEvaluator::unImpOpEvaluator,     // TR::brem
    TR::TreeEvaluator::unImpOpEvaluator,     // TR::srem
    TR::TreeEvaluator::iremEvaluator,        // TR::iurem
+   TR::TreeEvaluator::lremEvaluator,        // TR::lurem
```
0xdaryl: I think this should be left as `TR::TreeEvaluator::unImpOpEvaluator`.  I don't believe the ARM long division/remainder helpers can handle unsigned operations.

I will be creating an issue shortly to complete the implementation of unsigned arithmetic on all code generators and I will include this particular case as part of that.

compiler/p/codegen/BinaryEvaluator.cpp
Leonardo2718: For once, this is not my spelling error 😆 ... but I'll fix it.

fvtest/compilertriltest/ArithmeticTest.cpp
Leonardo2718: I think you're right that it's theoretically possible. Currently, tests are only compiled at warm, so that might explain why I haven't seen the failure when I run the tests locally.

0xdaryl:
Hi @Leonardo2718...  What is the current status of this PR? Would you consider closing it while you work on addressing the review comments and then re-open when ready?

0xdaryl:
@Leonardo2718 : same question as above.  :-)

Leonardo2718:
Heh... yeah ok :sweat_smile: . I'll close this until I have more time to work on this.

Issue #2168, opened 2017-12-20
Should TR::checkILCondition trap or throw an exception upon failure?


dchopra001:
While investigating some tril failures I found that an IlValidator DeathTest ( IlValidatorTest/IllformedTrees) spits out 14 core files (1 for each test). This is expected behavior because `TR::checkILCondition` is designed to trap if it detects a validation error:
https://github.com/eclipse/omr/blob/38f19f5a9254a3a14b1258c8c0dee634319a8198/compiler/ras/ILValidationUtils.cpp#L45-L64

My question is whether we want this behavior, or should we instead throw an exception and then change the GoogleTest from Death tests. If we want to keep this behavior, then we need to discuss what to do to make sure the cores are cleaned up as it seems wasteful to generate them right now. 

@rbanerjee @Leonardo2718 @mgaudet thoughts?

fjeremic:
IMO we do not want this behavior. Running this Tril test on a system with a `ulimit -c unlimited` set will result in tons of cores generated and tons of wasted space. I would think if we encountered invalid IL instead of trapping we should abort the compile via a C++ exception.

In debug mode uncaught aborted compiles should be fatal while in production mode they are silent failures only observed in verbose log as compilation failures.

IMO this death test should be changed to catch a compilation abort exception due to invalid IL rather than expect a crash.

@mpirvu @dsouzai thoughts?

mgaudet:
So, yes, definitely. Though, I do wonder if this may be a bit of a google test bug. 

Charlie has investigated issues with death tests on OS/X and switching to exceptions would be the resolution there too.

dchopra001:
Side Note: CommoningDeathTest will also need to be changed if we decide to go the exception route.

Leonardo2718:
Interesting. In the original implementation, I used a macro instead of calling `TR::trap()` directly. The macro mapped to `TR::trap()` only in DEBUG builds. In PROD builds, it threw an exception to abort compilation without taking down the VM.

I tend to agree that throwing an exception is preferable. Ideally, this exception should be able to propagate to the VM or test harness for testing purposes. In a production build, it would just abort compilation. This would make it easy to test that invalid IL is rejected. The death tests would be changed to expect a specific exception to be thrown.

However, it is also possible to check the return code of the compile method call. This is not ideal but will work for most testing purposes.

Maybe @dsouzai or @mpirvu have ideas for making the exception mechanism work?

rbanerjee:
[...]

PR #2222, opened 2018-01-15
Remove Death Tests in ILValidatorTest

dchopra001:
The IL Validator traps when it detects invalid IL. This also results in a core being produced. This can be wasteful when several tril tests are purposefully testing the IL Validator by feeding in incorrect IL. Given the above reasons it would be better if the IL Validator threw an exception as opposed to trapping. With this change, the Death Tests in the ILValidatorTest's tril tests don't make sense anymore and need to be changed to instead detect the correct return code instead. This commit adds that functionality.

Closes: #2168


dchopra001:
Review from at least @rbanerjee and @dsouzai please.

compiler/ras/ILValidationUtils.cpp
```diff
@@ -58,7 +58,8 @@ void TR::checkILCondition(TR::Node *node, bool condition,
       printILDiagnostic(comp, "\n");
       if (!comp->getOption(TR_ContinueAfterILValidationError))
          {
-         TR::trap();
+         comp->failCompilation<TR::ILValidationFailure>("Failed as expected");
+         //TR::trap();
```
Leonardo2718: That comment should be removed.

dchopra001: Done.

dchopra001:
Had to make another change to rename CommoningTest to InvalidCommoningTest because there's already a class named CommoningTest defined.

charliegracie:

dchopra001:
Builds passed. This should be good to merge now.

PR #2224, opened 2018-01-16
Replace trfprintf calls with traceMsg

dchopra001:
trfprintf calls inside of doInstructionSelection do not do a null check on the file handle before writing. This can cause crashes when the file handle is null (which is a possibility when we run out of file handles (ulimit -n)). These calls should be replaced with traceMsg, which does a null check before trying to print.


dchopra001:
@dsouzai review request

charliegracie:
@genie-omr build zlinux zos plinux

dchopra001:
The travis CI test failed, but I don't see any output in the Job Log 😕

charliegracie:
It was cancelled.  Travis is cancelling old builds because they have such a backlog.  I tried restarting and we will see if that does anything.

0xdaryl:
I believe this is safe to merge without a macOS build.  Merging.

PR #2230, opened 2018-01-16
Fix std::map::value_type to match the type of the allocator

fjeremic:
Some compilers require (via static_assert) that the type of the value
match the type of the allocator value. In this case it does not matter
whether either are const so we make the map value_type as such.

### DIFF
--- a/compiler/optimizer/LoadExtensions.hpp
+++ b/compiler/optimizer/LoadExtensions.hpp
@@ -225,9 +225,9 @@ class TR_LoadExtensions : public TR::Optimization
 
    private:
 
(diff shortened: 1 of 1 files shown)


fjeremic:
@charliegracie / @mstoodle could you please review? This is supposed to fix OS X builds on a version of the compiler that does not run on the build farm. Not sure how to test this really.

charliegracie:

fjeremic:
Referencing a relevant discussion we had in https://github.com/eclipse/omr/pull/2201/files#r159917286. Unfortunately I missed the second `const`. All in all the typed allocator has to be a `const` pointer to a `const` type.

charliegracie:
sigh the travis builds have not done anything after running for 2+ hours

charliegracie:
I have verified that this in fact fixes OS X compiles on my machine

PR #2231, opened 2018-01-16
Remove deprecated x86 assembly files

0dvictor:
These files are long deprecated and hence should be deleted.


compiler/x/runtime/VirtualGuardRuntime.cpp
```diff
@@ -25,8 +25,6 @@
 
 #define IS_32BIT_SIGNED(x)   ((x) == ( int32_t)(x))
 
-extern "C" void _patchingFence16(void *startAddr);
-
```
0xdaryl: Please update the copyright header in this file.

0xdaryl:
A genie build is not required as the files affected are x86-only.

compiler/x/runtime/VirtualGuardRuntime.cpp
0dvictor: Thank you for pointing it out. Updated.

PR #2233, opened 2018-01-16
An additonal tweak for Clang/Win

amitin:
Fix intializer which clang 5 doesn't like.


charliegracie:
@genie-omr build zlinux plinux

Issue #2232, opened 2018-01-16
64 bit Power processor recognition on 64 bit builds


IBMJimmyk:
Under certain conditions (described below) it is possible for a 64 bit build to think it is running on a 32 bit Power processor. This can cause problems in other parts of the code that assume that a 64 bit build can not run on a 32 bit processor.

The condition is triggered when a 64 bit build is run on a unrecognized Power processor. Unrecognized Power processors are given the internal value of "TR_DefaultPPCProcessor". During the runtime check to see if the processor is 64 bit or not, "TR_DefaultPPCProcessor" is considered 32 bit.

The proposed fix is for 64 bit builds to assume they are running on a 64 bit processor even if it does not recognize the processor beyond it being a Power processor. This is safe because a 64 bit build can not be run on a 32 bit processor to begin with.

Behavior on 32 bit builds will remain unchanged. It will continue to assume that an unrecognized Power processor is 32 bit.

0xdaryl:
PR is merged.

PR #2234, opened 2018-01-16
64 bit builds should assume a 64 bit processor on Power

IBMJimmyk:
This is a change targeting Power. The runtime check is64BitProcessor now
checks to see if it is a 64 bit build. If so, it just returns true since
a 64 bit build must be running on a 64 bit processor. If not, the function
will check the CPU properties as before.

Issue: #2232

### DIFF
--- a/compiler/p/codegen/OMRCodeGenerator.cpp
+++ b/compiler/p/codegen/OMRCodeGenerator.cpp
@@ -3926,7 +3926,18 @@ bool OMR::Power::CodeGenerator::supportsTransientPrefetch()
 
 bool OMR::Power::CodeGenerator::is64BitProcessor()
    {
-   return TR::Compiler->target.cpu.getPPCis64bit();
+   /*
+    * If the target is 64 bit, the CPU must also be 64 bit (even if we don't specifically know which Power CPU it is) so this can just return true.
+    * If the target is not 64 bit, we need to check the CPU properties to try and find out if the CPU is 64 bit or not.
+    */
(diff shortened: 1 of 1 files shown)


0xdaryl:
Right now the 64-bit CPU capability is determined via an identity check on the processor.  Is that the *only* way of determining this for Power?  Perhaps the better fix is to ask the processor itself if it supports 64-bit, irrespective of its identity (I'm not sure of the means on Power, but on x86 you'd generate capabilities via CPUID).

ymanton:
The equivalent of `cpuid` doesn't exist in the PPC ISA, processor identity is the only thing that can be detected at the ISA level (by reading the Processor Version Register).

If you want to get at processor capabilities you'd have to ask the OS via whatever facility it provides; it would use the PVR and static data based on processor identity as well, but it saves you from having to maintain those things yourself. On Linux that would be things like the aux vector, `/proc/cpuinfo`, sysfs, `lscpu`, etc. In this case the aux vector's `AT_HWCAP` would have something like: `archpmu vsx arch_2_06 dfp ic_snoop smt mmu fpu altivec ppc64 ppc32`.

There's code in the port library to get at the aux vector, unfortunately it looks like its in OpenJ9 and not OMR for whatever reason, but for 64-bit detection on a 64-bit build I don't think we really need to bother with a runtime check.

0xdaryl:
genie-omr build plinux

Power testing is sufficient because it is a Power-only change.  Also, this can be merged without a macOS build.

PR #2236, opened 2018-01-16
Resolve compiler warnings on OS X

charliegracie:
The changes fix a lot of the compiler warnings on OS X.

### DIFF
--- a/compiler/x/amd64/codegen/OMRMemoryReference.cpp
+++ b/compiler/x/amd64/codegen/OMRMemoryReference.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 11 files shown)


charliegracie:
@0dvictor can you look at the change in `compiler/x/runtime/X86Runtime.hpp`

0dvictor:
```compiler/x/runtime/X86Runtime.hpp``` looks good to me.

File: compiler/x/amd64/codegen/OMRMemoryReference.cpp [hunk 1]
```diff
@@ -392,7 +392,7 @@ OMR::X86::AMD64::MemoryReference::addMetaDataForCodeAddressWithLoad(
             {
             if (cg->needClassAndMethodPointerRelocations())
                cg->addAOTRelocation(new (cg->trHeapMemory()) TR::ExternalRelocation(displacementLocation, (uint8_t *)&_symbolReference,
-                                                                                        (uint8_t *)containingInstruction->getNode()->getInlinedSiteIndex(),
+                                                                                        (uint8_t *)(uintptr_t)containingInstruction->getNode()->getInlinedSiteIndex(),
```
fjeremic: We should try to prefer `static_cast` over C-style casts for additional compiler provided safety. It catches obvious bugs and makes refactoring easy. For example changing the type `containingInstruction` returns to something invalid will result in a compile time bug rather than runtime crash.

Also `getInlinedSiteIndex` seems to return an `int16_t`. Why are these casts necessary?

https://github.com/eclipse/omr/blob/3cda7d85e70ba406301453269cd578f76f68af1e/compiler/il/OMRNode.hpp#L700

charliegracie: You get this warning without it `warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]`.  You first have to cast it to an integer of the correct size and then you can cast it to a pointer

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build zos

compiler/x/amd64/codegen/OMRMemoryReference.cpp
fjeremic: We need to clean this up. I just looked at uses of `TR::ExternalRelocation::_targetAddress2` (bad name IMO) and it seems it is being used in all sorts of contexts. If the respective field is to be used as an arbitrary cast it should be of type `void*` to make this explicit.

To avoid all the casting at source we should provide `int16_t` constructors and cast to `void*` underneath the hood if need be.

fjeremic: https://github.com/eclipse/omr/issues/2237

charliegracie:

charliegracie:
@genie-omr build zos

0xdaryl:
@charliegracie , can you confirm you can build this locally on macOS?

PR #2238, opened 2018-01-17
Co-operative copy of large objects in CS

amicic:
For large objects, before trying to win forwarding, setup destination
header object to indicate that object is in progress. Then threads that
see the object is forwarded, but still being copied, will wait instead
of trying to create its own copy.

Further, while waiting, those threads will try to participate in
copying. Destination object header, beside the being-copied bit will
contantin progress info: amount of bytes yet to copy and count of
threads already participating. Threads atomically try to win a section
of object to copy. Once, everything is taken to copy (0 bytes yet to
copy), and all threads are done with the last section and decrement the
participation count, the winning threads will do the final steps of
fixup and replace the progress info with the valid object header.

### DIFF
--- a/gc/structs/ForwardedHeader.cpp
+++ b/gc/structs/ForwardedHeader.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2016 IBM Corp. and others
+ * Copyright (c) 2015, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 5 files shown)


amicic:
Please, do not merge before https://github.com/eclipse/openj9/pull/960 is approved. They should go together (for the sake of OpenJ9).

gc/structs/ForwardedHeader.cpp
```diff
@@ -185,4 +229,115 @@ MM_ForwardedHeader::restoreSelfForwardedPointer()
...
+	}
+
+	return 0;
+}
+
+void
+MM_ForwardedHeader::wait(uintptr_t *spinCount) {
+	/* TODO: consider yielding and/or sleeping, for high spinCount value */
+	for (uintptr_t i = *spinCount; i > 0; i--)	{
+		MM_AtomicOperations::nop();
```
charliegracie: This doesn't get optimized away as an empty loop?

amicic: The very same spin mechanism is used by omrgc_spinlock_acquire.

charliegracie: Ok thanks.

PR #2239, opened 2018-01-18
Generate sxtb/sxth and uxtb/uxth instructions for ARM

knn-k:
This commit adds code generation of ARM's sxtb/sxth instructions
and uxtb/uxth instructions for improving the instruction sequences
of sign extension and of zero extension, respectively.

### DIFF
--- a/compiler/arm/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/arm/codegen/OMRTreeEvaluator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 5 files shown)


compiler/arm/codegen/OMRTreeEvaluator.cpp
```diff
@@ -1454,20 +1454,64 @@ TR::Register *OMR::ARM::TreeEvaluator::conversionAnalyser(TR::Node          *nod
 
 static void signExtend(TR::Node *node, TR::Register *dst, TR::Register *src, int32_t bitsInDst, TR::CodeGenerator *cg)
    {
```
0xdaryl: `signExtend` and `zeroExtend` are largely the same function, save for three opcode differences (unless I'm missing something subtle).  In the interests of code size (always a concern on ARM), would it be better to have a single `generateSignOrZeroExtend` function that is parameterized by those opcodes and sharing the bulk of the logic?  The other code generators do similar things in other contexts.

knn-k: Updated the code following the review comment.

0xdaryl:
A genie build is not required as this is an ARM-specific change.  Also, this can be merged without waiting for the macOS build to finish.

PR #2241, opened 2018-01-18
Disable vmThread reclamation on 32-bit Windows

0xdaryl:
In preparation for removing the vmThread reclamation logic, it must first
be disabled on the 32-bit platforms where it is enabled to aid with its
surgical removal:

* force `allowVMThreadRematerialization` to always return `false`.  While
this affects both Windows and Linux 32-bit, in practice this will have no
effect on Linux 32-bit because querying this setting is always coupled with
one of the other reclamation options below that mask its effect.  This
function always returns `false` on 64-bit platforms.
* force `supportsFS0VMThreadRematerialization` to always return `false`.
This affects Win32 only.
* remove calls to `setSupportsVMThreadGRA`, thereby causing queries to
`getSupportsVMThreadGRA` to always return `false`.  This affects Win32
only.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


0xdaryl:
As external Eclipse OMR testing doesn't occur on 32-bit Windows and Linux, I ran an OpenJ9 build and tests against this PR to ensure there were no problems.

@0dvictor or @andrewcraik : I'd appreciate a review please.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -976,19 +976,12 @@ bool OMR::X86::CodeGenerator::supportsAddressRematerialization()         { stati
```
0dvictor: Will these two functions `allowVMThreadRematerialization()` and `supportsFS0VMThreadRematerialization()` be cleaned up once downstream projects completely stop querying them?

vijaysun-omr:
Simple preparation for VM thread logic removal. Merging.

PR #2242, opened 2018-01-18
Flush RS prior STW Scavenge just for Mutators

amicic:
Flushing Scavenger Remembered Set fragment within thread GC Environments
for all threads (including GC once), prior the second/last STW phase
within Concurrent Scavenger cycle, while concurrent phase is still in
progress is not correct - it interferes with GC background threads that
are still using this RS fragments.

Only RS fragments for mutators (that generate remembered objects through
Write Barrier) should be flushed.

Also doing minor re-org (moving RS aware code into Environment, rather
in VMInternface), and simplifying (removing redundant Enviroment
parameters).


charliegracie:

Issue #106, opened 2016-05-10
Issues in omr/port/unix/omrintrospect.c


keithc-ca:
* The body of close_wrapper() doesn't match the comments describing its intended behavior.
* The uses of setup_native_thread() don't use its return value properly.
* omrintrospect_threads_nextDo() sends an excessive number of signals while waiting for the next thread to signal that it has captured its call stack. The comments refer to a Linux scheduler bug that has long-since been fixed: there is no longer a need for this (badly behaving) workaround.

charliegracie:
I believe this issue was resolved via #109

keithc-ca:
The previous workaround is still necessary for Linux kernel releases 2.x, but not for the reasons described by the original comment in the code. Instead, signals are either being lost or some threads are chosen to receive more than one. Either way, not all other threads have reached the signal handler before the controller starts waiting for client threads to trace their stacks. Sending additional signals (and yielding) while the controller is waiting for a client thread to trace its stack deals with both kinds of problems.

PR #2248, opened 2018-01-22
Refine minimum Linux kernel version with reliable sigqueue

keithc-ca:
Processes have been reported to hang in omrintrospect_threads_nextDo() on Linux systems with kernel version 3.10. This changes the assumption of reliability of sigqueue() from `(version >= 3)` to `(version >= 3.11)`.

See #106.


charliegracie:
@genie-omr build plinux zlinux

PR #2249, opened 2018-01-23
Remove unused debug functions for ARM

knn-k:
This commit removes some debug functions for ARM that are no longer
used: printAsmDumpHeader() and printARM().

### DIFF
--- a/compiler/ras/Debug.hpp
+++ b/compiler/ras/Debug.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -426,9 +426,6 @@ public:
    virtual void         printHeader();
    virtual void         printMethodHotness();
    virtual void         printInstrDumpHeader(const char * title);
-#if defined(TR_TARGET_ARM)
-   virtual void         printAsmDumpHeader();
-#endif
 
    virtual void         printByteCodeAnnotations();
    virtual void         printAnnotationInfoEntry(J9AnnotationInfo *,J9AnnotationInfoEntry *,int32_t);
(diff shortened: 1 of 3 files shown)


compiler/ras/Debug.hpp
```diff
@@ -426,9 +426,6 @@ class TR_Debug
    virtual void         printHeader();
    virtual void         printMethodHotness();
    virtual void         printInstrDumpHeader(const char * title);
-#if defined(TR_TARGET_ARM)
-   virtual void         printAsmDumpHeader();
-#endif
```
0xdaryl: Please update the copyright header in this file.

0xdaryl:
re/ https://github.com/eclipse/omr/pull/2249#issuecomment-359657253

My cursory inspection of the code suggests that it isn't used.

This debug code is notoriously ugly and rife with platform `#ifdefs` which makes determining answers to questions like this difficult to answer.  Tools like ctags may guide you in this effort, but I suspect the easiest path is to remove the function and doing a personal build on all architectures to see if it breaks.  Unfortunately, we haven't settled on an API for OMR yet which means when you delete functions in OMR (even seemingly unused functions) it may cause downstream problems in projects we are unaware of (OpenJ9 is easy to check).  We can't control that yet, which is one of the reasons we want to move to releases for OMR.

Another option is to create a PR (or even in this PR) with that function removed and a committer can launch a genie build on all architectures (except ARM, unfortunately) to verify there are no issues prior to merging.

compiler/ras/Debug.hpp
knn-k: Updated the copyright year of Debug.hpp.

knn-k:
I will leave the empty function ``TR_Debug::print(TR::FILE*, uint8_t*, uint8_t*)`` for the time being.
My purpose was to remove the calls to ``printARM()`` and ``disassemble()``.  Then I can reduce the binary size by removing ARMDisassem.cpp from compiler/trj9/build/files/target/arm.mk of OpenJ9.

0xdaryl:
@genie-omr build all

0xdaryl:
This PR has been awaiting macOS testing for over 24 hours with a large macOS backlog still at TravisCI.  As all other testing has succeeded and these changes are all guarded with ARM `#ifdefs` I am confident this can be merged.  Merging...

PR #2259, opened 2018-01-25
DDRgen macro constant expression evaluation

### LINKED ISSUE #2100: Improve ddrgen handling of constant expressions
The tool ddrgen needs to capture constants in the blob it generates. Currently it doesn't do such a good job.

For example,
```
#define MASK (0x1 | 0x2 | 0x4)
```
only considers the first number it finds inside the inner-most set of parentheses and interprets `0x1` as `0`.

Even if those flags were not expressed in hexadecimal,
```
#define MASK (1 | 2 | 4)
```
it would see `1` which is still the wrong answer.

Several constants defined by OMR (or OpenJ9 which is likely to be the first consumer of ddrgen) use C operators: these need to be properly handled by ddrgen. The full list of C operators in order of precedence is available at http://en.cppreference.com/w/c/language/operator_precedence.

The problem is well-contained in Macro::getNumeric(long long *).

mikezhang1234567890:
Adds a recursive descent parser in order to parse macros that contain
constant expressions which evaluate down to integers.
The final constant value is returned as a `long long` in order to keep
compatibility with the rest of the ddr components.

Closes #2100

### DIFF
--- a/ddr/lib/ddr-ir/Macro.cpp
+++ b/ddr/lib/ddr-ir/Macro.cpp
@@ -21,53 +21,974 @@
 
 #include "ddr/ir/Macro.hpp"
 
-#include <stdlib.h>
-
-#if defined(_MSC_VER)
-/* Older Visual Studio compilers don't provide strtoll(). */
-#define strtoll(str, end, base) _strtoi64((str), (end), (base))
-#endif
+#include "ddr/std/sstream.hpp"
+#include <ctype.h>
+#include <string.h>
+#include <stdint.h>
+#include <algorithm>
+#include <vector>
 
 using std::string;
+using std::stringstream;
 
-DDR_RC
-Macro::getNumeric(long long *ret) const
+enum Symbol {
+	OR_OR,		/* || */
+	AND_AND,	/* && */
+	OR,			/* | */
+	XOR,		/* ^ */
+	AND,		/* & */
+	EQ_EQ,		/* == */
+	NOT_EQ,		/* != */
+	GT,			/* > */
+	LT,			/* < */
+	GT_EQ,		/* >= */
+	LT_EQ,		/* <= */
+	GT_GT,		/* >> */
+	LT_LT,		/* << */
+	PLUS,		/* + */
+	MINUS,		/* - */
+	MULT,		/* * */
+	DIV,		/* / */
+	REM,		/* % */
+	NOT,		/* ! */
+	TILDE,		/* ~ */
+	LPAREN,		/* ( */
+	RPAREN,		/* ) */
+	END			/* '\0' or null terminator */
+};
+
+class MacroScanner
 {
-	DDR_RC rc = DDR_RC_ERROR;
-	const string &value = getValue();
-	const char * const whitespace = "\t ";
+private:
+	char const *_cursor;
 
-	for (size_t start = 0, end = value.length() - 1; start <= end;) {
-		/* trim leading whitespace */
-		start = value.find_first_not_of(whitespace, start);
-		if (string::npos == start) {
-			break;
+	static bool isadigit(char c, int base);
+	static int digitvalue(char c);
+
+public:
+	MacroScanner(char const *cString) :
+		_cursor(cString)
+	{
+	}
+
+	bool validTypeCast(bool *isSigned, size_t *bitWidth);
+	bool atSymbol(Symbol sym);
+	bool readNumber(int64_t *ret);
+};
+
+class MacroParser
+{
+/* data members */
+private:
+	MacroScanner _scanner; /*handles reading characters from expression string*/
+
+/* function members */
+private:
+	bool logicalOr(int64_t *ret);
+	bool logicalAnd(int64_t *ret);
+	bool bitwiseOr(int64_t *ret);
+	bool bitwiseXor(int64_t *ret);
+	bool bitwiseAnd(int64_t *ret);
+	bool equalNeq(int64_t *ret);
+	bool comparison(int64_t *ret);
+	bool bitshift(int64_t *ret);
+	bool addSub(int64_t *ret);
+	bool multDivRem(int64_t *ret);
+	bool unaryExpression(int64_t *ret);
+	bool parentheses(int64_t *ret);
+	bool validSingleTerm(int64_t *ret);
+
+public:
+	MacroParser(char const *cExpression) :
+		_scanner(cExpression)
+	{
+	}
+
+	bool evaluate(int64_t *ret);
+};
+
+/**
+ * Helper function to read the value of a number from a string, this checks if
+ * a character is a digit in a specified base
+ *
+ * @param[in] c: character being checked
+ * @param[in] base: the base used to check if the char is valid
+ *
+ * @return: if the character is a valid digit in the specified base
+ */
+bool
+MacroScanner::isadigit(char c, int base)
+{
+	if (base > 10) {
+		return ((c >= '0') && (c <= '9')) || ((c >= 'a') && (c < ('a' + base - 10)))
+				|| ((c >= 'A') && (c < ('A' + base - 10)));
+	} else {
+		return (c >= '0') && (c < ('0' + base));
+	}
+}
+
+/**
+ * Helper function to read the value of a number from a string, this gets the
+ * value of a single digit, assuming digits go 0-9 then A-Z case-insensitive
+ * (36 possible digits)
+ *
+ * @param[in] c: digit whose value is sought
+ *
+ * @return: integer value of the digit
+ */
+int
+MacroScanner::digitvalue(char c)
+{
+	if ((c >= 'a') && (c <= 'z')) {
+		return 10 + c - 'a';
+	} else if ((c >= 'A') && (c <= 'Z')) {
+		return 10 + c - 'A';
+	} else {
+		return c - '0';
+	}
+}
+
+/**
+ * Checks if the cursor is at a valid set of type cast expressions, after
+ * encountering a left parenthesis. If it is, determine the signedness and bit
+ * width of the type cast. If not, move the cursor back to the left parenthesis.
+ *
+ * @param[out] isSigned: the signedness of the resulting typecast
+ * @param[out] bitWidth: the bit width of the resulting typecast
+ *
+ * @return: if the typecast expression is valid
+ */
+bool
+MacroScanner::validTypeCast(bool *isSigned, size_t *bitWidth)
+{
+	static const string fixedWidthTypes[] = {"int8_t", "int16_t", "int32_t", "int64_t",
+	 "uint8_t", "uint16_t", "uint32_t", "uint64_t",
+	 "intptr_t", "uintptr_t"};
+
+	if (!atSymbol(LPAREN)) {
+		return false;
+	}
+
+	char const *typecastStart = _cursor;
+	_cursor = strchr(_cursor, ')');
+	if (NULL == _cursor) {
+		_cursor = typecastStart - 1;
+		return false;
+	}
+
+	string typeExpr(typecastStart, _cursor);
+	stringstream ss(typeExpr);
+	std::vector<string> types;
+	string word;
+	bool validTypes = false;
+
+	while (ss >> word) {
+		types.push_back(word);
+	}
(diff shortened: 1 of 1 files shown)


mikezhang1234567890:
Reviewer: @keithc-ca

ddr/include/ddr/ir/Macro.hpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2017 IBM Corp. and others
+ * Copyright (c) 2015, 2018 IBM Corp. and others
```
keithc-ca: Please revert this change.

fjeremic: > Please revert this change.

From an outside perspective, any reason why? Was this file crafted in 2017 and unchanged in the new year?

keithc-ca: That's right; the copyright date only needs to be updated if the file was changed (in some other way).

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -26,56 +26,563 @@
...
+/* function members */
+private:
+	bool binaryExpression(int64_t *, binExpr_type);
+	bool unaryExpression(int64_t*);
+	bool parentheses(int64_t*);
+	bool readNumber(int64_t*);
+	void skipWhitespace();
+
+public:
+	MacroParser(std::string expression) : _expression(expression), _index(0)
```
charliegracie: The variables are being initialized in the wrong order.  They need to be initialized in declaration order since that is the order they will *actually* be initialized in. 

Could you also move them each onto their own line for clarity?

mikezhang1234567890:
@keithc-ca Can you review again please?

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+	RPAREN		/* ) */
+};
+
+class MacroScanner
 {
-	DDR_RC rc = DDR_RC_OK;
-	string value = getValue();
+private:
+	char const *_cursor;
+	char const *_end;
```
keithc-ca: I don't think this field (`_end`) is required.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+	MacroScanner(string expression) :
+		_cursor(expression.c_str()),
+		_end(_cursor + expression.length())
+	{
+	}
+
+	bool atSymbol(Symbol sym);
+	bool readNumber(int64_t *ret);
+
+	friend class MacroParser;
```
keithc-ca: Make the required methods public rather than use a friend declaration.

FYI, there should be a forward declaration for `class MacroParser` before it can be meaningfully referenced in a friend declaration.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
-			ss << value;
-			long long valueNumeric = 0;
-			ss >> valueNumeric;
-			if (ss.fail()) {
-				rc = DDR_RC_ERROR;
-			} else if (NULL != ret) {
-				*ret = valueNumeric;
-			}
+	MacroScanner(string expression) :
+		_cursor(expression.c_str()),
```
keithc-ca: Change the signature of the constructor so the argument is of type `const char *`. As written, `_cursor` points at memory that will soon be freed.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ * Helper function to read the value of a number from a string, this checks if
+ * a character is a digit in a specified base
+ *
+ * @param[in] c: character being checked
+ * @param[in] base: the base used to check if the char is valid
+ *
+ * @return: if the character is a valid digit in the specified base
+ */
+static bool
+isadigit(char c, int base)
```
keithc-ca: This function (and 'digitvalue') should be private static methods of MacroScanner.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+			return true;
+		}
+		break;
+	case RPAREN:
+		if (')' == _cursor[0]) {
+			_cursor += 1;
+			return true;
+		}
+		break;
+	}
```
keithc-ca: A switch statement should always have a default case.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+/**
+ * Reads an integer number starting at the cursor in the expression string.
+ * Number is read as a 64bit integer.
+ */
+bool
+MacroScanner::readNumber(int64_t *ret)
+{
+	uint64_t retval = 0;
+	int base = 0;
+	bool negative = false;
```
keithc-ca: A `-` character is not part of a numeric literal - it is the unary negation operator.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+	for (int i= 0; (_cursor < _end) && (i < 3); i++, _cursor++) {
+		if (('l' == _cursor[0]) || ('L' == _cursor[0]) ) {
+			numL += 1;
+		} else if (('u' == _cursor[0]) || ('U' == _cursor[0])){
+			numU += 1;
+		} else {
+			break;
+		}
+	}
+	if (2 < numL || 1 < numU) {
```
keithc-ca: Parentheses missing here.
I personally find it difficult to read such comparisons: I think `((numL > 2) || (numU > 1))` is clearer.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ * @return: whether the expression is valid or not
+ */
+bool
+MacroParser::evaluate(int64_t *ret)
+{
+	bool validExpr = logicalOr(ret);
+	if (validExpr) {
+		return true;
+	}
+	return false;
```
keithc-ca: This can be just `return logicalOr(ret);`.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ *
+ * @param[out] ret: the integer value the logical or expression evaluates to
+ *
+ * @return: whether the at the current index the expression can be evaluated
+ */
+bool
+MacroParser::logicalOr(int64_t *ret)
+{
+	int64_t term1 = 0;
+	int64_t term2 = 0;
```
keithc-ca: Why not just use `ret` and `term2`? What value is added by declaring `term1`?

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ * @return: whether the at the current index the expression can be evaluated
+ */
+bool
+MacroParser::multDivRem(int64_t *ret)
+{
+	int64_t term1 = 0;
+	int64_t term2 = 0;
+
+	if (unaryExpression(&term1) || parentheses(&term1)
+			|| _scanner.readNumber(&term1)) {
```
keithc-ca: I think another method is warranted to encapsulate the recurring pattern:
```
unaryExpression(X) || parentheses(X) || _scanner.readNumber(X)
```

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ * applies a unary typecast operator
+ *
+ * @param[in] typeExpr: string representing the types inside a typecast operator
+ * @param[in] operand: value of the operand the typecast is being applied to
+ * @param[out] ret: value after the typecast has been applied
+ *
+ * @return: whether the typecast is valid and has been successfully applied
+ */
+static bool
+typeCast(const string typeExpr, int64_t operand, int64_t *ret)
```
keithc-ca: I think we should first just ignore casts (besides, this doesn't handle things like `uintptr_t` or `UDATA`).

Removing the friendship between the scanner and parser classes will require some rework. I expect the scanner will need to recognize a sequence of words separated by whitespace for use by the parser to handle casts.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -25,57 +25,951 @@
...
+ */
+DDR_RC
+Macro::getNumeric(long long *ret)
+{
+	DDR_RC rc = DDR_RC_ERROR;
+	string value = getValue();
+	int64_t retVal = 0;
+	MacroParser parser(value);
+	if (parser.evaluate(&retVal)) {
+		*ret = retVal;
```
keithc-ca: This must first check that `NULL != ret`.

mikezhang1234567890:
@keithc-ca Can you please take another look?

charliegracie:
@genie-omr build all

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -21,53 +21,962 @@
...
+	bool bitshift(int64_t *ret);
+	bool addSub(int64_t *ret);
+	bool multDivRem(int64_t *ret);
+	bool unaryExpression(int64_t *ret);
+	bool parentheses(int64_t *ret);
+	bool validSingleTerm(int64_t *ret);
+
+public:
+	MacroParser(std::string expression) :
+		_scanner(expression.c_str())
```
keithc-ca: This leads to a dangling pointer.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -21,53 +21,962 @@
...
+				validTypes = true;
+			} else if ((1 >= numSigned + numUnsigned) && (0 == numChar)
+					&& (0 == numShort) && (0 == numInt) && (0 == numLong)) {
+				*bitWidth = 0;
+				validTypes = true;
+			}
 		}
+	}
+	if (!validTypes) {
+		_cursor = typecastStart - 1;
```
keithc-ca: Why the `- 1`? Shouldn't it restore _cursor to the value on entry?

mikezhang1234567890: It restores `_cursor` to the `(` symbol before the string of types.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -21,53 +21,962 @@
...
+		size_t bitWidth = 0;
+		bool isSigned = true;
+		if (_scanner.validTypeCast(&isSigned, &bitWidth)
+				&& _scanner.atSymbol(RPAREN) && validSingleTerm(&operand)) {
+			switch(bitWidth) {
+			case 0:
+				*ret = (isSigned) ? (signed) operand : (unsigned) operand;
+				break;
+			case 8:
+				*ret = (isSigned) ? (int8_t) operand : (size_t) operand;
```
keithc-ca: The second cast should be `(uint8_t) operand`.

ddr/lib/ddr-ir/Macro.cpp
keithc-ca: That would be reasonable if the scanner (rather than the parser) considered the opening parenthesis.

ddr/lib/ddr-ir/Macro.cpp
```diff
@@ -21,53 +21,975 @@
...
+ *
+ * return: DDR return code indicating success or error
+ */
+DDR_RC
+Macro::getNumeric(long long *ret) const
+{
+	DDR_RC rc = DDR_RC_ERROR;
+	string value = getValue();
+	int64_t retVal = 0;
+	MacroParser parser(value.c_str());
```
keithc-ca: There's no need to make a copy of `_value` here; the parser can be constructed from `_value.c_str()`.

charliegracie:
@mikezhang1234567890  can you squash the commits into the correct amount of logical changes (likely just 1)?  Thanks

mikezhang1234567890:
Made changes as requested and squashed commits. Is there anything else that needs addressing? @keithc-ca @charliegracie

PR #2260, opened 2018-01-26
CMake: Remove superfluous call to make_compiler_target

dnakamura:
testcompiler already exports the required include paths etc.
Since compilertest links against testcompiler it gets these for free


charliegracie:
@genie-omr build zlinux zos plinux

PR #2263, opened 2018-01-26
Added 16 bit padding in HashtableInputData 

Varun-garg:
Added a 16 bit padding variable in `struct HashtableInputData` in `algorithm_test_internal.h` to fill 32 byte alignment.
This fixes four (all) errors generated by `omralgotest` on valgrind when gtest tries to print `HashtableInputData` and supposedly using `sizeof()` for it. `sizeof()` returned 32 for struct, resulting in print operations using remaining  un-initialized 16 bits causing these errors. 

[Valgrind errors fixed](https://github.com/eclipse/omr/files/1668617/Algotest_logs.txt)

Signed-off-by: Varun Garg <varun.10@live.com>

### DIFF
--- a/fvtest/algotest/algorithm_test_internal.h
+++ b/fvtest/algotest/algorithm_test_internal.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2015 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -50,6 +50,7 @@ typedef struct HashtableInputData {
 	uint32_t listToTreeThreshold;
 	BOOLEAN forceCollisions;
 	BOOLEAN collisionResistant;
+	uint16_t padding;
 } HashtableInputData;
 
 /* ---------------- avltest.c ---------------- */


charliegracie:
@Varun-garg Can you add a comment to the structure discussing the padding field and what steps a developer should use to ensure the struct stays properly aligned if they need to change / add /remove fields?  Also Can you squash the commits into 1?  Thanks

youngar:
Hi, I agree with Charlie if we go with this solution, we need to document what the padding is there for.

However, I think that there might be another way to solve this problem: we can set a custom pretty printer for this structure to avoid Google Test dumping the bytes.  See [teaching-google-test-how-to-print-your-values](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#teaching-google-test-how-to-print-your-values).  If Google Test uses our pretty printing function, it shouldn't try to print uninitialized data, and we won't need the padding field.  Something along these lines should work:

```C++
::std::ostream& operator<<(::std::ostream& os, const HashtableInputData& input) {
	return os << input.hashtableName << " forceCollisions=" << input.forceCollisions;
}
```

Does this seem like it might solve the problem @Varun-garg?

Varun-garg:
Thanks @charliegracie and @youngar for good suggestions.

From what I understood about alignment [(from here)](http://www.songho.ca/misc/alignment/dataalign.html), alignment of a data structure means adding extra space (padding) in it to make its size a multiple of a number (depending on processor's architecture) so that it's address can be divisible by  1, 2, 4, or 8. This has a great impact on performance. 

The size of aligned struct is usually multiple of 4, 8, 16, 32 bytes (needs review).

GCC and some compilers automatically add this padding when compiling a program. This can be seen when I tried following:
```
typedef struct HashtableInputData {
	const char* hashtableName;  // 8bytes
	const uintptr_t* data;            // 8 bytes
	uintptr_t dataLength;            // 8 bytes
	uint32_t listToTreeThreshold; // 4 bytes
	bool forceCollisions;               // 1 byte
	bool collisionResistant;           // 1 byte
} HashtableInputData;                   // 30 bytes

cout <<sizeof(HashtableInputData) //32 bytes. (2 padding unitialized bytes added automatically)
```
This (automatic alignment compared to manual alignment) has 2 slight disadvantages

1. When we try to access memory of  the struct without knowing its contents with help of sizeof(), for example logging it, we will end up accessing (printing) these two uninitialized bytes. The problem may get worse when someone tries to debug the logs and confuses these two padding bytes with first two of next variable.
2. When we are already wasting away two bytes, why not declare them in the struct for future purposes.

If performance is not an issue, padding can be [disabled](https://stackoverflow.com/a/40642888/3276830) (in GCC) by adding `__attribute__((packed))` at end of structure.

@youngar That looks very good, it can solve our problem and improve readability of logs. I will try it out.

charliegracie:
I would not keep the padding if the new print function resolves the issue.

charliegracie:
@genie-omr build all

charliegracie:
@Varun-garg ping?

charliegracie:
If there are no issues I would suggest closing I guess.

Issue #2256, opened 2018-01-23
Remove VMThread reclamation logic


0xdaryl:
The deprecated VMThread reclamation logic will be removed from the Eclipse OMR compiler.  This is logic that was introduced by OpenJ9 many years ago to re-use the register normally reserved for the J9VMThread in areas of high register pressure. If the compiler can prove that the VMThread is not needed in a context where registers are required to avoid spilling then it may spill the VMThread register to free it for general register assignment. There is also logic to use the VMThread register in a GRA context, as well as a late tree-lowering pass (late edge splitting) to help with reclamation placement.

This is technology that was only ever enabled on 32-bit Windows for OpenJ9 and saw limited performance gains.  It has since been disabled (via #2241) in preparation for its removal.

The only known downstream project that has a dependency on this feature is OpenJ9, and the work in this issue will be staged with corresponding PRs in OpenJ9 to ensure all parts are removed with minimal disruption.

This is the umbrella epic to track all the removal work in Eclipse OMR.  This work includes, but is not expected to be limited to, the following:

- [x] disable VMThread reclamation (#2241)
- [ ] remove VMThread enablement queries and fold surrounding code
- [ ] remove VMThread GRA optimization and from lowerTrees
- [ ] remove late edge splitting optimization
- [ ] remove VMThread preservation/restoration/spill placement logic in CodeGenerator
- [ ] remove all `isVMThreadRequired`
- [ ] remove VMThread infrastructure in CodeGenerator

PR #2266, opened 2018-01-27
Remove more VMThread reclamation logic from the x86 CodeGenerator

0xdaryl:
* Disable late edge splitting
* Do not add VMThread pre/post conditions on x86
* Do not override initial register state of EBP on Win32
* Remove vmThread rematerialization logic from OutlinedInstructions
* Remove uses of getSupportsVMThreadGRA()

Issue: #2256

### DIFF
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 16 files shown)


0xdaryl:
Notes to reviewers:

* Because testing using the FV tests in TravisCI does not exercise these code paths extensively, I ran an OpenJ9 build within IBM on Linux and Windows 32/64-bit platforms using promotion-calibre tests and all tests were successful.
* I chose to stage this removal with multiple commits to make triage easier should OMR or downstream testing uncover problems relating to this in the future.
* Late edge splitting was only disabled, and code paths guarded by that option were folded so that further refactoring progress could be made.  All code relating to LES will be removed in subsequent staged commits.
* APIs to branch instruction generators have a boolean indicating whether the VMThread requires register dependencies.  The internal logic within those generators has been removed, so it doesn't matter what the caller passes in (true/false).  The API to these functions will be changed in a subsequent commit.

I would appreciate a review from @0dvictor please.

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1149,11 +1128,7 @@ void OMR::X86::TreeEvaluator::compareBytesForOrder(TR::Node *node, TR::CodeGener
 
 TR::Register *OMR::X86::TreeEvaluator::gotoEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   TR::Compilation *comp = cg->comp();
-   bool needVMThreadDep =
-      comp->getOption(TR_DisableLateEdgeSplitting) ||
-      !performTransformation(comp, "O^O LATE EDGE SPLITTING: Omit ebp dependency for %s node %s\n", node->getOpCode().getName(), cg->getDebug()->getName(node));
-   generateJumpInstruction(JMP4, node, cg, needVMThreadDep);
+   generateJumpInstruction(JMP4, node, cg, false);
```
fjeremic: Will the parameters of APIs be cleaned up in a later commit as part of #2256?

0xdaryl: Yes, that is my plan as part of my incremental approach to untangling this.  Perhaps I'm being overly-conservative, though, so I could be talked into changing it for this PR.

fjeremic: > Yes, that is my plan as part of my incremental approach to untangling this. Perhaps I'm being overly-conservative, though, so I could be talked into changing it for this PR.

No need. Just confirming it will happen in the future. Incremental deprecation is appreciated.

0dvictor:
LGTM.

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build all

Retrying again as I had to be added to a whitelist for this to work.

Issue #2068, opened 2017-11-28
Relocate createOrFindClonedNode to OpenJ9


0xdaryl:
The following `OMR::CodeGenerator` function only has relevance to the OpenJ9 project.  It should be relocated to the `J9::CodeGenerator` class in that project instead.
```
TR::Node *createOrFindClonedNode(TR::Node *node, int32_t numChildren);
```
Also, the `_uncommmonedNodes` field and member initialization must be relocated.

PR #2269, opened 2018-01-28
Relocate createOrFindClonedNode, _uncommmonedNodes to OpenJ9

shivammittal99:
Closes: #2068 

Signed-off-by: Shivam Mittal <shivammittal99@gmail.com>

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -182,7 +182,6 @@ OMR::CodeGenerator::CodeGenerator() :
       _implicitExceptionPoint(0),
       _localsThatAreStored(NULL),
       _numLocalsWhenStoreAnalysisWasDone(-1),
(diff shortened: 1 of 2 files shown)


charliegracie:
ping @0xdaryl and @mstoodle

charliegracie:
@genie-omr build all

mstoodle:
there's a dependent PR at OpenJ9 which seems like it would be needed or else merging this could break openj9, so I'll be nice to them and not "just" merge this one even though it seems ready.

In future, maybe we could try to stage changes like this by first making the OMR code #ifdef J9_PROJECT_SPECIFIC, then copy it to OpenJ9, and then remove it from OMR. I know that's more pull requests and more work on the writing of commits, but it would certainly simplify and streamline the process to merge this kind of change.

0xdaryl:
Thank you for the contribution.  This work was subsumed by #3857.  Closing.

PR #2186, opened 2017-12-25
Implement write barriers for arraycopy for ARM

knn-k:
OMR::ARM::TreeEvaluator::arraycopyEvaluator() asserts when passed
a non-simple arraycopy node.
This implements new functions genWrtbarForArrayCopy() and
VMCardCheckEvaluator() for handling non-simple arraycopy nodes.

Closes: #1996

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #2270, opened 2018-01-29
Enable generation of ReferenceArrayCopy for ARM

knn-k:
Call setSupportsReferenceArrayCopy() in the ARM JIT.
Write barrier for arraycopy was enabled for ARM by #2186.


0xdaryl:
A genie build is not required because this is an ARM-specific change.

0xdaryl:
While I agree with this change because it is symmetrical with the other code generators, I think setting this property should ultimately be set in the OpenJ9 code generators.  This PR need not accomplish that, but I will open an issue to track that work.

PR #2271, opened 2018-01-29
Fix multiple issues with ddrgen

keithc-ca:
This is in support of https://github.com/eclipse/openj9/issues/378.

Fixes & improvements
* blacklist pattern matching
* close list files
* avoid typedef cycles
* write proper enum size to blob
* handle CR/LF line-ends reading a list files
* pass const references
* favour constructor use over assignment
* formatting
* improve safety
* mark constructors explicit as appropriate
* improve DEBUGPRINTF and ERRMSG macros and use consistent format
* remove redundant or unnecessary #includes
* add constructor for Dwarf_Attribute_s
* add Dwarf_Attribute_s._flag
* add dwarf_formflag() for OSX
* omit static fields from expected superset.out
* improve handling of vtable pointers
* improve 'getmacros' script (and don't let it change timestamps)
* use 'find' to avoid scanning unintersting files
* pull regex definitions out of the inner loop
* fix references to ddr/tools/getmacros
* improve test code
* add an example of map_to_type= without following whitespace
* improve namespace handling
* improve getmacros script
* replace use of 'awk' with 'sed'
* fix @MACRO emissions
* tighten API of ddrgen
* fix validation that macro values are constant
* streamline MacroTool
* fix constant evaluation (including numeric prefixes like '0x')
* fix problem reading macrolist
* use const types where appropriate
* use canonical type expressions
* make as many *.i files as possible
* be careful constructing name to type map for macros
* implicitly blacklist illegal type names
* remove overloading in TypeVisitor
* allow writing of blob and superset independently
* simplify blacklist handling
* make reading blacklist more robust
* fix override support
* fix handling of #undef by getmacros
* ignore linkage_name DWARF attribute
* add ClassType::_isComplete
* construct all fields explicitly
* relocate constants to J9HashtableConstants
* repeat structure for typedef with distinct name

ddrgen
* consider macros in ClassType
* fix Type::renameFieldsAndMacros
* don't require values for enum literals
* don't trim trailing *[] from override value
* generalize support for opaque types
* include typedef with distinct name
* new strategy for eliminating duplicate types
* remove special handling for CLimits
* remove unused _offset field in Modifiers
* reset modifiers with field type override

Blob generation
* don't duplicate fields in blob
* extract common code
* propagate failure to parent process

getmacros
* fix handling of ddr_options: valuesandbuildflags
* improve include guard detection
* fix constant handling

would previously map to the following
(independent of whether A was defined or not)
```
@MACRO_A_DEFINED A_DEFINED 1
@MACRO_A_DEFINED A_DEFINED 0
```

### DIFF
--- a/ddr/lib/ddr-ir/Modifiers.cpp
+++ b/ddr/lib/ddr-ir/Modifiers.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2017 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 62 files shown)


charliegracie:

ddr/lib/ddr-ir/Modifiers.cpp
```diff
@@ -23,54 +23,56 @@
...
-		s += "&";
+	for (size_t i = 0; i < _referenceCount; ++i) {
+		s << "&";
 	}
 
-	if (isArray()) {
-		for (size_t i = 0; i < getArrayDimensions(); i++) {
-			s += "[]";
-		}
+	for (size_t i = getArrayDimensions(); i != 0; i -= 1) {
```
charliegracie: I guess getArrayDimensions() returns 0 if isArray() returns false?

ddr/lib/ddr-scanner/Scanner.cpp
```diff
@@ -19,93 +19,131 @@
...
+
+	return false;
+}
+
+static bool
+stringMatchesAny(const string &candidate, const set<string> &patterns)
+{
+	bool match = false; // patterns.find(candidate) != patterns.end();
+
+	if (!match) {
```
charliegracie: Can we get rid of match?

ddr/lib/ddr-ir/Modifiers.cpp
keithc-ca: Yes.

ddr/lib/ddr-scanner/Scanner.cpp
keithc-ca: Will do.

keithc-ca:
@charliegracie Any more feedback or questions?

PR #2272, opened 2018-01-29
On linux 390 make sure to target z9-190 as the architecture

charliegracie:
With the move to CMake some assembly files get assembled using
GCC instead of AS. On Linux 390 some of the instructions used in
the port library assembly files are only supported on newer
architectures so we should set `-march=z9-190`. The compiler
technology is already setting this value for assembler files.


charliegracie:
@fjeremic and @dnakamura can you give this a quick review?

charliegracie:

PR #2273, opened 2018-01-29
Facility for marking assumptions for later removal

klangman:
Allows for marking assumptions which will be removed later.
Used in OpenJ9 to speed up removal of assumptions during
class unloading.

Signed-off-by: Kevin Langman <langman@ca.ibm.com>

### DIFF
--- a/compiler/runtime/OMRRuntimeAssumptions.hpp
+++ b/compiler/runtime/OMRRuntimeAssumptions.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -48,6 +48,11 @@ enum RuntimeAssumptionCategory
    ValueModification,
    SentinelCategory,
    LastRuntimeAssumptionCategory = SentinelCategory
+   };
+
+enum AssumptionSameJittedBodyFlags
+   {
+   MARK_FOR_DELETE=1
    };
 
 class RuntimeAssumption : public TR_Link0<RuntimeAssumption>
@@ -106,9 +111,23 @@ class RuntimeAssumption : public TR_Link0<RuntimeAssumption>
       setNext(NULL);
       }
 
+   /**
+    * Mark this assumption for future removal
+    * The caller of this routine must insure that the assumption has been dequeueFromListOfAssumptionsForJittedBody
+    * prior to this call. Using this marking feature will allow for removing a number of items from the linked-list
+    * of assumptions using one pass through the linked-list
+    */
+   void markForDetach() { _nextAssumptionForSameJittedBody = (RuntimeAssumption *)(((uintptr_t)_nextAssumptionForSameJittedBody)|MARK_FOR_DELETE); }
+
+   /**
+    * Returns true when the assumption has previously been marked using the markForDetach routine
+    */
+   bool isMarkedForDetach() { return(((uintptr_t)(_nextAssumptionForSameJittedBody)&MARK_FOR_DELETE) == MARK_FOR_DELETE); }
+
    protected:
    RuntimeAssumption *_nextAssumptionForSameJittedBody; // links assumptions that pertain to the same method body
-                                                           // These should form a circular linked list with a sentinel
+                                                        // These should form a circular linked list with a sentinel
+                                                        // Lower bit is used to mark assumptions that will be removed after all marking is completed
    uintptrj_t _key; // key for searching in the hashtable
    };
 


klangman:
@mpirvu Can you review this change?

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -105,10 +105,13 @@ class RuntimeAssumption : public TR_Link0<RuntimeAssumption>
       _nextAssumptionForSameJittedBody = 0;
       setNext(NULL);
       }
```
0xdaryl: I'd like to see Doxygen comments for new functions added to the API please.

compiler/runtime/OMRRuntimeAssumptions.hpp
```diff
@@ -105,10 +105,13 @@ class RuntimeAssumption : public TR_Link0<RuntimeAssumption>
...
+   void markForDetach() { _nextAssumptionForSameJittedBody = (RuntimeAssumption *)(((uintptrj_t)_nextAssumptionForSameJittedBody)|1); };
```
fjeremic: Do we need to update `getNextAssumptionForSameJittedBody` and `setNextAssumptionForSameJittedBody`? It seems that if this is a true circular buffer we could end up dereferencing an assumption with the "removable" bit set which would cause runtime errors.

Is the overhead of an extra field too high in this scenario to warrant this bit hackery? At the very least this bit needs to be given a proper name via an `enum` for example and the qualified name should be used throughout rather than a magic constant.

compiler/runtime/OMRRuntimeAssumptions.hpp
fjeremic: Might as well also properly Doxygen `_nextAssumptionForSameJittedBody` since we are changing its meaning.

compiler/runtime/OMRRuntimeAssumptions.hpp
mpirvu: Perhaps an issue describing the problem we are trying to solve and a sketch of the solution is warranted here.
As Kevin explained to me, marking (setting the last bit of` _nextAssumptionForSameJittedBody`) is done when the body containing the assumption is about to be reclaimed. We start at the `sentinel` and end at the `sentinel`. From that moment on the circular list of assumptions belonging to the same body is no longer going to be traversed so the last bit being set should not cause problems.

The most elegant solution to the overhead encountered when detaching from the RAT (Runtime Assumption Table) is to make the buckets of the RAT a doubly linked list. This means adding another pointer to each runtime assumption which in turn will increase the footprint (assumptions could be in the millions).

I totally agree with having a symbolic value for that bit. That would be more readable and would allow searching.

fjeremic: > From that moment on the circular list of assumptions belonging to the same body is no longer going to be traversed so the last bit being set should not cause problems.

I don't think I'm understanding this part. Why mark the bit if the list will never be traversed again? Who tests this bit and when? Is it possible to outline the algorithm which sets and tests this bit? Can some entries have the bit set and others not?

I suppose my feedback here is that the intention of this bit is not clear at all to a potential downstream project attempting to consume this API. We need to exhaustively document what the exact intention is, how it is meant to be consumed, and if there are any implicit assumptions we are imposing on the consumers of this API.

klangman: The setting of the bit and the removal of the assumption is done in one stop-the-world event. Therefore it's not possible for any traversal of the linked-list when the bit is set on the nextAssumptionForSameJittedBody field. In this way we are able to mark all the assumptions one at a time, then remove all the assumption all in one pass over the assumption table resulting in a massive reduction in pause time when thousands of classes are being removed in one GC cycle. As Marius pointed out, this is a compromise that allows for good performance with minimal increase in footprint. The original method for removing assumptions remains, this is just another option that allows for better performance if implemented correctly.

fjeremic: So if the assumptions are marked and then removed in lock-step could we not use a local hashtable to keep track of which allocations to remove?

joransiu: Is the savings here that we can do a bulk remove of consecutive assumptions in a single post pass (i.e. 1 set of pointer updates per bulk remove), vs the cost of having to individually remove individual entries in the list (1 set of updates per entry removed)?

mpirvu: I was talking to Filip and one suggestion that emerged from the discussion (part from the need for documentation) was to completely kill the `_nextAssumptionForSameJittedBody` pointer with an eye catcher (like DEADFOOD) rather than setting the last bit. This way it's clear that we don't plan on using the pointer anymore.

klangman: For debugging purposes, I think we should leave it as a low bit tag. If we get a core file with 0xDEADFOOD as the next pointer we have less data to work with then we would if we left it as a valid next pointer with only the lower bit set.

vijaysun-omr:
genie-omr build all

PR #2275, opened 2018-01-30
Direct old-to-old-reference callback

amicic:
Bypass hook API tp reduce overhead, mostly by avoiding the hook internal
stats counter (which is using an atomic increment).


charliegracie:
@amicic do downstream projects also have to be modified since we removed a hook?

charliegracie:

amicic:
Perhaps yes - you should know better. I can leave the hook, if needed.

amicic:
Well, I'd be really surprised that anybody else would be using this hook. This hook is for a very specific purpose (Scavenger to Concurrent Global GC interaction), and all that code resides within OMR.

charliegracie:
No I do not see any references either.

PR #2276, opened 2018-01-30
CMake: modify add_compiler_subdirectory avoid creating new scope

dnakamura:
By switching the backend from `add_subdirectory` to `include` we can
avoid introducing a new scope.


charliegracie:
@genie-omr build plinux 

No need to test the other platforms yet since they are not using CMake and these are CMake specific changes.

cmake/modules/OmrCompilerSupport.cmake
```diff
@@ -329,23 +329,13 @@ function(create_omr_compiler_library)
 	# Add the contents of the macro to CORE_COMPILER_OBJECTS in this scope.
 	# The library name parameter is currently ignored.
 	macro(compiler_library libraryname)
-		# Since CMake *copies* the enviornment from parent to child scope,
-		# a PARENT_SCOPE set ultimately is not reflected for local lookups.
-		# As a result, we need to _also_ update the local scope here if we
-		# want this to succeed when called multiple times in a scope.
-		set(CORE_COMPILER_OBJECTS ${CORE_COMPILER_OBJECTS} ${ARGN}             )
-		set(CORE_COMPILER_OBJECTS ${CORE_COMPILER_OBJECTS} ${ARGN} PARENT_SCOPE)
+		list(APPEND CORE_COMPILER_OBJECTS ${ARGN})
```
charliegracie: Can you explain the differences between the lines you removed and the line you added?

dnakamura: The second line, `set(CORE_COMPILER_OBJECTS ${CORE_COMPILER_OBJECTS} ${ARGN} PARENT_SCOPE)` is no longer required since we are no longer introducing a new scope. 
`set(CORE_COMPILER_OBJECTS ${CORE_COMPILER_OBJECTS} ${ARGN})`, is effectively the same as the simple `list(APPEND ...)` although I feel the `list(APPEND...)` form more clearly shows our intent

charliegracie: @rwy0717 I think this is good.  Your thoughts?

rwy0717:
a little late but, LGTM.

PR #2277, opened 2018-01-30
Fix for malfunctional env var TR_DisableAVX on X86

0dvictor:
TR_DisableAVX was malfunctional since the mask to unset XSAVE bits were incorrect.
Fixing it.


0xdaryl:
A genie build is not necessary as this is purely an x86 change.

Issue #2280, opened 2018-01-31
Fix GRA for Vector registers on 32bit x86


IBMJimmyk:
Vector Register GRA can cause a problem that leads to the generation of an incorrect result. GRA for vector registers should be disabled on x86 until this issue is resolved.

The problem here affects both 64bit and 32bit. It is currently suspected to be a different issue than the one described in Issue #2035.

0xdaryl:
Would you mind articulating in greater detail what the problem is please?  The symptom is fairly vague, and since the solution is to disable the feature entirely I'd say that warrants a more detailed description of the problem.  Thanks.

IBMJimmyk:
While running a large Java test an incorrect result error was triggered. The incorrect result was being produced by a loop that increments a long type counter. Attempts to recreate the failure in a smaller and more manageable test case have unfortunately not been successful.

A binary search on change sets traced the problem back to the changes in PR #1766. The loop in question is an AutoSIMD candidate and #1766 expanded AutoSIMD support to Long type add and sub which suggests that AutoSIMD might be the problem. Rerunning the test with the disableautosimd option causes the problem to stop occurring.

Further analysis was performed to narrow down the problem. The trees generated by the SPMDParallelizer optimization that performs AutoSIMD looked to be correct but it proved difficult to fully verify the result of Vector GRA by inspection. An alternate approach was to try and disable Vector Register GRA and this resulted in the test passing. Further analysis is still required to determine what the exact problem in GRA is but due to the potential to trigger incorrect results, I am suggesting this part of the feature be disabled until Vector Register GRA has been fixed. At the current time this is the narrowest known workaround that causes the problem to stop occurring.

mstoodle:
Could we please revisit this problem as the vectorizing regression for regressions caused by #2281 has now come up in 3 separate investigations, two of which have been public reports for the OpenJ9 project (and it's also an issue preventing reduction vectorization from the JitBuilder library for OMR). What's the plan to address this issue @IBMJimmyk @gita-omr ?

PR #2281, opened 2018-01-31
Disable vector reg GRA and reduction on x86

IBMJimmyk:
Vector Reg GRA can cause a problem that leads to the generation of an
incorrect result. GRA for vector registers is temporarily being disabled
on x86 until this issue is resolved.

The removal of GRA may cause a performance problem in autosimd reductions
so reductions have also been disabled. This was done by temporarily
disabling the use of getvelem in autosimd on x86 which is needed
for reduction.

Issue: #2280

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -303,12 +303,16 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
     * GRA does not work with vector registers on 32 bit due to a bug where xmm registers are not being assigned.
     * This disables GRA for vector registers on 32 bit.
     * This code will be reenabled as part of Issue 2035 which tracks the progress of fixing the GRA bug.
+    * GRA does not work with vector registers on 64 bit either. So GRA is now being disabled vector registers.
+    * This code will be reenabled as part of Issue 2280
     */
+#if 0
    if (TR::Compiler->target.is64Bit())
       {
       self()->setFirstGlobalVRF(self()->getFirstGlobalFPR());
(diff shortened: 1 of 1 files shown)


0xdaryl:
Repeating my question from the issue in this PR so it's tracked alongside the code:

Would you mind articulating in greater detail what the problem is please?  Your symptom is fairly vague, and since the solution is to disable the feature entirely for all x86 OMR consumers I'd say that warrants a more detailed description of the problem. Thanks.

IBMJimmyk:
This is the same comment as from Issue #2280. I'm reposting it here so it is more visible:

While running a large Java test an incorrect result error was triggered. The incorrect result was being produced by a loop that increments a long type counter. Attempts to recreate the failure in a smaller and more manageable test case have unfortunately not been successful.

A binary search on change sets traced the problem back to the changes in PR #1766. The loop in question is an AutoSIMD candidate and #1766 expanded AutoSIMD support to Long type add and sub which suggests that AutoSIMD might be the problem. Rerunning the test with the disableautosimd option causes the problem to stop occurring.

Further analysis was performed to narrow down the problem. The trees generated by the SPMDParallelizer optimization that performs AutoSIMD looked to be correct but it proved difficult to fully verify the result of Vector GRA by inspection. An alternate approach was to try and disable Vector Register GRA and this resulted in the test passing. Further analysis is still required to determine what the exact problem in GRA is but due to the potential to trigger incorrect results, I am suggesting this part of the feature be disabled until Vector Register GRA has been fixed. At the current time this is the narrowest known workaround that causes the problem to stop occurring.

0xdaryl:
A genie build is not required because this is an x86-specific change.

PR #2284, opened 2018-02-02
Provide a meaningful x86 register debug name

0dvictor:
Provide a meaningful register name when its size is unknown.


0xdaryl:
A genie build is not required as this is an x86-only change.

PR #2290, opened 2018-02-07
Remove filename spaces

0xdaryl:
Remove spaces from filenames to make them consistent with the rest of
the codebase and to avoid requiring quoting in some contexts.


0xdaryl:
I could not find any references to these files in the OMR website or any other file in the codebase that would need to get fixed up.

fjeremic:
> I could not find any references to these files in the OMR website or any other file in the codebase that would need to get fixed up.

Thanks! I'll let this brew for another day or so just in case someone spots a dependency so they can speak out.

PR #2296, opened 2018-02-08
Change the value of the unneededConv flag

dchopra001:
The unneededConv flag used by TR::x2y nodes overlaps with the simpleDivCheck flag. During node recreation, flags are copied from the original node to the new node without checking if the flag will be valid for the new node being created. This can cause invalid instructions to be generated if the optimizer transforms an ldiv node of two i2l children to i2l of idiv during treeSimplification. To avoid this, this commit changes the unneededConv node's value so that it doesn't overlap with the simpleDivCheck flag.

Closes: #2297
Signed-off-by: Dhruv Chopra <dhruv.c.chopra@ibm.com>

### DIFF
--- a/compiler/il/OMRNode.hpp
+++ b/compiler/il/OMRNode.hpp
@@ -2042,7 +2042,7 @@ protected:
        *     needed because the value the conversion is acting on has been cleansed prior to reaching this evaluation
        *     point.
        */
-      unneededConv                          = 0x00008000, ///< Flag used by TR::x2y
+      unneededConv                          = 0x00000400, ///< Flag used by TR::x2y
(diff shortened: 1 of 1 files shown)


dchopra001:
Review request from @fjeremic at least.

fjeremic:
The new flag value seems to have the least number of conflicts with other existing flags. It looks like `nodeIsDebug` which also shares the same value is not used anywhere. We should deprecate this flag. I've opened https://github.com/eclipse/omr/issues/2299 to track this.

fjeremic:
@genie-omr build all zos,zlinux,plinux

Issue #2299, opened 2018-02-08
Deprecate nodeIsDebug node flag


fjeremic:
As part of #2296 we change the flag value of `unneededConv`. It now shares the same value as `nodeIsDebug`. A quick search through the codebase reveals this flag is not actually used by anyone. We should deprecate this flag.

Issue #2297, opened 2018-02-08
unnededConv flag overlaps with the simpleDivCheck flag


dchopra001:
A bug was recently discovered in openj9 that causes invalid instructions to be generated when, during treeSimplification, an `ldiv` node of two `i2l` children is transformed into an `i2l` of `idiv`. Here's an example from a tracelog showing what happens before and after the transformation is applied:

Before this transformation, the relevant trees look like the following:
```
 n201n     BBStart <block_27> (freq 1338) (extension of previous block)
 n184n     DIVCHK [#26]
 n183n       ldiv (cannotOverflow simpleDivCheck )
 n151n         ==>i2l
 n182n         i2l
 n181n           iload  java/util/stream/AbstractTask.LEAF_TARGET I[#633  final Static] [flags 0x20303 0x0 ] (cannotOverflow )
 n185n     lstore  <auto slot 2>[#634  Auto] [flags 0x4 0x0 ]
 n183n       ==>ldiv
 n190n     iflcmpgt --> block_26 BBStart at n188n ()
 n183n       ==>ldiv
 n187n       lconst 0 (highWordZero X==0 X>=0 X<=0 )
```
After the transformation, they look like below:
```
 n184n     treetop
 n232n       idiv
 n150n         ==>isub
 n181n         iload  java/util/stream/AbstractTask.LEAF_TARGET I[#633  final Static] [flags 0x20303 0x0 ] (cannotOverflow )
 n185n     lstore  <auto slot 2>[#634  Auto] [flags 0x4 0x0 ]
 n183n       i2l (unneededConv cannotOverflow )
 n232n         ==>idiv
 n190n     ificmpgt --> block_26 BBStart at n188n ()
 n232n       ==>idiv
 n187n       iconst 0 (Unsigned X==0 X>=0 X<=0 )
```

The issue is with `n183n`. During the conversion the node is recreated and the flags from the original node are copied into the new node. This is done without complete verification that the flags will be valid for the new node. For example, in the case above, the `unneededConv` and `simpleDivCheck` flag are overloaded. Since we didn't clear this flag during conversion, we end up setting it incorrectly (and unknowingly) for the new node. This leads to invalid instructions being generated during code generation.

dchopra001:
The problem occurs because during the aforementioned transformation, the node is recreated here:
[...]

Issue #2254, opened 2018-01-23
Enable automatic pull request copyright checks


0xdaryl:
Re-purpose the copyright checker tool from Eclipse OpenJ9 as part of the Eclipse OMR delivery process.  This will be far more convenient and less error-prone than manually inspecting each file on delivery.

Eclipse OpenJ9 PR that introduced the copyright checker:  https://github.com/eclipse/openj9/commit/ac8ac21a333a863cc928928fd7168e6763351e3b

Some input from @charliegracie specific to Eclipse OMR:

> I think the copyright checker needs some improvements though.  There is a lot of third party files in OMR that do not meet the copyright check.  If we modify that code I do not want the check to fail.  The script needs to be updated to whitelist the thirdparty folder or something similar.

0xdaryl:
@nbhuiyan

AdamBrousseau:
Are we open to the option of pulling the OpenJ9 Jenkins File directly and making any changes/improvements there rather than committing a ~duplicate script to OMR and having to maintain multiple copies? This is what I did for the OpenJ9 extensions repos.
We could likely come up with a process to pass a whitelist to the script rather than hardcoding multiple whitelists in the same file for each repo.

charliegracie:
I am not sure that is the right answer.  That puts a circular dependency on the projects that does not exist right now.  Currently OpenJ9 depends on OMR and not the other way around.  I think we should think about it a bit more before we make that decision.

AdamBrousseau:
I agree it is not the perfect solution, but neither is code duplication. I don't really want to suggest a "third party tools repo" but I suppose that is an option. It might be a case of the lesser of ~two~ three evils?

0xdaryl:
I agree with @charliegracie that OMR depending on a downstream project isn't the right answer.  Downstream projects already have a dependence on the upstream OMR.  Perhaps creating a separate tools repo under the Eclipse OMR project for shared infrastructure for runtime projects is the way to go.

mstoodle:
Could OMR be the source and OpenJ9 consumes it from OMR?

xliang6:
> We could likely come up with a process to pass a whitelist to the script rather than hardcoding multiple whitelists in the same file for each repo.

Can we generalize the script to take in repo-specific information like a whitelist filename? If so, this generalized script can reside in OMR. We can then create smaller scripts to call this script with the right filenames for OMR and OpenJ9, or maybe we can achieve this with Jenkins?

[...]

PR #2300, opened 2018-02-08
Deprecate nodeIsDebug node flag

fjeremic:
A quick search through the codebase reveals this flag is not actually
used by anyone so it has no reason to exist.

Fixes: #2299

### DIFF
--- a/compiler/il/OMRNode.hpp
+++ b/compiler/il/OMRNode.hpp
@@ -1011,10 +1011,6 @@ public:
    bool divisionCannotOverflow();
    bool isNonDegenerateArrayCopy();
 
-   // Should not be counted for metrics
-   bool isDebug();
-   void setIsDebug(bool v);
-
    // Flag used by arithmetic int/long operations
    bool cannotOverflow();
    void setCannotOverflow(bool v);
@@ -1812,7 +1808,6 @@ protected:
       visitedForHints                       = 0x00000080, ///< Used only during codegen phase
       nodeIsNonNegative                     = 0x00000100,
       nodeIsNonPositive                     = 0x00000200,
-      nodeIsDebug                           = 0x00000400, ///< Should not be counted for metrics
 
       //---------------------------------------- node specific flags---------------------------------------
 
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -5616,22 +5616,6 @@ OMR::Node::divisionCannotOverflow()
    return dividendIsNotMin || divisorIsNotNegativeOne;
    }
 
-
-
-bool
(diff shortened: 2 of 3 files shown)


compiler/il/OMRNode.hpp
```diff
@@ -1812,7 +1808,6 @@ class OMR_EXTENSIBLE Node
       visitedForHints                       = 0x00000080, ///< Used only during codegen phase
       nodeIsNonNegative                     = 0x00000100,
       nodeIsNonPositive                     = 0x00000200,
-      nodeIsDebug                           = 0x00000400, ///< Should not be counted for metrics
```
0xdaryl: Do you want to announce this flag's new availability by not deleting it outright but adding an availability comment instead?  e.g.,

```
// Available        =0x00000400,
```

fjeremic: I do not think this is a good idea. We have the starts of a discussion starting in https://github.com/eclipse/omr/issues/2297 which deal with problems of overloading flag values and getting unexpected behavior. One has to think and investigate very hard before adding a new flag overload at the moment. We should be looking for alternative solutions to this broad problem and encouraging more overloading of flag values is in my opinion dangerous.

I would prefer for us to start having less overloading moving forward until we setting on a more general solution to this not so trivial problem at hand. Do you agree with this direction?

fjeremic:
Travis failure is due to the infamous ThreadAbortTest.

compiler/il/OMRNode.hpp
0xdaryl: I absolutely fully support an overall effort to reduce node flag overloading.  I've been been burned by this confusion in years past.   In this case, I was fooled by the comment preceding this enum which puts `0x00000400` in the "general node flags" category which I incorrectly assumed it meant it couldn't/shouldn't be overloaded, but that's clearly not true.  Your code is fine as is.

0xdaryl:
@genie-omr build all

fjeremic:
> Sorry for catching this late. One small change: please update the copyright header on ras/DebugCounter.cpp and then it's good to go.

Right! I thought about this afterwards. Do we have an issue open to add a GitHub hook for a copyright date check? If so could you link it here? Rebasing and updating my commit now.

0xdaryl:
Yes, #2254.  Although @nbhuiyan owns the OMR issue, we're waiting on @adambrousseau to complete some necessary file whitelisting on the OpenJ9 copyright checker tools that we'll need for OMR.  Let's move any copyright checker discussion to #2254.

0xdaryl:
A previous genie build succeeded, and only a comment change was pushed since then.  Hence, skipping a repeat of the genie tests.

PR #2301, opened 2018-02-08
API to support a mutex and a condition variable

babsingh:
**Motivation:** OpenJ9 utilizes OMR. OMR has a thread library. Currently,
macros to utilize a mutex and a condition variable can be accessed by
including `thrdsup.h` in the parent project. Including `thrdsup.h` leads
to compilation errors in OpenJ9 on AIX and Windows. So, an API has been
developed to support a mutex and a condition variable without exposing
OMR's internal dependencies (`thrdsup.h`).

Two structs have been added, **OMRMutex** and **OMRCondVar** which are wrappers
for **J9OSMutex** and **J9OSCond** respectively. The definition of these structs
is only visible to functions in `omrthread.c`. Isolating definition of
these structs in `omrthread.c` prevents exposing large system header
files such as `windows.h` outside of OMR.

The API includes the following functions to utilize a mutex and a condition variable:
1) `omrthread_mutex_create`
2) `omrthread_mutex_destroy`
3) `omrthread_condvar_create`
4) `omrthread_condvar_destroy`
5) `omrthread_mutex_enter`
6) `omrthread_mutex_exit`
7) `omrthread_condvar_notify`
8) `omrthread_condvar_notify_all`
9) `omrthread_condvar_wait`

### DIFF
--- a/include_core/omrthread.h
+++ b/include_core/omrthread.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -47,6 +47,9 @@ typedef intptr_t omrthread_os_errno_t;
 typedef struct J9Thread *omrthread_t;
 typedef struct J9ThreadMonitor *omrthread_monitor_t;
 typedef struct J9Semaphore *j9sem_t;
+
+typedef struct OMRMutex *omrmutex_t;
+typedef struct OMRCondVar *omrcondvar_t;
 
 #include "omrthread_generated.h"
 
--- a/include_core/thread_api.h
+++ b/include_core/thread_api.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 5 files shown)


babsingh:
I am trying to expose `J9OSMutex` and `J9OSCond` to OpenJ9. Currently, I accomplish this by including `thrdsup.h` directly in OpenJ9. This works fine on Linux. But, it causes compilation issues on Windows and AIX.

@keithc-ca Can you provide some feedback/guidance on achieving this?

babsingh:
@0xdaryl @charliegracie @fjeremic These changes can be reviewed.

charliegracie:
Can you explain the motivation for this change?  I understand you want access to the mutex and condVar but the whole point of the thread library is to hide these details from the consumer (specifically for Windows and AIX).

babsingh:
Required for https://github.com/eclipse/openj9/pull/1079.

JVM_*Signal functions require a signal handler which can execute Java code. JLI Launch thread has no awareness of the OMR environment. When the signal handler is invoked from the JLI Launch thread, then omrthread_monitor_enter/exit functions start to segfault since MACRO_SELF() returns NULL. So, we need to access a mutex and a condition variable directly without relying upon monitor code in order to make the signal handler work in Java.

charliegracie:
Doesn't that mean the functions called by the JLI thread should just be attaching/detaching themselves properly to the thread library? The attach/detach calls will be *required* anyway if you want to be able to use the thread library APIs you are trying to create.

babsingh:
Not sure if improper attach/detach is causing the earlier seg faults in the JLI thread.

The new API works fine in the JLI thread.

charliegracie:
The reason the new APIs work is that your downstream changes do the allocation/free’ing on attached threads and the OS specific macros used by the functions called by the JLI thread do not require SELF. Would you be able to try the attach/detach to see if they work?  I am not comfortable exposing these APIs in the thread library.

babsingh:
I didn't find any references to JLI thread in OpenJ9. I think JLI thread is OpenJDK code [@pshipton confirmed this]. So, we won't be able to attach/detach JLI thread to the thread library. Also, Java applications, such as DB2, have their own native threads. We won't have control over such native application threads. Without this API, the Java signal handler won't work if a signal is received in a native application thread that is not attached to the thread library.

charliegracie:
The JLI thread is clearly calling a function in OpenJ9 code correct? The code which does the omrthread_monitor_enter / _exit needs to have an omthread_attach at the beginning and an omrrhread_detach at the end / all exits. Doing this I believe will resolve the original issues and not require the API changes. 

@NOTE
Calls to omrthread_attach and omrthread_detach stack to it is fine to call it if the thread is already attached.

pshipton:
The JLI thread is not calling a function in OpenJ9 code. The JLI thread can run a signal handler, as can any thread. The signal handler needs a way to notify another thread that the signal has occurred. Babneet wants to add these APIs as the mechanism for notifying the other thread. Not sure if the signal handler could call omrthread_attach() / detach(). If that works it could be the solution.

charliegracie:
I am assuming the function that is causing the issue is this one?

https://github.com/eclipse/openj9/pull/1079/files#diff-abe0cd1587b6a77eb44fb6920607a2c6R79

I assume that function had an omthread_monitor_enter/exit in the initial implementation?

If I am incorrect in either question can you please point me to the code that is causing the issue?

pshipton:
I believe its
https://github.com/eclipse/openj9/pull/1079/files#diff-abe0cd1587b6a77eb44fb6920607a2c6R223
as called from the signal handler
https://github.com/eclipse/openj9/pull/1079/files#diff-fed8775563862a9f1c52335a4fc98f81R4525

charliegracie:
Then it is in fact OpenJ9 code that is causing the issue.  The code called by the function handler needs to just attach/detach to the thread library right before the call to monitor_enter and right after the call to monitor_exit

babsingh:
Calling attach/detach functions, everytime a signal is received, will be very expensive; almost undesirable. I have been trying to avoid pthread lib usage as much as possible since blocking operations are not allowed in the signal handler. Taking this approach will increase blocking operations in the signal handler. Since a real semaphore is not available on zOS, I chose to use a mutex and condvar to notify the reporter thread. Once we have a proper semaphore implementation on zOS (https://www.ibm.com/developerworks/library/l-semaphore/index.html), we can move away from using a mutex and condvar in the signal handler. OMR already exposes a semaphore and its API. Why can't we similarly expose a mutex and a condvar? Directly having access to a mutex and a condvar is more footprint friendly, and yields better performance in some cases than using a monitor.

charliegracie:
The cost of multiple function calls deep to get to the signal handler is going to be overwhelmed by 2 more calls at the point you are going to notify another thread to go execute some Java code?  Sorry I just find that difficult to believe.  Have you tried the solution of thread attach/detach and using the monitors directly?  If you try the solution I suggested and show that it does not work or it causes a significant overhead compared to the current proposal we can work on another solution.  The current solution is not correct in terms of following the API contracts laid out by the thread library.  

The thread library is a wrapper for providing OS agnostic support for threads, monitors, etc. to runtimes.  These APIs should be applicable to most if not all platforms supported by the thread library. ALL thread library APIs require a thread to be attached.  Any new API that gets added can not / should not break this contract.  

The OMR project is getting close to a first release where we will be documenting and creating a stable API for components like the thread library.  We have to be very careful about what APIs get added as they will need to be supported for a long time.

babsingh:
I added the following code in the signal handler:
```
omrthread_attach(NULL);
...
omrthread_detach(NULL);
```
https://github.com/babsingh/openj9/commit/4c0616b74dd6c5fd1726d630c9aa783a425a07d7: It works.

charliegracie:
Have you been able to run testing / performance runs to see if it is ok?

pshipton:
Its not ok because the signal handler can't call apis that can block. What I think we should do is propose a new OMR signal API that OpenJ9 can use, since the current APIs are not sufficient.

charliegracie:
@babsingh Can we close this PR as I believe the design has changed and other PRs will be handling this?

PR #2302, opened 2018-02-09
Scan RS concurrently

amicic:
Move the Remembered Set (as roots) scan, partially to the concurrent
phase of Concurrent Scavenger. 

The RS in Scavenger for the most part remembers object Tenure to Nursery
references (direct ones). But it can also remember off-heap structures
that have Nursery references, where the off-heap structure has a proxy
heap object in Tenure (indirect ones). An example being a class object
in Tenure and a matching off-heap class structure in upstream OpenJ9
project.

This change will break the RS scan into two passes. The one where we
scan Indirect references will stay in the STW phase where the rest of
the roots are processed. The pass where scan Direct references is moved
to the concurrent phase. 

Iteration (and popping-up) of RS puddles should still be safe operation,
even though new puddles can be added while RS is still being iterated.
The popping occurs against the previous list, while all the pushing
(whether the ones that have just been processed or the newly added ones
) occurs to the current list. 

Also, there are no races between readers/writters on RS slots - mutator
or other GC scanning threads can only add new slots - never modify the
existing ones.

Moving Indirect pass to concurrent phase is also a possibly, but that
would require that VM employs read barrier of those indirect reference
slots, which ATM is still not the case for the only project using CS -
OpenJ9.

Also doing some minor naming cleanup with RS related constants.

### DIFF
--- a/gc/base/GCExtensionsBase.cpp
+++ b/gc/base/GCExtensionsBase.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -93,7 +93,7 @@ MM_GCExtensionsBase::initialize(MM_EnvironmentBase* env)
 	if (!rememberedSet.initialize(env, OMR::GC::AllocationCategory::REMEMBERED_SET)) {
 		goto failed;
 	}
-	rememberedSet.setGrowSize(J9_SCV_REMSET_SIZE);
+	rememberedSet.setGrowSize(OMR_SCV_REMSET_SIZE);
 #endif /* OMR_GC_MODRON_SCAVENGER */
 
 #if defined(J9MODRON_USE_CUSTOM_SPINLOCKS)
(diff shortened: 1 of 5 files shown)


amicic:
@fjeremic Just to update the year or something else?

fjeremic:
> @fjeremic Just to update the year or something else?

Just the year, nothing else.

fjeremic:
Thanks for the updates. Copyright checks are fine now. I'll delegate to @charliegracie or @dmitripivkine for the semantics review of this PR. Cheers!

fjeremic:
> This change should not be merged before eclipse/openj9#1145 is approved (for the sake of OpenJ9).

Could you please mark this with a "WIP: " prefix. The committers will not be able to accidentally or intentionally merge this until you have removed it. You can remove the "WIP: " after any upstream dependencies are merged or approved such as the one you mentioned.

amicic:
In general it sounds overkill - it could unnecessarily delay review and discussion on this PR. Perhaps we come up with some other label to signify this dependency (makes sure that the approver is aware of upstream project dependency)? What I normally do make the approver same for both PRs and let know via comments about the dependency.

fjeremic:
> In general it sounds overkill - it could unnecessarily delay review and discussion on this PR. Perhaps we come up with some other label to signify this dependency (makes sure that the approver is aware of upstream project dependency)? What I normally do make the approver same for both PRs and let know via comments about the dependency.

I agree, our current handling of dependencies in OMR may need some work. I've opened up a separate discussion on this in #2303 to avoid further polluting this PR.

Issue #2303, opened 2018-02-09
How to best handle dependent changes from downstream consumers


fjeremic:
As part of https://github.com/eclipse/omr/pull/2302#issuecomment-364489477 the contributor has signaled to the committers that they want the PR to be eligible for merging only after ~an upstream~ downstream dependency has been approved (pending OMR review of course). Our current process for dealing with this is to request the PR be marked "WIP: " (Work In Progress) until a time where the contributor believes it is ready for final reviews and merging.

In https://github.com/eclipse/omr/pull/2302#issuecomment-364492670 the contributor identified this may be a cause of churn as it could delay reviews of the OMR PR. In my opinion I don't think we should liberally be taking advantage that some of the OMR committers are also committers on the ~upstream~ downstream project as that creates coupling and bias towards a particular consumer of OMR. Is there a better way to handle these situations?

dmitripivkine:
In my opinion OpenJ9 is not a upstream but downstream project. The change itself would not break OMR but OpenJ9 might be broken. This fact of such dependence has been reflected in PR and committer has to be noticed. WIP however might be used for dependent OpenJ9 item.
Also I believe it is a good practice to submit both PRs at the same time to minimize time dependent project is broken.

0xdaryl:
I am of the opinion that the contributor of a change that may involve multiple projects should assume the bulk of the responsibility for shepherding it into the codebase(s).  Yes, an OMR committer ultimately has to push the OMR button after ensuring adequate reviews are complete and that the contributor guidelines are being followed, but staging a delivery across OMR and a downstream project should really be orchestrated by the contributor in my opinion.  This starts with the code itself and ensuring it is written in a staged manner that doesn't rely on coordination.  I'll concede there are situations where this isn't possible, but there have been examples in the past where it *was* possible but not done for whatever reason leading to downstream churn.

In my mind, the WIP scheme is simple and does the job we want it to do.  Namely, it allows the contributor of a PR to hold back the committers from merging until they have made all downstream preparations (e.g., ensuring a dependent PR is merged) or for any other reason for that matter.  A contributor can (and should) use the comment section of their own PR to explain their unique circumstances for merging (e.g., requesting a review even though it's marked WIP, informing a committer of their desire to coordinate when the PR is merged, etc.), and all OMR committers are willing to work with contributors and have demonstrated they are responsive to their requests.  Communication is key.  There are numerous examples of this over OMR's two year history.

[...]

PR #2306, opened 2018-02-12
Add downstream dependency section

fjeremic:
Describe the process of handling dependent changes in downstream
projects and the coordination expected from contributor and committer.

Fixes: #2303

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,6 +138,18 @@ Following these guidelines will help us to merge your pull requests smoothly:
    create a better end product. Committers will wait until after you've removed
    the WIP prefix to merge your changes.
 
+### Downstream Dependencies
+
+If a downstream project consuming OMR has a dependency on the contents of your pull 
+request, i.e. the pull request should not be merged until all downstream preparations 
+have been complete, you as the contributor of the pull request may use the WIP prefix in 
+addition to comment in the said pull request to inform the committer of your 
+particular circumstances for merging. That is, you may request a review of the pull
+request even though it's marked WIP, informing the committer of your desire to
+coordinate when the PR is merged. Once all downstream preparations have been complete
+it is your responsibility as the contributor to remove the WIP prefix and inform the
+committers that the pull request is ready for merging, pending reviews have been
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -138,6 +138,18 @@ Following these guidelines will help us to merge your pull requests smoothly:
    create a better end product. Committers will wait until after you've removed
    the WIP prefix to merge your changes.
 
+### Downstream Dependencies
+
+If the contents of your pull request has a dependency on a downstream project, i.e.
```
charliegracie: I do not believe OMR changes should have dependencies on downstream projects as it should be the other way around. If a change is being contributed to OMR the contributor may ask the OMR committers to wait for their downstream project to be ready but the OMR committers should also make other downstream projects aware of the changes that are coming so they can also be ready.

Dependencies and the requirements on downstream projects as I see them:
1. A new feature/API is added to OMR and it is completely self contained.  No work required by downstream projects to consume and take advantage of this feature.
2. A new feature/API is added to OMR but requires downstream projects to configure or provide support for it to be useful. This *should* not break downstream projects but until they are ready to provide the required extra bits they can not take advantage of the new feature.
3. An API change/removal which will break all downstream projects using the component/feature containing the change.  This will require downstream projects to modify their code to move to the latest version of OMR.  This *will* break consumers until they have updated.  This is where co-ordination of the OMR project committers/leads is required to ensure that all downstream projects are aware of the upcoming changes and facilitate, where possible, to help the downstream consumers update.  These types of changes will require a lot more oversight as there must be a very good reason for this.  In a lot of cases the `old` API and the `new` API can live side by side for a release to ensure downstream projects have enough time to update.

fjeremic: Fixed in https://github.com/eclipse/omr/pull/2306/commits/b003075004f5b75061a6073c3671f9a7730eda60

0xdaryl:
@fjeremic : A reminder to send an email to the omr-dev mailing list informing the community that the CG has changed with a brief description of the changes.  Thanks.

Issue #378, opened 2016-10-25
DefineFunction() doesn't accept the string literal in C++11


ShuyuLi-ca:
Some of the parameter types in DefineFunction() are char* but arguments are not to be modified. Started in C++11, DefineFunction() doesn't accept the string literal.

The parameter types should be changed in order to accommodate the string literal.

mstoodle:
This issue was addressed by PR #373 but the issue wasn't mentioned in the commit message.

Closing manually.

Thanks for the fix, @ShuyuLi-ca !

PR #2307, opened 2018-02-12
Use GNU debug symbols on PPC platforms

keithc-ca:
This is required for DDR support: #378 and makes ppc64le consistent with x86-64.


charliegracie:
Does this work if you compile PPC LE with `xlc` instead of GCC?

keithc-ca:
The change is in a section guarded by
```
ifeq (gcc,$(OMR_TOOLCHAIN))
```
so it should not apply for xlc.

charliegracie:
Thanks!  I missed that in th diff.

charliegracie:

Issue #2113, opened 2017-12-06
Occasional hangs in omrthreadtest on linux


charliegracie:
Occasionally the omrthreadtest hangs when running merge or pull request builds.  The test always hangs in the ThreadAbortTest.

Today I was able to log into the machine and get a backtrace for the process while it was hung. There was only 1 thread in the system. Here is the backtrace.
```
#0  0x000003ffaac8d7d4 in __pthread_cond_wait (cond=0x2aa31849ea8, mutex=0x2aa3184e4d8) at pthread_cond_wait.c:186
#1  0x000002aa30504de8 in monitor_wait_original (interruptible=0, nanos=0, millis=0, monitor=0x2aa31844be8, self=0x2aa31847d80) at ./common/omrthread.c:4435
#2  monitor_wait (interruptible=0, nanos=0, millis=0, monitor=0x2aa31844be8) at ./common/omrthread.c:4290
#3  omrthread_monitor_wait (monitor=0x2aa31844be8) at ./common/omrthread.c:4160
#4  0x000002aa304a31d4 in ThreadAbortTest_Waiting_Test::TestBody (this=<optimized out>) at abortTest.cpp:207
#5  0x000002aa304fc3fe in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (location=0x2aa305439e2 "the test body", method=<optimized out>, 
    object=<optimized out>) at ../../third_party/gtest-1.8.0/src/gtest.cc:2402
#6  testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=0x2aa3184d3c0, 
    method=(void (testing::Test::*)(testing::Test * const)) 0x2aa304a3108 <ThreadAbortTest_Waiting_Test::TestBody()>, location=0x2aa305439e2 "the test body")
    at ../../third_party/gtest-1.8.0/src/gtest.cc:2438
```

```
(gdb) info threads
  Id   Target Id         Frame 
* 1    Thread 0x3ffaae72720 (LWP 63510) "omrthreadtest" 0x000002aa304a31d4 in ThreadAbortTest_Waiting_Test::TestBody (this=<optimized out>) at abortTest.cpp:207
```

charliegracie:
One of the travis jobs for the #2115 just had the same hang. Sadly we can not get any diagnostic information from the failing travis build. Interestingly enough though none of the new asserts triggered so that gives some information.

charliegracie:
[...]

PR #2310, opened 2018-02-12
Improve debugging code for Issue #2113

charliegracie:
This issue has happened a few more times in the CI testing so I
am able to improve the debugging code by enabling more asserts.
Also I removed some code from the test which should never be
executed. I added another assert to make sure the expected code
path would always happen.


charliegracie:

PR #2311, opened 2018-02-12
Filter duplicate information from macroList

keithc-ca:
Without this filtering, macroList for openj9 would be nearly 400MB; with it, it is about 250kB.


charliegracie:
@genie-omr build all

PR #2312, opened 2018-02-13
Enable 3 tier locking in CMake builds

charliegracie:
The non CMake testing had 3 tier locking enabled so the CMake
testing should also test it.  To enable 3 tier locking the JLM
flag also has to be enabled.


charliegracie:
@genie-omr build all

PR #2313, opened 2018-02-13
Fix checks when printing RS Instructions

dchopra001:
The getSourceImmediate() API is being used incorrectly inside `TR_Debug::print(TR::FILE*,TR::S390RSInstruction*)`. The purpose of the query here is to check if _sourceImmediate is set. However, if it is set to 0, then the query evaluates as false. This can cause errors and crashes due to printing. This commit adds a new query to check if `_sourceImmediate` has been set, and cleans up the print function to correctly use the `hasMask()` query as well.

### DIFF
--- a/compiler/z/codegen/S390Instruction.hpp
+++ b/compiler/z/codegen/S390Instruction.hpp
@@ -2550,9 +2550,11 @@ class S390RSInstruction : public TR::S390RegInstruction
 
    int16_t                 _sourceImmediate;
    int8_t                  _maskImmediate;
-   /// _hasMaskImmediate evaluates to true if a value for the _maskImmediate field has been specified, false otherwise
+   /// Set to true if a mask value was specified during the construction of this object, indicating an RS instruction of RS-b format 
+   /// where the third operand is a mask
    bool                    _hasMaskImmediate;
-   /// _hasSourceImmediate evaluates to true if a value for the _sourceImmediate field has been specified, false otherwise
+   /// Set to true if a mask value was specified during the construction of this object, indicating an RS instruction of RS-b format 
+   /// where the third operand is a mask
    bool                    _hasSourceImmediate;
    int8_t                  _idx;
    Kind                    _kind;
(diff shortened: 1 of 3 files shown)


compiler/z/codegen/S390Instruction.hpp
```diff
@@ -2551,6 +2551,7 @@ class S390RSInstruction : public TR::S390RegInstruction
    int16_t                 _sourceImmediate;
    int8_t                  _maskImmediate;
    bool                    _hasMask;
+   bool                    _hasImmediate;
```
fjeremic: Would it be possible to add Doxygen documentation for the newly added field and API?

fjeremic:
@dchopra001 something seems off with your commit authorship. You may want to check that your GitHub profile matches the email you specified in your commit. You may wish to add alternative emails in your profile.

dchopra001:
@fjeremic I've added documentation (4769ec2) and done some renaming (c5cb3a6). PTAL.

compiler/z/codegen/S390Instruction.hpp
```diff
@@ -2550,7 +2550,10 @@ class S390RSInstruction : public TR::S390RegInstruction
 
    int16_t                 _sourceImmediate;
    int8_t                  _maskImmediate;
-   bool                    _hasMask;
+   /// _hasMaskImmediate evaluates to true if a value for the _maskImmediate field has been specified, false otherwise
+   bool                    _hasMaskImmediate;
```
fjeremic: I don't think this is a very descriptive comment. The name of the field already implies relation to `_maskImmediate`. The comment doesn't give any indication as to _why_ we have the two `_sourceImmediate` and `_maskImmediate` fields. Perhaps a better comment may describe why we make the distinction:

```
/// Determines whether a mask value was specified during the construction of this object, indicating an RS instruction of RS-b format where the third operand is a mask
bool _hasMaskImmediate;

/// Determines whether an immediate value was specified during the construction of this object, indicating an RS instruction of RS-a format where the third operand is a memory reference interpreted as an immediate value.
bool _hasSourceImmediate;

dchopra001: Updated here: 387883a

fjeremic:
@genie-omr build all

PR #2319, opened 2018-02-15
[CMake] Initial cmake script for building OMR static lib

amitin:
This is initial implementation for building static OMR lib.
Known issues:
1. Doesn't include static glue 
2. Tested on Windows, but *nix version should work too

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,7 @@ add_subdirectory(omrtrace)
 add_subdirectory(omr)
 add_subdirectory(third_party)
 add_subdirectory(omrsigcompat)
+add_subdirectory(omr_static_lib)
 
 if(OMR_GC)
 	add_subdirectory(gc)
--- a/omr_static_lib/CMakeLists.txt
+++ b/omr_static_lib/CMakeLists.txt
@@ -1,0 +1,67 @@
+###############################################################################
+# Copyright (c) 2017, 2018 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+# or the Apache License, Version 2.0 which accompanies this distribution and
+# is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# This Source Code may also be made available under the following
+# Secondary Licenses when the conditions for such availability set
+# forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+# General Public License, version 2 with the GNU Classpath
+# Exception [1] and GNU General Public License, version 2 with the
+# OpenJDK Assembly Exception [2].
+#
+# [1] https://www.gnu.org/software/classpath/license.html
+# [2] http://openjdk.java.net/legal/assembly-exception.html
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+###############################################################################
+
+set(OMR_STATIC_LIBNAME omrstatic)
+
+set(OMR_LIBS 
+	j9avl 
+	j9hashtable
+	j9hookstatic
+	j9pool
+	omrcore
+	omrtrace
+	omrutil
+)
+
+if(OMR_PORT)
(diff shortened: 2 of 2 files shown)


charliegracie:
@rwy0717 and @dnakamura can you review please?

rwy0717:
Hi @amitin ! Thanks for your patch :)

I have some ideas on how we can improve the patch, but before we get into it, what are you using this for? The giant static library was something we did in the old build system, but it was [a little crazy](https://github.com/eclipse/omr/blob/b9d1250dea50a28faf03024394ff36ed99c4c2d1/omr_static_lib/makefile#L70). Maybe there is another way we can tackle this.

Is the point that, this will make it easier to link with OMR in non-cmake buildsystems?

amitin:
@rwy0717, Well, I didn't know that the huge static lib is an obsolete way. :)
We use cmake to build our project with all third-party libs that we use. Some of the third-party libs (smaller ones) are included into 'all' target, but bigger (like OMR) are cmake '[external projects](https://cmake.org/cmake/help/latest/module/ExternalProject.html)'. This way requires to have some knowledge about which libs are required for, say, 'port' part of OMR, and these libs locations as well as includes. The ideal is what LLVM provides: a special '[llvm-config](http://llvm.org/docs/CommandGuide/llvm-config.html)' tool for aswering to these queries.
So as a workaround I added a single static lib - because I could see the same done for autoconf. :)

Unfortunately, using non-cmake build system is not what we want right now. Our main target platform is Windows x86/x86_64 and the most easy and convenient way to build our project is cmake.

dnakamura:
The problem still exists of getting proper include paths and compile flags  needed for linking properly. 
I am not opposed to the concept of this change, although I would like to see omrstatic excluded from all.  Partly because this approach will probably break on at least one of the oddball platforms we support. The other option would be wrapping `add_subdirectory(omr_static_lib)` with a config flag, although that has the downside of needing to reconfigure if you change your mind.

amitin:
Alternatively, we could use 'install' target with given CMAKE_INSTALL_PREFIX which provides known lib/includes structure. This could have all possible standalone OMR parts built separately as external project, and OMR glue - as part of our project.

charliegracie:
@dnakamura more thoughts?  I would like to move this forward.

dnakamura:
@charliegracie I think long term the right answer to the problem is figuring out a proper solution for exporting the required information. IN the mean time i would be ok merging this if `omrstatic` was no longer part of `all`

rwy0717:
@amitin what components of OMR are you using? IIRC it's primarily the jitbuilder library? It might be possible to find a simplified way of consuming the jitbuilder library.

Alternatively, you could add OMR as a subdirectory to your project, which would make it _much_ easier to work with.

amitin:
@rwy0717, it's port library and gc. Compiler/JitBuilder probably will be in use too, but not so soon. :)
Thank you for your suggestion, I'll give it a try.

CMakeLists.txt
```diff
@@ -132,6 +132,7 @@ add_subdirectory(omrtrace)
 add_subdirectory(omr)
 add_subdirectory(third_party)
 add_subdirectory(omrsigcompat)
+add_subdirectory(omr_static_lib)
```
charliegracie: Should this be happening as one of the last directories to add since it relies on the gc directory, etc?  I apologize for my lack of CMake knowledge in this area.  It is very possible CMake just builds up the dependencies and things will happen in the right order.

rwy0717:
Targets and dependencies can be defined in any order. However, variables, functions, and macros must be defined before used. Since this subdirectory only uses variables defined by OmrPlatform, it is order independent (which is how we've try to keep things).

youngar:
Hi @amitin, I was wondering if you still had a use-case for building a single static OMR lib.  It should not be hard to use OMR from another cmake project (especially as a subdirectory), but I know there is more we can do to support non-cmake build systems.

I really like the idea of creating something like `llvm-config`.  I have investigated  creating `pkgconfig` files for each library in OMR, but it seemed to be quite a bit of work.  I don't think that it would be hard to generate a single `pkgconfig` for all of OMR (which would link all libraries).

If this is no longer needed I would like to close the pull request.

If I can help in any way with integrating OMR with your build system, please let me know.

amitin:
@youngar, the goal I tried to achieve is to build OMR as an 'external project' using cmake. This cmake feature allows to fetch, patch and build some third-party library without including the source code of this library into the main project source tree. I think that now we could do it without building a static lib, so please close this pull request.

PR #2322, opened 2018-02-16
Fix field offset discovery

keithc-ca:
* rely on DWARF offsets which consider alignment requirements
* correct offsets of synthetic fields
* remove unused Field._sizeOf

Issue: eclipse/openj9#378


charliegracie:
@genie-omr build all

PR #2324, opened 2018-02-19
Apply MFENCE and CLFLUSH in patchFence16 on X86-32

### LINKED ISSUE #2323: Patching Fence missing memory fences and cache line flushes on X86-32
From https://github.com/eclipse/omr/blob/b9d1250dea50a28faf03024394ff36ed99c4c2d1/compiler/x/runtime/X86Runtime.hpp#L112
```
inline void patchingFence16(void* addr)
   {
#ifdef TR_HOST_64BIT
   _mm_mfence();
   _mm_clflush(addr);
   _mm_clflush(static_cast<char*>(addr)+8);
   _mm_mfence();
#endif
}
```
patchingFence16() only applies memory fences and cache line flashes on X86-64. However, it seems such memory fences and cache line flushes are also required; otherwise, the partially patched instruction may reach execution and causes crashes.

After researching the history of this function, it has been no-op for X86-32 since day one.
https://github.com/eclipse/omr/blob/0ab05844ed6b2afa165bed1b9e95b240f0eea846/compiler/x/i386/runtime/IA32AsmUtil.inc#L97
```
        align   16
_patchingFence16:             ; Entry point for C code
        ret
```

The intrinsic-based implementation is same as the old assembly implementation.

0dvictor:
Memory fences and cache line flushes are missing in
patchFence16() on X86-32. Fixing it.

Closes #2323

### DIFF
--- a/compiler/x/runtime/X86Runtime.hpp
+++ b/compiler/x/runtime/X86Runtime.hpp
@@ -30,9 +30,7 @@
 #define cpuidex(CPUInfo, EAXValue, ECXValue) __cpuidex(CPUInfo, EAXValue, ECXValue)
 #else
 #include <cpuid.h>
-#ifdef TR_HOST_64BIT
 #include <emmintrin.h>
-#endif
 #define cpuid(CPUInfo, EAXValue)             __cpuid(EAXValue, CPUInfo[0], CPUInfo[1], CPUInfo[2], CPUInfo[3])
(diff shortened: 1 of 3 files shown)


0xdaryl:
Are the changes to the build files related to the 32-bit fence fix?  If not, please move them to a separate commit.

0dvictor:
They are related since MFENCE and CLFLUSH instructions are introduced by SSE2, the makefiles need to be updated so that SSE2 is enabled.

0xdaryl:
@genie-omr build plinux zlinux

0dvictor:
Good catch. Both copyright headers are updated. Thank you.

PR #2115, opened 2017-12-06
Add debug code and asserts

charliegracie:
This code adds extra asserts and debugging to help track down an
occasional hang on Linux 390.

Issue: #2113

Signed-off-by: Charlie Gracie <charlie.gracie@gmail.com>

PR #2326, opened 2018-02-20
Fix intermittent TraceLogTest.stressTraceBufferManagement failure

jason-hall:
Fixes issue #2115. Trace points wrapping into the next record by 1 character were skipped by the trace point iterator. It was failing about 1 in 3 times on z/os and now passed 300 consecutive runs.


0xdaryl:
@genie-omr build all

PR #2331, opened 2018-02-21
Fix use of Macro::getNumeric as test

keithc-ca:
Return value should not depend on whether 'ret' is null; otherwise we get this failure:
```
Error: java/genBinaryBlob.cpp:770 addBlobConst - Adding more constants than enumerated
```


charliegracie:
@genie-omr build all

PR #2334, opened 2018-02-22
Fixing CR instruction handling on PPC

alenbadel:
Prior, instructions that required CR registers were not correctly processed. An offset is needed to handle CR registers. This commit includes enabling handling for 1 Target, 2 Source Register Immediate instructions to be correctly formulated.

The implementation of the function needs to be moved from the corresponding .hpp file to the .cpp. This was necessary to determine that the register used to construct the instruction is a CR.

Signed-off-by: Alen Badel <Alen.Badel@ibm.com>

### DIFF
--- a/compiler/p/codegen/OMRRealRegister.cpp
+++ b/compiler/p/codegen/OMRRealRegister.cpp
@@ -186,7 +186,14 @@ OMR::Power::RealRegister::setRegisterFieldRA(uint32_t *instruction)
    else
       *instruction |= fullRegBinaryEncodings[_registerNumber] << pos_RA;
    }
-
+void
+OMR::Power::RealRegister::setRegisterFieldRB(uint32_t *instruction)
+  {
+      if (self()->isConditionRegister())
+        *instruction |= fullRegBinaryEncodings[_registerNumber] << (pos_RB + 2);
+      else
+         *instruction |= fullRegBinaryEncodings[_registerNumber] << pos_RB;
+  }
 void
 OMR::Power::RealRegister::setRegisterFieldXS(uint32_t *instruction)
    {
--- a/compiler/p/codegen/OMRRealRegister.hpp
+++ b/compiler/p/codegen/OMRRealRegister.hpp
@@ -110,10 +110,7 @@ class OMR_EXTENSIBLE RealRegister : public OMR::RealRegister
 
    void setRegisterFieldRA(uint32_t *instruction);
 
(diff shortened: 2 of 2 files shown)


zl-wang:
To describe the issue more clearly:
    It is certainly not that instructions requiring CR registers cannot be handled correctly,  since we have done that correctly for the past dozens of years.  The real issue is that no instruction (we generated in the past) required CR in RB register field such that incorrect CR encoding in that field hasn't been caught.

zl-wang:
jenkins test plinux

zl-wang:
jenkins test sanity plinux

fjeremic:
@zl-wang removed some comments requesting jenkins builds (probably using Eclipse OpenJ9 syntax). I've kicked them off here.

@genie-omr build plinux

gita-omr:
The code changes LGTM.

compiler/p/codegen/OMRRealRegister.cpp
```diff
@@ -186,7 +186,14 @@ OMR::Power::RealRegister::setRegisterFieldRA(uint32_t *instruction)
...
    }
-
+void
+OMR::Power::RealRegister::setRegisterFieldRB(uint32_t *instruction)
+  {
+      if (self()->isConditionRegister())
+        *instruction |= fullRegBinaryEncodings[_registerNumber] << (pos_RB + 2);
+      else
+         *instruction |= fullRegBinaryEncodings[_registerNumber] << pos_RB;
+  }
```
0xdaryl: Am I wrong in assuming something like the following will generate smaller code on Power?  One wrinkle is the shift amount is no longer constant...
```
   int32_t RBshift = self()->isConditionRegister() ? pos_RB + 2 : pos_RB;
   *instruction |= fullRegBinaryEncodings[_registerNumber] << RBshift;
```

gita-omr: Some compilers may common fullRegBinaryEncodings[_registerNumber] between the two branches for the sake of space (not run time).

0xdaryl: Thanks for clarifying.

zl-wang:
Reviewers, please review this!

fjeremic:
@genie-omr build plinux

fjeremic:
AppVeyor build seems to have failed with infrastructure issues. The OMR plinux build passed though. Merging.

gita-omr:
@0xdaryl could you please add arch:power label?

PR #2335, opened 2018-02-22
Fix compilation issue on older versions of GCC

charliegracie:
On oldder versions of GCC abortTest.cpp was failing due to 'error:
comparison between signed and unsigned integer expressions'. The
compiler was treating the constant '0' as an int in the comparison.

The fix is to change the constant to '0u' to tell the compiler to
treat it as unsigned.


charliegracie:

PR #2336, opened 2018-02-22
Merge type size of a forward declaration with a later definition

keithc-ca:
DWARF information corresponding to a forward declaration does not include a size (nor can it).
This change merges the size of a subsequent definition (if found) so the generated blob will have the proper size information.

Issue: eclipse/openj9#378


charliegracie:

charliegracie:

PR #2337, opened 2018-02-22
Fix blob types for bit fields

keithc-ca:
A field
```
int bit : 1;
```
should be described in the blob as having type `int:1` instead of just `int`; this change corrects that.

Issue: eclipse/openj9#378


charliegracie:

PR #2344, opened 2018-02-23
Read Barrier on collected reference dereference

dsouzai:
The compiler must use read barriers when dereferencing collected references.

### DIFF
--- a/compiler/optimizer/VPConstraint.cpp
+++ b/compiler/optimizer/VPConstraint.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -1343,7 +1343,8 @@ TR::VPConstString *TR::VPConstString::create(OMR::ValuePropagation *vp, TR::Symb
 
    if (vpConstStringCriticalSection.hasVMAccess())
       {
-      void *string = *((void **) symRef->getSymbol()->castToStaticSymbol()->getStaticAddress());
+      uintptrj_t stringStaticAddr = (uintptrj_t) symRef->getSymbol()->castToStaticSymbol()->getStaticAddress();
+      uintptrj_t string = vp->comp()->fej9()->getStaticReferenceFieldAtAddress(stringStaticAddr);
       // with no vmaccess, staticAddress cannot be guaranteed to remain the same
       // during the analysis. so use a different hash input
       //
@@ -1352,11 +1353,11 @@ TR::VPConstString *TR::VPConstString::create(OMR::ValuePropagation *vp, TR::Symb
 
       // since vmaccess has been acquired already, chars cannot be null
       //
-      int32_t len = vp->comp()->fej9()->getStringLength((uintptrj_t)string);
+      int32_t len = vp->comp()->fej9()->getStringLength(string);
       int32_t i = 0;
       uint32_t hashValue = 0;
       for (int32_t i = 0; i < len && i < TR_MAX_CHARS_FOR_HASH; i++)
-         hashValue += TR::Compiler->cls.getStringCharacter(vp->comp(), (uintptrj_t)string, i);
+         hashValue += TR::Compiler->cls.getStringCharacter(vp->comp(), string, i);
 
       int32_t hash = (int32_t)(((uintptrj_t)hashValue) % VP_HASH_TABLE_SIZE);
 
@@ -1365,10 +1366,13 @@ TR::VPConstString *TR::VPConstString::create(OMR::ValuePropagation *vp, TR::Symb
       for (entry = vp->_constraintsHashTable[hash]; entry; entry = entry->next)
          {
          constraint = entry->constraint->asConstString();
-         if (constraint &&
-             string == *((void **) constraint->_symRef->getSymbol()->castToStaticSymbol()->getStaticAddress()))
+         if (constraint)
             {
-            return constraint;
+            uintptrj_t constraintStaticAddr = (uintptrj_t)constraint->_symRef->getSymbol()->castToStaticSymbol()->getStaticAddress();
+            if (string == vp->comp()->fej9()->getStaticReferenceFieldAtAddress(constraintStaticAddr))
+               {
+               return constraint;
+               }
             }
          }
       constraint = new (vp->trStackMemory()) TR::VPConstString(vp->comp()->getStringClassPointer(), vp->comp(), symRef);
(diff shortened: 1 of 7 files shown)


dsouzai:
@fjeremic @amicic @jdmpapin @andrewcraik could you please review.

compiler/env/OMRKnownObjectTable.cpp
```diff
@@ -115,7 +115,8 @@ OMR::KnownObjectTable::getIndexAt(uintptrj_t *objectReferenceLocation)
 #ifdef J9_PROJECT_SPECIFIC
    TR::VMAccessCriticalSection getIndexCriticalSection(self()->comp());
 #endif
-   uintptrj_t objectPointer = *objectReferenceLocation; // Note: object references held as uintptrj_t must never be compressed refs
+   // Note: object references held as uintptrj_t must never be compressed refs
+   uintptrj_t objectPointer = TR::Compiler->om.getReferenceAtAddress(self()->comp(), (uintptrj_t)objectReferenceLocation);
```
fjeremic: Could we use a `static_cast` / `reinterpret_cast` here (similarly elsewhere)?

compiler/env/OMRObjectModel.hpp
```diff
@@ -110,12 +111,43 @@ class ObjectModel
...
+    * \param
+    *     fieldAddress Address of field to be dereferenced
+    *
+    * \param
+    *     symRef symref of the node representing the field dereference
+    *
+    * \return
+    *     dereferenced field
+    */
+   uintptrj_t getStaticReferenceFieldAtAddress(TR::Compilation* comp, uintptrj_t fieldAddress, TR::SymbolReference *symRef) { return *(uintptrj_t *)fieldAddress; }
```
fjeremic: I would vouch to say this should assert via `notImplemented` rather than providing a default implementation. How you read static fields seems very downstream-project specific (similarly for below).

compiler/env/OMRObjectModel.hpp
```diff
@@ -110,12 +111,43 @@ class ObjectModel
...
    * @brief: Returns true if option for software read barriers is enabled in the VM's GC
    */
-   bool shouldReplaceGuardedLoadWithSoftwareReadBarrier() { return false; };
+   bool shouldReplaceGuardedLoadWithSoftwareReadBarrier() { return false; }
+
+   /**
+    * \brief Dereference a static reference field
+    *
+    * \param
+    *     comp TR::Compilation
```
fjeremic: Perhaps a better comment is needed here as to why this parameter is necessary or what it may be used for?

compiler/optimizer/VPConstraint.cpp
```diff
@@ -1343,7 +1344,8 @@ TR::VPConstString *TR::VPConstString::create(OMR::ValuePropagation *vp, TR::Symb
 
    if (vpConstStringCriticalSection.hasVMAccess())
       {
-      void *string = *((void **) symRef->getSymbol()->castToStaticSymbol()->getStaticAddress());
+      uintptrj_t stringStaticAddr = (uintptrj_t) symRef->getSymbol()->castToStaticSymbol()->getStaticAddress();
+      void *string = (void *)TR::Compiler->om.getStaticReferenceFieldAtAddress(vp->comp(), stringStaticAddr, symRef);
```
fjeremic: As far as I can tell everywhere this variable is being used it is used under a `uintptrj_t` context. Can we not change the type to `uintptrj_t` and avoid all these casts?

compiler/env/OMRKnownObjectTable.cpp
dsouzai: Hm not sure about `reinterpret_cast`, but `static_cast` doesn't like casting from pointer to non-pointer or vice versa.

compiler/env/OMRObjectModel.hpp
dsouzai: I'll have to check if there's any code not guarded by the `J9_PROJECT_SPECIFIC` macro that makes this call; if there is then we can't add that assert. There is also the counter proposal that the default behaviour (considering that we're only adding this API to account for read barriers of collected references, and not for things like the `TR_OpaqueClassBlock`) should be just a simple dereference.

compiler/env/OMRKnownObjectTable.cpp
fjeremic: https://ideone.com/Yj8T10

compiler/env/OMRObjectModel.hpp
dsouzai: Looks like every call to `getStaticReferenceFieldAtAddress` is protected by `J9_PROJECT_SPECIFIC` so I'll update the OMR implementation to require a downstream implementation.

amicic:
Looks good to me.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2387,16 +2383,13 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+                  if ((arrayObject != 0) &&
                       constrainIaloadCriticalSection.hasVMAccess())
                      {
-                     arrLength = *((int32_t *) (((uintptrj_t) *(void ***)p) + (uintptrj_t) vp->fe()->getOffsetOfContiguousArraySizeField()));
-
-                     if (arrLength == 0 && TR::Compiler->om.useHybridArraylets())
-                        {
-                        arrLength = *((int32_t *) (((uintptrj_t) *(void ***)p) + (uintptrj_t) vp->fe()->getOffsetOfDiscontiguousArraySizeField()));
-                        }
+                     //int32_t arrLength = TR::Compiler->om.getArrayLengthInElements(vp->comp(), arrayObject);
```
fjeremic: Is this line meant to be commented? Can it be removed?

dsouzai: I kept it commented because of 

https://github.com/eclipse/omr/pull/2344/files/0b3b80243b9c67f50c89fc700e308cdb12d67ffc#diff-30d6231397753ecc45846719f86be793R2401 and https://github.com/eclipse/omr/pull/2344/files/0b3b80243b9c67f50c89fc700e308cdb12d67ffc#diff-30d6231397753ecc45846719f86be793R2402

dsouzai:
Btw, I just realized this, but with the addition of `notImplemented` in the OMR version of `getStaticReferenceFieldAtAddress` , this PR now needs to be merged in tandem with the OpenJ9 PR (https://github.com/eclipse/openj9/pull/1260), otherwise, it will break OpenJ9.

compiler/optimizer/VPHandlers.cpp
fjeremic: Could you please add a comment above the commented out line saying why it is there so it is not removed by subsequent cleanup.

fjeremic:
> Btw, I just realized this, but with the addition of notImplemented in the OMR version of getStaticReferenceFieldAtAddress , this PR now needs to be merged in tandem with the OpenJ9 PR (eclipse/openj9#1260), otherwise, it will break OpenJ9.

Ok, that's fine. I've amended the PR title with a WIP prefix to prevent accidental merges. We can coordinate together to merge at an approriate time once both PRs are ready.

https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#downstream-dependencies

dsouzai:
@fjeremic Added requested comment. Also I changed all uses of `TR::Compiler.om->getStaticReferenceFieldAtAddress` to `comp->fej9()->getStaticReferenceFieldAtAddress` since all uses are protected by `J9_PROJECT_SPECIFIC` ; no point going through two function calls if we don't have to.

dsouzai:
Based on the discussion in https://github.com/eclipse/openj9/pull/1260 I've reworked the commit to only deal with static collected references (ie static fields).

Could you guys please re-review.

dsouzai:
@fjeremic @jdmpapin do you guys minding re-reviewing, especially https://github.com/eclipse/omr/pull/2344/commits/075eb20e8430a7a94f68605c8df75449fbc8b1c1

fjeremic:
@jdmpapin any concerns regarding this change?

jdmpapin:
LGTM

fjeremic:
@dsouzai please remove the WIP if this is ready to merge.

dsouzai:
@fjeremic Done.

PR #2347, opened 2018-02-26
Support for DDR on Linux s390 platforms

keithc-ca:
Issue: eclipse/openj9#378
* ddrgen needs to link with libelf.so
* produce both .so and .so.dbg files


charliegracie:
@genie-omr build all

PR #2351, opened 2018-02-28
RS direct reference pass

amicic:
In Concurrent Scavenger, move the preparation step for the second
(direct reference) pass to an earlier point, before concurrent phase
starts and more importatnly before Mutators resume and start adding new
references to Remembered Set.

If references are added to a an existing puddle that we are about the
iterate, we can encounter non-full fragments with NULL slots.

Now, any new references are guarantied to be remembered in a new puddle,
that we will not iterate in the second pass, which indeed is not
necessary and with high probability waste of time - those refs will be
subject to pruning in the next GC cycle, when they are more likely to be
pruned.


charliegracie:
@genie-omr build all zos

PR #2352, opened 2018-03-01
Fix handling of 'flag' macros

keithc-ca:
* fix regular expression for flag definitions so the following is properly recognized
```
    #define FLAG // comment
```
* generate *_DEFINED constants with buildflags option
* manage separate flag and constant namespaces
* put {flag} in the 'Constants' namespace and {flag}_DEFINED in the 'Flags' namespace


charliegracie:
@genie-omr build all zos

PR #2354, opened 2018-03-01
Improve support for cast expressions

keithc-ca:
* a type must be one or more C identifiers
* assume casts to unknown types can be safely ignored (e.g. `UDATA`)


charliegracie:
@genie-omr build all zos

PR #2356, opened 2018-03-06
Add trace points to portlib APIs related to cgroups

ashu-mehra:
Added trace points on error conditions and entry and exit
tracepoints for port library APIs related to cgroups.

### DIFF
--- a/port/common/omrport.tdf
+++ b/port/common/omrport.tdf
@@ -1013,3 +1013,42 @@ TraceExit=Trc_PRT_sysinfo_get_os_description_Exit Group=sysinfo Overhead=1 Level
 
 TraceEntry=Trc_PRT_sysinfo_os_has_feature_Entered Group=sysinfo Overhead=1 Level=5 NoEnv Template="sysinfo_os_has_feature: desc = %p, feature = %d"
 TraceExit=Trc_PRT_sysinfo_os_has_feature_Exit Group=sysinfo Overhead=1 Level=5 NoEnv Template="sysinfo_os_has_feature: returning with %zu."
+
+TraceException=Trc_PRT_isCgroupV1Available_statfs_failed Group=sysinfo Overhead=1 Level=1 NoEnv Template="isCgroupV1Available: statfs on %s failed with errno=%d"
+TraceException=Trc_PRT_isCgroupV1Available_tmpfs_not_mounted Group=sysinfo Overhead=1 Level=1 NoEnv Template="isCgroupV1Available: tmpfs is not mounted on %s"
+
+TraceException=Trc_PRT_addCgroupEntry_oom_for_cgroup_entry Group=sysinfo Overhead=1 Level=1 NoEnv Template="addCgroupEntry: failed to allocate memory for OMRCgroupEntry"
+TraceEvent=Trc_PRT_addCgroupEntry_added_new_entry Group=sysinfo Overhead=1 Level=3 NoEnv Template="addCgroupEntry: added new cgroup entry for subsystem %s and cgroup %s"
+
+TraceException=Trc_PRT_readCgroupFile_fopen_failed Group=sysinfo Overhead=1 Level=1 NoEnv Template="readCgroupFile: fopen failed for %s with errno=%d"
+TraceException=Trc_PRT_readCgroupFile_unexpected_format Group=sysinfo Overhead=1 Level=1 NoEnv Template="readCgroupFile: unexpected format of file %s"
+TraceEvent=Trc_PRT_readCgroupFile_available_subsystems Group=sysinfo Overhead=1 Level=3 NoEnv Template="readCgroupFile: available subsystems are 0x%llx"
+
+TraceException=Trc_PRT_readCgroupSubsystemFile_subsystem_not_available Overhead=1 Level=1 NoEnv Template="readCgroupSubsystemFile: requested subsystem 0x%llx is not available"
+TraceException=Trc_PRT_readCgroupSubsystemFile_oom_for_filename Overhead=1 Level=1 NoEnv Template="readCgroupSubsystemFile: failed to allocate memory for file name"
+TraceException=Trc_PRT_readCgroupSubsystemFile_fopen_failed Overhead=1 Level=1 NoEnv Template="readCgroupSubsystemFile: fopen failed for %s with errno=%d"
+TraceException=Trc_PRT_readCgroupSubsystemFile_unexpected_file_format Overhead=1 Level=1 NoEnv Template="readCgroupSubsystemFile: expected %d items to read, but read %d items"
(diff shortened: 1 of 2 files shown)


ashu-mehra:
fyi - @DanHeidinga

charliegracie:
@genie-omr build all zos

ashu-mehra:
I have made couple of trivial changes to the commit. Renamed
`Trc_PRT_addCgroupEntry_addedNewEntry` to `Trc_PRT_addCgroupEntry_added_new_entry` and 
`Trc_PRT_readCgroupFile_availableSubsystems` to `Trc_PRT_readCgroupFile_available_subsystems`
 to be consistent with other tracepoints I added.

port/common/omrport.tdf
```diff
@@ -1013,3 +1013,42 @@ TraceExit=Trc_PRT_sysinfo_get_os_description_Exit Group=sysinfo Overhead=1 Level
 
 TraceEntry=Trc_PRT_sysinfo_os_has_feature_Entered Group=sysinfo Overhead=1 Level=5 NoEnv Template="sysinfo_os_has_feature: desc = %p, feature = %d"
 TraceExit=Trc_PRT_sysinfo_os_has_feature_Exit Group=sysinfo Overhead=1 Level=5 NoEnv Template="sysinfo_os_has_feature: returning with %zu."
+
+TraceException=Trc_PRT_isCgroupV1Available_statfs_failed Group=sysinfo Overhead=1 Level=1 NoEnv Template="isCgroupV1Available: statfs on %s failed with errno=%d"
+TraceException=Trc_PRT_isCgroupV1Available_tmpfs_not_mounted Group=sysinfo Overhead=1 Level=1 NoEnv Template="isCgroupV1Available: tmpfs is not mounted on %s"
+
+TraceException=Trc_PRT_addCgroupEntry_oom_for_cgroup_entry Group=sysinfo Overhead=1 Level=1 NoEnv Template="addCgroupEntry: failed to allocate memory for OMRCgroupEntry"
+TraceEvent=Trc_PRT_addCgroupEntry_added_new_entry Group=sysinfo Overhead=1 Level=2 NoEnv Template="addCgroupEntry: added new cgroup entry for subsystem %s and cgroup %s"
```
DanHeidinga: Can the `noEnv` be removed from these?  `NoEnv` tracepoints have to do an expensive search to find the VMThread to get the tracebuffer.  For `TraceExceptions` this matters less and for asserts not at all.

Generally, if the code will need to get the VMThread for tracepoints, it should try to do that once and pass the thread along rather than having each tracepoint do the work.

DanHeidinga: Looking further - everything here is `NoEnv`so while the guidance is good for OpenJ9 tracepoints, it doesn't appear to apply here in OMR.

DanHeidinga: `Trc_PRT_addCgroupEntry_added_new_entry` This should probably be set to something higher than level=2.

port/unix/omrsysinfo.c
```diff
@@ -3607,12 +3618,16 @@ readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFl
...
+		Trc_PRT_readCgroupSubsystemFile_subsystem_not_available(subsystemFlag);
 		rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_UNAVAILABLE, "cgroup subsystem %s is not available", subsystemNames[subsystem]);
 		goto _end;
 	}
 	
 	cgroup = getCgroupNameForSubsystem(portLibrary, PPG_cgroupEntryList, subsystemNames[subsystem]);
 	if (NULL == cgroup) {
+		/* If the subsystem is available and supported, cgroup must not be NULL */
+		Trc_PRT_readCgroupSubsystemFile_missing_cgroup(subsystemFlag);
+		Trc_PRT_Assert_ShouldNeverHappen();
```
DanHeidinga: `Trc_PRT_Assert_ShouldNeverHappen` should be the last thing in the code path as it kills execution.
In this case, it should be moved to just before the `goto _end`.

port/unix/omrsysinfo.c
```diff
@@ -3768,8 +3796,11 @@ uint64_t
 omrsysinfo_cgroup_are_subsystems_available(struct OMRPortLibrary *portLibrary, uint64_t subsystemFlags)
 {
 #if defined(LINUX) && !defined(OMRZTPF)
+	Trc_PRT_sysinfo_cgroup_are_subsystems_available_Entry(subsystemFlags);
 	uint64_t available = portLibrary->sysinfo_cgroup_get_available_subsystems(portLibrary);
-	return (available & subsystemFlags);
+	uint64_t rc = available & subsystemFlags;
+	Trc_PRT_sysinfo_cgroup_are_subsystems_available_Exit(rc);
```
DanHeidinga: This code isn't c89 - the variables would need to be declared before the tracepoint.

port/unix/omrsysinfo.c
```diff
@@ -3789,8 +3820,10 @@ uint64_t
 omrsysinfo_cgroup_enable_subsystems(struct OMRPortLibrary *portLibrary, uint64_t requestedSubsystems)
 {
 #if defined(LINUX) && !defined(OMRZTPF)
+	Trc_PRT_sysinfo_cgroup_enable_subsystems_Entry(requestedSubsystems);
```
DanHeidinga: Declare the variable before the tracepoint.

ashu-mehra:
In addition to `Trc_PRT_addCgroupEntry_added_new_entry`, I have also changed the level of following trace points to 1:
`Trc_PRT_readCgroupFile_available_subsystems` and
`Trc_PRT_isRunningInContainer_container_detected`

ashu-mehra:
@DanHeidinga updated the change set as per the suggestions.

DanHeidinga:
> In addition to Trc_PRT_addCgroupEntry_added_new_entry,

By higher than level 2, I meant move them to level 3 or greater as they're good debug info but not necessary on every startup.

ashu-mehra:
> By higher than level 2, I meant move them to level 3 or greater as they're good debug info but not necessary on every startup.

If I understand correctly, you are suggesting this information is not critical enough to be considered for level 1. But I feel if some problem occurs, we would be better off having these details (like if we are in container or not, and subsystem/cgroup details) by default, rather than enabling these tracepoints explicitly to get these details. From that perspective, I felt it would be better to have these tracepoints at level 1.

DanHeidinga:
@ashu-mehra These tracepoints are good for debugging container detection issues and definitely should be available.  The common case is that the container detection will work correctly and this information can be determined from the API (in javacores, etc for languages like java) or from the core file using tools like DDR.

We need the tracepoints to make it easy to debug container detection issues, but they don't need to be enabled in every run.

ashu-mehra:
@DanHeidinga ok, I changed the level of those tracepoints to 3.

PR #2359, opened 2018-03-07
Fix existing PATH hack for Windows

ymanton:
The path list separator should be : not ; when running
under a Unix-like environment, otherwise this particular
path that's being added to the start of the list ends up
being prepended to the first path on the list and
invalidating both.

Also change back-slashes to forward-slashes to be
more consistent.

Signed-off-by: Younes Manton <ymanton@ca.ibm.com>

### DIFF
--- a/run_configure.mk
+++ b/run_configure.mk
@@ -83,7 +83,7 @@ endif
 ifneq (,$(findstring win_x86,$(SPEC)))
   include $(CONFIG_INCL_DIR)/configure_win_x86.mk
   # OMRTODO: this is J9 buildfarm specific code.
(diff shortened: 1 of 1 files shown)


charliegracie:
Was this breaking a  build somewhere?  This script is likely going away fairly soon and that line was a hack for internal builds that should be removed if we were going to keep this script.

ymanton:
It broke the first build I attempted on Windows because I had MSVC at the head of my PATH. Whether or not it breaks anything for anyone and how cryptic the subsequent errors are depends on what that first entry is, but I would think that it's going to be a build tool in most people's environments.

PR #2360, opened 2018-03-07
Downgrade to VS 10 as this is our minimum supported version

charliegracie:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, 2017 IBM Corp. and others
+# Copyright (c) 2017, 2018 IBM Corp. and others
 # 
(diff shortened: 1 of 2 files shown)


fjeremic:
Any background on why this is needed would be helpful.

charliegracie:
It is the current minimum VS level we support.  The project constantly gets Issues about breaking this old version of the compiler so we should be testing it.  *If* we could remove this as a supported version I would be happy to use a newer version as our standard compiler.

charliegracie:
I am going to open an Issue to have documentation written describing the minimum compiler levels we support on each platform.  It will also discuss what versions we test in our CI builds.

charliegracie:
Closing as there are some compiler issues with how we do CMake builds on windows that I need to investigate.  I will re-open once I have a working solution.

PR #2362, opened 2018-03-08
Replace CARD_SIZE & CARD_SIZE_SHIFT with macros

keithc-ca:
ddrgen associates literals of anonymous enums with the
enclosing scope. As there is no such scope in the case of
CARD_SIZE & CARD_SIZE_SHIFT these literals are not captured.
As macros they will be included in the superset & blob.

In support of https://github.com/eclipse/openj9/issues/378.

### DIFF
--- a/gc/include/omrmodroncore.h
+++ b/gc/include/omrmodroncore.h
@@ -38,10 +38,11 @@
  */
 typedef U_8 Card;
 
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all zos

PR #2364, opened 2018-03-08
Fix handling of literals of anonymous enum types

keithc-ca:
Fix
* promote _enumMembers from ClassType to NamespaceUDT
* collect literals of a nested anonymous enum in enclosing namespace

Cleanup
* change signature of Type::getSymbolKindName()
* check that IO operations are successful
* remove unnecessary #include directives
* remove useless method TypedefUDT::insertUnique()
* manage printing type header in JavaSupersetGenerator
* extract common code for visiting class and union types

Addition
* add TypePrinter for debugging (currently disabled)

This should be the final OMR change required to enable https://github.com/eclipse/openj9/issues/378 on Linux.

### DIFF
--- a/ddr/lib/ddr-ir/TypePrinter.cpp
+++ b/ddr/lib/ddr-ir/TypePrinter.cpp
@@ -1,0 +1,173 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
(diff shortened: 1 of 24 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build aix

charliegracie:
@genie-omr build all

ddr/lib/ddr-ir/TypePrinter.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "ddr/ir/EnumUDT.hpp"
+#include "ddr/ir/Field.hpp"
+#include "ddr/ir/TypedefUDT.hpp"
+#include "ddr/ir/UnionUDT.hpp"
+
+void
+TypePrinter::printIndent() const
+{
+	for (int32_t i = _indent; i > 0; --i) {
+		printf("  ");
```
charliegracie: Only question is should we be using the PortLibrary tty_printf instead of printf in this class and ddrgen.cpp?

keithc-ca: Perhaps that would be better, but this is only a debugging aid.

PR #2365, opened 2018-03-09
Fix incorrect xrelease prefix while tracing

cathyzhyi:
The **xrelease** prefix was added to S8MemImm4 while tracing. Fix it
to avoid confusion.


vijaysun-omr:
X86 change...so no need for builds on other platforms. Merging.

PR #2366, opened 2018-03-12
Reduce size of generated *.i files

keithc-ca:
Reducing the footprint of *.i files eliminates writing useless content and also avoids later, unnecessary reads by getmacros. In OMR alone, the reduction is from over 230MB to about 36MB.


charliegracie:
@genie-omr build all zos

PR #2372, opened 2018-03-15
Use sed instead of grep

keithc-ca:
Unlike grep, sed won't fail if there are no matching lines. Even though make is ignoring such failures, build logs are cluttered with noise like this:

```
../../omrmakefiles/rules.mk:400: recipe for target 'BaseVirtual.i' failed
make[9]: [BaseVirtual.i] Error 1 (ignored)
```

Improves upon https://github.com/eclipse/omr/pull/2366.


charliegracie:
@genie-omr build all zos

PR #2376, opened 2018-03-18
WIP: Optional Test Output for xUnit Jenkins Plugin

smlambert:
Googletest can generate output files that can be represented graphically using the xUnit plugin.  This helps see stacktrace and console output on a per test basis in Jenkins without having to scroll through console output.

### DIFF
--- a/buildenv/jenkins/Jenkinsfile
+++ b/buildenv/jenkins/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
                 timestamps {
                     echo 'Configure...'
                     sh '''pwd'''
-                    // sh'''cd omr'''
+                    sh "cd $WORKSPACE/omr"
                     sh'''make -f run_configure.mk OMRGLUE=./example/glue SPEC=linux_x86-64'''
                    
                     echo 'Compile...'
--- a/buildenv/jenkins/Jenkinsfile
+++ b/buildenv/jenkins/Jenkinsfile
@@ -19,12 +19,14 @@ pipeline {
         string (defaultValue: 'master', description: 'Branch name to use', name: 'BRANCH_NAME')
         string (defaultValue: '', description: 'Machine label expression for jenkins node, like Linux&&390 or aix&&ppc, etc', name: 'MACHINE_LABEL')
         choice (choices: 'linux_x86-64\nosx_x86-64\nlinux_ppc-64\nlinux_ppc-64_le_gcc\nlinux_armmak\nlinux_390-64\naix_ppc-64', description: 'SPEC?', name: 'SPEC')
-        string (defaultValue: '/usr/lib/ccache', description: 'PATH to use', name: 'PATH')
+        string (defaultValue: '/usr/lib/ccache', description: 'If SPEC=aix_ppc-64, PATH=/home/u0020236/tools, otherwise use default', name: 'PATH')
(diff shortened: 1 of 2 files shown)


smlambert:
jenkinsomrtestview

smlambert:
Can add to the pipeline script something like this:

    post {
        always {
            step([$class: 'XUnitBuilder',
                thresholds: [[$class: 'FailedThreshold', unstableThreshold: '1']],
                tools: [[$class: 'GoogleTestType', skipNoTestFiles: true, pattern: '**/test_output/**']]])

            echo 'Cleanup workspace'
            deleteDir()
        }
    }

charliegracie:
As we move to a complete `CMake` solution (just waiting on AIX and zOS) we will be using `CTest` and not `omrtest.mk` to drive the testing.  I think this should be easy enough to add into that environment but just want to make sure the idea doesn't get lost in the switch over.

charliegracie:
For the JenkinsFile I think that is a good start for the old style Merge builds.  I do not think it would work for the PR builds as I think the GIT SCM line is fairly different.  They may be able to be the same though. 

Would it be possible to update it for the CMake style builds? Add support for ccache output? Add support for updating the GitHub commit status?

charliegracie:
I actually just realized you probably can not even view the current Jenkins jobs :).  I can copy one of them for you :).  Or I can suggest the changes to your jenkinsfile here in the PR.

charliegracie:
As a prep step I installed the xUnit plugin and I am going to create a test job against your repo :)

smlambert:
Thanks, yes you are correct, I can't see your current Jenkinsfiles, so please just make detailed suggestions to this PR, or send me an example of what you are actually using.  :)

charliegracie:
Here are the current [Jenkins files](https://github.com/charliegracie/omr/tree/jenkinsfiles/scripts/jenkins) used for the OMR merge builds.

smlambert:
Cool thanks @charliegracie - I'll rework, using those as the starting point. I used buildenv/jenkins as that is what some of our other open projects use, but now I see your branch, I could shift to scripts/jenkins to avoid a new dir.

charliegracie:
Also if you look in the CMakeLists.txt in `fvtest\*`  you can see calls to `add_test` to add tests into the CTest testing.  The command lines are there so I believe it should be easy to create similar output files I think.

smlambert:
@youngar - can you or someone more familiar with cmake advise me on some syntax, and how to accomplish the same or similar thing as what i have in the .mk file in the PR?  I can/will dig into cmake doc, but you can likely shortcut this... also, the cmake doc apologies for not having documented all of their variables... looking for something that will provide me with unique filenames.  In the .mk case, I used $@, as it evaluates to the target name.

charliegracie:
@dnakamura can you help out with Shelley's question?  Thanks

smlambert:
Thanks @dnakamura!

0xdaryl:
Is there an update on the status of this WIP PR?  If this won't be ready for merge soon, may I suggest it be closed while you work on it in your personal fork and then re-open again when it's ready for final review/merge?  Thank you!

smlambert:
Apologies, I understood that @dnakamura was going to come back with some info re: cmake targets, and I was negligent in pursuing it.  I can close this.  Doesn't seem needed/wanted, I just wanted to show folks how you could get a better view of test results in Jenkins from a small change to the gtest output.

Issue #2377, opened 2018-03-18
DDR: cannot open file 'pthreads.lib' while building on Windows using CMake


samolisov:
## Environment
```
> cmake --version

cmake version 3.11.0-rc3

CMake suite maintained and supported by Kitware (kitware.com/cmake).

> clang-cl --version

clang version 6.0.0 (tags/RELEASE_600/final)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\llvm\bin
```

*Windows Kit 10.0.16299.0*

## Problem

If DDR is included in the build configuration, the following error appears just before the build end:

```bash
[414/415] Linking CXX executable omr_ddrgen.exe
FAILED: omr_ddrgen.exe
cmd.exe /C "cd . && "C:\Program Files\cmake\bin\cmake.exe" -E vs_link_exe --intdir=ddr\tools\ddrgen\CMakeFiles\omr_ddrgen.dir --manifests  -- lld-link /nologo ddr\tools\ddrgen\CMakeFiles\omr_ddrgen.dir\ddrgen.cpp.obj  /out:omr_ddrgen.exe 
...
util\pool\j9pool.lib pthread.lib psapi.lib pdh.lib kernel32.lib

...

lld-link: error: could not open pthread.lib: no such file or directory
[415/415] Linking CXX executable omr_blob_reader.exe
ninja: build stopped: subcommand failed.
```

**Note:** The problem isn't reproducible when `MSVC` is in use.

## Steps to reproduce

Using *Ninja* as a make tool:
```bash
> cmake -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl -DCMAKE_LINKER:STRING=lld-link -DOMR_GC=ON -DOMR_DDR=ON .. -G"Ninja"
> cmake --build .
```

Using *NMake* as a make tool:
```bash
> cmake -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl -DCMAKE_LINKER:STRING=lld-link -DOMR_GC=ON -DOMR_DDR=ON .. -G"NMake Makefiles"
> cmake --build .
```

PR #2379, opened 2018-03-18
CMake: Fix DDR on Windows building using Clang

samolisov:
When `clang-cl` is used as a C/C++ compiler on Windows to build OMR, CMake attempts to link `DDR` with the `pthread` library and the following error appears:

```bash
lld-link: error: could not open pthread.lib: no such file or directory
```

This commit fixes the error.

In order to improve code readability, a refactoring of system information detection was applied: a new CMake variable is introduced: `OMR_PLATFORM_THREAD_LIBRARY`. A corresponding value is assigned to the variable in a `cmake/modules/platform/toolcfg/${OMR_TOOLCONFIG}` file. This refactoring allows developers just to include the platform dependent thread library to a `target_link_libraries` invocation if the developed component should be linked with the thread library.

**Warning:** The following code is put to `gnu.cmake`:

```cmake
# we dont actually have a clang config and use gnu on non-Windows,
# so we have to detect Apple Clang here.
# see ../../OmrDetectSystemInformation.cmake
if(CMAKE_C_COMPILER_ID MATCHES "^(Apple)?Clang$")
       set(THREAD_LIBRARY pthread)
else()
       set(THREAD_LIBRARY -pthread)
endif()
```

This code should work on all non-Windows platforms when `clang` is used since we have no `clang.cmake` - a file with `Clang` config. I've tried to build `OMR` on Linux using `Clang 5` but builds fail - no working linker configuration, so even the `tracegen` tool hasn't been linked. I think a separated configuration for `Clang` is required.

Tested on `gcc/Linux` and `Clang/Windows` as well as `MSVC/Windows` **Must be tested** on `XLC` and `Clang/MacOS X`.

Closes: #2377

### DIFF
--- a/cmake/modules/platform/toolcfg/gnu.cmake
+++ b/cmake/modules/platform/toolcfg/gnu.cmake
@@ -79,3 +79,13 @@ set(PASM_FLAGS -x assembler-with-cpp -E -P)
 
 set(SPP_CMD ${CMAKE_C_COMPILER})
 set(SPP_FLAGS -x assembler-with-cpp -E -P)
+
+# Configure the platform dependent library for multithreading
+# We dont actually have a clang config and use gnu on non-Windows,
+# so we have to detect Apple Clang here.
+# see ../../OmrDetectSystemInformation.cmake
+if(CMAKE_C_COMPILER_ID MATCHES "^(Apple)?Clang$")
+	set(OMR_PLATFORM_THREAD_LIBRARY pthread)
+else()
+	set(OMR_PLATFORM_THREAD_LIBRARY -pthread)
+endif()
--- a/cmake/modules/platform/toolcfg/xlc.cmake
+++ b/cmake/modules/platform/toolcfg/xlc.cmake
@@ -96,3 +96,6 @@ list(APPEND TR_C_COMPILE_OPTIONS
 
 set(SPP_CMD ${CMAKE_C_COMPILER})
 set(SPP_FLAGS -E -P)
+
+# Configure the platform dependent library for multithreading
+set(OMR_PLATFORM_THREAD_LIBRARY -lpthread)
--- a/cmake/modules/platform/toolcfg/msvc.cmake
+++ b/cmake/modules/platform/toolcfg/msvc.cmake
@@ -69,6 +69,9 @@ elseif(OMR_ENV_DATA32)
 	)
 endif(OMR_ENV_DATA64)
 
+# Configure the platform dependent library for multithreading
+set(OMR_PLATFORM_THREAD_LIBRARY Ws2_32.lib)
+
 macro(omr_toolconfig_global_setup)
 	# Make sure we are building without incremental linking
 	omr_remove_flags(CMAKE_EXE_LINKER_FLAGS    /INCREMENTAL)
--- a/cmake/modules/platform/toolcfg/verify.cmake
+++ b/cmake/modules/platform/toolcfg/verify.cmake
@@ -20,3 +20,4 @@
 #############################################################################
 
 omr_assert(WARNING TEST DEFINED OMR_WARNING_AS_ERROR_FLAG)
+omr_assert(FATAL_ERROR TEST DEFINED OMR_PLATFORM_THREAD_LIBRARY)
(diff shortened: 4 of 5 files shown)


charliegracie:
On Windows when we are compiling with clang-cl shouldn't we still be linking against `Ws2_32.lib`?

charliegracie:
@dnakamura or @rwy0717 Can you review this please?

charliegracie:
@genie-omr build all

samolisov:
I've double checked when -`DCMAKE_LINKER` is assigned to `lld-link` and `Ws2_32.lib` is specified as the `THREAD_LIBRARY`, OMR applications and libraries are built OK including `DDR`. If `pthread`  are added to the linker's command line, the same error is here (I didn't specify the correct linker when got this error, although - my bad):

```bash
lld-link: error: could not open pthread.lib: no such file or directory
```

cmake/modules/platform/toolcfg/gnu.cmake
```diff
@@ -79,3 +79,13 @@ set(PASM_FLAGS -x assembler-with-cpp -E -P)
 
 set(SPP_CMD ${CMAKE_C_COMPILER})
 set(SPP_FLAGS -x assembler-with-cpp -E -P)
+
+#Configure the platform dependent library for multithreading
+#TODO: we dont actually have a clang config and use gnu on non-Windows,
+#so we have to detect Apple Clang here.
+#see ../../OmrDetectSystemInformation.cmake
+if(CMAKE_C_COMPILER_ID MATCHES "^(Apple)?Clang$")
+	set(THREAD_LIBRARY pthread)
```
rwy0717: Out of consistency, this variable should be called `OMR_PLATFORM_THREAD_LIBRARIES`, and we need to assert that this variable is defined in verify.cmake.

Also, it was a deliberate decision to classify any gnuish toolchain as "gnu", so this isn't really a TODO. If we want to revisit that decision, I'm open for discussing alternatives.

cmake/modules/platform/toolcfg/xlc.cmake
```diff
@@ -96,3 +96,6 @@ list(APPEND TR_C_COMPILE_OPTIONS
 
 set(SPP_CMD ${CMAKE_C_COMPILER})
 set(SPP_FLAGS -E -P)
+
+#Configure the platform dependent library for multithreading
+set(THREAD_LIBRARY -lpthread)
```
rwy0717: This should actually be just `pthread`. I believe the original code's `-l` was a typo.

rwy0717:
This is a nice improvement to the way our cmake is organized--thanks. I'm a little confused though, doesn't this patch preserve the original behaviour? How does this patch fix your link error?

samolisov:
This patch just changes pthread.lib to ws2_32.lib when clang is used for building on Windows. I've put the requested changes ASAP except the one for XLC: I believe it should be changed by a separated PR and commit.

rwy0717:
Oh, hah, I see what I was missing:

When we're picking the toolcfg, we usually just rely on the `CMAKE_C_COMPILER_ID`. What I didn't realize was, when the compiler is clang, we'll also check the `CMAKE_C_SIMULATE_ID` for `MSVC`. This patch makes sense now.

I'm happy with your solution, so you just have my variable-naming nitpick. And as you said, we can save the XLC fix for a later PR.

samolisov:
@rwy0717 I've changed the variable's name in the code as well as in the commit message and the PR. I've put the assert: it will throw a `FATAL_ERROR` if the `OMR_PLATFORM_THREAD_LIBRARY` (I think it should be one Library) variable is undefined. If you think `WARNING` will be enough, please let me know.

**PS:** It's interesting if I make a typo on `${OMR_PLATFORM_THREAD_LIBRARY}` in `omr\thread\CMakeLists.txt`, nothing will happen: DDR will still be built fine. I'm not so familiar with the OMR thread and port libraries but it looks like `pthread` or `Ws2_32` are not required at all...

rwy0717:
Looks good to me.

charliegracie:
@genie-omr build all

charliegracie:
Thanks for the great work!

PR #2382, opened 2018-03-21
Enable Windows stack protection (/GS flag)

theresa-m:
OpenJ9 PR: https://github.com/eclipse/openj9/pull/1494

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>


charliegracie:
This option was not present in the CMake builds so it did not have to be removed.

Issue #2333, opened 2018-02-22
Build OMR compiler technology with port library


0xdaryl:
Require the OMR port library to be built when the OMR compiler technology is integrated in a downstream project, the test compiler, or JitBuilder.  This is a pre-requisite to the compiler technology using more port library features.

Ensure the solution is flexible enough to not integrate the OMR port library if a consuming project chooses to provide their own port library.

0xdaryl:
@nbhuiyan @rwy0717

nbhuiyan:
@0xdaryl I changed the reference to this issue in the commit message from `Closes` to just `Issue` since I wasn't sure which of the 2 PRs would go first. I think this can be closed now.

fjeremic:
Closing due to above comment.

PR #2385, opened 2018-03-21
WIP: Build Port lib with compiler technology (Autotools)

nbhuiyan:
The commit message contains the details. An upcoming PR will add CMake support, after which we can begin to use the port library within the compiler in OMR. You may test this out locally.

### Commit message:
This commit is the first step towards leveraging more port
library features in the compiler. A new static library:
libcompilerportstatic.a combines the port library with its
dependencies to make it easier to link it to the compiler
products. The downstream projects consuming the compiler
are still free to use their own port library.

The JitBuilder example tests have been temporarily disabled
to prevent build failures. These example tests may need
some significant changes to the Makefile to properly
work with this new dependency.

Issue: #2333


nbhuiyan:
@0xdaryl @rwy0717

nbhuiyan:
@0xdaryl @rwy0717 Ready for review

GNUmakefile
```diff
@@ -187,7 +189,8 @@ postbuild_targets += tests
 main_targets := $(sort $(main_targets))
 test_targets := $(sort $(test_targets))
 
-targets += $(tool_targets) $(prebuild_targets) $(main_targets) omr_static_lib $(test_targets) ddr
+targets += $(tool_targets) $(prebuild_targets) $(main_targets) omr_static_lib $(test_targets) ddr $(compiler_prereqs)
```
rwy0717: My understanding is, this will always enable the compiler prereqs, which is not the right thing. Iff the compiler is enabled, you should be assert that it's prereqs are enabled in the configure script.

(you edit configure.ac, and compile it into configure using autoconf).

jitbuilder/release/Makefile
```diff
@@ -109,189 +108,203 @@ testall: all_goal all
 # Even experimental stuff:
 testexperimental: experimental_goal all
 
+# Linker args
+JITBUILDER_LINK_FLAGS=\
+	-L. \
```
rwy0717: nit: use spaces, not tabs.

GNUmakefile
```diff
@@ -157,20 +157,22 @@ test_targets += fvtest/vmtest
...
+  main_targets += jitbuilder
+  test_targets += fvtest/jitbuildertest jitbuilder/release
 endif
 
 ifeq (yes,$(ENABLE_DDR))
   postbuild_targets += ddr
   ddr:: staticlib
 endif
 
+compiler_prereqs += port thread util/pool util/omrutil util/avl util/hashtable util/hookable
```
rwy0717: nit: this should be `=` not `+=`

GNUmakefile
```diff
@@ -216,7 +219,7 @@ prebuild: tools
 tools:
 	@$(MAKE) -f GNUmakefile $(tool_targets)
 
-.PHONY: tools prebuild mainbuild staticbuild tests postbuild
+.PHONY: tools prebuild mainbuild staticlib tests postbuild
```
rwy0717: nice catch.

jitbuilder/release/Makefile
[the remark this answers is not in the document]
nbhuiyan: I used tabs for all indentations due to this section of the Coding Standard doc:  https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md#indent-code-using-tabs-not-spaces

Should I still go ahead and change it to spaces?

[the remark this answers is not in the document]
rwy0717: Yep, switch to 2 space indentation. Don't use tabs in makefiles, unless you're writing a recipe.

GNUmakefile
nbhuiyan: All of the compiler prereq targets get built by default, and `port` and `thread` can be optionally disabled. Adding `compiler_prereqs` to `targets` was a way to ensure that someone trying to build the Test Compiler or JITBuilder would also trigger building of the port and thread libraries even if they were set to be disabled. However, the issue with this is that it is removing the ability to explicitly disable the building of `port` and `thread`.

After discussing with @rwy0717 , we have decided that it is best not to assert enabling the compiler dependencies in configure.ac mainly for the following reasons:
* we did not think anyone would have a reason to explicitly disable the port and thread libraries if they want to build the Test Compiler and JITBuilder. Doing so would result in an error anyway if there are no port and thread libraries to link.
* Currently, configure.ac does not make any assertions to ensure top-level inter-component dependencies are enabled despite there being dependencies. It does not make sense to add a special assertion for the Test Compiler and JITBuilder.

Therefore, it is reasonably safe to remove `$(compiler_prereqs)` from `targets`. It would then allow explicitly disabling the building of `port` and `thread`, and will be alright if the test compiler and JITBuilder are not explicitly enabled.

fvtest/jitbuildertest/Makefile
```diff
@@ -47,10 +47,18 @@ MODULE_CXXFLAGS += -std=c++0x $(OMR_GTEST_CXXFLAGS)
 
 MODULE_STATIC_LIBS += \
   jitbuilder \
-  omrGtest
```
rwy0717: missed a copyright update :(

rwy0717:
one missing copyright update otherwise LGTM

nbhuiyan:
Totally missed that somehow. This kind of demonstrates the convenience we'd enjoy by having the [automatic copyright date checks](https://github.com/eclipse/omr/issues/2254) enabled for OMR.

PR #2389, opened 2018-03-23
Fix incorrect return value from AtomicFetchAndAdd

0dvictor:
AtomicFetchAndAdd returns incorrect result, fixing it.


0dvictor:
Currently this code is not used by any known project and hence no Issue was reported.

andrewcraik:
looks good to me - @0xdaryl over to you

0xdaryl:
Good catch.  A genie build is not needed because this is x86-specific.

PR #2390, opened 2018-03-24
Fix GS compile error by removing inline asm

theresa-m:
Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

### DIFF
--- a/port/win32/omrintrospect.c
+++ b/port/win32/omrintrospect.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


port/win32/omrintrospect.c
```diff
@@ -307,46 +307,9 @@ setup_native_thread(J9ThreadWalkState *state, CONTEXT *sigContext)
 		 * current thread. It seems to work, however, if we populate the EIP and EBP context
 		 * data manually (which is what the assembler does below).
 		 */
-#ifdef WIN64
```
keithc-ca: Perhaps the comment above should be updated given the removal of the assembly code?

theresa-m: Thanks :) updated.

charliegracie:
No need for further CI testing as this is in Windows only code.

DanHeidinga:
Did anyone confirm that the comment regarding the limitations from MSDB was no longer correct?

charliegracie:
We removed the usage of the call that the comment was talking about.  In the past we used the removed API on 32 bit since the API we are using for both 32/64 bit now was not available.  The API used now was made available on XP as far as I can see we should be ok.

DanHeidinga:
Reading https://bugs.chromium.org/p/crashpad/issues/detail?id=53 , which was updated in 2015, the function is broken on 32bit windows.  The claim is that it returns the caller's caller's values.

keithc-ca:
The one or two frames that might be omitted from that context aren't helpful; they'd be setup_native_thread and startDo/nextDo. The interesting frames are much deeper in the stack.

DanHeidinga:
Seriously?  You'd be happy working with an inconsistent set of registers?

keithc-ca:
The old assembly version was only capturing Eip and Ebp; I'm assuming RtlCaptureContext at least captures them from the same frame.

DanHeidinga:
Did you read the link I pasted?

DanHeidinga:
Also, the original code did two things: `GetThreadContext` to lookup the context and then the assembly code to fix up the `ebp` & `eip`

DanHeidinga:
I've opened an OpenJ9 issue to validate that OpenJ9's javacores, which use this code to fetch the native stack, are still correct once this change hits the OpenJ9 builds.

PR #2391, opened 2018-03-25
Make memory reference separate index register for PackedDecimal

NigelYiboYu:
In memory reference constructor, if the load/store is a packed decimal
load/store and when vector VSI instructions are used, the index register
need to be handled separately by an intermediate LA instruction because
VSI instructions don't support index registers.

### DIFF
--- a/compiler/z/codegen/S390GenerateInstructions.cpp
+++ b/compiler/z/codegen/S390GenerateInstructions.cpp
@@ -2123,6 +2123,9 @@ TR::Instruction * generateVSIInstruction(
                       TR::MemoryReference    * mr ,
                       uint8_t                  imm3)  /* 8 bits */
    {
+   TR_ASSERT_FATAL(mr != NULL, "NULL memory reference for VSI instruction\n");
+   mr->separateIndexRegister(n, cg, true, NULL);
+
    return new (INSN_HEAP) TR::S390VSIInstruction(cg, op, n, reg, mr, imm3);
    }
 


compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -578,16 +578,26 @@ OMR::Z::MemoryReference::MemoryReference(TR::Node * rootLoadOrStore, TR::CodeGen
          self()->createUnresolvedDataSnippet(rootLoadOrStore, cg, symRef, tempReg, isStore);
          }
 
+
+      // VSI vector load/store instructions, such as VSTRL, VLRL, and VSTEB, lack the I field and do not support
+      // index registers. As a result of this, we need to set separate index register so that an intermediate LA instruction
+      // is emitted to add up the index and base registers.
+      bool isPackedDecimalVectorLoadStore = rootLoadOrStore->getType().isAnyPacked()
+              && (TR::Compiler->target.cpu.getS390SupportsVectorPackedDecimalFacility() && !TR::Options::getCmdLineOptions()->getOption(TR_DisableVectorBCD))
+              && (rootLoadOrStore->getOpCode().isLoad() || rootLoadOrStore->getOpCode().isStore());
```
fjeremic:
```
rootLoadOrStore->getOpCode().isLoad() || rootLoadOrStore->getOpCode().isStore()
```

This is always true in this constructor, right? At least the code above this seems to assume so. Is it possible to fold this?

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -578,16 +578,26 @@ OMR::Z::MemoryReference::MemoryReference(TR::Node * rootLoadOrStore, TR::CodeGen
...
-      if ((rootLoadOrStore->getType().isInt64() || (symbol && symbol->getSize() == 8)) &&
-         !(TR::Compiler->target.is64Bit() || cg->use64BitRegsOn32Bit()) &&
-          !canUseRX &&
-          (storageReference == NULL))
+      bool separateIndexRegisterForRSLoadStore = (rootLoadOrStore->getType().isInt64() || (symbol && symbol->getSize() == 8)) &&
+                                                 !(TR::Compiler->target.is64Bit() || cg->use64BitRegsOn32Bit()) &&
+                                                 !canUseRX &&
+                                                 (storageReference == NULL);
+
+      if (isPackedDecimalVectorLoadStore || separateIndexRegisterForRSLoadStore)
```
fjeremic: I don't think this will work. For `LM` this works because there exists a long displacement `LMY` alternative. From `VSTRL` for example there does not exist a long displacement version. What happens if the displacement is more than 4096? 

I'd like to see a test case for this via Tril. I suspect this implementation will fail in undefined ways.

In my opinion this type of code does not belong in the MemoryReference constructor. The constructor should in theory not care about instruction formats. It should create the best possible memory reference given a node. If the instruction in which the memory reference will be used has certain restrictions, such as `VSTRL`, the `generateVSIInstruction` API should handle such restrictions.

See `generateRSInstruction` for example which supports the exact same memory reference format as `VSI` instructions:

https://github.com/eclipse/omr/blob/03270bf3b9e51e02c22fc8e6c5b88c60d928db7c/compiler/z/codegen/S390GenerateInstructions.cpp#L1086-L1100

We need to also be calling `enforce4KDisplacementLimit` to fixup the displacement if needed. See `enforceSSFormatLimits` as a relevant example.

compiler/z/codegen/OMRMemoryReference.cpp
NigelYiboYu: It should be fine to remove the load/store op-code check. Let me test it without this constraint.

fjeremic:
@genie-omr build zlinux zos

PR #2383, opened 2018-03-21
WIP: Tenure TLH memory recycling

RSalman:
Changes to preserve TLH remainders for reserved memory which can be
used for subsequent allocations. That is, references to the base and top
of allocated TLHs are saved (where appropriate) upon abandon. This
effectively reduces the amount of bytes discarded during GC (~35%).

Signed-off-by: Salman Rana <salman.rana@ibm.com>

PR #2392, opened 2018-03-26
Tenure Heap Recycling

RSalman:
Changes to preserve TLH remainders (scavenger copy cache) for reserved memory which can be reused for subsequent GCs. That is, references to the base and top of allocated TLH are saved (where appropriate) upon abandon.

This effectively reduces the tenure heap consumption, savings depend on 2 factors:
- Thread count
- Tenuring rate

For this reason, savings vary from negligible to magnitude of order lower tenure consumption rate. For instance, high tenuring (from non-generational workload) and low GC thread count results in lower savings, and opposite, very generational workload (low tenuring rate) with high GC thread count results in high savings.
 
For concurrent scavenger, we obtain additional savings from mutator threads as they create copy caches through a read barrier.
 
Additionally, some simplifications were done - the explicit copy cache flushing at the beginning of the 2nd STW phase of CS cycle was removed as it is already implicity done by VMinterface code

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -222,6 +222,8 @@ public:
 
 #if defined(OMR_GC_MODRON_SCAVENGER)
 	MM_Scavenger *scavenger;
+	void *_masterThreadTenureTLHRemainderBase;  /**< base and top pointers of the last unused tenure TLH copy cache, that will be loaded to thread env during master setup */
+	void *_masterThreadTenureTLHRemainderTop;
 #endif /* OMR_GC_MODRON_SCAVENGER */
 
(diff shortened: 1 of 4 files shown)


RSalman:
Please see PR https://github.com/eclipse/omr/pull/2383 for review comments that were previously made.

charliegracie:
@genie-omr build all zos

charliegracie:
@amicic are you ok with these changes now?

amicic:
@charliegracie Code is ok, but we are still doing regression testing, to confirm nothing is broken in Concurrent Scavenger world, due to some non-trivial simplifications/reorganizations.

charliegracie:
@genie-omr build all zos

Issue #2386, opened 2018-03-22
FVTest: rastest hangs when OMR is compiled with clang-cl on Windows


samolisov:
## Problem

FVTests hang when OMR is compiled with `clang-cl`  (Clang 6.0.0) on Windows 10. The problem is not reproduced when `MSVC` is used for compilation. 

The output of the unit tests is the follows:

```
4: [----------] 2 tests from CpuLoadAgentOpts/RASAgentTest
4: cpuLoadAgent: Agent_OnLoad(options="<NULL>")
4: cpuLoadAgent: BindCurrentThread passed, vmThread=0x000001FD16B62E30
4: Test GetProcessCpuLoad()
4: unknown file: error: SEH exception with code 0xc00000fd thrown in the test body.
```
(the command `ctest -V` is still running but nothing more happens).

## Steps to reproduce

```bash
> cmake -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl -DCMAKE_LINKER:STRING=lld-link -DCMAKE_BUILD_TYPE=Debug -DOMR_GC=ON -DOMR_EXAMPLE=ON -DOMR_DDR=ON -DOMR_FVTEST=ON  .. -G"Ninja"

> cmake --build .

> ctest -V
```

## Analysis

I've had a look at the sources of the `omr/fvtest/rastest/cpuLoadAgent.c` and see an allocation of a matrix on stack in the body of the `static void matrixSquare(void)` method: 

```c
#define MATRIX_SIZE 1000

...

static void
matrixSquare(void)
{
	/* Declare and initialize the matrix. memset works here since this is a true 2D
	 * array defined in the same scope as the memset. */
	uintptr_t matrix[MATRIX_SIZE][MATRIX_SIZE]; // SEH appears here!!!
	uintptr_t i = 0;
	uintptr_t j = 0;
...
```

When I put print statements before and after the method invocation, it becomes very clear that the test stops here. 

Do I right understand the stack overflow exception appears here because `clang-cl` use an 8MB stack by default and 1 000 000 of `__int64` can't be allocated there? I've checked, if the size of the matrix equals to `358 x 358` everything works fine.

I see the following ways to solve the problem:

 * Shrink the `MATRIX_SIZE` constant value
 * Allocate `matrix` on the heap
 * Let `clang-cl` increase the size of the stack using the `-stack_size` linker command argument.

What do you think?

charliegracie:
[...]

PR #2394, opened 2018-03-27
Allocate a matrix in the rastest on the heap

samolisov:
A matrix in the method `static void matrixSquare(void)` is allocated on
the stack:

```c
uintptr_t matrix[MATRIX_SIZE][MATRIX_SIZE]
```

Because the constant `MATRIX_SIZE` is defined as `1000`, the amount of
memory needed to allocate `matrix` equals to `8000000`. This value is more
than the default stack size for the major C/C++ toolchains. In order to
protect the test from the Stack Overflow error, the `matrix` should be
allocated on the heap as discussed in #2386.

Closes: #2386

### DIFF
--- a/fvtest/rastest/cpuLoadAgent.c
+++ b/fvtest/rastest/cpuLoadAgent.c
@@ -27,11 +27,11 @@
 
 #include "omragent.h"
 
-#define MATRIX_SIZE 1000
+#define MATRIX_ORDER 1000
 #define NUM_ITERATIONS_SYSTEM_CPU_BURN	3000
 
 static intptr_t dummy_omrsysinfo_get_CPU_utilization(struct OMRPortLibrary *portLibrary, struct J9SysinfoCPUTime *cpuTimeStats);
-static void matrixSquare(void);
+static void matrixSquare(OMR_VMThread *vmThread);
 static void systemTimeCPUBurn(void);
 static omr_error_t testProcessCpuLoad(OMR_VMThread *vmThread, OMR_TI const *ti, omr_error_t expectedRc, BOOLEAN dummyPort, BOOLEAN checkRc);
 static omr_error_t testSystemCpuLoad(OMR_VMThread *vmThread, OMR_TI const *ti, omr_error_t expectedRc, BOOLEAN dummyPort, BOOLEAN checkRc);
@@ -132,7 +132,7 @@ testProcessCpuLoad(OMR_VMThread *vmThread, OMR_TI const *ti, omr_error_t expecte
 	OMRPORT_ACCESS_FROM_OMRVMTHREAD(vmThread);
 
 	real_omrsysinfo_get_CPU_utilization = OMRPORTLIB->sysinfo_get_CPU_utilization;
-	matrixSquare();
+	matrixSquare(vmThread);
 	systemTimeCPUBurn();
 	if (TRUE == dummyPort) {
 		/* use dummy_omrsysinfo_get_CPU_utilization */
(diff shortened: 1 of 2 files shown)


fvtest/rastest/cpuLoadAgent.c
```diff
@@ -401,15 +402,17 @@ matrixSquare(void)
 {
 	/* Declare and initialize the matrix. memset works here since this is a true 2D
 	 * array defined in the same scope as the memset. */
-	uintptr_t matrix[MATRIX_SIZE][MATRIX_SIZE];
+	uint32_t size_of_matrix = MATRIX_SIZE * MATRIX_SIZE * sizeof(uintptr_t);
+	uintptr_t *matrix = (uintptr_t*) malloc(size_of_matrix);
 	uintptr_t i = 0;
 	uintptr_t j = 0;
-
```
charliegracie: Now that we are using malloc to allocate the memory can we add a NULL check on the result and then just print an error message and return if we fail to allocate?

charliegracie: Both of the callers have an OMR_VMTHread so you can pass that through as the parameter.  Once you have the thread you can insert the following macro to get access to the PortLibrary for printing `OMRPORT_ACCESS_FROM_OMRVMTHREAD(vmThread);`

Now that you have access to the PortLibrary you can use the `omrtty_printf` function to write information in the case where we fail to malloc memory.

samolisov:
*@charliegracie* What is your opinion about whether this method is required
at all? Please see the discussion here #2386 If the method is required,
I'll implement your suggestions.

fvtest/rastest/cpuLoadAgent.c
charliegracie: oh and also you should be using the port library `omrmem_allocate_memory` and `omrmem_free_memory` APIs

charliegracie:
I would keep the method for now.

samolisov:
@charliegracie thank you for the comments. I've applied what you suggested. One moment I don't understand clearly is memory categories, I use the `OMRMEM_CATEGORY_UNKNOWN` one:

```c
uintptr_t *matrix = (uintptr_t*) omrmem_allocate_memory(size_of_matrix,
			OMRMEM_CATEGORY_UNKNOWN);
```
Is it OK?

fvtest/rastest/cpuLoadAgent.c
rwy0717: I would change `MATRIX_SIZE` to `MATRIX_ROWS` (or something) and use the symbol `MATRIX_SIZE` to represent `size_of_matrix`.

samolisov:
I can change MATRIX_SIZE to whatever but MATRIX_ROW says only about a
number of rows but what about columns?

Wikipedia says:  An *n*-by-*n* matrix is known as a square matrix of order
*n*

*What about MATRIX_ORDER?*

2018-03-28 14:33 GMT+03:00 Robert Young <notifications@github.com>:

rwy0717:
This is not the first time we've had issues with CPU burners being optimized away. Maybe we should move the burn function to a separate compilation unit (EG rasTestHelpers.cpp), so the work is less likely to evaporate. We also have other burners in processTimeTest, we should really only have one common implementation, given how finicky they are to implement.

I prefer processTimeTest's implementation, since the atomic ops are less likely to be optimized away.

```c++
void
userTimeCpuBurn()
{
	for (size_t i = 0; i < 40000; i++) {
		VM_AtomicSupport::nop();
	}
}
```

And sure, matrix order sounds fine to me, if we keep the current implementation.

charliegracie:
Do not worry about the Atomic changes.  I think that would be best in another PR.  Thanks for the great work @samolisov

charliegracie:
@genie-omr build all

charliegracie:
Actually I missed one thing.  I missed the copyright headers, which need to be updated to 2018.  Can you push a new commit which updates the 2 copyright headers?  You can put `[ci skip]` in the commit title to skip all CI testing since it is just a documentation change.

samolisov:
@charliegracie Done!

PR #2396, opened 2018-03-27
Fix bug in KnowledgeLevel

liqunl:
KnowledgeLevel returns FIXED_CLASS if getFixedClass, which returns
_fixedClass, is not null. This wrong because _fixedClass and
getFixedClass() in TR_PrexArgument are used to store and get the class
of the argument regardless of its kind. This change will fix
KnowledgeLevel and use _class and getClass() to avoid misleading.

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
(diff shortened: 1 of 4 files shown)


liqunl:
@andrewcraik Can you review this PR? Will post the result of my pbuild once it finishes.

andrewcraik:
looks reasonable - I'd like to hear the performance results for DayTrader3 and Spark HiBench when considering this. While it should prevent inlining an implementation that won't run I do want to be sure we aren't going to regress anything while doing so unless we can justify that in some way.

liqunl:
The performance on DayTrader3 and Spark is on par with the baseline.

vijaysun-omr:
@genie-omr build all

PR #2402, opened 2018-03-28
Disable peeking under HCR

liqunl:
Disable peeking under HCR and introduce
genMethodILForPeekingEvenUnderHCR for optimizations that have protection
when HCR happens.

### DIFF
--- a/compiler/compile/Method.cpp
+++ b/compiler/compile/Method.cpp
@@ -671,3 +671,12 @@ void TR_ResolvedMethod::makeParameterList(TR::ResolvedMethodSymbol *methodSym)
 
    methodSym->setFirstJitTempIndex(methodSym->getTempIndex());
    }
+
+TR::SymbolReferenceTable*
+TR_ResolvedMethod::genMethodILForPeeking(TR::ResolvedMethodSymbol *methodSymbol, TR::Compilation  *comp, bool resetVisitCount, TR_PrexArgInfo  *argInfo)
+   {
+   if (comp->getOption(TR_EnableHCR))
+      return NULL;
+
+   return _genMethodILForPeeking(methodSymbol, comp, resetVisitCount, argInfo);
+   }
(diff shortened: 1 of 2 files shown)


liqunl:
@andrewcraik This is part of the change to reenable peeking globally and let each caller to check if it's safe to peek before calling to genMethodILForPeeking. pbuild has passed.

andrewcraik:
Please improve your pull request and commit message. You are enabling peeking under HCR rather than disabling it. I am not a fan of the nature of this change in general because there is an onus on the caller to do the necessary safety checks. I think it would be more clear to create a peek implementation method and then two public front-end methods normal peeking which has the HCR check (the current name) and one with EvenUnderHCR added to its name that does not have the checks. This then makes it clear which one you want to call and when and why.

liqunl:
I've considered that option. But it looks like HCR is a J9 specific thing, do we want to add a function with its name in OMR? The change in this commit is inside J9_PROJECT_SPECIFIC.

andrewcraik:
The new public entry points can be inside the same macro for now until they are refactored no?

liqunl:
Although HCR is only used in Java, this technique can be brought into other languages as well if they want. I'm proposing a function `genMethodILForPeekingEvenUnderHCR` in OMR to allow optimizations to peek when it's safe for them. @0xdaryl What do you think?

andrewcraik:
@0xdaryl So I think the notion of method redefinition is not OpenJ9 specific - many languages want to support methods being redefined so as a concept that is a useful thing to have in OMR. We could certainly change the EvenUnderHCR to EvenUnderMethodRedefinition. The logic for checking for method redefinition should not, in my opinion, be scattered across optimizations - at a given point you know whether peeking is safe or not under redefinition and so the proposal is to just have a call to express that intent to provide a single place to implement the check and to make the code in the optimizations more readable. Thoughts?

0xdaryl:
OK.  EvenUnderMethodRedefinition is better than HCR.  New functions in OMR should have Doxygen comments describing what they do and their parameters.

andrewcraik:
Sounds good @liqunl can you do the rename and add the comments requested please?

liqunl:
@0xdaryl @andrewcraik Changes have been made as per request.

0xdaryl:
@genie-omr build all

PR #2406, opened 2018-03-29
Fix build problems when cross-configuring and default compilers can't be found

ymanton:
These patches address problems that show up when autoconf can't find known compilers (e.g. `gcc`) in the current `PATH`. Running `configure` will currently fail in such circumstances.

`OMR_CROSS_CONFIGURE` is used to indicate that `configure` is not being run on the build system. When set to `yes` the user should use `CC` and `CXX` to point to the compilers that are to be used and set various other parameters (e.g. `OMR_TARGET_DATASIZE`), since we can't derive them on the non-build system. The current implementation will (rightfully) not try to invoke `CC` and `CXX` when cross-configuring but will then go on to look for known compilers in `PATH` and proceed to invoke and derive (useless, potentially incorrect) facts from them. With these changes we'll no longer run such tests when cross-configuring and instead check/validate that the user has supplied them. This allows building on systems where default compilers are not in the `PATH` and also allows `configure` to run a tad bit faster.

These changes also introduce the `OMR_TOOLS_CC` and `OMR_TOOLS_CXX` variables to allow the user to control which compilers are used to build the OMR tools when cross-compiling. The current implementation requires that known compilers exist in `PATH`.

:warning: These changes require that downstream projects that set `OMR_CROSS_CONFIGURE=yes` and cross-compile make changes to how they invoke `configure`; they'll need to specify their tool compilers and data size via `OMR_TOOLS_CC`, `OMR_TOOLS_CXX`, `OMR_BUILD_DATASIZE`, and `OMR_BUILD_TOOLCHAIN`. These changes can be made before this PR is merged to avoid breaking builds.


0xdaryl:
@rwy0717 , would you mind reviewing this PR please?

Apologies for the delay @ymanton--this seems to have slipped through the cracks.

ymanton:
OpenJ9 no longer uses `OMR_CROSS_CONFIGURE` so my original motivation for fixing this is somewhat diminished to be honest but I'll leave it open anyway if the project feels it's worth fixing.

rwy0717:
LGTM, and probably worth merging. Thanks @ymanton :+1:

0xdaryl:
@genie-omr build all

ymanton:
@0xdaryl I'd like to fix downstream before this is merged then. I'll comment when that's done.

ymanton:
Downstream has been fixed. Rebased and removed the WIP.

0xdaryl:
@genie-omr build all

ymanton:
Ping.

jjohnst-us:
@ymanton Hoping to get a little assistance on this one.  I believe this PR broke our builds.  I notice that our omr/tools/hookgen/HookGen.cpp no longer compiles because the compiler variable OMR_BUILD_CXX is no longer being set to 'gcc'.

I notice in your description that cross-configuration builds need to start setting OMR_TOOLS_CXX.  However if I set that, it doesn't seem to take.  I update the following file:
glue/configure_includes/configure_linux_ztpf_390.mk
CONFIGURE_ARGS += 'OMR_TOOLS_CC=gcc'
thanks for any guidance

jjohnst-us:
@yamaton Thanks, once I was updating the wrong file need to update openj9 version of configure*.mk no the omr version.  Once I did that was able to set the variables appropriately.
Thank you!

PR #2408, opened 2018-03-31
Relocatable Debug Counters

dsouzai:
Relocatable Debug Counter Support

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -86,6 +86,7 @@
 #include "il/symbol/LabelSymbol.hpp"                // for LabelSymbol
 #include "il/symbol/RegisterMappedSymbol.hpp"
 #include "il/symbol/ResolvedMethodSymbol.hpp"
+#include "il/symbol/StaticSymbol.hpp"
 #include "infra/Array.hpp"                          // for TR_Array
 #include "infra/Assert.hpp"                         // for TR_ASSERT
 #include "infra/Bit.hpp"                            // for isEven, etc
@@ -3038,10 +3039,19 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(TR::Instruction *curso
       return cursor;
    if (delta == 0)
       return cursor;
-   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp());
+   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp(), name);
    aggregatedCounters->aggregateStandardCounters(self()->comp(), cursor->getNode(), name, delta, fidelity, staticDelta);
    if (!aggregatedCounters->hasAnyCounters())
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !aggregatedCounters->initializeReloData(self()->comp(), delta, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = aggregatedCounters->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, aggregatedCounters);
+
    return self()->generateDebugCounterBump(cursor, aggregatedCounters, 1, NULL);
    }
 
@@ -3055,6 +3065,15 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(TR::Instruction *curso
    TR::DebugCounter *counter = TR::DebugCounter::getDebugCounter(self()->comp(), name, fidelity, staticDelta);
    if (!counter)
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !counter->initializeReloData(self()->comp(), 0, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = counter->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, counter);
+
    return self()->generateDebugCounterBump(cursor, counter, deltaReg, NULL);
    }
 
@@ -3066,10 +3085,19 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(const char *name, TR::
       return cursor;
    if (delta == 0)
       return cursor;
-   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp());
+   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp(), name);
    aggregatedCounters->aggregateStandardCounters(self()->comp(), cursor->getNode(), name, delta, fidelity, staticDelta);
    if (!aggregatedCounters->hasAnyCounters())
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !aggregatedCounters->initializeReloData(self()->comp(), delta, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = aggregatedCounters->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, aggregatedCounters);
+
    return self()->generateDebugCounterBump(cursor, aggregatedCounters, 1, &cond);
    }
 
@@ -3083,6 +3111,15 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(const char *name, TR::
    TR::DebugCounter *counter = TR::DebugCounter::getDebugCounter(self()->comp(), name, fidelity, staticDelta);
    if (!counter)
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !counter->initializeReloData(self()->comp(), 0, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = counter->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, counter);
+
    return self()->generateDebugCounterBump(cursor, counter, deltaReg, &cond);
    }
 
@@ -3094,10 +3131,19 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(const char *name, TR_S
       return cursor;
    if (delta == 0)
       return cursor;
-   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp());
+   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp(), name);
    aggregatedCounters->aggregateStandardCounters(self()->comp(), cursor->getNode(), name, delta, fidelity, staticDelta);
    if (!aggregatedCounters->hasAnyCounters())
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !aggregatedCounters->initializeReloData(self()->comp(), delta, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = aggregatedCounters->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, aggregatedCounters);
+
    return self()->generateDebugCounterBump(cursor, aggregatedCounters, 1, srm);
    }
 
@@ -3111,6 +3157,15 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(const char *name, TR::
    TR::DebugCounter *counter = TR::DebugCounter::getDebugCounter(self()->comp(), name, fidelity, staticDelta);
    if (!counter)
       return cursor;
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()) && !counter->initializeReloData(self()->comp(), 0, fidelity, staticDelta))
+      return cursor;
+
+   TR::SymbolReference *symref = counter->getBumpCountSymRef(self()->comp());
+
+   if (TR::DebugCounter::relocatableDebugCounter(self()->comp()))
+      self()->comp()->mapStaticAddressToCounter(symref, counter);
+
    return self()->generateDebugCounterBump(cursor, counter, deltaReg, srm);
    }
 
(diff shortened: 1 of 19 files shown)


compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3038,10 +3039,20 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(TR::Instruction *curso
       return cursor;
    if (delta == 0)
       return cursor;
-   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp());
+   TR::DebugCounterAggregation *aggregatedCounters = self()->comp()->getPersistentInfo()->getDynamicCounters()->createAggregation(self()->comp(), name);
    aggregatedCounters->aggregateStandardCounters(self()->comp(), cursor->getNode(), name, delta, fidelity, staticDelta);
    if (!aggregatedCounters->hasAnyCounters())
       return cursor;
+
+   if (self()->comp()->compileRelocatableCode() && !aggregatedCounters->getReloData(self()->comp(), delta, fidelity, staticDelta))
```
mstoodle: you're going to hate me, but I would prefer you create a new query here corresponding to debug counters need to be relocated that itself can just return `compileRelocatableCode()` . That will make it easier to identify this set of code changes later on.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3055,6 +3066,16 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(TR::Instruction *curso
    TR::DebugCounter *counter = TR::DebugCounter::getDebugCounter(self()->comp(), name, fidelity, staticDelta);
    if (!counter)
       return cursor;
+
+   if (self()->comp()->compileRelocatableCode() && !counter->getReloData(self()->comp(), 0, fidelity, staticDelta))
+      return cursor;
```
mstoodle: Just want to confirm that this case means: we're compiling relocatable code but don't know how to relocate this particular requested debug counter, so it silently just doesn't insert the counter?

compiler/codegen/Relocation.cpp
```diff
@@ -391,7 +391,7 @@ const char *TR::ExternalRelocation::_externalRelocationTargetKindNames[TR_NumExt
    "TR_RamMethodSequenceReg (23)",
    "TR_VerifyClassObjectForAlloc (24)",
    "TR_ConstantPoolOrderedPair (25)",
-   "TR_AbsoluteMethodAddressOrderedPair (36)",
+   "TR_AbsoluteMethodAddressOrderedPair (26)",
```
mstoodle: good catch :)

compiler/codegen/Relocation.cpp
```diff
@@ -418,11 +418,12 @@ const char *TR::ExternalRelocation::_externalRelocationTargetKindNames[TR_NumExt
    "TR_ValidateArbitraryClass (50)",
    "TR_EmitClass (51)",
    "TR_JNISpecialTargetAddress (52)",
-   "TR_VirtualRamMethodConst (53)"
+   "TR_VirtualRamMethodConst (53)",
    "TR_InlinedInterfaceMethod (54)",
    "TR_InlinedVirtualMethod (55)",
    "TR_NativeMethodAbsolute (56)",
    "TR_NativeMethodRelative (57)",
+   "TR_DebugCounter (58)"
```
mstoodle: also good catch :)

compiler/compile/OMRCompilation.hpp
```diff
@@ -676,6 +676,36 @@ class OMR_EXTENSIBLE Compilation
 
    void setIlVerifier(TR::IlVerifier *ilVerifier) { _ilVerifier = ilVerifier; }
 
+   typedef std::pair<void *, TR::DebugCounterBase *> DebugCounterEntry;
+   typedef TR::typed_allocator<DebugCounterEntry, TR::Allocator> DebugCounterMapAllocator;
+   typedef std::map<void *, TR::DebugCounterBase *, std::less<void *>, DebugCounterMapAllocator> DebugCounterMap;
```
mstoodle: please make the key type "const" to avoid compile failures on more recent compilers

compiler/compile/OMRCompilation.hpp
```diff
@@ -676,6 +676,36 @@ class OMR_EXTENSIBLE Compilation
...
+    * @param staticAddress the static address of the memory used to store the count value of the debug counter
+    * @return the debug counter associated with the static address
+    */
+   TR::DebugCounterBase *getCounterFromStaticAddress(void *staticAddress);
+
+   /**
+    * @brief getDebugCounterMap
+    * @return reference to the TR::DebugCounterMap map
+    */
+   DebugCounterMap &getDebugCounterMap() { return _debugCounterMap; }
```
mstoodle: You should not make this implementation detail public

compiler/compile/OMRCompilation.hpp
```diff
@@ -676,6 +676,36 @@ class OMR_EXTENSIBLE Compilation
...
+    * @brief mapStaticAddressToCounter
+    * @param staticAddress the static address of the memory used to store the count value of the debug counter
+    * @param counter pointer to the debug counter
+    *
+    * Maps the staticAddress to the debug counter. The map has to be between the static address and the
+    * debug counter, instead of the symref of the static address and the debug counter, because sometimes,
+    * transformations can create a new symref that will not exist in this map; however because the static
+    * address is constant, it is guaranteed to exist in the map.
+    */
+   void mapStaticAddressToCounter(void *staticAddress, TR::DebugCounterBase *counter);
```
mstoodle: Do you ever need this two functions in a place where you do not have the symbol reference?

If not, then I suggest changing the API to take a symbol reference, but assert if it isn't a static symbol reference and dig out the static address here as the key for the map.

compiler/x/codegen/OMRX86Instruction.cpp
```diff
@@ -1153,7 +1153,11 @@ void
 TR::X86RegImmSymInstruction::autoSetReloKind()
    {
    TR::Symbol *symbol = getSymbolReference()->getSymbol();
-   if (symbol->isConst())
+   if (getSymbolReference()->isDebugCounter())
```
mstoodle: minor nit, since nobody else who changed this code really thought about it, but putting this relatively rare case condition at the beginning if the if chain isn't the best performance choice :) .

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3055,6 +3066,16 @@ TR::Instruction *OMR::CodeGenerator::generateDebugCounter(TR::Instruction *curso
...
+
+   TR::SymbolReference *symref = counter->getBumpCountSymRef(self()->comp());
+   symref->setIsDebugCounter();
```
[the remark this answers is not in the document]
dsouzai: Heh yeah that is a better way of doing things, I'll make the appropriate changes.

compiler/codegen/OMRCodeGenerator.cpp
dsouzai: What it actually means is "we failed to allocate memory we need to store information about this counter at relo time". I chose the "silently fail to insert counter" path because of the previous check, which is also effectively a check to see if we managed to allocate the counter.

compiler/codegen/Relocation.cpp
dsouzai: Took 30 minutes to figure out why the logs weren't printing out "TR_DebugCounter (58)"...

compiler/compile/OMRCompilation.hpp
dsouzai: > Do you ever need this two functions in a place where you do not have the symbol reference?

No, I believe in every place that this function is called, I get the static address from the symref. I'll update the code as per your suggestion.

dsouzai:
@mstoodle Made the requested changes.

compiler/p/codegen/OMRMemoryReference.cpp
```diff
@@ -1522,7 +1522,13 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
          if (!GCRnode)
             GCRnode = cg->getCurrentEvaluationTreeTop()->getNode();
```
mstoodle: It doesn't seem like "GCRnode" is the right name for this variable anymore...

compiler/p/codegen/OMRMemoryReference.cpp
```diff
@@ -1605,7 +1611,13 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
...
```
mstoodle: same comment here

compiler/p/codegen/OMRMemoryReference.cpp
```diff
@@ -1699,11 +1711,26 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
 
          if (comp->getOption(TR_AOT))
```
mstoodle: i know it's not your fault, but this isn't really the right condition to use :(

compiler/p/codegen/OMRMemoryReference.cpp
dsouzai: The comment states `/* don't want to trash node prematurely by the code for handling guarded counting recompilations symbols */` , so I'm not sure what to rename the variable to then :S

compiler/p/codegen/OMRMemoryReference.cpp
dsouzai: Same concern as voiced in previous comment.

compiler/p/codegen/OMRMemoryReference.cpp
dsouzai: Heh, if that makes you sad, you should probably not grep for all the other instances of that query ;P. At any rate I'll fix it to use the canonical `compileRelocatableCode()` query.

dsouzai:
@mstoodle Made most of the requested changes; didn't know what to rename the `GCRnode` variable (as described in the comments above).

compiler/p/codegen/OMRMemoryReference.cpp
mstoodle: Sorry, I missed this update before.

So we can't correct the variable name because it's indirectly referenced in a comment :) ? The comment is already wrong! Let's choose not to perpetuate a conservation of confusion law :) .

My point was that `GCRNode` seems to be now used to hold all kinds of different nodes in this function, not *just* nodes that are related to GCR. So why not use a more generic name like `nodeForSymbol` or something like that? The comment can similarly be made more generic "...the code for handling symbol relocations" or some such.

dsouzai:
@mstoodle Does everything look ok to you now?

compiler/ras/DebugCounter.cpp
```diff
@@ -375,10 +478,36 @@ TR::DebugCounter *TR::DebugCounterGroup::findCounter(const char *nameChars, int3
...
-TR::DebugCounterAggregation *TR::DebugCounterGroup::createAggregation(TR::Compilation *comp)
+TR::DebugCounterAggregation *TR::DebugCounterGroup::findAggregation(const char *nameChars, int32_t nameLength)
+   {
+   if (nameChars == NULL)
+      return NULL;
+   char *name = (char*)alloca(nameLength+1);
+   strncpy(name, nameChars, nameLength);
+   name[nameLength] = 0;
+
+   // There's a race here if we do parallel compilation
```
mstoodle: Don't you mean there would be a race if we do parallel compilation? The critical region prevents the race, does it not? Or is there another race here you're referring to?

compiler/ras/DebugCounter.cpp
```diff
@@ -375,10 +478,36 @@ TR::DebugCounter *TR::DebugCounterGroup::findCounter(const char *nameChars, int3
...
+   }
+
+
+TR::DebugCounterAggregation *TR::DebugCounterGroup::createAggregation(TR::Compilation *comp, const char * name)
    {
-   TR::DebugCounterAggregation *aggregatedCounters = new (comp->trPersistentMemory()) TR::DebugCounterAggregation(comp->trMemory());
+   TR::DebugCounterAggregation *aggregatedCounters = new (comp->trPersistentMemory()) TR::DebugCounterAggregation(name, comp->trPersistentMemory());
    _aggregations.add(aggregatedCounters);
+
+   // There's a race here if we do parallel compilation
```
mstoodle: same as above comment

compiler/x/codegen/OMRX86Instruction.cpp
```diff
@@ -4003,7 +4007,11 @@ void
 TR::AMD64RegImm64SymInstruction::autoSetReloKind()
    {
    TR::Symbol *symbol = getSymbolReference()->getSymbol();
-   if (symbol->isConst())
+   if (getSymbolReference()->isDebugCounter())
+      {
```
mstoodle: could you make this consistent with the code below it (one way or the other) ?

compiler/ras/DebugCounter.cpp
dsouzai: I added that line because there's a similar comment here https://github.com/eclipse/omr/pull/2408/files/4f5f0fa137ffcb00f92a01f91654835fc6ff07bb#diff-d89c18db7b5f8e5d8da1080cb3818635R563

I guess the comment describes why there's a critical section

compiler/ras/DebugCounter.cpp
dsouzai: Same as above.

dsouzai:
@mstoodle Changes made

dsouzai:
@mstoodle ok, comments removed, this should be good to go (pending any genie builds :) )

mstoodle:
@genie-omr build all

dsouzai:
ugh, sorry, gotta change a couple of lines. I called `isNotDataAddress` instead of `setNotDataAddress`

dsouzai:
This is the diff of the commit i'll update; not sure if you'll want to kick off genie builds again.

```
diff --git a/compiler/compile/OMRSymbolReferenceTable.cpp b/compiler/compile/OMRSymbolReferenceTable.cpp
index b38e034a..1f5b162a 100644
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -1799,7 +1799,7 @@ OMR::SymbolReferenceTable::findOrCreateCounterSymRef(char *name, TR::DataType d,
    // No match
    //
    TR::StaticSymbol * sym = TR::StaticSymbol::createNamed(trHeapMemory(), d,address,name);
-   sym->isNotDataAddress();
+   sym->setNotDataAddress();
    result = new (trHeapMemory()) TR::SymbolReference(self(),sym);
    _debugCounterSymbolRefs.add(result);
    return result;
diff --git a/compiler/ras/DebugCounter.cpp b/compiler/ras/DebugCounter.cpp
index 9873658d..9e878945 100644
--- a/compiler/ras/DebugCounter.cpp
+++ b/compiler/ras/DebugCounter.cpp
@@ -439,7 +439,7 @@ TR::SymbolReference *TR::DebugCounterAggregation::getBumpCountSymRef(TR::Compila
       TR::StaticSymbol *symbol = TR::StaticSymbol::create(_mem,TR::Int64);
       TR_ASSERT(symbol, "StaticSymbol *symbol must not be null. Ensure availability of persistent memory");
       symbol->setStaticAddress(&_bumpCount);
-      symbol->isNotDataAddress();
+      symbol->setNotDataAddress();
       _symRef = new (_mem) TR::SymbolReference(comp->getSymRefTab(), symbol);
       TR_ASSERT(_symRef, "SymbolReference *_symRef must not be null. Ensure availability of persistent memory");
       symbol->setIsDebugCounter();
```

mstoodle:
@genie-omr build all

dsouzai:
My personal builds also passed, this should be good to merge.

mstoodle:
@dsouzai will this PR require any coordination of merging with https://github.com/eclipse/openj9/pull/1589 ?

dsouzai:
Nope, the OMR changes are self consistent, the OpenJ9 changes though need these changes in openj9-omr before they can be merged.

PR #2409, opened 2018-03-31
Silence ignored-return warnings

DanHeidinga:
This is a minimal change to remove the warnings when building. A
proper fix requires a deeper redesign of the code as not all of
these functions return a value and the callers aren't prepared
to handle the failure / error cases.

Silencing the warning seems like a pragmatic step towards less
noise in the builds.

### DIFF
--- a/port/unix/omrintrospect.c
+++ b/port/unix/omrintrospect.c
@@ -293,7 +293,9 @@ barrier_release_r(barrier_r *barrier, uintptr_t seconds)
 		 * converts this to a busy wait
 		 */
 		barrier->released = 1;
-		write(barrier->descriptor_pair[1], &byte, 1);
+		if (write(barrier->descriptor_pair[1], &byte, 1) != 1) {
+			/* Silence warning: needs a redesign to do better */
+		}
 	}
 
 	/* wait until all entrants have arrived */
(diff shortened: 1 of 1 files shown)


charliegracie:
Which platform / compiler version were you using when you saw this warning?  I know it is a legitimate warning but I thought that warning was disabled for this file.

DanHeidinga:
I've seen it in OpenJ9 builds using the ubuntu 16.04 docker images with gcc 4.8

rwy0717:
If the write fails, should we be treating it as a fatal error? Can we just assert the return value is as-expected?

DanHeidinga:
I don't think we can assert it given that some of the functions allow timeouts - ie:
```
It will block until all expected entrants have called this
function and barrier_release_r has been called, or the deadline has expired.
```

charliegracie:
For the few files/functions that contain these calls I think I would rather disable the warning explicitly than have the empty code blocks.

rwy0717:
Can we just cast to void?

charliegracie:
@DanHeidinga have you had a chance to look into this?  Thanks

DanHeidinga:
Casting to (void) doesn't fix the warnings by design according to gcc bug reports.

DanHeidinga:
For fwrite, glibc has removed the `_wur` (warn unused return) according to https://sourceware.org/bugzilla/show_bug.cgi?id=11959 so maybe this will be automatically cleaned up in newer glibc levels.

charliegracie:
I am ok with leaving these warnings for now and if it becomes an issue in the future we can try to do something else.

PR #2410, opened 2018-03-31
ASSERT before dereferencing asserted variable

DanHeidinga:
Adjusted code to do the ASSERT() before dereferencing the
variable.  This way the assert will fire before the code
will crash.


charliegracie:
@genie-omr build all

PR #2412, opened 2018-04-02
Add comments to JIT BinaryCommutativeAnalyser files for ARM

knn-k:
BinaryCommutativeAnalyser.hpp and .cpp for ARM are not currently used
from anywhere.  This commit adds comments to those files instead of
removing them entirely from the repository.


knn-k:
This PR came from the discussion in https://github.com/eclipse/openj9/pull/1571 .

vijaysun-omr:
@JBKingdon and @0xdaryl

JamesKingdon:
Many thanks Konno-san.

JamesKingdon:
Background for committer: we have a significant performance delta to track down on ARM, so I'd prefer we didn't delete code until we've had time to review and confirm that there is no reasonable opportunity to benefit from it.

Issue #2332, opened 2018-02-22
Improving/Extending OMR Signal API


babsingh:
The objective is to improve and extend the existing OMR Signal API so that signal support can be added to OpenJ9.

**Current OMR signal API implementation**: `omrsig_set_async_signal_handler`. In OMR, a list of `J9*AsyncHandlerRecord` a.k.a `asyncHandlerList` is maintained.

```
-- omrsig_set_async_signal_handler(...,  omrsig_handler_fn handler, void *handler_arg, uint32_t flags)
---- registerMasterHandlers
------ registerSignalHandlerWithOS //call to sigaction
```

```
typedef struct J9*AsyncHandlerRecord {
	OMRPortLibrary *portLib;
	omrsig_handler_fn handler;
	void *handler_arg;
	uint32_t flags;
	struct J9*AsyncHandlerRecord *next;
} J9*AsyncHandlerRecord;
```

Whenever `omrsig_set_async_signal_handler` is invoked, we look for a record associated to the specified handler (`omrsig_handler_fn`) in `asyncHandlerList`.

If a `J9*AsyncHandlerRecord` exists in `asyncHandlerList`, we update `flags` to associate the record with a signal. If no record exists, then we add a new record for the handler in `asyncHandlerList` and set `flags` appropriately. Multiple `J9*AsyncHandlerRecord`s can be associated with a signal. On Unix, we register handlers with the OS using `sigaction`. On Windows, we don't register signals with the OS using `signal`; it is assumed signals will only be sent using `Ctrl + ...` so `SetConsoleCtrlHandler` is used to set an handler (`consoleCtrlHandler`).

When an async signal is received, `masterASynchSignalHandler` is invoked by the signal on Linux/Unix platfroms. This function increments signal counts and wakes up the `asynchSignalReporter` thread. `asynchSignalReporter` thread decrements the signal count, finds all records in `asyncHandlerList` for the raised signal and invokes the `handler` in each record.

On Windows, handlers are not registered with the OS, and there is no thread to invoke handlers. The OS invokes `consoleCtrlHandler` which was set using `SetConsoleCtrlHandler`. `consoleCtrlHandler` finds all records in asyncHandlerList for the raised signal and invokes the handler in each record.

**What needs to be done in OMR?**

[...]

PR #2413, opened 2018-04-02
Add stubs for new omrsig functions

babsingh:
Stubs have been added for the following new `omrsig` functions:
1) `omrsig_set_single_async_signal_handler`
2) `omrsig_map_os_signal_to_portlib_signal`
3) `omrsig_map_portlib_signal_to_os_signal`

File `omrsig.c` doesn't exist. Instead, functions are defined in 
`omrsignal.c`. Update comments to use `omrsignal.c` instead of 
`omrsig.c`.

OMR Issue: https://github.com/eclipse/omr/issues/2332

### DIFF
--- a/port/common/omrsignal.c
+++ b/port/common/omrsignal.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2015 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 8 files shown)


charliegracie:
@genie-omr build all zos

port/common/omrsignal.c
```diff
@@ -166,11 +166,59 @@ omrsig_protect(struct OMRPortLibrary *portLibrary,  omrsig_protected_fn fn, void
...
+ *
+ * @param[in] portLibrary The port library
+ * @param[in] osSignalValue OS signal value
+ *
+ * @return port library signal flag on success and 0 on failure
+*/
+uint32_t
+omrsig_map_os_signal_to_portlib_signal(struct OMRPortLibrary *portLibrary, uint32_t osSignalValue)
+{
+	return 0;
```
charliegracie: Is `0` a signal value or are they all guaranteed to be `>` 0?

babsingh: If I execute `kill -l`, all signals are `>` than zero. I don't think I should use `0` for error since `SIG_ERR` is negative. I will change it to a negative number.

babsingh: `omrsig_map_os_signal_to_portlib_signal` returns a `port library signal flag`. a signal flag is used for bit operations so a signal flag will never be 0. so, 0 is acceptable as an error for `omrsig_map_os_signal_to_portlib_signal`.

`omrsig_map_portlib_signal_to_os_signal` returns an `OS signal value`. for this function, I changed the error to a negative value.

PR #2416, opened 2018-04-03
Fix for missing data snippet in trace log

0dvictor:
Data snippet was not properly printed in trace
log, fixing it.

### DIFF
--- a/compiler/x/codegen/DataSnippet.cpp
+++ b/compiler/x/codegen/DataSnippet.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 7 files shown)


compiler/x/codegen/DataSnippet.cpp
```diff
@@ -105,12 +105,21 @@ TR_Debug::print(TR::FILE *pOutFile, TR::IA32DataSnippet * snippet)
...
+      case 4:
+         trfprintf(pOutFile, "%s \t%s", ddString(), hexPrefixString());
+         break;
+      case 2:
+         trfprintf(pOutFile, "%s \t%s", dwString(), hexPrefixString());
+         break;
+      default:
+         trfprintf(pOutFile, "%s \t%s", dbString(), hexPrefixString());
+         break;
+      }
```
0xdaryl: Could we save some DLL code size here by rewriting this something like:
```
const char *toString;
switch (snippet->getDataSize())
   {
   case 8:  toString = dqString(); break;
   case 4:  toString = ddString(); break;
   // etc
   }
trfprintf(pOutFile, "%s \t%s", toString, hexPrefixString());
```

0dvictor: Sure, it has been updated.

PR #2417, opened 2018-04-03
Optimized fneg and dneg evaluator on X86

0dvictor:
Reduced one instruction (out of two instructions)
in generated code for fneg and dneg.


compiler/x/codegen/FPTreeEvaluator.cpp
```diff
@@ -591,6 +591,63 @@ TR::Register *OMR::X86::TreeEvaluator::fpBinaryArithmeticEvaluator(TR::Node
       }
    }
 
+TR::Register *OMR::X86::TreeEvaluator::fpNegEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   static uint8_t MASKOFFLOAT[] =
```
0xdaryl: Ideally, these arrays would be forced 16-byte aligned to improve the instruction that consumes them as a 16-byte quantity.  The C++11 `alignas` specifier would do the trick, but sadly not supported on all our build compilers. :-(

0xdaryl: Oops.  Wrong PR, but it applies to this one too.

0dvictor: Agreed. Once all our build compilers have the proper support, we should update these arrays with `alignas`.

0dvictor:
Yes, they are nearly identical. Let me mark this as WIP and common the two logics once PR #2415 is merged.

0dvictor:
I have updated this PR to adopt the common logic introduced by PR #2415.

0xdaryl:
This is an x86-only change, so a genie build is not required.

PR #2418, opened 2018-04-04
Add non-null constraint for indirect load of this$0 field from trusted classes

yanluo7:

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -2650,10 +2650,39 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
       {
       vp->comp()->fej9()->markHotField(vp->comp(), node->getSymbolReference(), base->getClass(), base->isFixedClass());
       }
+
+   if (node->getSymbolReference())
+      {
+      if (!node->getSymbolReference()->isUnresolved() &&
+          (node->getSymbolReference()->getSymbol()->getKind() == TR::Symbol::IsShadow) &&
+          (node->getSymbolReference()->getCPIndex() >= 0) &&
+          node->getSymbolReference()->getOwningMethod(vp->comp()))
+         {
+         int32_t fieldSigLen;
+         const char *fieldSig = node->getSymbolReference()->getOwningMethod(vp->comp())->fieldSignatureChars(
+                                      node->getSymbolReference()->getCPIndex(), fieldSigLen);
+         int32_t fieldNameLen;
+         const char *fieldName = node->getSymbolReference()->getOwningMethod(vp->comp())->fieldNameChars(
+                                      node->getSymbolReference()->getCPIndex(), fieldNameLen);
+         char *className = node->getSymbolReference()->getOwningMethod(vp->comp())->classNameChars();
+         uint16_t classNameLen = node->getSymbolReference()->getOwningMethod(vp->comp())->classNameLength();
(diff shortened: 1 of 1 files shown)


yanluo7:
Performance implication:

An instance of an inner class (i.e. non-static) has a hidden field named `this$0` that refers to the enclosing class instance associated with this inner class instance. It behaves much like a instance member of the enclosing class. Because of these properties the hidden field is always non-null -- otherwise the inner class can never be instantiated in the first place. It was discovered that during Explicit New Initialization pass, certain field stores are prevented from being recognized as valid stores because of null checks on the `this$0` field, as null checks are considered GC points. This change adds a non-null constraint on the hidden field if they are coming from trusted packages in `java/util`.

This along with another openj9 change in Explicit New Initialization showed about 0.3-0.4% throughput improvement for Day Trader 3 benchmark.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2653,10 +2653,39 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+         int32_t fieldSigLen;
+         const char *fieldSig = node->getSymbolReference()->getOwningMethod(vp->comp())->fieldSignatureChars(
+                                      node->getSymbolReference()->getCPIndex(), fieldSigLen);
+         int32_t fieldNameLen;
+         const char *fieldName = node->getSymbolReference()->getOwningMethod(vp->comp())->fieldNameChars(
+                                      node->getSymbolReference()->getCPIndex(), fieldNameLen);
+         char *className = node->getSymbolReference()->getOwningMethod(vp->comp())->classNameChars();
+         uint16_t classNameLen = node->getSymbolReference()->getOwningMethod(vp->comp())->classNameLength();
+         if (fieldName && !strncmp(fieldName, "this$0", 6) &&
+              className && !strncmp(className, "java/util/", 10))
```
fjeremic: This seems like very Java specific code that should IMO belong in OpenJ9.

yanluo7: Moved Java specific code inside  `J9_PROJECT_SPECIFIC` block as @0xdaryl suggested.

vijaysun-omr:
I am to reconcile the comment 

"Because of these properties the hidden field is always non-null -- otherwise the inner class can never be instantiated in the first place."

with the fact that the optimization only happens for trusted (java/util/ as per the description). 

Why do we need to restrict the optimization if the hidden field is in fact always non-null ?

yanluo7:
@vijaysun-omr We wanted to restrict this to trusted class as user program could theoretically contain artificially engineered classes with field named `this$0` (as apposed to `javac` a valid inner-class); under such cases we cannot be sure the field is the hidden field we are targeting. We could extend this to other packages under `java/` or even `sun/`, but from observation the vast majority of cases happen under `java/util` (including `java/util/concurrent`) package due to its prolific use of inner classes such as various `Node` and `Iterator`, so we restrict it, conservatively, to `java/util`.

yanluo7:
@0xdaryl In the original commit, strictly speaking the added logic does not need to be under the `else` statement, because currently with no logic in VP to analyze a load of `this$0`, I believe the node should not have `isNonNull()` or `isNull()` set. So moving it up under project-specific code should be functionally equivalent.

vijaysun-omr:
@0xdaryl given the answer to your last question, do you approve of this PR ?

vijaysun-omr:
@genie-omr build all

PR #2419, opened 2018-04-04
Enable DDR on Windows

keithc-ca:
Most of the changes here are to `PdbSCanner`, the Windows-specific debug information extraction code:
* avoid premature typedef expansion
* fix handling of blacklisted types and namespaces
* add explicit cases for char and wchar_t (treat char as unsigned like other platforms)
* capture values of literals
* fix handling of anonymous types and namespaces
* fix postponed type resolution
* fix handling of nested blacklisted types
* fix nested anonymous enum types (promote literals of anonymous enum to enclosing scope)
* removed duplicate, unused getName() function
* removed redundant getUDTname() function
* removed renameAnonymousTypes() functions which are no longer useful
* add MSDIA120 to the list of supported libraries
* remove broken progress reporting
* general cleanup surrounding error conditions and propagation of errors
* fix printing of type sizes (Visual Studio doesn't support printf format "%zu")
* merge grep filtering into sed script
* use the same DDR commands on all platforms

Constant discover script `getmacros` doesn't work well with mutual including header files - needed to adjust the include order in dllmain.c to avoid losing constants from thrtypes.h.

Some clean up was required to GC header files that were missing proper forward declarations.

### DIFF
--- a/ddr/include/ddr/scanner/pdb/PdbScanner.hpp
+++ b/ddr/include/ddr/scanner/pdb/PdbScanner.hpp
@@ -28,10 +28,10 @@
 #include <vector>
 
 #if defined(WIN32) || defined(WIN64)
-/* windows.h defined uintptr_t.  Ignore its definition */
+/* windows.h defines UDATA: Ignore its definition. */
 #define UDATA UDATA_win_
 #include "dia2.h"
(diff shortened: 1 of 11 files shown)


charliegracie:
@genie-omr build all zos

PR #2420, opened 2018-04-04
WIP: Concurrent Card Cleaning - Uncommitted Range Access Fix

RSalman:
Fix to resolve a bug that occurs when ranges belonging to uncommitted
card table memory are found to be valid accessible ranges during card cleaning.

### Bug
When determining card table ranges to be cleaned, ranges belonging to uncommitted card table memory were found to be valid cleanable ranges. This issue occurred when card cleaning took place in the middle of an expansion. Specifically, after the heap was expanded and added to the region manager, but before the card table was expanded (to accommodate for heap expansion). 

This left the card table cleaning operation with an inconsistent view of the heap structures as all structures were assumed to be expanded for regions found in the region manager. However, with the current implementation, the region manager is updated before the expansion of all the structures _(can't easily be rearranged given the current implementation)_, and as a result the region is seen as a valid memory space to clean  (even though card table has yet to be expanded). 

### Fix
A `resizingLock` has been introduced, which is used as follows:
- Lock before starting `collectorExpand`
- Lock before adding a card table range for cleaning

This will block card cleaning (during concurrent `determineCleaningRanges` and `collectorExpand`) until all structures of a newly expanded region are expanded. Hence, this ensures that the card table has been expanded for the range to be cleaned.


amicic:
Create an Issue, and move the Bug details there. Refer to issue in the body of the PR.

Be more specific, in that this only occurs with Tenure heap expansion from Scavenger, and also that the problem only occurs with Concurrent Scavenger overlapping with Concurrent Marking. This is only (known) way that Tenure heap expansion can race with Concurrent Card Cleaning.

Then, make the lock conditional on Concurrent Scavenger and/or Concurrent Marking. I think (but subject to discussion), definitions should be compile time guarded, and call sites should be both compile time and runtime (using isConcurrentScavengerActive() and likes).

amicic:
PR title must be max 50 char long (when WIP is removed)

charliegracie:
Do we need to do this via a lock?  Could we not just make sure we update things in the right order so that we never clean a card for memory that does not exist?

amicic:
Simple re-ordering of steps that are under critical section
1) expand HeapRegionManager ranges
2) expand and commit meta GC structures

what would likely fix the original problem, did not work, because expansion of meta GC structures also look into HRM ranges.

We are still looking into this approach, but trying to pass the ranges to  step 2 via methods arguments, instead of relying on HRM. Needs several APIs to be modified. 

We can wait to see how clean/messy the other solution is and compare them. Meanwhile, I was thinking to have at least a solution (before internal IBM stream split) and to be able to distinguish from other intermittent failures we've also seen .  Now that IBM dev stream just got split without the fix, waiting is plausible.

charliegracie:
I am ok with moving forward with this solution as a temporary fix to resolve the issue if the investigation is going to continue.

amicic:
Alternative investigation (with steps re-ordering instead of using a lock) is underway. We have a working fix, but code is somewhat messy. We are trying to simplify things.... Let's wait for a couple of days more, and than we'll compare them.

charliegracie:
Ok I will wait to hear back from you.

charliegracie:
Any update @RSalman?

RSalman:
@amicic and I have another solution, however we've uncovered another issue during testing. We believed it to be related to our fix, however, after looking into it, we found it to be an existing issue unrelated to our changes. 

Since this new issue comes up during testing, our attention has been to trying to get more details on the newly found issue before moving forward with a solution for  https://github.com/eclipse/omr/issues/2426.

For now, this PR can be merged as a temporary solution, however before we do so, I would like to run some more tests which I think are useful after our findings from the reworked solution.

PR #2421, opened 2018-04-04
Add new portlib signal flags to support new signals

babsingh:
OMR Issue: #2332


babsingh:
@charliegracie With this change, the `portlib signal flags` will become 64-bit. When `omrsignal.c::signalMap` is updated to include new signals, we will need to change `uint32_t` to `uint64_t` for fields where `portlib signal flag` is stored. do you have better recommendations to avoid this change?

charliegracie:
Is this going to break backwards compatibility in any way?  I do not think so but just asking the question out loud so we can think about it.

charliegracie:
@DanHeidinga thoughts?

babsingh:
> Is this going to break backwards compatibility in any way?

no. currently, existing APIs accept signal flags in the < 32-bit domain. existing APIs should work fine when the `flags` field is changed from `uint32_t` to `uint64_t`.

charliegracie:
@genie-omr build all zos

babsingh:
[linux_ppc-64_le_gcc](https://ci.eclipse.org/omr/job/PullRequest-linux_ppc-64_le_gcc/335/console) failure is a infrastructure setup failure; not an OMR functional issue.

charliegracie:
@genie-omr build plinux

PR #2424, opened 2018-04-05
Upgrade lint job environment on Travis CI to Trusty

nbhuiyan:
Currently, the OMRChecker lint job on Travis CI runs
in Precise. It is in the process of being decommisoned,
and the tentative official end of support for Precise
by Travis CI was April 1st 2018. This patch upgrades the
environment to Trusty, and also makes use of clang-3.8
instead of clang-3.4.

[Travis CI Blog post regarding this matter](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2016, 2017 IBM Corp. and others
+# Copyright (c) 2016, 2018 IBM Corp. and others
 # 
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


nbhuiyan:
@charliegracie @rwy0717 Would you be able to review this?

charliegracie:
I just re-started the lint job to verify the time it takes to run.  It seemed really long with these changes but it may have just been caused by ccache misses due to a new compiler.  If the time drops back down to the regular time I will merge.

nbhuiyan:
After the restart, the time taken to run the lint job has still been a longer than it usually took using clang-3.4 on Ubuntu Precise. The only things being done differently here is the additional `llvm-toolchain-trusty` source which adds a couple of seconds, and building the OMRChecker plugin using clang++-3.8 instead of g++. Do you have any other suspicions about the cause of this slowdown? 

I would expect a newer version of Clang would be faster, if not the same speed than the older version. I have been trying out a few things in the meantime to see if I can achieve an improvement.

charliegracie:
I am ok with the change as-is.  Merging.

PR #2428, opened 2018-04-06
Enable SSE/SSE2 on OMR as default

0dvictor:
All known downstream projects uses SSE/SSE2 for
floating point operations; however OMR itself
still kept X87 as default on X86-32.
This commit turns SSE/SSE2 on as default, allowing
OMR's tests exercising SSE/SSE2 route.
X87 support is deprecated, and will be removed.

Issue #2429

### DIFF
--- a/compiler/x/codegen/FPTreeEvaluator.cpp
+++ b/compiler/x/codegen/FPTreeEvaluator.cpp
@@ -510,8 +510,11 @@ TR::Register *OMR::X86::TreeEvaluator::fpReturnEvaluator(TR::Node *node, TR::Cod
    TR_ASSERT(returnRegister, "Return node's child should evaluate to a register");
    TR::Compilation *comp = cg->comp();
 
+   const TR::X86LinkageProperties &linkageProperties = cg->getProperties();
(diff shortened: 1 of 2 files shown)


compiler/x/codegen/FPTreeEvaluator.cpp
```diff
@@ -1146,7 +1145,11 @@ TR::Register *OMR::X86::TreeEvaluator::fpConvertToLong(TR::Node *node, TR::Symbo
    TR::Node *child = node->getFirstChild();
 
    // TODO: Perform f2l using SSE.
+#ifdef J9_PROJECT_SPECIFIC
    if (child->getOpCode().isDouble() && cg->useSSEForDoublePrecision())
+#else
+   if (false)
+#endif
```
0xdaryl: Could you explain this sequence of code and why OpenJ9 and OMR need to go down different paths now?

0dvictor: The SSE path uses a helper call that is not implemented by OMR. It is indeed a OpenJ9 specific code although it is not marked as `J9_PROJECT_SPECIFIC`.
We should definitely make the SSE path not relying on any J9 specific functions; however, I personally don't think we should hold enabling SSE/SSE2 solely because the path uses J9 specific code.

0dvictor:
Good point. I'm closing this PR now.

PR #2431, opened 2018-04-09
 Move addPrefetch code from OMR to OpenJ9

IBMJimmyk:
This removes addPrefetch from OMR. It is replaced by a call to the method
insertPrefetchIfNecessary. By default insertPrefetchIfNecessary just
returns. Extending projects can override it to add functionality.

The performance improvement changes that were initially intended to go into addPrefetch have now been moved to insertPrefetchIfNecessary inside OpenJ9.

iloadEvaluator and l2aEvaluator were also modified to now use
insertPrefetchIfNecessary instead of addPrefetch.

Prefetch was added to iloadEvaluator on Power because it was shown to
improve performance on an internal appserver benchmark.

The OpenJ9 changes will be handled by a pull request in OpenJ9.


IBMJimmyk:
That sounds like a good idea. I'll try it out.

IBMJimmyk:
I updated the change set to remove addPrefetch. The openj9 half of this change (https://github.com/eclipse/openj9/pull/1636) now has the addPrefetch code under the name insertPrefetchIfNecessary. The two calls to addPrefetch were updated to call insertPrefetchIfNecessary. There is also now a insertPrefetchIfNecessary in omr that intentionally does nothing. It can be overrided with a function that does do something.

IBMJimmyk:
@0xdaryl Do you have any other comments on the changes made?

IBMJimmyk:
@0xdaryl Sorry to bug you again but are there any other changes needed?

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -482,6 +482,12 @@ class OMR_EXTENSIBLE CodeGenerator
 
    TR::Recompilation *allocateRecompilationInfo() { return NULL; }
 
+   /*
+    * This determines if it is necessary to emit a prefetch instruction.
+    * If so, it also emits the prefetch instruction.
+    */
```
0xdaryl: Could you add the documentation for each of these parameters and what they are for so that a downstream implementing project knows what to do?  I must admit I'm confused by `targetRegister` because the one place this is used in OMR the argument passed in is named `source` !  Thanks.

IBMJimmyk: I added a comment saying what each parameter was. The names I used were the same as the ones in the original addPrefetch function that got moved to OpenJ9 and renamed to insertPrefetchIfNecessary. Your specific example of passing in "source" as the "targetRegister" was something that was already there. I think it is because it is a source register inside l2aEvaluator but it is used to generate the prefetch target inside insertPrefetchIfNecessary.

0xdaryl: Thanks.  But please use Doxygen syntax (i.e., `@brief`, `@param`, `@return` tags) in comments, and document the API in the header file (`OMRCodeGenerator.hpp`) not the implementation (`.cpp`) file.

IBMJimmyk: I moved the comment to the hpp file and changed it to use Doxygen syntax.

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build all

vijaysun-omr:
The appveyor build is still not done here (infra issue), but I am going to merge this since it's primarily a Power change and it has to be merged with some urgency I was told (for a downstream project).

PR #2432, opened 2018-04-09
Fix signal issues unix

babsingh:
Two issues are being addressed in this pull request:

1. **Fix issue with re-registering master handlers**. `OMRSIG_SIGACTION` can be called outside of `omrsignal.c` functions. A call to `OMRSIG_SIGACTION` will override the master handlers set by either `omrsig_protect` or `omrsig_set_async_signal_handler`. Setting a master handler once sets a bit in `signalsWithMasterHandlers`. If the master handler is overridden by an external call to `OMRSIG_SIGACTION`, then the bit in `signalsWithMasterHandlers` will not be un-set. Since the bit in `signalsWithMasterHandlers` isn't un-set once the master handler has been overridden, then subsequent calls to `omrsig_protect` and `omrsig_set_async_signal_handler` will not re-register the master handler. In subsequent calls, it will look as if the master handler is already registered. We mustn't rely on `signalsWithMasterHandlers` to re-register master handlers. It doesn't cause any harm to register master handlers every time a call is made to either `omrsig_protect` or `omrsig_set_async_signal_handler`.

2. **Initialize cached oldAction before invoking OMRSIG_SIGACTION**. Before registering handlers using `OMRSIG_SIGACTION`, `oldActions[unixSignalNo].action` (`struct sigaction`) needs to be properly initialized or reset (in case of re-registration).

OMR Issue: #2332

### DIFF
--- a/port/unix/omrsignal.c
+++ b/port/unix/omrsignal.c
@@ -322,7 +322,6 @@ omrsig_protect(struct OMRPortLibrary *portLibrary, omrsig_protected_fn fn, void 
 	omrthread_t thisThread = NULL;
 	uint32_t rc = 0;
 	uint32_t flagsSignalsOnly = 0;
-	uint32_t flagsWithoutMasterHandlers = 0;
 
 	Trc_PRT_signal_omrsig_protect_entered(fn, fn_arg, handler, handler_arg, flags);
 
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all zos

charliegracie:
I will spend more time to review this later today or tomorrow.  I need to understand the current code more to be able to understand the changes.  Sorry for the delay.

babsingh:
> Can you verify that you have tested this with OpenJ9 and that it does not cause any issues?

No failures were seen when these changes were tested with OpenJ9.

PR #2433, opened 2018-04-10
Optimized iabs and labs evaluator on X86

0dvictor:
Reduced one instruction (out of three) in generated
code for iabs and labs.


0xdaryl:
This is an x86-specific change so a genie build is not required.

Issue #2405, opened 2018-03-29
Many compilation units contain nonresolvable WINDOWS macrodefinition


samolisov:
When I try building `OMR` with enabled `JitBuilder` on Windows using `MSVC` or `clang`, the following error appears:

```bash
omr\compiler\cs2/timer.h(43): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory
```

If we open the `compiler/cs2/timer.h` file, the following macrodefinition will be found:

```c
#if defined(WINDOWS)
#include <time.h>
#include "windows_api.h"
#else
#include <sys/time.h>
#endif
```

So, `#include <sys/time.h>` is going to the compiled file regardless the build runs on Windows!

As described in the documentation, `doc/BuildingWithCMake.cmake`, on Windows we use the `MSVC` and `clang` toolchains to build `OMR`. I've checked, both toolchains define the following symbols on Windows (x64):

 * WIN32
 * _WIN32
 * _WIN64

No `WINDOWS` or `WIN64` are defined there!

My suggestion is to add the using of `WIN32` or `_WIN32` (but `WIN32` is much common for `OMR`, see later) in the coding standard and use this symbol where possible. I'm not sure whether the construction `#if defined(WIN32) || defined(WIN64)` does matter because `WIN64` is defined neither by `clang` nor by `MSVC` even on Windows x64. If we must be sure the code is compiled on (for) Windows x32 only, the construction `#if defined(WIN32) && !defined(OMR_ENV_DATA64)` looks like a good solution. 

Also, a number of compiler specific checks are in the project, I have no idea what to do with that (`MinGW`-related especially), so your opinion is needed!

---

The analysis shows the `WINDOWS`  word occurrences in the following files:
 * `compiler/compile/OMRCompilation.cpp`
 * `compiler/control/OMROptions.cpp`
 * `compiler/cs2/timer.h`
 * `compiler/cs2/windows_api.h`
 * `compiler/infra/Annotations.hpp`
 * `compiler/infra/Bit.hpp`
 * `compiler/infra/OMRMonitor.cpp`
 * `compiler/infra/ThreadLocal.h `
 * `compiler/ras/CallStack.cpp`
 * `compiler/ras/CallStackIterator.hpp`
 * `compiler/ras/Debug.cpp`
 * `compiler/ras/Debug.hpp`
 * `compiler/ras/IgnoreLocale.hpp`
 * `compiler/x/runtime/X86Runtime.hpp`

[...]

PR #2434, opened 2018-04-11
WIP: Windows compile definitions - OMR_WINDOWS introduction

samolisov:
Add the `OMR_WINDOWS` compile definition

As it's discussed with @youngar in the comments to issue #2405,
it would be good to start using macros prefixed with `OMR_`.
As more people use OMR, it would be nice if we didn't collide
with commonly used macro names like `WINDOWS`.

The JIT already uses `WINDOWS`, while the rest of the code would
need to be updated.

The constructions such as `#ifdef WIN32` will be replaced by
`#if defined(OMR_WINDOWS)`

The constructions such as `#ifdef WIN64` will be replaced by
`#if defined(OMR_WINDOWS) && defined(OMR_ENV_DATA64)`

The `OMR_WINDOWS` compile definition is added to the CMake
configuration as well as the autotools one.

Issue: #2405


youngar:
Hi, looks like a great start!

It seems like some of the Windows macros are used by msvc specify what version of Windows we are compiling for.  This makes sure we use the right APIs that were available on older systems.
```cmake
set(OMR_WINVER "0x501")
	-D_WIN_95	
	-D_WIN32_WINDOWS=0x0500	
	-D_WIN32_WINVER=${OMR_WINVER}	
	-D_WIN32_WINNT=${OMR_WINVER}
```
[These correspond to](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt):
```c++
#define _WIN32_WINNT_WIN2K                  0x0500 // Windows 2000  
#define _WIN32_WINNT_WINXP                  0x0501 // Windows XP 
```
I think if we want to make sure that OMR still supports Windows XP we should keep the macro `-D_WIN32_WINNT=${OMR_WINVER}`.  The rest should probably be deleted.  I'll ask around and see if we actually still want to target Win XP.

`_WIN32_DCOM` also appears to be special msvc macro, which we might need.  I don't have much experience with Windows programming, and I didn't understand the explanation from stackoverflow 😛 .

youngar:
Another thing I was wondering:  we might be to remove `WIN32` and `WIN64` defines as well. They should be automatically defined by `msvc`.  Not sure if they are defined by `clang` or `mingw`.

samolisov:
@youngar Yes, this is a goal of the PR to remove WIN32 and WIN64 but these definitions should be removed from the code before. I'm working on this.

About WIN32_WINNT_WINXP, I'm a bit confused: the definition in the cmake file is WIN32_WINNT, not  WIN32_WINNT_WINXP, so this is a reason why I've removed it. Since we don't specify WIN32_WINNT_WINXP explicitly, I assume MSVC or WinKit does it for us. Maybe a test on Win XP is needed.

youngar:
Hi, my understanding is the you set `#define _WIN32_WINNT xxx` to indicate what version of windows you want to build for.  By defining it to `0x0501`, we are saying that we want to build using the Windows XP version of Windows header files.  My understanding from the documentation is that `WIN32_WINNT_WINXP` is always defined in the Window's header file `SDKDDKVer.h`.  For some reason we are supposed to define `_WIN32_WINNT ` before we include that header file, or else we could have written `#define _WIN32_WINNT WIN32_WINNT_WINXP`.

samolisov:
I've seen a number of `defined(OMRZTPF)` occurrences in the code, so I get the following question: should we use `OMRWINDOWS` or `OMR_WINDOWS` to get the code is compiled on the Windows OS? `OMR_WINDOWS` looks a little bit readable but `OMRWINDOWS` is more consistent with already used `OMRZTPF` and `J9HAMMER`.

I've pushed the commit that changes `WINDOWS` to `OMR_WINDOWS` in the **compiler** module. I need to check the module on Windows but the module has a non-standard `CMakeLists.txt`: only children `CMakeLists` are included to the buildsystem and I know only one way - to enable OMR_JITBUILDER that can't be compiled itself due to a number of compilation errors.

samolisov:
Oops:

compiler/env/defines.h:

```cpp
/* Operating Systems */
#ifndef OMR_LINUX
#  define OMR_LINUX   201
#endif
#ifndef OMR_WINDOWS
#  define OMR_WINDOWS 202
#endif
#ifndef OMR_AIX
#  define OMR_AIX     203
#endif
#ifndef OMR_ZOS
#  define OMR_ZOS     204
#endif
#ifndef OMR_OSX
#  define OMR_OSX     205
#endif
```

so, `OMR_WINDOWS` is already defined for ALL operating systems.

I have no idea what to do... If we have a look at the `defines.h` the `OMR_WINDOWS` definition and its siblings are used to set the `HOST_OS` compile definition. Should we use `HOST_OS` and check something like: `#if HOST_OS == OMR_WINDOWS || HOST_OS == OMR_ZTPF` but it's very noisy, isn't it?

samolisov:
I've pushed the commits that change all occurrences of `WIN32`, `_WIN32`, `WIN32 || WIN64`, `WIN64` to either `OMRWINDOWS` or `OMRWINDOWS && [!]OMR_ENV_DATA64` within the whole codebase. Unit tests on Windows haven't been broken (at least those what worked fine).

youngar:
Hi! Thanks for putting in all this work!  I think that this approach is good for now, but we are still far away from standardizing our macro usages.

I agree that typing `HOST_OS == OMR_WINDOWS` every time is verbose, but there is a reason it was done that way:  The compiler supports both a `HOST_OS` and a `TARGET_OS`.  Theoretically,  the compiler supports cross compilation. `HOST_OS` is the os the compiler is running on, and `TARGET_OS` is what the compiled code will be running on.  After this is in, I'll probably create a new issue to move `compiler\env\defines.h` to be global to OMR.

youngar:
We need to have `-D_WIN32_WINNT=0x0501` added back and `-DWINVER=0x0501` added to make sure we are building for Windows XP.

samolisov:
@youngar Thank you for the comments. I've returned `-D_WIN32_WINNT=0x0501`  and put `-DWINVER=0x0501` into the `win.cmake` file.

samolisov:
I've removed the WIP marker. Please review the PR, this PR can be the first step to build the compiler on Windows.

youngar:
Hi, I launched a OpenJ9 build to check your changes, and it failed right away. I'm copying the error below, but you shouldn't expect it to be helpful.
```
     [exec] J9DDR Autoblob GenerateInputCFile
     [exec] Writing C file: /tmp/bld_385157/bld_win_x86-64_cmprssptrs/ddr/vmddrstructs.c
     [exec] C file written
     [exec] chmod 755 ../buildtools/extract_structures/linux_x86/extract_structures
     [exec] cpp -undef -nostdinc -include typestubs.h -Idummy_headers -DJAVA_SPEC_VERSION=9 -D_AMD64_=1 -DWIN64 -D_WIN64 -DWIN32 -D_WIN32 -DJ9HAMMER -D_CRT_SECURE_NO_WARNINGS -DCRTAPI1=_cdecl -DCRTAPI2=_cdecl -D_WIN95 -D_WIN32_WINDOWS=0x0500 -D_WIN32_DCOM -D_WIN32_IE=0x0500 -DWINVER=0x0501 -D_WIN32_WINNT=0x0501 -D_MT -D_WINSOCKAPI_ -D_DLL -I. -Iomr/include_core -I../rastrace -I../include -I../oti -I../gc_include -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/gc/include -I../shared_util -Ibcutil -Ishared_common -I../jvmti -I../rasdump -I../rastrace -I../util -I../shared_common -I../jcl/common -I../jcl -I../harmony/include -I../compiler/trj9/runtime -I../compiler/trj9 -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/compiler -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/compiler/base -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/compiler/env -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/port -I../port -I../shared_common/include -I../bcverify -I../vm -I../dbgext -I../bcutil -I../nls -I/tmp/bld_385157/bld_win_x86-64_cmprssptrs/omr/include_core -DUT_TRACE_OVERHEAD=0 -DTR_HOST_X86 -DTR_64Bit -DTR_HOST_64BIT -DTR_TARGET_64BIT -DTR_TARGET_X86 -DUT_DIRECT_TRACE_REGISTRATION -DTR_HOST_X86  -o vmddrstructs.i vmddrstructs.c
     [exec] /bluebird/bin/platform/linux386/ibm-jdk-1.8.0/bin/java -cp ../buildtools/j9ddr-autoblob.jar com.ibm.j9ddr.autoblob.StripGarbage vmddrstructs.i
     [exec] Cleaning vmddrstructs.i
     [exec] ../buildtools/extract_structures/linux_x86/extract_structures --c89 --no_warnings vmddrstructs.ic
     [exec] J9DDR C/C++ Type Extraction Utility
     [exec] Based on EDG C++ Front-End
     [exec] This binary is licensed for IBM Internal Use Only
     [exec] "../oti/j9port.h", line 298: error: identifier "SYS_FLOAT" is undefined
     [exec]    SYS_FLOAT cpuEntitlement;
     [exec]    ^
     [exec] 
     [exec] 1 error detected in the compilation of "vmddrstructs.ic".
     [exec] EDG finished with errors
     [exec] gmake[2]: *** [vmddrblob.c] Error 2
```

I'm guessing this is not a problem with your changes, we probably just need to update OpenJ9 to define OMRWINDOWS when it includes OMR header files.  I'll try it out, but unfortunately I won't be able to do that until tomorrow.  As a general rule we don't commit anything OMR until we are sure it works with OpenJ9.

dnakamura:
Rather than making a new define `OMRWINDOWS` why not rely on the standard `_WIN32`/`_WIN64`?

youngar:
The compiler component doesn't follow the same coding standard as the rest of OMR.  You added a lot of comments to the macro statements, for example:
```
+#endif /* defined(OMRWINDOWS) && defined(OMR_ENV_DATA64) */
```
I think its better to keep compiler component consistent with itself, and that the comments should be removed from anything in the `compiler` directory.

dnakamura:
I agree the semantics are a bit awkward (but it makes sense given that it originally was used to differentiate from the 16-bit windows api). However I feel that we should be standards / convention where possible.  I would propose relying on `_WIN32`  purely for checking the OS, just like you would use `__linux__` etc, and create an `OMR_ENV32`/ `OMR_ENV64` (which we may have had in the past).  In fact this already basically exists with `TR_HOST_32BIT` / `TR_HOST_64BIT`

gacholio:
Similar to the OMR_ARCH_xxx flags, I'd like to see OMR_OS_xxx flags, rather than a specific one for windows.  We already have something similar to this in the directory structure - for Linux, we could define OMR_OS_UNIX and OMR_OS_LINUX (we consider ZOS to be a variant of Unix, so it would also define two OS flags).

I would definitely prefer not to have new flags for 32-bit and 64-bit windows.  I'd prefer a combination of OMR_OS_WINDOWS and OMR_ENV_DATA64.

samolisov:
Colleagues, I'm working on fixing JIT compiler compilation errors under Windows and this PR should be the first step. As I see, there is no consensus about how to go forward: must we switch to enable the cross-compilation of the compiler as @dnakamura said or just rename the OS flags to `OMR_OS_WINDOWS`. I have no ideas about the cross-compilation for now but I'm ready to rename `OMR_WINDOWS` to `OMR_OS_WINDOWS` and even fix the others OS flags if it's required. What do you think?

(Another option is just to rename the `WINDOWS` compile definition to the `WIN32` one for the compiler code as a temporal solution until the decision is made).

charliegracie:
Hi @samolisov, sorry I was travelling for work and missed some updates here.  I would go forward with replacing OMR_WINDOWS with OMR_OS_WINDOWS and then we can get the changes delivered.  I added the compiler flag so @0xdaryl, @mstoodle, @vijaysun-omr and @fjeremic can comment / review the compiler changes but going forward this is the direction I believe the project wants / needs to go.

cmake/modules/platform/os/win.cmake
```diff
@@ -22,29 +22,15 @@
 set(OMR_WINVER "0x501")
```
youngar: Please delete this line, its not used anymore

samolisov: @youngar ok, I've deleted.

youngar:
The prerequisite change for OpenJ9 is here: https://github.com/eclipse/openj9/pull/1922

youngar:
I believe we still need to change the define in two more places:
```c++
fvtest/compilertest/build/toolcfg/host/win.mk:22:HOST_DEFINES+=WINDOWS
jitbuilder/build/toolcfg/host/win.mk:23:HOST_DEFINES+=WINDOWS
```

samolisov:
@youngar Oh, thank you for the comment. I've replaced `WINDOWS` to `OMR_OS_WINDOWS` there. A conflict in `port/common/omrstr.c` has also been fixed.

charliegracie:
@youngar can this be merged?

charliegracie:
@genie-omr build all

PR #2435, opened 2018-04-11
VRSb instruction memory reference needs separate index register

NigelYiboYu:
This pull request contains two commits:

1. The VRSb instruction format does not support index fields.
Fix the generateVRSbInstruction() API so it separates the memory
references index register. 

2.  Add an assert check to make sure the right node is used to generate memory reference. 
The generateS390MemoryReference(TR::Node*, TR::CodeGenerator*,bool)
API uses a constructor that handles load and store nodes with symbol
references. If the node has a no symbol reference, the compiler can crash in binary encoding.

### DIFF
--- a/compiler/z/codegen/OMRMemoryReference.cpp
+++ b/compiler/z/codegen/OMRMemoryReference.cpp
@@ -3696,6 +3696,12 @@ generateS390MemoryReference(TR::Register * br, int32_t disp, TR::CodeGenerator *
 TR::MemoryReference *
 generateS390MemoryReference(TR::Node * node, TR::CodeGenerator * cg, bool canUseRX)
    {
+   // The TR::MemoryReference(TR::Node*, TR::CodeGenerator*, bool); constructor (below) is made for store or load nodes
+   // that have symbol references.
+   // A symbol reference is needed to adjust memory reference displacement (TR::MemoryReference::calcDisplacement).
+   // If the node has no sym ref, this memory reference's gets a NULL symbol reference, which can lead to crashes in displacement
+   // calculations.
+   TR_ASSERT(node->getOpCode().hasSymbolReference(), "Memory reference generation API needs a node with symbol reference\n");
    return new (cg->trHeapMemory()) TR::MemoryReference(node, cg, canUseRX);
    }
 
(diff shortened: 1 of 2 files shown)


NigelYiboYu:
This change is being tested. Marking it WIP.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -3696,6 +3696,10 @@ generateS390MemoryReference(TR::Register * br, int32_t disp, TR::CodeGenerator *
 TR::MemoryReference *
 generateS390MemoryReference(TR::Node * node, TR::CodeGenerator * cg, bool canUseRX)
    {
+   // The TR::MemoryReference(TR::Node*, TR::CodeGenerator*, bool); constructor (below) is made for store or load nodes
+   // that have symbol references. If the node has no sym ref, this memory reference's symbolRef is null, whichs can lead
```
fjeremic: Typo on "whichs". Also can you describe what kind of problems can be encountered in the binary encoding phase? i.e. who will need this symbol reference.

NigelYiboYu: Fixed.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -3696,6 +3696,12 @@ generateS390MemoryReference(TR::Register * br, int32_t disp, TR::CodeGenerator *
 TR::MemoryReference *
 generateS390MemoryReference(TR::Node * node, TR::CodeGenerator * cg, bool canUseRX)
    {
+   // The TR::MemoryReference(TR::Node*, TR::CodeGenerator*, bool); constructor (below) is made for store or load nodes
+   // that have symbol references.
+   // A symbol reference is needed to adjust memory reference displacement (TR::MemoryReference::calcDisplacement).
+   // If the node has no sym ref, this memory reference's gets a NULL symbol reference, which can lead to crashes in displacement
+   // calculations.
```
joransiu: Based on the comment, would it make more sense to push the TR_ASSERT to the constructor instead?

fjeremic:
@NigelYiboYu if this is ready for merging please remove the WIP and I'll do a final review.

NigelYiboYu:
Tests passed. Ready to merge.

compiler/z/codegen/OMRMemoryReference.cpp
NigelYiboYu: This has been done.

fjeremic:
@genie-omr build zlinux, zos

PR #2436, opened 2018-04-11
Windows MSVC memory barriers affect only the compiler

gacholio:
On the Windows MSVC compiler, the memory barrier primitives were only
affecting compiler reordering, not processor reordering.

- ensure MSVC barriers emit appropriate instructions
- ensure compiler reordering does not occur in existing barriers
- add compiler reordering barrier

Signed-off-by: Graham Chapman <graham_chapma.ibm.com>


charliegracie:
@genie-omr build all zos

PR #2438, opened 2018-04-12
Stop versioning w.r.t. variable-step IVs

jdmpapin:
Loop versioner has been trying to identify induction variables (IVs)
that step by a non-constant (but still loop-invariant) amount each
iteration, so that BNDCHKs involving such a variable can be versioned.
However, the versioning tests for those checks are currently incorrect.
For the time being, this commit prevents versioner from recognizing IVs
unless they step by an amount that is a compile-time constant.


vijaysun-omr:
Looks like a safe change on it's own since it is only disabling a kind of versioning temporarily.

vijaysun-omr:
@genie-omr build all

PR #2439, opened 2018-04-12
Rewrite ificmp{eq,ne} of comparisons during VP

jdmpapin:
VP (value propagation) will now simplify conditional trees of the
following form into a single fused conditional branch.

    ificmp{eq,ne}
      xcmp{eq,ne}
        lhs
        rhs
      iconst {0,1}

This allows global VP to propagate null/non-null constraints when one
child of the conditional was originally an instanceof, and it was
transformed into a comparison against null earlier in the same pass.


vijaysun-omr:
The change looks okay to me. Is there some reason this only handles integer if compares and does not also include long for example ?

vijaysun-omr:
@genie-omr build all

PR #2440, opened 2018-04-12
Recognize String.value loads as non-null in VP

jdmpapin:


vijaysun-omr:
Looks safe to me. @fjeremic I assume there are no edge cases related to string compression that could cause String.value to be null ? Just thought I'd check before merging.

vijaysun-omr:
@genie-omr build all

PR #2441, opened 2018-04-12
Remove unused compressedRefs anchors in dead trees

jdmpapin:
Dead trees elimination removes indirect loads under compressedRefs
anchors when they have reference count 1, i.e. when their results are
unused. But when the result of such a load is used anywhere, its anchor
is left in place. So far this determination has been made only when the
pass encounters the anchor, which means that it counts later uses even
if those uses are themselves about to be removed.

Dead trees elimination now revisits compressedRefs anchors for loads at
the end of the pass, in order to determine whether the anchored loads
have become unused, and remove any that have.

### DIFF
--- a/compiler/optimizer/DeadTreesElimination.cpp
+++ b/compiler/optimizer/DeadTreesElimination.cpp
@@ -23,6 +23,7 @@
 
 #include <stddef.h>                             // for NULL
 #include <stdint.h>                             // for int32_t, uint32_t
+#include "infra/forward_list.hpp"               // for TR::forward_list
 #include "codegen/CodeGenerator.hpp"            // for CodeGenerator
 #include "codegen/FrontEnd.hpp"                 // for TR_FrontEnd, etc
 #include "compile/Compilation.hpp"              // for Compilation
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
@genie-omr build all

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -883,6 +901,45 @@ int32_t TR::DeadTreesElimination::process(TR::TreeTop *startTree, TR::TreeTop *e
...
       }
 
+   for (auto it = anchors.begin(); it != anchors.end(); ++it)
+      {
+      TR::Node *anchor = it->tree->getNode();
+      TR::Node *load = anchor->getChild(0);
+      if (load->getReferenceCount() > 1)
+         continue;
+
+      TR_ASSERT_FATAL(
```
vijaysun-omr: While it's likely that the child of a compressed refs anchor is an indirect load, is there a reason why the check at the point where the compressed refs anchor is added to the "anchors" data structure is only isLoad ? i.e. is there a possibility that this assert fails in some obscure case when the indirect load has been changed to a direct load and the compressed refs anchor has not been deleted ? Why not check for isLoadIndirect at the point of the addition to "anchors" ?

fjeremic: Tree verification passes do not run in production builds, while the above assert will fire. I do see your point however, and I agree.

jdmpapin: Updated to remove the assertion and instead look specifically for `isLoadIndirect()` when queueing up the trees

PR #2442, opened 2018-04-12
Share result temps for primitive return values

jdmpapin:
Starting in ade5a1cc, isNotCollected() returns true for primitives,
which causes this conditional to assign them all fresh temporaries.


vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks safe to me. Merging.

PR #2443, opened 2018-04-12
WIP: Avoid generating redundant gotos in simplifier

jdmpapin:
Simplifier's partialRedundantCompareElimination() could sometimes
generate redundant gotos, if the fallthrough successor of the original
conditional is just a goto. Now in that case it will try to skip over
the intermediate goto.

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -1979,6 +1979,39 @@ static TR::SymbolReference *getUniqueSymRef(TR::Node *node, bool &isLegal , TR::
    return savedSymRef;
    }
 
+static TR::TreeTop *forwardThroughGoto(TR::TreeTop *origTarget)
+   {
+   // Don't skip GlRegDeps
+   if (origTarget->getNode()->getNumChildren() > 0)
+      return origTarget;
+
+   // Don't skip loop preheaders
+   TR::Block *origBlock = origTarget->getNode()->getBlock();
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
@genie-omr build all

vijaysun-omr:
This looks okay on it's own but is'nt this kind of thing cleaned up by the rest of the optimizer afterwards ? Maybe it's just that in the specific case you were looking at, there is no pass afterwards in the opt strategy and it's easy enough to clean this up at the source, that you felt you should just clean it up.

jdmpapin:
In the case I saw, it doesn't yet get cleaned up afterwards. I was in the process of adding a very late pass of redundant goto elimination (see #2444), but with the intention that it should run only on particular blocks discovered by dead trees elimination. Since this goto is already redundant, it won't be discovered that way.

However, it turns out that there's already a later pass of redundant goto elimination in my case, and that pass also doesn't fix the gotos. The reason is that there's one predecessor that falls through, so redundant goto elimination can't delete the block from the CFG. I've since addressed that elsewhere: "Remove redundant gotos more opportunistically" in #2444 allows redundant goto elimination to redirect all of the *other* predecessors. So perhaps that change obviates this one.

vijaysun-omr:
Just checking : do you still want to go ahead with this change then ?

jdmpapin:
I've marked this WIP for the time being. Later depending on #2444 I'll either close this one or remove WIP.

jdmpapin:
Closing since #2444 has been merged, and as mentioned above one of its commits makes this unnecessary

PR #2444, opened 2018-04-12
Prepare redundant goto elim for post-GRA requests

jdmpapin:
Some very late passes of global dead store elimination and dead trees
elimination in openj9 can create opportunities for redundant goto
elimination after GRA (global register allocation). This series updates
redundant goto elimination so that it works on blocks that have register
dependencies, and then makes dead trees elimination request it on
promising blocks. This way it's possible to add an "if enabled" pass of
redundant goto elimination that only looks at blocks where these
opportunities are likely to have been created.

### DIFF
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -3432,36 +3432,31 @@ int32_t TR_EliminateRedundantGotos::process(TR::TreeTop *startTree, TR::TreeTop 
       if (block->getPredecessors().empty())
          continue;
 
-      bool gotoBlockRemovable = true;
+      if (block->isLoopInvariantBlock())
+         continue;
+
       TR::Block *destBlock = block->getSuccessors().front()->getTo()->asBlock();
-      auto inEdge = block->getPredecessors().begin();
-      while (inEdge != block->getPredecessors().end())
+      TR::CFGEdgeList fixablePreds(comp()->trMemory()->currentStackRegion());
+      auto preds = block->getPredecessors();
+      for (auto inEdge = preds.begin(); inEdge != preds.end(); ++inEdge)
          {
          if (((*inEdge)->getFrom() == cfg->getStart()) || ((*inEdge)->getFrom() == block))
-            {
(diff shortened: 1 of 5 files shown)


compiler/optimizer/LocalOpts.cpp
```diff
@@ -3386,6 +3386,49 @@ int32_t TR_EliminateRedundantGotos::process(TR::TreeTop *startTree, TR::TreeTop
...
+      if (rootStructure != NULL)
+         {
+         blockRegion = block->getParentStructureIfExists(cfg);
+         blockIsRegionEntry =
+            blockRegion != NULL && blockRegion->getNumber() == block->getNumber();
+         }
+
+      bool manuallyFixStructure = blockIsRegionEntry;
+
+      if (block->isExtensionOfPreviousBlock())
```
vijaysun-omr: Could you please elaborate on why we set the bool to manually fix structure to false in this case (I understand that the test guarantees that it cannot be a loop header but am missing the connection to why that matters wrt manually fixin structure).

vijaysun-omr:
@genie-omr build all

compiler/optimizer/LocalOpts.cpp
jdmpapin: The intention was to avoid unnecessarily failing [this test in the next commit](https://github.com/eclipse/omr/pull/2444/commits/9be5d861d71e74ca3f8113aa3455f1d342213a62#diff-bf9bdb533ff9e25408c9fac330ac7aedR3495), which says that when manually fixing structure, we have to be able to adjust all predecessors.

But come to think of it again, it seems it doesn't help to change `manuallyFixStructure` here. If the block is an extension, it has exactly one predecessor and that predecessor falls through. We only accept fallthroughs for empty blocks, so the block would fail that test because of `emptyBlock` anyway.

jdmpapin: Updated to remove the unnecessary change to `manuallyFixStructure` in this case

vijaysun-omr:
@genie-omr build all

PR #2448, opened 2018-04-13
Introduce RegMem form of PSHUFB on X86

0dvictor:
RegMem form of PSHUFB is added in case downstream projects need it.


0xdaryl:
This is an x86-specific change.  No genie build required.

PR #2457, opened 2018-04-15
Make extensible class macro naming more consistent and correct

0xdaryl:
* Fix concrete extensible class macro names
* Fix extensible class connector macros (remove COMPOSED)
* Fix extensible class connector macros (remove BASE)


fjeremic:
@genie-omr build all

PR #2458, opened 2018-04-16
Removing-Register-assignment-direction-queries

vignesh240:


vignesh240:
Unused Register assignment direction queries removed

Issue #2159, opened 2017-12-19
Add an IL Validation Rule that checks whether `aiadd` and `aiuadd` are valid on the target platform


rbanerjee:
The `aiadd` OpCode is only valid when the platform we are compiling on is a `32 bit` one.

This is because we expect binary operations to operate on fields of the same length.
However, in the OMR compiler IL, `integer` sized opcodes are always 32 bits.
Meaning on a non-32 bit platform this ends up trying to invoke a binary operand
on arguments with different length, which is undefined.

[The same reasoning applies to aiuadd, the unsigned counterpart of aiadd]

The process would involve implementing the steps outlined below:

- [ ] Create a `TR::NodeValidation` Rule that checks whether the OpCode in question is of type `aiadd | aiuadd` and the target platform is a `32` bit one.
- [ ] Make this `ILValidationRule` a part the "appropriate" `ILValidationStrategy` array.
  [This would require investigating where exactly in the Compilation phase does this Rule get 
    satisfied, I.e, `right after ILGen`, `immediately before going into codegen` etc.]
                     
- [ ] Add the associated  `Tril` `ILValidatorTest`s

Leonardo2718:
I'll just mention this for completeness: there will also need to be a rule to validate that `aladd` and `aluadd` are only used on 64-bit target platforms.

PR #2460, opened 2018-04-16
Fix for incorrect evaluation of TR::aladd on X86-32

0dvictor:
TR::aladd was incorrectly evaluated on X86-32 by adding address
register to the high register of the offset register pair.
Fixing it by using the low register of the offset register pair.

### DIFF
--- a/compiler/x/codegen/BinaryCommutativeAnalyser.cpp
+++ b/compiler/x/codegen/BinaryCommutativeAnalyser.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


vijaysun-omr:
I did'nt think that TR::aladd was expected to be generated on 32 bit platforms, instead TR::aiadd was what was expected as per the original intent. Could you please clarify what generates TR::aladd under 32-bit ?

0dvictor:
I didn't research who generated TR::aladd. However, according to this line https://github.com/eclipse/omr/blob/546c38660cdf15f559d3360599060ac3e853776b/compiler/x/i386/codegen/TreeEvaluatorTable.hpp#L599, TR::aladd is supported on 32-bit and hence should be fixed.

rbanerjee:
From my understanding `aladd` claims to add a `long` sized value to an address. In the OMR IL `long` is defined to be of size 64. So on a 32 bit machine this would entail adding a value of size 64 to something of size 32, and this should in turn generate an error (As binary Operations in the OMR IL must always be applied on fields of the same size) .

So IIUC (and as @vijaysun-omr points out) this is a bug in IL generation with open Issues on the subject: Namely https://github.com/eclipse/omr/issues/566 (which deals with `aiadd` but a symmetric argument can be applied to `aladd`) and also as @Leonardo2718 mentions here https://github.com/eclipse/omr/issues/2159#issuecomment-352838699

0dvictor:
Thank you both for sharing the background. If that is the case, I guess the proper way to fix the issues is removing the evaluator from the evaluator table and fixing whatever issues it causes.

0dvictor:
PR created to remove aiadd and aladd evaluators from the corresponding evaluator tables. https://github.com/eclipse/omr/pull/2473

0dvictor:
Closing this PR.

Issue #2459, opened 2018-04-16
Track differences between code generators


0xdaryl:
This issue is intended to document the functional and performance differences between the various code generators.  The idea is to document what's *missing* in each code generator based on knowledge of other code generators rather than what support it actually has.  For example, unimplemented IL opcodes, lack of register assignment functionality, etc.

The goal is to eventually close the functional and performance gaps between the code generators, ideally by sharing more code between them.  But to do that we need to understand where the limitations and gaps currently are.

Anyone is free to add to this issue, either from what they know offhand or what they discover while fixing or reviewing the code.  Ultimately, work items will be created from these differences.

## AArch32
* no support for live virtual registers or register interferences
* assocreg (register association hints to the register assigner)

## Power
* ???

## X86
* ???

## Z
* ???

fjeremic:
Some things off top my head:

- (https://github.com/eclipse/omr/issues/397) Debug counter support inside internal control flow is lacking on Power and x86 (may not be needed on x86?)
- We seem to be using `NoReg` vs. `AssignAny` in dependencies interchangeably on different codegens
- Use of ScratchRegisterManager is seldom across the board, but especially on Power
  - We should avoid passing "temp" registers to helper functions
- Codegen phases are a mess at the moment in my opinion
- I think at least one other codegen does not have a proper peephole pass
  - The code handling peepholes is in some other place IIRC

I'll update as I think of more things.

PR #2462, opened 2018-04-16
Add boundary checking for _registerFile

dchopra001:
`_registerFile` is a list of RealRegisters whose indices are grabbed from RealRegisterEnum.hpp for all RealRegisters. This enum defines PseudoRealRegisters as well as RealRegisters. PseudoRealRegisters are not defined to be in the bounds of `_registerFile`. Thus, boundary checking needs to be added wherever needed so that no one tries to access a PseudoRealRegister in `_registerFile`. Doing so would cause undefined behaviour. 

This PR does the following:

-Update the isRealReg and isPseudoReg queries to accurately reflect what is a real register and what is not.
-Add boundary checks using the above queries wherever needed for `_registerFile`.
-Remove the initialization of an element that is outside the bounds of `_registerFile` to avoid future bugs.


dchopra001:
Marking WIP till tests + reviews are done. I might do a squash and rebase afterwards as well.

Review Request for @fjeremic .

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -6124,7 +6124,6 @@ OMR::Z::Machine::initialiseRegisterFile()
 
    // Initialize GPRs
    _registerFile[TR::RealRegister::NoReg] = NULL;
-   _registerFile[TR::RealRegister::SpilledReg] = NULL;
```
fjeremic: This change scares me a little. I understand that this is technically incorrect as indexing with `TR::RealRegister::SpilledReg` will go out of bounds of the `_registerFile` and will likely point into `_registerAssociations`

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRMachine.hpp#L184-L190

Now I cannot say for certain the specific field in `_registerAssociations` that gets accessed will be modified, but it is very likely it will actually remain `NULL` as we clear the associations at every extended basic block:

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRTreeEvaluator.cpp#L11877-L11882

This is because register associations typically don't hardcode the real register. They are typically left as `AssignAny` dependencies. `TR::RealRegister::SpilledReg` maps to a value of 91, and `TR::RealRegister::NumRegisters` maps to 81, which means we will be accessing the 10th element of `_registerAssociations` which is GPR9. I don't think we ever hardcode dependencies on GPR9 even for linkage so it is very unlikely GPR9 will appear on a dependency list, and hence it is very unlikely it will appear on an association.

This means `_registerFile[TR::RealRegister::SpilledReg]` will be very likely to be `NULL`. There are several places in the codegen that check whether this value is indeed `NULL` and act on it. Removing the above will have undefined consequences even though it is clearly incorrect.

Looking over the Power codegen and in the spirit of https://github.com/eclipse/omr/issues/2459 we should map `TR::RealRegister::NumRegisters` to represent the actual size of the enum and zero initialize all of these arrays for values which are not valid.

See the Power implementation:

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/p/codegen/OMRMachine.cpp#L140-L148

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/p/codegen/RealRegisterEnum.hpp#L217-L220

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -3007,7 +3007,7 @@ OMR::Z::Machine::findBestFreeRegister(TR::Instruction   *currentInstruction,
       preference = 0;
       }
 
-   uint64_t prefRegMask = _registerFile[preference] ? _registerFile[preference]->getRealRegisterMask() : 0;
+   uint64_t prefRegMask = TR::RealRegister::isRealReg((TR::RealRegister::RegNum)preference) ? _registerFile[preference]->getRealRegisterMask() : 0;
```
fjeremic: Just to elaborate a bit more on my concern, I also see a few other array accesses of `_registerFile` in this same function a little further below. These also seem scary to me as `preference` could clearly be a non-real register. There are also numerous other array accesses of `_registerFile` in other functions in the codegen and I'm not certain if we are exceeding bounds on those as well.

This is enough cause for concern that I think we should properly size these arrays and `NULL` initialize values which we deem not valid (all non-real registers).

compiler/z/codegen/OMRMachine.cpp
dchopra001: I had similar concerns when I was removing that line. However I decided to go with the removal because this is technically undefined behaviour. If this causes future defects then those should be fixed as well. I did not see any issues in my testing and so felt this was a safe enough change. I did not do any performance testing to see if this would have an effect there. If that is worth checking then let me know. 

> This means _registerFile[TR::RealRegister::SpilledReg] will be very likely to be NULL. There are several places in the codegen that check whether this value is indeed NULL and act on it. Removing the above will have undefined consequences even though it is clearly incorrect.

Can you point me to some instances where we check if `_registerFile[TR::RealRegister::SpilledReg]` is `NULL`? Before submitting this PR I went through all occurrences of `_registerFile` inside z/codegen and didn't see any instances where we check if `_registerFile[TR::RealRegister::SpilledReg]` is `NULL` (I don't think we ever do a null check on `_registerFile`). However I could have missed it considering there's like 400+ occurrences. 

Regarding being in sync with how the Power implementation does things, we may have some other issues to deal with first. RealRegisterEnum in the Power implementation lists only real registers (that correspond to actual hardware) except for `NoReg` and `SpilledReg`. On our side, there are concepts such as  `TR::RealRegister::EvenOddPair`:
https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/RealRegisterEnum.hpp#L161-L174

If we change `TR::RealRegister::NumRegisters` to map the actual size of the enum then situations like the following will become problematic:

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRMachine.cpp#L7482-L7507

I'm pretty sure the loop in the above function is only concerned with actual registers in hardware. So trying to access PseudoRealRegisters defined in `RealRegisterEnum.hpp` such as `TR::RealRegister::LegalOddPair` will result in undefined behaviour currently and segmentation faults when we define `_registerFile[TR::RealRegister::LegalOddRegister] = NULL`. 

There are a couple of workarounds to this issue. We can instead bound the loop to the last "real Register" which in our case is `TR::RealRegister::LastHPR`:
https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/RealRegisterEnum.hpp#L156-L161

Instead of LastHPR, we can add a new entry into the enum which holds the number of "RealRegisters" inside RealRegisterEnum.hpp

We can also borrow from the Power implementation: 
https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/p/codegen/OMRMachine.cpp#L1919-L1926

However in our case we would probably need to do `TR::RealRegister::NumRegisters - 14` (or 15?). This is kind of messy IMO. 

We can also add bound checks using the `isRealReg` and `isPseudoRealReg` queries that I have updated in this PR. However this might cause a performance issue because we will have to add these in many different locations.

dchopra001: The best long term solution in my opinion is to remove these "PseudoRealRegisters" from the `RealRegisterEnum` and add them into their own new enum: `PseudoRealRegisters`. The first thing this will do is avoid confusion. When I first started looking into this, `TR::RealRegister::EvenOddPair` sounded like a *real* register. However, I learned later that it's in fact just a concept and it will allow the RA to map to a real register eventually. The second thing this will do is make it clear that `_registerFile` is a list of *real* registers. If we define _registerFile's size to be similar to how it is defined in the Power side, then _registerFile will be a list of real registers, and a bunch of NULL entries at the end for PseudoRealRegisters. I don't believe _registerFile is meant to be used this way. For example, we are never supposed to access the `_registerFile[TR::RealRegister::LegalEvenOfPair]` element. Consider how we handle such pseudo registers inside `OMR::Z::Machine::findBestFreeRegister`: 

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRMachine.cpp#L3078-L3102

We never actually try and access `_registerFile[TR::RealRegister::LegalEvenOfPair]`, but instead `LegalEvenOfPair` indicates that we need to find an actual legal even register. 

I haven't looked into the other PseudoRegisters, but I suspect they will be similar. Thoughts?

compiler/z/codegen/OMRMachine.cpp
dchopra001: I don't believe an invalid array access is possible inside this function. The above line will set `prefRegMask` to 0 if `preference` is a non-real register. After this line, in every case where we try to access `_registerFile[preference]` we first check if `prefRegMask` is not 0. 

I agree with the concern though. This could be a problem in the future if new code is added and  proper boundary checking is not done.

compiler/z/codegen/OMRMachine.cpp
fjeremic: I do like the idea of moving these pseudo registers into a different enum (they are not real registers, and some of them are not registers at all so the name is still up in the air). I do support that idea and it is a much cleaner solution than bundling them as we do today. I suspect we placed them there originally because of convenience.

Never the less I think the primary question we need to answer is how to represent dependencies? These "extra" registers are predominantly used to specify dependencies. Unlike virtual registers dependencies can be associated with "AssignAny" or "EvenOddPair" for example. Register dependencies are effectively tuples where the first element is a virtual register and the second element is _x_. What is _x_ if we go down the route you propose?

Currently _x_ is of type `OMR::RealRegister:RegNum` which is the union of all real registers and these pseudo register things.

dchopra001: Will this new enum replace the original `RegNum` enum? Or will it be effectively "inheriting" from the RegNum enum?

fjeremic: > Will this new enum replace the original RegNum enum? Or will it be effectively "inheriting" from the RegNum enum?

It would be

compiler/z/codegen/OMRMachine.cpp
fjeremic: Yes, you're correct. All the paths are guarded indeed. I glossed over this very important detail. Thanks!

compiler/z/codegen/OMRMachine.cpp
dchopra001: I've restored this line back into the codebase : 3e8ab62a63402bf83fdd30d629348b5ad51c6f97

I'll open up a new issue + PR to address this issue as this problem will need to be solved in a larger context. For now we should be okay to merge with this "bug" since it doesn't break anything. I'll add the link for the new issue here for record keeping as well.

fjeremic:
Jenkins test all

fjeremic:
@genie-omr build zlinux zos

PR #2466, opened 2018-04-17
Fix ddr (PdbScanner.cpp) compilation with clang

samolisov:
Resolve the error of conversion string to char*

`clang` can not convert a non-trivial object of type `std::string` to
`char*` to use inside the `fprintf` function:

PdbScanner.cpp(1307,58) error: cannot pass non-trivial object of type
'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char>
>') to variadic function; expected type from format string was 'char *'

ERRMSG("Unhandled symbol returned by get_symTag: %s", symTagToString(symTag));

An explicit type conversion using the `c_str()` method is implemented.


ddr/lib/ddr-scanner/pdb/PdbScanner.cpp
```diff
@@ -223,30 +223,30 @@ PdbScanner::loadDataFromPdb(const wchar_t *filename, IDiaDataSource **dataSource
 		}
 
 		if (DDR_RC_OK != rc) {
-			ERRMSG("Creating instance of IDiaDataSource failed with HRESULT = %08X\n", hr);
+			ERRMSG("Creating instance of IDiaDataSource failed with HRESULT = %08lX\n", hr);
```
charliegracie: Is this correct for both 32 and 64 bit Windows compiles?  Just verifying that we didn't switch one warning for another.

charliegracie:
@genie-omr build all

samolisov:
@charliegracie Microsoft [says](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx) HRESULT is defined as LONG on Windows 32 too (if a type is different on Win 32 vs Win 64, this fact is explicitly declared). The length sub-specifier for LONG is 'l' and as far as I know, the specifier is regardless of the length of LONG itself.

PR #2474, opened 2018-04-18
Reverse #2463 'Resolve link issue when using ld'

jdekonin:
* change makes assumptions on library path that does not exist for all xlc versions
* lower optimization on cpuLoadAgent test to work with xlc13 without above change

Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>

### DIFF
--- a/fvtest/rastest/cpuLoadAgent.mk
+++ b/fvtest/rastest/cpuLoadAgent.mk
@@ -31,6 +31,10 @@ ifeq (win,$(OMR_HOST_OS))
 	OBJECTS += $(top_srcdir)/omr.res
 endif
 MODULE_INCLUDES += ../util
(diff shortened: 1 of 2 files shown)


charliegracie:
@genie-omr build all

jdekonin:
The files in xlc13 (and 12??) install to ```ls -al /opt/IBM/xlc/<version>/lib``` not ```/usr/vac/lib``` which was default for older releases.

charliegracie:
@genie-omr build aix

jdekonin:
Pushed a new change...somehow I made that an '02' and not an 'O2'...slight difference.  I wrapped the exports for aix only as well. :)

Also added a MODULE_CFLAGS because this folder contains both c and c++.

PR #2478, opened 2018-04-18
CMake: Fix assembly flags when compiling the jit

dnakamura:
No longer use CMAKE_ASM_ATT_FLAGS since we are using ASM not ASM-ATT.
Also put flags into its own variable to prevent polluting flags for
non-OMR components.


charliegracie:
@genie-omr build all

PR #2479, opened 2018-04-18
Add Object class to example

rwy0717:
Add a new class, Object, which represents the data stored in an object.
Update the omrobjectptr_t class to point to the object. Update the
object model to use this type.

Begin using the OMRApp namespace for new application-specific code.
Eventually, all glue will be defined under this name.

/cc @charliegracie @youngar


example/include/OMR/App/Object.hpp
```diff
@@ -0,0 +1,141 @@
...
+	fomrobject_t _value;
+};
+
+class Object {
+public:
+	explicit Object(size_t sizeInBytes, ObjectFlags flags = 0)
+		: _header(sizeInBytes, flags) {
+	}
+
+	ObjectHeader& header() { return _header; }
```
rwy0717: This use of return-by-reference is questionable. I think the best alternative would be to just make the header public, which gives us the same semantics.

example/include/OMR/App/Object.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+typedef uint8_t ObjectFlags;
+
+class ObjectHeader {
+public:
+	explicit ObjectHeader(size_t sizeInBytes, ObjectFlags flags = 0) {
+		assign(sizeInBytes, flags);
+	}
+
+	size_t sizeInBytes() const {
```
rwy0717: I'm avoiding the words get/set in accessors. let me know if you want this changed.

example/include/OMR/App/Object.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+	bool operator==(const ObjectSlotIterator& rhs) const { return _slot == rhs._slot; }
+
+	bool operator!=(const ObjectSlotIterator& rhs) const { return _slot != rhs._slot; }
+
+	ObjectSlotIterator& operator++()
+	{
+		++_slot;
+		return *this;
+	}
```
rwy0717: This class is trying to implement an idiomatic c++ forward iterator, which normally uses references and overloaded operators.

My intention is to use this as:

```C++
for (ObjectSlotIterator it = object->begin(); it != object.end(); ++it) {
  ObjectSlotHandle slot = *it;
  visitor.edge(object, slot);
}
```

And I might as well mention, it works with range-based for loops in cxx11:

```c++
for(ObjectSlotHandle slot : object) {
  visitor.edge(object, slot);
}
```

rwy0717: I ended up just using pointers as iterators, for the sake of simplicity.

samolisov:
@rwy0717 I see `OMR_FVTEST` depends on `OMR_EXAMPLE`, should the `OMR_EXAMPLE` flag be tested in the `fvtests/CMakeLists.txt` as you are doing it for `OMR_FVTEST` in the `example/CMakeLists.txt` (and maybe I'm wrong but I can't understand why `OMR_FVTEST` is required for examples)?

0xdaryl:
Minor stylistic question:  why is the OMR directory capitalized?  Even the root OMR project directory is lowercase.

rwy0717:
The example program requires google test from the fvtest component. You're right, we should be asserting in fvtest that the example is enabled. I'll make that change.

rwy0717:
> Minor stylistic question: why is the OMR directory capitalized? Even the root OMR project directory is lowercase.

Since this is a mechanism for "namespacing" directories, I was going for a 1:1 with c++ namespaces. I have no problems switching to lowercase `<omr/app/Whatever.hpp>`,  if that's what we want.

There's the other question of, do we want the app namespace to also designate the component. Maybe I should be using the OMR::GCApp:: namespace for GC glue code? I wouldn't suggest OMR::GC::App or OMR::App::GC, each can cause uncertainty.

samolisov:
@rwy0717 I'm sorry, the review function is not available on my tablet PC. Are you sure the omr_example_core_glue library should "links" to the omrgc, not the omr_example_gc_glue (see the example/glue/CMakeLists.txt file).

rwy0717:
>I'm sorry, the review function is not available on my tablet PC

No worries @samolisov, thanks for taking a look!

The gc-glue provides interface sources, which are compiled and linked into the gc library. Linking a glue component more than once would make it's sources compile into multiple libraries, which is not what we want.

You do have a point though, using the core glue to indirectly link omrcore against omrgc is the wrong thing--omrcore should just link against omrgc. I've fixed this.

samolisov:
@rwy0717 My point was just remove omrgc from omcore and omr_core_glue_lib in order to make omrcore independent from omrgc if it's possible. Maybe omr_gc_glue_lib should depend on omrgc (or be linked to omrgc, I can be a confused a bit, the concept of glue libraries is a new for me, as I get this is a way for consumers to link their own code into OMR libraries). If we link omrcore and omrgc, could we also link omrcore with other its dependencies: omrtraces, hook library and so on? Could you have a look at issue #2481 please?

rwy0717:
OK, I see what you're saying. Lets move the discussion to the other issue, I think there's quite a bit to be fixed around omrcore.

samolisov:
@rwy0717 I've re-check the PR and now I'm not sure `FVTEST` mush fail if `OMR_EXAMPLE=OFF`. If a customer provides his/her glue libraries and therefore the corresponding tasks, `FVTEST` should build against these tasks and their parameters, the `example/glue` module won't be required. Am I missing something?

rwy0717:
Breaking this PR into smaller PRs--the first is #2495, which contains only the cmakelist changes.

rwy0717:
I've reduced this commit to just adding the `Object` class to the example.

charliegracie:
@genie-omr build all

example/glue/objectdescription.h
```diff
@@ -19,17 +19,26 @@
...
-typedef uintptr_t* languageobjectptr_t;
-typedef uintptr_t* omrobjectptr_t;
+
+namespace OMRApp {
+
+class Object;
+
+}  /* namespace OMRApp */
+
+typedef OMRApp::Object* languageobjectptr_t;
```
charliegracie: Will this typedef work if the file is included in a C file and not a C++ file?  I believe you have properly defined it but I am just wondering if you have verified that it will work in a C file.

charliegracie:
@rwy0717 do you see OMR::App::Object replacing MM_ObjectModel?  If so I think we may want to call it ObjectModel (at some point) as that more describes what it is.

rwy0717:
Sorry, I guess I should have given the motivation for this PR. This shows that a language can implement their ObjectModel in terms of an `Object` class. `Object` is for the application only, is not glue code, and is not used (directly) by the GC. This class can't/won't replace the ObjectModel--often, objects aren't representable by structs, and many operations simply do not work as member functions.  In the example's case, I see some basic ObjectModel operations becoming thin wrappers around `object->operation()` calls. This is largely a "superficial" change... the only functional addition are new slot iteration functions. I think having an `Object` class just makes the example a little more obvious and understandable.

example/glue/objectdescription.h
rwy0717: No this won't work--namespaces will break it. Is c compatibility needed/wanted in the example? I can change this, but it's nice that the `objectptr` types are literally `OMRApp::Object *`.

rwy0717:
Oh and there is one other improvement, the example now supports the compressed header and compressed references flags independently. Previous to this commit, both flags had to be enabled together.

example/glue/CollectorLanguageInterfaceImpl.cpp
```diff
@@ -207,7 +207,11 @@ MM_CollectorLanguageInterfaceImpl::scavenger_reverseForwardedObject(MM_Environme
 
 		/* Restore the original object header from the forwarded object */
 		GC_ObjectModel *objectModel = &(env->getExtensions()->objectModel);
-		objectModel->setObjectSizeAndFlags(originalObject, objectModel->getConsumedSizeInBytesWithHeader(forwardedObject), objectModel->getObjectFlags(forwardedObject));
+
```
rwy0717: There's some spacing here that I have to fix.

charliegracie:
There is a merge conflict @rwy0717.  once that is resolved I will merge

rwy0717:
OK. I can fix the conflict. Before accepting this PR, we should decide on whether to use the OMRApp namespace (which I am using in this PR). That or I can remove any namespacing from this PR, to give us time to figure it out. The discussion is #2496.

charliegracie:
I would remove the namespacing from this PR until we settle on naming.

charliegracie:
ping :)

charliegracie:
ping?

rwy0717:
OK I think this is good to go, just needs to pass CI.

charliegracie:
@genie-omr build all

PR #2483, opened 2018-04-19
Refactor options; add option to print blacklisted type names

keithc-ca:
When wildcards are used, it isn't always obvious which types are effectively blacklisted; this introduces options (`-sb` or `--show-blacklisted`) to request that set be shown.

The parameter list to `getOptions` was already very long; rather than add to it, I refactored option management into a new class.


charliegracie:
@genie-omr build all

Issue #2465, opened 2018-04-17
AArch64: Implement initial version of OMRMachine class


knn-k:
Start writing code for AArch64 with OMRMachine and related classes, based on the stub from https://github.com/eclipse/omr/pull/2453

PR #2480, opened 2018-04-19
AArch64: Initial version of RealRegister-related files

knn-k:
This commit implements the initial version of RealRegister-related
files for aarch64.
This is a part of #2465.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #2484, opened 2018-04-20
AArch64: Initial version of OMRMachine files

knn-k:
This commit implements the initial version of OMRMachine files for
aarch64.
This is a part of #2465.

### DIFF
--- a/compiler/aarch64/codegen/OMRMachine.cpp
+++ b/compiler/aarch64/codegen/OMRMachine.cpp
@@ -19,11 +19,547 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
+#include <stddef.h> // for NULL, etc
+#include <stdint.h> // for uint16_t, int32_t, etc
+
 #include "codegen/CodeGenerator.hpp"
 #include "codegen/Machine.hpp"
+#include "codegen/Machine_inlines.hpp"
+#include "codegen/RealRegister.hpp"
+#include "infra/Assert.hpp"
 
 OMR::ARM64::Machine::Machine(TR::CodeGenerator *cg) :
-      OMR::Machine(cg)
+   OMR::Machine(NUM_ARM64_GPR, NUM_ARM64_FPR), _cg(cg)
    {
+   _registerFile = (TR::RealRegister **)cg->trMemory()->allocateMemory(sizeof(TR::RealRegister *)*TR::RealRegister::NumRegisters, heapAlloc);
+   self()->initialiseRegisterFile();
+   }
 
+TR::RealRegister *OMR::ARM64::Machine::findBestFreeRegister(TR_RegisterKinds rk,
+                                                            bool considerUnlatched)
+   {
+   int32_t first;
+   int32_t last;
+
+   switch(rk)
+      {
+      case TR_GPR:
+         first = TR::RealRegister::FirstGPR;
+         last  = TR::RealRegister::LastAssignableGPR;
+         break;
+      case TR_FPR:
+         first = TR::RealRegister::FirstFPR;
+         last  = TR::RealRegister::LastFPR;
+         break;
+      default:
+         TR_ASSERT(false, "Unsupported RegisterKind.");
+   }
+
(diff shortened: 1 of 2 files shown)


knn-k:
This PR depends on https://github.com/eclipse/omr/pull/2480 .

compiler/aarch64/codegen/OMRMachine.cpp
```diff
@@ -19,11 +19,457 @@
...
+            first++;
+         last  = TR::RealRegister::LastAssignableGPR;
+         break;
+      case TR_FPR:
+         first = TR::RealRegister::FirstFPR;
+         last  = TR::RealRegister::LastFPR;
+         break;
+   }
+
+   uint32_t bestWeightSoFar = 0xffffffff;
```
0xdaryl: At some point we need to decide whether register weights make sense for ARM or would have any benefit.  We can't necessarily predict how some unforeseen linkages or code paths may want to influence the register selection choices, so I'm fine with leaving the basic infrastructure in place.

compiler/aarch64/codegen/OMRMachine.cpp
```diff
@@ -19,11 +19,457 @@
...
+
+   return NULL;
+   }
+
+void OMR::ARM64::Machine::initialiseRegisterFile()
+   {
+   _registerFile[TR::RealRegister::NoReg] = NULL;
+   _registerFile[TR::RealRegister::SpilledReg] = NULL;
+
+   _registerFile[TR::RealRegister::x0] = new (self()->cg()->trHeapMemory()) TR::RealRegister(TR_GPR,
```
0xdaryl: For all of these initializations you have both `self()->cg()` and `_cg`.  We should be consistent.  I'd say go with `self()->cg()`.

compiler/aarch64/codegen/OMRMachine.cpp
```diff
@@ -19,11 +19,457 @@
...
+         break;
+      case TR_FPR:
+         first = TR::RealRegister::FirstFPR;
+         last  = TR::RealRegister::LastFPR;
+         break;
+   }
+
+   uint32_t bestWeightSoFar = 0xffffffff;
+   TR::RealRegister *freeRegister = NULL;
+   for (int i = first; i <= last; i++)
```
0xdaryl: int32_t for consistency

Leonardo2718: How about `auto` instead? :grin:

0xdaryl:
A genie build is not needed as these files are AArch64-specific.

compiler/aarch64/codegen/OMRMachine.cpp
```diff
@@ -19,11 +19,457 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
+#include "aarch64/codegen/OMRMachine.hpp"
```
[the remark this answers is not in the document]
knn-k: This came from the "p" version of OMRMachine.cpp.

compiler/aarch64/codegen/OMRMachine.cpp
```diff
@@ -19,11 +19,457 @@
...
 OMR::ARM64::Machine::Machine(TR::CodeGenerator *cg) :
-      OMR::Machine(cg)
+      OMR::Machine(NUM_ARM64_GPR, NUM_ARM64_FPR)
+   {
+   _registerFile = (TR::RealRegister **)cg->trMemory()->allocateMemory(sizeof(TR::RealRegister *)*TR::RealRegister::NumRegisters, heapAlloc);
+   self()->initialiseRegisterFile();
+   }
+
+TR::RealRegister *OMR::ARM64::Machine::findBestFreeRegister(TR_RegisterKinds rk,
+                                                            bool excludeGPR0,
```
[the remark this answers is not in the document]
knn-k: Both "p" and "arm" versions have `excludeGPR0`, while the "arm" version does not seem to use it.
It is needed in "p" because `r0` is used as the zero register depending on the instruction?

knn-k:
Reflected review comments.
The functions with `"Not implemented yet."` asserts require `MemoryReference`, instruction generation, etc., so I leave them unimplemented for now.

PR #2487, opened 2018-04-20
Remove deprecated code related to ArrayCmp on X86

0dvictor:
Some code related to ArrayCmp are either long deprecated
or never used. Removing them.


0xdaryl:

PR #2476, opened 2018-04-18
Revert "Add new portlib signal flags to support new signals"

babsingh:
Reverts eclipse/omr#2421

OpenJ9 hangs during compilation - https://github.com/eclipse/openj9/issues/1711

[skip ci]

Signed-off-by: Babneet Singh <sbabneet@ca.ibm.com>

PR #2494, opened 2018-04-23
Extend OMR Signal API

babsingh:
1) Added two new port library signal flags in omrport.h: SIGHUP and
SIGCONT.

2) Changed the signature of omrsig_set_single_async_signal_handler.
Before, it returned the address to the old OS handler. Address to the
old OS handler can be NULL in non-error situations. So, it wasn't
possible to distinguish between error and non-error scenarios. Now, it
is changed to return 0 on success and non-zero on failure. The address
of the old OS handler is returned via an input reference variable.

3) Return macro OMRPORT_SIG_ERROR (-1) from
omrsig_map_portlib_signal_to_os_signal.

4) Added Unix (port/unix/omrsignal.c) implementation of
omrsig_set_single_async_signal_handler,
omrsig_map_os_signal_to_portlib_signal and
omrsig_map_portlib_signal_to_os_signal. Supporting trace points have
been added in omrport.tdf.

5) In Unix omrsignal.c, removed unused macro MAX_PORTLIB_SIGNAL_TYPES.

6) Fix issue with re-registering master handlers. OMRSIG_SIGACTION can
be called outside of omrsignal.c functions. A call to OMRSIG_SIGACTION
will override the master handlers set by either omrsig_protect or
omrsig_set_async_signal_handler. Setting a master handler once sets a
bit in signalsWithMasterHandlers. If the master handler is overridden by
an external call to OMRSIG_SIGACTION, then the bit in
signalsWithMasterHandlers will not be un-set. Since the bit in
signalsWithMasterHandlers isn't un-set once the master handler has been
overridden, then subsequent calls to omrsig_protect and
omrsig_set_async_signal_handler will not re-register the master handler.
In subsequent calls, it will look as if the master handler is already
registered. We mustn't rely on signalsWithMasterHandlers to re-register
master handlers. It doesn't cause any harm to register master handlers
every time a call is made to either omrsig_protect or
omrsig_set_async_signal_handler.

7) Initialize oldAction before invoking OMRSIG_SIGACTION. Before
registering handlers using OMRSIG_SIGACTION, oldAction (struct
sigaction) needs to be properly initialized or reset within
registerSignalHandlerWithOS.

8) omrsig_set_single_async_signal_handler needs access to the old OS
handler. omrsig_set_single_async_signal_handler,
omrsig_set_async_signal_handler and omrsig_protect call
registerMasterHandlers to register master handlers with the OS.
registerMasterHandlers calls registerSignalHandlerWithOS.
registerSignalHandlerWithOS has access to the old OS handler. The old OS
handler is returned via an input reference variable from
registerSignalHandlerWithOS to registerMasterHandlers, and then to any
caller of registerMasterHandlers. If the input reference variable is
NULL, then the input reference variable is not updated. To allow this,
the definitions of registerMasterHandlers and
registerSignalHandlerWithOS has been updated.

9) In Unix omrsignal.c, oldActions is used to cache the original OS
handler so that it can be restored during shutdown.
registerSignalHandlerWithOS registers a handler with the OS using
OMRSIG_SIGACTION. While registering a handler with the OS, it gets
access to the old OS handler. Now, registerSignalHandlerWithOS can be
called multiple times to register the master handlers (refer to point
6). So, it should cache the old (original) OS handler in oldActions
while registering the first handler. While registering subsequent
handlers, oldActions isn't updated.

10) Added trace point in mapUnixSignalToPortLib to print error for
unknown OS signal.

11) MAX_UNIX_SIGNAL_TYPES corresponds to the total number of signals
available. All signals have a value greater than 0.
MAX_UNIX_SIGNAL_TYPES should not be used to create an array for storing
information related to signals. Instead, (MAX_UNIX_SIGNAL_TYPES + 1)
should be used to create an array for storing information related
signals. This will allow us to properly store the information for the
MAX_UNIX_SIGNAL_TYPES(th) signal. Macro ARRAY_SIZE_SIGNALS has been
added to create arrays for storing information related to signals. 

12) In Unix omrsignal.c, signalMap has been updated to support three
more signals: SIGINT, SIGHUP and SIGCONT. All three are asynchronous
signals.

13) Semaphores are no longer used to track signal counts. Instead, an
array (signalCounts) has been added to keep track of signal counts.

14) masterASynchSignalHandler no longer updates semaphores to keep track
of signal counts. Instead, it updates the signal specific array element
within signalCounts via atomic operation. Then, it notifies the
asynchSignalReporterThread.

15) asynchSignalReporter has been modified. The code to invoke handlers
has been moved into another function named runHandlers.
asynchSignalReporter no longer relies upon semaphores to determine which
signal should be handled. Instead, it checks the array elements of
signalCounts to determine the signal to be handled. After invoking
associated handlers, it decrements the signal specific array element via
atomic operation.

16) The code to initialize and destroy unneeded semaphores has been
removed from initializeSignalTools and destroySignalTools.

17) registerMasterHandlers uses a constant 4 without a clear explanation. 
Added a macro OMRPORT_SIG_SMALLEST_SIGNAL_FLAG to represent the constant 4. 
Added an explanation for the usage of OMRPORT_SIG_SMALLEST_SIGNAL_FLAG in registerMasterHandlers.


babsingh:
@charliegracie This pull request includes changes which were reverted in https://github.com/eclipse/omr/pull/2475. 

The cause of the previous OpenJ9 intermittent hang was: https://github.com/eclipse/omr/pull/2476. 

The following OpenJ9 pull request depends on this pull request: https://github.com/eclipse/openj9/pull/1642 and https://github.com/eclipse/openj9/pull/1603.

charliegracie:
There seem to be a few logically distinct changes in this PR.  I am having trouble following the change and associate the different changes with the reason they are being made.  Would it be possible to break this into a few commits which could be reviewed individually?  You can still use this PR but instead of having one commit for everything just force push a few commits that break the change into its logical parts.

babsingh:
@charliegracie I have broken the large commit into smaller logical commits.

babsingh:
@charliegracie Currently, `Trc_PRT_signal_omrsig_set_single_async_signal_handler_exiting_did_nothing_possible_error` is labelled as a `TraceExit`, and `Trc_PRT_signal_mapOSSignalToPortLib_ERROR_unknown_signal` is labelled as a `TraceEvent`. Both trace points cover error cases. Should they be labelled as `TraceException`?

pshipton:
FYI TraceEntry must be pared with TraceExit. A TraceExit cannot be changed to TraceException, but to TraceExit-Exception.

charliegracie:
@genie-omr build all, zos

@pshipton are you ok with the changes as-is?

babsingh:
I would like to change `Trc_PRT_signal_omrsig_set_single_async_signal_handler_exiting_did_nothing_possible_error` to a `TraceExit-Exception` and `Trc_PRT_signal_mapOSSignalToPortLib_ERROR_unknown_signal` to `TraceException`. @charliegracie will you be opposed to this change?

charliegracie:
I think that makes sense.

babsingh:
I have changed `Trc_PRT_signal_omrsig_set_single_async_signal_handler_exiting_did_nothing_possible_error` to a `TraceExit-Exception` and `Trc_PRT_signal_mapOSSignalToPortLib_ERROR_unknown_signal` to a `TraceException`.

pshipton:
A few minor comments

pshipton:
Also note the port library signal tests will need to be updated, although a separate PR for this works for me.

port/unix/omrsignal.c
```diff
@@ -488,22 +471,114 @@ omrsig_set_async_signal_handler(struct OMRPortLibrary *portLibrary, omrsig_handl
...
 {
-	return NULL;
+	uint32_t rc = 0;
+	J9UnixAsyncHandlerRecord *cursor = NULL;
+	J9UnixAsyncHandlerRecord **previousLink = NULL;
+	BOOLEAN foundHandler = FALSE;
+
+	Trc_PRT_signal_omrsig_set_single_async_signal_handler_entered(handler, handler_arg, portlibSignalFlag);
+
+	if (0 != portlibSignalFlag) {
```
pshipton: Please update the function doc to explain the purpose of passing portlibSignalFlag of zero.

port/unix/omrsignal.c
```diff
@@ -699,119 +727,55 @@ asynchSignalReporter(void *userData)
...
-		/* we get woken up if there is a signal pending or it is time to shutdown */
-		if (0 != shutDownASynchReporter) {
-			break;
-		}
-#else /* defined(J9ZOS390) */
+	while (0 == shutDownASynchReporter) {
+		int unixSignal = 1;
+		uint32_t asyncSignalFlag = 0;
 
+#if !defined(J9ZOS390)
```
pshipton: this could just be an #else

babsingh: inside the `while` loop, there is no prior`ifdef J9ZOS390`. do you want me to move the `sem_wait` after the `for` loop along with the `cond_wait`?

port/unix/omrsignal.c
babsingh: updated  function doc.

port/unix/omrsignal.c
pshipton: No. Sorry, I misread the folded code.

port/common/omrsignal.c
```diff
@@ -174,21 +174,27 @@ omrsig_set_async_signal_handler(struct OMRPortLibrary *portLibrary, omrsig_handl
...
+ * entry corresponding to omrsig_handler_fn is removed, and related resources are freed. portlibSignalFlag can only
+ * have one signal flag set; otherwise, OMRPORT_SIG_ERROR is returned. One omrsig_handler_fn handler is registered
+ * with a signal at any time instead of multiple handlers. When associating a new omrsig_handler_fn with a signal,
+ * prior omrsig_handler_fn(s) are dissociated from the signal. The address of the old signal handler function is
+ * stored in oldOSHandler. This function supports all signals listed in OMRPORT_SIG_FLAG_SIGALLASYNC.
  *
  * @param[in] portLibrary The port library
  * @param[in] handler the function to call if an asynchronous signal arrives
  * @param[in] handler_arg the argument to handler
  * @param[in] portlibSignalFlag port library signal flag
```
pshipton: This line could be clearer. i.e. something like "a single port library signal constant, or 0 to remove the signal handler"

pshipton:
@charliegracie I'm good with these changes once Babneet makes one remaining doc update

port/common/omrsignal.c
babsingh: updated doc. changed to "a single port library signal flag, or 0 to remove the omrsig_handler_fn entry".

charliegracie:
@genie-omr build all

pshipton:
@charliegracie everything was addressed

PR #2495, opened 2018-04-23
Rework the Example cmakelists

rwy0717:
Make the default glue NOTFOUND, and assert that the glue is defined,
whenever it is required. Only use the example glue if OMR_EXAMPLE is on.
OMR_EXAMPLE is on by default.

Compile the example program. We are not able to run example as a test,
due to an old hang.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, 2017 IBM Corp. and others
+# Copyright (c) 2017, 2018 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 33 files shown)


samolisov:
@rwy0717 in the previous PR you linked omrcore against omrgc if OMR_GC is
enabled. Now you are use interface export of omrgc in
omr_example_core_glue? What do you think about return as it was? I have no
idea why core_glue must propagate omrgc headers.

>

rwy0717:
Oh right, I lost that change when I was splitting up my commits. Good catch, it's fixed.

samolisov:
@rwy0717 Great job! Waiting for the merge of this PR.
>
>

charliegracie:
@genie-omr build all

samolisov:
As you write, LGFM :)

rwy0717:
I've made a mistake, the example is explicitly enabled in the travis and appveyor builds, but is not on by default. Will push a fix.

dnakamura:
LGTM

rwy0717:
Fixed the merge conflict, and also enabled `OMR_FVTEST` by default, since it's required by the example.

charliegracie:
I would have expected FVTest would require example and not the other way around.  WHat does example rely on in FVTest?

rwy0717:
The example needs the `testMain` wrapper from the omrGtestGlue. Maybe some other things.

charliegracie:
hmmm I do not think it should.  I do not see it directly including `omrTest.h`. Would it be possible to quickly remove that?  It should not be using any FV features as it is an example and not an FV Test.

samolisov:
Maybe we should not consider example as a test (testMain) and it could be better just to implement example as a standalone demo application that demonstrates how to use GC and other components in an OMR based customer application?

rwy0717:
The omrGTestGlue provides a [fairly complex `main` function](https://github.com/eclipse/omr/blob/master/fvtest/omrGtestGlue/argmain.cpp). How do we feel about pulling this code out of FV, and into a standalone util library? Beyond testing and example, this could be used in our build tools, which seem to duplicate this code almost verbatim (here's [hookgen's version](https://github.com/eclipse/omr/blob/master/tools/hookgen/main.cpp)).

The library could be called `omr_main`, and it would provide a `main` function that translates arguments before calling out to a platform-agnostic `omr_main_entry` function. While it's an unusual utility, it could reduce code duplication by quite a bit.

charliegracie:
I would be ok with that as long as we shared it with hookgen, etc.

charliegracie:
@genie-omr build all zos

example/CMakeLists.txt
```diff
@@ -19,15 +19,35 @@
...
 add_executable(example
 	main.cpp
 )
 
 target_link_libraries(example
-	PRIVATE
-	omrgc
+	omr_main_function
+	pugixml
+	omrtestutil
```
charliegracie: Is omrtestutil still required now that you moved omr_main_function out into a separate library?

charliegracie: Is the thumbs up a yes that it is still required or a yes I will remove it since it isn't required anymore?

rwy0717: It needs to be removed. I'll get to it today.

rwy0717: done.

charliegracie:
@genie-omr build all

charliegracie:
Thank you @rwy0717 this is a great improvement!

PR #2498, opened 2018-04-24
CMake cleanup

charliegracie:
Resolve issues with empty GC compilation units. This was completed by deleting completely empty cpp files and also by only compiling files when their feature flag was enabled.  Not all were resolved in this changeset as there needs to be some discussion on the proper way to do the rest. 

Cleanup the AppVeyor cache.


charliegracie:
@genie-omr build all zos

samolisov:
@charliegracie <https://github.com/charliegracie> Why don't we use
`target_sources()` in order to configure a source list dependent on options?

2018-04-24 18:09 GMT+03:00 Charlie Gracie <notifications@github.com>:

charliegracie:
@samolisov I just pushed a new commit.  Is this what you mean?

@genie-omr build all

samolisov:
@charliegracie Yes, I think this CMakeLists.txt looks  better and matched to the new best practices.

fjeremic:
> @charliegracie Yes, I think this CMakeLists.txt looks better and matched to the new best practices.

@samolisov I'm a newbie when it comes to CMake. I can never find any official best practices or a guide to "the right way™" of doing CMake. Do you have a link to the best practices you're referring to?

dnakamura:
@charliegracie  LGTM
@fjeremic  I'm not sure its actually documented anywhere but  i believe `target_sources()` is preferred over building a list (which is a practice which originates from before `target_sources()` was added). However a lot of the cmake stuff builds a list variable for 2 reasons. 1) to work with our VPATH hack. 2) to simplify the porting since it maps 1:1 with the makefiles

youngar:
Initially, it looks good! I'm going to scan through the source lists vs compile flags and see if anything wrong jumps out at me.  After that I'll merge this.

samolisov:
@fjeremic It's not official and maybe "best practices" sounds very loud, I was excted about target_sources by this article: https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/

PR #2499, opened 2018-04-24
Fixing dcmplt and dcmpgt Instruction handling on PPC

alenbadel:
dcmplt and dcmpgt (Along with fcmpgt and fcmplt) use setbool in order to compare and set the required bit results.

Prior, the sequence setbool expansion was implemented incorrectly. 
-fcmpu compares input arguments, the result is stored in the conditional register (Bits LT,GT,EQ, and SO). 
-setb reads the condition register and set the bit result based on the LT and GT fields.
-setb checks bit set in the LT field will result to the output -1 (or 0xfff) if it's high. 

Therefore, in some situations the result was incorrect. Since when the LT bit is high, the result bit is not set correctly. No matter if dcmplt or dcmpgt is called. 

With this change:
-only with dcmplt and fcmplt , we move LT bit of the condition register to the GT bit. 
-When running dcmpgt and fcmpgt, moving is not needed since we already have the desired state set in the GT bit of the condition register
-The LT bit is now cleared using crxor. 

Now with setb, the result will be based off of GT. setb will set high if GT is high, and low if GT is low. 

The binary encoding for condition register logical instructions were also fixed on PPC. 

Signed-off-by: Alen Badel <Alen.Badel@ibm.com>

### DIFF
--- a/compiler/p/codegen/FPTreeEvaluator.cpp
+++ b/compiler/p/codegen/FPTreeEvaluator.cpp
@@ -2010,7 +2010,7 @@ TR::Register *OMR::Power::TreeEvaluator::dcmpltEvaluator(TR::Node *node, TR::Cod
    int64_t imm = 0;
    if (TR::Compiler->target.cpu.id() >= TR_PPCp9)
       {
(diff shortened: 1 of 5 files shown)


0xdaryl:
@gita-omr @ymanton : would one of you mind reviewing for Power correctness please?

0xdaryl:
@genie-omr build all

gita-omr:
Might be worth mentioning that SO bit was incorrectly ORed into the LT bit and that was fixed as well. Also, all these problems existed only on P9 and higher.

gita-omr:
@0xdaryl could you please add arch:power label?

PR #2501, opened 2018-04-25
CMake: Detect git SHA

dnakamura:
If git is found (and source dir is a git repo) set OMR_VERSION_STRING
to the git SHA. Otherwise fall back to old behavior of "<Unknown>"


charliegracie:
@dnakamura  If we tag the repo (for a release, etc.) and you checkout the tag will this pick up the tag name or just the SHA value?

dnakamura:
just the SHA

charliegracie:
sadness.  It would have been great if it would be the SHA for a normal checkout and if you checked out a TAG that the tag would show up.  Oh well this is better than what we had :)

dnakamura:
Its probably doable, I just need to dig into the docs for the various git commands. I just stole this from `openj9-openjdk-jdk9`

PR #2502, opened 2018-04-25
CMake: Enable exception handling

samolisov:
Clang/LLVM isn't able to compile the compilator's code when exception handling
is not enabled:

```
error: cannot use 'throw' with exceptions disabled
       throw Exception();
```

Warning C4291:

```
..\omr\OMR_Agent.cpp(145): warning C4291: 'void *OMR_Agent::operator
new(::size_t,void *)': no matching operator delete found; memory will
not be freed if initialization throws an exception
omr\include_core\OMR_Agent.hpp(136): note: see declaration of
'OMR_Agent::operator new'
```

has also been fixed.

### DIFF
--- a/cmake/modules/platform/toolcfg/msvc.cmake
+++ b/cmake/modules/platform/toolcfg/msvc.cmake
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, 2017 IBM Corp. and others
+# Copyright (c) 2017, 2018 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
@@ -71,6 +71,17 @@ endif(OMR_ENV_DATA64)
 
 # Configure the platform dependent library for multithreading
 set(OMR_PLATFORM_THREAD_LIBRARY Ws2_32.lib)
+
+# TR_CXX_COMPILE_OPTIONS are appended to CMAKE_CXX_FLAGS, and so apply only to
+# C++ file compilation
+list(APPEND TR_CXX_COMPILE_OPTIONS
+	/EHsc   # Enable exception handling (Clang doesn't enable exception handling by default)
+)
+
+# TR_C_COMPILE_OPTIONS are appended to CMAKE_C_FLAGS, and so apply only to
+# C file compilation
+list(APPEND TR_C_COMPILE_OPTIONS
+)
 
 macro(omr_toolconfig_global_setup)
 	# Make sure we are building without incremental linking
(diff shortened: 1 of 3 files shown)


rwy0717:
It would be better if exceptions were an optional feature, possibly called `OMR_CXX_EXCEPTIONS` in cmake. We would also want to respect this option on all supported platforms, not just windows.

To explain, most of our code does not use exceptions at all, and is not exception safe. Exceptions also require some symbols from the C++ std lib, which j9 can not link against. We are stuck supporting `-nostdlib`.

samolisov:
@rwy0717 Do you mean I should add `EHsc` to the compiler options only if the `OMR_CXX_EXCEPTIONS` parameter is enabled or I should define the parameter and use this as a cross-compiler option to enable exceptions? 

Most of the code doesn't use exceptions but the compiler code does and couldn't be built on clang without explicitly enabling the feature (I believe MSVC just enables a default exception handling mechanism as it's described [in the documentation](https://msdn.microsoft.com/en-us/library/1deeycx5.aspx) but clang doesn't have the defaults).

rwy0717:
Oh, you're trying to build the compiler. You should add the flag to `TR_CXX_COMPILE_OPTIONS` in the `cmake/platform/toolconfig/msvc.cmake`.

samolisov:
@rwy0717 I've moved the compilator flag to TR_CXX_COMPILE_OPTIONS, could you have a look again?

rwy0717:
tiny nitpick: you used 4 spaces to indent, should be 1 tab.
otherwise LGTM.

samolisov:
Oops! Fixed.

rwy0717:
Hey @samolisov, it looks like you're running up against the "Eclipse IP validation checker". Have you signed the eclipse contributor agreement? Here are the instructions, in case you haven't (found in our contributing guide):

1. [Create an account at Eclipse](https://dev.eclipse.org/site_login/createaccount.php).
2. Add your GitHub user name in your account settings.
3. [Log into the project's portal](https://projects.eclipse.org/) and sign the ["Eclipse ECA"](https://projects.eclipse.org/user/sign/cla).
4. Ensure that you [_sign-off_](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Signing_off_on_a_commit) your Git commits.
5. Ensure that you use the _same_ email address as your Eclipse account in commits.

samolisov:
@rwy0717 Thank you. It looks like a bug in the ECA checking mechanism: I've login on my Eclipse page and see that my ECA record is active until 2021. I've just pushed an updated commit to trigger the checking mechanism.

rwy0717:
LGTM

PR #2506, opened 2018-04-26
Store addressable physical memory in GC_ExtensionsBase

ashu-mehra:
Added a new field 'usablePhysicalMemory' in GC_ExtensionsBase to
store the value returned by omrsysinfo_get_addressable_physical_memory().

### DIFF
--- a/gc/base/GCExtensionsBase.cpp
+++ b/gc/base/GCExtensionsBase.cpp
@@ -102,6 +102,10 @@ MM_GCExtensionsBase::initialize(MM_EnvironmentBase* env)
 	excessiveGCStats.endGCTimeStamp = omrtime_hires_clock();
 	excessiveGCStats.lastEndGlobalGCTimeStamp = excessiveGCStats.endGCTimeStamp;
 
+        /* Get usable physical memory from portlibrary. This is used by computeDefaultMaxHeap().
+	 * It can also be used by downstream projects to compute project specific GC parameters.
+	 */
+	usablePhysicalMemory = omrsysinfo_get_addressable_physical_memory();
 
 	computeDefaultMaxHeap(env);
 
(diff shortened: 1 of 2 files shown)


ashu-mehra:
As per the comment here https://github.com/eclipse/openj9/pull/1699#discussion_r183835089 it would be useful if sysinfo_get_addressable_physical_memory() is stored in GC_ExtensionsBase, so that downstream projects won't have to call the portlib API multiple time.
fyi - @dmitripivkine @charliegracie

ashu-mehra:
This change would help to avoid at least two calls to `omrsysinfo_get_addressable_physical_memory()` in OpenJ9 PR https://github.com/eclipse/openj9/pull/1699 in `mminit.cpp:gcInitializeDefaults()` and `MM_GCExtensions::computeDefaultMaxHeap()`. Both these functions read the value from Extensions.

charliegracie:
I see no reason to not cache this so I will merge the change

charliegracie:
Hi @ashu-mehra there is a type mismatch between the variable `usablePhysicalMemory` and the function `omrsysinfo_get_addressable_physical_memory`. Can you please resolve this?

Also there appears to be a tabbing issue [here](https://github.com/eclipse/omr/pull/2506/files#diff-9f61477d168dc53989390dec0f61cff9R105) that I missed in the original review.  Can you also fix it please?

Also can you explain the difference between the fields `MM_GCExtensionsBase->memoryMax`, `MM_GCExtensionsBase->maxSizeDefaultMemorySpace` and `MM_GCExtensionsBase->memoryMax`?

ashu-mehra:
> Also can you explain the difference between the fields MM_GCExtensionsBase->memoryMax, and MM_GCExtensionsBase->maxSizeDefaultMemorySpace?

Going purely by their usage in current code base I am not sure if there is any difference between the two.
In OMR I don't see any use of `maxSizeDefaultMemorySpace`. Probably downstream projects like OpenJ9 may require, but even there I don't see how `maxSizeDefaultMemorySpace` is different from `memoryMax`.

PR #2507, opened 2018-04-26
 Modify shutdown sequence to fix CS issue

ebadyano:
Fixing two issues with Concurrent Scavenger:
1. Modified shutdown sequence to do Scavenger tear down later in the
sequence; after all of the mutator threads are done. 
2. Added a new case in flushGCCaches to clear a GC Slave thread
copy cache when starting up with many GC
threads and GC slave thread is brought up during active scavenger cycle
but doesn't do any actual work and so is actually acting like a mutator
thread until the next cycle.

### DIFF
--- a/gc/base/standard/ConfigurationGenerational.cpp
+++ b/gc/base/standard/ConfigurationGenerational.cpp
@@ -63,6 +63,19 @@ MM_ConfigurationGenerational::newInstance(MM_EnvironmentBase *env)
 	return configuration;
 }
 
+void
+MM_ConfigurationGenerational::tearDown(MM_EnvironmentBase* env)
+{
+	MM_GCExtensionsBase* extensions = env->getExtensions();
+
+	/* Set pointer to scavenger in extensions to NULL
+	 * before Scavenger is teared down as part of clean up of defaultMemorySpace
+	 * in MM_Configuration::tearDown. */
+	extensions->scavenger = NULL;
+
+	MM_ConfigurationStandard::tearDown(env);
+}
+
 MM_MemorySubSpaceSemiSpace *
 MM_ConfigurationGenerational::createSemiSpace(MM_EnvironmentBase *envBase, MM_Heap *heap, MM_Scavenger *scavenger, MM_InitializationParameters *parameters, UDATA numaNode)
 {
--- a/gc/base/standard/EnvironmentStandard.cpp
+++ b/gc/base/standard/EnvironmentStandard.cpp
@@ -97,8 +97,19 @@ MM_EnvironmentStandard::flushGCCaches()
 {
 #if defined(OMR_GC_CONCURRENT_SCAVENGER)
 	if (getExtensions()->concurrentScavenger) {
-		if (MUTATOR_THREAD == getThreadType()) {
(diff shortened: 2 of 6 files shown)


ebadyano:
related to https://github.com/eclipse/openj9/pull/1783

gc/base/standard/ConfigurationGenerational.cpp
```diff
@@ -63,6 +63,19 @@ MM_ConfigurationGenerational::newInstance(MM_EnvironmentBase *env)
...
 
+void
+MM_ConfigurationGenerational::tearDown(MM_EnvironmentBase* env)
+{
+	MM_GCExtensionsBase* extensions = env->getExtensions();
+
+	/* Set pointer to scavenger in extensions to NULL
+	 * before Scavenger is teared down as part of clean up of defaultMemorySpace
+	 * in MM_Configuration::tearDown. */
+	extensions->scavenger = NULL;
```
charliegracie: If we set this to NULL here how do we find it to free during defaultMemorySpace tearDown?  Is it based on the default collector?

charliegracie: Ok thanks!

amicic:
If understand things right... If we attempt to put OMR change without OpenJ9 change, we will attempt to kill default memory space twice (in Java VM). 

First kill will be in OpenJ9 (old), and then second in OMR (new). OpenJ9 code does not set defaultMemorySpace pointer to NULL after the kill. So, OMR code cannot properly determine that it has already been killed.

To remove that dependency, a small pre-req change would have to go to OpenJ9 (to set NULL) fisrt, but since that code would go away anyway in the second/main OpenJ9 change, it's probably simpler for the two changes to go together.

amicic:
I just realized that OMR_GC_ShutdownHeap already kills default memory space, the same way this changes does it (but unsure at which moment).  It's doing it just prior to killing Configuration.

ATM, I'd argue, this change does it at the right place, since it is Configuration constructor that creates it, and it's the one that should kill it. The on in OMR_GC_ShutdownHeap should be removed.

gc/base/standard/EnvironmentStandard.cpp
```diff
@@ -97,8 +97,16 @@ MM_EnvironmentStandard::flushGCCaches()
 {
 #if defined(OMR_GC_CONCURRENT_SCAVENGER)
 	if (getExtensions()->concurrentScavenger) {
-		if (MUTATOR_THREAD == getThreadType()) {
-			getExtensions()->scavenger->threadFinalReleaseCopyCaches(this, this);
+
+		/* When a GC Slave thread is created, a Thread object is placed in its CopyCache.
```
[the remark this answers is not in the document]
ebadyano: thank you! I will update the comment.

charliegracie:
Agreed @amicic I believe the one in OMR_GC_ShutdownHeap should be removed as part of this change.

ebadyano:
removed now duplicated code in OMR_GC_ShutdownHeap that kills the default memory subspace. new personal build and grinder links: http://vmfarm.ottawa.ibm.com/build_info.php?build_id=385866 http://vmfarm.ottawa.ibm.com/build_info.php?build_id=385879

charliegracie:
@genie-omr build all

PR #2509, opened 2018-04-27
Remove deprecated SSE2IfArraycmpEvaluator on X86

0dvictor:
SSE2IfArraycmpEvaluator is not used at all and contains bugs,
removing it.


0xdaryl:

PR #2510, opened 2018-04-28
Print data snippet label name log in tracelog

0dvictor:
Data snippet label name is missing in tracelog on X86, adding it.


0xdaryl:

PR #2511, opened 2018-04-30
WIP : Compile omr using musl on Alpine platform

Param-S:
Changes to compile omr using musl toolchain on Alpine platform

Signed-off-by: Parameswaran Selvam <parselva@in.ibm.com>


compiler/ras/CallStack.cpp
```diff
@@ -171,8 +171,10 @@ const char *TR_PPCCallStackIterator::getProcedureName()
       }
    }
 
-#elif defined(LINUX)
+#elif defined(LINUX) /* && defined(__GLIBC__) */
```
0xdaryl: Can you remove the comment from this line?

include_core/omrsig.h
```diff
@@ -32,14 +32,15 @@
 extern "C" {
 #endif /* __cplusplus */
 
-#if defined(OSX)
+#if defined(OSX) || !defined(__GLIBC__)
 #define __THROW
 #endif /* defined(OSX) */
```
mstoodle: please update the #endif comment to link back to the new  #if condition

charliegracie:
@Param-S any update on this?  I would like to see this change get merged.

Param-S:
@charliegracie Sorry for the delay. The changes enable us to compile on alpine linux.  Currently, I made the function which uses backtrace, backtrace_symbols as dummy function(specific to alpine platform). It needs to be rewritten using libunwind to get the same functionality. I will make those changes in separate PR.

charliegracie:
@genie-omr build all

charliegracie:
@mstoodle can you review the Compiler changes again? Also what are your thoughts on the magic number?

port/unix/omrsysinfo.c
```diff
@@ -63,7 +63,12 @@
 #include <nl_types.h>
 #include <langinfo.h>
 #if !defined(USER_HZ) && !defined(OMRZTPF)
+#if defined(LINUX) && defined(__GLIBC__)
 #define USER_HZ HZ
+#else
+#include <sys/acct.h>
+#define USER_HZ AHZ
+#endif
```
mstoodle: missing comment for defined(LINUX) && defined(__GLIBC__)

compiler/ras/CallStack.cpp
```diff
@@ -229,6 +230,11 @@ void TR_LinuxCallStackIterator::printStackBacktrace(TR::Compilation *comp)
       }
    free(symbols);
    }
+#else
+void TR_LinuxCallStackIterator::printStackBacktrace(TR::Compilation *comp)
+	{
+	}
+#endif /* defined(__GLIBC__) */
```
mstoodle: I'm ok with this change; just might impact some debugging when stuff crashes, but backtrace doesn't seem to be something musl wants to deal with.

port/unix/omrsysinfo.c
```diff
@@ -128,7 +133,15 @@
...
 
 #if defined(LINUX) && !defined(OMRZTPF)
+#if defined(__GLIBC__)
 #include <linux/magic.h>
+#else
+/* Linux distros which doesn't use glibc toolchain like
+ * Alpine Linux the magic.h file is not available
+ * Random number for TMPFS_MAGIC variable is taken from other distros
+ */
+#define TMPFS_MAGIC 0x01021994
```
mstoodle: What breaks if this magic number isn't here?

I'm confused that musl doesn't define it and yet somehow us defining it is the right solution to the problem caused by TPFS_MAGIC not being there...

As it stands, I don't think I want code "taken from other distros" as the comment states.

ymanton: These sorts of magic numbers are fine; they're returned by the `statfs()` call and are well documented, e.g. in `man statfs`. The header in question is not part of any C library, it's part of the Linux kernel so I'm not sure why it wouldn't exist on a musl system. Is there some sort of headers package on musl that you can/should install? If so, it might be better to rewrite this as `#ifndef TMPFS_MAGIC ...` rather than looking for `__GLIBC__` but even guarding the include with `__GLIBC__` might be problematic because it doesn't guarantee its presence.

The code in question seems to call `statfs()` on the typical cgroups root mount point and I guess wants to confirm that it is a tmpfs based file system because that's what it usually is, but I don't know if that's really a requirement. I don't see why it couldn't be any other kind of FS, in which case this could would break.

Param-S: TMPFS_MAGIC is one of the file system type. It is clearly documented in man page of statfs. http://man7.org/linux/man-pages/man2/statfs.2.html.  I will correct the comment by clearly stating that "It is one of the file system type and it is documented in statfs system call man page"

Param-S:
@mstoodle I have updated the PR with explaining about the macro.

ymanton:
I think I have to reiterate my point from earlier that there are still issues here. There are includes here that are incorrectly guarded with `__GLIBC__` (`magic.h` as already mentioned, `prctl.h` is another that I've just spotted); these are not glibc headers, they are Linux kernel headers and there is likely a separate package that you need to install to get them.

charliegracie:
@Param-S can you look into the headers / packages that @ymanton is talking about?

charliegracie:
@Param-S ping?

charliegracie:
@Param-S is this something that you are still trying to get merged? After Sepetember 8th this will have been idle for 90 days and will be closed.

PR #2513, opened 2018-04-30
Streamline getmacros script to extract constants for DDR

keithc-ca:
* reduce number of processes created
* extract work per file so only a single redirection is needed
* use BASH_REMATCH to extract macro names
* run make with '--jobs=8'
* add timestamps to progress messages


charliegracie:
@genie-omr build all

PR #2514, opened 2018-05-01
Fix data type of GC_ExtensionsBase::usablePhysicalMemory

ashu-mehra:
GC_ExtensionsBase::usablePhysicalMemory should be of same type as the value
returned by omrsysinfo_get_addressable_physical_memory() which is uint64_t.


ashu-mehra:
fyi - @charliegracie

charliegracie:
@genie-omr build all

PR #2516, opened 2018-05-02
Correct .gitignore entry for build/ directory

mstoodle:
The entry right now says "build/" but this will match any directory
called "build" including anything below, say, jitbuilder/build . Given
the latter directory is revision controlled, it's not correct for it
to be ignored.

Fix is to use /build/ as the entry (leading slash makes it only apply
to the top level directory in the repo).


charliegracie:
I believe no further CI testing is required for this.

mstoodle:
Indeed, I had meant to include a skip ci tag but was apparently too excited to open the PR :( .

PR #2517, opened 2018-05-02
Add more detailed messages creating IDiaDataSource

keithc-ca:
Makes diagnosing problems due to missing or unregistered msdia*.dll libraries easier.


charliegracie:
@genie-omr build win

PR #2518, opened 2018-05-02
AppVeyor build now tests 32bit build

charliegracie:


charliegracie:
Now that we have Windows machines in the OMR CI testing 64 bit Windows builds I am suggesting we change the AppVeyor builds to 32 bit.  In the future if we have enough capacity at the OMR CI we can remove AppVeyor builds completely but for now they are quick and providing the OMR project with a way to test both 32 and 64 bit builds.

PR #2521, opened 2018-05-02
Include MemorySpace.hpp in gc/base/Configure.cpp

mstoodle:
A recent change requires Configure.cpp to know the full shape
of a MemorySpace object, but Configure.cpp only #include's
MemorySpace.hpp via some other header file. Worse, it may
not include MemorySpace.hpp under some configuration settings.

So let's just explicitly #include it.

Signed-off-buy: Mark Stoodley <mstoodle@ca.ibm.com>


charliegracie:
@genie-omr build all

PR #2524, opened 2018-05-03
Extend OMR Signal API [Part 2]

babsingh:
9 new commits have been added to support OMR Signal API extension/improvement.

1) **Add new OMR Signal API: omrsig_register_os_handler**

    omrsig_register_os_handler will register a handler with the OS. Stubs
    have been added for omrsig_register_os_handler.

    On Unix variants, we cache the original OS handler in oldActions while
    registering the first handler with the OS. The original OS handler is
    restored during shutdown.

    If OMRSIG_SIGACTION is called outside OMR, then the original OS handler
    will be overwritten without being cached in oldActions. Subsequently,
    the oldActions may cache a user handler instead of the original OS
    handler. During shutdown, we may not restore the original OS handler
    incorrectly.

    The new API, omrsig_register_os_handler, will let a user register a
    handler with the OS while properly caching the original OS handler. This
    will allow us to properly restore the original OS handler during
    shutdown.

2) **Re-format comments' layout**

3) **Rename signalsWithMasterHandlers and masterHandlerMonitor**

    signalsWithMasterHandlers is used to determine if we have registered a
    handler with a signal. With the introduction of
    omrsig_register_os_handler, a user can register handlers other than the
    master handlers. signalsWithMasterHandlers has been renamed to
    signalsWithHandlers to better represent its purpose.

    masterHandlerMonitor is acquired whenever a handler is registered with
    the OS. Since a user can register any arbitrary handler with the OS
    using omrsig_register_os_handler, masterHandlerMonitor is renamed to
    registerHandlerMonitor to better represent its purpose.

    After restoring the original OS handler in sig_full_shutdown,
    oldActions[index].restore should be reset to 0.

4) **Add implementation of omrsig_register_os_handler [Unix]**

    Implementation of omrsig_register_os_handler has been added in
    unix/omrsignal.c.

    Supporting entry and exit trace points have been added.

    registerSignalHandlerWithOS returns OMRPORT_SIG_ERROR for OS signals
    that are included in signalMap.

5) **Fix omrsig_set_async_signal_handler signature**

    Currently, omrsig_set_async_signal_handler returns an uint32_t.

    But, unix and ztpf implementations of omrsig_set_async_signal_handler
    set return value to OMRPORT_SIG_ERROR (-1).

    In order to allow negative return values,
    omrsig_set_async_signal_handler's signature is changed to return an
    int32_t.

6) **Use error macros instead of undocumented constants for errors**

    In omrsig_set_async_signal_handler and
    omrsig_set_single_async_signal_handler, constants such as 1 and -1 are
    directly used as error codes. Using constants directly leads to
    inconsistency in implementation across platforms.

    Now onwards, omrsig_set_async_signal_handler and
    omrsig_set_single_async_signal_handler will consistently return
    OMRPORT_SIG_ERROR (-1) for error.

7) **Cache info if a master handler is registered with a signal**

    omrsig_protect and omrsig_set_*_async_signal_handler may be invoked
    multiple times. If a master handler is already registered with a signal,
    then we don't want to re-register the master handler with that signal.

    If a master handler is registered with a signal, then this information
    is cached in signalsWithMasterHandlers. The port library signal bit is
    set in signalsWithMasterHandlers.

    When a user overrides the master handler by invoking
    omrsig_register_os_handler, then the port library signal bit is un-set
    in signalsWithMasterHandlers.

    omrsig_set_single_async_signal_handler returns the oldOSHandler. The
    oldOSHandler can be the master handler if the user re-invokes
    omrsig_set_single_async_signal_handler for a signal. So, a user can
    invoke omrsig_register_os_handler using a master handler. In this case,
    we must set signalsWithMasterHandlers bits accordingly in
    omrsig_register_os_handler.

    This allows us to prevent OS calls to re-register master signal
    handlers.

8) **Properly handle NULL values for oldOSHandler**

    omrsig_set_single_async_signal_handler and omrsig_register_os_handler
    have been updated to properly handle NULL values for oldOSHandler.
    Function description has been updated for both of them. A trace point
    was accessing oldOSHandler without a NULL check. A NULL check has been
    added before the trace point accesses oldOSHandler.

9) **Remove unnecessary write barrier**

    At the end of registerSignalHandlerWithOS, a write barrier is issued
    before updating signalsWithHandlers. According to the comment, the write
    barrier is needed since registerHandlerMonitor is not acquired before
    updating signalsWithHandlers at some places.

    It has been confirmed that registerHandlerMonitor is acquired every time
    signalsWithHandlers is accessed. So, the write barrier at the end of
    registerSignalHandlerWithOS is no longer required.

    In sig_full_shutdown, registerHandlerMonitor is not acquired before
    updating signalsWithHandlers. This is an exception since no other code
    should execute when OMR is shutting down.

### DIFF
--- a/port/common/omrsignal.c
+++ b/port/common/omrsignal.c
@@ -180,8 +180,11 @@ omrsig_set_async_signal_handler(struct OMRPortLibrary *portLibrary, omrsig_handl
  * entry corresponding to omrsig_handler_fn is removed, and related resources are freed. portlibSignalFlag can only
  * have one signal flag set; otherwise, OMRPORT_SIG_ERROR is returned. One omrsig_handler_fn handler is registered
  * with a signal at any time instead of multiple handlers. When associating a new omrsig_handler_fn with a signal,
- * prior omrsig_handler_fn(s) are dissociated from the signal. The address of the old signal handler function is
- * stored in oldOSHandler. This function supports all signals listed in OMRPORT_SIG_FLAG_SIGALLASYNC.
+ * prior omrsig_handler_fn(s) are dissociated from the signal. This function supports all signals listed in
+ * OMRPORT_SIG_FLAG_SIGALLASYNC.
+ *
+ * The address of the old signal handler function is stored in *oldOSHandler. In case of error or if NULL is provided
+ * for oldOSHandler, then *oldOSHandler is not updated to reflect the old signal handler function.
  *
  * @param[in] portLibrary The port library
  * @param[in] handler the function to call if an asynchronous signal arrives
@@ -229,7 +232,11 @@ omrsig_map_portlib_signal_to_os_signal(struct OMRPortLibrary *portLibrary, uint3
  * @brief Register a handler with the OS. For an invalid portlibSignalFlag, an error is returned.
  * portlibSignalFlag is invalid: if it is zero or if multiple signal bits are specified or if the
  * specified flag is not supported. If OS fails to register newOSHandler for the specified signal,
- * then an error is returned. In error cases, oldOSHandler won't be updated.
+ * then an error is returned.
+ *
+ * The address of the old signal handler function is stored in *oldOSHandler. In case of error or
+ * if NULL is provided for oldOSHandler, then *oldOSHandler is not updated to reflect the old
+ * signal handler function.
  *
  * This function may override a master handler which was previously set by omrsig_protect or
  * omrsig_set_*_async_signal_handler variant. The records associated with the master handler
--- a/port/common/omrsignal.c
+++ b/port/common/omrsignal.c
@@ -163,12 +163,12 @@ omrsig_protect(struct OMRPortLibrary *portLibrary,  omrsig_protected_fn fn, void
  * @note A handler should not do anything to cause the reporting thread to terminate (e.g. call omrthread_exit()) as that may prevent
  *  future signals from being reported.
  *
- * @return 0 on success
+ * @return 0 on success or non-zero on failure
  */
-uint32_t
+int32_t
 omrsig_set_async_signal_handler(struct OMRPortLibrary *portLibrary, omrsig_handler_fn handler, void *handler_arg, uint32_t flags)
 {
-	return 1;
+	return OMRPORT_SIG_ERROR;
 }
 
 
--- a/port/common/omrsignal.c
+++ b/port/common/omrsignal.c
@@ -226,6 +226,51 @@ omrsig_map_portlib_signal_to_os_signal(struct OMRPortLibrary *portLibrary, uint3
 }
 
 /**
+ * @brief Register a handler with the OS. For an invalid portlibSignalFlag, an error is returned.
+ * portlibSignalFlag is invalid: if it is zero or if multiple signal bits are specified or if the
+ * specified flag is not supported. If OS fails to register newOSHandler for the specified signal,
+ * then an error is returned. In error cases, oldOSHandler won't be updated.
+ *
+ * This function may override a master handler which was previously set by omrsig_protect or
+ * omrsig_set_*_async_signal_handler variant. The records associated with the master handler
+ * for the portlibSignalFlag signal are left unchanged when this function overrides the master
+ * handler. When the master handler is re-registered with the portlibSignalFlag signal, then
+ * the records associated with master handler don't need to be restored. An example of records
+ * is J9*AsyncHandlerRecord(s) in asyncHandlerList.
+ *
+ * Each platform variant of omrsignal.c should have a signalMap. signalMap should have a list of
+ * signals, which are supported on a platform. This function supports all signals listed in
+ * omrsignal.c::signalMap.
+ *
+ * On Unix variants, sigaction is used to register the handler. SA_SIGINFO flag is set.
+ * So, the signature of the signal handling function should be void (*)(int, siginfo_t *, void *).
+ * On zLinux, the signature changes to void (*)(int, siginfo_t *, void *, uintptr_t).
+ *
+ * sigaction isn't available on Windows. On Windows, signal is used to register the handler.
+ * The signature of the signal handling function should be void (*)(int).
+ *
+ * On Unix variants, original OS handler is cached in oldActions while registering the first handler
+ * with the OS. The original OS handler is restored during shutdown. If OMRSIG_SIGACTION is called outside
+ * OMR, then the original OS handler can get overwritten without being cached in oldActions. Subsequently,
+ * the oldActions can cache a user handler instead of the original OS handler. During shutdown, the
+ * original OS handler can be restored incorrectly. omrsig_register_os_handler lets a user register a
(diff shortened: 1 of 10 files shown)


charliegracie:
@genie-omr build all

babsingh:
@charliegracie @pshipton these changes can be reviewed.

port/unix/omrsignal.c
```diff
@@ -1193,11 +1259,8 @@ registerSignalHandlerWithOS(OMRPortLibrary *portLibrary, uint32_t portLibrarySig
 		}
 	}
 
-	/* CMVC 96193 signalsWithMasterHandlers is checked without acquiring the masterHandlerMonitor */
-	issueWriteBarrier();
```
pshipton: According to the comment, the problem is not that registerSignalHandlerWithOS is called without acquiring registerHandlerMonitor, the problem is that signalsWithHandlers is used elsewhere without first acquiring registerHandlerMonitor, and all the actions previous to the write barrier need to be finished before the bit set.

pshipton: CMVC 96193 contains the following reasoning. The idea was to avoid unnecessary acquires and improve performance, change sig_protect to improve perf. 

"We have seen j9thread_monitor_entry and exit for the masterHandlerMonitor mutex contributing 914
instructions twice in the newDirectByteBuffer. 

High-level logic is
acquire masterHandlerMonitor mutex
check for -Xrs flag
if not set then call registerMasterHandlers
release masterHandlerMonitor mutex

If we can test the -Xrs flag first and only acquire/release the mutex if we are going to call 
registerMasterHandlers then we can strip 1828 instructions from the newDirectByteBuffer pathlength.
"

babsingh: will undo this change.

babsingh: the following commit: `Remove unnecessary write barrier`, has been removed from the pull request.

port/common/omrsignal.c
```diff
@@ -225,6 +225,51 @@ omrsig_map_portlib_signal_to_os_signal(struct OMRPortLibrary *portLibrary, uint3
 	return OMRPORT_SIG_ERROR;
 }
 
+/**
+ * @brief Register a handler with the OS. For an invalid portlibSignalFlag, an error is returned.
+ * portlibSignalFlag is invalid if it is zero or if multiple signal bits are specified. If
```
pshipton: Its also invalid if it sets an unsupported signal

port/common/omrsignal.c
```diff
@@ -225,6 +225,51 @@ omrsig_map_portlib_signal_to_os_signal(struct OMRPortLibrary *portLibrary, uint3
...
+ * for the portlibSignalFlag signal are left unchanged when this function overrides the master
+ * handler. When the master handler is re-registered with the portlibSignalFlag signal, then
+ * the records associated with master handler don't need to be restored. An example of records
+ * will be J9*AsyncHandlerRecord(s) in asyncHandlerList.
+ *
+ * Each platform variant of omrsignal.c should have a signalMap. signalMap should have a list of
+ * signals, which are supported on a platform. This function will support all signals listed in
+ * omrsignal.c::signalMap.
+ *
+ * On Unix variants, sigaction will be used to register the handler. SA_SIGINFO flag is set.
```
pshipton: Please remove use of "will",  "we", "us".
i.e.
will be -> is
will let -> lets or allows
we cache the original OS handler -> cache the original OS handler
This will allow us to properly restore the original OS handler during shutdown
-> The original OS handler is properly restored during shutdown

port/unix/omrsignal.c
```diff
@@ -89,6 +89,15 @@ static struct {
  */
 static uint32_t signalsWithHandlers;
 
+/* Records the (port library defined) signals for which we have registered a master
```
pshipton: Avoid using "we"
for which we have registered a master handler -> for which a master handler is registered.
then we should avoid -> then avoid

port/unix/omrsignal.c
```diff
@@ -596,6 +605,24 @@ omrsig_register_os_handler(struct OMRPortLibrary *portLibrary, uint32_t portlibS
 	} else {
 		omrthread_monitor_enter(registerHandlerMonitor);
 		rc = registerSignalHandlerWithOS(portLibrary, portlibSignalFlag, (unix_sigaction)newOSHandler, oldOSHandler);
+		if (0 == rc) {
+			/*  A user-specified handler has been successfully registered for a signal. */
+			if ((newOSHandler == (void *)masterSynchSignalHandler)
+				|| (newOSHandler == (void *)masterASynchSignalHandler)
+			) {
+				/* User-specified handler is a master handler. So, we should set the portlibSignalFlag bit in
```
pshipton: avoid we

port/unix/omrsignal.c
```diff
@@ -596,6 +605,24 @@ omrsig_register_os_handler(struct OMRPortLibrary *portLibrary, uint32_t portlibS
...
+				 * signalsWithMasterHandlers.
+				 */
+				signalsWithMasterHandlers |= portlibSignalFlag;
+			} else {
+				/* If the user-specified handler is not a master handler, then we should un-set the
```
pshipton: avoid "we" and "will". There are more places, I'm not going to point them all out.

port/unix/omrsignal.c
pshipton: hmm, I don't think it needed to be removed. I was commenting because you seemed to misunderstand the reason for the write barrier, and providing the context for the barrier. Looking at the sig_protect code, it doesn't use signalsWithHandlers to check the -Xrs flag, but signalOptionsGlobal. I assume the code was changed at some point. Perhaps there is opportunity for future optimization of the monitors, but this should be left out of this change.

babsingh: `signalsWithHandlers` and `signalsWithMasterHandlers` are always accessed with `registerHandlerMonitor`. there is a comment about this next to their description: `Access to this must be protected by the registerHandlerMonitor`. should i re-add the commit for removing the write barrier?

pshipton: Re-adding it is fine with me.

babsingh:
@pshipton 
1. Re-added commit: `Remove unnecessary write barrier`
2. Removed usage of `we`, `will` and `us` from comments added in this pull request.

pshipton:
Please fix the commit comment for "Remove unnecessary write barrier" to explain properly why the barrier is no longer needed.

babsingh:
@pshipton I have updated the commit comment for `Remove unnecessary write barrier`.

babsingh:
@charliegracie Peter has reviewed these changes. can you please review/merge this PR?

charliegracie:
@genie-omr build all

ashu-mehra:
@babsingh I think the new trace points added in this have CRLF line endings, which would break f87abc3d5017a2000dd8d17276c829d282d79faa that changed files to have Unix line-endings.

babsingh:
Removed CRLF pairs in tracepoints: https://github.com/eclipse/omr/pull/2552

Issue #2525, opened 2018-05-04
AArch64: Configure/Make/CMake Build Support/Configuration


aarongraham9:
Configure, Make and CMake build support and configuration for AArch64.

aarongraham9:
A PR has been opened for this in #2526.

aarongraham9:
This can be closed via #2526.

Issue #2327, opened 2018-02-20
Implement AArch64 support


0xdaryl:
This is the umbrella epic to track the work needed to provide language-agnostic ARM AArch64 support to Eclipse OMR.  Tasks to be decomposed include:

- [ ] configure/CMake support
- [ ] port library
- [ ] compiler
- [ ] fvtest
- [ ] set up an open-source test environment for 32-bit and 64-bit ARM build and test

aarongraham9:
Configure/CMake support can be tracked in #2525.

PR #2520, opened 2018-05-02
AArch64: Initial Port Library Changes

aarongraham9:
This PR implements the initial configure/make/build systems, port library and Travis CI cross-compile build for aarch64 (ARM 64-Bit).

Work by @aarongraham9 and @jeanlego.

This is a part of #2327 and #2445.

Signed-off-by: Aaron Graham aaron.graham@unb.ca

Issue #39, opened 2016-03-28
Configure downloads required files


charliegracie:
Currently the configure script downloads the configure files config.sub and config.guess.  If the scripts can not download these files you can not compile the source.

I am suggesting that we check these files into the source to improve consumability of the Eclipse OMR project.

charliegracie:
To check-in source from another project we need to get Eclipse IP approval.

The CQ can be found here:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11036

Since we have a +1 from the PMC we can go forward with committing these files.  Since these files have a different license we will also need to update the OMR LICENSE file

PR #2526, opened 2018-05-04
AArch64: Initial Configure and Make/Build Systems

aarongraham9:
This PR implements the initial configure/make/build systems for aarch64.

There is still work to be done; as subsequent PR(s):
- Improve Build system
  * Implement Build System for aarch32

Work by @aarongraham9 and @jeanlego.

Part of: #2327 and #2525


fvtest/compilertest/build/Makefile
```diff
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2000, 2016 IBM Corp. and others
+# Copyright (c) 2000, 2018 IBM Corp. and others
```
JamesKingdon: Better not bump the date if we didn't make any other changes?

omrmakefiles/rules.linux.mk
```diff
@@ -155,7 +150,16 @@ endif
 ## Debugging Information
 # Indicate that GNU debug symbols are being used
 ifeq (gcc,$(OMR_TOOLCHAIN))
-  ifneq (,$(filter arm ppc s390 x86,$(OMR_HOST_ARCH)))
+  ifeq (arm,$(OMR_HOST_ARCH))
```
JamesKingdon: Did we lose s390 from this list?

config.guess
```diff
@@ -171,6 +171,7 @@ case "${UNAME_MACHINE}:${UNAME_SYSTEM}:${UNAME_RELEASE}:${UNAME_VERSION}" in
 	UNAME_MACHINE_ARCH=`(/sbin/$sysctl 2>/dev/null || \
 	    /usr/sbin/$sysctl 2>/dev/null || echo unknown)`
 	case "${UNAME_MACHINE_ARCH}" in
+		aarch64) machine=aarch64-unknown ;;
```
JamesKingdon: ws/alignment

example/glue/configure_includes/configure_linux_aarch64.mk
```diff
@@ -0,0 +1,61 @@
...
+	CONFIGURE_ARGS += \
+		--host=aarch64-linux-gnu \
+		--target=aarch64-linux-gnu \
+		--build=x86_64-pc-linux-gnu \
+		--enable-OMRTHREAD_LIB_UNIX \
+		--enable-OMR_ARCH_AARCH64 \
+		--enable-OMR_ENV_LITTLE_ENDIAN \
+		--enable-OMR_GC_TLH_PREFETCH_FTA \
+		--enable-OMR_PORT_CAN_RESERVE_SPECIFIC_ADDRESS \
+	    --enable-OMR_THR_FORK_SUPPORT \
```
JamesKingdon: ws/alignment

glue/configure_includes/configure_linux_aarch64.mk
```diff
@@ -0,0 +1,61 @@
...
+#
+# [1] https://www.gnu.org/software/classpath/license.html
+# [2] http://openjdk.java.net/legal/assembly-exception.html
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+###############################################################################
+
+include $(CONFIG_INCL_DIR)/configure_common.mk
+
+#	TODO:AARCH64 Disable fvtest ???
```
JamesKingdon: What are the pros and cons here? I guess the issue is if we are cross-compiling we won't be able to execute fvtests on the build platform. Is there merit in building the fvtests (but not executing them, obviously) so that the build directory could then be mounted/copied to an aarch64 platform to run the tests? I guess other option is to take fvtest out of the cross compile spec and add a new spec for native compile which includes it.
Either way we should probably make a decision and implement whatever that implies for this PR before merging :)

omrmakefiles/rules.linux.mk
```diff
@@ -58,19 +59,19 @@ ifeq (ppc,$(OMR_HOST_ARCH))
...
 ifeq (gcc,$(OMR_TOOLCHAIN))
-ifeq (1,$(OMR_RTTI))
-    GLOBAL_CXXFLAGS+=-fno-exceptions -fno-threadsafe-statics
-else
-    GLOBAL_CXXFLAGS+=-fno-exceptions -fno-rtti -fno-threadsafe-statics
-endif
-else
-ifeq (1,$(OMR_RTTI))
-    GLOBAL_CXXFLAGS+=-qrtti -qnoeh
+    ifeq (1,$(OMR_RTTI))
```
JamesKingdon: It's best to fight the urge to make formatting improvements that don't directly relate to the functional change that the PR introduces, because it makes life hell for your reviewers, who have to spend a lot of time trying to figure out if a particular diff is functional or not. Prettifying unrelated code is best done in it's own PR so that reviewers know what to expect.

run_configure.mk
```diff
@@ -83,7 +86,7 @@ endif
 ifneq (,$(findstring win_x86,$(SPEC)))
   include $(CONFIG_INCL_DIR)/configure_win_x86.mk
   # OMRTODO: this is J9 buildfarm specific code.
-  export PATH:=$(DEV_TOOLS)/jtc-toolchain/java7/windows/mingw-msys/msys/1.0/bin:$(PATH)
+	export PATH:=$(DEV_TOOLS)/jtc-toolchain/java7/windows/mingw-msys/msys/1.0/bin:$(PATH)
```
JamesKingdon: ws/alignment

configure
```diff
@@ -786,6 +787,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
```
JamesKingdon: What's the purpose/need for the new directory?

glue/configure_includes/configure_linux_aarch64.mk
aarongraham9: I'd say we should build them so that they could be migrated to an AArch64 platform to be natively executed. Thoughts?

omrmakefiles/rules.linux.mk
aarongraham9: Yeah fair enough; I believe the IDE that @jeanlego and I are using does some auto whitespace stripping/formatting.

omrmakefiles/rules.linux.mk
aarongraham9: It would seem so. I'll be sure to put it back.

omrmakefiles/rules.linux.mk
jeanlego: No this was me. I've rearranged things so that's its clearer where to put new architecture in. (Change if/else ladders to if else elif format)

fvtest/compilertest/build/Makefile
aarongraham9: Agreed I'll fix that.

configure
aarongraham9: That is a good question. We re-ran `autoconf`and I think it generated that into the configure script. I can remove that.

glue/configure_includes/configure_linux_aarch64.mk
JamesKingdon: That sounds very reasonable to me.

jeanlego: My thoughts too. That was the idea behind this. At least if we use this for jenkins

config.guess
aarongraham9: Fixed.

example/glue/configure_includes/configure_linux_aarch64.mk
aarongraham9: Fixed.

run_configure.mk
aarongraham9: Fixed.

omrmakefiles/rules.linux.mk
aarongraham9: Should be cleaned up.

aarongraham9:
All comments should be addressed in the latest push/commit. This should be ready for re-review @JamesKingdon, @0xdaryl and @jeanlego.

jeanlego:
This should be ready for review again ! @JamesKingdon, @0xdaryl.

JamesKingdon:
Many thanks for all the clean-up, this looks great. @0xdaryl we should probably get someone who knows more about the build than me (especially the new cmake stuff) to take a look - any suggestions?

omrmakefiles/rules.linux.mk
```diff
@@ -155,7 +157,7 @@ endif
 ## Debugging Information
 # Indicate that GNU debug symbols are being used
 ifeq (gcc,$(OMR_TOOLCHAIN))
-  ifneq (,$(filter arm ppc s390 x86,$(OMR_HOST_ARCH)))
+  ifneq (,$(filter aarch64, arm ppc s390 x86,$(OMR_HOST_ARCH)))
```
0xdaryl: Any reason the `aarch64` gets a comma?

0xdaryl:
I'd appreciate a review from @rwy0717

omrmakefiles/rules.linux.mk
aarongraham9: No that slipped through. Fixed.

rwy0717: I think that's a typo which breaks the test.

glue/configure_includes/configure_linux_aarch64.mk
```diff
@@ -0,0 +1,59 @@
...
+	CONFIGURE_ARGS += \
+	--host=aarch64-linux-gnu \
+	--target=aarch64-linux-gnu \
+	--build=x86_64-pc-linux-gnu \
+	--enable-OMRTHREAD_LIB_UNIX \
+	--enable-OMR_ARCH_AARCH64 \
+	--enable-OMR_ENV_LITTLE_ENDIAN \
+	--enable-OMR_GC_TLH_PREFETCH_FTA \
+	--enable-OMR_PORT_CAN_RESERVE_SPECIFIC_ADDRESS \
+    --enable-OMR_THR_FORK_SUPPORT \
```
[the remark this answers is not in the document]
aarongraham9: Fixed.

omrmakefiles/rules.linux.mk
aarongraham9: Agreed. Fixed in the latest push.

rwy0717:
It looks like you updated only part of `config.guess` to support aarch64 detection. Maybe we should just update it to the latest from [savannah](https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD)?

fvtest/compilertest/Makefile
```diff
@@ -51,6 +51,13 @@ ifeq ($(PLATFORM),arm-linux-gcc-cross)
     SUBMAKE_EXTRAARGS=PLATFORM=arm-linux-gcc
 endif
 
+ifeq ($(PLATFORM),aarch64)
+    export CC_PATH?=aarch64-unknown-linux-gnu-gcc
+    export CXX_PATH?=aarch64-unknown-linux-gnu-g++
+    export AS_PATH?=aarch64-unknown-linux-gnue-as
```
rwy0717: I'm not sure, but is this a typo: `gnue`?

tools/configure
```diff
@@ -2263,6 +2263,8 @@ $as_echo_n "checking OMR_BUILD_ARCH... " >&6; }
 case $build_cpu in #(
   *arm*) :
     OMR_BUILD_ARCH=arm ;; #(
+  aarch64) :
+    OMR_BUILD_ARCH=aarch64 ;; #(
```
rwy0717: This is a generated file, and I don't think you updated the source `tools/configure.ac`.

rwy0717: Nevermind, this looks good! Jumped the gun a little :)

rwy0717:
I think there is some cmake platform detection in the wrong PR:
[cmake/modules/OmrDetectSystemInformation.cmake](https://github.com/eclipse/omr/pull/2520/files#diff-513069afb8ad1b8d4c252c6dafdc189f)

Also, you set these two compile options in autotools, but not cmake:
* `-march=armv8-a+simd`
* `-Wno-unused-but-set-variable`

Are we okay without setting them?

jeanlego:
they definitely should be set. This was my doing, unless cmake consumes this? (My cmake knowledge is very much lacking) I used everything arm build related that is in there as a template. So it could be missing from arm also.

rwy0717:
`config.guess` is gnu code that we have checked in to our build system. The original issue is #39. Our license has special notes about these files. Maybe @charliegracie knows more about what process needs to happen if we want to update these files. I suspect we can just check in the newer versions.

aarongraham9:
@rwy0717 I've added(moved) in the `cmake/modules/OmrDetectSystemInformation.cmake` and `util/omrutil/gettimebase.c` files in from the AArch64 Port PR. They were split over there inadvertently. They more belong here for sure.

I also added in the missing compile flags to the `cmake/modules/platform/arch/aarch64.cmake` file.

aarongraham9:
@rwy0717, @0xdaryl, @JamesKingdon and @charliegracie; aside from the on-going `config.guess` update discussion this should be good for re-review and/or merge.

Would it be better to move the `config.guess` update discussion to an issue? Perhaps that way we could merge this PR and revisit the code change when the discussion is resolved?

charliegracie:
to update config.guess we need to open an Eclipse CQ.  @0xdaryl can you look into that?

0xdaryl:
Thoughts on updating `config.sub` while we're at it?  @charliegracie @rwy0717 

I can add it to the CQ for permission, but we don't necessarily have to update it.

0xdaryl:
CQ submitted:  https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16278

0xdaryl:
@genie-omr build all

0xdaryl:
The CQ to update `config.sub` and `config.guess` has been approved.  Rather than bury their upgrade in this PR, I would rather create a separate PR to track their upgrade because they do have side effects beyond AArch64 (I'll create an issue).

Can this PR be merged without `config.guess` ?  If yes, then @aarongraham9 please remove these changes from this PR.

Are there any other technical issues anyone sees that should prevent this PR from being merged?  @charliegracie @rwy0717 @JamesKingdon

jeanlego:
It's essential for aarch64 to build.

aarongraham9:
Agreed @jeanlego. What is in this PR is required for AArch64 to build in it's current state.

aarongraham9:
But there shouldn't be any issue with overwriting it later with the Savannah version(s) down the road.

0xdaryl:
My only concern is that the CQ(s) were approved for unmodified files from savannah.  If nothing will break by delivering this PR without an updated `config.guess` (and nothing is immediately depending on this PR), and then we merge the new one in a subsequent PR, I propose we go that route.

aarongraham9:
Of fair enough. Let me re-verify if not having those changes breaks anything for this code submission.

jeanlego:
the updated travis needs this to be merged in. If you guys don't see any hurry in adding it in then it makes sense to me too!

0xdaryl:
Agreed.  But adding a Travis build is in another PR that will have to depend on  the`config.guess` PR being updated first.

aarongraham9:
Just confirmed that it won't break anything in this PR. Will push an updated version shortly.

aarongraham9:
Yes the Travis PR will have to wait on the `config.guess` PR but I don't think that will be a big deal.

0xdaryl:
Could you create a PR for the updated config files?  It will need to be reviewed and tested by others.  I already have an issue here #2538.

aarongraham9:
Will do @0xdaryl.

aarongraham9:
Revised code without and changes to `config.guess` has been pushed. Pending Appveroy/Travis(/Jenkins) builds now.

0xdaryl:
@genie-omr build all

For completeness.

aarongraham9:
There now seems to be a, non-error, error in the Travis CMake build. I don't believe the `config.guess` reversion should have caused this? Any thoughts @rwy0717?

```
[ 28%] Building CXX object jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/codegen/OMRRegister.cpp.o

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

aarongraham9:
I'm thinking it's just Travis/CMake running a bit slow and/or our timeout is a bit too low?

aarongraham9:
Is it possible to trigger another/re-run the Travis build manually without re-pushing the code again @0xdaryl?

0xdaryl:
I have waved my magic wand...

aarongraham9:
I was wrong @0xdaryl. It seems to have stalled again. I'm quite sure it's a Travis issue because it hung way before it even got to compiling code. I think even before pulling our code.

Issue #2426, opened 2018-04-05
Tenure heap expansion & Concurrent Card Cleaning Race Condition


RSalman:
When determining card table ranges to be cleaned, ranges belonging to uncommitted card table memory are found to be valid ranges. This issue occurs when card cleaning takes place at the same time as tenure heap expansion. Specifically when we have 
> Tenure heap expansion in a middle of Scavenging (failed attempt to Tenure, which isn't fully expanded yet), and also that the problem occurs only with Concurrent Scavenger cycle overlapping with Concurrent Marking cycle. This is only (known) way that Tenure heap expansion can race with Concurrent Card Cleaning.

This leaves the card table cleaning operation with an inconsistent view of the subspace structures as all structures are assumed to be expanded for regions found in the region manager. However, with the current implementation, the region manager is updated before the expansion of all the structures:

`omr/gc/base/PhysicalSubArenaVirtualMemoryFalt.cpp`
```C++
/* Reconstruct the region at its new size */
getHeapRegionManager()->resizeAuxillaryRegion(env, _region, _lowAddress, _highAddress);
Assert_MM_true(NULL != _region);

/* Update the owning subspace */
_subSpace->expanded(env, this, expandSize, lowExpandAddress, highExpandAddress, true);
_subSpace->heapReconfigured(env);
```

For this reason, the region addresses are available to clean, even though corresponding card table addresses may not be committed:

`omr/gc/base/standard/ConcurrentCardTable.cpp`

````C++
MM_HeapRegionManager *regionManager = heap->getHeapRegionManager();
GC_HeapRegionIterator regionIterator(regionManager);
while(NULL != (region = regionIterator.nextRegion())) {
	/* Get reference to owning subspace */
	MM_MemorySubSpace *subspace = region->getSubSpace();

	/* Do we need to include this segments cards  */
	if (subspace->isActive() && (_cleanAllCards || subspace->isConcurrentCollectable())) {
		numRanges += 1;

		/* If there is space in cleaningRanges array so add it... */
		if (numRanges <= _maxCleaningRanges) {
			nextRange->baseCard = heapAddrToCardAddr(env, region->getLowAddress());
			nextRange->topCard = heapAddrToCardAddr(env, region->getHighAddress());
			nextRange->nextCard = nextRange->baseCard;
			nextRange->numCards = (uintptr_t)(nextRange->topCard - nextRange->baseCard);
			_cardTableStats.totalCards += nextRange->numCards;
			nextRange += 1;
		}
[...]

PR #2527, opened 2018-05-04
Fix _initRanges race with CM and CS

amicic:
_initRanges is a table that drives parallel setting/reseting of Mark Map
in Tenure/Nursery at the begining of Concurrent Marking CM)
cycle/begining STW phase.

In presence of Concurrent Scavenger (CS), that can over lap with CM, and
if Tenure heap expansion occurs (due to failed tenuring), the table will
be updated. 

This change will make the heap expanding thread simply skip the table
update if it senses that init in is progress. The table update is left
for a later, safer point .

BTW, if initializaiton is already in progress, the thread doing heap
expansion will do Mark Map work for the expaned part of the heap (this
part, and deferral of the table update is actually already handled for
standard Scavenger), so it does not really matter if update of
_initRanges occurs just after Mark Map init
work is complete.

This fix has not direct relation with issue reported by
#2426. A few conditions are same
(CM and CS overlapping, CS doing Tenure expansion), but the key
difference is what CM is doing: in this scenario Mark Map initialization
vs Card Table cleaning as in the issue.

Some other minor changes:
- split the deferal flag (_rebuildInitWork) into two flags
(expand/contract) for easier potential future debugging (there is a
nasty case of heap contraction that might be compromised)
- renamed a flag to indicates STW phase of Concurernt Mark

### DIFF
--- a/gc/base/standard/ConcurrentGC.cpp
+++ b/gc/base/standard/ConcurrentGC.cpp
@@ -819,7 +819,6 @@ void
 MM_ConcurrentGC::determineInitWork(MM_EnvironmentBase *env)
 {
 	bool initDone= false;
-	uintptr_t initWork;
 
 	Trc_MM_ConcurrentGC_determineInitWork_Entry(env->getLanguageVMThread());
 	
@@ -901,7 +900,7 @@ MM_ConcurrentGC::determineInitWork(MM_EnvironmentBase *env)
 	}
 
 	/* Now count total initailization work we have to do */
-	initWork = 0;
+	uintptr_t initWork = 0;
 	for (uint32_t i = 0; i < _numInitRanges; i++) {
 		if (_initRanges[i].base != NULL) {
 			initWork += _initRanges[i].initBytes;
@@ -909,7 +908,8 @@ MM_ConcurrentGC::determineInitWork(MM_EnvironmentBase *env)
 	}
 
 	_stats.setInitWorkRequired(initWork);
-	_rebuildInitWork = false;
+	_rebuildInitWorkForAdd = false;
+	_rebuildInitWorkForRemove = false;
 
 	Trc_MM_ConcurrentGC_determineInitWork_Exit(env->getLanguageVMThread());
 }
@@ -1339,7 +1339,7 @@ MM_ConcurrentGC::tuneToHeap(MM_EnvironmentBase *env)
 	 */
 	if(0 == heapSize) {
 		Trc_MM_ConcurrentGC_tuneToHeap_Exit1(env->getLanguageVMThread());
-		assume0(!_globalCollectionInProgress);
+		assume0(!_stwCollectionInProgress);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


charliegracie:
@genie-omr build all

gc/base/standard/ConcurrentGC.cpp
```diff
@@ -1427,7 +1426,59 @@ MM_ConcurrentGC::tuneToHeap(MM_EnvironmentBase *env)
...
+				omrthread_monitor_exit(_initWorkMonitor);
+
+				omrthread_monitor_wait(_initWorkCompleteMonitor);
+
+				/* Ongoing initialization just (likely) completed. Re-acquire _initWorkMonitor to prevent any other thread to join (in case this was just a temporary condition
+				 *  where work was not complete), because we are about to examine if we can complete our work. We will not try to distinguish the case between cases initialization
+				 *  being still in progress, or being completed. Either way, we will do our work. It would not be correct to wait them to finish, because they might have given up
+				 *  VM access and might not come back. Reconfiguring table that has been already used to do some init work, will cause the init work to be redone,
+				 *  but that's a reasonable price to pay for such a rare scenario. */
+				omrthread_monitor_enter(_initWorkMonitor);
```
charliegracie: I believe this could cause a deadlock.  There are locations including the new code above that have the following pattern with monitors:
```
enter(_initWorkMonitor)
enter(_initWorkCompleteMonitor)
```

This new code will be holding `_initWorkCompleteMonitor` and then trying to enter `_initWorkMonitor` which will make the enter order:
```
enter(_initWorkCompleteMonitor)
enter(_initWorkMonitor)
```

If threads are entering the monitors in different orders I believe we could deadlock

amicic: Valid concern - but things are not so simple. Let's see....

This order 
enter(_initWorkCompleteMonitor)
enter(_initWorkMonitor)
occurs only at one point, in the new code, when the heap expanding thread, waiting for Mark Map initializer threads to complete the work, is notified that there are no more active initializers (0 ==_initializers). When this thread tries to acquire _initWorkMonitor again (already holding _initWorkCompleteMonitor), no other thread should really attempt to acquire _initWorkCompleteMonitor (while already holding _initWorkMonitor).

There are 3 spots that the later could potentially occur:
1) another thread trying to expand heap (entering this new code sequence). It is not possible, because it's prevented (even in standard Scavenger) by Subspace level lock that serialize such threads.

2) end of doConcurrentInitialization when new initalizers threads try to leave and potentially notify any other blocking thread (either initalizers ones or heap expanding ones). It should not be possible because there no active initializers threads (0 ==_initializers) nor new one should be able to join (see 3)

3) beginning of doConcurrentInitialization, when new initalizers threads try to join.
A new thread trying to join will quickly realize that it should not join, for the same reasons one of initalizers threads recently notified the heap expanding threads there are no more active initializers: init work complete or exclusive access has been requested. If any of two checks are true, the thread will release _initWorkMonitor and leave without attempting to acquire _initWorkCompleteMonitor. 
Things are, however, even more complicated with _determineInitWorkPending flag, because it is one of the conditions we notify heap expanding flag, but it is not the reason to prevent new initializes to join (because we want them to help and finish initialization as fast as possible). I'll think about this more, if could cause a deadlock or not... (I'll also think about possibly simplify things by making heap expanding thread not blocking at all, but just postpone the init table update for a later, safer point).

amicic:
Should not go before https://github.com/eclipse/omr/pull/2564 , since this change has some asserts that uncover existing problems in shutdown sequence that https://github.com/eclipse/omr/pull/2564 (and some dependent changes) fixes

PR #2529, opened 2018-05-04
Fix line-ends

keithc-ca:
With this change, all files have only Linux line-ends (no CRLF pairs). This should improve the situation of building OMR (and downstream projects like OpenJ9) on z/OS where more care is required so proper conversions to EBCDIC can occur.

See also eclipse/openj9#1743.


keithc-ca:
https://github.com/eclipse/omr/pull/2529/files?w=1 shows that, besides copyright dates, this includes only whitespace changes.

rwy0717:
Maybe not in this PR, but do we think we should check in a gitattributes file and force people to use the right line endings?

keithc-ca:
I would be in favour of having `.gitattributes` included in the repository, but I've never been able to make it work satisfactorily. Another option is to have a PR check for proper line-ends like there is in the eclipse/openj9 repository.

charliegracie:
@genie-omr build zos

Only whitespace and copyright date modifications.  Only testing zOS because it can fail due to column length, etc.

mstoodle:
Alas, burned by the whitespace police :( . Good change, though.

Issue #2497, opened 2018-04-23
Discuss: Split omrcore into startup, RAS, and base libraries


rwy0717:
I think the responsibilities of omr core are too far reaching, and results in a web of interdependencies between components. My suggestion is to split omr core into three libraries:

1. `omrbase`: The core library would contain the basic data structures all OMR components rely on, like OMR_VM and OMR_VMThread. `-I include_core/` would be provided by omrbase. omrbase has no (or minimal) dependencies. If we want to call this library `omrcore`, people should be aware that autotools tries to delete files named `core`.

2. `omrstartup`: The startup library would provide the monolithic startup/shutdown helpers. This library depends on all other OMR components. Breaking out startup into a separate library removes core's circular dependency on other components. Further, this means that startup can be an optional library for OMR consumers. If startup also initializes the JIT, than the GC will inherit omrcore's dependency on the JIT.

3. `omrras`: Lastly, it makes sense that RAS should be it's own component, because RAS actually has it's own configure option and glue code, and I think is important enough for it's own directory.

see also:
- #2481
- #1644

/cc @dnakamura @samolisov @charliegracie

samolisov:
Colleagues, what does RAS means?

@rwy0717 I agree, omrcore has many dependencies (OMR_TRACE_LIB, OMR_HOOK_LIB, even GC when OMR_GC is enabled, JIT in the future, etc.) for now and OMR looks like a monolithic application. I think we can go on the LLVM way: build OMR as a set of (mostly) independent libraries (as it is written in the short description of the project) to give developers an opportunity to pick only those that required.

rwy0717:
Just checking, the core tracing engine stuff belongs in omrbase, it's the TI and RAS code that moves to omrras?

charliegracie:
That is the hard distinction.  Technically tracing is a RAS feature but ALL components pretty much use the trace engine so if you put it in omrbase all components will require omrbase.  If you put it in RAS all components will require RAS.  I am not sure which is better.  I kind of feel like RAS is a component like util which could be required by all components.

rwy0717:
Right. Will keep it in RAS. After looking a little more at the API component, what do you think of calling the library `omrvm`?

rwy0717:
[...]

PR #2530, opened 2018-05-07
CMake: Fix linkage issues in omrcore

dnakamura:

### DIFF
--- a/omr/CMakeLists.txt
+++ b/omr/CMakeLists.txt
@@ -63,17 +63,19 @@ target_include_directories(omrcore
 		.
 )
 
+
+target_link_libraries(omrcore
+	# NOTE: Don't link any other libraries to omrcore
+	# Doing so will almost certainly create a circular dependency in OpenJ9
+	PRIVATE
+		${OMR_CORE_GLUE_TARGET}
+)
+
+
 if(OMR_GC)
 	add_dependencies(omrcore
 		omrgc_tracegen
 		omrgc_hookgen
-	)
-
-	target_link_libraries(omrcore
-		PUBLIC
-			omrgc
-		PRIVATE
-			${OMR_CORE_GLUE_TARGET}
 	)
 
 	target_include_directories(omrcore


rwy0717:
LGTM

omr/CMakeLists.txt
```diff
@@ -63,19 +63,21 @@ target_include_directories(omrcore
 		.
 )
 
+
+target_link_libraries(omrcore
+	# NOTE: Don't link any other libraries to omrcore
+	# Doing so will almost certainly create a circular dependency in OpenJ9
```
DanHeidinga: `in OpenJ9` -> `in dependent projects`?  Or maybe `in consuming projects`?

samolisov:
@dnakamura A comment from #2523. I don't know why linking omrcore with their dependencies create any problems for the downstream project. Could the reason be in hardcoded `gc` in `omr/CMakeLists.txt`? If so, one of the PR's commits changes hardcoded `gc` to `${OMR_GC_LIB}` and customers will be able to put their own gc library there.

What do you think about `${OMR_HOOK_LIB}` and `${OMR_TRACE_LIB}`? Must omrcore not be linked with those libraries too?

charliegracie:
@dnakamura can you provide more details about this change?  If there are circular dependencies or anything like that we need to resolve those.

dnakamura:
@samolisov  The issue i believe technically exists in omr, however cmake will not complain about circular dependencies if all the nodes are static libraries. However for OpenJ9 `${OMR_GC_LIB}` , `${OMR_HOOK_LIB}`, and `${OMR_TRACE_LIB}` are all shared libraries.  The actual error message I'm getting:
```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "omrcore" of type STATIC_LIBRARY
    depends on "omrgc" (weak)
    depends on "j9trc" (weak)
  "omrgc" of type STATIC_LIBRARY
    depends on "omrcore" (weak)
    depends on "j9trc" (weak)
  "j9trc" of type SHARED_LIBRARY
    depends on "omrcore" (weak)
    depends on "omrgc" (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```

rwy0717:
As I understand what @dnakamura just said:

J9's tracepoint shared library needs RAS stuff from `omrcore`, but the j9 GC glue needs tracepoints. So you end up with this circular dependency graph (this is just one possible interpretation):

```
j9trc -> core -> gc+glue -> j9trc
```

The circular dependency is ultimately between omrcore and omrgc. CMake will allows circular dependencies between static libraries, but complains once you introduce shared libraries into the mix, as in j9trc. The long term solution will be to break the core library into multiple components #2497.

dnakamura:
So after talking offline with @rwy0717 , we believe to root of the issue is that in the cmake world we are building `omrvmstartup` (ie the  `omr/startup` directory) as part of the `omrcore` library, instead of as a separate library. By doing this, in addition to moving `omrrasinit.c` and `omrtrcinit.c` into the startup directory, we can eliminate the bulk of, if no all, the circular dependency issues here. I am working on putting a PR together now

dnakamura:
Closed since #2533 is the better approach

PR #2532, opened 2018-05-08
Scavenger cleanup and performance improvements

charliegracie:
This change cleans up some of the code in the hot loops and removes some unnecessary tracing.

The aliasToCopyCache code was refactored to provide the same or very similar aliasing with much less cost.

Also the calculateOptimumCopyCacheSize was refactored to be based on waiting threads and work list counts.  The current algorithm limited maximum copy cache size to about 50% or less of the true maximum for almost every cache. The new algorithm could still be improved by copy cache sizes now reach maximum for large parts of the collection.


charliegracie:
@amicic and @dmitripivkine can you review this when you get a chance.

gc/base/standard/Scavenger.hpp
```diff
@@ -285,22 +287,29 @@ class MM_Scavenger : public MM_Collector
...
-
-	MMINLINE bool isWorkAvailableInCache(MM_CopyScanCacheStandard *scanCache);
 	MMINLINE void addCacheEntryToScanListAndNotify(MM_EnvironmentStandard *env, MM_CopyScanCacheStandard *newCacheEntry);
 
+	MMINLINE bool
+	isWorkAvailableInCache(MM_CopyScanCacheStandard *cache)
+	{
+		return ((NULL != cache) && (cache->scanCurrent < cache->cacheAlloc));
+	}
+
```
amicic: do you need the null check here too?

charliegracie:
Some GC times showing the improvements to Scavenge time as reported in verbose:gc.  All runs were completed with `-Xmx2G -Xms2G -Xmn256m -Xgc:scvTenureAge=1,scvNoAdaptiveTenure` command line options.  Runs were interleaved performing a default run followed by a modified run.
Without any changes (total time spent in GC is 4.3%):

|Run #|Mean time (ms)|Minimum time (ms)|Maximum time (ms)|Total time (ms)|
|-----:|:--------------:|:------------------:|:-------------------:|:--------------:|
|1        |11.8                   |2.78                          |49.3                          |9562                 |
|2       |11.9                   |2.58                          |41.2                           |9716                  |
|3       |11.9                   |2.96                          |41.7                           |9852                 |
|4       |11.8                   |3.88                          |46.9                          |9582                 |
|5       |11.6                   |2.44                          |35.0                          |9272                  |

With the changes (total time spent in GC is 4.0%):

|Run #|Mean time (ms)|Minimum time (ms)|Maximum time (ms)|Total time (ms)|
|-----:|:--------------:|:------------------:|:-------------------:|:--------------:|
|1        |10.9                  |2.8                             |44.1                           |9008                |
|2       |10.7                   |2.68                          |35.0                           |8725                |
|3       |10.7                   |3.29                          |40.5                           |8807                |
|4       |10.6                   |2.48                          |46.9                           |8834                |
|5       |10.9                   |2.5                            |36.2                            |8991                |

gc/base/standard/Scavenger.hpp
charliegracie: No good catch.  I missed removing it when I refactored the changes into separate commits for the PR.

charliegracie: Fixed in the latest commit

charliegracie:
There are further improvements that could also be made in the tight loop for hierarchical scanning. I will continue to look at those next week. Also during this investigation I noticed that the split list count for the packetlist and scanlist defaults to the number of GC threads.  I believe this is overkill and I will investigate modifying it to closer match the split allocation list count in MemoryPoolSplitAddressOrderedList.

gc/base/standard/Scavenger.cpp
```diff
@@ -3159,61 +3104,53 @@ MM_Scavenger::aliasToCopyCache(MM_EnvironmentStandard *env, GC_SlotObject *scann
...
+#if defined(J9MODRON_TGC_PARALLEL_STATISTICS)
+		env->_scavengerStats._releaseScanListCount += 1;
+#endif /* J9MODRON_TGC_PARALLEL_STATISTICS */
+		addCacheEntryToScanListAndNotify(env, env->_deferredScanCache);
+		env->_deferredScanCache = NULL;
 	}
 	return NULL;
-		}
-
+}
```
dmitripivkine: missed empty line

gc/base/standard/Scavenger.cpp
```diff
@@ -877,32 +829,61 @@ MM_Scavenger::calcGCStats(MM_EnvironmentStandard *env)
...
 
-	/* Fit result into allowable cache size range */
-	if (cacheSize < _extensions->scavengerScanCacheMinimumSize) {
-		cacheSize = _extensions->scavengerScanCacheMinimumSize;
-	} else if (cacheSize > _extensions->scavengerScanCacheMaximumSize) {
-		cacheSize = _extensions->scavengerScanCacheMaximumSize;
+	uintptr_t scanCacheCount = _scavengeCacheScanList.getApproximateEntryCount();
+	if (scanCacheCount < threadCount) {
+		uintptr_t cacheSizeBasedOnScanCacheCount = calculateCopyScanCacheSizeForQueueLength(maxCacheSize, threadCount, scanCacheCount);
+		cacheSize = OMR_MIN(cacheSizeBasedOnScanCacheCount, cacheSize);
```
amicic: I would prefer changes in calculateOptimumCopyScanCacheSize go separately (not just PRs, but actual nightly builds) so we can distinguish between the aliasing change that also might change the behavior.

amicic:
looks good

dmitripivkine:
looks reasonable for me

gc/base/standard/Scavenger.cpp
```diff
@@ -440,7 +438,6 @@ MM_Scavenger::workerSetupForGC(MM_EnvironmentStandard *env)
 	/* caches should all be reset */
 	Assert_MM_true(NULL == env->_survivorCopyScanCache);
 	Assert_MM_true(NULL == env->_tenureCopyScanCache);
-	Assert_MM_true(NULL == env->_scanCache);
```
dmitripivkine: Is it any way to have a similar check with new logic?

charliegracie: Not with the current changes.  We still verify that all of the copy caches are NULL.  Since there is technically not scanCache pointer in the env there is nothing to check.  I believe we still have an assert at the end of the GC to ensure that all caches are back on the empty list so I think we are covered.

charliegracie:
I fixed the spacing @dmitripivkine.  @amicic I removed the calculateOptimumCopyScanCacheSize from this PR.  I will open another PR with that change once this one is delivered.

charliegracie:
@genie-omr build all

charliegracie:
@amicic, @dmitripivkine and @youngar I believe this is ready to go

PR #2533, opened 2018-05-08
CMake: Split omrvmstartup into separate library

dnakamura:
See #2530 for more discussion


fvtest/vmtest/CMakeLists.txt
```diff
@@ -30,16 +30,16 @@ target_link_libraries(omrvmtest
...
-	omrport
 	j9hashtable
 	j9avl
 	j9pool
-
+	omrvmstartup
+	${OMR_HOOK_LIB}
+	${OMR_GC_LIB}
+	${OMR_PORT_LIB}
+	${OMR_TRHREAD_LIB}
```
rwy0717: typo

dnakamura: well i guess they dont need the thread lib then 😄

omr/startup/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+	trc_omrti
+)
+
+
+target_link_libraries(omrvmstartup
+	PRIVATE
+		omrcore
+		${OMR_HOOK_LIB}
+		${OMR_TRACE_LIB}
+)
```
rwy0717: Shouldn't these links be public?

dnakamura: i dont think so. plus on a static lib it doesnt make a difference anyway

rwy0717: I'm thinking more about include directory propagation.

dnakamura: I try to keep include directory propagation down to a minimum. I need to double check but I dont believe that you need any of  the hook or trace include directories in order to consume the apis exposed by this library

charliegracie:
@genie-omr build all

fvtest/vmtest/CMakeLists.txt
charliegracie: This may be the cause of the AIX compile failure :)

charliegracie:
@dnakamura there are compile failures on AIX that will need to be cleaned up

dnakamura:
@charliegracie  should be fixed

youngar:
@genie-omr build all

PR #2542, opened 2018-05-11
Add tracking of persistent memory in containers

dsouzai:
In the scenario where there is a need to add a class such as the following:
```
class Test
   {
public:
   Test(uintptr_t f1, uintptr_t f2)
      : _field1(f1),
        _field2(f2)
      { }

   uintptr_t _field1;
   uintptr_t _field2;
   };
```
into an STL container such as `std::list`. currently there's no way to know how much total persistent memory is used for `Test`. With this commit it is now possible to do so:

1. Add (or replace the existing `TR_ALLOC` with) `TR_ALLOC_SPECIALIZED` with the appropriate `TR_MemoryBase::ObjectType` enum value to the class definition
2. Declare and use the list, eg:
```
typedef TR::list<Test, Test::TrackedPersistentAllocator> TestList;
TestList myList(getTypedAllocator<Test, Test::TrackedPersistentAllocator>(Test::getPersistentAllocator()));
```

Because there is a concern of dll size increase and so since we don't necessarily want to replace all uses of `TR_ALLOC` with `TR_ALLOC_SPECIALIZED`, when you do step 2 above with an object that has a `TR_ALLOC` in its class definition, the allocation is attributed to `TR_MemoryBase::UnknownType` via a non templated class.

### DIFF
--- a/compiler/env/TRPersistentMemory.cpp
+++ b/compiler/env/TRPersistentMemory.cpp
@@ -39,6 +39,7 @@ namespace TR { class Compilation; }
 namespace TR { class PersistentInfo; }
 
 extern TR::Monitor *memoryAllocMonitor;
+extern const char * objectName[];
 
 namespace TR
    {
@@ -72,4 +73,27 @@ TR_PersistentMemory::TR_PersistentMemory(
    _persistentAllocator(TR::ref(persistentAllocator)),
    _totalPersistentAllocations()
    {
+   }
+
+void
+TR_PersistentMemory::printMemStats()
+   {
+   fprintf(stderr, "TR_PersistentMemory Stats:\n");
+   for (uint32_t i = 0; i < TR_MemoryBase::NumObjectTypes; i++)
+      {
+      fprintf(stderr, "\t_totalPersistentAllocations[%s]=%lu\n", objectName[i], (unsigned long)_totalPersistentAllocations[i]);
+      }
+   fprintf(stderr, "\n");
+   }
+
+void
+TR_PersistentMemory::printMemStatsToVlog()
+   {
+   TR_VerboseLog::vlogAcquire();
+   TR_VerboseLog::writeLine(TR_Vlog_MEMORY, "TR_PersistentMemory Stats:");
+   for (uint32_t i = 0; i < TR_MemoryBase::NumObjectTypes; i++)
+      {
+      TR_VerboseLog::writeLine(TR_Vlog_MEMORY, "\t_totalPersistentAllocations[%s]=%lu", objectName[i], (unsigned long)_totalPersistentAllocations[i]);
+      }
+   TR_VerboseLog::vlogRelease();
    }
--- a/compiler/env/TRMemory.hpp
+++ b/compiler/env/TRMemory.hpp
@@ -488,6 +488,9 @@ public:
 
    TR::PersistentInfo * getPersistentInfo() { return &_persistentInfo; }
 
+   void printMemStats();
+   void printMemStatsToVlog();
+
    uintptr_t _signature;        // eyecatcher
 
    friend class TR_Memory;
--- a/compiler/env/TRMemory.hpp
+++ b/compiler/env/TRMemory.hpp
@@ -503,6 +503,65 @@ public:
 
 extern TR_PersistentMemory * trPersistentMemory;
 
+class TR_TypedPersistentAllocatorBase
+   {
+public:
+
+   void *allocate(size_t size, void * hint = 0)
+      {
+      return trPersistentMemory->allocatePersistentMemory(size, TR_MemoryBase::UnknownType);
+      }
+   void deallocate(void * p, size_t sizeHint = 0) throw()
+      {
+      trPersistentMemory->freePersistentMemory(p);
+      }
+
+   friend bool operator ==(const TR_TypedPersistentAllocatorBase &left, const TR_TypedPersistentAllocatorBase &right)
+      {
+      return &left == &right;
+      }
+   friend bool operator !=(const TR_TypedPersistentAllocatorBase &left, const TR_TypedPersistentAllocatorBase &right)
+      {
+      return !operator ==(left, right);
+      }
+
+   // Enable automatic conversion into a form compatible with C++ standard library containers
+   template<typename T> operator TR::typed_allocator<T, TR_TypedPersistentAllocatorBase& >()
+      {
+      return TR::typed_allocator<T, TR_TypedPersistentAllocatorBase& >(*this);
+      }
+   };
+
+template <TR_MemoryBase::ObjectType O>
+class TR_TypedPersistentAllocator : public TR_TypedPersistentAllocatorBase
+   {
+public:
+
+   void *allocate(size_t size, void * hint = 0)
+      {
+      return trPersistentMemory->allocatePersistentMemory(size, O);
+      }
+   void deallocate(void * p, size_t sizeHint = 0) throw()
+      {
+      trPersistentMemory->freePersistentMemory(p);
+      }
+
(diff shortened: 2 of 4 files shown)


dsouzai:
@mpirvu could you review.

dsouzai:
If I try making use of this new functionality, GCC doesn't have a problem but xlC does. I need to follow up with that.

dsouzai:
Ugh, so with xlC, if I have a class like this
```
class Test
   {
public:
   TR_ALLOC_SPECIALIZED(TR_MemoryBase::Test)

   Test(uintptr_t f1, uintptr_t f2)
      : _field1(f1),
        _field2(f2)
      { }

   uintptr_t _field1;
   uintptr_t _field2;
   };
```
doing this:
```
TR::list<Test, Test::TrackedPersistentAllocator> myList(getTypedAllocator<Test, Test::TrackedPersistentAllocator>(Test::getPersistentAllocator()));
for (uint32_t i; i < 10; i++)
  {
  Test val(i*5, i*10);
  myList.push_front(val);
  }
fprintf(stderr, "Values in list are:\n");
for (TR::list<Test, Test::TrackedPersistentAllocator>::iterator i = myList.begin(); i != myList.end(); i++)
  {
  Test &val = *i;
  fprintf(stderr, "%d,%d\n", val._field1, val._field2);
  }
fprintf(stderr, "\n");
```
will cause this error
`omr/compiler/env/TypedAllocator.hpp", line 127.66: 1540-0218 (S) The call does not match any parameter list for "PMA_Test::operator new".`; the code in question is `static void construct(pointer p, const_reference prototype) { new(p) value_type(prototype); }` in TypedAllocator.hpp

However, doing this
```
TR::list<Test *, Test::TrackedPersistentAllocator> myList(getTypedAllocator<Test *, Test::TrackedPersistentAllocator>(Test::getPersistentAllocator()));
for (uint32_t i; i < 10; i++)
  {
  Test *val = new (PERSISTENT_NEW) Test(i*5, i*10);
  myList.push_front(val);
  }
fprintf(stderr, "Values in list are:\n");
for (TR::list<Test *, Test::TrackedPersistentAllocator>::iterator i = myList.begin(); i != myList.end(); i++)
  {
  Test *val = *i;
  fprintf(stderr, "%d,%d\n", val->_field1, val->_field2);
  }
fprintf(stderr, "\n");
```
is fine. For some reason xlC can't find the global `placement new` if the data in the container is not a pointer. The way to fix it is to add `void * operator new (size_t s, void *p) { return p; }`. Unfortunately, I can't add this to `TR_ALLOC_IMPL` because some classes already define this `placement new` (eg `OMR::Node::operator new(size_t s, void *ptr)`).

So, in the future, when this feature is used to allocate some class in a STL container, we're going to have to add the `placement new` definition into that class so that xlC builds. @mpirvu @0xdaryl are you guys ok with that compromise? The more general solution is to add the `placement new` definition to `TR_ALLOC_IMPL` but that requires cleaning up all other places where it is already defined.

mpirvu:
@dsouzai Could you please talk to someone from the xlC team to see whether they can provide a behavior similar to gcc?

> The more general solution is to add the placement new definition to TR_ALLOC_IMPL but that requires cleaning up all other places where it is already defined.

How many such places exist? Maybe my query is broken, but I don't find any.

dsouzai:
I'm not sure I understand why xlC is having issues, given https://github.com/eclipse/omr/blob/master/compiler/env/TypedAllocator.hpp#L32

dsouzai:
Turns out, if there other `operator new`s defined in a class, xlC will try to match a call to `operator new` only within the class; if it fails to match, it won't bother checking the global scope.

dsouzai:
Hm what's weird is this will fail on gcc:

```
#include <stdlib.h>
#include <stdint.h>
#include <new>
#define PERSISTENT_NEW_DECLARE uint32_t*
class X
{
   public:
      void* operator new(size_t s, PERSISTENT_NEW_DECLARE) { return malloc(s); }
      X(int n): id(n){ }
      ˜X(){ }
   private:
      int id;
      //        ...
};
int main()
{
   char* whole = new char[ 3 * sizeof(X) ];  // a 3-part buffer
   X * p1 = new((void*)whole) X(8);                 // fill the front
   char* seg2 = &whole[ sizeof(X) ];         // mark second segment
   X * p2 = new((void*)seg2) X(9);                  // fill second segment
   char* seg3 = &whole[ 2 * sizeof(X) ];     // mark third segment
   X * p3 = new((void*)seg3) X(10);                 // fill third segment
   p2->˜X();   // clear only middle segment, but keep the buffer
   // ...
   return 0;
}
```

So I'm not sure why in the more complicated scenario with templates gcc works fine. I wonder if gcc's behaviour is the more inconsistent one :S

lmaisons:
I talked with Irwin a bit, I think this may come down to an oversight on my part, where the `typed_allocator::construct` function which is currently:
```
static void construct(pointer p, const_reference prototype) { new(p) value_type(prototype); }
```
should instead be (notice the global scope resolution operator)
```
static void construct(pointer p, const_reference prototype) { ::new(p) value_type(prototype); }
```

dsouzai:
Thanks @lmaisons , using `::new` seems to do the trick.

dsouzai:
@mpirvu @0xdaryl  Could you please review.

lmaisons:
Would it make sense to re-order the commits? The operator new fix seems required in order for the primary commit to work?

dsouzai:
I ran pbuilds and also did some personal tests on multiple platforms; everything looks ok.

@mpirvu @vijaysun-omr This should be good to merge pending reviews/genie builds.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
I would like a review from @mpirvu please

mpirvu:
As far as I can tell the code looks good.
I have one suggestion and one question:
1) The example you gave in the documentation allocates the list itself on the stack while the list nodes are allocated using the persistent allocator. Most often we want both the list and the nodes to be allocated using the persistent allocator. Could you please add an example on how to do that?
2) The purpose of this work is to be able to track allocations in case some memory leak gets reported. Have you tested the implementation by allocating some list with a known number of elements and then tried to use some diagnostics routines (I forgot their names) which should return back exactly how much you added to the list?

dsouzai:
@mpirvu 

>The example you gave in the documentation allocates the list itself on the stack while the list nodes are allocated using the persistent allocator. Most often we want both the list and the nodes to be allocated using the persistent allocator. Could you please add an example on how to do that?

Added it to the doc.

> The purpose of this work is to be able to track allocations in case some memory leak gets reported. Have you tested the implementation by allocating some list with a known number of elements and then tried to use some diagnostics routines (I forgot their names) which should return back exactly how much you added to the list?

I added https://github.com/eclipse/omr/pull/2542/files#diff-3029c93c833245ff2f677c54428698c0R79 and I verified that it does report the total memory allocated by the container.

compiler/env/TRPersistentMemory.cpp
```diff
@@ -73,3 +74,26 @@ TR_PersistentMemory::TR_PersistentMemory(
...
    {
    }
+
+void
+TR_PersistentMemory::printMemStats()
+   {
+   fprintf(stderr, "TR_PersistentMemory Stats:\n");
+   for (uint32_t i = 0; i < TR_MemoryBase::NumObjectTypes; i++)
+      {
+      fprintf(stderr, "\t_totalPersistentAllocations[%s]=%d\n", objectName[i], _totalPersistentAllocations[i]);
```
mpirvu: `_totalPersistentAllocations[i]` is defined as `size_t` so  `%d` is not quite correct.
Ideally one should use `%zu`, but last time I tried (a few years back) one of the platforms didn't support it. The best way is probably to use `%lu` and cast it to `unsigned long`

dsouzai:
Changes made.

vijaysun-omr:
@genie-omr build all

PR #2544, opened 2018-05-14
AArch64: Stub files of InstOpCode

knn-k:
This commit implements stubs of InstOpCode files for aarch64.

### DIFF
--- a/compiler/aarch64/codegen/InstOpCode.hpp
+++ b/compiler/aarch64/codegen/InstOpCode.hpp
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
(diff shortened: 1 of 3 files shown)


knn-k:
Minimum stubs only.  InstOpCode.hpp is needed in implementing OMRLinkage.hpp.

compiler/aarch64/codegen/OMRInstOpCodeEnum.hpp
```diff
@@ -0,0 +1,9 @@
+/*
```
0xdaryl: This file needs a copyright header as well, even though it is an incomplete C++ file.

compiler/aarch64/codegen/InstOpCode.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef TR_ARM64_INSTOPCODE_INCL
+#define TR_ARM64_INSTOPCODE_INCL
+
+#include "codegen/OMRInstOpCode.hpp"
+
+#ifndef OPS_MAX
+#define OPS_MAX    TR::InstOpCode::ARM64NumOpCodes
```
0xdaryl: What will this be used for?  Nothing in the current OMR code generators (nor OpenJ9 for that matter) uses this macro even on P and Z which also define it.  I'd rather we don't define this until we actually have a use for it.

knn-k: This macro came from the p version of InstOpCode.  I am removing it.

knn-k:
Updated the files reflecting review comments.

PR #2546, opened 2018-05-14
Fix compiler compilation on Windows (part 1)

samolisov:
The compiler as well as JitBuilder aren't able to be compiled on Windows due to some compilation errors. This PR does the following:

 * Adds Windows-specific includes into CompileMethod (`process.h` vs `unistd.h`)
 * Fixes 'compiler/cs2/listof.h' compilation errors
 * Marks 'forward_list::iterator::*' and '->' as const (and align this to the `std::forward_list` implementation)
 * Fixes element to _element in forward_list.hpp
 * Fixes 'cs2/arrayof.h' compilation errors
 * Fixes compiling the output of `flex` on Windows: the `--wincompat` parameter changes `<unistd.h>` 
    unix include with `<io.h>` windows analog.
 * Removes the unused `dlfcn.h` header from the `jitbuilder/release` C++ files.
 * Removes the unused `math.h` header from the `fvtest/compilertest` C++ files.
 * Removes the unused `unistd.h` header from the `fvtest/compilertest` C++ files.
 * Eliminates an ambiguous call to `std::isnan` in `fvtest/compilertest/tests/OpCodesTest.hpp`.

The PR depends on #2434

### DIFF
--- a/compiler/control/CompileMethod.cpp
+++ b/compiler/control/CompileMethod.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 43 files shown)


samolisov:
I've added the WIP label: the PR must be updated once #2434 is merged.

samolisov:
I've updated commit 439ce76 and removed the WIP marker, the PR is ready for review and should be merged after #2434.

dibyendumajumdar:
Hi, I can confirm that this series indeed fixes compilation errors / build issues with Visual Studio 2017. It doesn't fully resolve issues as #2596  is also needed.

charliegracie:
@fjeremic can you take a quick look at this?

charliegracie:
@genie-omr build all

compiler/control/CompileMethod.cpp
```diff
@@ -21,11 +21,15 @@
 
 #include "control/CompileMethod.hpp"
 
+#if defined(OMR_OS_WINDOWS)
+#include <process.h>                           // for _getpid
```
fjeremic: Just validating - our minimum compiler version is VS2010 for Windows IIRC. Is `_getpid` supported there? I haven't been able to find the docs for it but it looks like it should be.

fjeremic:
> Fixes 'compiler/cs2/listof.h' compilation errors

The day we get to eliminate cs2 will be a triumphant day indeed. Thanks for fixing these - much appreciated!

compiler/control/CompileMethod.cpp
fjeremic: Did a bunch of digging. It looks like it is supported. The `_getpid()` call will boil down to a `GetCurrentProcessId()` WIN_API even on VS2010. I think we're good here.

PR #2549, opened 2018-05-15
AArch64: fix missing semi colon in atomic operations

jeanlego:
fixes the missing semi colon in include_core atomic operation that would prevent successful build for AArch64


0xdaryl:
A genie build is not required as this is an AArch64-specific fix.

rwy0717:
Thanks!

PR #2552, opened 2018-05-15
[ci skip] Fix line-ends

babsingh:
https://github.com/eclipse/omr/pull/2529 changed all files to have 
Linux line-ends (no CRLF pairs). I had copied old trace points which had
CRLF pairs. While rebasing, CRLF pairs were unnoticeable. Removing CRLF
pairs which were introduced unintentionally in
https://github.com/eclipse/omr/pull/2524.


babsingh:
fyi - @ashu-mehra

Issue #163, opened 2016-07-21
Create a simple logger for fvtest


renfeiw:
The main purpose of the logger is to provide different layers of logging level:
LEVEL_VERBOSE   -   detailed message for debug purpose.
LEVEL_INFO          -   default level.
LEVEL_WARN       -    warning message.
LEVEL_ERROR       -   error/failure message.
LEVEL_OFF            -    no message from logger.

Use function log(LogLevel ll, const char *format, ...) to write test message to stdout. If no log level is provided, the level defaults to LEVEL_INFO.

To control the finest log level to output, use command line option "-logLevel=". The accepted values are "verbose, info, warn, error, off" corresponding to the logging levels shown above. If no "-logLevel" is passed, the finest level is LEVEL_INFO.

We plan to change gctest to use logger first and set it as an example for the rest of the fvtests.

charliegracie:
closed via #169

PR #2553, opened 2018-05-15
Quiet output of rfib test

dnakamura:


charliegracie:
I think I would prefer just removing the prints from the test unless they are valuable for a developer that might run the test locally.  @mstoodle do you have any opinion?  I do think we want to do something here as this test adds about 4K lines of output.

smlambert:
Can you use the logger that is used by some of the other tests (and set most/all output to debug level?) #163

charliegracie:
It would be difficult to add the FV test logger to these examples since they are JitBuilder examples and only have a dependency of the JitBuilder library.  At this point I do not think we want to make them depend on any of the FV libraries.  @dnakamura I think if you just remove the 5 `Call("print*` calls (and also those functions) this test will be as quiet as it should be.  After looking into the code more I can not see any reason why we would want this output.  You could easily wrap those calls and their callers with a `#if defined(RFIB_DEBUG_OUTPUT)` and by default we do not have that enable we will not get the output.

charliegracie:
@dnakamura any update

PR #2554, opened 2018-05-17
Code fixes discovered from using new gcc

youngar:
Most of the the changes are related to improper initialization of non-trivial classes with `memset(buf, 0, ..)`.

Can you review @rwy0717?

### DIFF
--- a/gc/base/MemoryPoolSplitAddressOrderedListBase.cpp
+++ b/gc/base/MemoryPoolSplitAddressOrderedListBase.cpp
@@ -54,24 +54,18 @@
 #include "SweepPoolState.hpp"
 
 bool
-J9ModronHeapFreeList::initialize(MM_EnvironmentBase* env)
+J9ModronFreeList::initialize(MM_EnvironmentBase* env)
 {
-	if (!_lock.initialize(env, &env->getExtensions()->lnrlOptions, "J9ModronHeapFreeList:_lock")) {
+	if (!_lock.initialize(env, &env->getExtensions()->lnrlOptions, "J9ModronFreeList:_lock")) {
 		return false;
 	}
-	_freeList = NULL;
-	_timesLocked = 0;
 
-	_freeSize = 0;
-	_freeCount = 0;
+	_freeList = NULL;
 
 	/* Link up the inactive hints */
 	J9ModronAllocateHint* inactiveHint = (J9ModronAllocateHint*)_hintStorage;
 	J9ModronAllocateHint* previousInactiveHint = NULL;
 	uintptr_t inactiveCount = HINT_ELEMENT_COUNT;
-
-	_hintActive = NULL;
-	_hintLru = 0;
 
 	while (inactiveCount) {
 		inactiveHint->next = previousInactiveHint;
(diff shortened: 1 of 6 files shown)


gc/base/MemoryPoolSplitAddressOrderedListBase.cpp
```diff
@@ -54,25 +54,19 @@
 #include "SweepPoolState.hpp"
 
 bool
-J9ModronHeapFreeList::initialize(MM_EnvironmentBase* env)
+J9ModronFreeList::initialize(MM_EnvironmentBase* env)
```
charliegracie: Was the class name wrong?

gc/base/ParallelSweepChunk.hpp
```diff
@@ -89,7 +89,7 @@ class MM_ParallelSweepChunk : public MM_BaseNonVirtual
 	 */	
 	MMINLINE void clear()
 	{
-		memset(this, 0, sizeof (*this));
+		*this = MM_ParallelSweepChunk();
```
charliegracie: Are all of the fields in MM_ParallelSweepChunk properly initialized in the constructor?

rwy0717: The constructor is not initializing these fields:
```c++
#if defined(OMR_GC_CONCURRENT_SWEEP)
	MM_ParallelSweepChunk *_nextChunk;
	uintptr_t _concurrentSweepState;
#endif /* OMR_GC_CONCURRENT_SWEEP */
```

youngar: Woops, looks like two were missing:
```c++
#if defined(OMR_GC_CONCURRENT_SWEEP)
	MM_ParallelSweepChunk *_nextChunk;
	uintptr_t _concurrentSweepState;
#endif /* OMR_GC_CONCURRENT_SWEEP */
```

gc/base/MemoryPoolSplitAddressOrderedListBase.cpp
youngar: FWIW, it would not be hard to change the name to `J9ModronHeapFreeList`.  The usages of this class are in just a few places:
```c++
gc/base/MemoryPoolHybrid.cpp:12
gc/base/MemoryPoolHybrid.hpp:7
gc/base/MemoryPoolSplitAddressOrderedListBase.cpp:7
gc/base/MemoryPoolSplitAddressOrderedListBase.hpp:4
````
Any preference to a name?

charliegracie: keep your current changes as I had missed the mismatch in naming in the original code.

charliegracie:
@genie-omr build all

gc/base/MemoryPoolSplitAddressOrderedListBase.cpp
charliegracie: I would leave it the way you have it now.  I had originally missed the mismatch in names.  This is ok.

charliegracie:
Once @rwy0717 approves or gives a thumbs up and the CI testing finishes I will merge.

rwy0717:
LGTM

youngar:
These changes passed on an IBM internal Java build on all platforms.

PR #2555, opened 2018-05-17
Improve presentation of CI status in README file

### LINKED ISSUE #2395: too many build badges in the README
It'd be great if we presented the build status badges in a table, so we knew which build was which. This would also give developers a better idea of platform support.

Maybe something like:

| OS | Arch | Status |
|:---|:---|---|
| Linux | s390-64 |  |

FWIW here's what chakra does: https://github.com/Microsoft/ChakraCore/wiki/Build-Status

0xdaryl:
Fixes #2395

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ is available at https://www.apache.org/licenses/LICENSE-2.0.
 This Source Code may also be made available under the following
 Secondary Licenses when the conditions for such availability set
 forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
-General Public License, version 2 with the GNU Classpath 
+General Public License, version 2 with the GNU Classpath
 Exception [1] and GNU General Public License, version 2 with the
 OpenJDK Assembly Exception [2].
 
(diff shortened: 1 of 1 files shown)


0xdaryl:
@rwy0717 : any feedback?

0xdaryl:
Personally, I like the status right at the top so you don't have to go looking for it.  Not sure how others  feel...

0xdaryl:
But having said that, it does seem kind of "naked" on its own at the top.  Perhaps in a build-status section would look better but near the top.

rwy0717:
Yeah, that's what I was thinking too. Maybe we can put it right before we list off the OMR components. It could probably use a header too.

mstoodle:
maybe it would be less visually imposing if you run the platforms horizontally rather than vertically? Of course, I like to use a wide browser window, so maybe I'm especially biased that way...

0xdaryl:
A couple of alternatives:

**Vertical** -> https://github.com/0xdaryl/omr/tree/readme2
**Horizontal**  --> https://github.com/0xdaryl/omr/tree/readme3

I think the horizontal looks alright for now, but might be hard to scale as we add platforms.  I'm ok with it for now.

fjeremic:
> I think the horizontal looks alright for now, but might be hard to scale as we add platforms. I'm ok with it for now.

I'm in favor of the horizontal as well. I don't suspect we'll have scaling problems even if we double the size of the table as the images automatically scale as per examples [1] and [2].

[1] https://github.com/dotnet/roslyn
[2] https://github.com/dotnet/cli

0xdaryl:
PR has been updated with a horizontal presentation of build status at the top of the file.

fjeremic:
I'll wait until EOD in case anyone else has comments or concerns. I'll omit the genie builds as the change is only for the README.md.

charliegracie:
A few notes....

1. The AppVeyor build is actually Windows x86
2. Windows 64 bit is tested on the Eclipse OMR CI and can be found at this link: [Build Status](https://ci.eclipse.org/omr/job/Build-win_x86-64/)
3. The Travis link is actually testing the following platforms:
Linux x86-32
Linux x86-64 compressed references
OS X
4. Linux 64 bit is being tested on the Eclipse OMR CI here: [Build Status](https://ci.eclipse.org/omr/job/Build-linux_x86-64/)

Adding the Windows x86 platform (AppVeyor), updating the Linux x86-64/Windows x86-64 platforms to point to the OMR CI will mostly fix the chart.  I am not sure how to organize the Linux x86, Linux x86-64 compressed and OS X since they all use the same build link.

fjeremic:
@0xdaryl just making sure you took note of the concerns from @charliegracie. These seem like valid points to address.

0xdaryl:
Thanks @charliegracie .  I pushed a new version.  GFMD doesn't really give a lot of options for formatting tables (a line break would be super nice), but let me know what you guys think.

0xdaryl:
Maybe we should just call the first column `Travis CI`.  Because it's so eclectic I wonder if there's any  real value in itemizing each of the builds therein, and it just eats up real estate.

README.md
```diff
@@ -20,11 +20,12 @@ OpenJDK Assembly Exception [2].
...
 -->
 
-[![Travis Status](https://api.travis-ci.org/eclipse/omr.svg?branch=master)](https://travis-ci.org/eclipse/omr)
-[![Appveyor Status](https://ci.appveyor.com/api/projects/status/github/eclipse/omr?svg=true&branch=master)](https://ci.appveyor.com/project/eclipsewebmaster/omr)
-[![Build-linux_390-64 Status](https://ci.eclipse.org/omr/job/Build-linux_390-64/badge/icon)](https://ci.eclipse.org/omr/job/Build-linux_390-64/)
-[![Build-linux_ppc-64_le_gcc Status](https://ci.eclipse.org/omr/job/Build-linux_ppc-64_le_gcc/badge/icon)](https://ci.eclipse.org/omr/job/Build-linux_ppc-64_le_gcc/)
-[![Build-aix_ppc-64 Status](https://ci.eclipse.org/omr/job/Build-aix_ppc-64/badge/icon)](https://ci.eclipse.org/omr/job/Build-aix_ppc-64/)
+# Eclipse OMR
+
+| **Travis CI** (macOS, AArch64, x86-64 Compressed, Linter) | **Windows x86** | **Windows x86-64** | **Linux x86-64** | **Power 64-bit Linux** | **Power 64-bit AIX** | **Linux on Z 64-bit** |
```
fjeremic: I agree. I don't think we get much out of the `(macOS, AArch64, x86-64 Compressed, Linter)`. Let's just call it "Travis CI" and save the real estate.

PR #2556, opened 2018-05-17
Adding missed OS check for GS support in Port Library for zLinux

ebadyano:
Update make files to be able to use auxv.c helpers for s390 Linux
builds.


port/CMakeLists.txt
```diff
@@ -213,6 +213,10 @@ if(OMR_HOST_ARCH STREQUAL ppc)
 	if(OMR_HOST_OS STREQUAL linux)
 		list(APPEND OBJECTS auxv.c)
 	endif()
+	elsif(OMR_ARCH_S390)
+	if(OMR_HOST_OS STREQUAL linux)
+		list(APPEND OBJECTS auxv.c)
+	endif()
```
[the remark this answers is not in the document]
ebadyano: i think my indentation is wrong, one of the endif() is for the nested if() inside elseif(). I'll add a missing white space.

charliegracie:
@genie-omr build all

port/unix/auxv.c
```diff
@@ -31,7 +31,9 @@
 #include <elf.h>
 #include <link.h>
 #include <pthread.h>
+#if defined(LINUXPPC)
```
charliegracie: These changes seem to be enabling the compilation of this file on Linux 390.  How come we had to add new includes if we are compiling on Linux PPC?

ebadyano: the include was there before and I assume compiles fine on power, but for some reason that header wasn't found when compiling on s390 Linux. It didn't seem to be necessary for query_auxv(AT_HWCAP). @ymanton Younes, do you know <asm/elf.h> is used in the lib on Power?

ymanton: It's not necessary on ppc64le, probably not necessary on ppc64 or ppc32 either.

fjeremic: > It's not necessary on ppc64le, probably not necessary on ppc64 or ppc32 either.

@ebadyano we should probably just remove the `#if defined` and the `#include` as well then?

ebadyano: @ymanton Younce, is that okay with you, if we remove that include?

fjeremic:
@genie-omr build all

port/unix/auxv.c
ymanton: Yes it's fine.

Issue #2361, opened 2018-03-07
A nursery evacuation algorithm with improved throughput and localization of references


ktbriggs:
[TL;DR] [See also #2664]

[**Note** (Feb 1 2019): Evacuator development is being capped for the time being. I am leaving this comment intact for archaeologic reasons and preparing an updated overview of evacuator design, with discussion of the benchmarking results reported in #2664, in the comment subsequent to this, below. The update is a **work in progress** that I hope to complete over the next few days. This issue will be closed at that point.]

----------------------------------------------------------------------------------------------------------
I'm putting a new nursery evacuation algorithm for generational garbage collection forward for review and comments from the OMR community. Below is a sketch of evacuator design, intended to promote discussion within the OMR community. Existing source code, written offline to verify the evacuator design and to demonstrate evacuator simplicity and performance, will be made available soon, pending resolution of concerns relating to introducing a nearly completed work into OMR and OpenJ9.  I welcome constructive and pragmatic critique and recommendations.

# Background
Evacuator was developed off-line as proof of concept to verify a idea relating to the direction of copy during generational collections, ie the movement of objects from evacuation space to survivor and tenure regions. At bottom generational collection is a straightforward process that consumes most of its processing time in a single method, `memcpy()`. Specifically, evacuator selectively filters out large (>256 byte) objects for breadth-first (long-distance between referring pointer and referent object) copying and limits hierarchical (copying referent object close to referring pointer) copying to small objects, whereby the likelihood of hardware cache-line containment of referring pointer and referent head is increased for small objects. The likelihood of cache-line containment for reference arcs decreases with increasing distance between the object being scanned for referent objects in evacuation space and the next location available for receiving copied referents in survivor or tenure space. Scavenger's aliasing algorithm attempts to minimize this distance by opportunistically switching alias between survivor and tenure copy/scan caches, but this provides a limited degree of freedom for control and, with large copy/scan caches, frequently diverges to scan-copy distances that do nothing to improve locality of reference while also inhibiting production of shareable work between gc threads.

[...]

Issue #1638, opened 2017-09-13
Externalize slot object for GC_ObjectScanner


ktbriggs:
GC_ObjectScanner (omr/gc/base/ObjectScanner.hpp) contains a GC_SlotObject _slotObject member that it uses to communicate found reference slots to calling code in getNextSlot(). This arrangement prohibits calling code from using a subclass of GC_SlotObject that may contain additional object metadata, eg object size, that is important to its task. Also, it opens up the possibility of leaving the caller with a dangling GC_SlotObject pointer after the GC_ObjectScanner instance has been destroyed. 

A GC_SlotObject * _slotObjectPtr member (private) should be added to GC_ObjectScanner, and a new constructor. In the existing constructor, _slotObjectPtr should be set to the address of the existing _slotObject member, to maintain compatibility with existing code that uses GC_ObjectScanner. The new constructor should present all of the parameters of the existing construct plus a GC_SlotObject * valued parameter that can receive a pointer to a GC_SlotObject or subclass instance of the caller's choosing. In the new constructor, _slotObjectPtr must be set to the parameter's value.

The implementation of `GC_ObjectScanner::getNextSlot()` should then be refactored to use exclusively _slotObjectPtr, and any other direct references to _slotObject.x must be replaced with  _slotObjectPtr->x.

Object scanning is interruptible, so GC_ObjectScanner must present a new public method `setSlotObject(GC_SlotObject *x)` to set _slotObjectPtr = x, To be called when restoring interrupted scanning state. 

Additionally, all subclasses of GC_SlotObject must add a new constructor that receives a GC_SlotObject parameter and passes it down the inheritance chain to the new GC_ObjectScanner constructor. 

Finally, a comment must be added to the existing GC_ObjectScanner and subclass constructors to indicate that they are DEPRECATED and will be removed at some point in the future, with instructions to use the new constructor.

smlambert:
@amicic - do you feel this an interesting exercise for some one?

amicic:
On a quick pass, I'm not sure I understood the point. I'll pass on@rwy0717 and @dmitripivkine (who are anyway more familiar with it) to comment.

dmitripivkine:
I believe it is not

rwy0717:
I'm closing this issue, because I don't think it's a change we will want to make.

PR #2557, opened 2018-05-17
Evacuator (OMR integration)

ktbriggs:
Evacuator is the working name for a novel stack-based generational
collection algorithm. MM_Scavenger subclasses MM_EvacuatorController
which instantiates and controls MM_Evacuator instances. MM_Evacuator
instances are bound to and unbound from MM_Environment of gc threads
harnessed by MM_EvacuatorParallelTask.

Otherwise only minimal changes were required to plug the evacuator 
framework into scavenger. The intention is to minimize changes to
existing gencon reliability/performance/maintenance (eg, allowing
concurrent scavenger work to proceed with minimal regard to evacuator
development). 

Evacuator is disabled by default in this commit. To enable evacuator, 
so that the evacuator algorithm will run in place of 
`MM_Scavenger::workThreadGarbageCollect()`, set 
`MM_ExtensionsBase::evacuatorEnabled = true` before instantiating the 
generational collector.

Evacuator is capable of producing detailed tracing output for debugging
purposes and instrumentation. To enable this capability, and enable a 
number of debugging assertions, edit EvacuatorBase.hpp and replace 

```
#undef EVACUATOR_DEBUG
```

with

```
#define EVACUATOR_DEBUG
```

Debug tracing levels can then be controlled by exporting a environment
variable `EVACUATOR_DEBUG_FLAGS`, which is a bitmap of bit flags as 
defined in EvacuatorBase.hpp. 

Related to: #2361


charliegracie:
@genie-omr build all

ktbriggs:
This was getting a bit stale. Rebased on eclipse/omr/master and pushed.

@charliegracie can we get an update on this?

charliegracie:
Hi @ktbriggs , this is a very large PR which contains over 5kloc so it will take some time to complete the review.  This will require an Eclipse CQ to be approved before we can merge as well.  I will get the CQ opened tomorrow.

fvtest/gctest/GCConfigTest.cpp
```diff
@@ -300,18 +307,8 @@ GCConfigTest::allocateHelper(const char *objName, uintptr_t size)
...
 
-	uint8_t objectAllocationModelSpace[sizeof(MM_ObjectAllocationModel)];
-	MM_ObjectAllocationModel *noGc = new(objectAllocationModelSpace)
-			MM_ObjectAllocationModel(env, size, MM_ObjectAllocationModel::selectObjectAllocationFlags(false, false, false, true));
-	objEntry.objPtr = OMR_GC_AllocateObject(exampleVM->_omrVMThread, noGc);
-
-	if (NULL == objEntry.objPtr) {
-		gcTestEnv->log("No free memory to allocate %s of size 0x%llx, GC start.\n", objName, size);
-		MM_ObjectAllocationModel *withGc = new(objectAllocationModelSpace)
-				MM_ObjectAllocationModel(env, size, MM_ObjectAllocationModel::selectObjectAllocationFlags(false, false, false, false));
```
charliegracie: Can you explain this change in more detail?  It looks like we are removing part of the test that verifies we can allocate without a GC happening if we pass the right flags.

gc/base/EvacuatorBase.hpp
```diff
@@ -0,0 +1,214 @@
...
+#define Debug_MM_true1(env, assertion, format, arg)
+#define Debug_MM_true2(env, assertion, format, arg1, arg2)
+#define Debug_MM_true3(env, assertion, format, arg1, arg2, arg3)
+#define Debug_MM_true4(env, assertion, format, arg1, arg2, arg3, arg4)
+#endif /* defined(EVACUATOR_DEBUG) */
+
+#include "GCExtensionsBase.hpp"
+
+#if (DEFAULT_SCAN_CACHE_MINIMUM_SIZE > (DEFAULT_SCAN_CACHE_MAXIMUM_SIZE >> 3))
+#error "Scan cache default sizes must satisfy DEFAULT_SCAN_CACHE_MINIMUM_SIZE <= (DEFAULT_SCAN_CACHE_MAXIMUM_SIZE >> 3)"
```
charliegracie: Can you explain this more with a comment?

gc/base/EvacuatorCopyspace.hpp
```diff
@@ -0,0 +1,199 @@
...
+	 * is allocated from forge as contiguous block sized to contain requested number of elements and
+	 * must be freed using MM_Forge::free() when no longer needed.
+	 *
+	 * @param count the number of aray elements to instantiate
+	 * @return a pointer to instantiated array
+	 */
+	static MM_EvacuatorCopyspace *
+	newInstanceArray(MM_Forge *forge, uintptr_t count)
+	{
+		MM_EvacuatorCopyspace *copyspace = (MM_EvacuatorCopyspace *)forge->allocate(sizeof(MM_EvacuatorCopyspace) * count, OMR::GC::AllocationCategory::FIXED, OMR_GET_CALLSITE());
```
charliegracie: Where is this memory freed?

gc/base/EvacuatorHistory.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+	/* get the predecessor of the epoch that is being prepared for commit */
+	Epoch *last() { return &_history[((0 < _epoch) && (reports_per_cycle > _epoch)) ? (_epoch - 1) : epochToIndex()]; }
+
+	/* commit closing epoch and open a new one */
+	void commit(Epoch *epoch) { Debug_MM_true(epoch == &_history[epochToIndex()]); VM_AtomicSupport::add(&_epoch, 1); }
+
+	/* clear history for starting a gc cycle */
+	void
+	reset(uintptr_t gc = 0, uintptr_t tlhAllocationCeiling = 0, uintptr_t releaseThreshold = 0)
```
charliegracie: Are their any uses of this function that do not pass a GC parameter?

gc/base/EvacuatorParallelTask.cpp
```diff
@@ -0,0 +1,124 @@
...
+	MM_EnvironmentStandard *env = MM_EnvironmentStandard::getEnvironment(envBase);
+	OMRPORT_ACCESS_FROM_OMRPORT(envBase->getPortLibrary());
+	uint64_t startTime = omrtime_hires_clock();
+	_controller->waitToSynchronize(env->getEvacuator(), id);
+
+	bool result = MM_ParallelTask::synchronizeGCThreadsAndReleaseMaster(env, id);
+	uint64_t endTime = omrtime_hires_clock();
+	env->_scavengerStats.addToSyncStallTime(startTime, endTime);
+	
+	_controller->continueAfterSynchronizing(env->getEvacuator(), startTime, endTime, id);
```
charliegracie: same as above

gc/base/EvacuatorParallelTask.cpp
```diff
@@ -0,0 +1,124 @@
...
+	MM_EnvironmentStandard *env = MM_EnvironmentStandard::getEnvironment(envBase);
+	OMRPORT_ACCESS_FROM_OMRPORT(envBase->getPortLibrary());
+	uint64_t startTime = omrtime_hires_clock();
+	_controller->waitToSynchronize(env->getEvacuator(), id);
+
+	bool result = MM_ParallelTask::synchronizeGCThreadsAndReleaseSingleThread(env, id);
+	uint64_t endTime = omrtime_hires_clock();
+	env->_scavengerStats.addToSyncStallTime(startTime, endTime);
+
+	_controller->continueAfterSynchronizing(env->getEvacuator(), startTime, endTime, id);
```
charliegracie: same as above

gc/base/EvacuatorScanspace.hpp
```diff
@@ -0,0 +1,252 @@
...
+	 * is allocated from forge as contiguous block sized to contain requested number of elements and
+	 * must be freed using MM_Forge::free() when no longer needed.
+	 *
+	 * @param count the number of aray elements to instantiate
+	 * @return a pointer to instantiated array
+	 */
+	static MM_EvacuatorScanspace *
+	newInstanceArray(MM_Forge *forge, uintptr_t count)
+	{
+		MM_EvacuatorScanspace *scanspace = (MM_EvacuatorScanspace *)forge->allocate(sizeof(MM_EvacuatorScanspace) * count, OMR::GC::AllocationCategory::FIXED, OMR_GET_CALLSITE());
```
charliegracie: Where is this memory freed?

gc/base/EvacuatorWorklist.hpp
```diff
@@ -0,0 +1,325 @@
...
+/*
+ * Function members
+ */
+private:
+protected:
+public:
+	/**
+	 * Returns a pointer to the volatile sum of the number of bytes contained in work packets in the list
+	 */
+	MMINLINE volatile uint64_t *volume() { return &_volume; }
```
charliegracie: Is this returning the address of the value so other code can modify the contents?  If so wouldn't it make more set to add an API to update the volume?

gc/base/GCExtensionsBase.hpp
```diff
@@ -832,6 +833,16 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
 #endif /* defined(OMR_GC_MODRON_SCAVENGER) */
 	}
 
+	MMINLINE bool
+	isEvacuatorEnabled()
+	{
+#if defined(OMR_GC_MODRON_SCAVENGER)
+		return evacuatorEnabled && isScavengerEnabled() && !isConcurrentScavengerEnabled();
```
charliegracie: can the evacuator be enabled and not the scavenger?  I think it would be better if this code looked like this:
```
Assert_MM_true(isScavengerEnabled());
Assert_MM_true(!isConcurrentScavengerEnabled());
return evacuatorEnabled;
```

gc/base/ObjectScanner.hpp
```diff
@@ -55,15 +55,15 @@ class GC_ObjectScanner : public MM_BaseVirtual
 protected:
 	static const intptr_t _bitsPerScanMap = sizeof(uintptr_t) << 3;
 
-	omrobjectptr_t const _parentObjectPtr;	/**< Object being scanned */
 	uintptr_t _scanMap;						/**< Bit map of reference slots in object being scanned (32/64-bit window) */
+	fomrobject_t *_scanPtr;					/**< Pointer to base of object slots mapped by current _scanMap */
+	GC_SlotObject _slotObject;				/**< Create own SlotObject class to provide output */
```
charliegracie: Can you document why the order of the fields changed?  If there was a reason it would be beneficial to document it in the comments.

gc/base/ObjectScanner.hpp
```diff
@@ -309,6 +309,8 @@ class GC_ObjectScanner : public MM_BaseVirtual
 
 	MMINLINE bool hasMoreSlots() { return 0 == (GC_ObjectScanner::noMoreSlots & _flags); }
 
+	MMINLINE bool isEmpty() { return (0 == _scanMap) && !hasMoreSlots(); }
```
charliegracie: I would expect _scanMap to only be 0 is hasMoreSlots returns false.  Can you explain why this is not true?

gc/base/ParallelDispatcher.hpp
```diff
@@ -112,12 +112,11 @@ class MM_ParallelDispatcher : public MM_Dispatcher
 	
 	virtual void setThreadInitializationComplete(MM_EnvironmentBase *env);
 	
-	uintptr_t adjustThreadCount(uintptr_t maxThreadCount);
-	
 public:
 	virtual bool startUpThreads();
 	virtual void shutDownThreads();
 	
+	uintptr_t adjustThreadCount(uintptr_t maxThreadCount);
```
charliegracie: Can you explain why this is now public?

gc/stats/ScavengerStats.hpp
```diff
@@ -219,6 +216,18 @@ class MM_ScavengerStats
 		_copy_cachesize_sum += copyCacheSize;
 	}
 
+	MMINLINE void
+	countObjectSize(uintptr_t objectSize)
+	{
```
charliegracie: Is this only used for the Evacuator or does the Scavenger also track this data now?

charliegracie: If it is only used by the Evacuator we need to add a compile time define as this is growing the memory requirements of the Scavenger for each thread in the system

ktbriggs:
@charliegracie I will hold off on further commits until you have had an opportunity to complete your review.

fvtest/gctest/GCConfigTest.cpp
ktbriggs: You are correct, I will revert this changes and retest gctest. Thanks for spotting this.

gc/base/EvacuatorBase.hpp
ktbriggs: Will do. Evacuator wants to ensure that there is a reasonable range between min/max whitespace allocation sizes.

gc/base/EvacuatorCopyspace.hpp
ktbriggs: In MM_Evacuator::tearDown(). Otherwise MM_EvacuatorCopyspace would be required to cache a pointer to MM_Forge.

gc/base/EvacuatorHistory.hpp
```diff
@@ -0,0 +1,123 @@
...
+	uintptr_t epochToIndex() { return epochToIndex(_epoch); }
+
+protected:
+
+public:
+	/* number of committed epochs per this gc cyle */
+	uintptr_t count() { return _epoch; }
+
+	/* get a past (committed) epoch */
+	Epoch *epoch(uintptr_t epoch)
```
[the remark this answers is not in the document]
ktbriggs: I will fix this.

gc/base/EvacuatorParallelTask.cpp
```diff
@@ -0,0 +1,124 @@
...
+void
+MM_EvacuatorParallelTask::synchronizeGCThreads(MM_EnvironmentBase *envBase, const char *id)
+{
+	MM_EnvironmentStandard *env = MM_EnvironmentStandard::getEnvironment(envBase);
+	OMRPORT_ACCESS_FROM_OMRPORT(envBase->getPortLibrary());
+	uint64_t startTime = omrtime_hires_clock();
+	_controller->waitToSynchronize(env->getEvacuator(), id);
+	MM_ParallelTask::synchronizeGCThreads(env, id);
+	uint64_t endTime = omrtime_hires_clock();
+	_controller->continueAfterSynchronizing(env->getEvacuator(), startTime, endTime, id);
```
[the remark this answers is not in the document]
ktbriggs: This code is correct as is.

gc/base/EvacuatorParallelTask.cpp
ktbriggs: This code is correct as is.

gc/base/EvacuatorParallelTask.cpp
ktbriggs: This code is correct as is.

gc/base/EvacuatorScanspace.hpp
ktbriggs: In MM_Evacuator::tearDown().

gc/base/EvacuatorWorklist.hpp
ktbriggs: Worklist volume is adjusted whenever a work packet is added or removed. It is not externally settable.

gc/base/GCExtensionsBase.hpp
ktbriggs: I agree and will change this as you suggest.

gc/base/ObjectScanner.hpp
ktbriggs: After _scanMap, these fields are the hottest fields in the object scanner, so I moved them closer to _scanMap to ensure that they would be more likely to be contained in same cache line. 

I agree that more detailed object scanner documentation is required. @rwy0717 is pushing that envelope as well, and will likely force implementation of https://github.com/eclipse/omr/issues/1638 at least. I intend to write a boxnote to explain how to implement object scanners for languages with alignment and reference encoding special needs. For now I will  include a synopsis of this in ObjectScanner.hpp comment. Beyond that I suggest we treat this (object scanner documentation) as a separate issue.

gc/base/ObjectScanner.hpp
ktbriggs: Large objects (and split array segments) may contain >64 object references and will require >1 scan map to feed to object scanner. The _scanMap == 0 condition will occur with hasMoreSlots(true) for each scan map but the last after each scan map is exhausted.

gc/base/ParallelDispatcher.hpp
ktbriggs: Evacuator controller uses this to force determination of gc thread count before dispatching slave threads. Metronome also needs to know gc thread count before dispatching its slave threads. If not called directly as evacuator controller is doing, the determination of gc thread count is not made until after controller starts its parallel task threads in scavenge(). 

There are other ways to do this (see metronome), but this is the most straightforward way and it works fine with MM_ParallelDispatcher/MM_ParallelTask.

gc/stats/ScavengerStats.hpp
ktbriggs: Scavenger also records object size histograms. Scavenger and evacuator produce identical summary information, including object size histograms and performance-related metrics, at the end of each generational gc, when EVACUATOR_DEBUG_ALWAYS is defined in EvacuatorBase.hpp. 
Scavenger does not use object size stats in any other way,  but having these stats for both algorithms helps to eliminate object size differences as dependent factors contributing to observed differences in performance-related metrics when running scavenger vs evacuator benchmarking trials.

For evacuator, object size histograms may be useful in determining optimal values for some of the controllable parameters now declared as constants in MM_EvacuatorBase, eg, definition of "small object", ie an object that can be copied inside the stack. An upcoming phase of evacuator development will involve migrating some of these constants into parameters that can be set via command-line options. The object size stats will then likely be compiled in by default (ie without EVACUATAOR_DEBUG_ALWAYS).

charliegracie:
Due to the size of this contribution an Eclipse CQ needs to be approved . I have created the [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16676).

gc/base/EvacuatorWorklist.hpp
charliegracie: I see code using _volume directly so I am still confused why this API exists in its current form.

gc/base/EvacuatorScanspace.hpp
charliegracie: Could this be placed in a comment for the newInstanceArray so it is clear to everyone where it is being freed.

gc/base/EvacuatorCopyspace.hpp
charliegracie: Could this be placed in a comment for the newInstanceArray so it is clear to everyone where it is being freed.

gc/base/EvacuatorParallelTask.cpp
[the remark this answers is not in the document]
charliegracie: ok the symmetry is important :)

gc/base/EvacuatorWorklist.hpp
ktbriggs: _volume is a volatile private member of MM_EvacuatorWorklist and is modified directly only within that class. The volume() method is public to allow MM_Evacuator instances to assess the volume of other evacuator's worklist. It returns a pointer to this volatile member only to ensure that the caller obtains the volume value only when it dereferences the pointer. To return the value directly from the volume() method could result is use of a stale value when returned value is applied.

gc/base/EvacuatorScanspace.hpp
ktbriggs: As above.

gc/base/EvacuatorCopyspace.hpp
ktbriggs: Will do.

gc/base/GCExtensionsBase.hpp
ktbriggs: Unfortunately ModronAssertions.h cannot be included in GCExtensionsBase.hpp without compilation errors (redefined macro, unrelated to current PR changes). Leaving this test as is will simply force scavenger to be selected for generational GC if concurrent scavenger is enabled, without warning as long as scavenger is enabled, and evacuator enablement is effective only when gencon is selected as gc policy (because evacuator is implemented in MM_Scavenger, which is only instantiated for gencon). I don't know that printing an error message in this context is warranted or would be useful.

ktbriggs:
Last commit (Enable >64 GC threads) addresses a bone of contention that was brought up in reviews. I pushed it here in order to enable a build on the farm to obtain an SDK that can run on available machines with >64 cores, which require glibc2.1X (my personal build vm builds for glibc2.4), 

As stated in the commit message, I think this commit is a good candidate for inclusion with the PR. Testing with <64 gc threads looks fine and none of this will be involved in runtimes that do nor explicity request evacuator in their runtime. Review should be pretty straightforward as many of the changes relate to different means of printing debug output involving bitmaps. 

But if this commit will significantly (>2 days) impede PR merge I will back this commit onto a new branch and continue there -- @charliegracie let me know.

ktbriggs:
Rebased and pushed (no source changes). The previous continuous-integration/travis-ci/pr segfaulted on omrgctest almost immediately. Built omr from same branch/commit on a fyre vm and ran 100x grinder and 1000x grinder for built omrgctest without failure. I believe previous failure was unrelated to commit per se.

ktbriggs:
Squashed 3 commits related to enabling >64 gc threads and removed #define macro that enables debug output, which should not be enabled for ci/production builds for this PR in any case. Reasons for previous travis/ci failure in gctest (segfault) were related to debug output so this should let the travis build run without failing. 

Previous travis/ci failures fo omrgctest were not reproducible on a Linux Ubuntu 16.04 vm on fyre after cloning and building this branch (see previous comment).

ktbriggs:
I've completed changes relating to "Enable >64 GC threads" and rebased. Evacuator scaling up to 120 GC threads has been verified on a linux/x86-64 platform with 120 cores running SPECjbb2005 up to 120 warehouses. 

There is no apparent regression in evacuator performance for up to 8 CPUS. Typical results for SPECjbb2005 runoffs between scavenger and evacuator on 8-core linux/x86-64 look like this:

Scavenger:
```
collections: 2 global; 0 backouts; 3253 generational
idle time: 3.199% stall; 6.037% complete; .753% sync; 36542.617 gc-ms
throughput: 54067988736 allocated; 53643943432 evacuated; 36542.617 ms; 1433.577 kb/ms
fragmentation: 0 (0%) micro; 450032400 (.832%) macro; 424045304 (.785%) global
containment: 6.397% <64; 9.508% <128; 14.873% <256; 97.341% small (<= 256 bytes)
```
Evacuator:
```
collections: 2 global; 0 backouts; 3491 generational
idle time: 2.393% stall; 3.538% complete; .712% sync; 34783.991 gc-ms
throughput: 58201047400 allocated; 57708802976 evacuated; 34783.991 ms; 1620.177 kb/ms
fragmentation: 16547488 (.028%) micro; 468634864 (.805%) macro; 492244424 (.846%) global
containment: 13.682% <64; 18.655% <128; 25.583% <256; 97.486% small (<= 256 bytes)
```

More benchmarking and tuning is required to fully assess relative performance for >64 GC threads. Initial benchmarking with 120 threads showed evacuator more performant for SPECjvm2008/xml.validation but less so for SPECjbb2005. 

@charliegracie this PR is complete -- there will be no more source changes. Let me know if/when you would like to have these commits squashed.

gc/base/ParallelDispatcher.hpp
charliegracie: I am still confused as why the normal call to adjust does not work for Evacuator?

charliegracie:
The Eclipse CQ has been approved to be merged but it has not been completed.  There are currently discussing about doing an OMR release shortly.  The project can not have code from open CQs in the code base at the time of the release.  I am making some queries to see how long it will take to get the CQ completed.

charliegracie:
@genie-omr build all

example/glue/EvacuatorDelegate.hpp
```diff
@@ -0,0 +1,245 @@
...
+	 * @see MM_EvacuatorController::max_evacuator_public_flag (public bit mask is 0xffffffff)
+	 */
+	enum {
+		unused = 1
+	};
+private:
+	MM_EnvironmentStandard *_env;
+	MM_Scavenger *_controller;
+	MM_Evacuator *_evacuator;
+	bool _isCleared;
```
charliegracie: Can you document the instance variables please?

example/glue/EvacuatorDelegate.hpp
```diff
@@ -0,0 +1,245 @@
...
+		, _controller(NULL)
+		, _evacuator(NULL)
+		, _isCleared(false)
+	{ }
+
+#if defined(EVACUATOR_DEBUG_DELEGATE)
+	void debugValidateObject(omrobjectptr_t objectptr) { }
+	void debugValidateObject(MM_ForwardedHeader *forwardedHeader) { }
+	const char *
+	debugGetClassname(omrobjectptr_t objectptr, char *buffer, uintptr_t bufferLength)
```
charliegracie: What is the debugGetClassName used for?

gc/base/Evacuator.cpp
```diff
@@ -0,0 +1,1438 @@
...
+
+/**
+ * Per gc cycle setup. This binds the evacuator instance to a gc worker thread for the duration of the cycle.
+ *
+ * @param env the environment for the gc worker thread that is bound to this evacuator instance
+ */
+void
+MM_Evacuator::bindWorkerThread(MM_EnvironmentStandard *env)
+{
+	omrthread_monitor_enter(_mutex);
```
charliegracie: Why do you need a monitor to bing the threads?  This should be happening as part of the dispatch to the helper threads which is already under the ParallelDispatcher lock.

gc/base/Evacuator.cpp
```diff
@@ -0,0 +1,1438 @@
...
+
+	/* bind evacuator and delegate to executing gc thread */
+	_env = env;
```
charliegracie: All new GC algorithm code takes an Environment as a parameter rather than caching it. Can you modify the callers which rely on the cached env to take an env parameter instead?

gc/base/Evacuator.cpp
```diff
@@ -0,0 +1,1438 @@
...
+	_env->setEvacuator(this);
+	_delegate.cycleStart();
+
+	/* clear worker gc stats */
+	_stats = &_env->_scavengerStats;
+	_stats->clear(true);
+	_stats->_gcCount = _env->getExtensions()->scavengerStats._gcCount;
+
+	/* Reset the local remembered set fragment */
+	_env->_scavengerRememberedSet.count = 0;
```
charliegracie: Is this the remembered set cache that is used during the GC for the current thread?

example/glue/EvacuatorDelegate.hpp
ktbriggs: Will do.

example/glue/EvacuatorDelegate.hpp
ktbriggs: This used used with tracing (EVACUATOR_DEBUG) object copying. Classname and size are printed. This is never enabled in submitted evacuator PRs.

gc/base/Evacuator.cpp
ktbriggs: This is required as dispatcher may already started other evacuator threads and these may ping binding evacuator's worklist while it is initializing.

gc/base/Evacuator.cpp
ktbriggs: I am pushing back hard on this. We have discussed this at length. The evacuator-environment binding is 1-1, and ensures that an evacuator cannot work on any thread other that the one it is bound to. Passing any other env as parameter could break this binding. It would be redundant at best and misleading and possibly dangerous at worst.

gc/base/Evacuator.cpp
ktbriggs: It is.

example/glue/EvacuatorDelegate.hpp
ktbriggs: Done. Also removed redundant MM_EvacuatorDelegate::_env field and updated method comments to explain the persistence of the evacuator-delegate binding (1-1, invariant over delegate lifetime) and environment-(evacuator-delegate) binding (1-(1-1), invariant over duration of gc cycle, dissolved between gc cycles)

ktbriggs:
I added this to #2361 but I am reproducing these results here to highlight that work on this branch is piling up while waiting for this to be pulled (or flushed as the case may be). 

>Some results from preliminary benchmarking after allowing >64 GC threads and tuning for low-volume live sets (such as jbb2005). These results are from 5x2 (evacuator vs scavenger) SPECjbb2005 runs on an fyre ubuntu1604 vm with 8 cpus. Heap size is -Xms=-Xmx=1g. There no global collections except for the 2 that are always incurred with this benchmark during startup (these are small and precede any generational collections);

|metric|scavenger|evacuator|%difference|
|-----:|--------:|--------:|----------:|
|total SPEC score|142771|148396|3.9|
|total generational gc time (ms)|193536|168160|-15.1|
|average throughput (kb/ms)|1437|1721|19.8|

@charliegracie can you explain what is holding this decision (merge or flush) up? This has been dangling for >6 weeks here, and the "discussion" opened for internal review of the initial commit is >3months stale. If you do not wish to pursue this then please close it.

gc/base/ParallelDispatcher.hpp
charliegracie: Is there any more details?

charliegracie: Is there any more details?

gc/base/Evacuator.cpp
charliegracie: It would be nice if binding happened under the dispatcher lock and didn’t need this lock but we can look into that in the future.

gc/base/Evacuator.cpp
charliegracie: Kim sometimes I ask questions we have discussed in private so the decisions / discussion is documented for others.  It would be best if all discussion happened in the public so everyone can see.

ktbriggs:
I added this to #2361 but I am reproducing these results here to highlight that work on this branch is piling up while waiting for this to be pulled (or flushed as the case may be). 

>Some results from preliminary benchmarking after allowing >64 GC threads and tuning for low-volume live sets (such as jbb2005). These results are from 5x2 (evacuator vs scavenger) benchmark runs on an fyre ubuntu1604 vm with 8 cpus. Heap size is -Xms=-Xmx=1g. There no global collections except for the 2 that are always incurred with this benchmark during startup (these are small and precede any generational collections);

|metric|scavenger|evacuator|%difference|
|-----:|--------:|--------:|----------:|
|total benchmark score|142771|148396|3.9|
|total generational gc time (ms)|193536|168160|-15.1|
|average throughput (kb/ms)|1437|1721|19.8|

@charliegracie can you explain what is holding this decision (merge or flush) up? This has been dangling for >6 weeks here, and the "discussion" opened for internal review of the initial commit is >3months stale. If you do not wish to pursue this then please close it.

ktbriggs:
Charlie thank you for clarifying this. I of course do not wish to impede release schedules in any way. Please bear in mind that the intention for getting this into regular builds is to allow other teams (especially the performance team) to evaluate this in stable builds. Also that none of the evacuator functionality is enabled in builds that include evacuator unless explicitly requested. It is disabled by default in OpenJ9 builds and, similarly, in any other builds that consume OMR gencon. So I suggest that there is very low risk in including this in OpenJ9 and any other builds that involve OMR. This can be verified by reviewing changes in those files included in this PR orther than Evacuator*.*pp (ie, changes to pre-existing files).

Regarding evacuator scalability, benchmarking with SPECjbb2005 suggests that evacuator performance drops below scavenger beyond 16 gc threads. The reason for this is that each evacuator thread is left holding a chunk of reserved but unused memory at the end of each cycle (evacuator whitelists ensure that fragmentation occurring before end of cycle are recycled -- this internal fragmentation typically accounts for <0.01% of total fragmentation). This accounts for observed increase in frequency of nursery collections, as mutator exhausts available allocation space more quickly, and this effect increases as the number of gc threads increases.

To date I have tried a number of different heuristics to anticipate end of cycle but have not yet found a stable solution, but I have a prospective solution for this, I think, but I have  yet to try it out. I would like to return all unused memory to the allocation semispace (after flipping semispaces at the end of each nursery collection, using `MM_MemoryPoolAddressOrderedList::recycleHeapChunk()`. 

You can see this effect in results reported below. Each data point is obtained from 5 runs of SPECjbb2005 on lnxem64tvm9-resv.ottawa.ibm.com (120 CPUs) with the number of active cores limited via numactl to match number of gc threads. 

The stats below are derived from internal metrics reported by scavenger and evacuator at the end of each nursery collection in evacuator builds. The "% inside" statistic in the throughput row for the evacuator runs is the percentage of the total evacuated volume that is copied hierarchically inside stack frames (close copying) as opposed to copied breadth first to outside copyspaces. For scavenger this statistic is a close approximation to the percentage of evacuated volume that is copied within aliased caches. For the evacuator runs fragmentation is split between "micro" (incurred before end of cycle) and "macro" (unused memory at end of cycle). This distinction is not reported for scavenger at present and all fragmentation is charged to the "macro" statistic. 

```
15 gc threads ---

evacuator:
        collections: 10 global; 0 backouts; 109500 generational
        idle time: 1.158% stall; 3.435% complete; .417% sync; 2017823.353 gc-ms
        throughput: 3182745660696 allocated; 2994948051280 evacuated; 2373403459976 (79.246%) inside; 2017823.353 ms; 1449.459 kb/ms
        fragmentation: 129712136 (.004%) micro; 187604020808 (5.894%) macro; 187797609416 (5.901%) global
        containment: 15.263% <64; 20.528% <128; 27.196% <256; 97.431% small (<= 256 bytes)
scavenger:
        collections: 11 global; 0 backouts; 110734 generational (0.000% of runtime)
        idle time: 2.137% stall; 2.819% complete; .358% sync; 1995505.745 gc-ms
        throughput: 3044756815360 allocated; 3025614924312 evacuated; 611470076072 (20.209%) inside; 1995505.745 ms; 1480.678 kb/ms
        fragmentation: 0 (0%) micro; 20166561024 (.662%) macro; 19141891048 (.629%) global
        containment: 7.525% <64; 11.074% <128; 17.139% <256; 97.406% small (<= 256 bytes)
		
30 gc threads ---
		
evacuator:
        collections: 10 global; 0 backouts; 62124 generational
        idle time: 1.537% stall; 4.613% complete; .715% sync; 1467529.937 gc-ms
        throughput: 1902317111392 allocated; 1686467790776 evacuated; 1109529507384 (65.790%) inside; 1467529.937 ms; 1122.253 kb/ms
        fragmentation: 65739456 (.003%) micro; 215736398304 (11.340%) macro; 215849320616 (11.347%) global
        containment: 14.182% <64; 18.779% <128; 24.338% <256; 97.372% small (<= 256 bytes)
scavenger:
        collections: 10 global; 0 backouts; 62828 generational (0.000% of runtime)
        idle time: 2.601% stall; 4.176% complete; .538% sync; 1427204.082 gc-ms
        throughput: 1716732410880 allocated; 1700809190640 evacuated; 303194139960 (17.826%) inside; 1427204.082 ms; 1163.776 kb/ms
        fragmentation: 0 (0%) micro; 17413316232 (1.014%) macro; 15923220240 (.928%) global
        containment: 8.121% <64; 11.647% <128; 17.378% <256; 97.451% small (<= 256 bytes)
```

The stats in the table below are derived from values scraped from the respective verbosegc logs (T = #gc threads).

|algorithm|T|gc-count|average gc-ms|%cpu|kb/gc-ms|total gc-ms|total user-ms|total system-ms|
|--------:|-:|-----:|-----:|-----:|---:|------:|-------:|-----:|
|evacuator|15|109500|18.427|99.562|1450|2017764|28068599|123206|
|scavenger|15|110734|18.020|99.530|1481|1995451|27635373|130471|
|evacuator|30|62124|23.622|99.411|1122|1467498|39288834|232404|
|scavenger|30|62828|22.715|99.375|1164|1427168|38074601|239181|

ktbriggs:
@charliegracie in response to your questions above:

_Can you provide more details about the current set of improvements?_

Evacuator development is continuing on another branch (https://github.com/ktbriggs/omr/tree/evacuator-bench). The branch referenced in this pull request (https://github.com/ktbriggs/omr/tree/evacuator-bench) has been stable since June 18 but subject to rebasing from eclpise/master. The last rebase was June 26. Since June 18 work on the evacuator-bench branch includes:

- try to restore stack limit after each outside copy in flush instead of continuing to flush current frame
  -- flushing is triggered by presence of >0 stalled threads
- recycle survivor unused whitespace allocations into memory pool at end of each gc cycle
  -- this takes care of problematic large whitespace remainders in stack, copyspaces at end of each cycle
- changes to optimal white/work size calculations
  -- whitespace allocation sizes are now almost always maximal (128kb)
  -- workspace sizes are determined dynamically according to volume of work on evacuator worklist
- reduce default min/max work packet sizes to 50% of scavenger defaults
  -- this reduces stall times in evacuator
  -- use -XXgc scanCacheMin/MaxSize options to override default work packet size
- add breathFirstScanning to evacuator flags, ensure breadth-first scanning is supportable
  -- breadth-first scanning occurs when stack limit is set just above bottom stack frame
- add MM_ScavengerStats::countWorkPacketSize() and supporting fields
  -- produce work packet size histogram at end of each gc 
- include array header/padding in first array segment (set up in push())
  -- minor change to ensure that aggregate copied/scanned byte counts are current
- reduce volumes of work transferred from donors in findWork()
  -- receiver stops pulling before it receives more than is remaining for donor
- non-indexable object scanners preselect a hot field as first slot to return
  -- hot object referenced by preselected slot is then more likely to be copied close to parent

_How do the new changes compare?_

I ran 5 trials of gencon against each of 3 benchmarks selected for the proportions of small (<=256 bytes) objects in their workloads:

- transactional workload (~97% small objects)
- data compilation workload (~75% small objects)
- database processing workload (~50% small objects)

Each benchmark was tested with 3 gencon variants: 

- (master) scavenger built from omr and openj9 master branches (ie, without any of the changes included in this PR) 
- (scavenger) gencon + -Xgc hierarchicalScanOrdering built from omr and openj9 evacuator-bench branches
- (evacuator) gencon + -Xgc recursiveScanOrdering built from omr and openj9 evacuator-bench branches

The master variant is from a OpenJ9 build from the eclipse openj9 and omr master branches. It does not include any evacuator code and runs the production gencon algorithm. The scavenger and evacuator variants are both incorportated within the same SDK, This SDK was built from the evacuator-bench branches in my clones of eclipse omr and openj9, which were rebased on the corresponding master branches.

Results scraped from corresponding verbose gc logs are shown in the table below. The `gc-ms` statistic is the average gc time, `user-ms, duration-ms, system-ms` are totals, `gc-%cpu` is calculated as `user-ms/(user-ms + system-ms)`, and `gc-kb/ms` is average gc throughput. 

|algorithm|benchmark|gc-count|gc-ms|gc-%cpu|gc-kb/ms|user-ms|duration-ms|system-ms|
|--------:|--------:|---:|-------:|----:|------:|-------:|------:|--------:|
|master|transactional|17259|10.865|98.462|1482|187525|1163772|18176|
|scavenger|transactional|17969|11.268|98.517|1424|202483|1301032|19580|
|evacuator|transactional|17509|9.505|98.310|1697|166436|1089624|18728|
|scavenger|data compilation|8420|29.549|98.736|1640|248807|1572616|20124|
|evacuator|data compilation|8790|24.748|98.641|1962|217543|1426748|19648|
|scavenger|database processing|12958|2.607|89.645|566|33782|112024|12940|
|evacuator|database processing|12821|3.972|94.719|1106|50930|265484|14800|

For the SDK built from the evacuator-bench branch additional summary statistics are available for each nursery collection, including total bytes evacuated and proportions of reference arcs with 64-, 128-, and 256-byte cacheline containment. These internal metrics also provide a more detailed view of parallelism within nursery collections allowing stall (% of time spent waiting for work), end scan (% of time spent wating for other gc threads to complete heap scan), and sync (%time waiting for other gc threads to join at a synchronization point) percentages to be calculated. These are shown in the table below.

|algorithm|benchmark|evacuated-mb|%stall|%endscan|%sync|%<64|%<128|%<256|
|--------:|--------:|-----------:|-----:|-------:|----:|---:|----:|----:|
|scavenger|transactional|281620|5.387|3.209|0.766|6.370|9.521|15.031|
|evacuator|transactional|275883|3.999|2.881|0.735|13.234|18.226|25.224|
|scavenger|data compilation|398383|6.721|7.222|0.367|2.969|5.875|11.297|
|evacuator|data compilation|416910|7.718|4.597|0.317|7.619|15.626|25.488|
|scavenger|database processing|18668|18.598|15.319|9.089|7.652|8.914|9.867|
|evacuator|database processing|55010|7.242|8.108|5.163|26.417|29.616|31.473|

Cacheline containment and evacuation throughput were the main metrics that the evacuator algorithm is designed to improve. Substantial improvement in these metrics is evident in the results reported above. However, I cannot report any consistent and significant improvements in application performance. Please note with regard to the database processing results in the tables above that the evacuator runs evacuated almost 3X the amount evacuated in the scavenger runs with the evacuator-bench SDK.

I am in the process of completing the runs for master SDK with the data compilation and database processing benchmarks and will report results here when available.

charliegracie:
hmmm interesting results so far. I am concerned that there is still a lot of variability in the runs. It may be useful to disable dynamic tenure age adjustment, set tenure age to a specific value of say 1 which would help stabilize the runs. Would it be possible to add the average tenure age for each benchmark to your results above? If there is a difference we may want to re-run with the fixed tenure age to ensure apples to apples comparisons.  Depending on the values in the verbose:gc logs other options may help provide more consistent results as well. 

For the results a relative value of the application performance (using master as the 1.0 value) with a listing of higher or lower is better for each of the 3 benchmarks would be beneficial to fully understanding the results.  Also in the cache line containment I would be very interested in seeing this divided between the nursery and tenure space.  Since objects rarely move once they make it to tenure space cache line containment is very important to application performance.

ktbriggs:
Tenure age histograms from above runs, as requested. The 1st pair of (scavenger-evacuator) is for the transactional workload, 2nd is for data compilation, 3rd  is for database processing. 

|age|scavenger|evacuator|scavenger|evacuator|scavenger|evacuator|
|--:|--------:|--------:|--------:|--------:|--------:|--------:|
|2|5|5|6291|6555|78|79|
|3|26|23|1606|1796|125|135|
|4|68|74|117|75|403|405|
|5|48|44|52|53|606|673|
|6|7|11|25|37|373|348|
|7|2|2|31|26|200|222|
|8|0|0|28|13|20|47|
|9|0|0|14|8|4|4|
|10|10|10|5|10|47|47|
|11|0|1|27|12|34|21|
|12|0|0|38|44|52|38|
|13|0|2|65|52|23|29|
|14|17803|17337|121|109|10993|10773|

ktbriggs:
There was a bug in evacuator code that was preventing nursery objects from reaching tenure age. This caused surges in remembered set cardinality and corresponding bloat in evacuation volume for the evacuator runs with the database/data compilation/trnsactional workloads reported above. Ran same 3 2x5 trials again on same 8-core x86-64/Ubuntu1604 vm with 1g heap (Xms=Xmx). 

image

The _gc-count_, _gc-time_ and _evacuated-mb_ values are totals over all 5 runs for each workload. The evacuated volume in kbytes per gc thread per gc is expressed in the _kb/gc*threads_ value. The _stall%_ values reflect the percentage of available CPU bandwidth that was not consumed by evacuator threads. The _small%_ values represent the proportion of evacuated objects that were <=256 bytes in total size (for evacuator, these are eligible for copying inside stack frames and may so realize cache-line containment with referring parent and referent children). The _inside%_ values (evacuator only) represent the proportions of evacuated objects that were copied inside the stack. The cache-line containment (_contained-Nb_) are percentages of total number of evacuated objects that were copied inside the stack with referring pointer and head of referent copy collocated within the came N-byte cache line. The _fragmentation%_ values represent the proportion of survivor and tenure memory reserved but not used in nursery collections. For evacuator, this does not include survivor fragments that are successfully recycled back into the survivor subspace memory pool at t he end of each nursery collection or tenure fragments that are retained between collections for reuse in the next nursery collection (tenure fragments flushed before global collections would be included in the reported percentages, but no global collections occurred at any point after the first nursery collection in any of the trial runs).

With 8 threads the volume per thread is almost always too low in the database workload to feed 8 gc threads and  gc throughput is suboptimal for both scavenger and evacuator. The relatively low prevalence of small objects (eligible for recursive copying inside the stack) with the database workload reduces the frequency of inside copying, as does increased pressure from stalled evacuators. Running evacuator threads clamp the stack limit in the presence of stalled evacuator threads to force breadth-first flushing of all objects to outside copyspaces and produce distributable work, and scavenger inhibits aliasing in stalling contexts to force breadth-first copying while other threads are stalled. In spite of this, cache line containment was most prevalent for both scavenger and evacuator for the database workload runs. Performance for this workload could likely be improved by _reducing_ the number of gc threads until the stalling rate drops to <5%. (Sidebar: It is interesting to think of how this could be accomplished dynamically by maintaining a histogram with a bin for each gc thread and having each gc thread frequently sample the stalled thread count and record the number of times n threads were stalled for n<gc thread count.)

Evacuator demonstrated better performance with respect to the two main outcomes that it is designed to improve (evacuation throughput and cache-line containment) with the data compilation and transactional workloads. Throughput for evacuator was higher by about 12% and 21%, respectively, and cache-line containment was improved by 2-3X relative to scavenger. For these workloads there is an ample evacuation volume to ensure that gc threads receive enough scan work to keep busy copying maximally within their local context (aliased or within-stack operation) with low-level interruption from stalled threads. 

With evacuator, copying inside the stack increases the likelihood of cache-line containment, which significantly improves gc throughput as more copying is done with referring object and referent copy partially overlapping within the same cache-line. With scavenger, aliased copying sometimes achieves the same result but this is less effective because the distance between aliased cache scan and copy heads is often constrained only by the bounds of the cache. There were no significant differences between application (workload) performance between scavenger and evacuator for any of the workloads (average differences between scavenger and evacuator workload metrics were <-1%, 4% and <1% in favor of evacuator for the database, data compilation and transactional workloads, respectively.

ktbriggs:
Closing this as it has grown stale (since May 17 2018). Evacuator development has continued on another branch with a goal of improving performance and stabilizing the build. This branch is currently under test and will be published in my private repos when testing is complete:

I will publish snapshot of the current development branches, with urls for the omr and openj9 branches, when testing is complete. Please refer to the evacuator epic #2664 for this and subsequent developments, including benchmarking results, relating to evacuator.  

As always I encourage reviews of evacuator code and welcome comments. In order to keep feedback collocated please publish any and all comments to the evacuator discussion #2361, In particular, in the absence of a current PR, please do not hesitate to add comments relating to reviewed code to #2361.

I will open a new pull request when epic development yields significant and reproducible improvements in _application_ throughput (as shown above, evacuator produces consistent and significant improvement in internal GC metrics, but to date only a slight trend toward improved application throughput).

PR #2559, opened 2018-05-18
Avoid breaking CFG in redundant goto elimination

jdmpapin:
OMR::Block::changeBranchDestination() deletes the outgoing edge from the
control flow graph (CFG) when the branch already targets newDestination.
This means that an unnecessary call to changeBranchDestination() can
corrupt the CFG.

Redundant goto elimination can make such calls since eb94ed52.

Now changeBranchDestination() will safely ignore these calls, and
redundant goto elimination will avoid making them.

### DIFF
--- a/compiler/il/OMRBlock.cpp
+++ b/compiler/il/OMRBlock.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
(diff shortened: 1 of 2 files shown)


andrewcraik:
Seems a sensible change - @vijaysun-omr can you have a look?

fjeremic:
@genie-omr build all

Waiting on review by @vijaysun-omr before merging. Meanwhile the builds can be launched.

vijaysun-omr:
Looks safe to me.

PR #2560, opened 2018-05-18
Modify calculateOptimumCopyScanCacheSize

charliegracie:
CalculateOptimumCopyScanCacheSize can be broken into 2 different
checks. If there are threads waiting for work the GC should use
smaller copyCache to share work more aggressively. Also if the
work queue is empty the GC should also use smaller queues to share
work more aggressively.

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -829,32 +829,61 @@ MM_Scavenger::calcGCStats(MM_EnvironmentStandard *env)
  ****************************************
  */
 
+uintptr_t
+MM_Scavenger::calculateCopyScanCacheSizeForWaitingThreads(uintptr_t maxCacheSize, uintptr_t threadCount, uintptr_t waitingThreads)
+{
+	uintptr_t minCacheSize = _extensions->scavengerScanCacheMinimumSize;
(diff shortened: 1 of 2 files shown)


charliegracie:
FYI @amicic and @youngar

amicic:
The queue length criteria in some (contrived) cases might bring down the size too low: for example N threads scanning exactly N long linked lists (each thread one) will make N happy (fully parallel) threads (even with max cache size), but the queue size will be brought down to min, which might be unnecessary. In practice, it's probably not a big problem.

But I do think that existing ratio  of scanned/copied slots is not very relevant metric. It does nor really indicate production/consumption ratio, what it originally had been meant for.

Overall, this looks reasonable.

charliegracie:
I think you are correct about this change down to minimum + 1 step size in the worst case but that would just be for the first cache... If all threads do have work after that there will be N caches on the list so all threads will then use full a full cache size.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

PR #2562, opened 2018-05-18
fix warning in X86Debug.cpp due to missing return statements

mstoodle:
Repeated switch statements where default case does not return a value.
Generates lots of warnings, but probably has zero functional impact to
anyone anywhere.


fjeremic:
https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines:

> The first line [...], and should be written in _sentence case_ preferably not ending in a period.

fjeremic:
@genie-omr build xlinux

PR #2564, opened 2018-05-18
Killing default Memory Space

amicic:
Move the destruction of default Memory Space from kill() to teardown()
of base Configuration class.

Firstly, kill() should not be doing anything more than just calling the
destructor (teardown() method), and destroying the backing memory
(free() method).

Secondly, and more imporantly, doing it in kill() before teardown, would
do it too early, before specific teardown of Configuration subclasses
would be not be complete. For example
ConfigurationGenerational::tearDown would want to set scavenger pointer
in extensions to NULL before Scavenger instance is destroyed (inderectly
by MemorySpace and SubSpaceSemiSpace)  

### DIFF
--- a/gc/base/Configuration.cpp
+++ b/gc/base/Configuration.cpp
@@ -53,17 +53,6 @@
 void
 MM_Configuration::kill(MM_EnvironmentBase* env)
 {
-	MM_GCExtensionsBase *ext = env->getExtensions();
-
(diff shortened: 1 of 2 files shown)


amicic:
For the benefit of OpenJ9 project, this change should not be merged before: https://github.com/eclipse/openj9/pull/1948

PR #2565, opened 2018-05-19
Fix memory fences on x86

gacholio:
- dword rather than qword is sufficient for readWriteBarrier
- no memory fence is required for readBarrier
- remove lfence/sfence TODOs (not required for normal reads/writes)


gacholio:
Spoke with @0dvictor and confirmed that any locked instruction is sufficient for a memory barrier, and since the JIT and the Windows __faststorefence primitive also use dword, we may we as well line up.

LFENCE/SFENCE are only required for ordering of non-temporal SSE instructions.

include_core/AtomicSupport.hpp
```diff
@@ -259,12 +259,7 @@ class VM_AtomicSupport
 #elif defined(_MSC_VER)
 		_ReadWriteBarrier();
 #elif defined(__GNUC__)
-#if defined(J9X86) || defined(J9HAMMER)
-		asm volatile("":::"memory");
-		/* TODO investigate whether or not we should call this
-		asm volatile("sfence" ::: "memory");
-		*/
-#elif defined(ARM) /* defined(J9X86) || defined(J9HAMMER) */
+#if defined(ARM)
```
DanHeidinga: line 268 needs an update `#endif /* defined(J9X86) || defined(J9HAMMER) */` to fix the comment

include_core/AtomicSupport.hpp
```diff
@@ -292,24 +287,8 @@ class VM_AtomicSupport
...
-		/* TODO investigate whether or not we should call this instead
-		asm volatile("lfence":::"memory");
-		*/
-#elif defined(J9HAMMER)
-		asm volatile("lock orq $0x0,(%%rsp)" ::: "memory");
-		/* TODO investigate whether or not we should call this instead
-		asm volatile("lfence":::"memory");
-		*/
-#elif defined(ARM) /* defined(J9HAMMER) */
+#if defined(ARM)
```
DanHeidinga: The comment on line 297 needs to be updated as well

charliegracie:
@genie-omr build all

DanHeidinga:
> Did I read this correctly?

I had miss'd the last case was an `#else`, not another `#elif`

include_core/AtomicSupport.hpp
gacholio: Comments addressed.

gacholio:
Accidentally merged other changes - closing and starting again,

gacholio:
New PR: https://github.com/eclipse/omr/pull/2572

PR #2568, opened 2018-05-21
Add .gitattributes file to allow correct format of files

groeges:
when they are download on z/OS system by Rocket Software's
Git program. These changes are needed when building OMR
within OpenJDK with OpenJ9

Signed-off-by: Steve Groeger <groeges@uk.ibm.com>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+# This .gitattributes file will cause all text files EXCEPT for
+# git's .gitattributes and .gitignore files to be encoded as EBCDIC.
+# Selected binary files will not be translated at all.
+# The default for text files
+* git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+
+# git's files (which MUST be ASCII)
+.gitattributes git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+.gitignore git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+*.hdf git-encoding=iso8859-1 working-tree-encoding=ibm-1047
+*.tdf git-encoding=iso8859-1 working-tree-encoding=ibm-1047
+
+# Binary files
+*.jpg git-encoding=BINARY working-tree-encoding=BINARY
+*.png git-encoding=BINARY working-tree-encoding=BINARY
+*.gif git-encoding=BINARY working-tree-encoding=BINARY
+*.zip git-encoding=BINARY working-tree-encoding=BINARY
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,17 @@
+# This .gitattributes file will cause all text files EXCEPT for
+# git's .gitattributes and .gitignore files to be encoded as EBCDIC.
+# Selected binary files will not be translated at all.
+# The default for text files
+* git-encoding=iso8859-1 working-tree-encoding=iso8859-1
```
charliegracie: I am not that familiar with `.gitattributes` but the comments are not correct.  The comments about EBCDIC are clearly only true on zOS.  Is this going to be an issue on platforms other than zOS?

charliegracie:
@genie-omr build all

.gitattributes
groeges: @charliegracie No, the extra git-encoding in .gitattributes is only used when using the version of git that is available for z/OS from Rocket Software. All other platforms will ignore these options.

groeges: Have modified the comments in this file to make it clear that this only effects z/OS platform.

charliegracie:
@genie-omr build all

groeges:
@charliegracie Looking at the build that is failing I think it may have something to do with the .tdf files being converted twice. With the .gitattributes file, the hdf and tdf files are marked to be downloaded as EBCDIC, however in your build process you also convert them to ENCDIC, hence they are converted twice.  With the change in adding the .gitattributes your build should not need to convert the files to EBCDIC.  This is only my thoughts on what the issue with the build might be and am happy to be proved wrong.

charliegracie:
It looks like you are correct.  Give me a day or so to see if I can figure out a way to check to see if they were converted via the `.gitattributes` and not convert them again.

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

charliegracie:
@genie-omr build zos

groeges:
@charliegracie I see the zos_390-64 build succeeded. Does this mean that you found a way of checking to see if the files were already converted and to then not convert. If so, does this mean the PR is now OK and able to be merged.

charliegracie:
Yes I was just waiting for a PR to finish that does not have your change to make sure it didn't break :).  Now that [Job 199](https://ci.eclipse.org/omr/job/PullRequest-zos_390-64/199/) passed successfully this is ready to be merged :)

PR #2572, opened 2018-05-22
Fix memory fences on x86

gacholio:
- dword rather than qword is sufficient for readWriteBarrier
- no memory fence is required for readBarrier
- remove lfence/sfence TODOs (not required for normal reads/writes)


gacholio:
Spoke with @0dvictor and confirmed that any locked instruction is sufficient for a memory barrier, and since the JIT and the Windows __faststorefence primitive also use dword, we may we as well line up.

LFENCE/SFENCE are only required for ordering of non-temporal SSE instructions.

charliegracie:
@genie-omr build all

PR #2574, opened 2018-05-23
Clang: disable unused private field and function

samolisov:
There are many warnings `unused-private-field` about the
`OMRNode::NodeExtensionStore::_flag` appear during compilation
with clang as well as some false positive warnings `unused-function`
about the `isJ9()` function (the function is used!) do. To suppress the
warnings, the `-no-unused-private-field -no-unused-function` command
line was added to clang in the MSVC toolchain configuration.


fjeremic:
@genie-omr build all

PR #2576, opened 2018-05-23
Fix more warnings and errors

youngar:
Mostly fixing usages of `strncpy`, and some improper initialization of class instances.

### DIFF
--- a/example/glue/VerboseManagerImpl.cpp
+++ b/example/glue/VerboseManagerImpl.cpp
@@ -61,22 +61,21 @@ bool
 MM_VerboseManagerImpl::configureVerboseGC(OMR_VM *omrVM, char *filename, uintptr_t fileCount, uintptr_t iterations)
 {
 	OMRPORT_ACCESS_FROM_OMRVM(omrVM);
-	if (MM_VerboseManager::configureVerboseGC(omrVM, filename, fileCount, iterations)) {
-		this->fileCount = fileCount;
-		this->iterations = iterations;
-		size_t len = strlen(filename);
-		this->filename = (char *)omrmem_allocate_memory(len+1, OMRMEM_CATEGORY_MM);
-		strncpy(this->filename, filename, len);
-		this->filename[len] = '\0';
-		return true;
-		if (NULL != this->filename) {
-			return false;
-			strncpy(this->filename, filename, len);
-			this->filename[len] = '\0';
-			return true;
-		}
+	if (!MM_VerboseManager::configureVerboseGC(omrVM, filename, fileCount, iterations)) {
+		return false;
 	}
-	return false;
+	this->fileCount = fileCount;
+	this->iterations = iterations;
+	size_t len = strlen(filename);
+
+	this->filename = (char *)omrmem_allocate_memory(len+1, OMRMEM_CATEGORY_MM);
+	if (NULL == this->filename) {
+		return false;
+	}
+
+	strcpy(this->filename, filename);
+
+	return true;
 }
 
 bool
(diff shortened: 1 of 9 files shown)


youngar:
@rwy0717 can you review this please?

youngar:
@genie-omr build all

example/glue/VerboseManagerImpl.cpp
```diff
@@ -61,22 +61,21 @@ bool
...
+	this->iterations = iterations;
+	size_t len = strlen(filename);
+
+	this->filename = (char *)omrmem_allocate_memory(len+1, OMRMEM_CATEGORY_MM);
+	if (NULL == this->filename) {
+		return false;
 	}
-	return false;
+
+	strcpy(this->filename, filename);
```
charliegracie: can we use strncpy since we do have the length?  It will stop some of the static analysis tooling from complaining. If you think it really does not make sense we can leave it as is.

example/glue/VerboseManagerImpl.cpp
```diff
@@ -61,22 +61,21 @@ bool
...
 	OMRPORT_ACCESS_FROM_OMRVM(omrVM);
-	if (MM_VerboseManager::configureVerboseGC(omrVM, filename, fileCount, iterations)) {
-		this->fileCount = fileCount;
-		this->iterations = iterations;
-		size_t len = strlen(filename);
-		this->filename = (char *)omrmem_allocate_memory(len+1, OMRMEM_CATEGORY_MM);
-		strncpy(this->filename, filename, len);
-		this->filename[len] = '\0';
-		return true;
-		if (NULL != this->filename) {
```
charliegracie: thank you for removing this dead code

tools/tracegen/TDFParser.cpp
```diff
@@ -135,7 +135,7 @@ TDFParser::parse()
 			goto failed;
 		}
 
-		strncpy(tdf->header.datfilename, DEFAULT_DAT_FILE, strlen(DEFAULT_DAT_FILE));
+		strcpy(tdf->header.datfilename, DEFAULT_DAT_FILE);
```
charliegracie: were newer versions of GCC complaining about this?

thread/common/omrthreadmem.cpp
```diff
@@ -58,11 +59,11 @@ typedef struct J9ThreadMemoryHeader {
 void
 omrthread_mem_init(omrthread_library_t threadLibrary)
 {
-	memcpy(&threadLibrary->threadLibraryCategory, &threadCategoryTemplate, sizeof(OMRMemCategory));
-	memcpy(&threadLibrary->nativeStackCategory, &nativeStackCategoryTemplate, sizeof(OMRMemCategory));
+	new (&threadLibrary->threadLibraryCategory) OMRMemCategory(threadCategoryTemplate);
```
charliegracie: This doing a new using the copy constructor?

example/glue/VerboseManagerImpl.cpp
youngar: GCC is complaining about this in my version.  The complaint in question is that we are using the length of the source buffer as an argument meant to be the length of the destination buffer.    Although `len` is technically the correct length of the destination, GCC can tell we are using the result of `strlen(filename)`.  The current advice for this issue is to just use `strcpy`, since it is guaranteed to work in this situation.

It seems with these new warnings that GCC is trying to look a little too hard into how these functions were meant to be used.  Our usage is completely valid, but `strncpy` was originally meant to copy non-null-terminated strings.  I have run into other warnings where we use `strncpy` to make sure that we don't copy the `\0` off the end of a source string, and it produces a warning that complains we are truncating the source string.  Instead, I have switched instances of that to use `strcat`.

tools/tracegen/TDFParser.cpp
youngar: Yeah

thread/common/omrthreadmem.cpp
youngar: Yeah, placement new with a copy constructor

example/glue/VerboseManagerImpl.cpp
youngar: :+1:

Issue #2491, opened 2018-04-20
Remove PseudoRealRegisters from RealRegisterEnum.hpp


dchopra001:
Inside the z/codegen's RealRegisterEnum.hpp, we have some dependencies/PseudoRealRegisters (not sure which one is the officially accepted name here) defined inside RealRegisterEnum.hpp:
https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/RealRegisterEnum.hpp#L159-L174

These dependencies are not backed by real hardware, but instead serve as hints to the register allocator. For example, `TR::RealRegister::LegalEvenOfPair` does not serve a similar purpose as `TR::RealRegister::GPR0`. 

We should separate these dependency definitions from real registers. This enum is indirectly used in certain places to size lists that represent Hardware backed registers. This results in confusion as to what the enum represents (is it a list of dependencies or hardware backed registers or both?), and potential bugs. For example, `_registerFile` is defined below to have size 81 (i.e. `TR::RealRegister::NumRegisters` maps to 81):

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRMachine.hpp#L184-L190

And when initialized, 80 real registers and a NoReg value is initialized:

https://github.com/eclipse/omr/blob/de70faf87f67ffe9cd9571d30e5a378009b85b7c/compiler/z/codegen/OMRMachine.cpp#L6122-L6132

However, as can be seen above, we also do the following: `_registerFile[TR::RealRegister::SpilledReg] = NULL;` This line is probably important to the RA, but is also undefined behaviour because `SpilledReg` maps to index 91 ( and _registerFile's size is 81). More context as to why this doesn't cause a crash here:[1]

We need to separate these dependencies from real register enumerations because they are clearly being used in different contexts. While this may not currently create bugs, it does create confusion and  can cause unwarranted bugs in the future.

For more context on this problem, please refer to the discussion in [1] when I initially tried to remove the `_registerFile[TR::RealRegister::SpilledReg] = NULL;` above:

[1] https://github.com/eclipse/omr/pull/2462#discussion_r182447301

dchopra001:
Current proposed implementation to address this issue is here:
https://github.com/eclipse/omr/pull/2462#discussion_r182606476

I'm working on trying this out currently.

0xdaryl:
Good writeup @dchopra001 !

PR #2577, opened 2018-05-23
Define PseudoRealRegisters using new RegDep enum

dchopra001:
Certain registers defined inside RealRegisterEnum.hpp are not hardware backed registers
(ex. TR::RealRegister::NumRegisters), but instead are used as hints for the code generator or as dependencies for real registers. This is quite prevalent in the z/codegen.This commit separates these pseudo registers from the real register enum, and updates the APIs where needed. TR::RealRegister::MISC is also deprecated because it represents the same thing as TR::RealRegister::LastHPR.

Closes: #2491

### DIFF
--- a/compiler/aarch64/codegen/PseudoRegisterEnum.hpp
+++ b/compiler/aarch64/codegen/PseudoRegisterEnum.hpp
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+/*
+ * This file will be included within an enum.  Only comments and enumerator
+ * definitions are permitted.
+ */
+
--- a/compiler/codegen/OMRRealRegister.hpp
+++ b/compiler/codegen/OMRRealRegister.hpp
@@ -58,6 +58,7 @@ class OMR_EXTENSIBLE RealRegister : public TR::Register
       Locked    = 4
       } RegState;
 
+   // All hardware backed registers
    typedef enum
       {
       #include "codegen/RealRegisterEnum.hpp"
@@ -69,6 +70,22 @@ class OMR_EXTENSIBLE RealRegister : public TR::Register
       #include "codegen/RealRegisterMaskEnum.hpp"
 
       } RegMask;
+
+    // PseudoRegisters are the union of all hardware backed registers (ex. GPR0)
+    // and constructs such as TR::RealRegister::NoReg or TR::RealRegister::AssignAny. 
+    // These pseudo registers are used to assign register dependencies during the 
+    // Register Allocation phase. The enum below is meant to hold all pseudo registers.
+    // However, to avoid compilation failures, we cannot include 
+    // codegen/RealRegisterEnum.hpp again. In order to include it twice, these two enums
+    // must be scoped. However, not all of our minimum compiler toolchains support this
+    // feature yet (MSVC 2010). So we only include PseudoRegisterEnum below, and cast any 
+    // RegNum to RegDep when using it to set a register dependency. If MSVC 2010 support is 
+    // no longer required, then this can be cleaned up by using scoped enums instead. An issue
+    // to track this is open here: https://github.com/eclipse/omr/issues/2590 
+    typedef enum
+       {
+       #include "codegen/PseudoRegisterEnum.hpp"
+       } RegDep;
 
 
    protected:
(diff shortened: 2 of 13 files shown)


dchopra001:
@fjeremic Review request

dchopra001:
@fjeremic I left a couple of things out to first get your thoughts on them before proceeding.

1. Should TR::RealRegister::NoReg be a pseudo register or a real register? Right now it's defined as a real register but I'm leaning towards it being a pseudoRegister.

2. This commit still doesn't fix the problem discussed in #2491 where `_registerFile[TR::RealRegister::SpilledReg]` is an array out of bounds check. If we are to keep this check in place, then here's a proposed solution:

We define a new value, NumPseudoRegisters, inside PseudoRegisterEnum.hpp to be the summation of all hardware backed registers (i.e. TR::RealRegister::NumRegisters) and all non-hardware backed registers (ex. TR::RealRegister::AssignAny). In the z/codegen, this value will be around 94-95. We then redefine `_registerFile[TR::RealRegister::NumRegisters]` to `_registerFile[TR::RealRegister::NumPseudoRegisters]`, and then initialize the non-hardware backed registers to `NULL.` 
While we're at it, we can rename `TR::RealRegister::NumRegisters` to `TR::RealRegister::NumRealRegisters`.

Throughout the codebase, if we ever iterate through `_registerFile` we use an upper bound of `TR::RealRegister::NumRegisters` or constructs such as `TR::RealRegister::LastAssignableGPR`. So redefining the size of the array shouldn't cause any problems.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -1048,20 +1048,20 @@ lDivRemGenericEvaluator(TR::Node * node, TR::CodeGenerator * cg, bool isDivision
...
-      dependencies->addPostCondition(tempReg6, TR::RealRegister::GPR11);
+      dependencies->addPostCondition(oddRegister, (TR::RealRegister::RegDep)TR::RealRegister::GPR3);
+      dependencies->addPostCondition(tempReg5, (TR::RealRegister::RegDep)TR::RealRegister::GPR10);
+      dependencies->addPostCondition(tempReg6, (TR::RealRegister::RegDep)TR::RealRegister::GPR11);
 
       // Setup regs for call out
-      dependencies->addPostCondition(EPReg, cg->getEntryPointRegister());
-      dependencies->addPostCondition(RAReg, cg->getReturnAddressRegister());
+      dependencies->addPostCondition(EPReg, (TR::RealRegister::RegDep)cg->getEntryPointRegister());
+      dependencies->addPostCondition(RAReg, (TR::RealRegister::RegDep)cg->getReturnAddressRegister());
```
fjeremic: I don't think delegating these casts to the callee is a good idea. Do you agree that a much less invasive change would be to define two overloads to `addPostCondition` (and similar APIs) and hide these casts under the hood. Once we get scoped enum support all of this can be cleaned up.

dchopra001: Yup, that sounds a lot cleaner. I've backed out the casts and introduced the new APIs in 
42a4102 to 6da2dce

compiler/aarch64/codegen/PseudoRegisterEnum.hpp
```diff
@@ -0,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corp. and others
```
fjeremic: I think we need to update these copyright dates to "2018, 2018" as this is a completely new file. Similarly elsewhere.

compiler/codegen/OMRRealRegister.hpp
```diff
@@ -70,6 +71,21 @@ class OMR_EXTENSIBLE RealRegister : public TR::Register
...
+    // PseudoRegisters are the union of all hardware backed registers (ex. GPR0)
+    // and constructs such as TR::RealRegister::NoReg or TR::RealRegister::AssignAny. 
+    // These pseudo registers are used to assign register dependencies during the 
+    // Register Allocation phase. The enum below is meant to hold all pseudo registers.
+    // However, to avoid compilation failures, we cannot include 
+    // codegen/RealRegisterEnum.hpp again. In order to include it twice, these two enums
+    // must be scoped. However, not all of our minimum compiler toolchains support this
+    // feature yet (MSVC 2010). So we only include PseudoRegisterEnum below, and cast any 
+    // RegNum to RegDep when using it to set a register dependency. If MSVC 2010 support is 
+    // no longer required, then this can be cleaned up by using scoped enums instead.
```
fjeremic: Can we open up an item to track this issue and reference it in the comment?

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -10142,7 +10142,7 @@ OMR::Z::CodeGenerator::addVMThreadPreCondition(TR::RegisterDependencyConditions
       //TODO: using addPreConditionIfNotAlreadyInserted() is slower than using addPreCondition()
       //think about changing the name of this method to addVMThreadPreConditionIfNotAlreadyInserted()
       //and add one for addVMThreadPreCondition, which uses deps->addPreCondition()
-      deps->addPreConditionIfNotAlreadyInserted(reg,(TR::RealRegister::RegNum)self()->getVMThreadRegister()->getAssociation());
+      deps->addPreConditionIfNotAlreadyInserted(reg,(TR::RealRegister::RegDep)self()->getVMThreadRegister()->getAssociation());
```
fjeremic: Space after 'reg,' needed. Also we should avoid C style casts and use `static_cast` instead. Once we move to scoped enums you won't have to search for these types of casts as they'll be compile time errors (if using `static_cast` - not if using unsafe C style cast). Similarly elsewhere.

compiler/aarch64/codegen/PseudoRegisterEnum.hpp
dchopra001: Done: 3d92f53441d758ed3d79bdfa0e9e86f4d07968d9

compiler/codegen/OMRRealRegister.hpp
dchopra001: Done: 63095b3dcc67752c33319c4f88cde2b520816c16

Issue is here: https://github.com/eclipse/omr/issues/2590

compiler/z/codegen/OMRCodeGenerator.cpp
dchopra001: Done: a2e8dd2

fjeremic:
@dchopra001 let me know when this one is ready. Please remove the WIP: once you are done with the changes.

dchopra001:
@fjeremic Ready for review. I'll do a squash + rebase after the review.

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build all

PR #2578, opened 2018-05-24
Add portlib api sysinfo_cgroup_is_memlimit_set

ashu-mehra:
Added new API sysinfo_cgroup_is_memlimit_set that returns TRUE if
cgroup memory limit is set, FALSE otherwise

Signed-off-by: Ashutosh Mehra asmehra1@in.ibm.com

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -349,6 +349,7 @@ static int32_t readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, BOOLE
 static OMRCgroupSubsystem getCgroupSubsystemFromFlag(uint64_t subsystemFlag);
 static int32_t readCgroupSubsystemFile(struct OMRPortLibrary *portLibrary, uint64_t subsystemFlag, const char *fileName, int32_t numItemsToRead, const char *format, ...);
 static int32_t isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer);
+static int32_t getCgroupMemoryLimit(struct OMRPortLibrary *portLibrary, uint64_t *limit);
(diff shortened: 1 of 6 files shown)


ashu-mehra:
fyi - @charliegracie

charliegracie:
@genie-omr build all

port/unix/omrsysinfo.c
```diff
@@ -3750,6 +3751,40 @@ isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer)
...
+
+	Trc_PRT_sysinfo_cgroup_get_memlimit_Entry();
+
+	rc = readCgroupSubsystemFile(portLibrary, OMR_CGROUP_SUBSYSTEM_MEMORY, "memory.limit_in_bytes", numItemsToRead, "%lu", &cgroupMemLimit);
+	if (0 != rc) {
+		Trc_PRT_sysinfo_cgroup_get_memlimit_memory_limit_read_failed("memory.limit_in_bytes", rc);
+		goto _end;
+	}
+
+	physicalMemLimit = getPhysicalMemory(portLibrary);
```
charliegracie: Do we need to get this data and do the checks below to decide if CGroups are set?  Isn't the above code enough to know that CGroups are set?

ashu-mehra: Problem is if cgroup memory limit is not set, then the value in `memory.limit_in_bytes` file is a very large number (actually close to LONG_MAX value). Therefore, the need to explicitly check the value in file against available physical memory to know if the limit is set or not.

charliegracie: sigh... ok the cgroup stuff seems like such a hack on the container side... This change is ok.

PR #2580, opened 2018-05-24
Wrap omrtrace lib within build flags in OMR_Initialize_VM

nbhuiyan:
Also, the corresponding wrapping was done in OMR_ShutdownVM. Not
everything that wants to initialize/shutdown OMR components may
want to initialize/shutdown the health center monitoring tools.

Also fixed a potential bug where the port library's omrtty_printf
member was being used before the port library was initialized.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -138,24 +138,26 @@ test_targets += fvtest/porttest/aixbaddep
 endif
 endif
 
+ifeq (1, $(OMR_RAS_TDF_TRACE))
 # RAS Libraries
 main_targets += omrtrace
+# RAS Tests
+test_targets += fvtest/rastest
+endif
 
 # OMR Startup
 main_targets += omr omr/startup
 
-# RAS Tests
-test_targets += fvtest/rastest
-
 # OMR Example Targets
 ifeq (1,$(OMR_EXAMPLE))
 test_targets += example
+main_targets += omrtrace
 endif
 
 # VM Tests
 test_targets += fvtest/vmtest
 
-# List of targets that are needed for the test compil
+# List of targets that are needed for the test compiler and jitbuilder
 compiler_prereqs = port thread util/pool util/omrutil util/avl util/hashtable util/hookable
 
 # Test Compiler
@@ -181,6 +183,7 @@ DO_TEST_TARGET := no
 ifeq (yes,$(ENABLE_FVTEST_AGENT))
 DO_TEST_TARGET := yes
 test_targets := $(test_prereqs)
+main_targets += omrtrace
 test_targets += fvtest/rastest fvtest/util
 endif
 endif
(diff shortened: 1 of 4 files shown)


nbhuiyan:
@0xdaryl requesting review.

charliegracie:
@genie-omr build all

GNUmakefile
```diff
@@ -138,24 +138,26 @@ test_targets += fvtest/porttest/aixbaddep
...
 # OMR Startup
 main_targets += omr omr/startup
 
-# RAS Tests
-test_targets += fvtest/rastest
-
 # OMR Example Targets
 ifeq (1,$(OMR_EXAMPLE))
 test_targets += example
+main_targets += omrtrace
```
charliegracie: why is omrtrace getting adding multiple times?  It is here and above in the OMR_RAD_TDF_TRACE above.

GNUmakefile
```diff
@@ -181,6 +183,7 @@ DO_TEST_TARGET := no
 ifeq (yes,$(ENABLE_FVTEST_AGENT))
 DO_TEST_TARGET := yes
 test_targets := $(test_prereqs)
+main_targets += omrtrace
```
charliegracie: again why is this being added again?

omr/startup/omrvmstartup.cpp
```diff
@@ -325,7 +325,7 @@ OMR_Initialize_VM(OMR_VM **omrVMSlot, OMR_VMThread **omrVMThreadSlot, void *lang
 	OMR_VM *omrVM = NULL;
 	OMR_VMThread *vmThread = NULL;
 	OMRPORT_ACCESS_FROM_OMRPORT(&portLibrary);
-
+
```
charliegracie: Can we remove the tab here?

GNUmakefile
nbhuiyan: The `example` target links against omrtrace, and there exists no checks to ensure it has been enabled. This would prevent the errors that may arise when `OMR_RAS_TDF_TRACE` is disabled. Adding that simply helps with preventing build errors from being unable to find `omrtrace.a`. It may not be necessary to have this as a part of this PR, as it does not completely solve all the potential issues that may arise when omrtrace is disabled. There are more dependency-related issues that we could solve through assertions (through CMake only!) as well as more `#ifdef` guards across multiple components that use the trace engine.

GNUmakefile
nbhuiyan: Similar answer to the above. fvtests require omrtrace. The duplicated entries get removed anyway a few lines below.

omr/startup/omrvmstartup.cpp
nbhuiyan: That was an oversight. I will fix that as a part of any revisions I do in this PR.

nbhuiyan:
Thanks @charliegracie ! The commit has been amended.

PR #2581, opened 2018-05-24
ConcurrentGC STW phase detection for CS

amicic:
At a couple of spots Concurrent Scavenger needs to know if the
overlapping Concurrent GC is in pure concurrent mode or it transitioned
to the final STW phase.

Originally, the *mode* variable was used, but this appears to be not
quite correct: there is a small time hole when mode is
CONCURRENT_FINAL_COLLECTION, while exclusive access has not been
obtained yet, and there is still concurrent marking activity (it would
terminate as soon as the tread realizes the mode has changed).

We now rely on isStwCollectionInProgress() API, which will correctly
return true only if ConcurrentGC operatates while having exclusive
access.

All other changes are non-functional and just shuffling the code to
reduce amount of type casting, break mutual #include dependancy, and
generally make compiler happy.

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -52,7 +52,6 @@
 
 class MM_CardTable;
 class MM_ClassLoaderRememberedSet;
-class MM_Collector;
 class MM_CollectorLanguageInterface;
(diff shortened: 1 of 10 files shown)


charliegracie:
@genie-omr build all

PR #2584, opened 2018-05-24
WIP: Virtualized ELFRelocationResolver

samasri:
Signed-off-by: Samer AL Masri <almasri@ualberta.ca>

This is an example of how to virtualize class hierarchies in Eclipse OMR.

### DIFF
--- a/compiler/codegen/ELFRelocationResolver.hpp
+++ b/compiler/codegen/ELFRelocationResolver.hpp
@@ -33,7 +33,7 @@ namespace TR {
 /**
  * @brief The ELFRelocationResolver class translates between a TR::StaticRelocation and the ELF relocation type required for the platform.
  */
-class OMR_EXTENSIBLE ELFRelocationResolver : public ::OMR::ELFRelocationResolverConnector
+class /*OMR_EXTENSIBLE*/ ELFRelocationResolver : public ::OMR::ELFRelocationResolverConnector
    {
    };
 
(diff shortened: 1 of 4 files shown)


mstoodle:
Do we really need both OMR_EXTENSIBLE and OMR_EXTENSION ? Could we get away with just OMR_EXTENSIBLE? Presumably the annotation processor is smart enough to be able to identify the annotation is on a class versus on an instance function?

xliang6:
If we only use one macro to both signify a class being statically extensible and mark a function virtual or a part of the OMR API, what happens if we want to redefine the new macro to `final` in a downstream project for performance reasons?

IMHO, the two macros having almost the same name is confusing. Maybe we can name the new macro OMR_API or something to that effect?

charliegracie:
Is this PR still active or has it been super-ceded by newer work

samasri:
It is still active. This [wiki page](https://github.com/samasri/omr/wiki/Final-State-of-Project) describes the current state of the project for which this PR belongs.

0xdaryl:
This PR was created as part of a research project led by Dr. Sarah Nadi at the University of Alberta.  I discussed the resolution of this family of PRs with Dr. Nadi last week.  We agreed that since an effective alternate solution to address the extensibility requirements of the OMR compiler has not yet been determined (investigative work to do so will be part of a research project as described by Dr. Nadi at the Feb 25, 2019 OMR Architecture Meeting in #3607), plus the fact that the classes touched by these PRs represent only a partial and incomplete "revirtualization" solution, and that further progress on these PRs is unlikely in the near term, that we will close these PRs in the main OMR repository.

Issue #2397, opened 2018-03-27
Define JitBuilder client API


mstoodle:
One of the high level goals for the JitBuilder library is to enable different language bindings for the JitBuilder API. Right now, that effort is sabotaged a bit because the public JitBuilder API is mixed together with the JitBuilder implementation classes public APIs. We should really separate the implementation from the client API.

I have been working on this for a while under another umbrella item (earlier prototypes have been discussed a bit in the context of #1819), and will submit a PR soon. The solution I've come down on creates a parallel hierarchy of "client" objects that define the API independently of the "implementation" objects that implement the API. The current set of TR::IlBuilder class and friends basically become the "implementation" classes, with each implementation object having a pointer to its corresponding client object (and vice versa). The implementation classes also define functions that other implementation classes call.

So, clients will create a (for example) `MethodBuilder` client object which automatically creates an implementation `TR::MethodBuilder` object and cross links them so that anytime a service is called on the client object, it can be forwarded to the appropriate implementation object (and vice versa for callbacks like `IlBuilder::buildIL()` ). You can think of the client object as a shim to call services against the implementation object.

Although it feels a bit cumbersome to have a parallel set of classes, I have a strategy to make it more manageable :) . That's not the focus of this issue (link to be added to relevant issue discussing that strategy).

There will be some breaking fallout from this kind of change, so I wanted to open this issue so we can discuss that fallout.

JitBuilder classes are currently OMR compiler classes, which means you use the class name as mapped into the TR namespace (e.g. TR::IlBuilder, TR::MethodBuilder). After the client objects are created, clients will work with those classes which are *not* OMR compiler classes. That means we can put them anywhere we like (except into the TR namespace :) ).

I see a few options:
1. Put them into the global namespace
2. Put them into a long namespace like `JitBuilder::` or `jitbuilder::`
3. Put them into an abbreviated namespace like `JB::`

I don't think 1 is the best idea. It's easy, but it just doesn't feel responsible or like it's good naming hygiene. 2 feels most satisfying to me, but makes code look a bit more "wordy" unless you rely on "using". 3 is just another two-letter acronym so the code will look more or less similar to what it is now, but "JB" could be a lot of different things and doesn't "read" well to my eye.

If I had to decide myself, I would probably come down on `JitBuilder::` or `jitbuilder::` because I know not everyone likes those pesky shift keys, but I thought I'd open it up to people who are using JitBuilder to place a vote or even to suggest an alternative I haven't thought of.

[...]

PR #2587, opened 2018-05-25
Change custom-allocated maps to pointers in ilgen/MethodBuilder and ilgen/TypeDictionary

samolisov:
The `compiler/ilgen's` `MethodBuilder` and `TypeDictionary` classes
contain a number of class members, maps, that use a custom allocator
(`TR::typed_allocator<...> _memoryRegion`). The allocator itself is
deleted in the destructors of `MethodBuilder` and `TypeDictionary`:

```cpp
_memoryRegion->~Region();
::operator delete(_memoryRegion, TR::Compiler->persistentAllocator());
```

Then, the class destructors automagically invoke the destructors for
member objects: `_structsByName`, `_unionsByName`, `_symbols`, etc.
Eventually, the method `_Node::_Freenode0` of the `_Tree_node` STL
class (std::map is implemented as a tree) for the head node will be
invoked. The method uses the allocator (`_memoryRegion` in our case, the
destroyed yet) to deallocate the head node and, since the allocator has
been destroyed yet in the destructor of the `MethodBuilder` or
`TypeDictionary` class, an access violation error appears:

```
[----------] 2 tests from JITILBuilderTest
[ RUN      ] JITILBuilderTest.ControlFlowTest
unknown file: error: SEH exception with code 0xc0000005 thrown in the
test body.
[  FAILED  ] JITILBuilderTest.ControlFlowTest (351 ms)
```

To avoid the situation, the `map` members of the `MethodBuilder` and
`TypeDictionary` classes are replaced with the pointers to maps and
the allocated memory is explicitly reclaimed in the destructors before
the allocator is being deleted.

### DIFF
--- a/compiler/ilgen/MethodBuilder.cpp
+++ b/compiler/ilgen/MethodBuilder.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2016 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -87,13 +87,13 @@ MethodBuilder::MethodBuilder(TR::TypeDictionary *types, OMR::VirtualMachineState
    _methodName("NoName"),
    _returnType(NoType),
    _numParameters(0),
-   _symbols(str_comparator, *_memoryRegion),
-   _parameterSlot(str_comparator, *_memoryRegion),
-   _symbolTypes(str_comparator, *_memoryRegion),
-   _symbolNameFromSlot(std::less<int32_t>(), *_memoryRegion),
-   _symbolIsArray(str_comparator, *_memoryRegion),
-   _memoryLocations(str_comparator, *_memoryRegion),
-   _functions(str_comparator, *_memoryRegion),
+   _symbols(new SymbolMap(str_comparator, *_memoryRegion)),
+   _parameterSlot(new ParameterMap(str_comparator, *_memoryRegion)),
+   _symbolTypes(new SymbolTypeMap(str_comparator, *_memoryRegion)),
+   _symbolNameFromSlot(new SlotToSymNameMap(std::less<int32_t>(), *_memoryRegion)),
+   _symbolIsArray(new ArrayIdentifierSet(str_comparator, *_memoryRegion)),
+   _memoryLocations(new MemoryLocationMap(str_comparator, *_memoryRegion)),
+   _functions(new FunctionMap(str_comparator, *_memoryRegion)),
    _cachedParameterTypes(0),
    _definingFile(""),
    _newSymbolsAreTemps(false),
@@ -113,14 +113,21 @@ MethodBuilder::MethodBuilder(TR::TypeDictionary *types, OMR::VirtualMachineState
 MethodBuilder::~MethodBuilder()
    {
    // Cleanup allocations in _memoryRegion *before* its destroyed below (see note in constructor)
-   _symbols.clear();
-   _parameterSlot.clear();
-   _symbolTypes.clear();
-   _symbolNameFromSlot.clear();
-   _symbolIsArray.clear();
-   _memoryLocations.clear();
-   _functions.clear();
-
+   _symbols->clear();
+   ::operator delete(_symbols);
+   _parameterSlot->clear();
+   ::operator delete(_parameterSlot);
+   _symbolTypes->clear();
+   ::operator delete(_symbolTypes);
+   _symbolNameFromSlot->clear();
+   ::operator delete(_symbolNameFromSlot);
+   _symbolIsArray->clear();
+   ::operator delete(_symbolIsArray);
+   _memoryLocations->clear();
+   ::operator delete(_memoryLocations);
+   _functions->clear();
+   ::operator delete(_functions);
+   // Cleanup the alocators
    _trMemory->~TR_Memory();
    ::operator delete(_trMemory, TR::Compiler->persistentAllocator());
    _memoryRegion->~Region();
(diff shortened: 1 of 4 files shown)


fjeremic:
@dsouzai this seems like your area of expertise. I would like to delegate the reviews to you.

dsouzai:
I'm a bit torn on this. On the one hand, the changes do fix the problem. But on the other hand, it seems like that encapsulation in TR::TypedDictionary is broken, at least wrt to the allocator. I think the right thing to do would be to fix the underlying problem. One way to fix it would be to something like this:

```
SomeTest::doTest()
   {
   int32_t rc = 0;
   uint8_t *entry=0;

   TR::SegmentProvider segmentProvider(1 << 16, TR::Compiler->rawAllocator);
   TR::Region memoryRegion(segmentProvider, TR::Compiler->rawAllocator) );
      {
      TR::TypeDictionary types(memoryRegion);
      }
   }
```

where the constructor of `TypeDictionary` is modified to take in a `TR::Region &` and the segment provider and region are locals in an outer scope. I don't think there are too many places where we initialize a `TR::TypeDictionary` local object, so it shouldn't be too much work.

@mpirvu @lmaisons What do you think?

lmaisons:
I don't know if things have changed, but what I remember is that we were not supposed to assume `::operator new` was safe to use in this context, instead delegating to it in the implementation of the underlying allocator if desired.

lmaisons:
@mstoodle It might be worth reconsidering the decision to not include the memory allocation related headers in the client-visible include set, given that this contortion is driven by that requirement.

samolisov:
I see about 40 of references to the constructor of `TR::TypeDictionary`, the most are in JitBuilder. Some references are created in macros (`JBTestUtil.hpp`, `DEFINE_TYPES`) and it could take some time to inject an instance of `TR::Region` there and delete them then.

mstoodle:
JitBuilder was purposefully designed not to expose compiler implementation details like the memory allocation stuff. I would prefer to keep it that way.

IMO the TR::TypeDictionary object has the longest lifetime, so the ultimate creation/destruction of the underlying memory region should be tied to the TypeDictionary's lifetime. The TR::TypeDictionary object and TR::MethodBuilder object's lifetimes are dictated by the client, but the client should be expected to destroy MethodBuilder objects before the TypeDictionary, or they should die at the "same time" via a closing scope.

I think @samolisov 's changes are in line with this position, unless I've misread something @dsouzai @lmaisons ?

dsouzai:
> JitBuilder was purposefully designed not to expose compiler implementation details like the memory allocation stuff. I would prefer to keep it that way.

Ah, I didn't know this is one of the tenets of JitBuilder. In that case, couldn't we wrap the `_segmentProvider` and `_memoryRegion` in a struct, something like

```
struct TypeDictMemory
   {
   TypeDictMemory()
      {
      _segmentProvider( new(TR::Compiler->persistentAllocator()) TR::SystemSegmentProvider(1 << 16, TR::Compiler->rawAllocator) ),
      _memoryRegion( new(TR::Compiler->persistentAllocator()) TR::Region(*_segmentProvider, TR::Compiler->rawAllocator) ),
      }
   ~TypeDictMemory()
      {
      _memoryRegion->~Region();
      ::operator delete(_memoryRegion, TR::Compiler->persistentAllocator());
      static_cast<TR::SystemSegmentProvider *>(_segmentProvider)->~SystemSegmentProvider();
      ::operator delete(_segmentProvider, TR::Compiler->persistentAllocator());
      }
   TR::SegmentProvider *_segmentProvider;
   TR::Region *_memoryRegion;
   }
```
And have `TypeDictMemory` as the first member of `TypeDictionary`, so that it is the last one to get destroyed?

samolisov:
@dsouzai I've tried the proposed solution and can say this works fine: the struct is defined as the first member of a corresponding class and so that it is the last one to get destroyed, I see no `SEH exception with code 0xc0000005` more. I can open a PR with the commit but I'm not sure is have allocators as class members better than to use pointers on them?

samolisov:
@dibyendumajumdar #2596 also may be required to build the compiler on Windows.

dibyendumajumdar:
@samolisov  Yes I had already done a version of those - re #2434 I just defined WINDOWS 

My version is here: https://github.com/dibyendumajumdar/nj
The goal is to create a cutdown version of OMR with just the compiler/JitBuilder. In my case less is more.

I have applied some of your patches - really appreciate the work you are doing.

dsouzai:
>I'm not sure is have allocators as class members better than to use pointers on them?

I think they have to be pointers in order to minimize the number of header files included.

>I've tried the proposed solution and can say this works fine: the struct is defined as the first member of a corresponding class and so that it is the last one to get destroyed, I see no SEH exception with code 0xc0000005 more. I can open a PR with the commit

We'll have to ask @mstoodle since he's the owner of this code; Mark, does the idea of wrapping the allocators in a struct embedded into TypeDictionary (and potentially MethodBuilder) sound ok to you?

mstoodle:
@dsouzai yes, I'm fine with it, along with instructive comments to explain why it's being done that way :) .

On this particular "tenet" of JitBuilder: soon this point will become less important with the separation of the client API from the JitBuilder implementation classes (see #2397 which has been making unsatisfyingly slow progress over the last couple of weeks due to my distractedness :( ). I would still like to avoid making the implementation any (much?) more complicated than it already is.

Until the client API lands, any additional header file dependency for the JitBuilder header files will require extra build work to copy those headers over into the `jitbuilder/release` directory. And that will be throw-away work because the client API breaks such dependencies.

samolisov:
@dsouzai @mstoodle Should I close this PR and open a new one with an implementation of the solution proposed by @dsouzai?

dsouzai:
Probably best to just open a new PR and link this one to it, would be a lot cleaner I think.

samolisov:
@dsouzai OK, #2597 is opened.

dsouzai:
@samolisov Could you close this PR since you opened a new one?

PR #2595, opened 2018-05-28
AArch64: Initial version of MemoryReference files

knn-k:
This commit implements the initial version of MemoryReference files
for aarch64.


knn-k:
[Memo]
MemoryReference class for other platforms has other types of constructors, that take Node, SymbolReference, another MemoryReference, and so on, as arguments.  But their forms are different from platform to platform, so I only implemented the ones that take base register along with other arguments in the initial version.
Implementation of those constructors that don't take registers is a future work.

Issue #1952, opened 2017-11-03
JItBuilder on Windows


mgaudet:
In service of #1859, I've been investigating building JitBuilder on Windows. 

While longer term I want to use CMake to do these builds, I wanted to build on the already working JIT Compilation story for OpenJ9, so [I have borrowed the Makefile fragments from OpenJ9 that drive windows compilation for the J9 JIT](https://github.com/mgaudet/omr/commit/23b957a3357bce130c830ee5b448540222afec37). 

There's a couple of issues that I've found. Some easier to tackle than others: I want to use this issue to gather notes on the status of this, and how to proceed. 

##  C++ Issues 

- [ ] Some minor type errors, where container allocator types don't quite match what's expected. The one places that has caught this, using `auto` is a simple correction. 

- [ ] Deleted copy constructors aren't supported, so type dictionary needs to have its copy constructor made private instead. 

## Platform Issues: 

- [ ] The compiler technology uses `<unistd.h>` for `getpid()` in `compiler/control/CompileMethod.cpp` as part of `writePerfToolEntry`.  I've partially dealt with this issue by wrapping the body the function in `#ifndef WINDOWS`, but it seems that this should be at least wrapped in `#ifdef LINUX`. 

- [ ] The build version of windows we have doesn't _seem_ to have a definition of `snprintf`, so I ended up temporarily working around this with 
    
    ```
    +#ifdef WINDOWS
    +#define snprintf _snprintf
    +#endif
    ```

    but that doesn't feel like the right answer. 

- [ ] What to do about `mmap`? The JitBuilder CodeCache uses `mmap` directly, and as well as using `mmap` in 

    https://github.com/eclipse/omr/blob/6abf600ca9219a2d26e679c4e1eee21c2cdb9173/compiler/env/FEBase_t.hpp#L98-L112

The platform issues seem like an argument in favour of reprioritizing https://github.com/eclipse/omr/issues/1642

mgaudet:
Current sticking point is code cache.

lmaisons:
The mmap equivalent on Windows is [VirtualAlloc](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366887(v=vs.85).aspx)

Note:
> To execute dynamically generated code, use VirtualAlloc to allocate memory and the VirtualProtect function to grant PAGE_EXECUTE access.

dsouzai:
As Luc mentioned you can use VirtualAlloc on Windows as is done in DebugSegmentProvider:

[...]

PR #2596, opened 2018-05-28
Use the VirtualAlloc API to allocate code cache

samolisov:
JitBuilder, as well as the compiler test infrastructure, use the
`sys/mman.h` memory management api to allocate an executable code cache
for the compiler. Windows doesn't provide the `sys/mman.h` API, so the
`VirtualAlloc` function from the Windows API should be used instead of
`mmap`.

Currently, the method `FEBase<Derived>::allocateRelocationData` (see
`compiler/env/FEBase_t.hpp`) is used only from the
`TR_PPCTableOfConstants` class (so, won't be compiled under Windows) but
in order to compiler the method, the `VirtualAlloc` function are used
there too.

TODO: the use of `VirtualAlloc`/`mmap` has to be replaced with the
`omrvmem_reserve_memory_ex` function from the port library.

The PR depends on #2434

Issue: #1952

### DIFF
--- a/compiler/env/FEBase_t.hpp
+++ b/compiler/env/FEBase_t.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


mstoodle:
LGTM as a stepping stone towards the right portlib based solution.

charliegracie:
@genie-omr build all

fjeremic:
> TODO: the use of VirtualAlloc/mmap has to be replaced with the omrvmem_reserve_memory_ex function from the port library.

Referencing https://github.com/eclipse/omr/issues/1952 for this one. I suppose we partly completed it but need to push the code down to the portlib and start using the portlib API.

PR #2597, opened 2018-05-28
Wrap memory allocators in a struct

samolisov:
The `compiler/ilgen's` `MethodBuilder` and `TypeDictionary` classes
contain a number of class members, maps, that use a custom allocator
(`TR::typed_allocator<...> _memoryRegion`). The allocator itself is
deleted in the destructors of `MethodBuilder` and `TypeDictionary`:

```cpp
_memoryRegion->~Region();
::operator delete(_memoryRegion, TR::Compiler->persistentAllocator());
```

Then, the class destructors automagically invokes the destructors for
member objects: `_structsByName`, `_unionsByName`, `_symbols`, etc.
Eventually, the method `_Node::_Freenode0` of the `_Tree_node` STL
class (std::map is implemented as a tree) for the head node will be
invoked. The method uses the allocator (`_memoryRegion` in our case,
the destroyed yet) to deallocate the head node and, since the allocator
has been destroyed yet in the destructor of the `MethodBuilder` or
`TypeDictionary` class, an access violation error appears:

```
[----------] 2 tests from JITILBuilderTest
[ RUN      ] JITILBuilderTest.ControlFlowTest
unknown file: error: SEH exception with code 0xc0000005 thrown in the
test body.
[  FAILED  ] JITILBuilderTest.ControlFlowTest (351 ms)
```

To avoid the situation, the memory allocators `TR::SegmentProvider`,
`TR::Region` and `TR_Memory` are wrapped in a struct and the struct is
defined as the first member of `MethodBuilder` and `TypeDictionary`
classes respectively, so that it is the last one to get destroyed.

### DIFF
--- a/compiler/ilgen/MethodBuilder.cpp
+++ b/compiler/ilgen/MethodBuilder.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2016 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


samolisov:
The solution is proposed by @dsouzai in #2587, the discussion there is very informative too.

dsouzai:
@mstoodle For your review.

mstoodle:
@genir-omr build all\

mstoodle:
@genie-omr build all

PR #2605, opened 2018-05-30
Fix container detection with systemd

Ali-Ni:
- Fix container detection with systemd

  Up until now, `isRunningInContainer` assumed the process was operating
in a container if init (pid 1) wasn't in the root cgroup in any subsystem.
This is broken in systemd, as it encapsulates pid 1 in a special
"init.scope" group.   

  Added the exception of cgroup "init.scope" to fix for systemd
version 226 and onwards.  

  [See here](https://github.com/systemd/systemd/blob/master/NEWS#L2951) 

- Fix /proc/[pid]/cgroup file reading with systemd

   Systemd adds an entry in /proc/[pid]/cgroup files that has
no subsystem name. This currently breaks the functions that try
to read them, as those functions use fscanf with a format string
expecting a subsystem name.   

   This change puts each file entry into a buffer and first attempts
to read expecting a subsystem name. If this fails, it tries
again expecting no subsystem name. Both `isRunningInContainer` and
`readCgroupFile` are affected by this change.   

Signed-off-by: Mohammad Ali Nikseresht <anikser@ibm.com>

### DIFF
--- a/port/common/omrport.tdf
+++ b/port/common/omrport.tdf
@@ -1072,3 +1072,6 @@ TraceException=Trc_PRT_signal_omrsig_register_os_handler_invalid_portlibSignalFl
 
 TraceEntry=Trc_PRT_signal_omrsig_is_master_signal_handler_entered Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_master_signal_handler: Entered, osHandler=%p"
 TraceExit=Trc_PRT_signal_omrsig_is_master_signal_handler_exiting Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_master_signal_handler: Exiting, rc=%d"
+
+TraceException=Trc_PRT_readCgroupFile_fgets_failed Group=sysinfo Overhead=1 Level=3 NoEnv Template="readCgroupFile: fgets failed for %s with errno=%d"
+TraceException=Trc_PRT_isRunningInContainer_fgets_failed Group=sysinfo Overhead=1 Level=3 NoEnv Template="isRunningInContainer: unexpected format of file %s with errno=%d"
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -257,6 +257,7 @@ struct  {
 
 #define OMR_CGROUP_V1_MOUNT_POINT "/sys/fs/cgroup"
 #define ROOT_CGROUP "/"
+#define SYSTEMD_INIT_CGROUP "/init.scope"
 #define OMR_PROC_PID_ONE_CGROUP_FILE "/proc/1/cgroup"
 
 /* An entry in /proc/<pid>/cgroup is of following form:
@@ -3739,7 +3740,7 @@ isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer)
 				goto _end;
 			}
 
-			if (0 != strcmp(ROOT_CGROUP, cgroup)) {
+			if ((0 != strcmp(ROOT_CGROUP, cgroup)) && (0 != strcmp(SYSTEMD_INIT_CGROUP, cgroup))) {
 				*inContainer = TRUE;
 				break;
 			}
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -267,6 +267,7 @@ struct  {
  * 	7:cpuacct,cpu:/mycgroup
  */
 #define PROC_PID_CGROUP_ENTRY_FORMAT "%d:%[^:]:%s"
+#define PROC_PID_CGROUP_SYSTEMD_ENTRY_FORMAT "%d::%s"
 
 /* Currently 12 subsystems or resource controllers are defined.
  */
@@ -1189,7 +1190,6 @@ omrsysinfo_get_number_CPUs_by_type(struct OMRPortLibrary *portLibrary, uintptr_t
 				}
 			}
 		}
-
 		if (0 == toReturn) {
 			Trc_PRT_sysinfo_get_number_CPUs_by_type_failedBound("errno: ", errno);
 		}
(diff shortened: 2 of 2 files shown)


Ali-Ni:
@charliegracie requesting review

Ali-Ni:
Running into other issues with the isRunningInContainer function... marking this as WIP until I can sort them out.

Ali-Ni:
@charliegracie Everything seems to be working, can you review this please?

charliegracie:
@genie-omr build all

port/common/omrport.tdf
```diff
@@ -1021,6 +1021,7 @@ TraceException=Trc_PRT_addCgroupEntry_oom_for_cgroup_entry Group=sysinfo Overhea
 TraceEvent=Trc_PRT_addCgroupEntry_added_new_entry Group=sysinfo Overhead=1 Level=3 NoEnv Template="addCgroupEntry: added new cgroup entry for subsystem %s and cgroup %s"
 
 TraceException=Trc_PRT_readCgroupFile_fopen_failed Group=sysinfo Overhead=1 Level=1 NoEnv Template="readCgroupFile: fopen failed for %s with errno=%d"
+TraceException=Trc_PRT_readCgroupFile_fgets_failed Group=sysinfo Overhead=1 Level=3 NoEnv Template="readCgroupFile: fgets failed for %s with errno=%d"
```
charliegracie: All new tracepoints have to be added to the end of the file.  This allows the tool generating the tracepoints to keep the same IDs for tracepoints release to release.  Can you move the new tracepoints to the bottom?

port/unix/omrsysinfo.c
```diff
@@ -3517,25 +3518,46 @@ readCgroupFile(struct OMRPortLibrary *portLibrary, int pid, BOOLEAN inContainer,
 	}
 
 	while (0 == feof(cgroupFile)) {
+		char buffer[PATH_MAX + 1024];
```
charliegracie: The PATH_MAX variable implies that is the maximum buffer size.  How come you are adding 1024 to this size?

port/unix/omrsysinfo.c
```diff
@@ -3722,24 +3744,44 @@ isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer)
...
+				}
+				subsystems[0] = '\0';
 			} else if (3 != rc) {
 				Trc_PRT_isRunningInContainer_unexpected_format(OMR_PROC_PID_ONE_CGROUP_FILE);
 				rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_READ_FAILED, "unexpected format of %s file", OMR_PROC_PID_ONE_CGROUP_FILE);
 				goto _end;
 			}
 
-			if (0 != strcmp(ROOT_CGROUP, cgroup)) {
+			if (0 != strcmp(ROOT_CGROUP, cgroup) && 0 != strcmp(SYSTEMD_INIT_CGROUP, cgroup)) {
```
charliegracie: if there is more than one complex conditional statement each statement needs to be wrapped in brackets.  This line should be:
```
if ((0 != strcmp(ROOT_CGROUP, cgroup)) && (0 != strcmp(SYSTEMD_INIT_CGROUP, cgroup))) {
```

port/unix/omrsysinfo.c
```diff
@@ -3722,24 +3744,44 @@ isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer)
...
+			} else if (1 == rc) {
+				rc = sscanf(buffer, PROC_PID_CGROUP_SYSTEMD_ENTRY_FORMAT, &hierId, cgroup);
+				Assert_PRT_true(strlen(cgroup) < PATH_MAX); 
+
+				if (2 != rc) {
+					Trc_PRT_isRunningInContainer_unexpected_format(OMR_PROC_PID_ONE_CGROUP_FILE);
+					rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_READ_FAILED, "unexpected format of %s file", OMR_PROC_PID_ONE_CGROUP_FILE);
+					goto _end;
+				}
+				subsystems[0] = '\0';
```
charliegracie: why is subsystems being modified?

port/unix/omrsysinfo.c
```diff
@@ -3722,24 +3744,44 @@ isRunningInContainer(struct OMRPortLibrary *portLibrary, BOOLEAN *inContainer)
...
+				rc = portLibrary->error_set_last_error_with_message_format(portLibrary, OMRPORT_ERROR_SYSINFO_PROCESS_CGROUP_FILE_READ_FAILED, "fgets failed to read %s file stream with errno=%d", OMR_PROC_PID_ONE_CGROUP_FILE, osErrCode);
+				goto _end;
+			} else if (NULL == buffer) {
+				break;
+			}
+			rc = sscanf(buffer, PROC_PID_CGROUP_ENTRY_FORMAT, &hierId, subsystems, cgroup);
 			/* Ensure we didn't overflow */
 			Assert_PRT_true(strlen(subsystems) < 1024);
-			Assert_PRT_true(strlen(cgroup) < PATH_MAX);
+			Assert_PRT_true(strlen(cgroup) < PATH_MAX);
```
charliegracie: Can you remove the extra tab(s)

port/unix/omrsysinfo.c
Ali-Ni: This is somewhat arbitrary - previously group and subsystems were of max size PATH_MAX and 1024 respectively, and since buffer is to contain both of them, this is the size I chose. Realistically, no entry will get close to 1024, and PATH_MAX itself is a macro for 1024. Should I just change all of these to be of length PATH_MAX?

port/unix/omrsysinfo.c
Ali-Ni: If it gets to this point, it means subsystems is empty - this is to overwrite the subsystem from the previous iteration. Subsystems is not being used here, but I left it in since I don't know how this function may be changed in the future. Should I remove this for the sake of efficiency?

charliegracie: If it is not required it would be great if it could be removed.

port/unix/omrsysinfo.c
Ali-Ni: Done. Also removed now unnecessary overflow checks (as fgets is capped at the number of characters it will read)

PR #2508, opened 2018-04-27
AArch64: Initial version of OMRCodeGenerator

knn-k:
This commit implements the initial version of OMRCodeGenerator files
for aarch64.
This is a part of #2503.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #2606, opened 2018-05-31
AArch64: Remove BOM from OMRMachine.cpp

knn-k:
This commit removes the unnecessary byte order mark (BOM) from
OMRMachine.cpp.


knn-k:
Another file, OMRCodeGenerator.cpp for aarch64, also contains a byte order mark.  I am overwriting the file by #2508.
As far as I searched in OMR, no other file contains BOM.

PR #2607, opened 2018-05-31
Fix tril's Xcall related tests

samolisov:
Tril parses hexadecimal numbers if and only if they start with the `0x`
prefix. Unfortunately, some C/C++ compilers put the prefix for the `%p`
format specifier (so, they translate `%p` to something like
`0x0123456789ABCDEF`) while others don't. If the `0x` prefix is absent
before the address attribute value of the `Xcall` node, tril
interprets leading digits as a decimal number and the rest starting from
not a digital symbol (A,B,C,D,E or F) as an anonymous attribute. As a
result, the formed `Xcall` node will try to invoke a function by a wrong
address and an Access Violation Error appears.

The commit fixes the problem explicitly adding the `0x` prefix before the
format specifier in all code for `Xcall` tests. Output for the ilgen trace
of the `Xcall` opcode is fixed too.

### DIFF
--- a/fvtest/compilertriltest/CallTest.cpp
+++ b/fvtest/compilertriltest/CallTest.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -28,8 +28,8 @@ int32_t oracleBoracle(int32_t x) { return x + 1123; } // Randomish number to avo
 
 TEST_F(CallTest, icallOracle) { 
     char inputTrees[200] = {0};
-    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=%p args=[Int32] (iload parm=0)) )  ))";
-    std::snprintf(inputTrees, 200, format_string, &oracleBoracle);
+    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=0x%jX args=[Int32] (iload parm=0)) )  ))";
+    std::snprintf(inputTrees, 200, format_string, reinterpret_cast<uintmax_t>(&oracleBoracle));
     auto trees = parseString(inputTrees);
 
     ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
--- a/fvtest/compilertriltest/LinkageTest.cpp
+++ b/fvtest/compilertriltest/LinkageTest.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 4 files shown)


Leonardo2718:
LGTM! :+1:

fjeremic:
@genie-omr build all

fjeremic:
@samolisov is there more changes pending for this PR? If not I'm happy to merge once WIP: is removed.

samolisov:
I'm not sure, I see a strange behavior when the code is built using MSVC: the LinkageTest hangs when we are trying to call a function with 4 arguments, Clang doesn't provide the same thing: the all tests run but many errors are there, access violation and some others. I wish to solve the hang issue and then remove the WIP status.

samolisov:
I've changed lines like the following:

```cpp
    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=0x%jX args=[Int32] (iload parm=0)) )  ))";
```

with the construction:

```cpp
#if defined(OMR_ENV_DATA64)
    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=0x%016llX args=[Int32] (iload parm=0)) )  ))";
#else
    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=0x%08lX args=[Int32] (iload parm=0)) )  ))";
#endif /* OMR_ENV_DATA64 */
```

(the format specifier `%jX` writes a warning when the code is being compiled using clang: the argument must be cast to `uintmax_t` but it's not allowed for a function pointer.)

The `omrcfg.h` header is included in the files where the `OMR_ENV_DATA64` compile definition is in use.

I've removed the `WIP` label: the strange behaviour of the LinkageTest when the code is built using MSVC is a story for another issue and PR, I'm going to open the issue ASAP and attach the code generator traces. I think some help from more experienced colleagues is needed to fix it.

Leonardo2718:
> (the format specifier %jX writes a warning when the code is being compiled using clang: the argument must be cast to uintmax_t but it's not allowed for a function pointer.)

Really? It works fine for me: https://godbolt.org/g/vwN2JV.

fjeremic:
> Really? It works fine for me: godbolt.org/g/vwN2JV.

I tried all the versions as well and I don't get the error either.

samolisov:
@fjeremic @Leonardo2718 There is no error but a warning is generated:

```
..\fvtest\compilertriltest\CallTest.cpp(37,51):  warning: format specifies type 'uintmax_t' (aka 'unsigned long long') but the argument has type 'int32_t (*)(int32_t)' (aka 'int (*)(int)') [-Wformat]
    std::snprintf(inputTrees, 200, format_string, &oracleBoracle);
                                   ~~~~~~~~~~~~~  ^~~~~~~~~~~~~~
..\fvtest\compilertriltest\CallTest.cpp(33,100):  note: format string is defined here
    const auto format_string = "(method return=Int32 args=[Int32] (block (ireturn (icall address=0x%jX args=[Int32] (iload parm=0)) )  ))";
```

But `%016llX` is no a solution also, a similar warning will be there. Clang 6.0.0. What do you think, should I return back the code with `%jX`? The new code requires `omrcfg.h` as well as a conditional compilation but returns a little bit better-formatted addresses. On the other hand, the generated tril code is machine-readable only.

Leonardo2718:
The warning should go away if you use `reinterpret_cast<uintmax_t>(&oracleBoracle)`.

samolisov:
I tried but MSVC can't compile LinkageTest.cpp if I cast the `fourthArg` function to `uintmax_t` using the `reinterpret_cast` operator (Clang can, CallTest.cpp also is compiled by both compilers even if `oracleBoracle` is reinterpteted as `uintmax_t`):

```
..\fvtest\compilertriltest\LinkageTest.cpp(205): error C2440: 'reinterpret_cast': cannot convert from 'T (__cdecl *)(T,T,T,T)' to 'uintmax_t'
```

I think the generics (oh, sorry, templates) is the reason.

fjeremic:
Dare I say we use a C-style unsafe cast for this case?

samolisov:
@fjeremic No, `static_cast` is used:

```cpp
static_cast<TypeParam (*)(TypeParam,TypeParam,TypeParam,TypeParam)>(fourthArg<TypeParam>),// address
```

Leonardo2718:
Oh! In that case, try using two casts (newlines added for readability):

```c++
reinterpret_cast<reinterpret_cast>(
  static_cast<TypeParam (*)(TypeParam,TypeParam,TypeParam,TypeParam)>(
    fourthArg<TypeParam>
  )
)
```

It's ugly, but this might be one of those awkward cases where you really have to hold the compilers hand (often happens with older compilers).

samolisov:
@Leonardo2718 Wow! Thank you, it looks like a solution. 

@Leonardo2718 @fjeremic The commit has been updated, please have a look.

charliegracie:
@genie-omr build all

PR #2608, opened 2018-05-31
Two small compile fixes for the GC

rwy0717:
- missing scavenger ifdef
- missing link libraries for GC

/cc @charliegracie


charliegracie:
@dnakamura can you give this a thumbs up?

charliegracie:
@genie-omr build all

PR #2610, opened 2018-05-31
Use compiler defined size_t instead of std::size_t

fjeremic:
For extensible classes such as IlGeneratorMethodDetails when overridden
in downstream projects we may encounter compilation issues as the
extended class will look for a suitable delete operator which should be
compatible with the types on the new operator.

size_t and std::size_t however are not compatible for all compilers
(gcc) and will cause errors complaining that no suitable delete
operator was found for the extensible class.

### DIFF
--- a/compiler/env/TRMemory.hpp
+++ b/compiler/env/TRMemory.hpp
@@ -948,6 +948,11 @@ namespace TR
 
    typedef CS2::shared_allocator < GlobalSingletonAllocator > GlobalAllocator;
 
+   static GlobalAllocator globalAllocator(const char *name = NULL)
+      {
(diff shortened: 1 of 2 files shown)


fjeremic:
This should fix the compilation issues for OpenJ9. Leaving WIP: until I've tested it does indeed resolve things. Meanwhile I suppose we can launch builds to accelerate things.

@genie-omr build all

fjeremic:
@genie-omr build all

charliegracie:
@genie-omr build xlinux

fjeremic:
Issues seem addressed. I removed the WIP: as this is ready for merge if committers are satisfied.

PR #2611, opened 2018-06-01
Improve setting signalsWithMasterHandlers and fix omrsig_protect

babsingh:
Currently, signalsWithMasterHandlers is set at two places:
omrsig_register_os_handler and registerMasterHandlers.

With this changeset, signalsWithMasterHandlers will only be set within
registerSignalHandlerWithOS after an OS handler has been successfully
registered.

A write barrier is issued before updating signalsWithHandlers and
signalsWithMasterHandlers within registerSignalHandlerWithOS. The write
barrier will allow us to read signalsWithHandlers and
signalsWithMasterHandlers without acquiring registerHandlerMonitor.

omrsig_protect serves asynchronous signals, and OS handlers for
asynchronous signals are generally registered once at startup. Now
onwards, signalsWithMasterHandlers is read without acquiring
registerHandlerMonitor within omrsig_protect. This serves two purposes. 

First, omrsig_protect won't acquire registerHandlerMonitor if a handler
is already registered. From past performance numbers, this would save
1828 instructions from the newDirectByteBuffer pathlength. 

Second, thread->monitor value ends up being NULL if the thread gets
aborted during the call to
omrthread_monitor_enter(registerHandlerMonitor). This impacts the
behavior of omrthread.c::interrupt_waiting_server. In OpenJ9,
intermittent segmentation faults are observed during calls to
Java_java_lang_Thread_interruptImpl. By not acquiring
registerHandlerMonitor if an OS handler is already registered,
omrsig_protect avoids conflict with the thread library and prevents
segmentation fault in OpenJ9.

### DIFF
--- a/port/unix/omrsignal.c
+++ b/port/unix/omrsignal.c
@@ -321,6 +321,7 @@ omrsig_protect(struct OMRPortLibrary *portLibrary, omrsig_protected_fn fn, void 
 	omrthread_t thisThread = NULL;
 	uint32_t rc = 0;
 	uint32_t flagsSignalsOnly = 0;
+	uint32_t flagsWithoutMasterHandlers = 0;
 
 	Trc_PRT_signal_omrsig_protect_entered(fn, fn_arg, handler, handler_arg, flags);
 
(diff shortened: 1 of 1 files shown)


babsingh:
fyi - @pshipton this is the fix for the gpf issue. i am testing changes in personal builds. i will remove the WIP after testing.

port/unix/omrsignal.c
```diff
@@ -1274,10 +1258,22 @@ registerSignalHandlerWithOS(OMRPortLibrary *portLibrary, uint32_t portLibrarySig
 		}
 	}
 
+	issueWriteBarrier();
+
 	/* Set the portLibrarySignalNo bit in signalsWithHandlers to record successful registration
 	 * of the handler. */
 	signalsWithHandlers |= portLibrarySignalNo;
```
pshipton: Seems these updates need to be atomic. Otherwise without a monitor bits could be lost.

babsingh: a monitor is always acquired before calling registerSignalHandlerWithOS.

babsingh:
Removed the WIP tag. Personal builds and local testing have succeeded.

charliegracie:
@genie-omr build all

port/unix/omrsignal.c
charliegracie: Is there a lock being held at this point while the flags are being modified?

babsingh: calls to `registerSignalHandlerWithOS` always acquire `registerHandlerMonitor`.

PR #2613, opened 2018-06-01
Create getter for method being compiled

mpirvu:
The Compiler object stores the method to be compiled in a field called
_method, but currently there is no way to retrieve it. This commit adds
that ability.


fjeremic:
@genie-omr build zlinux

mpirvu:
@fjeremic  All checks have passed, so it's ready to be merged

PR #2614, opened 2018-06-01
Add missed object alignment to optimum cache size

dmitripivkine:
calculateCopyScanCacheSizeForQueueLength() can return value which is not
aligned to object alignment. Round output to proper alignment.


charliegracie:
@genie-omr build all

PR #2615, opened 2018-06-01
Remove unnecessary includes of omr.h in compiler

nbhuiyan:
In 2 different files in the compiler, omr.h is included,
but not used at all.


fjeremic:
@genie-omr build all

PR #2617, opened 2018-06-01
Remove stale code related to Object.clone from VP code

mpirvu:
Some stale VP code related to Object.clone adds overhead for no good reason
and it should be removed.
The actual optimizations related to Object.clone are currently performed in
transformObjectCloneCall()


fjeremic:
@andrewcraik could I please delegate a review to you for this one?

andrewcraik:
So this code is dead as @mpirvu states - the current clone optimization is in the _objectCloneCalls/_objectCloneTypes lists and the _arrayCloneCalls and _arrayCloneTypes lists. The transformToOptimizedCloneCall removed here is guarded by an option that I don't see being enabled anywhere. I think I am fine with this being cleaned up.

fjeremic:
@genie-omr build all

PR #2624, opened 2018-06-04
Fix JitBuild compilation error on GCC version < 4.6

0dvictor:
GCC support of nullptr was introduced with version 4.6; therefore, avoid nullptr for code that may compile with an older version of GCC.


charliegracie:
Is auto available on all of our compilers?  I think it would be better to just move the `uint8_t *` down to that line instead of using `auto`.

0dvictor:
```auto``` was introduced with GCC 4.4, our minimal supported GCC version. It is also widely used in JIT and test utilities.

0xdaryl:
@genie-omr build all

mstoodle:
Sorry, I should have caught that on the review. Thanks for fixing it quickly, @0dvictor !

PR #2625, opened 2018-06-04
Skip handling of Scan Cache reminder size zero

dmitripivkine:
In MM_Scavenger::clearCache() handling of Scan Cache reminder can be
skipped if it's size is zero. It should prevent calling
abandonHeapChunk() unnecessary.


charliegracie:
@genie-omr build all

PR #2612, opened 2018-06-01
Remove clean TestJitConfig.cpp from compilertest

samolisov:
The translation module `fvtest/compilertest/runtime/TestJitConfig.cpp`
contains no executable code and looks only like a wrapper for
the `runtime/TestJitConfig.hpp` header. The file may be removed and this
won't break the compilation.

The commit was missing from #2546.

Signed-off-by: Pavel Samolysov <samolisov@gmail.com>

PR #2628, opened 2018-06-05
Enable JitBuilder and TestCompiler for Windows CI

samolisov:
After #2434, #2546, #2596 and #2612 have been merged, the JitBuilder as
well as TestCompiler components are ready to be built on Windows. The
Appveyor CMake configuration is updated to enable the components.

`OMR_GLUE` variable is not used by the project, so the variable is
removed from the configuration.


samolisov:
`win_bison` and `win_flex` must be installed on the Windows CI server, the programs are required for `tril`. The software can be downloaded from [sourceforge](https://sourceforge.net/projects/winflexbison/). I have no information (and as I get the access) to the buildserver, could someone from the committer team install `win_bison` and `win_flex` and add them to the `PATH` environment variable, please.

charliegracie:
For the AppVeyor machines we will need to figure out how to install it.  On the Eclipse OMR CI Windows machines I will look into installing win_bison and win_flex

charliegracie:
For AppVeyor can you try modifying the appveyor.yml to add a new  section after `environment`:
```
install:
  - cinst winflexbison
```

You may also need to add the following 2 lines to the `before_build` section:
```
  - rename c:\ProgramData\chocolatey\bin\win_flex.exe flex.exe
  - rename c:\ProgramData\chocolatey\bin\win_bison.exe bison.exe
```

charliegracie:
@genie-omr build win

charliegracie:
@genie-omr build win

charliegracie:
@genie-omr build win

charliegracie:
@genie-omr build win

charliegracie:
@genie-omr build win

charliegracie:
@genie-omr build win

charliegracie:
Hmmm even with win_bison and win_flex installed CMake is not finding them... I will keep digging

samolisov:
Are the path to them in the PATH? Locally I use CMake 3.11 it can find
win_flex as well as flex but only in PATH (as I see).

charliegracie:
@genie-omr build win

charliegracie:
Yeah they are both on the path.  When I just tried manually on the machine it worked. Maybe the Jenkins machine config is changing the path.... I will check that next

charliegracie:
@genie-omr build win

charliegracie:
When I log into the machine and open a cgwin terminal they are definitely on the path.  When I print the PATH variable from Jenkins it is not.  I tried disconnecting and reconnecting the node but that did not work.  I will keep trying... Hopefully by the time you get the AppVeyor file changed I will have figured this out.

charliegracie:
@genie-omr build win

charliegracie:
yay it seems to be compiling this time.   Wow there are a lot of compiler warnings in the compiler folder when compared to the OS X or Linux warnings.

charliegracie:
It looks like some files are missing `#include <stdio.h>` and `#include <string>`.  Also though the compiler test may be using C++ features not available in the minimum version of Visual Studio OMR supports.

samolisov:
@charliegracie Thank you very much for your effort! I've added the `winflexbison` component to the AppVeyor configuration but, as you mentioned before, some strange compilation errors appear:

```
C:\projects\omr\compiler\ilgen/MethodBuilder.hpp(117): error C3861: 'snprintf': identifier not found
```

On MSVC 2017 the error doesn't appear at all, I've double checked the out of my compilation using MSVC as well as clang.

charliegracie:
Yeah I think there are some headers included by default on newer compilers or these headers are included by other headers.

charliegracie:
I am going to try to create a commit based on top your branch that fixes these.  Is it ok if I open a PR against your repo if I get something working so the changes can flow through this PR?

samolisov:
Charlie, no problem.

samolisov:
I've rebased the branch on master but there are two conflicts: `compiler/ilgen/MethodBuilder.hpp` and `compiler/ilgen/TypeDictionary.hpp`.  I see new compiler directives in the files: `PUT_OMR_TYPEDICTIONARY_INTO_TR` and `PUT_OMR_METHODBUILDER_INTO_TR`, could you let me know how to enable them, please?

charliegracie:
@mstoodle can you provide some help here?  Thanks

samolisov:
My bad, I've missed the [commit](https://github.com/eclipse/omr/commit/26dd25ef94f0653d3c9e9ac4e86719fdfc64e237#diff-eaeee08d3e02a873c5f5106f25d0a0b2) by @mstoodle and resolved the conflict wrong. After a fix, the compiler is built ok on my machine. The conflicts are resolved.

charliegracie:
In the merging I think you lost a change in OMRMethodBuilder.hpp (the code was in MethodBuilder.hpp before). Windows does not provide `snprintf` in older version so we need to defined it ourselves.  Right after the `#include ...` in can you add the following lines:
```
#if defined (_MSC_VER) && _MSC_VER < 1900
#define snprintf _snprintf
#endif
```

samolisov:
Yes, I understand, the structure of the classes has been significant
changed. I'll restore the definition ASAP. Also the deleted copy
constructor should be checked (in the skipped commit during the rebase).

samolisov:
@charliegracie I've returned the `snprintf` definition to the `OMRMethodBuilder.hpp` file but as I see there are a few files in the code where `snprintf` also is in use. I pushed `OMRMethodBuilder.hpp` only, let's see how appveyor deals with it. 

Also, the commit 'Remove the delete keyword from the constructor' was restored.

charliegracie:
@genie-omr build all

charliegracie:
@mstoodle, @0xdaryl, @vijaysun-omr or @fjeremic I made some simple changes to `compiler` code to help get this working.  Can one of you take a quick look at that code?

compiler/ilgen/OMRMethodBuilder.hpp
```diff
@@ -28,6 +28,10 @@
 #include "ilgen/IlBuilder.hpp"
 #include "env/TypedAllocator.hpp"
 
+#if defined (_MSC_VER) && _MSC_VER < 1900
```
mstoodle: I know it's more files to change, but I would prefer this code go into the .cpp files, which will also mean moving some code out of OMRMethodBuilder.hpp that uses snprintf. I discussed with Charlie and he has offered to do that work as an immediate follow-up PR so that we can get this one merged.

PR #2631, opened 2018-06-05
Add new memory object types for stats purposes

mpirvu:
This commit adds two new memory object types that will be used
in downstream projects.


fjeremic:
@genie-omr build all

PR #2632, opened 2018-06-05
Add option to disable recompilation when inlined callees get redefined

cathyzhyi:
Add option to disable recompilation when inlined callees get redefined

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -490,6 +490,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableReadMonitors",                 "O\tdisable read monitors",                         SET_OPTION_BIT(TR_DisableReadMonitors), "F"},
    {"disableRecognizedCallTransformer",    "O\tdisable recognized call transformer",           TR::Options::disableOptimization, recognizedCallTransformer, 0, "P"},
    {"disableRecognizedMethods",            "O\tdisable recognized methods",                    SET_OPTION_BIT(TR_DisableRecognizedMethods), "F"},
+   {"disableRecompDueToInlinedMethodRedefinition", "O\tdisable recompilation for method body with patched HCR guard",  SET_OPTION_BIT(TR_DisableRecompDueToInlinedMethodRedefinition), "F"},
    {"disableReducedPriorityForCustomMethodHandleThunks",  "R\tcompile custom MethodHandle invoke exact thunks at the same priority as normal java methods", SET_OPTION_BIT(TR_DisableReducedPriorityForCustomMethodHandleThunks), "F", NOT_IN_SUBSET},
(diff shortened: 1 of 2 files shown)


cathyzhyi:
This changeset depends on https://github.com/eclipse/openj9/pull/2075

andrewcraik:
I don't think this depends on OpenJ9 as it will build and work fine in OMR in isolation - you mean that OpenJ9 will need a matching change to handle the changes proposed in the PR otherwise their build will break right?

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -69,6 +69,7 @@
 #include "ras/Debug.hpp"                        // for TR_Debug
 #include "runtime/Runtime.hpp"
 #include "infra/SimpleRegex.hpp"
+#include "optimizer/TransformUtil.hpp"
```
fjeremic: Could we place this optimizer include in alphabetical order with the other optimizer includes?

andrewcraik:
I don't see anything in this change about HCR guards. This looks to be adding code to a generic OSR path. I don't like that the change is to a generic API and is unconditionally adding this recompilation call when building OpenJ9 when the path could be used for many purposes.

Could the code be put in J9 after it calls induceOSRAfter or be made conditional - preferably by a wrapping API? An alternative would be to insert the call tree in after the OSR infrastructure is added/built in OpenJ9.

fjeremic:
I'll delegate official review to @andrewcraik. Will not merge until you have an approval given this is your area of expertise. I agree though if we can avoid another `J9_PROJECT_SPECIFIC` ifdef that would be amazing.

cathyzhyi:
@andrewcraik  thanks for the suggestion. moved the code to openJ9 after induceOSRAfter.

fjeremic:
> @andrewcraik thanks for the suggestion. moved the code to openJ9 after induceOSRAfter.

Could you please amend the commit message and PR contents to match that of the new change? Do we even need this option to be in OMR? Can it be added to OpenJ9 instead?

cathyzhyi:
> Could you please amend the commit message and PR contents to match that of the new change? Do we even need this option to be in OMR? Can it be added to OpenJ9 instead?

I will need to double check the J9Options part. Tried to move the option to openJ9 but seems the SET_OPTION_BIT  in openJ9 doesn't work.  Will move to openJ9 or open an issue once I come to a conclusion.

andrewcraik:
Yeah option subset processing doesn't work in OpenJ9 so the option will likely need to stay in OMR for now (sadly), but moving the addition of the compile call down is a good thing.

cathyzhyi:
@andrewcraik Hey Andrew, could you help to merge this PR so that the openJ9 part can pass?

fjeremic:
@genie-omr build all

PR #2633, opened 2018-06-05
Fix Werror flag usage in port library

dnakamura:
For files where we stil see Unused result warnings, change behaviour to
disable Werror for that specific warning, rather than disable werror
completely


charliegracie:
No extra testing required since the autotools build is only used in Travis Linux builds

Issue #2603, opened 2018-05-30
GC: Rework object slot visitation in OMR


rwy0717:
# Motivation

Object traversal can be quite complex, especially in dynamic languages.
* Walking can be modal--an object could be composed of multiple regions of data. For example, an object might have header slots, then normal/inline slots, and then side-allocated overflow slots. This does not work well with heterogenous iterators.
* Object layouts change dynamically. Keeping up to date slot maps is infeasibly slow.
* Reference slots are often tagged, sometimes in nontrivial ways. The encodings can even be mixed across slots.

These are all real issues I am running up against while implementing javascript.

# Slot Encoding

To solve the issue with slot encoding, I want to add a new `SlotHandle` type to the glue. `SlotHandle` is a pointer to an object slot containing a reference. `SlotHandle` provides a small API:
* `readReference`
* `writeReference`
* `atomicWriteReference` 
* maybe: `clearReference`

Implementing this class in the glue gives languages the power to implement their own slot encoding. The `SlotHandle` glue replaces the `GC_SlotObject` type in OMR.

Here is an example `SlotHandle` from Om, which builds on Om's Nunboxing `Value` type:

```c++
/// A pointer into a object's slot.
class ValueSlotHandle
{
  public:

	constexpr ValueSlotHandle(Value* slot) : slot_(slot) {}

	void writeReference(Cell *ref) const { slot_->setRef(ref); }

	void atomicWriteReference(Cell *ref) const { assert(0); }

	Cell *readReference() const { return reinterpret_cast<Cell*>(slot_->getRef()); }

  private:
	Value *slot_;
};
```

## Issue: Compressed references

One of the big contentions with this design is, this makes compressed references a concern of the application's object model, and is no longer provided "for free" by the GC. Unfortunately, I believe that crefs can't work without support from the object model, and the added cost of implementing crefs is necessary for the sake of flexibility.

To make implementing basic crefs easy on application writers, I want to provide a `CompressedSlotHandle` in OMR. Applications like Java are free to use this type when implementing their object model on 64bit systems. We could also provide a `BasicSlotHandle` for uncompressed, untagged reference slots. These are just utilities classes that applications can build their object model on top of.

What I don't know is, how can we make it easy for applications to configure the heap for compressed references. Currently, I am imagining some kind of helper for building a heap configuration and choosing a shift value, based on a user-specified max address. Ultimately, I would like it to be feasible for applications to implement an ObjectModel that combines both crefs and tagging.

# Walking Objects

[...]

PR #2635, opened 2018-06-05
Settable slot object instance for object scanners

ktbriggs:
Adds a private _slotObjectPtr pointer and setSlotObject() setter method
to enable scanner client to provide own instance of GC_SlotObject
subclass to receive found referent slots in getNextSlot().

This is a much simpler way to implement and use specialized slot
object containers that extend existing GC_SlotObject than that
described in #1638.

In testing so far (and in production testing) the default slot object
is GC_SlotObject::_slotObject and no changes to other code to override
this using new GC_ObjectScanner::setSlotObject() method are included
with this commit.

Downstream work relating to this will be undertaken in evacuator 
development stream and then possibly migrated to other contexts.

Closes: #1638

### DIFF
--- a/gc/base/ObjectScanner.hpp
+++ b/gc/base/ObjectScanner.hpp
@@ -46,6 +46,13 @@
  * and allows scanning for most objects (<32/64 slots) to complete inline, without
  * requiring a virtual method call (getNextSlotMap()). Otherwise at least one such call
  * is required before getNextSlot() will return NULL.
+ *
+ * The effective slot object pointer points to the instance of a GC_SlotObject or subclass
+ * that will receive found reference slots. This base class provides a default GC_SlotObject
+ * instance for this purpose and presets the effective slot object pointer to point to this
+ * instance. The effective slot pointer can be reset to point to any GC_SlotObject instance
+ * by calling setSlotObject() after scanner instantiating and after resuming an interrupted
+ * object scan.
  */
 class GC_ObjectScanner : public MM_BaseVirtual
 {
@@ -61,7 +68,8 @@ protected:
 	uintptr_t _leafMap;						/**< Bit map of reference slots in object that refernce leaf objects */
 #endif /* defined(OMR_GC_LEAF_BITS) */
 	fomrobject_t *_scanPtr;					/**< Pointer to base of object slots mapped by current _scanMap */
-	GC_SlotObject _slotObject;				/**< Create own SlotObject class to provide output */
+	GC_SlotObject *_slotObjectPtr;			/**< Points to effective slot object for receiving refernce slots */
+	GC_SlotObject _slotObject;				/**< Local GC_SlotObject instance to provide default for _slotObjectPtr */
 	uintptr_t _flags;						/**< Scavenger context flags (scanRoots, scanHeap, ...) */
 	uintptr_t _hotFieldsDescriptor;			/**< Hot fields descriptor for languages that support hot field tracking */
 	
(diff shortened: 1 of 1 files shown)


gc/base/ObjectScanner.hpp
```diff
@@ -61,7 +68,8 @@ class GC_ObjectScanner : public MM_BaseVirtual
 	uintptr_t _leafMap;						/**< Bit map of reference slots in object that refernce leaf objects */
 #endif /* defined(OMR_GC_LEAF_BITS) */
 	fomrobject_t *_scanPtr;					/**< Pointer to base of object slots mapped by current _scanMap */
-	GC_SlotObject _slotObject;				/**< Create own SlotObject class to provide output */
+	GC_SlotObject *_slotObjectPtr;			/**< Points to effective slot object for receiving refernce slots */
+	GC_SlotObject _slotObject;				/**< Local GC_SlotObject instance to provide default for _slotObjectPtr */
```
charliegracie: Why is there an embedded SlotObject if users are expected to set a pointer?

gc/base/ObjectScanner.hpp
```diff
@@ -61,7 +68,8 @@ class GC_ObjectScanner : public MM_BaseVirtual
 	uintptr_t _leafMap;						/**< Bit map of reference slots in object that refernce leaf objects */
 #endif /* defined(OMR_GC_LEAF_BITS) */
 	fomrobject_t *_scanPtr;					/**< Pointer to base of object slots mapped by current _scanMap */
-	GC_SlotObject _slotObject;				/**< Create own SlotObject class to provide output */
+	GC_SlotObject *_slotObjectPtr;			/**< Points to effective slot object for receiving refernce slots */
```
charliegracie: Can you fix the spacing on the comment?

charliegracie:
I would need to see this being used somewhere to understand it better.  Also I would like to complete the discussion in #2603 before we start making other changes to Object iteration as it may radically change object iteration.

gc/base/ObjectScanner.hpp
ktbriggs: That is to maintain backwards compatibility with existing object scanner implementations and use, as well as a default option for use cases to do not require/want to use their own slot object. The default constructor sets the new _slotObjectPtr to point to this instance, the implementation now performs all slot operations using _slotObjectptr. 

This commit adds 8 bytes to the size of the GC_SlotObject. In future we might want to add a GC_SlotObject * parameter to the default GC_ObjectScanner constructor. In that case we can eliminate the GC_SlotObject instance (8 or 16 bytes, compressed or uncompressed). 

I expect GC_ObjectScanner subclasses will not be concerned with this. This change is being introduced to allow GC components that use object scanners to specify specialized slot object instances. This will allow using components to include fields in the slot object for values (eg, object size) that would otherwise need to be calculated >1 times. Eg, component needs object size in >1 context but must calculate it each time because there is no place to cache result of first calculation.

gc/base/ObjectScanner.hpp
ktbriggs: I think what you are seeing is an artifact of html layout (tab size). It looks fine in eclipse:

image

ktbriggs:
Do not merge this. I'm holding off on this as submitted commit will not work with an upcoming change to return a hot field first in getNextSlot() enumeration, which will require specialized slot object to be provided in the constructor (if specialized slot object is warranted). This change should benefit both scavenger and evacuator.

PR #2636, opened 2018-06-05
Allow fabricated loadaddr for classes in AOT

jdmpapin:
More specifically for classes loaded by the bootstrap class loader,
which are unambiguously identifiable at relocation time.

This allows the JIT compiler to make up references to types about which
it has special knowledge. In particular, it's now possible to create a
loadaddr for java/lang/Class "out of thin air," without a corresponding
constant pool entry available.

At a high level, the appropriate relocations are carried out as follows:

  1. Allow TR_ClassAddress ExternalRelocations to be created without a
  constant pool index.

  2. Before grouping into IteratedExternalRelocations, identify these
  and set their kind to the new TR_ArbitraryClassAddress.

  3. Write the relocation record for TR_ArbitraryClassAddress with the
  same data as would be required for TR_ClassPointer.

  4. Relocate TR_ArbitraryClassAddress by finding the class as though
  for TR_ClassPointer, and then patching the code as though for
  TR_ClassAddress (to deal with materialization sequences).

Note that because the compilation has the class pointer, it must have
been through validateArbitraryClass(). In case the class can't be found
at relocation time, the validation record will cause relocation to fail,
so the value used for missing classes is irrelevant.

Power code generation is updated to generate a materialization sequence
and relocation when it encounters a fabricated class symbol reference in
an AOT (ahead of time) compilation. Previously it used an unresolved
snippet, which requires a constant pool index.

### DIFF
--- a/compiler/codegen/OMRAheadOfTimeCompile.hpp
+++ b/compiler/codegen/OMRAheadOfTimeCompile.hpp
@@ -37,6 +37,7 @@ namespace OMR { typedef OMR::AheadOfTimeCompile AheadOfTimeCompileConnector; }
 #include "runtime/Runtime.hpp"      // for TR_ExternalRelocationTargetKind
 
 class TR_Debug;
+namespace TR { class ExternalRelocation; }
 namespace TR { class IteratedExternalRelocation; }
 namespace TR { class AheadOfTimeCompile; }
 
@@ -88,6 +89,11 @@ class OMR_EXTENSIBLE AheadOfTimeCompile
 
    void traceRelocationOffsets(uint8_t *&cursor, int32_t offsetSize, const uint8_t *endOfCurrentRecord, bool isOrderedPair);
 
+   /**
+    * Do project-specific processing of an AOT relocation just before combining
+    * it into an IteratedExternalRelocation.
+    */
+   static void interceptAOTRelocation(TR::ExternalRelocation *relocation) { }
 
    private:
    TR::Compilation *                           _comp;
--- a/compiler/p/codegen/OMRMemoryReference.cpp
+++ b/compiler/p/codegen/OMRMemoryReference.cpp
@@ -1166,6 +1166,33 @@ uint8_t *OMR::Power::MemoryReference::generateBinaryEncoding(TR::Instruction *cu
             }
          else
             {
+            if (comp->compileRelocatableCode()
+                && symbol->isStatic()
+                && symbol->isClassObject())
+               {
+               TR::Node *node = currentInstruction->getNode();
+               TR_RelocationRecordInformation *recordInfo =
+                  (TR_RelocationRecordInformation*)comp->trMemory()->allocateMemory(
+                     sizeof(TR_RelocationRecordInformation),
+                     heapAlloc);
+               recordInfo->data1 = (uintptr_t)self()->getSymbolReference();
+               recordInfo->data2 = node == NULL ? -1 : node->getInlinedSiteIndex();
+               recordInfo->data3 = fixedSequence1;
+               cg->addAOTRelocation(
+                  new (cg->trHeapMemory()) TR::ExternalRelocation(
+                     cursor,
+                     (uint8_t*)recordInfo,
+                     TR_ClassAddress,
+                     cg),
+                  __FILE__,
+                  __LINE__,
+                  node);
+
+               // The relocation will OR its values into the immediates, so
+               // they have to be zero beforehand
+               addr = 0;
+               }
+
             *wcursor = 0x3c000000;                        // lis target, bits0-15
             target->setRegisterFieldRT(wcursor);
             *wcursor |= (addr>>48) & 0x0000ffff;
@@ -1509,6 +1536,21 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
       return;
       }
 
+   bool useUnresSnippetToAvoidRelo = cg->comp()->compileRelocatableCode();
+   if (useUnresSnippetToAvoidRelo)
+      {
+      // The unresolved snippet can't handle fabricated symrefs, so we'll have
+      // to do a relocation. For now, keep using the unresolved snippet for
+      // typical class symrefs even though they could generate relocations too.
+      bool reloIsRequired =
+         symbol->isStatic()
+         && symbol->isClassObject()
+         && ref->getCPIndex() == -1;
+
+      if (reloIsRequired)
+         useUnresSnippetToAvoidRelo = false;
+      }
+
    if (TR::Compiler->target.is64Bit())
       {
       TR::Node *topNode = cg->getCurrentEvaluationTreeTop()->getNode();
@@ -1569,7 +1611,8 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
       //  1. the load hasn't been resolved yet
       //  (otherwise optimizer will remove the ResolveCHK and we can be sure pTOC will contain the resolved address),
       //  2. we don't have a PTOC slot, we must always take the slow path
-      if ((ref->isUnresolved() || cg->comp()->compileRelocatableCode()) &&
+
+      if ((ref->isUnresolved() || useUnresSnippetToAvoidRelo) &&
           (topNode->getOpCodeValue() == TR::ResolveCHK || tocIndex == PTOC_FULL_INDEX))
          {
          snippet = new (cg->trHeapMemory()) TR::UnresolvedDataSnippet(cg, node, ref, isStore, false);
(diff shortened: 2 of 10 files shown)


jdmpapin:
This is the omr part of eclipse/openj9#2083.

andrewcraik:
@vijaysun-omr this class pointer relocation is an important piece of work - the refactoring @charliegracie has highlighted is likely a bigger issue (though one that needs to be addressed) - I think this change is just keeping related code together pending a refactor. Could you take a look.

jdmpapin:
That's a good point about OpenJ9-specific code in OMR. Much of the code here would be hard to move, but I can probably move some of the worst offenders.

If I change `TR::ExternalRelocation::addAOTRelocation()` to call a new extension point, something like 

    // please excuse this name for the moment
    TR::AheadOfTimeCompile::transformAOTRelocation(self());

then I could move `classAddressUsesReloRecordInfo()` and `getClassAddressSymRef()` entirely into OpenJ9.

The assertion in `SymbolReferenceTable::findOrCreateClassSymbol()` is probably also not too hard to move, but that whole method looks somewhat OpenJ9-specific to me. It has two arguments with names starting `cpIndex` referring to constant pools.

andrewcraik:
@vijaysun-omr could you consider this change - the refactoring could be done in a follow-on PR since I think there are other changes that need this new capability waiting to go in.

jdmpapin:
Updated to avoid adding `classAddressUsesReloRecordInfo()` and `getClassAddressSymRef()` to OMR, as described above.

[Diff from the previous version of this commit](https://github.com/eclipse/omr/files/2085145/aot-fab-loadaddr.patch.txt)

vijaysun-omr:
This change addresses classes loaded by bootstrap classloader; are we functionally correct for classes loaded by other classloaders because we will not fabricate a loadaddr of such a class ?

andrewcraik:
line 1315 of the OMRSymbolReferenceTable.cpp is an ASSERT_FATAL which will run even in production and cause a core dump. If anyone ever tries to request a class to be relocated in this way that is not loaded by the bootstrap loader.

compiler/p/codegen/PPCAOTRelocation.cpp
```diff
@@ -39,6 +39,15 @@ void TR::PPCPairedRelocation::mapRelocation(TR::CodeGenerator *cg)
...
       {
+      // The validation and inlined method ExternalRelocations are added after
+      // binary encoding is finished so that their actual relocation records
+      // will end up at the start. This way other relocations can depend on the
+      // appropriate validations already having completed.
+      //
+      // This relocation may also need to depend on previous validations, but
+      // those are already in the AOT relocation list by this point. Add it at
+      // the beginning to maintain the correct relative ordering.
+
```
ymanton: This is a little unclear to me. You've added a mechanism to allow callers of `addAOTRelocation()` to specify where to place the relocation, and this is the only place that uses `TR::AOTRelocationAtFront`, but the comment suggests that it doesn't really need it. Is this mechanism actually necessary? If not, is there any value in introducing it?

jdmpapin: The `ExternalRelocation`s get (roughly) reversed in the process of grouping them into `IteratedExternalRelocation`s, so adding this one at the end (like usual) will cause its relocation record to appear at the very beginning, before the validations. Adding it at the beginning here will put the relocation record at the end.

ymanton: OK, sounds reasonable, though it's a little odd that this is only needed on Power. How does this manage to work on other platforms? Wouldn't they have the same sort of ordering constraints on dependent relocations? Or are class address and validation relocations not dependent on other platforms?

jdmpapin: `ExternalRelocation`s are typically created earlier, during binary encoding. `TR::PPCPairedRelocation`s (which of course are specific to Power) are generated during evaluation instead, so their locations are `TR::Instruction*` rather than pointers into the binary code. They're translated into `ExternalRelocation` here, after binary encoding is complete, but also after the validations have been generated.

ymanton: Fair enough, thanks. Shame it has to be yet another platform-specific oddity.

LGTM.

vijaysun-omr:
@genie-omr build all

compiler/p/codegen/OMRMemoryReference.cpp
```diff
@@ -1609,106 +1652,76 @@ void OMR::Power::MemoryReference::accessStaticItem(TR::Node *node, TR::SymbolRef
...
+            symbol->isClassObject(),
+            "expected a class symbol for symref #%d\n",
+            ref->getReferenceNumber());
+
+         TR_RelocationRecordInformation *recordInfo = ( TR_RelocationRecordInformation *)comp->trMemory()->allocateMemory(sizeof( TR_RelocationRecordInformation), heapAlloc);
+         recordInfo->data1 = (uintptr_t)ref;
+         recordInfo->data2 = node ? (uintptr_t)node->getInlinedSiteIndex() : (uintptr_t)-1;
+         recordInfo->data3 = orderedPairSequence1;
+         staticRelocation = new (cg->trHeapMemory()) TR::PPCPairedRelocation(rel1, NULL, (uint8_t *)recordInfo, TR_ClassAddress, node);
+         cg->getAheadOfTimeCompile()->getRelocationList().push_front(staticRelocation);
```
mstoodle: is there a reason you didn't want to use addAOTRelocation() here ?

compiler/codegen/OMRAheadOfTimeCompile.hpp
```diff
@@ -88,6 +89,11 @@ class OMR_EXTENSIBLE AheadOfTimeCompile
 
    void traceRelocationOffsets(uint8_t *&cursor, int32_t offsetSize, const uint8_t *endOfCurrentRecord, bool isOrderedPair);
 
+   /**
+    * Do project-specific processing of an AOT relocation just before combining
+    * it into an IteratedExternalRelocation.
+    */
+   static void interceptAOTRelocation(TR::ExternalRelocation *relocation) { }
```
mstoodle: Couple of points here:
1) Why is this implemented on AheadOfTimeCompile instead of ExternalRelocation?

2) I think you could have come up with a more informative name here (at least there's a comment to describe it :) ). It's especially weird that an extending project is also expected to call this "interceptAOTRelocation()" back in OMR.

So, maybe ExternalRelocation::finalProjectSpecificHandling() or something like that?

mstoodle:
@jdmpapin I know this has been merged already, but I had a couple of comments that I would like to see addressed as follow-up (either convince me I'm wrong or do some cleanup).

I would also like to echo @charliegracie's comment, in general. I appreciate the willingness you showed to adapt your code changes to reduce the amount of OpenJ9 specific code needed. The more of this refactoring that people can fit in as they visit the code, the faster we'll get these projects in the shape they're supposed to be in. I know not all problems can be entirely fixed when a PR touches some code, but I will always applaud someone who tries to leave the code in better shape than they found it.

compiler/p/codegen/OMRMemoryReference.cpp
jdmpapin: This is pushing onto `getRelocationList()`, not `getAOTRelocationList()`. The relocation will be picked up later from the former list, and its `mapRelocation()` will translate it into an `ExternalRelocation`, which will then be passed to `addAOTRelocation()`.

I know these lists' names aren't very evocative, but I "merely" didn't rename them.

compiler/codegen/OMRAheadOfTimeCompile.hpp
jdmpapin: `AheadOfTimeCompile` is already `OMR_EXTENSIBLE`. I find that the proliferation of files and identically-named classes required for the extensible class pattern make the code harder to work with, so I'd prefer not to convert `ExternalRelocation` if possible. I'm not sure that `ExternalRelocation` is a very natural place for logic either. It feels to me more like a bag of bits. It doesn't even know what type the data is supposed to be, so everything gets cast to `uint8_t*`.

I agree the name could have been better. I was trying to (partially) address @charliegracie's comment quickly, since this change was holding up a bug fix.

You're right as well that there's no reason why a downstream project should have to call the implementation from OMR, since OMR can always just put code next to the call, instead of inside this function. I called OMR from OpenJ9 because of a more general thought that a child project should probably call the parent's implementation unless it specifically knows otherwise.

So what I'd like to do is the following:
- Keep this method on `AheadOfTimeCompile`,
- Document that the default implementation is a no-op,
- Remove the call to OMR from OpenJ9, and
- Rename to sound more like the hook this is, e.g. `beforeCombiningAOTRelocation()`.

Renaming without breaking the build will take three PRs, but the other changes should be easily folded into those.

PR #2637, opened 2018-06-05
Stop versioner from leaving out needed type tests

jdmpapin:
In OpenJ9, versioner could miss necessary type tests, either when
compiling ahead of time, or when determining how to protect a load of
classFromJavaLangClass. These tests are now generated as required.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -7576,19 +7576,81 @@ void TR_LoopVersioner::collectAllExpressionsToBeChecked(List<TR::TreeTop> *nullC
 
             TR_OpaqueClassBlock *cl = NULL;
             if (sig && (len > 0))
-               cl = fe()->getClassFromSignature(sig, len, otherSymRef->getOwningMethod(comp()));
+               {
+               // Currently it's not possible to use this class to eliminate
+               // the type check in AOT. That would require a verification
+               // record to check the subtyping relationship.
+               cl = fe()->getClassFromSignature(sig, len, symRef->getOwningMethod(comp()));
+               }
 
             int32_t otherLen;
             char *otherSig = otherSymRef->getOwningMethod(comp())->classNameOfFieldOrStatic(otherSymRef->getCPIndex(), otherLen);
             dumpOptDetails(comp(), "For node %p len %d other len %d sig %p other sig %p\n", node, len, otherLen, sig, otherSig);
-            if (otherSig)
-               otherClassObject = otherSymRef->getOwningMethod(comp())->getDeclaringClassFromFieldOrStatic(comp(),otherSymRef->getCPIndex());
+            TR::ResolvedMethodSymbol *owningMethodSym = otherSymRef->getOwningMethodSymbol(comp());
+            int32_t classCPI = 0;
+            int32_t fieldCPI = otherSymRef->getCPIndex();
+            instanceOfReqd = false;
+            if (otherSymRef->getSymbol()->isShadow() && fieldCPI >= 0)
+               {
+               TR_ResolvedMethod *owningMethod = owningMethodSym->getResolvedMethod();
+               classCPI = owningMethod->classCPIndexOfFieldOrStatic(fieldCPI);
+               bool aotOK = true;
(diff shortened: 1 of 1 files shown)


andrewcraik:
Jenkins test sanity

jdmpapin:
This requires eclipse/openj9#2082, eclipse/openj9#2083, and #2636.

jdmpapin:
I tested this using [ReproPreventingSymRefReuseForAot.java.txt](https://github.com/eclipse/omr/files/2077497/ReproPreventingSymRefReuseForAot.java.txt), which I ran like so:

    $ java -Xrs -XX:-EnableHCR -Xjit:'limit={*.go(*)*},count=128,disableAsyncCompilation,optLevel=hot,disableOptTransformations={O^O LOOP VERSIONER:  Creating test outside loop for checking if * is conditional*},disableSuffixLogs,compilationThreads=1,traceFull,log=go.log' ReproPreventingSymRefReuseForAot

HCR is disabled here so that the result of the call to `isClass` is straightforwardly the result of `instanceof` after inlining, so that VP can remove the `checkcast`. If it weren't eliminated, the `checkcast` would be versioned and so versioner would happen to generate the necessary type test.

The `disableOptTransformations` option here prevents a transformation whereby versioner duplicates the loop in order to remove an invariant conditional from it. This transformation would again have versioner happen to generate the necessary type test.

For me this program (run as described) reliably produces a segmentation fault, which is prevented by this change.

This test program can also be run with AOT compilation by adding `-Xshareclasses` and `-Xaot:'forceAOT,...'`, where `...` are the options given in `-Xjit`. Doing so engages the new AOT relocation kind `TR_ArbitraryClassAddress` added by #2636 and eclipse/openj9#2083.

andrewcraik:
all dependencies have been merged - @vijaysun-omr can you please review this? the change seems reasonable to me

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -7576,63 +7576,114 @@ void TR_LoopVersioner::collectAllExpressionsToBeChecked(List<TR::TreeTop> *nullC
...
+                        }
+                     break;
+                     }
+                  }
+               }
+#endif
 
-            //the call to findOrCreateClassSymbol is safe even though we pass CPI of -1 since the call above getClassFromSignature is not vetted for AOT
-            if (cl && otherClassObject && (fe()->isInstanceOf(cl, otherClassObject, true) == TR_yes))
+            if (otherClassObject == NULL)
```
vijaysun-omr: Why can we continue with the versioning in this case but not in the j/l/Class case ?

jdmpapin: The typical reason to have `otherClassObject == NULL` here would be that it wasn't set just above, because `otherSymRef` is either:

- an array shadow, which is okay because array accesses are handled elsewhere;
- the VFT, which is okay because all objects have it; or
- a field of a native struct, e.g. `classRomPtrSymbol`, in which case there is no way to generate a test.

In the last case, `firstChild` has to be such that, if it's safe to evaluate, then the result is always a `J9Class` (in this example), or more generally whatever it is that's expected by `node`. This is the same condition required for verifiability in (OpenJ9's) `transformIndirectLoadChain()`:

> For Java object references:
> - Verifiable if it is known to point at an object or null
> - Verified if it is non-null and points to an object that contains the field being loaded
>
> _For C structs pointers:_
> - _Verifiable if it is known to point to a struct of the proper type or null_
> - _Verified trivially by virtue of being verifiable and non null_

(emphasis mine)

But now that you mention it, it's also possible for `getClassFromConstantPool` to return null in AOT even though `isVettedForAOT=true`. I'll broaden the condition for bailing out to include circumstances where we should have gotten the class from the constant pool.

jdmpapin: Updated

vijaysun-omr:
I am okay with the changes now but wondering if you could clarify when getClassFromConstantPool returns NULL in an ahead of time (AOT) compilation.

vijaysun-omr:
Okay so it's relatively rare even in AOT then. Thanks. I will merge this now.

vijaysun-omr:
@genie-omr build all

PR #2640, opened 2018-06-06
Support exception throw from outlined code on X86

0dvictor:
GC Map and Exception Table are now properly setup for outlined instructions on X86, so that exception can now be throwed.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -2016,19 +2016,15 @@ void OMR::X86::CodeGenerator::doBinaryEncoding()
 
    // Create exception table entries for outlined instructions.
    //
-   auto oiIterator = self()->getOutlinedInstructionsList().begin();
-   while (oiIterator != self()->getOutlinedInstructionsList().end())
+   for(auto oiIterator = self()->getOutlinedInstructionsList().begin(); oiIterator != self()->getOutlinedInstructionsList().end(); ++oiIterator)
       {
       uint32_t startOffset = (*oiIterator)->getFirstInstruction()->getBinaryEncoding() - self()->getCodeStart();
       uint32_t endOffset   = (*oiIterator)->getAppendInstruction()->getBinaryEncoding() - self()->getCodeStart();
 
-      TR::Block * block = (*oiIterator)->getBlock();
-      bool      needsETE = (*oiIterator)->getCallNode() && (*oiIterator)->getCallNode()->getSymbolReference()->canCauseGC();
-
-      if (needsETE && block && !block->getExceptionSuccessors().empty())
+      TR::Block* block = (*oiIterator)->getBlock();
+      TR::Node*  node  = (*oiIterator)->getCallNode();
+      if (block && node && !block->getExceptionSuccessors().empty() && node->canGCandExcept())
          block->addExceptionRangeForSnippet(startOffset, endOffset);
-
-      ++oiIterator;
       }
 
 #ifdef J9_PROJECT_SPECIFIC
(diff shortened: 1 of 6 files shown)


compiler/x/codegen/OutlinedInstructions.hpp
```diff
@@ -145,4 +150,15 @@ class TR_OutlinedInstructions
    void setRematerializeVMThread() { _rematerializeVMThread = true; }
 
    };
+
+class TR_OutlinedGenerator
```
fjeremic: This looks like a useful utility which I'd definitely want to make use of in the Z codegen. Can this be moved to a common location and Doxygen documented?

0dvictor: Good suggestion. I've added the documents.
I also would like to make it a common utility; however, I prefer to keep it inside X scope for now. Its implementation is closely tied to X's ```TR_OutlinedInstructions```, and currently contains workaround that avoiding breaking downstream projects in the intermediate stage. For example using the deprecated constructor for now. Once downstream projects adopt it and the deprecated constructor is removed, ```TR_OutlinedGenerator``` will reach to its final stage, and we can move it to a common location.
I'm expecting another 1-2 pull requests to clean up deprecated code.

compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -2016,19 +2016,15 @@ void OMR::X86::CodeGenerator::doBinaryEncoding()
...
+   for(auto oiIterator = self()->getOutlinedInstructionsList().begin(); oiIterator != self()->getOutlinedInstructionsList().end(); ++oiIterator)
       {
       uint32_t startOffset = (*oiIterator)->getFirstInstruction()->getBinaryEncoding() - self()->getCodeStart();
       uint32_t endOffset   = (*oiIterator)->getAppendInstruction()->getBinaryEncoding() - self()->getCodeStart();
 
-      TR::Block * block = (*oiIterator)->getBlock();
-      bool      needsETE = (*oiIterator)->getCallNode() && (*oiIterator)->getCallNode()->getSymbolReference()->canCauseGC();
-
-      if (needsETE && block && !block->getExceptionSuccessors().empty())
+      auto block = (*oiIterator)->getBlock();
```
andrewcraik: I appreciate auto can be super helpful for things like iterators or other complex types - I don't know if I am a fan of loosing the type for simple things like block and node - surely the type helps make this code more clear?

0dvictor: sure, I've updated it with ```TR::Block*``` and ```TR::Node*```.

compiler/x/codegen/OutlinedInstructions.hpp
```diff
@@ -145,4 +152,32 @@ class TR_OutlinedInstructions
...
 
    };
+
+/**
+   @class TR_OutlinedGenerator
+   @brief This class is used to switch to outlined code generation in the manner of the RAII idiom.
+
+   The main use of this class is to generate instruction in the outlined code region, for example, cold paths codes.
+*/
+class TR_OutlinedGenerator
```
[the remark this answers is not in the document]
0dvictor: Sure, I've updated the name.

compiler/x/codegen/OutlinedInstructions.hpp
```diff
@@ -84,6 +86,11 @@ class TR_OutlinedInstructions
    // do in a TR_OutlinedInstructions what you could have done in a normal
    // in-line internal control flow region.
    //
+   // NOTE: use this constructor directly is deprecated, for new development,
+   // use TR_OutlinedGenerator instead.
+   // TR_OutlinedGenerator will setup all necessary information for gc map and
+   // exception table.
+   //
```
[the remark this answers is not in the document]
0dvictor: Good point and I do have a plan to make the constructor `private`. I chose not change it with this PR because downstream projects may currently use this constructor, and I want to give them a time window to adopt this change without break. I'll open a follow-up PR to make this constructor `private` once we are satisfied ourselves that downstream projects have deprecated the usage of this constructor.

0xdaryl:
This is an x86-specific change.  A genie build is not required.

PR #2643, opened 2018-06-06
WIP: Eliminate calls in more appropriate way in VP

liqunl:
Only the handler of the treetop node can decide whether the tree is
needed or not. This commit will fix the call handler in places where it
changes the tree's node to null when it finds the call is not needed
based on the knowledge of the callee.


liqunl:
@andrewcraik Could you review?

compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -386,3 +386,32 @@ OMR::TransformUtil::removeTree(TR::Compilation *comp, TR::TreeTop * tt)
...
+ *
+ * \parm child
+ *    The child for node after the transformation. It's the receiver of the call or any node for a static call.
+ *
+ * \note
+ *    The call node's refcount has to be 1 since the transformation is equivalent to eliminating the call,
+ *    otherwise the resulting node will cause problems in other optimizations.
+ */
+void
+OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::Node* node, TR::TreeTop * anchorTree, TR::Node* child)
```
fjeremic: `TransformUtil` seems to be an extensible class. We should provide API documentation in the header so downstream projects implementing the APIs have fully documented expected behavior. In general, headers are for API inheritance documentation, object files are for implementation detail documentation.

andrewcraik:
my main question is does this belong in transformutils or does it belong in Node? @vijaysun-omr thoughts?

vijaysun-omr:
@liqunl Could you please elaborate on what might go wrong if this change was not done ?

compiler/optimizer/OMRTransformUtil.cpp
liqunl: Done

vijaysun-omr:
@andrewcraik I have never been completely clear on the purpose of the TransformUtils class. It seems to have some static convenience routines that are queried during transformations but quite why routines like fieldShouldBeCompressed or isNoopConversion need to live there and not in Node is'nt obvious to me at least. The rest of the methods in this class seem to be around transforming a node "in place" and if that's the kind of routine that the class was originally created for, then the new routine @liqunl is adding would seem to belong in there too.

liqunl:
Current way of eliminating a call is done by removing the call node, then set current tree's node to null in the call's handler. If the treetop node is a NULLCHK, it'll try to get the reference child from the call node, which will crash since the call node has been removed. One solution will be to add safety check in NULLCHK's handler, but @andrewcraik and I think it's the NULLCHK's or other treetop nodes' handlers instead of the call's handler that can decide whether the tree can be removed.

andrewcraik:
Yes so I think this was originally created for all the transform load chain logic since it was big, complex and not a simple 'node' operation. I'm not clear on why the other helpers ended up here if they are not used in the load chain logic. I guess we can put this new one here for now, but some clear statement like "tools to transform a node in place" would seem a good idea to keep in mind going forward.

liqunl:
Modified the comment to stress this API will transform a call in place.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks okay to me and tests have passed. Merging.

PR #2644, opened 2018-06-06
Support for rewinding code cache allocation pointers

mpirvu:
CodeCache will implement two new routines:
       void setAllocationMark();
       void resetToAllocationMark();
The first one will save the current values of _warmCodeAlloc and _coldCodeAlloc,
while the second will restore these pointers to the saved valued.
resetToAllocationMark also increments the free space in the code cache repository.


0xdaryl:
"Tossing code" isn't a concept that OMR knows about.  If it is something to introduce into OMR (and I think there are some advantages of it) I think we need to architect and define exactly what gets tossed so that the meaning and behaviour is clear.

In your particular PR, you reset ("toss") the allocation pointers, but what about the trampoline reservations that have occurred in the code cache in the meantime?  Are there other scenarios in OMR where resetting these pointers is useful?

I think at this point it probably makes more sense that these changes be sunk into the OpenJ9 extension of CodeCache rather than OMR.  What do you think?

mpirvu:
Thanks Daryl. I just haven't had the time to re-evaluate this yet. Will do so soon.

PR #2646, opened 2018-06-07
new tarokKickoffHeadroomRegionRate

LinHu2016:
- update tarokKickoffHeadroomRegionCount to tarokKickoffHeadroomInBytes
	for allowing the headroom be fraction of region count
	- for OpenJ9 recalculating tarokKickoffHeadroom base
	on the rate of current free memory to avoid GMP kickoff too
	early or too late for some special cases.
	- update related tracepoints
	- step1


charliegracie:
@genie-omr build all

charliegracie:
@amicic thoughts on this change?

amicic:
Looks good. It's just a part of a larger change set, that will allow us for finer granularity (bytes vs regions) control of concurrent kickoff head room in region based collectors, which will be helpful with configurations with low region counts.

charliegracie:
@genie-omr build win

Issue #2567, opened 2018-05-21
Issue with not detecting gcc when ccache is enabled in the adoptopenjdk+openj9 builds


sxa555:
Ref https://github.com/eclipse/openj9/issues/1919#issuecomment-389807665 which shows where this is set up.

It appears that if the C compiler (although in this case it's ccache) is set to have environment variables preceding it then the OMR component then configure gets confused. Haven't checked the logic but it could be that openj9 are only setting it up like that on `s390x`:

Failing job on `Linux/s390x`: https://ci.adoptopenjdk.net/job/openjdk10_openj9_build_s390x_linux/58/consoleFull
Successful job on `Linux/x86-64` where compiler is showing as `cc`: https://ci.adoptopenjdk.net/job/openjdk10_openj9_build_x86-64_linux/54/consoleFull

sxa555:
`./make/autoconf/generated-configure.sh` and `./make/autoconf/build-performance.m4` in the main openjdk repository both appear to add the extra variables to the start of the CCACHE

smlambert:
I'd like to raise the priority of this issue, is someone working on it presently?

DanHeidinga:
@keithc-ca Are you handling this as part of eclipse/openj9#1919?

keithc-ca:
No, I expect this is unrelated to that, but I'd be happy to be wrong.

DanHeidinga:
Thanks Keith for checking.

@dnakamura Can you take a look at this?

dnakamura:
This root issue is that dash (default `/bin/sh` on ubuntu), does not properly handle setting variables  when included from an expansion inside of `eval`. 
Boiled down code:
```sh
CC="var=1 true"
ac_compiler=\$CC
eval "$ac_compiler"
```
^ works in bash but not dash. Currently working on a fix

sxa555:
Fix doesn't appear to be resolving the problem based on the last attempt to build with the fix

irinarada:
@dnakamura - any new updates - this is blocking the 090 binaries testing.

dnakamura:
Opened a PR for JDK9.  working on porting to other versions now

sxa555:
Any update @dnakamura ? It's JDK10 that's first in our priority list.

keithc-ca:
Devin's new fix is merged for JDK10: https://github.com/ibmruntimes/openj9-openjdk-jdk10/pull/44.

sxa555:
[...]

PR #2647, opened 2018-06-07
Call configure using the CONFIG_SHELL variable

dnakamura:
If CONFIG_SHELL not set, default to 'sh' as before

partial solution to #2567 

### DIFF
--- a/run_configure.mk
+++ b/run_configure.mk
@@ -47,6 +47,10 @@ help:
 # Remove all built-in rules
 .SUFFIXES:
 
+
+# Set a default value for CONFIG_SHELL
+CONFIG_SHELL ?= sh
+
 ###
 ### SPEC Specific Configure Arguments
 ###
@@ -128,7 +132,7 @@ clean: clean-environment-variables
 ###
 
 define CONFIGURE_RECIPE
-sh configure --disable-auto-build-flag 'OMRGLUE=$(OMRGLUE)' 'SPEC=$(SPEC)' $(CONFIGURE_ARGS)
+$(CONFIG_SHELL) configure --disable-auto-build-flag 'OMRGLUE=$(OMRGLUE)' 'SPEC=$(SPEC)' $(CONFIGURE_ARGS)
 # Force the timestamps of unchanged files to be updated
 touch $(CONFIGURE_OUTPUT_FILES)
 endef


run_configure.mk
```diff
@@ -47,6 +47,10 @@ help:
 # Remove all built-in rules
 .SUFFIXES:
 
+
+# Set a default value for CONFIG_SHELL
+CONFIG_SHELL ?= sh
```
keithc-ca: Even though it used to use `sh`, perhaps the default should be `bash`?

dnakamura: there is no guarantee that bash will be installed on the machine (or on the path), but posix requires that `sh` exists

charliegracie:
@genie-omr build zos

charliegracie:
Only the Travis and zos builds are currently using the auto tools build story.

run_configure.mk
keithc-ca: We already assume bash is available:

https://github.com/eclipse/omr/blob/916cfcfa682c496a7f8a5a1fb611fb7d0a2d8c0c/ddr_artifacts.mk#L36
https://github.com/eclipse/omr/blob/916cfcfa682c496a7f8a5a1fb611fb7d0a2d8c0c/fvtest/omrtest.mk#L48

dnakamura: That is only DDR. I would argue that we should really try to keep things posix compliant where possible. (Not to mention the risk of breaking builds on some of the more o
bscure systems)

charliegracie: is bash required by autotools (configure)?

dnakamura: It is not. They go to great lengths to work oon just about any posix-ish shell. (although I imagine bash is preferred and gets the most testing)

keithc-ca: The only reason we're having this discussion, though, is because some platforms have `sh` that *isn't* good enough (e.g. Ubuntu 16.04 on x86 or 390x).

dnakamura: only in a somewhat weird edge case. Also, I need to look into it, but i believe that may actually just be a bug in dash

keithc-ca: Ubuntu 18.04 on x86 has the same behavior as 16.04.

dnakamura: right but as far as I can tell, this should work according to the POSIX spec, thus it is logically a bug in the shell

charliegracie:
I am going to merge this as-is.  If we decide in the future that bash or something else is the right shell we can revisit.

PR #2649, opened 2018-06-07
Fix crashes with option TR_DisableKnownObjectTable

nwoeanhinnogaehr:
This commit was written by Omer for JITaaS and is necessary because we use `TR_DisableKnownObjectTable`. Currently the optimizer will crash without a known object table because it does not check for its existence.

Signed-off-by: Noah Weninger <Noah.Weninger@ibm.com>

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1524,7 +1524,8 @@ TR_InlinerBase::createVirtualGuard(
    if (guard->_kind == TR_MutableCallSiteTargetGuard)
       {
       TR::KnownObjectTable *knot = comp()->getOrCreateKnownObjectTable();
-      heuristicTrace(tracer(),"  createVirtualGuard: MutableCallSite.epoch is %p.obj%d (%p.%p)", guard->_mutableCallSiteObject, guard->_mutableCallSiteEpoch, *guard->_mutableCallSiteObject, *knot->getPointerLocation(guard->_mutableCallSiteEpoch));
+      if (knot)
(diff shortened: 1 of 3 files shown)


0xdaryl:
@nwoeanhinnogaehr , can you re-sign the commit yourself and push again?  You'll also have to have signed the ECA.

nwoeanhinnogaehr:
@0xdaryl I resigned the commit, hopefully that convinces the check tool that it is ok. I didn't change the author of the commit. ECA should already have been signed. thanks

0xdaryl:
@genie-omr build all

compiler/optimizer/LocalOpts.cpp
```diff
@@ -7594,6 +7594,9 @@ int32_t TR_InvariantArgumentPreexistence::perform()
 static bool specializeInvokeExactSymbol(TR::Node *callNode, TR::KnownObjectTable::Index receiverIndex, TR::Compilation *comp, TR::Optimization *opt)
```
andrewcraik: The check for the knot is valid, but surely there needs to be some logic to stop calling this if you don't have a valid knot index for the receiverIndex - we'd just be wasting compile time otherwise...

nwoeanhinnogaehr: I noticed the place where this is called is guarded by `receiverInfo.hasKnownObjectIndex()` so perhaps checking for knot here is unnecessary.

nwoeanhinnogaehr: Commit amended

andrewcraik:
So there seems to be something a bit strange happening here - the getOrCreateKnownObjectTable seems to be assumed to work in a number of places and this change is adding guards around those places to test an option. Is the effect of this change to say thou shalt not call getOrCreateKnownObjectTable if a given option is set? Would it be more clean to make the code need to check the result of the getOrCreate call for null and to handle the disable fo the create in the getOrCreate code?

It seems like we are introducing a second convention around an API which will spread control logic farther and wider than was intended...

nwoeanhinnogaehr:
I agree. Updated

0xdaryl:
@genie-omr build all

PR #2651, opened 2018-06-07
Add frontend method getVFTEntry

nwoeanhinnogaehr:
The optimizer should ideally not fetch class fields directly. Factor it out into a frontend method instead (implemented in openj9).

This commit was originally written by Omer Sheikh and is required for JITaaS support.

Depends on https://github.com/eclipse/openj9/pull/2126

Signed-off-by: Noah Weninger <Noah.Weninger@ibm.com>

### DIFF
--- a/compiler/env/OMRClassEnv.hpp
+++ b/compiler/env/OMRClassEnv.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -110,6 +110,15 @@ public:
 
    char *classSignature_DEPRECATED(TR::Compilation *comp, TR_OpaqueClassBlock * clazz, int32_t & length, TR_Memory *) { return NULL; }
    char *classSignature(TR::Compilation *comp, TR_OpaqueClassBlock * clazz, TR_Memory *) { return NULL; }
+
+   /**
+    * Get the virtual function table entry at a specific offset from the class
+    *
+    * @param clazz The RAM class pointer to read from
+    * @param offset An offset into the virtual function table (VFT) of clazz
+    * @return The entry point of the method at the given offset
+    */
+   intptrj_t getVFTEntry(TR::Compilation *comp, TR_OpaqueClassBlock* clazz, int32_t offset);
 
    };
 
--- a/compiler/env/OMRClassEnv.cpp
+++ b/compiler/env/OMRClassEnv.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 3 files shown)


andrewcraik:
OpenJ9 2126 has been merged

andrewcraik:
I am fine with this change so maybe @fjeremic you could take a look?

0xdaryl:
"FrontEnd" is a deprecated interface from OMR's perspective.  Please consider either moving this into the `ClassEnv` structure (which can answer "frontend" questions about a class) or sink it down into OpenJ9 where I think this query will ultimately be used.

compiler/codegen/FrontEnd.hpp
```diff
@@ -200,6 +200,8 @@ class TR_FrontEnd : public TR_Uncopyable
    virtual TR_OpaqueClassBlock * getClassFromMethodBlock(TR_OpaqueMethodBlock *mb);
    virtual int32_t getNewArrayTypeFromClass(TR_OpaqueClassBlock *clazz);
 
+   virtual intptrj_t getVFTEntry(TR_OpaqueClassBlock *clazz, int32_t offset);
```
fjeremic: In addition to @0xdaryl's comments it would be great if we could Doxygen document this new API so that customers of the class know what the expected behavior is.

nwoeanhinnogaehr:
@0xdaryl `ClassEnv` only contains static methods so for JITaaS we need to insert an if statement (with a thread local read!) into each one of them we want to "override". `TR_Frontend` makes this a lot easier as it can just be a virtual call.

If we were to move the call to this new method completely into Openj9, should we do that by moving the `TR_MethodTest` case into J9ValuePropagation.cpp with a `#ifdef J9_PROJECT_SPECIFIC` in its place? I'm not too familiar with the conventions for splitting code among the projects.

andrewcraik:
The methodtest case should not move down to OpenJ9 - method test guards exist in OMR and are a valid cross language construct - the bit that would potentially need to move down would be how to check if the LHS and RHS of the equality in this case are equal

0xdaryl:
You obviously have more requirements on this simple API than this pull request provides.  Please be aware that I'm approaching it from a language-agnostic perspective and what fits in with OMR.

There is a precedent for accomplishing what I think you're doing.  You can add the API to the `ClassEnv` structure that also takes in a `TR::Compilation` object.  You then add your virtual "FrontEnd" query in `TR_J9VMBase` in the OpenJ9 project.  In the OpenJ9 specialization of that `ClassEnv` function you grab the `fej9()` pointer from the Compilation object and call your virtual function from there (see `env/J9ClassEnv.cpp` in OpenJ9 for existing examples).  The call to the J9 ClassEnv implementation should be resolvable at compile-time so you may realize some efficiencies there.

Admittedly, this isn't a great overall solution for some downstream projects.  The FrontEnd refactoring is only partially complete and will need to undergo further evolution in OMR to support the notion of different "frontend" specializations like shared classes and JIT-as-a-service.  I expect to be architecting this in the second half of this year.

nwoeanhinnogaehr:
That makes sense. I updated the commit to use ClassEnv instead.

We can then revert the corresponding change to OpenJ9.

vijaysun-omr:
@0xdaryl are you okay with this commit ?

vijaysun-omr:
@genie-omr build all

compiler/env/OMRClassEnv.hpp
```diff
@@ -111,6 +111,9 @@ class OMR_EXTENSIBLE ClassEnv
    char *classSignature_DEPRECATED(TR::Compilation *comp, TR_OpaqueClassBlock * clazz, int32_t & length, TR_Memory *) { return NULL; }
    char *classSignature(TR::Compilation *comp, TR_OpaqueClassBlock * clazz, TR_Memory *) { return NULL; }
 
+   // Get the virtual function table entry at a specific offset from the class
```
0xdaryl: Please provide a Doxygen comment for the API (i.e., the parameters and return value as well).

nwoeanhinnogaehr:
@0xdaryl updated

0xdaryl:
@genie-omr build all

PR #2653, opened 2018-06-08
Use 32bit object size when compressing the object header

rwy0717:

### DIFF
--- a/example/glue/Object.hpp
+++ b/example/glue/Object.hpp
@@ -32,8 +32,10 @@ typedef uint8_t ObjectFlags;
 
 #if defined(OMR_INTERP_COMPRESSED_OBJECT_HEADER)
 typedef uint32_t RawObjectHeader;
+typedef uint32_t ObjectSize;
 #else
 typedef uintptr_t RawObjectHeader;
+typedef uintptr_t ObjectSize;
 #endif
 
 /**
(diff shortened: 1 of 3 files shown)


example/glue/Object.hpp
```diff
@@ -32,8 +32,10 @@ typedef uint8_t ObjectFlags;
 
 #if defined(OMR_INTERP_COMPRESSED_OBJECT_HEADER)
 typedef uint32_t RawObjectHeader;
+typedef uint32_t ObjectSize;
```
charliegracie: Is ObjectSize technically fj9object_t?

charliegracie:
@genie-omr build all

example/glue/Object.hpp
rwy0717: I'm not sure. `fomrobject_t` is compressed by the crefs flag. In PR #2479 I was trying to keep compressed headers and crefs completely independent.

instead of introducing a new `ObjectSize` type, we can just change `ObjectHeader::assign` to cast a `size_t` to the header width:

```c++
void assign(size_t sizeInBytes, ObjectFlags flags) {
  _value = (RawObjectHeader(sizeInBytes) << SIZE_SHIFT) | flags;
}
```

This is basically what the [original implementation](https://github.com/eclipse/omr/pull/2479/files#diff-95625fb90bd88f837d42321c62fe39a7L76) did, but keeps the compressed header and crefs flags independent.

rwy0717: Unless keeping the flags independent is bogus. After all, why would you want to compress the header without compressing the slots? If we want, I can remove `RawObjectHeader` too, and use `fomrobject_t`.

PR #2655, opened 2018-06-08
Fix warnings and stop ignoring them

keithc-ca:
This way we don't need to deal with differences in the support for `-Wno-error` in various versions of gcc.

* in omrintrospect.c:
  -- record RESUME_FAILURE on failure to destroy a barrier
  -- use strcpy instead of strcat to initialize the /proc path


charliegracie:
@genie-omr build all

PR #2661, opened 2018-06-11
Fix compile errors and warnings

ashu-mehra:
- Avoid using nullptr as it is not supported on older gcc.
nullptr is available since gcc 4.6.
- Use J9_IGNORE_RETURNVAL to call mbtowc() when its return value
is not required.


mstoodle:
@genie-omr build all

charliegracie:
@mstoodle I thought there was another PR for the `nullptr` changes, but I do not see it right now...

charliegracie:
I can not find the PR I was thinking about so I am going to merge.

Issue #1860, opened 2017-10-26
Move use64BitRegsOn32Bit to OMRCPUEnv


0xdaryl:
Move the CodeGenerator query `use64BitRegsOn32Bit` to env/OMRCPUEnv.cpp.

I suggest the API in OMRCPUEnv class look like this:
```
bool use64BitRegsOn32Bit(TR::Compilation *comp)
   {
   return false;
   }
```
CPU architectures can extend this class and provide and overriding method that specializes the function for that architecture.  The Compilation parameter is there in case it is needed.  Presently, only Z requires an extension.

Fix up all uses of the original function.

Uses of this migrated function would therefore look like:
```
TR::Compiler->target.cpu.use64BitRegsOn32Bit(comp)
```

fjeremic:
We're currently looking into potentially folding `use64BitRegsOn32Bit` to always return `true`. The option was first introduced to work around obsolete OSs not preserving the high order 32-bits of 64-bit registers under a 32-bit process when a context switch occurs. This caused weird bugs if data was kept in the upper 32-bits.

Since the Linux distros have dropped 32-bit support on Z and no such limitations exist on z/OS so there is no reason why this option should by default return `true`. We'll likely be folding it away. However rigorous testing needs to be done after such a change as all long evaluators will be affected.

PR #2662, opened 2018-06-12
Disable HPR GRA unless register pressure simulation is enabled

fjeremic:
Register pressure simulation is a prerequisite for HPR GRA because GRA and local RA need to make consistent choices and register pressure simulation is the only part of GRA that is HPR aware. As concrete examples, among others, consider the following:

1. A collected reference coming in as a parameter

In this case GRA needs to know that on 64-bit such a register candidate should not be considered for HPRs, since they are really 32-bit registers. However GRA does not know anything about this. It is the register pressure simulation algorithm [1] that coordinates with the codegen on whether collected references are HPR elligible.

2. A valid HPR candidate is being used as a return value

In this case GRA needs to be aware of the choices local RA will make. Because a value feeds into a return point of a method local RA must enforce that the virtual register corresponding to the return value is 64-bit [3]. Otherwise the high order half of the register may get locally allocated to an HPR spill, and of course this would not be valid. As such GRA must know this fact and it must not globally allocate the return value to an HPR, otherwise we will get into an impossible scenario where local RA is forced to coerce a 64-bit GPR (the return value) into a 32-bit HPR.

The register pressure algorithm is once again aware of these interactions and prevents such values feeding into return points from being globally HPR allocated [2].

[1] https://github.com/eclipse/omr/blob/9d1d8cf3048781bc6d87e6a1079167586cc5aa4d/compiler/codegen/CodeGenRA.cpp#L2691-L2702
[2] https://github.com/eclipse/omr/blob/9d1d8cf3048781bc6d87e6a1079167586cc5aa4d/compiler/codegen/CodeGenRA.cpp#L2889-L2903
[3] https://github.com/eclipse/omr/blob/9d1d8cf3048781bc6d87e6a1079167586cc5aa4d/compiler/z/codegen/ControlFlowEvaluator.cpp#L1098-L1102

### DIFF
--- a/compiler/z/codegen/OMRMachine.cpp
+++ b/compiler/z/codegen/OMRMachine.cpp
@@ -6518,7 +6518,7 @@ OMR::Z::Machine::initializeGlobalRegisterTable()
             }
 
 
-         for (i = linkage->getNumIntegerArgumentRegisters(); i >= 0; i--)
+         for (i = linkage->getNumIntegerArgumentRegisters() - 1; i >= 0; i--)
             p = self()->addGlobalReg(linkage->getIntegerArgumentRegister(i), p);
 
          self()->setLastLinkageGPR(p-1);
--- a/compiler/z/codegen/OMRMachine.cpp
+++ b/compiler/z/codegen/OMRMachine.cpp
@@ -6487,9 +6487,7 @@ OMR::Z::Machine::initializeGlobalRegisterTable()
          }
       else
          {
-      TR::Linkage *linkage = self()->cg()->getS390Linkage();
-         int32_t i;
-         TR::RealRegister::RegNum reg;
+         TR::Linkage *linkage = self()->cg()->getS390Linkage();
          self()->setFirstGlobalGPRRegisterNumber(0);
 
          if (linkage->isZLinuxLinkageType())
@@ -6510,7 +6508,7 @@ OMR::Z::Machine::initializeGlobalRegisterTable()
 
          if (linkage->isZLinuxLinkageType()) // ordering seems to make crashes on zos.
             {
-            for (i = linkage->getNumIntegerArgumentRegisters(); i >= 0; i--)
+            for (uint32_t i = linkage->getNumIntegerArgumentRegisters(); i >= 0; i--)
                {
                if (!linkage->getPreserved(linkage->getIntegerArgumentRegister(i)))
                   p = self()->addGlobalReg(linkage->getIntegerArgumentRegister(i), p);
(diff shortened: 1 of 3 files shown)


fjeremic:
@r30shah could I please get a review?

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -6444,487 +6444,252 @@ OMR::Z::Machine::initializeGlobalRegisterTable()
...
+   self()->setLastLinkageFPR(p-1);
 
-   self()->setLastGlobalGPRRegisterNumber(25);        // Index of last global GPR
-   self()->setLast8BitGlobalGPRRegisterNumber(25);    // Index of last global 8bit Reg
-
-   // additional (forced) restricted regs.
-   // Similar code in TR::S390PrivateLinkage::initS390RealRegisterLinkage() for RA
-
-   for (int32_t i = self()->getFirstGlobalGPRRegisterNumber(); i < self()->getLastGlobalGPRRegisterNumber(); ++i)
+   // Preserved regs, vmthread last
```
r30shah: Looks like this comment is obsolete now as vmThread is added to register file earlier in this function (Line 6545)

fjeremic: You're right. Fixed in a686991

fjeremic:
Marking as WIP: until I've done some additional testing. Committers / reviewers please feel free to proceed with reviews however.

fjeremic:
@vijaysun-omr I'm going to leave this as WIP for about two weeks until I'm back to work on it. My changes in RegiserCandidate.cpp to relax the conditions at which we exclude HPRs is exposing a local RA bug on Linux on Z 32-bit. 

I'm looking into either fixing this bug first to enable this change or circumventing the bug altogether by implementing https://github.com/eclipse/omr/issues/1860#issuecomment-369652978.

I've also validated that reverting the RegisterCandidate.cpp change fixes all the issues and enables this PR to merge but this is taking a step backwards. I'd rather fix the root problem and generate better code.

fjeremic:
@vijaysun-omr the local RA bug mentioned by my previous comment was already hit and has been fixed in #2698. I have verified the fix from #2698 indeed solves the problems I was experiencing. Removing the WIP: tag as this is ready for final review and merge. I'm simultaneously running more testing to do my due diligence.

@genie-omr build all

fjeremic:
Internal testing complete with no errors. This one should be good to go. Travis CI build failed because of infrastructure reasons. See job details for more information.

vijaysun-omr:
Looks a safe change to me. Merging.

PR #2663, opened 2018-06-12
Select hot field for first slot in object scanner

ktbriggs:
The aliasing technique employed by scavenger and evacuator has a higher
likelihood of copying the first slot returned by parent object scanner
close to the parent. Likelihood of close copying then tends to drop as 
subsequent child slots are copied. The OpenJ9 JIT has a mechanism to identify
hot fields, but this capability is used at cross-purposes in scavenger
and not at all in present evacuator implementation. This change forces
object scanners to preselect a hot field as the first slot to be 
returned in the first call to getNextSlot() after the object scanner is
instantiated. 

Other languages that can track hot fields can use this feature
by including a bitmap (descriptor) of hot fields when they instantiate object 
scanners. For all languages, the scope of the hot fields descriptor is limited 
to the first 31 or 63 slots following the object header for 32 or 64-bit 
platforms. Only one hot field is selected for first slot to scan, so if there is
a capability to identify the hottest field, the hot field descriptor should
include 1 bit marking the hottest object only.

This change requires introduction of a new pointer-valued field
_hotScanPtr in GC_ObjectScanner. This will have a greater impact on
system memory footprint for scavenger (which includes an object scanner
in every work packet) than on evacuator (32 object scanner instances per
gc thread).

To test this, I ran 2x32 benchmark trials using evacuator with (hot) and 
without (cold) this change. Application performance **improved by 1.5%** with 
hot field selection. However it must be noted that there is a high degree in variance
in measurements of appplication performane with this benchmark, so this result is 
not statistically significant. 

Similar results may be expected for scavenger, since it uses object scanners in an identical manner. Object scanners are also employed in the standard marking scheme, where this change will incur a small amount of overhead (hot field selection and initialization and a one-time test of a pointer for nullity) that is not recovered by cache-line containment of parent object and hot slot object, since marking uses breadth first traversal to mark referent objects. This will change if marking adopts evacuator's stack-based traversal algorithm, which is intended for future investigation.

### DIFF
--- a/gc/base/ObjectScanner.hpp
+++ b/gc/base/ObjectScanner.hpp
@@ -63,6 +63,7 @@ protected:
 	fomrobject_t *_scanPtr;					/**< Pointer to base of object slots mapped by current _scanMap */
 	GC_SlotObject _slotObject;				/**< Create own SlotObject class to provide output */
 	uintptr_t _flags;						/**< Scavenger context flags (scanRoots, scanHeap, ...) */
+	fomrobject_t *_hotScanPtr;				/** saved value of initial _scaanPtr if there is a hot field */
 	uintptr_t _hotFieldsDescriptor;			/**< Hot fields descriptor for languages that support hot field tracking */
 	
 public:
@@ -107,6 +108,7 @@ protected:
 		, _scanPtr(scanPtr)
 		, _slotObject(env->getOmrVM(), NULL)
 		, _flags(flags | headObjectScanner)
+		, _hotScanPtr(NULL)
 		, _hotFieldsDescriptor(hotFieldsDescriptor)
 	{
 		_typeId = __FUNCTION__;
@@ -135,6 +137,17 @@ protected:
 	MMINLINE void
 	initialize(MM_EnvironmentBase *env)
 	{
+		uintptr_t hotField = (_scanMap & _hotFieldsDescriptor);
+		if (0 != hotField) {
+			uintptr_t zeros = MM_Bits::leadingZeroes(hotField);
+			_slotObject.writeAddressToSlot(&_scanPtr[zeros]);
+			_scanMap &= ~((uintptr_t)1 << zeros);
+#if defined(OMR_GC_LEAF_BITS)
+			_leafMap &= ~((uintptr_t)1 << zeros);
+#endif /* defined(OMR_GC_LEAF_BITS) */
+			_hotScanPtr = _scanPtr;
+			_scanPtr = NULL;
+		}
 	}
 
 	/**
@@ -173,9 +186,6 @@ public:
 	 */
 	MMINLINE bool isLeafObject() { return (0 == _scanMap) && !hasMoreSlots(); }
 
-
-	MMINLINE uintptr_t getHotFieldsDescriptor() { return _hotFieldsDescriptor; }
-
 	/**
 	 * Return base pointer and slot bit map for next block of contiguous slots to be scanned. The
 	 * base pointer must be fomrobject_t-aligned. Bits in the bit map are scanned in order of
@@ -198,7 +208,7 @@ public:
 	{
 		while (NULL != _scanPtr) {
 			/* while there is at least one bit-mapped slot, advance scan ptr to a non-NULL slot or end of map */
-			while ((0 != _scanMap) && ((0 == (1 & _scanMap)) || (0 == *_scanPtr))) {
+			while ((0 != _scanMap) && ((0 == ((uintptr_t)1 & _scanMap)) || (0 == *_scanPtr))) {
 				_scanPtr += 1;
 				_scanMap >>= 1;
 			}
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

charliegracie:
Would you be able to test this with the scavenger to see if it makes the same improvement?

ktbriggs:
Botched the scavenger runs -- the SDK did not include the feature under test. Will rerun tonight. But if the result is negative I will suggest disabling feature for scavenger by passing 0 for hot field descriptor to object scanner constructor when scavenger aliasing algorithm is selected for the collector (this descriptor is unused except in the implementation of feature). Evacuator evidently gets a boost from it, and this will be improved as cache-line containment improves.

ktbriggs:
Rebased (commit 47dd40a).

ktbriggs:
Rebased (commit 47dd40a). Here  are the results for scavenger 2x32 SPECjbb2015. This was done on the same 8 cpu/8g fyre VM as the evacuator trials. Heap size in each case was 1g, to avoid global gcs.

||cold|hot|t-score|
|--:|---:|--:|------:|
|stdev|1536.61|1729.18|0.39|
|mean|31148.27|30800.00||
|sum|1027893|1016400||

This is almost a negative mirror image of evacuator results. I think this reflects the low cache-line containment rate for scavenger (~10% vs ~20% for x86-64) - close hot copies are not frequent enough to offset the _hotScanPtr initialization overhead. I will disable this feature for scavenger and retain it for evacuator. Disablement will mean scavenger will carriy a NULL _hotScanPtr field in its object scanners and incurs one extra NULL pointer check per object.

charliegracie:
I am hesitant to grow this structure as it will impact ALL GC policies.  Can this be refactored so that it only affects Evacuator (ie put it under sigh... a build flag)?

amicic:
Wondering if this is doable without the extra field (_hotScanPtr )....

One idea: in getNextSlot()  to replace the outer while condition that uses _scanPtr to rely on hasMoreSlots() instead. Initially, if there is a hot field, hasMoreSlots() would return false, and _scanPtr  would point at the real first scan slot, as normally it would. Then, at the end of  getNextSlot(), instead of checking _hotScanPtr, to rely on _hotFieldsDescriptor and if non-0 do the work of finding the hot field pointer, resetting noMoreSlots flag and initializing and returning slotObject (but no need to touch _scanPtr).

Then, there is a chance with some code shuffling to 'merge' that new hasMoreSlots() with the existing one in the middle of the outer while loop (perhaps an aggressive compiler might be able to do it even without shuffling)

If for some reason I missed something and this idea is not doable, I'm questioning the existence of field _hotFieldsDescriptor (and its getter method). Does not seem to be used, at least not in OMR and OpenJ9.

ktbriggs:
Sorry I haven't been paying enough attention to this. I'm returning to it now because it could provide a win for both scavenger and evacuator. Older generations of scavenger used object slot iterators that walked backwards through slots. The reason for this is that the last slot is closest to the copy head in the scan cache/frame that will receive the copied referent. This feature can be used to select the last slot as the hot slot, which would result in increased cache-line containment for small object reference arcs,. 

I would like to do this for evacuator at least, but it will provide some benefit for scavenger in OpenJ9 Java so I would suggest allowing it without a build flag. It will impact scavenger/evacuator only as object scanners are presently used only in standard marking scheme. If @charlie_gracie you think a build flag is required I will withdraw this PR and implement in evacuator stream #2557 for experimentation, using hot field map to preselect last slot are first slot to scan. Otherwise, I will rebase to pick this up in #2557 and implement for scavenger. 

Jon and I are working on replacing object slot iterators with scanners and can disable this feature simply by using default (0) hot field map in scanner instantiation, and enabling it when ready. As @amicic  points out object scanners do not otherwise use the hot field map, and no other component requires hot field information from object scanners, so using default (0) hot field map is a safe and effective way to manage the impact of this change. 

Aleks I think this implementation is optimal. The overhead for hot field preselection is to cook the initial slot map and preset scanptr and hotfieldptr plus an additional NULL pointer check per object.

charliegracie:
The Scavenger walks backwards through an object to perform object scanning? For mixed objects or arrays?

ktbriggs:
Scavenger circa 2016 used object _iterators_  to walk backwards. I did not understand the purpose of this when I started refactoring Scavenger into OMR. Scavenger today uses object _scanners_, which walk forward. This feature that I am proposing will force mixed object scanners to select the slot closest to the end of the (mixed) object to copy first, thereby increasing likelihood of cache-line containment of copied referent with referring slot.

charliegracie:
I think you are confused about the object iterators the Scavenger used to use.  The mixed object iterator walked the object forwards but the indexable (array) iterator walked backwards due to changes made for arraylets which accidentally walked the array backwards.

I am not sure a mixed object should be forced to mark/copy-foward the last slot in the object first as that is an arbitrary decision to make a single object closest to the referencing slot. That may slightly improve GC times (maybe...) but it is just as likely to cause mutator degradation as it is to help.

Do you have any data to back that this helps application throughput? When running gencon? When running Evacuator?

PR #2666, opened 2018-06-12
Implement breakOnEntry on the Z codegen

fjeremic:
Given the analysis done by our friends over at the V8 project the
software breakpoint in GDB as per [1] is encoded as 0x0001. We add a
new BREAK pseudo instruction and generate this software breakpoint so
users can consistently use the breakOnEntry option cross platform.

[1] https://github.com/v8/v8/blob/a062708467a313d64818b5f8a0ac37dae8775959/src/s390/assembler-s390.cc#L611-L614

### DIFF
--- a/compiler/z/codegen/S390GenerateInstructions.cpp
+++ b/compiler/z/codegen/S390GenerateInstructions.cpp
@@ -2131,23 +2131,40 @@ TR::Instruction * generateVSIInstruction(
    }
 
 /************************************************************ Misc Instructions ************************************************************/
-TR::Instruction *
-generateS390PseudoInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op, TR::Node * n, TR::Node * fenceNode, TR::Instruction * preced)
+TR::Instruction*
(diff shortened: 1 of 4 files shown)


compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2131,23 +2131,33 @@ TR::Instruction * generateVSIInstruction(
...
+      cursor = new (INSN_HEAP) TR::S390PseudoInstruction(op, n, fenceNode, cg);
+      }
 
+   switch (op)
+      {
+      case TR::InstOpCode::PROC:
+         {
+         if (cg->comp()->getOption(TR_EntryBreakPoints))
+            {
+            cursor = new (INSN_HEAP) TR::S390EInstruction(TR::InstOpCode::BREAK, n, cursor, cg);
```
0xdaryl: Apologies for not being 100% familiar with the order in which the Z codegen emits things.  Will this breakpoint appear after the prologue then?  Is that the behaviour you want?

fjeremic: Yes it should appear after the prologue. You're right it is probably best if the break got inserted before. I'll swap the order of generation.

fjeremic: Fixed in 1359fee

0xdaryl:
@genie-omr build zlinux,zos

PR #2668, opened 2018-06-13
Fix vlog options

r30shah:
verbose={} options was broken causing segmentation fault. This commit fixes that.


r30shah:
@fjeremic  While using verbose={compil*} options to get vlog, I found that it is broken with recent builds. We were just missing a `,` in the initialized array. 
Can you please review this change?

fjeremic:
@genie-omr build all

PR #2669, opened 2018-06-14
Fix compile errors on 32-bit Windows

keithc-ca:
Explicit casts are needed on 32-bit platforms where size_t and intptr_t are both narrower than int64_t.


charliegracie:
@genie-omr build all

charliegracie:
hmmm DDR must not be enabled in our CI 32 bit windows build

PR #2670, opened 2018-06-14
Replace PassThrough with its child in VP

liqunl:
Replace PassThrough node with its child in its non-NullCheck parents

### DIFF
--- a/compiler/optimizer/ValuePropagationCommon.cpp
+++ b/compiler/optimizer/ValuePropagationCommon.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -4109,6 +4109,7 @@ TR::LocalValuePropagation::optDetailString() const throw()
 
 void OMR::ValuePropagation::launchNode(TR::Node *node, TR::Node *parent, int32_t whichChild)
    {
+   TR::Node *newNode = node;
    if (node && node->getVisitCount() != _visitCount)
       {
       int32_t valueNumber = getValueNumber(node);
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik @jdmpapin Could you review?

compiler/optimizer/VPHandlersCommon.cpp
```diff
@@ -69,8 +70,25 @@ TR::Node *constrainChildren(OMR::ValuePropagation *vp, TR::Node *node)
...
    for (int32_t i = node->getNumChildren()-1; i >= 0; i--)
       {
+      TR::Node* child = node->getChild(i);
       vp->setCurrentParent(node);
-      vp->launchNode(node->getChild(i), node, i);
+      vp->launchNode(child, node, i);
+
+      // Replace PassThrough with its child
+      if (child->getOpCodeValue() == TR::PassThrough
+          && !node->getOpCode().isNullCheck())
```
jdmpapin: It's possible for `launchNode()` to replace the child, in which case `child` wouldn't be right anymore.

I thought `constrainChildren()` might be able to get away without having to test for null checks, but if not, perhaps this is better done inside `launchNode()`.

Should we use `performTransformation()` here?

compiler/optimizer/VPHandlersCommon.cpp
```diff
@@ -69,8 +70,25 @@ TR::Node *constrainChildren(OMR::ValuePropagation *vp, TR::Node *node)
...
+         {
+         TR::Node* grandChild = child->getFirstChild();
+         node->setAndIncChild(i, grandChild);
+
+         if (child->getReferenceCount() == 1)
+            child->recursivelyDecReferenceCount();
+         else
+            child->decReferenceCount();
```
jdmpapin: Couldn't this conditional be replaced with an unconditional call to `recursivelyDecReferenceCount()`?

liqunl:
@jdmpapin Changes have been updated.

liqunl:
@andrewcraik Change has passed pbuild.

andrewcraik:
LGTM - @vijaysun-omr over to you

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks ok to me. Will merge if tests pass.

PR #2674, opened 2018-06-14
Use the right way to check if an argument has fixed type

liqunl:
We should ask `classIsFixed()` on the argument to check if it has fixed class constraint.

### DIFF
--- a/compiler/optimizer/LocalOpts.cpp
+++ b/compiler/optimizer/LocalOpts.cpp
@@ -7442,9 +7442,15 @@ int32_t TR_InvariantArgumentPreexistence::perform()
 
                parmInfo.setSymbol(p);
                TR_OpaqueClassBlock *clazz = arg->getClass();
-               if (clazz)
+               TR_OpaqueClassBlock *clazzFromMethod = fe()->getClassFromSignature(sig, len, feMethod);
+               TR_ASSERT(!clazz ||
+                         !clazzFromMethod ||
+                         clazz == clazzFromMethod ||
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik Could you review?

andrewcraik:
Looks good to me - @vijaysun-omr can you take a look?

compiler/optimizer/LocalOpts.cpp
```diff
@@ -7462,18 +7461,12 @@ int32_t TR_InvariantArgumentPreexistence::perform()
                   }
                else
                   {
-                  clazz = fe()->getClassFromSignature(sig, len, feMethod);
+                  clazz = clazz ? clazz : fe()->getClassFromSignature(sig, len, feMethod);
```
vijaysun-omr: Should you be taking the more specific of clazz or getClassFromSignature ? i.e. maybe both are non-null but one provides "better" type info than the other (?)

liqunl: Inliner will try to refine argument types and put the result in argInfo. The type from argInfo should be the same as or more specific than the type from method symbol.

andrewcraik: should we have an assert for that then maybe just to verify it is true and warn ourselves is something breaks that invariant unintentionally later on?

liqunl: Assertion is added

vijaysun-omr:
@genie-omr build all

PR #2676, opened 2018-06-15
WIP: Om in OMR

rwy0717:


charliegracie:
@rwy0717 can you rebase?  @youngar can you open an Eclipse CQ for this?

example/glue/LanguageVMGlue.c
```diff
@@ -23,7 +23,8 @@
 #include "omrvm.h"
 
 omr_error_t
-OMR_Glue_BindCurrentThread(OMR_VM *omrVM, const char *threadName, OMR_VMThread **omrVMThread)
+
```
charliegracie: what happened to the function name?

gc/CMakeLists.txt
```diff
@@ -167,6 +167,9 @@ add_library(omrgc STATIC
 	stats/ScavengerStats.cpp # TODO only compile if scavenger or VLHGC. Is this actually used by VLHGC?
 	stats/SweepStats.cpp
 
+	stats/ConcurrentGCStats.cpp
+	stats/CompactStats.cpp
```
charliegracie: Why did these have to move out of the proper location for their compile flags?

om/CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+	TEST OMR_GC # AND OMR_JITBUILDER
+	MESSAGE "Om requires the OMR_GC and OMR_JITBUILDER to be enabled"
+)
+
+omr_assert(FATAL_ERROR
+	TEST NOT OMR_EXAMPLE
+	MESSAGE "The OMR example code is temporarily incompatible with Om"
+)
+
+set(CMAKE_CXX_STANDARD 11 PARENT_SCOPE)
```
charliegracie: Can this be removed?  We should not be forcing C++11 as not all compiler support this.  Extended support is allowed as long as all of the features are available on Visual Studio 10 and xlC 11 (I think it is xlC 11)

rwy0717: This isn't the cleanest way to set the standard, but should be okay. This will cause cmake to choose the best flag for the compiler toolchain, to get as close as possible to c++11.  Cmake will still succeed even If the compiler doesn't support c++11. To treat this as an error, you'd turn `CMAKE_CXX_STANDARD_REQUIRED` on, but I'm not doing that here.

Lately, I have been working to restrict om to the language or stdlib features supported by our target specs (toolchain & target platform). There is more work to do here, as om has never actually been compiled for most specs (and actually used to target c++14).

On older versions of cmake, or on some toolchains, cmake is bad at picking the flag. It's not something I'm worrying about now, but will probably go wrong if we ever port Om to eg xlc.

Eventually, I would like to set the standard flag for the omr_om/omrgc target only, but that requires that the omrgc target is already defined. Due to the way I've structured the top level cmakelists.txt, that isn't yet possible. This probably needs a cleaner solution. Possibly, an OMR_GC_CXX_STANDARD variable?

gc/CMakeLists.txt
rwy0717: IIRC The stats aren't ifdef'd properly, this was giving me a compile error when compact or concurrent were disabled.

example/glue/LanguageVMGlue.c
rwy0717: d'oh. not sure. will fix.

rwy0717:
I will periodically rebase this branch to keep it up to date :+1:

rwy0717:
There are a few small utilities in om that could be generally useful. I placed them under the `OMR::Infra` namespace. Is there any interest in making a new "infra" component for shared c++ utilities? Should this code just go under the util directory? Or should I just roll it into om?

The headers are [here](https://github.com/b9org/omr/tree/0a2e6965b007f109452d447a1a7ece7f8ef9ee35/om/include/OMR/Infra).

charliegracie:
I would roll those headers into OM.

@rwy0717 any insight on getting it compiling?

0xdaryl:
Presumably before this is ultimately merged the commits will be squashed down into a much smaller subset?

rwy0717:
closing to contribute in smaller PRs.

PR #2678, opened 2018-06-15
Include OMR build flags in DDR macros

mikezhang1234567890:
Change DDR getmacros script to pick up undef'd flags commented out by
configure.
Treat omrcfg.h as containing both values and flags so unset flags show
up.

Reviewer: @keithc-ca

### DIFF
--- a/ddr/tools/getmacros
+++ b/ddr/tools/getmacros
@@ -30,6 +30,7 @@ default_regex="@ddr_namespace: *default"
 map_to_type_regex="@ddr_namespace: *map_to_type="
 flag_define_regex="^ *# *define +(${id}) *($|//|/\*)"
 flag_undef_regex="^ *# *undef +(${id}) *($|//|/\*)"
+flag_undef_regex_comment="^/\* #undef (${id}) \*/"
 value_define_regex="^ *# *define +(${id}) +[^ /].*($|//|/\*)"
 include_guard_regex=".*_[Hh]([Pp][Pp])?_? *($|//|/\*)"
 
@@ -162,6 +163,7 @@ process_one_policy_file() {
 		-e '/@ddr_namespace:/p' \
 		-e '/^ *# *define /p' \
 		-e '/^ *# *undef /p' \
+		-e '/^\/\* #undef /p' \
 	| while read -r line; do
 		if [[ ${line} =~ ${default_regex} ]]; then
 			set_default_namespaces
@@ -173,7 +175,7 @@ process_one_policy_file() {
 		elif [[ ${addFlags} -ne 0 && ${line} =~ ${flag_define_regex} ]]; then
 			# Print a define flag as a constant of 1.
 			define_flag_macro "${BASH_REMATCH[1]}"
-		elif [[ ${addFlags} -ne 0 && ${line} =~ ${flag_undef_regex} ]]; then
+		elif [[ ${addFlags} -ne 0 && ${line} =~ ${flag_undef_regex} || ${line} =~ ${flag_undef_regex_comment} ]]; then
 			# Print an undef flag as a constant of 0.
 			undef_flag_macro "${BASH_REMATCH[1]}"
 		fi
--- a/include_core/omrcfg.h.in
+++ b/include_core/omrcfg.h.in
@@ -29,7 +29,10 @@
 #if !defined(OMRCFG_H_)
 #define OMRCFG_H_
 
-/* @ddr_namespace: map_to_type=OmrBuildFlags */
+/**
+ * @ddr_namespace: map_to_type=OmrBuildFlags
+ * @ddr_options: valuesandbuildflags
+ */
 
 #undef OMR_GC
 #undef OMR_JIT
--- a/omrcfg.CMakeTemplate.h
+++ b/omrcfg.CMakeTemplate.h
@@ -28,7 +28,10 @@
 #if !defined(OMRCFG_H_)
 #define OMRCFG_H_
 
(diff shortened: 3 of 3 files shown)


mikezhang1234567890:
After discussion with Keith, since `configure` is a generated file, need to move the code to pick up undefs somewhere else, probably into the DDR getmacros script.

configure
```diff
@@ -8632,7 +8632,7 @@ cat >>$CONFIG_STATUS <<_ACEOF || ac_write_fail=1
     # Replace #undef with comments.  This is necessary, for example,
     # in the case of _POSIX_SOURCE, which is predefined and required
     # on some systems where configure will not decide to define it.
-    if (defundef == "undef") {
+    if (defundef == "undef" && macro !~ /^OMR/) {
```
keithc-ca: This is a derived artifact: this change will be overwritten if `HAS_AUTOCONF=1` is in the build environment.

mikezhang1234567890:
@keithc-ca can you take another look at this? I've moved the searching for undefs to the DDR `getmacros` script.

ddr/tools/getmacros
```diff
@@ -30,6 +30,7 @@ default_regex="@ddr_namespace: *default"
 map_to_type_regex="@ddr_namespace: *map_to_type="
 flag_define_regex="^ *# *define +(${id}) *($|//|/\*)"
 flag_undef_regex="^ *# *undef +(${id}) *($|//|/\*)"
+flag_undef_regex_comment="^/\* # *undef +(${id}) \*/"
```
keithc-ca: We want to be stricter here: don't accept spaces between `#` and `undef`.

ddr/tools/getmacros
```diff
@@ -162,6 +163,7 @@ process_one_policy_file() {
 		-e '/@ddr_namespace:/p' \
 		-e '/^ *# *define /p' \
 		-e '/^ *# *undef /p' \
+		-e '\|^/\* # *undef |p' \
```
keithc-ca: I don't see how this can work; the pattern must be enclosed with two `/` characters.
Also we don't want to match spaces between `#` and `undef`.

ddr/tools/getmacros
mikezhang1234567890: Alright, should the other regexes be changed too?

ddr/tools/getmacros
mikezhang1234567890: From https://www.gnu.org/software/sed/manual/sed.html#Regexp-Addresses

```
\%regexp%
(The % may be replaced by any other single character.)

This also matches the regular expression regexp, but allows one to use a different delimiter than /. 
```

ddr/tools/getmacros
keithc-ca: No, the others are correct as is - they recognize patterns in regular code.

ddr/tools/getmacros
keithc-ca: Thanks, I wasn't aware of that usage. I wouldn't be surprised it that's a GNU extension even though that page doesn't describe it as such. I think we should use `/` and escape as necessary.

youngar: As far as I can tell, it seems like it is indeed an extension to the posix standard for sed: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

Interestingly, it seems the posix standard suggests but does not specify that sed uses `/` as a delimiter for addresses.

It looks like bsd sed supports this behavior: https://www.freebsd.org/cgi/man.cgi?query=sed&sektion=&n=1

ddr/tools/getmacros
```diff
@@ -30,6 +30,7 @@ default_regex="@ddr_namespace: *default"
 map_to_type_regex="@ddr_namespace: *map_to_type="
 flag_define_regex="^ *# *define +(${id}) *($|//|/\*)"
 flag_undef_regex="^ *# *undef +(${id}) *($|//|/\*)"
+flag_undef_regex_comment="^/\* #undef (${id}) \*/"
```
keithc-ca: This should be `"^/\/\* #undef (${id}) \*\//"` (note the leading and trailing (escaped) `/`).

mikezhang1234567890: That matches a literal `//*` at the start of a line, and a `*//` at the end of our match.

This is the result of my test(on AIX), expanding out $id:
```
-bash-4.4$ pattern="^/\/\* #undef ([A-Za-z_][A-Za-z0-9_]*) \*\//"
-bash-4.4$ echo "$pattern"
^/\/\* #undef ([A-Za-z_][A-Za-z0-9_]*) \*\//
-bash-4.4$ string="/* #undef FLAG */"
-bash-4.4$ echo "$string"
/* #undef FLAG */
-bash-4.4$ if [[ $string =~ $pattern ]]; then echo "found"; else echo "not found"; fi;
not found
-bash-4.4$ string="//* #undef FLAG *//"
-bash-4.4$ if [[ $string =~ $pattern ]]; then echo "found"; else echo "not found"; fi;
found
-bash-4.4$ string="/* #undef FLAG */"
-bash-4.4$ pattern="^/\* #undef ([A-Za-z_][A-Za-z0-9_]*) \*/"
-bash-4.4$ if [[ $string =~ $pattern ]]; then echo "found"; else echo "not found"; fi;
found
```

keithc-ca: My apologies, I thought that pattern was for `sed`, not the shell: the current `flag_undef_regex_comment` is good.

keithc-ca:
@youngar can you please do a final review and merge if you're satisfied? Thanks.

youngar:
@genie-omr build all

PR #2681, opened 2018-06-17
Document how to Program OMR IL

dibyendumajumdar:
This is an initial commit in what I hope will create a user guide for programmers working with or intending to work with OMR IL. The document is aimed more towards front-end language developers who need to generate OMR IL.


dibyendumajumdar:
Will retry

PR #2683, opened 2018-06-18
Re-order checks in fixupForwardedSlot

amicic:
Check if Concurrent Marking is in STW mode before attempting to
recoginize/fixup forwarded object


charliegracie:
@genie-omr build all

PR #2685, opened 2018-06-19
Improved the performance of omrtime_current_time_millis

gaolibj:


gaolibj:
Commit with wrong user id.

PR #2690, opened 2018-06-21
Disable Cgroup Support for z/TPF

jjohnst-us:
z/TPF does not support cgroup.  The cgroup support could
potentially be supported in the future depending on priorities.

Signed-off-by: James Johnston <jjohnst@us.ibm.com>


charliegracie:
@genie-omr build all

PR #2692, opened 2018-06-22
Bug fixes, cleanups, and improvements for external relocations and AOT

ymanton:
These changes primarily make external relocations a little less AOT-specific (to pave the way for their use in other contexts) by renaming `addAOTRelocation()`. The old API is kept intact so that downstream consumers (OpenJ9) will still build until they are fixed. Also included is a change to the x86 codegen to allow for finer grained control of which kinds of statics are relocatable. Finally, a few bugs are fixed.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -250,7 +250,7 @@ OMR::CodeGenerator::CodeGenerator() :
      _collectedSpillList(getTypedAllocator<TR_BackingStore*>(TR::comp()->allocator())),
      _allSpillList(getTypedAllocator<TR_BackingStore*>(TR::comp()->allocator())),
      _relocationList(getTypedAllocator<TR::Relocation*>(TR::comp()->allocator())),
-     _aotRelocationList(getTypedAllocator<TR::Relocation*>(TR::comp()->allocator())),
+     _externalRelocationList(getTypedAllocator<TR::Relocation*>(TR::comp()->allocator())),
      _staticRelocationList(_compilation->allocator()),
      _breakPointList(getTypedAllocator<uint8_t*>(TR::comp()->allocator())),
      _jniCallSites(getTypedAllocator<TR_Pair<TR_ResolvedMethod,TR::Instruction> *>(TR::comp()->allocator())),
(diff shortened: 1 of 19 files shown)


compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -3158,37 +3158,47 @@ void OMR::CodeGenerator::addRelocation(TR::Relocation *r)
 
 void OMR::CodeGenerator::addAOTRelocation(TR::Relocation *r, const char *generatingFileName, uintptr_t generatingLineNumber, TR::Node *node, TR::AOTRelocationPositionRequest where)
    {
-   TR_ASSERT(generatingFileName, "AOT relocation location has improper NULL filename specified");
+   self()->addExternalRelocation(r, generatingFileName, generatingLineNumber, node, static_cast<TR::ExternalRelocationPositionRequest>(where));
+   }
+
+void OMR::CodeGenerator::addAOTRelocation(TR::Relocation *r, TR::RelocationDebugInfo* info, TR::AOTRelocationPositionRequest where)
```
mstoodle: are you keeping these around until downstream projects can be changed to use the new name?

mstoodle: Ah! just saw the comment in the commit message. Other than the IP check, I have no concerns with this PR.

mstoodle:
Coming up to speed now.... GitHub just shows he check on the last commit of the PR, but it's actually the third commit (3d32ba8) it's complaining about. @ymanton, since Omer was an IBM employee at the time, I think you'll have to re-submit that commit under your own name (since the email Omer used is probably no longer valid).

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -233,10 +233,15 @@ class TR_ClobberEvalData
 
 namespace TR
    {
+   enum ExternalRelocationPositionRequest
+      {
+      ExternalRelocationAtFront,
+      ExternalRelocationAtBack,
+      };
    enum AOTRelocationPositionRequest
```
fjeremic: Will we be cleaning up the duplicates afterward?

ymanton: Yes, the methods and enums with AOT in their names will be removed once downstream is fixed.

ymanton:
Fixed IP check and rebased.

ymanton:
Ping.

ymanton:
@mstoodle This is ready.

mstoodle:
@ymanton the @ mention works better than a random ping :) . I'm reviewing it now.

mstoodle:
@genie-omr build all

PR #2693, opened 2018-06-22
Remove .gitattributes file as it breaks travis builds

DanHeidinga:
Travis builds have been updated to newer versions of git which recognize
the worktree-encoding attribute. This means the current files are
invalid as they force the encoding of all files as ebcdic.

Other builds, and developers, will be broken when they update to
newer versions of git.

This solution won't work long term. Removing the file now so we
can unblock builds while thinking about the right long term solution.

This backs out part of #2568


DanHeidinga:
Similar PR for OpenJ9 under discussion: https://github.com/eclipse/openj9/pull/2254

charliegracie:
This PR should not remove the binary file ones.  #2695 has been opened to just remove the EBCDIC bindings.

DanHeidinga:
Closing this as #2695 is a better solution

PR #2695, opened 2018-06-22
Clear gitattributes

dnakamura:
git 2.18 starts honoring working-tree-encoding. This causes
means files will be checked out as ebcdic.


dnakamura:
@genie-omr build zos

charliegracie:
@genie-omr build all

PR #2696, opened 2018-06-25
Implicit master for CS STW

amicic:
Stop-the-World Concurrent Scavenger phases of a cycle will now use
implicit master thread (mutator thread that triggered the phase through
Allocation Failure).

MasterGCThread class is modified to now support both implicit and
explicit master GC thread for STW phases. The later (for now) is still
used for Balanced GC (in OpenJ9).

Another optional behavior is whether to acquire or not VM access during
concurrent phase. The default is not to (again, still used by Balanced
GC, during concurrent marking), while CS acquires it (it must because it
moves objects).

The benefit of having implicit master is simpler start/end of a STW
increment, since there is no need for use of assume/relinquish exclusive
access API, which in various ways complicates some corner cases of VM
access synchornization (like hand-off or simultanious safe-access and
exclusive VM access ) with also complex GC scenarios (where two
different STW increments/cycle are executed back to back, like percolate
GC).

Another changes (and simplification) is that the transition from first
STW phase and concurrent phase of a CS cycle will not have the deferred
VM access 'protocol'. Simply, at the end of STW, exclusive VM access
will be released completely (including master GC thread releasing VM
access, instead of deferring the release). Then master GC thread will
simply acquire VM access prior to starting the concurrent phase. No
special tricks or APIs are used. Master GC thread will continue to keep
VM access during the whole concurrent phase, although in future for
fairness, it may occassionally try to release it even before finishing
the phase. So, there is now a gap between first STW and concurrent
phase, where master GC holds neither exclussive VM access nor thread VM
access, and other parties are ok to acquire exclusive VM access (or
safeAccess) at that time (for example to walk heap and replace classes).
The same gap also exists (used to exist even before), between the end of
the concurrent phase and start of the second (last) STW phase. 

Clearly, concurrent phase of CS will continue to use explict master GC
thread.

Another change is a fix to a startup scenario, where explicit master
thread during attach-thread procedure acts as a mutator, and if (rarely)
happens during already active concurrent CS phase, the GC specific
buffers (like barrier ones) would be populated but never flushed (at the
time of flushing the thread would already turn into a GC thread and not
treated properly as a mutator). The fix is to explicitely flush these
buffers after attaching and before really becoming a GC thread. This
used to be a problem only with slave threads, but now that STW phase
runs with implicit master, similar problem occurs while brining up the
explicit master. The old fix is removed, and this new fix is now applied
on both master and slave thread.

### DIFF
--- a/gc/base/EnvironmentBase.hpp
+++ b/gc/base/EnvironmentBase.hpp
@@ -521,11 +521,6 @@ public:
 	bool isInlineTLHAllocateEnabled() { return _delegate.isInlineTLHAllocateEnabled(); }
 #endif /* OMR_GC_THREAD_LOCAL_HEAP */
 
-	/* When a GC thread is created _workUnitIndex is initialized to 0, when a task is started _workUnitIndex is reset to 1.
-	 * _workUnitIndex can be used to check if a GC thread has done any work since it was created.
-	 */
-	MMINLINE bool isGCSlaveThreadActivated() { return _workUnitIndex != 0; }
-
 	MMINLINE uintptr_t getWorkUnitIndex() { return _workUnitIndex; }
 	MMINLINE uintptr_t getWorkUnitToHandle() { return _workUnitToHandle; }
 	MMINLINE void setWorkUnitToHandle(uintptr_t workUnitToHandle) { _workUnitToHandle = workUnitToHandle; }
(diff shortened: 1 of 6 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@amicic can you update a spelling mistake in the commit message?  "efault" instead of "default"? Also part way through there is a "STw" instead of "STW".

charliegracie:
@dmitripivkine can you give this a quick review please?

charliegracie:
@genie-omr build all

charliegracie:
I am merging without waiting for the AppVeyor build as they seem to taking over a day to complete.  This has successfully passed the Windows testing on the OMR CI.

PR #2698, opened 2018-06-25
Spill target register if it is the only one available

dchopra001:
If targetReg's state is blocked and no other registers
are available for spilling, then bypass the blocked state to
spill the blocked register, then reverse spill it back when needed.

### DIFF
--- a/compiler/z/codegen/OMRMachine.cpp
+++ b/compiler/z/codegen/OMRMachine.cpp
@@ -5327,40 +5327,70 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction                       
          }
       else
          {
-         self()->cg()->traceRegAssigned(currentTargetVirtual, spareReg);
+         // if spareReg is still NULL, then it means we did not find any free Reg, and no
+         // other register is available to spill. So we must spill the target register.
 
-         // virtual register is not assigned yet, copy register
-         cursor = self()->registerCopy(currentInstruction, currentTargetVirtualRK, targetRegister, spareReg, self()->cg(), instFlags);
-
-         spareReg->setState(TR::RealRegister::Assigned);
-         spareReg->setAssignedRegister(currentTargetVirtual);
-         currentTargetVirtual->setAssignedRegister(spareReg);
-
-         if (enableHighWordRA && currentTargetVirtual->is64BitReg())
+         // In general if a register is blocked, the virtual register
+         // associated with a target register would have been assigned as part of that instruction.
+         // So we would not do this (i.e. spill the blocked register). However, in this case all
+         // pre-assigned registers in the dependency are blocked before individiual dependencies are coerced
+         // into their respective target registers (inside TR_S390RegisterDependencyGroup::assignRegisters((..)).
+         // In this situation it's possible that the target register of a register dependency is currently
+         // blocked (as in here) and occupied by another virtual register and that there is no other available
+         // register to free.
+         // Hence, we must spill the blocked target reg as it is the only available candidate.
+         if (spareReg == NULL)
             {
-            //TR_ASSERT(targetRegisterHW->getState() == TR::RealRegister::Blocked,
-            //        "currentTargetVirtual is blocked and is fullsize, but the HW is not blocked?");
+            self()->cg()->setRegisterAssignmentFlag(TR_RegisterSpilled);
+            virtualRegister->block();
 
-            spareReg->getHighWordRegister()->setState(TR::RealRegister::Assigned);
-            spareReg->getHighWordRegister()->setAssignedRegister(currentTargetVirtual);
-            targetRegister->getLowWordRegister()->setState(TR::RealRegister::Unlatched);
-            targetRegister->getLowWordRegister()->setAssignedRegister(NULL);
-            targetRegisterHW->setState(TR::RealRegister::Unlatched);
-            targetRegisterHW->setAssignedRegister(NULL);
-            }
-         if (virtualRegister->getTotalUseCount() != virtualRegister->getFutureUseCount())
-            {
-            self()->cg()->setRegisterAssignmentFlag(TR_RegisterReloaded);
-            self()->reverseSpillState(currentInstruction, virtualRegister, targetRegister);
+            self()->spillRegister(currentInstruction, currentTargetVirtual);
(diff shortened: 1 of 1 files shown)


dchopra001:
@joransiu @r30shah

0xdaryl:
@genie-omr build zlinux,zos

joransiu:
This is an interesting scenario that we did not handle previously.  Normally, if a register is blocked, the virtual register associated with a target register would have been assigned as part of that instruction previously.  As such, we would not want to spill a blocked register.  In this case, all pre-assigned registers in the dependency are blocked by [1], before individual dependencies are coerced to their respective target register.   As such, it is possible that the target register of a reg dep is currently blocked and occupied by another virtual register... and that there is no available register to free.  

[1] https://github.com/eclipse/omr/blob/master/compiler/z/codegen/OMRRegisterDependency.cpp#L1055-L1090

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -5327,40 +5327,58 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction
...
+            virtualRegister->setAssignedRegister(targetRegister);
+            targetRegister->setState(TR::RealRegister::Assigned);
+
+            if (targetRegister->isHighWordRegister() && currentTargetVirtual->is64BitReg() &&
+                targetRegister->getRegisterNumber() == spareReg->getHighWordRegister()->getRegisterNumber())
+               doNotRegCopy = true;
+            virtualRegister->unblock();
             }
-         else
+         else
```
joransiu: nit. extra space at the end of line.

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -5327,40 +5327,58 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction
...
-
-         // virtual register is not assigned yet, copy register
-         cursor = self()->registerCopy(currentInstruction, currentTargetVirtualRK, targetRegister, spareReg, self()->cg(), instFlags);
-
-         spareReg->setState(TR::RealRegister::Assigned);
-         spareReg->setAssignedRegister(currentTargetVirtual);
-         currentTargetVirtual->setAssignedRegister(spareReg);
-
-         if (enableHighWordRA && currentTargetVirtual->is64BitReg())
+         if(spareReg == NULL)
```
joransiu: nit.  Missing a space between `if` and `(`.   This code to handle the `spareReg == NULL` case probably deserves a comment.

compiler/z/codegen/OMRMachine.cpp
dchopra001: Fixed here: 
bb2f4d9

compiler/z/codegen/OMRMachine.cpp
dchopra001: Space fixed here: 
bb2f4d9

Comment: 
35913c9d6f78176b5617110005966b9613d4efdd

joransiu:
lgtm

dchopra001:
Squashed + rebased. This is ready to merge!

0xdaryl:
@genie-omr build zlinux,zos

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -5327,40 +5327,70 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction
...
-            targetRegisterHW->setState(TR::RealRegister::Unlatched);
-            targetRegisterHW->setAssignedRegister(NULL);
-            }
-         if (virtualRegister->getTotalUseCount() != virtualRegister->getFutureUseCount())
-            {
-            self()->cg()->setRegisterAssignmentFlag(TR_RegisterReloaded);
-            self()->reverseSpillState(currentInstruction, virtualRegister, targetRegister);
+            if (targetRegister->isHighWordRegister() && currentTargetVirtual->is64BitReg() &&
+                targetRegister->getRegisterNumber() == spareReg->getHighWordRegister()->getRegisterNumber())
+               doNotRegCopy = true;
```
fjeremic: @dchopra001 this looks incorrect. `spareReg` is `NULL` so we have a null pointer dereference here. What is the intention of this `if` block? I do not see any path following this block which leads to a use of `doNotRegCopy`.

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -5327,40 +5327,70 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction
...
+         // pre-assigned registers in the dependency are blocked before individiual dependencies are coerced
+         // into their respective target registers (inside TR_S390RegisterDependencyGroup::assignRegisters((..)).
+         // In this situation it's possible that the target register of a register dependency is currently
+         // blocked (as in here) and occupied by another virtual register and that there is no other available
+         // register to free.
+         // Hence, we must spill the blocked target reg as it is the only available candidate.
+         if (spareReg == NULL)
+            {
+            self()->cg()->setRegisterAssignmentFlag(TR_RegisterSpilled);
+            virtualRegister->block();
```
fjeremic: `currentAssignedRegister` is `NULL` at this point so why do we need the block `virtualRegister` here? It should not be associated with any real register so blocking doesn't do anything.

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -5327,40 +5327,70 @@ OMR::Z::Machine::coerceRegisterAssignment(TR::Instruction
...
-            targetRegisterHW->setAssignedRegister(NULL);
-            }
-         if (virtualRegister->getTotalUseCount() != virtualRegister->getFutureUseCount())
-            {
-            self()->cg()->setRegisterAssignmentFlag(TR_RegisterReloaded);
-            self()->reverseSpillState(currentInstruction, virtualRegister, targetRegister);
+            if (targetRegister->isHighWordRegister() && currentTargetVirtual->is64BitReg() &&
+                targetRegister->getRegisterNumber() == spareReg->getHighWordRegister()->getRegisterNumber())
+               doNotRegCopy = true;
+            virtualRegister->unblock();
```
fjeremic: Is this correct? `virtualRegister` was just assigned and the dependency has been satisfied, shouldn't we block the register instead?

PR #2700, opened 2018-06-25
More VMThread reclamation cleanup

0xdaryl:
* Misc. VMThread reclamation cleanup
* Remove VMThread reclamation parameter from generateLabelInstruction


0xdaryl:
Note to reviewers:  a stub `generateLabelInstruction` that accepts a VMThread boolean was left in the code to assist the removal of this parameter from downstream projects.  These redirecting implementations will be removed when downstream projects have been updated with this change.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Safe cleanup of VM thread logic that is'nt needed.

PR #2706, opened 2018-06-26
Add missing const to std::map allocator type

rwy0717:

### DIFF
--- a/compiler/compile/OMRCompilation.hpp
+++ b/compiler/compile/OMRCompilation.hpp
@@ -675,7 +675,7 @@ public:
 
    void setIlVerifier(TR::IlVerifier *ilVerifier) { _ilVerifier = ilVerifier; }
 
-   typedef std::pair<const void *, TR::DebugCounterBase *> DebugCounterEntry;
+   typedef std::pair<const void * const, TR::DebugCounterBase *> DebugCounterEntry;
    typedef TR::typed_allocator<DebugCounterEntry, TR::Allocator> DebugCounterMapAllocator;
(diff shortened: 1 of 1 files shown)


compiler/compile/OMRCompilation.hpp
```diff
@@ -675,7 +675,7 @@ class OMR_EXTENSIBLE Compilation
 
    void setIlVerifier(TR::IlVerifier *ilVerifier) { _ilVerifier = ilVerifier; }
 
-   typedef std::pair<const void *, TR::DebugCounterBase *> DebugCounterEntry;
+   typedef std::pair<const void * const, TR::DebugCounterBase *> DebugCounterEntry;
    typedef TR::typed_allocator<DebugCounterEntry, TR::Allocator> DebugCounterMapAllocator;
    typedef std::map<const void *, TR::DebugCounterBase *, std::less<const void *>, DebugCounterMapAllocator> DebugCounterMap;
```
dsouzai: Does the `const void * const` also have to be added to the `std::map` and `std:;less` template params?

0xdaryl:
@genie-omr build all

compiler/compile/OMRCompilation.hpp
rwy0717: The synopsis of `std::map` is:

```c++
template<
    class Key,
    class T,
    class Compare = std::less<Key>,
    class Allocator = std::allocator<std::pair<const Key, T> >
> class map;
```
It's only the allocator that works with `const Key`. You don't need to change map's `Key` or `Compare` unless you really want the map to use constant keys as input.

PR #2709, opened 2018-06-27
Disable Compiler Time Profiler and ATOE Mapping for zTPF

jjohnst-us:
z/TPF currently doesn't support the Compiler Time Profiling
like Linux.  Potentially in the future depending on priorities.
Also, z/TPF gets a compilation failure in ILValidationRules
because of ATOE translation mapping for remove.  The guard
macro prevents this for z/TPF only.


jjohnst-us:
@fjeremic 
Minor updates, similar updates exist already in the code.
tia

fjeremic:
@genie-omr build all

PR #2711, opened 2018-06-27
Fix VP handlers for [il]abs when low bound is MIN

jdmpapin:
For values not constrained to be constant, if the low bound was the
minimum value of the type (hereafter referred to as MIN), VP (value
propagation) would negate it, which is undefined behaviour. In the best
case, the program would continue using -MIN == MIN, which is correct for
two's complement, but which causes VP to produce a bad constraint.

For a positive high bound, the resulting constraint would exclude all
negative values, even though MIN is possible.

For a non-positive high bound, the resulting constraint's low and high
bounds would be out of order (i.e. high == MIN < low). This constraint
is not only incorrect but malformed, and it is liable to cause VP to
misbehave.

The handlers now deal separately with the possibility that the child is
the minimum value, effectively treating an input constraint [MIN, high]
instead as {MIN} union [MIN+1, high]. The range is processed as before,
but if MIN was possible, then the result is a VPMergedConstraints that
includes MIN as well.

Additionally, these handlers will now cast to unsigned before negating
when computing the constraint for a constant, in order to ensure that
there is no signed overflow.

Finally, if the child is unconstrained, the result is always known to be
either MIN or non-negative.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -7134,25 +7134,36 @@ TR::Node *constrainIabs(OMR::ValuePropagation *vp, TR::Node *node)
 
    bool isGlobal;
    TR::VPConstraint *child = vp->getConstraint(node->getFirstChild(), isGlobal);
-   if (child)
+   if (child == NULL)
       {
-      if (child->asIntConst())
+      vp->addGlobalConstraint(
+         node,
+         TR::VPMergedConstraints::create(
+            vp,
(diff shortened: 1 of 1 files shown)


jdmpapin:
ping?

vijaysun-omr:
I would like a review from @andrewcraik please.

andrewcraik:
A very interesting edge case indeed - one I had not considered either. @vijaysun-omr change looks good to me.

PR #2712, opened 2018-06-27
Fix DDR blob_reader so it respects byte ordering

### LINKED ISSUE #2697: DDR blob_reader doesn't respect byte ordering
A blob created on z Linux is unreadable on x or p; conversely, blobs created on little-endian systems are unreadable on big-endian systems. The format of a blob specifically encodes the byte ordering of the originating system: blob_reader should act accordingly.

keithc-ca:
Fixes #2697.


charliegracie:
@genie-omr build all

PR #2715, opened 2018-06-28
Fix incorrect usage of TR::Options

harryyu1994:
- during compilation, the per-method TR::Options should be used instead of the global one
- mostly happens at the Optimizer and Code Generation stage
- simplify comp()->getOptions()->getOption(TR_XXX) to comp()->getOptions(TR_XXX), they are equivalent
- switch _options usage to options in OMRCompilation constructor

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -326,14 +326,14 @@ OMR::Compilation::Compilation(
    // while the PersistentMethodInfo is allocated during codegen creation
    // Access to this list must be performed with assumptionTableMutex in hand
    //
-   if (!TR::Options::getCmdLineOptions()->getOption(TR_DisableFastAssumptionReclamation))
+   if (!options.getOption(TR_DisableFastAssumptionReclamation))
       _metadataAssumptionList = new (m->trPersistentMemory()) TR::SentinelRuntimeAssumption();
 #endif
 
    //Random fields must be set before allocating codegen
    _primaryRandom = new (m->trHeapMemory()) TR_RandomGenerator(options.getRandomSeed());
    _adhocRandom = new (m->trHeapMemory()) TR_RandomGenerator(options.getRandomSeed());
-   if (_options->getOption(TR_RandomSeedSignatureHash))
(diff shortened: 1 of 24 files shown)


compiler/compile/OMRCompilation.cpp
```diff
@@ -326,13 +326,13 @@ OMR::Compilation::Compilation(
    // while the PersistentMethodInfo is allocated during codegen creation
    // Access to this list must be performed with assumptionTableMutex in hand
    //
-   if (!TR::Options::getCmdLineOptions()->getOption(TR_DisableFastAssumptionReclamation))
+   if (!_options->getOption(TR_DisableFastAssumptionReclamation))
```
mpirvu: The parameter `options` is the same thing here.

compiler/compile/OMRCompilation.cpp
```diff
@@ -326,13 +326,13 @@ OMR::Compilation::Compilation(
...
       _metadataAssumptionList = new (m->trPersistentMemory()) TR::SentinelRuntimeAssumption();
 #endif
 
    //Random fields must be set before allocating codegen
-   _primaryRandom = new (m->trHeapMemory()) TR_RandomGenerator(options.getRandomSeed());
-   _adhocRandom = new (m->trHeapMemory()) TR_RandomGenerator(options.getRandomSeed());
+   _primaryRandom = new (m->trHeapMemory()) TR_RandomGenerator(_options->getRandomSeed());
```
mpirvu: The code was good as it was because `options` is the per compilation Options object

harryyu1994: yeah I changed it because I saw many more instances of `_options` than `options` so I changed everything to `_options` to avoid confusion. Do you prefer using `options` or just want this part of the code unchanged?

mpirvu: Yes, I would prefer to use the parameter `options`

harryyu1994:
Hi @vijaysun-omr, could you help merge this change as well as https://github.com/eclipse/openj9/pull/2290, I think they are ready to go in, thanks a lot.

vijaysun-omr:
@genie-omr build all

PR #2717, opened 2018-06-28
Initial changes to refactor snapshot at the beginning barrier

ebadyano:
SATB barrier is used in metronome gc policy. We would like to enable it
for other gc polices under option -Xgc:snapshotAtTheBeginningBarrier.
These are initial changes to add the option and start decoupling barrier
code from metronome gc.

### DIFF
--- a/gc/base/Configuration.hpp
+++ b/gc/base/Configuration.hpp
@@ -93,6 +93,22 @@ public:
 	bool initializeHeapRegionDescriptor(MM_EnvironmentBase *env, MM_HeapRegionDescriptor *region) { return _delegate.initializeHeapRegionDescriptorExtension(env, region); }
 	void teardownHeapRegionDescriptor(MM_EnvironmentBase *env, MM_HeapRegionDescriptor *region) { _delegate.teardownHeapRegionDescriptorExtension(env, region); }
 
+	MMINLINE MM_GCWriteBarrierType getBarrierType() { return _writeBarrierType; }
+
+	MMINLINE bool
+	isSnapshotAtTheBeginningBarrierEnabled()
+	{
+		return (getBarrierType() == gc_modron_wrtbar_satb_and_oldcheck) ||
+				(getBarrierType() == gc_modron_wrtbar_satb);
+	}
+
+	MMINLINE bool
+	isIncrementalUpdateBarrierEnabled()
+	{
+		return (getBarrierType() == gc_modron_wrtbar_cardmark_and_oldcheck) ||
+				(getBarrierType() == gc_modron_wrtbar_cardmark);
+	}
+
 	/**
 	 * Delegated method to determine when to start tracking heap fragmentation, which should be inhibited
 	 * until the heap has grown to a stable operational size.
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -70,6 +70,7 @@ class MM_ReferenceChainWalkerMarkMap;
 class MM_RememberedSetCardBucket;
 #if defined(OMR_GC_STACCATO)
 class MM_RememberedSetWorkPackets;
+class MM_RememberedSetSATB;
 #endif /* OMR_GC_STACCATO */
 #if defined(OMR_GC_MODRON_SCAVENGER)
 class MM_Scavenger;
@@ -162,6 +163,7 @@ public:
 	bool _forceOptionConcurrentMark; /**< true if Concurrent Mark option is forced in command line */
 	bool _forceOptionConcurrentSweep; /**< true if Concurrent Sweep option is forced in command line */
 	bool _forceOptionLargeObjectArea; /**< true if Large Object Area option is forced in command line */
+	bool _forceOptionWriteBarrierSATB; /**< Set with -Xgc:snapshotAtTheBeginningBarrier */
 
 	MM_ConfigurationOptions()
 		: MM_BaseNonVirtual()
@@ -170,6 +172,7 @@ public:
 		, _forceOptionConcurrentMark(false)
 		, _forceOptionConcurrentSweep(false)
 		, _forceOptionLargeObjectArea(false)
+		, _forceOptionWriteBarrierSATB(false)
 	{
 		_typeId = __FUNCTION__;
 	}
@@ -250,7 +253,9 @@ public:
 	MM_SublistPool rememberedSet;
 #endif /* OMR_GC_MODRON_SCAVENGER */
 #if defined(OMR_GC_STACCATO)
+	// TODO: SATB remove staccatoRememberedSet after initial SATB changes are merged
 	MM_RememberedSetWorkPackets* staccatoRememberedSet; /**< The Staccato remembered set used for the write barrier */
+	MM_RememberedSetSATB* sATBBarrierRememberedSet; /**< The snapshot at the beginning barrier remembered set used for the write barrier */
 #endif /* OMR_GC_STACCATO */
 	ModronLnrlOptions lnrlOptions;
 
@@ -1231,6 +1236,7 @@ public:
 		, gcmetadataPageFlags(OMRPORT_VMEM_PAGE_FLAG_NOT_USED)
 #if defined(OMR_GC_STACCATO)
 		, staccatoRememberedSet(NULL)
+		, sATBBarrierRememberedSet(NULL)
 #endif /* OMR_GC_STACCATO */
 
 		, heapBaseForBarrierRange0(NULL)
--- a/gc/base/MarkingScheme.cpp
+++ b/gc/base/MarkingScheme.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -31,12 +31,17 @@
 #include "ConcurrentGC.hpp"
 #include "ConcurrentGCStats.hpp"
 #endif /* defined(OMR_GC_MODRON_CONCURRENT_MARK) */
+#include "Configuration.hpp"
 #include "EnvironmentBase.hpp"
 #include "GCExtensionsBase.hpp"
 #include "Heap.hpp"
 #include "MarkMap.hpp"
 #include "MarkingScheme.hpp"
 #include "Task.hpp"
+#if defined(OMR_GC_REALTIME)
+#include "WorkPacketsSATB.hpp"
+#endif /* defined(OMR_GC_REALTIME) */
+#include "RememberedSetSATB.hpp"
 #if defined(OMR_GC_MODRON_CONCURRENT_MARK)
 #include "WorkPacketsConcurrent.hpp"
 #else
@@ -387,11 +392,18 @@ MM_WorkPackets *
 MM_MarkingScheme::createWorkPackets(MM_EnvironmentBase *env)
 {
 	MM_WorkPackets *workPackets = NULL;
-
 	if (_extensions->isConcurrentMarkEnabled()) {
+		if (_extensions->configuration->isSnapshotAtTheBeginningBarrierEnabled()) {
+#if defined(OMR_GC_REALTIME)
+			MM_WorkPacketsSATB *workPacketsSATB = MM_WorkPacketsSATB::newInstance(env);
+			_extensions->sATBBarrierRememberedSet = MM_RememberedSetSATB::newInstance(env, workPacketsSATB);
+			workPackets = workPacketsSATB;
+#endif /* defined(OMR_GC_REALTIME) */
+		} else {
 #if defined OMR_GC_MODRON_CONCURRENT_MARK
-		workPackets = MM_WorkPacketsConcurrent::newInstance(env);
+			workPackets = MM_WorkPacketsConcurrent::newInstance(env);
 #endif /* defined OMR_GC_MODRON_CONCURRENT_MARK */
+		}
 	} else {
 		workPackets = MM_WorkPacketsStandard::newInstance(env);
 	}
--- a/gc/base/standard/ConcurrentGC.cpp
+++ b/gc/base/standard/ConcurrentGC.cpp
@@ -53,6 +53,7 @@
 #if defined(OMR_GC_CONCURRENT_SWEEP)
 #include "ConcurrentSweepScheme.hpp"
 #endif /* OMR_GC_CONCURRENT_SWEEP */
+#include "Configuration.hpp"
 #include "CycleState.hpp"
 #include "Debug.hpp"
 #include "Dispatcher.hpp"
@@ -75,6 +76,10 @@
 #include "SublistPuddle.hpp"
 #include "SublistSlotIterator.hpp"
 #include "WorkPacketsConcurrent.hpp"
+
+#if defined(OMR_GC_REALTIME)
+#include "RememberedSetSATB.hpp"
+#endif /* defined(OMR_GC_REALTIME) */
 
 typedef struct ConHelperThreadInfo {
 	OMR_VM *omrVM;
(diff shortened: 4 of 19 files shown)


ebadyano:
This is work in progress. Opened the pull request to get feedback and comments. 
@amicic @dmitripivkine

gc/base/MarkingScheme.cpp
```diff
@@ -392,6 +393,8 @@ MM_MarkingScheme::createWorkPackets(MM_EnvironmentBase *env)
 #if defined OMR_GC_MODRON_CONCURRENT_MARK
 		workPackets = MM_WorkPacketsConcurrent::newInstance(env);
 #endif /* defined OMR_GC_MODRON_CONCURRENT_MARK */
+	} else if (_extensions->isSnapshotAtTheBeginningBarrierEnabled()) {
+		workPackets = MM_WorkPacketsSATBBarrier::newInstance(env);
```
amicic: Something about ext->concurrentMark flag should change. Either, it should should be renamed to something incrementalUpdateBarrier is enabled or keep it true even when SATB barrier is enabled (in which case isSATBBarrierEnebled flag could be upgraded to writeBarrierMode enum, which would then make this is similar code have another else branch asserting the mode is one of the two possible).

Dont have a strong opinion which of the two approaches (let's see if anybody else has an opinion), but bottom line concurrentMark has fairly generic name and somewhat implies it could mean SATB being enabled is also a posiblity.

I understand this would affect not only but a few other places in the code (including OpenJ9)

gc/base/standard/ConcurrentGC.hpp
```diff
@@ -381,10 +360,26 @@ class MM_ConcurrentGC : public MM_ParallelGlobalGC
 	ConHelperRequest getConHelperRequest(MM_EnvironmentBase *env);
 
 protected:
-	bool initialize(MM_EnvironmentBase *env);
+	virtual bool initialize(MM_EnvironmentBase *env);
```
amicic: Don't make this virtual (it would be making like a C++ constructor virtual) - none or rarely they are virtual in the code base . At may be not as elegant, but at the call site you should specifically instantiate/initialize SATB or IU variant of this class. There actually probably only called by their specifc newInstance call, so just even removing 'virtual' might work!?

gc/base/standard/RememberedSetWorkPackets.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+#include "modron.h"
+#include "Debug.hpp"
+#include "RememberedSetWorkPackets.hpp"
+#include "WorkPackets.hpp"
+
+/**
+ * Object creation and destruction 
+ *
+ */
```
amicic: This could be under REALTIME build flag - up for a debate.

gc/base/standard/WorkPacketsSATBBarrier.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+#include "j9.h"
+#include "j9protos.h"
+#include "j9port.h"
+#include "omrthread.h"
+
+#include "Debug.hpp"
+#include "GCExtensions.hpp"
+#include "IncrementalOverflow.hpp"
+
```
amicic: This should definitely be under REALTIME build flag.

gc/base/GCExtensionsBase.hpp
```diff
@@ -250,7 +250,7 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
 	MM_SublistPool rememberedSet;
 #endif /* OMR_GC_MODRON_SCAVENGER */
 #if defined(OMR_GC_STACCATO)
-	MM_RememberedSetWorkPackets* staccatoRememberedSet; /**< The Staccato remembered set used for the write barrier */
+	MM_RememberedSetWorkPackets* sATBBarrierRememberedSet; /**< The snapshot at the beginning barrier remembered set used for the write barrier */
```
amicic: To make this change self contained (not dependent  related OpenJ9 changes), you should keep both staccatoRememberedSet and sATBBarrierRememberedSet. Once this is merged, and after OpenJ9 is merged, you will have another small cleanup OMR change to remove staccatoRememberedSet.

amicic: same applies for other name changing including methods and classes - all should have temporary duplicates

gc/base/standard/ConcurrentGC.cpp
```diff
@@ -749,35 +726,6 @@ MM_ConcurrentGC::tearDown(MM_EnvironmentBase *env)
...
-	if(NULL != _cardTable) {
-		result = true;
-		/* Set card table address in GC Extensions */
-		_extensions->cardTable = _cardTable;
-	}
-
-	return result;
-}
-
-
```
amicic: this would be removed in the followup cleanup OMR change

gc/base/WorkPackets.hpp
```diff
@@ -110,6 +110,14 @@ class MM_WorkPackets: public MM_BaseVirtual
...
 	void reuseDeferredPackets(MM_EnvironmentBase *env);
 
+	/* methods used for Snapshot At the Beginning Barrier */
+	virtual MM_Packet *getBarrierPacket(MM_EnvironmentBase *env) { return NULL; }
+	virtual void removePacketFromInUseList(MM_EnvironmentBase *env, MM_Packet *packet) {}
+	virtual void putFullPacket(MM_EnvironmentBase *env, MM_Packet *packet) {}
+	virtual void putInUsePacket(MM_EnvironmentBase *env, MM_Packet *packet) {}
+	virtual void moveInUseToNonEmpty(MM_EnvironmentBase *env){}
+	virtual bool inUsePacketsAvailable(MM_EnvironmentBase *env){return false;}
+
```
amicic: Do these methods need to be virtual? in RembemberSetWorkPackets, if you make _workPackets be a pointer in to more specific WorkPacketsSATB class, then I believe they don't. These methods should not be used in outside of SATB specific code.

gc/base/standard/RememberedSetWorkPackets.hpp
```diff
@@ -0,0 +1,93 @@
...
+class EnvironmentModron;
+
+class MM_RememberedSetWorkPackets : public MM_BaseNonVirtual
+{
+/* Data members & types */
+public:
+	J9VMGCRememberedSet _rememberedSetStruct; /**< The VM-readable struct containing the remembered set "global" indexes. */
+protected:
+private:
+	MM_WorkPackets *_workPackets; /**< The workPackets struct used as backing store for the rememberedSet */
```
amicic: this should point to more specific MM_WorkPacketsSATB class

gc/base/standard/ConcurrentGC.cpp
```diff
@@ -2716,7 +2672,7 @@ MM_ConcurrentGC::doConcurrentTrace(MM_EnvironmentBase *env,
 
 	if (!isGcOccurred) {
 		/* If no more work left (and concurrent scanning is complete or disabled) then switch to exhausted now */
-		if (((MM_ConcurrentCardTable *)_cardTable)->isCardCleaningComplete() &&
+		if (NULL != _cardTable && ((MM_ConcurrentCardTable *)_cardTable)->isCardCleaningComplete() &&
```
dmitripivkine: Would you please add extra pair of brackets around (NULL != _cardTable) - here and below ?

gc/base/standard/ConcurrentGCIncrementalUpdate.hpp
```diff
@@ -0,0 +1,81 @@
...
+	void recordCardCleanPass2Start(MM_EnvironmentBase *env);
+
+public:
+	
+	/*
+	 * Function members
+	 */
+
+protected:
+	virtual bool initialize(MM_EnvironmentBase *env);
```
dmitripivkine: Should not be virtual as well

gc/base/standard/ConcurrentGCSATB.cpp
```diff
@@ -0,0 +1,298 @@
...
+	env->_workStack.reset(env, _markingScheme->getWorkPackets());
+	Assert_MM_true(env->_cycleState == NULL);
+	Assert_MM_true(CONCURRENT_OFF < _stats.getExecutionMode());
+	Assert_MM_true(_concurrentCycleState._referenceObjectOptions == MM_CycleState::references_default);
+	env->_cycleState = &_concurrentCycleState;
+
+	uintptr_t sizeTraced = 0;
+	while(NULL != (objectPtr = (omrobjectptr_t)env->_workStack.popNoWait(env))) {
+		/* Check for array scanPtr..if we find one ignore it*/
+		if ((uintptr_t)objectPtr & PACKET_ARRAY_SPLIT_TAG){
```
dmitripivkine: Is it correct? If we never split array should it be assert here when we see one? Or, if we do split how we are going to rescan reminder?

gc/base/MarkingScheme.cpp
charliegracie: I agree with @amicic that something should be done here.  My initial thought is that we leave the boolean as is and if concurrentMark is enabled then use the writebarriertype to decide which of the two WorkPacket structures to create.  I assume that pattern would get used in a few place.

gc/base/standard/ConcurrentGCSATB.cpp
dmitripivkine: Ok, this is correct but not optimum. Split array is handled in Work Packet by two sequential records:
- special recored with PACKET_ARRAY_SPLIT_TAG set and array element index we should start at followed by
- standard WP record for indexable object
The split index is ignored here so we start scan at index 0 always (and stop at array end or number of bytes to scan exceeds). The scanning of elements marked already should be relatively fast however does it mean that if number of bytes to scan is large we would have array_size_in_elements/split_amount attempts to rescan this array?

gc/base/WorkPackets.hpp
charliegracie: Agreed I think all of these could just be on the specific sublcass for SATB

gc/base/standard/ConcurrentGC.cpp
```diff
@@ -2756,7 +2712,9 @@ MM_ConcurrentGC::concurrentFinalCollection(MM_EnvironmentBase *env, MM_MemorySub
 {
 	/* Switch to FINAL_COLLECTION; if we fail another thread beat us to it so just return */
 	if	(_stats.switchExecutionMode(CONCURRENT_EXHAUSTED, CONCURRENT_FINAL_COLLECTION)) {
-
+		if(_extensions->snapshotAtTheBeginningBarrierEnabled) {
+		_extensions->sATBBarrierRememberedSet->preserveGlobalFragmentIndex(env);
```
charliegracie: tabbing here

gc/base/GCExtensionsBase.hpp
```diff
@@ -395,6 +395,8 @@ class MM_GCExtensionsBase : public MM_BaseVirtual {
 	bool gcThreadCountForced; /**< true if number of GC threads is specified in java options. Currently we have a few ways to do this:
 										-Xgcthreads		-Xthreads= (RT only)	-XthreadCount= */
 
+	bool snapshotAtTheBeginningBarrierEnabled;
```
charliegracie: Do we need this boolean?  Can we just use the write barrier type?

gc/base/standard/ConcurrentGC.cpp
ebadyano: fixed

gc/base/standard/ConcurrentGCIncrementalUpdate.hpp
ebadyano: fixed

gc/base/standard/ConcurrentGC.cpp
ebadyano: fixed

gc/base/standard/RememberedSetWorkPackets.hpp
ebadyano: fixed. I also re-named RememberedSetWorkPackets.hpp to RememberedSetWorkPacketsSATB.hpp so it doesn't conflict with the one in openj9 (which will get removed after openj9 changes are in)

gc/base/WorkPackets.hpp
ebadyano: Moved these methods to WorkPacketsSATB and made _workPackets  pointer to be WorkPacketsSATB.

gc/base/GCExtensionsBase.hpp
ebadyano: added staccatoRememberedSet back untill openj9 changes are in and we can remove it in the next set of changes.

gc/base/standard/ConcurrentGC.hpp
ebadyano: removed virtual.

gc/base/GCExtensionsBase.hpp
ebadyano: For now we want to use Xgc option to specify when to use sanpshot at the beginning. writeBarrierType is a field in Configuration and it seems that configuration is initialized after Xgc options are parsed. I added a check for snapshotAtTheBeginningBarrierEnabled in MM_ConfigurationDelegate to set the barrier type.. Is there a better way?

amicic: I think @ebadyano is right that we do need something in Extensions to at least temporary store the result  of parsing command line options, till the writeBarrierType is initialized in the Delegate.

I can only suggest minor modifications:
1) move the field to extensions->configurationOptions struct (where there are a few other high level  params controlled by command line options). Try make the name consistant with others in the struct,  for example: _forceOptionWriteBarrierSATB.
2) do not use that field anywhere else in the code, but fetch the barrier type from the Delegate instead.

gc/base/standard/RememberedSetWorkPackets.cpp
ebadyano: added REALTIME build flag.

gc/base/standard/WorkPacketsSATBBarrier.cpp
ebadyano: added REALTIME build flag.

gc/base/GCExtensionsBase.hpp
ebadyano: added _forceOptionWriteBarrierSATB under extensions->configurationOptions.

gc/base/MarkingScheme.cpp
ebadyano: I moved isSnapshotAtTheBeginningBarrierEnabled() to MM_Configuration and modified it to check for _writeBarrierType which is set when configuration is instantiated.

gc/base/Configuration.hpp
```diff
@@ -93,6 +93,23 @@ class MM_Configuration : public MM_BaseVirtual
...
+	{
+		return getBarrierType() == gc_modron_wrtbar_satb_and_oldcheck ||
+				getBarrierType() == gc_modron_wrtbar_satb;
+	}
+
+	MMINLINE bool
+	isIncrementalUpdateBarrierEnabled()
+	{
+		return getBarrierType() == gc_modron_wrtbar_cardmark_and_oldcheck ||
+				getBarrierType() == gc_modron_wrtbar_oldcheck ||
```
amicic: it would be only the 2 conditions with _cardmark_. Remove the pure gc_modron_wrtbar_oldcheck case.

amicic:
This is cleaner now. I expect more/better sharing and separation between ConcurrentGC class and its two subclasses, but that can be done in future changes.

@charliegracie @dmitripivkine please, re-review it.

gc/base/Configuration.hpp
```diff
@@ -93,6 +93,22 @@ class MM_Configuration : public MM_BaseVirtual
 	bool initializeHeapRegionDescriptor(MM_EnvironmentBase *env, MM_HeapRegionDescriptor *region) { return _delegate.initializeHeapRegionDescriptorExtension(env, region); }
 	void teardownHeapRegionDescriptor(MM_EnvironmentBase *env, MM_HeapRegionDescriptor *region) { _delegate.teardownHeapRegionDescriptorExtension(env, region); }
 
+	MMINLINE MM_GCWriteBarrierType getBarrierType() { return _writeBarrierType; }
+
+	MMINLINE bool
+	isSnapshotAtTheBeginningBarrierEnabled()
+	{
+		return getBarrierType() == gc_modron_wrtbar_satb_and_oldcheck ||
+				getBarrierType() == gc_modron_wrtbar_satb;
```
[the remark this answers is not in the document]
ebadyano: fixed.

gc/base/MarkingScheme.cpp
```diff
@@ -31,12 +31,17 @@
...
 #include "EnvironmentBase.hpp"
 #include "GCExtensionsBase.hpp"
 #include "Heap.hpp"
 #include "MarkMap.hpp"
 #include "MarkingScheme.hpp"
 #include "Task.hpp"
+#if defined(OMR_GC_REALTIME)
+#include "WorkPacketsSATB.hpp"
+#endif /* defined(OMR_GC_REALTIME) */
+#include "RememberedSetSATB.hpp"
```
dmitripivkine: Should this include be inside "if defined(OMR_GC_REALTIME)" ?

gc/base/GCExtensionsBase.hpp
```diff
@@ -70,6 +70,7 @@ class MM_ReferenceChainWalkerMarkMap;
 class MM_RememberedSetCardBucket;
 #if defined(OMR_GC_STACCATO)
 class MM_RememberedSetWorkPackets;
+class MM_RememberedSetSATB;
```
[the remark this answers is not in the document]
amicic: Right, I believe there is no difference between the two. But I think we should remove both and replace with something like OMR_GC_SATB_BARRIER. It would be a separate change.

ebadyano:
@charliegracie Charlie could you please review? Thank you!

charliegracie:
I have been looking through the changes and will have a complete review by Monday EOD

charliegracie:
One question is does this change force downstream changes?  Off hand it does not look like it but I just want to make sure.

amicic:
Downstream changes (mentioned at the top) are waiting for this but are not forced to go with this. Changes are made (like temporary duplicate fields etc) to break that dependency.

charliegracie:
@genie-omr build all

Issue #2309, opened 2018-02-12
Linking j9hashtable with gc_base


Varun-garg:
I want to use `j9hashtable` inside gc (#1759). I tired modifying `gc/base/makefile` with my limited makefile skills

```
MODULE_INCLUDES += ../include ../stats ../structs standard segregated j9hashtable j9pool $(OMRGLUE_INCLUDES)
MODULE_STATIC_LIBS += omrutil j9avl j9hashtable j9pool
```
Everything runs fine when I compile just omr, but when I compile openj9 (for which I need docker image). I get a segmentation error (as mentioned in #2189). I am able to analyze it by locating debug symbols and moving them to the `compressedrefs` directory. So the crash is: 
```(0x000014A84915CE7E [libj9gc29.so+0x118e7e]) pool_newElement + 697 in section .text 
(0x000014A84915950A [libj9gc29.so+0x11550a]) hashTableAddNodeInList + 246 in section 
(0x000014A849159361 [libj9gc29.so+0x115361]) hashTableAdd + 320 in section .text 
(0x000014A84919D852 [libj9gc29.so+0x159852]) valgrindMempoolAlloc(MM_GCExtensionsBase*, unsigned long, unsigned long) + 170 in section 
(0x000014A84921FAD9 [libj9gc29.so+0x1dbad9]) MM_Scavenger::incrementalScanCacheBySlot(MM_EnvironmentStandard*) + 15439 
(0x000014A849222986 [libj9gc29.so+0x1de986]) MM_Scavenger::completeScan(MM_EnvironmentStandard*) 
(0x000014A849222DA2 [libj9gc29.so+0x1deda2]) MM_Scavenger::workThreadGarbageCollect(MM_EnvironmentStandard*) 
(0x000014A84929C671 [libj9gc29.so+0x258671]) MM_ParallelScavengeTask::run(MM_EnvironmentBase*) 
(0x000014A84918F811 [libj9gc29.so+0x14b811]) MM_ParallelDispatcher::slaveEntryPoint(MM_EnvironmentBase*) 
(0x000014A84918F5F8 [libj9gc29.so+0x14b5f8]) dispatcher_thread_proc2(OMRPortLibrary*, void*)
(0x000014A84B2740F9 [libj9prt29.so+0x1f0f9])
(0x000014A84918F708 [libj9gc29.so+0x14b708]) dispatcher_thread_proc
(0x000014A84BDEA269 [libj9thr29.so+0x6269])
[...]

PR #2719, opened 2018-06-29
Enable DDR on AIX

mikezhang1234567890:
This pull request, in conjunction with the corresponding OpenJ9 pull request [#2309](https://github.com/eclipse/openj9/pull/2309), enables DDR on AIX, addressing eclipse/openj9#1769. This PR should be delivered before the
OpenJ9 PR.

The PR includes the following changes:

  * make _typeName field in MacroInfo non const since xlc on AIX fails to generate a copy assignment operator for const strings
  * replace string.front() with array access in Scanner since xlc does not support this method 
  * replace unsupported xargs command on AIX
  * initialize die sibling to NULL in AixSymbolTableParser.cpp
  * in AixSymbolTableParser.cpp, add missing DW_AT_data_member_location to fields parsed
  * fix missing digit in a number matching pattern in AixSymbolTableParser.cpp
  * fix AixSymbolTableParser incorrectly treating bool as 32bits instead of 8 bits
  * Create .dbg files for each of the shared libs on AIX. However, since objcopy does not seem to work properly on AIX, the debug libs are simply direct copies.
  * Generate file list on AIX even though AIX does not provide absolute paths. This allows the file blaclist to work on AIX.
  * Change the blacklist to only exclude top level types, in order to exclude the problematic type names that shows up on AIX. Compiling with this change on AIX and Linux both showed no apparent issues.

Testing was done on the OpenJ9 side.

### DIFF
--- a/ddr/include/ddr/scanner/dwarf/AixSymbolTableParser.hpp
+++ b/ddr/include/ddr/scanner/dwarf/AixSymbolTableParser.hpp
@@ -50,7 +50,7 @@ const string BUILT_IN_TYPES_THAT_ARE_TYPEDEFS[NUM_BUILT_IN_TYPES + 1] = {"", "",
 "", "", "integer*1", "integer*2", "integer*4", "wchar", "", "", "logical*8", "integer*8", "", "", ""
 };
 const int BUILT_IN_TYPE_SIZES[NUM_BUILT_IN_TYPES +1 ] = {0, 32, 8, 16, 32, 8, 8, 16, 32, 32,
-32, 0, 32, 64, 128, 32, 32, 32, 64, PTR_SIZE, 8, 8, 16, 32, 32, 64, 128,
+32, 0, 32, 64, 128, 32, 8, 32, 64, PTR_SIZE, 8, 8, 16, 32, 32, 64, 128,
 8, 16, 32, 16, 64, 64, 64, 64, 64, 64, 64};
 
 const string START_OF_FILE[3] = {"debug", "3", "FILE"};
--- a/ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
+++ b/ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
@@ -97,6 +97,15 @@ DwarfScanner::getSourcelist(Dwarf_Die die)
 	_fileNamesTable = NULL;
 	_fileNameCount = 0;
 
+	/* Get the dwarf source file names. */
+	if (DW_DLV_ERROR == dwarf_srcfiles(die, &_fileNamesTable, &_fileNameCount, &error)) {
+		ERRMSG("Failed to get list of source files: %s\n", dwarf_errmsg(error));
+		goto Done;
+	} else if (_fileNameCount < 0) {
+		ERRMSG("List of source files has negative length: %lld\n", _fileNameCount);
+		goto Done;
+	}
+
 	if (!hasAttr) {
 		/* The DIE didn't have a compilationDirectory attribute so we can skip
 		 * over getting the absolute paths from the relative paths.  AIX does
@@ -104,15 +113,6 @@ DwarfScanner::getSourcelist(Dwarf_Die die)
 		 */
 		goto Done;
 	} else {
-		/* Get the dwarf source file names. */
-		if (DW_DLV_ERROR == dwarf_srcfiles(die, &_fileNamesTable, &_fileNameCount, &error)) {
-			ERRMSG("Failed to get list of source files: %s\n", dwarf_errmsg(error));
-			goto Done;
-		} else if (_fileNameCount < 0) {
-			ERRMSG("List of source files has negative length: %lld\n", _fileNameCount);
-			goto Done;
-		}
-
 		/* Get the CU directory. */
 		if (DW_DLV_ERROR == dwarf_attr(die, DW_AT_comp_dir, &attr, &error)) {
 			ERRMSG("Getting compilation directory attribute: %s\n", dwarf_errmsg(error));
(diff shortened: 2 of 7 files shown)


charliegracie:
@genie-omr build all

omrmakefiles/rules.aix.mk
```diff
@@ -156,6 +156,14 @@ endif
...
-  GLOBAL_CXXFLAGS+=-g
-  GLOBAL_CFLAGS+=-g
+  GLOBAL_CXXFLAGS+=-g -qsuppress=1540-0825
+  GLOBAL_CFLAGS+=-g -qsuppress=1540-0825
+  GLOBAL_CPPFLAGS+=-qsuppress=1540-0825
+endif
+
+#DDR compile
+ifeq (yes,$(ENABLE_DDR))
+  GLOBAL_CXXFLAGS+=-qsuppress=1540-0825
```
charliegracie: Can you provide an example of a file which generates the warning and the reason for the warning?  It seems very heavy handed to add this option to the global flags when DDR or DEBUG is enabled.

mikezhang1234567890: ddrgen's getmacro script annotates header files with macros it finds with lines starting with `@`. During the preprocessor stage, xlC generates invalid character warnings for each line like that it finds, while GCC did not.

charliegracie: Does it annotate the files for normal compilation or just for compilation during ddrgen processing?  If it is the later then these options should only be added for one of the ddr phases and not everywhere.

ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
```diff
@@ -209,6 +209,12 @@ DwarfScanner::blackListedDie(Dwarf_Die die, bool *dieBlackListed)
 	bool blacklistedFile = true;
 	bool blacklistedType = true;
 
+/* Since AIX does not provide a DW_AT_comp_dir attribute, so we will not have
+ * a file list to look through, therefore don't blacklist based on file name
+ */
+#if defined(AIXPPC)
+	blacklistedFile = false;
```
keithc-ca: Isn't this change redundant? If there's no `DW_AT_decl_file` attribute, the outcome should be the same.

ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
```diff
@@ -256,6 +262,7 @@ DwarfScanner::blackListedDie(Dwarf_Die die, bool *dieBlackListed)
 			}
 		}
 	}
+#endif /* !defined(AIXPPC) */
```
keithc-ca: Nit: the convention is to use the same condition as is used in the `#if` directive (remove `!`).

omrmakefiles/rules.aix.mk
keithc-ca: I think some dependencies are missing in openj9: if they were added, the compiler wouldn't be invoked while the header files are mutated by getmacros.

ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
mikezhang1234567890: We set the error status to `DDR_RC_ERROR` and exit the function if the attribute isn't found. 

Looking at the function again, it would probably be better to just remove the throw error part:

https://github.com/mikezhang1234567890/omr/blob/21386ad081a48644c241725a7689dcf706a00aa5/ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp#L235-L236

keithc-ca: The call to `dwarf_hasattr` should not fail, but it might say the attribute is not present.

omrmakefiles/rules.aix.mk
charliegracie: It still feels very heavy handed to set these flags globally if they are not needed.  Also as OpenJ9 moves towards CMake do we need changes in the CMake build system as well?

ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
mikezhang1234567890: I was wrong there. The `DW_AT_comp_dir` is the attribute that's missing, so it fails the check on line 255: `if (declFile <= (Dwarf_Unsigned)_fileNameCount)`

omrmakefiles/rules.aix.mk
keithc-ca: I spoke with @mikezhang1234567890 about reorganizing the makefiles in OpenJ9 so we won't have such warnings.

charliegracie:
@mikezhang1234567890 ping?

mikezhang1234567890:
Working on moving the ddrgen invocation out of the OpenJ9 building VM step at the moment, so hopefully the suppress warning flag will no longer be needed.

charliegracie:
If this PR will still be required can you mark it as WIP in the title until you are ready?  Thanks

ddr/include/ddr/scanner/dwarf/AixSymbolTableParser.hpp
```diff
@@ -50,7 +50,7 @@ const string BUILT_IN_TYPES_THAT_ARE_TYPEDEFS[NUM_BUILT_IN_TYPES + 1] = {"", "",
 "", "", "integer*1", "integer*2", "integer*4", "wchar", "", "", "logical*8", "integer*8", "", "", ""
 };
 const int BUILT_IN_TYPE_SIZES[NUM_BUILT_IN_TYPES +1 ] = {0, 32, 8, 16, 32, 8, 8, 16, 32, 32,
-32, 0, 32, 64, 128, 32, 32, 32, 64, PTR_SIZE, 8, 8, 16, 32, 32, 64, 128,
+32, 0, 32, 64, 128, 32, 8, 32, 64, PTR_SIZE, 8, 8, 16, 32, 32, 64, 128,
```
keithc-ca: I think `BUILT_IN_TYPES` and `BUILT_IN_TYPE_SIZES ` should be replaced by an array of structs that pair the names and sizes together to make this sort of error easier to spot, but that can wait for a separate PR.

keithc-ca:
@youngar Andrew, can you have a look at this and merge as you see fit? Thanks.

youngar:
@genie-omr build all

youngar:
It looks like we're not compiling DDR in the OMR builds, is this correct?

PR #2721, opened 2018-07-02
Add .gitattributes for zos platform

groeges:
This .gitattributes.zos file will not be picked up by the latest versions of git, so git will not attempt to encode/decode files and current clone/checkout will not be affected.

On a zos platform the openjdk extensions file ./closed/get_j9_source.sh will be modified to copy this .gitattributes.zos file to .git/info/attributes and the files refreshed. This will then encode/decode the files.

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

### DIFF
--- a/.gitattributes.zos
+++ b/.gitattributes.zos
@@ -1,0 +1,22 @@
+# This .gitattributes file will cause all text files EXCEPT for
+# git's .gitattributes and .gitignore files to be encoded as EBCDIC
+# when cloned using git (from Rocket Software) on z/OS platform.
+#
+# Selected binary files will not be translated at all.
+#
+# The default for text files
+* git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+
+# git's files (which MUST be ASCII)
+.gitattributes git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+.gitignore git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+
+# omr files that need to be encoded as EBCDIC
+*.hdf git-encoding=iso8859-1 working-tree-encoding=ibm-1047
+*.tdf git-encoding=iso8859-1 working-tree-encoding=ibm-1047
+
+# Binary files
+*.jpg git-encoding=BINARY working-tree-encoding=BINARY
+*.png git-encoding=BINARY working-tree-encoding=BINARY
+*.gif git-encoding=BINARY working-tree-encoding=BINARY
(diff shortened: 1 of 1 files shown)


.gitattributes.zos
```diff
@@ -0,0 +1,22 @@
+# This .gitattributes file will cause all text files EXCEPT for
+# git's .gitattributes and .gitignore files to be encoded as EBCDIC
+# when cloned using git (from Rocket Software) on z/OS platform.
```
charliegracie: Is this comment correct?  To me it looks like all files will be forced to ASCII except *.hdf and *.tdf files (and of course the binary files)

charliegracie:
@dnakamura thoughts?

dnakamura:
Looks good, but a couple of minor nits
- I don't think we actually need the explicit default tag of `iso8859-1` from the Rocket git docs it looks like that is the default anyway.
- For binary files we should probably do it the standard git way. eg `*.png binary`
- The git encoding should maybe be `UTF-8` ( I think there are a couple files which have non-ascii characters), although this could just end up being more hassle than its worth

.gitattributes.zos
groeges: I will update the comment.

groeges:
@dnakamura I would prefer to specifically specify the default encoding. 
The specification for the binary files was copied from the existing .gitattributes.
I think the encoding should be ISO8859-1 instead of UTF-8 as that is what is used on z.OS.

.gitattributes.zos
```diff
@@ -0,0 +1,23 @@
...
+# git (from Rocket Software) on z/OS platform.
+#
+# Selected binary files will not be translated at all.
+#
+# The default for text files
+* git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+
+# git's files (which MUST be ASCII)
+.gitattributes git-encoding=iso8859-1 working-tree-encoding=iso8859-1
+.gitignore git-encoding=iso8859-1 working-tree-encoding=iso8859-1
```
keithc-ca: Aren't these redundant, given they specify the same encodings as the default above? I suggest they be removed.

groeges: @keithc-ca Have removed the encoding definitions that are covered by the default definition

groeges:
@genie-omr build all

groeges:
In order to not impact the OMR repository with changes for z/OS we have implemented the required change within the z/OS OpenJDK Extensions repository. This PR is therefore no longer required so closing it.

PR #2722, opened 2018-07-03
Add Evaluator for max and min opcodes in Z codegen

dhong44:
Add evaluator to handle max and min opcodes in Z codegen
The specific opcodes are imax, imin, lmax and lmin.

### DIFF
--- a/compiler/z/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/z/codegen/ControlFlowEvaluator.cpp
@@ -898,7 +898,7 @@ static TR::Register * maxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool i
    TR::Register *registerA = cg->gprClobberEvaluate(node->getFirstChild());
    TR::Register *registerB = cg->evaluate(node->getSecondChild());
    // Mask is 4 to pick b when a is Lower for max, 2 to pick b when a is higher for min
-   uint8_t mask = isMax ? 0x4 : 0x2;
+   const uint8_t mask = isMax ? 0x4 : 0x2;
 
    if (node->getOpCodeValue() == TR::imax || node->getOpCodeValue() == TR::imin)
       {
--- a/compiler/z/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/z/codegen/ControlFlowEvaluator.cpp
@@ -893,6 +893,28 @@ lcmpHelper(TR::Node * node, TR::CodeGenerator * cg)
    return targetRegister;
    }
 
+/**
+ *  \brief
+ *     Compares 2 numbers are returns the greater of the 2.
+ *     ONLY SUPPORTS imax, imin, lmax, lmin
+ *
+ *  \detail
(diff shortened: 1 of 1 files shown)


dhong44:
@fjeremic Could I please get a review

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -893,19 +893,63 @@ lcmpHelper(TR::Node * node, TR::CodeGenerator * cg)
    return targetRegister;
    }
 
+static TR::Register * maxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMax)
+   {
+   TR::Register *registerA = cg->gprClobberEvaluate(node->getFirstChild());
+   TR::Register *registerB = cg->evaluate(node->getSecondChild());
+   // Mask is 4 to pick b when a is Lower for max, 2 to pick b when a is higher for min
+   uint8_t mask = isMax ? 0x4 : 0x2;
```
fjeremic: Can this be `const`?

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -893,19 +893,63 @@ lcmpHelper(TR::Node * node, TR::CodeGenerator * cg)
...
+
+   if (node->getOpCodeValue() == TR::imax || node->getOpCodeValue() == TR::imin)
+      {
+      generateRRInstruction(cg, TR::InstOpCode::CR, node, registerA, registerB);
+      generateRRFInstruction(cg, TR::InstOpCode::LOCR, node, registerA, registerB, mask, true);
+      }
+   else if (node->getOpCodeValue() == TR::lmax || node->getOpCodeValue() == TR::lmin)
```
fjeremic: I'd like to see an `else` with a `TR_ASSERT_FATAL` in case someone tries to use this helper for `bmax` for example (an opcode that is not supported by this function).

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -893,19 +893,63 @@ lcmpHelper(TR::Node * node, TR::CodeGenerator * cg)
    return targetRegister;
    }
 
+static TR::Register * maxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool isMax)
```
fjeremic: Could we please Doxygen document this function and it's limitations (only handles 32-bit and 64-bit min/max).

compiler/z/codegen/ControlFlowEvaluator.cpp
dhong44: 6080fa1

compiler/z/codegen/ControlFlowEvaluator.cpp
dhong44: d420fe4

fjeremic:
This is a Z only change, no point waiting for Travis CI and AppVeyor builds to pass.

PR #2724, opened 2018-07-04
Move uses of snprintf into cpp

charliegracie:
Having the uses of snprintf in the header forces the define for
snprintf to be in the header for Windows platforms. Moving limits
the scope of the definition into the cpp.


mstoodle:
sorry, @charliegracie I missed the review request in vacation churn :( .

mstoodle:
@genie-omr build all

PR #2727, opened 2018-07-05
Make hookgen/tracegen deal with EBCDIC on z/OS

dnakamura:
Add a general flag `USE_NATIVE_ENCODING` to the makefiles, which only has
a defined meaning on z/OS at the moment. If it is set to 1, the target is
built as a "native" executable. Ie, dont use ascii literals, and dont use
the a2e library. In addition, `OMR_EBCDIC` is defined when this flag is
active on z/OS


youngar:
@genie-omr build all

PR #2728, opened 2018-07-06
z/TPF Specific Build Updates for refactoring build options

jjohnst-us:
Today zTPF build does not integrate very well for Java builds.
These updates refactor build options in an effort to cleanup
z/TPF Builds so that the build intergrates nicely with the rest
of the z/TPF build environment.


omrmakefiles/rules.mk
```diff
@@ -197,7 +197,7 @@ $(AR) $(ARFLAGS) $(MODULE_ARFLAGS) $(GLOBAL_ARFLAGS) rcv $@ $(OBJECTS)
 endef
 
 define CLEAN_COMMAND
--$(RM) $(OBJECTS) $(OBJECTS:$(OBJEXT)=.i) *.d
+-$(RM) $(OBJECTS) $(OBJECTS:$(OBJEXT)=.i) $(OBJECTS:$(OBJEXT)=.lst) *.d
```
charliegracie: Does this change need to only happen for ZTPF? Will the RM fail on platforms is there is no .lst files?

charliegracie:
@genie-omr build zos zlinux plinux win

omrmakefiles/rules.mk
jjohnst-us: Nice catch Charlies, yes its possible other platforms wouldn't have *.lst.  Testing another update which I will push once I can verify it works.  Would look something like below:
``
define CLEAN_COMMAND
-$(RM) $(OBJECTS) $(OBJECTS:$(OBJEXT)=.i) *.d
ifeq (linux_ztpf,$(OMR_HOST_OS))
-$(RM) $(OBJECTS) $(OBJECTS:$(OBJEXT)=.lst) *.d
endif
endef

jjohnst-us:
@charliegracie 
Made an update to address your review comment.  My personal build had one unrelated failure because I needed to update the OMR_HOST_ARCH from s390x to s390.

Let me know if you need anything else.

charliegracie:
@genie-omr build zos zlinux

PR #2729, opened 2018-07-06
Add flag to prevent AIX vmem from using shmat

gacholio:
Add a flag to the omrvmem_reserve_memory mode to indicate that
omrvmem_numa_set_affinity will not be used on the resulting allocation.
On AIX, this ensures that mmap is used to allocate the memory rather
than shmat.  All platforms are updated such that
omrvmem_numa_set_affinity fails if the new flag is set.


charliegracie:
@genie-omr build all

charliegracie:
@LinHu2016 can you review this?

LinHu2016:
LGTM

PR #2731, opened 2018-07-09
Add options for jProfiling

r30shah:
This commits adds following options.

1. jProfilingLoopRecompThreshold : To specify base threshold for loop which will be adjusted as per loop depth and will be compared against raw loop count.
2. jProfilingMethodRecompThrehold : To specify number of method invocations required to trip recompilation of profiling body.

### DIFF
--- a/compiler/control/OMROptions.hpp
+++ b/compiler/control/OMROptions.hpp
@@ -1944,6 +1944,12 @@ public:
 
    int32_t getJitMethodEntryAlignmentBoundary(TR::CodeGenerator *cg);
    void setJitMethodEntryAlignmentBoundary(int32_t boundary) { _jitMethodEntryAlignmentBoundary = boundary; }
+/**   \brief Returns a threshold on the profiling method invocations to trip recompilation
+ */
+   int32_t getJProfilingMethodRecompThreshold() { return _jProfilingMethodRecompThreshold; }
+/**   \brief Returns a base threshold for loop to trip recompilation
+ */
+   int32_t getJProfilingLoopRecompThreshold() { return _jProfilingLoopRecompThreshold; }
 
    inline static float getMinProfiledCheckcastFrequency() { return _minProfiledCheckcastFrequency/((float)100.0); }
    static bool isQuickstartDetected() { return _quickstartDetected; }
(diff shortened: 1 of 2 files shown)


r30shah:
@andrewcraik @fjeremic Can you please review this. I added options related to jProfiling in OMR layer such that I can use them per method as well.

andrewcraik:
do you have a depends on OMR PR number then based on your last comment?

r30shah:
Changes in this PR does not depend on any changes in OMR. OpenJ9 counter part is https://github.com/eclipse/openj9/pull/2348

andrewcraik:
@r30shah can you be clear - can this change merge without the openj9 change merging or will openj9 break if the PR you mentioned is not merged at the same time? If the latter is true can we stage the change to avoid this or will that add a lot of overhead to delivering what should be a simple change?

andrewcraik:
change look reasonable to me - I look forward to when we can move the J9 project specific stuff down to OpenJ9 but until the J9 options processor can handle method selectors then this will have to do. Over to you @vijaysun-omr

r30shah:
@andrewcraik Yes you are right, merging this won't break **OMR** but it will definitely break **OpenJ9** if https://github.com/eclipse/openj9/pull/2348 changes are not merged together. I should have been more clear, sorry about that.

compiler/control/OMROptions.hpp
```diff
@@ -1945,6 +1945,9 @@ class OMR_EXTENSIBLE Options
    int32_t getJitMethodEntryAlignmentBoundary(TR::CodeGenerator *cg);
    void setJitMethodEntryAlignmentBoundary(int32_t boundary) { _jitMethodEntryAlignmentBoundary = boundary; }
 
+   int32_t getJProfilingMethodRecompThreshold() { return _jProfilingMethodRecompThreshold; }
+   int32_t getJProfilingLoopRecompThreshold() { return _jProfilingLoopRecompThreshold; }
```
fjeremic: Could we Doxygen document these new APIs?

vijaysun-omr:
@genie-omr build all

r30shah:
I removed changes to add `disableJProfilingRecompLoopTest`  which would break OpenJ9 without https://github.com/eclipse/openj9/pull/2348. Going to open up separate PR for tracing and Disabling the loop recompilation test generation for jProfiling body

andrewcraik:
I think separating the changes so that this change is not co-dependent on an openJ9 change to not break OpenJ9 is a good idea. @vijaysun-omr could you rebuild now that the review feedback has been incorporated?

vijaysun-omr:
@genie-omr build all

PR #2734, opened 2018-07-10
Optimized Simplifier related to abs/neg

0dvictor:
Simplified following patterns:
```
1.
node foo                         node foo
   [ilfd]abs                  =>    non-negative node bar
      non-negative node bar
node foo                         node foo
   [ilfd]abs                  =>    [ilfd]neg
      negative node bar                negative node bar
 * Note that transformation of [fd]abs is default off *
2.
node foo                         node foo
   [ilfd]abs                  =>    [ilfd]abs
      [ilfd]abs                        node bar
         node bar
3.
node foo                         node foo
   [ilfd]abs                  =>    [ilfd]abs
      [ilfd]neg                        node bar
         node bar
4.
node foo                         node foo
   dneg                       =>    node bar
      dneg
         node bar
```

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -9805,12 +9805,12 @@ TR::Node *fnegSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
    // Make this  work for -(0-0)
    //
    if (firstChild->getOpCodeValue() == TR::fneg)
-     {
+      {
       if (performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] --A -> A\n", s->optDetailString(), node))
-        {
-        node= s->replaceNode(node,firstChild->getFirstChild(), s->_curTree); // remove neg
-        }
-     }
+         {
+         node= s->replaceNode(node,firstChild->getFirstChild(), s->_curTree); // remove neg
+         }
+      }
    // don't just do -A op B -> -(A op B) as not always a win, so catch opportunity here
    else if ((firstChild->getOpCodeValue() == TR::fmul ||
            firstChild->getOpCodeValue() == TR::fdiv ||
@@ -9835,14 +9835,14 @@ TR::Node *fnegSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
          childNo=1;
          }
       if (child2Change && child2Change->getReferenceCount() == 1 && performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] -(-A op B) -> A op B (op=*,/,%%)\n", s->optDetailString(), node))
-        {
-        // remove neg
-        TR::Node * newChild= s->replaceNode(child2Change,child2Change->getFirstChild(), s->_curTree);
-        firstChild->setChild(childNo,newChild);
-        // now replace current node with fmul,fdiv or frem
-        node= s->replaceNode(node,node->getFirstChild(), s->_curTree);
-        }
-   }
+         {
+         // remove neg
+         TR::Node * newChild= s->replaceNode(child2Change,child2Change->getFirstChild(), s->_curTree);
+         firstChild->setChild(childNo,newChild);
+         // now replace current node with fmul,fdiv or frem
+         node= s->replaceNode(node,node->getFirstChild(), s->_curTree);
+         }
+      }
    else if (s->comp()->cg()->supportsNegativeFusedMultiplyAdd())
       {
       // unless we've already got an inserted multiply, transformit
@@ -9889,11 +9889,17 @@ TR::Node *dnegSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
       }
 
    // Make this work for -(0-0)
-
-   if (s->comp()->cg()->supportsNegativeFusedMultiplyAdd())
+   if (firstChild->getOpCodeValue() == TR::dneg)
+      {
+      if (performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] --A -> A\n", s->optDetailString(), node))
+         {
+         node= s->replaceNode(node,firstChild->getFirstChild(), s->_curTree); // remove neg
+         }
+      }
+   else if (s->comp()->cg()->supportsNegativeFusedMultiplyAdd())
       {
       if ((firstChild->getOpCode().isAdd() || firstChild->getOpCode().isSub()) &&
-     !(firstChild->getFirstChild()->isFPStrictCompliant() || firstChild->getSecondChild()->isFPStrictCompliant()) &&
+          !(firstChild->getFirstChild()->isFPStrictCompliant() || firstChild->getSecondChild()->isFPStrictCompliant()) &&
           performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] -(-A +/- B) -> -((A*1)+/-B)\n", s->optDetailString(), node))
          {
          TR::Node * newConst = TR::Node::create(firstChild->getFirstChild(), TR::dconst, 0);
@@ -9907,7 +9913,7 @@ TR::Node *dnegSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
          newMul->setIsFPStrictCompliant(true);
          }
       else if (firstChild->getOpCode().isMul() &&
-          performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] -(A*B) -> -((A*B)-0)\n", s->optDetailString(), node))
+               performTransformation(s->comp(), "%sTransforming [" POINTER_PRINTF_FORMAT "] -(A*B) -> -((A*B)-0)\n", s->optDetailString(), node))
          {
          TR::Node * newConst = TR::Node::create(firstChild, TR::dconst, 0);
          TR::Node * newAdd = TR::Node::create(firstChild, TR::dsub, 2);
@@ -9962,28 +9968,56 @@ TR::Node *lconstSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * 
    return node;
    }
 
-TR::Node *iabsSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
+TR::Node *ilfdabsSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
    {
    simplifyChildren(node, block, s);
+
+   static bool AllowFlagBasedTransformationForFloatOrDouble = feGetEnv("TR_AllowFlagBasedTransformationForFloatOrDouble") != NULL;
+
+   TR::ILOpCodes abs = node->getOpCodeValue();
+   TR::ILOpCodes neg = TR::BadILOp;
+   bool performFlagBasedTransformation = false;
+   switch (abs)
+      {
+      case TR::iabs:
+         neg = TR::ineg;
+         performFlagBasedTransformation = true;
+         break;
+      case TR::labs:
+         neg = TR::lneg;
+         performFlagBasedTransformation = true;
+         break;
+      case TR::fabs:
+         neg = TR::fneg;
+         performFlagBasedTransformation = AllowFlagBasedTransformationForFloatOrDouble;
+         break;
+      case TR::dabs:
+         neg = TR::dneg;
+         performFlagBasedTransformation = AllowFlagBasedTransformationForFloatOrDouble;
+         break;
+      default:
+         TR_ASSERT(0, "unexpected opcode %d\n", node->getOpCodeValue());
+      }
    auto child = node->getChild(0);
    if (child->isNonNegative() &&
-       performTransformation(s->comp(), "%sSimplify iabs of non-negative child at [" POINTER_PRINTF_FORMAT "]\n",
-             s->optDetailString(), node))
+       performFlagBasedTransformation &&
+       performTransformation(s->comp(), "%sSimplify abs of non-negative child at [" POINTER_PRINTF_FORMAT "]\n", s->optDetailString(), node))
       {
       return s->replaceNodeWithChild(node, child, s->_curTree, block);
       }
-   return node;
-   }
-
-TR::Node *labsSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
-   {
-   simplifyChildren(node, block, s);
-   auto child = node->getChild(0);
-   if (child->isNonNegative() &&
-       performTransformation(s->comp(), "%sSimplify labs of non-negative child at [" POINTER_PRINTF_FORMAT "]\n",
-             s->optDetailString(), node))
+   else if (child->isNegative() &&
+            performFlagBasedTransformation &&
+            performTransformation(s->comp(), "%sSimplify abs of non-positive child at [" POINTER_PRINTF_FORMAT "]\n", s->optDetailString(), node))
       {
-      return s->replaceNodeWithChild(node, child, s->_curTree, block);
+      TR::Node::recreate(node, neg);
+      return s->simplify(node, block);
+      }
+   else if ((child->getOpCodeValue() == abs || child->getOpCodeValue() == neg) &&
+            performTransformation(s->comp(), "%sSimplify abs of abs/neg child at [" POINTER_PRINTF_FORMAT "]\n", s->optDetailString(), node))
+      {
+      node->setAndIncChild(0, child->getFirstChild());
+      child->recursivelyDecReferenceCount();
+      return s->simplify(node, block);
       }
    return node;
    }
--- a/fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
+++ b/fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
@@ -1,0 +1,217 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include <cmath>
+#include "JitTest.hpp"
+#include "default_compiler.hpp"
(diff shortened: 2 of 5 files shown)


andrewcraik:
Given the great care and discussion in #2711 I am not sure that the neg of MIN_INT is equal to the abs of MIN_INT.

Pattern 1 could be invalid if the negative node bar is i or l and has the MIN value no? neg of MIN is not required to be the same as abs of MIN I don't think since it isn't representable in twos compliment in the integer range (eg 2^(n-1) -> 2^(n-1)-1).

Pattern 2 should be ok abs of abs should be redundant I think.

Pattern 3 could also be invalid if the neg were of the MIN value since abs of MIN and neg of MIN do not have to do the same thing...

Now for f and d I'll need to have a think - mathematically one would say yes, but I'll need to consider all the special values like the signed zeros etc to make sure the floating point versions of these are true.

@jdmpapin FYI given your involvement in #2711

0dvictor:
Existing OMR logic has already assumed abs(MIN_INT) equals to -MIN_INT because they both equal to MIN_INT.
For example, after #2711, OMR's VP constrainIabs assumes abs(MIN_INT) equals to MIN_INT; VP constrainIneg also assumes -MIN_INT equals MIN_INT

https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/VPHandlers.cpp#L7131-L7206
https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/VPHandlers.cpp#L7007-L7063

VP constrains on Long are similar.

andrewcraik:
so that OMR logic says abs(MIN_INT) == MIN_INT not -MIN_INT. The neg opcode (at least on x86) assume just straight twos compliment negation and will overflow. As a result iabs -> ineg, for example, means that iabs(MIN_INT) == ineg(MIN_INT) which I don't believe to be true (at least with the current evaluators). Now there may be some consistent state we can arrive at that makes the above valid, but I think there needs to be a bigger consideration of the MIN_INT corner case before we can unleash these opt changes as we are likely to get bitten by a corner case on some platform somewhere.

jdmpapin:
> abs(MIN_INT) == MIN_INT not -MIN_INT

As Victor already mentioned,

    abs(MIN_INT) = -MIN_INT = MIN_INT

which I write in that order because it shows the justification for `abs(MIN_INT) = MIN_INT`. The equality on the left holds because `MIN_INT < 0`, and we want to define `abs` so that `abs(x) = -x` whenever `x < 0`. The equality on the right holds because we use two's complement.

I still need to look at the particular changes here, but whatever they are, `-MIN_INT != MIN_INT` is not an objection to them (because it isn't true)

0dvictor:
OMR's OpCodeTest also requires `abs(MIN) == -MIN`:
https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/fvtest/compilertest/tests/OpCodesTest.hpp#L915-L916

`neg(MIN_INT)` is evaluated to `-MIN_INT`;
`abs(MIN_INT)` is evaluated to `MIN_INT >= 0 ? MIN_INT : -MIN_INT`, which is `-MIN_INT`.
Therefore, whatever `-MIN_INT` equals to, `abs(MIN_INT)` always equals to `-MIN_INT`.

jdmpapin:
While I think that's the intention in the OpCodeTest snippet, I should mention that if `T` is a signed type at least as large as `int`, and if the platform uses two's complement (like all of our platforms do), then the results of `neg<T>()` and `abs<T>()` on the minimum value of `T` cannot be relied upon in C++ because negating the minimum value is a signed overflow, which is undefined behaviour unless the opcode tests are compiled with `-fwrapv` or equivalent. And `T` is likely to be signed at least for `abs<T>()`, since otherwise `a >= 0` will always hold.

andrewcraik:
Right and my concern is not that mathematically these things are wrong but that we may have a problem with undefined behavior (either in software or hardware) related to this which may bite us hence my caution in saying that the folding is safe in the MIN_INT case. If you have been through the evaluators and the MIN_INT cases are ok then I'm not opposed, I'm just very worried about a subtle bug slipping through.

0dvictor:
For OpCodeTest, assuming we define `MIN_INT` as the value of the minimum possible value of a 32-bit signed integer, I agree `T` must be a signed type to make `abs(MIN_INT) == neg(MIN_INT)`. However, the representation or behavior of negative numbers is not a requirement. As long as `MIN_INT >= (T)0` has a valid behavior and returns false, `abs(MIN_INT)` is reduced to `-MIN_INT` and hence has the same behavior of `-MIN_INT` regardless what such behavior is or whether such behavior is defined.

andrewcraik:
@jdmpapin without regard to my comments what is your assessment of this change since you have spent a lot of time thinking about the boundary conditions on these various types and the operations associated there with.

jdmpapin:
Patterns 1-3 are all valid for two's complement integers, and as far as I'm aware the compiler assumes two's complement throughout. UB is not an issue here, since it only applies to C and C++ programs, and overflow is perfectly well-defined in IL as wrapping mod a power of two.

I believe pattern 4 is good, and in any case it's at least as good as the existing transformation done in the handler for `fneg`.

There is an additional transformation that newly applies to floating-point not mentioned in the commit message:

    [df]abs           =>       bar (X>=0)
      bar (X>=0)

This transformation and pattern 1 rely on `isNonNegative()` and `isNegative()`, but the precise meaning of these queries for floating-point values isn't obvious to me, and I'm not sure the compiler has an existing convention. For example, they could be based on the usual IEEE754 ordering comparison, or they could be based on the sign bit.

For patterns 2 and 3 in floating-point, these work as long as `[fd]neg` flips the sign bit, and `[fd]abs` clears it. I don't have IEEE754 handy, but I highly suspect these behaviours are allowed or even required. The following all seem to agree on negation:

- the `fneg` and `dneg` bytecode instructions in the JVMS,
- the code generated by GCC for unary minus on my system even with `-O0` (`xorps` to flip the bit), and
- the unary minus operator in Python 2.7.5 (based on printing results for zeroes, infinities, and NaN).

The following all seem to agree on the absolute value:

- the Javadoc for `Math.abs(F)F` and `Math.abs(D)D`,
- the code generated by GCC for `fabs` on my system even with `-O0` (`andps` to clear the bit), and
- the `abs` function in Python 2.7.5 (based on printing results on zeroes, infinities, and NaN).

While I'm not sure that floating-point semantics are nailed down for OMR IL, and it's possible that somebody could have a language in which, for example, -NaN has to produce _the same_ NaN, I think the semantics described above should be pretty uncontroversial.

andrewcraik:
Thanks @jdmpapin for the review. Given the above discussion I think I am happy with Pattern 2 and Pattern 3. I think that because the meaning of the flags for Pattern 1 is unclear we should not be this aggressive about optimization and should leave this alone for the time being - discussions about the floating point model are ongoing so this can be revisited later. Same goes for [df]abs of bar => bar.

With regards to the flipping/clearing of the sign bit - I did check the standard and the operations are defined in terms of the sign bit so these should be good. Pattern 4 should be good as a result.

So I would be happy saying yes to 2 3 and 4. Exclude 1 and the [df]abs of bar => bar for now - ie do not rely on flags whose meaning is poorly defined until we define their meaning as part of defining the OMR floating point model.

jdmpapin:
@ajcraik Just to be clear, you're still okay pattern 1 for integers, yes?

0dvictor:
The way OMR set `isNonNegative()` and `isNegative()` is to re-interpret the floating number bits as if it was an integer of its corresponding size, and set it the two flags accordingly. Therefore, `isNonNegative()` means the sign bit is cleared and `isNegative()` means the sign bit is set.

https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/ValuePropagationTable.hpp#L167-L170
https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/VPHandlers.cpp#L878-L883
https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/VPHandlers.cpp#L845-L864
https://github.com/eclipse/omr/blob/9a62d4f832d828ea6b9fc1485e7f205eca13c735/compiler/optimizer/VPHandlers.cpp#L885-L904

0dvictor:
As @jdmpapin pointed out, I updated the commit message to include the missing transformation of non-negative child of abs.

andrewcraik:
pattern 1 for integers is fine. Flags have proven a bit of a sore point for the compiler because node recreation etc does not always handle them properly (API improvements have been made to improve this situation, but I don't think the treatment is universal. I would prefer that we get what we can - the majority and we can revisit this last bit as part of the floating point VP work that is going at the moment.

0dvictor:
I see your concern about the counter-intuitive meanings and possible misbehaved flags on float/double nodes. I updated the proposed change so that flag-based transformations on float/double node is turned off by default.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -9962,29 +9968,55 @@ TR::Node *lconstSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier *
...
-TR::Node *iabsSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
+TR::Node *ilfdabsSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
    {
    simplifyChildren(node, block, s);
-   auto child = node->getChild(0);
-   if (child->isNonNegative() &&
-       performTransformation(s->comp(), "%sSimplify iabs of non-negative child at [" POINTER_PRINTF_FORMAT "]\n",
-             s->optDetailString(), node))
+
+   static auto AllowFlagBasedTransformationForFloatOrDouble = (bool)feGetEnv("TR_AllowFlagBasedTransformationForFloatOrDouble");
```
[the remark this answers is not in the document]
0dvictor: Sure, updated.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -9962,29 +9968,55 @@ TR::Node *lconstSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier *
...
+
+   auto abs = node->getOpCodeValue();
+   auto neg = TR::BadILOp;
```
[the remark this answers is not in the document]
0dvictor: Updated.

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -9962,29 +9968,55 @@ TR::Node *lconstSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier *
...
       return s->replaceNodeWithChild(node, child, s->_curTree, block);
       }
+   else if (child->isNegative() &&
+            performFlagBasedTransformation &&
+            performTransformation(s->comp(), "%sSimplify abs of non-positive child at [" POINTER_PRINTF_FORMAT "]\n", s->optDetailString(), node))
+      {
+      TR::Node::recreate(node, neg);
+      return s->simplify(node, block);
+      }
+   else if ((abs == child->getOpCodeValue() || neg == child->getOpCodeValue()) &&
```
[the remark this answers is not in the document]
0dvictor: Updated.

charliegracie:
@genie-omr build all

vijaysun-omr:
@andrewcraik are you satisfied with the changes made based on your review ?

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -9962,29 +9968,55 @@ TR::Node *lconstSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier *
...
-   auto child = node->getChild(0);
-   if (child->isNonNegative() &&
-       performTransformation(s->comp(), "%sSimplify iabs of non-negative child at [" POINTER_PRINTF_FORMAT "]\n",
-             s->optDetailString(), node))
+
+   static bool AllowFlagBasedTransformationForFloatOrDouble = feGetEnv("TR_AllowFlagBasedTransformationForFloatOrDouble") != NULL;
+
+   TR::ILOpCodes abs = node->getOpCodeValue();
+   TR::ILOpCodes neg = TR::BadILOp;
+   bool performFlagBasedTransformation = true;
```
andrewcraik: I think this bool is misnamed - it would be better as isFlagBasedTransformation defaulting to false and set to true. You would then later test AllowFlagBasedTransformationForFloatOrDobule and this condition. This is functionally correct, just harder to read.

andrewcraik:
modulo the comment on the variable name I think I am now happy with this given all the revisions.

compiler/optimizer/OMRSimplifierHandlers.cpp
0dvictor: Good point, updated.

0dvictor:
This PR is updated and re-verified.

vijaysun-omr:
@0dvictor is it easy to create some test cases for these cases you are now catching ?

vijaysun-omr:
@genie-omr build all

0dvictor:
Sorry that I missed the first message regarding to test cases. I'm not sure if OMR has a test framework to verify a specific IL transformation; however, the code involved in this PR is covered by [ilfd]NegConst tests.  Are they sufficient for the purpose of verify the correctness of these transformations?

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                                  "    (%sabs (%sload parm=0))))))",
+                  nameForType<TypeParam>(), nameForType<TypeParam>(),
+                  prefixForType<TypeParam>(), prefixForType<TypeParam>(), prefixForType<TypeParam>(), prefixForType<TypeParam>());
+
+    auto trees = parseString(inputTrees);
+
+    ASSERT_NOTNULL(trees);
+
+    Tril::DefaultCompiler compiler{trees};
+    NoAndIlVerifier verifier;
```
Leonardo2718: I don't think this validator is appropriate for these tests. It should be fairly straightforward to create special ones for each of these.

0dvictor: Thank you for pointing it out. I've updated the verifies, do you mind take a look once more?

0dvictor:
> My only comment is that they should use IlVerifiers specific to these tests. NoAndIlVerifier just checks that a method contains no *and nodes, which is not relevant to these tests.

Thank you for pointing it out. I've updated the verifiers and hence now each test has a customized one. Do you mind take a look once more?

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+   virtual int32_t verifyNode(TR::Node *node) override
+      {
+      if (node->getOpCode().isAbs())
+         {
+         if (node->getChild(0)->getOpCode().isAbs())
+            {
+            return 1;
+            }
+         }
+      return 0;
```
Leonardo2718: Couldn't this be simplified to `return node->getOpCode().isAbs() && node->getChild(0)->getOpCode().isAbs()`?

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+   virtual int32_t verifyNode(TR::Node *node) override
+      {
+      if (node->getOpCode().isAbs())
+         {
+         if (node->getChild(0)->getOpCode().isNeg())
+            {
+            return 1;
+            }
+         }
+      return 0;
```
Leonardo2718: Couldn't this be simplified to `return node->getOpCode().isAbs() && node->getChild(0)->getOpCode().isNeg()`?

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+   virtual int32_t verifyNode(TR::Node *node) override
+      {
+      if (node->getOpCode().isNeg())
+         {
+         if (node->getChild(0)->getOpCode().isNeg())
+            {
+            return 1;
+            }
+         }
+      return 0;
```
Leonardo2718: Couldn't this be simplified to `return node->getOpCode().isNeg() && node->getChild(0)->getOpCode().isNeg()`?

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+       * using `addOptimization`, or add a group using
+       * `addOptimizations(omrCompilationStrategies[warm])`.
+       * This could also be done in test cases themselves.
+       */
+      addOptimization(OMR::treeSimplification);
+      }
+
+   };
+
+using Types = ::testing::Types<int32_t, int64_t, float, double>;
```
Leonardo2718: While I tend to agree that this kind of `using` declaration is easier to read than a `typedef`, it's a C++11 feature that I don't think is supported by all our build compilers.

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+TYPED_TEST_CASE(SimplifierFoldAbsNegTest, Types);
+
+
+class NoAbsAbsIlVerifier : public SimplifierFoldAbsNegTestIlVerifierBase
+   {
+   protected:
+   virtual int32_t verifyNode(TR::Node *node) override
```
Leonardo2718: I would love to see more uses of `override` in our code. Unfortunately, it is also one of those C++11 features that is not supported by all our build compilers. (Also, as a general comment, you should only have [exactly one of `virtual` or `override`](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override), not both :wink: )

0dvictor: I see, let me remove the `override` keyword.

> Also, as a general comment, you should only have exactly one of virtual or override, not both

It is an interesting topic, examples from MSDN use both https://msdn.microsoft.com/en-us/library/jj678987.aspx

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
0dvictor: Good idea, updated.

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
0dvictor: Good idea, updated.

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
0dvictor: Good idea, updated.

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
0dvictor: Thank you for point it out, I've updated this part to `typedef`.

fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
0dvictor: I've updated these functions not using `override` keywords.

mstoodle:
genie-omr build all

mstoodle:
genie-omr build all

mstoodle:
ok, that wasn't good enough. @0dvictor would you please force push to get the IP validation on the books, and maybe the appveyor test will also choose to proceed just for completeness. Sorry for the CI churn!

0dvictor:
@mstoodle Sure, I've rebased the changeset. IP validation, Travis-CI and AppVeyor have started.

vijaysun-omr:
Thanks @0dvictor for adding the tests. The compiler changes looked fine too. Merging.

PR #2736, opened 2018-07-10
Code cleanup related to X86 OutlinedInstruction

0dvictor:
1. Removing deprecated code;
2. Completing return result copying support.


vijaysun-omr:
Could you please elaborate on what is being cleaned up ? 
I think I need a longer description to tie it to the code changes.

0dvictor:
Updated.

compiler/x/codegen/OutlinedInstructions.cpp
```diff
@@ -175,18 +120,34 @@ void TR_OutlinedInstructions::generateOutlinedInstructionsDispatch()
...
-         generateRegRegInstruction(_targetRegMovOpcode, _callNode, targetRegPair->getLowOrder(),  resultRegPair->getLowOrder(),  _cg);
-         generateRegRegInstruction(_targetRegMovOpcode, _callNode, targetRegPair->getHighOrder(), resultRegPair->getHighOrder(), _cg);
+         generateRegRegInstruction(MOVRegReg(), _callNode, targetRegPair->getLowOrder(),  resultRegPair->getLowOrder(),  _cg);
+         generateRegRegInstruction(MOVRegReg(), _callNode, targetRegPair->getHighOrder(), resultRegPair->getHighOrder(), _cg);
          }
       else
          {
-         TR_ASSERT(!resultRegPair, "OutlinedInstructions: resultReg is a register pair and targetReg is not");
-         generateRegRegInstruction(_targetRegMovOpcode, _callNode, _targetReg, resultReg, _cg);
+         switch (resultReg->getKind())
```
vijaysun-omr: This switch seems to not fall in either category of change you mentioned in the commit message. Could you please clarify what this code is for ?

0dvictor: The previous version does not copy the result correctly for fcall/dcall/vcall, therefore I fix it.
In case if you are courious, the existing codes in OMR and known downstream projects do not use fcall/dcall/vcall with outlined instructions.

vijaysun-omr: Ok thanks for clarifying the reason (it's more than unused code cleanup).

vijaysun-omr:
@0xdaryl could you please review this too ?

compiler/x/codegen/OutlinedInstructions.cpp
```diff
@@ -175,18 +120,34 @@ void TR_OutlinedInstructions::generateOutlinedInstructionsDispatch()
    if (_targetReg)
       {
       TR_ASSERT(resultReg, "assertion failure");
+      TR_ASSERT(resultReg->getKind() == _targetReg->getKind(), "OutlinedInstructions: targetReg must have same time as resultReg.");
```
0xdaryl: Do you mean "same kind" rather than "same time" ?

compiler/x/codegen/OutlinedInstructions.cpp
```diff
@@ -175,18 +120,34 @@ void TR_OutlinedInstructions::generateOutlinedInstructionsDispatch()
...
+               generateRegRegInstruction(MOVRegReg(), _callNode, _targetReg, resultReg, _cg);
+               break;
+            case TR_FPR:
+            case TR_VRF:
+               generateRegRegInstruction(MOVDQURegReg, _callNode, _targetReg, resultReg, _cg);
+               break;
+            default:
+               TR_ASSERT(false, "OutlinedInstructions: unsupported result register kind.");
+               break;
+            }
```
0xdaryl: You'll save some code space by having the `switch` just select the opcode and then having one `generateRegRegInstruction` call after the switch.

compiler/x/codegen/OutlinedInstructions.cpp
0dvictor: Good catch, it has been updated.

compiler/x/codegen/OutlinedInstructions.cpp
0dvictor: Sure, updated.

0dvictor:
@0xdaryl I've updated the proposed changes according to your request, could you take one more look at it?

PR #2737, opened 2018-07-10
Common up logic for handling circular register dependencies between P and Z codegens

fjeremic:
Currently as part of #2698 we introduced a change to the Z codegen to violate the blocked register state and purposely evict spill a blocked register when we encounter a circular register dependency with all assignable registers blocked.

To avoid violating the definition of blocked registers we implement the same algorithm we have in the Power codegen for handling this case. Namely to handle circular dependencies, we block a real register if:

1. It is already assigned to a correct virtual register or
2. If it is assigned to one register in the list but is required by another

However, if all available registers are requested, we do not block in case 2. to avoid all registers being blocked.

To implement this algorithm we had to common up some utility classes into the common register dependency implementations. The commit messages detail what has been done. I'll further outline some of the changes via self review and pose some questions to other members of the community for things I've encountered while implementing this.


fjeremic:
@0xdaryl could I please get a review for the overall platform code commoning and documentation? @ymanton could I please get a review for the Power portion of the change. Note it should effectively be a NOP in terms of semantics. I merely documented the API and upstreamed it to a common place. @dchopra001 / @joransiu could I please get a review for the Z portion of the change since you have been involved with the fix to #2698.

compiler/z/codegen/OMRRegisterDependencyStruct.hpp
```diff
@@ -42,59 +42,16 @@ namespace OMR { typedef OMR::Z::RegisterDependencyExt RegisterDependency; }
...
 namespace OMR
 {
 namespace Z
 {
-struct RegisterDependencyExt
+struct RegisterDependencyExt : OMR::RegisterDependencyExt
    {
-   uint32_t                   _realRegister:7;
-   uint32_t                   _flags:2;
-   uint32_t                   _virtualRegIndex:23;
```
fjeremic: @ymanton @0dvictor @andrewcraik @0xdaryl note that the Z codegen was actually different than all the other codegens in this respect. We used a bitflag to represent a register dependency to save on space. Each register dependency is 32-bits in memory. We did this with a tradeoff to access the real register will require an extra indirection via `cg->getRegisterArray()`

I've changed it now to do it in the same way as Power and x86, though I'm not sure which is the better way as the size of each dependency is now 128-bits on 64-bit platforms (one pointer and two integer fields which will be aligned). So it is a trade off between space vs. time in accessing the real register. Which do we think is better overall?

compiler/z/codegen/OMRRegisterDependencyStruct.hpp
```diff
@@ -42,59 +42,16 @@ namespace OMR { typedef OMR::Z::RegisterDependencyExt RegisterDependency; }
...
-   uint32_t resetRefsRegister() {return _flags &= ~ReferencesDependentRegister;}
-
-   TR::Register * getRegister(TR::CodeGenerator *cg)
-      {
-       if (_virtualRegIndex==BAD_VIRTUAL_IDX)
-          return NULL;
-       else
-          return (cg->getRegisterArray())[_virtualRegIndex];
-      }
+   TR::RealRegister::RegNum  _realRegister;
```
fjeremic: @0xdaryl note this can and should be commoned however I think x86 is blocking this at the moment because we do not use `RegNum` (last I looked) in the x86 codegen.

fjeremic:
Note there may be failures on non-Z platforms at the moment. I wanted to get the ball rolling on the reviews. I'll be fixing the compile issues momentarily as the x86 and Power changes should semantically be NOPs.

fjeremic:
@0xdaryl @ymanton @dchopra001 please see review requests in https://github.com/eclipse/omr/pull/2737#issuecomment-403919152. This change is blocking further progress in improving register allocation on Z so reviews would be greatly appreciated to move this guy along. Thank you!

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -2609,21 +2609,59 @@ OMR::Z::CodeGenerator::prepareRegistersForAssignment()
...
+      TR::RealRegister* realReg = machine->getS390RealRegister((TR::RealRegister::RegNum)i);
+      if (realReg->getState() == TR::RealRegister::Locked)
+         ++lockedVRFs;
+      }
+
+   machine->setNumberOfLockedRegisters(TR_GPR, lockedGPRs);
+   machine->setNumberOfLockedRegisters(TR_HPR, lockedHPRs);
+   machine->setNumberOfLockedRegisters(TR_FPR, lockedFPRs);
+   machine->setNumberOfLockedRegisters(TR_VRF, lockedVRFs);
+
```
ymanton: I think I realized at some point in the past that we really don't need to count the number of locked regs because they are locked from the outset of the compilation and never change. It was a minor issue that was never really revisited, but since you're in this part of the code base doing good things can you look into it? The locked regs are determined by the linkage; each linkage sets the register file up in the state that it wants, so that might be a good place to count the number of locked regs as well.

fjeremic: @ymanton indeed this is how it is on Power. We simply initialize with a constant. Unfortunately on Z it is a little bit more complicated since there are many places where we lock register depending on the state of the JIT and whether some optimizations ran [1] [2] [3].

With this in mind the most sane thing to do is just do a single pass through the real registers and count them up for each compilation. I agree this is not the most efficient thing to do, but the alternative would be to keep incrementing and decrementing the counts via `getNumberOfLockedRegisters` / `setNumberOfLockedRegisters` each time we change a locked state.

I will be simplifying the logic in this area in following PRs so hopefully we can get it to a manageable state so we can do exactly what Power does and simply initialize the number of locked registers with a constant. This is definitely in my crosshairs.

[1] https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/z/codegen/OMRCodeGenerator.cpp#L2568-L2610

[2] https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/z/codegen/SystemLinkage.cpp#L2327-L2336

[3] https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/z/codegen/SystemLinkage.cpp#L90-L121

compiler/codegen/OMRMachine.hpp
```diff
@@ -46,21 +47,79 @@ class OMR_EXTENSIBLE Machine
...
 
-   Machine(TR::CodeGenerator *cg) :
-         _cg(cg)
+   Machine(TR::CodeGenerator *cg)
+      :
+      _cg(cg),
+      numLockedGPRs(-1),
+      numLockedHPRs(-1),
+      numLockedFPRs(-1),
+      numLockedVRFs(-1)
```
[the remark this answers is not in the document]
fjeremic: I think it is because of the asserts we have on Power in OMRCodeGenerator.cpp which in a debug build will count the number of locked registers and validate that whatever is returned by `getNumberOfLockedRegisters` is actually correct. The counts there start at 0 hence you will never get false positives, i.e. you never initialized the numbers via `getNumberOfLockedRegisters` but it just so happens that you had no locked registers. This is still a bug.

compiler/z/codegen/OMRCodeGenerator.cpp
ymanton: OK, that's a little different than what I assumed, but still salvageable. Is the number of locked regs known and fixed by the time we do register allocation? That's kind of what it looks like based on the above.

If so, we can/should still iterate through the reg file to count locked regs; what I was suggesting was that it doesn't need to be done every time we assign regs in a set of reg deps as it is now, it can simply be done once before RA. Based on what Z is doing it might be best to do it in common OMR code immediately before RA starts, that way we don't have to do it in every linkage in OMR and downstream and can accommodate cases where the number of locked regs is determined later (after the linkage initializes the register file) as well.

As long as all code generators agree on the meaning of `Locked`, that it should be for registers that don't participate in register allocation and that registers don't change from locked to some other state or vice versa during RA it's all good.

fjeremic: > what I was suggesting was that it doesn't need to be done every time we assign regs in a set of reg deps as it is now, it can simply be done once before RA

@ymanton this is exactly what happens with the above change. The code modified is in `OMR::Z::CodeGenerator::prepareRegistersForAssignment` which only gets called once per compilation before we do RA:

https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/codegen/OMRCodeGenPhase.cpp#L338-L352

ymanton: Yes, but it it's still in a Z-codegen-specific place, while on Power a hardcoded count is still in all of the linkages (both OMR and downstream) -- what I'm getting at (perhaps poorly) is that we can simply do the counting in a common pre-RA place, like ` OMR::CodeGenPhase::performRegisterAssigningPhase()`. It will introduce common behaviour on all platforms regardless of whether or not the count is a constant property of the linkage (like on Power) or whether it's somewhat more dynamic (like on Z), make it not possible for someone writing a new codegen or new linkage to forget to set it, etc. Does that make sense or am I misunderstanding?

ymanton: (I see a bit of where my own confusion came from, once upon a time on Power the locked regs used to be counted up in `TR_PPCRegisterDependencyGroup::assignRegisters()` but at some point we moved it into the linkage and got rid of the loop and left that bit of paranoid code that's now guarded by `#DEBUG`. For some reason I imagined that you somehow introduced that to Z. My mistake, sorry!)

vijaysun-omr:
@genie-omr build all

fjeremic:
@0dvictor FYI in case this is applicable to x86. Though I dare not step foot inside the x86 register allocator :joy:

vijaysun-omr:
@ymanton are you okay with this change ?

ymanton:
Yes, no objections.

PR #2738, opened 2018-07-10
Support both numa affinity and numa interleave memory independently

LinHu2016:
- numa_platform_supports_numa flag is only for identifying if numa
	affinity is enabled.
	- new numa_platform_interleave_memory flag for identifying if
	interleave memory is enabled.

Issue: eclipse/openj9#2367
	

charliegracie:
@genie-omr build all

PR #2746, opened 2018-07-12
Correctly set OMR_JITBUILDER_TEST flag in cmake.config

### LINKED ISSUE #2745: JitBuilder and Tril tests are not being run as part of the Travis CI builds
It appears the JitBuilder and Tril tests are not running as part of the regular Travis CI builds. The reason is that `OMR_JITBUILDER_TEST` is not currently enabled in `omr/cmake/caches/Travis.cmake`.

Leonardo2718:
This flag enables the JitBuilder and Tril tests in the Travis CI builds.

Fixes #2745

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -38,7 +38,6 @@ set(OMR_FVTEST ON CACHE BOOL "Enable the FV Testing.")
 set(OMR_GC ON CACHE BOOL "Enable the GC")
 set(OMR_JIT OFF CACHE BOOL "Enable building the JIT compiler")
 set(OMR_JITBUILDER OFF CACHE BOOL "Enable building JitBuilder")
-set(OMR_JITBUILDER_TEST OFF CACHE BOOL "Enable building JitBuilder Tests")
 set(OMR_OMRSIG ON CACHE BOOL "Enable the OMR signal compatibility library")
 set(OMR_PORT ON CACHE BOOL "Enable portability library")
 set(OMR_TEST_COMPILER OFF CACHE BOOL "Enable building the test compiler")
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build all

charliegracie:
Hmmm that is strange.  It should have been added if jitbuilder was enabled without requiring this change.

Leonardo2718:
Yeah, I noticed that. But, for some reason, it's not being enabled, even in local builds.

0xdaryl:
@genie-omr build all

charliegracie:
I would rather not merge this yet... I would like to know why the original change was not working

dnakamura:
ahh looked more closely at the old code to see what was going on. The original code did not work, because  `set( ... CACHE)` will not overwrite a value in the cache.  Since `OMR_JITBUILDER_TEST` is already given a default value (https://github.com/eclipse/omr/blob/master/cmake/config.cmake#L41),  later when we try to overwrite it (https://github.com/eclipse/omr/blob/master/cmake/config.cmake#L57) its a no-op.  One option is to change the default to: `set(OMR_JITBUILDER_TEST "${OMR_JITBUILDER}" CACHE BOOL "Enable building JitBuilder Tests")`, however it does have some limitations I'm still not crazy about (the tests dont later get auto-enabled if jitbuilder is enabled,  and its still possible to enable the tests without enabling jitbuilder). 
The best solution in this case is to probably use `cmake_dependent_option()`, which also has some limitations, which is why its not already being used, but the limitations are fairly minor

Leonardo2718:
> I believe the issue is that I set a default value for OMR_JITBUILDER_TEST to OFF a few lines above. Since the check is not forcing it on when OMR_JITBUILDER is set (which is the correct thing to do) we do not get the OMR_JITBUILDER_TEST enabled. I believe the proper fix is to delete setting the default value on line 41.

Does this mean that it's also not needed in `Appveyor.cmake` https://github.com/eclipse/omr/blob/170c4b4eb87bf7391a4360443bb8f8ee520f4f72/cmake/caches/Appveyor.cmake#L26 or is that one needed for a different reason?

charliegracie:
No AppVeyor should still have it forced to off in the cache.  On Windows we can compile JitBuilder but we can not compile the tests due to c++11 features that are used :( .  We need to fully remove the c++11 so we can support the tests on xlc and visual studio at some point.

0xdaryl:
@genie-omr build all

0xdaryl:
Change looks fine to me.  @charliegracie, can I get your thumbs up?

PR #2747, opened 2018-07-12
Re-add SIGQUIT portlib signal type support on Windows

babsingh:
In one of the previous changes, `SIGQUIT` portlib signal type was replaced
with `SIGBREAK` portlib signal type on Windows. This removed support for
`SIGQUIT` portlib signal type on Windows. Downstream projects register a
handler using the `SIGQUIT` portlib signal type on Windows. Without
`SIGQUIT` portlib signal type support on Windows, downstream projects will
break. In order to fix downstream projects, support for `SIGQUIT` portlib
signal type is re-added on Windows. `SIGBREAK` portlib signal type is not
used anymore so it is removed.


babsingh:
@charliegracie @pshipton @AdamBrousseau There is a two way dependency between https://github.com/eclipse/omr/pull/2716 and https://github.com/eclipse/openj9/pull/2293. OMR Acceptance build won't pass until it has https://github.com/eclipse/openj9/pull/2293. We can't merge https://github.com/eclipse/openj9/pull/2293 until https://github.com/eclipse/omr/pull/2716 promotes. So, the temporary fix in this PR will allow OMR Acceptance build to promote. Once https://github.com/eclipse/openj9/pull/2293 is merged in OpenJ9, we can remove this temporary fix.

babsingh:
I have set this PR to WIP. I will remove the WIP once my OpenJ9 personal builds pass.

charliegracie:
Can you change the commit message to describe that the previous change removed a signal type that is required by downstream projects and that we are adding the old signal type back so that down stream projects are not broken. I personally think that this is not a temporary solution but that it should probably stay.

babsingh:
If we are planning to support the old flag on Windows, then the new flag can be removed and we can just keep using the old flag. @charliegracie is this fine?

babsingh:
Taking a closer look, we won't be able to associate two portlib signal types to one OS signal. registration will work fine. but, the handler will not be invoked for one portlib signal type on win64amd. so, i think we should add support for the old SIGQUIT portlib signal type on Windows and remove the new SIGBREAK portlib signal type.

charliegracie:
Ok that is likely the best solution.  Thanks for looking into this

charliegracie:
@genie-omr build all

babsingh:
@charliegracie Removed WIP. Personal builds passed. Local signal testing also passed.

PR #2748, opened 2018-07-12
Fix z codegen generic load helper

NigelYiboYu:
Fix z generic load helper to use the correct generate APIs. The genericLoadHelper() looks up an array for the correct load instructions. The load instructions in the array come in different instruction formats and this helper is not using the correct API in all cases. This change corrects the APIs genericLoadHelper() uses.

Also, cleanup and refactor genericLoadHelper() and remove its architecture level support. The codebase's minimum supported z/Architecture level is z9 and genericLoadHelper() uses only instructions from z9 and below.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -4970,20 +4970,26 @@ genericLoadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference *
       TR::InstOpCode::Mnemonic load = loadInstrs[form][numberOfBytesLog2][isSourceSigned][numberOfExtendBits/32-1];
       if (form == RegReg)
          {
-         generateRRInstruction(cg, load, node, targetRegister, srcRegister);
+         if (TR::InstOpCode::getInstructionFormat(load) == RR_FORMAT)
+            generateRRInstruction(cg, load, node, targetRegister, srcRegister);
+         else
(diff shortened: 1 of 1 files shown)


NigelYiboYu:
FYI @fjeremic 
Marking this WIP until fully tested.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5058,20 +4970,26 @@ genericLoadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference *
       TR::InstOpCode::Mnemonic load = loadInstrs[form][numberOfBytesLog2][isSourceSigned][numberOfExtendBits/32-1];
       if (form == RegReg)
          {
-         generateRRInstruction(cg, load, node, targetRegister, srcRegister);
+         if (load == TR::InstOpCode::LR)
```
fjeremic: Are we able to make this check more generic and use the instruction formats already stored in the OMRInstOpCodeProperties.hpp tables? i.e. we can check if the instruction is RR format rather than checking a specific opcode. Let's also add an assert in the fallback that the instruction is in fact  an RRE format instruction.

NigelYiboYu: Fixed to use format checking e.g. `getInstructionFormat() == RX_FORMAT`.

The `loadInstrs` array only contains two formats for `RegReg` load and two formats for `MemReg` load. So no need to add assert for this.

fjeremic:
@genie-omr build zlinux,zos

PR #2752, opened 2018-07-12
Don't build useDef info in LoadExtensions at warm

dchopra001:
Building useDef info inside LoadExtensions is an expensive operation and doesn't yield a performance benefit for methods compiled at the warm optimization level. This change will now allow useDef info to only be built by load extensions at hot or higher optimization levels.

### DIFF
--- a/compiler/optimizer/LoadExtensions.cpp
+++ b/compiler/optimizer/LoadExtensions.cpp
@@ -62,7 +62,7 @@ TR_LoadExtensions::TR_LoadExtensions(TR::OptimizationManager *manager)
 
 int32_t TR_LoadExtensions::perform()
    {
-   if (comp()->getOptLevel() >= warm && !optimizer()->cantBuildGlobalsUseDefInfo())
+   if (comp()->getOptLevel() >= hot && !optimizer()->cantBuildGlobalsUseDefInfo())
       {
       if (!comp()->getFlowGraph()->getStructure())
          {


dchopra001:
@fjeremic

fjeremic:
@dchopra001 can you please post some data here justifying the performance benefits of this change for future reference?

dchopra001:
Sure. 

The benchmark I used was DayTrader7 and it was running using OpenJ9, which is a consumer of the OMR project, and hence the LoadExtensions optimization. 

Below is a summary of the performance of the benchmark when building useDef information was restricted at warm, hot, veryHot, and scorching compile levels:

screen shot 2018-07-12 at 11 14 40 pm

The change to achieve building useDef info at different optLevels was done here: 
https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/optimizer/LoadExtensions.cpp#L63-L67

Based on the above data, we can see that if we limit building useDef information at hot or higher compiles, we get a 1.9% improvement in compile time while not having any significant impact on throughput. So it does not make sense to enable this optimization for a warm compile, because there is little to no benefit in throughput, and a significant cost in compilation time. I'm also attaching the raw data here.
[useDefInLoadExtensions_Final.xlsx](https://github.com/eclipse/omr/files/2191180/useDefInLoadExtensions_Final.xlsx)


**Why this is enabled at hot or higher:** 
The third row, GRA Ticks, is the number of times we hit a use case where we actually took advantage (i.e. performed a transformation) of the usedef info built by the load extensions optimization. It was collected via the number of times we hit the following static debug counter:
https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/optimizer/LoadExtensions.cpp#L435-L438

The GRA ticks show that we barely get any hits of the transformation for methods compiled at `hot` or higher. Methods are predominantly compiled in the DayTrader7 benchmark at the warm optimization level. This explains why we barely get any ticks at optimization levels higher than warm (because not much is compiled at those opt levels!). Hence, this set of performance runs is effectively telling us that we do not see a benefit in throughput, while paying a significant amount in compilation time, when building useDef information at the warm optimization level. There isn't enough data provided by these runs to determine what the ideal optLevel is where we _should_ enable useDef building (we would need another benchmark that compiles more methods at those optLevels for that). So for now, I have enabled building that information at `hot` or higher optLevels. This way we get the compile time improvement on a warm benchmark like DayTrader7, but also don't negatively impact any workload that might be benefitting from useDef info being built at `hot` or higher.

fjeremic:
@genie-omr build all

PR #2756, opened 2018-07-13
Revert "Call configure using the CONFIG_SHELL variable"

dnakamura:
Reverts eclipse/omr#2647

Code did not ultimately solve the problem it intended to.  Also the generated configure script will
re-launch itself with $CONFIG_SHELL if set (autoconf documentation indicated that feature was not yet
implemented).

### DIFF
--- a/run_configure.mk
+++ b/run_configure.mk
@@ -47,10 +47,6 @@ help:
 # Remove all built-in rules
 .SUFFIXES:
 
-
-# Set a default value for CONFIG_SHELL
-CONFIG_SHELL ?= sh
-
 ###
 ### SPEC Specific Configure Arguments
 ###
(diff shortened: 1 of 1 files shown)


dnakamura:
Behavior can be verified
```
$ sed -i '2i printf "Configure script running in %s\\n" \$(cat /proc/$$/cmdline)' configure
$ ./configure 
Configure script running in /bin/sh./configure
Configure script running in /bin/bash./configure
Configure script running in /bin/bash./configure
```

0xdaryl:
FYI... @shivammittal99 confirms this fixes his OpenJ9 build problem where `tracegen` and `hookgen` were invoked before they were finished building when running with more than one build job.

PR #2757, opened 2018-07-15
Add matched user-defined placement delete operator through the compiler code

samolisov:
When memory is allocated for an object with operator new, the object's
constructor is called. If the constructor throws an exception, any
memory that was allocated for the object should be deallocated. This
cannot take place unless an operator `delete` function exists that matches
the operator `new`.

The compiler code generates a lot of **C4291** warnings because no placement
form of operator `delete` has been defined that matches the placement form
of operator `new`.

The commit adds missed placement forms of operator `delete` to the
compiler code.

### DIFF
--- a/compiler/env/TRMemory.hpp
+++ b/compiler/env/TRMemory.hpp
@@ -746,10 +746,14 @@ TR_HeapMemory::allocate(size_t size, TR_MemoryBase::ObjectType ot)
    static void   jitPersistentFree(void *mem)                                 {TR_Memory::jitPersistentFree(mem); }
 
 #define TR_PERSISTENT_NEW(a) \
-   void * operator new   (size_t s, PERSISTENT_NEW_DECLARE)   throw()   {return TR_Memory::jitPersistentAlloc(s, a);} \
-   void * operator new[] (size_t s, PERSISTENT_NEW_DECLARE)   throw()   {return TR_Memory::jitPersistentAlloc(s, a);} \
-   void * operator new   (size_t s, TR_PersistentMemory * m)  throw()   {return m->allocatePersistentMemory(s, a);} \
-   void * operator new[] (size_t s, TR_PersistentMemory * m)  throw()   {return m->allocatePersistentMemory(s, a);} \
+   void * operator new    (size_t s, PERSISTENT_NEW_DECLARE)   throw()   { return TR_Memory::jitPersistentAlloc(s, a); } \
+   void operator delete   (void *p, PERSISTENT_NEW_DECLARE)    throw()   { TR_Memory::jitPersistentFree(p); } \
+   void * operator new[]  (size_t s, PERSISTENT_NEW_DECLARE)   throw()   { return TR_Memory::jitPersistentAlloc(s, a); } \
+   void operator delete[] (void *p, PERSISTENT_NEW_DECLARE)    throw()   { TR_Memory::jitPersistentFree(p); } \
+   void * operator new    (size_t s, TR_PersistentMemory * m)  throw()   { return m->allocatePersistentMemory(s, a); } \
+   void operator delete   (void *p, TR_PersistentMemory *m)    throw()   { m->freePersistentMemory(p); } \
+   void * operator new[]  (size_t s, TR_PersistentMemory * m)  throw()   { return m->allocatePersistentMemory(s, a); } \
+   void operator delete[] (void *p, TR_PersistentMemory *m)    throw()   { m->freePersistentMemory(p); } \
    void operator delete  (void *p, size_t s) throw() { TR_ASSERT(false, "Invalid use of operator delete"); }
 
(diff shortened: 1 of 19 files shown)


samolisov:
So if we have a look in the `UseDefInfo.hpp` file, we will see the following pattern:

```cpp
   static void *operator new(size_t size, TR::Allocator a)
      { return a.allocate(size); }
   static void  operator delete(void *ptr, size_t size)
      { ((TR_UseDefInfo*)ptr)->allocator().deallocate(ptr, size); } /* t->allocator() better return the same allocator as used for new */
```
The matched form of the operator `new(size_t size, TR::Allocator a)` will have the following signature: `delete(void *ptr, TR::Allocator a)` and we would have been able to call the `a.allocate(size);` if the `size` had been saved somewhare. So the question is: would it be a good idea to store the `size` in a field when the operator `new` is invoked?

samolisov:
I've removed the ` OMROptimizationManager.hpp`  from the list of solved compilation units: I think it's not a good idea implement the `new/delete` pair for each extensible class.

fjeremic:
@dsouzai is definitely the most experienced in the area of memory management in OMR. I think he is away for another week however. Hopefully we can sit tight with this change @samolisov. Sorry for the delay. I'll kick off some builds in the mean while.

@genie-omr build all

samolisov:
@fjeremic thank you for the information.

Unfortunately, some builds fail, I'll work on them. Also some ideas about how to handle the situation I've described in the first comment are welcome.

samolisov:
I save the size into additional bytes allocated along with the class
instance:

```cpp
static void *operator new(size_t size, TR::Allocator a)
   {
   size_t allocated_size = size + sizeof(size_t);
   void *ptr = a.allocate(size);
   *(reinterpret_cast<size_t *>(ptr)) = size;
   return reinterpret_cast<size_t *>(ptr) + 1;
   }
```

and then use it during the delete of the object:

```cpp
static void  operator delete(void *ptr, TR::Allocator a)
   {
   size_t *size_ptr = reinterpret_cast<size_t *>(ptr) - 1;
   size_t allocated_size = *size_ptr + sizeof(size_t);
   a.deallocate(size_ptr, allocated_size);
   }
```

Please, review the file `OMROptimization.hpp`.

samolisov:
I have no idea why the AppVeyor build is failed, can someone re-trigger my PR building?

fjeremic:
> I have no idea why the AppVeyor build is failed, can someone re-trigger my PR building?

I'm not sure we can re-trigger those. A force push will do it though.

charliegracie:
I restarted it :)

dsouzai:
Hi @samolisov,

Sorry for the delay, I just got back to workin' today. 

I believe there isn't a need to have the `SizeAwareAllocatorDecorator`. Your changes affect objects that exist only during a compilation. If there is an exception through during construction, the compilation will be aborted, and all memory associated with that compilation will get freed (when `TR::SegmentAllocator &scratchSegmentProvider` goes out of scope).

That said, in order to remove all the warnings, I'm fine with adding all the various placement `delete`s, but they should be empty and do nothing (perhaps containing a comment stating why they exist).

By allocating a header for the payload, `SizeAwareAllocatorDecorator` does something that should really be hidden away inside the allocator, and so, in my opinion anyway, increases the complexity of the allocator framework that exists. We should really change `TR::Allocator` from being a `CS2::shared_allocator < ThreadLocalAllocator >` and instead have it be a `TR::Region` (which does not care about the size). I haven't gotten around to doing this yet because of time constraints - it requires measuring the impact to performance.

samolisov:
@dsouzai Thank you very much for this brilliant explanation! I've reset my commits since the first one and just put the matched placement `delete`s as empty functions adding the reason why. One does not simply walk throughout the allocator hierarchy in OMR :)

I've changed the PR's description and removed the 'WIP' marker, I think the PR is ready for review.

dsouzai:
You're welcome :)

>One does not simply walk throughout the allocator hierarchy in OMR :)

Heh, I've written some documentation [here](https://github.com/eclipse/omr/blob/master/doc/compiler/memory/MemoryManager.md#compiler-memory-manager-hierarchy) and [here](https://github.com/eclipse/omr/blob/master/doc/compiler/memory/MemoryManager.md#how-the-compiler-allocates-memory) which gives a very high level idea of the memory manager if you're curious.

fjeremic:
@genie-omr build all

samolisov:
AppVeyor returns ` 5108 Warning(s) 0 Error(s)` and then dead. Travis just was not able to run a container. What a day!

fjeremic:
> AppVeyor returns 5108 Warning(s) 0 Error(s) and then dead. Travis just was not able to run a container. What a day!

OMR genie builds to the rescue :) I'm happy to merge this. Thanks @dsouzai for the reviews and @samolisov for the fix!

PR #2759, opened 2018-07-16
Add isEmpty() API to Checklist

cathyzhyi:
Add isEmpty() API to Checklist to check if the list contains
any elements at all.


fjeremic:
@genie-omr build all

PR #2766, opened 2018-07-18
WIP: Don't sink stores of pd2i* and pd2l* nodes

dchopra001:
Sinking stores of pd2i/pd2iOverflow and pd2l/pd2lOverflow breaks the assumption in the z codegen that all such nodes must have a BCDCHKHandlerLabel. This commit prevents Store Sinking from making such transformations and hence breaking this assumption.

### DIFF
--- a/compiler/optimizer/SinkStores.cpp
+++ b/compiler/optimizer/SinkStores.cpp
@@ -2460,6 +2460,16 @@ bool TR_SinkStores::treeIsSinkableStore(TR::Node *node, bool sinkIndirectLoads, 
             traceMsg(comp(), "         can't move store of pinning array reference or with UseOnlyAliases\n");
          return false;
          }
+#ifdef J9_PROJECT_SPECIFIC
+      if (node->getOpCode().isStore() && node->getFirstChild()->getOpCode().isBinaryCodedDecimalOp() &&
+              node->getFirstChild()->getOpCode().isConversion() &&
+             (node->getFirstChild()->getOpCode().getDataType() == TR::Int32 || node->getFirstChild()->getOpCode().getDataType() == TR::Int64))
+          {
+          if (trace())
+             traceMsg(comp(), "         can't move store of a BCD Conversion Operation\n");
+          return false;
+          }
+#endif
       }
 
    if (!comp()->cg()->getSupportsJavaFloatSemantics() &&


dchopra001:
Can I please get a review from one or more of @NigelYiboYu @fjeremic @vijaysun-omr

compiler/optimizer/SinkStores.cpp
```diff
@@ -2460,6 +2460,16 @@ bool TR_SinkStores::treeIsSinkableStore(TR::Node *node, bool sinkIndirectLoads,
             traceMsg(comp(), "         can't move store of pinning array reference or with UseOnlyAliases\n");
          return false;
          }
+#ifdef J9_PROJECT_SPECIFIC
+      if (node->getOpCode().isStore() && node->getFirstChild()->getOpCode().isBinaryCodedDecimalOp() &&
+              node->getFirstChild()->getOpCode().isConversion() &&
+             (node->getFirstChild()->getOpCode().getDataType() == TR::Int32 || node->getFirstChild()->getOpCode().getDataType() == TR::Int64))
```
fjeremic: I'm not a fan of introducing additional `J9_PROJECT_SPECIFIC` ifdefs into OMR. @0xdaryl FYI. I see that `node->dontEliminateStores()` is called a few lines above. This is implemented here:

https://github.com/eclipse/omr/blob/8bbe28cea066becdd271db94888c331ae6a4601d/compiler/il/OMRNode.cpp#L2306-L2316

It seems to query `OMRSymbol` which is an extensible class:

https://github.com/eclipse/omr/blob/8bbe28cea066becdd271db94888c331ae6a4601d/compiler/il/symbol/OMRSymbol.hpp#L70-L77

So that begs the question: Could we do this in the upstream OpenJ9 project without introducing additional `J9_PROJECT_SPECIFIC` into the OMR codebase?

fjeremic: In addition it seems this change may be too strict. Rather than checking for `TR::Int32` and `TR::Int64` explicitly we should check simply that the node is of an integral type since this is the case you really care about. I'd also like to see a comment describing why this restriction is being put in place directly in the code. i.e. describe why sinking integral stores for packed operations which could potentially be under a BCDCHK operation is not safe.

0xdaryl: +1

dchopra001: @fjeremic Yes, this would work better. I think the best place we can do this in the `Node` class. I think we should be able to override `OMR::Node::dontEliminateStores` and have a `J9::Node::dontEliminateStores`. Then something like the following will probably work:
https://github.com/dchopra001/openj9/blob/f99987b1293c2bf4fc804b0bea03e5511b4a62f6/runtime/compiler/il/J9Node.cpp#L131-L138

vijaysun-omr:
@dchopra001 just checking that this is still intended to be WIP at the moment, i.e. that there are changes you are planning ?

vijaysun-omr:
Any update on this @dchopra001 ?

vijaysun-omr:
@shivammittal99 any update on this ? It is still marked WIP and I just wanted to make sure that was accurate.

fjeremic:
Assigning myself to this one as well so it doesn't slip my radar. I believe @dchopra001 is away for another week or so, so he won't get to this until then.

fjeremic:
@dchopra001 I think this is still a real bug which we should fix. Have you had any time to make progress on it?

dchopra001:
@fjeremic I wasn't able to reproduce this problem due to infrastructure issues. Those issues seem to have been resolved now. The fix is ready. I just need to test it to ensure that it solves the problem. I'll get that done this week.

fjeremic:
> @fjeremic I wasn't able to reproduce this problem due to infrastructure issues. Those issues seem to have been resolved now. The fix is ready. I just need to test it to ensure that it solves the problem. I'll get that done this week.

Ok, please keep in mind the above review comment. This code really belongs in OpenJ9.

charliegracie:
@dchopra001 ping?

dchopra001:
@charliegracie This is a failure that occurs on z14 and higher machines at scorching compile levels for DAA tests. Due to ongoing infrastructure issues, I have been unable to either reproduce this in the right environment or perform any post-mortem analysis of a core dump.

vijaysun-omr:
@dchopra001 any update on this ?

fjeremic:
@dchopra001 can we migrate this code to OpenJ9 as per the review? This is definitely still a bug. Even though we cannot reproduce the original problem it is still a lurking issue which we ought to fix.

dchopra001:
I'm going to close this PR and re-open in OpenJ9 once I have a reproduction.

PR #2769, opened 2018-07-18
WIP: Remove deferredVMAccessRelease

amicic:
Now that MasterGCThread class does not use deferred VM access mechanism
(https://github.com/eclipse/omr/pull/2696), we can remove the argument
from relinquishExclusiveVMAccess call.


charliegracie:
@genie-omr build all

charliegracie:
Will this break downstream projects?

amicic:
Prep work not to break OpenJ9 has been done: https://github.com/eclipse/openj9/pull/2423
Noone else should be affected.

PR #2770, opened 2018-07-18
Keeping size of LOA is bigger than largeObjectMinimumSize or 0

LinHu2016:
- maintain new LOA size either bigger than largeObjectMinimumSize or 0
during LOA resize(both expand and contract cases), the size, which is
smaller than largeObjectMinimumSize, is meaningless and could
potentially trigger the assertion. 
	- create inline method for minimizing duplicated logic


dmitripivkine:
@amicic FYI

amicic:
I'm fine with change

dmitripivkine:
Looks good to me

charliegracie:
@genie-omr build all

PR #2771, opened 2018-07-18
Print float/double more precisely in trace logs

0dvictor:
Print float/double in scientific notation when necessary, as well as its raw bits.

### DIFF
--- a/compiler/ras/Tree.cpp
+++ b/compiler/ras/Tree.cpp
@@ -184,12 +184,13 @@ static bool valueIsProbablyHex(TR::Node *node)
    }
 
 void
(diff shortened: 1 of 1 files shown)


fjeremic:
@r30shah given your recent experience in the area could you please review?

0dvictor:
This is to resolve the frustration that @r30shah and I have encountered: we saw a `dconst 0.0000` in the trace log but its actual value was not zero.

0dvictor:
By printing the bits, we can inspect the very details of the value. There exist multiple different bits that maps to the same floating point value, for example `NaN`.

PR #2772, opened 2018-07-18
Arbitrary size data snippet support on X86

0dvictor:
X86 now can have data snippets with arbitrary size.


0xdaryl:
@genie-omr build win

compiler/x/codegen/OMRCodeGenerator.hpp
```diff
@@ -594,7 +594,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
 
    bool nodeIsFoldableMemOperand(TR::Node *node, TR::Node *parent, TR_RegisterPressureState *state);
 
-   TR::IA32ConstantDataSnippet     *findOrCreateConstant(TR::Node *, void *c, uint8_t size);
+   TR::IA32DataSnippet         *createData(TR::Node *, void *c, uint8_t size);
```
0xdaryl: Doxygen comments for these two functions please.

0xdaryl: If this is the API for creating an arbitrary size data snippet (the small 4,8,16 byte primitives have a handy wrapper), I wish the name was a little more descriptive than just `createData`.  Even `createDataSnippet` might be better.

0xdaryl:
This class should really be renamed to `X86ConstantDataSnippet`.  Perhaps you could create an OMR issue to track that as well (it won't affect the merging of this PR).

compiler/x/codegen/OMRCodeGenerator.hpp
0dvictor: Good suggestion, I was just to keep the old names. They are updated now.

0dvictor: Updated.

0dvictor:
> This class should really be renamed to X86ConstantDataSnippet. Perhaps you could create an OMR issue to track that as well (it won't affect the merging of this PR).

An Issue has been raised: https://github.com/eclipse/omr/issues/2792

0xdaryl:
@genie-omr build win

PR #2775, opened 2018-07-19
Store a pointer to the original regex string in SimpleRegex

nwoeanhinnogaehr:
This pointer only remains valid for the lifetime of the original string. However, in practice SimpleRegex objects are only created from strings in enviroment variables or command line options, which have a static lifetime.

Needed to pass regex options from client to server for JITaaS.

Signed-off-by: Noah Weninger <noah.weninger@ibm.com>

### DIFF
--- a/compiler/infra/SimpleRegex.cpp
+++ b/compiler/infra/SimpleRegex.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


0xdaryl:
@genie-omr build all

0xdaryl:
Apologies for being obtuse, but can you elaborate on why the original string is still needed?

nwoeanhinnogaehr:
Sure, I should have been more clear. When we do a remote compilation in JITaaS, we want to send debug regex values like `onlyInline` from the client to the server. As it is today, the SimpleRegex class would be difficult to serialize (mutual recursion) and would take up a lot of space (in the Component struct, there is `uint64_t bit_map[BWORD(256)]`).

Sending the original string across the wire is much simpler. Apart from this implementation, the other options are to modify the parsing code (also in OMR) to store a pointer to the string somewhere else, or to duplicate and edit all of the parsing code for JITaaS. I chose this as it seems to be the cleanest of the options.

0xdaryl:
CI failure on zOS is an infra issue.  Merging.

PR #2777, opened 2018-07-20
Avoid generating MVC for unresolved symbols

NigelYiboYu:
The store operations can be optimized to use just a MVC instruction
instead of a load-store operation to save one instruction. For store
instructions, an UnresolvedDataSnippets is needed so that it can
resolve the symbols at program run-time and patch that store
instruction.

Run-time store instruction patching does not work with MVC. Hence, avoid
generating MVC instructions if the source or target memory locations
are from unresolved symbols.


NigelYiboYu:
Marking this WIP until fully tested.

fjeremic:
@genie-omr build zlinux,zos

Issue #2761, opened 2018-07-17
Aarch64: Duplicated symbols in ARM64Ops.hpp


knn-k:
I found some more duplicated symbols in `omr/compiler/aarch64/codegen/ARM64Ops.hpp` merged with #2600 as shown below.

`sturb` and `ldurb`:
```
	/* Load/store register (unscaled immediate) */
		sturb,                                                          	/* 0x38000000	STURB     	 */
		ldurb,                                                          	/* 0x38400000	LDURB     	 */
		ldursbx,                                                        	/* 0x38800000	LDURSB    	 */
		ldursbw,                                                        	/* 0x38C00000	LDURSB    	 */
		sturb,                                                          	/* 0x3C000000	STUR      	 */
		ldurb,                                                          	/* 0x3C400000	LDUR      	 */
```

`sturh` and `ldurh`:
```
		sturh,                                                          	/* 0x78000000	STURH     	 */
		ldurh,                                                          	/* 0x78400000	LDURH     	 */
		ldurshx,                                                        	/* 0x78800000	LDURSH    	 */
		ldurshw,                                                        	/* 0x78C00000	LDURSH    	 */
		sturh,                                                          	/* 0x7C000000	STUR      	 */
		ldurh,                                                          	/* 0x7C400000	LDUR      	 */
```

knn-k:
FYI. @aarongraham9 and @jeanlego

aarongraham9:
Thanks for the good find @knn-k! Is this an issue that's blocking you or something that you need imminently for your own PR(s)? Otherwise we'll just correct them in the next Binary Encoding related PR (which work on should start shortly anyway).

knn-k:
@aarongraham9 I am going to make some changes in OpBinary.cpp in #2754, and perhaps it will affect ARM64Ops.hpp, too.  But I don't need this issue to be fixed immediately.

knn-k:
The STURB and LDURB instructions are encoded as `0x38000000` and `0x38400000`, respectively.  I am trying to identify the instructions with the encoding `0x3C000000` (STUR ?) and `0x3C400000` (LDUR ?), but not successful so far.

It is the same for STURH/LDURH.

aarongraham9:
This can be closed via #2779.

PR #2779, opened 2018-07-20
AArch64: This Fixes duplicate opcode in ARM64Ops

jeanlego:
this should fix error #2761 
by renaming sturb and ldurb duplicate name
 for clarity we use the "_bt" notation when accessing a byte of the
register and only "b" when the target is the byte size fp register


0xdaryl:
For the cases where the SIMD mnemonics are aliased with the non-SIMD mnemonics, what do you think about prefixing the SIMD-form instructions with a `v` instead?  It makes it somewhat consistent with the other vector (SIMD) instructions, and is more intuitive when choosing the right opcode (in my mind at least).  For instance,
```
vsturb (SIMD&FP)
sturb
```

0xdaryl:
^^^ @knn-k @jeanlego @aarongraham9 @JamesKingdon

knn-k:
I agree with adding the prefix "v" for SIMD and FP instructions.

jeanlego:
added the changes proposed by @0xdaryl 
should be good to go! @aarongraham9 ? @knn-k

knn-k:
Looks good to me.

aarongraham9:
LGTM as well @jeanlego and @0xdaryl.

PR #2780, opened 2018-07-20
Hoist registerFile to base Machine class

0xdaryl:
Hoist the registerFile array to the common base class to
share the implementation.

### DIFF
--- a/compiler/codegen/OMRMachine.hpp
+++ b/compiler/codegen/OMRMachine.hpp
@@ -34,6 +34,7 @@ namespace OMR { typedef OMR::Machine MachineConnector; }
 #include <stddef.h>                            // for NULL
 #include <stdint.h>                            // for uint8_t, uint32_t
 #include "codegen/RegisterConstants.hpp"
+#include "codegen/RealRegister.hpp"
 #include "env/TRMemory.hpp"                    // for TR_Memory, etc
 #include "infra/Annotations.hpp"               // for OMR_EXTENSIBLE
 
@@ -45,7 +46,6 @@ namespace OMR
 
 class OMR_EXTENSIBLE Machine
    {
-   TR::CodeGenerator *_cg;
 
    int16_t numLockedGPRs;
    int16_t numLockedHPRs;
@@ -68,6 +68,9 @@ class OMR_EXTENSIBLE Machine
 
    inline TR::Machine * self();
 
+   /**
+    * \return : the cached TR::CodeGenerator object
+    */
    TR::CodeGenerator *cg() {return _cg;}
 
    /** \brief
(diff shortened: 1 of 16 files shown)


0xdaryl:
@fjeremic @ymanton @0dvictor @knn-k : I'd appreciate a review on your platform of expertise please.

0dvictor:
Changes on X86 LGTM.

0xdaryl:
@genie-omr build all

knn-k:
It looks OK for ARM32 and ARM64.

compiler/x/i386/codegen/OMRMachine.hpp
```diff
@@ -64,7 +64,6 @@ class OMR_EXTENSIBLE Machine : public OMR::X86::Machine
       IA32_MAX_GLOBAL_FPRS      = 8,
       };
 
-   TR::RealRegister  *_registerFileStorage[TR_X86_REGISTER_FILE_SIZE];
```
fjeremic: Should we deprecate `TR_X86_REGISTER_FILE_SIZE` as well? Will this not always equal `TR::RealRegister::NumRegisters`?

compiler/codegen/OMRMachine.hpp
```diff
@@ -119,6 +122,29 @@ class OMR_EXTENSIBLE Machine
...
+   TR::RealRegister **registerFile() { return _registerFile; }
+
+   /**
+    * \brief Retrieves the TR::RealRegister object for the given real register
+    *
+    * \param[in] rn : the desired real register
+    *
+    * \return : the desired TR::RealRegister object
+    */
+   TR::RealRegister *realRegister(TR::RealRegister::RegNum rn) { return _registerFile[rn]; }
```
fjeremic: This seems to deviate from the rest of the codebase which the majority of uses the `get*` syntax for getters. Personally if I am loking t oretrieve something from a class I type `get` in my IDE and find the right getter which I want. This would definitely stump me. I am ok with the change as long as this is the direction we're heading in. Thoughts?

compiler/z/codegen/OMRMachine.hpp
```diff
@@ -640,7 +639,7 @@ class OMR_EXTENSIBLE Machine : public OMR::Machine
       }
    bool supportLockedRegisterAssignment();
 
-   TR::RealRegister *getRegisterFile(int32_t i);
+   TR::RealRegister *realRegister(int32_t rn) { return _registerFile[(TR::RealRegister::RegNum)rn]; }
```
fjeremic: I don't think this is necessary. Won't we pick this API up from the extensible class? They are identical (modulo some casting needed for the argument).

compiler/codegen/OMRMachine.hpp
fjeremic: It seems we are indeed to quote @0xdaryl consistently inconsistent in the area of naming these getters. I'm fine with the change as is. However I look forward to the other two reviews being addressed.

fjeremic:
@0xdaryl just making sure you haven't forgot about this PR.

0xdaryl:
Thanks.  I haven't forgotten, just no time to work on it at the moment.  Hopefully by the weekend.

fjeremic:
Looks like all AIX nodes are off. Given the tests have succeeded on Linux PPC I'm comfortable merging this before the infrastructure issues are resolved.

PR #2783, opened 2018-07-23
Locking Changes

babsingh:
1) **Revert "Improve locking"**

	This commit reverts the commit in #1979.

	The purpose of the original change was to reduce compare and swap
	operations in the spinlock loop; thus, reducing spinning overhead. The
	premise was to improve throughput by reducing spinning overhead. In the
	initial benchmarking, no throughput improvements were noticed because of
	this change. Recently, this change has resulted in regressing
	throughput. ~30% throughput drop is observed in one of the OpenLiberty
	benchmarks. Thus, it has been decided to revert the original change.

### DIFF
--- a/thread/common/threadhelpers.cpp
+++ b/thread/common/threadhelpers.cpp
@@ -70,16 +70,20 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
 	uintptr_t spinCount2Init = monitor->spinCount2;
 	uintptr_t spinCount1Init = monitor->spinCount1;
 
-	uintptr_t spinCount3 = spinCount3Init;
-	uintptr_t spinCount2 = spinCount2Init;
-
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
+	BOOLEAN spinning = TRUE;
  	if (monitor->spinThreads < lib->maxSpinThreads) {
  		VM_AtomicSupport::add(&monitor->spinThreads, 1);
  	} else {
- 		goto exit;
+ 		spinCount1Init = 1;
+ 		spinCount2Init = 1;
+ 		spinCount3Init = 1;
+ 		spinning = FALSE;
  	}
 #endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
+
+	uintptr_t spinCount3 = spinCount3Init;
+	uintptr_t spinCount2 = spinCount2Init;
 
 	for (; spinCount3 > 0; spinCount3--) {
 		for (spinCount2 = spinCount2Init; spinCount2 > 0; spinCount2--) {
@@ -130,10 +134,11 @@ update_jlm:
 #endif /* OMR_THR_JLM */
 
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
-	VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
-
-exit:
+	if (spinning) {
+		VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
+	}
 #endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
+
 	return result;
 }
 


charliegracie:
This needs a lot more details before it can be merged.  The original change was made to improve locking performance with large numbers of threads. This code has shipped for many months so many users may be enjoying the performance wins from the original change.

babsingh:
@charliegracie I have added more details for reverting the change in the commit message.

fyi - @pshipton @vijaysun-omr

thread/common/threadhelpers.cpp
```diff
@@ -70,17 +70,23 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
...
-	uintptr_t spinCount2 = spinCount2Init;
-
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
- 	if (monitor->spinThreads < lib->maxSpinThreads) {
- 		VM_AtomicSupport::add(&monitor->spinThreads, 1);
- 	} else {
- 		goto exit;
- 	}
+	BOOLEAN spinning = TRUE;
+	if (OMRTHREAD_IGNORE_SPIN_THREAD_BOUND != lib->maxSpinThreads) {
```
charliegracie: What is the default for maxSpinThreads?

If maxSpinThreads == OMRTHREAD_IGNORE_SPIN_THREAD_BOUND then shouldn't the code still do the new behaviour?  If not what is the point of this new code?

thread/common/threadhelpers.cpp
```diff
@@ -70,17 +70,23 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
...
+		if (monitor->spinThreads < lib->maxSpinThreads) {
+			VM_AtomicSupport::add(&monitor->spinThreads, 1);
+		} else {
+			spinCount1Init = 1;
+			spinCount2Init = 1;
```
charliegracie: Why are you just not setting the affected variables directly here instead of setting the Init values?

thread/common/threadhelpers.cpp
```diff
@@ -70,17 +70,23 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
...
+			spinCount3Init = 1;
+			spinning = FALSE;
+		}
+	}
 #endif /* defined(OMR_THR_SPIN_WAKE_CONTROL) */
 
+	uintptr_t spinCount3 = spinCount3Init;
```
charliegracie: This is effectively turning spinning off correct?

thread/common/threadhelpers.cpp
```diff
@@ -130,10 +136,11 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
 #endif /* OMR_THR_JLM */
 
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
-	VM_AtomicSupport::subtract(&monitor->spinThreads, 1);
-
-exit:
+	if (spinning && (OMRTHREAD_IGNORE_SPIN_THREAD_BOUND != lib->maxSpinThreads)) {
```
charliegracie: I do not think this is correct, which could be causing monitor->spinThreads to be inflated.

babsingh:
@charliegracie I have backed out the changes for the new `-Xthr:maxSpinThreads=0` option. I will push the changes for the new option in a new PR.

thread/common/threadhelpers.cpp
babsingh: the purpose of `OMRTHREAD_IGNORE_SPIN_THREAD_BOUND != lib->maxSpinThreads` is to remove the bound on the number of spin threads, and allow unlimited spin threads. this takes us back to early 2017 or late 2016 when there was no bound on the number of spin threads. default on `maxSpinThreads` varies on the number of CPUs on a machine: https://github.com/eclipse/openj9/pull/553/commits/d4823cdb8f91e9219124b78115f2d91929d22c72.

thread/common/threadhelpers.cpp
babsingh: the Init values are needed for calculation of JLM counts at the end. Init values need to preserve the spin counts at the beginning of the spin loop.

thread/common/threadhelpers.cpp
babsingh: if `spinCount*Init`(s) are set to 1, then threads will attempt one CAS operation before exiting the spin loop. this is equivalent to turning spinning off. otherwise, `spinCount*Init`retain the global values 256/32/45.

thread/common/threadhelpers.cpp
babsingh: `(OMRTHREAD_IGNORE_SPIN_THREAD_BOUND != lib->maxSpinThreads)` - this should stop atomic increment/decrement of `monitor->spinThreads` and stop the (`monitor->spinThreads < lib->maxSpinThreads`) check to allow unlimited spin threads.

charliegracie:
@genie-omr build all

PR #2787, opened 2018-07-24
Create a temp for the Operand[stack/array] base name

charliegracie:
Currently the name for the operand stack/array base was always
Operand<Type>_Base which means you can only ever use one of them
per application. This change allows multiple to be used if the
consumer specifies a base name.

### DIFF
--- a/compiler/ilgen/OMRVirtualMachineOperandArray.cpp
+++ b/compiler/ilgen/OMRVirtualMachineOperandArray.cpp
@@ -40,13 +40,7 @@ OMR::VirtualMachineOperandArray::VirtualMachineOperandArray(TR::MethodBuilder *m
    _elementType(elementType),
    _arrayBaseRegister(arrayBaseRegister)
    {
-   int32_t numBytes = _numberOfElements * sizeof(TR::IlValue *);
-   _values = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
-   memset(_values, 0, numBytes);
-
-   // store current operand stack pointer base address so we can use it whenever we need
-   // to recreate the stack as the interpreter would have
-   mb->Store("OperandArray_base", arrayBaseRegister->Load(mb));
+   init();
    }
 
 OMR::VirtualMachineOperandArray::VirtualMachineOperandArray(TR::VirtualMachineOperandArray *other)
@@ -54,7 +48,8 @@ OMR::VirtualMachineOperandArray::VirtualMachineOperandArray(TR::VirtualMachineOp
    _mb(other->_mb),
    _numberOfElements(other->_numberOfElements),
    _elementType(other->_elementType),
-   _arrayBaseRegister(other->_arrayBaseRegister)
+   _arrayBaseRegister(other->_arrayBaseRegister),
+   _arrayBaseName(other->_arrayBaseName)
    {
    int32_t numBytes = _numberOfElements * sizeof(TR::IlValue *);
    _values = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build all

compiler/ilgen/OMRVirtualMachineOperandArray.hpp
```diff
@@ -73,6 +73,12 @@ class VirtualMachineOperandArray : public TR::VirtualMachineState
     * @param numOfElements the number of elements in the array
     */
    VirtualMachineOperandArray(TR::MethodBuilder *mb, int32_t numOfElements, TR::IlType *elementType, TR::VirtualMachineRegister *arrayBase);
+   /**
+    * @brief public constructor, must be instantiated inside a compilation because uses heap memory
+    * @param mb TR::MethodBuilder object of the method currently being compiled
+    * @param numOfElements the number of elements in the array
```
0xdaryl: Please add docs for the other parameters too: elementType, arrayBase, arrayBaseName

0xdaryl:
Just noticed now that @mstoodle was requested as a reviewer on this.  Mark, if you want to become the assignee on this one please take it from me.  Your review is still welcome either way.

charliegracie:
@genie-omr build all

charliegracie:
@0xdaryl comments have been updated

mstoodle:
@charliegracie can you expand on why you want this? I presume the goal is to support multiple operand stacks in the same method, right?

If that's the use case, maybe instead of choosing a particular name we should just create a temporary to hold the base address rather than forcing the user to come up with a name?

charliegracie:
It is to support using to OperandArrays at the same time so I figured why not add it to both... A temporary would be just fine.  I do not need to know the name.

charliegracie:
@mstoodle do you mean something like this?

charliegracie:
@genie-omr build all

compiler/ilgen/OMRVirtualMachineOperandArray.cpp
```diff
@@ -181,3 +176,26 @@ OMR::VirtualMachineOperandArray::Move(TR::IlBuilder *b, int32_t dstIndex, int32_
...
+   int32_t numBytes = _numberOfElements * sizeof(TR::IlValue *);
+   _values = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
+   memset(_values, 0, numBytes);
+
+   TR::Compilation *comp = TR::comp();
+   // Create a temp for the OperandArray base
+   TR::SymbolReference *symRef = comp->getSymRefTab()->createTemporary(_mb->methodSymbol(), _mb->typeDictionary()->PointerTo(_elementType)->getPrimitiveType());
+   symRef->getSymbol()->setNotCollected();
+   char *name = (char *) comp->trMemory()->allocateHeapMemory((2+10+1) * sizeof(char)); // 2 ("_T") + max 10 digits + trailing zero
+   sprintf(name, "_T%u", symRef->getCPIndex());
```
mstoodle: Maybe `_ArrayBase_%u` as the name rather than `_T%u` ?

compiler/ilgen/OMRVirtualMachineOperandStack.cpp
```diff
@@ -227,3 +222,26 @@ OMR::VirtualMachineOperandStack::grow(int32_t growAmount)
...
+   int32_t numBytes = _stackMax * sizeof(TR::IlValue *);
+   _stack = (TR::IlValue **) TR::comp()->trMemory()->allocateHeapMemory(numBytes);
+   memset(_stack, 0, numBytes);
+
+   TR::Compilation *comp = TR::comp();
+   // Create a temp for the OperandStack base
+   TR::SymbolReference *symRef = comp->getSymRefTab()->createTemporary(_mb->methodSymbol(), _mb->typeDictionary()->PointerTo(_elementType)->getPrimitiveType());
+   symRef->getSymbol()->setNotCollected();
+   char *name = (char *) comp->trMemory()->allocateHeapMemory((2+10+1) * sizeof(char)); // 2 ("_T") + max 10 digits + trailing zero
+   sprintf(name, "_T%u", symRef->getCPIndex());
```
mstoodle: ...and `_StackBase_%u` here?

charliegracie:
@genie-omr build all

charliegracie:
@mstoodle I believe it is good now :)

PR #2789, opened 2018-07-24
[CMake] Build an arch-specific compilertest only

samolisov:
Compilation units which specific for other environments, different from the current one, are
excluded from the compilertest cmake-managed build.


samolisov:
I've restored the original file format, only the changed lines have been added in the diff.

0xdaryl:
@Leonardo2718 : would you mind reviewing these proposed changes?

0xdaryl:
@genie-omr build all

PR #2793, opened 2018-07-24
Remove unused query for CMOV support on X86

0dvictor:
supportsCMOV() is not in-use and unnecessary, removing it.


0dvictor:
OMR requires SSE, and all chips that supports SSE also supports CMOV; therefore, querying for CMOV support is unnecessary.

0xdaryl:
@genie-omr build win

PR #2794, opened 2018-07-25
Add feature to ignore spin thread bound

babsingh:
In OMR, a user can set `lib->maxSpinThreads` to 0. In the current
implementation, no threads will spin if `lib->maxSpinThreads` is set to 0.
This is a redundant feature since spinning can also be disabled by
setting `lib->defaultMonitorSpinCount1/2/3` to 1.

So, the behavior for `lib->maxSpinThread = 0` is re-purposed. If
`lib->maxSpinThreads` is set to 0, then an upper bound won't be enforced
on spin threads, and unlimited spin threads will be allowed.

This is an experimental feature. A consumer of OMR can enable this
feature by setting `lib->maxSpinThread` to 0 via
`omrthread_global("maxSpinThreads")`. It will be used to create heuristics
for locking workloads which benefit from no bound on the number of spin
threads.

### DIFF
--- a/include_core/omrthread.h
+++ b/include_core/omrthread.h
@@ -109,6 +109,7 @@ or omrthread priority value.
 
 #define OMRTHREAD_MINIMUM_SPIN_THREADS 1
 #define OMRTHREAD_MINIMUM_WAKE_THREADS 1
+#define OMRTHREAD_IGNORE_SPIN_THREAD_BOUND 0
 
 #include "thread_api.h"
 
--- a/thread/common/threadhelpers.cpp
+++ b/thread/common/threadhelpers.cpp
@@ -72,14 +72,16 @@ omrthread_spinlock_acquire(omrthread_t self, omrthread_monitor_t monitor)
 
 #if defined(OMR_THR_SPIN_WAKE_CONTROL)
 	BOOLEAN spinning = TRUE;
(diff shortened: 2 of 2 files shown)


charliegracie:
@genie-omr build all

charliegracie:
Where is lib->maxSpinThread set? How was now allowing the value to be 0 enforced?

babsingh:
@charliegracie 
> Where is lib->maxSpinThread set?

`lib->maxSpinThreads` is initialized in **omrthread.c::init_spinParameters**.
```
  	lib->maxSpinThreads = OMRTHREAD_MINIMUM_SPIN_THREADS;
  	if (init_threadParam("maxSpinThreads", &lib->maxSpinThreads)) {
  		return -1;
  	}
```
In OMR, `lib->maxSpinThreads` is initialized to OMRTHREAD_MINIMUM_SPIN_THREADS (1). The consumer of OMR can override the value by setting `omrthread_global("maxSpinThreads") `.

> How was not allowing the value to be 0 enforced?
After reading code, this condition is not enforced in OMR but through OpenJ9. If `lib->maxSpinThreads` is set to 0 in OMR, then no threads would spin which is undesirable. I will update the commit message to accurately reflect the OMR changes.

babsingh:
@charliegracie I have updated the commit message to better justify the proposed changes.

PR #2795, opened 2018-07-25
Make runtime helper table class do lazy address translation

NigelYiboYu:
Runtime helper table no longer unwrap function pointers at initialization
time. It now keeps an array of function pointers and only unwraps the raw
entry points whenever a user calls the getAddress() API.

Also, add a getEnvironment() API to get the function pointers'
environments. This can be helpful for platforms whose C functions have
environment pointers.

### DIFF
--- a/compiler/runtime/Runtime.hpp
+++ b/compiler/runtime/Runtime.hpp
@@ -170,31 +170,70 @@ enum TR_RuntimeHelper
 class TR_RuntimeHelperTable
    {
 public:
-   void* getAddress(TR_RuntimeHelper h)
-      {
-      return h < TR_numRuntimeHelpers ? _helpers[h] : (void *) (uintptr_t) 0xDEADB00F;
-      }
+
+    static const intptrj_t INVALID_FUNCTION_POINTER = 0xdeadb00f;
+
+    /**
+     * \brief
+     *
+     * Looks up the given helper table and returns the function pointer if it's a function index.
+     * For constant number entries, return INVALID_FUNCTION_POINTER.
+     */
+    void* getFunctionPointer(TR_RuntimeHelper h);
+
+   /**
+    * \brief
+    *
+    * For helper functions, this API returns the raw entry address. If the runtime
+    * helper's corresponding linkage is not a function, return the helper as data.
+    *
+    */
+   void* getFunctionEntryPointOrConst(TR_RuntimeHelper h);
+
+   /**
+    * \brief
+    *
+    * Returns the linkage type of the given helper index. The linkage types are set during
+    * the JIT initilization phase.
+    */
    TR_LinkageConventions getLinkage(TR_RuntimeHelper h)
       {
       return h < TR_numRuntimeHelpers ? _linkage[h] : TR_None;
       }
-   // Linkage convention is essential when calling a helper
-   // For example, on X86, there are private linkage, System V ABI, fastcall, cdecl, etc.
+
+   /**
+    * \brief
+    *
+    * Sets the address and linkage convention.
+    *
+    * Linkage convention is essential when calling a helper
+    * For example, on X86, there are private linkage, System V ABI, fastcall, cdecl, etc.
+    */
    void setAddress(TR_RuntimeHelper h, void * a, TR_LinkageConventions lc = TR_Helper)
       {
-      _helpers[h] = translateAddress(a);
+      _helpers[h] = a;
       _linkage[h] = lc;
       }
 
+   /**
+    * \brief
+    *
+    * Stores the given pointer as a helper constant.
+    */
    void setConstant(TR_RuntimeHelper h, void * a)
       {
       _helpers[h] = a;
+      _linkage[h] = TR_None;
       }
 
 private:
-   // translate address is to allow each platform converting a C function pointer
-   // to an address callable by assembly, which is essential for P and Z.
+   /**
+    * \brief
+    * Translate address is to allow converting a C function pointer
+    * to an address callable by assembly.
+    */
    void* translateAddress(void* a);
+
    void*                 _helpers[TR_numRuntimeHelpers];
    TR_LinkageConventions _linkage[TR_numRuntimeHelpers];
    };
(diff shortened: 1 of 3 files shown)


NigelYiboYu:
This is the OMR portion of the runtime helper table change. The idea is to store function pointers on JIT initialization instead of storing unwrapped raw function entry addresses. In doing so, the JIT compiler is able to extract and use helper functions' environment pointers via the new `getEnvironment()` API. This is important for JITCODE to C function calls on PPC AIX and z/OS.

FYI @fjeremic

compiler/runtime/Runtime.hpp
```diff
@@ -170,31 +170,66 @@ enum TR_RuntimeHelper
...
+    * helper's corresponding linkage is not a function, return the helper as data.
+    */
+   void* getAddress(TR_RuntimeHelper h);
+
+   /**
+    * \brief
+    *
+    * Looks up the given helper index and returns the environment pointer, if there is any.
+    * The environment pointer of a function is applicable to PPC AIX and z/OS.
+    */
```
ymanton: I suggest removing the _applicable to ... AIX and z/OS_ part because it will go stale if/when other platforms that use this convention are added. For example, Linux ppc64 also has this.

Also, I'm a little confused about what's supposed to be returned here. Is it a pointer to the function descriptor? Or a particular field (that isn't the entry point because that's what `getAddress` seems to be doing)?

The "environment pointer" in PPC ABIs is a field in the function descriptor that's mostly useless to us at the moment, so that's another source of confusion, and `TOC_UNWRAP_ENV` part that you added seems to suggest you're returning a particular field after all. Is it the TOC pointer?

compiler/runtime/Runtime.hpp
```diff
@@ -170,31 +170,66 @@ enum TR_RuntimeHelper
...
       }
 
 private:
-   // translate address is to allow each platform converting a C function pointer
-   // to an address callable by assembly, which is essential for P and Z.
+   /**
+    * \brief
+    * Translate address is to allow each platform converting a C function pointer
+    * to an address callable by assembly, which is essential for P and Z.
+    */
```
ymanton: Same as above, "P and Z" might go stale.

compiler/runtime/Runtime.hpp
NigelYiboYu: Removed the statement `applicable to ... AIX and z/OS` and updated the comment to state explicitly that this returns the `environment field extracted from the function descriptor on applicable platforms`.

compiler/runtime/Runtime.hpp
NigelYiboYu: Fixed.

compiler/runtime/Runtime.hpp
NigelYiboYu: This goes without saying that the `getEnvironment()` is an empty stub in the OMR project. So it's another reason to remove the words on AIX and z/OS. The actual environment unwrapping should be project specific.

NigelYiboYu:
This change passed my tests. Ready to review.

compiler/runtime/Runtime.hpp
ymanton: Would it be better to only have APIs to return the function descriptor and the entry point, rather than a specific field and the entry point? That way the caller can get access to any field in the descriptor that they know and care about, rather than just one particular field. The way it is now is OK until someone decides they really want to access 2 or more fields from their descriptors or want access to the descriptor itself.

The most natural approach to me is to have `getFunctionPointer` and `getEntryPoint` and to have those return

* the function descriptor and entry point on platforms that use function descriptors, or
* the entry point and entry point on platforms that don't.

Anyone who wants to call the function from C will naturally use `getFunctionPointer`, anyone who knows about the platform ABI will know whether or not function pointers are pointers to descriptors or entry points and will use `getFunctionPointer` and know what fields are available in descriptors, and anyone who wants the actual entry point will call `getEntryPoint` without caring. The bonus is that you also have a convenient and obvious way to tell if function pointers are pointers to code by testing `getFunctionPointer() == getEntryPoint()`.

I'd also use `getEntryPoint` rather than `getAddress` because there is less room for confusion and most ABIs I've read refer to entry points.

compiler/runtime/Runtime.hpp
```diff
@@ -170,31 +170,69 @@ enum TR_RuntimeHelper
...
+   void* getAddress(TR_RuntimeHelper h);
+
+   /**
+    * \brief
+    *
+    * Looks up the given helper index and returns the environment pointer on applicable platforms.
+    * Certain platforms have function descriptors for each C function. And the environment pointer
+    * is a field in the function descriptor.
+    */
+   void* getEnvironment(TR_RuntimeHelper h);
```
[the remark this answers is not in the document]
fjeremic: I fully support @ymanton's review as well. These all sound like excellent improvements!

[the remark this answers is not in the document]
NigelYiboYu: There's another source of confusion I want to mention: the array in `TR_RuntimeHelperTable` contains a mixture of entry points and constant values. How about naming them `getFunctionPointer` and `getEntryPointOrConst`.

1. `getFunctionPointer` returns function pointers for helper functions. It shall return `#define INVALID_FUNC_POINTER 0xdeadbeaf` if the index refers to a constant value. This allows API users (CHelperLinkage on z, for instance) to get function descriptors and extract whatever fields in there.
2. `getFunctionEntryPointOrConst` will replace the current `getAddress()` and return unwrapped entry points or constant numbers. The rest of the codebase assumes it can query raw entry point or constant via the same API.

Also, I want to give `TR_RuntimeHelperTable` a constructor to initialize the whole linkage array to TR_None. This class lacks initialization.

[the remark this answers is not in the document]
ymanton: That's fine. I'm a little surprised to hear that constants are being kept in there, but if that's the case then yes that makes sense.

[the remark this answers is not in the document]
fjeremic: I was too. I don't think they belong there and we should eventually weed them out. Until then I agree we should make the API explicit about the kinds of things we can (currently) extract.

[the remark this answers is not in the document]
NigelYiboYu: Grep for `SET_CONST` and you'll see a dozen of helpers are just constants.

NigelYiboYu:
Addressed all previous concerns and re-tested. This PR is ready again.

ymanton:
LGTM.

fjeremic:
@genie-omr build all

PR #2796, opened 2018-07-25
Avoid linear search if OMRPORT_VMEM_ALLOC_QUICK option is specified

keithc-ca:
On some Linux systems (e.g. s390x) the initial text segment is loaded at virtual address `0x80000000`. (I suspect, but don't know how to verify, that this is new behavior on zLinux which lead to https://github.com/eclipse/openj9/issues/2388.)

A 2GB heap (either because the system has 8GB of RAM or the user explicitly requested that size) cannot be allocated below 4GB (to avoid the need for a non-zero shift for compressed references).

The old (documented) behavior of the `OMRPORT_VMEM_ALLOC_QUICK` option is to fall back to a linear search, which in our example will make 512k attempts to allocate blocks at multiples of 4k - all of which will fail. This only serves to delay returning to `MM_MemoryManager::createVirtualMemoryForHeap` where it can retry with larger shifts for compressed references.

### DIFF
--- a/fvtest/porttest/omrmemTest.cpp
+++ b/fvtest/porttest/omrmemTest.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2017 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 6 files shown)


keithc-ca:
Updated to avoid linear search only if `findAvailableMemoryBlockNoMalloc` returns NULL.

pshipton:
fyi an OpenJ9 Pull Request build with this change took about 1.5 hours. Reduced from about 7 hours for running the sanity tests without this change. Details in https://github.com/eclipse/openj9/issues/2329

charliegracie:
@genie-omr build all

charliegracie:
The port library vmem test is failing on linux ppc and the linux 64 bit compressed references is failing to compile for some really weird reason...

charliegracie:
I restarted the Linux x64 compressed references autotools build and it failed to compiled in the exact same way.

keithc-ca:
Change updated:
* search for free blocks now takes into account whether `OMRPORT_VMEM_STRICT_ADDRESS` option was specified (so the behavior, other than performance, should be the same whether `OMRPORT_VMEM_ALLOC_QUICK` was specified or not)
* fixed some tests to free all memory that was allocated directly

charliegracie:
@genie-omr build all

PR #2797, opened 2018-07-25
Initialize CS background GC thread count

amicic:
This is done now in Configuration class, while it used to be initialized
in downstream projects like OpenJ9.

It is done just after initilization of GC thread count for STW
phase. Then Concurrent Scavenger can initialize its own
background count as a function of STW GC thread count.

The background default is set to appox 1/4 of STW GC thread count.

Also, reporting the counts in verbose GC <initialized stanza. It can now
have two more lines (adding for Concurrent Mark, too):

```

  <attribute name="gcthreads" value="24" />
  <attribute name="gcthreads Concurrent Scavenger" value="6" />
  <attribute name="gcthreads Concurrent Mark" value="1" />
  
```  

### DIFF
--- a/gc/base/standard/ConfigurationStandard.cpp
+++ b/gc/base/standard/ConfigurationStandard.cpp
@@ -87,6 +87,26 @@ MM_ConfigurationStandard::initialize(MM_EnvironmentBase* env)
 	return result;
 }
 
+
+void
+MM_ConfigurationStandard::initializeGCThreadCount(MM_EnvironmentBase* env)
+{
+
+	MM_Configuration::initializeGCThreadCount(env);
+
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+	MM_GCExtensionsBase* extensions = env->getExtensions();
+
+	/* If not explicitly set, concurrent phase of CS runs with approx 1/4 the thread count (relative to STW phases thread count */
+	if (!extensions->concurrentScavengerBackgroundThreadsForced) {
+		extensions->concurrentScavengerBackgroundThreads = OMR_MAX(1, (extensions->gcThreadCount + 1) / 4);
+	} else if (extensions->concurrentScavengerBackgroundThreads > extensions->gcThreadCount) {
+		extensions->concurrentScavengerBackgroundThreads = extensions->gcThreadCount;
+	}
+#endif
+}
+
+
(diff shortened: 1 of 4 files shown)


gc/base/standard/ConfigurationStandard.cpp
```diff
@@ -87,6 +87,25 @@ MM_ConfigurationStandard::initialize(MM_EnvironmentBase* env)
 	return result;
 }
 
+
+void
+MM_ConfigurationStandard::initializeGCThreadCount(MM_EnvironmentBase* env)
+{
+	MM_GCExtensionsBase* extensions = env->getExtensions();
```
charliegracie: This line needs to mo e down into the `#if defined`

gc/verbose/VerboseHandlerOutput.cpp
```diff
@@ -240,6 +240,7 @@ MM_VerboseHandlerOutput::handleInitialized(J9HookInterface** hook, uintptr_t eve
 	MM_InitializedEvent* event = (MM_InitializedEvent*)eventData;
 	MM_VerboseWriterChain* writer = _manager->getWriterChain();
 	MM_EnvironmentBase* env = MM_EnvironmentBase::getEnvironment(event->currentThread);
+	MM_GCExtensionsBase* ext = MM_GCExtensionsBase::getExtensions(env->getOmrVM());
```
charliegracie: Do you need a new variable?  Isn't _extensions a GCExtensions?

gc/base/standard/ConfigurationStandard.cpp
amicic: fixed

gc/verbose/VerboseHandlerOutput.cpp
amicic: fixed

charliegracie:
@genie-omr build all

amicic:
thinking about removing
```
  <attribute name="concurrentScavenger" value="true" />
```

It used be useful to know if CS is capably of running (h/w and OS meet the requirements) and if actually running. But now it's implicitly clear by CS thread count if it is enabled or not (the thread count line will show up only if CS is enabled). 

The line is not documented.

@charliegracie what do you think?

PR #2802, opened 2018-07-26
Reduce output from hookgen and tracegen

youngar:
Reduce the output of both of these tools to none, unless something goes
wrong.  Add a verbose output to give the old output.

### DIFF
--- a/tools/hookgen/HookGen.cpp
+++ b/tools/hookgen/HookGen.cpp
@@ -521,19 +521,32 @@ HookGen::processFile()
 void
 HookGen::displayUsage()
 {
-	fprintf(stderr, "hookgen {fileName}\n");
+	fprintf(stderr, "hookgen [-v] fileName\n");
 }
 
 RCType
 HookGen::parseOptions(int argc, char *argv[])
 {
-	if (argc != 2) {
+	for (int i = 1; i < argc; i++) {
+		printf("%d: %s", i, argv[i]);
+		if(0 == strcmp(argv[i], "-v")) {
+			_verbose = true;
+		} else {
+			if (NULL != _fileName) {
+				fprintf(stderr, "Please specify only one filename\n");
+				displayUsage();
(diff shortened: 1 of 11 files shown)


charliegracie:
@genie-omr build all

tools/hookgen/HookGen.cpp
```diff
@@ -564,7 +577,9 @@ startHookGen(int argc, char *argv[])
 		goto finish;
 	}
 
-	fprintf(stderr, "Processed %s to create public header %s and private header %s\n", hookGen.getFileName(), hookGen.getPublicFileName(), hookGen.getPrivateFileName());
+	if(true == hookGen.isVerbose()) {
```
charliegracie: can you remove the `true == `?  since it is a bool value you do not need it.

tools/tracegen/CFileWriter.cpp
```diff
@@ -50,7 +50,9 @@ CFileWriter::writeOutputFiles(J9TDFOptions *options, J9TDFFile *tdf, J9TDFGroup
 		return RC_OK;
 	}
 
-	printf("Creating c file: %s\n", fileName);
+	if (true == options->verboseOutput) {
```
charliegracie: same as above

tools/tracegen/DATFileWriter.cpp
```diff
@@ -54,7 +54,9 @@ DATFileWriter::writeOutputFiles(J9TDFOptions *options, J9TDFFile *tdf)
 		return RC_OK;
 	}
 
-	printf("Creating pdat file: %s\n", fileName);
+	if (true == options->verboseOutput) {
```
charliegracie: same...

youngar:
@charliegracie I'll go through and change it, I hated it the way it was :)

youngar:
@genie-omr build all

PR #2803, opened 2018-07-27
Deprecate Access Register (AR) support in Z codegen

fjeremic:
The compiler currently has AR support on some minor spectrum of
usability, though in practice the feature has been turned off for
several years, and has never been used outside of niche areas.

Supporting ARs even when disabled (as is the default) does have
technical debt and compile time overhead as we have to query for them
all the time, and in fact we have a separate register allocation pass
just for ARs.

Given these registers were originally envisioned to be spill slots
however times have changed since their original inception:

1. We don't have as many spills due to codegen and RA improvements
2. Storing out the stack is quite cheap
3. There is a lot of weirdness of using ARs around calls (linkage)
4. No support on Linux on Z

With these in mind this commit purges the codebase of all (modulo any
places I may have missed) AR support.


fjeremic:
@0xdaryl could I please get a review? This is blocking additional work in the area of register allocation on Z. I'll resolve the merge conflict momentarily.

fjeremic:
Merge conflicts resolved.

0xdaryl:
Three more files to fix copyright:
```
z/codegen/OMRRealRegister.cpp
z/codegen/OMRRealRegister.hpp
codegen/RegisterConstants.hpp
```

fjeremic:
@0xdaryl fixed all 13 copyright mistakes.

0xdaryl:
@genie-omr build all

One more build since the record of the previous run was lost.

0xdaryl:
The last change was only comment updates.  I don't think this PR needs to be held up any longer waiting for the AppVeyor build as this code should have been covered sufficiently by other builds.

PR #2804, opened 2018-07-28
Fix incorrect hasMandatoryPrefix() on X86

0dvictor:
hasMandatoryPrefix() should answer true when there is one prefix;
however, its answer was reversed. Fixing it.


0xdaryl:
@genie-omr build win

PR #2805, opened 2018-07-28
WIP: Allow transformation of branch instruction on X86

0dvictor:
Three API are introduced to ease transformation of branch instructions:
```
    getConditionCode()      : Obtain the condition test field, a.k.a. tttn field
    changeBranchCondition() : Change the branch condition
    reverseBranch()         : Change the branch to its opposite
```
One API have been updated to eliminate the position dependent hack:
```
    convert4ByteBranchTo1Byte()
```


charliegracie:
ping @0xdaryl

charliegracie:
@genie-omr build xlinux win

mstoodle:
This one needs a review from @0xdaryl who is away for another week, so we'll have to sit on it for the time being.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -167,6 +166,36 @@ inline TR_X86OpCodes SizeParameterizedOpCode(bool is64Bit =
 #define MOVSMemReg     SizeParameterizedOpCode<MOVSDMemReg     , MOVSSMemReg     >
 #define MOVSRegMem     SizeParameterizedOpCode<MOVSDRegMem     , MOVSSRegMem     >
 
+// Aliases to Jcc
```
andrewcraik: Hmm isn't defining these a bit dangerous since they are global replacements of one string with another? It might make finding uses of the opcode in future harder...

0dvictor: Good point, updated.

compiler/x/codegen/X86Ops.hpp
```diff
@@ -33,10 +33,38 @@ enum TR_X86OpCodes : uint32_t
 #define INSTRUCTION(name, mnemonic, binary, property0, property1) name
 #include "codegen/X86Ops.ins"
 #undef INSTRUCTION
+   // Aliases to Jcc
```
andrewcraik: I'm still not sure if this is a great idea - there is going to be a set of names that can appear in the code that won't match what the RAS will print right? That being the case having a canonical name makes more sense? I know the opcode set allows both, but just for simplicity of debugging and understanding not having to remember the conversions would seem to have an attraction to me... Would be interested to hear the thoughts of others on this point. If we do want to do the aliasing then this is the way to do it rather than the previous macro implementation.

0dvictor: We already need to remember the conversions now, as different alias makes more sense in different context. For example: `JB` is the same as `JC`
`STC`
`JC L_SOMEPLACE`
In this context, `JC` makes more sense than `JB`
On the other hand, in following code:
`CMP RAX, RCX`
`JB L_SOMEPLACE`
`JB` makes more sense than `JC`.
Currently, we only allow one form, so that we need to remember the conversion.
In addition, in the debugger and coredumps, the disassembled code only shows one of the alias and it is out of our control. The reader must remember the conversion.

However, I see your point about RAS prints and I totally agree the drawbacks. What about printing all alias of the `Jcc` in its comment? For instance:
`je	Snippet Label L0018		; JE4	; (Check Failure Snippet)`
will become
`je	Snippet Label L0018		; JE4/JZ4	; (Check Failure Snippet)`

andrewcraik: That is a good counter argument. I think having the RAS print the alternative representations might be a good compromise between having the aliases and making the code readable. I was wondering if there was a way to print the one written in the code inline, but because they are aliases I don't think that that is possible. That is to say I was wondering if in your `STC JC` case we could print `jc` in the RAS trace and in the `CMP RAX, RCX, JB` case we could print `jb` in the actual assembly with the comment. As I say I don't see a way with a true alias, but thought I would throw the idea out there. If printing what was written in the codegen isn't possible (as I suspect) then the RAS change you propose will address my concern and I think we can go ahead with the aliasing.

0dvictor: I'm marking this as WIP and I'll update the RAS prints to include all alternative representations.

0dvictor:
Sorry that I haven't got a chance working on the updates related to RAS yet. Will update as soon as possible.

0xdaryl:
This PR hasn't seen an update in a while.  @0dvictor, are you still planning on making the requested RAS changes in the near future?  If not, perhaps we can close this PR while you work on it in your private fork and then re-open when the changes are ready for review again.

0xdaryl:
Closing, per the previous comment.

PR #2806, opened 2018-07-28
Deprecate VMifInstanceOf helper on X86

0dvictor:
Requested by OpenJ9, VMifInstanceOf helper is deprecated. Their old
implementation contains potential functional issues.

Note that VMifInstanceOf is a J9 project specific feature although
it is not yet guarded by J9_PROJECT_SPECIFIC. TR::instanceOf itself
is also a OpenJ9 project specific IL Op.

### DIFF
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -1310,6 +1310,9 @@ TR::Register *OMR::X86::TreeEvaluator::iternaryEvaluator(TR::Node *node, TR::Cod
 
 static bool canBeHandledByIfInstanceOfHelper(TR::Node *node, TR::CodeGenerator *cg)
    {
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build win

0xdaryl:
This change by itself is fine.  Ideally, these evaluators would query a project-neutral API to ask whether or not their handling should be delegated to some project.  This instanceof code as well as the BigDecimal code should move under the OpenJ9-specific implementation of that API.

0xdaryl:
Actually, if the original code contains functional issues, what's the value in keeping this code in OMR at all?  If you can't delete it for some reason, what's the timeframe for it being in "deprecated" state before being ultimately removed?

cathyzhyi:
This deprecated version in omr will be deleted once the new version is believed to be stable enough. LGTM.

0dvictor:
This code will be deleted as soon as OpenJ9 removes its old implementation. I personally don’t think it will take longer than a week.
By keeping the code disabled for a short  period of time, I believe it is helpful ease the debugging in case if the new implementation is not stable.

PR #2807, opened 2018-07-28
Allow accessing TR_OutlinedInstructions on X86

0dvictor:
An API is introduced to allow accessing underlying TR_OutlinedInstructions
from a TR_OutlinedInstructionsGenerator.


0xdaryl:
@genie-omr build win

0dvictor:
In certain cases, user may want to modify/query some properties of underlying TR_OutlinedInstructions; although majority of cases do not. For example, user may want to set the Outlined Path Register Usage List and Mainline Path Register Usage List.

Issue #2755, opened 2018-07-13
CodeGenerator functions should acquire the CodeGenerator object from self()


0xdaryl:
There are several places in the CodeGenerator (common codegen and individual architectures) where the CodeGenerator object is retrieved indirectly within CodeGenerator member functions rather than simply from the immediately available `self()` function.

Here is just one example:

https://github.com/eclipse/omr/blob/fcfd166b487d3796b525be3def5afe8b97c2c022/compiler/codegen/OMRCodeGenerator.cpp#L1154

In this case, `self()` is sufficient rather than `self()->comp()->cg()` as they should return the same pointer.

All CodeGenerator member functions in common code (including files such as `CodeGenRA.cpp`) and each architecture should be scrutinized for this pattern and changed to use `self()`.  Just searching for `cg` or `cg()` in the CodeGenerator files is a good way to find many of these occurrences.

PR #2809, opened 2018-07-29
Use immediately available self() in CodeGenerator

samolisov:
There are several places in the CodeGenerator (common codegen and
individual architectures) where the CodeGenerator object is retrieved
indirectly within CodeGenerator member functions rather than simply from
the immediately available `self()` function. In this case, `self()` is
sufficient rather than `self()->comp()->cg()` as they should return the
same pointer.

Closes: #2755


0xdaryl:
@genie-omr build all

PR #2813, opened 2018-07-31
Revert part of fix incorrect usage of TR::Options

harryyu1994:
- Backing out this commit because of 2% performance regression on daytrader

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -326,7 +326,7 @@ OMR::Compilation::Compilation(
    // while the PersistentMethodInfo is allocated during codegen creation
    // Access to this list must be performed with assumptionTableMutex in hand
    //
-   if (!options.getOption(TR_DisableFastAssumptionReclamation))
+   if (!TR::Options::getCmdLineOptions()->getOption(TR_DisableFastAssumptionReclamation))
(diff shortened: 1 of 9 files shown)


charliegracie:
@genie-omr build all

charliegracie:
Can you provide more details on how this change caused a performance regression?  Was it getting the options wrong? The change should effectively be the same so more details would be appreciated.  Thanks

harryyu1994:
The original change(Fix incorrect usage of TR::Options) exposed a bug somewhere that will be investigated later. I haven't located which line of change it is that changed the program behavior, so I'm reverting to get rid of the regression for now. Somewhere in the program TR::Options::getCmdLineOptions()->getOptions(TR_XXX) and options->getOptions(TR_XXX) are giving different results and DayTrader didn't like the result from options->getOptions(TR_XXX). @charliegracie

harryyu1994:
FYI @vijaysun-omr, Hi Vijay, since you merged the original commit, could you review&merge this one too?

harryyu1994:
@mpirvu FYI; Hi Marius, due to the urgent situation, I'm going to revert this change for now and investigate further later. Could you help reviewing this as well? thanks!

vijaysun-omr:
Looks ok to me. Merging.

pshipton:
fyi original change was #2715

Issue #2799, opened 2018-07-26
Compiler: ELF output is unnecissarily large.


dnakamura:
I got the compiler to generate a object file which defines a single, 7-byte function.  However, the resulting object file is 17MB. 

My best guess is the entirety of the code cache is put into the object file, rather than  just the section which is actually being used

nbhuiyan:
@dnakamura Would you be able to briefly talk about how you were planning on using the generated obj file? You are right about the fact that the entire the code cache gets put into the object file.

dnakamura:
@nbhuiyan The use case I am working on is prototyping out some jitbuilder-as-a-service stuff. 
eg.

Client        Server
JB IL   ->  
         <-  compiled code

The issue is that the address the code executes from on the client, is going to be different from the address of the code cache on the server. Long term there is a better solution to this problem, but in the name of a quick prototype, my solution is to have the server output an ELF object, which can then be loaded into the client.

I have put together a test fix. The generated objects are still much larger than they really need to be, but are still an order of magnitude smaller
https://github.com/dnakamura/omr/tree/elf_fix

mstoodle:
I agree that it's a pain.

Note that copying only the used part only works when trampolines aren't used, because trampolines are allocated from the other end of the code cache and it's not easy to relocate them. POWER and Z rely on trampolines. On top of that, multiple code caches can be allocated within the code cache repository (i.e. the 16MB slab). The initial implementation just writes out the entire repository to avoid dealing with these complications.

PR #2816, opened 2018-07-31
Compiler: Improve ELF output size

dnakamura:
Only output the part of the code cache which is actually in use, rather
that the entire code cache.

### DIFF
--- a/compiler/codegen/ELFGenerator.cpp
+++ b/compiler/codegen/ELFGenerator.cpp
@@ -32,15 +32,15 @@
 #include "runtime/CodeCacheManager.hpp"
 
 TR::ELFExecutableGenerator::ELFExecutableGenerator(TR::RawAllocator rawAllocator,
-                            uint8_t const * codeStart, size_t codeSize):
-                            ELFGenerator(rawAllocator, codeStart, codeSize)
+                            TR::CodeCacheMemorySegment *codeCache):
+                            ELFGenerator(rawAllocator, codeCache)
                             {
                                 initialize();
                             }
 
 TR::ELFRelocatableGenerator::ELFRelocatableGenerator(TR::RawAllocator rawAllocator,
-                            uint8_t const * codeStart, size_t codeSize):
-                            ELFGenerator(rawAllocator, codeStart, codeSize)
+                            TR::CodeCacheMemorySegment *codeCache):
+                            ELFGenerator(rawAllocator, codeCache)
                             {
                                 initialize();
                             }
@@ -295,7 +295,7 @@ TR::ELFGenerator::writeSectionNameToFile(::FILE *fp, char * name, uint32_t size)
 void 
 TR::ELFGenerator::writeCodeSegmentToFile(::FILE *fp)
 {
-    fwrite(static_cast<const void *>(_codeStart), sizeof(uint8_t), _codeSize, fp);
+    fwrite(_codeCache->segmentBase(), sizeof(uint8_t),_codeCache->segmentAlloc() - _codeCache->segmentBase(), fp);
 }
 
 void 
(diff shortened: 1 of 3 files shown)


dnakamura:
See #2799

0xdaryl:
@nbhuiyan , please review.

compiler/codegen/ELFGenerator.cpp
```diff
@@ -419,10 +419,10 @@ TR::ELFExecutableGenerator::initializePHdr(void)
...
     _programHeader->p_type = PT_LOAD; //should be loaded in memory
     _programHeader->p_offset = sizeof(ELFEHeader); //offset of program header from the first byte of file to be loaded
-    _programHeader->p_vaddr = (ELFAddress) _codeStart; //virtual address to load into
-    _programHeader->p_paddr = (ELFAddress) _codeStart; //physical address to load into
-    _programHeader->p_filesz = _codeSize; //in-file size
-    _programHeader->p_memsz = _codeSize; //in-memory size
+    _programHeader->p_vaddr = (ELFAddress) _codeCache->segmentBase(); //virtual address to load into
+    _programHeader->p_paddr = (ELFAddress) _codeCache->segmentBase();; //physical address to load into
+    _programHeader->p_filesz = 0; //this will be set at output time when we know the sioze of the code cache
+    _programHeader->p_memsz = 0;  //this will be set at output time when we know the sioze of the code cache
```
nbhuiyan: typo: size

compiler/codegen/ELFGenerator.cpp
```diff
@@ -419,10 +419,10 @@ TR::ELFExecutableGenerator::initializePHdr(void)
 {
     _programHeader->p_type = PT_LOAD; //should be loaded in memory
     _programHeader->p_offset = sizeof(ELFEHeader); //offset of program header from the first byte of file to be loaded
-    _programHeader->p_vaddr = (ELFAddress) _codeStart; //virtual address to load into
-    _programHeader->p_paddr = (ELFAddress) _codeStart; //physical address to load into
-    _programHeader->p_filesz = _codeSize; //in-file size
-    _programHeader->p_memsz = _codeSize; //in-memory size
+    _programHeader->p_vaddr = (ELFAddress) _codeCache->segmentBase(); //virtual address to load into
+    _programHeader->p_paddr = (ELFAddress) _codeCache->segmentBase();; //physical address to load into
+    _programHeader->p_filesz = 0; //this will be set at output time when we know the sioze of the code cache
```
nbhuiyan: typo: size

charliegracie:
ping @dnakamura

dnakamura:
updated

mstoodle:
Can you please explain how you tested this change @dnakamura ?  For example, there is a UseCall code sample that isn't currently run by default in our testing. Have you made sure that continues to work?

I am still worried that this does not work on any platform but x86 where we do not use trampolines and will be broken even there if we start using the helper infrastructure (see diagram here: https://github.com/eclipse/omr/blob/master/compiler/runtime/OMRCodeCache.cpp#L57-L74 ).

dnakamura:
@mstoodle  I ran the a couple of the jitbuilder samples (mandelbrot, recfib, iterfib), with the appropriate `TR_Options` to get it to output the obj file. For each one I wrote a small C test harness (basically a small stub just to invoke the function and dump output), and linked against the appropriate object files. Executing programs resulted in expected behavior.

However given your earlier feedback on trampolines. I'm thinking that I should re-test on some power/z hardware

mstoodle:
@dnakamura i would appreciate that. It may just mean an adjustment to calculate the upper bound of the cache that depends on the platform (on x86 we can get away with the `alloc` pointer but by default maybe we should store the whole cache).

PR #2819, opened 2018-07-31
Consolidate data size check helpers on X86

0dvictor:
Move data size check helpers into shared header file: env/jittypes.h

### DIFF
--- a/compiler/codegen/OMRTreeEvaluator.hpp
+++ b/compiler/codegen/OMRTreeEvaluator.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


0xdaryl:
I wonder if such macros would be generally useful outside of the x86 codegen (there's nothing x86-specific about them).  Perhaps relocating them to `env/jittypes.h` would be a better place?

0dvictor:
> I wonder if such macros would be generally useful outside of the x86 codegen (there's nothing x86-specific about them). Perhaps relocating them to `env/jittypes.h` would be a better place?

Good suggestion, I've moved them into `env/jittypes.h`. Note that I did not move `IS_32BIT_RIP` as it is a X86-specific query.

0xdaryl:
@genie-omr build all

0xdaryl:
Please update the copyright header in jittypes.h

0dvictor:
> Please update the copyright header in jittypes.h

Updated.

0xdaryl:
@genie-omr build all

Sorry, I have to restart these as they were lost once the new commit was pushed and I didn't have a chance to see what their status was after the first set of builds.

PR #2820, opened 2018-07-31
WIP: Create VectorLoopBuilder and IlBuilder::VectorForLoop

mstoodle:
To make it easier to build vectorized loops, this pull request creates a new class called a VectorLoopBuilder which takes care of the mundane aspects of managing a vectorized loop along with a residue loop. Basically VectorLoopBuilder manages two internal builder objects, one each corresponding to the vector loop body and the residue loop body. Services called on the VectorLoopBuilder are passed along to *both* the vector and residue loop copy, enabling users to create fully correct vectorized loops while only writing one version of the loop code.

Also provided is a new service on IlBuilder called VectorForLoop() which constructs a VectorLoopBuilder object and connects its internal builder objects into the control flow structure for a vectorized loop. It takes a type which will determine the vector length used in the loop as well as the `initial` and `end` values that describe the loop iterations. It returns a `TR::VectorLoopBuilder` object which can then be used to fill in the loop. If the provided type cannot be vectorized, then only the residue loop will be generated.

Along with this new service, there are two new code samples: Loop which multiplies two 1D vectors together, and VectorMatMult which vectorizes the middle loop of a straightforward ijk matrix multiplication loop nest.

This service is just a start; it's not complete in that only Add, Sub, and Mul, various flavours of Loads/Stores, and ForLoop are implemented on VectorLoopBuilder at the moment, and the vectorization support depends on what the OMR compiler component can do. It will not work on all platforms but it runs just fine on my MacBook and should be good on the x86 platform, at least.

### DIFF
--- a/compiler/ilgen/CMakeLists.txt
+++ b/compiler/ilgen/CMakeLists.txt
@@ -29,6 +29,7 @@ compiler_library(ilgen
 	${CMAKE_CURRENT_LIST_DIR}/OMRMethodBuilder.cpp
 	${CMAKE_CURRENT_LIST_DIR}/OMRThunkBuilder.cpp
 	${CMAKE_CURRENT_LIST_DIR}/OMRTypeDictionary.cpp
+	${CMAKE_CURRENT_LIST_DIR}/OMRVectorLoopBuilder.cpp
 	${CMAKE_CURRENT_LIST_DIR}/OMRVirtualMachineOperandArray.cpp
 	${CMAKE_CURRENT_LIST_DIR}/OMRVirtualMachineOperandStack.cpp
 	${CMAKE_CURRENT_LIST_DIR}/OMRVirtualMachineState.cpp
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -45,6 +45,7 @@
 #include "ilgen/IlInjector.hpp"
 #include "ilgen/MethodBuilder.hpp"
 #include "ilgen/BytecodeBuilder.hpp"
+#include "ilgen/VectorLoopBuilder.hpp"
 #include "infra/Cfg.hpp"
 #include "infra/List.hpp"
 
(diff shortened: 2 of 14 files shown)


mstoodle:
I just changed the "Loop" sample to use 32-bit integers. I had thought to use 8 bit integers, but apparently the evaluators aren't all there for the VectorInt8 type. For now, I think 32-bit integers are good enough.

mstoodle:
So, this approach obviously works, so long as you know exactly what you're doing. But it may not be a very practical final solution. In the end, clients can't really know if an operation is supported as a vector instruction, and so I think this VectorLoopBuilder thing may not be easy to use. I have been pondering a slightly different approach where the actual loop control flow does not get generated until `buildIL()` has been called on the loop builder. It would start by leaving builders in place where the vector loop and control flow *could* be added, but it won't actually be added until countBlock() comes along to find out what's in that build. At which point, it will decide whether it should use the vector loop or not. That way, VectorLoopBuilder can track whether an operation was inserted by the client that did not fit into the "vector" scheme and so it can either leave just a residue loop in place or tack in the (already constructed) vector loop body and control flow.

I'm marking this PR as WIP while I noodle that a bit more. Suggestions and feedback welcome!

mstoodle:
Finally managed to spend a bit more time to think about this one.

What I am thinking is that the VectorForLoop service will inject something like the following into the builder `B` it is called on before it returns:

```
Store("vectorIterationsCompleted", initial);    // vector loop will start here or maybe do nothing
AppendBuilder(possibleVectorLoop);             // left empty until *after* client's buildIL has returned
AppendBuilder(possibleMergeBack);
ForLoopUp(<residue loop>,
     Load("vectorIterationsCompleted",
     <normal step>
     end);
`````

Services called on the VectorLoopBuilder will inject a scalar operation into the residue loop and also a vector operation into the vector loop, which has been created but not connected to the original builder `B` yet. If any service is called that does not (yet?) have a supported vector operation, then the VectorForLoop will remember that and basically abandon the vector loop from that point forward and the control flow graph will stay just as it was when the code was created by VectorForLoop: the initial setting of `vectorIterationsCompleted` will simply cause all of the loop iterations to execute in the residue loop as scalar operations (which is the fall-back case).

If, however, no unsupported service is invoked, then "at the end of the `MethodBuilder::buildIL()` function, the `VectorLoopBuilder` object will add this extra call:
```
possibleVectorLoop->Goto(actualVectorLoop);
```
...and the `actualVectorLoop` sequence will end in `Goto(possibleMergeBack);`

In fact, it may not have to be done with builders (it could accomplished with actual IL blocks and a `goto` node), but the end result is the same. To accomplish this bit, I think each `VectorLoopBuilder` object will probably register itself with its `MethodBuilder` for this "post" updating trick.

Questions around how to represent whether vector operations are supported yet, since that's a per-platform question that will obviously need to be maintained as the code generators mature; a code generator query seems like an obvious approach. Another question is whether this support should be done as a new kind of "VectorLoopBuilder" class, which will obviously need to mimic the services in IlBuilder, or become folded into `IlBuilder`, or if we should do these kinds of things via something like the listener pattern that's being looked at for the client API.

Any thoughts?

mstoodle:
Closing until I can spend time re-implementing the approach.

PR #2823, opened 2018-08-04
Enable aliasing of auto variables that have their address taken

dibyendumajumdar:
In languages like C it is possible for local (auto) variables to have their address taken
and then passed by reference to another function. Example:

int foo(int*i, long long*l)
{
	*i = (int) *l;
	return 0;
}

int simplelocals(void)
{
	int a = 5;
	long long b = 42;
	foo(&a, &b);	// Just to fool the optimiser
	return a+(int)b;
}

Currently OMR's localCSE ignores the fact that the two locals were passed by reference to foo() and therefore
could have been changed by foo(). This results in incorrect optimization. The solution is to alias the autos
for "a" and "b" with the call "foo".

To enable this a new BitVector _addressTakenAutos is introduced in OMR::AliasBuilder. This is made available
to front-ends via OMR::AliasBuilder::addressTakenAutos(). The front-end is expected to set the symbols for
which addresses have been taken into this set.

The second part of this change is to use the aliasing information. The aliasing api that controls what the
rest of the optimizer sees is getUseDefAliasesBV in
https://github.com/eclipse/omr/blob/master/compiler/il/Aliases.cpp

In this routine the case for:

case TR::Symbol::IsResolvedMethod

Is updated to OR the addressTakenAliases with the method's aliases.

This change relates to the bug logged in #issue 2768.

Signed-off-by: Dibyendu Majumdar <mobile@majumdar.org.uk>

### DIFF
--- a/compiler/il/Aliases.cpp
+++ b/compiler/il/Aliases.cpp
@@ -580,11 +580,18 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
             return aliases;
             }
 #endif
-         TR_BitVector * aliases = new (aliasRegion) TR_BitVector(bvInitialSize, aliasRegion, growability);
-         *aliases |= symRefTab->aliasBuilder.addressTakenAutos();
-         TR_BitVector *methodAliases = symRefTab->aliasBuilder.methodAliases(self());
-         *aliases |= *methodAliases;
-         return aliases;
+         if (!symRefTab->aliasBuilder.addressTakenAutos().isEmpty())
+            {
+               TR_BitVector * aliases = new (aliasRegion) TR_BitVector(bvInitialSize, aliasRegion, growability);
+               *aliases |= symRefTab->aliasBuilder.addressTakenAutos();
+               TR_BitVector *methodAliases = symRefTab->aliasBuilder.methodAliases(self());
+               *aliases |= *methodAliases;
+               return aliases;
+            }
+         else 
+            {
+               return symRefTab->aliasBuilder.methodAliases(self());
+            }
          }
       case TR::Symbol::IsShadow:
          {
--- a/compiler/il/Aliases.cpp
+++ b/compiler/il/Aliases.cpp
@@ -580,8 +580,11 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
             return aliases;
             }
 #endif
-
-         return symRefTab->aliasBuilder.methodAliases(self());
+         TR_BitVector * aliases = new (aliasRegion) TR_BitVector(bvInitialSize, aliasRegion, growability);
+         *aliases |= symRefTab->aliasBuilder.addressTakenAutos();
+         TR_BitVector *methodAliases = symRefTab->aliasBuilder.methodAliases(self());
+         *aliases |= *methodAliases;
+         return aliases;
          }
       case TR::Symbol::IsShadow:
          {
--- a/compiler/compile/OMRAliasBuilder.cpp
+++ b/compiler/compile/OMRAliasBuilder.cpp
@@ -55,6 +55,7 @@ OMR::AliasBuilder::AliasBuilder(TR::SymbolReferenceTable *symRefTab, size_t size
      _refinedNonIntPrimitiveArrayShadows(1, c->trMemory(), heapAlloc, growable),
      _refinedAddressArrayShadows(1, c->trMemory(), heapAlloc, growable),
      _refinedIntArrayShadows(1, c->trMemory(), heapAlloc, growable),
+     _addressTakenAutos(1, c->trMemory(), heapAlloc, growable),
      _litPoolGenericIntShadowHasBeenCreated(false),
      _userFieldMethodDefAliases(c->trMemory(), _numNonUserFieldClasses),
      _conservativeGenericIntShadowAliasingRequired(false),
--- a/compiler/compile/OMRAliasBuilder.hpp
+++ b/compiler/compile/OMRAliasBuilder.hpp
@@ -108,6 +108,8 @@ public:
    TR_BitVector & refinedAddressArrayShadows() { return _refinedAddressArrayShadows; }
    TR_BitVector & refinedIntArrayShadows() { return _refinedIntArrayShadows; }
 
+   TR_BitVector & addressTakenAutos() { return _addressTakenAutos; }
+
    bool litPoolGenericIntShadowHasBeenCreated(){ return _litPoolGenericIntShadowHasBeenCreated; }
    void setLitPoolGenericIntShadowHasBeenCreated(){ _litPoolGenericIntShadowHasBeenCreated = true; }
 
@@ -184,6 +186,8 @@ protected:
    TR_BitVector _refinedAddressArrayShadows;
    TR_BitVector _refinedIntArrayShadows;
    TR_BitVector _refinedNonIntPrimitiveArrayShadows;
+
+   TR_BitVector _addressTakenAutos;
 
    bool _litPoolGenericIntShadowHasBeenCreated;
 
--- a/compiler/optimizer/UseDefInfo.cpp
+++ b/compiler/optimizer/UseDefInfo.cpp
@@ -143,7 +143,13 @@ void TR_UseDefInfo::prepareUseDefInfo(bool requiresGlobals, bool prefersGlobals,
       }
 
    bool canBuild = false;
-   _hasCallsAsUses = false;
+   // OMR sets this to false, but we need this to be enabled to ensure call nodes
+   // participate in use/def analysis, and thereby allow any aliased locals
+   // to be handled correctly; this is WIP as for Java like languages this is
+   // not needed, as locals cannot be aliased by function calls. However for C
+   // like languages this is needed; therefore this needs to be controlled via
+   // a parameter
+   _hasCallsAsUses = true;  
    _uniqueIndexForDefsOnEntry = false;
 
    if (comp()->cg()->getGRACompleted() && conversionRegsOnly)
--- a/compiler/optimizer/UseDefInfo.cpp
+++ b/compiler/optimizer/UseDefInfo.cpp
@@ -143,13 +143,6 @@ void TR_UseDefInfo::prepareUseDefInfo(bool requiresGlobals, bool prefersGlobals,
       }
 
    bool canBuild = false;
-   // OMR sets this to false, but we need this to be enabled to ensure call nodes
-   // participate in use/def analysis, and thereby allow any aliased locals
-   // to be handled correctly; this is WIP as for Java like languages this is
-   // not needed, as locals cannot be aliased by function calls. However for C
-   // like languages this is needed; therefore this needs to be controlled via
-   // a parameter
-   _hasCallsAsUses = true;  
    _uniqueIndexForDefsOnEntry = false;
 
    if (comp()->cg()->getGRACompleted() && conversionRegsOnly)
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -48,9 +48,9 @@ add_executable(comptest
 	CompareTest.cpp
 	TypeConversionTest.cpp
 	TernaryTest.cpp
-	# Following test crashes at present because
+	# Following test fails at present because
 	# it requires optimizer callsAsUses flags enabled
-	#LocalPassByRefTest.cpp
+	LocalPassByRefTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -48,6 +48,9 @@ add_executable(comptest
 	CompareTest.cpp
 	TypeConversionTest.cpp
 	TernaryTest.cpp
+	# Following test crashes at present because
+	# it requires optimizer callsAsUses flags enabled
+	#LocalPassByRefTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/LocalPassByRefTest.cpp
+++ b/fvtest/compilertriltest/LocalPassByRefTest.cpp
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "JitTest.hpp"
+#include "default_compiler.hpp"
+
+class LocalPassByRefTest : public TRTest::JitTest {};
+
+int foo(int*i, long long*l)
+{
+	*i = (int) *l;
+	return 0;
+}
+
+TEST_F(LocalPassByRefTest, passByRefTest) { 
+    char inputTrees[512] = {0};
+	const auto format_string = "(method return=Int32 (block (istore temp=\"a\" (iconst 5)) (lstore temp=\"b\" (lconst 42)) (treetop (icall address=0x%jX args=[Address, Address] (loadaddr temp=\"a\") (loadaddr temp=\"b\") )) (ireturn (iadd (l2i (lload temp=\"b\")) (iload temp=\"a\")))))";
+    std::snprintf(inputTrees, sizeof inputTrees, format_string, reinterpret_cast<uintmax_t>(&foo));
+    auto trees = parseString(inputTrees);
+
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we 
+    // do not have trampoline support, and so this call may be out of 
+    // range for some architectures. 
+#ifdef TR_TARGET_X86
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+    auto entry_point = compiler.getEntryPoint<int32_t (*)()>();
+
+    EXPECT_EQ(84, entry_point());
+#endif
+}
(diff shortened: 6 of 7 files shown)


fjeremic:
Suggesting @vijaysun-omr and @andrewcraik for review.

dibyendumajumdar:
Hi - note that this initial pull is probably incomplete - it is causing a break in one of my tests. In UseDefInfo.hpp, I am getting an out of bounds error at line 1660: 

```
 aux._sideTableToUseDefMap[node->getLocalIndex()] = node->getUseDefIndex();
```

Any tips on what I am missing?

vijaysun-omr:
Regarding your last question, it may be useful to see what the "node" is in the case when you are hitting the error. If I had to guess, maybe you have a node that is'nt a direct load or a direct store referring to an auto.

compiler/il/Aliases.cpp
```diff
@@ -542,8 +542,11 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
             return aliases;
             }
 #endif
-
-         return symRefTab->aliasBuilder.methodAliases(self());
+         TR_BitVector * aliases = new (aliasRegion) TR_BitVector(bvInitialSize, aliasRegion, growability);
+         *aliases |= symRefTab->aliasBuilder.addressTakenAutos();
+         TR_BitVector *methodAliases = symRefTab->aliasBuilder.methodAliases(self());
```
vijaysun-omr: I wonder if you could check if addressTakenAutos is non empty before create the new "aliases" bit vector in this case. i.e. if it was empty, then we could just return the same answer as in the original code without creating a new bit vector (especially for front ends that will never have bits set in addressTakenAutos).

dibyendumajumdar: Yes makes sense. I will also create a test case before resubmitting.

dibyendumajumdar: BTW I am getting an out of bounds error again if I run with more optimizations ... so there is probably something missing that I need to add.

[trtrace.log](https://github.com/eclipse/omr/files/2280363/trtrace.log)

vijaysun-omr: Which optimization causes your error ? It may be that some optimization passes need some (hopefully small) change to account for the fact that autos could be in the alias sets returned for calls. This kind of aliasing may not have happened before in the front ends that are running the optimizations that you are running.

vijaysun-omr: Okay I see you were doing global copy propagation when the crash happened. Which IL node was that analysis looking at and what was the stack trace for the crash ?

mstoodle: @dibyendumajumdar just a gentle reminder there's an outstanding question for you here...

dibyendumajumdar: Hi - yes, sorry just been busy with other things. I will do some more tests and report back.

vijaysun-omr: @dibyendumajumdar just a friendly ping to check in whether you had some more information on the crash (the earlier question) so that we can try to help. Thanks.

dibyendumajumdar: @vijaysun-omr yes sorry for the delay. I have finally got some time to look at this. Creating test case and then will update this pull request with more info. Thanks

dibyendumajumdar: @vijaysun-omr Stack trace:
```
>	sparse-omrjit.exe!std::deque<unsigned int,TR::typed_allocator<unsigned int,TR::Region &> >::operator[](unsigned __int64 _Pos) Line 1411	C++
 	sparse-omrjit.exe!TR::deque<unsigned int,TR::Region &>::operator[](unsigned __int64 index) Line 108	C++
 	sparse-omrjit.exe!TR_UseDefInfo::insertData(TR::Block * block, TR::Node * node, TR::Node * parent, TR::TreeTop * treeTop, TR_UseDefInfo::AuxiliaryData & aux, CS2::ASparseBitVector<CS2::shared_allocator<CS2::heap_allocator<65536,12,TRMemoryAllocator<1,12,28> > > > & expandedLIndexes, bool considerImplicitStores) Line 1660	C++
 	sparse-omrjit.exe!TR_UseDefInfo::insertData(TR::Block * block, TR::Node * node, TR::Node * parent, TR::TreeTop * treeTop, TR_UseDefInfo::AuxiliaryData & aux, CS2::ASparseBitVector<CS2::shared_allocator<CS2::heap_allocator<65536,12,TRMemoryAllocator<1,12,28> > > > & expandedLIndexes, bool considerImplicitStores) Line 1559	C++
 	sparse-omrjit.exe!TR_UseDefInfo::fillInDataStructures(TR_UseDefInfo::AuxiliaryData & aux) Line 485	C++
 	sparse-omrjit.exe!TR_UseDefInfo::prepareUseDefInfo(bool requiresGlobals, bool prefersGlobals, bool cannotOmitTrivialDefs, bool conversionRegsOnly) Line 323	C++
 	sparse-omrjit.exe!TR_UseDefInfo::TR_UseDefInfo(TR::Compilation * comp, TR::CFG * cfg, TR::Optimizer * optimizer, bool requiresGlobals, bool prefersGlobals, bool loadsShouldBeDefs, bool cannotOmitTrivialDefs, bool conversionRegsOnly, bool doCompletion) Line 103	C++
 	sparse-omrjit.exe!OMR::Optimizer::performOptimization(const OptimizationStrategy * optimization, int firstOptIndex, int lastOptIndex, int doTiming) Line 1850	C++
 	sparse-omrjit.exe!OMR::Optimizer::optimize() Line 1129	C++
 	sparse-omrjit.exe!OMR::Compilation::performOptimizations() Line 1275	C++
 	sparse-omrjit.exe!OMR::Compilation::compile() Line 1058	C++
 	sparse-omrjit.exe!compileMethodFromDetails(OMR_VMThread * omrVMThread, TR::IlGeneratorMethodDetails & details, TR_Hotness hotness, int & rc) Line 373	C++
 	sparse-omrjit.exe!nj::FunctionBuilder::compile(int opt_level) Line 261	C++
 	sparse-omrjit.exe!JIT_Compile(JIT_FunctionBuilder * fb, int opt_level) Line 451	C++
 	sparse-omrjit.exe!output_fn(dmr_C * C, JIT_Context * module, entrypoint * ep) Line 2175	C
 	sparse-omrjit.exe!compile(dmr_C * C, JIT_Context * module, symbol_list * list) Line 2210	C
 	sparse-omrjit.exe!dmrC_omrcompile(int argc, char * * argv, JIT_Context * module, const char * inputbuffer) Line 2249	C
 	sparse-omrjit.exe!main(int argc, char * * argv) Line 90	C
 	[External Code]	
```

dibyendumajumdar: Node is icall:

```

  | Name | Value | Type
-- | -- | -- | --
◢ | node | 0x000001e54b593bb0 {...} | TR::Node *
  | ◢ OMR::Node | {_opCode={...} _numChildren=2 _visitCount=17 ...} | OMR::Node
  | ◢ _opCode | {...} | TR::ILOpCode
  | ▶ OMR::ILOpCode | {_opCode=icall (61) } | OMR::ILOpCode
  | _numChildren | 2 | unsigned short
  | _visitCount | 17 | unsigned short
  | _globalIndex | 17 | unsigned int
  | ▶ _flags | {_flags=0 } | OMR::FlagsTemplate<unsigned int>
  | _localIndex | 4 | unsigned int
  | ▶ _byteCodeInfo | {_doNotProfile=0 _isSameReceiver=0 _callerIndex=-1 ...} | TR_ByteCodeInfo
  | _referenceCount | 1 | unsigned int
  | ▶ _unionA | {_useDefIndex=3 _register=0x0000000000000003 {...} } | OMR::Node::UnionA
  | ▶ _unionBase | {_constValue=2084323277616 _fpConstValue=14236464.0 _dpConstValue=1.029792526297e-311#DEN ...} | OMR::Node::UnionBase
  | ▶ _unionPropertyA | {_symbolReference=0x000001e54b59c810 {...} _regLoadStoreSymbolReference=0x000001e54b59c810 {...} _branchDestinationNode=...} | OMR::Node::UnionPropertyA

```

dibyendumajumdar: aux._sideTableToUseDefMap has size 4, so presumably this is why there is an out of bounds exception.

dibyendumajumdar: So I reckon the bug must be in getExpandedTotalNode()?

 aux._sideTableToUseDefMap.resize(getExpandedTotalNodes());

vijaysun-omr: Thanks, could you please upload a "current" version of the log with traceUseDefs added to your command line options ? It was'nt clear to me if stuff had changed in your build since the last log was generated (?)

vijaysun-omr: Calls "expand" into N bits in the use def info analysis, one for each symbol reference that may be aliased to the call. This can probably be explained most easily with an example:

call foo // aliases t1, t2 and t3 (say)
...
t1 = rhs;
...
load t1 // the def is the above store 
...
load t2 // the def is the above call
...
load t3 // the def is the above call

So we need a mechanism that kills the call as a def for t1 alone when the store to t1 is encountered and this is what the expanded index achieves...by giving one bit per sym ref that may alias the call, there are separate bits for t1, t2, t3 associated with the call. 

Normally the set of symbol references that may be aliased to the call only contains shadows and statics. So, we may need to check how many bits the call is expanding out to and which aliases are being taken into account in that expansion. In particular are the new address taken autos included in the set of symbol reference that may be aliased when deciding how many bits the call's aliasing should expand to.

dibyendumajumdar: I have saved it here:
https://github.com/dibyendumajumdar/dmr_c/blob/master/tests/omrjit/traces/trtrace_traceUseDefs.log

I will add another one with full all other options added

dibyendumajumdar: [trtrace.log](https://github.com/eclipse/omr/files/2441191/trtrace.log)

This is trace with TR_Options=traceUseDefs,log=trtrace.log,traceIlGen,traceFull,traceAliases

vijaysun-omr: Thanks, are you constructing usedef info passing true to requiresGlobals and prefersGlobals ?
i.e. in
TR_UseDefInfo::TR_UseDefInfo(TR::Compilation *comp, TR::CFG *cfg, TR::Optimizer *optimizer,
bool requiresGlobals, bool prefersGlobals, bool loadsShouldBeDefs, bool cannotOmitTrivialDefs, bool conversionRegsOnly, bool doCompletion)

I think we may not even be indexing calls as participants in usedef info analysis until those flags are set. I'm still not certain it will all work with just those flags set but I'd be interested in seeing a fresh log and info about any crash that you run into.

dibyendumajumdar: Hi @vijaysun-omr , it seems that requiresGlobals was set to true, and prefersGlobals set to false. So I changed latter to true too, and attached is the trace log after that.

[trtrace.log](https://github.com/eclipse/omr/files/2454183/trtrace.log)

I would like to understand the aliasing code and fix it for this scenario, but it isn't very clear how to get started. I don't tend to have a lot of time to delve into it so that doesn't help. If you could point me to bit where calls are analyzed then maybe I can start there.

vijaysun-omr: @dibyendumajumdar I understand and I will try to help as much as possible to allow you to move forward. The new trace log does not really show the call node as participating in the use def info analysis (i.e. it did not change anything). 

The aliasing info shown in your log file may not necessarily be the problem; from my quick scan it looked fine to me. The issue may be the usage of aliasing info in the specific analysis you are crashing in, namely use def info analysis. The call node that you have in your test case has to participate in the use def info analysis info for it's aliasing info to be even consulted. 

The routine inside use def info analysis that I feel is'nt picking up call nodes is:

bool TR_UseDefInfo::indexSymbolsAndNodes(AuxiliaryData &aux)

This routine is responsible for figuring out which nodes need to participate in the analysis and which index to assign each node and symbol. Not assigning a node an index means that it does not participate in the analysis in any way, i.e. it is considered neither a use nor a definition of anything. 

As an example from the Java usage in OpenJ9, one would expect load and store nodes for auto symbols to participate in the analysis but given that we don't run use def info analysis for globals (i.e. shadows and statics) in OpenJ9, no other kinds of symbols participate in the analysis. The reason why we don't do use def analysis for "globals", i.e. statics and shadows in OpenJ9 is simply a choice that we have made given that aliasing tends to be quite conservative for calls (that can more or less kill any global) and so we don't really anticipate the info to lead to a significant enough throughput performance gain to be worth the compile time cost.

For other OMR-based language compilers that want to track uses and definitions of "global" symbols as well, i.e. statics and shadows, the first step would be to set those two flags I mentioned in the use def info as you are doing already it seems. In that case I would have expected you to index your call node around the following line in UseDefInfo.cpp:

      if (sym->isMethod())
               {
               if (!_hasCallsAsUses)
                  continue;
               }
            else if (!isValidAutoOrParm(symRef))
               continue;

            // Index this symbol
            //
            if (trace())
               {
                  traceMsg(comp(), "Non Trivial symbol [%p] has index %d\n", sym, _numSymbols);
               }
            _sideTableToSymRefNumMap[_numSymbols] = symRefNumber;
            sym->setLocalIndex(_numSymbols++);
_numNonTrivialSymbols++;

We have had other compilers use the "globals" implementation of use def info in the past but because it's not used in OpenJ9, I'll probably have to do some of the discovery of why this mode of running use def info is'nt working for you, i.e. it's not a path I'm as familiar with. That having been said, I am pretty sure we can work together to find the root cause and it probably won't be that hard to fix.

dibyendumajumdar: Hi @vijaysun-omr, thank you for the detailed answer. I will spend some time on this over the coming weekend (hopefully!) and report back on progress.

vijaysun-omr: Also, the following code is where the call node ought to be indexed. So you can check what we did if/when we reached here for your call now.

 else if (opCode.isCall() || opCode.isFence())
         {
         if (num_aliases > 0)
            {
            if (symRefToLocalIndexMap[symRef->getReferenceNumber()] == 0)
               {
               localIndex = _numExpandedDefUseNodes;
               _numExpandedDefUseNodes += num_aliases;
               //traceMsg(comp(), "num_aliases #%d  %d %d %d\n", node->getSymbolReference()->getReferenceNumber(), num_aliases, opCode.isFence(), node->getSymbolReference()->getSymbol()->isMethod());
               useDefIndex = _numDefUseNodes++;
               //             traceMsg(comp(), "UDI: setting useDefIndex to %d _numDefUseNodes = %d\n",useDefIndex,_numDefUseNodes);

               symRefToLocalIndexMap[symRef->getReferenceNumber()] = localIndex;
               }
            else
               {
               localIndex = symRefToLocalIndexMap[symRef->getReferenceNumber()];
               useDefIndex = _numDefUseNodes++;
               }
}

dibyendumajumdar: > In that case I would have expected you to index your call symbol around the following line in UseDefInfo.cpp:
> 
> ```
>   if (sym->isMethod())
>            {
>            if (!_hasCallsAsUses)
>               continue;
>            }
>         else if (!isValidAutoOrParm(symRef))
>            continue;
>

It seems that _hasCallsAsUses is never set anywhere - perhaps some code related to this was removed?
Anyway, so the reason nothing happens here is because of that.
 
>         // Index this symbol
>         //
>         if (trace())
>            {
>               traceMsg(comp(), "Non Trivial symbol [%p] has index %d\n", sym, _numSymbols);
>            }
>         _sideTableToSymRefNumMap[_numSymbols] = symRefNumber;
>         sym->setLocalIndex(_numSymbols++);
> ```
> 
> So the next step would be to understand if/why the call symbol was not indexed at this point since this probably holds you back from indexing the call node later from:

dibyendumajumdar: So if I set _hasCallsAsUses to true then it doesn't crash. The test passes so presumably nothing bad happens. So what should be setting this to true? Presumably I need a way to test whether the Call node aliased locals and set this to true in that case?

dibyendumajumdar: @vijaysun-omr @mstoodle @Leonardo2718  Apart from fixing the issues with optimization, I also need to provide an API to set the address taken flag on a local symbol, so that the alias info is updated. Any views on what would be a preferred interface? In my project I provide an api such as:

```
/**
 * Inform the compiler that the automatic variable had its 
 * address taken, hence the variable should be considered as aliased
 * when calling functions etc.
 */
extern void JIT_SetAutoAddressTaken(JIT_ILInjectorRef ilinjector,
	JIT_SymbolRef symbol);
```

I also need a mechanism in Tril to set this for the test case. 

Thanks and Regards

vijaysun-omr: @dibyendumajumdar can you please upload a log from the run where you set _hasCallAsUses to true and the test passed ? With all the same tracing options as you did previously.

dibyendumajumdar: Hi @vijaysun-omr  

As requested: 
[trtrace.log](https://github.com/eclipse/omr/files/2480526/trtrace.log)

vijaysun-omr: Looks like the aliasing did get through to use def info as was intended. As I mentioned before, on OpenJ9 we probably did not need to set that boolean to true since locals cannot be aliased with calls in Java. This boolean should really be set in a language-specific way somehow so as to avoid hardcoding it to either true or false (not needed for Java for compile time reasons and not functionally correct for your language implementation).

andrewcraik: so to answer the first question setting _hasCallAsUses should be handled as part of the constructor if it needs to be configured by different languages. This is the first time we are going to have a language specific override of a flag like this for UseDefInfo. Looking at how it is currently used instances of TR_UseDef_info are allocated in the OMROptimizer.cpp, ValueNumberInfo.cpp and LoadExtensions.cpp. There is some variation between them in how they configure UseDefInfo for globals. I think the best way to handle this is to create a new API on the Optimizer object (eg OMROptimizer.hpp and OMROptimizer.cpp) which all current allocation sites switch to calling - the method will accept as parameters the bools which vary between the different request sites within the compiler. The value for _hasCallAsUses could be either stored as a flag in the optimizer or by overriding the build method to change how UseDefInfo is built. I would prefer the later since it means languages have a common place to configure UseDefInfo according to their needs. This small refactor would ideally be done as a separate commit and pull request ahead of changes introducing the new flag to the constructor. I hope that makes sense - if it doesn't do keep asking questions and I'll try to clarify more on how this would work.

An added bonus of this is that we could improve the UseDefInfo memory management in a follow-on item which would be a another good reason to refactor the optimizer to allocate UseDefInfo through a central location.

dibyendumajumdar: Hi @andrewcraik 

Thanks for the feedback. I will open a new pull request for the suggested refactoring.

Regards

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@dibyendumajumdar are you planning any other changes in this PR (apart from maybe squashing the commits) ?

dibyendumajumdar:
Hi @vijaysun-omr, I am planning to add a way to set the _hasCallAsUses in a language specific way. But this is dependent upon the refactoring suggested by @andrewcraik which I haven't done yet.

Additionally the test case I added needs to be enhanced to set both the _hasCallAsUses flag as well as inform the compiler that the locals have had their address taken; I am not yet sure how to do that in a tril testcase.

dibyendumajumdar:
Hi all, I still haven't had a time window to work on this pull request, as it needs a few days of effort to a) add an api to control the _hasCallsAsUses flag and also to add some mechanism  in Tril to inform the optimizer about the address being taken so that the test case can be handled correctly. It doesn't look like I will get time until later this year, but this is critical change to support languages such as C so I intend to finish it at some point. Sorry for the delay.

compiler/il/Aliases.cpp
dibyendumajumdar: Hi @andrewcraik @vijaysun-omr , I have submitted https://github.com/eclipse/omr/pull/3866. This contains the refactoring suggested by Andrew above.

dibyendumajumdar:
Hi @vijaysun-omr @andrewcraik @Leonardo2718 

I have updated this PR. The only outstanding thing is to find a way to customize the optimizer flag to set callsAsUses flag. The test case I have created will fail or crash without this, hence it is not yet enabled in the build.

I would prefer to separately submit a change to add an option to set the optimizer flag as that might require adding an option controlled via a compiler option.

dibyendumajumdar:
This PR is not adequate as has been revealed by https://github.com/eclipse/omr/issues/4487 . Further changes are needed

PR #2825, opened 2018-08-07
Add support for CUDA 9.1 and 9.2

keithc-ca:
According to [1], CUDA Toolkit version 9.1 was released in December 2017 and version 9.2 was released in March 2018.

[1] https://developer.nvidia.com/cuda-toolkit-archive


charliegracie:
@genie-omr build all

PR #2826, opened 2018-08-07
Fix incorrect assertions

liqunl:
This commit does the following
 - Stop assertion for mergedWithOSRGuard in creating nopsite.
 - Class from arg info is not guaranteed to be more specific than class
   from method, thus choose the more specific one of the two and fix the
   assertion.


liqunl:
@cathyzhyi @andrewcraik Can you review?

0xdaryl:
@genie-omr build all

liqunl:
This PR needs more work since an issue is exposed in openj9.

cathyzhyi:
LGTM

liqunl:
@0xdaryl Could you help start a build and merge?

vijaysun-omr:
@liqunl could you please elaborate on the rationale for the change ? i.e. what is the assertion for and why are the categories of "merging" that are being tested for (both originally and with your change) okay here ?

liqunl:
@vijaysun-omr I updated the description for why we need to stop firing the assertion for mergedWithOSRGuard. I think the assertion's original purpose is to assure the guard is nopable, then we added an exception for mergedWithHCRGuard, now we need another exception for mergedWithOSRGuard. mergedWithOSRGuard should have been exempted during the work of nextGenHCR, I think we forgot to add it.

vijaysun-omr:
Okay thanks, could you also update the commit message with the more elaborate explanation please ?

liqunl:
Done.

PR #2827, opened 2018-08-08
AArch64: Initial version of OMRMemoryReference.generateBinaryEncoding()

knn-k:
This commit implements the initial version of generateBinaryEncoding()
in OMRMemoryReference for aarch64 and related functions.

### DIFF
--- a/compiler/aarch64/codegen/OMRMemoryReference.cpp
+++ b/compiler/aarch64/codegen/OMRMemoryReference.cpp
@@ -23,6 +23,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include "codegen/ARM64Instruction.hpp"
 #include "codegen/CodeGenerator.hpp"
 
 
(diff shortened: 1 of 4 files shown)


knn-k:
I used `TR::Instruction.getOpCode().getOpCodeBinaryEncoding()` in `generateBinaryEncoding()` for categorizing LD/ST instructions.  On the other hand, `TR::Instruction` is unavailable in `estimateBinaryLength()`.

The p version of `MemoryReference::estimateBinaryLength()` takes `TR::CodeGenerator` as its argument, while the 32-bit arm version of the function takes `TR_ARMOpCodes` instead.

The aarch64 version of `estimateBinaryLength()` should imitate the 32-bit arm version?

compiler/aarch64/codegen/OMRMemoryReference.cpp
```diff
@@ -86,17 +88,136 @@ bool OMR::ARM64::MemoryReference::useIndexedForm()
...
+         }
+      }
 
    return cursor;
    }
 
 
 uint32_t OMR::ARM64::MemoryReference::estimateBinaryLength(TR::CodeGenerator& codeGen)
    {
-   TR_ASSERT(false, "Not implemented yet.");
```
0xdaryl: I suppose the question is why this function takes a `TR::CodeGenerator` parameter (passed by reference no less) when it's not even used or needed.  I'd say remove it until it's needed (unless you believe some of the unimplemented paths are going to need it, in which case I'd say pass in the pointer to be consistent with most of the other code generators).

compiler/aarch64/codegen/OMRMemoryReference.cpp
```diff
@@ -23,7 +23,9 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include "codegen/ARM64Instruction.hpp"
 #include "codegen/CodeGenerator.hpp"
+#include "codegen/Instruction.hpp"
```
0xdaryl: Do you need to include this as well as `ARMInstruction.hpp` ?  ARMInstruction is an extension of Instruction so this must be included already.  Please remove.

knn-k:
Updated reflecting review comments.

knn-k:
I implemented `estimateBinaryLength()` taking `TR::InstOpCode` as its argument in #2887.

PR #2829, opened 2018-08-08
Add peephole replacing load-store with mem-mem instruction

dhong44:
Add peephole optimization before Z code gen to replace
Load-Store sequences such as L-ST and LG-STG with an
MVC instruction.

This will need to be done before RA since load-store
sequences can only be reduced if the use count of a
register is known. Since currently the peephole phase
is performed after RA, this change also adds a new
pre RA peephole phase.


dhong44:
@fjeremic

dhong44:
I ran a [daytrader7 benchmark](https://github.com/WASdev/sample.daytrader7) both with and without the load-store to MVC peephole.
image
The results show very little change between the baseline and with the optimization and differences in performance are statistically insignificant. Theoretically, replacing 2 instructions with a memory to memory instruction should improve performance and save the usage of a register and the results show a lot of previously missed opportunities being captured.

compiler/z/codegen/OMRCodeGenPhase.hpp
```diff
@@ -51,12 +51,15 @@ class OMR_EXTENSIBLE CodeGenPhase : public OMR::CodeGenPhase
...
    static void performSetBranchOnCountFlagPhase(TR::CodeGenerator * cg, TR::CodeGenPhase *);
+   static void performPreRAPeepholePhase(TR::CodeGenerator * cg, TR::CodeGenPhase *);
 
    // override base class implementation because new phases are being added
    static int getNumPhases();
    const char * getName();
    static const char* getName(PhaseValue phase);
 
+   // override base class to call new post RA peephole
+   static void performPeepholePhase(TR::CodeGenerator * cg, TR::CodeGenPhase *);
```
fjeremic: What does this comment mean exactly? Also can we move up this declaration up with the others?

compiler/z/codegen/S390Peephole.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+private:
+   /** \brief
+    *     Attempts to reduce L[' '|FH|G] R,MR1  ST[' '|FH|G] R,MR2 sequences to MVC MR2, MR1
+    *     to save a register and instruction.
+    *
+    *  \return
+    *     true if the reduction was successful; false otherwise.
+    */
+   bool LoadStoreReduction(TR::InstOpCode::Mnemonic storeOpCode, uint16_t size);
```
fjeremic: Although we are pretty inconsistent already what do you think of imposing a convention for these peepholes and naming this one `attemptLoadStoreReduction`? It is an attempt after all and is not guaranteed to succeed. Can we also document the parameters?

compiler/z/codegen/S390Peephole.cpp
```diff
@@ -0,0 +1,3248 @@
...
+
+void
+TR_S390PreRAPeephole::perform()
+   {
+   TR::Delimiter d(comp(), comp()->getOption(TR_TraceCG), "preRAPeephole");
+
+   if (comp()->getOption(TR_TraceCG))
+      printInfo("\nPost PreRegister Assignment Peephole Optimization Instructions:\n");
+
+   bool moveInstr;
```
fjeremic: I don't see a purpose for the existence of this variable. Are we able to fold this away?

compiler/z/codegen/S390Peephole.cpp
```diff
@@ -0,0 +1,3248 @@
...
+   if (_cursor->getNext()->getOpCodeValue() == storeOpCode)
+      {
+      TR::S390RXInstruction* loadInst = static_cast<TR::S390RXInstruction*> (_cursor);
+      TR::S390RXInstruction* storeInst = static_cast<TR::S390RXInstruction*> (_cursor->getNext());
+
+      if ( loadInst->getMemoryReference()->getSymbolReference() && loadInst->getMemoryReference()->getSymbolReference()->isUnresolved() ||
+          storeInst->getMemoryReference()->getSymbolReference() && storeInst->getMemoryReference()->getSymbolReference()->isUnresolved())
+         {
+         return false;
+         }
```
fjeremic: I think this deserves a one liner comment why it is necessary as it is non-obvious at first glance.

compiler/z/codegen/S390Peephole.cpp
```diff
@@ -0,0 +1,3248 @@
...
+      // Register must only be used in the load-store sequence, possibly used in the memory reference
+      // of the load however, not in the store since the the register would have to be loaded first.
+      ncount_t uses = 2;
+      uses += reg == loadInst->getMemoryReference()->getBaseRegister();
+
+      // Bail out if register could have future uses, or if either instruction has dependencies.
+      // Since 2 instructions are replaced with one before RA, we can't handle things like merging
+      // dependencies.
+      if (reg->getTotalUseCount() != uses || reg->isDependencySet() ||
+         loadInst->getDependencyConditions() || storeInst->getDependencyConditions())
```
fjeremic: This is the first time I'm seeing a use of `isDependencySet`. Why do we need this check? Is it not sufficient to just check whether `loadInst` and `storeInst` have dependencies on them? By the way a register being a part of a dependency list will increment its total use count so we cannot have the case where `reg->getTotalUseCount() == uses && reg->isDependencySet()` is true.

compiler/z/codegen/S390Peephole.cpp
dhong44: 79eadfb

compiler/z/codegen/S390Peephole.cpp
dhong44: 76e39d2

compiler/z/codegen/S390Peephole.cpp
dhong44: f497fd6

compiler/z/codegen/S390Peephole.hpp
dhong44: a456077

compiler/z/codegen/OMRCodeGenPhase.hpp
dhong44: 80d2864

fjeremic:
@genie-omr build all

PR #2830, opened 2018-08-08
Build blob with types that match superset

keithc-ca:
* stop including typedefs in superset
* refactor handling of opaque types
* make sure names are fully qualified
* omit `union` prefix for field types


charliegracie:
@genie-omr build all

keithc-ca:
The failure in https://travis-ci.org/eclipse/omr/jobs/413693029 looks like network troubles: can that job be restarted or is the success of the other travis jobs sufficient?

PR #2831, opened 2018-08-08
Fix JitBuilder Call to return the right type for Int8 and Int16

charliegracie:
Currently if a function returns a type smaller than Int32 the
returned value will be of type Int32 and not Int16 or Int8. I
added some tests that will assert in JitBuilder if the type of
the return value does not match the declaration. This is achieved
by adding the result to a Const<type> of the appropriate type.

The solution is to convert the result if callNode->getDataType()
does not match the defined return type. To make the change simpler
I modified the parameters of convertTo to take a TR::DataType
instead of a TR::IlType.

### DIFF
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -972,7 +972,7 @@ OMR::IlBuilder::ConvertTo(TR::IlType *t, TR::IlValue *v)
       TraceIL("IlBuilder[ %p ]::%d is ConvertTo (already has type %s) %d\n", this, v->getID(), t->getName(), v->getID());
       return v;
       }
-   TR::IlValue *convertedValue = convertTo(t, v, false);
+   TR::IlValue *convertedValue = convertTo(typeTo, v, false);
    TraceIL("IlBuilder[ %p ]::%d is ConvertTo(%s) %d\n", this, convertedValue->getID(), t->getName(), v->getID());
    return convertedValue;
    }
@@ -987,22 +987,21 @@ OMR::IlBuilder::UnsignedConvertTo(TR::IlType *t, TR::IlValue *v)
       TraceIL("IlBuilder[ %p ]::%d is UnsignedConvertTo (already has type %s) %d\n", this, v->getID(), t->getName(), v->getID());
       return v;
       }
-   TR::IlValue *convertedValue = convertTo(t, v, true);
+   TR::IlValue *convertedValue = convertTo(typeTo, v, true);
    TraceIL("IlBuilder[ %p ]::%d is UnsignedConvertTo(%s) %d\n", this, convertedValue->getID(), t->getName(), v->getID());
    return convertedValue;
    }
 
 TR::IlValue *
-OMR::IlBuilder::convertTo(TR::IlType *t, TR::IlValue *v, bool needUnsigned)
+OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned)
    {
    TR::DataType typeFrom = v->getDataType();
-   TR::DataType typeTo = t->getPrimitiveType();
 
    TR::ILOpCodes convertOp = ILOpCode::getProperConversion(typeFrom, typeTo, needUnsigned);
-   TR_ASSERT(convertOp != TR::BadILOp, "Builder [ %p ] unknown conversion requested for value %d (TR::DataType %d) to type %s", this, v->getID(), (int)typeFrom, t->getName());
+   TR_ASSERT(convertOp != TR::BadILOp, "Builder [ %p ] unknown conversion requested for value %d %s to %s", this, v->getID(), typeFrom.toString(), typeTo.toString());
 
    TR::Node *result = TR::Node::create(convertOp, 1, loadValue(v));
-   TR::IlValue *convertedValue = newValue(t, result);
+   TR::IlValue *convertedValue = newValue(typeTo, result);
    return convertedValue;
    }
 
@@ -2006,6 +2005,9 @@ OMR::IlBuilder::genCall(TR::SymbolReference *methodSymRef, int32_t numArgs, TR::
    if (returnType != TR::NoType)
       {
       TR::IlValue *returnValue = newValue(callNode->getDataType(), callNode);
+      if (returnType != callNode->getDataType())
+         returnValue = convertTo(returnType, returnValue, false);
+
       return returnValue;
       }
 
--- a/fvtest/jitbuildertest/CallReturnTest.cpp
+++ b/fvtest/jitbuildertest/CallReturnTest.cpp
@@ -1,0 +1,185 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ * or the Apache License, Version 2.0 which accompanies this distribution and
+ * is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following
+ * Secondary Licenses when the conditions for such availability set
+ * forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ * General Public License, version 2 with the GNU Classpath
+ * Exception [1] and GNU General Public License, version 2 with the
+ * OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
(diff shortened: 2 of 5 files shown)


charliegracie:
@genie-omr build all

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -985,22 +985,21 @@ OMR::IlBuilder::UnsignedConvertTo(TR::IlType *t, TR::IlValue *v)
...
 TR::IlValue *
-OMR::IlBuilder::convertTo(TR::IlType *t, TR::IlValue *v, bool needUnsigned)
+OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned)
    {
    TR::DataType typeFrom = v->getDataType();
-   TR::DataType typeTo = t->getPrimitiveType();
 
    TR::ILOpCodes convertOp = ILOpCode::getProperConversion(typeFrom, typeTo, needUnsigned);
-   TR_ASSERT(convertOp != TR::BadILOp, "Builder [ %p ] unknown conversion requested for value %d (TR::DataType %d) to type %s", this, v->getID(), (int)typeFrom, t->getName());
+   TR_ASSERT(convertOp != TR::BadILOp, "Builder [ %p ] unknown conversion requested for value %d (TR::DataType %d) to (TR::DataType %d)", this, v->getID(), (int)typeFrom, (int)typeTo);
```
Leonardo2718: I would much prefer to have the actual type names be printed here.

Can we use `typeFrom.toString(), typeTo.toString()` instead of `(int)typeFrom, (int)typeTo`?

charliegracie:
I will look into adding it to the jitbuilder fvtest but if it is going to to make a long time to figure it out I will not be able to do it.

compiler/ilgen/OMRIlBuilder.cpp
charliegracie: I will give it a try.  thanks!

charliegracie:
@genie-omr build all

fvtest/jitbuildertest/CallReturnTest.cpp
```diff
@@ -0,0 +1,173 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 IBM Corp. and others
```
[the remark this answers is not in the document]
charliegracie: No but most of the code was copied from code that was which means the initial copyright date needs to flow with it I believe.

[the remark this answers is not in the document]
Leonardo2718: I see. Fair enough.

fvtest/jitbuildertest/CallReturnTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+   }
+
+static int8_t
+testInt8(int8_t val)
+   {
+   #define TEST_INT8_LINE LINETOSTR(__LINE__)
+   return val;
+   }
+
+DEFINE_BUILDER(TestInt64RturnType,
```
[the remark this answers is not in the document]
charliegracie: Fixed all cases of the missing `e`

Leonardo2718:
Re the failures on `linux_390-64` and `linux_ppc-64_le_gcc` test failures, I believe these are due to missing trampoline support (@0xdaryl) and are unrelated to this change set. Do we have an easy way of filtering which tests are executed on different platforms? (In this case, I think adding `-ReturnTypeTest*` as a gtest filter should do the trick.)

charliegracie:
Not easily..... sadly that is not controlled via the repo yet.  I will modify the tests themselves to just not run them....   Sadly, this is becoming much more time consuming that I had originally thought it would be :(

charliegracie:
@genie-omr build all

charliegracie:
note there are lots of Tril Validation errors... Not sure if they are caused by my changes or not.

Leonardo2718:
Yes, those validation failures are expected. I believe the messages are not suppressed because they are printed to stderr.

charliegracie:
@genie-omr build pLinux,zLinux

fvtest/jitbuildertest/CallReturnTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+   ASSERT_COMPILE(TR::TypeDictionary, TestInt64ReturnType, testFunction);
+   ASSERT_EQ(testFunction(1), 2);
+   }
+
+typedef int32_t (*Int32ReturnType)(int32_t);
+TEST_F(ReturnTypeTest, Int32_Test)
+   {
+   Int32ReturnType testFunction;
+   ASSERT_COMPILE(TR::TypeDictionary, TestInt32ReturnType, testFunction);
+   ASSERT_EQ(testFunction(1), 2);
```
mstoodle: Maybe use an input value that requires more than 16 bits, just for completeness :) .

fvtest/jitbuildertest/CallReturnTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+   }
+
+typedef int16_t (*Int16ReturnType)(int16_t);
+TEST_F(ReturnTypeTest, Int16_Test)
+   {
+   Int16ReturnType testFunction;
+   ASSERT_COMPILE(TR::TypeDictionary, TestInt16ReturnType, testFunction);
+   ASSERT_EQ(testFunction(1), 2);
```
mstoodle: Same: value bigger than 8 bits?

charliegracie:
I am not sure how to test that the values are of the appropriate size explicitly.... By performing a math operation on the result and a value of the right size is the only way I know to test this. If the wrong type is returned the JitBuilder code will fail an assert that checks that both values in the Add operation are the same size.

ymanton:
Correct me if I'm wrong, but you're generating code to do a truncation at runtime however the problem is not in the runtime values, it's in our IL. The values returned from the call at runtime are in the correct range (if the function being called was compiled by a working compiler), it's our representation in the IL that's wrong because we don't have anything smaller than an `icall`.

ymanton:
> I think the bcall and scall implementations on most platforms would just end up doing a runtime truncation of the actual return value anyway would they not?

No, that wouldn't be necessary, the callee will return the appropriately sized value if it was compiled by a compiler that follows the platform's ABI.

The only potential exception to that are functions that are generated by the OMR compiler, because we don't have anything smaller than an `ireturn`. However if you were generating OMR IL for a function returning a 16- or 8-bit int and only had `ireturn` at your disposal you'd probably just do a `b2i` or `s2i`, so the value you'd return would still likely be in the correct range.

This is more of a typing problem in the compiler, especially on the caller side. Generating a truncation of the result at runtime will add a needless instruction, however I don't know how you can fix properly other than to implement the smaller call and return IL ops. If this isn't causing any problems at runtime you can always ignore it I suppose.

As far as testing goes, I'd suggest creating functions that take a bigger type and return a smaller one, that way your test can pass in values that fit and don't fit the return type and check that they are appropriately truncated. For example

```static int16_t testInt16(int32_t val) { return (int16_t)val; }```

which can be tested with parameters like

```
(int32_t)INT16_MIN - 1
INT16_MIN
0
INT16_MAX
(int32_t)INT16_MAX + 1
```

charliegracie:
Are there any technical issues with the change ignoring testing?

The current issue is an assertion in JitBuilder when trying to perform a math operation (Add) on the result of a call at runtime during method compilation.  The call is defined to a return an `Int8` and the value being added to it is `ConstInt8(1)`. I would like my code to work without explicitly adding casts, in my user code, to cast the result to an Int8. I properly defined the function to return an Int8 and in fact the helper I am calling does return an Int8. With my proposed change I no longer get the assertion during method compilation. I think that the OMR compiler should support b/sCall and b/sReturn but adding those featured is a little outside of my comfort zone and I do not have enough spare cycles in my spare time to complete that work in the foreseeable future.

ymanton:
The only technical issue I can see is that it creates an unnecessary instruction, but I agree that fixing the actual problem is no small task.

I assume the assert you're referring to is this one or one like it?

https://github.com/eclipse/omr/blob/3cc353d8375f38e82231516e9869bd3e43ac4030/compiler/ilgen/OMRIlBuilder.cpp#L1044-L1065

What happens if you make the assert a little looser and allow the types to differ if at least one of the operands is a call node, or for calls returning 16- and 8-bit integers substitute the return type of the method symbol in place of the type of the call node so the assert will succeed? If there's only one or a couple of these assertions I'd suggest considering that.

If/when the underlying issue is fixed properly someone would have to come back and change the code regardless of which temporary fix is adopted here, but at least with the modified assert there are no unnecessary instructions being generated to slow us down in the meantime.

ymanton:
Yes it will. That would be a good reason to accept this patch as it is I admit.

charliegracie:
@mstoodle ping

mstoodle:
Discussed test values with Charlie and I believe he'll be updating them, at which point I think this is ready to be merged.

charliegracie:
@genie-omr build all

charliegracie:
I am not sure the appveyor build will run in a reasonable amount of time... I am thinking about removing it as the resources seem very scarce these days.

mstoodle:
Thanks for all the work to fix this problem and construct tests to make sure it stays fixed @charliegracie !

Issue #2470, opened 2018-04-17
Early allocation of CodeCacheHashEntrySlab


0xdaryl:
I'm trying to understand this comment:

https://github.com/eclipse/omr/blob/2f938a7b9ac34708a16ae170bc92157eed944fb1/compiler/runtime/OMRCodeCache.cpp#L1697-L1698

I'm considering sinking that allocation down to where it actually needs to be as part of some broader work on trampolines (rather than allocating it before the code cache and then passing it down as a parameter), but I'm cautious about the implications and side effects.  That comment is from shortly after Testarossa's Planck Epoch, so I'm not sure if its message is still correct or relevant.

From reading the code, I believe the reason the allocation is done earlier (i.e., here) is because:
1)  If the allocation fails there's no point in even attempting to allocate the code cache because it will also fail.  Hence, this is a quicker way to determine allocation failure, and move on.
2)  If there is an allocation failure during code cache creation (which involves several functions), it will be easier to de-allocate the `CodeCacheHashEntrySlab` on failure in this function.

I don't believe 1) is really an interesting path to be optimizing, and I'm not convinced that 2) is really a complicated scenario that can't be easily solved.

Do any of the other grey beards have a better recollection on why this allocation happens here and the implications of sinking it?

@zl-wang @mpirvu @mstoodle

mpirvu:
That's exactly how I interpret the comment as well (I don't remember to have written that comment).
If the slab is allocated later we must be careful with exposing the newly created code cache to the outside world as it it not fully initialized (the slab is used in `codeCache->initialize()`. I believe that once you attach it to global list with `manager->addCodeCache(codeCache)` it's up for grabs, though marking it as `reserved` prevents other parties from being interested in it.

mstoodle:
Like @mpirvu , I do not recall the actual reasons, but from glancing at the code it does appear that quite a bit of stuff happens in the initialization of an allocated code cache, including updating at least one field in the manager object. I also wonder if deallocation could be guaranteed not to "interfere" with another thread allocating a code cache at the same time from the repository (i.e. leaving a gap that the repository would then need to manage).

There currently isn't a `deallocateCodeCacheObject()`, but it would need to be kept in sync with any code updates to `initialize()` which may have been deemed less attractive than simply allocating two objects in one order rather than the other way around. 

Is there some natural ordering that suggests the code cache should really be allocated before the hash entry slab?

zl-wang:
[...]

PR #2832, opened 2018-08-09
Move allocation of CodeCacheHashEntrySlab to CodeCache initialization

0xdaryl:
Currently, the initial CodeCacheHashEntrySlab is allocated before the
CodeCache object is allocated and then passed in as a parameter to the
CodeCache object initialization.  This is done mainly for the reasons
discussed in Issue #2470.

This commit moves the allocation of the CodeCacheHashEntrySlab object
and the first slab to the CodeCache::initialize() function.  Handling
of allocation failures has also been adjusted to accommodate the new
location to ensure memory is reclaimed.

This change is necessary to consolidate trampoline artifact creation
to a single location to facilitate further refactoring in this area.

### DIFF
--- a/compiler/runtime/OMRCodeCacheManager.cpp
+++ b/compiler/runtime/OMRCodeCacheManager.cpp
@@ -100,7 +100,7 @@ OMR::CodeCacheManager::initialize(
 #endif // HOST_OS == OMR_LINUX
 
    TR::CodeCacheConfig &config = self()->codeCacheConfig();
-      
+
    if (allocateMonolithicCodeCache)
       {
       size_t size = config.codeCacheTotalKB() * 1024;
@@ -156,7 +156,7 @@ OMR::CodeCacheManager::initialize(
             config.codeCacheTotalKB() <= REACHEABLE_RANGE_KB)
 #endif
         );
-   
+
    _lowCodeCacheSpaceThresholdReached = false;
 
    _initialized = true;
@@ -220,7 +220,7 @@ OMR::CodeCacheManager::destroy()
       int numCharsWritten = snprintf(elfFilename, maxElfFilenameSize, "/tmp/perf-%d.jit", jvmPid);
       if (numCharsWritten > 0 && numCharsWritten < maxElfFilenameSize)
       {
-         TR_ASSERT(_elfExecutableGenerator->emitELF((const char*) &elfFilename, 
+         TR_ASSERT(_elfExecutableGenerator->emitELF((const char*) &elfFilename,
                                 _symbolContainer->_head,
                                 _symbolContainer->_numSymbols,
                                 _symbolContainer->_totalSymbolNameLength),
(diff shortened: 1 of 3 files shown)


0xdaryl:
@ymanton @zl-wang @fjeremic : I'd appreciate a review please.

This is a WIP while some deeper testing completes and to get some early review comments.

A change also needs to be introduced into OpenJ9 to change its implementation of the `CodeCache::initialize` method to match this API.  Admittedly, I'm still working on the best way of staging these two changes.

compiler/runtime/OMRCodeCacheManager.hpp
```diff
@@ -144,8 +144,7 @@ class OMR_EXTENSIBLE CodeCacheManager
    void  freeMemory(void *memoryToFree);
 
    TR::CodeCache * allocateCodeCacheObject(TR::CodeCacheMemorySegment *codeCacheSegment,
-                                           size_t codeCacheSizeAllocated,
-                                           CodeCacheHashEntrySlab *hashEntrySlab);
+                                           size_t codeCacheSizeAllocated);
```
fjeremic: This parameter sounds incorrect. Shouldn't it be `codeCacheSize` or `codeCacheSizeToAllocate`? "Allocated" implies a return value and this parameter is not of a reference type. Might be a good chance to improve this.

0xdaryl: Yes, that's a good point.  I think it was named as such because at its call site its argument is named `codeCacheSizeAllocated` which is a reference argument to `getNewCacheMemorySegment`.  However, there's no need to carry that naming convention to the parameter of this API.  I'll change it.

0xdaryl:
Here is the integration plan to stage these changes to prevent build breakages:

1.  Introduce `OMR::CodeCache::initialize(TR::CodeCacheManager *manager, TR::CodeCacheMemorySegment *codeCacheSegment, size_t codeCacheSizeToAllocate)` in its own Eclipse OMR PR (#TBD)
2. Introduce `J9::CodeCache::initialize(TR::CodeCacheManager *manager, TR::CodeCacheMemorySegment *codeCacheSegment, size_t codeCacheSizeToAllocate)` in its own Eclipse OpenJ9 PR (#TBD)
3.  This PR, modulo the changes from 1)
4.  Remove the now deprecated 4-child `J9::CodeCache::initialize( ... )` from Eclipse OpenJ9 (#TBD).

zl-wang:
Looks good to me ...

compiler/runtime/OMRCodeCacheManager.cpp
```diff
@@ -254,15 +254,18 @@ OMR::CodeCacheManager::freeMemory(void *memoryToFree)
...
    if (codeCache)
       {
       new (codeCache) TR::CodeCache();
-      if (!codeCache->initialize(self(), codeCacheSegment, codeCacheSizeAllocated, hashEntrySlab))
+      if (!codeCache->initialize(self(), codeCacheSegment, codeCacheSizeAllocated))
          {
+         if (codeCache->hashEntrySlab())
+            {
+            codeCache->hashEntrySlab()->free(self());
+            }
```
ymanton: Now that the `CodeCache` object allocates the slab itself it seems like a bad idea for the caller to have to clean it up. Can't this be done in `initialize()` if it fails?

0xdaryl: Yes it can and I agree that's a better place for it.  I'll move it.

fjeremic:
@ymanton any other concerns?

ymanton:
No, LGTM.

fjeremic:
@genie-omr build all

PR #2561, opened 2018-05-18
Consolidate InstOpCode tables on Z and improve documentation

fjeremic:
This change consists of several logical parts outlined in the commits and summarized as follows:

_Consolidate S390OpProp and S390OpProp2 properties into one list_

Since the deprecation of unused instruction properties as part of #2151
we now have space to consolidate the two property lists. We migrate all
properties into a single list and deprecate S390OpProp2 properties.

_Merge encodings, names, and properties into a single table_

Refactor OpCodeBinaryEntry to include all the properties of an
instruction. An example of the BRC instruction is as follows:

```
   {
   /* .mnemonic    = */ TR::InstOpCode::BRC,
   /* .name        = */ "BRC",
   /* .description = */ "Branch On Condition",
   /* .opcode[0]   = */ 0xA7,
   /* .opcode[1]   = */ 0x04,
   /* .format      = */ RI_FORMAT,
   /* .minimumALS  = */ TR_S390ProcessorInfo::TR_z900,
   /* .properties  = */ S390OpProp_BranchOp |
                        S390OpProp_ReadsCC
   },
```

To accomplish this a script was used (detailed in the associated PR) to
parse all the data from ArchInstOpCode* files and consolidate them into
OMROpCodeProperties.hpp.

_Order instructions by minimum ALS then by mnemonic_

Partition all instructions by architecture level set (ALS) first and
then alphabetize based on mnemonic within each partition. This
simplifies human search for instructions as we typically look for
whether an instruction is supported on a particular architecture.

_Deprecate InstOpCodeTables and rename OpCodeBinaryEncoding_

InstOpCodeTables is now consolidated into a single table the
declaration of which can be moved to InstOpCode.cpp. Further we rename
OpCodeBinaryEncoding to OpCodeMetaData to better reflect the contents
of each entry.

We also take this opportunity to further document the InstOpCode class.

Signed-off-by: Filip Jeremic <fjeremic@ca.ibm.com>

PR #2840, opened 2018-08-13
Update PPA format to RRF2

fjeremic:
According to PoPs the PPA instruction is an RRF-c format instruction
with no third register and a mask, hence it should be labeled as an
RRF2 instruction.

Note that the .format field in the table modified is only used as a
reference and has no bearing on any execution path.

### DIFF
--- a/compiler/z/codegen/OMRInstOpCodeProperties.hpp
+++ b/compiler/z/codegen/OMRInstOpCodeProperties.hpp
@@ -13105,7 +13105,7 @@
    /* .description = */ "Perform Processor Assist",
    /* .opcode[0]   = */ 0xB2,
    /* .opcode[1]   = */ 0xE8,
-   /* .format      = */ RRF_FORMAT,
+   /* .format      = */ RRF2_FORMAT,
    /* .minimumALS  = */ TR_S390ProcessorInfo::TR_zEC12,
    /* .properties  = */ S390OpProp_UsesTarget
    },


fjeremic:
@zl-wang could I please get a review?

zl-wang:
are you sure the format is irrelevant?  i saw RRFInstruction::generateBinaryEncoding() has both cases though:
       case TR::Instruction::IsRRF:
           ...
       case TR::Instruction::IsRRF2:
           ...

fjeremic:
> are you sure the format is irrelevant?

Yes, I'm certain. There are no uses of either `RRF_FORMAT` or `RRF2_FORMAT` in the entire codebase.



> i saw RRFInstruction::generateBinaryEncoding() has both cases though:
> case TR::Instruction::IsRRF:
> ...
> case TR::Instruction::IsRRF2:
> ...

The code you are referencing is here:

https://github.com/eclipse/omr/blob/ca27bb0ebec3ecdc4dfefc965b6cc0122916023e/compiler/z/codegen/S390Instruction.cpp#L1702-L1800

The switch case relies on a query to `getKind()` which is defined here:

https://github.com/eclipse/omr/blob/ca27bb0ebec3ecdc4dfefc965b6cc0122916023e/compiler/z/codegen/S390Instruction.hpp#L1916-L1937

Hence the "format" of an instruction is not determined by the `.format` within the `OpCodeMetaData` struct (the code this PR modifies) but rather by whether certain arguments are present when constructin the instruction object. i.e. whether the `_isMask3Present` and `_isMask4Present` fields are set. These are set in the constructors, for example:

https://github.com/eclipse/omr/blob/ca27bb0ebec3ecdc4dfefc965b6cc0122916023e/compiler/z/codegen/S390Instruction.hpp#L1885-L1895

If you wish to use the `PPA` instruction you can simply use the following API passing `true` to `isMask3` argument:

https://github.com/eclipse/omr/blob/104ab0bab2e858967d64de59c68fdff8816fd082/compiler/z/codegen/S390GenerateInstructions.cpp#L635

which will use this constructor:

https://github.com/eclipse/omr/blob/104ab0bab2e858967d64de59c68fdff8816fd082/compiler/z/codegen/S390Instruction.hpp#L1821-L1835

And hence `_isMask3Present == true && _isMask4Present == false` meaning `getKind()` will return `IsRRF2`:

https://github.com/eclipse/omr/blob/104ab0bab2e858967d64de59c68fdff8816fd082/compiler/z/codegen/S390Instruction.hpp#L1938-L1944

There is no mention of `.format` anywhere in the above logic.

0xdaryl:
@genie-omr build zlinux,zos

0xdaryl:
Looks OK to me.  But I'll wait for @zl-wang's thumbs up on your reply before merging.

zl-wang:
Interesting derivation of instr format ... I am fine with it.

fjeremic:
> Interesting derivation of instr format ... I am fine with it.

I agree! It definitely shouldn't be this cumbersome. We only recently added the `.format` field as part of #2561. We need to find some time to clean this up and simply use the `.format`, but as you noted it may not always be correct so it is definitely a non-trivial quantity of work.

fjeremic:
I've opened up #2848 to address the issues mentioned above so we don't forget. We'll get to it eventually.

Issue #2180, opened 2017-12-22
Provide an implementation for findOrCreateLiteral in Z CodeGenerator


0xdaryl:
The only implementation of `findOrCreateLiteral()` in the FrontEnd always ends in an unimplemented assertion, despite being on seemingly active code paths in the Z code generator.  No known downstream project provides an implementation either.

This issue was discussed with @fjeremic prior to open-sourcing this code, and the following is a summary of our discussion at that time:

> The `FrontEnd::findOrCreateLiteral` API is currently unimplemented and will assert when called. This API is used in several DFP (Decimal Floating Point) paths which are off by default.  Historically, this function was only ever implemented by a static language compiler that was not open-sourced that leveraged those DFP features.
> 
> The task here is to deprecate the `FrontEnd::findOrCreateLiteral` API.  We need to use `OMR::Z::CodeGenerator::findOrCreateConstant` instead which will return a constant data snippet.  Note that the instructions currently using `FrontEnd::findOrCreateLiteral` (namely RX instruction format) do not support using constant data snippets, and this would also need to be implemented in such cases.
> 
> We do not have any tests in any downstream project to exercise these paths.  Hence, part of the work for this task will be to create tests that can exercise them, possibly with Tril.

  I will leave this issue marked as a "Discussion" until we decide on how to proceed.  If we decide to provide an implementation then this will become an Epic task to track the work items.

0xdaryl:
In the very short term, the `findOrCreateLiteral` function can move from the FrontEnd class into the `OMR::Z::CodeGenerator` class and the references fixed up.  It will still be unimplemented.

PR #2843, opened 2018-08-13
Create an implementation of findOrCreateLiteral in Z CodeGenerator

0xdaryl:
This commit does not provide an implementation for findOrCreateLiteral
but simply produces an equivalent stub in the Z CodeGenerator class.
This is the first step in a series of staged commits to safely remove
it from the FrontEnd class.

Issue #2180


fjeremic:
@genie-omr build zlinux,zos

Issue #2844, opened 2018-08-14
Migrate isInlineableNativeMethod from FrontEnd to inliner policy


0xdaryl:
The deprecated FrontEnd class contains a function `isInlineableNativeMethod` that is used exclusively by the inliner and should be part of the inliner policy class.

The work to refactor the FrontEnd interface will be starting up again, and moving existing FrontEnd functions to where they should actually live is one step in that process.  The following sequence of PRs will be used to stage the integration of this migration with no project breakages:

- [ ] Introduce the virtual `isInlineableNativeMethod` API in `OMR_InlinerPolicy` (PR TBD)
- [ ] Introduce the virtual `isInlineableNativeMethod` API in `TR_J9InlinerPolicy` whose implementation closely follows the version in VMJ9.cpp (PR TBD)
- [ ] Redirect OMR caller to the new policy API and remove the FrontEnd API (PR TBD)
- [ ] Remove the OpenJ9 FrontEnd API (PR TBD)

PR #2845, opened 2018-08-14
Introduce isInlineableNativeMethod API in OMR_InlinerPolicy

0xdaryl:
The deprecated FrontEnd interface contains a function `isInlineableNativeMethod`
that should be part of the `OMR_InlinerPolicy` class.  Introduce a
new virtual function into that policy class as the first step
to migrating it from the FrontEnd interface.

Issue #2844


0xdaryl:
@vijaysun-omr @andrewcraik : I'd appreciate a review and merge please.

andrewcraik:
this is an appropriate API - we'll have to see if anything more than the recognized method is required to be able to answer the query, but I agree with the direction - it is a policy query and belongs there in the long run.

vijaysun-omr:
Looks like a valid basis to build on.

PR #2847, opened 2018-08-14
 Resolve requirement to have a pointerTo(pointerTo(type)) 

charliegracie:
Currently in OMRVirtualMachineRegisterInStruct the field in the struct is required to be a pointer to a pointer to a type or you will crash during compilation. The only requirement is that the field is a pointer. Fetch elementType as the baseType for the field and if it is a pointer then the elementType is actually Word.

I corrected OperandStackTest to properly define it's struct and also cleaned up OperandArrayTest.

I also enabled the extra JitBuilder tests for the 2 platforms they can run on by default so we do not miss any failures introduced into these tests.


charliegracie:
@Leonardo2718 any thoughts?

mstoodle:
@jduimovich I could be wrong (my memory is hazy on this point) but I seem to recall you coming across this issue at some point, so interested to hear your thoughts on it. Perhaps you just made the same observation @charliegracie is making, but I want to make sure there isn't some reasonable motivating reason among our stakeholders to define the field in this particular way.

charliegracie:
@genie-omr build all

Leonardo2718:
LGTM. I think this change makes a lot of sense.

0xdaryl:
@jduimovich : not sure if you saw @mstoodle's question above?

jduimovich:
@0xdaryl  thanks for the ping.
@mstoodle  yes this is the same issue I was having. SP (to me) is a pointerTo(stack_elementType) so this is more  correct than pointerTo(pointerTo(stack_elementType)). With the check for the StackElementType being a pointers this should be backwards compatible assuming Word is *always* pointer-size and this continues to allow users who want stack elements to be defined as pointers instead of Int(xx) to work .

In terms of forward compatibility, I don't think I'll  hit any new verify checks due to SP references that used to return pointers when I change my code to the new way and anyone who does have stacks of pointer elements is already using it "correctly".    I use tagged "pointers" so I already have the casting convertTo(s) for integer tag checks and the object dereferences once I know what the thing is ... and of course, these fine preparatory changes due to the last set of verify "upgrades" so thanks to @Leonardo2718  :)

compiler/ilgen/OMRVirtualMachineRegisterInStruct.hpp
```diff
@@ -67,7 +67,11 @@ class VirtualMachineRegisterInStruct : public TR::VirtualMachineRegister
       _fieldName(fieldName),
       _localNameHoldingStructAddress(localNameHoldingStructAddress)
       {
-      _elementType = b->typeDictionary()->GetFieldType(structName, fieldName)->baseType()->baseType();
+      _elementType = b->typeDictionary()->GetFieldType(structName, fieldName)->baseType();
```
mstoodle: I find this code confusing because _elementType is named for what we store there, but it's not well named for the purpose for which we use it (to cook up integer values when we want to adjust the register's value). Confusing because the _elementType is not Word even if the field's base type is a pointer type. But it's a perfectly reasonable thing to do once you look at how _elementType is used in OMR::VirtualMachineRegister.
 
I'd like to propose that we also change the name of _elementType (in OMR::VirtualMachineRegister) to something like _integerTypeForAdjustments, which can be initialized as _elementType is being initialized below. At that point, I don't think I'll find this code confusing at all, which will at least make me happier.

Can I ask you to update the pull request with that change, please, @charliegracie ?

jduimovich:
FYI I tested Charlies pull request and it works for my use case. I tested the config with no change to my code as well as after removal of the extra pointerTo that had been required.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
FYI @mstoodle I changed GetWord to getWord but also found another issue with the code when I tried to represent a "PC" as a VirtualMachineRegisterInStruct so I refactored it some more.

charliegracie:
@mstoodle I removed the change to enable testing as that is better suited in a PR on it's own.  I will open a PR for that sometime soon.

mstoodle:
Thanks Charlie, that resolved my only concern that the condition wasn't quite the right one to use. I'll merge after pedantically waiting for the testing to prove nothing untoward has happened. Please feel free to ping me if you notice it's done before I do.

PR #2851, opened 2018-08-14
Change implementation of JitBuilder's IfAnd and IfOr services

Leonardo2718:
As was done in #2821 for JitBuilder's `Switch()` service, the new class `JBCondition` is added to encapsulate the different pieces needed to use the `IfAnd()` and `IfOr()` services. This avoids having to keep track of the different pieces independently in the implementation and also simplifies generating language bindings for the services.

This is a breaking change as the API is not backwards compatible. Code of the form:

```c++
IfAnd(&returnTrue, &returnFalse, 3,
    equiv1, rc12Equal,
    equiv2, rc23Equal,
    equiv3, rc34Equal);
```
must be changed to
```c++
IfAnd(&returnTrue, &returnFalse, 3,
   MakeCondition(equiv1, rc12Equal),
   MakeCondition(equiv2, rc23Equal),
   MakeCondition(equiv3, rc34Equal));
```

FYI @mstoodle @jduimovich @charliegracie @rwy0717

### DIFF
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -1640,6 +1640,14 @@ OMR::IlBuilder::IfOr(TR::IlBuilder **anyTrueBuilder, TR::IlBuilder **allFalseBui
    setComesBack();
    }
 
+TR::IlBuilder::JBCondition *
+OMR::IlBuilder::MakeCondition(TR::IlBuilder *conditionBuilder, TR::IlValue *conditionValue)
+   {
+   TR_ASSERT(conditionBuilder != NULL, "MakeCondition needs to have non-null conditionBuilder");
+   TR_ASSERT(conditionValue != NULL, "MakeCondition needs to have non-null conditionValue");
+   return new (_comp->trHeapMemory()) JBCondition(conditionBuilder, conditionValue);
+   }
+
 TR::IlValue *
 OMR::IlBuilder::EqualTo(TR::IlValue *left, TR::IlValue *right)
    {
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -1532,10 +1532,12 @@ OMR::IlBuilder::UnsignedShiftR(TR::IlValue *v, TR::IlValue *amount)
 
 /*
  * @brief IfAnd service for constructing short circuit AND conditional nests (like the && operator)
- * @param allTrueBuilder builder containing operations to execute if all conditional tests evaluate to true
+ * @param allTrueBuilder builder containing operations to execute if all conditional tests evaluate 
+ *        to true (automatically allocated if pointed-to pointer is null)
  * @param anyFalseBuilder builder containing operations to execute if any conditional test is false
+ *        (automatically allocated if pointed-to pointer is null)
  * @param numTerms the number of conditional terms
- * @param ... for each term, provide a TR::IlBuilder object and a TR::IlValue object that evaluates a condition (builder is where all the operations to evaluate the condition go, the value is the final result of the condition)
+ * @param terms array of JBCondition instances that evaluate a condition
  *
  * Example:
  * TR::IlBuilder *cond1Builder = OrphanBuilder();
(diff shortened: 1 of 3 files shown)


compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1640,6 +1673,50 @@ OMR::IlBuilder::IfOr(TR::IlBuilder **anyTrueBuilder, TR::IlBuilder **allFalseBui
...
+ * TR::IlBuilder *cond1Builder = OrphanBuilder();
+ * TR::IlValue *cond1 = cond1Builder->LessThan(
+ *                      cond1Builder->   Load("x"),
+ *                      cond1Builder->   Load("lower"));
+ * TR::IlBuilder *cond2Builder = OrphanBuilder();
+ * TR::IlValue *cond2 = cond2Builder->GreaterOrEqual(
+ *                      cond2Builder->   Load("x"),
+ *                      cond2Builder->   Load("upper"));
+ * TR::IlBuilder *inRange = NULL, *outOfRange = NULL;
+ * IfOr(&outOfRange, &inRange, 2, MakeCondition(cond1Builder, cond1), Makecondition(cond2Builder, cond2));
```
0xdaryl: Minor typo:   2nd `MakeCondition`

0xdaryl:
@genie-omr build all

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1582,12 +1580,49 @@ OMR::IlBuilder::IfAnd(TR::IlBuilder **allTrueBuilder, TR::IlBuilder **anyFalseBu
...
+   va_end(args);
+
+   IfAnd(allTrueBuilder, anyFalseBuilder, numTerms, terms);
+   }
+
 /*
  * @brief IfOr service for constructing short circuit OR conditional nests (like the || operator)
- * @param anyTrueBuilder builder containing operations to execute if any conditional test evaluates to true
- * @param allFalseBuilder builder containing operations to execute if all conditional tests are false
+ * @param anyTrueBuilder builder containing operations to execute if any conditional test evaluates to true (automatically allocated if pointed-to pointer is null)
```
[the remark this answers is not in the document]
Leonardo2718: Good point. Fixed.

compiler/ilgen/OMRIlBuilder.cpp
Leonardo2718: Fixed.

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1640,6 +1679,52 @@ OMR::IlBuilder::IfOr(TR::IlBuilder **anyTrueBuilder, TR::IlBuilder **allFalseBui
...
+   va_end(args);
+
+   IfOr(anyTrueBuilder, allFalseBuilder, numTerms, terms);
+   }
+
+TR::IlBuilder::JBCondition *
+OMR::IlBuilder::MakeCondition(TR::IlBuilder *conditionBuilder, TR::IlValue *conditionValue)
+   {
+   TR_ASSERT(conditionBuilder != NULL, "MakeCondition needs to have non-null conditionBuilder");
+   return new (_comp->region()) JBCondition(conditionBuilder, conditionValue);
```
mstoodle: any reason you're using region() here rather than allocateHeapMemory() like everywhere else in the file?

While region() is probably the right answer here (and everywhere), I'd prefer to see all of the allocateHeapMemory() calls changed to _comp->region() calls in a separate commit please.

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1640,6 +1679,52 @@ OMR::IlBuilder::IfOr(TR::IlBuilder **anyTrueBuilder, TR::IlBuilder **allFalseBui
...
+      }
+   va_end(args);
+
+   IfOr(anyTrueBuilder, allFalseBuilder, numTerms, terms);
+   }
+
+TR::IlBuilder::JBCondition *
+OMR::IlBuilder::MakeCondition(TR::IlBuilder *conditionBuilder, TR::IlValue *conditionValue)
+   {
+   TR_ASSERT(conditionBuilder != NULL, "MakeCondition needs to have non-null conditionBuilder");
```
mstoodle: should assert on null conditionValue as well...

Leonardo2718: Yes... yes it should. Fixed.

compiler/ilgen/OMRIlBuilder.cpp
Leonardo2718: Well, my thinking was that this is not a "change" but an "addition", so might as well use the correct thing. Fixed.

PR #2852, opened 2018-08-14
Support lcall on X86-32

0dvictor:
lcall is now supported on X86-32.


0xdaryl:
@genie-omr build all

PR #2854, opened 2018-08-14
Implement omrsig_is_signal_ignored

babsingh:
1) **Add stubs for omrsig_is_signal_ignored**

	omrsig_is_signal_ignored checks if the signal is ignored by the OS. If
	the signal is ignored, isSignalIgnored is set to TRUE. Otherwise,
	isSignalIgnored is set to FALSE.

	If an error is encountered during evaluation, OMRPORT_SIG_ERROR is
	returned. 0 is returned on success. If OMRPORT_SIG_ERROR is returned,
	then the value of isSignalIgnored should be ignored.

2) **Implement omrsig_is_signal_ignored [unix]**

	For unix, implementation for omrsig_is_signal_ignored has been added.

	For win32 and win64amd, sigaction doesn't exist. There is no way to
	check if a signal is ignored on win32/win64amd. omrsig_is_signal_ignored
	should always return FALSE on win32/win64amd. The stub implementation is
	sufficient on win32/win64.

	For ztpf, an implementation of omrsig_is_signal_ignored will be
	required. ztpf implementation should be identical to the unix
	implementation.


charliegracie:
@genie-omr build all

PR #2855, opened 2018-08-14
Allow OMRPORT_OMRSIG_SUPPORT to be enabled and disabled globally

babsingh:
These changes will allow a consumer of OMR to enable and disable
OMRPORT_OMRSIG_SUPPORT via the configure files.


babsingh:
@dnakamura can you verify if I covered all CMake dependencies?

charliegracie:
@dnakamura ping

dnakamura:
LGTM

cmake/config.cmake
```diff
@@ -161,6 +161,7 @@ set(OMR_PORT_CAN_RESERVE_SPECIFIC_ADDRESS ON CACHE BOOL "TODO: Document")
 set(OMR_PORT_NUMA_SUPPORT OFF CACHE BOOL "TODO: Document")
 set(OMR_PORT_ALLOCATE_TOP_DOWN OFF CACHE BOOL "TODO: Document")
 set(OMR_PORT_ZOS_CEEHDLRSUPPORT OFF CACHE BOOL "TODO: Document")
+set(OMRPORT_OMRSIG_SUPPORT OFF CACHE BOOL "TODO: Document")
```
charliegracie: Should this be defaulted to OFF or ON? @dnakamura didn't you think this should just be on all of the time?

babsingh: by default, `OMRPORT_OMRSIG_SUPPORT` is turned off in OMR. OpenJ9 enables `OMRPORT_OMRSIG_SUPPORT` while building DDR. in OMR, i feel that `OMRPORT_OMRSIG_SUPPORT` should be disabled by default. if a consumer of OMR wants signal chaining, then they should enable it while glueing OMR.

charliegracie:
@genie-omr build all

PR #2862, opened 2018-08-15
Remove checking REX.W when encoding on X86-32

0dvictor:
The assumption which REX.W can only be set on X86-64 is no-longer true.
For example, REX.W of FMA instructions is set even on X86-32.
Therefore, the assertion should be removed.


0xdaryl:
@genie-omr build win

PR #2863, opened 2018-08-15
Build AArch64 Compiler Files

aarongraham9:
Build AArch64 Compiler Files

- The Makefile in the JITBuilder was trying to use a wildcard operation to get all of the .cpp files in the AArch64 Compiler directory but it was failing to do so.
- Now explicitly lists all of the .ccp files as is convention.

For Issue #2839

Signed-off-by: Aaron Graham <aaron.graham@unb.ca>

PR #2868, opened 2018-08-15
AArch64: Remove CodeGenerator parm from MemoryReference::estimateBinaryLength()

knn-k:
This commit adds CodeGenerator to the parameter list of
OMR::ARM64::MemoryReference::estimateBinaryLength().

Issue: #2863


knn-k:
I once removed CodeGenerator from OMR::ARM64::MemoryReference::estimateBinaryLength() following the review comment in #2827.  Reverting the change.

0xdaryl:
Why revert the change rather than change the caller?  I thought you were going to remove the unneeded CodeGenerator reference argument in ARM64BinaryEncoding instead.

knn-k:
I thought reverting the change was the quickest way to avoid the build error for the time being.  I will have to add something else to MemoryReference::estimateBinaryLength() parameters later.

I changed the callers in ARM64BinaryEncoding.cpp.

0xdaryl:
AArch64-only change.  No need to wait for the AppVeyor build.  Merging.

PR #2869, opened 2018-08-16
Define missing instruction formats in the Z codegen

fjeremic:
Define missing instruction formats as outlined in in PoPs Chapter 5: Program Execution - Instructions - Instruction Formats. The formats are defined in the same order as listed in the PoPs document.

### DIFF
--- a/compiler/z/codegen/InstOpCode.cpp
+++ b/compiler/z/codegen/InstOpCode.cpp
@@ -153,7 +153,6 @@ OMR::Z::InstOpCode::copyBinaryToBufferWithoutClear(uint8_t *cursor, TR::InstOpCo
       case RXE_FORMAT:
       case RXY_FORMAT:
       case RXF_FORMAT:
-      case RSE_FORMAT:
       case RSY_FORMAT:
       case RSL_FORMAT:
       case RIE_FORMAT:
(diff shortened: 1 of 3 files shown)


fjeremic:
@r30shah could I please get a review? Note we could not eliminate some of the CC properties as they are actually being used in the codegen to avoid generating compare instructions if possible. For example when evaluating an `ificmpeq` the codegen will check if we are comparing against 0 and if the register containing the value we are about to compare is a register which was defined by an instruction which sets CC info.

For example if the register we are about to compare was last referenced in an `ALG` instruction because we tagged `ALG` with `S390OpProp_SetsZeroFlag` we will know that we have valid arithmetic CC information at the time we are about to evaluate the `ifcmpeq` and hence we can omit to generate a compare instruction since we are comparing for 0 and the `ALG` instruction sets the zero flag if the result was zero. This avoids generating an extra compare.

https://github.com/eclipse/omr/blob/fb3c9e04d4f32578ad8c45ceb6ca470c3087d5ca/compiler/z/codegen/OMRTreeEvaluator.cpp#L1011-L1023

fjeremic:
Travis ARM failure is due to infrastructure issues. I'll kick off the genie builds preemptively here.

@genie-omr build zlinux,zos

fjeremic:
Travis ARM failure is due to infrastructure issues. I'll kick off the genie builds preemptively here.

@genie-omr build zlinux,zos

fjeremic:
Builds passed. @0xdaryl could I please get a review?

0xdaryl:
Strange.  The merge conflicts just appeared.  Please resolve.

PR #2872, opened 2018-08-16
Read barrier stats

amicic:
Introduce two Concurrent Scavenger specific read barrier stats counts:
- number of objects being copied 
- number of reference slots being updated

Maintain them in ScavengerCopyScanRatio structure, so that we can report
progress of the counts throughout a GC cycle (reporting itself is done
by downstream project, like OpenJ9).


charliegracie:
@genie-omr build all

PR #2874, opened 2018-08-17
Add IncompatibleClassChangeError helper aliasing

jdmpapin:


jdmpapin:
Nothing yet. With eclipse/openj9#2648, OpenJ9 will put it on a `ZEROCHK` that checks the type of the receiver of an `invokeinterface` bytecode instruction that doesn't go through interface dispatch

Issue #2812, opened 2018-07-30
Adding new IL Opcode xrdbar/xrdbari 


cathyzhyi:
We are proposing to add xrdbar/xrdbari opcodes because 2 recent development items need to monitor instance or static fields reads and writes. One is field watch in debugging mode where we need to report fields accesses to the VM. The other one is pauseless GC where every instance field accesses might need to call to the GC to fix up the pointer. In both cases, the load have side affects and we need new opcodes to represent this situation.

The **x**rdbar/**x**rdbari need to represent both the side effects of the load as well as the value of the load. They are the same as the corresponding **x**load/**x**loadi except that read barrier nodes need to be anchored under a treetop or chk at their point of first evaluation (like a call). Then they can be referenced in subsequent trees.  

The children and symbol of **indirect** read barriers are the same as their corresponding load. It would be something like:

```
treetop
  xrdbari <#xxx A.instanceB>
    aload <auto 3>
...
  ==>xrdbari
```

The children and symbol of **direct** read barriers are the same as their corresponding load with one additional child which is static field class. The extra child is for simplifying code evaluation for unresolved case where the load of unresolved static field class hangs under a treetop. 
```
resolved case:
treetop
  xrdbar <#xxx A.staticB>
     Class of A
...
  ==>xrdbar

unresolved case:
ResolveCHK
    loadaddr  Class of A (unresolved static)
ResolveCHK
    irdbar
      ==>loadaddr
...
  ==>xrdbar
```

#2875

mstoodle:
Presumably these are typed like `xload`/`xloadi` ?

cathyzhyi:
@mstoodle yes. updated the description.

cathyzhyi:
Documenting here the discussion at the architecture meeting #2817 about the new opcode
 - [ ] the opcode itself should be able to be commoned or moved and the optimizations applies to load should be able to apply to rdbar as well.
- [ ] the only difference between the rdbar and loads is that rdbar has to hang under a treetop before the first use to represent the side effect it's gonna have on evalutor. We want to avoid the complication of dealing with a call in the middle of evaluating another treetop.
- [ ] If there is any restrictions on optimization coming from different modes, debugging mode etc, they should be addressed under different mode.  The optimizations like LocalCSE, PRE, loop versioner might need to special case rdbar under those different modes. The special case should be grouped into properties rather than adding another | to a group of other opcode values. Aliasing can be different depending on modes as well.
[...]

PR #2875, opened 2018-08-17
Add rdbar/rdbari opcodes

cathyzhyi:
The xrdbar/xrdbari opcode are added to represent both the value of the
load as well as its side effects like needing to notify the GC of the
read, notify the VM of the read or other similar activities.

This PR adds the most basic supports for read barrier opcodes. By default
read barriers use load handlers and evaluators.

A treetop of indirect read barriers looks like the following:

treetop
&nbsp;&nbsp;xrdbari <#xxx A.instanceB>
           &nbsp;&nbsp;&nbsp;&nbsp;aload <auto 3>
...
&nbsp;&nbsp;==>xrdbari

A treetop of direct read barriers looks like the following:

treetop
&nbsp;&nbsp;xrdbar <#xxx A.staticB>
&nbsp;&nbsp;&nbsp;&nbsp;Class of A
...
&nbsp;&nbsp;==>xrdbar

Refer to #2812 for more design details


compiler/arm/codegen/TreeEvaluatorTable.hpp
```diff
@@ -38,13 +38,27 @@
...
    TR::TreeEvaluator::bloadEvaluator,       // TR::bloadi
    TR::TreeEvaluator::sloadEvaluator,       // TR::sloadi
    TR::TreeEvaluator::lloadEvaluator,       // TR::lloadi
+   TR::TreeEvaluator::iloadEvaluator,       // TR::irdbar
+   TR::TreeEvaluator::floadEvaluator,       // TR::frdbar
+   TR::TreeEvaluator::dloadEvaluator,       // TR::drdbar
+   TR::TreeEvaluator::aloadEvaluator,       // TR::ardbar
+   TR::TreeEvaluator::bloadEvaluator,       // TR::brdbar
+   TR::TreeEvaluator::sloadEvaluator,       // TR::srdbar
+   TR::TreeEvaluator::lloadEvaluator,       // TR::lrdbar
```
Leonardo2718: Should all these comments have an "i" at the end (i.e. `TR::irdbar` &rarr; `TR::irdbari`)?

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -376,6 +488,118 @@
...
+   {
+   /* .opcode               = */ TR::drdbari,
+   /* .name                 = */ "drdbari",
+   /* .properties1          = */ ILProp1::LoadVar | ILProp1::Indirect | ILProp1::HasSymbolRef,
+   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::SupportedForPRE| ILProp2::MayUseSystemStack | ILProp2::ReadBarrierLoad,
+   /* .properties3          = */ ILProp3::LikeUse,
+   /* .properties4          = */ 0,
+   /* .dataType             = */ TR::Double,
+   /* .typeProperties       = */ ILTypeProp::Size_8 | ILTypeProp::Floating_Point,
+   /* .childProperties      = */ ONE_CHILD(TR::Address),
```
fjeremic: Shouldn't the child here be a double? Similarly for others. This is likely a copy/paste mistake. Can we fix all of them?

compiler/il/ILProps.hpp
```diff
@@ -210,7 +210,7 @@ namespace ILProp2
       ResolveCheck                 = 0x00000040,
       BndCheck                     = 0x00000080,
       CheckCast                    = 0x00000100,
-      // Available                 = 0x00000200,
+      ReadBarrierLoad              = 0x00000200,
```
fjeremic: Why do we need this exactly? Isn't this implied by the opcode?

compiler/il/OMRILOpCodesEnum.hpp
```diff
@@ -53,13 +53,32 @@
    bload,    // load byte
    sload,    // load short integer
    lload,    // load long integer
+
+   //Read barrier is used to represent loads with side effects like check for GC, debugging etc.
+   //It is the same as the corresponding load except that it needs to be anchored under a
+   //treetop. The children and symbol of a read barrier are the same as the corresponding load.
```
fjeremic: @0xdaryl do we have any template for how to properly document IL? IMO we should add some structure to the documentation.

compiler/p/codegen/TreeEvaluatorTable.hpp
```diff
@@ -38,13 +38,27 @@
...
+   TR::TreeEvaluator::floadEvaluator,                   // TR::frdbar
+   TR::TreeEvaluator::dloadEvaluator,                   // TR::drdbar
+   TR::TreeEvaluator::aloadEvaluator,                   // TR::ardbar
+   TR::TreeEvaluator::bloadEvaluator,                   // TR::brdbar
+   TR::TreeEvaluator::sloadEvaluator,                   // TR::srdbar
+   TR::TreeEvaluator::lloadEvaluator,                   // TR::lrdbar
    TR::TreeEvaluator::iloadEvaluator,                   // TR::iloadi
    TR::TreeEvaluator::floadEvaluator,                   // TR::floadi
    TR::TreeEvaluator::dloadEvaluator,                   // TR::dloadi
    TR::TreeEvaluator::aloadEvaluator,                   // TR::aloadi ibm@59591
```
fjeremic: While we're in this area can we remove the "ibm@59591" whatever that is?

compiler/il/ILProps.hpp
cathyzhyi: It's for later changes in optimizer when we want to special case for this entire group of read barrier opcodes. In dead tree elimination for example, we need to keep the anchoring tree top if it's rdbars. This also mirrors the WriteBarrierStore we already have.

cathyzhyi:
@andrewcraik the more detailed description is in the original issue. Should I copy the description to this PR as well?

compiler/p/codegen/TreeEvaluatorTable.hpp
cathyzhyi: removed!

andrewcraik:
I don't think the whole discussion needs to be here, but it would be nice to see something in the PR/commit saying these are used to handle read operations which may have side-effects - for example, needing to notify the GC of the read, notify the VM of the read or other similar activities. Just makes the pull more self contained for someone searching history.

compiler/il/ILOpCodeProperties.hpp
```diff
@@ -264,6 +264,118 @@
    /* .ifCompareOpCode      = */ TR::BadILOp,
    },
 
+   {
+   /* .opcode               = */ TR::irdbar,
+   /* .name                 = */ "irdbar",
+   /* .properties1          = */ ILProp1::LoadVar | ILProp1::HasSymbolRef,
+   /* .properties2          = */ ILProp2::ValueNumberShare | ILProp2::SupportedForPRE| ILProp2::MayUseSystemStack | ILProp2::ReadBarrierLoad,
```
andrewcraik: Does PRE really handle these?

cathyzhyi: PRE hasn't been updated to include the support yet. Another PR will address optimizer changes to support this opcode but as a property for the opcode all the optimizations applied to loads should be able to be applied to rdbars.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -754,6 +754,36 @@ TR::Register *OMR::X86::TreeEvaluator::sloadEvaluator(TR::Node *node, TR::CodeGe
    return reg;
    }
 
+TR::Register *iReadBarrierEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   return OMR::X86::TreeEvaluator::iloadEvaluator(node,cg);
```
0dvictor: It may be better to use `TR::TreeEvaluator::iloadEvaluator` so that it can reach the `iloadEvaluator` overridden by downstream projects.

compiler/x/amd64/codegen/OMRTreeEvaluator.cpp
```diff
@@ -126,6 +126,11 @@ TR::Register *OMR::X86::AMD64::TreeEvaluator::lloadEvaluator(TR::Node *node, TR:
    return reg;
    }
 
+TR::Register *OMR::X86::AMD64::TreeEvaluator::lReadBarrierEvaluator(TR::Node *node, TR::CodeGenerator *cg)
```
0dvictor: This evaluator may go into common x86 code, it is same as the 32-bit one.

compiler/x/i386/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3203,6 +3203,11 @@ TR::Register *OMR::X86::I386::TreeEvaluator::lloadEvaluator(TR::Node *node, TR::
    return reg;
    }
 
+TR::Register *OMR::X86::I386::TreeEvaluator::lReadBarrierEvaluator(TR::Node *node, TR::CodeGenerator *cg)
```
0dvictor: This evaluator may go into common x86 code, it is same as the 64-bit one.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -754,6 +754,36 @@ TR::Register *OMR::X86::TreeEvaluator::sloadEvaluator(TR::Node *node, TR::CodeGe
    return reg;
    }
 
+TR::Register *iReadBarrierEvaluator(TR::Node *node, TR::CodeGenerator *cg)
```
0dvictor: Do you mean `OMR::X86::TreeEvaluator::iReadBarrierEvaluator` ?

compiler/x/codegen/OMRTreeEvaluator.cpp
cathyzhyi: changed according to the comments.

compiler/x/amd64/codegen/OMRTreeEvaluator.cpp
cathyzhyi: changed according to the comments

compiler/x/i386/codegen/OMRTreeEvaluator.cpp
cathyzhyi: changed according to the comments.

compiler/x/codegen/OMRTreeEvaluator.cpp
cathyzhyi: thanks for catch this. changed accordingly.

0dvictor:
FYI, https://github.com/eclipse/omr/pull/2935 is required by this changeset. An existing bug prevent `iReadBarrierEvaluator` being correctly executed.

compiler/x/amd64/codegen/TreeEvaluatorTable.hpp
```diff
@@ -35,17 +35,31 @@
    TR::TreeEvaluator::iloadEvaluator,                   // TR::iload
    TR::TreeEvaluator::floadEvaluator,                   // TR::fload
    TR::TreeEvaluator::dloadEvaluator,                   // TR::dload
-   TR::TreeEvaluator::aloadEvaluator,                 // TR::aload
+   TR::TreeEvaluator::aloadEvaluator,                   // TR::aload
    TR::TreeEvaluator::bloadEvaluator,                   // TR::bload
    TR::TreeEvaluator::sloadEvaluator,                   // TR::sload
-   TR::TreeEvaluator::lloadEvaluator,                 // TR::lload
+   TR::TreeEvaluator::lloadEvaluator,                   // TR::lload
+   TR::TreeEvaluator::iReadBarrierEvaluator,            // TR::irdbar
```
fjeremic: This is a little unorthodox naming scheme. I believe the convention is to name the evaluators `<ilOpcode>Evaluator` across all the codegens so it is easy to grep around and search for the implementations. This seems to deviate from the naming scheme of all the other IL evaluators (see above and below these ones for example).

cathyzhyi: thanks for the suggestion! modified accordingly.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -754,6 +754,55 @@ TR::Register *OMR::X86::TreeEvaluator::sloadEvaluator(TR::Node *node, TR::CodeGe
    return reg;
    }
 
+TR::Register *OMR::X86::TreeEvaluator::irdbarEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   if (node->getSymbolReference()->getSymbol()->isStatic())
+      cg->decReferenceCount(node->getFirstChild());
```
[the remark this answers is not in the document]
cathyzhyi: For direct rdbar there is one more child (the static field class) than the normal load so the load evaluator called after this won't know about that child and decrease its reference count.

compiler/x/i386/codegen/TreeEvaluatorTable.hpp
```diff
@@ -39,13 +39,27 @@
    TR::TreeEvaluator::bloadEvaluator,                   // TR::bload
    TR::TreeEvaluator::sloadEvaluator,                   // TR::sload
    TR::TreeEvaluator::lloadEvaluator,                  // TR::lload
-   TR::TreeEvaluator::iloadEvaluator,                   // TR::iloadi
-   TR::TreeEvaluator::floadEvaluator,                   // TR::floadi
-   TR::TreeEvaluator::dloadEvaluator,                   // TR::dloadi
+   TR::TreeEvaluator::irdbarEvaluator,                 // TR::irdbar
+   TR::TreeEvaluator::frdbarEvaluator,                 // TR::frdbar
+   TR::TreeEvaluator::drdbarEvaluator,                 // TR::drdbar
+   TR::TreeEvaluator::ardbarEvaluator,                 // TR::ardbar
```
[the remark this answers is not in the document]
cathyzhyi: It's just the design for static rdbar changed a little bit (because of the extra child) so I wanted to wait until we agree on the x evaluators first and apply them to other platforms.

andrewcraik:
@Leonardo2718 @fjeremic can you please have a look at this again now that I believe a number of your comments have been addressed? This has been sitting for a bit and we need to keep it moving forward.

fjeremic:
@genie-omr build all

PR #2876, opened 2018-08-20
Avoid amount in heap-resize tag due to idle tuning

ashu-mehra:
During idle phase when GC de-commits free heap page, the amount of
memory de-committed is not correctly calculated. As it gets displayed
in gc logs as part of heap-resize tag, it can cause confusion.
So its better to not display this value in the heap-resize tag.

Issue https://github.com/eclipse/openj9/issues/2312

### DIFF
--- a/gc/verbose/VerboseHandlerOutput.cpp
+++ b/gc/verbose/VerboseHandlerOutput.cpp
@@ -926,7 +926,12 @@ MM_VerboseHandlerOutput::outputHeapResizeInfo(MM_EnvironmentBase *env, uintptr_t
 
 	getTagTemplate(tagTemplate, sizeof(tagTemplate), omrtime_current_time_millis());
 
-	writer->formatAndOutput(env, indent, "<heap-resize id=\"%zu\" type=\"%s\" space=\"%s\" amount=\"%zu\" count=\"%zu\" timems=\"%llu.%03llu\" reason=\"%s\" %s />", id, resizeTypeName, getSubSpaceType(subSpaceType), resizeAmount, resizeCount, timeInMicroSeconds / 1000, timeInMicroSeconds % 1000, reasonString, tagTemplate);
+	if (HEAP_RELEASE_FREE_PAGES == resizeType) {
+		/* Do not display 'resizeAmount' as it is not accurate for HEAP_RELEASE_FREE_PAGES type */
+		writer->formatAndOutput(env, indent, "<heap-resize id=\"%zu\" type=\"%s\" space=\"%s\" count=\"%zu\" timems=\"%llu.%03llu\" reason=\"%s\" %s />", id, resizeTypeName, getSubSpaceType(subSpaceType), resizeCount, timeInMicroSeconds / 1000, timeInMicroSeconds % 1000, reasonString, tagTemplate);
+	} else {
+		writer->formatAndOutput(env, indent, "<heap-resize id=\"%zu\" type=\"%s\" space=\"%s\" amount=\"%zu\" count=\"%zu\" timems=\"%llu.%03llu\" reason=\"%s\" %s />", id, resizeTypeName, getSubSpaceType(subSpaceType), resizeAmount, resizeCount, timeInMicroSeconds / 1000, timeInMicroSeconds % 1000, reasonString, tagTemplate);
+	}
 	writer->flush(env);
 }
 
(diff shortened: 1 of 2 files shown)


amicic:
in schema.xml, you'll need to update heap-resize tag to have amount attribute as optional (currently it's marked are required)

ashu-mehra:
@amicic updated the schema as suggested.

gc/verbose/VerboseHandlerOutput.cpp
```diff
@@ -926,7 +926,12 @@ MM_VerboseHandlerOutput::outputHeapResizeInfo(MM_EnvironmentBase *env, uintptr_t
 
 	getTagTemplate(tagTemplate, sizeof(tagTemplate), omrtime_current_time_millis());
 
-	writer->formatAndOutput(env, indent, "<heap-resize id=\"%zu\" type=\"%s\" space=\"%s\" amount=\"%zu\" count=\"%zu\" timems=\"%llu.%03llu\" reason=\"%s\" %s />", id, resizeTypeName, getSubSpaceType(subSpaceType), resizeAmount, resizeCount, timeInMicroSeconds / 1000, timeInMicroSeconds % 1000, reasonString, tagTemplate);
+	if (HEAP_RELEASE_FREE_PAGES == resizeType) {
+		/* Do not display 'resizeAmount' as it is not accurate for HEAP_RELEASE_FREE_PAGES type */
```
charliegracie: Why is resizeAmount not accurate for HEAP_RELEASE_FREE_PAGES?

charliegracie: is this call being made because we have resized the heap? If yes then we know exactly how much we reduced or increased the heap size by.  If we are using this to talk about requesting the OS to release memory for previously freed memory (and the heap size is not changing) then we need a new verbose stanza completely as that is not what this is meant for.

amicic: yes, it's sort of abused stanza/tag. it has however its own type that is printed 'release free pages' or so, as opposed to 'expand' or 'contract'. now without 'amount' being reported, it serves just to report the event of some physical memory being released due to app idleness. it could indeed be a separate tag - I personally have a strong opinion if it should

charliegracie: ok makes sense.... I missed that it had its own tag

amicic: Here is a sample stanza from the parent Issue. Reminded me, it also has unique reason:

<heap-resize id="759" type="release free pages" space="tenure" amount="1261850624" count="1" timems="76.973" reason="idle" timestamp="2018-07-11T15:04:24.175" />

If we remove 'amount' here, we should report it elsewhere (tracepoint).

But I'm not completely convinced we should remove it - another posible alternative (to removing or 'fixing') is just to leave as is and document it: it's a superset, pages that we are releasing, including the ones that might have already been released. It's still valuable info so we can tell what percentage of the heap is unused.

charliegracie: ok I guess I need more details on the value printed in `amount` right now for the `release free pages`? Is it the amount of memory that was decommitted during this resize? Is the confusion that the size may include memory that was previous contracted?

ashu-mehra: @charliegracie whenever we do `heap-resize` with type as `release free pages`, we traverse all entries in the free list and de-commit the pages. Code is present [here](https://github.com/eclipse/omr/blob/91adae92b06795822df6a92435e2763f21d4bec2/gc/base/MemoryPoolAddressOrderedListBase.cpp#L189). We don't know whether the pages being de-committed have a mapping in physical memory or not. If they don't map to physical pages, then de-committing such pages wouldn't have any effect on the memory footprint.
So, the confusion arises because the `amount` is treated as the memory de-committed and is expected to indicate reduction in footprint, which may not be case.

charliegracie: I think having no value is even more confusing.  If we document what the value means when the type is `heap free pages` then I think the current value is ok.

amicic: I would document at both places. For a start, at the place you just referred.

I'd also add it to the verbose GC docs. I don't know if we have any for OpenJ9 yet, but I'd start with IBM one's, and eventually all should be propagated to OpenJ9, too.

I'd open a whole new section here:
https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/mm_gc_pd_verbosegc.html

and call it heap resizing and document all existing types (I can help with the content) and than add the one in question here.

charliegracie: Any update @ashu-mehra

ashu-mehra: I guess if we have decided to just document this behavior, then this PR is no longer required.
I will close it.

PR #2877, opened 2018-08-20
Do GC on idle every time

ashu-mehra:
Currently JVM goes into idle state, it performs GC only if the amount
of memory allocated in the java heap since last GC cycle is above a
threshold specified by 'gcOnIdleRatio'.
This is very conservative approach. Since JVM is going into idle state
it can be aggressive in doing GC.
With this change JVM would do GC unconditionally everytime it goes into
idle state.

Issue https://github.com/eclipse/openj9/issues/2312

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -724,7 +724,6 @@ public:
 #if defined(OMR_GC_IDLE_HEAP_MANAGER)
 	uintptr_t idleMinimumFree;   /**< percentage of free heap to be retained as committed, default=0 for gencon, complete tenture free memory will be decommitted */
 	uintptr_t lastGCFreeBytes;  /**< records the free memory size from last Global GC cycle */
-	uintptr_t gcOnIdleRatio; /**< the percentage of allocation since the last GC allocation determines the invocation of global GC, default global GC is invoked if allocation is > 20% */
 	bool gcOnIdle; /**< Enables releasing free heap pages if true while systemGarbageCollect invoked with IDLE GC code, default is false */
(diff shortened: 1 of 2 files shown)


amicic:
I don't have a strong opinion on this. It will vary from scenario to scenario. 

Perhaps, we could introduce a heuristic like: do system GC only if there at least 1 (or 3) global GCs since the previous event of VM going into the idle state.  But then, it's just more complexity in the code, that may make sense to some-one/some-workload and not to the others.

@charliegracie

mstoodle:
another option might be to make it dynamically configurable by the runtime, and OpenJ9 can simply configure it to be its default behaviour. Of course, that leaves the question of what the default should be for OMR...

dmitripivkine:
I believe this change is good. It eliminates extra complexity (which is not so effective as expected) introduced for Heap Idle Manager and returns back to basics.

charliegracie:
Can you provide more background on this feature again? So the idle detection in the JVM was triggering a system.gc() call and the system.gc() call was doing checks to see if it should actually do the GC? Now you are proposing that if the idle detection calls system.gc() you just want the system.gc() to happen?

ashu-mehra:
> Can you provide more background on this feature again?

Initially we decided to do GC during idle only if the application has allocated certain amount heap memory since last GC cycle. We set it to 20%, that is if more than 20% of tenure heap is allocated since last heap, we would do GC. But this has not proved to be a very good heuristic, especially for simple programs that people write out to test idle tuning feature.
In addition, since we are in idle state, it shouldn't hurt to do GC unconditionally.
@dmitripivkine do you want to add anything more to it?

dmitripivkine:
this change looks good to me

charliegracie:
@genie-omr build all

PR #2880, opened 2018-08-21
Allow relocation of ClassUnload assumptions

dmitry-ten:
AOT and JITaaS require assumptions to be relocated.
Add a new kind of `ExternalRelocation` to relocate ClassUnload assumptions.
Add a new query to `OMRClassEnv` to check whether relocation is needed.

Signed-off-by: Dmitry Ten <Dmitry.Ten@ibm.com>


dmitry-ten:
This is the omr part of eclipse/openj9#2672

dmitry-ten:
@ymanton could you review this, please?

compiler/x/codegen/DataSnippet.cpp
```diff
@@ -48,19 +49,29 @@ void
 TR::IA32DataSnippet::addMetaDataForCodeAddress(uint8_t *cursor)
    {
    // add dummy class unload/redefinition assumption.
+   bool needRelocation = cg()->comp()->compileRelocatableCode() || TR::Compiler->cls.classUnloadAssumptionNeedsRelocation();
```
ymanton: Can we simply use one query here? Can compilers doing AOT also return true for `classUnloadAssumptionNeedsRelocation()`?

compiler/x/codegen/DataSnippet.cpp
```diff
@@ -48,19 +49,29 @@ void
...
+                               __FILE__, __LINE__, self()->getNode());
+      }
+
    if (_isClassAddress)
       {
       if (TR::Compiler->target.is64Bit())
          {
-         cg()->jitAddPicToPatchOnClassUnload((void*)-1, (void *) cursor);
+         if (!needRelocation)
+            cg()->jitAddPicToPatchOnClassUnload((void*)-1, (void *) cursor);
```
ymanton: I don't know if this will work as intended. IIRC it's possible to AOT-compile a method at runtime and then run it without processing the relocations. If that's still possible then you will need to register the assumption because the relocation may not be processed. Perhaps @dsouzai knows if that's still possible.

dsouzai: Even on the AOT compile run we go through and process the relocations; in fact I think some code depends on the relocations being processed because, for example, when a `lea` gets converted to a `mov + lea`, the immediate value of the `mov` will be garbage (even on the AOT compile run) until it gets relocated.

compiler/x/codegen/DataSnippet.cpp
```diff
@@ -50,17 +51,27 @@ TR::IA32DataSnippet::addMetaDataForCodeAddress(uint8_t *cursor)
    // add dummy class unload/redefinition assumption.
    if (_isClassAddress)
       {
```
mstoodle: this doesn't seem right to me. If there is supposedly a class address written at cursor, then why shouldn't we be creating some kind of `ArbitraryClassAddress` or `ClassObject` relocation? Typically you would create a relocation for the class pointer and that would take care of creating the class unload assumption at AOT load time.

There are other implementations of `addMetaDataForCodeAddress` sprinkled around the x86 code generator that do not require a new relocation record type to be created and that follow the model I outlined above.

ymanton: I think in this case the class address is "right", i.e. it was communicated to the compiler that generated the code by the current compiler that's loading the method and so the only thing necessary to relocate the address is to register an unload assumption against it. I'm not sure if that means we can reuse one of the existing relocations and just write have the loading compiler write the same value to the location or if that has other repercussions.

dmitry-ten: So, after investigating the option of relocating the class pointer, I think it's not possible to
do from this data snippet, because the node, at which the relocation happens, is not related
to the class node. E.g.  older version of OpenJ9 used to create snippet for `instanceof` node.
 Thus we do not know anything about the class that will be stored in the snippet.

@dsouzai says that creating this kind of class pointer relocation will be possible with one of the future changes. So maybe this change should be postponed until then.

dmitry-ten:
Comments have been addressed

mstoodle:
Still unfortunate we have to do relocations this way (we ultimately need to move this stuff all the way into OpenJ9), but the change looks reasonable enough to me. I'll just do a full run of tests and if that passes will merge. ( genie-omr build all)

Thanks @dmitry-ten

PR #2881, opened 2018-08-22
Disable OSRLiveRangeAnalysis when mimic interpreter

cathyzhyi:
Mimic interpreter frame shape doesn't need to zero out dead slots after
OSR out and therefore OSRLiveRangeAnalysis is not needed.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -729,10 +729,16 @@ bool TR_OSRLiveRangeAnalysis::shouldPerformAnalysis()
    else if (comp()->getOSRMode() == TR::involuntaryOSR)
       {
       static const char* disableOSRPointDeadslotsBookKeeping = feGetEnv("TR_DisableOSRPointDeadslotsBookKeeping");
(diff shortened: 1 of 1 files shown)


andrewcraik:
makes sense to me - @vijaysun-omr could you review?

vijaysun-omr:
Is the reason for not requiring this analysis to run if mimic interpreter shape is set is that there are no explicit exits to paths that do an OSR transition out from compiled code in this mode ? Could you please elaborate the rationale for the change so that it's documented here even for someone who is'nt very familiar with our OSR technology ?

vijaysun-omr:
@genie-omr build all

cathyzhyi:
Mimic Interpreter frame shape also executes OSR exit path to populate OSR buffer because VM needs it to lift constraint related to method monitor temp. 

`OSRLiveRangeAnalysis` is for book keep dead autos in bytecode to prevent the case where optimizer optimizing out dead stores leaving some autos uninitialized and crashing the VM while debugging. Mimic interpreter frame shape doesn't need this because the stack slots are always the same as interpreter and the optimizer is prevent from removing dead stores and leaving an auto uninitialized.

vijaysun-omr:
Looks like a safe change to me. Merging.

PR #2882, opened 2018-08-22
Deprecate many unused Z codegen APIs

fjeremic:
- Deprecate isCallSnippet() from the common code generator
- Deprecate setUpArgumentsInRegister as it is not used anywhere
- Deprecate loadArgumentItem as it is not used anywhere
- Deprecate string support in S390ConstantDataSnippet
- Deprecate isRefed in S390ConstantInstructionSnippet
- Deprecate S390(TargetAddress|LookupSwitch)Snippet
- Deprecate S390LabelTableSnippet as it is not being used
- Deprecate S390LabelTableSnippet as it is not being used
- Remove several undeclared functions
- Remove many unused defines in OMRCodeGenerator
- Deprecate (get|set)NodeAddressOfCachedStaticTree
- Deprecate getBucketPlusIndexRegisters
- Deprecate (get|set)CurrentDEPEND as it is unused
- Deprecate labelHashTable as it is unused
- Deprecate (get|set)ARegisterIterator as it is unused
- Simplify supportsTrapsInTMRegion
- Deprecate biasDecimalFloatFrac as it is unused
- Deprecate isMemcpyWithPadIfFoldable as it is unused
- Reinstate block and unblock 


fjeremic:
I'm trying to keep things manageable. This is the first of several PRs of cleanup. Once all the cruft is gone out of the codegen there will be some major refactoring and a lot of documentation added.

vijaysun-omr:
@genie-omr build all

compiler/ras/Debug.hpp
```diff
@@ -141,7 +141,6 @@ namespace TR { class X86FPMemRegInstruction;               }
 namespace TR { class X86FPRegMemInstruction;               }
 class TR_X86RegisterDependencyGroup;
 namespace TR { class X86RestartSnippet; }
-namespace TR { class S390LookupSwitchSnippet; }
```
vijaysun-omr: Could you please elaborate on what a target address snippet or lookup switch snippet was envisaged for ? I know these concepts are not being used in the codebase currently (hence your deletion of this code) but a different question is whether they could be useful down the line. Is there a short reason why you feel the notions are themselves not useful ?

fjeremic: It seems to have targeted a key/value switch table, likely to compare against certain keys and jump to certain addresses. It required some infrastructure from a specific node which as far as I can tell no longer exists in OMR.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -2527,21 +2493,11 @@ OMR::Z::MemoryReference::getSnippet()
       {
       self()->setMemRefAndGetUnresolvedData(snippet);
       }
-   else // add writable data snippet address relocation
-   if (self()->getConstantDataSnippet() != NULL)
+   // add writable data snippet address relocation
```
mstoodle: nit: shouldn't comment be inside the conditional?

fjeremic: I don't think these comments should exist at all. They do not provide any additional information than can be implied by the source code lines following them. Fixed in 9133acf

fjeremic:
@genie-omr build all

mstoodle:
@vijaysun-omr is away for a week, so I'll take over this "assignment" to merge. I don't believe removing the two comments will materially affect the test results, but since @fjeremic started them I will wait for them to finish before merging. Thanks for the hard work to push this one through @fjeremic !

fjeremic:
Looks like the TravisCI build failed due to an `apt-get`. I'd say this PR should be ready to go @mstoodle.

mstoodle:
Yes. I restarted it and it passed. Merging now.

PR #2883, opened 2018-08-22
Fix incorrect usage of TR::Options

harryyu1994:
- the per-compilation TR::Options should be used instead of the global TR::Options during the compilation stages (i.e. Optimizer, CodeGen) 
- this problem was uncovered in a downstream project where the global and the per-compilation TR::Options were significantly different and the per-compilation TR::Options was heavily relied on during Optimization & Code Generation Stage.
- this commit was actually merged but reverted due to a performance regression (DayTrader3 on X platform)
- the problematic line was excluded from this commit

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -326,7 +326,7 @@ OMR::Compilation::Compilation(
    // while the PersistentMethodInfo is allocated during codegen creation
    // Access to this list must be performed with assumptionTableMutex in hand
    //
-   if (!TR::Options::getCmdLineOptions()->getOption(TR_DisableFastAssumptionReclamation))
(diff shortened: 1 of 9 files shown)


harryyu1994:
The problematic change from last time:
```
// Temporarily exclude PPC due to perf regression
      if( (self()->getMethodHotness() <= warm))
         {
         if (!TR::Compiler->target.cpu.isPower())
            TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);
         }
      self()->getOptions()->setOption(TR_DisablePartialInlining);
```
When `TR::Options::getCmdLineOptions()` was changed to `self()`, it causes a performance regression.

mstoodle:
@genie-omr build all

PR #2884, opened 2018-08-22
Change insert prefetch to insert BPP after ICF

dhong44:
Change insert prefetch to insert BPP after ending internal
control flow, this will prevent BPP from being inserted
within a ICF.


fjeremic:
@genie-omr build zlinux,zos

PR #2889, opened 2018-08-23
Use port library for printing types

keithc-ca:
Previously, `printf` was used directly which triggers a warning from VS2017 about providing a value of type `size_t` for a format `%ul` which expects `unsigned long`.

> printf("union '%s' size(%lu) {\n", type->_name.c_str(), type->_sizeOf);

Using the port library allows us to use the format `%zu` for values of type `size_t`.


charliegracie:
@genie-omr build all

PR #2892, opened 2018-08-24
Rename IA32DataSnippet and IA32ConstantDataSnippet

### LINKED ISSUE #2792: Rename IA32DataSnippet and IA32ConstantDataSnippet
IA32DataSnippet and IA32ConstantDataSnippet are actually shared by both X86-32 and X86-64 CodeGen; however, the "IA32" in their names is misleading as "IA32" usually means X86-32. Ideally, they should be renamed to X86DataSnippet and X86ConstantDataSnippet respectively.

0dvictor:
IA32DataSnippet and IA32ConstantDataSnippet are actually shared by both X86-32 and X86-64 CodeGen;
however, the "IA32" in their names is misleading as "IA32" usually means X86-32.
Therefore, they are renamed to X86DataSnippet and X86ConstantDataSnippet respectively.

Closes #2792

### DIFF
--- a/compiler/x/amd64/codegen/MemoryReference.hpp
+++ b/compiler/x/amd64/codegen/MemoryReference.hpp
@@ -28,7 +28,7 @@
 #include <stdint.h>        // for uint8_t
 #include "env/jittypes.h"  // for intptrj_t
 
-namespace TR { class IA32DataSnippet; }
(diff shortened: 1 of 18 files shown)


mstoodle:
change looks ok to me, let's see how the testing goes

PR #2895, opened 2018-08-28
Fix Struct/Union fieldName resolution

charliegracie:
If a struct of union contains fields whose name are a sub string
of another field TypeDictionary will return the wrong field for
one of the names. The fix is to use strcmp instead of strncmp to
ensure the strings are exactly the same.

I have added a simple test for each of these that verifies the
types of each field returned by GetFieldType and GetUnionFieldType
matches the declared type for that field.

### DIFF
--- a/fvtest/jitbuildertest/FieldNameTest.cpp
+++ b/fvtest/jitbuildertest/FieldNameTest.cpp
@@ -1,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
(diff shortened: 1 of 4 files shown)


fvtest/jitbuildertest/FieldNameTest.cpp
```diff
@@ -0,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2017 IBM Corp. and others
```
Leonardo2718: Please update the copyright year :slightly_smiling_face:

charliegracie: Fixed

charliegracie:
@genie-omr build all

PR #2898, opened 2018-08-28
Fix 'error=class-memaccess' warnings with gcc 8

keithc-ca:
Address diagnostics of the form:
```
error: 'void* memset(void*, int, size_t)' clearing an object with no trivial copy-assignment; use assignment or value-initialization instead [-Werror=class-memaccess]
```

### DIFF
--- a/ddr/tools/blob_reader/blob_reader.cpp
+++ b/ddr/tools/blob_reader/blob_reader.cpp
@@ -148,6 +148,17 @@ struct Structure {
 	uint32_t size;
 	vector<Field *> fields;
 	vector<Constant *> constants;
+
+	Structure()
+		: name(NULL)
+		, nameLength(0)
+		, superName(NULL)
+		, superNameLength(0)
+		, size(0)
+		, fields()
+		, constants()
+	{
+	}
 };
 
 bool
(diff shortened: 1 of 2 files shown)


ddr/tools/blob_reader/blob_reader.cpp
```diff
@@ -294,7 +294,7 @@ main(int argc, char *argv[])
 			currentStruct += sizeof(BlobStruct);
 
 			Structure *builtStruct = (Structure *)malloc(sizeof(Structure));
-			memset(builtStruct, 0, sizeof(Structure));
+			memset((void *)builtStruct, 0, sizeof(Structure));
```
charliegracie: I can not see why this code would generate that warning.  Structure is a simple Struct.... I thought this warning was for non-POD types

keithc-ca: I think the fields of type `vector` declassify it as a POD type.

keithc-ca:
Updated blob_reader.cpp so `Structure` has a constructor that we can use instead of `memset`.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

PR #2899, opened 2018-08-28
Remove deprecated X86ScratchArgHelperCallSnippet

0dvictor:
X86ScratchArgHelperCallSnippet has been long deprecated, removing it.


mstoodle:
Amusingly, it looks like the implementation for this snippet is in OpenJ9, but these vestigial references remain in OMR. But nothing is using the snippet even in OpenJ9, so it can be removed once these references go away (see https://github.com/eclipse/openj9/pull/2738).

PR #2903, opened 2018-08-29
Constant folding a node with new constraint added

liqunl:
In constrainCmpeqne, a node is folded into a constant without adding a
new constraint. The parent node's handler will not do certain
optimizations without a constraint on the child node.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -11039,17 +11039,12 @@ static TR::Node *constrainCmpeqne(OMR::ValuePropagation *vp, TR::Node *node, boo
    TR::VPConstraint *constraint = NULL;
    if (result >= 0)
       {
-      if ((lhsGlobal || vp->lastTimeThrough()) &&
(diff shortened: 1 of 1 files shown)


liqunl:
@jdmpapin Can I have a review?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -11039,17 +11039,12 @@ static TR::Node *constrainCmpeqne(OMR::ValuePropagation *vp, TR::Node *node, boo
    TR::VPConstraint *constraint = NULL;
    if (result >= 0)
       {
-      if ((lhsGlobal || vp->lastTimeThrough()) &&
-          performTransformation(vp->comp(), "%sChanging node [%p] %s into constant %d\n", OPT_DETAILS, node, node->getOpCode().getName(), result))
+      constraint = TR::VPIntConst::create(vp, result/*, isUnsigned*/);
+      if (lhsGlobal || vp->lastTimeThrough())
```
jdmpapin: Not necessarily a problem, but I don't think this `if` is needed anymore, since `replaceByConstant()` already checks the same condition before actually folding. At this point it should be fine to unconditionally `replaceByConstant()` and then return

mstoodle:
genie-omr build all

PR #2906, opened 2018-08-30
AArch64: Add constructors for ARM64AdminInstruction

knn-k:
This commit adds constructors with RegisterDependencyConditions for
ARM64AdminInstruction.

### DIFF
--- a/compiler/aarch64/codegen/ARM64Instruction.hpp
+++ b/compiler/aarch64/codegen/ARM64Instruction.hpp
@@ -711,8 +711,38 @@ class ARM64AdminInstruction : public TR::Instruction
     * @param[in] cg : CodeGenerator
     */
    ARM64AdminInstruction(TR::InstOpCode::Mnemonic op, TR::Node *node, TR::Node *fenceNode,
-                          TR::Instruction *precedingInstruction, TR::CodeGenerator *cg)
+                         TR::Instruction *precedingInstruction, TR::CodeGenerator *cg)
       : TR::Instruction(op, node, precedingInstruction, cg), _fenceNode(fenceNode)
+      {
+      }
+
+   /*
+    * @brief Constructor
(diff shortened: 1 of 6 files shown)


knn-k:
This is preparation for generating `ret` instruction with register dependency.

compiler/aarch64/codegen/ARM64Instruction.hpp
```diff
@@ -711,11 +711,41 @@ class ARM64AdminInstruction : public TR::Instruction
...
       : TR::Instruction(op, node, precedingInstruction, cg), _fenceNode(fenceNode)
       {
       }
 
+   /*
+    * @brief Constructor
+    * @param[in] op : instruction opcode
+    * @param[in] cond : register dependency conditions
+    * @param[in] node : node
+    * @param[in] fenceNode : fence node
```
0xdaryl: Can you provide more details on what the fence node is and what a consumer of this instruction is expected to provide?

knn-k: There are examples in omr/compiler/arm/codegen/OMRTreeEvaluator.cpp passing a fence node to `generateAdminInstruction()`.  The fence node comes from `TR::Node::createRelative32BitFenceNode()` .  The p version of OMRTreeEvaluator.cpp does it in a similar way.

0xdaryl: OK.  I studied this code a bit more and my confusion came from the overloaded nature of an "admin" instruction.  A fence node is only needed for (surprise!) fence instructions, and it's a fabricated node to hold the kind of relocation represented by the fence.  I believe using a node rather than some other structure is historical.  Anyway, I'm fine with this.

0xdaryl:
This is an AArch64-specific change, a genie build is not required.

PR #2909, opened 2018-08-30
Remove unused queries in CodeGen

0dvictor:
Following two queries are not used in anywhere, removing them:
    supportsDebugCounters
    incrementEventCounter


mstoodle:
genie-omr build all

PR #2910, opened 2018-08-30
Deprecate many unused Z codegen APIs - Part II

fjeremic:
This is a continuation of #2882.

- Deprecate isOperandLW as it is unused
- Deprecate S390OpProp_(Target|Src|Src2)LW
- Deprecate isBeginBlock as it is unused
- Deprecate TR_S390AutoMarkers and TR_S390CompareConventions
- Deprecate auto marker symbol in the optimizer
- Deprecate SpecialArgumentRegisters
- Fold several PLX related queries which always return false
- Deprecate StrictestAutoSymbolAlignment
- Fold alreadySaved(...) to always return false
- Deprecate useCachedStaticAreaAddresses
- Deprecate several unused functions in OMRLinkage
- Deprecate cachedStackRegisterForOutgoingArguments
- Fold getLengthStartForSSInstruction to be 1 always 


fjeremic:
@mstoodle could I please get a review for this one as well? There are more parts coming after.

compiler/z/codegen/OMRLinkage.hpp
```diff
@@ -89,12 +76,6 @@ enum TR_S390LinkageConventions
    TR_JavaPrivate          = 0x1,   ///< Java private linkage
    TR_JavaHelper           = 0x2,   ///< Java helper linkage
```
fjeremic: This `enum TR_S390LinkageConventions` shouldn't actually exist. We need to fold this enum into `TR_LinkageConventions` which is an extensible enum exactly for this purpose. I've opened up #2911 to track this.

fjeremic:
> genie-omr build all

Cool! Didn't know you didn't need the '@' before genie-omr.

compiler/z/codegen/OpMemToMem.cpp
```diff
@@ -450,7 +450,7 @@ MemToMemConstLenMacroOp::generateRemainder()
    uint64_t len = (uint64_t)_length;
    TR::Instruction * cursor = (_cursor == NULL ? _cg->getAppendInstruction() : _cursor);
 
-   if ((len >= (uint64_t)(_cg->getS390Linkage())->getLengthStartForSSInstruction()) && len > 0)
+   if (len >= 1)
```
mstoodle: Maybe set this into an appropriately named const uint64_t rather than leaving as the almost as generic-as-you-get magic constant '1' ?

fjeremic: I agree. Fixed in e46cb2d

fjeremic:
@genie-omr build all

mstoodle:
Assuming tests pass, I think this can be merged. @fjeremic feel free to ping me to merge if you happen to notice tests finishing before I do.

fjeremic:
@mstoodle I think this one should be ready.

PR #2916, opened 2018-08-31
Enable 64 bit registers on 32 bit for Linux on Z

dhong44:
Enable 64 bit registers on 32 bit for Linux on Z.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -701,7 +701,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableLateCleanFolding",            "O\tfold pdclean flags into pdstore nodes right before codegen",  SET_OPTION_BIT(TR_EnableLateCleanFolding), "F"},
    {"enableLinkagePreserveStrategy2",              "O\tenable linkage strategy 2", SET_OPTION_BIT(TR_LinkagePreserveStrategy2), "F"},
    {"enableLocalVPSkipLowFreqBlock",     "O\tSkip processing of low frequency blocks in localVP", SET_OPTION_BIT(TR_EnableLocalVPSkipLowFreqBlock), "F" },
-   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "T"},
+   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "F"},
    {"enableLongRegAllocationHeuristic",   "O\tenable heuristic for long register allocation",   SET_OPTION_BIT(TR_Enable64BitRegsOn32BitHeuristic), "F"},
    {"enableLoopEntryAlignment",            "O\tenable loop Entry alignment",                          SET_OPTION_BIT(TR_EnableLoopEntryAlignment), "F"},
    {"enableLoopVersionerCountAllocFences", "O\tallow loop versioner to count allocation fence nodes on PPC toward a profiled guard's block total", SET_OPTION_BIT(TR_EnableLoopVersionerCountAllocationFences), "F"},
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -701,7 +701,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableLateCleanFolding",            "O\tfold pdclean flags into pdstore nodes right before codegen",  SET_OPTION_BIT(TR_EnableLateCleanFolding), "F"},
    {"enableLinkagePreserveStrategy2",              "O\tenable linkage strategy 2", SET_OPTION_BIT(TR_LinkagePreserveStrategy2), "F"},
    {"enableLocalVPSkipLowFreqBlock",     "O\tSkip processing of low frequency blocks in localVP", SET_OPTION_BIT(TR_EnableLocalVPSkipLowFreqBlock), "F" },
-   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "F"},
+   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "T"},
    {"enableLongRegAllocationHeuristic",   "O\tenable heuristic for long register allocation",   SET_OPTION_BIT(TR_Enable64BitRegsOn32BitHeuristic), "F"},
    {"enableLoopEntryAlignment",            "O\tenable loop Entry alignment",                          SET_OPTION_BIT(TR_EnableLoopEntryAlignment), "F"},
    {"enableLoopVersionerCountAllocFences", "O\tallow loop versioner to count allocation fence nodes on PPC toward a profiled guard's block total", SET_OPTION_BIT(TR_EnableLoopVersionerCountAllocationFences), "F"},
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1189,7 +1189,6 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"traceLocalLiveVariablesForGC",     "L\ttrace local live variables for GC",            TR::Options::traceOptimization, localLiveVariablesForGC, 0, "P"},
    {"traceLocalReordering",             "L\ttrace local reordering",                       TR::Options::traceOptimization, localReordering, 0, "P"},
    {"traceLocalVP",                     "L\ttrace local value propagation",                TR::Options::traceOptimization, localValuePropagation, 0, "P"},
-   {"traceLongRegAllocationHeuristic",  "L\ttrace long register allocation",               TR::Options::traceOptimization, longRegAllocation, 0, "P"},
    {"traceLookahead",                   "O\ttrace class lookahead",                        SET_OPTION_BIT(TR_TraceLookahead), "P"},
    {"traceLoopAliasRefiner",            "L\ttrace loop alias refiner",                     TR::Options::traceOptimization, loopAliasRefiner, 0, "P"},
    {"traceLoopCanonicalization",        "L\ttrace loop canonicalization",                  TR::Options::traceOptimization, loopCanonicalization, 0, "P"},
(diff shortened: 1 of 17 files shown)


dhong44:
@fj

compiler/control/OMROptions.cpp
```diff
@@ -701,7 +701,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableLateCleanFolding",            "O\tfold pdclean flags into pdstore nodes right before codegen",  SET_OPTION_BIT(TR_EnableLateCleanFolding), "F"},
    {"enableLinkagePreserveStrategy2",              "O\tenable linkage strategy 2", SET_OPTION_BIT(TR_LinkagePreserveStrategy2), "F"},
    {"enableLocalVPSkipLowFreqBlock",     "O\tSkip processing of low frequency blocks in localVP", SET_OPTION_BIT(TR_EnableLocalVPSkipLowFreqBlock), "F" },
-   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "F"},
+   {"enableLongRegAllocation",            "O\tenable allocation of 64-bit regs on 32-bit",      SET_OPTION_BIT(TR_Enable64BitRegsOn32Bit), "T"},
```
fjeremic: I don't think this does what you may think it does. Changing the `"F"` to `"T"` does not make this option enabled by default. i.e. `comp()->getOption(TR_Enable64BitRegsOn32Bit)` will return false. This doesn't sound like what the commit title says is happening here.

I'm a little worried here. How did you test your change? Did you hardcode somewhere to enable this option manually?

compiler/z/codegen/OMRLinkage.cpp
```diff
@@ -1585,15 +1583,34 @@ OMR::Z::Linkage::copyArgRegister(TR::Node * callNode, TR::Node * child, TR::Regi
...
+         {
+         debugObj->addInstructionComment(toS390RRInstruction(cursor), REG_PARAM);
+         }
+
+      argRegister = self()->cg()->allocateConsecutiveRegisterPair(tempRegL, tempRegH);
+      self()->cg()->stopUsingRegister(argRegister);
+      }
+   else if (!self()->cg()->canClobberNodesRegister(child, 0))
+      {
+      if (!(TR::Compiler->target.is64Bit()) && child->getDataType() == TR::Int64)
```
fjeremic: Can we simplify this to:

```
if (TR::Compiler->target.is32Bit() && child->getDataType() == TR::Int64)
```

compiler/z/codegen/OMRRegisterDependency.cpp
```diff
@@ -1051,7 +1051,9 @@ TR_S390RegisterDependencyGroup::assignRegisters(TR::Instruction   *currentInstru
 
       dependentRegNum = _dependencies[i].getRealRegister();
 
-      if (dependentRegNum != TR::RealRegister::SpilledReg)
+      if (dependentRegNum != TR::RealRegister::SpilledReg &&
+          dependentRegNum != TR::RealRegister::KillVolHighRegs &&
+          dependentRegNum != TR::RealRegister::NoReg)
```
fjeremic: Can you explain this change?

compiler/control/OMROptions.cpp
dhong44: When testing the change I hardcoded `return true` into `use64BitRegsOn32Bit` but before opening this pull request and running my final tests I thought this option should be respected so I tried to modify the default value of this option. It looks like this would be incorrect anyway as this option should only be enabled by default on z. Since I had previously hardcoded return true some of my tests were still valid and I did have this working at one point but I would like to fix this and run some more tests on my own before this is merged. 

How would you suggest settings this option, should it be done in `J9CodeGenerator.cpp`, `OMRCodeGenerator.cpp`, or should I hard code return true?

fjeremic: I don't think the option is used on other platforms at all. You can confirm this. What I suggest we do is we rename the option to `TR_Disable64BitRegsOn32Bit` and change the code to be:

```
OMR::CodeGenerator::use64BitRegsOn32Bit()
   {
#ifdef TR_TARGET_S390
   if (TR::Compiler->target.is64Bit())
      return false;
   else
      {
      return !self()->comp()->getOption(TR_Disable64BitRegsOn32Bit);
      }
#endif // TR_TARGET_S390
   return false;
}
```

Then redo all your testing so we have a way to disable the feature if things go wrong. Please test by both enabling (which should be the default with your change) and disabling (by hardcoding the disable option in `jitPreProcess` for example) and running all testing. I'd like to make sure that when we enable this even if nothing breaks we can still disable the feature in the next little while to in case bugs arise down the testing line.

compiler/z/codegen/OMRRegisterDependency.cpp
dhong44: Around linkages we add a register dependency `KillVolHighRegs` only on 32 bit Java is 64 bit registers are allowed. After killing the high regs we change this register dependency to a NoReg. When these register dependencies get here we weren't ignoring them and counted them as  HPRs resulting in an incorrect count for HPRs.

https://github.com/eclipse/omr/blob/b9f595237b4f73dc0008bbf748c394e789fce17e/compiler/z/codegen/OMRLinkage.cpp#L2589-L2594

https://github.com/eclipse/omr/blob/b9f595237b4f73dc0008bbf748c394e789fce17e/compiler/z/codegen/OMRRegisterDependency.cpp#L1191-L1197

compiler/z/codegen/OMRLinkage.cpp
dhong44: 303e1f5

compiler/control/OMROptions.cpp
dhong44: It looks like a `TR_Disable64BitRegsOn32Bit` options already exists.

https://github.com/eclipse/omr/blob/master/compiler/control/OMROptions.cpp#L411

I don't think I can reuse this option since it is used in an optimization and might have unintended side effects if we were to later enable the disabling. Should a different but still similar name be used, such as `TR_Disable64BitRegistersOn32Bit`?

https://github.com/eclipse/omr/blob/master/compiler/optimizer/LocalOpts.cpp#L5984-L5994

fjeremic: I think we should be fine to reuse the same option that already exists. If you check this `` optimization it's end result is deciding whether `useLongRegAllocation` will return true or false. This function is only used in `OMR::CodeGenerator::use64BitRegsOn32Bit()` - a use which you end up deleting as part of this PR. This effectively makes `TR_LongRegAllocation` dead code and any associated code in `TR_Rematerialization`. Sounds like there is more potential for cleanup. Do you want to take care of this in a new commit?

fjeremic: Also please validate my logic here was correct in case I've missed something.

compiler/z/codegen/OMRLinkage.cpp
```diff
@@ -1585,15 +1583,34 @@ OMR::Z::Linkage::copyArgRegister(TR::Node * callNode, TR::Node * child, TR::Regi
...
+         {
+         debugObj->addInstructionComment(toS390RRInstruction(cursor), REG_PARAM);
+         }
+
+      argRegister = self()->cg()->allocateConsecutiveRegisterPair(tempRegL, tempRegH);
+      self()->cg()->stopUsingRegister(argRegister);
+      }
+   else if (!self()->cg()->canClobberNodesRegister(child, 0))
+      {
+      if (TR::Compiler->target.is32Bit() && child->getDataType() == TR::Int64)
```
dhong44: In a previous discussion you mentioned that if we enable 64 bit registers on 32 bit, no long values should be passed in register pairs and we should add an assertion here since this would be dead code. Before, the assertion used here was `TR_ASSERT_FATAL(0, "Long values should not be passed in register pairs, they can be stored in a single GPR");`. However, this would cause failures when disabling these changes, Should I add an assertion such as `TR_ASSERT_FATAL(self()->cg()->use64BitRegsOn32Bit(), "Long values should not be passed in register pairs, they can be stored in a single GPR");` or do you think it is unnecessary?

compiler/control/OMROptions.cpp
dhong44: Remove `TR_LongRegAllocation`: 752e355
Use `TR_Disable64BitRegsOn32Bit`: 31a08e1

compiler/z/codegen/OMRLinkage.cpp
fjeremic: I don't think it hurts to have the assert added if that is the expectation, and AFAIK it should be. Please make sure this is tested when you add the assert back.

compiler/optimizer/LocalOpts.cpp
```diff
@@ -5700,324 +5572,6 @@ void TR_Rematerialization::rematerializeNode(TR::TreeTop *treeTop, TR::Node *par
...
-         traceMsg(comp(), "\tFails test %d\n", testNum);
-      return ;
-      }
-   else*/ if (totalRatio > LONGREG_TOTAL_HIGH_WATERMARK &&
-            spillRatio < LONGREG_SPILL_THRESHOLD &&
-            argRatio < LONGREG_ARG_THRESHOLD)
-      {
-      if (trace())
-         traceMsg(comp(), "\tPasses test %d\n", testNum);
-      comp()->setUseLongRegAllocation(true);
```
fjeremic: Do we also need to remove this API and the associated fields?

compiler/z/codegen/OMRLinkage.cpp
dhong44: In tests from the discussion https://github.com/eclipse/omr/pull/2916#discussion_r217520385 I had the assert when 64 bit regs were enabled and commented out the assert otherwise. This path should be well tested. 

57d3672

compiler/z/codegen/OMRLinkage.cpp
```diff
@@ -1609,6 +1609,7 @@ OMR::Z::Linkage::copyArgRegister(TR::Node * callNode, TR::Node * child, TR::Regi
       {
       if (TR::Compiler->target.is32Bit() && child->getDataType() == TR::Int64)
          {
+	 TR_ASSERT_FATAL(!self()->cg()->use64BitRegsOn32Bit(), "Long values should not be passed in register pairs, they can be stored in a single GPR");
```
fjeremic: Looks like a tab spaces issue here.

dhong44: b040b77

fjeremic:
One last thing left is to clean up `setUseLongRegAllocation` as per above review and I think we should be good to go here.

PR #2917, opened 2018-09-02
Remove use of C++11 nullptr

gacholio:
Breaks older compilers - use NULL instead.

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #2918, opened 2018-09-02
Prevent rare shutdown crashes in well-behaved applications 

gacholio:
Well-behaved applications ensure that all forked threads have terminated
before shutting down the thread library (for example, by using
omrthread_exit with a monitor to signal completion rather than simply
falling off the end of the thread proc).

This change ensures that the thread library shutdown will wait for all
omrthread_exit calls to complete before starting to shut down.

A subtle side-effect of this change is that TLS finalizers will be
called earlier on threads which use omrthread_exit (they will be called
before exitting held monitors rather than after). This is to avoid
calling the finalizers with the global mutex locked. It should have no
effect.


gacholio:
Note: based on #2917, so the second commit should vanish once that's merged.

gacholio:
Note this change does not affect omrthread_exit calls that happen after thread library shutdown has begun (there is no reliable way to do that). It assumes that all threads have been terminated before the shutdown starts, and ensures that those terminations take place before library shutdown.

gacholio:
The appveyor build appears to be broken by a pre-existing issue with debug builds.

```static intptr_t
init_threadParam(char *name, uintptr_t *pDefault)
{
	uintptr_t *p;

	ASSERT(name);
	ASSERT(pDefault);
	ASSERT(0 != *pDefault);

	p = omrthread_global(name);
	if (NULL == p) {
		return -1;
	}
	ASSERT(0 == *p);
	*p = (uintptr_t)pDefault;

	return 0;
}
```

```
ASSERT(0 != *pDefault);
```

appears to be bogus - why would it be illegal for a default value to be 0?

Looks like the source of the problem is `init_spinParameters`:

```
	lib->adaptSpinHoldtime = 0;
	if (init_threadParam("adaptSpinHoldtime", &lib->adaptSpinHoldtime)) {
```

Seems pretty blatantly broken.

charliegracie:
@babsingh thoughts on @gacholio's question?

gacholio:
Were debug builds added recently? This should have been failing on every OMR PR build.

charliegracie:
As far as I can see there have been no changes in the AppVeyor builds in a very long time.

babsingh:
I agree with GAC. `lib->adaptSpin*` variables are set to 0 before `init_threadParam` is used to initialize `"adaptSpin*"` global variables. `ASSERT(0 != *pDefault)` is not needed.

Currently, `ASSERT(0 != *pDefault)` should always fail on all platforms. Are asserts working properly on all platforms?

gacholio:
They're not real asserts - THREAD_ASSERTS must be defined for them to generate any code.

charliegracie:
@genie-omr build all

gacholio:
The problem is not the bad assert, it seems specific to windows.

gacholio:
Assertion will be fixed in #2925

gacholio:
Looks like only the debug build (whatever that entails) on windows fails.

gacholio:
Windows crash is due to shutting the library down multiple times, fix now added to this PR.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
@babsingh can you review this as well please?

thread/common/omrthread.c
```diff
@@ -2121,14 +2145,30 @@ threadInternalExit(void)
...
-	GLOBAL_UNLOCK_SIMPLE(lib);
-
-	/* We couldn't clear TLS earlier because the debug version of
-	 * GLOBAL_UNLOCK_SIMPLE() uses omrthread_self().
+#if defined(THREAD_ASSERTS)
+	/* We can't clear TLS before unlocking the global monitor because
+	 * the debug version of GLOBAL_UNLOCK_SIMPLE() uses omrthread_self().
+	 *
+	 * This opens up a small window of time where the thread library
+	 * could be shut down before this thread has completed exitting.
```
babsingh: typo: exitting -> exiting

thread/common/omrthread.c
```diff
@@ -2121,14 +2145,30 @@ threadInternalExit(void)
...
+	{
+		TLSKEY tlsKey = lib->self_ptr;
+		GLOBAL_UNLOCK_SIMPLE(lib);
+		if (detached) {
+			TLS_SET(tlsKey, NULL);
+		}
+	}
+#else /* THREAD_ASSERTS */
 	if (detached) {
 		TLS_SET(lib->self_ptr, NULL);
```
babsingh: i think `TLS_SET/pthread_setspecific/TlsSetValue` are thread-safe. so, the TLS can be always cleared after releasing the global lock. the `ifdefs THREAD_ASSERTS` can be removed.

thread/common/omrthread.c
```diff
@@ -2071,9 +2091,12 @@ threadFree(omrthread_t thread, int globalAlreadyLocked)
  * Exit from the current thread.
  *
  * If the thread has been detached it is destroyed.
+ *
+ * @param[in] globalAlreadyLocked indicates whether the thread library global mutex is already locked
+ *            Also implies that omrthread_tls_finalize was called before the global mutex was locked
  */
```
babsingh: missing documentation: `@return none` or `return void`

thread/common/omrthread.c
```diff
@@ -679,11 +679,17 @@ init_spinParameters(omrthread_library_t lib)
...
-void
+static void
 omrthread_shutdown(void)
 {
 	omrthread_library_t lib = GLOBAL_DATA(default_library);
 	ASSERT(lib);
+	/* Acquire the global lock here to wait until all outstanding omrthread_exit calls
+	 * have completed. omrthread_exit holds the global lock until the thread is ready
+	 * to exit to the OS.
+	 */
```
babsingh: The goal is to wait here until all outstanding `omrthread_exit` calls have completed. `GLOBAL_LOCK_SIMPLE` uses a mutex. A mutex won't guarantee that all `omrthread_exit` calls have completed. Shouldn't a condition variable be used over here?

thread/common/omrthread.c
gacholio: This has nothing to do with thread safety - the comment clearly states why TLS can't be cleared in the assert case.

thread/common/omrthread.c
gacholio: It was like that when I got here :)

thread/common/omrthread.c
gacholio: I only care about omrthread_exit calls that were in progress when the shutdown begins. Any thread library call issued after shutdown begins is going to crash/fail randomly.

gacholio: The correct pattern is to inform all your forked threads of the need to exit, and wait for them to indicate that with a monitor protecting a count variable or somesuch. Each thread calls omrthread_exit with the monitor after indicating completion. Once they've all checked in, the library shutdown can begin. This code ensures that each thread executing omrthread_exit has reached the point of OS termination (and never needs to reference the thread library structures again). In pratice, only one thread could be in this state when shutdown begins.

thread/common/omrthread.c
gacholio: See below.

gacholio:
I will fix the spelling and return comment provided we don't need to launch another round of builds.

thread/common/omrthread.c
babsingh: i meant ... can't we just use the below code block for all scenarios. this way ... we can avoid duplicate code in the ifdefs. the new scope can be avoided if the TLS key is retrieved at the start of the function. 
```
	{
		TLSKEY tlsKey = lib->self_ptr;
		GLOBAL_UNLOCK_SIMPLE(lib);
		if (detached) {
			TLS_SET(tlsKey, NULL);
		}
	}
```

gacholio: I see what you're saying, but this is still an issue. Using the tlsKey after the unlock still leaves a tiny window where the shutdown could free the key before the TLS_SET takes place. If the key were somehow reallocated, we'd be NULLing someone else's data.

PR #2920, opened 2018-09-04
Do not skip generation of LA in aiadd evaluator

fjeremic:
When evaluating the following tree we currently have a bug where we do
not generate the actual add because the displacement is too large to be
handled by long displacement instructions:

```
aiadd
  aconst 0x784b8500
  iconst 28
```

This is specific to 31-bit. The reason is because we do not generate
the LA instruction to add the 28 to the previous memory reference. This
bug was introduced in commit 564fb25 as part of #2882. In fact the code
that was removed here is actually an optimization on 31-bit that does
not exist for 64-bit but it really should.

This will be addressed in a separate Issue which will be opened as part
of the PR containing this commit.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -7101,19 +7101,7 @@ OMR::Z::TreeEvaluator::aiaddEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
    aiaddMR->populateAddTree(node, cg);
    aiaddMR->eliminateNegativeDisplacement(node, cg);
-
-   // this is partial extract from enforceDisplacementLimit in order to remove the need for the final LA in some cases
-   if (aiaddMR->getOffset() >= MAXLONGDISP && aiaddMR->getOffset() <= TR::getMaxSigned<TR::Int32>() && TR::Compiler->target.is32Bit())
-      {
-      TR::MemoryReference *tempMR = generateS390MemoryReference(aiaddMR->getBaseRegister(), aiaddMR->getIndexRegister(), 0, cg);
-      tempMR->setSymbolReference(aiaddMR->getSymbolReference());
-      generateRXInstruction(cg, TR::InstOpCode::LA, node, targetRegister, tempMR);
-      generateS390ImmOp(cg, TR::InstOpCode::getAddOpCode(), node, targetRegister, targetRegister, (int32_t)aiaddMR->getOffset());
-      }
-   else
-      {
-      aiaddMR->enforceDisplacementLimit(node, cg, NULL);
-      }
+   aiaddMR->enforceDisplacementLimit(node, cg, NULL);
 
    if (node->getOpCodeValue() == TR::aiadd && node->isInternalPointer())
       {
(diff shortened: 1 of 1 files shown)


fjeremic:
@genie-omr build zlinux,zos

vijaysun-omr:
It is a bit uncommon to have an aiadd with two constant children; I'm curious how you ended up with that IL tree (i.e. which pass created it) ? Obviously we do need to handle it correctly and so I don't disagree with your commit.

vijaysun-omr:
Since #2882 was mostly about removing (implicitly or explicitly) unused apis could you please elaborate on how that change actually resulted in this change in behaviour ?

fjeremic:
> It is a bit uncommon to have an aiadd with two constant children; I'm curious how you ended up with that IL tree (i.e. which pass created it) ? Obviously we do need to handle it correctly and so I don't disagree with your commit.

See details in #2923. It was noticed in OpenJ9. Reproducible by any SDK with the following command:

```
./bin/java -Xjit:'limit={java/lang/System.setFieldImpl(Ljava/lang/String;Ljava/lang/Object;)V}(tracefull,log=setFieldImpl.trace)' -version
```

The tree gets generated at ILGen and no subsequent optimization (including simplifier) simplifies the tree.

> Since #2882 was mostly about removing (implicitly or explicitly) unused apis could you please elaborate on how that change actually resulted in this change in behaviour ?

The behavior change was the result of the following line being deleted:
https://github.com/eclipse/omr/pull/2882/files#diff-019309d732e02fc0678eb74f2d192590L7150

Commit 564fb2516755d332a0cdd053dc9fe8defc24689f was supped to deprecate the use of `getBucketPlusIndexRegisters`, however a mistake was made in the `aiadd` evaluator where I did not notice the above line in the `if` block. As such I thought the entire block was dead and removed the `skipFinalLA` variable. This is how the bug got introduced.

This should have really been caught at the OMR level though, hence the need for Tril testing in #2923 that I opened.

vijaysun-omr:
Thanks for answering both questions. 

Regarding the first question, I looked at #2923 and while I got a little bit more information on the origin of the tree (from IL gen and for accessing some instance field off a class) I am still a bit unclear on the context in which this IL is possible. One obvious question is why does'nt the IL gen just generate an aconst instead of this aiadd. If the point is to expose the address of the class separately for ensuring that HCR or AOT can see the address as something to patch up, then I wonder if a loadaddr of that class might have been a better choice of IL (though maybe this bug would probably still have been seen).

fjeremic:
> Regarding the first question, I looked at #2923 and while I got a little bit more information on the origin of the tree (from IL gen and for accessing some instance field off a class) I am still a bit unclear on the context in which this IL is possible. One obvious question is why does'nt the IL gen just generate an aconst instead of this aiadd. If the point is to expose the address of the class separately for ensuring that HCR or AOT can see the address as something to patch up, then I wonder if a loadaddr of that class might have been a better choice of IL (though maybe this bug would probably still have been seen).

Probably a question for the OpenJ9 community. I've opened up an issue [1] and tagged the appropriate people.

[1] https://github.com/eclipse/openj9/issues/2767

vijaysun-omr:
Thanks for the issue for answering the question. Merging this change itself since it is a valid fix regardless.

PR #2924, opened 2018-09-04
Update documents related to C++11 nullptr

0dvictor:
nullptr is not supported by g++ 4.4.7, the oldest gcc/g++ that OMR supports.


fjeremic:
Just making a note that this is in reference to #2917. @0dvictor in the future please add the `[ci skip]` tag to avoid launching builds for documentation changes. Merging without running builds as this does not affect any code.

PR #2925, opened 2018-09-04
Remove incorrect assertion

gacholio:
init_threadParam was asserting that the default value not be 0, which is
incorrect.


charliegracie:
@genie-omr build all

gacholio:
Never shows up anywhere since THREAD_ASSERTS needs to be defined. As soon as it is, there's an immediate crash on startup.

PR #2927, opened 2018-09-04
Remove S390InterfaceCallDataSnippet

NigelYiboYu:
Move S390InterfaceCallDataSnippet out of the OMR project.


compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -406,71 +406,81 @@ uint8_t *
...
-//  0x00000100211cfe80     DC  0x00000000       ; Flags (resolved)
-//  0x00000100211cfe88     DC  0x100211cfe90    ; lastCachedSlotField
+//  DC  0x100211cf58a    ; Call Site RA
+//  DC  0x8020b380       ; Address Of Constant Pool
+//  DC  0x00000040       ; CP Index
+//  DC  0x00000000       ; Interface class
+//  DC  0x00000000       ; Method Index    (contains Method ptr if the target is direct)
+//  DC  0x00000000       ; Thunk address   (used if the target is direct dispatch)
+//  DC  0x00000000       ; Flags (resolved=1, unresolved = 0)
+//  DC  0x100211cfe90    ; lastCachedSlotField
```
fjeremic: Can we pad the 64-bit fields with zeros to signify to the reader that they are 64-bit. For example currently it is non-obvious that "Address Of Constant Pool" is a 64-bit address.

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -406,71 +406,81 @@ uint8_t *
...
+//  Note: some fields and PIC slots will be unused if interface call turns out to be a direct dispatch.
 //
-//  0x00000100211cfe80     DC  0x00000000       ; Flags (resolved)
-//  0x00000100211cfe88     DC  0x100211cfe90    ; lastCachedSlotField
+//  DC  0x100211cf58a    ; Call Site RA
+//  DC  0x8020b380       ; Address Of Constant Pool
+//  DC  0x00000040       ; CP Index
+//  DC  0x00000000       ; Interface class
+//  DC  0x00000000       ; Method Index    (contains Method ptr if the target is direct)
+//  DC  0x00000000       ; Thunk address   (used if the target is direct dispatch)
```
fjeremic: Thunk address of what exactly? Can we better document this?

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -406,71 +406,81 @@ uint8_t *
...
+//  DC  0x00000000       ; Flags (resolved=1, unresolved = 0)
```
fjeremic: Does "resolved=1" mean bit 1, i.e. second bit, indicates resolved? We can better document this I think. Perhaps "Bit 0 indicates whether the call is resolved"? Are there any other bits in use?

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -406,71 +406,81 @@ uint8_t *
...
 
 //  31-Bit Layout.
-//  0x00000000211cfe50     DC  0x0000610ff8     ; Address Of Dispatch Glue
-//  0x00000000211cfe54     DC  0x00211cf58a     ; Call Site RA
-//  0x00000000211cfe58     DC  0x2020b380       ; Address Of Constant Pool
-//  0x00000000211cfe5c     DC  0x00000040       ; CP Index
-//  0x00000000211cfe60     DC  0x00000000       ; Interface class
-//  0x00000000211cfe64     DC  0x00000000       ; Method Index
+//  DC  0x0000610ff8     ; Address Of Dispatch Glue
+//  DC  0x00211cf58a     ; Call Site RA
```
fjeremic: Can we remove the leading zeros so the address is actually 31-bit? Also is there anything different for 31-bit layout? If there are no different fields and just that all fields are 32-bit rather than 64-bit I say we omit this altogether to avoid confusion.

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -406,71 +406,81 @@ uint8_t *
...
+   // For 64-bit single dynamic slot, LPQ and STPQ instructions will be emitted; and they require quadword alignment
+   //
+   // 4-slot snippet has 10 DC fields before the 1st class pointer; whereas single dynamic snippet has 6.
+   // To align class pointers to double/quad word boundaries, we can just align the cursor to boundaries instead
+   // of aligning the the first PIC.
+   int32_t alignment = comp->getOption(TR_enableInterfaceCallCachingSingleDynamicSlot) ? 16 : 8;
+   if (((intptrj_t)cursor % alignment) != 0)
+      {
+      cursor += (alignment - ((intptrj_t)cursor % alignment));
+      }
```
fjeremic: Could you please add a fatal assert following this alignment to validate that `cursor` is aligned properly? This will guard against future modifications to the snippet layout misaligning things.

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -498,6 +508,14 @@ TR::S390InterfaceCallDataSnippet::emitSnippetBody()
    *(intptrj_t *) cursor = 0;
    cursor += TR::Compiler->om.sizeofReferenceAddress();
 
+   // thunk to transition to the interpreter
+   // TODO: enable AOT once thunk relocation is supported.
```
fjeremic: Do we have an issue we can link to in the code which tracks this work? If not can we open one and tag the appropriate people working on getting this done?

compiler/z/codegen/ConstantDataSnippet.cpp
fjeremic: Actually I see there is one below but it only checks for 8 byte alignment? Don't we need 16 byte?

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -1010,90 +1029,99 @@ TR_Debug::print(TR::FILE *pOutFile, TR::S390ConstantDataSnippet * snippet)
...
-         bufferPos += refSize;
-         }
+      printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
+      trfprintf(pOutFile, "DC   \t0x%016lx  # Interpreter thunk", *(intptrj_t*)bufferPos);
+      bufferPos += refSize;
 
-      // non-single cache slots
       bool isSingleDynamicSlot = comp()->getOption(TR_enableInterfaceCallCachingSingleDynamicSlot);
-      if (!isSingleDynamicSlot)
+      bool isZeroSlot =  static_cast<TR::S390InterfaceCallDataSnippet*>(snippet)->getNumInterfaceCallCacheSlots() == 0;
```
fjeremic: What is the point of this variable? I only see one use of it. Is it possible to inline it into the use?

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -1010,90 +1029,99 @@ TR_Debug::print(TR::FILE *pOutFile, TR::S390ConstantDataSnippet * snippet)
...
-         printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
-         trfprintf(pOutFile, "DC   \t0x%016lx  # flags", *(intptrj_t*)bufferPos);
-         bufferPos += refSize;
-         }
+      printPrefix(pOutFile, NULL, bufferPos, sizeof(intptrj_t));
+      trfprintf(pOutFile, "DC   \t0x%016lx  # Interpreter thunk", *(intptrj_t*)bufferPos);
+      bufferPos += refSize;
 
-      // non-single cache slots
       bool isSingleDynamicSlot = comp()->getOption(TR_enableInterfaceCallCachingSingleDynamicSlot);
```
fjeremic: This variable is only used on the else path of the following block, so there is no need to have the preemptive `comp()->getOption` call outside of the else path.

fjeremic:
By the way @NigelYiboYu you can add single bullet lines in your commit body and GitHub will format them nicely:

>Add thunk field to interface call data snippet
>
>- Add a new thunk to interface call data snippet to support
>direct dispatch to interpreted methods.
>- Update interface call data snippet comment to reflect this new change.
>- Fix interfaceCallDataSnippet trace printing function for zero-slot and
>single dynamic slot cases.
>
>Signed-off-by: Nigel Yu <yunigel@ca.ibm.com>

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: I was a bit hesitant to mention `J2I` because that's Java specific.

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: Will update the comment.
The whole flag is pointer sized and is either 0 or 1. This indicates if the call is resolved or not. It's sized like this to make sure PIC slots are aligned properly.

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: There's no difference with 31-bit except the fields are all word sized. Will remove this.

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: I'll update the assert to handle 16-byte alignment check.

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: Fixed.

compiler/z/codegen/ConstantDataSnippet.cpp
NigelYiboYu: Fixed.

NigelYiboYu:
Updated the commit to address all previous concerns.

fjeremic:
@NigelYiboYu given that we just moved the entire snippet to https://github.com/eclipse/openj9/pull/2665 can we re-purpose this PR to delete it from OMR?

NigelYiboYu:
The commit has been updated to remove `TR::S390InterfaceCallDataSnippet` from the OMR project. Marking this WIP until all tests pass.

NigelYiboYu:
@fjeremic This is ready for review, and OMR sanity tests.

fjeremic:
@genie-omr build all

fjeremic:
Z only change. I think we have the deleted lines in Debug.hpp covered by the OMR builds. I'm merging this without waiting for Travis CI or AppVeyor.

PR #2928, opened 2018-09-04
Add experimental GC context and object scanner APIs

rwy0717:
This commit introduces two experimental APIs to the GC. Both APIs are
only supported under cmake, and are permanantly disabled in autotools.

## Experimental Context

The first is a high level wrapper for initializing and tearing down the
GC. To enable this API, use the flag OMR_GC_EXPERIMENTAL_CONTEXT.

In the experimental context APIs, the  GC is structured into three
tiers:

1. OMR::Runtime - RAII-style process-wide initialization. Creating a
   Runtime will bring  up the thread library, port library, and
   OMR_Runtime. There should only be one instance of the runtime in a
   program.
2. OMR::GC::System - A managed heap. Provides it's own shared-access
   mechanism. Typically used on a per-VM basis.
3. OMR::GC::Context - per-thread context attached to a GC::System. Most
   GC APIs require a GC::Context. Having a context implies shared heap
   access. Contexts can be upgraded to exclusive contexts.

This code is grouped under the `gc/api` directory. New public
headers are grouped under `gc/api/include/OMR/GC/*`. They are meant to be
included as `#include <OMR/GC/Context.hpp>`.

### Heap access

GC::System will implement it's own shared access mechanism. Currently unimplemented.

### Built in rooting

The experimental context also provides a new collection of utilities for rooting references. The APIs provided in this PR are:
1. Root scanning callback registration. MarkingFns can registered on each GC::Context or each GC::System. The marking functions are called at the beginning of a GC.
3. StackRoots, an RAII rooting mechanism for rooting references on the C++ stack. The stack roots are scanned and fixed up as part of the GC.

The rooting API also provides a GC::Handle, which acts as a reference to a root. The Handle provides a pointer-like API, which under the hood will indirect through root slots.

### GC-provided glue

In order to implement the heap access and rooting mechanisms, enabling the context API will  turn on some GC-provided implementations of glue, including: 
* MarkingDelegate
* EnvironmentDelegate

To support this notion of gc-provided glue, this PR rearranges the
example's glue headers into include and include_extended.
include_extended is only added to the ipath if
OMR_GC_EXPERIMENTAL_CONTEXT is off, to ensure there is only ever one
version of each glue header available.

## Experimental object scanner

The second API is a new piece of glue, the OMRClient::GC::ObjectScanner.
This is an attempt to write an object scanner API that is simpler for
clients to implement, while  supporting custom reference encodings. The
new glue APIs are placed under the OMRClient namespace. Public headers
are included as <OMRClient/GC/ObjectScanner.hpp>.

The experimental object scanner makes compressed object pointers a
concern of the application. The long term goal is to remove the
compressed references flag from OMR entirely. Through utility classes
provided by OMR, implementing compressed references should be trivial
for applications to implement.

The object scanner, if successful, will deprecate:
* fomrobject_t
* GC_SlotObject


charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build aix,win,zos

charliegracie:
@genie-omr build aix,win,zos

gc/CMakeLists.txt
```diff
@@ -167,6 +167,9 @@ add_library(omrgc STATIC
 	stats/ScavengerStats.cpp # TODO only compile if scavenger or VLHGC. Is this actually used by VLHGC?
 	stats/SweepStats.cpp
 
+	stats/ConcurrentGCStats.cpp
```
charliegracie: Shouldn't these two files only be included if the proper define is set?

gc/api/.clang-format
```diff
@@ -0,0 +1,114 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Mozilla
```
charliegracie: Was this file meant to be included?

gc/api/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+		include/
+)
+
+# https://cmake.org/cmake/help/v3.6/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES
+target_compile_features(omrgc
+	PUBLIC
+		cxx_constexpr
+		# cxx_auto_type
+		cxx_deleted_functions
+		cxx_delegating_constructors
```
charliegracie: Are all of these available on the OMR minimum supported compiler levels?

gc/CMakeLists.txt
rwy0717: I think this was an old compilation fix that's not needed any more. I'm not sure since I've squashed my history. I'll remove it.

gc/api/.clang-format
rwy0717: Yeah, this is the format I've been using for the API directory. I figured it'd be useful for anyone making a change in the code. Do you want me to drop it from the PR?

gc/api/CMakeLists.txt
rwy0717: Most of these features are actually for Om, not the object scanner / context. I can probably get rid of most (if not all) for this PR.

To answer the question though, I believe most of these features are _not_ supported, so that will be a problem for the Om PR.

rwy0717:
I suppose it's worthwhile to mention that the jenkins builds are only testing with the two new features disabled.

gc/api/CMakeLists.txt
rwy0717: Well after looking over the code, this is what I'm seeing:
* noexcept: in my case, I can just use the deprecated `throw()` syntax, or just get rid of the exception specifiers entirely, or define a OMR_NOEXCEPT symbol depending on the compiler support.
* nullptr: can be replaced by `NULL`, except in the `OMR::GC::Ref<T>`, where it is used in the interface. I can remove Ref from this PR, and deal with it later.
* delegating constructors: I can rewrite some code to not use this feature, it shouldn't be difficult.
* constexpr: this one is probably going to hurt to remove, I'm using constexpr a lot. APIs in Om will have to drop constexpr too. I think it's worthwhile to find a way to support constexpr when available. Maybe we can test for `__cplusplus` to define`OMR_CXX11_CONSTEXPR`, and fall back to `inline` if its not available in the current compilation unit? That might break ODR though.
* Static assert / type traits: they're such a good safety net that I'd be sad to remove them. Maybe I can hide the static asserts under an ifdef like `OMR_HAVE_CXX11_STATIC_ASSERT`, or define an `OMR_STATIC_ASSERT` macro. It's harder to replace the type traits though. I'm using `std::is_standard_layout`.
* range-based for loop: I'm using it to iterate the marking functions attached to the context, but I can get rid of this no-problem.

rwy0717: There is a module in cmake for automatically defining portable wrappers around c++ features, called [WriteCompilerDetectionHeader](https://cmake.org/cmake/help/v3.12/module/WriteCompilerDetectionHeader.html#module:WriteCompilerDetectionHeader).

It looks extremely useful, but it would (using a symbol prefix of OMR) define symbols such as `OMR_COMPILER_IS_SCO`, which might conflict with the compiler project. Alternatively, using a prefix of `OMR_CXX` would define symbols like `OMR_CXX_COMPILER_CXX_CONSTEXPR` which is not great either.

Something like this though, would be a good way to conditionally support newer cxx features when available.

rwy0717: I opened an issue for implementing portable c++11 feature wrappers:
https://github.com/eclipse/omr/issues/2955

rwy0717: There is some new discussion ongoing about how (or if) we'll be making use of C++11 in the future. Until we land on a decision, I'd like to leave the C++11 isms in this PR as-is. Since this code is disabled by default, and explicitly opt-in only, this should have no impact on existing consumers of OMR. When we figure out what to do about C++11, I'll update this code to follow our new coding practices.

ktbriggs:
> some GC glue is actually provided by the GC, such as the EnvironmentLanguageInterface

Note that EnvironmentLanguageInterface, ConfiguraitonLanguageInterface and CollectorLanguageInterface are deprecated. The first two do not exist in OMR source base anymore, and the third has been whittled down to a deprecated rump of scavenger_*() methods that are still being used by scavenger. 

Instead, EnvironmentDelegate, ConfigurationDelegate and a collection of collector-related delegates provide the integration points for consumers of OMR::GC components.

rwy0717:
You're right, it's actually `MarkingDelegate.cpp` and `EnvironmentDelegate.cpp` that are provided by the this API.

charliegracie:
Can you enable this in one of the travis tests? Likely not OS X since it only has one test but one of the other CMake Linux jobs would work if there are any.

charliegracie:
@rwy0717 can you resolve the conflict?

rwy0717:
resolved.

rwy0717:
Closing to contribute in smaller PRs.

PR #2929, opened 2018-09-04
Change getmacros annotation to not use @ symbol

mikezhang1234567890:
Instead, change the sed scripts to use the DDRFILE_BEGIN and DDRFILE_END markers to denote the annotations to scrape.

This is needed since on AIX, xlC throws illegal character warnings for cpp source files during the preprocessor.

This should be merged after the corresponding OpenJ9 change (https://github.com/eclipse/openj9/pull/2762) which will accept the new pattern.

### DIFF
--- a/omrmakefiles/rules.mk
+++ b/omrmakefiles/rules.mk
@@ -206,12 +206,15 @@ define CLEAN_COMMAND
 endef
 endif
 
+DDR_SED_COMMAND := \
+    sed -n -e '/^DDRFILE_BEGIN /,/^DDRFILE_END /s/^/@/p'
+
 define DDR_C_COMMAND
-$(CC) $(CFLAGS) $(MODULE_CPPFLAGS) $(GLOBAL_CPPFLAGS) -E $< | sed -n -e '/^@/p' > $@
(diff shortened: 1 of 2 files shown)


charliegracie:
@genie-omr build all

omrmakefiles/rules.mk
```diff
@@ -206,12 +206,15 @@ define CLEAN_COMMAND
 endef
 endif
 
+DDR_SED_COMMAND := \
+    sed -n -e '/^DDRFILE_BEGIN /,/^DDRFILE_END /s/^/@/p'
```
keithc-ca: I thought the point of this was to stop producing files with `@` characters?

keithc-ca: Ignore my last comment: we still want the `@` characters in `*.i` files otherwise we'd have to change `MacroTool.cpp`.

mikezhang1234567890: We just can't have the `getmacros` script write `@` characters into the `.cpp` source files, since that's what xlC consumes. The rest of the pipeline still expects the macro info scraped from the annotated and preprocessed source to have `@` characters to begin lines.

PR #2931, opened 2018-09-05
Implement JitBuilder bit-wise type conversion services

### LINKED ISSUE #2921: Add a `reinterpret_cast`-like service to JitBuilder
JitBuilder should have a service that allows "reinterpreting" bits from one type to another. Specifically, this service would be used to generate the following IL opcodes:

* `ibits2f`
* `fbits2i`
* `lbits2d`
* `dbits2l`

Leonardo2718:
This commit adds the `ConvertBitsTo` service to JitBuilder. It allows an
integer value's bit pattern to be reinterpreted as a floating point
value and vice versa. It maps directly to the following OMR compiler
IL opcodes:

- `ibits2f`
- `fbits2i`
- `lbits2d`
- `dbits2l`

Fixes #2921

### DIFF
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -1005,6 +1005,30 @@ OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned
    return convertedValue;
    }
 
+TR::IlValue*
+OMR::IlBuilder::ConvertBitsTo(TR::IlType* t, TR::IlValue* v)
+   {
+   TR::DataType typeFrom = v->getDataType();
+   TR::DataType typeTo = t->getPrimitiveType();
+
+   if (typeTo == typeFrom)
+      {
+      TraceIL("IlBuilder[ %p ]::%d is ConvertBitsTo (already has type %s) %d\n", this, v->getID(), t->getName(), v->getID());
+      return v;
+      }
+
+   TR::ILOpCodes convertOpcode = TR::DataType::getDataTypeBitConversion(typeFrom, typeTo);
+   TR_ASSERT(convertOpcode != TR::BadILOp && TR::DataType::getSize(typeTo) == TR::DataType::getSize(typeFrom),
+             "Builder [ %p ] requested bit conversion for value %d from type of size %d (%s) to type of size %d (%s) (consider using ConvertTo() to for narrowing/widening)",
+             this, v->getID(), TR::DataType::getSize(typeFrom), typeFrom.toString(), TR::DataType::getSize(typeTo), typeTo.toString());
+   TR_ASSERT(convertOpcode != TR::BadILOp, "Builder [ %p ] unknown bit conversion requested for value %d (%s) to type %s", this, v->getID(), typeFrom.toString(), t->getName());
+
+   TR::Node *result = TR::Node::create(convertOpcode, 1, loadValue(v));
+   TR::IlValue *convertedValue = newValue(t, result);
+   TraceIL("IlBuilder[ %p ]::%d is CoerceTo(%s) %d\n", this, convertedValue->getID(), t->getName(), v->getID());
+   return convertedValue;
+   }
+
 TR::IlValue *
 OMR::IlBuilder::unaryOp(TR::ILOpCodes op, TR::IlValue *v)
    {
--- a/compiler/ilgen/OMRIlBuilder.hpp
+++ b/compiler/ilgen/OMRIlBuilder.hpp
@@ -255,6 +255,18 @@ public:
    TR::IlValue *ConvertTo(TR::IlType *t, TR::IlValue *v);
    TR::IlValue *UnsignedConvertTo(TR::IlType *t, TR::IlValue *v);
 
+   /**
+    * @brief Convert the bit representation of an IlValue to a given type
+    * @param type is the target type of the conversion
+    * @param value is the value to be converted
+    * @return the IlValue converted to the specified type
+    *
+    * This service allows the bit representation of a floating-point
+    * IlValue to be converted to an integer IlValue and vice versa.
+    * The floating-point and integer types must be of the same bit-width.
+    */
+   TR::IlValue* ConvertBitsTo(TR::IlType* type, TR::IlValue* value);
+
    // memory
    TR::IlValue *CreateLocalArray(int32_t numElements, TR::IlType *elementType);
    TR::IlValue *CreateLocalStruct(TR::IlType *structType);
--- a/fvtest/jitbuildertest/Makefile
+++ b/fvtest/jitbuildertest/Makefile
@@ -36,7 +36,8 @@ OBJECTS := \
 	WorklistTest \
 	IfThenElseTest \
 	CallReturnTest \
-	FieldNameTest
+	FieldNameTest \
+	ConvertBitsTest
 
 OBJECTS := $(addsuffix $(OBJEXT),$(OBJECTS))
 
--- a/compiler/il/OMRDataTypes.cpp
+++ b/compiler/il/OMRDataTypes.cpp
@@ -203,6 +203,23 @@ OMR::DataType::getDataTypeConversion(TR::DataType t1, TR::DataType t2)
    return OMR::conversionMap[t1][t2];
    }
 
+TR::ILOpCodes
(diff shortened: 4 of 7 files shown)


compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1005,6 +1005,31 @@ OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned
...
+   else return TR::BadILOp;
+   }
+
+TR::IlValue*
+OMR::IlBuilder::CoerceTo(TR::IlType* t, TR::IlValue* v)
+   {
+   TR::DataType typeFrom = v->getDataType();
+   TR::DataType typeTo = t->getPrimitiveType();
+
+   TR::ILOpCodes coerceOpcode = getCoercionOpcode(typeFrom, typeTo);
```
mstoodle: should insert an `if (toType != fromType)` around all this code...if the type is already correct, then you'll get a BadILOp, but there's no real reason to assert in that scenario.

compiler/ilgen/OMRIlBuilder.hpp
```diff
@@ -255,6 +255,18 @@ class IlBuilder : public TR::IlInjector
...
+    * @brief Coerce an IlValue to a given type
+    * @param type is the target type of the coercion
+    * @param value is the value to be coerced
+    * @return the IlValue coerced to the specified type
+    *
+    * This service acts as a kind of `reinterpret_cast` for JitBuilder IlValues.
+    * Given some input value, it produces a new value that represents the same
+    * bit pattern as the original one, but with a different type.
+    */
+   TR::IlValue* CoerceTo(TR::IlType* type, TR::IlValue* value);
```
mstoodle: Maybe we should borrow the verb from C++ here and call it `ReinterpretAs` or `ReinterpretCast` ?

The primary hit on google for "coerce a type" yields examples that are more like our ConvertTo service.

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1005,6 +1005,31 @@ OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned
    return convertedValue;
    }
 
+static TR::ILOpCodes
+getCoercionOpcode(TR::DataType fromType, TR::DataType toType)
+   {
+   if (fromType == TR::Int32 && toType == TR::Float) return TR::ibits2f;
```
mstoodle: This code isn't indented properly.

Also, I think a better place for this kind of utility function would be in `compiler/il/OMRILOps.[ch]pp` and just called from here.

compiler/ilgen/OMRIlBuilder.hpp
rwy0717: Agree, coerce to me usually means "different representation, equivalent value".  What about
What about some derivative of "retype bits" or "force type to" ?

rwy0717:
You’re not supporting implicit widening/narrowing, which I think is the right thing for a "bitwise reinterpret", but maybe we can put out a better error message when the widths don’t match, and suggest which APIs could do the narrowing/widening?

compiler/ilgen/OMRIlBuilder.hpp
Leonardo2718: Actually, as I explained in the issue, I specifically wanted to avoid borrowing C++'s `reinterpret_cast` because it specifically behaves differently (see https://en.cppreference.com/w/cpp/language/reinterpret_cast and https://godbolt.org/z/jZI_7q).

rwy0717: Not sure the pointer issue you mentioned is legit--reinterpret seems to be a straightforward (if restrictive) type-changing operation. FWIW, llvm calls this op [bitcast to](https://llvm.org/docs/LangRef.html#bitcast-to-instruction), and there's a new [`bit_cast`](https://en.cppreference.com/w/cpp/numeric/bit_cast) in c++20.

Leonardo2718: Perhaps I didn't explain my concern well enough. While it's true that the type changing aspect of`reinterpret_cast` is straightforward, the restrictions on it make use-cases for it are very different than the use-cases for the existing "CoerceTo" service.

I'd be okay with renaming it to `BitCastTo()` to maybe `ConvertBitsTo(). @mstoodle thoughts?

mstoodle: I'd go with `ConvertBitsTo()` for synergy with `ConvertTo()`. Any opinions on whether we should create a new `ConvertValueTo()` and deprecate `ConvertTo()` at some later date?

Leonardo2718: :+1: for `ConvertValueTo()`

compiler/ilgen/OMRIlBuilder.cpp
Leonardo2718: Fixed!

compiler/ilgen/OMRIlBuilder.hpp
Leonardo2718: Fixed! Settled on `ConvertBitsTo()`.

compiler/ilgen/OMRIlBuilder.cpp
Leonardo2718: Done!

Leonardo2718:
@mstoodle @rwy0717 I think I've addressed all your concerns and also added some tests.

compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -1005,6 +1005,30 @@ OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned
...
+
+   if (typeTo == typeFrom)
+      {
+      TraceIL("IlBuilder[ %p ]::%d is CoerceTo (already has type %s) %d\n", this, v->getID(), t->getName(), v->getID());
+      return v;
+      }
+
+   TR::ILOpCodes coerceOpcode = TR::DataType::getDataTypeBitConversion(typeFrom, typeTo);
+   TR_ASSERT(coerceOpcode != TR::BadILOp && TR::DataType::getSize(typeTo) == TR::DataType::getSize(typeFrom),
+             "Builder [ %p ] requested bit conversion for value %d from type of size %d (%s) to type of size %d (%s) (consider using ConverTo() to for narrowing/widening)",
```
mstoodle: typo "ConverTo"

Leonardo2718: :man_facepalming: Fixed!

compiler/ilgen/OMRIlBuilder.hpp
```diff
@@ -255,6 +255,18 @@ class IlBuilder : public TR::IlInjector
    TR::IlValue *ConvertTo(TR::IlType *t, TR::IlValue *v);
    TR::IlValue *UnsignedConvertTo(TR::IlType *t, TR::IlValue *v);
 
+   /**
+    * @brief Convert the bit representation of an IlValue to a given type
+    * @param type is the target type of the coercion
+    * @param value is the value to be coerced
+    * @return the IlValue coerced to the specified type
```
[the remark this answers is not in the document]
mstoodle: and don't forget the commit/PR message please :)

fvtest/jitbuildertest/Makefile
```diff
@@ -36,7 +36,8 @@ OBJECTS := \
 	WorklistTest \
 	IfThenElseTest \
 	CallReturnTest \
-	FieldNameTest
+	FieldNameTest \
+  ConvertBitsTest
```
charliegracie: Is there a spacing issue here or is it just GitHub?

compiler/ilgen/OMRIlBuilder.hpp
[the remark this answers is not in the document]
Leonardo2718: Yep, definitely. Fixed!

fvtest/jitbuildertest/Makefile
Leonardo2718: Nope, definitely a spacing issue :sweat_smile:. Thanks for pointing it out. Fixed!

Leonardo2718:
Ping @mstoodle

mstoodle:
genie-omr build all

PR #2934, opened 2018-09-05
Introduce CodeGen support queries for intrinsics

0dvictor:
Two CodeGen queries are introduced:
```
    ilOpCodeForIntrinsicMethod()
    isIntrinsicMethodSupported()
```
Two CodeGen queries are updated:
```
    isILOpCodeSupported()
    suppressInliningOfRecognizedMethod()
```

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -489,8 +489,38 @@ class OMR_EXTENSIBLE CodeGenerator
    bool isRegisterClobberable(TR::Register *reg, uint16_t count);
 
    // ilgen
-   bool ilOpCodeIsSupported(TR::ILOpCodes); // no virt, default, cast
 
+   /**
+    * @brief Returns if an IL OpCode is supported by current CodeGen
+    *
+    * @param op The IL OpCode being checked.
+    *
+    * @return True if the IL OpCode is supported otherwise false.
+    */
+   static bool isILOpCodeSupported(TR::ILOpCodes op);
+
+   /**
+    * @brief Returns the corresponding IL OpCode for an intrinsic method
+    *
+    * This query maps an intrinsic method to an IL OpCode, with the requirement that
+    * the method's child(ren) corresponds to the OpCode's child(ren) exactly.
+    * It is usually used by the IL Gen transforming the intrinsic method to IL OpCode
+    * so that it can leverage existing framework for better optimization.
+    *
+    * @param method The intrinsic method being checked.
+    *
+    * @return The corresponding IL OpCode for the intrinsic method.
+    */
+   static TR::ILOpCodes ilOpCodeForIntrinsicMethod(TR::RecognizedMethod method) { return TR::BadILOp; }
+
+   /**
+    * @brief Returns if an intrinsic method is supported by current CodeGen
+    *
+    * @param method The intrinsic method being checked.
+    *
+    * @return True if the intrinsic method is supported otherwise false.
+    */
+   static inline bool isIntrinsicMethodSupported(TR::RecognizedMethod method);
 
 
    TR::Recompilation *allocateRecompilationInfo() { return NULL; }
@@ -592,7 +622,7 @@ class OMR_EXTENSIBLE CodeGenerator
     * @param method : the recognized method to consider
     * @return true if inlining should be suppressed; false otherwise
     */
-   bool suppressInliningOfRecognizedMethod(TR::RecognizedMethod method) {return false;}
+   bool suppressInliningOfRecognizedMethod(TR::RecognizedMethod method);
 
    // --------------------------------------------------------------------------
    // Optimizer, not code generator
--- a/compiler/codegen/OMRCodeGenerator_inlines.hpp
+++ b/compiler/codegen/OMRCodeGenerator_inlines.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -24,10 +24,19 @@
 
 #include "codegen/OMRCodeGenerator.hpp"
 
-TR::CodeGenerator*
-OMR::CodeGenerator::self()
+inline TR::CodeGenerator* OMR::CodeGenerator::self()
    {
    return static_cast<TR::CodeGenerator*>(this);
+   }
+
+inline bool OMR::CodeGenerator::isIntrinsicMethodSupported(TR::RecognizedMethod m)
+   {
+   return TR::CodeGenerator::isILOpCodeSupported(TR::CodeGenerator::ilOpCodeForIntrinsicMethod(m));
+   }
+
+inline bool OMR::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod method)
+   {
+   return TR::CodeGenerator::isIntrinsicMethodSupported(method);
    }
 
 #endif
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -2961,7 +2961,7 @@ OMR::CodeGenerator::canNullChkBeImplicit(TR::Node *node, bool doChecks)
    return false;
    }
 
-bool OMR::CodeGenerator::ilOpCodeIsSupported(TR::ILOpCodes o)
+bool OMR::CodeGenerator::isILOpCodeSupported(TR::ILOpCodes o)
    {
 	return (_nodeToInstrEvaluators[o] != TR::TreeEvaluator::unImpOpEvaluator) &&
 	      (_nodeToInstrEvaluators[o] != TR::TreeEvaluator::badILOpEvaluator);
--- a/compiler/p/codegen/OMRCodeGenerator.cpp
+++ b/compiler/p/codegen/OMRCodeGenerator.cpp
@@ -3212,9 +3212,9 @@ TR::Snippet *OMR::Power::CodeGenerator::findSnippetInstructionsFromLabel(TR::Lab
    return NULL;
    }
 
-bool OMR::Power::CodeGenerator::ilOpCodeIsSupported(TR::ILOpCodes o)
+bool OMR::Power::CodeGenerator::isILOpCodeSupported(TR::ILOpCodes o)
    {
-   if (!OMR::CodeGenerator::ilOpCodeIsSupported(o))
+   if (!OMR::CodeGenerator::isILOpCodeSupported(o))
       {
       return false;
       }
--- a/compiler/p/codegen/OMRCodeGenerator.hpp
+++ b/compiler/p/codegen/OMRCodeGenerator.hpp
@@ -436,7 +436,7 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    int64_t getSmallestPosConstThatMustBeMaterialized() {return 32768;}  // maximum 16-bit signed int plus 1
    bool shouldValueBeInACommonedNode(int64_t); // no virt, cast
 
-   bool ilOpCodeIsSupported(TR::ILOpCodes);
+   static bool isILOpCodeSupported(TR::ILOpCodes);
    // Constant Data update
    bool checkAndFetchRequestor(TR::Instruction *instr, TR::Instruction **q);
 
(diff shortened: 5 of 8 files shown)


compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -489,8 +489,30 @@ class OMR_EXTENSIBLE CodeGenerator
...
-   bool ilOpCodeIsSupported(TR::ILOpCodes); // no virt, default, cast
-
+   /**
+    * @brief Returns if an IL OpCode is supported by current CodeGen
+    *
+    * @param op The IL OpCode being checked.
+    *
+    * @return True if the IL OpCode is supported otherwise false.
+    */
+   static bool ilOpCodeIsSupported(TR::ILOpCodes op); // no virt, default, cast
```
fjeremic: Isn't the whole point of OMR that all IL be platform agnostic? Why do we need this query exactly?

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -489,8 +489,30 @@ class OMR_EXTENSIBLE CodeGenerator
...
+   /**
+    * @brief Returns the corresponding IL OpCode for an intrinsic function
+    *
+    * @param op The intrinsic function being checked.
+    *
+    * @return The corresponding IL OpCode for the intrinsic function.
+    */
+   static TR::ILOpCodes ilOpCodeForIntrinsicFunction(TR::RecognizedMethod m) { return TR::BadILOp; }// no virt, default, cast
```
fjeremic: This doesn't sound like an OMR type query since its very downstream dependent. What is the envisioned use for this API?

compiler/codegen/OMRCodeGenerator.hpp
0dvictor: Each platform only supports a subset of OMR IL OpCode; therefore, the Optimizer need know whether an OpCode is supported before utilizing it. One example is that Power supports many math OpCode while X86 doesn't, i.e. `dsin`, `dcos`, `dlog`, etc.

compiler/codegen/OMRCodeGenerator.hpp
0dvictor: The concept of intrinsic function exists in most if not all programming languages, to expose features that is not trivial in the language itself, for example, parallelization, complex mathematics, and synchronization. OMR providing this query can help in two aspects: 1. providing a unified framework for downstream projects to better organize their intrinsic functions; 2. giving Optimizer a better idea about an intrinsic function.

0dvictor: `suppressInliningOfRecognizedMethod()` is a perfect example for using this query so that I've updated the change to include `suppressInliningOfRecognizedMethod()`.

compiler/codegen/OMRCodeGenerator_inlines.hpp
```diff
@@ -24,10 +24,19 @@
...
    {
    return static_cast<TR::CodeGenerator*>(this);
    }
 
+inline bool OMR::CodeGenerator::intrinsicFunctionSupported(TR::RecognizedMethod m)
+   {
+   return TR::CodeGenerator::ilOpCodeIsSupported(TR::CodeGenerator::ilOpCodeForIntrinsicFunction(m));
+   }
+
+inline bool OMR::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod method)
```
fjeremic: Why should the code generator care about this? Shouldn't the inliner contain such a query and possibly delegate a question to the code generator if it needs help? i.e. shouldn't this function be defined in the inliner as follows:

```
inline bool OMR::CodeGenerator::suppressInliningOfRecognizedMethod(TR::RecognizedMethod method)
   {
   return cg->intrinsicFunctionSupported(method);
   }
```

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Thanks for the explanation.

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Ok, sounds reasonable. One comment though in that to my knowledge for symmetry we seem to prefer to use "method" rather than "function" in OMR, i.e. `RecognizedMethod`, `OMRMethod`, `TrustedMethod`, `MethodInliner`, etc. To stay consistent what do you think about renaming this API to `ilOpCodeForIntrinsicMethod` to match the type name of it's only argument?

compiler/codegen/OMRCodeGenerator_inlines.hpp
fjeremic: Or I suppose now that `intrinsicFunctionSupported` is static it would be defined exactly the same, except in the inliner which cares about inlining rather than the code generator which cares about how to evaluate IL to produce machine code.

ymanton:
Is there an obvious benefit to intrinsics having their own IL op? Why not just leave them as direct calls to intrinsic function symbols that the code generator can evaluate however it likes once the method gets to Instruction Selection?

compiler/codegen/OMRCodeGenerator_inlines.hpp
0dvictor: Inliner, or Optimizer in general, need consult CodeGen whether certain feature is supported/enabled unless the feature is assumed to be supported by all CodeGen. In this case, reasons behind whether to suppress inlining of a recognized method vary so that it make sense for CodeGen to have a dedicated query.

0dvictor:
> Is there an obvious benefit to intrinsics having their own IL ops? Why not just leave them as direct calls to intrinsic function symbols that the code generator can evaluate however it likes once the method gets to Instruction Selection?

There are definitely benefits for certain ones although not for all. Many intrinsics are directly map to existing IL ops, by transforming the call into IL OpCde early, it can go through existing Optimizer and hence potentially getting better code.

ymanton:
> There are definitely benefits for certain ones although not for all. Many intrinsics are directly map to existing IL ops, by transforming the call into IL OpCde early, it can go through existing Optimizer and hence potentially getting better code.

But that benefit only exists for intrinsics that map to an existing IL op that the OMR optimizer already knows about. Is this facility only limited to that case?

What about the more general case of adding an intrinsic that doesn't map to an existing IL op? I assumed you expect down-stream consumers to add a corresponding IL op such intrinsics. If so, representing it as a new IL op has far design and maintenance implications and should really be thought through; I don't think we want an explosion of IL ops. Representing it as a call has much less burden, all of the infrastructure to call arbitrary symbols already exists, you can have as many as you want at no extra cost, and they can be optimized the same as any other IL op.

Since it's not clear exactly how OMR compiler consumers can or should extend the IL I'm not sure it makes sense to add a mechanism that makes use of that before we've even decided how it should work or if it's even the right decision because once that proliferates it will be hard to undo.

0dvictor:
> But that benefit only exists for intrinsics that map to an existing IL op that the OMR optimizer already knows about. Is this facility only limited to that case?

Yes, this facility is intended to use only when the intrinsic method maps to an existing IL OpCode.

>What about the more general case of adding an intrinsic that doesn't map to an existing IL op? I assumed you expect down-stream consumers to add a corresponding IL op for such intrinsics. If so, representing it as a new IL op has far reaching design and maintenance implications and should really be thought through; I don't think we want an explosion of IL ops. Representing it as a call has much less burden, all of the infrastructure to call arbitrary symbols already exists, you can have as many as you want at no extra cost, and they can be optimized the same as any other IL op.

This is a good point; however it is beyond the scope of this changeset. This changeset does not give any indication whether a new IL OpCode should be introduced or avoided.

>Since it's not clear exactly how OMR compiler consumers can or should extend the IL I'm not sure it makes sense to add a mechanism that makes use of that before we've even decided how it should work or if it's even the right decision, because once that proliferates it will be hard to undo.

Again this is beyond the scope of this changeset. By providing a facility to aid mapping intrinsic methods to existing IL OpCodes does not indicate which way the consumers should go. Consumers can easily propose to add a new IL OpCode without this facility. Whether such a IL OpCode should be introduced must be considered case-by-case.

ymanton:
OK, I can see value in that. Can you add a line or two to the doxygen comment that tells the reader what these are for and how you intend them to be used? I would not have been able to guess what you had in mind had I not asked. Some justification and concrete motivation for this sort of thing would be nice, otherwise it's hard to provide a meaningful review.

Also, this is only suitable for cases where the method's children correspond to the IL ops children exactly, right? Is that satisfactory for your use cases and have you given any thought to the other cases?

0dvictor:
> OK, I can see value in that. Can you add a line or two to the doxygen comment that tells the reader what these are for and how you intend them to be used? I would not have been able to guess what you had in mind had I not asked. Some justification and concrete motivation for this sort of thing would be nice, otherwise it's hard to provide a meaningful review.

Good suggestion, I've updated the comment to include its exact behavior and usual usage.

> Also, this is only suitable for cases where the method's children correspond to the IL ops children exactly, right? Is that satisfactory for your use cases and have you given any thought to the other cases?

Yes, this is only for those method whose children exactly correspond to an IL OpCode. Such intrinsic methods are the most intuitive ones and have significant percentage in downstream projects like OpenJ9. In fact, it is a common design for many programming languages like Intel intrinsics for C/C++ and math utilities in Java.
For methods whose children do not exactly correspond to an IL OpCode, it is truely downstream project specific, OMR cannot and should not put any constraint or assumptions on their shape. Downstream projects' IL Gen need do the transformation on its own.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -489,8 +489,35 @@ class OMR_EXTENSIBLE CodeGenerator
...
+    * This query maps an intrinsic method to an IL OpCode, with the requirement that
+    * the method's child(ren) corresponds to the OpCode's child(ren) exactly.
+    * It is usually used by the IL Gen transforming the intrinsic method to IL OpCode
+    * so that it can leverage existing framework for better optimization.
+    *
+    * @param method The intrinsic method being checked.
+    *
+    * @return The corresponding IL OpCode for the intrinsic method.
+    */
+   static TR::ILOpCodes ilOpCodeForIntrinsicMethod(TR::RecognizedMethod method) { return TR::BadILOp; }// no virt, default, cast
```
fjeremic: Can we add some spacing between the function declarations. Also please remove the "no virt, default" it is not useful anymore as it was remanence of old comments used for extensible classes.

0dvictor: For sure, updated.

fjeremic:
@ymanton I'm satisfied with the contents of the PR. I'll wait for an official review from you before merging in case you have further comments.

ymanton:
LGTM.

fjeremic:
@genie-omr build all

0xdaryl:
@fjeremic : please give me an hour to review before merging.  Looking at it now.

fjeremic:
Relaunching windows build meanwhile.

@genie-omr build win

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -489,8 +489,38 @@ class OMR_EXTENSIBLE CodeGenerator
...
-   bool ilOpCodeIsSupported(TR::ILOpCodes); // no virt, default, cast
 
+   /**
+    * @brief Returns if an IL OpCode is supported by current CodeGen
+    *
+    * @param op The IL OpCode being checked.
+    *
+    * @return True if the IL OpCode is supported otherwise false.
+    */
+   static bool ilOpCodeIsSupported(TR::ILOpCodes op);
```
0xdaryl: My main nit is with the naming of these functions.  Since this is a question of the code generator, wouldn't `isILOpCodeSupported` be more appropriate?  The current name is more of a statement than a question.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -489,8 +489,38 @@ class OMR_EXTENSIBLE CodeGenerator
...
+   static TR::ILOpCodes ilOpCodeForIntrinsicMethod(TR::RecognizedMethod method) { return TR::BadILOp; }
+
+   /**
+    * @brief Returns if an intrinsic method is supported by current CodeGen
+    *
+    * @param method The intrinsic method being checked.
+    *
+    * @return True if the intrinsic method is supported otherwise false.
+    */
+   static inline bool intrinsicMethodSupported(TR::RecognizedMethod method);
```
0xdaryl: This function is named inconsistently too.  Since it answers a question, how about `isIntrinsicMethodSupported` ?

compiler/codegen/OMRCodeGenerator.hpp
0dvictor: This is an existing query and its name was not changed. Luckily, no known downstream projects use this query so that we can rename it. I've updated its name to `isILOpCodeSupported`.

compiler/codegen/OMRCodeGenerator.hpp
0dvictor: Good suggestion, I've updated its name to `isIntrinsicMethodSupported`.

0xdaryl:
@genie-omr build all

0dvictor:
All tests have passed, the changes are ready to be merged.

PR #2935, opened 2018-09-06
Fix for incorrect conversionAnalyser evaluation on X86

0dvictor:
conversionAnalyser() should use cg->evaluate() instead of calling the iload evaluator directly.


0xdaryl:
@genie-omr build win

PR #2938, opened 2018-09-06
Introduce a utility to manage real registers on X86

0dvictor:
A utility, RealRegisterManager, is introduced to manage virtual registers
that mapped with predefined real registers.

### DIFF
--- a/compiler/x/codegen/RealRegisterManager.hpp
+++ b/compiler/x/codegen/RealRegisterManager.hpp
@@ -1,0 +1,111 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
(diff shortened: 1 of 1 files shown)


0dvictor:
This is a useful helper utility separated from https://github.com/eclipse/omr/pull/1899. I separated it out so that it won't be hold out because of the comprehensive review required by #1899.

0dvictor:
It is a revised work from OpenJ9 https://github.com/eclipse/openj9/blob/e67661f65a2b2a28be0fd93f477cc67e794dc3dc/runtime/compiler/x/codegen/X86HelperLinkage.cpp#L40-L99

0xdaryl:
Can you provide more documentation for this class?  In particular, please describe why someone would want to use it, how one would use it, and what its limitations are (e.g., limited to GPRs and FPRs).  Individual functions need Doxygen comments too please.

Also, as its a stand-alone class, what are your thoughts on how this functionality can be tested?  e.g., unit test to to create one of these, add registers of different kinds, etc.

Is the first copyright date really 2000?

0dvictor:
> Can you provide more documentation for this class? In particular, please describe why someone would want to use it, how one would use it, and what its limitations are (e.g., limited to GPRs and FPRs). Individual functions need Doxygen comments too please.

Good suggestion, I've update the documentation.

> Also, as its a stand-alone class, what are your thoughts on how this functionality can be tested? e.g., unit test to to create one of these, add registers of different kinds, etc.

The code is intensively used by OpenJ9 for quite awhile, and it has been tested with #1899. I'm not sure how it can fit into existing OMR test frameworks, any suggestions?

> Is the first copyright date really 2000?

I believe so, the code is a minor modification (only name change of member functions) of OpenJ9's RealRegisterManager utility. The OpenJ9 file has copyright date from 2000.
https://github.com/eclipse/openj9/blob/e67661f65a2b2a28be0fd93f477cc67e794dc3dc/runtime/compiler/x/codegen/X86HelperLinkage.cpp#L40-L99

> Further to my last point about testing, nothing references this header file in the OMR code base so it won't actually be built into anything in OMR (CI testing is useless). We need to be able to validate it somehow.

Agreed. It is currently tested with #1899 as it was plan to be introduced as part of #1899, but I feel it is a fairly isolated utility that worth discussed separately.
If testing as part of #1899 is not sufficient, do you have any suggestions?

andrewcraik:
@0xdaryl any response to the comments from @0dvictor ?

andrewcraik:
@0xdaryl ping?

andrewcraik:
@0xdaryl ping - seems like a useful utility

PR #2939, opened 2018-09-07
Optimized atomic memory update inlining on X86

0dvictor:
1. Optimized simple atomic operations to avoid allocating register when unnecessary.
2. Change the helper inlineAtomicMemoryUpdate() to a static local function to ease C++ compilation.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -2978,20 +2978,30 @@ static TR::Register * inlineSinglePrecisionSQRT(TR::Node *node, TR::CodeGenerato
   return node->getRegister();
 }
 
-TR::Register* OMR::X86::TreeEvaluator::performSimpleAtomicMemoryUpdate(TR::Node* node, TR_X86OpCodes op, TR::CodeGenerator* cg)
+/** \brief
+ *    Generate instructions to do atomic memory update.
+ *
+ *  \param node
+ *     The tree node
+ *
+ *  \param op
+ *     The instruction op code to perform the memory update
+ *
+ *  \param cg
+ *     The code generator
+ */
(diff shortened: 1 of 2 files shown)


andrewcraik:
looks good to me

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2978,20 +2978,30 @@ static TR::Register * inlineSinglePrecisionSQRT(TR::Node *node, TR::CodeGenerato
...
+ */
+static TR::Register* inlineAtomicMemoryUpdate(TR::Node* node, TR_X86OpCodes op, TR::CodeGenerator* cg)
    {
    TR_ASSERT((!TR_X86OpCode(op).hasLongSource() && !TR_X86OpCode(op).hasLongTarget()) || TR::Compiler->target.is64Bit(), "64-bit instruction not supported on IA32");
    TR::Register* address = cg->evaluate(node->getChild(0));
-   TR::Register* value   = cg->evaluate(node->getChild(1));
-   TR::Register* result  = cg->allocateRegister();
+   TR::Register* value   = cg->gprClobberEvaluate(node->getChild(1), MOVRegReg());
 
-   generateRegRegInstruction(MOVRegReg(), node, result, value, cg);
```
vijaysun-omr: This looks like a suspicious instruction to just have spuriously. 
Are you sure there is'nt any purpose to it (not even when the type of 'value' and 'result' don't match ? e.g. could 'value' be an address type (and therefore a collected reference register or is that impossible ?

0dvictor: The purpose of this instruction was to preserve the value in Register 'value'. It was a conservative approach that preserve it regardless whether is is used later. The proposed changes use `gprClobberEvaluate()` instead of `evaluate()`, so that the `value` is preserved whenever needed. Therefore, we can safely remove this `MOVRegReg()` instruction.

vijaysun-omr:
This change only affects X86 and so it is okay for the testing done to be restricted to that platform.

PR #2940, opened 2018-09-07
Implement ILBuilder Negate Service

charliegracie:
The OMR compiler supports negate opcodes so it makes sense to
provide a Negate service on ILBuilder. This change also includes
a jitbuildertest for negating.

### DIFF
--- a/compiler/ilgen/OMRIlBuilder.cpp
+++ b/compiler/ilgen/OMRIlBuilder.cpp
@@ -993,6 +993,20 @@ OMR::IlBuilder::UnsignedConvertTo(TR::IlType *t, TR::IlValue *v)
    }
 
 TR::IlValue *
+OMR::IlBuilder::Negate(TR::IlValue *v)
+   {
+   TR::DataType dataType = v->getDataType();
+
+   TR::ILOpCodes negateOp = ILOpCode::negateOpCode(dataType);
+   TR_ASSERT(negateOp != TR::BadILOp, "Builder [ %p ] cannot negate value %d of type %s", this, v->getID(), dataType.toString());
+
+   TR::Node *result = TR::Node::create(negateOp, 1, loadValue(v));
+   TR::IlValue *negatedValue = newValue(dataType, result);
+   TraceIL("IlBuilder[ %p ]::%d is Negated %d\n", this, negatedValue->getID(), v->getID());
+   return negatedValue;
+   }
+
+TR::IlValue *
 OMR::IlBuilder::convertTo(TR::DataType typeTo, TR::IlValue *v, bool needUnsigned)
    {
    TR::DataType typeFrom = v->getDataType();
--- a/compiler/ilgen/OMRIlBuilder.hpp
+++ b/compiler/ilgen/OMRIlBuilder.hpp
@@ -254,6 +254,7 @@ public:
    TR::IlValue *UnsignedGreaterOrEqualTo(TR::IlValue *left, TR::IlValue *right);
    TR::IlValue *ConvertTo(TR::IlType *t, TR::IlValue *v);
    TR::IlValue *UnsignedConvertTo(TR::IlType *t, TR::IlValue *v);
+   TR::IlValue *Negate(TR::IlValue *v);
 
    // memory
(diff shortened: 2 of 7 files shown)


compiler/ilgen/OMRIlBuilder.cpp
```diff
@@ -992,6 +992,20 @@ OMR::IlBuilder::UnsignedConvertTo(TR::IlType *t, TR::IlValue *v)
    return convertedValue;
    }
 
+TR::IlValue *
+OMR::IlBuilder::Negate(TR::IlValue *v)
+   {
+   TR::DataType dataType = v->getDataType();
+
+   TR::ILOpCodes negateOp = ILOpCode::negateOpCode(dataType);
+   TR_ASSERT(negateOp != TR::BadILOp, "Builder [ %p ] can not negate value %d of type %s", this, v->getID(), dataType.toString());
```
Leonardo2718: So, I'm not someone who regularly points out grammatical errors, but since I happened to notice this one:

"can not" &rarr; "cannot"

compiler/ilgen/OMRIlBuilder.hpp
```diff
@@ -254,6 +254,7 @@ class IlBuilder : public TR::IlInjector
    TR::IlValue *UnsignedGreaterOrEqualTo(TR::IlValue *left, TR::IlValue *right);
    TR::IlValue *ConvertTo(TR::IlType *t, TR::IlValue *v);
    TR::IlValue *UnsignedConvertTo(TR::IlType *t, TR::IlValue *v);
+   TR::IlValue *Negate(TR::IlValue *v);
```
Leonardo2718: I can't decide whether this particular service needs some accompanying documentation. The name really says all that needs to be said. However, I don't know if this could be confused with a boolean NOT service.

Leonardo2718:
Curious, Power doesn't seem to have evaluators for `sneg` and `bneg`: https://github.com/eclipse/omr/blob/f29d15851676f9fbfb94a86289fe062ed209937e/compiler/p/codegen/TreeEvaluatorTable.hpp#L114-L119

Leonardo2718:
Looking more closely, it seems like negation should also work for floating pointer types. @charliegracie would you mind adding a couple more tests to cover these types?

charliegracie:
@Leonardo2718 and not surprising the PPC tests failed!

charliegracie:
I will add the float and double tests.

charliegracie:
@genie-omr build all

compiler/ilgen/OMRIlBuilder.cpp
charliegracie: fixed

charliegracie:
I believe I addressed all of the comments and excluded the test on PPC

charliegracie:
ping @mstoodle

ymanton:
I don't think there is such a query, but `bneg` and `sneg` should be doable with the same instruction as `ineg`.

@charliegracie can you try using the `inegEvaluator` for `bneg` and `sneg`? You can change them at the spot @Leonardo2718 pointed to:

https://github.com/eclipse/omr/blob/f29d15851676f9fbfb94a86289fe062ed209937e/compiler/p/codegen/TreeEvaluatorTable.hpp#L114-L119

On Power values have to be promoted to the full width of the register because there are no sized arithmetic or logical instructions, there's only one `neg` so it works for every size of integer as long as we make sure the upper bits of the register always correctly represent the sign.

charliegracie:
Modified the code to add in the inegEvaluator and the Negate tests seems to be working but not the CompilerTest test is failing because it was expecting those to fail lol.  I will clean that test up now :)

charliegracie:
@genie-omr build all

charliegracie:
ILBuilder Negate test has been added back to PPC, support has been enabled for b/sneg using the inegEvaluator and the compiler test which was ensuring those two opcodes were not supported have been modified to properly test they now work.

mstoodle:
I really wish there was a way to re-trigger the IP check; it's been failing to show up an annoying number of times. But the sign-off looks right to me, so I'll merge.

PR #2941, opened 2018-09-07
Provide an API to get the constant values from an ILValue

charliegracie:
I have been working on an InterpreterBuilder class for generating
an Interpreter using JitBuilder. One of the main goals is to allow
the BytecodeBuilder objects used by the InterpreterBuilder to be
consumed without any changes to generate a JIT method. When using
the VirtualMachineState hierarchy some operations require getting
the constant value out of an ILValue so that calls to VMOS Push
and Pop can index into its array of IlValues.

### DIFF
--- a/compiler/ilgen/OMRIlValue.cpp
+++ b/compiler/ilgen/OMRIlValue.cpp
@@ -120,6 +120,18 @@ OMR::IlValue::storeOver(TR::IlValue *value, TR::Block *block)
       }
    }
 
+int32_t
+OMR::IlValue::get32bitConstValue()
+   {
+   return _nodeThatComputesValue->get32bitIntegralValue();
+   }
+
+int64_t
+OMR::IlValue::get64bitConstValue()
+   {
+   return _nodeThatComputesValue->get64bitIntegralValue();
+   }
+
 void *
 OMR::IlValue::client()
    {
--- a/compiler/ilgen/OMRIlValue.hpp
+++ b/compiler/ilgen/OMRIlValue.hpp
@@ -91,6 +91,18 @@ public:
       }
 
    /**
+    * @brief returns the 32 bit constant value for this IlValue. Caller should ensure that this IlValue represents
+    * a 32 bit (or less) constant value before it calls this function or an assertion will be triggered
+    */
+   int32_t get32bitConstValue();
+
+   /**
+    * @brief returns the 64 bit constant value for this IlValue. Caller should ensure that this IlValue represents
+    * a 64 bit (or less) constant value before it calls this function or an assertion will be triggered
+    */
+   int64_t get64bitConstValue();
+
+   /**
     * @brief associates this object with a particular client object
     */
    void setClient(void *client)


charliegracie:
ping @mstoodle and @Leonardo2718

mstoodle:
@charliegracie can you please flesh out the use case for these utilities a bit more?

charliegracie:
I am not sure how to explain it.  An ILValue has a node which could be a constant load.  If it is a constant load convinces may want to get access to the constant value.  These APIs provide access to those constant values.

charliegracie:
Having these types of APIs available will help other JitBuilder utilities. Imagine if the `VMOA` had an improved API set.
1. Get(int32_t index)
2. Get(TR::IlValue *index);
3. Set(int32_t index, TR::IlValue *value);
4. Set(TR::IlValue *index, TR::IlValue *value);

APIs 2 and 4 would be documented such that the index parameter has to be a constant value so that the constant could be pulled out of the ILValue using the API I added in this PR.

Why would you want these new APIs (2 and 4) on `VMOA`? Currently JitBuilder only provides a simple way to simulate stores to an array and these values are only ever stored in the actual array if `Commit` is called. This is great for simulating a VM state when generating code for a known method. In this case the values are known integral values so the APIs can easily just work with int32_t data types. If you were generating code for an interpreter or another general purpose method  you would like not be simulating reads/writes but in fact actually handling each value immediately. In this case you would want a version of the VMOA that does not simulate the vm state but writes/reads it into/from the actual memory. I am envisioning a super class or subclass for VMOA and VMOS that works this way. If you want to be able to share code for individual bytecode handlers between an Interpreter and a compiled method the public APIs on the VMOA and VMOS hierarchy needs to match. Since the version used by the Interpreter will be reading values at runtime they will need to have APIs that use ILValues and not int32_t. Again to be able to share the implementation now the JIT version using the VMOA and VMOS directly will also need to be able to calls APIs that take ILValues.... If these APIs require ILValues for some of their APIs there needs to be a way to get the actual constant value out of the ILValue.

Also I would like to make this information available to consumers of JitBuilder so it could be used for optimizations.  Downstream projects like OpenJ9 already make use of the APIs on Node to perform language specific optimizations. I feel that it is important for consumers of the JitBuilder API to also be able to make these types of optimizations if they would like.

mstoodle:
I think this API sets a precedent I'm not particularly excited about: it assumes that the JitBuilder client is tracking which IlValue's correspond to constant values. If it's already tracking that, then it should also be able to track the constant value. Yes, it might be convenient for the IlValue to just hand back the value it already has, but I don't want the client to be able to query an arbitrary IlValue to get its "value".

I also don't think this will work for any IlValue that has been merged with another IlValue at a merge point, since I don't *believe* that _nodeThatComputesValue is updated when the merge happens (it only updates the symref needed to load the merged value).

charliegracie:
ok I do not mind not going forward with this but I will need some help in solving an otherwise unsolvable issue in my opinion.

imagine you the second simplest bytecode (simplest would probably be `nop`) you could have `dup`. The code will want to peek from the stack and push the result back onto the stack.
```
void
execute()
    {
    TR::VirtualMachineStack *stack = vmState()->_stack;
    TR::IlValue *value = stack->Top(this);
    stack->Push(this, value);
    }
```
The above code would easily be shared between an InterpreterBuilder and a JitMethodBuilder (new class I created for generating JIT'd methods to make it easier). Now lets think about a bytecode like `push_local`.
```
void
execute()
    {
    TR::VirtualMachineArray *locals = vmState()=>_locals;
    TR::VirtualMachineStack *stack = vmState()->_stack;

    TR::IlValue *value = locals->Get(this, `value from immediates`);
    stack->Push(this, value);
    }
```
Now when you look at the `push_local` code it looks like everything can be re-used but that is not true. The `value from immediates` is an integral value in the case of a JIT'd method and an actual IlValue for an InterpreterBuilder. To make things simple I spec'd a new API that both the InterpreterBuilder and JitMethodBuilder implement called `GetImmediate` that returns an IlValue. This value is used as an ILValue index for the VirtualMachineArray used for the InterpreterBuilder but it has to be an integral value so it can index into the TR::IlValue ** array of TR::IlValue representing the locals for the VirtualMachineArray used by the JitMethodBuilder.

I am open to any and all suggestions on how to do this in a different way.  I really really do not want to concede that you can not share 100% of the code for generating the Interpreter and a JIT'd method.

mstoodle:
Thanks for the additional context, @charliegracie ..... I'll noodle that for a bit :) .

charliegracie:
@mstoodle anymore thoughts?

mstoodle:
Sorry *again* for the looooooooooooooooooooooooooooooooooooooooooooooooooooong delay in responding.

So the model I had in my head for what you were doing, which is possibly also not workable because it was a mental one not actually tried, was that the BytecodeBuilder would become "stateful" in a sense and go through a sequence of maybe 3 steps (simplest case I could imagine):
   1) fetch operands to `IlValue`s
   2) perform bytecode action
   3) store `IlValue` results

To implement a `BytecodeBuilder`, you'd add fields to hold each operand needed by the "action" piece (# 2). Then you would implement two versions of # 1 (one for the interpreter to use, and one for the JIT to use) and two versions of # 3 (one for the interpreter to use, and one for the JIT to use).

More concretely, imagine `BytecodeBuilder` gets these new virtual functions:
```
   virtual void FetchOperandsForInterpreter();
   virtual void FetchOperandsForJIT();
   virtual void Execute();
   virtual void StoreResultsForInterpreter();
   virtual void StoreResultsForJIT();
```

(maybe returning `bool` to allow for problems to be encountered causing helpfully messaged bailing).

We can codify the conventions for the interpreter and JIT builders separately so that these functions must be called consistently:
```
   void buildForInterpreter() {
      FetchOperandsForInterpreter();
      Execute();
      StoreResultsForInterpreter();
   }

  void buildForJIT() {
     FetchOpernadsForJIT();
     Execute();
     StoreResultsForInterpreter();
   }
```

I know what you're thinking: crap, I don't have just one implementation anymore! But *most* of the actual semantic code is the same, and the remaining code should be pretty mechanical and with a bit more thought we may be able to make it easier for that part to be constructed with less manual effort.

Thoughts? Really sorry again for how unresponsive I have been here...

charliegracie:
My thoughts are if it is mechanical and easy to hide in the APIs then why force a user to do it?

Things are complicated right now and I am hoping to simplify them.  With the current MethodBuilder APIs every runtime which wants to implement a JIT needs to write the same basic loop over each bytecode using the Worklist and write almost identical code as every other JIT using JitBuilder.  Moving a lot of the code into the InterpreterBuilder and new JitMethodBuilder class which are sublcasses of the new RuntimeBuilder class which relies on the abliity to hide the Operand differences from the user very much simplifies building an Interpreter and/or a JIT.  To implement both you just need to fill in a small number of functions (ie what is the bytecode handler for this bytecode index and a few others) and you have a fully working Interpreter and/or JIT. 

If you look at the sample Interpreter I have [here](https://github.com/charliegracie/omr/tree/interpreterbuilder/jitbuilder/release/src/interpreter) and in particular the [DupBuilder](https://github.com/charliegracie/omr/blob/interpreterbuilder/jitbuilder/release/src/interpreter/DupBuilder.cpp) you will easily see that if the builder has to deal implement 2 new APIs for handle Operands it will more than double the amount of code required.

I am not sure what the right answer is but I feel like writing more code than needed is a big turn off.  If I look at the OpenJ9 JIT it uses the constant value of Nodes all over the place and I think once JitBuilder moves forward and JitBuilder specific OPTs are added this type of feature will become  necessity and not just a nice to have.

I do not want to push the code base into supporting APIs that do not make sense but I do not see an easy way to write the code I am trying to write without a bunch of new support.... This is the least `crazy` thing my Interpreter builder does so I am scared to even open the next PRs.  

Can you look through the changes in the interpreter_builder branch I mentioned above and maybe make suggestions there?  If users have to basically write 2 different pieces of code for the Interpreter and the JIT I doubt that is a great answer.

mstoodle:
well, it was a thought experiment "in progress" ... I guess i was hoping we can iterate on moving forward. I'll take a closer look at the branches to see if I can imagine a different way forward.

mstoodle:
Maybe we could implement one, two, three argument handlers and one (maybe even two?) result handlers automatically. So in `execute()` you could reference `operand1()`, `operand2()`, and `operand3()` (and set some field in the constructor to say how many you need) and store results to `saveResult1(TR::IlValue *v)`, `saveResult2(TR::IlValue *v)` and set some field in the constructor to say how many results you need). You could even create "degenerate" / "convenience" subclasses to expand out the combinations (`_op1_none`, `_op1_res1`, `_op2_res1`, etc.) so clients would only need to choose the appropriate class and then write their `execute()` .

Some macros along similar lines to the DEFINE_STRUCT and DEFINE_FIELD ones could help reduce the work when you need to go further outside the predefined boxes....

How does that sound?

mstoodle:
In fact, it strikes me that a lot of boilerplate in the construction of these C++ classes could be hidden behind some convenience macros. Since most of that boilerplate isn't code that's very interesting from a debugging perspective, I think we could make the typical debugging experience still reasonable while dramatically reducing the amount of code written. You might even be able to get it down to having all the bytecode bodies being described in one file....

mstoodle:
Just imagine the savings in copyright header de-duplication alone ;)

mstoodle:
Ok, that works best for arguments that are on the stack; i wasn't considering the fields from an instruction (which is what originally motivated this PR!). That's harder to imagine "automating" though maybe we could still reduce the boilerplate...(he said, hopefully). Maybe some simple cases could still be "automated" (Int16OperandFollowingBytecode) (yes: absolutely awful name) ...

mstoodle:
Oh, and one clarification on your comment: "If I look at the OpenJ9 JIT it uses the constant value of Nodes all over the place"

Every single one of those will be looking at a node with an appropriate constant opcode and there are asserts if you ask for the wrong kind of constant or for the constant value of a node that doesn't have a constant opcode. There are comprehensive analysis passes (e.g. value propagation) that determine whether the value that reaches some other kind of node is, in fact, a constant and will replace the node with the constant value. But that is not safe to do at the time JitBuilder is operating, in particular because you don't know at the time a value is consumed whether some other producer for that value will be later produced and directed to the consuming operation (via a control flow merge).

charliegracie:
I believe 100% of the locations where I am using this on an ILValue I know it is a constant and the same asserts above will trigger if I am wrong (I know because they triggered when I was wrong). I think one of the major issues here is that these files that are only used for JitBuilder are included in the OMR compiler. I believe they are there so they can `cheat` and access APIs and members that would not normally be visible. I believe these files should be in the `jitbuilder/ilgen` folder the same as `JBIlGeneratorMethodDetails.[c,h]pp`. If files like IlValue moved into that folder it would force the ability for me to be able to make the calls I need to make due to the restructuring that would have to happen.

At this point my InterpreterBuilder code is so out of date it will be almost as much work to re-write it from scratch as it would be to rebase it. Also I do not see a path forward with my current implementation so I am closing the active PRs and I will revisit the idea in the future again if I have the time.

mstoodle:
The implementation 100% "cheats" because it's implementing a compiler and needs to know all the details. That's why we did the "client API" work recently: to stop exposing those "cheats" to clients of JitBuilder. Those files could "cheat" just as easily if they were in the `jitbuilder` directory, but they're in the `compiler` directory so they can be more easily reused and extended by other projects.

I just don't like creating a new API through that fundamentally sometimes works and sometimes doesn't, particularly since what determines when it works is very much a compiler internals kind of thing which JitBuilder is supposed to be abstracting away from. I'm sorry you felt that giving up was the only possible option.

PR #2942, opened 2018-09-07
Simplify NaN normalization fbits2i

dchopra001:
fbits2iEvaluator is used for type-coercion (float to int bits). The evaluator normalizes NaN and infinity values if needed. Since the evaluator assumes that floating point data conforms to the IEEE-754 standard, some simplification can be done here. Mainly, the simplification is that infinity values don't need to be normalized as there is only one possible representation for them anyway.


dchopra001:
Tagging @fjeremic for review

fjeremic:
@genie-omr build zlinux,zos

PR #2943, opened 2018-09-07
Add bcLength to BytecodeBuilder

charliegracie:
BytecodeBuilders have an bcIndex which describes the index for the
bytecode. Similarly a bytecode should be able to describe how many
pc indexes each bytecode represents.

### DIFF
--- a/compiler/ilgen/BytecodeBuilder.hpp
+++ b/compiler/ilgen/BytecodeBuilder.hpp
@@ -29,8 +29,8 @@ namespace TR
    class BytecodeBuilder : public OMR::BytecodeBuilder
       {
       public:
-         BytecodeBuilder(TR::MethodBuilder *methodBuilder, int32_t bcIndex, char *name=NULL)
-            : OMR::BytecodeBuilder(methodBuilder, bcIndex, name)
+         BytecodeBuilder(TR::MethodBuilder *methodBuilder, int32_t bcIndex, char *name=NULL, int32_t bcLength=-1)
+            : OMR::BytecodeBuilder(methodBuilder, bcIndex, name, bcLength)
             { }
          void initialize(TR::IlGeneratorMethodDetails * details,
                            TR::ResolvedMethodSymbol     * methodSymbol,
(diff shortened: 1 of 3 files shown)


charliegracie:
Ping @mstoodle.

mstoodle:
Sorry for the reviewing delay @charliegracie ; I've been sick for a few days.

It introduces an awkward parameter ordering on the constructor, but I know you did it to avoid breaking compatibility. `bcLength` won't be used by BytecodeBuilder at all, it's just carrying information along. I wonder if some of the functionality you're creating for `InterpreterBuilder` would make sense as subclasses rather than adding seemingly unused things into the existing classes? It would give you more flexibility to introduce new things.

It's just a thought; I'm still noodling this one.

charliegracie:
I believe BytecodeBuilder is doing more than carrying it a long.  Right now ALL users of BytecodeBuilder have to write the exact same loop to generate a method. As part of my InterpreterBuilder I am introducing another class (which needs a better name) JitMethodBuilder which significantly simplifies creating a JIT'd method for a runtime and it also uses this new bcLength... I had a subclass at one point (not sure what to call it) but I did not think it made sense.  BytecodeBuilder was introduced to make it easier to generated JIT'd methods (added the work list, etc) and I see this just as an extension of that....

I guess I could just open the giant PR with 15+ commits to make it easier to follow why I am making these changes but I thought opening the small PRs would make it easier to review. Also having the smaller PRs keeps the different discussion points located in a single spot so different "threads" do not get intermixed... let me know if you would prefer that I close these smaller PRs and just open the big PR that has all of the changes.

Leonardo2718:
Perhaps one big PR is not needed, but I think there should be an issue describing this work and all the needed changes.

charliegracie:
Here is a link to the [Issue](https://github.com/eclipse/omr/issues/2988)

charliegracie:
@mstoodle any more thoughts?

mstoodle:
Nothing platform-specific about this change, so the auto testing should be adequate...Merging....

Issue #2885, opened 2018-08-22
Segmentation fault detected when running ddrgen


ChengJin01:
Segmentation fault occurred in ddrgen when compiling the latest version of OpenJDK11/OpenJ9/OMR with VS2017.

```
../ddrgen --blob ../j9ddr.dat --superset ../superset.dat --blacklist blacklist --macrolist ../mac    roList --overrides overrides --show-empty  \
        ../j9ddr_misc29.pdb ../j9gc29.pdb ../j9jvmti29.pdb ../j9prt29.pdb ../j9shr29.pdb ../j9thr    29.pdb ../j9trc29.pdb ../j9vm29.pdb ../jclse11_29.pdb
make[5]: *** [run_omrddrgen.mk:73: ../j9ddr.dat] Segmentation fault
make[5]: Leaving directory '/cygdrive/c/temp1/openj9-openjdk-jdk11/build/windows-x86_64-normal-se    rver-release/vm/ddr'
```
It worked good last month after fixing all issues with OpenJDK11 on Windows.

I am not too sure whether the problem was caused by recent changes merged in omr/ddr

@keithc-ca,  @DanHeidinga

keithc-ca:
I was able to reproduce this. It fails because it can't resolve typedefs like `omr_error_t` immediately.
The `PostponedType` mechanism handles delayed resolution of field types, but `PdbScanner::createTypedef` also want to copy the type size into the typedef. Perhaps another (later) pass can update sizes of typedefs.

keithc-ca:
@ChengJin01 This should be fixed by #2944: please close if you agree.

PR #2944, opened 2018-09-07
Enable DDR to work with VS2017

keithc-ca:
Only add enum literals to enum types

* Consider the source pattern `typedef enum E { Ea, Eb } E;`: With Visual Studio 2017, the PDB description for the typedef may precede that of the enum. Thus `_typeMap["E"]` will refer to a TypedefUDT object which is not safe to cast to an EnumUDT.

Minor cleanup

* Be careful capturing the size of a typedef: The call to setType() isn't guaranteed to set _aliasedType immediately so it's not safe to dereference it. Typedef types are omitted from the generated blob and superset files so we don't really need to capture the size of typedef types.

* Explicitly construct base types and initialize all fields. The compiler ought to do the same implicitly, but there's no cost to being explicit.

* Use Modifiers::addArrayDimension. This is preferable to accessing the field _arrayLengths directly.

* Explicitly mark virtual destructors. If a class has a virtual destructor, then destructors of derived classes should be implicitly virtual as well. Being explicit makes the code easier to understand.

* Add Type::getTypeNameKey(). Refactor the static function from Symbol_IR.cpp; it's still only used by Symbol_IR at this time. Type::getFullName() was made a 'const' method so getTypeNameKey() can be const.

### DIFF
--- a/ddr/include/ddr/ir/Field.hpp
+++ b/ddr/include/ddr/ir/Field.hpp
@@ -37,6 +37,7 @@ public:
 	bool _isStatic;
 
 	Field();
+	virtual ~Field();
 
 	const std::string &getTypeName() const;
 };
--- a/ddr/include/ddr/ir/Members.hpp
+++ b/ddr/include/ddr/ir/Members.hpp
@@ -30,6 +30,7 @@ class Members
 public:
 	std::string _name;
 
+	Members() : _name() {}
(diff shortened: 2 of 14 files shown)


keithc-ca:
This fixes #2885.

charliegracie:
@genie-omr build all

PR #2947, opened 2018-09-09
Modify DeadTreesElimination to support rdbars

cathyzhyi:
A treetop hanging a read barrier can be removed if there is no other
real use of the read barrier. Otherwise, the treetop has to be kept to
anchor the read barrier.
          
This change contains 2 parts:
                 
The first part changes trivial DeadTreesElimination to remove the
first treetop from trees in the following form:
<pre>
treetop   
   xrdbari  
anchor    
   =>xrdbari
</pre>

The second part prevents rdbar hanging under treetop with referenceCount > 1 from being removed.

isssue #2812 

### DIFF
--- a/compiler/optimizer/DeadTreesElimination.cpp
+++ b/compiler/optimizer/DeadTreesElimination.cpp
@@ -141,6 +141,11 @@ static bool fixUpTree(TR::Node *node, TR::TreeTop *treeTop, TR::NodeChecklist &v
    return containsFloatingPoint;
    }
 
+static inline bool isReadBarreirUnderTreetop(TR::Node *node)
+   {
+   return node->getOpCodeValue() == TR::treetop && node->getFirstChild()->getOpCode().isReadBar();
+   }
+
 bool collectSymbolReferencesInNode(TR::Node *node,
                                    TR::SparseBitVector &symbolReferencesInNode,
                                    int32_t *numDeadSubNodes, vcount_t visitCount, TR::Compilation *comp,
@@ -480,6 +485,21 @@ int32_t TR::DeadTreesElimination::performOnBlock(TR::Block *block)
    return 0;
    }
 
+typedef std::pair<ncount_t const, TR::TreeTop* > ReadBarToTreeTopMapEntry;
+typedef TR::typed_allocator<ReadBarToTreeTopMapEntry, TR::Region &> ReadBarToTreeTopMapAlloc;
+typedef std::map<ncount_t, TR::TreeTop *, std::less<ncount_t>, ReadBarToTreeTopMapAlloc> ReadBarToTreeTopMap;
+
+static void findReadBarInSubTree(TR::Node *node, TR::NodeChecklist &visitedNodesInCurrentTree, TR::list<TR::Node*> &rdbarsInCurrentSubTree)
+   {
+   if (visitedNodesInCurrentTree.contains(node))
+      return;
+   visitedNodesInCurrentTree.add(node);
+   if (node->getOpCode().isReadBar())
+      rdbarsInCurrentSubTree.push_back(node);
+   for (int i = 0; i < node->getNumChildren(); i++)
+      findReadBarInSubTree(node->getChild(i), visitedNodesInCurrentTree, rdbarsInCurrentSubTree);
+   }
+
 void TR::DeadTreesElimination::prePerformOnBlocks()
    {
    _cannotBeEliminated = false;
@@ -487,13 +507,46 @@ void TR::DeadTreesElimination::prePerformOnBlocks()
 
    _targetTrees.deleteAll();
 
-   // Walk through all the blocks to remove trivial dead trees of the form
-   // treetop
-   //   => node
-   // The problem with these trees is in the scenario where the earlier use
-   // of 'node' is also dead.  However, our analysis won't find that because
-   // the reference count is > 1.
+   /*
+    * Walk through all the blocks to remove trivial dead trees in the following forms:
+    *
+    * case 1:
+    * treetop
+    *   => node
+    *
+    * case 2:
+    * treetop
+    *    xrdbari
+    * anchor
+    *    =>xrdbari
+    *
+    * The problem with these trees is in the scenario where the earlier use
+    * of 'node' is also dead.  However, our analysis won't find that because
+    * the reference count is > 1.
+    *
+    * Here are some clarification about case 2:
+    * 1. Case 2 is seen very often because ilgen creates trees in the following form:
+    *    NULLCHK
+    *      ardbari
+    *          aload
+    *    anchor
+    *      => ardbari
+    *    And the NULLCHK can be optimized away by optimizations like value propagation and
+    *    turned into a treetop.
+    * 2. We do not remove the treetop node if there is any other tree whose subtree
+    *    references to that rdbar before the anchor node, like the following:
+    *    treetop
+    *      ardbari
+    *    SOMETREE (that's not anchor nor treetop)
+    *      => ardbari
+    *    anchor
+    *      => ardbari
+    *    Because those SOMETREE would need a treetop to anchor the rdbar node as the first evaluation point.
+    *
+    */
    vcount_t visitCount = comp()->incOrResetVisitCount();
+   ReadBarToTreeTopMap rdbar2ttMap(std::less<ncount_t>(), comp()->trMemory()->currentStackRegion());
+
    for (TR::TreeTop *tt = comp()->getStartTree();
         tt != 0;
         tt = tt->getNextTreeTop())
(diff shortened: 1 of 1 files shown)


cathyzhyi:
this PR is dependent on #2875

0xdaryl:
> this PR is dependent on #2875

If that's the case, this PR should be marked as a WIP until #2875 is merged.

andrewcraik:
@cathyzhyi #2875 is merged can you rebase and un-WIP so this can be progressed?

cathyzhyi:
@andrewcraik removed the WIP and rebased on latest build. thanks!

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Needs a review from @andrewcraik

vijaysun-omr:
@andrewcraik friendly ping to review when you get a chance

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -930,34 +973,56 @@ int32_t TR::DeadTreesElimination::process(TR::TreeTop *startTree, TR::TreeTop *e
...
+       *    ...
+       *
+       * to:
+       *
+       * treetop
+       *    baseObjectNode
+       * treetop
+       *    child1
+       *    ...
+       */
```
andrewcraik: Eliminating a load indirect is not as simple as saying it has no other uses - if the symref were volatile for example it creates a happens before edge that may be required to continue to exist. This doesn't really feel like a dead trees operation - there is a valid operation here and there needs to be some analysis/thought to remove it. What is your reasoning for doing this here?

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -480,20 +485,36 @@ int32_t TR::DeadTreesElimination::performOnBlock(TR::Block *block)
...
+
+    * case 2:
+    * treetop
+    *    rdbar
+    * anchor
+    *    =>rdbar
+    *
+    * The problem with these trees is in the scenario where the earlier use
+    * of 'node' is also dead.  However, our analysis won't find that because
+    * the reference count is > 1.
```
andrewcraik: Can/does case 2 happen for aloadi as well? What situations are resulting in this? is it sloppy ilgen or are opts specifically not caring the tree is already anchored by the anchor tree and re-anchoring it?

andrewcraik:
@cathyzhyi any commnet on the questions above?

compiler/optimizer/DeadTreesElimination.cpp
cathyzhyi: Thanks for pointing this out I confused the other treetop with real use.  This part of the code won't kick in for rdbar because it implies a rdbar is hanging under 2 treetops which is the case handled by trivial deadtree elimination already. Removing rdbar handling code here.

cathyzhyi:
@andrewcraik made changes according to the comment. Thanks!

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -480,20 +485,37 @@ int32_t TR::DeadTreesElimination::performOnBlock(TR::Block *block)
    return 0;
    }
 
+typedef std::pair<TR::Node* const, TR::TreeTop* const> ReadBarToTreeTopMapEntry;
+typedef TR::typed_allocator<ReadBarToTreeTopMapEntry, TR::Region&> ReadBarToTreeTopMapAlloc;
+typedef std::map<TR::Node*, TR::TreeTop *, std::less<TR::Node*>, ReadBarToTreeTopMapAlloc> ReadBarToTreeTopMap;
```
andrewcraik: This is going to need some memory profililng - std::map can be very inefficient. Alternative could be std::deque where we index using the node global index to look up a TreeTop or an array sized for the max node global index. Profiling the memory footprint increase is certainly something that needs to be done along with the high water mark.

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -618,6 +653,17 @@ namespace
       };
    }
 
+static bool treeCanNotBeRemoved(TR::Node *node)
```
andrewcraik: This logic isn't easy to read - I think you could factor these conditions more clearly with some nice comments.

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -508,6 +530,12 @@ void TR::DeadTreesElimination::prePerformOnBlocks()
          TR::TransformUtil::removeTree(comp(), tt);
          removed = true;
          }
+      else if (node->getOpCode().isAnchor() && visitedReadbarsUnderTreeTop.isSet(node->getFirstChild()->getGlobalIndex()))
+         {
+         TR::TreeTop *ttToRemove = rdbar2ttMap[node->getFirstChild()->getGlobalIndex()];
```
andrewcraik: is there a potential bug here if you have
treetop
   rdbar
CHK1
   =>rdbar
CHK2
   => rdbar

cathyzhyi: I think it should be fine. The `rdbar2ttMap` only record nodes hanging under a `TreeTop` but not any other checks. The condition checking`if (!removed && isReadBarreirUnderTreetop(node))` when adding the rebar's treetop to the deque guarantee only the first treetop hanging the rebar will be added.

cathyzhyi: right, the condition should check for `checks` as well.

cathyzhyi: @andrewcraik Hey Andrew, I am reconsidering this comment https://github.com/eclipse/omr/pull/2947#discussion_r230922890. I think it shouldn't be a problem after all. Checks like `resolveCHK` or `nullCHK` has to be the first appearance of the first child so no treetop with the same child should appear before them and this pattern won't happen for those checks. For checks like `Checkcast` this tree pattern is possible and we do need the treetop as an anchor node.

vijaysun-omr:
@cathyzhyi is this a WIP given the last comment asking for some more perf testing ?

cathyzhyi:
> @cathyzhyi is this a WIP given the last comment asking for some more perf testing ?

yes, change it to WIP pending on memory profiling result.

andrewcraik:
@cathyzhyi any updates?

vijaysun-omr:
@genie-omr build all

compiler/optimizer/DeadTreesElimination.cpp
vijaysun-omr: I'm unclear on the last comment. It says case 2 happens when VP removes a check, however, case 2 as described in the comment talks about a check being the second reference to the rdbar. If the check was removed how did we end up in case 2 ? 

Also what exactly do you mean by "anchor" in your comment about case 2 ? 

Finally, is the following case also possible (and therefore ought to be handled):

    * anchor/check
     *    rdbar
     * treetop
     *    =>rdbar

cathyzhyi: Sry for the confusion. I should actually remove the `check` from the comment in the code for case 2. 

- The case this code catches is the following:
```
case 2:
treetop
   rdbar
anchor
   =>rdbar
```
which is the result of VP removing the NullCHK for the following case
```
NullCHK
   rdbar
compressedrefs
   =>rdbar
```

- `anchor` means `isAnchor` query returns true which in our current code means `TR::compressedRefs`

-  this case is possible 
```
anchor/check
    rdbar
 treetop
    =>rdbar
```

It falls into existing case 1 andis handled by this existing if and doesn't need extra handling https://github.com/eclipse/omr/blob/f81a86655fae3c72e47ed45de7d66e1073913416/compiler/optimizer/DeadTreesElimination.cpp#L504-L510

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@andrewcraik review approval is a prereq here before I consider merging. I still want to do another pass over the code myself.

compiler/optimizer/DeadTreesElimination.cpp
```diff
@@ -480,20 +485,37 @@ int32_t TR::DeadTreesElimination::performOnBlock(TR::Block *block)
    return 0;
    }
 
+typedef std::pair<ncount_t , TR::TreeTop* > ReadBarToTreeTopMapEntry;
```
Leonardo2718:
```suggestion
typedef std::pair<const ncount_t , TR::TreeTop* > ReadBarToTreeTopMapEntry;
```

I believe this will fix the test failure.

cathyzhyi: Thanks a lot, Leo!

Leonardo2718:
@genie-omr build all

compiler/optimizer/DeadTreesElimination.cpp
cathyzhyi: For `aloadi` this is handled in the normal dead tree elimination because for normal `aloadi` a treetop can always be removed without looking at its reference count. For `xrdbari`, we only apply normal dead tree elimination when the reference count is one (in `treeCanPossiblyBeRemoved` method). If we don't remove the first treetop in trivial dead tree elimination,  normal dead tree elimination won't kick in.

compiler/optimizer/DeadTreesElimination.cpp
andrewcraik: @cathyzhyi what about a SpineCHK - I don't think that requires the array child to be a first appearance so couldn't that create the above pattern?

compiler/optimizer/DeadTreesElimination.cpp
andrewcraik: ok so this is a result of an optimizationand not sloppy ilgen so it seems reasonable to handle this here

compiler/optimizer/DeadTreesElimination.cpp
cathyzhyi: Had an offline discussion with Andrew and added code to make the code more conservative. Do not remove the treetop if there is any other tree in between the first treetop and the anchoring treetop. something like:
treetop
=>rdbar
ANYTREE
=>rdbar
anchor
=>rdbar

Leonardo2718:
@genie-omr build all

vijaysun-omr:
@genie-omr build all

compiler/optimizer/DeadTreesElimination.cpp
andrewcraik: @vijaysun-omr I think this more conservative version of just handling the treetop and anchor pattern should be sufficient - thoughts?

vijaysun-omr: I think this is safe since it's unclear what ANYTREE assumptions may be wrt anchoring earlier.

andrewcraik:
Given the review from @vijaysun-omr on the specific question I raised - I am happy for this to merge now.

PR #2951, opened 2018-09-11
WIP: AArch64: Implement return evaluators

knn-k:
This commit implements some basic evaluators such as ireturn for
aarch64 as the starting point.
This depends on #2906.

Update


knn-k:
Marked as WIP until #2906 is merged.

knn-k:
#2906 was merged, and I removed "WIP:" prefix.  It is ready for review.

0xdaryl:
These evaluators are nearly identical except for the choice of return register, register kind, and the ReturnInfo word.  Is there any way you can common the bulk of the logic up and only parameterize based on the properties that change?  It should save some DLL size.

knn-k:
Updated files reflecting the review comment, introducing genericReturn().  Also implemented returnEvaluator() for void return.

PR #2952, opened 2018-09-11
Skip cold blocks in recognized call transformer

liqunl:
Cold blocks are blocks that haven't run or blocks generated by
optimizations that are not to be optimized furthur.

### DIFF
--- a/compiler/optimizer/OMRRecognizedCallTransformer.cpp
+++ b/compiler/optimizer/OMRRecognizedCallTransformer.cpp
@@ -38,6 +38,12 @@ int32_t OMR::RecognizedCallTransformer::perform()
    TR::NodeChecklist visited(comp());
    for (auto treetop = comp()->getMethodSymbol()->getFirstTreeTop(); treetop != NULL; treetop = treetop->getNextTreeTop())
       {
+      if (treetop->getEnclosingBlock()->isCold())
+         {
+         treetop = treetop->getEnclosingBlock()->getExit();
+         continue;
+         }
+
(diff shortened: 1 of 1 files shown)


liqunl:
@0dvictor Since you wrote RecognizedCallTransformer, can you review?

fjeremic:
Could you please give some background information on why this change is necessary? Why is it beneficial not to do this? It sounds like a potentially scary performance change in case the block frequencies are wrong.

0dvictor:
I'm not sure whether it is correct to turn it off for all cold blocks. If certain methods shouldn't not be transformed when they are in cold blocks, their corresponding `isInlineable()` should check and return false.

liqunl:
>Could you please give some background information on why this change is necessary? Why is it beneficial not to do this?

This is to avoid transforming the call multiple times if the call remains in the tree after the optimization. This happens if transformation does something similar to inliner and the resulting trees have a slow path containing the call. Unfortunately, `RecognizedCallTransformer` is an ilgen opt, @0dvictor told me that we can't add the method to the inlining table in ilgen opt, thus we can't have the inliner way to prevent the slow path being optimized, which is querying `isTheVirtualCallNodeForAGuardedInlinedCall`.

liqunl:
>I'm not sure whether it is correct to turn it off for all cold blocks. If certain methods shouldn't not be transformed when they are in cold blocks, their corresponding isInlineable() should check and return false.

Thanks for the advice, closing this PR and put the check in `isInlineable()`

PR #2956, opened 2018-09-12
Change for supporting CopyForwardHybrid

LinHu2016:
CopyForwardHybrid is hybrid mode of CopyForward Scheme that 
 GC can run both copyforward and markcompact on the collection set
 simultaneously (for balanced PGC only). During precollection, 
 GC can reserve some regions for markcompact only(non evacuated 
 regions -- such as jni critical regions) and copyforward for 
 the rest of collection set. 

 - new global flag tarokEnableCopyForwardHybrid for indicating if
 CopyForwardHybrid mode is on, default is off.
 - new global test option fvtest_forceCopyForwardHybridRatio, which 
can set the percentage of collection set to be reserved for markcompact.
 - update workpackets for allow synchronizing with the threads pull
 the work from another queue(such as copycache).
 -  add new nonEvacuateRegionCount in CopyForwardStats for verbose
 gc reporting.
 
### DIFF
--- a/gc/base/WorkStack.cpp
+++ b/gc/base/WorkStack.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2015 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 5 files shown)


charliegracie:
@genie-omr build all

gc/base/WorkStack.cpp
```diff
@@ -297,3 +297,41 @@ MM_WorkStack::flushOutputPacket(MM_EnvironmentBase *env)
...
+	if (NULL == _inputPacket) {
+		/* If the output packet contains at least a free entry - invert the input/output */
+		if((NULL != _outputPacket) && !_outputPacket->isEmpty()) {
+			/* swap the input packet with the output packet */
+			_inputPacket = _outputPacket;
+			_outputPacket = NULL;
+			env->_workPacketStats.workPacketsExchanged += 1;
+		}
+	}
+	return (NULL != _inputPacket);
```
amicic: this pattern is repeated a couple of more times  in this file. perhaps you could reuse this method?

charliegracie:
@genie-omr build all

gc/base/WorkStack.cpp
LinHu2016: done

LinHu2016: expose and use WorkPackets' monitor and waitCount for CopyForwardScheme scan queue synchronization

charliegracie:
@amicic has @LinHu2016 addressed all of your concerns?

amicic:
this change now looks good

charliegracie:
@genie-omr build all

PR #2959, opened 2018-09-12
Add a transform util to create alternate path for existing operations

liqunl:
The transformation is:

          elseBlock               ifBlock
              |                   /     \
             ...      ===>       /       \
              |               elseBlock  thenBlock
          mergeBlock             |        /
                                ...      /
                                  \     /
                                   \   /
                                 mergeBlock

### DIFF
--- a/compiler/optimizer/OMRTransformUtil.cpp
+++ b/compiler/optimizer/OMRTransformUtil.cpp
@@ -394,3 +394,56 @@ OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::No
    node->removeAllChildren();
    node = TR::Node::recreateWithoutProperties(node, TR::PassThrough, 1, child);
    }
+
+void
+OMR::TransformUtil::createConditionalAlternatePath(TR::Compilation* comp,
+                                                   TR::TreeTop *ifTree,
+                                                   TR::TreeTop *thenTree,
+                                                   TR::Block* elseBlock,
+                                                   TR::Block* mergeBlock,
+                                                   TR::CFG *cfg,
+                                                   bool markCold)
+   {
+   cfg->setStructure(0);
+
+   TR::Block* ifBlock = elseBlock;
+   ifBlock->prepend(ifTree);
+   elseBlock = ifBlock->split(ifTree->getNextTreeTop(), cfg, false /*fixupCommoning*/, true /*copyExceptionSuccessors*/);
+
+   TR::Block * thenBlock = TR::Block::createEmptyBlock(thenTree->getNode(), comp, 0, elseBlock);
+   if (markCold)
+      {
+      thenBlock->setFrequency(UNKNOWN_COLD_BLOCK_COUNT);
+      thenBlock->setIsCold();
+      }
+   else
+      {
+      thenBlock->setFrequency(elseBlock->getFrequency());
+      }
+
+   cfg->addNode(thenBlock);
+
+   TR::Block *cursorBlock = mergeBlock;
+   while (cursorBlock && cursorBlock->canFallThroughToNextBlock())
+      {
+      cursorBlock = cursorBlock->getNextBlock();
+      }
+
+   if (cursorBlock)
+      {
+      TR::TreeTop *cursorTree = cursorBlock->getExit();
+      TR::TreeTop *nextTree = cursorTree->getNextTreeTop();
+      cursorTree->join(thenBlock->getEntry());
+      thenBlock->getExit()->join(nextTree);
+      }
+   else
+      cfg->findLastTreeTop()->join(thenBlock->getEntry());
+
+
+   thenBlock->append(thenTree);
+   thenBlock->append(TR::TreeTop::create(comp, TR::Node::create(thenTree->getNode(), TR::Goto, 0, mergeBlock->getEntry())));
+   ifTree->getNode()->setBranchDestination(thenBlock->getEntry());
+   cfg->addEdge(TR::CFGEdge::createEdge(thenBlock, mergeBlock, comp->trMemory()));
+   cfg->addEdge(TR::CFGEdge::createEdge(ifBlock, thenBlock, comp->trMemory()));
+   cfg->copyExceptionSuccessors(elseBlock, thenBlock);
+   }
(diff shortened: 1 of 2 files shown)


mstoodle:
Can you please add more detail to the commit title? "Add an API" is pretty information free :) .

Also, just in the PR comments, can you give a concrete example where such an API is useful? Not doubting it is, i just want to get a good example in my head.

compiler/il/OMRBlock.cpp
```diff
@@ -1143,6 +1143,95 @@ OMR::Block::createConditionalBlocksBeforeTree(TR::TreeTop * tree,
    return remainderBlock;
    }
 
+/**
```
Leonardo2718: Could you add a `\brief`, "briefly" describing what this function does?

compiler/il/OMRBlock.cpp
```diff
@@ -1143,6 +1143,95 @@ OMR::Block::createConditionalBlocksBeforeTree(TR::TreeTop * tree,
...
+ *     start block_D
+ *        ifTree
+ *        goto the only successor of block_A
+ *     end block_D
+ *
+ *   \param this          the block to be used as the else block
+ *   \param compareTree   the compare tree to use in constructing the conditional blocks and cfg
+ *   \param ifTree        the tree to use on the 'if' branch when constructing the blocks and cfg
+ *   \param cfg           the cfg to update with the new information
+ *   \parm markCold       whether to mark the if block cold
```
Leonardo2718: For completeness, please add `\parm changeBlockExtensions`.

compiler/il/OMRBlock.cpp
```diff
@@ -1143,6 +1143,95 @@ OMR::Block::createConditionalBlocksBeforeTree(TR::TreeTop * tree,
    return remainderBlock;
    }
 
+/**
+ * calling this function on block_A will convert
```
Leonardo2718: Could you add a note explaining that block_A is assumed to only have one successor (and maybe why this is the case)? For example `calling this function on block_A, which must have only one successor, will convert`.

Although the assumption is mentioned later on, I think that calling it out explicitly earlier on makes it easier to understand.

0xdaryl:
Perhaps to @mstoodle's point, is this a common-enough pattern that we'd expect it to be useful to other Block users?  Just wondering aloud whether this should just be part of your optimization, or even a transformation utility (in the TransformUtil class).

liqunl:
>is this a common-enough pattern that we'd expect it to be useful to other Block users?

This won't be very common. I put it in Block because of the nature of the code, thinking that someone else might find it useful or we might use it in the future. My optimization is in Inliner, I don't think this API belongs there. I can move it TransformUtil if that's more acceptable for you @mstoodle @Leonardo2718 @0xdaryl

0xdaryl:
That seems fine by me.

liqunl:
@mstoodle @Leonardo2718 Updated the commit, could you continue the review?

liqunl:
@mstoodle Done

mstoodle:
genie-omr build all

liqunl:
The CI build failure is
```
../compiler/codegen/OMRCodeGenerator.cpp:138:1:error:static assertion failed: NodeToInstrEvaluators is not the correct size
static_assert(TR::NumIlOps ==
^~~~~~~~~~~~~
```
Note related to this commit.

liqunl:
Rebased the commit to pick up https://github.com/eclipse/omr/pull/2875 which is causing the Travis CI build failure.

compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -394,3 +394,57 @@ OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::No
...
+                                                   TR::Block* mergeBlock,
+                                                   TR::CFG *cfg,
+                                                   bool markCold)
+   {
+   cfg->setStructure(0);
+
+   TR::Block* ifBlock = elseBlock;
+   ifBlock->prepend(ifTree);
+   elseBlock = ifBlock->split(ifTree->getNextTreeTop(), cfg, false /*fixupCommoning*/, true /*copyExceptionSuccessors*/);
+   elseBlock->setIsExtensionOfPreviousBlock(true);
```
vijaysun-omr: Not every optimization can cope with extended basic blocks in the compiler. What happens if this routine is called before block extension can be tolerated given that the above line sets "elseBlock" to be an extension of the previous block and does not fix up commoning.

vijaysun-omr: I see. So no nodes were actually being kept commoned across the 2 blocks anyway.

liqunl: Yes

vijaysun-omr:
I am fine with the changes now that the block extension is'nt being done. I will merge this once checks pass.

compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -394,3 +394,56 @@ OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::No
...
    node = TR::Node::recreateWithoutProperties(node, TR::PassThrough, 1, child);
    }
+
+void
+OMR::TransformUtil::createConditionalAlternatePath(TR::Compilation* comp,
+                                                   TR::TreeTop *ifTree,
+                                                   TR::TreeTop *thenTree,
+                                                   TR::Block* elseBlock,
+                                                   TR::Block* mergeBlock,
+                                                   TR::CFG *cfg,
```
fjeremic: Out of curiosity why do we need to provide a CFG here? Is it possible for a CFG to be "rogue" and live outside the `comp` object? Or to have multiple CFGs? I'm wondering if this parameter could be removed and just extract the CFG from the compilation?

liqunl: Good point. I was using the existing Block API as a template when writing this one, didn't think of why we need a CFG as a parameter.

Looking at the usage of `OMR::Block::createConditionalBlocksBeforeTree`, we use it in openj9's ilgen where we pass the inlined method's CFG, which is different than the top level method's CFG.

liqunl:
A summary of the build. There was only one failure, which is a compilation failure on aarch64, caused by https://github.com/eclipse/omr/pull/2875, and has been fixed by https://github.com/eclipse/omr/pull/2975. I rebased the commit this morning to pick up the fix. The util is not used in OMR but I tested it in openj9.

compiler/optimizer/OMRTransformUtil.cpp
fjeremic: Thanks, did not think about the inlined method CFG case. No further comments from me.

PR #2965, opened 2018-09-14
Add TR_J2IVirtualThunkPointer relocation kind

jdmpapin:
This kind can be used in downstream projects to describe a relocation
for JIT-compiled code that contains a pointer to a code sequence that
transitions control into an interpreter.


jdmpapin:
@irinarada FYI

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Could we please update the description to avoid the downstream project reference in this case ?

jdmpapin:
I've updated the commit message

vijaysun-omr:
All checks passed (apart from appveyor that is still waiting). 
Given the simple nature of the change, I am merging this now that the commit message has been updated.

PR #2966, opened 2018-09-14
Fix overloaded virtual function hidden error

babsingh:
While compiling **OpenJ9** with **Clang 9.1.0** on **OSX**, the following error was
seen:

```
In file included from MemorySubSpaceMetronome.cpp:39:
./MemorySubSpaceMetronome.hpp:69:8: error:
'MM_MemorySubSpaceMetronome::collect' hides overloaded virtual function
[-Werror,-Woverloaded-virtual]
        void collect(MM_EnvironmentBase *env, MM_GCCode gcCode);
             ^
../omr/gc/base/segregated/MemorySubSpaceSegregated.hpp:92:15: note:
hidden overloaded virtual function
'MM_MemorySubSpaceSegregated::collect' declared here: type mismatch at
2nd parameter ('MM_AllocateDescription *' vs 'MM_GCCode')
        virtual void collect(MM_EnvironmentBase *env,
MM_AllocateDescription *allocDescription);
```
 
The definition of `collect` differs in **MemorySubSpaceSegregated** (base
class) and **MemorySubSpaceMetronome** (derived class). This leads to the
above error.

In the base class, the virtual function `collect` doesn't get used. In
order to fix the above error, the unused definition of `collect` in the base class
has been removed.

### DIFF
--- a/gc/base/segregated/MemorySubSpaceSegregated.cpp
+++ b/gc/base/segregated/MemorySubSpaceSegregated.cpp
@@ -212,13 +212,6 @@ MM_MemorySubSpaceSegregated::largestDesirableArraySpine()
 #endif /* defined(OMR_GC_ARRAYLETS) */
 
 void
-MM_MemorySubSpaceSegregated::collect(MM_EnvironmentBase *env, MM_AllocateDescription *allocDescription)
-{
-	Assert_MM_unreachable();
-}
-
-
-void
 MM_MemorySubSpaceSegregated::abandonHeapChunk(void *addrBase, void *addrTop)
 {
 }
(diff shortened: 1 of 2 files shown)


dmitripivkine:
I believe this is correct fix if we need to keep collect() method in MM_MemorySubSpaceSegregated.
An alternative solution might be this method removal

DanHeidinga:
> I believe this is correct fix if we need to keep collect() method in MM_MemorySubSpaceSegregated.
> An alternative solution might be this method removal

@dmitripivkine which solution do you prefer?  Does collect() have to stay in MM_MemorySubSpaceSegregated?

dmitripivkine:
I would remove unused function from OMR code. However it is up to Charlie to decide

youngar:
@DanHeidinga @babsingh After talking with Dmitri, we think its better to delete the function, no point in keeping dead code around.

babsingh:
@youngar @dmitripivkine I have updated the PR to remove the unused `collect` method.

youngar:
@genie-omr build all

youngar:
@babsingh I just realized we forgot to update the copyright headers.  Would you mind opening a PR to fix that?

PR #2967, opened 2018-09-14
Use logical comparison in minEvaluator

dchopra001:
When comparing long values in register pairs, only the high word of the registers should do a signed comparison. The low words should be evaluated logically as the high word comparison would have indicated if there was a sign difference between the two.

### DIFF
--- a/compiler/z/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/z/codegen/ControlFlowEvaluator.cpp
@@ -946,7 +946,7 @@ static TR::Register * maxMinHelper(TR::Node *node, TR::CodeGenerator *cg, bool i
          cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRNE, node, done);
          cursor->setStartInternalControlFlow();
 
-         generateRRInstruction(cg, TR::InstOpCode::CR, node, registerA->getLowOrder(), registerB->getLowOrder());
+         generateRRInstruction(cg, TR::InstOpCode::CLR, node, registerA->getLowOrder(), registerB->getLowOrder());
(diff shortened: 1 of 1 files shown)


fjeremic:
@genie-omr build zlinux,zos

dchopra001:
I'm going to add an example case here that illustrates the problem this commit is trying to solve. 

Consider the following scenario (where `long` values are 64-bit integers):

```
long Value1 = 2147483648;
long Value2 = 2147483647;
```

On a 64-bit compiler, these will look as follows in registers:
Value1 = 0x0000000080000000
Value2 = 0x000000007FFFFFFF

The evaluator here will do a comparison using a `CGR` instruction, it will tell us that Value1 is larger. This is correct. 

Now consider that we are on a 32-bit compiler. In this case, we place the high word and low words for each value in a different register. Example:

Value1:
r0 = 0x00000000
r1 = 0x80000000

Value2:
r2 = 0x00000000
r3 = 0x7FFFFFFF

In this scenario, must compare the two high words to determine which value to pick. However, in the event that both the high words are equal, we must compare the low words to determine which value is bigger or smaller. 

The minEvaluator currently will generate a `CR` instruction to compare r1 and r3 (the low words). The `CR` instruction treats its operands as signed 32-bit integers. Based on this, it would say that r1 is smaller than r3 (because r1 would be a negative number). **This is obviously incorrect**. We cannot view R1 as a separate 32-bit value. It is a part of a 64-bit value.

The fix here would be to do a logical comparison (`CLR`) instead of a signed comparison for the low word. We only need to do a signed comparison for the high word.

fjeremic:
Travis build failed due to some certificate issue and AppVeyor is not necessary as this is a Z only change. Merging.

PR #2971, opened 2018-09-17
Add an option to disable guarded static final field folding

liqunl:


liqunl:
@andrewcraik Can I have a review? I'll add code to set the option bit in openj9

andrewcraik:
I'm fine with adding this option - usage will be reviewed in OpenJ9 where it is being used for the time being.

0xdaryl:
@genie-omr build all

liqunl:
CI build failure on aarch64 is caused by https://github.com/eclipse/omr/pull/2875, fixed by https://github.com/eclipse/omr/pull/2975. Rebased to pick up the fix.

PR #2973, opened 2018-09-18
Update name of TR_referenceArrayCopy

0dvictor:
The name of TR_referenceArrayCopy is renamed as jitReferenceArrayCopy,
to matching the naming scheme of other JIT helpers.


fjeremic:
TravisCI failures on AArch due to read barrier IL which as fixed in #2975. Merging.

PR #2974, opened 2018-09-18
Support load effective address of memory on X86

0dvictor:
Expand insertLoadMemory() to return effective address instead of content.


vijaysun-omr:
Could you please elaborate on "why" you are making this change ?

0dvictor:
> Could you please elaborate on "why" you are making this change ?

During the development of OpenJ9, I found there are some situations that we want to get the address of a field instead of loading the content of the field. I guess this may be demanded sometime in other downstream projects.

vijaysun-omr:
So you are making the change to also allow a LEA to be generated instead of a load but are looking to reuse some of this code in that case.

vijaysun-omr:
@0xdaryl please review.

0xdaryl:
@genie-omr build xlinux,win

PR #2977, opened 2018-09-18
Improvements for Relocatable Compilations Part 1

dsouzai:
This PR introduces changes in OMR required to enable optimizations that are currently implicitly disabled for relocatable compilations.

Issue tracking all the PRs: https://github.com/eclipse/openj9/issues/2921


fjeremic:
Could we please get a high level summary of the changes in the PR description?

dsouzai:
@mstoodle @jdmpapin could you please review?

compiler/optimizer/VPConstraint.cpp
```diff
@@ -1293,6 +1293,8 @@ TR::VPKnownObject *TR::VPKnownObject::create(OMR::ValuePropagation *vp, TR::Know
    {
    TR::KnownObjectTable *knot = vp->comp()->getKnownObjectTable();
    TR_ASSERT(knot, "Can't create a TR::VPKnownObject without a known-object table");
```
jdmpapin: In light of the newly added conditional just below, I was going to say that this assertion is now bogus. But on second thought I think maybe we should keep the assertion, and push the null test out into the callers. The caller claims to have passed us a known object index, which is nonsensical in the absence a known object table.

compiler/control/OMROptions.cpp
```diff
@@ -2215,6 +2215,19 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
             if (_coldUpgradeSampleThreshold == TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD)
                _coldUpgradeSampleThreshold = 2;
             }
+
+         // disable DelayRelocationForAOTCompilations feature because with higher
+         // method counts, the JIT collects enough IProfiler info prior to
+         // compilation that it doesn't need to wait any longer before running the
```
jdmpapin: I think the end of this sentence was accidentally

compiler/control/OMROptions.cpp
```diff
@@ -2215,6 +2215,19 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
...
+
+         if (self()->getOption(TR_UseHigherMethodCounts))
+            {
+            self()->setOption(TR_DisableDelayRelocationForAOTCompilations, true);// If scount has not been changed on the command line, adjust it here
+            if (self()->getInitialSCount() == TR_INITIAL_SCOUNT)
```
jdmpapin: Not something that has to be part of this change, but it strikes me that it would be good to have a way to determine whether a value was set explicitly or defaulted. At the moment, we'll second guess the user's `scount` if they pass `-Xjit:scount=20,useHigherMethodCounts` (with the current value of 20 for `TR_INITIAL_SCOUNT`)

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -2280,6 +2280,7 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
          case TR_DirectMethodGuard:
          case TR_NonoverriddenGuard:
          case TR_InterfaceGuard:
+         case TR_AbstractGuard:
          case TR_MethodEnterExitGuard:
          case TR_HCRGuard:
          //case TR_AbstractGuard:
```
jdmpapin: Should we delete this one? Or uncomment here instead of adding it above?

vijaysun-omr:
@mstoodle and @mpirvu your reviews would be good too.

dsouzai:
@mstoodle friendly reminder for review :)

compiler/compile/OMRCompilation.hpp
```diff
@@ -351,6 +351,8 @@ class OMR_EXTENSIBLE Compilation
    bool compilationShouldBeInterrupted(TR_CallingContext) { return false; }
 
    void verifySymbolHasBeenValidated(void *symbol) { return; }
+   void incrementHeuristicRegion() { return; }
```
mstoodle: Could you add a description of what a "heuristic region" is?

compiler/x/amd64/codegen/OMRMemoryReference.cpp
```diff
@@ -400,10 +400,26 @@ OMR::X86::AMD64::MemoryReference::addMetaDataForCodeAddressWithLoad(
          if (sr.getSymbol()->isStatic())
             {
             if (cg->needClassAndMethodPointerRelocations())
-               cg->addExternalRelocation(new (cg->trHeapMemory()) TR::ExternalRelocation(displacementLocation, (uint8_t *)srCopy,
-                                                                                         (uint8_t *)(uintptr_t)containingInstruction->getNode()->getInlinedSiteIndex(),
-                                                                                         TR_ClassAddress, cg),__FILE__, __LINE__,
-                                                                                         containingInstruction->getNode());
+               {
+               if (cg->comp()->getOption(TR_UseSymbolValidationManager))
```
mstoodle: Wouldn't these relocation changes also be needed for other platforms?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -1665,13 +1696,27 @@ TR::X86RegImmSymInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
...
             setReloKind(TR_RamMethod);
+         symbolKind = TR::SymbolType::typeMethod;
          // intentional fall-through
       case TR_ClassPointer:
-         cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
-                                                           (uint8_t*)getNode(),
-                                                           (TR_ExternalRelocationTargetKind)getReloKind(),
-                                                           cg()),
-                                __FILE__, __LINE__, getNode());
+         if (symbolKind == TR::SymbolType::typeClass && cg()->comp()->getOption(TR_UseSymbolValidationManager))
```
mstoodle: why only typeClass? are you intentionally not handling typeMethod with the SymbolManager?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -1524,13 +1540,27 @@ TR::X86RegImmInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
...
          case TR_RamMethod:
+            symbolKind = TR::SymbolType::typeMethod;
             // intentional fall-through
          case TR_ClassPointer:
-            cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
-                                                              (uint8_t*)getNode(),
-                                                              (TR_ExternalRelocationTargetKind) _reloKind,
-                                                              cg()),
-                                   __FILE__, __LINE__, getNode());
+            if (symbolKind == TR::SymbolType::typeClass && cg()->comp()->getOption(TR_UseSymbolValidationManager))
```
mstoodle: why only typeClass? are you intentionally not handling typeMethod with the SymbolManager?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -883,15 +884,29 @@ TR::X86ImmInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
...
                // intentional fall-through
             case TR_ClassPointer:
-               cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
-                                                                           (uint8_t*)getNode(),
-                                                                           (TR_ExternalRelocationTargetKind) _reloKind,
-                                                                           cg()),
-                  __FILE__,
-                  __LINE__,
-                  getNode());
+               if (symbolKind == TR::SymbolType::typeClass && cg()->comp()->getOption(TR_UseSymbolValidationManager))
```
mstoodle: why only typeClass? are you intentionally not handling typeMethod with the SymbolManager?

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -2589,13 +2635,27 @@ TR::AMD64RegImm64Instruction::addMetaDataForCodeAddress(uint8_t *cursor)
...
             case TR_RamMethod:
+               symbolKind = TR::SymbolType::typeMethod;
                // intentional fall-through
             case TR_ClassPointer:
-               cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
-                                                                 (uint8_t*)getNode(),
-                                                                 (TR_ExternalRelocationTargetKind) _reloKind,
-                                                                 cg()),
-                                      __FILE__, __LINE__, getNode());
+               if (symbolKind == TR::SymbolType::typeClass && cg()->comp()->getOption(TR_UseSymbolValidationManager))
```
mstoodle: why only typeClass? are you intentionally not handling typeMethod with the SymbolManager?

compiler/optimizer/Inliner.cpp
```diff
@@ -5689,6 +5689,8 @@ TR_CallSite::addTarget(TR_Memory* mem, TR_InlinerBase *inliner, TR_VirtualGuardS
 
    addTarget(result);
 
+   _comp->validateImplementer(implementer);
```
mstoodle: seems strange to ignore the return value...

compiler/compile/OMRCompilation.hpp
```diff
@@ -353,6 +353,7 @@ class OMR_EXTENSIBLE Compilation
    void verifySymbolHasBeenValidated(void *symbol) { return; }
    void incrementHeuristicRegion() { return; }
    void decrementHeuristicRegion() { return; }
+   bool validateImplementer(TR_ResolvedMethod *implementer) { return true; }
```
mstoodle: Maybe comment that this facility isn't used directly in OMR, but can be implemented by consuming projects, along with a description of how the return value will be interpreted by OMR?

compiler/codegen/Relocation.cpp
```diff
@@ -423,6 +423,43 @@ const char *TR::ExternalRelocation::_externalRelocationTargetKindNames[TR_NumExt
...
+   "TR_ValidateInterfaceMethodFromCP (89)",
+   "TR_ValidateMethodFromClassAndSig (90)",
+   "TR_ValidateStackWalkerMaySkipFramesRecord (91)",
+   "TR_ValidateArrayClassFromJavaVM (92)",
+   "TR_ValidateClassInfoIsInitialized (93)",
+   "TR_ValidateMethodFromSingleImplementer (94)",
+   "TR_ValidateMethodFromSingleInterfaceImplementer (95)",
+   "TR_ValidateMethodFromSingleAbstractImplementer (96)",
+   "TR_ValidateImproperInterfaceMethodFromCP (97)",
+   "TR_SymbolFromManager (98)",
```
mstoodle: impressive addition to the list...

compiler/codegen/Relocation.cpp
```diff
@@ -423,6 +423,43 @@ const char *TR::ExternalRelocation::_externalRelocationTargetKindNames[TR_NumExt
    "TR_DebugCounter (59)",
    "TR_ClassUnloadAssumption (60)",
    "TR_J2IVirtualThunkPointer (61)",
+   "TR_InlinedAbstractMethodWithNopGuard (62)",
```
mstoodle: This one doesn't look like it should fall into the category of "Validation type"

compiler/codegen/Relocation.cpp
mstoodle: Is TR_SymbolFromManager actually a validation relocation type?

mstoodle: i confess i'm pretty worried that this is going to result in pretty extreme overkill in the common cases we're hopefully already getting right without this degree of complexity

compiler/x/amd64/codegen/OMRMemoryReference.cpp
dsouzai: Yes they would, I just didn't have the time to develop and test this on all platforms; for now this code only works on x86-64. The reason being that I only added a `symbolFromManager` relo type that is used to patch a pointer; however, this isn't sufficient on a platform like P that patches pointers differently depending on whether the relo record was `TR_ClassAddress` or `TR_ClassPointer`. In the next few months I have to make all of this work on all platforms.

compiler/x/codegen/X86BinaryEncoding.cpp
dsouzai: In the beginning I did try to have this work for even the other method relocation record types, but I ran into some issues that I never got around to dealing with (for reasons of time constraints); so I figured I'd leave those alone for now until the future. I can open some issues to track all of these TODOs.

compiler/x/codegen/X86BinaryEncoding.cpp
dsouzai: Same as above.

compiler/x/codegen/X86BinaryEncoding.cpp
dsouzai: Same as above.

compiler/x/codegen/X86BinaryEncoding.cpp
dsouzai: Same as above.

compiler/codegen/Relocation.cpp
dsouzai: > Is TR_SymbolFromManager actually a validation relocation type?

No it's a relocation record, it will patch a pointer and add class unload/redefinition assumptions.


>i confess i'm pretty worried that this is going to result in pretty extreme overkill in the common cases we're hopefully already getting right without this degree of complexity

I understand the concerns; when I first started this work (namely try to improve AOT performance), I wanted to not have to go down this road because of the obvious risks of adding this much code. However, it quickly became apparent to me that for most cases I couldn't, in any reasonable amount of time, convince myself that a particular code path was safe to enable under AOT. Additionally, as the code stands, it becomes very difficult to deal with any code path that works with methods or classes that have a cpIndex of -1. Overall, in order to make AOT better in the long run, I decided to go down this road to make life easier for myself and anyone else in the future dealing with AOT. That's why @jdmpapin and myself came up with this approach, which in essence is a symbol table used by linkers for relocations, to simply the compiler; our goal was to ensure that no one ever has to say "we didn't enable this code under AOT because it is too hard".

compiler/codegen/Relocation.cpp
dsouzai: You're right, it isn't; it's just at the bottom because moving it to where the other inlined methods are would involve shifting and renaming (the number portion) or all of the other relocation names, which in the interest of time I decided was not worth doing.

compiler/optimizer/Inliner.cpp
dsouzai: I know why; it's a bad reason but I remember why I ignored this. It's because the code locations that addTarget don't really have a fall back path; once it's gotten to calling `addTarget`, it just assumes adding to the list of targets should just work; getting the concept of "oh we might not want to do so for some reason" requires a bit of work; I'll open an issue to track this.

dsouzai: FWIW, to give OpenJ9 as an example, because we're validating the implementer which is a method symbol, there is no way for a validation to fail (a class symbol validation can fail if we can't remember the class). However, you're right in that in general, we shouldn't ignore this.

compiler/compile/VirtualGuard.cpp
```diff
@@ -196,6 +196,8 @@ TR_VirtualGuard::createVftGuard
    aconstNode->setInlinedSiteIndex(calleeIndex);
    aconstNode->setByteCodeIndex(0);
 
+   comp->verifySymbolHasBeenValidated(static_cast<void *>(thisClass));
+
```
vijaysun-omr: Is there something particular about VFT (class) guards that make them important to handle but not (say) method guards ?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -4870,6 +4870,9 @@ static void devirtualizeCall(OMR::ValuePropagation *vp, TR::Node *node)
       return;
       }
 
+   if (!vp->comp()->fe()->canDevirtualizeDispatch())
+      return;
+
```
vijaysun-omr: Is the situation you described (where a virtual call is not de-virtualized and yet has it's CP index changed to -1) one where we convert an interface call to a virtual call ? 

Regardless, is it avoiding precisely that case that cannot be handled by AOT or also avoiding cases when it could have devirtualized without causing any issues wrt AOT ? e.g. the simplest case when the method mentioned in the virtual call is changed to direct call of the same method, i.e. no new CP index is needed ?

compiler/runtime/Runtime.hpp
```diff
@@ -413,6 +413,7 @@ typedef struct TR_RelocationRecordInformation {
    uintptr_t data2;
    uintptr_t data3;
    uintptr_t data4;
+   uintptr_t data5;
```
vijaysun-omr: Could you please describe in brief why the new field is being added ? The names of these variables are not descriptive enough for me to assess what they are used for eventually. Maybe update the commit message too.

compiler/compile/VirtualGuard.cpp
dsouzai: This is some debug code that doesn't need to exist any more; I'll clean this stuff up.

compiler/optimizer/VPHandlers.cpp
dsouzai: >Is the situation you described (where a virtual call is not de-virtualized and yet has it's CP index changed to -1) one where we convert an interface call to a virtual call ?

I don't know about that. I would have to re-try running my tests with this code removed to see what the condition was that resulted in a virtual call where the J2I Thunk had a cpIndex of -1.

>Regardless, is this avoiding precisely that case that cannot be handled by AOT or also avoiding cases when it could have devirtualized without causing any issues wrt AOT ? e.g. the simplest case when the method mentioned in the virtual call is changed to direct call of the same method, i.e. no new CP index is needed ?

It is definitely avoiding the case that cannot be handled by AOT. However, I don't know if this now also avoids the cases when it could have been devirtualized; it is something that I moved on from due to time constraints and so I suppose there is opportunity to not be as conservative as I have been here. That said, this change is consistent with other parts of (at least OpenJ9) code that check do a front end query to see if devirtualization is possible.

I'll open an issue to revisit this.

dsouzai:
https://github.com/eclipse/openj9/issues/2921 contains issues to track future cleanup. Although this is from the OpenJ9 project, as and when the issues are tackled, OMR issues will be created.

compiler/optimizer/VPHandlers.cpp
jdmpapin: I suspect this is because the J2I virtual thunk relocations work based on a constant pool entry. To properly interact with the "symbol validation manager," they should instead take a method ID.

compiler/x/codegen/X86BinaryEncoding.cpp
mstoodle: Well, ok. It seems like a strange way to structure it (switch that separates out method from class handling, but all cases default fall-through to a class-specific case that then checks if it's typeClass before using the new or the old infra). Why not put the old infra into the lowest fall-through for the method cases (reordering if needed) and the same for the newer symbol manager handling at the bottom of all the class cases?

Oh, I see, because you want to be able to selectively not use the symbol manager at all.... Ok. Somewhat yuck, but I don't see an appreciably cleaner way and it should mostly resolve itself when you figure out the method pointer issues. Also feel mildly queasy that methods are flaky but classes aren't. Are you sure it wasn't just early implementation glitches that have since been resolved? I wouldn't have thought these two fundamental ones would behave significantly differently...

compiler/compile/OMRCompilation.hpp
mstoodle: ...`incrementHeuristicRegion()` should be called on entry (and `decrementHeuristicRegion()` on exit) to such a heuristic region. Actually, I think "enter" and "exit" are better verbs for this facility but that's something that can be cleaned up later.

compiler/compile/OMRCompilation.hpp
mstoodle: I can't help asking "a valid implementer of what?" If the question is really inlining specific, could the name instead be `validateTargetToBeInlined()` or something a little more meaningful?

compiler/optimizer/Inliner.cpp
mstoodle: Can you at least assert on true return value?

compiler/optimizer/VPConstraint.cpp
mstoodle: It does seem odd that this function can be called when there is no KNOT, unless there's an "undefined/unknown" index value. And presumably every caller is just about to look at whether the return value is NULL...

compiler/codegen/Relocation.cpp
dsouzai: Yeah I did, because there wasn't really much to be gained by logically separating the commits that effectively did the same thing, namely add new relo kinds.

dsouzai:
In the interest of time, I'm closing this PR since https://github.com/eclipse/omr/pull/2978 will contain all the relocatable compilation improvements.

PR #2978, opened 2018-09-18
Improvements for Relocatable Compilations

dsouzai:
This PR introduces changes in OMR required to enable optimizations that are currently implicitly disabled for relocatable compilations.

Issue tracking all the PRs: eclipse/openj9#2921


0xdaryl:
Thanks for the contribution.  This should be marked WIP until #2977 is merged.

compiler/compile/VirtualGuard.cpp
```diff
@@ -196,8 +196,6 @@ TR_VirtualGuard::createVftGuard
    aconstNode->setInlinedSiteIndex(calleeIndex);
    aconstNode->setByteCodeIndex(0);
 
-   comp->verifySymbolHasBeenValidated(static_cast<void *>(thisClass));
```
mstoodle: It would be better to squash this change with the one in #2977 that introduced the function in the first place...

compiler/x/codegen/DataSnippet.cpp
```diff
@@ -80,6 +79,16 @@ TR::X86DataSnippet::addMetaDataForCodeAddress(uint8_t *cursor)
...
+
+      TR_OpaqueClassBlock *clazz = getData<TR_OpaqueClassBlock *>();
+      if (clazz && cg()->comp()->compileRelocatableCode() && cg()->comp()->getOption(TR_UseSymbolValidationManager))
+         {
+         cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
+                                                               (uint8_t *)clazz,
+                                                               (uint8_t *)TR::SymbolType::typeClass,
+                                                               TR_SymbolFromManager,
+                                                               cg()),  __FILE__, __LINE__, getNode());
+         }
```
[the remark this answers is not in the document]
dsouzai: Yeah that's a good idea; although before this code, we never created a DataSnippet for a class pointer under AOT - this code is now active under AOT because checkcast inlining was enabled.

compiler/compile/VirtualGuard.cpp
dsouzai: Yeah you're right, I was trying to avoid merge conflicts :) . I'll try and clean this up.

compiler/x/codegen/X86BinaryEncoding.cpp
```diff
@@ -868,6 +868,7 @@ TR::X86ImmInstruction::addMetaDataForCodeAddress(uint8_t *cursor)
 
       if (getReloKind() != -1) // TODO: need to change Body info one to use this
          {
+         TR::SymbolType symbolKind = TR::SymbolType::typeClass;
```
mstoodle: again, this commit should have a stand-alone title and message.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1599,6 +1599,7 @@ TR::Node *constrainAload(OMR::ValuePropagation *vp, TR::Node *node)
                sym->isFinal()))
             haveClassLookaheadInfo = true;
 
+         bool allowForAOT = vp->comp()->getOption(TR_UseSymbolValidationManager);
```
mstoodle: Given only classes are actually covered properly by the symbol validation manager (methods still use the "old" approach)...have you convinced yourself that the enhanced knowledge about classes doesn't end up enabling an incorrect runtime state due to the methods derived from those classes? Or will that be protected because the queries on the resulting method won't let things get out of hand?

compiler/optimizer/VPConstraint.cpp
```diff
@@ -1215,8 +1215,9 @@ TR::VPClassType *TR::VPClassType::create(OMR::ValuePropagation *vp, const char *
    if (classObject)
       {
       bool isClassInitialized = false;
+      bool allowForAOT = vp->comp()->getOption(TR_UseSymbolValidationManager);
```
mstoodle: another commit that should have a more meaningful commit title and message

compiler/x/codegen/DataSnippet.cpp
```diff
@@ -54,9 +54,8 @@ TR::X86DataSnippet::addMetaDataForCodeAddress(uint8_t *cursor)
    if (_isClassAddress)
       {
       bool needRelocation = TR::Compiler->cls.classUnloadAssumptionNeedsRelocation(cg()->comp());
-      if (needRelocation)
+      if (needRelocation && !cg()->comp()->compileRelocatableCode())
```
mstoodle: another commit that should have a more meaningful title and message...These changes do not "Enable checkcast inlining under AOT"

mstoodle: `needRelocation` should really be renamed to `assumptionNeedsRelocation` ; I'll allow it in a subsequent PR.

compiler/optimizer/LocalOpts.cpp
```diff
@@ -8058,7 +8058,9 @@ TR_ColdBlockMarker::hasNotYetRun(TR::Node *node)
       TR::SymbolReference *symRef = node->getSymbolReference();
       bool isUnresolved;
 
-      if (comp()->compileRelocatableCode() && !comp()->getOption(TR_DisablePeekAOTResolutions))
+      if (comp()->compileRelocatableCode() &&
+          !comp()->getOption(TR_UseSymbolValidationManager) &&
```
mstoodle: Are you just trying to preserve the old behaviour here by overloading on `TR_UseSymbolValidationManager`? The downstream code doesn't seem dependent on the symbol manager to me?

compiler/compile/Method.cpp
```diff
@@ -385,6 +385,7 @@ bool         TR_ResolvedMethod::isPublic()                                 { not
 bool         TR_ResolvedMethod::isFinal()                                  { notImplemented("isFinal"); return false; }
 bool         TR_ResolvedMethod::isStrictFP()                               { notImplemented("isStrictFP"); return false; }
 bool         TR_ResolvedMethod::isInterpreted()                            { notImplemented("isInterpreted"); return false; }
+bool         TR_ResolvedMethod::isInterpretedForHeuristics()               { notImplemented("isInterpretedForHeuristics"); return false; }
```
mstoodle: ...makes me wonder if `ForHeuristics` (or its other state) is a concept worth promoting more broadly...

For example, the `isUnresolved()` change you had to make in another commit...

mstoodle:
As a general statement, it would be nicer to see these things propagate into the code base in smaller chunks and not all at once. I know Eclipse OpenJ9 is very keen to get these changes into their 0.11 release and these changes are needed to enable even bigger things in OpenJ9, but I suspect at least some of these commits could have been submitted earlier than the full batch was and could have been merged early (i.e. not all of these changes are connected to the new symbol validation manager).

Anyway, it would be nice to see these improvements to the code base flowing in via a more incremental and hopefully spread out process, is my hopeful request for future submissions.

compiler/optimizer/VPHandlers.cpp
dsouzai: From what I've seen, `TR_MethodPointer` is only generated when dealing with Profiled Pointers, so it _should_ be ok using the "old" approach due to the class chain + first class loaded from classloader validations. Additionally, `TR_RamMethod` seems to be used only (or more specifically, can *only* be used) for the method being compiled, so that too should be ok using the "old" approach. That said, I'm doing some runs now with these changes made to the appropriate places
```
-               if (symbolKind == TR::SymbolType::typeClass && cg()->comp()->getOption(TR_UseSymbolValidationManager))
+               if (cg()->comp()->getOption(TR_UseSymbolValidationManager))
```
since I can't think of any good reason why the "new" approach shouldn't working, Towards the end of this work, I did find a bug in this code that I fixed, so perhaps I won't see that crash now.

compiler/optimizer/LocalOpts.cpp
dsouzai: > Are you just trying to preserve the old behaviour here by overloading on TR_UseSymbolValidationManager?

Yeah that's exactly it. I wanted there to be a way for the old code to run exactly as it did without all the new changes.

dsouzai: Although... the `isInterpretedForHeuristics` change also affects the "old" behaviour... Though, it only affects it from a heuristic point of view and, if anything, results in better decisions being made, so I suppose (like https://github.com/eclipse/omr/pull/2978/commits/1e645c3d91ccae4927e6c7f1de50163d07145847 is just a generally good thing to have)

mstoodle:
genie-omr build all

mstoodle:
I have opened #3029 to start documenting work that should be happen going forward to tease OpenJ9 and OMR further apart in their mutual dependency in specifically the relocation infrastructure. With that in place, I think we can merge this one now.

PR #2980, opened 2018-09-18
Improve maxMinHelper in z/codegen 

dhong44:
This pull request improves the `maxMinHelper` evaluator
in Z codegen in 3 ways.
- Add guards for z10  in the evaluator and lower and disable max/min optimisations on unsupported platforms
- Rewrites the evaluator for lmin and lmax when using register pairs
- Add Tril tests to capture corner cases for this evaluator

### DIFF
--- a/fvtest/compilertriltest/MaxMinTest.cpp
+++ b/fvtest/compilertriltest/MaxMinTest.cpp
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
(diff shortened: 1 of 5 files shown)


dhong44:
@fjeremic, @dchopra001

fvtest/compilertriltest/MaxMinTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    ASSERT_NOTNULL(trees);
+
+    Tril::DefaultCompiler compiler{trees};
+
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+    auto entry_point = compiler.getEntryPoint<int32_t (*)(void)>();
+    volatile auto exp = param.oracle(param.lhs, param.rhs);
+    volatile auto act = entry_point();
```
Leonardo2718: Do these really need to be volatile? I think just having `ASSERT_EQ(eparam.oracle(param.lhs, param.rhs), entry_point())` should work.

fvtest/compilertriltest/MaxMinTest.cpp
```diff
@@ -85,7 +120,7 @@ TEST_P(Int32MaxMin, UsingLoadParam) {
 
 
 INSTANTIATE_TEST_CASE_P(MaxMin, Int32MaxMin, ::testing::Combine(
-    ::testing::ValuesIn(TRTest::combine(TRTest::const_values<int32_t>(), TRTest::const_values<int32_t>())),
+    ::testing::ValuesIn(TRTest::combine(intMaxMinTestValues(), intMaxMinTestValues())),
```
fjeremic: Why not combine both of these? Or better yet why don't we scrap the two custom functions you have here (`intMaxMinTestValues` and `longMaxMinTestValues`) and put the constants directly in `TRTest::const_values` so all other tests get the benefit of testing the additional constants?

dhong44:
These Trill tests will not work on z10 and lower since conditional loads are not supported. Currently Tril does not support platform specific testing, so we cannot guard against this. https://github.com/eclipse/omr/issues/1843 Is open to track this.

fjeremic:
I'll preemptively launch the builds. However given the Tril changes I'll invite @Leonardo2718 to review again. Thanks for addressing all the concerns @dhong44!

@genie-omr build all

fjeremic:
Travis AArch build failed due to infrastructure. Given this is a Z change I'm ok to merge.

PR #2981, opened 2018-09-18
Consolidate register iterators

samolisov:
Each code generator implements a RegisterIterator class which appears to be implemented solely to assist register assignment tracing dumps in ras/Debug.cpp. The DNA of this class across all the code generators must be at least 97% identical, and I have no doubt that with a bit of thought a single, common implementation shared by all code generators could be developed.

A generic implementation of the 'TR::RegisterIterator' is implemented
and all platform-specific 'OMR::ARCH::RegisterIterator' classes are
removed from the codebase. The signature of the class constructor is
changed a bit: the constructor takes the machine as well as the first
and last assignable registers, no TR_RegisterKinds is used:

```cpp
self()->setGPRegisterIterator(new (self()->trHeapMemory()) TR::RegisterIterator(self()->machine(),
    TR::RealRegister::FirstGPR, TR::RealRegister::LastAssignableGPR));
self()->setFPRegisterIterator(new (self()->trHeapMemory()) TR::RegisterIterator(self()->machine(),
    TR::RealRegister::FirstXMMR, TR::RealRegister::LastXMMR));
```

**Warning**:  such methods as `getARCHITECTURERealRegister()` were renamed to `getRealRegister` and this affects the downstream projects (OpenJ9 at least). PR eclipse/openj9#2936 in Eclipse/openJ9 is opened to reflect the changes.

Building on all supported platforms is needed.


samolisov:
Eventualy, the PR is successful built on Travis and AppVeyor. @0xdaryl could you review?

charliegracie:
@mstoodle can you or another compiler committer review or suggest someone else to review?

samolisov:
@0xdaryl I'm sorry for the late response. Firstly, thank you very much for your comments.

I have a couple of questions:

About the **getRealRegister** method, the code generator for z Systems has two versions of the method (and all other ones have only the first one):

```cpp
TR::RealRegister *getRealRegister(TR::RealRegister::RegNum regNum)
      {
      return _registerFile[regNum];
      }

   TR::RealRegister *getRealRegister(int32_t regNum)
      {
      return _registerFile[regNum];
      }
```

I can move both to `compiler/codegen/OMRMachine.hpp/cpp` but because the second is used only within the z codegen I think it should still be left in the `compiler/z/codegen/OMRMachine.hpp`. Do you agree?

---

About a platform-specific wrappers for the `getRealRegister` method. I've opened a PR for OpenJ9 ( eclipse/openj9#2936) to reflect these changes. Please let me know if you think this is not enough and I'll add an additional commit where introduce the platform-specific wrappers for the method. Then, after OpenJ9 is changed, we will be able just to revert this commit and remove the methods.

About testing, I enabled tracing for register allocation and see the log (during an execution of the compilertest.exe test): all X86 (I tested only on X86) general purpose registers were there.

samolisov:
I have an idea to split this PR on a series and introduce smaller changes to the codebase.

PR #2983, opened 2018-09-18
Avoid uses of memset() with a constant zero length

keithc-ca:
This fixes the failure seen in https://ci.adoptopenjdk.net/job/openjdk11_openj9_build_ppc64le_linux/11/console
```
/usr/include/bits/string3.h:81:7: error: call to ‘__warn_memset_zero_len’ declared with attribute warning: memset used with constant zero length parameter; this could be due to transposed parameters [-Werror]
       __warn_memset_zero_len ();
       ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/port/common/omrstr.c
+++ b/port/common/omrstr.c
@@ -943,16 +943,18 @@ writeStringToBuffer(char *buf, uintptr_t bufLen, uint64_t width, uint64_t precis
 	if (leftPadding > bufLen) {
 		leftPadding = bufLen;
 	}
-	if (buf && (0 != leftPadding)) {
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

charliegracie:
Is this a new warning in a newer version of GCC?

keithc-ca:
The build I referenced says gcc reports 7.3.0 (which works for me on Ubuntu 16.04), but some header files are linked to the O/S version (CentOS in this case), so it may be the combination of gcc7 and CentOS that leads to the warning.

Issue #2922, opened 2018-09-04
Add Tril tests for aiadd / aladd evaluators


fjeremic:
As part of #2920 we let a bug escape because we had no tests for the `aiadd` and `aladd` evaluators. Even a simple test case would have caught the issue in the aforementioned PR. The task here is to implement some Tril tests for these evaluators. 

One can test the changes against backing out the change in #2920 and validating that it does indeed fail.

PR #2964, opened 2018-09-14
AArch64: Add utility functions for generating MOV, MUL, etc.

knn-k:
This commit implements some more utility functions in
GenerateInstructions.* for aarch64 for generating instructions
that take xzr as one of target/source registers, in addition to the
functions added in #2888.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #2963, opened 2018-09-14
AArch64: Implement loadConstant32() and loadConstant64()

knn-k:
This commit implements loadConstant32() and loadConstant64()
utility functions for loading a integer value to a register for
aarch64.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #2985, opened 2018-09-19
WIP: AArch64: Implement iaddEvaluator() and some others

knn-k:
This commit implements some evaluators in BinaryEvaluator.cpp,
such as iaddEvaluator() and imulEvaluator(), for aarch64.
This depends on #2963, and #2964.


knn-k:
#2964 is now merged, and this one is ready for review.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -38,10 +76,27 @@ OMR::ARM64::TreeEvaluator::laddEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
-	}
+   {
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *src1Reg = cg->evaluate(firstChild);
+   TR::Node *secondChild = node->getSecondChild();
+   TR::Register *trgReg = cg->allocateRegister();
+
+   if (secondChild->getOpCode().isLoadConst() && secondChild->getRegister() == NULL)
+      {
+      addConstant32ToInteger(node, trgReg, src1Reg, -secondChild->getInt(), cg);
```
0xdaryl: AArch64 has an immediate form subtract instruction that does this kind of operation internally.  What is the advantage of manually converting the instruction and constant operand into an addition?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -50,12 +105,63 @@ OMR::ARM64::TreeEvaluator::lsubEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+      }
+   else if (value == -1)
+      {
+      generateNegInstruction(cg, node, treg, sreg);
+      }
+   else
+      {
+      TR::Register *tempReg = cg->allocateRegister();
+      loadConstant32(cg, node, value, tempReg);
+      generateMulInstruction(cg, node, treg, sreg, tempReg);
```
0xdaryl: If `tempReg` is a temporary register, you'll have to manually kill its live range here:
`cg->stopUsingRegister(tempReg)`.

compiler/aarch64/codegen/BinaryEvaluator.cpp
knn-k: I thought it would save the code size of the JIT compiler itself.
I changed the function `addConstant32ToInteger()`  to `addOrSubConstant32ToInteger()` that takes an extra parameter.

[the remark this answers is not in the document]
knn-k: With the change from `addConstant32ToInteger()` to `addOrSubConstant32ToInteger()`, there is no need to worry about `-MIN_INT`.

knn-k:
Updated files reflecting review comments.

compiler/aarch64/codegen/ARM64Instruction.hpp
```diff
@@ -46,6 +46,16 @@ inline bool constantIsImm9(int32_t intValue)
    return (-256 <= intValue && intValue < 256);
    }
 
+/*
+ * @brief Answers if the signed integer value can be placed in 12-bit field
+ * @param[in] intValue : signed integer value
+ * @return true if the value can be placed in 12-bit field, false otherwise
+ */
```
0xdaryl: My understanding of the ADDImm and SUBImm instructions is that the 12-bit immediate field is zero-extended to 32/64 bits (regardless of whether it is shifted or not).  This means that only unsigned (positive) values 0<=x<=4095 are suitable.  The code here allows for negative values which is not correct.

fjeremic:
I hate to be "that guy" but this seems like the perfect opportunity to fix https://github.com/eclipse/omr/issues/2922. @knn-k how do you feel about writing some tests and helping the entire community as a whole?

0xdaryl:
@fjeremic : Unit testing in general for ARM and AArch64 is weak (non-existent?), and we do plan on writing Tril-based unit tests for the opcodes that would be shared across all platforms (and thereby retiring the platform-specific unit tests).

This particular PR deals will iadd/isub only, so aiadd/aladd will have to wait.

Having said that, it is timely to look at implementing iadd/isub Tril tests for ARM/AArch64 (and therefore all other platforms).  I'll speak to @knn-k @aarongraham9 @jeanlego @marufunb about moving that work up.

fjeremic:
> @fjeremic : Unit testing in general for ARM and AArch64 is weak (non-existent?), and we do plan on writing Tril-based unit tests for the opcodes that would be shared across all platforms (and thereby retiring the platform-specific unit tests).

Uuuups, right! We're talking about `iadd` not `aiadd` here. I must have "borrowed" one of the a's from AArch when reading the title. Carry on...

compiler/aarch64/codegen/ARM64Instruction.hpp
knn-k: You are correct.  The immediate field of these instructions are zero-extended.

knn-k:

0xdaryl:
@genie-omr build aarch64

PR #2990, opened 2018-09-19
Fix address type alignment in OMRIlTypes

0xdaryl:
Use correct alignment for 32-bit and 64-bit targets.


charliegracie:
@genie-omr build all

PR #2995, opened 2018-09-20
Fix typos in documentation of return values

keithc-ca:
The return values are `EsIsDir` and `EsIsFile` (with an uppercase 'I' instead of a lowercase 'l').

[ci skip] comment change only


charliegracie:
Merging without AppVeyor as this is a documentation only change.

FYI [skip ci] / [ci skip] in the first line of the commit message will work for both AppVeyor and Travis... Sadly AppVeyor does require it to be in the first line.

PR #2996, opened 2018-09-20
Make OMR an installable library

youngar:
This is the start of a series of changes intended to make OMR more
easily consumable.  This change is prototyped using the standalone AVL
library, although it should be trivial to add support for other
libraries. Some of the ways we hope OMR can be consumed are as follows:

 1. Installable OMR.  By creating proper install targets for OMR using
    standard paths, it should be trivial to package OMR for any linux
    distribution.  Users will also be able to install OMR into their
    home directories, using a properly configured PREFIX.
 2. CMake User Registry.  With a little bit of care, we can support the
    CMake User Registry.  This means that a user can build OMR anywhere
    on their system, and the build directory will be registered in
    `~/.cmake/packages/Omr`.  This will allow users to consume OMR using
    `find_package(Omr)`, without having to install OMR.
 3. As a subproject.  Having OMR as a subdirectory is the most common
    way to consume OMR.  We intend to continue to support and enhance
    this method of consumption.

We also hope to add versioning support to OMR in the future.


youngar:
This work still needs to be tested with OpenJ9.  I'm going to run CI to ensure I'm not using unsupported CMake features.
@genie-omr build all

ymanton:
Providing binary libraries seems like something that warrants a lot of discussion and a careful weighing of the pros and cons. ABI backward compatibility has a lot of baggage.

ymanton:
We don't have to worry about ABI compatibility at the moment (except maybe JitBuilder), we only distribute source code so only have to worry about source-level backward compatibility. ABI compatibility would come into play if we defined any sort of consumable binary object that was independent of the consumers.

I guess I don't quite understand what you mean by "Installable OMR" and "tested with OpenJ9". If we provide the ability to build an installable OMR library (static or shared), we'll still be on the hook for ABI backward compatibility, even if we don't provide the actual library binaries. Perhaps I'm misunderstanding what exactly is intended to be built here and how it is supposed to be consumed.

youngar:
Hey Younes, I was a careless and didn't distinguishing between ABI and source compatibility, sorry about that.

By "test with OpenJ9", I was just making sure that none of my changes broke OpenJ9 cmake builds (and they actually did).

I don't think that providing a proper install target puts us on the hook for backwards abi compatibility.  I think that if someone wants to install a development version of OMR from git on their system, they should be able to.  I don't see the problem with that.

youngar:
By the way, one of the main benefits of properly supporting installing, is that we gain the ability to export our project.  This is how item 2 is supported, and the most common way for cmake libraries to be consumed. You can read more about installing and exporting here if you want to know more: 
https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html
https://cliutils.gitlab.io/modern-cmake/chapters/install/exporting.html

ymanton:
> I don't think that providing a proper install target puts us on the hook for backwards abi compatibility. I think that if someone wants to install a development version of OMR from git on their system, they should be able to. I don't see the problem with that.

What happens when you update that installed version a week later? Will all software that was previously built against it have to be rebuilt? If so it's an unstable library and it doesn't make much sense to install it as a separate, independent binary. If nothing has to be rebuilt then it has a stable ABI.

ymanton:
I guess the answer to that depends on whether we're talking about static or dynamic libraries and in this case you have a static library, so that changes the question of ABI compatibility to "if you update that installed version a week later will all software that was previously able to link against it still link against it if rebuilt?"

ymanton:
Sorry, but isn't that what is implied by (1) in the PR? If I install it into `/home/ymanton/lib/` all of the same problems exist, just not globally.

My apologies if I'm totally misunderstanding your intentions, but a clear description or example of what is produced and installed and how it's supposed to be consumed by downstreams would be helpful.

youngar:
@rwy0717 @dnakamura Can you guys review this?

youngar:
In the commit message, I was listing things that this PR could possibly enable
down the road.  I agree that installing OMR libraries into your home directory
can be problematic.  For me personally, I would be comfortable with the risks,
and I believe that it would helpful to me for local development. If it makes
you more comfortable, I can remove (1) from the list of possible uses. This PR
is still useful, even without considering (1).

DanHeidinga:
As a(n OpenJ9) consumer of OMR, I'm still struggling to understand the use case here.  If most uses of OMR are like ours, don't they depend on the `glue` code?  What would an installable library be without the glue implementation?

Put another way, what does it mean for a source-only project to be an installable library?

rwy0717:
You could "install" OMR into a JVM tree, for example. We're already doing this to put omrsig into the right place.

youngar:
As well, If another consumer wanted to create a package for their runtime, the install targets in OMR would enable this.

rwy0717:
Yeah for sure. I could use the OMR install targets to build a Base9 package too. Without this, every consumer of OMR is stuck writing nearly identical installation targets.

As well, we've already mentioned that some OMR "components" really act more like consumers, like jitbuilder, om, or the example. Why wouldn't you want to install these targets?

rwy0717:
Installing OMR's build tools could be useful as well. That would be an easy way to support cross compilation. Currently, we have a hacky mechanism for importing build tools from another local build, but being able to install buildtools (possibly to /tmp or wherever) seems more flexible and less problematic.

ymanton:
> You could "install" OMR into a JVM tree, for example. We're already doing this to put omrsig into the right place.

It's deeply confusing to call this an installation; this is creating a build artifact and putting it somewhere as part of a larger build. This is distinctly different than building a standalone library, intended to be installed in a shared location and updated independently of downstream consumers.

Build artifacts don't have ABI issues so forget about everything I said about ABIs. Build artifacts however really have to be driven by their consumer, the consumer has to provide the compiler, linker, and options that the library is to be built with -- you're building it on their behalf and it has to be ABI-compatible with whatever they're linking it with in addition to inheriting things like debug flags. Does using CMake functionality that is intended for installable, independent libraries for this make sense? I don't know CMake as well as I'd like, but I would think you'd want a CMake module or something that downstream can *include* in their CMake build system so that downstream's compiler, linker, etc settings get applied. Essentially the things provided by OMR would then be "definitions" of libraries and source files that make up those libraries, the location of corresponding headers, and other such things.

charliegracie:
@youngar I think the terminology may be what is confusing people but I believe that is the terminology used by CMake.  I personally see this a big step forward that will help downstream projects consume OMR.

Also technologies like PortLibrary, ThreadLibrary, Util, etc that have stable APIs could easily be "installed" on a system and used directly.  As JitBuilder and GCBuilder become stable it may also make sense for them to become installable libraries as well.  I know that is outside of the scope of this PR and hopefully it does not add to the confusion.

charliegracie:
@genie-omr build all

DanHeidinga:
> I think the terminology may be what is confusing people 

This is definitely the case.  Can someone describe the usecases more clearly then?

I may be missing the big picture but from my perspective, this looks like a solution in search of a problem.  

Questions like the ones below were brought up in person but I'm not sure I've followed the answer:
* Does this apply to OMR consumers with glue code?  
* While JITBuilder was mentioned as possible user of this, doesn't it use custom makefiles already?  Is the intention to introduce this and migrate JITBuilder to it?
* Some of the example uses - portlib, threadlib, util - are libraries that are typically statically linked.  Does this apply to static link libs?

I guess I'm trying to understand the big picture and the benefits this brings to the project.

Leonardo2718:
@DanHeidinga 

> * While JITBuilder was mentioned as possible user of this, doesn't it use custom makefiles already? Is the intention to introduce this and migrate JITBuilder to it?

The JitBuilder makefiles follow the process required to build the OMR compiler into any downstream project. They also "install" the JitBuilder library into the `jitbuilder/release` directory. By "install" I mean that the compiled library and headers are copied into the directory. The CMake files for JitBuilder do this as well. IMHO, using CMake install targets is a much more elegant and flexible solution. Also, being able to use `find_package(Omr)` could make consuming JitBuilder much easier (I'm thinking of [wasmjit-omr](https://github.com/wasmjit-omr/wasmjit-omr) as good example).

DanHeidinga:
@Leonardo2718 I'm glad to hear other consumers of OMR see value in this PR for their runtimes.  I think I the understand the use case where this is a purely local install into an existing runtime build though I'm still unclear on how it works with any runtime requiring glue code.

Leonardo2718:
@DanHeidinga 

If it helps, I will note that CMake's `find_package()` can be used for more than just including and linking a library. It can also be used to make helper utilities available to a project. For example, `find_package(BISON)` and `find_package(FLEX)` are used to "import" the `BISON_TARGET()` and `FLEX_TARGET()` functions. However, flex and bison are not actually linked or even shipped as part of the project (for obvious reasons). The Tril [CMakeLists.txt file](https://github.com/eclipse/omr/blob/master/fvtest/tril/tril/CMakeLists.txt) shows how these are used in practice.

For projects that require glue code, I think of the install target as making the OMR code and CMake utilities "visible" to a consuming project. For example, in the context of the OMR compiler, the install target might just copy the contents of `omr/compiler` into the install directory. `find_package(Omr)` would then make the compiler source files visible to the consumer and give them access to the `make_compiler_target()` function. I'm imagining a similar scenario would work for the other OMR components.

charliegracie:
@youngar do you think this change is ready?  If so can you remove the WIP, please?

youngar:
If no one has any complaints, I'll remove the WIP.  We need to get the OpenJ9 side merged first https://github.com/eclipse/openj9/pull/2998.

ymanton:
I believe both I and @DanHeidinga asked for some sort of example scenario or use case. I'd like to know what this produces, how/where it's supposed to be installed, and how downstream(s) are supposed to consume it.

youngar:
@ymanton I think @Leonardo2718 covered jitbuilder as a use case very well in https://github.com/eclipse/omr/pull/2996#issuecomment-425518185.

ymanton:
That's not exactly what I was asking for, I was asking for a high level description of what *this* patch does (what it produces, what it installs, etc), what downstream is supposed to do to consume it, and so on.

However I can piece some of it together from the accompanying OpenJ9 PR. From what I've gathered, downstream defines the install locations for binaries/libs/headers/etc and that's it, nothing else needed thus far. Is that correct?

OpenJ9 doesn't define the header install path, so does that mean the headers are consumed from the OMR source tree or are they installed and consumed from wherever `${CMAKE_INSTALL_INCLUDEDIR}` points to? If things are installed wherever downstream points them to how do things get rebuilt when one edits the code in OMR? Does downstream effectively have to do a `make install` on every build to get things to update incrementally? If downstream consumes the "installed" headers CMake would have to only "reinstall" updated headers when you do an incremental build; if it simply installs all headers, even ones that didn't change, all downstream files that have a dependency on all OMR headers will get rebuilt. Also, how is downstream supposed to choose which binaries and libraries get built and which don't, and more importantly, which toolchain and flags to use?

DanHeidinga:
> If a downstream project uses find_package(Omr) to use OMR, they won't have any access to the install targets of OMR anyways.

It's not unusual when working with runtimes like OpenJ9 to have multiple versions checked out at once.  Will find_package(Omr) find the package in the local source tree or some other package (maybe the first built one)?

> If a project is using find_package(Omr), then OMR will have already been fully configured and built. There is no way to change what is built and how without rebuilding OMR. It is up to CMake to make sure that the OMR build flags are passed on to downstream.

Is this usable by projects that have glue code?  If OMR is fully configured and built, then how does the consuming project adapt the OMR library for it's needs?

youngar:
>Will find_package(Omr) find the package in the local source tree or some other package (maybe the first built one)?

`find_package(Omr)` will not (necessarily) find the OMR in the local source tree.  Since OpenJ9 uses OMR as a subdirectory, it does not make sense for it to use `find_package`.

> Is this usable by projects that have glue code? If OMR is fully configured and built, then how does the consuming project adapt the OMR library for it's needs?

Right now, this is not usable by projects that have GC glue code.  The compiler component has a CMake macro `make_compiler_target()`, which allows for multiple compilers to be created.  We could copy this model for the GC.  This would allow OMR to compile the GC against multiple glues in one build, and allow any project to create a GC library compiled against their glue code.

rwy0717:
There’s lots of interesting talk about how or what kind of functionality could be provided by an OMR package. As long as we all agree that we want to package _something_, this PR seems like a reasonable first step.

People are suggesting we could take the OMR GC or compiler and produce some kind of installable “OMR development kit”. This is a really cool idea, and I can see how it might work. Still, I have my reservations. This would take a lot of work and experimentation to develop, and I don’t believe it should ever replace our original “OMR as a subdirectory” consumption story. It’s also pretty different from the more basic library packaging I’m interested in implementing today.

I would like us to build off this PR by adding install commands for glueless elements like jitbuilder and it’s dependencies. We could let users install the build tools, which would be helpful for cross-compile scenarios. I even think that install commands could be used to help consumers like J9 produce their own binary packages.

I don’t see why we can’t support all these things, eventually. If we’re smart with module and component decomposition, it should be easy to let users pick and choose what gets installed.

charliegracie:
So I feel like there is now a consensus that this is the right direction.. I will merge this PR later today if no one has any objections.

PR #2998, opened 2018-09-26
Fix RPATH and dylib install_name on OSX

babsingh:
On OSX, -rpath,\\$$ORIGIN doesn't work. -rpath needs to correspond to
either @executable_path, @loader_path or some other path. More
information can be seen in the man pages of dyld. I have chosen to set
rpath as "-rpath,@loader_path/.". @loader_path points to the directory
where dylib being loaded exists.

When naming dylibs, @rpath/ should be appended to dylib names. Example:
@rpath/libomrsig.dylib. Appending @rpath to the dylib name instructs the
dynamic linker to search the list of run paths for the dylib.

Reference:
https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath

### DIFF
--- a/omrmakefiles/rules.osx.mk
+++ b/omrmakefiles/rules.osx.mk
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2016, 2016 IBM Corp. and others
+# Copyright (c) 2016, 2018 IBM Corp. and others
 # 
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


babsingh:
fyi - @charliegracie, @DanHeidinga

babsingh:
Related to https://github.com/eclipse/openj9/issues/36.

babsingh:
@youngar can you help merge this PR?

youngar:
Our CI only tests using the CMake based build system, so there is no point in waiting for these jobs to finish.

charliegracie:
What about CMake? Very soon the autotools config is going to be removed... @babsingh, @youngar and @dnakamura what should we be doing for CMake so this is not an issue a few weeks from now?

youngar:
Hey, I was talking about this with @babsingh, the answer for cmake would be to set these variables:

```
set(CMAKE_INSTALL_RPATH "@loader_path")
set(MACOSX_RPATH TRUE)
# possible this:
SET(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
```

It is my opinion that setting these variables belongs in the OpenJ9 build system, not OMR.

Issue #2958, opened 2018-09-12
Codegen Support for atomic method symbols


liqunl:
x86 now supports atomic method symbols whose semantics are equivalent to common atomic operations such as fetch-and-add. 

Currently we have the following method symbols
```
atomicAdd32BitSymbol
atomicAdd64BitSymbol
atomicFetchAndAdd32BitSymbol
atomicFetchAndAdd64BitSymbol
atomicSwap32BitSymbol
atomicSwap64BitSymbol
atomicCompareAndSwap32BitSymbol
atomicCompareAndSwap64BitSymbol
```

Two more symbols @0dvictor is trying to add on x86 are (see https://github.com/eclipse/omr/pull/2174) 
```
atomicCompareAndSwapReturnValue32BitSymbol
atomicCompareAndSwapReturnValue64BitSymbol
```

They should be supported on all platforms. FYI @fjeremic @gita-omr @JamesKingdon

fjeremic:
Do we have tests added which exercise these symbols? How exactly do we know we've implemented them correctly?

0dvictor:
> Do we have tests added which exercise these symbols? How exactly do we know we've implemented them correctly?

Unfortunately we don't have tests now. They are only being tested in OpenJ9 through OpenJ9's tests.

fjeremic:
> Unfortunately we don't have tests now. They are only being tested in OpenJ9 through OpenJ9's tests.

I think we need to push the tests down into OMR then. If we wish to support an OMR API we should also have tests written against the respective API, otherwise the semantics of the APIs will diverge between codegens and non-OpenJ9 users of the APIs could potentially introduce changes which will break OpenJ9.

liqunl:
Not related to the discussion of adding test, but to show an idea of how large the item will be. Calls with the atomic method symbols share the same evaluator helper on X, which is https://github.com/eclipse/omr/blob/master/compiler/x/codegen/OMRTreeEvaluator.cpp#L2981

zl-wang:
Although the existing evaluators for Unsafe.CAS or JUC's Atomic classes methods can be largely reused,
the question, as I can see it, needs to be clarified whether these method symbols carry Unsafe semantic implication of memory ordering (such that we put in expensive memory barriers, treating the underlying accessed memory as java volatiles). I hope they don't.

liqunl:
>Although the existing evaluators for Unsafe.CAS or JUC's Atomic classes methods can be largely reused,
[...]

PR #2999, opened 2018-09-26
Implement atomic add and swap codegen support for Z

fjeremic:
In support of #2958 we implement the existing intrinsics for Z. Another follow-up PR will address the new intrinsics which are proposed in #2174.

- Implement Z codegen support for several atomic symbols
- Document atomic symbol semantics
- Document supportsAtomicAdd and introduce supportsAtomicSwap 

Issue: #2958

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -389,10 +389,8 @@ class OMR_EXTENSIBLE CodeGenerator
    /** \brief
     *     Determines whether intrinsics on the following atomic symbols is supported:
     *
-    *     - atomicAdd32BitSymbol
-    *     - atomicAdd64BitSymbol
-    *     - atomicFetchAndAdd32BitSymbol
-    *     - atomicFetchAndAdd64BitSymbol
+    *     - atomicAddSymbol
+    *     - atomicFetchAndAddSymbol
     *
     *  \return
     *     \c true if intrinsics are supported; \c false otherwise.
@@ -405,12 +403,9 @@ class OMR_EXTENSIBLE CodeGenerator
    /** \brief
     *     Determines whether intrinsics on the following atomic symbols is supported:
     *
-    *     - atomicSwap32BitSymbol
-    *     - atomicSwap64BitSymbol
-    *     - atomicCompareAndSwap32BitSymbol
-    *     - atomicCompareAndSwap64BitSymbol
-    *     - atomicCompareAndSwapReturnValue32BitSymbol
-    *     - atomicCompareAndSwapReturnValue64BitSymbol
+    *     - atomicSwapSymbol
+    *     - atomicCompareAndSwapSymbol
+    *     - atomicCompareAndSwapReturnValueSymbol
     *
     *  \return
     *     \c true if intrinsics are supported; \c false otherwise.
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -43,6 +43,7 @@ namespace OMR { typedef OMR::CodeGenerator CodeGeneratorConnector; }
 #include "codegen/StorageInfo.hpp"
 #include "codegen/TreeEvaluator.hpp"
 #include "compile/Compilation.hpp"              // for Compilation
+#include "compile/SymbolReferenceTable.hpp"
 #include "control/Options.hpp"
 #include "control/Options_inlines.hpp"
 #include "cs2/hashtab.h"                        // for HashTable, etc
@@ -387,30 +388,15 @@ class OMR_EXTENSIBLE CodeGenerator
    TR_HasRandomGenerator randomizer;
 
    /** \brief
-    *     Determines whether intrinsics on the following atomic symbols is supported:
+    *     Determines whether the code generator supports inlining intrinsics for \p symbol.
     *
-    *     - atomicAddSymbol
-    *     - atomicFetchAndAddSymbol
-    *
-    *  \return
-    *     \c true if intrinsics are supported; \c false otherwise.
-    */
-   bool supportsAtomicAdd()
-      {
-      return false;
-      }
-
-   /** \brief
-    *     Determines whether intrinsics on the following atomic symbols is supported:
-    *
-    *     - atomicSwapSymbol
-    *     - atomicCompareAndSwapSymbol
-    *     - atomicCompareAndSwapReturnValueSymbol
+    *  \param symbol
+    *     The symbol which to check.
     *
     *  \return
-    *     \c true if intrinsics are supported; \c false otherwise.
+    *     \c true if intrinsics on \p symbol are supported; \c false otherwise.
     */
-   bool supportsAtomicSwap()
+   bool supportsNonHelper(TR::SymbolReferenceTable::CommonNonhelperSymbol symbol)
       {
       return false;
       }
--- a/compiler/codegen/OMRCodeGenerator.hpp
+++ b/compiler/codegen/OMRCodeGenerator.hpp
@@ -386,7 +386,40 @@ class OMR_EXTENSIBLE CodeGenerator
 
    TR_HasRandomGenerator randomizer;
 
-   bool supportsAtomicAdd() {return false;}
+   /** \brief
+    *     Determines whether intrinsics on the following atomic symbols is supported:
+    *
+    *     - atomicAdd32BitSymbol
+    *     - atomicAdd64BitSymbol
+    *     - atomicFetchAndAdd32BitSymbol
+    *     - atomicFetchAndAdd64BitSymbol
+    *
+    *  \return
+    *     \c true if intrinsics are supported; \c false otherwise.
+    */
+   bool supportsAtomicAdd()
+      {
+      return false;
+      }
+
+   /** \brief
+    *     Determines whether intrinsics on the following atomic symbols is supported:
+    *
+    *     - atomicSwap32BitSymbol
+    *     - atomicSwap64BitSymbol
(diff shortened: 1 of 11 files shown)


fjeremic:
Requesting review from @liqunl @0dvictor for x86 code changes and verification of documentation, and @r30shah for Z codegen piece.

compiler/compile/OMRSymbolReferenceTable.hpp
```diff
@@ -186,11 +186,65 @@ class SymbolReferenceTable
...
+       *
+       *  Which performs the following operation atomically:
+       *
+       *  \code
+       *    [address] = [address] + <value>
+       *    return <value>
+       *  \endcode
+       */
       atomicAdd32BitSymbol,
       atomicAdd64BitSymbol,
```
fjeremic: @0xdaryl @0dvictor personally I don't think we should distinguish between 32-bit and 64-bit since this can be implied by the data type of `<value>`. I'm in favor of folding these into one `atomicAddSymbol` and doing a type check on the second child at codegen time to determine whether the addition is 32-bit or 64-bit. Is there any downside to this? I don't want to bloat these symbol tables with unnecessary enums.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -386,7 +386,40 @@ class OMR_EXTENSIBLE CodeGenerator
...
+    *     - atomicSwap64BitSymbol
+    *     - atomicCompareAndSwap32BitSymbol
+    *     - atomicCompareAndSwap64BitSymbol
+    *     - atomicCompareAndSwapReturnValue32BitSymbol
+    *     - atomicCompareAndSwapReturnValue64BitSymbol
+    *
+    *  \return
+    *     \c true if intrinsics are supported; \c false otherwise.
+    */
+   bool supportsAtomicSwap()
```
fjeremic: I added this query because on Z we can use compare and swap on all supported architecture levels where as atomic addition is only available on z196 onwards. We also need to be using the existing `supportsAtomicAdd` query in downstream projects.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -386,7 +386,40 @@ class OMR_EXTENSIBLE CodeGenerator
...
+    *
+    *     - atomicAdd32BitSymbol
+    *     - atomicAdd64BitSymbol
+    *     - atomicFetchAndAdd32BitSymbol
+    *     - atomicFetchAndAdd64BitSymbol
+    *
+    *  \return
+    *     \c true if intrinsics are supported; \c false otherwise.
+    */
+   bool supportsAtomicAdd()
```
0dvictor: I don't whether this query should cover support for all above 4 symbols. There may be a processor that supports one but not the other, and hence the codegen may choose to support one but not the other. Though I can't think of any current example, but the original 8086 processor supports only atomic add but not atomic fetch and add.

Why don't we make it a generic query like `bool supportsNonHelper(CommonNonhelperSymbol symbol)` so that support for each individual symbol can be queried?

compiler/codegen/OMRCodeGenerator.hpp
0dvictor: Similar to above, hardware support for "atomic swap" and "atomic compare and swap" varies, so that codegen may not support both.
In addition, "atomic swap" and "atomic compare and swap" are conceptional different as the later one is a conditional operation while the other isn't.
If we believe a generic query for each symbol isn't necessary, I would still request to separate the queries for "atomic swap" and "atomic compare and swap".

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: I definitely agree. That's a great idea. I'm working on this now.

compiler/compile/OMRSymbolReferenceTable.hpp
fjeremic: Fixed in 5da969d and 8d30aae

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Fixed in a7704d1

compiler/codegen/OMRCodeGenerator.hpp
fjeremic: Fixed in a7704d1

fjeremic:
@0dvictor could you please review again? All comments have been addressed.

compiler/compile/OMRSymbolReferenceTable.hpp
```diff
@@ -186,15 +186,75 @@ class SymbolReferenceTable
...
+       *  \code
+       *    temp = [address]
+       *    [address] = [address] + <value>
+       *    return temp
+       *  \endcode
+       *
+       *  The data type of \c <value> indicates the width of the operation.
+       */
+      atomicFetchAndAddSymbol,
       atomicFetchAndAdd32BitSymbol,
```
0dvictor: I'm assuming these legacy symbol are deprecated, they will be removed once we are satisfied that known downstream projects have adopted this change?

0dvictor:
Changes in common code and X86 CodeGen looks good to me.

liqunl:
I like the idea of having the query `supportsNonHelper`. Changes in common and X86 look good to me. But will there be a follow-up PR to add the write barrier and read barrier for Object type?

compiler/compile/OMRSymbolReferenceTable.hpp
fjeremic: That's exactly why they're still here. A subsequent PR will remove them once the downstream dependencies are gone.

fjeremic:
> But will there be a follow-up PR to add the write barrier and read barrier for Object type?

I'm not sure what you're referring to here. Could you elaborate what these are and where they should be added?

liqunl:
Sorry, what I really wanted to ask is: does current implementation support object fields that are reference type? Reading from and writing to such fields needs read/write barrier.

fjeremic:
> Sorry, what I really wanted to ask is: does current implementation support object fields that are reference type? Reading from and writing to such fields needs read/write barrier.

No it doesn't. According to the documentation an address must be provided to these atomic symbols. Presumably whoever is providing the address has ensured that it is correct, i.e. if the address was calculated from a field load whoever is providing the address must ensure the read barrier was done.

Our evaluators for the atomic symbols are simple. The onus is on whoever generates them to ensure they follow the semantics according to the documentation.

compiler/z/codegen/OMRTreeEvaluator.hpp
```diff
@@ -687,6 +687,69 @@ class OMR_EXTENSIBLE TreeEvaluator: public OMR::TreeEvaluator
 
    private:
 
+   /** \brief
```
r30shah: @fjeremic Changes in Z Codegen looks fine to me. Comments we have on SymbolReferenceTable makes it really easy to understand what we are inlining. Is it possible to have similar small explanation here as well, Just makes it easier to understand the code.

fjeremic: Fixed in 9b514d3

fjeremic:
All tagged reviews have approved. Off to @0xdaryl.

fjeremic:
@0xdaryl / @vijaysun-omr all suggested reviewers have approved. Any concerns from committers side?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -3066,27 +3066,37 @@ TR::Register *OMR::X86::TreeEvaluator::directCallEvaluator(TR::Node *node, TR::C
...
+         auto op = node->getChild(1)->getDataType().isInt32() ? XCHG4MemReg : XCHG8MemReg;
+
+         return inlineAtomicMemoryUpdate(node, op, cg);
+         }
+      else if (comp->getSymRefTab()->isNonHelper(SymRef, TR::SymbolReferenceTable::atomicSwap32BitSymbol))
          {
          return inlineAtomicMemoryUpdate(node, XCHG4MemReg, cg);
          }
-      if (comp->getSymRefTab()->isNonHelper(SymRef, TR::SymbolReferenceTable::atomicSwap64BitSymbol))
+      else if (comp->getSymRefTab()->isNonHelper(SymRef, TR::SymbolReferenceTable::atomicSwap64BitSymbol))
```
0xdaryl: I will postulate that it is far better for overall code footprint that you simply collect the opcode in each `else` block and then have one single call to `inlineAtomicMemoryUpdate` after the if/else ladder.  It looks like each call takes the same arguments except for the opcode.

fjeremic: Fixed in 5d7999b

fjeremic:
@0xdaryl any other concerns?

0xdaryl:
@genie-omr build all

0dvictor:
> I find the name NoHelper a bit confusing. I would rather called it InlineHelper.

There is an outstanding Issue (#3050) about renaming `NonHelper`. Though I believe everyone agreed to change its name, the final decision on its new name has not been made yet. The discussion was brought in one of complier architecture meetings. https://github.com/eclipse/omr/issues/3115

andrewcraik:
I don't think any special aliasing is required because we default to a conservative default method aliasing which should be conservative enough to prevent unwanted code motion (https://github.com/eclipse/omr/blob/9cf9e1c660823c3f2ad333c9e6aca9c791581051/compiler/il/Aliases.cpp#L390).

gita-omr:
Yes, I meant something more optimistic than the default aliasing. But I agree, it can be a separate feature.

PR #3000, opened 2018-09-27
AArch64: CMake build settings

knn-k:
This commit adds some changes to CMake build settings for aarch64.


knn-k:
1. The GNU toolchain for aarch64 does not accept the option `-m64`.  `gnu.cmake` is changed so that it adds `-m64` option only when the target processor is not `aarch64`.
2. Some C/C++ files in OMR expect a symbol `AARCH64` instead of `J9AARCH64` or `AARCH64GNU`.  Added `-DAARCH64` to `aarch64.cmake`.

knn-k:
You need a toolchain file for running `cmake` for cross-compilation.  Where should the toolchain file be in the OMR source tree?

knn-k:
Removed "WIP:".

0xdaryl:
Changes look reasonsonable to me.  Could one of @Leonardo2718 or @rwy0717 please eyeball them?

rwy0717:
Does `TR_HOST_ARCH` describe the build environment, or the runtime environment?

knn-k:
@rwy0717 
I am trying to build the JIT that runs on aarch64 by running a cross-compiling GCC on x86-64, and I have already built `triltest` that runs on aarch64 linux in my local environment using the changes in this PR.

I thought `TR_HOST_ARCH` was the build environment.
I remember the old J9 on x86 generated JXE files with AOT code for another processor.  Does the comment "Currently not doing cross, so assume HOST == TARGET" in `OmrCompilerSupport.cmake` mean such case ?

rwy0717:
OK, it appears that TR_HOST_ARCH is set but not used anywhere--it's meaningless. It's possible someone was thinking about producing  a cross-jit, but didn't get anywhere...

Where and how are you setting the TR_TARGET_ARCH? In a toolchain file?

knn-k:
`TR_TARGET_ARCH` is set in `cmake/modules/platform/arch/aarch64.cmake`.  Other platforms set `TR_HOST_ARCH` in `cmake/modules/platform/arch/*.cmake`.

knn-k:
Updated the files.

rwy0717:
Toolchain looks good.

I checked with others, and `host` definitely refers to the runtime environment. It looks like you've removed the check already 👍 

With the changes currently in the PR, are you still having trouble getting cmake to cross compile?

knn-k:
No, my trouble was solved.  I can build for aarch64.

rwy0717:
LGTM

0xdaryl:
@genie-omr build all

charliegracie:
@genie-omr build xcplinux

0xdaryl:
@charliegracie : What is the xcplinux build?  Some kind of cross-compile build?

charliegracie:
that was the temporary name I picked for the Linux x86-64 compressed references build... I have since switched it to xcrlinux as to not conflict with plinux.  Once we have all of the builds move to the OMR CI I will open an Issue to discuss how we want to name the builds so appropriate builds can be launched with the bot.

PR #3002, opened 2018-09-27
Restore Nursery allocation after aborted Scavenge

amicic:
Fixing a regression where an unnecessary scavenge cycle would be
triggered right after aborted scavenge (and percolate global GC), and
only then allocation in Nursery would be restored.
Typically, it was a really minor perf penalty (relative to already
expensive but also very infrequent aborted handling). But in some very
malicious cases if a subsequnt Scavenge also aborts, it would cause a
whole series of frequent aborts. 

The regression was introduced as a part of aborted Concurrent Scavenger
handling, but would affect just the standard (non-CS) Scavenger.

Also adding a lot of trace points in SubSpace allocate path.
SubSpace Generic path which is frequently executed (like TLH refresh),
has the lower trace level of 3 (hence, won't show by default), while the
rest of them, which are rather infreqent (a few per a GC cycle) has
default 1 (and will show by default).


charliegracie:
@genie-omr build all

charliegracie:
@youngar once the CI testing is finished can you give this a quick review and merge once you are happy with it?

Issue #3004, opened 2018-09-28
AArch64: Undefined reference error caused by OMRCompilerEnv


knn-k:
The file `compiler/aarch64/env/OMRCompilerEnv.hpp` was introduced for aarch64 by #2622, but it lacks the implementation of its `initializeTargetEnvironment()` function, and that causes an undefined reference error in linking an executable as shown below:
```
[100%] Linking CXX executable triltest
../../../jitbuilder/libjitbuilder.a(OMRCompilerEnv.cpp.o): In function `OMR::CompilerEnv::initialize()':
OMRCompilerEnv.cpp:(.text+0xe4): undefined reference to `OMR::ARM64::CompilerEnv::initializeTargetEnvironment()'
```
The ARM 32bit version implements its own flag in the function, and that is the only reason of having the `OMR::ARM::CompilerEnv` class.
I don't think AArch64 needs platform-specific `CompilerEnv` class at this point.  I propose to delete the file `OMRCompilerEnv.hpp` from the aarch64 tree.

PR #3005, opened 2018-09-28
AArch64: Remove env/OMRCompilerEnv.hpp from aarch64

knn-k:
This commit fixes the undefined reference problem in #3004.

Closes: #3004


0xdaryl:
An AArch64-specific change.  No need to run a genie build.  Merging.

PR #3008, opened 2018-09-28
Supporting rdbar/wrtbar changes for field watch

cathyzhyi:
This change adds field report helpers and jit resolve fields helpers, and add enableFieldWatch option

issue: #2812


compiler/optimizer/PartialRedundancy.cpp
```diff
@@ -230,6 +230,10 @@ int32_t TR_PartialRedundancy::perform()
    if (comp()->getProfilingMode() == JitProfiling && comp()->getHCRMode() != TR::none && _numProfilingsAllowed == 0)
       return 0;
 
+   // PRE is not supported for field watch yet
```
andrewcraik: I think this needs at least a reason - why isn't it supported, what is the TODO that needs to be addressed or the issue encountered?

vijaysun-omr:
Can an issue be created to refactor the code such that the query is asked of the language "runtime" in some sense ?

andrewcraik:
Ok so functionally I'm fine with this. The JIT helper names are a bit esoteric for my taste, but I can live with them. I do want a reason for PRE being disabled to be documented. I agree that we need an issue to factor the canGCandReturn query to facilitate language overrides since it may change between languages what the specific semantics of some nodes (like read and write barriers) are.

cathyzhyi:
@andrewcraik @vijaysun-omr Thanks for the suggestion. I am keeping the OMR layer of canGCAndReturn as before as the default implementation and move the checking for `readbar` and `wrtbar` into the sub projects.

compiler/optimizer/PartialRedundancy.cpp
cathyzhyi: added reason and TODO.

cathyzhyi:
opened issue #3026 to further cleanup canGCAndReturn.

vijaysun-omr:
What is so unique about PRE that it is disabled for readbar/writebar for field watch ? 
I mean, there are some other opts such as field privatizer for example that do some similar operations and I was wondering if those need to disabled too (or you have made the changes in those case) ?

andrewcraik:
+1 to @vijaysun-omr - the comment isn't saying why it is disabled - the answer to the above question is what is needed in the comment in addition to addressing the other opts which do similar things.

andrewcraik:
So it looks like this PR has now changed to just adding helpers and options - opt disables and other related changes have been removed. I'm fine with the helpers and options so review can continue anew on this. @cathyzhyi given the disable is removed can you just comment on if that is going to be in a separate PR or if it is un-needed now?

compiler/ras/Debug.cpp
```diff
@@ -3912,6 +3912,14 @@ TR_Debug::getRuntimeHelperName(int32_t index)
          case TR_stackOverflow:             return "jitStackOverflow";
          case TR_reportMethodEnter:         return "jitReportMethodEnter";
          case TR_reportStaticMethodEnter:   return "jitReportStaticMethodEnter";
+         case TR_jitReportInstanceFieldRead: return "jitReportInstanceField";
+         case TR_jitReportStaticFieldRead:  return "jitReportStaticFieldRead";
+         case TR_jitReportInstanceFieldWrite:  return "jitReportInstanceWrite";
```
0xdaryl: `"jitReportInstanceFieldWrite"` for consistency.

compiler/ras/Debug.cpp
```diff
@@ -3912,6 +3912,14 @@ TR_Debug::getRuntimeHelperName(int32_t index)
          case TR_stackOverflow:             return "jitStackOverflow";
          case TR_reportMethodEnter:         return "jitReportMethodEnter";
          case TR_reportStaticMethodEnter:   return "jitReportStaticMethodEnter";
+         case TR_jitReportInstanceFieldRead: return "jitReportInstanceField";
```
0xdaryl: `"jitReportInstanceFieldRead"`

compiler/ras/Debug.cpp
cathyzhyi: Thanks for pointing this out. Changed according to the suggestion.

compiler/ras/Debug.cpp
cathyzhyi: Thanks for pointing this out. Changed according to the suggestion.

cathyzhyi:
> So it looks like this PR has now changed to just adding helpers and options - opt disables and other related changes have been removed. I'm fine with the helpers and options so review can continue anew on this. @cathyzhyi given the disable is removed can you just comment on if that is going to be in a separate PR or if it is un-needed now?

It will be addressed in a separate PR.

andrewcraik:
I'm fine with this - @vijaysun-omr I think this should be good to test now. The scope has been reduced and the areas of concern delayed to another pull request.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Since there are no more (partial) optimizer changes in this PR and the remaining changes just add helpers and an option, I am fine with it.

vijaysun-omr:
@0xdaryl are you okay with the changes made ?

PR #3018, opened 2018-10-01
Add support for inlining new atomic methods on Power

hzongaro:
Added support for inlining of the new atomic intrinsics
atomic{AddAndFetch|FetchAndAdd|Swap}{32|64}BitSymbol on Power.  This is
largely a copy of similar support in OpenJ9 for inlining methods of
AtomicInteger and AtomicLong on Power.

Issue:  #2958


hzongaro:
May I ask @0dvictor @zl-wang @gita-omr to review?

One thing I'll point out is that I put `OMR::Power::CodeGenerator::inlineDirectCall` in `OMRTreeEvaluator.cpp` simply because the corresponding `J9::Power::CodeGenerator::inlineDirectCall` is defined in `J9TreeEvaluator.cpp`, rather than in `J9CodeGenerator.cpp`, which feels odd.

There will be a separate PR in OpenJ9 to take advantage of this change.

compiler/p/codegen/OMRCodeGenerator.hpp
```diff
@@ -171,6 +171,8 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
 
    bool getSupportsIbyteswap();
 
+   bool supportsAtomicAdd() {return true;}
```
0dvictor: FYI, https://github.com/eclipse/omr/pull/2999 is changing this query, it may need to be updated before merging.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+
+   TR::LabelSymbol *doneLabel = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
+   TR::LabelSymbol *loopLabel = TR::LabelSymbol::create(cg->trHeapMemory(),cg);
+
+   loopLabel->setStartInternalControlFlow();
+   deltaChild = node->getSecondChild();
+
+   if (deltaChild->getOpCode().isLoadConst()
+          && !deltaChild->getRegister()
+          && deltaChild->getDataType() == TR::Int32)
```
zl-wang: Could the second argument be 64bit constant and within [UPPER, LOWER] range at the same time? Don't want to miss that opt opportunity.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+   // We deviate from the VM helper here: no-store-no-barrier instead of always-barrier
+   generateInstruction(cg, TR::InstOpCode::sync, node);
+
+   TR::RegisterDependencyConditions *conditions;
+
+   //Set the conditions and dependencies
+   conditions = new (cg->trHeapMemory()) TR::RegisterDependencyConditions((uint16_t) numDeps, (uint16_t) numDeps, cg->trMemory());
+
+   addDependency(conditions, valueAddrReg, TR::RealRegister::NoReg, TR_GPR, cg);
+   conditions->getPreConditions()->getRegisterDependency(0)->setExcludeGPR0();
```
zl-wang: No need to exclude gr0 from this register, since you are explicitly putting it in the indexReg position.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+   if (isAddOp)
+      {
+      if (isDeltaImmediateShifted)
+         generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addis, node, deltaReg, resultReg, ((delta & 0xFFFF0000) >> 16));
+      else if (isDeltaImmediate)
+         generateTrg1Src1ImmInstruction(cg, TR::InstOpCode::addi, node, deltaReg, resultReg, delta);
+      else
+         generateTrg1Src2Instruction(cg, TR::InstOpCode::add, node, deltaReg, resultReg, deltaReg);
+      }
+
```
zl-wang: Maybe I misunderstood it, but AddAndFetch result should be the sum?  Here, you have the loaded value as the result always.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+      // If argument is an immediate value and operation is an add,
+      // it will be used as an immediate operand in an add immediate instruction
+      loadConstant(cg, node, delta, deltaReg);
+      }
+   else if (!isDeltaImmediate)
+      {
+      // For non-constant arguments, use evaluated register
+      // For non-immediate constants, evaluate since they may be re-used
+      numDeps++;
+      generateTrg1Src1Instruction(cg, TR::InstOpCode::mr, node, deltaReg, tempReg);
```
zl-wang: Should explicitly test if it will be re-used ... and you potentially can save this copy instruction.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+      delta = (int32_t)(deltaChild->getInt());
+      isDeltaConstant = true;
+
+      //determine if the constant can be represented as an immediate
+      if (delta <= UPPER_IMMED && delta >= LOWER_IMMED)
+         {
+         // avoid evaluating immediates for add operations
+         isDeltaImmediate = true;
+         }
+      else if (delta & 0xFFFF == 0 && (delta & 0xFFFF0000) >> 16 <= UPPER_IMMED && (delta & 0xFFFF0000) >> 16 >= LOWER_IMMED)
```
zl-wang: For a signed 32bit value, you can test  delta & 0xffff == 0  and don't need to test the latter two clauses (they are true always).

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,232 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+
+      if (doInline)
+         {
+         resultReg = inlineSimpleAtomicUpdate(node, isAddOp, isLong, isGetThenUpdate, cg);
+         }
+      }
+
+   return doInline;
+   }
+
```
zl-wang: High-level issue:  this will not work for 32bit JVM.  i.e.  this is assuming a 64bit JVM.

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Thanks, Julian @zl-wang .  Just to clarify, are you suggesting I remove just line 5308, or both lines 5308 and 5309?  Sorry for the basic questions - I'm not really familiar yet with how register dependencies are specified or how they work.

zl-wang: @hzongaro you can eliminate both 5308 and 5309.

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: @zl-wang Sorry for the confusion, Julian.  At lines #5338-5352, the code will return `resultReg` for `FetchAndAdd` and `deltaReg` for `AddAndFetch` - that's based on the value of `isGetThenUpdate`.  Would it help if I changed the name of `resultReg` to `prevValueReg`, say, and renamed `deltaReg` to `updatedValueReg`?

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: @zl-wang Julian, thanks for catching this.  Just so I'm clear, can this code be inlined on a 32 bit JVM when it's working with 32 bit operands, and inlined on a 64 bit JVM when it's working with either 32 bit and 64 bit operands?  Or can it only be inlined on a 64 JVM ever?

Something I just noticed this morning, I was playing with `AtomicLong.getAndAdd` and `AtomicInteger.getAndAdd` without my changes for these new intrinsics, and it looks like inline code is never generated for `AtomicLong.getAndAdd` for 32 bit or 64 bit JVMs, and always generated for `AtomicInteger.getAndAdd` for both 32 bit and 64 bit JVMs.  Is that expected?

zl-wang: @hzongaro I didn't expect that.  I expected existing JVM can inline CAS of both 32/64 operands in both 32/64 bit JVM.  The only assumption is the underlying CPU is 64bit (this has been true since java7.1 when we discontinued support of 32bit hardware).  The case for inlining CAS of 64bit operand in 32bit JVM needs special care though ... you can assemble 64bit operand in a register within the CAS range only, depending on the fact that CAS will definitely fail if there was a signal delivered in the between (i.e.  meaning the upper 32bit of the register can be trashed anytime).

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Sounds good.  Thanks, Julian!

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: @zl-wang Julian, I assume you mean only for the non-immediate constant case.  I think you're suggesting that in the case of a non-immediate constant, it should evaluate the constant in the loop, but for a non-constant, it should be evaluated outside of the loop, the value copied into another register to using an `mr`.  Do I have that right?

Now that I'm thinking about it, why does this code try to reuse the register that holds delta to hold the sum. . . ?

zl-wang: @hzongaro  certainly the new sequence is better.  previously, my comment, looking at the mr instruction, meant to say why not testing the delta child referenceCount to decide if delta can be reused.  In your new sequence, if  deltaChild->getReferenceCount() == 1, you can even eliminate the creation of GPR_88 and use GPR_99 in its place.  But this is not strictly necessary.

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: @zl-wang Julian, I've been working on a revision that handles 64-bit values running in a 32-bit JVM, but it makes the code quite ugly.  I’ll pull that out as a separate pull request, and I’ll let you decide whether it’s an important enough case to handle, as well as point out any problems with the way that I tackled the problem.

compiler/p/codegen/OMRCodeGenerator.hpp
hzongaro: Revised in commit https://github.com/eclipse/omr/pull/3018/commits/8077bb18b5170cb1e62f23fef75258c0dc006d5d

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Addressed in commit https://github.com/eclipse/omr/pull/3018/commits/e8493aa681a66b85e9d83cca9cbe1f7dfdda4c1b

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Fixed up variable names to clarify what they hold.  Revised in commit https://github.com/eclipse/omr/pull/3018/commits/e8493aa681a66b85e9d83cca9cbe1f7dfdda4c1b

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Revised generated code so that the `mr` in the loop is no longer used.  Changed in commit https://github.com/eclipse/omr/pull/3018/commits/e8493aa681a66b85e9d83cca9cbe1f7dfdda4c1b

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Eliminated the redundant test.  Fixed in commit https://github.com/eclipse/omr/pull/3018/commits/e8493aa681a66b85e9d83cca9cbe1f7dfdda4c1b

hzongaro:
@zl-wang @fjeremic Julian, Filip - I've made revisions that I think addresses your review comments.  May I ask you to re-review when you have some time?  The one exception is handling 64-bit values on a 32-bit JVM, which I will submit in a follow up PR.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+      // If argument is immediate, but the operation is not an add,
+      // the value must still be loaded into a register
+      // If argument is an immediate value and operation is an add,
+      // it will be used as an immediate operand in an add immediate instruction
+      numDeps++;
+      deltaReg = cg->allocateRegister();
+      loadConstant(cg, node, delta, deltaReg);
+      }
+
+   uint8_t len = isLong ? 8 : 4;
```
fjeremic: This should be `const`

compiler/p/codegen/OMRCodeGenerator.hpp
```diff
@@ -384,6 +386,9 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    bool supportsFusedMultiplyAdd() {return true;}
    bool supportsNegativeFusedMultiplyAdd() {return true;}
 
+   bool supportsNonHelper(TR::SymbolReferenceTable::CommonNonhelperSymbol symbol
+);
```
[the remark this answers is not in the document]
hzongaro: Oops!  Didn't notice that working with an 80 column window.

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Corrected in commit https://github.com/eclipse/omr/pull/3018/commits/cf67e3ee2bc9ec34b0170de4e4dc302555d71b9a

compiler/p/codegen/OMRCodeGenerator.hpp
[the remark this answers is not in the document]
hzongaro: Corrected in commit https://github.com/eclipse/omr/pull/3018/commits/cf67e3ee2bc9ec34b0170de4e4dc302555d71b9a

fjeremic:
@hzongaro I think you accidentally checked in some huge file as part of your latest commit.

compiler/p/code
```diff
@@ -0,0 +1,3682 @@
...
+         }
+      instructionCursor = instructionCursor->getNext();
+      }
+   }
+// end of PPC Codegen Peephole routines
+
+bool OMR::Power::CodeGenerator::supportsAESInstructions()
+   {
+    if ( TR::Compiler->target.cpu.getPPCSupportsAES() && !self()->comp()->getOption(TR_DisableAESInHardware))
+      ret
```
zl-wang: This looks like you accidentally copied OMRCodeGenerator.cpp to this compiler/p/code file?

compiler/p/codegen/OMRCodeGenerator.cpp
```diff
@@ -3660,3 +3660,23 @@ bool OMR::Power::CodeGenerator::getSupportsIbyteswap()
...
+OMR::Power::CodeGenerator::supportsNonHelper(TR::SymbolReferenceTable::CommonNonhelperSymbol symbol)
+   {
+   bool result = false;
+
+   switch (symbol)
+      {
+      case TR::SymbolReferenceTable::atomicAddSymbol:
+      case TR::SymbolReferenceTable::atomicFetchAndAddSymbol:
+      case TR::SymbolReferenceTable::atomicSwapSymbol:
+         {
```
zl-wang: Does it matter that you missed cases here of data-type-specific symbols, e.g., atomicSwap32BitSymbol?

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+   bool reuseDeltaReg = false;
+
+   int32_t numDeps = 3;
+
+   int32_t delta = 0;
+
+   // Memory barrier --- NOTE: we should be able to do a test upfront to save this barrier,
+   //                          but Hursley advised to be conservative due to lack of specification.
+   generateInstruction(cg, TR::InstOpCode::lwsync, node);
+
```
zl-wang: I am wondering these symbols are only intended to map to the corresponding Unsafe methods.  If yes, the note on memory barriers is justified.  If not, the note is pessimistic regarding memory barriers.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+   loopLabel->setStartInternalControlFlow();
+   deltaChild = node->getSecondChild();
+
+   // Determine whether immediate instruction can be used.  For 64 bit operand,
+   // value has to fit in 32 bits to be considered for immediate operation
+   if (deltaChild->getOpCode().isLoadConst()
+          && !deltaChild->getRegister()
+          && (deltaChild->getDataType() == TR::Int32
+                 || (deltaChild->getDataType() == TR::Int64
+                        && deltaChild->getInt() == deltaChild->getLongInt())))
```
zl-wang: This test is most likely incorrect to one of the BigEndian or LittleEndian cases, i.e.  it cannot be correct for both BE and LE.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+      {
+      addDependency(conditions, deltaReg, TR::RealRegister::NoReg, TR_GPR, cg);
+      numDeps--;
+      }
+
+   doneLabel->setEndInternalControlFlow();
+   generateDepLabelInstruction(cg, TR::InstOpCode::label, node, doneLabel, conditions);
+
+   cg->decReferenceCount(valueAddrChild);
+   cg->stopUsingRegister(valueAddrReg);
```
zl-wang: As a general rule, you don't need to stop-using a register if it is carried in a node, since decReferenceCount on the node automatically takes care of the liveness.  Otherwise, if the node is referenced by multiple nodes, you run the risk of prematurely ending its life.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+   cg->stopUsingRegister(cndReg);
+
+   if (deltaReg)
+      {
+      cg->stopUsingRegister(deltaReg);
```
zl-wang: Same issue here as above.

compiler/p/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5189,6 +5191,262 @@ TR::Register *OMR::Power::TreeEvaluator::directCallEvaluator(TR::Node *node, TR:
...
+      }
+
+   if (deltaChild)
+      {
+      cg->decReferenceCount(deltaChild);
```
zl-wang: Furthermore, deltaReg might be returned as resultReg.  You had better setRegister before this decReferenceCount, because it can prematurely end that register's life (if this were the last reference to deltaChild).

compiler/p/codegen/OMRCodeGenerator.cpp
hzongaro: Thanks for raising that question.  I based this on the implementations of `OMR::X86::CodeGenerator::supportsNonHelper` and `OMR::Z::CodeGenerator::supportsNonHelper`.  @0dvictor @fjeremic Victor, Filip - should `supportsNonHelper` include the 32-bit and 64-bit-specific nonhelper symbols?

compiler/p/codegen/OMRTreeEvaluator.cpp
0dvictor: They are not. Unsafe is only one of many use cases of these symbols; other use cases include automatic reference count in Swift, as well as similar synchronization helpers for other languages.

compiler/p/codegen/OMRCodeGenerator.cpp
0dvictor: I believe the data-type-specific ones are deprecated and will be deleted soon. AFAIK, none of them is in-use as of now.

compiler/p/codegen/OMRTreeEvaluator.cpp
fjeremic: I think an assert would have caught this in DEBUG mode. Another reason why we need to start moving towards running testing with asserts enabled like we used to in the past.

compiler/p/codegen/OMRCodeGenerator.cpp
hzongaro: Thank you!

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Fixed in https://github.com/eclipse/omr/pull/3018/commits/a453a47d5b03cd028f0c9be8e0ed55b752ea2d70

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Thanks, Julian!  That really clarifies for me the proper use of those methods.  Fixed in https://github.com/eclipse/omr/pull/3018/commits/a453a47d5b03cd028f0c9be8e0ed55b752ea2d70

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Fixed in https://github.com/eclipse/omr/pull/3018/commits/a453a47d5b03cd028f0c9be8e0ed55b752ea2d70

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: Fixed in https://github.com/eclipse/omr/pull/3018/commits/a453a47d5b03cd028f0c9be8e0ed55b752ea2d70

compiler/p/codegen/OMRTreeEvaluator.cpp
hzongaro: I spoke with Julian, and he indicated that his comment was about the use of memory barriers themselves rather than about the comment in the code.  He indicated that as these symbols are used for `Unsafe` (in addition to other uses) the memory barriers are required.

hzongaro:
I believe all review comments have been addressed, so I will remove "WIP" from the title.

fjeremic:
@genie-omr build plinux,aix

PR #3019, opened 2018-10-02
Provide classes to record JitBuilder API calls

bragaigor:
Recorder is part of the record and replay functionality.
The ultimate goal for such feature is to remove the JIT from
runtimes and provide it as a service. The recorder part of this
feature, as the name implies, will record the IL at the client side
and send it to be compiled in the server. The server then
replays/parses the IL and compiles it, to then send machine code
back to the client so that it can execute the requested IL. The IL
contains services that have recording support (e.g. Add, Sub, Xor,
Store, Loop, etc. More to come). This commit is dedicated to the
Recorder initial functionality. More details as follows.

JitBuilderRecorder defines the basic structure of a JitBuilder IL
(ljbil) representation. The following files provide implementations
that write this representation to a binary buffer, binary file, text
file, respectively:
OMRJitBuilderRecorderBinaryBuffer
OMRJitBuilderRecorderBinaryFile
OMRJitBuilderRecorderTextFile

To enable replay compilation and a host of other useful facilities,
I have implemented a very simple format that can capture the calls
made to JitBuilder services on builder objects in a readily persisted
(stored to disk, saved in buffers, etc) form.

More details:

This commit defines the basic structure of a JitBuilder IL (ljbil)
representation along with representations as a binary buffer,
binary file, and text file.

The fundamental elements of the jbil representation are (currently):
	String (length then characters)
	Number (8, 16, 32, 64, float, double)
	ID (32-bit)
		Statement
		Type
		Value
		Builder
		Location

An ID is just a number used to identify different Strings, Builder
objects, Values, Types, Locations, and Statements. A std::map is
used to associate an ID with pointers corresponding to these
different categories. Rather than embed pointers into the jbil
output, we embed IDs.

IlBuilder, MethodBuilder, and BytecodeBuilder objects will need
to create some jbil output for each service that is called on
these objects. New objects need to be assigned IDs (by calling
StoreID() ) before they can be referenced by their ID.

Each time an IlBuilder object is created, for example, the NewIlBuilder
can be used to assign it an ID, and that ID can be referenced by
subsequent statements to call new services on it.

Every jbil output starts with (see start() function):
	String(RECORDER_SIGNATURE); // 4 (16 bits) "JBIL"
	Number(VERSION_MAJOR);      // 0 (16 bits)
	Number(VERSION_MINOR);      // 0 (16 bits)
	Number(VERSION_PATCH);      // 0 (16 bits)

This header can be encoded in as little as 12 bytes. This version
can be used to understand how the jbil was produced and enables
processing tools to precisely understand the format of the jbil.

Following the header is a sequence of statements.  There are only
two predefined statements. The first predefined statement assigns
an ID to a string:
	0 "Statement String"

All statements begin with the builder object they apply to.
The predefined statement references the special builder ID
0 which cannot be used by any other builder object. This
statement causes the provided string to be assigned the
next available ID (starting at 1). Subsequent statements
are all of the form:

	<BuilderID> <StatementID> <required values>

Depending on the statement, any number of additional
values may be required, but each statement has a
particular set of values it requires.

The statement strings are defined in StatementNames.hpp
as STATEMENT_<statement name> . For example, STATEMENT_LOAD
represents a call to IlBuilder::Load() while STATEMENT_ADD
represents a call to IlBuilder::Add().

The sequence of statements ends when the second predefined
statement is encountered using the special Builder(1)
object followed by the JBIL_COMPLETE string ("Done");

At the end of the representation, the following trailer is
added (see end() function):
	Builder(1);
	String(JBIL_COMPLETE);      // 4 (16 bits) "Done"

This commit doesn't show how IlBuilder, MethodBuilder, or
BytecodeBuilder can use JitBuilderRecorder to record
jbil; that will come in later commits.

Co-authored-by: Sharon Wang <sharon-wang-cpsc@outlook.com>
Co-authored-by: Mark Stoodley <mstoodle@ca.ibm.com>
Co-authored-by: Andrew Young <youngar17@gmail.com>
Co-authored-by: Charlie Gracie <charlie.gracie@gmail.com>

### DIFF
--- a/compiler/ilgen/JitBuilderRecorder.hpp
+++ b/compiler/ilgen/JitBuilderRecorder.hpp
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef TR_JITBUILDERRECORDER_INCL
+#define TR_JITBUILDERRECORDER_INCL
+
+#include "ilgen/OMRJitBuilderRecorder.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorder : public OMR::JitBuilderRecorder
+      {
+      public:
+         JitBuilderRecorder(const TR::MethodBuilder *mb, const char *fileName)
+            : OMR::JitBuilderRecorder(mb, fileName)
+            { }
+         virtual ~JitBuilderRecorder()
+            { }
+      };
+
+} // namespace TR
+
+#endif // !defined(TR_JITBUILDERRECORDER_INCL)
--- a/compiler/ilgen/JitBuilderRecorderBinaryBuffer.hpp
+++ b/compiler/ilgen/JitBuilderRecorderBinaryBuffer.hpp
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef TR_JITBUILDERRECORDER_BINARYBUFFER_INCL
+#define TR_JITBUILDERRECORDER_BINARYBUFFER_INCL
+
+#include "ilgen/OMRJitBuilderRecorderBinaryBuffer.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderBinaryBuffer : public OMR::JitBuilderRecorderBinaryBuffer
+      {
+      public:
+         JitBuilderRecorderBinaryBuffer(const TR::MethodBuilder *mb, const char *fileName)
+            : OMR::JitBuilderRecorderBinaryBuffer(mb, fileName)
+            { }
+         virtual ~JitBuilderRecorderBinaryBuffer()
+            { }
+      };
+
+} // namespace TR
+
+#endif // !defined(TR_JITBUILDERRECORDER_BINARYBUFFER_INCL)
--- a/compiler/ilgen/JitBuilderRecorderBinaryFile.hpp
+++ b/compiler/ilgen/JitBuilderRecorderBinaryFile.hpp
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef TR_JITBUILDERRECORDER_BINARYFILE_INCL
+#define TR_JITBUILDERRECORDER_BINARYFILE_INCL
+
+#include "ilgen/OMRJitBuilderRecorderBinaryFile.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderBinaryFile : public OMR::JitBuilderRecorderBinaryFile
+      {
+      public:
+         JitBuilderRecorderBinaryFile(const TR::MethodBuilder *mb, const char *fileName)
+            : OMR::JitBuilderRecorderBinaryFile(mb, fileName)
+            { }
+         virtual ~JitBuilderRecorderBinaryFile()
+            { }
+      };
+
+} // namespace TR
+
+#endif // !defined(TR_JITBUILDERRECORDER_BINARYFILE_INCL)
--- a/compiler/ilgen/JitBuilderRecorderTextFile.hpp
+++ b/compiler/ilgen/JitBuilderRecorderTextFile.hpp
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef TR_JITBUILDERRECORDER_TEXTFILE_INCL
+#define TR_JITBUILDERRECORDER_TEXTFILE_INCL
+
+#include "ilgen/OMRJitBuilderRecorderTextFile.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderTextFile : public OMR::JitBuilderRecorderTextFile
+      {
+      public:
+         JitBuilderRecorderTextFile(const TR::MethodBuilder *mb, const char *fileName)
+            : OMR::JitBuilderRecorderTextFile(mb, fileName)
+            { }
+         virtual ~JitBuilderRecorderTextFile()
+            { }
+      };
+
+} // namespace TR
+
+#endif // !defined(TR_JITBUILDERRECORDER_TEXTFILE_INCL)
--- a/compiler/ilgen/OMRJitBuilderRecorder.cpp
+++ b/compiler/ilgen/OMRJitBuilderRecorder.cpp
@@ -1,0 +1,158 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "infra/Assert.hpp"
+#include "ilgen/JitBuilderRecorder.hpp"
+
+namespace TR { class MethodBuilder; }
+
+OMR::JitBuilderRecorder::JitBuilderRecorder(const TR::MethodBuilder *mb, const char *fileName)
+: _mb(mb), _nextID(0), _idSize(8),  _file(fileName, std::fstream::out | std::fstream::trunc)
+   {
+   start(); // initializes IDs 0 and 1 (reserved)
+   }
+
+OMR::JitBuilderRecorder::~JitBuilderRecorder()
+   {
+   }
+
+void
+OMR::JitBuilderRecorder::start()
+   {
+   // special reserved value, must do it first !
+   StoreID(0);
+
+   // another special reserved value, so do it first
+   StoreID((const void *)1);
+
+   String(StatementName::RECORDER_SIGNATURE);
+   Number(StatementName::VERSION_MAJOR);
+   Number(StatementName::VERSION_MINOR);
+   Number(StatementName::VERSION_PATCH);
+   EndStatement();
+   }
+
+void 
+OMR::JitBuilderRecorder::Close()                                       
+   { 
+   end();
+   EndStatement();
+   }
+
+OMR::JitBuilderRecorder::TypeID
+OMR::JitBuilderRecorder::getNewID()
+   {
+   // support for variable sized ID encoding
+   //  to avoid any synchronization issues in how decoders/encoders count IDs, use a statement to signal change
+   // check if we're close to 8-bit / 16-bit boundary
+   // need to leave space for one ID because these statements will not have been defined yet
+   //  and defining the statement needs an ID!
+   if (_nextID == (1 << 8) - 2)
+      {
+      _idSize = 16;
+      Statement(StatementName::STATEMENT_ID16BIT);
+      }
+   else if (_nextID == (1 << 16) - 2)
+      {
+      _idSize = 32;
+      Statement(StatementName::STATEMENT_ID32BIT);
+      }
+
+   return _nextID++;
+   }
+
+OMR::JitBuilderRecorder::TypeID
+OMR::JitBuilderRecorder::myID()
+   {
+   return lookupID(_mb);
+   }
+
+bool
+OMR::JitBuilderRecorder::knownID(const void *ptr)
+   {
+   TypeMapID::iterator it = _idMap.find(ptr);
+   return (it != _idMap.end());
+   }
+
+OMR::JitBuilderRecorder::TypeID
+OMR::JitBuilderRecorder::lookupID(const void *ptr)
+   {
+   TypeMapID::iterator it = _idMap.find(ptr);
+   if (it == _idMap.end())
+      TR_ASSERT_FATAL(0, "Attempt to lookup a builder object that has not yet been created");
+
+   TypeID id = it->second;
+   return id;
+   }
+
+void
+OMR::JitBuilderRecorder::ensureStatementDefined(const char *s)
+   {
+   if (knownID(s))
+      return;
+
+   StoreID(s);
+   Builder(0);
+   Statement(s);
+   String(s);
+   EndStatement();
+   }
+
+void
+OMR::JitBuilderRecorder::end()
+   {
+   ID(1); // reserved ID to indicate end of file
+   String(StatementName::JBIL_COMPLETE);
+   }
+
+void
+OMR::JitBuilderRecorder::BeginStatement(const char *s)
+   {
+   BeginStatement(_mb, s);
+   }
+
+void
+OMR::JitBuilderRecorder::BeginStatement(const TR::MethodBuilder *b, const char *s)
+   {
+   ensureStatementDefined(s);
+   Builder(b);
+   Statement(s);
+   }
+
+void
+OMR::JitBuilderRecorder::StoreID(const void *ptr)
+   {
+   TypeMapID::iterator it = _idMap.find(ptr);
+   if (it != _idMap.end())
+      TR_ASSERT_FATAL(0, "Unexpected ID already defined for address %p", ptr);
+
+   _idMap.insert(std::make_pair(ptr, getNewID()));
+   }
+
+bool
+OMR::JitBuilderRecorder::EnsureAvailableID(const void *ptr)
+   {
+   if (knownID(ptr)) // already available, so just return
+      return true;
+
+   StoreID(ptr);
+   return false; // ID was not available, but is now
+   }
--- a/compiler/ilgen/OMRJitBuilderRecorder.hpp
+++ b/compiler/ilgen/OMRJitBuilderRecorder.hpp
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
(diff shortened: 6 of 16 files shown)


compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,212 @@
...
+   virtual void BeginStatement(const char *s);
+   virtual void EndStatement()                                { }
+
+   void StoreID(const void *ptr);
+   bool EnsureAvailableID(const void *ptr);
+
+   /**
+    * @brief constant strings used to define output format
+    */
+   const char *STATEMENT_DEFINENAME                   = "DefineName";
```
charliegracie: To fix the Windows compile failure all of these fields need to be static.
`static const char *STATEMENT_DEFINENAME                   = "DefineName";`

but to be even more pedantic I think it makes more sense to declare them as const pointers to const data.
`static const char *const STATEMENT_DEFINENAME                   = "DefineName";`

CMakeLists.txt
```diff
@@ -156,6 +156,11 @@ if(OMR_JITBUILDER)
 	add_subdirectory(jitbuilder)
 endif(OMR_JITBUILDER)
 
+if(OMR_IMPERIUM)
```
charliegracie: Can you remove this from the commit since the imperium folder does not exist?

charliegracie:
@bragaigor can you resolve the merge issue? @mstoodle  or @Leonardo2718  do you have any suggestions to resolve the compilation issue on Windows?

compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,212 @@
...
+namespace TR { class IlType; }
+namespace TR { class IlValue; }
+
+namespace OMR
+{
+
+class JitBuilderRecorder
+   {
+   public:
+   const int16_t VERSION_MAJOR=0;
```
Leonardo2718: This kind of initialization is not allowed prior to C++11.

charliegracie: but is it available in extended0x... the rest of the Compiler code is making use of extended0x so I see no reason to block this...

charliegracie:
For the `compiler/ilgen/OMRJitBuilderRecorder.hpp` I think the list of const char * variables should be moved into a separate file.. Envisioning the Replay side of this work you will need to have exactly the same list...

Similarly I think the VERSION information should also move... 

Maybe we need a new class that has a list of instance variables that are all initialized in a helper function called from the constructor instead of being set directly.

compiler/ilgen/OMRJitBuilderRecorder.hpp
Leonardo2718: Hm, yes you're right. For some reason, I thought this was the cause of the build failure. Looking at it again, it's clearly not :sweat_smile: .

youngar:
@genie-omr build all

charliegracie:
Thanks for fixing the compile issue! @mstoodle , @0xdaryl and @Leonardo2718 now that everything is working can we get some reviews so we can get this delivered... The follow on PR for the text file replay is already opened and getting stale as well... The sooner these 2 PRs get in the sooner the OMR Compiler will be able to support a JITaaS.

.vscode/settings.json
```diff
@@ -0,0 +1,17 @@
...
+        "new": "cpp",
+        "stdexcept": "cpp",
+        "type_traits": "cpp",
+        "typeinfo": "cpp",
+        "algorithm": "cpp",
+        "functional": "cpp",
+        "tuple": "cpp",
+        "utility": "cpp"
+    }
+}
```
fjeremic: Similarly to #3056 I'm not sure we should check in user VSCode settings into the repo.

compiler/compile/OMRCompilation.hpp
```diff
@@ -1225,6 +1226,8 @@ class OMR_EXTENSIBLE Compilation
 
    BitVectorPool _bitVectorPool; //MUST be declared after _trMemory
 
+   bool _shouldCompile;
```
fjeremic: Can we please Doxygen document what this field is used for? A naive reader such as myself may be wondering why someone may instantiate an `OMR::Compilation` object only to "not compile".

charliegracie:
@bragaigor I believe you have another branch on your repo that contains an updated versions that includes the merge. I believe at this point it is likely best just to push that code to this branch to facilitate reviews.

Leonardo2718:
I've posted a comment in https://github.com/eclipse/omr/issues/1818#issuecomment-441317606 that describes at a (very) high-level an alternative design for the recording mechanism @mstoodle and I have been discussing.

Essentially, instead of having `IlBuilderRecorder`, `MethodBuilderRecorder`, and `BytecodeBuilderRecorder` be part of the IlBuilder class heirarchy, they would become separate classes that are generated in a similar way as the C++ client API. These generated classes would act as listeners for `IlBuilder`, `MethodBuilder`, and `BytecodeBuilder`, respectively, following the Listener/Observer design pattern. Other parts of the recorder implementation, such as `JitBuilderRecorder` would stay the same.

mstoodle:
Given that new high level picture, I wonder if we could pull out, say, the `JitBuilderRecorder` classes, from this PR so that we can get those into the code base? I think we're going to have to do some different things to get the "Record" function merged, which @Leonardo2718 and I will work on. I think we would then reconstruct the replay functionality (which I know isn't in this PR) in the same way using the generator approach we're going to build for the record function, just structuring the result differently.

@bragaigor would you be willing to take on that first part to get the `JitBuilderRecorder` contributed, at least? (the other parts are much less well structured at the moment so won't ask you to do anything there for the time being)

.gitignore
```diff
@@ -188,6 +188,7 @@ subscriberAgentWithJ9threadTrace2.out
 tfileTest*.tst
 traceLogTest.trc
 VerboseGC_*.xml*
+*.ilt
```
mstoodle: What's an "ilt" file?

compiler/compile/Compilation.hpp
```diff
@@ -49,7 +49,8 @@ class OMR_EXTENSIBLE Compilation : public OMR::CompilationConnector
          TR::Options &options,
          TR::Region &heapMemoryRegion,
          TR_Memory *memory,
-         TR_OptimizationPlan *optimizationPlan) :
+         TR_OptimizationPlan *optimizationPlan,
+         bool shouldCompile = true) :
```
mstoodle: Would you please also remove this `shouldCompile` support from the pull request? With the listener approach, I don't *think* it will be needed (which is another advantage of that approach).

compiler/ilgen/JitBuilderRecorderBinaryBuffer.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define TR_JITBUILDERRECORDER_BINARYBUFFER_INCL
+
+#include "ilgen/OMRJitBuilderRecorderBinaryBuffer.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderBinaryBuffer : public OMR::JitBuilderRecorderBinaryBuffer
+      {
+      public:
+      //    JitBuilderRecorderBinaryBuffer(const TR::MethodBuilderRecorder *mb)
```
mstoodle: Rather than commenting out the intended call because there is no longer a `TR::MethodBuilderRecorder` class, please just change it to plain old `TR::MethodBuilder` because I think that's what it should be anyway.

compiler/ilgen/JitBuilderRecorderBinaryFile.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define TR_JITBUILDERRECORDER_BINARYFILE_INCL
+
+#include "ilgen/OMRJitBuilderRecorderBinaryFile.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderBinaryFile : public OMR::JitBuilderRecorderBinaryFile
+      {
+      public:
+      //    JitBuilderRecorderBinaryFile(const TR::MethodBuilderRecorder *mb, const char *fileName)
```
mstoodle: same here: please just use `TR::MethodBuilder`

compiler/ilgen/JitBuilderRecorderTextFile.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define TR_JITBUILDERRECORDER_TEXTFILE_INCL
+
+#include "ilgen/OMRJitBuilderRecorderTextFile.hpp"
+
+namespace TR
+{
+   class JitBuilderRecorderTextFile : public OMR::JitBuilderRecorderTextFile
+      {
+      public:
+      //    JitBuilderRecorderTextFile(const TR::MethodBuilderRecorder *mb, const char *fileName)
```
mstoodle: and here...I'll stop pointing it out now, but please change all the places where you commented the "correct" thing out and just remove the `Recorder` name in the class.

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+   {
+   // support for variable sized ID encoding
+   //  to avoid any synchronization issues in how decoders/encoders count IDs, use a statement to signal change
+   // check if we're close to 8-bit / 16-bit boundary
+   // need to leave space for one ID because these statements will not have been defined yet
+   //  and defining the statement needs an ID!
+   if (_nextID == (1 << 8) - 2)
+      {
+      _idSize = 16;
+      Statement(StatementName::STATEMENT_ID16BIT); // TODO
```
mstoodle: what's "TODO" about this ?

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+      Statement(StatementName::STATEMENT_ID32BIT); // TODO
+      }
+
+   return _nextID++;
+   }
+
+OMR::JitBuilderRecorder::TypeID
+OMR::JitBuilderRecorder::myID()
+   {
+//    return lookupID(_mb);
```
mstoodle: please uncomment and fix these ones too :) .

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+   }
+
+void
+OMR::JitBuilderRecorder::ensureStatementDefined(const char *s)
+   {
+   if (knownID(s))
+      return;
+
+   StoreID(s);
+   // Builder(0);
```
mstoodle: is this needed? If it isn't please delete it

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+void
+OMR::JitBuilderRecorder::end()
+   {
+   ID(1);
+   String(StatementName::JBIL_COMPLETE);
+   }
+
+// void
+// OMR::JitBuilderRecorder::BeginStatement(const char *s)
```
mstoodle: needed? if not, please remove

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+void
+OMR::JitBuilderRecorder::StoreID(const void *ptr)
+   {
+   TypeMapID::iterator it = _idMap.find(ptr);
+   if (it != _idMap.end())
+      TR_ASSERT_FATAL(0, "Unexpected ID already defined for address %p", ptr);
+
+   _idMap.insert(std::make_pair(ptr, getNewID()));
+   // std::cout << "map: " << _idMap << '\n';
```
mstoodle: please remove these comments

compiler/ilgen/JitBuilderRecorder.hpp
```diff
@@ -0,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 IBM Corp. and others
```
mstoodle: I think these files should really be 2018, 2018

compiler/ilgen/JitBuilderRecorderBinaryFile.hpp
```diff
@@ -0,0 +1,43 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 IBM Corp. and others
```
mstoodle: another that should be 2018, 2018. Please change all the new files to use these copyright dates.

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include <stdint.h>
+#include <iostream>
```
mstoodle: are these headers all needed for the code in this file? If they aren't, please remove them. i'm particularly interested in iostream and fstream which I suspect are only here because of the commented out `std::cout` calls (that I've also asked that you remove :) ).

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <fstream>
+
+#include "infra/Assert.hpp"
+#include "ilgen/JitBuilderRecorder.hpp"
+// #include "ilgen/MethodBuilderRecorder.hpp"
```
mstoodle: I don't think you should need to `#Include` even `MethodBuilder.hpp` here, since I think you only use it as a pointer and so you can just forward declare the type in the header file.

compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,97 @@
...
+ *******************************************************************************/
+
+#ifndef OMR_JITBUILDERRECORDER_INCL
+#define OMR_JITBUILDERRECORDER_INCL
+
+#include <iostream>
+#include <fstream>
+#include <map>
+#include "ilgen/StatementNames.hpp"
+// #include "ilgen/VirtualMachineState.hpp"
```
mstoodle: please remove: I don't think you need it

compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include "ilgen/IlInjector.hpp"
```
mstoodle: I don't think you need `IlInjector.hpp` in the header and possibly not at all. If you do need it, please move to the `.cpp`

compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+namespace TR { class IlBuilderRecorder; }
```
mstoodle: please delete this forward declaration of `IlBuilderRecorder`.

compiler/ilgen/OMRJitBuilderRecorder.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef OMR_JITBUILDERRECORDER_INCL
+#define OMR_JITBUILDERRECORDER_INCL
+
+#include <iostream>
+#include <fstream>
```
mstoodle: Not sure you need `iostream` and `fstream` here, do you?

compiler/ilgen/OMRJitBuilderRecorderBinaryBuffer.hpp
```diff
@@ -0,0 +1,67 @@
...
+ *******************************************************************************/
+
+#ifndef OMR_JITBUILDERRECORDER_BINARYBUFFER_INCL
+#define OMR_JITBUILDERRECORDER_BINARYBUFFER_INCL
+
+#include "ilgen/JitBuilderRecorder.hpp"
+#include <vector>
+
+namespace TR { class IlBuilderRecorder; }
+// namespace TR { class MethodBuilderRecorder; }
```
mstoodle: please delete this reference and the one to `IlBuilderRecorder`.

compiler/ilgen/OMRJitBuilderRecorderBinaryFile.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *******************************************************************************/
+
+#ifndef OMR_JITBUILDERRECORDER_BINARYFILE_INCL
+#define OMR_JITBUILDERRECORDER_BINARYFILE_INCL
+
+#include "ilgen/JitBuilderRecorderBinaryBuffer.hpp"
+#include <fstream>
+
+// namespace TR { class IlBuilderRecorder; }
+// namespace TR { class MethodBuilderRecorder; }
```
mstoodle: remove both, please

compiler/ilgen/OMRJitBuilderRecorderTextFile.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <cstring>
+
+#include "infra/Assert.hpp"
+#include "ilgen/JitBuilderRecorderTextFile.hpp"
+
+// OMR::JitBuilderRecorderTextFile::JitBuilderRecorderTextFile(const TR::MethodBuilderRecorder *mb, const char *fileName)
+OMR::JitBuilderRecorderTextFile::JitBuilderRecorderTextFile(const char *fileName)
+   : TR::JitBuilderRecorder(), _file(fileName, std::fstream::out | std::fstream::trunc)
+   {
+   start(); // initializes IDs 0 and 1 (reserved)
```
mstoodle: i would prefer this to be done by `JitBuilderRecorder`'s constructor

compiler/ilgen/OMRJitBuilderRecorderBinaryBuffer.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <cstring>
+
+#include "ilgen/JitBuilderRecorderBinaryBuffer.hpp"
+#include "infra/Assert.hpp"
+
+// OMR::JitBuilderRecorderBinaryBuffer::JitBuilderRecorderBinaryBuffer(const TR::MethodBuilderRecorder *mb)
+OMR::JitBuilderRecorderBinaryBuffer::JitBuilderRecorderBinaryBuffer()
+   : TR::JitBuilderRecorder(), _buf()
+   {
+   start();
```
mstoodle: please move down to the base class so not needed here?

compiler/ilgen/OMRJitBuilderRecorderBinaryBuffer.cpp
```diff
@@ -0,0 +1,179 @@
...
+   }
+
+void
+OMR::JitBuilderRecorderBinaryBuffer::Close()
+   {
+   end();
+   EndStatement();
```
mstoodle: I also think this should be moved down to the base class

compiler/ilgen/OMRJitBuilderRecorderBinaryFile.cpp
```diff
@@ -0,0 +1,43 @@
...
+OMR::JitBuilderRecorderBinaryFile::JitBuilderRecorderBinaryFile(const char *fileName)
+   : TR::JitBuilderRecorderBinaryBuffer(), _file(fileName, std::fstream::out | std::fstream::app)
+   {
+   }
+
+void
+OMR::JitBuilderRecorderBinaryFile::Close()
+   {
+   end();
+   EndStatement();
```
mstoodle: Call the superclass' implementation to get `end(); EndStatement();` then do the write and close the file here.

compiler/ilgen/OMRJitBuilderRecorderTextFile.cpp
```diff
@@ -0,0 +1,130 @@
...
+   }
+
+void
+OMR::JitBuilderRecorderTextFile::Close()
+   {
+   end();
```
mstoodle: same here: please call `JitBuilderRecoder::Close()` to accomplish `end(); EndStatement();` and then close the file.

compiler/ilgen/StatementNames.hpp
```diff
@@ -0,0 +1,152 @@
...
+#define OMR_STATEMENTNAMES_INCL
+
+#include <cstdint>
+
+namespace OMR
+{
+
+namespace StatementName {
+
+static const int16_t VERSION_MAJOR              = 0;
```
mstoodle: these values should be pulled from the jitbuilder api definition rather than being defined here, somehow. @Leonardo2718 what's the best way to do that?

jitbuilder/build/rules/common.mk
```diff
@@ -159,6 +159,30 @@ $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/OMRVirtualMachineRegisterInStruct.
 $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineRegisterInStruct.hpp: $(FIXED_SRCBASE)/$(JIT_OMR_DIRTY_DIR)/ilgen/VirtualMachineRegisterInStruct.hpp $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen
 	cp $< $@ || cp $< $@
 
+$(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen/OMRJitBuilderRecorder.hpp: $(FIXED_SRCBASE)/$(JIT_OMR_DIRTY_DIR)/ilgen/OMRJitBuilderRecorder.hpp $(RELEASE_INCLUDE)/$(JIT_OMR_DIRTY_DIR)/ilgen
```
mstoodle: I didn't think we were doing this kind of thing anymore. but if we still are, then please remove the `*Recorder` versions of these files from the list

.gitignore
bragaigor: Those were the files generated by the recorder: ilt -> intermediate language text. I'll remove it.

fvtest/compilertest/build/files/common.mk
```diff
@@ -255,6 +255,10 @@ JIT_PRODUCT_SOURCE_FILES+=\
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRIlType.cpp \
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRIlValue.cpp \
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRMethodBuilder.cpp \
+    $(JIT_OMR_DIRTY_DIR)/ilgen/OMRJitBuilderRecorder.cpp \
```
mstoodle: nit: not alphabetical

jitbuilder/build/files/common.mk
```diff
@@ -224,6 +224,10 @@ JIT_PRODUCT_BACKEND_SOURCES+=\
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRIlType.cpp \
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRIlValue.cpp \
     $(JIT_OMR_DIRTY_DIR)/ilgen/OMRMethodBuilder.cpp \
+    $(JIT_OMR_DIRTY_DIR)/ilgen/OMRJitBuilderRecorder.cpp \
```
mstoodle: nit: not alphabetical

compiler/ilgen/StatementNames.hpp
Leonardo2718: Well, I'm not sure about the "best way" :stuck_out_tongue: .

It seems to me like whole file can/should be generated from the JitBuilder API description.

A few lines of Python should do the trick; something like (not thoroughly test, ideally with less hardcoded strings and a better design):

```python
import genutils

if __name__ == "__main__":
    api_file = "jitbuilder.api.json"
    dest_file = "StatementNames.hpp"

    with open(api_file) as api_src:
        api_description = genutils.APIDescription.load_json_file(api_src)

    with open(dest_file, "w") as writer:
        writer.write("#ifndef OMR_STATEMENTNAMES_INCL\n")
        writer.write("#define OMR_STATEMENTNAMES_INCL\n")

        writer.write("#include <cstdint>\n")

        writer.write("namespace OMR {\n")
        writer.write("namespace StatementNames {\n")

        version = api_description.description["version"] # need to add methods for this to APIDescription
        writer.write("static const int16_t VERSION_MAJOR = {};\n".format(version["major"]))
        writer.write("static const int16_t VERSION_MINOR = {};\n".format(version["minor"]))
        writer.write("static const int16_t VERSION_PATCH = {};\n".format(version["patch"]))

        writer.write('static const char * const RECORDER_SIGNATURE = "JBIL";\n')
        writer.write('static const char * const JBIL_COMPLETE      = "Done";\n')

        for service in api_description.get_class_by_name("MethodBuilder").services():
            name = service.name()
            writer.write('static const char * const STATEMENT_{} = "{}";\n'.format(name.upper(), name))
        for service in api_description.get_class_by_name("IlBuilder").services():
            name = service.name()
            writer.write('static const char * const STATEMENT_{} = "{}";\n'.format(name.upper(), name))

        writer.write('static const char * const STATEMENT_ID16BIT = "ID16BIT";\n')
        writer.write('static const char * const STATEMENT_ID32BIT = "ID32BIT";\n')

        writer.write("} // namespace StatementNames\n")
        writer.write("} // namespace OMR\n")

        writer.write("#endif // OMR_STATEMENTNAMES_INCL\n")
```

Then, it's just a matter of adding a few lines to the build files to run this code. :slightly_smiling_face:

mstoodle:
for the record, when I call something a "nit" I don't *need* it to be fixed, but always welcome it for "bonus points" :)

compiler/compile/OMRCompilation.cpp
```diff
@@ -371,6 +371,18 @@ OMR::Compilation::Compilation(
          );
    _isServerInlining = !options.getOption(TR_NoOptServer);
 
+   // TR_DisableInternalPointers must be set before allocateCodeGenerator(self()) is called because
```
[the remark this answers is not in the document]
bragaigor: It looks like this file is exactly the same as the one in master. Am I missing something?

compiler/ilgen/OMRJitBuilderRecorder.cpp
```diff
@@ -0,0 +1,158 @@
...
+   Builder(0);
+   Statement(s);
+   String(s);
+   EndStatement();
+   }
+
+void
+OMR::JitBuilderRecorder::end()
+   {
+   ID(1);
```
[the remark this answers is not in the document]
bragaigor: I was thinking something along the lines "// reserved ID to indicate end of file"?

compiler/compile/OMRCompilation.cpp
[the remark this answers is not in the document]
mstoodle: I didn't expect to see a change shown in this PR if the files are identical. The changes looked very similar but in different places and using slightly different ways to express the condition.

compiler/ilgen/StatementNames.hpp
mstoodle: yeah, that's a good point about the services being autogenerated. I do like that better. Thanks for the suggestion!

mstoodle: Maybe we leave the improvement of this part of the infrastructure to a second PR, though. For now, @bragaigor, let's just leave the major.minor.revision levels hard-coded and we'll fix it in a subsequent PR.

mstoodle:
Please also address @fjeremic's comments on the commit title and the commit message is no longer appropriate given the changes that have been made.

compiler/compile/OMRCompilation.cpp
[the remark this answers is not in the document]
bragaigor: Oh I see. I'll try to remove those differences!

mstoodle:
The only thing is that github is still showing a change in the "opt server" heuristic code that shouldn't be in this PR, but maybe that's just a hiccup?

Please squash the commits now and we'll run the final round of tests to make sure you didn't introduce any failures. Hopefully that will iron out this unexpected change as well, unless you know what's causing it.

mstoodle:
genie-omr build all

mstoodle:
@fjeremic do you still have concerns?

fjeremic:
> One last try to get @fjeremic to mark his approval (or to request additional changes)?

I fixed up the PR title to match that of the commit and the body as well since it had changed with additional documentation added so as to not omit anything. I made the change on @bragaigor's behalf. Thanks for the contribution!

mstoodle:
Congratulations! :)

PR #3022, opened 2018-10-02
Clean up GC test shutdown; don't run GC tests for disabled features

rwy0717:
In the GC tests, If the collector fails to initialize, then the testcase will abort before a VM thread is attached. Unfortunately, some of the GC shutdown APIs require a valid thread.  This dependency is bogus--the thread is not really needed. This PR adds some new shutdown APIs that instead take an OMR VM. With these new APIs, the tests can bring everything down safely, even when the test aborts very early.

This error condition was exposed when trying to run testing for GC features that are disabled. This is bad--obviously a scavenger test will fail when the scavenger is ifdef'd out. This PR moves the offending tests under the relevant feature ifdefs, so we only run tests when they're actually supported.

/cc @youngar @charliegracie

### DIFF
--- a/fvtest/gctest/GCConfigTest.cpp
+++ b/fvtest/gctest/GCConfigTest.cpp
@@ -40,14 +40,21 @@
 #define STRINGFY(str) DO_STRINGFY(str)
 #define DO_STRINGFY(str) #str
 
-const char *gcTests[] = {"fvtest/gctest/configuration/sample_GC_config.xml",
-                                "fvtest/gctest/configuration/test_system_gc.xml",
-                                "fvtest/gctest/configuration/gencon_GC_config.xml",
-                                "fvtest/gctest/configuration/gencon_GC_backout_config.xml",
-                                "fvtest/gctest/configuration/scavenger_GC_config.xml",
-                                "fvtest/gctest/configuration/scavenger_GC_backout_config.xml",
-                               	"fvtest/gctest/configuration/global_GC_config.xml",
-								"fvtest/gctest/configuration/optavgpause_GC_config.xml"};
+const char *gcTests[] = {"fvtest/gctest/configuration/sample_GC_config.xml"
+                        , "fvtest/gctest/configuration/test_system_gc.xml"
+                        , "fvtest/gctest/configuration/global_GC_config.xml"
+#if defined(OMR_GC_MODRON_CONCURRENT_MARK)
+                        , "fvtest/gctest/configuration/optavgpause_GC_config.xml"
+#endif
+#if defined(OMR_GC_MODRON_SCAVENGER)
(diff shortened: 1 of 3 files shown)


fvtest/gctest/GCConfigTest.cpp
```diff
@@ -44,10 +44,15 @@ const char *gcTests[] = {"fvtest/gctest/configuration/sample_GC_config.xml",
                                 "fvtest/gctest/configuration/test_system_gc.xml",
                                 "fvtest/gctest/configuration/gencon_GC_config.xml",
                                 "fvtest/gctest/configuration/gencon_GC_backout_config.xml",
```
charliegracie: Is the backout test a gencon test as well?

gc/startup/omrgcstartup.cpp
```diff
@@ -229,15 +229,18 @@ OMR_GC_InitializeDispatcherThreads(OMR_VMThread* omrVMThread)
...
 {
-	MM_GCExtensionsBase *extensions = MM_GCExtensionsBase::getExtensions(omrVMThread->_vm);
 	omr_error_t rc = OMR_ERROR_NONE;
 
-	if (NULL != extensions->dispatcher) {
-		extensions->dispatcher->shutDownThreads();
-		extensions->dispatcher->kill(MM_EnvironmentBase::getEnvironment(omrVMThread));
-		extensions->dispatcher = NULL;
-	}
+	if ((NULL != omrVMThread) && (NULL != omrVMThread->_vm)) {
```
charliegracie: No one should ever pass a NULL OMRVMThread into this API.

rwy0717: OK I wasn't sure about this one--I'll move the null check to the gc test shutdown code.

fvtest/gctest/GCConfigTest.cpp
rwy0717: Yep. I've redone these ifdefs.

charliegracie:
@genie-omr build all

gc/startup/omrgcstartup.cpp
```diff
@@ -335,3 +344,9 @@ OMR_GC_ShutdownHeap(OMR_VM *omrVM)
 	return OMR_ERROR_NONE;
 }
 
+omr_error_t
+OMR_GC_ShutdownHeapAndCollector(OMR_VM* omrVM) {
+	OMR_GC_ShutdownCollector(omrVM);
+	OMR_GC_ShutdownHeap(omrVM);
+	return OMR_ERROR_NONE;
+}
```
charliegracie: Can you add a newline at the end of the file please?

rwy0717: Fixed.

PR #3025, opened 2018-10-03
Return true for rdbar for canGCandReturn query

cathyzhyi:
GC can happen at rdbar vmhelpers.

#2812 

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -2258,17 +2258,20 @@ OMR::Node::canGCandReturn()
       return true;
       }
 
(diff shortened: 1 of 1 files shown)


andrewcraik:
The definition of canGCandReturn is really a bit language specific, but this has not been factored this way yet. Treating a rdbar as a potential GC point won't be wrong, it just may be conservative. I'm fine with this change in this light. @vijaysun-omr could you take a look please?

andrewcraik:
This was subsumed by #3008

vijaysun-omr:
Can an issue be created to refactor the code such that the query is asked of the language "runtime" in some sense ?

Issue #2948, opened 2018-09-10
Extending the current wrtbar opcode to xwrtbar/xwrtbari


cathyzhyi:
Write barrier opcodes are needed to represent a store as well as the extra side effects like notify the VM of the write for debugging purpose. There are current already `wrtbar` and `wrtbari` for address type direct and indirect stores and we are extending the opcodes and the syntax to other types.

Indirect write barrier looks like the following:

````
xwrtbari <#xxx A.instanceB>
  baseObject
  valueToStore
  extra Info needed by the extra side affect
````

Direct write barrier looks like the following:
````
xwrtbar <#xxx A.staticB>
  valueToStore
  extra Info needed by the extra side affect
````

see PR #3027 for actual implementation

cathyzhyi:
related work has been completed

PR #3027, opened 2018-10-04
Add wrtbar/wrtbari opcodes

cathyzhyi:
The xwrtbar/xwrtbari opcodes are added to represent both the store
as well as its side effects like checking object address against
tenure address range for GC, notifying the VM of the store or
other similar activities.

This PR adds the most basic supports for write barrier opcodes. The
exist wrtbar/wrtbari are change to awrtbar/awrtbari for consistency.
By default write barriers use store handlers and evaluators.

Issue: #2948


cathyzhyi:
NOTE this PR is dependent on sub projects changes.

compiler/optimizer/OMRSimplifierTableEnum.hpp
```diff
@@ -63,17 +63,29 @@
    dftSimplifier,           // TR::fstore
    dftSimplifier,           // TR::dstore
    astoreSimplifier,        // TR::astore
-   dftSimplifier,           // TR::wrtbar
    dftSimplifier,           // TR::bstore
    dftSimplifier,           // TR::sstore
+   directStoreSimplifier,   // TR::iwrtbar
```
andrewcraik: why is this one different?

andrewcraik:
Since some of the code generators cannot handle the new bytecode - should there be a codegen support query added with this so that people can know if the platform supports this expanded family of bytecodes?

andrewcraik:
Could you reference the design issue please?

andrewcraik:
Overall I'm find with the new opcodes - the additions to the optimizer seem to be the minimum to let the new opcodes exist. Are there any tests for the new op code?

compiler/optimizer/OMRSimplifierTableEnum.hpp
cathyzhyi: The wrtbars are using the same simplifiers as the corresponding stores except the awrtbar which used dftSimplifier in the old implemenation already so I kept the old implmentation.

cathyzhyi:
Created issue #3028 for adding proper testing for the wrbar opcodes.

cathyzhyi:
#2948 is the original issue for proposing wrtbar opcodes which is referenced from the sub project issue implementing field watch.

cathyzhyi:
> Since some of the code generators cannot handle the new bytecode - should there be a codegen support query added with this so that people can know if the platform supports this expanded family of bytecodes?

There is a general query for this purpose https://github.com/eclipse/omr/blob/6864b36de56b5f3f2a31099be1d70670f3daeaaa/compiler/codegen/OMRCodeGenerator.cpp#L2960

vijaysun-omr:
Do we need to check for the appropriate wrtbar variant opcode wherever we check for the corresponding store opcodes today in the optimizer (say) ? e.g. if there is some transformation that checks for an bstore opcode, are you aiming to have it also engage if we see a bwrtbar ?

cathyzhyi:
> Do we need to check for the appropriate wrtbar variant opcode wherever we check for the corresponding store opcodes today in the optimizer (say) ? e.g. if there is some transformation that checks for an bstore opcode, are you aiming to have it also engage if we see a bwrtbar ?

yes, technically we would like all the opts applying to stores apply to wrtbars as well but we would like to postpone catching all the cases until performance matters for wrtbar. A quick scan of the code doesn't show a lot of such cases probably because address type wrtbar exist before and most code is aware of the variance already.

vijaysun-omr:
Okay, could you please open an issue to update those other places ? It's unclear to me what all purposes a write barrier could be used for down the line (and whether performance might matter in those cases) and it's probably better to do the work to treat it the same as a store everywhere so that any future users of the write barrier family of opcodes are not surprised by the lack of optimizations. I'd like that "cleanup" to be done down the line, and hence my request to open an issue.

cathyzhyi:
> Okay, could you please open an issue to update those other places ? It's unclear to me what all purposes a write barrier could be used for down the line (and whether performance might matter in those cases) and it's probably better to do the work to treat it the same as a store everywhere so that any future users of the write barrier family of opcodes are not surprised by the lack of optimizations. I'd like that "cleanup" to be done down the line, and hence my request to open an issue.

Opened issue #3033 for further work.

andrewcraik:
@vijaysun-omr this has been updated again with the follow-on item you requested. Can you review again and approve if there is nothing else? @fjeremic has offered to help with this merge because of build coordination needs.

vijaysun-omr:
I see a few places where you have renamed the iwrtbar opcode to awrtbari but have you considered that the places that dealt with iwrtbar may now have to deal with the other write barrier opcodes as well ? i.e. if we went down to the else path where we are not an awrtbari opcode maybe the code was assuming it might be a store opcode that is not a write barrier before (since there was only one indirect write barrier opcode). However this will no longer hold true for a bwrtbari for example. I pointed out one instance of this potential problem above but I saw others when I looked at other changed places in this commit and wanted to ask if you had thought about those places already.

vijaysun-omr:
@genie-omr build all

compiler/il/OMRIL.cpp
```diff
@@ -584,7 +628,7 @@ OMR::IL::opCodeForCorrespondingIndirectStore(TR::ILOpCodes storeOpCode)
       case TR::fstorei:  return TR::floadi;
       case TR::dstorei:  return TR::dloadi;
       case TR::astorei:  return TR::aloadi;
-      case TR::wrtbari:  return TR::aloadi;
+      case TR::awrtbari:  return TR::aloadi;
```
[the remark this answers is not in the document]
cathyzhyi: I considered adding other types of write barriers but wrtbar's mapping might be different under different mode. If field watch is enabled, it should be mapped to rdbar but for other GC mode it should be mapped to a normal load.  This PR is only intended to add the opcodes and changes to make SURE other places ARE aware of wrtbars will be more broadly addressed in another PR https://github.com/eclipse/omr/pull/3008.

vijaysun-omr:
If I understood your last response correctly, you are only adding opcodes here and not actually creating IL nodes with the new opcodes in this PR; therefore you do not need to address all the correctness concerns in this PR but you intend to do so via the other PR you referenced ?

cathyzhyi:
@vijaysun-omr yes, the new opcodes cannot be created with this PR and a serious of later changes will follow to guarantee correctness.

vijaysun-omr:
Okay I am fine with these changes then. FYI @andrewcraik and @fjeremic

PR #3030, opened 2018-10-04
Add two helpers for OSR related optimizations

liqunl:
- Add a helper named osrFearPointHelper to mark an OSR fear point, a
  place that has been optimized with runtime assumptions and requires
  protection of OSR points. This helper helps to locate actual fear
  points, which as part of the program can be optimized, and allows the
  separation of OSR dependent optimization and OSR guard insertion.
  
- Add a helper named potentialOSRPointHelper to serve as a marker of OSR
  points. Currently, OSR points consist of calls, asyncchecks and
  monitor enters. As part of the program, they can be optimized and no
  longer be identified as OSR points. However, places where they reside
  may still be OSR possible. This helper works as a bookkeeping tool to
  track place that are OSR possible.
 
- Add a markdown file that covers a basic introduction to OSR.

- Fix some node APIs such that doNotProfile is set only for nodes
  created during ilgen. doNotProfile on a bytecode info is looked at by
  OSR APIs to determine if a node is created in ilgen. Only nodes
  created in ilgen will have the OSR bookkeeping  and are able to
  support OSR transition. Thus, nodes created after ILGen supposed to
  have doNotProfile set and cannot be OSR points.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -648,9 +648,15 @@ bool OMR::Compilation::isPotentialOSRPoint(TR::Node *node, TR::Node **osrPointNo
       else if (node->getOpCode().isCall())
          {
          TR::SymbolReference *callSymRef = node->getSymbolReference();
-         if (callSymRef->getReferenceNumber() >=
+         if (node->isPotentialOSRPointHelperCall())
+            {
+            potentialOSRPoint = true;
+            }
+         else if (callSymRef->getReferenceNumber() >=
              self()->getSymRefTab()->getNonhelperIndex(self()->getSymRefTab()->getLastCommonNonhelperSymbol()))
+            {
             potentialOSRPoint = (disableGuardedCallOSR == NULL);
+            }
          }
       else if (node->getOpCodeValue() == TR::monent)
          potentialOSRPoint = (disableMonentOSR == NULL);
@@ -774,6 +780,11 @@ OMR::Compilation::getOSRInductionOffset(TR::Node *node)
    if (!self()->isPotentialOSRPoint(node, &osrNode))
       {
       TR_ASSERT(0, "getOSRInductionOffset should only be called on OSR points");
+      }
+
+   if (osrNode->isPotentialOSRPointHelperCall())
+      {
+      return osrNode->getOSRInductionOffset();
       }
 
    if (osrNode->getOpCode().isCall())
--- a/compiler/il/Aliases.cpp
+++ b/compiler/il/Aliases.cpp
@@ -105,6 +105,33 @@ OMR::SymbolReference::getUseonlyAliasesBV(TR::SymbolReferenceTable * symRefTab)
       case TR::Symbol::IsMethod:
          {
          TR::MethodSymbol * methodSymbol = _symbol->castToMethodSymbol();
+
+         // Aliasing for potentialOSRPointHelper
+         // A potentialOSRPointHelper call is an exception point that may go to OSR catch block ( see
+         // Node API exceptionsRaised), the control flow constraint imposed by the exception edge will
+         // apply to all the global optimizations that may move things around. Local optimizations also
+         // ask exceptionsRaised to determine if a code motion across certain point is safe. So aliasing
+         // is not necessary. However, we'd like to add aliasing here to cause the compiler to be more
+         // conservative about reordering this helper with other operations. The aliasing can always be
+         // relaxed when necessary.
+         //
+         if (symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::potentialOSRPointHelperSymbol))
+            {
+            return &symRefTab->aliasBuilder.defaultMethodUseAliases();
+            }
+
+         // Aliasing for osrFearPointHelper
+         // Preventing the reordering of fear point helper w.r.t. OSR points and yield/invalidation points is
+         // the minimum requirement of aliasing for OSR fear point helper. These reorderings would in almost
+         // all cases be naturally disallowed simply due to the fact that the fear point is represented as a
+         // call, which even without aliasing could e.g. perform I/O. Thus the following is a highly conservative
+         // aliasing and can be relaxed later when necessary
+         //
+         if (symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::osrFearPointHelperSymbol))
+            {
+            return &symRefTab->aliasBuilder.defaultMethodUseAliases();
+            }
+
          if (!methodSymbol->isHelper())
             {
             return &symRefTab->aliasBuilder.defaultMethodUseAliases();
@@ -285,8 +312,12 @@ OMR::SymbolReference::getUseDefAliasesBV(bool isDirectCall, bool includeGCSafePo
          if (!methodSymbol->isHelper())
             return symRefTab->aliasBuilder.methodAliases(self());
 
-         if (symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::arraySetSymbol))
+         if (symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::arraySetSymbol) ||
+             symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::osrFearPointHelperSymbol) ||
+             symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::potentialOSRPointHelperSymbol))
+            {
             return &symRefTab->aliasBuilder.defaultMethodDefAliases();
+            }
 
          if (symRefTab->isNonHelper(self(), TR::SymbolReferenceTable::arrayCmpSymbol))
             return 0;
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -200,6 +200,8 @@ OMR::Node::Node(TR::Node *originatingByteCodeNode, TR::ILOpCodes op, uint16_t nu
       // estimate code size to be able to propagate frequencies
       //else
       //   TR_ASSERT(0, "no byte code info");
+
+      _byteCodeInfo.setDoNotProfile(1);
       }
       if(comp->getDebug())
         comp->getDebug()->newNode(self());
@@ -266,6 +268,16 @@ OMR::Node::Node(TR::Node * from, uint16_t numChildren)
 
    if(comp->getDebug())
       comp->getDebug()->newNode(self());
+
+   TR_IlGenerator * ilGen = comp->getCurrentIlGenerator();
+   if (ilGen)
+      {
+      _byteCodeInfo.setDoNotProfile(0);
+      }
+   else
+      {
+      _byteCodeInfo.setDoNotProfile(1);
+      }
 
    if (from->getOpCode().isBranch() || from->getOpCode().isSwitch())
       _byteCodeInfo.setDoNotProfile(1);
@@ -528,6 +540,9 @@ OMR::Node::recreateAndCopyValidPropertiesImpl(TR::Node *originalNode, TR::ILOpCo
    TR_ASSERT(originalNode != NULL, "trying to recreate node from a NULL originalNode.");
    if (originalNode->getOpCodeValue() == op)
       {
+      if (!originalNode->hasSymbolReference() || newSymRef != originalNode->getSymbolReference())
+         originalNode->_byteCodeInfo.setDoNotProfile(1);
+
       // need to at least set the new symbol reference on the node before returning
       if (newSymRef)
          originalNode->setSymbolReference(newSymRef);
@@ -575,6 +590,7 @@ OMR::Node::recreateAndCopyValidPropertiesImpl(TR::Node *originalNode, TR::ILOpCo
 
    // TODO: copyValidProperties is incomplete
    TR::Node::copyValidProperties(originalNodeCopy, node);
+   originalNode->_byteCodeInfo.setDoNotProfile(1);
 
    // add originalNodeCopy back to the node pool
    comp->getNodePool().deallocate(originalNodeCopy);
@@ -1191,6 +1207,38 @@ OMR::Node::createArraycopy(TR::Node *first, TR::Node *second, TR::Node * third, 
    return node;
    }
 
+
+TR::Node *
+OMR::Node::createPotentialOSRPointHelperCallInILGen(TR::Node* originatingByteCodeNode, int32_t osrInductionOffset)
+   {
+   TR::Compilation* comp = TR::comp();
+   // The following are assertions to prevent misuses of this helper
+   //
+   TR_ASSERT(comp->getCurrentIlGenerator(), "This API must be called during ILGen");
+   TR_ASSERT(!comp->isPeekingMethod(), "Can not generate the helper call during peeking");
+   TR_ASSERT(comp->supportsInduceOSR(), "Can not create the helper without OSR support");
+
+   TR::Node* callNode = TR::Node::createWithSymRef(originatingByteCodeNode, TR::call, 0, TR::comp()->getSymRefTab()->findOrCreatePotentialOSRPointHelperSymbolRef());
+   callNode->setOSRInductionOffset(osrInductionOffset);
+
+   // Node created outside of ilgen will have doNotProfile set, this results in OSR infrastructure believing it
+   // can not OSR at this node. Reset this flag since a potentialOSRPointHelper is a safe OSR transition point
+   //
+   callNode->getByteCodeInfo().setDoNotProfile(0);
+   return callNode;
+   }
+
+TR::Node *
+OMR::Node::createOSRFearPointHelperCall(TR::Node* originatingByteCodeNode)
+   {
+   TR::Compilation* comp = TR::comp();
+
+   TR_ASSERT(!comp->isPeekingMethod(), "Can not generate the helper call during peeking");
+   TR_ASSERT(comp->supportsInduceOSR(), "Can not create the helper without OSR support");
+
+   TR::Node* callNode = TR::Node::createWithSymRef(originatingByteCodeNode, TR::call, 0, TR::comp()->getSymRefTab()->findOrCreateOSRFearPointHelperSymbolRef());
+   return callNode;
+   }
 
 
 TR::Node *
@@ -3599,7 +3647,7 @@ OMR::Node::exceptionsRaised()
          break;
 #endif
       default:
-       if (node->getOpCode().isCall())
+       if (node->getOpCode().isCall() && !node->isOSRFearPointHelperCall())
             {
             possibleExceptions |= TR::Block::CanCatchOSR;
             if (node->getSymbolReference()->canGCandExcept()
@@ -4042,12 +4090,16 @@ void
 OMR::Node::setByteCodeInfo(const TR_ByteCodeInfo &bcInfo)
    {
    _byteCodeInfo = bcInfo;
+   if (!TR::comp()->getCurrentIlGenerator())
+      _byteCodeInfo.setDoNotProfile(1);
    }
 
 void
 OMR::Node::copyByteCodeInfo(TR::Node * from)
    {
    _byteCodeInfo = from->_byteCodeInfo;
+   if (!TR::comp()->getCurrentIlGenerator())
+      _byteCodeInfo.setDoNotProfile(1);
    }
 
 uint32_t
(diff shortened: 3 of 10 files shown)


liqunl:
@andrewcraik Can you review?

compiler/il/OMRNode.cpp
```diff
@@ -8642,3 +8642,55 @@ OMR::Node::resetFlagsForCodeMotion()
...
+          sym->isStaticField())
+         return true;
+      }
+   return false;
+   }
+
+/**
+ * \brief
+ *    Return true of the node is a call with osrFearPointSymbol
+ */
```
fjeremic: I'd like to see the API documentation move to the interface in the .hpp file so that downstream projects extending the classes know how to implement the specification without reading the OMR implementation.

compiler/optimizer/Inliner.cpp
```diff
@@ -103,6 +103,7 @@
 #include "ras/DebugCounter.hpp"
 #include "ras/LogTracer.hpp"                              // for heuristicTrace, etc
 #include "runtime/Runtime.hpp"
+#include "compile/SymbolReferenceTable.hpp"               // for osrFearPointSymbol, etc
```
fjeremic: Can we keep the alphabetical order and move this up to the other `"compile/"` includes?

liqunl: looks like it is already included, i deleted this line.

compiler/il/OMRNode.cpp
```diff
@@ -8642,3 +8642,55 @@ OMR::Node::resetFlagsForCodeMotion()
...
+bool
+OMR::Node::isOSRFearPoint()
```
andrewcraik: Since this is only checking for calls to the special OSRFearPoint helper I think this needs a better name because this sounds very generic when this is actually very specific.

compiler/il/OMRNode.cpp
```diff
@@ -8642,3 +8642,55 @@ OMR::Node::resetFlagsForCodeMotion()
...
+
+   return false;
+   }
+
+/**
+ * \brief
+ *    Return true of the node is a call with potentialOSRPointSymbol
+ */
+bool
+OMR::Node::isPotentialOSRPoint()
```
andrewcraik: Since this is only checking for calls to the special PotentialOSRPoint helper I think this needs a better name because this sounds very generic when this is actually very specific.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -588,62 +588,87 @@ OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR_ByteCo
...
-      if (remainderTree->getNode()->getOpCodeValue() != TR::BBEnd)
+      self()->genInduceOSRAfter(insertionPoint, induceBCI, branch, extendRemainder, offset, lastTreeTop);
+      return true;
+      }
+
+   return false;
+   }
+
+bool
+OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR::Node* osrNode, TR::TreeTop* branch,
```
andrewcraik: Where is this new API being used? inducing an OSR only conceptually needs and insertion point and a BCI so I'm not sure what this new API is going to get us? A more detailed discussion of this API and why it would be used would be helpful. Given an induction is something you expect to happen I'm not sure why the true false etc here is helpful...

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1512,6 +1537,31 @@ OMR::ResolvedMethodSymbol::supportsInduceOSR(TR_ByteCodeInfo &bci,
    return true;
    }
 
+bool
+OMR::ResolvedMethodSymbol::supportsInduceOSR(TR::Node* osrNode,
```
andrewcraik: why do we need this rather than isPotentialOSRPoint or isPotentialOSRPointWithSupport?

compiler/compile/OMRCompilation.cpp
```diff
@@ -702,7 +704,7 @@ bool OMR::Compilation::isPotentialOSRPointWithSupport(TR::TreeTop *tt)
          TR_ByteCodeInfo &bci = osrNode->getByteCodeInfo();
          TR::ResolvedMethodSymbol *method = bci.getCallerIndex() == -1 ?
             self()->getMethodSymbol() : self()->getInlinedResolvedMethodSymbol(bci.getCallerIndex());
-         potentialOSRPoint = method->supportsInduceOSR(bci, tt->getEnclosingBlock(), self(), false);
+         potentialOSRPoint = method->supportsInduceOSR(osrNode, tt->getEnclosingBlock(), self(), false);
```
andrewcraik: why do we care about the node? We can either OSR at a BCI or not which is what this was originally asking...

liqunl:
Made some changes according to Andrew's comments. I agree that the OSR infrastructure need not know about nodes, everything should be based on bytecodes. The changes are made on the fact that the OSR infrastructure is utilizing the `doNotProfile` flag on bytecode info. This flag basically tells us whether a bytecode info on a node is created during ilgen. Two bytecode info objects can have the same bytecode index and caller index, but they may have different flags. The `potentialOSRPointHelper` call will have `doNotProfile` flag set to false on its bytecode info such that the ability to transition at the helper call is not affected by when the node is created.

@andrewcraik Can you continue the review process?

liqunl:
WIP this PR since I need to rework on the alias set for the two helpers.

liqunl:
@jdmpapin @andrewcraik May I ask for a review?

andrewcraik:
So there should be an assert in the creation of helper call nodes that we have not yet removed the OSR infrastructure otherwise the nodes will have been created after the OSR setup has been completed - see osrInfrastructureRemoved on OMRCompilation - this will help prevent misuse.

andrewcraik:
I think this change warrants a markdown file added to the docs/compiler directory explaining what the helpers are, what their bytecode indices represent and what they do since it is not obvious for someone removed from the design of this new capability.

compiler/il/Aliases.cpp
```diff
@@ -105,6 +105,20 @@ OMR::SymbolReference::getUseonlyAliasesBV(TR::SymbolReferenceTable * symRefTab)
       case TR::Symbol::IsMethod:
          {
          TR::MethodSymbol * methodSymbol = _symbol->castToMethodSymbol();
+
+         // No aliasing is needed for potentialOSRPointHelper and osrFearPointHelper because:
+         // The two helper calls are considered exception points because the Node API exceptionsRaised
+         // returns non-zero value for any call. The control flow constraint imposed by the exception
+         // edge will apply to all the global optimizations that may move things around. Local
+         // optimizations also ask `exceptionsRaised` to determine if a movement across certain point
+         // is dangerous.
```
andrewcraik: So I buy that this would work for the fear points the way they are intended to be used - that is we create a fear point targeting a bytecode index in a given method when we believe the block already has the infrastructure to do that transition. I am not sure that this will hold for a fear point. The fear point is not a point that we explicitly kept track of during ilgen and so you are relying on the block's OSR path to keep things live for a point that was not tracked during ilgen. Now if the fear points are also potential OSR points then I can see how that would work, but there is an implicit assumption on the relationship between the fear point and the potential points.

How can we be sure the exception edge for the fear point a) will exist and b) will hold all the right things live?

andrewcraik:
There seems to be an implicit assumption that a fear point call node will be created for an existing OSR point (and it has to be for the transition to work). I would prefer some code to verify this - namely you cannot create an potential OSR point helper call and stick it in a block without the right exception setup and without the right book keeping being in place for the call. My opinion is that a preventative check like this will prevent a number of cases of potential misuse / undefined behavior later on and should make development use of the feature fail early rather than late with a more helpful error.

compiler/il/Aliases.cpp
jdmpapin: The exception control flow is relevant for the OSR points, but not (directly) for the fear points.

Because a fear point helper call must be protected, and because it represents no computation of its own, I believe the only constraints on it have to do with this protection. If it's unprotectable, it must _remain_ unprotectable. If OTOH it's protectable, the set of OSR points protecting it must be (roughly speaking) preserved. I say roughly speaking because any such OSR point that isn't a yield point (e.g. a potential OSR point helper call), if it could itself be protected by earlier OSR points, can be replaced with those. Protectable fear point helper calls are also allowed to become unprotectable, although that's undesirable because it would cause spurious compilation restarts.

I think these invariants could be maintained while allowing reordering of fear point helper calls w.r.t. anything other than OSR points and yield/invalidation points. These reorderings would in almost all cases be naturally disallowed simply due to the fact that the fear point is represented as a call, which even without aliasing could e.g. perform I/O.

That said, I think it's perfectly fine to add aliasing here to cause the compiler to be more conservative about reordering these helpers with other operations, so long as there is a comment to indicate that the aliasing is known to be more conservative than necessary. That would hedge against any holes in the account I just gave, and it can always be relaxed later if necessary.

liqunl:
@andrewcraik @jdmpapin Made some changes according to your comments.

andrewcraik:
@jdmpapin since I'm pretty busy can you make a first pass and I'll have a look after that?

liqunl:
Forgot to mention that the call node will have an iconst child which stores the bytecode offset of the induction target.

compiler/il/OMRNode.cpp
```diff
@@ -1192,6 +1208,38 @@ OMR::Node::createArraycopy(TR::Node *first, TR::Node *second, TR::Node * third,
...
+   TR_ASSERT(!comp->isPeekingMethod(), "Can not generate the helper call during peeking");
+   TR_ASSERT(comp->supportsInduceOSR(), "Can not create the helper without OSR support");
+
+   TR::Node* offsetNode = TR::Node::iconst(originatingByteCodeNode, offset);
+   TR::Node* callNode = TR::Node::createWithSymRef(originatingByteCodeNode, TR::call, 1, offsetNode, TR::comp()->getSymRefTab()->findOrCreatePotentialOSRPointHelperSymbolRef());
+
+   // Node created outside of ilgen will have doNotProfile set, this results in OSR infrastructure believing it
+   // can not OSR at this node. Reset this flag since a potentialOSRPointHelper is a safe OSR transition point
+   //
+   callNode->getByteCodeInfo().setDoNotProfile(0);
```
jdmpapin: Now that this function is restricted to ilgen, does it still need to `setDoNotProfile(0)`?

compiler/il/OMRNode.hpp
```diff
@@ -499,6 +530,22 @@ class OMR_EXTENSIBLE Node
 
    bool                   isClassUnloadingConst();
 
+   /**
+    * \brief
+    *    Return true of the node is a load of static final field
```
jdmpapin: Looks like "of" is a typo for "if", and again in the two just below

doc/compiler/osr/OSR.md
```diff
@@ -0,0 +1,45 @@
...
+# OSR transition
+
+OSR transition is the process of transfering the execution from jitted code to the interpreter. OSR transition involves leaving the jitted code, reconstructing the runtime VM state and resuming the execution where the jitted code left off in the interpreter.
+
+# OSR point
+An OSR point is a place in the jitted code where OSR transition is supported.
+
+At an OSR point, the VM state can be correctly reconstructed and is consistent with the VM state needed by the interpreter to resume the execution at a given bytecode index.
+
+To achieve that, bookkeeping is required at all possible OSR points during ILGen. Due to the cost and complexity of bookkeeping, OSR transition is limited to some well-defined points, which are are calls, asyncchecks and monitor enters, etc.
```
jdmpapin: Typo: are are

compiler/il/OMRNode.cpp
liqunl: The Node constructor sets `doNotProfile` if an `originatingByteCodeNode` is passed in.

liqunl:
@andrewcraik @jdmpapin Fixed the typo and moved the offset to the Node union. Will create an issue for the work to print the offset in the trace log.

compiler/il/OMRNode.cpp
```diff
@@ -267,6 +269,16 @@ OMR::Node::Node(TR::Node * from, uint16_t numChildren)
    if(comp->getDebug())
       comp->getDebug()->newNode(self());
 
+   TR_IlGenerator * ilGen = comp->getCurrentIlGenerator();
+   if (ilGen)
```
vijaysun-omr: What is the rationale for this change ? While I understand that you would be allowing more nodes to be profiled with this change (in particular, the ones that are created during IL gen), the intent of the original code may have been to avoid profiling any nodes that are created based on some other node, i.e. not created directly from bytecodes. I guess I'm wondering what the case is that you are going after with this change and whether this change only affects the case you are interested in.

compiler/il/OMRNode.cpp
```diff
@@ -528,6 +540,9 @@ OMR::Node::recreateAndCopyValidPropertiesImpl(TR::Node *originalNode, TR::ILOpCo
    TR_ASSERT(originalNode != NULL, "trying to recreate node from a NULL originalNode.");
    if (originalNode->getOpCodeValue() == op)
       {
+      if (!originalNode->hasSymbolReference() || newSymRef != originalNode->getSymbolReference())
```
vijaysun-omr: Again could you please elaborate on why you are making this change ?

compiler/il/OMRNode.cpp
```diff
@@ -8642,3 +8694,43 @@ OMR::Node::resetFlagsForCodeMotion()
 /**
  * Node flags functions end
  */
+
+bool
+OMR::Node::isLoadOfStaticFinalField()
+   {
+   if (self()->getOpCode().isLoadDirect() &&
```
vijaysun-omr: You may want to check that a static field is loaded via a direct load on all platforms, in particular on System Z. I believe we have an api to load statics indirectly as well (exposing the address as the child of the indirect load on platforms where we want to common the address).

vijaysun-omr:
It might be helpful for those reading this PR and not as familiar with OSR if you spelled out how the functionality implemented by the two new helper calls you are introducing was implemented prior to these changes. This could also help motivate the need for these changes.

compiler/il/OMRNode.cpp
liqunl: `supportsInduceOSR` looks at `doNotProfile` flag on a bci to see if a node is created during ilgen/directly from bytecodes, to determine if OSR is possible at a given bci. The intention of this change is to ensure that the flag is correctly set, such that the `supportsInduceOSR` does not respond true to nodes that are created after ilgen through this API.

liqunl: I wonder if we can initialize `doNotProfile` in `TR_ByteCodeInfo`'s constructor to 1 such that we only need to set it to 0 when the node is created directly from bytecode.

compiler/il/OMRNode.cpp
liqunl: The rational behind this change and the change above is that `doNotProfile` is used by `supportsInduceOSR` to determine if a node is created in ILGen. Nodes created in ilgen implies that the semantics of the node is equivalent of the semantics of the bytecode. If a node is transmuted, its semantics may have changed and OSR may not be possible.

compiler/il/OMRNode.cpp
vijaysun-omr: I'm still not completely clear. I assume you are concerned about ```doNotProfile``` being set to false in cases when it ought to be set to true, judging by your comment around not wanting ```supportsInduceOSR``` to return true for nodes created after IL gen. Can you give me an example of code that creates a node after IL gen but was setting ```doNotProfile``` to false ?

vijaysun-omr: To be clear my prior comment was a response to your statement : "The intention of this change is to ensure that the flag is correctly set, such that the supportsInduceOSR does not respond true to nodes that are created after ilgen through this API."...I was asking about an example that went through the node api but was presumably creating a node after IL gen that had ```doNotProfile``` set to false.

liqunl: A few examples from existing code are
https://github.com/eclipse/omr/blob/master/compiler/optimizer/LoopVersioner.cpp#L5309
https://github.com/eclipse/omr/blob/master/compiler/optimizer/OMRValuePropagation.cpp#L6485
https://github.com/eclipse/omr/blob/master/compiler/optimizer/OMRValuePropagation.cpp#L7958
https://github.com/eclipse/omr/blob/master/compiler/optimizer/VPHandlers.cpp#L4193
https://github.com/eclipse/omr/blob/master/compiler/optimizer/LoopVersioner.cpp#L5176

They are a lot of places like above where a node is created without passing an originating bytecode node. `doNotProfile` flag on their bci is not guaranteed to be set.

compiler/il/OMRNode.cpp
liqunl: Thanks for pointing this out, we do use indirect load for a static field on Z9. I will remove this check. Just for a record, the tree format of an indirect load of static field is
```
xloadi <static field symref>
   loadaddr <class statics>
```

compiler/il/OMRNode.cpp
vijaysun-omr: I believe if you create a new node with at least one child, the bytecode used is from the first child. If a node is being created without any children, then an originating bytecode node has to be passed in explicitly (and I don't know of cases when it's not). So in a sense, whatever the node you are creating, you do have an originating bytecode node available that ought to be used to call the node api that should be setting ```doNotProfile``` to true for the node being created.

vijaysun-omr: From your list of links, I think:

https://github.com/eclipse/omr/blob/master/compiler/optimizer/OMRValuePropagation.cpp#L7958
https://github.com/eclipse/omr/blob/master/compiler/optimizer/VPHandlers.cpp#L4193
https://github.com/eclipse/omr/blob/master/compiler/optimizer/LoopVersioner.cpp#L5176

are concerning since we do see examples of a node without a child being created without an originating bytecode node being passed in. I am not sure what the bytecode info would be for such nodes (you can check in the log file for the compilation). In my opinion these cases ought to be fixed...maybe separately from this PR but it is kind of related to the work done in this PR for handling the problematic cases when originating node is not available.

liqunl: I've checked that `doNotProfile` is only set to `true` when an originating bytecode node is passed, and for branch nodes. This is why I have changes to set `doNotProfile` in Node APIs that don't take an originating bytecode node. One solution for `doNotProfile` is to initialize it to `true` in `TR_ByteCodeInfo`'s constructor and allow Node APIs to set it to false.  Only nodes created in ilgen need it to be set to `false`.

For bytecode info on a node to be created, without an originating bytecode node, I found it's the same as its first child's bytecode info (from trace logs). I haven't checked nor tested the bytecode info in cases where the first child is not available either, but I think the bytecode index can be arbitrary value because the bytecode info of the node is created using the default constructor, which doesn't initialize bytecode index nor `doNotProfile` flag.

There are too many places creating a node without passing an originating bytecode node after IL gen. Node API users are not aware of the need of originating bytecode node, there's no notes in the documentation nor assertion requiring such node. The fix you mentioned will need to touch basically every opt. It will also require knowing what the opt is doing and passing the right originating bytecode node, thus it is a large item. IMO, apart from fixes to where the Node API is used, the Node API itself should also be clear on what are needed. Maybe we should not allow all Node APIs to be usable after ILGen. In summary, I don't think it's this PR's job to fix that.

vijaysun-omr: ..."but I think the bytecode index can be arbitrary value because the bytecode info of the node is created using the default constructor"...is a somewhat worrisome comment. 

I believe we did mandate that we pass in an originating bytecode node to node creation APIs in the past but it's possible some revision of that API in the last few years removed that mandatory requirement. I say this to suggest (hope ?) that there are'nt too many cases of creating nodes without using an API requiring an originating bytecode node but I agree it's probably better to deal with that in a separate PR. I would appreciate it if you would please create an issue for that and work on it asap once you have completed your current tasks because it's a dangerous state to be in to have uninitialized bytecode info on some nodes. On occasions the owning method is looked up based on the caller index that is part of the bytecode info and to have an uninitialized index to use to do the lookup could cause failures in some cases.

vijaysun-omr:
@genie-omr build all

compiler/il/OMRNode.cpp
vijaysun-omr: I would appreciate @andrewcraik opinion on the above discussion as well.

liqunl:
@vijaysun-omr Build has passed, and I've created #3153 to address the bytecode info issue. Could you merge this one?

compiler/il/OMRNode.cpp
andrewcraik: So the node creation/recreation APIs are a mess - there are a number of different versions with a number of different conventions and they do need to be cleaned up and rationalized, but I feel this is outside of the scope of this contribution discussion since @liqunl is proposing a perfectly reasonable pair of helpers as part of the OSR infrastructure.

The cases where the bytecode info won't be initialized - the cases of `createWithSymRef` which pass a NULL originating bytecode are especially problematic because the children are not available for the bytecode info. In general, bugs around the creation of nodes with uninitialized BCI have been seen a few times in the last year and indicate the API needs tightening up. That being said I don't think making the profiling flag handling explicit like has been done is wrong - it makes it clear what value it should have. I would argue that taking BCI info from a child isn't the most clear behavior either.

The issue with the node API is a much bigger one, but it will not be a quick fix. I think this is a good candidate for an architecture meeting and some help wanted items once we figure out which APIs are dangerous and need to be removed and what conventions we will apply to the handling of bytecodeinfo on recreate etc.

vijaysun-omr: @liqunl please open an issue for fixing cases when we have uninitialized bytecode info on nodes. 

@0xdaryl please consider this for discussion at the one of the upcoming OMR architecture meetings.

liqunl: https://github.com/eclipse/omr/issues/3153 is created for further discussion on bytecode info correctness.

PR #3031, opened 2018-10-04
Fix jit crash in Symbol staticName

yanluo7:
Caller of Symbol::staticName typically assumes the underlying symbol
represents a static field. However, the condition that guards the
call currently only checks for a staic symbol, which is insufficient
as other types like Constant Dynamic. Method Handle, Method Type all
use static symbols. Use the correct check Symbol::isStaticField which
precludes the above mentioned non-static field types.


0xdaryl:
Change looks ok, but I have a couple of comments:

* your commit title misled me to what your fix actually was.  I propose a title something like "Ensure staticName() symbols are static fields"

* if the callers of `staticName` are assuming they're asking about a static field, perhaps `staticName` should be renamed to `staticFieldName` to prevent this kind of mistake from occurring in the future?  Not for this PR, but might be worth an issue to track it.

0xdaryl:
@genie-omr build all

yanluo7:
Open https://github.com/eclipse/omr/issues/3034 to track renaming `staticName` to `staticFieldName`.

0xdaryl:
@genie-omr build all

Issue #2569, opened 2018-05-22
Develop a specification for TR IL


0xdaryl:
As the intermediate representation consumed by the Eclipse OMR compiler technology (known as TR IL) becomes consumed in language runtimes beyond OpenJ9, it is important to ensure its semantics are clearly defined and well understood to prevent inconsistency and ambiguity.  This is important not only for consistency between projects consuming OMR but for different architecture backends within a particular project.  The goal of this effort is to produce stand-alone documentation (either in the GitHub repo or on the Eclipse OMR website) for TR IL as well as improve the documentation in the code.

Some ideas for content and tasks for this effort include:

* Provide a clear definition of what the TR IL is and what its components are:
   * Treetop
   * Node
   * IL opcode
   * Symbol
   * SymbolReference

* Define the data types and their semantics
   * primitive scalar
   * aggregate
   * vector

* Standardize the rules around how trees are structured
   * treetop anchor nodes
   * side effects
   * acyclic
   * type consistency

* Symbols and SymbolReferences
   * kinds and their expected use cases
   * semantics of properties (e.g., volatile, unresolved)

* IL opcodes 
   * typed vs typeless opcodes
   * define opcode properties

* IL opcode semantics
   * memory load/store
   * arithmetic
   * conversion
   * control flow
      * unconditional
      * conditional
      * tables
      * exceptions
      * method dispatch
   * objects
      * allocation
      * introspection
   * misc
	
* Aliasing

* How to extend TR IL in consuming projects (?)

* Testing
   * Creating IL test cases with Tril
   * IL validation

* Should Methods and ResolvedMethods be included (?)

0xdaryl:
@vijaysun-omr @mstoodle @andrewcraik @jdmpapin @Leonardo2718 @ymanton @fjeremic @mpirvu @gita-omr @xliang6

A discussion topic for the May 23 Compiler Architecture meeting.

mstoodle:
Unfortunately, I won't be able to attend this meeting. It looks like you've got more than enough to discuss in a single meeting, though :) .

Some quick observations:

1. I think Block needs to be included as a part of the IL, since you need to be able to talk about how our trees (DAGs) can reach backwards through an extended basic block but not outside an extended block.

[...]

PR #3035, opened 2018-10-05
WIP: Fix for incorrectly marked collected reference on X86

0dvictor:
The result register of TR::aiadd/TR::aladd was incorrectly marked as a
collected reference, when the aiadd/aladd node is not internal pointer.
As a result, an invalid GC map entry may be introduced and causes GC assertions.
The result of TR::aiadd/TR::aladd can never be a collected reference, fixing it.

### DIFF
--- a/compiler/x/codegen/BinaryCommutativeAnalyser.cpp
+++ b/compiler/x/codegen/BinaryCommutativeAnalyser.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -860,30 +860,17 @@ TR::Register *TR_X86BinaryCommutativeAnalyser::integerAddAnalyserImpl(TR::Node  
       }
    else if (getCopyRegs())
       {
-      TR::Register *tempReg;
+      TR::Register* tempReg = _cg->allocateRegister();
       if (firstRegister->containsCollectedReference() ||
           secondRegister->containsCollectedReference() ||
           firstRegister->containsInternalPointer() ||
           secondRegister->containsInternalPointer())
          {
-         if (root->isInternalPointer())
+         if (root->isInternalPointer() && root->getPinningArrayPointer())
             {
-            tempReg = _cg->allocateRegister();
-            if (root->getPinningArrayPointer())
-               {
-               tempReg->setContainsInternalPointer();
-               tempReg->setPinningArrayPointer(root->getPinningArrayPointer());
-               }
+            tempReg->setContainsInternalPointer();
+            tempReg->setPinningArrayPointer(root->getPinningArrayPointer());
             }
-         else if (comp->generateArraylets() && root->getOpCodeValue() == TR::aiadd)
-            // arraylets: aiadd is technically internal pointer into spine object, but isn't marked as internal pointer
-            tempReg = _cg->allocateRegister();
-         else
-            tempReg = _cg->allocateCollectedReferenceRegister();
-         }
-      else
-         {
-         tempReg = _cg->allocateRegister();
          }
 
       targetRegister = tempReg;
--- a/compiler/x/codegen/BinaryEvaluator.cpp
+++ b/compiler/x/codegen/BinaryEvaluator.cpp
@@ -789,23 +789,8 @@ TR::Register *OMR::X86::TreeEvaluator::integerAddEvaluator(TR::Node *node, TR::C
                (internalPointerMismatch || targetRegister->containsCollectedReference()))))
             {
             TR::Register *firstOperandReg = targetRegister;
+            targetRegister = cg->allocateRegister();
 
-            // For a non-internal pointer TR::aiadd created when merging news (for example)
-            // commoning is permissible across a GC point; however the register
-            // needs to be marked as a collected reference for GC to behave correctly;
-            // note that support for internal pointer TR::aiadd (e.g. array access) is not present
-            // still
-            //
-            if (targetRegister->containsCollectedReference() &&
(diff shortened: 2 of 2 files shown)


liqunl:
The commit message is worrying me and seems wrong. The result of aiadd and aladd can be a collected reference if they're internal pointer. There is an API to compute the collectedness of a data. It's `OMR::Node::computeIsCollectedReference`. But this API might not work after lower tree since we do a lot of a2l.

0dvictor:
Could you give some examples when the result of TR::aiadd/TR::aladd should return a collected reference?
If the address node is a collected reference, the result of [address+offset] cannot be another collected reference; if the address node is not a collected reference, how can we possibly know the result is a collected reference?

andrewcraik:
@vijaysun-omr could you review? I think this is popping up a fair number of failures in recent builds.

compiler/x/codegen/BinaryCommutativeAnalyser.cpp
```diff
@@ -860,30 +860,17 @@ TR::Register *TR_X86BinaryCommutativeAnalyser::integerAddAnalyserImpl(TR::Node
...
-               tempReg->setContainsInternalPointer();
-               tempReg->setPinningArrayPointer(root->getPinningArrayPointer());
-               }
+            tempReg->setContainsInternalPointer();
+            tempReg->setPinningArrayPointer(root->getPinningArrayPointer());
             }
-         else if (comp->generateArraylets() && root->getOpCodeValue() == TR::aiadd)
-            // arraylets: aiadd is technically internal pointer into spine object, but isn't marked as internal pointer
-            tempReg = _cg->allocateRegister();
-         else
```
vijaysun-omr: I presume this allocateCollectedReferenceRegister section of code is the one that was causing problems ? 
Could you please elaborate on the actual problem that you are fixing ?

vijaysun-omr:
It seems the title of the PR may have been changed based on @liqunl earlier comment but the PR description and the commit message are still using the words that she requested a change to. 

Can you please change the commit message and be more descriptive in terms of why this change was needed ? i.e. under what situation were we using a collected reference register for an aiadd/aladd ?

vijaysun-omr:
@genie-omr build xlinux win

liqunl:
>It seems the title of the PR may have been changed based on @liqunl earlier comment but the PR description and the commit message are still using the words that she requested a change to.

I have no problem with the wording now, but I agree that we need a more descriptive commit message.

compiler/x/codegen/BinaryCommutativeAnalyser.cpp
0dvictor: The result register (`tempReg` here) was incorrectly marked as a collected reference, when the `aiadd`/`aladd` node is not internal pointer. As a result, an invalid GC map entry may be introduced and causes GC assertions.
I cannot think of any valid reasons that this result could be collected; therefore, all branches are now to allocate a "non-collected" register instead of "collected" ones.

0dvictor:
Commit messages have been updated. Sanity tests may need be restarted.

compiler/x/codegen/BinaryCommutativeAnalyser.cpp
vijaysun-omr: It seems we would have used an (uncollected) register if the isInternalPointer flag would have been "true" on this aiadd/aladd. I expect this would be set on all aiadd/aladd in Java but I guess you saw some case where that node flag was not set (?) If so, how did such a node originate ?

0dvictor: It is from Value Propagation where a `TR::arraycopy` gets generated. The source and destination children are `TR::aladd`.

vijaysun-omr: If these aladds created by arraycopy generation in value propagation don't have the internal pointer flag set, then we should consider setting the flag and see if that fixes the failure.

0dvictor: But either way `aladd` shouldn't return a collected reference. It doesn't make sense to create a collected reference from [address+offset].

vijaysun-omr: Agreed, but it's preferable to fix a bug in common code if we can so that each platform does'nt have to make changes.

0dvictor: Good suggestion. @liqunl kindly agreed to investigate why `root->isInternalPointer()` answered no as a separate issue.

liqunl: It seems like VP arraycopy transformation creates an aiadd/aladd without setting the flag. I'm trying to find where the node is created.

liqunl:
Created https://github.com/eclipse/openj9/issues/3183 to set internal pointer node flags in VP.

andrewcraik:
@vijaysun-omr and @0dvictor  I'm not sure if this is the fix we want in general. #3046 was just merged to fix the missing flag. In general we seem to have a principle of offsets to a collected reference result in a collected reference unless the result is explicitly marked as an internal pointer not to be reported to the GC. removing the functionality to allow collected reference to be generated from collected references via explicit offsets not marked internalPointer because Java doesn't use the feature doesn't seem to be quite the right solution. Now there may be another good reason for removing it, but VP failing to set a flag that is required by current convention doesn't seem like a problem the codegen should be trying to fix as alluded to earlier.

0dvictor:
I don't this this PR is needed to fix #3040 either. This is a generic feature fix, while #3046 is the fix for root cause of #3040.

0dvictor:
@andrewcraik pointed a good point. I think our different points of views came from different understanding of the semantics of `TR::aiadd/TR:aladd`. I suggest to wait for #2569 before making decision whether this changeset should be merged.

Marking this PR as WIP for now.

0dvictor:
Not yet. I don't think we can progress this further until the semantics of `TR::iadd/TR::ladd` is clear.

PR #3038, opened 2018-10-05
Fix for compilation error under certain compilers

0dvictor:
Type information getChild(0) doesn't seem to be propagated properly
when compiling with xlC under AIX. Making the the type of child
explicit seems succificient to get compiled.


vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks safe to me. Will wait for the checks to pass before considering merge.

PR #3042, opened 2018-10-09
CMake: Add support for pasm2asm on NASM source files

dnakamura:

### DIFF
--- a/cmake/modules/OmrCompilerSupport.cmake
+++ b/cmake/modules/OmrCompilerSupport.cmake
@@ -95,10 +95,10 @@ function(pasm2asm_files out_var compiler)
 	set(result "")
 	foreach(in_f ${ARGN})
 		get_filename_component(extension ${in_f} EXT)
-		if (extension STREQUAL ".pasm") # Requires preprocessing
+		if (extension MATCHES "\\.pn?asm") # Requires preprocessing
 			get_filename_component(absolute_in_f ${in_f} ABSOLUTE)
 
-			string(REGEX REPLACE ".pasm" ".asm" out_f ${in_f})
+			string(REGEX REPLACE "\\.p(n?asm)" ".\\1" out_f ${in_f})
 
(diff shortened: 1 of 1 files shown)


dnakamura:
`.pnasm` files do exist in OpenJ9. (https://github.com/eclipse/openj9/blob/master/runtime/compiler/x/runtime/X86PicBuilder.pnasm)

It looks like its just so that it can consume `j9cfg.h`

0xdaryl:
I believe there is only one #define that's needed from j9cfg.h.  @nbhuiyan planned on modifying the jilconsts generator to add the needed definition in there to avoid this dependency.  That work is planned for by end of October.  The initial version, however, still relies on the C preprocessor unfortunately.

charliegracie:
From a quick look at `jilconsts.h` and `jilconsts.c` it would be a smaller change to add this one define to be written out in `writeConstants` than it would be to make this change.  This change is small but adding 3 lines to jilconsts.c would be just as easy and would happen in the project that requires the code and not pus work down onto OMR which technically does not belong in this project.

I could be missing something about jilconsts?

charliegracie:
I would personally prefer to see new work for a downstream project happen in the downstream project and not in OMR whenever possible. If it is a matter of adding 1 new constant to `jilconsts` would that actually be easier / simpler than making a change in OMR and forcing cpp to be run on the files?

0xdaryl:
@nbhuiyan : you've studied that code more than I have.  Is the change that straightforward?  If so, please make it in OpenJ9 so we can avoid this OMR dependence.

nbhuiyan:
@0xdaryl @charliegracie @dnakamura eclipse/openj9#3243 would remove the need for the C preprocessor for NASM files.

0xdaryl:
Thanks.  This PR should be closed then.

charliegracie:
Closing as per the comment from @nbhuiyan

PR #3043, opened 2018-10-09
Differentiate memory settings for CS HW support

dmitripivkine:
Concurrent Scavengers runs without HW support (using Software Barriers)
don't require strict heap geometry alignments used for Guarded Storage.
Add new flag to extensions concurrentScavengerHWSupported and
isConcurrentScavengerHWSupported() helper to use in places where Guarded
Storage compatible settings are necessary.
Temporary make isConcurrentScavengerHWSupported() return value of
concurrentScavenger flag for compatibility purposes.


dmitripivkine:
@amicic FYI

charliegracie:
@genie-omr build all

charliegracie:
If there is a +1 from @amicic and the testing has passed any committer can merge.

amicic:
looking good

PR #3045, opened 2018-10-10
Compressedrefs - Map memory below 4GB - OSX

babsingh:
**1) Allow memory to be mapped below 4GB on OSX**

  On OSX 64-bit, linker sets page zero size to 4GB. This prevents memory
  to be mapped below 4GB.

  On OSX 32-bit, page zero size is set to 4KB by default.

  This change will set page zero size to 4KB on 64-bit architectures. This
  will allow memory to be mapped below 4GB.

**2) Enable support for compressedrefs feature in OMR**

  For compressedrefs, memory needs to be mapped below 4GB.

  Currently, code for allocating/mapping memory below 4GB is disabled on
  OSX.

  Specifying -pagezero_size <SIZE> while creating main executables will
  allow us to map memory below 4GB.

  Thus, re-enabling code for allocating/mapping memory below 4GB:
  omrmem_allocate_memory32 and supporting test code.

### DIFF
--- a/cmake/modules/platform/os/osx.cmake
+++ b/cmake/modules/platform/os/osx.cmake
@@ -28,3 +28,10 @@ list(APPEND OMR_PLATFORM_DEFINITIONS
 # of the OMR code should be heavily reduced. In the mean time, we keep
 # the distinction
 list(APPEND TR_COMPILE_DEFINITIONS -DSUPPORTS_THREAD_LOCAL -DOSX)
+
+if(OMR_ENV_DATA64)
+	# Set page zero size to 4KB for mapping memory below 4GB.
+	list(APPEND OMR_PLATFORM_EXE_LINKER_OPTIONS
+		-pagezero_size 0x1000
+	)
+endif()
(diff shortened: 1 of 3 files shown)


babsingh:
fyi - @DanHeidinga @charliegracie @youngar @dnakamura

cmake/modules/platform/os/osx.cmake
```diff
@@ -28,3 +28,10 @@ list(APPEND OMR_PLATFORM_DEFINITIONS
 # of the OMR code should be heavily reduced. In the mean time, we keep
 # the distinction
 list(APPEND TR_COMPILE_DEFINITIONS -DSUPPORTS_THREAD_LOCAL -DOSX)
+
+if(OMR_ENV_DATA64)
```
charliegracie: this should be if 64 bit and compressed references I think

charliegracie:
The commit message for `Enabled compressedrefs feature in OMR` is a little misleading. That commit is not actually enabling compressed references in OMR or OMR testing but rather allowing the feature to be enabled.

charliegracie:
@genie-omr build all

DanHeidinga:
related to eclipse/openj9#36

PR #3046, opened 2018-10-10
Set aladd internal pointer flag for VP arraycopy

### LINKED ISSUE #3040: Correctly set node flags for internal pointer in VP
Internal pointers - aiadd/aladd nodes pointing to the middle of an object should have `isInternalPointer` flag set. Not setting it and its pinning array will result in the evaluator go to a wrong path.

jdmpapin:
These nodes are internal pointers and flagging them as such prevents
code generators from including their registers in GC maps.

generateArrayletAddressTree() is left unchanged because it's unclear
whether the leaf element address should be considered an internal
pointer, but this function is currently unused anyhow: it is only called
when generateArraylets(), which with hybrid arraylets is always false.

Fixes #3040

### DIFF
--- a/compiler/optimizer/ValuePropagationCommon.cpp
+++ b/compiler/optimizer/ValuePropagationCommon.cpp
@@ -2015,6 +2015,7 @@ TR::Node *generateArrayAddressTree(
       }
 
    array = TR::Node::create(is64BitTarget ? TR::aladd : TR::aiadd, 2, objNode, array);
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build all

vijaysun-omr:
This is a safe one line change to mark nodes created by VP with the same flag as the IL gen sets and the codegen expects. Checks have passed. Merging.

PR #3052, opened 2018-10-12
WIP Map OS name "Darwin" to "Mac OS X"

pdbain-ibm:
This is required for Java runtime compatibility.

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -500,8 +500,6 @@ omrsysinfo_get_OS_type(struct OMRPortLibrary *portLibrary)
 #else
 	if (NULL == PPG_si_osType) {
 		int rc;
-		int len;
-		char *buffer;
 		struct utsname sysinfo;
 
 #ifdef J9ZOS390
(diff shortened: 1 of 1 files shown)


port/unix/omrsysinfo.c
```diff
@@ -510,14 +508,19 @@ omrsysinfo_get_OS_type(struct OMRPortLibrary *portLibrary)
 		rc = uname(&sysinfo);
 #endif
 		if (rc >= 0) {
-			len = strlen(sysinfo.sysname) + 1;
+			size_t len = 0;
+			char *buffer = NULL;
+			char *sysName = sysinfo.sysname;
+			if (0 == strcmp(sysName, "Darwin")) { /* Translate this to "OS X" */
```
DanHeidinga: This should probably be ifdef'd so it only runs on osx.

fjeremic:
Why don't we just rename to "Mac OS X" all together rather than map it? @mstoodle @0xdaryl @charliegracie thoughts?

port/unix/omrsysinfo.c
pdbain-ibm: I could hard-code `sysName` and skip the call to `uname()`

charliegracie: I think we should create a `#elif OSX` (using the proper define) and inside the code do something similar to this:
```
return "Mac OS X";
```

In the future if we have to change this based on the version of OS X or Mac OS running we already have a place to do it.

pdbain-ibm:
Per Charlie's suggestion 
https://github.com/eclipse/omr/pull/3052#discussion_r224817638, closing this and opening a (much simpler) pull request https://github.com/eclipse/omr/pull/3053.

pdbain-ibm:
Further to 
https://github.com/eclipse/omr/pull/3052#issuecomment-429353059: why not do the same for z/OS at the same time?

PR #3054, opened 2018-10-12
Report if CS is H/W assisted

amicic:
Verbose GC change, 3 cases
1) CS enabled, running interpreted or Jitted with S/W based Read Barrier:
```
  <attribute name="concurrentScavenger" value="enabled" />
```
2) CS enabled, running Jitted with H/W based Read Barrier:
```
  <attribute name="concurrentScavenger" value="enabled, H/W assisted" />
```
3) CS disabled, the line is not reported

### DIFF
--- a/gc/include/omrmm.hdf
+++ b/gc/include/omrmm.hdf
@@ -204,7 +204,7 @@ typedef uintptr_t (*condYieldFromGCFunctionPtr) (OMR_VMThread *omrVMThread, uint
 		<data type="struct OMR_VMThread*" name="currentThread" description="current thread" />
 		<data type="uint64_t" name="timestamp" description="time of event" />
 		<data type="const char*" name="gcPolicy" description="-Xgcpolicy value" />
-		<data type="uintptr_t" name="concurrentScavenger" description="Scavenger running Concurrently flag" />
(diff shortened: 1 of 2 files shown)


fjeremic:
Could I please get a review from @dmitripivkine?

amicic:
This is a minor downstream dependency with  https://github.com/eclipse/openj9/pull/3251.
No compile dependancy, just runtime.
This change requires for downstream project to supply info if JIT lib is loaded, For OpenJ9, without the downstream change incorrect info will be supplied, so the verbose GC line may be incorrect, but there will no effect on program or JVM execution.

dmitripivkine:
LGTM

fjeremic:
@genie-omr build all

fjeremic:
Tested on z14 and output is as expected.

joransiu:
I have a concern for users running ConcurrentScavenger on platforms that support hardware assist.  Despite the intention to use hardware support, they may have misconfigured one of the underlying OS, hypervisor, h/w configurations.   By seeing `<attribute name="concurrentScavenger" value="enabled" />`, they might be misled to think that they are getting the full benefit of CS w/ hardware assist, without knowing better.

PR #3055, opened 2018-10-12
Workaround for unsupported GCC features on OSX

babsingh:
When using GCC 4.2.1 on OSX 10.10 Yosemite, the following errors are
seen:
```
In file included from /usr/include/os/object.h:27,
                 from /usr/include/dispatch/dispatch.h:48,
                 from .å./omr/include_core/./unix/thrdsup.h:74,
                 from ../omr/include_core/thrdsup.h:35,
                 from ../omr/include_core/thrtypes.h:32,
                 ...
/usr/include/os/base.h:113:20: error: missing binary operator before
token "("
/usr/include/os/base.h:119:54: error: missing binary operator before
token "("
In file included from /usr/include/dispatch/dispatch.h:51,
                 from ../omr/include_core/./unix/thrdsup.h:74,
                 from ../omr/include_core/thrdsup.h:35,
                 from ../omr/include_core/thrtypes.h:32,
                 ...
/usr/include/dispatch/object.h:325:20: error: missing binary operator
before token "("
/usr/include/dispatch/object.h:363:20: error: missing binary operator
before token "("
/usr/include/dispatch/object.h:391:20: error: missing binary operator
before token "("
/usr/include/dispatch/object.h:423:20: error: missing binary operator
before token "("
```
The above error happens because __has_extension and __has_feature macros
are undefined when using GCC 4.2.1 on OSX 10.10 Yosemite. The above
error also happens with other version of GCC.

Clang supports the missing macros so the above errors are not seen with
Clang.

For supporting OpenJ9 JDK8 on OSX, OMR needs to be built with GCC 4.2.1
on OSX 10.10 Yosemite.

Adding stubs for the missing macros before including dispatch.h resolves
the above errors.


babsingh:
fyi - @DanHeidinga @charliegracie

babsingh:
Related to https://github.com/eclipse/openj9/issues/36.

include_core/unix/thrdsup.h
```diff
@@ -71,6 +71,15 @@ typedef WRAPPER_TYPE (*WRAPPER_FUNC)(WRAPPER_ARG);
 #include <semaphore.h>
 typedef sem_t OSSEMAPHORE;
 #elif defined(OSX) /* defined(LINUX) || defined(AIXPPC) */
+#if defined(__GNUC__)
+#ifndef __has_extension
```
DanHeidinga: @babsingh Can you add a comment to this code that describes why this is required?

charliegracie:
@genie-omr build all

0xdaryl:
Is GCC 4.2.1 the GCC version that's required to build on macOS?  The JIT/compiler won't build with anything earlier than GCC 4.3.x (I don't quite recall what 'x' is), primarily due to dependencies on the allocator it uses for STL containers.  It is a non-trivial exercise to change, but please clarify the compiler-level requirement.

include_core/unix/thrdsup.h
babsingh: added a comment

babsingh:
@0xdaryl I am using the following instructions to build OpenJ9 JDK8 on OSX: https://github.com/manasthakur/jdk-tips/wiki/Building-OpenJDK-8-on-Mac-OS-X-Yosemite. Xcode4 is needed to built OpenJDK8. Xcode4 comes with `gcc version 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)`.

DanHeidinga:
@charliegracie can you rekick the build?  A new commit with an updated comment was pushed...

charliegracie:
I am curious if the OMR compiler component compiles at all with this compiler? @babsingh do you have any insight into that since you appear to be doing builds?

babsingh:
based upon the discussion with @0xdaryl, OpenJ9 JIT needs gcc-4.4.x+. So, we will need to upgrade gcc version using brew to support the JIT.

@0xdaryl does the gcc-4.4.x+ requirement also apply to the OMR compiler component?

The issue reported in this pull request was also seen with gcc-4.9 on Yosemite: https://stackoverflow.com/questions/26527077/compiling-with-accelerate-framework-on-osx-yosemite.

DanHeidinga:
@charliegracie @0xdaryl The builds have passed and @babsingh has clarified that this is required even on gcc 4.9.  Is there anything else needed for this to be merged?

PR #3058, opened 2018-10-12
Correct the comment of indirectCallEvaluator on X86

0dvictor:
Comment above indirectCallEvaluator was incorrect, fixing.


0xdaryl:
@genie-omr build xlinux,win

PR #3059, opened 2018-10-13
RecognizedCallTransformer to provide a default action

0dvictor:
RecognizedCallTransformer has a perfect use case of #2934: to transform
a supported intrinsic method to its corresponding IL op.

### DIFF
--- a/compiler/optimizer/OMRRecognizedCallTransformer.cpp
+++ b/compiler/optimizer/OMRRecognizedCallTransformer.cpp
@@ -27,6 +27,8 @@
 #include "il/TreeTop_inlines.hpp"
 #include "infra/Checklist.hpp"       // for NodeChecklist, etc
 #include "codegen/CodeGenerator.hpp" // for CodeGenerator
+#include "codegen/CodeGenerator_inlines.hpp"
+#include "optimizer/TransformUtil.hpp"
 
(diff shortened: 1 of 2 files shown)


compiler/optimizer/OMRRecognizedCallTransformer.cpp
```diff
@@ -55,6 +57,20 @@ int32_t OMR::RecognizedCallTransformer::perform()
    return 0;
    }
 
+bool OMR::RecognizedCallTransformer::isInlineable(TR::TreeTop* treetop)
+   {
```
0xdaryl: Who is the expected consumer of this API?  The optimizer inliner or a code generator?

0dvictor: Its only consumer is `OMR::RecognizedCallTransformer::perform()`. The API is for downstream projects to provide information about whether a call node should be transformed. `OMR::RecognizedCallTransformer::perform()` applies transformation according to the return of this API.

0dvictor: This is an ILGen Opt.

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build xlinux

0xdaryl:
@genie-omr build xlinux

The xLinux builds appear to have stalled for connectivity issues.  Re-starting.

PR #3060, opened 2018-10-13
Remove J9 project specific arraycopy code on X86

0dvictor:
J9 project specific code related to ArrayCopy has been deprecated,
removing them.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -2280,39 +2280,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
 
 TR::Register *OMR::X86::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   if (node->isReferenceArrayCopy() && !node->isNoArrayStoreCheckArrayCopy())
-      {
-      return TR::TreeEvaluator::VMarrayStoreCheckArrayCopyEvaluator(node, cg);
-      }
-   // There are two cases.
-   // In the first case we know that a simple memmove or memcpy operation can
-   // be done. For this case there are 3 children:
-   //    1) The byte source pointer
-   //    2) The byte destination pointer
-   //    3) The byte count
+   // There are two variants of TR::arraycopy: one has 5 children, the other has 3 children. Details can be found from
+   // compiler/il/ILOpCodeProperties.hpp
    //
-   // In the second case we must generate run-time tests to see if a simple
-   // byte copy can be done or if an element-by-element copy is needed.
-   // For this case there are 5 children:
-   //    1) The original source object reference
-   //    2) The original destination object reference
-   //    3) The byte source pointer
-   //    4) The byte destination pointer
-   //    5) The byte count
+   // In most, if not all, cases, the 5-child variant requires language specific semantics, which OMR is not aware of. The fact 
+   // that a 5-child arraycopy is generated indicates at least one of the first two children must be needed when performing the
+   // copy; otherwise a 3-child arraycopy should be generated instead. Interpreting the meanings of the first two children 
+   // definitely requires language specific semantics. For example, the first two children may be for dealing with an arraycopy 
+   // where the Garbage Collector may need to be notified about the copy or something to that affect.
    //
+   // Therefore, this OMR evaluator only handles the 3-child variant, is an operation equivalent to C++'s std::memmove().
+   // Should a downstream project need the 5-child variant evaluator, it needs to override this evaluator in its own TreeEvaluator,
+   // and delegates the 3-child variant back to this evaluator by calling OMR::TreeEvaluatorConnector::arraycopyEvaluator.
+
+   TR_ASSERT_FATAL(node->getNumChildren() == 3, "This evaluator is for the 3-child variant of arraycopy.");
 
    // ALL nodes need be evaluated to comply argument evaluation order;
    // since size node is the last node, its evaluation can be delayed for further optimization
-   TR::Node* sizeNode = node->getLastChild();
-   for (int32_t i = 0; i < node->getNumChildren()-1; i++)
-      {
-      cg->evaluate(node->getChild(i));
-      }
+   TR::Node* srcNode  = node->getChild(0); // the address of memory to copy from
+   TR::Node* dstNode  = node->getChild(1); // the address of memory to copy to
+   TR::Node* sizeNode = node->getChild(2); // the size of memory to copy, in bytes
 
-   TR::Register* srcReg = cg->allocateRegister();
-   TR::Register* dstReg = cg->allocateRegister();
-   generateRegRegInstruction(MOVRegReg(), node, srcReg, node->getChild(node->getNumChildren() - 3)->getRegister(), cg);
-   generateRegRegInstruction(MOVRegReg(), node, dstReg, node->getChild(node->getNumChildren() - 2)->getRegister(), cg);
+   TR::Register* srcReg = cg->gprClobberEvaluate(srcNode, MOVRegReg());
+   TR::Register* dstReg = cg->gprClobberEvaluate(dstNode, MOVRegReg());
 
    TR::DataType dt = node->getArrayCopyElementType();
    uint32_t elementSize = 1;
@@ -2358,8 +2348,9 @@ TR::Register *OMR::X86::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::Co
       }
    else
       {
-      TR::Register* sizeReg = TR::TreeEvaluator::intOrLongClobberEvaluate(sizeNode, TR::TreeEvaluator::getNodeIs64Bit(sizeNode, cg), cg);
-      if (TR::Compiler->target.is64Bit() && !TR::TreeEvaluator::getNodeIs64Bit(sizeNode, cg))
+      bool isSize64Bit = TR::TreeEvaluator::getNodeIs64Bit(sizeNode, cg);
+      TR::Register* sizeReg = cg->gprClobberEvaluate(sizeNode, MOVRegReg(isSize64Bit));
+      if (TR::Compiler->target.is64Bit() && !isSize64Bit)
          {
          generateRegRegInstruction(MOVZXReg8Reg4, node, sizeReg, sizeReg, cg);
          }
@@ -2381,16 +2372,8 @@ TR::Register *OMR::X86::TreeEvaluator::arraycopyEvaluator(TR::Node *node, TR::Co
 
    cg->stopUsingRegister(dstReg);
    cg->stopUsingRegister(srcReg);
-#ifdef J9_PROJECT_SPECIFIC
-   if (node->isReferenceArrayCopy())
-      {
-      TR::TreeEvaluator::generateWrtbarForArrayCopy(node, cg);
-      }
-#endif
-   for (int32_t i = 0; i < node->getNumChildren()-1; i++)
-      {
-      cg->decReferenceCount(node->getChild(i));
-      }
+   cg->decReferenceCount(dstNode);
+   cg->decReferenceCount(srcNode);
    return NULL;
    }
 
@@ -4108,12 +4091,6 @@ OMR::X86::TreeEvaluator::tabortEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    //TODO: Find a way to detect the real abort reason here
    generateImmInstruction(XABORT, node, 0x04, cg);
    return NULL;
-   }
-
-TR::Register *
-OMR::X86::TreeEvaluator::VMarrayStoreCheckArrayCopyEvaluator(TR::Node*, TR::CodeGenerator*)
-   {
-   return 0;
    }
 
 bool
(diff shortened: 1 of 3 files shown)


andrewcraik:
It worries me that the 5 child variant of arraycopy has no OMR support on x86 from what I can see - am I right that the only arraycopy evaluator is a primitive arraycopy evaluator? Yet I still see supports reference array copy being set by the OMR codegen. The need for a 5 child copy is an OMR-level concept so I'm not sure why we are removing all mention of it?

If it is a case that the 5 child copy is not supported in OMR because the semantics are too language specific at the very least the comments about the 5 child copy existing should not be expunged so that people generating these and finding no evaluator know they need to implement it. Having a hook point for them to implement may also remain a good idea - hence why there was a stub for it...

0dvictor:
> It worries me that the 5 child variant of arraycopy has no OMR support on x86 from what I can see - am I right that the only arraycopy evaluator is a primitive arraycopy evaluator? Yet I still see supports reference array copy being set by the OMR codegen. The need for a 5 child copy is an OMR-level concept so I'm not sure why we are removing all mention of it?

Good point. The 3-child variant should not mean primitive arraycopy. I've updated the comments to avoid this confusion. Now the it states there are two variants of arraycopy, but OMR can only do the 3-child variant and downstream project should take care of the 5-child variant.

> Yet I still see supports reference array copy being set by the OMR codegen.

This is a good question. In fact OMR does not fully support reference array copy. I'm not sure declaring such compatibility is proper in OMR.

> If it is a case that the 5 child copy is not supported in OMR because the semantics are too language specific at the very least the comments about the 5 child copy existing should not be expunged so that people generating these and finding no evaluator know they need to implement it.

Yes, 5-child arraycopy cannot be supported in OMR because language specific semantics is required to interpret the first two children. The fact that a 5-child arraycopy is generated indicates at least one of the first two children must needed when performing the copy; otherwise a 3-child arraycopy should be generated instead.

> Having a hook point for them to implement may also remain a good idea - hence why there was a stub for it...

IMO, the empty stub is more confusing than useful. It looks more like a uncompleted default behavior rather than an internal API that a downstream project should implement.
Anyway, the hook point is not need at all. A downstream project should override `arraycopyEvaluator` and do the language specific work. It can also choose to delegate handling of the 3-child variant to back to OMR. OpenJ9 has an example that OpenJ9's arraycopy evaluator checks if the arraycopy type and delegates back to OMR when necessary.
https://github.com/eclipse/openj9/blob/fa48180f21facc6ac1156a31d97adc5512d2c1ef/runtime/compiler/x/codegen/J9TreeEvaluator.cpp#L1513

andrewcraik:
Can we keep the comment that describes the meaning of the children - that is a very useful thing that has been purged...

andrewcraik:
I'm fine with OMR not supporting it but we probably need to move the support for reference array copy setting out of OMR for consistency. We should also then make sure that the comment in the arraycopyEvaluator tells you that you need to set the flag to indicate support for the 5-child variant when when you want it.

I think it also bears mentioning in the comment why the 5 child version needs language specific semantics - eg that it is to deal with array copies where the garbage collector may need to be told about the copy being done or something to that affect.

0dvictor:
> Can we keep the comment that describes the meaning of the children - that is a very useful thing that has been purged...

Good point. Description of the meaning of the children is very useful; however, such description already exists in "compiler/il/ILOpCodeProperties.hpp". I feel duplicating the information is redundant and  reduces maintainability. I have added comments to direct readers where to find this information.
https://github.com/eclipse/omr/blob/de7e6dc30504cb34cade6c98a7637483938ec5c8/compiler/il/ILOpCodeProperties.hpp#L9863-L9884 

> I'm fine with OMR not supporting it but we probably need to move the support for reference array copy setting out of OMR for consistency.

For sure, we should move the setting out of OMR; however, it is beyond the scope of this pull request. OMR doesn't fully support reference arraycopy even without this pull request. Therefore, relocating the settings isn't related to this pull request. I'll create a separate pull request to move it out.

> We should also then make sure that the comment in the arraycopyEvaluator tells you that you need to set the flag to indicate support for the 5-child variant when when you want it.

I'm not sure if "5-child arraycopy" is an interchangeable concept to "reference arraycopy". The 3-child variant may be used for copying `TR_Address` under certain conditions. On the other side, a downstream project may need use the 5-child variant for copying primitive arrays in theory. Therefore, I feel improper to mention "reference arraycopy" in this evaluator.

> I think it also bears mentioning in the comment why the 5 child version needs language specific semantics - eg that it is to deal with array copies where the garbage collector may need to be told about the copy being done or something to that affect.

Good suggestion, comments have been updated accordingly.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2249,39 +2249,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
...
-   //    5) The byte count
+   // In most, if not all, cases, the 5-child variant requires language specific semantics, which OMR is not aware of.
+   // The fact that a 5-child arraycopy is generated indicates at least one of the first two children must needed when
+   // performing the copy; otherwise a 3-child arraycopy should be generated instead. Interpreting the meanings of the
+   // first two children definitely requires language specific semantics. For example, the first two children may be for
+   // dealing with arraycopy where Garbage Collector may need be notified about the copy or something to that affect.
    //
+   // Therefore, this OMR evaluator only handles the 3-child variant, is an operation equivalent to C++'s std::memmove().
+   // Should a downstream project need the 5-child variant evaluator, it need override this evaluator in its own TreeEvaluator,
+   // and delegates the 3-child variant back to this evaluator by calling OMR::TreeEvaluatorConnector::arraycopyEvaluator.
```
andrewcraik: The upward delegation should be through the TR namespace for the composition model with extensible classes to work properly I think.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2249,39 +2249,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
...
-   // In the second case we must generate run-time tests to see if a simple
-   // byte copy can be done or if an element-by-element copy is needed.
-   // For this case there are 5 children:
-   //    1) The original source object reference
-   //    2) The original destination object reference
-   //    3) The byte source pointer
-   //    4) The byte destination pointer
-   //    5) The byte count
+   // In most, if not all, cases, the 5-child variant requires language specific semantics, which OMR is not aware of.
+   // The fact that a 5-child arraycopy is generated indicates at least one of the first two children must needed when
```
andrewcraik: must _be_ needed - should we mention the feature flags gating the evaluator here?

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2249,39 +2249,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
...
+   // performing the copy; otherwise a 3-child arraycopy should be generated instead. Interpreting the meanings of the
+   // first two children definitely requires language specific semantics. For example, the first two children may be for
+   // dealing with arraycopy where Garbage Collector may need be notified about the copy or something to that affect.
```
andrewcraik: with _an_ arraycopy where _the_ Garbage Collector may need _to_ be notified

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2249,39 +2249,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
...
    //
+   // Therefore, this OMR evaluator only handles the 3-child variant, is an operation equivalent to C++'s std::memmove().
+   // Should a downstream project need the 5-child variant evaluator, it need override this evaluator in its own TreeEvaluator,
```
andrewcraik: it _needs to_ override

andrewcraik:
Ok I had a few grammar edits on the comments - I see your point about not reproducing the documentation - at least mentioning the variants and that the first two children are special addresses my concern in regards to giving people that context when the first find themselves looking at this evaluator. We might want to consider a comment on the feature flags gating the generation of the versions so people know what they are and how they relate to the evaluator.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Updated.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Updated.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Updated.

0dvictor:
> Ok I had a few grammar edits on the comments - I see your point about not reproducing the documentation - at least mentioning the variants and that the first two children are special addresses my concern in regards to giving people that context when the first find themselves looking at this evaluator. 

Thank you and the comments have been updated.

> We might want to consider a comment on the feature flags gating the generation of the versions so people know what they are and how they relate to the evaluator.

Good suggestion. I added this into "ILOpCodeProperties.hpp" since I feel it makes more sense to find such information right besides its specs.

compiler/x/codegen/OMRTreeEvaluator.cpp
andrewcraik: @0dvictor please fix this comment

andrewcraik:
I think this now makes sense and is the right balance of OMR support, OMR guidance, and not having language specific implementation in OMR. @0xdaryl I'd appreciate your thoughts.

compiler/x/codegen/OMRTreeEvaluator.cpp
0dvictor: Sorry that I forgot to response this comments. I believe OMR is the correct namespace when the upward delegation needs to call this OMR implementation.
Here is two examples in OpenJ9:
https://github.com/eclipse/openj9/blob/fc85619c5e2df8bd045a6aa94d169b55461fcd15/runtime/compiler/codegen/J9CodeGenerator.cpp#L572
https://github.com/eclipse/openj9/blob/fc85619c5e2df8bd045a6aa94d169b55461fcd15/runtime/compiler/codegen/J9CodeGenPhase.cpp#L151

andrewcraik:
ping to @vijaysun-omr as the owner since this has been sitting for a while now - thoughts on this change?

compiler/x/codegen/OMRTreeEvaluator.cpp
0xdaryl: Yes, calling the version in `OMR::TreeEvaluatorConnector` is the correct way to call the parent project's implementation.

compiler/x/codegen/OMRTreeEvaluator.cpp
```diff
@@ -2249,39 +2249,29 @@ static void arraycopyForShortConstArrayWithoutDirection(TR::Node* node, TR::Regi
...
+   // that a 5-child arraycopy is generated indicates at least one of the first two children must be needed when performing the
+   // copy; otherwise a 3-child arraycopy should be generated instead. Interpreting the meanings of the first two children 
+   // definitely requires language specific semantics. For example, the first two children may be for dealing with an arraycopy 
+   // where the Garbage Collector may need to be notified about the copy or something to that affect.
    //
+   // Therefore, this OMR evaluator only handles the 3-child variant, is an operation equivalent to C++'s std::memmove().
+   // Should a downstream project need the 5-child variant evaluator, it needs to override this evaluator in its own TreeEvaluator,
+   // and delegates the 3-child variant back to this evaluator by calling OMR::TreeEvaluatorConnector::arraycopyEvaluator.
+
+   TR_ASSERT(node->getNumChildren() == 3, "This evaluator is for the 3-child variant of arraycopy.");
```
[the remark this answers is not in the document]
0dvictor: Good suggestion, it has been updated.

0xdaryl:
@genie-omr build xlinux,win

fjeremic:
> Is there a plan to remove/sink 5-child arraycopy support to J9 for non-x86 platforms? I'd hate to introduce an asymmetry amongst the code generators. At the very least create an umbrella issue to track this work please. @gita-omr @fjeremic

@r30shah given your recent work in the area can you tackle this one? Let's open up an item to track (one in OMR and one on our board).

r30shah:
@fjeremic  Just opened up issue : https://github.com/eclipse/omr/issues/3179 contains short summary of what is needed. Changes just requires some trivial shuffling.

0xdaryl:
Thanks @fjeremic.  I'm not going to merge this until I hear from Power (@gita-omr) or see an issue tracking this on Power.

gita-omr:
> Thanks @fjeremic. I'm not going to merge this until I hear from Power (@gita-omr) or see an issue tracking this on Power.

Going through the comments right now. Generally, would appreciate a bit more background in the description or some summary at the end.

gita-omr:
@0xdaryl I created a P issue.

0dvictor:
The original intention was to cleanup X86's `arraycopy` evaluator by removing J9-project-specific code for the following three reasons:
1. A downstream project specific code shouldn't exist in OMR;
2. These J9-project-specific code in `arraycopy` evaluator have been deprecated by OpenJ9; they are not used anymore;
3. Cleaning these up allows further cleanup in OpenJ9.

In the discussion of the pull request, it is concluded that handling 5-child variant of `arraycopy`
 always requires project-specific logic that OMR is hardly aware of, and hence 5-child variant related logic should be sunk to downstream projects.

PR #3062, opened 2018-10-13
Align comments in tree evaluator tables on X86

0dvictor:
Comments in tree evaluator tables are annoyingly misaligned, fixing.


0xdaryl:
@genie-omr build xlinux,win

PR #3066, opened 2018-10-15
Software RangeCheck  Read Barrier

amicic:
Renaming flag S/W Evacuate RB to S/W RangeCheck RB. 
Evacuate is an action that is required by RB in Concurrent Scavenger.
It's always executed in software. What has alternative implementation in
S/W vs H/W is the trigger mechanism, which is the range check against
Evacuate area boundaries (or a set of areas).

Temporarily, keeping the old variable and accessor, until downstream
projects are updated to use the new ones.

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -418,6 +418,7 @@ public:
 	bool scavengerRsoScanUnsafe;
 #if defined(OMR_GC_CONCURRENT_SCAVENGER)
 	bool softwareEvacuateReadBarrier; /**< enable software read barrier instead of hardware guarded loads when running with CS */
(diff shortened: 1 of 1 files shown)


fjeremic:
Could I please request a review from @dmitripivkine for this one?

fjeremic:
@genie-omr build xlinux

fjeremic:
@genie-omr build xlinux,zlinux,zos

dmitripivkine:
LGTM

charliegracie:
What is the difference between `softwareEvacuateReadBarrier` and `softwareRangeCheckReadBarrier`?

amicic:
Naming only. I want to rename the command line option (in OpenJ9) that controls it, but then the name of the internal flag to match it too.

PR #3067, opened 2018-10-15
Align heap for GS if necessary only

dmitripivkine:
Do heap memory alignments reqiured for GS in case if CS runs on
supported HW only. Request of using Software Barriers overwrites it.


dmitripivkine:
WIP, require https://github.com/eclipse/openj9/pull/3280 to be promoted

dmitripivkine:
@charliegracie FYI: OpenJ9 part has been promoted and this PR might be merged

PR #3069, opened 2018-10-15
Properly set array related feature flags on X86

0dvictor:
Guard following feature flags with "canGenerateArraylets":
1. SupportsArrayCmp
2. SupportsPrimitiveArrayCopy
3. SupportsArraySet
4. SupportsArrayTranslateTRTO
5. SupportsArrayTranslateTROT
6. SupportsArrayTranslateTROTNoBreak

Remove SupportsReferenceArrayCopy as OMR doesn't actually
fully support reference arraycopy.

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -369,13 +369,34 @@ OMR::X86::CodeGenerator::initialize(TR::Compilation *comp)
    self()->addSupportedLiveRegisterKind(TR_VRF);
    self()->setLiveRegisters(new (self()->trHeapMemory()) TR_LiveRegisters(comp), TR_VRF);
 
-   self()->setSupportsArrayCmp();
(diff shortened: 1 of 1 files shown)


andrewcraik:
This change makes sense to me - I assume that the major downstream consumer of these features, OpenJ9, avoids seeing these opcodes by disabling the opts that generate them under arraylets which is why the feature flag misconfiguration wasn't seen by them or do they disable these features themselves?

0dvictor:
> I assume that the major downstream consumer of these features, OpenJ9, avoids seeing these opcodes by disabling the opts that generate them under arraylets which is why the feature flag misconfiguration wasn't seen by them or do they disable these features themselves?

Yes, consumers of these opcode always check `canGenerateArraylets` before generating the opcode. This proposed changeset won't affect the behavior of existing implementation.

andrewcraik:
LGTM - @0xdaryl as the usual codegen committer could you have a look at this?

0xdaryl:
@genie-omr build xlinux,win

PR #3070, opened 2018-10-15
Fix OSX implementation of omrsig_info()

keithc-ca:
The code was returning addresses into a _copy_ of parts of the thread state eliminating the possibility that a signal handler might update that state before allowing a thread to resume.

An example modification is in OpenJ9's handler for division by zero which wants to alter the instruction pointer to a trap handler, but writing to that copy of thread state is ineffective.

### DIFF
--- a/port/osx/omrsignal_context.c
+++ b/port/osx/omrsignal_context.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2016 IBM Corp. and others
+ * Copyright (c) 2016, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


DanHeidinga:
@charliegracie @0xdaryl This looks reasonable to me and addresses allows OpenJ9 to take the correct actions (ie: run special trap handlers) in response to signals on mac, just as it does on linux.

Can one of you review this?  We need it for OpenJ9's 0.11.0 release which is imminent.

port/osx/omrsignal_context.c
```diff
@@ -26,7 +26,7 @@
 #include <unistd.h>
 
 void
-fillInUnixSignalInfo(struct OMRPortLibrary *portLibrary, void *contextInfo, struct OMRUnixSignalInfo *j9Info)
+	fillInUnixSignalInfo(struct OMRPortLibrary *portLibrary, void *contextInfo, struct OMRUnixSignalInfo *j9Info)
```
charliegracie: why is this tabbed in?

0xdaryl:
Seems like a reasonable change to me.

charliegracie:
@genie-omr build all

PR #3073, opened 2018-10-16
Remove bogus assertion in TR_IndirectCallSite

jdmpapin:
In findCallTargetUsingArgumentPreexistence(), it is possible for
getResolvedMethod() to fail to find the target. In this case, we should
simply skip adding the target, as already happens in builds without
assertions.


0xdaryl:
@genie-omr build all

0xdaryl:
Seems fine to me.  @andrewcraik, any objections?

andrewcraik:
no objection - looks bogus to me.

PR #3074, opened 2018-10-16
Guard code with SVM check

dsouzai:
This commit is for J9_PROJECT_SPECIFIC code. With the new Symbol
Validation Manager code, it is possible for the compiler to create a
class symbol with cpIndex -1 under AOT even if the class is not loaded
by the system class loader. However, as the code stands, this will
trigger a TR_ASSERT_FATAL. This commit ensures that the code that checks
for the system class loader only runs when the SVM is not enabled.

### DIFF
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -1303,7 +1303,7 @@ OMR::SymbolReferenceTable::findOrCreateClassSymbol(
    sym->setClassObject();
 
 #ifdef J9_PROJECT_SPECIFIC
-   if (cpIndex == -1 && comp()->compileRelocatableCode())
+   if (cpIndex == -1 && comp()->compileRelocatableCode() && !comp()->getOption(TR_UseSymbolValidationManager))
       {
       // Restricting TR_ArbitraryClassAddress to classes loaded by the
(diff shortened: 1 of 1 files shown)


dsouzai:
@jdmpapin could you review. @vijaysun-omr fyi.

vijaysun-omr:
Okay so this assert was present before the new symbol validation code was added, and it happened to be on a path where it would execute regardless of whether new symbol validation option was set or not, but it's no longer valid to execute with the new symbol validation option set ?

vijaysun-omr:
@genie-omr build all

dsouzai:
Yes; when using the symbol validation manager, this code should not execute because it is now valid for the compiler to generate a class symbol with cpIndex -1 even if the class is not loaded
by the system class loader - if the svm isn't used, then under AOT the compiler should not be generating a class symbol with cpIndex of -1 unless it is loaded by the system class loader.

vijaysun-omr:
Okay, got it. So it's a conservative change that limits when the assert is run to when the old AOT is in use since that was the original intent anyway (we did'nt have the symbol validation code at all when the assert was first added). The symbol validation explicitly wants to relax the condition the assert is checking and that's exactly what this commit does.

vijaysun-omr:
Merged this since all checks passed and it is a conservative change.

PR #3080, opened 2018-10-17
Prevent direct memory operation on rdbar

0dvictor:
Read Barriers have side effects and hence they cannot be skipped.

### DIFF
--- a/compiler/codegen/Analyser.cpp
+++ b/compiler/codegen/Analyser.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -52,12 +52,13 @@ void TR_Analyser::setInputs(TR::Node     *firstChild,
       setReg2();
       }
 
-   // The VFT pointer can be low-tagged, so generally we can't operate on it
-   // directly in memory.
-   //
+   // The VFT pointer can be low-tagged, so generally we can't operate on it directly in memory.
+   // Read Barriers have side effects and hence they cannot be skipped.
+
    TR::SymbolReference *vftPointerSymRef = TR::comp()->getSymRefTab()->element(TR::SymbolReferenceTable::vftSymbol);
 
    if (  firstChild->getOpCode().isMemoryReference()
+      && !firstChild->getOpCode().isReadBar()
       && firstChild->getSymbolReference() != vftPointerSymRef
       && firstChild->getReferenceCount() == 1
       && !lockedIntoRegister1)
@@ -66,6 +67,7 @@ void TR_Analyser::setInputs(TR::Node     *firstChild,
       }
 
    if (  secondChild->getOpCode().isMemoryReference()
+      && !secondChild->getOpCode().isReadBar()
       && secondChild->getSymbolReference() != vftPointerSymRef
       && secondChild->getReferenceCount() == 1
       && !lockedIntoRegister2)


andrewcraik:
This looks good and the change is only going to have an impact in modes that use rdbars so the risk is low and the change is pretty easy to verify through code inspection. I think this needs to be merged soon - OpenJ9 may want it for the 0.11 release since they are now starting to use the rdbar opcodes. Attn @vijaysun-omr @pshipton

fjeremic:
Could you please elaborate what this `TR_Analyzer` class does and why it is possible to skip a read barrier here? Perhaps I'm looking for some context here to understand the change.

fjeremic:
Sounds like we need a fatal tree verification pass to validate that all read barriers are properly anchored, otherwise there is no telling what other part of the optimizer may play the same tricks and avoid evaluation in favor of in memory operations which will result in a skipped read barrier. Those problems are brutal to debug so I'm heavily in favor of adding a verification pass.

andrewcraik:
@fjeremic totally agree, @yanluo7 is going to look at the anchoring issue and we should add that verification pass as an optional extra under an option just due to the compile-time it takes.

andrewcraik:
Since this is a workaround for anchoring the rdbar nodes properly I'm opening an issue to track fixing the anchoring and adding an optional verification pass which I'll link here. Once fixed this change could be reverted or left as the project sees fit.

fjeremic:
> @fjeremic totally agree, @yanluo7 is going to look at the anchoring issue and we should add that verification pass as an optional extra under an option just due to the compile-time it takes.

Cool! In light of this I'm fine with merging this as a temporary fix until the anchoring is improved. Assigning to @vijaysun-omr and recording my approval.

fjeremic:
@genie-omr build all

vijaysun-omr:
Okay so this is a conservative change to use a direct memory form in a compare ? Is it possible to "fold" the evaluation in this manner in any other places in the codegen ?

andrewcraik:
It _may_ be possible, but on x86 the majority of paths flow through the `TR_X86CompareAnalyser` which derives from `TR_Analyser`. I can't promise this will fix all cases, but it fixes a broad category and a proper fix based on validating correct anchoring is being built under issue #3081 which will ensure there is no need to play whack-a-mole in future. This is more rightly an interim fix which should have been called out earlier.

vijaysun-omr:
Given that we have compressed refs anchors in 64-bit compressed refs builds,  is the reference count going to be >= in comp refs ?

0dvictor:
I agree that we probably should anchor read barriers to comply the specs, but it does not solve all problems this pull request solves.
IMO even if read barriers are anchored, this change should still remain. Although `getReferenceCount() == 1` may accidentally exclude the case where the node is a read barrier, it doesn't mean the condition is proper. `getOpCode().isMemoryReference()` gives an impression that a "MemoryReference" node may be transformed to direct memory operation. This is no-longer true when read barrier is introduced so that read barriers should be excluded.

andrewcraik:
I believe so, especially when shift > 0.

vijaysun-omr:
I believe in a compressed refs scenario we have a bunch of other nodes that come "in between" the readbar and any comparison parent (or any other use that would be a parent) i.e. the nodes that do the compression/decompression. In other words, the pattern for direct memory memory operation would be broken invariably in a compressed refs scenario. Since pause-less readbars are used only under compressed refs in OpenJ9 v0.11 I do not think we need to do workarounds for the fact that readbars are not being anchored when they ought to be. In other words, let's fix this issue by doing the anchoring of readbars.

fjeremic:
> Although `getReferenceCount() == 1` may accidentally exclude the case where the node is a read barrier, it doesn't mean the condition is proper.

I disagree. The whole reason we initially imposed the restriction that the first occurrence of a read barrier must be anchored under a treetop is precisely to avoid this situation where we would sprinkle `isReadBar()` all over the codebase to avoid doing the wrong thing.

We have similar checks in the Z codegen where for a compressed refs shift of 1 we can fold away the shift by using the compressed object in both the base and index registers of the memory reference (the address is calculated as `base + index + offset`, so if `base == index` it's `base * 2 == base << 1`). We do not want this for read barriers. There are also lots of other codegen examples where we do this. These tricks are impossible in all situations if the node is anchored. i.e. the read barrier must be performed, no exceptions or tricks.

0dvictor:
> I believe in a compressed refs scenario we have a bunch of other nodes that come "in between" the readbar and any comparison parent (or any other use that would be a parent) i.e. the nodes that do the compression/decompression. In other words, the pattern for direct memory memory operation would be broken invariably in a compressed refs scenario. Since pause-less readbars are used only under compressed refs in OpenJ9 v0.11 I do not think we need to do workarounds for the fact that readbars are not being anchored when they ought to be. In other words, let's fix this issue by doing the anchoring of readbars.

> I disagree. The whole reason we initially imposed the restriction that the first occurrence of a read barrier must be anchored under a treetop is precisely to avoid this situation where we would sprinkle isReadBar() all over the codebase to avoid doing the wrong thing.

> We have similar checks in the Z codegen where for a compressed refs shift of 1 we can fold away the shift by using the compressed object in both the base and index registers of the memory reference (the address is calculated as base + index + offset, so if base == index it's base * 2 == base << 1). We do not want this for read barriers. There are also lots of other codegen examples where we do this. These tricks are impossible in all situations if the node is anchored. i.e. the read barrier must be performed, no exceptions or tricks.

I think there are misunderstandings of the intention of this changeset. It has NO indication on whether read barriers should be anchored. The first condition check `getOpCode().isMemoryReference()` seems improper to me. It should really be "if the opcode is one of `xload(i)` or `xstore(i)`", which is no-longer equivalent to `isMemoryReference` when read barrier is introduced. Therefore, the condition should be updated from a point of view of defensive programming.

fjeremic:
> I think there are misunderstandings of the intention of this changeset. It has NO indication on whether read barriers should be anchored. The first condition check `getOpCode().isMemoryReference()` seems improper to me. It should really be "if the opcode is one of `xload(i)` or `xstore(i)`", which is no-longer equivalent to `isMemoryReference` when read barrier is introduced. Therefore, the condition should be updated from a point of view of defensive programming.

I understand your point now, however what makes this location special? I see 10 more occurrences of `isMemoryReference()` in all platform codegens. All of them have a `getReferenceCount() == 1` check as well. It seems all these locations should also be looking for `xload(i)` or `xstore(i)` then. This is what I was talking about previously about sprinkling `!isReadBar()` all over the codebase. There are probably so many similar examples.

The fact that the tree is anchored already solves all the problems because wherever the second or subsequent reference of the read barrier occurs it's a register value at that point and you can do whatever optimization you normally do on register values. No need for extra checks and paranoia making the code harder to read.

0dvictor:
> I understand your point now, however what makes this location (in this PR) special? I see 10 more occurrences of isMemoryReference() in all platform codegens. All of them have a getReferenceCount() == 1 check right beside as well. It seems all these locations should also be looking for xload(i) then. This is what I was talking about previously about sprinkling !isReadBar() all over the codebase. There are probably so many similar examples.

Good point. Ideally there should be a query that answers true if and only if the opcode is one of `xload(i)` and all places that looks for `xload(i)` should use the new query. I agree that nothing makes this particular one special, all of them should either be updated in one shot or left unchanged.

> The fact that the tree is anchored already solves all the problems because wherever the second or subsequent reference of the read barrier occurs it's a register value at that point and you can do whatever optimization you normally do on register values. No need for extra checks and paranoia making the code harder to read.

Although anchoring makes all problems disappeared, it doen't mean this piece is robust. If anyone works in this part in the future, it is quite easy to miss the nontrivial information that 1. `isMemoryReference()` answers true not only for `xload(i)` but also `xrdbar(i)`; 2. `getReferenceCount() == 1` happens to filter out `xrdbar(i)` but it won't work if one wants to check a refcount != 1.

Anyway, I think we all agreed that this changeset neither fix any existing problem nor enhance any feature. I'm closing this PR.

PR #3020, opened 2018-10-02
AArch64: Implement BBStart/BBEnd evaluators

knn-k:
This commit implements BBStart/BBEnd evaluators for aarch64.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #3082, opened 2018-10-18
AArch64: Add ARM64RegBranchInstruction for br/blr/ret

knn-k:
This commit adds ARM64RegBranchInstruction for aarch64 branch
instructions that take a register as the target address.
It also introduces a new opcode that represents a return from a method,
which is separate from the opcode for aarch64 machine instruction "ret".

### DIFF
--- a/compiler/aarch64/codegen/ARM64BinaryEncoding.cpp
+++ b/compiler/aarch64/codegen/ARM64BinaryEncoding.cpp
@@ -102,11 +102,27 @@ uint8_t *TR::ARM64CompareBranchInstruction::generateBinaryEncoding()
    return cursor;
    }
 
+uint8_t *TR::ARM64RegBranchInstruction::generateBinaryEncoding()
+   {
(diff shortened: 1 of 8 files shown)


knn-k:
This commit also removes the redundant `_conditions` member and its related functions from `ARM64DepInstruction` because the parent `OMR::ARM64::Instruction` class has them.

knn-k:
The new opcode is named as `retn`.  It has to be separate from the machine instruction name `ret`, and the full-spelled `return` causes a C++ compilation error.

knn-k:
Opcode `fence` is introduced in #3020.

knn-k:
Removed "WIP:".

knn-k:
The Travis CI build failed with this message:
```
```

knn-k:
Rebased to the latest level for resolving conflicts caused by merging #3020.

0xdaryl:
@genie-omr build aarch64

Issue #3084, opened 2018-10-18
Crash on z/OS


hzongaro:
We're seeing an intermittent crash on z/OS in `OSRData.cpp`.  Here's a typical backtrace:

```
j> 19:09:22 TR_OSRCompilationData::writeMetaData(unsigned char*) const+0x156
j> 19:09:22 createMethodMetaData(TR_J9VMBase&,TR_ResolvedMethod*,TR::Compilation*)+0xb94
j> 19:09:22 TR::CompilationInfoPerThreadBase::compile(J9VMThread*,TR::Compilation*,TR_ResolvedMethod*,T...+0x148c
j> 19:09:22 TR::CompilationInfoPerThreadBase::wrappedCompile(J9PortLibrary*,void*)+0x1c3c
j> 19:09:22 omrsig_protect+0x79c
j> 19:09:22 TR::CompilationInfoPerThreadBase::compile(J9VMThread*,TR_MethodToBeCompiled*,J9::J9SegmentP...+0x4da
j> 19:09:22 TR::CompilationInfoPerThread::processEntry(TR_MethodToBeCompiled&,J9::J9SegmentProvider&)+0x8de
j> 19:09:22 TR::CompilationInfoPerThread::processEntries()+0x62a
j> 19:09:22 protectedCompilationThreadProc(J9PortLibrary*,TR::CompilationInfoPerThread*)+0x664
j> 19:09:22 omrsig_protect+0x79c
j> 19:09:22 compilationThreadProc(void*)+0x692
j> 19:09:22 thread_wrapper+0x9ba
j> 19:09:22 CELQPCMM+0xf80
j> 19:09:22 Unhandled exception
j> 19:09:22 Type=Segmentation error vmState=0x0005ff0b
j> 19:09:22 J9Generic_Signal_Number=00000004 Signal_Number=0000000b Error_Value=00000000 Signal_Code=00000035
j> 19:09:22 Handler1=000000500871F8F8 Handler2=000000500864F3F0
j> 19:09:22 gpr0=0000000000000000 gpr1=0000005022D4CC10 gpr2=000000507FFFE90C gpr3=00000050234BBDB8
j> 19:09:22 gpr4=0000005022AF8A80 gpr5=0000004F80000008 gpr6=0000004F80000008 gpr7=00000050234C33C0
j> 19:09:22 gpr8=FFFFFFFF00000008 gpr9=FFFFFFFF80000018 gpr10=00000050234C3F90 gpr11=0000000000000007
[...]

PR #3085, opened 2018-10-18
Avoid inlining TR_ScratchBufferInfo::writeToBuffer

hzongaro:
If TR_ScratchBufferInfo::writeToBuffer is inlined, it appears to expose
a bug in code generated by the IBM xlC compiler on z-Series.  The same
(or a similar) problem was previously seen on p-Series compiling with
xlC, and the work-around was to avoid inlining when compiling with xlC
on POWER.  This extends the same work-around to z compiling with xlC

Fixes: #3084

### DIFF
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -1490,7 +1490,9 @@ bool TR_OSRCompilationData::TR_ScratchBufferInfo::operator==(const TR_ScratchBuf
       scratchBufferOffset == info.scratchBufferOffset && symSize == info.symSize;
    }
 
-#if defined(TR_HOST_POWER) && defined(TR_TARGET_POWER) && defined(__IBMCPP__)
+#if (defined(TR_HOST_POWER) && defined(TR_TARGET_POWER)                       \
+           || defined(TR_HOST_S390) && defined(TR_TARGET_S390))               \
+        && defined(__IBMCPP__)
 __attribute__((__noinline__))
 //This tiny function when inlined by xlC 12 at -O3 breaks java -version
 #endif
--- a/compiler/compile/OSRData.cpp
+++ b/compiler/compile/OSRData.cpp
@@ -1494,7 +1494,8 @@ bool TR_OSRCompilationData::TR_ScratchBufferInfo::operator==(const TR_ScratchBuf
            || defined(TR_HOST_S390) && defined(TR_TARGET_S390))               \
         && defined(__IBMCPP__)
 __attribute__((__noinline__))
-//This tiny function when inlined by xlC 12 at -O3 breaks java -version
+//This tiny function when inlined by xlC 12 or later at -O3 breaks java -version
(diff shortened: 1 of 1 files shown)


compiler/compile/OSRData.cpp
```diff
@@ -1490,7 +1490,9 @@ bool TR_OSRCompilationData::TR_ScratchBufferInfo::operator==(const TR_ScratchBuf
       scratchBufferOffset == info.scratchBufferOffset && symSize == info.symSize;
    }
 
-#if defined(TR_HOST_POWER) && defined(TR_TARGET_POWER) && defined(__IBMCPP__)
+#if (defined(TR_HOST_POWER) && defined(TR_TARGET_POWER)                       \
+           || defined(TR_HOST_S390) && defined(TR_TARGET_S390))               \
+        && defined(__IBMCPP__)
 __attribute__((__noinline__))
 //This tiny function when inlined by xlC 12 at -O3 breaks java -version
```
fjeremic: I believe you observed this issue with xlC 2.1.1 hence this comment has not aged well. Can you improve the comment and mention it was originally seen on 1.2 on Power but reproduced on 2.1.1 on z/OS so that we have more context?

fjeremic:
@hzongaro FYI if you add the "Fixes:" keyword [1] followed by the issue number in the PR comment GitHub will automatically close the corresponding issue when this PR is merged. I've amended your PR comment with the appropriate syntax (note the dots under the "Fixes:" word and hover your mouse over it).

[1] https://help.github.com/articles/closing-issues-using-keywords/

0xdaryl:
These kinds of build compiler guards will be forgotten very quickly after getting buried in the source, leaving a potential performance opportunity on the table when the build compiler is eventually (quietly) fixed.  I'd like to request an issue be created to track this guard's eventual removal.  A GH issue is much easier to search than source code.  Now that I think of it, I could even create a special "Build Compiler Bug" label to highlight these kinds of problems.

Has this issue ever been raised with xlC?

fjeremic:
> Has this issue ever been raised with xlC?

I believe @hzongaro is in the process of doing this.

compiler/compile/OSRData.cpp
hzongaro: Thanks, Filip @fjeremic.  I've updated that comment in https://github.com/eclipse/omr/pull/3085/commits/42f72e85c25fa8cfd9d5c282835b72b8c419be69

fjeremic:
@hzongaro can you open up an issue for the the problem as mentioned by @0xdaryl and write down some details on the bug report you raise with the xlC teams so we can track this moving forward?

@genie-omr build all

hzongaro:
> @hzongaro can you open up an issue for the the problem as mentioned by @0xdaryl and write down some details on the bug report you raise with the xlC teams so we can track this moving forward?
Opened github issue #3088 to track the build compiler problem.

PR #3086, opened 2018-10-18
Avoid OOL instruction generation during OOM on X86

0dvictor:
OutlinedInstructionsGenerator to not generate instructions
when there is a uncaught exception.


0xdaryl:
Seems reasonable to me.  This is a pitfall in destructors that we need to raise awareness of somehow.

@dsouzai , can I have a review please?

0xdaryl:
@genie-omr build xlinux,win

JamesKingdon:
The fix missed an important corer case where the destructor itself is the first occurence of OOM which still leads to crashes.

PR #3090, opened 2018-10-18
Fix MSVC error C2398 requires narrowing conversion

samolisov:
The 'SimpifierFoldAbsNegTest' unit test from 'compilertriltest' contains
an initialization block for a vector of floats:

std::vector<  float> fTestData =
    { 0, 1, 2, -1, -2, 3.14, -3.14,
      std::numeric_limits<float>::min(), std::numeric_limits<float>::min() };

During a compilation with MSVC the following errors appears:

.\..\fvtest\compilertriltest\SimplifierFoldAbsNegTest.cpp(45): error
C2398: Element '6': conversion from 'double' to 'float' requires a
narrowing conversion
..\..\fvtest\compilertriltest\SimplifierFoldAbsNegTest.cpp(45): error
C2398: Element '7': conversion from 'double' to 'float' requires a
narrowing conversion

Literals '3.14' and '-3.14' can be represented as well 'double' as
'float', the 'f' suffix after the literal must be present to fix the
compilation error.

### DIFF
--- a/fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
+++ b/fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp
@@ -42,7 +42,7 @@ template <> char* nameForType< double>() { return "Double"; }
 
 std::vector<int32_t> iTestData = { 0, 1, 2, -1, -2, 99999, -99999, std::numeric_limits<int32_t>::max(), std::numeric_limits<int32_t>::min() };
 std::vector<int64_t> lTestData = { 0, 1, 2, -1, -2, 99999, -99999, std::numeric_limits<int64_t>::min(), std::numeric_limits<int64_t>::min() };
-std::vector<  float> fTestData = { 0, 1, 2, -1, -2, 3.14, -3.14, std::numeric_limits<float>::min(), std::numeric_limits<float>::min() };
+std::vector<  float> fTestData = { 0, 1, 2, -1, -2, 3.14F, -3.14F, std::numeric_limits<float>::min(), std::numeric_limits<float>::min() };
 std::vector< double> dTestData = { 0, 1, 2, -1, -2, 3.14, -3.14, std::numeric_limits<double>::min(), std::numeric_limits<double>::min() };
 
 template <typename T> std::vector<T> dataForType();


samolisov:
I see the all initialization blocks except `std::vector<int32_t> iTestData` contains `std::numeric_limits<int64_t>::min()` as both last elements:

```cpp
std::vector<int64_t> lTestData = { ..., std::numeric_limits<int64_t>::min(), std::numeric_limits<int64_t>::min() };
std::vector<  float> fTestData = { ..., std::numeric_limits<float>::min(), std::numeric_limits<float>::min() };
std::vector< double> dTestData = { ..., std::numeric_limits<double>::min(), std::numeric_limits<double>::min() };
```

Shall we use std::numeric_limits<int64_t>::**max()** for the second-to-last element of each vector?

0xdaryl:
@genie-omr build all

0xdaryl:
I don't see why we couldn't (shouldn't?) use `std::numeric_limits::max()` as part of the input data set for this test.  In fact, for the integer types, I think using `std::numeric_limits::min()` may actually be unsafe or incorrect for the two tests that use a "neg" operation because -MIN_INT is undefined behaviour.  Unless I'm reading this test incorrectly...  @Leonardo2718

Leonardo2718:
`std::numeric_limits::max()` should definitely be part of the input data set.

As for `std::numeric_limits::min()` in the integer case, the only *C++ undefined behaviour* in the tests is at line 133 when `std::abs(test)` is called:

https://github.com/eclipse/omr/blob/e8f65c40c0037a1be4f3db68ac54b3a2434852ca/fvtest/compilertriltest/SimplifierFoldAbsNegTest.cpp#L133

The real question should be: is -MIN_INT undefined behaviour in TR?

The correct thing to do is to replace the call to `std::abs(test)` with something that matches whatever the defined behaviour of `(abs (abs (<some value>))` is in the OMR compiler. If the behaviour is identical to the C++ function `std::abs()`, then yes `std::numeric_limits::min()` should be removed from the input data set for the integer cases.

0xdaryl:
I'm going to move the discussion around the use of `std::numeric_limits::max()` and `std::numeric_limits::min()` in this test into a separate issue as it doesn't affect the mergability of this PR.

PR #3091, opened 2018-10-19
Explicit thread count for parallel GC task

amicic:
Fixing 2 problems, when overriding thread count for a task (via run()
API), which may occur in a scenario when initial heap size is so small
to cause an early GC before even all GC threads are up and running: 

1) major: the overriding value should be no bigger than
_threadCount/_adjustedThreadCount, since there might not be
more slave threads available, yet. Otherwise, the task would hang
waiting for non-existant threads to complete.
2) minor: do not record the overriding value into Dispatcher's
_threadCount, since early at startup _threadCount might still be
constructed. Just pass the value to the task. Otherwise, the final
constructed value could be lost and subsequent GCs may run with reduced
GC counts.

Temporarily, keeping both old and new variant of prepareThreadsForTask()
API until downstream OMR projects start using the new API.

### DIFF
--- a/gc/base/Dispatcher.cpp
+++ b/gc/base/Dispatcher.cpp
@@ -57,7 +57,7 @@ MM_Dispatcher::initialize(MM_EnvironmentBase *env)
 }
 
 void
-MM_Dispatcher::prepareThreadsForTask(MM_EnvironmentBase *env, MM_Task *task)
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@youngar can you review this and merge once you are happy with the changes?  Thanks

PR #3093, opened 2018-10-19
Flush GC caches for concurrent marker thread

amicic:
GC threads at startup act briefly as mutators since they allocate from
heap (various thread specific objects). Before transitioning the thread
type to its specific GC role, they need to flush any GC specific caches,
as any mutators do.

Specifically, in Concurrent Scavenger, if a GC thread is created during
a very early cycle, it may trigger a read barrier and create a copy
cache. This cache has to by flushed (to scan queue for further scanning
in final STW GC phase).

This problem has already been addressed for Master and Slave GC threads
(https://github.com/eclipse/omr/pull/2696). Here Concurrent Marking
Background threads are treated with the identical fix. 


charliegracie:
@genie-omr build all

charliegracie:
Once the CI builds complete successfully any committer can feel free to merge.

PR #3094, opened 2018-10-19
State explicitly CS witout H/W assistance

amicic:
Slighly modify the verbose GC stanza where we state if Concurrent
Scavenger is enabled:

1) on H/W capable platforms we will state one of two:

enabled, with H/W assistance
enabled, without H/W assistance

2) on plaforms with no H/W assistance we will just state:

enabled

3) if CS is not enabled, regardless of platform, the stanza is not
issued

Since there is an overlap, also piggybacking to this change removal of
unused softwareEvacuateReadBarrier flag and retirment of jitLoaded field
in Initialized hook.


fjeremic:
@genie-omr build all

PR #3095, opened 2018-10-19
Add support for CUDA 10.0

keithc-ca:
According to [1], CUDA Toolkit version 10.0 was released in September 2018.

[1] https://developer.nvidia.com/cuda-toolkit-archive


charliegracie:
@genie-omr build all

PR #3099, opened 2018-10-22
AArch64: Add guard to Machine::assignOneRegister()

knn-k:
This code adds a guard code to Machine::assignOneRegister() for aarch64.
Do nothing and just return when a real register is already assigned to
a virtual register.

### DIFF
--- a/compiler/aarch64/codegen/OMRMachine.cpp
+++ b/compiler/aarch64/codegen/OMRMachine.cpp
@@ -443,30 +443,34 @@ TR::RealRegister *OMR::ARM64::Machine::assignOneRegister(TR::Instruction *curren
                                                          TR::Register *virtualRegister)
    {
    TR_RegisterKinds rk = virtualRegister->getKind();
-   TR::RealRegister *assignedRegister;
+   TR::RealRegister *assignedRegister = virtualRegister->getAssignedRealRegister();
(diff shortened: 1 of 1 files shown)


knn-k:
AppVeyor build failed with this message:
```
AppVeyor failed to assign job to 'gce' build environment after 3 attempt(s). The envronment is either unavailable or there are no free workers, however alternate build environment was not configured by the system administrator. Please try again later.
```

knn-k:
arm32 version does the same in `OMR::ARM::Machine::assignSingleRegister()`.

On the other hand, `OMR::Power::Machine::assignOneRegister()` does not check the real register in the function, but the callers of `assignOneRegister()` do that like this:
```
void TR::PPCConditionalBranchInstruction::assignRegisters(TR_RegisterKinds kindToBeAssigned)
   {
   TR::Register           *virtualRegister = getConditionRegister();
   TR::RealRegister       *assignedRegister = virtualRegister->getAssignedRealRegister();
   TR::Machine *machine = cg()->machine();

   if (assignedRegister == NULL)
      {
      assignedRegister = machine->assignOneRegister(this, virtualRegister, false);
      }
```

0xdaryl:
@genie-omr build aarch64

PR #3102, opened 2018-10-22
CI Update OMR README

aarongraham9:
CI: Update OMR README
-Remove old, removed, Travis CI Matrix Builds
-Add in new Eclipse OMR Jenkins CI Builds for AArch64 and ARM
-Add other missing Eclipse OMR Jenkins Builds

[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,19 +24,19 @@ SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-excepti
 
 | Build | Status |
 | ---------------------- | -------------------- |
-| **Travis CI (Overall)** | [![Travis Overall Status](https://api.travis-ci.org/eclipse/omr.svg?branch=master)](https://travis-ci.org/eclipse/omr) |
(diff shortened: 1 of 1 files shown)


aarongraham9:
As a follow up to the Eclipse OMR Jenkins Build Farm work. Any thoughts @0xdaryl, @Leonardo2718 and @charliegracie?

charliegracie:
Thanks! It is missing a few builds I think... There should be 10 OMR CI builds on Jenkins that get executed for every Push/Merge plus the single AppVeyor build and the single Travis build which makes 12 build links.

charliegracie:
I think the build ones.

aarongraham9:
Updated based on your comments. Any further thoughts @charliegracie, @Leonardo2718 or @0xdaryl?

youngar:
I double checked the list, it looks good to me.  It is strange that we have a working "PR/merge" build for z/os, but not a working "branch/regular" build.  I'm guessing there is no technical reason, it is just an oversight.

Please change `x86_64` to `x86-64` in `Linux x86_64 Compressed Pointers`.  I also think it might be better to change `LINT` to `Lint` in `Linux x86-64 LINT (Travis)`

aarongraham9:
Thanks, @youngar; I've updated the PR based on your comments. Also rebased off of the latest on `master`.

youngar:
Thanks @aarongraham9

PR #3104, opened 2018-10-22
WIP: Insert a GC map at catch block entry on X86

0dvictor:
A GC map may be needed at the entry of a catch block.

### DIFF
--- a/compiler/x/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/x/codegen/OMRTreeEvaluator.cpp
@@ -3434,6 +3434,9 @@ TR::Register *OMR::X86::TreeEvaluator::BBStartEvaluator(TR::Node *node, TR::Code
 
       inst->setNeedToClearFPStack(true);
 
+      if (block->isCatchBlock())
+         inst->setNeedsGCMap(0xFF00FFFF);
+
       node->getLabel()->setInstruction(inst);
       block->setFirstInstruction(inst);
 


0dvictor:
@andrewcraik @fjeremic @gacholio FYI.

andrewcraik:
@0dvictor I assume you are going to leave the non-merging to a common code change? Also could you at least reference the design in the PR message to make it clear what this is in relation to please?

gacholio:
Yes, there should be a "fixes: " in the title, referring to the issue.

fjeremic:
> Yes, there should be a "fixes: " in the title, referring to the issue.

I don't think this will work in this case because the issue this is fixing is in OpenJ9. However we should associate the work by adding the "Issue: eclipse/openj9#3374" in the PR description right above the sign-off tag.

gacholio:
Sorry, missed the cross-repo reference.

Interestingly, I did not fix it this way - I modified somewhere else that was testing FSD, so I'm wondering if this is the proper change. I think we may be leaning towards a new flag to control this, rather than enabling it always, which will be more complex if the change is made on OMR.

andrewcraik:
My initial gut feel is that having a flag on a gcmap saying it must be preserved may be the way to go - at the OMR level it would be a useful utility for a language and then we can decouple required preservation from FSD where preservation is implied...

fjeremic:
> Interestingly, I did not fix it this way - I modified somewhere else that was testing FSD, so I'm wondering if this is the proper change. I think we may be leaning towards a new flag to control this, rather than enabling it always, which will be more complex if the change is made on OMR.

Do you have the location of where you made your change for reference? We can examine the tradeoffs.

gacholio:
It's detailed in the epic.

0dvictor:
Reading through the above valuable comments, I feel this support shouldn't be part of OMR. Closing this PR.

Issue #3034, opened 2018-10-05
Rename Symbol::staticName to staticFieldName


yanluo7:
Currently the callers of staticName are assuming they're asking about a static field name. So to avoid confusion, staticName should be renamed to staticFieldName to prevent user of this call to guard it under insufficient conditions (See https://github.com/eclipse/omr/pull/3031).

PR #3108, opened 2018-10-23
Rename Symbol::staticName to staticFieldName

nanjekyejoannah:
Rename Symbol::staticName to staticFieldName

This PR renames staticName  to staticFieldName.

Closes: #3034

Signed-off-by: nanjekyejoannah <nanjekyejoannah@gmail.com>

### DIFF
--- a/compiler/compile/Method.cpp
+++ b/compiler/compile/Method.cpp
@@ -445,10 +445,10 @@ int32_t      TR_ResolvedMethod::classCPIndexOfFieldOrStatic(int32_t)       { not
 const char * TR_ResolvedMethod::signature(TR_Memory *, TR_AllocationKind)  { notImplemented("signature"); return 0; }
 const char * TR_ResolvedMethod::externalName(TR_Memory *, TR_AllocationKind)  { notImplemented("signature"); return 0; }
 char *       TR_ResolvedMethod::fieldName (int32_t, TR_Memory *, TR_AllocationKind kind)           { notImplemented("fieldName"); return 0; }
-char *       TR_ResolvedMethod::staticName(int32_t, TR_Memory *, TR_AllocationKind kind)           { notImplemented("staticName"); return 0; }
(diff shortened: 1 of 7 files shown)


nanjekyejoannah:
@andrewcraik , I understand the implications of this change after this review. Thanks for the insight. Could I have misunderstood the issue? or there is  a better solution to fix the issue?

 If it is long term that requires downstream users to migrate their code to the new API as you say , then am glad to close this as it does not seem feasible for now.

andrewcraik:
I don't think you need to close this as such - I would suggest you do the change in two steps - in this commit do the rename as you have done, but add a staticName which calls staticFieldName. This way downstream  consumers can continue to use the old API for a period of time. The downstream projects (primarily Eclipse OpenJ9 in this case I think) can update to use the new staticFieldName method. Once that migration is done you can delete the staticName delegating methods you added for compatibility as a second pull request.

The rename as descried in #3034 is certainly something we would like to see done and it is great that you are helping out. Even if you just do step one above it would be progress towards getting rid of the somewhat misleading staticName method.

nanjekyejoannah:
@andrewcraik +1 for that. 

I will look at this in 10 days as am currently abit busy atm

Leonardo2718:
@nanjekyejoannah any updates? :slightly_smiling_face:

0xdaryl:
@nanjekyejoannah, there hasn't been an update on this PR in some time.  Are you still planning on making progress with it in the near future?  If not, I suggest closing it while you work on it in your private fork and then re-open when it is ready for review.  There are downstream dependencies to this change as well (i.e., in OpenJ9) so this PR will need to take on a bit of a different form (as discussed above) to accommodate staging those changes so that builds don't break.

nanjekyejoannah:
Closing this as am short on bandwidth at the moment. I can't promise to look at this now. When I do in the future, I will reopen.

0xdaryl:
OK, thanks.  Closing.

PR #3110, opened 2018-10-23
Use _extensions in VGC handler

amicic:
There is no local extensions declared in Initialized staza handler. Use
the class member instead.


fjeremic:
@genie-omr build zlinux

PR #3111, opened 2018-10-23
Remove pointless obtaining of GR1

shingarov:
Signed-off-by: Boris Shingarov <shingarov@labware.com>

### DIFF
--- a/compiler/p/codegen/OMRCodeGenerator.cpp
+++ b/compiler/p/codegen/OMRCodeGenerator.cpp
@@ -1712,9 +1712,7 @@ void OMR::Power::CodeGenerator::doBinaryEncoding()
    self()->setBinaryBufferCursor(temp);
    self()->alignBinaryBufferCursor();
 
-   TR::Instruction *nop;
    bool skipLabel = false;
-   static int count=0;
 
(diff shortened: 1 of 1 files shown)


compiler/p/codegen/OMRCodeGenerator.cpp
```diff
@@ -1713,7 +1713,6 @@ void OMR::Power::CodeGenerator::doBinaryEncoding()
    self()->alignBinaryBufferCursor();
 
    TR::Instruction *nop;
-   TR::Register *gr1 = self()->machine()->getPPCRealRegister(TR::RealRegister::gr1);
    bool skipLabel = false;
    static int count=0;
```
fjeremic: While we're in the area can we remove `count` too which doesn't appear to be used? Can we push `nop` down into the two `if` blocks which use the variable thus significantly limiting it's scope. There is no reason for it to be declared way up here.

On a more stylistic note please try to avoid periods on commit headers as per the commit guidelines [1] as it makes reading `git log` much easier if all commits are consistent:

> The first line should be less than 70 characters, where reasonable, and should be written in sentence case preferably not ending in a period. Leave a blank line between the first line and the message body.

[2] https://chris.beams.io/posts/git-commit/

fjeremic:
@genie-omr build plinux,aix

Issue #2848, opened 2018-08-14
Use the instruction property `.format` to extract format in Z codegen


fjeremic:
As discussed in https://github.com/eclipse/omr/pull/2840#issuecomment-412677381 we currently have a very cumbersome way of getting to the instruction format. As part of #2561 we introduced a `.format` property for every instruction we support. We need to make use of this field and get rid of currently logic for determining formats.

fjeremic:
TODOs in #2869 have been addressed in #2932. I think we have all the tools necessary to close off this issue now. Working towards this as we speak.

PR #3113, opened 2018-10-23
Prepare to unify instruction formats and instruction kinds in Z codegen

fjeremic:
This PR consists of several incremental steps. The high level explanation of the changes requires a little bit of background. As part of #2848 we're trying to unify the `TR::Instruction::Kind` enum values to those of the instruction formats defined in OMRInstOpCode.hpp. 

To do this however we must make sure that `TR::Instruction` objects have an immutable "kind". Before this PR this was not the case as the Z codegen had a `setKind(...)` API which is used as a hack to support long displacement. Before this PR we would at instruction selection time (via `useSourceMemoryReference`) and at binary encoding time (via `S390MemoryReference::generateBinaryEncoding`) change the "kind" of an RX,RS,SI instruction to that of their long displacement counterparts.

This is incorrect on so many levels as we would then have potentially an `S390RXInstruction` instance that really represents an RXY instruction. The encoding and all the APIs called know only about RX instructions and there were hacks in there to handle long displacement.

The end goal here is to eliminate the `setKind(...)` APIs, which means a lot of work in the area of handling long displacement properly. We effectively do two things in this PR:

1. Push the long displacement logic up into the `generate*Instruction` APIs and return the correct instruction to begin with rather than modify existing instructions to "fake" long displacement support
2. Modify binary encodings of RX,RS,SI instructions to handle long displacement in case we reach binary encoding and determine we cannot encode the calculated displacement

Issue: #2848


fjeremic:
Given the size of the change I'd like to request reviews from multiple people including @joransiu, @r30shah, and @NigelYiboYu. Please proceed with the reviews as I suspect it may take some time. Meanwhile I have been testing the change in the background and I will carry out performance tests as well to make sure I didn't do anything silly to introduce regressions.

compiler/z/codegen/OMRInstruction.hpp
```diff
@@ -144,8 +144,7 @@ class OMR_EXTENSIBLE Instruction : public OMR::Instruction
    void setDebugHookOp() {_flags.set(DebugHookOp);}
 
    bool hasLongDisplacementSupport();
-   TR::InstOpCode::Mnemonic opCodeCanBeAdjustedTo(TR::InstOpCode::Mnemonic);
-   void attemptOpAdjustmentForLongDisplacement();
+   static TR::InstOpCode::Mnemonic opCodeCanBeAdjustedTo(TR::InstOpCode::Mnemonic);
```
fjeremic: In subsequent PRs I think I want to rename this and probably push it out to OMRInstOpcode.cpp.

compiler/z/codegen/OMRLinkage.hpp
```diff
@@ -567,9 +566,8 @@ enum TR_DispatchType
    virtual TR::RealRegister::RegNum getPrivateStaticBaseRegister()        { return _privateStaticBaseRegister; }
    virtual TR::RealRegister *getPrivateStaticBaseRealRegister();
 
-   virtual TR::RealRegister::RegNum setExtCodeBaseRegister (TR::RealRegister::RegNum r)   { return _extCodeBaseRegister = r; }
-   virtual TR::RealRegister::RegNum getExtCodeBaseRegister()    { return _extCodeBaseRegister; }
-   virtual TR::RealRegister *getExtCodeBaseRealRegister();
+   // TODO (#2848): Once the downstream dependency is removed we can fold this setter
+   virtual TR::RealRegister::RegNum setExtCodeBaseRegister (TR::RealRegister::RegNum r)   { return TR::RealRegister::GPR7; }
```
fjeremic: We'll clean this up right after this PR is merged. I'll have the downstream dependency PR up soon.

fjeremic: https://github.com/eclipse/openj9/pull/3419

fjeremic:
Marking as WIP to prevent accidental merge until I've done the performance measurements.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -2942,254 +2941,231 @@ OMR::Z::MemoryReference::generateBinaryEncoding(uint8_t * cursor, TR::CodeGenera
...
-
-   //calculate an estimate for the displacment of this reference
-   disp = self()->calcDisplacement(cursor, instr, cg);
+   int32_t displacement = self()->calcDisplacement(cursor, instr, cg);
 
    //add project specific relocations for specific instructions
-   self()->addInstrSpecificRelocation(cg, instr, disp, cursor);
+   self()->addInstrSpecificRelocation(cg, instr, displacement, cursor);
+
+   // We may ned up "upgrading" the instruction in case of long displacement, so this variable may change
```
r30shah: Seems like a typo here.

r30shah:
Thanks @fjeremic  for making these changes. I have started reviewing changes and as this is pretty big changes going through the commits one by one. So far reviewed following commits.
1. https://github.com/eclipse/omr/pull/3113/commits/1fcd64607605d0c4a64adf527accf63b53c86639 : Looks good to me, know where in codebase I see use case. 
2. https://github.com/eclipse/omr/pull/3113/commits/121da10771739d16cad74cb28e0b2caf19aace34 : LGTM
3. https://github.com/eclipse/omr/pull/3113/commits/e97aabd8d39a54f25f8493675f4db65a479f635b : LGTM

I also looked into https://github.com/eclipse/omr/pull/3113/commits/d2aeed80bfa529491a7aa48666cb7ce8f503daf1 apart from one typo looked fine. I will go through this one and other closely tied commits once again.

fjeremic:
A note to reviewers that this change actually exposes some existing bugs in the codebase. In particular I have an example from the downstream OpenJ9 project [1] where we're incorrectly using `generateRXInstruction` to generate an RXY instruction! This is a big mistake which would have worked prior to this change as we would "fix up" the RX into an RXY as soon as we realized the displacement was negative. 

This is really bad! This bug was not easy to reproduce either, so to avoid these "silent" issues I'm going to be adding another commit which will add fatal asserts inside of the constructors of our instruction classes to validate that the supplied mnemonic is of the correct format.

[1] https://github.com/eclipse/openj9/blob/29161a8487e12f0633ff0de9ff81fbabdae9a341/runtime/compiler/z/codegen/S390Recompilation.cpp#L369-L371

compiler/z/codegen/OMRMemoryReference.cpp
fjeremic: Fixed as part of e14011a

fjeremic:
Removing WIP. I've done all the internal testing I could and ran performance numbers. All tests are clean and performance looks good. I've done my due diligence to convince myself the change is sound. Off to the reviewers now @joransiu, @r30shah, @NigelYiboYu.

### Performance Evaluation

I built the change with OpenJ9 and ran the ODM ILOG and Liberty DayTrader7 benchmarks with debug counters enabled on the builds to spit out how many time we had to upgrade instructions for long and huge displacement.

Overall the throughput of the benchmarks was on par (which is expected). The static debug counters show no extra instructions being upgraded and no huge displacement being used, which is also expected. 

ODM ILOG 5-FastPath:

```
Baseline:
Sum Throughput (TPS) per thread for run = 12526.48
Global Throughput (TPS) for run = 12526.48
Average response time for run = 0.3193240599744408 ms
Min Execution Duration: 0
Max Execution Duration: 90
Min TPS for one thread: 3126.2025
Max TPS for one thread: 3136.9425

== Static debug counters ==
  2: z/memref/long-displacement-upgrade                                                          |           20 |    100.00% |   100.00% |  __   2 __
  3: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/InternalSHA.computeBlock()V) |            8 |     40.00% |    40.00% |    40.00% |  __   3 __
  4: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/SHA.implCompress([BI)V)      |           12 |     60.00% |    60.00% |    60.00% |  __   4 __

Experimental:
Sum Throughput (TPS) per thread for run = 12549.91
Global Throughput (TPS) for run = 12549.91
Average response time for run = 0.31872773908850127 ms
Min Execution Duration: 0
Max Execution Duration: 52
Min TPS for one thread: 3133.5425
Max TPS for one thread: 3142.245

== Static debug counters ==
  2: z/memref/long-displacement-upgrade                                                          |           20 |    100.00% |   100.00% |  __   2 __
  3: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/InternalSHA.computeBlock()V) |            8 |     40.00% |    40.00% |    40.00% |  __   3 __
  4: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/SHA.implCompress([BI)V)      |           12 |     60.00% |    60.00% |    60.00% |  __   4 __
```

Liberty DayTrader7:

```
Baseline:
10100.8/s

== Static debug counters ==
  2: z/memref/long-displacement-upgrade                                                          |           24 |    100.00% |   100.00% |  __   2 __
  3: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/InternalSHA.computeBlock()V) |            8 |     33.33% |    33.33% |    33.33% |  __   3 __
  4: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/SHA.implCompress([BI)V)      |            8 |     33.33% |    33.33% |    33.33% |  __   4 __
  5: z/memref/long-displacement-upgrade/(com/ibm/ws/sib/utils/CryptoHash.calculate([B)[I)        |            8 |     33.33% |    33.33% |    33.33% |  __   5 __

Experimental:
10103.5/s

== Static debug counters ==
  2: z/memref/long-displacement-upgrade                                                          |           28 |    100.00% |   100.00% |  __   2 __
  3: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/InternalSHA.computeBlock()V) |            8 |     28.57% |    28.57% |    28.57% |  __   3 __
  4: z/memref/long-displacement-upgrade/(com/ibm/security/bootstrap/SHA.implCompress([BI)V)      |           12 |     42.86% |    42.86% |    42.86% |  __   4 __
  5: z/memref/long-displacement-upgrade/(com/ibm/ws/sib/utils/CryptoHash.calculate([B)[I)        |            8 |     28.57% |    28.57% |    28.57% |  __   5 __
  ```

File: compiler/z/codegen/OMRMemoryReference.cpp [hunk 4]
```diff
@@ -2940,282 +2935,121 @@ OMR::Z::MemoryReference::generateBinaryEncoding(uint8_t * cursor, TR::CodeGenera
...
+   auto instructionFormat = instr->getKind();
 
-   // Most cases we will know at inst selection phase that the disp is big.  But it is still possible
-   // for us to hit bin phase and realize the DISP is large (eg when the stack is large).
-   if (disp<0 || disp>=MAXDISP)
+   // In most cases we will know at instruction selection that the displacement is large. But it is still possible for
+   // us to reach binary encoding without knowing a displacement will be large. A typical example of this is when 
+   // referencing stack symbols whose displacement will be unknown until the stack is mapped; an operation that happens
+   // after instruction selection and before binary encoding.
+   if (displacement < 0 || displacement >= MAXDISP)
```
NigelYiboYu: Maybe we can use the newly added `MINDISP` instead of a 0?

compiler/z/codegen/S390Instruction.cpp
```diff
@@ -1057,57 +1057,7 @@ TR::S390BranchOnIndexInstruction::generateBinaryEncoding()
...
-            {
-            (*(uint32_t *) cursor) = boi(0x86007000 | (regMask << 16));
-            }     // BXH Rx,Ry,0(r7)
-         else // TR::InstOpCode::BRXLE
-         {
-            (*(uint32_t *) cursor) = boi(0x87007000 | (regMask << 16));
-            }     // BXLE Rx,Ry,0(r7)
-         cursor += 4;
-         }
+      TR_ASSERT_FATAL(false, "Cannot encode branch on index instruction because distance (%d) is out of range", distance);
```
NigelYiboYu: By removing this block, we are unable to jump to labels that are more than 32K away. +-32K is usually enough but should we keep this long jump  just in case?

r30shah:
Checked last added commits and also skimmed through old commits. LGTM

compiler/z/codegen/S390Instruction.cpp
fjeremic: The branch on index instructions are only used for tight arraycopy loops (~6-10 instructions in length). The code removed is based on the assumption that the extended code base register is available, which is definitely not the case always. The change here strengthens this assumption by asserting in production so we don't go off and encode uses of GPR7 when we're not supposed to.

As to whether we should implement this path? I suppose so though I don't see a particular use case at the moment. However this is outside the scope of this PR. I've left the code in a better shape than it was before.

compiler/z/codegen/OMRMemoryReference.cpp
fjeremic: Good catch. Thanks! Fixed in 0514fc6

fjeremic:
@0xdaryl could I get this assigned to you as the committer?

0xdaryl:
@fjeremic , yes.  I'm short on cycles the next couple of days, but will get to it mid-week if that's ok.

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -1094,81 +1109,156 @@ TR::Instruction *
...
 
-   TR::Instruction *instr;
-   if (preced)
-      instr = new (INSN_HEAP) TR::S390RSInstruction(op, n, treg, mf, preced, cg);
-   else
-      instr = new (INSN_HEAP) TR::S390RSInstruction(op, n, treg, mf, cg);
+   TR::Instruction* result = NULL;
 
-   return instr;
+   int32_t displacement = mf->getOffset();
```
0xdaryl: This pattern repeats across the various generateRS instructions (or at least it appears to be very similar).  Can it be pulled into a separate function to save some code space?

compiler/z/codegen/BinaryAnalyser.cpp
```diff
@@ -227,7 +227,17 @@ TR_S390BinaryAnalyser::genericAnalyser(TR::Node * root,
                tempMR->enforce4KDisplacementLimit(secondChild, cg(), NULL);
                }
 
-            generateRXInstruction(cg(), memToRegOpCode, root, thirdReg, tempMR);
+            auto instructionFormat = TR::InstOpCode(memToRegOpCode).getInstructionFormat();
```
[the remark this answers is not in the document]
fjeremic: The choice between RX and RXE instructions should really exist. Looking at the two use cases of this pattern in BinaryAnalyser.cpp and BinaryCommutativeAnalser.cpp the code is organized in such a way that forces us to distinguish between the two. Examining the code further it looks like it can definitely be simplified so such checks are not even necessary.

They are added here in this way to satisfy the asserts which are added to make sure we're not using the wrong `generate*Instruction` API. I would prefer not to over complicate these APIs as it would set a dangerous precedent (i.e. why not have a `generateRXorRXEorRXYorRXSInstruction`? - which one are we generating then?).

RX and RXE instructions are inherently different formats. One has a mask where the other one doesn't. As opposed to RX and RXY which have the exact same operands except the RXY form has a slightly larger displacement.

compiler/z/codegen/SystemLinkage.cpp
```diff
@@ -1453,26 +1451,19 @@ TR::S390SystemLinkage::mapStack(TR::ResolvedMethodSymbol * method, uint32_t stac
...
-         }
-      if (comp()->getOption(TR_TraceCG))
-         traceMsg(comp(), "\n\nOffsetToLongDispSlot = %d\n", getOffsetToLongDispSlot());
+      //zLinux has a special reserved slot in the linkage convention
+      setOffsetToLongDispSlot(TR::Compiler->target.is64Bit() ? 8 : 4);
       }
    else
       {
-      setOffsetToLongDispSlot(0);
+      //setOffsetToLongDispSlot(stackIndex -= cg()->sizeOfJavaPointer());
```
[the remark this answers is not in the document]
fjeremic: Fixed in 15d4d15

compiler/z/codegen/S390GenerateInstructions.cpp
fjeremic: Fixed in 5c826f3

fjeremic:
> - in the commit title for 522521d, please put a space between TR_DisableLongDispNodes and as
> - in commit message for 86c153e, first sentence "ned" -> "end"

These are fixed in b0ea9b6 and ba5621d respectively.

compiler/z/codegen/BinaryAnalyser.cpp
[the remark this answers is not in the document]
fjeremic: I'm going to try and fix this up in the followup PRs where I will clean up the remanences of things we deprecate here (extended code base register, etc.).

0xdaryl:
@genie-omr build zlinux

0xdaryl:
There's a CI failure:
```
21:40:46 22: Assertion failed at /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/compiler/z/codegen/S390GenerateInstructions.cpp:830: instructionFormat == RXYa_FORMAT
21:40:46 22: 	Mnemonic (ADB) is incorrectly used as an RXY instruction
```

fjeremic:
I'll take a look at this tomorrow. This is an assertion added in this PR to catch these sorts of preexisting bugs. This looks like another occurrence which I haven't seen before in my testing.

0xdaryl:
+1.  Another example of why we need more asserts and asserts builds.  :-)

fjeremic:
@genie-omr build zlinux

fjeremic:
@0xdaryl issues fixed in 6b90dd6. We should be good to go here.

PR #3118, opened 2018-10-24
Set Specialized for arraycopy direction on IBM Z

r30shah:
ValuePropagation in OMR uses this flag to generate specialized trees for array copy direction if this flag is set. Currently Z platform expects optimizer to generate test for array copy direction. Set this flag in Z CodeGenerator.


r30shah:
@fjeremic Can you please review this. I am moving setting this flag from OpenJ9 (https://github.com/eclipse/openj9/blob/7b34bfd2bed88c5d0e10118c1a6bab50ffb84248/runtime/compiler/z/codegen/J9CodeGenerator.cpp#L140) to OMR where it rightfully sits.

fjeremic:
@genie-omr build zlinux,zos

PR #3119, opened 2018-10-24
Store sync before forwarding in CS

amicic:
For Concurrent Scavenger, on weak memory model, ensure that a candidate
copy is visible before (potentially) winning forwarding.

This covers the more common case of small object forwarding where we
allow duplicates (no cooperative copying of long objects).

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -1463,7 +1463,13 @@ MM_Scavenger::copy(MM_EnvironmentStandard *env, MM_ForwardedHeader* forwardedHea
 	/* Concurrent Scavenger can update forwarding pointer only after the object has been copied
 	 * (since mutator may access the object as soon as forwarding pointer is installed) */
 	}
-	if (originalDestinationObjectPtr == (allowDuplicate ? (destinationObjectPtr = forwardedHeader->setForwardedObject(destinationObjectPtr)) : destinationObjectPtr)) {
+	if (allowDuplicate) {
+		/* On weak memory model, ensure that this candidate copy is visible
+		 * before (potentially) winning forwarding */
+		MM_AtomicOperations::storeSync();
+		destinationObjectPtr = forwardedHeader->setForwardedObject(destinationObjectPtr);
+	}
+	if (originalDestinationObjectPtr == destinationObjectPtr) {
 		/* Succeeded in forwarding the object */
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

amicic:
If I read it correctly, it is a no-op on X86:
https://github.com/eclipse/omr/blob/master/include_core/AtomicSupport.hpp#L254

On Z it ends up being a fence. @fjeremic , does this effectively equal to no-op?

fjeremic:
> On Z it ends up being a fence. @fjeremic , does this effectively equal to no-op?

This only pertains to z/OS. According to page 376 of [1] `__fence()` is a compiler build-in which acts as a barrier for ordering optimizations done by the compiler. This should reduce to a no-op. A `J9ZOSRWB` would actually issue a `BCR 0,15` instruction which is a hard read write flush barrier in hardware.

fjeremic:
It is interesting however that thw following two are **not** identical:

```
readWriteBarrier();
```
vs.
```
readBarrier();
writeBarrier();
```

gc/base/standard/Scavenger.cpp
```diff
@@ -1463,7 +1463,13 @@ MM_Scavenger::copy(MM_EnvironmentStandard *env, MM_ForwardedHeader* forwardedHea
 	/* Concurrent Scavenger can update forwarding pointer only after the object has been copied
 	 * (since mutator may access the object as soon as forwarding pointer is installed) */
 	}
-	if (originalDestinationObjectPtr == (allowDuplicate ? (destinationObjectPtr = forwardedHeader->setForwardedObject(destinationObjectPtr)) : destinationObjectPtr)) {
+	if (allowDuplicate) {
+		/* On weak memory model, ensure that this candidate copy is visible
+		 * before (potentially) winning forwarding */
+		MM_AtomicOperations::storeSync();
+		destinationObjectPtr = forwardedHeader->setForwardedObject(destinationObjectPtr);
```
charliegracie: I am surprised this is not enforced with the compare and swap that happens in the setForwardedObject.

charliegracie: Ok after some digging around I agree that this barrier is required on PPC HW. The compare and swap instruction following does not actually mean the data in the object will be visible but rather just that every thread would see the forwarding pointer. This storeSync forces the data in the newly copied object to be visible. Thanks @amicic

PR #3120, opened 2018-10-24
Remove unused ParallelDispatcher API

amicic:
Now, that there are no more users of old prepareThreadsForTask()
(without explicit threadCount), remove it.


youngar:
@genie-omr build all

Issue #3121, opened 2018-10-25
A lot of errors during compilation OMROptions.cpp using Clang 8


samolisov:
I use **Clang 8** on Windows (actually built from the latest *master* on 24.10.2018) to build OMR with the compiler and get a lot of the following compilation errors on `OMROptions.cpp`:

```
..\..\compiler\control\OMROptions.cpp(760,94):  error: constant expression evaluates to -2147483648 which cannot be narrowed to type 'uintptrj_t' (aka 'unsigned long long') [-Wc++11-narrowing]
   {"enableTraps",                        "C\tenable trap instructions",                     RESET_OPTION_BIT(TR_DisableTraps), "F"},
                                                                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I see the `RESET_OPTIONS_BIT` macro is defined as:

```cpp
#define RESET_OPTION_BIT(x) TR::Options::resetBit, offsetof(OMR::Options,_options[(x)&TR_OWM]), ((x)&~TR_OWM)
```

The latest component, **((x)&~TR_OWM)**, evaluates at the compile time and is used as a substitution for the `uintptrj_t parm2` field of the `OptionTable` struct (see `OptionsUtil.hpp`). So, as I understand, `uintptrj_t` is defined as **unsigned** *long long* and is not able to take a signed value.

For example, `TR_DisableTraps` is defined as `0x80 00 00 00 + 7` what equals to `0x80 00 00 07`, `TR_OWM` is `0x00 00 00 1F`, the inversion of `TR_OWM` (`~TR_OWM)` equals to `0xFF FF FF E0`, and the conjunction will be equal to `0x8F FF FF E0`. So this is a negative value for a 32 bits system but not for a 64 bits one.

I believe the error is about the constant can be represent by a less width value and the compiler 'wants' to get a constant value of the right with and doesn't know what to do with the sign.

The problem can be solved using explicit casting in the following 3 macros:
 - `SET_OPTION_BIT(x)`
 - `RESET_OPTION_BIT(x)`
 - `SET_TRACECG_BIT(x)`

PR #3122, opened 2018-10-25
Fix cannot be narrowed to type 'uintptrj_t' error

samolisov:
A lot of the following compilation errors occur on OMROptions.cpp during
compilation with Clang 8:

..\..\compiler\control\OMROptions.cpp(760,94):  error: constant expression evaluates to -2147483648 which cannot be narrowed to type 'uintptrj_t' (aka 'unsigned long long') [-Wc++11-narrowing]
{"enableTraps", "C\tenable trap instructions", RESET_OPTION_BIT(TR_DisableTraps), "F"},

The reason is a narrowing the latest component of the RESET_OPTION_BIT
macro '((x)&~TR_OWM)' to the 'uintptrj_t' aka
'unsigned long long' type. The expression must be explicitly converted
to 'uintptrj_t'.

Closes: #3121

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -58,9 +58,9 @@
 
 using namespace OMR;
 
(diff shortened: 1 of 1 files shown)


fjeremic:
@genie-omr build all

PR #3125, opened 2018-10-26
Avoid unnecessary MemRef materialization on X86-64

0dvictor:
It is safe to assume that the offset field of a memory reference to
a data snippet always fits in 32-bit signed integer range. Hence the
extra register for materialization is not necessary.

### DIFF
--- a/compiler/x/amd64/codegen/OMRMemoryReference.cpp
+++ b/compiler/x/amd64/codegen/OMRMemoryReference.cpp
@@ -142,26 +142,20 @@ void OMR::X86::AMD64::MemoryReference::finishInitialization(
    // Figure out whether we need to allocate a register for the address
    //
    bool mightNeedAddressRegister;
-   if (!self()->getBaseRegister() &&
-       !self()->getIndexRegister() &&
-       (cg->needRelocationsForStatics() ||
-        cg->needClassAndMethodPointerRelocations() ||
-        cg->needRelocationsForBodyInfoData() ||
-        cg->needRelocationsForPersistentInfoData()))
-      {
-      mightNeedAddressRegister = true;
-      }
-   else if (sr.getSymbol() != NULL && sr.isUnresolved())
-      {
-      // Once resolved, the address could be anything, so be conservative.
-      //
-      mightNeedAddressRegister = true;
-      }
-   else if (self()->getDataSnippet())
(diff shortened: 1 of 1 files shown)


compiler/x/amd64/codegen/OMRMemoryReference.cpp
```diff
@@ -142,27 +142,20 @@ void OMR::X86::AMD64::MemoryReference::finishInitialization(
...
-      mightNeedAddressRegister = true;
-      }
-   else if (self()->getDataSnippet())
+   if (self()->getDataSnippet())
       {
       // Assume snippets are in RIP range
       //
       mightNeedAddressRegister = false;
       }
+   else if (!self()->getBaseRegister() && !self()->getIndexRegister() &&
```
andrewcraik: can we put the getIndexRegister on its own line since it will make reading this a lot easier

compiler/x/amd64/codegen/OMRMemoryReference.cpp
```diff
@@ -171,27 +164,24 @@ void OMR::X86::AMD64::MemoryReference::finishInitialization(
...
       }
-   else if (comp->getOption(TR_EnableHCR) && sr.getSymbol() && sr.getSymbol()->isClassObject())
+   else if (sr.getSymbol() != NULL && sr.isUnresolved())
       {
-      // Can't do any displacement-based tests because the displacement can change
+      // Once resolved, the address could be anything, so be conservative.
       //
       mightNeedAddressRegister = true;
       }
-   else if (self()->getBaseRegister() == NULL)
```
andrewcraik: it isn't clear to me why this case was eliminated? same for the getIndexRegister case?

0dvictor: The information whether the base or index register is NULL is not related to whether address register may be needed. Therefore querying if either is NULL is useless.

compiler/x/amd64/codegen/OMRMemoryReference.cpp
0dvictor: Sure, it has been updated.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
All checks have passed other than zos. This is an x86 specific change and so it should not hold up this merge.

PR #3126, opened 2018-10-26
Improve ArrayCopy Sequence Part on IBM Z

r30shah:
This commit contains following changes.
1. Instead of expecting diamond for array copy direction from
optimizer generates a runtime test for a given single array copy
tree top.
2. Vectorize backward array copy sequence to speed up

### DIFF
--- a/compiler/codegen/ScratchRegisterManager.cpp
+++ b/compiler/codegen/ScratchRegisterManager.cpp
@@ -78,7 +78,6 @@ bool TR_ScratchRegisterManager::reclaimScratchRegister(TR::Register *reg)
 
 TR::Register *TR_ScratchRegisterManager::findOrCreateScratchRegister(TR_RegisterKinds rk)
    {
-   TR_ASSERT(rk != TR_VRF,"VRF: RA (findOrCreateScratchRegister) unimplemented");
    // Check for free registers in the list.  If there are none then create
    // one if there is enough room.
    //
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -666,8 +666,6 @@ OMR::Z::CodeGenerator::CodeGenerator()
    self()->setSupportsBCDToDFPReduction();
    self()->setSupportsIntDFPConversions();
 
-   self()->setSupportsPostProcessArrayCopy();
-
    if (_processorInfo.supportsArch(TR_S390ProcessorInfo::TR_z10))
       {
       self()->setSupportsTranslateAndTestCharString();
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -13429,6 +13429,161 @@ OMR::Z::TreeEvaluator::generateMemToMemElementCopy(TR::Node *node, TR::CodeGener
 #undef iComment
    }
 
+void 
+OMR::Z::TreeEvaluator::genLoopForwardArrayCopy(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *loopIterReg, bool isConstantLength)
+   {
+   TR::LabelSymbol *processingLoop = generateLabelSymbol(cg);
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, processingLoop);
+   generateSS1Instruction(cg, TR::InstOpCode::MVC, node, 255,
+      generateS390MemoryReference(byteDstReg, 0, cg),
+      generateS390MemoryReference(byteSrcReg, 0, cg)); 
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteSrcReg, generateS390MemoryReference(byteSrcReg, 256, cg));
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteDstReg, generateS390MemoryReference(byteDstReg, 256, cg));
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRCT, node, loopIterReg, processingLoop);
+   // If it is known at compile time that this is forward array copy and also length is known at compile time then internal control flow starts here.
+   if (node->isForwardArrayCopy() && isConstantLength)
+      processingLoop->setStartInternalControlFlow();
+   }
+
+
+void
+OMR::Z::TreeEvaluator::forwardArrayCopySequenceGenerator(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR::Node *byteLenNode, TR_S390ScratchRegisterManager *srm, TR::LabelSymbol *mergeLabel)
+   {
+#ifdef J9_PROJECT_SPECIFIC
+   bool mustGenerateOOLGuardedLoadPath = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() &&
+                                         node->getArrayCopyElementType() == TR::Address;
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      // It might be possible that we have constant byte lenght load and it is forward array copy. 
+      // In this case if we need to do guarded Load then need to evaluate byteLenNode.
+      if (byteLenReg == NULL)
+         byteLenReg = cg->gprClobberEvaluate(byteLenNode);
+      TR::TreeEvaluator::genGuardedLoadOOL(node, cg, byteSrcReg, byteDstReg, byteLenReg, mergeLabel, srm, true);
+      }
+#endif
+   if (byteLenNode->getOpCode().isLoadConst())
+      {
+      // TODO: IMPLEMENT MVCL VERSION OF FORWARD ARRAY COPY
+      // If the length is at the THRESHOLD=77825 i.e., 4096*19+1, MVCL becomes a better choice compared to MVC in loop.
+      // In general, MVCL shows performance gain over LOOP with MVCs when the length is
+      // within [4K*i+1, 4K*i+4089], i=19,20,...,4095.
+      // Notice that within the small range [4K*i-7, 4K*i], MVCL is significantly degraded. (3 times slower than normal detected)
+      // In order to use MVCL, we have also to make sure that the use is safe. i.e., src and dst are NOT aliased.
+      int64_t byteLen = byteLenNode->getConst<int64_t>();
+      if (byteLen > 256)
+         {
+         TR::Register *loopIterReg = srm->findOrCreateScratchRegister();
+         int64_t loopIter = byteLenNode->getConst<int64_t>() >> 8;
+         genLoadLongConstant(cg, node, loopIter, loopIterReg);
+         TR::TreeEvaluator::genLoopForwardArrayCopy(node, cg, byteSrcReg, byteDstReg, loopIterReg, true);
+         srm->reclaimScratchRegister(loopIterReg);
+         }
+      // In following cases following MVC instruction should be generated. 
+      // 1. byteLength <= 256
+      // 2. byteLength > 256 && byteLength is not multiple of 256
+      uint8_t residueLength = static_cast<uint8_t>(byteLen & 0xFF);
+      if (byteLen <= 256 || residueLength != 0)
+         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, residueLength-1, 
+            generateS390MemoryReference(byteDstReg, 0, cg),
+            generateS390MemoryReference(byteSrcReg, 0, cg)); 
+      }
+   else
+      {
+      TR::Register *loopIterReg = srm->findOrCreateScratchRegister();
+      generateRIInstruction(cg, TR::InstOpCode::getAddHalfWordImmOpCode(), node, byteLenReg, -1);
+      generateShiftRightImmediate(cg, node, loopIterReg, byteLenReg, 8);
+      TR::LabelSymbol *exrlInstructionLabel = generateLabelSymbol(cg);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BZ, node, exrlInstructionLabel);
+      TR::TreeEvaluator::genLoopForwardArrayCopy(node, cg, byteSrcReg, byteDstReg, loopIterReg, false);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, exrlInstructionLabel);
+      TR::LabelSymbol *exrlTargetLabel = generateLabelSymbol(cg);
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, exrlTargetLabel);
+      TR::Instruction *residueMVC = generateSS1Instruction(cg, TR::InstOpCode::MVC, node, 0, 
+                                       generateS390MemoryReference(byteDstReg, 0, cg),
+                                       generateS390MemoryReference(byteSrcReg, 0, cg)); 
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, exrlInstructionLabel);
+      if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10) && !cg->comp()->getOption(TR_DisableInlineEXTarget))
+         {
+         generateRILInstruction(cg, TR::InstOpCode::EXRL, node, byteLenReg, exrlTargetLabel);
+         }
+      else
+         {
+         TR::Register *regForResidueMVCTarget = srm->findOrCreateScratchRegister();
+         generateEXDispatch(node, cg, byteLenReg, regForResidueMVCTarget, residueMVC);
+         srm->reclaimScratchRegister(regForResidueMVCTarget);
+         }
+      srm->reclaimScratchRegister(loopIterReg);
+      }
+   }
+
+TR::RegisterDependencyConditions *
+OMR::Z::TreeEvaluator::backwardArrayCopySequenceGenerator(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR::Node *byteLenNode, TR_S390ScratchRegisterManager *srm, TR::LabelSymbol *mergeLabel)
+   {
+#ifdef J9_PROJECT_SPECIFIC
+   bool mustGenerateOOLGuardedLoadPath = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() &&
+                                         node->getArrayCopyElementType() == TR::Address;
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      TR::TreeEvaluator::genGuardedLoadOOL(node, cg, byteSrcReg, byteDstReg, byteLenReg, mergeLabel, srm, false);
+      }
+#endif
+   TR_Debug *debug = cg->getDebug();
+#define iComment(str) if (debug) debug->addInstructionComment(cursor, (str));
+   TR::Instruction *cursor = NULL;
+   TR::RegisterDependencyConditions *deps = NULL;
+   bool genStartICFLabel = node->isBackwardArrayCopy() && byteLenNode->getOpCode().isLoadConst();
+   if (!cg->getSupportsVectorRegisters())
+      {
+      deps = TR::TreeEvaluator::generateMemToMemElementCopy(node, cg, byteSrcReg, byteDstReg, byteLenReg, srm, false, false, genStartICFLabel);
+      }
+   else
+      {
+      TR::Register *loopIterReg = srm->findOrCreateScratchRegister();
+      cursor = generateShiftRightImmediate(cg, node, loopIterReg, byteLenReg, 4); 
+      iComment("Calculate loop iterations");
+         
+      TR::LabelSymbol *handleResidueLabel = generateLabelSymbol(cg);
+      if (genStartICFLabel)
+         {
+         TR::LabelSymbol *cFlowRegionStart = generateLabelSymbol(cg);
+         generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, cFlowRegionStart);
+         cFlowRegionStart->setStartInternalControlFlow();
+         }
+      cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BZ, node, handleResidueLabel); 
+      iComment("if length <= 16, jump to handling residue");
+      
+      TR::LabelSymbol *loopLabel = generateLabelSymbol(cg);
+      cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, loopLabel); 
+      iComment("Vector Load and Strore Loop");
+      cursor = generateRIInstruction(cg, TR::InstOpCode::getAddHalfWordImmOpCode(), node, byteLenReg, -16);
+      
+      TR::Register *vectorBuffer = srm->findOrCreateScratchRegister(TR_VRF);
+      generateVRXInstruction(cg, TR::InstOpCode::VL , node, vectorBuffer, generateS390MemoryReference(byteSrcReg, byteLenReg, 0, cg));
+      generateVRXInstruction(cg, TR::InstOpCode::VST, node, vectorBuffer, generateS390MemoryReference(byteDstReg, byteLenReg, 0, cg), 0);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRCT, node, loopIterReg, loopLabel);
+      
+      cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, handleResidueLabel); 
+      iComment("Handle residue");
+      
+      cursor = generateRIInstruction(cg, TR::InstOpCode::getAddHalfWordImmOpCode(), node, byteLenReg, -1); 
+      iComment("Adjust remaining bytes for vectore load/store length instruction");
+      TR::LabelSymbol *skipResidueLabel = generateLabelSymbol(cg);
+    /** TODO: Exploit constant length in vectorized implementation
+      *  For non vectorized version, we are going to copy bytes as per data type so there is no advantage of exploiting constant length unless we unroll the loop
+      *  Current vectorized implementation for backward array copy uses byteLen as index to copy bytes from source to destination in backward direction.
+      *  So even if length is known at compile time we can use it to avoid generating residue control logic.
+      */
+      generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), node, byteLenReg, 0, TR::InstOpCode::COND_BL, skipResidueLabel, false);
+      generateVRSbInstruction(cg, TR::InstOpCode::VLL , node, vectorBuffer, byteLenReg, generateS390MemoryReference(byteSrcReg, 0, cg));
+      generateVRSbInstruction(cg, TR::InstOpCode::VSTL, node, vectorBuffer, byteLenReg, generateS390MemoryReference(byteDstReg, 0, cg), 0);
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, skipResidueLabel);
+      srm->reclaimScratchRegister(loopIterReg);
+      srm->reclaimScratchRegister(vectorBuffer);
+      }
+   return deps;
+   #undef iComment
+   }
+
 
 TR::Register*
 OMR::Z::TreeEvaluator::arraycopyEvaluator(TR::Node* node, TR::CodeGenerator* cg)
@@ -13462,9 +13617,9 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
    TR::Register* byteSrcReg = NULL;
    TR::Register* byteDstReg = NULL;
    TR::Register* byteLenReg = NULL;
-
+   bool isConstantByteLen = byteLenNode->getOpCode().isLoadConst();
    // Check for array copy of zero bytes or negative number of bytes
-   if (byteLenNode->getOpCode().isLoadConst() && byteLenNode->getConst<int64_t>() <= 0)
+   if (isConstantByteLen && byteLenNode->getConst<int64_t>() <= 0)
       {
       cg->evaluate(byteSrcNode);
       cg->evaluate(byteDstNode);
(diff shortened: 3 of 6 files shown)


compiler/codegen/ScratchRegisterManager.cpp
```diff
@@ -78,7 +78,6 @@ bool TR_ScratchRegisterManager::reclaimScratchRegister(TR::Register *reg)
 
 TR::Register *TR_ScratchRegisterManager::findOrCreateScratchRegister(TR_RegisterKinds rk)
    {
-   TR_ASSERT(rk != TR_VRF,"VRF: RA (findOrCreateScratchRegister) unimplemented");
```
r30shah: Main aim for Scratch Register is to provide a limited pool of scratch registers that can be used by evaluators without worrying about keeping track of them to add into register dependecy list or mark them dead. I did not seen in code that it needs to do anything special for vector registers. If it has requested a TR::Register of TR_VRF it needs to allocate register of that kind. Which should be handled here (https://github.com/eclipse/omr/blob/7694d089de6060a33d7b1a08fd092f7c40bf660e/compiler/codegen/ScratchRegisterManager.cpp#L105)
@fjeremic  I checked other utilities within scratch register manager and did not find any special attention required for vector registers. Do you know any other corner case because of which we have an assert here?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,268 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+#ifdef J9_PROJECT_SPECIFIC
+   bool mustGenerateOOLGuardedLoadPath = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() &&
+                                         node->getArrayCopyElementType() == TR::Address;
+   if (mustGenerateOOLGuardedLoadPath)
+      {
+      // It might be possible that we have constant byte lenght load and it is forward array copy. 
+      // In this case if we need to do guarded Load then need to evaluate byteLenNode.
+      if (byteLenReg == NULL)
+         byteLenReg = cg->gprClobberEvaluate(byteLenNode);
+      TR::TreeEvaluator::genGuardedLoadOOL(node, cg, byteSrcReg, byteDstReg, byteLenReg, mergeLabel, srm, true);
```
r30shah: @fjeremic  This call here is implemented in https://github.com/eclipse/openj9/pull/3465. This is something on what I want your view, I moved the implementation as it handles special case for Concurrent Scavenge GC which really belongs to OpenJ9.

compiler/codegen/ScratchRegisterManager.cpp
fjeremic: I don't recall any restriction or why it is there. Does it just work™ if you enable it?

r30shah: Yes it works fine. 
I removed that  and used the scratch register manager to allocate vector register. Reason I needed to go through the scratch register manager is that there are so many parts for array copy and if we are generating tests for direction in codegen, we are left it with an Big ICF which includes both forward array copy and backward array copy. Scratch register Manager becomes handy in this case as generating dependencies for merge point we can use SRM utility function to add all the scratch registers allocated through it to dependency condition.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
-            {
-            byteLenReg = cg->evaluateLengthMinusOneForMemoryOps(byteLenNode, true, alreadyEvaluatedLengthMinusOne);
-            }
-
-         MemCpyVarLenMacroOp op(node, byteDstNode, byteSrcNode, cg, byteLenReg, byteLenNode, alreadyEvaluatedLengthMinusOne);
-
-         op.setUseEXForRemainder(true);
-         op.generate(byteDstReg, byteSrcReg);
-         }
+      // Optimizer has already identified that this is array copy in forward direction.
```
fjeremic: I don't think this comment is true or useful for that matter. It could have very well been the IL generator from whichever downstream project is consuming OMR which sets this flag. I think we should remove the comment as it serves no purpose explaining _why_ the code is the way it is. i.e. it doesn't give any more information than the `if (node->isForwardArrayCopy())` line proceeding it.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
-      // Switch back from OOL Slow Path
-      outOfLineCodeSection->swapInstructionListsWithCompilation();
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, startICFLabel);
+      startICFLabel->setStartInternalControlFlow();
+      cursor = generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpOpCode(), node, byteLenReg, static_cast<int8_t>(0x0), TR::InstOpCode::COND_BNH, mergeLabel, false);
+      iComment("byteLen <= 0 goto mergeLabel");
       }
-#endif
 
    if (node->isForwardArrayCopy())
```
fjeremic: Don't we also have `isBackwardArrayCopy()`? Are we able to make use of this to eliminate the runtime check?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
    else
       {
-      // Don't clobber evaluate byteLenNode again if it was already evaluated
-      if (!alreadyEvaluatedByteLenNode)
+      // We need to decide direction of array copy at runtime.
+      TR::LabelSymbol *forwardArrayCopyLabel = generateLabelSymbol(cg);
+      cursor = generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpRegOpCode(), node, byteSrcReg, byteDstReg, TR::InstOpCode::COND_BNL, forwardArrayCopyLabel, false);
+      iComment("if byteSrcPointer >= byteDstPointer then GoTo forwardArrayCopy");
+      
+      // If it is contant length bytes array copy then Internal Control Flow starts here.
```
fjeremic: This comment is not useful as it is implied by the following 3 lines.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
-      TR::Register* strideRegister = cg->allocateRegister();
-
-      MemCpyVarLenTypedMacroOp vtOp(node, byteDstNode, byteSrcNode, cg, node->getArrayCopyElementType(), byteLenReg, byteLenNode, false, node->isForwardArrayCopy());
-
-      vtOp.generate(byteDstReg, byteSrcReg, strideRegister);
-      vtOp.cleanUpReg();
-
-      cg->stopUsingRegister(strideRegister);
+      TR::Register *checkBoundReg = srm->findOrCreateScratchRegister();
+      if (byteLenReg == NULL)
```
fjeremic: Braces needed here and for the other `byteLenReg == NULL` check below.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
+      // Optimizer has already identified that this is array copy in forward direction.
+      TR::TreeEvaluator::forwardArrayCopySequenceGenerator(node, cg, byteSrcReg, byteDstReg, byteLenReg, byteLenNode, srm, mergeLabel);
       }
    else
       {
-      // Don't clobber evaluate byteLenNode again if it was already evaluated
-      if (!alreadyEvaluatedByteLenNode)
+      // We need to decide direction of array copy at runtime.
+      TR::LabelSymbol *forwardArrayCopyLabel = generateLabelSymbol(cg);
+      cursor = generateS390CompareAndBranchInstruction(cg, TR::InstOpCode::getCmpRegOpCode(), node, byteSrcReg, byteDstReg, TR::InstOpCode::COND_BNL, forwardArrayCopyLabel, false);
```
fjeremic: See PoPs 7-44 where `BPP` instruction is documented. Check the Programming Notes about assumed branch-prediction logic. I'm not sure what instruction ends up being generated here and for the branch below but as I see the most common scenario will be a forward array copy so one of the branches you generate will always (more or less) be taken and one will always fail. Let's not introduce unwanted branch mispredictions and make sure the branches we're generating here are the optimal ones.

i.e. my question is should this be a `TR::InstOpCode::COND_BL` branch?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13577,92 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
+      
+      cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, forwardArrayCopyLabel);iComment("Forward Array Copy Sequence");
+      TR::TreeEvaluator::forwardArrayCopySequenceGenerator(node, cg, byteSrcReg, byteDstReg, byteLenReg, byteLenNode, srm, mergeLabel);
+      cursor = generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, mergeLabel); iComment("Jump to MergeLabel");
+      
+      cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, backwardArrayCopyLabel); iComment("Backward Array Copy Sequence");
+      deps = TR::TreeEvaluator::backwardArrayCopySequenceGenerator(node, cg, byteSrcReg, byteDstReg, byteLenReg, srm, mergeLabel);
+      }
+   cursor = generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, mergeLabel);
+   if (!(isConstantByteLen && node->isForwardArrayCopy() && byteLenNode->getConst<int64_t>() < 256))
```
fjeremic: I'm not really following what we're doing here exactly. Can you explain? This code seems out of place.

fjeremic:
I'd also like to see this PR referenced by the BumbleBench PR for the arraycopy benchmark if you've made changes there or if not what benchmark was used to generate your results?

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -679,8 +679,6 @@ OMR::Z::CodeGenerator::CodeGenerator()
    self()->setSupportsBCDToDFPReduction();
    self()->setSupportsIntDFPConversions();
 
-   self()->setSupportsPostProcessArrayCopy();
```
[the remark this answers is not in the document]
r30shah: Yes that is `todo`. VP code needs to be cleaned up once we have this in as there is no need to generate diamond in optimizer. Let me open up an issue in OMR to do that.

compiler/z/codegen/OMRTreeEvaluator.cpp
r30shah: So in case we have a constant length array copy in forward direction with length  within MVC instruction range, we will not have an ICF. This condition avoids putting generating endICF label if we do not have ICF

fjeremic: Is there any downside to just putting it in there in terms of the actual emitter binary? I would prefer to keep the code clean at the expense of 2 labels being generated.

r30shah: Code inside this if do not generate a label. It sets end internal control flow to merge label and also creates dependencies for ICF. Also setting EndInternalControl Flow without start of internal control flow will hit an assert.  And main reason I am doing all this in if is to avoid generating dependencies in case there is no ICF.

compiler/z/codegen/OMRTreeEvaluator.cpp
r30shah: The sequence favors executing arrayCopy Sequence. I generated here are like this.
```
if (srcPointer >= dstPointer)
   goto forwardSequenceLabel
else if (srcPointer+byteLength > dstPointer)
   goto backwardSequenceLabel

forwardSequenceLabel: 
...
Jump mergeLabel
backwardSequenceLabel:
...
mergeLabel:
```
So in summary, it first checks if srcPointer is greater or equal to dstPointer and if that is case then it can do arrayCopy in forward direction even if there is an overlap and that is what I check first and in this case I won't even need to know if overlap exists or not.
If srcPointer is less than dstPointer then I need to check if there is overlap and in case of overlap does arraycopy in backward direction.

r30shah:
@fjeremic To concentrate runs to only ArrayCopy I used BumbleBench and added a benchmark (https://github.com/r30shah/bumblebench/commit/130f0aad07374a5afa2d3301a0c225836d2da0aa). 
I can open up PR to add this benchmark to BumbleBench. (I only have one concern, this is too trivial benchmark to test isolate these arraycopy changes. Not sure if I should commit them) . I am pasting row results from my runs here.
screen shot 2018-11-02 at 3 10 07 pm

compiler/z/codegen/OMRCodeGenerator.cpp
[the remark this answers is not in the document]
fjeremic: Let's open up an item and link it here.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: I see two calls to `genGuardedLoadOOL`, one here and one in the backwards evaluator. I also see we call both of them in the generic path where we don't know if we are forwards or backwards. Won't this generate two OOL sequences for the fallback CS path when we only really need one?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,276 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+   TR::LabelSymbol *processingLoop = generateLabelSymbol(cg);
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, processingLoop);
+   generateSS1Instruction(cg, TR::InstOpCode::MVC, node, 255, 
+      new (cg->trHeapMemory()) TR::MemoryReference(byteDstReg, 0, cg), 
+      new (cg->trHeapMemory()) TR::MemoryReference(byteSrcReg, 0, cg));
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteSrcReg, new (cg->trHeapMemory()) TR::MemoryReference(byteSrcReg, 256, cg));
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteDstReg, new (cg->trHeapMemory()) TR::MemoryReference(byteDstReg, 256, cg));
+   generateS390BranchInstruction(cg, TR::InstOpCode::BRCT, node, loopIterReg, processingLoop);
+   // If it is known at compile time that this is forward array copy and also length is known at compile time then internal control flow starts here.
+   if (node->isForwardArrayCopy() && isConstantLength)
```
fjeremic: Aren't we always a forward arraycopy in here, so why do we need to check that `node->isForwardArrayCopy()` is true?

r30shah: We are always will be generating forward array copy sequence here that is true. But ICF starts here only where we have constant byte length and node is known to be doing array copy in forward direction. If direction is not known and byteLength is constant, arraycopy ICF starts early when we check out overlapping region for array copy

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,276 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+      generateShiftRightImmediate(cg, node, loopIterReg, byteLenReg, static_cast<int32_t>(8));
+      TR::LabelSymbol *exrlInstructionLabel = generateLabelSymbol(cg);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BZ, node, exrlInstructionLabel);
+      TR::TreeEvaluator::genLoopForwardArrayCopy(node, cg, byteSrcReg, byteDstReg, loopIterReg, false);
+      generateS390BranchInstruction(cg, TR::InstOpCode::BRC, TR::InstOpCode::COND_BRC, node, exrlInstructionLabel);
+      TR::LabelSymbol *exrlTargetLabel = generateLabelSymbol(cg);
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, exrlTargetLabel);
+      generateSS1Instruction(cg, TR::InstOpCode::MVC, node, 0, 
+         new (cg->trHeapMemory()) TR::MemoryReference(byteDstReg, 0, cg), 
+         new (cg->trHeapMemory()) TR::MemoryReference(byteSrcReg, 0, cg));
```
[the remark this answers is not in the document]
r30shah: Changed in https://github.com/eclipse/omr/pull/3126/commits/3d0f4b7a6e1f7d0f3fa0b4ec92cf94d314a28974

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,276 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+      generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, exrlInstructionLabel);
+      generateRILInstruction(cg, TR::InstOpCode::EXRL, node, byteLenReg, exrlTargetLabel);
```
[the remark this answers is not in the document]
r30shah: Good catch. I added a check to see if it is z10+ then generating EXRL other wise use generateEXDispatch to generate EX Instruction
Change in https://github.com/eclipse/omr/pull/3126/commits/3d0f4b7a6e1f7d0f3fa0b4ec92cf94d314a28974

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13584,97 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
 
    cg->decReferenceCount(byteDstNode);
    cg->decReferenceCount(byteSrcNode);
    cg->decReferenceCount(byteLenNode);
 
-   cg->stopUsingRegister(byteSrcReg);
-   cg->stopUsingRegister(byteDstReg);
-   cg->stopUsingRegister(byteLenReg);
+   // All registers used in this evaluator is added to the Register Dependency Condition deps.
+   // Helper function stopUsingDepRegs can be used mark registers dead.
```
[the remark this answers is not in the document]
r30shah: Changes in https://github.com/eclipse/omr/pull/3126/commits/3d0f4b7a6e1f7d0f3fa0b4ec92cf94d314a28974

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13314,113 +13584,97 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
+   if (deps != NULL)
+      {
+      deps->stopUsingDepRegs(cg);
+      }
+   else
+      {
+      cg->stopUsingRegister(byteSrcReg);
+      cg->stopUsingRegister(byteDstReg);
+      if (byteLenNode != NULL)
+         cg->stopUsingRegister(byteLenReg);
```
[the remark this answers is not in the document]
r30shah: Removed if. https://github.com/eclipse/omr/pull/3126/commits/3d0f4b7a6e1f7d0f3fa0b4ec92cf94d314a28974

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,276 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+      // In this case if we need to do guarded Load then need to evaluate byteLenNode.
+      if (byteLenReg == NULL)
+         byteLenReg = cg->gprClobberEvaluate(byteLenNode);
+      TR::TreeEvaluator::genGuardedLoadOOL(node, cg, byteSrcReg, byteDstReg, byteLenReg, mergeLabel, srm, true);
+      }
+#endif
+   if (byteLenNode->getOpCode().isLoadConst())
+      {
+      // TODO: IMPLEMENT MVCL VERSION OF FORWARD ARRAY COPY
+      // If the length is at the THRESHOLD=77825 i.e., 4096*19+1, MVCL becomes a better choice compared to MVC in loop.
```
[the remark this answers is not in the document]
r30shah: Yes, We can find comments regarding to previous experiments here : https://github.com/eclipse/omr/blob/e7406af2360b1cecbf1018111685524224e28fde/compiler/z/codegen/OpMemToMem.cpp#L338

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,301 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+         genLoadLongConstant(cg, node, loopIter, loopIterReg);
+         TR::TreeEvaluator::genLoopForwardArrayCopy(node, cg, byteSrcReg, byteDstReg, loopIterReg, true);
+         srm->reclaimScratchRegister(loopIterReg);
+         }
+      // In following cases following MVC instruction should be generated. 
+      // 1. byteLength <= 256
+      // 2. byteLength > 256 && byteLength is not multiple of 256
+      uint8_t residueLength = static_cast<uint8_t>(byteLen & 0xFF);
+      if (byteLen <= 256 || residueLength != 0)
+         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, static_cast<uint8_t>(residueLength-1),
```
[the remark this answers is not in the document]
r30shah: Removed unnecessary use of `static_cast` in https://github.com/eclipse/omr/pull/3126/commits/5b932b12ff5d03944b3da0ecad3a43d2d901f6c1

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13266,6 +13266,301 @@ OMR::Z::TreeEvaluator::arraysetEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+void 
+OMR::Z::TreeEvaluator::genLoopForwardArrayCopy(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *loopIterReg, bool isConstantLength)
+   {
+   TR::LabelSymbol *processingLoop = generateLabelSymbol(cg);
+   generateS390LabelInstruction(cg, TR::InstOpCode::LABEL, node, processingLoop);
+   generateSS1Instruction(cg, TR::InstOpCode::MVC, node, 255,
+      generateS390MemoryReference(byteDstReg, 0, cg),
+      generateS390MemoryReference(byteSrcReg, 0, cg)); 
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteSrcReg, new (cg->trHeapMemory()) TR::MemoryReference(byteSrcReg, 256, cg));
+   generateRXInstruction(cg, TR::InstOpCode::LA, node, byteDstReg, new (cg->trHeapMemory()) TR::MemoryReference(byteDstReg, 256, cg));
```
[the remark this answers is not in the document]
r30shah: Changed in https://github.com/eclipse/omr/pull/3126/commits/5b932b12ff5d03944b3da0ecad3a43d2d901f6c1 
There still some places where I needed to use Constructor to generate new dependency adding old deps. Same as https://github.com/eclipse/openj9/pull/3465#discussion_r231923959 , we need to add new functions to RegisterDependency

r30shah:
Created a separate PR for helper functions from OMR needed by this change. https://github.com/eclipse/omr/pull/3167 We can merge this safely once OMR part is merged.

r30shah:
FYI @dmitripivkine , @DanHeidinga  
These is the Pull Request vectorizing backward array copy sequence on Z. Instructions used in vectorizing sequence is block concurrent. Here block size could be halfword, word, double word or quadword depending on the way storage operand is referenced. I am loading quadword at time so there should be no issue regarding to atomicity. 
To test this out, I picked up logic from MultithreadedArrayCopyTest and created a unit test in which multiple threads continuously rotates array (Right) forcing it to follow backward array copy sequence and ran it with these change. Test passes. 
Here is the test case I wrote : https://github.com/r30shah/openj9/commit/dc74ef29bcd6f8686b3a1f0d962bd0e62c5bc8c9 Checking out comments and everything and will create PR to add this to openj9 test. 
Let me know if you have any concerns.

r30shah:
PR to add bumble bench mentioned in https://github.com/eclipse/omr/pull/3126#issuecomment-435479866 : 
https://github.com/AdoptOpenJDK/bumblebench/pull/11

r30shah:
Created a PR for Multi Threaded array copy in backward direction mentioned in https://github.com/eclipse/omr/pull/3126#issuecomment-437195277 : 
https://github.com/eclipse/openj9/pull/3623

charliegracie:
I am very nervous of these changes. About every 1.5-2 years someone tries to improve array copy and those changes always include using instructions which move less than Object word size or not a multiple of Object word size at a time. These types of copies are not valid for object array copies since threads may be concurrently performing array copy operations and reading the array values. If a thread is performing an array copy on a reference array (an array of objects) the smallest amount you can copy in full 64 bit is 64 bits at a time, in compressed references it is 32 bits at a time and in 32 bit it is 32 bits at a time. Also those copies have to be aligned to the same boundaries. IF that is not the case application threads and concurrent GC threads may crash because they will not read proper references out of the array. 

Can you explain in thorough detail how the forwardArrayCopy and backwardArrayCopy changes you are proposing ensure that these criteria I mentioned above are met?

r30shah:
@charliegracie I would like to clarify couple of things on this PR. 
Let me start with backward array copy sequence which is main improvement here, as instructions uses vector instructions which are block concurrent where block size could be halfword, word, double word. I am loading quadword at time (for non compressedref objects - 2 at time, for compressed refs objects - 4 at a time) and block concurrent means accesses to all bytes within blocks appears to be block concurrent as observed by other CPUs and channel programs.  This means when one thread is accessing - Fetching/Storing bytes within blocks for an example if one thread is fetching bytes within block, no other CPU or channel program would be able to store within that block. Similarly if while storing bytes within blocks, no other CPU or channel program will be allowed to fetch from the blocks.  Also load and store in this sequence is in multiple of Object reference sizes. 
  
For forward ArrayCopy sequence. I did not modify any sequence, just moved it from https://github.com/eclipse/omr/blob/e706e5cc9533ad8da26449f7e70a51adc2bced71/compiler/z/codegen/OpMemToMem.cpp#L508
The main reason was. I wanted to get rid of Diamond generated by optimizer and move logic to decide array copy direction to codegen. That API was not that extensible to handle Internal Control Flow I am introducing.

charliegracie:
@fjeremic I completely disagree with your [comment](https://github.com/eclipse/omr/pull/3126#issuecomment-437942709).  This is 100% not a testing issue but rather an issue where someone can provable explain to me via doc that these instructions are supposed to copy data in a way that works for the Java memory model. I have been on the receiving end of debugging these issues over the years and I am just trying to prevent others from suffering the same pain. 

I am all for innovation and if these changes are correct I am very excited for the performance wins. @r30shah is there a doc you can link to that shows your claims above for the backward array copy? I want to ensure there are no sneaky alignment concerns. For the forward array copy why are you not using the same vector instructions? I see that you are moving the code directly from OpenJ9 for forward array copy. Reading my concerns above do you believe the forward array copy is correct? I am not concerned if it passes tests I am concerned whether or not those instructions make the guarantees that I expressed above.

fjeremic:
> This is 100% not a testing issue but rather an issue where someone can provably explain to me via doc that these instructions are supposed to copy data in a way that works for the Java memory model.

And how would you determine this not to be the case? You would write a test with exercises the code and makes it fail. Or in a TDD approach you write the tests first.

> I have been on the receiving end of debugging these issues over the years and I am just trying to prevent others from suffering the same pain.

I agree, such bugs would be a real pain to debug. I'm all for more documentation and explanation but at the end of the day code in the codebase must be exercised via automated testing, not a bunch of people looking at it and signing off. I don't think reviews are enough, especially on our non-trivial codebase.

> For the forward array copy why are you not using the same vector instructions?

Forward arraycopy is achieved with an `MVC` instruction which is the fastest way to copy memory on z/Architecture. When destructive overlap does not exist, as is the case here, `MVC` is block-concurrent on in a left-to-right direction, and all bytes accessed within a double-word appear to be accessed concurrently as observed by other CPUs [1 - Chapter 5.128 _Block-Concurrent References_].

Currently we do not have a "backwards-MVC", however instead of copying references word (or doubleword) at a time we can do multiple of them within a vector register. The vector load and store instructions used to implement the backwards arraycopy are also block-concurrent on the integral boundary of up to 8 bytes. You can find this in the same section 5.128 a few paragraphs below the `MVC` comment in [1].

[1] http://www-01.ibm.com/support/docview.wss?uid=isg2b9de5f05a9d57819852571c500428f9a

r30shah:
I was looking for accessible link to documents. Thanks @fjeremic for pointing it out. In addition to Filip's explanation regarding to Block Concurrent instructions, 
@charliegracie  I think there is slight misunderstanding regarding to forward array copy code. I did not moved code from OpenJ9 to OMR. It was already sitting in omr/compiler/z/codegen/OpMemToMem.cpp but was not extensible to handle both forward and backward case together. To improve on comilation time, we removed burden of generating direction test (find if region is destructive) from Optimizer and moved it to CodeGen. 
Look at this comment in ValuePropagation : https://github.com/eclipse/omr/blob/e706e5cc9533ad8da26449f7e70a51adc2bced71/compiler/optimizer/OMRValuePropagation.cpp#L5706
ArrayCopy Transformation was done only on Z where it generates test for direction in IL while CodeGen on Power and x86 was taking care of it.

charliegracie:
@fjeremic again I disagree with your [comment](https://github.com/eclipse/omr/pull/3126#issuecomment-438006954) as I believe you are misquoting 
 or misunderstanding what I am saying.  I am saying my concerns are with wanting to see documentation for the instructions so that I can verify by reading the documentation that what was explained to me is what is supposed to happen. After myself and other reviewers have verified this, then yes testing should be completed. There is no time that I would suggest against adding new testing. In fact, I feel like I am the one pushing most items in OMR to include more testing and to make sure we do not break downstream projects. 

To be clear.
1. Testing for this type of change is definitely the right approach and I am glad it is being added.
1.1 I would like to see the test modified in the way Dmitri described in a comment on the test.
1.2 I believe picking different locations for the copy to start and selecting different amounts of slots to be copied would make this test even better.
2. Reviewing the technical documentation for the instructions for such a change is also important. I am not claiming anyone is missing something. I just want to be able to verify things myself and make sure that the proper documentation is available for others when they try to read the code to make further changes/ improvements.  Sadly the document you linked requires and IBM ID/Password.. It looks like I can get access but that is a pain for external people trying to make changes in the future.

Thanks @r30shah for the clarification.  I would still like to verify that the MVC instruction is doing the right thing since I can see in the past we specifically changed code that was performing MVC instructions as the writes were not atomic. Maybe that is related to the `block` size being mentioned but I would still like to read just to be sure.

fjeremic:
> To be clear.
> 
>  1. Testing for this type of change is definitely the right approach and I am glad it is being added.
>        1.1 I would like to see the test modified in the way Dmitri described in a comment on the test.
>        1.2 I believe picking different locations for the copy to start and selecting different amounts of slots to be copied would make this test even better.
> 
>  2. Reviewing the technical documentation for the instructions for such a change is also important. I am not claiming anyone is missing something. I just want to be able to verify things myself and make sure that the proper documentation is available for others when they try to read the code to make further changes/ improvements.  Sadly the document you linked requires and IBM ID/Password.. It looks like I can get access but that is a pain for external people trying to make changes in the future.

Thanks for the clarification @charliegracie. I fully understand now and I agree with all of the above! I'm sorry for the misunderstanding and the resulting churn. @r30shah, some great points are made above which we should act on. Please let me know if you need help.

> 2\. Sadly the document you linked requires and IBM ID/Password.. It looks like I can get access but that is a pain for external people trying to make changes in the future.

You're right! Thanks for catching this. Here is a direct link for current and future readers:
http://publibfi.boulder.ibm.com/epubs/pdf/dz9zr011.pdf

fjeremic:
> Thanks @r30shah for the clarification. I would still like to verify that the MVC instruction is doing the right thing since I can see in the past we specifically changed code that was performing MVC instructions as the writes were not atomic. Maybe that is related to the `block` size being mentioned but I would still like to read just to be sure.

I'd like to request a review from @joransiu at this point as well since he has historical context on such bugs in the past. This was before I started working on the project so he would likely be the next best person to review.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13478,113 +13633,99 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
+      }
+   else if (node->isBackwardArrayCopy())
+      {
+      deps = TR::TreeEvaluator::backwardArrayCopySequenceGenerator(node, cg, byteSrcReg, byteDstReg, byteLenReg, byteLenNode, srm, mergeLabel);
       }
    else
       {
-      // Don't clobber evaluate byteLenNode again if it was already evaluated
-      if (!alreadyEvaluatedByteLenNode)
+      // We need to decide direction of array copy at runtime.
```
r30shah: @charliegracie  This is the part where it checks for the destructive region.

r30shah:
Removing WIP tag as this change can be merged if all reviews are done.

fjeremic:
@genie-omr build zlinux,zos

Issue #3047, opened 2018-10-10
Remove usages of LA instruction when loading constants in Z codegen


dhong44:
In many cases such as [1], a LA instruction is used to load a constant into a register. This LA instruction is used often in `FPTreeEvaluator.cpp`. Instead a load immediate instruction should be used or when loading 0 into a register, a xor instruction could be used. LA instructions should be used for loading memory address locations, this has caused errors before [2] when enabling 64 bit registers on 32 bit since LA would not properly load 0 into registers.

[1] https://github.com/eclipse/omr/blob/master/compiler/z/codegen/FPTreeEvaluator.cpp#L463
[2] https://github.com/eclipse/omr/pull/2916/files#diff-936032f2a7d7fc08b8410b38ab2d1fc6

PR #3127, opened 2018-10-26
Use XGR to load 0 into a register, and Load Immediate to load a constant instead of LA

muzaffar1331:
This patch replaces TR::InstOpCode::LA with TR::InstOpCode::XOR when
loading 0 into a register. XOR is faster than LA when loading 0 in a
register. See issue #3047.

Closes: #3047
Signed-off-by: Muzaffar Auhammud <muzaffar@cyberstorm.mu>


fjeremic:
Thanks for your contribution @muzaffar1331! @dhong44 could I please ask for a review given you opened the original issue?

fjeremic:
Kicking off a preliminary build:

@genie-omr build zlinux,zos

dhong44:
Thanks for the contribution @muzaffar1331! Just a couple things.

- The also mentions replacing LA instructions when loading constant with load immediate instructions. Load immediate instructions could be used instead when we load a constant value into a register. Sometimes we use a LA instruction to values like 1 or 2 into a register when instead we could use a load immediate instruction. If you'd like you can take care of replacing LA with load immediate instructions as well, there's a couple instances, especially in `FPTreeEvaluator.cpp` where we use LA for constants.
https://github.com/eclipse/omr/blob/8015650e248925792b1630947d4dfd95f51be557/compiler/z/codegen/FPTreeEvaluator.cpp#L1802

- It also looks like there's still a couple times where we use load address instead of xor that you've missed. I'd suggest grepping the codebase for `TR::InstOpCode::LA` and just checking if it's loading a constant or 0. https://github.com/eclipse/omr/blob/8015650e248925792b1630947d4dfd95f51be557/compiler/z/codegen/OMRTreeEvaluator.cpp#L10069 
There's another example of where we do that, I'd suggest just grepping the codebase for `TR::InstOpCode::LA` and checking if each usage can be replaced with xor or load immediate. 

- The code right now is failing to compile because we don't actually have an instruction named XOR. Instead we have XGR and XR to perform exclusive or, XGR is for 64 bit exclusive or and XR is for 32 bit exclusive or. Another issues is that we have to change the type of instruction from RII to RRI, there's an example in [1] where `generateRIIInstruction` was replaced with `generateRRIInstruction`. The error code was:
> FPTreeEvaluator.cpp:2004:30: error: ‘XOR’ is not a member of ‘TR::InstOpCode’



If you search for Z Principles of Operation there are a couple links floating around on the web. That is the documentation for Z/architecture where all the instructions available are described. In appendix B there is a list of the instructions, it also details the type of instruction such as RII and RRI. At the right of that table there is a clickable link to the page which will give more detail about each instruction. The instructions and their properties can also be found in `OMRInstOpCodeEnum.hpp` and `OMRInstOpCodeProperties.hpp`. 
https://github.com/eclipse/omr/blob/8015650e248925792b1630947d4dfd95f51be557/compiler/z/codegen/OMRInstOpCodeEnum.hpp#L519
https://github.com/eclipse/omr/blob/8015650e248925792b1630947d4dfd95f51be557/compiler/z/codegen/OMRInstOpCodeProperties.hpp#L6829-L6841

Here's a blog I found that hashttps://linuxmain.blogspot.com/2012/09/zarchitecture-principles-of-operation.html

muzaffar1331:
Thanks for the review and the useful notes @dhong44 :) I made the changes concerning XGR and using Load Immediate when loading constants into a register.

dhong44:
When I mentioned load immediate instructions I meant any instruction that loads immediate values. There are many more instructions than just LGFI and when loading small constants such as 1 and 2, it would be better to use LHI for 32 bit and LGHI for 64 bit. The difference is that LGFI can load a 32-bit value but LHI and LGHI can only load 16-bit values and the benefit of LHI is that the instruction only takes 4 bytes to encode instead of 6 which saves space.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -10283,14 +10283,14 @@ OMR::Z::TreeEvaluator::arraycmpHelper(TR::Node *node,
             {
             if (return102)
                {
-               generateRIInstruction(cg, TR::InstOpCode::LA, node, retValReg, 2);
+               generateRIInstruction(cg, TR::InstOpCode::LGFI, node, retValReg, 2);
```
dhong44: Similar to https://github.com/eclipse/omr/pull/3127#discussion_r228721789 we should use `TR::InstOpCode::getLoadHalfWordImmOpCode()` to use LGHI or LHI. https://github.com/eclipse/omr/blob/a44b67653214e2884d9890937b08e65e60d3d308/compiler/z/codegen/InstOpCode.cpp#L274

muzaffar1331:
@dhong44 I made some changes. Can you please review them and suggest if anything else needs to be modified? Thanks :)

compiler/z/codegen/FPTreeEvaluator.cpp
```diff
@@ -459,7 +459,8 @@ convertToFixed(TR::Node * node, TR::CodeGenerator * cg)
    else  // Signed Target Path
       {
       //1) Reset targetRegister
-      generateRIInstruction(cg, TR::InstOpCode::LA, node, targetRegister, 0);
+
+      cursor = generateRRInstruction(cg, TR::InstOpCode::XGR, node, targetRegister, targetRegister);
```
dhong44: A different pull request removed the cursor variable which is causing compilation errors, cursor doesn't have to be set and can safely be removed here.

dhong44:
I think the OMR project prefers rebasing instead of merge commits when fixing merge conflicts. We might have to double check with @fjeremic but I would prefer if these commits were rebased and squashed.

muzaffar1331:
@dhong44 Thanks for the review! Made changes to use the helper functions instead of the ternary operator. Closing this PR due to conflicts and rebase issues. Will submit the commit in a new PR.

fjeremic:
> @dhong44 Thanks for the review! Made changes to use the helper functions instead of the ternary operator. Closing this PR due to conflicts and rebase issues. Will submit the commit in a new PR.

@muzaffar1331 you are able to rebase and force push to the branch corresponding to this PR. GitHub will naturally update the PR accordingly. It is useful to keep the same PR for archeological purposes as all the reviews have already happened here. In case there is a problem with the change in the future we can look back and figure out what we missed. My suggestion is to re-open this PR if possible.

muzaffar1331:
@fjeremic I can't access the Reopen button for this PR as I had deleted the repository associated with it earlier. I haphazardly pushed the commits on the master branch in the previous repository, and when I merged incoming the changes and made the final commits on top of it, I probably did something wrong during the rebase process, which wrecked the workflow. With little forethought, I deleted the original fork, forked a new repo and re-applied the changes there :/

PR #3129, opened 2018-10-29
AArch64: Implement iabs/labsEvaluator()

knn-k:
This commit implements iabs/labsEvaluator() for aarch64.

### DIFF
--- a/compiler/aarch64/codegen/UnaryEvaluator.cpp
+++ b/compiler/aarch64/codegen/UnaryEvaluator.cpp
@@ -81,17 +81,35 @@ TR::Register *OMR::ARM64::TreeEvaluator::lnegEvaluator(TR::Node *node, TR::CodeG
 	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
 	}
 
+static TR::Register *commonIntegerAbsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *reg = cg->gprClobberEvaluate(firstChild);
+   TR::Register *tempReg = cg->allocateRegister();
+
+   bool is64bit = node->getDataType().isInt64();
+   TR::InstOpCode::Mnemonic eorOp = is64bit ? TR::InstOpCode::eorx : TR::InstOpCode::eorw;
+   TR::InstOpCode::Mnemonic subOp = is64bit ? TR::InstOpCode::subx : TR::InstOpCode::subw;
+
+   generateArithmeticShiftRightImmInstruction(cg, node, tempReg, reg, is64bit ? 63 : 31);
+   generateTrg1Src2Instruction(cg, eorOp, node, reg, reg, tempReg);
+   generateTrg1Src2Instruction(cg, subOp, node, reg, reg, tempReg);
+
+   cg->stopUsingRegister(tempReg);
+   node->setRegister(reg);
+   cg->decReferenceCount(firstChild);
+   return reg;
+   }
+
 TR::Register *OMR::ARM64::TreeEvaluator::iabsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
-	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::iabsEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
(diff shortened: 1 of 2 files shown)


knn-k:
These evaluators are used in `comptest`.

compiler/aarch64/codegen/UnaryEvaluator.cpp
```diff
@@ -81,17 +81,36 @@ TR::Register *OMR::ARM64::TreeEvaluator::lnegEvaluator(TR::Node *node, TR::CodeG
 	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
 	}
 
+static TR::Register *commonAbsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
```
0xdaryl: Can you call this `commonIntegerAbsEvaluator` to make it clear this is for integer operations?

compiler/aarch64/codegen/UnaryEvaluator.cpp
```diff
@@ -81,17 +81,36 @@ TR::Register *OMR::ARM64::TreeEvaluator::lnegEvaluator(TR::Node *node, TR::CodeG
...
+   {
+   TR::Register *trgReg = cg->allocateRegister();
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *srcReg = cg->evaluate(firstChild);
```
0xdaryl: Could this be a clobberEvaluate to save a register?  I think you still need a temp register, but the target and source could potentially be shared.

compiler/aarch64/codegen/OMRTreeEvaluator.cpp
```diff
@@ -343,10 +343,11 @@ OMR::ARM64::TreeEvaluator::checkcastAndNULLCHKEvaluator(TR::Node *node, TR::Code
 
 TR::Register *
 OMR::ARM64::TreeEvaluator::treetopEvaluator(TR::Node *node, TR::CodeGenerator *cg)
-	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::treetopEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
-	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
-	}
+   {
```
0xdaryl: can you split the treetopEvaluator into a separate commit?  It's not related to abs.

knn-k: I separated treetopEvaluator() as #3151.

knn-k:

compiler/aarch64/codegen/UnaryEvaluator.cpp
```diff
@@ -81,17 +81,35 @@ TR::Register *OMR::ARM64::TreeEvaluator::lnegEvaluator(TR::Node *node, TR::CodeG
...
 	}
 
+static TR::Register *commonIntegerAbsEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *reg = cg->gprClobberEvaluate(firstChild);
+   TR::Register *tempReg = cg->allocateRegister();
+
+   bool is64bit = node->getDataType().isInt64();
+   TR::InstOpCode::Mnemonic eorOp = is64bit ? TR::InstOpCode::eorx : TR::InstOpCode::eorw;
```
andrewcraik: While this doesn't need to be fixed here, it might be worth considering some helpers somewhere in the opcode implementation which will return the right opcode for an operation based on a width. We have some of those in the x86 where you can ask for a load or other operator for a given width which centralizes the opcode selection and reduces the ternary operators in the codegen code. It also provides a central place to update opcode choices later if needed. Again, more of a suggestion as the aarch64 implementation moves forward rather than saying this code itself should not be accepted.

0xdaryl: I agree with this suggestion.  It's been handy on x86 for sure.  @knn-k , please open an issue for this so it can get on the backlog.

0xdaryl:
@genie-omr build aarch64

compiler/aarch64/codegen/UnaryEvaluator.cpp
knn-k: Opened Issue #3269.

PR #3132, opened 2018-10-29
Rename evacuate read barrier type

amicic:
Rename the barrier type to range check, which represents the trigger
criteria, rather then the action that barrier will perform. Evacuate is
a specific action performed by Concurrent Scavenger read barrier. Some
other GCs or profiling/debug tools could use the same read barrier
trigger but with its own action (that is not evacuate).

Temporarily keeping the old types, until downstream projects (OpenJ9)
are updated to use the new type.


charliegracie:
@genie-omr build all

PR #3133, opened 2018-10-30
AArch64: Separate SP and XZR registers

knn-k:
Registers SP and XZR share the same binary encoding in aarch64, but
it is better to separate these registers from the viewpoints of
register allocation and debug output.


knn-k:
The AppVeyor build failed with "The remote server returned an error: (500) Internal Server Error."

0xdaryl:
@genie-omr build aarch64

0xdaryl:
A full genie build is not required as this is an AArch64-specific change.

PR #3134, opened 2018-10-30
Disable CGroup Calls in omrsysinfo for TPF

jjohnst-us:
A semi-recent update in the omrsysinfo segment added CGroup calls in the
retrieveLinuxCgroupMemoryStats function.  TPF currently doesn't support
CGroup calls and needs guard macros around such calls so that it can be
configured out during TPF builds.

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -1511,6 +1511,7 @@ _cleanup:
 #define CGROUP_MEMORY_STAT_CACHE "cache"
 #define CGROUP_MEMORY_STAT_CACHE_SZ (sizeof(CGROUP_MEMORY_STAT_CACHE)-1)
 
+#if !defined(OMRZTPF)
 /**
  * Function collects memory usage statistics from the memory subsystem of the process's cgroup.
  *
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@bharathappali can you quickly review since it changes the default return value for one of the functions? Thanks

charliegracie:
@jjohnst-us can you resolve the merge conflict?

ashu-mehra:
@jjohnst-us I think you can put whole of `retrieveLinuxCgroupMemoryStats` under `!defined(OMRZTPF)
` to prevent including it in TPF builds.There is only one call to `retrieveLinuxCgroupMemoryStats` which is already protected by `!defined(OMRZTPF)`

bharathappali:
I have few views over the issue mentioned in the description.

The only caller for the function `retrieveLinuxCgroupMemoryStats` is `retrieveLinuxMemoryStats` and its initiating the call in the part guarded by `#if !defined(OMRZTPF)` so the function `retrieveLinuxCgroupMemoryStats` will not be called in the `TPF mainframe platform`.

But as it might be needed to avoid any failures(compiling or runtime) due this call, I'm thinking if the whole function `retrieveLinuxCgroupMemoryStats` could be guarded by `#if !defined(OMRZTPF)`.

Please correct me if I'm wrong.

charliegracie:
@genie-omr build all

PR #3136, opened 2018-10-30
Get ddrgen working on OSX

mikezhang1234567890:
The following changes will enable ddrgen to create the generated DDR
classes on OSX.

Create dSYM bundle for OMR shared libraries and dont require .dSYM
extension in parser since the bundle is passed in as the .dbg file.

Dwarfdump displays flag attribute values as a string true/false on OSX
instead of 1 or 0. Parse the value as a string and fall back to looking
for 1 or 0 if it doesn't find either "true" or "false" string.

Parse bit offset field for OSX since the data member location field is
missing if the field is a bit field, and instead we look at the
data_bit_offset field.

Include unknown dies in die tree when parsing dwarfdump output, since
there is an issue with parsing based on last indentation with parent
dies that the parser does not recognize. We then remove the unknown
dies from tree afterwards.

Escape sequences don't work in getmacros script for OSX, so replace
`\t` with character class `[[:space:]]` and `\U` with bash scripting to
upper case characters. These solutions are much more widely supported
and seem to work with all AIX, BSD, GNU and Windows based tools we use.

Replace // with / in filepath for getmacros script since // comments
out the rest of the line in C/C++.

### DIFF
--- a/ddr/lib/ddr-scanner/dwarf/DwarfParser.cpp
+++ b/ddr/lib/ddr-scanner/dwarf/DwarfParser.cpp
@@ -23,6 +23,7 @@
 
 /* Statics to create: */
 static void deleteDie(Dwarf_Die die);
+static void cleanUnknownDiesInCU(Dwarf_CU_Context *cu);
 static int parseDwarfInfo(char *line, Dwarf_Die *lastCreatedDie, Dwarf_Die *currentDie,
 	size_t *lastIndent, unordered_map<Dwarf_Die *, Dwarf_Off> *refToPopulate, Dwarf_Error *error);
 static int parseCompileUnit(char *line, size_t *lastIndent, Dwarf_Error *error);
@@ -79,7 +80,7 @@ dwarf_init(int fd,
 	FILE *fp = NULL;
 	if (DW_DLV_OK == ret) {
 		stringstream ss;
-		ss << "dwarfdump " << filepath << ".dSYM" << " 2>&1";
+		ss << "dwarfdump " << filepath << " 2>&1";
 		fp = popen(ss.str().c_str(), "r");
 		if (NULL == fp) {
 			ret = DW_DLV_ERROR;
@@ -104,6 +105,7 @@ dwarf_init(int fd,
 			}
 		}
 		free(buffer);
+		cleanUnknownDiesInCU(Dwarf_CU_Context::_currentCU);
 	}
 	if (DW_DLV_OK == ret) {
 		for (unordered_map<Dwarf_Die *, Dwarf_Off>::iterator it = refToPopulate.begin(); it != refToPopulate.end(); ++it) {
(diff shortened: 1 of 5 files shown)


mikezhang1234567890:
@keithc-ca can you review please?

mikezhang1234567890:
Ddrgen when run as part of the OpenJ9 build process will fail without those changes. It's not strictly necessary since I don't believe ddrgen is run anywhere else, and will not run for OSX by default anyways. Should I remove the dependency then?

keithc-ca:
I don't think we need hold this up because of https://github.com/eclipse/openj9/pull/3489. As long as this doesn't break any existing use cases it should be fine: building OpenJ9 on OSX with ddr support is not an existing use case.

ddr/lib/ddr-scanner/dwarf/DwarfParser.cpp
```diff
@@ -296,42 +344,36 @@ parseDwarfDie(char *line, Dwarf_Die *lastCreatedDie,
...
+		newDie->_attribute = NULL;
+		newDie->_context = Dwarf_CU_Context::_currentCU;
+		newDie->_child = NULL;
+		newDie->_sibling = NULL;
+		if (0 == *lastIndent) {
+			/* No last indent indicates that this Die is at the start of a CU. */
+			newDie->_parent = NULL;
+			Dwarf_CU_Context::_currentCU->_die = newDie;
+		} else if (spaces > *lastIndent) {
+			/* If the Die is indended farther than the last Die, it is a first child. */
```
keithc-ca: typo: indended -> indented

ddr/tools/getmacros
```diff
@@ -200,6 +202,9 @@ parse_namespace_policy_files() {
 	local filenum=0
 
 	for f in "${annotated_files[@]}"; do
+		#remove any double // which comments out the rest of the line in C/C++
+		f=${f//\/\//\/}
```
keithc-ca: This doesn't look right. I think it should be `${f/\/\//\/}`, but that would still only do one substitution.
These are pathnames in the output of `grep -l`; perhaps a better solution would be to add
`| sed -e 's|//+|/|g'` to the end of the pipeline that produces `annotated_files`.

ddr/tools/getmacros
```diff
@@ -73,7 +73,9 @@ begin_namespace() {
 
 set_default_namespaces() {
 	# Remove directory and extension; make first letter uppercase; append 'Constants'.
-	basename=$(echo ${f} | sed -e 's|^.*/||' -e 's|\..*$||' -e 's|\(.\)|\U\1|')
+	basename=${f##*/}
+	basename=${basename%.*}
+	basename=${basename^}
```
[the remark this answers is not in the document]
mikezhang1234567890: I'll change it to `basename=${basename%%.*}` which will provide the same results as the sed script.

ddr/tools/getmacros
mikezhang1234567890: The two slashes at the start matches all instances.

From https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion :
```
${parameter/pattern/string}
The pattern is expanded to produce a pattern just as in filename expansion. Parameter is expanded and the longest match of pattern against its value is replaced with string. If pattern begins with ‘/’, all matches of pattern are replaced with string. Normally only the first match is replaced. 
```

keithc-ca: Thanks, I missed that subtlety as I read the bash manual.
I'm still unclear how we wind up with paths containing `//`.

mikezhang1234567890: We pass in `../` as the parameter for the top level search directory. In the `get_all_annotated_files` function, it looks like OSX doesn't "fix" the path and just leaves the extra slash in.

keithc-ca: Perhaps it would be better to remove any trailing slash from that directory rather than apply a fix to every file name.

ddr/lib/ddr-scanner/dwarf/DwarfScanner.cpp
```diff
@@ -673,7 +673,7 @@ DwarfScanner::addDieToIR(Dwarf_Die die, Dwarf_Half tag, NamespaceUDT *outerUDT,
 
 	rc = getOrCreateNewType(die, tag, &newType, outerUDT, &isNewType);
 
-	if ((DDR_RC_OK == rc) && isNewType && NULL == outerUDT) {
+	if ((DDR_RC_OK == rc) && isNewType) {
```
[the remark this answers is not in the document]
keithc-ca: Thanks, Mike. I believe you've addressed all my comments except for this one.

With this change, we might blacklist nested types that weren't blacklisted before: see commit a6c45b43757b5054831c98a61b2993f1eec32944:
> Change the blacklist to only exclude top level types, in order to
> exclude the problematic type names that shows up on AIX. Compiling
> with this change on AIX and Linux both showed no apparent issues.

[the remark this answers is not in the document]
keithc-ca: Thanks, Mike.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
@mikezhang1234567890 is the PR ready?

mikezhang1234567890:
Yes it is.

PR #3138, opened 2018-10-31
Fix result loading from unresolved class data snippet

NigelYiboYu:
An unresolved class is resolved at runtime by a PicBuilder routine and the
result (resolved class) is stored in a patchable literal pool
slot referenced by the last slot of an 'unresolved data snippet'.

On Z hardware lower than z10, the JIT compiled code does not use the
correct slot of the snippet and, as a result of this, crashes when loading
the resolved class from the data snippet. This bug is fixed by this
commit.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -11465,43 +11465,45 @@ OMR::Z::TreeEvaluator::loadaddrEvaluator(TR::Node * node, TR::CodeGenerator * cg
    if (symRef->isUnresolved())
       {
 #ifdef J9_PROJECT_SPECIFIC
+      TR::UnresolvedDataSnippet * uds = new (cg->trHeapMemory()) TR::UnresolvedDataSnippet(cg, node, symRef, false, false);
+      cg->addSnippet(uds);
+
+      // Generate branch to the unresolved data snippet
+      TR::Instruction * cursor = generateRegUnresolvedSym(cg, TR::InstOpCode::getLoadOpCode(), node, targetRegister, symRef, uds);
+      uds->setBranchInstruction(cursor);
+
+      // Create a patchable data in litpool
+      TR::S390WritableDataSnippet * litpool = cg->CreateWritableConstant(node);
+      litpool->setUnresolvedDataSnippet(uds);
+
+      TR::S390RILInstruction * lrlInstr = NULL;
       if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10))
          {
-         TR::UnresolvedDataSnippet * uds;
-         TR::Instruction * cursor;
-
-         // create UnresolvedDataSnippet
-         uds = new (cg->trHeapMemory()) TR::UnresolvedDataSnippet(cg, node, symRef, false, false);
-         cg->addSnippet(uds);
-
-         // generate branch to the unresolved data snippet
-         cursor = generateRegUnresolvedSym(cg, TR::InstOpCode::getLoadOpCode(), node, targetRegister, symRef, uds);
-
-         uds->setBranchInstruction(cursor);
-
-         // create a patchable data in litpool
-         TR::S390WritableDataSnippet * litpool = cg->CreateWritableConstant(node);
-         litpool->setUnresolvedDataSnippet(uds);
-
-         TR::S390RILInstruction * LRLinst;
-         LRLinst = (TR::S390RILInstruction *) generateRILInstruction(cg, TR::InstOpCode::getLoadRelativeLongOpCode(), node, targetRegister, reinterpret_cast<void*>(0xBABE), 0);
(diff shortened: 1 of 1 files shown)


compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11465,43 +11465,45 @@ OMR::Z::TreeEvaluator::loadaddrEvaluator(TR::Node * node, TR::CodeGenerator * cg
    if (symRef->isUnresolved())
       {
 #ifdef J9_PROJECT_SPECIFIC
-      if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10))
-         {
-         TR::UnresolvedDataSnippet * uds;
-         TR::Instruction * cursor;
+      // create UnresolvedDataSnippet
```
fjeremic: This comment seems not useful. We can remove it.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11465,43 +11465,45 @@ OMR::Z::TreeEvaluator::loadaddrEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+      TR::UnresolvedDataSnippet * uds = new (cg->trHeapMemory()) TR::UnresolvedDataSnippet(cg, node, symRef, false, false);
+      cg->addSnippet(uds);
 
-         // create UnresolvedDataSnippet
-         uds = new (cg->trHeapMemory()) TR::UnresolvedDataSnippet(cg, node, symRef, false, false);
-         cg->addSnippet(uds);
+      // generate branch to the unresolved data snippet
```
fjeremic: Capitalization needed.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11465,43 +11465,45 @@ OMR::Z::TreeEvaluator::loadaddrEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
 
-         uds->setBranchInstruction(cursor);
-
-         // create a patchable data in litpool
-         TR::S390WritableDataSnippet * litpool = cg->CreateWritableConstant(node);
-         litpool->setUnresolvedDataSnippet(uds);
-
-         TR::S390RILInstruction * LRLinst;
-         LRLinst = (TR::S390RILInstruction *) generateRILInstruction(cg, TR::InstOpCode::getLoadRelativeLongOpCode(), node, targetRegister, reinterpret_cast<void*>(0xBABE), 0);
+      TR::S390RILInstruction * LRLinst;
```
fjeremic: I'd prefer if we initialize these pointers to `NULL` so we catch errors fast in case we don't initialize on some paths. Obviously for simple code such as this it we know it is initialized however it's good practice to initialize such pointers everywhere (for more complicated code paths where it is not obvious the variable is initialized on all paths).

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11465,43 +11465,45 @@ OMR::Z::TreeEvaluator::loadaddrEvaluator(TR::Node * node, TR::CodeGenerator * cg
...
+      if (cg->getS390ProcessorInfo()->supportsArch(TR_S390ProcessorInfo::TR_z10))
+         {
+         LRLinst = static_cast<TR::S390RILInstruction *>(generateRILInstruction(cg, TR::InstOpCode::getLoadRelativeLongOpCode(), node, targetRegister, reinterpret_cast<void*>(0xBABE), 0));
```
fjeremic: There is a reason `generateRILInstruction` does not return a `TR::S390RILInstruction*` type. In general one should not assume this because arbitrary instructions may be inserted before the instruction you wanted to generate, and sometimes even after. If you want to do this case I'd like to see a fatal assert validating that `LRinst` is of the correct instruction kind by checking via `getKind`. Similarly for the `else` path.

NigelYiboYu: Added a fatal assert to check for instruction format.

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: Variable initialized to NULL.

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: Fixed.

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: Fixed.

fjeremic:
@genie-omr build zlinux,zos

fjeremic:
AppVeyor build failed with infrastructure issues. This is a Z only change though and the Eclipse builds passed. I'm comfortable enough to merge.

PR #3139, opened 2018-11-01
AArch64: Fix generation of instructions with zero register

knn-k:
This commit fixes the problem in register assignment for zero register.

### DIFF
--- a/compiler/aarch64/codegen/ARM64Instruction.cpp
+++ b/compiler/aarch64/codegen/ARM64Instruction.cpp
@@ -165,6 +165,83 @@ void TR::ARM64Trg1Src2Instruction::assignRegisters(TR_RegisterKinds kindToBeAssi
    setSource2Register(assignedSource2Register);
    }
 
+// TR::ARM64Trg1Src3Instruction:: member functions
+
+bool TR::ARM64Trg1Src3Instruction::refsRegister(TR::Register *reg)
+   {
+   return (reg == getTargetRegister() ||
+           reg == getSource1Register() ||
+           reg == getSource2Register() ||
+           reg == getSource3Register());
+   }
+
+bool TR::ARM64Trg1Src3Instruction::usesRegister(TR::Register *reg)
+   {
+   return (reg == getSource1Register() || reg == getSource2Register() || reg == getSource3Register());
+   }
+
+bool TR::ARM64Trg1Src3Instruction::defsRegister(TR::Register *reg)
+   {
+   return (reg == getTargetRegister());
+   }
+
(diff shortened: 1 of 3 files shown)


knn-k:
Removed "WIP:".  Ready for review.

knn-k:
`generateSrc1ImmInstruction()` and `generateSrc2Instruction()` in `GenerateInstructions.cpp` work as expected, but they use a temporary register as the target register at this point.  Using the zero register in those instruction is a future work.

0xdaryl:
@genie-omr build aarch64

0xdaryl:
Code looks ok, but can you be more descriptive in your commit message about what the "register assignment problem" is for zero registers?

knn-k:
`generateTrg1ZeroSrc1Instruction()` used to set a register dependency for `TR::RealRegister::xzr` for both pre- and post-conditions.  It caused a NULL access in `reverseSpillState()` with totalUseCount=1 and futureUseCount=0 for the virtual register.
Then I changed `generateTrg1ZeroSrc1Instruction()` to set only the pre-condition for `xzr`, and it resulted in the following register assignment.
```
GPR_0017(2/2)=w1 GPR_0018(1/1)=w2 !GPR_0018(1/1)=xzr [         0x1802bc0]	0 	orrw 	w0, w2, w1
```
GPR_0018 is supposed to be the zero register here, but the JIT assigns `w2` to GPR_0018 first, then assigns `xzr` to the same virtual register based on the pre-condition.

The new code in this PR does not set the register dependency nor call `assignOneRegister()` for `xzr`, but just set the bit pattern for the zero register in the instruction.  This requires a new class `ARM64Trg1ZeroSrc1Instruction`, but I think it is the most simple solution.

0xdaryl:
I've been thinking about the original problem you're trying to solve and wonder if there isn't something more problematic going on with the register dependencies.

Would you happen to have a log (with traceRA) of the original case where you set pre- and post- conditions for xzr and caused a crash?  It doesn't make sense to me that the total use count of the virtual should be 1 when it appears in two register dependency conditions (unless you've investigated and already have an answer for that).  Please upload to the public Slack arm channel if you do.  I'm hoping there might be a more natural solution to this problem than introducing a new instruction kind.

knn-k:
@0xdaryl I don't have the log file of the original problem any more.  I tried to recreate the problem using `generateMulInstruction()`, but it resulted in unexpected failure.

knn-k:
Interim conclusion from discussion with @0xdaryl:
The initial value of the total use count was wrong for the virtual register associated with the zero register with the implementation before fix.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3140, opened 2018-11-01
Remove unnecessary cast

gacholio:
The cast serves no purpose and causes a compile error on some
compilers.

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -4509,7 +4509,7 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
 	}
 	if (state->multiLineCounter < subsystemMetricMapElement->metricElementsCount) {
 		for (i = 0; i < state->multiLineCounter; i++) {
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

rwy0717:
What's the point of incrementing in a loop? Is this just weird code?

fjeremic:
> What's the point of incrementing in a loop? Is this just weird code?

I agree. I don't get it. The type is defined as:

https://github.com/eclipse/omr/blob/66a192ee120d047bf499d804ab0fa42a3a6f353a/port/unix/omrsysinfo.c#L318-L323

So how isn't:

```
	for (i = 0; i < state->multiLineCounter; i++) {
			currentElement++;
		}
```

equivalent to:

```
currentElement += state->multiLineCounter
```

I think we should improve this code.

bharathappali:
@fjeremic i have made the changes in this commit.
https://github.com/bharathappali/omr/commit/03d283552950eece3258e809da5d77fc0a7685b3

will be raising a PR. Thanks

PR #3141, opened 2018-11-01
Add warning flags to thread objects

rwy0717:

### DIFF
--- a/thread/CMakeLists.txt
+++ b/thread/CMakeLists.txt
@@ -305,17 +305,17 @@ target_compile_definitions(j9thr_obj
 		$<TARGET_PROPERTY:omr_base,INTERFACE_COMPILE_DEFINITIONS>
 )
 
-add_library(j9thrstatic STATIC
-	$<TARGET_OBJECTS:j9thr_obj>
-)
-
 if(OMR_WARNINGS_AS_ERRORS)
-	target_compile_options(j9thrstatic PRIVATE ${OMR_WARNING_AS_ERROR_FLAG})
+	target_compile_options(j9thr_obj PRIVATE ${OMR_WARNING_AS_ERROR_FLAG})
 endif()
 
 if(OMR_ENHANCED_WARNINGS)
-	target_compile_options(j9thrstatic PRIVATE ${OMR_ENHANCED_WARNING_FLAG})
+	target_compile_options(j9thr_obj PRIVATE ${OMR_ENHANCED_WARNING_FLAG})
 endif()
+
+add_library(j9thrstatic STATIC
+	$<TARGET_OBJECTS:j9thr_obj>
+)
 
 target_include_directories(j9thrstatic
 	INTERFACE
--- a/thread/common/omrthread.c
+++ b/thread/common/omrthread.c
@@ -5022,12 +5022,15 @@ monitor_notify_three_tier(omrthread_t self, omrthread_monitor_t monitor, int not
 	MONITOR_LOCK(monitor, CALLER_NOTIFY_ONE_OR_ALL);
 	queue = monitor->waiting;
 	if (queue) {
+#if defined(THREAD_ASSERTS)
 		intptr_t state;
 
 		state = omrthread_spinlock_swapState(monitor, J9THREAD_MONITOR_SPINLOCK_EXCEEDED);
 		ASSERT((state == J9THREAD_MONITOR_SPINLOCK_OWNED)
 			|| (state == J9THREAD_MONITOR_SPINLOCK_EXCEEDED));
-
+#else
+		omrthread_spinlock_swapState(monitor, J9THREAD_MONITOR_SPINLOCK_EXCEEDED);
+#endif
 		if (notifyall) {
 			/* set all the thread flags */
 			do {
--- a/thread/common/omrthread.c
+++ b/thread/common/omrthread.c
@@ -1374,7 +1374,6 @@ void
 omrthread_detach(omrthread_t thread)
 {
 	uintptr_t destroy = 0;
-	uintptr_t attached = 0;
 
 	if (thread == NULL) {
 		thread = MACRO_SELF();
@@ -1391,7 +1390,7 @@ omrthread_detach(omrthread_t thread)
 				 * detached.  Mark it dead so that it can be destroyed.
 				 */
 				thread->flags |= J9THREAD_FLAG_DEAD;
-				attached = destroy = 1;
+				destroy = 1;
 			} else {
 				/* A j9-created thread should never be dead here */
 				destroy = thread->flags & J9THREAD_FLAG_DEAD;
--- a/thread/common/omrthread.c
+++ b/thread/common/omrthread.c
@@ -1518,7 +1518,6 @@ thread_wrapper(WRAPPER_ARG arg)
 {
 	omrthread_t thread = (omrthread_t)arg;
 	omrthread_library_t lib = NULL;
-	uintptr_t flags;
 	int globalAlreadyLocked = GLOBAL_NOT_LOCKED;
 
 	ASSERT(thread);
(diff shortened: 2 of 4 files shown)


thread/CMakeLists.txt
```diff
@@ -293,6 +293,14 @@ add_library(j9thr_obj OBJECT
...
 )
 
+if(OMR_WARNINGS_AS_ERRORS)
+	target_compile_options(j9thr_obj PRIVATE ${OMR_WARNING_AS_ERROR_FLAG})
+endif()
+
+if(OMR_ENHANCED_WARNINGS)
+	target_compile_options(j9thr_obj PRIVATE ${OMR_ENHANCED_WARNING_FLAG})
+endif()
+
```
charliegracie: Do we want to remove the setting of the option from below that was setting it on the library?

rwy0717: Sure, I can remove it.

rwy0717: Seeing as the j9thrstatic doesn't have any objects itself, adding compilation flags to it doesn't make sense.

rwy0717: done

charliegracie:
@genie-omr build all

youngar:
There are some errors that need to be fixed before we can merge this, luckily they don't seem very bad.
```
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c: In function ‘omrthread_detach’:
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c:1377:12: error: variable ‘attached’ set but not used [-Werror=unused-but-set-variable]
15:20:39   uintptr_t attached = 0;
15:20:39             ^
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c: In function ‘thread_wrapper’:
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c:1522:12: error: variable ‘flags’ set but not used [-Werror=unused-but-set-variable]
15:20:39   uintptr_t flags;
15:20:39             ^
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c: In function ‘threadDestroy’:
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c:2001:22: error: variable ‘lib’ set but not used [-Werror=unused-but-set-variable]
15:20:39   omrthread_library_t lib;
15:20:39                       ^
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c: In function ‘monitor_notify_three_tier’:
15:20:39 /home/jenkins/jenkins-agent/workspace/PullRequest-linux_390-64/thread/common/omrthread.c:5026:12: error: variable ‘state’ set but not used [-Werror=unused-but-set-variable]
15:20:39    intptr_t state;
15:20:39             ^
```

charliegracie:
it is strange that the test is failing on Linux_x86-64 which is one of the last auto-tools builds and this change does not change anything in that build tree.

youngar:
I think its only the `travis x86_64` build that is running autotools, which is still passing.

charliegracie:
I was confused.... It was the compressed pointers one that is passing which looks like it is auto-tools... going to open a PR shortly to resolve the issues

charliegracie:
Actually it is probably better to just make all of the fixes in this PR... @rwy0717 and @youngar here is a patch I believe should resolve the issues.  I do not have the right GCC right now to test the fixes.  There are 2 fixes...
patch 1 to fix the ASSERT macro:
```
diff --git a/thread/common/threaddef.h b/thread/common/threaddef.h
index 1f9ac66..11dd1e3 100644
--- a/thread/common/threaddef.h
+++ b/thread/common/threaddef.h
@@ -80,7 +80,7 @@
 #else /* THREAD_ASSERTS */
 
 #ifndef ASSERT
-#define ASSERT(ignore) ((void)0)
+#define ASSERT(ignore) ((void)sizeof(ignore))
 #endif /* ASSERT */
 
 #ifndef ASSERT_DEBUG
```

patch to fix the 2 actual unused variables:
```
diff --git a/thread/common/omrthread.c b/thread/common/omrthread.c
index ead0f23..14c26c7 100644
--- a/thread/common/omrthread.c
+++ b/thread/common/omrthread.c
@@ -1374,7 +1374,6 @@
 omrthread_detach(omrthread_t thread)
 {
 	uintptr_t destroy = 0;
-	uintptr_t attached = 0;
 
 	if (thread == NULL) {
 		thread = MACRO_SELF();
@@ -1391,7 +1390,7 @@
 				 * detached.  Mark it dead so that it can be destroyed.
 				 */
 				thread->flags |= J9THREAD_FLAG_DEAD;
-				attached = destroy = 1;
+				destroy = 1;
 			} else {
 				/* A j9-created thread should never be dead here */
 				destroy = thread->flags & J9THREAD_FLAG_DEAD;
@@ -1519,7 +1518,6 @@
 {
 	omrthread_t thread = (omrthread_t)arg;
 	omrthread_library_t lib = NULL;
-	uintptr_t flags;
 	int globalAlreadyLocked = GLOBAL_NOT_LOCKED;
 
 	ASSERT(thread);
@@ -1556,7 +1554,6 @@
 		OMROSCOND_WAIT_LOOP();
 	}
 	thread->flags |= J9THREAD_FLAG_STARTED;
-	flags = thread->flags;
 
 	/* Set the numa affinity if it is pending on this thread */
 #if defined(OMR_PORT_NUMA_SUPPORT)
```

rwy0717:
I'll add the patches to this PR later today (or worst cast tomorrow). Thanks :+1:

rwy0717:
I just updated the PR with fixes.

I decided not to use `sizeof(ignore)` fix to the `ASSERT` macro, which seemed a little dirty to me. Rather, I wrapped the offending assert in an ifdef.

There was also some other numa code which wasn't correctly ifdef'd.

thread/common/omrthread.c
```diff
@@ -1998,11 +1995,16 @@ threadAllocate(omrthread_library_t lib, int globalIsLocked)
...
+#if defined(OMR_OS_WINDOWS)
 	omrthread_library_t lib;
+#endif
 
 	ASSERT(thread);
+
+#if defined(OMR_OS_WINDOWS)
 	lib = thread->library;
 	ASSERT(lib);
+#endif
```
rwy0717: The var `lib`  is only "actually" used in windows, however this `ASSERT(lib)` line was enabled on every platform. With this change, the assert will only be on windows debug builds. Should I leave the assert enabled on non-windows debug builds? I can switch the ifdef to:

```c++
#if defined(OMR_OS_WINDOWS) || defined(THREAD_ASSERTS)
```

youngar:
@genie-omr build all

youngar:
@genie-omr build all

youngar:
@genie-omr build plinux

thread/common/omrthread.c
charliegracie: Having to #if def all uses of the ASSERT macro can be quite tedious and error prone.  Is there anyway to make the ASSERT no generate a warning if THREAD_ASSERTS is not defined?

charliegracie: If the macro ASSERT can not be fixed then I think the above code should have both #if defs included to be clear

rwy0717: Updated to include both ifdefs.

rwy0717:
This is ready for another "build all"

charliegracie:
@genie-omr build all

youngar:
@genie-omr build all

PR #3146, opened 2018-11-02
Removes unnecessary loop & makes pointer incremented to certain level

bharathappali:
This fix provides the optimization like removing the for loop which
performs increment operation on a pointer till certain level and
incrementing the pointer directly to desired level.


fjeremic:
Thanks for fixing this. Just need some small changes on your commit message and we'll be good to go [1]

> The first line describes the change made. It is written in the imperative mood, and should say what happens when the patch is applied. Keep it short and simple. The first line should be less than 70 characters, where reasonable, and should be written in **sentence case** preferably not ending in a period.

fjeremic:
@genie-omr build all

PR #3148, opened 2018-11-02
Provide workaround when dllload fails to open lib

tajila:
Provide workaround when dllload fails to open lib

Linux and MacOS use the directory of the portLibrary as a
fall back for when a library is not able to be loaded in the 
specified path. z/OS does not support the necessary functions 
(ie. dladdr) to do the same. As a result, a second attempt to 
load a dll will be done with just the name of the library which 
will trigger a search on the filesystem.

Signed-off-by: Tobi Ajila <tobi_ajila@ca.ibm.com>


tajila:
@youngar Can you please review this

youngar:
@genie-omr build zos

Issue #3112, opened 2018-10-23
Rename getIA32*Instruction to getX86*Instruction


0dvictor:
The following methods have `IA32` in their names but are used for both 32-bit and 64-bit. This Issue is to track renaming them.

https://github.com/eclipse/omr/blob/2739e6d0c5e0b37bc4184d9d277f9e9f92cd55f3/compiler/x/codegen/OMRInstruction.hpp#L146
https://github.com/eclipse/omr/blob/2739e6d0c5e0b37bc4184d9d277f9e9f92cd55f3/compiler/x/codegen/OMRInstruction.hpp#L148
https://github.com/eclipse/omr/blob/2739e6d0c5e0b37bc4184d9d277f9e9f92cd55f3/compiler/x/codegen/OMRInstruction.hpp#L154

PR #3150, opened 2018-11-03
Rename getIA32*Instruction to getX86*Instruction

muzaffar1331:
This patch renames the methods getIA32RegInstruction(), getIA32LabelInstruction(),
getIA32ImmInstruction() to getX86RegInstruction(), getX86LabelInstruction(),
getX86ImmInstruction() as requested in issue #3112 since they are used for both
32-bit and 64-bit.

Closes: #3112
Signed-off-by: Muzaffar Auhammud <muzaffar@cyberstorm.mu>

### DIFF
--- a/compiler/x/codegen/BinaryEvaluator.cpp
+++ b/compiler/x/codegen/BinaryEvaluator.cpp
@@ -2625,7 +2625,7 @@ TR::Register *OMR::X86::TreeEvaluator::integerDivOrRemEvaluator(TR::Node *node, 
                                                       divisionLabel,
                                                       node->getOpCode(),
                                                       node->getType(),
-                                                      divideInstr->getIA32RegInstruction()->getIA32RegRegInstruction(),
+                                                      divideInstr->getX86RegInstruction()->getIA32RegRegInstruction(),
                                                       cg));
          }
 
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -180,7 +180,7 @@ bool isConditionCodeSetForCompareToZero(TR::Node *node, bool justTestZeroFlag)
         prevInstr;
         prevInstr = prevInstr->getPrev())
       {
-      prevRegInstr = prevInstr->getIA32RegInstruction();
+      prevRegInstr = prevInstr->getX86RegInstruction();
 
       // The register must be equal and the node size must be equal in order to
       // insure the instruction is setting the condition code based on the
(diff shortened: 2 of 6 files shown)


andrewcraik:
@0dvictor can you make a first pass on this one as the originator of #3112?

0dvictor:
It's great to see this being cleaned up. The changes look good to me.

One FYI, this is one usage of the changed API in OpenJ9, you might want to create an Issue there to let them know. https://github.com/eclipse/openj9/blob/05ef11976cae67ca8172d7e8370d946ed3abd0ff/runtime/compiler/x/codegen/J9TreeEvaluator.cpp#L2402

andrewcraik:
Given the downstream use shouldn't this be a two stage change? Rename and add the original methods calling the renamed methods and, in a second commit once the downstream migration is done, remove the old IA32 names?

0dvictor:
I'm okay with either approach. Is there any guidelines about introducing changes that potentially require downstream projects to update their code?

andrewcraik:
Well in general we should avoid breaking changes and stage changes in such a way to that consumers can migrate cleanly would be a general principle. It is certainly one that @charliegracie and others have espoused on Slack and in various other forums repeatedly.

0dvictor:
Good suggestion. Although I don't know how we can avoid the situation in general, let's at least not break OpenJ9 this time.

0dvictor:
I went through OpenJ9 code, and found its usage of `getIA32RegInstruction()` is unnecessary. I created a pull request to remove the usage (https://github.com/eclipse/openj9/pull/3557). One it is merged, we can safely merge this PR.

andrewcraik:
>Good suggestion. Although I don't know how we can avoid the situation in general, let's at least not break OpenJ9 this time.

I agree there are certainly times where avoiding the break is very difficult or the cost of doing so prohibitive to the point where we want to work with downstream projects for a coordinated change. This would be a case where I don't think that level of coordination is required.

andrewcraik:
@0dvictor please mark WIP pending merge of eclipse/openj9#3557

0dvictor:
eclipse/openj9#3557 has been merged. This PR should be ready to go.

muzaffar1331:
@andrewcraik @0xdaryl Can someone review this PR please? Thanks

Leonardo2718:

Leonardo2718:
@0xdaryl do you mind if I re-assign this to myself (well, actually just assign this to myself because no-one has taken ownership yet, just thought I'd ask any ways since your review was requested :slightly_smiling_face: )?

0xdaryl:
Sure, thanks.

PR #3154, opened 2018-11-07
Improve byte simplification

charliegracie:
The b<add/sub>Simplifier routines were only simplifying math
operations on 2 constants. The short simplifier handles way more
cases and uses a generic helper routine that should work for byte
operations as well.  This change switches the 2 APIs above to use
the common helper.

Also fix the spacing in the s<add/sub>Simplifier.

### DIFF
--- a/compiler/optimizer/OMRSimplifierHandlers.cpp
+++ b/compiler/optimizer/OMRSimplifierHandlers.cpp
@@ -6833,24 +6833,12 @@ TR::Node *daddSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
 
 TR::Node *baddSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
    {
-   simplifyChildren(node, block, s);
-
-   TR::Node * firstChild = node->getFirstChild(), * secondChild = node->getSecondChild();
-
-   if (firstChild->getOpCode().isLoadConst() && secondChild->getOpCode().isLoadConst())
-      {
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
@genie-omr build all

vijaysun-omr:
The change being made here is a good one and should be merged.

More generally, I wonder if there are other places in the compiler where byte operations get treated sub-optimally in the compiler as well. e.g. folding of operations other than badd/bsub. Java bytecodes widen the sub-integer types to 32-bit integers and arithmetic operations are done on integers. Is this a design point that you considered ?

charliegracie:
It appears that other simplifications for mul/div, etc are also not comprehensive for byte and short data types.  I looked at improving those but it was going to be a lot more complicated.  I am hoping to have more time in the near future to come back and hopefully make some improvements there as well.

vijaysun-omr:
Yes and when I check the value propagation table for bmul and bdiv, I see:

 constrainChildren,        // TR::bmul
...
 constrainChildren,        // TR::bdiv

which basically means no special handling. Contrasted with the int equivalents:

 constrainImul,            // TR::imul
...
  constrainIdiv,            // TR::idiv

There may also be issues in passes like induction variable analysis detecting an induction variable that is narrower than int. So, those may be places to look at as well.

charliegracie:
any future changes would be in another PR. Just to be clear

vijaysun-omr:
Merging this PR since it's a good improvement to simplifier on it's own.

PR #3156, opened 2018-11-07
fix error propagation from init_thread_library in omrthread_attach_ex

aandyl:
Without this change, I get a segfault dereferencing a NULL pointer retrieved from a TLS entry that was never set. With this change I get:

Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

### DIFF
--- a/thread/common/omrthread.c
+++ b/thread/common/omrthread.c
@@ -1183,11 +1183,12 @@ omrthread_attach(omrthread_t *handle)
 intptr_t
 omrthread_attach_ex(omrthread_t *handle, omrthread_attr_t *attr)
 {
-	intptr_t retVal = J9THREAD_SUCCESS;
+	intptr_t retVal = J9THREAD_ERR;
 	omrthread_t thread;
 	omrthread_library_t lib;
 
 	if (init_thread_library()) {
+		retVal = J9THREAD_ERR;
 		goto cleanup0;
 	}
 


charliegracie:
@aandyl this is testing an error condition where the Thread Library fails to be initialized?  I think there needs to be an error code set there but it is very strange to me that code would be calling `omrthread_attach_ex` before the thread library was started. I do notice that the older `omrthread_attach` API does exactly what you are proposing so it likely makes sense but I was recently thinking about removing the `init_thread_library` call from both `omrthread_attach_ex` and `omrthread_attach` and just add an assert that the thread library was initialized instead.  Can you explain more how this situation is occurring?

charliegracie:
As an aside you will need to sign the commit and sign the Eclipse CLA as per the [Contributing.md](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#submitting-a-contribution) and [here](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#legal-considerations)

aandyl:
I am not sure exactly what's resulting in the call to omrthread_attach_ex. If I remember, my observation was that it was happening while loading the j9prt library, so might be some behavior of port library initialization. The thread library initialization was failing due to an unrelated platform problem, but I figured this independently is a reasonable fix for better error reporting.

I've completed the CLA now, sorry for the delay.

charliegracie:
@genie-omr build all

PR #3157, opened 2018-11-07
Refactor recognition of standard types

keithc-ca:
Types names in the blob and the superset should be consistent. Previously, `unsigned short int` would appear in the blob while the superset would say `U16`.

The code from `Macro.cpp` has moved to `Type.cpp` and has been improved to handle `const` and `volatile`.

Also some minor cleanup:
* remove unnecessary #includes from header files
* break #include cycle involving Symbol_IR.hpp and Type.hpp
* add #includes to *.cpp files as needed


charliegracie:
@genie-omr build all

PR #3158, opened 2018-11-07
Call generateCatchBlockBBStartPrologue in ARM BBStartEvaluator

fjeremic:
We need to call the common code generator API in the BBStartEvaluator
on ARM as downstream projects may choose to override behavior of what
happens in catch blocks.

This synchronizes all the platforms in terms of catch block handling.


0xdaryl:
@knn-k , you'll need to implement this in the BBStartEvaluator for AArch64 as well.

0xdaryl:
@genie-omr build arm

PR #3159, opened 2018-11-07
WIP: Rename GC Mode to Write Barrier Kind [Phase 1/2]

0dvictor:
The old GC mode field and its related queries are actually
for write barrier kind; fixing the naming.

### DIFF
--- a/compiler/control/OMROptions.hpp
+++ b/compiler/control/OMROptions.hpp
@@ -1890,25 +1890,30 @@ public:
    static bool createDebug();
    static TR_Debug * findOrCreateDebug();
 
-   TR_WriteBarrierKind getGcMode()           { return _gcMode; }
+   /**   \brief Returns the type of write barriers
+   */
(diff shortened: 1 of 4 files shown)


andrewcraik:
I think the deprecated comments need to be doxygen of some kind. I agree with the renaming taking place - it makes it more clear what you are asking about.

0dvictor:
Thank you both for valuable comments. I'm refactoring this code mainly because GC now also has a "Read Barrier Kind" similar to "Write Barrier Kind", and either one by itself is a "GC mode". I'm working support to query and cache the "Read Barrier Kind" now.
@0xdaryl has a good point that the Option class may not be a suitable place for either "Write Barrier Kind" or "Read Barrier Kind". I'm marking this as WIP for now and I'll rethink the design.

0dvictor:
I've created #3168 to start moving write barrier kind to Object Model. This PR is no longer needed; closing.

PR #3087, opened 2018-10-18
AArch64: Binary Encoding Fix and Addition of 'B' and 'BL' Instructions

marufunb:
Some binary encoding values mismatched because of the mismatch in opcodes.
Those values are being updated. 'B' and 'BL' instructions are being added
 to ORMInstOpCodeEnum.cpp and OpBinary.cpp

Fixes #3017 
Signed-off-by: marufunb <mrahma15@unb.ca>

PR #3161, opened 2018-11-08
AArch64: Separate OpCodeNames.hpp from ARM64Debug.cpp

knn-k:
This commit moves the array members of opCodeToNameMap[] in ARM64Debug.cpp
to a new file OpCodeNames.hpp for easier maintenance.


knn-k:
The idea of separation came from #3087.

knn-k:
AppVeyor build failed with the following message:
```
winflexbison not installed. An error occurred during installation:
 The remote server returned an error: (503) Server Unavailable. Service Unavailable
```

0xdaryl:
This is an AArch64-specific change.  The AppVeyor failure can be ignored.

0xdaryl:
@genie-omr build aarch64

0xdaryl:
Apologies for being obtuse, but what is the purpose of this separation?  We've done this in other areas of the compiler that use extensible classes or enums, but in this particular case that doesn't appear to be why this is done.

I understand from #3087 that it was a comment of mine that motivated this, but I cannot remember what the comment was nor find a reference to it.

knn-k:
I thought it would be clearer that OpCode names for debug also need to be updated when you add/change OpCodes in OpBinary.cpp and OMRInstOpCodeEnum.hpp.

knn-k:
Cancelling this PR.

PR #3163, opened 2018-11-08
Implement InstOpCode::getMnemonicName for z codegen

dhong44:
Implement InstOpCode::getMnemonicName for z codegen so that
the mnemonic name can be retrieved without going through
the debug object.


dhong44:
@fjeremic could I get a review please?

compiler/z/codegen/InstOpCode.hpp
```diff
@@ -42,6 +42,7 @@ class InstOpCode: public OMR::InstOpCodeConnector
       OMR::InstOpCodeConnector(m) { }
 
    /* OMR TODO: Delegating method; need to be removed and replace with getMnemonic */
+   const char* getMnemonicName()                                        { return TR::InstOpCode::metadata[_mnemonic].name; }
    TR::InstOpCode::Mnemonic getOpCodeValue()                            { return _mnemonic; }
    TR::InstOpCode::Mnemonic setOpCodeValue(TR::InstOpCode::Mnemonic op) { return _mnemonic = op; }
```
fjeremic: Was this TODO meant for the two functions below your change? If so we should move it down.

fjeremic:
Do we need to fix up the S390Debug.cpp to call this new API? Or even better are we able to remove the Debug API now?

compiler/z/codegen/InstOpCode.hpp
dhong44: 27171ed

fjeremic:
Yeah I say we consolidate to one API. We're asking about the name of an `OMR::InstOpCode` instance so that's where the API belongs (the change in this PR). We should remove all other ones so people don't shoot themselves in the foot and start using the wrong API in the future.

dhong44:
Removed the other APIs in cb2df3e.

fjeremic:
Before I merge @dhong44 have we checked if there are any downstream consumers of the APIs we're removing here?

@genie-omr build all

dhong44:
The only downstream consumer I am familiar with is openj9. The jdk builds without any issues with openj9 and this branch of OMR but I haven't ran any testing. I can mark as WIP and run some tests.

fjeremic:
@dhong44 looks like we have some jitbuilder failures on Linux on z which we'll need to fix. Please see the failing build.

dhong44:
I think the error was a result of declaring the virtual funciton `getOpCodeName` since it was never implemented, 0509dfa should fix this.

fjeremic:
@genie-omr build all

fjeremic:
Change looks good to me, I'll merge once WIP is removed.

fjeremic:
@genie-omr build all

PR #3164, opened 2018-11-08
Inflate all available free regions for expanded size

LinHu2016:
- due to the start address of heap memory for each numa node 
need to alignment with page size and region size, free region counts
might be not distributed evenly among numa nodes. 
inflate all available regions from numa nodes Round Robin until
inflating all of regions or reaching expanded size.


gc/base/PhysicalSubArenaRegionBased.hpp
```diff
@@ -108,6 +108,11 @@ class MM_PhysicalSubArenaRegionBased: public MM_PhysicalSubArena
 	 */
 	void validateNumaSymmetry(MM_EnvironmentBase *env);
 
+	void resetAllNumaNodesInflated();
+
+	void setNumaNodeInflated(uintptr_t idx);
+
+	BOOLEAN areAllNumaNodesInflated();
```
charliegracie: This should be `bool` and not `BOOLEAN`

charliegracie:
@genie-omr build all

charliegracie:
@amicic or @dmitripivkine can one of you review this as well? Thanks

LinHu2016:
update with precompile option OMR_GC_VLHGC

dmitripivkine:
I don't like two moments:
- `inflated` slot located in global `J9MemoryNodeDetail` structure but this information is very local to subarena inflation process (means "free region list exhausted for node" so name can be better) and has no use anywhere else
- we call `manager->acquireSingleTableRegion()` for case when we know it will return `NULL` because this node was recognized as `inflated` in previous iterations.

However I understand why it designed this way. I would appreciate if somebody suggest how to implement it better way but I am ok with current code (with type of areAllNumaNodesInflated() changed to `bool` as Charlie suggested).

gc/base/PhysicalSubArenaRegionBased.hpp
LinHu2016: fixed

LinHu2016:
As Dmitri's suggestion, inflated flag has been removed from global  structure, instead using freeregionCountFromAllNumaNode to confirm if all free regions has been inflated.

charliegracie:
@genie-omr build all

dmitripivkine:
This implementation is better but ugly part is walk regions table list to count number of elements.
What about to have decrementing counter for NUMA nodes with non-empty region lists, not for regions. We can create helper `manager->areFreeRegionsForNode()` (check `_freeRegionTable[numaNode]`) and call `manager->acquireSingleTableRegion()` for case if it is true (we do have free regions). So any returned NULL means free regions list is exhausted and we can decrement counter by 1. We can iterate until reach point none of NUMA nodes have non-empty region lists. Will this idea work? Did I miss something?

LinHu2016:
updated to use decrementing counter for NUMA nodes to avoid iterating all of free regions from numa nodes.

gc/base/PhysicalSubArenaRegionBased.hpp
```diff
@@ -107,7 +107,6 @@ class MM_PhysicalSubArenaRegionBased: public MM_PhysicalSubArena
 	 * @param env[in] the current thread
 	 */
 	void validateNumaSymmetry(MM_EnvironmentBase *env);
-
```
dmitripivkine: I would exclude changes in this file from this pull request

gc/base/HeapRegionManagerTarok.cpp
```diff
@@ -86,6 +86,16 @@ MM_HeapRegionManagerTarok::tearDown(MM_EnvironmentBase *env)
 	MM_HeapRegionManager::tearDown(env);
 }
 
+bool
+MM_HeapRegionManagerTarok::areFreeRegionsForNode(MM_EnvironmentBase *env, uintptr_t numaNode)
+{
+	bool ret = false;
```
amicic: this could be just one line: return (... != ...);

LinHu2016: done

gc/base/PhysicalSubArenaRegionBased.hpp
LinHu2016: done

gc/base/PhysicalSubArenaRegionBased.cpp
```diff
@@ -191,16 +191,44 @@ MM_PhysicalSubArenaRegionBased::doExpandInSubSpace(MM_EnvironmentBase *env, uint
...
 	uintptr_t regionSize = manager->getRegionSize();
 
+	/*
+	 due to heap address for each numa node has to alignment with region size and page size, free region count might be not distributed evenly among numa nodes.
+	 use numaNodeCountWithFreeRegions and areFreeRegionsForNode() to inflate all available regions from numa nodes round robin until reaching expanded size.
+	 */
+	uintptr_t numaNodeCountWithFreeRegions = 0;
+#if defined (OMR_GC_VLHGC)
+	numaNodeCountWithFreeRegions = _affinityLeaderCount;
+#	for (uintptr_t idx=0; idx<_affinityLeaderCount; idx++) {
```
charliegracie: There is something wrong with this line. See the `#` at the beginning

LinHu2016: oOps! I have removed '#', Thanks @charliegracie

charliegracie:
@genie-omr build all

PR #3165, opened 2018-11-08
Simplify child index calculation code in PRE

cathyzhyi:
The child index of rhsNode of a store is more relevant to whether the
store is indirect or direct rather than whether it's a wrtbar. Simplify
the calculation to make it more clear.


andrewcraik:
Looks reasonable to me, but I'd like an opinion from @vijaysun-omr since he has more familiarity with PRE.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks ok to me. Merging.

PR #3168, opened 2018-11-08
Provide read/write barrier kind info in Object Model

0dvictor:
Object Model now support querying for the kind of read and write barriers.

### DIFF
--- a/compiler/env/OMRObjectModel.hpp
+++ b/compiler/env/OMRObjectModel.hpp
@@ -32,6 +32,7 @@ namespace OMR { typedef OMR::ObjectModel ObjectModelConnector; }
 #endif
 
 #include <stdint.h>        // for int32_t, int64_t, uint32_t
+#include "omrgcconsts.h"
 #include "env/jittypes.h"  // for uintptrj_t, intptrj_t
 #include "il/DataTypes.hpp"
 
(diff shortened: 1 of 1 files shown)


0dvictor:
Good point, although I tested on OpenJ9 and it compiled successfully, there is a theoritical risk that adopting change breaks a downstream project. I'm marking it as WIP for now and I'll make updates to avoid moving the enum.

0dvictor:
The changes have been updated so that existing `TR_WriteBarrierKind` enum is kept. It is ready continue to be reviewed now.

0xdaryl:
I'm wondering if we should be taking the opportunity to improve the synergy with the GC and using the constants it uses, rather than the JIT defining its own.  In particular, `include_core/omrgcconsts.h` defines `MM_GCWriteBarrierType` and `MM_GCReadBarrierType` already which your modes map to.

I would be loathe to do this if it introduced a build dependence on the GC, but I don't believe it does.  It simply requires including header files from OMR core.  @charliegracie FYI.

0dvictor:
> I'm wondering if we should be taking the opportunity to improve the synergy with the GC and using the constants it uses, rather than the JIT defining its own.

Good idea. Let me give it a try. In the meantime, I'm marking this pull request as WIP.

0dvictor:
The pull request has been updated and re-verified. Now it uses types from `include_core/omrgcconsts.h` @0xdaryl could you take a look at this change once more?

0xdaryl:
Before merging, I'd appreciate @charliegracie 's +1 on this approach.

charliegracie:
I really like the idea of using the GC header file directly.  I think this is a great start but when the GC is enabled in OMR most consumers are going to use gencon which has a write barrier. Should these new APIs actually just query the GC to ask what the write barrier / read barrier types are?

0xdaryl:
@charliegracie, you're right that the GC should be consulted on what barrier is in play.  One way to do it with the extensible class mechanism is for the consuming project to extend this ObjectModel class with its own implementations of `readBarrierType` and `writeBarrierType` which would ask the GC, there by decoupling the OMR compiler from GC.  The drawback of this is that if this is a common pattern for multiple downstream projects (and it likely is), then they will each end up having to implement similar (if not identical) code to query the GC.  A cleaner solution to this will need some thought, but not as part of this PR (which is a great start to improving the synergy between OMR components).

PR #3170, opened 2018-11-09
Fix Concurrent Scavenge GC compilation errors

0dvictor:
Now Concurrent Scavenge GC should compile with all X86 specs.

### DIFF
--- a/gc/structs/ForwardedHeader.cpp
+++ b/gc/structs/ForwardedHeader.cpp
@@ -281,7 +281,7 @@ MM_ForwardedHeader::copyOrWaitOutline(omrobjectptr_t destinationObjectPtr)
 
 			if (0 == remainingSizeToCopy) {
 				if (participatingInCopy) {
-					fomrobject_t newValue = ((outstandingCopies - 1) << OUTSTANDING_COPIES_SHIFT) | _beingCopiedTag;
+					fomrobject_t newValue = (fomrobject_t)(((outstandingCopies - 1) << OUTSTANDING_COPIES_SHIFT) | _beingCopiedTag);
 					if (oldValue != lockCompareExchangeObjectHeader(&objectHeader->slot, oldValue, newValue)) {
 						continue;
 					}
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -4627,7 +4627,7 @@ MM_Scavenger::scavengeRoots(MM_EnvironmentBase *env)
 {
 	Assert_MM_true(concurrent_state_roots == _concurrentState);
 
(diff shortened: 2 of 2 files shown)


charliegracie:
@amici Can you review this please?

@0dvictor before your change which x86 specs were supported? After your change which specs do you believe are now supported?

amicic:
looks good

0dvictor:
> before your change which x86 specs were supported?

Only `linux_x86-64` and `linux_x86-64_cmprssptrs` is supported before this change.

> After your change which specs do you believe are now supported?

This change only makes other X86 specs compile, but doesn't make any of them fully functional. More debugging and development will be required to enable them. The specs I tested for compilation are `linux_x86`, `win_x86`, `win_x86-64` and `win_x86-64_cmprssptrs`. I anticipate other X86 specs will also compile successfully, but I do not have the right environment to try it out,

amici:
> @amici Can you review this please?

I was wondering when did I join this group... but you actually needed some other `amici`. ;)

gc/structs/ForwardedHeader.cpp
```diff
@@ -281,7 +281,7 @@ MM_ForwardedHeader::copyOrWaitOutline(omrobjectptr_t destinationObjectPtr)
 
 			if (0 == remainingSizeToCopy) {
 				if (participatingInCopy) {
-					fomrobject_t newValue = ((outstandingCopies - 1) << OUTSTANDING_COPIES_SHIFT) | _beingCopiedTag;
+					fomrobject_t newValue = (fomrobject_t)(((outstandingCopies - 1) << OUTSTANDING_COPIES_SHIFT) | _beingCopiedTag);
```
charliegracie: This code still seems wrong but it is wrong for compressed references.  In a compressed references VM an object slot is only 32 bits so performing math on a uintptr_t and then casting it to fomrobject_t is wrong. The `outstandingCopies` variable should be of the appropriate type or should have some sort of mask applied to it before we store it into `newValue`. @amicic what are your thoughts?

amicic: (fomrobject_t) in this case plays the role of the mask

amicic: we can put comment now that with (fomrobject_t) downcasting in case of CR, we don't lose significant digits

charliegracie: Ok thanks... I am ok with this change but I would like to see the existing issues investigated. Thanks.

charliegracie:
@genie-omr build all

gc/structs/ForwardedHeader.cpp
amicic: https://github.com/eclipse/omr/issues/3175 
Please assign to me, and tag appropriately.

PR #3173, opened 2018-11-09
Check register properties for generating load

r30shah:
While resolving register dependencies at split point (GlRegDeps) we check type of the node to decide whether if we need 32 Bit or 64 Bit load instruction. If we have evaluated this node to use 64 Bit register it should use 64 Bit load instruction to generate necessary instruction to satisfy register dependencies at split point.


fjeremic:
@genie-omr build zlinux,zos

PR #3177, opened 2018-11-12
AArch64: Fix createPrologue()

knn-k:
createPrologue() has saved the argument registers using strimmx
or vstrimmd instructions, regardless of the data size.  On the other
hand, the stack offset for the argument registers takes the data
size into consideration.
This commit changes the createPrologue() instruction generation, so
that 32-bit and smaller arguments are saved by strimmw or vstrimms
instructions.


0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3178, opened 2018-11-12
Allows iterator to complete a cycle irrespective of file availability

bharathappali:
Its observed that the cgroup information iterator is going to infinity loop if it doesn't find the cgroup metric file in the `/sys/fs/cgroup/<subsystem>/<file>` so added changes to skip and proceed if a file is not available so that other metric can be accessed uninterrupted.

Added a macro (CGROUP_METRIC_FILE_CONTENT_MAX_LIMIT) to define the max
limit of cgroup metric file content and removed duplicate documentation


charliegracie:
@genie-omr build all

charliegracie:
@DanHeidinga could you give this a quick once over since you reviewed the original change?

port/unix/omrsysinfo.c
```diff
@@ -4497,21 +4497,21 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
...
-		rc = readCgroupMetricFromFile(portLibrary, state->subsystemid, subsystemMetricMapElement->metricFileName, currentElement->metricKeyInFile, metricElement->value, &(*state).fileContent, isSingleLineKeyValue, sizeRef);
+		rc = readCgroupMetricFromFile(portLibrary, state->subsystemid, subsystemMetricMapElement->metricFileName, currentElement->metricKeyInFile, metricElement->value, &(state->fileContent), isSingleLineKeyValue, sizeRef);
 		if (0 == rc) {
 			if(NULL == state->fileContent) {
 				state->multiLineCounter = SINGLE_CGROUP_METRIC;
 			} else {
 				state->multiLineCounter = 0;
 			}
 		} else {
+			state->count = state->count + 1;
```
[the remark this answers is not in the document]
bharathappali: @DanHeidinga  Thanks for the review.

> `rc` will be an error code in this case. Won't that terminate correctly coded uses of the iterator?

In my implementation I'm using these api and printing the metrics based on `rc` value but not terminating the iterator. Iterator termination is completely based on `omrsysinfo_cgroup_subsystem_iterator_hasNext` api which is internally dependent on the check for `state->count`. So if the `file is not found` and `state->count` is unchanged its ending up in searching for same file again and again (A bug in the way i implemented these api's).

> If the intention is to allow the iterator to continue, and I'm not clear on why it should

I thought it would be good if iterator continues irrespective of `rc` value as there might be errors in accessing the metric from file such as `file not found` or `metric not found` in file, these shouldn't interrupt the iterator from reading other metrics in the list which might be available. Please correct me if I'm wrong 

> then the rc will need to be changed.

Changing the `rc` would be a good option as it will alert the caller that the current metric is not available. 

> Also, use `state->count += 1`

Sure will be using it.

port/unix/omrsysinfo.c
```diff
@@ -4526,8 +4526,13 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
 				goto _end;
 			}
 	  	} while ((part = strtok_r(NULL, "\n", &saveptr)) != NULL);
-	} else {
-		portLibrary->mem_free_memory(portLibrary, state->fileContent);
+		state->multiLineCounter = state->multiLineCounter + 1;
```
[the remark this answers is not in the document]
bharathappali: > Shouldn't the increment happen unconditionally in the loop?

if the token matches `currentElement->metricKeyInFile` then the multiLineCounter is incremented and goes to the `_end` block of the function. In case if the metric is missing from that file then as `currentElement` is dependent on multiLineCounter (`currentElement += state->multiLineCounter;`) it always end up in the same metric to search again and again as the condition `if (state->multiLineCounter < subsystemMetricMapElement->metricElementsCount)` will be satisfied always. So if the metric is not found after parsing the whole cached file content i'm incrementing the `state->multiLineCounter` outside the loop. 

Please correct me if i'm wrong

Thanks

bharathappali:
@DanHeidinga made changes to set the `metricElement.key` value in case of error so that it would be handy to the caller to display that specific metric is `unavailable` if not found. For example usecase like [this](https://github.com/eclipse/openj9/pull/2954)

Thanks @ashu-mehra for suggesting a optimized way to handled edge cases

port/unix/omrsysinfo.c
```diff
@@ -4470,65 +4451,72 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
...
-		if (0 == rc) {
-			if(NULL == state->fileContent) {
-				state->multiLineCounter = SINGLE_CGROUP_METRIC;
-			} else {
-				state->multiLineCounter = 0;
-			}
-		} else {
+		rc = readCgroupMetricFromFile(portLibrary, state->subsystemid, subsystemMetricMapElement->metricFileName, currentElement->metricKeyInFile, &(state->fileContent), metricElement->value, sizeRef);
+
+		// In error condition or single metric condition we increment the counter to continue to next metric
```
charliegracie: `//` comments are not c89 compliant. Can you change these to `/* */` comments?

port/unix/omrsysinfo.c
```diff
@@ -4470,65 +4451,72 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
...
-				goto _end;
+				rc = 0;
+				break;
 			}
 	  	} while ((part = strtok_r(NULL, "\n", &saveptr)) != NULL);
-	} else {
-		portLibrary->mem_free_memory(portLibrary, state->fileContent);
-		state->count = state->count + 1;
+
+		// if we complete parse the fileContent and couldn't find the metric return error code
```
charliegracie: `//` comments are not c89 compliant. Can you change these to `/* */` comments?

port/unix/omrsysinfo.c
```diff
@@ -4552,6 +4539,7 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
 			}
 		}
 	}
+	metricElement->key = currentElement->metricTag;
```
DanHeidinga: Why write this in the case of an error?

bharathappali: To have it handy for caller to display the value for this particular key is `unavailable`.

port/unix/omrsysinfo.c
bharathappali: sure @charliegracie will be changing it.

port/unix/omrsysinfo.c
bharathappali: sure will be changing it.

port/unix/omrsysinfo.c
bharathappali: > A cleaner approach is to have a separate call that provides the metric name and one that gets the data.

@DanHeidinga Will be adding an api to get the metric name seperately. Thank you

DanHeidinga: Are there other usecases or pieces of data that need to be available even in error cases?  If so, let's get them described here so the right api can be designed upfront rather than through review comments.

bharathappali: @DanHeidinga As of now i haven't came across any requirement for making other pieces of data  available in that context.

charliegracie:
@genie-omr build all

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,104 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		if (NULL == part) {
+			rc = OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_METRIC_NOT_AVAILABLE;
+		}
+
+		state->multiLineCounter += 1;
+	}
+
+	if (state->multiLineCounter >= subsystemMetricMapElement->metricElementsCount) {
+		if (NULL != state->fileContent) {
+			portLibrary->mem_free_memory(portLibrary, state->fileContent);
```
DanHeidinga: when does this get freed if someone doesn't iterate all the elements?  Does the iterator need a destroy function to cleanup any resources if the not everything is walked?

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,104 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
-	} else {
-		portLibrary->mem_free_memory(portLibrary, state->fileContent);
-		state->count = state->count + 1;
+
+		/* if we complete parse the fileContent and couldn't find the metric return error code */
+		if (NULL == part) {
+			rc = OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_METRIC_NOT_AVAILABLE;
+		}
+
+		state->multiLineCounter += 1;
```
DanHeidinga: Trying to understand the `multi` part of this name.... Is it a lineCounter?  What makes it `multi` line?  Is there something that distinguishes single vs multi lines?

port/common/omrport.c
```diff
@@ -274,6 +274,7 @@ static OMRPortLibrary MasterPortLibraryTable = {
 	omrsysinfo_is_running_in_container, /* sysinfo_is_running_in_container */
 	omrsysinfo_cgroup_subsystem_iterator_init, /* sysinfo_cgroup_subsystem_iterator_init */
 	omrsysinfo_cgroup_subsystem_iterator_hasNext, /* omrsysinfo_cgroup_subsystem_iterator_hasNext */
+	omrsysinfo_cgroup_subsystem_get_metric_key, /* omrsysinfo_cgroup_subsystem_get_metric_key */
```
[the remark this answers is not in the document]
bharathappali: @DanHeidinga Thanks for the review.

will be naming it to `omrsysinfo_cgroup_subsystem_iterator_metric_key`

include_core/omrport.h
```diff
@@ -1544,6 +1543,8 @@ typedef struct OMRPortLibrary {
 	int32_t (*sysinfo_cgroup_subsystem_iterator_init)(struct OMRPortLibrary *portLibrary, uint64_t subsystem, struct OMRCgroupMetricIteratorState *state);
 	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_subsystem_iterator_hasNext "omrsysinfo_cgroup_subsystem_iterator_hasNext"*/
 	BOOLEAN (*sysinfo_cgroup_subsystem_iterator_hasNext)(struct OMRPortLibrary *portLibrary, const struct OMRCgroupMetricIteratorState *state);
+	/** see @ref omrsysinfo.c::omrsysinfo_cgroup_subsystem_get_metric_key "omrsysinfo_cgroup_subsystem_get_metric_key"*/
+	int32_t (*sysinfo_cgroup_subsystem_get_metric_key)(struct OMRPortLibrary *portLibrary, const struct OMRCgroupMetricIteratorState *state, char **metricKey);
```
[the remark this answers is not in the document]
bharathappali: Yeah it needs to be `const char **` thanks for pointing it. Will change it.

port/unix/omrsysinfo.c
bharathappali: > Does the iterator need a destroy function to cleanup any resources if the not everything is walked?

Yes, sorry for missing it in the first place. Will add an api (something like `omrsysinfo_cgroup_subsystem_iterator_clean_up`) for clearing all the allocated memory.

port/unix/omrsysinfo.c
```diff
@@ -4538,7 +4557,6 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
 		 * which is not required so we could remove it 
 		 */
 		size_t len = strlen(metricElement->value);
-		metricElement->key = currentElement->metricTag;
 		metricElement->units = currentElement->metricUnit;
 		if ((len > 0) && (metricElement->value[len-1] == '\n')) {
 			metricElement->value[--len] = '\0';
```
DanHeidinga: `--len` use `len-1` just like the code above as `len` is never used again.

port/unix/omrsysinfo.c
```diff
@@ -4557,6 +4575,15 @@ omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, st
 	return rc;
```
DanHeidinga: `*printUnits = FALSE;` just above this - can this just ensure that  `metricElement->units` is null?  And then `printUnits` can be removed?

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		break;
+	case OMR_CGROUP_SUBSYSTEM_CPUSET:
+		subsystemMetricMap = omrCgroupCpusetMetricMap;
+		break;
+	default:
+		rc = OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_UNAVAILABLE;
+		goto _end;
+	}
+	if (state->multiLineCounter < subsystemMetricMap[state->count].metricElementsCount){
+		*metricKey = (subsystemMetricMap[state->count].metricInfoElementList + state->multiLineCounter)->metricTag;
```
DanHeidinga: Can the `rc = 0` occur here so that all the "success" case is together?

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
 	return check;
 }
 
+int32_t
+omrsysinfo_cgroup_subsystem_iterator_metric_key(struct OMRPortLibrary *portLibrary, const struct OMRCgroupMetricIteratorState *state, const char **metricKey)
+{
+	int32_t rc = OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_METRIC_NOT_AVAILABLE;
+#if defined(LINUX) && !defined(OMRZTPF)
```
DanHeidinga: Code should validate that `metricKey` is not null.

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+	} else {
+		rc = OMRPORT_ERROR_SYSINFO_CGROUP_SUBSYSTEM_METRIC_NOT_AVAILABLE;
+	}
+_end:
+#endif
+	return rc;
+}
+
 int32_t
 omrsysinfo_cgroup_subsystem_iterator_next(struct OMRPortLibrary *portLibrary, struct OMRCgroupMetricIteratorState *state, struct OMRCgroupMetricElement *metricElement, BOOLEAN *printUnits, uint64_t sizeRef)
```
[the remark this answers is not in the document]
bharathappali: >  How is it different than checking whether the `OMRCgroupMetricElement->units` filed is null or not?

Even if `OMRCgroupMetricElement->units` is not null some of the metrics doesn't have units to measure like `bytes` or `microseconds` etc. `printUnits` jus tells if the metric has units to measure, if its true then it prints the unit of measure in javacore.

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		rc = readCgroupMetricFromFile(portLibrary, state->subsystemid, subsystemMetricMapElement->metricFileName, currentElement->metricKeyInFile, &(state->fileContent), metricElement->value, sizeRef);
+
+		/* In error condition or single metric condition we increment the counter to continue to next metric */
+		if ((0 != rc) || (NULL == state->fileContent)) {
+			state->count += 1;
 			goto _end;
 		}
-		
+
+		state->multiLineCounter = 0;
```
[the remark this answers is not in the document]
bharathappali: > Can we rename this variable fileMetricCounter to make it clear that this a count of the number of metrics in this file?

Sure will be naming it to `fileMetricCounter`

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
bharathappali: Its used in call for `readCgroupMetricFromFile`.

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
-				state->multiLineCounter = SINGLE_CGROUP_METRIC;
-			} else {
-				state->multiLineCounter = 0;
-			}
-		} else {
+		rc = readCgroupMetricFromFile(portLibrary, state->subsystemid, subsystemMetricMapElement->metricFileName, currentElement->metricKeyInFile, &(state->fileContent), metricElement->value, sizeRef);
+
+		/* In error condition or single metric condition we increment the counter to continue to next metric */
+		if ((0 != rc) || (NULL == state->fileContent)) {
+			state->count += 1;
```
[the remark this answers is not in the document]
bharathappali: Will be adding `state->fileContent = NULL;`

port/unix/omrsysinfo.c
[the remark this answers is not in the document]
DanHeidinga: It's passed to `readCgroupMetricFromFile` but never used in that function.  The code that used to use it was removed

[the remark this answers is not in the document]
DanHeidinga: My point is that `printUnits` should be a property of the `OMRCgroupMetricElement` and is preferably something that can be determined from the `units` field. 

> Even if OMRCgroupMetricElement->units is not null some of the metrics doesn't have units to measure 

If an metric doesn't have any units, why wouldn't `units` be null?

bharathappali:
@DanHeidinga I have made changes as per your review comments.

port/unix/omrsysinfo.c
bharathappali: added validation

port/unix/omrsysinfo.c
bharathappali: moved `rc = 0`

port/unix/omrsysinfo.c
bharathappali: changed `--len` to `len-1`

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		char *tempBuff = portLibrary->mem_allocate_memory(portLibrary, CGROUP_METRIC_FILE_CONTENT_MAX_LIMIT, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
 		char *part = "";
 		char *saveptr = NULL;
 		currentElement += state->multiLineCounter;
 		strcpy(tempBuff, state->fileContent);
 		part = strtok_r(tempBuff, "\n", &saveptr);
 	  	do {
-			if(0 == strncmp(part, currentElement->metricKeyInFile, strlen(currentElement->metricKeyInFile))) {
+			if (0 == strncmp(part, currentElement->metricKeyInFile, strlen(currentElement->metricKeyInFile))) {
 				part += strlen(currentElement->metricKeyInFile);
```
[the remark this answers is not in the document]
bharathappali: made `strlen(currentElement->metricKeyInFile)` to a local variable

port/unix/omrsysinfo.c
```diff
@@ -4465,70 +4446,108 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
 			goto _end;
 		}
-		
+
+		state->multiLineCounter = 0;
 	}
+
 	if (state->multiLineCounter < subsystemMetricMapElement->metricElementsCount) {
-		char *tempBuff = portLibrary->mem_allocate_memory(portLibrary, 1024, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
+		char *tempBuff = portLibrary->mem_allocate_memory(portLibrary, CGROUP_METRIC_FILE_CONTENT_MAX_LIMIT, OMR_GET_CALLSITE(), OMRMEM_CATEGORY_PORT_LIBRARY);
```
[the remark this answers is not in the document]
bharathappali: added a search mechanism to remove allocation

bharathappali:
@DanHeidinga I have made all the changes by addressing your review comments. I request you to have a look and suggest me changes needed if any. 

Thanks!

charliegracie:
@genie-omr build all

bharathappali:
@DanHeidinga I would like to request you to suggest me the changes required after your review. Thanks!

charliegracie:
@DanHeidinga are you ok with these changes now?

charliegracie:
@genie-omr build all

port/unix/omrsysinfo.c
```diff
@@ -4525,69 +4505,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+			} else if (tempPtr[index] == '\n') {
+				lineEnd = index;
+				if (0 == strncmp(&tempPtr[lineStart], currentElement->metricKeyInFile, metricKeyLen)) {
+					memset(metricElement->value, 0, MAX_LINE_LENGTH);
+					strncpy(metricElement->value, &tempPtr[lineSpace + 1], lineEnd - lineSpace);
+					isMetricFound = TRUE;
+					rc = 0;
+					break;
+				}
+				lineStart = index + 1;
```
DanHeidinga: I think this would be much clearer if written using pointers to search the string and would avoid some of the signed / unsigned comparisons.  Something along the lines of:

```c
char *start = state->fileContent;
char *current = state->fileContent;
char *end = state->fileContent + strlen(state->fileContent) - metricKeyLen;  /* Must ensure there is at least metric worth of data */
while (current < end) {
    if (0 == strncmp(start, currentElement->metricKeyInFile, metricKeyLen)) {
        char *newLine = current;
        current += metricKeyLen;
        /* advance past any whitespace */
        while ((current != end) &&  (' ' == *current)) {
            current += 1;
        }
        /* find the newline */
        newLine = current;       
        while ((newLine != end) && ('\n' != *newLine)) {
            newLine += 1;
        }
        strncpy(metricElement->value, current, newLine - current);
    }
    current += 1;
}
```

DanHeidinga:
(Sorry, realized I've had a pending review comments sitting here.  Thought I'd finished that off already)

port/unix/omrsysinfo.c
```diff
@@ -4466,69 +4446,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
-				rc = 0;
-				portLibrary->mem_free_memory(portLibrary, tempBuff);
-				state->multiLineCounter = state->multiLineCounter + 1;
-				goto _end;
+
+		state->fileMetricCounter = 0;
+	}
+
+	if (state->fileMetricCounter < subsystemMetricMapElement->metricElementsCount) {
+		currentElement += state->fileMetricCounter;
```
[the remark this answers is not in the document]
bharathappali: @DanHeidinga Thank you for the review. Made the declarations before the expression.

port/unix/omrsysinfo.c
```diff
@@ -4466,69 +4446,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		int32_t metricKeyLen = strlen(currentElement->metricKeyInFile);
+		BOOLEAN isMetricFound = FALSE;
+		int32_t lineStart = 0;
+		int32_t lineSpace = 0;
+		int32_t lineEnd = 0;
+		int32_t index = 0;
+		char *tempPtr = state->fileContent;
+		while (index < strlen(state->fileContent)) {
```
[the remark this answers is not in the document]
bharathappali: Changed this part of code to the one you have suggested, so this implementation will no longer be in it. Will be correcting this mistake in future code contributions. Thank you.

port/unix/omrsysinfo.c
```diff
@@ -4466,69 +4446,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+			if (tempPtr[index] == ' ') {
+				lineSpace = index;
+			} else if (tempPtr[index] == '\n') {
+				lineEnd = index;
+				if (0 == strncmp(&tempPtr[lineStart], currentElement->metricKeyInFile, metricKeyLen)) {
+					memset(metricElement->value, 0, MAX_LINE_LENGTH);
```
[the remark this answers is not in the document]
bharathappali: Initially i have found the results of previous metrics getting printed along with the current one as the number chars copied over might be less than the previous one so did a `memset`. I have removed `memset` and placing the last char as `\0` to avoid such errors.

port/unix/omrsysinfo.c
```diff
@@ -4525,69 +4505,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		state->fileMetricCounter = 0;
+	}
+
+	if (state->fileMetricCounter < subsystemMetricMapElement->metricElementsCount) {
+		currentElement += state->fileMetricCounter;
+		int32_t metricKeyLen = strlen(currentElement->metricKeyInFile);
+		BOOLEAN isMetricFound = FALSE;
+		int32_t lineStart = 0;
+		int32_t lineSpace = 0;
+		int32_t lineEnd = 0;
```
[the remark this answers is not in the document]
bharathappali: Changed this part of code to the one you have suggested, so this implementation will no longer be in it. Will be correcting this mistake in future code contributions. Thank you.

port/unix/omrsysinfo.c
```diff
@@ -4525,69 +4505,118 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+		int32_t index = 0;
```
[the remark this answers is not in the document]
bharathappali: Changed this part of code to the one you have suggested, so this implementation will no longer be in it. Will be correcting this mistake in future code contributions. Thank you.

port/unix/omrsysinfo.c
bharathappali: @DanHeidinga Thank you for suggesting this method, I have implemented it with minor changes. Please let me know any changes required.

bharathappali:
@DanHeidinga I have addressed the review comments, Please let me know if any changes are required. Thank you.

port/unix/omrsysinfo.c
```diff
@@ -4525,69 +4505,121 @@ omrsysinfo_cgroup_subsystem_iterator_hasNext(struct OMRPortLibrary *portLibrary,
...
+			if (0 == strncmp(current, currentElement->metricKeyInFile, metricKeyLen)) {
+        			char *newLine = current;
+        			current += metricKeyLen;
+        			/* advance past any whitespace */
+        			while ((current != end) &&  (' ' == *current)) {
+            				current += 1;
+        			}
+        			/* find the newline */
+        			newLine = current;       
+       		 		while ('\n' != *newLine) {
```
DanHeidinga: minor indentation nitpick

youngar:
@genie-omr build all

youngar:
Merging and I'll just open up a separate fix for the whitespace

PR #3182, opened 2018-11-12
Move variable declarations to top of scope for c89 compilers

DanHeidinga:


charliegracie:
@genie-omr build xlinux,pinux,zlinux,osx

PR #3183, opened 2018-11-13
Adding a .gitignore entry for emacs save files

vijaysun-omr:
emacs save files ending in ~ should not be tracked by git.
This commit adds them to the .gitignore file for the project.


0xdaryl:
A genie build is not necessary as it will not exercise this change.

PR #3184, opened 2018-11-13
AArch64: Add is64bit parameter to generateCompareInstruction()

knn-k:
This commit adds is64bit parameter to generateCompareInstruction(),
generateTestInstruction(), etc.
Those functions used to look at node->getDataType().isInt64() to
determine whether 64-bit or not.  It is inappropriate for lcmpeq
or other lcmp* nodes that compare 64-bit values and return 32-bit
value.


0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3166, opened 2018-11-08
Cleanup CMake defines and modify the README.md

charliegracie:
Clean up a bunch of CMake defines in the config, caches and correct some defaults.

Modify the README.md to use CMake to build instead of auto-tools

PR #3185, opened 2018-11-13
Add -fno-strict-aliasing to compiler options

pdbain-ibm:
Add this option to ARM, PPC linux, and X86 linux.

This is related to Issue https://github.com/eclipse/openj9/issues/2872


charliegracie:
@pdbain-ibm could you also add those options to the CMake build system since most OMR CI builds will not be impacted by the current change as CMake is the default. Once #3166 is merged that will be more obvious :)

charliegracie:
The change would be to add the option to the `omr/cmake/modules/toolcfg/gnu.cmake` I believe. @dnakamura can you provide some insight? Also if we add it to the gnu.cmake (or another global location) you can remove the GC specific addition of it in `omr/gc/CMakeLists.txt`

charliegracie:
Also we should think about adding the `xlC` option for disabling it as well.

pdbain-ibm:
It looks like the option is already in the cmake files for GCC (cf. : runtime/cmake/platform.cmake).
Am I missing something (besides xlc)?

>     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread -O3 -fno-strict-aliasing")
>     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -O3 -fno-strict-aliasing -fno-exceptions -fno-rtti -fno-threadsafe-statics")
>

charliegracie:
`platform.cmake` is an OpenJ9 file so that will not resolve issues in OMR for its general use.

pdbain-ibm:
Pull request updated (but still tentative).  @dnakamura would you please review?

gc/CMakeLists.txt
```diff
@@ -323,7 +323,7 @@ if(OMR_ENHANCED_WARNINGS)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
```
youngar: We can delete this whole block of cmake, the xlc option `-qalias=noansi` is set for all files in [xlc.cmake ](https://github.com/eclipse/omr/blob/589f93bb883463c3c3444d58710826cb7c81d15e/cmake/modules/platform/toolcfg/xlc.cmake#L28)

omrmakefiles/rules.linux.mk
```diff
@@ -187,21 +187,25 @@ ifeq (x86,$(OMR_HOST_ARCH))
         GLOBAL_CXXFLAGS+=-m32 -msse2 -I/usr/include/nptl
         GLOBAL_CPPFLAGS+=-DJ9X86
     endif
+    GLOBAL_CFLAGS+=-fno-strict-aliasing
+    GLOBAL_CXXFLAGS+=-fno-strict-aliasing
 
 else ifeq (aarch64,$(OMR_HOST_ARCH))
-    GLOBAL_CFLAGS+=-march=armv8-a+simd -Wno-unused-but-set-variable
```
youngar: I think it would be better to set this flag any time we're using gcc
```
ifeq (gcc,$(OMR_TOOLCHAIN))
  GLOBAL_CFLAGS+= -fno-strict-aliasing
  GLOBAL_CXXFLAGS += -fno-strict-aliasting
endif
```

pdbain-ibm:
I updated the cmake files.  Would you like me to run a personal build to test them?

youngar:
@genie-omr build all

pdbain-ibm:
Build failure, still needs work: `../../omrmakefiles/rules.linux.mk:47: *** invalid syntax in conditional.`
Fixing...

youngar:
@genie-omr build all

Issue #3145, opened 2018-11-01
Bad type punning in port/unix/omrtty.c:174


rwy0717:
```
../port/unix/omrtty.c: In function 'omrtty_available':
../port/unix/omrtty.c:174:3: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
   return *(int32_t *)&avail;
   ^
```

rwy0717:
The comment in this function claims that sometimes, FIONREAD can write out a 64bit integer, depending on the device.  AFAICT from public documentation, FIONREAD always takes an `int*` (which is typically a 32bit datatype). There should be no need to cast like this, we should be able to just use an int to represent `avail`. Does anyone have a counter example?

FIONREAD takes an int:
* http://man7.org/linux/man-pages/man2/ioctl_list.2.html
* https://www.freebsd.org/cgi/man.cgi?query=ioctl&sektion=2
* https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxbd00/ioctl.htm
* https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.basetrf1/ioctl32.htm
* https://docs.oracle.com/cd/E19683-01/806-6546/kermes8-28/index.html

PR #3187, opened 2018-11-13
Use an int to store the output of ioctl FIONREAD

rwy0717:
All publicly available documentation indicates that FIONREAD outputs an
int, not a pointer-width value. Use an int, and avoid bad aliasing.
There is a comment indicating that _some_ devices output a 64 bit value,
but I cannot see how or when that would ever be true. This PR removes
that comment.

Fixes: #3145

### DIFF
--- a/port/unix/omrtty.c
+++ b/port/unix/omrtty.c
@@ -154,7 +154,7 @@ omrtty_available(struct OMRPortLibrary *portLibrary)
 	/* PR94604 */
 	int rc;
 	off_t curr, end;
-	intptr_t avail = 0;
+	int avail = 0;
 
 	/* when redirected from a file */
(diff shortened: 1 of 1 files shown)


rwy0717:
@pdbain-ibm There's a comment here that says some devices may output a 64bit value on an FIONREAD ioctl. As far as I can tell, that's not true, it's always an int. Do you have any idea when this could be true, or where this comment came from?

There are some documentation links to the original issue #3145 .

Thanks

pdbain-ibm:
I don't know where the comment came from; it looks very old, given the PR number.

rwy0717:
hehe yeah, I wonder if it was some old cell phone or something. Does this change look OK to you @pdbain-ibm ?

pdbain-ibm:
Have you tested it on some BE platforms like AIX or z/OS?

pdbain-ibm:
I would also check with @gacholio .

youngar:
@genie-omr build all

youngar:
@genie-omr build all

PR #3189, opened 2018-11-13
Rework extensible enums

dnakamura:
Currently, the way extensible enum files are named, requires having

my_project/compiler/Foo.enum:
```
#include "path_to_omr/compiler/Foo.enum"

SOME_ENUM_VAL
OTHER_ENUM_VAL
...
```

However, this approach breaks if the path to omr sources ever changes.  Instead this PR restructures
extensible enums to work the same way as extensible classes. ie

my_project/compiler/Foo.enum:
```
#include "compiler/OMRFoo.enum"
...
```

### DIFF
--- a/compiler/codegen/LinkageConventions.enum
+++ b/compiler/codegen/LinkageConventions.enum
@@ -1,0 +1,22 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "codegen/OMRLinkageConventions.enum"
--- a/compiler/optimizer/DataFlowAnalysis.enum
+++ b/compiler/optimizer/DataFlowAnalysis.enum
@@ -1,0 +1,22 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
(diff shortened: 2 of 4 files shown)


Leonardo2718:
@genie-omr build all

compiler/codegen/LinkageConventions.enum
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2018, 2018 IBM Corp. and others
```
charliegracie: is this correct?

compiler/optimizer/DataFlowAnalysis.enum
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2018, 2018 IBM Corp. and others
```
charliegracie: again I am wondering why the initial copyright date was changed?

dnakamura: git didnt really track the changes properly. DataFlowAnalysis,enum was renamed to OMRDataFlowAnalysis.enum, and  a new file was created with the old name. I was hoping that by keeping the rename in one commit, and the creation of the new file in a separate commit, git would properly track the file move, but that seems to have failed

dnakamura:
@Leonardo2718 Can you point to an instance where the enum is in a .hpp file? I'm hesitant to make that change since its not really valid as a standalone header. From what I can tell they are designed to be used like so:

SomeEnum.hpp:
```
enum SomeEnum{
#include "SomeEnum.enum"
};
```

Leonardo2718:
@dnakamura 

That usage is correct. Other examples of extensible enums that `.hpp` are:

- `compiler/il/OMRILOpCodesEnum.hpp`
- `compiler/il/DataTypesEnum.hpp`
- `compiler/codegen/OMRRecognizedMethodsEnum.hpp`
- `compiler/codegen/OMRInstructionFlagEnum.hpp`
- `compiler/optimizer/OMRSimplifierTableEnum.hpp `
- `compiler/{p,z,arm,aarch64,x}/codegen/OMRInstructionKindEnum.hpp`'

They're fairly easy to find by grepping for files with `Enum` in their name.

`.hpp` may not be the best extension but I think it's the most used. If we decide to change it to something else later on, it should be easy enough to change it on all the files.

charliegracie:
Are there other *.enum files already?  To be clear @dnakamura did not create the `.enum` files but rather just changed their names to make it easier to work with.

Leonardo2718:
> Are there other *.enum files already? To be clear @dnakamura did not create the .enum files but rather just changed their names to make it easier to work with.

The files modified in this PR are the only ones I know of that use `.enum`. My thinking was that since the files are being renamed to follow the pattern of other extensible enums (and classes), also changing the file extensions so they match makes sense.

That said, I do see why using `.enum` makes sense. If others don't see value in changing the extension, I'll accept the changes as they currently are.

charliegracie:
ping @Leonardo2718

Leonardo2718:
As I mentioned in #3213, I would like to see an umbrella issue open to track why these changes are needed.

The extensible enum mechanism has historically been the source of much discussion so I want to make sure we're tracking the reasons for changing their structure.

0xdaryl:
My preference going forward is to not use `.hpp` file extensions for entities that are used to compose enums and static tables.  The reason is they are not complete header files and can only be included in very carefully guarded circumstances (e.g., within the braces of an enum).  A different file extension makes that clear.

I support the renaming to an `.enum` extension for the enum files and think it should proliferate throughout OMR.  I also think we need a different extension for the static table files (e.g., `.table` or `.init`).

charliegracie:
@dnakamura ping?

dnakamura:
#3328

dnakamura:
@charliegracie I believe opening an umbrella issue was the only outstanding request on this PR. Feel free to correct me if I am mistaken

PR #3190, opened 2018-11-14
Remove deprecated J9 snippets on X86 [Phase 1/2]

0dvictor:
Following J9-specific snippets are deprecated, cleaning up.
1. IA32WriteBarrierSnippet
2. AMD64WriteBarrierSnippet
3. X86JNIPauseSnippet
4. X86PassJNINullSnippet

Note that this commit only removes all OMR usage of these snippets,
a subsequent commit will remove them completely.


andrewcraik:
Since these snippets are being cleaned up through their having been re-written as out of line sequences in OpenJ9 getting them out of OMR makes sense. I'll leave the detailed review to @0xdaryl since he is following the various x86 clean-ups happening in OMR.

0xdaryl:
@genie-omr build all

Issue #3089, opened 2018-10-18
Investigate Scavenger's Aliasing Inhibiting Condition


RSalman:
### Summary
The goal of this investigation is to relax Scavenger's aliasing inhibiting condition _(mechanism used to improve parallelism & GC pause time)_  to increase object locality and ultimately improve application throughput. From Scavenger's aliasing perspective, object locality and parallelism/pause times are inversely proportional. A gain in one will cause a loss in the other, however throughput gains can result from improvements in either. As a result, with correct tuning between the two, optimal throughput gains can be achieved. That is, by tuning aliasing activity, throughput gains can be achieved in two ways:

1. Increasing **parallelism** (results in a decrease in **object locality**)
2. Increasing **object locality** (results in a decrease in **parallelism**)

**By relaxing Scavenger's aliasing inhibiting condition, a better  "equilibrium" between the the two can achieved for optimal throughput gains.** 

_A background has been provided to outline work that's been done in the past which relates to the investigation discussed in this issue._
### Background
In the context of Parallel Scavenger, a thread is said to be stalled if it has no caches to scan. In other words, a thread is left ideal and "starved" when it has exhausted all work. These stalled threads are accounted for by  Scavenger's `_waitingCount` member, which is incremented for each stalled thread.

In the past, an increase in mean GC pause times (Approx. 18%) was observed, it was found to be related to a decrease in parallelism resulting from stalled threads. Overall, this was determined to be a consequence of hierarchical scan aliasing which forces threads to "consume their own produce". Threads with scan work resulting in greater number of cache copies continued running while other threads with less copy work eventually stalled. As a result, threads were stalled for a proportion of GC, parallelism dropped and pause times increased.

To alleviate thread stalling, work sharing was introduced by making aliasing conditional. This would act as a mechanism to inhabit aliasing  and release newly copied work when threads were found to be stalled _(i.e., wait count is greater than 0)_, as outlined in the following code: 

```C++
MMINLINE MM_CopyScanCacheStandard *
MM_Scavenger::aliasToCopyCache(MM_EnvironmentStandard *env, GC_SlotObject *scannedSlot, MM_CopyScanCacheStandard* scanCache, MM_CopyScanCacheStandard* copyCache)
{
	/* Only alias a copy cache IF there are 0 threads waiting.  If the current thread is the only producer and
	 * it aliases a copy cache then it will be the only thread able to consume.  This will alleviate the stalling issues
	 * described in VMDESIGN 1359.
	 *
[...]

PR #3194, opened 2018-11-14
GC: Relax Scavenger's Aliasing Inhibiting Condition

RSalman:
waitingCountAliasThreshold introduced to relax copy cache alias inhibiting - relaxed to 20% from 0% threads that are allowed to be stalled.

See https://github.com/eclipse/omr/issues/3089

### DIFF
--- a/gc/base/Dispatcher.cpp
+++ b/gc/base/Dispatcher.cpp
@@ -89,8 +89,9 @@ MM_Dispatcher::cleanupAfterTask(MM_EnvironmentBase *env)
 void
 MM_Dispatcher::run(MM_EnvironmentBase *env, MM_Task *task, uintptr_t newThreadCount)
 {
+	uintptr_t activeThreads = recomputeActiveThreadCount(env, task, newThreadCount);
 	task->masterSetup(env);
-	prepareThreadsForTask(env, task, newThreadCount);
+	prepareThreadsForTask(env, task, activeThreads);
 	acceptTask(env);
 	task->run(env);
 	completeTask(env);
@@ -107,4 +108,10 @@ MM_Dispatcher::startUpThreads()
 void 
 MM_Dispatcher::shutDownThreads() 
 {
+}
+
+uintptr_t
+MM_Dispatcher::recomputeActiveThreadCount(MM_EnvironmentBase *env, MM_Task *task, uintptr_t newThreadCount)
+{
+	return 1;
 }
--- a/gc/base/standard/ParallelScavengeTask.cpp
+++ b/gc/base/standard/ParallelScavengeTask.cpp
@@ -39,6 +39,13 @@ MM_ParallelScavengeTask::run(MM_EnvironmentBase *envBase)
 }
 
 void
+MM_ParallelScavengeTask::masterSetup(MM_EnvironmentBase *env)
+{
+	uintptr_t calculatedAliasThreshold = (uintptr_t)(getThreadCount() * env->getExtensions()->aliasInhibitingThresholdPercentage);
+	_collector->setAliasThreshold(calculatedAliasThreshold);
+}
+
+void
 MM_ParallelScavengeTask::setup(MM_EnvironmentBase *env)
 {
 	if (env->isMasterThread()) {
--- a/gc/base/Dispatcher.hpp
+++ b/gc/base/Dispatcher.hpp
@@ -60,6 +60,8 @@ protected:
 	virtual void completeTask(MM_EnvironmentBase *env);
 	virtual void cleanupAfterTask(MM_EnvironmentBase *env);
 
+	virtual uintptr_t recomputeActiveThreadCount(MM_EnvironmentBase *env, MM_Task *task, uintptr_t newThreadCount);
+
 public:
 	static MM_Dispatcher *newInstance(MM_EnvironmentBase *env);
 	virtual void kill(MM_EnvironmentBase *env);
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -452,6 +452,7 @@ public:
 	double dnssMinimumExpansion;
 	double dnssMinimumContraction;
 	bool enableSplitHeap; /**< true if we are using gencon with -Xgc:splitheap (we will fail to boostrap if we can't allocate both ranges) */
+	double aliasInhibitingThresholdPercentage; /**< percentage of threads that can be blocked before copy cache aliasing is inhibited (set through aliasInhibitingThresholdPercentage=) */
 
 	enum HeapInitializationSplitHeapSection {
 		HEAP_INITIALIZATION_SPLIT_HEAP_UNKNOWN = 0,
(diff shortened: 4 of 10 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@RSalman and @amicic if this is ready can you remove the WIP and I will get it merged

RSalman:
Ready to be merged!

gc/base/standard/ParallelScavengeTask.cpp
```diff
@@ -41,6 +41,14 @@ MM_ParallelScavengeTask::run(MM_EnvironmentBase *envBase)
 void
 MM_ParallelScavengeTask::setup(MM_EnvironmentBase *env)
 {
+	/* The alias threshold must be set before the scavenge tasks start but after the number of active threads
```
charliegracie: The `Task::masterSetup(MM_EvnironmentBase *env)` call is likely the right place to put this functionality.

In `Task.hpp` see the following function description:
```
	/**
	 * Single call setup routine for tasks invoked by the master thread before the task is dispatched.
	 */
	virtual void masterSetup(MM_EnvironmentBase *env);
```

The `masterSetup` function is called in `Dispachter::run` before `prepareThreadsForTask` is called which means the number of threads that will be active in this GC bu after the override of the thread count set a few lines above. I think some slight refactoring of this code could clean things up and make sure the new code you are introducing is only being called once per GC

amicic: There was actually intention to call this after prepareThreadsForTask which can change the number of actual threads used by the task.

There is no place after prepareThreadsForTask  that will ensure that any code will be executed by only one thread, but still before any of GC threads start doing work (without some extra tricks, like using atomics etc.)

ktbriggs:
@RSalman @charliegracie @amicic Has any benchmarking been done to demonstrate improvement with this change? Note that to get an accurate picture of stalling behavior you need to use the wait time stats recorded in global MM_ScavengerStats:

```
uint64_t scavengeMicros = omrtime_hires_delta(_extensions->incrementScavengerStats._startTime, 
	_extensions->incrementScavengerStats._endTime, OMRPORT_TIME_DELTA_IN_MICROSECONDS);
omrtty_printf("%5llu      : idle time; %llu %llu %llu %llu %llu %llu %llu\n", stats->_gcCount,
	stats->_workStallCount, stats->_syncStallCount, stats->_completeStallCount,
	omrtime_hires_delta(0, stats->_workStallTime, OMRPORT_TIME_DELTA_IN_MICROSECONDS),	
	omrtime_hires_delta(0, stats->_syncStallTime, OMRPORT_TIME_DELTA_IN_MICROSECONDS),
	omrtime_hires_delta(0, stats->_completeStallTime, OMRPORT_TIME_DELTA_IN_MICROSECONDS),
	scavengeMicros);
```

The stats reported in vgc logs are bogus as they count monitor spin time as user time and charge idle CPU time to the running process if not taken by other process. They almost always come out as >98% CPU utilization.

charliegracie:
@ktbriggs the stats in verbose:gc are not bogus.  They are reporting exactly what the were meant to report. The values in verbose:gc are the overall CPU cycles being used as reported by the proper OS APIs. As you suggest from your comment those are not the number you are looking for and that is why we have other numbers. The numbers in verbose:gc are very very useful for debugging issues and describing how much CPU is being consumed.

charliegracie:
@RSalman do you think it will be possible to refactor the code as per my [comment](https://github.com/eclipse/omr/pull/3194#discussion_r235043218).  If so, how much work do you predict it will be? If the work is going to be too significant then we can open an Issue to re-work the calculation for the number of GC threads that take part in a task and as part of that task we could switch your new code into the proper `masterSetup` call.  I would prefer to make the change now but I do not want to bottleneck your improvements.

gc/base/standard/ParallelScavengeTask.cpp
amicic: as for refactoring.... are you suggesting to extract recomputeActiveThreadCount (and few more lines around it) from prepareThreadsForTask  and call it just before masterSetup (still within Dispatcher::run() sequence)?

charliegracie: Something like that... and it would be great if the recomputeActiveThreadCount used the value passed into the task setup that was added recently so there is only one place that thread count is changed.

RSalman:
#### Refactored

`recomputeActiveThreadCount` is moved out from  `prepareThreadsForTask`. It's now called first in   `dispatcher::run` followed by `masterSetup` which now consumes the active thread count and does the threshold calculation. The active thread count is then passed to `prepareThreadsForTask` to wake up the threads.

charliegracie:
@genie-omr build all

gc/base/Dispatcher.cpp
```diff
@@ -108,3 +110,9 @@ void
 MM_Dispatcher::shutDownThreads() 
 {
 }
+
+uintptr_t
+MM_Dispatcher::recomputeActiveThreadCount(MM_EnvironmentBase *env, MM_Task *task, uintptr_t newThreadCount){
+	Assert_MM_unreachable();
```
charliegracie: This should not assert.  IF a GC was not using the parallel Dispatcher then it has to fall to the Dispatcher version.  I am fine with this just returning 1 with no assert.

charliegracie:
@amicic thoughts?

gc/base/Dispatcher.cpp
RSalman: It would be more correct to just return `_activeThreadCount` rather than 1. But that's not possible without another rework.... As Alek mentioned no collector instantiates dispatcher it self, so the little implementation that Dispatcher has can be moved to PD.

Most of the common dispatcher code seems to be in PD anyways, such as
```C++
uintptr_t _threadCountMaximum; /**< maximum threadcount - this is the size of the thread tables etc */
uintptr_t _threadCount; /**< number of threads currently forked */
uintptr_t _activeThreadCount; /**< number of threads actively running a task */
```
I don't thin Dispatcher should even have any idea of `recomputeActiveThreadCount` when it doesn't have the `_activeThreadCount` info. However, it had to be added here since Dispatcher run method is used by PD.

charliegracie: Agreed.  The change is ok as-is.  The dispatcher / parallel dispatcher should be re-worked in a separate PR.  Thanks.

RSalman: I think a larger rework is required in general to clean all this up, an issue beyond what this PR addresses. A separate issue should be opened to do a larger refactor/cleanup in addition to what Alek pointed out below

RSalman: I'll make the change to remove the Assert and return 1... it's more consistent with the other Dispatcher methods 

 `dispatcher.hpp`
```C++
	MMINLINE virtual uintptr_t threadCount() { return 1; }
	MMINLINE virtual uintptr_t threadCountMaximum() { return 1; }
	MMINLINE virtual uintptr_t activeThreadCount() { return 1; }
```

charliegracie:
@genie-omr build all

PR #2822, opened 2018-08-02
Enable SSE/SSE2 on OMR X86-32

0dvictor:
All known downstream projects uses SSE/SSE2 for floating point operations;
however OMR itself still kept X87 as default on X86-32.
This commit turns SSE/SSE2 on as default, allowing OMR's tests exercising SSE/SSE2 route.
X87 support is deprecated, and will be removed.

Closes #2429

Signed-off-by: Victor Ding <dvictor@ca.ibm.com>

Issue #2645, opened 2018-06-07
Consolidate implementations of RegisterIterator


0xdaryl:
Each code generator implements a `RegisterIterator` class which appears to be implemented solely to assist register assignment tracing dumps in `ras/Debug.cpp`.  The DNA of this class across all the code generators must be at least 97% identical, and I have no doubt that with a bit of thought a single, common implementation shared by all code generators could be developed.

Perhaps the Machine class itself should provide an iterator that is initialized upon request for a particular register kind.

0xdaryl:
It's also not clear to me why this was made an extensible class in the first place.  Its hardly a class I would think one would want to extend.

samolisov:
@0xdaryl there is a `TR_X86FPStackIterator` class defined inside `compiler/x/codegen/OMRCodeGenerator.hpp', the class is derived from `TR::RegisterIterator` and (as I assume) uses the dynamic polymorphism (overrides all virtual methods in `OMR::X86::RegisterIterator`. By the way, this design forces the constructor of `OMR::X86::RegisterIterator` to check whether the `kind` equals to `TR_X87` otherwise the assert in the last else block takes place (so the `traceRA={states}` doesn't work on X).

I've added the check but haven't opened the PR because I wish to create a unified version of the RegisterIterator. The question I have is about polymorphism: may I make `OMR::RegisterIterator` (or  `TR::RegisterIterator`?) not extensible but keep the virtual methods inside and just define two derived classes, for instance, `OMR::RegiterFileIterator` (the iterator over general, MMX, and so on registers) and a platform-specific `TR_X86FPStackIterator` (or `OMR::X86::FPStackIterator`?, the iterator over X87 register stack) but this solution brings the dynamic polymorphism to the compiler.

What is your (and other compiler gurus') opinion?

samolisov:
As I know, the x87 support is deprecated in OMR, so we can just remove `TR_X86FPStackIterator` from the codebase.

samolisov:
@0xdaryl Thank you for the comment. I was busy a bit renaming all architecture-specific `getXXXRealRegister` into `getRealRegister` but now am ready to move the task forward.

I wish to remove the switch over `TR_RegisterKinds` from the constructor: not all architectures support the full set of register kinds, for example, there is no `TR_HPR` for X86. If we take `TR_RegisterKinds` into account, we will have to define a static method for each kind of registers in a corresponding `RealRegister` class. 

[...]

PR #3200, opened 2018-11-15
Remove TR_X86FPStackIterator (X87 FP) class

samolisov:
As it is discussed in #2822, X87 support is deprecated, and will be
removed. This fact lets us the opportunity to remove
'TR_X86FPStackIterator' from the codebase as the first step to
consolidate implementations of RegisterIterator.

Issue: #2645


0xdaryl:
@genie-omr build xlinux,win

PR #3203, opened 2018-11-15
Fix build break

r30shah:
When node is set to be backwardArrayCopy node, we generate backward array copy sequence only. When this is combined with constant length bytelength we end up with NULL byteLenReg which causes segmentation fault.

### DIFF
--- a/compiler/z/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/z/codegen/OMRTreeEvaluator.cpp
@@ -13483,7 +13483,7 @@ OMR::Z::TreeEvaluator::forwardArrayCopySequenceGenerator(TR::Node *node, TR::Cod
       // 2. byteLength > 256 && byteLength is not multiple of 256
       uint8_t residueLength = static_cast<uint8_t>(byteLen & 0xFF);
       if (byteLen <= 256 || residueLength != 0)
-         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, residueLength-1, 
+         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, static_cast<uint8_t>(residueLength-1), 
             generateS390MemoryReference(byteDstReg, 0, cg),
             generateS390MemoryReference(byteSrcReg, 0, cg)); 
       }
@@ -13656,6 +13656,8 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
       }
    else if (node->isBackwardArrayCopy())
       {
+      if (byteLenReg == NULL)
+         byteLenReg = cg->gprClobberEvaluate(byteLenNode);
       deps = TR::TreeEvaluator::backwardArrayCopySequenceGenerator(node, cg, byteSrcReg, byteDstReg, byteLenReg, byteLenNode, srm, mergeLabel);
       }
    else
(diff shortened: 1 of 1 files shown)


r30shah:
@fjeremic  I am waiting for test to get finished but this change should fix https://github.com/eclipse/openj9/issues/3688

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -13483,7 +13483,7 @@ OMR::Z::TreeEvaluator::forwardArrayCopySequenceGenerator(TR::Node *node, TR::Cod
       // 2. byteLength > 256 && byteLength is not multiple of 256
       uint8_t residueLength = static_cast<uint8_t>(byteLen & 0xFF);
       if (byteLen <= 256 || residueLength != 0)
-         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, residueLength-1, 
+         generateSS1Instruction(cg, TR::InstOpCode::MVC, node, static_cast<uint8_t>(residueLength-1),
```
r30shah: @fjeremic `generateSS1Instruction` expects 16 bit unsigned integer argument. When we have constant byte length without explicit casting it sends `65535`  instead of `255` to generateSS1Instruction. Just explicitly cast value here without speculating binary encoding generate current value for us.

fjeremic: Not sure I'm following here. Why does explicit casting here change anything? `residueLength` is already a `uint8_t`, and hence so is `residueLength - 1` so the cast doesn't change the answer at all I would think.

fjeremic:
I'd also like to see a more descriptive commit subject line. Sure I can tell it's fixing a build break but on what?

compiler/z/codegen/OMRTreeEvaluator.cpp
r30shah: Looking at the function declaration 
https://github.com/eclipse/omr/blob/c2ec9f716193f5c995ee22619ebedf27bdab1d48/compiler/z/codegen/S390GenerateInstructions.hpp#L647
 In case when `residueLength-1` will be casted to uint16_t. So is `byteLen` is multiple of `256` means `residueLength` will be `0`. In this case here our intended argument is `255` but it will send `65535` instead. 
Verified this in the instruction generation time : 
```
[0x787909c0]   MVC     #638 0(65536,GPR_0044),#639 0(GPR_0042)
```

However it does generate correct binary encoding but looking at the Instruction Generation it is misleading. I took a look at `SS-a` type instructions. I do not see any one uses 16 bit length. Is it possible to clean up SS1 instruction generator to reflect this?

r30shah:
@fjeremic Changed the Commit title. Also explained my reasons regarding to explicit static casting while passing length for MVC instruction. Local tests passes.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: I see. Thanks for the explanation. We should avoid these headaches and modify the `generateSS1Instruction` API to take a `uint8_t` since the `L` field of the SS encoding is 8 bits wide and unsigned. Not sure why we even give the option to the user to shoot themselves in the foot with the `uint16_t` parameter type. Can you open up an item for this in OMR and on our ZenHub?

fjeremic:
@genie-omr build zlinux,zos

PR #3205, opened 2018-11-16
Concurrent Scavenger yield

amicic:
If there is a request to yield, abort the concurrent scan phase. The
rest of the work will be done in final STW phase.

There could be an arbitrary criteria, which this change does not impose;
it just provides API to check if yield is necessary.

An example of a criteria is raised exclusive VM access request (which
could be language specific). In such a case it is fair to yield and
don't have a third party to wait 100s of milisec before the concurrent
phase completes.

Another example would be exhausted hybrid survivor/allocate space.
Application cannot make progress anyway, so it's better to complete the
cycle in the final STW phase with all GC threads available (by default
4x more than number of threads in concurrent phase).

Yield checks are still relatively coarse - they are done before fetching
next work unit (scan cash), which in a bad case may still take a few
milisec to respond to yield request.


amicic:
Related, but not dependent on: https://github.com/eclipse/openj9/pull/3711

charliegracie:
@genie-omr build all

PR #3209, opened 2018-11-19
DDR: fix ddrgen bugs involving empty structs

dnakamura:
Fix a couple of bugs with ddrgen related to empty structs

* structs are considered based on the total number of constants/fields
seen globally rather than the number of fields/constants seen in that
struct

* Empty structs are still actually output, even when -e is not specified.
This leads to an assertion failure because the number of structs output
is greater than _buildInfo.header.structureCount

* _buildInfo.header.structDataSize is incremented for empty structs which
should not be output.


charliegracie:
@genie-omr build all

PR #3213, opened 2018-11-19
Rename extensible header files in the the compiler

dnakamura:
A number of header files in the compiler are designed to be extened by
the consuming runtime. For these headers we prefix the names with OMR
and provide a new default implementation of the header which simply
includes the OMR prefixed header. (this is in line with existing
extensible headers)


charliegracie:
The Lint build breaks with this change

0xdaryl:
Your commit titles have a double "the".  Please fix.

Leonardo2718:
Changes LGTM.

Is there a plan to make similar changes to the corresponding files in downstream projects (e.g. OpenJ9)?

dnakamura:
@Leonardo2718  Yes plan is to make changes to downstream projects. Changes are being made so we can do out of tree builds properly (where OMR isnt a sub-directory of consuming project)

Leonardo2718:
@genie-omr build all

charliegracie:
ping @dnakamura and @Leonardo2718

dnakamura:
@0xdaryl I saw downstream consumer extending it in the same manner as the normal extensible headers. I assumed that they were relying on the fact that the omr headers will include "Runtime.hpp" and get their copy (which then includes the omr copy), but I will do a bit more digging to see if that is the case

charliegracie:
@dnakamura ping?

dnakamura:
For some reason I thought this pr was only dealing with Runtime.hpp which is why i closed it, however there are other useful changes here

compiler/il/ILProps.hpp
```diff
@@ -18,262 +18,4 @@
...
-      // Available                = 0x08000000,
-      // Available                = 0x10000000,
-      // Available                = 0x20000000,
-      // Available                = 0x40000000,
-      // Available                = 0x80000000,
-      };
-   }
-
-#endif
+#include "compiler/il/OMRILProps.hpp"
```
Leonardo2718: In extensible classes, we normally do not use "compiler/" at the start of includes (there is one exception but it doesn't apply in this case):

```suggestion
#include "il/OMRILProps.hpp"
```

Is there a reason for needing them here?

dnakamura: just an oversight on my part

Leonardo2718:
` compiler/aarch64/codegen/TreeEvaluatorTable.hpp` and ` compiler/arm/codegen/TreeEvaluatorTable.hpp` still have `compiler/` in the includes.

dnakamura:
@0xdaryl poke

Leonardo2718:
@genie-omr build all

DanHeidinga:
Whats the next step to get this merged?  I see the builds have been run and succeeded.

dnakamura:
@DanHeidinga  I believe its just waiting for someone to click the merge button. I believe I have addressed everyone's issues

Leonardo2718:
All comments have been addressed. Everyone who reviewed has approved. All checks have passed. Merging.

PR #3216, opened 2018-11-20
WIP: More strict condition for isOSRInductionHelper

cathyzhyi:
The old condition is not sufficient to tell if the method is induceOSR
helper since the reference index for normal methods and helper methods
can overlap.


vijaysun-omr:
@genie-omr build all

cathyzhyi:
the reference number should be unique and this is not the cause of the intermittent failure. closing...

PR #3221, opened 2018-11-20
cmake: Add OmrOption, a util for declaring build options

rwy0717:
OmrOption is a module which provides a declarative syntax for specifying
and validating ON/OFF build options. The long-term goal is to declare
and validate all OMR build options using this module.

The general syntax is:
```
  omr_add_option_group(<group>)

  omr_add_option(<option> <option-group>
    COMMENT <comment>
    DEFAULT <expression>
    REQUIRES <option>...
    REQUIRES_EXPR <expr>
    CONFLICTS <option>...
  )

  omr_validate_option_group(<group>)
```
Options are defined as BOOL cache variables.

DEFAULT is an expression, which is evaluated and coerced to an on/off
value. The default value is only evaluated on the first run, and cached
in subsequent runs. Since options are regular cmake cache variables, the
value can be manually overridden by users at any point. A default value
must always be provided.

REQUIRES allows users to pass a list of required flags or variables,
which must be ON if the option is ON.

Similarly, CONFLICTS is a list of options which must be OFF, if the
option is ON.

REQUIRES_EXPR is an expression which must be TRUE, if the option is
ON. This can be used for expressing more complex option requirements.
Each option can have only one associated REQUIRES_EXPR.

Options are registered into a group, which must be explicitly validated
with a call to omr_validate_option_group. Option groups allow consuming
projects to create and validate their own option sets. By making
validation explicit, we allow the user (us OMR developers) to delay
validation until after all configuration variables have been set.
Options are validated on every run of cmake, which ensures that the user
never puts OMR into an illegal configuration.

The phrasing "option" was chosen over "flag" to follow existing cmake
conventions, where add_option / add_dependent_option create ON/OFF
bool cache variables.

There is a similar module built in to cmake called DependentOption.
This module was not used because:
1. A dependent option is only show to the user if it's dependencies are
   satisfied. OmrOption makes the choice to always present ALL flags in
   the cache.
2. A dependent option will revert to it's default value if it's
   requirements aren't satisfied, overriding any user provided value.
   OmrOption never tries to override the user, and relies on option
   validation instead.
3. OmrOption provides an easier to read, more obvious syntax.
4. The default value in OmrOption is an expression, rather than a simple
   on/off constant. This gives us a succinct way to implement implicitly
   enabled/disabled flags, while still being user overridable.
   Dependent options have a dependency expression, but the default value
   is a constant, which cannot be overridden by the user.


rwy0717:
(updated some comments)

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+#
+#  omr_add_option(JS_NUNBOXED_VALUES JS_OPTIONS
+#    COMMENT "Encode properties as double immediates, exploiting the format of NaNs to embed additional types."
+#    DEFAULT NOT JS_LOW_TAGGED_INTEGERS
+#    REQUIRES OMR_ENV64 IEEE_754
+#    CONFLICTS JS_LOW_TAGGED_INTEGERS
+#    REQUIRES_EXPR OMR_HOST_CPU IN_LIST "X86" "x86_64"
+#  )
+#
+#  omr_validate_option_group(JS_OPTIONS)
```
Leonardo2718: I think some of the content from your PR description would be good to also have here (especially the comparison with `DependentOption`) so that we don't need to look back at the PR to find that of information in the future.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	omr_assert(
+		TEST is_option_group
+		MESSAGE "${group} is not an options group"
+	)
+endfunction(omr_assert_is_option_group)
+
+# omr_group_options(<group> [<options>...])
+# Register a set of options into an options group. Options must have already been declared.
+# see also: omr_declare_option, omr_add_option_group, omr_add_option.
+function(omr_group_options group)
```
Leonardo2718: Can you add some guidelines for how this function should be used? It's not clear to me whether this is only intended for use in the implementation of this utility or for general, "public" use. If the latter, it might also be useful to add a short example.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	omr_assert_is_option_group(${group})
+	set_property(GLOBAL APPEND PROPERTY ${group} ${ARGN})
+endfunction(omr_group_options)
+
+# omr_get_options_group(<output> <group>)
+# Store the list of options in <group> into the variable <output>.
+function(omr_get_option_group output group)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	omr_assert_is_option_group(${group})
+	get_property(${output} GLOBAL PROPERTY ${group})
+	set(${output} ${${output}} PARENT_SCOPE)
+endfunction()
+
+# omr_declare_option(<option>)
+# Declare a new option. Defines the three properties, option_REQUIRES, option_CONFLICTS, and option_REQUIRES_EXPR.
+# See also: omr_add_option.
+function(omr_declare_option option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	define_property(GLOBAL PROPERTY ${option}_REQUIRES_EXPR
+		BRIEF_DOCS "An expression which must be true if ${option} is enabled."
+		FULL_DOCS " "
+	)
+endfunction(omr_declare_option)
+
+# omr_option_requires(<option> [<requirement>...])
+# Add requirements to <option>. If the variable named by option is true,
+# each variable named in the requirements must also be true.
+function(omr_option_requires option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	set_property(GLOBAL APPEND PROPERTY ${option}_REQUIRES ${ARGN})
+endfunction(omr_option_requires)
+
+# omr_option_conflicts(<option> [<conflict>...])
+# Add conflicts to <option>. If the variable named by option is true,
+# each variable named in the conflicts must not be true.
+function(omr_option_conflicts option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	set_property(GLOBAL APPEND PROPERTY ${option}_CONFLICTS ${ARGN})
+endfunction(omr_option_conflicts)
+
+# omr_option_requires_expr(<option> <expr>...)
+# Register an expr which must evaluate to true if option is enabled. Unfortunately, it is only possible to register
+# a single expr for each option. Calling this more than once for an option is an error.
+function(omr_option_requires_expr option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+	endif()
+
+	if(DEFINED ARG_REQUIRES_EXPR)
+		omr_option_requires_expr(${option} ${ARG_REQUIRES_EXPR})
+	endif()
+endfunction(omr_add_option)
+
+# omr_check_option_requirements(<option>)
+# Assert that every required option is enabled. See also: omr_add_option, omr_option_requires.
+function(omr_check_option_requirements option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+		omr_assert(
+			TEST ${requirement}
+			MESSAGE "${option} requires ${requirement}"
+		)
+	endforeach(requirement)
+endfunction(omr_check_option_requirements)
+
+# omr_check_option_conflicts(<option>)
+# Assert that every conflicting option is disabled. See also: omr_add_option, omr_option_conflicts.
+function(omr_check_option_conflicts option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+		omr_assert(
+			TEST NOT ${conflict}
+			MESSAGE "${option} conflicts with ${conflict}"
+		)
+	endforeach(conflict)
+endfunction(omr_check_option_conflicts)
+
+# omr_check_option_requirement_expr(<option>)
+# if the option has a REQUIRE_EXPR defined, assert that the expression is true.
+function(omr_check_option_requirement_expr option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,245 @@
...
+		omr_assert(
+			TEST ${expr}
+			MESSAGE "${option} REQUIRES_EXPR failed"
+		)
+	endif(expr_set)
+endfunction(omr_check_option_requirement_expr)
+
+# omr_validate_option(<option>)
+# Validate the dependencies of option.
+function(omr_validate_option option)
```
Leonardo2718: Same comment as above.

cmake/modules/OmrOption.cmake
Leonardo2718: Also, for completeness, can you add an example of actually reading one of the defined options?

rwy0717:
I was thinking about adding some tests, but I'm not sure how to actually implement them. I can look at using `add_test` to run cmake scripts, which can use omr_assert to check that the module is functioning correctly. However, it would be difficult to test invalid configurations--I don't know how to check that a script aborted in the "right" way.

Leonardo2718:
> However, it would be difficult to test invalid configurations--I don't know how to check that a script aborted in the "right" way.

Hm... yeah that does seem a bit tricky.

cmake/modules/OmrOption.cmake
rwy0717: For the most part, I wrote these functions to be publicly callable... although there are a few where it doesn't always make sense, like `omr_check_option_requirements `. There aren't any conventions in cmake for private functions, so I suppose the right thing to do is document the complete public API in the module's documentation comment. I can move the signature comments for public functions up into the module docs, if that works?

Leonardo2718: Sure, I think that works. Maybe also add an example usages, where it makes sense?

rwy0717:
changes:
1. added lots of documentation to the header
2. added a leading underscore to internal function names
3. allow `omr_add_option_group` to declare multiple groups at once
4. slightly modified error output
5. fixed bug in `omr_validate_option`: correctly check `if(${option})`, not `if(option)`.

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+  omr_option_requires(<option> [<requirement>...])
+
+Add requirements to ``<option>``. If the option is ``ON``, each variable
+named in the requirements must also be true.
+
+::
+  omr_option_conflicts(<option> [<conflict>...])
+
+Add conflicts to ``<option>``. If the option is ``ON``, each variable named
+in the conflicts must _not_ be true.
```
Leonardo2718: Would it be clearer to just say "must be false"?

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+    COMMENT <comment>
+    DEFAULT <expression>
+    REQUIRES <option>...
+    REQUIRES_EXPR <expr>
+    CONFLICTS <option>...
+  )
+
+Create a new option. The new option is a ``BOOL`` cache variable.
+The option will be registered in ``<group>``. When ``<group>`` is
+validated, this options requirements will be tested.
```
Leonardo2718: Should `options` be `option's`?

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+At this point, it is possible to register requirements on the option,
+put the option into a group, or validate the option (despite having no
+requirements).
+
+::
+  omr_group_options(<group> <option>...)
+
+Register a set of options into an options group. Options must have already
+been declared. It's possible (but not recommended) to register an option in
+to multiple groups.
```
Leonardo2718: "in to" &rarr; "into"

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+    message(STATUS "not encoding immediate values")
+  endif()
+
+Comparison to :module:`DependentOptions`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There is a similar module built in to cmake called DependentOption.
+This module was not used because:
+
+1. A dependent option is only show to the user if it's dependencies are
```
Leonardo2718:
```suggestion
1. A dependent option is only shown to the user if it's dependencies are
```

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+		MESSAGE "${group} is not an options group"
+	)
+endfunction(_omr_assert_is_option_group)
+
+function(omr_group_options group)
+	_omr_assert_is_option_group(${group})
+	set_property(GLOBAL APPEND PROPERTY ${group} ${ARGN})
+endfunction(omr_group_options)
+
+# omr_get_options_group(<output> <group>)
```
Leonardo2718:
```suggestion
# _omr_get_options_group(<output> <group>)
```

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+	if(DEFINED ARG_CONFLICTS)
+		omr_option_conflicts(${option} ${ARG_CONFLICTS})
+	endif()
+
+	if(DEFINED ARG_REQUIRES_EXPR)
+		omr_option_requires_expr(${option} ${ARG_REQUIRES_EXPR})
+	endif()
+endfunction(omr_add_option)
+
+# omr_check_option_requirements(<option>)
```
Leonardo2718:
```suggestion
# _omr_check_option_requirements(<option>)
```

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+	get_property(requirements GLOBAL PROPERTY ${option}_REQUIRES)
+	foreach(requirement IN LISTS requirements)
+		omr_assert(
+			TEST ${requirement}
+			MESSAGE "option ${option} requires ${requirement}"
+		)
+	endforeach(requirement)
+endfunction(_omr_check_option_requirements)
+
+# omr_check_option_conflicts(<option>)
```
Leonardo2718:
```suggestion
# _omr_check_option_conflicts(<option>)
```

cmake/modules/OmrOption.cmake
```diff
@@ -0,0 +1,372 @@
...
+	get_property(conflicts GLOBAL PROPERTY ${option}_CONFLICTS)
+	foreach(conflict IN LISTS conflicts)
+		omr_assert(
+			TEST NOT ${conflict}
+			MESSAGE "option ${option} conflicts with ${conflict}"
+		)
+	endforeach(conflict)
+endfunction(_omr_check_option_conflicts)
+
+# omr_check_option_requirement_expr(<option>)
```
Leonardo2718:
```suggestion
# _omr_check_option_requirement_expr(<option>)
```

PR #3222, opened 2018-11-20
cmake: add omr_base: a library defining basic configuration

rwy0717:
OMR uses include_directories and compile_definitions to set up some
basic, universal compile options. The problem is, these only apply to
targets defined in OMR, and will not propagate out to consuming
projects. This patch is a step towards making every OMR component export
all its needed options.

omr_base is an interface library that exports all the core configuration
that most OMR components need. This gives us a centralized location to
configure basic properties of the omr project. omr_base replaces any
usage of include_directories or compile_definitions.

Virtually every library in OMR should link (directly or indirectly)
against omr_base publicly. By doing so, we can count on these properties
correctly propagating out to consuming projects through linking, without
any additional boilerplate on the consumer's end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,16 +70,32 @@ include(cmake/config.cmake)
 
 enable_testing()
 
-include_directories(
-	${PROJECT_BINARY_DIR}
-	./include/
-	./include_core/
-	./third_party/
+###
+### omr_base: basic includes and definitions needed everywhere
+###
+
+add_library(omr_base INTERFACE)
+
+target_include_directories(omr_base
+	INTERFACE
+		$<BUILD_INTERFACE:${omr_SOURCE_DIR}/include_core>
+		$<BUILD_INTERFACE:${omr_BINARY_DIR}>
+		$<INSTALL_INTERFACE:${OMR_INSTALL_INC_DIR}>
 )
 
-add_definitions(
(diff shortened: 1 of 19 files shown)


Leonardo2718:
@genie-omr build all

rwy0717:
The issue is that object libraries can't link against omr_base, so I'll have to manually copy the properties over. It's funny, because I thought it would be an issue, but it was working fine locally. I'll fix this pronto.

charliegracie:
@genie-omr build all

rwy0717:
d'oh!

Leonardo2718:
@genie-omr build all

Leonardo2718:
Seems straight forward enough :slightly_smiling_face: 

The only thing that isn't quite clear to me is why some targets linke `omr_base` using `INTERFACE` or `PRIVATE` instead of `PUBLIC`. I would have thought that everything should should use `PUBLIC` in order to maintain the same behaviour as before. As I'm not familiar with the parts of the code that don't use `PUBLIC`, I'm just looking for confirmation that the use of `INTERFACE` and `PRIVATE` is correct/intentional in those particular cases.

rwy0717:
Most libraries are linking publicly. There are a few other _base libraries which are also interface libraries, so have to link against omr_base as INTERFACE. The exception is omrsig, which has no public interface or headers, so it links against omr_base privately.

PR #3223, opened 2018-11-21
Add induceOSRAfterAndRecompile API in OMR

cathyzhyi:
Add induceOSRAfterAndRecompile API to generate the tree for helper call
to trigger both OSR and recompilation.

The old API induceOSRAfter is still useful if only OSR transition
is needed.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -518,7 +518,10 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
    //there is no guarantee that the future block containing the induceOSR call still will have an exception
    //edge to the OSR catch block.
    TR::SymbolReferenceTable* symRefTab = self()->comp()->getSymRefTab();
-   TR::SymbolReference *induceOSRSymRef = symRefTab->findOrCreateRuntimeHelper(TR_induceOSRAtCurrentPC, true, true, true);
+   TR::SymbolReference *induceOSRSymRef = symRefTab->findOrCreateRuntimeHelper(TR_induceOSRAtCurrentPC,
+                                                                               true /* canGCandReturn */,
+                                                                               true /* canGCandExcept */,
+                                                                               true /* preservesAllRegisters */);
    // treat jitInduceOSR like an interpreted call so that each platform's codegen generate a snippet for it
    induceOSRSymRef->getSymbol()->getMethodSymbol()->setInterpreted();
    TR::Node *refNode = insertionPoint->getNode();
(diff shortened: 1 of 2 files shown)


compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -579,11 +582,44 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
...
+bool
+OMR::ResolvedMethodSymbol::induceOSRAfterAndRecompile(TR::TreeTop *insertionPoint, TR_ByteCodeInfo induceBCI, TR::TreeTop* branch,
+    bool extendRemainder, int32_t offset, TR::TreeTop ** lastTreeTop)
+   {
+   TR_ASSERT(self()->comp()->allowRecompilation(), "request OSR and recompilation at node %p when recomp is disabled\n", insertionPoint);
+   TR::TreeTop *induceOSRCallTree = induceOSRAfterImpl(insertionPoint, induceBCI, branch, extendRemainder, offset, lastTreeTop);
+   if (!induceOSRCallTree)
+      return false;
+   TR::Node *induceOSRCallNode = induceOSRCallTree->getNode()->getFirstChild();
+   induceOSRCallNode->setSymbolReference(self()->comp()->getSymRefTab()->findOrCreateRuntimeHelper(TR_induceOSRAtCurrentPCAndRecompile,
```
andrewcraik: can we have an assert before setting the symbolreference that the node is a call and that it has the non-recomp symbol reference. This should be a fatal assert for safety.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -579,11 +582,44 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
    return insertionPoint->getPrevTreeTop();
    }
 
+/*
+ *  \brief Generate the helper call to request for OSR transition and recompilation of the current compiled method.
```
andrewcraik: Please document the meaning of the bool return condition for the public APIs you have doxygened in this PR.

andrewcraik:
I'm fine with this design - passing the symbol / symref to use for the induce down through all the layers of the OSR infrastructure will be super messy so replacing the symbol reference on the returned tree seems an elegant solution. Once the appropriate safety assert is added I'm good with this. @vijaysun-omr

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
cathyzhyi: added documentation to explain the return value. Thanks!

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
cathyzhyi: Added the assert to verify the callNode we get is as expected.

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -579,11 +582,54 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
...
+   {
+   TR_ASSERT(self()->comp()->allowRecompilation(), "request OSR and recompilation at node %p when recomp is disabled\n", insertionPoint);
+   TR::TreeTop *induceOSRCallTree = induceOSRAfterImpl(insertionPoint, induceBCI, branch, extendRemainder, offset, lastTreeTop);
+   if (!induceOSRCallTree)
+      return false;
+   TR::Node *induceOSRCallNode = induceOSRCallTree->getNode()->getFirstChild();
+   TR::SymbolReference *symRef = induceOSRCallNode->getSymbolReference();
+   TR_ASSERT_FATAL(induceOSRCallNode->getOpCode().isCall() &&
+                   symRef->getReferenceNumber() == TR_induceOSRAtCurrentPC,
+                   "induceOSRCallNode %p under induceOSRCallTree %p should be a call node with TR_induceOSRAtCurrentPC helper call", induceOSRCallNode, induceOSRCallTree);
```
Leonardo2718: To help with debugging, would be useful to also print the node id for `induceOSRCallNode`?

Leonardo2718:
@genie-omr build all

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
cathyzhyi: added id for `induceOSRCallNode` into assert msg. Thanks!

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -579,11 +582,54 @@ OMR::ResolvedMethodSymbol::genInduceOSRCallNode(TR::TreeTop* insertionPoint,
...
 OMR::ResolvedMethodSymbol::induceOSRAfter(TR::TreeTop *insertionPoint, TR_ByteCodeInfo induceBCI, TR::TreeTop* branch,
     bool extendRemainder, int32_t offset, TR::TreeTop ** lastTreeTop)
    {
+   return induceOSRAfterImpl(insertionPoint, induceBCI, branch, extendRemainder, offset, lastTreeTop) != NULL;
+   }
+
+/*
+ *  \brief Internal implementation for generating the helper call to request for OSR transition.
+ *
+ *  \return The treetop of the inserted induceOSR helper call
```
Leonardo2718: Sorry, one other knit: can this comment also specify under what conditions `NULL` is returned?

cathyzhyi: added explanation for return value NULL. thanks!

Leonardo2718:
@genie-omr build all

andrewcraik:
@vijaysun-omr ping and @cathyzhyi please address the missing self()->

Leonardo2718:
@genie-omr build all

vijaysun-omr:
Looks okay to me. Merging.

Issue #3033, opened 2018-10-05
Make sure opimizations are aware about wrtbar/rdbars


cathyzhyi:
Wrtbar and rdbar opcodes are added in https://github.com/eclipse/openj9/pull/3138 and more investigation is needed to make sure the opts working for a normal store/load also work for wrtbar/rdbar.

0xdaryl:
@cathyzhyi:  just to clarify, this is for performance rather than functional correctness?

cathyzhyi:
@0xdaryl yes, it's for performance and the opcodes are functionally working now.

cathyzhyi:
the PR disabled PRE and FieldPrivatizer for functional correctness https://github.com/eclipse/omr/pull/3224

PR #3224, opened 2018-11-21
Guarantee functional correctness of rdbar/wrtbar

cathyzhyi:
Map rdbar/wrtbars to corresponding opcode to normal opcode to keep
consistency with the existing wrtbar opcode. Subprojects can overwrite
the mapping decisions.

PRE and FieldPrivatizer are disabled for now if the trees might contain
newly added rdbar/wrtbar. All the transformations applied to normal
load/write s should be applied to rd/wrtbar s. However, we need to
be careful about the difference in the shape of the trees and the
mapping relationship between different loads and writes. More
investigation is needed to make sure those 2 opts would do the right
thing for rdbar/wrtbars.

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -2734,3 +2734,8 @@ void OMR::Compilation::invalidateAliasRegion()
       new (&self()->_aliasRegion) TR::Region(_heapMemoryRegion);
       }
    }
+
+bool OMR::Compilation::incompleteOptimizerSupportForReadWriteBarriers()
+   {
+   return false;
+   }
--- a/compiler/il/OMRIL.cpp
+++ b/compiler/il/OMRIL.cpp
@@ -600,13 +600,17 @@ OMR::IL::opCodeForCorrespondingIndirectLoad(TR::ILOpCodes loadOpCode)
       case TR::iuload: return TR::iustore;
       case TR::iuloadi: return TR::iustorei;
       case TR::luloadi: return TR::lustorei;
-      case TR::brdbari: return TR::bwrtbari;
-      case TR::srdbari: return TR::swrtbari;
-      case TR::irdbari: return TR::iwrtbari;
-      case TR::lrdbari: return TR::lwrtbari;
-      case TR::frdbari: return TR::fwrtbari;
-      case TR::drdbari: return TR::dwrtbari;
-      case TR::ardbari: return TR::awrtbari;
+      case TR::brdbari:
+      case TR::srdbari:
+      case TR::irdbari:
+      case TR::lrdbari:
+      case TR::frdbari:
+      case TR::drdbari:
+      case TR::ardbari:
+         //There is not necessarily a guaranteed symmetry about whether an indirect rdbar should be mapped to
+         //an indirect wrtbar or a normal indirect store. The mapping of rdbar/ wrtbar totally depends on the
+         //actual use in subprojects and should be undefined in OMR level.
+         TR_ASSERT_FATAL(0, "xrdbari can't be used with global opcode mapping API at OMR level\n");
       default: break;
       }
 
@@ -628,13 +632,23 @@ OMR::IL::opCodeForCorrespondingIndirectStore(TR::ILOpCodes storeOpCode)
       case TR::fstorei:  return TR::floadi;
       case TR::dstorei:  return TR::dloadi;
       case TR::astorei:  return TR::aloadi;
-      case TR::awrtbari:  return TR::aloadi;
+      case TR::awrtbari: return TR::aloadi;
       case TR::vstorei:  return TR::vloadi;
-      case TR::cstorei: return TR::cloadi;
+      case TR::cstorei:  return TR::cloadi;
       case TR::bustorei: return TR::buloadi;
-      case TR::iustore: return TR::iuload;
+      case TR::iustore:  return TR::iuload;
       case TR::iustorei: return TR::iuloadi;
       case TR::lustorei: return TR::luloadi;
+      case TR::bwrtbari:
+      case TR::swrtbari:
+      case TR::iwrtbari:
+      case TR::lwrtbari:
+      case TR::fwrtbari:
+      case TR::dwrtbari:
+         //There is not necessarily a guaranteed symmetry about whether an indirect wrtbar should be mapped to
+         //an indirect rdbar or a normal indirect load. The mapping of rdbar/wrtbar totally depends on the
+         //actual use in subprojects and should be undefined in OMR level.
+         TR_ASSERT_FATAL(0, "xwrtbari can't be used with global opcode mapping API at OMR level\n");
 
       default: break;
       }
--- a/compiler/optimizer/FieldPrivatizer.cpp
+++ b/compiler/optimizer/FieldPrivatizer.cpp
@@ -85,6 +85,15 @@ TR_FieldPrivatizer::TR_FieldPrivatizer(TR::OptimizationManager *manager)
 
 int32_t TR_FieldPrivatizer::perform()
    {
+
+   // Need to confirm places calling opCodeFor* APIs are doing the right
+   // thing for readbar and wrtbar. All the transformations applied to normal
+   // load/store s should be applied to rd/wrtbar s. However, we need to be
+   // careful about the difference in the shape of the trees and the
+   // mapping relationship between different loads and stores
+   if (comp()->incompleteOptimizerSupportForReadWriteBarriers())
+      return 0;
+
    TR::StackMemoryRegion stackMemoryRegion(*trMemory());
 
    _postDominators = NULL;
--- a/compiler/optimizer/PartialRedundancy.cpp
+++ b/compiler/optimizer/PartialRedundancy.cpp
@@ -230,6 +230,14 @@ int32_t TR_PartialRedundancy::perform()
    if (comp()->getProfilingMode() == JitProfiling && comp()->getHCRMode() != TR::none && _numProfilingsAllowed == 0)
       return 0;
 
+   // Need to confirm places calling opCodeFor* APIs are doing the right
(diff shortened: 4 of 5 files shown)


Leonardo2718:
@genie-omr build all

compiler/compile/OMRCompilation.cpp
```diff
@@ -2730,3 +2730,12 @@ void OMR::Compilation::invalidateAliasRegion()
       new (&self()->_aliasRegion) TR::Region(_heapMemoryRegion);
       }
    }
+
+// This query is for temporarily disable opts not supporting
+// the newly added rdbar/wrtbar opcodes. Subprojects can overrite
+// the answer. This query should be deleted eventually if
+// all the optimizations support the opcodes.
```
Leonardo2718: Can you use doxygen formatting for this comment block?

compiler/optimizer/FieldPrivatizer.cpp
```diff
@@ -85,6 +85,15 @@ TR_FieldPrivatizer::TR_FieldPrivatizer(TR::OptimizationManager *manager)
 
 int32_t TR_FieldPrivatizer::perform()
    {
+
+   // TODO: need to confirm places calling opCodeFor* APIs are doing the right
```
Leonardo2718: Can you create an issue for this and remove the `TODO`?

compiler/optimizer/PartialRedundancy.cpp
```diff
@@ -230,6 +230,14 @@ int32_t TR_PartialRedundancy::perform()
    if (comp()->getProfilingMode() == JitProfiling && comp()->getHCRMode() != TR::none && _numProfilingsAllowed == 0)
       return 0;
 
+   // TODO: need to confirm places calling opCodeFor* APIs are doing the right
```
Leonardo2718: Can you create an issue for this and remove the `TODO`?

Leonardo2718:
Also, can you find a better commit title please (I'd try to give a suggestion but I can't think of anything at the moment :sweat_smile:)?

compiler/compile/OMRCompilation.cpp
cathyzhyi: Changed to use doxygen formatting.

cathyzhyi:
> Also, can you find a better commit title please (I'd try to give a suggestion but I can't think of anything at the moment )?

changed the title as discussed. thanks!

compiler/optimizer/FieldPrivatizer.cpp
cathyzhyi: removed the `TODO`. this is actually already an issue tracking this. I am putting the issue in the comment.

cathyzhyi:
An issue was created to track further work in the optimizer https://github.com/eclipse/omr/issues/3033

Leonardo2718:
@andrewcraik please review.

andrewcraik:
Hmm this raises an interesting philosophical issue - if you need to load from something that was stored with a wrtbari do you need to do a rdbari? There is not necessarily a guaranteed symmetry - that which needs a wrtbar may not need a rdbar and vice versa. From this perspective having the wrtbar map to a normal load and a rdbar to a normal store makes sense. I do worry, however, that there might be an unintended change in behavior with the change in mapping. @cathyzhyi did you do some kind of survey of usage to know that the change in mapping for wrtbar will not result in functional issues?

At the very least if a wrtbar does not always, but may need to, become a rdbar then we need a mapping of loads to rdbars and stores to wrtbars for ops to use and adding that now would make sense if we agree a wrtbar does not need to become a rdbar. Given that the side-effects of these side-effecting loads and stores is not known apriori we can't have a general rule so need to provide for the possibility IMO.

cathyzhyi:
After discussing with Andrew, the mapping of rdbar/ wrtbar totally depends on the actual use and should be undefined in OMR level. Eventually, the optimizations involved will need to specifically look at what's get loaded or stored to decide what opcode to use rather than using a global mapping API like this.  I am changing the mapping of the rdbar/wrtbar to assert fatal to make sure no one is using the mapping API without realizing the complexity behind.

compiler/optimizer/FieldPrivatizer.cpp
```diff
@@ -85,6 +85,15 @@ TR_FieldPrivatizer::TR_FieldPrivatizer(TR::OptimizationManager *manager)
 
 int32_t TR_FieldPrivatizer::perform()
    {
+
+   // Need to confirm places calling opCodeFor* APIs are doing the right
+   // thing for readbar and wrtbar. All the transformations applied to normal
+   // load/write s should be applied to rd/wrtbar s. However, we need to be
+   // careful about the difference in the shape of the trees and the
+   // mapping relationship between different loads and writes
+   if (comp()->mayContainUnsupportedReadbarOrWritebar())
```
vijaysun-omr: Where is the code that returns true for this query ? The implementation in this commit returns false.

cathyzhyi: the subprojects will overwrite the answer to return true depending on options enabled.

compiler/compile/OMRCompilation.cpp
```diff
@@ -2730,3 +2730,18 @@ void OMR::Compilation::invalidateAliasRegion()
...
+ * \brief
+ *    This query tells whether the trees might contain rdbar/wrtbar opcodes
+ *
+ * \note
+ *    This query is for temporarily disable opts not supporting
+ *    the newly added rdbar/wrtbar opcodes. Subprojects can overrite
+ *    the answer. This query should be deleted eventually if
+ *    all the optimizations support the opcodes.
+ */
+bool OMR::Compilation::mayContainUnsupportedReadbarOrWritebar()
```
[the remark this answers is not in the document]
cathyzhyi: I was hoping to avoid confusion with the existing `awrtbari` and`awrtbar`. Those wrtbars are supported by all the opts in context of GC. `mayContainReadbarOrWritebar` sounds like it's referring to all read and write bars in all cases.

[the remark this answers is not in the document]
0xdaryl: My objection is around the generality of this query.  "unsupported" leaves one to wonder where and how they are unsupported.  I think if the context could be more clearly defined then it would be ok.  For example, if this is to prevent optimizations from running that may not competely implement support for all barriers then the somewhat wordy `incompleteOptimizerSupportForReadWriteBarriers` might be a less ambiguous name for this query.

Ideally, we should prevent optimizations from even running if we generate read/write barriers during IL gen that we know have incomplete support in certain optimizations, but we don't (yet!) have that capability.

compiler/il/OMRIL.cpp
```diff
@@ -600,13 +600,14 @@ OMR::IL::opCodeForCorrespondingIndirectLoad(TR::ILOpCodes loadOpCode)
...
-      case TR::drdbari: return TR::dwrtbari;
-      case TR::ardbari: return TR::awrtbari;
+      case TR::brdbari:
+      case TR::srdbari:
+      case TR::irdbari:
+      case TR::lrdbari:
+      case TR::frdbari:
+      case TR::drdbari:
+      case TR::ardbari:
+         TR_ASSERT_FATAL(0, "xrdbari can't be used with global opcode mapping API at OMR level\n");
```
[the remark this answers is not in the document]
cathyzhyi: Added comments to explain the restriction.

compiler/il/OMRIL.cpp
```diff
@@ -628,13 +629,20 @@ OMR::IL::opCodeForCorrespondingIndirectStore(TR::ILOpCodes storeOpCode)
...
+      case TR::iustore:  return TR::iuload;
       case TR::iustorei: return TR::iuloadi;
       case TR::lustorei: return TR::luloadi;
+      case TR::bwrtbari:
+      case TR::swrtbari:
+      case TR::iwrtbari:
+      case TR::lwrtbari:
+      case TR::fwrtbari:
+      case TR::dwrtbari:
+         TR_ASSERT_FATAL(0, "xwrtbari can't be used with global opcode mapping API at OMR level\n");
```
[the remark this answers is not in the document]
cathyzhyi: Added comments to explain the restriction.

compiler/compile/OMRCompilation.cpp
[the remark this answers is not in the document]
cathyzhyi: Changed the function name as suggested to be more clear in the intention.

cathyzhyi:
@Leonardo2718 I've addressed all the concerns in the comments. Can you review again?

Leonardo2718:
@genie-omr build all

compiler/compile/OMRCompilation.cpp
```diff
@@ -2734,3 +2734,19 @@ void OMR::Compilation::invalidateAliasRegion()
...
+ *    This query tells whether the trees might contain rdbar/wrtbar opcodes
+ *    that are not fully supported by optimizer yet
+ *
+ * \note
+ *    This query is for temporarily disable opts not supporting
+ *    the newly added rdbar/wrtbar opcodes. Subprojects can overrite
+ *    the answer. This query should be deleted eventually if
+ *    all the optimizations support the opcodes.
+ */
+bool OMR::Compilation::incompleteOptimizerSupportForReadWriteBarriers()
```
fjeremic: Thanks for the documentation. I'd like to see this moved to the header file however as the header defines the API to consuming projects. Projects consuming OMR should not have to dig through the OMR implementation to find out what a particular API is supposed to achieve.

cathyzhyi: Moved the documentation to header file as suggested.

Leonardo2718:
@genie-omr build all

vijaysun-omr:
I am okay with this now. @0xdaryl are you also willing to approve this ?

vijaysun-omr:
All reviewers who had comments on this have approved and checks have passed.

PR #3225, opened 2018-11-21
Handle signals blocked by a parent process

babsingh:
If a process has blocked signals, then the signals stay blocked in the
sub-processes across fork(s) and exec(s). Blocked signals prevent signal
handlers to be invoked. Signals listed in signalMap should be unblocked
during initialization in case a parent process has blocked any of those
signals.

Issue: https://github.com/eclipse/openj9/issues/3468

### DIFF
--- a/port/common/omrport.tdf
+++ b/port/common/omrport.tdf
@@ -1083,3 +1083,7 @@ TraceEvent=Trc_PRT_retrieveLinuxMemoryStats_CgroupMemoryStatsFailed Group=sysinf
 
 TraceEntry=Trc_PRT_signal_omrsig_is_signal_ignored_entered Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_signal_ignored: Entered, portLibrarySignalFlag=0x%X"
 TraceExit=Trc_PRT_signal_omrsig_is_signal_ignored_exiting Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_signal_ignored: Exiting, rc=%d, isSignalIgnored=%d"
+
+TraceException=Trc_PRT_signal_unblockSignals_sigemptyset_failed Group=signal Overhead=1 Level=1 NoEnv Template="omrsignal unblockSignals: sigemptyset failed with rc=%d and errno=%d"
+TraceException=Trc_PRT_signal_unblockSignals_sigaddset_failed Group=signal Overhead=1 Level=1 NoEnv Template="omrsignal unblockSignals: sigaddset failed with signal=0x%X, rc=%d and errno=%d"
+TraceException=Trc_PRT_signal_unblockSignals_pthread_sigmask_failed Group=signal Overhead=1 Level=1 NoEnv Template="omrsignal unblockSignals: pthread_sigmask failed with rc=%d and errno=%d"
--- a/port/unix/omrsignal.c
+++ b/port/unix/omrsignal.c
@@ -36,6 +36,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <setjmp.h>
+#include <errno.h>
 #include "omrsignal_context.h"
 
 #if !defined(J9ZOS390)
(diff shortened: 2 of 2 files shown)


babsingh:
fyi - @pshipton @charliegracie

charliegracie:
@genie-omr build all

babsingh:
I think we should use `pthread_sigmask` instead of `sigprocmask`. `sigprocmask`'s behavior is unspecified for multi-threaded programs. `pthread_sigmask` is just like `sigprocmask` but it supports multi-threaded applications. I will update the PR after testing the new changes. Setting PR to WIP until the new changes are added.

`sigprocmask` - http://man7.org/linux/man-pages/man2/sigprocmask.2.html
`pthread_sigmask` - http://man7.org/linux/man-pages/man3/pthread_sigmask.3.html

port/unix/omrsignal.c
```diff
@@ -1795,3 +1807,37 @@ omrsig_chain_at_shutdown_and_exit(struct OMRPortLibrary *portLibrary)
...
+			return FALSE;
+		}
+	}
+
+	/* Unblock the signals listed in signalMap. */
+	if (0 != sigprocmask(SIG_UNBLOCK, &signalSet, NULL)) {
+		return FALSE;
+	}
+
+	return TRUE;
```
DanHeidinga: Can you reformat this to have a single return point?

DanHeidinga: Please also add some tracepoints to this function, particularly for the error cases.

babsingh: Tracepoints won't work in initialization code. Asserts are used instead of tracepoints at such locations. I will add asserts.

DanHeidinga: I'm not sure asserts help here.  The goal is to make it easy to understand what failed and why - not to take the runtime down quicker :)

port/common/omrport.tdf
```diff
@@ -1083,3 +1083,7 @@ TraceEvent=Trc_PRT_retrieveLinuxMemoryStats_CgroupMemoryStatsFailed Group=sysinf
 
 TraceEntry=Trc_PRT_signal_omrsig_is_signal_ignored_entered Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_signal_ignored: Entered, portLibrarySignalFlag=0x%X"
 TraceExit=Trc_PRT_signal_omrsig_is_signal_ignored_exiting Group=signal Overhead=1 Level=3 NoEnv Template="omrsig_is_signal_ignored: Exiting, rc=%d, isSignalIgnored=%d"
+
+TraceException=Trc_PRT_signal_unblockSignals_sigemptyset_failed Group=signal Overhead=1 Level=3 NoEnv Template="omrsignal unblockSignals: sigemptyset failed with rc=%d and errno=%d"
```
DanHeidinga: These are all exception paths so they can be Level=1

babsingh: updated to Level=1.

DanHeidinga:
@charliegracie If you're OK with this change, can you kick off the genie builds?

charliegracie:
Hmmm I am not sure we should have switched the `sigprocmask` to `pthread_sigmask`.  This code is supposed to be called during startup before there are any other threads.  Using the pthread one may mask issues if the code is ever moved and it would make it easier to move since this is supposed to thread safe now.

babsingh:
> This code is supposed to be called during startup before there are any other threads

New threads inherit properties from their parent thread. The above constraint is only applied so that we unblock signals **once** in the main thread. Then, we won't have to unblock signals for the child threads spawned from the main thread.

This code should be location independent. I expect it to work without issues from anywhere. 

From my readings, functionally `sigprocmask` == `pthread_sigmask`. It's safer to use `pthread_sigmask` since `sigprocmask` may yield unexpected behavior in a multi-threaded setting. Instead of `pthread_sigmask`, I feel that `sigprocmask` may mask issues due to unexpected behavior.

charliegracie:
@genie-omr build all

PR #3171, opened 2018-11-09
AArch64: Implement icmpeqEvaluator(), etc.

knn-k:
This commit implements icmpeqEvaluator() and similar evaluators for
aarch64.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #3227, opened 2018-11-22
Tril tests for icmpeq, etc.

knn-k:
This commit adds tril test cases for icmpeq and similar evaluators.


0xdaryl:
@genie-omr build aarch64

Now that #3171 has merged.

0xdaryl:
Duh...  This will still only test the builds, not runtime.

knn-k:
Is my understanding below correct?

I need to add a new template class other than `BinaryOpTest` in `OpCodeTest.hpp` for writing test cases for opcodes like `lcmpeq`, which have different types with input (int64_t) and output (int32_t).
The same can be said for type conversion opcodes, such as `i2l`.

Leonardo2718:
It's looking good @knn-k !

> Is my understanding below correct?
>
> I need to add a new template class other than BinaryOpTest in OpCodeTest.hpp for writing test cases for opcodes like lcmpeq, which have different types with input (int64_t) and output (int32_t).
The same can be said for type conversion opcodes, such as i2l.

Actually, you should be able to use `OpCodeTest` instead of `BinaryOpTest`. It's a variadic template so it can (in theory) handle an arbitrary number of arguments. The format for it is `OpCodeTest<ReturnT, Arg1T, Arg2T, etc.>`. So, in your case, you might use it like this:

```c++
class Int64Compare : public TRTest::OpCodeTest<int32_t,int64_t,int64_t> {};
```

I believe the `to_struct()` helper should still work.

Is this ready for merging, or are you planning on adding the lcmp tests in this PR?

knn-k:
Thanks  @Leonardo2718.  I think I will add lcmp tests in another PR.

PR #3230, opened 2018-11-22
Simplify instanceof/guard type constraint creation

jdmpapin:
...where genTypeResult() was used.

Rather than attempt not to intersect constraints, this logic now simply
generates the constraint for objects that pass the type test, and tries
to add that constraint on the appropriate path, implicitly intersecting
it with any pre-existing constraint. The branch can be folded away if
one of the paths is found to be unreachable due to an empty constraint
intersection.

The previous logic could sometimes fold away a profiled guard against
java/lang/Class even in the absence of any other type information,
generating an unconditional goto to the taken side (fall-back path),
which is functionally correct but undesirable. This now only happens
when an existing type constraint precludes the object from being an
instance of java/lang/Class.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -8998,169 +8998,67 @@ static void removeConditionalBranch(OMR::ValuePropagation *vp, TR::Node *node, T
       vp->_edgesToBeRemoved->add(branchEdge);
    }
 
-// check for compatibility between types
-//
+// Constrain objects that pass a type test, given a constraint on the class
+// against which the object is to be tested.
 static TR::VPConstraint*
-genTypeResult(OMR::ValuePropagation *vp, TR::VPConstraint *objectRefConstraint,
-               TR::VPConstraint *classConstraint, bool &applyType, bool isInstanceOf)
+passingTypeTestObjectConstraint(
+   OMR::ValuePropagation *vp,
+   TR::VPConstraint *classConstraint,
+   bool testingForFixedType)
    {
-   TR::VPConstraint *newConstraint = NULL;
-   if (objectRefConstraint)
+   TR_ASSERT_FATAL(
+      classConstraint->isClassObject() == TR_yes,
+      "expected a instanceof classConstraint to be a 'ClassObject'");
+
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

jdmpapin:
The AArch64 build failed to `git fetch`. Can it be restarted?

Leonardo2718:
@genie-omr build aarch64

andrewcraik:
This makes sense to me

andrewcraik:
@Leonardo2718 looks like something is up with the test builds - lots are still pending...

Leonardo2718:
@genie-omr build all

compiler/optimizer/VPHandlers.cpp
```diff
@@ -8998,169 +8998,67 @@ static void removeConditionalBranch(OMR::ValuePropagation *vp, TR::Node *node, T
...
+   TR::VPConstraint *classConstraint,
+   bool testingForFixedType)
+   {
+   TR_ASSERT_FATAL(
+      classConstraint->isClassObject() == TR_yes,
+      "expected a instanceof classConstraint to be a 'ClassObject'");
+
+   TR::VPClassType *type = classConstraint->getClassType();
+   TR_ASSERT_FATAL(
+      type != NULL,
```
vijaysun-omr: Class.isInstance gets mapped to an instanceof IL opcode in the IL gen in some cases, and in the most general case, this could generate an instanceof with the second child not being a constant class. In this case, is it possible that we don't know the type ? i.e. we know it's a ClassObject but we don't know of what type.

jdmpapin: The only caller is `constrainIfcmpeqne()`. At both call sites it passes `instanceofConstraint`, and only when `instanceofConstraint` has been set, in which case it always has a `getClassType()`. We could make `passingTypeTestObjectConstraint()` more general, so that it handles completely unknown types, but it would only be able to tell us that the object is non-null, and the generality would go unused without further changes elsewhere.

vijaysun-omr: Okay so the callers of the routine currently always pass in a class type. Thanks for the clarification.

PR #3231, opened 2018-11-22
Constrain VFT of ClassObject to be java/lang/Class

jdmpapin:
It was already so constrained in cases where the class represented by
the ClassObject is known exactly, but the represented class is
irrelevant here. Now (isClassObject() == TR_yes) takes precedence over
isFixedClass() so a location of {JavaLangClassObject} is sufficient to
cause VP to create the constraint.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -2124,33 +2124,30 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
       if (base && node->getOpCode().hasSymbolReference() &&
           (node->getSymbolReference() == vp->comp()->getSymRefTab()->findVftSymbolRef()))
          {
-         TR::VPClassPresence *nonnull = TR::VPNonNullObject::create(vp);
-         if (!base->isFixedClass())
-            {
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

andrewcraik:
@Leonardo2718 looks like something is up with the test builds - lots are still pending...

Leonardo2718:
@genie-omr build all

vijaysun-omr:
The enhancement to the type propagation logic here looks valid in the case when base object is a ClassObject. The rest of the changes are primarily rearranging the code which also look equivalent to me.

vijaysun-omr:
Is the title accurate ? The VFT load is not a java/lang/Class and is not being constrained as such in the code.

jdmpapin:
> Is the title accurate ? The VFT load is not a java/lang/Class and is not being constrained as such in the code. I think maybe you mean the type of a ClassObject value is being constrained to java/lang/Class ?

Right, I don't mean that the VFT is _a_ `java/lang/Class`, but rather than it is the _`J9Class` of_ `java/lang/Class` (in J9 terms). Commit titles are very cramped and in this case I don't know that I can squeeze that distinction into the space available. I don't think I'd say the title is inaccurate though: in a certain way the constraint says that the VFT is `java/lang/Class` itself.

PR #3233, opened 2018-11-22
Misc startup/shutdown fixes for Concurrent Scavenger

amicic:
- explicitly call default constructor for various stats structs in
GCExtensions 
- initialize MasterGCThread struct earlier, so that Scavenge can be
triggered on early Nursery Allocation Failure (with small Xms/Xmns),
rather then fullfilling the request from Tenure (and slowing down the
startup) 
- on shutdown free up previous RS list, too. It will be non-empty, if
shutdown is requested in a middle of CS cycle

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -280,6 +280,14 @@ MM_Scavenger::initialize(MM_EnvironmentBase *env)
 
 	_cacheLineAlignment = CACHE_LINE_SIZE;
 
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+	if (_extensions->concurrentScavenger) {
+		if (!_masterGCThread.initialize(this, true, true)) {
+			return false;
+		}
+	}
+#endif /* OMR_GC_CONCURRENT_SCAVENGER */
+
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build all

gc/base/standard/Scavenger.cpp
```diff
@@ -280,6 +280,14 @@ MM_Scavenger::initialize(MM_EnvironmentBase *env)
 
 	_cacheLineAlignment = CACHE_LINE_SIZE;
 
+#if defined(OMR_GC_CONCURRENT_SCAVENGER)
+	if (_extensions->concurrentScavenger) {
+		if (!_masterGCThread.initialize(this, true, true)) {
+			return false;
```
charliegracie: Just curious why this had to move from `collectorStartup` to the `initialize` method?

amicic: It was too late to do it collectorStartup (but did not investigate why). As a minimum MasterGCThread needs to know about collector (the init method passes 'this' pointer). 
When triggering a GC, the path goes through Collector then (if CS is enabled) through  MasterGCThread (to figure out if explicit Master thread is needed, based on if it's STW phase or concurrent phase) and back to Collector. Possibly things could be simplified (for STW phase of CS) to not go through MasterGCThread, but it goes always (for CS) for unified code. Bottom line, MasterGCThread needs to know about Collector from very early points to be able to trigger at least STW phase.

Otherwise, very early, with very small Xms, MasterGCThread would just return and do nothing, We would not trigger OOM though... MSSGenerational would redirect the AF to Tenure. So we could see 10s or 100s of AFs satisfied by Tenure, instead of properly triggering Scavenge. Things would stabilize, but it would just slow startup down and pollute verbose GC. 

This was simplest possible fix - we now trigger Scavenge on first AF, although first couple of cycles may not have concurrent phase, since creation of explicit master thread may be further delayed.

amicic:
These fixes help with https://github.com/eclipse/openj9/issues/3451.

PR #3234, opened 2018-11-23
VGC: report Concurrent Scavenger phase termination

amicic:
If concurrent phase is prematurely terminated, a warning message is
reported by verbose GC.

Distinguish (for now) between:
- externally requested termination (JVMTI and similar)
- GC requested (exhausted allocate/survivor, global STW, etc.)

Example of exhausted allocate/survivor:

```
<concurrent-end id="13810" type="scavenge" contextid="13800"
timestamp="2018-11-23T18:07:54.482">
<warning details="termination requested by GC" />  <<<<<<<<<<<<<<<<<<<<
<gc-op id="13811" type="scavenge" timems="9.128"
contextid="13800" timestamp="2018-11-23T18:07:54.482">
  <memory-copied type="nursery" objects="46034"
bytes="7511576" bytesdiscarded="142136" />
  <memory-copied type="tenure" objects="458" bytes="17376"
bytesdiscarded="496" />
  <copy-failed type="nursery" objects="49" bytes="3624" />
</gc-op>
</concurrent-end>

<exclusive-start id="13812"
timestamp="2018-11-23T18:07:54.482" intervalms="13.230">
  <response-info timems="2.577" idlems="0.784" threads="3"
lastid="000000000141C500" lastname="AbstractDriver Thread Spawner for
jobset Mauve" />
</exclusive-start>
<af-start id="13813" threadId="0000000001FFB088"
totalBytesRequested="48" timestamp="2018-11-23T18:07:54.483"
intervalms="13.249" type="nursery" />
<gc-start id="13814" type="scavenge" contextid="13800"
timestamp="2018-11-23T18:07:54.484">
  <mem-info id="13815" free="45614456" total="241762304"
percent="18">
    <mem type="nursery" free="0" total="140247040"
percent="0">
      <mem type="allocate/survivor" free="0"
total="14024704" percent="0" />      <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
      <mem type="evacuate" free="0" total="126222336"
percent="0" />
    </mem>
   
```

### DIFF
--- a/gc/verbose/VerboseHandlerOutput.cpp
+++ b/gc/verbose/VerboseHandlerOutput.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2016 IBM Corp. and others
+ * Copyright (c) 1991, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build all

gc/verbose/schema.xsd
```diff
@@ -612,6 +612,7 @@ xmlns:vgc="http://www.ibm.com/j9/verbosegc">
 	<complexType name="concurrent-end">
 		<sequence>
 			<element ref="vgc:concurrent-mark-end" maxOccurs="1" minOccurs="1" />
+			<element ref="vgc:warning" maxOccurs="unbounded" minOccurs="0" />
```
charliegracie: slight nitpick. Can there be more than warning per concurrent GC end?

amicic: yes, there could be a gc-op. clearly, missed from a long ago

amicic: added

amicic: This particular warning could show up only once, but since I'm reusing generic warning message, perhaps there could be some other ones too (overflow etc), although I'm not sure (perhaps others could show up only in the final STW gc-op). I can investigate if that's the case....

For gc-op, well, clearly it's a part of it (see example below), although it does not have to show up in context of Balanced GC (hence, min=0).

charliegracie: Thanks for the clarification!

charliegracie:
@genie-omr build all

gc/verbose/VerboseHandlerOutput.cpp
```diff
@@ -861,6 +861,33 @@ MM_VerboseHandlerOutput::handleConcurrentStart(J9HookInterface** hook, UDATA eve
 	exitAtomicReportingBlock();
 }
 
+
+void
+MM_VerboseHandlerOutput::handleConcurrentEndInternal(J9HookInterface** hook, UDATA eventNum, void* eventData)
+{
+	MM_ConcurrentPhaseEndEvent *event = (MM_ConcurrentPhaseEndEvent *)eventData;
+	MM_ConcurrentPhaseStatsBase *stats = (MM_ConcurrentPhaseStatsBase *)event->concurrentStats;
```
charliegracie: Sorry I missed these earlier.... Are these stats / event structs under the concurrent scavenger build define? Should they be?

amicic: Implementation/body of handleConcurrentEndInternal (in base VGC handler) could indeed be under CONCURRENT_SCAVENGER compile flag - it's all CS specific. It does not have to be though, since MM_ConcurrentPhaseStatsBase (or anything else in the method) is not CS specific (the stats struct is also used by Balanced GC).

Should I add it?

charliegracie: As long as nothing is CS specific and non CS specs will compile it is ok. I would rather not have things under if def if possible.

PR #3236, opened 2018-11-24
Create an issue template for new compiler IL opcode requests

0xdaryl:


0xdaryl:
@vijaysun-omr @fjeremic @Leonardo2718 @andrewcraik @mstoodle : I've taken your suggestions and incorporated them into the template.  Please review.

I'd appreciate @mstoodle or @charliegracie to sign off on the use of an issue template for this.

Here is a sample screenshot of what a user will see when they open a new issue:

image
I must admit I don't particularly like the fact that the issue template is given so much prominence (big font in a box) and there is only a small link for a regular issue (i.e., >99% of the issues created).  One way around this if we don't like it is to create a dummy issue template for "regular" issues that will be a similar size.

.github/ISSUE_TEMPLATE/new-compiler-il-opcode-request.md
```diff
@@ -0,0 +1,129 @@
...
+discussed and reviewed with your participation during a [Compiler Architecture Meeting](https://github.com/eclipse/omr/issues/2316).
+Please provide as much detail as possible to assist with the evaluation.
+
+The title of your issue should follow the format:
+```
+New IL Opcode Request: <name of IL opcode>
+```
+
+In the template below, you may delete the text that is provided in this template as
+you replace it with the information requested.
```
Leonardo2718: Should this text also be italicized?

fjeremic:
> I must admit I don't particularly like the fact that the issue template is given so much prominence (big font in a box) and there is only a small link for a regular issue (i.e., >99% of the issues created). One way around this if we don't like it is to create a dummy issue template for "regular" issues that will be a similar size.

Perhaps we should have a _Generic Issue_ template which would be the first option?

0xdaryl:
New IL opcodes are not created that often (thankfully?), but tend to generate a lot of discussion when they do.  I count only 6 PRs in the ~2 years that the compiler component has been open-sourced that introduced or proposed a new IL opcode.

You're right that this may be infrequent enough to not warrant a specialized template.  However, if we can make the generic case more prominent, ordered properly, and documented correctly it may stem some of those problems.  For example, what do you think of this?

image

fjeremic:
I'm in favor of the generic template. It also gives us the ability to advertise our Slack workspace for example or any other useful information for newcomers. An example of this can be found on the [babel](https://github.com/babel/babel) GitHub project. For example [1] displays a support question template and it is defined in [2]. Right above the title you can customize the _about_ section:

```
If you have a question 💬, please check out our Slack or StackOverflow!
```

And in the prepopulated text section you can put in a Markdown comment which will not appear in the submitted issue, for example:

```
<!---
This will not appear in the issue but I can put in whatever text I want here that is useful to the user.
-->
```

[1] https://github.com/babel/babel/issues/new?template=Support_question.md
[2] https://github.com/babel/babel/blob/master/.github/ISSUE_TEMPLATE/Support_question.md

charliegracie:
I am ok with the generic template as I think it can add a lot of valuable information for users who are trying to open issues. I think we should add 1 or a few generic templates.

I think the OMR should have a few different templates similar to projects like `babel` but I think the component feature specific templates will start to become very intrusive as components see the IL opcode template and want to add their own for specific features of their component.

I would abstain from a vote on this topic so if everyone else is ok with adding the Il opcode template then it is ok with me. I think we may end up revisiting the idea in the future if the number of templates starts to get out of hand which would also confuse users who are trying to open an Issue.

Leonardo2718:
ping @0xdaryl

fjeremic:
There seems to be a `0xdaryl-patch-1` [1] branch on the global repo associated with this PR. Do we still plan to move forward with this?

[1] https://github.com/eclipse/omr/branches

PR #3237, opened 2018-11-24
Remove deprecated reg-reg mov instruction list on X86

0dvictor:
Out-of-dated reg-reg mov instruction list has been deprecated, removing.


0xdaryl:
@genie-omr build xlinux,win

PR #3239, opened 2018-11-26
AArch64: Call decFutureUseCount() in Machine::assignOneRegister()

knn-k:
decFutureUseCount() has not been called upon register assignment for
aarch64 instructions. This commit adds a call to decFutureUseCount()
in OMR::ARM64::Machine::assignOneRegister().


knn-k:
ARM32 version calls `decFutureUseCount()` in its `assignSingleRegister()` in the same way as this commit, while P version calls `decFutureUseCountAndUnlatch()` every time after calling `assignOneRegister()` in `PPCInstruction.cpp`.

Which style is preferred, ARM32 or P?

knn-k:
Memo: #3139 handles a problem with TotalUseCount for virtual registers in dependency conditions.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3243, opened 2018-11-27
AArch64: Change the operand order of store instructions in trace file

knn-k:
This commit changes the order of operands of store instructions
from "str [sp, 0], lr" to "str lr, [sp, 0]", which is the same order
as that in AArch64 assembler syntax.


knn-k:
This is separated from #3139.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3248, opened 2018-11-27
WIP - Add extra assertions for atomic non-helpers on Power

hzongaro:
Added assertion that we're running a 64-bit JVM when trying to inlinevone of the atomic non-helpers, as 32-bit code generation is not yet supported for it.

Also if a call to codegen inlineDirectCall from directCallEvaluator does not inline a call, added an assertion that the symbol is not one of the atomic non-helpers.

### DIFF
--- a/compiler/compile/OMRSymbolReferenceTable.hpp
+++ b/compiler/compile/OMRSymbolReferenceTable.hpp
@@ -313,6 +313,13 @@ class SymbolReferenceTable
    //
    bool isNonHelper(TR::SymbolReference *, CommonNonhelperSymbol);
    bool isNonHelper(int32_t, CommonNonhelperSymbol);
+   bool isNonHelper(TR::SymbolReference *);
+   bool isNonHelper(int32_t);
+
+   // Look up non-helper symbol reference
+   CommonNonhelperSymbol getNonHelperSymbol(TR::SymbolReference *);
+   CommonNonhelperSymbol getNonHelperSymbol(int32_t);
+
 
    // Total number of symbols (known and dynamic) in the SRT
    //
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -1843,6 +1843,27 @@ bool OMR::SymbolReferenceTable::isNonHelper(int32_t ref, CommonNonhelperSymbol s
       }
    }
 
+bool OMR::SymbolReferenceTable::isNonHelper(TR::SymbolReference *symRef)
+   {
+   return isNonHelper(symRef->getReferenceNumber());
+   }
+
+bool OMR::SymbolReferenceTable::isNonHelper(int32_t ref)
+   {
+   return (ref >= _numHelperSymbols && ref < _numHelperSymbols + getLastCommonNonhelperSymbol());
+   }
+
+OMR::SymbolReferenceTable::CommonNonhelperSymbol OMR::SymbolReferenceTable::getNonHelperSymbol(TR::SymbolReference *symRef)
+   {
+   return getNonHelperSymbol(symRef->getReferenceNumber());
+   }
+
+OMR::SymbolReferenceTable::CommonNonhelperSymbol OMR::SymbolReferenceTable::getNonHelperSymbol(int32_t ref)
+   {
+   return isNonHelper(ref) ? (CommonNonhelperSymbol) (ref - _numHelperSymbols)
+                           : getLastCommonNonhelperSymbol();
+   }
+
 int32_t OMR::SymbolReferenceTable::getNonhelperIndex(CommonNonhelperSymbol s)
    {
    TR_ASSERT(s <= getLastCommonNonhelperSymbol(), "assertion failure");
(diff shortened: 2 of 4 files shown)


hzongaro:
@gita-omr Gita, this PR adds the extra assertions you suggested for pull #3018.  I will open a second pull request for a corresponding assertion in OpenJ9.

*cc:* Victor @0dvictor, Filip @fjeremic FYI

compiler/compile/OMRSymbolReferenceTable.hpp
```diff
@@ -313,6 +313,13 @@ class SymbolReferenceTable
    //
    bool isNonHelper(TR::SymbolReference *, CommonNonhelperSymbol);
    bool isNonHelper(int32_t, CommonNonhelperSymbol);
+   bool isNonHelper(TR::SymbolReference *);
+   bool isNonHelper(int32_t);
```
fjeremic: The APIs are now confusing. What is the difference between the existing ones vs. the new ones? It is definitely non-obvious to me looking at the definitions here. I'd like to see all these new APIs Doxygen documented. From a quick glance it seems the existing APIs looked at whether a symbol reference is a particular non-helper, where as the new APIs look for whether the symbol reference is any non-helper.

It also appears that the implementations of these APIs do not use any object data from the `OMRSymbolReferenceTable` object, which prompts the question of why they are not declared as `static`?

gita-omr: I actually think the the methods are very simple and self explanatory. The nonHelper name is the only thing that is confusing :)

fjeremic: i.e. the reason we need documentation is if you wish to ship OMR as a static library with header-only definitions the documentation for external public APIs should be documented right in the header because you may not even have the source implementations to look at.

gita-omr: Sure, one line of documentation does not hurt. BTW: why don't we inline these types of methods?

hzongaro: > It also appears that the implementations of these APIs do not use any object data from the OMRSymbolReferenceTable object, which prompts the question of why they are not declared as static?

Filip @fjeremic, I think they depend on the value of `_numHelperSymbols`, which is instance data from `OMRSymbolReferenceTable`, if I understand correctly.

I have added Doxygen comments for the existing `isNonHelper` methods as well as the new methods in <https://github.com/eclipse/omr/pull/3248/commits/d6d364d25b9abcf445765d9ec972fe8086768c7b>.  I would appreciate your reviewing them for correct terminology, clarity, _etc._

hzongaro: > why don't we inline these types of methods?

@gita-omr I don't know why they weren't inlined to begin with.  I am happy to replace all six methods with inline implementations if you would prefer and nobody objects.

gita-omr: @0xdaryl what do you think?

fjeremic: I agree we can make them inline.

hzongaro: Gita @gita-omr, Filip @fjeremic, I've been struggling trying to separate out the `isNonHelper` and `getNonHelperSymbol` methods into separate `*SymbolReferenceTable_inlines.hpp` header files.  My difficulty has been with trying to avoid requiring the various .cpp files that depend on `isNonHelper` and `getNonHelperSymbol` including these new `_inlines.hpp` files because of dependencies between `SymbolReference` and `SymbolReferenceTable`.

If I do tackle this by adding `#include`s for the new `_inlines.hpp` files to the `.cpp` files, I'll need to do this in stages:

1. Add empty `*SymbolReferenceTable_inlines.hpp` to OMR
2. Add `#include`s for the new `_inlines.hpp` files to OpenJ9
3. Actually add inline method implementations to OMR and `#include` for the new `_inlines.hpp` files to OMR (or the second half of this step could be done at step 1.)

If we want to avoid requiring `#include`s for the new `_inlines.hpp` files to the `.cpp` files, which I think is desirable, I'll need some insights or advice.

Any thoughts?  Either way, I'd suggest we go ahead with this PR and then the corresponding [OpenJ9 PR 3856](https://github.com/eclipse/openj9/pull/3856)

gita-omr: I agree to go ahead with this PR and postpone the inlining.

hzongaro: Thanks, Gita.  I will remove the WIP from this pull request then.

hzongaro: I forgot to mention here that I had opened issue #3285 to track the task of making those methods in `SymbolReferenceTable` inline.

hzongaro: @gita-omr Gita, were there any other changes you wanted to see in this pull request?

fjeremic:
@genie-omr build win

PR #3249, opened 2018-11-27
Clarify copyright updates in CONTRIBUTING.md

0xdaryl:
File renames or moves do not require a copyright header change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,7 +129,8 @@ Following these guidelines will help us to merge your pull requests smoothly:
    copyright/license requirements outlined below.
 
 4. For each file that you modify, ensure that the current copyright year in the
-   header at the top of the file is up-to-date with the current year.
+   header at the top of the file is up-to-date with the current year.  Note that
+   renaming, moving, or copying a file does not require a copyright change.
 
 5. Follow the coding style and format of the code you are modifying (see the
    [coding standards](doc/CodingStandard.md)). The code base is yet to be unified


youngar:
@0xdaryl What about copying (duplicating) a file?

0xdaryl:
To me, that's akin to creating a new empty file and copying new content into it (which should require a forward copyright change because it's a separate entity).  Other opinions?

charliegracie:
Creating an exact copy is the same as moving the file to me, so I wouldn't update the copyright... I think this issue is going to get filled with personal preferences and not facts.  Do the Eclipse Committer webpages provide any guidance around this?

youngar:
I will admit that I don't know what the answer is.  I have been using the original copyright dates of a copied file, even if it is modified after.  I remember reading something that said to do it that way, but I don't remember where and I don't think it was specified by Eclipse.  A quick search didn't really yield me any results. [GNU specifies](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html) "If you copy a file into the package from some other program, keep the copyright years that come with the file."

I think that this change is accurate as-is, but it may be that we want to open an issue for file duplication.

0xdaryl:
Not too much in the [Eclipse Project Handbook](https://www.eclipse.org/projects/handbook/#ip-copyright-headers) on how to update the headers, but they do say the dates in the copyright need to reflect:

> The first year is when the contents of the file were first created and the last year is when the contents were last modified.

If the filename is not part of the contents, then renaming or moving a file doesn't need a copyright update.

It does not provide clarity on what it means to "copy" a file.  I seem to have a philosophical difference on what copying a file means, but one way of interpreting the Eclipse text would be that if the actual contents of two files are the same (regardless of the file name) then the copyright need not change.  If we agree on this then I can amend this PR to add file copying to the list of activities that does not need a copyright update.

charliegracie:
I agree that if the code contents do not change then neither copyright date should be updated. 

IF you copy code from a file into a new file with modifications the initial copyright date should persist from the original file and new last updated copyright date should be update to the current year.

PR #3250, opened 2018-11-27
Initialize Suballocator variables all platforms

dmitripivkine:
Suballocator related variables suballocatorInitialSize and
suballocatorCommitSize in MM_GCExtensionsBase are defined for all
platforms but initialized in class constructor only for Compressed Refs.
It cause operations with non-initialized memory might crash Memory
Check. Extending scope of initialization of these variables for all
platforms. As far as Suballocator happen to be used for Compressed Refs
platforms only it is irrelevant which values these variables are
initialized on other platforms. Set them exactly the same as in CR to
have cleaner code. Again, this would not have any impact for non-CR
platforms footprint.

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -1541,9 +1541,9 @@ public:
 #endif
 		, heapInitializationFailureReason(HEAP_INITIALIZATION_FAILURE_REASON_NO_ERROR)
 		, scavengerAlignHotFields(true) /* VM Design 1774: hot field alignment is on by default */
-#if defined(OMR_GC_COMPRESSED_POINTERS)
 		, suballocatorInitialSize(SUBALLOCATOR_INITIAL_SIZE) /* default for J9Heap suballocator initial size is 200 MB */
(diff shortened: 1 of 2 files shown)


charliegracie:
With these changes, will the GC now request that a sub allocator memory block is allocated for all GC policies since the value will be non-zero?

dmitripivkine:
> With these changes, will the GC now request that a sub allocator memory block is allocated for all GC policies since the value will be non-zero?

No. An initialization of Suballocator `		omrport_control(OMRPORT_CTLDATA_ALLOCATE32_COMMIT_SIZE, extensions->suballocatorCommitSize);
` and allocation `omrmem_ensure_capacity32(extensions->suballocatorInitialSize)` would be called for CR platforms only.

dmitripivkine:
We can set them to 0 but it makes code ugly. I preferred this way because it is more common.
There is one line in OpenJ9 in mminit.cpp line uses SUBALLOCATOR_ALIGNMENT definition enabled for all platforms by this change wrapped for CR only. This wrapping can be removed as well later

charliegracie:
@genie-omr build all

PR #3254, opened 2018-11-28
AArch64: Remove unnecessary ARM64Dep*Instruction classes

knn-k:
This commit removes some ARM64Dep*Instruction classes.
The base OMR::ARM64::Instruction class already has the common
RegisterDependencyConditions field in it, and you don't need to
define separate classes like ARM64DepLabelInstruction from
ARM64LabelInstruction.


knn-k:
New constructors with dependency conditions are added to `ARM64LabelInstruction` and `ARM64ConditionalBranchInstruction` classes.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3255, opened 2018-11-28
MM_Scavenger::aliasToCopyCache Update Comment

RSalman:
Update comment to reflect the changes made to
MM_Scavenger::aliasToCopyCache

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -3132,11 +3132,11 @@ MM_Scavenger::copyCacheDistanceMetric(MM_CopyScanCacheStandard* copyCache)
 MMINLINE MM_CopyScanCacheStandard *
 MM_Scavenger::aliasToCopyCache(MM_EnvironmentStandard *env, GC_SlotObject *scannedSlot, MM_CopyScanCacheStandard* scanCache, MM_CopyScanCacheStandard* copyCache)
 {
-	/* Only alias a copy cache IF there are 0 threads waiting.  If the current thread is the only producer and
-	 * it aliases a copy cache then it will be the only thread able to consume.  This will alleviate the stalling issues
-	 * described in VMDESIGN 1359.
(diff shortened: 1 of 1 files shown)


charliegracie:
@RSalman you can add `[skip ci] to the  first line of the commit message for future commits that only change comments to skip the Travis and AppVeyor builds.

gc/base/standard/Scavenger.cpp
```diff
@@ -3132,11 +3132,11 @@ MM_Scavenger::copyCacheDistanceMetric(MM_CopyScanCacheStandard* copyCache)
...
 {
-	/* Only alias a copy cache IF there are 0 threads waiting.  If the current thread is the only producer and
-	 * it aliases a copy cache then it will be the only thread able to consume.  This will alleviate the stalling issues
-	 * described in VMDESIGN 1359.
+	/* Only alias a copy cache IF there are <= _waitingCountAliasThreshold threads waiting (defaulted to 20% of active GC threads, option XXgc:aliasInhibitingThresholdPercentage).
+	 * If the current thread is the only producer and it aliases a copy cache then it will be the only thread able to consume.
+	 * This will alleviate the stalling issues.
 	 *
-	 * @NOTE this is likely too aggressive and should be relaxed.
+	 * @NOTE See GIT Issue 3089 (Investigate Scavenger's Aliasing Inhibiting Condition)
```
youngar: I personally feel like we shouldn't be referencing external documents from the code.  The important details should be copied into the comments.  I'm not sure if this view is held by everyone.  At the very least you should say `Github issue` instead of `GIT issue`.

RSalman: Changed it to `Github`. 

This is already referencing some very old external documentation (VMDESIGN 1359 - as seen in the PR change), this comment change will remove that but reference the issue #. I was trying to stay consistent with the way it was done before by adding the issue #, however I can remove it if that's what we want.

youngar: Thanks for the fix!

PR #3256, opened 2018-11-28
Update LOAresize to avoid potential issue

LinHu2016:
-add the protection to prevent the potential issue, 
   which expected contract LOA size could be bigger than 
   current LOA size. (resizeLOA() pass "new loa size" as unsigned int
   parameter to method isSizeEnoughForLOA() for checking if the size
   is big enough for LOA, in case calculated new size is smaller than
   0, it could cause isSizeEnoughForLOA() return wrong decision).
 
  -change Xgc:debugLOAResive option to standard trace point 
   for better debugging option

  

charliegracie:
Can you modify the commit message.  The issue that the PR is fixing and the solution is not clear in the commit message.  Also can you remove the "depends on" line since technically the OMR project does not depend on the OpenJ9 project.

charliegracie:
@genie-omr build all

PR #3263, opened 2018-11-29
Remove UT_DIRECT_TRACE_REGISTRATION

charliegracie:


charliegracie:
Marked as WIP until I investigate whether or not any down stream projects were relying on the old behaviour.  Since the old behaviour was a temporary fix for OpenJ9 and included a JavaVM specific header file I believe only the OpenJ9 needs to be investigated.

charliegracie:
This PR makes it easier to consume OMR

0xdaryl:
The compiler and JitBuilder changes look fine to me.

Would you mind adding a commit message that describes why you are doing this (e.g., it is not used anymore), or reference an issue that does?

youngar:
Looks like a jenkins problem made the linux_x86 build fail.
```
Error when executing always post condition:
hudson.remoting.RequestAbortedException: java.io.IOException: Unexpected termination of the channel
```

It failed right after running CMake, which seems weird to me. I'm restarted the build for now.

charliegracie:
@genie-omr build xlinux,x32linux

youngar:
I think this is good to go.  Does any code or build in OpenJ9 use the old tracegen (written in Java?)  It seems there are a couple places in OpenJ9 that could some cleanup, but it's very simple: [UT_DIRECT_TRACE_REGISTRATION](https://github.com/eclipse/openj9/search?q=UT_DIRECT_TRACE_REGISTRATION&unscoped_q=UT_DIRECT_TRACE_REGISTRATION), and [directRegistration](https://github.com/eclipse/openj9/search?q=directRegistration&unscoped_q=directRegistration)

charliegracie:
this code was a temporary “fix” for OpenJ9 when the original split happened.  No other downstream project should require “jni.h”. I tested OpenJ9 and it still compiled and passes all tests with this change.

0xdaryl:
No outstanding objections.  Merging.

PR #3264, opened 2018-11-29
Add comment regarding disabled CS wrtbar optimization

fjeremic:
The TODO is addressed by adding a comment describing why a particular
optimization related to wrtbars is disabled under concurrent scavenge.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -2941,7 +2941,18 @@ TR::Node *constrainWrtBar(OMR::ValuePropagation *vp, TR::Node *node)
 
    if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
       {
-      // TODO (GuardedStorage): Why do we need this restriction?
+      // The optimization below targets the following type of code:
+      // 
+      //     X x = new X();
+      //     x.a = new A();
+      //
+      // Most of the time both newly allocated objects would be in the nursery and there is no need for a generational
+      // barrier. However, in a rare case that a GC occurs in between the two allocations above the premise can be
+      // violated (by tenuring the first, but not the second object). The GC will compensate for the lack of the
+      // barrier by auto-remembering all objects that are tenured and directly stack referenced.
+      //
+      // This optimization is currently disabled when read barriers are required as it is currently unknown whether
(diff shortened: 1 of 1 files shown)


fjeremic:
@amicic could I please get a review here as well to make sure the wording is correct. This is also an FYI that the piece of code _is_ indeed currently disabled in the JIT until we figure out whether it is safe.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2941,7 +2941,18 @@ TR::Node *constrainWrtBar(OMR::ValuePropagation *vp, TR::Node *node)
...
+      //     x.a = new A();
+      //
+      // Most of the time both newly allocated objects would be in the nursery and there is no need for a generational
+      // barrier. However, in a rare case that a GC occurs in between the two allocations above the premise can be
+      // violated (by tenuring the first, but not the second object). The GC will compensate for the lack of the
+      // barrier by auto-remembering all objects that are tenured and directly stack referenced.
+      //
+      // This optimization is currently disabled when read barriers are required as it is currently unknown whether
+      // this optimization is safe under such a GC mode.
       doOpt = false;
```
vijaysun-omr: Is it possible to be sure(r) one way or another ? I am guessing it's not an easy case to reason through due to complexities with the CS implementation but wanted to understand if that is in fact the case or we are reasonably sure it is not a safe transformation in this GC mode.

amicic: At this point I don't have better info than: I'm unsure if it safe. Some testing in indicated it was not, but don't remember details. I do plan to revisit it.

Another aspect is if this is a desirable optimization at all in the context of low pause GCs, like Concurrent Scavenge is... As a part of 'compensate' action, GC will do rescan of all threads (all done STW),  once the need for auto-remembering is detected. Again, I don't have much info how often that will occur.

amicic:
It's ok comment best to my knowledge. Although, just realized one detail.... 'second object not tenured'. At the point of GC, second object has not been allocated/constructed yet, that's why GC was triggered. So it was a bit silly to call it an object, and claim it was not tenured for the particular scenario.

vijaysun-omr:
Change only added comments to the code. I am going to merge now that basic checks have passed.

Issue #2886, opened 2018-08-22
Performance regression with setOption(TR_DisableInternalPointers)


harryyu1994:
The following code snippet (from omr/compiler/compile/OMRCompilation.cpp) gives us performance regression when `TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);` is changed to `self()->setOption(TR_DisableInternalPointers);`.
```
// Temporarily exclude PPC due to perf regression
      if( (self()->getMethodHotness() <= warm))
         {
         if (!TR::Compiler->target.cpu.isPower())
            TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);
         }
      self()->getOptions()->setOption(TR_DisablePartialInlining);
```

In theory, `self()->setOption(TR_DisableInternalPointers)` is the correct way to do this as it only sets the TR_DisableInternalPointers bit for the current compilation. Using `TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);` is a bug because changing the global TR::Options object means changing the options for all the compilations that follow. The per-compilation TR::Options object is constructed based on the global TR::Options object. 

However, performance results have shown that correcting this code leads to 4% performance regression on X. While removing this line `TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);` also gives us a slight performance regression as well. Further investigation is needed.

0xdaryl:
@nbhuiyan , fyi for your compiler options rework.

PR #3266, opened 2018-11-29
Restrict setting TR_DisableInternalPointer to current compilation

nbhuiyan:
Previously, this option was set only when `OMR::Compilation::compile()`
was called. That resulted in the option not having an effect on
CodeGenerator's `_disableInternalPointers` member as that is
initialized much earlier in its constructor by checking a few
compilation-specific options, including `TR_DisableInternalPointers`.

For this reason, not setting `TR_DisableInternalPointer` option
globally was resulting in performance regression in some cases.
Setting this option  globally is not the correct thing to do,
as this causes the option to be applied to all subsequent
compilations regardless of whether it should be set.

By setting `TR_DisableInternalPointer` to the current
compilation-specific options in the the constructor body of
`OMR::Compilation` before `allocateCodeGenerator(TR::Compilation)`
is called, we no longer need to set this option globally.

Fixes: #2886

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -371,6 +371,18 @@ OMR::Compilation::Compilation(
          );
    _isServerInlining = !options.getOption(TR_NoOptServer);
 
+   // TR_DisableInternalPointers must be set before allocateCodeGenerator(self()) is called because
+   // CodeGenerator's _disableInternalPointers member is set in its constructor and this is one of
+   // options that is checked for
+   if (_isOptServer)
+      {
+      if(self()->getMethodHotness() <= warm)
+         {
+         if (!TR::Compiler->target.cpu.isPower()) // Temporarily exclude PPC due to perf regression
+            self()->setOption(TR_DisableInternalPointers);
+         }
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build all

compiler/compile/OMRCompilation.cpp
```diff
@@ -930,13 +942,7 @@ int32_t OMR::Compilation::compile()
...
    if (self()->isOptServer())
       {
-      // Temporarily exclude PPC due to perf regression
-      if( (self()->getMethodHotness() <= warm))
-         {
-         if (!TR::Compiler->target.cpu.isPower())
-            TR::Options::getCmdLineOptions()->setOption(TR_DisableInternalPointers);
-         }
-      self()->getOptions()->setOption(TR_DisablePartialInlining);
+         self()->setOption(TR_DisablePartialInlining);
```
0xdaryl: Please fix indenting to be style consistent.

nbhuiyan: done

0xdaryl:
@genie-omr build all

PR #3271, opened 2018-11-30
AArch64: Add is64bit parameter to generateNegInstruction()

knn-k:
This commit adds is64bit parameter to generateNegInstruction().
I found a case where the Node has the type Int64 while the
intended operation is 32-bit negation.


0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3272, opened 2018-11-30
Add Bytes utility header

rwy0717:
Add a set of convenience functions for working with bytes. Useful for
specifying large sizes like heap minimum size, or maximum memory
overhead.

The kibi/mebi/gibi prefix was chosen to minimize confusion, some people
might consider a kilobyte to be 1000 or 1024 bytes, but a kibibyte is
unambiguously 1024 bytes.

### DIFF
--- a/fvtest/coretest/main.cpp
+++ b/fvtest/coretest/main.cpp
@@ -1,0 +1,28 @@
+/*******************************************************************************
+ * Copyright (c) 2014, 2018 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR
+ *LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#include "omrTest.h"
+
+extern "C" int omr_main_entry(int argc, char** argv, char** envp) {
+	::testing::InitGoogleTest(&argc, argv);
+	return RUN_ALL_TESTS();
+}
--- a/fvtest/CMakeLists.txt
+++ b/fvtest/CMakeLists.txt
@@ -27,6 +27,7 @@ add_subdirectory(util)
 
 add_subdirectory(omrGtestGlue)
 add_subdirectory(algotest)
+add_subdirectory(coretest)
 add_subdirectory(porttest)
 add_subdirectory(rastest)
 add_subdirectory(sigtest)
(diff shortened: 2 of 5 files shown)


0xdaryl:
Perhaps this should live in the `util` directory (or some kind of `core` directory) so other OMR components could avail themselves of it if they need to?

rwy0717:
How about I put this header at `include_core/OMR/Bytes.hpp`, which would be used as `#include <OMR/Bytes.hpp>`.

0xdaryl:
Yes -- in general, I'm fine with this living somewhere under `include_core`.

So `OMR` is a new subdirectory of `include_core` that will be used for such utility files in the future?  I'm just trying to understand the naming convention.

rwy0717:
The intent is to scope our public headers under an `OMR/*` directory, rather than use an `OMR_*` prefix. So rather than doing `#include <Bytes.hpp>` or `#include <OMR_Bytes.hpp>`, programmers would do `#include <OMR/Bytes.hpp>`.

Leonardo2718:
This may be beyond the scope of this PR's intention, but I think using an approach similar to [`std::chrono::duration`](https://en.cppreference.com/w/cpp/chrono/duration) (using [`std::ratio`](https://en.cppreference.com/w/cpp/numeric/ratio/ratio)) would be a good fit for what these functions are intended to do.

Maybe for future work? :slightly_smiling_face:

rwy0717:
Using std::ratio is an interesting idea. While I would agree that it's outside the scope of this PR, I'd still like to explore the idea 👍 

One thing to get out of the way now is, it's very unlikely that every compiler will provide the ratio utility, so it's something we'd have to recreate ourselves. Luckily, I don't see that strongly affecting the interface.

I have to admit, I've never found std::ratio to be obvious or easy to use. Especially around negative or nonintegral counts and ratios. With a ratio-based size type, it might be possible to define a -0.5 MiB size, or use a 1/4 byte unit. Is this a good thing? The improved expressibility of the type might not be worth all the new kinds of errors we're creating.

One very nice thing about duration is, the unit is encoded in the type. Unfortunately, the benefit is diminished when the size is immediately converted to a size-in-bytes. I don't foresee people keeping their scaled sizes around.

I'm still trying to think up a compelling use case for ratio-based sizes. In the meantime I wrote a [strawman implementation](https://gcc.godbolt.org/z/7DS6Yu) of a ratio-based API. The count is always a size_t, for simplicity's sake, but that can change. [Here](https://gcc.godbolt.org/z/XJIG-J) is a version that relies on integral size factors, rather than ratios. This implementation trades off on flexibility for ensuring the size is always a nonnegative integer.

0xdaryl:
Thanks.  I'll never turn down an offer for unit tests, assuming you were offering.  :-)

rwy0717:
Are we comfortable if I add a new group of tests called coretest, and put the Bytes tests there?

0xdaryl:
Fine with me!

fvtest/coretest/main.cpp
```diff
@@ -0,0 +1,28 @@
+/*******************************************************************************
+ * Copyright (c) 2014, 2018 IBM Corp. and others
```
rwy0717: Since this is a copy of an older file, this copyright date was unmodified.

rwy0717:
Updated with a few simple tests, just enough to exercise the API. Added the coretest directory and "standard" helpers for creating unit tests.

rwy0717:
Minor stylistic updates:
* line wrap before braces around `TEST` bodies
* use `stdint.h`, not `cstdint`

youngar:
@genie-omr build all

youngar:
```
include_core/OMR/Bytes.hpp:32:1: error: 'constexpr' does not name a type
constexpr size_t
^
include_core/OMR/Bytes.hpp:32:1: note: C++11 'constexpr' only available with -std=c++11 or -std=gnu++11
```

rwy0717:
I added constexpr as a required feature to the tests (The test is only built + run with cmake). That should fix it. Originally, this header & test was going into the GC API, which would turn on C++0x.

rwy0717:
Stopped using constexpr, not available on MSVC 2012 (version 11) (first available in MSVC 2015). Removed the cmake compile feature cxx_constexpr as well.

Leonardo2718:
@genie-omr build all

rwy0717:
OK, this is ready for another build.

Leonardo2718:
@genie-omr build all

PR #3274, opened 2018-12-03
Add development flags for context and object scanner

rwy0717:
These are temporary development flags.


charliegracie:
I renamed the PR since it is tagged as a GC item the PR title does not need to include `gc`.

charliegracie:
@genie-omr build all

PR #3276, opened 2018-12-03
Fix coding style

keithc-ca:
* add missing parentheses
* space around binary operators and after keywords
* r-value on left of !=


youngar:
@genie-omr build all

PR #3277, opened 2018-12-03
Fix cache trashing in _shouldYield

amicic:
Set the value of the flag only if it really changes (from false to
true).

The flag was being written on every check from multiple threads/CPUs,
causing cash trashing and slowing down GC with plafroms with CS enabled.
The flag (regression) was introduced here:
https://github.com/eclipse/omr/pull/3205.


charliegracie:
@genie-omr build all

charliegracie:
Once the PR testing has passed any committer can feel free to merge

PR #3278, opened 2018-12-03
Introduce a CG phase to simplify instructions

0dvictor:
Instruction Simplification Phase provides opportunities after
Instruction Selection to simplify certain patterns that may not be
trivial in the evaluators.

### DIFF
--- a/compiler/codegen/CodeGenPhaseToPerform.hpp
+++ b/compiler/codegen/CodeGenPhaseToPerform.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -30,6 +30,7 @@
     SetupForInstructionSelectionPhase,
     RemoveUnusedLocalsPhase,
     InstructionSelectionPhase,
+    InstructionSimplificationPhase,
     CreateStackAtlasPhase,
     RegisterAssigningPhase,
     MapStackPhase,
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -312,6 +312,37 @@ OMR::CodeGenPhase::performBinaryEncodingPhase(TR::CodeGenerator * cg, TR::CodeGe
    }
 
 
+void
+OMR::CodeGenPhase::performInstructionSimplificationPhase(TR::CodeGenerator * cg, TR::CodeGenPhase * phase)
+   {
+   TR::Compilation* comp = cg->comp();
+   bool trace = comp->getOption(TR_TraceCG);
+   phase->reportPhase(InstructionSimplificationPhase);
+
+   TR::LexicalMemProfiler mp(phase->getName(), comp->phaseMemProfiler());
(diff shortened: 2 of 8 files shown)


vijaysun-omr:
How is this related to any other "peepholing" pass that might already exist in the codegen (perhaps implemented only on other platforms) ?

0dvictor:
The existing Peephole Phase performs optimization after register assignment, while the proposed Instruction Selection Phase performs optimization before register assignment. They are for different scenarios. The major advantage of doing simplification before register assignment is that it can possibly reduce the number of registers required and hence lead to better register assignment.
One optimization fits in this phase in my mind is to use EFLAG information to eliminate sequences like:
```
SETcc GPR_A
MOVZX GPR_B, GPR_A
TEST GPR_B, GPR_B
JNZ LABEL
```
to
`Jcc LABEL`

vijaysun-omr:
I think input from @0xdaryl @zl-wang @andrewcraik would be good to have since there may be philosophical reasons such a codegen pass was not implemented historically.

andrewcraik:
@0dvictor what is the status on this?

0dvictor:
Yes, this proposal is indeed the equivalent of Z's current "Pre-RA Peephole" phase.
This PR supersets most of the Z CodeGen's infrastructure introduced by https://github.com/eclipse/omr/commit/396efd006bebafdd34f26b0bd799bbebce2dbe8e, which allows other platform, for example X86, sharing the same framework to apply similar CodeGen level transformation/optimization.

0dvictor:
FYI, I'd like to name this proposed phase "Instruction Simplification" rather than "Pre-RA Peephole" mainly because I'd like to avoid indications of the phase's relation to RA from its name. "Pre-RA Peephole" sounds like the phase immediately before RA; however, theoretically there could be other phases between this Instruction Simplification phase and RA. The best suitable place should be immediately after Instruction Selection; and hence "Post-InstructionSelection Peephole" would be a more suitable name to precisely describe this proposed phase than "Pre-RA Peephole". "Post-InstructionSelection Peephole" is a long and tedious name, so that I think "Instruction Simplification" would be better as it describes this phase's intention and avoids incorrect indications from its relationship to RA.

fjeremic:
> FYI, I'd like to name this proposed phase "Instruction Simplification"

Sounds reasonable to me.

0xdaryl:
I don't quite understand the reasoning not to call this a "peephole" optimization, a well-understood term in compiler theory, literature, and existing compiler implementations.  Will this phase do something different that precludes calling it that?  I'd rather not stray from the popular vernacular unless there's a good reason.

I do understand and support removing the other "phase" information from the name.

0dvictor:
> I don't quite understand the reasoning not to call this a "peephole" optimization, a well-understood term in compiler theory, literature, and existing compiler implementations. Will this phase do something different that precludes calling it that? I'd rather not stray from the popular vernacular unless there's a good reason.

This proposed phase is exactly the peephole optimization in compiler theory. The reason why I proposed a different name is because the name "Peephole Phase" has been taken and used as the peephole phase between Register Assignment and Binary Encoding. Therefore this proposed a peephole phase between Instruction Selection and Register Assignment needs a new name.

0xdaryl:
I would like to see some harmonization of the naming of the peephole pass(es) in the code generator.  Just because one pass has used the term "peephole" does not preclude its use for others.

As this PR has been inactive for a while, it requires work to be merge-ready, and the contributor has moved on to other things and is unlikely to build upon the skeleton work proposed here, I am going to close this PR now.

PR #3283, opened 2018-12-04
Disable OMR_THR_YIELD_ALG on MacOS

babsingh:
OMR_THR_YIELD_ALG should only be used with Completely Fair Scheduler
(CFS). CFS is only available on Linux.

MacOS schedulers can be classified as multilevel feedback queue
schedulers. MacOS doesn't use CFS. Thus, disabling OMR_THR_YIELD_ALG on
MacOS.

### DIFF
--- a/example/glue/configure_includes/configure_osx.mk
+++ b/example/glue/configure_includes/configure_osx.mk
@@ -39,7 +39,6 @@ CONFIGURE_ARGS += \
   --enable-OMR_TEST_COMPILER \
   --enable-OMR_THR_FORK_SUPPORT \
   --enable-OMR_THR_THREE_TIER_LOCKING \
-  --enable-OMR_THR_YIELD_ALG \
   --enable-OMR_GC_ARRAYLETS \
   --enable-OMR_THR_SPIN_WAKE_CONTROL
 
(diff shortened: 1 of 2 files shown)


babsingh:
fyi - @charliegracie @DanHeidinga

babsingh:
Related to https://github.com/eclipse/openj9/issues/36

charliegracie:
The `OMR_THR_YIELD_ALG` flag is now being turned off for OS X in the auto-tools build system. The auto-tools build system is not used for OS X testing in OMR. When looking into this change I noticed that ALL linux platforms are enabling this option in the auto-tools build system. Again the default OMR CI testing uses CMake on all Linux platforms so technically OMR is not testing this feature at all.  Should this be enabled by default on all Linux platforms?

babsingh:
> Should `OMR_THR_YIELD_ALG` be enabled by default on all Linux platforms?

`OMR_THR_YIELD_ALG` introduces optimizations for Completely Fair Scheduler
(CFS). Since Linux uses CFS, `OMR_THR_YIELD_ALG` should be enabled by default on all Linux platforms.

charliegracie:
No further CI testing required since only the OS X auto-tools build system was modified and there is not CI testing that uses it.

PR #3286, opened 2018-12-05
Add an extension point for optimizations of direct load in VP

liqunl:
Add transformDirectLoad where downstream project can add language
specific optimizations of direct load.

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.cpp
+++ b/compiler/optimizer/OMRValuePropagation.cpp
@@ -7165,6 +7165,19 @@ void OMR::ValuePropagation::constrainRecognizedMethod(TR::Node *node)
    }
 
 
+/** \brief
+ *     Extension point for language specific optimizations on direct load
+ *
+ * \parm node
+ *     The direct load node to be constrained
+ *
+ * \return
+ *     True if node is transformed, false otherwise
+ */
+bool OMR::ValuePropagation::transformDirectLoad(TR::Node *node)
+   {
+   return TR::TransformUtil::transformDirectLoad(comp(), node);
+   }
 
 bool OMR::ValuePropagation::checkAllUnsafeReferences(TR::Node *node, vcount_t visitCount)
    {
--- a/compiler/optimizer/OMRValuePropagation.hpp
+++ b/compiler/optimizer/OMRValuePropagation.hpp
@@ -589,6 +589,7 @@ class ValuePropagation : public TR::Optimization
    void transformReferenceArrayCopy(TR_TreeTopWrtBarFlag *);
    void transformReferenceArrayCopyWithoutCreatingStoreTrees(TR_TreeTopWrtBarFlag *arrayTree, TR::SymbolReference *srcObjRef, TR::SymbolReference *dstObjRef, TR::SymbolReference *srcRef, TR::SymbolReference *dstRef, TR::SymbolReference *lenRef);
    virtual void constrainRecognizedMethod(TR::Node *node);
+   virtual bool transformDirectLoad(TR::Node *node);
 
    struct ObjCloneInfo {
       TR_ALLOC(TR_Memory::ValuePropagation)
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -1564,7 +1564,7 @@ TR::Node *constrainAload(OMR::ValuePropagation *vp, TR::Node *node)
       return node;
 
    // before undertaking aload specific handling see if the default load transformation helps
-   if (TR::TransformUtil::transformDirectLoad(vp->comp(), node))
+   if (vp->transformDirectLoad(node))
       {
       constrainNewlyFoldedConst(vp, node, false /* !isGlobal, conservative */);
       return node;


liqunl:
@andrewcraik Can I have a review?

liqunl:
Instead of adding a hook that does similar but not identical things to `constrainAload`, I made `constrainAload` a virtual function of VP. Downstream project can override this function to provide their own implementation. Since most handlers are not member functions, a non-member function is needed to wrap the virtual one. The macro `VPHandler(name)` is introduced to define such function. The function defined by the macro just calls to the VP member function with the same name. When most handlers are moved to VP, we can consider calling the member functions directly via member function pointers.

@vijaysun-omr @mstoodle  Could you take a look and see if you're ok with the design? The change and commit message are incomplete, it is just to show you how the design look.

liqunl:
Yes, Openj9's implementation will have most of the code today's OMR implementation has. It can call the OMR's implementation at the end of its implementation to reduce the code duplication in the new design.

mstoodle:
Did you consider just putting a virtual call at the point of the original `J9_PROJECT_SPECIFIC` code that could be empty in OMR but filled in for OpenJ9? That might make it easier to augment what's in OMR without also having to take that same code to OpenJ9.

liqunl:
The purpose of this PR is to have an extension point to host my code for guarded static final field folding, a new feature I'm trying to add for Openj9. I originally added a virtual call `constrainDirectLoad` to replace `TR::TransformUtil::transformDirectLoad` at the beginning of `constrainAload`. But after discussion with @andrewcraik @jdmpapin , we think having two APIs doing similar things is too messy and hard to maintain. If anyone wants to add new opts for aload, they might be confused about where to put their opts. With this design, only one API is explicitly declared and it's clear where the opts for aload belong.

mstoodle:
Maybe I'm misunderstanding the origin of confusion, but the proposed solution seems to favour projects providing special Aload constraints rather than favouring the building of a "common" set of Aload constraints in OMR that can be easily reused by other projects (including OpenJ9).

I think if you establish a naming convention up front, it won't be too terribly confusing to manage having two functions (one that's in OMR and is responsible for calling the second one).

@andrewcraik @jdmpapin what do you think? I don't really want to establish a pattern where a project absorbs OMR by first copying everything in OMR and then having to periodically re-copy things any new things that get added to OMR....

andrewcraik:
So my concern with putting a handle in the middle of the constrainALoad is that almost the entire content of constrainAload should move to OpenJ9 - there is only a small bit of code at the start and at the end. If we proceed with adding a hook of that kind it suggests that we really need one for almost every constrain where any kind of project specific override is needed - super ugly. The handlers being flat functions as they are means that it is not easy to extend the in the child. One way would be to make them all member functions and put member functions in the table, but that is a big change and not something we really want to do wholesale. Another way is to make the handler implementation virtual so you can override it and call back up to OMR to do the OMR stuff and then do what you want to do either before or after (when there isn't a very specific hook we would want to add deep in a handler). That kind of thinking resulted in this kind of approach to try and find a nice way to let a project replace a handler with a way to call back to the OMR handler if it wants to. I'm really not a fan of a hook which represents pretty much the whole handler. Note that this approach does not require the downstream project to copy OMR's implementation it can call up to its super's implementation.

mstoodle:
@andrewcraik said:
> Note that this approach does not require the downstream project to copy OMR's implementation it can call up to its super's implementation.

I'm not sure how you can make that 100% faithfully work given the current ordering of the current bits: there's something that's clearly documented as "supposed to go first" and another bit that seems like it should really go last (because the middle part can return). Or are you suggesting these can be rearranged?

> If we proceed with adding a hook of that kind it suggests that we really need one for almost every constrain where any kind of project specific override is needed - super ugly

I would prefer we treat these case-by-case and try hard to reorganize the code towards the simplest possible project vs. OMR ordering (i.e. project "specials" first, then OMR or vice versa). Project-specific ordering constraints are ugly, but if we rule them out entirely then you're going to have to reorder some things introducing potential for regressions in OpenJ9. For example, we could reorder the code in the current function along the lines that you're suggesting and test for regressions...

Or are OpenJ9 regressions due to refactoring just a consequence you're willing to accept a priori to avoid any ugliness :) ?

andrewcraik:
Since @liqunl is just trying to get a specific bit of functionality in and not grow the J9 code in VP in OMR let's just go with the hook in the middle of the function for now which is uncontroversial so she can do what she needs to do while we work through the design of VP's extension points.

In reply to the above:
> Or are you suggesting these can be rearranged?
I think some of that could be rearranged - some of the if the node is null / non-null handle the constraints are a bit out of place and we could clean that up.

> Or are OpenJ9 regressions due to refactoring just a consequence you're willing to accept a priori to avoid any ugliness :)
Regressions are not something we want to cause, but I am also not convinced a sea of mid-method hooks is going to be easy for anyone to wrap their head around and may well lead to bugs and regressions if the flow of code is woven between two files in two projects ...

I agree that we want to keep large bulks of functionality together and keep as much in OMR as possible. I think this is an area worthy of discussion in an architecture meeting at some point - @0xdaryl could you add it to the topic list?

liqunl:
@mstoodle `constrainAload` is written in a way that the project specific stuff has to go in the middle. It can be rearranged with a little bit of refactoring. I'll cook up the code in a bit.

0xdaryl:
I will add this topic to tomorrow's CAM agenda.

liqunl:
As per discussion from @andrewcraik @mstoodle, this area needs more thinking and discussion, we'll leave it to the architecture meeting and go with the hook implementation for now.

liqunl:
@andrewcraik @mstoodle Could you review the hook implementation?

mstoodle:
@genie-omr build all

andrewcraik:
Looks good to me too

mstoodle:
@vijaysun-omr as assignee, are you ok with this PR?

liqunl:
@vijaysun-omr ping?

vijaysun-omr:
Looking at the commit in the end, it looks good to me. The function is the same at the OMR layer and a "virtual" extension point has been created for downstream projects.

Issue #2084, opened 2017-11-30
Document proper use of the TR:: namespace


0xdaryl:
Provide documentation (under `doc/compiler`) describing the expected use cases for the `TR::` namespace in the context of extensible classes and in a more general compiler context.

mgaudet:
[This is currently in doc/compiler/Extensible_Classes_in_OMR.md](https://github.com/eclipse/omr/blob/master/doc/compiler/extensible_classes/Extensible_Classes_in_OMR.md)

> We use the TR:: namespace to serve this purpose. Outside of the layered construction of an extensible class, all compiler code uses the TR:: namespace (rooted in the code name for the compiler technology which was Testasrossa).
> 
> In the case of extensible classes, we expose them to consumers by providing a terminating subclass in the TR:: namespace, which we call the Concrete class.
> 
> In the case of classes that aren't extensible, It is perfectly legitimate to create a class directly in the TR:: namespace if it's complete and can and should be used as-is.

Not sufficient?

mgaudet:
(This feels like a dupe of #788)

0xdaryl:
@mgaudet:  yes, the discussion that preceded those updates to the Extensible Class documentation slipped my mind.  I think I want to surface that text somewhere other than the Extensible Classes documentation though as the `TR::` namespace has uses beyond it.  I'll keep this issue open in the meantime.

PR #3287, opened 2018-12-05
Consolidate all register iterators

samolisov:
Consolidate implementations of RegisterIterator

Each code generator implements a `RegisterIterator` class which appears
to be implemented solely to assist register assignment tracing dumps in
`ras/Debug.cpp`. The DNA of these classes across all the code generators
must be at least 90% identical, and there is no doubt that with a bit of
thought a single, common implementation shared by all code generators
could be developed.

A generic implementation of the `TR::RegisterIterator` is implemented
and all platform-specific `OMR::ARCH::RegisterIterator` classes are
removed from the codebase. The signature of the class constructor is
changed a bit: the constructor takes the machine as well as the first
and last assignable registers, no TR_RegisterKinds is used:

```cpp
self()->setGPRegisterIterator(new (self()->trHeapMemory()) TR::RegisterIterator(self()->machine(),
    TR::RealRegister::FirstGPR, TR::RealRegister::LastAssignableGPR));
self()->setFPRegisterIterator(new (self()->trHeapMemory()) TR::RegisterIterator(self()->machine(),
    TR::RealRegister::FirstXMMR, TR::RealRegister::LastXMMR));
```

Closes: #2645


samolisov:
@fjeremic @0xdaryl I've opened a new PR where compleated task #2645. Because in the previous PRs related to the task I didn't remove the `getXXXRealRegister()` methods, the downstream projects should feel any problems. Once the PR is merged, the related PR in OpenJ9 (eclipse/openj9#2936) will be ready to review and then the wrappers may be removed from the codebase.

compiler/codegen/OMRMachine.hpp
```diff
@@ -68,11 +68,22 @@ class OMR_EXTENSIBLE Machine
...
+      {
+      return _registerFile[regNum];
+      }
+
    /**
     * \return : the cached TR::CodeGenerator object
     */
    TR::CodeGenerator *cg() {return _cg;}
 
+
```
fjeremic: Extra new line here.

compiler/codegen/OMRRegisterIterator.cpp
```diff
@@ -19,47 +19,28 @@
...
-#include "codegen/RealRegister.hpp"
-#include "codegen/Register.hpp"
-#include "codegen/RegisterIterator.hpp"
-#include "infra/Assert.hpp"
+#include <stddef.h>                        // for NULL
+#include "codegen/Machine.hpp"             // for Machine
+#include "codegen/RealRegister.hpp"        // for RealRegister, etc
+#include "codegen/Register.hpp"            // for Register
+#include "codegen/OMRRegisterIterator.hpp" // for RegisterIterator
+#include "infra/Assert.hpp"                // for TR_ASSERT
```
fjeremic: I never really found these comments particularly useful and they have a high maintenance with code moving around. How do other committers feel about not including them?

compiler/codegen/OMRRegisterIterator.hpp
```diff
@@ -22,36 +22,61 @@
...
- * This interface must be implemented by platform-specific descendents,
- * using either List and ListIterator, or some cheaper alternatives.
+/**
+ * \brief
+ * An implementation of the iterator pattern to allow users (for example RAS)
+ * traverse the platform-specific list of a certain kind of registers (GPR,
+ * FPR, VRF, etc.)
  */
-class OMR_EXTENSIBLE RegisterIterator
+class RegisterIterator
```
fjeremic: One byproduct of this change is that the class is no longer extensible. I don't particularly see a problem with this and it can be extended at a later point if needed, but wanted to explicitly point this out to other committers.

compiler/codegen/OMRMachine.hpp
samolisov: Thanks. Fixed

compiler/codegen/OMRRegisterIterator.cpp
samolisov: I'm don't mind removing the comments, no problem. @Leonardo2718 @0xdaryl @0dvictor What is your opinion about comments after includes, do we need them there?

compiler/codegen/OMRRegisterIterator.hpp
samolisov: In issue #2645 @0xdaryl said:

> It's also not clear to me why this was made an extensible class in the first place. It's hardly a class I would think one would want to extend.

So, as I get, there is no reason to mark the class as `OMR_EXTENSIBLE`.

compiler/codegen/OMRRegisterIterator.cpp
Leonardo2718: I'm okay with removing them. I've never found these kinds of comments useful either.

compiler/codegen/OMRRegisterIterator.hpp
```diff
@@ -22,36 +22,61 @@
...
-namespace OMR { class RegisterIterator; }
-namespace OMR { typedef OMR::RegisterIterator RegisterIteratorConnector; }
-#endif
-
 #include "env/TRMemory.hpp"  // for TR_Memory, etc
+
+namespace TR { class Machine; }
 namespace TR { class Register; }
 
 namespace OMR
```
Leonardo2718: Given `RegisterIterator` is no longer an extensible class, I think its implementation should be moved to the `TR` namespace, like the other non-extensible classes.

samolisov: @Leonardo2718 Initially I've implemented `RegisterIterator` in the `TR` namespace but @0xdaryl suggested to move it into the `OMR` one (please have a look at https://github.com/eclipse/omr/pull/2981#discussion_r222771376) Fortunately, I've used a separate commit for these changes, so if you think the class must be in the `OMR` namespace, I can just reset the commit.

@0xdaryl @Leonardo2718 @fjeremic Which namespace should be the home for `RegisterIterator`?

Leonardo2718: Hm... alright.

According to our current documentation in [Extensible_Classes_in_OMR.md](https://github.com/eclipse/omr/blob/master/doc/compiler/extensible_classes/Extensible_Classes_in_OMR.md#namespaces) (emphasis mine):

>We use the TR:: namespace to serve this purpose. Outside of the layered construction of an extensible class, all compiler code uses the TR:: namespace (rooted in the code name for the compiler technology which was Testasrossa).
>
>In the case of extensible classes, we expose them to consumers by providing a terminating subclass in the TR:: namespace, which we call the Concrete class.
>
>**In the case of classes that aren't extensible, It is perfectly legitimate to create a class directly in the TR:: namespace if it's complete and can and should be used as-is**.

Has there been some discussion to change this that I've missed? I see #2084 is still open, but it hasn't had any activity in quit some time. :slightly_smiling_face:

fjeremic: I thought `TR::` was reserved _for_ extensible classes and `OMR::` for non-extensible. I think @samolisov's change here is correct.

compiler/codegen/OMRRegisterIterator.cpp
samolisov: I've removed the comments from the files touched by PR.

compiler/codegen/OMRRegisterIterator.hpp
samolisov: @Leonardo2718 do you agree with @fjeremic about the namespace?

Leonardo2718: It seems like a deeper discussion is needed to figure out what namespace we want non-extensible classes to be in, so I've opened #3306.

In the mean time, I think this PR shouldn't have to wait for that discussion to conclude and should just follow our currently documented guideline, which is to use the `TR` namespace (at least, that's how I'm interpreting the paragraph I highlighted above).

@fjeremic does this sound reasonable?

samolisov: @Leonardo2718 @fjeremic OK, I've moved `RegisterIterator` back to the `TR` namespace. The class may be moved to the `OMR` namespace as well as all the related ones when the decision is made. If you have no more objections, please approve the PR.

compiler/codegen/RegisterIterator.hpp
```diff
@@ -22,20 +22,61 @@
...
+namespace TR { class Register; }
 
 namespace TR
 {
 
-class OMR_EXTENSIBLE RegisterIterator : public OMR::RegisterIteratorConnector
+/**
+ * \brief
+ * An implementation of the iterator pattern to allow users (for example RAS)
+ * traverse the platform-specific list of a certain kind of registers (GPR,
```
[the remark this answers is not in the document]
samolisov: @Leonardo2718 thank you for the suggestion, I've edit the commit in this way.

samolisov:
All builds have been passed. @Leonardo2718 @fjeremic can someone merge the PR?

compiler/codegen/OMRRegisterIterator.cpp
0xdaryl: Those `#include` comments are remnants from when we ran Include-What-You-Use over the code base.  They are impossible to maintain and keep up-to-date manually, and really only suitable if we regularly run IWYU on our code (which we don't).  I'm in favour of removing them.

Leonardo2718:
@0xdaryl can you approve this PR as well, since you authored #2645? I want at least two people with codegen experience to approve this because it touches all the codegens.

PR #2778, opened 2018-07-20
AArch64: Use OMR::ARM64::InstOpCode instead of TR_ARM64OpCode

knn-k:
Change the name space TR_ARM64OpCode:: to OMR::ARM64::InstOpCode::
for aligning to OMR style.

Issue: #2754

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #3290, opened 2018-12-06
AArch64: Remove ARM64Ops.hpp

knn-k:
This commit removes the file ARM64Ops.hpp, which is no longer used.


knn-k:
The contents of `ARM64Ops.hpp` were moved to `OMRInstOpCodeEnum.hpp` by #2778.

PR #3292, opened 2018-12-06
AArch64: Fix register allocation for FP xload

knn-k:
This comment fixes register allocation for fload/dloadEvaluator()
for aarch64 so that they use VFP registers.


0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3293, opened 2018-12-06
Fix wrtbar loop versioning range check trees

yanluo7:
- Use the correct aconst node creation API Node::aconst to ensure
the correct address-type constant is represented. The current API
Node:create takes in a 32-bit int value, which results in 64-bit
addresses being truncated and sign-extended.
- Fix erroneous logic on X86 where erroneously we always generate
compile time constant for tenure range boundaries.


Leonardo2718:
@genie-omr build all

andrewcraik:
The previous code is clearly buggy - if the heap base is not fixed the bounds cannot be treated as compile-time constants. The change in node constructor is necessary to make sure the constant is not truncated or sign extended incorrectly. LGTM @vijaysun-omr

vijaysun-omr:
Fixes fix error in the logic on X86 and use the right api regardless of 32/64 bit platforms.

PR #3299, opened 2018-12-06
Fix warnings in multiple (mostly compiler/optimizer) files

mstoodle:
Newer compilers complain about a few things that are not syntactically
incorrect, but might be bugs. For example, not wrapping && expressions
in parenthesis when used in || conditionals. This commit fixes a number
of such warnings found primarily in the compiler/optimizer files.

### DIFF
--- a/compiler/optimizer/VirtualGuardHeadMerger.cpp
+++ b/compiler/optimizer/VirtualGuardHeadMerger.cpp
@@ -340,9 +340,9 @@ int32_t TR_VirtualGuardHeadMerger::perform() {
             {
             // TODO handle moving code earlier in the block down below the guard
             // tail split
-            if ((block->getNextBlock()->getSuccessors().size() == 1) ||
-                ((block->getNextBlock()->getSuccessors().size() == 2) &&
-                 block->getNextBlock()->getLastRealTreeTop()->getNode()->getOpCode().isBranch()) &&
+            if ((block->getNextBlock()->getSuccessors().size() == 1 ||
+                 (block->getNextBlock()->getSuccessors().size() == 2 &&
+                  block->getNextBlock()->getLastRealTreeTop()->getNode()->getOpCode().isBranch())) &&
                 performTransformation(comp(), "%sCloning block_%d and placing clone after block_%d to reduce HCR guard nops\n", OPT_DETAILS, block->getNextBlock()->getNumber(), cold1->getNumber()))
                tailSplitBlock(block, cold1);
             }
--- a/compiler/x/codegen/OMRMemoryReference.cpp
+++ b/compiler/x/codegen/OMRMemoryReference.cpp
@@ -523,8 +523,13 @@ OMR::X86::MemoryReference::populateMemoryReference(
        subTree->getSymbolReference()->getSymbol()->isMethodMetaData())
       evalSubTree = false;
 
-   if (evalSubTree &&
-       subTree->getReferenceCount() > 1 || subTree->getRegister() != NULL || (self()->inUpcastingMode() && !subTree->cannotOverflow()))
+   // This condition was previously formatted in a misleading way given the
+   //  conditional being evaluated. I believe the condition is correct (rather,
+   //  I have no strong evidence that it is incorrect other than the formatting)
+   //  so I added parentheses to eliminate a compiler warning and reformatted
+   if ((evalSubTree && subTree->getReferenceCount() > 1)
+       || (subTree->getRegister() != NULL)
+       || (self()->inUpcastingMode() && !subTree->cannotOverflow()))
       {
       if (_baseRegister != NULL)
          {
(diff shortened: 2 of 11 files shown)


mstoodle:
genie-omr build all

mstoodle:
@andrewcraik would you please help out by double-checking these changes, since most of them fall in the optimizer component

charliegracie:
@mstoodle did you use a tool or the output from the compiler to make these changes?

mstoodle:
@charliegracie I fixed each one individually manually.

charliegracie:
hmm ok.. I could trivially see that most of them were just adding the brackets to force exactly what the compiler was already doing but a few of them I was not so sure with a quick read. If you had of used a tool to put the brackets where the compiler was already enforcing the operator rules I would have just approved but I guess I need to spend some more time reading the code.

Thanks for reducing the compiler warnings! It makes the compiles easier to follow and also improves compilation speed!

mstoodle:
There are 2 that change functionality and that's because (I believe) they are bugs. I'll be doing more of this warning reduction in the compiler component because the extra messages annoy me :) .

@andrewcraik another ping for your review please.

compiler/optimizer/VirtualGuardHeadMerger.cpp
```diff
@@ -342,8 +342,8 @@ int32_t TR_VirtualGuardHeadMerger::perform() {
             // tail split
             if ((block->getNextBlock()->getSuccessors().size() == 1) ||
                 ((block->getNextBlock()->getSuccessors().size() == 2) &&
-                 block->getNextBlock()->getLastRealTreeTop()->getNode()->getOpCode().isBranch()) &&
-                performTransformation(comp(), "%sCloning block_%d and placing clone after block_%d to reduce HCR guard nops\n", OPT_DETAILS, block->getNextBlock()->getNumber(), cold1->getNumber()))
+                 block->getNextBlock()->getLastRealTreeTop()->getNode()->getOpCode().isBranch() &&
```
andrewcraik: I'm going a little cross-eyed trying to read this, but I think the intention is for the performTransformation to be applied to both size == 1 and size == 2 which is not what the change does - it just bind the perform to the size == 2 case

compiler/x/codegen/OMRMemoryReference.cpp
```diff
@@ -524,7 +524,7 @@ OMR::X86::MemoryReference::populateMemoryReference(
       evalSubTree = false;
 
    if (evalSubTree &&
-       subTree->getReferenceCount() > 1 || subTree->getRegister() != NULL || (self()->inUpcastingMode() && !subTree->cannotOverflow()))
+       (subTree->getReferenceCount() > 1 || subTree->getRegister() != NULL || (self()->inUpcastingMode() && !subTree->cannotOverflow())))
```
andrewcraik: hmm I'm not sure if this is right - the intent might have been to say if the refcount is greater than 1 and we want to eval OR if we have a register OR other condition since you won't eval in the other two cases... @0dvictor thoughts?

0dvictor: I think the logic is correct, though it might not be quite obvious. It lists three conditions:
1. The reference count is greater than 1;
2. The tree has been evaluated;
3. CG is in upcasting mode (in the processing of evaluating `i2l`, `s2l`, `s2i`), AND the tree may overflow.

Obviously, we should evaluate the tree under case 1; otherwise we may incorrectly evaluate multiple times for the subsequent occurrences.
For case 2, since the tree has been evaluated, "re-evaluate" won't generate any instructions but simply returns its result register. The use of the register is same as if we freshly evaluate the tree.
For case 3, we need insert the proper sign-extension instruction so that we have to evaluate the tree regardless of its reference count.
So the condition looks correct to me.

compiler/optimizer/VirtualGuardHeadMerger.cpp
mstoodle: yeah...looking at it again, I think you're right. I'll fix it up.

compiler/x/codegen/OMRMemoryReference.cpp
andrewcraik: The issue I have is that evalSubTree is only true for some specific loadaddr so I'm trying to figure out why that is and why it should feed through as it is doing. I'm not saying the analysis above is wrong, but I'm not sure if there is an unintended change here (see 521 to 524 above).

0dvictor: I see your point, and I agree three may be an unintended change. The original code is equivalent to
```
if (   (evalSubTree && 
        subTree->getReferenceCount() > 1)       || subTree->getRegister() != NULL || (self()->inUpcastingMode() && !subTree->cannotOverflow()))
```
`&&` has higher precedence than `||` so that `evalSubTree && subTree->getReferenceCount() > 1` is evaluated first.
Maybe we should move `subTree->getReferenceCount() > 1` into the earlier condition?

mstoodle: I have fixed this one as I agree the earlier association of && makes more sense, especially given the archeology @andrewcraik went through to track back how this condition has evolved over the years (though we didn't go *all* the way back :) ). I also added a comment to capture that we're preserving the original code's behaviour and overriding the indentation that suggested a different intent.

mstoodle:
@andrewcraik if you give me the thumbs up on the two additional changes, I'll squash and then we can run more comprehensive testing....

mstoodle:
I'll also update the commit message and title since I'm not sure I actually found or fixed any bugs (other than my own).

mstoodle:
genie-omr build all

Leonardo2718:
@andrewcraik if you can give a :+1: I will merge this.

PR #3300, opened 2018-12-07
AArch64: Implement integer extension evaluators: i2l, etc.

knn-k:
This commit implements i2lEvaluator() and other integer signed/unsigned
extension evaluators.


knn-k:
AppVeyor build failed with the following message:

>[NuGet] An error occurred while loading packages from 'https://chocolatey.org/api/v2/': The remote server returned an error: (403) Forbidden.
> winflexbison not installed. The package was not found with the source(s) listed.

knn-k:
Implemented `l2iEvaluator()` (also handles `i2b`, `l2s`, etc.).

Further performance improvement is possible in the same way as other platforms do:
If the first child of the node is a load from memory, the reference count of the child is 1, and the child has no register assigned yet, instructions like `SXTB` and `UXTH` can be omitted.  You can extend (signed or unsigned) the data size when you load the value from memory into register.

knn-k:
Tril tests are added in #3312.

knn-k:
Updated files resolving merge conflicts.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3303, opened 2018-12-07
Make x86-64 table jump position independent in AOT

jdmpapin:
The `TR_MethodAddress` relocation which has until now been generated for the displacement of the memory reference of the jump dispatching a `table` node is incorrect on 64-bit.

First, the displacement is only 32 bits, but the relocation works on a pointer-sized value (64 bits). The high half of the "address" is really the first four bytes of the next instruction. If the address adjustment changes the high half, then the next instruction will be modified.

Second, the 32-bit displacement gets sign-extended. If the code were originally compiled at a location below 2GB, and later loaded above 2GB (but below 4GB), the address adjustment will set the sign bit, so that the sign-extension is no longer the same as the zero-extension (i.e. the correct address).

To prevent these problems, the table evaluator now generates a position-independent sequence for the jump table dispatch, which requires no (external) relocation:

    lea rJumpTable, [rip + OFFSET_TO_JUMP_TABLE]
    jmp qword ptr [rJumpTable + 8*rIndex]

This sequence is only generated for 64-bit relocatable code. 32-bit code doesn't have access to `rip`-relative addressing, but it's immune to the above relocation problems anyway, since the pointer size is 32 bits, and there is no sign- nor zero-extension. For non-relocatable code, the existing lone `jmp` is preferable to a two-instruction sequence.

The addresses contained in the jump table remain position-dependent, and they continue to get relocated just as before.

### DIFF
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -462,12 +462,33 @@ TR::Register *OMR::X86::TreeEvaluator::tableEvaluator(TR::Node *node, TR::CodeGe
          cg->evaluate(secondChild->getFirstChild()); // evaluate the glRegDeps
       }
 
-   TR::MemoryReference *tempMR = generateX86MemoryReference((TR::Register *)NULL,
-                                                            selectorReg,
-                                                            (uint8_t)(TR::Compiler->target.is64Bit()? 3 : 2),
-                                                            (intptrj_t)branchTable, cg);
+   TR::MemoryReference *jumpMR = NULL;
+   TR::Register *branchTableReg = NULL;
+   if (TR::Compiler->target.is64Bit() && cg->comp()->compileRelocatableCode())
+      {
+      // Generate position-independent code so that no (external) relocation is
+      // necessary:
+      //
+      //    lea rBranchTable, [rip + OFFSET_TO_JUMP_TABLE]
+      //    jmp qword ptr [rBranchTable + 8*rIndex]
+      //
+      TR::LabelSymbol *label = generateLabelSymbol(cg);
+      label->setCodeLocation(reinterpret_cast<uint8_t*>(branchTable));
+      TR::MemoryReference *branchTableLeaMR = generateX86MemoryReference(label, cg);
+      branchTableReg = cg->allocateRegister();
+      generateRegMemInstruction(LEA8RegMem, node, branchTableReg, branchTableLeaMR, cg);
+      jumpMR = generateX86MemoryReference(branchTableReg, selectorReg, 3, cg);
+      }
+   else
+      {
+      jumpMR = generateX86MemoryReference(
+         (TR::Register *)NULL,
+         selectorReg,
+         (uint8_t)(TR::Compiler->target.is64Bit()? 3 : 2),
+         (intptrj_t)branchTable, cg);
 
-   tempMR->setNeedsCodeAbsoluteExternalRelocation();
+      jumpMR->setNeedsCodeAbsoluteExternalRelocation();
+      }
 
    TR::X86MemTableInstruction *jmpTableInstruction = NULL;
 
@@ -481,11 +502,11 @@ TR::Register *OMR::X86::TreeEvaluator::tableEvaluator(TR::Node *node, TR::CodeGe
          deps->stopAddingConditions();
          }
 
-      jmpTableInstruction = generateMemTableInstruction(JMPMem, node, tempMR, numBranchTableEntries, deps, cg);
+      jmpTableInstruction = generateMemTableInstruction(JMPMem, node, jumpMR, numBranchTableEntries, deps, cg);
       }
    else
       {
-      generateMemInstruction(JMPMem, node, tempMR, cg);
+      generateMemInstruction(JMPMem, node, jumpMR, cg);
       }
 
    for (i = 2; i < node->getNumChildren(); ++i)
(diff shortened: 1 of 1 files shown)


Leonardo2718:

jdmpapin:
AppVeyor didn't start due to 403 Forbidden for winflexbison

jdmpapin:
> 32-bit code doesn't have access to rip-relative addressing, but it's immune to the above relocation problems anyway, since the pointer size is 32 bits, and there is no sign- nor zero-extension. For non-relocatable code, the existing lone jmp is preferable to a two-instruction sequence.

0dvictor:
> 32-bit code doesn't have access to rip-relative addressing, but it's immune to the above relocation problems anyway, since the pointer size is 32 bits, and there is no sign- nor zero-extension.

Right; however, following sequences would achieve the same:
```
CALL 0
POP jump_table_reg
LEA jump_table_reg, [jump_table_reg + OFFSET_TO_JUMP_TABLE]
```

> For non-relocatable code, the existing lone jmp is preferable to a two-instruction sequence.

I doubt the extra LEA instruction and short-lived register has measurable perf impact. IMO, the inconsistency between 64-bit AOT and the rest code path would make debugging and maintaining related code harder.

jdmpapin:
> following sequences would achieve the same: *[...]*

I guess we already expand label memory references to include the `call`, `pop` sequence on 32-bit? Otherwise, this change would still need to distinguish 32- from 64-bit in order to determine whether to emit `call`, `pop`.

> I doubt the extra LEA instruction and short-lived register has measurable perf impact.

Yes, the difference should be very small. @andrewcraik what do you think?

andrewcraik:
So I like the cleanliness of the RIP relative model - my concern is around the performance cost of the `lea` or `call; pop 0; lea`. I think unifying the AOT cases so the relocations all work the same should be the goal for now. Can we add an issue to track trying to unify the normal and AOT cases to the RIP relative addressing - we just need to make sure we only do that after careful perf testing rather than trying to clean the code up too much prematurely.

jdmpapin:
> I guess we already expand label memory references to include the `call`, `pop` sequence on 32-bit?

We don't already have this, and I don't particularly want to add it. At the moment `_addressRegister` doesn't exist on 32-bit, so it would have to move. We would need to determine the right condition under which to force position-independent code. At the moment it *looks* like I'm the only (x86) user of label memory references, so it could be `_label != NULL`, but there might be other users outside of OpenJ9. And finally, when instructions expand themselves into multiple-instruction sequences during binary encoding, it can be difficult to follow what's going on.

I could test for 32-bit and generate `call`, `pop` here, which would "unify" the AOT cases in the sense that they use corresponding relocations (none, regardless of bitness), but doing so would add _more_ cases to this particular code. Furthermore, this strategy has its own complications around fixing up the relative offset. We could ensure a perfectly predictable length for the instructions between `call` (exclusive) and the use of the result (inclusive) so that the offset from the point of use can be adjusted. This would mean preventing encodings from varying, and preventing anything (e.g. RA) from inserting instructions in this range. Alternately, we could keep a pointer to the `pop` instruction in order to compute the correct offset directly. Either way, the memory reference containing the offset needs to keep additional state.

With all this in mind, do we still want to try to unify the AOT paths?

andrewcraik:
After much discussion of all the tradeoffs I think that the current form is the best tradeoff - making bout 32 and 64 use a relative addressing scheme for the jump table location would add a second new code path, unification of AOT and non-AOT is likely to cause performance problems, varying relocations because of instructions sequence are probably also inevitable. As a result, this is the cleanest of the solutions and I'd like to proceed with this. Over to @vijaysun-omr - I believe the fixed 64bit relocations are required for OpenJ9's OSX effort so if you could have a look sooner rather than later it would be appreciated.

0dvictor:
Agreed, after discussions, I also think this approach looks the best tradeoff in terms of performance among all different approaches that I can think of. This pull request looks good to me.

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -462,12 +462,33 @@ TR::Register *OMR::X86::TreeEvaluator::tableEvaluator(TR::Node *node, TR::CodeGe
...
+   if (TR::Compiler->target.is64Bit() && cg->comp()->compileRelocatableCode())
+      {
+      // Generate position-independent code so that no (external) relocation is
+      // necessary:
+      //
+      //    lea rJumpTable, [rip + OFFSET_TO_JUMP_TABLE]
+      //    jmp qword ptr [rJumpTable + 8*rIndex]
+      //
+      TR::LabelSymbol *label = generateLabelSymbol(cg);
+      label->setCodeLocation(reinterpret_cast<uint8_t*>(branchTable));
```
0xdaryl: Minor pedantic comment.  You seem to now be referring to the previously named "branchTable" as a "jumpTable".  I'm fine with that, but can we keep the naming consistent in this function?  Simply renaming `branchTable` to `jumpTable` is probably easiest as there are only a couple of references in this function.

jdmpapin: Renamed using "branch" instead of "jump" to stay consistent with not just `branchTable` but also `numBranchTableEntries` and `addMetaDataForBranchTableAddress()`

0xdaryl:
@genie-omr build xlinux,win

0xdaryl:
This is an x86-specific change.  A full genie build is not required.

PR #3304, opened 2018-12-07
Add basic type_traits-like utility to OMR

rwy0717:
Type traits are useful in templated code to make queries or assertions
about type parameters. For example, a container might assert that it
never contains a reference type.  A smart pointer won't define
pointer-operators when it's referent is void.

Our supported compilers don't provide the standard C++11 header
<type_traits> in OMR. Even looking at planned compiler upgrades, it's
unlikely we'll ever get a standard implementation of type_traits on XLC.

This header provides some easily implemented type traits without any
hacks or compiler support. As well, this header follows OMR coding
conventions, rather than try to simulate the standard type_traits
header.

This PR implements the following:

* Static constants:
  * `IntegralConstant<T, X>`
  * `BoolConstant<B>` / `TrueConstant` / `FalseConstant`
  * `TypeAlias<T>`
* types queries:
  * `IsSame<T, U>`
  * `IsReference<T>`
  * `IsVoid<T>`
  * `IsPointer<T>`
* type modifications:
  * `RemoveConst<T>`
  * `RemoveVolatile<T>`
  * `RemoveCv<T>`
  * `RemoveReference<T>`
  * `RemoveCvRef<T>`

There's lots of opportunities for expanding the traits provided in this
initial commit, but this is a good base for defining container types in
OMR.

### DIFF
--- a/fvtest/coretest/TestTypeTraits.cpp
+++ b/fvtest/coretest/TestTypeTraits.cpp
@@ -1,0 +1,116 @@
+#include <OMR/TypeTraits.hpp>
+#include <gtest/gtest.h>
+
+namespace OMR
+{
+
+TEST(TestTypeTraits, IntegralConst)
+{
+	EXPECT_EQ(0, (IntegralConstant<int, 0>::VALUE));
+	EXPECT_EQ(0xdeadbeef, (IntegralConstant<unsigned int, 0xdeadbeef>::VALUE));
+}
+
+TEST(TestTypeTraits, BooleanConstant)
+{
+	EXPECT_TRUE((BoolConstant<true>::VALUE));
+	EXPECT_TRUE(TrueConstant::VALUE);
+	EXPECT_FALSE((BoolConstant<false>::VALUE));
+	EXPECT_FALSE(FalseConstant::VALUE);
+}
+
+TEST(TestTypeTraits, IsSame)
+{
+	EXPECT_TRUE((IsSame<int, int>::VALUE));
+	EXPECT_FALSE((IsSame<int, double>::VALUE));
+	EXPECT_TRUE((IsSame<void*, void*>::VALUE));
+	EXPECT_FALSE((IsSame<void* const, void*>::VALUE));
+	EXPECT_FALSE((IsSame<void*, const void*>::VALUE));
+}
+
+TEST(TestTypeTraits, TypeAlias)
+{
+	EXPECT_TRUE((IsSame<void, TypeAlias<void>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<void*, TypeAlias<void*>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<const int, TypeAlias<const int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int&, TypeAlias<int&>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveConst)
+{
+	EXPECT_TRUE((IsSame<int, RemoveConst<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveConst<const int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<volatile int, RemoveConst<const volatile int>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveVolatile)
+{
+	EXPECT_TRUE((IsSame<int, RemoveVolatile<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveVolatile<volatile int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<const int, RemoveVolatile<const volatile int>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveCv)
+{
+	EXPECT_TRUE((IsSame<int, RemoveCv<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCv<const int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCv<const volatile int>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveReference)
+{
+	EXPECT_TRUE((IsSame<int, RemoveReference<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveReference<int&>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<const int, RemoveReference<const int&>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int*, RemoveReference<int*&>::Type>::VALUE));
+	// TODO: EXPECT_TRUE((IsSame<int, RemoveReference<int&&>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveCvRef)
+{
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<const int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<const volatile int>::Type>::VALUE));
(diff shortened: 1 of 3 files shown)


fvtest/coretest/TestTypeTraits.cpp
```diff
@@ -0,0 +1,121 @@
...
+{
+	EXPECT_TRUE((IsSame<int, RemoveCv<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCv<const int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCv<const volatile int>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, RemoveReference)
+{
+	EXPECT_TRUE((IsSame<int, RemoveReference<int>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveReference<int&>::Type>::VALUE));
```
Leonardo2718: It might be worth to also add:

```c++
	EXPECT_TRUE((IsSame<const int, RemoveReference<const int&>::Type>::VALUE)); // or volatile
	EXPECT_TRUE((IsSame<int*, RemoveReference<int*&>::Type>::VALUE));
```

fvtest/coretest/TestTypeTraits.cpp
```diff
@@ -0,0 +1,121 @@
...
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<int&>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<const int&>::Type>::VALUE));
+	EXPECT_TRUE((IsSame<int, RemoveCvRef<const volatile int&>::Type>::VALUE));
+	// TODO: EXPECT_TRUE((IsSame<int, RemoveCvRef<int&&>::Type>::VALUE));
+}
+
+TEST(TestTypeTraits, IsReference)
+{
+	EXPECT_FALSE((IsReference<int>::VALUE));
+	EXPECT_TRUE((IsReference<int&>::VALUE));
```
Leonardo2718: Maybe also add

```c++
EXPECT_FALSE((IsReference<int *>::VALUE));
EXPECT_TRUE((IsReference<const int &>::VALUE));
```

include_core/OMR/TypeTraits.hpp
```diff
@@ -0,0 +1,161 @@
...
+/// Helper for IsPointer.
+template <typename T>
+struct IsPointerBase : FalseConstant {};
+
+template <typename T>
+struct IsPointerBase<T*> : TrueConstant {};
+
+/// IsPointer<T>::VALUE is true if T is a (possibly cv qualified) primitive pointer type.
+template <typename T>
+struct IsPointer : IsPointerBase<typename RemoveCvRef<T>::Type> {};
```
Leonardo2718: Interesting. This actually deviates from `std::is_pointer`, which only remove cv qualifiers not ref.

rwy0717: Hah good eye. This was deliberate, but after some offline talk today, I'll probably change this to match the standard type traits.

fvtest/coretest/TestTypeTraits.cpp
rwy0717: Sure! Will do.

include_core/OMR/TypeTraits.hpp
rwy0717: I stopped treating `T*&` as pointers, and updated the tests to reflect that.

fvtest/coretest/TestTypeTraits.cpp
rwy0717: added.

fvtest/coretest/TestTypeTraits.cpp
rwy0717: added.

include_core/OMR/TypeTraits.hpp
```diff
@@ -0,0 +1,161 @@
...
+template <typename T>
+struct RemoveReference<T&> : RemoveReference<T> {};
+
+// TODO: Handle RValue references in typetraits
+// template <typename T>
+// struct RemoveReference<T&&> : RemoveReference<T> {};
+
+/// Remove cv-qualifiers and one layer of references.
+template <typename T>
+struct RemoveCvRef : TypeAlias<T> {};
```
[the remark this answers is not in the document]
rwy0717: Updated.

include_core/OMR/TypeTraits.hpp
```diff
@@ -0,0 +1,161 @@
...
+/// RemoveVolatile<T>::Type is T without volatile qualifiers.
+template <typename T>
+struct RemoveVolatile : TypeAlias<T> {};
+
+template <typename T>
+struct RemoveVolatile<volatile T> : RemoveVolatile<T> {};
+
+/// RemoveCv<T>::Type is T without const/volatile qualifiers.
+template <typename T>
+struct RemoveCv : TypeAlias<T> {};
```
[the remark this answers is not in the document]
rwy0717: Updated.

rwy0717:
Re tril/enable_if: that would be great!

charliegracie:
I thinking keeping it in the OMR format is the best answer.  If it is written to be a shim for `std` then it will set expectations that it works exactly like the `std` version.

charliegracie:
@Leonardo2718 are you good with the changes now?

include_core/OMR/TypeTraits.hpp
```diff
@@ -0,0 +1,140 @@
...
+/// RemoveCv<T>::Type is T without const/volatile qualifiers.
+template <typename T>
+struct RemoveCv : RemoveVolatile<typename RemoveConst<T>::Type> {};
+
+/// RemoveReference<T>::Type is T as a non-reference.
+template <typename T>
+struct RemoveReference : TypeAlias<T> {};
+
+template <typename T>
+struct RemoveReference<T&> : RemoveReference<T> {};
```
Leonardo2718: Is there a reason for defining the template recursively, instead of:

```suggestion
struct RemoveReference<T&> : TypeAlias<T> {};
```

Unless there's something I'm missing, the two should be equivalent because reference collapsing prevents having a reference to a reference.

include_core/OMR/TypeTraits.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+// TODO: Handle RValue references in typetraits
```
Leonardo2718: So, I don't know how other's feel about TODO comments, but I generally prefer GitHub issues because they are easier to track (I know we're actively trying to eliminate TODOs from the compiler component).

charliegracie: To me it depends on the TODO and how long I think it will live in the code base.  When people have the code checked out (maybe even from another project like OpenJ9) knowing that there are GitHub Issues associated with the code you are looking at is very difficult. The TODO in the code allows users to know there is outstanding work. I personally do not like the idea of having hundreds if not thousands of GitHub Issues that may sit forgotten for years until we decide to close them when a simple TODO would suffice. I think committers can use their best judgement to ask for an Issue or leave the TODO.

include_core/OMR/TypeTraits.hpp
rwy0717: It was just a stylistic choice. In my initial PR, most types were defined recursively. It makes less sense now.

include_core/OMR/TypeTraits.hpp
Leonardo2718: I see your point, although I think TODO's are just as likely to be forgotten and deleted :grin: .

In this case, I think I'm ok with the TODOs because it's very clear what the task to be done is. At most, I expect any discussion to be about what support/use for rvalue references OMR has. We can cross that bridge when we get to it.

Leonardo2718:
@genie-omr build all

rwy0717:
The OSX CI completed successfully a few hours ago, but it looks like the status here didn't update properly.

Leonardo2718:
@genie-omr build osx (?)

Leonardo2718:
@genie-omr build osx

Leonardo2718:
Restarting the build fixed it. Merging.

PR #3305, opened 2018-12-07
Fix invalid option definitions causing crashes in the JIT

fjeremic:
The `TR::OptionTable::msg` field must be non-null, otherwise when
printing the options via the `verbose` command we run into null pointer
dereferences which causes the JIT to crash.

Issue: eclipse/openj9#3955


fjeremic:
@dsouzai could I please get a review?

vijaysun-omr:
@genie-omr build all

PR #3307, opened 2018-12-07
Provide Dockerfile for easier x86 build failure reproduction

mstoodle:
While working on a recent PR on my MacBook, I needed to reproduce
a failure in the Linux x86 testing. To do that, I built a Docker
container to build and run tests and to diagnose the failure.

It seemed like something others might find useful, so I cleaned
it up a bit to make it a bit more consumable. It's not perfect
but its gets you to having a workin Linux x86 build with only
a couple of commands.  The basic directions are provided in
the Dockerfile, which I'll reproduce here to save a bit of time:

Basic usage (adapt to your environment and scenario):
  git clone https://github.com/eclipse/omr (or cd to a directory that contains an OMR repository)
  cd omr/buildenv/docker && docker build -f x86_64/ubuntu16/Dockerfile -t buildomr .
  cd ../../../
  docker run --privileged -v $PWD/omr:/omr -it buildomr

The container will automatically run cmake to configure itself and
then initiate the build.  Once the build completes, you'll be left
sitting at a prompt in the container so you can run more commands.
I had the container also install gdb, vim, and emacs but we could
add some other useful commands too.

### DIFF
--- a/buildenv/docker/x86_64/ubuntu16/Dockerfile
+++ b/buildenv/docker/x86_64/ubuntu16/Dockerfile
@@ -1,0 +1,61 @@
+# Copyright (c) 2018, 2018 IBM Corp. and others
+#
+# This program and the accompanying materials are made available under
+# the terms of the Eclipse Public License 2.0 which accompanies this
+# distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+# or the Apache License, Version 2.0 which accompanies this distribution and
+# is available at https://www.apache.org/licenses/LICENSE-2.0.
+# 
+# This Source Code may also be made available under the following
+# Secondary Licenses when the conditions for such availability set
+# forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+# General Public License, version 2 with the GNU Classpath
+# Exception [1] and GNU General Public License, version 2 with the
+# OpenJDK Assembly Exception [2].
+# 
+# [1] https://www.gnu.org/software/classpath/license.html
+# [2] http://openjdk.java.net/legal/assembly-exception.html
+# 
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+#
+#
+# Create OMR build environment with Ubuntu 16.04
+# Basic usage (adapt to your environment and scenario):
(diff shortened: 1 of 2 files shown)


mstoodle:
I mimicked the OpenJ9 directory structure, but this file is different than the ones OpenJ9 puts into their `buildenv` directory so maybe that's not the right approach.

I welcome feedback and look forward to making this Dockerfile even better!

mstoodle:
Note that this commit doesn't affect any testing, so the appveyor failure is spurious (and due to missing winflexbison on the test machine...)

mstoodle:
@charliegracie any thoughts on this  convenience mechanism?

charliegracie:
This seems like a good change to help users get up and going quickly. I am not sure having these directly in the main repo is the right place but I do not have a better suggestion so we should start here :)

buildenv/docker/x86_64/ubuntu16/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# 
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+#
+#
+# Create OMR build environment with Ubuntu 16.04
+# Basic usage (adapt to your environment and scenario):
+#   git clone https://github.com/eclipse/omr (or cd to a directory that contains an OMR repository)
+#   cd omr/buildenv/docker && docker build -f x86_64/ubuntu16/Dockerfile -t buildomr .
+#   cd ../../../
+#   docker run --privileged -v $PWD/omr:/omr -it buildomr
```
charliegracie: Is the --privileged specified to allow mounting of the disk?

mstoodle: It enables you to use gdb inside the container

mstoodle:
I put it in the repo so it would be version controlled and so it's available as soon as you clone the repo :) .

mstoodle:
Still not sure about the directory I put it in, though...

buildenv/docker/x86_64/ubuntu16/Dockerfile
charliegracie: Ahh I knew it must have had a reason :)

charliegracie:
Files can always be moved :).  I am happy to merge as-is if you think it is ready

charliegracie:
I believe the change is fine.  If we want to move it we can open another PR in the future.

PR #3312, opened 2018-12-10
Tril tests for b2i, s2l, etc.

knn-k:
This commit adds tril test cases for integer extension evaluators
such as b2i and s2l.

### DIFF
--- a/fvtest/compilertriltest/CMakeLists.txt
+++ b/fvtest/compilertriltest/CMakeLists.txt
@@ -46,6 +46,7 @@ add_executable(comptest
 	BitPermuteTest.cpp
 	MaxMinTest.cpp
 	CompareTest.cpp
+	TypeConversionTest.cpp
 )
 
 target_link_libraries(comptest
--- a/fvtest/compilertriltest/OpCodeTest.hpp
+++ b/fvtest/compilertriltest/OpCodeTest.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2018 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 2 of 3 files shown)


knn-k:
I had to add a new template class `TRTest::UnaryOpWithRetTypeTest` in OpCodeTest.hpp, because using `TRTest::OpCodeTest<int32_t, int8_t>` results in compilation failure with `TRTest::to_struct(GetParam())`.

knn-k:
How can I represent unsigned types like `uint8_t` in Tril, for writing tests of `bu2i`, etc.?

Leonardo2718:
Re `bu2i` et al.: Testarossa types don't have "signedness" (usually), operations do. The types are really just used to specify bit-width. So, it's completely valid to compile, say

```
(method return=Int32 args=[Int8] (block (ireturn (bu2i (bload parm=0)))))
```

and to retrieve the entry point using

```c++
compiler.getEntryPoint<uint32_t (*)(uint8_t)>()
```

Does that help? :slightly_smiling_face:

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -167,3 +167,177 @@ INSTANTIATE_TEST_CASE_P(DivArithmeticTest, Int32Arithmetic, ::testing::Combine(
...
+    ::testing::ValuesIn(TRTest::const_values<int16_t>()),
+    ::testing::Values(
+        std::make_tuple("s2l", s2l) )));
+
+class Int32ToInt64 : public TRTest::UnaryOpWithRetTypeTest<int64_t,int32_t> {};
+
+TEST_P(Int32ToInt64, UsingLoadParam) {
+    auto param = TRTest::to_struct(GetParam());
+
+    char *inputTrees = "(method return=Int64 args=[Int32] (block (lreturn (i2l (iload parm=0)))))";
```
fjeremic: Could we format these in a "tree-like" form as you'd find in trace files? See [1] for example. It makes it much easier to read and it feels natural as if you were reading a trace log.

[1] https://github.com/eclipse/omr/blob/8a555b1f455096e71b9155989444ef9550da21a3/fvtest/compilertriltest/IfxcmpgeReductionTest.cpp#L97-L114

knn-k:
Updated files reflecting review comments and adding tests for `bu2i`, etc.

knn-k:
Renamed `ArithmeticTest` in `INSTANTIATE_TEST_CASE_P()` to `TypeConversionTest`.

knn-k:
Added `UsingConst` variants.

Leonardo2718:
@genie-omr build all

PR #3314, opened 2018-12-10
Add isPow2, aligned, and align utility to Bytes.hpp

rwy0717:
The align utility is for quickly rounding sizes up to a power of two.
Aligning is done everywhere, so it makes sense to have one common
implementation. Note that this is not a general rounding utility, as it
can only round certain specially crafted inputs.

### DIFF
--- a/fvtest/coretest/TestBytes.cpp
+++ b/fvtest/coretest/TestBytes.cpp
@@ -48,4 +48,140 @@ TEST(TestBytes, CompareOneUnitToSmallerUnit)
 	EXPECT_EQ(mebibytes(1024), gibibytes(1));
 }
 
+TEST(TestBytes, IsPow2)
+{
+	EXPECT_TRUE(isPow2(1));
+	EXPECT_TRUE(isPow2(2));
+	EXPECT_TRUE(isPow2(4));
+	EXPECT_TRUE(isPow2(8));
+	EXPECT_TRUE(isPow2(size_t(1) << ((sizeof(size_t)*8) - 1)));
+
+	EXPECT_FALSE(isPow2(0));
+	EXPECT_FALSE(isPow2(3));
+	EXPECT_FALSE(isPow2(5));
+	EXPECT_FALSE(isPow2(6));
+	EXPECT_FALSE(isPow2(7));
+	EXPECT_FALSE(isPow2(9));
+	EXPECT_FALSE(isPow2(SIZE_MAX));
+}
+
+TEST(TestBytes, AlignedUnsafe)
+{
+	EXPECT_TRUE(alignedNoCheck(0, 0));
+	EXPECT_FALSE(alignedNoCheck(1, 0));
+	EXPECT_FALSE(alignedNoCheck(2, 0));
+	EXPECT_FALSE(alignedNoCheck(gibibytes(1), 0));
+}
+
+TEST(TestBytes, Aligned)
+{
+	EXPECT_TRUE(aligned(1, 1));
+	EXPECT_TRUE(aligned(2, 1));
+
+	EXPECT_FALSE(aligned(1, 2));
+	EXPECT_TRUE(aligned(2, 2));
+	EXPECT_FALSE(aligned(3, 2));
+}
+
+TEST(TestBytes, AlignToZero)
+{
+	// Zero is considered an invalid input, since the calculation always results in zero.
+	// This test is validating that behaviour, but users should NOT be aligning to zero.
+	// the minimum valid alignment is 1.
+	EXPECT_EQ(0, alignNoCheck(0, 0));
+	EXPECT_EQ(0, alignNoCheck(1, 0));
+	EXPECT_EQ(0, alignNoCheck(2, 0));
+	EXPECT_EQ(0, alignNoCheck(3, 0));
+	EXPECT_EQ(0, alignNoCheck(gibibytes(1), 0));
+}
+
+TEST(TestBytes, AlignToOne)
+{
+	EXPECT_EQ(0, align(0, 1));
+	EXPECT_EQ(1, align(1, 1));
+	EXPECT_EQ(2, align(2, 1));
+	EXPECT_EQ(3, align(3, 1));
+	EXPECT_EQ(gibibytes(1), align(gibibytes(1), 1));
+}
+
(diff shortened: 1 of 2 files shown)


include_core/OMR/Bytes.hpp
```diff
@@ -53,6 +53,25 @@ gibibytes(size_t n)
...
+
+/// Round a size up to a multiple of alignment.
+/// Alignment must be a positive multiple of two, EG 1, 2, 4, 8, or 16.
+/// The maximum unaligned size is intrinsically related to the alignment.
+/// As a conservative measure, users should not use aligned sizes greater than
+/// UNALIGNED_MAX_SIZE, nor use alignments larger than ALIGNMENT_MAX.
+inline size_t
+align(size_t size, size_t alignment)
+{
+	return (size + (alignment - 1)) & ~(alignment - 1);
```
fjeremic: Given the "shoot yourself in the foot" likelyhood of someone using this for non-powers of two, can we add an assert in here to check that `alignment` is a power of two rather than relying on the user reading the documentation. We can use something like [1] to check for it.

[1] https://graphics.stanford.edu/~seander/bithacks.html#DetermineIfPowerOf2

0xdaryl: +1.   That query alone should be a utility in this core library somewhere.  The power-of-2 question is already asked in the JIT, so we can eventually replace those questions with whatever becomes available in core.

charliegracie: If this is going to be used by all components, including the GC, which currently has the exact code above I would rather keep the code to a minimum as to not impact the performance in tights loops in the GC that calls this code.  There is tons of complicated code that requires users to read the API documentation before they call it. If preferred we can rename the API to something that implies a power of 2 alignment but I would rather not slow down the code for no reason.

include_core/OMR/Bytes.hpp
```diff
@@ -53,6 +53,25 @@ gibibytes(size_t n)
...
+#if defined(OMR_ENV_DATA64)
+static const size_t ALIGNMENT_MAX = size_t(1) << 63;
+static const size_t UNALIGNED_SIZE_MAX = size_t(1) << 63;
+#else
+static const size_t ALIGNMENT_MAX = size_t(1) << 31;
+static const size_t UNALIGNED_SIZE_MAX = size_t(1) << 31;
+#endif
+
+/// Round a size up to a multiple of alignment.
+/// Alignment must be a positive multiple of two, EG 1, 2, 4, 8, or 16.
```
jdmpapin: Should this say "*power* of two"?

include_core/OMR/Bytes.hpp
rwy0717: Yeah, my intention was for this to be a "quick and dirty" call. Especially when you consider the use cases--fast path allocator for example--it makes sense to keep it small. 

I could split this into a safe `align` and an `alignUnsafe` helper. I can also add `isAligned(n, a)` and `isPow2(x)` to the header, while I'm here.

There aren't any assertion utilities in OMR core (yet, AFAIK), so any checks would have to be done with the standard C assert macro. cassert is disabled in release builds, which I think is the behaviour we want, although these kinds of assertions tend to mislead people when debugging.

rwy0717: Do we want to accept an alignment of zero, or not?

include_core/OMR/Bytes.hpp
rwy0717: Ah haha for sure. Will fix.

rwy0717:
The planned functions:

* `isPow2(x)` and `isPow2OrZero(x)`
  * determine if x is a power of two.
  * OrZero variant does not check for zero.
* `aligned(n, a)` and `alignedUnsafe(n, a)`
  * determine if `n` is aligned to `a`.
  * The unsafe variant doesn't validate arguments.
* `align(n, a)` and `alignUnsafe(n, a)`
  * round `n` up to `a`.

rwy0717:
The value of `isPow2OrZero` is dubious... unless someone has a specific use case, I won't add it.

include_core/OMR/Bytes.hpp
rwy0717: Fixed, thanks.

rwy0717:
Updated and ready for another look.

I'm concerned that the range checking is too restrictive. Currently it allows alignments or sizes up to `(SIZE_MAX / 2) + 1`. Are we happy with these limits? Or I could just drop the range check all together, if that's what people want.

rwy0717:
I can always do a more flexible overflow check, and nix the max constants.

include_core/OMR/Bytes.hpp
charliegracie: Yes do not accept 0

rwy0717:
changes:
* renamed `alignUnsafe`/`alignedUnsafe` to `alignNoCheck`/`alignedNoCheck`, which is a more conventional naming scheme.
* replaced `size` range check with more permissive overflow check.
* removed` alignment` range check, any `size_t` larger than `ALIGNMENT_MAX` will fail the `isPow2` check anyways.

I left the `UNALIGNED_SIZE_MAX` and `ALIGNMENT_MAX` constants in the header, in case anyone finds them useful.

rwy0717:
Updated the comments. Ready for another look.

charliegracie:
@genie-omr build all

fvtest/coretest/TestBytes.cpp
```diff
@@ -48,4 +48,140 @@ TEST(TestBytes, CompareOneUnitToSmallerUnit)
...
+}
+
+TEST(TestBytes, AlignToEight)
+{
+	EXPECT_EQ(0, align(0, 8));
+	EXPECT_EQ(8, align(1, 8));
+	EXPECT_EQ(8, align(7, 8));
+	EXPECT_EQ(8, align(8, 8));
+	EXPECT_EQ(16, align(9, 8));
+	EXPECT_EQ(gibibytes(1), align(gibibytes(1), 4));
```
youngar:
```suggestion
	EXPECT_EQ(gibibytes(1), align(gibibytes(1), 8));
```

youngar:
@genie-omr build all

PR #3317, opened 2018-12-10
Add support for SIGWINCH

babsingh:
`SIGWINCH` is only available on Unix platforms. `SIGWINCH` is needed to
support Ammonite-REPL: https://ammonite.io/#Ammonite-REPL.

Fixes: https://github.com/eclipse/openj9/issues/3976


include_core/omrport.h
```diff
@@ -776,7 +776,11 @@ typedef struct J9ProcessorInfos {
 #define OMRPORT_SIG_FLAG_DOES_NOT_MAP_TO_POSIX  0x200000
 #define OMRPORT_SIG_FLAG_SIGHUP  0x400000
 #define OMRPORT_SIG_FLAG_SIGCONT  0x800000
-#define OMRPORT_SIG_FLAG_SIGALLASYNC  (OMRPORT_SIG_FLAG_SIGQUIT | OMRPORT_SIG_FLAG_SIGABRT | OMRPORT_SIG_FLAG_SIGTERM | OMRPORT_SIG_FLAG_SIGRECONFIG | OMRPORT_SIG_FLAG_SIGXFSZ | OMRPORT_SIG_FLAG_SIGINT | OMRPORT_SIG_FLAG_SIGHUP | OMRPORT_SIG_FLAG_SIGCONT)
+#define OMRPORT_SIG_FLAG_SIGWINCH  0x1000000
+#define OMRPORT_SIG_FLAG_SIGALLASYNC \
+			(OMRPORT_SIG_FLAG_SIGQUIT | OMRPORT_SIG_FLAG_SIGABRT | OMRPORT_SIG_FLAG_SIGTERM \
+				| OMRPORT_SIG_FLAG_SIGRECONFIG | OMRPORT_SIG_FLAG_SIGXFSZ | OMRPORT_SIG_FLAG_SIGINT \
```
DanHeidinga: The multiple levels of indentation here aren't my favourite but I won't object if @charliegracie is ok with this.

charliegracie: I would have preferred the `|` to be at the end of the line and then each new line lines up with the line above. ie...
```
#define OMRPORT_SIG_FLAG_SIGALLASYNC ( \
			OMRPORT_SIG_FLAG_SIGQUIT | OMRPORT_SIG_FLAG_SIGABRT | OMRPORT_SIG_FLAG_SIGTERM | \
			OMRPORT_SIG_FLAG_SIGRECONFIG | OMRPORT_SIG_FLAG_SIGXFSZ | OMRPORT_SIG_FLAG_SIGINT | \
			OMRPORT_SIG_FLAG_SIGHUP | OMRPORT_SIG_FLAG_SIGCONT | OMRPORT_SIG_FLAG_SIGWINCH)
```

port/unix/omrsignal.c
```diff
@@ -200,14 +200,14 @@ static struct {
 	{OMRPORT_SIG_FLAG_SIGXFSZ, SIGXFSZ},
 	{OMRPORT_SIG_FLAG_SIGINT, SIGINT},
 	{OMRPORT_SIG_FLAG_SIGHUP, SIGHUP},
-	{OMRPORT_SIG_FLAG_SIGCONT, SIGCONT}
+	{OMRPORT_SIG_FLAG_SIGCONT, SIGCONT},
+	{OMRPORT_SIG_FLAG_SIGWINCH, SIGWINCH}
```
charliegracie: While we are adding `SIGWINCH` are their other signals that we were not supporting that we should also support?

babsingh: We need to support all async signals seen under `kill -l` or [SunMiscSignalTest](https://github.ibm.com/runtimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/sun/misc/SunMiscSignalTest.java#L107-L135).

Currently, we use a 32-bit flag in the `omrsignal` implementation: [code](https://github.com/eclipse/omr/blob/master/port/unix/omrsignal.c#L112). Only ~7bits are remaining in the 32-bit flag: [code](https://github.com/eclipse/omr/blob/master/include_core/omrport.h#L749-L779). To support the remaining 15 signals, we need 15 more bits. The 32-bit flag will be converted into a 64-bit flag to support the remaining async signals.

I don't know how to prioritize unsupported async signals. Not sure how to use the remaining 7 bits in the 32-bit flag. Should I leave them unused for this PR? In the short term, we can use them as users request support for specific signals.

In the long term, the 32-bit flag will be converted into a 64-bit flag to support the remaining async signals.

charliegracie: ok I think this change is ok as-is so we can unlock users who need to handle this signal. I will approve and merge once the testing is complete but can you open an Issue that describe the work / requirement for switching this 32 bit flag to a 64 bit flag?

charliegracie:
@genie-omr build all

port/unix/omrsignal.c
babsingh: An issue already exists: https://github.com/eclipse/omr/issues/2332. 

> Each bit in J9*AsyncHandlerRecord->flags represents a signal. 32 bits can only represent 32 signals. On some platforms, there are more than 32 signals. If we use uint64_t to represent flags, then we will be able to represent 64 signals ...

PR #3318, opened 2018-12-11
Supporting options and code to allow AOT at cheap warm level

mpirvu:
This commit adds a new option -Xaot:disableAOTAtCheapWarm to control the
logic that will force AOT compilations to be performed at a "cheap"
warm level. A "cheap" warm level is an optimization level where the
inliner has been subdued somewhat (the same inliner changes are applied
to the Xtune:virtualized mode in OpenJ9 JVM).
Upgrades of AOT bodies through sampling are also subdued if the
guarded counting recompilation (GCR) is enabled.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -230,6 +230,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"disableAllocationSinking",           "O\tdon't delay object allocations until immediately before the corresponding constructor calls", TR::Options::disableOptimization, allocationSinking, 0, "P"},
    {"disableAndSimplification",           "O\tdisable and simplification",                     TR::Options::disableOptimization, andSimplification, 0, "P"},
    {DisableAnnotations,                   "O\tdisable annotation support",                     RESET_OPTION_BIT(TR_EnableAnnotations), "F"},
+   {"disableAOTAtCheapWarm",              "O\tdisable AOT with cheap warm opt level", SET_OPTION_BIT(TR_DisableAotAtCheapWarm), "F", NOT_IN_SUBSET},
    {"disableAOTCheckCastInlining",        "O\tdisable AOT check cast inlining",                SET_OPTION_BIT(TR_DisableAOTCheckCastInlining), "F"},
    {"disableAOTColdCheapTacticalGRA",   "O\tdisable AOT cold cheap tactical GRA",                      SET_OPTION_BIT(TR_DisableAOTColdCheapTacticalGRA), "F"},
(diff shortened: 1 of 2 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -2217,6 +2218,14 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
             if (_coldUpgradeSampleThreshold == TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD)
                _coldUpgradeSampleThreshold = 2;
             }
+         else
+            {
+            // For AOT with GCR enabled we can be more conservative with 
+            // upgrades through sampling because we can rely on GCR
+            if (!self()->getOption(TR_DisableGuardedCountingRecompilations) &&
+              _coldUpgradeSampleThreshold == TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD)
+               _coldUpgradeSampleThreshold = 10;
```
vijaysun-omr: Is the purpose here to reduce the likelihood of sampling causing upgrade compilations during startup ? i.e. GCR would be expected to cause many more upgrades outside of startup anyway.

compiler/control/OMROptions.cpp
```diff
@@ -2217,6 +2218,14 @@ OMR::Options::jitLatePostProcess(TR::OptionSet *optionSet, void * jitConfig)
             if (_coldUpgradeSampleThreshold == TR_DEFAULT_COLD_UPGRADE_SAMPLE_THRESHOLD)
                _coldUpgradeSampleThreshold = 2;
             }
+         else
+            {
+            // For AOT with GCR enabled we can be more conservative with 
+            // upgrades through sampling because we can rely on GCR
```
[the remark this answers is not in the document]
mpirvu: The new option is going to be checked in the downstream project, but regardless of the new option's value, if AOT and GCR are enabled, upgrading through sampling is going to be more conservative. I felt this is a good change to make in any case.
As I mentioned in private conversation, upgrades during startup can cause temporary backups of compilations: once the upgrade starts the JIT is not going to cancel it and while this upgrade is in progress many AOT loads are just waiting to be server (GCR recomps don't have this problem because they are allowed only after startup).
With this solution we allow upgrades through sampling only for methods that receive a "significant" amount of samples and thus they should be worthy of an upgrade.

vijaysun-omr:
@genie-omr build all

compiler/control/OMROptions.cpp
mstoodle: Could we maybe add a comment to explain `10` ?

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Merging this since review comments have been addressed and checks have passed.

PR #3324, opened 2018-12-11
Add static symbol for constant pool address

cathyzhyi:
Create a static symbol to represent the constant pool address for a
resolved method so that a loadaddr node can be created when the CP
address is needed in trees and AOT can relocate the address value
correctly.

### DIFF
--- a/compiler/il/symbol/OMRSymbol.hpp
+++ b/compiler/il/symbol/OMRSymbol.hpp
@@ -340,6 +340,9 @@ public:
    inline void setGCRPatchPoint();
    inline bool isGCRPatchPoint();
 
+   inline void setConstantPoolAddress();
+   inline bool isConstantPoolAddress();
+
    // flag methods specific to resolved
    //
    inline bool isJittedMethod();
@@ -507,6 +510,7 @@ public:
       StartPC                   = 0x04000000,
       CountForRecompile         = 0x02000000,
       RecompilationCounter      = 0x01000000,
+      ConstantPoolAddress       = 0x00800000,
       GCRPatchPoint             = 0x00400000,
 
       //Only Used by Symbols for which isResolvedMethod is true;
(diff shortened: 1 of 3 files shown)


0xdaryl:
This seems very Java-specific.  Can this not be implemented in the OpenJ9 extension of Symbol?

cathyzhyi:
> This seems very Java-specific. Can this not be implemented in the OpenJ9 extension of Symbol?

Yea, I agree the Java-specific symbols should be added to Openj9 but currently, I think we don't have support for this. The existing symbols like method handle, constant dynamic are added in OMR layer as well.  I am not very sure what's involved in adding the support. We probably can just add an enum flag field the OpenJ9 extension of Symbol?

0xdaryl:
It depends on how you're going to use the symbol I suppose.  This PR doesn't give any clues to that.

For example, rather than using an identity query on a symbol (e.g., is this a static symbol for a constant pool address, or is this a constant dynamic symbol?) I think we should be striving for capability based queries instead (i.e., what is the property or capability provided by this symbol that's necessary for you to make a decision about something?).

cathyzhyi:
The constant pool symbol added here will be used for adding AOT relocation in this PR https://github.com/eclipse/omr/pull/3334. The code for AOT relocation for constant pool is mostly in OMR layer as well.

andrewcraik:
@0xdaryl your thoughts in light of the above?

0xdaryl:
If the usage of this symbol will be in non-J9_PROJECT_SPECIFIC code in OMR and is in line with the usage other existing, related symbols then I can accept that.  It was this clarification that I was asking for originally.

In general, I'm not in favour of explicitly adding any new J9_PROJECT_SPECIFIC macros.  At some point, however, we must stop kicking the can down the road and actually commit to a proper design that aligns with the OMR project's goals.

0xdaryl:
@genie-omr build all

PR #3327, opened 2018-12-12
Fix get_OS_version for MacOS

pdbain-ibm:
Also define platform-specific constants to reduce the complexity of the ifdefs
and reduce duplicated code.

### DIFF
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -40,6 +40,9 @@
 #include <mach-o/dyld.h>
 #include <sys/param.h>
 #include <sys/mount.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
 #endif /* defined(LINUX) && !defined(OMRZTPF) */
 
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


charliegracie:
Can you add some comments to the code explaining why OS X is not using the same code as other `unix` style platforms? Thanks

pdbain-ibm:
@charliegracie 
Thanks for the suggestion.  Done.

port/unix/omrsysinfo.c
```diff
@@ -581,46 +584,66 @@ omrsysinfo_get_OS_type(struct OMRPortLibrary *portLibrary)
...
 	if (NULL == PPG_si_osVersion) {
 		int rc;
+#if !defined (OSX)
 		struct utsname sysinfo;
+#endif
 
 #ifdef J9ZOS390
 		rc = __osname(&sysinfo);
+#elif defined (OSX)
+			/* uname() on MacOS returns the version of the Darwin kernel.
```
charliegracie: It looks like there is an extra tab for the comment. Could you align the comment with the code?  Thanks

Thanks for the comment as it is exactly what I was looking for!

charliegracie:
@genie-omr build all

PR #3330, opened 2018-12-12
Reduce size of *.i files

keithc-ca:
Omit useless lines with only '@'.

### DIFF
--- a/omrmakefiles/rules.mk
+++ b/omrmakefiles/rules.mk
@@ -207,7 +207,7 @@ endef
 endif
 
 DDR_SED_COMMAND := \
-    sed -n -e '/^DDRFILE_BEGIN /,/^DDRFILE_END /s/^/@/p'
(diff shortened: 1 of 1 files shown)


omrmakefiles/rules.mk
```diff
@@ -207,7 +207,7 @@ endef
 endif
 
 DDR_SED_COMMAND := \
-    sed -n -e '/^DDRFILE_BEGIN /,/^DDRFILE_END /s/^/@/p'
+    sed -n -e '/^DDRFILE_BEGIN /,/^DDRFILE_END /s/^/@/' -e '/^@./p'
```
charliegracie: Is there a similar change required for the CMake infrastructure?

keithc-ca: I don't believe DDR is supported with CMake yet.

charliegracie:
@genie-omr build zos,aarch64

PR #3331, opened 2018-12-12
 TGC root scan time

amicic:
1) Flags for reducing  printing

Introduce two flags (global and per thread) that will assist in reducing
the amount of prints for TGC root scan time. They will be printed only
if any thread worked on any of roots for a given increment. This will be
particularly useful for Metronome (in OpenJ9) since stats are reported
for each GC increment, and most of increments do not deal with roots.

2) Max Increment time

Introduce fields root scan stats for max increment time and entity that
that owns that max increment. This values will be reported by
incremental GCs, like Metronome GC, where scan time spent for one entity
could span over several increments.


charliegracie:
@genie-omr build all

amicic:
More code added, please re-review.

charliegracie:
@genie-omr build all

PR #3334, opened 2018-12-13
Add AOT support on X86 for CP address symbol

cathyzhyi:
Add relocation record for constant pool address symbol and
the existing AOT infra can relocate the address value correctly.

### DIFF
--- a/compiler/x/codegen/OMRX86Instruction.cpp
+++ b/compiler/x/codegen/OMRX86Instruction.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


compiler/x/codegen/OMRX86Instruction.cpp
```diff
@@ -1137,7 +1137,8 @@ void
 TR::X86RegImmSymInstruction::autoSetReloKind()
    {
    TR::Symbol *symbol = getSymbolReference()->getSymbol();
-   if (symbol->isConst())
+   if (symbol->isConst() ||
+       symbol->isConstantPoolAddress())
```
Leonardo2718: Since these are OpenJ9 queries, shouldn't they be guarded with `J9_PROJECT_SPECIFIC` (as much as it pains me to add more of those)?

This is what's causing the CI failures.

Leonardo2718: Oh wait, this requires #3324 ?

andrewcraik: I would note that the AOT infrastructure has not been factored for OMR and OpenJ9 yet (and is missing extension points etc). I agree J9 code should go in J9, imposing the complete re-architecture as an overhead to contribution is impractical given the scope of the code being discussed here - the refactor is a several large items itself.

cathyzhyi:
@Leonardo2718 https://github.com/eclipse/omr/pull/3324 is merged and @dsouzai is OK with the AOT part. could you help to merge this one? Thanks!

Leonardo2718:

PR #3336, opened 2018-12-13
Perform wrtbar versioning conservatively

yanluo7:
If a GC point is triggered mid-loop, destination object of
the wrtbar can move/tenure. And if it is not remembered, upon return
from GC the remaining loop iterations will not run any wrtbars
since we had versioned them out. The fix here is to be conservative
when versioning wrtbars - for correctness, if there is a GC
point in the loop, wrtbar versioning will not be performed.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -433,6 +433,7 @@ int32_t TR_LoopVersioner::performWithoutDominators()
          naturalLoop->computeInvariantExpressions();
          }
 
+      _skipWrtbarVersion = false;
       _seenDefinedSymbolReferences->empty();
       _writtenAndNotJustForHeapification->empty();
       _additionInfo->empty();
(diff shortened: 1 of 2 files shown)


yanluo7:
Perf results skipWrtbar vs baseline :
SPECjbb2015: no impact on max-jops, critical-jops dropped 1% w/ this change.
SPECjbb2005/power: no impact on throughput
Hibench: no impact on scores.

yanluo7:
@andrewcraik FYI

vijaysun-omr:
Please remove Java specific benchmark information from this OMR PR.

compiler/optimizer/LoopVersioner.cpp
```diff
@@ -3285,6 +3292,13 @@ bool TR_LoopVersioner::detectChecksToBeEliminated(TR_RegionStructure *whileLoop,
          _nullCheckReference = NULL;
          _inNullCheckReference = false;
 
+         // If the node is a GC point (either cause GC and return, or cause GC and throw),
+         // be conservative and do not version wrtbars. Applies to top level nodes/checks
+         // and child call nodes.
+         if (currentNode->canCauseGC() || ((currentNode->getNumChildren() > 0) &&
```
vijaysun-omr: Is this really necessary to check both the parent and the child node for canCauseGC ? 
Maybe this kind of checking on the child ought to be folded into the canCauseGC routine so that every caller does not have to do logic such as this.

yanluo7: Upon inspecting `Node::canCauseGC` more closely, it does take into account both top level and child nodes, so calling it on the top level node is sufficient.

vijaysun-omr: Thanks, the code looks simpler and cleaner now in this area.

vijaysun-omr:
I would also like a review from @andrewcraik on this.

vijaysun-omr:
@genie-omr build all

andrewcraik:
@vijaysun-omr looks good to me - nice and simple to achieve the affect desired

vijaysun-omr:
Looks safe to me. Merging.

PR #3337, opened 2018-12-13
 Correctly fill 64-bit registers on 31-bit in OOL

fjeremic:
When a register is spilled in the mainline but used in the out of line
(OOL) code section we are forced to perform a spill and a fill in the
OOL such that the register state remains the same on both paths of
execution.

The way this is handled on Z is that `SpilledReg` register dependencies
are added to the OOL entry label which represent the state of the
machine at the branch to the OOL entry label in the mainline.

Because a register may have been spilled in the mainline and used in
the OOL we must perform a fill (generate a load). When 64-bit registers
are available for use in 31-bit targets the register kind will be a
`TR_GPR64`, however currently the `is64BitReg` flags are not properly
set for these register kinds and as such we cannot use this flag as an
"end-all" answer to whether we should generate a 32-bit fill or a
64-bit fill (decision to use `L` vs. `LG` instructions).

Until this is resolved so that we use the `is64BitReg` flag on both
31-bit and 64-bit we must honor the decision made by the register kind
when doing a fill for an OOL section.

See OutOfLineCodeSection.hpp for more details on how local RA handles
OOL code sections.

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -3962,56 +3962,44 @@ OMR::Z::CodeGenerator::gprClobberEvaluate(TR::Node * node, bool force_copy, bool
    if (!self()->canClobberNodesRegister(node, 1, &data, ignoreRefCount) || force_copy)
       {
       char * CLOBBER_EVAL = "LR=Clobber_eval";
-      if ((node->getType().isInt64() || TR::Compiler->target.is64Bit() )
-         && TR::Compiler->target.is32Bit())
+      if (TR::Compiler->target.is32Bit() && !self()->use64BitRegsOn32Bit() && node->getType().isInt64())
          {
-         if (self()->use64BitRegsOn32Bit())
+         TR::RegisterPair * trgRegPair = (TR::RegisterPair *) srcRegister;
+         TR::Register * lowRegister = srcRegister->getLowOrder();
+         TR::Register * highRegister = srcRegister->getHighOrder();
+         bool allocatePair=false;
+         if (!data.canClobberLowWord())
             {
-            TR::Register * targetRegister = self()->allocate64bitRegister();
-
-            cursor = generateRRInstruction(self(), TR::InstOpCode::LGR, node, targetRegister, srcRegister);
-
-            return targetRegister;
+            lowRegister = self()->allocateRegister();
+            self()->stopUsingRegister(lowRegister); // Allocate pair will make these live again
+            allocatePair = true;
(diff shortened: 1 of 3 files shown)


fjeremic:
Requesting review from @NigelYiboYu or @sehirst. @dhong44 FYI.

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -4083,7 +4062,7 @@ OMR::Z::CodeGenerator::gprClobberEvaluate(TR::Node * node, bool force_copy, bool
          return targetRegister;
          }
       }
-   return self()->evaluate(node);
+   return srcRegister;
```
sehirst: Can you explain why it's ok to no longer evaluate the node and instead just return `srcRegister`?

compiler/z/codegen/OMRRegisterDependency.cpp
```diff
@@ -959,6 +959,11 @@ TR_S390RegisterDependencyGroup::assignRegisters(TR::Instruction   *currentInstru
                      break;
                   case TR_GPR:
                      opCode = TR::InstOpCode::getLoadOpCode();
+                     
+                     if (virtReg->is64BitReg())
+                        {
+                        opCode = TR::InstOpCode::LG;
```
sehirst: Is this not an issue for spill instructions, i.e. there's no corresponding change needed for ST/STG?

compiler/z/codegen/OMRCodeGenerator.cpp
fjeremic: The `node` has already been evaluated at the start of the evaluator. Calling `evaluate` on it again will simply return the same register as before, so let's save ourselves the function call and return it directly.

compiler/z/codegen/OMRRegisterDependency.cpp
fjeremic: I don't think so. See commit body of 55761eb for an explanation. This piece of code is very special for the way we handle spilled registers in the mainline but filled in the OOL code path. The fill portion of the puzzle happens in [1] where we check for `TR_GPR64` and correctly generate a 64-bit store to fill.

[1] https://github.com/eclipse/omr/blob/552260112c0871db8981a2b2e31f148b98c41015/compiler/z/codegen/OMRMachine.cpp#L4485-L4488

fjeremic:
@genie-omr build zos zlinux

fjeremic:
@0xdaryl / @vijaysun-omr could I get one of you to pick this one up?

PR #3341, opened 2018-12-14
AArch64: Implement aconst/aloadEvaluator()

knn-k:
This commit implements evaluators for aconst and aload for aarch64.

### DIFF
--- a/compiler/aarch64/codegen/OMRTreeEvaluator.cpp
+++ b/compiler/aarch64/codegen/OMRTreeEvaluator.cpp
@@ -225,10 +225,41 @@ OMR::ARM64::TreeEvaluator::iloadEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 // also handles aloadi
 TR::Register *
 OMR::ARM64::TreeEvaluator::aloadEvaluator(TR::Node *node, TR::CodeGenerator *cg)
-	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::aloadEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
-	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
-	}
+   {
+   TR::Compilation *comp = cg->comp();
+   TR::Register *tempReg;
+
+   if (!node->getSymbolReference()->getSymbol()->isInternalPointer())
+      {
+      if (node->getSymbolReference()->getSymbol()->isNotCollected())
+         tempReg = cg->allocateRegister();
+      else
+         tempReg = cg->allocateCollectedReferenceRegister();
+      }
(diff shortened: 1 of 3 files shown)


compiler/aarch64/codegen/OMRTreeEvaluator.cpp
```diff
@@ -221,12 +221,57 @@ OMR::ARM64::TreeEvaluator::iloadEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
 TR::Register *
 OMR::ARM64::TreeEvaluator::aloadEvaluator(TR::Node *node, TR::CodeGenerator *cg)
-	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::aloadEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
-	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
-	}
+   {
+   TR::Compilation *comp = cg->comp();
+   // NEW to check for it feeding to a single vgdnop
+   static bool disableLoadForVGDNOP = (feGetEnv("TR_DisableLoadForVGDNOP") != NULL);
```
0xdaryl: I realize this was inherited from ARM32, but before we introduce this into AArch64 can you explain what this does (or will do when it's finished)?

compiler/aarch64/codegen/UnaryEvaluator.cpp
```diff
@@ -54,10 +54,42 @@ TR::Register *OMR::ARM64::TreeEvaluator::cconstEvaluator(TR::Node *node, TR::Cod
...
+      TR::Register *tempReg = node->setRegister(cg->allocateRegister());
+      TR::Instruction *cursor = loadConstant64(cg, node, address, tempReg);
+      if (isPicSite)
+         {
+         if (node->getOpCode().isCall() || node->isClassPointerConstant())
+            comp->getStaticPICSites()->push_front(cursor);
+         else if (node->isMethodPointerConstant())
+            comp->getStaticMethodPICSites()->push_front(cursor);
+         }
+      return tempReg;
```
0xdaryl: Same comment as above.  This is potentially pulling in a lot of technical debt into AArch64.  Are you able to confirm that all of this will work this way on AArch64 and is required?

knn-k:
I removed VGDNOP-related, PicSite-related, and AddressConstantInSnippet-related code from my previous commit.
I will revisit them when I understand what they are and how they should be implemented for aarch64.
I leave `aconstEvaluator()` separate from `lconstEvaluator()`.  They are mostly the same at this point, but I think `aconstEvaluator()` will need additional code in it later.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3342, opened 2018-12-14
Tril tests for i2f, etc.

knn-k:
This commit implements Tril test cases for i2f, etc.


knn-k:
This commit contains tests for integer->FP conversion.
I am planning to add tests for f2d, d2f, and FP->integer, but I need to wait for discussion in #3332 on how to pass FP values in `UsingConst` variants.

Leonardo2718:
@genie-omr build all

Leonardo2718:
The failure on ppc is due to #3369. @knn-k can you `#ifndef TR_TARGET_POWER` out the test case for power and then open an issue to remove the guard once #3369 is resolved?

knn-k:
Opened Issue #3375 for removing the guard for ppc later.

PR #3343, opened 2018-12-14
Add C++ wrappers for OMR global singletons

rwy0717:
This PR adds three utilities:
1. Runtime: Wraps up the Platform interface and the OMR_Runtime.
2. PlatformInterface: Wrapper for the port library and the threading interface
3. ThreadingInterface / Thread: Wrapper for the thread library.

OMR::Runtime bundles together the global singletons which every VM-level class (GC, JIT, RAS, etc) would depend on.

### DIFF
--- a/fvtest/gc_api_test/CMakeLists.txt
+++ b/fvtest/gc_api_test/CMakeLists.txt
@@ -39,3 +39,7 @@ function(omr_add_gc_test testname)
 		COMMAND "${testname}"
 	)
 endfunction(omr_add_gc_test)
+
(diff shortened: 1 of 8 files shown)


rwy0717:
Yep, they are very related. This is just a super thin C++ wrapper around initializing the OMR_Runtime, portlib, and the thread library. The "real" code is still in "omr/omr". There's no equivalent to the VM (yet).

charliegracie:
@genie-omr build all

rwy0717:
Closing in favour of #3806, a different take on the same idea.

PR #3344, opened 2018-12-14
Add ErrorCode, ErrorCategory, Unique(Singleton)

rwy0717:
OMR has a lot of independent components, and a lot of sources of error.
There is no single view / enum of all possible errors across all
components. In the VM startup, this is a particular problem, because the
VM is responsible for initializing all other OMR components. The VM
needs to communicate what component failed, and how.

This PR adds a new ErrorCode and ErrorCategory utility to OMR. Each
instance of the ErrorCategory class is a static global singleton.
ErrorCategory values are comparable by identity (address), so are only
equal to themselves.  This allows each component (or even OMR consumers)
to create their own globally unique ErrorCategory values, without any
centralized coordination.

The ErrorCode class associates an integer code with an ErrorCategory,
and can be used as a more general error code. Through it's category,
it's possible to translate an ErrorCode to a string using the message()
member function.

ErrorCode and ErrorCategory's design is based off the C++11 std
error_condition and error_category. The differences are:

0. OMR-style naming.
1. No bool operator for ErrorCode. Instead, use the member functions
   ok(), bad(), or just test the error's value().
2. ErrorCategories translate error codes into static const char* data,
   rather than return a std::string, which we can not use.
3. No special support for enum classes, which our compilers don't yet
   support.

As long as OMR does not link against the standard library, we
will not be able to use std error_code.

The ErrorCategory is built off a new Unique utility, which is a general
utility class for defining globally unique values. In the future, Unique
could be used to create new kinds of extensibility in OMR.

### DIFF
--- a/fvtest/coretest/CMakeLists.txt
+++ b/fvtest/coretest/CMakeLists.txt
@@ -38,5 +38,7 @@ function(omr_add_core_test testname)
 endfunction(omr_add_core_test)
 
 omr_add_core_test(TestBytes)
+omr_add_core_test(TestErrorCode)
 omr_add_core_test(TestIntrusiveList)
 omr_add_core_test(TestTypeTraits)
+omr_add_core_test(TestUnique)
--- a/fvtest/coretest/TestErrorCode.cpp
+++ b/fvtest/coretest/TestErrorCode.cpp
@@ -1,0 +1,133 @@
+/*******************************************************************************
+ *  Copyright (c) 2018, 2018 IBM and others
+ *
+ *  This program and the accompanying materials are made available under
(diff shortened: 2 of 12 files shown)


rwy0717:
This is more of an RFC than a real PR. 

Basically, ErrorCode is an (int, category) pair, and the category can translate the error into a string message through a virtual function call. This gives us a way to take error codes from multiple C libraries (port, thread, ras, etc?) and roll them into a single extensible value type. If we're looking at representing errors in C++ libraries, there are probably better ways to do this.

rwy0717:
Added a missing copyright

PR #3346, opened 2018-12-14
Split the example glue into base and extended

rwy0717:
When the OMR GC API is enabled, certain elements of the glue will be
provided by the GC. To ensure that every component finds the correct
copy of the glue headers, the example's version has to be taken out of
the include path.

To facilitate this, the example's glue headers are split into an include
and an include_extended directory. The include_extended directory is
only added to the includes when OMR_GC_API is OFF.

As well, some GC glue sources are only added if the GC API is off, since
those are provided by rthe GC as well.


charliegracie:
@dnakamura and @youngar can you review please?

0xdaryl:
It doesn't appear much has happened with this PR since it was opened.  @dnakamura and @youngar could you review as requested?  A rebase is in order too.

rwy0717:
This PR isn't needed, because I'm not going to do the "high level GC API" aka gcbuilder anymore (at least for the time being).

Issue #3206, opened 2018-11-18
NotEqualTo evaluates to false for NaNs


aviansie-ben:
It seems that the `IlBuilder::NotEqualTo` service has surprising behaviour with respect to NaN when comparing floating-point types. This service ends up returning 0 when comparing a value against NaN because it uses the `fcmpne`/`dcmpne` opcodes, which evaluate to false for unordered values, instead of the more appropriate `fcmpneu`/`dcmpneu` opcodes.

This behaviour is not in line with the IEEE 754 standard, which dictates that `x != NaN` should be true for all values of x. Considering that most (if not all) languages agree on this, the behaviour of the JitBuilder service is incredibly surprising and this behaviour is probably never what the user intends.

mstoodle:
It looks to me like no one in our consumer list is relying on compareOp() to return the "non-u" versions of dcmp/fcmp opcodes, so there shouldn't be any issue "changing" OMR's decision on this front. I believe @Leonardo2718 mentioned he had tried this in the past and there were some issues that came up in the evaluators (like, maybe they're not all implemented :) ).

@Leonardo2718 can you please outline what you found here, so we can capture what we think is needed here to move forward?

mstoodle:
bah just noticed @aviansie-ben 's cross-link. Never mind @Leonardo2718 !

Leonardo2718:
@mstoodle Yes, the cross-linked issue covers what I had mentioned earlier.

PR #3347, opened 2018-12-14
Fix IlBuilder::NotEqualTo for floating-point NaNs

aviansie-ben:
Previously, the IlBuilder::NotEqualTo service was generating fcmpne and
dcmpne opcodes when comparing floating-point values. However, these
opcodes return false when their operands are unordered with respect to
each other (i.e. when one or both are NaN). This result is unexpected
and is almost never what a JitBuilder user wants to generate, as the
IEEE 754 standard (whose behaviour is mandated by many languages) says
that non-equality should return true for unordered operands.

The fcmpneu and dcmpneu opcodes do exhibit the expected behaviour with
regards to unordered operands. Thus, in order to better conform to user
expectations, IlBuilder::NotEqualTo will now emit those opcodes instead
when comparing floating-point values.

Fixes: #3206

### DIFF
--- a/compiler/il/OMRILOps.cpp
+++ b/compiler/il/OMRILOps.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corp. and others
+ * Copyright (c) 2000, 2018 IBM Corp. and others
  *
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

PR #3349, opened 2018-12-14
Add OMR::GC::Handle<T> value type

rwy0717:
Handle<T> is a pointer-like value that indirects through an externally
rooted slot. This allows users to pass rooted gc-references through
function calls, independently of how the object is actually rooted.  The
main utility of this class is that it _signifies_ that the caller has
rooted the parameters of a function.

It is possible for Handle<T>'s to outlive their roots, which is an
extremely dangerous error. In the future, we may chose to add "handle
counting" to specialized root structures, to ensure there are no
outstanding handles when a root dies.

Another potential future enhancement would be to parameterize the root
type, to allow for different gc-reference encodings. The value of doing
this is questionable though, since the main function of the handle is to
hide the rooting mechanism, and doing this puts it right into the type.

### DIFF
--- a/gc/api/include/OMR/GC/Handle.hpp
+++ b/gc/api/include/OMR/GC/Handle.hpp
@@ -169,4 +169,4 @@ operator!=(T* lhs, Handle<U> rhs)
 } /* namespace GC */
 } /* namespace OMR */
 
-#endif // OMR_GC_HANDLE_HPP_
+#endif /* OMR_GC_HANDLE_HPP_ */
--- a/gc/api/include/OMR/GC/Handle.hpp
+++ b/gc/api/include/OMR/GC/Handle.hpp
@@ -1,0 +1,172 @@
+/*******************************************************************************
+ *  Copyright (c) 2018, 2019 IBM and others
+ *
+ *  This program and the accompanying materials are made available under
+ *  the terms of the Eclipse Public License 2.0 which accompanies this
+ *  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ *  or the Apache License, Version 2.0 which accompanies this distribution and
+ *  is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ *  This Source Code may also be made available under the following
+ *  Secondary Licenses when the conditions for such availability set
+ *  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ *  General Public License, version 2 with the GNU Classpath
+ *  Exception [1] and GNU General Public License, version 2 with the
+ *  OpenJDK Assembly Exception [2].
+ *
+ *  [1] https://www.gnu.org/software/classpath/license.html
+ *  [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ *  SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#if !defined(OMR_GC_HANDLE_HPP_)
+#define OMR_GC_HANDLE_HPP_
+
+#include <OMR/TypeTraits.hpp>
+#include <cstddef>
+#include <omrcfg.h>
+
+namespace OMR
+{
+namespace GC
+{
+
+template <typename T>
+class HandleBase
+{
+public:
+	HandleBase() {}
+
+	HandleBase(T* const volatile* root) : _root(root) {}
+
+	/**
+	 * Convert this Handle to a raw indirect pointer. The address of the rooting slot.
+	 */
+	T* const volatile* root() const { return this->_root; }
+
+	/**
+	 * Convert this Handle to a raw pointer.
+	 */
+	T* get() const { return *_root; }
+
+protected:
+	T* const volatile* _root;
+};
+
+template <typename T, bool DEFINE_OPERATORS = !IsVoid<T>::VALUE>
+class HandlePtrOps;
+
+template <typename T>
+class HandlePtrOps<T, true> : public HandleBase<T>
+{
+public:
+	HandlePtrOps() : HandleBase<T>() {}
+
+	HandlePtrOps(T* const volatile* root) : HandleBase<T>(root) {}
+
+	T& operator*() const { return *this->get(); }
+
+	T* operator->() const { return this->get(); }
+};
+
+template <typename T>
+class HandlePtrOps<T, false> : public HandleBase<T>
+{
+public:
+	HandlePtrOps() : HandleBase<T>() {}
+
+	HandlePtrOps(T* const volatile* root) : HandleBase<T>(root) {}
+};
+
+/**
+ * The Handle<T> provides pointer semantics by indirecting through roots.
+ *
+ * A Handle<T> is associated with a single root slot, such as an instance of StackRoot<T>. When a Handle<T> is
+ * dereferenced, it will access the target indirectly through the associated root. Because Handles always indirect
+ * through roots, they are safe to hold across a collection. A Handle is invalid when it's root is destroyed. Handles
+ * are fast and flexible to use, but care must be taken so a Handle does not outlive it's root.
+ */
+template <typename T>
+class Handle : public HandlePtrOps<T>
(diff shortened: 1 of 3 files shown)


rwy0717:
There's one funny issue where Handle<T> is not directly comparable to NULL, since NULL should never be compared to an object. NULL has to be cast to a pointer first.

rwy0717:
@dnakamura can I get a review please?

rwy0717:
Removed a stale union in HandleBase that wasn't needed.

charliegracie:
@genie-omr build all

charliegracie:
Waiting for @dnakamura and CI testing to finish before merging

rwy0717:
I think the AIX build might have failed due to an infrastructure issue, in cleanup workspace.

dnakamura:
Overall LGTM, however I'm not crazy about the naming of `HandleMixin` given that the way that its used is not really a mixin. Also I think the class could be eliminated entirely by a refactor, but id have to try it to be sure

rwy0717:
> I'm not crazy about the naming of HandleMixin

How about `HandlePtrOps` ? I was thinking, it _could_ be specialized to add functionality for a specific type, but that's probably a bad idea. Should I move HandleBase and HandleMixin to an Internal or Detail namespace?

>  I think the class could be eliminated entirely by a refactor

You're welcome to try! I couldn't find a way.

rwy0717:
Changes:
- Add test for assignment, converting assignment
- Disable converting copy constructor when T==U, allowing the compiler to define a default/trivial copy constructor.
- Renamed HandleMixin to HandlePtrOps.
- Separated HandlePtrOps into three pieces: the declaration, and the two specializations for void or not-void.

rwy0717:
Fixed a typo.

charliegracie:
@dnakamura are you good with these changes now?

charliegracie:
@genie-omr build all

rwy0717:
Well, I found a way to do this without inheritance. I'm not sure it's a better solution. To me it seems like we're entering metaprogramming hell, but I have never loved enable_if.

```c++
template <typename T>
class Handle
{
public:
	/// Uninitialized handle object. Very dangerous.
	Handle() : _root() {}

	template <typename U>
	explicit Handle(U* const volatile* root) : _root(root) {}

	/// Copy constructor. Disabled when T = U, where we rely on the implicit copy constructor.
	template <typename U, typename = typename EnableIf<!IsSame<T, U>::VALUE>::Type>
	Handle(const Handle<U>& other) : _root(other.root()) {}

	/// Convert this Handle to a raw indirect pointer. The address of the rooting slot.
	T* const volatile* root() const { return _root; }

	/// Convert this Handle to a raw pointer.
	T* get() const { return *_root; }

	/// Dereference operator. Only available when T is not void.
	template <typename U = T, typename = typename EnableIf<!IsVoid<U>::VALUE>::Type>
	U& operator*() const { return *get(); }

	/// Dereference operator. Only available when T is not void.
	template <typename U = T, typename = typename EnableIf<!IsVoid<U>::VALUE>::Type>
	U* operator->() const { return get(); }

private:
	T* const volatile* _root;
};
```

The trick is to force the template to be dependent on a template parameter, which stops the compiler from statically instantiating the operator templates.

Do we like this solution better? It certainly is shorter.

Leonardo2718:
An alternative that does use inheritance (with the CRTP) but is still fairly short:

```c++
template<typename T, typename D>
class WithPtrOp {
public:
	/// Dereference operator. Only available when T is not void.
	T& operator*() const { return *static_cast<D>(this)->get(); }

	/// Dereference operator. Only available when T is not void.
	T* operator->() const { return static_cast<D>(this)->get(); }
};

template<typename D>
class WithPtrOp<void, D> {};

template <typename T>
class Handle: public WithPtrOp<T, Handle<T> >
{
public:
	/// Uninitialized handle object. Very dangerous.
	Handle() : _root() {}

	template <typename U>
	explicit Handle(U* const volatile* root) : _root(root) {}

	/// Copy constructor. Disabled when T = U, where we rely on the implicit copy constructor.
	template <typename U, typename = typename EnableIf<!IsSame<T, U>::VALUE>::Type>
	Handle(const Handle<U>& other) : _root(other.root()) {}

private:
	T* const volatile* _root;
};
```

gc/api/include/OMR/GC/Handle.hpp
```diff
@@ -0,0 +1,147 @@
...
+{
+namespace GC
+{
+
+/// A CRTP mixin which adds self() calls which will cast this to t
+template <typename D, typename T>
+class Mixin
+{
+private:
+	friend class D;
```
Leonardo2718: I don't think this does what you intended. `D` is not a class name.

gc/api/include/OMR/GC/Handle.hpp
```diff
@@ -0,0 +1,147 @@
...
+namespace OMR
+{
+namespace GC
+{
+
+/// A CRTP mixin which adds self() calls which will cast this to t
+template <typename D, typename T>
+class Mixin
+{
+private:
```
Leonardo2718:
```suggestion
protected:
```

rwy0717: I went with private, so that mixins wouldn't expose the self() method to the base class. I have a suspicion that it would not play out well when a class inherits from multiple CRTP mixins.

rwy0717:
So, do we like the mixin-based version the best? Is this what I should go with?

gc/api/include/OMR/GC/Handle.hpp
rwy0717: Argh! It seems your right, but this was compiling! There is apparently a workaround for pre-c++11:
```c++
friend class typename TypeAlias<D>::Type;
```

Should I go with something like this?

gc/api/include/OMR/GC/Handle.hpp
Leonardo2718: The thing is, private fields are not accessible by derived classes.

rwy0717:
After reviewing the alternatives, I feel that the original, being the most straightforward, is the best implementation for now. At some point down the road, we may wish to revisit using a CRTP mixin to package pointer operators, or conditionally enable the operators with EnableIf, or (most likely) both. I've reverted the PR to the original gc-handle implementation.

gc/api/include/OMR/GC/Handle.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+template <typename T>
+class HandleBase
+{
+public:
+	HandleBase() {}
+
+	HandleBase(T* const volatile* root) : _root(root) {}
+
+	/// Convert this Handle to a raw indirect pointer. The address of the rooting slot.
```
charliegracie: Why are all of the comments using `///` instead of `//` or `/* .. */`?

rwy0717: They're meant to be doxygen comments.

rwy0717: Although, after skimming the doxygen documentation, it looks like these comments need to be at least 2 lines before doxygen will pick them up? Should I change to javadoc-style comments? eg
```c++
/**
 * Convert this Handle to a raw indirect pointer. The address of the rooting slot.
 */
```

charliegracie: Yes I would convert them to the javadoc style.  Thanks

rwy0717:
Updated comment style, updated copyright, and rebased.

gc/api/include/OMR/GC/Handle.hpp
```diff
@@ -0,0 +1,172 @@
...
+bool
+operator!=(T* lhs, Handle<U> rhs)
+{
+	return lhs != rhs.get();
+}
+
+} /* namespace GC */
+} /* namespace OMR */
+
+#endif // OMR_GC_HANDLE_HPP_
```
youngar:
```suggestion
#endif /* OMR_GC_HANDLE_HPP_ */
```

youngar:
@genie-omr build all

rwy0717:
@youngar can you please relaunch the zos build?

youngar:
@genie-omr build zos

rwy0717:
This PR isn't needed anymore.

PR #3338, opened 2018-12-13
Tril Tests for iand and ixor

aarongraham9:
Tril Tests for iand and ixor

Signed-off-by: Aaron Graham <aaron.graham@unb.ca>

PR #3351, opened 2018-12-14
WIP: AArch64: Implement iand/ior/ixor Evaluators

aarongraham9:
AArch64: Implement iandEvaluator, iorEvaluator and ixorEvaluator
-Enable Corresponding Evaluators in TreeEvaluatorTable


aarongraham9:
Corresponding Tril Tests were merged in #3338.

aarongraham9:
- Commoned-up the functionality in my `iandEvaluator`, `iorEvaluator` and `ixorEvaluator` Evaluators.
- Tightened up the use of virtual registers. It now re-uses one if possible; otherwise, it allocates a new one.  
- Rebased this PR off of the `master` branch.
- Related Tril Regression Tests for `iand`, `ior` and `ixor` OMR IL OpCodes can be found in PR: #3338.

- Tril Test Results for `iand`, `ior` and `ixor` OMR IL OpCodes/Evaluators are as follows:
`./comptest --gtest_filter=*LogicalTest*Int*LogicalBinary*`:
```
Note: Google Test filter = *LogicalTest*Int*LogicalBinary*
[==========] Running 675 tests from 1 test case.
[----------] 675 tests from LogicalTest/Int32LogicalBinary
[----------] 675 tests from LogicalTest/Int32LogicalBinary (10972 ms total)

[==========] 675 tests from 1 test case ran. (10998 ms total)
[  PASSED  ] 675 tests.
[  ALL TESTS PASSED  ]
```

Removing the WIP as this should be good to go for review;. please let me know if any of you have any comments or concerns: @0xdaryl, @knn-k, @Leonardo2718, @jeanlego, @marufunb and @hsoontie.

aarongraham9:
Would you be able to take a look and if you're happy launch a `@genie-omr build aarch64` Jenkins CI build @0xdaryl and/or @Leonardo2718?

Any feedback from @knn-k, @jeanlego, @marufunb and @hsoontie would also be appreciated.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -366,23 +366,41 @@ OMR::ARM64::TreeEvaluator::lxorEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
 	}
 
+// Evaluate iand, ior and ixor
+static inline TR::Register *
+evaluateBooleanTypeInteger(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg)
+   {
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Register *src1Reg = cg->gprClobberEvaluate(firstChild);
+   TR::Node *secondChild = node->getSecondChild();
+   TR::Register *src2Reg = cg->gprClobberEvaluate(secondChild);
```
0xdaryl: A clobber evaluation for both of these source registers is not what you want here.  A clobber evaluation is useful when the register itself will be modified by the instruction but yet it's original value is still needed (because of commoning, for example).  Clobber evaluation will insert an instruction to copy the register to a temporary register if its reference count is > 1 (i.e., its value is still consumed elsewhere) and skip the copy if this is the last reference it.

Simple calls to `evaluate` are what you want here.

However, there is an opportunity to optimize the target register though.  If one of the source registers "dies" at this instruction then you can re-use it for the target register rather than allocating a new register.  So, if the reference count on either source register == 1 then you can conclude that this is the last reference to that node and its register live range ends here and can be reused.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -366,23 +366,41 @@ OMR::ARM64::TreeEvaluator::lxorEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
 	}
 
+// Evaluate iand, ior and ixor
+static inline TR::Register *
+evaluateBooleanTypeInteger(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg)
```
0xdaryl: I find this name confusing.  How about `evaluateBitwiseLogical` instead to keep the terminology consistent throughout the compiler?

compiler/aarch64/codegen/BinaryEvaluator.cpp
aarongraham9: Ok fair enough. Agreed. I'll check the reference count at this point to see about re-use.

knn-k:
You can combine 32-bit evaluator and 64-bit evaluator code like I did in other evaluators.
See `icmpHelper()` in compiler/aarch64/codegen/ControlFlowEvaluator.cpp for example.

PR #3352, opened 2018-12-14
Add RefSlotHandle types to GC

rwy0717:
The RefSlotHandle is "A handle to a reference-slot in an object". The
RefSlotHandle is a default implementation of a SlotHandle for bare,
unencoded gc-references.

This PR also adds a CompressedRefSlotHandle, for working with compressed
references. The reference shift is dynamic.

In the future, these classes will be used by OMR client's object
scanners to give the GC read-write capabilities on objects.

### DIFF
--- a/gc/api/include/OMR/GC/CompressedRefSlotHandle.hpp
+++ b/gc/api/include/OMR/GC/CompressedRefSlotHandle.hpp
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ *  Copyright (c) 2018, 2019 IBM and others
+ *
+ *  This program and the accompanying materials are made available under
+ *  the terms of the Eclipse Public License 2.0 which accompanies this
+ *  distribution and is available at https://www.eclipse.org/legal/epl-2.0/
+ *  or the Apache License, Version 2.0 which accompanies this distribution and
+ *  is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ *  This Source Code may also be made available under the following
+ *  Secondary Licenses when the conditions for such availability set
+ *  forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
+ *  General Public License, version 2 with the GNU Classpath
+ *  Exception [1] and GNU General Public License, version 2 with the
+ *  OpenJDK Assembly Exception [2].
+ *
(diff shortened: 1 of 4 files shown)


gc/api/include/OMR/GC/CompressedRefSlotHandle.hpp
```diff
@@ -0,0 +1,65 @@
...
+	explicit CompressedRefSlotHandle(uint32_t* slot, uint32_t shift) : _slot(slot), _shift(shift) {}
+
+	void* toAddress() const { return _slot; }
+
+	omrobjectptr_t readReference() const { return decode(*_slot, _shift); }
+
+	void writeReference(omrobjectptr_t value) const { *_slot = encode(value, _shift); }
+
+private:
+	static omrobjectptr_t decode(uint32_t data, uint32_t shift)
```
charliegracie: Is it possible to modify the clang format for the GC such that this function would look like:
```
static omrobjectptr_t
decode(uint32_t data, uint32_t shift)
{
	return (omrobjectptr_t)(((uint64_t)data) << shift);
}
```

since that more closely matches the coding standards?

charliegracie: Note that single line functions are ok just if it is a multi line function it should follow the above format.

rwy0717: You know, it's actually not as simple as I would have thought. Putting the type onto it's own line stops clang-format from putting short functions on one line. I'll just manually fix it up.

rwy0717:
* Fixed formatting.
* Removed `#error` in normal refslot handle header, which was ensuring compressed-refs are disabled. It's not a valid test.

rwy0717:
This is ready for a build

charliegracie:
@genie-omr build all

charliegracie:
Just to make sure is the `explicit` keyword available on all of our compilers?

rwy0717:
I haven't actually tested it, but it's part of the original c++98 standard, so I assume so.

rwy0717:
* Updated comments, copyright
* Rebased
* Removed explicit from CompressedRefSlotHandle, since it has no affect. The RefSlotHandle is still explicit.

rwy0717:
This PR isn't needed anymore.

PR #3353, opened 2018-12-14
Add more type traits to OMR/TypeTraits.hpp to support JitBuilder and Tril

### LINKED ISSUE #3323: Add missing utilities to OMR/TypeTraits.hpp and reimplement toIlType and is_supported
Now that OMR has its own implementation of the `type_traits` library as `OMR/TypeTraits.hpp`, `TypeDictionary::toIlType` and `TypeDictionary::is_supported` should be reimplemented in terms of these new utilities instead of the standard library ones, which are not available on all supported build compilers. To do so, the following utilities will have to be implemented:

- [x] IsIntegral: replaces `std::is_integral`
- [x] IsFloatingPoint: replaces `std::is_floating_point`
- [x] IsArithmetic: replaces `std::is_arithmetic` (maybe find a better name?)
- [x] IsVoid: replaces `std::is_void`
- [x] IsPointer: replaces `std::is_pointer`
- [x] RemovePoineter: replaces `std::remove_pointer`
- [x] EnableIf: replaces `std::enable_if`

Leonardo2718:
This changeset adds the following type traits:

- IsIntegral: replaces `std::is_integral`
- IsFloatingPoint: replaces `std::is_floating_point`
- IsArithmetic: replaces `std::is_arithmetic`
- RemovePoineter: replaces `std::remove_pointer`

In addition, the JitBuilder `toIlType` and `is_supported` services and Tril implementation are refactored to make use of these newly implemented utilities, breaking their dependency on the C++11 `type_traits` standard library.

Closes #3323

### DIFF
--- a/fvtest/tril/tril/ast.hpp
+++ b/fvtest/tril/tril/ast.hpp
@@ -26,7 +26,6 @@
 
 #include "OMR/TypeTraits.hpp"
 
-#include <type_traits>
 #include <assert.h>
 #include <cstring>
 
@@ -41,8 +40,8 @@
  *
  * Type name     | Used for              | Compatible with (C++ types)
  * --------------|-----------------------|--------------------------------------
- * Integer       | integral values       | any type satisfying `std::is_integral`
- * FloatingPoint | floating point values | any type satisfying `std::is_floating_point`
+ * Integer       | integral values       | any type satisfying `OMR::IsIntegral`/`std::is_integral`
+ * FloatingPoint | floating point values | any type satisfying `OMR::IsFloatingPoint`/`std::is_floating_point`
  * String        | character strings     | `const char*`
  *
  * Names for these types are defined in the ASTType enum.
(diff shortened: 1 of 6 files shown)


include_core/OMR/TypeTraits.hpp
```diff
@@ -98,6 +98,17 @@ struct RemoveCvRef : RemoveCv<typename RemoveReference<T>::Type> {};
 // template <typename T>
 // struct RemoveCvRef<T&&> : RemoveCvRef<T> {};
 
+/// Remove one layer of non-cv-qualified pointers
+template <typename T>
+struct RemoveNonCvPointer : TypeAlias<T> {};
```
Leonardo2718: I use a different naming convention here than is used for `IsPointer` (i.e. `RemoveNonCvPointer` instead of `RemovePointerBase`). My thinking is that names are clearer if they describe what a type trait does rather than how it's used. What do others think?

fvtest/tril/tril/ast.hpp
```diff
@@ -41,8 +40,8 @@
  *
  * Type name     | Used for              | Compatible with (C++ types)
  * --------------|-----------------------|--------------------------------------
- * Integer       | integral values       | any type satisfying `std::is_integral`
- * FloatingPoint | floating point values | any type satisfying `std::is_floating_point`
+ * Integer       | integral values       | any type satisfying `std::IsIntegral`
```
rwy0717: Honestly, you should probably have both: `std::is_integral`/`OMR::IsIntegral`.

Leonardo2718:
Autotools build is failing because JitBuilder examples don't have `include_core` in their build path.

include_core/OMR/TypeTraits.hpp
rwy0717: Yeah, that makes sense to me. I guess we should update the others to match? Do you want to do it or should I?

rwy0717:
LGTM 👍

PR #3355, opened 2018-12-14
Multiple ddrgen fixes

keithc-ca:
* add OSX support for DW_TAG_ptr_to_member_type
* use port library consistently if DEBUG_PRINT_TYPES is defined
* add missing use of Type::isStandardType() in blob generator
* recognize more known types and handle them consistently
-- fix bug in Type::isStandardType()
-- e.g. 'unsigned short int' was handled, but not 'unsigned short'
-- treat char as unsigned unless explicitly 'signed'
-- handle SRP types; e.g. 'J9SRP(U_32)' -> 'J9SRP(U32)'


charliegracie:
@genie-omr build all

PR #3356, opened 2018-12-14
[WIP] Replace GCD semaphores with Mach semaphores on OSX

babsingh:
On OSX, GCD (Grand Central Dispatch) semaphores cause issues such as
hangs and crashes:
1. thrstatetest failures: https://github.com/eclipse/openj9/issues/3822
2. vmRuntimeState failures:
https://github.com/eclipse/openj9/issues/3878

Also, GCD semaphores are not signal-safe.

Mach semaphores resolve the hangs and crashes seen with GCD semaphores. 

POSIX semaphores are signal safe. POSIX semaphores in OSX are
implemented using Mach semaphores. So, it is safe to assume that Mach
semaphores are signal safe.

Thus, GCD semaphores are replaced with Mach semaphores on OSX.


babsingh:
Removed WIP. @charliegracie can you please review this PR?
fyi - @DanHeidinga @pshipton

Leonardo2718:
@genie-omr build osx

charliegracie:
@genie-omr build all

charliegracie:
the zOS failure is due to no machines being available.

charliegracie:
@babsingh you provided a link to the Swift semaphore.c but I do not see any `mach` APIs being called from there.  Can you explain why you provided that link?

babsingh:
> you provided a link to the Swift semaphore.c but I do not see any mach APIs being called from there. Can you explain why you provided that link?

GCD semaphores are built on top of Mach semaphores. Digging slightly deeper into the source code reveals the relationship between GCD and Mach semaphores. Example: [dispatch_semaphore_signal](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/semaphore.c#L104) -> ... -> [_dispatch_sema4_signal](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/shims/lock.c#L136) -> [semaphore_signal](https://github.com/st3fan/osx-10.9/blob/master/xnu-2422.1.72/libsyscall/mach/semaphore.c#L37) [mach API].

> why OMR should switch to Mach semaphores?

- We should use Mach semaphores for signal safety. There are discussions where it is argued that [GCD semaphores might be unsafe when used in signal handlers](https://github.com/adrienverge/openfortivpn/issues/105). 
- GCD semaphores are built on top of Mach semaphores. GCD semaphores add an extra layer of overhead on top of Mach semaphores.
- GCD semaphores are the cause of a lot of failures in OpenJ9's **thrstatetest** and **vmRuntimeState** test suites, which test the OMR thread library. Mach semaphores resolve the failures seen with GCD semaphores.
- The implementation of GCD semaphores has a fast-path and a slow-path. This suggests some form of JIT compilation. Not sure if we want this feature in the presence of our own JIT.

babsingh:
ping @charliegracie.

charliegracie:
I guess the piece I am missing is what is broken with the current implementation.  I understand that there are some downstream projects that are failing tests but I am not sure we understand why they are failing.  If the GCD semaphores are built on top of Mach semaphores then I would expect the GCD semaphores to work properly.  Do we actually know what the issue is why GCD semaphores? Can tests be introduced into OMR that show the problem? If there are noticeable differences in the implementations that use GCD and Mach semaphores will this PR break OMR users that are relying on the current behaviour?

Your comment about GCD semaphores having a fast-path and a slow-path implying there is some sort of JIT does not make sense to me. Lots of code has fast and slow paths without having JITs compiling code for the fast path.  In fact the OMR GC has lots of fast and slow path implementations without having a JIT.

babsingh:
@charliegracie 
> what is broken with the current implementation? Do we actually know what the issue is with GCD semaphores?

Here is a stack trace from an [OpenJ9 failure](https://github.com/eclipse/openj9/issues/4046#issuecomment-447506420):
```
  * frame #0: 0x00007fff97484286 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff952a59f9 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff944fa9b3 libsystem_c.dylib`abort + 129
    frame #3: 0x0000000008467868 libj9prt29.dylib`omrdump_create(portLibrary=<unavailable>, filename=<unavailable>, dumpType=<unavailable>, userData=<unavailable>) + 280 at omrosdump.c:191
    frame #4: 0x00000000084b0e49 libj9dmp29.dylib`doSystemDump(agent=0x00007f9d51702a70, label="/Users/jenkins/babsing_workspace/openj9-openjdk-jdk8/build/macosx-x86_64-normal-server-release/vm/", context=<unavailable>) + 297 at dmpagent.c:718
    frame #5: 0x00000000084b5955 libj9dmp29.dylib`protectedDumpFunction(portLibrary=<unavailable>, userData=<unavailable>) + 21 at dmpagent.c:2869
    frame #6: 0x0000000008468be7 libj9prt29.dylib`omrsig_protect(portLibrary=0x0000000008185c58, fn=(libj9dmp29.dylib`protectedDumpFunction at dmpagent.c:2866), fn_arg=0x00007fff57bda560, handler=(libj9dmp29.dylib`signalHandler at dmpagent.c:2881), handler_arg=0x0000000000000000, flags=125, result=0x00007fff57bda578) + 983 at omrsignal.c:384
    frame #7: 0x00000000084b5556 libj9dmp29.dylib`runDumpAgent [inlined] runDumpFunction(agent=0x00007f9d51702a70, label="/Users/jenkins/babsing_workspace/openj9-openjdk-jdk8/build/macosx-x86_64-normal-server-release/vm/", context=0x00007fff57bdaa30) + 121 at dmpagent.c:2843
    frame #8: 0x00000000084b54dd libj9dmp29.dylib`runDumpAgent(vm=0x00007f9d52802e20, agent=0x00007f9d51702a70, context=0x00007fff57bdaa30, state=0x00007fff57bdaa68, detail="", timeNow=<unavailable>) + 669 at dmpagent.c:2769
    frame #9: 0x00000000084cb3d1 libj9dmp29.dylib`triggerDumpAgents(vm=0x00007f9d52802e20, self=<unavailable>, eventFlags=131072, eventData=<unavailable>) + 1393 at trigger.c:1028
    frame #10: 0x00000000084b9367 libj9dmp29.dylib`abortHandler(sig=<unavailable>) + 327 at dmpsup.c:245
    frame #11: 0x00007fff9b7e9f1a libsystem_platform.dylib`_sigtramp + 26
    frame #12: 0x00007fff97484287 libsystem_kernel.dylib`__pthread_kill + 11
    frame #13: 0x00007fff952a59f9 libsystem_pthread.dylib`pthread_kill + 90
    frame #14: 0x00007fff944fa9b3 libsystem_c.dylib`abort + 129
    frame #15: 0x000000000846a012 libj9prt29.dylib`masterSynchSignalHandler(signal=<unavailable>, sigInfo=0x00007fff57bdb830, contextInfo=0x00007fff57bdb898) + 738 at omrsignal.c:1143
    frame #16: 0x00007fff9b7e9f1a libsystem_platform.dylib`_sigtramp + 26
    frame #17: 0x00007fff9b7c3613 libdispatch.dylib`_dispatch_semaphore_signal_slow + 123
    frame #18: 0x000000000803877c thrstatetest`test_cleanup(env=0x0000000008b6f700, ignoreExpectedFailures=<unavailable>) + 140 at testsetup.c:445
    frame #19: 0x000000000802500a thrstatetest`signalProtectedMain [inlined] runTest(ignoreExpectedFailures=1) + 49 at runtest.c:441
    frame #20: 0x0000000008024fd9 thrstatetest`signalProtectedMain [inlined] runAllTests(env=0x0000000008b6f700) + 69 at runtest.c:473
    frame #21: 0x0000000008024f94 thrstatetest`signalProtectedMain(portLibrary=<unavailable>, arg=<unavailable>) + 916 at runtest.c:629
    frame #22: 0x0000000008468be7 libj9prt29.dylib`omrsig_protect(portLibrary=<unavailable>, fn=<unavailable>, fn_arg=<unavailable>, handler=<unavailable>, handler_arg=<unavailable>, flags=<unavailable>, result=<unavailable>) + 983 at omrsignal.c:384
    frame #23: 0x0000000008024a63 thrstatetest`main(argc=<unavailable>, argv=<unavailable>, envp=<unavailable>) + 323 at cmain.c:99
    frame #24: 0x00007fff939005c9 libdyld.dylib`start + 1
```
SIGABRT is generated from `_dispatch_semaphore_signal_slow` (frame 17). The failures result in a hang. Also, failures are intermittent. The hangs and the intermittent nature of failures prevent me to study the GCD semaphore behavior during the failure. 

I have found another project (openfortivpn) which encountered issues with similar symptoms when using GCD semaphores. The issues were resolved after replacing GCD semaphores with Mach semaphores. References:
- Issue 1 - Locking issue resulting in a hang: https://github.com/adrienverge/openfortivpn/pull/132
- Issue 2 - GCD semaphores unsafe for signal handlers: https://github.com/adrienverge/openfortivpn/issues/105
- Fix for Issue 1 and Issue 2 - replace GCD semaphores with Mach semaphores: https://github.com/adrienverge/openfortivpn/pull/135.
- Summary of Issue 1, Issue 2 and Fix: https://github.com/adrienverge/openfortivpn/pull/135#issuecomment-307048969.

The above references support the decision to replace GCD semaphores with Mach semaphores in OMR.

> If the GCD semaphores are built on top of Mach semaphores then I would expect the GCD semaphores to work properly. 

Please refer to the openfortivpn case mentioned above where GCD semaphores didn't work but Mach semaphores worked.

> Can tests be introduced into OMR that show the problem?

For OMR, tests can be derived from the failing OpenJ9 testcases in the long run. I will create an OMR issue to add tests for reproducing the problem seen with GCD semaphores.

Note: Currently, there are no tests for the j9sem API since it just wraps calls to OS specific semaphore implementations.

> If there are noticeable differences in the implementations that use GCD and Mach semaphores will this PR break OMR users that are relying on the current behavior?

OMR's semaphore (j9sem) API should behave the same with Mach semaphores. No OMR users would be negatively impacted.

charliegracie:
Sorry @babsingh but to me that looks like maybe there is some bad code in the test since a call to dispatch is crashing. Do you know what the code is doing that causes the crash?  When I map the line number from your stack trace to code it does not appear to be anything to do with semaphores so clearly our code does not match up. Is it a call to `j9sem_post`?

If this is an intermittent failure what was the failure rate for the GCD semaphores? How many tests have you run with the mach semaphores?

charliegracie:
@genie-omr build zos

babsingh:
@charliegracie 

> Do you know what the code is doing that causes the crash? When I map the line number from your stack trace to code it does not appear to be anything to do with semaphores so clearly our code does not match up. Is it a call to j9sem_post?

The abort signal is generated from `libdispatch.dylib::_dispatch_semaphore_signal_slow`, which is [j9sem_post](https://github.com/eclipse/omr/blob/master/include_core/unix/thrdsup.h#L390-L391). `j9sem_post` is invoked in `thrstatetest` from here: [thrstate/testsetup.c#L442](https://github.com/eclipse/openj9/blob/master/runtime/tests/thread/thrstate/testsetup.c#L442).

```
#define SEM_POST(smP)	\
(dispatch_semaphore_signal((dispatch_semaphore_t)(smP)))
```

> If this is an intermittent failure what was the failure rate for the GCD semaphores? How many tests have you run with the mach semaphores?

The `thrstatetest` suite fails every time with GCD semaphores since multiple test cases fail in the `thrstatetest` suite. A test case may fail 80% of the time. Also, the test behavior varies every time a failure re-occurs. https://github.com/eclipse/openj9/issues/3878 contains the list of all failing test cases in the `thrstatetest` suite. The source code for the `thrstatetest` suite is here: [LINK](https://github.com/eclipse/openj9/tree/master/runtime/tests/thread/thrstate).


The `thrstatetest` suite passes with Mach semaphores every time. In addition, the entire OpenJ9 sanity test suite has been run with Mach semaphores. The following pull request contains links to the test jobs: https://github.com/eclipse/openj9/pull/4149. No failures related to Mach semaphores are seen in these test jobs.

> Also since OpenJ9 does funky things with recompiling the port/thread libraries and uses its own defines are you sure the thread / port defines which may impact this code are consistent between the two projects?

OpenJ9 uses the `j9sem` API from here: [thread/common/j9sem.c](https://github.com/eclipse/omr/blob/master/thread/common/j9sem.c). `j9sem` is consistent between OpenJ9 and OMR. OpenJ9 doesn't do any funky things when building `j9sem`.

**More details about the issue:**

`thrstatetest` is a suite with multiple test cases. The testcases in the `thrstatetest` suite set fields in `J9VMThread` (`setVMThreadState`), `J9Thread` (`setJ9ThreadState`) and `omrthread_monitor_t` (`setJ9MonitorState`) structures: [thrstate/testsetup.c](https://github.com/eclipse/openj9/blob/master/runtime/tests/thread/thrstate/testsetup.c). Then, OpenJ9 functions, such as `getVMThreadObjectState` and `getVMThreadRawState`, are used to verify the thread state. These functions are defined in [runtime/util/thrinfo.c](https://github.com/eclipse/openj9/blob/master/runtime/util/thrinfo.c). These functions extract information from `J9VMThread`, `J9Thread` and `omrthread_monitor_t` structures. 


I looked into a couple of failures in detail. An example of actual vs. expected behavior is shown below:
```
getVMThreadRawState failed:
	actual: vmstate 2 lockObj 00000000FFE3A420 rawLock 00007FC8350A1040 lockOwner 000000000AFB8200 count 1
	expect: vmstate 2 lockObj 00000000FFE3A420 rawLock 00007FC8350A1040 lockOwner 000000000AFB4700 count 1
```
For the above failure, the `lockOwner` doesn't match between the actual and expected results. In other failing testcases, I observed that `vmstate/rawLock/count` doesn't match the actual and expected results. When I reproduce the failures, the testcases fail differently such that `lockOwner/vmstate/rawLock/count` may not have the same values seen in the previous failure. In some reruns, a failure may not even reoccur. These are symptoms of memory/data corruption.

> [revisiting] what is broken with the current implementation? 

After some reading, I found that the GCD and POSIX pthreads are two different approaches for concurrency: https://www.objc.io/issues/2-concurrency/concurrency-apis-and-pitfalls/. GCD and POSIX pthreads are replacements for each other. They are not supposed to be used together. On OSX, the OMR thread library uses POSIX structures for mutexes, condvars and threads. But, it uses the GCD library for semaphores. The usage of POSIX and GCD libraries together causes the undefined behavior described above. The POSIX named semaphores use Mach semaphores. So, the POSIX library works fine with Mach semaphores. This is why replacing GCD semaphores with Mach semaphores resolves the above-mentioned failures.

PR #3357, opened 2018-12-17
Tril tests for iucmpeq, lucmpeq, etc.

knn-k:
This commit adds tests for unsigned integer comparison, such as
iucmpeq and lucmpgt.

### DIFF
--- a/fvtest/compilertriltest/CompareTest.cpp
+++ b/fvtest/compilertriltest/CompareTest.cpp
@@ -52,7 +52,14 @@ TEST_P(Int32Compare, UsingConst) {
     auto param = TRTest::to_struct(GetParam());
 
     char inputTrees[120] = {0};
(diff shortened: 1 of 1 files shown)


fvtest/compilertriltest/CompareTest.cpp
```diff
@@ -52,11 +52,16 @@ TEST_P(Int32Compare, UsingConst) {
...
+       "(method return=Int32 "
+         "(block "
+           "(ireturn "
+             "(%s "
+               "(iconst %d) "
+               "(iconst %d)))))",
+       param.opcode.c_str(), param.lhs, param.rhs);
     auto trees = parseString(inputTrees);
 
-    ASSERT_NOTNULL(trees);
```
Leonardo2718: Why is this being removed? I think it's still needed.

knn-k: That was not an intended change.  Updated the file.

Leonardo2718:
@genie-omr build all

PR #3359, opened 2018-12-17
AArch64: Remove redundant call to decFutureUseCount()

knn-k:
This commit fixes TR::ARM64MemSrc1Instruction::assignRegisters().
It calls decFutureUseCount() on sourceVirtual, but that is not
needed because assignOneRegister() has already done it.


knn-k:
This is my mistake in #2919.  See its review comment.  I had to remove the block starting with `if (sourceVirtual->decFutureUseCount() == 0)` when I rewrote the code to use `assignOneRegister()`.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3368, opened 2018-12-18
Only check doNotProfile flag under voluntaryOSR

cathyzhyi:
The flag only needs to be checked under voluntaryOSR. Yield
points in involuntaryOSR are always OSRPoints and there is always OSR
support for each yield point under this mode.

### DIFF
--- a/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
+++ b/compiler/il/symbol/OMRResolvedMethodSymbol.cpp
@@ -1636,8 +1636,12 @@ OMR::ResolvedMethodSymbol::cannotAttemptOSRDuring(int32_t callSite, TR::Compilat
             break;
             }
 
-         // Check the caller existed during ILGen
-         if (callSiteInfo._byteCodeInfo.doNotProfile())
+         // In voluntaryOSR mode, check the caller existed during ILGen because
+         // there is no OSR support for call nodes created outside of ILGen.
+         //
+         // In involuntaryOSR node, yield points are always OSR points and there should be OSR
+         // support for very yield point under this mode.
+         if (callSiteInfo._byteCodeInfo.doNotProfile() && comp->getOSRMode() == TR::voluntaryOSR)
             {
             if (comp->getOption(TR_TraceOSR))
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

compiler/il/symbol/OMRResolvedMethodSymbol.cpp
```diff
@@ -1637,7 +1637,7 @@ OMR::ResolvedMethodSymbol::cannotAttemptOSRDuring(int32_t callSite, TR::Compilat
             }
 
          // Check the caller existed during ILGen
-         if (callSiteInfo._byteCodeInfo.doNotProfile())
+         if (callSiteInfo._byteCodeInfo.doNotProfile() && comp->getOSRMode() == TR::voluntaryOSR)
```
fjeremic: Perhaps I may not be as experienced in this part of the codebase but the check added here at least to me seems quite non-obvious. I think it should deserve a comment on why it is there.

andrewcraik: to those versed in the area it does make sense, but a comment is fine - basically you only have a choice about OSR when you are in voluntaryOSR - under involuntary the semantics of the node regardless of this flag determine if it is an OSR point or not.

cathyzhyi: Added more comments.

andrewcraik:
Change looks good to me

Leonardo2718:
@genie-omr build all

vijaysun-omr:
@cathyzhyi and/or @andrewcraik could'nt whatever "problem" occurred in involuntary OSR (since that's the only mode that was affected by this change) also occur in voluntary OSR ? Maybe that's where the fact that it's "voluntary" OSR comes in and the fall back back is a safe one that does not impact functional correctness, whereas with "involuntary" OSR there is no fall back.

andrewcraik:
right so under voluntary OSR we expect that points may not be able to OSR and the code has to cope with that. Under involuntary OSR we have to treat all potential points as OSR points since we don't control them - answering that a point which can OSR does not support it under involuntary OSR is a paradox and that is the reason this change 'works'.

PR #3371, opened 2018-12-18
Migrate CodeCache switching logic to CodeGenerator

0xdaryl:
* Move NumReservedIPICTrampolines getter/setters to X86 CodeGenerator
* Move codeCacheSwitched getter/setter to CodeGenerator
* Implement switchCodeCacheTo() in CodeGenerator

### DIFF
--- a/compiler/compile/OMRCompilation.cpp
+++ b/compiler/compile/OMRCompilation.cpp
@@ -2026,7 +2026,7 @@ void OMR::Compilation::switchCodeCache(TR::CodeCache *newCodeCache)
    {
    TR_ASSERT( self()->getCurrentCodeCache() != newCodeCache, "Attempting to switch to the currently held code cache");
    self()->setCurrentCodeCache(newCodeCache);  // Even if we signal, we need to update the reserved code cache for recompilations.
-   _codeCacheSwitched = true;
+   self()->cg()->setCodeCacheSwitched(true);
 
 #ifdef TR_TARGET_X86
    self()->cg()->setNumReservedIPICTrampolines(0);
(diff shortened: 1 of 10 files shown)


0xdaryl:
Note to reviewers:

This PR is written in such a way as to prevent breakages in downstream projects.  Once this PR is merged and the necessary changes made to known downstream projects then the (hopefully obvious) scaffolding will be removed in OMR.

One of the by-products of the scaffolding left in place for now are the `TR_TARGET_X86` macros that guard calls to CodeGenerator `setNumReservedIPICTrampolines` and `getNumReservedIPICTrampolines` in some places.  The correct way to fix this would be to provide X86-specialized classes but in the case of the Compilation class there isn't already an X86 specialization.  This means that by adding one would cause a downstream breakage anyway because its build system would need to be modified to be aware of the new file.  Since that isn't in the spirit of the scaffolding, I simply guarded the calls with the target macro to avoid introducing a new file.  This code should be extremely transient.

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build all

compiler/compile/OMRCompilation.cpp
```diff
@@ -2739,3 +2742,29 @@ bool OMR::Compilation::incompleteOptimizerSupportForReadWriteBarriers()
...
+   return self()->cg()->getNumReservedIPICTrampolines();
+#else
+   return 0;
+#endif
+   }
+
+void OMR::Compilation::setNumReservedIPICTrampolines(int32_t n)
+   {
+#ifdef TR_TARGET_X86
+   return self()->cg()->setNumReservedIPICTrampolines(n);
```
Leonardo2718: I guess this technically works, but given that the function returns void, would it make more sense to skip the `return`?

```suggestion
   self()->cg()->setNumReservedIPICTrampolines(n);
```

0xdaryl:
@genie-omr build all

PR #3378, opened 2018-12-19
AArch64: Add VFP instructions for float/double

knn-k:
This commit adds aarch64 VFP instructions for float/double scalar
operations.


0xdaryl:
@genie-omr build aarch64

PR #3379, opened 2018-12-19
Revert "Multiple ddrgen fixes"

DanHeidinga:
Reverts eclipse/omr#3355

An OpenJ9 contributor created the initial PR to address some issues with how ddrgen understands types.  Unfortunately, the changes introduced some issues in our builds when OpenJ9 upgraded to the  OMR version with the changes.

We haven't gotten to the bottom of the why the changes are incorrect yet but they seem to trigger a corner case resulting in incorrect behaviour.  See example below.

I'm requesting that these changes be reverted as they've introduced functional issues that affect OpenJ9's ability to take new OMR levels and may cause issues for any other ddrgen users.

We'll follow up on getting the changes fixed and re-delivered in the new year.

Example problem:
```
 [OUT] > Problem running command:
 [OUT] U64 contains value larger than Long.MAX_VALUE
 [OUT] > 
```
https://ci.eclipse.org/openj9/job/Test-sanity.functional-JDK8-linux_390-64_cmprssptrs/527/

### DIFF
--- a/ddr/include/ddr/blobgen/java/genSuperset.hpp
+++ b/ddr/include/ddr/blobgen/java/genSuperset.hpp
@@ -23,6 +23,11 @@
 #define GENSUPERSET_HPP
 
 #include "ddr/blobgen/genBlob.hpp"
+#include "ddr/std/unordered_map.hpp"
+
+#include <set>
+
+using std::set;
 
 class Field;
 class SupersetFieldVisitor;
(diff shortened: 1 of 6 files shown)


DanHeidinga:
@charliegracie @fjeremic @mstoodle as OMR committers, can one of you review and hopefully merge this?

Leonardo2718:
Thanks @DanHeidinga for providing a fix.

It looks like you didn't sign-off in the commit message so the ip validation is failing. Could you please correct that?

Also, are there any plans to revisit the changes from #3355 to find a better solution? If yes, could you open an issue so we can keep track of that work, maybe linking to a related OpenJ9 issue?

Leonardo2718:
@genie-omr build all

DanHeidinga:
> It looks like you didn't sign-off in the commit message so the ip validation is failing. Could you please correct that?

I used the webui's revert button.  I'm not aware of a way to add the appropriate signed-off by lines when doing a revert like this.  Historically, we've ignored the ip-validation failure in this case.

> Also, are there any plans to revisit the changes from #3355 to find a better solution? If yes, could you open an issue so we can keep track of that work, maybe linking to a related OpenJ9 issue?

Yes, we definitely want to fix the issues exposed here.  I've opened https://github.com/eclipse/openj9/issues/4114 to track the work to get these changes fixed and re-delivered to OMR

Leonardo2718:
AppVeyor build ssems to have failed because of infrastructure issues. Restarting.

DanHeidinga:
Thanks everyone!

PR #3380, opened 2018-12-19
Clarify meaning of return value in omrsysinfo_is_running_in_container

DanHeidinga:
Questions were raised on whether the result of the function could be
trusted when an error was encountered.  This change clarifies the
functions contract to ensure that FALSE is always returned if an
error occurs.

Checked all three places this function is defined:
* omr/port/common/omrsysinfo.c
* omr/port/win32/omrsysinfo.c (empty implemention, no comment)
* omr/port/unix/omrsysinfo.c


Leonardo2718:
Since this a documentation-only change, no PR builds are needed. So, I will cancel the AppVeyor and Travis CI builds to save resources.

PR #3381, opened 2018-12-19
Add omrthread_numa_get_current_node()

dnakamura:
Function returns the numa node number the thread is currently executing
on

### DIFF
--- a/include_core/thread_api.h
+++ b/include_core/thread_api.h
@@ -1411,6 +1411,14 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
 void
 omrthread_numa_set_enabled(BOOLEAN enabled);
 
+
+/**
+ * Gets the NUMA node the caller is being executed on
+ * @return uintptr_t
+ */
+uintptr_t
+omrthread_numa_get_current_node();
+
 /* -------------- rasthrsup.c ------------------- */
 /**
  * @brief
--- a/thread/linux/omrthreadnuma.c
+++ b/thread/linux/omrthreadnuma.c
@@ -35,6 +35,7 @@
 #include <inttypes.h>
 #include <sched.h>
 #include <sys/stat.h>
+#include <sys/syscall.h>
 #include <stdio.h>
 
 #include "omrcfg.h"
@@ -546,3 +547,17 @@ dumpNumaInfo() {
 
 }
 #endif
+
+
+uintptr_t
+omrthread_numa_get_current_node(){
+    unsigned node = 0;
+#if OMR_PORT_NUMA_SUPPORT
+    if (0 == syscall(SYS_getcpu, NULL, &node, NULL)) {
+        ++node;
+    } else {
+        node = 0;
+    }
+#endif
+    return node;
+}
--- a/thread/linux/omrthreadnuma.c
+++ b/thread/linux/omrthreadnuma.c
@@ -548,9 +548,9 @@ dumpNumaInfo() {
 }
 #endif
 
-
 uintptr_t
-omrthread_numa_get_current_node(){
+omrthread_numa_get_current_node()
+{
     unsigned node = 0;
 #if OMR_PORT_NUMA_SUPPORT
     if (0 == syscall(SYS_getcpu, NULL, &node, NULL)) {
--- a/thread/win32/omrthreadnuma.c
+++ b/thread/win32/omrthreadnuma.c
@@ -148,3 +148,17 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
 	*nodeCount = 0;
 	return J9THREAD_NUMA_ERR_AFFINITY_NOT_SUPPORTED;
 }
+
+uintptr_t
+omrthread_numa_get_current_node(){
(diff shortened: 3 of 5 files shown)


DanHeidinga:
Related to https://github.com/eclipse/openj9/issues/2813

thread/linux/omrthreadnuma.c
```diff
@@ -546,3 +547,17 @@ dumpNumaInfo() {
...
 }
 #endif
+
+
+uintptr_t
+omrthread_numa_get_current_node(){
+    unsigned node = 0;
+#if OMR_PORT_NUMA_SUPPORT
+    if (0 == syscall(SYS_getcpu, NULL, &node, NULL)) {
+        ++node;
```
DanHeidinga: This feels like there should be a comment here.  Is the numbering schedule 0 based vs 1 based elsewhere?

thread/win32/omrthreadnuma.c
```diff
@@ -148,3 +148,16 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
 	*nodeCount = 0;
 	return J9THREAD_NUMA_ERR_AFFINITY_NOT_SUPPORTED;
 }
+
+uintptr_t
+omrthread_numa_get_current_node(){
+    USHORT node = 0;
+
+    PROCESSOR_NUMBER processorNumber;
```
DanHeidinga: tab vs space

thread/win32/omrthreadnuma.c
```diff
@@ -148,3 +148,16 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
...
+	GetCurrentProcessorNumberEx(&processorNumber);
+	if (GetNumaProcessorNodeEx(&processorNumber, &node)) {
+		return node + 1;
```
DanHeidinga: same comment needed here as above

include_core/thread_api.h
```diff
@@ -1411,6 +1411,13 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
 void
 omrthread_numa_set_enabled(BOOLEAN enabled);
 
+
+/**
+ * Gets the NUMA node the caller is being executed on
```
DanHeidinga: Can you add the `  * @return uintptr_t ` stanza to this comment?

thread/win32/omrthreadnuma.c
```diff
@@ -148,3 +148,16 @@ omrthread_numa_get_node_affinity(omrthread_t thread, uintptr_t *numaNodes, uintp
...
+	}  else {
+		return 0;
```
DanHeidinga: Should be possible to do this with a single return statement similar to the linux version

charliegracie:
The Windows compile failure looks real.

charliegracie:
@genie-omr build win

dnakamura:
yeah I believe the windows compile failure is just a matter of a missing include file. Hunting it down now

dnakamura:
After some digging it seems the `GetCurrentProcessorNumberEx` function was introduced in win. And we currently target windows xp (`057c2f9ba8`). @samolisov Is there a reason you put back the preprocessor defines to target win XP?

samolisov:
@dnakamura we had a conversation with @youngar related to PR #2434 Please have a look at the comment https://github.com/eclipse/omr/pull/2434#issuecomment-381816845

dnakamura:
Opened PR to update minimum version to win 7 (#3445) since AFAIK it is the minimum supported windows  version

charliegracie:
Both of the Windows builds failed.  Any thoughts @dnakamura ?

dnakamura:
@charliegracie I believe the windows CI builds ran before #3445 landed.

Also I have added the aix support code, removing WIP

include_core/thread_api.h
DanHeidinga: @dnakamura Can you address this review comment?

thread/linux/omrthreadnuma.c
DanHeidinga: @dnakamura was this comment addressed somewhere else?

DanHeidinga:
@dnakamura I took a look at this again and I don't think the review comments were addressed.  Can you update the change and I can take another look at it?

dnakamura:
@DanHeidinga  should be resolved

DanHeidinga:
@dnakamura I'm probably missing something obvious but it's not clear to me why the node is being incremented before returning it.  ie:
```c
    if (0 == syscall(SYS_getcpu, NULL, &node, NULL)) {
         ++node;
```
Can you clarify that?  (preferably with the comment in the code so we don't lose the explanation)

dnakamura:
@DanHeidinga  The OS numbers numa nodes starting from 0, but in OMR, we use a value of 0 to indicate no associated numa node

pshipton:
@charliegracie is there something else that needs to be addressed before this is ready to merge?

charliegracie:
@genie-omr build all

rwy0717:
@genie-omr build aarch64
rebuilding due to infrastructure issues

PR #3382, opened 2018-12-19
Add shouldSetCPUAffinity field to NUMAManager

dnakamura:
Field idicates if gc should be modifying the affinity of mutator threads

### DIFF
--- a/gc/base/NUMAManager.hpp
+++ b/gc/base/NUMAManager.hpp
@@ -164,5 +164,5 @@ public:
 	{}
 };
 
-#endif /* OMR_GC_VLHGC */
+#endif /* NUMAMANAGER_HPP_ */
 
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Related to https://github.com/eclipse/openj9/issues/2813

gc/base/NUMAManager.hpp
```diff
@@ -146,6 +152,7 @@ class MM_NUMAManager
 	 */
 	MM_NUMAManager()
 		: _physicalNumaEnabled(false)
+		, _shouldSetCPUAffinity(true)
```
DanHeidinga: This changes the behaviour from default setAffinity, to not setting affinity.  Intentional?

charliegracie: The NUMAManager is only used for balanced where I think the GC wants to be setting affinity by default.  I think this change is good but I need to review the OpenJ9 change that is associated with this.

dnakamura: @DanHeidinga  `_shouldSetCPUAffinity` is a new field which has been added. We default the value to true because old behavior was to always allow setting the cpu affinity (if numa is available and enabled)

PR #3385, opened 2018-12-20
Tril tests for f2i, f2d, etc.

knn-k:
This commit adds Tril test cases for FP->integer and FP->FP conversion.

### DIFF
--- a/fvtest/compilertriltest/TypeConversionTest.cpp
+++ b/fvtest/compilertriltest/TypeConversionTest.cpp
@@ -908,3 +908,395 @@ INSTANTIATE_TEST_CASE_P(TypeConversionTest, Int64ToDouble, ::testing::Combine(
     ::testing::Values(
         std::make_tuple("l2d", l2d) )));
 #endif
+
+template <typename F, typename I>
+bool fp_filter(F a)
+   {
+   // workaround: avoid failure caused by snprintf("%f") or potential undefined behaviour
+   return (abs(a) < 0.01 && a != 0.0) ||
+      (a > static_cast<F>(std::numeric_limits<I>::max())) ||
+      (a < static_cast<F>(std::numeric_limits<I>::min())) ;
+   }
+
+int32_t f2i(float x) {
+    return static_cast<int32_t>(x);
+}
+
+int64_t f2l(float x) {
+    return static_cast<int64_t>(x);
+}
+
+int32_t d2i(double x) {
+    return static_cast<int32_t>(x);
+}
+
+int64_t d2l(double x) {
+    return static_cast<int64_t>(x);
+}
+
+class FloatToInt32 : public TRTest::UnaryOpTest<int32_t,float> {};
+
+TEST_P(FloatToInt32, UsingConst) {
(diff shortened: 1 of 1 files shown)


fvtest/compilertriltest/TypeConversionTest.cpp
```diff
@@ -908,3 +908,386 @@ INSTANTIATE_TEST_CASE_P(TypeConversionTest, Int64ToDouble, ::testing::Combine(
...
+
+    auto entry_point = compiler.getEntryPoint<int32_t (*)(float)>();
+    volatile auto exp = param.oracle(param.value);
+    volatile auto act = entry_point(param.value);
+    ASSERT_EQ(exp, act);
+}
+
+INSTANTIATE_TEST_CASE_P(TypeConversionTest, FloatToInt32, ::testing::Combine(
+    ::testing::ValuesIn(
+        TRTest::filter(TRTest::const_values<float>(), smallFp_filter<float>)),
```
Leonardo2718: Although I agree it makes sense to use a template function for `smallFp_filter`, I believe some of our older build compiler's don't handle converting template function instantiations to function pointers correctly. So, a separate function needs to be defined for each type instead :disappointed: . (This is also the reason template functions are not used for the oracle functions)

knn-k: Of course I can change `smallFp_filter()` not to use template.
But fvtest/compilertriltest/ArithmeticTest.cpp already uses a filter function with template:
```
        TRTest::filter(TRTest::const_value_pairs<int32_t, int32_t>(), div_filter<int32_t> )),
```
Does the `div_filter()` cause problems anywhere?

Leonardo2718: Hm... so it does. Ok, I'll accept the changes as they are then.

I'll need to take a closer look at what was not working before. If this does turn out to be a problem, I will fix it when there's concrete evidence.

Leonardo2718:
@genie-omr build all

Leonardo2718:
[sigh]

The OSX failure is due to the fact that converting a floating-point value to an integer type that is too narrow to hold the value is undefined behaviour.

As a fix, I recommend changing `smallFp_filter` to something like:
```c++
#include <limits>

template <typename F, typename I>
bool fp_filter(F a)
   {
   // avoid failure caused by snprintf("%f") or potential undefined behaviour
   return (abs(a) < 0.01) || 
      (a > static_cast<F>(std::numeric_limits<I>::max())) || 
      (a < static_cast<F>(std::numeric_limits<I>::min())) ;
   }

/* ... */

INSTANTIATE_TEST_CASE_P(TypeConversionTest, FloatToInt32, ::testing::Combine(
    ::testing::ValuesIn(
        TRTest::filter(TRTest::const_values<float>(), fp_filter<float, int32_t>)),
    ::testing::Values(
        std::make_tuple("f2i", f2i) )));
```

I _think_ that should do it.

knn-k:
Updated the file reflecting the comment.

Leonardo2718:
@genie-omr build all

PR #3386, opened 2018-12-20
AArch64: Improve code generation of lcmpeq, lucmpeq, etc.

knn-k:
This commit improves code generation of comparison evaluators of
64-bit integers.
Their logic is commoned with 32-bit integer comparison. As its
result, it uses subsimmx instruction when the second child is a
small constant (less than 4096), and that saves one instruction.


0xdaryl:
@genie-omr build aarch64

0xdaryl:

Issue #3366, opened 2018-12-18
Calling convention violation on X86_64/Windows


samolisov:
Component: **comptest** (`fvtest/compilertriltest/LinkageTest.cpp`)

The `TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg)` test fails (just stops execution) when OMR are compiled with MSVC and passes when OMR are compiled with Clang. The single parameter test passed as it's expected.

The last string in the test log:

```
[----------] 3 tests from LinkageTest/0, where TypeParam = <type>
```

---

### Some debug information

**Invoke the `fourthArg` function from a JITed by OMR one**

**MSVC**

Unit test:
```asm
00007ff7`fbfaa653 41b900040000    mov     r9d,400h
00007ff7`fbfaa659 4533c0          xor     r8d,r8d
00007ff7`fbfaa65c 33d2            xor     edx,edx
00007ff7`fbfaa65e 33c9            xor     ecx,ecx
00007ff7`fbfaa660 ff942468030000  call    qword ptr [rsp+368h]
```

JITed function:

```asm
0000019c`d6d30034 4153            push    r11
0000019c`d6d30036 49bb37d8d4fbf77f0000 mov r11,offset comptest!ILT+51250(??$fourthArgHYAHHHHHZ) (00007ff7`fbd4d837)
0000019c`d6d30040 41ffd3          call    r11
0000019c`d6d30043 415b            pop     r11
0000019c`d6d30045 c3              ret
```

(no one saves non-volatile registers and stack frame)

Template handler (MSVC specific):

```asm
comptest!ILT+51250(??$fourthArgHYAHHHHHZ):
00007ff7`fbd4d837 e904332500      jmp     comptest!fourthArg<int> (00007ff7`fbfa0b40)
```

comptest!fourthArg<int>:

```asm
comptest!fourthArg<int>:
00007ff7`fbfa0b40 44894c2420      mov     dword ptr [rsp+20h],r9d ss:00000004`0473f1d8=00000000
00007ff7`fbfa0b45 4489442418      mov     dword ptr [rsp+18h],r8d
00007ff7`fbfa0b4a 89542410        mov     dword ptr [rsp+10h],edx
00007ff7`fbfa0b4e 894c2408        mov     dword ptr [rsp+8],ecx
00007ff7`fbfa0b52 57              push    rdi
00007ff7`fbfa0b53 8b442428        mov     eax,dword ptr [rsp+28h]
[...]

PR #3389, opened 2018-12-20
Add linkage tests for Microsoft x64 calling convention

samolisov:
Microsoft x64 calling convention brings addition limitations for x86-64:
only four registers are available for arguments (not four for integer
and four for float but four for both). Also the caller has
responsibility to allocate 32 bytes of "shadow space" on the stack right
before calling the function (regardless of the actual parameters used).

Two tests are added to check these circumstances: the first just
demonstrates how to invoke a method with five arguments while the second
one invokes a method who is actively using the stack and can corrupts
the return address if no "shadow space" has been allocated.

Issue: #3366

### DIFF
--- a/fvtest/compilertriltest/LinkageTest.cpp
+++ b/fvtest/compilertriltest/LinkageTest.cpp
@@ -215,14 +215,14 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFiveArg) {
  */
 template <typename T>
 T stackUser(T a, T b, T c, T d, T e) {
-	volatile T x, y, z, w;
-	for (int i = 0; i < a; i+= b) {
-	    w = (c + b) * (i + e);
-		x = a * (i + d + 1);
-		y = b - d + c;
-		z = c * 2 * (i + 1);
-	}
-	return x + y + z + w;
+    volatile T x, y, z, w;
+    for (int32_t i = 0; i < a; i+= b) {
+        w = (c + b) * (i + e);
+        x = a * (i + d + 1);
+        y = b - d + c;
+        z = c * 2 * (i + 1);
+    }
+    return x + y + z + w;
 }
 
 /**
@@ -283,5 +283,76 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFiveArgToStackUser) {
     EXPECT_EQ(static_cast<TypeParam>(2053),     entry_point(1,1,1,1,static_cast<TypeParam>(1024)))     << "Input Trees: " << inputTrees;
     EXPECT_EQ(static_cast<TypeParam>(3),        entry_point(1,1,1,1,static_cast<TypeParam>(-1)))       << "Input Trees: " << inputTrees;
     EXPECT_EQ(static_cast<TypeParam>(0x1e1e23), entry_point(1,1,1,1,static_cast<TypeParam>(0xf0f0f)))  << "Input Trees: " << inputTrees;
+#endif
+}
+
+class LinkageWithMixedTypesTest : public TRTest::JitTest {};
+
+int32_t fourthArgFromMixedTypes(double a, int32_t b, double c, int32_t d) { return d; }
+
+typedef int32_t (*FourMixedArgumentFunction)(double,int32_t,double,int32_t);
+
+TEST_F(LinkageWithMixedTypesTest, SystemLinkageParameterPassingFourArgWithMixedTypes) {
+    char inputTrees[400] = {0};
+    const auto format_string = "(method return=Int32 args=[Double,Int32,Double,Int32]"
+                               "  (block (ireturn (icall address=0x%jX args=[Double,Int32,Double,Int32] linkage=system"
+                                 " (dload parm=0)"
+                                 " (iload parm=1)"
+                                 " (dload parm=2)"
+                                 " (iload parm=3)"
+                                 ") )  ))";
+    std::snprintf(inputTrees, 400, format_string,
+        reinterpret_cast<uintmax_t>(static_cast<FourMixedArgumentFunction>(fourthArgFromMixedTypes)));
+
+    auto trees = parseString(inputTrees);
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+    auto entry_point = compiler.getEntryPoint<FourMixedArgumentFunction>();
+
+    EXPECT_EQ(1024,    entry_point(0.0,0,0.0,1024))     << "Input Trees: " << inputTrees;
+    EXPECT_EQ(-1,      entry_point(0.0,0,0.0,-1))       << "Input Trees: " << inputTrees;
+    EXPECT_EQ(0xf0f0f, entry_point(0.0,0,0.0,0xf0f0f))  << "Input Trees: " << inputTrees;
+#endif
+}
+
+double fifthArgFromMixedTypes(double a, int32_t b, double c, int32_t d, double e) { return e; }
+
+typedef double (*FiveMixedArgumentFunction)(double,int32_t,double,int32_t,double);
+
+TEST_F(LinkageWithMixedTypesTest, SystemLinkageParameterPassingFiveArgWithMixedTypes) {
+    char inputTrees[400] = {0};
+    const auto format_string = "(method return=Double args=[Double,Int32,Double,Int32,Double]"
+                               "  (block (dreturn (dcall address=0x%jX args=[Double,Int32,Double,Int32,Double] linkage=system"
+                                 " (dload parm=0)"
+                                 " (iload parm=1)"
+                                 " (dload parm=2)"
+                                 " (iload parm=3)"
+                                 " (dload parm=4)"
+                                 ") )  ))";
+    std::snprintf(inputTrees, 400, format_string,
+        reinterpret_cast<uintmax_t>(static_cast<FiveMixedArgumentFunction>(fifthArgFromMixedTypes)));
+
+    auto trees = parseString(inputTrees);
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+    auto entry_point = compiler.getEntryPoint<FiveMixedArgumentFunction>();
+
+    EXPECT_DOUBLE_EQ(1024.3, entry_point(0.0,0,0.0,0,1024.3)) << "Input Trees: " << inputTrees;
+    EXPECT_DOUBLE_EQ(-1.7,   entry_point(0.0,0,0.0,0,-1.7))   << "Input Trees: " << inputTrees;
+    EXPECT_DOUBLE_EQ(0.001,  entry_point(0.0,0,0.0,0,0.001))  << "Input Trees: " << inputTrees;
 #endif
 }
--- a/fvtest/compilertriltest/LinkageTest.cpp
+++ b/fvtest/compilertriltest/LinkageTest.cpp
@@ -105,7 +105,6 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingSingleArg) {
 template <typename T>
 T fourthArg(T a, T b, T c, T d) { return d; }
 
-
 TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
     char inputTrees[400] = {0};
     const auto format_string = "(method return=%s args=[%s,%s,%s,%s] (block (%sreturn (%scall address=0x%jX args=[%s,%s,%s,%s] linkage=system"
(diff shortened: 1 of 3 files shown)


Leonardo2718:
@genie-omr build win

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,138 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+#endif
+}
+
+/*
+ * It doesn't matter what this function exactly do. It is requred just to demonstrate what can be expected if a callee
+ * actively uses the stack (has a lot of local variables).
+ */
+template <typename T>
+T stackUser(T a, T b, T c, T d, T e) {
+	T x, y, z, w;
```
Leonardo2718: For the sake of being absolutely sure the build compiler doesn't mess with this function too much, it might be worth making these variables volatile:

```suggestion
	volatile T x, y, z, w;
```

samolisov: @Leonardo2718 Thank you for the suggestion, I've added the `volatile` keyword.

samolisov:
I've also added a test with mixed parameters: (double, int, double, int, double) and (double,int,double,int) and now are working on a JITed-to-JITed test (and get a fail now). So, I think it's better to mark the PR as `WIP`.

Do I right understand, currently JITed-to-JITed methods also must use the `system` linkage?

Leonardo2718:
Thanks @samolisov. Those are definitely useful data points.

I don't see a need to put the tests with mixed parameters in a separate file as they are testing the same behaviour as the other ones (even though they are not typed tests). :slightly_smiling_face: 

> Do I right understand, currently (on OMR) JITed-to-JITed methods also must use the system linkage?

That's correct. Currently, although support for different linkages exists, the OMR compiler does not distinguish between different kinds of calls so everything defaults to using system linkage.

fjeremic:
Probably a good idea if we get @0dvictor to review this as well given he has fixed some wicked bugs in this area very recently in OpenJ9.

0dvictor:
Thank you for providing these valuable tests. The tests look good to me. My only concern is that OMR they test areas that OMR currently does not have proper support. Maybe we shouldn't merge this PR until #3366 is resolved? I'm working on #3366 and I'm preparing the fix/implementation. It should be ready next week.

samolisov:
@Leonardo2718 The tests in `LinkageTest.cpp` are typed while the tests in `LinkageWithMixedTypesTest` are not. I think they can be in a separate file.

Leonardo2718:
Thanks @0dvictor! I actually think it would be okay to merge the tests right away, as long as they are dissabled. Then, when the underyling problem is solved, we can just renable them. :slightly_smiling_face: 

@samolisov 

> The tests in `LinkageTest.cpp` are typed while the tests in `LinkageWithMixedTypesTest` are not.

Yes, that's true. However, I prefer putting related tests together in one file, even if they use different tools or approaches (typed vs. "un-typed"). Unless there are so many tests in one file that it becomes difficult to manage (which I don't think is the case here), putting them in the same file makes them easier to browes. :slightly_smiling_face:

samolisov:
@Leonardo2718 Ok, I think you are right. The tests from `LinkageWithMixedTypeTest.cpp` are merged into `LinkageTest.cpp`.

charliegracie:
@genie-omr build win

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+                                                  );
+
+    auto trees = parseString(inputTrees);
+
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
```
Leonardo2718: I'm not sure there is much value in executing everything that came before this. Please skip this test by adding the following at the beginning of the test case body:

```c
SKIP_IF(OMRPORT_ARCH_HAMMER == omrsysinfo_get_CPU_architecture(), MissingImplementation)
   << "Test is skipped on non-x86-64 platforms because calls are not currently supported on other platforms (see issue #1645)";
```

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
```
Leonardo2718: Please use `SKIP_IF()` at the start of test body.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+
```
Leonardo2718: Are the extra blank lines intentional?

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+#endif
+}
+
+/*
+ * It doesn't matter what this function exactly do. It is requred just to demonstrate what can be expected if a callee
+ * actively uses the stack (has a lot of local variables).
+ */
+template <typename T>
+T stackUser(T a, T b, T c, T d, T e) {
+	volatile T x, y, z, w;
```
Leonardo2718: Indentation should use spaces here not tabs.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+                                                  );
+
+    auto trees = parseString(inputTrees);
+
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
```
Leonardo2718: Please use `SKIP_IF()` at the start of test body.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+
```
Leonardo2718: Are the extra blank lines intentional

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    auto entry_point = compiler.getEntryPoint<TypeParam (*)(TypeParam,TypeParam,TypeParam,TypeParam,TypeParam)>();
+
+    EXPECT_EQ(static_cast<TypeParam>(2053),     stackUser<TypeParam>(1,1,1,1,static_cast<TypeParam>(1024)))     << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(3),        stackUser<TypeParam>(1,1,1,1,static_cast<TypeParam>(-1)))       << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(0x1e1e23), stackUser<TypeParam>(1,1,1,1,static_cast<TypeParam>(0xf0f0f)))  << "Input Trees: " << inputTrees;
+
+    EXPECT_EQ(static_cast<TypeParam>(2053),     entry_point(1,1,1,1,static_cast<TypeParam>(1024)))     << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(3),        entry_point(1,1,1,1,static_cast<TypeParam>(-1)))       << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(0x1e1e23), entry_point(1,1,1,1,static_cast<TypeParam>(0xf0f0f)))  << "Input Trees: " << inputTrees;
```
Leonardo2718: I'm not sure I understand this. Isn't the point of `stackUser()` to do something like

```c++
EXPECT_EQ(stackUser<TypeParam>(1,1,1,1,static_cast<TypeParam>(1024)),
          entry_point(1,1,1,1,static_cast<TypeParam>(1024)))
      << "Input Trees: " << inputTrees;
```

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+#endif
+}
+
+class LinkageWithMixedTypesTest : public TRTest::JitTest {};
+
+int fourthArgFromMixedTypes(double a, int b, double c, int d) { return d; }
```
Leonardo2718: By convention, we tend to use `int32_t` for integer types instead of just `int` in the OMR compiler and related tests.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    std::snprintf(inputTrees, 400, format_string,
+        reinterpret_cast<uintmax_t>(static_cast<FourMixedArgumentFunction>(fourthArgFromMixedTypes)));
+
+    auto trees = parseString(inputTrees);
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
```
Leonardo2718: Please use `SKIP_IF()` at the start of test body.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    std::snprintf(inputTrees, 400, format_string,
+        reinterpret_cast<uintmax_t>(static_cast<FiveMixedArgumentFunction>(fifthArgFromMixedTypes)));
+
+    auto trees = parseString(inputTrees);
+    ASSERT_NOTNULL(trees) << "Trees failed to parse\n" << inputTrees;
+
+    // Execution of this test is disabled on non-X86 platforms, as we
+    // do not have trampoline support, and so this call may be out of
+    // range for some architectures.
+#ifdef TR_TARGET_X86
```
Leonardo2718: Please use `SKIP_IF()` at the start of test body.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -151,3 +150,291 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    Tril::DefaultCompiler compiler{trees};
+    ASSERT_EQ(0, compiler.compile()) << "Compilation failed unexpectedly\n" << "Input trees: " << inputTrees;
+
+    auto entry_point = compiler.getEntryPoint<FourMixedArgumentFunction>();
+
+    EXPECT_EQ(1024,    entry_point(0.0,0,0.0,1024))     << "Input Trees: " << inputTrees;
+    EXPECT_EQ(-1,      entry_point(0.0,0,0.0,-1))       << "Input Trees: " << inputTrees;
+    EXPECT_EQ(0xf0f0f, entry_point(0.0,0,0.0,0xf0f0f))  << "Input Trees: " << inputTrees;
```
Leonardo2718: Would it be possible to turn this into a value-parameterize test?

fvtest/compilertriltest/LinkageTest.cpp
0xdaryl: FYI... `SKIP_IF` was recently introduced in #3539.

samolisov: Maybe you mean `SKIP_IF(OMRPORT_ARCH_HAMMER != omrsysinfo_get_CPU_architecture(), MissingImplementation)`? Otherwise test fails on my Windows X64. I see in `port\win32\omrsysinfo.c` the following code:

```cpp
#elif defined( _AMD64_)
	return OMRPORT_ARCH_HAMMER;
```

fvtest/compilertriltest/LinkageTest.cpp
samolisov: @Leonardo2718 I use different values for both tests in `LinkageWithMixedTypesTest` since one parameter list ends with integer while the other ends with double. There could be two value-parameterized tests or I can pass a vector of values (but with different types and it could make the tests more compilcated) but I see no reason for it: the tests will be more difficult for reading and understanding what is being testing there.

fvtest/compilertriltest/LinkageTest.cpp
samolisov: Thanks, fixed.

fvtest/compilertriltest/LinkageTest.cpp
samolisov: Thanks, fixed.

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -136,18 +135,288 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+    EXPECT_EQ(static_cast<TypeParam>(0x1e1e23), stackUser<TypeParam>(1,1,1,1,static_cast<TypeParam>(0xf0f0f)))  << "Input Trees: " << inputTrees;
+
+    EXPECT_EQ(static_cast<TypeParam>(2053),     entry_point(1,1,1,1,static_cast<TypeParam>(1024)))     << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(3),        entry_point(1,1,1,1,static_cast<TypeParam>(-1)))       << "Input Trees: " << inputTrees;
+    EXPECT_EQ(static_cast<TypeParam>(0x1e1e23), entry_point(1,1,1,1,static_cast<TypeParam>(0xf0f0f)))  << "Input Trees: " << inputTrees;
+}
+
+class LinkageWithMixedTypesTest : public TRTest::JitTest {};
+
+int fourthArgFromMixedTypes(double a, int32_t b, double c, int32_t d) { return d; }
```
Leonardo2718:
```suggestion
int32_t fourthArgFromMixedTypes(double a, int32_t b, double c, int32_t d) { return d; }
```

Forgot about the return type :stuck_out_tongue:

fvtest/compilertriltest/LinkageTest.cpp
```diff
@@ -136,18 +135,288 @@ TYPED_TEST(LinkageTest, SystemLinkageParameterPassingFourArg) {
...
+
+typedef int (*FourMixedArgumentFunction)(double,int32_t,double,int32_t);
```
Leonardo2718:
```suggestion
typedef int32_t (*FourMixedArgumentFunction)(double,int32_t,double,int32_t);
```

fvtest/compilertriltest/LinkageTest.cpp
Leonardo2718: I see, you're right.

As an alternative then to increase input coverage, how about something like:

```c++
    auto inputs = TRTest::const_values<int32_t>();
    for (auto i = inputs.cbegin(); i != inputs.cend(); ++i) {
        SCOPED_TRACE(*i);
        EXPECT_EQ(*i, entry_point(0.0,0,0.0, *i))  << "Input Trees: " << inputTrees;
    }
```

fvtest/compilertriltest/LinkageTest.cpp
samolisov: @Leonardo2718 thank you, fixed.

fvtest/compilertriltest/LinkageTest.cpp
samolisov: @Leonardo2718 Thank you, fixed.

fvtest/compilertriltest/LinkageTest.cpp
samolisov: @Leonardo2718 I've defined two additional `const_values` template implementations: 

```cpp
template <>
inline std::vector<float> const_values<float>()
```

and

```cpp
template <>
inline std::vector<double> const_values<double>()
```

and changed the three test 'expections' in every test to a check against each const_value except the `SystemLinkageParameterPassingFiveArgToStackUser` test (because the expected value is not matched to the last argument).

```cpp
    auto inputs = TRTest::const_values<TypeParam>();
    for (auto i = inputs.cbegin(); i != inputs.cend(); ++i) {
        SCOPED_TRACE(*i);
        EXPECT_EQ(static_cast<TypeParam>(*i), entry_point(0,0,0,0, *i))  << "Input Trees: " << inputTrees;
    }
```

fvtest/compilertriltest/LinkageTest.cpp
samolisov: I've add a comment, the goal is just to test the fact the compiled function itself works fine.

samolisov:
@Leonardo2718 would you like all of the commits be squashed into one?  I think we can use individual commits for every small logical change such as 'Add linkage tests with mixed arg types' or 'Add a test against a constant array'.

Leonardo2718:
@samasri 

>  I think we can use individual commits for every small logical change such as 'Add linkage tests with mixed arg types' or 'Add a test against a constant array'.

Sorry. Yes, that's what I meant. Only the small commits (e.g. the one that changes `int` &rarr; `int32_t`) should be squashed since they just change code that was added by another commit.

fvtest/compilertriltest/LinkageTest.cpp
Leonardo2718: For some reason, it looks like the return type is `int` again. :confused:

fvtest/compilertriltest/LinkageTest.cpp
Leonardo2718: For some reason, it looks like the return type is `int` again. :confused:

fvtest/compilertriltest/LinkageTest.cpp
Leonardo2718: Hm... Ok, I see.

fvtest/compilertriltest/LinkageTest.cpp
samolisov: @Leonardo2718 my bad, sorry. Fixed.

samolisov:
@Leonardo2718 I've fixed a bug with 'int' in returns and squashed the 'int -> int32_t'  commit with that where the function was introduced. If tests are passed on Travis and Appveyor, the commits will be ready to a merge.

Leonardo2718:

samolisov:
'''
10:30:46 /Users/jenkins/workspace/PullRequest-osx_x86-64/fvtest/compilertriltest/LinkageTest.cpp:59:36: error: use of undeclared identifier 'privateOmrPortLibrary'
10:30:46     SKIP_IF(OMRPORT_ARCH_HAMMER != omrsysinfo_get_CPU_architecture(), MissingImplementation)
'''

I think I miss getting of the OMR port library before using 'omrsysinfo_get_CPU_architecture()'. How to get the port library?

samolisov:
@Leonardo2718 could you re-trigger builds? I see the `omrsysinfo_get_CPU_architecture()` macro could not be expanded well from the `SKIP_IF` macro. I see in the other tests `omrsysinfo_get_CPU_architecture()` is invoked before the `SKIP_IF` macro. After a successful build, I'll squash the last commit.

Leonardo2718:
Oh! I've actually seen this problem before. The fix is to add

```c++
OMRPORT_ACCESS_FROM_OMRPORT(TRTest::TestWithPortLib::privateOmrPortLibrary);
```

at the beginning of every typed test (before the `SKIP_IF`). I'm not sure why it's needed but it seems to work.

Sorry you had to run into that.

0xdaryl:
> I'm not sure why it's needed

It is a macro that essentially defines a "private" port library pointer local variable that is initialized from the parameter to the macro.  In this case, the parameter is the port library itself.  However, there are other macros available where the port library can be initialized from a VMThread or an OMR_VM (for example) if that's all you had on hand.  In that case the macro would know how to dig out the port library from whatever structure you passed into the macro.

Leonardo2718:
> It is a macro that essentially defines a "private" port library pointer local variable that is initialized from the parameter to the macro. In this case, the parameter is the port library itself. However, there are other macros available where the port library can be initialized from a VMThread or an OMR_VM (for example) if that's all you had on hand. In that case the macro would know how to dig out the port library from whatever structure you passed into the macro.

:sweat_smile: Well, I knew that part. The thing is, I explicitly defined the `TRTest::TestWithPortLib` class so that no `OMRPORT_ACCESS_FROM_*` macros would be needed within a test. What I don't understand is why that mechanism in `TestWithPortLib` doesn't work with TypedTests. It works fine for other tests, eg:

https://github.com/eclipse/omr/blob/1537cc08bd468d71d214ad621a2d7d6df3b4445c/fvtest/compilertriltest/BitPermuteTest.cpp#L380-L384

samolisov:

Leonardo2718:
@genie-omr build all

samolisov:
Some tests are failed on `linux_x86-64` and `osx_x86-64` but no LinkageTest:

```
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/0, where TypeParam = int
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/0 (4 ms total)
16:41:52  27: 
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/1, where TypeParam = long long
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/1 (4 ms total)
16:41:52  27: 
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/2, where TypeParam = float
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/2 (4 ms total)
16:41:52  27: 
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/3, where TypeParam = double
16:41:52  27: [0;32m[----------] [m6 tests from LinkageTest/3 (5 ms total)
16:41:52  27: 
16:41:52  27: [0;32m[----------] [m2 tests from LinkageWithMixedTypesTest
16:41:52  27: [0;32m[----------] [m2 tests from LinkageWithMixedTypesTest (1 ms total)
```

Leonardo2718:
Hm... that's curious. Can you try rebasing on top of the latest master?

Leonardo2718:
Actually, I know what the problem is. The new specialization of `const_values<>()` for the type `float` is exposing an existing problem. :man_facepalming: 

Please add

```C++
SKIP_IF((param.value == static_cast<float>(0xFFFFFF0F)) &&
        (OMRPORT_ARCH_HAMMER != omrsysinfo_get_CPU_architecture()), 
        KnownBug)
     << "f2i test behaves unexpectedly on x86-64 with certain high input values (see issue #3602)";
```

here https://github.com/eclipse/omr/blob/4077a2cddf6bd503b76b0646d725a7b358dee5b0/fvtest/compilertriltest/TypeConversionTest.cpp#L973-L974 and here https://github.com/eclipse/omr/blob/4077a2cddf6bd503b76b0646d725a7b358dee5b0/fvtest/compilertriltest/TypeConversionTest.cpp#L998-L999

Leonardo2718:
@genie-omr build all

samolisov:
@Leonardo2718 I've added the skip to the TypeConversionTest but it looks like not enough:
```
 [0;32m[----------] [m20 tests from TypeConversionTest/FloatToInt32
17:00:39  27: /Users/jenkins/workspace/PullRequest-osx_x86-64/fvtest/compilertriltest/TypeConversionTest.cpp:998: Failure
17:00:39  27:       Expected: exp
17:00:39  27:       Which is: -2147483648
17:00:39  27: To be equal to: act
17:00:39  27:       Which is: 2147483647
17:00:39  27: [0;31m[  FAILED  ] [mTypeConversionTest/FloatToInt32.UsingConst/8, where GetParam() = (2.14748e+09, ("f2i", 0x68da870)) (1 ms)
17:00:39  27: /Users/jenkins/workspace/PullRequest-osx_x86-64/fvtest/compilertriltest/TypeConversionTest.cpp:1024: Failure
```

samolisov:
```
    std::string arch = omrsysinfo_get_CPU_architecture();
    SKIP_IF((param.value == static_cast<float>(0xFFFFFF0F)) && (OMRPORT_ARCH_HAMMER != arch), KnownBug)
```

May be `OMRPORT_ARCH_HAMMER == arch`?

Leonardo2718:
>  May be `OMRPORT_ARCH_HAMMER == arch`?

:disappointed: Sorry @samolisov. Yes, you're right. It should be `OMRPORT_ARCH_HAMMER == arch`.

samolisov:
@Leonardo2718 No problem, I've changed and pushed.

Leonardo2718:
@genie-omr build all

samolisov:
```
Which is: -2147483648
17:18:47  27: To be equal to: act
17:18:47  27:       Which is: 2147483647
```
2147483647 is not 0xFFFFFF0F. Should I rebase on Master to get change pass the tests?

Leonardo2718:
> 2147483647 is not 0xFFFFFF0F. Should I rebase on Master to get change pass the tests?

Yes, you're right about 0xFFFFFF0F. Rebasing is a good idea, though I don't think it's going to make the tests pass. Maybe also try:

```c++
SKIP_IF( (param.value < static_cast<float>(std::numeric_limits<int32_t>::min()) || param.value > static_cast<float>(std::numeric_limits<int32_t>::max()) && (OMRPORT_ARCH_HAMMER != arch), KnownBug)
```

Leonardo2718:
... or maybe

```c++
SKIP_IF( (param.value <= static_cast<float>(std::numeric_limits<int32_t>::min()) || param.value >= static_cast<float>(std::numeric_limits<int32_t>::max()) && (OMRPORT_ARCH_HAMMER != arch), KnownBug)
```

(using `<=`/`>=` instead of `<`/`>`)

samolisov:
@Leonardo2718 Ok, done. Let's try without rebasing.

Leonardo2718:
@genie-omr build all

samolisov:
@Leonardo2718 `osx_x86-64` and `list_x86-64` are passed well, wow.

samolisov:
@Leonardo2718 `continuous-integration/eclipse-omr/pr/zos_390-64` are waiting for a machine for building. Do you agree, the PR can be merged?

Leonardo2718:
> `continuous-integration/eclipse-omr/pr/zos_390-64 are waiting for a machine for building`. Do you agree, the PR can be merged?

Sure. The zos builds do not currently run Tril tests anyways, so waiting for that build to complete adds no value. Merging.

PR #3390, opened 2018-12-20
Add clang format

rwy0717:
The formatting style is heavily based on webkit, with only a few differences.

The webkit style can be found here: https://webkit.org/code-style-guidelines/

The differences are:

* Use a column limit of 120, rather than unlimited. This is needed to allow
  clang-format to reindent comments. 120 was chosen because it is a good
  middleground, and happens to be the max that github displays well.  In the
  future, this may become a soft (unenforced) limit.
* Sorting includes is disabled, because rearranging headers in OMR causes
  compilation issues. In the future, this may be turned on for new code.
* The Standard is set to C++03. This means we put a space between angle
  brackets (IE ` > >`) when closing a template. This is needed for xlc.
* Do not indent inside any namespaces in OMR. Rather, we add a comment when
  closing any namespace.
* Do not split empty functions/classes/blocks into multiple lines.
* Do not reflow the spdx license identifier.
* Place template declarations on their own line. In webkit, this is optional,
  but for the sake of reflowing templated code, this is forced behaviour. If
  the column length ever becomes a soft limit, this might also disappear.
* Left align escaped newlines (backslashes in macros), which gets us
  closer to the webkit style while using a limited column width.

The exact diff between our clang-format and webkit follows:

```diff
--- ./scratch/webkit.clang-format	2018-12-18 13:43:32.028092627 -0500
+++ .clang-format	2018-12-20 11:43:09.080364796 -0500
@@ -3 +3 @@
-# BasedOnStyle:  WebKit
+BasedOnStyle:  WebKit
@@ -8 +8 @@
-AlignEscapedNewlines: Right
+AlignEscapedNewlines: Left
@@ -20 +20 @@
-AlwaysBreakTemplateDeclarations: MultiLine
+AlwaysBreakTemplateDeclarations: Yes
@@ -36,3 +36,3 @@
-  SplitEmptyFunction: true
-  SplitEmptyRecord: true
-  SplitEmptyNamespace: true
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
@@ -40 +40 @@
-BreakBeforeBraces: WebKit
+BreakBeforeBraces: Custom
@@ -48,2 +48,2 @@
-ColumnLimit:     0
-CommentPragmas:  '^ IWYU pragma:'
+ColumnLimit:     120
+CommentPragmas:  '^ IWYU pragma:|SPDX-License-Identifier:'
@@ -58 +58 @@
-FixNamespaceComments: false
+FixNamespaceComments: true
@@ -82 +82 @@
-NamespaceIndentation: Inner
+NamespaceIndentation: None
@@ -97 +97 @@
-SortIncludes:    true
+SortIncludes:    false
@@ -114 +114 @@
-Standard:        Cpp11
+Standard:        Cpp03
```


rwy0717:
I've formatted the codebase and pushed to [`rwy0717/omr/tree/clang-format`](https://github.com/rwy0717/omr/tree/clang-format).

Here are a few choice examples:
* [functions with many parameters](https://github.com/rwy0717/omr/blob/clang-format/util/hashtable/hashtable.c#L234-L238)
* [Multiline macros](https://github.com/rwy0717/omr/blob/clang-format/gc/base/ModronAssertions.h)
* [Constructors](https://github.com/rwy0717/omr/blob/clang-format/gc/base/EnvironmentBase.hpp#L663-L708)
* [Template metaprogramming](https://github.com/rwy0717/omr/blob/clang-format/include_core/OMR/TypeTraits.hpp)
* [Big if expressions](https://github.com/rwy0717/omr/blob/clang-format/compiler/optimizer/LocalOpts.cpp#L1977-L1982)

samolisov:
I had a look at the formatted code, looks good for me but I see brackets after a function declaratin are put on a new line while afrer a block construction (if/while/try,...) are stay on the same line. Also, in the compiler many classes are forward declared as 

namespace TR { class CodeGeneration; }
namespace TR { class Linkage; }

Now the namespace declarations will be on new lines:

namespace TR {
class CodeGenerator;
}

namespace TR {
class Linkage;
}

I think it is a good idea just to put all forward declarations in a file under one namspace umbrella.

rwy0717:
I've also formatted the openj9 runtime in the same style, it's available at https://github.com/rwy0717/openj9-reformat. Here is the [bytecode interpreter](https://github.com/rwy0717/openj9-reformat/blob/594367c71f6e22e5df5d0faedd5a84d54eb0c2d1/runtime/vm/BytecodeInterpreter.hpp#L610).

samolisov:
@rwy0717 for me personally, new brace on a new line after many parameters is better, but if method has not so many parameters and the signature can be placed on one line, the brace on the same line looks more "usual". Even though I'm not sure whether having two rules for braces in function declarations is better than only one.

rwy0717:
> Was the column limit set only because of comments? If so we can format once with the 120 column limit to get the comments fixed then format again with column limit 0 and ReflowComments: false so the comments stay put but code can have unlimited column width.

Once you switch to "unlimited columns", the formatter will stop reflowing the code, leaving any newlines in place. This means that the second formatting pass should (hopefully) have no effect on the codebase. We _could_ use an extremely large column limit in the second pass, but I haven't experimented with this yet. I suspect 

> The compiler options processing table looks to have regressed [3]. Is there any way to keep the assignments within an enum aligned? I see there is AlignConsecutiveAssignments but that seems it will align stuff in functions too, which is not what we want. Only enums. 

You're right.

> Is there a way to do that?

I don't believe so. If we want this to be aligned, I would use `// clang-format off` to preserve the tabled formatting.

> Curious for [4] why some functions are one line and others are multiple lines? Is it because the one-liners are only one statement/expression?

Yep, as long as it fits within the column-limit.  The WebKit default will single-line-ify any type of short function. When the column-limit is very large, the formatter will put some surprisingly big functions on a single line. If we want, it's possible to enable "short functions on single line" only for inline member functions.

rwy0717:
Since we never committed to this formatting style, I'm closing this PR.

PR #3394, opened 2018-12-21
AArch64: Implement ARM64Src1/Src2Instruction

knn-k:
This commit implements ARM64Src1Instruction and ARM64Src2Instruction
for use with "fcmp" instructions.


knn-k:
This is a preparation for evaluators such as `iffcmpeq` and `dcmplt`.
Floating-point comparison instruction has a different instruction format from that of integer comparison, which is actually an alias of `subs` instruction with zero register as the target.
So adding ARM64Instructions with no target register.

knn-k:
This one is already tested with #3402, #3405 and their test cases.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3396, opened 2018-12-23
WIP: Simplify lowest level compiler allocation classes

mstoodle:
This PR prereqs a change in OpenJ9 to avoid breaking OpenJ9, so
it has been marked WIP. Reviews are welcome at this stage.

The lowest level memory allocation classes in the OMR compiler
have a complicated hierarchy and relationship that many people
have struggled to understand and maintain. The complexity came
about trying to hold a unified picture of memory allocation
between OMR (which just uses a raw malloc allocator) and OpenJ9
(which uses J9MemorySegments allocated from the VM a.k.a.
"the system"). Unfortunately, this unity did not materialize
and so these lower level allocation classes are not actually
shared with OpenJ9.

The set of changes in this commit is motivated by the effort
to try to compile the basic JitBuilder files in compiler/ilgen
as part of OpenJ9. In this commit, I brought over the code from
OpenJ9's SystemSegmentProvider and refactored it so that it
could operate in OMR using TR::MemorySegments rather than
J9MemorySegments. I changed its name to SegmentAllocator and
declare the SegmentAllocator as the base layer for the higher
level allocation support for e.g. regions and STL and other
objects in the compiler component.

This commit is the first stage in a series of coordinated
changes to OMR and OpenJ9 to migrate from the "old" memory
allocation scheme to the "new" memor allocation scheme. In
this change, *most* of the new code is implemented inside
a new compiler/env/mem directory which contains any files
from compiler/env that required substantial modification.
This commit also introduces a temporary macro called,
uninspiringly, OLD_MEMORY: if OLD_MEMORY is defined then
the old OMR memory allocation classes will be built rather
than the new ones. By default after this commit, OMR will
build with the *new* memory allocation classes (e.g. into
jitbuilder and compilertest). Most consumers of the OMR
compiler should experience *zero* functional differences
as the memory classes being changed are at the lowest levels
of the compiler implementation (well lower than the facilities
most compiler developers use directly).

The SegmentAllocator will hold a BackingMemoryAllocator and a
RawAllocator.  In fact, the BackingMemoryAllocator in OMR is
an only slightly beefed up RawAllocator which returns memory
wrapped in a TR::MemorySegment. I deleted the SegmentPool code
because nobody was using it. The DebugSegmentProvider code has
been moved into DebugRawAllocator (subclass of TR::RawAllocator)
as well as DebugSegmentAllocator (subclass of TR::SegmentAllocator).

I added some tracing output to the SegmentAllocator and
BackingMemoryAllocator to track segments as they are allocated
and deallocated (see MEMLOG macros in each file). This debug
output uses printfs due to the primordial environment of the
memory allocators, and is activated by environment variables
OMRDebug_BackingMemory=1 or 2 or 3    (increasing detail)
OMRDebug_SegmetnAllocator=1 or 2 or 3 (increasing detail)

While debugging the memory allocation layers, I discovered that
the original list management for segments was not very efficient
in that new segments were typically added at the end of lists but
searched starting from the beginning. This commit changes the
order of search to require fewer probes (less than half the
probes required in some very unscientific studies). I doubt
it will have any performance impact in most workloads but
it felt better to fix it.

BackingMemoryAllocator, SegmentAllocator, and DebugSegmentAllocator
are all extensible classes to facilitate a similar refactoring
extravaganza to be inflicted upon OpenJ9 very soon.


mstoodle:
genie-omr build all

mstoodle:
genie-omr build all

mstoodle:
genie-omr build all

mstoodle:
genie-omr build arm

mstoodle:
Note to reviewers; I still have to update the compiler memory documentation in this PR.

compiler/control/CompileMethod.cpp
```diff
@@ -276,14 +282,35 @@ compileMethodFromDetails(
...
+   TR::BackingMemoryAllocator backingMemoryAllocator(rawAllocator, 1 << 16);;
+
+   TR::DebugSegmentAllocator debugSegmentAllocator(backingMemoryAllocator, 1 << 16);
+   TR::SegmentAllocator defaultSegmentAllocator(backingMemoryAllocator, 1 << 16);
+   TR::SegmentAllocator &segmentAllocator =
+      TR::Options::getCmdLineOptions()->getOption(TR_EnableScratchMemoryDebugging) ?
+      debugSegmentAllocator : defaultSegmentAllocator;
+#endif
+
+   TR::Region dispatchRegion(segmentAllocator, rawAllocator);
```
dsouzai: `segmentAllocator` isn't defined when using `OLD_MEMORY`; I guess the variable is named `segmentProvider`.

dsouzai: Given the change below (ie using `segmentProvider` for `OLD_MEMORY`, and `segmentAllocator` for the new stuff), it seems like you're going to have to duplicate the initialization of the `TR::Region` in each of the `ifdef` sections

compiler/env/mem/OMRBackingMemoryAllocator.cpp
```diff
@@ -0,0 +1,125 @@
...
+TR::MemorySegment &
+OMR::BackingMemoryAllocator::allocate(size_t requiredSize, void * hint)
+   {
+   size_t sizeToAllocate = std::max(requiredSize, _minimumAllocationSize);
+   MEMLOG(1, "allocate request %lu will allocate at least %lu\n", requiredSize, sizeToAllocate);
+
+   void *p = _rawAllocator.allocate(sizeToAllocate, hint);
+   TR::MemorySegment *newSegment = new (_rawAllocator) TR::MemorySegment(p, sizeToAllocate);
+   if (newSegment == NULL)
+      throw std::bad_alloc();
```
dsouzai: Unless `std::nothrow_t tag` is specified as a parameter to the `placement new` (which in this case it isn't), `OMR::RawAllocator` throws a `std::bad_alloc()`, this is redundant.

compiler/env/mem/OMRBackingMemoryAllocator.cpp
```diff
@@ -0,0 +1,125 @@
...
+      auto ref = *it;
+      MEMLOG(3, "_allocatedSegments iterating segment %p\n", &(ref.get()));
+      if (segment == ref.get())
+         {
+         _allocatedSegments.erase(it);
+         found=true;
+         break;
+         }
+      }
+   TR_ASSERT(found, "Request to deallocate segment not found in allocated segment list");
```
dsouzai: How "fatal" is it if we can't find the segment in this deque?

compiler/env/mem/OMRBackingMemoryAllocator.cpp
```diff
@@ -0,0 +1,125 @@
...
+   _rawAllocator.deallocate(&segment, sizeof(TR::MemorySegment));;
+   }
+
+void
+OMR::BackingMemoryAllocator::deallocateSegments()
+   {
+   while (!_allocatedSegments.empty())
+      {
+      TR::MemorySegment &segment = _allocatedSegments.front().get();
+      deallocate(segment);
```
dsouzai: Just curious as to whether `BackingMemoryAllocator` can be extensible, and if so, whether we want to enforce that the `OMR` version of this is guaranteed to get called.

compiler/env/mem/OMRBackingMemoryAllocator.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#define MEMLOG(n,s, ...) do { if ((LOGLEVEL != NULL) && ((int)(*LOGLEVEL) >= (int)(n+'0'))) { printf("BckMem %p : " s, this, __VA_ARGS__); } } while (0)
+#define MEMLOG0(n,s)     do { if ((LOGLEVEL != NULL) && ((int)(*LOGLEVEL) >= (int)(n+'0'))) { printf("BckMem %p : " s, this); } } while (0)
+
+
+#include "env/mem/BackingMemoryAllocator.hpp"
+
+
+TR::MemorySegment &
+OMR::BackingMemoryAllocator::allocate(size_t requiredSize, void * hint)
```
dsouzai: This method doesn't check the `_allocationLimit` even though that concept exists at the `BackingMemoryAllocator` level.

dsouzai:
What is the reason `OMRCompilerEnv.cpp` is in the `mem` dir?

compiler/env/mem/OMRBackingMemoryAllocator.cpp
mstoodle: At best it's going to be a memory leak, and at worst I suppose it's a sign that memory corruption is likely happening in the compilation. I suppose a `TR_ASSERT_FATAL` would be a better option.

compiler/env/mem/OMRBackingMemoryAllocator.cpp
mstoodle: I'm not sure that's appropriate: if someone creates an extension of BackingMemoryAllocator that implements a different `allocate()` function then presumably they should have the ability to also provide a different `deallocate()` function that matches?

Someone who did that would probably want to further break down `deallocate()` so they could reuse some parts of it without necessarily using other parts (exercise left for those who need it :) ?)

compiler/env/mem/OMRBackingMemoryAllocator.cpp
mstoodle: Um, yeah...good point :)

jitbuilder/control/Jit.cpp
```diff
@@ -120,12 +123,14 @@ initializeJitBuilder(TR_RuntimeHelper *helperIDs, void **helperAddresses, int32_
    // Create a bootstrap raw allocator.
    //
    TR::RawAllocator rawAllocator;
+   TR::BackingMemoryAllocator backingAllocator(rawAllocator, 1<< 20);
```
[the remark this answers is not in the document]
mstoodle: They don't *have* to be linked together. That's just how they're used currently....I don't believe there's a requirement for the `BackingMemoryAllocator` to die just because the `SegmentAllocator` dies...

compiler/env/mem/OMRBackingMemoryAllocator.cpp
dsouzai: Looking at this a second time, wouldn't this result in the code going through the `_allocatedSegments` list for every iteration of the loop?

compiler/env/mem/OMRSegmentAllocator.cpp
```diff
@@ -0,0 +1,257 @@
...
+   {
+   // round up to a multiple of the default allocation size
+   size_t const roundedSize = ( ( ( requiredSize + (_allocationBlockSize - 1) ) / _allocationBlockSize ) * _allocationBlockSize );
+   MEMLOG(1, "allocate request %lu will allocate at least %lu\n", requiredSize, roundedSize);
+
+   // if we can recycle a free segment, then do that
+   if (!_freeSegments.empty() && roundedSize == _allocationBlockSize)
+      {
+      TR::MemorySegment &recycledSegment = _freeSegments.back();
+      TR_ASSERT(recycledSegment.size() == roundedSize, "unexpected size on free segment");
```
dsouzai: I think in the non-debug build, this should result in a `std::bad_alloc` exception, at least when `recycledSegment.size()` is less than `roundedSize`.

compiler/env/mem/OMRSegmentAllocator.cpp
```diff
@@ -0,0 +1,257 @@
...
+      }
+   TR_ASSERT(found, "deallocate() request for segment not in list of allocated segments");
+   // we cannot erase segmentToRemove from _allocatedSegments yet because _allocatedSegments owns the TR::MemorySegment object
+   //  so if we were to erase it now, that memory would be reclaimed before we're done with the freeing operations below
+
+   auto ref = *segmentToRemove;
+   MEMLOG(2, "deallocate %p ( %p ) was found in allocatedSegments\n", &segment, &(ref.get()));
+
+   // small segments that match the block size can just be put onto free list
+   if (segmentSize == _allocationBlockSize)
```
dsouzai: This is a pedantic comment,so ignore at will :): isn't this just a special case of `else if (_backingMemoryAllocator.fitsInMinimumSizeSegment(segmentSize))`?

dsouzai:
Overall, lgtm.

Was curious about this q though (whch I think got lost among all the other comments)

compiler/env/mem/OMRBackingMemoryAllocator.cpp
mstoodle: no, because they start at the same end of the list :). It asks what the head is, then removes the head (so `deallocate` just picks off the first segment). It might be possible to refactor `deallocate` to not rely on it and it probably deserves a comment, but i guess I went the easy way.

For the record, I found the list iteration in the earlier implementation, which I initially duplicated, was not as efficient as it should be, so I tweaked it a bit to make it better. So good catch :) .

compiler/env/mem/OMRSegmentAllocator.cpp
mstoodle: fair point: it may also justify `TR_ASSERT_FATAL` since SegmentAllocator is clearly doing something it shouldn't...

compiler/env/mem/OMRSegmentAllocator.cpp
mstoodle: well, remember that the backing memory allocator and the segment allocator don't need to use the same segment allocation size. And the segment allocator can satisfy a request with a segment that's a multple of the segment allocator's default size but still less than the backing memory allocator's default size.

For example, the backing memory allocator may be allocating no smaller than 1MB segments and the segment allocator may be allocating no smaller than 64KB segments. The segment allocator can handle a memory request for 256KB (to satisfy a request that's any larger than 192KB) from the current backing memory chunk. When that segment comes back, it can't just go onto the free list: it needs to be carved up into 64KB segments first.

dsouzai: Ah right; heh that's a subtlety I missed :)

dsouzai:
Well I guess I was asking more in terms of why it was in the "memory" dir, given that `OMRCompilerEnv` is about more than just the memory.

compiler/env/mem/OMRBackingMemoryAllocator.cpp
dsouzai: Heh I see, sneaky :P

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+
+The remainder of this document describes the OMR implemtation of the
+two levels of memory allocator management in the compiler.
+
 ### Low Level Allocators
-Low Level Allocators cannot be used by STL containers. They are 
-generally used by High Level Allocators as the underlying 
-provider of memory.
+Low level allocators interface directly with whatever memory provider
+is available outside of the JIT. In a languqage runtime environment,
```
dsouzai: typo: `languqage`

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+that may mean allocating memory from services that track memory allocated
+by the virtual machine. Or the memory provider can simply be "malloc".
+At its lowest level, the OMR memory allocator currently employ the
+system equivalent of malloc/mmap.
+
+There are three fundamental pieces to the lowest memory allocation
+layer: `TR::RawAllocator`, `TR::BackingMemoryAllocator`, and
+`TR::SegmentAllocator`. Each of these classes builds upon the
+earlier classes: one constructs a `TR::BackingMemoryAllocator`
```
dsouzai: Pedantic: it isn't immediately obvious that "one" here is meant to be the proverbial "one"; because of the previous sentence it could read as "one of the classes constructs a....".

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+              |                                 |
+              +---------------------------------+
+```
+
+As you can see, a `TR::SegmentAllocator` object contains a reference
+to a `TR::BackingMemoryAllocator` object, which itself contains a
+reference to a `TR::RawAllocator` object. The lifetimes of these objects
+are therefore clear: the raw allocator must outlive the other two, and
+the backing memory allocator must outlive the segment allocator. Only
+one segment allocator can use a backing memory allocator at a time,
```
dsouzai: Might be worth opening an issue as a backlog/beginner item to enforce this.

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+cause the compiler to allocate relatively large and consistently
+sized memory regions from its surrounding environment.
+Controlling allocations in this way can help to reduce fragmentation
+and simplifies core memory reuse without managing free lists (the
+higher level memory allocators do employ free lists to manage
+smaller chunks of memory).
+
+Each backing memory allocator uses a `TR::RawAllocator` to allocate
+memory when needed. Each such memory chunk is represented by
+a `TR::MemorySegment`. This TR::MemorySegment object is allocated
```
dsouzai: Missing backticks for `TR::MemorySegment`

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+directly using the raw allocator which must also be freed when
+the segment is deallocated. To ensure no memory leaks, the
+backing memory allocator maintains an `std::deque` with references
+to the memory segments that have been returned from `allocate()`.
+
+All requested backing memory allocations (i.e. calls to
+`allocate(size_t requestedSize)` will round up the allocation size
```
dsouzai: Missing `)` before "will round..."

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+When the backing memory object is destructed, it will first free
+all the `TR::MemorySegment`s as well as the raw memory each of
+of those segments represents. That memory can also be returned
+immediately by calling `deallocateSegments()`.
+
+The backing memory allocator tracks how much memory has been allocated
+via `_bytesAllocated` which can be queried via `bytesAllocated()`. The
+backing memory allocator can also be directed to enforce an
+allocation limit via a constructor option which sets the `_allocationLimit`
+field. By default, the backing memory allocator will not limito
```
dsouzai: typo: `limito`

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +15,378 @@ by the Compiler; there are too many participants for informal trust,
...
+
+A second new feature is that the segment allocator will try to
+reuse segments that have been passed to `deallocate()` without
+returning them to the backing memory allocator. To avoid the
+overhead of OS level memory allocators, the segment allocator
+tries hard not to allocate more memory (with the consequence
+that it will hold onto allocated memory longer than strictly
+required by the compiler).
+
+This reuse has one restrictions to simplify the reuse algorithm:
```
dsouzai: grammar: `restrictions`

dsouzai:
Btw, forgot to mention earlier, but the copyrights need to be updated :)

mstoodle:
Ah, now I understand the source of your question: the `mem` directory is temporary (yes, poorly named but hopefully not long lived): it holds the files requiring substantial changes between "old" and "new" memory management.

doc/compiler/memory/MemoryManager.md
mstoodle: agree, will fix

mstoodle: that is definitely an awkward sentence that I seem to have missed in my own review....thanks!

mstoodle:
genie-omr build all

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
-`TR_EnableScratchMemoryDebugging` option is enabled.
+The full set of classes used for lower level memory allocation, along
+with their relationships, is shown below:
+
+
+```
+RawAllocator classes:
++----------------------------+         +-----------------------------+
+|                            |  using  |                             |
+|     TR::RawAllocator       +<--------+     OMR::RawAllocator       |
```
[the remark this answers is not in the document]
mstoodle: I didn't see a good way to represent it, so "confusing" is a perfectly apt word to use.

`OMR::RawAllocator` is placed into the `TR` namespace via a `using` statement:
`namespace TR { using OMR::RawAllocator };`

mstoodle:
genie-omr build all

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+
+If the current backing memory chunk cannot satisfy the request,
+then a new backing memory allocation request will be needed
+to provide this memory to the compiler. There are two scenarios
+here: 1) the memory request is a large memory request that will
+allocate a dedicated segment for this allocation i.e. it will not
+later be reused, or 2) the memory request is a "regular" one
+but the current backing memory chunk has simply been exhausted.
+In the first case, a dedicated memory segment will be allocated
+and directly returned to the compiler.
```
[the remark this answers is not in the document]
mstoodle: i reorganized a bit to hopefully make it less repetitive...thanks!

mstoodle:
genie-omr build all

mstoodle:
genie-omr build all

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+backing memory allocator maintains an `std::deque` with references
+to the memory segments that have been returned from `allocate()`.
+
+All requested backing memory allocations (i.e. calls to
+`allocate(size_t requestedSize)`) will round up the allocation size
+to a multiple of the default segment allocation size to reduce
+fragmentation from the raw allocator. For example, the default
+method compilation requests will construct a backing memory allocator
+with a default segment size of 64KB (`1 << 16`), which means all 
+allocations from the backing memory allocator will rounded up to
```
Leonardo2718:
```suggestion
allocations from the backing memory allocator will be rounded up to
```

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+
+When the backing memory object is destructed, it will first free
+all the `TR::MemorySegment`s as well as the raw memory each of
+those segments represents. That memory can also be returned
+immediately by calling `deallocateSegments()`.
+
+The backing memory allocator tracks how much memory has been allocated
+via `_bytesAllocated` which can be queried via `bytesAllocated()`. The
+backing memory allocator can also be directed to enforce an
+allocation limit via a constructor option which sets the `_allocationLimit`
```
Leonardo2718: When you say "constructor option", do you mean "constructor parameter"?

```suggestion
allocation limit via a constructor parameter which sets the `_allocationLimit`
```

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+satisfied by requesting a dedicated segment from the backing
+memory allocator, and such a segment will be immediately sent
+to the backing memory allocator to be deallocated when the
+compiler asks the segment allocator to deallocate it.
+
+To facilitate reuse, the segment allocator maintains a list of
+free segments, which are `TR::MemorySegment`s that have already
+been cut up into the default allocation size. Because all
+allocation requests are rounded up to this size anyway (except
+for larger memory requests), any segment on this free list can
```
Leonardo2718: I find the phrasing of this sentence slightly confusing because, if I understand correctly, larger memory requests *are* rounded up to a *multiple* of the base size. Can the sentence be changed to something like:

```
Because all allocation requests smaller than the base are rounded up, ...
```

?

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+be immediately returned to satisfy any "regular" request, and
+the segment allocator prefers to use such a free segment if
+one is available.
+
+If no such free segment is available, the next possible source
+of memory is _currentBackingMemoryChunk, which is the last
+`TR::MemorySegment` allocated by the backing memory allocator
+that has not been fully consumed yet. If the request can
+be satisfied from this memory segment, then a new
+`TR::MemorySegment` will be created (using the raw allocator)
```
Leonardo2718: Might also be worth mentioning where the raw allocator is coming from. I assume it's the one inside the backing allocator but I'd prefer to have that explicitly stated for clarity.

doc/compiler/memory/MemoryManager.md
```diff
@@ -15,90 +37,378 @@ by the Compiler; there are too many participants for informal trust,
...
+A second new feature is that the segment allocator will try to
+reuse segments that have been passed to `deallocate()` without
+returning them to the backing memory allocator. To avoid the
+overhead of OS level memory allocators, the segment allocator
+tries hard not to allocate more memory (with the consequence
+that it will hold onto allocated memory longer than strictly
+required by the compiler).
+
+This reuse has one restriction to simplify the reuse algorithm:
+only segments smaller than the default segment size (which will
```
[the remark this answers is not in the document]
mstoodle: yes, that's correct

doc/compiler/memory/MemoryManager.md
mstoodle: that makes sense.

doc/compiler/memory/MemoryManager.md
mstoodle: i'll try to reword that phrase to make it less awkward

mstoodle:
There are a couple more changes coming already, inspired by the work in OpenJ9:
1. I have now realized the awkwardness in the `using` construct for mapping an extensible class into the `TR` namespace, so I'll be promoting many of these classes to full OMR_EXTENSIBLE classes and also bring in `self()`
2. some of the fields in OMR classes are marked as `private` but are useful to reference directly in subclasses, so I'll be promoting those fields to `protected`

...probably more to come...

mstoodle:
This pull request no longer reflects the direction I think this work should take. Following the compiler architecture meeting where we discussed this design, I received significant input from others that motivated a somewhat different take. I also gained an improved understanding of how RawAllocator interacts with STL allocators that makes some of this design, as written, problematic. I'm going to close this PR at this point, since it's obviously stale.

Note to any (optimistically presumed) subsequent readers: please feel free to contact me to discuss this work if you're interested :) .

Issue #3397, opened 2018-12-23
Add awareness of Xrdbar operations for Power


IBMJimmyk:
These are Power specific changes.
Update evaluateNULLCHKWithPossibleResolve to know about read barrier ops.
Add default evaluators for xrdbar ops.

For more design details regarding the xrdbar opcodes see:
https://github.com/eclipse/omr/issues/2812

PR #3398, opened 2018-12-23
Add awareness of Xrdbar operations for Power

IBMJimmyk:
Updated evaluateNULLCHKWithPossibleResolve for Power to recognize read
barrier opcodes.

Fixed a bug in the Power version of evaluateNULLCHKWithPossibleResolve
where reference counts were not handled properly. If the child has already
been evaluated, it is not necessary to decrement the reference counts on
the reference node.

Added default xrdbarEvaluators for Power.

TM is not currently compatible with read barriers. If read barriers are
needed, TM is disabled.

Closes: #3397

### DIFF
--- a/compiler/p/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/p/codegen/ControlFlowEvaluator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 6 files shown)


vijaysun-omr:
@genie-omr build all

vijaysun-omr:
I would like reviews from @ymanton and @zl-wang please.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -3856,17 +3858,28 @@ TR::Register *OMR::Power::TreeEvaluator::evaluateNULLCHKWithPossibleResolve(TR::
...
        !firstChild->getSymbolReference()->isUnresolved())
       {
       cg->decReferenceCount(firstChild);
       cg->decReferenceCount(reference);
       }
+   else if (opCode.isLoad() && firstChild->getReferenceCount()==1 &&
+       !firstChild->getSymbolReference()->isUnresolved())
+      {
+      cg->decReferenceCount(firstChild);
+      }
```
ymanton: Can we do this instead?

```
if (opCode.isLoad() && firstChild->getReferenceCount()==1 &&
    !firstChild->getSymbolReference()->isUnresolved())
   {
   cg->decReferenceCount(firstChild);
   if (firstChild->getRegister()==NULL)
      cg->decReferenceCount(reference);
   }
```

IBMJimmyk: Yes, I can make that change.

IBMJimmyk:
I added in the suggested code change and also updated the comment in evaluateNULLCHKWithPossibleResolve.

vijaysun-omr:
@genie-omr build all

zl-wang:
I actually had two sessions of code review about this in the last week.  It is ok to me.

vijaysun-omr:
Looks okay to me too. Merging since checks have passed and reviews are done.

PR #3406, opened 2018-12-27
AArch64: Fix code generation of isub

knn-k:
There is a case where isubEvaluator() for aarch64 generates wrong
"addw" instruction.  This commit fixes it.


knn-k:
Fix is tested with ArithmeticTest.cpp in fvtest/compilertriltest.

0xdaryl:
@genie-omr build aarch64

0xdaryl:
This is an AArch64-specific fix.  A full genie build is not required.

PR #3407, opened 2018-12-27
AArch64: Implement directCallEvaluator()

knn-k:
This commit implements directCallEvaluator() for aarch64.
It also changes the layout of stack frame.

### DIFF
--- a/compiler/aarch64/codegen/ARM64SystemLinkage.cpp
+++ b/compiler/aarch64/codegen/ARM64SystemLinkage.cpp
@@ -24,6 +24,8 @@
 #include "codegen/ARM64Instruction.hpp"
 #include "codegen/GenerateInstructions.hpp"
 #include "codegen/MemoryReference.hpp"
+#include "env/StackMemoryRegion.hpp"
+#include "il/Node_inlines.hpp"
 #include "il/symbol/AutomaticSymbol.hpp"
 
 
@@ -172,13 +174,22 @@ TR::ARM64SystemLinkage::getRightToLeft()
    }
 
 
-static void mapSingleParameter(TR::ParameterSymbol *parameter, uint32_t &stackIndex)
+static void mapSingleParameter(TR::ParameterSymbol *parameter, uint32_t &stackIndex, bool inCalleeFrame)
    {
-   auto size = parameter->getSize();
-   auto alignment = size <= 4 ? 4 : 8;
-   stackIndex = (stackIndex + alignment - 1) & (~(alignment - 1));
-   parameter->setParameterOffset(stackIndex);
-   stackIndex += size;
+   if (inCalleeFrame)
+      {
+      auto size = parameter->getSize();
+      auto alignment = size <= 4 ? 4 : 8;
+      stackIndex = (stackIndex + alignment - 1) & (~(alignment - 1));
+      parameter->setParameterOffset(stackIndex);
+      stackIndex += size;
+      }
+   else
+      { // in caller's frame -- always 8-byte aligned
+      TR_ASSERT((stackIndex & 7) == 0, "Unaligned stack index.");
+      parameter->setParameterOffset(stackIndex);
+      stackIndex += 8;
+      }
    }
 
 
@@ -189,6 +200,8 @@ TR::ARM64SystemLinkage::mapStack(TR::ResolvedMethodSymbol *method)
    uint32_t stackIndex = 0;
    ListIterator<TR::AutomaticSymbol> automaticIterator(&method->getAutomaticList());
    TR::AutomaticSymbol *localCursor = automaticIterator.getFirst();
+
+   stackIndex = 8; // [sp+0] is for link register
 
    // map non-long/double automatics
    while (localCursor != NULL)
@@ -237,7 +250,6 @@ TR::ARM64SystemLinkage::mapStack(TR::ResolvedMethodSymbol *method)
          stackIndex += 8;
          }
       }
-   stackIndex += 8; // for link register
 
    /*
     * Because the rest of the code generator currently expects **all** arguments
@@ -265,7 +277,7 @@ TR::ARM64SystemLinkage::mapStack(TR::ResolvedMethodSymbol *method)
             if (nextIntArgReg < getProperties().getNumIntArgRegs())
                {
                nextIntArgReg++;
-               mapSingleParameter(parameter, stackIndex);
+               mapSingleParameter(parameter, stackIndex, true);
                }
             else
                {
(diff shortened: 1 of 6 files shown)


knn-k:
Arguments passed via the stack is always placed at 8-byte boundary even when their types are 32-bit integer or floating-point.  This is compatible with the AArch64 procedure call standard.
32-bit local variables in the stack are packed at 4-byte boundary to reduce the stack frame size.

0xdaryl:
@genie-omr build aarch64

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
```diff
@@ -516,17 +534,313 @@ int32_t TR::ARM64SystemLinkage::buildArgs(TR::Node *callNode,
...
+   int32_t totalSize = buildArgs(callNode, dependencies);
+   if (totalSize > 0)
+      {
+      if (constantIsUnsignedImm12(totalSize))
+         {
+         generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::subimmx, callNode, sp, sp, totalSize);
+         }
+      else
+         {
+         TR_ASSERT(false, "Too many arguments.");
```
0xdaryl: This should be a fatal assert.

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
```diff
@@ -516,17 +534,313 @@ int32_t TR::ARM64SystemLinkage::buildArgs(TR::Node *callNode,
...
+
+   if (totalSize > 0)
+      {
+      if (constantIsUnsignedImm12(totalSize))
+         {
+         generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::addimmx, callNode, sp, sp, totalSize);
+         }
+      else
+         {
+         TR_ASSERT(false, "Too many arguments.");
```
0xdaryl: This should be a fatal assert.

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
```diff
@@ -516,17 +534,313 @@ int32_t TR::ARM64SystemLinkage::buildArgs(TR::Node *callNode,
...
+                  else
+                     tempReg = cg()->allocateRegister();
+                  generateMovInstruction(cg(), callNode, tempReg, argRegister);
+                  argRegister = tempReg;
+                  }
+               if (numIntegerArgs == 0 &&
+                  (resType.isAddress() || resType.isInt32() || resType.isInt64()))
+                  {
+                  TR::Register *resultReg;
+                  if (resType.isAddress())
```
0xdaryl: The data type is not sufficient to determine if the return value is a collected reference or not (`TR_Address` can be both a managed and unmanaged pointer).  You should be asking the method symbol if it's a collected reference or not and creating the appropriate register based on that answer.

compiler/aarch64/codegen/OMRCodeGenerator.hpp
```diff
@@ -283,14 +283,25 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
...
    TR_GlobalRegisterNumber _gprLinkageGlobalRegisterNumbers[TR::RealRegister::NumRegisters]; // could be smaller
    TR_GlobalRegisterNumber _fprLinkageGlobalRegisterNumbers[TR::RealRegister::NumRegisters]; // could be smaller
 
    private:
 
    enum // flags
       {
-      IsOutOfLineHotPath    = 0x00000200,
+      HasCall               = 0x00000200,
+      IsOutOfLineHotPath    = 0x00004000,
```
0xdaryl: Can these values be numbered a little more compactly?  Nothing functionally wrong with this, but they are interesting choices.  :-)

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
```diff
@@ -516,17 +534,313 @@ int32_t TR::ARM64SystemLinkage::buildArgs(TR::Node *callNode,
...
+                  addDependency(dependencies, argRegister, properties.getIntegerArgumentRegister(numIntegerArgs), TR_GPR, cg());
+                  }
+               }
+            else
+               {
+               // numIntegerArgs >= properties.getNumIntArgRegs()
+               if (childType == TR::Address || childType == TR::Int64)
+                  {
+                  op = TR::InstOpCode::strpostx;
+                  mrefSize = 8;
```
0xdaryl: Why is `mrefSize` tracked?  It's not used anywhere...

compiler/aarch64/codegen/OMRCodeGenerator.hpp
knn-k: I thought it might be better to use the same values as those used in `p` and `arm` versions.

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
knn-k: Originally `mrefSize` was added to `argSize`.  I changed the stack offsets for outgoing arguments, and `argSize` is now always incremented by 8.  Removing the unused variable.

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
knn-k: Hmm.  Then both `p` and `arm` versions have the same potential issue.

compiler/aarch64/codegen/ARM64SystemLinkage.cpp
```diff
@@ -516,17 +534,313 @@ int32_t TR::ARM64SystemLinkage::buildArgs(TR::Node *callNode,
...
+      child = callNode->getChild(i);
+      childType = child->getDataType();
+
+      switch (childType)
+         {
+         case TR::Int8:
+         case TR::Int16:
+         case TR::Int32:
+         case TR::Int64:
+         case TR::Address: // have to do something for GC maps here
```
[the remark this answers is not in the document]
knn-k: I copied the comment from the `p` version, and just forgot to remove it.
Neither `p` nor `arm` versions do anything special for GC maps in this function.

knn-k:
Re: collected references with `resType.isAddress()` in one of the review comments:
Shall I open an issue to track it for not only aarch64, but p and arm?

0xdaryl:
> Shall I open an issue to track it for not only aarch64, but p and arm?

Yes please.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3411, opened 2019-01-02
Remove early return for OMRPORT_VMEM_ALLOC_QUICK

dmitripivkine:
Theoretically smartAddress might be NULL in case when allocation is
impossible only. However we do have a few customers report
findAvailableMemoryBlockNoMalloc() returns NULL intermittently when slow
allocation path succeeds. We can not reproduce problem in house so the
reason is unknown for now. Service is working on details. In order to
prevent possible failures remove early return to allow slow allocation
path unconditionally temporary until findAvailableMemoryBlockNoMalloc()
is fixed.

### DIFF
--- a/port/linux/omrvmem.c
+++ b/port/linux/omrvmem.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@dmitripivkine this issue here may be related to the issue that required this change to be released:
https://github.com/google/sanitizers/issues/708

dmitripivkine:
> @dmitripivkine this issue here may be related to the issue that required this change to be released:
> [google/sanitizers#708](https://github.com/google/sanitizers/issues/708)

lines in `maps` look like:
```
01b02000-01b23000 rw-p 00000000 00:00 0                                  [heap]
01b23000-04d23000 rw-p 00000000 00:00 0
04d23000-0e323000 ---p 00000000 00:00 0
```
This issue is about missed p/s and the end of permissions. We are not parsing line so far. We are looking for space between records like ranges [previous end, current start] so we read `start address`, skip `-` and read `end address' and nothing more from the line.

PR #3365, opened 2018-12-18
WIP: Add C version of the Simple example to Jitbuilder sample programs

daytonallen:
The C version of the Simple example in this commit serves as a proposal
of how the C API would work. The main aim is to gain community feedback
and foster discussions surrounding the ergonomics of the API.

Issue: #3197
Signed-off-by: Dayton J Allen daytonjallen@gmail.com

Issue #3197, opened 2018-11-14
Write JitBuilder C API generator


Leonardo2718:
As a step towards addressing #2049, the following work needs to be completed:

- [ ] Refactor C++ generator so code can be shared more easily with C generator
  - [ ] Create a "generator" class
  - [ ] Create a source file writer class
  - [ ] Common "generator driving" code
- [ ] Migrate  C++ JitBuilder examples to C
- [ ] Write generator for C
  - [ ] Generate header files with forward declarations
  - [ ] Generate source files with implementation

daytonallen:
Hey, just to get the discussion started on what the C API implementation should look like, I've gone ahead and translated the _Simple.cpp_ & _Simple.hpp_ examples from the test suite to C as a proposal of how the API might look.


_Simple.h_
``` c
#ifndef SIMPLE_H
#define SIMPLE_H

#include "Jitbuilder.h"

typedef struct SimpleMethod {
    /* Struct embedding for inheritance? */
    MethodBuilder *methodBuilder;
} SimpleMethod;

SimpleMethod* SimpleMethod_init(TypeDictionary *);
bool SimpleMethod_buildIL(SimpleMethod *);

#endif SIMPLE_H
```

_Simple.c_
```c
#include "Simple.h"

#include <stdio.h>
#include <stdbool.h>
#include <stdlib.h>

#define TOSTR(x) #x
#define LINETOSTR(x) TOSTR(x)

int main(int argc, char *argv[]) {
    puts("Step 1: initialize JIT");
    bool initialized = JB_initializeJit();
    if (!initialized) {
        fprintf(stderr, "FAIL: could not initialized JIT\n");
        exit(-1);
    }

    puts("Step 2: define type dictionary");
    TypeDictionary *types;
    puts("Step 3: compile method builder");
    SimpleMethod *method = SimpleMethod_init(types);

    void *entry = 0;
    int32_t rc = JB_compileMethodBuilder(method->methodBuilder, &entry);
    if (rc != 0) {
        fprintf(stderr, "FAIL: compilation error %d\n", rc);
        exit(-2);
    }

    puts("Step 4: invoke compiled code and print results\n");
    typedef int32_t (SimpleMethodFunction)(int32_t);
    SimpleMethodFunction *increment = (SimpleMethodFunction *) entry;

    int32_t v;
    v = 0;
    printf("increment(%d) == %d\n", v, increment(v));

    v = 1;
    printf("increment(%d) == %d\n", v, increment(v));

[...]

PR #3412, opened 2019-01-02
WIP: Implement JitBuilder C API generator

Leonardo2718:
These commits implement an initial C API generator for JitBuilder. Unit tests are added for the new generator and some refactoring is done to enhance code-reuse between the C++ generator and the C generator.

While I welcome initial reviews, the following items still need to be worked on before this PR is ready for merging:

- naming needs to be updated according to the discussions in #3365 and #3197 
- figure out how to best handle default callbacks and setting callbacks
- properly test using C JitBuilder examples

### DIFF
--- a/jitbuilder/apigen/cgen.py
+++ b/jitbuilder/apigen/cgen.py
@@ -1,0 +1,567 @@
+#! /usr/bin/env python
+
+###############################################################################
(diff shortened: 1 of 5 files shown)


fjeremic:
Do we plan on making further progress here? If not we should close the PR.

PR #3413, opened 2019-01-02
Use numeric_limits, not stdint.h limits

rwy0717:
The stdint.h limits aren't available without a macro definition on older
compilers, but this behaviour is deprecated as-of C++11. Avoid the whole
mess and stick with numeric_limits, the C++  way to get a maximum.


Leonardo2718:
@genie-omr build all

Leonardo2718:
For good measure

@genie-omr build all

PR #3415, opened 2019-01-02
Heap size startup hints

amicic:
Some pre-req work to support Heap size startup hints feature in OpenJ9
- swapped initial order of allocate and survivor in Nursery Physical
Arena to be able to expand early
- introduce new reasons for expansion, which would be reported by
verbose GC during early startup
- introduce two tuning parameters, that will control how aggressively
the feature is used. heapSizeStatupHintWeightNewValue is for now set to
0 (which effectively means the feature is disabled), but plan is to used
more value of 0.8 or so, after some more testing is peformed


charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

PR #3417, opened 2019-01-04
Recover x86 jit builder testing

mstoodle:
I recently discovered that many of our extended JitBuilder sample tests (which cannot be run safely on all platforms) are not being run on the x86 platforms where they can run safely. On top of that, one of the samples (OperandStackTests) was broken.

I also found that the JitBuilder API description for VirtualMachineOperandStack constructors was not quite right (and we probably didn't discover that because the tests weren't being run).

This PR includes a fix for the OperandStackTests failure, a fix for the API description, and the extended JitBuilder testing is enabled for X86 platforms so hopefully we won't end up in this state again.

Finally, I also fixed a couple more compiler warnings (one in the optimizer, which I *believe* is the last exposed warning in the entire optimizer, which is nice, and one of many more to be fixed in the code generator).

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -62,6 +62,13 @@ endif()
 ## Do NOT force it since it is explicitly disabled on Windows for now.
 if(OMR_JITBUILDER)
 	set(OMR_JITBUILDER_TEST ON CACHE BOOL "")
+
+	## Enable additional JitBuilder tests if running on a supported platform
+        #  (which currently means 64-bit x86 platforms except for Windows)
+	if((OMR_HOST_ARCH STREQUAL "x86" AND NOT OMR_OS_WINDOWS) AND OMR_TEMP_DATA_SIZE STREQUAL "64")
+    		set(OMR_JITBUILDER_TEST_EXTENDED ON)
+	endif()
+
 else()
     # if JitBuilder isn't enabled, the tests can't be built
     set(OMR_JITBUILDER_TEST OFF CACHE BOOL "" FORCE)
--- a/compiler/optimizer/GlobalRegisterAllocator.cpp
+++ b/compiler/optimizer/GlobalRegisterAllocator.cpp
@@ -1433,7 +1433,7 @@ TR_GlobalRegisterAllocator::transformNode(
          // internal pointer must always be present on the stack.
          //
          //
-         if (!fpStore && !symRef->storeCanBeRemoved()
+         if ((!fpStore && !symRef->storeCanBeRemoved())
              || (comp()->getOption(TR_MimicInterpreterFrameShape) && comp()->getOption(TR_FSDGRA)))
             {
             gr->setValue(0);
@@ -3961,9 +3961,9 @@ TR_GlobalRegisterAllocator::markAutosUsedIn(
          }
 
       if (!canSkip &&
-          ((node->getOpCodeValue() == TR::iadd) || (node->getOpCodeValue() == TR::isub) ||
-           (node->getOpCodeValue() == TR::iand) || (node->getOpCodeValue() == TR::iloadi) || (node->getOpCodeValue() == TR::ixor)) ||
-          ((node->getOpCodeValue() == TR::iload) /* && !node->getSymbolReference()->getSymbol()->isAutoOrParm() */))
+          (((node->getOpCodeValue() == TR::iadd) || (node->getOpCodeValue() == TR::isub) ||
+            (node->getOpCodeValue() == TR::iand) || (node->getOpCodeValue() == TR::iloadi) || (node->getOpCodeValue() == TR::ixor)) ||
+           ((node->getOpCodeValue() == TR::iload) /* && !node->getSymbolReference()->getSymbol()->isAutoOrParm() */)))
          {
          TR::Node *storeNode = NULL;
          TR::Node *loadNode = NULL;
@@ -4290,9 +4290,7 @@ TR_GlobalRegisterAllocator::markAutosUsedIn(
       TR::CFG *cfg = comp()->getFlowGraph();
 
       TR::SymbolReference *symRef = node->getSymbolReference();
-      if ((symRef->getSymbol()->isAutoOrParm() ||
-           (0) && symRef->getSymbol()->getDataType() != TR::NoType)
-          && isSymRefAvailable(symRef, blocksInLoop))
+      if (symRef->getSymbol()->isAutoOrParm() && isSymRefAvailable(symRef, blocksInLoop))
          {
          TR_RegisterCandidate *rc = registerCandidates[symRef->getReferenceNumber()];
          if (!rc)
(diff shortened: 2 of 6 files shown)


mstoodle:
genie-omr build all

mstoodle:
Oh, almost forgot: credit to @Leonardo2718 for suggesting the cmake update to re-introduce the tests! Thanks, Leonardo!

jitbuilder/release/cpp/samples/OperandStackTests.cpp
```diff
@@ -585,7 +585,8 @@ OperandStackTestUsingStructMethod::OperandStackTestUsingStructMethod(OMR::JitBui
    : OperandStackTestMethod(d)
    {
    d->DefineStruct("Thread");
-   d->DefineField("Thread", "sp", d->PointerTo(d->PointerTo(STACKVALUEILTYPE)), offsetof(Thread, sp));
+   //d->DefineField("Thread", "sp", d->PointerTo(d->PointerTo(STACKVALUEILTYPE)), offsetof(Thread, sp));
```
charliegracie: Can you remove the commented out line?

cmake/config.cmake
```diff
@@ -62,6 +62,12 @@ endif()
 ## Do NOT force it since it is explicitly disabled on Windows for now.
 if(OMR_JITBUILDER)
 	set(OMR_JITBUILDER_TEST ON CACHE BOOL "")
+
+	## Enable additional JitBuilder tests if running on a supported platform
+	if(OMR_HOST_ARCH STREQUAL "x86" AND OMR_TEMP_DATA_SIZE STREQUAL "64")
+    		set(OMR_JITBUILDER_ADDITIONAL ON)
```
charliegracie: I am not sure this works on Windows

charliegracie: Also since we are making changes here could we change the name of this flag to something like OMR_JITBUILDER_TEST_EXTENDED or something similar so it is more clear that it is controlling tests and not JITBuilder features?

jitbuilder/release/cpp/samples/OperandStackTests.cpp
mstoodle: bah, yes of course....that was poor cleanup on my part :(

mstoodle:
@charliegracie @Leonardo2718 this one is ready for another look at

Leonardo2718:
There's a typo in the first commit message: `parmaeters` &rarr; `parmeters`.

charliegracie:
@genie-omr build all

compiler/optimizer/GlobalRegisterAllocator.cpp
```diff
@@ -3961,9 +3961,9 @@ TR_GlobalRegisterAllocator::markAutosUsedIn(
          }
 
       if (!canSkip &&
-          ((node->getOpCodeValue() == TR::iadd) || (node->getOpCodeValue() == TR::isub) ||
-           (node->getOpCodeValue() == TR::iand) || (node->getOpCodeValue() == TR::iloadi) || (node->getOpCodeValue() == TR::ixor)) ||
-          ((node->getOpCodeValue() == TR::iload) /* && !node->getSymbolReference()->getSymbol()->isAutoOrParm() */))
+          (((node->getOpCodeValue() == TR::iadd) || (node->getOpCodeValue() == TR::isub) ||
+            (node->getOpCodeValue() == TR::iand) || (node->getOpCodeValue() == TR::iloadi) || (node->getOpCodeValue() == TR::ixor)) ||
+           ((node->getOpCodeValue() == TR::iload) /* && !node->getSymbolReference()->getSymbol()->isAutoOrParm() */)))
```
Leonardo2718: If I'm matching these parentheses correctly, the new code is not equivalent to the old one because the `!canSkip &&` did not apply to the `(node->getOpCodeValue() == TR::iload)`. Looking through the rest of the code quickly, I _think_ the fix is correct. I'd like confirmation that the change was intentional, since the commit message says "These changes fix no bugs, just eliminate some warnings".

Leonardo2718:
@mstoodle ping?

compiler/optimizer/GlobalRegisterAllocator.cpp
mstoodle: it was intentional (I forgot about it when I updated the comment <blush>).

@andrewcraik @vijaysun would you please confirm that you agree with this change, please?

mstoodle:
I also fixed the typo in the first commit message

compiler/optimizer/GlobalRegisterAllocator.cpp
andrewcraik: It is not equivalent in that canSkip did not apply to TR::iload and yet canSkip is initialized to false and the entire if that could set it to true is off due to `0 &&`. From a functional perspective right now this should be fine. @vijaysun-omr may have better context to know if there is a potential problem, but it does look ok to me in the context.

Leonardo2718:
@genie-omr build all

compiler/optimizer/GlobalRegisterAllocator.cpp
mstoodle: ok, had a discussion with @andrewcraik and @vijaysun on this code change. It turns out the change is actually irrelevant due to other disabled code blocks (so `canSkip` is actually false for this if condition). There are other significant blocks of code in this function that have been effectively disabled in some distant past investigation and nobody knows of any plan to re-enable. Vijay has promised to take a gander through that code to see if he believes there is value in keeping any of it. If not, then I'll adapt my commit to just rip out the dead code to simplify the function, which should have the side benefit of also eliminating the warning :) .

vijaysun-omr: I think the main piece of sign extension code that I know was used (and still seems to be enabled) is at    

``` if (!enableNewSignExtGRA && /* !hasCatchBlock && */ node->getOpCode().isLoadVarDirect() && ```

The snippet of code under discussion here is from some prototype that was never enabled (I don't remember details now) but I think it's safe to delete the code for "enableNewSignExtGRA" being true in my opinion because it was never enabled and given that it's several years since we last touched this area of the codebase, I don't think it needs to clutter the codebase.

mstoodle:
removed the `enableNewSignExtGR` (which was quite therapeutic) and pushed. I was lazy (ok, really I pressed return too fast) and didn't build it first, so *possibly* the changes will break the build. Let's hold off on additional jenkins testing until the Travis build comes back successful.

Leonardo2718:
@mstoodle Looks like both Travis and AppVeyor builds completed. :slightly_smiling_face:

mstoodle:
rebased and pushed. hopefully this one is very close now

mstoodle:
@genie-omr build all

vijaysun-omr:
@mstoodle the cleanup of ```enableNewSignExtGRA=true``` code looks good. There appear to be some changes requested based on the reviews done. I will look at this again once those changes have been done.

mstoodle:
I believe those changes have now been completed. @Leonardo2718 would you please revisit to see if I got everything you pointed out?

PR #3419, opened 2019-01-05
Percolate on exhausted Concurrent Mark

amicic:
Introduce new Scavenge to Global GC percolate event/reason.

If Concurrent Mark work is exhausted and about to transition to final
STW phase, and at the same time Allocate in Nursery is depleted so that
Scavenge is triggered, skip doing the Scavenge and go straight to Global
GC. That Global would occur anyway immediately after the Scavenge and
since it will sweep Nursery, after which Allocation Failure will be
likely be satisfied, there is no point in doing the Scavenge before that
STW Global.

An example, as observed from verbose GC:
```
<af-start id="2154" threadId="0000000001D0A488"
totalBytesRequested="520" timestamp="2019-01-04T18:39:56.593"
intervalms="64.692" type="nursery" />   <<<<<<<<<<<<<< AF in Nursery
<percolate-collect id="2155" from="nursery" to="global"
reason="concurrent mark exhausted" timestamp="2019-01-04T18:39:56.593"/>
<gc-start id="2156" type="global" contextid="2151"
timestamp="2019-01-04T18:39:56.593">  <<<<<<<<<<<<<<<< percolating to
Global instead
  <mem-info id="2157" free="57489872" total="434503680" percent="13">
    <mem type="nursery" free="0" total="141295616" percent="0">
      <mem type="allocate" free="0" total="70647808" percent="0" />
<<<<<<<<<<< indeed depleted Nursery triggered Scavenge
      <mem type="survivor" free="0" total="70647808" percent="0" />
    </mem>
...
  </mem-info>
</gc-start>
....
<concurrent-halted id="2158" timestamp="2019-01-04T18:39:56.593">
  <halted state="Complete" status="exhausted" /> ++++ indeed Concurrent
Mark work was exhausted ++++++
  <traced bytesTarget="107803108" bytesTotal="129982220"
bytesByMutator="126991328" bytesByHelper="2990892" percent="120" />
  <cards cleaned="130" thresholdBytes="5688730" />
</concurrent-halted>
....
<cycle-end id="2166" type="global" contextid="2151"
timestamp="2019-01-04T18:39:56.621" />
<allocation-satisfied id="2167" threadId="0000000001D09B00"
bytesRequested="520" />
<af-end id="2168" timestamp="2019-01-04T18:39:56.621"
threadId="0000000001D0A488" success="true" from="nursery"/> <<<<<<<<<<
original AF (same thread, same size) indeed satisfied from Nursery
```


charliegracie:
@genie-omr build all

charliegracie:
I think this is a net win.  There may be an edge case where the scavenge may still be required directly following the global but I agree that in most cases the scavenge is not required.  This should improve STW times.

PR #3420, opened 2019-01-07
Added portlibrary global to make isRunningInContainer called at startup

bharathappali:
OMR PortLibrary Api `omrsysinfo_is_running_in_container` is required by every runtime built on top of OMR and also might be having many places where this would be called, As this api internally calls other function, its get invoked everytime we make a call to the api so i thought it would be good to have a PortLibrary global variable `PPG_isRunningInContainer` for is running in container check and also that function to be called at the portLibrary startup which assigns the value for the portlibrary global `PPG_isRunningInContainer` and would decrease the calls for `isRunningInContainer` function.

### DIFF
--- a/port/common/omrsysinfo.c
+++ b/port/common/omrsysinfo.c
@@ -373,6 +373,7 @@ omrsysinfo_shutdown(struct OMRPortLibrary *portLibrary)
 int32_t
 omrsysinfo_startup(struct OMRPortLibrary *portLibrary)
 {
+	PPG_isRunningInContainer = FALSE;
 	return 0;
 }
 
@@ -955,14 +956,12 @@ omrsysinfo_get_cgroup_subsystem_list(struct OMRPortLibrary *portLibrary)
  *
  * @param[in] portLibrary pointer to OMRPortLibrary
  *
- * @param[in] errorCode int32_t pointer to state error code from internal calls
- *
  * @return TRUE if Runtime is running in a container and FALSE if not or if an error occurs
  */
 BOOLEAN
-omrsysinfo_is_running_in_container(struct OMRPortLibrary *portLibrary, int32_t *errorCode)
+omrsysinfo_is_running_in_container(struct OMRPortLibrary *portLibrary)
 {
-	return FALSE;
+	return PPG_isRunningInContainer;
 }
 
 /**
--- a/port/unix/omrsysinfo.c
+++ b/port/unix/omrsysinfo.c
@@ -2040,6 +2040,7 @@ omrsysinfo_shutdown(struct OMRPortLibrary *portLibrary)
 int32_t
 omrsysinfo_startup(struct OMRPortLibrary *portLibrary)
 {
+	PPG_isRunningInContainer = FALSE;
 	/* Obtain and cache executable name; if this fails, executable name remains NULL, but
 	 * shouldn't cause failure to startup port library.  Failure will be noticed only
 	 * when the omrsysinfo_get_executable_name() actually gets invoked.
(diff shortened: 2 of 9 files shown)


port/unix/omrsysinfo.c
```diff
@@ -4476,11 +4477,11 @@ omrsysinfo_get_cgroup_subsystem_list(struct OMRPortLibrary *portLibrary)
  * Returns TRUE if running inside a container and FALSE if not or if an error occurs
  */
 BOOLEAN
-omrsysinfo_is_running_in_container(struct OMRPortLibrary *portLibrary, int32_t *errorCode)
+omrsysinfo_is_running_in_container(struct OMRPortLibrary *portLibrary)
 {
 	BOOLEAN inContainer = FALSE;
 #if defined(LINUX) && !defined(OMRZTPF)
-	*errorCode = isRunningInContainer(portLibrary, &inContainer);
+	inContainer = PPG_isRunningInContainer;
```
ashu-mehra: This function can just return `PPG_isRunningInContainer`. I don't think there is any need for `inContainer` here.

bharathappali:
As the current implementation is making the result of `isRunningInContainer` global and returned in `omrsysinfo_is_running_in_container`, to have a error code sent back it again needs a global variable to record it at startup which is not that required. So to make the functionality simple we could just return either `TRUE` or `FALSE` for `omrsysinfo_is_running_in_container`, if there was any error in the call for `isRunningInContainer` it makes the `BOOLEAN` var passed to it as `FALSE`. So returning TRUE or FALSE could make it simple and easy to use.

Please provide your views on it @ashu-mehra . Thanks!

port/unix/omrsysinfo.c
charliegracie: Would it make sense to have the global on all platforms and always just return the global?  Platforms that do not support containers can just always set the global to false.  This way in the future if other platforms do support containers there would be less code to change.

port/common/omrsysinfo.c
```diff
@@ -960,7 +960,7 @@ omrsysinfo_get_cgroup_subsystem_list(struct OMRPortLibrary *portLibrary)
  * @return TRUE if Runtime is running in a container and FALSE if not or if an error occurs
```
DanHeidinga:
```c
 * @param[in] errorCode int32_t pointer to state error code from internal calls
```
Needs to be removed as well

bharathappali: Removed description for `errorCode` param.

port/unix/omrsysinfo.c
bharathappali: @charliegracie That sounds good to me, Will be making the changes. Thanks !

port/unix/omrsysinfo.c
```diff
@@ -2058,6 +2059,7 @@ omrsysinfo_startup(struct OMRPortLibrary *portLibrary)
 		}
 	}
 	attachedPortLibraries += 1;
+	isRunningInContainer(portLibrary, &PPG_isRunningInContainer);
```
charliegracie: Would it make more sense for this code to look like this:
`PPG_isRunningInContainer = isRunningInContainer(portLibrary);`

charliegracie: Any update on this question?

bharathappali: Sorry for delayed reply @charliegracie I have missed to reply to this question earlier.

`isRunningInContainer` is an internal method which is used by many functions and some of them (now or in future) might need the `return code` of it. So I'm thinking it would be better to have the current implementation to have the error code handy. Please let me know your views on it. Thanks!

charliegracie: No problem :). Ah I missed the other callers.  Since the other callers are using the return value I think it makes sense to leave the code as is.  Thanks!

charliegracie:
@genie-omr build all

bharathappali:
@charliegracie [continuous-integration](https://ci.eclipse.org/omr/job/PullRequest-linux_ppc-64_le_gcc/1087/console) is failing to fetch the origin. Will triggering the build again solves the problem ?

charliegracie:
@genie-omr build plinux

PR #3421, opened 2019-01-07
Implement dynamic array clone in value propagation

yanluo7:
Dynamic array cloning transforms refArrayObject.clone()
into anewarray followed by arraycopy, where refArrayObject
is a resolved, not fixed array class. This is safe as cloning
of an array object can safely be devirtualized and replaced with
a new array allocation followed by an array copy. The caveat is
that since the array class is not fixed, the actual class type of the
cloned array cannot be known at compile time, hence the class
child of the anewarray opcode needs to be dynamically evaluated
at runtime, instead of being represented as a runtime constant under
normal, statically allocated anewarray cases.

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.hpp
+++ b/compiler/optimizer/OMRValuePropagation.hpp
@@ -600,10 +600,19 @@ class ValuePropagation : public TR::Optimization
          : _clazz(clazz), _isFixed(isFixed)  { }
    };
 
+   struct ArrayCloneInfo {
+      TR_ALLOC(TR_Memory::ValuePropagation)
+
+      TR_OpaqueClassBlock *_clazz;
+      bool _isFixed;
+      ArrayCloneInfo(TR_OpaqueClassBlock *clazz, bool isFixed)
+         : _clazz(clazz), _isFixed(isFixed)  { }
+   };
+
 #ifdef J9_PROJECT_SPECIFIC
    void transformConverterCall(TR::TreeTop *);
    void transformObjectCloneCall(TR::TreeTop *, ObjCloneInfo *cloneInfo);
-   void transformArrayCloneCall(TR::TreeTop *, TR_OpaqueClassBlock *j9class);
+   void transformArrayCloneCall(TR::TreeTop *, ArrayCloneInfo *cloneInfo);
 #endif
 
 
@@ -905,7 +914,7 @@ class ValuePropagation : public TR::Optimization
    List<TR::TreeTop> _objectCloneCalls;
    List<TR::TreeTop> _arrayCloneCalls;
    List<ObjCloneInfo> _objectCloneTypes;
-   List<TR_OpaqueClassBlock> _arrayCloneTypes;
+   List<ArrayCloneInfo> _arrayCloneTypes;
 
    int32_t    *_parmInfo;
    bool       *_parmTypeValid;
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -5762,6 +5762,8 @@ TR::Node *constrainAcall(OMR::ValuePropagation *vp, TR::Node *node)
       if (!node->getOpCode().isIndirect())
          {
          static char *enableDynamicObjectClone = feGetEnv("TR_enableDynamicObjectClone");
+         // Dynamic cloning kicks in when we attempt to make direct call to Object.clone  
+         // or J9VMInternals.primitiveClone where the cloned object is an array.
          if (method->getRecognizedMethod() == TR::java_lang_Object_clone
              || method->getRecognizedMethod() == TR::java_lang_J9VMInternals_primitiveClone)
             {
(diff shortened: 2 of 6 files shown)


yanluo7:
The desired tree after transformation looks like following:
Note the aloadi <componentClass> child of anewarray node where we obtain the component class from the class of the receiver array object of object.clone() :

treetop ()
  anewarray  jitANewArray
    iloadi  <contiguous-array-size>
      ==>l2a (array object to be cloned)
    aloadi  <componentClass>
      ==>aloadi (vftload of array object)
treetop ()
  arraycopy  <arraycopy>
    ==>l2a
    ==>anewarray
    aladd
      ==>l2a
      lconst 8
    aladd 
      ==>anewarray
      ==>lconst 8
    lshl 
       i2l 
         ==>iloadi
       iconst 2

yanluo7:
This PR is dependent on https://github.com/eclipse/openj9/pull/4179 for X86 codegen support.

vijaysun-omr:
Review from @jdmpapin or @andrewcraik would be good.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -5802,22 +5804,43 @@ TR::Node *constrainAcall(OMR::ValuePropagation *vp, TR::Node *node)
...
                       && !vp->_objectCloneCalls.find(vp->_curTree))
                      {
+                     if (vp->trace())
+                        traceMsg(vp->comp(), "Object Clone: Resolved Class of node %p object clone\n", node);
                      vp->_objectCloneCalls.add(vp->_curTree);
                      vp->_objectCloneTypes.add(new (vp->trStackMemory()) OMR::ValuePropagation::ObjCloneInfo(constraint->getClass(), false));
                      }
+                  // Currently enabled for X86 as the required codegen support is implemented on X86 only.
+                  // Remove the condition as other platforms receive support.
+                  else if (TR::Compiler->target.cpu.isX86()
```
vijaysun-omr: Is it better to have a "capability" based query in the codegen object here instead of checking that we are on X86 explicitly ? When the codegens other than X86 implement the necessary function they can return true to the capability based query and this code won't have to change.

vijaysun-omr:
@genie-omr build all

andrewcraik:
@jdmpapin could you have a look please?

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -3548,11 +3548,18 @@ void OMR::ValuePropagation::transformArrayCloneCall(TR::TreeTop *callTree, TR_Op
...
+
    TR_OpaqueClassBlock *j9class = comp()->fe()->getComponentClassFromArrayClass(j9arrayClass);
 
    TR::DebugCounter::prependDebugCounter(comp(), TR::DebugCounter::debugCounterName(comp(), "inlineClone.location/array/(%s)", comp()->signature()), callTree);
    int32_t classNameLength;
    char *className = TR::Compiler->cls.classNameChars(comp(), j9arrayClass, classNameLength);
+   if (TR::Compiler->cls.isPrimitiveClass(comp(), j9class))
+      {
+      TR_ASSERT_FATAL(isFixedClass, "Primitive array should be of fixed constraint");
+      }
```
jdmpapin: I agree that for arrays of primitives we should generate fixed class constraints, but I'm not sure that a mere bound should be considered _wrong_ per se. We can conclude right here that a primitive element type implies `isFixedClass`, whether or not the original constraint told us so.

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -600,10 +600,19 @@ class ValuePropagation : public TR::Optimization
          : _clazz(clazz), _isFixed(isFixed)  { }
    };
 
+   struct ArrayCloneInfo {
```
jdmpapin: Is this needed? It's exactly the same as `ObjCloneInfo`

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -3587,9 +3594,21 @@ void OMR::ValuePropagation::transformArrayCloneCall(TR::TreeTop *callTree, TR_Op
...
+      else
+         {
+         // Load the component class of the cloned array as 2nd child of anewarray as expected by the opcode.
+         TR::Node * arrayClassNode = TR::Node::createWithSymRef(callNode, TR::aloadi, 1, objNode, comp()->getSymRefTab()->findOrCreateVftSymbolRef());
+         classNode = TR::Node::createWithSymRef(callNode, TR::aloadi, 1, arrayClassNode, comp()->getSymRefTab()->findOrCreateArrayComponentTypeSymbolRef());
+         }
       TR::SymbolReference *symRef = comp()->getSymRefTab()->findOrCreateANewArraySymbolRef(objNode->getSymbolReference()->getOwningMethodSymbol(comp()));
-      TR::Node::recreateWithoutProperties(callNode, TR::anewarray, 2, lenNode, loadaddr, symRef);
+      TR::Node::recreateWithoutProperties(callNode, TR::anewarray, 2, lenNode, classNode, symRef);
+      //YAN
```
jdmpapin: Stray comment

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -3587,9 +3594,21 @@ void OMR::ValuePropagation::transformArrayCloneCall(TR::TreeTop *callTree, TR_Op
...
+         }
+      else
+         {
+         // Load the component class of the cloned array as 2nd child of anewarray as expected by the opcode.
+         TR::Node * arrayClassNode = TR::Node::createWithSymRef(callNode, TR::aloadi, 1, objNode, comp()->getSymRefTab()->findOrCreateVftSymbolRef());
+         classNode = TR::Node::createWithSymRef(callNode, TR::aloadi, 1, arrayClassNode, comp()->getSymRefTab()->findOrCreateArrayComponentTypeSymbolRef());
+         }
       TR::SymbolReference *symRef = comp()->getSymRefTab()->findOrCreateANewArraySymbolRef(objNode->getSymbolReference()->getOwningMethodSymbol(comp()));
-      TR::Node::recreateWithoutProperties(callNode, TR::anewarray, 2, lenNode, loadaddr, symRef);
+      TR::Node::recreateWithoutProperties(callNode, TR::anewarray, 2, lenNode, classNode, symRef);
```
jdmpapin: I was concerned about the behaviour of `anewarray` given a primitive as the component type (exception? UB?). But in order to reach this point, the cloned object must have been constrained to be a reference array. If the object could be either a reference array or a primitive array, then the type bound would not have been an array type, and `isArray()` would have been `TR_maybe`. So for the purpose of this change the primitive `anewarray` question is moot :+1:

compiler/optimizer/OMRValuePropagation.hpp
yanluo7: I thought since the related code seems to have a naming convention in distinction between "object" and "array", like `transformObjectCloneCall` and `transformArrayCloneCall`, adding the `ArrayCloneInfo` struct would lend bit more clarity and consistency. But yes, the structs are the currently same.

vijaysun-omr:
@genie-omr build all

yanluo7:
This PR requires https://github.com/eclipse/openj9/pull/4179 being merged first.

vijaysun-omr:
Is that dependency not possible to break by not enabling the new function on X86 in this PR ?

vijaysun-omr:
@genie-omr build all

yanluo7:
@vijaysun-omr If we let the capability query return false on x86, that should break the dependency.

compiler/optimizer/ValuePropagationCommon.cpp
yanluo7: Removed this assert.

andrewcraik:
just waiting on eclipse/openj9#4179 to finish sanity

andrewcraik:
eclipse/openj9#4179 has merged - I believe this is ready to merge

vijaysun-omr:
Looks like we have checks passing and reviews are done. Merging this one.

PR #3425, opened 2019-01-08
AArch64: Add is64bit flag to generateMovInstruction()

knn-k:
Caller of generateMovInstruction() should pass is64bit flag to the
function.  It is true by default.


knn-k:
It is sometimes inappropriate to judge the required bit size from `node->getDataType().isInt64()`.

0xdaryl:
@genie-omr build aarch64

0xdaryl:

PR #3427, opened 2019-01-08
Add a direct method call relocation and clean up some related code

ymanton:
This PR adds a new relocation identifier, `TR_MethodCallAddress` and uses it during BinaryEncoding on x86 to create the corresponding project-specific relocation on direct call instructions. The implementation will be downstream, but the high level description is that it will be used to relocate direct calls where the call target is known at compile time but the address of the call instruction is not, thereby necessitating the use of a relocation to generate the appropriate relative offset for the call once its position is known.

Also included are some self-explanatory cleanups.

FYI @mpirvu.

### DIFF
--- a/compiler/codegen/OMRAheadOfTimeCompile.cpp
+++ b/compiler/codegen/OMRAheadOfTimeCompile.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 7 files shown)


Leonardo2718:
The changes seems reasonable to me.

I don't quite see why the implementation needs to be downstream. From the description given, it seems like a general implementation usable by most downstream projects should be possible. I can see uses for this new relocation in OMR, so I'd like to understand what is limiting the implementation.

ymanton:
The implementation is downstream because all of the other implementations are downstream. :grin:

What exists at the OMR level is the compile-time representation of external relocations of various kinds (i.e. the `TR::ExternalRelocation` class and the `TR_ExternalRelocationTargetKind` enum) and their creation at binary encoding time, but that alone isn't enough. There's no code in the OMR compiler to emit, consume, or apply any of these external relocations, that's all currently done per-downstream project.

Leonardo2718:
> The implementation is downstream because all of the other implementations are downstream. :grin:

I see. I suppose that's a fair reason.

I would still like to see this functionality added to OMR eventually because there is a need for it.

That's said, I'm not familiar enough with the code to know whether it makes sense to do that in this change or whether it is even feasible. So, I won't push to hard for it here. :slightly_smiling_face:

ymanton:
Copyrights have been updated.

vijaysun-omr:
Is the addition of this direct call relocation something that non-X86 ones code generators need to be aware of ? I want to just understand if this is a functional requirement; maybe direct calls have to be treated as unresolved without this relocation and if so, it won't be a functional requirement...more of a way to improve AOT performance when done on the other codegens as well (?)

ymanton:
No, other code generators and will not need to be aware of it, they need to explicitly create this relocation during their binary encoding passes. And yes, direct calls are currently treated as unresolved by all of the OpenJ9 codegens in AOT mode, so they won't ever reach this code unless/until they decide to take advantage of it.

vijaysun-omr:
fyi @dsouzai @r30shah @gita-omr @fjeremic of this change and the AOT use case.

dsouzai:
> And yes, direct calls are currently treated as unresolved by all of the OpenJ9 codegens in AOT mode, so they won't ever reach this code unless/until they decide to take advantage of it.

Actually, in AOT w/ SVM, direct calls are treated as a resolved interpreted call; with the SVM, nothing is implicitly unresolved under AOT.

PR #3423, opened 2019-01-08
Tril tests for fabs, fneg, etc.

knn-k:
This commit adds Tril test cases for fabs, fneg, and their double-
precision versions.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #3429, opened 2019-01-09
Tril tests: Use floating-point abs function in filters

knn-k:
I have used the abs() function in some test case filters for
floating-point workaround, but that is an integer function.
This commit corrects it by replacing with std::abs().


knn-k:
The filter added in #3423 already uses `fabs()`.

Leonardo2718:
Ok, given my comment in https://github.com/eclipse/omr/pull/3423#discussion_r246402674, I think we should be using `std::abs` from `<cmath>` instead of `<math.h>`.

knn-k:

PR #3430, opened 2019-01-09
compiler: build: Allow dollars in identifiers

rwy0717:
Apple patches libstdc++v3 with additional support for dtrace. These
dtrace symbols use dollar sign identifiers, and are incompatible with
the compile option -fno-dollars-in-identifiers. This commit removes the
option.


Leonardo2718:
@genie-omr build all

Leonardo2718:
From a private discussion with @0xdaryl, it appears the `-fno-dollars-in-identifiers` option was added several years ago, long before the project was open sourced, for reasons that are no longer relevant. We see no reason to continue using this option. Merging.

PR #3433, opened 2019-01-10
Optimize sequential recognized object shadow field loads in VP

yanluo7:
Optimize the generalized pattern `Field1 f1 = o.field1; if (f1==null) initialize_f1; Field2 f2 = f1.field2; if (f2==null) initialize_f2` where `o` is a known object. If value of `f1` is null, it is initialized just once to a valid non-null value and does not change; similarly, if `f2` is null, it is also initialized to a valid non-null value and also does not change during entire run time. If `o` is a known object. VP loads `f1` at compile time and checks for null. A non-null value means it has been initialized so we treat it as a known object and attach known object, non-null, and fixed class constraints. In the loading of `f2`, the newly propagated known object constraint is attached to the base object `f1`, so again we do a compile-time load of `f2` and attach the appropriate constraints. The "initialize-just-once-and-never-change" nature of the field loads are represented as special `recognized known object` fields, with the shadow sym refs specially tagged. 

A Java example is `MyEnumClass.class.getEnumConstants()`. The method retrieves the enum constants array via 2 indirect loads: EnumVars enumVars = classObject.enumVars followed by Object[] enumConstants = enumVars.cachedEnumConstants. Both fields are initialized once and never change.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -2151,6 +2151,61 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
          node->setIsNonNull(true);
          }
       else if (base
+               && base->getKnownObject()
+               && !symRef->isUnresolved() && symRef->getSymbol()->isRecognizedKnownObjectShadow()
+               && !vp->comp()->compileRelocatableCode())
+         // Base is known object, and field is recognized known object shadow,
+         // hence can be treated as known object.
+         {
+         TR::VMAccessCriticalSection constrainIaloadCriticalSection(vp->comp(),
+               TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+         if (constrainIaloadCriticalSection.hasVMAccess())
+            {
+            TR::KnownObjectTable *knot = vp->comp()->getOrCreateKnownObjectTable();
+            uintptrj_t baseObject = knot->getPointer(base->getKnownObject()->getIndex());
+            if (baseObject)
+               {
+               // Check if baseObject is actually the recognized field Class
+               TR_OpaqueClassBlock *baseObjectClazz = TR::Compiler->cls.objectClass(vp->comp(), baseObject);
+               int32_t recognizedClassNameLength;
+               const char *recognizedClassName = symRef->getSymbol()->owningClassNameCharsForRecognizedField(recognizedClassNameLength);
+               if (recognizedClassName)
+                  {
+                  TR_OpaqueClassBlock *recognizedClazz = vp->fe()->getClassFromSignature(recognizedClassName,
+                        recognizedClassNameLength, symRef->getOwningMethod(vp->comp()));
+                  if (recognizedClazz && (TR_yes == vp->fe()->isInstanceOf(baseObjectClazz, recognizedClazz, false)))
+                     {
+                     uintptrj_t fieldObject = vp->comp()->fej9()->getReferenceFieldAtAddress(baseObject + node->getSymbolReference()->getOffset());
+                     if (0 != fieldObject)
+                        {
+                        TR_OpaqueClassBlock *fieldObjectClazz = TR::Compiler->cls.objectClass(vp->comp(), fieldObject);
+                        if(!TR::Compiler->cls.isClassArray(vp->comp(), fieldObjectClazz))
+                           {
+                           traceMsg(vp->comp(), "Recognized known object field node %p \n", node);
+                           TR::KnownObjectTable::Index fieldObjectKnotIndex = knot->getIndex(fieldObject);
+                           // Global constraints should work here, as field loads get fresh value numbers
+                           constraint = TR::VPClass::create(vp,
+                                 TR::VPKnownObject::create(vp, fieldObjectKnotIndex),
+                                 TR::VPNonNullObject::create(vp), NULL, NULL,
+                                 TR::VPObjectLocation::create(vp, TR::VPObjectLocation::HeapObject));
+                           vp->addGlobalConstraint(node, constraint);
+                           }
+                        else
+                           {
+                           int32_t arrLength = TR::Compiler->om.getArrayLengthInElements(vp->comp(), fieldObject);
+                           traceMsg(vp->comp(), "Recognized known array field node %p length %d\n", node, arrLength);
+                           // Global constraints should work here, as field loads get fresh value numbers
+                           vp->addGlobalConstraint(node, TR::VPFixedClass::create(vp, fieldObjectClazz));
+                           vp->addGlobalConstraint(node, TR::VPNonNullObject::create(vp));
+                           vp->addGlobalConstraint(node, TR::VPArrayInfo::create(vp, arrLength, arrLength, TR::Compiler->om.getArrayElementWidthInBytes(vp->comp(), fieldObject)));
+                           }
+                        }
+                     }
+                  }
+               }
+            }
+         }
+      else if (base
                && base->isJavaLangClassObject() == TR_yes
                && base->isNonNullObject()
                && base->getKnownObject())
--- a/compiler/il/symbol/OMRSymbol.hpp
+++ b/compiler/il/symbol/OMRSymbol.hpp
@@ -354,6 +354,8 @@ public:
 
    inline bool isRecognizedShadow();
 
+   inline bool isRecognizedKnownObjectShadow();
+
    inline void setArrayletShadowSymbol();
    inline bool isArrayletShadowSymbol();
 
@@ -512,7 +514,7 @@ public:
       ArrayShadow               = 0x80000000,
       RecognizedShadow          = 0x40000000, // recognized field
       ArrayletShadow            = 0x20000000,
-      // Available              = 0x10000000,
+      RecognizedKnownObjectShadow = 0x10000000,
       GlobalFragmentShadow      = 0x08000000,
       MemoryTypeShadow          = 0x04000000,
       Ordered                   = 0x02000000,
--- a/compiler/il/symbol/OMRSymbol_inlines.hpp
+++ b/compiler/il/symbol/OMRSymbol_inlines.hpp
@@ -446,6 +446,12 @@ OMR::Symbol::isRecognizedShadow()
    return self()->isShadow() && _flags.testAny(RecognizedShadow);
    }
 
+bool
+OMR::Symbol::isRecognizedKnownObjectShadow()
+   {
+   return self()->isRecognizedShadow() && _flags.testAny(RecognizedKnownObjectShadow);
+   }
+
 void
 OMR::Symbol::setArrayletShadowSymbol()
    {


yanluo7:
Performance implication:
In a micro benchmark that repeatedly calling `SomeClass.class.getEnumConstants` and iterates on the returned constants array, this change improved throughput to 200%.

vijaysun-omr:
Two initial comments:

1. Can the descriptions be changed to be less Java-heavy ? This is not the place for discussing Java issues in depth. 

2. Did you put some thought into how this might be possible to structure in the code such that at least some of this Java-specific logic goes into the downstram project ? If not, please consider this a request to do so.

yanluo7:
@vijaysun-omr 
1. I've changed the description to be more generic and updated in the comment above. I will change the commit message as well when I have a new version of code to commit.
2. I think I can use recognized field to make it more reusable: if the base is known object and field is recognized to be init-once-and-never-change, we do the compile time load and attach the appropriate constraints.

yanluo7:
openj9 PR: https://github.com/eclipse/openj9/pull/4346

0xdaryl:
Thanks for making some changes to lessen the amount of OpenJ9-specific code in this PR.  As OMR PRs should be independent of any downstream project please refrain from referencing OpenJ9 PRs in OMR.

If the OMR PR is actually dependent on work needing to happen downstream first, as the PR author you can mark the OMR PR as a "WIP" so that it won't be merged until all other required work is completed (but that work shouldn't be tracked in the OMR PR).

fjeremic:
> If the OMR PR is actually dependent on work needing to happen downstream first, as the PR author you can mark the OMR PR as a "WIP" so that it won't be merged until all other required work is completed (but that work shouldn't be tracked in the OMR PR).

[1] https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#downstream-dependencies

yanluo7:
This PR has no dependency, can be merged after review -- no WIP needed.

vijaysun-omr:
Can the commit title typo "sequencial" be fixed please ?

vijaysun-omr:
Review from @andrewcraik and/or @jdmpapin would be good for this change.

andrewcraik:
Ok so this is interesting it is an alternative to how OpenJ9 has implemented a lot of 'known' field handling where the known fields are handled in the J9TransformationUtils.cpp with lists of whitelisted fields. I know this is an OMR review, but I mention this because the OpenJ9 implementation while not contributed to OMR originates from OMR hooks. Now there are actually 3 kinds of 'shadows' that the OpenJ9 code deals with: known object fields pointing to other known objects, known object fields pointing at representable native structs (eg natives that an address could be materialized) and final fields pointing at unrepresentable native structs (eg natives which can be dereferenced but which cannot be represented with a compile-time constant like a known object trusted field - the first case). This is essentially a mechanism for expressing 1. Now this does it by annotating a symref with a flag rather than having a list of structed fields which we examine when doing folding. Additionally, this has a benefit of creating a known object constraint from a known object constraint which seems appealing. If we are going to add this mechanism then the existing known object -> known object dereferences should switch to using it and the appropriate code updated to make sure we don't have two ways of representing the same thing. We also need the transformation utilities to be enhanced to handle creating the derived knownobject without having to use VP - eg if you have a symref with a known object and one of these shadow flags set we should create the known object immediately and not wait for VP. Given we want to be aggressive in putting the known object constraint on do we need this symref flag to do it - that is more dubious in my mind since it is easy to find structurally and serves to encode the same inforamtion in different ways...

yanluo7:
@andrewcraik 
The benefit of annotating the recognized shadow sym ref is the "recognition/annotation logic" is done early on during ilgen, we can use existing the recognized field mechanism to encode the information we want, this would make optimizer code cleaner I think without all the pattern recognition and class name comparison code scattered in different places. But I'm OK if we decide to move some of the logic to existing code in J9TransformationUtils.cpp.

The pattern described in this change is different from a dereferencing chain like a.b.c.d. As after each dereference there's control flow that initializes the field if null is read. We would still need to reply on VP to put on and propagate non-null constraint (and other constraint like array info constraint if the field is array) to remove/optimize extraneous control flow (which is measured to contributes to the perf gain as well).

vijaysun-omr:
@yanluo7 there are review comments outstanding on this PR. 
Just bringing it to your attention since there has'nt been much activity on it for a few days.

andrewcraik:
So I think this is a fine first step on moving the notion of what symrefs can be traversed into symref rather than being held outside the symref with matching logic scattered around the codebase (though currently centralizing in the various TransformUtil helper classes).

It is certainly possible that during ilgen we could attach both the known object shadow symref and mark the node non-null if we know, based on language semantics, that the node is non-null. That would allow folding and traversal earlier in the optimization sequence.

I think we probably do want to eventually add the corresponding representable native shadows and dereferenceable native shadow notions so that we can move away from pattern matching / symref matching entirely. That doesn't need to be done in this PR, but it is definitely something we should do.

Could you create an issue to track the implementation of this new infrastructure? Could we also start a markdown document with some description of what you have in this PR so it is easier for people to find and we can add other special shadows etc there to make it easier for people to understand how all of this is working?

andrewcraik:
@yanluo7 ping?

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,47 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
 
          node->setIsNonNull(true);
          }
+      else if (base
+               && base->isNonNullObject()
+               && base->getKnownObject())
```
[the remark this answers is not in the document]
yanluo7: Fixed the conditions.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,47 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+         if (!symRef->isUnresolved() && symRef->getSymbol()->isRecognizedKnownObjectShadow()
+               && !vp->comp()->compileRelocatableCode())
+            {
+            TR::VMAccessCriticalSection constrainIaloadCriticalSection(vp->comp(),
+                  TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+            if (constrainIaloadCriticalSection.hasVMAccess())
+               {
+               TR::KnownObjectTable *knot = vp->comp()->getOrCreateKnownObjectTable();
+               uintptrj_t baseObject = knot->getPointer(base->getKnownObject()->getIndex());
+               uintptrj_t fieldObject = vp->comp()->fej9()->getReferenceFieldAtAddress(baseObject + node->getSymbolReference()->getOffset());
```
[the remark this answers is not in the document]
yanluo7: Added additional check to ensure the baseObject class type is the same as the owning class type of the recognized field.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,47 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+                  TR_OpaqueClassBlock *fieldObjectClazz = TR::Compiler->cls.objectClass(vp->comp(), fieldObject);
+                  if(!TR::Compiler->cls.isClassArray(vp->comp(), fieldObjectClazz))
+                     {
+                     traceMsg(vp->comp(), "Recognized known object field node %p \n", node);
+                     TR::KnownObjectTable::Index fieldObjectKnotIndex = knot->getIndex(fieldObject);
+                     constraint = TR::VPClass::create(vp,
+                           TR::VPKnownObject::create(vp, fieldObjectKnotIndex),
+                           TR::VPNonNullObject::create(vp), NULL, NULL,
+                           TR::VPObjectLocation::create(vp, TR::VPObjectLocation::HeapObject));
+                     vp->addGlobalConstraint(node, constraint);
```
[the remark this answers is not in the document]
yanluo7: Added comment.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,47 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+                     }
+                  else
+                     {
+                     int32_t arrLength = TR::Compiler->om.getArrayLengthInElements(vp->comp(), fieldObject);
+                     traceMsg(vp->comp(), "Recognized known array field node %p length %d\n", node, arrLength);
+                     vp->addGlobalConstraint(node, TR::VPFixedClass::create(vp, fieldObjectClazz));
+                     vp->addGlobalConstraint(node, TR::VPNonNullObject::create(vp));
+                     vp->addGlobalConstraint(node, TR::VPArrayInfo::create(vp, arrLength, arrLength, TR::Compiler->om.sizeofReferenceField()));
```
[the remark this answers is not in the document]
yanluo7: Use API to get array element size.

yanluo7:
@andrewcraik Just revised the change based on comments received. Working on the documentation to track and expand the implementation of this new infrastructure.

andrewcraik:
I think that has addressed the concerns - @jdmpapin are you happy with this now? do you agree we will likely want shadows for other kinds of dereferenceable shadows eventually to migrate the whole indirect transformation infrastructure away from hardcoded symbols / symrefs?

yanluo7:
Created issue https://github.com/eclipse/openj9/issues/4703 to track the implementation of using symbol reference flags for foldable fields.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,57 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+         // hence can be treated as known object.
+         {
+         TR::VMAccessCriticalSection constrainIaloadCriticalSection(vp->comp(),
+               TR::VMAccessCriticalSection::tryToAcquireVMAccess);
+         if (constrainIaloadCriticalSection.hasVMAccess())
+            {
+            TR::KnownObjectTable *knot = vp->comp()->getOrCreateKnownObjectTable();
+            uintptrj_t baseObject = knot->getPointer(base->getKnownObject()->getIndex());
+            // Check if baseObject is actually the recognized field Class
+            TR_OpaqueClassBlock *baseObjectClazz = TR::Compiler->cls.objectClass(vp->comp(), baseObject);
```
jdmpapin: We should check before loading the class pointer that `baseObject != 0`. VP tries not to create known object constraints for null, but there is a null known object index.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -2150,6 +2150,57 @@ TR::Node *constrainIaload(OMR::ValuePropagation *vp, TR::Node *node)
...
+            int32_t baseObjClassNameLength;
+            char *baseObjClassName = TR::Compiler->cls.classNameChars(vp->comp(), baseObjectClazz, baseObjClassNameLength);
+            int32_t recognizedClassNameLength;
+            const char *recognizedClassName = symRef->getSymbol()->owningClassNameCharsForRecognizedField(recognizedClassNameLength);
+            if (recognizedClassName && (baseObjClassNameLength == recognizedClassNameLength)
+                  && !strncmp(baseObjClassName, recognizedClassName, recognizedClassNameLength))
```
jdmpapin: This name-based test will work as long as the type can be uniquely identified by name (in Java that means: has a name that is only allowed to be defined by the bootstrap loader), and as long as we don't want to read the field from an instance of a subclass. If possible, I think it would be better to get the `TR_OpaqueClassBlock*` for the class containing the recognized field, and see whether `isInstanceOf(baseObjectClazz, recognizedFieldDefiningClazz, true, true) == TR_yes`

jdmpapin:
> do you agree we will likely want shadows for other kinds of dereferenceable shadows eventually to migrate the whole indirect transformation infrastructure away from hardcoded symbols / symrefs?

I do think it's cleaner for properties like this one (known object knowledge can propagate to a non-null field) to be a property of the symref that gets queried explicitly, instead of something that consuming code knows about special symrefs and applies by comparing to special symrefs

compiler/optimizer/VPHandlers.cpp
yanluo7: Actually I thought you might suggest this. The recognized fields defining class we currently have are all basic bootloader classes so I thought this literal compare of the class name would work. Anyway, added isInstanceOf check.

compiler/optimizer/VPHandlers.cpp
yanluo7: Added null check

compiler/optimizer/VPHandlers.cpp
jdmpapin: I think it would make more sense to pass `objectTypeIsFixed=true`, but that doesn't affect whether we get `TR_yes`

vijaysun-omr:
@yanluo7 is this still a WIP ?

yanluo7:
@vijaysun-omr Still a WIP, but the PR it depends on is going to get merged soon. Once that happens I will take off the WIP and add a comment here.

yanluo7:
@vijaysun-omr The dependent PR has been merged. This PR is now ready to be merged.

yanluo7:
The AppVeyor build failure appears to be unrelated infra issue. This PR should be ready to merge. @vijaysun-omr

vijaysun-omr:
@genie-omr build all

Leonardo2718:
Appveyor build failed due to infrastructure issues so I restarted it.

vijaysun-omr:
Code looks okay to me and reviewers have approved. Checks have also passed. Merging.

PR #3434, opened 2019-01-10
Tril test case for imulh

knn-k:
This commit adds a test case for imulh.

### DIFF
--- a/fvtest/compilertriltest/ArithmeticTest.cpp
+++ b/fvtest/compilertriltest/ArithmeticTest.cpp
@@ -36,6 +36,11 @@ int32_t imul(int32_t l, int32_t r) {
     return l*r;
 }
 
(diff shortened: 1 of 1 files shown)


fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -34,6 +34,10 @@ int32_t imul(int32_t l, int32_t r) {
     return l*r;
 }
 
+int32_t imulh(int32_t l, int32_t r) {
+    return (int32_t)(((int64_t)l * (int64_t)r) >> 32);
```
Leonardo2718: To make the code more readable, can you split it into multiple lines using temporary variables? Also, please use `static_cast` instead of c-style casts.

Leonardo2718:
@genie-omr build all

PR #3436, opened 2019-01-10
SemiSpace active size

amicic:
Remove SemiSpace specific implementation of getActiveMemorySize() API,
that was returning only the size of Allocate part of SemiSpace.

Now getActiveMemorySize() will rely on base class MemorySubSpace and
return same what getActiveMemorySize(MEMORY_TYPE_NEW) or
getActiveMemorySize(MEMORY_TYPE_OLD | MEMORY_TYPE_NEW) used to return
(and will still be unchanged) and that is the total size of SemiSpace
(Allocate and Nursery).

This will affect a couple of spots in OMR and downstream objects that
specifically want Allocate Size.

Most of calls like heap->getActiveMemorySize() should not be affected,
because they would eventually call getActiveMemorySize(MEMORY_TYPE_OLD |
MEMORY_TYPE_NEW) against SemiSpace (which used to return total size even
before and will still do).


amicic:
Minor, but risky change. Please @LinHu2016 and @dmitripivkine, take a look as well, if this could affect beyond what I believe it could.

LinHu2016:
LGTM, memory related MxBeans try to collect similar information, but only use heap level APIs, so there is no side affection from MxBeans APIs.

charliegracie:
What is the reason for this?  The change looks good and I do not see any issues with it in OMR but I am just curious why the change is being made.

charliegracie:
@genie-omr build all

charliegracie:
Thank you for the explanation.  Looks good to me.

PR #3438, opened 2019-01-11
Check and fix to make sure rdbars are anchored

cathyzhyi:
This change adds check to make sure the rdbars are anchored
and anchor a rdbar if it's found unanchored.

Ideally all optimizations should anchor a rdbar when it's created.
This API is for fixing the unanchored rdbars in case some
optimization forget to anchor the node by mistake. Optimizations
should still try to generate correct rdbar trees instead of relying
on this API.

A proper ILValdiationRule will be added in a separate PR to only
validate the tree without fixing them.


compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,62 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+      {
+      TR::Node *node = tt->getNode();
+      if (node->getOpCodeValue() == TR::treetop ||
+          node->getOpCode().isResolveOrNullCheck() ||
+          node->getOpCode().isAnchor())
+         {
+         if (node->getFirstChild()->getOpCode().isReadBar())
+            anchoredRdbarNodes.add(node->getFirstChild());
+         }
+      verifyAndCollectUnanchoredRdbarsInSubtree(node, tt, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
```
Leonardo2718: Linter error:
```suggestion
      self()->verifyAndCollectUnanchoredRdbarsInSubtree(node, tt, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
```

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,62 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+      if (!anchoredRdbarNodes.contains(node))
+         {
+         TR_ASSERT(0, "node (n%dn) %p is rdbar but not anchored\n", node->getGlobalIndex(), node);
+         traceMsg(self(), "node (n%dn) %p is rdbar but not anchored\n", node->getGlobalIndex(), node);
+         unanchoredRdbarNodes.push_back(std::make_pair(node, currentTreeTop));
+         }
+      }
+   visitedNodes.add(node);
+   for (int i=0; i < node->getNumChildren(); i++)
+      verifyAndCollectUnanchoredRdbarsInSubtree(node->getChild(i), currentTreeTop, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
```
Leonardo2718: Linter error:
```suggestion
      self()->verifyAndCollectUnanchoredRdbarsInSubtree(node->getChild(i), currentTreeTop, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
```

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,62 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+         }
+      verifyAndCollectUnanchoredRdbarsInSubtree(node, tt, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
+      }
+
+   for (auto it = unanchoredRdbarNodes.begin(); it != unanchoredRdbarNodes.end(); it++)
+      {
+      TR::Node *node = (*it).first;
+      TR::TreeTop *tt = (*it).second;
+      tt->insertBefore(TR::TreeTop::create(self(), TR::Node::create(TR::treetop, 1, node)));
+      traceMsg(self(), "node (n%dn) %p is an unanchored readbar, anchor it now\n", node->getGlobalIndex(), node);
```
Leonardo2718: Might be useful to also print the id of the new treetop node.

compiler/compile/OMRCompilation.hpp
```diff
@@ -704,6 +704,12 @@ class OMR_EXTENSIBLE Compilation
    void verifyBlocks(TR::ResolvedMethodSymbol *s = 0);
    void verifyCFG(TR::ResolvedMethodSymbol *s = 0);
 
+   typedef TR::typed_allocator<std::pair<const TR::Node *, TR::TreeTop *>, TR::Region &> NodeTreeTopPairAllocator;
```
Leonardo2718: Type given to `typed_allocator` must match type given to `std::list`.

```suggestion
   typedef TR::typed_allocator<std::pair<TR::Node *, TR::TreeTop *>, TR::Region &> NodeTreeTopPairAllocator;
```

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,62 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+      }
+   }
+
+void OMR::Compilation::verifyAndCollectUnanchoredRdbarsInSubtree(TR::Node *node, TR::TreeTop *currentTreeTop, TR::NodeChecklist &anchoredRdbarNodes, NodeTreeTopPairList &unanchoredRdbarNodes, TR::NodeChecklist &visitedNodes)
+   {
+   if (visitedNodes.contains(node))
+      return;
+   if (node->getOpCode().isReadBar())
```
Leonardo2718: If a property was created for "nodes that must be anchored", I think this code could be easily generalized to cover other opcodes as well (not sure it's worth doing in this PR though).

compiler/compile/OMRCompilation.cpp
cathyzhyi: corrected. thanks!

compiler/compile/OMRCompilation.cpp
cathyzhyi: added `self()`. Thanks!

compiler/compile/OMRCompilation.cpp
cathyzhyi: add printing id of the new treetop node. thanks!

compiler/compile/OMRCompilation.hpp
cathyzhyi: corrected.

compiler/compile/OMRCompilation.cpp
cathyzhyi: that's a good idea. opened an issue to generalize the code with a property. https://github.com/eclipse/omr/issues/3441

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,63 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+         {
+         if (node->getFirstChild()->getOpCode().isReadBar())
+            anchoredRdbarNodes.add(node->getFirstChild());
+         }
+      self()->verifyAndCollectUnanchoredRdbarsInSubtree(node, tt, anchoredRdbarNodes, unanchoredRdbarNodes, visitedNodes);
+      }
+
+   for (auto it = unanchoredRdbarNodes.begin(); it != unanchoredRdbarNodes.end(); it++)
+      {
+      TR::Node *node = (*it).first;
```
Leonardo2718: Why not:
```suggestion
      TR::Node *node = it->first;
```

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,63 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+      TR::TreeTop *tt = (*it).second;
```
Leonardo2718: Why not:
```suggestion
      TR::TreeTop *tt = it->second;
```

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2081,63 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+ *    This API is for fixing the unanchored rdbars in case some optimizations forget to
+ *    anchor the node by mistake. Optimizations should still try to generate correct rdbar
+ *    trees instead of relying on this API.
+ */
+void OMR::Compilation::verifyAndFixRdbarAnchors()
+   {
+   TR::NodeChecklist anchoredRdbarNodes(self());
+   TR::NodeChecklist visitedNodes(self());
+   NodeTreeTopPairList unanchoredRdbarNodes(self()->trMemory()->currentStackRegion());
+   for (TR::TreeTop *tt = self()->getStartTree(); tt; tt = tt->getNextTreeTop())
```
Leonardo2718: I wonder if the code would be cleaner if it used IL iterators from `infra/ILWalk.hpp` instead.

cathyzhyi: Didn't know the API exists. Changed to use the node iterator to make the code more clear.

cathyzhyi:
@Leonardo2718 Hey Leo, do you think the change is ok now?

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2082,53 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+         if (!anchoredRdbarNodes.contains(node))
+            {
+            TR_ASSERT(0, "node (n%dn) %p is rdbar but not anchored\n", node->getGlobalIndex(), node);
+            traceMsg(self(), "node (n%dn) %p is rdbar but not anchored\n", node->getGlobalIndex(), node);
+            unanchoredRdbarNodes.push_back(std::make_pair(node, iter.currentTree()));
+            }
+         }
+      }
+
+   for (auto it = unanchoredRdbarNodes.begin(); it != unanchoredRdbarNodes.end(); it++)
```
andrewcraik: do we have to save them all up or could we just do it as we go? it would take less resources to do it as we go - eg put the anchors in before the current walk treetop so as to not mess up the iterator?

compiler/compile/OMRCompilation.cpp
```diff
@@ -87,6 +87,7 @@
 #include "infra/CfgEdge.hpp"                   // for CFGEdge
 #include "infra/Timer.hpp"                     // for TR_SingleTimer
 #include "infra/ThreadLocal.h"                 // for tlsDefine
+#include "infra/ILWalk.hpp"                    // for PreorderNodeIterator
```
fjeremic: Let's try alphabetize these as we add them.

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2085,44 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
+ * \brief
+ *    Check to make sure the rdbars are anchored and anchor a rdbar if it's found unanchored
+ *
+ * \note
+ *    Ideally all optimizations should anchor a rdbar when it's created.
+ *    This API is for fixing the unanchored rdbars in case some optimizations forget to
+ *    anchor the node by mistake. Optimizations should still try to generate correct rdbar
+ *    trees instead of relying on this API.
+ */
+void OMR::Compilation::verifyAndFixRdbarAnchors()
```
fjeremic: Please prefer to add API Doxygen information in the header file such that consumers looking at the API interface (the header) can understand what a function does. Users may not always have access to the source code of the implementation and they should not need it to understand API semantics.

compiler/compile/OMRCompilation.cpp
```diff
@@ -1084,6 +1085,8 @@ int32_t OMR::Compilation::compile()
          }
 #endif
 
+      self()->verifyAndFixRdbarAnchors();
```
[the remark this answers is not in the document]
cathyzhyi: Changed to run this only when rdbars can actually exist or under debug or assumes to avoid increasing compilation time.

compiler/compile/OMRCompilation.cpp
cathyzhyi: Changed to fix the treetop as we go instead of in another pass.

compiler/compile/OMRCompilation.cpp
cathyzhyi: changed to respect alphabetic order

compiler/compile/OMRCompilation.cpp
cathyzhyi: moved Doxygen comment into header file. Thanks!

compiler/compile/OMRCompilation.cpp
```diff
@@ -2079,6 +2085,33 @@ void OMR::Compilation::verifyCFG(TR::ResolvedMethodSymbol *methodSymbol)
...
 
+void OMR::Compilation::verifyAndFixRdbarAnchors()
+   {
+   TR::NodeChecklist anchoredRdbarNodes(self());
+   for (TR::PreorderNodeIterator iter(self()->getStartTree(), self()); iter.currentTree() != NULL; ++iter)
+      {
+      TR::Node *node = iter.currentNode();
+      if (node->getOpCodeValue() == TR::treetop ||
+          node->getOpCode().isResolveOrNullCheck() ||
+          node->getOpCode().isStoreReg() ||
```
vijaysun-omr: Is being under a store reg really to be considered "safe" ? 
It's morally like being under a store to an auto which I don't think you consider safe.

cathyzhyi: I thought reg store was not a real store but only an indicator for register allocation. The opcode doesn't return true for `isStore` either.

vijaysun-omr: It is not a real store but neither is to be considered an "anchor". i.e. usually a reg store is present where the original auto/parm store was and if it was illegal to have a readbar under a store for the first occurrence then it's equally illegal  for a readbar under a reg store to be the first occurrence.

vijaysun-omr:
@genie-omr build all

compiler/compile/OMRCompilation.cpp
cathyzhyi: Thanks for the explanation. removed the store reg check.

Issue #3440, opened 2019-01-11
Add awareness of read barriers to an arraycopy transformation


IBMJimmyk:
There is a transformation that changes a reference arraycopy node into a primitive arraycopy node if neither an arraystore check nor a write barrier are required. This transformation should also not occur if a read barrier is required.

transformArrayCopyCall inside ValuePropagationCommon.cpp needs to be changed to know about read barriers and elect not to transform the reference arraycopy node if read barriers are needed.

gita-omr:
@pshipton this is a prerequisite for https://github.com/eclipse/openj9/issues/3421

gita-omr:
PR: https://github.com/eclipse/omr/pull/3442

PR #3442, opened 2019-01-11
transformArrayCopyCall needs to know about read barriers

IBMJimmyk:
There is a transformation that changes a reference arraycopy node into a
primitive arraycopy node if neither an arraystore check nor a write
barrier are required. This transformation should also not occur if a read
barrier is required.

transformArrayCopyCall was changed to check if read barriers are required
and elect not to transform the reference arraycopy node if this is the
case.

Closes: #3440

### DIFF
--- a/compiler/optimizer/ValuePropagationCommon.cpp
+++ b/compiler/optimizer/ValuePropagationCommon.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


gita-omr:
@andrewcraik could you please review?

jdmpapin:
I thought we might also want to check whether read barriers are needed before setting `changeExistingNodeToPrimitiveCopy` [just below](https://github.com/eclipse/omr/blob/2f369b0d6af51cbf4ab8af6becf768ed1bacf089/compiler/optimizer/ValuePropagationCommon.cpp#L1870-L1873), but it looks like `elementSize == 0` is guaranteed there (since `!referenceArray1 && !referenceArray2 && !primitiveArray1 && !primitiveArray2`).

compiler/optimizer/ValuePropagationCommon.cpp
```diff
@@ -1103,6 +1104,11 @@ void OMR::ValuePropagation::transformArrayCopyCall(TR::Node *node)
    if (!comp()->getOptions()->needWriteBarriers())
       needWriteBarrier = false;
 
+   if (TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+      {
+      needReadBarrier = true;
+      }
```
Leonardo2718: Couldn't this code be simplified to

```c++
needReadBarrier = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads();
```

or even

```c++
bool needReadBarrier = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads();
```

IBMJimmyk: Yes. I've changed it to your second suggestion.

Leonardo2718:
@genie-omr build all

vijaysun-omr:
Looks safe to me. Merging.

PR #3445, opened 2019-01-11
Update target windows  version to win 7

dnakamura:

### DIFF
--- a/cmake/modules/platform/os/win.cmake
+++ b/cmake/modules/platform/os/win.cmake
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, 2018 IBM Corp. and others
+# Copyright (c) 2017, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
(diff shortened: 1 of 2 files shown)


Leonardo2718:
@genie-omr build win

youngar:
Change looks good.  I would consider putting an explanation of why we upgraded (for `GetCurrentProcessorNumberEx`)

charliegracie:
@samolisov If I remember correctly you added the code to require XP. Does this change break anything in your environment?

charliegracie:
@genie-omr build win

youngar:
> @samolisov If I remember correctly you added the code to require XP. Does this change break anything in your environment?

That change set it to WindowsXP because that is how it was in the old build system, and I think it was missing in cmake.

@dnakamura [It looks like openj9 is still using 0x501](https://github.com/eclipse/openj9/blob/ede34ef90806363e8ccae5abcb8f99bbdacc6761/runtime/makelib/targets.mk.win.inc.ftl#L186), is this going to be a problem?

PR #3446, opened 2019-01-11
Avoid revisit node in TR_ParameterToArgumentMapper

cathyzhyi:
The old way of iterating nodes can visit a node more than once. Change
to use PreorderNodeIterator to avoid revisiting any node.


vijaysun-omr:
Looks safe to me. Merging since checks have passed.

PR #3447, opened 2019-01-11
DDR: allow ddr-scanner to handle exe/dll files

dnakamura:
If filename does not end with '.pdb', assume the file is an exe/dll file


charliegracie:
@genie-omr build all

charliegracie:
The commit message explains the actual change but I am curious as to why we need the change?

charliegracie:
@dnakamura is the plan to include the debug data in the exe/dll instead of creating pdb files?

dnakamura:
The debug info always lives in .pdb files, however the .exe/.dll files has info which points to the location of the pdb file

charliegracie:
Ahhh that makes sense.  Thanks!

PR #3451, opened 2019-01-14
Add JITaaS option to do local compiles for remote compiles

harryyu1994:
This option enables JITaaS (a downstream project) clients to perform corresponding local compilations for their remotely compiled methods. The purpose of this is to generate and compare the trace logs of a method's remote compilation and local compilation.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


compiler/control/OMROptions.cpp
```diff
@@ -691,6 +691,7 @@ TR::OptionTable OMR::Options::_jitOptions[] = {
    {"enableIVTT",                         "O\tenable IV Type Transformation", TR::Options::enableOptimization, IVTypeTransformation, 0, "P"},
    {"enableJCLInline",                    "O\tenable JCL Integer and Long methods inlining", SET_OPTION_BIT(TR_EnableJCLInline), "F"},
    {"enableJITaaSHeuristics",             "O\tenable JITaaS heuristics", SET_OPTION_BIT(TR_EnableJITaaSHeuristics), "F"},
+   {"enableJITaaSDoubleCompile",          "O\tenable JITaaS to do both local and remote compilations", SET_OPTION_BIT(TR_EnableJITaaSDoubleCompile), "F"},
```
Leonardo2718: Options have to be order alphabetically, so I think `enableJITaaSDoubleCompile` needs to come before `enableJITaaSHeuristics`.

Leonardo2718:
I find using the word "Double" a bit confusing. What about "Hybrid" or "Dual" instead?

mstoodle:
Is the intention for the option to allow both local and remote compilations, or to request both local and remote compilations? Which one is used?

Since JitBuilder is on its way to providing JITaaS at OMR, I'd like to see this option be more clearly defined.

harryyu1994:
The intention is to request both local & remote compilations of the same methods. I guess we should go with "Dual"..?

mstoodle:
So the purpose of the local compilation is to provide a validating log for the remote compilation? The locally compiled code is not used, presumably?

How about `doLocalCompilesForRemoteCompiles" ? It's a mouthful, but a bit more clear (assuming I've guessed the right meaning?)

harryyu1994:
@mstoodle @Leonardo2718 may I get this reviewed again? thanks!

Leonardo2718:
@genie-omr build all

Leonardo2718:
Approved and checks passed. Merging.

PR #3452, opened 2019-01-14
Store barrier in co-operative copy setup

amicic:
This barrier is at the beginning of co-operative copy, before a thread
tries to win it, to ensure that other participating threads (that lost
the race) see the valid control info (like remaining bytes left).

Contrary to initial belief, the atomic operation (on source object
header) following the setup step (which mutates the destination object
header) is not an implicit store barrier for the setup. Hence, an
explicit one is introduced.

Closely related to https://github.com/eclipse/omr/pull/3262, which makes
sure that barrier is also executed at the end of co-operative copy.


amicic:
@dmitripivkine please, have a look

vijaysun-omr:
@genie-omr build all

PR #3454, opened 2019-01-14
Combine enter/exit tracing queries into one method

a7ehuo:
In practice we never enable methodEnterTracing without methodExitTracing.
isMethodEnterTracingEnabled() and isMethodExitTracingEnabled() are
implemented as the same. As proposed in issue eclipse/openj9#3585,
combine them into one method - isMethodTracingEnabled(), which should
be used to replace all usages of isMethodEnterTracingEnabled() and
isMethodExitTracingEnabled().

Signed-off-by: Annabelle Huo <Annabelle.Huo@ibm.com>

### DIFF
--- a/compiler/codegen/FrontEnd.hpp
+++ b/compiler/codegen/FrontEnd.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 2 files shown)


a7ehuo:
@mpirvu FYI

mpirvu:
Maybe it's worth saying that the end goal is to replace all usages of `isMethodEnterTracingEnabled()` and `isMethodExitTracingEnabled()` with `isMethodTracingEnabled()` because in practice we never enable methodEnterTracing without methodExit tracing.

compiler/codegen/FrontEnd.hpp
```diff
@@ -150,6 +150,7 @@ class TR_FrontEnd : public TR_Uncopyable
    virtual uint32_t offsetOfIsOverriddenBit();
 
    // Needs VMThread
+   virtual bool isMethodTracingEnabled(TR_OpaqueMethodBlock *method);
```
Leonardo2718: I think a comment stating that `isMethodTracingEnabled()` is intended to replace the other methods and why that's the case would be appropriate here.

a7ehuo:
@Leonardo2718 @mpirvu Please review the new update. Thank you.

Leonardo2718:
@genie-omr build all

PR #3462, opened 2019-01-15
Remove stale IVs in IVA

jdmpapin:
Induction variable analysis (IVA) should not leave stale induction variables (IVs) on the structure that may no longer be valid. IVA now removes all preexisting IVs from structure so that the only IVs remaining afterward are those discovered in the current pass.


jdmpapin:
For eclipse/openj9#3771

Leonardo2718:
@genie-omr build plinux

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks pretty safe to me. Merging now since checks have passed.

PR #3388, opened 2018-12-20
Tril tests for ificmpeq, etc.

knn-k:
This commit adds tests for integer conditional branches, such as ificmpeq.

Signed-off-by: knn-k <konno@jp.ibm.com>

PR #3463, opened 2019-01-15
Free the code cache repository segment in the Test Compiler and JitBuilder on JIT shutdown

Leonardo2718:
This change adds the ability to free the code cache memory segments to the compiler. In the Test Compiler and JitBuilder, the code cache repository segment is now freed when the JIT shuts down.

This should resolve the test failure seen in  #3388.

### DIFF
--- a/compiler/runtime/OMRCodeCacheManager.cpp
+++ b/compiler/runtime/OMRCodeCacheManager.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -237,6 +237,7 @@ OMR::CodeCacheManager::destroy()
       self()->freeMemory(codeCache);
       codeCache = nextCache;
       }
+   self()->freeCodeCacheSegment(_codeCacheRepositorySegment);
    _initialized = false;
    }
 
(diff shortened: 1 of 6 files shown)


Leonardo2718:
@genie-omr build all

0xdaryl:
I'll review, but I'd appreciate a once-over from at least one of @mpirvu or @dsouzai please.

Leonardo2718:
@genie-omr build all

compiler/runtime/OMRCodeCacheManager.cpp
```diff
@@ -237,6 +237,7 @@ OMR::CodeCacheManager::destroy()
       self()->freeMemory(codeCache);
       codeCache = nextCache;
       }
+   self()->freeCodeCacheSegment(_codeCacheRepositorySegment);
```
mpirvu: Does the codeCacheRepository always exist? In OpenJ9 sometimes (when the virtual address space is small) we do not allocate a repository. Instead, each code cache will have their own segment. Maybe adding a check is safer.

mpirvu:
oops. Took me too long to chase whether that statement `munmap(memSegment->_base, memSegment->_top - memSegment->_base + sizeof(TR::CodeCacheMemorySegment));` is correct or not

0xdaryl:
@mpirvu : I did the math from first principles, and it appears to be correct.  :-)

compiler/runtime/OMRCodeCacheManager.cpp
Leonardo2718: In the JitBuilder and the Test Compiler, the code cache repository always exists. However, the option to not use a repository is available at the OMR level. So, you are correct, it would probably be safer to add a check. I will open a new PR with a fix.

PR #3464, opened 2019-01-15
Revert back to not setting skip zero init on anewarray node

yanluo7:
On Power, setting skip zero init caused assertion in GC. Revert
back the logic for now.


andrewcraik:
LGTM - @vijaysun-omr could you review and merge? OpenJ9 acceptance build are failing because the flag is triggering GC assertions that are being investigated so best to be a bit less aggressive in this new opt for now.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Clearly a safe change since the flag to skip zero inits is not being set as a result of this revert. i.e. we get lesser optimization with this change. This was a very recent change committed into the codebase and so reverting should not disrupt any downstream projects.

vijaysun-omr:
Merging this change now since checks have passed.

yanluo7:
@andrewcraik @vijaysun-omr After talking to Dmitri, it's my understanding that for us to allocate an object array without zero init, the GC would require some "significant" work to support that -- so as it stands now, this change is the functionally correct one.
FYI, Dmitri told me the GC code that handles non-zero init object array used to be there, but was later disabled/deleted as "JIT were unhappy about perf bursts occur occasionally."

PR #3467, opened 2019-01-16
Implemented lmul evaluator

hsoontie:
Implemented lmul evaluator for AArch64

### DIFF
--- a/compiler/aarch64/codegen/BinaryEvaluator.cpp
+++ b/compiler/aarch64/codegen/BinaryEvaluator.cpp
@@ -115,6 +115,30 @@ static void mulConstant32(TR::Node *node, TR::Register *treg, TR::Register *sreg
       }
    }
 
+// Multiply a register by a 64-bit constant
+static void mulConstant64(TR::Node *node, TR::Register *treg, TR::Register *sreg, int64_t value, TR::CodeGenerator *cg)
+   {
+   if (value == 0)
+      {
+      loadConstant64(cg, node, 0, treg);
+      }
+   else if (value == 1)
+      {
+      generateMovInstruction(cg, node, treg, sreg);
+      }
+   else if (value == -1)
+      {
+      generateNegInstruction(cg, node, treg, sreg);
+      }
+   else
+      {
+      TR::Register *tmpReg = cg->allocateRegister();
+      loadConstant64(cg, node, value, tmpReg);
+      generateMulInstruction(cg, node, treg, sreg, tmpReg);
+      cg->stopUsingRegister(tmpReg);
+      }
+   }
+
 TR::Register *
 OMR::ARM64::TreeEvaluator::imulEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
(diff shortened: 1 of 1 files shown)


compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -159,10 +184,37 @@ OMR::ARM64::TreeEvaluator::imulhEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+      {
+      int64_t value = secondChild->getLongInt();
+      if (value > 0 && cg->convertMultiplyToShift(node))
+         {
+         trgReg = cg->evaluate(node);
+         return trgReg;
+         }
+      else
+         {
+         trgReg = cg->allocateRegister();
```
0xdaryl: You may not need to allocate a new register for the target if one of the source children dies at this instruction.  If the reference count on either source child is 1 (and it's evaluated into a register) then you should re-use it.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -159,10 +184,37 @@ OMR::ARM64::TreeEvaluator::imulhEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+         mulConstant64(node, trgReg, src1Reg, value, cg);
+         }
+      }
+   else
+      {
+      TR::Register *src2Reg = cg->evaluate(secondChild);
+      trgReg = cg->allocateRegister();
```
0xdaryl: Same comment about the target register above.

hsoontie:
@0xdaryl Here is an update! It's all together now as well.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -160,9 +184,48 @@ OMR::ARM64::TreeEvaluator::imulhEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+      {
+      trgReg = cg->allocateRegister();
+      }
+
+   if (secondChild->getOpCode().isLoadConst() && secondChild->getRegister() == NULL)
+      {
+      int64_t value = secondChild->getLongInt();
+      if (value > 0 && cg->convertMultiplyToShift(node))
+         {
+         trgReg = cg->evaluate(node);
```
0xdaryl: I think it is possible to reach here by taking the path through L203, which allocates a target register.  If that's the case, then that register will be lost.  Is it possible to move this logic where you decide to convert a multiply to a shift before the initial analysis starting at L193?  Alternatively, can you avoid allocating a target register if you know you'll be coming down this path?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -160,9 +184,48 @@ OMR::ARM64::TreeEvaluator::imulhEvaluator(TR::Node *node, TR::CodeGenerator *cg)
...
+   TR::Register *src2Reg = NULL;
+   TR::Register *trgReg = NULL;
+
+   if(1 == firstChild->getReferenceCount())
+      {
+      trgReg = src1Reg;
+      }
+   else if(1 == secondChild->getReferenceCount() && secondChild->getRegister() != NULL)
+      {
+      trgReg = src2Reg;
```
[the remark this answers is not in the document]
hsoontie: Thank you, I believe you're right, I think I'd meant to set it as I made the check.

hsoontie:
@0xdaryl 
Here is a small update. I moved the part of the logic you mentioned upwards.

hsoontie:
@0xdaryl The whitespace should be fixed up now! I will look into creating an issue for imul.

0xdaryl:
@genie-omr build aarch64

PR #3468, opened 2019-01-16
Only free the code cache repository segment when it's used

Leonardo2718:
Pull request #3463 added functionality to free the code cache repository
segment when the compiler's code cache manager is destroyed. However, it
is possible to create a code cache manager that doesn't use a
repository. This commit therefore adds to check to only free the
repository segment if it's actually being used.


Leonardo2718:
@mpirvu since you pointed out the issue in https://github.com/eclipse/omr/pull/3463#pullrequestreview-193250045, would you please review this change as well? :slightly_smiling_face:

Leonardo2718:
@genie-omr build all

PR #3469, opened 2019-01-16
Copy OSR exception edge from caller when block is not split

liqunl:
Inliner unconditionally merges the last block of callee and the caller's
block immediately after and copies the caller block's non-OSR exception
successors to callee blocks. Then it splits the merged block when
needed. Split is not needed when no guard and no OSR blocks are
generated for the callee. However, the OSR exception successor from the
caller, must be copied onto the merged block, otherwise the OSR points
from the caller block will be missing OSR exception edges.

This commit does the following:
1. Copy the missing OSR exception edge
2. Ask OSR method data whether OSR blocks have been generated for the
   callee instead of checking the trees and crossed basic blocks
3. Improve the code readability by changing variables names and adding
   comments
   - Rename `blockOfCaller` to `mergedBlock`
   - Rename `blockOfCallerInCalleeCFG` to `blockOfTreesAfterTheCall`
   - A static function `succIsOSRCatchBlock` is added to only copy OSR
   exception successor

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -165,6 +165,11 @@ static bool succAndPredAreNotOSRBlocks(TR::CFGEdge * edge)
       !toBlock(edge->getFrom())->isOSRCodeBlock() &&
       !toBlock(edge->getTo())->isOSRCatchBlock() &&
       !toBlock(edge->getFrom())->isOSRCatchBlock();
+   }
+
+static bool succIsOSRCatchBlock(TR::CFGEdge * edge)
+   {
+   return toBlock(edge->getTo())->isOSRCatchBlock();
    }
 
 int32_t
@@ -5003,6 +5008,7 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
    //
    TR::Block * firstCalleeBlock = calleeSymbol->getFirstTreeTop()->getNode()->getBlock();
    TR::Block * lastCalleeBlock = 0;
+   bool treesAfterCallMergedIntoLastCalleeBlock = false;
    if (tif->firstBBEnd())
       {
       // hook up the previous BBStart in the caller with the first BBEnd in the callee
@@ -5019,6 +5025,7 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
       lastCalleeBlock = tif->lastMainLineTreeTop()->getNode()->getBlock();
       nextBBEndInCaller->getNode()->setBlock(lastCalleeBlock);
       lastCalleeBlock->setExit(nextBBEndInCaller);
+      treesAfterCallMergedIntoLastCalleeBlock = true;
       }
 
    // merge cfgs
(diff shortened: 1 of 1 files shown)


compiler/optimizer/Inliner.cpp
```diff
@@ -5114,22 +5114,36 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
...
+
+         TR::Block * blockOfCallerInCalleeCFG = blockOfNextBBEndInCaller->split(callNodeTreeTop, callerCFG);
+         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfCallerInCalleeCFG);
+         }
+      else
+         {
+         debugTrace(tracer(), "Copy OSR exception edge from caller block_%d to block_%d\n", blockContainingTheCall->getNumber(), blockOfNextBBEndInCaller->getNumber());
+
+         // Need to copy the exception edge from caller block containing the call to the last callee block after merge
+         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfNextBBEndInCaller);
```
andrewcraik: The copy should not be bind - there should be an assert here that there isn't already an exception successor on the block where this is being copied to.

liqunl: Do you mean there isn't already an exception successor to OSR catch block? We copy exception edges except OSR exception edge from `blockContainingTheCall` to all the callee blocks that are not OSR catch block or OSR code block [here](https://github.com/eclipse/omr/blob/master/compiler/optimizer/Inliner.cpp#L5070). So there can be exception successor on the merged block at this point.

compiler/optimizer/Inliner.cpp
```diff
@@ -5114,22 +5114,36 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
...
+         debugTrace(tracer(), "Splitting block to have distint exception edges for callee %d\n", comp()->getCurrentInlinedSiteIndex());
+
+         TR::Block * blockOfCallerInCalleeCFG = blockOfNextBBEndInCaller->split(callNodeTreeTop, callerCFG);
+         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfCallerInCalleeCFG);
+         }
+      else
+         {
+         debugTrace(tracer(), "Copy OSR exception edge from caller block_%d to block_%d\n", blockContainingTheCall->getNumber(), blockOfNextBBEndInCaller->getNumber());
+
+         // Need to copy the exception edge from caller block containing the call to the last callee block after merge
```
cathyzhyi: could you add some comments to explain that `blockOfNextBBEndInCaller ` is merged with the last block in inlined callee so it loses all the original exception edges and therefore we need to copy those exception edges back to this block. To people not familiar with the code, `last callee block after merge` seems not clear enough to justify why we want to copy the exception edges from caller to callee.

liqunl: Made some changes to improve the comment and code readability.

compiler/optimizer/Inliner.cpp
andrewcraik: So we need to make sure that the exception edges are consistent and at the very least there needs to be a fatal assert, but probably some conditional code that will abort the merge if both blocks have OSR exception successors or conflicting exception edges.

liqunl: `copyExceptionSuccessors` will not copy existing successors. IMO, I don't think we need any assert.
The merge happens [here](https://github.com/eclipse/omr/blob/master/compiler/optimizer/Inliner.cpp#L5022). The trees from the last callee block and the caller trees immediately after the call are merged, they are all put into `lastCalleeBlock`. Then all non-OSR exception successors from `blockContainingTheCall` are copied to every callee block. So `lastCalleeBlock` has all exception successors from the callee and non-OSR exception successors from the caller. Then in the code I touched, we call `copyExceptionSuccessors` to copy the rest of exception successors, which is the OSR exception successor. At this point, I don't think there are conflicting exception edges.

I can pass in a filter function to `copyExceptionSuccessors`  to only copy OSR exception successor to make the code clear.

liqunl: @andrewcraik Added code to assert if there is a mismatch between the osr meta data and CFG.

andrewcraik:
@vijaysun-omr can you please review - we need this to fix some correctness issues

compiler/optimizer/Inliner.cpp
```diff
@@ -5143,14 +5143,22 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
...
 
-         debugTrace(tracer(), "Splitting block to have distint exception edges for callee %d\n", comp()->getCurrentInlinedSiteIndex());
 
-         TR::Block * blockOfTreesAfterTheCall = mergedBlock->split(callNodeTreeTop, callerCFG);
-         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfTreesAfterTheCall, succIsOSRCatchBlock);
+         debugTrace(tracer(), "Splitting block_%d to have distint exception edges for callee %d", mergedBlock->getNumber(), comp()->getCurrentInlinedSiteIndex());
+         // mergedBlock may have exception successors from the callee and non-OSR exception successors from the caller, after split,
+         // `blockOfTreesAfterTheCall` only contains trees from the caller, should not have exception successors from the callee.
+         TR::Block* blockOfTreesAfterTheCall = mergedBlock->split(callNodeTreeTop, callerCFG, false/*fixupCommoning*/, false/*copyExceptionSuccessors*/);
+         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfTreesAfterTheCall);
```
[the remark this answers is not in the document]
liqunl: It is just to improve readability. As we have copied non-OSR exception edges, here we only want to copy successor of OSR exception edge, even though `copyExceptionSuccessors` without passing the third argument would ignore existing successors.

compiler/optimizer/Inliner.cpp
```diff
@@ -5134,15 +5145,17 @@ bool TR_InlinerBase::inlineCallTarget2(TR_CallStack * callStack, TR_CallTarget *
 
          debugTrace(tracer(), "Splitting block to have distint exception edges for callee %d\n", comp()->getCurrentInlinedSiteIndex());
 
-         TR::Block * blockOfCallerInCalleeCFG = blockOfNextBBEndInCaller->split(callNodeTreeTop, callerCFG);
-         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfCallerInCalleeCFG);
+         TR::Block * blockOfTreesAfterTheCall = mergedBlock->split(callNodeTreeTop, callerCFG);
+         callerCFG->copyExceptionSuccessors(blockContainingTheCall, blockOfTreesAfterTheCall, succIsOSRCatchBlock);
```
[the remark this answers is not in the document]
liqunl: Yes, a lot of them is just for readability. Will update the commit message.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Checks have passed and the review comments have all been addressed. Merging.

PR #3470, opened 2019-01-16
WIP: CMake target library name override

jeanlego:
This PR will help build OpenJ9 and OMR with CMake,
OpenJ9 expect to be able to rename the target library but these were hardcoded in OMR.
This PR should fix this issue
 
### DIFF
--- a/gc/CMakeLists.txt
+++ b/gc/CMakeLists.txt
@@ -66,7 +66,7 @@ target_compile_definitions(omrgc_tracegen
 		$<TARGET_PROPERTY:omr_base,INTERFACE_COMPILE_DEFINITIONS>
 )
 
-add_library(omrgc STATIC
+add_library(${OMR_GC_LIB} STATIC
 	base/AddressOrderedListPopulator.cpp
 	base/AllocationContext.cpp
 	base/AllocationInterfaceGeneric.cpp
@@ -203,14 +203,14 @@ add_library(omrgc STATIC
 )
 
 if(OMR_GC_OBJECT_MAP)
-	target_sources(omrgc
+	target_sources(${OMR_GC_LIB}
 		PRIVATE
 			base/ObjectMap.cpp
 	)
 endif()
 
 if(OMR_GC_MODRON_STANDARD)
-	target_sources(omrgc
+	target_sources(${OMR_GC_LIB}
 		PRIVATE
 			base/standard/ConfigurationFlat.cpp
 			base/standard/ConfigurationStandard.cpp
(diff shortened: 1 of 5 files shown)


youngar:
Hi @jeanlego Does this have a corresponding OpenJ9 issue we can look at?

jeanlego:
@youngar not yet. It is in the works
I will raise the issue once I have the PR up in openJ9 too.
This is part of the effort to get an Aarch64 OMR.

Looking at the inProgress cards in https://github.com/eclipse/openj9/projects/9#card-15181106
More specifically see Issue https://github.com/eclipse/openj9/issues/3725, https://github.com/eclipse/openj9/issues/3725#issuecomment-439911310

to build OpenJ9 and omr with cmake, OpenJ9 seems to expect OMR to be inside the runtime directory of OpenJ9 and OpenJ9 renames the OMR_*_LIB i.e.: omrgc gets renamed to j9gc
https://github.com/eclipse/openj9/blob/master/runtime/CMakeLists.txt#L210-L232

When cloning OMR inside runtime, omr will complain that it cannot find the target j9gc since internally it is hardcoded to omrgc.

We seem to allow to change the name to some extend
as can be seen https://github.com/eclipse/omr/blob/39d6895ed93be3b1cf2c74859bdadec7c60e6450/cmake/config.cmake#L78-L95
throughout the code some places are using OMR_GC_LIB some are using omrgc

Using this PR CMake gets past this error.
This does not fully build yet but I wanted to keep the changes small as they affect the whole build.

jeanlego:
@0xdaryl know anyone on the OpenJ9 side of things? seems @dnakamura has been active on the CMake.

dnakamura:
These library names are meant to be hard coded, and Openj9 does not expect to be able to rename the libraries (AFAIK).  Openj9 builds its own version of port/thread/trace libraries (which are basically the omr versions with some extra stuff added). The variables were added so that omr components could link against the Openj9 versions of the libraries if we are building openj9

jeanlego:
Trying to get cmake to start a build for openj9 and omr on the aarch64 platform.

jeanlego:
native for now. cross compiling is in the scope. The Aarch64 port of OpenJ9 and OMR effort need CMake to test and compile the code. This was my first stab at it, maybe a bad one! If there's anything to do to help i'd be please to offer

dnakamura:
From an omr side  everything should be pretty close to being good to go, (many months ago i did have a prototype which was cross compiling on travis). The only thing to add would possibly be compiler flags etc. On the openj9 side its probably just adding source files. eg 
```
if(ARM)
target_sources(someLibrary arm/some_file.c)
endif()
```
Will take a quick look at it to see what is needed

aarongraham9:
Hey all (@jeanlego, @youngar, @dnakamura, @Leonardo2718, @0xdaryl)! Is this still valid/useful? Is so we should update it and fix the merge conflicts. If not then can we close it @jeanlego?

jeanlego:
can close

aarongraham9:
Hey all, @jeanlego says we can close this. Unless there's anything you'd like to salvage @youngar, @dnakamura, @Leonardo271 and, @0xdaryl we'll close this.

PR #3473, opened 2019-01-17
Support Symbol Validation Manager on Z

r30shah:
This pull request contains three main changes. 
1. Update relocation on Z codegen to support Symbol Validation Manager on Z if it is enabled. 
2. SideEffect guard is not supported in AOT yet though it was generated if Symbol Validation Manager, this caused a potential bug where we are left with Sife Effect guard node in AOT compiled body with no relocations. 
3. Fix x86 Codegen to check if it is compiling relocatable method first so that we do not miss assert in case we encounter a virtual guard that is not supported under AOT. 

### DIFF
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -2287,11 +2287,7 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
       }
 
    TR_VirtualGuardSite *site = NULL;
-   if (node->isSideEffectGuard())
-      {
-      site = comp->addSideEffectNOPSite();
-      }
-   else if (cg->needClassAndMethodPointerRelocations())
+   if (cg->needClassAndMethodPointerRelocations())
       {
       site = (TR_VirtualGuardSite *)comp->addAOTNOPSite();
       TR_AOTGuardSite *aotSite = (TR_AOTGuardSite *)site;
@@ -2304,10 +2300,9 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
          case TR_DirectMethodGuard:
          case TR_NonoverriddenGuard:
          case TR_InterfaceGuard:
-         case TR_AbstractGuard:
          case TR_MethodEnterExitGuard:
          case TR_HCRGuard:
-         //case TR_AbstractGuard:
+         case TR_AbstractGuard:
             aotSite->setGuard(virtualGuard);
             break;
 
@@ -2315,13 +2310,17 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
             break;
 
          default:
-            TR_ASSERT(0, "AOT guard in node but not one of known guards supported for AOT. Guard: %d", reloKind);
+            TR_ASSERT_FATAL(0, "AOT guard in node but not one of known guards supported for AOT. Guard: %d", reloKind);
             break;
          }
       }
-   else
+   else if (!node->isSideEffectGuard())
       {
       site = virtualGuard->addNOPSite();
+      }
+   else 
+      {
+      site = comp->addSideEffectNOPSite();
       }
 
    List<TR::Register> popRegisters(cg->trMemory());
--- a/compiler/z/codegen/ConstantDataSnippet.cpp
+++ b/compiler/z/codegen/ConstantDataSnippet.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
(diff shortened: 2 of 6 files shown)


r30shah:
@dsouzai  As you were expecting, with this change I do not see any performance difference in Daytrader 7 throughput performance enabling SVM.  
@fjeremic  @dsouzai  @jdmpapin  Can I please get a review?

compiler/z/codegen/ConstantDataSnippet.cpp
```diff
@@ -170,24 +197,38 @@ TR::S390ConstantDataSnippet::addMetaDataForCodeAddress(uint8_t *cursor)
...
-         uint8_t * targetAdress2 = NULL;
-         if (getNode()->getOpCodeValue() != TR::aconst)
+         if (cg()->comp()->getOption(TR_UseSymbolValidationManager))
             {
-            if (TR::Compiler->target.is64Bit())
-               targetAdress2 = (uint8_t *) *((uint64_t*) cursor);
-            else
-               targetAdress2 = (uint8_t *) *((uintptrj_t*) cursor);
+            TR_ASSERT_FATAL(getDataAs8Bytes(), "Static Sym can not be NULL");
+            cg()->addExternalRelocation(new (cg()->trHeapMemory()) TR::ExternalRelocation(cursor,
```
dsouzai: should this be `*((uint64_t*) cursor)`?

dsouzai: nvm, this is the relolocation...

compiler/z/codegen/ControlFlowEvaluator.cpp
```diff
@@ -169,6 +169,7 @@ virtualGuardHelper(TR::Node * node, TR::CodeGenerator * cg)
          case TR_DirectMethodGuard:
          case TR_NonoverriddenGuard:
          case TR_InterfaceGuard:
+         case TR_AbstractGuard:
          case TR_MethodEnterExitGuard:
          case TR_HCRGuard:
          //case TR_AbstractGuard:
```
jdmpapin: Can we either uncomment this one to use instead, or else delete it? The list on x86 is like this too

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -294,6 +294,11 @@ genLoadLongConstant(TR::CodeGenerator * cg, TR::Node * node, int64_t value, TR::
       TR::Instruction * temp = cursor;
       cursor = generateRegLitRefInstruction(cg, TR::InstOpCode::LG, node, targetRegister, value, TR_GlobalValue, cond, cursor, base);
       }
+   else if (comp->compileRelocatableCode() && sym && (sym->isRecompilationCounter()))
+      {
+      TR::Instruction * temp = cursor;
+      cursor = generateRegLitRefInstruction(cg, TR::InstOpCode::LG, node, targetRegister, value, TR_BodyInfoAddress, cond, cursor, base);
+      }
```
jdmpapin: How were recompilation counters handled before? Perhaps we didn't generate them?

jdmpapin:
The casts converting from `int64_t` (the result of `getDataAs8Bytes()`) to `uint8_t*` will probably cause warnings on 32-bit. I know that for the moment they won't *run* on 32-bit, but I thought I'd mention it

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -2287,11 +2287,7 @@ static bool virtualGuardHelper(TR::Node *node, TR::CodeGenerator *cg)
       }
 
    TR_VirtualGuardSite *site = NULL;
-   if (node->isSideEffectGuard())
-      {
-      site = comp->addSideEffectNOPSite();
-      }
-   else if (cg->needClassAndMethodPointerRelocations())
+   if (cg->needClassAndMethodPointerRelocations())
```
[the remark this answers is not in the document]
r30shah: Changed to FATAL assert in https://github.com/eclipse/omr/pull/3473/commits/2618d981a67e09dc3034e1d02931795e8b2232d3

compiler/z/codegen/ControlFlowEvaluator.cpp
r30shah: Changed in https://github.com/eclipse/omr/pull/3473/commits/8617dd3b42d21960eb7135295da2fd32e956fe2c

compiler/z/codegen/OMRTreeEvaluator.cpp
r30shah: We used to have different trees previously for recompilation counters which roughly looks like following 
```
iloadi
    loadaddr <recompilation_counter>
```
Relocatable code for `loadaddr` on Z is handled here : https://github.com/eclipse/omr/blob/22db7f45eb2d2593f711e87c59c7688064335443/compiler/z/codegen/OMRTreeEvaluator.cpp#L11528

Because of this change https://github.com/eclipse/openj9/pull/4311/files#diff-eeab0f6a82448f563e51b3e7529826f7 where we now have `iload <recompilation_counter>` and come to this function : https://github.com/eclipse/omr/blob/22db7f45eb2d2593f711e87c59c7688064335443/compiler/z/codegen/OMRTreeEvaluator.cpp#L279 which does not check if sym ref is of recompilation counter.

r30shah:
@jdmpapin  I agree. I have this cast where I am adding relocation (Looking at the function signature : https://github.com/eclipse/omr/blob/22db7f45eb2d2593f711e87c59c7688064335443/compiler/codegen/Relocation.hpp#L393)

fjeremic:
@genie-omr build all

PR #3474, opened 2019-01-18
Added changes to resolve the balanced gc policy crash in container

bharathappali:
As mentioned in the OpenJ9 issue [#3797](https://github.com/eclipse/openj9/issues/3767) [PortLibrary](https://github.com/eclipse/omr/blob/b21d26c497f0a49736bb39897cc8e012e45df2cf/port/linux/omrvmem.c#L174) and [ThreadLibrary](https://github.com/eclipse/omr/blob/f8fc704032f95cbc0f337eb6be54f2b06e58a773/thread/linux/omrthreadnuma.c#L149) are not having a consistent view on NUMA which is making JVM to crash if its running in container (default docker or any environment where `getmempolicy` system call is restricted) with balanced gc policy. 

This Fix would allow the PortLibrary to get node count and set the memorystate to `J9NUMA_DENIED` if there is a failure in `getmempolicy` system call due to only `security restriction like EPERM - Operation Not Permitted`

FYI @LinHu2016 @ashu-mehra 

### DIFF
--- a/port/linux/omrvmem.c
+++ b/port/linux/omrvmem.c
@@ -180,6 +180,7 @@ initializeNumaGlobals(struct OMRPortLibrary *portLibrary)
 	int schedReturnCode = 0;
 	int mempolicyReturnCode = 0;
 	int useAllNodes = 0;
+	PPG_numaSyscallNotAllowed = FALSE;
 
 	memset(&PPG_numa_available_node_mask, 0, sizeof(J9PortNodeMask));
 	PPG_numa_max_node_bits = 0;
@@ -192,7 +193,9 @@ initializeNumaGlobals(struct OMRPortLibrary *portLibrary)
 	PPG_numa_policy_mode = -1;
 	memset(&PPG_numa_mempolicy_node_mask, 0, sizeof(J9PortNodeMask));
 	mempolicyReturnCode = do_get_mempolicy(&PPG_numa_policy_mode, PPG_numa_mempolicy_node_mask.mask, maxNodes, (unsigned long)NULL, 0);
-
+	if ((0 != mempolicyReturnCode) && (EPERM == errno)) {
+		PPG_numaSyscallNotAllowed = TRUE;
+	}
 	if (0 == mempolicyReturnCode) {
 
 		long anySet = 0;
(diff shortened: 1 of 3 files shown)


port/linux/omrvmem.c
```diff
@@ -219,8 +222,10 @@ initializeNumaGlobals(struct OMRPortLibrary *portLibrary)
 		Trc_PRT_vmem_omrvmem_initializeNumaGlobals_get_mempolicy_failure(errno);
 	}
 
-	/* only proceed if we could successfully look up the memory policy and default affinity */
-	if ((0 == schedReturnCode) && (0 == mempolicyReturnCode)) {
+	/* Proceed if we could successfully look up the default affinity.
+	 * We proceed  to get node count even if `getmempolicy` fails due to security restrictions.
```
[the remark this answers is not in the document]
bharathappali: @charliegracie Thanks for the review. I have removed the extra space.

port/linux/omrvmem.c
```diff
@@ -192,7 +193,9 @@ initializeNumaGlobals(struct OMRPortLibrary *portLibrary)
 	PPG_numa_policy_mode = -1;
 	memset(&PPG_numa_mempolicy_node_mask, 0, sizeof(J9PortNodeMask));
 	mempolicyReturnCode = do_get_mempolicy(&PPG_numa_policy_mode, PPG_numa_mempolicy_node_mask.mask, maxNodes, (unsigned long)NULL, 0);
-
+	if ((0 != mempolicyReturnCode) && (EPERM == errno)) {
+		PPG_syscallNotAllowed = TRUE;
```
charliegracie: Shouldn't this global have `numa` somewhere in the name or does this mean that no `syscall` can be made at all?

bharathappali: @charliegracie I would like to rename it to something which states [these calls](https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile) are not allowed. NUMA is one of the consumer of these calls and someother functionality might also get benefit from this global if it needs to perform any of these calls.

I would like to take your view of how we could name it ? If we can just keep it to the NUMA context as currently its the only consumer of it we can add NUMA in it. Thank you

bharathappali: @charliegracie I'm sorry i'm looking in the docker container perspective only, As OMR is not bound to any such things naming it having `numa` in it would be good (As we are using it in NUMA perspective only). Will `PPG_numaSyscallsNotAllowed` be fine ?

bharathappali:
@LinHu2016 Are you referring to call `omrthread_numa_set_enabled(FALSE)` in `omrvmem_startup`
Something like this : 
```
#if defined(OMR_PORT_NUMA_SUPPORT)
        if (0 == initializeNumaGlobals(portLibrary)) {
                PPG_numa_platform_supports_numa = 1;
        } else {
                PPG_numa_platform_supports_numa = 0;
                omrthread_numa_set_enabled(FALSE);/* Disabling thread NUMA */
        }
#endif

```

LinHu2016:
LGTM, BTW there is duplicate code in ztpf/omrvmem.c, please make sure if you need to update it too.

bharathappali:
Updated `ztpf/omrvmem.c` to resolve the crash.

charliegracie:
@genie-omr build all

PR #182, opened 2016-08-16
FIX #127: Remove MM_ScanClassesMode and dependencies from OMR

ktbriggs:
Implement VM access functions for example/glue. Complete ConcurrentGC
integration with example glue and GCTest.

Signed-off-by: Kim Briggs <briggs@ca.ibm.com>

Travis build (3e5edb4): https://travis-ci.org/ktbriggs/omr/jobs/152670523

PR #3475, opened 2019-01-18
Implement fremEvaluator

hsoontie:
WIP Implementing fremEvaluator for AArch64
Other changes:
- Added new floating point <-> integer conversion opcodes


hsoontie:
Notes about this PR: This doesn't yet work, but passes some tests. The rounding follows the IEEE "round to nearest, tied to even" method.

hsoontie:
This is now passing all tests and should be good to go (note that it doesn't check for division by zero right now). If there are any problems let me know!

knn-k:
When you move or add opcodes in OMRInstOpCodeEnum.hpp and OpBinary.cpp, please do the same with the `opCodeToNameMap[]` table in ARM64Debug.cpp.  Thanks.

compiler/aarch64/codegen/ARM64Debug.cpp
```diff
@@ -435,32 +435,79 @@ static const char *opCodeToNameMap[] =
...
    "fcmpd",
    "fcmpd_zero",
    "fmovs",
-   "fmovd",
-   "fabss",
-   "fabsd",
-   "fnegs",
-   "fnegd",
+	"fabss",
+	"fnegs",
```
knn-k: Indentation of lines from "fabss" to "fcvt_htod" is different from that of other lines.

hsoontie:
@0xdaryl Could you also take a look at this when you have a moment? Thanks :)

hsoontie:
@0xdaryl This PR's merge conflicts have been fixed and it should be ready for further review.

aarongraham9:
This one should be good for a review @0xdaryl.

hsoontie:
Hello all,
This PR has been on hold for a while. To get things out there, here is what I am dealing with at the moment:
I believe fremEvaluator() worked before I performed a rebase to fix some merge conflicts a while ago. Testing things now results in some strange issues that I cannot pinpoint the source of:

- Running it against the frem tril test code in #3500, it will pass over 400 test cases but fail 32, all `UsingConst`. Examples of some failures:
```
 Expected: exp
      Which is: 0
To be equal to: act
      Which is: -1.49012e-08
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/59, where GetParam() = ((1, 0.1), ("frem", 0xe7eb04)) (23 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:288: Failure
      Expected: exp
      Which is: -0
To be equal to: act
      Which is: 1.49012e-08
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/89, where GetParam() = ((-1, 0.1), ("frem", 0xe7eb04)) (9 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:288: Failure
      Expected: exp
      Which is: 0
To be equal to: act
      Which is: -6.25849e-07
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/119, where GetParam() = ((42, 0.1), ("frem", 0xe7eb04)) (8 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:288: Failure
      Expected: exp
      Which is: -0
To be equal to: act
      Which is: 6.25849e-07
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/149, where GetParam() = ((-42, 0.1), ("frem", 0xe7eb04)) (9 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:288: Failure
      Expected: exp
      Which is: 0
To be equal to: act
      Which is: -40
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/167, where GetParam() = ((3.40282e+38, 95), ("frem", 0xe7eb04)) (8 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:288: Failure
      Expected: exp
      Which is: 0
To be equal to: act
      Which is: -1.8228e+08
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/171, where GetParam() = ((3.40282e+38, 2.1642e+09), ("frem", 0xe7eb04)) (9 ms)

```
- Commenting out all the GenerateInstruction calls in the evaluator except the first one and changing the frem() call in the ArithmeticTest.cpp tril tests to return a simple division shows that the evaluator cannot properly generate a divide as its first step. This produced 256 passes and 194 failures when I tested it. Meanwhile, fdivEvaluator works fine with nearly identical code. Examples of some of the failures:
```
 Expected: exp
      Which is: 1
To be equal to: act
      Which is: 0
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/33, where GetParam() = ((1, 1), ("frem", 0xe7ea74)) (23 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:289: Failure
      Expected: exp
      Which is: -1
To be equal to: act
      Which is: 0
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/35, where GetParam() = ((1, -1), ("frem", 0xe7ea74)) (9 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:289: Failure
      Expected: exp
      Which is: 0.0238095
To be equal to: act
      Which is: 1
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/37, where GetParam() = ((1, 42), ("frem", 0xe7ea74)) (8 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:289: Failure
      Expected: exp
      Which is: -0.0238095
To be equal to: act
      Which is: 1
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/39, where GetParam() = ((1, -42), ("frem", 0xe7ea74)) (8 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:289: Failure
      Expected: exp
      Which is: 2.93874e-39
To be equal to: act
      Which is: 1
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/41, where GetParam() = ((1, 3.40282e+38), ("frem", 0xe7ea74)) (9 ms)
../../fvtest/compilertriltest/ArithmeticTest.cpp:289: Failure
      Expected: exp
      Which is: 1
To be equal to: act
      Which is: 0
[  FAILED  ] ArithmeticTest/FloatArithmetic.UsingConst/43, where GetParam() = ((1, 1), ("frem", 0xe7ea74)) (8 ms)
```
I will be working on fixing this on my own, but if anyone has suggestions for me I would be glad to hear them.

0xdaryl:
You asked a few questions in the public Slack channel.  I'll answer them here:

> When running tril tests, a function (getLinkage() in \compiler\runtime\Runtime.hpp) is indexing into the _linkage array using the runtime helper TR_AMD64floatRemainder (index #182). _linkage[182] returns 0, which corresponds to the linkage type TR_None. This ultimately leads to a segmentation fault later after looking for the linkage type.
>
>Some questions I have:
> - Could anybody explain what "linkage convention" means in general, or what the different linkage conventions mean specifically? (Ex: TR_None, TR_Private, TR_System, TR_Helper, TR_J9JNILinkage)

"Linkage" refers to the ABI to use when one function calls another.  It is more commonly referred to as a "calling convention" in other compiler projects, but the term "Linkage" is engrained in the history of the OMR compiler.  There are a few linkages that are known to the OMR compiler, but not all are implemented in the OMR compiler:
* `TR_None` : really means "unknown", or "not yet known"
* `TR_Private` : this is a language-specific linkage, generally the linkage in use for one method in a language runtime to call another.
* `TR_System` : this is the System ABI.  It varies by operating system and architecture.  However, it is generally the linkage in place when one C function calls another.  This should be the default linkage in my opinion.
* `TR_Helper` : this is generally the ABI to use when calling "helper functions".  Helper functions are usually those that you find in the runtime helper table.  However, as you'll see below, the linkage of a helper in that table may not always be `TR_Helper`.  The `TR_Helper` linkage can be used for those helpers that have a different calling requirement.
* `TR_J9JNILinkage` : a Java-specific linkage for JNI dispatches.  This really shouldn't be in OMR.

> - Does anybody know what the _linkage array is used for?

For each helper that is known, we keep track of its entry point (in the `_helpers` table) and its linkage (in the `_linkage` table).

> - Should using TR_AMD64floatRemainder as an index into _linkage result in TR_None in the first place?

Not likely, if you're expecting something valid to be there.   Because most helpers are project-specific, each project is responsible for initializing the helpers table.  For example, in JitBuilder, see:
https://github.com/eclipse/omr/blob/99100d4c32d296f349a791068329f8754ece4a57/jitbuilder/control/Jit.cpp#L57
There isn't a runtime version of float remainder on x86 in OMR (JitBuilder or CompilerTest), which is the function you're looking for and probably why you're getting `TR_None` back for the linkage.  That's one of the gaps we have with floating point unfortunately that we need to fix.  OpenJ9 provides a function, but it's written in assembler.  Such a function can't be directly used in OMR because there isn't a common assembler on x86 available as part of the toolchain (and I'm not convinced there should be).

hsoontie:
We have investigated the issue further today.
When testing in the way described above and printing text to a file every time fremEvaluator() is called, the text is printed 240 times, which seems to indicate that it does not get to the evaluator every time it is supposed to (450 test cases are being run). 256 tests passed, and @aarongraham9 guessed that 16 of them stopped at the optimizer level, meaning that the print might only happen in the result of a passed test. I will continue to look into this on Monday.

hsoontie:
Another short update:
With the above setup, all the failed tests are `UsingConst`. The few `UsingConst` cases that pass use a 0 denominator.

hsoontie:
@0xdaryl , to let you know where I am at, here is a shorter summary of the issue I am encountering here:

Context:
- An undetermined issue is causing tests of fremEvaluator() to produce strange results.

Setup:
- fremEvaluator()'s code was changed to call and return the result of fdivEvaluator(), which works normally when testing fdiv.
- The frem() function in ArithmeticTest.cpp was changed to call and return the result of fdiv(), which works normally when testing fdiv.

Results:
- All UsingLoadParam test cases pass.
- A large majority of UsingConst test cases fail.
- The UsingConst cases that pass seem to involve 0 or 1 denominators (perhaps this has to do with optimization?).
- 450 test cases run.
- 256 cases passed.
- 194 cases failed.
- fremEvaluator() called 240 times.

aarongraham9:
Any thoughts on this one @Leonardo2718 or @0xdaryl?

aarongraham9:
Any further thoughts on the run issues in this PR @Leonardo2718 and @0xdaryl?

aarongraham9:
Thoughts @Leonardo2718 and @0xdaryl.

hsoontie:
Rebasing & testing this showed no improvements on my end (it still fails the same number of tests). for this new version I just made a minor correction to the evaluator itself (removed a line that was unnecessary) and pushed the code that enables frem because I realized I had forgotten it.

aarongraham9:
Any word on the Issue(s) to link this agains @0xdaryl?

aarongraham9:
Any word on the Issue(s) to link this against @0xdaryl?

PR #3477, opened 2019-01-18
Move Metronome Reference Stats from OpenJ9 to OMR

jason-hall:
Move the reference stats from MM_MarkJavaStats to MM_MetronomeStats as these fields are only used by metronome. After adding them to MetronomeStats, the usage is to be changed in OpenJ9.

Signed-off-by: Jason Hall <jasonhal@ca.ibm.com>

### DIFF
--- a/gc/stats/MetronomeStats.cpp
+++ b/gc/stats/MetronomeStats.cpp
@@ -25,4 +25,7 @@
 void
 MM_MetronomeStats::merge(MM_MetronomeStats* statsToMerge)
 {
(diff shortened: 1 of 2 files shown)


jason-hall:
Would you prefer to move the pre-existing Javs-specific out of gc/stats/MetronomeStats? classLoaderUnloaderCount, classesUnloadedCount, anonymousClassesUnloadedCount, etc are in there.

charliegracie:
Any cleanup around the stats structures would be appreciated

PR #3479, opened 2019-01-20
Provide a consistent interface for asking if trampolines are required

0xdaryl:
* Add CPU-specific tests for relative branch reachability
* Remove unused AlwaysUseTrampoline CodeGenerator option
* Add directCallRequiresTrampoline CodeGenerator query
* Use directCallRequiresTrampoline CodeGenerator query

### DIFF
--- a/compiler/aarch64/codegen/OMRCodeGenerator.hpp
+++ b/compiler/aarch64/codegen/OMRCodeGenerator.hpp
@@ -286,6 +286,17 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
    TR_GlobalRegisterNumber _gprLinkageGlobalRegisterNumbers[TR::RealRegister::NumRegisters]; // could be smaller
    TR_GlobalRegisterNumber _fprLinkageGlobalRegisterNumbers[TR::RealRegister::NumRegisters]; // could be smaller
 
+   /**
+    * @brief Answers whether a trampoline is required for a direct call instruction to
+    *           reach a target address.
+    *
+    * @param[in] targetAddress : the absolute address of the call target
+    * @param[in] sourceAddress : the absolute address of the call instruction
+    *
+    * @return : true if a trampoline is required; false otherwise.
+    */
+   bool directCallRequiresTrampoline(intptrj_t targetAddress, intptrj_t sourceAddress);
+
    private:
 
    enum // flags
--- a/compiler/aarch64/env/OMRCPU.cpp
+++ b/compiler/aarch64/env/OMRCPU.cpp
@@ -20,3 +20,12 @@
  *******************************************************************************/
 
 #include "env/CPU.hpp"
+#include "env/jittypes.h"
+
+bool
+OMR::ARM64::CPU::isTargetWithinUnconditionalBranchImmediateRange(intptrj_t targetAddress, intptrj_t sourceAddress)
+   {
+   intptrj_t range = targetAddress - sourceAddress;
+   return range <= self()->maxUnconditionalBranchImmediateForwardOffset() &&
+          range >= self()->maxUnconditionalBranchImmediateBackwardOffset();
+   }
--- a/compiler/arm/codegen/ARMBinaryEncoding.cpp
+++ b/compiler/arm/codegen/ARMBinaryEncoding.cpp
@@ -69,17 +69,14 @@ uint32_t encodeHelperBranchAndLink(TR::SymbolReference *symRef, uint8_t *cursor,
 
 uint32_t encodeHelperBranch(bool isBranchAndLink, TR::SymbolReference *symRef, uint8_t *cursor, TR_ARMConditionCode cc, TR::Node *node, TR::CodeGenerator *cg)
    {
-   int32_t  distance;
-   uint32_t target;
+   intptrj_t target = (intptrj_t)symRef->getMethodAddress();
 
-   target = (uintptr_t)symRef->getMethodAddress();
-   distance = target - (uintptr_t) cursor;
-   if (distance > BRANCH_FORWARD_LIMIT || distance < BRANCH_BACKWARD_LIMIT)
+   if (cg->directCallRequiresTrampoline(target, (intptrj_t)cursor))
       {
       target = cg->fe()->indexedTrampolineLookup(symRef->getReferenceNumber(), (void *)cursor);
-      TR_ASSERT(((int32_t)target - (int32_t)cursor) <= BRANCH_FORWARD_LIMIT &&
-             ((int32_t)target - (int32_t)cursor) >=  BRANCH_BACKWARD_LIMIT,
-             "CodeCache is more than 32MB.\n");
+
+      TR_ASSERT_FATAL(TR::Compiler->target.cpu.isTargetWithinBranchImmediateRange(target, (intptrj_t)cursor),
+                      "Target address is out of range");
       }
 
    cg->addExternalRelocation(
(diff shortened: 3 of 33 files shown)


0xdaryl:
This is a WIP while further testing completes.

In the meantime, I'd appreciate a review from @fjeremic @ymanton @0dvictor @knn-k for the platform-specific changes please.

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build all

0xdaryl:
@genie-omr build aarch64

compiler/aarch64/env/OMRCPU.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "env/CPU.hpp"
+#include "env/jittypes.h"
+
+bool
+OMR::ARM64::CPU::isTargetWithinUnconditionalBranchImmediateRange(intptrj_t targetAddress, intptrj_t sourceAddress)
+   {
+   intptrj_t range = targetAddress - sourceAddress;
+   return (range <= self()->maxUnconditionalBranchImmediateForwardOffset() &&
+           range >= self()->maxUnconditionalBranchImmediateBackwardOffset()) ? true : false;
```
fjeremic: Do we need the ternary operator here? Omitting `? true : false` accomplishes the exact same thing.

compiler/codegen/OMRCodeGenerator.hpp
```diff
@@ -1340,16 +1340,32 @@ class OMR_EXTENSIBLE CodeGenerator
...
+    * @param[in] sourceAddress : the absolute address from which the call displacement
+    *              range is measured.  The meaning of this address is architecture
+                   specific.
+    *
+    * @return : true.  The default value is true because it should only result in
+    *              correct (but possibly suboptimal) behaviour.
+    */
+   bool directCallRequiresTrampoline(intptrj_t targetAddress, intptrj_t sourceAddress)
+      {
+      TR_ASSERT(0, "An architecture specialization of this function must be provided.");
```
fjeremic: Sounds like a fatal assert. Can we make it as such?

compiler/z/codegen/OMRSnippet.cpp
```diff
@@ -149,7 +149,8 @@ OMR::Z::Snippet::generatePICBinary(TR::CodeGenerator * cg, uint8_t * cursor, TR:
          }
 #endif
 
-      TR_ASSERT(CHECK_32BIT_TRAMPOLINE_RANGE(destAddr, cursor),  "Helper Call is not reachable.");
+      TR_ASSERT(TR::Compiler->target.cpu.isTargetWithinBranchRelativeRILRange(destAddr, (intptrj_t)cursor),
+                "Helper Call is not reachable.");
```
fjeremic: Let's make this a fatal assert. I'd rather have a compile time assert failure than unpredictable runtime crash.

fjeremic: Similarly elsewhere in Z codegen where asserts were touched.

compiler/aarch64/env/OMRCPU.cpp
fjeremic: Similarly elsewhere in this PR.

compiler/z/codegen/OMRSnippet.cpp
0xdaryl: Agreed.

compiler/codegen/OMRCodeGenerator.hpp
0xdaryl: Yes, I'm ok with that suggestion.  However, the idea behind returning `true` was that it should force the user of this API to do the right thing because the target is not apparently reachable (i.e., its a fail safe).  A fatal assert, however, would make it more clear to an implementation that it should be overriding this function.

compiler/aarch64/env/OMRCPU.cpp
0xdaryl: Agreed it is not necessary.  I'll fix up everywhere.

compiler/aarch64/codegen/OMRCodeGenerator.hpp
```diff
@@ -286,6 +286,18 @@ class OMR_EXTENSIBLE CodeGenerator : public OMR::CodeGenerator
...
+    * @brief Answers whether a trampoline is required for a direct call to reach a
+    *        target address from some source address.
+    *
+    * @param[in] targetAddress : the absolute address of the call target
+    * @param[in] sourceAddress : the absolute address from which the call displacement
+    *              range is measured.
+    *
+    * @return : true if a trampoline is required; false otherwise.
+    */
+   bool directCallRequiresTrampoline(intptrj_t targetAddress, intptrj_t sourceAddress);
```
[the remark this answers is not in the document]
0xdaryl: It does make sense to have a parameter with a clearly defined definition.  The CodeGenerator implementation on each platform should be able to determine what adjustments are needed.  In theory, all platforms except x86 should be passing in the instruction address already based on my understanding of how the displacements are calculated for their respective branch instructions.  X86 will be require a bit of rework that may reach into downstream projects.  I'll look into this deeper and report back.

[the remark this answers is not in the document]
0xdaryl: I've made changes to the PR where the sourceAddress of the `directCallRequiresTrampoline` is the address of the call instruction.  For the most part, no changes were required except on x86, and even in that case the changes were minimal.  I'll push those changes shortly.

@fjeremic : in the process of doing this I discovered a couple of places in the Z codegen where I believe the wrong source address is being used already to calculate whether a trampoline is needed.  Here are the two places that I found in the current code base:

https://github.com/eclipse/omr/blob/7a0ff628cb5c514ebaa88c1bb3d884fd8b4ca6fb/compiler/z/codegen/OMRSnippet.cpp#L142

https://github.com/eclipse/omr/blob/7a0ff628cb5c514ebaa88c1bb3d884fd8b4ca6fb/compiler/z/codegen/S390HelperCallSnippet.cpp#L113

In both cases, I believe `cursor` is two bytes off because it has already been advanced once the instruction opcode was written.  If you confirm this is an issue then I'll fix up as part of this work.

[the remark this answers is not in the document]
0xdaryl: I should point out that in both of the Z cases above the `cursor` was correctly adjusted during the actual encoding of the instruction (it was just the trampoline test where it is different).

compiler/arm/codegen/ARMBinaryEncoding.cpp
```diff
@@ -69,17 +69,14 @@ uint32_t encodeHelperBranchAndLink(TR::SymbolReference *symRef, uint8_t *cursor,
...
-   if (distance > BRANCH_FORWARD_LIMIT || distance < BRANCH_BACKWARD_LIMIT)
+   if (cg->directCallRequiresTrampoline(target, (intptrj_t)cursor))
       {
       target = cg->fe()->indexedTrampolineLookup(symRef->getReferenceNumber(), (void *)cursor);
-      TR_ASSERT(((int32_t)target - (int32_t)cursor) <= BRANCH_FORWARD_LIMIT &&
-             ((int32_t)target - (int32_t)cursor) >=  BRANCH_BACKWARD_LIMIT,
-             "CodeCache is more than 32MB.\n");
+
+      TR_ASSERT(TR::Compiler->target.cpu.isTargetWithinBranchImmediateRange(target, (intptrj_t)cursor),
+                "Target address is out of range.  CodeCache is more than 32MB.");
```
0dvictor: It seems the asserting condition does not have any indication that CodeCache must not be more than 32MB. It may be more clear removing "CodeCache is more than 32MB."

0xdaryl: Sure.

knn-k:
For AArch64, `TR::ARM64ImmSymInstruction::generateBinaryEncoding()` and `ARM64LabelInstruction::generateBinaryEncoding()` in ARM64BinaryEncoding.cpp can be rewritten using `directCallRequiresTrampoline()`.

0xdaryl:
I'm running some final, private stress tests now.  All look good so far.  I'll un-WIP this when they're done so hopefully this can be merged then (merge conflicts are killing me!)

0xdaryl:
Un-WIPing as my internal stress testing is clean.  I believe I've addressed all the comments.

PR #3480, opened 2019-01-20
WIP: Add new RawMemory allocators but don't use them yet

mstoodle:
This commit adds the new TR::RawAllocator hierarchy of classes used
for allocating TR::RawMemory (void *).

TR::RawAllocator is an abstract base class for all of these allocators
and  is the class to be used by the rest of the compiler (as opposed
to referencing a subclass directly). RawAllocator defines the following
basic API:
	allocate() to allocate requested size in bytes
	deallocate() to deallocate a previously allocated chunk of RawMemory
	protect() to remove read access permission to a previously
                  allocated chunk of RawMemory

TR::MallocAllocator implements the API using malloc() and free() from the
C library. It does not implement protect(). TR::MallocAllocator is the
equivalent of the current TR::RawAllocator class.

TR::DebugAllocator implements the API using e.g. mmap/munmap on platforms
that support it. The protect() function is also implemented. TR::DebugAllocator
is used when debugging use-after-free memory problems because it will
trap when memory is used after supposedly being freed.

TR::CustomAllocator takes arguments at construction time to implement the
API. It declares AllocateFunction, DeallocateFunction, and ProtectFunction
function types that match the signature of allocate(), deallocate(), and
protect(), respectively. TR::CustomAllocator offers more flexibility but
at the cost of an extra pointer indirection per call to the API.

These classes are all part of a refactoring of the lower level memory
allocation layers in the OMR compiler. They are not yet used by any
other code in OMR but will eventually replace TR::RawAllocator as well
as functionality in DebugSegmentAllocator.

The new memory allocator classes are created in compiler/env/newmemory.
When they are finally enabled by default, these files will move from
newmemory into compiler/env and replace the implementations there.
For now, the new files will not be referenced or even built by anything,
to avoid any functional issues before the integration code is completed.

### DIFF
--- a/compiler/env/newmemory/CustomAllocator.hpp
+++ b/compiler/env/newmemory/CustomAllocator.hpp
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#ifndef OMR_CUSTOMALLOCATOR_INCL
+#define OMR_CUSTOMALLOCATOR_INCL
+
+#pragma once
+
+#include <stddef.h>  // for size_t
+#include <cstdlib>   // for free, malloc
+#include <new>       // for bad_alloc, nothrow, nothrow_t
+#include "infra/Assert.hpp" //  for OMR_ASSERT_FATAL
+#include "env/newmemory/RawAllocator.hpp"
+
+// CustomAllocator is a more flexible implementation of TR::RawAllocator
+// but at the performance cost of an extra pointer indirection for every
+// API call. Upon construction, the functions to call to implement
+// allocate() (AllocatFunction), deallocate (DeallocateFunction), and
+// optionally noLongerUsed() (NoLongerUsedFunction) must be specified.
+
+namespace OMR
+{
+
+class CustomAllocator :  public TR::RawAllocator
+   {
+public:
+   typedef RawMemory (*AllocateFunction)(size_t, RawMemory);
+   typedef void (*DeallocateFunction)(RawMemory);
+   typedef void (*NoLongerUsedFunction)(RawMemory, size_t);
+
+   CustomAllocator(AllocateFunction allocateFunction,
+                   DeallocateFunction deallocateFunction,
+                   NoLongerUseFunction noLongerUsedFunction=0)
+      : _allocateFunction(allocateFunction)
+      , _deallocateFunction(deallocateFunction)
+      , _noLongerUsedFunction(noLongerUsedFunction)
+      {
+      OMR_ASSERT_FATAL(_allocateFunction && _deallocateFunction, "Custom allocator needs valid allocateFunction and deallocateFunction");
+      }
+
+   virtual TR::RawAllocator & clone()
+      {
+      return *(new (*this) CustomAllocator(_allocateFunction, _deallocateFunction, _noLongerUsedFunction));
+      }
+
+   virtual RawMemory allocate(size_t size, const std::nothrow_t tag, void * hint = 0) const throw()
+      {
+      return (*_allocateFunction)(size, hint);
+      }
+
+   virtual void deallocate(RawMemory p) const throw()
+      {
+      (*_deallocateFunction)(p);
+      }
+
+   virtual void noLongerUsed(RawMemory p, size_t size) const throw()
+      {
+      if (_noLongerUsedFunction)
+         (*_noLongerUsedFunction)(p, size);
+      }
+
+private:
+   AllocateFunction   _allocateFunction;
+   DeallocateFunction _deallocateFunction;
+   NoLongerUsedFunction    _noLongerUsedFunction;
+   };
+
+} // namespace OMR
+
+#endif // OMR_CUSTOMALLOCATOR_INCL
--- a/compiler/env/newmemory/DebugAllocator.cpp
+++ b/compiler/env/newmemory/DebugAllocator.cpp
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception [1] and GNU General Public
+ * License, version 2 with the OpenJDK Assembly Exception [2].
+ *
+ * [1] https://www.gnu.org/software/classpath/license.html
+ * [2] http://openjdk.java.net/legal/assembly-exception.html
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+ *******************************************************************************/
+
+#if (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX)
+#include <sys/mman.h>
+#if defined(__APPLE__) || !defined(MAP_ANONYMOUS)
+#define MAP_ANONYMOUS MAP_ANON
+#endif
+#elif defined(OMR_OS_WINDOWS)
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#else
+#include <string.h>
+#endif /* defined(OMR_OS_WINDOWS) */
+
+#include <stddef.h>  // for size_t
+#include <cstdlib>   // for free, malloc
+#include <new>       // for bad_alloc, nothrow, nothrow_t
+#include "infra/Assert.hpp"
+#include "env/mem/DebugAllocator.hpp"
+
+void *
+OMR::DebugAllocator::allocate(size_t size, const std::nothrow_t tag,void * hint) const throw()
+   {
+#if (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX)
+   void *p = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+   return (p == MAP_FAILED) ? NULL : p;
+#elif defined(OMR_OS_WINDOWS)
+   return  VirtualAlloc(NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+#else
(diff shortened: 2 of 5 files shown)


compiler/env/newmemory/DebugAllocator.cpp
```diff
@@ -0,0 +1,87 @@
...
+void *
+TR::DebugAllocator::allocate(size_t size, const std::nothrow_t tag,void * hint) const throw()
+   {
+#if (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX)
+   void *p = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+   return (p == MAP_FAILED) ? NULL : p;
+#elif defined(OMR_OS_WINDOWS)
+   return  VirtualAlloc(NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+#else
+   return this->TR::RawAllocator::allocate(size, hint);
```
Leonardo2718: I don't think this can work. If the intention is to cause a failure, can `static_assert` be used instead?

compiler/env/newmemory/DebugAllocator.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+void
+TR::DebugAllocator::deallocate(void* p) const throw()
+   {
+#if (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX)
+   TR_ASSERT(0, "DebugAllocator does not support deallocation without size parameter on this platform");
+#elif defined(OMR_OS_WINDOWS)
+   VirtualFree(p, 0, MEM_RELEASE);
+#else
+   this->TR::RawAllocator::deallocate(p);
```
Leonardo2718: Same comment as for `allocate()`.

compiler/env/newmemory/DebugAllocator.cpp
```diff
@@ -0,0 +1,87 @@
...
+#else
+   return this->TR::RawAllocator::allocate(size, hint);
+#endif /* (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX) */
+   }
+
+void
+TR::DebugAllocator::deallocate(void* p) const throw()
+   {
+#if (defined(LINUX) && !defined(OMRZTPF)) || defined(__APPLE__) || defined(_AIX)
+   TR_ASSERT(0, "DebugAllocator does not support deallocation without size parameter on this platform");
```
Leonardo2718: It seems like this could be problematic when debugging. Would it be possible to make it so that users must always pass a size and avoid this problem altogether?

compiler/env/newmemory/CustomAllocator.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace TR {
+
+// CustomAllocator is a more flexible implementation of TR::RawAllocator
+// but at the performance cost of an extra pointer indirection for every
+// API call. Upon construction, the functions to call to implement
+// allocate() (AllocatFunction), deallocate (DeallocateFunction), and
+// optionally protect() (ProtectFunction) must be specified.
+
+class CustomAllocator :  public TR::RawAllocator
```
Leonardo2718: I'm not sure I understand the purpose of this class. If a user wanted to use custom memory management functions, would it not be easier for them to just created a new sub-class of `TR::RawAllocator`?

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+class RawAllocator
+   {
+public:
+   RawAllocator()
+      { }
+
+   // virtual "copy constructor" used primarily to simplify lifetime management for RawAllocator objects
+   virtual TR::RawAllocator & clone() = 0;
+
+   virtual RawMemory allocate(size_t size, const std::nothrow_t tag, void * hint = 0) const throw() =  0;
```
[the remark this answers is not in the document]
mstoodle: yep

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+   RawMemory allocate(size_t size, void * hint = 0) const
```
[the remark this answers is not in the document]
mstoodle: yep

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+      {
+      void * const alloc = allocate(size, std::nothrow, hint);
+      if (!alloc) throw std::bad_alloc();
+      return alloc;
+      }
+
+   virtual void deallocate(RawMemory p) const throw() = 0;
+
+   // default simplification: deallocate() does not need a size
+   virtual void deallocate(RawMemory p, const size_t size) const throw()
```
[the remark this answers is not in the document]
mstoodle: it was like that when I got here :) . I'll try to track down why it exists.

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+   // Try to make the given RawMemory region inaccessible so future access will trap
+   // Not all subclasses will override this empty implementation
+   virtual void protect(RawMemory p, size_t size) const throw()
+      { }
+
+   friend bool operator ==(const RawAllocator &left, const RawAllocator &right)
+      {
+      return true;
```
vijaysun-omr: Is ```true``` the best default one can have here ?

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+   // default simplification: deallocate() does not need a size
+   virtual void deallocate(RawMemory p, const size_t size) const throw()
+      {
+      deallocate(p);
+      }
+
+   // Try to make the given RawMemory region inaccessible so future access will trap
+   // Not all subclasses will override this empty implementation
+   virtual void protect(RawMemory p, size_t size) const throw()
```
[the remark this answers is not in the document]
mstoodle: The purpose of protect is so that memory allocation, deallocation, and protection can have consistent implementations in some particular RawAllocator. I suppose another option would be for *only* TR::DebugAllocator to have a protect function, but that may require upcasting somewhere else in the compiler from TR::RawAllocator. The intention was for the entire rest of the compiler to consistently use TR::RawAllocator as its interface.

compiler/env/newmemory/DebugAllocator.cpp
mstoodle: That is another option, but the option still exists to paint the memory in `prrotect()` even if it won't fault on re-use. Asserting here would prevent that option.

compiler/env/newmemory/DebugAllocator.cpp
mstoodle: and same response :) .

compiler/env/newmemory/DebugAllocator.cpp
mstoodle: it's the way the code works right now. We can change how it works later; I would prefer to keep it as functionally similar as possible for now.

compiler/env/newmemory/CustomAllocator.hpp
mstoodle: they could do that, yes. OpenJ9 has to call malloc and free via function pointers anyway for its "debugger" scenario, so I generalized it to allow any function pointers to be used. I felt that was less specialized.

compiler/env/newmemory/DebugAllocator.cpp
Leonardo2718: I'm not sure I follow. Calling `TR::RawAllocator::allocate(size_t, void *)` will call `TR::RawAllocator::allocate(size_t, nothrow_t, void*)`, which will be delegated back to `TR::DebugAllocator::allocate(size_t, nothrow_t, void *)`, resulting in an infinit loop... unless I've missed something?

There's also the issue of calling something that can throw an exception inside a `throw ()` function without handling the exception.

compiler/env/newmemory/DebugAllocator.cpp
Leonardo2718: In this case, `TR::RawAllocator::deallocate(RawMemory)` has no implementation so calling it is undefined behaviour (though it will most likely result in a linking error).

compiler/env/newmemory/RawAllocator.hpp
mstoodle: Actually, it was a better answer before we had RawAllocator subclasses. This answer tries to encode the "instances of RawAllocator are equivalent" statement. But with subclasses in the mix, it is not actually the safest answer (although, in any one execution run, it is reasonable that only one sublass would ever be created, so the `return true` is *probably* a reasonable answer). I'll think about this one more.

compiler/env/newmemory/MallocAllocator.hpp
```diff
@@ -0,0 +1,62 @@
...
+   {
+public:
+   MallocAllocator()
+      : TR::RawAllocator()
+      { }
+
+   // virtual "copy constructor" used primarily to simplify lifetime management for RawAllocator objects
+   virtual TR::RawAllocator & clone()
+      {
+      return *(new (*this) TR::MallocAllocator());
```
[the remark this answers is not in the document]
mstoodle: > is the expectation that cloned instances must be manually freed?

yes, and it should be documented. Cloning is used to remove implicit lifetime assumptions (so any object O that keeps a reference to a RawAllocator should actually clone that RawAllocator and clean the instance up when O's lifetime ends).

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+      { }
+
+   friend bool operator ==(const RawAllocator &left, const RawAllocator &right)
```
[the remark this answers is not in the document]
mstoodle: yeah, they made more sense in the original code where there weren't subclasses of RawAllocator. I think I'll remove these.

compiler/env/newmemory/RawAllocator.hpp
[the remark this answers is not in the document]
mstoodle: There are quite a few places in the compiler that pass a size to a `deallocate()` function, though I suspect few of them go all the way through to the `RawAllocator`. Possibly this is a hanger-on from `cs2` allocators whose deallocate function takes a size (and an optional name!).

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,133 @@
...
+//
+
+class RawAllocator
+   {
+public:
+   RawAllocator()
+      { }
+
+   // virtual "copy constructor" used primarily to simplify lifetime management for RawAllocator objects
+   virtual TR::RawAllocator & clone() = 0;
```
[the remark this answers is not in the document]
mstoodle: i don't think it *has* to be removed, though if anyone uses it they will want to be careful (just like with any other object referred to via its (abstract in this case) base class).

compiler/env/newmemory/DebugAllocator.cpp
mstoodle: See comment above: `TR::DebugAllocator` is supposed to be a subclass of `TR::MallocAllocator` which I think resolves this issue. Note for interested readers:`TR::MallocAllocator` used to be in `TR::RawAllocator` :)

compiler/env/newmemory/CustomAllocator.hpp
mstoodle: Well, OMR doesn't *need* a CustomRawAllocator, but I guess I thought it might be easier for someone to show up and create a CustomRawAllocator rather than write up a new class, figure out where to add new files to build for which build system they choose to use.

Another motivating example: if the JIT is compiled into a dynamic library, then you'll only have pointers to the allocate/free functions anyway (assuming they're not in the JIT's library, which is the case in OpenJ9), so why force everyone to create a new class to enable dynamic linking?

compiler/env/newmemory/RawAllocator.hpp
mstoodle: I've removed these functions

compiler/env/newmemory/DebugAllocator.cpp
mstoodle: yean, that's wrong. `TR::DebugAllocator `should extend `TR::MallocAllocator` and those calls should go back to `TR::MallocAllocator` <blush>

mstoodle: good catch on the missing `tag` parameter...i've added it to the call. Thanks!

compiler/env/newmemory/RawAllocator.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+   // allocate() tries to allocate a RawMemory block that can hold at least size bytes.
+   // On failure, allocate() returns NULL.
+   virtual RawMemory allocate(size_t size, const std::nothrow_t tag, void * hint = 0) const throw() =  0;
+
+   // allocate() tries to allocate a RawMemory blocvk that can hold at least size bytes.
```
Leonardo2718: Typo:
```suggestion
   // allocate() tries to allocate a RawMemory block that can hold at least size bytes.
```

mstoodle:
Apologies for the churn, but I'm withdrawing this PR. Some changes were necessary to handle the requirements imposed upon our allocation strategy by the STL (now wishing even more that we hadn't gone down this path). I'll be submitting another PR hopefully soon (I have it building and passing tests for OMR now, but still need to make sure it can accommodate existing downstream projects like OpenJ9).

PR #3481, opened 2019-01-21
Add missing RememberedSet structures from OpenJ9

jason-hall:
Code which uses structure J9VMGCRememberedSet has been added to OMR, but not the structure. The intended plan is to add MM_GCRememberedSet to OMR and then replace the J9VMRememberedSet structure in OpenJ9 with a typedef to MM_GC_RememberedSet. This should allow changing usage of J9VMRememberedSet to MM_GCRememberedSet in OMR and OpenJ9 separately without breaking builds.

Signed-off-by: Jason Hall <jasonhal@ca.ibm.com>

### DIFF
--- a/include_core/j9nongenerated.h
+++ b/include_core/j9nongenerated.h
@@ -212,4 +212,22 @@ typedef struct J9MemorySegmentList {
 	uintptr_t flags;
 } J9MemorySegmentList;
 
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

PR #3483, opened 2019-01-22
Set CURRENTLY REFERENCED object state atomically

dmitripivkine:
Currently in MM_Scavenger::rescanThreadSlot() there is two stage process
to set Remembered State to object header and add object to Remembered
Set:
1) call rememberObject() would try to change Remembered State from 0
(not remembered) to STATE_REMEMBERED atomically and add object to the
Remembered Set if succeeds.
2) upgrade STATE_REMEMBERED to
OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED non-atomically

I believe an upgrade to OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED
might be lost on Weak Ordering platforms and might be a reason for
premature removal of the object from RS.
 
So proposed fix sets OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED to
the object header atomically. If replaced (old) Remembered State was not
0 (object has been added to RS already) no further action necessary.
If object has not been remembered before it should be added to RS

### DIFF
--- a/gc/base/standard/Scavenger.cpp
+++ b/gc/base/standard/Scavenger.cpp
@@ -2756,8 +2756,16 @@ MM_Scavenger::rescanThreadSlot(MM_EnvironmentStandard *env, omrobjectptr_t *obje
 			Assert_MM_true(!isObjectInNewSpace(tenuredObjectPtr));
 
 			*objectPtrIndirect = tenuredObjectPtr;
-			rememberObject(env, tenuredObjectPtr);
-			_extensions->objectModel.setRememberedBits(tenuredObjectPtr, OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED);
+
+			/*
+			 * This call sets OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED Remembered state in the object header:
+			 * - if object has any Remembered state set already (it means object has been added to RS) just upgrade it to OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED
+			 * - if object has no Remembered state set add object to the Remembered Set here.
+			 */
+			if (_extensions->objectModel.atomicSwitchReferencedState(tenuredObjectPtr, OMR_TENURED_STACK_OBJECT_CURRENTLY_REFERENCED)) {
+				/* Allocate an entry in the remembered set */
+				addToRememberedSetFragment(env, tenuredObjectPtr);
+			}
 		}
 	}
 }


dmitripivkine:
@charliegracie @amicic Would you please review?

charliegracie:
@genie-omr build all

amicic:
Some quick thoughts (without completely understanding the nature of the problem)
a) if this problem is weak memory model related, I'd expect a fix to deal with store (or read) fence (memory barrier). Doing atomic store on a memory location will effectively do full memory barrier on that same location, but generally speaking doing an atomic to produce fence effect is overkill (and plus it would not be as clear why that atomic is done, as if fence is done)
b) if the problem is race related, this fix looks reasonable. But then, it would imply 1) the problem should be visible on non-weak memory models and 2) there might be ways to improve reproducibility of the problem by injecting some short pauses between the atomic remember and non-atomic state change.

All in all, I'd like to understand better the nature of the problem (who can try to read that new state, but incorrectly conclude it's not set) to properly judge the fix,

amicic:
Looked a bit more trying to understand how CURRENTLY_REFERENCED state might be (temporarily) not visible or (permanently) lost 

1) if it is weak memory related... there should be another thread (actually CPU) consuming this memory location (the object header) but not see the up-to-date state. The first consumer of this, after the thread rescan is processReferenceSlot in pruning pass. It's guarantied to have a list one lock related operation (for the thread level synchronization)  in-between that will do a store fence. Hence pruning pass should see up-to-date value.
2) if it is race related... there should be two threads racing to change the value of the header. If two threads are doing the same steps (auto-remembering it), the state will not be lost - the slower one is guarantied to set the state to CURRENTLY_REFERENCED (and also faster one will end up adding it to RS) I also don't see that any other thread will try to change the header with some other values/flags (object is already copied into tenure and header flags are all set).

Please enlighten me with a specific scenario, where the problem is...

dmitripivkine:
We do have atomic state change in current code already: from "not remembered" to "regular remembered" (with adding of object to RS if won). After this next line changes state from "regular remembered" to "currently referenced remembered" non-atomically (without fence). I am not sure which exactly side effects this sequence might have as well as I can not reproduce a problem. However it is obvious that remembered state might be changed from "not remembered" to "currently referenced remembered" atomically directly (if won, with adding element to RS) or just upgrade "regular remembered" to "currently referenced remembered" otherwise. As far as we do have right helper already (used in barrier code) and I do not introduce extra atomic operation I believe it reasonable to change code as a part of general cleaning (even this PR does not fix anything). I doubt it is reasonable to try to reproduce original problem. Also I am not suggesting this is necessarily the fix for it.

>All in all, I'd like to understand better the nature of the problem (who can try to read that new state, but incorrectly conclude it's not set) to properly judge the fix

If somehow "currently referenced remembered" might be lost (just have "regular remembered" instead) it might be a reason for premature removal of object from RS. The goal of currently/recently extended remembered states is artificially prolong time of remembering the object. As I understand the barriers on some objects (for example thread stack references) might be omitted intentionally, so this is a mechanism to keep them alive.

dmitripivkine:
>if it is race related... there should be two threads racing to change the value of the header. If two threads are doing the same steps (auto-remembering it), the state will not be lost - the slower one is guarantied to set the state to CURRENTLY_REFERENCED (and also faster one will end up adding it to RS) I also don't see that any other thread will try to change the header with some other values/flags (object is already copied into tenure and header flags are all set).

I think (if it is a problem) it is race related. I agree it is not clear how exactly it might fail (if it might) but there is a way to avoid questionable code by going straight to desired state.

amicic:
Still don't see if this fixes anything, but I do agree that code looks more reasonable.

atomicSwitchReferencedState appears to do the right thing. Besides changing the state to currently remembered, it returns false if object is already remembered, but in any other remembered states (plain remembered and recently remembered), which would prevent adding to RS again. atomicSetRememberedState, which is similar, would be incorrect to use - it would refuse to change the state if already in those two remembered states (which tells me atomicSetRememberedState, should dont even take an argument - it should and indeed is only used to set plain remembered state).

charliegracie:
I believe this change is technically correct and cleaner than the previous code.  I am not sure it is resolving any issues but I think it is a code improvement anyway.

charliegracie:
@amicic are you ok with these changes as-is to cleanup the code?

amicic:
yes, the change  looks good

PR #3486, opened 2019-01-22
Remove unused _start_time field from FEBase

0xdaryl:
Not used in OMR, and known downstream projects that create a FrontEnd.


0xdaryl:
@genie-omr build all

PR #3488, opened 2019-01-22
Consolidate 64-bit vs. 32-bit virtual register handling on Z codegen

fjeremic:
Due to the sheer size of the change it has been broken up into logical pieces each of which is in a separate commit to make reviews and code archeology easier to manage. Each commit body details what and why the change was made.

At a high level this change consolidates how we handle 64-bit virtual registers in the Z codegen. Previously we had two ways of making this distinction:

1. Using the `Is64BitReg` register flag
2. Using the `TR_GPR64` register kind

This change deprecates the `TR_GPR64` register kind in favor of the `Is64BitReg` register flag. The reason behind this choice is that the `TR_GPR64` register kind was "bolted on" to the Z register allocator at the same time we added support for using 64-bit registers on 32-bit. The register kind was only meant to be used by the 32-bit targets. The 64-bit targets were completely unaware of the `TR_GPR64` register kind, because they used a different mechanism to tell apart 64-bit vs. 32-bit virtual registers.

This caused all sorts of problems where the register allocator would get "out of sync" when both of these concepts were used to track 64-bit virtual registers. A catalyst which uncovered all these issues was enabling the use of 64-bit registers on 32-bit as part of #2916.

In the new world both 32-bit and 64-bit targets will use the `Is64BitReg` register flag to determine whether a particular virtual register holds a 32-bit or a 64-bit value. All registers allocated are assumed to be 32-bit. When a register is used within a 64-bit instruction (as marked in OMRInstOpCodeProperties.hpp) the virtual register will automatically be marked as a 64-bit register. Outside from this the developer need not care whether a virtual register is 32-bit or 64-bit as everything is handled automatically. The only exceptions to this is around special evaluators such as the GRA evaluators (`lRegLoad`, `aRegLoad`, etc.) and the extension/narrowing evaluators (`l2i`, `a2l`, etc.).


fjeremic:
Given the Z-specifics of this change and the sheer breadth I don't expect any one single committer to fully review the change so I'm kindly requesting reviews from @joransiu, @NigelYiboYu, @r30shah, @dchopra001, and @sehirst if you are familiar with any piece of the register allocator which I changed.

fjeremic:
I'm going to begin doing some self reviews now to help the committers/reviewers understand why certain changes were made.

compiler/codegen/CodeGenRA.cpp
```diff
@@ -452,8 +452,8 @@ OMR::CodeGenerator::allocateSpill(bool containsCollectedReference, int32_t *offs
 TR_BackingStore *
 OMR::CodeGenerator::allocateSpill(int32_t dataSize, bool containsCollectedReference, int32_t *offset, bool reuse)
    {
-   TR_ASSERT(dataSize <= 16, "assertion failure");
-   TR_ASSERT(!containsCollectedReference || (dataSize == TR::Compiler->om.sizeofReferenceAddress()), "assertion failure");
+   TR_ASSERT_FATAL(dataSize <= 16, "assertion failure");
+   TR_ASSERT_FATAL(!containsCollectedReference || (dataSize == TR::Compiler->om.sizeofReferenceAddress()), "assertion failure");
```
fjeremic: A lot of the assertions in RA have been made fatal to catch problems at compile time rather than exhibit weird runtime crashes. The cost for these assertions is so small that it is definitely worth while to enable them.

compiler/codegen/CodeGenRA.cpp
```diff
@@ -700,18 +698,10 @@ TR::Register * OMR::CodeGenerator::allocateRegister(TR_RegisterKinds rk)
    self()->addAllocatedRegister(temp);
    if (self()->getDebug())
       self()->getDebug()->newRegister(temp);
-   if (  rk == TR_GPR
-      && (  !self()->getDebug()
-         || !self()->getTraceRAOption(TR_TraceRASpillTemps)
-         || performTransformation(self()->comp(), "O^O SPILL TEMPS: Set UpperHalfIsDead on %s\n", self()->getDebug()->getName(temp)))) // allocateRegister is called for the vmthread during initialization when getDebug has not been initialized yet
-      {
-      temp->setIsUpperHalfDead();
```
fjeremic: These weren't used anywhere so we removed them. It had a few dependencies on some x86 code which was dead as a result.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -1142,7 +1138,7 @@ OMR::CodeGenerator::getNumberOfGlobalRegisters()
 #ifdef TR_HOST_S390
 uint16_t OMR::CodeGenerator::getNumberOfGlobalGPRs()
    {
-   if (self()->supportsHighWordFacility() && !self()->comp()->getOption(TR_DisableHighWordRA))
+   if (self()->supportsHighWordFacility())
```
fjeremic: Pretty much all the code in the Z codegen checked for both whether we have high word support and whether the option was not enabled. Checking the option is rather expensive as it usually involves a few indirections. We changed this to only check the option when we set that we have high word support and if the option is specified we simply don't enable support.

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -1171,18 +1167,6 @@ TR::Register *OMR::CodeGenerator::allocateSinglePrecisionRegister(TR_RegisterKin
    return temp;
    }
 
-TR::Register *OMR::CodeGenerator::allocate64bitRegister()
```
fjeremic: There is no longer a difference in the `allocateRegister` APIs. All registers are assumed to be 32-bit unless used in a 64-bit context. These are all deprecated. There is likely some code which used the `allocate64bitRegister` API but was changed to use the `allocateRegister` API since they are synonymous now. This results in a bunch of code which looks like this:

```
if (<some condition>) {
    reg = cg->allocateRegister();
} else {
    reg = cg->allocateRegister(); // Used to be allocate64bitRegister
}
```

Such code should be folded away if you see it in an area you're working with.

compiler/codegen/OMRRegister.hpp
```diff
@@ -187,7 +184,6 @@ class OMR_EXTENSIBLE Register
       IsLive                        = 0x0010, // Register is currently live
       ContainsInternalPointer       = 0x0080,
       IsSinglePrecision             = 0x0400,
-      UpperHalfIsDead               = 0x0400, // GPRS ONLY -- no need to save/restore the high-order bits when spilled; aliased with IsSinglePrecision
```
fjeremic: These two (`UpperHalfIsDead` and `IsSinglePrecision`) are aliased for some reason. I'm not sure if this is intentional or if this is mistake. Regardless I've checked that the `IsSinglePrecision` flag is only used for FPRs while the `UpperHalfIsDead` flag only affects GPRs so they shouldn't interfere.

compiler/il/OMRNode.cpp
```diff
@@ -2829,21 +2829,12 @@ OMR::Node::isRematerializable(TR::Node *parent, bool onlyConsiderOpCode)
    return false;
    }
 
-
-
 bool
 OMR::Node::canEvaluate()
```
fjeremic: Not sure what the purpose of this API was. It returns true on all paths. Sounds like it should be deprecated.

compiler/z/codegen/BinaryAnalyser.cpp
```diff
@@ -292,36 +292,26 @@ TR_S390BinaryAnalyser::genericAnalyser(TR::Node * root,
...
-   TR::Node * firstChild;
-   TR::Node * secondChild;
    TR::Instruction * cursor = NULL;
    TR::RegisterDependencyConditions * dependencies = NULL;
    bool setsOrReadsCC = NEED_CC(root) || (root->getOpCodeValue() == TR::lusubb);
    TR::InstOpCode::Mnemonic regToRegOpCode;
    TR::InstOpCode::Mnemonic memToRegOpCode;
    TR::Compilation *comp = TR::comp();
 
-   if (TR::Compiler->target.is64Bit() || cg()->use64BitRegsOn32Bit())
```
fjeremic: You'll see a lot of these. On 64-bit `use64BitRegsOn32Bit` will always return false, and on 32-bit `use64BitRegsOn32Bit` will always return `true` so the query above will be true in all cases so we can fold away the resulting code.

compiler/z/codegen/BinaryCommutativeAnalyser.hpp
```diff
@@ -69,20 +69,11 @@ class TR_S390BinaryCommutativeAnalyser : public TR_Analyser
...
-                            TR::InstOpCode::Mnemonic lowMemToRegOpCode,
-                            TR::InstOpCode::Mnemonic highMemToRegOpCode,
-                            TR::InstOpCode::Mnemonic copyOpCode);
-
    void integerAddAnalyser(TR::Node       *root,
                            TR::InstOpCode::Mnemonic regToRegOpCode,
                            TR::InstOpCode::Mnemonic memToRegOpCode,
                            TR::InstOpCode::Mnemonic copyOpCode);
 
-   void longAddAnalyser(TR::Node *root, TR::InstOpCode::Mnemonic copyOpCode);
```
fjeremic: These analyzers become dead as they're only used for when 64-bit values must go into register pairs, which is no longer true on any target.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -1881,15 +1379,7 @@ TR::Register *
...
-
-   if (TR::Compiler->target.is64Bit() || cg->use64BitRegsOn32Bit())
-      {
-      return laddHelper64(node, cg);
-      }
-   else
-      {
-      return laddHelper(node, cg);
-      }
+   return laddHelper64(node, cg);
```
fjeremic: We need to rename these to remove the "64".

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -801,7 +796,7 @@ OMR::Z::CodeGenerator::CodeGenerator()
 bool
 OMR::Z::CodeGenerator::getSupportsBitPermute()
    {
-   return TR::Compiler->target.is64Bit() || self()->use64BitRegsOn32Bit();
+   return true;
```
fjeremic: We should be able to fold this now.

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -2583,33 +2558,12 @@ OMR::Z::CodeGenerator::doRegisterAssignment(TR_RegisterKinds kindsToAssign)
...
-            rks = (TR_RegisterKinds) ((int)rks | TR_HPR_Mask);
+         TR_RegisterKinds rks = (TR_RegisterKinds)(TR_GPR_Mask | TR_HPR_Mask | TR_FPR_Mask | TR_VRF_Mask);
 
          self()->getDebug()->startTracingRegisterAssignment("backward", rks);
          }
       }
 
-   // pre-pass to set all internal control flow reg deps to 64bit regs
-   TR::Instruction * currInst = instructionCursor;
-   if (self()->supportsHighWordFacility() && !self()->comp()->getOption(TR_DisableHighWordRA) && TR::Compiler->target.is64Bit())
```
fjeremic: I can't figure out why this code was here in the first place. I can't think of why anything can go wrong so I've removed it. My testing shows there is no issue with this. Perhaps there was an age old bug which was fixed and this code was just left in place?

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -111,136 +111,105 @@ getRegisterName(TR::Register * reg, TR::CodeGenerator * cg)
...
-   TR_Debug * debugObj = cg->getDebug();
+   // TODO: In reality we should be validating that the the source is NULL and target is non-NULL however we cannot
+   // currently do this because the registerCopy and registerExchange APIs are not updating the register states as
+   // registers are shuffled around.
+   //
+   // A typical example of when this assert would fire is when attempting a register exchange (which uses this API)
+   // via a spare reigster. We will generate 3 register shuffles but the register state is only updated later, thus
+   // during the second shuffle the spare register will the the target in the above parameter and as such we would
+   // assert since the spare register presumably did not have any virtual register associated with it.
+   TR_ASSERT_FATAL(sourceReg->getAssignedRegister() != NULL || targetReg->getAssignedRegister() != NULL, "Attempting register copy with source (%s) and target (%s) real registers not corresponding to any virtual register", getRegisterName(sourceReg, cg), getRegisterName(targetReg, cg));
```
fjeremic: I'll be working on this next, following the merge of this PR.

compiler/z/codegen/OMRMachine.cpp
```diff
@@ -1927,46 +1782,30 @@ bool
...
 
-         //self()->cg()->traceRegWeight(_registerFile[i], _registerFile[i]->getWeight());
-         //self()->cg()->traceRegWeight(_registerFile[i + 1], _registerFile[i + 1]->getWeight());
-
-         if (enableHighWordRA && rk == TR_GPR64)
+         // TODO: This is a little too conservative. This API needs to be taught whether the register pair requires
+         // 64-bit registers or not and if and only if we require 64-bit registers do we need to check the highword.
+         bool highWordPairIsFree = false;
+         if ((_registerFile[i + 0]->getHighWordRegister()->getState() == TR::RealRegister::Free || _registerFile[i + 0]->getHighWordRegister()->getState() == TR::RealRegister::Unlatched) &&
+             (_registerFile[i + 1]->getHighWordRegister()->getState() == TR::RealRegister::Free || _registerFile[i + 1]->getHighWordRegister()->getState() == TR::RealRegister::Unlatched))
```
fjeremic: Lots of opportunity for improvement. I've left TODOs all around which I'll be addressing after this PR is merged.

compiler/z/codegen/OMRRegister.cpp
```diff
@@ -70,8 +70,12 @@ void
 OMR::Z::Register::setPlaceholderReg()
    {
 #if defined(TR_TARGET_64BIT)
-   self()->setIs64BitReg(true);
+   if (self()->getKind() == TR_GPR)
+      {
+      self()->setIs64BitReg(true);
+      }
```
fjeremic: We added an assert to make sure the `is64BitReg()` query is only used for GPRs, however we do have placeholder FPR registers so we avoid setting this flag unless it's a GPR.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -4758,32 +4567,12 @@ genericLoad(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference * tempM
 
 
 /**
- * This monster function handles all loads on zos and there are many, many variations. If we had a guarantee that we are executing
- * on GoldenEagle or better, this function would had been just couple of lines to pick the right instruction of the loadInstr
- * array(!!!):
```
fjeremic: Finally! I've gone ahead and did this and indeed this template function is very easy to implement.

0xdaryl:
@genie-omr build all

compiler/codegen/RegisterConstants.hpp
```diff
@@ -39,7 +39,6 @@ enum TR_RegisterKinds
    TR_VRF           = 4,
    TR_VSX_SCALAR    = 5,
    TR_VSX_VECTOR    = 6,
-   TR_GPR64         = 7,
    TR_SSR           = 8,   // used for TR_PseudoRegisters for SS (storage to storage) instructions to return results
```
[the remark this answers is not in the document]
fjeremic: Fixed in fc05273

compiler/optimizer/OMRSimplifierHandlers.cpp
```diff
@@ -11619,7 +11619,7 @@ TR::Node *lxorSimplifier(TR::Node * node, TR::Block * block, TR::Simplifier * s)
       }
 
 
-   // For 32 bit platforms (p and z), evaluator due to historical reasons, evaluator for rotate does not support 64 bit regs on a 32 bit platform by default
+   // For 32 bit platforms (p), evaluator due to historical reasons, evaluator for rotate does not support 64 bit regs on a 32 bit platform by default
```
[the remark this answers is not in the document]
fjeremic: Fixed in 0ed7512

compiler/il/OMRNode.cpp
fjeremic: Fixed in 67da1aa

fjeremic:
All reviews so far have been addressed.

0xdaryl:
Can you fix the copyrights on OMRLocalCSE.cpp and PartialRedundancy.cpp please?  Otherwise it looks good to me.

fjeremic:
> Can you fix the copyrights on OMRLocalCSE.cpp and PartialRedundancy.cpp please? Otherwise it looks good to me.

Thanks. Fixed as part of 35ce2d8

fjeremic:
@genie-omr build all

joransiu:
Went through the commits and the changes look good to me.  

I suspect the " get(Used|Defined|Killed)Registers" methods used by the Instruction Scheduler in the past.   Do we envision any future uses of this type of data from an Instruction (maybe some peephole)?

I do want to echo Daryl's point about stress testing the changes in this PR.

Thanks for the thorough job of cleaning and simplifying our 64-bit virtual register handling!

fjeremic:
> I do want to echo Daryl's point about stress testing the changes in this PR.

Yeah I've been thoroughly testing them for the last few weeks in OpenJ9. Most of the bugs (pre-existing and ones I've introduced) were caught by the various fatal asserts I've enabled and added. So far my personal testing looks clean. I suppose once all the reviews are complete I'll run another giant batch. I've also been more or less daily running builds against these changes to hopefully find any intermittent bugs. So far nothing has arisen though.

compiler/optimizer/OMRLocalCSE.cpp
```diff
@@ -114,13 +114,7 @@ bool OMR::LocalCSE::shouldCopyPropagateNode(TR::Node *parent, TR::Node *node, in
 
 bool OMR::LocalCSE::shouldCommonNode(TR::Node *parent, TR::Node *node)
    {
-   if (!isTreetopSafeToCommon())
-      return false;
-
-   if (node->getType().isAggregate())
```
vijaysun-omr: Is this aggregate condition removed intentionally ?

vijaysun-omr:
I'm fine with these changes (I only reviewed the common code parts) and the other reviewers have also signed off. Are you planning to do more testing before you would like this merged @fjeremic ?

fjeremic:
> I'm fine with these changes (I only reviewed the common code parts) and the other reviewers have also signed off. Are you planning to do more testing before you would like this merged @fjeremic ?

I'll do two more rounds of testing today and write back with an all clear if everything goes well.

fjeremic:
@vijaysun-omr we are looking good. All testing is green. You have the all clear. OpenJ9 change which depends on this one is clean as well and ready for merge.

PR #3493, opened 2019-01-22
Move scavenger-related CLI functionality to new ScavengerDelegate

rwy0717:
This is a first pass. The example and the empty glue version of the
delegate are implemented. The code changes are under the
OMR_GC_SCAVENGER_DELEGATE ifdef, and disabled by default.

### DIFF
--- a/example/glue/ScavengerDelegate.cpp
+++ b/example/glue/ScavengerDelegate.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2019 IBM Corp. and others
+ * Copyright (c) 2019, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -19,9 +19,14 @@
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
  *******************************************************************************/
 
+#include "omrcfg.h"
+
+#if defined(OMR_GC_SCAVENGER_DELEGATE)
+
+#include "ScavengerDelegate.hpp"
+
 #include "j9nongenerated.h"
 #include "modronbase.h"
-
 #if defined(OMR_GC_MODRON_CONCURRENT_MARK)
 #include "CardTable.hpp"
 #endif /* OMR_GC_MODRON_CONCURRENT_MARK */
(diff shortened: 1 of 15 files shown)


example/glue/ScavengerDelegate.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include "omr.h"
+#include "omrExampleVM.hpp"
+#include "omrvm.h"
+#include "OMRVMInterface.hpp"
+#include "ParallelGlobalGC.hpp"
+#include "Scavenger.hpp"
+#include "SlotObject.hpp"
+
+void
+MM_ScavengerDelegate::masterSetupForGC(MM_EnvironmentBase *env)
```
rwy0717: The implementation of this file was taken from the example's CollectorLanguageInterfaceImpl, with the following changes:

* wrapped body in ifdef(OMR_GC_SCAVENGER_DELEGATE)
* renamed CLI to MM_ScavengerDelegate
* deleted virtual keyword

example/glue/ScavengerDelegate.hpp
```diff
@@ -0,0 +1,260 @@
...
+class MM_CompactScheme;
+class MM_EnvironmentStandard;
+class MM_ForwardedHeader;
+class MM_MarkMap;
+class MM_MemorySubSpaceSemiSpace;
+
+class MM_ScavengerDelegate : public MM_BaseVirtual {
+
+private:
+	MM_CollectorLanguageInterface* _cli;
```
rwy0717: The example delegate is holding a pointer to the cli, which isn't actually used. The pointer is initialized in the constructor. I want to assess if openj9's scavenger will need the CLI before I remove.

example/glue/ScavengerDelegate.hpp
```diff
@@ -0,0 +1,260 @@
...
+#endif /* defined(OMR_ENV_DATA64) && !defined(OMR_GC_COMPRESSED_POINTERS) */
+
+	/**
+	 * This method will be called on the master GC thread after each scavenger cycle, successful or
+	 * otherwise. It may be used for reporting or other tasks as required.
+	 *
+	 * @param[in] env The environment for the calling thread.
+	 * @param[in] scavengeSuccessful Indicates whether the scavenge cycle completed normally or aborted.
+	 */
+	void reportScavengeEnd(MM_EnvironmentBase * env, bool scavengeSuccessful);
```
rwy0717: function names were taken from the CLI, with the `scavenger_` prefix removed.

glue/ScavengerDelegate.cpp
```diff
@@ -0,0 +1,170 @@
+/*******************************************************************************
```
rwy0717: The implementation of this file was taken from the empty glue's CLI, with functions and classes renamed.

gc/base/standard/Scavenger.cpp
```diff
@@ -382,7 +382,11 @@ MM_Scavenger::masterSetupForGC(MM_EnvironmentStandard *env)
 	clearCycleGCStats(env);
```
rwy0717: In this file, every usage of the CLI is replaced by the corresponding call to the delegate. The code should be functionally equivalent.

example/glue/ScavengerDelegate.hpp
rwy0717: Removed the CLI pointer.

rwy0717:
Added an initialize and tearDown member-functions to the delegate, called during scavenger init/teardown. There are no newInstance/kill functions, because the delegate is embedded in the scavenger, not a separate allocation.

youngar:
@genie-omr build all

youngar:
Nothing here seems too complicated, just a refactoring.  Hopefully we get this in soon and can update the OpenJ9 side of things.

youngar:
@charliegracie Any comments, or can I merge this?

rwy0717:
Regenerated the configure script

charliegracie:
@genie-omr build all

example/glue/ScavengerDelegate.cpp
ktbriggs: Can we also remove the remaining `*_` prefixes from the function names? They too seem superfluous at this point...

PR #3499, opened 2019-01-23
Use the Test Compiler with Tril tests instead of JitBuilder

Leonardo2718:
The dependency on JitBuilder is broken by only providing the Test
Compiler as a target link library. A few workarounds for JitBuilder's
new client API are also removed. Lastly, the function to initialize the
Test Compiler with specific options is also exposed.

### DIFF
--- a/fvtest/compilertest/Jit.hpp
+++ b/fvtest/compilertest/Jit.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2017 IBM Corp. and others
+ * Copyright (c) 2017, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -26,5 +26,6 @@ namespace TR { class MethodBuilder; }
 class TR_Memory;
 
 extern "C" bool initializeJit();
+extern "C" bool initializeJitWithOptions(char * options);
 extern "C" uint32_t compileMethodBuilder(TR::MethodBuilder *m, uint8_t **entry);
 extern "C" void shutdownJit();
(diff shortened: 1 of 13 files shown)


Leonardo2718:
@genie-omr build all

charliegracie:
Hmmm I thought originally tril was using the `testcompiler` but then it was moved to use `jitbuilder` as we were deprecating the `testcompiler`.  I could very well be confusing this with something different.

charliegracie:
I am also not sure the `testcompiler` is properly compiled on all of the platforms where `jitbuilder` currently compiles.

Leonardo2718:
> Hmmm I thought originally tril was using the `testcompiler` but then it was moved to use `jitbuilder` as we were deprecating the `testcompiler`.

Nope. Tril has always used `jitbuilder` as the default backend, though there was an option added a while ago to use the `testcompiler` as an alternative.

We're deprecated the `testcompiler` tests (IlInjector-based tests) in favour of Tril-based tests. However, the "compiler" part of the `testcompiler` is going to stay because it can still be useful for Tril. (I should really publicize what my plans are to avoid this kind of confusion :stuck_out_tongue: )

> I am also not sure the `testcompiler` is properly compiled on all of the platforms where `jitbuilder` currently compiles.

I'd be surprised if platform support was very different because there's a lot of duplicate code between them.

I believe the test failures we're seeing are due to a few things I just forgot to remove. :sweat_smile:

Leonardo2718:
@genie-omr build plinux, zlinux

Leonardo2718:
@genie-omr build plinux

Leonardo2718:
@genie-omr build plinux

Leonardo2718:
@genie-omr build zlinux

Leonardo2718:
@genie-omr build zlinux

Leonardo2718:
@genie-omr build zlinux, plinux

Leonardo2718:
@genie-omr build zlinux, plinux

Leonardo2718:
@genie-omr build all

0xdaryl:
Code changes look fine.  Please double check the copyright headers for 2019 on some of the files.

Leonardo2718:
Fixed the few copyright headers I missed.

@genie-omr build all

Leonardo2718:
(AppVeyor build failed because of infrastructural problems. Restarted)

PR #3500, opened 2019-01-23
Implemented frem tril tests

hsoontie:
Implemented frem tril tests

### DIFF
--- a/fvtest/compilertriltest/ArithmeticTest.cpp
+++ b/fvtest/compilertriltest/ArithmeticTest.cpp
@@ -24,6 +24,7 @@
 #include "omrformatconsts.h"
 
 #include <cmath>
+#include <limits>
 
 #if defined(J9ZOS390)
 namespace std
@@ -432,10 +433,60 @@ float fdiv(float l, float r) {
     return l/r;
 }
 
+bool isnan(float x) {
+    return x != x; // NANs never compare equal
+}
+
+bool isinf(float x) {
+    return std::abs(x) == std::numeric_limits<float>::infinity();
+}
+
+float frem_copysign(float x, float y) {
+    // technically not completely correct because -0.0 == +0.0
+    // also doesn't handle NaN's correctly because NaN == NaN is always false
+    return (std::abs(y) == y ? +1.0f : -1.0f)*x;
+}
+
+float frem_round(float x) {
+    float n = std::ceil(x);
+    if ((n - x > 0.5f) || (n - x == 0.5f && std::fmod(n, 2.0f) != 0.0f)) {
+        n = std::floor(x);
+    }
+    return n;
+}
+
+float frem(float x, float y) {
+    if (isnan(x) || isnan(y)) {
+        return std::numeric_limits<float>::quiet_NaN();
+    }
+    else if (std::abs(y) == 0.0f || isinf(x)) {
+        return std::numeric_limits<float>::signaling_NaN();
+    }
+    else if (isinf(y)) {
+        return x;
+    }
+
+    float rem = x - frem_round(x / y) * y;
+
+    if (std::abs(rem) == 0.0f) {
+        rem = frem_copysign(rem, x);
+    }
+
+    return rem;
+}
+
 class FloatArithmetic : public TRTest::BinaryOpTest<float> {};
 
 TEST_P(FloatArithmetic, UsingConst) {
     auto param = TRTest::to_struct(GetParam());
+
+    std::string arch = omrsysinfo_get_CPU_architecture();
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_HAMMER == arch), MissingImplementation)
+        << "The x86-64 codegen does not yet support frem.";
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_PPC64LE == arch), MissingImplementation)
+        << "The PPC-64_LE codegen does not yet support frem.";
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_S390X == arch), MissingImplementation)
+        << "The S390X codegen does not yet support frem.";
 
     char inputTrees[1024] = {0};
     std::snprintf(inputTrees, sizeof(inputTrees),
@@ -469,6 +520,14 @@ TEST_P(FloatArithmetic, UsingConst) {
 
 TEST_P(FloatArithmetic, UsingLoadParam) {
     auto param = TRTest::to_struct(GetParam());
+
+    std::string arch = omrsysinfo_get_CPU_architecture();
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_HAMMER == arch), MissingImplementation)
+        << "The x86-64 codegen does not yet support frem.";
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_PPC64LE == arch), MissingImplementation)
+        << "the PPC-64_LE codegen does not yet support frem.";
+    SKIP_IF(param.opcode == "frem" && (OMRPORT_ARCH_S390X == arch), MissingImplementation)
+        << "The S390X codegen does not yet support frem.";
 
     char inputTrees[1024] = {0};
     std::snprintf(inputTrees, sizeof(inputTrees),
@@ -505,7 +564,8 @@ INSTANTIATE_TEST_CASE_P(ArithmeticTest, FloatArithmetic, ::testing::Combine(
         std::make_tuple<const char*, float (*)(float, float)>("fadd", static_cast<float (*)(float, float)>(fadd)),
         std::make_tuple<const char*, float (*)(float, float)>("fsub", static_cast<float (*)(float, float)>(fsub)),
         std::make_tuple<const char*, float (*)(float, float)>("fmul", static_cast<float (*)(float, float)>(fmul)),
-        std::make_tuple<const char*, float (*)(float, float)>("fdiv", static_cast<float (*)(float, float)>(fdiv))
+        std::make_tuple<const char*, float (*)(float, float)>("fdiv", static_cast<float (*)(float, float)>(fdiv)),
+        std::make_tuple<const char*, float (*)(float, float)>("frem", static_cast<float (*)(float, float)>(frem))
     )));
 
 double dadd(double l, double r) {


fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -221,6 +221,10 @@ float fdiv(float l, float r) {
     return l/r;
 }
 
+float frem(float l, float r) {
+    return remainder(l, r);
```
Leonardo2718: I believe `std::remainder()` is new in C++11. Are we sure it's available on all our supported build compilers?

hsoontie: Thanks for the heads-up! I'll look for a different solution.

hsoontie: @Leonardo2718 I just replaced it with remainderf(), which I believe is the C99 version.

Leonardo2718: I'm not 100% sure yet that's going to be enough.

The C99 standard library is, technically, only "officially" available in C++ since C++11.

However, because most C++03 compilers ship with a C99 compiler as well, I suspect the C99 standard library *is* available pre-C++11 in most compilers.

Leonardo2718:
@genie-omr build all

fvtest/compilertriltest/ArithmeticTest.cpp
hsoontie: @Leonardo2718 Here is another solution done using floor() and ceil(), does this work?

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -221,6 +221,31 @@ float fdiv(float l, float r) {
...
+        }
+        else if(ans == anstrunc + 0.5)
+        {
+            (int)anstrunc % 2 == 0 ? ans = anstrunc : ans = anstrunc + 1;
+        }
+        else
+        {
+            ans = anstrunc; 
+        }
+    }
```
Leonardo2718: I don't think this algorithm works. Testing it with `frem(-5.1, 3.0)` I get -2.1 instead of 0.9.

I'm still trying to investigate whether `remainderf`/`remainder` is supported on all our build platforms, so that option is not off the table yet. Still working on it though.

hsoontie: Oops! Thanks for that, I think I found the issue. 
I may put this on hold until you're sure about remainderf(), since using that would be ideal, if we can. Could you send a message or make another comment when you do find out?

Leonardo2718: Sorry @hsoontie. Still looking into it.

Leonardo2718: After some investigation, I've concluded that `std::remainderf()` should work on all our supported compilers. Sorry for taking so long to reply.

hsoontie: Awesome, thank you! I will revert it soon.

hsoontie: @Leonardo2718 Done!

aarongraham9:
It seems something went wrong with the Travis CI build (seems the build actually passed) a while back. Would you be able to restart the Travis CI build @0xdaryl or @Leonardo2718.

Other than that this PR should be good for a next review/merge.

Leonardo2718:
Hmm... for some reason, the Travis build completed successfully but the status isn't being updated.

@hsoontie Sorry for the delay and for flip-flopping back and forth on this PR (figuring out what is and isn't supported by our build compilers is many times more difficult than it ought to be). As it turns out, the version of MSVC we use in our builds doesn't have `remainderf`.

As such, please update the test with the following `frem()` implementation:

```c++
#include <cmath>
#include <limits>

bool isnan(float x) {
    return x != x; // NANs never compare equal
}

bool isinf(float x) {
    return std::abs(x) == std::numeric_limits<float>::infinity();
}

float copysign(float x, float y) {
    // technically not completely correct because -0.0 == +0.0
    // also doesn't handle NaN's correctly because NaN == NaN is always false
    return (std::abs(y) == y ? +1.0f : -1.0f)*x;
}

float frem_round(float x) {
    float n = std::ceil(x);
    if ((n - x > 0.5f) || (n - x == 0.5f && (int(n) & 0x1))) {
        n = std::floor(x);
    }
    return n;
}

float frem(float x, float y) {
    if (isnan(x) || isnan(y)) {
        return std::numeric_limits<float>::quiet_NaN();
    }
    else if (std::abs(y) == 0.0f || isinf(x)) {
        return std::numeric_limits<float>::signaling_NaN();
    }
    else if (isinf(y)) {
        return x;
    }

    float rem = x - frem_round(x / y) * y;

    if (std::abs(rem) == 0.0f) {
        rem = copysign(rem, x);
    }

    return rem;
}

```

hsoontie:
@Leonardo2718 , I added that code in. Please let me know if there's anything else I can do!

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -221,8 +222,46 @@ float fdiv(float l, float r) {
     return l/r;
 }
 
-float frem(float l, float r) {
-    return remainderf(l, r);
+bool isnan(float x) {
```
fjeremic: Worried about this one on z/OS as well because `isnan` is defined in `math.h`

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -221,8 +222,46 @@ float fdiv(float l, float r) {
...
+    }
+    return n;
+}
+
+float frem(float x, float y) {
+    if (isnan(x) || isnan(y)) {
+        return std::numeric_limits<float>::quiet_NaN();
+    }
+    else if (std::abs(y) == 0.0f || isinf(x)) {
+        return std::numeric_limits<float>::signaling_NaN();
```
fjeremic: Not 100% sure this will compile on z/OS.

fjeremic:
@Leonardo2718 could we hold off on the merge? I'm going to test this change against #3693 to make sure it doesn't break things as I don't think #3693 will be merged before this.

fvtest/compilertriltest/ArithmeticTest.cpp
fjeremic: I checked this one and it is fine.

fvtest/compilertriltest/ArithmeticTest.cpp
fjeremic: I checked this one and it is fine.

fvtest/compilertriltest/ArithmeticTest.cpp
```diff
@@ -221,6 +222,48 @@ float fdiv(float l, float r) {
...
 
+bool isnan(float x) {
+    return x != x; // NANs never compare equal
+}
+
+bool isinf(float x) {
+    return std::abs(x) == std::numeric_limits<float>::infinity();
+}
+
+float copysign(float x, float y) {
```
fjeremic: This seems to be the only problem on z/OS. `copysign` is already defined in `math.h` so we cannot redefine here. Either use the standard library function or rename this.

Leonardo2718:
@fjeremic 

> could we hold off on the merge? I'm going to test this change against #3693 to make sure it doesn't break things as I don't think #3693 will be merged before this.

Sure, no worries. I think spending the extra time to make sure we don't break #3693 is worthwhile.

And thanks for testing this out, btw!

hsoontie:
@Leonardo2718 @fjeremic Function name has been changed. Thank you!

Leonardo2718:
@genie-omr build all

Leonardo2718:
Hm, some of the tests seem to be failing.

hsoontie:
Looking at the logs, I'm not really sure where it's going wrong. To me, the code you suggested looks like it should work. Would anybody have a suggestion?

Leonardo2718:
It looks like it's failing on linux x86_64. As a start, I'd suggest running the test locally with a debugger.

hsoontie:
@0xdaryl has answered some questions I had about this PR's x86 failure over in #3475 .
For those who don't know yet, here is a short summary of what I know about the issue, posted on slack:

_When running tril tests, a function (getLinkage() in \compiler\runtime\Runtime.hpp) is indexing into the linkage array using the runtime helper TR_AMD64floatRemainder (index #182). linkage[182] returns 0, which corresponds to the linkage type TR_None. This ultimately leads to a segmentation fault later after looking for the linkage type._

The main part of Daryl's response is as follows:

_There isn't a runtime version of float remainder on x86 in OMR (JitBuilder or CompilerTest), which is the function you're looking for and probably why you're getting TR_None back for the linkage. That's one of the gaps we have with floating point unfortunately that we need to fix. OpenJ9 provides a function, but it's written in assembler. Such a function can't be directly used in OMR because there isn't a common assembler on x86 available as part of the toolchain (and I'm not convinced there should be)._

@0xdaryl , is this problem something I can fix right now?

aarongraham9:
Any thoughts on this one @Leonardo2718, @fjeremic or @0xdaryl?

Leonardo2718:
> There isn't a runtime version of float remainder on x86 in OMR (JitBuilder or CompilerTest), which is the function you're looking for and probably why you're getting TR_None back for the linkage. That's one of the gaps we have with floating point unfortunately that we need to fix. OpenJ9 provides a function, but it's written in assembler. Such a function can't be directly used in OMR because there isn't a common assembler on x86 available as part of the toolchain (and I'm not convinced there should be).

It sounds like an issue needs to be opened for this if there isn't one already.

Until a solution is found for the issue, the frem tests should be disabled using the `SKIP_IF` macro (see other tests in `ArithmeticTest.cpp` for examples).

hsoontie:
@Leonardo2718 Rebased and disabled the tests for x86_64. I investigated the PPC-64_LE failure as well and it looked to be having the same issue, so I ended up disabling those tests as well. I am still not sure about the other two.

Leonardo2718:
Thanks @hsoontie. Would you mind opening an issue about adding frem support to PPC64LE and x86-64?

@genie-omr build all

hsoontie:
It looks like there is one more segfault left. I will investigate this ASAP.

hsoontie:
So, after looking at the Jenkins build I am thinking that the failure here is not because of the same issue Daryl mentioned. The tril test actually looks like it's running all the way through, but the expected or actual values are coming out wrong (I am not sure why). I think I remember this happening on the failed osx x86_64 build as well, which I can't see anymore because disabling the linux x86_64 build also disabled the osx one. 
@aarongraham9 suggested that I create an issue for the 4 failures and disable the remaining test as I did with the others, because the problems seem to us like they're unrelated to the tril tests. If I do take this route, I would also like to be able to see the old failed Jenkins osx build if possible to make sure the issue I create is accurate.
Do you have any thoughts on that idea, @Leonardo2718  or @0xdaryl ?

hsoontie:
@Leonardo2718 @0xdaryl I opened issue #4135 just now. Please let me know when you have a minute if it should be changed/if you would like me to take another course of action.

aarongraham9:
Thoughts on issue #4135 for the failing `ppc*`, `x86*`, `s390x` and OS support @Leonardo2718 and @0xdaryl?

Also thoughts on getting this merged for `AArch64` @Leonardo2718 and @0xdaryl?

Leonardo2718:
Before this can be merged, the failure on z Linux either needs to be resolved or the tests need to be disabled on that platform, referencing #4135 as the reason for disabling.

aarongraham9:
Apologies @Leonardo2718 & @0xdaryl; I was under the impression the `Z` testhad been disabled at this point.

@hsoontie; can you update this PR with disabling the `Z` platform test, "referencing #4135 as the reason for disabling", and re-push to the PR? A rebase probably wouldn't be a bad idea either. :)

hsoontie:
Unfortunately I could not test this directly due to a build issue, but the change I just made should disable the s390x tests as well. I will be watching the build status to confirm.
@Leonardo2718, I have now updated my commit message to reference the issue. Please let me know if there's anything else I should do.

Leonardo2718:
It's customary to also reference the issue number the "skip message" so that someone who sees the message can know where to go look for more information.

@genie-omr build zlinux

hsoontie:
@Leonardo2718 It properly skipped the tests, but AppVeyor failed. It seems like "winflexbison" failed to install.

Leonardo2718:
I've restarted the AppVeyor build.

hsoontie:
@Leonardo2718 @0xdaryl All tests have passed now. The issue number is in the commit message.

hsoontie:
@Leonardo2718 @0xdaryl Hello, does anything else need to be done here in relation to #4135?

aarongraham9:
I believe this is good to go as-is and the remaining items can be completed against #4135. Thoughts @Leonardo2718 and @0xdaryl?

0xdaryl:
@Leonardo2718:  if @hsoontie added architecture exclusion guards for `OMRPORT_ARCH_AARCH64` and `OMRPORT_ARCH_ARM`  like she did for the other OMR backends could we land this code?  We'll revisit those guards when #3475 is merged.

Leonardo2718:
As Filip explained in https://github.com/eclipse/omr/issues/4135#issuecomment-516981240, the oracle function appears to be incorrect. I don't want to merge this until we have a proper solution for the oracle as it's causing false positives on several platforms.

0xdaryl:
I think the best course of action is to close this PR while the larger problem of determining the expected results is worked out.  However, I want to open an issue for the larger problem of defining and validating floating point semantics in OMR and reference it in this PR before closing.

aarongraham9:
Any word on the Issue(s) to link this agains @0xdaryl?

aarongraham9:
Any word on the Issue(s) to link this against @0xdaryl?

PR #3502, opened 2019-01-23
Eliminate FrontEnd calls to inquire about CodeCache bounds

0xdaryl:
Eliminate calls to the FrontEnd functions `getCodeCacheBase()` and `getCodeCacheTop()`
in all their forms and instead ask the CodeCache directly for its bounds.

The Z code generator was the only consumer of these APIs.  In all cases it contained a
contingency for the situation where a CodeCache had not been allocated yet, which can
occur if the Z code generator function was called during optimization.  There were no
situations in which this occurred so the code was simplified to only handle the case
where there is an allocated CodeCache.

Note that for OMR projects that did not specialize the FrontEnd functions, the bounds
of the CodeCache will now be more precise.  The default FrontEnd implementations in
OMR simply return `0` for the CodeCache base and `(uint8_t*)(~0)` for the CodeCache
top.  The changes in this commit will return the actual base and top addresses of
the current CodeCache.  This is arguably the correct behaviour.

The vestigal FrontEnd functions will be cleaned up in a subsequent PR.

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build zlinux,zos

This is a Z-specific change.  Full builds are not required.

0xdaryl:
@genie-omr build zlinux,zos

fjeremic:
Modified files are in the Z codegen which AppVeyor does not run. Merging.

PR #3504, opened 2019-01-23
Enhance JIT version generation script

keithc-ca:
When invoked with an optional second argument, it will write to the named (output) file only if the contents need to change. This avoids recompiling the generated source file and downstream (linking, etc.) when no JIT changes have been made.

### DIFF
--- a/jitbuilder/build/rules/gnu/common.mk
+++ b/jitbuilder/build/rules/gnu/common.mk
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2016, 2016 IBM Corp. and others
+# Copyright (c) 2016, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
@@ -63,7 +63,7 @@ $(call RULE.cpp,$(JIT_PRODUCT_BUILDNAME_OBJ),$(JIT_PRODUCT_BUILDNAME_SRC))
 .phony: $(JIT_PRODUCT_BUILDNAME_SRC)
 $(JIT_PRODUCT_BUILDNAME_SRC):
 	@mkdir -p $(dir $@)
-	$(PERL_PATH) $(GENERATE_VERSION_SCRIPT) $(PRODUCT_RELEASE) > $@
+	$(PERL_PATH) $(GENERATE_VERSION_SCRIPT) $(PRODUCT_RELEASE) $@
 
 jit_clean::
 	rm -f $(JIT_PRODUCT_BUILDNAME_SRC)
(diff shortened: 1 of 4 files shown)


jitbuilder/build/rules/gnu/common.mk
```diff
@@ -63,7 +63,7 @@ $(call RULE.cpp,$(JIT_PRODUCT_BUILDNAME_OBJ),$(JIT_PRODUCT_BUILDNAME_SRC))
 .phony: $(JIT_PRODUCT_BUILDNAME_SRC)
```
youngar: I'm guessing that it was specifically this phony rule which generates the version unconditionally that was the problem?  I think that `add_custom_command()` should have been working fine, is that correct?

youngar:
@genie-omr build all

jitbuilder/build/rules/gnu/common.mk
keithc-ca: I think that should be `.PHONY`, not `.phony`.

The main problem was that (when the rule ran), it would modify the target unconditionally. The new version script is given a chance to read the target file so it can be left untouched when the contents are correct.

Assuming the cmake custom command ran at the right times (I'm not sure that's the case, but I'd prefer to leave that to a separate PR), it would also unconditionally modify the file.

fjeremic:
I also think this is the only `perl` dependency in OMR? Wondering if we should eventually phase this out in place of something else.

youngar:
Another option would be to pick up the `snapshot_name` during "configure" time.  It wouldn't update the `snapshot_name` down to the minute every time you run `make`, but is that a problem?

keithc-ca:
I'd argue that doing it at configuration time would be too infrequent. One use case is via setting the environment variable `TR_BUILD_NAME` (openj9 uses this strategy) which might be different from the previous build.

I'd be (a little) surprised if this is the only use of `perl` in OMR, but I agree that removing it as a dependency is a worthy goal. Perhaps it would be easiest to switch to python which is required by jitbuilder, or replace uses of python with perl?

fjeremic:
> I'd argue that doing it at configuration time would be too infrequent. One use case is via setting the environment variable `TR_BUILD_NAME` (openj9 uses this strategy) which might be different from the previous build.

For historical reference the reason this whole thing was added because we tried to get ccache to play nice with the JIT. In the before world we supplied the name via a `-D` directly to make, however this was always invalidating ccache entries as the commandline is used as part of the hash and the commandline would always change (up to a minute granularity).

youngar:
@fjeremic Thanks for the note!  I can see how this helps.

@keithc-ca  I had forgotten that re-running UMA did not re-configure OMR.  I can't wait until both are using CMake!

I think a `sh` script would probably be the easiest for this script, given how simple it is. Either way, I think this change is good to go as-is.

keithc-ca:
It's unclear to me whether any changes are requested; what, if anything, do I need to do before this can be merged?

youngar:
No changes are requested by me.

PR #3510, opened 2019-01-24
Introduce empty CPU extension classes for ARM and AArch64

0xdaryl:
To prevent downstream project breakages, introduce empty extensions
to the CPU hierarchy for ARM and AArch64.  Once merged, downstream
projects can make changes to their build systems to accommodate the new
files before new CPU functionality is added in #3479.

### DIFF
--- a/compiler/aarch64/env/OMRCPU.hpp
+++ b/compiler/aarch64/env/OMRCPU.hpp
@@ -1,0 +1,57 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2019 IBM Corp. and others
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at http://eclipse.org/legal/epl-2.0
+ * or the Apache License, Version 2.0 which accompanies this distribution
+ * and is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License, v. 2.0 are satisfied: GNU General Public License,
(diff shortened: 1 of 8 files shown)


0xdaryl:
@genie-omr build all

compiler/aarch64/env/OMRCPU.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OMR
+{
+
+namespace ARM64
+{
+
+class CPU : public OMR::CPU
+   {
+protected:
```
Leonardo2718: Indentation seems off:
```suggestion
   protected:
```

compiler/arm/env/OMRCPU.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OMR
+{
+
+namespace ARM
+{
+
+class CPU : public OMR::CPU
+   {
+protected:
```
Leonardo2718: Indentation seems off again:

```suggestion
   protected:
```

compiler/aarch64/env/OMRCPU.hpp
0xdaryl: The compiler code is wildly inconsistent on this, to the point where it's a localized matter of preference.  I used this style because that's what's used in the parent class `OMR::CPU`.

compiler/arm/env/OMRCPU.hpp
0xdaryl: Ditto, as above.

compiler/aarch64/env/OMRCPU.hpp
Leonardo2718: Hm... Ok.

Leonardo2718:
Change seems straightforward enough. Merging.

PR #3513, opened 2019-01-25
Acquire current CodeCache directly from CodeGenerator

0xdaryl:
Replace calls to Compilation get/setCurrentCodeCache with a direct
inquiry to the CodeGenerator.  The Compilation functions always delegate
to the CodeGenerator.  Delete the Compilation functions.

There is one instance in `CompileMethodFromDetails` where the CodeGenerator
object could be NULL (during an AOT load) and that scenario is handled
by passing a NULL CodeCache as the original Compilation function would have.

### DIFF
--- a/compiler/env/FEBase_t.hpp
+++ b/compiler/env/FEBase_t.hpp
@@ -50,19 +50,20 @@ uint8_t *
 FEBase<Derived>::allocateCodeMemory(TR::Compilation *comp, uint32_t warmCodeSize, uint32_t coldCodeSize,
                             uint8_t **coldCode, bool isMethodHeaderNeeded)
    {
-   TR::CodeCache *codeCache = static_cast<TR::CodeCache *>(comp->getCurrentCodeCache());
+   TR::CodeGenerator *cg = comp->cg();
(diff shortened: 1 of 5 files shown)


compiler/env/FEBase_t.hpp
```diff
@@ -50,19 +50,20 @@ uint8_t *
 FEBase<Derived>::allocateCodeMemory(TR::Compilation *comp, uint32_t warmCodeSize, uint32_t coldCodeSize,
                             uint8_t **coldCode, bool isMethodHeaderNeeded)
    {
-   TR::CodeCache *codeCache = static_cast<TR::CodeCache *>(comp->getCurrentCodeCache());
+   TR::CodeGenerator *cg = comp->cg();
+   TR::CodeCache *codeCache = static_cast<TR::CodeCache *>(cg->getCodeCache());
```
Leonardo2718: I don't think this `static_cast` is needed because we already get a `TR::CodeCache *`:

```suggestion
   TR::CodeCache *codeCache = cg->getCodeCache();
```

Leonardo2718:
@genie-omr build all

Issue #3514, opened 2019-01-25
Initial tenure age


amicic:
Current initial tenure age is 10.

Persistent objects which are typically created at the startup, may get flipped many times, before GC adjusts and reduce the tenure threshold. With larger heaps, it takes long time (in absolute terms) to adjust, Until then GCs will be more expensive than necessary (due to unnecessary flipping of mostly persistent objects). Consequently it may slow down the startup process. Additionally, Nursery may expand more than necessary (since expand heuristic relies on Scavenge overhead).

amicic:
Experiments that demonstrate noticeable footprint reduction and some startup improvements with larger initial heaps and initial tenure age 1, using recent OpenJ9 VM and various Liberty based applications.

![sufp_tenureage1](https://user-images.githubusercontent.com/18264984/51759193-ad3a3600-2095-11e9-8c88-616da4fd2517.png)

What is not specified here is initial heap size. It is actually not explicitly set (with -Xms or similar option), but it's learnt from previous runs (as part of https://github.com/eclipse/openj9/issues/3743). Initial sizes end up being in neighborhood of 40M for Nursery (wheres default initial Nursery size is only 2M).

amicic:
This is from another experiment, with default initial heap size, and lower initial tenure age of 6. Lower tenure age does not seem to either help or hurt. Since Scavenges are very frequent with small initial Nursery, the age threshold adapts very quickly (in absolute time terms).

![sufp_defaultinitheap_tenureage6](https://user-images.githubusercontent.com/18264984/51759648-b7106900-2096-11e9-861e-57799703adfe.png)

PR #3515, opened 2019-01-25
Initial Tenure Age

amicic:
Initial Tenure Age will be a function of initial Nursery size. With
larger initial Nursery heaps, initial Tenure Age will be lower. This
helps promoting persistent objects (typically created at startup)
sooner, which reduces startup time and Nursery growth. See:
https://github.com/eclipse/omr/issues/3514

For Nursery Sizes of
256K, 512K, 1M, 2M, 4M, 8M, 16M, 32M, >=64M

initial Tenure Age is set respectively to:

9, 8, 7, 6, 5, 4, 3, 2, 1.

For OpenJ9, which starts with 2M Nursery, this will reduce the default
initial Tenure Age to 6 (from 10).


charliegracie:
The rate at which the initial tenure age is decreased seems very aggressive to me.  Most normal sized heap workloads (min/max heap set to 2G) would have a nursery of 512M.  This change would force an initial tenure age of 1 on all of those workloads.  

Since users can override the initial tenure age and still allow dynamic tenure age adjustment there is an easy "fix" if this causes issues.

amicic:
Initial age 1 was intentionally set (even for not so large inital heaps). Measurements here (https://github.com/eclipse/omr/issues/3514#issuecomment-457635715) for initial heap size of about 40M Nursery clearly show that age 1 helped. Similar experiment was done with age 3 (not shown in the issue though), but it did not help as much.

I've also done some informal measurements of GC times for a known Java benchmark with larger initial heaps (order of GB Nursery) and it demonstrated a drop in pauses for the first few GC cycle (except first one which is same, since all objects are flipped at least once) till age converges to a somewhat stable value. 

Possibly age 1 is optimal across all initial sizes (even as low as 1M Nursery or so), but due to lack of data for very small initial Nursery sizes (order of 1M),  I just left it somewhere in between what we used to have and aggressive setting of 1. 

Btw, I'm unsure what you meant by 'this' in you last statement. Did you mean issue with this change or issue with high current default age value of 10 (ie are you endorsing the change or not)?

charliegracie:
the `this` was this change.  I was sort of endorsing the change :)

With your further explanation I am ok with the change.

charliegracie:
@genie-omr build all

PR #3517, opened 2019-01-28
Implemented binary evaluator helper for AArch64

hsoontie:
WIP Implementing blanket binary evaluator helper for AArch64

### DIFF
--- a/compiler/aarch64/codegen/BinaryEvaluator.cpp
+++ b/compiler/aarch64/codegen/BinaryEvaluator.cpp
@@ -30,6 +30,80 @@
 #include "il/Node_inlines.hpp"
 #include "infra/Bit.hpp"
 
+/**
+ * Generic helper that handles a number of similar binary operations.
+ * @param[in] node : calling node
+ * @param[in] regOp : the target AArch64 instruction opcode
+ * @param[in] regOpImm : the matching AArch64 immediate instruction opcode 
+ * regOpImm == regOp indicates that the passed opcode has no immediate form.
+ * @param[in] cg : codegenerator
+ * @return target register
+ */
+static inline TR::Register *
+genericBinaryEvaluator(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::InstOpCode::Mnemonic regOpImm, bool is64Bit, TR::CodeGenerator *cg)
+   {
+   TR::Node *firstChild = node->getFirstChild();
+   TR::Node *secondChild = node->getSecondChild();
+   TR::Register *src1Reg = cg->evaluate(firstChild);
+   TR::Register *src2Reg = NULL;
+   TR::Register *trgReg = NULL;
+   int64_t value = 0;
+
+   if(1 == firstChild->getReferenceCount())
+      {
+      trgReg = src1Reg;
+      }
+   else if(1 == secondChild->getReferenceCount() && secondChild->getRegister() != NULL)
+      {
+      trgReg = src2Reg;
+      }
+   else
+      {
+      trgReg = cg->allocateRegister();
+      }
+
+   if (secondChild->getOpCode().isLoadConst() && secondChild->getRegister() == NULL)
+      {
+      if(is64Bit)
+         {
+         value = secondChild->getLongInt();
+         }
+      else
+         {
+         value = secondChild->getInt();
+         }
+      /* When regOp == regOpImm, an immediate version of the instruction does not exist. */
+      if(constantIsUnsignedImm12(value) && regOp != regOpImm)
+         {
+         generateTrg1Src1ImmInstruction(cg, regOpImm, node, trgReg, src1Reg, value);
+         }
+      else
+         {
+         src2Reg = cg->allocateRegister();
+         if(is64Bit)
+            {
+            loadConstant64(cg, node, value, src2Reg);
+            }
+         else
+            {
+            loadConstant32(cg, node, value, src2Reg);
+            }
+         generateTrg1Src2Instruction(cg, regOp, node, trgReg, src1Reg, src2Reg);
+         cg->stopUsingRegister(src2Reg);
+         }
+      }
+   else
+      {
+      src2Reg = cg->evaluate(secondChild);
+      generateTrg1Src2Instruction(cg, regOp, node, trgReg, src1Reg, src2Reg);
+      }
+   
+   firstChild->decReferenceCount();
+   secondChild->decReferenceCount();
+   node->setRegister(trgReg);
+   return trgReg;
+   }
+
 static TR::Register *addOrSubInteger(TR::Node *node, TR::CodeGenerator *cg)
    {
    TR::Node *firstChild = node->getFirstChild();
@@ -68,27 +142,25 @@ static TR::Register *addOrSubInteger(TR::Node *node, TR::CodeGenerator *cg)
 TR::Register *
 OMR::ARM64::TreeEvaluator::iaddEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   return addOrSubInteger(node, cg);
+   return genericBinaryEvaluator(node, TR::InstOpCode::addw, TR::InstOpCode::addimmw, false, cg);
    }
 
 TR::Register *
 OMR::ARM64::TreeEvaluator::laddEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::laddEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
-	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
+	return genericBinaryEvaluator(node, TR::InstOpCode::addx, TR::InstOpCode::addimmx, true, cg);
 	}
 
 TR::Register *
 OMR::ARM64::TreeEvaluator::isubEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   return addOrSubInteger(node, cg);
+   return genericBinaryEvaluator(node, TR::InstOpCode::subw, TR::InstOpCode::subimmw, false, cg);
    }
 
 TR::Register *
 OMR::ARM64::TreeEvaluator::lsubEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 	{
-	// TODO:ARM64: Enable TR::TreeEvaluator::lsubEvaluator in compiler/aarch64/codegen/TreeEvaluatorTable.hpp when Implemented.
-	return OMR::ARM64::TreeEvaluator::unImpOpEvaluator(node, cg);
+	return genericBinaryEvaluator(node, TR::InstOpCode::subx, TR::InstOpCode::subimmx, true, cg);
 	}
 
 // Multiply a register by a 32-bit constant
@@ -279,7 +351,7 @@ OMR::ARM64::TreeEvaluator::lmulhEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 TR::Register *
 OMR::ARM64::TreeEvaluator::idivEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   return idivHelper(node, false, cg);
+   return genericBinaryEvaluator(node, TR::InstOpCode::sdivw, TR::InstOpCode::sdivw, false, cg);
    }
 
 TR::Register *
@@ -291,7 +363,7 @@ OMR::ARM64::TreeEvaluator::iremEvaluator(TR::Node *node, TR::CodeGenerator *cg)
 TR::Register *
 OMR::ARM64::TreeEvaluator::ldivEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
-   return idivHelper(node, true, cg);
+   return genericBinaryEvaluator(node, TR::InstOpCode::sdivx, TR::InstOpCode::sdivx, true, cg);
    }
 
 TR::Register *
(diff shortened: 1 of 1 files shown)


hsoontie:
This PR contains work based on our proposal to simplify BinaryEvaluator.cpp code.
(Myself, @jeanlego & @marufunb all contributed.)

The helper functions in this PR have been tested to cover the following 2-source evaluators:
- iadd
- ladd
- isub
- lsub
- idiv
- ldiv
- iand
- land
- ior
- lor
- ixor
- lxor

Notes:
- Other binary evaluators (such as irem and lrem) contain extra steps and don't match this format quite as closely, and we are thinking those evaluators may be best left seperate.
- As can be seen, we are looking into expanding its usability to cover single and 3-source operations, and were wondering if this is reasonable/possible. If not, it can be left as a 2-source helper and those parts of the code can be removed.
- Work still needs to be done, such as adding special cases for immediate loads, and maybe adding checks for divide by zero for idiv/ldiv at the same time. 

Any ideas or suggestions would be greatly appreciated!

0xdaryl:
FYI...  If there are multiple authors on a commit, you can list them in the commit by adding a line like `Co-authored-by: name <name@example.com>` for each contributor.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
...
+static inline TR::Register *
+evaluateBooleanTypeInteger(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg, uint8_t sizeOfOperation)
+   {
+   TR::Register *srcReg[3];
+   RegisterStatus srcStatus[3];
+   TR::Node *children[3];
+   TR::Register *trgReg;
+   bool assigned = false;
+
+   if(sizeOfOperation > 0)
```
0xdaryl: Can `sizeOfOperation` ever be 0?  If so, what is the semantics of that?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
...
+	}
+   return status;
+   }
+
+static inline TR::Register *
+evaluateBooleanTypeInteger(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg, uint8_t sizeOfOperation)
+   {
+   TR::Register *srcReg[3];
+   RegisterStatus srcStatus[3];
+   TR::Node *children[3];
```
0xdaryl: Do you need the `children` array?  It seems the children always map to a particular child on the Node and there's no need to keep them in a separate array.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
...
+	else
+   {
+      *childReg = cg->evaluate(child);
+      status = DO_NOT_TOUCH;
+	}
+   return status;
+   }
+
+static inline TR::Register *
+evaluateBooleanTypeInteger(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg, uint8_t sizeOfOperation)
```
0xdaryl: Can you make the CodeGenerator parameter the last parameter?

0xdaryl: I'm not sure I understand the name of this function.  What does "BooleanTypeInteger" mean?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
 #include "il/Node_inlines.hpp"
 #include "infra/Bit.hpp"
 
+typedef enum RegisterStatus {
+   OK_FOR_REUSE,
+   DO_NOT_TOUCH
+} RegisterStatus;
+
```
0xdaryl: Please add Doxygen comments for the functions that provide a @brief, @param, and @return as appropriate.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
...
+      srcStatus[2] = evaluateRegister(&srcReg[2], children[2], node, cg);
+   }
+   
+   for(int i=0; i < sizeOfOperation && !assigned; i++)
+   {
+      if(srcStatus[i] == OK_FOR_REUSE)
+	   {
+	      trgReg = srcReg[i];
+		   assigned = true;
+		   srcStatus[i] = DO_NOT_TOUCH;
```
0xdaryl: Shouldn't this break from the loop, or do you always want the target to point to the last srcReg found?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,116 @@
...
+      value = child->getLongInt();
+      if (constantIsUnsignedImm12(value))
+      {
+         //TODO: Cover special cases for immediate loads
+      }
+      else
+      {
+         *childReg = cg->allocateRegister();
+         loadConstant64(cg, parent, value, *childReg);
+         status = OK_FOR_REUSE;
```
0xdaryl: The register re-use mechanism should be driven by the reference count on the node whose value the register represents.  When you evaluate a node, its reference count represents the number of remaining uses of that node in subsequent trees (including the current one).  You can derive some properties of the register you need based on the reference count on the node.  For instance, if the reference count on the node you are evaluating is 1 then you know that this is the last unseen reference to the node in the trees and the register that represents its result effectively goes dead after this point (i.e., the register can be reused).  Furthermore, if the reference count on a register is >=1 and there hasn't been a register yet associated with a node then that tells you that this is the first occurrence of that node in the trees and hence this is the first evaluation of it.  A node with a reference count >1 tells you that the live range of the register associated with it does not die at this point.

I think your register status mechanism should be driven by the reference count and the register status on the node.  Perhaps you can think about whether the status mechanism is really required (and if so, why is it needed over and above the reference counting mechanism).

compiler/aarch64/codegen/BinaryEvaluator.cpp
0xdaryl: Is `sizeOfOperation` needed?  Why not ask the opcode how many children it has?

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: I believe we included an array to reduce the amount of code (some of the repeated if statements with the children can be converted into for loops, and we did this in a slightly newer version we have). If you would prefer to keep arrays out of it, we can change it. :)

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: In our current version of this, if 0 is passed into the sizeOfOperation field when the second child in the operation is 0 (in idiv and ldiv), it will handle division by zero by simply allocating a register and returning it (it will skip creating any instructions at all because it never enters any for loops or ifs). We were wondering if you wanted that case to be handled elsewhere (in the evaluator functions themselves), or if it would be okay for the helper to do it for us.

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: I think this loop will break when assigned is set to true (as soon as a target is found).

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: Thank you for the illustration, it is very helpful! We'll look into changing this as well.

compiler/aarch64/codegen/BinaryEvaluator.cpp
0xdaryl: In this implementation, the elements of `children` always map to the corresponding ordinal child position on the node, hence it is not needed.  If there is newer code that hasn't been pushed then I can reserve this comment until I see that code.

compiler/aarch64/codegen/BinaryEvaluator.cpp
0xdaryl: Yes, you're right.  Sorry I missed that.

compiler/aarch64/codegen/BinaryEvaluator.cpp
0xdaryl: This isn't obvious from the implementation provided.  From the current implementation, it appears this parameter maps to the number of children on the node.  If that's the case, then passing in "0" is special case.  If you have newer code that shows what you have in mind then please push that.

hsoontie:
@0xdaryl Here's a new commit. We decided to remove the passing of the number of children and the issue of the special 0 case because in the end it would be more efficient, and involve just a bit more code outside of this helper to handle divide by zero.
The array was also removed.
Let us know your thoughts on this version!

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+   if(child->getReferenceCount() == 1 && NULL == trgReg)
+      {
+      trgReg = *childReg;
+      }
+   }
+
+/**
+ * Generic helper that handles a number of similar binary operations.
+ * @param[in] node : calling node
+ * @param[in] regOp : the matching opcode of the caller
```
0xdaryl: This is the AArch64 instruction opcode you want to use.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+ * @param[in] cg : codegenerator
+ * @return target register
+ */
+static inline TR::Register *
+genericBinaryEvaluator(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg)
+   {
+   int sizeOfOperation = node->getNumChildren();
```
0xdaryl: Please use `int32_t` for integers to be consistent with rest of codebase.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
 #include "il/Node_inlines.hpp"
 #include "infra/Bit.hpp"
 
+/**
+ * Helper that evaluates a child node.
+ * @param[in] childReg : the register of the child node being looked at
```
0xdaryl: [in, out] ?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+      {
+      value = child->getLongInt();
+      //if(constantIsUnsignedImm12(value))
+         //{
+         //TODO: Cover special cases for immediate loads
+         //}
+      //else
+         //{
+         *childReg = cg->allocateRegister();
+         loadConstant64(cg, parent, value, *childReg);
```
0xdaryl: I think these two statements should just be: `*childReg = cg->evaluate(child)`.  It should do the same thing.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+ * @param[in] trgReg : the target register
+ * @param[in] child : the child node being looked at
+ * @param[in] parent : the parent of the child node being looked at
+ * @param[in] cg : codegenerator
+ */
+static inline void
+evaluateRegister(TR::Register **childReg, TR::Register *trgReg, TR::Node *child, TR::Node *parent, TR::CodeGenerator *cg)
+   {
+   int64_t value;
+	if (child->getOpCode().isLoadConst() && child->getRegister() == NULL)
```
0xdaryl: Fix indenting.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+
+/**
+ * Generic helper that handles a number of similar binary operations.
+ * @param[in] node : calling node
+ * @param[in] regOp : the matching opcode of the caller
+ * @param[in] cg : codegenerator
+ * @return target register
+ */
+static inline TR::Register *
+genericBinaryEvaluator(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::CodeGenerator *cg)
```
0xdaryl: If this is a binary evaluator, why isn't the number of children always "2" ?

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+         loadConstant64(cg, parent, value, *childReg);
+         //}
+      }
+   else
+      {
+      *childReg = cg->evaluate(child);
+      }
+   if(child->getReferenceCount() == 1 && NULL == trgReg)
+      {
+      trgReg = *childReg;
```
0xdaryl: It looks like the intent is to pass `trgReg` back to the caller.  If that's the case then I think the parameter should be declared as a `TR:Register **trgReg`, otherwise just the local pointer is updated.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,113 @@
...
+      generateTrg1Src2Instruction(cg, regOp, node, trgReg, srcReg[0], srcReg[1]);
+      }
+   else
+      {
+      generateTrg1Src3Instruction(cg, regOp, node, trgReg, srcReg[0], srcReg[1], srcReg[2]);
+      }
+
+   if(0 < sizeOfOperation && firstChild->getReferenceCount() > 1)
+      {
+      cg->stopUsingRegister(srcReg[0]);
```
0xdaryl: Calling `stopUsingRegister` manually should only be done for "scratch registers".  That is, registers that have a limited live range and are not associated with a node (i.e., they don't appear when you ask `node->getRegister()`).  When you decrement the reference count on a node and it determines that its reference count is 1 then it will implicitly call `stopUsingRegister` already.

So, you only want to be calling `stopUsingRegister` here for the scratch registers you created (via `allocateRegister`) that will not live past this evaluator.

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: Our intent was to make this function able to handle other simple evaluators outside of the binary evaluators, though I'm not sure:
1) if that is something you wanted us to do, or if you would rather us stick to binary evaluators for now
2) how many unary or ternary operations actually exist that are simple and would fit this function (The ones that I have seen myself require extra steps).

Even if we remove this functionality, I think the function can easily be reincorporated later if similar patterns are showing up elsewhere in other evaluators. :)

0xdaryl: I think limiting this to just binary operators is sufficient for now.  Over-generalizing it to one or three children may overcomplicate it, and in fact cause us to miss opportunities or introduce bugs.

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: Does calling evaluate() handle the constant special case more easily than doing what's there? (I put this in because it follows the idea of the others, and because I thought it was meant to simplify things.)

hsoontie:
@0xdaryl Here's another update, not including compressing the special case code to just using evaluate(). Once I had made the appropriate changes and went to the immediate case I realized it may be better to merge the two functions into one (especially because I'm converting it back to binary cases only).
Going off of your comment on Aaron's ladd/lsub PR, I tried to check appropriately for reusable registers.
Please let me know your thoughts! Thank you!

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,67 @@
 #include "il/Node_inlines.hpp"
 #include "infra/Bit.hpp"
 
+/**
+ * Generic helper that handles a number of similar binary operations.
+ * @param[in] node : calling node
+ * @param[in] regOp : the target AArch64 instruction opcode
+ * @param[in] regOp : the matching AArch64 immediate instruction opcode
```
0xdaryl: regOpImm

0xdaryl: Also, please document here the assumption from below that regOp == regOpImm implies there isn't an immediate form instruction.

compiler/aarch64/codegen/BinaryEvaluator.cpp
```diff
@@ -30,6 +30,67 @@
...
+      /* When regOp = regOpImm, an immediate version of the instruction does not exist. */
+      if(constantIsUnsignedImm12(value) && regOp != regOpImm)
+         {
+         generateTrg1Src1ImmInstruction(cg, regOpImm, node, trgReg, src1Reg, value);
+         isImm = true;
+         }
+      else
+         {
+         src2Reg = cg->allocateRegister();
+         loadConstant64(cg, node, value, src2Reg);
```
0xdaryl: Because `loadConstant64` has the potential of generating multiple instructions, should you be distinguishing between 32-bit and 64-bit constants here?  i.e., if you're loading an integer constant, will you sometimes end up generating more instructions than you need because you are using `loadConstant64` rather than `loadConstant32` ?

0xdaryl:
@genie-omr build aarch64

compiler/aarch64/codegen/BinaryEvaluator.cpp
hsoontie: I think it may be worth it in the long run to check, yes.
@aarongraham9 @jeanlego @marufunb , any thoughts on this?

aarongraham9: Yes. I think checking for and distinguishing between 32-bit and 64-bit `const` loads would be a better idea here. Any further thoughts @jeanlego & @marufunb?

marufunb:
Yes I also think checking for both 32 bit and 64 bit constant loads is good idea..

Thanks,
Md. Mahbubur Rahman (Maruf)
Graduate Academic Assistant at
IBM/UNB Centre of Advanced Studies- Atlantic (CASA)
Phone: 5064714698
Faculty of Computer Science,
University of New Brunswick

________________________________

hsoontie:
@0xdaryl Here is another update. Please let me know your thoughts!

0xdaryl:
Code changes look good.  Please squash the commits, remove the word "blanket" from the commit title (it's not really a term we use in the compiler :-) ), and remove the WIP.  I'll re-run CI and this should be ready to merge.

0xdaryl:
@genie-omr build aarch64

knn-k:
See #4161 for a fix in `genericBinaryEvaluator()`.

Issue #3326, opened 2018-12-12
Agenda for Dec. 12, 2018 Compiler Architecture Meeting


0xdaryl:
### Agenda

1.  Create an issue template for new compiler IL opcode requests  (#3236) [ @0xdaryl   ]
2.  API to ask about the type a compare opcode operates on (#3302) [ @andrewcraik ]
3.  TR namespace usage (#3306) [ @Leonardo2718 ]

[Replay](https://www.youtube.com/watch?v=aKLJzEaWMXg)

### Details
Wednesday, Dec. 12, 2018 @ 15:30 EST (GMT -5), 90 minutes maximum

Join [Webex](https://ibm.webex.com/ibm/j.php?MTID=m0b56f0ccbf2ebb90a67094b48b7e1ec8)

Meeting number (access code): 929 612 565
Meeting password: omrrocks!

North America call-in number: 1-844-531-0958

[Other global call-in numbers (for audio only)](https://ibm.webex.com/cmp3300/webcomponents/widget/globalcallin/globalcallin.do?siteurl=ibm&serviceType=MC&eventID=729622812&tollFree=1)

**This meeting will be recorded and posted to the [Eclipse OMR YouTube channel](https://www.youtube.com/channel/UCO8MkTRrFnLefi5nqReOMMA).**

PR #3520, opened 2019-01-28
Refine the guidance on using the TR namespace

### LINKED ISSUE #2084: Document proper use of the TR:: namespace

### LINKED ISSUE #3306: Revisit philosophy and use of the TR:: Namespace
#3287 has raised some questions about whether non-extensible classes in the OMR compiler component should go in the `TR` namespace or the `OMR` namespace. Although a discussion that happened over a year ago provides some guidance for this, the lack of consensus on the topic now suggests that further clarification is needed and a potential change of guideline may be warranted.

To summarize various issues related to the topic:

I believe the original discussion about when to use `TR::` vs. `OMR::` began in #788. The conclusion resulted in #817 which added the following paragraph in [Extensible_Classes_in_OMR.md](https://github.com/eclipse/omr/blob/master/doc/compiler/extensible_classes/Extensible_Classes_in_OMR.md#namespaces):

> In the case of classes that aren't extensible, It is perfectly legitimate to create a class directly in the TR:: namespace if it's complete and can and should be used as-is.

The paragraph can be interpreted to mean that non-extensible classes *should* go in the `TR` namespace (although that's not strictly what it says). A follow up issue like #1650 seems to support this interpretation.

Although not directly related, #1706 and #1885 discuss using component-level namespaces and moving class in the compiler component from `OMR::` to `OMR::Compiler::`. The issues don't address what goes and doesn't go in the `TR` namespace but do mention some subtleties of how the namespace is used.

Finally, #2084 proposes to clarify "proper use" of the `TR` namespace and to move the documentation outside of [Extensible_Classes_in_OMR.md](https://github.com/eclipse/omr/blob/master/doc/compiler/extensible_classes), since the namespace is also used for non-extensible classes.

Recent discussions, including #3287, have put into question whether the guideline to put non-extensible classes in the `TR` namespace is appropriate. This issue aims to track people's current thoughts on the topic and to develop more complete and precise guidelines for the use of `TR::`.

A non-executive list of points to consider:

- what are the implications of using the `TR` vs. `OMR` (vs. `OMR::Compiler`) namespaces for non-extensible classes in downstream projects?
- what work will have to be done to move existing non-extensible classes from `TR::` to `OMR::` (or `OMR::Compiler::`?
- what should happen with classes that use the `TR_` prefix in stead of `TR::` (a convention used in the compiler component long before it was opensourced)?

FYI @0xdaryl @mstoodle @fjeremic @vijaysun-omr

0xdaryl:
Per the discussion at the Compiler Architecture Meeting (#3326)
and the resolution in #3306, the `TR` namespace shall be used
for the public API of the compiler component that is consumable
by downstream projects.  This PR clarifies the language and
references the umbrella issue that tracks the work to migrate
the existing public API to the `TR` namespace.

Closes #3306
Closes #2084

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2016, 2018 IBM Corp. and others
+Copyright (c) 2016, 2019 IBM Corp. and others
 
 This program and the accompanying materials are made available under
 the terms of the Eclipse Public License 2.0 which accompanies this
@@ -10,7 +10,7 @@ is available at https://www.apache.org/licenses/LICENSE-2.0.
 This Source Code may also be made available under the following
 Secondary Licenses when the conditions for such availability set
 forth in the Eclipse Public License, v. 2.0 are satisfied: GNU
-General Public License, version 2 with the GNU Classpath 
+General Public License, version 2 with the GNU Classpath
 Exception [1] and GNU General Public License, version 2 with the
 OpenJDK Assembly Exception [2].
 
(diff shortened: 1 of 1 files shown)


0xdaryl:
@Leonardo2718, would you mind reviewing/merging please?

Leonardo2718:
Documentation only change so no builds are required. Merging.

PR #3523, opened 2019-01-29
Allow alternative path to merge to else block

liqunl:
If `elseBlock` and `mergeBlock` are identical, the user wants to create
an alternate path that after completion will jump back to the fall
through of the if test.

```
      ifBlock
       |    \
       |     \
       |   thenBlock
       |    /
       |   /
   mergeBlock/elseBlock
```
  
Under current implementation, `ifBlock` is constructed by splitting
`elseBlock` after prepending `ifTree` to `elseBlock`. The old
`elseBlock` will become `ifBlock`, `elseBlock` will get a new value
after split. If `mergeBlock` is the same as `elseBlock`, then it will be
the same as `ifBlock` after split. The alternate path `thenBlock` will
merge to `ifBlock` if we don't update `mergeBlock`, resulting in a
infinite loop. Fix it.

### DIFF
--- a/compiler/optimizer/OMRTransformUtil.cpp
+++ b/compiler/optimizer/OMRTransformUtil.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corp. and others
+ * Copyright (c) 2000, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -406,9 +406,15 @@ OMR::TransformUtil::createConditionalAlternatePath(TR::Compilation* comp,
    {
    cfg->setStructure(0);
 
+   bool mergeToElseBlock = (elseBlock == mergeBlock);
+
    TR::Block* ifBlock = elseBlock;
    ifBlock->prepend(ifTree);
    elseBlock = ifBlock->split(ifTree->getNextTreeTop(), cfg, false /*fixupCommoning*/, true /*copyExceptionSuccessors*/);
+
+   // Since `elseBlock` is changed above, update `mergeBlock`
+   if (mergeToElseBlock)
+      mergeBlock = elseBlock;
 
    TR::Block * thenBlock = TR::Block::createEmptyBlock(thenTree->getNode(), comp, 0, elseBlock);
    if (markCold)


compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -406,10 +406,15 @@ OMR::TransformUtil::createConditionalAlternatePath(TR::Compilation* comp,
    {
    cfg->setStructure(0);
 
+   bool mergeToElseBlock = elseBlock == mergeBlock ? true : false;
```
Leonardo2718: Why not just:

```suggestion
   bool mergeToElseBlock = elseBlock == mergeBlock;
```
Or maybe:
```suggestion
   bool mergeToElseBlock = (elseBlock == mergeBlock);
```

compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -406,10 +406,15 @@ OMR::TransformUtil::createConditionalAlternatePath(TR::Compilation* comp,
...
+
    TR::Block* ifBlock = elseBlock;
    ifBlock->prepend(ifTree);
    elseBlock = ifBlock->split(ifTree->getNextTreeTop(), cfg, false /*fixupCommoning*/, true /*copyExceptionSuccessors*/);
 
+   if (mergeToElseBlock)
```
Leonardo2718: Is there a reason for using a dedicated variable instead of just:

```suggestion
   if (elseBlock == mergeBlock)
```

Also, given that `elseBlock` and `mergeBlock` are pointers, if `elseBlock == mergeBlock` then isn't `mergeBlock = elseBlock` a no-op?

liqunl: `elseBlock` is changed at line 413. `split` creates a new block and assign it to `elseBlock`.

Leonardo2718: I see. Thank you for clarifying that. :slightly_smiling_face:

Leonardo2718:
@genie-omr build all

liqunl:
@Leonardo2718 My bad, forgot the copyright. Updated the commit with copyright change.

Leonardo2718:
@genie-omr build all

vijaysun-omr:
While I think I understand the change in the narrow sense of what you changed with this PR, I feel it would be useful to understand better 
a) the control flow in the case when "else block" is the same as the "merge block" 
and 
b) what was going wrong that necessitated this change.

i.e. the change has the feel of a functional bug fix but it would help if that bug was clearly explained.

liqunl:
@vijaysun-omr Updated the commit message to explain the problem this change is trying to fix.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
The longer description gives a better sense for the problem being solved by this code change. Merging since reviews are done and checks have passed.

PR #3524, opened 2019-01-29
WIP: Add rdbar handling in NullChk evaluator

dchopra001:
NullChk's special case in compressed pointers where the node to be null checked is not the iiload but its child (ex. an aload) will now also handle the case where the child is rdbar node.


fjeremic:
@genie-omr build all

dchopra001:
> This code really belongs in OpenJ9. I see the x86 implementation already exists there. @dchopra001 could you open up a beginner issues to migrate this code on all codegens to OpenJ9?
https://github.com/eclipse/omr/issues/3530

PR #3531, opened 2019-01-30
Fix vector load/store instruction large displacement bug

NigelYiboYu:
Fix vector load/store instruction large displacement bug

Fix a problem in VRS, VSI, and VRV vector-storage instruction
formats to handle large displacements correctly.

The fix consists of two parts: an instruction selection phase change and
a binary encoding change. Both of these changes aim to use a scratch register
to hold the base+displacement value if a vector load/store instruction has a
displacement that exceeds the 12-bit signed integer range.

### DIFF
--- a/compiler/z/codegen/OMRMemoryReference.cpp
+++ b/compiler/z/codegen/OMRMemoryReference.cpp
@@ -2034,6 +2034,11 @@ OMR::Z::MemoryReference::alignmentBumpMayRequire4KFixup(TR::Node * node, TR::Cod
    return false;
    }
 
+/**
+ * \brief
+ *  Enforces format limits on several instruction formats that have a displacement and base register without
+ *  an index register. This routine was originally meant for SS formats and is now used on SS, VRS, VRV and VSI formats.
+*/
 TR::Instruction *
 OMR::Z::MemoryReference::enforceSSFormatLimits(TR::Node * node, TR::CodeGenerator * cg, TR::Instruction *preced)
    {
@@ -2047,11 +2052,7 @@ OMR::Z::MemoryReference::enforceSSFormatLimits(TR::Node * node, TR::CodeGenerato
 TR::Instruction *
 OMR::Z::MemoryReference::enforceRSLFormatLimits(TR::Node * node, TR::CodeGenerator * cg, TR::Instruction *preced)
    {
-   bool forceDueToAlignmentBump = self()->alignmentBumpMayRequire4KFixup(node, cg);
-   // call separateIndexRegister first so any large offset is handled along with the index register folding
-   preced = self()->separateIndexRegister(node, cg, true, preced, forceDueToAlignmentBump); // enforce4KDisplacementLimit=true
-   preced = self()->enforce4KDisplacementLimit(node, cg, preced, false, forceDueToAlignmentBump);
-   return preced;
+   return self()->enforceSSFormatLimits(node, cg, preced);
    }
 
 TR::Instruction *
@@ -3032,7 +3033,18 @@ OMR::Z::MemoryReference::generateBinaryEncoding(uint8_t * cursor, TR::CodeGenera
              instructionFormat == TR::Instruction::IsSS1 ||
              instructionFormat == TR::Instruction::IsSS2 ||
              instructionFormat == TR::Instruction::IsSS4 ||
-             instructionFormat == TR::Instruction::IsS)
+             instructionFormat == TR::Instruction::IsS   ||
+             instructionFormat == TR::Instruction::IsRIS ||
+             instructionFormat == TR::Instruction::IsRRS ||
+             instructionFormat == TR::Instruction::IsSSE ||
+             instructionFormat == TR::Instruction::IsSSF ||
+             // Vector instruction formats that have 12-bit displacements and no index registers
+             instructionFormat == TR::Instruction::IsVSI ||
+             instructionFormat == TR::Instruction::IsVRV ||
+             instructionFormat == TR::Instruction::IsVRSa||
+             instructionFormat == TR::Instruction::IsVRSb||
+             instructionFormat == TR::Instruction::IsVRSc||
+             instructionFormat == TR::Instruction::IsVRSd)
             {
             TR_ASSERT_FATAL(base != NULL, "Expected non-NULL base register for long displacement on %s [%p] instruction", instr->getOpCode().getMnemonicName(), instr);
 
--- a/compiler/z/codegen/S390GenerateInstructions.cpp
+++ b/compiler/z/codegen/S390GenerateInstructions.cpp
@@ -2051,24 +2051,38 @@ TR::Instruction * generateVRRiInstruction(
 /* Note subtle differences between register types and optionality of masks between these 3*/
 TR::Instruction *
 generateVRSaInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op, TR::Node * n, TR::Register * targetReg, TR::Register * sourceReg, TR::MemoryReference * mr,
-                        uint8_t mask4 /* 4 bits */)
+                        uint8_t mask4 /* 4 bits */, TR::Instruction* preced)
    {
-   return new (INSN_HEAP) TR::S390VRSaInstruction(cg, op, n, targetReg, sourceReg, mr, mask4);
+   preced = mr->enforceSSFormatLimits(n, cg, preced);
+
+   if (preced)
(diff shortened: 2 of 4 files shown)


fjeremic:
We really ought to add Tril testing for such cases. This would have been caught long ago. @Leonardo2718 do we have a timeline when we may support processor detection for Tril testing? Such a test for this bug would have required z13 machines.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -3032,7 +3032,14 @@ OMR::Z::MemoryReference::generateBinaryEncoding(uint8_t * cursor, TR::CodeGenera
...
              instructionFormat == TR::Instruction::IsSS4 ||
-             instructionFormat == TR::Instruction::IsS)
+             instructionFormat == TR::Instruction::IsS   ||
+             // Vector instruction formats that have 12-bit displacements and no index registers
+             instructionFormat == TR::Instruction::IsVSI ||
+             instructionFormat == TR::Instruction::IsVRV ||
+             instructionFormat == TR::Instruction::IsVRSa||
+             instructionFormat == TR::Instruction::IsVRSb||
+             instructionFormat == TR::Instruction::IsVRSc||
+             instructionFormat == TR::Instruction::IsVRSd)
```
fjeremic: @NigelYiboYu did you confirm these are the only formats without an index register?

NigelYiboYu: Yes, these are the vector-storage formats (z13 and above) that don't have an index register. Their memory reference are **D(B)**, with D being 12-bits signed.

The only vector-storage format that has **D(X,B)** is VRS format.

Let me check non-vector formats....

NigelYiboYu: For non-vector instructions, we missed some trivial formats: **RIS, RRS, SSE and SSF.** I'll add them just in case. However, note that these are either not used in the JIT code base or not used for large jumps in practice. 

There are only 8 RIS and RRS instructions and they are all compare-and-branch instructions such as CIB, CGIB, and etc. They usually don't jump to far.

We have no use of SSE in the JIT.

SSF format consists of 5 instructions, two of which are used: LDP and LGPD.

fjeremic: This is certainly an oversight. We should fix all the ones you noted as well. Let's add them to the list.

Leonardo2718:
@fjeremic 

> @Leonardo2718 do we have a timeline when we may support processor detection for Tril testing? Such a test for this bug would have required z13 machines.

#3539 :slightly_smiling_face:

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -2047,11 +2052,7 @@ OMR::Z::MemoryReference::enforceSSFormatLimits(TR::Node * node, TR::CodeGenerato
 TR::Instruction *
 OMR::Z::MemoryReference::enforceRSLFormatLimits(TR::Node * node, TR::CodeGenerator * cg, TR::Instruction *preced)
    {
-   bool forceDueToAlignmentBump = self()->alignmentBumpMayRequire4KFixup(node, cg);
```
NigelYiboYu: This is exactly the same as `enforceSSFormatLimits`. Remove redundant code.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -2076,9 +2077,7 @@ OMR::Z::MemoryReference::enforce4KDisplacementLimit(TR::Node * node, TR::CodeGen
    if (self()->ignoreNegativeOffset())
       return preced;
 
-   TR_ASSERT(!forcePLXFixup, "This logic is only used for markAndAdjustForLongDisplacementIfNeeded. You probably want the other flag.");
```
NigelYiboYu: Remove code associated with `forcePLXFixup` since it's dead.

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2051,24 +2051,38 @@ TR::Instruction * generateVRRiInstruction(
 /* Note subtle differences between register types and optionality of masks between these 3*/
 TR::Instruction *
 generateVRSaInstruction(TR::CodeGenerator * cg, TR::InstOpCode::Mnemonic op, TR::Node * n, TR::Register * targetReg, TR::Register * sourceReg, TR::MemoryReference * mr,
-                        uint8_t mask4 /* 4 bits */)
+                        uint8_t mask4 /* 4 bits */, TR::Instruction* preced)
    {
-   return new (INSN_HEAP) TR::S390VRSaInstruction(cg, op, n, targetReg, sourceReg, mr, mask4);
+   if (mr != NULL) preced = mr->enforceSSFormatLimits(n, cg, preced);
+
+   if (preced)
```
NigelYiboYu: A comment on the use of two different constructors here because I can't find a better place to put it. We have to use two constructors here for a lot of instruction formats because the root `OMR::Instruction` has two constructors and they use different `preceding instructions`. We can misplace an instruction if the `preceding instruction` is not used properly.

compiler/z/codegen/OMRMemoryReference.cpp
fjeremic: Are you certain? I recall @dchopra001 tried to remove it at one point and it caused issues. I forget what the particular issue was though. Perhaps touch base with him.

NigelYiboYu: Will revert this change.

fjeremic: For reference I do want it removed if we can do so without introducing bugs as it is clearly PLX specific.

compiler/z/codegen/OMRMemoryReference.cpp
```diff
@@ -2047,11 +2052,7 @@ OMR::Z::MemoryReference::enforceSSFormatLimits(TR::Node * node, TR::CodeGenerato
 TR::Instruction *
 OMR::Z::MemoryReference::enforceRSLFormatLimits(TR::Node * node, TR::CodeGenerator * cg, TR::Instruction *preced)
    {
-   bool forceDueToAlignmentBump = self()->alignmentBumpMayRequire4KFixup(node, cg);
-   // call separateIndexRegister first so any large offset is handled along with the index register folding
-   preced = self()->separateIndexRegister(node, cg, true, preced, forceDueToAlignmentBump); // enforce4KDisplacementLimit=true
-   preced = self()->enforce4KDisplacementLimit(node, cg, preced, false, forceDueToAlignmentBump);
-   return preced;
+   return enforceSSFormatLimits(node, cg, preced);
```
Leonardo2718: Linter is suggesting:
```suggestion
   return self()->enforceSSFormatLimits(node, cg, preced);
```

compiler/z/codegen/S390GenerateInstructions.cpp
```diff
@@ -2078,17 +2092,28 @@ generateVRSdInstruction(
                       TR::Node                * n          ,
                       TR::Register            * targetReg  ,   /* VRF */
                       TR::Register            * sourceReg3 ,   /* GPR R3 */
-                      TR::MemoryReference     * mr        )
+                      TR::MemoryReference     * mr         ,
+                      TR::Instruction         * preced)
    {
-   return new (INSN_HEAP) TR::S390VRSdInstruction(cg, op, n, targetReg, sourceReg3, mr);
+   if (mr != NULL) preced = mr->enforceSSFormatLimits(n, cg, preced);
```
fjeremic: For all of the changes in S390GenerateInstructions.cpp the memory reference argument cannot be NULL. Can we remove these checks?

fjeremic: Whoops did not mean to approve. That was an accident. We'll need to make this change before we can merge as well as the change from Leo above.

compiler/z/codegen/OMRMemoryReference.cpp
NigelYiboYu: There's no issue in my testing. However, this flag is actually used in a few places so it's not safe to blindly remove it.

compiler/z/codegen/S390GenerateInstructions.cpp
NigelYiboYu: NULL checks removed.

compiler/z/codegen/OMRMemoryReference.cpp
NigelYiboYu: Fixed.

fjeremic:
@genie-omr build all

fjeremic:
AppVeoyr failed due to infrastructure reasons.

PR #3537, opened 2019-01-31
Dont crash when seeing invalid tree while tracing

cathyzhyi:
When seeing NULLChk without NULLChkReference while printing trees
to tracing log, we should try to continue printing the complete
trees which makes debugging easier.

### DIFF
--- a/compiler/ras/Tree.cpp
+++ b/compiler/ras/Tree.cpp
@@ -1585,7 +1585,9 @@ TR_Debug::printNodeInfo(TR::Node * node, TR_PrettyPrinterString& output, bool pr
 
    if (node->getOpCode().isNullCheck())
       {
-      output.append(" on %s%dn", globalIndexPrefix.getStr(), node->getNullCheckReference()->getGlobalIndex());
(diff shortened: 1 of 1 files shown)


compiler/ras/Tree.cpp
```diff
@@ -1585,7 +1585,9 @@ TR_Debug::printNodeInfo(TR::Node * node, TR_PrettyPrinterString& output, bool pr
 
    if (node->getOpCode().isNullCheck())
       {
-      output.append(" on %s%dn", globalIndexPrefix.getStr(), node->getNullCheckReference()->getGlobalIndex());
+      if (node->getNullCheckReference())
+         output.append(" on %s%dn", globalIndexPrefix.getStr(), node->getNullCheckReference()->getGlobalIndex());
+      else output.append(" on null NullCheckReference ----- INVALID tree!!\n");
```
0xdaryl: From a consistency POV, this message ends with a newline but all others in a similar context don't.  Arguably, trying to print an invalid tree is breaking the mold, but can you confirm that the newline is what you want here?

PR #3539, opened 2019-02-01
 Implement a primitive SKIP_IF macro to allow tests to be skipped

Leonardo2718:
This change set introduces the SKIP_IF macro, which allows Tril-based compiler tests to be conditionally skipped without failing or disabling the test. This commit only intends to introduce the syntax for skipping tests and the underlaying machinery needed for skipping test execution. Support for features such as logging and reporting of skipped tests is still primitive. No built-in support for platform detection is provided.

The basic syntax is for skipping a test is:

    SKIP_IF(<condition>, <reason>);

The <condition> parameter can be any boolean expression and specifies when a test should be skipped. The <reason> parameter must be a value from the `SkipReason` enum; it is mainly intended for use in logging and reporting, although the current implementation makes little use of it.

Optionally, a "skip message" can also be streamed using the same syntax as is used Google Test's ASSERT*() macros:

    SKIP_IF(<condition>, <reason>) << <message>;

When a test is skipped it will print a message stating the reason, the test name, and the skip message. For example:

    [----------] 2 tests from SkipTest
    Unsupported Feature: Skipping test: SkipIfTrue
        Test is itentionally skipped to verify that skipping works
    [----------] 2 tests from SkipTest (0 ms total)

Issue #1843

### DIFF
--- a/fvtest/compilertriltest/JitTest.hpp
+++ b/fvtest/compilertriltest/JitTest.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corp. and others
+ * Copyright (c) 2017, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -24,7 +24,7 @@
 
 #include <gtest/gtest.h>
 #include <vector>
-#include <stdexcept> 
+#include <stdexcept>
 #include "control/Options.hpp"
 #include "optimizer/Optimizer.hpp"
 #include "ilgen/MethodBuilder.hpp"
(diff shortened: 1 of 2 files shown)


Leonardo2718:
@genie-omr build all

Leonardo2718:
As I've mentioned in the commit message, this PR only adds the basic machinery needed to skip tests; built-in platform detection is not yet implement. However, I'm hoping that we can start by using the port library to write skip conditions that will satisfy our needs.

fvtest/compilertriltest/JitTest.hpp
```diff
@@ -312,4 +313,111 @@ std::vector<std::tuple<L,R>> const_value_pairs()
...
+ *
+ * These values are intended to be short descriptions of why a test is skipped
+ * and are mostly useful for logging and reporting purposes. Additional explenations
+ * should be specified in the skip message.
+ */
+enum SkipReason {
+   KnownBug,               // test is skipped because of a known bug
+   MissingImplementation,  // feature under test is not implemented
+   UnsupportedFeature,     // feature under test is not supported
+   NumSkipReasons_,        // DO NOT USE IN USER CODE
```
fjeremic: Are we able to use `std::numeric_limits<enum SkipReason>::max()` or do not all our targets support it?

Leonardo2718: I don't think that ever works: https://stackoverflow.com/questions/9201865/why-is-c-numeric-limitsenum-typemax-0 . :disappointed:

PR #3540, opened 2019-02-01
Replace visit count in Inliner with node checklist

cathyzhyi:
Inliner and several local transformations inside Inliner are using
visitcount to track visited nodes or call nodes that has already been
inlined and the code logic interferes with each other. Changing the
local transformations in Inliner to use NodeChecklist. Global use of
visitcount in Inliner needs more investigation and will be fixed in a
separate commit

https://github.com/eclipse/openj9/issues/3340

### DIFF
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1364,30 +1364,15 @@ TR_DumbInliner::analyzeCallSite(
 static TR::TreeTop * cloneAndReplaceCallNodeReference(TR::TreeTop *, TR::Node *, TR::Node *, TR::TreeTop *, TR::Compilation *);
 static void addSymRefsToList(List<TR::SymbolReference> &, List<TR::SymbolReference> &);
 
-static TR::Node *findPotentialDecompilationPoint(TR::Node *node, TR::Compilation *comp)
-   {
-   if (node->getVisitCount() == comp->getVisitCount())
-      return NULL;
-   else
-      node->setVisitCount(comp->getVisitCount());
-
-   TR::Node *result = NULL;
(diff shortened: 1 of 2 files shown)


Leonardo2718:
@genie-omr build all

cathyzhyi:
There is a small chance that this change might affect inliner heuristic. Tested with major benchmarks and didn't see any difference in performance.

vijaysun-omr:
Change looks safe to me but a review from @andrewcraik would be good.

andrewcraik:
@vijaysun-omr this is good to merge - switching away from visit counts and cleaning up the extra params and stuff can only be a good thing.

andrewcraik:
@Leonardo2718 or @fjeremic or @0xdaryl could one of you get this merged please? Vijay has approved, I'm good with it and the problem it fixes is causing OpenJ9 some grief in testing.

0xdaryl:
I would prefer not to break the committer covenant we have established in the OMR committer guide and merge another committer's self-assigned PR without their explicit consent to do so.  I have just asked @vijaysun-omr privately to either commit this or allow another committer to do so.  Hopefully this can happen today.

Presumably, this commit can be applied directly to the openj9-omr repo if needed urgently downstream.

fjeremic:
Second bullet in [1] for reference. We've had situations before where the assigned committer had "inside knowledge" unbeknownst to others and a merge broke things downstream. This is why we put this new rule in place.

[1] https://github.com/eclipse/omr/blob/master/doc/GuideForCommitters.md#general-guidelines

andrewcraik:
@0xdaryl and @fjeremic ok thank you for the reply - I only asked because of the > 1 week since the approval and request for my comment. I can discuss with the other committers at OpenJ9 about whether we want to do a pre-emptive pick into openj9-omr or continue to wait.

PR #3542, opened 2019-02-04
Misc. CodeCacheManager API refactoring

0xdaryl:
* Relocate CodeCache::allocate function to CodeCacheManager
* Delete unused functionality from CodeCacheManager
* Remove reservationInterfaceCache from CodeCacheManager

### DIFF
--- a/compiler/runtime/OMRCodeCacheManager.hpp
+++ b/compiler/runtime/OMRCodeCacheManager.hpp
@@ -171,6 +171,23 @@ public:
                                 bool needsToBeContiguous,
                                 bool isMethodHeaderNeeded=true);
 
+   /**
+    * @brief Allocate and initialize a new code cache from a new memory segment
+    *
+    * @param[in] segmentSizeInBytes : segment size to create the code cache from
+    * @param[in] reservingCompilationTID : thread id of the requestor
+    *               if (reservingCompilationTID)
+    *                  <= -2 : no reservation is requested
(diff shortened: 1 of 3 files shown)


compiler/runtime/OMRCodeCacheManager.hpp
```diff
@@ -171,6 +171,24 @@ class OMR_EXTENSIBLE CodeCacheManager
...
+   /**
+    * @brief Allocate and initialize a new code cache from a new memory segment
+    *
+    * @param[in] segmentSizeInBytes : segment size to create the code cache from
+    * @param[in] reservingCompilationTID : thread id of the requestor
+    *               if (reservingCompilationTID)
+    *                  >= -1 : the new code cache will be reserved
+    *                  <= -2 : no reservation is requested
+    *                  == -1 : the application thread is requesting the allocation
+    *                  >=  0 : a compilation thread is requesting the allocation
```
Leonardo2718: I find listing `>= -1 : the new code cache will be reserved` a bit confusing. Could use something like this instead:

```
    *                  <= -2 : no reservation is requested
    *                  == -1 : the application thread is requesting the allocation; new code cache will be reserved
    *                  >=  0 : a compilation thread is requesting the allocation; new code cache will be reserved
```

0xdaryl: Agreed.  Fixed and pushed.

Leonardo2718:
AppVeyor build failed because of infra problems. Restarted.

PR #3549, opened 2019-02-06
Add getDelegate to Scavenger

rwy0717:

### DIFF
--- a/gc/base/standard/Scavenger.hpp
+++ b/gc/base/standard/Scavenger.hpp
@@ -599,6 +599,10 @@ public:
 	static MM_Scavenger *newInstance(MM_EnvironmentStandard *env, MM_HeapRegionManager *regionManager);
 	virtual void kill(MM_EnvironmentBase *env);
 
+#if defined (OMR_GC_SCAVENGER_DELEGATE)
+	MM_ScavengerDelegate* getDelegate() { return &_delegate; }
+#endif /* OMR_GC_SCAVENGER_DELEGATE */
+
 	/* Read Barrier Verifier specific methods */
 #if defined(OMR_ENV_DATA64) && !defined(OMR_GC_COMPRESSED_POINTERS)
 	virtual void scavenger_poisonSlots(MM_EnvironmentBase *env);


gc/base/standard/Scavenger.hpp
```diff
@@ -599,6 +599,11 @@ class MM_Scavenger : public MM_Collector
 	static MM_Scavenger *newInstance(MM_EnvironmentStandard *env, MM_HeapRegionManager *regionManager);
 	virtual void kill(MM_EnvironmentBase *env);
 
+#if defined (OMR_GC_SCAVENGER_DELEGATE)
```
charliegracie: Why are there two? Also why does anyone need direct access to the delegate?

rwy0717: > Why are there two?

One is for accessing a constant delegate when the scavenger is const. I'll get rid of it.


> Also why does anyone need direct access to the delegate?

In j9, I want to move the following scavenge-specific flags from clij to the scavenger delegate:

```c++
#if defined(OMR_GC_MODRON_SCAVENGER)
       volatile bool _scavenger_shouldScavengeFinalizableObjects; /**< Set to true at the beginning of a collection if there are any pending finalizable objects */
       volatile bool _scavenger_shouldScavengeUnfinalizedObjects; /**< Set to true at the beginning of a collection if there are any unfinalized objects */
       volatile bool _scavenger_shouldScavengeSoftReferenceObjects; /**< Set to true if there are any SoftReference objects discovered */
       volatile bool _scavenger_shouldScavengeWeakReferenceObjects; /**< Set to true if there are any WeakReference objects discovered */
       volatile bool _scavenger_shouldScavengePhantomReferenceObjects; /**< Set to true if there are any PhantomReference objects discovered */
#if defined(J9VM_GC_FINALIZATION)
       bool _scavenger_finalizationRequired; /**< Scavenger variable used to determine if finalization should be triggered */
#endif /* J9VM_GC_FINALIZATION */
#endif /* OMR_GC_MODRON_SCAVENGER */
```

the `ScavengerRootScanner` and `ScavengerRootClearer` need access to these flags.

charliegracie:
@genie-omr build all

PR #3554, opened 2019-02-07
Remove dead code

NigelYiboYu:
Remove isAddressOfStaticSymRefWithLockedReg() and
isAddressOfPrivateStaticSymRefWithLockedReg() since they always return
false.


0xdaryl:
@genie-omr build zlinux,zos

Only Z-specific files are changed so a full genie build is not required.

PR #3555, opened 2019-02-08
Add double map API for LINUX and balanced GC policy

bragaigor:
Include one test in fvtest to test reserve API with file handles

This is the initial step for the upcoming arraylet double mapping API
which will be included in incremental subsequent PRs

Co-authored-by: Charlie Gracie <charlie.gracie@gmail.com>

### DIFF
--- a/fvtest/porttest/omrvmemTest.cpp
+++ b/fvtest/porttest/omrvmemTest.cpp
@@ -45,6 +45,8 @@
 #define TWO_GIG_BAR 0x7FFFFFFF
 #define ONE_MB (1*1024*1024)
 #define FOUR_KB (4*1024)
+#define SIXTEEN_KB (16*1024)
+#define SIXTEEN_MB (0x1000000LL)
 #define TWO_GB	(0x80000000LL)
 #define SIXTY_FOUR_GB (0x1000000000LL)
 
@@ -54,6 +56,8 @@
 #define D512M (512*1024*1024)
 #define D256M (256*1024*1024)
 #define DEFAULT_NUM_ITERATIONS 50
+#define ARRAYLET_COUNT 8
+#define LEAF_SIZE_LIMIT (ONE_MB * 2)
 
 #define MAX_ALLOC_SIZE 256 /**<@internal largest size to allocate */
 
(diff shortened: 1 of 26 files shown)


fvtest/porttest/omrvmemTest.cpp
```diff
@@ -388,6 +388,122 @@ TEST(PortVmemTest, vmem_test1)
...
+	int32_t lastErrorNumber = 0;
+
+	reportTestEntry(OMRPORTLIB, testName);
+
+	/* First get all the supported page sizes */
+	pageSizes = omrvmem_supported_page_sizes();
+#if defined(J9ZOS390)
+	pageFlags =
+#endif /* J9ZOS390 */
+		omrvmem_supported_page_flags();
```
charliegracie: shouldn't this call be inside the `#if defined`?  It looks like you copied most of this from another test but I would still clean up that in this new version.

fvtest/porttest/omrvmemTest.cpp
```diff
@@ -388,6 +388,122 @@ TEST(PortVmemTest, vmem_test1)
...
+		getPortLibraryMemoryCategoryData(OMRPORTLIB, &initialBlocks, &initialBytes);
+
+		/* reserve and commit */
+#if defined(J9ZOS390)
+		/* On z/OS skip this test for newly added large pages as obsolete omrvmem_reserve_memory() does not support them */
+		if (isNewPageSize(pageSizes[i], pageFlags[i])) {
+			continue;
+		}
+#endif /* J9ZOS390 */
+		memPtr = (char *)omrvmem_reserve_memory(
```
charliegracie: What happens when a platform does not support `OMRPORT_VMEM_MEMORY_MODE_FILE_HANDLE`?

Right now this will "work" on all platforms but only Linux (or Unix based?) platforms will actually use a backing file. Can we switch the other platforms to return an unsupported return code? And then modify the test to return success if unsupported is returned? I think that would be cleaner.

@youngar thoughts?

charliegracie:
@genie-omr build all

port/linux/omrvmem.c
```diff
@@ -964,49 +965,75 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
-		if (0 != (OMRPORT_VMEM_MEMORY_MODE_COMMIT & mode)) {
-			protectionFlags = get_protectionBits(mode);
-		} else {
-			flags |= MAP_NORESERVE;
+	}
+
+	char filename[FILE_NAME_SIZE];
+	if (useBackingFile) {
+		char int_str[FILE_NAME_SIZE / 2];
+		sprintf(int_str, "%09d", getpid()); // Max pid in 64bit system is 4194304. Pads with zeros.
```
youngar: `sprintf()` appends a NULL byte to the string, so the needed buffer here is `int_str[FILE_NAME_SIZE/2 + 1]`.

port/linux/omrvmem.c
```diff
@@ -964,49 +965,75 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+
+		fd = portLibrary->file_open(portLibrary, filename, EsOpenRead | EsOpenWrite | EsOpenCreate, 0);
+		identifier->fd = fd;
+		if (-1 == fd) {
+			Trc_PRT_vmem_default_reserve_failed(address, byteAmount);
+			update_vmemIdentifier(identifier, NULL, NULL, 0, 0, 0, 0, 0, NULL);
+			Trc_PRT_vmem_default_reserve_exit(result, address, byteAmount);
+			return result;
 		}
+		ftruncate(fd, byteAmount);
```
youngar:
```
./linux/omrvmem.c:1001:12: error: ignoring return value of ‘ftruncate’, declared with attribute warn_unused_result [-Werror=unused-result]
  ftruncate(fd, byteAmount);
```
From the manpage: On success, zero is returned. On error, -1 is returned, and errno is set appropriately.

port/linux/omrvmem.c
```diff
@@ -964,49 +965,75 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+		useBackingFile = TRUE;
 #endif
-	{
-		if (0 != (OMRPORT_VMEM_MEMORY_MODE_COMMIT & mode)) {
-			protectionFlags = get_protectionBits(mode);
-		} else {
-			flags |= MAP_NORESERVE;
+	}
+
+	char filename[FILE_NAME_SIZE];
```
youngar: Did you try using `memfd_create()` instead of opening a named file?  I can't exactly tell what versions of Linux support this call. http://man7.org/linux/man-pages/man2/memfd_create.2.html

bragaigor: I haven't, I'll try that and test to see if it works.

port/linux/omrvmem.c
```diff
@@ -964,49 +965,75 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
-			omrmem_categories_increment_counters(category, byteAmount);
-			if (0 != (OMRPORT_VMEM_MEMORY_MODE_COMMIT & mode)) {
-				if (NULL == omrvmem_commit_memory(portLibrary, result, byteAmount, identifier)) {
-					/* If the commit fails free the memory */
-					omrvmem_free_memory(portLibrary, result, byteAmount, identifier);
-					result = NULL;
-				}
+	if(useBackingFile) {
+		portLibrary->file_close(portLibrary, fd); //Switch to unlink?
+		unlink(filename);
```
youngar: You can `unlink()` the file from the file system right after the file is opened.   Also there is the portlibrary call for this, `portLibrary->file_unlink()`.

fvtest/porttest/omrvmemTest.cpp
youngar: I think it's a good idea.  We already have the error code specified `OMRPORT_ERROR_VMEM_NOT_SUPPORTED` in omrporterror.h:218.

youngar: omrvmem.c:
```
	if(mode & OMRPORT_VMEM_MEMORY_MODE_FILE_HANDLE) {
  		return portLibrary->error_set_last_error(portLibrary,  errno, OMRPORT_ERROR_VMEM_NOT_SUPPORTED);
		return NULL;
	}
```

fvtest/porttest/omrvmemTest.cpp
```diff
@@ -388,6 +388,122 @@ TEST(PortVmemTest, vmem_test1)
...
+						pageSizes[i], OMRMEM_CATEGORY_PORT_LIBRARY);
+
+
+		/* did we get any memory? */
+		if (memPtr == NULL) {
+			lastErrorMessage = (char *)omrerror_last_error_message();
+			lastErrorNumber = omrerror_last_error_number();
+			outputErrorMessage(PORTTEST_ERROR_ARGS, "unable to reserve and commit 0x%zx bytes with page size 0x%zx.\n"
+					"\tlastErrorNumber=%d, lastErrorMessage=%s\n", pageSizes[i], pageSizes[i], lastErrorNumber, lastErrorMessage);
+
```
youngar:
```
if (OMRPORT_ERROR_VMEM_NOT_SUPPORTED == lastErrorNumber) {
 return SUCCESS;
}
```

fvtest/porttest/omrvmemTest.cpp
```diff
@@ -388,6 +388,122 @@ TEST(PortVmemTest, vmem_test1)
...
+						0, pageSizes[i], &vmemID,
+						OMRPORT_VMEM_MEMORY_MODE_READ | OMRPORT_VMEM_MEMORY_MODE_WRITE | OMRPORT_VMEM_MEMORY_MODE_COMMIT | OMRPORT_VMEM_MEMORY_MODE_FILE_HANDLE,
+						pageSizes[i], OMRMEM_CATEGORY_PORT_LIBRARY);
+
+
+		/* did we get any memory? */
+		if (memPtr == NULL) {
+			lastErrorMessage = (char *)omrerror_last_error_message();
+			lastErrorNumber = omrerror_last_error_number();
+			outputErrorMessage(PORTTEST_ERROR_ARGS, "unable to reserve and commit 0x%zx bytes with page size 0x%zx.\n"
```
youngar: before we output the error message, which will cause the test to fail:
```
if (OMRPORT_ERROR_VMEM_NOT_SUPPORTED == lastErrorNumber) {
 goto exit;
}
```

bragaigor:
@genie-omr build all

youngar:
@genie-omr build all

port/win32/omrvmem.c
```diff
@@ -791,6 +791,10 @@ getMemoryInRange(struct OMRPortLibrary *portLibrary, struct J9PortVmemIdentifier
 	static int count = 0;
 #endif
 
+	if(mode & OMRPORT_VMEM_MEMORY_MODE_FILE_HANDLE) {
+		portLibrary->error_set_last_error(portLibrary,  errno, OMRPORT_ERROR_VMEM_NOT_SUPPORTED);
+		return result;
```
bragaigor: Change to `return memoryPointer;`

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

youngar:
@genie-omr build all

charliegracie:
@youngar whenever you are ok with the changes you can merge them.

port/linux/omrvmem.c
```diff
@@ -961,52 +962,79 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+	char filename[FILE_NAME_SIZE];
+	if (useBackingFile) {
+		char int_str[FILE_NAME_SIZE / 2 + 1];
+		sprintf(int_str, "%09d", getpid()); // Max pid in 64bit system is 4194304. Pads with zeros.
+		strcpy(filename, "tempfile");
+		strcat(filename, int_str);
+
+		fd = portLibrary->file_open(portLibrary, filename, EsOpenRead | EsOpenWrite | EsOpenCreate, 0);
+		identifier->fd = fd;
+		omrfile_unlink(portLibrary, filename);
```
charliegracie: I dont think you should be calling the `macro` version here.  This should likely call through the portLibrary variable the same as file_open.

charliegracie:
@genie-omr build all

port/linux/omrvmem.c
```diff
@@ -961,52 +962,79 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+		useBackingFile = TRUE;
 #endif
-	{
-		if (0 != (OMRPORT_VMEM_MEMORY_MODE_COMMIT & mode)) {
-			protectionFlags = get_protectionBits(mode);
-		} else {
-			flags |= MAP_NORESERVE;
+	}
+
+	char filename[FILE_NAME_SIZE];
```
charliegracie: Is this variable used outside of the `if (backingFile)` block below?

bragaigor: No it's not. I think I left outside because I thought it would be used further down. I'll move it inside

port/linux/omrvmem.c
```diff
@@ -961,21 +962,51 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
 #else
-	fd = portLibrary->file_open(portLibrary, "/dev/zero", EsOpenRead | EsOpenWrite, 0);
-	if (-1 != fd)
+		useBackingFile = TRUE;
 #endif
-	{
+	}
+
+	if (useBackingSharedFile) {
+		flags = MAP_SHARED;
```
charliegracie: This has to move below the declarations or up into the other if

fvtest/porttest/omrvmemTest.cpp
```diff
@@ -388,6 +388,127 @@ TEST(PortVmemTest, vmem_test1)
...
+			if (OMRPORT_ERROR_VMEM_INSUFFICENT_RESOURCES == lastErrorNumber) {
+				portTestEnv->log(LEVEL_ERROR, "Portable error OMRPORT_ERROR_VMEM_INSUFFICENT_RESOURCES...\n");
+				portTestEnv->changeIndent(1);
+				portTestEnv->log(LEVEL_ERROR, "REBOOT THE MACHINE to free up resources AND TRY THE TEST AGAIN\n");
+				portTestEnv->changeIndent(-1);
+			}
+			goto exit;
+		} else {
+			/* Need to close the file handle since it was not closed by reserve API */
+			OMRPORTLIB->file_close(OMRPORTLIB, vmemID.fd);
```
charliegracie: This should move into the into omrvmem_free I believe

port/linux/omrvmem.c
```diff
@@ -961,21 +962,51 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+		char filename[FILE_NAME_SIZE + 1];
+		sprintf(filename, "tempfile%09d", getpid()); // Max pid in 64bit system is 4194304.
```
charliegracie: This has to change to be more random so that 2 users inside the same process to not get the same file.

port/linux/omrvmem.c
```diff
@@ -961,21 +966,54 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+
+	if (useBackingSharedFile) {
+		char filename[FILE_NAME_SIZE + 1];
+		sprintf(filename, "omrvmem_temp_%zu_%09d_%zu",  omrtime_current_time_millis(portLibrary), 
+								getpid(), 
+								pthread_self()); 
+		fd = shm_open(filename, O_RDWR | O_CREAT, 0600);
+		identifier->fd = fd;
+		shm_unlink(filename);
+		ft = ftruncate(fd, byteAmount);
```
charliegracie: Is it safe to call `ftruncate` if `fd` == -1?

port/linux/omrvmem.c
```diff
@@ -1005,8 +1043,10 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
 				}
 			}
 		}
+
```
charliegracie: Can we remove the extra blank line?

port/linux/omrvmem.c
```diff
@@ -1005,8 +1043,10 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
 	}
 
+
```
charliegracie: Can we remove the extra blank line?

port/linux/omrvmem.c
bragaigor: The documentation does not say much, but I tested and when we call `ftruncate` and `fd == -1` the call returns -1 which is a fail and sets `errno` to 9 which is `EBADF` (Bad file number)

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build zlinux,plinux

charliegracie:
@genie-omr build all

charliegracie:
The zOS failure is due to infrastructure issues

charliegracie:
@genie-omr build all

gc/base/MemoryManager.cpp
```diff
@@ -95,6 +95,12 @@ MM_MemoryManager::createVirtualMemoryForHeap(MM_EnvironmentBase* env, MM_MemoryH
 		}
 	}
 
+#if defined(LINUX)
+	if(extensions->isVLHGC() && extensions->indexableObjectModel.isDoubleMappingEnabled()) {
```
youngar: I think we need to define `bool isDoubleMappingEnabled()` in [ArrayletObjectModel.hpp](https://github.com/eclipse/omr/blob/f8fc704032f95cbc0f337eb6be54f2b06e58a773/example/glue/ArrayletObjectModel.hpp) 

```c++
	MMINLINE bool
	isDoubleMappingEnabled()
	{
		return false;
	}
```

port/linux/omrvmem.c
```diff
@@ -1009,13 +1049,127 @@ default_pageSize_reserve_memory(struct OMRPortLibrary *portLibrary, void *addres
...
+				printf("Map failed to provide the correct address. nextAddress %p != %p\n", nextAddress, address);
+				fflush(stdout);
+#endif
+				break;
+			}
+		}
+	}
+
+	if (!successfulContiguousMap) {
+		if(shouldUnmapAddr)
```
youngar: Our coding standard says to always put brackets around if statements.

youngar:
@genie-omr build linux

gc/base/VirtualMemory.cpp
```diff
@@ -161,6 +161,16 @@ MM_VirtualMemory::reserveMemory(J9PortVmemParams* params)
...
 }
 
+void*
+MM_VirtualMemory::doubleMapArraylet(MM_EnvironmentBase *env, void* arrayletLeaveAddrs[], UDATA arrayletLeafCount, UDATA arrayletLeafSize, UDATA byteAmount, struct J9PortVmemIdentifier *newIdentifier, UDATA pageSize, OMRMemCategory *category)
+{
+	OMRPORT_ACCESS_FROM_OMRVM(_extensions->getOmrVM());
+	struct J9PortVmemIdentifier *oldIdentifier = &_identifier;
+	uintptr_t mode = OMRPORT_VMEM_MEMORY_MODE_READ | OMRPORT_VMEM_MEMORY_MODE_WRITE | OMRPORT_VMEM_MEMORY_MODE_COMMIT;
+
+	return omrvmem_get_contiguous_region_memory(arrayletLeaveAddrs, arrayletLeafCount, arrayletLeafSize, byteAmount, oldIdentifier, newIdentifier, mode, pageSize, category);
```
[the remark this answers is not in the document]
bragaigor: The being passed in was `OMRMEM_CATEGORY_PORT_LIBRARY`. Should I use `OMRMEM_CATEGORY_MM_RUNTIME_HEAP` instead?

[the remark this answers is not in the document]
youngar: For now add it to OMRMEM_CATEGORY_MM.  I think what we really need is to create a new category of memory and double check how memory is reported in the javacore file.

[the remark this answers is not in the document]
youngar: For now add it to OMRMEM_CATEGORY_MM.  I think what we really need is to create a new category of memory and double check how memory is reported in the javacore file.

youngar:
@genie-omr build all

youngar:
@genie-omr build all

youngar:
@genie-omr build plinux

youngar:
@genie-omr build all

youngar:
@genie-omr build all

Issue #3559, opened 2019-02-12
AppVeyor builds are failling due to an error running cmake


Leonardo2718:
AppVeyor builds seem to be failing consistently with the following output:

```
Build started
Fetching repository commit (e7f9916)...OK
Total: 43.5 MB in 3109 files
Running Install scripts
cinst winflexbison
Chocolatey v0.10.11
Installing the following packages:
winflexbison
By installing you accept licenses for the packages.
Progress: Downloading winflexbison 2.4.9.20170215... 100%
winflexbison v2.4.9.20170215 [Approved]
winflexbison package files install completed. Performing other installation steps.
WARNING: Url has SSL/TLS available, switching to HTTPS for download
Downloading winflexbison 
  from 'https://sourceforge.net/projects/winflexbison/files/old_versions/win_flex_bison-2.4.9.zip'
Progress: 100% - Completed download of C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\winflexbison\2.4.9.20170215\win_flex_bison-2.4.9.zip (683.24 KB).
Download of win_flex_bison-2.4.9.zip (683.24 KB) completed.
Hashes match.
Extracting C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\winflexbison\2.4.9.20170215\win_flex_bison-2.4.9.zip to C:\ProgramData\chocolatey\lib\winflexbison\tools...
C:\ProgramData\chocolatey\lib\winflexbison\tools
 ShimGen has successfully created a shim for win_bison.exe
 ShimGen has successfully created a shim for win_flex.exe
 The install of winflexbison was successful.
  Software installed to 'C:\ProgramData\chocolatey\lib\winflexbison\tools'
Chocolatey installed 1/1 packages. 
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
del "C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.targets\ImportAfter\Xamarin.Common.targets"
set PATH=C:\Python37;%PATH%
cmake -Wdev -G "Visual Studio 11 2012" -DOMR_ENV_DATA32=ON -Ccmake\caches\Appveyor.cmake.
CMake Error: No source or binary directory provided
Command exited with code 1
```

For example: https://ci.appveyor.com/project/eclipsewebmaster/omr

rwy0717:
to me, it looks like we're losing a space before the dot at the end of this line:
```
[...]

PR #3560, opened 2019-02-12
Fix typo in specifying cmake source dir

rwy0717:
As of cmake 13.3, users must specify a source directory. Due to an old
typo in the appvyeyor config, the source directory "./" was attached to
the end of the cache name.

This PR adds back the missing space between the initial cache file and
the source directory.

Fixes: #3559

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,7 +36,7 @@ install:
 before_build:
   - del "C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.targets\ImportAfter\Xamarin.Common.targets"
   - set PATH=C:\Python37;%PATH%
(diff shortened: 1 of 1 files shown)


fjeremic:
The prefixes in the PR titles/commits are not very useful given that we tag the PRs anyway with the `ci` tag. Furthermore the commit message could use some loving [1] [2]:

> The first line describes the change made. It is written in the imperative mood, and should say what happens when the patch is applied. Keep it short and simple. The first line should be less than 70 characters, where reasonable, and **should be written in sentence case** preferably not ending in a period. Leave a blank line between the first line and the message body.

[2] https://chris.beams.io/posts/git-commit/

Leonardo2718:
AppVeyor build passed previously and changes to the commit did not affect the change itself. The change only impacts AppVeyor so no further testing is required. Merging.

PR #3562, opened 2019-02-12
Remove callingMethod from getMethodFromName()

jdmpapin:
`getMethodFromName()` is used for fabricating calls to methods about which the compiler has special knowledge. Since these names originate in the compiler (and not in any particular method we happen to be compiling), the lookups should behave as such, so the `callingMethod` is irrelevant.


0xdaryl:
@genie-omr build all

0xdaryl:
The TravisCI build actually passed, but for some reason its completion status is not being reflected in this PR.  Merging.

PR #3563, opened 2019-02-12
Remove dead evaluators from OMRTreeEvaluator.cpp in Z codegen

fjeremic:
The OMRTreeEvaluator.cpp file is very large, totaling at over 18000
lines. GitHub does not play nice with large files as we hit certain
line limits and size limits [1]. As a first step in cleaning this up we
remove all unused evaluators which we do not foresee using in the
future.

Next step will be to do a big code migration of a ton of OpenJ9
evaluators into the OpenJ9 codebase. This commit however should
alleviate the immediate issue.

[1] https://help.github.com/articles/limits-for-viewing-content-and-diffs-in-a-repository/


0xdaryl:
@genie-omr build zlinux,zos

This is a Z-specific change.  A full genie build is not required.

PR #3564, opened 2019-02-13
WIP : Add a flag to virtual guard performing method test

r30shah:
Add a flag to the virtual guard performing method test at virtual call site.

### DIFF
--- a/compiler/compile/VirtualGuard.hpp
+++ b/compiler/compile/VirtualGuard.hpp
@@ -225,6 +225,8 @@ class TR_VirtualGuard
    TR_VirtualGuardKind     getKind()        { return _kind; }
    TR_OpaqueClassBlock    *getThisClass()   { return _thisClass; }
    void                    setThisClass(TR_OpaqueClassBlock *thisClass)   { _thisClass = thisClass; }
+   void setIsInterfaceMethodGuard(bool b) { _isInterfaceMethodGuard = b; }
+   bool getIsInterfaceMethodGuard() { return _isInterfaceMethodGuard; }
 
    uintptrj_t *mutableCallSiteObject()
       {
@@ -296,6 +298,7 @@ class TR_VirtualGuard
    bool                      _evalChildren;
    bool                      _mergedWithHCRGuard;
    bool                      _mergedWithOSRGuard;
+   bool                      _isInterfaceMethodGuard;
 
    // These reference locations are non-null only for MutableCallSiteGuards
    uintptrj_t                *_mutableCallSiteObject;
--- a/compiler/compile/VirtualGuard.cpp
+++ b/compiler/compile/VirtualGuard.cpp
@@ -306,11 +306,16 @@ TR_VirtualGuard::createMethodGuardWithReceiver
    TR_OpaqueClassBlock *realThisClass = thisClass;
    if (thisClass && TR::Compiler->cls.isInterfaceClass(comp, thisClass) && callSymRef->getSymbol()->castToMethodSymbol()->isInterface())
       realThisClass = calleeSymbol->getResolvedMethod()->classOfMethod();
-
-   if (realThisClass && !TR::Compiler->cls.isInterfaceClass(comp, realThisClass) && callSymRef->getSymbol()->castToMethodSymbol()->isInterface())
+   bool isInterfaceMethodGuard = false;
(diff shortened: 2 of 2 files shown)


compiler/compile/VirtualGuard.hpp
```diff
@@ -296,6 +298,7 @@ class TR_VirtualGuard
    bool                      _evalChildren;
    bool                      _mergedWithHCRGuard;
    bool                      _mergedWithOSRGuard;
+   bool                      _isInterfaceMethodGuard;
```
fjeremic: Why do we need this? Can't we already tell this from `TR_VirtualGuardKind::TR_InterfaceGuard` using the `_kind` field?

r30shah: I think the location where I found the problem, it was generating `TR_ProfiledGuard` 
https://github.com/eclipse/omr/blob/a2cb2ebd66c5996c430a3aa1ba035008ef020c88/compiler/compile/VirtualGuard.hpp#L42
This is where it generates the Virtual Guard : https://github.com/eclipse/omr/blob/a2cb2ebd66c5996c430a3aa1ba035008ef020c88/compiler/compile/VirtualGuard.cpp#L328

I think Virtual Guard kind could be anything (I am not sure that it is guaranteed to be used to generate `TR_ProfiledGuard` kind of guards so added a flag to make sure it has generated a Method Test for the interface. 
This is the type of test it generates. 
https://github.com/eclipse/omr/blob/a2cb2ebd66c5996c430a3aa1ba035008ef020c88/compiler/compile/VirtualGuard.hpp#L65

fjeremic: This seems oddly specific to a single case you're trying to tackle. I'm quite weary of special casing these sorts of things given the precedent it sets. Subscribing @0xdaryl and @andrewcraik for their thoughts. Let's try not to complicate the API more than needed if existing plumbing can handle it.

0xdaryl: Are the VirtualGuard kinds not mutually exclusive then?  i.e., the case you've found is really an InterfaceGuard, but someone created it as a ProfiledGuard.  Will something now break because it was expecting this guard to be a ProfiledGuard?  Are you sure this shouldn't have been created as an interface guard in the first place?

In general, I'm supportive of capability queries rather than identity queries, so I'd like to understand the code that led to this situation.

andrewcraik: so the question of which guard kind you have to worry about comes down to if the guard is nopable or not - if the guard is nop'd then we don't need the range check. If the guard is not nop'd and we are doing a method test (eg an indirection into the vft) then we need to essentially bound check that access. Now we can omit the bound check for virtual dispatches because we know the check will always pass so the question is can we tell if the original call was on an interface - I need to check the inlining table to see if we store the call node symref or the callnode so we can get it later - I suspect the answer is yes, but I need to find it.

Assuming that information is available from the inlining table then we should not have a bool flag on the guard and we should base the need for the bound check on the nopability, the guard kind and and the callsite's dispatch kind (eg was it an invokeinterface).

r30shah: Thanks a lot @andrewcraik  for the clarification. Apart from information stored in inlined table, is it possible to use information on VirtualGuard? 
I am looking into VirtualGuard constructor, I see that we store the SymbolReference of the callNode (If it was passed that callNode - Which should be for the method test), I should be able to take advantage of this information in https://github.com/eclipse/openj9/pull/4721 where I actually use this information and we can avoid the bool on virtual guard.

andrewcraik: You can use information on the guard - I believe the symref should be reliable for inline method guards which are the kind you are concerned with. You can add a fatal assert so that if, in testing, there is ever a case where we don't have the symref you fail early and violently.

andrewcraik: Assuming this all works out as I expect you likely won't need an OMR change to allow you to capture the information you need to make the fix you need to make in OpenJ9 - I think there should already be enough data.

r30shah:
I have made the Changes in the OpenJ9 code to identify Virtual Guard performing MethodTest  is guard for inlined method of interface call. https://github.com/eclipse/openj9/pull/4721 
Closing this PR as we do not need it.

PR #3565, opened 2019-02-13
Transpose parameters to findHelperTrampoline

0xdaryl:
It is more readable and intuitive if the helper you're asking about
is the first parameter.  Fixup uses of this API.

The original API remains in the code and simply redirects to the new
API, and will be removed once downstream projects have been changed.


0xdaryl:
@genie-omr build all

PR #3568, opened 2019-02-13
Migrate randomgen architecture disabling from OpenJ9 to OMR

fjeremic:
Issue: eclipse/openj9#3576

### DIFF
--- a/compiler/z/codegen/OMRCodeGenerator.cpp
+++ b/compiler/z/codegen/OMRCodeGenerator.cpp
@@ -583,6 +583,48 @@ OMR::Z::CodeGenerator::CodeGenerator()
       _processorInfo.disableArch(TR_S390ProcessorInfo::TR_zNext);
       }
 
+   // Randomly disable an architecture if we are in randomGen mode
+   if (comp->getOption(TR_Randomize))
+      {
+      switch (randomizer.randomInt(TR::Compiler->target.cpu.id() - TR_s370gp7))
+         {
+         case 1:
+            {
+            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z196);
+            traceMsg(comp, "RandomGen: Disabling z196 processor architecture.");
+            break;
+            }
+
+         case 2:
+            {
+            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_zEC12);
+            traceMsg(comp, "RandomGen: Disabling zEC12 processor architecture.");
+            break;
+            }
+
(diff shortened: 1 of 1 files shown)


fjeremic:
@genie-omr build zlinux,zos

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -583,6 +583,48 @@ OMR::Z::CodeGenerator::CodeGenerator()
...
+            }
+
+         case 3:
+            {
+            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z13);
+            traceMsg(comp, "RandomGen: Disabling z13 processor architecture.");
+            break;
+            }
+
+         case 3:
```
Leonardo2718: I think this is supposed to be:
```suggestion
         case 4:
```

compiler/z/codegen/OMRCodeGenerator.cpp
```diff
@@ -583,6 +583,48 @@ OMR::Z::CodeGenerator::CodeGenerator()
...
+            {
+            _processorInfo.disableArch(TR_S390ProcessorInfo::TR_z14);
+            traceMsg(comp, "RandomGen: Disabling z14 processor architecture.");
+            break;
+            }
+
+         case 4:
```
Leonardo2718: And here
```suggestion
         case 5:
```

compiler/z/codegen/OMRCodeGenerator.cpp
Leonardo2718: Unless the cases were supposed to start at 0...

fjeremic:
Thanks for catching that @Leonardo2718. I might add at his point that it is unfortunate that most of the processor detection currently sits in OpenJ9, unrightfully so. I'm slowly working on pushing this code down into OMR.

[1] https://github.com/eclipse/openj9/blob/35c44ead49a53f8c0ad1ca3e63c18711d3a7b26f/runtime/compiler/env/ProcessorDetection.cpp#L854-L857

Leonardo2718:
@genie-omr build zlinux,zos

Leonardo2718:
> I might add at his point that it is unfortunate that most of the processor detection currently sits in OpenJ9, unrightfully so. I'm slowly working on pushing this code down into OMR.

Glad to hear your working on moving that code to OMR :smile: . Not having access to it is a major drawback right now.

PR #3569, opened 2019-02-13
Update paths to tracegen'd files

dnakamura:
Specify full path to the files so that they are handled properly by ddr


charliegracie:
@genie-omr build all

PR #3570, opened 2019-02-14
Provide explicit cast for findHelperTrampoline

0xdaryl:
The cast is required in order to change the return type of
`findHelperTrampoline` to a `intptrj_t` to allow downstream
projects to continue to build without breaking.

### DIFF
--- a/compiler/runtime/OMRCodeCacheManager.cpp
+++ b/compiler/runtime/OMRCodeCacheManager.cpp
@@ -518,7 +518,7 @@ OMR::CodeCacheManager::findHelperTrampoline(int32_t helperIndex, void *callSite)
 OMR::CodeCacheTrampolineCode *
 OMR::CodeCacheManager::findHelperTrampoline(void *callingPC, int32_t helperIndex)
    {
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build all

compiler/runtime/OMRCodeCacheManager.cpp
```diff
@@ -518,7 +518,7 @@ OMR::CodeCacheManager::findHelperTrampoline(int32_t helperIndex, void *callSite)
 OMR::CodeCacheTrampolineCode *
 OMR::CodeCacheManager::findHelperTrampoline(void *callingPC, int32_t helperIndex)
    {
-   return self()->findHelperTrampoline(helperIndex, callingPC);
+   return (OMR::CodeCacheTrampolineCode *)self()->findHelperTrampoline(helperIndex, callingPC);
```
Leonardo2718: I think I would prefer a C++ style cast here.

0xdaryl: Fair enough.  Fixed.

0xdaryl:
@genie-omr build all

PR #3571, opened 2019-02-14
Refactor rdbar and wrtbar evaluators into the common codegen

dchopra001:
This PR does the following:

Implement rdbar/wrtbar evaluators in the Z codegen.
Separate indirect and direct rdbar/wrtbar evaluators.
Refactor all applicable rdbar/wrtbar evaluators into the common codegen so that they can be shared across all codgens.


Leonardo2718:
@genie-omr build zlinux, zos

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -6495,6 +6495,112 @@ OMR::Z::TreeEvaluator::bstoreEvaluator(TR::Node * node, TR::CodeGenerator * cg)
    return NULL;
    }
 
+TR::Register *
+OMR::Z::TreeEvaluator::irdbarEvaluator(TR::Node *node, TR::CodeGenerator *cg)
+   {
+   if (node->getSymbolReference()->getSymbol()->isStatic())
+      cg->decReferenceCount(node->getFirstChild());
+   return TR::TreeEvaluator::iloadEvaluator(node,cg);
+   }
```
fjeremic: It appears that now all codegens (bar ARM) implement these evaluators in the exact same way. Would it be possible to move these evaluators up to the common code generator and we all use the same common implementation?

fjeremic: Also space needed between the two arguments `node,cg`

fjeremic:
@dchopra001 can we squash this and resolve the conflicts?

compiler/z/codegen/OMRTreeEvaluator.cpp
dchopra001: This should no longer be an issue in the latest version.

compiler/codegen/OMRTreeEvaluator.cpp
```diff
@@ -51,6 +51,418 @@
...
+   cg->evaluate(sideEffectNode);
+
+   // Note: We only decrement the reference count for the side effect node. This is because
+   // the store evaluator is not aware of wrtbar's side effect node and so will not evaluate/decrement the 
+   // reference count for it. However it is aware and will do this work for the other children. So we skip
+   // decrementing the reference count for the other children here and let the store evaluator handle it in
+   // order to avoid double decrementing the reference count.
+   cg->decReferenceCount(sideEffectNode);
+
+   return TR::TreeEvaluator::bstoreEvaluator(node, cg);
```
fjeremic: The write barrier evaluators are too aggressive. They should only be evaluating nodes which they are handling. In this case the side effect node is the only one we should be evaluating. The delegated store evaluator will take care of evaluating the store portion of the write barrier. These evaluators should look like this:

```
TR::Register *
OMR::TreeEvaluator::bwrtbariEvaluator(TR::Node *node, TR::CodeGenerator *cg)
   {
   // A bwrtbari IL has a side effect node which is currently not used at the OMR level. We first handle the side
   // effect, then call the appropriate store evaluator to properly evaluate the opcode.
   TR::Node *sideEffectNode = node->getThirdChild();
   cg->evaluate(sideEffectNode);

   // Note: We only decrement the reference count for the side effect node. This is because the store evaluator is not
   // aware of wrtbar's side effect node and so will not evaluate/decrement the reference count for it. However it is
   // aware and will do this work for the other children. So we skip decrementing the reference count for the other
   // children here and let the store evaluator handle it in order to avoid double decrementing the reference count.
   cg->decReferenceCount(sideEffectNode);

   return TR::TreeEvaluator::bstoreEvaluator(node, cg);
   }
```

fjeremic: Similarly for all other write barrier evaluators.

compiler/codegen/OMRTreeEvaluator.cpp
```diff
@@ -51,6 +51,418 @@
...
+   // the side effect, then call the appropriate load evaluator to properly evaluate the opcode.
+
+   // Currently brdbar nodes don't have any side effects and are effectively just a load operation. 
+   // However we must evaluate brdbar's child here since it is the responsibility of the evaluator
+   // to evaluate all of its children. Additionally, the bloadEvaluator is not aware of brdbar's child
+   // and will not use it. It is only needed for the side effect (i.e. by this evaluator).
+   TR::Node *firstChild = node->getFirstChild();
+   cg->evaluate(firstChild);
+   cg->decReferenceCount(firstChild);
+   return TR::TreeEvaluator::bloadEvaluator(node, cg);
```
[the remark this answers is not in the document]
fjeremic: Similarly for all other read barrier evaluators.

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build all

PR #3572, opened 2019-02-14
Fix ddrgen parsing signed data attributes on OSX

mikezhang1234567890:
Parser was writing signed data to udata field instead of sdata.


charliegracie:
@genie-omr build all

PR #3574, opened 2019-02-14
Remove TR::X86UnresolvedVirtualCallSnippet

jdmpapin:
...and the corresponding `IsUnresolvedVirtualCall` enumerator. This OpenJ9-specific type is unused. The declaration of its `TR_Debug::print` method is retained for the moment to avoid breaking builds.


0xdaryl:
@genie-omr build all

jdmpapin:
Updated the copyright header

PR #3583, opened 2019-02-19
Update genie-omr build targets in committer guide

0xdaryl:
Document all the genie build targets.

[skip ci]

### DIFF
--- a/doc/GuideForCommitters.md
+++ b/doc/GuideForCommitters.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2018, 2018 IBM Corp. and others
+Copyright (c) 2018, 2019 IBM Corp. and others
 
 This program and the accompanying materials are made available under
 the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


doc/GuideForCommitters.md
```diff
@@ -88,9 +88,24 @@ If the change affects multiple platforms or 64-bit Windows, you must initiate a
...
+   @genie-omr build [ all | {platform-list} ]
    ```
-   Combinations of platforms can be specified separated by commas.
+Where `{platform-list}` is a comma-separated list of platforms chosen from the
+following tags:
+
+| Tag      | Platform |
+| :----    | :------- |
+| zlinux   | 64-bit Linux on Z |
+| zos      | 64-bit zOS |
```
fjeremic: Interesting. So we have no 32-bit Z targets. Anyone have an idea if this is hard to setup? I woudln't mind running all the Tril tests on 32-bit Linux on Z for example as we need coverage there.

charliegracie: Interesting.  I had not noticed there was not 32 bit zOS. This is something we can likely change. The zOS build is still using the autoconf and JIT static makefiles so it should be relatively easy to switch.

0xdaryl: I created #3591 to track this.

PR #3585, opened 2019-02-19
Fix missing omrVM in example ScavengerDelegate

rwy0717:
This PR fixes a typo where the _omrVM member was missing from the
delegate. This error made it through because the delegate hasn't been
enabled alongside compressed references & compressed object header.

The omrVM used to be cached in the CLI. In the ScavengerDelegate, only
keep the extensions, and just get the VM out of the extensions when
needed.


youngar:
@genie-omr build all

PR #3586, opened 2019-02-20
AArch64: Use cg->directCallRequiresTrampoline() for checking branch range

knn-k:
This is a follow-up of #3479 for aarch64.
Two functions in ARM64BinaryEncoding.cpp to use
cg->directCallRequiresTrampoline() for checking the range of branch
instructions.


0xdaryl:
@genie-omr build aarch64

PR #3587, opened 2019-02-20
Move SwitchAnalyzer from OpenJ9 to OMR

charliegracie:
Recent testing using JB Switch did not show the performance I was
expecting. Moving the SwitchAnalyzer from OpenJ9 to OMR provides
a 25-30% improvement to my testing.

### DIFF
--- a/compiler/optimizer/CMakeLists.txt
+++ b/compiler/optimizer/CMakeLists.txt
@@ -85,6 +85,7 @@ compiler_library(optimizer
 	${CMAKE_CURRENT_LIST_DIR}/OMRSimplifierHandlers.cpp
 	${CMAKE_CURRENT_LIST_DIR}/StructuralAnalysis.cpp
 	${CMAKE_CURRENT_LIST_DIR}/Structure.cpp
+	${CMAKE_CURRENT_LIST_DIR}/SwitchAnalyzer.cpp
 	${CMAKE_CURRENT_LIST_DIR}/TranslateTable.cpp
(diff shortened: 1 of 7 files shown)


charliegracie:
@0xdaryl this PR likely needs changes to ensure it happens correctly.  This was the simple change that got my performance improvement.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@andrewcraik a quick review please since this is in the optimizer.

andrewcraik:
An init of the optimization in the OMROptimizer.cpp would be standard for an optimization living at the OMR level. Normally you would need the opt enum to move as well but the enum seems to already be in OMR for whatever reason. Otherwise seems like a standard move of code upwards.

Moving this code up may require changes in OpenJ9 to keep things compatible.

charliegracie:
I believe I updated all of @andrewcraik comments. I will try to check out whether or not downstream projects need a change.  I do not think any down stream projects use CMake for their builds so they will not use the new copy of SwitchAnalyzer. I believe the OpenJ9 project does do a single build that builds using CMake so I may just have to add this file to it's blacklist.

charliegracie:
@genie-omr build all

andrewcraik:
Rather than black listing the OMR version wouldn't it be more correct to switch it to consume the OMR version? I guess you mean you will blacklist, then this will go in and then you will do the switch?

charliegracie:
That is what I was thinking so that the changes could flow independently.

PR #3589, opened 2019-02-20
Improve assertion output for large object stats

dmitripivkine:
Log() calculation can be broken in OS or output of the function might be
corrupted. This change improves assertions output.


dmitripivkine:
@amicic Would you please review?

charliegracie:
@genie-omr build all

PR #3594, opened 2019-02-21
WIP: Return a reference to a dummy address variable

rwy0717:
This code is unimplemented, and should crash before trying to return a
dummy result. Unfortunately, the dummy result is created by
dereferencing a null pointer, and clang will still emit a warning that
this is undefined behaviour. This patch causes the function to return a
valid reference to a global dummy value, which silences the error.


rwy0717:
I'm closing this PR, because I think the better solution is to insert an unconditional fatal assertion failure. The compiler won't require a return result if the function doesn't return.

PR #3595, opened 2019-02-21
Explicitly widen the source immediate before casting to pointer

rwy0717:
This patch fixes a pile of bad casts where we're converting from a 32bit
integer to a 64bit pointer. The warnings are silenced by performing an
intermediate cast to a uintptr_t, before casting to a pointer.


charliegracie:
@genie-omr build xcrlinux xlinux win osx

Issue #1699, opened 2017-09-20
Multiple TR_ALLOC definition warning


0xdaryl:
Building Tril compiler tests (e.g., on x86-64) yields the following warning:

```
In file included from /home/travis/build/eclipse/omr/compiler/il/OMRNode.hpp:40:0,
                 from /home/travis/build/eclipse/omr/compiler/il/Node.hpp:25,
                 from /home/travis/build/eclipse/omr/fvtest/compilertriltest/SimplifierFoldAndTest.cpp:24:
/home/travis/build/eclipse/omr/compiler/env/TRMemory.hpp:707:0: warning: "TR_ALLOC" redefined [enabled by default]
 #define TR_ALLOC(a) \
 ^
In file included from /home/travis/build/eclipse/omr/compiler/ilgen/IlBuilder.hpp:33:0,
                 from /home/travis/build/eclipse/omr/compiler/ilgen/TypeDictionary.hpp:38,
                 from /home/travis/build/eclipse/omr/fvtest/tril/tril/ilgen.hpp:25,
                 from /home/travis/build/eclipse/omr/fvtest/tril/tril/method_info.hpp:26,
                 from /home/travis/build/eclipse/omr/fvtest/tril/tril/method_compiler.hpp:25,
                 from /home/travis/build/eclipse/omr/fvtest/tril/tril/jitbuilder_compiler.hpp:25,
                 from /home/travis/build/eclipse/omr/fvtest/compilertriltest/SimplifierFoldAndTest.cpp:23:
/home/travis/build/eclipse/omr/compiler/ilgen/IlInjector.hpp:56:0: note: this is the location of the previous definition
 #define TR_ALLOC(x)
 ^
```

Although `TRMemory.hpp` itself is guarded for multiple includes, the macros defined within it are not.  I'm not convinced that protecting those macros from multiple inclusion is the right answer yet.  This will need some investigation.

Leonardo2718:
Actually, I believe this happens when building JitBuilder in general, not just Tril.

mstoodle:
Yeah, it's because these files are also used independently of the JIT (i.e. in jitbuilder/release/include). Rather than slurping in the entire TR_Memory set of headers, I just provided this escape hatch to define the macro (since it isn't neeed in the other scenario). I could probably have wrapped in a #ifndef block :) .

0xdaryl:
This is fixed by #3596.

PR #3596, opened 2019-02-21
Define TR_ALLOC everywhere

rwy0717:
With the JitBuilder public API in place, env/TRMemory.hpp is now available everywhere, and 
we can fully define the TR_ALLOC macro everywhere.

As a work around in the internals of the JitBuilder API, we have to use the global new operator to allocate internal compiler types.

### DIFF
--- a/compiler/ilgen/IlInjector.hpp
+++ b/compiler/ilgen/IlInjector.hpp
@@ -30,6 +30,7 @@
 
 #include <stdint.h>
 #include "env/jittypes.h"
+#include "env/TRMemory.hpp"
 #include "il/ILOpCodes.hpp"
 #include "ilgen/IlGen.hpp"
 
@@ -50,11 +51,6 @@ namespace TR { class SymbolReferenceTable; }
 namespace TR { class TreeTop; }
 namespace TR { class IlType; }
 namespace TR { class TypeDictionary; }
-
-// This macro reduces dependencies for this header file
-#ifndef TR_ALLOC
-#define TR_ALLOC(x)
-#endif
 
 namespace OMR
 {
--- a/compiler/ilgen/OMRIlType.hpp
+++ b/compiler/ilgen/OMRIlType.hpp
@@ -22,13 +22,8 @@
 #ifndef OMR_ILTYPE_INCL
 #define OMR_ILTYPE_INCL
 
+#include "env/TRMemory.hpp"
 #include "il/DataTypes.hpp"
-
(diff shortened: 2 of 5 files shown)


0xdaryl:
This is #1699 I believe.

0xdaryl:
^^^ Uh, I think.  @rwy0717, correct me if I'm wrong.

rwy0717:
Yep, I think that would be the same issue :+1:

mstoodle:
Now that we no longer use the header files from the OMR compiler in the JitBuilder client API, I would have thought a better solution would be to get rid of this empty `#define` altogether...

rwy0717:
I can do that, if you want--I figured it was still important for one reason or another. Just double checking... does that mean that each place where we have the empty define, we can instead include `TRMemory.hpp`?

mstoodle:
`env/TRMemory.hpp` yes

charliegracie:
@rwy0717 any updates?

rwy0717:
When we fully define  `TR_ALLOC` everywhere, it causes compilation errors in the generated jitbuilder API code (which I'm going to call jitbuilder-api).

The Problem: jitbuilder-api uses new-expressions to allocate compiler classes, but those compiler classes use `TR_ALLOC` to define custom new operators. The issue is, these user-defined new operators hide the normal new operator for those types. This means that the types must be allocated through TR::Memory, and the basic new expressions will fail to compile. Since `TR::Memory` is internal, it's actually impossible for jitbuilder-api to allocate those types correctly.

I'm hoping we can sidestep this restriction by calling the global new operator explicitly with the syntax `::new TR::TypeDictionary()`. This would only be a band-aid solution, and I haven't actually tried it.

The long term solution may be to allow consumers to allocate transient memory associated with a compilation through some restricted, user-facing API. I'm betting @mstoodle has some ideas on how this could work.

fjeremic:
@rwy0717 what state is this PR in? Is this safe to merge as the current workaround? We have some work in #3693 depending on this change.

fjeremic:
@rwy0717 what is blocking us to progress this PR? I'm set to get CMake up and running on our z/OS farm machines today and this PR is blocking the progress of #3693.

rwy0717:
@Leonardo2718 can you take a look at this?

Leonardo2718:
@genie-omr build all

fjeremic:
Thanks @rwy0717! Much appreciated for fixing this.

Issue #1720, opened 2017-09-22
Refactor assertions handling


mgaudet:
When compiling with clang, you'll see 

```
[13/262] Building CXX object jitbuilder/CMakeFiles/jitbuilder.dir/__/compiler/infra/Assert.cpp.o
../compiler/infra/Assert.cpp:171:7: warning: function declared 'noreturn' should not return [-Winvalid-noreturn]
      }
     
```

This is because the OMR_NORETURN annotation is applied to the fatal assertion, however, the function it calls is not also annotated with OMR_NORETURN, because it's implementation is shared with non-fatal assertions.

Correcting this is best done by refactoring the contents of Assert.cpp to keep the fatal and non-fatal portions better isolated such that the annotations are correct. The annotation is intended to make static analysis of the codebase more accurate. 

(This is part of #1640)

0xdaryl:
Fixed in #3597.

PR #3597, opened 2019-02-21
 Make fatal_assertion actually "no return"

rwy0717:
The function which actually triggers a trap, `assumeDontCallMeDirectlyImpl`, cannot be marked as no-return, because it only traps when handling fatal assertions.

Since fatal_assertion doesn't call any no-return functions, it cannot be marked as no-return.

Remove the OMR_NORETURN attribute from fatal_assertion and silence a compiler warning.

### DIFF
--- a/compiler/infra/Assert.cpp
+++ b/compiler/infra/Assert.cpp
@@ -75,20 +75,12 @@ void OMR_NORETURN TR::trap()
       }
    }
 
-static void assumeDontCallMeDirectlyImpl( const char * file, int32_t line, const char *condition, const char *s, va_list ap, bool fatal)
+
+static void traceAssertionFailure(const char * file, int32_t line, const char *condition, const char *s, va_list ap)
    {
    TR::Compilation *comp = TR::comp();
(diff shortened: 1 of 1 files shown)


rwy0717:
I can refactor the assertion code more aggressively, so that `fatal_assertion` actually can be marked no_return. The change would involve invoking `TR::Trap`directly from both `assertion` and `fatal_assertion`, so `assumeDontCallMeDirectly` just does logging. I think that would be the preferable solution. I'm going to push this version in a moment.

rwy0717:
Requires #3613

0xdaryl:
I believe this addresses #1720.

rwy0717:
This is ready for a review.

0xdaryl:
@genie-omr build all

0xdaryl:
Looks reasonable to me.  @Leonardo2718, would you mind giving this a once-over please?

0xdaryl:
@genie-omr build plinux,xlinux

PR #3598, opened 2019-02-21
Make mcc_printf a variadic function macro

rwy0717:
Make the mcc_printf macro an inline function

mcc_printf is a variable-like macro which aliases the token printf.
However, when CODECACHE_DEBUG is disabled, mcc_printf is defined to
nothing. When this happens, the arguments at each "call site" will stick
around as a parenthesized expression. These unused expressions generate
compiler warnings.

So a call like:
    mcc_printf("hello %s\n", my_name);

is transformed to the expression:
    ("hello %s\n", my_name);

This PR makes mcc_printf a variadic function-like macro that calls
printf when CODECACHE_DEBUG is on, and expanding to nothing otherwise.
By expanding the entire call to nothing, we can ensure that there are no
leftover artifacts of the disabled code.

This PR also drops the additional test against the variable
CodeCacheDebug, which isn't defined anywhere. Another define,
mcc_hashprintf, was deleted, because it simply doesn't appear in any
codebase.

### DIFF
--- a/compiler/runtime/CodeCacheTypes.hpp
+++ b/compiler/runtime/CodeCacheTypes.hpp
@@ -26,19 +26,23 @@
 #define __TPF_DO_NOT_MAP_ATOE_REMOVE
 #endif
 
+#undef CODECACHE_DEBUG
+
 #include <stddef.h>
 #include <stdint.h>
+#if defined(CODECACHE_DEBUG)
+#include <stdio.h>
+#endif /* CODECACHE_DEBUG */
+
 #include "runtime/MethodExceptionData.hpp"
 
 /*
  *  Debugging help
  */
-#ifdef CODECACHE_DEBUG
-#define mcc_printf if (CodeCacheDebug) printf
-#define mcc_hashprintf /*printf*/
+#if defined(CODECACHE_DEBUG)
+#define mcc_printf(fmt, ...) printf(fmt, ##__VA_ARGS__)
 #else
-#define mcc_printf
-#define mcc_hashprintf
+#define mcc_printf(fmt, ...) ((void)0)
 #endif
 
 class TR_OpaqueMethodBlock;
--- a/compiler/runtime/CodeCacheTypes.cpp
+++ b/compiler/runtime/CodeCacheTypes.cpp
@@ -37,7 +37,7 @@ CodeCacheHashTable::dumpHashUnresolvedMethod(void)
       {
       entry = _buckets[i];
       if (entry)
-         mcc_printf("index = %d, constPool = 0x%x  cpIndex = %d\n", i, entry->_info._unresolved._constPool,entry->_info._unresolved._constPoolIndex);
+         mcc_printf("index = %d, constPool = 0x%p  cpIndex = %d\n", i, entry->_info._unresolved._constPool,entry->_info._unresolved._constPoolIndex);
       }
    return;
    }
--- a/compiler/runtime/OMRCodeCacheManager.cpp
+++ b/compiler/runtime/OMRCodeCacheManager.cpp
@@ -37,6 +37,7 @@
 #include "infra/Assert.hpp"
 #include "infra/CriticalSection.hpp"
 #include "infra/Monitor.hpp"
+#include "omrformatconsts.h"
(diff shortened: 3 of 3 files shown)


Leonardo2718:
> The better solution would have been to use a variadic macro to delete the entire call AND arguments, but not all compilers support VA_ARGS.

Actually, I believe `##__VA_ARGS__` is available on all the build compilers OMR supports because it's used in the compiler (e.g. see definition of the `TR_ASSERT` macro). :slightly_smiling_face: 

https://github.com/eclipse/omr/blob/1537cc08bd468d71d214ad621a2d7d6df3b4445c/compiler/infra/Assert.hpp#L100-L102

rwy0717:
D'oh! You're right, it's everywhere in the compiler 👍 I'm going to switch to a macro, then.

rwy0717:
Updated.

Another detail: the macro no longer checks the variable `CodeCacheDebug`, which isn't defined anywhere.

rwy0717:
rebased, and moved mcc_printf's format arg into the varargs. This should deal with the case where varargs are empty.

```diff
- #define mcc_printf(format, ...) printf(format, __VA__ARGS___)
+ #define mcc_printf(...) printf(__VA__ARGS___)
```

compiler/runtime/CodeCacheTypes.hpp
```diff
@@ -26,19 +26,23 @@
...
 #include "runtime/MethodExceptionData.hpp"
 
 /*
  *  Debugging help
  */
-#ifdef CODECACHE_DEBUG
-#define mcc_printf if (CodeCacheDebug) printf
-#define mcc_hashprintf /*printf*/
+#if defined(CODECACHE_DEBUG)
+#define mcc_printf(...) printf(__VA_ARGS__)
```
Leonardo2718: Hm... so in other parts of the compiler, we use `##__VA_ARGS__` (with the leading `##`). I'm not sure if that matters. I know the first is technically a non-standard compiler extensions but it seems to work on all our build compilers.

0xdaryl:
@genie-omr build all

0xdaryl:
@rwy0717, did you do a build with `CODECACHE_DEBUG` enabled to verify the output produced from `mcc_printf` is what you expect it to be?

rwy0717:
I did turn the flag on, but only to verify that the code still compiles. I did see lots of new printf-formatting warnings :) Is there anything I can run to verify that the output actually happens?

compiler/runtime/CodeCacheTypes.hpp
rwy0717: using VA_ARGS is probably cleaner, and since it is supported, I'll push a version that uses it.

rwy0717:
I've switched to use `##__VA_ARGS__` rather than just `__VA_ARGS__`. I've also corrected some format string errors that were hidden because the mcc_printf macro was disabled.

Leonardo2718:
@genie-omr build all

Leonardo2718:
All tests have now passed. For completeness, I will point out that msvc is now warning about "empty controlled statements" (an extraneous semicolon; see see examples in [1] and [2]). @rwy0717 do you think this is worth fixing in this PR? The warning is:

```
C:\projects\omr\compiler\runtime\CodeCacheTypes.cpp(41): warning C4390: ';' : empty controlled
statement found;
```

[1] https://ci.appveyor.com/project/eclipsewebmaster/omr/builds/23553095?fullLog=true
[2] https://ci.eclipse.org/omr/job/PullRequest-win_x86-64/1010/consoleFull

rwy0717:
Maybe we can turn mcc_printf into an inline function:

```c++
void mccprintf(const char* fmt, ...) {
#if defined CODECACH_DEBUG
  va_args args;
  va_start(args, fmt);
  vfprintf(stderr, fmt, args);
  va_end(args);
#endif
}
```

Leonardo2718:
I think defining `mcc_printf` to `(void)0` _might_ also work.

rwy0717:
Lets try it. I just pushed a new version with `((void)0)`

0xdaryl:
@genie-omr build all

PR #3599, opened 2019-02-21
Clean up c-style formatting

rwy0717:
Introduce a new header, `omrformatconsts.h`, which defines a set of portable constants for building format strings for special integral types. This header  provides platform-independent format specifiers for the following data types:

- int8_t / uint8_t
- int16_t / uint16_t
- int32_t / uint32_t
- int64 / uint64_t
- size_t
- intptr_t / uintptr_t
- ptrdiff_t

The following formats are provided in this PR:
- x: unsigned hexadecimal format
- u: unsigned decimal format
- d: signed decimal format

(notably missing: i)

Normally, there are either special macros defined by the C library, or
special format characters in the case of size_t / ptrdiff_t.
Unfortunately, because of poor standards compliance across our
compilers, these aren't universally available.

The names of the constants follow the convention of those normally
provided by stdint.h.

As well, I've gone through and fixed a number of locations with format strings that generate errors.

### DIFF
--- a/compiler/control/CompileMethod.cpp
+++ b/compiler/control/CompileMethod.cpp
@@ -56,6 +56,7 @@
 #include "ras/Debug.hpp"
 #include "env/SystemSegmentProvider.hpp"
 #include "env/DebugSegmentProvider.hpp"
+#include "omrformatconsts.h"
 #include "runtime/CodeCacheManager.hpp"
 
 #if defined (_MSC_VER) && _MSC_VER < 1900
@@ -80,7 +81,7 @@ writePerfToolEntry(void *start, uint32_t size, const char *name)
       static const int maxPerfFilenameSize = 15 + sizeof(jvmPid)* 3; // "/tmp/perf-%ld.map"
       char perfFilename[maxPerfFilenameSize] = { 0 };
 
-      int numCharsWritten = snprintf(perfFilename, maxPerfFilenameSize, "/tmp/perf-%ld.map", jvmPid);
+      int numCharsWritten = snprintf(perfFilename, maxPerfFilenameSize, "/tmp/perf-%" OMR_PRId64 ".map", static_cast<int64_t>(jvmPid));
       if (numCharsWritten > 0 && numCharsWritten < maxPerfFilenameSize)
          {
          perfFile = fopen(perfFilename, "a");
--- a/compiler/runtime/OMRCodeCache.cpp
+++ b/compiler/runtime/OMRCodeCache.cpp
@@ -37,6 +37,7 @@
 #include "infra/Assert.hpp"
 #include "infra/CriticalSection.hpp"
 #include "infra/Monitor.hpp"
+#include "omrformatconsts.h"
 #include "runtime/CodeCache.hpp"
 #include "runtime/CodeCacheManager.hpp"
 #include "runtime/CodeCacheMemorySegment.hpp"
@@ -1275,8 +1276,8 @@ void
 OMR::CodeCache::dumpCodeCache()
    {
    printf("Code Cache @%p\n", this);
-   printf("  |-- warmCodeAlloc          = 0x%08x\n", _warmCodeAlloc );
-   printf("  |-- coldCodeAlloc          = 0x%08x\n", _coldCodeAlloc );
+   printf("  |-- warmCodeAlloc          = 0x%08" OMR_PRIxPTR "\n", (uintptr_t)_warmCodeAlloc );
+   printf("  |-- coldCodeAlloc          = 0x%08" OMR_PRIxPTR "\n", (uintptr_t)_coldCodeAlloc );
    printf("  |-- tempTrampsMax          = %d\n",     _tempTrampolinesMax );
    printf("  |-- flags                  = %d\n",     _flags );
    printf("  |-- next                   = 0x%p\n",   _next );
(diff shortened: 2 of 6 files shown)


compiler/control/CompileMethod.cpp
```diff
@@ -80,7 +80,7 @@ writePerfToolEntry(void *start, uint32_t size, const char *name)
       static const int maxPerfFilenameSize = 15 + sizeof(jvmPid)* 3; // "/tmp/perf-%ld.map"
       char perfFilename[maxPerfFilenameSize] = { 0 };
 
-      int numCharsWritten = snprintf(perfFilename, maxPerfFilenameSize, "/tmp/perf-%ld.map", jvmPid);
+      int numCharsWritten = snprintf(perfFilename, maxPerfFilenameSize, "/tmp/perf-%ld.map", (long)jvmPid);
```
Leonardo2718: In the compiler code, we prefer using `int*_t` rather than `int`, `long`, etc. Would it be possible to instead have:

```suggestion
      int numCharsWritten = snprintf(perfFilename, maxPerfFilenameSize, "/tmp/perf-%" OMR_PRId64 ".map", (int64_t)jvmPid);
```

?

fvtest/compilertriltest/BitPermuteTest.cpp
```diff
@@ -146,8 +147,8 @@ TEST_P(lBitPermuteTest, ConstAddressLengthTest)
       "  (lreturn                         "
       "  (lbitpermute                     "
       "   (lload parm=0)                  "
-      "   (aconst %llu)                   "
-      "   (iconst %llu)                   "
+      "   (aconst %p)                     "
```
Leonardo2718: Why is `%p` used here instead of `OMR_PRIuPTR` with a cast, as is done bellow?

include_core/omrformatconsts.h
```diff
@@ -0,0 +1,182 @@
+#ifndef OMRFORMATCONSTANTS_H_
```
Leonardo2718: This file needs a copyright header. :slightly_smiling_face:

fvtest/compilertriltest/BitPermuteTest.cpp
rwy0717: Oops you're right, this should be printing in unsigned decimal (`PRIuPTR`), not `p` (which is usually hex).  By the way, does tril support hex integer constants?

compiler/control/CompileMethod.cpp
rwy0717: OK, I'll fix this, and change the cast to reinterpret_cast.

compiler/runtime/OMRCodeCache.cpp
```diff
@@ -1275,8 +1276,8 @@ void
 OMR::CodeCache::dumpCodeCache()
    {
    printf("Code Cache @%p\n", this);
-   printf("  |-- warmCodeAlloc          = 0x%08x\n", _warmCodeAlloc );
-   printf("  |-- coldCodeAlloc          = 0x%08x\n", _coldCodeAlloc );
+   printf("  |-- warmCodeAlloc          = 0x%08" OMR_PRIxPTR "\n", (uintptr_t)_warmCodeAlloc );
+   printf("  |-- coldCodeAlloc          = 0x%08" OMR_PRIxPTR "\n", (uintptr_t)_coldCodeAlloc );
```
rwy0717: These casts match the surrounding code, so I figure I'll leave them? Or should these use reinterpret too??

Leonardo2718: I'd leave them. Other parts of the compiler use C-style casts too so I don't think it's worth changing these. I only insist on C++ style casts in tests, since that all "new" code. :slightly_smiling_face:

fvtest/compilertriltest/BitPermuteTest.cpp
Leonardo2718: > By the way, does tril support hex integer constants?

It sure does! :smile: 

I think these tests used to use hex literals but got changed at some point with other things. I recall there being issues with signedness at one point, but that shouldn't affect address literals.

include_core/omrformatconsts.h
rwy0717: Added, thanks.

fvtest/compilertriltest/BitPermuteTest.cpp
```diff
@@ -178,10 +179,10 @@ TEST_P(lBitPermuteTest, ConstAddressTest)
       "  (lreturn                                "
       "  (lbitpermute                            "
       "   (lload parm=0)                         "
-      "   (aconst %llu)                          "
+      "   (aconst %" OMR_PRIuPTR ")              "
       "   (iload parm=1)                         "
       "  ))))                                    ",
-      param.arrayAddress
+      uintptr_t(param.arrayAddress)
```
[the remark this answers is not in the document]
rwy0717: The casts are fixed.

fvtest/compilertriltest/BitPermuteTest.cpp
rwy0717: Heh nice. I'm leaving the pointer formatted in decimal (using `OMR_PRIuPTR`) but that's good to know :+1:

compiler/control/CompileMethod.cpp
rwy0717: fixed, and I changed the cast to use `static_cast`.

Leonardo2718:
@genie-omr build all

fvtest/compilertriltest/MaxMinTest.cpp
```diff
@@ -21,6 +21,7 @@
```
Leonardo2718: Copyright date needs to be updated.

rwy0717: OK I might wait until CI passes, then push an amended commit that updates the copyrights, if that's OK?

Leonardo2718: Sure, although the zos machines are down so it might be a while before that job passes.

rwy0717: Fixed.

rwy0717:
* Replaced `__xlc__` test with `__xlC__`.
* Fixed copyright in MaxMinTest.cpp
* Rebased changes

Leonardo2718:
@genie-omr build all

Leonardo2718:
@genie-omr build zos

Leonardo2718:
Alright, the zos build is failing because of infrastructural problems. These changes are not z specific and all the other builds (including zlinux) passed. Merging.

PR #3603, opened 2019-02-22
Run OSR live range analysis and OSR def analysis first

liqunl:
OSR live range analysis and OSR def analysis gathers information to
decide what to copy into the OSR buffer during OSR transition. This
should not change regardless of when they are performed. Optimizations
can create nodes with incorrect bytecode index. Those nodes are usually
not important and side effect free, such as `iconst` node. However, such
nodes bring difficulty on OSR live range analysis and OSR def analysis
since the two opts rely highly on the correctness of bytecode info.
Thus, move then up as the first two ilgen opts.

Also fix two minor issues found in OSR live range analysis.

### DIFF
--- a/compiler/optimizer/OSRDefAnalysis.cpp
+++ b/compiler/optimizer/OSRDefAnalysis.cpp
@@ -1403,6 +1403,8 @@ int32_t TR_OSRLiveRangeAnalysis::fullAnalysis(bool includeParms, bool containsPe
    vcount_t visitCount = comp()->incVisitCount();
    block = comp()->getStartBlock();
    _visitedBCI->empty();
+   int32_t callerIndex = comp()->getCurrentInlinedSiteIndex();
+
    while (block)
       {
       blockNum    = block->getNumber();
@@ -1507,8 +1509,9 @@ int32_t TR_OSRLiveRangeAnalysis::fullAnalysis(bool includeParms, bool containsPe
             && (comp()->isOSRTransitionTarget(TR::preExecutionOSR) || comp()->requiresAnalysisOSRPoint(node)))
             {
             TR_ByteCodeInfo bcInfo = node->getByteCodeInfo();
+            TR_ASSERT_FATAL(bcInfo.getCallerIndex() == callerIndex, "Found node with invalid caller index");
             TR_OSRPoint *osrPoint = comp()->getMethodSymbol()->findOSRPoint(bcInfo);
-            TR_ASSERT(osrPoint != NULL, "Cannot find a pre OSR point for node %p", node);
+            TR_ASSERT_FATAL(osrPoint != NULL, "Cannot find a pre OSR point for node %p", node);
 
             buildOSRLiveRangeInfo(node, _liveVars, osrPoint, liveLocalIndexToSymRefNumberMap,
                numBits, osrMethodData, containsPendingPushes);
@@ -1799,7 +1802,7 @@ void TR_OSRLiveRangeAnalysis::buildOSRLiveRangeInfo(TR::Node *node, TR_BitVector
             *deadSymRefs |= *_workDeadSymRefs; 
          }
 
-      if (newlyAllocated && !deadSymRefs->isEmpty())
+      if (newlyAllocated && deadSymRefs && !deadSymRefs->isEmpty())
          osrMethodData->addLiveRangeInfo(byteCodeIndex, deadSymRefs);
       }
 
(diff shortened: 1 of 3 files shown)


liqunl:
@andrewcraik could you review?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1503,7 +1505,7 @@ int32_t TR_OSRLiveRangeAnalysis::fullAnalysis(bool includeParms, bool containsPe
              //In involuntaryOSR mode, there are cases where the first evaluation point of a node is under a treetop  
              //that is not an OSRPoint but another treetop with the same bci is an OSRPoint. Need to call buildOSRLiveRangeInfo
              //in this case for the first evaluation point to make sure the liveness info is bookkept correctly for this bci
-             (_visitedBCI->isSet(node->getByteCodeInfo().getByteCodeIndex()) && comp()->getOSRMode() == TR::involuntaryOSR)) 
+             (node->getByteCodeInfo().getCallerIndex() == callerIndex && _visitedBCI->isSet(node->getByteCodeInfo().getByteCodeIndex()) && comp()->getOSRMode() == TR::involuntaryOSR))
```
andrewcraik: the safety check is good, but should we ever see one of these? Should this be a fatal assert since we could miss an OSR point with a bad BCI?

liqunl: We shouldn't after this change. Will add an assert

liqunl:
@andrewcraik Can you review again?

compiler/optimizer/OSRDefAnalysis.cpp
```diff
@@ -1507,6 +1509,7 @@ int32_t TR_OSRLiveRangeAnalysis::fullAnalysis(bool includeParms, bool containsPe
             && (comp()->isOSRTransitionTarget(TR::preExecutionOSR) || comp()->requiresAnalysisOSRPoint(node)))
             {
             TR_ByteCodeInfo bcInfo = node->getByteCodeInfo();
+            TR_ASSERT_FATAL(bcInfo.getCallerIndex() == callerIndex, "Found node with invalid caller index");
             TR_OSRPoint *osrPoint = comp()->getMethodSymbol()->findOSRPoint(bcInfo);
             TR_ASSERT(osrPoint != NULL, "Cannot find a pre OSR point for node %p", node);
```
andrewcraik: This should be an ASSERT_FATAL as well no? to continue with an osrPoint will just result in a segv….

andrewcraik:
I'm fine with the opt reordering - just need the TR_ASSERT for the osrPoint to be made fatal as well I think. Any concerns @vijaysun-omr ?

vijaysun-omr:
You mention that there may be difficulties due to (presumably) some transformation done by the IL gen opts that we moved to be run after the two OSR analyses. Can you please give example(s) of this ? I ask, because some small number of transformations are still done in the IL gen and depending on your example(s) I want us to be sure that none of the transformations in the IL gen can lead to the same kind of a problem.

liqunl:
@vijaysun-omr The problem I'm trying to fix is a crash in running live range info for a node with invalid bytecode info. This node is not a potential OSR point, but due to this [code](https://github.com/eclipse/omr/blob/master/compiler/optimizer/OSRDefAnalysis.cpp#L1502), we may attempt to build OSR live range info for it. We won't be able to find OSR point for such node, so we'll end up in a crash in `buildOSRLiveRangeInfo`.

One example is in FSD, the method enter hook at bytecode index 0 is an OSR point, any node with invalid bytecode info will also have 0 as their bytecode index.

We have to either guarantee bytecode info correctness on all nodes generated by ILGen opts before the two OSR analysis, or we move the two OSR analysis up to avoid dealing with such nodes. This PR does the later one.

vijaysun-omr:
@liqunl thanks for the answer. I understand the case you are fixing for the problem that was caused by IL gen opts now.

In theory a similar problem could occur from IL gen itself and I'd appreciate an issue at the project to check if there is any case in the IL gen that we need to fix as well to avoid a future bug. Once this is done, I will merge in this change.

liqunl:
@vijaysun-omr I have created one issue for openj9 as it's the only consumer of OSR.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
There is some issue with the zos_390-64 CI builds but it's not related to this PR. 
But the linux_390-64 build has passed as have builds on all other platforms. 
This is a common code change and given all the testing that has run has passed and I consider the risk of zos issues to be very low, I am merging this now.

Issue #840, opened 2017-03-07
Investigate and fix/resolve memcpy compiler warnings in OMRMachine.cpp


mstoodle:
Recently caught in a fast scrolling build log on OSX:

...
clang++ -std=c++0x -fno-rtti -fno-threadsafe-statics -Wno-deprecated -Wno-enum-compare -Wno-invalid-offsetof -Wno-write-strings -O3 -pthread -fomit-frame-pointer -fasynchronous-unwind-tables -Wreturn-type -fno-dollars-in-identifiers -Wno-parentheses -Werror=header-guard   -DBITVECTOR_BIT_NUMBERING_MSB -DUT_DIRECT_TRACE_REGISTRATION -DJITTEST -DJITBUILDER_SPECIFIC -DPROD_WITH_ASSUMES -DTR_HOST_X86 -DTR_HOST_64BIT -DBITVECTOR_64BIT -DOSX -DTR_TARGET_X86 -DTR_TARGET_64BIT -DSUPPORTS_THREAD_LOCAL -D_LONG_LONG -I'../include_core' -I'../jitbuilder/x/amd64' -I'../jitbuilder/x' -I'../jitbuilder' -I'../compiler/x/amd64' -I'../compiler/x' -I'../compiler' -I'..' -MMD -MF ../build/compiler/x/codegen/OMRMachine.o.depend.mk -MT ../build/compiler/x/codegen/OMRMachine.o -MP -o ../build/compiler/x/codegen/OMRMachine.o -c ../compiler/x/codegen/OMRMachine.cpp
../compiler/x/codegen/OMRMachine.cpp:1778:14: warning: destination for this 'memcpy' call is a pointer to dynamic class 'RealRegister';
      vtable pointer will be overwritten [-Wdynamic-class-memaccess]
      memcpy(registerFileClone[i], registerFile[i], sizeof(TR::RealRegister));
      ~~~~~~ ^
../compiler/x/codegen/OMRMachine.cpp:1778:14: note: explicitly cast the pointer to silence this warning
      memcpy(registerFileClone[i], registerFile[i], sizeof(TR::RealRegister));
             ^
             (void*)
../compiler/x/codegen/OMRMachine.cpp:1926:14: warning: destination for this 'memcpy' call is a pointer to dynamic class 'RealRegister';
      vtable pointer will be overwritten [-Wdynamic-class-memaccess]
      memcpy(registerFileClone[i], _registerFile[i], sizeof(TR::RealRegister));
[...]

PR #3604, opened 2019-02-22
Do not memcpy registers objects, copy by assignment

rwy0717:
Registers objects have vtables that should not be overwritten. Memcpy
will overwrite the entire memory of the object, including the vtable,
and clang doesn't like this. Just use regular assignment instead, which
should be equivalent.


0xdaryl:
Fixes #840

charliegracie:
@genie-omr build all

charliegracie:
Ignore the zOS issue since it is an infrastructure issue.

PR #3605, opened 2019-02-22
Add missing default case to switch statement in X CodeGenerator

rwy0717:

### DIFF
--- a/compiler/x/codegen/OMRCodeGenerator.cpp
+++ b/compiler/x/codegen/OMRCodeGenerator.cpp
@@ -1076,17 +1076,18 @@ OMR::X86::CodeGenerator::supportsMergingGuards()
 bool
 OMR::X86::CodeGenerator::supportsNonHelper(TR::SymbolReferenceTable::CommonNonhelperSymbol symbol)
    {
-   bool result = false;
+   bool result;
 
    switch (symbol)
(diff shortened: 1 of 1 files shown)


compiler/x/codegen/OMRCodeGenerator.cpp
```diff
@@ -1083,10 +1083,11 @@ OMR::X86::CodeGenerator::supportsNonHelper(TR::SymbolReferenceTable::CommonNonhe
       case TR::SymbolReferenceTable::atomicAddSymbol:
       case TR::SymbolReferenceTable::atomicFetchAndAddSymbol:
       case TR::SymbolReferenceTable::atomicSwapSymbol:
-         {
          result = true;
          break;
-         }
+      default:
+         result = false;
```
0xdaryl: Do you need to assign it to `false` because that's what it is initialized to?  I'd say either just `break` or remove the initialization to `false` from L1079.

rwy0717: No need for both, I removed the initialization at L1079.

0xdaryl:
@genie-omr build xlinux,x32linux,xcrlinux,win

This is an x86-specific change.  No need for a full genie build.

PR #3608, opened 2019-02-25
Implemented divide by zero check in BinaryEvaluator

hsoontie:
WIP Implementing divide by zero check in BinaryEvaluator

### DIFF
--- a/compiler/aarch64/codegen/BinaryEvaluator.cpp
+++ b/compiler/aarch64/codegen/BinaryEvaluator.cpp
@@ -104,6 +104,42 @@ genericBinaryEvaluator(TR::Node *node, TR::InstOpCode::Mnemonic regOp, TR::InstO
    return trgReg;
    }
 
+/**
+ * Helper that checks for division by zero before evaluating.
+ * @param[in] node : calling node
+ * @param[in] regOp : the target AArch64 instruction opcode
+ * @param[in] regOp : the matching AArch64 immediate instruction opcode
+ * @param[in] is64Bit : true when it is 64-bit operation
+ * @param[in] cg : codegenerator
+ * @return target register
+ */
+static inline TR::Register *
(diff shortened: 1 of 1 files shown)


hsoontie:
Please note that this PR borrows code from #3517 .

hsoontie:
@0xdaryl Fixed conflicts with master, squashed and removed WIP. This should now be good for review!

aarongraham9:
This should be a good for a review @0xdaryl.

0xdaryl:
An integer division-by-zero operation is undefined.  I imagine different language environments will choose to handle it differently.  Hence, it is not clear to me what your expected result is when a division by zero is detected.  You avert any kind of hardware exception by checking for a zero divisor, but then you allocate a register and don't initialize it with anything.  This will lead to all sorts of strange behaviour.  What behaviour were you trying to achieve?  I'm not sure if we could come up with a standardized response to a division by zero that would make sense.

Also, the division-by-zero case is also exceptional.  Any sort of handling of it should not pollute the mainline code, which is why the processing of it should live in an out-of-line instruction sequence or a snippet.

The preferred way of handling integer division is by performing it under a `DIVCHK` node (an IL generator would have to enforce this).  This separates the divisor check from the actual division operation and different language environments can choose to implement the check differently depending on their environment's semantics.  The DIVCHK nodes can have a symbol reference that represents a function to call in the language environment for processing.  If performing an explicit test for a zero divisor (as you have done), then the test and branch in the mainline path would take you to a snippet (off the mainline path) to likely call back into the language environment for processing.  The optimizer already knows about DIVCHKs as well and can eliminate them if proven safe.

hsoontie:
@0xdaryl It's been a while, but I think I was just trying to skip calling the evaluator in the case of division by zero. 
The code does include some unnecessary register allocation, and if using DIVCHK is the right way to go, I am happy with just closing this one and moving on if you are, Daryl.

0xdaryl:
Yes, I think this issue should be closed.  Individual language environments need to implement `TR::DIVCHK` evaluators to handle division-by-zero cases in a manner unique to their environment.

Issue #2345, opened 2018-02-24
Consider fatal asserts for unimplemented functions


0xdaryl:
Many functions in base classes do not have default implementations but are expected to have sensible implementations provided by either architecture specializations or downstream projects (or both).  This is especially true in the extensible class hierarchies where pure virtual functions are not used and asserts (i.e., `TR_ASSERT`) are used to provide that functionality at runtime.

However, `TR_ASSERT`s do not trigger in all build configurations (e.g., production builds) and it may be appropriate to replace them with fatal asserts in some contexts.

For this issue:

* Investigate the use of `notImplemented` macros in "default" functions in various hierarchies.  If the intention of the function is to act like a pure virtual function or if providing a single default implementation does not make sense then execution of that function should result in a fatal assert (`TR_ASSERT_FATAL`) regardless of the build configuration (prod, debug, prod with assumes).

* Define the `notImplemented` macros in a common header file (e.g., `infra/NotImplemented.hpp` or even in `infra/Assert.hpp` itself) to avoid repeating the definition across all source files.

0xdaryl:
Fixed in #2345.

PR #3610, opened 2019-02-25
Add a new TR_UNIMPLEMENTED() assertion to the compiler

rwy0717:
The new `TR_UNIMPLEMENTED` macro is a fatal assert, used to indicate that a function has not been implemented. Since this is a fatal assertion, callers are not obligated to return values, because the program will abort.

In as many places as I could find, I've replace code that looks like this:

```c++
TR_ASSERT(false, "Not implemented yet.");
return NULL;
```

with this:
```c++
TR_UNIMPLEMENTED();
return NULL;
```


0xdaryl:
@genie-omr build all

0xdaryl:
zOS failure is infrastructural, and the LoZ jobs have passed.  Merging.

0xdaryl:
Fixes: #2345

PR #3611, opened 2019-02-25
WIP: Enable TransparentHugePage for Nursery region under [madvise]

fengxue-IS:
When Linux TransparentHugePage is set to [madvise], set the `MADV_HUGEPAGE` flag for nursery region using `madvise()` call. This change apply to Gencon policy only.

- add portlibrary api `omrvmem_advise_memory()`
- add `adviseHugepage()` wrapper function to `MM_VirtualMemory` class

Signed-off-by: Jack Lu <Jack.S.Lu@ibm.com>

### DIFF
--- a/gc/base/MemoryManager.cpp
+++ b/gc/base/MemoryManager.cpp
@@ -375,6 +375,12 @@ MM_MemoryManager::createVirtualMemoryForHeap(MM_EnvironmentBase* env, MM_MemoryH
 		}
 	}
 
+#if defined(LINUX)
+	if ((NULL != instance) && (gc_policy_gencon == extensions->configurationOptions._gcPolicy)) {
+		instance->adviseHugepage();
+	}
+#endif /* defined(LINUX) */
+
 #if defined(OMR_VALGRIND_MEMCHECK)
 	//Use handle's Memory Base to refer valgrind memory pool
 	valgrindCreateMempool(extensions, env, (uintptr_t)handle->getMemoryBase());
--- a/gc/base/VirtualMemory.cpp
+++ b/gc/base/VirtualMemory.cpp
@@ -21,6 +21,7 @@
  *******************************************************************************/
 
 #include <string.h>
+#include <sys/mman.h>
 
 #include "omrcomp.h"
 #include "omrport.h"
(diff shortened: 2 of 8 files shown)


youngar:
@genie-omr build all

dmitripivkine:
I believe this implementation is not good enough.

Just to summarize results of out discussion. In case if MADVISE mode is provisioned we need to create infrastructure to overwrite it. We need:
- Add VM option to enable/disable madvise() feature. Default value is "enabled"
- Add API to Port Library. The general call should exist all platforms (caller should not wrap call to platform ifdef). An implementation on unsupported platforms should be empty
- Call should take page size for allocated memory as parameter. The decision is this page available for madvise() should be done inside function.
- low and high addresses of memory range should be provided as parameters. Keep in ming that THP can be created at Huge Page size aligned address only. My suggestion is to allow to provide unaligned addresses as parameters however do alignment internally (low up, top down with check that alignedLow < alignedHigh).
- as soon as implementation is ready somebody from GC team will add call of this API in proper place(s)
- I believe somebody from JIT team will add support for JIT structures.

For the record: it is possible to implement an original request to "overwrite" MADVISE mode to ALWAYS by calling of madvise() for every virtual memory reservation request fulfilled in small pages based on added VM command line option. The benefit of this approach is no need to modify number of callers of Virtual Memory consumers. However it seems is not flexible enough (yes or no across JVM). But it will do it.

charliegracie:
I think that the right solution is the simple solution. Add code in the port library which calls madvise if a port library option is set. I would enable the option by default and users of the port library can modify the option accordingly.

charliegracie:
Looking at the documentation for transparent huge pages it will only create pages for memory that is aligned to transparent huge page size.

For most linux systems transparent huge pages is set to `always`. With this setting the PortLibrary does not have to do anything to allow the virtual memory to use transparent huge pages. If the value is set to `madvise` then the PortLibrary should call `madvise` on the virtual memory. If the value is set to never than there is nothing for the PortLibrary to do.

to check the transparent huge pages setting you can read the following file:
`/sys/kernel/mm/transparent_hugepage/enabled`

The output value in that file *should* be in the following format:
`always [madvise] never`

The contents *should* always have all 3 possible choices and `[]` wrapping the currently active option.

A new helper function should be written in the port library on Linux that reads this file and sets a port global stating what the transparent huge pages setting is.

Another port global should be added which states whether or not transparent huge pages `madvise` should be honoured. This should have a default value of true.

In the port library call `omrvmem_reserve_memory` on Linux, if the allocated page size is 4k, and transparent huges pages is set to `madvise` and the port library global for honouring transparent huge pages madvise is true then the madvise call should be made.

Due to alignment concerns the value passed the address passed to madvise should be rounded up to 2M and the size should be rounded down to 2M.

fengxue-IS:
New Design Opened : https://github.com/eclipse/omr/pull/3647

Keeping this PR for reference to comments. Will close once all comments have been addressed in new PR

fengxue-IS:
closing this PR as #3647 is ready

PR #3613, opened 2019-02-26
Remove call to VMJ9 traceAssumeFailure

rwy0717:
In J9, [traceAssumeFailure](https://github.com/eclipse/openj9/blob/master/runtime/compiler/env/VMJ9.cpp#L5443-L5446) has been disabled, so calling it has no
effect. This is the only call site, so once it is removed,
traceAssumeFailure can be removed from the J9 VM.


0xdaryl:
@genie-omr build all

0xdaryl:
zOS failure is infrastructural.  LoZ tests pass.  Merging.

PR #3616, opened 2019-02-26
Correct register dependencies for x86 tableEvaluator

0xdaryl:
For non-Java projects the register dependencies for GlRegDeps were not
added to the indirect jump instruction, leading to incorrect register
assignment.  Add the appropriate register dependencies.

### DIFF
--- a/compiler/x/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/x/codegen/ControlFlowEvaluator.cpp
@@ -491,22 +491,23 @@ TR::Register *OMR::X86::TreeEvaluator::tableEvaluator(TR::Node *node, TR::CodeGe
       }
 
    TR::X86MemTableInstruction *jmpTableInstruction = NULL;
+   TR::RegisterDependencyConditions *deps = NULL;
+
+   // Add GlRegDep dependencies to indirect jump
+   //
+   if (secondChild->getNumChildren() > 0)
+      {
+      deps = generateRegisterDependencyConditions(secondChild->getFirstChild(), cg, 0, NULL);
+      deps->stopAddingConditions();
+      }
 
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build x32linux,xlinux,xcrlinux,win,osx

This is an x86-specific change.  A full genie build is not required.

0xdaryl:
@andrewcraik , I'd appreciate a review please.

compiler/x/codegen/ControlFlowEvaluator.cpp
```diff
@@ -491,22 +491,23 @@ TR::Register *OMR::X86::TreeEvaluator::tableEvaluator(TR::Node *node, TR::CodeGe
       }
 
    TR::X86MemTableInstruction *jmpTableInstruction = NULL;
+   TR::RegisterDependencyConditions *deps = NULL;
 
-   if (cg->getLinkage()->getProperties().getMethodMetaDataRegister() != TR::RealRegister::NoReg)
```
vijaysun-omr: @0xdaryl was this the condition that was somehow only allowing this logic for Java ?

vijaysun-omr:
@andrewcraik this is awaiting your review please.

compiler/x/codegen/ControlFlowEvaluator.cpp
0xdaryl: Yes.  I don't like it and it will be changed in the future.

0xdaryl: Just to be clear, the meta data register is only set to a non-NoReg register on OpenJ9-specific linkages.

andrewcraik:
I think this is ok - I'm not a deep expert on the code but I think this should achieve what @0xdaryl is looking to achieve. FYI @0dvictor when you are back

PR #3617, opened 2019-02-26
Remove obsolete limitation on zLinux

dmitripivkine:
Usage of GC Concurrent Background thread use to create long pauses on
Exclusive Access. This limitation is not necessary any more. 

### DIFF
--- a/gc/base/GCExtensionsBase.hpp
+++ b/gc/base/GCExtensionsBase.hpp
@@ -1490,11 +1490,7 @@ public:
 		, optimizeConcurrentWB(true)
 		, dirtCardDuringRSScan(false)
 		, concurrentLevel(8)
-#if defined(LINUX) && defined(S390)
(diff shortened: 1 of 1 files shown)


dmitripivkine:
I tested briefly on one of our 4-way machines. I can not see extra delays for Exclusive

fjeremic:
@dmitripivkine will we be able to observe benefit of this change over in OpenJ9?

dmitripivkine:
> @dmitripivkine will we be able to observe benefit of this change over in OpenJ9?

Yes, this change allows to use Concurrent Mark Background thread which was disabled at zLinux (only) due  delays at acquireExclusiveVMAccess(). It was related to zLinux kernel many years ago. So I believe kernel performance problem has been resolved since as well as actual code was rewritten.

charliegracie:
@genie-omr build all

charliegracie:
You can ignore the zOS failure as there are currently infrastructure issues.

PR #3618, opened 2019-02-27
Increase _coldUpgradeSampleThreshold to 20

dsouzai:
`_coldUpgradeSampleThreshold` is currently only used in the OpenJ9 project. Because the SVM has resulted in better AOT code quality, under `AGGRESSIVE_AOT`, this PR makes the upgrade sample threshold more conservative by increasing it to 20.

### DIFF
--- a/compiler/control/OMROptions.cpp
+++ b/compiler/control/OMROptions.cpp
@@ -5108,7 +5108,7 @@ void OMR::Options::setGlobalAggressiveAOT()
    // setOption(TR_NoOptServer); // questionable
 
    // conservative upgrades
-   _coldUpgradeSampleThreshold = 10; // instead of 3 or even 2
+   _coldUpgradeSampleThreshold = 20; // instead of 3 or even 2
 
    //setOption(TR_VaryInlinerAggressivenessWithTime); // aggressiveness will go gradually down; for non-AOT warm compilations
 


dsouzai:
@vijaysun-omr could you please review?

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@dsouzai do we need to add a check such that it uses the old threshold value if the improved AOT implementation (aka Symbol Validation Manager, SVM) is used (I believe it is possible to disable SVM via an option) ?

dsouzai:
Hm yeah that's a good point, I'll see if that change can be made in this PR as well (ie, if the change is possible entirely in OMR).

dsouzai:
Sigh, looks like I can't make the change in OMR; I'll have to make https://github.com/eclipse/openj9/pull/4900 dependent on this PR.

dsouzai:
z/os job failed with 

```
08:16:31  f run_configure.mk OMRGLUE=./example/glue SPEC=zos_390è(&È_øËÇÀÅÀ/ãÀâÀ f run_configure.mk OMRGLUE=./example/glue SPEC=zos_390?øÁ>:À,¦Á>,Ñ>ËÏ?Ê,Ëø/ÄÁ&Í%%êÁÉÍÁËÈ:?Ë^ È_øÀÍÊ/Â%ÁÁ/ËÄÊÑøÈËÇãëí(>?ÈÃ?Í>Àmake 64make 64?øÁ>:À,¦Á>,Ñ>ËÏ?Ê,Ëø/ÄÁ&Í%%êÁÉÍÁËÈ:?Ë^ È_øÀÍÊ/Â%ÁÁ/ËÄÊÑøÈËÇãëí(>?ÈÃ?Í>ÀCannot contact ZISVJD03: java.io.IOException: corrupted content in /openzdk/jenkins/workspace/PullRequest-zos_390-64@tmp/durable-230e20a9/jenkins-result.txt: java.lang.NumberFormatException: For input string: ""
```

Since I've added the appropriate change to use the old threshold when the SVM isn't enabled to https://github.com/eclipse/openj9/pull/4900, I think this should be ok to be merged @vijaysun-omr

vijaysun-omr:
@mpirvu can I get your okay for this change as well ?

dsouzai:
Yeah that makes sense. However I think that should probably be done in a different PR.

dsouzai:
Opened PR for renaming https://github.com/eclipse/omr/pull/3623

@vijaysun-omr this PR should be good to be merged I think

mpirvu:
Maybe I am missing something, but I don't think that #3623 addresses the issue that we want `_coldUpgradeSampleThreshold=20` when Xtune and AOT are enabled and want `_coldUpgradeSampleThreshold=10` when just Xtune is enabled.

dsouzai:
Ohh, it wasn't apparent that's what you meant in your comment:
>`setGlobalAggressiveAOT` is used for AOT and non-AOT alike (we should change its name). Thus we would make upgrades more conservative for all compilations under Xtune:virtualized regardless of AOT.

I thought you were just suggesting that we should rename the method to be more representative of what it does.

I'll look into what's involved to get what you described in https://github.com/eclipse/omr/pull/3618#issuecomment-468531525 working.

dsouzai:
I opened https://github.com/eclipse/openj9/pull/5010, since I realized that all the concerns brought up in this PR can be addressed entirely in OpenJ9.

PR #3622, opened 2019-02-28
Fix warning `hides overloaded virtual function`

charliegracie:
The first parameter to VirtualMachineState::MergeInto is declared
as `VirtualMachineState *`. In the client API and in
VirtualMachineOperandStack the first parameter is defined to be
the subclass type which Clang warns about.

### DIFF
--- a/compiler/ilgen/OMRVirtualMachineOperandStack.cpp
+++ b/compiler/ilgen/OMRVirtualMachineOperandStack.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 IBM Corp. and others
+ * Copyright (c) 2016, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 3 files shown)


charliegracie:
@genie-omr build all

Leonardo2718:
@mstoodle Had a fix for this in #3417 (see https://github.com/eclipse/omr/pull/3417/commits/a6f96493ef014cf37ad7e2fcd15d5c65b2f9c538).

charliegracie:
@Leonardo2718 believe these are 2 different issues being resolved

Leonardo2718:
Oh, yeah you're. I haden't looked closely enough :stuck_out_tongue:

compiler/ilgen/OMRVirtualMachineOperandStack.cpp
```diff
@@ -109,8 +109,9 @@ OMR::VirtualMachineOperandStack::Reload(TR::IlBuilder* b)
    }
 
 void
-OMR::VirtualMachineOperandStack::MergeInto(TR::VirtualMachineOperandStack* other, TR::IlBuilder* b)
+OMR::VirtualMachineOperandStack::MergeInto(TR::VirtualMachineState* o, TR::IlBuilder* b)
    {
+   TR::VirtualMachineOperandStack *other = (TR::VirtualMachineOperandStack *)o;
```
Leonardo2718: Can you use a static cast here:

```suggestion
   TR::VirtualMachineOperandStack *other = static_cast<TR::VirtualMachineOperandStack *>(o);
```

I know VirtualMachineOperandArray uses a C-style cast but these can be error-prone so I would prefer to use C++ style casts in new/modified code.

charliegracie: I have updated it.  I actually do not believe it is any more or less error-prone.  I didn't initially use the C++ static case because so much (almost all) of the code outside of tril uses C-style casts and I feel lightly spreading the static cast throughout just makes the code less consistent. The GC team experimented with the C++ static casts instead of C-style casts and quickly decided it was not worth it to switch everything as it did not catch any issues that C-style casts did not catch.

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

PR #3623, opened 2019-02-28
WIP: Rename Aggressive AOT methods

dsouzai:
1. This change is only affects the OpenJ9 project
2. `setGlobalAggressiveAOT` is used for AOT and non-AOT alike; this PR is opened to rename it so that it is more representative of the conditions under which it is called.


dsouzai:
@mpirvu is this along the lines of what you were looking for? I have to leave the existing methods because they're referenced in the OpenJ9 project...

Leonardo2718:
@dsouzai For the benefit of other people looking through the backlog of open PRs, can you add a quick description?

dsouzai:
Closing, because the concerns brought up in https://github.com/eclipse/omr/pull/3618  can be entirely addressed in https://github.com/eclipse/openj9/pull/5010.

PR #3628, opened 2019-03-01
Add missing dereference in copying register arrays

rwy0717:
There are a few places where we perform a deep copy of arrays of
register objects in the X OMRMachine. This used to be implemented by
memcpy'ing the elements over copies. In translating this code from using
memcpy to simple assignment, I forgot to dereference the register
pointers in the array. The result is that the array is shallow-copied,
which is very wrong. This PR adds the missing dereference operators, so
now the objects will be deep copied into the new array as intended.

The PR that introduced the error was #3604.


0xdaryl:
@genie-omr build all

PR #3629, opened 2019-03-01
Include system headers in jitbuilder API headers

rwy0717:
The jitbuilder public headers use types from stdint.h, without including
it. This patch adds the missing include, so users won't have to deal
with getting their includes in the right order. I've also included
stddef.h, if only for good measure. stddef defines NULL, size_t, and
ptrdiff_t.

### DIFF
--- a/jitbuilder/apigen/cppgen.py
+++ b/jitbuilder/apigen/cppgen.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
 
 ###############################################################################
-# Copyright (c) 2018, 2018 IBM Corp. and others
+# Copyright (c) 2018, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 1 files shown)


jitbuilder/apigen/cppgen.py
```diff
@@ -773,7 +776,11 @@ def write_common_decl(self, writer, api_desc):
         writer.write("#define {}_INCL\n\n".format(api_desc.project()))
 
         # write some needed includes and defines
-        writer.write("#include <stdint.h>\n\n")
+
```
Leonardo2718: To (try to) keep the style consistent, please remove this blank line between the comment and the code.

jitbuilder/apigen/cppgen.py
```diff
@@ -889,6 +896,9 @@ def write_class_header(self, writer, class_desc, namespaces, class_names):
         writer.write("#ifndef {}_INCL\n".format(class_desc.name()))
         writer.write("#define {}_INCL\n\n".format(class_desc.name()))
 
+        for header in self.get_common_system_includes():
+            writer.write(self.generate_include(header))
```
Leonardo2718: Might be good to add a `writer.write("\n")` after the loop to aid in the readability of the generated code.

rwy0717: I'm actually relying on line 907, just below, to output a newline after all the header includes.

jitbuilder/apigen/cppgen.py
rwy0717: Fixed.

rwy0717:
Fixed the copyright.

Leonardo2718:
@genie-omr build all

PR #3632, opened 2019-03-01
Node API computeIsCollectedReference support for ternary node

yanluo7:
Current implementation of Node::computeIsCollectedReference cannot handle
ternary nodes as it is unaware of the two value children, both of which
may need to be checked in order to answer its collected-ness. This change
provides support for ternary nodes by recursively calling the API on the
true-value child, and then, if the answer is inconclusive (i.e. true-value
node is a NULL constant), on the false-value child node. The recursion
continues if a value child is itself a ternary node. The existing
implementation logic is preserved and reused in the tail call to answer
the collected-ness of the value child nodes. The current implementation of
computeIsCollectedReference is refectored into a private method
`computeIsCollectedReferenceImpl` which preserves the logic and existing
behavior for all supported nodes and answers either `TR_yes` or `TR_no`;
in adding support for ternary nodes, a `TR_maybe` answer is possible if
both its value children returns inconclusive answer - e.g. when both
children are NULL constant, which is not impossible during some
optimization pass. Under such case, a `TR_maybe` is considered not
collected.

### DIFF
--- a/compiler/il/OMRNode.cpp
+++ b/compiler/il/OMRNode.cpp
@@ -2416,14 +2416,64 @@ OMR::Node::computeIsInternalPointer()
 bool
 OMR::Node::computeIsCollectedReference()
    {
+   TR::Compilation * comp = TR::comp();
+   TR::NodeChecklist processedNodesCollected(comp);
+   TR::NodeChecklist processedNodesNotCollected(comp);
+   return (self()->computeIsCollectedReferenceImpl(processedNodesCollected, processedNodesNotCollected) != TR_no);
+   }
+
+static TR_YesNoMaybe
+recordProcessedNodeResult(TR::Node *node, TR_YesNoMaybe collectedness, TR::NodeChecklist &processedNodesCollected, TR::NodeChecklist &processedNodesNotCollected)
+   {
+   switch (collectedness)
+      {
+      case TR_yes:
+         processedNodesCollected.add(node);
+         break;
+      case TR_no:
+         processedNodesNotCollected.add(node);
+         break;
+      case TR_maybe:
+         processedNodesCollected.add(node);
+         processedNodesNotCollected.add(node);
+         break;
+      default:
+         TR_ASSERT_FATAL(false, "Invalid collectedness result for Node %p\n", node);
+         break;
+      }
+   return collectedness;
+   }
+
+TR_YesNoMaybe
+OMR::Node::computeIsCollectedReferenceImpl(TR::NodeChecklist &processedNodesCollected, TR::NodeChecklist &processedNodesNotCollected)
+   {
    TR::Node *curNode = self();
+   TR::Node *receiverNode = curNode;
    if (curNode->getOpCode().isTreeTop())
-      return false;
+      return TR_no;
+
+   // In order to prevent from walking the same node repeatedly when
+   // one is referenced multiple times in a very deep tree structure
+   // (e.g. ternary whose child is a ternary and so on),
+   // Use 2 checklists to record following states:
+   // -- The node is not contained in either collected or uncollected checklists - first time encounter, process the node
+   //    and add it to the appropriate checklist(s) based on the result.
+   // -- Node is seen in both checklitsts - previously processed with result of TR_maybe
+   // -- Node is seen in processedNodesCollected - previously processed with result of TR_yes
+   // -- Node is seen in processedNodesNotCollected - previously processed with result of TR_no
+   bool ternarySeenCollected = processedNodesCollected.contains(receiverNode);
+   bool ternarySeenNotCollected = processedNodesNotCollected.contains(receiverNode);
+   if (ternarySeenCollected && ternarySeenNotCollected)
+      return TR_maybe;
+   else if (ternarySeenCollected)
+      return TR_yes;
+   else if (ternarySeenNotCollected)
+      return TR_no;
 
    while (curNode)
       {
       if (curNode->isInternalPointer())
-         return true;
+         return recordProcessedNodeResult(receiverNode, TR_yes, processedNodesCollected, processedNodesNotCollected);
 
       TR::ILOpCode op = curNode->getOpCode();
       TR::ILOpCodes opValue = curNode->getOpCodeValue();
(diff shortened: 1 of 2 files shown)


compiler/il/OMRNode.cpp
```diff
@@ -2472,17 +2488,37 @@ OMR::Node::computeIsCollectedReference()
...
+            else
+               {
+               // null constant under aladd, return false.
+               if (self() != curNode)
+                  return TR_no;
+               else
+                  // Null constant under ternary (i.e. we reached here via recursive calls), return maybe
+                  // to indicate need to check the other child.
+                  // Otherwise return yes for top level Null const nodes, to preserve previous logic.
+                  return (toplevel ? TR_yes : TR_maybe);
```
andrewcraik: Shouldn't this be a constant maybe? If the top level says maybe the result doesn't have to be collected does it?

yanluo7: Under previous logic, we had:
```
case TR::aconst:
    // aconst null can either be collected or uncollected. The collectedness of aconst
    // null can be determined by usedef analysis, which is too expensive for a very
    // basic query. Treating aconst null as a collected reference is always safe. The
    // only problem is: we might mark an aladd/aiadd on a temp whose value is null as
    // an internal pointer, thus creating a temp, pinning array and internal pointer
    // map for it.
    return self() == curNode && curNode->getAddress() == 0;
```
This says for Null aconst node, and self() == curNode meaning the aconst node is not under `aladd/aiadd`, we return true. 
In the new logic, if `toplevel ` is true, it means the aconst node is not under a ternary; in this case, I thought we should preserve the same logic as before and return true.

andrewcraik: I'm not sure this will be correct in all cases - surely we have to look at the children to know whether the overall result needs to be collected? if there is a large tree with many ternary operations and at least one is collected we need the result to be collected otherwise not. How can you know the situation without checking down to the leaves?

yanluo7: Only when we recursively call this API on ternary value child is `toplevel` passed in as false - so if `toplevel` is false, we know this current node is a child of ternary (as this is a recursive call) and we return TR_maybe, and the caller will then do another recursive call on the other child and also passing in `toplevel` false. 

Only when user code calls this API, `toplevel` is passed in as true, so we know we are not called recursively therefore receiver node is not the value child of ternary - so if we end up here and `toplevel` is true, we can return TR_yes as previously implemented.

andrewcraik: I still don't buy it - you could have 5 levels of aternary and all the leaves are aconst null except one - you must visit that one to know what the result should be.

yanluo7: After discussing this with Andrew, we believe this change should be able to handle nested ternary tree structure with multiple `aconst null` as leaf nodes. However, if a ternary node is referenced multiple times in a very deep tree structure, we would have to walking that same ternary node trees repeatedly as there is no indicator whether a node had already been processed. We discussed a solution where we use 2 checklists to record the result for ternary nodes that have been processed.

yanluo7:
@andrewcraik I revised the code based on our discussion using NodeChecklist avoid walking duplicate nodes in a deep tree. Could you review?

compiler/il/OMRNode.cpp
```diff
@@ -2415,26 +2415,55 @@ OMR::Node::computeIsInternalPointer()
...
+         return;
+      case TR_no:
+         processedNodesNotCollected.add(node);
+         return;
+      case TR_maybe:
+         processedNodesCollected.add(node);
+         processedNodesNotCollected.add(node);
+         return;
+      default:
+         return;
```
andrewcraik: This should assert fatal

compiler/il/OMRNode.cpp
andrewcraik: This toplevel thing still seems to bei n the code - it should disappear now you have the checklists

andrewcraik:
To minimize walking the bitvectors should be set for all processed nodes - this will ensure we don't revisit anything. I would change the name of the helper to something like recordNodeProcessingResult or something like that.

andrewcraik:
@yanluo7 a few review comments to address, but this is close to done now - thanks for adding the checklists to prevent the potential exponential walks.

compiler/il/OMRNode.cpp
yanluo7: Added fatal assert for invalid result.

yanluo7:
@andrewcraik Addressed the comments, and results are recorded in NodeChecklist for all processed nodes.

compiler/il/OMRNode.cpp
yanluo7: I think `toplevel` flag is still needed to differentiate whether a particular `aconst null` node is a root node or a leaf node under a ternary node, when it is first encountered. Since this call can be recursive, if we end up at this line, we wouldn't have known if the receiver node is a root node or leaf node under ternary, so we use this flag to tell us.
If it's a leaf node, we return `TR_yes`, otherwise return `TR_maybe` for lead node under ternary. For subsequent encounters, we check the nodelists and retrieve the previous result directly.

compiler/il/OMRNode.cpp
```diff
@@ -2472,15 +2556,45 @@ OMR::Node::computeIsCollectedReference()
...
+                  {
+                  recordProcessedNodeResult(receiverNode, TR_no, processedNodesCollected, processedNodesNotCollected);
+                  return TR_no;
+                  }
+               else
+                  {
+                  // Null constant under ternary (i.e. we reached here via recursive calls), return maybe
+                  // to indicate need to check the other child.
+                  // Otherwise return yes for top level Null const nodes, to preserve previous logic.
+                  TR_YesNoMaybe result = (toplevel ? TR_yes : TR_maybe);
```
andrewcraik: this use of toplevel is now redundant I believe and the parameter should be eliminated.

compiler/il/OMRNode.cpp
```diff
@@ -2415,26 +2415,84 @@ OMR::Node::computeIsInternalPointer()
...
 OMR::Node::computeIsCollectedReference()
+   {
+   TR::Compilation * comp = TR::comp();
+   TR::NodeChecklist processedNodesCollected(comp);
+   TR::NodeChecklist processedNodesNotCollected(comp);
+   return (self()->computeIsCollectedReferenceImpl(true, processedNodesCollected, processedNodesNotCollected) != TR_no);
+   }
+
+static void
+recordProcessedNodeResult(TR::Node *node, TR_YesNoMaybe collectedness, TR::NodeChecklist &processedNodesCollected, TR::NodeChecklist &processedNodesNotCollected)
```
andrewcraik: could I suggest that this method return the collectedness so that in the cases where the result type is hardcoded you only need to do `return recordProcessedNodeResult(…)` which prevents duplicate editing errors where the result has to change?

compiler/il/OMRNode.cpp
yanluo7: removed toplevel, it does look like redundant.

compiler/il/OMRNode.cpp
yanluo7: Addressed.

andrewcraik:
Thanks for the fixes I think this is looking good now

yanluo7:
@andrewcraik thanks, can this be merged? then we can start the review on ternary evaluator changes

andrewcraik:
@yanluo7 I am not an OMR committer so I do not have a final say on whether something can be merged. I believe the change is correct and achieves the desired aim in its current form. FYI @vijaysun-omr

compiler/il/OMRNode.cpp
```diff
@@ -2443,6 +2493,18 @@ OMR::Node::computeIsCollectedReference()
...
+            TR_YesNoMaybe thirdChildResult = curNode->getThirdChild()->computeIsCollectedReferenceImpl(processedNodesCollected, processedNodesNotCollected);
+            return recordProcessedNodeResult(receiverNode, thirdChildResult, processedNodesCollected, processedNodesNotCollected);
+            }
+         else
+            return recordProcessedNodeResult(receiverNode, secondChildResult, processedNodesCollected, processedNodesNotCollected);
+         }
+
       // opcodes associated with a symref, we should
       // be able to tell its collectedness from its symref.
       if (op.isLoadVar() || op.isLoadAddr() || op.isLoadReg())
```
vijaysun-omr: I worry when I see a register load's symbol reference being accessed since technically I don't think a register load opcode returns true for hasSymbolReference. I know there may be other places in the codebase that may be doing something similar but I want to ask if the register load evaluator figures out if the value is collected or not via some other logic and if so, if that is what you could be using here.

yanluo7: I checked the regLoad evaluator, it also assumes and queries the symbol to check for collected reference. So the logic seems consistent.

vijaysun-omr:
@genie-omr build all

PR #3634, opened 2019-03-04
Support of i/a typed ternary opcode on P codegen

yanluo7:
Currently ternary opcodes are not widely used, however there are upcoming
optimizer changes in CFG simplification that will generate more integer
and address-typed ternary opcoe (i.e. iternary, lternary, aternary). This
change fixes long ternary (lternary) evaluation on 32 bit platforms where
the result and operands need to be represented by register pairs, in all
various cases handled in evaluator; also, for address-typed ternary
(aternary), it marks the result register(s) as collected reference if
required.

### DIFF
--- a/compiler/p/codegen/ControlFlowEvaluator.cpp
+++ b/compiler/p/codegen/ControlFlowEvaluator.cpp
@@ -1095,11 +1095,20 @@ static TR::InstOpCode::Mnemonic cmp2cmpi(TR::ILOpCodes op, TR::CodeGenerator *cg
 TR::Register *OMR::Power::TreeEvaluator::iternaryEvaluator(TR::Node *node, TR::CodeGenerator *cg)
    {
    TR::DataType type = node->getType();
-   TR::Register *resultReg = type.getDataType() == TR::Float ?
+   bool two_reg = (TR::Compiler->target.is32Bit()) && type.getDataType() == TR::Int64;
+   TR::Register *resultReg = two_reg ?
+                            cg->allocateRegisterPair(cg->allocateRegister(),cg->allocateRegister()) :
+                            (type.getDataType() == TR::Float ?
                             cg->allocateSinglePrecisionRegister() :
                             (type.getDataType() == TR::Double ?
                             cg->allocateRegister(TR_FPR) :
-                            cg->allocateRegister(TR_GPR));
+                            cg->allocateRegister(TR_GPR)));
+
+   // Mark the result reg as collected reference if required.
+   if (!node->isNotCollected())
+      {
+      resultReg->setContainsCollectedReference();
+      }
 
    TR::InstOpCode::Mnemonic move_opcode = (type.isIntegral() || type.isAddress()) ? TR::InstOpCode::mr : TR::InstOpCode::fmr;
    TR::Node * firstChild = node->getFirstChild();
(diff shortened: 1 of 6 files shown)


yanluo7:
@gita-omr Hi Gita, could you review this PR?

vijaysun-omr:
@gita-omr could you please review this ?

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1155,34 +1164,82 @@ TR::Register *OMR::Power::TreeEvaluator::iternaryEvaluator(TR::Node *node, TR::C
...
+            }
          }
       else if (compare_type.isFloatingPoint())
          {
          cmp2Reg = cg->evaluate(firstChild->getSecondChild());
          if (!skip_compare)
             generateTrg1Src2Instruction(cg, TR::InstOpCode::fcmpu, node, ccr, cmp1Reg, cmp2Reg);
-         generateTrg1Src1Instruction(cg, move_opcode, node, resultReg, trueReg);
+
+         if (two_reg)
```
gita-omr: FloatingPoint will never use two registers (I think) so I would rather remove it. The same at line 1208.

compiler/p/codegen/ControlFlowEvaluator.cpp
```diff
@@ -1155,34 +1164,82 @@ TR::Register *OMR::Power::TreeEvaluator::iternaryEvaluator(TR::Node *node, TR::C
...
+            }
          }
       else
          {
          TR_ASSERT(false, "Unsupported compare type for ternary\n");
          }
 
-      TR::RegisterDependencyConditions *dep = new (cg->trHeapMemory()) TR::RegisterDependencyConditions(0, 5, cg->trMemory());
-      dep->addPostCondition(resultReg, TR::RealRegister::NoReg);
+      TR::RegisterDependencyConditions *dep = two_reg ?
```
gita-omr: please use 'two_reg ? 8:5' inside new instead of having two news

compiler/p/codegen/PPCInstruction.cpp
```diff
@@ -1571,16 +1571,43 @@ void TR::PPCControlFlowInstruction::assignRegisters(TR_RegisterKinds kindToBeAss
...
+            cg()->traceRAInstruction(cursor = generateTrg1Src1Instruction   (cg(), getOpCode2Value(), currentNode, getTargetRegister(1), getSourceRegister(1), cursor));
+            }
          cg()->traceRAInstruction(cursor = generateConditionalBranchInstruction(cg(), TR::InstOpCode::bne, currentNode, label2, getTargetRegister(0), cursor));
 
-         if (getNumSources() == 4)
+         if (useRegPair())
             {
-            cg()->traceRAInstruction(cursor = generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::cmpi4, currentNode, getTargetRegister(0), getSourceRegister(3), 0, cursor));
-            cg()->traceRAInstruction(cursor = generateConditionalBranchInstruction(cg(), TR::InstOpCode::bne, currentNode, label2, getTargetRegister(0), cursor));
+            if (getNumSources() == 6)
```
gita-omr: @yanluo7 could you please remind me what is the different number of iternary sources all about (even without pairs).

compiler/p/codegen/ControlFlowEvaluator.cpp
yanluo7: @gita-omr The two_reg flag applies to the 2nd/3rd child nodes where the result could be long value on 32 bit; the isFloatingPoint applies to the 1st compare child which could be a compare node of type float. e.g. 

lternaty
   dcompare
      some regload
      1.23456
   lconst -1
   lconst -2
The result of the ternary is a long (either -1 or -2 based on the double comparison) so on 32bit a reg pair is needed.

compiler/p/codegen/ControlFlowEvaluator.cpp
yanluo7: @gita-omr Fixed.

compiler/p/codegen/PPCInstruction.cpp
yanluo7: @gita-omr  This was existing logic, it seems to come from https://github.com/eclipse/omr/blob/6e1c999e6109f80c8e09abe629610fe4b27a66ea/compiler/p/codegen/ControlFlowEvaluator.cpp#L1284-L1285
It says if the result of the first child is a long value and we are on 32 bit, it needs to add the high order of the reg pair to the source register list. In the code above it says if we have the high order register added, in addition to comparing the lower order reg we must also compare the additional reg in order to fully test the condition.

compiler/p/codegen/ControlFlowEvaluator.cpp
gita-omr: got it

compiler/p/codegen/PPCInstruction.cpp
gita-omr: Thanks. Sorry for being picky but then it becomes quite confusing what useRegPair() means. Does it mean for the first child or for the last two. We use it for the last two and then decide on the first one based on the number of sources which is really hard to follow inside TR::PPCControlFlowInstruction::assignRegisters()

Maybe we could add two queries: useRegPairForCond() and useRegPairForResult() or somethinglike that...

yanluo7: @gita-omr I agree I think using query makes for better code -- added useRegPairForCond and renamed useRegPairForResult.

yanluo7:
@gita-omr Hi Gita, are you OK with the revision? If so, I can start the merge process. Thank you.

andrewcraik:
@gita-omr ping?

compiler/p/codegen/PPCInstruction.cpp
```diff
@@ -1571,16 +1571,43 @@ void TR::PPCControlFlowInstruction::assignRegisters(TR_RegisterKinds kindToBeAss
          {
          cg()->traceRAInstruction(cursor = generateTrg1Src1ImmInstruction(cg(), TR::InstOpCode::cmpi4, currentNode, getTargetRegister(0), getSourceRegister(0), 0, cursor));
 
-         cg()->traceRAInstruction(cursor = generateTrg1Src1Instruction   (cg(), getOpCode2Value(), currentNode, getTargetRegister(1), getSourceRegister(1), cursor));
+         if (useRegPairForResult())
+            {
```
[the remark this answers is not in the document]
gita-omr: Was hoping this can be improved since this change added more variations in child nodes numbering..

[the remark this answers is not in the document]
gita-omr: Alright, as long as you double checked we can move it to another item.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
All checks had passed before @yanluo7 changed the commit message.
So I do not want to re-run full testing.

PR #3635, opened 2019-03-04
WIP: Move StorageInfo into Z CodeGenerator

0xdaryl:
This feature is used exclusively by the Z CodeGenerator.

Keep a stub file `codegen/StorageInfo.cpp` that simply includes the
relocated version to prevent downstream breakages.  This stub will be
removed once changes in known downstream projects are made.


0xdaryl:
@genie-omr build all

0xdaryl:
WIP only because known downstream projects need modifications to prevent build breakages.  I'll un-WIP when those dependencies are resolved.

0xdaryl:
Downstream dependencies are a bit harder to resolve than I first suspected.  I'm closing this PR for now until those can be fixed cleanly.

fjeremic:
@0xdaryl FYI this should likely live in OpenJ9 as that is where all the Packed Decimal IL lives. I don't think StorageInfo is relevant to OMR.

PR #3637, opened 2019-03-04
Add tril tests for iudiv, iurem, and ludiv

aviansie-ben:
This PR adds tril tests that exercise the `iudiv`, `iurem`, and `ludiv` opcodes that implement unsigned division and remainder. Additionally, it tweaks the default constants used in tests that make use of unsigned 32- or 64-bit integers to cover more corner cases.

Issue #2227, opened 2018-01-16
Provide Power implementations of TR::ludiv and TR::lurem opcodes


0xdaryl:
Provide efficient code generator evaluators for the `TR::ldiv` and `TR::ludiv` opcodes for 32-bit and 64-bit Power and add them to the TreeEvaluator table.

aviansie-ben:
Once support for the `ludiv` opcode is complete, the relevant tril tests (see #3637) and the JitBuilder tests for the `UnsignedDiv` and `UnsignedRem` services (see #3638) using `Int64` should be re-enabled on Power.

PR #3638, opened 2019-03-04
 Add JitBuilder services for unsigned division and remainder 

aviansie-ben:
This PR adds `UnsignedDiv` and `UnsignedRem` services on `IlBuilder` to perform unsigned division and remainder. Unfortunately, due to the lack of an `lurem` opcode at this time, a hack is necessary to allow taking the unsigned remainder of two 64-bit numbers: namely, the remainder is calculated as `lhs - ((lhs / rhs) * rhs)`.

Additionally, a couple of bugs in the value propagator handlers for division and remainder have been fixed. Specifically, their special cases for `INT_MIN / -1` are no longer erroneously applied when the opcode being used is unsigned.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -6623,7 +6623,7 @@ TR::Node *constrainIdiv(OMR::ValuePropagation *vp, TR::Node *node)
          {
          int32_t lhsConst = lhs->asIntConst()->getInt();
          int32_t rhsConst = rhs->asIntConst()->getInt();
-         if (lhsConst == TR::getMinSigned<TR::Int32>() && rhsConst == -1)
+         if (lhsConst == TR::getMinSigned<TR::Int32>() && rhsConst == -1 && !isUnsigned)
             constraint = TR::VPIntConst::create(vp, TR::getMinSigned<TR::Int32>());
          else if (rhsConst != 0)
             {
@@ -6706,10 +6706,15 @@ TR::Node *constrainLdiv(OMR::ValuePropagation *vp, TR::Node *node)
       TR::VPConstraint *constraint = NULL;
       int64_t lhsConst = lhs->asLongConst()->getLong();
       int64_t rhsConst = rhs->asLongConst()->getLong();
-      if (lhsConst == TR::getMinSigned<TR::Int64>() && rhsConst == -1)
+      if (lhsConst == TR::getMinSigned<TR::Int64>() && rhsConst == -1 && !isUnsigned)
          constraint = TR::VPLongConst::create(vp, TR::getMinSigned<TR::Int64>());
       else if (rhsConst != 0L)
-         constraint = TR::VPLongConst::create(vp, TR::Compiler->arith.longDivideLong(lhsConst, rhsConst));
+         {
+         if (isUnsigned)
+            constraint = TR::VPLongConst::create(vp, ((uint64_t)lhsConst/(uint64_t)rhsConst));
+         else
+            constraint = TR::VPLongConst::create(vp, TR::Compiler->arith.longDivideLong(lhsConst, rhsConst));
+         }
       if (constraint)
          {
          vp->replaceByConstant(node, constraint, lhsGlobal);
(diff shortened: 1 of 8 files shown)


compiler/optimizer/VPHandlers.cpp
```diff
@@ -6706,10 +6706,15 @@ TR::Node *constrainLdiv(OMR::ValuePropagation *vp, TR::Node *node)
...
       int64_t lhsConst = lhs->asLongConst()->getLong();
       int64_t rhsConst = rhs->asLongConst()->getLong();
-      if (lhsConst == TR::getMinSigned<TR::Int64>() && rhsConst == -1)
+      if (lhsConst == TR::getMinSigned<TR::Int64>() && rhsConst == -1 && !isUnsigned)
          constraint = TR::VPLongConst::create(vp, TR::getMinSigned<TR::Int64>());
       else if (rhsConst != 0L)
-         constraint = TR::VPLongConst::create(vp, TR::Compiler->arith.longDivideLong(lhsConst, rhsConst));
+         {
+         if (isUnsigned)
+            constraint = TR::VPLongConst::create(vp, ((uint64_t)lhsConst/(uint64_t)rhsConst));
```
0xdaryl: Why isn't this unsigned divide an operation in the Arith class like in the signed case?  e.g. `longUnsignedDivideLong` or some variant.

aviansie-ben: As far as I can tell, the only purpose of the `ArithEnv` class is to allow OpenJ9 to override the behaviour of these operations by calling C helpers instead. Since OpenJ9 does not use unsigned opcodes, there's no reason for it to be overriding this behaviour.

mstoodle:
genie-omr build all

mstoodle:
@aviansie-ben would you consider writing some tests of these new services to go along with your implementations?

aviansie-ben:
@mstoodle I've added some tests for these services and fixed a minor issue with the `Int8` and `Int16` implementations

mstoodle:
genie-omr build all

mstoodle:
> 11:26:41  25: Assertion failed at /home/jenkins/workspace/PullRequest-linux_ppc-64_le_gcc/compiler/p/codegen/BinaryEvaluator.cpp:2021: node->getOpCodeValue() != TR::ludiv
11:26:41  25: 	TR::ludiv is not impelemented yet for 64-bit target

@aviansie-ben would you take a look at that, please?

aviansie-ben:
Ah, sorry. As I recently discovered with #3637 when adding tril tests for the unsigned div and rem opcodes, the Power codegen doesn't seem to correctly support any of the unsigned division and remainder opcodes right now. Because of that, these new tests need to be excluded on Power until that can be resolved...

aviansie-ben:
I've gone ahead and disabled the new JitBuilder tests on Power until the necessary opcodes are implemented properly.

mstoodle:
ok, just to be pedantic then: @genie-omr build all

mstoodle:
Please also add a comment to #2227 that these tests should be re-enabled once the underlying issues are addressed. Does #3674 also warrant similar treatment here?

aviansie-ben:
I've added comments to #2227 and #3673 to re-enable the relevant tests on Power once this is addressed. As for #3674, none of the test cases I added seem to be exercising the particular corner case that the Z codegen has problems with. Perhaps I should add test cases to cover that case? Or are the tril tests sufficient?

fjeremic:
@mstoodle FYI the Z issues in #3674 have been fixed in #3686 which is looking for a committer. Do you want to be the committer for that one so we don't need any "Z workarounds"?

aviansie-ben:
I've gone ahead and added test cases that should exercise `INT_MIN / -1` edge cases. With the issue with Z having been fixed, only Power needs these new tests disabled.

mstoodle:
@genie-omr build all

mstoodle:
The one failure shown looks like an infrastructure problem. 

This looks reasonable to me now.

Thanks (especially for your patience) @aviansie-ben !

Issue #3579, opened 2019-02-18
Support system limit on open files


pdbain-ibm:
Further to https://github.com/eclipse/omr/issues/3501  and https://github.com/eclipse/omr/pull/3511.

There is a problem with RLIMIT_NOFILE support in omrsysinfo_set_limit: if a user does:
```
omrsysinfo_get_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | J9PORT_LIMIT_HARD, &limit);
omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
```
in the case where `omrsysinfo_get_limit()` returns -1 (unlimited) as the limit, the `omrsysinfo_set_limit()` will fail because it exceeds the system-defined limit.

Two possible resolutions:
1. have `omrsysinfo_set_limit()` check for -1 as an argument, look up the system limit and use that
2. add a separate API to get the system limit. 

 On MacOS the limit is sysctl property `kern.maxfilesperproc` from `sysctlbyname()`, or possibly a property list on earlier versions. On Linux, it is `/proc/sys/fs/file-max`.

pdbain-ibm:
@charliegracie @youngar your thoughts?

youngar:
If the hard limit is unlimited, I don't think there should be a problem setting the soft limit to unlimited.  The only issue I can think of is that we are not properly checking the return value of `get_limit()`.  When we return `OMRPORT_LIMIT_UNKNOWN`, limit is set to `OMRPORT_LIMIT_UNKNOWN_VALUE` which is also `-1`.  This means the code should actually look like this:
```
uint64_t limit;
int rc = omrsysinfo_get_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | J9PORT_LIMIT_HARD, &limit);
if (rc != OMRPORT_LIMIT_UNKNOWN) {
  omrsysinfo_set_limit(OMRPORT_RESOURCE_FILE_DESCRIPTORS | OMRPORT_LIMIT_SOFT, limit);
}
```
Is it possible this is the problem?

pdbain-ibm:
Setting the soft limit (`rlim_cur`) to -1 or anything above the system-defined maximum (kern.maxfilesperproc: 24576 on my Mac) gives an error.

```
#include <stdio.h>
#include <sys/resource.h>
int main(int argc, char *argv[]) {
	struct rlimit lim = {0, 0};
[...]

PR #3641, opened 2019-03-05
Use system limit on file descriptors  for soft limit

pdbain-ibm:
If user tries to set the soft limit of file descriptors to unlimited on MacOS,
use the system limit instead.

Add unit  test.

Fixes https://github.com/eclipse/omr/issues/3579

### DIFF
--- a/fvtest/porttest/si.cpp
+++ b/fvtest/porttest/si.cpp
@@ -832,9 +832,10 @@ TEST(PortSysinfoTest, sysinfo_test_sysinfo_set_limit_CORE_FILE)
 	{
 		OMRPORT_ACCESS_FROM_OMRPORT(portTestEnv->getPortLibrary());
 		const char *testName = "omrsysinfo_test_sysinfo_set_limit_FILE_DESCRIPTORS";
-		intptr_t rc = -1;
+		uint32_t rc = OMRPORT_LIMIT_UNKNOWN;
 		uint64_t originalSoftLimit = 0;
 		uint64_t finalSoftLimit = 0;
+		uint64_t softSetToHardLimit = 0;
 		uint64_t originalHardLimit = 0;
 		uint64_t currentLimit = 0;
 		const uint64_t descriptorLimit = 256;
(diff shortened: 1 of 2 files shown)


charliegracie:
@genie-omr build all

pdbain-ibm:
Hmmm. The failures are at `porttest/si.cpp line  917: omrsysinfo_test_sysinfo_set_limit_FILE_DESCRIPTORS omrsysinfo_set_limit set hard limit FAILED rc=-1`, which is for the test running as root.

pdbain-ibm:
I am trying to get on one of the failing machines to investigate.  The test passes on a local Linux box.

pdbain-ibm:
- Added more diagnostic output
- Use a more conservative value for setting the file descriptor hard limit.

Tested on MacOS and Ubuntu.

charliegracie:
@genie-omr build all

pdbain-ibm:
The test fails if the hard and soft limits are the same finite value.  Fixed.

fvtest/porttest/si.cpp
```diff
@@ -835,6 +835,7 @@ TEST(PortSysinfoTest, sysinfo_test_sysinfo_set_limit_CORE_FILE)
 		intptr_t rc = -1;
```
charliegracie: shouldn't this be uint32_t to match the APIs?

fvtest/porttest/si.cpp
```diff
@@ -880,18 +882,52 @@ TEST(PortSysinfoTest, sysinfo_test_sysinfo_set_limit_CORE_FILE)
...
+		}
 
 		/* lowering the hard limit is irreversible unless privileged */
 		if (0 != geteuid()) { /* normal user */
 			/* setting the hard limit from unlimited to a finite value has unpredictable results:
 			 * the actual value may be much smaller than requested.
 			 * In that case, just try setting it to the same value.
 			 */
-			uint64_t newHardLimit =  (OMRPORT_LIMIT_UNLIMITED == rc) ? originalHardLimit: originalHardLimit - 1;
+			uint64_t newHardLimit =  ((OMRPORT_LIMIT_UNLIMITED == rc) || (originalSoftLimit == softSetToHardLimit))
```
charliegracie: This seems odd.  Can you explain this in more detail?

fvtest/porttest/si.cpp
charliegracie: I know it pre-existed your changes.  If it needs to be this type for code before your changes maybe a new variable is required for your new changes.

pdbain-ibm: set_limit returns -1 under certain conditions, which is wrong: it should be OMRPORT_LIMIT_UNKNOWN.  Updated.

fvtest/porttest/si.cpp
pdbain-ibm: Comments updated.

charliegracie:
@genie-omr build all

Issue #2082, opened 2017-11-30
Improve Linkage class conceptual integrity


0xdaryl:
The Linkage hierarchy exists to describe the ABI for passing control between caller and callee functions and returning.  Originally, its purpose was to describe the linkage to dispatch between compiled Java methods.  Over time it evolved to support multiple processor architectures, different system ABIs, and combinations of linkages.  Unfortunately, although they originated from a common root, the evolution of linkages on each platform evolved somewhat independently and each solved similar problems in different ways.

This is the umbrella epic to refactor the Linkage hierarchy, resolve its conceptual integrity issues, and provide ample documentation.  This epic is a work-in-progress, and sub-task issues will be added to this epic as this refactoring evolves.

- [ ] Discuss with the Eclipse OMR community whether "Linkage" is the proper name for this family of classes, as that term has a different connotation in other compiler technology projects and literature.   Perhaps a name based on "ABI" or "CallingConvention" is more appropriate and would be understood more broadly.

- [ ] Architect a LinkageProperties structure that can be shared across all linkages for all architectures.  Right now, each architecture defines its own unique properties class and there is much overlap.  It would be easier to ask property questions on a generic Linkage class if the means by which the properties were implemented were common.  I suspect this is why some very platform and linkage-specific functions were added to the base class in the hierarchy.

- [ ] A long term goal of this work is to make available fully-functional system ABIs for compiled methods that runtimes can use to dispatch to and call from.  Some system linkages in the current implementation are one-sided (e.g., the System ABI on x86) that support call-out only, or a subset of the ABI (e.g., no vararg support, lack of support for vector registers, etc.).

- [ ] Move functionality from CodeGenerator's and TreeEvaluator's that should be the responsibility of the Linkage hierarchy to the Linkage hierarchy.

- [ ] Develop Tril-based tests to validate the functionality of all linkages.

janvrany:
An experiment to refactor and cleanup of linkage properties: #5836, disussed on OMR Arch meeting #5816

PR #3643, opened 2019-03-06
Misc. Linkage base class cleanup

0xdaryl:
* remove obsolete functions not used in OMR nor any known downstream project
* use TR_UNIMPLEMENTED rather than bespoke macros
* logically re-order function declarations


0xdaryl:
@genie-omr build all

compiler/codegen/OMRLinkage.hpp
```diff
@@ -75,27 +75,17 @@ class OMR_EXTENSIBLE Linkage
 
    virtual void setParameterLinkageRegisterIndex(TR::ResolvedMethodSymbol *method, List<TR::ParameterSymbol> &parm)
```
0xdaryl: In the spirit of other functions in this base class, these might be made pure virtual functions.  However, I wasn't about to start pulling that thread in this basic cleanup.  That can be explored as part of #2082 eventually.

Issue #3619, opened 2019-02-27
Assertions in VerboseBuffer.cpp with CS


dmitripivkine:
I noticed assertions
```
** ASSERTION FAILED ** j9vgc.1 at VerboseBuffer.cpp:169 Assert_VGC_true(('\0' == _bufferAlloc[0]))
** ASSERTION FAILED ** j9vgc.1 at VerboseBuffer.cpp:178 Assert_VGC_true(('\0' == _bufferAlloc[0]))
```
when run with `-Xgc:concurrentScavenge -Xverbosegclog:<file>`.
I use Harmony test but I believe any serious load would reproduce the problem
In Harmony failure rate is ~4/100

fjeremic:
Which platform did this fail on? Is it software or hardware CS?

dmitripivkine:
> Which platform did this fail on? Is it software or hardware CS?

Software CS with -Xint on zLinux

dmitripivkine:
> > Which platform did this fail on? Is it software or hardware CS?
> 
> Software CS with -Xint on zLinux
Looks like output buffer expected to be used single threaded but it is not

dmitripivkine:
@amicic FYI

dmitripivkine:
option -Xverbosegclog:test.log makes it happen more often

amicic:
@RSalman please, investigate

RSalman:
Looking into this

fjeremic:
@dmitripivkine this sounds like an OpenJ9 issue. Shall we move it there?

dmitripivkine:
> @dmitripivkine this sounds like an OpenJ9 issue. Shall we move it there?

Actual assertion is in VerboseBuffer.cpp in OMR. There is no prove that it is OpenJ9 usage of OMR functionality yet. I believe it should be in OMR size for now.

RSalman:
> There is no prove that it is OpenJ9 usage of OMR functionality yet. I believe it should be in OMR size for now

Yup, it's an OMR issue... It's an intermittent issue that occurs when multiple threads write to the verbose buffer at the same time. Simple fix is to make this atomic like Balanced and Metronome have. Working on getting a PR in for the fix.

PR #3645, opened 2019-03-06
Fix assertions in VerboseBuffer.cpp with CS

RSalman:
Dedicated master GC thread in CS, which can report concurrent-start/end stanza races with some other asynchronous reporting (e.g. kickoff for concurrent mark)

1. Implementation for `enterAtomicReportingBlock` and `exitAtomicReportingBlock` with a lightweight lock. (Already implemented for Balanced and Metronome.. separate PR will be opened for OpenJ9 to make this common once OMR changes are in)

2. Introduced handleScavengeEndNoLock (wrapped by handleScavengeEnd), required by CS to avoid deadlock due to lock nesting. `handleConcurrentEnd` calls `handleScavengeEnd` with attempt to acquire a lock again, hence `handleScavengeEndNoLock` is called instead.

Fixes: #3619

### DIFF
--- a/gc/verbose/VerboseHandlerOutput.cpp
+++ b/gc/verbose/VerboseHandlerOutput.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
@@ -56,7 +56,8 @@ MM_VerboseHandlerOutput::newInstance(MM_EnvironmentBase *env, MM_VerboseManager 
 
 
 MM_VerboseHandlerOutput::MM_VerboseHandlerOutput(MM_GCExtensionsBase *extensions) :
-	_extensions(extensions)
+	_reportingLock()
+	,_extensions(extensions)
 	,_omrVM(NULL)
 	,_mmPrivateHooks(NULL)
 	,_mmOmrHooks(NULL)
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build all

charliegracie:
@amicic can you review this as well please?

amicic:
Lock is really required only for Concurrent Scavenger, not for STW Scavenger. but since it's not much used (contended lock), I'm fine with just keeping it simple and have it always enabled (and then remove it from Balanced and Realtime subclasses).
Lock nesting could be solved with using a mutex, but this is equally good solution.
Overall, this looks good to me.

fjeremic:
@RSalman FYI you can use GitHub PR keywords to automatically close an issue if this PR fixes it [1]. So in your case since this PR is a fix for #3619 you would simply add "Fixes: #3619" before the sign-off tag as per the OMR contribution guidelines [2]. This avoids churn of manually closing the issue.

Also I may suggest to avoid including tags in the PR and commit titles ("GC") as we already label the PRs with the GC tag. It also seems your PR description is different than your commit body, and there only seems to be one commit in this PR. It seems the PR has more information than the commit message, which should typically not be the case (it makes viewing/searching `git log` much easier). I may suggest you amend the message.

[1] https://help.github.com/en/articles/closing-issues-using-keywords
[2] https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines

RSalman:
Thanks @fjeremic for the suggestions and info, I've made the changes!

charliegracie:
@genie-omr build all

charliegracie:
@genie-omr build all

charliegracie:
Can you squash the commits into one?  Thanks

youngar:
@genie-omr build all

PR #3646, opened 2019-03-07
new element <regions> in gc-op-copy-forward

LinHu2016:
- add new element <regions> under gc-op-copy-forward to report
	the numbers of eden regions, nonEden regions in collection set
	and the 	numbers of regions are scheduled for evacuation and mark.

### DIFF
--- a/gc/verbose/schema.xsd
+++ b/gc/verbose/schema.xsd
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!--
-Copyright (c) 2010, 2018 IBM Corp. and others
+Copyright (c) 2010, 2019 IBM Corp. and others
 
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

gc/verbose/schema.xsd
```diff
@@ -598,6 +599,13 @@ xmlns:vgc="http://www.ibm.com/j9/verbosegc">
 		<attribute name="bytes" type="integer" use="required" />
 	</complexType>
 
+	<complexType name="regions">
```
charliegracie: Does the declaration order matter?

LinHu2016: no, it can be appended at the end of complexTypes

PR #3647, opened 2019-03-07
WIP: Add Transparent HugePage support to portLibrary

fengxue-IS:
- Add Transparent HugePage status to PortLibrary
- Read THP status from OS and store as port global
- Add Port option to enable/disable THP
- Add omrvmem_advise_hugepage API

Signed-off-by: Jack Lu <Jack.S.Lu@ibm.com>


port/linux/omrvmem.c
```diff
@@ -869,6 +877,17 @@ reserveLargePages(struct OMRPortLibrary *portLibrary, struct J9PortVmemIdentifie
 	return memoryPointer;
 }
 
+uintptr_t
+omrvmem_advise_hugepage(struct OMRPortLibrary *portLibrary, void* address, uintptr_t byteAmount)
```
charliegracie: I am not sure about this new API. This forces all users of VMem to specifically call this API at exactly the right time. I think this should happen in the reserve / commit call directly based on the same or similar decision.

port/common/omrportcontrol.c
```diff
@@ -289,6 +289,13 @@ omrport_control(struct OMRPortLibrary *portLibrary, const char *key, uintptr_t v
 		return 0;
 	}
 
+	if (0 == strcmp(OMRPORT_CTLDATA_VMEM_ADVISE_HUGEPAGE, key)) {
+#if defined(LINUX)
+		/* set value to advise OS about vmem to consider for Transparent HugePage (Only for Linux) */
+		portLibrary->portGlobals->vmemEnableMadvise &= value;
```
charliegracie: Why is the value being `&` and not just directly stored?

fengxue-IS: `&` is to make sure that `MADV_HUGEPAGE` flag is only set when THP setting is `[madvise]`, there is not point calling madvise for other cases even if user specified to enable this.

port/linux/omrvmem.c
fengxue-IS: This api should be called after `mmap()` call succeeded, given that this API is only intended for use under `[madvise]` setting, I am not sure we should force all vmem allocation to be flagged for THP. As enabling/disabling this option will make no difference from setting `THP => always/never` on the user side. 

For prototyping, if I change this to an internal function that is triggered after a successful `mmap()` call, would that be better?

charliegracie: yes I believe it would be

fengxue-IS: I added a call to the API after `getMemoryInRangeForDefaultPages` successfully completed to set hugepage flag after `mmap()` and `mprotect()` succeeded.

charliegracie:
Can `omrvmem_advise_hugepage` move into a helper instead of having it as a full public API.  This is currently only used on Linux so making it a private helper for Linux makes the most sense to me.

fengxue-IS:
Update madvise function wrapper to a helper function within `linux/omrvmem.c`, removed public API declaration.

fengxue-IS:
@charliegracie can I get a re-review please

FYI @DanHeidinga @dmitripivkine

charliegracie:
Can you rebase on the latest code from master please?

fengxue-IS:
@charliegracie Rebased on master

Issue #3644, opened 2019-03-06
Windows Server 2019 recognized as "Windows Server 2016" and version "10.0"


pshipton:
Windows Server 2019 available from Oct 2, 2018
https://cloudblogs.microsoft.com/windowsserver/2018/10/02/windows-server-2019-now-generally-available/

See https://github.com/eclipse/omr/blob/3c6f54e31bec06c23fec77f56f45db8e25250f8d/port/win32/omrsysinfo.c#L170

and https://github.com/eclipse/omr/blob/3c6f54e31bec06c23fec77f56f45db8e25250f8d/port/win32/omrsysinfo.c#L391

pshipton:
@pdbain-ibm is looking at this

pdbain-ibm:
Microsoft has stopped updating the major/minor release numbers and is not adding new `IsWindows<N>OrGreater()`functions: https://techcommunity.microsoft.com/t5/Windows-Server-Insiders/Windows-Server-2019-version-info/td-p/234472
The preferred method to get the version is to query the registry.

pdbain-ibm:
I did a quick-and-dirty hack on `omrsysinfo.c` and it now produces this:

> java version "1.8.0_211"
> Java(TM) SE Runtime Environment (build 8.0.6.0 - pwa6480sr6-20190220_01(SR6))
> IBM J9 VM (build 2.9, JRE 1.8.0 **Windows Server 2019 Datacenter** amd64-64-Bit Compressed References 20190307_411761 (JIT enabled, AOT enabled)
> OpenJ9   - d87ca9f
> OMR      - 0727558
> IBM      - 62c4d31)
> JCL - 20190219_01 based on Oracle jdk8u211-b76

pdbain-ibm:
Windows 10  gives: `IBM J9 VM (build 2.9, JRE 1.8.0 **Windows 10 Enterprise** amd64-64-Bit Compressed References ...` and Windows Server 2016: `IBM J9 VM (build 2.9, JRE 1.8.0 **Windows Server 2016 Standard** amd64-64-Bit Compressed References ...`

pdbain-ibm:
Looking at `omrsysinfo.c`:
```
#if !defined(_WIN32_WINNT_WINBLUE) || !(_WIN32_WINNT_MAXVER >= _WIN32_WINNT_WINBLUE)
		OSVERSIONINFOEX versionInfo;
#endif
```
[...]

PR #3648, opened 2019-03-08
Get Windows version information from registry if major version is 10

pdbain-ibm:
Fixes https://github.com/eclipse/omr/issues/3644

### DIFF
--- a/port/win32/omrsysinfo.c
+++ b/port/win32/omrsysinfo.c
@@ -169,7 +169,7 @@ omrsysinfo_get_env(struct OMRPortLibrary *portLibrary, const char *envVar, char 
 const char *
 omrsysinfo_get_OS_type(struct OMRPortLibrary *portLibrary)
 {
-
+BOOLEAN isServerMajorVersion10 = FALSE;
 /*
 WIN32_WINNT version constants :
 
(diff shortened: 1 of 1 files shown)


pdbain-ibm:
Running personal builds.  IBM Java 8 compiles and passes sanity tests.  OpenJ9 compiles successfully, sanity and extended tests are in progress.

pdbain-ibm:
Ready for review.

pshipton:
@pdbain-ibm Please don't take this out of draft state, it needs more discussion. I don't think this solution is compatible with what Java expects. Java has compatibility related expectations for the values that omrsysinfo_get_OS_type() returns (i.e. it expects the names which are already hard coded). If this changes, then Java will need to change as well so it continues to use the expected names.

pshipton:
The expected value for Server 2019 is "Windows Server 2019".

pdbain-ibm:
> Java has compatibility related expectations for the values that omrsysinfo_get_OS_type() returns (i.e. it expects the names which are already hard coded). If this changes, then Java will need to change as well so it continues to use the expected names.

Possible solutions:

1. get the `ReleaseId` from the registry and map the resulting strings to a product name, e.g. "1809" or "1803" -> "Windows Server 2019".
2. get the "ProductName" from the registry and match on that, e.g. strncmp() on "Windows Server 2019", "Windows 10", etc. and return a hard-coded name.
3. let OMR return the actual version and do the mapping in the JVM runtime code.

Option 1 is slow (many strcmps()) and hard to maintain with a faster release cadence.  Option 2, while kludgy, seems the most practical.

pshipton:
Is the ProductName always in English, even for localized versions of Windows?

pdbain-ibm:
> Is the ProductName always in English, even for localized versions of Windows?

Good point: not necessarily.  From "https://docs.microsoft.com/en-us/windows/desktop/sysinfo/structure-of-the-registry":

> Key names are not localized into other languages, although values may be.

According to https://techcommunity.microsoft.com/t5/Windows-Server-Insiders/Windows-Server-2019-version-info/td-p/234472:

> Windows Server 2016 will always be 1607. Once a product is released, that ID will not change.

Looks like we will need to parse the `ReleaseId` and look for known releases.

pdbain-ibm:
Proposal:
If the `ReleaseId` is "1607", return "Windows Server 2016", if "1809", "Windows Server 2019"
Options for handling other values:
1. treat the version as unknown and return the ProductName property
2. treat anything after 1809 as `Server 2019`
3. return a string of the form "Windows Server, version <ReleaseId>".  This handles customers on the Semi-Annual Channel. 

I favour 3.

> Java has compatibility related expectations for the values that omrsysinfo_get_OS_type() returns (i.e. it expects the names which are already hard coded). 

@pshipton : Apart from dtfj, where else do we check against hard-coded strings?

pshipton:
@pdbain-ibm this seems to be out of draft state. Please set it as WIP, it doesn't look ready for review yet.

By "Windows Server, version" do you mean for example "Windows Server, 1809" for Server 2019? If so I suggest "Windows Server, ReleaseId 1809" for an unknown release.

If you are asking about the Java use of omrsysinfo_get_OS_type(), the value is used to set the os.name system property, which any application could be checking.

pdbain-ibm:
> By "Windows Server, version" do you mean for example "Windows Server, 1809" for Server 2019? If so I suggest "Windows Server, ReleaseId 1809" for an unknown release.

Sorry: I misformatted the comment.  I have fixed it. 
Per the format used in https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info, I suggest `Windows Server, version 1709` for unrecognized release.  Known releases such as 1809 will have hard coded names, in this case `Windows Server 2019`

pdbain-ibm:
I downloaded the HotSpot Java 11 and tested it on Windows Server 2019:

>openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.2+9)
OpenJDK Server VM AdoptOpenJDK (build 11.0.2+9, mixed mode, emulated-client)
os.name = Windows Server 2016
os.version = 10.0

pdbain-ibm:
Tested and ready for review.

pshipton:
Looking at the example, from a Java perspective the comma in `Windows Server, version 1607` isn't ideal and may cause issues.

pdbain-ibm:
> Looking at the example, from a Java perspective the comma in `Windows Server, version 1607` isn't ideal and may cause issues.

Okay.  Comma removed.

charliegracie:
@genie-omr build win

pdbain-ibm:
Then failure in the CI build doesn't look related.  Could we re-run the test to rule out machine problems?

charliegracie:
@genie-omr build win

PR #3650, opened 2019-03-08
Stop printing the CMAKE_SRC_DIR when running cmake

youngar:
The message was being printed as a diagnostic.


charliegracie:
@dnakamura thoughts?

dnakamura:
i don't think we need the flexibility of `PROJECT_SOURCE_DIR`. The code in question checks that you haven't run `configure` in the source tree before, simply as a continence since it creates a lot of weird build failures

youngar:
I removed the change to use `PROJECT_SOURCE_DIR`, since it was unnecessary.  The real point of this change is to remove the diagnostic message.  This is a part of a larger set of changes to clean up and reduce the output from when running cmake.

charliegracie:
@genie-omr build all

PR #3651, opened 2019-03-08
Fix jitbuildertest CMake warning

youngar:
By making the source private, we no longer have to worry about giving an
absolute path to the source.

CMake Warning (dev) at fvtest/jitbuildertest/CMakeLists.txt:42 (target_sources):
    Policy CMP0076 is not set: target_sources() command converts relative paths
    to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
    the cmake_policy command to set the policy and suppress this warning.

    An interface source of target "jitbuildertest" has a relative path.
    This warning is for project developers.  Use -Wno-dev to suppress it.


Leonardo2718:
@genie-omr build all

PR #3652, opened 2019-03-08
Change OmrTracegen message from a warning to a status

youngar:
Old warning message was quite verbose, and not a real warning.  By
changing the message to status the output is less alarming.

From:
```
CMake Warning at cmake/modules/OmrTracegen.cmake:32 (message):
  OMR: No trace root has been set.  Defaulting to
  '/home/aryoung/wsp/omr/build'
Call Stack (most recent call first):
  CMakeLists.txt:43 (include)
```

To:
```
-- OMR: No trace root has been set. Defaulting to '/home/aryoung/wsp/omr/build'
```


charliegracie:
The AppVeyor build uses CMake so no further testing is required

PR #3656, opened 2019-03-11
Fix GlobalVP handling of backedge constraints

andrewcraik:
Global value propagation currently handles loops using a two pass
strategy. On the first pass we do not consider constraints entering
the loop and compute a set of backedge constraints. This set of
constraints represents the information produced by an iteration of the
loop. We then process the loop again using the entry constraints and
selectively merge the backedge constraints where required. The current
def constraint merging logic considers backedge constraints only when a
def has not been seen on the current iteration of the loop. This is
insufficient - a definition must have been seen on all paths from the
loop header to the def constraint merge point to be able to ignore the
constraints on the backedge. The intuition here is that if a def has not
happened on some path from the backedge the def from a previous
iteration could reach and so we must consider it and merge them in.

This change adds a map and bitvectors of value numbers to the global VP
representation. These bitvectors track the value numbers defined on all
paths for loops and these are consulted as part of merging definitions.
Exception edges are handled conservatively in this analysis at present -
we know nothing - this can be improved if required later.

Note that failing to produce a constraint for a node is represented with
with a NULL constraint. This constraint matches the unprocessed default value.
When merging def constraints it is only safe to do this at the point of a node's
first evaluation tracking of first evaluation point is included for correctness.

### DIFF
--- a/compiler/optimizer/OMRValuePropagation.cpp
+++ b/compiler/optimizer/OMRValuePropagation.cpp
@@ -1848,6 +1848,15 @@ TR::VPConstraint *OMR::ValuePropagation::mergeDefConstraints(TR::Node *node, int
    {
    isGlobal = true; // Will be reset if local constraints found
 
+   if (_defMergedNodes->get(node->getGlobalIndex()))
+      {
+      if (trace())
+         traceMsg(comp(), "Node n%dn has already been processed by mergeDefConstraints - returning NULL\n", node->getGlobalIndex());
+      return NULL;
+      }
+
+   _defMergedNodes->set(node->getGlobalIndex());
+
    // If the node is a use node, look at its def points and merge constraints
    // from them.
    //
@@ -1963,6 +1972,16 @@ TR::VPConstraint *OMR::ValuePropagation::mergeDefConstraints(TR::Node *node, int
             continue;
 
          defValueNumber = getValueNumber(defNode);
+         // before we can look at def nodes to build up a set of contraints we had better know that set of defs
+         // dominates the use we are considering - if not we treat the use as an unknown
+         if (node->getSymbol()->isAutoOrParm()
+             && _curDefinedOnAllPaths
+             && !_curDefinedOnAllPaths->get(node->getSymbolReference()->getReferenceNumber()))
+            {
+            if (trace())
+               traceMsg(comp(), "symRef %d is not stored on all paths - treating as unknown\n", defNode->getSymbolReference()->getReferenceNumber());
+            unseenDefsFound = true;
+            }
 
          // Only consider def nodes that are stores; other def nodes
          // (e.g. calls) will make this value unconstrained.
@@ -2293,6 +2312,13 @@ TR::VPConstraint *OMR::ValuePropagation::mergeDefConstraints(TR::Node *node, int
          }
       }
 
+   if (relative != AbsoluteConstraint)
+      {
+      if (trace())
+         traceMsg(comp(), "we are processing a relative constraint and merging with backedge constraints is not supported\n");
+      return NULL;
+      }
+
    // Now go through the defs again and look for back-edge constraints on
    // the unseen def nodes
    //
@@ -2311,8 +2337,14 @@ TR::VPConstraint *OMR::ValuePropagation::mergeDefConstraints(TR::Node *node, int
 
       sym = defNode->getSymbol();
       defValueNumber = getValueNumber(defNode);
-      if (!hasBeenStored(defValueNumber, sym, _curConstraints))
+      // if we haven't seen a def on this iteration along all paths we must consider the backedge constraints
+      if (node->getSymbol()->isAutoOrParm() 
+          && !unseenDefsFound
+          && !_curDefinedOnAllPaths->get(defNode->getSymbolReference()->getReferenceNumber()))
          {
+         if (trace())
+            traceMsg(comp(), "symref %d is not defined on all paths - consulting backedge constraints\n", node->getSymbolReference()->getReferenceNumber());
+
          // this def is unseen
          unseenDefsFound = true;
          LoopInfo *loopInfo;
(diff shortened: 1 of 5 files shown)


andrewcraik:
I am running performance validation on this right now. I have tried DayTrader which showed now regression and I have tried an IBM internal app server benchmark with no regression. I will continue running performance benchmarks but would like to start a review in parallel in case major changes are required. FYI @vijaysun-omr since I think you are the committer with the greatest knowledge of this area of the code.

andrewcraik:
FYI @JamesKingdon because you were looking at a problem that was related to this issue.

compiler/optimizer/OMRValuePropagation.cpp
```diff
@@ -1848,6 +1848,15 @@ TR::VPConstraint *OMR::ValuePropagation::mergeDefConstraints(TR::Node *node, int
    {
    isGlobal = true; // Will be reset if local constraints found
 
+   if (_defMergedNodes->get(node->getGlobalIndex()))
```
vijaysun-omr: This is just a performance optimization unrelated to the functional bug you are fixing, correct ? i.e. avoiding repeatedly looking at the same store node.

vijaysun-omr: _defMergedNodes seems to be initialized on a "per structure" basis. Will this work as you intend for a nested loop for example ? i.e. are we sure we won't end up having the processing of the outer loop and inner loop interfere in some way ?

compiler/optimizer/OMRValuePropagation.cpp
```diff
@@ -4341,6 +4380,14 @@ void TR::GlobalValuePropagation::processNaturalLoop(TR_StructureSubGraphNode *no
 
       _visitCount--;
       processRegionSubgraph(node, false, true, true);
+
```
vijaysun-omr: Do exception predecessors also need to be considered here ?

compiler/optimizer/OMRValuePropagation.cpp
```diff
@@ -4395,12 +4442,81 @@ void TR::GlobalValuePropagation::processNaturalLoop(TR_StructureSubGraphNode *no
       collectInductionVariableEntryConstraints();
 
    processRegionSubgraph(node, lastTimeThrough, true, true);
+
```
vijaysun-omr: Do exception predecessors also need to be considered here ?

compiler/optimizer/OMRValuePropagation.cpp
andrewcraik: _defMergedNodes is not an optimization it is required for correctness - it prevents us from merging backedge constraints for a value has already been evaluated and at the point of first evaluation was unconstrained resulting in us storing a NULL. The processing won't interfere because it is an evaluation order check done locally.

compiler/optimizer/OMRValuePropagation.cpp
andrewcraik: No I don't believe so - an expcetion predecessor will mean that we just say that the _definedOnAllPath interest is empty because we assume nothing was defined from that predecessor/

vijaysun-omr:
@genie-omr build all

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -90,6 +90,9 @@ template <class T> class TR_Stack;
 typedef TR::Node* (* ValuePropagationPtr)(OMR::ValuePropagation *, TR::Node *);
 extern const ValuePropagationPtr constraintHandlers[];
 
+typedef TR::typed_allocator<std::pair<TR::CFGEdge *, TR_BitVector*>, TR::Region &> DefinedOnAllPathsMapAllocator;
```
Leonardo2718:
```suggestion
typedef TR::typed_allocator<std::pair<TR::CFGEdge * const, TR_BitVector*>, TR::Region &> DefinedOnAllPathsMapAllocator;
```

I believe this will fix the build failure.

andrewcraik: Thanks @Leonardo2718 I have updated my commit - sorry I have been away

compiler/optimizer/OMRValuePropagation.hpp
```diff
@@ -90,6 +90,9 @@ template <class T> class TR_Stack;
...
+typedef std::map<TR::CFGEdge * const, TR_BitVector *, std::less<TR::CFGEdge *>, DefinedOnAllPathsMapAllocator> DefinedOnAllPathsMap;
```
Leonardo2718: I think it's the `TR::typed_allocator` that needs the `TR::CFGEdge * const`, not the `std::map` :slightly_smiling_face: . See the following example: https://github.com/eclipse/omr/blob/9f86693a1d1fc70e7c998d018869d064816e8680/compiler/optimizer/CopyPropagation.hpp#L110-L113 :

andrewcraik: I see - changed

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@Leonardo2718 are you okay with this commit now ?

andrewcraik:
So I think given that this fixes a functional issue we probably need to get it in and if there are regressions we will have to worry about them later - some of them may be false performance since the transformations done without this change are not guaranteed to be correct. Would you agree @vijaysun-omr ?

gdipersio:
Anyone know if this defect could affect the following code?

After JIT (AOT disabled), the call to someOtherMethod() on line 3 appears to be passed the *prior* loop value of "computed".  I.e., "computed" seems to escape the loop and insert itself down the stack.  The actual code has far more params and locals.

private compute( double alwaysZero ) {
    for ( int i=0; i<8; i++ ) {
        double computed = someOtherMethod( alwaysZero );
    }
}

@andrewcraik @vijaysun-omr @Leonardo2718

andrewcraik:
hard to say for your actual code. The basic code you have there computed is a dead store and alwaysZero is likely loop invariant so would not likely be impacted. Depending on what else you code does then it could. GlobalVP is quite sensitive to what the compiler IR looks like at the point where it runs.

andrewcraik:
if computed is passed to someOtherMethod and appears to be carried across the loop then there is a possibility it is affected, but Global VP currently doesn't do as much with floating point values so again I can't say with certainty just from the example.

PR #3657, opened 2019-03-11
Convert RuntimeAssumptionTable to mark with lazy sweep

andrewcraik:
The current RuntimeAssumptionTable design relies on eager clean-up
and the clean-up code is dispresed in the codebase. This change does
two things: 1) it changes the runtime assumption table to use a
self-contained clean-up mechanism defaulting to lazy cleanup and 2)
isolate the API to traverse the RAT to simplify caller implementaiton
and ensure the clean-up is done in a consistent manner.

To faciliate the lazy clean-up of the RAT the special bit in the
next assumption for the JITed body is modified to be always useable.
APIs are added to traverse assumptions both ignoring and respecting
assumptions marked for removal.

### DIFF
--- a/compiler/runtime/OMRRuntimeAssumptions.hpp
+++ b/compiler/runtime/OMRRuntimeAssumptions.hpp
@@ -55,19 +55,124 @@ enum AssumptionSameJittedBodyFlags
    MARK_FOR_DELETE=1
    };
 
-class RuntimeAssumption : public TR_Link0<RuntimeAssumption>
+class RuntimeAssumption
    {
    friend class ::TR_DebugExt;
+   friend class ::TR_RuntimeAssumptionTable;
 
    protected:
    RuntimeAssumption(TR_PersistentMemory *, uintptrj_t key)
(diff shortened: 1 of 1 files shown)


andrewcraik:
This change is marked WIP because performance evaluation is ongoing, but I would like some early feedback from reviewers.

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@andrewcraik it would be good to move this work forward and get it committed soon.

andrewcraik:
I have completed evaluation of the new API in the OpenJ9 context and things look good there so I think we are ready to move forward. Note that changes tot his API will affect OMR consumers who use the RAT so collaboration on the delivery is likely the best course.

fjeremic:
@vijaysun-omr I've been asked over in eclipse/openj9#5052 to shepard these two PRs due to the interdependence between them. Would it be ok if I take committer ownership of this PR? If so please re-assign me.

vijaysun-omr:
Thanks @fjeremic for handling this commit. I've assigned to you.

PR #3660, opened 2019-03-13
Provide Compilation API for checking for recursive calls

0xdaryl:
Provide an API in the Compilation class for checking whether calling a
given method symbol represents a recursive call to the method currently
being compiled.  Use the API throughout the code generators.

### DIFF
--- a/compiler/arm/codegen/ARMDebug.cpp
+++ b/compiler/arm/codegen/ARMDebug.cpp
@@ -423,7 +423,7 @@ TR_Debug::print(TR::FILE *pOutFile, TR::ARMImmSymInstruction * instr)
 
    bool longJump = imm &&
                    _cg->directCallRequiresTrampoline((intptrj_t)imm, (intptrj_t)bufferPos) &&
-                   (!callee || !callee->isSameMethod(_comp->getCurrentMethod()));
+                   !_comp->isRecursiveMethodTarget(sym);
 
    if (bufferPos != NULL && longJump)
       {
--- a/compiler/arm/codegen/ARMBinaryEncoding.cpp
+++ b/compiler/arm/codegen/ARMBinaryEncoding.cpp
@@ -256,9 +256,6 @@ uint8_t *TR::ARMImmSymInstruction::generateBinaryEncoding()
 
    if (getOpCodeValue() == ARMOp_bl)
       {
-      TR::ResolvedMethodSymbol *sym = getSymbolReference()->getSymbol()->getResolvedMethodSymbol();
-      TR_ResolvedMethod *resolvedMethod = (sym==NULL)?NULL:sym->getResolvedMethod();
-
       label = getSymbolReference()->getSymbol()->getLabelSymbol();
 
       if (cg()->hasCodeCacheSwitched())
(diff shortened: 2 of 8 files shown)


0xdaryl:
@genie-omr build all

compiler/arm/codegen/ARMDebug.cpp
```diff
@@ -423,7 +423,7 @@ TR_Debug::print(TR::FILE *pOutFile, TR::ARMImmSymInstruction * instr)
 
    bool longJump = imm &&
                    _cg->directCallRequiresTrampoline((intptrj_t)imm, (intptrj_t)bufferPos) &&
-                   (!callee || !callee->isSameMethod(_comp->getCurrentMethod()));
+                   !_comp->isRecursiveMethodTarget(sym);
```
vijaysun-omr: Is the change in behaviour here okay (I suspect it is) ? There is no isDLT test here but the new routine does.

vijaysun-omr: Can you please clarify the thinking behind passing in a method symbol to the new api ? 
Many of these places being changed seem to have been working with a method at the precise spot that the new api is being called. I see that there was a method symbol available in all these spots that have been changed but a method symbol always has a pointer to a method, whereas the opposite is not true.
i.e. the new api may be something that could be used in more contexts in the future if it was passing in a method instead of a symbol.

0xdaryl: One of the reasons for introducing this API was to eliminate the work to dig out the Method when starting from a Symbol.  In many of these cases, I think the only use of the Method is to help answer this recursion question and the intermediate steps to go from a Symbol to a Method are not used (I think there are a couple of minor exceptions).  So, I thought having one place that takes you from a Symbol to a Method would help make the code a bit simpler.

vijaysun-omr: Ok perhaps pulling out the code after 

```bool OMR::Compilation::isRecursiveMethodTarget(TR::Symbol *targetSymbol)
   {
   bool isRecursive = false;

   if (targetSymbol)
      {
      TR::MethodSymbol *methodSymbol = targetSymbol->isMethod() ? targetSymbol->castToMethodSymbol() : NULL;
      TR::ResolvedMethodSymbol *resolvedSymbol = methodSymbol ? methodSymbol->getResolvedMethodSymbol() : NULL;
      TR_ResolvedMethod *resolvedMethod  = resolvedSymbol ? resolvedSymbol->getResolvedMethod() : NULL;
```
into it's own routine taking a ```resolvedMethod``` as an argument that the above routine calls would be a good minor change that would add another method to the api for use from some other place in the codebase where we have a resolved method but not a resolved method symbol.

vijaysun-omr: There was also another question about a potential change in behaviour that I did'nt know was intentional or not.

0xdaryl: > Ok perhaps pulling out the code after

Sure, I'll do that.

> potential change in behaviour

I think that was just an oversight in the original code in the ARMDebug code in the first place (missing the DLT check) because the actual binary encoding does do the check. 
 However, this debug code doesn't seem to map entirely to the actual binary encoding operations anyway.  Upon reflection, I would rather not disturb this code at this point and leave this change out of this function.  ARM32 needs a fair bit of cleanup anyway and investigating this instruction class should be part of it.

0xdaryl:
@genie-omr build all

vijaysun-omr:
Thanks for refactoring. Checks have passed. Merging.

PR #3661, opened 2019-03-13
Add missing -c option when  compiling zos assembly files

dnakamura:


dnakamura:
@fjeremic

fjeremic:
Just for context without this `xlc` will think it's linking an executable and invoke `c89` twice, once to compile and once to link. Unfortunately this is not what we want! As a byproduct of not having the `-c` flag we get corrupted object files which will fail to link in targets causing "weird" looking issues.

fjeremic:
Merging this as is because we don't have CMake testing on z/OS. I've tested the fix manually as working.

PR #3662, opened 2019-03-13
Fix bugs in Type-Coercion Evaluator

dchopra001:
When normalizing values in the type-coercion evaluator on Z, we need to shift a 32-bit integer left into the high word. This is required to ensure the Test Data class floating-point instruction behaves correctly. Additionally, logic for generating literal pool values and then loading it in a register is removed. This work can be performed by the generateRegLitRefInstruction that picks the best method to load a constant.


fjeremic:
Do we plan on making further progress here? If not we should close the PR.

PR #3663, opened 2019-03-13
Fix double initialization of MUTEX in OMRMonitor.cpp

mstoodle:
When the Windows support was added, or maybe after some subsequent
bad merge happened, it looks like a #if for the Windows platform
was missed or removed from OMR::Monitor::init(), leaving a double
call to MUTEX_INIT. This commit wraps the code in the proper
conditional #if...#else...#endif analogous to similar code in
the destroy(), enter(), and exit() functions.


fjeremic:
@genie-omr build all

PR #3664, opened 2019-03-14
Reduce default output of tracegen

youngar:
Make tracegen not output any information unless verbose is enabled.


tools/tracegen/CFileWriter.cpp
```diff
@@ -45,7 +45,9 @@ CFileWriter::writeOutputFiles(J9TDFOptions *options, J9TDFFile *tdf, J9TDFGroup
 	time_t targetFileMtime = FileUtils::getMtime(fileName);
 
 	if ((false == options->force) && (targetFileMtime > sourceFileMtime)) {
-		printf("C file is already up-to-date: %s\n", fileName);
+		if (true == options->verboseOutput) {
```
charliegracie: since the variable is a boolean can you remove the `true == ` part?

youngar: Fixed. There are also some cases I saw where `false == options->force` could be replaced with `!options->force` if we wanted to.

charliegracie:
@genie-omr build all

PR #3667, opened 2019-03-15
Deprecate PRINT_ME and TRACE_EVAL macros on Z

fjeremic:
These macros have multiple definitions and are causing issues in debug
mode on z/OS. However the macros themselves are not very useful and are
actually expensive during tracing while providing no real value.

### DIFF
--- a/compiler/z/codegen/BinaryEvaluator.cpp
+++ b/compiler/z/codegen/BinaryEvaluator.cpp
@@ -67,21 +67,6 @@
 #include "z/codegen/S390GenerateInstructions.hpp"
 #include "z/codegen/S390Instruction.hpp"
 
-
-//#define TRACE_EVAL
-#if defined(TRACE_EVAL)
-#define EVAL_BLOCK
-#if defined (EVAL_BLOCK)
-#define PRINT_ME(string,node,cg) TR::Delimiter evalDelimiter(TR::comp(), TR::comp()->getOption(TR_TraceCG), "EVAL", string)
-#else
-extern void PRINT_ME(char * string, TR::Node * node, TR::CodeGenerator * cg);
-#endif
-#else
-#define PRINT_ME(string,node,cg)
-#endif
-
-#define ENABLE_ZARCH_FOR_32    1
-
 /**
  * lnegFor32bit - 32 bit code generation to negate a long integer
  */
@@ -183,9 +168,7 @@ laddConst(TR::Node * node, TR::CodeGenerator * cg, TR::RegisterPair * targetRegi
    TR::Register * lowOrder = targetRegisterPair->getLowOrder();
    TR::Register * highOrder = targetRegisterPair->getHighOrder();
 
-   if ( ENABLE_ZARCH_FOR_32 &&
-         performTransformation(comp, "O^O Use AL/ALC to perform long add.\n")
-      )
+   if (performTransformation(comp, "O^O Use AL/ALC to perform long add.\n"))
       {     // wrong carry bit to be set
       if ( l_value != 0 )
          {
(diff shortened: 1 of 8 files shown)


fjeremic:
Requesting review from @r30shah.

fjeremic:
@genie-omr build zlinux,zos

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -533,34 +516,25 @@ generic32BitAddEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
-         case TR::Int8:
+         case TR::Int32:
             {
-            PRINT_ME("bconst", node, cg);
-            value = (int32_t) secondChild->getByte();
+            value = secondChild->getInt();
             break;
             }
+         case TR::Address:
+            TR_ASSERT_FATAL(TR::Compiler->target.is32Bit(), "Unexpected address data type for generic32BitAddEvaluator");
```
Leonardo2718: In the old implementation, the case for `TR::Address` falls through to the `TR::Int32` case (assuming the assert passes) because of the missing `break;`. However, with this change, the `TR::Address` cases falls through to the `default` case that will unconditionally fatal assert. Is this change in behaviour intentional?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -658,34 +632,25 @@ generic32BitSubEvaluator(TR::Node * node, TR::CodeGenerator * cg)
...
+            }
+         case TR::Int32:
             {
-            PRINT_ME("bconst", node, cg);
-            value = (int32_t) secondChild->getByte();
+            value = secondChild->getInt();
             break;
             }
+         case TR::Address:
+            TR_ASSERT_FATAL(TR::Compiler->target.is32Bit(), "Unexpected address data type for generic32BitSubEvaluator");
```
Leonardo2718: Same comment as above.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -658,34 +632,25 @@ generic32BitSubEvaluator(TR::Node * node, TR::CodeGenerator * cg)
 
       switch (secondChild->getDataType())
          {
-         case TR::Address:
-            TR_ASSERT( TR::Compiler->target.is32Bit(), "generic32BitSubEvaluator(): unexpected data type for 64bit code-gen!");
-         case TR::Int32:
+         case TR::Int8:
```
Leonardo2718: Not necessary for this change, but given this code seems to be a duplicate of the code above it might be nice to move it to its own function.

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: I did check this in fact. The `generic32BitAddEvaluator` is called from the following functions:

- `iaddEvaluator`
- `baddEvaluator`
- `saddEvaluator`
- `caddEvaluator`
- `addrAddHelper`

Out of these the only one with a `TR::Address` should be coming from `addrAddHelper` which does not have any callers. This function is dead. I checked the tree evaluator table [1] and the address add evaluators have their own separate implementation.

There seems to be a lot of duplicate code here which should get cleaned up. I'll open up an issue for this. Incidentally though had there been a mistake it would have been caught the Tril test in [2] I believe. Do we have explicit address add/sub tests anywhere? I thought I saw a PR at some point?

[1] https://github.com/eclipse/omr/blob/4cee14984f633e4229ac2edbb12d19d0954d6847/compiler/z/codegen/TreeEvaluatorTable.hpp#L653-L654

[2] https://github.com/eclipse/omr/blob/f8fc704032f95cbc0f337eb6be54f2b06e58a773/fvtest/compilertriltest/ILValidatorTest.cpp#L77

fjeremic: I'm going to actually add another commit to remove the `TR::Address` case and fall down to the default until the cleanup is performed to handle all of this correctly.

fjeremic: > I thought I saw a PR at some point?

Ah of course it is my vague memory. It was not a PR but rather an issue opened by yours truly in #2922. All the more reason to hope on that. I'll see if I can find some time for it.

r30shah: Actually this switch case is for extracting constant from the second child of the node which should be integer for `aiadd`so even if this is called from `addrAddHelper`, second child should be int . We should not call this function when we are adding constant address.  I agree with @fjeremic , It should be unexpected type and fall down to default

fjeremic: Fixed in 86b86e8 and opened #3668.

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Opened #3668.

Leonardo2718:
@genie-omr build zlinux, zos

PR #3669, opened 2019-03-17
CMake: Add OMR_EXE_LAUNCHER variable

dnakamura:
Add OMR_EXE_LAUNCHER variable which is prepended to the command line before
running executables compiled as part of the build.


charliegracie:
@genie-omr build all

charliegracie:
What does this PR add other than a variable that appears to never be set?  Can you explain it in a bit more detail?  Thanks

dnakamura:
It becomes useful for cross compiling. It allows you to run hook/trace gen even if they are compiled for a different architecture. It behaves the same as `CMAKE_CROSSCOMPILING_EMULATOR`, however there are bugs that have not been fixed until relatively recent versions of cmake

PR #3670, opened 2019-03-17
Fix use after free bug in Port::omrfile_stat

dnakamura:
If the provided path is longer than `EsMaxPath`, the buffer containing the
unicode path will be prematurely freed


youngar:
@genie-omr build all

PR #3671, opened 2019-03-18
Remove workaround for missing remainderf on older MSVC versions

Leonardo2718:
The older versions of MSVC that did not have `std::remainder()` and
`std::remainderf()` are no longer supported, so the work around is no
longer needed.


Leonardo2718:
@genie-omr build all

fjeremic:
@Leonardo2718 any progress?

PR #3672, opened 2019-03-18
Optimize floating point store of constant +0.0

0xdaryl:
For 64-bit x86 targets simply XOR the register rather than using a
10-byte immediate load to the register.


0xdaryl:
@genie-omr build x32linux,xlinux,win,osx

This is an x86-specific change, a full genie build is not required.

0xdaryl:
@Leonardo2718, FYI.

vijaysun-omr:
Look very safe and checks have passed. Merging.

Issue #2835, opened 2018-08-09
Rework Options Processing Framework in OMR Compiler Technology


nbhuiyan:
Reworking the options processing in OMR Compiler Technology has been a topic of discussion a number of times in the past. As an open source project, it is now more important than ever before to rework the options processing framework so that new developers find it easier to work with.

This issue will be used to track the progress and discussions surrounding this rework.

## Some limitations of the current options processing framework

1. `TR::Options` does a lot more work than it is supposed to be doing. It involves command line parsing, trace and verbose log handling, option set and regex handling and many others that should really be split into other more specialized classes.
2. The compiler uses a variety of paths to answer whether a particular feature is enabled or disabled with differing ways of controlling whether the feature is enabled or disabled.  For example, Options, Optimizer, Compilation, FrontEnd, and CodeGenerator are all asked whether features are enabled or disabled, oftentimes delegating to one another to answer the question. This makes it terribly confusing for a developer to track down who’s answering the question about a particular feature.
3. The option names do not follow a consistent pattern. The lack of a consistent option naming scheme can cause confusions for developers when they want to add new options. eg. for options such as `TR_EnableFeatureX` and `TR_DisableFeatureX`, we could perhaps do with one `TR_FeatureX` option.
4. Options are not available very early in the compiler initialization, preventing us from being able to control certain aspects of the compiler initialization due to not being able to query `TR::Options`.
5. OMROptions contain a number of OpenJ9-specific options, which should not be the case. Allowing this practice to continue could potentially make the kitchen sink problem in OMROptions worse as more projects make use of the OMR compiler
6. Option table items need to be sorted in alphabetical order for the options to work properly.
7. There is no way for a user to output the list of options that are in effect/turned off.
8. Lack of an option compatibility checking mechanism.
9. Adding new options isn't always easy or straightforward due to lacking documentation on this topic. There are different environment variables for setting compiler options, with no clear guidelines on their usage.

## Some desired characteristics of the reworked options processing framework

1. Compiler features/options encapsulated in one place to be queried from. 
2.  Allow new options to be added easily and a consistent way of enabling/disabling them.  For example, one should not have to create separate “enable” and “disable” options for the same piece of functionality.
3.  Options should be settable from a variety of sources:
    - JIT command-line
    - an options file (for more involved options settings)
    - through environment variables
4. Command-line option and environment variable settings should be available very early (i.e, JIT startup) and very late (i.e., JIT shutdown) in the compilation process.
[...]

PR #3675, opened 2019-03-18
WIP: Reworked boolean-only Option processing framework for the OMR Compiler

nbhuiyan:
This PR introduces some of the main components of the new option processing framework for the OMR Compiler that aims to replace the existing Options class. This was motivated by the fact that the current Options class is difficult for developers to work with and not in an ideal state for an open-source project (see #2835). This is the first of a handful of PRs in both OMR and OpenJ9 necessary to migrate to a new option processing framework. During the transition period, the new and old implementations will co-exist, but not used at the same time by default.

# Overview

Some of the main components of the new options processing framework being introduced in this PR are the following:

* OMROptions.json
* options-gen
* A new option table design
* OptionProcessors
* OptionsBuilder
* CompilerOptionsManager (will be renamed as OptionsManager once existing Options class is deprecated)
* CompilerOptions (will be renamed as Options once existing Options class is deprecated)

The sections below discusses the roles of each of these components, and how they work together.

## OMROptions.json

OMROptions.json contains all the OMR compiler options as an array of JSON objects and is consumed by the options-gen script at build time. Here is an example of an entry in the JSON array followed by the description of the fields:

```json
{
    "name": "x86UseMFENCE",
    "category": "M",
    "desc": "Enable to use mfence to handle volatile store",
    "option-member": "TR_X86UseMFENCE",
    "type": "bool",
    "default": "false",
    "processing-fn": "setTrue",
    "subsettable": "no"
}
```

* `name`: The name field is used as an identifier for an option. Every entry should have a unique name. Options provided in the command line and the environment are matched against this field.
* `category`: The category field is used to group the options into different categories (eg, codegen, optimization, logging, etc). In this case, `"M"` stands for miscellaneous options. This information will be used when outputting options help text.
* `desc`: Option description text explaining what this option does. It will be used when outputting options help text.
* `option-member`: This field is used to identify the field in the options class that stores the value of the option.
* `type`: Type of the option data.
* `default`: This is the value the option is initialized as, and will remain that way unless changed through the command line, environment variable, or some other logic after options have been processed. For boolean options, the default is false. Numeric and textual data may use this field to set a default value, and will also have the option of using an additional field in the json object to provide as a parameter to the processing function.
* `processing-fn`: This is the name of the function in the `TR::OptionProcessors` class that will be used to process this option if it appears on the command line or environment variable.
* `subsettable`: This field is used to set whether an option can be a part of an option subset to apply to specific method compilations. This capability will be added in the near future, so currently this field is not used during the processing of options.

To add a new option, all that you would need to do is add a new entry anywhere in the json file and re-build. If the available option processing functions do not meet your requirements, you may define a new function in `TR::OptionProcessors` and specify its name in the `processing-fn` field. Currenlty, `OMROptions.json` is a very large file because it contains downstream project options as well. Downstream project options should exist in their own `Options.json`. Since there are a lot of options that can be moved downstream, this migration will happen over time as downstream projects get support for the new option processing framework.

## options-gen

Simplifying the process of adding new options was possible through the use of the options-gen script run at build time. The options-gen script takes 2 input files: `OMROptions.json` containing the OMR options, and `Options.json` (or any other name you may define as a command line option to the script) containing the project-specific options. Other command-line options include the path to the input files, as well as the `-platform` option that will exclude or include platform-specific options (this mechanism is yet to be implemented in the near future, but it will basically be another field in the JSON options object allowing you to specify platform(s) to apply an option to)

The script takes the entries in `OMROptions.json` and `Options.json`, combining them into a single list, and processing them into an internal representation as an instance of class `OptionTable`. The internal representation is then used to generate a hash table that is used at run-time when processing the options. At build time, options-gen determines the index The name of each entry is hashed using a hashing algorithm that is also used at run-time to get the index into the bucket of the hash table containing these processed entries.

The `options-gen` script outputs 4 files:

* `OptionCharMap.inc`: an array of case-insensitive values associated with the characters used in the hashing function to determine the index into the hash table. This array is indexed into using the ascii values.
* `OptionTableProperties.inc`: contains macro definitions regarding the option table properties such as min/max hash value, table size, etc
* `OptionTableEntries.inc`: contains an aggregate initialization list for the hash table
* `Options.inc`: contains the data members of the CompilerOptions class

## A new option table design

Previously, the option table was an array alphabetically ordered by the option names. Finding the entry corresponding to a command line option string in the array required a binary search in the array, comparing the string each time. In the new framework, the option table is a hash table that is indexed into using a hash function.

```
OLD DESIGN:

{ {aa},
  {ab},
  {ad},
  {ba},
  .
  .
  NULL
}

NEW DESIGN:

{ {{aa}},
  {{ab},{ba}},
  {},
  {{ad}},
  .
  .
  {}
}

```

In the example above, notice how the hash table is not perfect, or minimal, although a lot of work was done to get close to that trying to mimic `gperf`'s perfect hash function generating mechanism. This is a possible area of improvement we could perhaps revisit later to completely eliminate string comparisons during the processing of command line options. The hash function simply gives us the index into a bucket in the hash table that contain a single entry, or a number of entries that hash to the same value, requiring us to perform string comparisons to get the right entry.

## OptionProcessors

OptionProcessors is an extensible class that contains the option processing functions. There will be a set of standard processing functions (eg, setTrue, setFalse, setInt32, setStringFromOptionArg, etc) that you may use for processing an option you add to the json file. If you would need your option to be processed in a very specific way not handled by existing processors (eg, option affecting multiple members of the option class), you may simply define your own and specify its name in the `processing-fn` field of the json object.

## OptionsBuilder

This class handles the processing of command line and environment variable options and processing of the options during the initialization.

## CompilerOptions

This class contains all of the options data members by including the Options.inc file generated by options-gen. 

## CompilerOptionsManager

CompilerOptionsManager is an extensible class handling the initialization of options, looking up entries in the option table, obtaining the CompilerOptions object specific to a compilation, setting a project's default options (ones that can only be determined at runtime).

---

In addition to the components above, this PR also proposes a solution using preprocessor macros to have both old and new option processing frameworks coexist, and verifying that the outcome of querying CompilerOptions is same as the existing Options class.
You may build using CMake to test adding and querying of options. I will soon be adding more documentation, perform some cleaning up and update Makefiles and CMake build files while in the draft state.


fjeremic:
> Simplifying the process of adding new options was possible through the use of the options-gen script run at build time.

@dnakamura FYI we'll likely need to add support for this in the CMake build system. Not sure which will come first though as this is a rather large change.

0xdaryl:
I know this doesn't strictly need it (according to GitHub), but would you mind rebasing this?  I'd like to schedule a few tests with it.

nbhuiyan:
@0xdaryl I have rebased the branch. To get a build that uses the new options, you will need to run CMake configure with `-DOMR_COMPILER_NEW_OPTIONS=ON` as the new options are off by default.

0xdaryl:
This PR introduces a new methodology in the core OMR compiler for constructing some of the data structures needed to integrate the compiler technology in different language environments.  Namely, the use of build-time tooling to synthesize static data tables from a metadata description of each according to the requirements of the target language environment.  As this marks a new direction in the compiler architecture and will impose a tooling dependency on downstream consuming projects, I will provide some more context around this design choice.

The compiler functionality in OMR is intended to be integrated into the fabric of language runtimes.  It was architected to be extendable by providing a set of APIs, data structures, and default functionality but allows consuming projects to extend or override them if needed.  Many operations in the compiler are driven by data in static tables or enumerated lists.  For example, IL opcodes, IL datatypes, compiler options, IL opcode evaluator tables, instruction opcodes, etc.  However, the ability to construct this information in such a way as to allow extension and customization is challenging due to limitations in the C++ language and a desire to represent these structures in a format that is straightforward to understand.  Generating such data at build-time driven by a metadata description of the structure in JSON format allows the build to synthesize uniform C++ structures composed of OMR, optional processor architecture, and optional downstream project specializations.

Python was selected as the tool to transform the JSON representation from each specialization and produce the C++ files to be included in the build.  The specific choice of Python 3.x was driven by a number of factors:
* The availability of a rich set of standard modules integrated with Python as well as language features for easily performing a number of tasks required by this PR (e.g., JSON evaluation, command-line parsing)
* The ubiquity of Python as a DevOps tool and its availability in all build environments supported by OMR
* OMR already has a build-time dependency on Python 3.x via the JitBuilder language bindings introduced in 2018, although this dependency was only exposed to downstream projects if they are consuming JitBuilder and using a language binding
* Developer familiarity with Python syntax and semantics over other scripting languages
* Python 3.x was chosen over Python 2.7 due to the end-of-life of Python 2.7 on January 1, 2020

One of the implications of this design choice is that downstream projects consuming and building with the OMR compiler will now require Python 3.x as part of their build pipeline (this is a build-only dependence, not runtime).  This is not unlike other compiler frameworks such as LLVM that impose a build-time dependency on Python.

During the design discussions for this PR, developers raised concerns about the possible extension of this approach to generating C++ functions.  The consensus, however, was that at this time this approach should be limited to auto-generating data only as no reasonable use cases for functions were conceived (other than getters/setters and simple initializers).  Any future proposal for auto-generated functions should be discussed by the community for merit.

Looking beyond this PR, this technique of generating data tables at build-time to facilitate extension and integration with downstream projects will be applied to other data structures and enums in the OMR compiler, such as the IL opcode table, IL data types, simplifier|value propagation|tree evaluator tables, etc.  Furthermore, the availability of Python as part of the DevOps toolchain will allow the OMR compiler (or other parts of OMR) to explore other ways to simplify the build, test, or integration of OMR into downstream projects.

DanHeidinga:
Some discussion about changes to option processing from a downstream consumer of OMR: https://github.com/eclipse/openj9/issues/6919#issuecomment-781461073

PR #3677, opened 2019-03-19
CMake: Search for windows message compiler

dnakamura:
Don't just assume that its on the path

### DIFF
--- a/cmake/modules/platform/os/win.cmake
+++ b/cmake/modules/platform/os/win.cmake
@@ -36,3 +36,15 @@ list(APPEND OMR_PLATFORM_DEFINITIONS
 	-D_WIN32_WINNT=0x0601
 	-DWINVER=0x0601
 )
+
+get_filename_component(kit_dir "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot]" REALPATH)
+if(OMR_ENV_DATA64)
+	set(kit_dir "${kit_dir}/bin/x64")
+else()
+	set(kit_dir "${kit_dir}/bin/x86")
+endif()
+# find the message compiler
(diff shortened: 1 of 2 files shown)


DanHeidinga:
Actually, looking at the appveyor build, I see it failed due to:
```
'CMAKE_MC_COMPILER-NOTFOUND' is not recognized as an internal or external command,
1207         operable program or batch file.
```

@dnakamura Is this a path setup issue?

dnakamura:
@DanHeidinga  The issue was that when cmake was run, `mc` was not on the path, but it was at build time. Have fixed by adding a hint which points to the windows SDK dir

charliegracie:
@genie-omr build win xlinux

cmake/modules/platform/os/win.cmake
```diff
@@ -36,3 +36,15 @@ list(APPEND OMR_PLATFORM_DEFINITIONS
...
 )
+
+get_filename_component(kit_dir "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot]" REALPATH)
+if(OMR_ENV_DATA64)
+	set(kit_dir "${kit_dir}/bin/x64")
+else()
+	set(kit_dir "${kit_dir}/bin/x86")
+endif()
+# find the message compiler
+find_program(CMAKE_MC_COMPILER mc.exe HINTS "${kit_dir}")
```
charliegracie: Could the `kit_dir` be some call that a user could specify if they have it installed in a random location that they do not want to have on their path?

dnakamura: no matter where they install it, the location should still be stored in the registry (which is where we pull `kit_dir` from). If for whatever reason its not in the registry you could still specify `-DCMAKE_MC_COMPILER=xxxx` on the command line.

PR #3678, opened 2019-03-19
CMake: add support for exporting symbols from libraries

dnakamura:
Currently only supported for MSVC

### DIFF
--- a/cmake/modules/OmrPlatform.cmake
+++ b/cmake/modules/OmrPlatform.cmake
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, 2017 IBM Corp. and others
+# Copyright (c) 2017, 2019 IBM Corp. and others
 #
 # This program and the accompanying materials are made available under
 # the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


charliegracie:
@genie-omr build win xlinux

charliegracie:
@youngar thoughts?

cmake/modules/OmrPlatform.cmake
```diff
@@ -112,6 +112,12 @@ macro(omr_platform_global_setup)
 	if(COMMAND omr_toolconfig_global_setup)
 		omr_toolconfig_global_setup()
 	endif()
+
+	# Create default implementation of omr_process_exports
+	if(NOT COMMAND omr_process_exports)
+		function(omr_process_exports)
```
charliegracie: Is omr_process_exports a toolchain only function?

charliegracie: @dnakamura ^^^^ ?

dnakamura: I'm not entirely sure I understand what you mean. My logic is that it would be implemented at the toolchain level (since each toolchain would require different logic for exporting symbols). However I also want to create a default implementation so that a toolchain file does not need to implement the function if nothing is required to export symbols (ie linux)

youngar:
I see how this mimics the older system, but only supporting Windows. Does Windows not export symbols from DLLs without a module definition? Would [GenerateExportHeader()](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) have worked for this? Maybe we could also use [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) in some other libraries?

dnakamura:
@youngar   GenerateExportHeader() would require source level changes, and has no chance of supporting zos.  also WINDOWS_EXPORT_ALL_SYMBOLS seems to completely fall apart when building under cygwin

cmake/modules/OmrPlatform.cmake
charliegracie: I am not sure I like the empty declaration here. I guess it beats the alternative of having to check to see if the function is defined at all of the callers.  If there was only going to be one caller I would suggest we just move the check there but I believe this is required by every share library correct?

dnakamura: Correct. But i was thinking on it, and I believe I'm going to do some refactoring. There will be a `omr_process_exports` function which performs some basic housekeeping (ensuring arguments are sane, etc) which would then call the toolchain specific function.

charliegracie:
@genie-omr build all

youngar:
Lgtm!

PR #3680, opened 2019-03-20
Add new evaluator and refactor for ardbari opCode

NigelYiboYu:
Add ardbarEvaluator() for ardbari node.

Remove the hardware concurrent scavenge peephole pass
replaceGuardedLoadWithSoftwareReadBarrier(). This is OpenJ9 specific and
isn't a proper software concurrent scavenge implementation.

Refactor aloadHelper() by moving some of its code into separate functions so
that OMR dependant projects can make use of them.

Change evaluateNULLCHKWithPossibleResolve() to avoid evaluating ardbari
if it's not evaluated. ardbari has side-effects and should be evaluted
before null checked.


compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -439,7 +439,9 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
    {
    TR::Compilation *comp = cg->comp();
 
-   if (TR::Compiler->target.cpu.isZ() && TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+   if (TR::Compiler->target.cpu.isZ()
+           && !TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier()
+           && TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
```
fjeremic: Why do we still need to do this anchoring business at all for hardware/software? I thought we are deprecating this code for both hardware and software and using the new common lowering phase for generating read barriers?

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -5016,56 +5014,15 @@ aloadHelper(TR::Node * node, TR::CodeGenerator * cg, TR::MemoryReference * tempM
...
-      if (!symbol->isInternalPointer()  &&
-          !symbol->isNotCollected()     &&
-          !symbol->isAddressOfClassObject() &&
-          // LowerTrees transforms unloadable aconsts to iaload <gen. int shadow> / aconst NULL.  All aconsts are never collectable.
-          !(symRef->isLiteralPoolAddress() && (node->getOpCodeValue() == TR::aloadi) && constNode->isClassUnloadingConst()))
-         {
-         tempReg->setContainsCollectedReference();
-         }
-      }
+   tempReg = OMR::Z::TreeEvaluator::checkAndAllocateReferenceRegister(node, cg, dynLitPoolLoad);
```
fjeremic: I don't think we should be using `OMR::Z::`, I think this should be `TR::` so we make use to the extensible class mechanism.

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11486,7 +11487,8 @@ void
 OMR::Z::TreeEvaluator::forwardArrayCopySequenceGenerator(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR::Node *byteLenNode, TR_S390ScratchRegisterManager *srm, TR::LabelSymbol *mergeLabel)
    {
 #ifdef J9_PROJECT_SPECIFIC
-   bool mustGenerateOOLGuardedLoadPath = TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() &&
+   bool mustGenerateOOLGuardedLoadPath = !TR::Compiler->om.shouldReplaceGuardedLoadWithSoftwareReadBarrier() &&
+                                         TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads() &&
```
fjeremic: Why are the hardware/software paths different here?

compiler/codegen/OMRCodeGenPhase.cpp
NigelYiboYu: This has been removed.

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: This is how the new software concurrent scavenge refArrayCopy differ from the old hardware concurrent scavenge arrayCopy. 
The old array copy uses LGG loop to copy arrays; whereas the new SW-CS simply calls a helper if it's in concurrent scavenge GC cycle. Hence, the difference between HW and SW paths.

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: Fixed.

compiler/z/codegen/OMRTreeEvaluator.cpp
fjeremic: I'm not sure I'm following. Have we not decided that on the OpenJ9 side we will fall back to the reference arraycopy helper even for the HW-CS? I thought we determined it will likely be faster to just call the helper to do the array copy rather than potentially hitting the read barrier on every single loop iteration.

fjeremic: Did we carry out this performance experiment?

NigelYiboYu: We haven't carried out the performance experiments. Without performance results, it's better to keep the HW-CS path as is until one is proven to be better than the other. Maybe I should evaluate these two and finalize the design now?

fjeremic: Yes, I think we should do this now. No better time IMO.

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build zos

compiler/z/codegen/OMRTreeEvaluator.cpp
```diff
@@ -11824,39 +11827,9 @@ OMR::Z::TreeEvaluator::primitiveArraycopyEvaluator(TR::Node* node, TR::CodeGener
...
-      TR::TreeEvaluator::primitiveArraycopyEvaluator(node, cg, byteSrcNode, byteDstNode, byteLenNode);
-
-#ifdef J9_PROJECT_SPECIFIC
-      TR::TreeEvaluator::genWrtbarForArrayCopy(node, byteSrcObjReg, byteDstObjReg, byteSrcNode->isNonNull(), cg);
-#endif
-      }
-
-   cg->decReferenceCount(byteSrcObjNode);
-   cg->decReferenceCount(byteDstObjNode);
+   TR_UNIMPLEMENTED();
```
fjeremic: Given our offline chat it sounded like reference arraycopy does not belong in OMR at all. We should just remove it altogether and override `arraycopyEvaluator` in OpenJ9. In OpenJ9 we can make the `node->isReferenceArrayCopy()` checks, etc.

NigelYiboYu: `referenceArrayCopy` is removed in 309a44a182123a3cee305f32616f5a3b107ce541

compiler/z/codegen/OMRTreeEvaluator.cpp
NigelYiboYu: Performance evaluation on helper vs load-store-loop has been done. My current pull requests use load-store-loops when CS is active for both HW and SW CS.

compiler/z/codegen/OMRTreeEvaluator.hpp
```diff
@@ -609,7 +617,7 @@ class OMR_EXTENSIBLE TreeEvaluator: public OMR::TreeEvaluator
     *     Register depdendecy conditions containg registers allocated within Internal Control Flow
     * 
     */
-   static TR::RegisterDependencyConditions* generateMemToMemElementCopy(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR_S390ScratchRegisterManager *srm, bool isForward, bool needsGuardedLoad, bool genStartICFLabel=false);
+   static TR::RegisterDependencyConditions* generateMemToMemElementCopy(TR::Node *node, TR::CodeGenerator *cg, TR::Register *byteSrcReg, TR::Register *byteDstReg, TR::Register *byteLenReg, TR_S390ScratchRegisterManager *srm, bool isForward, bool needsGuardedLoad, bool genStartICFLabel=false, TR::RegisterDependencyConditions* deps = NULL);
```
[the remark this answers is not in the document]
NigelYiboYu: Added in 6147d9132bb72ad7869883a717a590da47bfbbc6

fjeremic:
@NigelYiboYu can you fix the conflicts? We should be ready to go here otherwise.

fjeremic:
@genie-omr build all

PR #3681, opened 2019-03-20
Fix incorrect ldiv magic values on BE platforms

aviansie-ben:
When performing 64-bit signed division by a constant which is not a
power of 2, the division can be replaced with a multiplication using
some fixed point arithmetic trickery. Our previous implementation of
doing this had some hard-coded magic values for common divisors. For
unknown historical reasons, the table containing the 64-bit magic values
was actually an array of 32-bit integers which were being read as 64-bit
integers.

However, the order of 32-bit integers in the aforementioned array was
assuming that the platform was little endian. On big endian platforms,
the values of this table were being read with the two 32-bit parts of
each of the 64-bit integers swapped. As a result, 64-bit signed divides
of non-constant values by certain large constants such as 0x300000000
were being optimized using incorrect magic values on big endian
platforms.

This issue has been resolved by using an array of 64-bit integers
instead of an array of 32-bit integers read as 64-bit integers. Doing
this results in the endianness of the values being correctly handled at
compile time.

### DIFF
--- a/compiler/codegen/OMRCodeGenerator.cpp
+++ b/compiler/codegen/OMRCodeGenerator.cpp
@@ -2059,22 +2059,17 @@ OMR::CodeGenerator::compute64BitMagicValues(
 
    // Cache some common denominators and their magic values.  The key values in this
    // array MUST be in numerically increasing order for the binary search to work.
-   //
-   // The table is composed of 32-bit values because the compiler seems to have a problem
-   // statically initializing it with int64_t constant values.
 
    #define NUM_64BIT_MAGIC_VALUES 6
-   #define TOINT64(x) (*( (int64_t *) &x))
-   static uint32_t div64BitMagicValues[NUM_64BIT_MAGIC_VALUES][6] =
+   static int64_t div64BitMagicValues[NUM_64BIT_MAGIC_VALUES][3] =
+   //     Denominator                     Magic Value   Shift
 
-   //     Denominator        Magic Value          Shift
-
-      { {    3, 0,    0x55555556, 0x55555555,    0, 0 },
-        {    5, 0,    0x66666667, 0x66666666,    1, 0 },
-        {    7, 0,    0x24924925, 0x49249249,    1, 0 },
-        {    9, 0,    0x71c71c72, 0x1c71c71c,    0, 0 },
-        {   10, 0,    0x66666667, 0x66666666,    2, 0 },
-        {   12, 0,    0xaaaaaaab, 0x2aaaaaaa,    1, 0 } };
+      { {           3, CONSTANT64(0x5555555555555556),      0 },
+        {           5, CONSTANT64(0x6666666666666667),      1 },
+        {           7, CONSTANT64(0x4924924924924925),      1 },
+        {           9, CONSTANT64(0x1c71c71c71c71c72),      0 },
+        {          10, CONSTANT64(0x6666666666666667),      2 },
+        {          12, CONSTANT64(0x2aaaaaaaaaaaaaab),      1 } };
 
    // Quick check if 'd' is cached.
    first = 0;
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -2059,36 +2059,31 @@ OMR::CodeGenerator::compute64BitMagicValues(
 
    // Cache some common denominators and their magic values.  The key values in this
    // array MUST be in numerically increasing order for the binary search to work.
-   //
-   // The table is composed of 32-bit values because the compiler seems to have a problem
-   // statically initializing it with int64_t constant values.
```
0xdaryl: I wrote that comment in 2000 because the Microsoft Windows compiler at the time wasn't initializing this table correctly with 64-bit constants.  I'll just assume, 19 years later, that this problem is fixed in current MSVC's...

0xdaryl:
@gita-omr, would you mind reviewing please?

compiler/codegen/OMRCodeGenerator.cpp
gita-omr: Wow, great to know the reason. Thanks!

gita-omr:
LGTM

compiler/codegen/OMRCodeGenerator.cpp
```diff
@@ -2059,36 +2059,31 @@ OMR::CodeGenerator::compute64BitMagicValues(
...
+      if (div64BitMagicValues[mid][0] == d)
          {
-         *m = TOINT64(div64BitMagicValues[mid][2]);
-         *s = TOINT64(div64BitMagicValues[mid][4]);
+         *m = div64BitMagicValues[mid][1];
+         *s = div64BitMagicValues[mid][2];
          return;
          }
-      else if (d > TOINT64(div64BitMagicValues[mid][0]))
+      else if (d > div64BitMagicValues[mid][0])
```
0xdaryl: I think you still need the `int64_t` conversion when extracting the divisor from this table of UNSIGNED constants to ensure the binary search operates the way it's intended to.   This is because these magic values are to be used for SIGNED division and hence the divisors (and the divisor column) are signed quantities.

I might be mistaken with the C++ implicit type conversion rules (ahem @Leonardo2718), but I believe this comparison will be done as an unsigned comparison.  This means that if this table is ever expanded to include negative divisors then it may not work as expected.

This applies to the other use above as well.

gita-omr: Could we make it a table of  SIGNED constants?

0xdaryl: Since they're initialized with hex constants then I'd say yes.   In fact, the API to this function returns the table values as signed quantities anyway.  However, I will caution the author to do due diligence on this just to be sure it's a correct thing to do.

aviansie-ben: I've taken a closer look and I don't see anything that relies on the fact that the values are unsigned. In fact, all uses of these values besides the comparisons during the binary search seem to immediately convert the values to signed integers and the constants being loaded in are actually signed integers being converted to unsigned integers.

On that note, `compute32BitMagicValues` seems to have the same issue, so I'll go ahead and fix that while I'm at it.

Leonardo2718:
@genie-omr build all

0xdaryl:
Please investigate the CI build failures.

Leonardo2718:
@genie-omr build all

0xdaryl:
Your commit message does a great job describing the problem, but doesn't say anything about the fix this PR introduces.  Can you add a sentence about that please?

aviansie-ben:
Fixed

Leonardo2718:
@genie-omr build all

PR #3683, opened 2019-03-21
Prevent object movement in Nursery contract

amicic:
In a rare case (with very small or no Nursery tilt), that Nursery
contraction triggers object movement (sliding), refuse to contract, if
Concurrent Scavenger (CS) is enabled.

Firstly, it may cause significant pauses, negating the low pause
benefits of CS.
Secondly, and more importantly, it's not safe to do in presence of CS.
The area being moved is hybrid Survivor and Allocate. Objects that have
just been allocated during concurrent phase of the cycle, and have been
hashed, would need to grow (by a hash slot). The current reference fix
up logic does not account of that, since STW Scavenger would move only
pure Survived objects, that already grew, if really needed.

Adjusting the fix up logic is not trivial if possible at all (would
probably need an additional pass to figure out what is to grow).
Adjusting tilt ratio to contract by smaller amount so to avoid object
movement, is probably possible, but messy and semantically not really
that important - it's just simpler to defer the contract attempt for
another Scavenger cycle (if we are very little tilted, we should
probably end up expanding, not contracting).
In a forced case of very limited or disabled tilting, however, it may
mean that we may not ever (or rather rarely) end up contracting Nursery.

### DIFF
--- a/gc/base/standard/PhysicalSubArenaVirtualMemorySemiSpace.cpp
+++ b/gc/base/standard/PhysicalSubArenaVirtualMemorySemiSpace.cpp
@@ -547,6 +547,19 @@ MM_PhysicalSubArenaVirtualMemorySemiSpace::contract(MM_EnvironmentBase *env, uin
 
 		/* Check if the move location actually does in fact move any valid heap */
 		if(allocateSegmentBase > allocateLeadingFreeTop) {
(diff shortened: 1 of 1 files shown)


charliegracie:
@genie-omr build all

PR #3684, opened 2019-03-22
Change known obj constraint to block constraint

cathyzhyi:
A value number is known obj in a block doesn't mean it's known object
every where in the program.


andrewcraik:
looks good - the notion being a guard or other condition may allow us to know a VN is equal to a known object in a give block when we don't know that to be the case globally. FYI @vijaysun-omr

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Looks like a simple change to differentiate between the case when the known object property is only known locally in a block (due to a comparison test) vs known globally.

Issue #3674, opened 2019-03-18
Incorrect unsigned division special case on Z


aviansie-ben:
Currently, the Z codegen's implementations of `iudiv`, `iurem`, and `ludiv` incorrectly apply a special case for handling `INT_MIN / -1`. This special case is meant to ensure that `INT_MIN / -1 = INT_MIN` and `INT_MIN % -1 = 0` as required by the Java spec, however this logic is being erroneously applied to unsigned division.

This can be seen by the test failures encountered when adding tril tests for these opcodes in #3637 at https://ci.eclipse.org/omr/job/PullRequest-linux_390-64/1383/consoleFull.

fjeremic:
Do you have a trace for the failure? i.e. what the trees look like in one of the failing cases?

Leonardo2718:
Here's a log for an iudiv test (`--gtest_filter="DivArithmeticTest/UInt32Arithmetic.UsingLoadParam/410"`)
[iudiv.trace.log](https://github.com/eclipse/omr/files/2988842/iudiv.trace.log)

And here's one for an iurem test  (`--gtest_filter="DivArithmeticTest/UInt32Arithmetic.UsingLoadParam/411"`)
[iurem.trace.log](https://github.com/eclipse/omr/files/2988845/iurem.trace.log)

Issue #3685, opened 2019-03-22
Remove Java-ism in div evaluators on Z


fjeremic:
As part of #3674 we've identified some "extra" checks in our `*div` evaluators which are needed for Java. Unfortunately they apply to OMR today which is not really what we want. This code needs to be pushed out of OMR and into OpenJ9.

Code in question can be found in [1], [2], and [3].

[1] https://github.com/eclipse/omr/blob/3e2bef8e9e0840ef3ddb61abc3f2bf5a8b4e0bab/compiler/z/codegen/BinaryEvaluator.cpp#L1040-L1050

[2] https://github.com/eclipse/omr/blob/3e2bef8e9e0840ef3ddb61abc3f2bf5a8b4e0bab/compiler/z/codegen/BinaryEvaluator.cpp#L904-L924

[3] https://github.com/eclipse/omr/blob/3e2bef8e9e0840ef3ddb61abc3f2bf5a8b4e0bab/compiler/z/codegen/BinaryEvaluator.cpp#L741

hzongaro:
I tripped across code that mentions this issue.  Is this really a Java-ism, or should we expect that the following IL will produce the same result on all platforms &mdash; for OMR itself and for all downstream projects &mdash; without an exception being raised?

```
n99 idiv
n98   iconst 0x80000000
n97   iconst -1
```

PR #3686, opened 2019-03-22
Fix handling of unsigned 32-bit and 64-bit division on Z

fjeremic:
There is some Java-ism built into the Z *div evaluators in which we
special case a few corner cases for Java. Technically this code does
not belong in OMR, however that is a separate issue (#3685).

In this commit we disable the need to generate such checks for unsigned
division.

Fixes: #3674

### DIFF
--- a/compiler/z/codegen/BinaryEvaluator.cpp
+++ b/compiler/z/codegen/BinaryEvaluator.cpp
@@ -799,7 +799,10 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
          if (isDivision)
             {
             firstRegister = cg->gprClobberEvaluate(firstChild);
-            if (!firstChild->isNonNegative())
+
+            // This adjustment only applies to signed division as we are effctively trying to skirt the sign bit during
+            // the shift operation below
+            if (!isUnsigned && !firstChild->isNonNegative())
                {
                TR::LabelSymbol * cFlowRegionStart = generateLabelSymbol(cg);
                TR::LabelSymbol * skipSet = generateLabelSymbol(cg);
@@ -832,7 +835,10 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
 
             TR::LabelSymbol * done = generateLabelSymbol(cg);
             TR::RegisterDependencyConditions *deps = NULL;
-            if (!firstChild->isNonNegative())
+
+            // This adjustment only applies to signed division as we are effctively trying to skirt the sign bit during
+            // the shift operation below
+            if (!isUnsigned && !firstChild->isNonNegative())
                {
                TR::Register * tempRegister1 = cg->allocateRegister();
                TR::Register * tempRegister2 = cg->allocateRegister();
@@ -901,7 +907,8 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
    dependencies->addPostCondition(remRegister, TR::RealRegister::LegalEvenOfPair);
    dependencies->addPostCondition(quoRegister, TR::RealRegister::LegalOddOfPair);
 
-   if (!doConditionalRemainder &&
+   // TODO: Remove this Java-ism from OMR and push it into OpenJ9
+   if (!isUnsigned && !doConditionalRemainder &&
          !node->isSimpleDivCheck() &&
          !firstChild->isNonNegative() &&
          !secondChild->isNonNegative() &&
@@ -1045,7 +1052,8 @@ iDivRemGenericEvaluator(TR::Node * node, TR::CodeGenerator * cg, bool isDivision
    //    R1+1 -> Quotient
    //
    // Note:  D and DR require that R1=0 before executing the division
-   bool needCheck = (!node->isSimpleDivCheck()            &&
+   // TODO: Remove this Java-ism from OMR and push it into OpenJ9
+   bool needCheck = (!node->getOpCode().isUnsigned() && !node->isSimpleDivCheck()            &&
                      !firstChild->isNonNegative()         &&
                      !secondChild->isNonNegative());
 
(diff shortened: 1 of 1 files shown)


fjeremic:
@aviansie-ben I've confirmed this fixes all the Z failures in #3637. I'm leaving this in draft mode until I've verified we don't regress OpenJ9 accidentally.

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build all

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -901,7 +907,8 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
    dependencies->addPostCondition(remRegister, TR::RealRegister::LegalEvenOfPair);
    dependencies->addPostCondition(quoRegister, TR::RealRegister::LegalOddOfPair);
 
-   if (!doConditionalRemainder &&
+   // TODO: Remove this Java-ism from OMR and push it into OpenJ9
```
0xdaryl: If this is covered by #3685 then why not mention the issue here rather than a TODO?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -1045,7 +1052,8 @@ iDivRemGenericEvaluator(TR::Node * node, TR::CodeGenerator * cg, bool isDivision
    //    R1+1 -> Quotient
    //
    // Note:  D and DR require that R1=0 before executing the division
-   bool needCheck = (!node->isSimpleDivCheck()            &&
+   // TODO: Remove this Java-ism from OMR and push it into OpenJ9
```
0xdaryl: If this is covered by #3685 then why not mention the issue here rather than a TODO?

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -832,7 +835,10 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
 
             TR::LabelSymbol * done = generateLabelSymbol(cg);
             TR::RegisterDependencyConditions *deps = NULL;
-            if (!firstChild->isNonNegative())
+
+            // This adjustment only applies to signed division as we are effctively trying to skirt the sign bit during
```
0xdaryl: sp: effectively

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -799,7 +799,10 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
          if (isDivision)
             {
             firstRegister = cg->gprClobberEvaluate(firstChild);
-            if (!firstChild->isNonNegative())
+
+            // This adjustment only applies to signed division as we are effctively trying to skirt the sign bit during
```
0xdaryl: sp: effectively

fjeremic: Fixed in 7e888b5

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Fixed in 7e888b5

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Ah yes, the issue was opened after-the-fact. Totally forgot to go back and update the reference in the code. Fixed in 7e888b5

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Ah yes, the issue was opened after-the-fact. Totally forgot to go back and update the reference in the code. Fixed in 7e888b5

0xdaryl:
@genie-omr build zos,zlinux

This is a Z-specific change.  A full genie build is not required.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -901,7 +907,8 @@ lDivRemGenericEvaluator64(TR::Node * node, TR::CodeGenerator * cg, bool isDivisi
    dependencies->addPostCondition(remRegister, TR::RealRegister::LegalEvenOfPair);
    dependencies->addPostCondition(quoRegister, TR::RealRegister::LegalOddOfPair);
 
-   if (!doConditionalRemainder &&
+   // TODO(#3686): Remove this Java-ism from OMR and push it into OpenJ9
```
0xdaryl: Sorry to be a pest.  The issue number is actually #3685.  This PR is #3686.  Ditto for the other comment below.

compiler/z/codegen/BinaryEvaluator.cpp
```diff
@@ -1045,7 +1052,8 @@ iDivRemGenericEvaluator(TR::Node * node, TR::CodeGenerator * cg, bool isDivision
    //    R1+1 -> Quotient
    //
    // Note:  D and DR require that R1=0 before executing the division
-   bool needCheck = (!node->isSimpleDivCheck()            &&
+   // TODO(#3686): Remove this Java-ism from OMR and push it into OpenJ9
+   bool needCheck = (!node->getOpCode().isUnsigned() && !node->isSimpleDivCheck()            &&
```
mstoodle: pretty minor point: wouldn't it make more sense in this change and the previous one to add the `!isUnsigned` check next to the `!isNonNegative` check rather than adding it "next" to whatever the first condition happens to be?

mstoodle:
> I'm leaving this in draft mode until I've verified we don't regress OpenJ9 accidentally.

@fjeremic I don't see any WIP, so I think you've done it, but can you please confirm you've verified this point?

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Fixed in 36995ad

compiler/z/codegen/BinaryEvaluator.cpp
fjeremic: Fixed in 36995ad

fjeremic:
> @fjeremic I don't see any WIP, so I think you've done it, but can you please confirm you've verified this point?

Yes, but I'd like to launch another one as I'm always paranoid. Leaving WIP until it's complete. Feel free to do a final review.

0xdaryl:
@genie-omr build zos,zlinux

fjeremic:
Looks good. Removing WIP.

PR #3688, opened 2019-03-23
Prepare to consolidate trampoline space reservation functions

0xdaryl:
Consolidate `CodeCache::reserveSpaceForTrampoline` and
`CodeCache::reserveNTrampolines` into a single function.  The
error handling functionality from `reserveNTrampolines` will
be incorporated in the merged function as current consumers
of `reserveSpaceForTrampoline` performed the same functionality
now anyway.

The merged function does not unreserve the code cache on failure
(as the current `reserveNTrampolines` function does) as that
functionality is beyond the scope of this function.  Downstream
consumers will need to be modified if they're expecting that
behaviour.

This PR introduces a temporary "bridge" function while necessary
downstream changes are made.  It will become the permanent function
when such changes are complete.


0xdaryl:
@genie-omr build all

PR #3689, opened 2019-03-23
Add misc CodeCache getters and setters

0xdaryl:
For the following fields:

* _tempTrampolineBase
* _tempTrampolineTop
* _tempTrampolineNext
* _trampolineAllocationMark
* _trampolineReservationMark
* _trampolineBase

### DIFF
--- a/compiler/runtime/OMRCodeCache.hpp
+++ b/compiler/runtime/OMRCodeCache.hpp
@@ -313,6 +313,106 @@ public:
     */
    void setFreeBlockList(CodeCacheFreeCacheBlock *fcb) { _freeBlockList = fcb; }
 
+   /**
+    * @brief Getter for the base address of temporary trampolines
+    *
+    * @returns The base address of temporary trampolines
+    */
+   uint8_t *tempTrampolineBase() { return _tempTrampolineBase; }
+
+   /**
+    * @brief Setter for the base address of temporary trampolines
+    *
+    * @param[in] address : The new base address of temporary trampolines
+    */
+   void setTempTrampolineBase(uint8_t *address) { _tempTrampolineBase = address; }
+
+   /**
+    * @brief Getter for the top address of temporary trampolines
+    *
+    * @returns The top address of temporary trampolines
+    */
+   uint8_t *tempTrampolineTop() { return _tempTrampolineTop; }
+
+   /**
+    * @brief Setter for the top address of temporary trampolines
+    *
+    * @param[in] address : The new top address of temporary trampolines
+    */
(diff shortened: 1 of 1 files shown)


0xdaryl:
@genie-omr build all

compiler/runtime/OMRCodeCache.hpp
```diff
@@ -313,6 +313,106 @@ class OMR_EXTENSIBLE CodeCache : public CodeCacheBase
...
+    */
+   void setTrampolineAllocationMark(uint8_t *address) { _trampolineAllocationMark = address; }
+
+   /**
+    * @brief Adjusts the trampoline allocation mark by the specified amount
+    *
+    * @param[in] numBytes : The number of bytes to adjust the allocation mark by.  Negative
+    *              values are permitted
+    */
+   void adjustTrampolineAllocationMark(size_t numBytes) { _trampolineAllocationMark += numBytes; }
```
Leonardo2718: The documentation says that negative values for `numBytes` are permitted. However, `size_t` is defined as an unsigned type, so it cannot have a negative value. :slightly_smiling_face: 

The fix is probably to just use a different type.

compiler/runtime/OMRCodeCache.hpp
```diff
@@ -313,6 +313,106 @@ class OMR_EXTENSIBLE CodeCache : public CodeCacheBase
...
+    */
+   void setTrampolineReservationMark(uint8_t *address) { _trampolineReservationMark = address; }
+
+   /**
+    * @brief Adjusts the trampoline reservation mark by the specified amount
+    *
+    * @param[in] numBytes : The number of bytes to adjust the reservation mark by.  Negative
+    *              values are permitted
+    */
+   void adjustTrampolineReservationMark(size_t numBytes) { _trampolineReservationMark += numBytes; }
```
Leonardo2718: Same comment as for `adjustTrampolineAllocationMark`.

compiler/runtime/OMRCodeCache.hpp
0xdaryl: Fixed in f7ac529

0xdaryl:
@genie-omr build all

vijaysun-omr:
This obviously seems like a good change. I was just wondering if there was a reason you were thinking of adding in these specific APIs in question.

0xdaryl:
The trampoline code is being refactored out of the CodeCache class, but some remnants will be left behind (at least for now).  Those remnants are mostly the pointers in the code cache that control the dimensions and current allocation pointers to various trampoline structures.  To get access to those remnants I need getters/setters for them that can be used by a new, more generic trampoline manager class.

PR #1579, opened 2017-08-29
CMake improvements

charliegracie:
- Refactor the glue into multiple smaller libraries.
-- This drastically cleans up the dependency graph.
- Add platform support for Linux z, Linux P, AIX and OSX

PR #3693, opened 2019-03-25
Migrate z/OS to a CMake build system

fjeremic:
This PR adds support for building JitBuilder on z/OS. It has various fixes and some workarounds for issues either related to xlc toolchain not supporting C++11 or z/OS workarounds for unsupported APIs. Note we cannot actually run JitBuilder tests at the moment because the JIT needs to implement proper XPLINK linkage which we have somehow lost in the refactoring process.

This PR simply gets us into a state where we can build JitBuilder. It will enable us to hook up the build process to our PR testing to ensure developers are not checking in C++11 code which does not work on z/OS. In parallel I'll be working on adding proper XPLINK support back so that we can start executing tests as well.

The PR is broken up into many commits, each of which was a hurdle I had to overcome in the development process. Each commit description outlines why the change had to be made. I'll now start off with some self-reviews for major hacks I had to place in which require changes outside of my expertise for which I'll need help with.

Fixes: #3710


fjeremic:
Also requesting a review from @dnakamura.

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
```diff
@@ -227,10 +227,10 @@ JavaBlobGenerator::stringTableOffset(BlobHeader *blobHeader, J9HashTable *string
 	DDR_RC rc = DDR_RC_OK;
 	StringTableEntry exemplar;
 
-	*offset = UINT32_MAX;
+	*offset = U_32_MAX;
```
fjeremic: While this works and is an ok change I think we should think about building non-compiler sources on z/OS with `-D_ISOC99_SOURCE` which I believe overcomes these issues. However such a change is outside the scope of this PR.

util/omrutil/CMakeLists.txt
```diff
@@ -27,16 +27,6 @@ omr_assert(
...
-#TODO port following vpath code:
-#ifeq (zos,$(OMR_HOST_OS))
-	#OBJECTS += zbarrier$(OBJEXT)
-	#ifeq (0,$(OMR_ENV_DATA64))
-		#vpath zbarrier.s unix/zos/31
-	#else
-		#vpath zbarrier.s unix/zos/64
-	#endif
-#endif
-
```
fjeremic: @dnakamura for review for these. I think they are addressed below.

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
Leonardo2718: In other parts of the code, we've been replacing uses of `UINT32_MAX` with `std::numeric_limits<uin32_t>::max()` from the `<limits>` header. It's pre-C++11 and I'm fairly certain it's supported by all our build compilers.

compiler/env/FEBase_t.hpp
```diff
@@ -65,6 +65,13 @@ FEBase<Derived>::allocateRelocationData(TR::Compilation* comp, uint32_t size)
             size,
             MEM_COMMIT,
             PAGE_READWRITE));
+// TODO: Why is there no OMR_OS_ZOS? Or any other OS for that matter?
+#elif defined(J9ZOS390)
+   // TODO: This is an absolute hack to get z/OS JITBuilder building and even remotely close to working. We really
+   // ought to be using the port library to allocate such memory. This was the quickest "workaround" I could think
+   // of to just get us off the ground.
+   return reinterpret_cast<uint8_t *>(
+         malloc(size));
```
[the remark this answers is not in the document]
mstoodle: Created #3694 as this stuff should really be handled more properly which would eliminate the need for such a hack.

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
fjeremic: Unfortunately `std::numeric_limits<uint32_t>::max()` does not work on z/OS unless `-D_ISOC99_SOURCE` which is currently not the case for non-compiler C++ sources which is why we need to look into this (outside the scope of this PR).

fjeremic: Example:

```
[ 42%] Building CXX object ddr/lib/ddr-blobgen/CMakeFiles/omr_ddr_blobgen.dir/java/genBinaryBlob.cpp.o
"/trjit/sandbox/jeremif/developement/jitbuilder-zos/omr/ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp", line 230.19: CCN5130 (S) "std::numeric_limits" is not declared.
```

fjeremic:
@genie-omr build all

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
Leonardo2718: I see. :disappointed:

charliegracie: I would prefer to add that option to the DDR code instead of changing the code

charliegracie:
I would also like to see working zOS builds on the OMR farm before moving too far forward with these changes since I can not verify they work in a normal zOS environment.  Also I believe there are 1 or 2 PRs that should be merged before this. I think there is a PR to resolve the return values after TR_UNIMPLEMENTED as a minimum

fjeremic:
> I would also like to see working zOS builds on the OMR farm before moving too far forward with these changes since I can not verify they work in a normal zOS environment. Also I believe there are 1 or 2 PRs that should be merged before this. I think there is a PR to resolve the return values after TR_UNIMPLEMENTED as a minimum

I agree on both fronts. I'll work with the owners of the dependent PRs to move them along.

fjeremic:
@genie-omr build all

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
fjeremic: I don't think it's as easy as adding the flag as we currently have no separation of "DDR only defines" so we would have to add it to the global defines, but then that causes problems elsewhere with with C++ vs C99.

That is really the direction we want to end up in, i.e. we use the same defines and same flags to build all of OMR. Unfortunately we are not there yet. I believe such a change is outside the scope of what we're trying to accomplish in this PR.

As a compromise I've created issue #3710 which will track this work.

compiler/env/TRMemory.hpp
```diff
@@ -795,6 +795,9 @@ TR_HeapMemory::allocate(size_t size, TR_MemoryBase::ObjectType ot)
    void operator delete[](void * p, TR::Region &region) { region.deallocate(p); } \
    static TrackedPersistentAllocator getPersistentAllocator() { return TrackedPersistentAllocator(); }
 
+// TODO: This is actually handled by eclipse/omr#3596 which is in a WIP state. We should verify what is going on with
+// that PR and likely remove this line if it gets merged. For now leaving it in here as we cannot compile without it.
+#undef TR_ALLOC
```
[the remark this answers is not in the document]
fjeremic: I've removed the commit (dcc13463) now that #3596 has been merged.

fjeremic:
Removed the workaround due to #3956 being merged. I've verified everything works as expected now.

@genie-omr build all

mstoodle:
@fjeremic thanks for all the effort here!

Like @charliegracie, I would like to wait until the z/OS machines are configured so that we can actually activate the compiler/jitbuilder builds as part of merging these changes. As it stands, I believe we've only verified (at this point) that these changes do not affect any of the non-compiler componentry building on z/OS.

I guess that means getting an appropriate version of `cmake` installed and operational on our z/OS machines so that these changes can be activated to actually build the jitbuilder library?

fjeremic:
> I guess that means getting an appropriate version of `cmake` installed and operational on our z/OS machines so that these changes can be activated to actually build the jitbuilder library?

Yes, I agree. We've actually open sourced the z/OS support for CMake 3.5.1 [1]. I'm currently working on bootstrapping `cmake` on our OMR z/OS build machines. Currently running into some infrastructure issues with `/tmp` so I'm working with IT to resolve those. I'll update once I have CMake working on the farm.

This PR will receive another commit which enables CMake tests on the build farm.

[1] https://github.com/fjeremic/CMake/

fjeremic:
Just a short update on this. I've been working with IT behind the scenes to install various tooling we'll need to get these builds going on the Eclipse OMR z/OS machines hooked up for PR testing. I've now got this PR building completely on the PR machines. The Jenkins user is now setup for CMake builds as well. Next step will be the addition of a new commit which will enable CMake z/OS builds. I'll be working on this in the next day or two.

fjeremic:
So I ended up going down a deep rabbit hole to properly implement XPLINK linkage on z/OS so we can actually run the tests that we build here. With the change in #3779 along with the one here we can now from end-to-end build and run all the JitBuidler tests.

Before going back to polish off the system linkage changes further I'll take pause now and get back to working on this PR to hook it up to the Jenkins builds. Once #3779 is merged this PR should follow which will enable the actual testing.

fjeremic:
I just realized that to get the CMake builds setup we require a prerequisite step to actually host the pipelines in the repository itself. I've opened up #3784 to address this. Once that PR is merged we can add a commit here which modifies `buildenv/jenkins/jobs/pull-requests/PullRequest-zos_390-64` to start taking advantage of CMake to build JitBuilder.

ddr/lib/ddr-blobgen/java/genBinaryBlob.cpp
fjeremic: This got fixed by @keithc-ca in #3790.

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build zos

fjeremic:
@genie-omr build all

fjeremic:
@charliegracie @Leonardo2718 @dnakamura @mstoodle @0xdaryl @youngar

I've added 23 new commits and have converted the entire OMR build on z/OS to CMake. The genie build launched as part of the PR is now running on the new CMake build for z/OS. As part of this migration we are now using an almost identical Jenkins pipeline file for z/OS as we have on Linux. That means we are now compiling and linking all the OMR libraries on z/OS (including JitBuilder, Tril, CompilerTest, CompilerTrilTest, Port, RAS, etc.).

We in fact able to run all the tests via `ctest` as well however there are several things failing since we have never ran them before. Instead of fixing these now I have found this to be a sane checkpoint to stop and get this merged as is since I've migrated all the currently running tests in the old Makefile builds to run on the new CMake build (omralgotest , omrvmtest, omrutiltest, omrsigtest, omrrastest, omrsubscribertest, omrtraceoptiontest). Once we are happy with this PR and we have started using the CMake builds I'll be working on fixing up the rest of OMR so we can run all the tests we run on Linux.

We are now one major step closer to a full CMake migration at the OMR project. I will over the next day proceed to self-review my PR to explain some of the decisions made, however I would appreciate if you could review the changes within your subject matter domain so we can move forward. Thank you very much!

fjeremic:
@genie-omr build all

cmake/modules/platform/os/zos.cmake
```diff
@@ -20,12 +20,17 @@
...
 	-D_ALL_SOURCE
-	-D_XOPEN_SOURCE_EXTENDED
-	-DIBM_ATOE
+	-D_OPEN_THREADS=2
 	-D_POSIX_SOURCE
+	-D_XOPEN_SOURCE_EXTENDED
+	-D_ISOC99_SOURCE
+	-DLONGLONG
+	-DJ9ZOS390
+	-DOMR_EBCDIC
```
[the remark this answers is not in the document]
keithc-ca: The module option `USE_NATIVE_ENCODING` should select either `-DIBM_ATOE` or `-DOMR_EBCDIC` like https://github.com/eclipse/omr/blob/master/omrmakefiles/rules.zos.mk#L82 does.

fvtest/compilertest/tests/OptionSetTest.hpp
```diff
@@ -29,7 +29,7 @@
 
 namespace TestCompiler {
 
-typedef std::map<const std::string, int> MethodSets;
+typedef std::map<std::string, int> MethodSets;
```
Leonardo2718: I think a previous commit in this PR added the `const`. Could the history be changed so the `const` isn't added in the first place?

cmake/modules/FindLEX.cmake
```diff
@@ -0,0 +1,73 @@
+find_program(LEX_EXECUTABLE NAMES lex DOC "path to the lex executable")
```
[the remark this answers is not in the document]
fjeremic: Flex does not exist on z/OS so it is not an option. Lex is POSIX so it is available on z/OS as well as all other platforms OMR supports.

cmake/modules/FindYACC.cmake
```diff
@@ -0,0 +1,86 @@
+find_program(YACC_EXECUTABLE NAMES yacc DOC "path to the YACC executable")
```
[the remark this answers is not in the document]
fjeremic: Similarly to Flex, Bison does not exist on z/OS. It is available however through Rocker Software open source tools, however it is a super old version. Yacc is POSIX and available on all platforms OMR supports.

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
fjeremic: As far as I gather from this comment from @charliegracie (https://github.com/eclipse/omr/issues/1830#issuecomment-340785443) and my experience implementing this PR is that the only reason `atoe.h` and that whole utility exists is because the project was not previously hosted on Git and files were not properly tagged via `chtag` to specify whether they're ASCII or EBCDIC.

This combined with specifying the `CONVLIT` compiler option runs us into the weird state of ASCII/EBCDIC where we need to do runtime conversions. With Git on z/OS all files are properly filetagged as ASCII on checkout and the existing tools (via auto-conversion) transparently just work.

I too like @charliegracie hope to one day soon eliminate that utility once downstream projects start consuming this PR and building the source by checking out via Git.

cmake/modules/FindLEX.cmake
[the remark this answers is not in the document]
dnakamura: I meant more along the line of tricking/forcing cmake into accepting lex as  flex so that we can avoid having a duplicate `[F]LEX_TARGET()`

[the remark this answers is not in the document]
fjeremic: This is not possible unfortunately as `lex` and `flex` are not command line option compatible.

util/a2e/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ target_include_directories(j9a2e
 
 add_custom_command(TARGET j9a2e
 	POST_BUILD
-	COMMAND mv j9a2e.x ${CMAKE_BINARY_DIR}/j9a2e.x
+	COMMAND mv libj9a2e.x ${CMAKE_BINARY_DIR}/libj9a2e.x
```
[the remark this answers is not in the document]
fjeremic: Do you have a suggestion of how to properly do this?

jitbuilder/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@
 
 include(OmrCompilerSupport)
 
-find_package(PythonInterp 3.1 REQUIRED)
+find_package(PythonInterp REQUIRED)
```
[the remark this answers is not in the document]
Leonardo2718: nvm. Done in a different commit.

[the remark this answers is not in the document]
fjeremic: Yeah sorry I modified way too many files to go back to each individual one and squash the copyright fix in there. Just avoided the churn and did it all in one go at the end.

cmake/modules/FindLEX.cmake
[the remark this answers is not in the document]
fjeremic: Also I don't think we have duplicate targets. The commits here merely migrate OMR to use `lex` and `yacc` instead of `flex` and `bison` since the former are supported on all POSIX systems. There should be no usage of the latter following this change.

In fact I think we can likely contribute FindLEX.cmake and FindYACC.cmake to upstream CMake project.

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
fjeremic: > I'm hesitant about enabling OMR_EBCDIC by default, as it is largely untested

It is running testing as part of this PR and seems to work fine.

cmake/modules/platform/os/zos.cmake
```diff
@@ -20,12 +20,17 @@
...
+	-DSUPPORTS_THREAD_LOCAL
+	-DUT_DIRECT_TRACE_REGISTRATION
```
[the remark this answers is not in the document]
fjeremic: Fixed in dafd9e0834201cb817f6e552091c04d2dade6410

cmake/modules/platform/os/zos.cmake
```diff
@@ -20,12 +20,17 @@
 #############################################################################
 
 list(APPEND OMR_PLATFORM_DEFINITIONS
-	-DJ9ZOS390
-	-DLONGLONG
 	-D_ALL_SOURCE
-	-D_XOPEN_SOURCE_EXTENDED
-	-DIBM_ATOE
+	-D_OPEN_THREADS=2
```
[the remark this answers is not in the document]
fjeremic: It is already defined as part of OpenJ9 [1] and a few other places in OMR port library. I think we should be fine here.

[1] https://github.com/eclipse/openj9/blob/2673d23e5c7525b80254356df1635565f1642ef0/runtime/compiler/build/toolcfg/zos-xlc/common.mk#L75

cmake/modules/platform/toolcfg/xlc.cmake
```diff
@@ -19,30 +19,145 @@
...
+		-qfuncsect
+		-qsuppress=1540-1087:1540-1088:1540-1090:1540-029:1500-029
+		-qdebug=nscrep
 	)
-else()
-	#-qarch should be there for 32 and 64 C/CXX flags but the C compiler is used for the assembler and it has trouble with some assembly files if it is specified
+
+	# Configure the platform dependent library for multithreading
+	set(OMR_PLATFORM_THREAD_LIBRARY -lpthread)
+elseif(OMR_HOST_ARCH STREQUAL "s390")
```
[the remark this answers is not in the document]
fjeremic: This is in the xlc CMake tool configuration. Although I cannot imagine why you would attempt to build OMR on Linux with xlc it technically is an option. I don't see why the same configuration we use for xlc on z/OS is not wanted on xlc for Linux.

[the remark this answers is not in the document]
fjeremic: Note OMR currently only supports building with gcc on Linux on z.

util/a2e/CMakeLists.txt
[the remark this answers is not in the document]
fjeremic: I was actually copying already existing similar code [1] which was checked in as part of #1579. Do you mean to modify it to something like this:

```
ADD_CUSTOM_COMMAND(TARGET j9a2e
    POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E rename libj9a2e.x ${CMAKE_BINARY_DIR}/libj9a2e.x
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
)
```

If so I can update both the existing locations to do it the new way. Let me know what you prefer.

[1] https://github.com/eclipse/omr/blob/051896be071812db8a3b71485b014be26c80e7f4/util/a2e/CMakeLists.txt#L52-L56

cmake/modules/FindLEX.cmake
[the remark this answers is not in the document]
fjeremic: Fixed in 4f12625063e31bfa59ab4f2721fdc77cfad34eb2

cmake/modules/FindYACC.cmake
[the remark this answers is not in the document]
fjeremic: Fixed in 4f12625063e31bfa59ab4f2721fdc77cfad34eb2

cmake/modules/platform/toolcfg/xlc.cmake
```diff
@@ -61,32 +176,13 @@ if(OMR_HOST_OS STREQUAL "aix")
 		set(CMAKE_C_ARCHIVE_CREATE "<CMAKE_AR> -X64 cr <TARGET> <LINK_FLAGS> <OBJECTS>")
 		set(CMAKE_C_ARCHIVE_FINISH "<CMAKE_RANLIB> -X64 <TARGET>")
 	endif()
-
+
```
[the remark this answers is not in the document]
fjeremic: Fixed in 4f12625063e31bfa59ab4f2721fdc77cfad34eb2

fvtest/tril/tril/CMakeLists.txt
```diff
@@ -23,18 +23,17 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../cmake/modules"
...
 
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-list(APPEND TRIL_FLEX_FLAGS
-	--yylineno
-	--header-file=${CMAKE_CURRENT_BINARY_DIR}/tril.scanner.h
-)
+list(APPEND TRIL_FLEX_FLAGS)
```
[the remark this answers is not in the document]
fjeremic: We do seem to append `--wincompat` under Windows OS a few lines below so it is needed I think.

fvtest/tril/tril/simple_compiler.cpp
```diff
@@ -73,7 +73,23 @@ int32_t Tril::SimpleCompiler::compileWithVerifier(TR::IlVerifier* verifier) {
...
+       {
+          uint64_t environment;
+          void* func;
+       };
+
+       FunctionDescriptor* fd = new FunctionDescriptor();
+       fd->environment = 0;
+       fd->func = entry_point;
+
+       entry_point = (uint8_t*) fd;
```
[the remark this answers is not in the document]
fjeremic: Indeed it is quite scary. We had a similar thing checked in for JitBuilder (https://github.com/eclipse/omr/pull/3779#discussion_r278331680) and an accompanying discussion with @mstoodle. Unfortunately impleemnting this properly at the compiler level is going to require thought from the JitBuilder/API perspective to properly design. I believe this is outside the scope of what we're trying to accomplish here. Do you think we need an issue to track this?

third_party/gtest-1.8.0/include/gtest/internal/gtest-port.h
```diff
@@ -689,6 +689,22 @@ struct _RTL_CRITICAL_SECTION;
 #  include "gtest/internal/gtest-tuple.h"  // IWYU pragma: export  // NOLINT
 # elif GTEST_ENV_HAS_STD_TUPLE_
 #  include <tuple>
+
+#if defined(J9ZOS390)
```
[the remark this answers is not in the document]
fjeremic: Yes this is the only modification. There were previous modifications in this area as well in the current codebase [1] as part of #1429. It seems we did a CQ at the time because it was a large contribution. I don't suspect we need one here. I would however assume that there would be no licensing implications given the originally Google Test code was changed with the addition of [1] and that passed the CQ already.

[1] https://github.com/eclipse/omr/blob/7574eec990ab5d396bcec38b02a160f9381e8b55/third_party/gtest-1.8.0/include/gtest/internal/gtest-port.h#L2277-L2296

fvtest/tril/examples/incordec/main.cpp
```diff
@@ -41,7 +41,7 @@ int main(int argc, char const * const * const argv) {
     printTrees(stdout, trees, 0);
 
     // assume that the file contians a single method and compile it
-    Tril::DefaultCompiler incordecCompiler{trees};
+    Tril::DefaultCompiler incordecCompiler(trees);
```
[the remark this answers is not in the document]
fjeremic: Fixed and squashed the two into the new commit 9a077fb35

compiler/z/codegen/OMRInstOpCodeProperties.hpp
```diff
@@ -1033,7 +1033,7 @@
    /* .description = */ "BRANCH ON CONDITION",
    /* .opcode[0]   = */ 0x47,
    /* .opcode[1]   = */ 0x00,
-   /* .format      = */ RXb_FORMAT,
+   /* .format      = */ RXa_FORMAT,
```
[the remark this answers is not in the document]
fjeremic: I agree. I've opened #3900 to track this issue.

fvtest/compilertest/tests/OptionSetTest.hpp
fjeremic: I don't think this is the case. The `const` exists in the master branch [1].

[1] https://github.com/eclipse/omr/blob/051896be071812db8a3b71485b014be26c80e7f4/fvtest/compilertest/tests/OptionSetTest.hpp#L32

omrtrace/omrtracewrappers.cpp
```diff
@@ -56,7 +56,7 @@ twThreadSelf(void)
 omr_error_t
 twE2A(char *str)
 {
-#if defined(J9ZOS390)
+#if defined(J9ZOS390) && !defined(OMR_EBCDIC)
```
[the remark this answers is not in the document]
fjeremic: I'm not sure I see the value of decoupling here. As far as I could tell from the existing uses we are trying to accomplish the same thing, i.e. assume native encoding and do not do any runtime ASCII/EBCDIC conversions which is exactly what we're trying to accomplish in the runtime. If files are properlly filetagged with `chtag` (which they are automatically if checked out via Git for z/OS) we do not need any runtime conversions between ASCII/EBCDIC and we should use the platform native APIs.

fjeremic:
@Leonardo2718 @dnakamura I have addressed all the review comments thus far. There are some questions and uncertainties I've noted which require your input. I'll be awaiting further reviews. Thanks for investing the time to review this PR!

@genie-omr build all

third_party/gtest-1.8.0/include/gtest/internal/gtest-port.h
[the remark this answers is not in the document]
Leonardo2718: I see. That answered my question. Thanks.

fvtest/compilertest/tests/OptionSetTest.hpp
Leonardo2718: Hm. You're right. I must be misremembering something.

fjeremic:
@Leonardo2718 @dnakamura any other concerns which were not addressed above?

fjeremic:
@Leonardo2718 @dnakamura kind reminder for above. This work is blocking progression for the rest of the build. Also would be good if a committer was assigned for this one @0xdaryl @Leonardo2718 @charliegracie any takers?

compiler/ilgen/OMRMethodBuilder.cpp
```diff
@@ -620,7 +620,7 @@ OMR::MethodBuilder::getParameterTypes()
       TR_ASSERT_FATAL(symNamesIterator != _symbolNameFromSlot.end(), "No symbol found in slot %d", p);
       const char *name = symNamesIterator->second;
 
-      std::map<const char *, TR::IlType *, StrComparator>::iterator symTypesIterator = _symbolTypes.find(name);
+      auto symTypesIterator = _symbolTypes.find(name);
```
Leonardo2718: For the record, I don't think the old code is necessarily "valid C++" as is suggested by the commit message. The type of `_symbolTypes` is `SymbolTypeMap`, which is a typedef for 

```c++
std::map<const char *, TR::IlType *, StrComparator, SymbolTypeMapAllocator>
```

Note how the custom allocator `SymbolTypeMapAllocator` is missing from:

```c++
std::map<const char *, TR::IlType *, StrComparator>::iterator symTypesIterator = _symbolTypes.find(name);
```

Not specifying the type of the allocator means it will be defaulted to the standard allocator. The different allocator types means the two `std::map` template instances are technically different types. So, the code is actually trying to assign an iterator from a container of one type to an interator of a container of a different type, which is undefined behaviour.

A possible reason the code works with the other compilers is that the allocator type isn't used in the definition of the iterator type in those STL implementations.

An alternative fix that should also work is to make `symTypesIterator`'s type be `SymbolTypeMap::iterator`. However, I am fine with using `auto` as is done in this change.

fvtest/tril/tril/simple_compiler.cpp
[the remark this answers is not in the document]
Leonardo2718: I see, thanks.

Yes, we should definitely open an issue to track this. I agree that trying to resolve this now is far outside the scope of this PR.

Leonardo2718:
As @fjeremic mentioned, this PR is blocking progress on other work, so I am going to merge this at the end of this week if no one has any further comments.

FYI @dnakamura

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
dnakamura: but that applies it to just the jit, not the entirety of omr/openj9, however looking at the docs, it seems it just exposes additional functions rather than changing behavior of any existing functions, so this seems safe enough

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
dnakamura: THe main thing im worried about is openj9. Is it possible to leave this off be default, but enable it as part of omr CI builds?

cmake/modules/platform/toolcfg/xlc.cmake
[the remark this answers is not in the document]
dnakamura: just because we aren't doing it/its not supported deosnt mean that somebody wont try it.
Should be easy enough to change from `elseif(OMR_HOST_ARCH STREQUAL "s390")` to `elseif(OMR_HOST_OS STREQUAL "zos")`

omrtrace/omrtracewrappers.cpp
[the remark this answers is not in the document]
dnakamura: the logic was that enabling native ebcdic handling is a safe enough flag to enable in production (since any failures would show up at build time). Where as enabling it in the runtime is more tricky and requires more thorough testing to ensure there are no bugs that will only show up on users machines

util/a2e/CMakeLists.txt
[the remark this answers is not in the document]
dnakamura: what i meant is that its something that cmake should be handling by itself, but thats obviously outside of the scope of this pr (my comment was really more of a reminder to myself that I should dig around in the cmake sources to see how to make that happen)

omrtrace/omrtracewrappers.cpp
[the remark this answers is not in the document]
fjeremic: Fixed as part of 5682d6e and 91f1ee5

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
fjeremic: Fixed as part of 5682d6e and 91f1ee5

fjeremic:
@genie-omr build all

fjeremic:
@genie-omr build all

cmake/caches/Travis.cmake
```diff
@@ -34,3 +34,6 @@ set(OMR_THR_CUSTOM_SPIN_OPTIONS ON CACHE BOOL "")
 set(OMR_THR_FORK_SUPPORT ON CACHE BOOL "")
 set(OMR_THR_SPIN_WAKE_CONTROL ON CACHE BOOL "")
 set(OMR_THR_THREE_TIER_LOCKING ON CACHE BOOL "")
+
+set(OMR_USE_NATIVE_ENCODING ON CACHE BOOL "")
+set(OMR_TOOLS_USE_NATIVE_ENCODING ON CACHE BOOL "")
```
keithc-ca: The old (autotools) defaults for both of these was `OFF`.

cmake/config.cmake
```diff
@@ -82,6 +85,10 @@ set(OMR_TOOLS_IMPORTFILE "IMPORTFILE-NOTFOUND" CACHE FILEPATH
 	"Point it to the ImportTools.cmake file of a native build"
 )
 
+set(OMR_TOOLS_USE_NATIVE_ENCODING FALSE CACHE BOOL
```
keithc-ca: This should be more consistent: use `OFF` / `ON` or `FALSE` / `TRUE` instead of a mixture.

compiler/infra/OMRMonitor.hpp
```diff
@@ -47,18 +47,23 @@ class Monitor
    static TR::Monitor *create(char *name);
    static void destroy(TR::Monitor *monitor);
    void enter();
-   int32_t try_enter() { TR_UNIMPLEMENTED(); }
+   int32_t try_enter() { TR_UNIMPLEMENTED(); return 0; }
    int32_t exit(); // returns 0 on success
    void destroy();
    void wait() { TR_UNIMPLEMENTED(); }
-   intptr_t wait_timed(int64_t millis, int32_t nanos) { TR_UNIMPLEMENTED(); }
+   intptr_t wait_timed(int64_t millis, int32_t nanos) { TR_UNIMPLEMENTED(); return static_cast<intptr_t>(0); }
```
keithc-ca: Why is a `static_cast` needed to `intptr_t`, but not `int32_t` (above)?

cmake/caches/Travis.cmake
fjeremic: @dnakamura can you comment on these? This commes from the latest two commits.

compiler/infra/OMRMonitor.hpp
fjeremic: I've removed commit dc392586d which included these `return` statements as #3610 has properly fixed `TR_UNIMPLEMENTED` to mark it with `OMR_NORETURN`. These changes no longer exist.

cmake/config.cmake
fjeremic: Fixed in b7ecc9a to use `OFF` / `ON` as the rest of the file uses the same.

fjeremic:
@genie-omr build all

cmake/caches/Travis.cmake
Leonardo2718: @dnakamura ping?

dnakamura: the default for cmake is still off. `Travis.cmake` specifies options we set for travis/ci builds

Leonardo2718:
@keithc-ca Have your comments been addressed?

I intend to merge this PR today, so if anyone still wants to review this work, please do so ASAP.

cmake/caches/Travis.cmake
keithc-ca: The native encoding properties still say `ON` despite the apparent agreement above.

cmake/modules/platform/os/zos.cmake
```diff
@@ -20,12 +20,15 @@
...
```
keithc-ca: Why should `_OPEN_THREADS` be `2`? [1] suggests it only needs to be defined.

[1] https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxbd00/ptonce.htm

cmake/modules/platform/toolcfg/xlc.cmake
```diff
@@ -19,32 +19,53 @@
...
+	if(OMR_ENV_DATA64)
+		list(APPEND OMR_PLATFORM_COMPILE_OPTIONS
+			-q64
+		)
+	else()
+		# -qarch should be there for 32 and 64 C/CXX flags but the C compiler is used for the assembler and it has trouble with some assembly files if it is specified
+		list(APPEND OMR_PLATFORM_COMPILE_OPTIONS
+			-q32
+			-qarch=ppc
```
keithc-ca: `-qarch=ppc` seems wrong (at least on z/OS)

cmake/modules/platform/toolcfg/xlc.cmake
```diff
@@ -19,32 +19,53 @@
...
+	endif()
+
+	# Testarossa build variables. Longer term the distinction between TR and the rest
+	# of the OMR code should be heavily reduced. In the mean time, we keep
+	# the distinction
+
+	# TR_COMPILE_OPTIONS are variables appended to CMAKE_{C,CXX}_FLAGS, and so
+	# apply to both C and C++ compilations.
+	list(APPEND TR_COMPILE_OPTIONS
+		-qarch=pwr7
```
keithc-ca: `-qarch=pwr7` seems wrong for z/OS

fvtest/compilertest/build/toolcfg/gnu/common.mk
```diff
@@ -143,13 +143,13 @@ endif
 
 ifeq ($(HOST_ARCH),z)
     ifeq ($(HOST_BITS),32)
```
keithc-ca: OpenJ9 defines `HOST_BITS` as 31 or 64 for z/OS, perhaps this should be `ifneq $(HOST_BITS),64)`?

omrsigcompat/CMakeLists.txt
```diff
@@ -86,6 +86,12 @@ if(OMR_HOST_OS MATCHES "linux|osx")
 endif()
 if(OMR_HOST_OS STREQUAL "zos")
 	target_link_libraries(omrsig PRIVATE j9a2e)
+
+	add_custom_command(TARGET omrsig
+		POST_BUILD
+		COMMAND mv libomrsig.x ${CMAKE_BINARY_DIR}/libomrsig.x
```
keithc-ca: This should use `cp -f`.

port/unix/omrfile.c
```diff
@@ -354,9 +354,9 @@ omrfile_open(struct OMRPortLibrary *portLibrary, const char *path, int32_t flags
 	} else {
 		fd = open(path, realFlags, mode);
 	}
-#else
+#else /* defined(J9ZOS390)  && !defined(OMR_EBCDIC) */
 	fd = open(path, realFlags, mode);
-#endif
+#endif /* defined(J9ZOS390)  && !defined(OMR_EBCDIC) */
```
keithc-ca: nit: extra spaces in closing comments

port/unix/omrfiletext.c
```diff
@@ -404,12 +404,12 @@ file_write_using_iconv(struct OMRPortLibrary *portLibrary, intptr_t fd, const ch
...
 {
-#if (defined(J9ZOS390))
+#if defined(J9ZOS390) && !defined(OMR_EBCDIC)
 	const char eol = a2e_tab['\n'];
 	char *tempStr = NULL;
-#else
+#else /* defined(J9ZOS390)  && !defined(OMR_EBCDIC) */
 	const static char eol = '\n';
-#endif /* defined(J9ZOS390) */
+#endif /* defined(J9ZOS390)  && !defined(OMR_EBCDIC) */
```
keithc-ca: nit: extra spaces in closing comments

port/zos390/omrgetjobname.c
```diff
@@ -52,13 +52,19 @@ omrget_jobname(struct OMRPortLibrary *portLibrary, char *jobname, uintptr_t leng
 	if (tmp_jobname) {
 		memset(tmp_jobname, '\0', J9_MAX_JOBNAME);
 		_JOBNAME(tmp_jobname);  /* requires <31bit address */
+#if !defined(OMR_EBCDIC)
```
keithc-ca: Why just `!defined(OMR_EBCDIC)` here, but `defined(J9ZOS390) && !defined(OMR_EBCDIC)` elsewhere?

port/zos390/omrosdump.c
```diff
@@ -259,10 +259,14 @@ tdump_wrapper(struct OMRPortLibrary *portLibrary, char *filename, char *dsnName)
 	/* the filename must be entirely uppercase for IEATDUMP requests */
 	convertToUpper(portLibrary, filename, strlen(filename));
 
+#if !defined(OMR_EBCDIC)
 	/* Convert filename into EBCDIC... */
 	dsnName = a2e_func(filename, strlen(filename) + 1);
 	err = tdump(portLibrary, filename, dsnName, &returnCode, &reasonCode);
 	free(dsnName);
+#else
```
keithc-ca: nit: missing comment

util/a2e/CMakeLists.txt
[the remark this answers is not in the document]
keithc-ca: Like elsewhere, this should use `cp -f`, not `mv`.

cmake/modules/platform/os/zos.cmake
dnakamura: https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxbd00/ftms.htm

>When defined to 1, this indicates that symbols required by POSIX.1, POSIX.1a, and POSIX.4a(draft 6) are made visible.
>
>When defined to 2, additional pthread functions introduced in z/OS V1R07 from Single UNIX Specification, Version 3 are made visible, along with those made visible when this is defined to 1.

port/zos390/omrgetjobname.c
dnakamura: given that it is a zos specific file, its safe to assume that `J9ZOS390` is defined.

cmake/modules/platform/os/zos.cmake
keithc-ca: Elsewhere on that page it uses the word 'tolerate':
> _ALL_SOURCE does not expose functionality first introduced in Single UNIX Specification, Version 3 under macro definitions _XOPEN_SOURCE 600 or _POSIX_C_SOURCE 200112L, although it does tolerate interfaces made visible by defining _OPEN_THREADS to 2 or 3.

This doesn't sound ideal, but I suppose it's necessary because we use `pthread_sigmask()` (and probably other functions) in the list made available when `_OPEN_THREADS == 2`.

buildenv/jenkins/jobs/pull-requests/PullRequest-zos_390-64
```diff
@@ -14,41 +14,23 @@ pipeline {
...
-                    sh'''for file in `find . -name '*.tdf'`;do; chtag -p "$file" | grep -q "ISO8859-1"; if [ $? -eq 0 ]; then; mv "$file" "$file.iso88591";iconv -f ISO8859-1 -t IBM-1047 "$file.iso88591" >"$file" && chtag -tc IBM-1047 "$file"; fi; done'''
-                    
-                    echo 'Compile...'
-                    sh'''make -j4'''
-                    
-                    //echo 'Output CCACHE stats after running'
-                    //echo '''ccache -s'''
+                    dir('build') {
+                        echo 'Configure...'
+                        sh '''cmake -Wdev -C../cmake/caches/Travis.cmake -DCMAKE_C_COMPILER=/bin/c89 -DCMAKE_CXX_COMPILER=/bin/xlc -DOMR_DDR=0 -DOMR_THR_FORK_SUPPORT=0 ..'''
```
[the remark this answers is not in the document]
fjeremic: Indeed they should, however CMake seems to search for `/bin/c++` first, and on the OMR build machines this is a non-working compiler version:

```
JENKINS:/tmp/fjeremic/omr/build: >cmake -Wdev -C../cmake/caches/Travis.cmake -DOMR_DDR=0 -DOMR_THR_FORK_SUPPORT=0 ..
loading initial cache file ../cmake/caches/Travis.cmake
-- Starting with CMake version 3.5.1
-- Found Git: /u/jenkins/jenkins-zos/git (found version "2.3.5_os390_b013")
-- The CXX compiler identification is unknown
-- The C compiler identification is zOS
-- Check for working CXX compiler: /bin/c++
-- Check for working CXX compiler: /bin/c++ -- broken
CMake Error at /openzdk/jenkins/CMake/Modules/CMakeTestCXXCompiler.cmake:54 (message):
  The C++ compiler "/bin/c++" is not able to compile a simple test program.

  It fails with the following output:

   Change Dir: /tmp/fjeremic/omr/build/CMakeFiles/CMakeTmp
```

We can likely fix this by defaulting to `xlc` first for z/OS. I'll have to find where to make such a change in the port in [1] however for the time being given these pipeline scripts are tailored towards the two z/OS machines hooked up to OMR this seems to be a happy medium. After this PR is merged I'll look to fixing this and removing the two defines. Is that reasonable?

[1] https://github.com/fjeremic/CMake

buildenv/jenkins/jobs/pull-requests/PullRequest-zos_390-64
```diff
@@ -14,41 +14,23 @@ pipeline {
...
-                    sh'''make -j4'''
-                    
-                    //echo 'Output CCACHE stats after running'
-                    //echo '''ccache -s'''
+                    dir('build') {
+                        echo 'Configure...'
+                        sh '''cmake -Wdev -C../cmake/caches/Travis.cmake -DCMAKE_C_COMPILER=/bin/c89 -DCMAKE_CXX_COMPILER=/bin/xlc -DOMR_DDR=0 -DOMR_THR_FORK_SUPPORT=0 ..''' 
+                       
+                        echo 'Compile...'
+                        sh '''make -j4'''
```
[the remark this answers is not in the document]
fjeremic: There is no "easy" way to determine the number of CPUs on a z/OS machine. Such a USS command does not exist. Similarly to above this value is hardcoded towards our two build machines hooked up to the OMR CI. If you know of a bullet-proof way of finding out the number of CPUs available without dependencies I'm happy to make an appropriate change.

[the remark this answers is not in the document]
fjeremic: I have a note filed under my random notes drawer with the following:

> This is so much harder than it needs to be. If you ever get stuck on a z/OS machine for which you do not have TSO access, no JVM installed you can use the following one liner to find the number of cores (CPUs via `D M=CPU`) on the system:
> ```
>echo "int main() {return (int)((char * __ptr32 * __ptr32 * __ptr32 *)16)[0][165][53];}" > numcpus.cpp && xlc numcpus.cpp && ./a.out || echo $?
>```

It's a super-hack involving compiling and running a C program just to determine the number of cores available! And it doesn't even take into account SMT, etc. I'd rather not be doing something like this.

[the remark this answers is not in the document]
fjeremic: Note on z/OS `make` does not understand `MAKEFLAGS` either so at best what we can do is change this to something like `make $MAKEFLAGS` and at the Jenkins node setup level export `MAKEFLAGS` to the appropriate `-j` value for the particular Jenkins node. Does that sound reasonable?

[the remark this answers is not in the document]
keithc-ca: `oeconsol D M=CPU` lists CPUs which can be counted.

[1] ftp://ftp.www.ibm.com/s390/zos/tools/oeconsol/oeconsol.txt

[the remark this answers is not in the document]
fjeremic: `oeconsol` is a supervisor command. It is not available to users unless spcial permissions are given.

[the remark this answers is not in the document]
fjeremic: I believe `oeconsol` requires special permissions. It is not a usermode tool.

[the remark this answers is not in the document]
keithc-ca: I don't think this needs to hold up this PR; I note `make -j4` was here before.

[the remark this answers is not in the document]
fjeremic: Looking into it more indeed we need RACF permissions to issue a `DISPLAY` TSO command. This is a no-go.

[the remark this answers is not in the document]
fjeremic: Looking at the existing pipeline files in [1] pretty much all of them hardcode a `-j` value. IMO if we really ought to fix this we need to do it on all platforms, so it's outside the scope of what we're trying to accomplish in this PR. 

[1] https://github.com/eclipse/omr/tree/master/buildenv/jenkins/jobs/pull-requests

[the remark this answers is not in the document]
fjeremic: I've opened #3972 to track this.

cmake/modules/platform/toolcfg/xlc.cmake
fjeremic: This block is guarded by `OMR_HOST_ARCH STREQUAL "ppc"`. Note xlC exists on Power as well. This is the xlc.cmake tool configuration (both for Power and Z).

cmake/modules/platform/toolcfg/xlc.cmake
fjeremic: Similarly to above this is guarded by `OMR_HOST_ARCH STREQUAL "ppc"`

port/zos390/omrosdump.c
fjeremic: Fixed in 370fd8ea

port/unix/omrfiletext.c
fjeremic: Fixed in 370fd8ea

omrsigcompat/CMakeLists.txt
fjeremic: Fixed in 370fd8ea

fvtest/compilertest/build/toolcfg/gnu/common.mk
fjeremic: Fixed in 370fd8ea

util/a2e/CMakeLists.txt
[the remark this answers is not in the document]
fjeremic: Fixed in 370fd8ea (changed `mv` to `cp -f`). @dnakamura do you want to open an issue to track the proper fix for this as mentioned by your previous comment?

port/unix/omrfile.c
fjeremic: Fixed in 370fd8ea

fjeremic:
@keithc-ca I've addressed all latest reviews.

@genie-omr build all

fvtest/compilertest/build/toolcfg/gnu/common.mk
```diff
@@ -142,15 +142,13 @@ ifeq ($(HOST_ARCH),p)
...
-    ifeq ($(HOST_BITS),32)
-        CX_DEFINES+=S390 FULL_ANSI
-        CX_FLAGS+=-m31 -fPIC -fno-strict-aliasing -mtune=z10 -march=z9-109 -mzarch
-        CX_FLAGS_DEBUG+=-gdwarf-2
-    endif
-    
     ifeq ($(HOST_BITS),64)
         CX_DEFINES+=S390 FULL_ANSI S39064
         CX_FLAGS+=-fPIC -fno-strict-aliasing -mtune=z10 -march=z9-109 -mzarch
+	else
```
keithc-ca: inconsistent indent

cmake/caches/Travis.cmake
keithc-ca: @fjeremic Are planning to fix this? I think it should say:
```
set(OMR_USE_NATIVE_ENCODING OFF CACHE BOOL "")
set(OMR_TOOLS_USE_NATIVE_ENCODING OFF CACHE BOOL "")
```

fvtest/compilertest/build/toolcfg/gnu/common.mk
fjeremic: Fixed in 2c6cfc1

Leonardo2718:
@keithc-ca I believe your last concern has been addressed so I will merge this once the CI build completes.

@genie-omr build all

keithc-ca:
I didn't see any response to my concern about `Travis.cmake`.

cmake/caches/Travis.cmake
fjeremic: That will turn off those variables for the CI builds which is not what we want. We want to enable those two flags on the builds but keep them disabled otherwise. IMO I would personally just enable them everywhere as it's the correct thing to do.

fjeremic:
I'm going to start resolving what I believe are addressed reviews on behalf of reviewers as there are so many threads on this PR and it's not easy to see what is active and what is not.

cmake/caches/Travis.cmake
keithc-ca: I understood the meaning of 'USE_NATIVE_ENCODING' to mean 'use EBCDIC on z/OS and ASCII elsewhere', in which case, it's *not* the right thing for the travis build (on z/OS).

cmake/modules/platform/os/zos.cmake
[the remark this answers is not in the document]
fjeremic: See also related discussion in [2].

[2] https://github.com/eclipse/omr/pull/3693#discussion_r291634203.

cmake/caches/Travis.cmake
fjeremic: > it's _not_ the right thing for the travis build (on z/OS).

Can you please explain why you believe that? Why would we not want to use EBCDIC on z/OS?

rwy0717: Yeah as @fjeremic and @dnakamura said, these options only affect CI jobs, the [default is off](https://github.com/eclipse/omr/pull/3693/files#diff-370bf43c6501eaf4d5fde675c00cbc5fR51). In the old autotools system, these options enabled in CI through the [example config](https://github.com/eclipse/omr/blob/master/example/glue/configure_includes/configure_zos_390.mk#L36). To me, this looks like the same behaviour as before.

keithc-ca: The `--enable-native-encoding` configuration option is not the same thing as `USE_NATIVE_ENCODING`, see [1]. We do want `--enable-native-encoding` on z/OS, but we don't want `USE_NATIVE_ENCODING` specified globally. Some makefiles say `USE_NATIVE_ENCODING := 1` but many do not.

[1] https://github.com/eclipse/omr/blob/master/omrmakefiles/rules.zos.mk#L23

fjeremic: I'm not sure what the answer here is or what is wanted. As far as the _CMake_ OMR build system is concerned we want native encoding enabled by default as files are being checked out via Git for z/OS (Rocket Software). In such a configuration all files are file tagged as ISO8859-1 and because of `_BPX_AUTOCVT=ON` all disk operations get transparently converted from ISO8859-1 to IBM1047 transparently.

As such I cannot envision a scenario in which we would not want to use the native encoding always, since everything will be in EBCDIC in the z/OS case. The CMake z/OS system will not work unless `OMR_EBCDIC` is defined (it gets defined via `USE_NATIVE_ENCODING` in this case).

Leonardo2718: @keithc-ca Can you explain how setting `OMR_USE_NATIVE_ENCODING` and `OMR_TOOLS_USE_NATIVE_ENCODING` to `ON` affects the CI builds, compared to the autotools builds? Given that the genie builds have passed (several times), I'm trying to udnerstand whether any difference between the autotools and cmake builds actually matter.

cmake/modules/platform/os/zos.cmake
Leonardo2718: Is the consensus here that `_OPEN_THREADS == 2` is correct? Can this discussion be resolved?

cmake/caches/Travis.cmake
keithc-ca: You'll notice that tools like `hookgen`, `tracegen` and `tracemerge` set `USE_NATIVE_ENCODING := 1` in their makefiles. On the other hand no other makefiles do because the expectation is that text files will be read and written as ASCII (given OMR's heritage as part of J9).

`OMR_TOOLS_USE_NATIVE_ENCODING` seems to be specific to cmake builds and seems to be based on the incorrect assumption that all tools will use the same encoding (`ddrgen` is an exception (a tool not located in the `tools` folder) in that it needs to universally operate in ASCII).

I agree that the scenario that @fjeremic describes is a reasonable one (if we have our OMR blinders on), but I don't think his description of `_BPX_AUTOCVT=ON` is accurate.

I suggest it should be the choice of the consuming project (e.g. openj9) as to which parts of OMR interpret text files as ASCII or EBCDIC. This particular file relates only to the OMR travis build, but the concepts encoded in `OMR_USE_NATIVE_ENCODING` and `OMR_TOOLS_USE_NATIVE_ENCODING ` don't align with the autotools capabilities and freedoms.

fjeremic: I'll try to summarize this as best as I can from my current understanding.

The OMR project defines a target called j9a2e [1]. This target defines conversion functions between the ASCII and EBCDIC (default on z/OS) encodings, but that's not all. It also overrides libc APIs by defining header files which will replace standard library calls [2] with wrapper calls in the j9a2e target which will transparently under the hood convert ASCII inputs into EBCDIC before actually calling the real standard library functions (example in [3] for the `fopen` API).

The purpose of this target is so that when OpenJ9 consumes OMR it can link against libj9a2e.so and treat everything internally as ASCII which the files on disk remain EBCDIC. This is presumably so that targets which are built around ASCII (such as the natives in OpenJDK) function correctly. Please correct me if I'm wrong on this one as it is a pure guess on my part.

Now naturally if all internal operations are assuming ASCII and literal strings inside the compilation unit also have to be in ASCII so things are consistent. This is why we specify the `-Wc,"convlit(ISO8859-1)"` compiler flag [4] so that it converts all literal strings in every file into ASCII.

> The old (autotools) defaults for both of these was OFF.

Now the default for `USE_NATIVE_ENCODING` as seen in [4] is 0. This is precisely because we only ever built 8 targets in the autotools build system. These are the 8 targets we currently run in our CI (omralgotest, omrvmtest, omrutiltest, omrsigtest, omrrastest, omrsubscribertest, omrtraceoptiontest). **Note** that targets such as jitbuilder, tril, and comptest are not part of that list. The reason being is because these target _cannot_ work unless `USE_NATIVE_ENCODING = 1` because of the heavy use of C++ streams.

For example in [5] we use `std::cout` to print something to the screen. If `USE_NATIVE_ENCODING = 0` then this object file gets compiled with the `-Wc,"convlit(ISO8859-1)"` flag and links against libj9a2e.so. However C++ streams do not operate through libc and as such the functions we override in [2] will not be executed. Now because of `convlit` that literal string is encoded as ASCII and thus we attempt to print an ASCII string to the screen on z/OS which uses EBCIDC as the native encoding. The result of this on my tests is that the terminal session I execute such a simple program in simply hangs. I'm not yet sure if it's because of my terminal or something deeper went wrong. Regardless the issue is well understood.

---

To me the sane solution for OMR is to set `USE_NATIVE_ENCODING = 1` by default so that we use the native encoding of EBCIDC on z/OS because otherwise we may be restricting consumers of OMR into only using the OMR port library to do any IO, not to mention that such consumers of OMR will be limited to not using C++ IO features.

This is a heavy handed restriction IMO and not one I would advise.

For projects such as OpenJ9 which want to consume OMR and work internally with ASCII are free to do so by simply overriding `USE_NATIVE_ENCODING` to 0. 

Note that as part of this PR there are two defines:
   - `OMR_USE_NATIVE_ENCODING` which determines whether the `omr_base` will use native encoding under the hood (`-DOMR_EBCDIC`) or not (`convlit` + link with libj9a2e.so)
   - `OMR_TOOL_USE_NATIVE_ENCODING` which determines whether the 3 tools (hookgen, tracegen, tracemerge) will use native encoding or not. These tools seem to assume they will be working internally with the native encoding of the platform, so EBCDIC in case of z/OS.

Both of these can be overridden by consumers of OMR.

@keithc-ca @Leonardo2718 @dnakamura @rwy0717 do you agree or disagree? If you disagree please be specific about which part and try to provide an alternative if possible.

[1] https://github.com/eclipse/omr/blob/daa8b14b6a37a1a94aacdde5b0d9578600eec5fe/util/a2e/CMakeLists.txt#L23
[2] https://github.com/eclipse/omr/blob/master/util/a2e/headers/stdio.h#L77-L115
[3] https://github.com/eclipse/omr/blob/daa8b14b6a37a1a94aacdde5b0d9578600eec5fe/util/a2e/atoe.c#L690-L717
[4] https://github.com/eclipse/omr/blob/daa8b14b6a37a1a94aacdde5b0d9578600eec5fe/omrmakefiles/rules.zos.mk#L84-L89
[5] https://github.com/eclipse/omr/blob/daa8b14b6a37a1a94aacdde5b0d9578600eec5fe/jitbuilder/release/cpp/samples/Simple.cpp#L38-L41

Leonardo2718: I agree with @fjeremic.

OMR should not, by default, restrict users from using the tools they know (C++ I/O streams in this case). If a consumer of OMR requires deiferent behaviour, they can override the default.

keithc-ca: I agree users of the OMR project should be free to select the desired behavior with respect to (text) file encoding.

The fundamental difference (and problem) I see is that under the proposed cmake system, `USE_NATIVE_ENCODING` becomes a *global* choice while today with autotools it is only applied to a few folders within OMR. So OpenJ9, for example, would have to set `USE_NATIVE_ENCODING = OFF` to get the desired behavior of the port library but then tools hookgen, tracegen and tracemerge will work differently. It's possible that OpenJ9 could adapt to that change in behavior, but my main point is just that: it's a change of behavior.

fjeremic: According to [4] above it is a global setting, unless I'm misinterpreting the meaning of `GLOBAL_CPPFLAGS` in the autotools build system. Regardless I agree what I'm proposing is a change in behavior which deviates from autotools but enables us to build more OMR targets on z/OS (jitbuilder, tril, comptest, and pretty much the entire rest of OMR) and it standardizes on the notion of using the native encoding everywhere _in OMR_, not mixing bits and pieces to be on or off.

> but then tools hookgen, tracegen and tracemerge will work differently.

This isn't true because `OMR_USE_NATIVE_ENCODING` has no bearing on the use of native encoding for those 3 specific tools. `OMR_TOOLS_USE_NATIVE_ENCODING` is what controls whether we use native encoding there. Example in hookgen in [5].

[5] https://github.com/eclipse/omr/pull/3693/files#diff-5c657e288329114e6f63ea85ebefcaadR32

keithc-ca: Each folder with a makefile can independently select the appropriate value of `USE_NATIVE_ENCODING` before `rules*.mk` is included.

dnakamura: I would argue that having the libraries built as EBCDIC by default is probably the best solution, given that it is also the least surprising. If a consumer wants to override that and use a different character encoding than the OS (and take on all the associated headaches) then they are free to do that. By having it an opt-in means that by doing so, they take responsibility of the repercussions of that decision (eg. not being able to use C++ easily,  having to include `atoe.h` etc.)

dnakamura: >Each folder with a makefile can independently select the appropriate value of USE_NATIVE_ENCODING before rules*.mk is included.

True, but its not really a magic bullet. Each folder needs to ensure that every thing it links against was built with the same value of `Each folder with a makefile can independently select the appropriate value of `USE_NATIVE_ENCODING` (unless it doesnt use any api involving strings).  However since it is only ever set by hook/tracegen (which have minimal dependencies) its not that hard to keep track of at the moment

fjeremic:
@genie-omr build all

Leonardo2718:
The latest changes LGTM.

Given that these changes as a whole enable more extensive testing of OMR components, I think the change in behaviour compared to autotools is acceptable.

Since this PR is holding back other work, I am going to merge it today and will request that further changes and discussions happen in followup issues and PRs.

PR #3695, opened 2019-03-25
Loop versioner support for dynamic checkcast

yanluo7:
Currently when collecting checkcast nodes for loop invariant detection, loop versioner asserts the cast class should always be a loadaddr. With the upcoming CFG simplifier change, the cast class child can be dynamic. So this restriction is removed to allow versioner to detect loop invariant for dynamic checkcast nodes as well. The existing infrastructure appears to be already in-place to handle the new cases.

### DIFF
--- a/compiler/optimizer/LoopVersioner.cpp
+++ b/compiler/optimizer/LoopVersioner.cpp
@@ -3445,21 +3445,29 @@ bool TR_LoopVersioner::detectChecksToBeEliminated(TR_RegionStructure *whileLoop,
                }
             else if (currentOpCode.isCheckCast()) // (currentOpCode.getOpCodeValue() == TR::checkcast)
                {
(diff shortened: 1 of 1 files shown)


yanluo7:
@jdmpapin hi Devin, could you review?

andrewcraik:
looks good - over to @vijaysun-omr

vijaysun-omr:
@genie-omr build all

yanluo7:
@vijaysun-omr should be ready to merge.

vijaysun-omr:
I assume you have tested this new code path ? I doubt there is much testing in our existing buckets that even gets us to the point where such a checkcast without a constant class second child is generated.

yanluo7:
@vijaysun-omr This has gone through personal builds which have dynamic checkcast enabled to test the code path.

vijaysun-omr:
Thanks, checks have passed and reviews are done. Merging.

PR #3697, opened 2019-03-26
CMake: Add validation to OMR_THR_YIELD_ALG option

dnakamura:
Check that platform is supported if OMR_THR_YIELD_ALG is enabled.
Previously, if it were enabled on an unsupported platform, the result
would be a link error due to undefined reference to `THREAD_YIELD_NEW`


youngar:
@genie-omr build all

PR #3700, opened 2019-03-27
Use known object temps for inlined callee

cathyzhyi:
When inliner inlines a method and replaces the loads of parms with loads
of priv arg temps, the known object info we got from prex arguments are
lost. GlobalVP needs to run to propagate the known object info again
which is expensive and doesn't kick in at warm level. This change replaces
the priv args with known object temps when feasible.

### DIFF
--- a/compiler/il/symbol/OMRSymbol_inlines.hpp
+++ b/compiler/il/symbol/OMRSymbol_inlines.hpp
@@ -214,19 +214,6 @@ OMR::Symbol::isPinningArrayPointer()
    }
 
 void
-OMR::Symbol::setAutoAddressTaken()
-   {
-   TR_ASSERT(self()->isAuto(), "assertion failure");
-   _flags.set(AutoAddressTaken);
-   }
-
-bool
-OMR::Symbol::isAutoAddressTaken()
-   {
-   return self()->isAuto() && _flags.testAny(AutoAddressTaken);
-   }
-
-void
 OMR::Symbol::setSpillTempLoaded()
    {
    if (self()->isAuto()) // For non auto spills (ie. to original location) we can't optimize removing spills
--- a/compiler/optimizer/Inliner.cpp
+++ b/compiler/optimizer/Inliner.cpp
@@ -1364,7 +1364,6 @@ TR_DumbInliner::analyzeCallSite(
 //---------------------------------------------------------------------
 
 static TR::TreeTop * cloneAndReplaceCallNodeReference(TR::TreeTop *, TR::Node *, TR::Node *, TR::TreeTop *, TR::Compilation *);
-static void addSymRefsToList(List<TR::SymbolReference> &, List<TR::SymbolReference> &);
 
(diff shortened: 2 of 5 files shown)


andrewcraik:
Why only handle the receiver shouldn't we be able to propagate known objects for all arguments? Is there some performance consideration for not doing it for all arguments?

cathyzhyi:
On a second thought, we should be able to propgate not only the receiver. Will modified the code to be more general.

vijaysun-omr:
@cathyzhyi is this still a WIP ?

0xdaryl:
@genie-omr build all

0xdaryl:
@cathyzhyi , is this ready for final review by @vijaysun-omr ?

cathyzhyi:
@vijaysun-omr Hey Vijay, this change is finally ready. could you help to review? Thanks!

compiler/optimizer/Inliner.cpp
```diff
@@ -2354,45 +2354,58 @@ TR_ParameterToArgumentMapper::initialize(TR_CallStack *callStack)
...
                TR::SymbolReference * symRef = NULL;
-
+               const static bool disableUseKnownObjectTempsForParms = feGetEnv("TR_DisableUseKnownObjectTempsForParmsInCallee") ? true: false;
+               int argOrdinal = argIndex - _callNode->getFirstArgumentIndex();
+               TR_PrexArgument * prexArgument = _argInfo->get(argOrdinal);
+               // use known object temp if the argument is a known object
+               if (!disableUseKnownObjectTempsForParms && prexArgument
+                   && !parmMap->_parmIsModified
+                   && TR_PrexArgument::knowledgeLevel(prexArgument) == KNOWN_OBJECT
+                   && !prexArgument->isTypeInfoForInlinedBody())
```
vijaysun-omr: What is this ```isTypeInfoForInlinedBody``` check for ? Might be good to capture in the comment so that the reasoning is apparent in the code.

compiler/il/symbol/OMRSymbol_inlines.hpp
```diff
@@ -213,19 +213,6 @@ OMR::Symbol::isPinningArrayPointer()
    return self()->isAuto() && _flags.testAny(PinningArrayPointer);
    }
 
-void
```
vijaysun-omr: There is a small typo "propgate" in the commit message. If you can fix that up and at the same time mention this cleanup of the below code in the commit message, that would be great.

vijaysun-omr:
@cathyzhyi a friendly reminder that there are a couple of outstanding review comments to address.

cathyzhyi:
@vijaysun-omr sry I missed the notifications. Just cleaned up the code as suggested. Thanks!

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
@genie-omr build all

vijaysun-omr:
Thanks, I see that comments have been added near the relevant variable declaration to explain regarding the question I brought up. I also see that the commit message is now updated. Checks have also passed and so I am merging this.

PR #3701, opened 2019-03-27
Enable CopyForwardMarkCompactHybrid mode

LinHu2016:
- enable CopyForwardMarkCompactHybrid mode by default for fully
	testing.
	CopyForwardMarkCompactHybrid mode intends to avoid MarkCompactOnly
	PGCs, which are caused by jniCritical or abort case during GMP.
	it would improve the GC performance for the most cases, but in
	some cases, it might be no performance difference or slight draw back.


amicic:
Most of the functionality that this flag enables in OpenJ9, which I already reviewed. This should be ok to enable.

PR #3705, opened 2019-03-29
CMake: Add export info

dnakamura:
Add export info for port, thread, and hookable. While these arent built
as shared libraries, they are provided as object libraries so that a
consumer can repackage them. In addition the makefiles currently maintain
lists of symbols to export for these targets


youngar:
@genie-omr build win

PR #3711, opened 2019-04-03
Thread final/non-final Release Caches

amicic:
threadFinalReleaseCaches method is generalized to be used in a middle of
a Scavenger cycle, when we do not want to flush Copy Cache Remainders,
since they can be re-used later.

Additional parameter is introduced to control, if we want to be final
release (when remainders are abandoned) or not.

At all existing call sites we set the parameter to true and we used to
abandon them always, what makes this a non-functional change. Future
users will set the parameter to false when needed.

### DIFF
--- a/gc/base/standard/Scavenger.hpp
+++ b/gc/base/standard/Scavenger.hpp
@@ -624,10 +624,10 @@ public:
 	/* methods used by either mutator or GC threads */
 	/**
 	 * All open copy caches (even if not full) are pushed onto scan queue. Unused memory is abondoned.
-	 * @param env Invoking thread. Could be master thread on behalf on mutator threads (threadEnvironment) for which copy caches are to be released, or could be mutator or GC thread itself.
-	 * @param threadEnvironment Thread for which copy caches are to be released. Could be either GC or mutator thread.
+	 * @param env Invoking thread, for which copy caches are to be released. Could be either GC or mutator thread.
+	 * @param final If true (typically at the end of a cycle), abandon TLH remainders, too. Otherwise keep them for possible future copy cache refresh.
 	 */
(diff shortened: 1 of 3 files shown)


gc/base/standard/Scavenger.hpp
```diff
@@ -624,10 +624,10 @@ class MM_Scavenger : public MM_Collector
 	/* methods used by either mutator or GC threads */
 	/**
 	 * All open copy caches (even if not full) are pushed onto scan queue. Unused memory is abondoned.
-	 * @param env Invoking thread. Could be master thread on behalf on mutator threads (threadEnvironment) for which copy caches are to be released, or could be mutator or GC thread itself.
-	 * @param threadEnvironment Thread for which copy caches are to be released. Could be either GC or mutator thread.
+	 * @param env Invoking thread, for which copy caches are to be released. Could be either GC or mutator thread.
+	 * @param final If true (typically at the end of a cycle), abandon TLH remainders, too. Otherwise keep them for possible future copy cache refresh.
 	 */
-	void threadFinalReleaseCaches(MM_EnvironmentBase *env);
+	void threadReleaseCaches(MM_EnvironmentBase *env, bool final = true);
```
dmitripivkine: This change is good, however I am not a strong supporter of set parameters by default. For me would be better to explicitly add extra parameter at the caller side. You touch caller side any way so this would not involve wide spread change

youngar:
@genie-omr build all

gc/base/standard/Scavenger.hpp
amicic: makes sense - will remove the default value.

youngar:
@genie-omr build all

PR #3712, opened 2019-04-03
Format output of jitbuilder apigen

dnakamura:
Add basic formatting (indentation) to the output from apigen


youngar:
@genie-omr build xlinux

Leonardo2718:
Tests are not platform specific so the xlinux testing should be enough. But, for good measure:

@genie-omr build win, osx, plinux, zlinux

PR #3718, opened 2019-04-05
Don't inline thunk archetype in VP

liqunl:
Leave the thunk archetype to methodHandleInvokeInliningGroup if it is
requested.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -5032,6 +5032,14 @@ static void devirtualizeCall(OMR::ValuePropagation *vp, TR::Node *node)
    if (methodSymbol->isComputed())
       {
 #ifdef J9_PROJECT_SPECIFIC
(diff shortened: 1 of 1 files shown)


liqunl:
@andrewcraik Can I have a review?

0xdaryl:
Please fix typo of "inline" in your commit title.

andrewcraik:
What is the danger of VP doing this inlining? it would just reduce the number of iterations of targeted inlining need to reach the end of a handle chain unless it is breaking a pattern?

liqunl:
@andrewcraik VP doesn't only inline the thunk, but also inlines callees of the thunk. Some of the callees can be folded if not inlined.

andrewcraik:
So isn't the fix to stop inlining things that can be folded in general rather than blacklisting when it happens?

liqunl:
@andrewcraik True. Close this PR and will put the fix in inliner.

PR #3719, opened 2019-04-05
Make arraytranslate return correct # of translated chars

dchopra001:
The arraytranslateEncodeSIMDEvaluator helper function invoked by the arraytranslateEvaluator does not correctly add the number of residue characters succesfully translated if a saturated value is encountered in the second input vector. This commit adds an add instruction to correctly account for all the values successfuly translated.


fjeremic:
@genie-omr build zlinux,zos

PR #3720, opened 2019-04-05
Anchor children when anchorTree is non-null

liqunl:
Allow user to decide whether children need to be anchored.

### DIFF
--- a/compiler/optimizer/OMRTransformUtil.cpp
+++ b/compiler/optimizer/OMRTransformUtil.cpp
@@ -390,7 +390,8 @@ OMR::TransformUtil::removeTree(TR::Compilation *comp, TR::TreeTop * tt)
 void
 OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::Node* node, TR::TreeTop * anchorTree, TR::Node* child)
    {
-   opt->anchorAllChildren(node, anchorTree);
+   if (anchorTree)
+      opt->anchorAllChildren(node, anchorTree);
    node->removeAllChildren();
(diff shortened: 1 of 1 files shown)


Leonardo2718:
@genie-omr build all

compiler/optimizer/OMRTransformUtil.cpp
```diff
@@ -390,7 +390,8 @@ OMR::TransformUtil::removeTree(TR::Compilation *comp, TR::TreeTop * tt)
 void
 OMR::TransformUtil::transformCallNodeToPassThrough(TR::Optimization* opt, TR::Node* node, TR::TreeTop * anchorTree, TR::Node* child)
    {
-   opt->anchorAllChildren(node, anchorTree);
+   if (anchorTree)
```
vijaysun-omr: Why leave the door open for a functional bug by not forcing the anchoring of children ?

liqunl: @vijaysun-omr If the call can be transformed to a passthrough, its children are not needed. Maybe the anchoring is never needed for this transformation.

vijaysun-omr: I'm not convinced yet. The nature of what is being transformed does'nt affect whether we can safely delete the children (having reference count > 1) or not, unless you are sure that the children are all ONLY accessed from the treetop being transformed OR the treetop in question is guaranteed to be later than the first evaluation points for those children.

liqunl: No, those are not guaranteed. In the case of unnecessary anchoring, other opts can easily get rid of dead trees if the anchoring is not needed. I'll close this PR as per discussion.

PR #3724, opened 2019-04-08
Guard double mapping api under config flag

bragaigor:
Guards double mapping GC api under config flag


youngar:
Looks like you got everywhere!  Technically our [coding standard]( https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md#comment-conditional-compilation-directives-if-ifdef-ifndef-else-elif-endif) says to close `#ifdef`s like this:
```diff
-#endif /* OMR_GC_DOUBLE_MAP_ARRAYLETS */
+#endif /* defined(OMR_GC_DOUBLE_MAP_ARRAYLETS) */
```
I'm probably one of the worst offenders for getting this wrong, but would you mind switching it over?

youngar:
@genie-omr build all

PR #3728, opened 2019-04-09
Add APIs for creating temp symRef with known obj

cathyzhyi:
This change adds APIs for creating temps with known obj index so
that the various local optimizations can see the known obj info
for temps without needing expensive global optimization like
global VP and global copy propagation.

Notice the existing API findOrCreateSymRefWithKnownObject always create
a symRef with known object index based on the given original symRef so
the symbols are shared by the 2 symRefs. It doesn't work for temps
because each temp symbol can only have one symbol reference.

### DIFF
--- a/compiler/compile/OMRSymbolReferenceTable.cpp
+++ b/compiler/compile/OMRSymbolReferenceTable.cpp
@@ -989,30 +989,50 @@ OMR::SymbolReferenceTable::findOrCreateSymRefWithKnownObject(TR::SymbolReference
    }
 
 TR::SymbolReference *
-OMR::SymbolReferenceTable::findOrCreateSymRefWithKnownObject(TR::SymbolReference *original, TR::KnownObjectTable::Index objectIndex)
+OMR::SymbolReferenceTable::findTempSymRefWithKnownObject(TR::KnownObjectTable::Index knownObjectIndex)
    {
-   TR_BitVector *bucket = _knownObjectSymrefsByObjectIndex[objectIndex];
+   return findSymRefWithKnownObject(NULL, knownObjectIndex);
+   }
+
+TR::SymbolReference *
+OMR::SymbolReferenceTable::findSymRefWithKnownObject(TR::Symbol *symbol, TR::KnownObjectTable::Index knownObjectIndex)
+   {
+   TR_BitVector *bucket = _knownObjectSymrefsByObjectIndex[knownObjectIndex];
    if (!bucket)
-      {
-      bucket = new (trHeapMemory()) TR_BitVector(baseArray.size(), trMemory(), heapAlloc, growable, TR_MemoryBase::SymbolReference);
-      _knownObjectSymrefsByObjectIndex[objectIndex] = bucket;
-      }
+      return NULL;
 
    TR_BitVectorIterator bvi(*bucket);
    while (bvi.hasMoreElements())
       {
       TR::SymbolReference *symRef = getSymRef(bvi.getNextElement());
-      if (symRef->getSymbol() == original->getSymbol())
+      if (symbol && symRef->getSymbol() == symbol)
+         return symRef;
+      if (!symbol && symRef->getSymbol()->isAuto()) //For temp, symbol can be NULL. Find any temp with the given knownObjectIndex.
          return symRef;
       }
+   return NULL;
+   }
+
+TR::SymbolReference *
+OMR::SymbolReferenceTable::findOrCreateSymRefWithKnownObject(TR::SymbolReference *originalSymRef, TR::KnownObjectTable::Index knownObjectIndex)
+   {
+   TR::SymbolReference *symRef = findSymRefWithKnownObject(originalSymRef->getSymbol(), knownObjectIndex);
+   if (symRef)
+      return symRef;
 
    // Need a new one
-   TR::SymbolReference *result = new (trHeapMemory()) TR::SymbolReference(self(), *original, 0, objectIndex);
+   TR_BitVector *bucket = _knownObjectSymrefsByObjectIndex[knownObjectIndex];
+   if (!bucket)
+      {
+      bucket = new (trHeapMemory()) TR_BitVector(baseArray.size(), trMemory(), heapAlloc, growable, TR_MemoryBase::SymbolReference);
+      _knownObjectSymrefsByObjectIndex[knownObjectIndex] = bucket;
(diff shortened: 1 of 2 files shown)


Leonardo2718:
@genie-omr build all

compiler/compile/OMRSymbolReferenceTable.cpp
```diff
@@ -989,30 +989,46 @@ OMR::SymbolReferenceTable::findOrCreateSymRefWithKnownObject(TR::SymbolReference
...
+OMR::SymbolReferenceTable::findTempSymRefWithKnownObject(TR::KnownObjectTable::Index knownObjectIndex)
    {
-   TR_BitVector *bucket = _knownObjectSymrefsByObjectIndex[objectIndex];
+   return findSymRefWithKnownObject(NULL, knownObjectIndex, true /* isTemp */);
+   }
+
+TR::SymbolReference *
+OMR::SymbolReferenceTable::findSymRefWithKnownObject(TR::Symbol *symbol, TR::KnownObjectTable::Index knownObjectIndex, bool isTemp)
+   {
+   TR_BitVector *bucket = _knownObjectSymrefsByObjectIndex[knownObjectIndex];
```
andrewcraik: why does find allocate the bucket - if we are finding and there is no bucket we should just return NULL right?

andrewcraik: can't we tell from the symbol if we are dealing with a temp rather than having a Boolean flag?

compiler/compile/OMRSymbolReferenceTable.cpp
```diff
@@ -1023,6 +1039,14 @@ OMR::SymbolReferenceTable::findOrCreateSymRefWithKnownObject(TR::SymbolReference
    return result;
    }
 
+TR::SymbolReference *
+OMR::SymbolReferenceTable::createTempSymRefWithKnownObject(TR::Symbol *symbol, mcount_t owningMethodIndex, int32_t slot, TR::KnownObjectTable::Index knownObjectIndex)
```
andrewcraik: For non-temps we should start with the original symref and not the symbol - we'll need to be able to do the reverse and starting with the original symref will simplify changes to add this reverse capability in a follow-on change.

cathyzhyi: yes, this one is for temps only. Another `findOrCreateSymRefWithKnownObject` is for non-temps which takes the original sym ref as first argument. I can change the name `findOrCreateSymRefWithKnownObject` to `findOrCreateNoTempSymRefWithKnownObject` to be more explicit.

compiler/compile/OMRSymbolReferenceTable.cpp
cathyzhyi: Actually the symbol would be NULL if it's a temp. I thought passing a flag would be more straight forward than just checking if the symbol is NULL it must be a temp.

vijaysun-omr:
@genie-omr build all

compiler/compile/OMRSymbolReferenceTable.cpp
andrewcraik: I'm still not sure I like the bool and the symbol needing to agree - it seems redundant....

compiler/compile/OMRSymbolReferenceTable.cpp
andrewcraik: Can we add a fatal assert at least to check the symbol is a temp or something to stop someone using this by accident for a non-temp? The name does imply what you are saying, but I missed it reading the code so someone else might and a fatal assert would be the best thing if possible.

Leonardo2718:
AppVeyor build failed because of infra issues so I restarted it.

compiler/compile/OMRSymbolReferenceTable.cpp
cathyzhyi: yea, I agree it does seem redundant... changed to check if the symbol is NULL and added a comment to avoid confusion.

compiler/compile/OMRSymbolReferenceTable.cpp
cathyzhyi: Added assert to avoid confusion.

vijaysun-omr:
I've reviewed the code and it looks okay. One question I had is around the safety of this given that temp sym refs can be shared for more than one use in the original program. e.g. you may know that the inliner recycles temp sym refs based on it's knowledge of non-overlapping live ranges. 

i.e. what prevents us from considering a temp sym ref that was created with a known object in a different context that it was reused in where it is not a known object in that second context ?

vijaysun-omr:
@genie-omr build all

cathyzhyi:
> I've reviewed the code and it looks okay. One question I had is around the safety of this given that temp sym refs can be shared for more than one use in the original program. e.g. you may know that the inliner recycles temp sym refs based on it's knowledge of non-overlapping live ranges.
> 
> i.e. what prevents us from considering a temp sym ref that was created with a known object in a different context that it was reused in where it is not a known object in that second context ?

That part is in a subsequent PR https://github.com/eclipse/omr/pull/3700/files where in https://github.com/eclipse/omr/blob/343b653e82a4658cdc78fcab37c3a39bf3253661/compiler/compile/OMRSymbolReferenceTable.cpp#L1656
it's prevented to reuse temps with known obj index.

vijaysun-omr:
Thanks @cathyzhyi Do we need to merge that change at the same time as this one to reduce the likelihood of functional bugs if this one went in first ?

cathyzhyi:
> Thanks @cathyzhyi Do we need to merge that change at the same time as this one to reduce the likelihood of functional bugs if this one went in first ?

Yea, actually it makes more sense to move that small piece of code to this PR to make sure it 100% safe. Just did that.

vijaysun-omr:
Thanks for adding that change in. Does the inliner do recycling of temps on it's own, outside of findAvailableAuto in the symbol reference table ? If so, does the inliner need to be changed too ?

vijaysun-omr:
@genie-omr build all

cathyzhyi:
> Thanks for adding that change in. Does the inliner do recycling of temps on it's own, outside of findAvailableAuto in the symbol reference table ? If so, does the inliner need to be changed too ?

Inliner uses the same API to find temps to reuse so changing `findAvailableAuto` takes care of inliner as well.

PR #3730, opened 2019-04-09
Fix the signature of 'proc_setattr'

keithc-ca:
It should be:
>  int proc_setattr(pid_t, procattr_t*, size_t);


charliegracie:
@genie-omr build aix

charliegracie:
@genie-omr build plinux

PR #3731, opened 2019-04-09
Fix omrvmem_free_memory

bragaigor:
Some users store the identifier in the allocated
memory call from vmem_reserve_memory(). We therefore store the
mode in a temp, clear the identifier, then free the memory.

Add test against use-after-free memory usage.
One such example is the usage from J9JNIRedirectionBlock
where J9PortVmemIdentifier is put into a memory region
allocated using j9vmam_reserve_memory()

Update omrvmem_free_memory for aix, linux and zos390

### DIFF
--- a/fvtest/porttest/omrvmemTest.cpp
+++ b/fvtest/porttest/omrvmemTest.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1991, 2018 IBM Corp. and others
+ * Copyright (c) 1991, 2019 IBM Corp. and others
  *
  * This program and the accompanying materials are made available under
  * the terms of the Eclipse Public License 2.0 which accompanies this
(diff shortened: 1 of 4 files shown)


youngar:
@genie-omr build all

youngar:
In OpenJ9, memory for `J9JNIRedirectionBlock` is allocated using `j9vmem_reserve_memory()`. After allocating the memory they copy the `J9PortVmemIdentifier` into the reserved memory.  When `j9vmem_free_memory()` is called, we free the memory and then attempt to read flags from `J9PortVmemIdentifier`, resulting in a use-after-free bug.

youngar:
@genie-omr build all

youngar:
@genie-omr build all

youngar:
small problem:
```sh
23:03:00  omrvmemTest.cpp: In member function ‘virtual void PortVmemTest_vmem_test_free_memory_Test::TestBody()’:
23:03:00  omrvmemTest.cpp:416:7: error: unused variable ‘allocName’ [-Werror=unused-variable]
23:03:00    char allocName[allocNameSize];
23:03:00         ^
23:03:00  cc1plus: all warnings being treated as errors
```

youngar:
@genie-omr build all

youngar:
AIX and ZOS have similar use-after-free bugs in their `omrvem_free()` implementations, the rest of the platforms look okay.

youngar:
@genie-omr build all

youngar:
@charliegracie can you give this another set of eyes?

youngar:
@genie-omr build all

PR #3733, opened 2019-04-10
Adopt new read/write barrier kind

0dvictor:
eclipse/omr#3168 has introduced new read/write barrier kind; and known
downstream projects have adopted it; therefore OMR is fully moving to
this new read/write barrier kind.

### DIFF
--- a/compiler/codegen/OMRCodeGenPhase.cpp
+++ b/compiler/codegen/OMRCodeGenPhase.cpp
@@ -439,7 +439,7 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator * 
    {
    TR::Compilation *comp = cg->comp();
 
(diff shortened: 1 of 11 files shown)


compiler/codegen/OMRCodeGenPhase.cpp
```diff
@@ -439,7 +439,7 @@ OMR::CodeGenPhase::performSetupForInstructionSelectionPhase(TR::CodeGenerator *
    {
    TR::Compilation *comp = cg->comp();
 
-   if (TR::Compiler->target.cpu.isZ() && TR::Compiler->om.shouldGenerateReadBarriersForFieldLoads())
+   if (TR::Compiler->target.cpu.isZ() && TR::Compiler->om.readBarrierType() != gc_modron_readbar_none)
```
fjeremic: @NigelYiboYu FYI if this conflicts with your change in #3680. Similarly for other Z related changes.

Leonardo2718:
@genie-omr build all

PR #3732, opened 2019-04-10
Remove uses of OMR_INTERP_COMPRESSED_OBJECT_HEADER

gacholio:
Replace with OMR_GC_COMPRESSED_POINTERS. In every build, either both or
neither the flags are set. Removal of the flag definitions will follow
later.

[ci skip]

Signed-off-by: Graham Chapman <graham_chapman@ca.ibm.com>

PR #3734, opened 2019-04-10
CMake: remove use of OMR_INTERP_COMPRESSED_OBJECT_HEADER

dnakamura:
see #3732


charliegracie:
@genie-omr build all

charliegracie:
I believe this is ready to go? It should not break any downstream projects correct? What if downstream projects attempt to set the flag?

dnakamura:
Yes works fine. If any downstream projects set this flag, it simply does nothing. The only change is that `OMR_INTERP_COMPESSED_OBJECT_HEADER` will no longer be defined in `omrcfg.h`. However this does not appear to be used anyway

PR #3736, opened 2019-04-10
Improved Value Propagation for Constant Arrays

andrewcraik:
This change adds additional handling in value propagation for known
object arrays which have constant content. Enhancemetns include
folding of element access as well as folding of arraylength
operations.

### DIFF
--- a/compiler/optimizer/VPHandlers.cpp
+++ b/compiler/optimizer/VPHandlers.cpp
@@ -1151,7 +1151,6 @@ TR::Node *constrainBCDAggrLoad(OMR::ValuePropagation *vp, TR::Node *node)
 TR::Node *constrainAnyIntLoad(OMR::ValuePropagation *vp, TR::Node *node)
    {
    TR::DataType dataType = node->getDataType();
-
    // Optimize characters being loaded out of the values array of a constant string
    //
    if (dataType == TR::Int16 && node->getOpCode().isIndirect() &&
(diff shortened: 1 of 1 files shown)


vijaysun-omr:
@genie-omr build all

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1160,21 +1159,49 @@ TR::Node *constrainAnyIntLoad(OMR::ValuePropagation *vp, TR::Node *node)
...
-         TR::VPConstraint *baseVPConstraint = vp->getConstraint(base, isGlobal);
-         if (baseVPConstraint && baseVPConstraint->isConstString())
+         if (array->getOpCode().isIndirect())
             {
-            TR::VPConstString *constString = baseVPConstraint->getClassType()->asConstString();
+            TR::Node* base = array->getFirstChild();
+            TR::VPConstraint *baseVPConstraint = vp->getConstraint(base, isGlobal);
+            if (baseVPConstraint)
+               {
+               if (baseVPConstraint->isConstString())
```
vijaysun-omr: Is this case still needed if a known object constraint is going to be present for a constant string anyway ?

andrewcraik: Yes so it is possible to get const string when we don't have a known object. A constant string simply means we know what data is going to be in the string, not the identify of the string while a know object says we have the object identity and so can inspect it. We need to maintain both and perhaps generalizes const string to const object or something like that down the line.

compiler/optimizer/VPHandlers.cpp
```diff
@@ -1160,21 +1159,49 @@ TR::Node *constrainAnyIntLoad(OMR::ValuePropagation *vp, TR::Node *node)
...
+            TR::KnownObjectTable *knot = vp->comp()->getKnownObjectTable();
+            TR::VPKnownObject *kobj = arrayVPConstraint->getKnownObject();
+            if (knot && kobj)
                {
-               vp->replaceByConstant(node, TR::VPIntConst::create(vp, ch), true);
-               return node;
+               TR::KnownObjectTable::Index idx = kobj->getIndex();
+               if (kobj->isArrayWithConstantElements(vp->comp()))
+                  {
+                  TR::VMAccessCriticalSection constrainAnyIntLoadCriticalSection(vp->comp(),
```
Leonardo2718: All builds are failing compilation at this line with:

```
/home/jenkins/workspace/PullRequest-linux_x86-64/compiler/optimizer/VPHandlers.cpp:1195:19: error: ‘VMAccessCriticalSection’ is not a member of ‘TR’
                   TR::VMAccessCriticalSection constrainAnyIntLoadCriticalSection(vp->comp(),
                   ^
```

andrewcraik: Fixed

vijaysun-omr:
@genie-omr build all

PR #3737, opened 2019-04-11
 Add support for new signals

babsingh:
Support for the following 10 new signals: `SIGPIPE, SIGALRM, SIGCHLD, SIGALRM, SIGUSR1, SIGUSR2, SIGURG, SIGXCPU, SIGVTALRM and SIGPROF`, is added.


babsingh:
Using the `31st` bit is equivalent to hitting the bound for `uint32_t` (2^32); overflow will happen if other bits are set along with the `31st` bit in a `uint32_t` bit-mask. So, I couldn't use the `31st` bit (0x80000000) in `OMRPORT_SIG_FLAG_SIG*` flags. 

Also, functions such as `omrsig_set_async_signal_handler` allow multiple signals to be specified in a bit-mask; using the `31st` bit in the bit-mask along with other bits will cause overflow since the bit-mask is stored in `uint32_t`.

babsingh:
@babsingh (test email notification)

babsingh:
Reviewers: @charliegracie @youngar. This is a low impact change; it can be quickly reviewed and merged. I have tested these changes with one of the downstream projects (OpenJ9), and no issues were seen. Also, a user is impacted by missing signal support. So, it will be nice to have these changes merged soon.

fyi: @DanHeidinga @pshipton

youngar:
@genie-omr build all

PR #3739, opened 2019-04-12
CMake: Fix option handling for MSVC

dnakamura:
CMAKE_CONFIGURATION_TYPES is only set for multi-configuration generators
(eg Visual Studio). For single configuration generators (eg makefiles)
this will be empty, but we still need to remove configuration specific
flags.


charliegracie:
@genie-omr build win

